PR #722, opened 2020-05-06
PT-BR Translation

cyruzin:
Translation for brazilian portuguese.

### DIFF
--- a/translations/pt-br.po
+++ b/translations/pt-br.po
@@ -1,0 +1,691 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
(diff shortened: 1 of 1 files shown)


AJenbo:
The translations have now been intergrated:
image

(the incorrectly rendered non a-z symbols Is a known issue an will be solved later)

I have also added the shop menus to the translation file if you are up for working some more on it:
https://github.com/diasurgical/devilutionX/blob/master/Translations/pt-BR.po

julealgon:
why the missing translations in that screenshot?

AJenbo:
Because only 47% of the strings have been translated, the translation was done before we had any hellfire support. Also there was a bug that caused some garbled text, it and a partial solution to the non English symbols has been applied.

image

Still this translation needs to be updated and someone needs to finish https://github.com/diasurgical/devilutionX/issues/1686

PR #726, opened 2020-05-09
Add Croatian translation

muzena:


AJenbo:
The translations have now been intergrated:
image

I have also added the shop menus to the translation file if you are up for working some more on it:
https://github.com/diasurgical/devilutionX/blob/master/Translations/hr.po

AJenbo:
Hey @muzena we have finally implemented the needed fonts for rendering Croatian so now would be an excellent time to resume the work on the translation :)

image

muzena:
@AJenbo 
Ok, thanks.
I forgot procedure for how to translate, can I enable localization/translations during compilation from git master?
I compile from git master and get error:

Error
Failed to open file:
fonts\24-00.bin

File not found

AJenbo:
Remember to copy devilutionx.mpq from Packaging/resources if you have an old copy.

To switch to Croatian you must se the language to `hr` in diablo.ini

When you save the translation in Poedit it will create hr.mo, you just move this file next to the game and then the translation will be updated when you start the game.

PR #730, opened 2020-05-13
Xbox port

AJenbo:
This is a rebase of https://github.com/marty28/DiabloX by @marty28 on our master.

Unfortunately, Visual Studio 2003 + XDK appears to be the only option for things to work properly on Xbox atm. Eventually, we might be able to move it to [OpenXDK](https://github.com/XboxDev/nxdk), but for now, it has some outstanding issues. In effect, this means that we are limited to C++03 for supporting the port.

Remaning issues:
- [x] Crash if hero name is longer then 7
- [x] Clicking on list items in the menus with the mouse (focuse but no apply)
- [x] Scrollbars do not scroll
- [x] Credits are not rendered
- [x] Crash when going back from the hero menu
- [x] Last item (new hero) does not appear on the list
- [x] Last item on lists are incorrectly highlighed
- [x] Title in error dialogs as an incorrect shadow color
- [ ] mpqapi.cpp needs to be ported, either the main has to be ported to C++03, or the one used here has to be verified on all platforms.
- [ ] Move file Xbox path code to SDL_GetBasePath implementation
- [ ] Clean up unessesery `#ifdef _XBOX`
- [x] music and dialogs should be streamed instead of preloaded to save on memory usage

Here is a video of the game running on the original Xbox: https://www.youtube.com/watch?v=h1_g9-TK3Kc


marty28:
Hi Anders,

I have now tested this PR, it builds and plays fine.

Correct, Visual Studio 03 compiler, unfortunately this is the newest compiler the official OG Xbox SDK supports so I'm unable to build with VS2008 or newer. There is another unofficial Xbox SDK that has a newer Clang compiler but still early alpha and not usable anytime soon IMHO.

Cheers

AJenbo:
Thanks for testing. Could you also test the https://github.com/diasurgical/devilutionX/tree/xbox-next branch where i have merged the latest master.

Are you referring to https://github.com/XboxDev/nxdk ? The project looks really promising with it having CI for it's samples and an SDL2 port already. It appears to be under active development so hopefully they will get in to a suitable state before too long.

marty28:
No worries, tested the xbox-next branch and found two build errors:

SourceX\DiabloUI\fonts.h : Line 20 - unmatched #endif
SourceX\controls\menu_controls.cpp : Line 49 - Cannot convert parameter 1 from int* to SDLKey*

After removing the #endif and changing "auto sym" to "SDLKey sym" it builds and plays well.

Yep NXDK, I agree it does look promising, I tried to use it at the start of the port and found it still needs to mature a bit tho, I'm keeping an eye on it and look forward to using it.

Cheers

AJenbo:
Are the .hpp file name in Radon an issue for VC2003?

marty28:
I just reverted to .hpp and can confirm this is not an issue for VS03.

SourceX/storm/storm.cpp
```diff
@@ -55,10 +56,14 @@ void GetBasePath(char *buffer, size_t size)
 
 	snprintf(buffer, size, "%s", path);
 	SDL_free(path);
+#else
+	snprintf(buffer, size, "%s", "D:\\");
```
AJenbo, line 60: @marty28 this should be moved to the SDL2 implementation.

SourceS/sdl2_to_1_2_backports.h
```diff
@@ -1,13 +1,15 @@
 #pragma once
```
AJenbo, line 1: Afaik the Xbox uses SDL2, so none of the changes to this file should be nessesery?

3rdParty/StormLib/src/FileStream.cpp
```diff
@@ -18,8 +18,8 @@
 #include "StormCommon.h"
 #include "FileStream.h"
 
-#ifdef _MSC_VER
-#pragma comment(lib, "wininet.lib")             // Internet functions for HTTP stream
+#if 1//def _MSC_VER
```
AJenbo, line 21: @marty28 the changes to StormLib are a bit rough handed, could you try and clean it up a bit so that it doesn't break other platforms?

3rdParty/StormLib/src/FileStream.cpp
```diff
@@ -807,6 +810,7 @@ static void BaseHttp_Close(TFileStream * pStream)
 #endif
 }
 
+#ifndef NONET
```
AJenbo, line 813: NONET is a DevilutionX concept so it shouldn't be used inside of StormLib. I think the right solution would be to make `PLATFORM_WINDOWS && PLATFORM_XBOX` result in the unsupported scenario.

Source/mainmenu.cpp
```diff
@@ -155,10 +155,20 @@ BOOL mainmenu_init_menu(int type)
...
+#if NONET
+void UiSelOkDialog(const char *title, const char *body, bool background);
+#endif
+
 BOOL mainmenu_multi_player()
 {
 	gbMaxPlayers = MAX_PLRS;
+
+#if NONET
+	UiSelOkDialog(NULL, "There is no network support in this Xbox port at the momment...sorry", false);
```
AJenbo, line 167: Is the Xbox not able to run the loopback multiplayer option?

Source/mpqapi.cpp
```diff
@@ -3,349 +3,44 @@
  *
```
AJenbo, line 3: Looks like the unmodified Devilution version of mpqapi.cpp was used for this port (likely to save time on porting DevilutionX's C++11 to C++03). This is likely one of the major obstacles to merging the Xbox code. I have hacked together a cross-platform version from an older version of DevilutionX before we did a major rework on this code (afaik this was needed for stable Amiga support), this should be good enough to get us 90% of the way before we have to address it.

SourceX/storm/storm.cpp
marty28: I'm currently using SDL1.2, we also have the port of SDL2 I did for the MS Xbox SDK recently that we could use but it's slower and not as mature as 1.2 on Xbox.

I have not tested it with SDL2 since DevilutionX started supporting 1.2. I need to test it again, it might just work.

Maybe I should move this into SDL_GetBasePath?

SourceS/sdl2_to_1_2_backports.h
marty28: As above, I'm currently using 1.2 but we do have both.

3rdParty/StormLib/src/FileStream.cpp
marty28: yeah sorry, I would have kept things a bit nicer but at the time I did not expect it to get merged back.Thanks for doing the cleanup :)

Source/mainmenu.cpp
marty28: I think the Xbox can do loopback multiplayer I was just focused on the single player experience at the time.

No worries, I should be able to get the multiplayer dialogs converted this weekend.

Source/mpqapi.cpp
marty28: Yep was taken from unmodified Devilution version. I have made the changes required for Xbox to build your cross-platform version.

Below are the changes to get the latest head revision building on Xbox, tested and working well: https://github.com/marty28/devilutionX/commit/39ef7de1c85af787c25c4b65dc8f78fcfde390ea

Cheers

SourceX/storm/storm.cpp
AJenbo: Ok, that explains some things. Using SDL1.2 makes sens. Yeah this should be moved to the our SDL1.2 SDL_GetBasePath compat.

SourceS/sdl2_to_1_2_backports.h
AJenbo: ok, i'll revert some of the changes then. I'm thinking the Xbox port won't be working at this point since i just reverted everything about this file.

3rdParty/StormLib/src/FileStream.cpp
AJenbo: I think there are still some unessesery ifdefs left in there that you could maybe clean up :)

Source/mainmenu.cpp
AJenbo: Sounds good, I will focuse on other areas then.

Source/drlg_l2.cpp
```diff
@@ -357,8 +357,13 @@ static void DRLG_L2PlaceRndSet(BYTE *miniset, int rndper)
 			}
 			kk = sw * sh + 2;
 			if (found == TRUE) {
+#ifdef _XBOX
+				for (yy = max(sy - sh, 0); yy < min(sy + 2 * sh, DMAXY) && found == TRUE; yy++) {
+					for (xx = max(sx - sw, 0); xx < min(sx + 2 * sw, DMAXX); xx++) {
+#else
 				for (yy = std::max(sy - sh, 0); yy < std::min(sy + 2 * sh, DMAXY) && found == TRUE; yy++) {
```
AJenbo, line 364: @marty28 could you check if defining `#define NOMINMAX 1` allows you to use `std::min` in VS03?

marty28: @AJenbo I found it works. I made a commit reverting to the STL. Thx https://github.com/marty28/devilutionX/commit/bef88e29616df745ffc564a4fc446a186bec3c1d

AJenbo: excelent, that will gets us a bit closer :)

SourceX/DiabloUI/selyesno.cpp
```diff
@@ -24,18 +24,16 @@ void selyesno_Free()
 		UiListItem* pUIListItem = vecSelYesNoDialogItems[i];
 		if(pUIListItem)
 			delete pUIListItem;
-
-		vecSelYesNoDialogItems.clear();
 	}
+	vecSelYesNoDialogItems.clear();
 
 	for(int i = 0; i < (int)vecSelYesNoDialog.size(); i++)
```
glebm, line 30: There are some weird casts here. Perhaps use a `size_t` index?

```c++
for (std::size_t i = 0; i < vecSelYesNoDialog.size(); ++i)
```

(or `size_t` instead of `std::size_t` if the XBox compiler doesn't like it)

AJenbo: `std::size_t` was used in a few other places so I made the change to all the remaning loops as well as some other cleanups and a partial fix to the list not being clickable.

AJenbo:
@marty28 can I have you look at what is causing this error on Switch and Amiga:
```
/SourceX/dvlnet/abstract_net.cpp:37:17: error: expected primary-expression before ';' token
  delete loopback;
```

marty28:
> @marty28 can I have you look at what is causing this error on Switch and Amiga:
> 
> ```
> /SourceX/dvlnet/abstract_net.cpp:37:17: error: expected primary-expression before ';' token
>   delete loopback;
> ```
> 
> The line is only for NONET targets, but also not for _XBOX, which is probably wrong at this point.

@AJenbo I had a look and tried to get it to compile on Xbox as I agree we need it there as well, had the same problem. I found it needs a pointer to the be deleted. I tried to create a abstract_net class member "static abstract_net* m_pConn;" turns out it does not like been used like this statically, when global it works fine using "abstract_net* g_pConn" for the object been created and deleted. Is this an acceptable fix? Or we need to keep the global namespace as clean as possible? Cheers

AJenbo:
I think that will be fine, diablo uses globals heavily anyway

marty28:
Cool, have made a PR https://github.com/diasurgical/devilutionX/pull/743

Cheers

AJenbo:
@marty28 could you check if this branch now builds correctly on Xbox again?

AJenbo:
Cleaned up the commits and pushed it to a new branch/PR: https://github.com/diasurgical/devilutionX/pull/760

AJenbo:
Nased on a quick chat with the XboxDev group it appears that nxdk is now fare enough along to be able to support the game. Hopefully this means that we can switch to it and have automated builds.

glebm:
What's the status of nxdk today?

https://github.com/XboxDev/nxdk says:

> Modern C / C++ standards and compiler features.

Hopefully that means we can drop support for C++98 and go back to C++11?

AJenbo:
SDL_Mixer still hasent landed :(
https://github.com/XboxDev/nxdk/tree/master/lib/sdl

last i heard they only miss mod and mp3 support, and there are a forks that we might be able to use.
https://github.com/fgsfdsfgs/nxdk/tree/dev/lib/sdl
https://github.com/JayFoxRox/nxdk/tree/sdl_mixer/lib/sdl

I haven't heard from @marty28 in a while so i consider upstreaming of this port stalled and non-blocking for other activities. last i heard there was an issue running this branch and he was considering redoing it using nxdk.

Since we have no CI that checking if we break c++03 I don't think we adhere strictly to it. TBH use what ever the build in CI can handle but maybe don't deliberately refactor things in a way that will be hard to dumb down later if needed.
If you can figure out what limits there are for doing a PSP port that would be nice to stick to as well.

AJenbo:
Also now that we support music streaming it reduces the SDL_Mixer needs to just .wav support and removes the need for converting the music to ogg/disabling music.

PR #732, opened 2020-05-15
Fix building with latest SDL2 on Switch

rsn8887:
See also discussion here:
https://github.com/devkitPro/pacman-packages/issues/143


AJenbo:
thanks, having this not build has been a nightmare :)

PR #737, opened 2020-05-17
Converted connection UI to C++03, switched to STL for min/max

marty28:
Tested the Xbox with loopback and found it gets to the loading screen but then we get a crash in GetArtStringWidth, it's getting a bad pointer for the string passed in. Have been unable to find the cause yet.


AJenbo:
I think it was simply that selgame.cpp still wasn't converted and so the menu items were invalid when getting to the difficulty selection screen. I did the conversion and I can now start both TCP and Loopback games. Still getting a crash for joining a TCP game though.

marty28:
Oh yep makes sense, thanks for finishing the conversion on selgame.cpp. Tested and loopback is now working well on Xbox. I need to look at how different winsock is on Xbox, might be minimal changes to get it working.

Issue #720, opened 2020-05-02
Option to overwrite save folder


AJenbo:
Similar to being able to overwrite the standard data folder some users would like the option to overwrite the standard save folder. This would enable having the game run from a self-contained folder, for example a USB.

This is where the `--data-dir` code starts:
https://github.com/diasurgical/devilutionX/blob/791aa60cab453899dfaa7456cde1e8c23fcbee79/Source/diablo.cpp#L387

The new option should be `--save-dir`

AJenbo:
This has now been implemented by @ahwayakchih, so for anyone wanting to run the game off of a self-contained USB you can now make a shortcut that starts the game with `--save-dir ./` as a parameter.

PR #740, opened 2020-05-20
Implement `--save-dir` CLI option (#720)

ahwayakchih:
This implements support for `--save-dir` (as described in #720).

Unfortunately `ini` file is initialized as static variable, before we can change `prefPath`: https://github.com/diasurgical/devilutionX/blob/0fddb655e3ef855b43166e720252ba1736197ccb/SourceX/storm/storm.cpp#L39

`SourceX/storm/storm.cpp` and/or `3rdParty/radon/File` may need to be modified to workaround that (i was thinking about initializing `radon::File` only at first access to ini file, but i am not sure if that would be ok).

@AJenbo would further modifications to SourceX/storm/storm.cpp` be OK?

### DIFF
--- a/3rdParty/Storm/Source/storm.h
+++ b/3rdParty/Storm/Source/storm.h
@@ -5,6 +5,7 @@
 namespace dvl {
 
 extern std::string basePath;
+extern std::string prefPath;
 
 // Note to self: Linker error => forgot a return value in cpp
 
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -352,6 +352,7 @@ static void print_help_and_exit()
 	printf("    %-20s %-30s\n", "-h, --help", "Print this message and exit");
 	printf("    %-20s %-30s\n", "--version", "Print the version and exit");
 	printf("    %-20s %-30s\n", "--data-dir", "Specify the folder of diabdat.mpq");
+	printf("    %-20s %-30s\n", "--save-dir", "Specify the folder of save files");
 	printf("    %-20s %-30s\n", "-n", "Skip startup videos");
 	printf("    %-20s %-30s\n", "-f", "Display frames per second");
 	printf("    %-20s %-30s\n", "-x", "Run in windowed mode");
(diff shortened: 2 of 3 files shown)


AJenbo:
Thanks for doing this, looks good.

We actually intended to change that very code as it is required for Android support. You can cherry pick the following commit from the android-wip branch for this: https://github.com/diasurgical/devilutionX/commit/0890e37bcd6f50515b539203f6618777b46163b7

The reason that hasn't been merged yet is that it caused instability on RG300, but that was likely just a random result of OOB issues that were more favorably aligned without the change. Hopefully at this point, we have fixed whatever was the cause so that it can be safely merged.

Only loading the ini file at first access sounds nice, but since it's accessed during the initialization for audio volume, fullscreen state and video state there no point in doing so. We know we will always need it very early in the process.

ahwayakchih:
@AJenbo thanks for quick response :). I cherry picked the commit you suggested, but... won't this reload ini file every time a single value is accessed? So, on settings screen, with more than 3 values used, ini will be reloaded multiple times, right?

I did not write C/C++ for a long time now, but isn't it possible to set static variable inside a function, so it's still lazy loaded, but only once? So we could have something like this:

```cpp
radon::File& getIni() {
  static radon::File ini(getIniPath());
  return ini;
}
```

and then:

```cpp
bool getIniValue(const char *sectionName, const char *keyName, char *string, int stringSize, int *dataSize)
{
	radon::Section *section = getIni().getSection(sectionName);
```

and:

```cpp
void setIniValue(const char *sectionName, const char *keyName, char *value, int len)
{
	radon::File ini = getIni();
```

ahwayakchih:
@AJenbo i implemented static stuff i described above to prevent reloading ini file every time. I tested it on x86_64 (Linux) - it seems to be working fine. I do not have Android to test :(. I can revert the change (of course), if needed.

AJenbo:
Well the important one is RG300, I'll test it later this week and then hopuflly merge it :)

AJenbo:
Complete level 1 on RG300 with no crashes.

Issue #735, opened 2020-05-15
Add support for looking server up by host name


kai4785:
**Describe**
When choosing a multiplayer server, a hostname cannot be used and users are forced to use an IP address.

**To Reproduce**
From the top level menu, choose:
1) Multi Player
2) Client-Server (TCP)
3) < Choose any player, or create a new one >
4) Enter IP
5) < Enter a hostname like `localhost` >
6) < Enter a password >

At the end, you will be presented with an error screen and the message "Invalid argument".

**Expected behavior**
Allow users to enter hostnames in step 5 above.
As an added bonus, change Step 4 to "Join Server", and the prompt in step 5 to "Enter hostname or IP address."

**Screenshots**
![image](https://user-images.githubusercontent.com/729693/82059357-25f64480-9683-11ea-8233-2faf667fdebc.png)

![image](https://user-images.githubusercontent.com/729693/82059396-34dcf700-9683-11ea-993f-7c300345c16d.png)

![image](https://user-images.githubusercontent.com/729693/82058627-3b1ea380-9682-11ea-843f-d976f22e8360.png)

**Additional context**

AJenbo:
This will require implementing a DNS lookup and probably pattern matching to test if what was entered was an IP or not. It does look like ASIO has the needed functionality to implement this.

kai4785:
Awesome, thanks for the quick fix! This will make it much easier on my kids playing multiplayer with each other. No more looking up IPs!

AJenbo:
@ahwayakchih in case you didn't see the message above :)

ahwayakchih:
Thanks :D

PR #741, opened 2020-05-21
#735 - Use asio to resolve host name (or IP)

ahwayakchih:
Use asio API to resolve host name (and/or IP), as suggested by @AJenbo (see #735).
This is for TCP only, at least for now. Tested with "localhost" and "127.0.0.1" (both cases connecting to local game started in separate process).

I don't like converting `int` to `string` for `default_port`. It should probably be defined in one place for both types (int and str), but i'm not sure where/how to implement that (please let me know :).


SourceX/DiabloUI/selgame.cpp
```diff
@@ -73,7 +73,7 @@ UiItem ENTERIP_DIALOG[] = {
 	SELUDPGAME_TITLE,
 	SELUDPGAME_DESCRIPTION_LABEL,
 	SELGAME_DESCRIPTION,
-	UiArtText("Enter IP", { PANEL_LEFT + 305, 211, 285, 33 }, UIS_CENTER | UIS_BIG),
+	UiArtText("Enter hostname or IP", { PANEL_LEFT + 305, 211, 285, 33 }, UIS_CENTER | UIS_BIG),
```
AJenbo, line 76: Maybe more user frindly/generic:

```suggestion
	UiArtText("Enter address", { PANEL_LEFT + 305, 211, 285, 33 }, UIS_CENTER | UIS_BIG),

```

ahwayakchih: Thanks, suggestion applied :).

ahwayakchih:
Thanks, resolved :)

Issue #725, opened 2020-05-08
Crash in SDL Cleanup


NiteKat:
**Describe**
Game crashes when exiting via the main menu after performing specific steps on line 5432 of SDL_malloc.c. Crash seems to be silent, as no pop-ups happen. Only noticed it when running devilutionx.exe out of VS 19 in both x64-Debug and x64-Release configurations (cannot test x86-Debug and x86-Release as VS 19 won't let me run those for some reason). Crash is "Exception thrown: read access violation. it was 0xFFFFF..."

**To Reproduce**
1. Clone fresh copy of DevilutionX repo.
2. Open in Visual Studio 2019.
3. Choose either x64-Debug or x64-Release configuration.
4. Set devilutionx.exe as Startup Item.
5. Click selected Startup Item to run.
6. Create a multiplayer game with either a new character or existing character.
7. Walk to and click on Griswold.
8. Select Buy Basic Items.
9. Exit store.
10. Open menu, and chose New Game.
11. Back out to the main menu.
12. Choose exit game. The game will crash on the line in Description.

**Expected behavior**
Game should cleanly exit without crashing during cleanup.

**Screenshots**
![crash](https://user-images.githubusercontent.com/28605404/81433414-855bce00-9132-11ea-998d-8ed268ee34f1.png)

**Additional context**
N/A

AJenbo:
Can you see what is calling SDL_free in this case?

NiteKat:
This is the Call Stack at time of crash. Let me know if you meant something else.
ucrtbased.dll!check_bytes(const unsigned char * const first, const unsigned char value, const unsigned __int64 size) Line 194
	at minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp(194)
ucrtbased.dll!is_block_an_aligned_allocation(const void * const block) Line 252
	at minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp(252)
ucrtbased.dll!free_dbg_nolock(void * const block, const int block_use) Line 886
	at minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp(886)
ucrtbased.dll!_free_dbg(void * block, int block_use) Line 1030
	at minkernel\crts\ucrt\src\appcrt\heap\debug_heap.cpp(1030)
ucrtbased.dll!free(void * block) Line 32
	at minkernel\crts\ucrt\src\appcrt\heap\free.cpp(32)
[...]

Issue #717, opened 2020-04-25
[Linux ppc64le] SDL malloc segfault upon exiting the game


runlevel5:
**Describe**
Seems to me SDL malloc has issues when the game exits.

**To Reproduce**
1. Fetch master branch / commit 936b618338c02408426c2a36d34471c0ac804211
2. Start then quit the game from the main menu

**Expected behavior**
Clean exit with no errors

**Actual behaviour**
Segmentation fault

**Additional context**

```
$ coredumpctl debug 2103612
           PID: 2103612 (devilutionx)
           UID: 1000 (tle)
           GID: 1000 (tle)
        Signal: 11 (SEGV)
     Timestamp: Sat 2020-04-25 22:24:28 AEST (3min 56s ago)
  Command Line: ./devilutionx
    Executable: /home/tle/src/devilutionX/build/release/devilutionx
 Control Group: /user.slice/user-1000.slice/user@1000.service/apps.slice/apps-org.gnome.Terminal.slice/vte-spawn-d902afd2-7af0-4859-bd22-c20c1fd7c7d9.scope
          Unit: user@1000.service
     User Unit: vte-spawn-d902afd2-7af0-4859-bd22-c20c1fd7c7d9.scope
         Slice: user-1000.slice
     Owner UID: 1000 (tle)
       Boot ID: 29003e772cb44e319c84ccbe6ad7a138
    Machine ID: 5632f07729a648c49d05933910ac9490
      Hostname: orion.dev
       Storage: /var/lib/systemd/coredump/core.devilutionx.1000.29003e772cb44e319c84ccbe6ad7a138.2103612.1587817468000000000000.lz4
       Message: Process 2103612 (devilutionx) of user 1000 dumped core.
                
                Stack trace of thread 2103612:
                #0  0x00007fff84da9ce8 __GI___libc_free (libc.so.6 + 0xa9ce8)
                #1  0x00007fff8516a26c SDL_free_REAL (libSDL2-2.0.so.0 + 0xba26c)
                #2  0x00007fff8510cc28 SDL_free (libSDL2-2.0.so.0 + 0x5cc28)
                #3  0x00007fff852ac9e8 Mix_FreeChunk (libSDL2_mixer-2.0.so.0 + 0xc9e8)
                #4  0x0000000010047c18 n/a (/home/tle/src/devilutionX/build/release/devilutionx + 0x47c18)
[...]

PR #744, opened 2020-05-27
Prevent double free of sound chunk in storm

ahwayakchih:
This does not check why `SFileFreeChunk` and/or `SFileDdaDestroy` were called for the same chunk of data. It simply makes sure that once the chunk is freed it's set to `NULL`, so next call will not trigger `Mix_FreeChunk` and error.

This fixes cleanup crash i could reproduce every time by simply starting/loading game, gossiping with NPC/Townsfolk and quitting game (as described at https://github.com/diasurgical/devilutionX/issues/717#issuecomment-634628920).

This may fix both #717 and #725.

### DIFF
--- a/SourceX/storm/storm.cpp
+++ b/SourceX/storm/storm.cpp
@@ -42,7 +42,7 @@ radon::File& getIni() {
   return ini;
 }
 
-static Mix_Chunk *SFileChunk;
+static Mix_Chunk *SFileChunk = NULL;
 
 void GetBasePath(char *buffer, size_t size)
 {
@@ -102,6 +102,9 @@ BOOL SFileDdaBeginEx(HANDLE hFile, DWORD flags, DWORD mask, unsigned __int32 lDi
 		SDL_Log(SDL_GetError());
 		return false;
 	}
+	if (SFileChunk) {
+		SFileFreeChunk();
+	}
 	SFileChunk = Mix_LoadWAV_RW(rw, 1);
 	free(SFXbuffer);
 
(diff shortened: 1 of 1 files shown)


ahwayakchih:
@AJenbo don't know why (CircleCI requires login), but testflow failed on amigaos :(. Travis passed but  status was not updated on this PR page for some reason.

AJenbo:
travis has a tendency to fail randomly, amiga is failing because it can't download freetype from https://mirrors.sarata.com/non-gnu/freetype/freetype-2.10.1.tar.gz

SourceX/storm/storm.cpp
```diff
@@ -102,6 +102,9 @@ BOOL SFileDdaBeginEx(HANDLE hFile, DWORD flags, DWORD mask, unsigned __int32 lDi
 		SDL_Log(SDL_GetError());
 		return false;
 	}
+	if (SFileChunk) {
+		SFileFreeChunk();
```
AJenbo, line 106: did we make sure to stop the sound first?

ahwayakchih: Good point. No i did not add checking for that. Actually i added those lines just in case.
Changes in other two functions were enough to prevent crashes.

Should i copy solution from `SoundSample::Stop()` to check if chunk is playing, and if it is, stop it first? Or will simple `Mix_GetChunk(0) == SFileChunk && Mix_Playing(0)` be enough, since AFAIK storm functions use hardcoded channel 0? Or maybe just calling `SFileDdaEnd()` before `SFileFreeChunk()` in this case will be enough?

ahwayakchih: @AJenbo I just added a call to `SFileDdaEnd(hFile);`, for now. Let me know if something more is needed :).

AJenbo:
Thanks for digging into this issue :+1:

ahwayakchih:
Thanks :).

PR #746, opened 2020-05-31
Set Single Player Difficulty via ini

genevanmeter:
Adds "SP Difficulty" to ini which is used to set gnDifficulty when Single Player is selected.
Eliminates the need to do the Multiplayer New Game bug originally used before.  Multiplayer is unaffected as it will set gnDifficulty when player selects the difficulty. 
Going back and selecting a Single Player game will again read the ini and set gnDifficulty back to the ini setting.


AJenbo:
Thanks for the PR, but we plan to add the difficulty menu to the single player menu like it was in Hellfire. This will be much more intuitive an user friendly to players.

PR #751, opened 2020-06-19
Add an option to choose a display where to run the game

Meetoul:
Signed-off-by: Ihor Ivanov <toulmee@gmail.com>


AJenbo:
Thanks for the PR.
Looks like this is SDL2 only, could you implement it on the SDL1.2 side as well, or make sure not to present the option for SDL1.2 builds?

PR #753, opened 2020-06-20
Fix PPA debian building version and git commit

muzena:


muzena:
@AJenbo

muzena:
@AJenbo 
Why this check switch failed?

AJenbo:
All switch builds fail currently because of changes to the build server that we are not in control of.

PR #756, opened 2020-06-30
add SDL2-static and libstdc++-static fedora libs to readme

Knotschi:
I couldn't build the project without these libs installed

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ sudo apt-get install cmake g++ libsdl2-mixer-dev libsdl2-ttf-dev libsodium-dev
 ```
 ### Installing dependencies on Fedora
 ```
-sudo dnf install cmake glibc-devel SDL2-devel SDL2_ttf-devel SDL2_mixer-devel libsodium-devel libasan libubsan
+sudo dnf install cmake glibc-devel SDL2-devel SDL2_ttf-devel SDL2_mixer-devel libsodium-devel libasan libubsan SDL2-static libstdc++-static
 ```
 ### Compiling
 ```


runlevel5:
@Knotschi which Fedora version are you using? I could still compile successfully without those 2 static libs

Knotschi:
hmm, weird...
I am using fedora 31

without `SDL2-static` installed I am always getting:
```
[  3%] Built target smacker
[  6%] Built target Radon
[  8%] Built target PKWare
[ 16%] Built target StormLib
[ 65%] Built target devilution
make[2]: *** No rule to make target '/usr/lib64/libSDL2main.a', needed by 'devilutionx'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:245: CMakeFiles/devilutionx.dir/all] Error 2
make: *** [Makefile:172: all] Error 2
```

and with `SDL2-static` but without `libstdc++-static` installed I am always getting:
```
[  2%] Built target PKWare
[  5%] Built target Radon
[ 12%] Built target smacker
[ 16%] Built target StormLib
[ 65%] Built target devilution
[ 66%] Linking CXX executable devilutionx
/usr/bin/ld: cannot find -lstdc++
make[2]: *** [CMakeFiles/devilutionx.dir/build.make:834: devilutionx] Error 1
make[1]: *** [CMakeFiles/Makefile2:245: CMakeFiles/devilutionx.dir/all] Error 2
make: *** [Makefile:172: all] Error 2
```
when running `make -j$(nproc)`

AJenbo:
Hmm it should only be Nintendo Switch that makes use of a statically linked SDL2

yuripourre:
@Knotschi you are right, I cannot compile in Fedora 31 anymore.

yuripourre:
@Knotschi @AJenbo 
Replacing (in CMakeLists.txt)
`find_package(SDL2 CONFIG QUIET)` to `find_library(SDL2_LIBRARY NAME SDL2)` worked for me.
I have no idea if that would break other platforms.

yuripourre:
@mr-p666 With the change I mentioned I could compile without static packages. About gcc-c++ I guess you are right, it is missing.

madmann91:
I have a similar problem, but I think there is no need to link statically to anything. Notice that the command to find SDL2 is:
```
find_package(SDL2 CONFIG QUIET)
```
(in `CMakeLists.txt:150` on c009b72d9df9dd5ecf053325f7d27a6a74e5bbe2)

This means that SDL2 is loaded from a config file (coming from a SDL2 installed on the system), not from the `FindSDL2.cmake` file. This is a problem because the semantics of the targets exported by `FindSDL2.cmake` and the config file are different: In particular, the `SDL2::SDL2main` target obtained from the config file is _always_ the `SDL2main.a` file, **even if that file does not exist**. This is a problem because the CMake logic in DevilutionX uses `SDL2::SDL2main` as if it was `SDL2` combined with the potential other dependencies (e.g. `SDL2main`, when present).

To fix this, I recommend to use:
```
find_package(SDL2 MODULE QUIET)
```
This forces the use of `FindSDL2.cmake` instead of relying on the system's SDL2 config file. Doing so fixes this issue for me.

PR #760, opened 2020-07-19
Xbox next

AJenbo:
This is the next step for mainlining the Xbox port, I have cleaned up the commmit history, but it still needs to be tested by @marty28 before we can safly replace the old PR.


glebm:
This port may be possible with OpenXDK now, because:

1. OpenXDK does not have SDL_Mixer => We no longer use `SDL_Mixer`.
2. OpenXDK does not support exceptions => We can now build without exceptions (`set(NONET ON)`).
3. OpenXDK supports modern C++, no need for C++03

We probably don't need any of the patches here anymore, it should be able to build the OpenXDK version like we do any other port.

AJenbo:
Since marty28 appears to have abandoned GitHub I think we can just close this at this point as it was mostly untested, reported to possibly have an issue, and heavily outdated at this point, no CI.

PR #761, opened 2020-07-19
Ui refactor

AJenbo:
Refactor DiabloUI to dynamic C++03 compatible code
    
This will help with porting the game to the original Xbox, and also make it easier to implement dynamic resolutions and translations.

### DIFF
--- a/SourceX/DiabloUI/art_draw.cpp
+++ b/SourceX/DiabloUI/art_draw.cpp
@@ -11,12 +11,12 @@ void DrawArt(int screenX, int screenY, Art *art, int nFrame, Uint16 srcW, Uint16
 	if (screenY >= SCREEN_HEIGHT || screenX >= SCREEN_WIDTH || art->surface == NULL)
 		return;
 
-	SDL_Rect src_rect = {
-		0,
-		static_cast<Sint16>(nFrame * art->h()),
-		static_cast<Uint16>(art->w()),
-		static_cast<Uint16>(art->h())
-	};
+	SDL_Rect src_rect;
+	src_rect.x = 0;
+	src_rect.y = nFrame * art->h();
+	src_rect.w = art->w();
+	src_rect.h = art->h();
+
 	ScaleOutputRect(&src_rect);
 
 	if (srcW && srcW < src_rect.w)
@@ -36,7 +36,8 @@ void DrawArt(int screenX, int screenY, Art *art, int nFrame, Uint16 srcW, Uint16
 		ErrSdl();
 }
 
-void DrawAnimatedArt(Art *art, int screenX, int screenY) {
+void DrawAnimatedArt(Art *art, int screenX, int screenY)
+{
 	DrawArt(screenX, screenY, art, GetAnimationFrame(art->frames));
 }
 
--- a/SourceX/DiabloUI/credits.cpp
+++ b/SourceX/DiabloUI/credits.cpp
@@ -33,31 +33,26 @@ struct SurfaceDeleter {
 	}
 };
 
-using SurfacePtr = std::unique_ptr<SDL_Surface, SurfaceDeleter>;
-
 struct CachedLine {
-	CachedLine() = default;
-
-	explicit CachedLine(std::size_t index, SurfacePtr surface)
-	    : index(index)
-	    , surface(std::move(surface))
-	    , palette_version(pal_surface_palette_version)
+	CachedLine(std::size_t index, SDL_Surface *surface)
 	{
+		m_index = index;
+		m_surface = surface;
 	}
 
-	std::size_t index;
-	SurfacePtr surface;
+	std::size_t m_index;
+	SDL_Surface *m_surface;
 	unsigned int palette_version;
 };
 
-SurfacePtr RenderText(const char *text, SDL_Color color)
+SDL_Surface *RenderText(const char *text, SDL_Color color)
 {
 	if (text[0] == '\0')
 		return NULL;
 	SDL_Surface *result = TTF_RenderUTF8_Solid(font, text, color);
 	if (result == NULL)
 		SDL_Log(TTF_GetError());
-	return SurfacePtr(result);
+	return result;
 }
 
 CachedLine PrepareLine(std::size_t index)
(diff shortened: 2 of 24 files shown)


AJenbo:
@glebm would you be willing to review this. I think there might be a couple of non-ideal parts of the new code. Especially in relation to the credit screen, I'm not sure that it is correctly handeling caching anymore?

SourceX/DiabloUI/credits.cpp
```diff
@@ -154,10 +148,10 @@ class LinesBuffer {
 			empty_ = true;
 	}
 
-	void push_back(CachedLine &&line)
+	void push_back(CachedLine line)
```
AJenbo, line 151: @glebm this is the change that I suspect will have a negative effect on performance.

SourceX/DiabloUI/credits.cpp
```diff
@@ -33,31 +33,26 @@ struct SurfaceDeleter {
...
+	CachedLine(std::size_t index, SDL_Surface *surface)
 	{
+		m_index = index;
+		m_surface = surface;
 	}
 
-	std::size_t index;
-	SurfacePtr surface;
+	std::size_t m_index;
+	SDL_Surface *m_surface;
```
glebm, line 44: `SurfacePtr` handled surface deletion. With a regular pointer, deletion now has to be done manually.

Can add accessor methods and put `delete` there and the destructor.

SourceX/DiabloUI/credits.cpp
glebm: `CachedLine` is small so shouldn't make a difference

SourceX/DiabloUI/credits.cpp
```diff
@@ -236,23 +230,22 @@ void CreditsRenderer::Render()
 	// We use unscaled coordinates for calculation throughout.
 	Sint16 dest_y = VIEWPORT.y - (offset_y - lines_begin * LINE_H);
 	for (std::size_t i = 0; i < lines_.size(); ++i, dest_y += LINE_H) {
-		auto &line = lines_[i];
```
glebm, line 239: Does C++03 not have references?

SourceX/DiabloUI/diabloui.cpp
```diff
@@ -694,91 +720,94 @@ void Render(const UiScrollBar &ui_sb)
...
-void RenderItem(UiItem *item)
+void RenderItem(UiItemBase *item)
 {
 	if (item->has_flag(UIS_HIDDEN))
 		return;
-	switch (item->type) {
+	switch (item->m_type) {
 	case UI_TEXT:
-		Render(&item->text);
+		Render((UiText *)item);
```
glebm, line 755: `static_cast<UiText *>` is better because it at least checks that `UiText` is a subclass of `UiItemBase`.

SourceX/DiabloUI/diabloui.cpp
```diff
@@ -694,91 +720,94 @@ void Render(const UiScrollBar &ui_sb)
...
 		break;
 	case UI_EDIT:
-		Render(item->edit);
+		Render((UiEdit *)item);
 		break;
 	}
 }
 
-bool HandleMouseEventArtTextButton(const SDL_Event &event, const UiArtTextButton &ui_button)
+bool HandleMouseEventArtTextButton(const SDL_Event &event, UiArtTextButton *ui_button)
```
glebm, line 781: Is it necessary to remove `const` here (and throughout)?

SourceX/DiabloUI/art_draw.cpp
```diff
@@ -11,12 +11,12 @@ void DrawArt(int screenX, int screenY, Art *art, int nFrame, Uint16 srcW, Uint16
...
 
-	SDL_Rect src_rect = {
-		0,
-		static_cast<Sint16>(nFrame * art->h()),
-		static_cast<Uint16>(art->w()),
-		static_cast<Uint16>(art->h())
-	};
+	SDL_Rect src_rect;
+	src_rect.x = 0;
+	src_rect.y = static_cast<Sint16>(nFrame * art->h());
```
glebm, line 16: No need for `static_cast` if not doing aggregate initialization (`{...}`), an implicit cast happens on assignment.

SourceX/DiabloUI/dialogs.cpp
```diff
@@ -215,12 +216,16 @@ void Deinit()
 		UnloadTtfFont();
 	if (textInputWasActive)
 		SDL_StartTextInput();
-	for (std::size_t i = 0; i < dialogItemsSize; ++i) {
-		dialogItems[i].FreeCache();
+
+	for (std::size_t i = 0; i < vecOkDialog.size(); i++) {
+		UiItemBase *pUIItem = vecOkDialog[i];
+		if (pUIItem)
```
glebm, line 222: `delete NULL;` is safe

Generally, when checking for nullptr it's better to be explicit (`if (pUIItem != NULL)`) because implicit boolean conversion operator can be overriden.

SourceX/DiabloUI/ui_item.h
```diff
@@ -38,316 +38,321 @@ enum UiFlags {
...
+		return m_vecItems[i];
 	}
+
+	//private:
+	Sint16 m_x, m_y;
+	Uint16 m_width, m_height;
+	std::vector<UiListItem *> m_vecItems;
 };
 
+typedef std::vector<UiItemBase *> vUiItemBase;
```
glebm, line 356: Typedefing `vector` makes the code harder to read. It also makes it unobvious when something expensive happens, such as an entire vector is being copied (e.g. passed-by-value). Or when `delete` isn't done on each element.

SourceX/DiabloUI/ui_item.h
```diff
@@ -38,316 +38,321 @@ enum UiFlags {
...
+		m_iFlags |= flag;
 	}
 
 	void remove_flag(UiFlags flag)
 	{
-		flags &= ~flag;
+		m_iFlags &= ~flag;
+	}
+
+	virtual void FreeCache()
```
glebm, line 83: Why this is on the base class now?
Do we want to call it in the destructor?

SourceX/DiabloUI/diabloui.cpp
AJenbo: @marty28 I'm going to revert this, let me know if it's needed for VS2003

AJenbo: @glebm at least `Render(UiText/UiList` have issues with using `const` even on GCC:
```
SourceX/DiabloUI/diabloui.cpp:698:38: error: passing ‘const dvl::UiList’ as ‘this’ argument discards qualifiers [-fpermissive]
  698 |   SDL_Rect rect = ui_list->itemRect(i);
```

SourceX/DiabloUI/credits.cpp
AJenbo: It doesn't have `auto`, so my guess is that this was just the quick way to solve it. I changed it to `CachedLine`

SourceX/DiabloUI/credits.cpp
AJenbo: I'm not quite sure how to do that, at least not at this time.

I have applied your other suggestions as best able.

SourceX/DiabloUI/ui_item.h
```diff
@@ -38,316 +38,314 @@ enum UiFlags {
...
+		m_x = x;
+		m_y = y;
+		m_width = item_width;
+		m_height = item_height;
+	};
 
-	bool has_any_flag(int flags) const
+	~UiList() {};
+
+	SDL_Rect itemRect(int i)
```
glebm, line 322: This function should be `const`

SourceX/DiabloUI/diabloui.cpp
glebm: The `itemRect` function should be marked `const`

SourceX/DiabloUI/credits.cpp
AJenbo: Got some help from @Lundar in solving this :)

Tests look good so I have now merged this.

PR #774, opened 2020-08-12
Adding vim .swp files to gitignore

sixcy:
Should I conclude I'm the only vim user out here? ;)


qndel:
probably ;)

AJenbo:
I feel like people should just add thins to there global ignore, my editor isn't in the ignore list for this reason :P

Issue #661, opened 2020-03-16
Multiplayer difficulty selection with keyboard arrows


MysteryFTG:
This is just a very minor issue, not really a bug, but still something that's different from Diablo 1.09.

When starting a new multiplayer game the normal difficulty is selected by default. In 1.09 I can press "up" on the keyboard to wrap down to hell difficulty. This is also how the main menu behaves.

The main menu works correctly in devilutionX, however the multiplayer difficulty selection doesn't wrap around and I have to press down twice to get from normal to hell.

yuripourre:
This is the first time I dive deep into the menu code, so I may be wrong.
My guess is that is something related to this function: https://github.com/diasurgical/devilutionX/blob/master/Source/gmenu.cpp#L133

AJenbo:
@yuripourre that would be the ingame menu. For DiabloUI you define if the menu should wrap or not via the last parameter for `UiInitList` https://github.com/diasurgical/devilutionX/blob/master/SourceX/DiabloUI/selgame.cpp#L152

yuripourre:
Thanks @AJenbo. I am trying to understand the code.

PR #777, opened 2020-08-16
Modified UiInitList declaration to allow selection wrapping

JeefJoof:
This is my first pull request so I apologize if anything is done incorrectly.

This will fix Issue #661 regarding the lack of selection wrapping when using the arrow keys to select the difficulty for multiplayer sessions. Rather than allow wrapping only when choosing difficulty, I modified the default parameter in the declaration for UiInitList so the behavior would be more consistent across applicable menus. However, this does not work in the 'Select Hero' menu because its behavior is handled separately.

Thank you.

### DIFF
--- a/SourceX/DiabloUI/diabloui.h
+++ b/SourceX/DiabloUI/diabloui.h
@@ -54,7 +54,7 @@ void SetMenu(int MenuId);
 void UiFocusNavigationSelect();
 void UiFocusNavigationEsc();
 void UiFocusNavigationYesNo();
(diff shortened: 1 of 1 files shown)


AJenbo:
I think that rather then defaulting to true this should be done by explicitly parsing true for the affected menu. Else it will affet more menus then it should

AJenbo:
This has now been fixed via this commit: https://github.com/diasurgical/devilutionX/commit/6a6a12da385daaa62cf56f8ab1cdc0ca03b14439

PR #783, opened 2020-09-01
Softcode VSYNC setting. Enabled by default.

### LINKED ISSUE #782: VSYNC option should be softcoded
One of the very first things I noticed when running DevilutionX was the laggy mouse cursor on the main menu. I immediately realized vsync was turned on and proceeded to check in the code where we were setting that.

I found the setting, but it is currently hardcoded inside `display.cpp`.

I propose we softcode that setting, allowing configuration from the `diablo.ini` file.

julealgon:
This PR softcodes the 'VSYNC' option when creating the SDL renderer, allowing one to change the vsync setting through standard `diablo.ini` configuration.

The new setting is called `vsync` and can have values `0` (disabled) or `1` (enabled). As with other settings, if the flag is not found, it will be defaulted (to `1` in this case to retain old behavior) and saved back to configuration.

This resolves #782 .

### DIFF
--- a/SourceX/display.cpp
+++ b/SourceX/display.cpp
@@ -116,7 +116,15 @@ bool SpawnWindow(const char *lpWindowName, int nWidth, int nHeight)
 #ifdef USE_SDL1
 		SDL_Log("upscaling not supported with USE_SDL1");
 #else
-		renderer = SDL_CreateRenderer(ghMainWnd, -1, SDL_RENDERER_PRESENTVSYNC | SDL_RENDERER_ACCELERATED);
+		Uint32 rendererFlags = SDL_RENDERER_ACCELERATED;
+
+		int vsyncEnabled = 1;
+		DvlIntSetting("vsync", &vsyncEnabled);
+		if (vsyncEnabled) {
+			rendererFlags |= SDL_RENDERER_PRESENTVSYNC;
+		}
+
+		renderer = SDL_CreateRenderer(ghMainWnd, -1, rendererFlags);
 		if (renderer == NULL) {
 			ErrSdl();
 		}
--- a/SourceX/display.cpp
+++ b/SourceX/display.cpp
@@ -118,7 +118,7 @@ bool SpawnWindow(const char *lpWindowName, int nWidth, int nHeight)
 #else
 		Uint32 rendererFlags = SDL_RENDERER_ACCELERATED;
 
-		int vsyncEnabled = 1;
+		vsyncEnabled = 1;
 		DvlIntSetting("vsync", &vsyncEnabled);
 		if (vsyncEnabled) {
 			rendererFlags |= SDL_RENDERER_PRESENTVSYNC;
--- a/SourceX/display.h
+++ b/SourceX/display.h
@@ -8,6 +8,7 @@
 
 namespace dvl {
 
+extern int vsyncEnabled;
 extern int refreshDelay; // Screen refresh rate in nanoseconds
 extern SDL_Window *window;
 extern SDL_Renderer *renderer;
--- a/SourceX/dx.cpp
+++ b/SourceX/dx.cpp
@@ -17,6 +17,7 @@ int locktbl[256];
 #endif
 static CCritSect sgMemCrit;
 
+int vsyncEnabled;
 int refreshDelay;
 SDL_Renderer *renderer;
 SDL_Texture *texture;
@@ -280,6 +281,10 @@ void RenderPresent()
 			ErrSdl();
 		}
 		SDL_RenderPresent(renderer);
+
+		if (!vsyncEnabled) {
+			LimitFrameRate();
+		}
 	} else {
 		if (SDL_UpdateWindowSurface(ghMainWnd) <= -1) {
 			ErrSdl();


julealgon:
Does anyone know why testflow is failing for windows and switch? I'm not sure how to access CircleCI there and can't see any logs. Any help is appreciated.

AJenbo:
Nice. Will this cause the CPU to run at 100%? Do you think this is acceptable/expected behavior?

Windows build is failed because of an SSL error when trying to download the dependencies from download.libsodium.org.

Switch has been giving an odd docker error lately, of cause it hit on the same day as we solved the cmake build issue for the same platform :/ I haven't had time to investigate it yet.

You should be able to gain access to CircleCI by logging in with your GitHub account.

julealgon:
@AJenbo 
> Will this cause the CPU to run at 100%?

Yeah it does lock an entire thread to 100%:

image

> Do you think this is acceptable/expected behavior?

I think it's mostly fine for a simple game like this. It's not ideal for sure, but expected if there is no special handling (a frame limiter in the game).

Considering it is disabled by default and that results are similar to other games, I think this should be ok.

We could look into implementing a frame limiter later on as a separate effort IMHO.

AJenbo:
We do have an implementation of a frame limiter that is used when there isn't hardware support for vsync. It's not high precision, but unless your monitor runs at 20hz it should still give you better results than the original game.
https://github.com/diasurgical/devilutionX/blob/f66339a9ac89cbeae02476e583390dd79e8515c1/SourceX/dx.cpp#L242

julealgon:
> You should be able to gain access to CircleCI by logging in with your GitHub account.

Nice! I have access now, thanks.

> Windows build is failed because of an SSL error when trying to download the dependencies from download.libsodium.org.

So I know this is completely off-topic here, but why are we firing those downloads manually like that? Can't we use `vcpkg` to fetch all those dependencies? I apologize in advance for my ignorance of C++ development, but I would assume there would be a way to configure the dependencies to download automatically on build as needed, kinda like .Net NuGet packages? This would streamline development itself as well as no manual steps would be needed to install `vcpkg` and all that.

Alternatively, do you know why this one download in particular is failing the SSL check? Can't we just use `http` instead of `https` for that particular one (not sure who can edit the pipeline steps there, sorry if obvious)?

julealgon:
Interesting. I assumed there was no framelimiter, didn't know it was just not enabled in this flow. Sorry, still very new to this codebase, I barely checked a handful of files 😆 

Forcing it to execute seems to get CPU down to ~8% (which is actually surprisingly high still to me). 
image

I could check if I can enable it conditionally on the vsync setting. Do you think it's worth pursuing as part of this PR?

julealgon:
Interestingly enough, while testing the second commit here, I noticed that CPU utilization is actually really high when VSYNC is enabled:
image

The second commit should ensure we run the frame limiter when vsync is disabled to limit CPU usage down to ~8% (an overall improvement surprisingly).

The use of global variables is a bit scary to an enterprise developer like myself but I figured this was the way to go considering we are calling `RenderPresent` from several places right now... any attempt of parameter passing would create a huge cascade of changes 😨

AJenbo:
> Can't we use `vcpkg` to fetch all those dependencies?
This is a cross-compiled build, for Windows, done on a Linux machine. We use vcpkg for the Windows-based builds. We test this pipeline as that is how I do the releases as well, as I don't have easy access to a Windows machine, but here I don't have to download it each time. It would be good if we could adjust the build so that it caches the dependencies between builds, but it hasn't been high enough on the priority list since it works™ (for the most part).

> no manual steps would be needed to install `vcpkg` and all that.
vcpkg doesn't come with Windows and a lot of users have said they struggle to install it and get it working. For the native Windows builds we use a pre-existing image from AppVeyor where it has already been installed, but I don't know if there is a similar Linux image.

> Can't we just use `http` instead of `https` for that particular one (not sure who can edit the pipeline steps there, sorry if obvious)?
Not much of the internet is accessible over plain http at this point :) Any one can propose changes to the build pipeline by editing the respective config file. That way if you make a PR that requires changes to the build process you can have the changes included in the PR so that the build passes. I'm actually working on solving the issue right now:
https://github.com/diasurgical/devilutionX/pull/787

AJenbo:
> The second commit should ensure we run the frame limiter when vsync is disabled to limit CPU usage down to ~8% (an overall improvement surprisingly).

Nice work :) That should make the PR ready for merge, I'll give it another review tomorrow.

> The use of global variables is a bit scary to an enterprise developer like myself but I figured this was the way to go considering we are calling `RenderPresent` from several places right now... any attempt of parameter passing would create a huge cascade of changes fearful

I second that notion. The original code made heavy use of globals and we want to be able to merge improvements from upstream (like Hellfire support) so at least for now, we will have to close our eyes and accept the chaos :)

AJenbo:
> Forcing it to execute seems to get CPU down to ~8% (which is actually surprisingly high still to me).

If this is a debug build that might explain it as it causes a much higher CPU load. The rendering is more ore less exclusively done on the CPU so that's the reason for the high load, additionally to the original, it also has to do an 8bit -> 32bit conversion and upload to the GPU each frame as upscaling is done on the GPU in 32bit mode.

AJenbo:
Looks good, let's do this

Issue #784, opened 2020-09-01
Support proper integer scaling


julealgon:
The second thing I noticed after running DevilutionX (the first being the laggy cursor #783) was that the image was extremely blurry. I tried playing around with the settings but it seems that the game is not scaling using the most adequate algorithm for pixel art, resulting in a somewhat blurry image even when the screen resolution is a multiple of 640x480.

We should try to add an integer scaling option with pixel multiplication (no bilinear filtering or anything like that) and make sure the game scales to integer multiples of the standard resolution based on target resolution.

This is how the launcher from GOG works currently and the difference is night and day.

| GOG integer scale x 2 (1280x960) | DevilutionX (1280x960) |
| --- | --- |
| ![image](https://user-images.githubusercontent.com/461579/91782672-21bf4a00-ebd4-11ea-8d74-672409d0b3d5.png) | ![image](https://user-images.githubusercontent.com/461579/91782504-a5c50200-ebd3-11ea-832c-e814bf6710db.png) |

I actually had to use windows capture instead of in-game screenshot for DevilutionX. For some reason, when doing an in-game print, the final PCX file appears correctly. However, when looking at the game, the picture above is what I see.

I don't know yet why this is happening. If anybody has any idea just let me know.

julealgon:
Found it. This seems to be a bug in SDL when the native Windows DPI scaling is not set to the 100% default value.

I get the following results if I set my main monitor back to 100% scale:

![image](https://user-images.githubusercontent.com/461579/91783375-6f3cb680-ebd6-11ea-937c-42b260362fcf.png)

Found this post where folks are discussing it:
https://stackoverflow.com/a/41746429

I'll investigate further to see if it is possible to make the game render properly when windows scaling is used.

julealgon:
Ok I just realized that DPI-awareness is a separate issue entirely and that integer scaling is still required.

These are the results with and without integer scaling on a 1080p display:

| integer x 2 (1080p -> 960p) | non-integer scale |
| --- | --- |
| ![image](https://user-images.githubusercontent.com/461579/91783889-d60e9f80-ebd7-11ea-9055-a760f45cddde.png) | ![image](https://user-images.githubusercontent.com/461579/91783895-d9a22680-ebd7-11ea-9352-5fa09e635074.png) |

AJenbo:
Diablo technically isn't pixel art but prerendered 3D, some prefer blurry scaling others integer scaling so really good that you are implementing this as options.

[...]

PR #785, opened 2020-09-01
Add integer scaling support. Disabled by default.

julealgon:
This PR introduces the option to enable integer scaling in DevilutionX. It can be enabled by setting a new `diablo.ini` flag called `integer scaling`, which can have both a `0` (disabled) and `1` (enabled) values. The default is `0` to maintain current behavior.

This results in optimal scaling of 2D assets by ensuring that every pixel is stable during scaling: each pixel will always end up being N times bigger in the scaled image, no matter its position. i.e. a single pixel in the original 640x480 frame becomes 4 pixels (2x2) on 1280x960 or 9 pixels (3x3) on 1440p.

On a normal non-integer scaling, each column and row on the screen will have to fit a wider area that is not a perfect multiple of the original image. This introduces rounding differences resulting in variable pixel widths and/or heights, which in turn dramatically degrade picture quality by adding constant shimmering, especially with scrolling images.

Here is a real example (windows screenshot over game frame, not in-game screenshot):

| integer x 2 (1080p -> 960p) | non-integer scale (fullscreen 1080p) |
| --- | --- |
| image | image |

Notice how, while both images are clearly aliased (due to low resolution), the integer scaled version has all segments exactly identical, while the other image is all over the place: some 2 pixel rows, some 1 pixel rows, etc.

Enabling integer scaling manually ensures SDL will scale the image to the maximum possible multiple of 480p over your desktop resolution. This removes the need to configure the desktop resolution to an integer multiple before opening the game.

This implements #784

### DIFF
--- a/SourceX/display.cpp
+++ b/SourceX/display.cpp
@@ -126,6 +126,12 @@ bool SpawnWindow(const char *lpWindowName, int nWidth, int nHeight)
 			ErrSdl();
 		}
 
+		int integerScalingEnabled = 0;
+		DvlIntSetting("integer scaling", &integerScalingEnabled);
+		if (integerScalingEnabled && SDL_RenderSetIntegerScale(renderer, SDL_TRUE) < 0) {
+			ErrSdl();
+		}
+
 		if (SDL_RenderSetLogicalSize(renderer, nWidth, nHeight) <= -1) {
 			ErrSdl();
 		}


AJenbo:
Nice work, I had no idea that this option existed in SDL, I thought we would have to do a custom implementation.

julealgon:
> I had no idea that this option existed in SDL, I thought we would have to do a custom implementation.

Same to be honest. I had no idea it would be this easy 😆 I actually thought the method wasn't working properly at first, but it was all because of the DPI scaling issue. Once I set my desktop scaling back to 100% then it worked flawlessly.

PR #790, opened 2020-09-03
Remove attack/cast animation skipping

### LINKED ISSUE #779: Remove animation skipping on player attacks
When attacking, the game allows the player to override the attack animation right after the "hit frame" with a new attack. This new attack then abruptly interrupts the previous attack animation and starts a new one. This makes the attack animation look bad as all the frames after the hit frame are lost.

I'm not entirely sure if this was done on purpose by Blizzard (I assume it was, since it is quite obvious), but this doesn't seem to exist for monster attacks or spells, only for player attacks.

Additionally, it seems like frames after the hit-frame are not affected by attack speed modifiers, which also seems inconsistent and discourages kiting. This easy to test with a `fastest attack speed` weapon by performing a single attack: the initial frames of animation are much faster, then we get the hit frame somewhere in the middle of the animation, and then all the frames after that (could be called attack recovery frames) are the same speed as in a normal attack.

It would be nice to have this quirk removed from the game so that we have a smoother, more consistent play experience.

Optionally, one could increase the overall speed of the whole animation so as to maintain the same attacks per second as the original version.

I'd like to help with this, if at all possible.

EDIT:

Here are a few animations to showcase what I'm proposing:

| Vanilla (animation skipping) | Smooth (full animation) |
| --- | --- |
| warrior-skip | warrior-smooth |
| sorcerer-skip | sorcerer-smooth |

julealgon:
This PR removes special handling for animation skipping on attacks (normal, ranged, spell). Effectively, this results in slower overall attack speeds but ensures all attack motions always play properly (running all their frames of animation).

After this change, attack commands work similarly to any other command in the game that result in a specific animation.

I'm considering the updated version "correct behavior" and thus no flags of any kind are introduced to turn the functionality on and off.

I am not touching on animation speed in this MR, so standard and fast (quick, fast, faster, fastest) logic remains the same, with the exception that attack "recovery" frames (frames that happen after the hit-frame) always play. This will result in a proportionally less significant upgrade for fast weapons. That's because attack speed modifiers remove a constant amount of frames no matter the number of total frames of animation.

For example:

- F/H - Frames per hit
- IAS - Increased attack speed %

| | Warrior 1H attack (frames) | Hit Frame | F/H before / IAS | F/H after / IAS |
| --- | --- | --- | --- | --- |
| Standard | 16 | 9 | 10 / 0% | 16 / 0% |
| Quick | 15 | 8 | 9 / 10% | 15 / 6.25% |
| Fast | 14 | 7 | 8 / 20% | 14 / 12.5% |

This resolves #779. Check the issue for comparison gifs.


julealgon:
Converted to draft as per discussion on issue. Will be kept here for future reference.

AJenbo:
Closing this as I consider it fixed by https://github.com/diasurgical/devilutionX/commit/c7b9ffec1f1ddd95d92fc1bc0bac8258bf7ebfe4

julealgon:
> Closing this as I consider it fixed by c7b9ffe

You might be confusing this with the 1st frame skipped issue @AJenbo . This one was about the purposeful animation skipping that was implemented to speed up sequential attacks in the game. It is not related to animation and logic running together or anything like that.

Not a big deal for it to be closed however, as it would be a significant balance change.

Just wanted to clarify that they are not related.

AJenbo:
Yeah, I understand that they are not 1:1, but the relevant part for me is where they overlap: Play all frames in the animation. That should afaik be solved by c7b9ffe afaik.

julealgon:
> but the relevant part for me is where they overlap: Play all frames in the animation.

This one is unrelated to playing all frames in the animation 😛

AJenbo:
Ok, well lets just say then that I'm closing some of the PRs that won't get merged so that we can lower the number of open PRs :D
But feel free to create a branch and add it to the mod page of the wiki :)

PR #797, opened 2020-09-09
simplified theme room creation

pionere:
unnecessary checks eliminated, frequently used values are stored in (constant) locals.

### DIFF
--- a/Source/gendung.cpp
+++ b/Source/gendung.cpp
@@ -380,61 +380,35 @@ BOOL DRLG_WillThemeRoomFit(int floor, int x, int y, int minSize, int maxSize, in
 void DRLG_CreateThemeRoom(int themeIndex)
 {
 	int xx, yy;
+	const int lx = themeLoc[themeIndex].x;
+	const int ly = themeLoc[themeIndex].y;
+	const int hx = lx + themeLoc[themeIndex].width;
+	const int hy = ly + themeLoc[themeIndex].height;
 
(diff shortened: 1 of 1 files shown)


AJenbo:
Much cleaner, have you tried verifying if it's bin exact to the original?

If it's not then I would prefer if tests were added to check that the function behaves the same as before.

qndel:
I'll verify the bin exactness for you

AJenbo:
Most of your changes are trivial, but this part `(yy == ly || yy == hy - 1)` is what I'm most unsure of, besides using a var to hold the value you also removed some conditions, so if you can show that they are the same for one of these then I think it's acceptable.

Or you can do the simple 1:1 var replacment for now and we can look at simplifying the `if` at a later point.

pionere:
if you do a simple var replacement, you get a condition like:
```
for (yy = ly; yy < hy; yy++) {
    for (xx = lx; xx < hx; xx++) {
        if (yy == ly && xx >= lx && xx <= hx ||
            yy == hy - 1 && xx >= lx && xx <= hx) { ... }
```
and from the loop expression it is obvious that the second and third checks are not necessary (always TRUE), thus:
```
for (yy = ly; yy < hy; yy++) {
    for (xx = lx; xx < hx; xx++) {
        if (yy == ly || yy == hy - 1) { ... }
```

AJenbo:
Looks like your right, but definitely not the way they did it originally, it's kinda interesting that the compiler didn't figure out to optimize this aways.

PR #802, opened 2020-09-11
correct 64-bit spell-check in one place

pionere:
Spell availability check is done in CalcPlrScrolls, which makes the ones after item-removal unnecessary.
The check in CalcPlrScrolls is fixed to work with 64bit values.


AJenbo:
Thanks for reporting this and updating the PR now that the code was changed by another PR

PR #806, opened 2020-09-12
Expose monster type on monster info

### LINKED ISSUE #805: Show monster type: undead / animal / demon
All monsters in the game are categorized among the 3 monster types:
- Undead
- Animal
- Demon

Various game mechanics are dependent on monster type. For example, swords deal 50% more damage to animal monsters, but 50% less damage against undead. Conversely, blunt weapons deal 50% extra damage to undead monsters, but 50% less damage to animals.

Same goes for spells. Holy Bolt only damages undead monsters, while going directly through any other monster type.

Sometimes it is hard to tell precisely what the monster type is for a given monster. Thus, I propose we expose this information in the monster data displayed at the bottom of the screen.

| | Current | Proposal |
| --- | --- | --- |
| Normal | image | image |
| Unique | image  | image|

_*PS*: Screenshots with #803 applied_

I purposefully reduced "TOTAL KILLS" to just "KILLS" to try and use the space a little better in the second line. I thought leaving the original text resulted in too big of a line and it looked strange to me.

Additionally, it felt natural to me because of how items are displayed: damage and durability on the same line. So I think it doesn't feel out of place. 

Lastly, I proposed using the "KILLS" like to put the monster type because normal monsters already use up all 5 lines for some information:

| Line | Text |
| --- | --- |
| Line1 | {Name} |
| Line2 | TOTAL KILLS: {n} |
| Line3 (> 30 kills) | HIT POINTS: {n0 - n1} |
| Line4 (> 15 kills) | RESISTS: {magic} {fire} {lightning} |
| Line5 (> 15 kills) | IMMUNE: {magic} {fire} {lightning} |

If we add the enemy type as a brand new line, then we get into situations where the full info has 6 lines, which cannot be displayed in the panel (maximum of 5 lines).

The total number of lines for unique monsters is less, because it doesn't have the total kill line and the hit points line.

julealgon:
This PR introduces one extra information when displaying monster details in the bottom info panel: the monster type.

| | Disabled | Enabled |
| --- | --- | --- |
| Normal (on-screen bar) | image | image |
| Normal (info panel) | image | image | image |
| Unique (on-screen bar) | image | image |
| Unique | image  | image|

The information is added to the second info line to preserve space and for line limit considerations.

A new feature toggle was added to control this behavior, `Show monster type information`, disabled by default.

This resolves #805

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -437,6 +437,7 @@ static void SaveOptions()
 	setIniInt("Game", "Auto Equip Shields", sgOptions.Gameplay.bAutoEquipShields);
 	setIniInt("Game", "Auto Equip Jewelry", sgOptions.Gameplay.bAutoEquipJewelry);
 	setIniInt("Game", "Enable All Quests for Single Player", sgOptions.Gameplay.bAllQuests);
+	setIniInt("Game", "Show Monster Type", sgOptions.Gameplay.bShowMonsterType);
 
 	setIniValue("Network", "Bind Address", sgOptions.Network.szBindAddress);
 }
@@ -491,6 +492,7 @@ static void LoadOptions()
 	sgOptions.Gameplay.bAutoEquipShields = getIniBool("Game", "Auto Equip Shields", false);
 	sgOptions.Gameplay.bAutoEquipJewelry = getIniBool("Game", "Auto Equip Jewelry", false);
 	sgOptions.Gameplay.bAllQuests = getIniBool("Game", "Enable All Quests for Single Player", false);
+	sgOptions.Gameplay.bShowMonsterType = getIniBool("Game", "Show Monster Type", false);
 
 	getIniValue("Network", "Bind Address", sgOptions.Network.szBindAddress, sizeof(sgOptions.Network.szBindAddress), "0.0.0.0");
 }
--- a/Source/diablo.h
+++ b/Source/diablo.h
@@ -98,6 +98,8 @@ typedef struct GameplayOptions {
 	bool bAutoEquipJewelry;
 	/** @brief Enable All Quests for Single Player */
 	bool bAllQuests;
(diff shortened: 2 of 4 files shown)


julealgon:
Retry DeepCode

AJenbo:
Been disappointed about DeepCode, it's detected very few issues and has errored a lot, it's removed now.

AJenbo:
How about having this behind an ini flag?

julealgon:
> How about having this behind an ini flag?

We can definitely take a look at that. I'll rebase this later and add the flag.

Source/diablo.cpp
```diff
@@ -437,6 +437,7 @@ static void SaveOptions()
 	setIniInt("Game", "Auto Equip Shields", sgOptions.Gameplay.bAutoEquipShields);
 	setIniInt("Game", "Auto Equip Jewelry", sgOptions.Gameplay.bAutoEquipJewelry);
 	setIniInt("Game", "Enable All Quests for Single Player", sgOptions.Gameplay.bAllQuests);
+	setIniInt("Game", "Show monster type information", sgOptions.Gameplay.bShowMonsterType);
```
glebm, line 440: Is there other information other than the type itself?
If not, consider renaming to "Show monster type"

julealgon:
Rebased and added toggle for new behavior, disabled by default.

I also made the toggle control the same behavior on the on-screen health bar, if it is enabled.

| Toggle Off | Toggle On |
| --- | --- |
| image | image |
| image | image |

Source/diablo.cpp
AJenbo: Call it "Show Monster Type" so the casing is also consistent.

julealgon: Renamed. And fixed casing to match other setting names.

PR #813, opened 2020-09-12
align inventory cell backgrounds with sprite

Lundar:
Change Inventory cell table so that cells line up consistently. Previously Cells would drift out of X alignment from left to right, with right most cells being 2-3 pixels out. Y alignment on the 4th inventory row was also 1 pixel out of alignment, resulting in the dividing line above the 4th row being red when occupied instead of while like the other rows.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -288,7 +288,7 @@ void DrawInv()
 		}
 
 		if (plr[myplr].InvBody[INVLOC_HAND_LEFT]._iLoc == ILOC_TWOHAND) {
-			InvDrawSlotBack(RIGHT_PANEL_X + 247, 160 + SCREEN_Y, 2 * INV_SLOT_SIZE_PX, 3 * INV_SLOT_SIZE_PX);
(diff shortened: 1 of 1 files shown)


Lundar:
screen00
screen01

Lundar:
Before:
screen02

After:
screen00

AJenbo:
Thanks for the before and after, the issue (and fix) is much clear now. I'll test this with a game controller before merging just to be fure, but other then that this looks good to merge.

mewmew:
Great PR @Lundar. I also appreciate the before and after screenshots.

Cheers,
Robin

PrisonOfMirrors:
Imho the better way to handle this would be like they did in Diablo 2: specify the top left corner of the box and then the number of rows/columns assigned to that rectangle and let the game dynamically construct the coordinates. This keeps fro. Having to hardcode every single box and ensures everything is proper. Really don't understand the design decisions of the developers short of needing to release the game by a deadline

AJenbo:
But that would require math :o

PrisonOfMirrors:
!(image)[https://i.pinimg.com/originals/24/c0/23/24c0237d9f330e31e11b7d4c781fabcb.png]

PrisonOfMirrors:
image

PR #819, opened 2020-09-17
fixed memory corruption when using an item

pionere:
prevent the possible negative-indexing of plr[pnum].InvList

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -2107,7 +2107,7 @@ BOOL UseInvItem(int pnum, int cii)
 		return TRUE;
 	if (stextflag != STORE_NONE)
 		return TRUE;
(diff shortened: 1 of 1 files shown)


AJenbo:
Looks like you managed to tame Git :)

AJenbo:
Nice find, was hard to read even with a diff :D

galaxyhaxz:
A little background about this bug. The check is doing `<= 5` because originally there were only 5 items. They added amulets very late in development and forgot to update that to 6, and didn't have an enum in place so the raw value remained.

Slot 6 is the armor slot, and because it falls right under the inventory items it will try to use the armor when right clicking and not cause any issues. However, if you move say the belt underneath the inventory, then right clicking the armor will use potion slot 8.

PR #820, opened 2020-09-17
correct stat checks when calculating the item bonuses

pionere:
check the stats of the given player, since it might not be the same as the mc


AJenbo:
Thanks

PR #823, opened 2020-09-17
use defs instead of constants

pionere:
some constants are replaced with the appropriate defines.


AJenbo:
Most of this (since it doesn't affect the compiled result) should be applied to devilution instead of devilutionx, it will then land in devilutionx on the next sync.

AJenbo:
This also needs a rebase at this point, sorry :/

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/486be4efebbf8df664db670844456b7df6165299, but looks like most of it was already done via Devilution.

PR #824, opened 2020-09-17
prevent possible long/infinite loops by not relying on random

pionere:
use better loops to find matching shrine/stat/item.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -354,15 +354,13 @@ void InitRndBarrels()
 			t = 0;
 			found = FALSE;
 			while (TRUE) {
-				if (t >= 3)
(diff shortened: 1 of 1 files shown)


AJenbo:
please provide unit test with these cleanups so that we can easily verify the behavior.

pionere:
how exactly do you expect to unit test a code which depends on random()?

yuripourre:
@pionere if you need deterministic tests, you can provide a seed to random

qndel:
be aware that by modifying/changing the sequence of random calls you might affect everything else, I think we managed to keep things the same so far = if you played devilutionx and vanilla side by side I'd expect same things to happen

PR #828, opened 2020-09-21
pass to strncpy size of buffer instead of size of pointer

pal666:
d749e5e6f268748ebd6f5acf72420840d2c11339 broken hero names longer than 7 characters

### DIFF
--- a/SourceX/DiabloUI/selhero.cpp
+++ b/SourceX/DiabloUI/selhero.cpp
@@ -496,7 +496,7 @@ BOOL UiSelHeroDialog(
 	} while (selhero_navigateYesNo);
 
 	*dlgresult = selhero_result;
-	strncpy(name, selhero_heroInfo.name, sizeof(name) - 1);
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
@pal666 Can you solve the merge conflict?

AJenbo:
This was fixed in https://github.com/diasurgical/devilutionX/commit/9fd34ec9276b91f36895c0fae638a2d8267a4f2d + https://github.com/diasurgical/devilutionX/commit/25e0e2079794227577769f8d0a851183de765719

This PR is invalid as it allowed for names of 31 char, but it should only be 15.

pal666:
in my checkout it's declared `char gszHero[16]`, which makes its sizeof-1 15 bytes, but as long as it's fixed, i'm fine

AJenbo:
Sorry I received another fix for it using `PLR_NAME_LEN` and got them confused.

PR #830, opened 2020-09-23
minor code cleanups

pionere:
eliminate useless random calls
use do-while instead of a while loop
eliminate unused local variable


qndel:
the "useless random calls" look like some define for wall generation chance, wondering what'd happen if we set it to 50 :D

pionere:
interesting. Care to elaborate on that?
I think the random to be meaningful at least there should be a SetRndSeed beforehand.

PR #831, opened 2020-09-23
mWhoHit and xp gain fixes

pionere:
set the mWhoHit flag of the correct monster (in M2MStartHit).
add xp if the monster id is MAX_PLRS. (in MonstStartKill)
sync M2MStartKill with MonstStartKill a bit


AJenbo:
Ouch, looks like there was a line change on this file. Any chance you could redo it?

AJenbo:
// BUGFIX: what about Q_GARBUD like in MonstStartKill

Hadent noticed this issue, but it turns out that was actually fixed in Hellfire :)

AJenbo:
Remaning changes merged via https://github.com/diasurgical/devilutionX/commit/5e84fd715b94a96a0a68340c23604db1c57dff9f

PR #832, opened 2020-09-23
correct fallenfear direction

pionere:
fallenfear direction fixed to use the position of the correct monster
+ some minor code cleanup


AJenbo:
Was a little hard to see what the issue was since the clean-up is done in the same commit. But yeah they were basically picking the opposite direction on a random monster and not them selfs.

PR #833, opened 2020-09-23
prevent quest items to be sold at the smith

pionere:


AJenbo:
Could you rebase this fix now that HF has been merged :)

qndel:
what quest items could be sold to griswold? ; o

pionere:
> what quest items could be sold to griswold? ; o

and here I thought devx aim is to be a modding base...

qndel:
1. this doesn't even compile, you are comparing ItemStruct with an int
2. this wouldn't let you sell items like leoric's crown / butcher's cleaver etc.

PR #835, opened 2020-09-23
use pnum instead of myplr in bone chamber

pionere:


Source/objects.cpp
```diff
@@ -2908,26 +2908,26 @@ void OperateBook(int pnum, int i)
...
+		    plr[pnum]._pdir,
 		    MIS_GUARDIAN,
 		    TARGET_MONSTERS,
-		    myplr,
+		    pnum,
 		    0,
 		    0);
 	}
-	if (setlevel && setlvlnum == SL_VILEBETRAYER) {
+	if (setlvlnum == SL_VILEBETRAYER) {
```
AJenbo, line 2930: Why is the check for `setlevel` removed from this quest?

pionere: because of line 2907:
```
if (!setlevel)
		return;
```

PR #840, opened 2020-09-29
Add missing 1st frame for animations

### LINKED ISSUE #838: Missing 1st animation frame on all player/monster/item animations
Due to the way the game handles animation instantiation and increments, all animations in the game end up starting from frame # 2 instead of starting from the beginning (frame # 1). 

This impacts: 
- All Attacks: melee, ranged and spells
- All item flips
- All monster animations

Animations that loop are also affected, however only on the first loop. The looping logic ends up "correcting" the behavior automatically after the first loop ends.

Here is an example of the difference we are talking about. The frame on the left is frame 2 of the warrior's attack animation, which is what plays today when the attack starts. The frame on the right is frame 1, which is never displayed in-game today. Note that it is _not_ equal to one of the standing frames (the shield is slightly raised):

image

This problem happens because all animations are created and initialized with frame 1 in place, but before the first frame is ever rendered, the update loops execute and increment this value to 2 (`ProcessPlayers`, `ProcessItems`, `ProcessMonsters`).

### LINKED ISSUE #839: 'quick' and 'haste' speed modifiers do not work as intended
Because of the way the game handles frame skipping to increase attack speeds, some modifiers in the game end up having wrong effects on the final attack speeed.

For reference, here is the logic for each attack type that is affected by speed modifiers:
| Function | Logic |
| --- | --- |
| PM_DoAttack | https://github.com/diasurgical/devilutionX/blob/4a4486d6161028e69702bedec6cb071e6a5354c3/Source/player.cpp#L2617-L2629 |
| PM_DoRangedAttack | https://github.com/diasurgical/devilutionX/blob/4a4486d6161028e69702bedec6cb071e6a5354c3/Source/player.cpp#L2703-L2709 |

Notice how all frame skipping for both attacks always check frame 1 to decide how many frames to skip. However, these 2 methods are most of the time not fired for frame 1, and instead only begin to execute on frame 2 onwards. This has the effect that all frame skipping logic that is based on frame 1 is actually discarded. Effectively:

| Suffix | Modifier | Intended frame skips | Effective frame skips |
| --- | --- | --- | --- |
| "of readiness" | `quick attack` | 1 frame | 0 frames |
| "of swiftness" | `fast attack` | 2 frames | 1 frame |
| "of speed" | `faster attack` | 3 frames | 2 frames |
| "of haste" | `fastest attack` | 4 frames | 2 frames |

This behavior is caused by the way the game handles animation increments (see #838) as well as how attacks work when starting from a standstill. When performing player updates, animation transitions from several player modes cause the logic to repeat during the same frame depending on the return value of each processing function:
 https://github.com/diasurgical/devilutionX/blob/4a4486d6161028e69702bedec6cb071e6a5354c3/Source/player.cpp#L3416-L3454

Notice how each processing function can decide whether or not the loop should continue running, or if it should be aborted (variable `tplayer`). However, the `CheckNewPath` call does not provide this same functionality, as it does not affect `tplayer` which will always be `false` when it returns from the standing mode, as can be seen here:

https://github.com/diasurgical/devilutionX/blob/4a4486d6161028e69702bedec6cb071e6a5354c3/Source/player.cpp#L2075-L2078

`CheckNewPath` is where all requests to attack come from, meaning, whenever an attack animation is triggered from a standstill, the loop finishes and animation is incremented to frame 2 before `PM_DoAttack` has a chance to execute. Thus all logic that depends on frame 1 inside the attack functions will never execute.

There are a few ways to have frame 1 dependent logic to execute today, for example, attacking after walking. However, this is not how most normal attacks take place and thus why this problem is so impactful to weapon speed.

Here are a couple of saves to test the weapon speed behavior:
- Warrior (5 swords, one for each speed): [single_0.zip](https://github.com/diasurgical/devilutionX/files/5296086/single_0.zip)
- Rogue (3 bows, one for each speed): [single_1.zip](https://github.com/diasurgical/devilutionX/files/5296087/single_1.zip)

julealgon:
This PR makes it so that the 1st frame of animation for monsters, players and items are shown and processed properly. To achieve that effect, it relies on a minimally invasive approach by leveraging the animation delay system for players and monsters to force the first frame to remain unchanged for 1 game frame, and uses a simpler approach for items since they already have an animation flag by starting their animations on frame 0 instead of 1.

Today, the 1st frame of animation is always skipped, meaning that for any CEL animation file, only frames 2 onwards are displayed.

This change also fixes a problem related to how speed affixes affect weapons, since now the first frame will be taken into consideration for skipping logic for both melee and ranged attacks.

Feel free to use the save files provided in #839 to test this change out. For a fully smooth animation, consider applying #790 as well.

This resolves #838
This resolves #839

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1116,7 +1116,7 @@ void NewMonsterAnim(int i, AnimStruct *anim, int md)
 	MonsterStruct *Monst = monster + i;
 	Monst->_mAnimData = anim->Data[md];
 	Monst->_mAnimLen = anim->Frames;
-	Monst->_mAnimCnt = 0;
+	Monst->_mAnimCnt = -2;
 	Monst->_mAnimFrame = 1;
 	Monst->_mAnimDelay = anim->Rate;
 	Monst->_mFlags &= ~(MFLAG_LOCK_ANIMATION | MFLAG_ALLOW_SPECIAL);
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1812,7 +1812,7 @@ void SetupItem(int i)
 	item[i]._iPostDraw = FALSE;
 
 	if (!plr[myplr].pLvlLoad) {
-		item[i]._iAnimFrame = 1;
+		item[i]._iAnimFrame = 0;
 		item[i]._iAnimFlag = TRUE;
 		item[i]._iSelFlag = 0;
 	} else {
@@ -2410,7 +2410,7 @@ void RespawnItem(int i, BOOL FlipFlag)
 	item[i]._iPostDraw = FALSE;
 	item[i]._iRequest = FALSE;
 	if (FlipFlag) {
-		item[i]._iAnimFrame = 1;
+		item[i]._iAnimFrame = 0;
 		item[i]._iAnimFlag = TRUE;
 		item[i]._iSelFlag = 0;
 	} else {
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -422,7 +422,7 @@ void NewPlrAnim(int pnum, BYTE *Peq, int numFrames, int Delay, int width)
 	plr[pnum]._pAnimData = Peq;
 	plr[pnum]._pAnimLen = numFrames;
 	plr[pnum]._pAnimFrame = 1;
-	plr[pnum]._pAnimCnt = 0;
+	plr[pnum]._pAnimCnt = -1;
 	plr[pnum]._pAnimDelay = Delay;
 	plr[pnum]._pAnimWidth = width;
 	plr[pnum]._pAnimWidth2 = (width - 64) >> 1;


Source/monster.cpp
```diff
@@ -1116,7 +1116,7 @@ void NewMonsterAnim(int i, AnimStruct *anim, int md)
 	MonsterStruct *Monst = monster + i;
 	Monst->_mAnimData = anim->Data[md];
 	Monst->_mAnimLen = anim->Frames;
-	Monst->_mAnimCnt = 0;
+	Monst->_mAnimCnt = -2;
```
julealgon, line 1119: We need to use `-2` instead of `-1` due to an off-by-one mistake in `monsters.cpp` inside `ProcessPlayers`:

https://github.com/diasurgical/devilutionX/blob/4a4486d6161028e69702bedec6cb071e6a5354c3/Source/monster.cpp#L4618-L4634

Notice the check for animation length here:

https://github.com/diasurgical/devilutionX/blob/4a4486d6161028e69702bedec6cb071e6a5354c3/Source/monster.cpp#L4620

Instead of testing with `>` they use `>=`. If we start with `mAnimCnt` as -1, it will be incremented to 0 right away and satisfy this check right on the first game frame, defeating the purpose of the change.

A value of -2 guarantees that it will only increment a frame on the next tick.

julealgon:
Converting to draft as per `wontfix` tag. Might be useful for some folks.

AJenbo:
Closing this as I consider it fixed by https://github.com/diasurgical/devilutionX/commit/c7b9ffec1f1ddd95d92fc1bc0bac8258bf7ebfe4

PR #849, opened 2020-09-30
Block buying "red" consumables with full inventory

### LINKED ISSUE #848: Attempting to buy item that cannot be used while inventory is full results in lost item
While working on belt-related logic, I noticed that Adria/Pepin store checks do not currently check for items that cannot be used due to unmet requirements when purchasing items from them.

Surely enough, I was able to find a bug here.

## Repro steps
1. Fill inventory with loot. Ensure there are no free spaces (all 40 positions populated with some item)
2. Ensure there is at least one empty slot in the belt
3. Go to Adria and purchase a scroll whose requirements you do not meet (easy to do with Warrior)
4. Notice the gold is spent, but the item is not added to your inventory or belt

ItemMissingAfterBuying

Sample savefile ready to display this bug:
[single_0.zip](https://github.com/diasurgical/devilutionX/files/5308206/single_0.zip)

This happens because the game checks whether or not you can carry the item before presenting you with the confirmation dialog. During this check, done inside `SpecialAutoPlace`, the item is checked against the belt even if the requirements are not met by your character. However, items you cannot use due to missing requirements _cannot_ be put into the belt.

https://github.com/diasurgical/devilutionX/blob/09bd5f0bd88b07464fa55f76be5f74e284798cb0/Source/inv.cpp#L536-L547

The section above is executed right after testing all inventory positions. If the whole inventory is full, then the game checks to see if there is any empty belt slot where the item could be put. However, it only checks item sizes before doing this, and does not check whether the item's requirements are met. 

This means the confirmation dialog is shown, but the game fails when attempting to actually add the item to the belt later as it uses a more robust function to do that that tests whether the requirements for the item are met before allowing the belt to hold it.

## Expected Behavior
There should be a message saying you don't have space to put the item (same as if the belt was already full) for red items.

julealgon:
This fixes a problem that causes items to vanish after being bought from Adria by ensuring the inventory space verification logic uses consistent logic between checking for space before buying, and checking for space when adding the item to the player after purchase.

Right before confirming a purchase, the game performs a thorough check on the player's inventory and belt to see if there is any space left to house the item that will be bought. However, due to a poor guard in the belt verification logic, items that the player cannot currently use due to unmet requirements are still considered as valid candidates for putting the item in the belt: the store then allows the player to purchase the item, but fails when actually adding the item to the player as items with unmet requirements cannot be placed on the belt.

Pepin's store call is also modified for consistency, even though he does not sell any item with stat requirements, and thus would not be affected by the bug today.

This resolves #848

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -580,6 +580,21 @@ void DrawInvBelt()
 	}
 }
 
+/**
+ * @brief Gets the size, in inventory cells, of the given item.
+ * @param item The item whose size is to be determined.
+ * @return The size, in inventory cells, of the item.
+ */
+InvXY GetInventorySize(const ItemStruct &item)
+{
+	int itemSizeIndex = item._iCurs + CURSOR_FIRSTITEM;
+
+	return {
+		InvItemWidth[itemSizeIndex] / INV_SLOT_SIZE_PX,
+		InvItemHeight[itemSizeIndex] / INV_SLOT_SIZE_PX,
+	};
+}
+
 BOOL AutoPlace(int pnum, int ii, int sx, int sy, BOOL saveflag)
 {
 	int i, j, xx, yy;
@@ -635,7 +650,7 @@ BOOL AutoPlace(int pnum, int ii, int sx, int sy, BOOL saveflag)
 	return done;
 }
 
-BOOL SpecialAutoPlace(int pnum, int ii, int sx, int sy)
+BOOL SpecialAutoPlace(int pnum, int ii, const ItemStruct &item)
 {
 	int i, j, xx, yy;
 	BOOL done;
@@ -645,7 +660,9 @@ BOOL SpecialAutoPlace(int pnum, int ii, int sx, int sy)
 	if (yy < 0) {
 		yy = 0;
 	}
-	for (j = 0; j < sy && done; j++) {
+
+	InvXY itemSize = GetInventorySize(item); 
+	for (j = 0; j < itemSize.Y && done; j++) {
 		if (yy >= NUM_INV_GRID_ELEM) {
 			done = FALSE;
 		}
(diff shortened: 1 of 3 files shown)


AJenbo:
A bit invasive, feel like writing some tests :)

Source/inv.cpp
```diff
@@ -1563,7 +1570,7 @@ void AutoGetItem(int pnum, int ii)
 			if (w == 1 && h == 1) {
 				idx = plr[pnum].HoldItem.IDidx;
 				if (plr[pnum].HoldItem._iStatFlag && AllItemsList[idx].iUsable) {
-					for (i = 0; i < 8 && !done; i++) {
+					for (i = 0; i < MAXBELTITEMS && !done; i++) {
```
qndel, line 1573: this should go to devilution in a separate PR

julealgon: https://github.com/diasurgical/devilution/pull/2071

julealgon:
> A bit invasive, feel like writing some tests :)

@AJembo I can give it a shot. I might ping you for questions on how the tests work though.

AJenbo:
@julealgon guess this one conflicted with the clean up you also had waiting :)

AJenbo:
@julealgon any reason this isn't ready?

julealgon:
@julealgon any reason this isn't ready?

Ready now. Using the `:X:` version of the emojis, please check if that works for you @AJenbo .

AJenbo (approved):
Very nice

AJenbo:
No issues with the emojis this time :)

Source/inv.cpp
```diff
@@ -580,6 +580,70 @@ void DrawInvBelt()
 	}
 }
 
+/**
+ * @brief Gets the size, in inventory cells, of the given item.
+ * @param item The item whose size is to be determined.
+ * @return The size, in inventory cells, of the item.
+ */
+InvXY GetInventorySize(const ItemStruct &item)
```
glebm, line 588: These functions should be marked `static` or placed in an anonymous namespace

PR #858, opened 2020-10-03
Fix entering player's name with SDL 1.

ColinPitrat:
The static bool means there's one copy of the variable per compilation
unit using the header. So changing its value in one compilation unit
doesn't affect the other compilation units. This is clearly not what is
wanted here.

Fixing this requires adding a CPP in SourceS. This is the first one
there, I hope it's not an issue. The sdl2_to_1_2_backports should only
be used with SDL1 so I only include it in the list of sources for SDL1
builds. I'm not familiar with cmake so I hope I don't do anything stupid
there!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,6 +301,11 @@ set(devilutionx_SRCS
   Packaging/resources/CharisSILB.ttf
   Packaging/windows/devilutionx.rc)
 
+if(USE_SDL1)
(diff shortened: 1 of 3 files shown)


AJenbo:
Hmm I can't come up with a better solution, alternatly the file could be placed in SourceX, but I think it's better that it lives with the rest of sdl2_to_1_2_backports.

Thank you for addressing this issue.

PR #859, opened 2020-10-04
Fix typos

ColinPitrat:


AJenbo:
pease target upstram with this typ of corrections https://github.com/diasurgical/devilution

ColinPitrat:
Ack, will do.
Some of them are not upstream though, it seems there are differences between their Source directories. Is that expected?

AJenbo:
Yes, DevilutionX applies some level of refactoring, most often to simplify code to make it easier to fix a particular bug or introduce a new feature. So in those cases, the corrections should still be applied directly in this repo. It does mean a bit of extra work, but it keeps things cleaner overall.

ColinPitrat:
Thanks, I'll do a separate PR later for the devilutionX only parts.

PR #861, opened 2020-10-05
Autoequip all items on pickup

julealgon:
This change introduces automatic equipping of items picked up in the ground when nothing is currently equipped in the corresponding body slots.

This standardizes this behavior for all items instead of only allowing weapons to be auto-equipped.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -643,6 +643,164 @@ bool CanBePlacedOnBelt(int playerNumber, const ItemStruct &item)
 	return false;
 }
 
+/**
+ * @brief Checks whether the given item can be equipped. Since this overload doesn't take player information, it only considers
+ * general aspects about the item, like if its requirements are met and if the item's target location is valid for the body.
+ * @param item The item to check.
+ * @return 'True' in case the item could be equipped in a player, and 'False' otherwise.
+ */
+bool CanEquip(const ItemStruct &item)
+{
+	return
+		item._iLoc != ILOC_INVALID &&
+		item._iLoc != ILOC_NONE &&
+		item._iLoc != ILOC_UNEQUIPABLE &&
+		item._iStatFlag;
+}
+
+/**
+ * @brief A specialized version of 'CanEquip(int, ItemStruct&, int)' that specifically checks whether the item can be equipped
+ * in one/both of the player's hands.
+ * @param playerNumber The player number whose inventory will be checked for compatibility with the item.
+ * @param item The item to check.
+ * @return 'True' if the player can currently equip the item in either one of his hands (i.e. the required hands are empty and
+ * allow the item), and 'False' otherwise.
+ */
+bool CanWield(int playerNumber, const ItemStruct &item)
+{
+	if (!CanEquip(item) || (item._iLoc != ILOC_ONEHAND && item._iLoc != ILOC_TWOHAND))
+		return false;
+
+	PlayerStruct &player = plr[playerNumber];
+	ItemStruct &leftHandItem = player.InvBody[INVLOC_HAND_LEFT];
+	ItemStruct &rightHandItem = player.InvBody[INVLOC_HAND_RIGHT];
+
+	if (leftHandItem.isEmpty() && rightHandItem.isEmpty()) {
+		return true;
+	}
+
+	if (!leftHandItem.isEmpty() && !rightHandItem.isEmpty()) {
+		return false;
+	}
+
+	ItemStruct &occupiedHand = !leftHandItem.isEmpty() ? leftHandItem : rightHandItem;
+
+	// Barbarian can wield two handed swords and maces in one hand, so we allow equiping any sword/mace as long as his occupied
+	// hand has a shield (i.e. no dual wielding allowed)
+	if (player._pClass == PC_BARBARIAN) {
+		if (occupiedHand._itype == ITYPE_SHIELD && (item._itype == ITYPE_SWORD || item._itype == ITYPE_MACE))
+			return true;
+	}
+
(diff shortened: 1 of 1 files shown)


Source/inv.cpp
```diff
@@ -450,6 +450,99 @@ void DrawInvBelt()
...
+{
+	return item._itype == ITYPE_NONE;
+}
+
+BOOL CanEquip(const ItemStruct &item)
+{
+	return item._iLoc != ILOC_INVALID && item._iLoc != ILOC_NONE && item._iLoc != ILOC_UNEQUIPABLE && item._iStatFlag;
+}
+
+BOOL CanWield(int playerNumber, const ItemStruct &item)
```
AJenbo, line 463: Could you add a breaf description of what this does. I'm unsure what the difference between this and CanEquip is just by the names.

Source/inv.cpp
```diff
@@ -450,6 +450,99 @@ void DrawInvBelt()
...
+	if (!CanEquip(playerNumber, item, bodyLocation)) {
+		return FALSE;
+	}
+
+	plr[playerNumber].InvBody[bodyLocation] = item;
+
+	NetSendCmdChItem(FALSE, bodyLocation);
+	CalcPlrInv(playerNumber, TRUE);
+	if (playerNumber == myplr) {
+		PlaySFX(ItemInvSnds[ItemCAnimTbl[item._iCurs]]);
```
AJenbo, line 525: If you can remove the added sfx I think I'm ok merging it for 1.2.0

Source/inv.cpp
```diff
@@ -637,32 +730,6 @@ BOOL GoldAutoPlace(int pnum)
 	return done;
 }
 
-BOOL WeaponAutoPlace(int pnum)
```
AJenbo, line 640: Did you check if the affected functions are bin exact for hellfire?

Source/inv.cpp
julealgon: I used the term `wield` on purpose since it means "using with your hands" in English (basically). It is a more specific version of `equip` in Diablo's case.

So `CanEquip(item, location_left_hand)` means `left_hand.IsEmpty && CanWield(item)`

I can add comments on it, but I thought it wasn't necessary if you understand the term, felt self-explanatory to me. Let me know if you want me to add comments though.

Source/inv.cpp
julealgon: Can you elaborate a bit on why? I felt like adding the sounds resulted in a really nice experience here and is consistent with the sound it makes when you manually equip the item.

I feel like in this case it improves on the game (without being invasive) while resulting in simpler code: every call to equip will play the sound unconditionally (I intend to use these methods for other things later).

Let me know your thoughts.

Source/inv.cpp
julealgon: Honestly didn't... I feel like it _will_ end up affecting Hellfire, since the Bard class allows dual wielding... let me check and update here.

Source/inv.cpp
AJenbo: I get that, but I still find it a bit ambigious if I should not use weapons with `CanEquip`, or will work for both. Basically, it's not clear if `CanEquip` is just a poorly named `CanWear` since there is a special function for weapons. If `CanEquip` works for weapons, then why would I ever use `CanWield`?

Names of functions definatly should be descriptive, but if we at some point come in doubth about what the 50 lines of code do, it emensly helpful to have a short sentince that explains the context.

Source/inv.cpp
AJenbo: Because I want to be as conservative as possible when changing behaviour. Atm. this PR makes one consisten, but also introduces a new one.

Adding audio feedback is nice, yes. But it's not something the game does. Adding this open the gate for also having sounds when you level up, items break, stafs run out of charges etc.

Source/inv.cpp
julealgon: There is a bunch of Hellfire customizations in there as I suspected:

https://github.com/diasurgical/devilution/blob/81d57286991127c1f41a8da10d2d1e744cce7ab1/Source/inv.cpp#L989-L1002

What do you suggest in this case? Wait for Hellfire?

Source/inv.cpp
julealgon: I disagree on this being ambiguous but I understand this is highly subjective. Will add the comments as suggested.

Source/inv.cpp
AJenbo: Hmm, the changes aren't too bad. If I'm reading it right they made the Monk never auto equip weapons, and bard and barbarian only if it's a mace or sword (implemented in two completely different ways...).
I don't think that will be hard to implement in the change you have made here. And the conflict looks to be reasonable to deal with.

Source/inv.cpp
julealgon: Commit for playing audio removed from the branch.

Source/inv.cpp
julealgon: Added comments on all new methods.

julealgon:
Rebased and adapted for Hellfire classes.

Source/inv.cpp
julealgon: Changes have now been translated into the new functions.

mewmew:
Would be fun if you were to pick up an already identified magic item which has cursed item powers; e.g. given -10 to vitality, thus killing you :)

AJenbo:
@julealgon rebase?

julealgon:
> @julealgon rebase?

Depends on #1075 now.

julealgon:
Good to go now @AJenbo . Give it a whirl.

Issue #857, opened 2020-10-02
Allow quick equipment management with shift+click


julealgon:
This proposal is to add new functionality to the inventory, allowing quicker management of equipment by leveraging shift+clicking.

This does not conflict with "move to/from belt" (#844) as items that can go in the belt are never equipable and vice-versa.

## Scenario 1 - Shift+click on equiped item

Shift clicking on an equiped item should unequip that item and move it to inventory.
If inventory is full, "no room/overburdened" sound should be played and nothing should happen.

---

## Scenario 2 - Shift+click on equipment in inventory, corresponding equip slot empty

Shift clicking on an equipment item in the inventory should try to equip that item piece
1. If the item cannot be equiped due to requirements, "i cant use this, yet" plays and nothing happens
2. If the item can be equiped, equip the item, moving it from the inventory to the body

---

## Scenario 3 - Shift+click on equipment in inventory, corresponding equip slot has an item already

The equipment from inventory is equiped, and the equiped item is moved to inventory.
1. If the item can be equiped, but only due to attributes granted by the currently equiped item(s), follow existing logic: equip the item, but it immediately becomes red (questionable if this is intended behavior, but that's how it is today)
2. If there is not enough room in inventory after removing inv item and putting the existing equipment in the inventory, play "overburdened" sound and do nothing.

---

## Scenario 4 - Help screen

Help screen should have information about the new functionality. Differentiate from moving to/from belt

---

## Scenario 5 - Tooltip

Tooltip related to shift+clicking could be added to elligible items.
1. Mouse over equipped item: "Shift+click to unequip"
2. Mouse over equipment in inventory: "Shift+click to equip"
3 (possible) Mouse over equipment in inventory + equipment already exists: "Shift+click to swap equipment"

Challenge with tooltip is number of lines limit in info panel: magic item with 2 affixes and requirements occupies all 5 lines

AJenbo:
I would prefer that this be split up in a PR for the functionality and one for tooltips.

The hardest part is probably checking that there is room when swapping 1 two handed item for two items.

julealgon:
> I would prefer that this be split up in a PR for the functionality and one for tooltips.

That's probably what I'll do indeed. The tooltips for these equipments is a tough problem to solve due to spacing.

> The hardest part is probably checking that there is room when swapping 1 two handed item for two items.

I have already extracted a function for moving an item to inventory with the other belt->inventory feature, so this should not be a problem

AJenbo:
[...]

PR #864, opened 2020-10-06
Equip/unequip items with shift click.

julealgon:
This change allows equipping and unequipping items by shift+clicking into the item in the inventory.

This initial implementation does not replace existing items, that is, it works to remove equiped items and to equip items into empty slots.

This partially implements #857


julealgon:
Leaving as draft for now as this depends on a few other changes being merged first.

Source/diablo.cpp
```diff
@@ -654,6 +654,8 @@ LRESULT GM_Game(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
 
 BOOL LeftMouseDown(int wParam)
 {
+	BOOL shiftIsHeld = wParam == DVL_MK_SHIFT + DVL_MK_LBUTTON;
```
AJenbo, line 657: You probably want to treat this as a bitflag. Also `BOOL` is just an int, might want to use bool for any new code.
```suggestion
	bool shiftIsHeld = wParam & (DVL_MK_SHIFT | DVL_MK_LBUTTON);
```

julealgon: > You probably want to treat this as a bitflag.

Oh, that looks better indeed. I didn't really pay a lot of attention to this line since I just extracted the existing logic into a variable for reuse.

> Also BOOL is just an int, might want to use bool for any new code.

`bool` for new code it is. Was using `BOOL` because I thought it was some sort of project-wide standard.

AJenbo: It was for the original Diablo, but it's actually a WinAPI type that isn't understood by compilers and so gets less error checking and optimizations. I hope to one day replace all cases of `BOOL` (`int`) and `BOOLEAN` (`char`) with `bool` :)

AJenbo (approved):

PR #714, opened 2020-04-19
experimental xbrz filtering

Chance4us:
So that we no longer see the void in the galaxy this is the follow up to  #696.
It crashes e.g. when upscale = 0 in diablo.ini.
The main menu is now scaled as well.
The cutscenes are now scaled but hardcoded for 16:9 display.
The credits are not displayed correctly.

Images for comparison:
pos1-town-pepin
pos1-xbrz-town-pepin
pos1-town-char-inv-unique
pos1-xbrz-town-char-inv-unique
pos1-town-pepin-inv
pos1-xbrz-town-pepin-inv
butcher
xrbz-butcher
pos2-dungeon
pos2-xbrz-dungeon
pos2-dungeon-inv
pos2-xbrz-dungeon-inv
pos3-caves
pos3-xbrz-caves
warlord
xbrz-warlord

PR #866, opened 2020-10-08
experimental xbrz filtering (#714)

AJenbo:
Recover the branch from https://github.com/diasurgical/devilutionX/pull/714

See the previous PR for comparison screenshots.


escalade:
Needs rebase :)

AJenbo:
@escalade I'm not going to be rebasing it since I'm not the original author and also have way to much else to look into atm. If you or @Chance4us are interested in doing so you would be more than wellcome.

PR #872, opened 2020-10-11
Handle DragonFly BSD in CMakeLists.txt

kraileth:
Using Ravenports, devilutionX has been available on DragonFly BSD, too, for a while now. This proposed change would allow us to drop one local compatibility patch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,7 @@ elseif(TARGET_PLATFORM STREQUAL "cpigamesh")
   include(cpigamesh_defs)
 endif()
 
-if(${CMAKE_SYSTEM_NAME} STREQUAL FreeBSD OR ${CMAKE_SYSTEM_NAME} STREQUAL OpenBSD)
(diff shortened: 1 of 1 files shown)


AJenbo:
I was wondering where you guys where :)

kraileth:
@AJenbo Can't say much about the other *BSD people, but I'm still here watching DevilutionX (especially hoping for Hellfire support to land! BTW: Is there some HF support in the current release?) However time for more than reading what happens is scarce. Wanted to upstream this patch for a while, glad that it finally worked out.

(And no, CI stuff didn't drop off my list -- it's just the same thing as with almost everybody: Life happens and then you don't get to things you want to do.)

qndel:
no hellfire support atm ;)

kraileth:
@qndel Thanks for confirming! Then I don't even have to try. ;)

AJenbo:
@kraileth Hellfire incoming

kraileth:
@AJenbo Really looking forward to that! Thanks for all the work that everybody involved here is doing. :+1:

AJenbo:
you can now build and try out Hellfire, it still has some missing parts, but we are working on it.

PR #875, opened 2020-10-12
better direction calculation

pionere:
- no limitation on the vector's length
- no need for error-prone tables

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -261,113 +261,130 @@ int GetSpellLevel(int id, int sn)
 	return result;
 }
 
+/**
+ * @brief Returns the direction a vector from p1(x1, y1) to p2(x2, y2) is pointing to.
+ *
+ *      W    SW     S
+ *            ^
+ *            |	
+ *     NW ----+---> SE
+ *            |
+ *            |
+ *      N    NE     E
+ *
+ * @param x1 the x coordinate of p1
+ * @param y1 the y coordinate of p1
+ * @param x2 the x coordinate of p2
+ * @param y2 the y coordinate of p2
+ * @return the direction of the p1->p2 vector
+*/
 int GetDirection8(int x1, int y1, int x2, int y2)
 {
-	BYTE Dirs[16][16] = {
-		{ 99, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 },
-		{ 2, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 },
-		{ 2, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0 },
-		{ 2, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0 },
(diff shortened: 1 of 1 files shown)


glebm:
I wonder if this function is a hotspot? The table version has fewer branches.

pionere:
> I wonder if this function is a hotspot? The table version has fewer branches.

what are you talking about?
table version: 6 branches in any case
new version: 4 branches at most

glebm:
I was estimating the number of branches by looking at the level of nesting. After looking more carefully, I see that it shouldn't be an issue. Perhaps this improves performance even?

AJenbo:
Wrote a test, there is a slight difference in how `GetDirection16` works now:
```
SourceT/missiles_test.cpp:50: Failure
Expected equality of these values:
  1
  dvl::GetDirection16(0, 0, 8, 11)
    Which is: 0
SourceT/missiles_test.cpp:58: Failure
Expected equality of these values:
  9
  dvl::GetDirection16(8, 11, 0, 0)
    Which is: 8
SourceT/missiles_test.cpp:66: Failure
Expected equality of these values:
  13
  dvl::GetDirection16(0, 8, 11, 0)
    Which is: 12
```

AJenbo:
It looks like the main cardinal directions become narrower with this change:
dir8

Source/missiles.cpp
```diff
@@ -261,113 +261,130 @@ int GetSpellLevel(int id, int sn)
...
+			md = 2; // DIR_W
+		} else {
+			my = -my;
+			if (5 * mx <= (my << 1))
+				return 5; // DIR_NE
+			md = 4; // DIR_N
+		}
+		if (5 * my <= (mx << 1))
+			md = 3; // DIR_NW
+	}
```
julealgon, line 314: Seems to me like the `if` block and the `else` block are pretty much identical. Is there no way to share the logic here?

I see the original code was using `abs`. Can't we do something similar to avoid having to duplicate all the checks?

julealgon (commented):
This code has a very considerable amount of logic duplicated and is incredibly hard to follow (at least for me).

galaxyhaxz:
Personally I'm in favor of it, with a bit of tweaking to get it closer to vanilla. Replacing hardcoded tables and other crap is a good thing, and I prefer the code method!

AJenbo:
After some investigation we have concluded that the original isn't very accurate:
image

The new one pretty much fixes this issue:
image

In terms of gameplay impact, this only consequence seems to be that charging monsters will now be slightly more accurate when moving in some directions.

AJenbo (approved):

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/79ed3c897bb0abd03024c5433ed47939bd738d06

Source/missiles.cpp
AJenbo: @pionere did clean it up, but never updated the PR so some one will  have to go dig it out:
https://github.com/pionere/devilutionX/blob/master/Source/missiles.cpp#L285

AJenbo: @pionere has made a more minimalistic version, but some one would have to go dig it out:
https://github.com/pionere/devilutionX/blob/master/Source/missiles.cpp#L285

PR #878, opened 2020-10-12
Add hellfire cmake target

mmaret:
May be not desirable to merge before the devilutionx ui get update for it too


AJenbo (approved):
This works for now :+1:

PR #879, opened 2020-10-12
set macOS deployment target to 10.12.0 (macOS Sierra)

spitfire:
Target compiles and runs on 10.12, also tested on 10.13 and 10.15 - this will let people play devilutionX on older Macs (dating back to 2010).
Because it is linked to libraries from homebrew it should be compiled on 10.12.x with Xcode 9.2


AJenbo (approved):
oh very nice work :+1:

PR #882, opened 2020-10-13
Fix OpenDingux icon

glebm:
OpenDingux / RetroFW icons must be 32x32 because the menu does no scaling.

Icon resized from the large version.

Before:
PXL_20201013_085706964


AJenbo:
thanks, after release i noticed that retrofw just showed a ?, initial searches indicated it should have stayed a 32x32 icon, but i wasn't sure.

Issue #378, opened 2019-10-20
Radon fails to compile on NetBSD


teknokatze:
**Describe the bug**
/usr/work/games/devilutionx/work/devilutionX-0.5.0/3rdParty/Radon/Radon/source/Key.cpp:33:18: error: use of undeclared identifier 'atof'
                return (float)(atof(value.data()));
                               ^
1 error generated.

**To Reproduce**
with pkgsrc, cd to games/devilutionx, make

This is with clang from NetBSD 9.99.15 base (clang 7).
None of the include files in Radon seem to include <stdlib.h> which provides atof.

AJenbo:
@ng-0 we currently don't officially support NetBSD, would you be able to make a PR with the needed changes?

ng-0:
Hi,

Okay, I will do so or defer to the maintainer of the pkgsrc package to make the PR.

niacat:
NetBSD support depends on ladislav-zezula/StormLib/pull/160

Otherwise it's very simple. I haven't tested with clang yet (the default system compiler is gcc).

PR #888, opened 2020-10-19
Support building on NetBSD.

niacat:
#378 can be closed.


AJenbo:
Oh forgot about NetBSD, wonder if we are missing any other BSDs?

Cool that you have UBSAN

README.md
```diff
@@ -63,6 +63,19 @@ cmake ..
...
 </details>
+<details><summary>NetBSD</summary>
+
+### Installing dependencies
+```
+pkgin install cmake SDL2_mixer SDL2_ttf libsodium
+```
+### Compiling
+```
+cd build
```
glebm, line 74: If cmake is reasonably recent on NetBSD, `cd` can be avoided:

```bash
cmake -S. -Bbuild ..
cmake --build build -j $(sysctl -n hw.ncpu)
```

niacat: Cool. It should be the most recent cmake release.

niacat:
@AJenbo We've also got ASAN but building with that I got a segfault immediately after launching devilutionx, unfortunately

Every currently maintained BSD descendant should now be covered by the CMakeLists, there's some older ones that aren't maintained any more though.

kraileth:
Well, MidnightBSD is actually actively maintained and has it's own kernel name, but it's extremely exotic. The others that I can think of are either customized versions of one of the big others (and don't use a different kernel ident) or are obviously / most likely inactive.

PR #891, opened 2020-10-20
Fix linking error with libSDL2main

madmann91:
This is the proper fix for the error seen on Fedora machines:

`make[2]: *** No rule to make target '/usr/lib64/libSDL2main.a', needed by 'devilutionx'.  Stop.`

The only change this PR makes is to force the use of the local `FindSDL2.cmake` file that is present in the repository, instead of relying on the installed `sdl2-config.cmake` file. A detailed explanation as to why this is required can be found on the previous PR: https://github.com/diasurgical/devilutionX/pull/756#issuecomment-712832940.


AJenbo:
Thanks for solving this, looks to be working with Switch and other targets so indeed this appears to be the proper solution.

AJenbo (approved):

PR #897, opened 2020-10-25
sharing updates

KonstantineLazarenko:


AJenbo:
This is impossible to merge, there are many changes that shouldn't be there and also this is manually patched from a very old branch.

Instead, I'll rebase the android branch and send you a PR, to begin with. Please create a branch in your repo that 1:1 with this repos master.

PR #899, opened 2020-10-26
possible fix for broken light and desyncs

qndel:
Fixes multiple vanilla issues related to light:
- player light being off after level transitions except when entering for the first time
- unique monster light being away from the monster till it moves
- monster lights returning to their spawn points after reentering the level
- light sources spawned dynamically (for example opening the final room in valor) being overwritten by default level light
- fix player being knocked back out of own light/vision radius
- fix inconsistency between light radius and vision
- fix restoring dFlags including player's vision - player could see stuff near town portal etc.
- properly update vision while moving
- tweak vCrawlTable to be more consistent - see pic below

Vision with light radius 2 before and after vCrawlTable tweak
It also affects novas and guardian but I think the overall result is the same - it would only differ if they only searched in radius up to 2, which is not the case, so the only ingame difference should be extended vision on light radius 2 (-80% as the default one is 100% - 10)
testrad2


Source/diablo.cpp
```diff
@@ -1288,7 +1288,19 @@ void PressChar(int vkey)
...
 		difficulties[2] = "Hell";
 		sprintf(pszStr, "%s, mode = %s", gszProductName, difficulties[gnDifficulty]);
-		NetSendCmdString(1 << myplr, pszStr);
+		char sad[169];
+		for (int i = 0; i < nummonsters; i++) {
+			MonsterStruct *mon = &monster[monstactive[i]];
+			if (mon->mlid != 0) {
+				LightListStruct *lid = &LightList[lightactive[mon->mlid]];
+
+				float d = ((float)mon->_mx - (float)lid->_lx) * ((float)mon->_mx - (float)lid->_lx) + ((float)mon->_my - (float)lid->_ly) * ((float)mon->_my - (float)lid->_ly);
```
AJenbo, line 1297: Why float, there are no devisions

qndel: good point :D

Source/diablo.cpp
```diff
@@ -1288,7 +1288,19 @@ void PressChar(int vkey)
 		difficulties[1] = "Nightmare";
 		difficulties[2] = "Hell";
 		sprintf(pszStr, "%s, mode = %s", gszProductName, difficulties[gnDifficulty]);
-		NetSendCmdString(1 << myplr, pszStr);
```
mmaret, line 1291: You could remove pszStr declaration then

Source/diablo.cpp
```diff
@@ -1288,7 +1288,19 @@ void PressChar(int vkey)
...
 		difficulties[2] = "Hell";
 		sprintf(pszStr, "%s, mode = %s", gszProductName, difficulties[gnDifficulty]);
-		NetSendCmdString(1 << myplr, pszStr);
+		char sad[169];
+		for (int i = 0; i < nummonsters; i++) {
+			MonsterStruct *mon = &monster[monstactive[i]];
+			if (mon->mlid != 0) {
+				LightListStruct *lid = &LightList[lightactive[mon->mlid]];
+
+				float d = (mon->_mx - lid->_lx) * (mon->_mx - lid->_lx) + (mon->_my - lid->_ly) * (mon->_my - lid->_ly);
```
mmaret, line 1297: "d" Can be declared as int right ?

qndel: ugh yes xDD

Source/diablo.cpp
qndel: this code is only for verifying that this is a proper fix, if it works, then it will be removed later

mmaret: :+1:

structs.h
```diff
@@ -602,7 +602,7 @@ typedef struct MonsterStruct { // note: missing field _mAFNum
 	unsigned char leader;
 	unsigned char leaderflag;
 	unsigned char packsize;
-	unsigned char mlid;
```
AJenbo, line 605: There are 200 monsters and in Hellfire, all monsters can receive light (if baserked). To avoid having to deal with the save game compatibility of expanding this to `Sint16`/`Sint32` I recommend that we set `NO_LIGHT` to be `UCHAR_MAX`. This may in fact even be bin exact.

qndel: I believe I've tried that and it didn't work, light code takes an int which is -1, you can't play around that with 255

AJenbo: You're also right that I got this wrong. I thought it was the light tracking the monster id, but it's the other way around so a signed char is plenty big.

Source/lighting.cpp
```diff
@@ -431,31 +431,33 @@ char *pCrawlTable[19] = {
...
-	{ 0, 1, 1, 2, 1, 3, 2, 4, 2, 5, 3, 6, 3, 7, 3, 8, 4, 9, 4, 10, 5, 11, 5, 12, 6, 13, 6, 14, 0, 0 },
-	{ 0, 1, 1, 2, 1, 3, 1, 4, 2, 5, 2, 6, 3, 7, 3, 8, 3, 9, 4, 10, 4, 11, 4, 12, 5, 13, 5, 14, 0, 0 },
-	{ 0, 1, 1, 2, 1, 3, 1, 4, 1, 5, 2, 6, 2, 7, 2, 8, 3, 9, 3, 10, 3, 11, 3, 12, 4, 13, 4, 14, 0, 0 },
-	{ 0, 1, 0, 2, 1, 3, 1, 4, 1, 5, 1, 6, 1, 7, 2, 8, 2, 9, 2, 10, 2, 11, 2, 12, 3, 13, 3, 14, 3, 15 },
-	{ 0, 1, 0, 2, 0, 3, 1, 4, 1, 5, 1, 6, 1, 7, 1, 8, 1, 9, 1, 10, 1, 11, 2, 12, 2, 13, 2, 14, 2, 15 },
-	{ 0, 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 1, 8, 1, 9, 1, 10, 1, 11, 1, 12, 1, 13, 1, 14, 1, 15 },
-	{ 0, 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8, 0, 9, 0, 10, 0, 11, 0, 12, 0, 13, 0, 14, 0, 15 }
+	{ 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8, 0, 9, 0, 10, 0,  11, 0,  12, 0,  13, 0,  14, 0,  15, 0 },
+	{ 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8, 1, 9, 1, 10, 1,  11, 1,  12, 1,  13, 1,  14, 1,  15, 1 },
+	{ 1, 1, 2, 0, 3, 0, 4, 1, 5, 1, 6, 1, 7, 1, 8, 1, 9, 1, 10, 1,  11, 1,  12, 2,  13, 2,  14, 2,  15, 2 },
```
AJenbo, line 438: This is going to affect how a lot of missiles behave so I would rather get a better understanding of why the is as limited as it is. Probably we simply need to cap it in the calculation.

Source/diablo.cpp
```diff
@@ -1649,26 +1662,27 @@ void LoadGameLevel(BOOL firstflag, int lvldir)
 		if (leveltype != DTYPE_TOWN) {
 			if (firstflag || lvldir == ENTRY_LOAD || !plr[myplr]._pLvlVisited[currlevel] || gbMaxPlayers != 1) {
 				HoldThemeRooms();
-				glMid1Seed[currlevel] = GetRndSeed();
+				glMid1Seed[currlevel] = sglGameSeed;
```
AJenbo, line 1665: I don't think I get why this is needed?

AJenbo:
Cherry-picked some of the lighter stuff and rebased the rest so it's a bit easier to ready the change history on this :)

I'll look at the rest of the changes later and merge them once I fully understand what is going on here :D

Source/player.cpp
```diff
@@ -1084,6 +1084,7 @@ void InitPlayer(int pnum, BOOL FirstTime)
 
 		if (pnum == myplr) {
 			plr[pnum]._plid = AddLight(plr[pnum]._px, plr[pnum]._py, plr[pnum]._pLightRad);
+			ChangeLightXY(plr[myplr]._plid, plr[myplr]._px, plr[myplr]._py); // fix for a bug where old light is still visible at the entrance after reentering level
```
AJenbo, line 1087: I have opted to fix this in `UpdateMonsterLights()` as that function can screw up light in existing saves and was called after `InitPlayer()`.

AJenbo: And it's reverted with your fix and a better one for handling the monsters with incorrect light id in old saves :) https://github.com/diasurgical/devilutionX

Source/diablo.cpp
qndel: because GetRndSee() modifies the seed?

galaxyhaxz: It can screw up the seed. IMO best way to handle it to have to RNG's one strictly for dungeon generation and another for the rest that can be called without worry.

qndel: this was the only rng desync I've spotted so it works fine in general I'd say except this where they did not one but two mistakes :D

Source/player.cpp
```diff
@@ -2474,6 +2476,8 @@ BOOL PM_DoWalk3(int pnum)
 
 	if (plr[pnum]._pVar8 >= anim_len) {
 		dPlayer[plr[pnum]._px][plr[pnum]._py] = 0;
+		if (leveltype != DTYPE_TOWN)
```
AJenbo, line 2479: Should check if pnum == myplr ;)

AJenbo:
The improper shape of the vision cone generated from the crawl table is also present at other sizes:
image

AJenbo:
Here is the crawl table represented as light as best I understand it:
image

And raw:
image

AJenbo:
Another related issue is that lights never go fully dark on the edge of the vision cone:
image

I'm thinking this is very similar to the UnDovision() that needed to be added to the walk routine.

PR #533, opened 2020-01-05
Implement translation support

AJenbo:
This brings translation support to the game, it will probably need a bit of adjustment on different platforms and for the handling of static texts. Currently, also only the strings from DiabloUI and appfat.cpp are marked as translatable.

Fixes #66

Skærmbillede fra 2020-01-07 23-02-49

PR #902, opened 2020-10-28
Implement translation support using tinygettext

AJenbo:
This implements translation support using tinygettext instead of gnu gettext to try and solve bulding it on all supported platofrms.
Previous attempt using gnu gettext: https://github.com/diasurgical/devilutionX/pull/533

Only the version info that is displayed ingame is translatable for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -379,6 +379,27 @@ if(RUN_TESTS)
   gtest_add_tests(${BIN_TARGET} "" AUTO)
 endif()
 
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/external/cmake-modules/)
+include(ExternalProject)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -379,6 +380,28 @@ if(RUN_TESTS)
   gtest_add_tests(${BIN_TARGET} "" AUTO)
 endif()
 
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/external/cmake-modules/)
+include(ExternalProject)
+set(TINYGETTEXT_PREFIX ${CMAKE_BINARY_DIR}/tinygettext/)
+ExternalProject_Add(tinygettext
```
glebm, line 386: Perhaps this should be in a separate file under `./CMake/`?

PR #904, opened 2020-10-29
Implement timedemo

AJenbo:
This allows you to record a session and then replay it later as fast as the system CPU will allow

This can be used to benchmark the game, as well as do automatic testing of the game after changing parts of the code to verify that it behaves identically before and after the change.

Attached is a save game and recording that can be used for testing
[demo0.zip](https://github.com/diasurgical/devilutionX/files/5470234/demo0.zip)

`./devilutionx --demo 0 --timedemo -f`

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -824,17 +824,29 @@ void RunGameLoop(interface_mode uMsg)
 		}
 		if (!gbRunGame)
 			break;
-		if (!nthread_has_500ms_passed()) {
-			ProcessInput();
+
+		bool drawGame = true;
+		bool processInput = true;
+		bool runGameLoop = demoMode ? GetDemoRunGameLoop(drawGame, processInput) : nthread_has_500ms_passed();
+		if (!runGameLoop) {
+			if (recordDemo != -1)
+				demoRecording << static_cast<uint32_t>(DemoMsgType::Rendering) << "," << gfProgressToNextGameTick << "\n";
+			if (processInput)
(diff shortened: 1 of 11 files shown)


Josegto1 (approved):

AJenbo:
Not perfect, but very useful - merged

Chance4us:
really cool

AJenbo:
@Chance4us it would be super helpful if you could do a play through a large portion of the game and upload it as a recording.

The thing is a little rough atm and we will likely change the process and format down the road, so only do it if you are ok with it having to be redone at some point.

Here is the current process:
- Create a hero, start a game, save and exit.
- Start the game with the following parameters `--record #` (replace # with the number in the save game you just created)
- When the recording is done simply exit the game (do not save!)

You can test the recording by with the following command parameters `--demo #`, the playback will try and respect the speed set in your ini. You can also play back at max speed that your CPU can handle by adding `--timedemo` at the end you will get a status in the terminal saying how fast it was able to play the demo:
`8389 frames, 56.72 seconds: 147.9 fps`

Once you are happy with the recording you need to upload 3 files `demo_#.dmo`, `single_#.#sv` and `diablo.ini`.

This way we can replicate your play through when we make changes to the game :D

I would suggest avoiding use of things we know that we are going to change, like the phasing spell, but don't let that hold you back to much :)

A handy thing, during playback you can press esc to resume control of the game at any state. This could also make it really useful for bug reporting :D

PR #911, opened 2020-11-07
valor fix

qndel:
Fixes treating quest items as dupes and the effect of valor quest bugging out - the cause is still unknown


Source/objects.cpp
```diff
@@ -3396,11 +3396,11 @@ void OperatePedistal(int pnum, int i)
 	BYTE *mem;
 	int iv;
 
-	if (object[i]._oVar6 != 3) {
+	if (object[i]._oVar6 != 3 || object[i]._oAnimFrame != 4) {
 		if (PlrHasItem(pnum, IDI_BLDSTONE, &iv) != NULL) {
 			RemoveInvItem(pnum, iv);
+			object[i]._oVar6 = object[i]._oAnimFrame; // should fix broken valor quest by syncing oVar6 with the pedestal frame
```
AJenbo, line 3402: Do you know of a way to replicate this issue?

qndel: nope, if I did, I would've fixed the cause. It's kind of rare, most visible when speedrunners get it and from what I've heard, they had different results, like opening all valor rooms with 1 stone (means initial counter - oVar6 - had 2 value) or none at all (counter at 3+) so we know there's something that ruins it

AJenbo: K, I'll try and see if I can figure it out :)

AJenbo:
@qndel rebase?

qndel:
didn't galaxy create a proper fix?

AJenbo:
This one? https://github.com/diasurgical/devilutionX/commit/b64beca78bd9d5977d0fcfa02971c1be78567564

Wasn't sure if there were some parts that were or weren't covered :)

PR #917, opened 2020-11-14
Add a possibility to select SDL audio device other than default

pgielda:
Currently devilutionx is using SDL_OpenAudioDevice with first parameter set as NULL which makes it always default to audio device 0. Same behaviour is with Mix_OpenAudio that under the hood does the same.

This PR allows to override this setting to allow playback on other audio devices. E.g. in my use case is the HDMI output of my monitor has id=4.

(use "aplay -l" to list the audio devices in Linux)

I hope it does not break anything but I did only limited testing on Debian due to lack of time. I still think this is a useful feature to add.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -493,6 +493,9 @@ foreach(target devilution ${BIN_TARGET} SourceS)
       ${SDL_LIBRARY})
     target_compile_definitions(${target} PRIVATE USE_SDL1)
   else()
+    if(SDL2_mixerPC_VERSION VERSION_GREATER_EQUAL 2.0.2)
+      target_compile_definitions(${target} PRIVATE SDL2_MIXER_VERSION_AT_LEAST_2_0_2)
+    endif()
     target_link_libraries(${target} PRIVATE
       SDL2::SDL2main
       SDL2::SDL2_ttf
--- a/SourceX/sound.cpp
+++ b/SourceX/sound.cpp
@@ -6,6 +6,7 @@
 #include "all.h"
 #include "../3rdParty/Storm/Source/storm.h"
 #include "stubs.h"
+#include "DiabloUI/diabloui.h"
 #include <SDL.h>
 #include <SDL_mixer.h>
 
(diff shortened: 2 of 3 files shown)


pgielda:
Ok, seems that CI is failing so I probably broke something. Not sure how can I check CI output without giving circleci extensive permissions to my github account?

AJenbo:
I don't believe it requires extensive access, but you could run the CircleCI build scripts locally if you do not trust them. https://github.com/diasurgical/devilutionX/blob/master/.circleci/config.yml

AJenbo:
SDL2
```
/root/repo/SourceX/sound.cpp: In function 'void dvl::snd_init(dvl::HWND)':
/root/repo/SourceX/sound.cpp:171:132: error: 'Mix_OpenAudioDevice' was not declared in this scope
  int result = Mix_OpenAudioDevice(22050, AUDIO_S16LSB, 2, 1024, SDL_GetAudioDeviceName(audio_device, 0), SDL_AUDIO_ALLOW_ANY_CHANGE);
```
SDL1.2
```
/root/repo/SourceX/sound.cpp: In function 'void dvl::snd_init(dvl::HWND)':
/root/repo/SourceX/sound.cpp:171:103: error: 'SDL_GetAudioDeviceName' was not declared in this scope
  int result = Mix_OpenAudioDevice(22050, AUDIO_S16LSB, 2, 1024, SDL_GetAudioDeviceName(audio_device, 0), SDL_AUDIO_ALLOW_ANY_CHANGE);
                                                                                                       ^
/root/repo/SourceX/sound.cpp:171:106: error: 'SDL_AUDIO_ALLOW_ANY_CHANGE' was not declared in this scope
  int result = Mix_OpenAudioDevice(22050, AUDIO_S16LSB, 2, 1024, SDL_GetAudioDeviceName(audio_device, 0), SDL_AUDIO_ALLOW_ANY_CHANGE);
                                                                                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~
/root/repo/SourceX/sound.cpp:171:132: error: 'Mix_OpenAudioDevice' was not declared in this scope
  int result = Mix_OpenAudioDevice(22050, AUDIO_S16LSB, 2, 1024, SDL_GetAudioDeviceName(audio_device, 0), SDL_AUDIO_ALLOW_ANY_CHANGE);
```

pgielda:
Ok, it might be harder on SDL1.2 indeed.

As for SDL_mixer, according to https://www.libsdl.org/projects/SDL_mixer/, it seems this function was added in 2.0.2.
CI seems to be using Debian stretch and it has  libsdl2-mixer-dev (2.0.1+dfsg1-1)
I was testing on Debian buster and it has libsdl2-mixer-dev (2.0.4+dfsg1-1) which seems to be enough.

I can do some ifdefs to test for the right version and fallback to the old method (where its hardcoded to device 0, maybe with a warning that it will play on device 0 if the user has a different setting in .ini) in case Mixer is too old -- if that is an acceptable solution.

AJenbo:
sure, have a look at how we generally have handled SDL versions

pgielda:
Its not lack of trust, more a problem with github always wanting too much access. All those repos are public so those CIs should also show it without being forced to log in and give access to public an *private* repos. Its circleci fault of course ;)


The tricky part here is that its not SDL2 version problem only (SDL has macros to check version so its easy) but also SDL_Mixer version (I am unware of such a mechanism for SDL_Mixer so I've added some code to CMakeLists).

I am not super happy with the result (too much copy-paste code) but its all I have time for now... 

Also something seems to fail on MacOS, without a clear error message. I have no easy way currently to test on a Mac :/
Is it possible to get more output for the MacOS fail somehow?

(Sorry for such a crappy contribution on my side ;) I think this project is great btw, keeps one of my favorite games of all time alive)

AJenbo:
@glebm do you know if this is something we could implement with SDL_audiolib?

StephenCWills:
SDL_audiolib accepts the device name as a string.

https://github.com/realnc/SDL_audiolib/blob/master/include/aulib.h#L59-L60

We can include it in the call to `Aulib::init()` here.

https://github.com/diasurgical/devilutionX/blob/1047e408bda059c10b38f90ceb9fef09dd477bb0/Source/sound.cpp#L204

AJenbo:
Replace by #4638

@pgielda sorry for the long review time on this, it unfortunately meant that things have diverged so much in the mean time that rebasing wasn't an option :(

pgielda:
That is completely ok @AJenbo. I really admire the energy in this project to push things forward and do understand that some things get obsolete because of that. Also originally my PR was failing on Mac (I only had a Linux machine around at that point) and I had no time to test it myself so it was more my fault than anyone else that it was not in a shape to be merged.

PR #922, opened 2020-11-19
commands for benchmarking

qndel:
* --nofpslimit - Disables FPS limit (use with vsync=0 in config.ini)
* --timer <#> - Shuts the game down after # seconds
* --skipmenu <#> - Skips intro and main menu and loads a hero named #

This will make benchmarking much easier as you can rerun the game with same parameters, not having to worry about going through menu etc.


glebm:
This is very useful indeed :heavy_check_mark:

PR #924, opened 2020-11-20
fix saves in x64-release

qndel:
Fixes saves in x64-Release in visual studio 2017 (and older?) - it worked properly in vs2019, hopefully this fix makes it work in the cases where it didn't and doesn't affect when it did xD


AJenbo:
Nice work debugging this.

PR #934, opened 2020-11-28
Controller: Fix L2/R2 handling

glebm:
Broken in https://github.com/diasurgical/devilutionX/commit/01401a16c70213db7a9a84196dee354a2ebc11a3

`ToControllerButtonEvent` must only be called once per SDL event because of how it handles L2/R2

### DIFF
--- a/SourceX/controls/controller.h
+++ b/SourceX/controls/controller.h
@@ -9,6 +9,8 @@ struct ControllerButtonEvent {
 	bool up;
 };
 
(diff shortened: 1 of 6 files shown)


SourceX/controls/controller.h
```diff
@@ -9,6 +9,8 @@ struct ControllerButtonEvent {
 	bool up;
 };
 
+// NOTE: Not idempotent because of how it handles axis triggers.
```
qndel, line 12: never heard this word before lol

PR #936, opened 2020-11-29
fix menu pics

qndel:
I had hfopt1.mpq which contained modified heros.pcx (an extra warrior frame with red "BARB" on it) so the total height of the pic was different and number of frames was hardcoded which resulted in bigger frames = ruined menu.
This hardcodes frame height to avoid such problems


SourceX/DiabloUI/diabloui.cpp
```diff
@@ -426,9 +426,9 @@ void LoadUiGFX()
 	LoadMaskedArt("ui_art\\focus42.pcx", &ArtFocus[FOCUS_BIG], 8);
 	LoadMaskedArt("ui_art\\cursor.pcx", &ArtCursor, 1, 0);
 #ifdef HELLFIRE
-	LoadArt("ui_art\\heros.pcx", &ArtHero, 6);
+	LoadArt("ui_art\\heros.pcx", &ArtHero, 6, NULL, 76);
 #else
-	LoadArt("ui_art\\heros.pcx", &ArtHero, 4);
+	LoadArt("ui_art\\heros.pcx", &ArtHero, 4, NULL, 76);
```
glebm, line 431: No need to hardcode here, right?

SourceX/DiabloUI/diabloui.cpp
```diff
@@ -426,9 +426,9 @@ void LoadUiGFX()
 	LoadMaskedArt("ui_art\\focus42.pcx", &ArtFocus[FOCUS_BIG], 8);
 	LoadMaskedArt("ui_art\\cursor.pcx", &ArtCursor, 1, 0);
 #ifdef HELLFIRE
-	LoadArt("ui_art\\heros.pcx", &ArtHero, 6);
+	LoadArt("ui_art\\heros.pcx", &ArtHero, 6, NULL, 76);
```
glebm, line 429: Please add a comment similar to that in PR description

glebm (changes requested):

SourceX/DiabloUI/diabloui.cpp
qndel: yeah, but what if someone creates Patch_rt.mpq that will modify heros.pcx for vanilla? :P Then we'll have the same issue :P

glebm (approved):

SourceX/DiabloUI/diabloui.cpp
glebm: OK, I'm not really sure how this is supposed to work
Wouldn't adding a hero require changes elsewhere in the code as well?

qndel: yeah but having the frame height hardcoded isn't harmful in any way atm so why not do it? :P we know how many frames we want, we know their height, so there's no need to calculate their height by dividing total height by the number of frames, which might come out wrong like in my case

SourceX/DiabloUI/art.cpp
```diff
@@ -3,7 +3,7 @@
 
 namespace dvl {
 
-void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette)
+void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette, int frame_height)
```
AJenbo, line 6: plz make a variant that takes frame height instead of frames, it's a bit odd to have it take both.

ferndog530 (approved):

SourceX/DiabloUI/art.cpp
qndel: what's wrong with current optional argument? I wanted to make the diff as small as possible ;p

AJenbo: Sometimes it's nicer to have it as clean as possible :)

I'm actually thinking that the best option is to split it into an array of single portraits since this will allow adding new ones without editing the others.

galaxyhaxz: I'm thinking the best thing is to use stick figures instead for portraits as it makes it more flexible when adding new classes.

AJenbo: image

PR #938, opened 2020-11-29
Preserve cursor position on vendors

### LINKED ISSUE #937: QOL change: Save Cursor position on vendors after purchasing
In vanilla, the selection cursor is always reset to the top position when shopping after any action is taken, including purchasing an item. This results in a very poor user experience when purchasing many of the same item (potions), or many items close to each other (e.g. scroll of identify + scroll of town portal).

This is especially troublesome for purchasing potions when the desired potion is not the topmost item in the vendor inventory. People tend to use a combination of mouse and keyboard for purchasing these potions in bulk, somewhat like this:
1. Click "Full Healing Potion"
2. Press enter to select
3. Press enter to confirm
4. Back to 1

This is usually repeated many times, sometimes as long as it takes to fill in the player inventory. Thus, even if the actions are still relatively fast, they become really slow compared to just holding the "Enter" key.

The proposal here is to persist the cursor position in the same vendor screen, so as to allow people to buy items more easily.

The persistence should only happen while purchasing a specific vendor section. For example:
1. Open griswold normal items
2. Scroll to bottom
3. Purchase item
4. Vendor screen is refreshed, and cursor stays at bottom
5. Back to main griswold menu
6. Select Premium items
7. Cursor should be at the top
8. Back to main griswold menu
9. Open normal items again
10. Cursor should also be at the top

The cursor should point "to the next item available" when the item is bought. If the cursor was pointing to the last item, then it should keep pointing to the new last item.

julealgon:
This PR adds a new behavior to all vendors so that the cursor position is preserved when interacting with them.

This affects the following actions:
- Buying items (Adria, Pepin and Griswold. Wirt is not affected since he only sells 1 item)
- Selling items (Adria and Griswold)
- Repairing items
- Recharging items

The implementation was done in a noninvasive manner, by manually scrolling the vendor sections back to the previously interacted with item index. This is a less than ideal solution but I feel other approaches would require significant refactoring at this point, as the vendor logic is extremely convoluted.

This should dramatically improve the user experience when purchasing and selling items from my experience.

Resolves #937

### DIFF
--- a/Source/stores.cpp
+++ b/Source/stores.cpp
@@ -1962,7 +1962,11 @@ void TakePlrsMoney(int cost)
 	}
 }
 
-void SmithBuyItem()
+/**
+ * @brief Purchases an item from the smith and returns the index into the vendors inventory where the item was purchased from.
+ * @return A zero-based index into the vendor's inventory where the purchased item was.
+ */
+int SmithBuyItem()
 {
 	int idx;
 
@@ -1971,6 +1975,7 @@ void SmithBuyItem()
 		plr[myplr].HoldItem._iIdentified = FALSE;
 	StoreAutoPlace();
 	idx = stextvhold + ((stextlhold - stextup) >> 2);
+	int purchasedItemIndex = idx;
 	if (idx == SMITH_ITEMS - 1) {
 		smithitem[SMITH_ITEMS - 1]._itype = ITYPE_NONE;
 	} else {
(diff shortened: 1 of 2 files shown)


AJenbo:
Indeed this should be done in the least invasive way possible at this point as we are still missing some parts of the hellfire for this section.

AJenbo:
StoreSellItem, SmithRepairItem, WitchRechargeItem and StoryIdItem will all be affected by the hellfire changes so we are currently blocked by that being done first:
https://github.com/diasurgical/devilution/issues?page=1&q=is%3Aissue+is%3Aopen

julealgon:
Rebased against latest master in case you want to get this one in.

AJenbo:
Merged manually via https://github.com/diasurgical/devilutionX/commit/5d2af51d01a31865e7384ba1fb0d24a2b2c55d26 since I figured out an even less invasive way to implement this.

Issue #941, opened 2020-12-09
Missing Walls


Trihedraf:
Walls that are supposed to be able to be semi transparent are just completely invisible

This happens on the current version when built from source. I tried just downloading 1.1.0 release and it did not do it. So I built 1.1.0 and then current with same libraries and everything and it only does it in the current version.

1.1.0 Source Built Version
![1 1 0](https://user-images.githubusercontent.com/26759970/101598055-6b29ed00-39ac-11eb-94d8-33e240ae4bbc.png)

Current Source Built Version
![Current](https://user-images.githubusercontent.com/26759970/101598065-711fce00-39ac-11eb-9d64-697dbad7b59c.png)

AJenbo:
Can you give some additional info about what system and compiler you are using, also please see if this commit is related:
https://github.com/diasurgical/devilutionX/commit/8e1c61a6f7a76bacedc7311ffce9e4b79c2aab8a

Trihedraf:
I am on Windows 10. Built with VS19 and MinGW.

I downloaded that git version and the one before and built both it is that commit that broke it. https://github.com/diasurgical/devilutionX/commit/8e1c61a6f7a76bacedc7311ffce9e4b79c2aab8a

qndel:
can u upload save?

Trihedraf:
[single_0.zip](https://github.com/diasurgical/devilutionX/files/5664305/single_0.zip)

AJenbo:
@madmann91 could you look in to this?

> I am on Windows 10. Built with VS19 and MinGW.
Are you using VS19 as an IDE and MinGW as the compiler, or are you saying that you used both compiler and had the same issue with both?

Trihedraf:
I have built it with both to make sure it wasn't a bug with one or the other. And I have tried it on multiple computers as well.

spidertoast:
I can reproduce this on Gentoo Linux when I apply the RenderLine improvements to 1.1.0. 
Running a HP mini 110 without graphics acceleration as there's no driver for the integrated SGX545 PowerVR graphics of the Atom N2800 and I thought this to be a unique case to this outdated hardware. 
@Trihedraf you can try to apply the RenderLine improvements to v1.1.0 via this patch and see if the issue reoccurs:
[...]

PR #942, opened 2020-12-09
`RenderLine()` function fix for issue #941

madmann91:
This should fix issue #941, and introduce compile-time and runtime checks for corner cases. The fallback for `count_leading_zeros` when the compiler is not gcc or clang has also been improved using binary search.

### DIFF
--- a/Source/render.cpp
+++ b/Source/render.cpp
@@ -179,7 +179,8 @@ inline static void RenderLine(BYTE **dst, BYTE **src, int n, BYTE *tbl, DWORD ma
 		// So we can limit it by ANDing the mask with another mask that only keeps
 		// iterations that are lower than n. We can now avoid testing if i < n
 		// at every loop iteration.
-		mask &= ((((DWORD)1) << n) - 1) << ((sizeof(DWORD) * CHAR_BIT) - n);
+		assert(n != 0 && n <= sizeof(DWORD) * CHAR_BIT);
+		k &= DWORD(-1) << ((sizeof(DWORD) * CHAR_BIT) - n);
 
 		if (light_table_index == lightmax) {
 			foreach_set_bit(mask, [=] (int i) { (*dst)[i] = 0; });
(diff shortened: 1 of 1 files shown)


madmann91:
The Travis CI build is not working for some unrelated reason on ppc64:
```ccache: error: /home/travis/.ccache/ccache.conf: No such file or directory```

**Comment by AJenbo** [comment]
Do you still see a performance improvement compared to the old code?

**Comment by madmann91** [comment]
The runtime checks are only enabled in Debug mode using `assert` (they should always be true anyway --- it's just here as documentation and sanity check). I have tried the fix on my machine, and the speed up seems to be comparable there for the saves I have tested (one in the cathedral, and one in Hell).

I have also thought about testing this in isolation, since the FPS counters fluctuate a lot and it's difficult to really assess how much of a speed-up there is. Hence, I have come up with this simple program to measure performance: https://gist.github.com/madmann91/a12d5c65bfb33573b0f8509d9042bec5. The results of this program on my machine are (compiled with `-march=native -O3`):

```
Old version: 8188/8296/8411us (Q1/Q2(=med)/Q3)
New version (fallback, with bug): 7217/7270/7351us (Q1/Q2(=med)/Q3)
New version (fallback, fixed): 7365/7404/7484us (Q1/Q2(=med)/Q3)
New version (builtin, with bug): 5012/5048/5092us (Q1/Q2(=med)/Q3)
New version (builtin, fixed): 5059/5086/5133us (Q1/Q2(=med)/Q3)
```

These numbers are much more accurate since they are the result of 100 runs of 100000 calls to RenderLine in different configurations, using random data for `src` and `dst`, and actual Diablo masks for `mask`. The results show the 1st, 2nd (a.k.a. the median), and third quartiles of the timing distribution, so that outliers are elimininated. The `fallback` case indicates when the compiler is not gcc or clang and `count_leading_zeros` default to a binary search. With that, you can see that the new version is consistently faster than the old one, even when using the fallback code for `count_leading_zeros`. The version with the bug fix is only a bit slower than the version without, so it seems the optimization still pays off, both when using the fallback, and when using the `__builtin_clz()` function.

AJenbo:
Thanks for fixing this and doing the benchmark

PR #945, opened 2020-12-16
PSP Vita v2

AJenbo:
A new and cleaner effort for porting the game to PSP Vita


SourceX/display.cpp
```diff
@@ -116,6 +121,12 @@ void CalculatePreferdWindowSize(int &width, int &height, bool useIntegerScaling)
 
 bool SpawnWindow(const char *lpWindowName)
 {
+
+#ifdef __vita__
+scePowerSetArmClockFrequency(444);
+#endif
+
+	SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
```
AJenbo, line 129:
```suggestion
#if !(SDL_VERSION_ATLEAST(2,0,6))
	SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
#endif
```

isage: Umm. Why `!` (not) ? 
Also, that hint first appeared in 2.0.10

isage:
Anything i can do to help get that merged? (move README back, i suppose)

SourceX/display.cpp
AJenbo: Oops, the `!` was a copy-paste error.

But the hint first appears in 2.0.6, https://github.com/spurious/SDL-mirror/blob/master/WhatsNew.txt#L329

What happens in 2.0.10 is that `SDL_HINT_ANDROID_SEPARATE_MOUSE_AND_TOUCH` is removed in favor of the then-existing `SDL_HINT_TOUCH_MOUSE_EVENTS`.

isage:
I've created separate branch and pull-request (#947) to allow easier merging

AJenbo:
Thanks

PR #950, opened 2020-12-26
Spend all stat points with shift click

julealgon:
This PR allows spending all remaining stat points in a single attribute by shift-clicking on the "+" button in the character screen. This makes it easier to assign stat points, especially since spending the attributes on a single stat on any given level up is a common scenario.

This functionality also exists like this in Diablo 2 and works the same way there.


julealgon:
Rebased.

PR #951, opened 2020-12-26
make cursor stay in the same position after buying

qndel:
effectively the same as https://github.com/diasurgical/devilutionX/pull/938 but less changes

### DIFF
--- a/Source/stores.cpp
+++ b/Source/stores.cpp
@@ -43,6 +43,7 @@ ItemStruct smithitem[SMITH_ITEMS];
 int stextdown;
 char stextscrlubtn;
 char stextflag;
+BOOL forceSelection = FALSE;
 
 /** Maps from towner IDs to NPC names. */
 const char *const talkname[9] = {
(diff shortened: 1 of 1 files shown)


Source/stores.cpp
```diff
@@ -2543,6 +2547,18 @@ void S_ConfirmEnter()
 			break;
 		}
 		StartStore(stextshold);
+		if (forceSelection) {
+			stextsel = stextlhold;
+			stextsval = stextvhold; //making the selection stay after buying an item
+			if (stextsval >= 1) {
+				// this trick is the best way to use existing mechanics and make sure the selection is still valid
```
glebm, line 2554: Let's not use the trick and ensure in-range selection directly?
Then we won't need the `if (!forceSelection)` SFX disabling either

qndel: are you saying I have any clue what's going on? :P

galaxyhaxz: Seeing are you say what am I saying and see??

AJenbo:
Thanks but this code is confusing enough already :D

Did this instead: https://github.com/diasurgical/devilutionX/commit/5d2af51d01a31865e7384ba1fb0d24a2b2c55d26

PR #952, opened 2020-12-27
Add key to stop walking and cancel pending actions

julealgon:
This PR introduces a new action that the player can now perform by pressing an assigned key (not set by default):
- Stops any walking after the current tile
- Cancels any queued up/buffered action, like attacking, talking, interacting with objects, etc

This allows for more fine grained control of the character without having to resort to things like attacking in-place or casting a skill to stop the current action/stop walking.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -2240,6 +2240,11 @@ void initKeymapActions()
 	    },
 	});
 #endif
+	keymapper.addAction({
+		"StopHero",
+		DVL_VK_INVALID,
+		[] { plr[myplr].Stop(); },
+	});
 }
 
 } // namespace devilution
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4078,7 +4078,7 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
 			if (p == myplr)
 				NewCursor(CURSOR_TELEPORT);
 		} else {
-			ClrPlrPath(p);
+			ClrPlrPath(&plr[p]);
 			plr[p]._pSpell = spl;
 			plr[p]._pSplType = RSPLTYPE_INVALID;
 			plr[p]._pSplFrom = 3;
@@ -4096,7 +4096,7 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
 			if (p == myplr)
 				NewCursor(CURSOR_TELEPORT);
 		} else {
-			ClrPlrPath(p);
+			ClrPlrPath(&plr[p]);
 			plr[p]._pSpell = spl;
 			plr[p]._pSplType = RSPLTYPE_INVALID;
 			plr[p]._pSplFrom = 3;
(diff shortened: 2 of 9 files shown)


qndel (changes requested):
I don't think any other function takes PlayerStruct as an argument so just pass player number and do plr[pnum] inside the function

AJenbo:
Hmm, would this be properly communicated in multiplayer?

julealgon:
@qndel 
> I don't think any other function takes PlayerStruct as an argument so just pass player number and do plr[pnum] inside the function

There are already functions in recent changes using C++ references. Check `spells.cpp` for a few examples. Using references like this leads to considerably cleaner code. The only reason they were not being used before is that most of the codebase is pure C and not C++.

julealgon:
@AJenbo 
> Hmm, would this be properly communicated in multiplayer?

Interesting question... I'm not 100% sure of that, but I suspect the messages are sent when the actions are executed. i.e. on each tile walk when `walkpath` is inspected, or on the final destination action when it runs. Thus, if I cancel those out early, there won't be any messages sent forward and it will effectively cancel it on MP as well. That's only an assumption though.

Do you have suggestions on how I could test this?

AJenbo:
Run two instances of the game in widescreen windowed, start and join a multiplayer game, have one player walk on a straight path across the screen where the other play can see, can press E to stop walking right at the start.

Notice if the other player sees the player stop or if it keeps walking to the original destination, or even one step further.

julealgon:
@AJenbo 
> Run two instances of the game in widescreen windowed

Nice. I didn't know you could run multiple instances of the game at once!

> ... start and join a multiplayer game...

As mentioned on discord, the instance connecting to the created game is crashing to desktop for me. If one of you could do a quick test of this it would be appreciated, otherwise I'd just have to wait it out until that is fixed as I have zero idea of what is happening there:

image

qndel:
works fine for me

julealgon:
Rebased and modified slightly to reuse an existing function to clear the player's walk path.

galaxyhaxz:
> Stops any walking after the current tile

$1 million challenge: make it stop on the current pixel exactly when you press it ;)

qndel (approved):
Looks good but I can't see the difference with it and without xD must be using it wrong

julealgon:
Rebased again to remove conflict.

qndel (approved):
Looks good 👍

julealgon:
Bump @AJenbo

julealgon:
Rebased again and updated the implementation slightly.

qndel:
I'm a huge fan of this PR but personally I'd go with changing as little as possible 🤷 why the sudden rewrite? :P

julealgon:
> I'd go with changing as little as possible 🤷 why the sudden rewrite? :P

I follow the "make the change easy then make the easy change" mantra when it comes to development, so i airways try to do some refactoring work whenever i make any changes.

They should always be in separate commits, however.

Why are you against making changes @qndel ? Now that we are free from the shackles of Hellfire, we can massively improve on the codebase.

julealgon:
Rebased.

julealgon:
Rebased.

julealgon:
Rebased again.

AJenbo:
please rebase and make it integrate with the keymapper (no default key for now)

julealgon:
> please rebase and make it integrate with the keymapper (no default key for now)

Done. Also updated PR title, commit title, and removed references in help since it does not have a standard key anymore.

AJenbo (approved):

PR #956, opened 2020-12-30
Always use desktop fullscreen mode

viciious:
Exclusive fullscreen mode is a thing of the past. Borderless windowed mode is the de-facto standard for games now because it allows quick switching between applications and doesn't mess up your screen resolution and/or sound.


AJenbo:
I don't see a reason to only support one set of systems, using desktop resolution is already the default.

julealgon:
Can we have this as an opt-in setting perhaps? It sounds promissing for those who can use it.

AJenbo:
This is how the game works when upscaling is enabled, so there is already a setting for it.

julealgon:
>This is how the game works when upscaling is enabled, so there is already a setting for it.

Ah gotcha. Thought there was more to it.

PR #959, opened 2020-12-31
Add --allquests debug switch for Singleplayer

Trihedraf:
This would add a switch --allquests which would enable all the quests to spawn in Singleplayer. Multiplayer would be unaffected.

### DIFF
--- a/Source/quests.cpp
+++ b/Source/quests.cpp
@@ -144,7 +144,7 @@ void InitQuests()
 		quests[z]._qmsg = questlist[z]._qdmsg;
 	}
 
-	if (gbMaxPlayers == 1, allquests == 0) {
+	if (gbMaxPlayers == 1 && allquests == 0) {
 		SetRndSeed(glSeedTbl[15]);
 		if (random_(0, 2) != 0)
 			quests[Q_PWATER]._qactive = QUEST_NOTAVAIL;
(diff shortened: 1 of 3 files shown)


Source/quests.cpp
```diff
@@ -144,7 +144,7 @@ void InitQuests()
 		quests[z]._qmsg = questlist[z]._qdmsg;
 	}
 
-	if (gbMaxPlayers == 1) {
+	if (gbMaxPlayers == 1, allquests == 0) {
```
qndel, line 147: what the fuck

Trihedraf: What is inside of that if is the code that removes one of the quests from each set of quests for the randomization. So if the switch is used it changes allquests to 1 and skips that whole section and all the quests are used.

qndel: yes, but who codes like this? xD use proper && / || instead of this voodo, + why would you make a 0/1 variable int?

Trihedraf: Sorry, not much of coder, more just hacking around with it. 
So it should (gbMaxPlayers == 1 || allquests == 0)?
and would it be a bool then?

AJenbo: I think wat qndel is suggesting is to make allquests a bool instead of it, and then check if it's false with `!allquests` instead of `allquests == 0`

qndel: yes + using TRUE/FALSE instead of 0/1

Trihedraf: ok implemented those changes

AJenbo (approved):
Thanks this is handy for testing

PR #962, opened 2021-01-02
Cancel targeted attack on dead monster

julealgon:
This PR introduces an optimization to improve the overall gameplay experience and responsiveness of characters while attacking. When a monster dies, any player that was currently targeting it for an attack as his next action will have that action cancelled.

This effectively prevents wasted shots when continuously attacking an enemy: the slower the action, the more pronounced this effect becomes.

Here is a comparison of how this works in practice:

| Standard | Attack cancel |
| --- | --- |
| Standard | BufferCancel |

Notice how in the standard version, each monster takes 2 hits: one to die, and another missed one after death. On the buffer cancelled version, the second hit does not occur since it is cancelled right on the frame where the monster dies, so it is easier to retarget to other monsters even while clicking constantly.

Clicking constantly on the standard version is guaranteed to always land at least 2 hits: the current action, and a buffered one.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1758,6 +1758,13 @@ void MonstStartKill(int i, int pnum, bool sendmsg)
 	M_FallenFear(monst->position.tile);
 	if ((monst->MType->mtype >= MT_NACID && monst->MType->mtype <= MT_XACID) || monst->MType->mtype == MT_SPIDLORD)
 		AddMissile(monst->position.tile, { 0, 0 }, 0, MIS_ACIDPUD, TARGET_PLAYERS, i, monst->_mint + 1, 0);
+
+	if (pnum >= 0 && pnum < MAX_PLRS) {
+		PlayerStruct &player = Players[pnum];
+		if ((player.destAction == ACTION_ATTACKMON || player.destAction == ACTION_RATTACKMON || player.destAction == ACTION_SPELLMON) && player.destParam1 == i) {
+			player.destAction = ACTION_NONE;
+		}
+	}
 }
 
 void M2MStartKill(int i, int mid)


julealgon:
Rebased.

julealgon:
Rebased

julealgon:
Rebased.

AJenbo:
I'm stalling a bit on this since it changes behavior. Was wondering if it made more sense to pare with the hold to attack feature instead.

julealgon:
> I'm stalling a bit on this since it changes behavior.

Fair enough.

> Was wondering if it made more sense to pare with the hold to attack feature instead.

I consider them completely independent. Actually, watching people play using a gamepad really makes me cringe and want this even more (if you watch that video from the Brazilian guy playing, you'll notice how many wasted swings it generates: basically, all hits).

sumyunggai:
From an outsider purist standpoint, this feel more like a mod because it changes the difficulty of the game, albeit indirectly. Your character will become more responsive, as you indicated @julealgon, which also means you are more efficient in combat.

From my standpoint, I would hate to see this rolled into the default implementation of hold LMB to attack, and kept as an option.

julealgon:
@sumyunggai 
> From an outsider purist standpoint, this feels more like a mod because it changes the difficulty of the game, albeit indirectly. Your character will become more responsive, as you indicated @julealgon, which also means you are more efficient in combat.

Are you implying this is significant enough to change game balance? If so, I strongly disagree. This is mostly inconsequential due to the fact that the original game is already incredibly easy. This change only makes it less annoying at the end of the day.

> From my standpoint, I would hate to see this rolled into the default implementation of hold LMB to attack

As I alluded to earlier, this has nothing to do with a feature to keep attacking on mouse hold and can be seen as an indendepent feature. Yes, it _also_ would impact the "hold to attack" feature, but they are not coupled together.

> ...and kept as an option

I think this would be completely unnecessary for the reasons above. I'd rather we keep such toggles to a minimum to make the code as clean as possible.

To me, it would be equivalent to having a toggle to enable "shift clicking to belt", since you also can do that in vanilla albeit much slower.

I'm of the opinion that not everything needs a toggle, even if it results in behavior that didn't exist in vanilla. It needs to be analyzed on a case-by-case basis and, in this case, IMHO it doesn't need one.

sumyunggai:
@julealgon 

> > From an outsider purist standpoint, this feels more like a mod because it changes the difficulty of the game, albeit indirectly. Your character will become more responsive, as you indicated @julealgon, which also means you are more efficient in combat.
> 
> Are you implying this is significant enough to change game balance? If so, I strongly disagree. This is mostly inconsequential due to the fact that the original game is already incredibly easy. This change only makes it less annoying at the end of the day.

No, I am not. But it will change how combat plays, and make it easier - especially when surrounded by enemies that are killed in few hits. If surrounded by 10 enemies that die to 1 hit each, vanilla would take 20 swings to eliminate them, with an associated loss of life commensurate with the time those swings take. You are reducing the time by 50% with this change. Same scenario, 2 hits to kill each enemy, vanilla would take 30 swings to eliminate, this change reduces it to 20; or a 33% reduction in time to take damage. This change is like equipping the player with a dynamic "of haste" perk.

(the above is, of course, theoretical. real impact would be lessened i imagine, due to misses, enemies taking more hits, etc).

but it also changes how combat feels. As someone who has played the warrior for about 23 years, melee combat has what feels like an inertial dampening factor to it; after reading your post, I think it's the behavior you undo with this PR. Regardless of difficulty, changing how combat feels should certainly be a factor for consideration when evaluating what is (1) merged into the main branch and (2) without the option to disable, short of source modification. 

Separately, Is the threshold for inclusion in the main branch "[does not] significantly [change the game] enough to change game balance"? I can't believe that's the case, as it would include a whole host of questionable and unwanted changes, more the realm of mod territory than the project's goals - as I understand them. 

> > From my standpoint, I would hate to see this rolled into the default implementation of hold LMB to attack
> 
> As I alluded to earlier, this has nothing to do with a feature to keep attacking on mouse hold and can be seen as an indendepent feature. Yes, it _also_ would impact the "hold to attack" feature, but they are not coupled together.

Thanks.

> > ...and kept as an option
> 
> I think this would be completely unnecessary for the reasons above. I'd rather we keep such toggles to a minimum to make the code as clean as possible.
> 
> To me, it would be equivalent to having a toggle to enable "shift clicking to belt", since you also can do that in vanilla albeit much slower.
> 
> I'm of the opinion that not everything needs a toggle, even if it results in behavior that didn't exist in vanilla. It needs to be analyzed on a case-by-case basis and, in this case, IMHO it doesn't need one.

It's definitely a value judgment, no doubt. But there is a time and a place for toggle-able features, and this seems like an obvious one. Similar to your logic you used in the auto-sort PR discussion, except with you holding the opposite position in this case.

Ultimately this is a community project, and you're clearly part of the community and driving how things are / are not implemented, etc. I'm just voicing my opinion as a long time D1 player who is new to this community and has an outsider's perspective. Perhaps you see value in that, perhaps not. 

FWIW, I would definitely play a warrior with this change to see how it feels. I'm advocating for the plutonic ideal version of Diablo in the vanilla, as-shipped devilutionX build. Something a first time player could stumble upon, fire up, and get a really good, authentic experience without the hassle of making 1997 code run in the 20s. This version would also have a robust set of options to enable, containing much requested community additions and modifications (LMB hold to attack, better automap, autosort inventory, improved item highlight, improved automap, stop attack on death frame, etc).

julealgon:
@sumyunggai 
> No, I am not. But it will change how combat plays, and make it easier

For sure, it will. This is undeniable. What I'm factoring in here is that, _even though it does make it easier_, it is _still_ inconsequential enough. You see, vanilla was already an incredibly easy game. This makes it a tiny bit easier. It doesn't really matter at the end of the day in terms of raw difficulty.

---

> especially when surrounded by enemies that are killed in few hits. If surrounded by 10 enemies that die to 1 hit each, vanilla would take 20 swings to eliminate them, with an associated loss of life commensurate with the time those swings take.

Not really, honestly. If you are in this situation, either you'll die anyway, or you'll live mostly unscathed. The way balance works in the original is very spiky. When you reach certain damage/armor thresholds, you become nearly invincible to mostly everything melee.

This change will just make the situation less _time consuming/frustrating_, but won't change much besides that. Also, keep in mind that most enemies will not normally take a single hit to be killed. They mostly do in the beginning, sure, but later on they take at the very least 2 to 3 on a normal, non-exploitative playthrough.

> You are reducing the time by 50% with this change. Same scenario, 2 hits to kill each enemy, vanilla would take 30 swings to eliminate, this change reduces it to 20; or a 33% reduction in time to take damage. This change is like equipping the player with a dynamic "of haste" perk.

Again I think you are going into the absolute worst possible scenarios with this. You are mostly never surrounded by this many monsters, and they mostly never take as few hits as this to die, _and_ you are mostly killing enemies in melee in chokepoints, facing a single direction (where this change has zero effect), _and_ most people overswing anyway like dumb motherfuckers (I've seen people overswing as much as 4 times a few times before changing directions). If anything, this change will impact axes the most which are completely and utterly underpowered anyways, _and only when surrounded by more than 2 or so enemies_ (which should not be a good situation to be in for a character without shield)

> (the above is, of course, theoretical. real impact would be lessened i imagine, due to misses, enemies taking more hits, etc).

And a miriad of other factors... I sincerly think you are overemphasizing the gameplay impact of this besides overall smoothness and cleaniness.

---

> but it also changes how combat feels. As someone who has played the warrior for about 23 years, melee combat has what feels like an inertial dampening factor to it; after reading your post, I think it's the behavior you undo with this PR. Regardless of difficulty, changing how combat feels should certainly be a factor for consideration when evaluating what is (1) merged into the main branch and (2) without the option to disable, short of source modification.

Everybody who knows me (in relation to this game) knows I'm a big fan of the warrior and melee gameplay. I've also been playing this game heavily since release (a few years after release... took a while for it to come out in Brazil IIRC). All I have to say is, please give it a try. The game feels _so much better_ with this in place, it's not even funny. I can now immediately feel the difference, both while playing as well as when watching others play. It feels like the game "makes more sense" suddenly. It really is a very small, but transformative change, for the better. This is why I don't think a toggle makes sense: it is just a net improvement to a very "hidden" mechanic that many people (the ones who don't pay a lot of attention while playing) don't even notice. The fact that I have to explain the situation to a lot of people when it comes to this behavior is enough reason for me to believe that a lot of people don't even realize the inneficiency.

Also, keep in mind that this will not prevent you from whiffing when holding shift. Holding shift will still cause you to "hit the ground" if the monster is killed and you click on it again after it is dead. Most people just mash click while holding shift, and the behavior will be less pronounced when they do that. Where this shines is when you "consciously attack", that is, you carefully click for every attack, queue them, and stop attacking as soon as a monster is dead. This won't magically prevent you from attacking the ground if you target the ground. If you hold shift and click behind enemies (a common thing), the game will _still_ queue your attack, since you did not target a monster. It only works when you are explicit about it: if you _specifically target a monster_ and the game can track down your queued attack is against that monster when it dies.

---

> Separately, Is the threshold for inclusion in the main branch "[does not] significantly [change the game] enough to change game balance"? I can't believe that's the case, as that would encompass a host of questionable and unwanted changes, more the realm of mod territory than the project's goals - as I understand them.

I think it is honestly impossible to precisely define the threshold. No, it's not only about the balance changes, but it's a contributing factor for sure. Any change that "drastically" (and this is, again, subjective) changes game balance will mostly automatically be rejected. "Questionable" changes are also rejected, and what is "questionable" is very, _very_ subjective. The goal of the project is "to stay true to the original vision" and _that_ is also subjective. There is no escaping subjectivity on any of this, imho.

---

> It's definitely a value judgment, no doubt. But there is a time and a place for toggle-able features, and this seems like an obvious one. 

As you can imagine, I disagree. I actually think it's obvious, but in the other direction: that it should _not_ be a toggle. It is a core mechanic that provides net improvements and has zero drawbacks, so there is no reason for a toggle behind it. I put this in the exact same category as "60 FPS cursor" and "shift+click to/from belt". Net, obvious improvements, that people will be unlikely to play without.

---

> Similar to your logic you used in the auto-sort PR discussion, except here you hold the opposite position.

The difference between that and this, to me, is how dramatically the change affects gameplay. Inventory management is a core element of the game, and automatically sorting everything goes against that quite a bit. On the other hand, this change here doesn't change any specific game mechanics, it just improves the responsiveness and fluidity of the game in some situations, it makes the game "make more sense": if I constantly attack an enemy, it is pretty obvious to me that, if the monster dies, I should stop trying to attack it. It is a logical change.

---

> Ultimately this is a community project, and you're clearly part of the community and driving how things are / are not implemented, etc. I'm just voicing my opinion as a long time D1 player who is new to this community and has an outsider's perspective. Perhaps you see value in that, perhaps not.

I see a lot of value in these discussions. I hope you do to.

I always try to clarify though: my opinion is my own and doesn't _necessarily_ represent those of DevilutionX as a project (sometimes they match, sometimes they don't). It is fairly normal for my opinion to clash with @AJenbo 's for example, and I think that's ok. As I said earlier, some of this stuff is just impossible to put in "guidelines" that are always 100% objective: it boils down to each persons' common sense. He is the maintainer however, and ultimately he makes the final decision. 

What I try to do is to steer the changes in the direction I think will be the most beneficial, not to me personally, but for the project as a whole, both in the structure and maintainability of the codebase, as well as the community interacting with the game. When I see some change that makes the game feel a lot better to play (such as this one), is easy to implement elegantly (such as this one) and has minimal to no impact on balance and game features (such as this one, imho), I'm tempted to try and push it into DevilutionX as much as possible because I want others to have the same improved experience as I do. If Ajenbo disagrees with something and doesn't want it merged, that's fair game and I reserve the modification for a mod I'll create later that will contain the more drastic changes I want to implement alongside any rejected change I propose here.

My point with the above is: everybody should do the same, if they are passionate enough. If you have a strong opinion about something and you want to argue for or against it, I'd not consider your opinion more or less based on who you are. I'll look into the raw arguments themselves and be convinced or not. I'd encourage you to keep pushing for what you believe is the best and, assuming everybody is doing the same from their point of view, we'll come to some conflicts sometimes, some aggreements in others, and some compromises in yet other times. I don't see any other "magical solution" for it.

> FWIW, I would definitely play a warrior with this change to see how it feels. 

Again, please do. Let me know how it feels once you do.

> I'm advocating for the platonic ideal version of Diablo in the vanilla, as-shipped devilutionX build. Something a first time player could stumble upon, fire up, and get a really good, authentic experience without the hassle of making 1997 code run in the 20s. 

I don't see it as a goal for people to have the "authentic" vanilla experience. It already is not authentic if you ask me after some improvements to framerate, widescreen, etc. As long as the game stays true to the spirit of the original, it's fair game to me. If people really want to play the original, they can play vanilla itself (it still runs just fine on Windows 10) or even the original devilution, but DevilutionX is _more_ than vanilla, even by default.

---

> This version would also have a robust set of options to enable, containing much requested community additions and modifications (LMB hold to attack, better automap, autosort inventory, improved item highlight, improved automap, stop attack on death frame, etc).

I think some options should never be exposed to the player. I think over-exposing options makes games worse sometimes and a lot more convoluted. This particular change here falls into the category of "not a setting" to me for those (and the above) reasons.

BTW, I trully appreciate the civil discussion, and the fact that you seem to care about this enough to discuss these finer points. The more people who care we have looking into this project, the best it will tend to be.

galaxyhaxz:
>  It already is not authentic if you ask me after some improvements to framerate, widescreen, etc. As long as the game stays true to the spirit of the original, it's fair game to me. If people really want to play the original, they can play vanilla itself (it still runs just fine on Windows 10) or even the original devilution, but DevilutionX is more than vanilla, even by default.

I agree with you @julealgon and for that matter vanilla Diablo runs slightly different on Windows 95 than modern windows/WINE. As Ajenbo and I confirmed the UI in particular runs at a different speed (and actually works/looks better on win 9x) but is buggy on modern windows.

The definition of pure vanilla is overrated IMO. Would you consider the 1998 PSX port of Diablo vanilla? They made many improvements to the game, including the double-attack nuance addressed in this PR. Is playing Diablo on a computer made after 2004 vanilla? Is playing it with a game controller which the PC version was never designed to do vanilla?

I think there are small changes to Diablo that make the game miles better and stick to the original design, and this is one of them. Hellfire on the other hand is a different beast, because that interfered far more with the story and game mechanics, but devilutionX supports both.

FitzRoyX:
I think that if a bug contributes to difficulty, that's artificial difficulty. D1 is now more moddable than ever and anyone can make real difficulty changes via real values like speed and stats. These forced swings make the combat clunkier and it hurts the warrior the most. You can somewhat avoid it by timing your clicks, so really the game logic is just being lazy here. This behavior was never going to survive the change to one-click attacking.

vlolteanu:
I think the benefits might not be as clear-cut when wielding a bow.

Suppose you're firing at a monster that is part of a pack. In this case, a cancelled shot would be a missed opportunity to hit someone behind your newly-deceased target.

vlolteanu:
> For sure, it will. This is undeniable. What I'm factoring in here is that, _even though it does make it easier_, it is _still_ inconsequential enough. You see, vanilla was already an incredibly easy game. This makes it a tiny bit easier. It doesn't really matter at the end of the day in terms of raw difficulty.

@julealgon I'm going to disagree with you here. If you're playing on hell difficulty without grinding a bit, getting lucky and/or knowing what gear to look for, you're going to find yourself having a tough time. Some monsters can two-shot you in the second half of the game.

julealgon:
> I think the benefits might not be as clear-cut when wielding a bow.
>
> Suppose you're firing at a monster that is part of a pack. In this case, a cancelled shot would be a missed opportunity to hit someone behind your newly-deceased target.

I suggest you play with it first @vlolteanu . This is not an issue for many reasons:
1. It would be very rare for a bow shot to be cancelled, since this behavior does _not_ cancel already-happening attacks, only queued ones. When firing with a bow, it would be _very_ likely that you have already started the next attack before the monster is hit by the missile and dies from it.

2. It only impacts the monster you are targeting, so if you are targeting the monster behind, your attack will not be cancelled

3. It does not influence shift clicking on the ground, so if you are attacking behind monsters, your attack is not cancelled when the monster in front of you dies. Similarly, if you are shift+clicking a monster, even though your queued attack _would_ be canceled, since you are still clicking, your next click will target the ground and the attack will still be re-queued.

> @julealgon I'm going to disagree with you here. If you're playing on hell difficulty without grinding a bit, getting lucky and/or knowing what gear to look for, you're going to find yourself having a tough time. Some monsters can two-shot you in the second half of the game.

We agree to disagree, then. To me, vanilla is still incredibly easy (in every difficulty) and the very (_very_) minor additional help from this feature here is inconsequential.

AJenbo (approved):

AJenbo (approved):
Adjusted to work properly with multiplayer

AJenbo:
I have finaly decided that I'm for this change. But after thinking a bit about it I think MonstStartKill() is a really strange place to do this and it would fail if multiple players are attacking the same monster. Instead it should probably be handled `PM_DoAttack`, `PM_DoRangeAttack` and `PM_DoSpell` (by returning `false`).

AJenbo (changes requested):
I have finaly decided that I'm for this change. But after thinking a bit about it I think MonstStartKill() is a really strange place to do this and it would fail if multiple players are attacking the same monster. Instead it should probably be handled `PM_DoAttack`, `PM_DoRangeAttack` and `PM_DoSpell` (by returning `false`).

julealgon:
> I think MonstStartKill() is a really strange place to do this

MonstStartKill is the method that effectively kills the monster. Why do you think it is a strange place to put logic that depends on the monster diying? I think this is by far the best spot to handle it, since it is the very source of the event the action depends upon. By moving it elsewhere, you'll just make it more complex to track down and link the 2 things.

> and it would fail if multiple players are attacking the same monster

That's true indeed, but your change doesn't address that. For this to work with all players, we'd need to loop over all players in the game and check each one, not use `pnum`. `pnum` is just the player that killed the monster, but every player could be targeting it and needs to have it's queue cleared.

AJenbo:
> > I think MonstStartKill() is a really strange place to do this
> 
> MonstStartKill is the method that effectively kills the monster. Why do you think it is a strange place to put logic that depends on the monster diying? I think this is by far the best spot to handle it, since it is the very source of the event the action depends upon. By moving it elsewhere, you'll just make it more complex to track down and link the 2 things.

Because it's part of the monster routine manipulating the player behavior.

> > and it would fail if multiple players are attacking the same monster
> 
> That's true indeed, but your change doesn't address that. For this to work with all players, we'd need to loop over all players in the game and check each one, not use `pnum`. `pnum` is just the player that killed the monster, but every player could be targeting it and needs to have it's queue cleared.

I know, that's why I canceled the merge and suggest that this is instead implemented in the player logic (check if target is still valid before executing a queued attack).

julealgon:
> Because it's part of the monster routine manipulating the player behavior.

Well yeah, because we don't have an event system. With events, you naturally decouple the source of the event from the actions that need to handle it. This is why I was advocating for reactive programming on discord at some point and @glebm mentioned being against it. IIRC Glebm actually suggested at the time to do pretty much what I did here: deal with the actions at the point of the event itself.

> I know, that's why I canceled the merge and suggest that this is instead implemented in the player logic (check if target is still valid before executing a queued attack).

Fair enough. In that case, if you think moving the logic away from the event is the way to go, feel free to do that change. I think it is inherently worse to do it that way so I don't have any intention of changing this myself. By shifting the logic to the attack, you will just delay something that can be resolved earlier in the pipeline (by cancelling the action).

I also feel like cancelling the queued attack, instead of cancelling the attack itself, is a lot more intuitive and easy to reason about, it feels more "direct" to me.

AJenbo:
> By shifting the logic to the attack, you will just delay something that can be resolved earlier in the pipeline (by cancelling the action).

The problem is that there are potentially ways you could schedule and attack after the monster has died.
- A attacks monster 5 and sends the message to B
- B kills the monster 5 and sends the death to A
- B receives the message about A attacking monster 5
- A receives the message and cancel the attack locally
- A makes the attack on player B's system

The game would then be out of sync and will stay so until action dies down and things align.

> I also feel like cancelling the queued attack, instead of cancelling the attack itself, is a lot more intuitive and easy to reason about, it feels more "direct" to me.

You can do that in `ProcessPlayers` before starting the action.

julealgon:
> - B kills the monster 5 and sends the death to A
> - B receives the message about A attacking monster 5

Shouldn't B just ignore this attack since monster 5 doesn't exist in his game? How is this handled today? It seems like a separate change that would make sense even without the change in this PR.

AJenbo:
`OnAttackMonster()` does not check if the target is dead and so will move to and attack it even if it was already killed.

PR #964, opened 2021-01-02
Engine: Functions to render to any buffer

glebm:
Refactor the rendering functions to allow rendering to any output buffer.

New functions that accept a buffer have the `To` suffix.

The functions that render to the global buffer now always accept coordinates.

/cc @viciious

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -5,6 +5,8 @@
  */
 #include "all.h"
 
+#include <cstddef>
+
 DEVILUTION_BEGIN_NAMESPACE
 
 BYTE sgbNextTalkSave;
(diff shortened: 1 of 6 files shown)


AJenbo:
First of all this is defiantly something we ant to do, with that said I do have some concerns.
- what is the expected performance impact of this?
- this touches a lot of files, did you make sure that it doesn't touch the about 40 functions in the 6 files that still have missing Hellfire code?
- What is your goal with this?

glebm:
> What is your goal with this?

In #955 (hwcursor), the author (@viciious) renders the cursor to the global buffer and then copies it from there. The code in that PR can be simpler if we render the cursor to a separate buffer to begin with.

> what is the expected performance impact of this?

I expect it to be neutral or very minimal. Perhaps worth measuring anyway.

> this touches a lot of files, did you make sure that it doesn't touch the about 40 functions in the 6 files that still have missing Hellfire code?

I did not. Only a few changes were needed elsewhere but perhaps this should wait until hellfire?

AJenbo:
I'm currently delaying looking at a lot of PRs because of the impending changes related to Hellfire, if you don't mind this also being added to the queue and then we look at it once Hellfire is implemented then that would be preferred from my side. Just don't take the silence as rejection :)

galaxyhaxz:
*If you do the hardest part first, then it makes the rest seem like a stroll on the moon*

AJenbo:
@glebm The codebase is ready for this now if you have time to rebase :)

glebm:
Rebased
No difference in FPS

PR #968, opened 2021-01-05
[WIP] Discord integration

heinermann:
Drafting for visibility.

### Goals
- Feature-rich Discord integration
- Be as unintrusive as possible to the rest of the codebase

### Done so far
- Discord rich presence, showing the following:
  - "In Menu" when in menus.
  - Line 1: Character name, level, class
  - Line 2: Dungeon type, level, difficulty, current/max players in multiplayer game
  - Line 3: Elapsed time in individual game instance
  - Character image asset showing the character, equipment, and weapons being wielded the same as in Diablo. (108/150 assets used)
  - Can send game invites to Discord

### TODO
- Discord lobby usage (create/join games and send invites to players and channels through Discord)
- ~~Improved Discord rich presence with graphics showing which dungeon type they are in~~
  - ~~Investigate possibility of showing their character as well in their Discord profile~~
  - ~~There is a 150 asset limitation.~~
- Discord lobby network code
- Fix "In Menu" presence not being activated (possibly because it isn't updated every frame or so)
- Indicator for whether they've beaten normal/nightmare/hell using the small image asset.

### Notes
- Location image was scrapped in favour of the character image. The small asset is too small to show the location.

### Screenshots
2021-01-06 18_36_43-Window

2021-01-06 18_38_47-Window

2021-01-06 18_46_28-Window


Source/drlg_l3.cpp
```diff
@@ -1970,7 +1970,7 @@ void FenceDoorFix()
 
 static void DRLG_L3Wood()
 {
-	int i, j, x, y, xx, yy, rt, rp, x1, y1, x2, y2;
+	volatile int i, j, x, y, xx, yy, rt, rp, x1, y1, x2, y2;
```
qndel, line 1973: that's not a part of discord features ;)

SourceX/discord/discord.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+const char *difficulties[] = { "Normal", "Nightmare", "Hell" };
+const char *DifficultyToStr(int difficulty)
+{
+	if (difficulty < 0 || difficulty > NUM_DIFFICULTIES)
+		return "Unknown";
+	return difficulties[difficulty];
+}
+
+const char *playerClasses[] = { "Warrior", "Rogue", "Sorcerer", "Monk", "Bard", "Barbarian" };
```
qndel, line 25: no need to redefine them and the function below is useless, just do 
```cpp
ClassStrTbl[plr[pnum]._pClass]
```
if taking name from player id or
```cpp
ClassStrTbl[classid]
```
if using class id directly

Source/drlg_l3.cpp
heinermann: Yeah sorry, can't compile without it, I'll remove it later.

qndel: :D god bless vs2019

galaxyhaxz:
Glad to see you around @heinermann :)

Source/drlg_l3.cpp
galaxyhaxz: Maybe, it's an Easter egg as dedicated to M_LazHelp

3rdParty/Discord/cpp/activity_manager.cpp
```diff
@@ -0,0 +1,177 @@
...
+    std::unique_ptr<std::function<void(Result)>> cb{};
+    cb.reset(new std::function<void(Result)>(std::move(callback)));
+    internal_->clear_activity(internal_, cb.release(), wrapper);
+}
+
+void ActivityManager::SendRequestReply(UserId userId,
+                                       ActivityJoinRequestReply reply,
+                                       std::function<void(Result)> callback)
+{
+    static auto wrapper = [](void* callbackData, EDiscordResult result) -> void {
```
glebm, line 122: `static constexpr auto wrapper` throughout

3rdParty/Discord/cpp/activity_manager.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "core.h"
+
+#include <cstring>
+#include <memory>
+
+namespace discord {
+
+class ActivityEvents final {
+public:
+    static void OnActivityJoin(void* callbackData, char const* secret)
```
glebm, line 16: We use snake case for parameters, local variables, and fields
i.e. `callback_data`

3rdParty/Discord/cpp/activity_manager.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    static void OnActivityJoinRequest(void* callbackData, DiscordUser* user)
+    {
+        auto* core = reinterpret_cast<Core*>(callbackData);
+        if (!core) {
+            return;
+        }
+
+        auto& module = core->ActivityManager();
+        module.OnActivityJoinRequest(*reinterpret_cast<User const*>(user));
```
glebm, line 46: Please add a comment explaining why this safe or use a safer cast (`static_cast`?)

3rdParty/Discord/cpp/application_manager.cpp
```diff
@@ -0,0 +1,78 @@
...
+    internal_->get_current_locale(internal_, reinterpret_cast<DiscordLocale*>(locale));
+}
+
+void ApplicationManager::GetCurrentBranch(char branch[4096])
+{
+    if (!branch) {
+        return;
+    }
+
+    internal_->get_current_branch(internal_, reinterpret_cast<DiscordBranch*>(branch));
```
glebm, line 44: Why is this safe? From the signature it seems that `branch` is char-aligned but what are `DiscordBranch` alignment requirements?

glebm (changes requested):

glebm:
Is the code in `3rdparty/Discord` from another repo?
If so, you can disregard my code review comments but please add that repo as a submodule instead of copying the code

glebm:
@heinermann Please use `ExternalProject_Add` to download https://dl-game-sdk.discordapp.net/2.5.6/discord_game_sdk.zip

There is an example of using a remote zip file here: https://cmake.org/cmake/help/latest/module/ExternalProject.html#examples

heinermann:
> @heinermann Please use CMake's `ExternalProject_Add` to download https://dl-game-sdk.discordapp.net/2.5.6/discord_game_sdk.zip
> 
> There is an example of using a remote zip file here: https://cmake.org/cmake/help/latest/module/ExternalProject.html#examples

Will this work even when the project in the zip is not a CMake project, only raw source files?

AJenbo:
Looks like this is updating each frame, mabe better to call the update on level load and level up.

Does it not work for Linux?

heinermann:
> Looks like this is updating each frame, mabe better to call the update on level load and level up.

Per the Discord guide:

> Make sure to call `discord.RunCallbacks()` in your main game loop; that's your `Update()` function.



> Does it not work for Linux?

I just haven't set up Linux because I can't test it right now, I'd need to set up a VM with Linux etc.

glebm:
@heinermann

> Will this work even when the project in the zip is not a CMake project, only raw source files?

Yes, you can specify exactly how it should be built via `CONFIGURE_COMMAND` and `BUILD_COMMAND`.

Alternatively, `FetchContent` may be easier to use in this case:

https://cmake.org/cmake/help/latest/module/FetchContent.html

> EDIT: It also appears that feature is for a later version of CMake

Why do you think that? I can see it in the 3.10 documentation:

https://cmake.org/cmake/help/v3.10/module/ExternalProject.html#examples

`FetchContent` is 3.11. It's OK to bump the minimum version to 3.11 if `FetchContent` is more appropriate here.
https://cmake.org/cmake/help/v3.11/module/FetchContent.html

glebm:
> 3rdParty/Discord/lib/x86/discord_game_sdk.dll

Can this library be built from source or is it a proprietary blob?

heinermann:
> > 3rdParty/Discord/lib/x86/discord_game_sdk.dll
> 
> Can this library be built from source or is it a proprietary blob?

Pretty sure it's proprietary.

> Why do you think that? I can see it in the 3.10 documentation

I made an assumption based on
```
1> [CMake] CMake Error at C:\Users\Adam Heinermann\diablo\devilutionX\CMakeLists.txt:191 (ExternalProject_Add):
1> [CMake]   Unknown CMake command "ExternalProject_Add".
```

Version shows `CMake 3.17.20032601-MSVC_2` so no idea...

glebm:
Add `include(ExternalProject)` for `ExternalProject_Add`, `include(FetchContent)` for `FetchContent_Declare`

CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
-cmake_minimum_required(VERSION 3.10) # CMP0083 NEW
+cmake_minimum_required(VERSION 3.11) # CMP0083 NEW
```
AJenbo, line 1: Could we limit this to only the DISCORD_INTEGRATION clause, I belive both android studio and visual studio 2017 are limited to 3.10

glebm: From what I can tell:

* Visual Studio 2017 was already at CMake 3.12 in 2018
* Android Studio is at CMake 3.18.1 as of version 4.2

CMakeLists.txt
```diff
@@ -186,6 +188,36 @@ add_library(PKWare STATIC
...
+    SOURCE_DIR 3rdParty/Discord
+  )
+
+  FetchContent_Populate(discordsrc)
+
+  if(NOT discordsrc_POPULATED)
+    message(FATAL_ERROR "Failed to retrieve Discord SDK.")
+  endif()
+
+  add_library(discord STATIC
```
glebm, line 204: Let's glob the sources to make updating easier:

```cmake
file(GLOB discord_SRCS ${discordsrc_SOURCE_DIR}/cpp/*.cpp)
add_library(discord STATIC ${discord_SRCS})
```

SourceX/discord/discord.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "../Source/player.h"
+
+namespace dvl {
+
+DiscordManager::DiscordManager()
+{
+	discord::Core::Create(DISCORD_DEVIX_APPID, DiscordCreateFlags_NoRequireDiscord, &core);
+}
+
+const char *difficulties[] = { "Normal", "Nightmare", "Hell" };
```
glebm, line 18: `difficulties`, `DifficultyToStr` etc should either be marked `static` or placed in an anonymous namespace

SourceX/discord/discord.h
```diff
@@ -0,0 +1,49 @@
...
+private:
+	std::string getPlayerInfoString();
+	std::string getLocationString();
+	std::string getDetailString();
+	std::string getStateString();
+
+private:
+	discord::Core *core = nullptr;
+
+	int numPlayers = 0;
```
glebm, line 34: Please name private member fields like this (Google style):

```c++
int num_players_ = 0;
int max_players_ = 0;
...
```

SourceX/discord/discord.h
```diff
@@ -0,0 +1,49 @@
...
+	int maxPlayers = 0;
+	int playerClass = 0;
+	int playerLevel = 0;
+	int difficulty = 0;
+	int dungeonType = 0;
+	int dungeonLevel = 0;
+
+	std::string characterName = "";
+
+	int64_t startTime = 0;
```
glebm, line 44: `std::int64_t start_time_ = 0;` and `#include <cstdint>`

SourceX/discord/discord.h
```diff
@@ -0,0 +1,49 @@
...
+
+	int numPlayers = 0;
+	int maxPlayers = 0;
+	int playerClass = 0;
+	int playerLevel = 0;
+	int difficulty = 0;
+	int dungeonType = 0;
+	int dungeonLevel = 0;
+
+	std::string characterName = "";
```
glebm, line 42: No need for `= ""`, `std::string` is a class type so it gets initialized with its default constructor.

SourceX/discord/discord.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#ifdef DISCORD
+
+#include <string>
+
+#define DISCORD_DIABLO_APPID 496571953147150354
```
glebm, line 7: Unused

SourceX/discord/discord.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#ifdef DISCORD
+
+#include <string>
+
+#define DISCORD_DIABLO_APPID 496571953147150354
+#define DISCORD_DEVIX_APPID 795760213524742205
```
glebm, line 8: Move to the cpp file?

glebm (changes requested):
Please run clang-format on the new files

SourceX/discord/discord.cpp
AJenbo: Isn't there a global array with tease strings already?

heinermann: They are hardcoded inside of various functions.

SourceX/discord/discord.h
heinermann: Where are these styles outlined for this project?

SourceX/discord/discord.h
heinermann: I'd rather keep it because it is Discord's id for Diablo, in case it can be used later.

SourceX/discord/discord.cpp
heinermann: Just noticed that Bard/Barbarian are called Rogue/Warrier in this array so I'll keep the one I had.

SourceX/discord/discord.h
glebm: Nowhere yet, there is no consistent style guide or a definitive clang-format

Some other classes are written this way, e.g. https://github.com/diasurgical/devilutionX/blob/master/SourceX/controls/devices/game_controller.h

CMakeLists.txt
```diff
@@ -387,6 +428,16 @@ target_include_directories(${BIN_TARGET} PRIVATE
   3rdParty/Radon/Radon/include
   3rdParty/libsmacker)
 
+if(DISCORD_INTEGRATION AND WIN32)
+  target_include_directories(${BIN_TARGET} PRIVATE "${discordsrc_SOURCE_DIR}/..")
+
+  set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86_64")
```
glebm, line 434: `${CMAKE_HOST_SYSTEM_PROCESSOR}` instead of `x86_64`, so that it works for the 32-bit BUILD as well

CMakeLists.txt
```diff
@@ -186,6 +188,36 @@ add_library(PKWare STATIC
 add_library(SourceS STATIC
   SourceS/paths.cpp)
 
+if(DISCORD_INTEGRATION AND WIN32)
```
glebm, line 191: Let's test in a separate `if` that the platform is supported and error out explicitly if it isn't. Then we can remove `AND WIN32` throughout

CMakeLists.txt
```diff
@@ -387,6 +428,16 @@ target_include_directories(${BIN_TARGET} PRIVATE
   3rdParty/Radon/Radon/include
   3rdParty/libsmacker)
 
+if(DISCORD_INTEGRATION AND WIN32)
+  target_include_directories(${BIN_TARGET} PRIVATE "${discordsrc_SOURCE_DIR}/..")
+
+  set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86_64")
+
+  target_link_libraries(discord PRIVATE "${DISCORD_LIB_DIR}/discord_game_sdk.dll.lib")
```
glebm, line 436: WIN32-only

CMakeLists.txt
```diff
@@ -387,6 +428,16 @@ target_include_directories(${BIN_TARGET} PRIVATE
...
   3rdParty/libsmacker)
 
+if(DISCORD_INTEGRATION AND WIN32)
+  target_include_directories(${BIN_TARGET} PRIVATE "${discordsrc_SOURCE_DIR}/..")
+
+  set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86_64")
+
+  target_link_libraries(discord PRIVATE "${DISCORD_LIB_DIR}/discord_game_sdk.dll.lib")
+  target_link_libraries(${BIN_TARGET} PRIVATE discord)
+  file(COPY "${DISCORD_LIB_DIR}/discord_game_sdk.dll" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
```
glebm, line 438: Replace `dll` with `${CMAKE_SHARED_LIBRARY_SUFFIX}` so that this also works for Mac and Linux

glebm (changes requested):

SourceX/discord/discord.h
heinermann: I'd like to hear from the other devs what style you guys would prefer as a group as I am personally not a fan of prefix/postfix underscores. Classes in [here](https://github.com/diasurgical/devilutionX/tree/master/SourceX/dvlnet) have a different style. @AJenbo @qndel 

I'll at least switch the variables to snake case naming which is more consistent between both examples.

CMakeLists.txt
heinermann: `${CMAKE_HOST_SYSTEM_PROCESSOR}` prints AMD64 on both x86 and x64 configurations in VS. Will fix CMake for all platforms eventually since this is still a WIP.

glebm: Oh, it's a configuration in VS which means we probably have to use a generator expression

I've found this answer from 2017 but hopefully there is a better solution these days https://cmake.org/pipermail/cmake-developers/2017-November/030451.html

heinermann:
Both 32 and 64 bit builds work on Windows (through VS).

Linking fails on Linux because it can't find `discord_game_sdk` and I don't know why.

```
/usr/bin/ld: cannot find -ldiscord_game_sdk
make[2]: *** [CMakeFiles/devilutionx.dir/build.make:832: devilutionx] Error 1
make[1]: *** [CMakeFiles/Makefile2:285: CMakeFiles/devilutionx.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
```

CMakeLists.txt
```diff
@@ -425,8 +425,14 @@ if(DISCORD_INTEGRATION)
     set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86_64")
   endif()
 
-  target_link_libraries(discord PRIVATE "${DISCORD_LIB_DIR}/discord_game_sdk${CMAKE_SHARED_LIBRARY_SUFFIX}${CMAKE_LINK_LIBRARY_SUFFIX}")
-  target_link_libraries(${BIN_TARGET} PRIVATE discord)
+  find_library(DISCORD_LIB
+    NAMES discord_game_sdk${CMAKE_SHARED_LIBRARY_SUFFIX}
+    HINTS ${DISCORD_LIB_DIR})
```
heinermann, line 430: This can use the short form, will update it

qndel:
Nice progress on this :D so clean <3

heinermann:
Will revisit and repurpose this as "Discord presence integration" (i.e. no game invites/ask to join), so that the project can take advantage of what was done here. Game invites can then be a separate PR. Also need to give ownership of the Discord game control panel (where all the graphics and stuff are). Further Discord integration can be developed later.

AJenbo:
Sounds like a good plan :)

heinermann:
Yeah but I trailed off and gave up on it shortly after. I threw up whatever WIP I had on my system to [this branch](https://github.com/heinermann/devilutionX/commits/discord-multiplayer).

heinermann:
Superseded by https://github.com/diasurgical/devilutionX/pull/3711 . This one can be referred to when looking how to get invites working.

PR #973, opened 2021-01-10
fix vs2019 release builds

qndel:
good job microsoft!

### DIFF
--- a/Source/drlg_l3.cpp
+++ b/Source/drlg_l3.cpp
@@ -1971,7 +1971,8 @@ void FenceDoorFix()
 static void DRLG_L3Wood()
 {
 #if (_MSC_VER >= 1920) 
-	volatile // visual studio 2019 throws internal compiler error without it
+	volatile // visual studio 2019 throws internal compiler error without it, see #708
+
 #endif
 	int i, j, x, y, xx, yy, rt, rp, x1, y1, x2, y2;
 	BOOL skip;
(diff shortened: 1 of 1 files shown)


AJenbo:
Nice, but is it needed for all the variables?

qndel:
no but it won't be a nice 1liner otherwise :P

Source/drlg_l3.cpp
```diff
@@ -1970,6 +1970,9 @@ void FenceDoorFix()
 
 static void DRLG_L3Wood()
 {
+#if (_MSC_VER >= 1920) 
+	volatile // visual studio 2019 throws internal compiler error without it
```
AJenbo, line 1974:
```suggestion
	volatile // visual studio 2019 throws internal compiler error without it, see #708

```

qndel:
```cpp
#if (_MSC_VER >= 1920) 
	volatile int j; // visual studio 2019 throws internal compiler error without it
#else
	int j;
#endif
	int i, x, y, xx, yy, rt, rp, x1, y1, x2, y2;
	```

**Comment by AJenbo** [comment]
Could also do:
```cpp
#if (_MSC_VER >= 1920) 
	volatile // visual studio 2019 throws internal compiler error without it, see #708
#endif
	int j;
	int i, x, y, xx, yy, rt, rp, x1, y1, x2, y2;
```
But on second thought let's not rearrange the code just to cater to a buggy compiler.

Issue #894, opened 2020-10-23
Lazarus Staff Skip


NiteKat:
**Describe**
On some game seeds, you can skip finding Lazarus' staff and taking it to Cain. The game puts the trigger tile for going to Lazarus' set level a certain distance away from where it is supposed to appear when the player has activated it. If that position has an x and y value high enough, it's possible the "idle" position for this trigger will land in a section of the map where the player can path to. If this happens, and the player stands still on the tile, they will be taken to Lazarus' set level without having to take find and take the staff to Cain. This is single player only. I only scanned 1,470 seeds, and 13.26% of those scanned have this bug, so I think it happens more often than one would think it does.

**To Reproduce**
1. Start the game with a seed where this bug exists.
2. Work your way to Level 15.
3. Find the tile where the portal is supposed to spawn.
4. Subtract 16 from the x and y, then divide by 2 to find where the trigger is sitting prior to activating it.
5. Walk the character onto that calculated position, and stand still. You will be brought to Lazarus' set level.

**Expected behavior**
A player should not be able to access Lazarus' set level until the staff is found and brought to Cain on all seeds.

**Additional context**
This code here is where the trigger tile is moved when it is supposed to be active: https://github.com/diasurgical/devilutionX/blob/master/Source/quests.cpp#L202 . I think the idle tile gets set somewhere in the L4 dungeon generation code. This could probably be changed so that the tile is always set to (0, 0) when not active, and then the game can scan for the tile type it's supposed to spawn on/near to move it there (I forget if it's the tile it spawns on that is a unique tile ID, or the tile below it). There are probably other ways to solve this problem, but that's the first that jumps to mind for me.

NiteKat:
Forgot to mention, this is a vanilla bug, so it should probably have the vanilla tag (I don't think i can add those)

AJenbo:
This is a hilarious bug, a nice find :D It's particularly funny since in most cases the game will check the quest status to prevent things like that, and doing so would have prevented this issue without needing a temp position in the first place :)
Good find.

NiteKat:
[...]

PR #978, opened 2021-01-17
Fix lazarus staff skip bug

galaxyhaxz:
Sometimes on level 15 the player can teleport to lazarus' lair by walking on a "random" tile because that tile hasn't been converted to a mini tile yet. This ports the fix from the PSX version of Diablo. Since the dungeon is generated each time, it only sets the coordinates when the portal is active.

ref: https://github.com/diabpsx/skeleton/blob/master/JAP_1998_05_29/DIABPSX/SOURCE/DRLG_L4.CPP#L2683

Bug found by @NiteKat

### DIFF
--- a/Source/drlg_l4.cpp
+++ b/Source/drlg_l4.cpp
@@ -1383,7 +1383,7 @@ static BOOL DRLG_L4PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx,
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Thanks for the PR :)

galaxyhaxz:
Forgot to mention, #894 can be closed now. I'm not particularly proud of sending 1 liner PR's but hey :P

PR #982, opened 2021-01-21
Adding 'w' key to switch between weapons. 

joewis:
This pull requests adds a shortcut 'w' to switch between weapon sets. 

Items are moved to the existing inventory, to avoid changes to the save file format. If there is insufficient space in the inventory, the switch stops with the switching item in HoldItem.

User Stories:
1. Warrior Arrg likes his Shadowhawk sword for the life leech, but he also likes to cast Chain Lightning, and keeps running out of Mana. He found a sword with Vampire suffix but his player finds it too cumbersome to keep changing the sword in the midst of things. The shield can either remain or switch with a different shield.
2. Sometimes Arrg likes to play with a bow but still needs the option to quickly pull out his sword and shield, to not die.  
3. Sorcerer Claudus likes to switch between staves so he can alternate between Chain Lightning and Fireball without using his Mana resources.


galaxyhaxz:
I love this idea, as having dual weapons adds a new level to gameplay tactics and makes many more builds open up. If Ajenbo is fine with breaking compatibility, you may greatly simplify your implementation by having proper slots like Tchernobog does. (:

joewis:
Haha yes I tried that implementation first and realized I would not be able to keep my current characters.
Also it's still kind of vanilla, you could always do this manually.

Source/help.cpp
```diff
@@ -384,13 +384,19 @@ const char gszHelpText[] = {
...
 	"+ / -: Zoom Automap|"
 	"1 - 8: Use Belt item|"
+<<<<<<< HEAD
+	"F5, F6, F7, F8:     Set hot key for skill or spell|"
+	"Shift + Left Click: Attack without moving|"
+=======
 	"F5, F6, F7, F8:     Set hotkey for skill or spell|"
 	"Shift + Left Click: Attack without moving|"
 	"Shift + Left Click (on character screen): Assign all stat points|"
+>>>>>>> f49b55cbe1e56744faf616cdde64b3a8202b1024
```
AJenbo, line 399: Conflict sad :(

joewis: No idea what just happened... it's definitely a git user error on my side though. 
I'll take it as a learning experience but not sure I can recover this... May have to start with a fresh branch.

AJenbo: Just edit the file, correct the duplicate lines, commit and put.

joewis: Thanks - Done that and it seems to build. Let's see if it runs :)

PR #983, opened 2021-01-22
Optional feature to auto-refill belt from inventory

joewis:
Functionality to "auto-refill" the belt from inventory when using a belt item. 
The code looks for an identical potion / scroll (oil / rune / etc.) in the inventory and uses that item instead of the belt item.

The feature should be optional and is default off. I would love to use the config.ini that is planned for "All the cool QOL stuff" but for now I have added  hotkey Shift+A to switch it on and off again.

I play on a laptop without mouse, and have the strong impression this is not what the original developers had in mind. Right clicking potions in the inventory is not really a feasible option.
This feature allows to utilize more slots for scrolls without losing access to quick Life / Mana refill.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -2508,6 +2508,16 @@ BOOL UseInvItem(int pnum, int cii)
 		speedlist = TRUE;
 	}
 
+	//If selected speedlist item exists in InvList, use the InvList item.
+	for (int i = 0; i < plr[pnum]._pNumInv && plr[pnum].autoRefillBelt; i++) {
+		if (plr[myplr].InvList[i]._iMiscId == Item->_iMiscId && plr[myplr].InvList[i]._iSpell ==Item->_iSpell){
+            c = i;
+            Item = &plr[pnum].InvList[c];
+            speedlist = FALSE;
+            break;
+        }
+	}
+
 	switch (Item->IDidx) {
 	case IDI_MUSHROOM:
 		sfxdelay = 10;
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1295,6 +1295,14 @@ static void PressChar(WPARAM vkey)
 		}
 		return;
 #endif
+	case 'A':
+        plr[myplr].autoRefillBelt = !plr[myplr].autoRefillBelt;
+        if(plr[myplr].autoRefillBelt)
+            NetSendCmdString(1 << myplr, "Auto-refill belt activated");
+        else
+            NetSendCmdString(1 << myplr, "Auto-refill belt de-activated");
+        PlaySFX(IS_IGRAB);
+        return;
 	}
 }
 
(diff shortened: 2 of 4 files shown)


galaxyhaxz:
Good idea, this is exactly how it works in the game Darkstone (often called a D1 ripoff). Might I also suggest code to draw a number in the item slot for how many are left in the inventory, like Darkstone?

Here's a screenshot. You can see how many of each scroll/potion are left:
darkstone

joewis:
The remaining potion count would be good but I haven't at all looked into the UI code yet.
Maybe we take the shock as part of the fun when suddenly a potion was your last potion?

galaxyhaxz:
<details><summary>Is that so?</summary>

image

</details>

julealgon:
> but for now I have added hotkey Shift+A to switch it on and off again.

Not a fan of this. Why not add a flag to ini like other toggleable features?

> This feature allows to utilize more slots for scrolls without losing access to quick Life / Mana refill.

To me, this just makes the belt much stronger. I feel like this is MOD territory,

joewis:
> > but for now I have added hotkey Shift+A to switch it on and off again.
> 
> Not a fan of this. Why not add a flag to ini like other toggleable features?
> 
> > This feature allows to utilize more slots for scrolls without losing access to quick Life / Mana refill.
> 
> To me, this just makes the belt much stronger. I feel like this is MOD territory,

That's what I did in today's commit - removed the Shift+A shortcut and made it a toggle feature. At least that was my intention.

ThomasChr:
@joewis Do you want to clean up the conflicts? After that I think this can be merged...

joewis:
I'll have a look - not sure how much has changed in the meantime

Source/options.h
```diff
@@ -89,8 +89,9 @@ struct GameplayOptions {
 	bool bAutoEquipJewelry;
 	/** @brief Only enable 2/3 quests in each game session */
 	bool bRandomizeQuests;
-	/** @brief Indicates whether or not monster type (Animal, Demon, Undead) is shown along with other monster information. */
```
qndel, line 92: where did this go?

Source/inv.cpp
```diff
@@ -483,15 +483,15 @@ void DrawInv(CelOutputBuffer out)
 					    InvRect[j + SLOTXY_INV_FIRST].X + RIGHT_PANEL_X,
 					    InvRect[j + SLOTXY_INV_FIRST].Y - 1,
 					    pCursCels, frame, frame_width,
-						false);
```
qndel, line 486: please avoid causing differences unrelated to the PR, a clang format commit can come later

joewis: Yes..  I did this before figuring out that I can apply clang-format only to the changes of the PR. Applying to the whole file send to move more than intended.

joewis: Yes..  I did this before figuring out that I can apply clang-format only to the changes of the PR. Applying to the whole file send to move more than intended.

AJenbo: We are working on solving that problem today :)

PR #985, opened 2021-01-23
Dynamic speed scrolling text

galaxyhaxz:
Given my increased laziness and the likely huge delay of anything, I've decided to go ahead and share some of my work here.

This feature is ported from the PSX. It calculates how fast to scroll text in the quest dialogs based on the number of lines and how long the speech is. It will always end up with exactly one line left when the speaker finishes.

It is much smoother than vanilla which only used integers and as a result some text finishes almost too quickly whereas other speeches are finished long before the text.

The big advantage here is supporting more languages, so they didn't have to fine tune pixel counts for 5 languages each with 250 speeches.

### DIFF
--- a/Source/effects.cpp
+++ b/Source/effects.cpp
@@ -1410,4 +1410,24 @@ void effects_play_sound(const char *snd_file)
 	}
 }
 
+DWORD GetSoundFrames(int n)
+{
+	HANDLE hFile;
+	BYTE temp[64];
+
+	WOpenFile(sgSFX[n].pszName, &hFile, FALSE);
+	WReadFile(hFile, temp, sizeof(temp), sgSFX[n].pszName);
+	WCloseFile(hFile);
+
+	/* lazy hack: read the size of the wave file */
+	PCMWAVEFORMAT *pcm = (PCMWAVEFORMAT *)((BYTE *)&temp[0x14]);
+	DWORD size = *(DWORD *)((BYTE *)&temp[0x28]);
+	/* compute seconds based on size */
+	float seconds = (float)size;
+	seconds /= pcm->wf.nSamplesPerSec * pcm->wf.nChannels * pcm->wBitsPerSample / 8;
+
+	/* return number of frames this wave lasts */
+	return (DWORD)(seconds * 20);
+}
+
 DEVILUTION_END_NAMESPACE
--- a/Source/effects.cpp
+++ b/Source/effects.cpp
@@ -1410,24 +1410,4 @@ void effects_play_sound(const char *snd_file)
 	}
 }
 
-DWORD GetSoundFrames(int n)
-{
-	HANDLE hFile;
-	BYTE temp[64];
-
-	WOpenFile(sgSFX[n].pszName, &hFile, FALSE);
-	WReadFile(hFile, temp, sizeof(temp), sgSFX[n].pszName);
-	WCloseFile(hFile);
-
-	/* lazy hack: read the size of the wave file */
-	WAVEFORMATEX *pcm = (WAVEFORMATEX *)((BYTE *)&temp[0x14]);
-	DWORD size = *(DWORD *)((BYTE *)&temp[0x28]);
-	/* compute seconds based on size */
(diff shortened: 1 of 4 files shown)


qndel:
epic 🎉

galaxyhaxz:
Also, the field `txtspd` can be removed from the struct and that data can be wiped from `textdat.cpp`. I just got lazy and didn't want to manually remove each one due to bad formatting.

Source/effects.cpp
```diff
@@ -1410,4 +1410,37 @@ void effects_play_sound(const char *snd_file)
...
+	WORD	nChannels;
+	DWORD	nSamplesPerSec;
+	DWORD	nAvgBytesPerSec;
+	WORD	nBlockAlign;
+} WAVEFORMAT;
+
+typedef struct pcmwaveformat_tag {
+	WAVEFORMAT	wf;
+	WORD		wBitsPerSample;
+} PCMWAVEFORMAT;
```
galaxyhaxz, line 1424: This structure doesn't seem to exist under SDL? Makes no sense, how else are you supposed to read information about a WAV file?

AJenbo: Audio is converted to output format on load, so you can check the output device for this info.

galaxyhaxz: I couldn't find anything on google. Are all the wave files initialized like they were in vanilla? We just need the seconds of the file (with at least 2 decimal points precision) or the header info to calculate it.

AJenbo: From a game logic perspective yes, but wave.cpp has been replaced with SDL.

Would it be funny if Blizzard was dealing with this exact issue 14 years ago?
https://discourse.libsdl.org/t/time-length-of-sdl-mixer-chunks/12852/5

Source/minitext.cpp
```diff
@@ -67,19 +64,77 @@ void InitQuestText()
...
+	}
+
+	return lines;
+}
+
+void CalcTextSpeed(int n)
+{
+	DWORD SfxFrames, TextHeight;
+
+	SfxFrames = GetSoundFrames(n);
```
AJenbo, line 118: I cleaned up the audio code and added a helper function so this can now be done the following way:
```suggestion
	SfxFrames = GetSFXLength(n) * 20 / 1000;
```

Source/effects.cpp
galaxyhaxz: :smile:

Source/minitext.cpp
galaxyhaxz: nice! what do you mean btw? isn't it still drawn at 20 FPS? the reason for this is to compensate lag at any given frame rate.
```c
double qtextDelay = (currTime - sgLastScroll) / (1000.0 / 20);
scrolltexty -= (DWORD)(qtextSpd * qtextDelay);
```
This code right here determines what % of the frame we are on and adjusts the speed so if say you drag the window even though the speech is still playing the game stops drawing so when it returns the text will draw faster to stay in sync with the speaker. It took forever to implement this btw, since the PSX used hardware functions.

AJenbo: No the mouse and scrolling test is rendered at you monitors refresh rate, for extra smooth goodness.

galaxyhaxz: The text seems to still be rendered the same as everything else to me.

AJenbo: For any text with a qtextSpd of 0 or higher (txtspd 5 or lower), you won't be able to see a difference as the text won't move more than 1px pr 50 ms. But for some of the faster texts, there should be a perceptual difference.
I think Gillian usually has a speed of -1 so you should be able to start to make it out there when talking to her. Or talk to Farnham about the Poisoned Water quests for maximum effect :D

galaxyhaxz: Yeah it makes much more of a difference to the shorter texts.

One of the key points is that the lore speeches they added on the PSX are so long that an integer pixel count will eventually cause a massive de-sync between the text and the speaker. They had to implement decimal pixel count. Vanilla did not have any speeches long enough for this to happen.

PR #986, opened 2021-01-24
spectral elixir [usability] bugfix

galaxyhaxz:
One problem with the hack they used to make the Spectral Elixir go from not usable to usable is that it stays usable until Diablo is restarted. This bug was "fixed" on the PSX by having a separate array for item usability and resetting it each game. While this does work, if you kept the elixir and started a new game it will become unusable again.

Therefore I have a more simple fix, to have a usable "copy" of the elixir to replace the old one. This also allows the array to be `const` since it's no longer modified. The downside is that this breaks existing saves because hellfire took up all the empty item slots. In vanilla this would not be the case.

The only other way I could see to fix this would be to add the usability flag to each item and save it. This would require changing several structs and would break compatibility even more.


galaxyhaxz:
I know this is a dirty "hack" but I thought of another way to fix this. This would also allow the game to distinguish between quest items from previous games verse new ones.

Since the item seed is saved, the lowest bit of the seed could act like a flag. `seed &= ~1` would be the elixir/quest item as it spawned. `seed |= 1` would set the elixir to usable (or in the case of quest items, marked as spawned).

AJenbo:
> I know this is a dirty "hack" but I thought of another way to fix this. This would also allow the game to distinguish between quest items from previous games verse new ones.
> 
> Since the item seed is saved, the lowest bit of the seed could act like a flag. `seed &= ~1` would be the elixir/quest item as it spawned. `seed |= 1` would set the elixir to usable (or in the case of quest items, marked as spawned).

I like that idea :) Despite it being a bit of a hack it follows the way `_iCreateInfo` works in a way. By the way are all bits in `_iCreateInfo` already taken, or are there may be some that are meaningless for quest/unique items?

galaxyhaxz:
Yeah, sadly every last bit of the creation flags is used.
```c
// Item re-creation information
// Creation bits are as follows:
// bit# desc
// 1-6	Level
// 7	Item Goodonly (T/F)
// 8,9	Unique percentage (0 = 0%, 1 = 15%, 2 = 1%)
// 8&9	Useful item only
// 10	Unique item
// 11	Spawned by Blacksmith
// 12	Spawned by Blacksmith premium
// 13	Spawned by Pegboy
// 14	Spawned by Witch
// 15	Spawned by Healer
```

qndel:
@galaxyhaxz you can use a hack they did to mark potions and scrolls that spawn on dungeon floor - they combined two existing flags in an unusual way, so you could mark usable elixir by setting for example both pepin and adria flag ;)

FitzRoyX:
Maybe I'm in the minority, but I see no problem with the elixir becoming unusable on new games. There's nothing to do with it except use it the moment you get it, so... you snooze, you lose?

galaxyhaxz:
@FitzRoyX the problem is the other way around, after you finish the quest all elixirs in new games will be usable before talking to Adria. This is more of a quirk but either way the game shouldn't be modifying a global array in that way.

I think all quest items should be deleted when you make a new game.

FitzRoyX:
So the original bug is that it becomes unusable after a total restart of the program? Or will a simple save/load do it?

galaxyhaxz:
The global item array stores the flag to whether or not the elixir can be used. When you complete the quest it sets this flag to true. All elixirs are now usable until you close Diablo. Now when you restart diablo it goes back to unusable. So you could save once you get the elixir, then restart diablo and it won't be usable anymore. It's just an odd design quirk and a quick hack imo.

The way I'm going to fix this in my project is to save quest data flags along with the player. You'll only be able to complete most quests once per difficulty and reap their reward, just like Diablo 2.

PR #987, opened 2021-01-29
fix vanilla typo

qndel:
sound files match "men"


AJenbo:
we should figure out a way to deduplicate all the texts

PR #988, opened 2021-01-30
Transparency

FluffyQuack:
Adds a toggle ("Transparency = 1" in config) to replace the dithering-based transparency with proper transparency (this is accomplished by creating a lookup table listing what colours in the palette are the closest approximation of every set of 2 colours blended).

Screenshots of transparency:
https://cdn.discordapp.com/attachments/518562502250856489/802176578104066078/devilutionx_2021-01-22_14-59-56-125.png
https://cdn.discordapp.com/attachments/518562502250856489/802193938667929650/devilutionx_2021-01-22_16-11-38-472.png

I need feedback on these things:
- I have a tendency to add "//Fluffy: I did thing!" as a way to track what changes I've done when I program. Not sure if those should be kept.
- I added config.cpp which I'm thinking could be a nice place to add future config toggles. Not sure if this should be kept either.
- The new code in RenderLine() can probably be cleaned up, but I'm not sure how.
- I replaced Griswold with a cow. I feel it only makes sense now as we have support for proper transparency.
- ... Okay, fine I didn't replace Griswold with a cow. It was tempting, though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,7 +253,8 @@ add_library(devilution STATIC
   Source/town.cpp
   Source/towners.cpp
   Source/track.cpp
-  Source/trigs.cpp)
+  Source/trigs.cpp
+  Source/Misc/config.cpp)
 
 set(devilutionx_SRCS
   SourceX/dx.cpp
--- a/Source/Misc/config.cpp
+++ b/Source/Misc/config.cpp
@@ -1,0 +1,44 @@
+//Fluffy: For loading various options from the game's config file
+
+#include "..\all.h"
+#include "../../3rdParty/Storm/Source/storm.h"
+
+//TODO: We could create a struct which contains all general options which can be defined in the config file to make them tidier
+
+DEVILUTION_BEGIN_NAMESPACE
+
+static void LoadGameSetupVariableFromConfig(char *name, BOOL *variable)
+{
+	int temp = *variable;
+	if (SRegLoadValue("devilutionx", name, 0, &temp))
+		*variable = (BOOL)temp;
+	else
+		SRegSaveValue("devilutionx", name, 0, temp);
+}
+
+void LoadOptionsFromConfig()
+{
+	if (!SRegLoadValue("devilutionx", "game speed", 0, &ticks_per_sec)) {
+		SRegSaveValue("devilutionx", "game speed", 0, ticks_per_sec);
+	}
+	tick_delay_highResolution = SDL_GetPerformanceFrequency() / ticks_per_sec; //Fluffy
+
+	//Fluffy: Load speed modifiers from config, but if they don't exist, then we calculate based on the tick rate
+	if (!SRegLoadValue("devilutionx", "Game Simulation Speed Modifier", 0, &gSpeedMod)) {
+		gSpeedMod = ticks_per_sec / 20;
+	}
+	if (!SRegLoadValue("devilutionx", "Monster Speed Modifier", 0, &gMonsterSpeedMod)) {
+		gMonsterSpeedMod = ticks_per_sec / 20;
+	}
+	if (gSpeedMod < 1)
+		gSpeedMod = 1;
+	if (gMonsterSpeedMod < 1)
+		gMonsterSpeedMod = 1;
+
+	//Fluffy: Load game setup from config here when booting up singleplayer (if we fail to load it, then we save its default to the config)
+	LoadGameSetupVariableFromConfig("Fast Walking In Town", &gameSetup_fastWalkInTown);
+	LoadGameSetupVariableFromConfig("Allow Attacks In Town", &gameSetup_allowAttacksInTown);
+	LoadGameSetupVariableFromConfig("Transparency", &options_transparency);
+}
+
+DEVILUTION_END_NAMESPACE
--- a/Source/Misc/config.cpp
+++ b/Source/Misc/config.cpp
@@ -1,6 +1,6 @@
 //Fluffy: For loading various options from the game's config file
 
-#include "..\all.h"
+#include "../all.h"
 #include "../../3rdParty/Storm/Source/storm.h"
 
 //TODO: We could create a struct which contains all general options which can be defined in the config file to make them tidier
--- a/Source/Misc/config.cpp
+++ b/Source/Misc/config.cpp
@@ -21,23 +21,8 @@ void LoadOptionsFromConfig()
 	if (!SRegLoadValue("devilutionx", "game speed", 0, &ticks_per_sec)) {
 		SRegSaveValue("devilutionx", "game speed", 0, ticks_per_sec);
 	}
-	tick_delay_highResolution = SDL_GetPerformanceFrequency() / ticks_per_sec; //Fluffy
-
-	//Fluffy: Load speed modifiers from config, but if they don't exist, then we calculate based on the tick rate
-	if (!SRegLoadValue("devilutionx", "Game Simulation Speed Modifier", 0, &gSpeedMod)) {
-		gSpeedMod = ticks_per_sec / 20;
-	}
-	if (!SRegLoadValue("devilutionx", "Monster Speed Modifier", 0, &gMonsterSpeedMod)) {
-		gMonsterSpeedMod = ticks_per_sec / 20;
-	}
-	if (gSpeedMod < 1)
-		gSpeedMod = 1;
-	if (gMonsterSpeedMod < 1)
-		gMonsterSpeedMod = 1;
 
 	//Fluffy: Load game setup from config here when booting up singleplayer (if we fail to load it, then we save its default to the config)
-	LoadGameSetupVariableFromConfig("Fast Walking In Town", &gameSetup_fastWalkInTown);
-	LoadGameSetupVariableFromConfig("Allow Attacks In Town", &gameSetup_allowAttacksInTown);
 	LoadGameSetupVariableFromConfig("Transparency", &options_transparency);
 }
 
--- a/Source/Misc/config.cpp
+++ b/Source/Misc/config.cpp
@@ -3,8 +3,6 @@
 #include "../all.h"
 #include "../../3rdParty/Storm/Source/storm.h"
 
-//TODO: We could create a struct which contains all general options which can be defined in the config file to make them tidier
-
 DEVILUTION_BEGIN_NAMESPACE
 
 static void LoadGameSetupVariableFromConfig(char *name, BOOL *variable)
--- a/Source/Misc/config.cpp
+++ b/Source/Misc/config.cpp
@@ -1,4 +1,4 @@
-//Fluffy: For loading various options from the game's config file
+//For loading various options from the game's config file
 
 #include "../all.h"
 #include "../../3rdParty/Storm/Source/storm.h"
@@ -22,7 +22,7 @@ void LoadOptionsFromConfig()
 		SRegSaveValue("devilutionx", "game speed", 0, ticks_per_sec);
 	}
 
-	//Fluffy: Load game setup from config here when booting up singleplayer (if we fail to load it, then we save its default to the config)
+	//Load game setup from config here when booting up singleplayer (if we fail to load it, then we save its default to the config)
 	LoadGameSetupVariableFromConfig("Transparency", &options_transparency);
 }
 
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -56,6 +56,7 @@ char sgbMouseDown;
 int color_cycle_timer;
 int ticks_per_sec = 20;
 WORD tick_delay = 50;
+BOOL options_transparency = false; //Fluffy: If true, we apply proper transparency rather than dithering
 
 /* rdata */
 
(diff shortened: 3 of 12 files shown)


Source/Misc/config.cpp
```diff
@@ -0,0 +1,29 @@
+//Fluffy: For loading various options from the game's config file
+
+#include "..\all.h"
+#include "../../3rdParty/Storm/Source/storm.h"
```
qndel, line 4: inconsistent usage of backslash / forward slash ;)

CMakeLists.txt
```diff
@@ -253,7 +253,8 @@ add_library(devilution STATIC
   Source/town.cpp
   Source/towners.cpp
   Source/track.cpp
-  Source/trigs.cpp)
+  Source/trigs.cpp
+  Source/Misc/config.cpp)
```
qndel, line 257: creating a new file just for 1 function seems to be the very definition of an overkill

Source/diablo.cpp
```diff
@@ -498,6 +500,7 @@ void diablo_quit(int exitStatus)
 int DiabloMain(int argc, char **argv)
 {
 	diablo_parse_flags(argc, argv);
+	LoadOptionsFromConfig(); //Fluffy: Read options from config here
```
qndel, line 503: look how your function name stands out from the surrounding ones

AJenbo:
Don't add comments about who did the code it isn't really helpful and we can track it in git much better (see git blame)

galaxyhaxz:
Diablo stores the dungeon palette in 0x00-0x7F and the shared item/player palette in 0x80-FF. Therefore you can calculate the last 128 colors blended with each other only once, by using a static bool inside the function. This will cut the time to generate the blended palette in half ;)

Source/Misc/config.cpp
FluffyQuack: Fixed

qndel: not only in this place :P

CMakeLists.txt
FluffyQuack: The idea is to use this moving forward for more config options rather than scattering them throughout the code. And considering the source files in Diablo are absolutely enormous, I personally don't think it's a bad idea to start adding more source files rather than keeping adding more and more to the existing ones.

But I can remove this file and move the config check somewhere else.

Source/diablo.cpp
FluffyQuack: It doesn't stand out compared to "DiabloMain()"

Do we have standards for how functions should be named in Devilution?

FluffyQuack:
> Don't add comments about who did the code it isn't really helpful and we can track it in git much better (see git blame)

I personally really prefer seeing this in the code itself as I can immediately identify who did what, but I just removed all the fluffy mentions.

FluffyQuack:
> Diablo stores the dungeon palette in 0x00-0x7F and the shared item/player palette in 0x80-FF. Therefore you can calculate the last 128 colors blended with each other only once, by using a static bool inside the function. This will cut the time to generate the blended palette in half ;)

I was considering to do that, but then I started thinking there could be a chance that blended colours in the static part of the palette might actually be closest to something in the dynamic part. I didn't check if that happens in practice, but in theory it could happen.

AJenbo:
The way you did config.h only works for Visual Stuidio:
```
[ 42%] Built target Radon
/root/repo/Source/diablo.cpp:11:74: fatal error: misc\config.h: No such file or directory
 #include "misc\config.h" //For reading options from config during startup
                                                                          ^
compilation terminated.
CMakeFiles/devilution.dir/build.make:166: recipe for target 'CMakeFiles/devilution.dir/Source/diablo.cpp.o' failed
make[2]: *** [CMakeFiles/devilution.dir/Source/diablo.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[ 43%] Building CXX object CMakeFiles/devilution.dir/Source/monster.cpp.o
/root/repo/Source/mainmenu.cpp:9:62: fatal error: misc\config.h: No such file or directory
 #include "misc\config.h" //For re-reading options from config
                                                              ^
[ 43%] Building CXX object CMakeFiles/devilution.dir/Source/movie.cpp.o
compilation terminated.
CMakeFiles/devilution.dir/build.make:452: recipe for target 'CMakeFiles/devilution.dir/Source/mainmenu.cpp.o' failed
make[2]: *** [CMakeFiles/devilution.dir/Source/mainmenu.cpp.o] Error 1
CMakeFiles/Makefile2:195: recipe for target 'CMakeFiles/devilution.dir/all' failed
make[1]: *** [CMakeFiles/devilution.dir/all] Error 2
Makefile:151: recipe for target 'all' failed
make: *** [all] Error 2
```

galaxyhaxz:
> chance that blended colours in the static part of the palette might actually be closest to something in the dynamic part.

Hmm technically you're right, I didn't think about that. One thing I noticed when I used blended transparency in my project is an issue with gray scale colors. Let's say the actual blended color would be RGB (128,128,128). However the way distance is calculated, we have the color (128,120,121). Therefore the delta is 15. We have another color which is (123,123,123). This color also has a delta of 15, however since the other color came first it gets chosen.

So even though the deltas are the same, 123,123,123 is a gray color that more closely matches 128,128,128. Whereas 128,120,121 will appear slightly reddish. I noticed this issue when drawing black dialog boxes, it appeared red rather than black. I fixed this issue by writing a basic algorithm that puts a "weight" on each color, so it will instead apply a threshold for whatever color is the strongest. The result is more accurate colors.

FluffyQuack:
> The way you did config.h only works for Visual Stuidio:
> 
> ```
> [ 42%] Built target Radon
> /root/repo/Source/diablo.cpp:11:74: fatal error: misc\config.h: No such file or directory
>  #include "misc\config.h" //For reading options from config during startup
>                                                                           ^
> compilation terminated.
> CMakeFiles/devilution.dir/build.make:166: recipe for target 'CMakeFiles/devilution.dir/Source/diablo.cpp.o' failed
> make[2]: *** [CMakeFiles/devilution.dir/Source/diablo.cpp.o] Error 1
> make[2]: *** Waiting for unfinished jobs....
> [ 43%] Building CXX object CMakeFiles/devilution.dir/Source/monster.cpp.o
> /root/repo/Source/mainmenu.cpp:9:62: fatal error: misc\config.h: No such file or directory
>  #include "misc\config.h" //For re-reading options from config
>                                                               ^
> [ 43%] Building CXX object CMakeFiles/devilution.dir/Source/movie.cpp.o
> compilation terminated.
> CMakeFiles/devilution.dir/build.make:452: recipe for target 'CMakeFiles/devilution.dir/Source/mainmenu.cpp.o' failed
> make[2]: *** [CMakeFiles/devilution.dir/Source/mainmenu.cpp.o] Error 1
> CMakeFiles/Makefile2:195: recipe for target 'CMakeFiles/devilution.dir/all' failed
> make[1]: *** [CMakeFiles/devilution.dir/all] Error 2
> Makefile:151: recipe for target 'all' failed
> make: *** [all] Error 2
> ```

*screams*

How are new header files supposed to be added?

AJenbo:
> > Diablo stores the dungeon palette in 0x00-0x7F and the shared item/player palette in 0x80-FF. Therefore you can calculate the last 128 colors blended with each other only once, by using a static bool inside the function. This will cut the time to generate the blended palette in half ;)
> 
> I was considering to do that, but then I started thinking there could be a chance that blended colours in the static part of the palette might actually be closest to something in the dynamic part. I didn't check if that happens in practice, but in theory it could happen.

Well, you would still cut it in half if you only blend the dungeon colors, but look for matches in all 256.

BUT! There are also a few cases of transparent sprites, like when you have a two-handed weapon so you actually need to blend all 256.

Maybe some runtime optimization could be done by generating two 128x128 tables :shrug:

FluffyQuack:
> Hmm technically you're right, I didn't think about that. One thing I noticed when I used blended transparency in my project is an issue with gray scale colors. Let's say the actual blended color would be RGB (128,128,128). However the way distance is calculated, we have the color (128,120,121). Therefore the delta is 15. We have another color which is (123,123,123). This color also has a delta of 15, however since the other color came first it gets chosen.
> 
> So even though the deltas are the same, 123,123,123 is a gray color that more closely matches 128,128,128. Whereas 128,120,121 will appear slightly reddish. I noticed this issue when drawing black dialog boxes, it appeared red rather than black. I fixed this issue by writing a basic algorithm that puts a "weight" on each color, so it will instead apply a threshold for whatever color is the strongest. The result is more accurate colors.

I'm not sure if my code does the same. My first iteration definitely would have, but I applied one change suggested by AJenbo which should result in more accurate blending. I have a screenshot of a dialogue box in the first post. Does that look correct to you?

AJenbo:
> > Don't add comments about who did the code it isn't really helpful and we can track it in git much better (see git blame)
> 
> I personally really prefer seeing this in the code itself as I can immediately identify who did what, but I just removed all the fluffy mentions.

Your editor probably has build-in blame support where you can see who did it on the left side column.

CMakeLists.txt
AJenbo: New files should go in SourceX, shared headers got in SourceS

Source/diablo.cpp
AJenbo: CamelCase is the standard. The rest is names that should be cleaned up.

galaxyhaxz:
Yeah it looks fine. You don't have to beef up the algorithm if you don't want to, as it will also cost more CPU. But a few basic techniques could be check if R==G==B. If so, only blend with other R==G==B colors.

Regarding actual "colors", what I did was increased the delta for the strongest color. So if we have (150,140,140) to the naked human eye this is slightly red. So I increase that to (170,140,140) which means the delta is stronger. This makes it choose a stronger shade of red. Again, this is just for perfection, I think the method you have is fine otherwise, we had code like that nearly two years ago ;)

Source/diablo.cpp
```diff
@@ -8,6 +8,7 @@
 #include "../3rdParty/Storm/Source/storm.h"
 #include "../DiabloUI/diabloui.h"
 #include <config.h>
+#include "misc\config.h" //For reading options from config during startup
```
AJenbo, line 11: Probably this will help:
```suggestion
#include "misc/config.h" //For reading options from config during startup

```

FluffyQuack:
> Your editor probably has build-in blame support where you can see who did it on the left side column.

I just found a feature for that in VS, but it opens a separate window where you can't write code, so it's a bit out of the way. I think it'll be fine if I write comments as I'm used to in my own fork, and then I edit them for any PRs. On a related note, I hope my excessive comments isn't a bad thing. I like comments. It works well since my memory is really bad, so it's quick to understand what code does if there are comments for it.

AJenbo:
> > Your editor probably has build-in blame support where you can see who did it on the left side column.
> 
> I just found a feature for that in VS, but it opens a separate window where you can't write code, so it's a bit out of the way. I think it'll be fine if I write comments as I'm used to in my own fork, and then I edit them for any PRs. On a related note, I hope my excessive comments isn't a bad thing. I like comments. It works well since my memory is really bad, so it's quick to understand what code does if there are comments for it.

Yours aren't excessive. Ideally the code should be broken up in units that are easy to understand, but this is a good first step.

Source/mainmenu.cpp
```diff
@@ -6,6 +6,7 @@
 #include "all.h"
 #include "../3rdParty/Storm/Source/storm.h"
 #include "../DiabloUI/diabloui.h"
+#include "misc\config.h" //For re-reading options from config
```
AJenbo, line 9: There is at least one of this left.

galaxyhaxz: image

FluffyQuack: Fixed. I did a search, and I couldn't find any other instances in the code.

Source/palette.cpp
```diff
@@ -99,6 +100,44 @@ void LoadPalette(const char *pszFileName)
...
+	* We save this info in a lookup table we use during rendering for whenever we want this kind of transparency.
+	* 
+	*/
+	if (options_transparency == true) {
+		for (int i = 0; i < 256; i++) {
+			for (int j = 0; j < 256; j++) {
+				if (i == j) { //No need to calculate transparency between 2 identical colours
+					palette_transparency_lookup[i][j] = j;
+					continue;
+				} else if (i > j) { //Since there's a lot of redundancy ([i][j] will always have the same value as [j][i]), we skip calculating combinations which have already been calculated
```
AJenbo, line 116: Nice and effective optimization :)

Source/mainmenu.cpp
```diff
@@ -6,6 +6,7 @@
 #include "all.h"
 #include "../3rdParty/Storm/Source/storm.h"
 #include "../DiabloUI/diabloui.h"
+#include "misc/config.h" //For re-reading options from config
```
qndel, line 9: windows isn't case sensitive, other platforms are

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/790c8e17e507ec34d839836f9cb3d4e587110f7e

- I didn't feel comfortable about adding an extra check to each pixel during trans_rect so opted for duplicating the 8 lines instead.
- Made options a struct.
- Split up saving and loading of options.
- Fixed a bit of double rendering (not visible)
- A bit of formatting
- Since the options weren't compiling and still needs work I simply moved the few lines to diablo.cpp, later we can move it to SourceX and add all the other options that are loaded from the INI.

AJenbo:
@glebm you might want to perf test this on some of the lower-end devices to see if it needs to be turned off by default for them.

PR #992, opened 2021-02-02
[bugfix] Quest items from previous games breaking dungeon

galaxyhaxz:
If you held quest items over from a previous game, such as Arkaine's Valor, then bring them to a new game where that quest didn't spawn, it can erroneously set the quest status from not available to completed. If you spawn the dungeon, go to a different level, pick the item up, then come back, the dungeon will be messed up as it will spawn the quest set pieces.

As a good side effect "May the spirit of Arkaine protect me" will only play once now and then the quest finishes so never again. Whew, thank god ^^

Can't remember who found this bug, @Nitekat ?

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1567,7 +1567,7 @@ void CheckBookLevel(int pnum)
 
 void CheckQuestItem(int pnum)
 {
-	if (plr[pnum].HoldItem.IDidx == IDI_OPTAMULET)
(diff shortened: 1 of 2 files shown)


AJenbo:
Nice, especially with the added enum comments :)

galaxyhaxz:
Also this fix doesn't fully flesh out all possible flaws. For example you could still start the Valor quest, drop and pickup a valor from the previous game and it completes the quest. I don't think there is any effective way to fix this without reworking part of the game, such as binding a seed to the quest or having a flag stored in the item.

PR #993, opened 2021-02-03
Refactoring of walk functions for players and monsters

FluffyQuack:
There's a lot of duplicated code in movement code for players and monsters (the DoWalk and StartWalk functions), so I merged a bunch of functions. I did this for players and monsters, though I left the StartWalk functions for monsters untouched as those have almost no duplicated code.

I also added comments for parts of the functions (basically, the parts I fully understand).

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -2044,23 +2044,41 @@ BOOL M_DoStand(int i)
 	return FALSE;
 }
 
-BOOL M_DoWalk(int i)
+/**
+ * @brief Continue movement towards new tile
+ */
+bool M_DoWalk(int i, int variant)
 {
-	BOOL rv;
+	bool returnValue;
 
 	commitment((DWORD)i < MAXMONSTERS, i);
 	commitment(monster[i].MType != NULL, i);
 
+	//Check if we reached new tile
 	if (monster[i]._mVar8 == monster[i].MType->Anims[MA_WALK].Frames) {
-		dMonster[monster[i]._mx][monster[i]._my] = 0;
-		monster[i]._mx += monster[i]._mVar1;
-		monster[i]._my += monster[i]._mVar2;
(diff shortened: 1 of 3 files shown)


galaxyhaxz:
Good work, they did this on the playstation version as well. There they also changed walking to work the same in all direction, by moving 2 pixels at a time rather than a whole tile.

FluffyQuack:
> Good work, they did this on the playstation version as well. There they also changed walking to work the same in all directions, by moving 2 pixels at a time rather than a whole tile.

That's something I want to experiment with, adding movement similar to how it works in the PS1 version.

galaxyhaxz:
I've tried twice and failed both times. Came to the conclusion that first the engine has to be changed to render in that way.

IMO the movement scheme is vastly superior to what we have now, especially for controllers.

FluffyQuack:
> I've tried twice and failed both times. Came to the conclusion that first the engine has to be changed to render in that way.
> 
> IMO the movement scheme is vastly superior to what we have now, especially for controllers.

Better code for defining render order would be nice. It looks like the Diablo dev team fixed render order by changing what tile players/monsters belong to and thus made the movement code overly messy. If the render code ordered entities, that would make it possible to make the movement code far more sense and also make it viable to implement PS1-style movement.

It would be a lot of work overall, but maybe something which can be done down the road.

Another change I'd like to see (and I suspect the PS1 version did this) is separate shadows from the sprite assets. That would make it possible to have transparent shadows and have correct outlines (right now shadows are distorting outlines slightly).

galaxyhaxz:
The PS1 did have a sorting table rather than a back buffer. Each sprite/tile is assigned to a polygon and sorted based on the Y value, they also separated walls from floors. Together this fixed the crappy EFlag issues. Ajenbo has mostly mitigated it somewhat using a mask.

PS1 also had dynamic shadows for some things, not sure if for everything.

PR #995, opened 2021-02-05
Proper fix for dungeon overflow

galaxyhaxz:
In the dungeon generation for cathedral there is an overflow bug. The function for checking if the room width/height is in bounds has swapped values, which can then clear them as being ok, but then the width/height is applied in the proper position later which may overflow.

This breaks existing saves as the dungeon layouts are now different. For the hell version, the bug never broke any dungeons but here is one example from cathedral:

DIABLO_20210205_212035


AJenbo:
breaking save game compatibility isn't an option unfortunately.

galaxyhaxz:
Well Anders DevilutionX is your project, not mine. You will be holding yourself back on many levels if you plan to keep compatibility, as you will be unable to fix many bugs, unable to expand max amount of missiles or other entities, unable to rework the way items are saved to fix uniques, and more. The whole purpose of Devilution was to give us a starting point so we could give this old game the overhaul it needs, and I mean it needs a lot. The vast majority of this game is horribly outdated shit code. You have to break eventually and that's ok when you do.

FluffyQuack:
Maybe there could be a toggle for "vanilla compatibility"? With it off, it would apply changes like this. With it on, it will do the same thing as the original game.

The more I look at the code, the more I'm personally okay with compatibility being broken. There are so many things in the original game which behave so poorly and could be dramatically improved.

aydjay:
Yea - as a player and an interested nobody, fixing bugs and  adding qol features is more important than keeping compatibility with saves.

AJenbo:
If we break compatibility it becomes a lot harder to verify behavior. But also we would break save games from any prior version of DevilutionX as well. I think this would be a bit of a surprise for a lot of users since it's not common for game updates to require the user to start over.

It's true that this poses some limits on how we solve problems, but there are often still ways to deal with the issues.

The missile limits could be handled by saving the additional data in an expanded part of the save game.
Fix unique items morphing could be solved by reading the items from the `game` section instead of the `hero` part of the savegame.

Regarding changes like this, DevilutionX already has a flag (`pOriginalCathedral`) in the savegames for tracking if the Diablo (0) or Hellfire (1) version of the cathedral algorithm should be used. We could build on that and add a DevilutionX version (2) as well. This would only break compatibility in terms of loading save games in an older version of DeivlutionX or the original game, but players could still continue existing saves in newer versions.

galaxyhaxz:
> If we break compatibility it becomes a lot harder to verify behavior.

We achieved 100% bin exactness with Devilution, this means we have identical behavior to vanilla. I have verified the bugfix in this PR by testing several hundred thousand seeds and the broken tiles do not occur.

> Fix unique items morphing could be solved by reading the items from the game section instead of the hero part of the savegame.

But this wouldn't solve the problem? The problem is when you make a new game, it has to recreate the unique from a packed item struct which doesn't store the unique ID, this is what's causing all the morphing especially for uniques with multiple bases (i.e. if we want Lightforge and Crackrust to drop we need to fix it).

> The missile limits could be handled by saving the additional data in an expanded part of the save game.

Seems like a hack imo. The code will get very messy very quickly when you start throwing parts of arrays all over the place.

> tracking if the Diablo (0) or Hellfire (1) version of the cathedral algorithm should be used

I don't believe the hellfire cathedral was any different aside from one minor out-of-bounds fix that didn't' re-scramble the dungeon.

AJenbo:
> > Fix unique items morphing could be solved by reading the items from the game section instead of the hero part of the savegame.
> 
> But this wouldn't solve the problem? The problem is when you make a new game, it has to recreate the unique from a packed item struct which doesn't store the unique ID, this is what's causing all the morphing especially for uniques with multiple bases (i.e. if we want Lightforge and Crackrust to drop we need to fix it).

If you load the items from the `game` data instead of `hero` then you get the full item data and that solves the problem since it doesn't have to recreate it from the packed data.

> I don't believe the hellfire cathedral was any different aside from one minor out-of-bounds fix that didn't' re-scramble the dungeon.

The code for placing the stairs is different and so is the size they occupy, this causes levels to break when loading save games from Diablo in Hellfire.

galaxyhaxz:
> If you load the items from the game data instead of hero then you get the full item data and that solves the problem since it doesn't have to recreate it from the packed data.

Technically true, but then the packed data serves no purpose. And i don't believe multiplayer has the game data as only the hero is saved.

> The code for placing the stairs is different and so is the size they occupy, this causes levels to break when loading save games from Diablo in Hellfire.

Ah right, how could I forget since I was the one who did this PR. They changed the staircase yeah, or at least one of the unused ones to make it fit the crypt.

FitzRoyX:
>I think this would be a bit of a surprise for a lot of users since it's not common for game updates to require the user to start over.
>
That you've managed to avoid it thus far is quite generous of you, but it's an unrealistic expectation if users have it. The only way to guarantee the savegame format never breaks is to never make public releases until the very end. I don't think anyone wants to wait 5 years between releases, it's easier to just keep an old version of devx until that character is finished.

galaxyhaxz:
Another approach to mitigate this is do what Diablo 2 did: allow upgrading old saves to the new format, but not the other way around. Diablo 2 had several major game revisions and each one a new format. If you're not comfortable convoluting the code you could even pass it as a separate tool.

AJenbo:
> Another approach to mitigate this is do what Diablo 2 did: allow upgrading old saves to the new format, but not the other way around. Diablo 2 had several major game revisions and each one a new format.

Yeah, that's what I suggested in my second comment.

> EDIT: and to go along with what was already said, you should consider DevilutionX a beta still. Despite its v1.0 version moniker, it is still largely unfinished.

Beta refers to the level of know issues compared to the target goal, the target for 1.0.0 was parity with Diablo 1.09 which I believe we achieved.

The releases follow semantic versioning 0.y.z is for initial development and not considered stable.
Once 1.0.0 is released the version number indicates the following:
(compatibility).(features).(bug fixes)
Between each release, the next version goes through the development cycle again, but that doesn't change the status of previous releases. This ensures that we can periodically release something that is a buggy mess and in an endless state of beta testing.

We do have long-term plans for many releases that will greatly improve various aspects of the game, so I'm thinking that's what you are referring to :)

AJenbo:
> Yea - as a player and an interested nobody, fixing bugs and adding qol features is more important than keeping compatibility with saves.

We receive bug reports from users when save-games aren't loading correctly so at least to some users compatibility is important.

> That you've managed to avoid it thus far is quite generous of you, but it's an unrealistic expectation if users have it. The only way to guarantee the savegame format never breaks is to never make public releases until the very end.

Thanks, but I don't agree. Maintaining the ability to load old savegames isn't that much extra work.

PR #999, opened 2021-02-08
Quest bug fixes including Valor [proper]

galaxyhaxz:
Three new fixes here. First and foremost, I'd like to credit @qndel for finding the Valor bug. It took awhile to finally track down the actual cause.

#### Arkaine's Valor quest could sometimes be partially/fully completed

This is caused by the fact the number of stones in the pedestal are stored in `_oVar6`. This variable is never initialized and only used by books and pedestals. Let's say you read a book on level 4 and it sets Var6 to 3, now this book is in object slot `50`. You go down to level 5 and the pedestal also happens to spawn in slot 50. This means it will think there are 3 stones despite the fact there are none.

#### Blood Stones spawn all at once

Now they only spawn one at a time, which prevents people from being able to telekinesis them (sorry speedrunners!). This also as a side effect fixes the accidental deletion if you drop one.

#### Quest items don't spawn when 127 items are on floor

If you try to click on the mushroom patch for example, it will "use it" but not spawn anything, breaking the quest. Now these objects will not function until a slot is available.

All 3 of these bugs were fixed on the PSX port, and I took the same approach code-wise.

Capture

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1496,6 +1496,7 @@ void AddPedistal(int i)
 	object[i]._oVar2 = setpc_y;
 	object[i]._oVar3 = setpc_x + setpc_w;
 	object[i]._oVar4 = setpc_y + setpc_h;
+	object[i]._oVar6 = 0;
 }
 
 void AddStoryBook(int i)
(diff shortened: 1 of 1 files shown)


qndel (approved):
Looks good 👌

PR #1010, opened 2021-02-17
Remove items dropping before entering level

galaxyhaxz:
This makes the item stay put in the cursor when entering a new level/portal. It also keeps items from dropping when talking to towners. Overall the dropping items thing was pointless and hacky.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1781,10 +1781,8 @@ void AutoGetItem(int pnum, int ii)
 		dropGoldValue = 0;
 	}
 
-	if (ii != MAXITEMS) {
-		if (dItem[item[ii]._ix][item[ii]._iy] == 0)
-			return;
-	}
+	if (dItem[item[ii]._ix][item[ii]._iy] == 0)
+		return;
 
 #ifdef HELLFIRE
 	if (item[ii]._iUid != 0)
@@ -2606,17 +2604,6 @@ int CalculateGold(int pnum)
 	}
 
 	return gold;
-}
-
-BOOL DropItemBeforeTrig()
-{
-	if (TryInvPut()) {
-		NetSendCmdPItem(TRUE, CMD_PUTITEM, cursmx, cursmy);
-		NewCursor(CURSOR_HAND);
-		return TRUE;
-	}
-
-	return FALSE;
 }
 
 DEVILUTION_END_NAMESPACE
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -2606,4 +2606,15 @@ int CalculateGold(int pnum)
 	return gold;
 }
 
+BOOL DropItemBeforeTrig()
+{
+	if (TryInvPut()) {
+		NetSendCmdPItem(TRUE, CMD_PUTITEM, cursmx, cursmy);
+		NewCursor(CURSOR_HAND);
+		return TRUE;
+	}
+
+	return FALSE;
+}
+
 DEVILUTION_END_NAMESPACE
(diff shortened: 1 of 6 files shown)


Source/inv.cpp
```diff
@@ -1781,10 +1781,8 @@ void AutoGetItem(int pnum, int ii)
 		dropGoldValue = 0;
 	}
 
-	if (ii != MAXITEMS) {
```
qndel, line 1784: why did you remove this? :P

FitzRoyX:
I wouldn't say it's that pointless to drop held items when talking to towners. It's far easier to click on the wrong thing when your cursor is a giant item like a shield instead of a small pointy hand.

If you mean just changing the cursor and going back to the item when menu is closed, well now you've got the problem of the player possibly forgetting that he has an item held while transacting and buying 1 too many things.

Source/inv.cpp
galaxyhaxz:
```
if (pnum == myplr && pcurs >= CURSOR_FIRSTITEM) {
	item[MAXITEMS] = plr[myplr].HoldItem;
	AutoGetItem(myplr, MAXITEMS);
}
```
`if (ii != MAXITEMS)` was part of the hack added in 1.02 to prevent items disappearing from cursor when entering a portal. They store the item into the temporary slot and added that check to `AutoGetItem` so it will go into inventory. This check is pointless once the above code is removed because items on the ground can only be in slots 0-126, slot 127 is otherwise impossible and only used in the pack player code.

galaxyhaxz:
> If you mean just changing the cursor and going back to the item when menu is closed, well now you've got the problem of the player possibly forgetting that he has an item held while transacting and buying 1 too many things.

What happens is if you click on a towner, pick up an item from inventory, then when you reach the towner the item is dropped on the ground and if there isn't any room it cancels talking to them. This simply forces it to always fail talking to them while you have an item on the cursor, as imagine a multiplayer game where it drops your item and someone else snatches it.

Same concept applies to portals, Just like in Diablo 2, the item will stay on cursor instead of dropping before you enter.

galaxyhaxz:
Also note that that the previous code was buggy in nature. For example, when you enter a portal and pick up an item in cursor, you could see the item get copied into the inventory but still remain in cursor, creating a temporary dupe, and if there was no room to place in inventory the item vanishes.

Regarding the DSF buglist:

> If a character goes to pick up an item as he/she is entering a Town Portal, a dupe will appear on the other side, thus creating the same item in town and in the dungeon. Neither item will be destroyed.

I wasn't able to reproduce this bug before or after this PR. Though it definitely sounds to be related to the old code that is now removed.

NiteKat:
Something to note, some of the store coded uses the hold item slot as a temporary item holder (I think the confirm dialogs do this?). If you allowed someone to navigate a store while holding an item, they could morph their item.

qndel:
well, I bet stores restore the hand cursor after that? So the best you could do would be deleting your cursor item :D

NiteKat:
I don't think they mess with the cursor icon, since they just set values into the hold item slot, they don't use the usual functions that would place an item there and modify the cursor, but I could be wrong.

joewis:
If I remember right, purchased items go to HoldItem and from there via AutoPlace to inventory.
In that case, if you're holding something it will 'morph' into nothing.

galaxyhaxz:
These petty pet projects aren't helping me land a job let alone a fucking Interview. So no more coding, if you want something done you'll have to do it yourself.

PR #1011, opened 2021-02-18
Fix portals appearing on wrong levels

qndel:
To reproduce before fix:
Go to leoric's tomb, cast town portal, go back to dlvl 1, you can see your portal there.
That's because leoric's setlvlnum is 1

### DIFF
--- a/Source/portal.cpp
+++ b/Source/portal.cpp
@@ -68,7 +68,7 @@ void SyncPortals()
 			int lvl = currlevel;
 			if (setlevel)
 				lvl = setlvlnum;
-			if (portal[i].level == lvl)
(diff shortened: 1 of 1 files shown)


AJenbo:
This is half the reason quests do not work in multiplayer :)

galaxyhaxz:
They fixed this bug on the PSX and additionally got all quests working in multiplayer. Even though you have split screen, it technically still works the same as multiplayer and uses all the same code.

AJenbo:
The biggest issue is that players appear on multiple levels so that is unfair with split screen as it's not affected :)

AJenbo (approved):

PR #1013, opened 2021-02-18
Fix 4 tile and line checks

NiteKat:
Switches to a better line drawing algorithm for monsters that need to check if they have line of sight, and fixes situations in the ranged AI where they would just stand still, allowing a player to freely kill them with no resistance.

The old line drawing algorithm only worked in one direction, so if that direction worked, but the other direction didn't and the other direction is the one the projectile would actually travel from the monster, then it would erroneously think it could land a hit. This algorithm works in all directions, so it can safely check the proper path. There still may be an occasional instance where the correct line is not drawn, but I think I've only seen that in one or two instances in testing, much more rare than the vanilla code.

The other change, fixing the standing still behavior, just modifies that code path so the monster walks towards the players. In the original code there was a code path that if there was no line of sight, the monster would just stand still. This should probably do something more intelligent, like find a spot that has line of sight and move toward that, but in the spirit of how the AI originally behaves, I just made it walk to the player. This produces some pacing behavior, but eventually they might settle in a proper spot to attack the player. Better than just standing there. :)

This change will also make monsters that are on top of the player think they don't have a line of sight (to avoid division by 0), which can happen if a player spawns on the floor while a monster is on the spawn tile. I don't think this'll have a big impact on gameplay - most monsters that would end up in this position would want to run away due to the distance, and for those that wouldn't (like maelstroms and blood stones, etc.), it just means they'll stick to melee instead of using their ranged attack.

This may need some additional play testing to make sure there aren't any bad side effects. Thank you to @galaxyhaxz for help with the line algorithm! And thank you to @qndel for helping me stumble through my first pull request.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -3648,7 +3648,10 @@ void MAI_Ranged(int i, int missile_type, BOOL special)
 				else
 					M_StartRAttack(i, missile_type, 4);
 			} else {
-				Monst->_mAnimData = Monst->MType->Anims[MA_STAND].Data[md];
+				fx = Monst->_lastx;
+				fy = Monst->_lasty;
+				md = GetDirection(Monst->_mx, Monst->_my, fx, fy);
+				M_CallWalk(i, md);
 			}
 		}
 	} else if (Monst->_msquelch != 0) {
@@ -4912,83 +4915,50 @@ BOOL CheckNoSolid(int x, int y)
 
 BOOL LineClearF(BOOL (*Clear)(int, int), int x1, int y1, int x2, int y2)
 {
-	int xorg, yorg;
-	int dx, dy;
-	int d;
-	int xincD, yincD, dincD, dincH;
-	int tmp;
-	BOOL done = FALSE;
+	int dx = x2 - x1;
+	int dy = y2 - y1;
+	int steps;
+	BOOL done = false;
 
-	xorg = x1;
-	yorg = y1;
-	dx = x2 - x1;
-	dy = y2 - y1;
-	if (abs(dx) > abs(dy)) {
-		if (dx < 0) {
-			tmp = x1;
-			x1 = x2;
-			x2 = tmp;
-			tmp = y1;
-			y1 = y2;
-			y2 = tmp;
-			dx = -dx;
-			dy = -dy;
-		}
-		if (dy > 0) {
-			d = 2 * dy - dx;
-			dincD = 2 * dy;
-			dincH = 2 * (dy - dx);
-			yincD = 1;
+	if (abs(dx) > abs(dy))
+		steps = abs(dx);
+	else
+		steps = abs(dy);
+
+	if (steps == 0)
+		return false;
+
+	float Xinc = (float)dx / steps;
+	float Yinc = (float)dy / steps;
+
+	float x = x1;
+	float y = y1;
+
+	int dunX = 0;
+	int dunY = 0;
+
+	for (int i = 0; i <= steps; i++) {
+		x += Xinc;
+		y += Yinc;
+		if (abs(Xinc) < 1) {
+			dunY = (int)y;
+			if (Xinc < 0)
+				dunX = ceil(x);
+			else
+				dunX = floor(x);
 		} else {
-			d = 2 * dy + dx;
-			dincD = 2 * dy;
-			dincH = 2 * (dx + dy);
-			yincD = -1;
+			dunX = (int)x;
+			if (Yinc < 0)
+				dunY = ceil(y);
+			else
+				dunY = floor(y);
 		}
-		while (!done && (x1 != x2 || y1 != y2)) {
-			if ((d <= 0) ^ (yincD < 0)) {
-				d += dincD;
-			} else {
-				d += dincH;
-				y1 += yincD;
-			}
-			x1++;
-			done = ((x1 != xorg || y1 != yorg) && !Clear(x1, y1));
-		}
-	} else {
-		if (dy < 0) {
-			tmp = y1;
-			y1 = y2;
-			y2 = tmp;
-			tmp = x1;
-			x1 = x2;
-			x2 = tmp;
-			dy = -dy;
-			dx = -dx;
-		}
-		if (dx > 0) {
-			d = 2 * dx - dy;
-			dincD = 2 * dx;
-			dincH = 2 * (dx - dy);
-			xincD = 1;
-		} else {
-			d = 2 * dx + dy;
-			dincD = 2 * dx;
-			dincH = 2 * (dy + dx);
-			xincD = -1;
-		}
-		while (!done && (y1 != y2 || x1 != x2)) {
-			if ((d <= 0) ^ (xincD < 0)) {
-				d += dincD;
-			} else {
-				d += dincH;
-				x1 += xincD;
-			}
-			y1++;
-			done = ((y1 != yorg || x1 != xorg) && !Clear(x1, y1));
-		}
+		if (!Clear(dunX, dunY))
+			return false;
+		if (dunX == x2 && dunY == y2)
+			return true;
 	}
-	return x1 == x2 && y1 == y2;
+	return false;
 }
 
 BOOL LineClear(int x1, int y1, int x2, int y2)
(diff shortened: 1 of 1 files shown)


NiteKat:
I forgot to note in my pull request, this still does not take blocking objects into account (like Cauldrons).

julealgon:
The description seems to indicate 2 distinct behaviors on this one. Can we make it 2 separate changes?

AJenbo:
Apparently this has rounding issues and the following showed improve it:
```cpp
  bool checkLineofSight(int x, int y)
  {
    if (x == game->self().getX() && y == game->self().getY())
      return true;

    float x1 = game->self().getFutureX();
    float y1 = game->self().getFutureY();
    int dx = x - game->self().getFutureX();
    int dy = y - game->self().getFutureY();
    int dunX = 0;
    int dunY = 0;
    int steps = 0;
    bool done = false;

    if (abs(dx) > abs(dy))
      steps = abs(dx);
    else
      steps = abs(dy);

    float Xinc = (float)dx / steps;
    float Yinc = (float)dy / steps;

    for (int i = 0; i <= steps; i++)
    {
      x1 += Xinc;
      y1 += Yinc;
      if (abs(Xinc) < 1)
      {
        if (Xinc < 0)
        {
          dunX = ceil(x1);
          dunY = y1;
        }
        else
        {
          dunX = floor(x1);
          dunY = y1;
        }
      }
      else
      {
        if (Yinc < 0)
        {
          dunX = x1;
          dunY = ceil(y1);
        }
        else
        {
          dunX = x1;
          dunY = floor(y1);
        }
      }
      if (dunX == x && dunY == y)
        return true;
      if (!game->getDungeonPiece(dunX, dunY) ||
        !game->getDungeonPiece(dunX, dunY)->allowsMissile())
        return false;
    }
    return false;
    ```

File: Source/monster.cpp [hunk 1]
```diff
@@ -3648,7 +3648,10 @@ void MAI_Ranged(int i, int missile_type, BOOL special)
 				else
 					M_StartRAttack(i, missile_type, 4);
 			} else {
-				Monst->_mAnimData = Monst->MType->Anims[MA_STAND].Data[md];
+				fx = Monst->_lastx;
+				fy = Monst->_lasty;
+				md = GetDirection(Monst->_mx, Monst->_my, fx, fy);
+				M_CallWalk(i, md);
```
AJenbo, line 3654: Preferably the monster should not have decided to stand in the first place if this is not an appropriate action for it to take.

Do you have a save where this can be replicated?

NiteKat: I'll get a save together with a monster specifically falling into this issue. I don't think it's bad for the AI to fall to this section of the code, the problem was Blizz made this section "just stand there." Ideally, you might do something a bit more than "Walk toward target" since this tends to induce a back and forth pacing (since it waffles between too close, and backed up to the spot that started all the trouble).

I'll get a save setup with this specific scenario though for ya in the morning.

NiteKat:
> Apparently this has rounding issues (irrational numbers not adding up to x.0000 when they are "suppose" to) and the following showed improve it:
> 
> ```c++
>   bool checkLineofSight(int x, int y)
>   {
>     if (x == game->self().getX() && y == game->self().getY())
>       return true;
> 
>     float x1 = game->self().getFutureX();
>     float y1 = game->self().getFutureY();
>     int dx = x - game->self().getFutureX();
>     int dy = y - game->self().getFutureY();
>     int dunX = 0;
>     int dunY = 0;
>     int steps = 0;
>     bool done = false;
> 
>     if (abs(dx) > abs(dy))
>       steps = abs(dx);
>     else
>       steps = abs(dy);
> 
>     float Xinc = (float)dx / steps;
>     float Yinc = (float)dy / steps;
> 
>     for (int i = 0; i <= steps; i++)
>     {
>       x1 += Xinc;
>       y1 += Yinc;
>       if (abs(Xinc) < 1)
>       {
>         if (Xinc < 0)
>         {
>           dunX = ceil(x1);
>           dunY = y1;
>         }
>         else
>         {
>           dunX = floor(x1);
>           dunY = y1;
>         }
>       }
>       else
>       {
>         if (Yinc < 0)
>         {
>           dunX = x1;
>           dunY = ceil(y1);
>         }
>         else
>         {
>           dunX = x1;
>           dunY = floor(y1);
>         }
>       }
>       if (dunX == x && dunY == y)
>         return true;
>       if (!game->getDungeonPiece(dunX, dunY) ||
>         !game->getDungeonPiece(dunX, dunY)->allowsMissile())
>         return false;
>     }
>     return false;
> ```

Note this is from my AI in relation to DAPI, so some of this is not relevant to the DevilutionX code, but it's working on the same principles. If I remember correctly, this is slightly improved, as I was trying to fix a few things with the algorithm. This doesn't have the attempted fix for rounding issues in it though, as I realized that's on another branch I didn't copy from. For right now, I'm going to leave the PR with the rounding issues as it's still "better" when coupled with the AI fix (because while they might occasionally think they can't hit when they can, they will think they can hit when they can't in much fewer scenarios).

Another thing this doesn't check is for objects in the way, but the original game doesn't check that either. So that's something that should be added in to this fix eventually too.

qndel:
new lines for { } make me uncomfortable 😅

NiteKat:
[4tile.zip](https://github.com/diasurgical/devilutionX/files/6566415/4tile.zip)
[lineofsightwithdoor.zip](https://github.com/diasurgical/devilutionX/files/6566416/lineofsightwithdoor.zip)

Two save files, one with Fire Clan archers shooting at the player with most not able to hit through a door (common tactic used to minimize damage, as the player can hit all the archers from where they are standing), and the other with a Flesh clan 4-tiled (standing still, not attacking, not moving, and free to be attacked by the player safely)

Source/monster.cpp
```diff
@@ -4912,83 +4915,50 @@ BOOL CheckNoSolid(int x, int y)
...
-			xincD = 1;
+	if (abs(dx) > abs(dy))
+		steps = abs(dx);
+	else
+		steps = abs(dy);
+
+	if (steps == 0)
+		return false;
+
+	float Xinc = (float)dx / steps;
```
glebm, line 4931: Can we avoid floating point math for this?

Context: For the past 6 months I've been trying to make DevilutionX run well on a device without an FPU (RG99 jz4725).

https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm

NiteKat: > Can we avoid floating point math for this?
> 
> Context: For the past 6 months I've been trying to make DevilutionX run well on a device without an FPU (RG99 jz4725).
> 
> https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm

Bresenham's line algorithm is what original Diablo uses and the the cause of one of the "4 tile" scenarios, specifically the one where monsters shoot into walls/corners endlessly. Bresenham's is a special case of DDA that only draws in one direction, but an arrow's path is NOT the same path in both directions between two points. Ajenbo and I discussed this over discord, so below are example lines drawn by Bresenham's and a proper line algorithm to highlight the difference.

Bresenham's drawing from origin (5, 1) to (0, 0) will draw it as {0x0, 1x0, 2x0, 3x1 4x1, 5x1}, but that's a line from (0, 0) to (5, 1). A line from (5, 1) to (0, 0) should actually be {5x1, 4x1, 3x1, 2x1, 1x1, 0x0} if I'm doing the DDA in this PR correctly in my head. Notice how (2, 1) is being checked in the PRs algorithm while it is not in Bresenham's. If an arrow dies on tile (2, 1) but not (2, 0), a monster standing at (5, 1) will think it can hit (2, 0) while it cannot. Same with the difference between (1, 1) and (1, 0).

I think the line it draws is also wrong in general, as Diablo's missile physics aim to have missiles "just barely" enter the target tile, i.e. they get hit the very "corner" of what would be considered "in the tile" so in a case like this, from (0, 0) to (5, 1) where one axis changes by 1 and the other by more, the axis that changes by 1 is going to change on the very last step. Bresenham's was changing that axis too early, as shown in the example above.

You could maybe do away with the float math and division with ints and bit shifting like other parts of Diablo does? Not sure how that might impact math with rounding though, but this already has issues with rounding into irrational decimals and not stepping when it should in the shorter axis.

PR #1014, opened 2021-02-18
unify checks for single and multiplayer

qndel:
There were all kinds of checks for single/multi games
gbMaxPlayers == 1
gbMaxPlayers != 1
gbMaxPlayers <= 1
gbMaxPlayers > 1
etc.


AJenbo:
Looks like you broke the pause function in single player :)

AJenbo:
Changed it to use a `bool` instead since we are trying to reduce the amount of preprocessor code and migrate to more C`++` instead.

PR #1015, opened 2021-02-18
Unify calculation of gold icon

AJenbo:


qndel:
Nice :D always cool to see so much code get trashed ;)

AJenbo:
The last bit of Hellfire code is in how gold pick up is handled, having some issues so needed something that wasn't so cluttered to look at :)

PR #1027, opened 2021-02-20
numpad potions + chat

qndel:
This lets you drink potions/write numbers on chat with numpad if numlock is enabled


SourceX/miniwin/misc_msg.cpp
```diff
@@ -205,6 +205,8 @@ static int translate_sdl_key(SDL_Keysym key)
 		return DVL_VK_NUMPAD8;
 	case SDLK_KP_9:
 		return DVL_VK_NUMPAD9;
+	case SDLK_KP_000:
+	case SDLK_KP_00:
```
AJenbo, line 209:
```suggestion
#ifndef USE_SDL1
	case SDLK_KP_000:
	case SDLK_KP_00:
#endif
```

PR #1031, opened 2021-02-21
Correct missing hellfire level mapping in leveltype

AJenbo:
Hellfire never added the 2 new dungeon types, instead, it makes additional checks on currlevel. This is confusing and messy, so cleaned it up.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -43,7 +43,6 @@ BOOLEAN UseBarbarianTest;
 BOOLEAN UseMultiTest;
 int sgnTimeoutCurs;
 char sgbMouseDown;
-int color_cycle_timer;
 int ticks_per_sec = 20;
 WORD tick_delay = 50;
 /** Game options */
@@ -1474,17 +1473,10 @@ void LoadLvlGFX()
 		pSpecialCels = LoadFileInMem("Levels\\TownData\\TownS.CEL", NULL);
 		break;
 	case DTYPE_CATHEDRAL:
-		if (currlevel < 21) {
-			pDungeonCels = LoadFileInMem("Levels\\L1Data\\L1.CEL", NULL);
-			pMegaTiles = LoadFileInMem("Levels\\L1Data\\L1.TIL", NULL);
-			pLevelPieces = LoadFileInMem("Levels\\L1Data\\L1.MIN", NULL);
-			pSpecialCels = LoadFileInMem("Levels\\L1Data\\L1S.CEL", NULL);
-		} else {
-			pDungeonCels = LoadFileInMem("NLevels\\L5Data\\L5.CEL", NULL);
-			pMegaTiles = LoadFileInMem("NLevels\\L5Data\\L5.TIL", NULL);
-			pLevelPieces = LoadFileInMem("NLevels\\L5Data\\L5.MIN", NULL);
-			pSpecialCels = LoadFileInMem("NLevels\\L5Data\\L5S.CEL", NULL);
-		}
+		pDungeonCels = LoadFileInMem("Levels\\L1Data\\L1.CEL", NULL);
+		pMegaTiles = LoadFileInMem("Levels\\L1Data\\L1.TIL", NULL);
+		pLevelPieces = LoadFileInMem("Levels\\L1Data\\L1.MIN", NULL);
+		pSpecialCels = LoadFileInMem("Levels\\L1Data\\L1S.CEL", NULL);
 		break;
 	case DTYPE_CATACOMBS:
 		pDungeonCels = LoadFileInMem("Levels\\L2Data\\L2.CEL", NULL);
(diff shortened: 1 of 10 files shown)


Source/diablo.cpp
```diff
@@ -1142,43 +1141,42 @@ void CreateLevel(lvl_entry lvldir)
...
-			LoadRndLvlPal(DTYPE_NEST);
-		}
 		break;
 	case DTYPE_HELL:
 		CreateL4Dungeon(glSeedTbl[currlevel], lvldir);
 		InitL4Triggers();
 		Freeupstairs();
-		LoadRndLvlPal(DTYPE_HELL);
+		break;
+	case DTYPE_NEST:
```
ChaosMarc, line 1165: Not sure about the code style of this project as I haven't committed anything yet, but couldn't the cases DTYPE_CAVES and DTYPE_NEST as well as DTYPE_CATHEDRAL and DTYPE_CRYPT be merged as they're identical?
```
	case DTYPE_CATHEDRAL:
	case DTYPE_CRYPT:
		CreateL5Dungeon(glSeedTbl[currlevel], lvldir);
		InitL5Triggers();
		Freeupstairs();
		break;
	case DTYPE_CAVES:
	case DTYPE_NEST:
		CreateL3Dungeon(glSeedTbl[currlevel], lvldir);
		InitL3Triggers();
		Freeupstairs();
		break;
```

Source/loadsave.cpp
```diff
@@ -931,6 +931,17 @@ void LoadDroppedItems(LoadHelper &file, size_t savedItemCount)
 	}
 }
 
+int getHelfireLevelType(int type)
```
ChaosMarc, line 934: getHel **l** fireLevelType

Source/diablo.cpp
DakkJaniels: I think it makes sense not to combine them, so the level generation can be separated and modified separately, if not now, then in the future.

ChaosMarc: That's fine with me too, but shouldn't we split the cases in [Source/trigs.cpp](https://github.com/diasurgical/devilutionX/pull/1031/files#diff-4136c23fd95ecf1a6d9f9fafd598072a38ccc65de4ff352b550dc44758099cda) then too?

DakkJaniels: Probably makes sense.

AJenbo: Going to go with the combined condition :) It feels like the right choice for staying DRY.

Source/objects.cpp
```diff
@@ -1570,9 +1591,7 @@ void UpdateFlameTrap(int i)
 		if (Objects[i]._oVar4 != 0)
 			ActivateTrapLine(Objects[i]._otype, Objects[i]._oVar1);
 	} else {
-		int damage[4] = { 6, 8, 10, 12 };
-
-		int mindam = damage[leveltype - 1];
+		int mindam = 4 + leveltype * 2;
```
ephphatha, line 1594:
```suggestion
		int mindam = 4 + static_cast<int>(leveltype) * 2;
```

Using an enum as a numeric value, might as well be explicit about the conversion. This also means that traps are doing more damage on hive/crypt than hellfire does but I assume that's an acceptable/intended change.

Source/objects.h
```diff
@@ -170,7 +170,15 @@ struct Object {
 	 */
 	[[nodiscard]] constexpr bool IsBarrel() const
 	{
-		return IsAnyOf(_otype, _object_id::OBJ_BARREL, _object_id::OBJ_BARRELEX);
+		return IsAnyOf(_otype, _object_id::OBJ_BARREL, _object_id::OBJ_BARRELEX, _object_id::OBJ_POD, _object_id::OBJ_PODEX, _object_id::OBJ_URN, _object_id::OBJ_URNEX);
+	}
+
+	/**
+	 * @brief Check if this object contains explosives or caustic material)
```
ephphatha, line 177:
```suggestion
	 * @brief Check if this object contains explosives or caustic material
```

Source/themes.cpp
```diff
@@ -669,7 +669,7 @@ void Theme_Treasure(int t)
 					CreateRndItem({ xp, yp }, false, false, true);
 					ItemNoFlippy();
 				}
-				if (rv >= treasureType - 2 && leveltype != DTYPE_CATHEDRAL) {
+				if (rv >= treasureType - 2 && leveltype != DTYPE_CATHEDRAL && leveltype != DTYPE_CRYPT) {
```
ephphatha, line 672:
```suggestion
				if (rv >= treasureType - 2 && IsNoneOf(leveltype, DTYPE_CATHEDRAL, DTYPE_CRYPT)) {
```

ephphatha (commented):
Nits from reading the source, I'll have a play later this evening. Looks good though, only minor gameplay changes that likely wont be noticeable anyway. The object layout diverging is interesting since it doesn't look like the structure of that code has changed in a meaningful way but I didn't look into it too deeply.

AJenbo:
> The object layout diverging is interesting since it doesn't look like the structure of that code has changed in a meaningful way but I didn't look into it too deeply.

The reason for diverge is that theme room is no longer being initialized (was being discarded any way), this causes the RNG to diverge after that point.

Source/objects.cpp
AJenbo: This was a bit of a mistake on my part. Since i read this as meaning Crypt don't have traps, and forgot about the nest :sweat_smile: 
```cpp
		if (leveltype > DTYPE_CATHEDRAL)
			AddChestTraps();
```
But this only means that they are less frequent since  they can still be added by `AddObjTraps()`

AJenbo: Since the burning crosses never spawn in the Nest or Crypt it is a bit up to our own interpretation how they should behave for custom levels. I did go ahead and revert this and instead mirror Cave and Hell since that is what the balance seams to mostly match.

AJenbo:
@ephphatha I have addressed your comments as well as some other things that I noticed on further inspection.

Additionally I have:
- Made the first commit create identical levels to before this PR
- Made a second commit that fully disables Theme rooms for Nest and Crypt, and there by causing the diff
- Fixed a vanilla bug in themes.cpp
- Got side tracked working on the next PR :)

ephphatha (approved):
Looks fine from a quick run through one of each level type :D

AJenbo:
@ephphatha thanks for reviewing and testing, this is a big mile stone to get done in my book <3

PR #1032, opened 2021-02-21
Fix loading order of poison water palette

### LINKED ISSUE #593: Bad pallet at end of loading poisoned water
**Describe**
When entering the poisoned water level blue colors appear in the image as the screen fades.
[single_0.zip](https://github.com/diasurgical/devilutionX/files/4176447/single_0.zip)

**To Reproduce**
Load attached save game and walk in to the poisoned water level

**Expected behavior**
The colors should persist as the scree fade.

**Screenshots**
image

AJenbo:
Fixes #593

This also fixes the issues with lave/water not animating when behind something that is using transparency

### DIFF
--- a/Source/palette.cpp
+++ b/Source/palette.cpp
@@ -81,14 +81,19 @@ void palette_init()
 }
 
 /**
- * Generate lookup table for transparency
+ * @brief Generate lookup table for transparency
  *
  * This is based of the same technique found in Quake2.
  *
  * To mimic 50% transparency we figure out what colours in the existing palette are the best match for the combination of any 2 colours.
  * We save this into a lookup table for use during rendering.
+ *
+ * @param palette The colors to operate on
+ * @param skipFrom Do not use colors between this index and skipTo
+ * @param skipTo Do not use colors between skipFrom and this index
+ * @param toUpdate Only update the first n colors
  */
-void GenerateBlendedLookupTable()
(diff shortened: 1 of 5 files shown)


Source/palette.cpp
```diff
@@ -359,6 +398,7 @@ void palette_update_quest_palette(int n)
 	}
 	ApplyGamma(system_palette, logical_palette, 32);
 	palette_update();
+	GenerateBlendedLookupTable(logical_palette, 1, 31, 32 - n);
```
AJenbo, line 401:
```suggestion
	GenerateBlendedLookupTable(logical_palette, 1, 31, 32 - n); // Possible optimization would be to only update color 0 as only menus has transparency calves this level

```

Source/palette.cpp
```diff
@@ -359,6 +398,8 @@ void palette_update_quest_palette(int n)
 	}
 	ApplyGamma(system_palette, logical_palette, 32);
 	palette_update();
+	GenerateBlendedLookupTable(logical_palette, 1, 31, 32 - n); // Possible optimization would be to only update color 0 as only menus has transparency calves this level
```
AJenbo, line 401:
```suggestion
	GenerateBlendedLookupTable(logical_palette, 1, 31, 32 - n); // Possible optimization would be to only update color 0 as only the UI can overlap with transparency in this quest
```

FluffyQuack (commented):
I have looked at these changes and they have attained the Fluffy stamp of approval! I do not see any indication that they break the rendering of cows, which is the most important of all things.

PR #1037, opened 2021-02-21
fix slain hero (and other stuff?)

qndel:


AJenbo (approved):
Nice this fixes the issue :)

Issue #896, opened 2020-10-24
Build from source failed on Fedora 32 with error "/usr/bin/ld: cannot find -lstdc++"


sjmalice:
I resolved the error by installing the "libstdc++-static" package. Could we add this to the dependencies for Fedora?

AJenbo:
This appears to be a general issue with Fedora it shoudn't require statics, please see https://github.com/diasurgical/devilutionX/pull/756 and https://github.com/diasurgical/devilutionX/pull/891.

zboszor:
This also happens on Fedora 33 with no libstdc++-static package.

AJenbo:
Can you test this PR?
https://github.com/diasurgical/devilutionX/pull/1703

sjmalice:
libstdc++-staticlibstdc++-staticFinally getting around to trying this again. I've pulled the latest changes and removed libstdc++-static. But now when I try to build from source, it fails with error "No rule to make target '/usr/lib64/libSDL2main.a', needed by 'devilutionx'.  Stop."

AJenbo:
Looks like it is still attempting to statically link SDL2. Maybe just to see what happens could you try building for Sdl1.2?

sjmalice:
OK. So since I'm not entirely sure what I'm doing, please bare with me, here are all the steps I tried this morning.

1. I edited line 28 of "CMakeLists.txt" to set option "USE_SDL1" to "ON". 
2. Then I installed packages with the command "sudo dnf install SDL-devel SDL_ttf-devel SDL_mixer-devel".
3. Cleared the build folder and ran the commands to build from source again.

Unfortunately, now the build fails with the same error as before. "/usr/bin/ld: cannot find -lstdc++" :(

AJenbo:
Sorry, but I think we need someone who knows how things are supposed to work on Fedora.
It's a bit odd since this works on most other Linux distributions, Mac and windows.

yuripourre:
@sjmalice since Fedora 29 for some reason libc cannot be compiled using the same packages as before. Currently a workaround is to install `libstdc++-static`.

```
sudo dnf install -y libstdc++-static
```

yuripourre:
@AJenbo I think that's a bug on Fedora packages. It used to work a few years ago.

AJenbo:
Someone should probably report it to fedora maintenance

PR #1040, opened 2021-02-22
Update Fedora's dependencies

runlevel5:
Without `libstdc++-static`, the compilation would yield following error:

```
ld: cannot find -lstdc++
```

Tested and verified on Fedora 33


Trihedraf:
I just made a fresh VM and it did need that to successfully build, but also it needed g++ as well. I would suggest changing it to this.

`sudo dnf install cmake g++ glibc-devel SDL2-devel SDL2_ttf-devel SDL2_mixer-devel libsodium-devel libasan libubsan libstdc++-static`

AJenbo:
https://github.com/diasurgical/devilutionX/issues/896

PR #1041, opened 2021-02-22
no idea whats going on

qndel:
check if this is any good, I have no idea what's the expected behavior with staff of mana/charged bolt while moving saves at this point 😅


Source/pack.cpp
```diff
@@ -136,7 +136,7 @@ void UnPackItem(PkItemStruct *is, ItemStruct *id)
 	if (idx == 0xFFFF) {
 		id->_itype = ITYPE_NONE;
 	} else {
-		if (!gbIsHellfireSaveGame) {
+		if (gbIsHellfireSaveGame && !gbIsHellfire) {
```
AJenbo, line 139:
```suggestion
		if (gbIsHellfireSaveGame) {
```

Source/loadsave.cpp
```diff
@@ -183,7 +183,7 @@ static void LoadItemData(ItemStruct *pItem)
 	tbuff += 1; // Alignment
 	CopyInt(tbuff, &pItem->_iStatFlag);
 	CopyInt(tbuff, &pItem->IDidx);
-	if (!gbIsHellfireSaveGame) {
+	if (!gbIsHellfireSaveGame && gbIsHellfire) {
```
AJenbo, line 186: Mapping is to the ids in the static list and do not change depending if hellfire behavior is enabled or not.

```suggestion
	if (!gbIsHellfireSaveGame) {
```

qndel:
oh well, other than that I only added numbers to items array + a check if it's the sorc's starting staff before just setting charges to 40/18

Source/itemdat.cpp
```diff
@@ -11,174 +11,174 @@ DEVILUTION_BEGIN_NAMESPACE
...
+/*003-IDI_WARRCLUB   */ {  IDROP_NEVER,   ICLASS_WEAPON, ILOC_ONEHAND,     ICURS_CLUB,                        ITYPE_MACE,   UITYPE_SPIKCLUB,    "Club",                        NULL,          1,          20,       1,       6,      0,      0,       0,       0,       0, ISPL_NONE,         IMISC_NONE,      SPL_NULL,        FALSE,       20,        20 },
+/*004-IDI_ROGUE      */ {  IDROP_NEVER,   ICLASS_WEAPON, ILOC_TWOHAND,     ICURS_SHORT_BOW,                   ITYPE_BOW,    UITYPE_NONE,        "Short Bow",                   NULL,          1,          30,       1,       4,      0,      0,       0,       0,       0, ISPL_NONE,         IMISC_NONE,      SPL_NULL,        FALSE,      100,       100 },
+/*005-IDI_SORCEROR   */ {  IDROP_NEVER,   ICLASS_WEAPON, ILOC_TWOHAND,     ICURS_SHORT_STAFF,                 ITYPE_STAFF,  UITYPE_NONE,        "Short Staff of Charged Bolt", NULL,          1,          25,       2,       4,      0,      0,       0,      20,       0, ISPL_NONE,         IMISC_STAFF,     SPL_CBOLT,       FALSE,      520,       520 },
```
AJenbo, line 19: the list should match hellfire, if we start mixing it gets reallly complicated

AJenbo:
probably dont try to solve hellfire multiplayer heros before the game kows what that is :)
there is no header in mp saves so all saves looks like diablo saves for now.

PR #1044, opened 2021-02-23
update

ivan89el:
rus


ivan89el:
sorry

AJenbo:
No worries, translation support is still not quite ready, but if you are eager to get it done then here is a preliminary translation file you can work on.

[devilutionx-ru.zip](https://github.com/diasurgical/devilutionX/files/6030229/devilutionx-ru.zip)

Note that you will be able to use proper Cyrillic so no need for romanization :)

AJenbo:
You could probably use https://www.lexilogos.com/keyboard/russian_conversion.htm to convert the work you have done back in to proper Cyrillic to save some time.

ivan89el:
This is my inability to use github. Translation I'm almost done, translated:
automap.cpp
control.cpp
cursor.cpp
diablo.cpp
error.cpp
gamemenu.cpp
help.cpp
inv.cpp
itemdat.cpp
items.cpp
monstdat.cpp
monster.cpp
objects.cpp
quests.cpp
setmaps.cpp
spelldat.cpp
stores.cpp
textdat.cpp
towners.cpp
trigs.cpp
Almost the entire game!

galaxyhaxz:
If you'd like a more concise and easy way to translate the game for future use, consider using this. This file is a string dump from the PSX port with 5 languages. Just add `Rus=""` beneath each string and your translation for it. Whenever it is added in the future, it will be easy to reference!!

DOWNLOAD: [maintxt.txt](https://github.com/diasurgical/devilutionX/files/6031812/maintxt.txt)

ivan89el:
@AJenbo
@galaxyhaxz
What do you think is the best option?

galaxyhaxz:
Anders already has it setup to use the .PO file, so I'd say go with that.

I'm using the other method for my project as it's better in the long term. You can create name keys for each string and dynamically look them up like Diablo II does. For example:

```
strMod1: "Enhanced damage %d"
strMod2: "%d to Armor"
...
GetStringMod("strMod%d", ModId);
```

ivan89el:
Thank you very much!(I will use the .po file) Sorry my english.

galaxyhaxz:
Great!! If you are in need of a font, someone made a russian exocet font for Diablo II, as there was never an official translation.

ivan89el:
@galaxyhaxz 
Correct Cyrillic will not work? That's what AJenbo said.

galaxyhaxz:
As it sits the current font code only supports Windows-1252, and converts 0x80+ characters. That mod I linked stores the Russian font in unicode then converts it to a 256-character bitmap font. Shouldn't be too hard to eventually implement :P

ivan89el:
Yes, it will not be difficult. Thank you again.

AJenbo:
We are currently working on a new Cyrillic bitmap font and implement Unicode rendering directly. So for you it will be easy when we are done :)

> I'm using the other method for my project as it's better in the long term. You can create name keys for each string and dynamically look them up like Diablo II does. For example:

PO uses the original text as the message id, I'm not sure I get how D2 way of handling translation is better?

```c
msgid "Sound"
msgstr "Lyd"
...
_("Sound");
_(string);
```

One problem I see with the D2 format is that all languages are stored in the same file making it often conflict if multiple people work on it simultaneously. As well with PO you get an industry-tested format with support for things like Russian plural form, existing translation tools etc.

galaxyhaxz:
> One problem I see with the D2 format is that all languages are stored in the same file making it often conflict if multiple people work on it simultaneously.

_The PSX port of D1 stores everything in one file_, **D2 does not**. Sit back and relax, you are about to go for a ride on a D2 mod making journey. I'll try to explain it the best I can:
```c
/////////// quest strings
SECTION: Quest
NAME: Akara
"First quest string here" // strQuestAkara1
"another string here" // strQuestAkara2

#include "monsters.txt" // monster strings
#include "gossip.txt" // NPC gossip
/////////// item strings
SECTION: Item
NAME: Mod
"Enhanced damage %d" // strItemMod1
"Enhanced defense %d" // strItemMod2
NAME: Helms
"Skull Cap" // strItemHelms1
"Great Helmet" // strItemHelms2
NAME: Swords
"Broad Sword" // strItemSwords1
```
This is the compiled into a unicode binary with a key generated for each string (e.g. `strSectionNameX`). Breaking down the key allows us to fetch the string by building the key real-time:
```c
//char *GetString(int Section, int Name, int ID);
char *QuestString = GetString(STR_QUEST, GetNPC(), Quest->Id);
char *ItemMod = GetString(STR_ITEM, "Mod", Item->mod);
char *ItemName = GetStrFromStr(gTablesItemStr); // no need to build key, specified by mod maker
```
There are a few advantages here over PO files.
- The game doesn't have to be re-compiled after modifying strings.
- When changing the English string no other languages have to be changed, since the reference is a key and not the string itself.
- It *truly* isolates strings from code
- We can map out strings in the order of enums, which greatly simplifies code for getting the string.

That last one is probably the biggest of all. For example, Griswold has a total of 12 gossip strings. To randomly fetch a gossip for a specific NPS, all we'd have to do is call `GetGossip(TownerId, min, max)` where `GetGossip` will build a key `strGossipTownerX` and lookup the string for us.

Anyway, that's all I have. I realize there will be disagreements and that's fine, it's open source for all hands-taking. But for my own project I will use this method. Diablo II is a tried-and-true game for modmaking and has had an active community for 20 years because of how easy it is for anyone to jump right in and modify game data/strings. This is coming from someone who used to be a D2 mod maker at the age of 14.

AJenbo:
Probably you could make a tool that generates the text files from PO files and get the best of both.

galaxyhaxz:
I fancy that idea! Easy translations outside the game, optimized for management inside (:

AJenbo:
@ivan89el you can now start working on a proper Russian translation:
https://github.com/diasurgical/devilutionX/blob/master/Translations/devilutionx.pot

We are currently working on adding a cyrillic font

PR #1057, opened 2021-02-25
fix text scroll speed

qndel:
Fixes wrong scroll speeds for very long texts

### DIFF
--- a/SourceX/soundsample.cpp
+++ b/SourceX/soundsample.cpp
@@ -104,9 +104,9 @@ int SoundSample::GetLength()
 		bytePerSample = 1;
 	}
 
-	//alen * 1000 can overflow a 32bit int so using a 64bit one - after division it should still fit inside 32 bits.
-	Uint64 tmp = chunk->alen * 1000 / (frequency * channels * bytePerSample);
-	return (Uint32)tmp;
+	Uint64 ms = 1000; // milliseconds, 64bit to avoid overflow when multiplied by alen
+	Uint32 bps = frequency * channels * bytePerSample; // bytes per second
+	return (Uint32)(chunk->alen * ms / bps);
 };
 
 } // namespace dvl
--- a/SourceX/soundsample.cpp
+++ b/SourceX/soundsample.cpp
@@ -104,7 +104,9 @@ int SoundSample::GetLength()
 		bytePerSample = 1;
 	}
 
-	return chunk->alen * 1000 / (frequency * channels * bytePerSample);
+	//alen * 1000 can overflow a 32bit int so using a 64bit one - after division it should still fit inside 32 bits.
+	Uint64 tmp = chunk->alen * 1000 / (frequency * channels * bytePerSample);
+	return (Uint32)tmp;
 };
 
 } // namespace dvl


AJenbo:
Wow, that's a high value right there 😅

Chance4us:
Unfortunately, the change has no effect.

galaxyhaxz:
The original code I had used a float, would that fix it for you?
```c
float tmp = chunk->alen / (frequency * channels * bytePerSample);
return (Uint32)(tmp * 1000);
```

Chance4us:
Yes. thank you ;-)

galaxyhaxz:
BTW the reason this fix doesn't work is because it still does 32-bit math and stores it inside a 64-bit integer. You have to cast it so it does 64-bit math as well:
```c
Uint64 tmp = chunk->alen * 1000 / (frequency * channels * bytePerSample);
// change to
Uint64 tmp = (Uint64)chunk->alen * 1000 / (frequency * channels * bytePerSample);
// alternative
Uint64 ms = 1000;
Uint64 tmp = chunk->alen * ms / (frequency * channels * bytePerSample);
```

AJenbo:
How about this to avoid any casting?
```cpp
Uint32 rate = frequency * channels * bytePerSample / 1000; // bytes per ms
return chunk->alen / rate;
```

galaxyhaxz:
> How about this to avoid any casting?

Sadly that still won't work. Diablo uses `22050 freq, 1 channel, 8 bit` for basically everything. That means 22050 / 1000 = 22.05, so you'd have to cast it into a float in this case.

I wish there was a way to avoid the ugly cast, but there really isn't. Either you multiply by 1000 and go over 32 bits (`int64`), or you divide which leaves a decimal and multiply by 1000 (`float/double`).

AJenbo:
I'm aware that we would lose some precision, but we are talking less than 23ms over a 10 seconds clip. Still, let's just go for the high-precision version since it's only triggered at the start of dialogs :)

galaxyhaxz:
: I'm aware that we would lose some precision, but we are talking less than 23ms over a 10 seconds clip. Still, let's just go for the high-precision version since it's only triggered at the start of dialogs :)

In my project ive implemented the lore from the PSX. The longest one is 4 minutes 47 seconds, the speaker will desync towards the end if you aren't very precise. Hell even the current implementation in devilution X isn't precise enough, because you changed my PR to milliseconds/pixel and you can get a decimal precision that way.

SourceX/soundsample.cpp
```diff
@@ -104,7 +104,9 @@ int SoundSample::GetLength()
 		bytePerSample = 1;
 	}
 
-	return chunk->alen * 1000 / (frequency * channels * bytePerSample);
+	//alen * 1000 can overflow a 32bit int so using a 64bit one - after division it should still fit inside 32 bits.
+	Uint64 tmp = chunk->alen * 1000 / (frequency * channels * bytePerSample);
+	return (Uint32)tmp;
```
AJenbo, line 109:
```suggestion
	Uint64 ms = 1000; // milliseconds, 64bit to avoid overflow when multiplied by alen
	Uint32 bps = frequency * channels * bytePerSample; // bytes per second
	return (Uint32)(chunk->alen * ms / bps);
```

galaxyhaxz:
A really cheesy way of avoiding all casting and sticking to raw 32-bit, this will work with audio up to 32-minutes (16 for dual channel)
```
Uint32 rate = frequency * channels * bytePerSample / 10;
return chunk->alen * 100 / rate;
```

AJenbo:
Well don't forget that audio is resampled to the output (22050hz, 2 channel, 16 bit by default) during loading, so it's actually only 8 minutes on most systems :)

PR #1075, opened 2021-02-28
Extract 'ItemStruct.isEmpty()' function

julealgon:
This PR extracts a function to check whether a given item is empty or not and reuses it in all places where such check is made.

This improves readability, hides some of the nasty implementation details and makes it easier to reuse this behavior in the future.


AJenbo:
How about making this a struct method instead os that you can simply call `item.isEmpty()`?

julealgon:
> How about making this a struct method instead os that you can simply call item.isEmpty()?

Done.

PR #1083, opened 2021-03-01
automap: Migrate away from gpBuffer/gpBufEnd

glebm:
Migrating away from using these globals will simplify transition to using SDL in rendering, rendering UI to a separate buffer, etc.

### DIFF
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -330,14 +330,15 @@ void Cl2DrawLightTbl(int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth, char 
 void Cl2DrawLight(int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth);
 
 /**
- * @brief Draw a line on the back buffer
+ * @brief Draw a line in the target buffer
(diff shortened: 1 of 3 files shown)


Source/engine.h
```diff
@@ -330,14 +330,18 @@ void Cl2DrawLightTbl(int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth, char
...
  * @param x0 Back buffer coordinate
  * @param y0 Back buffer coordinate
  * @param x1 Back buffer coordinate
  * @param y1 Back buffer coordinate
- * @param col Color index from current palette
+ * @param color_index Color index from current palette
  */
-void DrawLine(int x0, int y0, int x1, int y1, BYTE col);
+void DrawLineTo(CelOutputBuffer out, int x0, int y0, int x1, int y1, BYTE color_index);
+inline void DrawLine(int x0, int y0, int x1, int y1, BYTE color_index) {
```
AJenbo, line 342: What's still using this at this point?

glebm: Nothing, removed

PR #1085, opened 2021-03-01
Enginify `trans_rect`

glebm:
Moves `trans_rect` to `engine` as `DrawHalfTransparentRectTo` and makes it operate on `CelOutputBuf`.

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -555,6 +555,43 @@ void DrawLineTo(CelOutputBuffer out, int x0, int y0, int x1, int y1, BYTE color_
 	}
 }
 
+static void DrawHalfTransparentBlendedRectTo(CelOutputBuffer out, int sx, int sy, int width, int height)
+{
+	BYTE *pix = out.at(sx, sy);
+	for (int row = 0; row < height; row++) {
+		for (int col = 0; col < width; col++) {
+			*pix = paletteTransparencyLookup[0][*pix];
+			pix++;
+		}
+		pix += out.line_width - width;
+	}
+}
+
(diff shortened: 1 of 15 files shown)


Source/engine.cpp
```diff
@@ -555,6 +555,32 @@ void DrawLineTo(CelOutputBuffer out, int x0, int y0, int x1, int y1, BYTE color_
 	}
 }
 
+void DrawHalfTransparentRectTo(CelOutputBuffer out, int sx, int sy, int width, int height)
+{
+	int row, col;
+	BYTE *pix = out.at(sx, sy);
+
+	if (sgOptions.bBlendedTransparancy) { // Blended
```
AJenbo, line 563: maybe we should split this in to two function?

AJenbo (approved):

Source/engine.cpp
glebm: Done

Source/engine.h
```diff
@@ -340,6 +340,23 @@ void Cl2DrawLight(int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth);
...
 void DrawLineTo(CelOutputBuffer out, int x0, int y0, int x1, int y1, BYTE color_index);
 
+
+/**
+ * Draws a half-transparent rectangle by blacking out odd pixels on odd lines,
+ * even pixels on even lines.
+ *
+ * If blended transparency is enabled, uses `paletteTransparencyLookup` table instead
+ * of blacking pixels out.
+ *
```
AJenbo, line 350: This is implementation-specific and global transparency handling is described elsewhere already.
```suggestion
 *
```

AJenbo, line 350: This is implementation-specific and global transparency handling is described elsewhere already.
```suggestion
 *
```

PR #1092, opened 2021-03-03
Add barmaid's unused/missing transcriptions

yuripourre:
It really deserves a review since English is not my native language.

Thanks to @galaxyhaxz I could find better transcriptions to almost all of the sounds.

For the following transcriptions I tried my best:
TSFX_BMAID28
TSFX_BMAID29
TSFX_BMAID30


galaxyhaxz:
These are all from the cut quests and thus should be commented out, but this is great to have! There also exists these quests from other NPCs too, as well as Tremain's full 7 speeches.

Source/textdat.cpp
```diff
@@ -645,6 +645,38 @@ const TextDataStruct alltext[] = {
...
+	    1, 5, TSFX_BMAID26 },
+	{ "I know that my grandmother’s story seems strange but the graveyard does have many mysteries surrounding it. It couldn’t hurt you to put something there see what happens, could it? Maybe you will find out something that will help us all."
+	    1, 5, TSFX_BMAID28 },
+	{ "My grandmother had a dream about you last night, she said that in her dream you used your bare hands to defeat one of those fallen monsters that lurk under the church. When it died it exploded into a fabulous treasure! Although she was a bit frightened by her dream, she was able to describe the beast quite clearly to me."
+	    1, 5, TSFX_BMAID29 },
+	{ "My grandmother had that exact same dream again. She said you are quite brave and seemed very pleased with your prize."
+	    1, 5, TSFX_BMAID30 },
+	{ "I really don’t know much about priest Tremain, he never visits the tavern preferring to keep into the company of Pepin and Cain perhaps it’s because they two have more scholarly pursuits."
```
galaxyhaxz, line 678: `it’s because they two have more scholarly pursuits`

Based on this speech, I hear it as `they too have`. In reference that "they also" have scholarly pursuits, whereas for two you'd have to say `because those two have more scholarly pursuits`.

galaxyhaxz: For reference, you can see how the creator of Diablo evolution has dubbed these here: https://diablo-evolution.net/index.php?pageid=q1b10

```
I really don't know much about the priest Tremain. He never visits the tavern, preferring to keep to the company of Pepin and Cain. Perhaps it is because they, too, have more scholarly pursuits.
```

yuripourre: @galaxyhaxz 
> `it’s because they two have more scholarly pursuits`
> 
> Based on this speech, I hear it as `they too have`. In reference that "they also" have scholarly pursuits, whereas for two you'd have to say `because those two have more scholarly pursuits`.

Oh, makes much more sense, I will update it!

Thank you for the link! Who could tell that there is much more story to be told by Diablo 1?

galaxyhaxz: Yes, there are many cut quests (wirts treasure, infested cellar, map of doom, shadowfang, grimspike, andariel, horizon). It's a shame they weren't given another year to work on the game or it's probably be much different.

And maybe the code would look like it was written by adults xD

yuripourre:
> Make sure that you match enums, texts, and sounds or it will break the game, this is what was missing for the cow sounds:
> [44a4224](https://github.com/diasurgical/devilutionX/commit/44a4224aad245db125b90aa3ca47595c9a5db5c5)
> 
> Also this should be done as comments or it will cause the game to load the additional sounds into memory.

Oh, right! Now I get what you meant

AJenbo:
@yuripourre could you also add the corresponding labels to `_speech_id` (must match the order of the alltext array)

yuripourre:
@AJenbo Just added the _speech_ids

AJenbo (approved):
indentation is a bit wonkey

PR #1101, opened 2021-03-03
Add coexistent Hellfire .desktop resources

M0Rf30:
This also add a light Italian localization for Comment field


AJenbo:
nice :)

PR #1102, opened 2021-03-04
casting animations in town

qndel:
```cpp
		if (plr[pnum]._pVar8 > plr[pnum]._pSFrames) {
			StartWalkStand(pnum);
			ClearPlrPVars(pnum);
			return TRUE;
		}
	} else if (plr[pnum]._pAnimFrame == plr[pnum]._pSFrames) {
```
not sure what's the purpose of this part of code but it causes spell animation to loop in town, works fine after removing so ...
no more healing without any effect in town ;)
inspired by @mewmew

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -287,25 +287,16 @@ void LoadPlrGFX(int pnum, player_graphic gfxflag)
 			pAnim = (BYTE *)p->_pHAnim;
 			break;
 		case PFILE_LIGHTNING:
-			if (leveltype == DTYPE_TOWN) {
-				continue;
-			}
 			szCel = "LM";
 			pData = p->_pLData;
 			pAnim = (BYTE *)p->_pLAnim;
 			break;
 		case PFILE_FIRE:
-			if (leveltype == DTYPE_TOWN) {
-				continue;
-			}
 			szCel = "FM";
 			pData = p->_pFData;
 			pAnim = (BYTE *)p->_pFAnim;
 			break;
 		case PFILE_MAGIC:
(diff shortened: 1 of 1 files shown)


AJenbo:
The animation has its weapon drawn as oppose to other town animations where the weapon is not in hand.

galaxyhaxz:
Yep, it's because they got lazy and didn't want to make variants of the player casting spells while the weapon is off hand. When casting a spell in town the player will temporarily have their helmet on and weapon up. Unfortunately they also caked the spell overlay on top of the sprite, or else you could draw it on top of the player and it wouldn't look so buggy.

The other reason was probably to coincide with town killing :P

galaxyhaxz:
My suggestion is we do what Diablo 2 did: eliminate town-specific animations entirely and just use the same ones as the dungeon. This fixes the odd quirks. I know it's not the ideal solution but we don't have the 3D models and unless someone wants to spend 100000000000 hours recreating all possible character frames we are screwed :S

qndel:
ye I was thinking about it too but didn't feel like dumping more into this PR : p

AJenbo:
Let's revisit this when we have 3D models

galaxyhaxz:
image

dcworldwide:
Soon enough I'm sure ML and generative ai could be used to train an algo to generate character models. Can dream.

AJenbo:
Processing...
image

Generating...
image

galaxyhaxz:
Quick, you have 8 months to beat Blizzard to the Diablo remaster punch!

PR #1107, opened 2021-03-05
Migrate all engine.h functions to accept a buffer

glebm:
Removes non-buffer versions from engine.h and migrates all remaining engine.h functions to accept a buffer.


glebm:
We can now remove the `To` suffix from these functions, I'll send a separate PR once this is merged for ease of review

AJenbo (dismissed):

AJenbo:
is gpBufEnd still used?

glebm:
Unfortunately, yes, there are still a few places where `gpBuffer` and `gpBufEnd` are used directly.

PR #1123, opened 2021-03-07
:sparkles: Preserve active spell and hotkeys across games

AJenbo:


galaxyhaxz:
Changing them to F1-F4 would also improve usability :)

AJenbo:
I'm hoping we can have key remapping instead 😉

galaxyhaxz:
Yes, eventually that would be lovely!

Source/loadsave.cpp
```diff
@@ -824,6 +916,36 @@ void ConvertLevels()
...
+		plr[myplr]._pSplTHotKey[i] = file.nextByte();
+	}
+	plr[myplr]._pRSpell = file.nextLE32();
+	plr[myplr]._pRSplType = file.nextByte();
+}
+
+void SaveHotkeys()
+{
+	SaveHelper file("hotkeys", sizeof(plr[myplr]._pSplHotKey) + sizeof(plr[myplr]._pSplTHotKey) + 5);
+
```
qndel, line 938: why that ugly 5 instead of sizeofs? :P

AJenbo: It's hardcoded to save as a 32bit and an 8bit value so sizeof() could actually give us the wrong value. Actually, that made me realize that the same should be done for the two others :o

AJenbo: Better? https://github.com/diasurgical/devilutionX/pull/1124

galaxyhaxz: Hmm, just curious. Why did you save them to a config file instead of just saving the data at the end of the save file?

You could have done:
`ISave(plr[myplr]._pSplTHotKey[i]`

AJenbo: There are explicit checks on the size of the `hero` file, so changing the size would cause it to show as an invalid save in vanila.

PR #1125, opened 2021-03-07
Set up inventory tests; add scroll usage tests

lkk7:


AJenbo:
Awsome! Thanks you :D

PR #1126, opened 2021-03-08
Loadsave refactoring

AJenbo:
This should properly port everything to various CPU architectures, as well as make it more resilient to changes in types of structure fields.


Source/loadsave.cpp
```diff
@@ -7,13 +7,41 @@
...
 DEVILUTION_BEGIN_NAMESPACE
 
-BYTE *tbuff;
-
 bool gbIsHellfireSaveGame;
 int giNumberOfLevels;
 int giNumberQuests;
 int giNumberOfSmithPremiumItems;
 
+inline Uint16 SwapLE(Uint16 in)
```
glebm, line 15: Anonymous `namespace` or `static` for these functions and load/save helper.

`inline` does nothing in a `cpp` file: it's not a hint to inline a function in modern C++ compilers -- it simply tells the compiler that all duplicate function definitions are the same and it's OK to deduplicate.

glebm (approved):

AJenbo:
@glebm thank you for reviewing this beast :D

I was thinking that SwapLE and SwapBE should be moved to some header file, but haven't decided where that would be?
Would doing so change the meaning of `inline` so something useful?

Source/mpqapi.cpp
```diff
@@ -312,7 +312,7 @@ void ByteSwapHdr(_FILEHEADER *hdr)
 void InitDefaultMpqHeader(Archive *archive, _FILEHEADER *hdr)
 {
 	std::memset(hdr, 0, sizeof(*hdr));
-	hdr->signature = LOAD_LE32("MPQ\x1A");
+	hdr->signature = '\x1AQPM';
```
mewmew, line 315: Above (at line 265) `LOAD_LE32("MPQ\x1A");` was replaced with `SDL_SwapLE32('\x1AQPM');`.

On this line, the same code was replaced with `'\x1AQPM'`, without a call to `SDL_SwapLE32`. Was this difference intentional? Same question for line 325.

Source/towners.h
```diff
@@ -23,11 +23,6 @@ typedef struct TownerStruct {
 	int _ttype;
 	int _tx;    // Tile X-position of NPC
 	int _ty;    // Tile Y-position of NPC
-	int _txoff; // Sprite X-offset (unused)
```
mewmew, line 26: Keep these fields for towner walking animations?

mewmew (commented):
Just a few comments.

Source/towners.h
galaxyhaxz: Agreed, or just comment them out. Those fields are needed to make the towners walk. I guess in modern times everyone is too obese to move around :s

AJenbo: was kinda hoping we wold have some saner movement logic first, but ok  i will keep i around until that time comes :)

Source/mpqapi.cpp
AJenbo: that part was a temporary revert to get things back in to working order on BE, but this is exactly the inconsistency that i'm trying to find the proper solution for.

Issue #1136, opened 2021-03-09
QoL changes from Hellfire v1.02.A5


Deathrizes:
Many great changes and additions for hellfire that makes the game much more enjoyable!

-[   ] Item Outlines: In the dungeon, the color drawn to outline items on the floor now match the type of item (e.g. Gold outline = unique item).
-[   ] Adria and Pepin automatically fill your mana and life, no longer needing an extra action.
-[   ] Spell Scrolls: Unlearnable spells (Identify, Infravision, Resurrect) have blue ribbons instead of red ones.
-[   ] Durability Warnings: Now appear 2 points earlier (yellow and red at 7 and 4, instead of 5 and 2).
-[   ] Gold Piles: Always up to 10k (Personally 50k which I seen many suggest seems better, less clutter).
-[   ] Town Respawn: You now respawn within sight of Town Portals (rejoin other players faster).
-[   ] More Player Information: To improve coop play, mousing over other players now additionally shows their Armor Class and Resistances.
-[   ] Maxed Resistances: Now display their value (75%) instead of MAX for better display consistency with the above feature.
-[   ] Griswold: Stops favoring the sale of item types you don't have (now I can stop calling him "Grisbald the Junksmith").
-[   ] Wirt: Stops favoring the sale of item types you don't have (i.e. if you carry 3 bows, he'd virtually never sell a bow).
-[   ] Blocked Doors: No longer highlighted for interaction by the mouse; it can no longer obscure items behind it.
-[   ] New Game/Quit: To prevent accidental exits, these options are now disabled while you are dead in a multiplayer game.
-[   ] Chain Lightning: No longer scans for targets further than the corners of your screen (reduces gapping).
-[   ] Defiler Speeches: You will no longer hear the Defiler speak before you actually encounter it (no more awful disembodied taunting).

AJenbo:
I corrected the spacing so that the list now appears correct with usable a checkmark :)

Pepin automatically healing you is actually a normal feature of Hellfire 1.00, so this is already a part of DevilutionX.
Adria never refilled mana for free so that is a bit different.

malvarenga123:
>Gold Piles: Always up to 10k (Personally 50k which I seen many suggest seems better, less clutter).
I would like a Diablo 2 approach where gold is just a number in the inventory

NeonLightning:
i gotta say all these changes would be nice(although i don't play multiplayer)

julealgon:
[...]

PR #1143, opened 2021-03-09
QoL: Fill mana when speaking with Adria

yuripourre:


qndel:
things like these are left for mods to do, we don't change the game like this

yuripourre:
> things like these are left for mods to do, we don't change the game like this

Sorry, my bad. I saw #1136 but didn't read all the comments.

yuripourre:
@AJenbo Added the config, I will be more careful next time.

Source/diablo.cpp
```diff
@@ -431,6 +431,7 @@ static void SaveOptions()
 	setIniInt("Game", "Experience Bar", sgOptions.bExperienceBar);
 	setIniInt("Game", "Enemy Health Bar", sgOptions.bEnemyHealthBar);
 	setIniInt("Game", "Auto Gold Pickup", sgOptions.bAutoGoldPickup);
+	setIniInt("Game", "Fill Mana Store", sgOptions.bFillManaStore);
```
AJenbo, line 434:
```suggestion
	setIniInt("Game", "Adria Refills Mana, sgOptions.bAdriaRefillsMana);
```

Source/diablo.cpp
```diff
@@ -476,6 +477,7 @@ static void LoadOptions()
 	sgOptions.bExperienceBar = getIniBool("Game", "Experience Bar", false);
 	sgOptions.bEnemyHealthBar = getIniBool("Game", "Enemy Health Bar", false);
 	sgOptions.bAutoGoldPickup = getIniBool("Game", "Auto Gold Pickup", false);
+	sgOptions.bFillManaStore = getIniBool("Game", "Fill Mana Store", false);
```
AJenbo, line 480:
```suggestion
	sgOptions.bAdriaRefillsMana = getIniBool("Game", "Adria Refills Mana", false);
```

Source/diablo.h
```diff
@@ -47,6 +47,7 @@ typedef struct Options {
 	bool bExperienceBar;  // Show the current level progress
 	bool bEnemyHealthBar; // Show enemy health at the top of the screen
 	bool bAutoGoldPickup; // Automatically pick up goald when walking on to it
+	bool bFillManaStore;  // Recover mana when talking to Adria
```
AJenbo, line 50:
```suggestion
	bool bAdriaRefillsMana;  // Recover mana when talking to Adria

```

Source/stores.cpp
```diff
@@ -811,8 +811,21 @@ void S_StartSRepair()
 	OffsetSTextY(22, 6);
 }
 
+static void FillManaPlayer()
+{
+	if (!sgOptions.bFillManaStore)
```
AJenbo, line 816:
```suggestion
	if (!sgOptions.bAdriaRefillsMana)
```

galaxyhaxz:
This is an interesting one. In Diablo 2 for example, there is one "healer" in each act and they heal both mana+life. By enabling healing of mana, it can potentially create a small balance issue:

Let's say you die, all your gear drops and you are out of gold to buy potions/new gear. If you have spells you could take to Adria, get mana and be able to make a few kills. Whereas without it, you'd literally be screwed and probably lose your gear :p

What I plan on doing instead is having passive mana regen like in Diablo 2. Since magic (a.k.a "energy") is derived from within, it would make sense that after resting it slowly comes back. Plus it creates cool builds with gear to keep their mana up.

PR #1148, opened 2021-03-10
:recycle: Make options struct hierarchical

julealgon:
This PR splits up the `Options` model into several smaller models, one for each configurable section, mimicking the structure of the ini file sections.

This makes each model a bit more manageable in code and improves readability.


AJenbo:
nice work, at some point options might start to makse sens :)

PR #1149, opened 2021-03-10
Add item sounds to auto-equip

julealgon:
Adds item equip sound to when the item is automatically equipped on pickup. Sound is disabled by default and can be turned on via ini toggle.


julealgon:
Depends on #1148 being merged first.

AJenbo (approved):

PR #1159, opened 2021-03-11
Fixes the jitter of the player on the automap when moving diagonally

FluffyQuack:


AJenbo:
neat. do other players also jitter?

FluffyQuack:
> neat. do other players also jitter?

I didn't even know DevilutionX showed other players on the automap. I just tested it now and it's kind of horribly broken, probably because of this PR. I'll fix it tomorrow.

FluffyQuack:
I did a bit more fiddling with this, and I think the core problem might be that the position of the automap itself jitters whenever you move diagonally. By default, the player arrow and the automap jitters unevenly which makes the player arrow stand out (that's what my fix addresses), but maybe this problem should be approached differently than what I'm doing.

I'll leave it up to you if you wanna incorporate the PR as it fixes a part of the problem, and not the entire problem. (It's also a very minor problem to begin with.)

AJenbo:
This ended up being fixed in a different way :shrug:

PR #1163, opened 2021-03-12
:sparkles: Autoequip items on purchase

julealgon:
This PR introduces auto-equip functionality for purchased items. 

If auto-equip is enabled for the corresponding item type, the item is checked against the player's current body slots: if the item can be equipped, the purchase is made and the item is equipped automatically.

Note that this means one can now puchase items even with a full inventory, provided he has the corresponding empty slot in the body for the item to be equipped.


AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/95ae8fd5c6fde8c9f40b4038caf68f258952b96c
Yay for Git Kraken being able to autoresolve this

PR #1165, opened 2021-03-12
🚩 Add ini toggle for FPS Limit. Enabled by default.

Trihedraf:
Add toggle for the FPS Limiter.

### DIFF
--- a/SourceX/dx.cpp
+++ b/SourceX/dx.cpp
@@ -251,7 +251,8 @@ void LimitFrameRate()
 	uint32_t v = 0;
 	if (frameDeadline > tc) {
 		v = tc % refreshDelay;
-		SDL_Delay(v / 1000 + 1); // ceil
+		if (sgOptions.Graphics.bFPSLimit)
+			SDL_Delay(v / 1000 + 1); // ceil
 	}
 	frameDeadline = tc + v + refreshDelay;
 }
--- a/SourceX/dx.cpp
+++ b/SourceX/dx.cpp
@@ -246,13 +246,14 @@ void Blit(SDL_Surface *src, SDL_Rect *src_rect, SDL_Rect *dst_rect)
  */
 void LimitFrameRate()
 {
+	if (!sgOptions.Graphics.bFPSLimit)
+		return;
 	static uint32_t frameDeadline;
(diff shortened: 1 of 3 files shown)


SourceX/dx.cpp
```diff
@@ -251,7 +251,8 @@ void LimitFrameRate()
 	uint32_t v = 0;
 	if (frameDeadline > tc) {
 		v = tc % refreshDelay;
-		SDL_Delay(v / 1000 + 1); // ceil
+		if (sgOptions.Graphics.bFPSLimit)
+			SDL_Delay(v / 1000 + 1); // ceil
```
AJenbo, line 255: Do an early return in the function instead of additionally nested ifs. We don't need to do any of the other calculations anyway.

Trihedraf: Like this?

if (frameDeadline > tc && sgOptions.Graphics.bFPSLimit) {
		v = tc % refreshDelay;
		SDL_Delay(v / 1000 + 1); // cei`

AJenbo: No that is again nesting conditions.

early return:
```
if (!sgOptions.Graphics.bFPSLimit)
  return;
```

AJenbo:
Oh one other thing, don't use actual Unicode emojis as some git clients choke on it.

https://gitmoji.dev/

use `:triangular_flag_on_post:` instead of 🚩

Trihedraf:
Ok. Is there a way to update previous commits to fix that?

AJenbo:
Yeah, you can amend. But I didn't notice it before merging so I had to force push to master which is a bit of a no-no :/

alican3437:
Are you artificial intelligence or human?

iOS için Outlook<https://aka.ms/o0ukef> uygulamasını edinin
________________________________
Gönderen: Anders Jenbo ***@***.***>
Gönderildi: Friday, March 12, 2021 11:19:35 AM
Kime: diasurgical/devilutionX ***@***.***>
Bilgi: Subscribed ***@***.***>
Konu: Re: [diasurgical/devilutionX] 🚩 Add ini toggle for FPS Limit. Enabled by default. (#1165)


Yeah, you can amend. But I didn't notice it before merging so I had to force push to master which is a bit of a no-no :/

PR #1166, opened 2021-03-12
fix infinite attack loop when force attacking a non-hostile character

alanbirtles:
Not sure if this is the right fix but fixes the problem for me. Reproduce by standing next to a non hostile character (I found the bug with Lachdanan), putting the cursor the opposite side of them then force attack. An infinite loop of attacks starts and there doesn't seem to be any way to break out of it.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2725,7 +2725,8 @@ BOOL PM_DoAttack(int pnum)
 				m = -(dMonster[dx][dy] + 1);
 			}
 			if (CanTalkToMonst(m)) {
-				plr[pnum]._pVar1 = 0;
+				StartStand(pnum, plr[pnum]._pdir);
+				ClearPlrPVars(pnum);
 				return FALSE;
 			}
 		}


alanbirtles:
could well be, its not one I remember encountering but it was 20 years ago!

qndel:
can't you test in vanilla? :P

alanbirtles:
don't think it runs on my machine

julealgon:
By "force attack" do you mean shift click in the opposite direction? Does this happen with every class/weapon? Does it happen consistently regardless of save (did you check with different characters)?

AJenbo:
looks like a reasonable fix, i actually found something similar the other day. If attack a barral next to Lachdanan and then click him before the animation is over you will perform attacks on him (he wong get hit)

galaxyhaxz:
Tested in vanilla, I can't reproduce this bug.

alanbirtles:
This was with bard dual welding swords, I didn't try other characters. Yes I do mean shift clicking, note that shift clicking on the character directly doesn't trigger an attack, you have to click behind the character so that you are attacking in their direction but not attempting to attack them directly. Looking at the code I imagine this will happen with any melee weapon.

galaxyhaxz:
Ok, so most likely it's specific to the barbarian and changes made with hellfire.

alanbirtles:
yes, it does seem to only affect the bard, I couldn't reproduce with a warrior. With a bard it happens with single or dual wielded melee weapons

julealgon:
Fix looks a bit arbitrary/hacky to me. I wonder if there is a cleaner way to go about this. If the problem only occurs with bard, it doesn't make sense to me to do this this way. Why does it happen only with the bard in the first place?

I don't feel comfortable in making fixes while we don't clearly understand the problem at hand.

julealgon:
@alanbirtles can you please provide a save file that can easily repro this issue?

kphoenix137:
3 year old PR for an issue nobody except the author can reproduce. Close?
@AJenbo

kphoenix137:
Tested in 1.6-dev in Hellfire Single Player with a Level 50 Bard using starting weapons, using mouse and keyboard. Stood on all sides of Lachdanan and shift clicked on the other side of him for each direction. Could not reproduce the bug stated in this PR.

AJenbo:
It's possible that if the issue did exist then it's been solved now by other logic cleanups that prevents this from happening in the first place.

PR #1170, opened 2021-03-13
Completely replace gpBuffer with CelOutputBuffer

glebm:
`CelOutputBuffer` now contains an `SDL_Surface` and an `SDL_Rect`.

We now have access to SDL surface manipulation functions.

`gpBuffer` and `gpBufEnd` are completely gone 🧹


AJenbo (approved):
Really exciting, should make it easier to  integrate some of the custom assets. Also i see this gets rid of some now redundant assertion.

PR #1172, opened 2021-03-13
Zoom: Fix off-by-1 and handle odd dimensions

glebm:
The original code went out of viewport bounds on every line copy (doesn't crash only because of the buffer padding).

It also did not handle odd viewport dimensions (the original game didn't have to).

This is a new version of the function that handles both and in a more obvious way.

### DIFF
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -946,38 +946,57 @@ static void scrollrt_draw(CelOutputBuffer out, int x, int y, int sx, int sy, int
 }
 
 /**
- * @brief Scale up the rendered part of the back buffer to take up the full view
+ * @brief Scale up the top left part of the buffer 2x.
  */
 static void Zoom(CelOutputBuffer out)
 {
-	int wdt = gnScreenWidth / 2;
-
-	int src_x = gnScreenWidth / 2 - 1;
-	int dst_x = gnScreenWidth - 1;
-
+	int viewport_width = out.w();
+	int viewport_offset_x = 0;
 	if (PANELS_COVER) {
 		if (chrflag || questlog) {
-			wdt >>= 1;
-			src_x -= wdt;
+			viewport_width -= SPANEL_WIDTH;
+			viewport_offset_x = SPANEL_WIDTH;
 		} else if (invflag || sbookflag) {
-			wdt >>= 1;
-			src_x -= wdt;
-			dst_x -= SPANEL_WIDTH;
+			viewport_width -= SPANEL_WIDTH;
 		}
 	}
 
-	BYTE *src = out.at(src_x, gnViewportHeight / 2 - 1);
-	BYTE *dst = out.at(dst_x, gnViewportHeight - 1);
+	// We round to even for the source width and height.
+	// If the width / height was odd, we copy just one extra pixel / row later on.
+	const int src_width = (viewport_width + 1) / 2;
+	const int doubleable_width = viewport_width / 2;
+	const int src_height = (out.h() + 1) / 2;
+	const int doubleable_height = out.h() / 2;
 
-	for (int hgt = 0; hgt < gnViewportHeight / 2; hgt++) {
-		for (int i = 0; i < wdt; i++) {
+	BYTE *src = out.at(src_width - 1, src_height - 1);
+	BYTE *dst = out.at(viewport_offset_x + viewport_width - 1, out.h() - 1);
+	const bool odd_viewport_width = (viewport_width % 2) == 1;
+
+	for (int hgt = 0; hgt < doubleable_height; hgt++) {
+		// Double the pixels in the line.
+		for (int i = 0; i < doubleable_width; i++) {
 			*dst-- = *src;
 			*dst-- = *src;
-			src--;
+			--src;
 		}
-		memcpy(dst - out.pitch(), dst, wdt * 2 + 1);
-		src -= out.pitch() - wdt;
-		dst -= 2 * (out.pitch() - wdt);
+
+		// Copy a single extra pixel if the output width is odd.
+		if (odd_viewport_width) {
+			*dst-- = *src;
+			--src;
+		}
+
+		// Skip the rest of the source line.
+		src -= (out.pitch() - src_width);
+
+		// Double the line.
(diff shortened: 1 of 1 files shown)


AJenbo:
Well, something is off by one. Without it the right side of the screen looks wrong:
image

Source/scrollrt.cpp
```diff
@@ -943,7 +943,7 @@ static void Zoom(CelOutputBuffer out)
 			*dst-- = *src;
 			src--;
 		}
-		memcpy(dst - out.pitch(), dst, wdt * 2 + 1);
+		memcpy(dst - out.pitch(), dst, wdt * 2);
```
AJenbo, line 946: There that fixes it:
```suggestion
		memcpy(dst - out.pitch() + 1, dst + 1, wdt * 2);
```

glebm: This makes the left side look wrong. I'll look into this a bit more (I think 1 should be added but not always)

glebm: I got here because I tried removing BORDER/SCREEN_XY paddings and got a `memcpy` overlap ASAN warning when I did that

AJenbo: You probably can't get rid of BORDER before all blotting happens via SDL, unless you modefy all biltter functions to be aware of the buffer edges.

Else sprites leaving on the left will appear on the right and vice versa (shifted 1 px in height).

AJenbo: > This makes the left side look wrong. I'll look into this a bit more (I think 1 should be added but not always)

Are you sure about that, it looked correct when I tested it.

AJenbo: Maybe there is some integer rounding going on that needs to be taking in to consideration and various resolutions (I was testing at 640x480)

glebm: I've had a detailed look at this function and it had a few issues.
I rewrote it with comments and fixed this issue and an issue with odd dimensions.

julealgon (changes requested):
Combining refactoring changes with bug fixes makes it much harder to review.

glebm:
@julealgon These are not refactoring changes, I rewrote the function mostly from scratch. It just happens to have a similar structure to the original function.

glebm:
Ah, the original PR description is very out of date

Source/scrollrt.cpp
```diff
@@ -946,38 +946,57 @@ static void scrollrt_draw(CelOutputBuffer out, int x, int y, int sx, int sy, int
...
 
-	for (int hgt = 0; hgt < gnViewportHeight / 2; hgt++) {
-		for (int i = 0; i < wdt; i++) {
+	for (int hgt = 0; hgt < doubleable_height; hgt++) {
+		// Double the pixels in the line.
+		for (int i = 0; i < doubleable_width; i++) {
 			*dst-- = *src;
 			*dst-- = *src;
-			src--;
+			--src;
```
julealgon, line 980: I know this is a tiny irrelevant detail, but I noticed you changed `src--` to `--src`. Why did you do it?

AJenbo (approved):

Source/scrollrt.cpp
```diff
@@ -946,38 +946,57 @@ static void scrollrt_draw(CelOutputBuffer out, int x, int y, int sx, int sy, int
...
+			viewport_width -= SPANEL_WIDTH;
 		}
 	}
 
-	BYTE *src = out.at(src_x, gnViewportHeight / 2 - 1);
-	BYTE *dst = out.at(dst_x, gnViewportHeight - 1);
+	// We round to even for the source width and height.
+	// If the width / height was odd, we copy just one extra pixel / row later on.
+	const int src_width = (viewport_width + 1) / 2;
+	const int doubleable_width = viewport_width / 2;
```
julealgon, line 967: Is `doubleable` a word? What does this mean exactly?

Source/scrollrt.cpp
```diff
@@ -946,38 +946,57 @@ static void scrollrt_draw(CelOutputBuffer out, int x, int y, int sx, int sy, int
...
-		dst -= 2 * (out.pitch() - wdt);
+
+		// Skip the rest of the source line.
+		src -= (out.pitch() - src_width);
+
+		// Double the line.
+		memcpy(dst - out.pitch() + 1, dst + 1, viewport_width);
+
+		// Skip the rest of the destination line.
+		dst -= 2 * out.pitch() - viewport_width;
```
julealgon, line 996: The number of comments and "+1", "-1" etc in this function makes it really scary and indicates to me that there must be a cleaner/more intuitive way to handle it.

Source/scrollrt.cpp
glebm: `src--` is a more complex operation semantically (and is usually implemented in terms of `--src` for more complex types). When the operand is a basic type it's a purely stylistic preference but it's nice to use operators consistently both for basic type (like pointers) and for complex types (like iterators).

Source/scrollrt.cpp
glebm: I couldn't come up with a better term, but this is the width the algorithm doubles.
Then there may be a single-pixel column/row that is copied as is, in case the viewport dimension is odd.

Source/scrollrt.cpp
glebm: I hope so, you're welcome to give it a go :)

AJenbo: Having poked it earlier, I take all the comments as a sign that this is deceptively non-simple :)

Source/scrollrt.cpp
julealgon: Very interesting... this must be why the iterator example I saw had ++X instead of X++. I actually changed it to X++ after copying it because I thought there was no need to ever use the prefixed version.

Source/scrollrt.cpp
julealgon: I don't mean to imply it is "simple", just that we may be lacking more abstraction to solve it in a cleaner, more semantic way.

I don't have a suggestion was well.. it was just a gut feeling of "something is wrong here".

PR #1181, opened 2021-03-14
Cleanup of playerstruct and automap

AJenbo:


Source/control.cpp
```diff
@@ -368,7 +368,8 @@ void DrawSpellList(CelOutputBuffer out)
 			c = SPLICONLAST + 2;
 			break;
 		}
-		for (spl = 1, j = 1; j < MAX_SPELLS; spl <<= 1, j++) {
+		Sint32 j = SPL_FIREBOLT;
+		for (spl = 1; j < MAX_SPELLS; spl <<= 1, j++) {
```
julealgon, line 372: Where's `j` coming from now? Can we really remove its initialization here?

Source/loadsave.cpp
```diff
@@ -284,17 +284,17 @@ static void LoadPlayer(LoadHelper *file, int p)
 {
 	PlayerStruct *pPlayer = &plr[p];
 
-	pPlayer->_pmode = file->nextLE<Sint32>();
+	pPlayer->_pmode = (PLR_MODE)file->nextLE<Sint32>();
```
julealgon, line 287: Doesn't this work?
```c++
nextLE<PLR_MODE>
```

Source/player.h
```diff
@@ -8,185 +8,223 @@
...
-	int _pTSpell;
-	char _pTSplType;
-	int _pRSpell;
+	PLR_MODE _pmode;
+	Sint8 walkpath[MAX_PATH_LENGTH];
+	bool plractive;
+	action_id destAction;
+	Sint32 destParam1;
+	Sint32 destParam2;
+	direction destParam3;
```
julealgon, line 61: Sure its safe to type this one? is it really always a direction?

julealgon (approved):

Source/control.cpp
AJenbo: Look at the line above...

Source/loadsave.cpp
AJenbo: It's not safe as we need the data to be read from the file in fixed sizes no matter how enums may be implemented on various systems.

Source/player.h
AJenbo: yep.

Source/loadsave.cpp
AJenbo: > An enum is only guaranteed to be large enough to hold int values. The compiler is free to choose the actual type used based on the enumeration constants defined so it can choose a smaller type if it can represent the values you define.

Source/control.cpp
julealgon: Missed it for some reason. Thanks.

Source/loadsave.cpp
julealgon: Ah got it. I got confused for a moment since C# enums allow you to explicitly define the underlying type. Seems to only be available in C++11 and newer though.

AJenbo: well, we are on C++11 so :shrug:

Source/missiles.cpp
```diff
@@ -809,13 +809,12 @@ BOOL PlayerMHit(int pnum, int m, int dist, int mind, int maxd, int mtype, BOOLEA
 				dam = 64;
 		}
 		if ((resper <= 0 || gbIsHellfire) && blk < blkper) {
+			direction dir = plr[pnum]._pdir;
 			if (m != -1) {
 				tac = GetDirection(plr[pnum]._px, plr[pnum]._py, monster[m]._mx, monster[m]._my);
```
AJenbo, line 814: oops https://github.com/diasurgical/devilutionX/commit/c899cabf983d64b6aecac001ec2f3f11de215ccd

Source/loadsave.cpp
julealgon: oh! then we should be able to use that XD

https://stackoverflow.com/a/1122109/1946412

AJenbo: yep, but I would rather not have the old savegame format dictate what types we use :)

PR #1183, opened 2021-03-14
Bump CMake to 3.13

AJenbo:


glebm (approved):
Note that we're doing this so that we can set options from toolchain files

https://cmake.org/cmake/help/latest/policy/CMP0077.html

PR #1184, opened 2021-03-14
Try uppercase DIABDAT.MPQ

glebm:
It's uppercase in GOG and on the retail CD.
Other MPQs are not known to be uppercase.


AJenbo:
We will have to fire the support team now.

PR #1185, opened 2021-03-14
Remove extern C from engine.h

### LINKED ISSUE #1178: CLang warning for CelOutputBuffer
When compiling on CLang the latest code can be a bit spammy :)

```
[ 61%] Building CXX object CMakeFiles/devilution.dir/Source/tmsg.cpp.o
In file included from /Users/ajenbo/Documents/devilutionX/Source/themes.cpp:6:
In file included from /Users/ajenbo/Documents/devilutionX/Source/all.h:42:
/Users/ajenbo/Documents/devilutionX/Source/dx.h:15:17: warning: 'GlobalBackBuffer' has C-linkage specified, but returns user-defined type 'dvl::CelOutputBuffer' which is
      incompatible with C [-Wreturn-type-c-linkage]
CelOutputBuffer GlobalBackBuffer();
                ^
```

glebm:
The headers don't actually work with C as they're now.

A better long-term solution should be able to use idiomatic C++ throughout
and expose things to other languages via a separate C API layer.

Fixes #1178

### DIFF
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -13,6 +13,9 @@
 #ifndef __ENGINE_H__
 #define __ENGINE_H__
 
+#include <algorithm>
+#include <cstddef>
+
 #include <SDL.h>
 
 #ifdef USE_SDL1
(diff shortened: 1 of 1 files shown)


AJenbo:
Thanks :)

PR #1188, opened 2021-03-14
Add default names for Bard and Barbarian

glebm:
Barbarian names taken from later Diablo lore.
Bard simply reuses Rogue names.


AJenbo:
Uh nice :)

PR #1189, opened 2021-03-14
Always prefill player name if controller is active

glebm:


AJenbo:
Nice, I had been considering doing this :)

galaxyhaxz:
It even does this on the PSX, though it uses `PLAYER 1/PLAYER 2` instead of the unique ones.

PR #1190, opened 2021-03-14
DiabloUI: Left thumbstick navigation

glebm:
Adds support for using the left thumb stick to go up and down the list

PR #1191, opened 2021-03-14
Thumb stick support in gmenu and questlog

glebm:
Allows gmenu and quest log navigation with the left thumb stick


glebm:
I'll send a new PR with both #1191 and #1190

PR #1196, opened 2021-03-15
Install devilutionx.mpq on Linux

glebm:
1. Installs devilutionx.mpq on Linux.
2. Falls back to finding it in the install location.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -24,10 +24,6 @@ end_of_line = lf
 [.gitignore]
 end_of_line = lf
 
-[CMakeLists.txt]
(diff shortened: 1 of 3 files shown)


AJenbo:
Thanks. I couldn't really get my self to bother with this but it will be appreciated by package managers I'm sure :)

AJenbo:
Possibly we should also look in some system folder for the official MPQs as it's common to make game data packages that install the data on system level on Linux.

https://wiki.debian.org/Games/GameDataPackager

glebm:
Yeah, could be good

PR #881, opened 2020-10-13
OpenDingux: Add PocketGo2 config

glebm:
PocketGo2 seems to have similar key bindings to GKD350h
but has a working IPU scaler like RG350.

Note: I don't have a PocketGo 2, this needs to be tested and controls
documented by volunteers.

[devilutionx-pocketgo2.opk.zip](https://github.com/diasurgical/devilutionX/files/5369828/devilutionx-pocketgo2.opk.zip)

PR #1199, opened 2021-03-15
Can build hellfire on OpenDingux by using `./build.sh rg350-hellfire`

john-parton:
This is a really small PR to the OpenDingux build script. I just added `*-hellfire` platforms that set the `-DHELLFIRE` cmake property.

So you can build hellfire on the rg350 like this: `./build.sh rg350-hellfire`

I was going to try to make it a flag (`./build.sh rg350 --helfire`), but I wasn't sure the best way to do that with bash. I usually make all my scripts in python.

It outputs an OPK with `-hellfire` in the name.

Let me know if there's a better way to do this.

Thanks.

----

I built and tested this on an RG350M, and it worked perfectly as far as I can tell.

### DIFF
--- a/Packaging/OpenDingux/build.sh
+++ b/Packaging/OpenDingux/build.sh
@@ -17,7 +17,14 @@ if ! check_target "$@"; then
   exit 64
 fi
 
-declare -r TARGET="$1"
+declare -r TARGET=$(echo "$1" | cut -d "-" -f 1)
+
+if [[ $(echo "$1" | cut -d "-" -f 2) == hellfire ]]; then
+	declare -r HELLFIRE=ON
+else
+	declare -r HELLFIRE=OFF
+fi
+
 declare -r BUILD_DIR="build-${TARGET}"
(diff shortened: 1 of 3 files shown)


john-parton:
Yeah, looks like it would conflict with a change like this one: https://github.com/diasurgical/devilutionX/pull/881

Probably better to have a `--hellfire` flag. I just have no idea how to do that.

glebm:
Hellfire is supported with simply `./build.sh rg350` as of https://github.com/diasurgical/devilutionX/pull/1137 (both Diablo and Hellfire are in the same opk)

john-parton:
Okay, that makes SO MUCH sense. I saw the -DHELLFIRE flag and assumed that was the only way to do it. Thanks for the clarification.

AJenbo:
> `-DHELLFIRE` isn't used for much anymore. 

infact it does nothing, its just there as a reminder that we need to find a solution for switch and 3ds where passing arguments isent an option

PR #1200, opened 2021-03-17
Stream music instead of loading it all at once

glebm:
RAM usage (my beefy desktop):

* Town: 176 MiB -> 121 MiB
* Cathedral: 191 MiB -> 138 MiB

Performance: 1170 FPS -> 1150 FPS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,6 +289,7 @@ set(devilutionx_SRCS
   SourceX/thread.cpp
   SourceX/sound.cpp
   SourceX/soundsample.cpp
+  SourceX/storm_sdl_rw.cpp
   SourceX/storm/storm.cpp
   SourceX/storm/storm_net.cpp
   SourceX/dvlnet/abstract_net.cpp
--- a/SourceS/miniwin/misc.h
+++ b/SourceS/miniwin/misc.h
@@ -62,6 +62,7 @@ bool PostMessage(UINT Msg, WPARAM wParam, LPARAM lParam);
 // MSCVRT emulation
 //
 
+#define DVL_FILE_BEGIN 0
 #define DVL_FILE_CURRENT 1
 #define DVL_FILE_END 2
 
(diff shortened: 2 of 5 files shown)


joewis:
Is this small FPS penalty consistently there or just a random fluctuation? Asking for less beefy hardware...

glebm:
@joewis It's not very consistent so hard to say. Either way 1150/1170 is <2%.

AJenbo (approved):

AJenbo:
wold this approach also work for the videos and dialogs?

malvarenga123:
Should this be enabled for PC? I can see the advantage for consoles/handhelds that have very limited RAM. A peak hard drive usage on PC would glitch the stream now, wouldn't it?

galaxyhaxz:
It is enabled for everything. The original game streamed the same way, I don't think having to read the MPQ more often should be an issue.

AJenbo:
This is also going to improve load time on PC. If your PC is unable to pull 75kb/s for the in-focus task it's probably hitting some other limits already. You can turn the music off if you think this is too much for your PC. Besides, modern systems will pull active content into RAM so for PC the only difference is probably going to be a slightly faster load time.

galaxyhaxz:
Even computers in 1996 could stream xD

malvarenga123:
> This is also going to improve load time on PC. If your PC is unable to pull 75kb/s for the in-focus task it's probably hitting some other limits already. You can turn the music off if you think this is too much for your PC. Besides, modern systems will pull active content into RAM so for PC the only difference is probably going to be a slightly faster load time.

I do have some trouble with other games (like Diablo 3) that require to be accessing the disk all the time, but I don't think Diablo 1 would have any problems. In fact, I always wondered why couldn't the whole diablo.mpq be loaded into memory, considering it's only ~600 MB.

glebm:
> wold this approach also work for the videos and dialogs?

Videos are probably streamed already.

It should be easy to migrate dialogs to use `SFileRw_FromStormHandle` but I haven't looked into it.

AJenbo:
> > wold this approach also work for the videos and dialogs?
> 
> Videos are probably streamed already.

I was thinking in terms of simplify the code and use sdl mixer instead of raw sdl

PR #1202, opened 2021-03-17
Rename sync_word_6AA708 to sgnMonsterPriority in Source/sync.cpp for clarity

Semperfis96:
To improve (a little bit) the code & begin to work on DevilutionX, I remove a strange hexa thing. Not a big deal, compile & test on multiplayer game, still work correctly.
 
PS: Sorry for bad english, it's not my native language.

### DIFF
--- a/Source/sync.cpp
+++ b/Source/sync.cpp
@@ -9,7 +9,7 @@ DEVILUTION_BEGIN_NAMESPACE
 
 namespace {
 
-Uint16 sync_word_6AA708[MAXMONSTERS];
+Uint16 sync_word[MAXMONSTERS];
 int sgnMonsters;
 Uint16 sgwLRU[MAXMONSTERS];
 int sgnSyncItem;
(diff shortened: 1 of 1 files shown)


julealgon:
Please try to be a bit more specific with PR/commit titles when possible.

Looks like you are renaming a very specific field here, so you might want to mention that in the message.

For example, "Rename X in Y module to Z" is a lot more descriptive than "Clean variable name".

Additionally, what did that suffix mean? I'd only approve this when i know what it means. Since i don't know I'll leave this to others to review.

AJenbo:
Thanks, but this doesn't actually help since it removes the context from the variable.

`Word` in this context is a reference to the data type located at `0x6AA708` in the original game, so one doesn't make sense without the other. 

In short, what you have done here is rename it from "Sync Value 5" to "Sync Value". Instead we need to figure out what "Value 5" (`word 0x6AA708`) represents.

From its usage, I think this is used for figuring out what monsters should be synced, by seeing ordering them by the distance to the player.

`sgnMonsterPriority` might be a reasonable name for it.

For x86 word is a 16bit unsigned integer:
https://en.wikipedia.org/wiki/Word_(computer_architecture).

galaxyhaxz:
> Sorry for the first title & the unclear PR, i'm quite new to github & i really want to be useful in Devilution

We all had to start somewhere :)

AJenbo:
Thanks for applying the name :)

Semperfis96:
Thank you for merging my small changes !
This is so good to participate to the code of a such mythical game !

julealgon:
> Sorry for the first title & the unclear PR, i'm quite new to github & i really want to be useful in Devilution

No need for apologies ;) Looking forward to your help, we don't have a lot of people maintaining the codebase.

PR #1204, opened 2021-03-17
Respect sfx_STREAM in PlaySFX

glebm:
Changes audio playback of sounds with the `sfx_STREAM` flag, such as towner lines, to stream.


AJenbo:
It's kinda funny it's not even that long ago that I basically gave up on this ever landing and threw away the interface for it: https://github.com/diasurgical/devilutionX/commit/88a715cdcc9be1d2a2124cb6d552e045439f99f2 But yours is even a nice bit cleaner than the original so it all worked out :)

PR #1206, opened 2021-03-17
Thumb stick / DPad repeating navigation in DiabloUI, gmenu, and quest log

glebm:
Implements thumb stick / dpad navigation and repeat-on-hold in DiabloUI, gmenu, and quest log.

1. Allows DiabloUI, gmenu, and quest log navigation with the left thumb stick.
    
2. De-duplicates repetition / throttling logic throughout.
    
3. Menus have separate repetition frequencies (inventory at 100ms, everything else at 200ms).
    
4. Movement mostly decoupled from menu navigation: the `dpad_hotkeys` setting used on the Vita no longer prevents dpad from navigating in menus.

### DIFF
--- a/SourceX/controls/axis_direction.cpp
+++ b/SourceX/controls/axis_direction.cpp
@@ -1,0 +1,55 @@
+#include "axis_direction.h"
+
+#include <SDL.h>
+
+namespace dvl {
+
+AxisDirection AxisDirectionRepeater::Get(AxisDirection axis_direction)
+{
+	const int now = SDL_GetTicks();
+	switch (axis_direction.x) {
+		case AxisDirectionX_LEFT:
+			last_right_ = 0;
+			if (now - last_left_ < min_interval_ms_) {
+				axis_direction.x = AxisDirectionX_NONE;
+			} else {
+				last_left_ = now;
+			}
+			break;
+		case AxisDirectionX_RIGHT:
+			last_left_ = 0;
+			if (now - last_right_ < min_interval_ms_) {
+				axis_direction.x = AxisDirectionX_NONE;
+			} else {
+				last_right_ = now;
+			}
+			break;
+		case AxisDirectionX_NONE:
+			last_left_ = last_right_ = 0;
+			break;
+	}
+	switch (axis_direction.y) {
+		case AxisDirectionY_UP:
+			last_down_ = 0;
+			if (now - last_up_ < min_interval_ms_) {
+				axis_direction.y = AxisDirectionY_NONE;
+			} else {
+				last_up_ = now;
+			}
+			break;
+		case AxisDirectionY_DOWN:
+			last_up_ = 0;
+			if (now - last_down_ < min_interval_ms_) {
+				axis_direction.y = AxisDirectionY_NONE;
+			} else {
+				last_down_ = now;
+			}
+			break;
+		case AxisDirectionY_NONE:
+			last_up_ = last_down_ = 0;
+			break;
+	}
+	return axis_direction;
+}
+
+} // namespace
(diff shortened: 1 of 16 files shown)


julealgon:
Any chance you could split your changes? This is a huge commit, its not trivial to review.

glebm:
It may seem that way due to the line count but I think it is not that difficult to review if you're familiar with the relevant parts of the codebase.

Splitting up a change like this up is a ton of work. I'll consider it if @AJenbo needs me to.

AJenbo:
isn't about 256 +/- which isn't to bad, and like you mentioned this isn't as hard to review when familiar with the given code. Just haven't had the time yet.

SourceX/controls/game_controls.cpp
```diff
@@ -14,10 +14,10 @@ namespace dvl {
 
 bool start_modifier_active = false;
 bool select_modifier_active = false;
-
-// gamepad dpad acts as hotkeys without holding "start"
+
+// gamepad dpad acts as hotkeys without holding "start"
```
AJenbo, line 18:
```suggestion
/** Gamepad dpad acts as hotkeys without holding "start" */

```

SourceX/controls/game_controls.cpp
```diff
@@ -14,10 +14,10 @@ namespace dvl {
 
 bool start_modifier_active = false;
 bool select_modifier_active = false;
-
-// gamepad dpad acts as hotkeys without holding "start"
+
+// gamepad dpad acts as hotkeys without holding "start"
 bool dpad_hotkeys = false;
-// shoulder gamepad buttons act as potions by default
+// shoulder gamepad buttons act as potions by default
```
AJenbo, line 20:
```suggestion
/** Shoulder gamepad buttons act as potions by default */
```

SourceX/controls/game_controls.h
```diff
@@ -22,7 +25,7 @@ enum GameActionType {
 };
 
 struct GameActionSendKey {
-	DWORD vk_code;
+	std::uint32_t vk_code;
```
AJenbo, line 28: Would prefer Uint32

SourceX/controls/menu_controls.h
```diff
@@ -21,4 +21,7 @@ enum MenuAction {
 
 MenuAction GetMenuAction(const SDL_Event &event);
 
+// Menu action from holding the left stick or DPad.
```
AJenbo, line 24: Use Doxygen style for documentation
```suggestion
/** Menu action from holding the left stick or DPad. */
```

SourceX/controls/game_controls.h
glebm: I think we should use standard C++ types instead of the SDL stuff that was added for compatibility with ancient C versions (not C++)?

SourceX/controls/axis_direction.cpp
```diff
@@ -0,0 +1,55 @@
+#include "axis_direction.h"
+
+#include <SDL.h>
+
+namespace dvl {
+
+AxisDirection AxisDirectionRepeater::Get(AxisDirection axis_direction)
```
AJenbo, line 7: Show we maybe set the default to 200 since it is only in one place that isn't the case?

or at least define them so they aren't just magical values.

glebm: Good idea, done

SourceX/controls/game_controls.h
AJenbo: Uint32 maps to uint32_t on systems where it is available, so would there be much of a difference in the cases you refer to?

AJenbo (approved):

SourceX/controls/game_controls.h
glebm: We don't support any compilers where it's not available and we never will
`<cstdint>` is ancient stuff

glebm: SDL just happens to be even more ancient

glebm: Anyway, if we change something like that it should be done in a separate PR throughout.

Changed to Uint32

PR #1209, opened 2021-03-17
Update 3rdParty/Radon

glebm:
From https://github.com/szewczukk/Radon/commit/9203ff8b029584d3026660714862976a7c13a97e


glebm:
This solves one of the issues @awesomekling encountered when porting DevilutionX to his OS https://youtu.be/ZOzZ8R4gphE?t=1782

galaxyhaxz:
Clearing out the C++ with C. More portability and human friendly. I like it :+1:

AJenbo:
Haha, I just applied once of his changes in my Clang clean up as well :D

AJenbo:
Wait I upstreamed those changes but forgot to update DevilutionX afterwards x(

3rdParty/Radon/Radon/source/File.cpp
```diff
@@ -47,15 +47,15 @@ namespace radon
 
 	Section *File::getSection(const std::string & name)
 	{
-		for (auto & section : sections)
+		for (int i = 0; i < (int)sections.size(); i++)
```
julealgon, line 50: We cannot use iterators? Can you elaborate on this?

3rdParty/Radon/Radon/source/Section.cpp
```diff
@@ -17,19 +17,21 @@ namespace radon
...
+		return NULL;
 	}
 
+
 	void Section::addKey(Key key)
 	{
 		keys.push_back(key);
 	}
-}
+}
```
julealgon, line 37: Are these really intentional? Do we have it setup like this in the project?

3rdParty/Radon/Radon/source/File.cpp
glebm: IIRC this was done for the attempt at an XBox port. XBox SDK at that point only supported C++98

glebm: (I'm not the author of these changes)

3rdParty/Radon/Radon/source/Section.cpp
glebm: Radon is not part of DevilutionX code. It's a vendored third-party library.

julealgon: Are all files under the `3rd-party` folder excluded from editorconfig and compiler warnings on our side?

glebm: They should be excluded from `.clang-format` and `.editor-config` (they're not currently -- this is an oversight).

However, we still want to have the warnings so that we can fix them upstream.

3rdParty/Radon/Radon/source/File.cpp
AJenbo: C++03

PR #1210, opened 2021-03-17
Fix majority of the Clang warnings

AJenbo:
Fixes https://github.com/diasurgical/devilutionX/issues/1201 (except for 4 warnings about switches not covering all enum values). Sorry that this is all in one chunk, but if you ignore whitespace changes (can be done in github) then it's only 150 lines changed in total.

### DIFF
--- a/3rdParty/Radon/Radon/include/Key.hpp
+++ b/3rdParty/Radon/Radon/include/Key.hpp
@@ -17,13 +17,13 @@ namespace radon
 
 		Key(const std::string & name, const std::string & value);
 
-		Key(const std::string & name, const float & value);
+		Key(const std::string & name, float value);
 
 		std::string getStringValue();
 
 		float getFloatValue();
 
-		void setValue(float & value);
+		void setValue(float value);
 
 		void setValue(std::string & value);
 
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -351,6 +351,7 @@ void DrawAutomapPlr(CelOutputBuffer out, int pnum)
 		DrawLineTo(out, x + AmLine16, y + AmLine8, x + AmLine8, y + AmLine8, playerColor);
 		break;
 	case DIR_S:
+	case DIR_OMNI:
 		DrawLineTo(out, x, y, x, y + AmLine16, playerColor);
 		DrawLineTo(out, x, y + AmLine16, x + AmLine4, y + AmLine8, playerColor);
 		DrawLineTo(out, x, y + AmLine16, x - AmLine4, y + AmLine8, playerColor);
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -367,6 +367,8 @@ void DrawSpellList(CelOutputBuffer out)
 			mask = plr[myplr]._pISpells;
 			c = SPLICONLAST + 2;
 			break;
+		case RSPLTYPE_INVALID:
+			break;
 		}
 		Sint32 j = SPL_FIREBOLT;
 		for (spl = 1; j < MAX_SPELLS; spl <<= 1, j++) {
@@ -440,6 +442,8 @@ void DrawSpellList(CelOutputBuffer out)
 						sprintf(tempstr, "%i Charges", plr[myplr].InvBody[INVLOC_HAND_LEFT]._iCharges);
 					AddPanelString(tempstr, TRUE);
 					break;
+				case RSPLTYPE_INVALID:
+					break;
 				}
 				for (t = 0; t < 4; t++) {
 					if (plr[myplr]._pSplHotKey[t] == pSpell && plr[myplr]._pSplTHotKey[t] == pSplType) {
@@ -510,6 +514,8 @@ void ToggleSpell(int slot)
 	case RSPLTYPE_CHARGES:
 		spells = plr[myplr]._pISpells;
 		break;
+	case RSPLTYPE_INVALID:
+		return;
 	}
 
 	if (spells & GetSpellBitmask(plr[myplr]._pSplHotKey[slot])) {
@@ -1067,6 +1073,8 @@ void CheckPanelInfo()
 				else
 					sprintf(tempstr, "%i Charges", plr[myplr].InvBody[INVLOC_HAND_LEFT]._iCharges);
 				AddPanelString(tempstr, TRUE);
+				break;
+			case RSPLTYPE_INVALID:
 				break;
 			}
 		}
@@ -1678,26 +1686,25 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
 	if (pItem->_iDurability > 5)
 		return x;
 	if (c == 0) {
-		if (pItem->_iClass == ICLASS_WEAPON) {
-			switch (pItem->_itype) {
-			case ITYPE_SWORD:
-				c = 2;
-				break;
-			case ITYPE_AXE:
-				c = 6;
-				break;
-			case ITYPE_BOW:
-				c = 7;
-				break;
-			case ITYPE_MACE:
-				c = 5;
-				break;
-			case ITYPE_STAFF:
-				c = 8;
-				break;
-			}
-		} else {
+		switch (pItem->_itype) {
+		case ITYPE_SWORD:
+			c = 2;
+			break;
+		case ITYPE_AXE:
+			c = 6;
+			break;
+		case ITYPE_BOW:
+			c = 7;
+			break;
+		case ITYPE_MACE:
+			c = 5;
+			break;
+		case ITYPE_STAFF:
+			c = 8;
+			break;
+		default:
 			c = 1;
+			break;
 		}
 	}
 	if (pItem->_iDurability > 2)
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -338,7 +338,7 @@ static void DrawSpell(CelOutputBuffer out)
 void DrawSpellList(CelOutputBuffer out)
 {
 	int x, y, c, s, t, v, lx, ly, trans;
-	unsigned __int64 mask, spl;
+	Uint64 mask, spl;
 
 	pSpell = SPL_INVALID;
 	infostr[0] = '\0';
@@ -491,7 +491,7 @@ void SetSpeedSpell(int slot)
 
 void ToggleSpell(int slot)
 {
-	unsigned __int64 spells;
+	Uint64 spells;
 
 	if (plr[myplr]._pSplHotKey[slot] == -1) {
 		return;
@@ -512,7 +512,7 @@ void ToggleSpell(int slot)
 		break;
 	}
 
-	if (spells & SPELLBIT(plr[myplr]._pSplHotKey[slot])) {
+	if (spells & GetSpellBitmask(plr[myplr]._pSplHotKey[slot])) {
 		plr[myplr]._pRSpell = plr[myplr]._pSplHotKey[slot];
 		plr[myplr]._pRSplType = plr[myplr]._pSplTHotKey[slot];
 		force_redraw = 255;
@@ -855,7 +855,6 @@ void DrawCtrlBtns(CelOutputBuffer out)
  */
 void DoSpeedBook()
 {
-	unsigned __int64 spell, spells;
 	int xo, yo, X, Y, i, j;
 
 	spselflag = TRUE;
@@ -866,6 +865,7 @@ void DoSpeedBook()
 
 	if (plr[myplr]._pRSpell != SPL_INVALID) {
 		for (i = 0; i < 4; i++) {
+			Uint64 spells;
 			switch (i) {
 			case RSPLTYPE_SKILL:
 				spells = plr[myplr]._pAblSpells;
@@ -880,7 +880,7 @@ void DoSpeedBook()
 				spells = plr[myplr]._pISpells;
 				break;
 			}
-			spell = (__int64)1;
+			Uint64 spell = 1;
 			for (j = 1; j < MAX_SPELLS; j++) {
 				if (spell & spells) {
 					if (j == plr[myplr]._pRSpell && i == plr[myplr]._pRSplType) {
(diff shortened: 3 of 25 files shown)


Source/inv.cpp
```diff
@@ -1776,39 +1772,37 @@ void CheckInvCut(int pnum, int mx, int my, bool automaticMove)
...
 							break;
-
 						case PC_SORCERER:
 							PlaySFX(PS_MAGE15, false);
 							break;
-
 						case PC_MONK:
 							PlaySFX(PS_MONK15, false);
 							break;
+						case NUM_CLASSES:
```
galaxyhaxz, line 1785: This is a bit odd, given this could obviously never happen.

SourceX/qol.cpp
```diff
@@ -160,11 +160,11 @@ void DrawXPBar(CelOutputBuffer out)
 	int curXp = ExpLvlsTbl[charLevel];
 	int prevXp = ExpLvlsTbl[charLevel - 1];
 	int prevXpDelta = curXp - prevXp;
-	int prevXpDelta_1 = plr[myplr]._pExperience - prevXp;
+	Uint64 prevXpDelta_1 = plr[myplr]._pExperience - prevXp;
```
glebm, line 163: I'd put this assignment after the `if` guard on the line below

defs.h
```diff
@@ -21,7 +21,7 @@
 #define MAX_LVLMTYPES			24
 #define MAX_SPELLS				52
 #define MAX_SPELL_LEVEL			15
-#define SPELLBIT(s) ((__int64)1 << (s - 1))
+#define SPELLBIT(s) ((Uint64)1 << (s - 1))
```
glebm, line 24: `1ULL`?

glebm (approved):

Source/inv.cpp
julealgon: The idea of adding an extra enum member to signal the number of elements is a terrible hack. Obviously the static analysis cannot tell that this is one such instance.

It's sad that we have to resort to this... is there something else we can do with C++11 to iterate on the enum without needing this hack @AJenbo ?

SourceX/qol.cpp
julealgon: Can you elaborate why? Would that be related to fixing the warnings?

defs.h
julealgon: I think we have this in function form, so would suggest dropping the macro.

julealgon:
>  AJenbo removed the request for review from julealgon 3 hours ago

Why? oO

3rdParty/Radon/Radon/include/Key.hpp
```diff
@@ -23,7 +23,7 @@ namespace radon
 
 		float getFloatValue();
 
-		void setValue(float & value);
+		void setValue(const float & value);
```
julealgon, line 26: Nice.

Source/automap.cpp
```diff
@@ -351,6 +351,7 @@ void DrawAutomapPlr(CelOutputBuffer out, int pnum)
 		DrawLineTo(out, x + AmLine16, y + AmLine8, x + AmLine8, y + AmLine8, playerColor);
 		break;
 	case DIR_S:
+	case DIR_OMNI:
```
julealgon, line 354: This looks wrong to me. Shouldn't we just throw some sort of error in this case? Unless the value is the same as `DIR_S`?

Source/control.cpp
```diff
@@ -893,7 +897,7 @@ void DoSpeedBook()
 						yo -= SPLICONLENGTH;
 					}
 				}
-				spell <<= (__int64)1;
+				spell <<= (Uint64)1;
```
julealgon, line 900: Consider using the native literal as it avoids casting: `1ULL`

Source/drlg_l1.cpp
```diff
@@ -1352,19 +1352,17 @@ static BOOL L5checkRoom(int x, int y, int width, int height)
...
 	int cw, ch, cx1, cy1, cx2;
 
-	dirProb = random_(0, 4);
+	int dirProb = random_(0, 4);
+	int num = 0;
 
-	switch (dir == 1 ? dirProb != 0 : dirProb == 0) {
-	case FALSE:
-		num = 0;
+	if ((dir != 1 || dirProb == 0) && (dir == 1 || dirProb != 0)) {
```
julealgon, line 1362: This looks hard to read. Would this be equivalent?
`(dir == 1 && dirProb == 0) || (dir != 1 && dirProb !=0)`

Source/inv.cpp
```diff
@@ -1776,39 +1772,37 @@ void CheckInvCut(int pnum, int mx, int my, bool automaticMove)
 						case PC_BARBARIAN:
 							PlaySFX(PS_WARR15, false);
 							break;
-
```
julealgon, line 1779: Can we change the formatting standard to leave a blank line between cases? It is much harder to read without them IMHO :(

Source/items.cpp
```diff
@@ -4558,8 +4562,6 @@ BOOL PremiumItemOk(int i)
 		rv = FALSE;
 	if (AllItemsList[i].itype == ITYPE_GOLD)
 		rv = FALSE;
-	if (AllItemsList[i].itype == ITYPE_FOOD)
```
julealgon, line 4561: Would be nice to have this `ITYPE_FOOD` in its own PR.

Source/multi.cpp
```diff
@@ -433,7 +433,7 @@ static void multi_process_tmsgs()
 	int cnt;
 	TPkt pkt;
 
-	while (cnt = tmsg_get((BYTE *)&pkt, 512)) {
+	while ((cnt = tmsg_get((BYTE *)&pkt, 512))) {
```
julealgon, line 436: The extra parenthesis make no sense to me here. Any specific reason why we are mandating them?

Source/objects.cpp
```diff
@@ -3978,10 +3981,10 @@ void OperateShrine(int pnum, int i, int sType)
 			return;
 		InitDiabloMsg(EMSG_SHRINE_GLOWING);
 		int playerXP = plr[myplr]._pExperience;
-		int xpLoss, magicGain;
+		Sint32 xpLoss, magicGain;
```
julealgon, line 3984: Isn't this supposed to be `Sint64`?

Source/stores.cpp
```diff
@@ -969,6 +965,9 @@ void S_StartConfirm()
 	case STORE_SREPAIR:
 		strcpy(tempstr, "Are you sure you want to repair this item?");
 		break;
+	default:
+		strcpy(tempstr, "Are you sure?");
```
julealgon, line 969: Is this possible? I wouldn't create logic for impossible enum values. Either we should just fail somehow, or do nothing at best.

Adding a different string here may mislead people into thinking there is an extra behavior.

SourceS/miniwin.h
```diff
@@ -14,13 +14,6 @@
 #include <string.h>
 #include <time.h>
 
-#ifndef _WIN32
-#define __int8 char
-#define __int16 short
-#define __int32 int
-#define __int64 long long __attribute__((aligned(8)))
-#endif
```
julealgon, line 22: Nice.

SourceT/writehero_test.cpp
```diff
@@ -172,10 +174,10 @@ static void PackPlayerTest(PkPlayerStruct *pPack)
 	for (auto i = 0; i < 7; i++)
 		pPack->InvBody[i].idx = -1;
 	for (auto i = 0; i < MAXBELTITEMS; i++)
-		PackItemFullRejuv(pPack->SpdList+i, i);
+		PackItemFullRejuv(pPack->SpdList + i, i);
 	for (auto i = 1; i < 37; i++) {
-		if(spelldat_vanilla[i] != -1) {
-			pPack->pMemSpells |= (__int64)1 << (i-1);
+		if (spelldat_vanilla[i] != -1) {
+			pPack->pMemSpells |= (Uint64)1 << (i - 1);
```
julealgon, line 180: `1ULL` 

FAKE EDIT: Can't we just reuse the function for getting the bit of a spell, or even the macro?

3rdParty/Radon/Radon/include/Key.hpp
glebm: It's better to pass primitive types by value (just `float`)

SourceX/qol.cpp
glebm: To make the code easier to understand (see what next line does)

Source/inv.cpp
glebm: Blank lines between cases are unusual and there's already the `break;` to visually separate the cases.

Source/multi.cpp
glebm: Extra parenthesis are the conventional way to indicate that the assignment is intentional, i.e. `=` is not a typo of `==`

SourceX/qol.cpp
glebm: Placing it after the if-guard makes it clear that the code does not overflow the unsigned integer

julealgon: > To make the code easier to understand

This PR is exclusively to fix warnings however, not code readability. I'm fine with your suggestion, but it should ideally be done in a separate change (at least a separate commit).

glebm: It's a non-trivial "warning fix only" change because an unsigned integer behaves differently from a signed integer here.

Source/multi.cpp
glebm: Are you not familiar with this convention? I remember it existing 20 years ago and it's part of every codebase I've ever worked on and is built into C and C++ linters.

I'm not sure how it came to be but it's rather well-established

glebm: An alternative to this that will read more clear to those unfamiliar with the convention is to make comparison explicit:

```c++
while ((cnt = tmsg_get((BYTE *)&pkt, 512)) != 0) {
```

glebm: When the comparison is explicit, the extra parentheses are required, which may be where this convention came from.

I personally prefer explicit comparisons in C++ code but this is a rather controversial opinion (especially in C code).

defs.h
AJenbo: Where?

AJenbo:
> > AJenbo removed the request for review from julealgon 3 hours ago
> 
> Why? oO

To not bother you, you can still review if you feel like it, I just don't consider it a blocker for me any more.

SourceS/miniwin.h
AJenbo: yeah :)

Source/stores.cpp
AJenbo: Hm it was just to make it easy for anyone why would try to extend the stores later. I can remove it again if you think that's better.

Source/multi.cpp
AJenbo: Actually that's what our code style states so I should just apply this.

Source/drlg_l1.cpp
AJenbo: I don't think so but I can setup some test and check some variants.

Source/inv.cpp
galaxyhaxz: I don't have the same compiler setup to test, but I'd imagine just using an empty default case should fix the warning.

Source/automap.cpp
AJenbo: It would be incorrect for the player to have DIR_OMNI as that is the unitialized direction meaningn they aren't looking in any direction (not possible for real people). But since it's a rather minor issues I think it's to agressive to terminate the application, so I was thinking it would be ok to just assume that they are looking in the direction that they spawn in.

3rdParty/Radon/Radon/include/Key.hpp
AJenbo: I guess that can be done.

Source/inv.cpp
AJenbo: This was the best I could come up with :D

galaxyhaxz: To be fair for class SFX code I replaced that with a single function and a global table. So the whole switch case is replaced with `GetPlayerSfx(VOICE_CANTUSE)` and the like.

AJenbo: > I don't have the same compiler setup to test, but I'd imagine just using an empty default case should fix the warning.

It would, but I prefer this as it would war anyone adding new classes that they need to add it to this switch.

Source/objects.cpp
AJenbo: PlayerXP is singed 32bit int so multiplied by 0.95 should still fit in a Sint32 variable. The original __int64 casting appears mostly redundant.

SourceT/writehero_test.cpp
AJenbo: I think it's sometimes better to have the test logic disconnected from the application logic so you have less chance of reusing broken logic to very that it's broken.

Source/items.cpp
AJenbo: I split it into its own commit, best I can do here.

Source/drlg_l1.cpp
AJenbo: It's the same, applied the change.

defs.h
julealgon: `spells.cpp` `GetSpellBitmask`.

Source/automap.cpp
julealgon: Thanks for clarifying. We'll make these constraints stronger over time so that invalid values are made impossible.

Source/stores.cpp
julealgon: >  I can remove it again if you think that's better.

I think it is best we go for YAGNI here and avoid implementing "stub" things when we can.

SourceT/writehero_test.cpp
julealgon: I actually didn't notice this was a test at first. I agree with your reasoning.

PR #1214, opened 2021-03-19
OpenDingux: rg99 port (experimental)

glebm:
work-in-progress, need to improve RAM usage further before this is playable even without sound.

Let's merge this as an experimental platform (without a release) for now.


AJenbo:
I'm just going to wait with merging this till we have `STREAM_ALL_AUDIO` implemented :D
Maybe also til it has been tested on rg99? :thinking:

glebm:
Yeah I'll keep it as draft until all of that is done!

AJenbo:
We can save another 1.1MB by getting rid of `InitPlrGFXMem` and use `LoadFileInMem` instead of `LoadFileWithMem` in `LoadPlrGFX`. Or alternately 0.6MB by taking class in to consideration in `GetPlrGFXSize`.

glebm:
Looks like for RG99 we'll also have to use a hardware 8-bit surface with YUV palette - no other modes support hardware scaling

Something I'll look into in 2-3 weeks once the device arrives

glebm:
Current status: it "works" but there just isn't enough RAM on the device to run DevilutionX as it is today at any acceptable performance

There is about 25 MiB free RAM and 20 MiB swap

AJenbo:
how much have we won on structure and enum changes so fare?

AJenbo:
Oh wow, with out multiple stream support this port would never have stood a chance :(

glebm:
The video streaming PR (#1778) allows intro videos to work just fine! Too bad it's Linux/BSD-only and so won't work on the PSP

AJenbo:
So close

AJenbo:
How is it looking at this point?

glebm:
FPS went from 17 to 19 in town when I switched from SWSURFACE to an unbuffered HWSURFACE
Unbuffered HWSURFACE is fine here because there is an internal buffer anyway

john-parton:
Does the RG99 use zram swap like the RS90?

You can try rebuilding OpenDingux and put the swap on an SD card. That will free up memory.

glebm:
@john-parton Yeah it does. Currently hoping for Clang-LTO kernel to get fixed (unknown but likely substantial savings), and to convince Paul to remove PRINTK and BUG from the kernel config (saves 800 KiB).

Putting the swap on the SD card is unlikely to happen by default for RG99 OpenDingux (installing a different image just to play Diablo is a bit much). Perhaps it can be made a setting in the current OpenDingux.

glebm:
Current issues:
1. Not enough RAM. Need to fix in OpenDingux: either disable `CONFIG_PRINTK` or wait for kernel LTO.
2. Intro video colors are now broken. No changes on DevilutionX side so it's likely a regression in OpenDingux.

Progress: Now at 23 FPS (full-speed for Diablo, which originally ran at 20 FPS)

glebm:
Intro video colors are fixed by #3737 but I've discovered a bug in OpenDingux that causes the game to crash after the intro video (https://github.com/OpenDingux/SDL/issues/4)

glebm:
The crash bug has been fixed, so now we're back to everything working perfectly except there's not enough RAM

AJenbo:
Any idear what the RAM usage looks like at this point?

glebm:
The heap profile says 8.6 MiB (but some stuff isn't measured there). I think it may have gotten a bit worse as I can't get into the dungeon with a warrior anymore.

AJenbo:
I feared that was the case, a few of the optional parts should be better as of late, but the core has grown a bit.

glebm:
Let's merge this as an experimental platform!

It is far from being fully supported yet but it's a fun thing to try to optimize things down to RG99 requirements.

AJenbo (approved):
Couldn't find this PR last I looked, thought it might already have been merged :sweat_smile:

PR #1217, opened 2021-03-20
Hellfire: Better error if some MPQs are missing

glebm:
image

Makes the issue more obvious.

### DIFF
--- a/Source/init.cpp
+++ b/Source/init.cpp
@@ -192,6 +192,13 @@ void init_archives()
 	hfvoice_mpq = init_test_access(paths, "hfvoice.mpq", "DiabloInstall", 8500, FS_PC);
 	hfopt1_mpq = init_test_access(paths, "hfopt1.mpq", "DiabloInstall", 8600, FS_PC);
 	hfopt2_mpq = init_test_access(paths, "hfopt2.mpq", "DiabloInstall", 8610, FS_PC);
+
+	if (gbIsHellfire && (hfmonk_mpq == NULL || hfmusic_mpq == NULL || hfvoice_mpq == NULL))
+	{
+		UiErrorOkDialog("Some Hellfire MPQs are missing", "Not all Hellfire MPQs were found.\nPlease copy all the hf*.mpq files.");
+		app_fatal(NULL);
+	}
+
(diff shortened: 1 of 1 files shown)


galaxyhaxz:
"Not all MPQs were found. Deleting C:/windows in 10 seconds..."

Well, I guess people will be sure they have everything setup right now xD

AJenbo (approved):

Source/init.cpp
```diff
@@ -192,6 +192,13 @@ void init_archives()
 	hfvoice_mpq = init_test_access(paths, "hfvoice.mpq", "DiabloInstall", 8500, FS_PC);
 	hfopt1_mpq = init_test_access(paths, "hfopt1.mpq", "DiabloInstall", 8600, FS_PC);
 	hfopt2_mpq = init_test_access(paths, "hfopt2.mpq", "DiabloInstall", 8610, FS_PC);
+
+	if (gbIsHellfire && (hfmonk_mpq == NULL || hfmusic_mpq == NULL || hfvoice_mpq == NULL))
+	{
+		UiErrorOkDialog("Some Hellfire MPQs are missing", "Not all Hellfire MPQs were found.\nPlease copy all the hf*.mpq files.");
+		app_fatal(NULL);
+	}
```
julealgon, line 200: Won't this prevent one from playing with fully extracted MPQs but without having the MPQs in the folder?

AJenbo: You were already required to have the diabdat.mpq and hellfire.mpq. Requiring 3 more (that you already have if you have the files) in order to not have people confused about the game crashing and reporting bugs is defiantly worth it. If you are trying to save on disk space then just have 5 empty mpq files.

julealgon: > You were already required to have the diabdat.mpq and hellfire.mpq.

Ah ok, I thought we supported having only the extracted files and no MPQs.

> Requiring 3 more (that you already have if you have the files) in order to not have people confused about the game crashing and reporting bugs is defiantly worth it.

Considering your clarification, absolutely agree.

> If you are trying to save on disk space then just have 5 empty mpq files.

Wasn't trying to save space or anything like that, just wondering if the change would make us less flexible.

Issue #1218, opened 2021-03-20
Debian package build problem


Vitexus:
**Important information**
Debian / 1.1.0 / compiles

**Describe**
Debian package wont build

**To Reproduce**
1. git clone  ...
2. ln -s Packaging/debian to ./debian
3. debuild -us -uc

```shell
vitex@exiv:~/Projects/tmp/devilutionX$ debuild -us -uc
 dpkg-buildpackage -us -uc -ui
dpkg-buildpackage: info: source package devilutionx
dpkg-buildpackage: info: source version 1.0.1
dpkg-buildpackage: info: source distribution UNRELEASED
dpkg-buildpackage: info: source changed by Goran Vidovic (gogo) <trebelnik2@gmail.com>
 dpkg-source --before-build .
dpkg-buildpackage: info: host architecture amd64
 fakeroot debian/rules clean
dh clean --buildsystem=cmake --builddirectory=obj-x86_64-linux-gnu
   dh_auto_clean -O--buildsystem=cmake -O--builddirectory=obj-x86_64-linux-gnu
   dh_autoreconf_clean -O--buildsystem=cmake -O--builddirectory=obj-x86_64-linux-gnu
   dh_clean -O--buildsystem=cmake -O--builddirectory=obj-x86_64-linux-gnu
 dpkg-source -b .
dpkg-source: info: using source format '3.0 (native)'
dpkg-source: info: building devilutionx in devilutionx_1.0.1.tar.xz
dpkg-source: info: building devilutionx in devilutionx_1.0.1.dsc
 debian/rules build
make: „build“ je aktuální.
 fakeroot debian/rules binary
dh binary --buildsystem=cmake --builddirectory=obj-x86_64-linux-gnu
   dh_update_autotools_config -O--buildsystem=cmake -O--builddirectory=obj-x86_64-linux-gnu
   dh_autoreconf -O--buildsystem=cmake -O--builddirectory=obj-x86_64-linux-gnu
   debian/rules override_dh_auto_configure
make[1]: Vstupuje se do adresáře „/home/vitex/Projects/tmp/devilutionX“
dh_auto_configure -- \
        -DNIGHTLY_BUILD=ON \
        -DVERSION_NUM=oject \
        -DVERSION_SUFFIX=-lutionX \
        #-DSPAWN=ON
[...]

PR #1219, opened 2021-03-20
Debian package build problem #1218

Vitexus:
Now we can use git to obtain commit version and app version


AJenbo:
Thanks this makes a lot more sense.

PR #1230, opened 2021-03-21
Handle D-Pad / Left stick in UI every frame

glebm:
Previously this was handled as part of `game_logic`, reducing responsiveness and making it impossible to run where `game_logic` isn't running, e.g. stores.

Bonus: Enables D-Pad / Left stick navigation in stores because the above makes it trivial

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -91,6 +91,10 @@ BOOL was_window_init = false;
 BOOL was_ui_init = false;
 BOOL was_snd_init = false;
 
+// Controller support:
+extern void plrctrls_every_frame();
+extern void plrctrls_after_game_logic();
+
 static void print_help_and_exit()
 {
 	printInConsole("Options:\n");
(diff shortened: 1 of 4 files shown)


AJenbo:
Oh nice, this has been an annoying inconsistency.

Source/diablo.cpp
```diff
@@ -322,6 +326,7 @@ static void run_game_loop(unsigned int uMsg)
 		if (!gbRunGame)
 			break;
 		if (!nthread_has_500ms_passed()) {
+			plrctrls_every_frame();
```
AJenbo, line 329: Why not call this from ProcessInput or FetchMessage?

glebm: Done

AJenbo (approved):

yuripourre:
Thanks @glebm, that will help me. I was having trouble with that particular code.

PR #1233, opened 2021-03-21
Inventory and Stores cleanup

julealgon:
This PR cleans some of the inventory logic by reusing existing methods, adding comments, and cleaning up function signatures, and then leverages those changes in stores.cpp to avoid code duplication.

No behavior change should be observed with these changes.

Review per commit is recommended.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1335,9 +1335,7 @@ void CheckInvPaste(int pnum, int mx, int my)
 				SetCursor_(plr[pnum].HoldItem._iCurs + CURSOR_FIRSTITEM);
 			else
 				SetICursor(plr[pnum].HoldItem._iCurs + CURSOR_FIRSTITEM);
-			done2h = FALSE;
-			for (i = 0; i < NUM_INV_GRID_ELEM && !done2h; i++)
-				done2h = AutoPlaceItemInInventorySlot(pnum, i, plr[pnum].HoldItem, true);
+			done2h = AutoPlaceItemInInventory(pnum, plr[pnum].HoldItem, true);
 			plr[pnum].HoldItem = tempitem;
 			if (pnum == myplr)
 				SetCursor_(plr[pnum].HoldItem._iCurs + CURSOR_FIRSTITEM);
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -2211,7 +2211,7 @@ void AutoGetItem(int pnum, int ii)
 		}
 	}
 	plr[pnum].HoldItem = items[ii];
-	RespawnItem(ii, TRUE);
+	RespawnItem(&items[ii], TRUE);
 	NetSendCmdPItem(TRUE, CMD_RESPAWNITEM, items[ii]._ix, items[ii]._iy);
 	plr[pnum].HoldItem._itype = ITYPE_NONE;
 }
@@ -2411,7 +2411,7 @@ int InvPutItem(int pnum, int x, int y)
 	items[ii] = plr[pnum].HoldItem;
 	items[ii]._ix = x;
 	items[ii]._iy = y;
-	RespawnItem(ii, TRUE);
+	RespawnItem(&items[ii], TRUE);
 
 	if (currlevel == 21 && x == CornerStone.x && y == CornerStone.y) {
 		CornerStone.item = items[ii];
@@ -2498,7 +2498,7 @@ int SyncPutItem(int pnum, int x, int y, int idx, WORD icreateinfo, int iseed, in
 
 	items[ii]._ix = x;
 	items[ii]._iy = y;
-	RespawnItem(ii, TRUE);
+	RespawnItem(&items[ii], TRUE);
 
 	if (currlevel == 21 && x == CornerStone.x && y == CornerStone.y) {
 		CornerStone.item = items[ii];
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -927,12 +927,12 @@ bool AutoPlaceItemInInventory(int playerNumber, const ItemStruct &item, bool per
 	return done;
 }
 
-BOOL AutoPlaceItemInInventorySlot(int pnum, int ii, const ItemStruct &item, BOOL saveflag)
+bool AutoPlaceItemInInventorySlot(int pnum, int ii, const ItemStruct &item, bool saveflag)
 {
 	int i, j, xx, yy;
-	BOOL done;
+	bool done;
 
-	done = TRUE;
+	done = true;
 	yy = 10 * (ii / 10);
 	if (yy < 0) {
 		yy = 0;
@@ -941,7 +941,7 @@ BOOL AutoPlaceItemInInventorySlot(int pnum, int ii, const ItemStruct &item, BOOL
 	InvXY itemSize = GetInventorySize(item);
 	for (j = 0; j < itemSize.Y && done; j++) {
 		if (yy >= NUM_INV_GRID_ELEM) {
-			done = FALSE;
+			done = false;
 		}
 		xx = ii % 10;
 		if (xx < 0) {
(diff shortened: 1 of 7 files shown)


Source/items.h
```diff
@@ -299,7 +299,7 @@ void SpawnRewardItem(int itemid, int xx, int yy);
 void SpawnMapOfDoom(int xx, int yy);
 void SpawnRuneBomb(int xx, int yy);
 void SpawnTheodore(int xx, int yy);
-void RespawnItem(int i, BOOL FlipFlag);
+void RespawnItem(ItemStruct &item, BOOL FlipFlag);
```
AJenbo, line 302: I think I would prefer `*` over `&` for references as it makes it clear on the calling end that it is by reference.

Source/msg.cpp
```diff
@@ -1392,7 +1392,7 @@ static DWORD On_REQUESTGITEM(TCmd *pCmd, int pnum)
 				if (p->bPnum != myplr)
 					SyncGetItem(p->x, p->y, p->wIndx, p->wCI, p->dwSeed);
 				else
-					InvGetItem(myplr, ii);
+					InvGetItem(myplr, items[ii]);
```
AJenbo, line 1395: I find `&items[ii]` more explicit.

Source/inv.cpp
```diff
@@ -2111,7 +2084,7 @@ void CleanupItems(int ii)
 
 	int i = 0;
 	while (i < numitems) {
-		if (itemactive[i] == ii) {
+		if (itemactive[i] == FindGetItem(item.IDidx, item._iCreateInfo, item._iSeed)) {
```
AJenbo, line 2087: What's the intent of this change?

Source/items.cpp
```diff
@@ -1388,7 +1388,7 @@ void CreatePlrItems(int p)
 		{
 			SetPlrHandItem(&plr[p].HoldItem, IDI_WARRCLUB);
 			GetPlrHandSeed(&plr[p].HoldItem);
-			AutoPlace(p, 0, 1, 3, TRUE);
+			AutoPlaceItemInInventory(p, plr[p].HoldItem, true);
```
AJenbo, line 1391: What's happening here?

AJenbo (changes requested):
I prefer using by pointer rather than by reference. Other than that this looks good. Didn't even realize this was also duplicated in stores.cpp

Source/items.h
julealgon: What? But the purpose of references is exactly to abstract the need for specifying pointers in the first place.

Are you really sure about this? Can you point me to supporting documentation on it?

I was actively avoiding raw pointers as much as possible everywhere, so seeing this suggestion doesn't make sense to me.

Source/inv.cpp
julealgon: We don't have the index for the item anymore, so we have to fetch it based on the item.

julealgon:
> I prefer using by pointer rather than by reference.

Please can you let me know why? I always thought that pointers were evil and references were introduced to simplify things overall, which to me is the case.

Source/items.h
AJenbo: "supporting documentation" on my personal preference? I can write some if you would like that.

> I was actively avoiding raw pointers as much as possible everywhere, so seeing this suggestion doesn't make sense to me.

why?

Source/inv.cpp
AJenbo: this could potentially find a duplicate, maybe better to also pass the index?

at least we should only do the search ones (outside the loop)

glebm:
Raw pointers are OK if the codebase only uses them to mean unowned pointers.

This is not the current state of the codebase but is where we eventually want to get to.

When something is const and cannot be null, a const reference is always better.

For mutable argument, until last year, the Google style guide recommended using raw pointers.
They have the advantage that you cannot pass in a temporary into such an argument (for reference you'd need to explicitly delete the rvalue-reference overload)
These days the Google Style Guide allows both.

Source/inv.cpp
julealgon: > this could potentially find a duplicate

I'm using the same function used in the duplication logic though. You sure this could get a duplicated item?

> at least we should only do the search ones (outside the loop)

Indeed, that was a big oversight. I'll change it.

julealgon:
> Raw pointers are OK if the codebase only uses them to mean unowned pointers.
> ...
> These days the Google Style Guide allows both

It's not clear to me your recommendation here. Being "OK" usually means "its not terrible...." to me.

> When something is const and cannot be null, a const reference is always better.

But isn't the "never null" a big deal of references? In all the cases that I used a reference parameter, the item should not be null. Wouldn't I need to start checking for `!= nullptr` everywhere if I start using raw pointers? Those functions are global after all.

Source/items.h
julealgon: > "supporting documentation" on my personal preference?

No, I meant technical reasons. I'm not basing my decision to use one or the other on "personal preference" here, but based on the fact that references are more robust and lead to less errors than raw pointers (in my understanding).

glebm:
Mutable reference: cannot be null but the language allows passing a temporary which is almost never correct.

Mutable pointer: can be null but cannot be a temporary.

It's a stylistic preference at this point.

Source/inv.cpp
AJenbo: What duplication logic ;)

julealgon: > What duplication logic ;)

The item duplication checks use this method's conditions to define what a duplicate item means.

julealgon:
@AJenbo , just to confirm: final say is "raw pointers for mutable arguments" then?

AJenbo:
> * Reference: heavy_minus_sign mutability not always obvious at the call site

This is my issue with it if that wasn't clear :shrug: 

I hadn't thought about the fact that it cannot be NULL which is a really good argument for it :shrug:

AJenbo:
> @AJenbo , just to confirm: final say is "raw pointers for mutable arguments" then?

Yeah let's say that :+1:

glebm:
> Another argument against it is that we can't provide C-headers for functions that use reference.

I'd say that if we offers a C API it should be completely separate anyway (and we want our functions in a namespace which C can't do) so that isn't a consideration.

julealgon:
Temporarily converting to draft while I clean this up.

It now depends on #1282 (trying to break it a bit to split the diffs and making it easier to review).

Source/inv.cpp
AJenbo: The code was a bit faulty and could end up marking legit items as duplicates (especially uniques), since we fixed the duplication bug the faulty anti-duplication code has been removed.

Source/items.cpp
julealgon: I hadn't seen this comment. Do you still have any questions about it?

Basically, we just remove the hardcoded grid position and let the function take care of placing the item in the inventory using the standard iteration logic.

AJenbo: Yeah i eventually figured it out when going over things by commit :P

julealgon:
Rebased and adjusted. Should be ready for another review round.

Source/inv.cpp
```diff
@@ -2189,7 +2162,7 @@ void AutoGetItem(int pnum, int ii)
 	}
 
 	if (done) {
-		CleanupItems(ii);
+		CleanupItems(&plr[pnum].HoldItem, ii);
```
AJenbo, line 2165: this would previously have been items[ii], so could you explain this change to me ?

julealgon: They are the same. A little bit above we assign holditem to items[ii]

AJenbo: Oh i see now on line 2143, I hade missed that previously when going over despite looking.

julealgon: Admittedly, this is not very intuitive... it all goes back to the damn "holditem". It should be used to represent "the item in the cursor" but it's used as this temporary storage everywhere. Once we get rid of it and make those "lists of ints" lists of pointers, i think this kind of stuff will be dramatically simplified.

One step at a time though...

PR #1234, opened 2021-03-21
Fix right stick motion with the FPS limiter off

glebm:
1. Moves right stick movement handler to run on every frame.
2. Changes hi-res accumulators to `float` to ensure they're always
   incremented.
3. Avoid calling `SetCursorPos` within the same SDL tick:
   this caused SDL to process all controller events before mouse events
  (out-of-order).


glebm:
This also makes right-stick-as-mouse work in `gmenu`

PR #1235, opened 2021-03-21
CMake: Fix sodium static build

glebm:
The 4th argument to `cmake_dependent_option` is when to enable the option, not when to disable it like I'd thought.
https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html


AJenbo:
The build still asks for libsodium-23.dll

glebm:
Thanks for checking and catching this!
Really fixed now:

> -- Found sodium: /usr/i686-w64-mingw32/lib/libsodium.a (found version "1.0.18")

PR #1236, opened 2021-03-21
Extract Gamepad class

yuripourre:
This is the first step to enable local multiplayer. I am working on more refactorings to enable local multiplayer and will be creating PRs so it will not be too hard to review.

I moved the stick variables from global scope to a class `Gamepad`. I plan to extract buttons as well and turn gamepad into an array. So will be possible to have local multiplayer.


glebm (changes requested):
The controller code is structured like this:

1. Individual devices are implemented in controls/devices/.
2. `controls/controller.h` provides a unified interfaces to them.

The individual devices already implement what you want but the unified controller interface doesn't.

Please rename the class to `Controller` and place it into `controls/controller.h`

The reason I prefer `Controller` name over `Gamepad` is that game controllers are not always gamepads (consistent with SDL_GAMECONTROLLER).

yuripourre:
> The controller code is structured like this:
> 
>     1. Individual devices are implemented in controls/devices/.
> 
>     2. `controls/controller.h` provides a unified interfaces to them.
> 
> 
> The individual devices already implement what you want but the unified controller interface doesn't.
> 
> Please rename the class to `Controller` and place it into `controls/controller.h`
> 
> The reason I prefer `Controller` name over `Gamepad` is that game controllers are not always gamepads (consistent with SDL_GAMECONTROLLER).

Thanks for reviewing it @glebm. Oh! It is more clear now, I am having a hard time understanding all the stack of calls, I plan to have some helper methods to help with that.

I just updated the code.

glebm (approved):

AJenbo (approved):

PR #1241, opened 2021-03-22
Build AppImages for Linux on CircleCi

Trihedraf:
Adds the building of AppImages for Linux x86 and x64 on the automated CircleCi builds.


glebm:
This is very cool!
One minor issue is that Support and Credits screen are empty (`CharisSILB.ttf` is missing or in the wrong place)

Trihedraf:
With #1253 and #1254 will solve the CharisSILB.ttf not working.

PR #1242, opened 2021-03-22
.clang-tidy: Document run-clang-tidy

glebm:
Turns out clang-tidy comes with a Python script that makes running it very easy and parallelizes the checks.

https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clang-tidy/tool/run-clang-tidy.py

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,13 +1,17 @@
 ---
 # clang-tidy configuration
 #
-# clang-tidy can be run manually over a set of files like this:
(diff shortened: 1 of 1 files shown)


AJenbo:
Wonder when we should make this part of the CI :thinking:

glebm:
Probably once we've fixed all the warnings, after the 1.2.0 release

AJenbo:
:cold_sweat:

PR #1244, opened 2021-03-22
Fix SDL_Rect warnings on SDL1

glebm:


AJenbo:
Thanks, was getting exhausted by the thought of tracking this down :D

Source/engine.h
```diff
@@ -137,7 +137,14 @@ struct CelOutputBuffer {
 	 */
 	CelOutputBuffer subregion(Sint16 x, Sint16 y, Uint16 w, Uint16 h) const
 	{
-		return CelOutputBuffer(surface, SDL_Rect { region.x + x, region.y + y, w, h });
+		// In SDL1 SDL_Rect x and y are Sint16. Cast explicitly to avoid a compiler warning.
+		using CoordType = decltype(SDL_Rect {}.x);
```
julealgon, line 141: Oh wow... is this the equivalent of `typeof` in C#?

glebm: I don't know C# but `decltype` returns the type of any given expression without evaluating it

PR #1250, opened 2021-03-22
Unify controller implementations

yuripourre:
This is another step toward Couch co-op. It's an attempt to make all controller implementations similar, I created a `KeyboardController` class so it could be treated the same as the other implementations.

My C++ skills are a bit rusty so feel free to suggest improvements/modifications.

The next step will be having a single vector handling all the controllers so would be easier to assign each controller to a player.
I am trying to keep the PRs short so it's easier to review.

Comments are welcome!

### DIFF
--- a/SourceX/controls/controller.cpp
+++ b/SourceX/controls/controller.cpp
@@ -127,7 +127,7 @@ ControllerButtonEvent ToControllerButtonEvent(const SDL_Event &event)
 	}
 #endif
 
-	Joystick *const joystick = Joystick::Get(event);
+	const Joystick *const joystick = Joystick::Get(event)
 	if (joystick != NULL)
 		result.button = joystick->ToControllerButton(event);
 
--- a/SourceX/controls/controller.cpp
+++ b/SourceX/controls/controller.cpp
@@ -127,7 +127,7 @@ ControllerButtonEvent ToControllerButtonEvent(const SDL_Event &event)
 	}
 #endif
 
-	const Joystick *const joystick = Joystick::Get(event)
+	const Joystick *const joystick = Joystick::Get(event);
 	if (joystick != NULL)
 		result.button = joystick->ToControllerButton(event);
 
--- a/SourceX/controls/controller.cpp
+++ b/SourceX/controls/controller.cpp
@@ -6,6 +6,94 @@
 
 namespace dvl {
 
+AxisDirection Controller::GetLeftStickOrDpadDirection(bool allow_dpad)
+{
+	const float stickX = leftStickX;
+	const float stickY = leftStickY;
+
+	AxisDirection result { AxisDirectionX_NONE, AxisDirectionY_NONE };
+
+	if (stickY >= 0.5 || (allow_dpad && IsControllerButtonPressed(ControllerButton_BUTTON_DPAD_UP))) {
+		result.y = AxisDirectionY_UP;
+	} else if (stickY <= -0.5 || (allow_dpad && IsControllerButtonPressed(ControllerButton_BUTTON_DPAD_DOWN))) {
+		result.y = AxisDirectionY_DOWN;
+	}
+
+	if (stickX <= -0.5 || (allow_dpad && IsControllerButtonPressed(ControllerButton_BUTTON_DPAD_LEFT))) {
+		result.x = AxisDirectionX_LEFT;
+	} else if (stickX >= 0.5 || (allow_dpad && IsControllerButtonPressed(ControllerButton_BUTTON_DPAD_RIGHT))) {
+		result.x = AxisDirectionX_RIGHT;
+	}
+
+	return result;
+}
+
+void Controller::ScaleJoysticks()
+{
+	const float rightDeadzone = 0.07;
+	const float leftDeadzone = 0.07;
+
+	if (leftStickNeedsScaling) {
+		leftStickX = leftStickXUnscaled;
+		leftStickY = leftStickYUnscaled;
+		ScaleJoystickAxes(&leftStickX, &leftStickY, leftDeadzone);
+		leftStickNeedsScaling = false;
+	}
+
+	if (rightStickNeedsScaling) {
+		rightStickX = rightStickXUnscaled;
+		rightStickY = rightStickYUnscaled;
+		ScaleJoystickAxes(&rightStickX, &rightStickY, rightDeadzone);
+		rightStickNeedsScaling = false;
+	}
+}
+
+void Controller::ScaleJoystickAxes(float *x, float *y, float deadzone)
+{
+	//radial and scaled dead_zone
+	//http://www.third-helix.com/2013/04/12/doing-thumbstick-dead-zones-right.html
+	//input values go from -32767.0...+32767.0, output values are from -1.0 to 1.0;
+
+	if (deadzone == 0) {
+		return;
+	}
+	if (deadzone >= 1.0) {
+		*x = 0;
+		*y = 0;
+		return;
+	}
+
+	const float maximum = 32767.0f;
+	float analog_x = *x;
+	float analog_y = *y;
+	float dead_zone = deadzone * maximum;
+
+	float magnitude = std::sqrt(analog_x * analog_x + analog_y * analog_y);
+	if (magnitude >= dead_zone) {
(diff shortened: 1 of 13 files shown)


yuripourre:
Hey @AJenbo @glebm if you have some time, I will be glad to hear your thoughts.

SourceX/controls/controller.cpp
```diff
@@ -36,7 +127,7 @@ ControllerButtonEvent ToControllerButtonEvent(const SDL_Event &event)
 	}
 #endif
 
-	const Joystick *joystick = Joystick::Get(event);
+	Joystick *const joystick = Joystick::Get(event);
```
glebm, line 130: Why? You don't seem to be modifying `joystick`?

yuripourre: Yeah, It could be reverted, I just wanted to keep the same order as in controller. Do you prefer if I rollback?

glebm: They mean different things:

1. `const Joystick *joystick`: a non-const pointer to a const `Joystick`
2. `Joystick *const joystick`: a const pointer to a non-const `Joystick`

In case of 1, you can't modify the joystick, though you can assign another `const Joystick *` to the `joystick` variable.

In case of 2, you can modify the joystick but not reassign the variable.

glebm: The most const-correct version here would be

```c++
const Joystick *const joystick = Joystick::Get(event)
```

glebm (dismissed):

SourceX/controls/controller.cpp
yuripourre: Right! Done.

yuripourre:
@glebm Hey, I missed your approval, because there was a missing semicolon. If you don't mind approving again :D

glebm (approved):

yuripourre:
@glebm Thanks!

SourceX/display.cpp
```diff
@@ -151,9 +151,12 @@ bool SpawnWindow(const char *lpWindowName)
 	// Always try to initialize the first joystick.
 	Joystick::Add(0);
 #ifdef __SWITCH__
-	// TODO: There is a bug in SDL2 on Switch where it does not repport controllers on startup (Jan 1, 2020)
+	// TODO: There is a bug in SDL2 on Switch where it does not report controllers on startup (Jan 1, 2020)
 	GameController::Add(0);
 #endif
+#endif
+#ifdef HAS_KBCTRL == 1
```
AJenbo, line 158: Should have been
```cpp
#if HAS_KBCTRL == 1
```

SourceX/controls/controller.cpp
```diff
@@ -6,6 +6,94 @@
 
 namespace dvl {
 
+AxisDirection Controller::GetLeftStickOrDpadDirection(bool allow_dpad)
```
julealgon, line 9: Can we make this method `const`?

SourceX/controls/controller.cpp
```diff
@@ -6,6 +6,94 @@
...
+
+	float magnitude = std::sqrt(analog_x * analog_x + analog_y * analog_y);
+	if (magnitude >= dead_zone) {
+		// find scaled axis values with magnitudes between zero and maximum
+		float scalingFactor = 1.0 / magnitude * (magnitude - dead_zone) / (maximum - dead_zone);
+		analog_x = (analog_x * scalingFactor);
+		analog_y = (analog_y * scalingFactor);
+
+		// clamp to ensure results will never exceed the max_axis value
+		float clamping_factor = 1.0f;
```
julealgon, line 79: `const`?

SourceX/controls/controller.cpp
```diff
@@ -22,9 +110,12 @@ ControllerButtonEvent ToControllerButtonEvent(const SDL_Event &event)
 	}
 
 #if HAS_KBCTRL == 1
-	result.button = KbCtrlToControllerButton(event);
-	if (result.button != ControllerButton_NONE)
-		return result;
+	KeyboardController *const keyboardController = KeyboardController::Get(event);
```
julealgon, line 113: Why not use a reference?

SourceX/controls/controller.cpp
```diff
@@ -22,9 +110,12 @@ ControllerButtonEvent ToControllerButtonEvent(const SDL_Event &event)
 	}
 
 #if HAS_KBCTRL == 1
-	result.button = KbCtrlToControllerButton(event);
-	if (result.button != ControllerButton_NONE)
-		return result;
+	KeyboardController *const keyboardController = KeyboardController::Get(event);
+	if (keyboardController != NULL) {
```
julealgon, line 114: Should we compare to `nullptr` for new code?

SourceX/controls/controller.h
```diff
@@ -2,6 +2,8 @@
 
 #include <SDL.h>
 
+#include "all.h"
```
julealgon, line 5: Can we avoid using `all.h` for new code?

SourceX/controls/controller.h
```diff
@@ -19,6 +21,19 @@ class Controller {
 	float leftStickXUnscaled, leftStickYUnscaled, rightStickXUnscaled, rightStickYUnscaled;
 	// Whether stick positions have been updated and need rescaling.
 	bool leftStickNeedsScaling, rightStickNeedsScaling;
+
+	// Returns direction of the left thumb stick or DPad (if allow_dpad = true).
+	AxisDirection GetLeftStickOrDpadDirection(bool allow_dpad = true);
+
+	// Normalize joystick values
+	void ScaleJoysticks();
```
julealgon, line 29: Should we consider renaming the method to `NormalizeJoysticks`? The comment seems to indicate the current name is less than ideal.

SourceX/controls/controller.h
```diff
@@ -19,6 +21,19 @@ class Controller {
...
+
+	// Returns direction of the left thumb stick or DPad (if allow_dpad = true).
+	AxisDirection GetLeftStickOrDpadDirection(bool allow_dpad = true);
+
+	// Normalize joystick values
+	void ScaleJoysticks();
+
+	// NOTE: Not idempotent because of how it handles axis triggers.
+	// Must be called exactly once for each SDL input event.
+	ControllerButton ToControllerButton(const SDL_Event &event) const;
```
julealgon, line 33: Doesn't a `const` method imply idempotency?

SourceX/controls/devices/kbcontroller.h
```diff
@@ -7,15 +7,25 @@
...
+	static KeyboardController *Get(const SDL_Event &event);
+
+	ControllerButton ToControllerButton(const SDL_Event &event) const;
 
-bool ProcessKbCtrlAxisMotion(const SDL_Event &event);
+	//bool ProcessKbCtrlAxisMotion(const SDL_Event &event);
+	bool ProcessAxisMotion(const SDL_Event &event);
+};
+
+bool IsKbCtrlButtonPressed(ControllerButton button);
```
julealgon, line 28: `const`?

SourceX/controls/controller.cpp
glebm: This can be null, references cannot be null

julealgon: Ah yeah... makes sense, I forgot about that.

SourceX/controls/controller.cpp
AJenbo: sure but simple to replace, so don't want to enforce before we are consistent.

SourceX/controls/controller.h
AJenbo: i don't mind having to clean it up

SourceX/controls/devices/kbcontroller.h
glebm: Where?

glebm: There's no need for `const` on the argument -- it is passed by value (whether the method modifies it or not is an implementation detail)
There's no need for `const` on the return type -- it is returned by value (what the caller does with the result is the caller's business)

julealgon: I meant `const` as in `pure`, but I realize now this is not a method but a global function. Disregard.

PR #1253, opened 2021-03-23
Add --ttf-dir to specify location of path to ttf

Trihedraf:


galaxyhaxz:
@AJenbo Couldn't you just embed the TTF file inside the devilutionx.mpq to save users the hassle?

malvarenga123:
Just letting you guys know there is a newer version of the font (5.000) available: https://software.sil.org/charis/download/
The one that comes with DevilutionX is 4.106.

Trihedraf:
@malvarenga123 Version 5.000 is messed up at least for DevilutionX

V4.106

V5.000

AJenbo:
I guess they dropped monochrome hints :/

Source/diablo.cpp
```diff
@@ -144,6 +145,8 @@ static void diablo_parse_flags(int argc, char **argv)
 			SetPrefPath(argv[++i]);
 		} else if (strcasecmp("--config-dir", argv[i]) == 0) {
 			SetConfigPath(argv[++i]);
+		} else if (strcasecmp("--ttf-file", argv[i]) == 0) {
```
AJenbo, line 148: The command is `--font-file`, but sets the `font path`, which one was intended?

Source/diablo.cpp
```diff
@@ -103,6 +103,7 @@ static void print_help_and_exit()
 	printInConsole("    %-20s %-30s\n", "--data-dir", "Specify the folder of diabdat.mpq");
 	printInConsole("    %-20s %-30s\n", "--save-dir", "Specify the folder of save files");
 	printInConsole("    %-20s %-30s\n", "--config-dir", "Specify the location of diablo.ini");
+	printInConsole("    %-20s %-30s\n", "--ttf-file", "Specify the location of CharisSLIB.ttf");
```
AJenbo, line 106: Best not to put the specific font tile name here. I can be different depending on builds.

Source/diablo.cpp
Trihedraf: I originally set it up to do the path but I could not get the TTF_FONT_NAME to append to the path so I changed it to file. But I guess I should change the variable from ttfPath to ttfFile?

Source/diablo.cpp
Trihedraf: Oh I thought that was the only font we used.

AJenbo: Some systems specify a different font to use less memory.

Source/diablo.cpp
AJenbo: If you make it specify an exact font we can get rid of the build time font options. So I guess that would be preferred.

Trihedraf: It looks like its only the amiga that uses a different font. So would we just do an #ifdef that sets it for the amiga otherwise use charissilb?

AJenbo: Ok, if you think that's better then make it `--font-dir`

Source/diablo.cpp
qndel: + typo in name too

glebm: I would prefer `ttf-dir`. TTF file can still be a separate setting.

If we swap out the font in the future, we don't want the distributions to have change the flag.

CMake/amiga_defs.cmake
```diff
@@ -3,7 +3,6 @@ set(UBSAN OFF)
 set(NONET ON)
 set(USE_SDL1 ON)
 set(SDL1_VIDEO_MODE_BPP 8)
-set(TTF_FONT_NAME \"LiberationSerif-Bold.ttf\")
```
glebm, line 6: These defines are a good way to set the default value for the platform without littering `#ifdef` in the code

glebm (changes requested):

CMake/amiga_defs.cmake
glebm: It's important to have these as CMake defines because it lets package maintainers configure this default without modifying devilutionx code

Trihedraf: AJenbo said "If you make it specify an exact font we can get rid of the build time font options. So I guess that would be preferred." So IDK at this point.

AJenbo: You changed my mind "It looks like its only the amiga that uses a different font. So would we just do an #ifdef that sets it for the amiga otherwise use charissilb?", sorry for the confusion :/

Trihedraf: I just put TTF_FONT_NAME back. It will also let somebody use any font they want by setting the TTF_FONT_NAME when building.

glebm: Can you please put `TTF_FONT_DIR` back as well (as the default for when no `--ttf-dir` argument is provided)?

Some systems use a different font directory. E.g. batocera.linux is a Linux but has its custom paths for everything, etc.

Source/diablo.cpp
```diff
@@ -144,6 +145,8 @@ static void diablo_parse_flags(int argc, char **argv)
 			SetPrefPath(argv[++i]);
 		} else if (strcasecmp("--config-dir", argv[i]) == 0) {
 			SetConfigPath(argv[++i]);
+		} else if (strcasecmp("--ttf-dir", argv[i]) == 0) {
+			SetTtfPath(argv[++i]);
```
glebm, line 149: The flag is called `--ttf-dir` but actually sets the full path to the TTF font.
Shouldn't it set the directory only?

SourceS/paths.cpp
```diff
@@ -60,6 +74,13 @@ const std::string &GetConfigPath()
 	return *configPath;
 }
 
+const std::string &GetTtfPath()
+{
+	if (ttfPath == NULL)
+		ttfPath = new std::string(TTF_FONT_DIR TTF_FONT_NAME);
```
glebm, line 80:
```suggestion
	if (ttfPath == NULL)
		SetTtfPath(TTF_FONT_DIR);
```

glebm (approved):

SourceX/DiabloUI/fonts.cpp
```diff
@@ -1,5 +1,6 @@
 #include "DiabloUI/fonts.h"
 #include "file_util.h"
+#include "../SourceS/paths.cpp"
```
glebm, line 3: Don't include `cpp` files
I guess you're including it for the `TTF_FONT_NAME` define.

Instead, expose `TTF_FONT_NAME` as `GetTtfFontName()` and do the concatenation here.

glebm (changes requested):

SourceX/DiabloUI/fonts.cpp
```diff
@@ -61,11 +62,7 @@ void LoadTtfFont() {
 		was_fonts_init = true;
 	}
 
-	const char* ttf_font_path = TTF_FONT_NAME;
-	if (!FileExists(ttf_font_path))
-	{
-		ttf_font_path = TTF_FONT_DIR TTF_FONT_NAME;
-	}
+	const char* ttf_font_path = GetTtfPath().data();
```
glebm, line 65:
```suggestion
	std::string ttf_font_path = GetTtfPath() + GetTtfName();
```

SourceS/paths.cpp
glebm, line 80:
```suggestion
		ttfPath = new std::string(TTF_FONT_DIR);
```

(see below)

SourceS/paths.cpp
```diff
@@ -82,4 +103,13 @@ void SetConfigPath(const char *path)
 	AddTrailingSlash(configPath);
 }
 
+void SetTtfPath(const char *path)
+{
+	if (ttfPath == NULL)
+		ttfPath = new std::string;
+	*ttfPath = path;
+	AddTrailingSlash(ttfPath);
+	AddFontName(ttfPath);
```
glebm, line 112:
```suggestion
```

(see below)

SourceX/DiabloUI/fonts.cpp
```diff
@@ -61,11 +62,7 @@ void LoadTtfFont() {
...
 	}
 
-	const char* ttf_font_path = TTF_FONT_NAME;
-	if (!FileExists(ttf_font_path))
-	{
-		ttf_font_path = TTF_FONT_DIR TTF_FONT_NAME;
-	}
+	const char* ttf_font_path = GetTtfPath().data();
 #ifdef __linux__
 	if (!FileExists(ttf_font_path))
```
glebm, line 67: The line below should then be changed to:

```c++
ttf_font_path = "/usr/share/fonts/truetype/" + GetTtfFontName();
```

glebm (approved):

SourceS/paths.cpp
```diff
@@ -6,13 +6,23 @@
 #include "sdl2_to_1_2_backports.h"
 #endif
 
+#ifndef TTF_FONT_DIR
+#define TTF_FONT_DIR ""
+#endif
+
+#ifndef TTF_FONT_NAME
+#define TTF_FONT_NAME "CharisSILB.ttf"
+#endif
```
julealgon, line 15: Why are we using defines for new code? Can't we just use consts instead?

SourceS/paths.cpp
```diff
@@ -60,6 +70,20 @@ const std::string &GetConfigPath()
 	return *configPath;
 }
 
+const std::string &GetTtfPath()
+{
+	if (ttfPath == NULL)
```
julealgon, line 75: Shouldn't we be using `nullptr` instead?

SourceS/paths.cpp
```diff
@@ -82,4 +106,19 @@ void SetConfigPath(const char *path)
...
+	*ttfPath = path;
+	AddTrailingSlash(ttfPath);
+}
+
+void SetTtfName(const char *path)
+{
+	if (ttfName == NULL)
+		ttfName = new std::string;
+	*ttfName = path;
+}
```
julealgon, line 122: Why don't we need `AddTrailingSlash` in this one?

SourceS/paths.cpp
glebm: This is so that they can overriden via CMake arguments

SourceS/paths.cpp
glebm: Yes, but we can wait for @AJenbo to configure .clang-tidy the way he likes and then automatically migrate all of these

SourceS/paths.cpp
glebm: No because this is a filename, not a directory path

SourceS/paths.cpp
julealgon: :(

SourceS/paths.cpp
julealgon: Whats the difference? Aren't they just both string that need to be null-terminated like any other string?

Can you elaborate further?

Trihedraf: It's not for null termination. The directory paths need the slash so the filename can be stuck on the end later. The file is the end and needs to not have a slash.

AJenbo: > Whats the difference? Aren't they just both string that need to be null-terminated like any other string?
> 
> Can you elaborate further?

null != slash 🤷

SourceS/paths.cpp
AJenbo: its only been moved, but yeah i wouuld also have prefered it being replaced, but i get why that isnt a good option. in some cases.

SourceS/paths.cpp
julealgon: Oh wow, I completely misread the other method and thought we were adding null termination. My bad, nothing to see here.

PR #1254, opened 2021-03-23
Update Linux AppImage to use devilutionx.mpq and charissilb.ttf

Trihedraf:
Update Linux AppImage to use devilutionx.mpq and charissilb.ttf. Depends on #1253


AJenbo (approved):
I'm starting to think this needs to be split into a couple of helper scripts that could also be run locally when doing a release or a custom build.

AJenbo (approved):

AJenbo:
Nice this is a good improvement in general now :+1: 

I guess this depends on #1253 ?

Trihedraf:
Yes

PR #1255, opened 2021-03-23
Resolve #1228

### LINKED ISSUE #1228: [Minor bug] Hitting ESC+Enter (Save game) causes the mouse pointer to jump to the upper left corner
Before ESC:
before

After save game:
after

Chance4us:
Resolve #1228 #1255 

Workaround.
Unclear whether there is a side effect.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1464,7 +1464,7 @@ void DisableInputWndProc(UINT uMsg, WPARAM wParam, LPARAM lParam)
 	case DVL_WM_SYSKEYDOWN:
 	case DVL_WM_SYSCOMMAND:
 	case DVL_WM_MOUSEMOVE:
-		GetMousePos(lParam);
+		//GetMousePos(lParam);
 		return;
 	case DVL_WM_LBUTTONDOWN:
 		if (sgbMouseDown != CLICK_NONE)


Source/diablo.cpp
```diff
@@ -1464,7 +1464,7 @@ void DisableInputWndProc(UINT uMsg, WPARAM wParam, LPARAM lParam)
 	case DVL_WM_SYSKEYDOWN:
 	case DVL_WM_SYSCOMMAND:
 	case DVL_WM_MOUSEMOVE:
-		GetMousePos(lParam);
+		//GetMousePos(lParam);
```
AJenbo, line 1467: Don't leave disabled code. If we need it back we can get it from git.

AJenbo:
You should give changes a proper title rather than just referring to some external thing it solves.

Issue: "#147: Mouse doesn't work"
Fix: "Disable mouse movement in DisableInputWndProc
Resolves #147"

Source/diablo.cpp
AJenbo, line 1467: Probably a better solution is to split the cases:
```cpp
	case DVL_WM_SYSCOMMAND:
		return;
	case DVL_WM_MOUSEMOVE:
		GetMousePos(lParam);
		return;
```

AJenbo: I have now implemented this, thanks for figuring out where the issue where.
https://github.com/diasurgical/devilutionX/commit/c5b407f7a69b8d8af6a0debb3f187db47bdeaba3

Chance4us: Thank you.

PR #1259, opened 2021-03-23
Self-saving/self-loading options mechanism

julealgon:
This PR introduces new Options infrastruture to better centralize management of all settings in the game, by creating a hierarchy of classes that can save and load from the ini by themselves.

This makes it easier to add/remove/update flags as their definition is centralized in only one place, avoiding code duplication and mistakes.

For this iteration, we only replace boolean flags with the new system. Other types of flags will come later (each setting is mostly independent of each other).

### DIFF
--- a/Source/options.h
+++ b/Source/options.h
@@ -140,7 +140,15 @@ public:
 	/** @brief Run in fullscreen or windowed mode. */
 	BooleanOption bFullscreen = { "Fullscreen", true };
 	/** @brief Scale the image after rendering. */
-	bool bUpscale;
+	BooleanOption bUpscale =
+	{
+		"Upscale",
+#if !defined(USE_SDL1) && !defined(__vita__)
+		true
+#else
+		false
+#endif
+	};
 	/** @brief Expand the aspect ratio to match the screen. */
 	bool bFitToScreen;
 	/** @brief See SDL_HINT_RENDER_SCALE_QUALITY. */
--- a/Source/options.h
+++ b/Source/options.h
@@ -2,7 +2,8 @@
 
 DEVILUTION_BEGIN_NAMESPACE
 
-typedef struct AudioOptions {
+class AudioOptions final {
+public:
 	/** @brief Movie and SFX volume. */
 	Sint32 nSoundVolume;
 	/** @brief Music volume. */
@@ -11,9 +12,10 @@ typedef struct AudioOptions {
 	bool bWalkingSound;
 	/** @brief Automatically equipping items on pickup emits the equipment sound. */
 	bool bAutoEquipSound;
-} AudioOptions;
+};
 
-typedef struct GraphicsOptions {
+class GraphicsOptions final {
+public:
 	/** @brief Render width. */
 	Sint32 nWidth;
 	/** @brief Render height. */
@@ -38,9 +40,10 @@ typedef struct GraphicsOptions {
 	bool bColorCycling;
 	/** @brief Enable FPS Limit. */
 	bool bFPSLimit;
-} GraphicsOptions;
+};
 
-typedef struct GameplayOptions {
+class GameplayOptions final {
+public:
 	/** @brief Game play ticks per secound. */
 	Sint32 nTickRate;
 	/** @brief Enable double walk speed when in town. */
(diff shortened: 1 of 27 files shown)


Source/options.h
```diff
@@ -0,0 +1,225 @@
...
+	void AddOption(OptionBase&);
+	void AddOptionGroup(OptionGroup&);
+
+private:
+	std::vector<OptionBase*> m_options;
+	std::vector<OptionGroup*> m_optionGroups;
+	std::string m_sectionName;
+};
+
+typedef struct AudioOptions : public OptionGroup {
```
glebm, line 114: `class AudioOptions`, not `typedef struct`

julealgon: I'll include the change in the commits that actually introduce the inheritance

Source/options.h
```diff
@@ -0,0 +1,230 @@
...
+	 * @brief Saves the data stored in this option group to the ini file.
+	*/
+	void Save();
+
+protected:
+	void AddOption(OptionBase&);
+	void AddOptionGroup(OptionGroup&);
+
+private:
+	std::vector<OptionBase*> m_options;
```
glebm, line 109: Did you run `clang-tidy` over this? It should've complained that the name does not follow the standard we've chosen.

Source/options.h
```diff
@@ -0,0 +1,230 @@
...
+ * @brief Base non-templated class for all options. Provides saving and loading functionality and a way to group every
+ * type of option together.
+*/
+class OptionBase {
+public:
+	/**
+	 * @brief Loads data from the ini into this instance.
+	 * @param sectionName The name of the section that contains the data for this option.
+	*/
+	virtual void Load(const std::string& sectionName) = 0;
```
glebm, line 21: Perhaps section name should be part of the option itself, instead of being passed into these methods? We don't expect to save the same option in different sections.

glebm (changes requested):

Source/options.h
julealgon: I don't know how to run that. Doesn't it run on the build?

I picked this format from the webkit page. I see no mention of member names in our clang-format file.

https://webkit.org/code-style-guidelines/
> Data members in C++ classes should be private. Static data members should be prefixed by “s_”. Other data members should be prefixed by “m_”.

Source/options.h
```diff
@@ -0,0 +1,230 @@
...
+	*/
+	void Save();
+
+protected:
+	void AddOption(OptionBase&);
+	void AddOptionGroup(OptionGroup&);
+
+private:
+	std::vector<OptionBase*> m_options;
+	std::vector<OptionGroup*> m_optionGroups;
```
glebm, line 110: If this class owns the options, these vectors should be `std::vector<std::unique_ptr<OptionBase>>` (and same for option groups)

Source/options.h
glebm: I've set up clang-tidy to follow the Google Style guide following our conversation in chat.

In the Google Style Guide this would be named `options_`.

Set up clang-tidy + Visual Studio integration and it should be able to automatically update all names to follow the configuration.

This is configured here: https://github.com/diasurgical/devilutionX/blob/7d926a4619a4400d1593056b9041e3bd99d05162/.clang-tidy#L61

glebm (changes requested):

Source/options.h
julealgon: This allows me to pass the section name from the parent. I disagree that the setting itself should know it's section, since the setting is inside a container object which itself is mapped to a section.

Right now, if we move a setting from one section to the other (from one `OptionGroup` class to another) it will automatically pick the section from that parent.

If we allow the setting itself to have its own section name, we open the door for inconsistencies like a setting being inside a parent class, but being part of a section for another class. That would be less than ideal.

Source/options.h
glebm: It does not run as part of CI yet as there would be too many warnings.
We probably shouldn't do a massive cleanup of the whole codebase until after 1.2.0 release but we should start running it on new code.

Source/options.h
julealgon: Can you give me a bit of context? Please understand that I'm mostly not used to C++. Why is the current approach less than ideal?

Source/options.h
glebm: OK

Source/options.h
glebm: As discussed in chat, this is not needed

julealgon:
Converting to draft until we rebase on top of #1260 and apply some template magic to reduce the manual steps when adding options.

AJenbo:
@julealgon does this still need rebasing?

julealgon:
> @julealgon does this still need rebasing?

I stopped working on this as there seemed to be disagreements on how it should be done. Felt too small of an issue to keep pushing on my end.

PR #1264, opened 2021-03-25
Disable streaming music and sound for 3DS / Switch

glebm:
Works around https://github.com/devkitPro/SDL/issues/72 at the cost of ~50 MiB extra RAM usage.

### DIFF
--- a/Source/effects.cpp
+++ b/Source/effects.cpp
@@ -1084,6 +1084,11 @@ void stream_stop()
 static void stream_play(TSFX *pSFX, int lVolume, int lPan)
 {
 	BOOL success;
+#ifndef DISABLE_STREAMING_SOUNDS
+	constexpr bool kAllowStreaming = true;
+#else
+	constexpr bool kAllowStreaming = false;
+#endif
 
 	assert(pSFX);
 	assert(pSFX->bFlags & sfx_STREAM);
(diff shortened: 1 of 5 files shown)


Source/effects.cpp
```diff
@@ -1084,6 +1084,11 @@ void stream_stop()
 static void stream_play(TSFX *pSFX, int lVolume, int lPan)
 {
 	BOOL success;
+#ifndef DISABLE_STREAMING_SOUNDS
```
julealgon, line 1087: Have you considered making this an option on all platforms? Aren't there any scenarios where streaming could be problematic and someone could want to turn it off?

glebm: Streaming should not be problematic elsewhere

glebm: The flag isn't platform specific though. It's an option in the the main cmake file.

julealgon: Yeah I know... was just wondering about the overhead of introducing `#ifdef`s at all... I really dislike those. Feels to me that it would be preferred to have as much of the codebase as possible not relying on those but on soft settings instead.

glebm: For settings I agree but this shouldn't be a setting - it's a workaround against a broken library platform library implementation and we shouldn't carry around this code to other platforms.

AJenbo (approved):

PR #1265, opened 2021-03-25
Fix a small memory leak in game_controller.cpp

glebm:


SourceX/controls/devices/game_controller.cpp
```diff
@@ -166,7 +166,9 @@ void GameController::Add(int joystick_index)
 	controllers_->push_back(result);
 
 	const SDL_JoystickGUID guid = SDL_JoystickGetGUID(sdl_joystick);
-	SDL_Log("Opened game controller with mapping:\n%s", SDL_GameControllerMappingForGUID(guid));
+	char *mapping = SDL_GameControllerMappingForGUID(guid);
+	SDL_Log("Opened game controller with mapping:\n%s", mapping);
+	SDL_free(mapping);
```
julealgon, line 171: Omg... we have to manually free stuff because of a log string. This looks so sad to me for some reason TT

Isnt there a more elegant way to solve this? Some sort of automatic scope guard that calls `SDL_free` when exiting the scope?

julealgon (approved):

SourceX/controls/devices/game_controller.cpp
glebm: Yep, a `unique_ptr` accepts a custom deleter as the second argument!

Defining such `unique_ptr` types is a common practice when using SDL from C++.

julealgon: Nice! I remember a decade or so ago when I created a custom ScopeGuard class to do something like that.

glebm: Sent https://github.com/diasurgical/devilutionX/pull/1267

PR #1266, opened 2021-03-25
Remove SFileDdaSetVolume

glebm:
It is only used in one place and does not actually control the volume of any specific file.

### DIFF
--- a/SourceX/storm/storm.cpp
+++ b/SourceX/storm/storm.cpp
@@ -40,13 +40,6 @@ radon::File &getIni()
 	return ini;
 }
 
-BOOL SFileDdaSetVolume(HANDLE hFile, signed int bigvolume, signed int volume)
-{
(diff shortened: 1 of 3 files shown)


SourceX/storm/storm.cpp
```diff
@@ -40,13 +40,6 @@ radon::File &getIni()
 	return ini;
 }
 
-BOOL SFileDdaSetVolume(HANDLE hFile, signed int bigvolume, signed int volume)
```
julealgon, line 43: Interesting how only one parameter was actually being used...

julealgon (approved):

SourceX/storm/storm.cpp
AJenbo: a combination of us only having to implement what diablo uses and having it all build as one big project.

PR #1267, opened 2021-03-25
Add std::unique_ptr specializations for SDL types

glebm:
Adds `SDLSurfaceUniquePtr` that simplifies handling of SDL surfaces (so we don't have to free them manually).

Migrates a few places to this new type as an example.

### DIFF
--- a/SourceX/DiabloUI/art.cpp
+++ b/SourceX/DiabloUI/art.cpp
@@ -32,38 +32,34 @@ void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette)
 		format = 0;
 		break;
 	}
-	SDL_Surface *art_surface = SDL_CreateRGBSurfaceWithFormat(SDL_SWSURFACE, width, height, bpp, format);
+	SDLSurfaceUniquePtr art_surface { SDL_CreateRGBSurfaceWithFormat(SDL_SWSURFACE, width, height, bpp, format) };
 
 	if (!SBmpLoadImage(pszFile, pPalette, static_cast<BYTE *>(art_surface->pixels),
 	        art_surface->pitch * art_surface->format->BytesPerPixel * height, 0, 0, 0)) {
 		SDL_Log("Failed to load image");
-		SDL_FreeSurface(art_surface);
-		art->surface = NULL;
 		return;
 	}
 
-	art->surface = art_surface;
 	art->logical_width = art_surface->w;
 	art->frame_height = height / frames;
 
-	ScaleSurfaceToOutput(&art->surface);
+	art->surface = ScaleSurfaceToOutput(std::move(art_surface));
 }
 
 void LoadMaskedArt(const char *pszFile, Art *art, int frames, int mask)
 {
 	LoadArt(pszFile, art, frames);
 	if (art->surface != NULL)
-		SDLC_SetColorKey(art->surface, mask);
+		SDLC_SetColorKey(art->surface.get(), mask);
 }
 
 void LoadArt(Art *art, const BYTE *artData, int w, int h, int frames)
 {
 	art->frames = frames;
-	art->surface = SDL_CreateRGBSurfaceWithFormatFrom(
-	    const_cast<BYTE *>(artData), w, h, 8, w, SDL_PIXELFORMAT_INDEX8);
+	art->surface = ScaleSurfaceToOutput(SDLSurfaceUniquePtr{SDL_CreateRGBSurfaceWithFormatFrom(
+	    const_cast<BYTE *>(artData), w, h, 8, w, SDL_PIXELFORMAT_INDEX8)});
 	art->logical_width = w;
 	art->frame_height = h / frames;
-	ScaleSurfaceToOutput(&art->surface);
 }
 
 } // namespace devilution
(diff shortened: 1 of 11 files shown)


SourceX/DiabloUI/art.cpp
```diff
@@ -31,38 +31,34 @@ void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette)
 		format = 0;
 		break;
 	}
-	SDL_Surface *art_surface = SDL_CreateRGBSurfaceWithFormat(SDL_SWSURFACE, width, height, bpp, format);
+	SDLSurfaceUniquePtr art_surface { SDL_CreateRGBSurfaceWithFormat(SDL_SWSURFACE, width, height, bpp, format) };
```
julealgon, line 34: Wouldn't it be possible to do this with `unique_ptr<SDL_Surface>` syntax?

SourceX/DiabloUI/text_draw.cpp
```diff
@@ -41,15 +41,14 @@ void DrawTTF(const char *text, const SDL_Rect &rectIn, int flags,
...
 	if (*render_cache == NULL) {
-		*render_cache = new TtfSurfaceCache();
 		const auto x_align = XAlignmentFromFlags(flags);
-		(*render_cache)->text = RenderUTF8_Solid_Wrapped(font, text, text_color, rect.w, x_align);
-		ScaleSurfaceToOutput(&(*render_cache)->text);
-		(*render_cache)->shadow = RenderUTF8_Solid_Wrapped(font, text, shadow_color, rect.w, x_align);
-		ScaleSurfaceToOutput(&(*render_cache)->shadow);
+		*render_cache = new TtfSurfaceCache {
+			/*.text=*/ScaleSurfaceToOutput(SDLSurfaceUniquePtr { RenderUTF8_Solid_Wrapped(font, text, text_color, rect.w, x_align) }),
+			/*.shadow=*/ScaleSurfaceToOutput(SDLSurfaceUniquePtr { RenderUTF8_Solid_Wrapped(font, text, shadow_color, rect.w, x_align) }),
```
julealgon, line 47: Wait... are these comments some ghetto way of using named arguments? oO

julealgon, line 47: Why do we have the exact same line twice? If it is needed, can we avoid the code duplication somehow?

SourceX/display.cpp
```diff
@@ -296,15 +296,13 @@ SDL_Surface *CreateScaledSurface(SDL_Surface *src)
 } // namespace
 #endif // USE_SDL1
 
-void ScaleSurfaceToOutput(SDL_Surface **surface)
+SDLSurfaceUniquePtr ScaleSurfaceToOutput(SDLSurfaceUniquePtr surface)
```
julealgon, line 299: Old parameter was a double pointer, and now is a normal one. Is this ok?

SourceX/sdl_ptrs.h
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @brief Deletes the object using `SDL_free`.
+ */
+template <typename T>
+struct SDLFreeDeleter {
+	static_assert(!std::is_same<T, SDL_Surface>::value,
+	    "SDL_Surface should use SDLSurfaceUniquePtr instead.");
+
+	void operator()(T *obj) const
```
julealgon, line 33: Is this really a `const` method? Doesn't freeing something cause side effects?

SourceX/DiabloUI/text_draw.cpp
AJenbo: yeah im not a fan either, cant the IDE be configured to show them but not add them to the file.

glebm: The lines are different - they use different colors
Refactoring them is out-of-scope for this PR

SourceX/display.cpp
glebm: Yes

SourceX/sdl_ptrs.h
glebm: Yes, it's a const method because it does not modify any of the struct's fields (it doesn't have any)

SourceX/DiabloUI/art.cpp
glebm: That would be `std::unique_ptr<SDL_Surface, SDLSurfaceDeleter>` which is verbose and error-prone

glebm: Using just `std::unique_ptr<SDL_Surface>` is not possible - specializing / defining things in `std` namespace is not allowed

SourceX/DiabloUI/text_draw.cpp
glebm: In C++20 this is valid syntax to initialize struct fields (without the comment markers): https://abseil.io/tips/172

IDEs can be configured to show these but they will show everything, whereas the programmer can only add these where it makes it clearer.

See https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments

julealgon: It is a shame that this is only available in C++20 indeed.... it can be especially confusing when dealing with raw types since there is zero semantic information on the value itself.

I'm fine with using this (albeit weird) comment convention if we standardize on it. It is better than not having anything.

AJenbo:
Rebase plz

glebm:
Rebased

glebm:
Rebased again

PR #1269, opened 2021-03-25
Fix Gamepad sticks

### LINKED ISSUE #1268: Gamepad sticks no longer work
@yuripourre Gamepad sticks are broken, can you please have a look?

UBSAN also issues the following warnings:

```
../SourceX/controls/controller.cpp:38:6: runtime error: load of value 18, which is not a valid value for type 'bool'
../SourceX/controls/controller.cpp:45:6: runtime error: load of value 251, which is not a valid value for type 'bool'
```

yuripourre:
#Closes #1268

### DIFF
--- a/SourceX/controls/controller.h
+++ b/SourceX/controls/controller.h
@@ -16,11 +16,18 @@ struct ControllerButtonEvent {
 class Controller {
 public:
 	// Raw axis values.
-	float leftStickX = 0, leftStickY = 0, rightStickX = 0, rightStickY = 0;
+	float leftStickX = 0;
+	float leftStickY = 0;
+	float rightStickX = 0;
+	float rightStickY = 0;
 	// Axis values scaled to [-1, 1] range and clamped to a deadzone.
-	float leftStickXUnscaled = 0, leftStickYUnscaled = 0, rightStickXUnscaled = 0, rightStickYUnscaled = 0;
+	float leftStickXUnscaled = 0;
+	float leftStickYUnscaled = 0;
+	float rightStickXUnscaled = 0;
+	float rightStickYUnscaled = 0;
 	// Whether stick positions have been updated and need rescaling.
-	bool leftStickNeedsScaling = false, rightStickNeedsScaling = false;
+	bool leftStickNeedsScaling = false;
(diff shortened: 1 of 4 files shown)


SourceX/controls/controller.h
```diff
@@ -16,11 +16,11 @@ struct ControllerButtonEvent {
 class Controller {
 public:
 	// Raw axis values.
-	float leftStickX, leftStickY, rightStickX, rightStickY;
+	float leftStickX = 0, leftStickY = 0, rightStickX = 0, rightStickY = 0;
```
julealgon, line 19: Can we declare each field in its own line?

julealgon (approved):

SourceX/controls/controller.h
yuripourre: Sure!

SourceX/controls/devices/game_controller.h
```diff
@@ -8,7 +7,7 @@
 #include "controls/controller.h"
 #include "controls/controller_buttons.h"
 
-#ifndef USE_SDL1
+#if !defined(USE_SDL1) || defined(__SWITCH__)
```
yuripourre, line 10: This was added because display.cpp needs GameController definition for Switch:

https://github.com/diasurgical/devilutionX/blob/master/SourceX/display.cpp#L147

AJenbo: Switch is SDL2 so is it actually needed?

glebm: It doesn't make sense for this to be needed.

@yuripourre You probably tried building with both SWITCH and USE_SDL1=ON, which is not supported

glebm:
Have you tested this with a gamepad on your PC?

SourceX/controls/devices/game_controller.h
AJenbo:
```suggestion
 #ifndef USE_SDL1
```

yuripourre:
@glebm I just did and it's not working yet. I will close this PR and open one to revert all the controller changes. That way I will have time to fix it for good.

PR #1278, opened 2021-03-25
Mark noreturn functions as such

AJenbo:


Source/diablo.cpp
```diff
@@ -612,7 +612,7 @@ static void diablo_deinit()
 		SDL_Quit();
 }
 
-void diablo_quit(int exitStatus)
+[[noreturn]] void diablo_quit(int exitStatus)
```
julealgon, line 615: Can you double check if we really need the attribute in the implementation of the functions as well? I believe having them only on the header would be sufficient.

julealgon (dismissed):
Love it.

PR #1281, opened 2021-03-27
:recycle: Move options-related structs to new 'options.h' file

julealgon:
This PR introduces a new header file called `options.h` and moves all of our option structs there.

All consuming code now directly references `options.h` as the extern variable has also been migrated there.


AJenbo:
I have a good feeling about 1.3.0's development cycle already.

julealgon:
> I have a good feeling about 1.3.0's development cycle already.

It's going to be glorious.

PR #1284, opened 2021-03-27
Standardize Options structs

julealgon:
This PR standardizes all options structs by removing unnecessary C `typedef`s

### DIFF
--- a/Source/options.h
+++ b/Source/options.h
@@ -2,19 +2,19 @@
 
 DEVILUTION_BEGIN_NAMESPACE
 
-typedef struct DiabloOptions {
+struct DiabloOptions {
 	/** @brief Play game intro video on startup. */
 	bool bInto;
-} DiabloOptions;
+};
 
-typedef struct HellfireOptions {
+struct HellfireOptions {
 	/** @brief Play game intro video on startup. */
 	bool bInto;
 	/** @brief Corner stone of the world item. */
 	char szItem[sizeof(PkItemStruct) * 2 + 1];
-} HellfireOptions;
+};
 
-typedef struct AudioOptions {
+struct AudioOptions {
 	/** @brief Movie and SFX volume. */
 	Sint32 nSoundVolume;
 	/** @brief Music volume. */
@@ -23,9 +23,9 @@ typedef struct AudioOptions {
 	bool bWalkingSound;
 	/** @brief Automatically equipping items on pickup emits the equipment sound. */
 	bool bAutoEquipSound;
-} AudioOptions;
+};
 
-typedef struct GraphicsOptions {
+struct GraphicsOptions {
 	/** @brief Render width. */
 	Sint32 nWidth;
 	/** @brief Render height. */
@@ -50,9 +50,9 @@ typedef struct GraphicsOptions {
 	bool bColorCycling;
 	/** @brief Enable FPS Limit. */
 	bool bFPSLimit;
-} GraphicsOptions;
+};
 
-typedef struct GameplayOptions {
+struct GameplayOptions {
 	/** @brief Game play ticks per secound. */
 	Sint32 nTickRate;
 	/** @brief Enable double walk speed when in town. */
@@ -91,9 +91,9 @@ typedef struct GameplayOptions {
 	bool bRandomizeQuests;
 	/** @brief Indicates whether or not mosnter type (Animal, Demon, Undead) is shown along with other monster information. */
 	bool bShowMonsterType;
-} GameplayOptions;
+};
 
-typedef struct ControllerOptions {
+struct ControllerOptions {
 	/** @brief SDL Controller mapping, see SDL_GameControllerDB. */
 	char szMapping[1024];
 	/** @brief Use dpad for spell hotkeys without holding "start" */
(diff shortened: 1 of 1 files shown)


Source/options.h
```diff
@@ -2,19 +2,22 @@
 
 DEVILUTION_BEGIN_NAMESPACE
 
-typedef struct DiabloOptions {
+class DiabloOptions final {
```
glebm, line 5: We should use `struct` (without typedef) for types like this.

`struct` in C++ is identical to `class` except the default visibility is `public`.

glebm (changes requested):

Source/options.h
```diff
@@ -2,19 +2,22 @@
...
-typedef struct DiabloOptions {
+class DiabloOptions final {
+public:
 	/** @brief Play game intro video on startup. */
 	bool bInto;
-} DiabloOptions;
+};
 
-typedef struct HellfireOptions {
+class HellfireOptions final {
```
glebm, line 11: It's rather uncommon (and perhaps a bit confusing) to use `final` in places where no inheritance chains are intended.

Inheritance is discouraged overall (for the most part) so the `final` modifier may be sending the wrong signal here.

Source/options.h
julealgon: > for types such as these.

Can you elaborate a bit on this? What caracteristic exactly are you looking for to decide?

glebm: All fields are public

Source/options.h
julealgon: > It's rather uncommon (and perhaps a bit confusing) to use final in places where inheritance is not at play.

I'm a bit confused by this. `final` is there exactly to prevent this class from being inherited. If one removes `final`, inheritance would become "possibly at play".

"Sealed by default" is considered a good practice in C# to avoid inadvertently exposing classes that are not "ready" for inheritance and to limit the design, so I was applying the same principle here.

> Inheritance is discouraged overall (for the most part) so the final modifier may be sending the wrong signal here.

But final not only discourages, but prevents it, which goes inline with what you mentioned is the recommendation (inheritance is discouraged in other languages too in favor of composition). How could this be sending the wrong signal? It sends the message that "you will not inherit from this". Isn't that what we want?

Source/options.h
glebm: Or even: has any public fields

Basically it's when something is more of a plain data container than encapsulated data and/or behaviour

julealgon: > All fields are public

So even if there is inheritance at play, it should be a `struct` if it only contains `public` fields?

glebm: Yes

Source/options.h
glebm: It sends a message that I have to think about inheritance.

It's like writing "Does not contain uranium" on a cereal box :)

Source/options.h
glebm: Of course, as soon as you have anything `virtual`, it's no longer a plain data container and so should be a `class`.

C++ programmers expect that `struct` are generally POD types, so e.g. can be copied with a simple memcpy. Inheritance is at odds with that and should be avoided for structs.

Source/options.h
julealgon: Don't you think it's good to know that the class will never have parent classes and that inheritance will not be possible? That's extremely valuable to me.

I'm really not following your reasoning here. It's the same to me as saying if I mark a method with `const` then "I'll have to think about purity".

Everybody knows inheritance is a thing in C++, people will not "forget about it" just because they don't see the `final` keyword anywhere.

Can you please reconsider this one? It offers so many compile-time benefits, it would be a shame to drop it IMHO.

glebm: > Don't you think it's good to know that the class will never have parent classes and that inheritance will not be possible? That's extremely valuable to me.

Well, this is my default assumption for all classes. It's a stylistic preference, I don't feel that strongly about it.

AJenbo: unless there is an explicit reason to mark some thing as final i would strongly prefer not to add it. I find that it can cause problems when testing (how will you mock a final class), and it makes presumptions about how something will be used in the future.

I'm not sure what you mean by "ready for inheritance", sure thought might not have been put in to that explicitly, but is there known issues with doing so in all circumstances? Do you remove final only when you can prove that all possible usages will work?

julealgon: https://docs.microsoft.com/en-us/archive/blogs/ericlippert/why-are-so-many-of-the-framework-classes-sealed

https://blog.ploeh.dk/2021/03/08/pendulum-swing-sealed-by-default/

https://stackoverflow.com/a/268287/1946412

Mostly all the reasons for sealing a class in C# to me are general reasons. It is considered a good practice there by extremely experienced and known people in the community, including the previous language designer Eric Lippert.

> how will you mock a final class)

You extract an interface for it and mock the interface. For these classes here, there is obviously no need to: you just create instances of them with the values you want.

> and it makes presumptions about how something will be used in the future.

It respects YAH I: you don't know if inheritance will be needed. If it is in the future, you remove the `final`. Note that adding the keyword from a class is a breaking change, while removing it is not.

> Do you remove final only when you can prove that all possible usages will work?

You remove it once you have decided it needs to participate in an hierarchy chain.

glebm: There is a difference between C# and C++ here.

In C#, all methods are virtual.

In C++, for classes without virtual methods, `final` is verbose and unnecessary IMO, because classes without virtual methods are almost never inherited to begin with.

julealgon: >. It's a stylistic preference

@glebm it's not stylistic, it's enforced by the compiler. It's a compile-time guarantee of what one wants.

AJenbo: I would say a lot of this doesn't strictly apply since we are not shipping a library/framework. We are not accountable for support in the same way since this comes with no warranty or guaranty. Also I have a feeling many modders wont know what to do with highly restrictive code. Besides with the planed refactoring i would say our API is at version -1.0.0 and should not be relied on staying the same between two bug fixes.

julealgon: `final` removed.

AJenbo (approved):

AJenbo:
Now if only we knew what causes it to not be built on CircleCI ...

PR #1285, opened 2021-03-27
Add pnum to functions that uses myplr by default

yuripourre:
This PR adds pnum to almost all functions that uses myplr.

The functions that already had pnum as parameter are untouched.

I had to make a few adjustments to keep the same logic in some places, I will describe it better in the future.


Source/automap.cpp
```diff
@@ -672,12 +672,12 @@ void DrawAutomap(CelOutputBuffer out)
...
 
 	for (int pnum = 0; pnum < MAX_PLRS; pnum++) {
-		if (plr[pnum].plrlevel == plr[myplr].plrlevel && plr[pnum].plractive) {
+		if (plr[pnum].plrlevel == plr[pnum].plrlevel && plr[pnum].plractive) {
 			DrawAutomapPlr(out, pnum);
 		}
 	}
 	if (AutoMapShowItems)
-		SearchAutomapItem(out);
+		SearchAutomapItem(out, myplr);
```
yuripourre, line 680: Wrong parameter, please change to `SearchAutomapItem(out, pnum);`

AJenbo:
breaking this up in to commits with related functions seams like a big help for reviewing here, things look very similar but there are lots of pit falls.

control_test needs to be updated also.

Im not sure it makes sens to do this for the current UI, instead you probably want to make a new one for this.

try seeing if it makes more sens to pass aroun a pointer to the player in some of the functions.

yuripourre:
This PR will be replaced soon. Since it is very old, rebasing would be harder than starting from scratch.

AJenbo:
cool :)

PR #1289, opened 2021-03-28
Bugfix: auricGold shouldn\'t determine MaxGold.

peterwankman:
To make my life a bit easier, I privately changed *GOLD_MAX_LIMIT* to 50000.
This yielded the unexpected result that when picking up gold, the stacks were still capped at 5000 gold, and I needed to merge stacks manually.

items.cpp first sets *MaxGold* to *GOLD_MAX_LIMIT* in line 55, but this gets overridden by *auricGold / 2* in line 1076.
This patch fixes this unintended behavior without impacting the original gameplay.


AJenbo (approved):
thanks

PR #1294, opened 2021-03-29
Apple MFi controller buttons with the example of Nimbus+. Hack for a known Nimbus' problem.

singalen:
This is not a complete PR (it's ugly to put a hack code into a core class), but rather a request for comment.
I was curious what it would take to get my controller working.

### DIFF
--- a/CMake/mfi_defs.cmake
+++ b/CMake/mfi_defs.cmake
@@ -1,0 +1,23 @@
+#SDL Joystick axis mapping (circle-pad)
+set(JOY_AXIS_LEFTX 0)
+set(JOY_AXIS_LEFTY 1)
+
+#SDL Joystick button mapping (A / B and X / Y inverted for MFi Nimbus Plus, at least)
+set(JOY_BUTTON_A 1)
+set(JOY_BUTTON_B 0)
+set(JOY_BUTTON_X 3)
+set(JOY_BUTTON_Y 2)
+set(JOY_BUTTON_LEFTSHOULDER 4)
+set(JOY_BUTTON_RIGHTSHOULDER 5)
+set(JOY_BUTTON_BACK 15)
+set(JOY_BUTTON_START 14)
+
+# No use for it yet!
+#set(JOY_BUTTON_MENU 16)
+
+set(JOY_BUTTON_DPAD_DOWN 11)
+set(JOY_BUTTON_DPAD_LEFT 13)
+set(JOY_BUTTON_DPAD_UP 10)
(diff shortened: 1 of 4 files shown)


CMake/mfi_defs.cmake
```diff
@@ -0,0 +1,23 @@
+#SDL Joystick axis mapping (circle-pad)
```
yuripourre, line 1: Imo those changes should be defined during the cmake step:

```
cmake .. -DJOY_AXIS_LEFTX=0
...
```
Maybe adding this to the wiki page would be a good idea. But let's see what others think.

SourceX/controls/devices/joystick.cpp
```diff
@@ -225,7 +225,14 @@ bool Joystick::ProcessAxisMotion(const SDL_Event &event)
 #endif
 #ifdef JOY_AXIS_LEFTY
 	case JOY_AXIS_LEFTY:
-		leftStickYUnscaled = -event.jaxis.value;
+		// Work around Nimbus that reports inverted Y. No known fix,
+		// AFAIK, and no SDL remapping.
+		// https://forums.libsdl.org/viewtopic.php?p=51409
+		if (nimbus_axis_hack_) {
```
yuripourre, line 231: Maybe this could be a more general flag option for the ini. Something like: "Joystick Invert Y-Axis".

It could even use it as a feature :man_shrugging:

SourceX/controls/devices/joystick.cpp
```diff
@@ -263,6 +270,11 @@ void Joystick::Add(int device_index)
 #ifndef USE_SDL1
 	result.instance_id_ = SDL_JoystickInstanceID(result.sdl_joystick_);
 #endif
+	if (strstr(SDL_JoystickNameForIndex(device_index), "Nimbus") != NULL) {
```
yuripourre, line 273: You can remove it if you move it to options.

yuripourre (changes requested):

yuripourre:
@singalen Thank you for your contributions, more devs and more controllers supports are always welcome!

SourceX/controls/devices/joystick.cpp
singalen: I was not sure about the usefulness of such a feature outside of this specific vendor, so I'd like to open a discussion.

yuripourre: Inverted YAxis might happen with other joysticks:
https://github.com/gabomdq/SDL_GameControllerDB/issues/162#issuecomment-355218426

PR #1295, opened 2021-03-29
Add option to disable clicking on character crippling shrines

AJenbo:


qndel:
why not just make them do nothing?

NikoVP:
What about yhe black death zombies? Might as well address those as well.

Eider-McDuck:
QOL changes should leave the game balance completely unchanged.

Making the three shrines do nothing would mean that players could now click on goat shrines and cauldrons since permanent bad effects would be excluded.

But what players usually did in the game was simply never click on any bad shrines, cauldrons or goat shrines. This option just makes this explicit. If a cat jumps on your laptop and clicks on the cauldron, it is as if it never happened.

Moreover, there might be other players in the same game who want to click on those shrines - for whatever reason. Also, games with different QOL options should be compatible (e.g. player with enabled health bar and without can play together, obviously).

With this options, you cannot click on the shrine, but if another player in the same game wants to, they can, if they have not enabled the option. This is as it should be.

---

(Concerning black deaths, it is a bit more complicated to add an option that is strictly QOL, but I have a few ideas. We can discuss this in the next pull request.)



Normally, players will not click on these shrines. Enabling this option makes this explicit, so clicking on these shrines is no longer possible

joewis:
When playing Rogue, black death zombies are kind of exciting. 

I think the problem with the character crippling features is more that there's no way to undo the damage, e.g. through quests, super rare elixirs or special therapy from Pepin. 
Generally, (semi-)permanent consequences could be interesting for gameplay.

NikoVP:
These features are exciting for old school veteran of diablo but not for a casual unsuspecting player. I think an option to disable them with default setting being their normal function is okay. A lot of people don't like permanent damage to character. We are not talking respec here...

Source/objects.cpp
```diff
@@ -4528,10 +4530,12 @@ void OperateObject(int pnum, int i, BOOL TeleFlag)
 		OperateArmorStand(pnum, i, sendmsg);
 		break;
 	case OBJ_GOATSHRINE:
-		OperateGoatShrine(pnum, i, LS_GSHRINE);
+		if(!sgOptions.Gameplay.bDisableCripplingShrines)
```
yuripourre, line 4533: This approach may sound like a bug because there would be no feedback for the users.

I suggest removing the boolean check here and move it to the OperateShine method instead (since both OperateGoatShrine and OperateCauldron call it).

```
void OperateShrine(int pnum, int i, int sType)
{
	...
	if (!deltaload) {
		PlaySfxLoc(sType, object[i]._ox, object[i]._oy);
		object[i]._oAnimFlag = 1;
		object[i]._oAnimDelay = 1;
	} else {
		object[i]._oAnimFrame = object[i]._oAnimLen;
		object[i]._oAnimFlag = 0;
	}

	if (ShrineIsCrippling(i)) && sgOptions.Gameplay.bDisableCripplingShrines) {
		InitDiabloMsg("Cowardice is also a sign of widsom.");
		return;
	}
	...
```

yuripourre (changes requested):

NikoVP:
> 
> 
> When playing Rogue, black death zombies are kind of exciting.
> 
> I think the problem with the character crippling features is more that there's no way to undo the damage, e.g. through quests, super rare elixirs or special therapy from Pepin.
> Generally, (semi-)permanent consequences could be interesting for gameplay.

That's a nice suggestion but we are going again in changing gameplay and this will reaise red flag for a lot of people in the community. Another option would be to add them as temporary debuffs until the next game session/newgame. And make them displayeable. So people actually know what hit them. And will be on the look out next time. Some type of schrine/event historial or an icon showing the active debuff. I think this will stimulate people to learn the specifics of the gameplay and some more cripling efects can be introduced in a mod and make this an enjoyable mechanic, similar to other roguelike games.

AJenbo:
> That's a nice suggestion but we are going again in changing gameplay and this will reaise red flag for a lot of people in the community. Another option would be to add them as temporary debuffs until the next game session/newgame. And make them displayeable. So people actually know what hit them. And will be on the look out next time. Some type of schrine/event historial or an icon showing the active debuff. I think this will stimulate people to learn the specifics of the gameplay and some more cripling efects can be introduced in a mod and make this an enjoyable mechanic, similar to other roguelike games.

That is not trivial to implement and is game changing in it self.

NikoVP:
It sure is game changing. But having a character let's say in in their 10-15 level and finding out that it's crippled is kind of putting you off, especially if you don't know why. And whatever option is introduced for disabling it is also game changing.

Eider-McDuck:
If two players with different settings cannot play together anymore that means we definitely have a non-QOL change. This rules out any changes to game mechanics.

NikoVP:
Maybe a system that synchronizes the game settings to the hosts for the session should be implemented.

Source/objects.cpp
Eider-McDuck: This changes the game mechanics because people would now consider clicking on goat shrines and cauldrons.

Not being able to click on them is not a bug, this is what all reasonable players would do, so they would not notice any difference.

I am open to displaying a message or something, but you need
1) to disable goat shrines and cauldrons as well, and
2) allow players in the same game who have not disabled to the shrines to click on them, even if the other player has already tried to click.

I think a message is not strictly necessary, however.

AJenbo:
> It sure is game changing. But having a character let's say in in their 10-15 level and finding out that it's crippled is kind of putting you off, especially if you don't know why. And whatever option is introduced for disabling it is also game changing.

I'm trusting that you actually understand what I meant?

> Maybe a system that synchronizes the game settings to the hosts for the session should be implemented.

Obviously, there is since you can select difficaulities.

NikoVP:
I am sure its quite difficult to implement. I am no developer but I can grasp the amount of effort put in the project. What I was suggesting is in no way a demand, more like a conversation/discussion.

Source/objects.cpp
yuripourre: @Eider-McDuck yeah, you are right, so maybe only disabling and showing the message to avoid more interaction.

julealgon:
Not a fan of this proposal at all... 

I'd just wait until we softcode shrines and then people will be able to remove whatever shrines they don't like. This is a very hacky "solution" so I'd rather not see it merged as it makes the code worse and the behavior is very unintuitive.

malvarenga123:
> Not a fan of this proposal at all...
> 
> I'd just wait until we softcode shrines and then people will be able to remove whatever shrines they don't like. This is a very hacky "solution" so I'd rather not see it merged as it makes the code worse and the behavior is very unintuitive.

Whenever that happens, I think a single option DISABLE_BAD_SHRINES or something would suffice.

Eider-McDuck:
> I'd just wait until we softcode shrines and then people will be able to remove whatever shrines they don't like. This is a very hacky "solution" so I'd rather not see it merged as it makes the code worse and the behavior is very unintuitive.

Softcoding shrines will not solve this probem. The shrines should not actually be removed from the game. This is for "pro" players who want to play the original game exactly as it was.

For them, it was and is clear that clicking on any of those three shrines or on cauldrons and goat shrines is *always* a mistake. So they would never do it. This just changes the interface accordingly.

(It is like we have auto-heal on Pepin now, since noone would go to him and not click on healing.)

Of course, I am open to have a configurable list of shrines on which you cannot click, but for the standard game it seems not necessary.

julealgon:
@Eider-McDuck 
> Softcoding shrines will not solve this probem. The shrines should not actually be removed from the game. This is for "pro" players who want to play the original game exactly as it was.

Sounds like they should be just playing vanilla then 🤣 

> For them, it was and is clear that clicking on any of those three shrines or on cauldrons and goat shrines is always a mistake. So they would never do it. This just changes the interface accordingly.

What's with this absurd logic? I can't even grasp it. "Doing something in the game is a mistake, so it should be prevented by the game itself". What? It's so incredibly absurd that it's hard to argue against... if you don't want to be able to make mistakes, I think you should play a "click the button to win" sort of game (some mobile games are like that, so there is clearly a market for it).

> Of course, I am open to have a configurable list of shrines on which you cannot click...

This should absolutely _never_  be a thing, because it makes absolutely ZERO sense. A "list of shrines I can't click on", what? Seriously... holy shit.

AJenbo:
> This should absolutely _never_ be a thing, because it makes absolutely ZERO sense. A "list of shrines I can't click on", what? Seriously... holy shit.

Changing the list would affect the level generator, I suppose you could add the same shrine multiple times :shrug:

NikoVP:
> 
> 
> > This should absolutely _never_ be a thing, because it makes absolutely ZERO sense. A "list of shrines I can't click on", what? Seriously... holy shit.
> 
> Changing the list would affect the level generator, I suppose you could add the same shrine multiple times 🤷

Or if the crippling shrines are disabled, have them replaced with the most useless shrine. lets say the one filling your inventory with gold.

julealgon:
> with the most useless shrine. lets say the one filling your inventory with gold.

The gold shrine is extremely powerful IMHO.

NikoVP:
The holy shrine then (teleporting). Will not upset balance or overpower you. Just might get you into trouble.

julealgon:
> The holy shrine(teleporting)? Will not upset balance or overpower you. Just might get you into trouble.

I think there shouldn't be any reason to "handpick" a specific shrine to replace. We just need to make a change that allows the generator to pick from a "pool" of shrines and do that randomly. The only difference is that we'd be removing shrines from the pool of available ones.

Any strategy that rely on specific hacks like "picking shrine X" is a really poor way to go about this IMHO.

AJenbo:
> > The holy shrine(teleporting)? Will not upset balance or overpower you. Just might get you into trouble.
> 
> I think there shouldn't be any reason to "handpick" a specific shrine to replace. We just need to make a change that allows the generator to pick from a "pool" of shrines and do that randomly. The only difference is that we'd be removing shrines from the pool of available ones.
> 
> Any strategy that rely on specific hacks like "picking shrine X" is a really poor way to go about this IMHO.

One difference is that this allows you to play with anyone else that isn't using a modded list, where a modded list would require everyone to use it to play togeather.

julealgon:
> One difference is that this allows you to play with anyone else that isn't using a modded list, where a modded list would require everyone to use it to play togeather.

Absolutely. To properly do it we'd need to implement sending the data from the host, like we discussed briefly earlier. Then whatever the host has configured will be what the game uses for everybody.

When we have dedicated server support, the configuration could be defined on the server (and shared to clients using a similar mechanism).

AJenbo:
> Absolutely. To properly do it we'd need to implement sending the data from the host, like we discussed briefly earlier. Then whatever the host has configured will be what the game uses for everybody.

I thought we concluded not to do so for modded data :D Also, that requires that the other players want to play that mod, or at least that you are the host.

julealgon:
>I thought we concluded not to do so for modded data :D

I don't remember that oO. I remember we saying it would be implemented later due to the complexity and stuff (in the meantime, people would need the same settings to be able to play together and the game could create a hash to represent and compare configurations on joining).

> Also, that requires that the other players want to play that mod, or at least that you are the host.

True. Host would have total control. Ideally, on the "join game" page, we should list the options that the host have set so that the person joining could opt to join or not.

Tremain84:
I would just change the effect to not lose permanently hitpoints (which can never be regained) but to lose vitality (which can be restored via lvl up or vitality elexir). Same goes for the permanent 10% mana reduction. It could be changed to 5-10 magic points loss, which one is also able to get rid of, but with a lot of time and money.

SourceX/qol.cpp
```diff
@@ -50,6 +50,20 @@ void FillRect(CelOutputBuffer out, int x, int y, int width, int height, BYTE col
...
+{
+        switch(object[i]._oVar1) {
+        case SHRINE_FASCINATING:
+        case SHRINE_ORNATE:
+        case SHRINE_SACRED:
+                return true;
+                break;
+        default:
+                return false;
+                break;
```
AJenbo, line 63:
```suggestion
        default:
                return false;
```

austinwagner:
Coincidentally, after accidentally clicking a bad cauldron while in combat, I implemented a similar change. It sounds like there isn't great agreement on where to draw the line on a change like this, but I'll just add what I went with in case it helps the discussion.

I added two configuration options:

- Filtered Shrines - A comma delimited list of shrines considered bad
- Shrine Filter Mode - Configures how the filter list is applied:
  - 0 = Off
  - 1 = Ignore clicks
  - 2 = Don't spawn (and ignore clicks if already spawned)

Mode 1 is what I'm using because for me it really is just a QoL change. I save scum the cauldrons already, so this change just streamlines it. Mode 2 obviously has a larger impact on the game because there will end up being more good shrines since the number of shrines spawned isn't proportionally decreased.

Maybe it makes sense to disable the filter for multiplayer or to make the click filter only work on normal shrines that an experienced player would know not to click on anyway.

Eider-McDuck:
It is clear that we cannot omit the shrine (this would be a mod).

Here is a new clean version of the disabling code. Can in principle be extended to locally disable any object.

Disabled objects are now displayed in red with the suffix "(disabled)".

Eider-McDuck:
Now also handling Black Death zombies. We add an option for an unintrusive warning in the automap. This is a small spoiler (like the health bar), but sufficient since monsters do not spawn near the level entrance.

julealgon:
> Now also handling Black Death zombies

Please do not do that @Eider-McDuck . They are completely different things and should not be done in the same PR under any circumstance (even if both concepts where approved for inclusion).

If you want to propose the black death warning, send that in a separate PR.

Eider-McDuck:
You are right of course, otherwise it will be difficult to suggest improvements in a clean way. Reverted.

yuripourre (dismissed):

Eider-McDuck:
Check for disabled object was at wrong place.

Eider-McDuck:
A little bit cleaner. Now it looks ready.

PR #1298, opened 2021-03-29
Cleanup appfat and stdargs

AJenbo:


Source/appfat.h
```diff
@@ -3,30 +3,20 @@
...
-extern "C" {
-#endif
+namespace dvl {
 
-#define TraceLastError SDL_GetError
 [[noreturn]] void app_fatal(const char *pszFmt, ...);
 void DrawDlg(const char *pszFmt, ...);
 #ifdef _DEBUG
-[[noreturn]] void assert_fail(int nLineNo, const char *pszFile, const char *pszFail);
+[[noreturn]] void assert_fail(Sint32 nLineNo, const char *pszFile, const char *pszFail);
```
AJenbo, line 15: Maybe I should make it Uint8 to discourage files longer than 255 lines :thinking:

Source/appfat.cpp
```diff
@@ -7,10 +7,12 @@
 #include "../3rdParty/Storm/Source/storm.h"
 #include <config.h>
 
-DEVILUTION_BEGIN_NAMESPACE
+namespace dvl {
+
+namespace {
```
julealgon, line 12: I'm confused by this second empty namespace. Is this some sort of feature in C++ that I'm not aware of?

Source/appfat.h
julealgon: Is it a good practice in C++ to use all integral types for domain purposes like that? That's considered a bad practice in C# (the recommendation is to just validate the range with an `ArgumentOutOfRangeException`).

julealgon (dismissed):

Source/appfat.cpp
AJenbo: This is how you mark things that are private to this translation unit. It allows the compiler to perform aggressively to the contained code since it knows it cannot be called from other parts of the program. It's similar to marking functions as static.

Source/appfat.h
AJenbo: If you are referring to me suggesting the use of `Uint8` then that was just a programmer joak.

julealgon: > If you are referring to me suggesting the use of Uint8

I was, it was an honest question. In C# integral types other than `int` and `long` are recommended to be left for interoperability scenarios only, so I was curious if that wasn't the case in C++ or if anything goes.

Source/appfat.cpp
julealgon: Very interesting (if weird... as with a lot of C++ stuff I guess). Thanks for clarifying!

Source/appfat.h
AJenbo: Since I was not serious I'm lacking some context here. Are you suggesting that `int` would be better than Sint32 because it's left to the compiler/platform what size to use?

In general, we are working on replacing int/long with 32bit size integers as that is what they were in the original Windows x86 build, but may not be on other platforms, which could result in overflow and other issues. And chasing down strange bugs (OOB) on obscure platforms doesn't sound like a fun time.

Additionally `char` is `singed` on x86, but `unsigned` on ARM :tada: 

So explicit types are preferred. I suppose in this instance it really isn't important, but for consistencies sake :shrug:

galaxyhaxz: > I know at least one CPU defines int as 16bit

Most likely this is a very old system that couldn't run Diablo anyway. I agree with using fixed type sizes otherwise, but using `int` for file lines is fine, as even on a 16 bit machine that's 65k lines. Maybe you could also use Uint64, for maximum line support xD

julealgon: > Are you suggesting that int would be better than Sint32 because it's left to the compiler/platform what size to use?

Since we already discussed this briefly with glebm on discord I won't elaborate too much on it. To me, as someone coming from a more "stable" ecosystem, seeing all these "int variations" is incredibly confusing and unnecessary. My preference towards `int` was just because it was the standard integral type in the language, so it felt more natural.

However, the C++ ecosystem is such an absolute mess that it seems you really have to decide between 10 variations of the same thing on each and every moment. That's just incredibly sad to me. I just wish the environment was a bit more straightforward. Just a pipe dream I guess.

The very fact that we are even having these discussions is a testament to the unnecessary complexity involved in these minor things from C. I honestly don't understand how people can like this kind of thing :S

galaxyhaxz: You are right, sadly this is a reminiscent of poor design choices made 50 years ago. While I do and will always love C for its portability, there were some things that were never decided upon and it left it up to the compiler to define and that led to portability issues. At one point `int` was 16-bit and `long` was 32, then int became 32 and long stayed 32 on some and moved to 64 for others. Modern languages have addressed many of these shortcomings, but sadly you can't just take a C# or Go program an compile it for everything like you can with C/C++.

My line of thinking is to keep it simple and use a regular `int` whenever possible and only use type-specified types when it is required that it must be of that size.

PR #1299, opened 2021-03-29
Durability icon gradually changing from gold to red

FluffyQuack:
By default, the durability icon is gold at values 5-3 and red at 2-1. This makes it so it gradually goes from gold to red between 5 and 2.

Visual example:
devilutionx 2021-03-29 22-10-15-034
devilutionx 2021-03-29 22-12-06-272

It also makes it so the durability icon is not shown if the item is fully repaired and has a low max of 5 to 3 (I can remove this change if that's not wanted).

This was supposed to be one single commit, but I forgot I'm really bad at using git. This could be squashed into one commit.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1691,11 +1691,13 @@ void ReleaseChrBtns(bool addAllStatPoints)
 	}
 }
 
-static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c)
+#define DURABILITY_RED 2
+#define DURABILITY_GOLD 5
+static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct* pItem, int x, int c)
 {
 	if (pItem->isEmpty())
 		return x;
-	if (pItem->_iDurability > 5)
+	if (pItem->_iDurability > DURABILITY_GOLD)
 		return x;
 	if (c == 0) {
 		switch (pItem->_itype) {
@@ -1719,9 +1721,45 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
 			break;
 		}
 	}
+
+	if (pItem->_iDurability > DURABILITY_RED && pItem->_iDurability == pItem->_iMaxDur) //Avoid showing durability icon if it's equal to a low max (and above "red" durability value)
+		return x;
+
+	int y = -17 + PANEL_Y;
+
+	//Calculate how much of the icon should be gold and red
+	int height = 32;
+	int max = (pItem->_iMaxDur > DURABILITY_GOLD ? DURABILITY_GOLD : pItem->_iMaxDur) - DURABILITY_RED;
+	int amount;
+	if (pItem->_iDurability <= DURABILITY_RED || max <= 0)
+		amount = 0;
+	else {
+		int cur = pItem->_iDurability - DURABILITY_RED;
+		amount = (height * cur) / max;
+	}
+
+	if (options_hwRendering) { //Fluffy: Render via SDL rendering
+		int renderX = x - BORDER_LEFT;
+		int renderY = y - (height - 1) - BORDER_TOP;
+		c -= 1;
+		if (amount)
+			Render_Texture_Crop(renderX, renderY + (height - amount), TEXTURE_DURABILITYWARNING, -1, height - amount, -1, -1, c + 8); //Gold icon
+		amount = height - amount;
+		if (amount)
+			Render_Texture_Crop(renderX, renderY, TEXTURE_DURABILITYWARNING, -1, -1, -1, amount, c); //Red icon
+		//TODO: This is more indirectly related, but we should make it so that when equipment is fully destroyed, we should show some kind of icon then as well (we could add a new variable to the player struct which defines if a slot is empty due to item destruction or not, and if true, show a durability icon for that slot)
+	} else {
+		if (amount)
+			CelDraw_CropY(x, y, pDurIcons, c + 8, height, 0, amount); //Gold icon
+		if (amount != height)
+			CelDraw_CropY(x, y, pDurIcons, c, 32, amount, height); //Red icon
+	}
 	if (pItem->_iDurability > 2)
 		c += 8;
 	CelDrawTo(out, x, -17 + PANEL_Y, pDurIcons, c, 32);
+		Render_Texture_FromBottom(x - BORDER_LEFT, -17 + PANEL_TOP, TEXTURE_DURABILITYWARNING, c - 1);
+	else
+		CelDraw(x, -17 + PANEL_Y, pDurIcons, c, 32);
 	return x - 32 - 8;
 }
 
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1739,10 +1739,16 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct* pItem, int x, int c
 	}
 
 	//Draw icon
-	if (amount)
-		CelDrawTo_CropY(out, x, y, pDurIcons, c + 8, height, 0, amount); //Gold icon
-	if (amount != height)
-		CelDrawTo_CropY(out, x, y, pDurIcons, c, 32, amount, height); //Red icon
+	if (amount) {
+		CelOutputBuffer buf = GlobalBackBuffer();
+		buf = buf.subregionY(y - amount, amount);
+		CelDrawTo(buf, x, amount, pDurIcons, c + 8, 32); //Gold icon
+	}
+	if (amount != height) {
+		CelOutputBuffer buf = GlobalBackBuffer();
+		buf = buf.subregionY(y - height, height - amount);
+		CelDrawTo(buf, x, height, pDurIcons, c, 32); //Red icon
+	}
 
 	return x - 32 - 8;
 }
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1692,13 +1692,14 @@ void ReleaseChrBtns(bool addAllStatPoints)
 	}
 }
 
-#define DURABILITY_RED 2
-#define DURABILITY_GOLD 5
 static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct* pItem, int x, int c)
 {
+	const int durabilityThresholdGold = 5;
+	const int durabilityThresholdRed = 2;
+
 	if (pItem->isEmpty())
 		return x;
-	if (pItem->_iDurability > DURABILITY_GOLD)
+	if (pItem->_iDurability > durabilityThresholdGold)
 		return x;
 	if (c == 0) {
 		switch (pItem->_itype) {
@@ -1728,14 +1729,14 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct* pItem, int x, int c
 	//Calculate how much of the icon should be gold and red
 	int height = 32;
 	int amount;
-	if (pItem->_iDurability <= DURABILITY_RED)
+	if (pItem->_iDurability <= durabilityThresholdRed)
 		amount = 0;
 	else {
 		if (!sgOptions.Graphics.bDurabilityIconGradual)
 			amount = height;
 		else {
-			int cur = pItem->_iDurability - DURABILITY_RED;
-			amount = (height * cur) / (DURABILITY_GOLD - DURABILITY_RED);
+			int cur = pItem->_iDurability - durabilityThresholdRed;
+			amount = (height * cur) / (durabilityThresholdGold - durabilityThresholdRed);
 		}
 	}
 
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1738,28 +1738,12 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct* pItem, int x, int c
 		amount = (height * cur) / max;
 	}
 
-	if (options_hwRendering) { //Fluffy: Render via SDL rendering
-		int renderX = x - BORDER_LEFT;
-		int renderY = y - (height - 1) - BORDER_TOP;
-		c -= 1;
-		if (amount)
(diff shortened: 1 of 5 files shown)


Source/engine.cpp
```diff
@@ -157,7 +167,41 @@ void CelBlitSafeTo(CelOutputBuffer out, int sx, int sy, BYTE *pRLEBytes, int nDa
 	}
 }
 
-void CelClippedDrawSafeTo(CelOutputBuffer out, int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth)
+void CelBlitSafeTo_CropY(CelOutputBuffer out, int sx, int sy, BYTE *pRLEBytes, int nDataSize, int nWidth, int startY, int endY)
```
AJenbo, line 170: Maybe instead of yet another blitter routine, we could limit the target buffer like in minitext.cpp: https://github.com/diasurgical/devilutionX/blob/master/Source/minitext.cpp#L147

FluffyQuack: Done.

julealgon:
> It also makes it so the durability icon is not shown if the item is fully repaired and has a low max of 5 to 3 (I can remove this change if that's not wanted).

This seems like a behavior change to me so I'd suggest not doing it. IMHO this should only be about showing the gradient.

Messing with "when the icon shows" should be a separate configurable thing (that we currently don't support).

yuripourre:
I have no strong opinion regarding this feature but if it could be enabled using an option flag I would vote for having it.

FitzRoyX:
> > It also makes it so the durability icon is not shown if the item is fully repaired and has a low max of 5 to 3 (I can remove this change if that's not wanted).

I see no problem showing it all the time if you've hammered the max dur down. If it's close to breaking, it's close to breaking.

FluffyQuack:
I did some tweaks. I made this feature a toggle in config, and I reverted the behaviour related to maximum durability so it's the same as the original game (that is, it will still appear on screen with a fully repaired item with a low max durability).

I've been toying with a new presentation style for the durability icon. AJenbo seemed positive on it (I believe he said something along the lines of "Holy heck are you trying to make the game into some kind of nightmare!?" which sounds like high praise as Diablo always had very strong horror vibes). Here's the new presentation:
devilutionx 2021-03-30 21-10-14-341

yuripourre:
> I did some tweaks. I made this feature a toggle in config, and I reverted the behaviour related to maximum durability so it's the same as the original game (that is, it will still appear on screen with a fully repaired item with a low max durability).
> 
> I've been toying with a new presentation style for the durability icon. AJenbo seemed positive on it (I believe he said something along the lines of "Holy heck are you trying to make the game into some kind of nightmare!?" which sounds like high praise as Diablo always had very strong horror vibes). Here's the new presentation:
> devilutionx 2021-03-30 21-10-14-341

I like the way the durability has more visibility now. It's easy to miss the regular icon sometimes.

Source/control.cpp
```diff
@@ -1691,11 +1692,13 @@ void ReleaseChrBtns(bool addAllStatPoints)
 	}
 }
 
-static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c)
+#define DURABILITY_RED 2
+#define DURABILITY_GOLD 5
```
julealgon, line 1696: Can we use `consts` for new code instead of `#define`s?

Source/control.cpp
```diff
@@ -1691,11 +1692,13 @@ void ReleaseChrBtns(bool addAllStatPoints)
 	}
 }
 
-static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c)
+#define DURABILITY_RED 2
```
julealgon, line 1695: I feel like `DURABILITY_RED/GOLD` are very poor names for these, as they represent the outcome and not the actual semantic.

Can we come up with better names to represent these values? Perhaps something like `WeakenedItemDurabilityThreshold` and `NearBreakingItemDurabilityThreshold`?

julealgon, line 1695: Why are these defined in `control.cpp`? These are item things and should be moved into a more appropriate header (not CPP), like `items.h`

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
 			break;
 		}
 	}
-	if (pItem->_iDurability > 2)
-		c += 8;
-	CelDrawTo(out, x, -17 + PANEL_Y, pDurIcons, c, 32);
+
+	int y = -17 + PANEL_Y;
+
+	//Calculate how much of the icon should be gold and red
```
julealgon, line 1728: Instead of comments, can we just extract properly named methods with this behavior?

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
 		}
 	}
-	if (pItem->_iDurability > 2)
-		c += 8;
-	CelDrawTo(out, x, -17 + PANEL_Y, pDurIcons, c, 32);
+
+	int y = -17 + PANEL_Y;
+
+	//Calculate how much of the icon should be gold and red
+	int height = 32;
```
julealgon, line 1729: Can we avoid this magic number? Isn't there a constant somewhere we can use? If not, can we define it?

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+	int amount;
```
julealgon, line 1730: Declaring a variable and then assigning it in multiple code branches makes the code usually much harder to navigate.

Instead, please consider extracting a function that returns the value, and assign its result directly to the variable.

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+	if (pItem->_iDurability <= DURABILITY_RED)
```
julealgon, line 1731: Can we add the `{` and `}` on this `if`?

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+		amount = 0;
+	else {
+		if (!sgOptions.Graphics.bDurabilityIconGradual)
```
julealgon, line 1734: `else` and `if` should be combined on the same line.

Source/control.cpp
julealgon, line 1730: Variable name is too cryptic. Amount of what? Can we name this in a bit more specific way?

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+			amount = height;
+		else {
+			int cur = pItem->_iDurability - DURABILITY_RED;
```
julealgon, line 1737: Avoid abbreviations. Please rename the variable.

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+			amount = (height * cur) / (DURABILITY_GOLD - DURABILITY_RED);
+		}
+	}
+
+	//Draw icon
+	if (amount) {
```
julealgon, line 1743: Please avoid using non-booleans directly as boolean expressions. Use the explicit comparison in this case: 
```c++
if (ammount > 0)
```

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+	else {
+		if (!sgOptions.Graphics.bDurabilityIconGradual)
+			amount = height;
+		else {
+			int cur = pItem->_iDurability - DURABILITY_RED;
+			amount = (height * cur) / (DURABILITY_GOLD - DURABILITY_RED);
+		}
+	}
+
+	//Draw icon
```
julealgon, line 1742: Same here. If a comment is really needed, consider just extracting a function.

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+	if (amount) {
+		CelOutputBuffer buf = GlobalBackBuffer();
+		buf = buf.subregionY(y - amount, amount);
+		CelDrawTo(buf, x, amount, pDurIcons, c + 8, 32); //Gold icon
```
julealgon, line 1746: Would be nice to get rid of this `c + 8` in favor of some properly named value. I know it was already like this before, but since we are touching it it could be a good opportunity.

Source/control.cpp
```diff
@@ -1719,9 +1722,35 @@ static int DrawDurIcon4Item(CelOutputBuffer out, ItemStruct *pItem, int x, int c
...
+			amount = height;
+		else {
+			int cur = pItem->_iDurability - DURABILITY_RED;
+			amount = (height * cur) / (DURABILITY_GOLD - DURABILITY_RED);
+		}
+	}
+
+	//Draw icon
+	if (amount) {
+		CelOutputBuffer buf = GlobalBackBuffer();
```
julealgon, line 1744: Avoid abbreviations.

Source/diablo.cpp
```diff
@@ -427,6 +427,7 @@ static void SaveOptions()
 	setIniInt("Graphics", "Gamma Correction", sgOptions.Graphics.nGammaCorrection);
 	setIniInt("Graphics", "Color Cycling", sgOptions.Graphics.bColorCycling);
 	setIniInt("Graphics", "FPS Limiter", sgOptions.Graphics.bFPSLimit);
+	setIniInt("Graphics", "Gradual Durability Warning", sgOptions.Graphics.bDurabilityIconGradual);
```
julealgon, line 430: Not a fan of this setting. My vote is still to leave it out.

Source/engine.cpp
```diff
@@ -42,7 +42,7 @@ void CelDrawTo(CelOutputBuffer out, int sx, int sy, BYTE *pCelBuff, int nCel, in
 	CelBlitSafeTo(out, sx, sy, pRLEBytes, nDataSize, nWidth);
 }
 
-void CelClippedDrawTo(CelOutputBuffer out, int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth)
+void CelClippedDrawTo(CelOutputBuffer out, int sx, int sy, BYTE* pCelBuff, int nCel, int nWidth)
```
julealgon, line 45: Can we undo this to reduce the diff? Unrelated change.

Source/options.h
```diff
@@ -50,6 +50,8 @@ struct GraphicsOptions {
 	bool bColorCycling;
 	/** @brief Enable FPS Limit. */
 	bool bFPSLimit;
+	/** @brief If true, makes durability icon on the HUD gradually turn from gold to red. */
+	bool bDurabilityIconGradual;
```
julealgon, line 54: We named the setting "...Durability Warning" in the config, but here it's "Icon".

Additionally, the flag name is weird to read right now.

Suggested rename: `bGradualDurabilityWarning`

julealgon (changes requested):

Source/control.cpp
AJenbo: Pretty sure the C way is to call it `gcbDurRd`

Source/control.cpp
AJenbo: We allow single line ifs in the style guyde.

Source/control.cpp
AJenbo: This is basically that.

Source/control.cpp
AJenbo: I think it's called `out` almost everywhere else, technically not an abbreviation, but :shrug: This will all go away in a couple of refactors of the drawing code.

AJenbo:
@julealgon I marked some of your comments as resolved as somewhere basically stating the same thing multiple times, or would be made irrelevant if others were addressed and it made things a bit cluttered and make it a bit daunting to go over, especially when it hits the point where GitHub starts not loading all of them.

Let's give it another iteration and see if that addresses most of the issues and then come back with additional feedback if there are more issues left.

Source/control.cpp
FluffyQuack: Done.

Source/diablo.cpp
FluffyQuack: I'm personally indifferent to this. Surely in the long run we'll have an ingame menu for these options, and then I'm more okay with the ability to tweak minor things.

Source/engine.cpp
FluffyQuack: Done. I don't remember doing this change. It must have been done automatically by VS2019.

Source/diablo.cpp
julealgon: > Surely in the long run we'll have an ingame menu for these options

For this one, I think an option in the menu is _also_ not warranted. If I believed this should be a setting on the menu, I'd be ok with it being a setting on the ini.

Source/engine.cpp
AJenbo: see if yo can get it to follow thw .clang-format configuration in the project

FluffyQuack:
I re-wrote parts of this to try to address all of the complaints above, except for these:

- It's a still config var. We'd need someone to make a decisive decision here if it should be a toggle or not.
- Magic number for height is still there. I agree it would be nice if that to be a reference, but we need to set up some kind of system for that. Maybe a big header file defining the size of various CELs. I don't think that's within the scope of this PR.

yuripourre (approved):

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/d27eafaf1575b963a8314c98ab14d17c52df7666

Sorry for the slow iteration, all the noise was making it a bit overwhelming.

PR #1300, opened 2021-03-29
Avoid static non-PODs

BardurArantsson:


BardurArantsson:
Obviously not an *ideal* solution (which would be avoiding the global altogether), but at least this doesn't fall foul of the Static Order Initialization Fiasco :)

AJenbo:
Hmm maybe the crash is dependant on the resolution, what do you run the game as?

BardurArantsson:
Oh, this has nothing to do with the crash. It's just about avoiding non-POD statics.

AJenbo:
Ik, but was curious about that part also.

BardurArantsson:
Oh! I am running at 800x600 as it happens :). Upscaled, of course.

julealgon:
What is a "POD"?

AJenbo:
Plain Old Data

AJenbo:
The crash on missing graphics should now also be handled

PR #1303, opened 2021-03-30
StormLib: Replace switch/nds/amiga hacks with a simpler one

glebm:
The previous hacks didn't work at all for streaming files.

This is because the previous hacks manually patched `BaseMap_Open` to
`malloc` a buffer but didn't actually read anything into the buffer, so
`BaseMap_Read` was just reading out garbage bytes.

This far simpler hack is to simply use `BaseFile` functions on
platforms that don't have mmap.

### DIFF
--- a/3rdParty/StormLib/src/FileStream.cpp
+++ b/3rdParty/StormLib/src/FileStream.cpp
@@ -18,14 +18,7 @@
 #include "StormCommon.h"
 #include "FileStream.h"
 
-#ifdef __vita__
-#include <psp2/rtc.h>
-#define mmap(ptr, size, c, d, e, f) malloc(size)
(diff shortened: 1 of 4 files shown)


AJenbo:
Could you rebase this so that it gets new builds?

glebm:
Rebased

AJenbo:
Latest feedback from Tumayto
> The audio glitch still occurs, but it seems to be triggered less now? I spammed Deckard, then Griswold, got nothing, then I went over to Pepin and after a bit of spamming it made the noise. I reloaded my save, went to spam again, and got it to produce on Griswold as well. But it's not as bad as it was with the last build I tested. :smile:

glebm:
Closing in favor of #1314

PR #1304, opened 2021-03-30
Fix Storm local file seeking (also fixes `SBmpLoadImage`)

glebm:
SBmpLoadImage tried to get the size of the rest of the file incorrectly
by subtracting the current file position from the total size.

This was incorrect because file positions do not always begin at 0.

The only semi-portable way to get a size from file positions is to
subtract two file positions.

### DIFF
--- a/SourceX/storm/storm.cpp
+++ b/SourceX/storm/storm.cpp
@@ -1,4 +1,5 @@
 #include <cstddef>
+#include <cstdint>
 #include <string>
 
 #include "all.h"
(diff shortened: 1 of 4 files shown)


SourceX/storm/storm.cpp
```diff
@@ -228,7 +246,7 @@ BOOL SBmpLoadImage(const char *pszFileName, SDL_Color *pPalette, BYTE *pBuffer,
 	}
 
 	if (pPalette && pcxhdr.BitsPerPixel == 8) {
-		SFileSetFilePointer(hFile, -768, 0, 1);
+		SFileSetFilePointer(hFile, -768, NULL, DVL_FILE_CURRENT);
```
AJenbo, line 249: is DVL_FILE_CURRENT the proper def to use when talking to storm?

glebm: We don't have any other defs for this as I understand

glebm: I thought you defined these specifically to talk to Storm?

AJenbo:
Still not working :)

image

glebm:
Fixed and sent a PR to fix StormLib upstream as well https://github.com/ladislav-zezula/StormLib/pull/204

PR #1306, opened 2021-03-30
DrawInv cleanup

qndel:
basically cleaning all the copy/pasted code for all item slots and putting it into a loop instead

would be cool if someone with a good eye could take a look at 
```cpp
			if (slot == INVLOC_HAND_LEFT) {
				screen_x += frame_width == INV_SLOT_SIZE_PX ? 14 : 0;
				screen_y += InvItemHeight[frame] == (3 * INV_SLOT_SIZE_PX) ? 0 : -14;
				screen_x = frame_width == INV_SLOT_SIZE_PX ? 13 : 1;
				screen_y += InvItemHeight[frame] == 3 * INV_SLOT_SIZE_PX ? 0 : -14;
			}
```
This matches the code we had before but I'm wondering if unifying these 1 pixel differences wouldn't actually look better?
dst_x is coming out of nowhere too


NikoVP:
When playing with the inventory tab I noticed that one  of the hand slots is bigger than the other by roughly one or two pixels. Hence this might be the pixel difference.

AJenbo:
If you solve the conflict i think this is good to merge, nice cleanup

PR #1313, opened 2021-04-01
Handle mixed item origins in multiplayer

AJenbo:
This is the last piece of the puzzle that is 1.2.0


qndel:
wouldn't it be good to have defines for last valid spell/affix at this point? or just check that in a different way than using a certain hardcoded spell/affix :D

AJenbo:
I did apply the `spell_id`, the rest I don't think has enums, but once they do it should be easy to add them here. Or where are you thinking of ones that are explicitly for the limits? Or were you thinking in the style of `IDI_FIRSTQUEST`? I guess that is an option as well.

Source/loadsave.cpp
```diff
@@ -183,6 +183,23 @@ class SaveHelper {
...
+void RemoveInvalidItem(ItemStruct *pItem)
+{
+	bool isInvalid = !IsItemAvailable(pItem->IDidx) || !IsUniqueAvailable(pItem->_iUid);
+
+	if (!gbIsHellfire) {
+		isInvalid = isInvalid || (pItem->_itype == ITYPE_STAFF && GetSpellStaffLevel(pItem->_iSpell) == -1);
+		isInvalid = isInvalid || (pItem->_iMiscId == IMISC_BOOK && GetSpellBookLevel(pItem->_iSpell) == -1);
+		isInvalid = isInvalid || pItem->_iDamAcFlags != 0;
+		isInvalid = isInvalid || pItem->_iPrePower > IPL_AC_CURSE;
+		isInvalid = isInvalid || pItem->_iSufPower > IPL_AC_CURSE;
```
qndel, line 195: I'm guessing IPL_AC_CURSE is the last valid affix? That's where I was thinking about some stuff other than hardcoded one

Source/spells.cpp
```diff
@@ -361,20 +363,26 @@ int GetSpellBookLevel(int s)
 		case SPL_NOVA:
 		case SPL_APOCA:
 			return -1;
+		default:
+			if (s > SPL_BONESPIRIT)
```
qndel, line 367: same here, some max spell index define/whatever? :P

AJenbo: Can add `SPL_LASTDIABLO` instead.

PR #1314, opened 2021-04-01
StormLib: Update to upstream

glebm:
Updates StormLib to https://github.com/ladislav-zezula/StormLib/commit/0a9057a09f083eb4791a55eb3d9538cea9822e37

Removes all DevilutionX changes except the `#ifdef FULL` functionality trimming.

This means we lose the Xbox changes but they'll likely need to be adjusted for the new Xbox SDK anyway.

### DIFF
--- a/3rdParty/Storm/Source/storm.h
+++ b/3rdParty/Storm/Source/storm.h
@@ -47,9 +47,12 @@ typedef struct _SNETEVENT {
 
 // Note to self: Linker error => forgot a return value in cpp
 
-// Storm API definition
-#ifndef STORMAPI
-#define STORMAPI
+// We declare the StormLib methods we use here.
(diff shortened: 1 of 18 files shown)


glebm:
The only remaining thing here is to test this on a platform without mmap (e.g. switch or 3ds)

AJenbo:
@MBeijer this is the PR that updates stormLib. It looks to be playing the intro video, but stops working after that. I seem to remember a similar issue at some point, where the simple workaround was to skip the splash with -n. Still we should try and solve it properly :)

glebm:
The intro video issue appears for me even without this PR

glebm:
Rebased

glebm:
This should be good to merge

PR #1315, opened 2021-04-01
Issue app_fatal() message if we cannot find QoL assets

BardurArantsson:


SourceX/qol.cpp
```diff
@@ -63,6 +63,12 @@ void InitQol()
 		LoadMaskedArt("data\\healthbox.pcx", &qolArt->healthBox, 1, 1);
 		LoadArt("data\\health.pcx", &qolArt->health);
 		LoadMaskedArt("data\\resistance.pcx", &qolArt->resistance, 6, 1);
+
+		if ((qolArt->healthBox.surface == nullptr) ||
+			(qolArt->health.surface == nullptr) ||
+			(qolArt->resistance.surface == nullptr)) {
+			app_fatal("Failed to load UI resources. Is devilutionx.mpq accessible and up to date?");
```
BardurArantsson, line 70: I'm not thrilled about the error message itself, perhaps we could point to the exact location as well? Don't know enough about the path lookup mechanism to know.

Regardless, should hopefully be enough to get people on the right track.

AJenbo: its not simple as it will look in multiple locations and its not the same son all system and the path can easilly be to long for the info box

PR #1318, opened 2021-04-01
:cow: Draw outline for color in index 0 for inventory icons.

FluffyQuack:
Default outline drawing:
devilutionx 2021-04-01 14-36-08-326

Fixed outline drawing:
devilutionx 2021-04-01 14-35-27-918

Colour in index 0 in the palette is typically used for shadows. When drawing outlines, the game skips this colour so it avoids making outlines around shadows, but since the colour is also sometimes used on the object itself as the colour black this makes the outline not properly follow the shape of many CEL frames.

Inventory icons never have any shadows, so for those, we could make sure to always draw outlines for #0 pixels.

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -461,7 +461,7 @@ void CelDrawUnsafeTo(CelOutputBuffer out, int x, int y, BYTE *pCelBuff, int nCel
 	}
 }
 
-void CelBlitOutlineTo(CelOutputBuffer out, BYTE col, int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth)
+void CelBlitOutlineTo(CelOutputBuffer out, BYTE col, int sx, int sy, BYTE *pCelBuff, int nCel, int nWidth, bool skipColorIndexZero)
 {
 	int nDataSize, w;
 	BYTE *src, *dst, *end;
@@ -480,22 +480,24 @@ void CelBlitOutlineTo(CelOutputBuffer out, BYTE col, int sx, int sy, BYTE *pCelB
 				if (dst < out.end() && dst > out.begin()) {
 					if (dst >= out.end() - out.pitch()) {
 						while (width) {
-							if (*src++) {
+							if (!skipColorIndexZero || *src > 0) {
 								dst[-out.pitch()] = col;
 								dst[-1] = col;
 								dst[1] = col;
 							}
+							src++;
 							dst++;
 							width--;
 						}
 					} else {
 						while (width) {
-							if (*src++) {
+							if (!skipColorIndexZero || *src > 0) {
 								dst[-out.pitch()] = col;
 								dst[-1] = col;
 								dst[1] = col;
 								dst[out.pitch()] = col;
 							}
+							src++;
 							dst++;
 							width--;
 						}
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -185,9 +185,9 @@ void DrawInv(CelOutputBuffer out)
 				color = ICOL_RED;
 			}
 			if (frame <= 179) {
-				CelBlitOutlineTo(out, color, RIGHT_PANEL_X + 133, 59, pCursCels, frame, frame_width);
+				CelBlitOutlineTo(out, color, RIGHT_PANEL_X + 133, 59, pCursCels, frame, frame_width, false);
 			} else {
-				CelBlitOutlineTo(out, color, RIGHT_PANEL_X + 133, 59, pCursCels2, frame - 179, frame_width);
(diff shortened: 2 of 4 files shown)


qndel:
what about making `skipColorIndexZero` an optional argument with default false value?

FluffyQuack:
> what about making `skipColorIndexZero` an optional argument with default false value?

I have a counter proposal! I'll make it true by default. That way its default behaviour matches the CL2 outline draw function, and also, once your DrawInv rewrite PR is accepted, then it's 2 calls to CelBlitOutlineTo() which needs to set skipColorIndexZero to false.

qndel:
yeah I meant true, whatever is more common <derp>

qndel:
whatever requires less code 🤪

Source/engine.cpp
```diff
@@ -480,22 +480,24 @@ void CelBlitOutlineTo(CelOutputBuffer out, BYTE col, int sx, int sy, BYTE *pCelB
 				if (dst < out.end() && dst > out.begin()) {
 					if (dst >= out.end() - out.pitch()) {
 						while (width) {
-							if (*src++) {
+							if (!skipColorIndexZero || *src > 0) {
```
yuripourre, line 483: You could change less lines using: `skipColorIndexZero && *src++`

AJenbo: I actually prefer the current form in this PR as it doesn't do multiple types of operation in a single line.

FluffyQuack: Someone correct me if I'm wrong, but wouldn't that lead to invalid behaviour? The src pointer needs to increase by one every pixel, but if the "if" statement is `skipColorIndexZero && *src++` then I think what's after "&&" is never ran whenever skipColorIndexZero is false.

galaxyhaxz: You could make the suggestion work if you flip it i.e. `*src++ && skipColorIndexZero` But I think the way you have it now makes it the most legible, as post-increment in this case could be confusing.

yuripourre: @FluffyQuack you are right, I missed that.

yuripourre (approved):

julealgon:
"BUT MY VANILLA HIGHLIGHTS!!! I'M A PRO DIABLO PLAYER, I NEED THE ORIGINAL HIGHLIGHT BEHAVIOR"

Source/engine.cpp
```diff
@@ -480,22 +480,24 @@ void CelBlitOutlineTo(CelOutputBuffer out, BYTE col, int sx, int sy, BYTE *pCelB
 				if (dst < out.end() && dst > out.begin()) {
 					if (dst >= out.end() - out.pitch()) {
 						while (width) {
-							if (*src++) {
+							if (!skipColorIndexZero || *src > 0) {
 								dst[-out.pitch()] = col;
 								dst[-1] = col;
 								dst[1] = col;
 							}
+							src++;
```
julealgon, line 488: Shouldn't we be incrementing this before the if, like the original code?

Source/inv.cpp
```diff
@@ -185,9 +185,9 @@ void DrawInv(CelOutputBuffer out)
 				color = ICOL_RED;
 			}
 			if (frame <= 179) {
-				CelBlitOutlineTo(out, color, RIGHT_PANEL_X + 133, 59, pCursCels, frame, frame_width);
+				CelBlitOutlineTo(out, color, RIGHT_PANEL_X + 133, 59, pCursCels, frame, frame_width, false);
 			} else {
-				CelBlitOutlineTo(out, color, RIGHT_PANEL_X + 133, 59, pCursCels2, frame - 179, frame_width);
+				CelBlitOutlineTo(out, color, RIGHT_PANEL_X + 133, 59, pCursCels2, frame - 179, frame_width, false);
```
julealgon, line 190: Making it `false` by default would help indeed, considering most cases rely on it.

julealgon (approved):

Source/engine.cpp
AJenbo: It's a post-increment, so it was not before the if but after... this is why I don't like having them inside other operations, everyone gets confused :joy:

julealgon: Oh... indeed. I'm dumb.

Source/inv.cpp
FluffyQuack: https://github.com/diasurgical/devilutionX/pull/1306

Once that PR is accepted, then it's only 2 or 3 calls to the function which needs the argument to be set to false.

AJenbo (approved):

AJenbo:
This one is too nice and clean to not include in 1.2.0 :+1:

PR #1326, opened 2021-04-03
diablo.ini: Rename "Fast Walk" to "Run in Town"

glebm:
This will probably be a popular option in 1.2.0.
Let's give it a self-explanatory name.


AJenbo:
Named it as such since it was what was in the original registry. Hopefully old school Hellfire players won't be confused :)

Issue #1329, opened 2021-04-03
[3DS] Game crashes every time a save occurs


warnwar:
**Important information**
On New 3DS with sys ver. 11.10.0-43U, using build 3965 from CircleCI 
https://app.circleci.com/pipelines/github/diasurgical/devilutionX/3965/workflows/62a48a74-b2ee-4e5a-84e0-d4c5052ce6f2/jobs/24941

**Describe**
Any time the game saves, it crashes. The save does complete, but the 3DS turns off because of the crash.

It happens in the menu when creating a character, and it also happens whenever you save from within the game. I have included the crash dumps.
[crashdump.zip](https://github.com/diasurgical/devilutionX/files/6251862/crashdump.zip)

**Expected behavior**
Game should not crash when saving

AJenbo:
@MrHuu are you able to see what is going on here?

AJenbo:
> It happens in the menu when creating a character, and it also happens whenever you save from within the game. I have included the crash dumps.
> [crashdump.zip](https://github.com/diasurgical/devilutionX/files/6251862/crashdump.zip)

Thia looks like an empty folder?

warnwar:
oops, sorry about uploading the wrong zip for the crashdumps. This one should actually have them:
[crashdump.zip](https://github.com/diasurgical/devilutionX/files/6251955/crashdump.zip)

AJenbo:
Could you try this version https://24485-143324737-gh.circle-artifacts.com/0/devilutionx.cia

what model do you have, i belive some of the early ones are not able to fully run the game.

MrHuu:
Tested both latest .3dsx and .cia circle-ci builds on my new3ds.
Both builds crash after saving.

I'll update my local builds and look into it.

warnwar:
@AJenbo I have a New 3DS XL. I thought the original 3DS and the other lower-powered ones were incompatible.

I tried that build you suggested, and it also crashes in the same way. The game saves properly, but then the system turns off.

AJenbo:
Kinda odd that this would fail, since it free is guarded by `if (p != null)` so I don't think this is a double-free :/

AJenbo:
I tried explicitly calling `flush()`, could you see if this helps?
CIA:
https://24981-143324737-gh.circle-artifacts.com/0/devilutionx.cia
3DSX:
https://24981-143324737-gh.circle-artifacts.com/0/devilutionx.3dsx

warnwar:
[...]

PR #1337, opened 2021-04-03
[3ds] Try not to crash

AJenbo:
test builds for https://github.com/diasurgical/devilutionX/issues/1329


Source/loadsave.cpp
```diff
@@ -1745,6 +1745,8 @@ void SaveHeroItems(PlayerStruct *pPlayer)
 	SaveItems(&file, pPlayer->InvBody, NUM_INVLOC);
 	SaveItems(&file, pPlayer->InvList, NUM_INV_GRID_ELEM);
 	SaveItems(&file, pPlayer->SpdList, MAXBELTITEMS);
+
+    file.flush();
```
Trihedraf, line 1749:
```suggestion
	file.flush();

```

indentation is off

AJenbo: This is just a quick hack to see if it solves the issue :)

AJenbo: It does not :(

PR #1343, opened 2021-04-03
Fix broken KeySupport for Gameshell (and others)

danie1kr:
Hi all,
I noticed that a recent change to controller.cpp broke the Gameshell support as `ToControllerButtonEvent` no longer reacts to key presses in the menu and later. With this commit, it is fixed.
Looking at the history, this change might have been forgotten when df56d6a03bdf1f148beafa8cbc8612eac4f6c34b was done.
Hope this fixes the issue also for other devices which need the `HAS_KBCTRL == 1` define.
Happy coding!

### DIFF
--- a/SourceX/controls/controller.cpp
+++ b/SourceX/controls/controller.cpp
@@ -26,7 +26,7 @@ ControllerButtonEvent ToControllerButtonEvent(const SDL_Event &event)
 	result.button = KbCtrlToControllerButton(event);
 	if (result.button != ControllerButton_NONE)
 		return result;
-#endifs
+#endif
 #ifndef USE_SDL1
 	GameController *const controller = GameController::Get(event);
 	if (controller != NULL) {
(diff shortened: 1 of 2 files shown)


AJenbo:
@danie1kr looks like this breaks compiling :(

SourceX/controls/controller.cpp
```diff
@@ -22,7 +22,11 @@ ControllerButtonEvent ToControllerButtonEvent(const SDL_Event &event)
 	default:
 		break;
 	}
-
+#if HAS_KBCTRL == 1
+	result.button = KbCtrlToControllerButton(event);
+	if (result.button != ControllerButton_NONE)
+		return result;
+#endifs
```
yuripourre, line 29: #endif

yuripourre:
@danie1kr Nice catch! Your code has a minor typo.

SourceX/controls/controller.cpp
AJenbo: image

AJenbo:
```suggestion
#endif
```

glebm (approved):

AJenbo:
@danie1kr thanks for fixing this.

danie1kr:
> @danie1kr thanks for fixing this.

sure, happy to help.
Sorry for the git-typo, it was late...
But it works on my device.

PR #1344, opened 2021-04-03
Applied a few fixes on Radon

Jmgr:
Applied a few fixes:
- applied const-correctness
- removed unused headers
- added a `virtual` destructor on inheritable classes
- added `final` on non-inheritable classes
- fixed includes (was including all the library in each file)
- added `default` if applicable
- replaced `friend` declarations with accessor
- added missing `const` on reference parameter
- replaced index based `for` with range-based ones
- replaced `NULL` with `nullptr`

The biggest issue in my opinion is that there is no way for the library to return an error when reading or writing a file.

### DIFF
--- a/3rdParty/Radon/Radon/include/File.hpp
+++ b/3rdParty/Radon/Radon/include/File.hpp
@@ -2,25 +2,26 @@
 
 #pragma once
 
+#include "Section.hpp"
+
 #include <string>
 #include <vector>
-#include <memory>
 
 namespace radon
 {
-	class Section;
-
-	class File
+	class File final
 	{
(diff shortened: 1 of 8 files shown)


Jmgr:
And now that I think of it, I would suggest setting the sections and keys containers as maps instead of vectors.

AJenbo:
I think it would be better to target https://github.com/szewczukk/Radon this this PR, then we can sync from there.

Jmgr:
I've heard that another INI library (simpleini) will be used anyway, so that makes this PR quite useless. Oh well. I'll close this one then.

PR #1345, opened 2021-04-03
Alt highlight

qndel:
image
hold alt to highlight, press ctrl to invert highlight (highlighted if you aren't pressing alt)


SourceX/qol.cpp
```diff
@@ -216,4 +216,157 @@ void AutoGoldPickup(int pnum)
...
+
+	int nameWidth = GetTextWidth((char *)textOnGround);
+	x += labelCenterOffsets[ItemCAnimTbl[it->_iCurs]];
+	y -= TILE_HEIGHT;
+	if (!zoomflag) {
+		x <<= 1;
+		y <<= 1;
+	}
+	x -= nameWidth >> 1;
+	char clr = COL_WHITE;
```
yuripourre, line 304: Minor: this attribute could be a text_color to avoid the cast later.

qndel: I tried but compiler cried so I just stopped caring lol

yuripourre:
@qndel amazing work! That's one of the most important features for me. Thanks

SourceX/qol.cpp
AJenbo: :sob:

yuripourre: I did that in a branch somewhere but I had to change the "Drawing Item" struct too.

qndel: oh well, worked this time 🗡️

julealgon:
Would it be possible to add some marging to the sides of the boxes? Notice how close to the border the text is:

image

I think it would look a bit better if a marging similar to the one above and below was used in both sides.

SourceX/qol.cpp
```diff
@@ -216,4 +216,156 @@ void AutoGoldPickup(int pnum)
...
+	    , color(color)
+	    , text(text)
+	{
+	}
+};
+
+std::vector<drawingQueue> drawQ;
+int drawMinX;
+int drawMaxX;
+bool altPressed = false;
```
glebm, line 240: It'd be good to separate out the input handling from functionality.
One reason is that for gamepads I think we want this to be a toggle.

Should we rename this to something like `bGroundItemsHighlight`?

qndel: sure, I just didn't think I'd get this to work today at all :D

yuripourre:
> Would it be possible to add some marging to the sides of the boxes? Notice how close to the border the text is:
> 
> image
> 
> I think it would look a bit better if a marging similar to the one above and below was used in both sides.

Yeah, it would be nice. Frontend people call it "padding".

SourceX/qol.cpp
julealgon: Making it just "highlight" might be misleading. Highlighting already exists both when you mouse over as well as when you use the search spell.

I think we need to come up with a different name for this behavior.

What about `ShowItemNameTooltips`?

Trihedraf:
Is there anyway to make this a toggle so you don't have to hold the button down all the time if you always want the highlighting on?

galaxyhaxz:
@Trihedraf would be really simple to code in, i'm sure Qndel will do it sooner or later.
```
if (/* alt key pressed */) {
  showitems = !showitems;
}
```

qndel:
Tchernobog is using control for this, that way you can 🤔

galaxyhaxz:
Maybe make it press f4 while holding alt to make it stick... xD

Tremain84:
ALT is an amazing button for highlightning items on the ground.

qndel:
> ALT is an amazing button for highlightning items on the ground.

yes but we are talking about an on/off toggle for it without having to press alt all the time

Chance4us:
@qndel The text is sometimes split in two parts at the edge when you move from left to right.
spilt
Is that possible to be improved?
Savegame for testing:
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6292703/single_0.hsv.zip)

galaxyhaxz:
How do we know there isn't a potion on both sides, secretly ? ;P

Chance4us:
> How do we know there isn't a potion on both sides, secretly ? ;P

I just wanted to keep you busy so you wouldn't be bored (joke).

I set the highlighting to 3 so that the text is always displayed. That only happens on the edge when you pace back and forth. Feel free to give it a try.

julealgon:
> Is that possible to be improved?

@Chance4us how would you improve on it? "Improve" is a big open in this case since you could potentially handle it in many different ways.

AJenbo:
I think there is a general consensus that thinks on the right side of the screen should not appear on the left?

It's the same issue that @glebm was having with the cursor the other day. One way to handle it here would be to calculate the width but only request rendering of things slightly past the edge of the screen. But it would add a lot of complexity, and really it should just be handled by the render, but that requires a decent amount of work to fix.

julealgon:
> I think there is a general consensus that thinks on the right side of the screen should not appear on the left?

Sure. But my point is that you have many different ways of handling this case when the item is at the edge of the screen. Here is at least 2:
- Rendering it partially: I think this looks terrible, but it is probably the easiest approach?
- Moving the whole label so it always fits on the screen until the item itself is outside of view:

galaxyhaxz:
> Moving the whole label so it always fits on the screen until the item itself is outside of view:

This is how Diablo 2 handled it, and how my implementation does as well. Though, it does get messy pretty quick, and all the overlapping detection code is a PITA.

Chance4us:
>     * Rendering it partially: I think this looks terrible, but it is probably the easiest approach?

>     * Moving the whole label so it always fits on the screen until the item itself is outside of view:

Third:
* Not show the label if it needs to be moved.

qndel:
I failed to reproduce 
image

SourceX/qol.cpp
```diff
@@ -237,4 +237,156 @@ void AutoGoldPickup(int pnum)
 	}
 }
 
+struct drawingQueue {
```
glebm, line 240: Let's put all the alt highlighting code into a separate file?

`drawingQueue` is a very general name making it very unclear what this code is about

glebm (changes requested):

SourceX/qol.cpp
```diff
@@ -237,4 +237,156 @@ void AutoGoldPickup(int pnum)
...
+		clr = COL_GOLD;
+	drawQ.push_back(drawingQueue(id, it->_ix, it->_iy, x, y, nameWidth, 13, clr, textOnGround));
+}
+
+void ShowItemNameLabels(CelOutputBuffer out)
+{
+	isLabelHighlighted = false;
+	const int borderX = 5;
+	for (unsigned int i = 0; i < drawQ.size(); ++i) {
+		std::map<int, bool> backtrace;
```
glebm, line 339: Looks like this is only used as a set. Use `std::unordered_set` instead

SourceX/qol.cpp
```diff
@@ -237,4 +237,156 @@ void AutoGoldPickup(int pnum)
...
+					canShow = false;
+					drawQ[i].x = newpos;
+					backtrace[newpos] = true;
+				}
+			}
+		} while (!canShow);
+	}
+
+	for (unsigned int i = 0; i < drawQ.size(); ++i) {
+		drawingQueue t = drawQ[i];
```
glebm, line 366: Doesn't look like you need to copy it out, or that you use the index;

```suggestion
	for (const drawingQueue &t : drawQ) {
```

SourceX/qol.cpp
```diff
@@ -237,4 +237,156 @@ void AutoGoldPickup(int pnum)
 	}
 }
 
+struct drawingQueue {
+	int itemID, row, col, x, y, width, height;
+	text_color color;
+	std::string text;
+	drawingQueue(int itemID, int row, int col, int x, int y, int width, int height, text_color color, std::string text)
```
glebm, line 244: Doesn't look like you need this constructor. Remove it and construct with aggregate initialization instead: `drawingQueue { itemId, row, col, ... }`

Iniquitatis: And why pass `std::string` by value instead of const reference?

qndel: > Doesn't look like you need this constructor. Remove it and construct with aggregate initialization instead: `drawingQueue { itemId, row, col, ... }`

Thanks :) I didn't know about such a thing

Chance4us:
Current values for highlighting items:
0 = never highlight
1 = highlight when alt pressed
2 = always highlight except when alt pressed
3 = always highlight

I recommend also following:
4 = toggle between "always highlight" and "never highlight" when hit alt key

Trihedraf:
> 3 = always highlight
> 
> I recommend also following:
> 4 = toggle between "always highlight" and "never highlight" when hit alt key

I would say that 3 should be an alt toggle. No need for a 4.

Chance4us:
> > 3 = always highlight
> > I recommend also following:
> > 4 = toggle between "always highlight" and "never highlight" when hit alt key
> 
> I would say that 3 should be an alt toggle. No need for a 4.

If you set "Alt Highlight Items=3" in diablo.ini, then highlight is always on and there is no possibility to switch of by hitting "Alt".

Trihedraf:
> If you set "Alt Highlight Items=3" in diablo.ini, then highlight is always on and there is no possibility to switch off by hitting "Alt".

Yes that’s how it is currently. But is there a point to having a always and always toggle? I think that we should just set it to always toggle.

qndel:
No more ini setting, hold alt to highlight, press ctrl to invert highlight (highlighted if you aren't pressing alt).
So if someone doesn't want highlight at all, just don't press ctrl/alt derp

Chance4us:
> No more ini setting, hold alt to highlight, press ctrl to invert highlight (highlighted if you aren't pressing alt).
> So if someone doesn't want highlight at all, just don't press ctrl/alt derp

Really nice feature. Thanks.

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,170 @@
...
+			}
+		} while (!canShow);
+	}
+
+	for (const itemLabelQueue &t : labelQ) {
+
+		if (t.x < 0 || t.x >= gnScreenWidth || t.y < 0 || t.y >= gnScreenHeight)
+			continue;
+
+		if (MouseX >= t.x && MouseX <= t.x + t.width && MouseY >= t.y - t.height && MouseY <= t.y) {
```
Iniquitatis, line 149: This and line 146 are great candidates for some sort of `IsPointInRect()`.

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,170 @@
...
+	isLabelHighlighted = false;
+	const int borderX = 5;
+	for (unsigned int i = 0; i < labelQ.size(); ++i) {
+		std::unordered_set<int> backtrace;
+
+		bool canShow;
+		do {
+			canShow = true;
+			for (unsigned int j = 0; j < i; ++j) {
+				if (abs(labelQ[j].y - labelQ[i].y) < labelQ[i].height + 2) {
```
Iniquitatis, line 124: Awfully lot of `labelQ[blah]`. Maybe assign them to constant references?

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,170 @@
...
+	if (it->_itype == ITYPE_GOLD)
+		sprintf(textOnGround, "%i gold", it->_ivalue);
+	else
+		sprintf(textOnGround, "%s", it->_iIdentified ? it->_iIName : it->_iName);
+
+	int nameWidth = GetTextWidth((char *)textOnGround);
+	x += labelCenterOffsets[ItemCAnimTbl[it->_iCurs]];
+	y -= TILE_HEIGHT;
+	if (!zoomflag) {
+		x <<= 1;
```
Iniquitatis, line 101: Why not just multiply by two? It would be more clear.

qndel: because it looks pro

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,170 @@
...
+		y <<= 1;
+	}
+	x -= nameWidth >> 1;
+	text_color clr = COL_WHITE;
```
glebm, line 105: Make a function along these lines:

```c++
text_color GetTextColor(const ItemStruct &it) {
	switch (it._iMagical) {
		case ITEM_QUALITY_MAGIC:
			return COL_BLUE;
		case ITEM_QUALITY_UNIQUE:
			return COL_GOLD
		default:
			return COL_WHITE;
	}
}
```

Then here:

```suggestion
	const text_color clr = GetTextColor(*it);
```

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,170 @@
...
+			} else {
+				isLabelHighlighted = true;
+				cursmx = t.row;
+				cursmy = t.col;
+				pcursitem = t.itemID;
+			}
+		}
+		int bgcolor = 0;
+		if (pcursitem == t.itemID)
+			bgcolor = 134;
```
glebm, line 163: Extract `134` into a named thing like we have for `COL_BLUE` etc

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,170 @@
...
+	int itemID, row, col, x, y, width, height;
+	text_color color;
+	std::string text;
+};
+
+std::vector<itemLabelQueue> labelQ;
+int drawMinX;
+int drawMaxX;
+bool altPressed = false;
+int generatingLabelsState = 0; // 0 - waiting for the game to start, 1 = started generating, 2 = finished generating
```
glebm, line 33: Can use named states instead of comments:

```suggestion
enum class State {
  NOT_STARTED = 0,
  STARTED,
  FINISHED,
};

State generatingLabelsState;
```

glebm (changes requested):

Source/qol/itemlabels.cpp
qndel: I think I actually did recently! :D

Source/qol/itemlabels.cpp
glebm: `SDL_IsPointInRect` -- may also want to change `y` to be the top coordinate and have an `SDL_Rect` directly in the struct instead of `x, y, width, height`

Source/qol/itemlabels.h
```diff
@@ -0,0 +1,25 @@
...
+* Adds Item Labels QoL feature
+*/
+#pragma once
+#include "items.h"
+
+namespace devilution {
+
+struct CelOutputBuffer;
+
+extern BYTE ItemAnimLs[];
```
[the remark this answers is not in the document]
qndel: well, that's how it was in xpbar.h, I copy/pasted the `struct CelOutputBuffer;` from it :D

SourceX/qol.cpp
qndel: > It'd be good to separate out the input handling from functionality.
> One reason is that for gamepads I think we may want this to be a toggle.
> 
> Should we rename this to something like `bGroundItemsHighlight`?

Gamepads can call ToggleItemLabelHighlight() to turn highlighting on/off now

Source/engine.cpp
```diff
@@ -145,7 +146,10 @@ void CelBlitSafeTo(CelOutputBuffer out, int sx, int sy, BYTE *pRLEBytes, int nDa
 			if (!(width & 0x80)) {
 				i -= width;
 				if (dst < out.end() && dst >= out.begin()) {
-					memcpy(dst, src, std::min(static_cast<ptrdiff_t>(width), out.end() - dst));
+					// GenerateItemLabelOffsets loops this function for all items, we don't actually want them on screen, just the min and max x position
```
glebm, line 149: We probably want to avoid putting this into a general-purpose rendering function. Let's make a separate function for measuring instead?

glebm (changes requested):
Almost there!

Source/engine.cpp
qndel: well, you measure by rendering :D

Source/qol/common.cpp
```diff
@@ -4,16 +4,64 @@
...
+void FastDrawHorizLine(const CelOutputBuffer &out, int x, int y, int width, Uint8 col)
 {
 	memset(out.at(x, y), col, width);
 }
 
+void FillRect(const CelOutputBuffer &out, int x, int y, int width, int height, Uint8 col)
+{
+	for (int j = 0; j < height; j++) {
+		FastDrawHorizLine(out, x, y + j, width, col);
+	}
```
glebm, line 62: Something like this will use optimized SIMD implementation of rect filling in SDL (if available for the target architecture):

```c++
SDL_Rect rect;
rect.x = out.region.x + x;
rect.y = out.region.y + y;
rect.w = std::max(out.region.w - x, width);
rect.h = std::max(out.region.h - y, height);
SDL_FillRect(out.surface, &rect, col);
```

Source/engine.cpp
glebm: That's fine but the rendering function shouldn't know anything about the particular game features

A new engine.cpp-internal function can be introduced and used in both the rendering and the measuring implementations

AJenbo: This is really my only gripe with this implementation. I think we can probably carve something useful out of `CheckCursMove()`, so waiting for that to happen one way or another.

ChaosMarc:
There are two small bugs with the magic rock. 
1. Before picking it up the first time there is no text displayed for it when ALT is pressed
2. If you throw the Magic Rock on the ground again after picking it up the ALT-text is flickering. I've attached my save standing next to the rock on the pedestal (not picked up yet)

At least the first one seems to be true for all or some quest items that are placed randomly in the dungeons. I think I remember the same behavior with the Black Mushroom in a previous game. I'm currently playing with a circleci precompiled windows x64 build (https://app.circleci.com/pipelines/github/diasurgical/devilutionX/4668/workflows/131724c1-59b6-45ea-a560-4bd726a73d0a/jobs/31956/artifacts)

[single_0.zip](https://github.com/diasurgical/devilutionX/files/6359717/single_0.zip)

qndel:
> There are two small bugs with the magic rock.
> 
> 1. Before picking it up the first time there is no text displayed for it when ALT is pressed
> 2. If you throw the Magic Rock on the ground again after picking it up the ALT-text is flickering. I've attached my save standing next to the rock on the pedestal (not picked up yet)
> 
> At least the first one seems to be true for all or some quest items that are placed randomly in the dungeons. I think I remember the same behavior with the Black Mushroom in a previous game. I'm currently playing with a circleci precompiled windows x64 build (https://app.circleci.com/pipelines/github/diasurgical/devilutionX/4668/workflows/131724c1-59b6-45ea-a560-4bd726a73d0a/jobs/31956/artifacts)
> 
> [single_0.zip](https://github.com/diasurgical/devilutionX/files/6359717/single_0.zip)

Thanks but no need, I literally found this myself yesterday : P that's because I'm only displaying the label if item is on the last animation frame to avoid displaying the label while the item is in the air and magic rock is still animating while on the ground

galaxyhaxz:
> I'm only displaying the label if item is on the last animation frame to avoid displaying the label while the item is in the air and magic rock is still animating while on the ground

Didn't even realize that this would have been a bug. D2 draws it while being in the air, but doesn't let you select it until it's on the ground.

oGkspAz:
ItemBorders

Another nice addition to highlighting would be using the current borders used by the game on mouseover. Also Showing the item name for the item being hovered over as per D2. Imagine the image above with the intended text above items.
Also just so I note it on here as well: Using the 50% black as per NPC conversations instead of the solid black behind item names.

qndel:
I can't understand the first part? : P

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+void GenerateItemLabelOffsets(const CelOutputBuffer &out)
+{
+	if (generatingLabelsState == State::FINISHED)
+		return;
+	generatingLabelsState = State::STARTED;
+	int itemTypes = gbIsHellfire ? ITEMTYPES : 35;
+	for (int i = 0; i < itemTypes; i++) {
+		drawMinX = gnScreenWidth;
```
glebm, line 86:
```suggestion
		drawMinX = out.w();

```

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,171 @@
...
+{
+	if (altPressed == invertHighlightToggle)
+		return;
+	ItemStruct *it = &items[id];
+
+	char textOnGround[64];
+	if (it->_itype == ITYPE_GOLD)
+		sprintf(textOnGround, "%i gold", it->_ivalue);
+	else
+		sprintf(textOnGround, "%s", it->_iIdentified ? it->_iIName : it->_iName);
```
glebm, line 104:
```suggestion
	if (it->_itype == ITYPE_GOLD)
		std::sprintf(textOnGround, "%i gold", it->_ivalue);
	else
		std::strcpy(textOnGround, it->_iIdentified ? it->_iIName : it->_iName);

```

Source/qol/itemlabels.cpp
```diff
@@ -0,0 +1,171 @@
...
+* Adds Item Labels QoL feature
+*/
+
+#include "common.h"
+#include "itemlabels.h"
+#include "inv.h"
+#include "render.h"
+#include "gmenu.h"
+#include "cursor.h"
+
```
glebm, line 13:
```suggestion

#include <cstring>
```

qndel: ?

qndel: ah nvm, github made me think you wanted to replace render/gmenu/cursor includes with cstring lol

Source/engine.cpp
glebm: You can find the width of the Cel sprite by simply going to the first 0x80; I'll add a function for you

glebm: I've sent #1779, it's untested and I'm not sure how correct it is or whether it needs to handle single-frame CELs differently but it should be a good starting point.

glebm: I see that this measures the count of non-transparent pixels.
This is *probably* not exactly what you want, you want the min/max x. Such a function should be easy to write as well.

qndel:
As I've said earlier, I'm not rebasing this :P feel free to do whatever you want

Chance4us:
The following function can no longer be compiled, what should it look like?
CelClippedDrawTo(out, out.pitch() / 2 - 16, 351, itemanims[i], ItemAnimLs[i], 96);

glebm:
@Chance4us 

```diff
-CelClippedDrawTo(out, out.pitch() / 2 - 16, 351, itemanims[i], ItemAnimLs[i], 96);
+CelClippedDrawTo(out, out.pitch() / 2 - 16, 351, *itemanims[i], ItemAnimLs[i]);
```

Chance4us:
> @Chance4us We now store sprite width with the sprite itself, so there is no need to pass it (`96`):
> 
> ```diff
> -CelClippedDrawTo(out, out.pitch() / 2 - 16, 351, itemanims[i], ItemAnimLs[i], 96);
> +CelClippedDrawTo(out, out.pitch() / 2 - 16, 351, *itemanims[i], ItemAnimLs[i]);
> ```

Thanks a lot!

PR #1346, opened 2021-04-03
Fix shift+clicking stats behavior with capped stats

### LINKED ISSUE #1342: Assigning all remaining stat points with shift+click wastes points when the cap is reached
**Important information**
Windows 10 / latest master https://github.com/diasurgical/devilutionX/commit/c0f57f3020566f6305a3c6e934e9c43b90aa1dcd

**Describe**
When there are more remaining stat points to assign than is needed to reach the attribute cap for a given attribute, shift+clicking will spend all remaining points but only add as many points as needed to reach the cap. Extra points are wasted in the process.

**To Reproduce**
1. Have >1 remaining stat points to spent
2. Be near the stat cap for a given attribute so that the difference to reach the cap is less than the remaining stat points
3. Shift+click on the attribute from 2
4. Observe all remaining stat points are spent even though less points were needed to reach the stat cap

**Expected behavior**
The number of points spent should be capped to what is actually needed to reach the cap for the attribute. Any extra points should remain as "points to distribute"

**Screenshots**
ShiftClickStatsWastes

**Additional context**
Realized the problem after Nitekat commented on doing this. A scenario regarding stat caps hasn't been tested before so this was missed completely.

julealgon:
This PR fixes a bug that caused extra remaining stats to vanish when shift+click was used to assign all remaining stat points to a near-capped attribute.

Fixes #1342

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1655,6 +1655,13 @@ void CheckChrBtns()
 	}
 }
 
+int CapStatPointsToAdd(int remainingStatPoints, const PlayerStruct &player, attribute_id attribute)
+{
+	int pointsToReachCap = player.GetMaximumAttributeValue(attribute) - player.GetBaseAttributeValue(attribute);
+
+	return std::min(remainingStatPoints, pointsToReachCap);
+}
+
 void ReleaseChrBtns(bool addAllStatPoints)
 {
 	int i;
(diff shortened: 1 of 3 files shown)


AJenbo:
Couldn't help but also extract some helpers as well could you :)

Once this is merged could you apply them to the relevant places (MaxStats, get_max_* etc)?

julealgon:
> Once this is merged could you apply them to the relevant places (MaxStats, get_max_* etc)?

Absolutely, that's the idea.

AJenbo:
I though that was the case :)

PR #1366, opened 2021-04-04
✨ Make shift+attack consistent with attack at close range

julealgon:
This PR changes the way the game handles SHIFT+Clicking to attack when monsters are targeted. Instead of always generating a "attack tile at X/Y" command, we'll now generate a "attack monster X" when the player targets a specific monster. This improves responsiveness as the game actually keeps track of the monster being targeted, making it a bit easier and more intuitive to hit monsters.

Making it target a monster also helps with this parallel change: 
- https://github.com/diasurgical/devilutionX/pull/962

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2960,7 +2960,9 @@ void CheckNewPath(int pnum, bool pmWillBeCalled)
 	int y = 0;
 
 	int i = -1;
-	if (player.destAction == ACTION_ATTACKMON) {
+
+	bool holdPosition = Players[pnum].destParam2;
+	if (player.destAction == ACTION_ATTACKMON && !holdPosition) {
 		i = player.destParam1;
 		MakePlrPath(pnum, Monsters[i].position.future, false);
 	}
@@ -3062,7 +3064,7 @@ void CheckNewPath(int pnum, bool pmWillBeCalled)
 		case ACTION_ATTACKMON:
 			x = abs(player.position.tile.x - Monsters[i].position.future.x);
 			y = abs(player.position.tile.y - Monsters[i].position.future.y);
-			if (x <= 1 && y <= 1) {
+			if ((x <= 1 && y <= 1) || holdPosition) {
 				d = GetDirection(player.position.future, Monsters[i].position.future);
 				if (Monsters[i].mtalkmsg != TEXT_NONE && Monsters[i].mtalkmsg != TEXT_VILE14) {
 					TalktoMonster(i);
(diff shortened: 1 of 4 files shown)


AJenbo:
>  Would it also be possible to include a hold left mousebutton + shift to keep attacking the same monster?

We plan to implement it, but it's complicated and will not be part of this PR.

Tremain84:
> 
> 
> > Would it also be possible to include a hold left mousebutton + shift to keep attacking the same monster?
> 
> We plan to implement it, but it's complicated and will not be part of this PR.

Ok great news :)
For the classic Diablo I used a program called: "DiabloClicker" which you had to run after you started Diablo. I still have a copy of that version, but it doesn't work with the devilutionX build. 
But maybe they had some good workaround. If anyone needs it let me know.

AJenbo:
Use a generic Auto Clicker instead.

julealgon:
Rebased.

julealgon:
Rebased again.

Source/player.cpp
```diff
@@ -3244,7 +3245,7 @@ void CheckNewPath(int pnum)
 			i = plr[pnum].destParam1;
 			x = abs(plr[pnum]._px - monster[i]._mfutx);
 			y = abs(plr[pnum]._py - monster[i]._mfuty);
-			if (x <= 1 && y <= 1) {
+			if (x <= 1 && y <= 1 || holdPosition) {
```
obligaron, line 3248: Should parentheses be added for clarity?
```suggestion
			if ((x <= 1 && y <= 1) || holdPosition) {

```

Source/player.cpp
```diff
@@ -3386,7 +3387,7 @@ void CheckNewPath(int pnum)
...
```
obligaron, line 3390: Should parentheses be added for clarity?
```suggestion
			if ((x <= 1 && y <= 1) || holdPosition) {

```

Iniquitatis: Definitely. Mixing && and || in such way is generally a bad idea.

Source/player.cpp
julealgon: Definitely. Added.

Source/player.cpp
AJenbo: Causes warnings on Clang :)

qndel:
> > > Would it also be possible to include a hold left mousebutton + shift to keep attacking the same monster?
> > 
> > 
> > We plan to implement it, but it's complicated and will not be part of this PR.
> 
> Ok great news :)
> For the classic Diablo I used a program called: "DiabloClicker" which you had to run after you started Diablo. I still have a copy of that version, but it doesn't work with the devilutionX build.
> But maybe they had some good workaround. If anyone needs it let me know.

I'm pretty sure they didn't have anything special 😆

julealgon:
Closing this as I can't make it work with the new hold-to-continuously-attack feature and there has been pushback against it in another discussion.

AJenbo:
Atleast i now understand what the intent here is :) I'll let you know if i spot the reason it might not have worked with hold click

PR #1371, opened 2021-04-05
Fixed order of evaluation in SBmpLoadImage size allocation.

StephenCWills:
I encountered crashes at startup in a VS2019 release build that didn't occur in the debug build. The compiler optimizations used a different ordering for evaluation of the operands in the subtraction, causing it to seek to `DVL_FILE_END` before attempting to load the image data.


glebm:
Oh wow

glebm (approved):

PR #1373, opened 2021-04-05
DiabloUI: Render to DiabloUiSurface()

glebm:
Use that instead of using `GetOutputSurface()` directly.

For the rg99 port, we'll want to render this to the `pal_surface`
instead of output surface directly.

Might also be useful for 3ds, perhaps to render to the bottom screen.

### DIFF
--- a/SourceX/DiabloUI/art_draw.cpp
+++ b/SourceX/DiabloUI/art_draw.cpp
@@ -1,10 +1,9 @@
 #include "DiabloUI/art_draw.h"
+
+#include "DiabloUI/diabloui.h"
 #include "display.h"
 
 namespace dvl {
-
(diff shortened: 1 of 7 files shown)


AJenbo:
Should also make perpixel lighting experiments a whole lot easier :D

AJenbo (approved):

AJenbo:
Um was going to wait with this till after 1.2.0 release (hence the 1.3.0 milestone). Lets hope this didn't make things jump around on the screen :D

glebm:
Oops, sorry; This by itself is a complete no-op as `DiabloUiSurface()` just returns `GetOutputSurface()`

AJenbo:
yeah concluded so after looking over it a second time :)

PR #1380, opened 2021-04-06
Save 1.3 MiB RAM in-game by freeing menu resources

glebm:
Running with a memory profiler revealed that we were keeping main menu UI resources around while the game was running:

image

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -377,6 +377,10 @@ BOOL StartGame(BOOL bNewGame, BOOL bSinglePlayer)
 			break;
 		}
 
+		// Save 1.3 MiB of RAM by freeing all main menu resources
+		// before starting the game.
+		UiDestroy();
+
 		gbSelectProvider = FALSE;
 
 		if (bNewGame || !gbValidSaveFile) {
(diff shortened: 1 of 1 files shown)


AJenbo:
Was always suspecting this was the case but wasn't sure if this was "by design" in accordance with the original.

Source/diablo.cpp
```diff
@@ -377,6 +377,10 @@ BOOL StartGame(BOOL bNewGame, BOOL bSinglePlayer)
 			break;
 		}
 
+		// Save 1.3 MiB of RAM by freeing all main menu resources
+		// before starting the game.
```
julealgon, line 381: Do you think the comments are really needed here? Isn't the best practice of "releasing as soon as possible" enough justification to do it?

Source/diablo.cpp
```diff
@@ -392,6 +396,10 @@ BOOL StartGame(BOOL bNewGame, BOOL bSinglePlayer)
 		run_game_loop(uMsg);
 		NetClose();
 		pfile_create_player_description(NULL, 0);
+
+		// If the player left the game into the main menu,
+		// initialize main menu resources.
+		if (gbRunGameResult) UiInitialize();
```
julealgon, line 402: Can we break the line?

glebm: I prefer short statement to be single-line.
Our auto-formatter breaks longer ones.

Source/diablo.cpp
glebm: I'd like to keep the comment because the control flow is very subtle here.

Source/diablo.cpp
AJenbo: Doesn't it also rewrite this into two lines?

glebm: Indeed it does, looks like I didn't run the formatter

I thought we had `AllowShortIfStatementsOnASingleLine` enabled :)

PR #1381, opened 2021-04-06
:bug: use exact size of save format struct

AJenbo:


Source/loadsave.cpp
```diff
@@ -1738,7 +1738,10 @@ static void SavePortal(SaveHelper *file, int i)
 void SaveHeroItems(PlayerStruct *pPlayer)
 {
 	size_t items = NUM_INVLOC + NUM_INV_GRID_ELEM + MAXBELTITEMS;
-	SaveHelper file("heroitems", items * sizeof(ItemStruct));
+	int itemSize = 368;
+	if (gbIsHellfire)
+		itemSize + 4;
```
julealgon, line 1743: Not too relevant to the change itself, but can we agree to always leave a blank line after nested statements like this as a convention?

It makes the code more readable and avoids the situation of inadvertently adding a "second" statement to a scopeless `if` thinking it would become part of the branch execution.

Source/loadsave.cpp
```diff
@@ -1738,7 +1738,10 @@ static void SavePortal(SaveHelper *file, int i)
 void SaveHeroItems(PlayerStruct *pPlayer)
 {
 	size_t items = NUM_INVLOC + NUM_INV_GRID_ELEM + MAXBELTITEMS;
-	SaveHelper file("heroitems", items * sizeof(ItemStruct));
+	int itemSize = 368;
```
julealgon, line 1741: The hardcode here is really unfortunate. Is there no way to get to this using proper `sizeof` calls?

Source/loadsave.cpp
AJenbo: Sure can, I personally think it should be mandatory when doing stuff like this.

Source/loadsave.cpp
AJenbo: No, unfortunately not, the size is that of the original struct with x86 padding. But we can define it as a private const for the file so that it at least gets a name.

Source/loadsave.cpp
glebm: All formatting should be configured in `.clang-format` and done automatically, for the most part.

julealgon: Just to make it clear as I'm not sure my request was understood:

I'm proposing we do:
```cpp
if (x)
    doSomething();

doSomethingElse();
```

Instead of 
```cpp
if (x)
    doSomething();
doSomethingElse();
```

AJenbo:
Hellfire extended the struct with an additional field. Don't worry I will define two separate numbers.

Source/loadsave.cpp
julealgon: Would it be possible to create a struct with fixed-size field types that replicated the original layout exactly, allowing us to use `sizeof` again?

I'd actually prefer that if it was possible and then mapping from one to the other.

Source/loadsave.cpp
AJenbo: I was by me. Can you figure out how to configure it in .clang-fomrat?

glebm: > inadvertently adding a "second" statement

I'm not sure this is possible. The formatter will out-dent the block. If you leave it indented, the compiler or `clang-tidy` will tell you too.

To be clear, I'm against any conventions that are not taken care of automatically.
I work with *many* codebases and can't remember all the conventions. That's what tooling is for.

If clang-format provides such an option, that's fine by me https://clang.llvm.org/docs/ClangFormatStyleOptions.html

Source/loadsave.cpp
AJenbo: Technically I suppose, but it would rely on `pack 1` and a bunch of alignment files that go unused. But definatly not happening now for this.

AJenbo:
For 4 lines that fixes a memory corruption right before a release; you guys sure are nitpicking this to death :joy:

AJenbo:
Can I point your attention to `FILEBUFF` which is what is used elsewhere when the size of the save is needed :P

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/28bb00360162210ff10ddcf5016ae40905063cd7

Source/loadsave.cpp
julealgon: > To be clear, I'm against any conventions that are not taken care of automatically.
I work with many codebases and can't remember all the conventions. That's what tooling is for.

I'm with you. Not sure if my suggestion sounded like I wanted to do it manually only?

> If clang-format provides such an option, that's fine by me https://clang.llvm.org/docs/ClangFormatStyleOptions.html

I'll take a look at that, thanks for the link.

glebm: > Not sure if my suggestion sounded like I wanted to do it manually only?

Yeah I think I thought this because you used the word convention

> I'll take a look at that, thanks for the link.

If there isn't an option and you're willing, clang-format supports plugins

AJenbo: agreed, automated or free for all

PR #1382, opened 2021-04-06
Cleanup DiabloAllocPtr

glebm:
1. Removes a lock around allocation. `malloc` is required to be thread-safe in C11.

2. Defines it as a macro so that:

   1. We provide the correct location for the OOM error.
   2. We get better attribution from memory profilers.


AJenbo:
1... 2... 1.3.0 go!

PR #1386, opened 2021-04-06
Cleanup namespace and headers

AJenbo:


Source/diablo.cpp
```diff
@@ -2094,4 +2094,4 @@ void diablo_color_cyc_logic()
 	}
 }
 
-DEVILUTION_END_NAMESPACE
+}
```
glebm, line 2097: it'd be good to have a terminating comment for namespaces, `} // namespace devilution`

clang-tidy will do this automatically https://clang.llvm.org/extra/clang-tidy/checks/llvm-namespace-comment.html

AJenbo: Sure

julealgon: > it'd be good to have a terminating comment for namespaces

Can you elaborate why @glebm ? This seems really unnecessary to me, especially because it _only_ applies to namespaces for some weird reason. I don't appreciate inconsistent rules like that at all :S

glebm: Well, I'm not really sure why, I'm just cargo-culting what I'm used to here.
No strong opinion.

It is applied automatically though, so once @AJenbo adjusts .clang-tidy rules it won't be something you'll have to thing about

PR #1387, opened 2021-04-06
GPERF: Dump heap profile of the first game iteration

glebm:
If `GPERF` and `GPERF_HEAP_FIRST_GAME_ITERATION` are set,
start tracking allocation in `main` and save the heap state of the first game iteration.

Here is a part of the current heap state of the first game iteration for SDL1 and no audio that this produces:

profile-sdl1-no-audio

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -442,3 +442,6 @@ exefs/main
 ddraw_settings.ini
 Brewfile.lock.json
 .vscode/settings.json
+
+# GPerf heap profile dumps
+*.heap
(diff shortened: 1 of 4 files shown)


AJenbo (approved):
I can see this having some value going forward for tracking the development and also for anyone who wants to port to a limited system. Feel free to merge.

glebm:
Added a separate option for this, `GPERF_HEAP_FIRST_GAME_ITERATION`

PR #1388, opened 2021-04-06
Save 1.5 MiB RAM in-game by freeing more DiabloUI art

glebm:
Heap state of an SDL1 build with no audio

Before:

profile-sdl1-no-audio

After:

profile-sdl1-no-audio-fixed

### DIFF
--- a/SourceX/DiabloUI/diabloui.cpp
+++ b/SourceX/DiabloUI/diabloui.cpp
@@ -1,23 +1,21 @@
-#include "all.h"
-#include "display.h"
-#include "stubs.h"
-#include "utf8.h"
-#include <string>
+#include "DiabloUI/diabloui.h"
+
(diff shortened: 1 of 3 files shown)


SourceX/DiabloUI/diabloui.cpp
```diff
@@ -489,6 +482,18 @@ void LoadUiGFX()
 	LoadHeros();
 }
 
+void UnloadUiGFX()
+{
+	ArtHero.Unload();
```
AJenbo, line 487: Is there any good reason ArtHero exists outside of selhero.cpp?

glebm: Probably not, but that's for a separate cleanup

AJenbo (approved):

PR #1368, opened 2021-04-04
Add a keymapper

Jmgr:
This adds a keymapper that allows players to change the default key used to trigger various actions. The config file now has a new section called `Controller.Keymapping` that will be automatically filled with all defined actions and their default value.

Limitations:
 * no support for having multiple keys bound to the same action (not sure if that's needed, but the original code does accept two different keys for the belt actions for instance, like '&' or '1'. '&' is on the same key as '1' on the French keyboard layout at least, and pressing this key without shift does produce '&' in most applications and not '1'. The SDL seem to have solved this though, as pressing this key on my keyboard produces '1' in the DevilutionX code. All of that to say that those alternative keys might not be useful after all.)
 * not everything has been converted to actions, notably pausing the game (because it has a few other conditions and possibly side-effects), escape, enter and moving up and down text and menus (but I don't imagine we want to accept rebinding those key anyway).

Happy to receive any comment, this can certainly be improved and I've probably missed some cases here and there. :)

Fixes #1050

PR #1389, opened 2021-04-06
Add a Quick Save and a Quick Load actions to the Keymapper

Jmgr:
This just adds two quick save and quick load actions that are bound to F2 and F3 by default. F3 was used previously to show debug info, so it has been moved to 'N'. Quest debugging has also been moved to 'E' for consistency. This also frees F4 for some future action.

Note that this is based on #1368, so the changes showed here are those from that PR except the ones from last commit. This will be fixed once #1368 is merged.


julealgon (changes requested):
> Quest debugging has also been moved to 'E' for consistency.

This will conflict with this: 
- https://github.com/diasurgical/devilutionX/pull/952

AJenbo:
im wondering if we should get rid of most of the debug keys, i find them of little use if any and the complicate changes like key mapping.

Jmgr:
> im wondering if we should get rid of most of the debug keys, i find them of little use if any and the complicate changes like key mapping.

My suggestion here would be to add actions that can be unbound by default to the Keymapper. That way developers could decide what debug feature they want to bind to a key. I would also add non-debug actions like `QuitGame` that would not be bound by default. In the config and unbound key would just not have a value, like `QuitGame=`. WDYT?

Jmgr:
> > Quest debugging has also been moved to 'E' for consistency.
> 
> This will conflict with this:
> 
>     * #952

Oops. I'll remove that binding.

AJenbo:
Sounds good, now you just need to get i building and rebase :)

Jmgr:
> Sounds good, now you just need to get i building and rebase :)

I think I'll wait until #1368 is merged. Maintaining PRs based on other non-merged PRs from a fork a bit painful as you can's set the target branch to be the one from the original PR.

julealgon:
> I think I'll wait until #1368 is merged

Consider marking this PR as "draft" until then @Jmgr .

AJenbo:
Oh, wasn't aware that they where linked. Don't worry about it until the other on is merged, but yeah mark this as a draft.

Jmgr:
> > im wondering if we should get rid of most of the debug keys, i find them of little use if any and the complicate changes like key mapping.
> 
> My suggestion here would be to add to the Keymapper actions that can be unbound by default. That way developers could decide what debug feature they want to bind to a key. I would also add non-debug actions like `QuitGame` that would not be bound by default. In the config and unbound key would just not have a value, like `QuitGame=`. WDYT?

Done.

Jmgr:
I have already added this PR's content in https://github.com/diasurgical/devilutionX/pull/1368, so closing this one.

AJenbo:
Works for me

PR #1393, opened 2021-04-06
Simplify InitPlrGfx

glebm:
It was tracking each type of graphics individually but they're always allocated together.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -11,21 +11,10 @@
 
 namespace devilution {
 
-int plr_lframe_size;
-int plr_wframe_size;
-BYTE plr_gfx_flag = 0;
-int plr_aframe_size;
 int myplr;
 PlayerStruct plr[MAX_PLRS];
-int plr_fframe_size;
-int plr_qframe_size;
 BOOL deathflag;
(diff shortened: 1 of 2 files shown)


AJenbo:
is there even any significant benefit to pre allocating this instead of doing it when loading individual sprites?

glebm:
I still haven't figured out how all of this works
It *might* be related to a few of these being different between town and dungeon

glebm:
I also don't understand how why it worked previously with multiple players

The `plr_gfx_flag` was shared across all players and never reset!

glebm:
OK, so previously it always allocated the size of the BLOCK buffer to all animations of all the players except the first one!

glebm:
Ah no, nevermind, it used the sizes it got when evaluating the first player

glebm:
So we'll hopefully save some RAM once we begin allocating exact buffer sizes (for which I added a FIXME in this PR)

AJenbo:
i guess this PR made Multiplayer start slightly slower :)

glebm:
@AJenbo This PR doesn't change any behaviour, only the follow-up #1396 does

PR #1395, opened 2021-04-06
Removal of redundant updating of gameplay variables

FluffyQuack:
When you click on singleplayer in the main menu, this function gets called: `mainmenu_single_player()`

That loads several gameplay variables from the config.

However, as you progress through the character selection menus and before going ingame, this function gets called: `NetInit()`

That loads the same gameplay variables from the config, which makes the actions in `mainmenu_single_player()` unnecessary. I should note that the `NetInit()` function gets called in both singleplayer and multiplayer modes (when I first saw its name, I was thinking it was run only in network mode).

In order to remove redundant code (and to make it less confusing as to when important variables are actually initialized), I think it would be a good change to remove these lines from `mainmenu_single_player()`

### DIFF
--- a/Source/mainmenu.cpp
+++ b/Source/mainmenu.cpp
@@ -53,12 +53,6 @@ static BOOL mainmenu_init_menu(int type)
 static BOOL mainmenu_single_player()
 {
 	gbIsMultiplayer = false;
-
-	gbRunInTown = sgOptions.Gameplay.bRunInTown;
-	gnTickRate = sgOptions.Gameplay.nTickRate;
-	gbTheoQuest = sgOptions.Gameplay.bTheoQuest;
-	gbCowQuest = sgOptions.Gameplay.bCowQuest;
-
 	return mainmenu_init_menu(SELHERO_NEW_DUNGEON);
 }
 


julealgon (changes requested):
Can't we just get rid of these variables altogether? All of them are used in very few places, so I actually see no reason for them to exist in the first place: just read them from the options directly where it's needed.

AJenbo:
they are there so that in multiplayer you will agree on a game mode without changing your config

julealgon:
> they are there so that in multiplayer you will agree on a game mode without changing your config

Makes sense. Ultimately, I think we need a settings override mechanism so that settings can be set based on a hierarchy, so that multiplayer broadcast settings override local settings seamlessly.

I really dislike the current hardcoded variables there but I understand this will take a while to create. I have no idea if there is a C++ settings library that would allow such a thing, something similar to [Microsoft.Extensions.Options](https://www.nuget.org/packages/Microsoft.Extensions.Options/)

AJenbo:
We could replace the globals with `sgGameInitInfo` with out issue.

AJenbo:
@julealgon see if you like this https://github.com/diasurgical/devilutionX/pull/1450/files

PR #1396, opened 2021-04-07
InitPlrGfx: Only alloc enough for the class

glebm:
Previously, the allocation size was the maximum size over all the classes.

RAM usage per player:

* Before: 2.2 MiB (any class)
* After:
  * Warrior: 2.0 MiB
  * Rogue: 1.5 MiB
  * Sorcerer: 1.6 MiB
  * Monk: 1.7 MiB

We now also only allocate memory in multi-player as needed.

This means the game will only use as much RAM for player graphics as there are players currently in the game. Also the RAM used for each player is lower because it now takes class into account.

Follow-up to #1393

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1778,11 +1778,6 @@ void LoadGameLevel(BOOL firstflag, int lvldir)
 		InitInv();
 		InitItemGFX();
 		InitQuestText();
-
-		int players = gbIsMultiplayer ? MAX_PLRS : 1;
-		for (i = 0; i < players; i++)
-			InitPlrGFXMem(i);
-
 		InitStores();
 		InitAutomapOnce();
 		InitHelp();
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -2208,6 +2208,7 @@ static DWORD On_PLAYER_JOINLEVEL(TCmd *pCmd, int pnum)
 	else {
 		plr[pnum]._pLvlChanging = FALSE;
 		if (plr[pnum]._pName[0] != 0 && !plr[pnum].plractive) {
+			InitPlrGFXMem(pnum);
 			plr[pnum].plractive = TRUE;
 			gbActivePlayers++;
 			EventPlrMsg("Player '%s' (level %d) just joined the game", plr[pnum]._pName, plr[pnum]._pLevel);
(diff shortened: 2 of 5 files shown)


julealgon:
LOL @glebm the memory master... so many memory-related changes lately, this is insane 😆 

Awesome stuff.

glebm:
Just trying to get it to fit into a tiny device, at least without audio

I guess we also need to handle multiplayer here, looks like a player joining may not call these functions

glebm:
It doesn't crash on a player re-joining as a larger class, weird

glebm:
All fixed. We now initialize and free the player graphics memory as needed in multiplayer.

AJenbo:
I guess this still allocates spaces for all variants of that class? Meaning if you play as a light armor with an axe, it will still allocate space for the heavy armor with a staff?

glebm:
Yeah

AJenbo (approved):

glebm:
Last force push only adjusted the commit message

glebm:
> I guess this still allocates spaces for all variants of that class?

It doesn't allocate space for all variants, it allocates enough space to fit the largest variant (from what I can tell)

AJenbo:
> > I guess this still allocates spaces for all variants of that class?
> 
> It doesn't allocate space for all variants, it allocates enough space to fit the largest variant (from what I can tell)

yeah thats what i meant

glebm:
@AJenbo Probably not, though I'm guessing they're of similar size and can't save much there

When the game loaded from CD, this might have been needed to avoid a delay when swapping armor

AJenbo:
I'm guessing all of this was done as a performance optimization to avoid having to allocate mem during the game's runtime. Just having a really hard time seeing ~8 free+malloc having much of an impact. Especially when it also has to load the sprites from the disk. I think this is one of the only parts of the code where `LoadFileWithMem` is being used.

AJenbo:
@glebm oh here is an easy 0.2mb save, in doom.cpp it preallocates `pDoomCel` to `0x39000` bytes. There really isn't a reason to do so, it's only used in town, in hellfire, if you have the crypt map, and it's probably not the size of the cel file anyway.

glebm:
Looks like `pDoomCel` is only allocated when the player uses `IMISC_MAPOFDOOM`

PR #1406, opened 2021-04-07
Generate Mac build artifact

lourinaldi:
Taking a crack at https://github.com/diasurgical/devilution/issues/2192

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
       with:
-        fetch-depth: 2
+        fetch-depth: 0
       #with:
         # The branch, tag or SHA to checkout. When checking out the repository that
         # triggered a workflow, this defaults to the reference or SHA for that event.
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -17,11 +17,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-      with:
+      #with:
         # The branch, tag or SHA to checkout. When checking out the repository that
         # triggered a workflow, this defaults to the reference or SHA for that event.
         # Otherwise, uses the default branch.
-        ref: 'latest'
+        #ref: 'latest'
         #ref: '1.2.0'
     
     - name: Create Build Environment
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -19,16 +19,8 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
-      #with:
-        # The branch, tag or SHA to checkout. When checking out the repository that
-        # triggered a workflow, this defaults to the reference or SHA for that event.
-        # Otherwise, uses the default branch.
-        #ref: 'latest'
-        #ref: '1.2.0'
     
     - name: Create Build Environment
-      # Some projects don't allow in-source building, so create a separate build directory
-      # We'll use this as our working directory for all subsequent commands
       run: brew bundle install
 
     - name: Configure CMake
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -42,9 +42,12 @@ jobs:
       working-directory: ${{github.workspace}}/build
       shell: bash
       # Execute the build.  You can specify a specific target with "--target <NAME>"
-      run: cmake --build . -j 2 ; zip -r devilutionx-mac.zip devilutionx.app
+      run: cmake --build . -j 2 --target package
     
-    - uses: actions/upload-artifact@v2
+    - name: Release
+      uses: softprops/action-gh-release@v1
+      if: startsWith(github.ref, 'refs/tags/')
       with:
-        name: devilutionx-mac
-        path: ${{github.workspace}}/build/devilutionx-mac.zip
+        files: ${{github.workspace}}/build/devilutionx.dmg
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


AJenbo:
looks good, could you also run the package step so that it generates a .dmg instead of generating a .zip?
`cmake --build build -j 2 --target package`

AJenbo:
Will this also run a test build each time we commit or will it only build on a release tag? We mostly need this so that we can see that a PR doesn't break Mac and get some build logs that can be helpful in catching issues.

lourinaldi:
@AJenbo based on the docs at https://github.com/softprops/action-gh-release it seems it will limit releases to pushes to tags, but will run the build every time, regardless. I do make one assumption that the `latest` tag expands to a semantic version in the GH runner. Artifact uploading of anything other than the dmg itself would need to be added.

lourinaldi:
@AJenbo also if you would prefer that it check out something other than the `latest` tag (a particular branch, for example), please let me know specifically and I will make that change.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,53 @@
...
+      run: brew bundle install ; cmake -E make_directory ${{github.workspace}}/build
+
+    - name: Configure CMake
+      # Use a bash shell so we can use the same syntax for environment variable
+      # access regardless of the host operating system
+      shell: bash
+      working-directory: ${{github.workspace}}/build
+      # Note the current convention is to use the -S and -B options here to specify source 
+      # and build directories, but this is only available with CMake 3.13 and higher.  
+      # The CMake binaries on the Github Actions machines are (as of this writing) 3.12
```
AJenbo, line 38: We require CMake 3.13 so if that is still the case this will be an issue.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,53 @@
...
+        # The branch, tag or SHA to checkout. When checking out the repository that
+        # triggered a workflow, this defaults to the reference or SHA for that event.
+        # Otherwise, uses the default branch.
+        ref: 'latest'
+        #ref: '1.2.0'
+    
+    - name: Create Build Environment
+      # Some projects don't allow in-source building, so create a separate build directory
+      # We'll use this as our working directory for all subsequent commands
+      run: brew bundle install ; cmake -E make_directory ${{github.workspace}}/build
```
AJenbo, line 29: The project includes an empty build folder
```suggestion
      run: brew bundle install
```

AJenbo:
I wish there was a way to trigger the build for the current branch so that we could see it working before the merge :(

.github/workflows/cmake.yml
lourinaldi: according to this: https://github.com/jwlawson/actions-setup-cmake

...it seems like 3.13 is no longer the newest available. Which is good news for us.

lourinaldi:
> I wish there was a way to trigger the build for the current branch so that we could see it working before the merge :(

That's why I asked which branch specifically you wanted to be checked out by the action.  We can set it to whatever you want.

AJenbo:
Well, preferably this should run on any pull request so that we know if it breaks *before* it gets merged. So it should just run on every commit to master and ever PR against master if that is the question you are asking? But I don't understand what the alternative would be?

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,54 @@
...
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        # The branch, tag or SHA to checkout. When checking out the repository that
+        # triggered a workflow, this defaults to the reference or SHA for that event.
+        # Otherwise, uses the default branch.
+        ref: 'latest'
+        #ref: '1.2.0'
```
AJenbo, line 25: If I'm reading it right you can simply remove this to get the build to work as intended.

.github/workflows/cmake.yml
AJenbo: ...3.12, but yes looks like they upgraded to 3.20 so that should do nicely for us.

.github/workflows/cmake.yml
lourinaldi: After removing the `with.ref` for checkout, the build fails with `error: expected expression sgGameInitInfo.versionPatch = PROJECT_VERSION_PATCH;`

A quick search shows that others working on the project have had similar problems. It seems we may need to provide the version manually to CMake since git isn't doing it properly?

AJenbo:
Looks like it is failing to find the version tag in git because it's checked out as a hallow clone:` --depth=1`

You can get around this by either making a cull clone, or providing the version via CMake parameters if it's available via env:
`cmake .. -DVERSION_NUM=<latest-tag> -DVERSION_SUFFIX=<SHORT-SHA>`

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,54 @@
...
+  build:
+    # The CMake configure and build commands are platform agnostic and should work equally
+    # well on Windows or Mac.  You can convert this to a matrix build if you need
+    # cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      #with:
```
AJenbo, line 20:
```suggestion
    - uses: actions/checkout@v2
      with:
        fetch-depth: 0
```

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,56 @@
...
+        # The branch, tag or SHA to checkout. When checking out the repository that
+        # triggered a workflow, this defaults to the reference or SHA for that event.
+        # Otherwise, uses the default branch.
+        #ref: 'latest'
+        #ref: '1.2.0'
+    
+    - name: Create Build Environment
+      # Some projects don't allow in-source building, so create a separate build directory
+      # We'll use this as our working directory for all subsequent commands
```
AJenbo, line 31: This nolonger applies

```suggestion
    
    - name: Create Build Environment
```

AJenbo:
Nice work!

PR #1409, opened 2021-04-08
:recycle: Reuse 'player.GetMaximumAttributeValue' for max stats

julealgon:
This PR refactors all places that access maximum player stats to rely on the same function in `PlayerStruct` to do it.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -115,17 +115,7 @@ int ToBlkTbl[NUM_CLASSES] = {
 	25,
 	30,
 };
-/** Maps from player_class to maximum stats. */
-int MaxStats[NUM_CLASSES][4] = {
-	// clang-format off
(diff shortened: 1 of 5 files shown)


Source/player.cpp
```diff
@@ -207,6 +197,17 @@ Sint32 PlayerStruct::GetBaseAttributeValue(attribute_id attribute) const
 
 Sint32 PlayerStruct::GetMaximumAttributeValue(attribute_id attribute) const
 {
+	static int MaxStats[NUM_CLASSES][4] = {
```
glebm, line 200: `static const int`?

glebm (dismissed):

Source/player.cpp
julealgon: Done, thanks.

glebm (approved):

PR #1411, opened 2021-04-08
Vita: set(PREFILL_PLAYER_NAME ON)

glebm:
Not strictly necessary because we always do this now if the gamepad is active, just adding for consistency with other handheld ports


julealgon (approved):

AJenbo:
> Not strictly necessary because we always do this now if the gamepad is active, just adding for consistency with other handheld ports

Smart move :)

PR #1413, opened 2021-04-08
Monster max HP scaling based on the number of active players

mjsonofharry:
By default, monsters in multiplayer have about 2x the max health of their singleplayer counterparts. This PR adds an alternative to this; instead, monster max health scales based on the number of active players. 

This is controlled via two new ini settings;

 - **Monster HP Scaling**: Boolean setting that controls whether or not the new scaling behavior should be used. By default it's disabled, which means the original calculations are used.
 - **Monster HP Scaling Percentage**: If the new scaling behavior is enabled, then this integer setting controls the percentage. By default it's 33.

If there is only one player in the game, then monsters will have 100% of their singleplayer max HP. If the scaling percentage is set to 33%, then monsters will gain an extra 33% of their singleplayer max HP for each player who joins.

Rationale:

Enabling this feature makes the multiplayer mode friendlier to newcomers. You can set the scaling percentage to 0% for singleplayer values, or to something like 15% or 25% for a modest difficulty increase. 33% was selected for the default because it approximately doubles monster health with 4 players. Veteran players could also set the modifier much higher if they want to actually increase the difficulty.

Drawbacks:

The max HP calculation is still only done once, during monster initialization. I think it would make sense to check the active player count repeatedly until discovery by a player. I am still learning how `monster.cpp` works though, so I am not sure how to tell if a monster has encountered a player.


julealgon:
Not a fan of having balance changes like this configurable in the ini. I think this is mod territory. Even if we added the infrastructure to do something like this, _actually_ doing it should be reserved for a mod.

mjsonofharry:
> 
> 
> Not a fan of having balance changes like this configurable in the ini. I think this is mod territory. Even if we added the infrastructure to do something like this, _actually_ doing it should be reserved for a mod.

Hey, thanks for the feedback. If this isn't a desirable addition to the project then I can go ahead and close out the PR.

julealgon:
> Hey, thanks for the feedback.

Cheers. I appreciate more people trying to contribute :)

> If this isn't a desirable addition to the project then I can go ahead and close out the PR.

Please don't take my opinion as the final say. I'm not a maintainer in the project, so that decision should ultimately be on them. Would recommend keeping your PR as you intend until further notice.

mjsonofharry:
> Cheers. I appreciate more people trying to contribute :)

I'm not used to making public PRs, so thanks for saying this; it's encouraging.

> Please don't take my opinion as the final say. I'm not a maintainer in the project, so that decision should ultimately be on them. Would recommend keeping your PR as you intend until further notice.

I will follow your advice and keep it open for now. Thanks again for taking the time to review this.

AJenbo:
I agree that this is entering into mod territory and game balancing. At the same time this is something that the game handles very poorly.

As i read this PR the monsters would only be affected on level generation? Meaning if only one player is present and goes to a level before 3 others join the monsters would be very weak, but if they go to the next level and everyone leaves then the monsters will be very tough for the remaning player?

I think this would make a really good base/example mod, but we unfortunately aren't really ready for that in the current state as things are still pretty much hard-coded.

mjsonofharry:
> As i read this PR the monsters would only be affected on level generation?

I believe this is correct. I only changed the existing max health calculation, which takes place during monster initialization. I think this only happens once, when the level is generated.

The main reason I'm not confident about this is that my testing methods are a bit crude. I haven't figured out how to get runtime logging / stdout, so right now I test my changes by shooting arrows at monsters.

> Meaning if only one player is present and goes to a level before 3 others join the monsters would be very weak, but if they go to the next level and everyone leaves then the monsters will be very tough for the remaning player?

When you write it out like that it sounds pretty bad actually. This does give me an idea though. During initialization I could store the base max HP value for each monster. Then, during each update cycle, I can check if the monster needs to update its max HP based on the current number of active players. If the max HP needs to change, I only need to save its current health percentage. So if it's 500/1000 hit points I save 0.5 as its current health. This means that if a player joins or leaves, the monster's health will increase or decrease respectively. But if you have health bars enabled you wouldn't see any change because, relatively speaking, there wasn't a change.

> I think this would make a really good base/example mod, but we unfortunately aren't really ready for that in the current state as things are still pretty much hard-coded.

Thank you! And, yeah, if there's no way to integrate these changes that's totally fair.

AJenbo:
> The main reason I'm not confident about this is that my testing methods are a bit crude. I haven't figured out how to get runtime logging / stdout, so right now I test my changes by shooting arrows at monsters.

SDL_Log()

AJenbo:
> During initialization I could store the base max HP value for each monster.

This is already accessible via the monster type

mjsonofharry:
> This is already accessible via the monster type

There's a max HP value for the monster type, but isn't each  individual monster instance's max HP [slightly randomized](https://github.com/diasurgical/devilutionX/blob/7ce1deb51439475bbb4187854acd404573dd34f6/Source/monster.cpp#L479)? That per-instance max HP is the value that I'd need to persist, in order to ensure consistency.

Also SDL_Log() is a lifesaver, thank you :pray:

AJenbo:
i guess an alternative method would be to devide the damage by the ammount of active players :D

qndel:
> i guess an alternative method would be to devide the damage by the ammount of active players :D

lol

galaxyhaxz:
Any plans to also increase experience for more players in the game? In Diablo 2 each player in the game adds 50% more monster life + experience iirc. It actually gives an incentive then to crank it up.

mjsonofharry:
> i guess an alternative method would be to devide the damage by the ammount of active players :D

Good point! But I think the most authentic method would be to use bitwise operations ;)

> Any plans to also increase experience for more players in the game? In Diablo 2 each player in the game adds 50% more monster life + experience iirc. It actually gives an incentive then to crank it up.

I thought of this and wasn't originally planning on doing it. But if max HP scaling is going to be a mod then I suppose there's no reason I can't go crazy and do experience scaling as well. I think I would implement it with the same kind of ini controls--one setting to enable it and another to control the scaling modifier.

Mods aside, I would still recommend considering some kind of user-facing control over the 2x monster HP buff. Replacing it with custom scaling behavior is one thing, but what about just having a flag for disabling it? IIRC there's also a cap on experience gains per kill in multiplayer. I think it'd be nice to be able to disable that as well.

julealgon:
@mjsonofharry , you might consider marking this PR as a draft while there are still discussions going on about the implementation.

mjsonofharry:
So I've been up to a few things this weekend. First of all, I have [a new feature branch](https://github.com/mjsonofharry/devilutionX/pull/3/files) that introduces two new ini settings:
 - Reduce Enemy Health: removes the 2x monster HP buff in multiplayer
 - Power Leveling: removes the power level cap in multiplayer

The ini settings are aimed at fulfilling the core intent of _this_ PR (allowing inexperienced players to remove some of the more arbitrary barriers to entry) without introducing custom behavior. I would be interested in opening a new PR for this.

Second, I started experimenting with a different approach to enemy health rescaling. Rescaling now happens on every update pass. This means players can join and leave at any time and any number of times, and the monsters will all change their hitpoints and max health values accordingly. The behavior is pretty much [what I described above](https://github.com/diasurgical/devilutionX/pull/1413#issuecomment-816497662). From my testing it appears pretty promising so far! You can see the code changes for yourself on this PR.

I also looked into approximating HP scaling using player damage debuffs. Someone with better algebra skills than me could probably do it, but I wasn't able to produce a high enough quality algorithm. The TTK would be slightly different in any of my solutions. So I've shelved that approach for now.

PR #1416, opened 2021-04-08
Network packets have network byte order

AJenbo:


Source/msg.cpp
```diff
@@ -1361,7 +1361,7 @@ static DWORD On_SBSPELL(TCmd *pCmd, int pnum)
 	TCmdParam1 *p = (TCmdParam1 *)pCmd;
 
 	if (gbBufferMsgs != 1) {
-		spell_id spell = (spell_id)p->wParam1;
+		spell_id spell = (spell_id)(Uint16)p->wParam1;
```
AJenbo, line 1364: Most of theaes should be Sint not Uint

Source/utils/endian_int.h
```diff
@@ -0,0 +1,118 @@
...
+	}
+};
+
+template<class N, class T, T (*f)(T)>
+N endian_cast(const endian_int<T, f>& in)
+{
+	return static_cast<N>(static_cast<T>(in));
+}
+
+typedef endian_int<Uint16, SwapBE> Uint16_BE;
```
glebm, line 97: code style: Use `using` instead of typedef (`using Uint16_BE = endian_int<Uint16, SwapBE>;`)

Source/utils/endian_int.h
```diff
@@ -0,0 +1,118 @@
+#pragma once
+
+namespace devilution {
+
+template <class T>
+T SwapLE(T in)
+{
+	static_assert((sizeof(T) == 1) || (sizeof(T) == 2) ||
+	              (sizeof(T) == 4) || (sizeof(T) == 8),
+	              "SwapLE called for unsupported size");
```
glebm, line 10: Perhaps it'd be cleaner to just have 8 functions?

```c++
inline std::int8_t SwapLE(std::int8_t in) { return in; }
inline std::uint8_t SwapLE(std::uint8_t in) { return in; }
inline std::int16_t SwapLE(std::int16_t in) { return static_cast<std::int16_t>(SDL_SwapLE16(static_cast<std::uint16_t>(in))); }
inline std::uint16_t SwapLE(std::uint16_t in) { return SDL_SwapLE16(in); }
...
```

glebm (approved):

glebm (dismissed):

Source/utils/endian_int.h
```diff
@@ -0,0 +1,114 @@
...
+
+#include <type_traits>
+
+namespace devilution {
+
+template<typename T>
+T SwapLE(T in)
+{
+	static_assert(std::is_integral_v<T> || std::is_enum_v<T>);
+	static_assert((sizeof(T) == 1) || (sizeof(T) == 2) || (sizeof(T) == 4) || (sizeof(T) == 8));
```
glebm, line 11: `static_assert` requires a message on older compilers

```suggestion
	static_assert(std::is_integral_v<T> || std::is_enum_v<T>, "unsupported type");
	static_assert((sizeof(T) == 1) || (sizeof(T) == 2) || (sizeof(T) == 4) || (sizeof(T) == 8), "unsupported type");
```

Source/utils/endian_int.h
```diff
@@ -0,0 +1,114 @@
...
+		return static_cast<T>(SDL_SwapLE32(static_cast<Uint32>(in)));
+	if constexpr (sizeof(T) == 8)
+		return static_cast<T>(SDL_SwapLE64(static_cast<Uint64>(in)));
+}
+
+template<typename T>
+T SwapBE(T in)
+{
+	static_assert(std::is_integral_v<T> || std::is_enum_v<T>);
+	static_assert((sizeof(T) == 1) || (sizeof(T) == 2) || (sizeof(T) == 4) || (sizeof(T) == 8));
```
glebm, line 26:
```suggestion
	static_assert(std::is_integral_v<T> || std::is_enum_v<T>, "unsupported type");
	static_assert((sizeof(T) == 1) || (sizeof(T) == 2) || (sizeof(T) == 4) || (sizeof(T) == 8), "unsupported type");
```

glebm:
Error:

>  Source/multi.cpp:647:57: error: cannot pass non-trivial object of type 'devilution::int32_net_t' (aka 'endian_int<int, SwapBE>') to variadic function; expected type from format string was 'int' [-Wnon-pod-varargs]
                        app_fatal("Invalid size of game data: %d", gameData->size);

This needs an explicit cast, implicit one doesn't work here.

Source/utils/endian_int.h
```diff
@@ -0,0 +1,114 @@
+#pragma once
+
+#include <type_traits>
+
+namespace devilution {
+
+template<typename T>
+T SwapLE(T in)
+{
+	static_assert(std::is_integral_v<T> || std::is_enum_v<T>, "unsupported type");
```
glebm, line 10: `std::is_integral_v` and `std::is_enum_v` are not available on older GCC, use `std::is_integral<T>::value` and `std::is_enum<T>::value` instead

glebm:
Looks like `if constexpr` is not support on GCC 6 at all

glebm (approved):

PR #1417, opened 2021-04-08
Optimizes some struct layouts

glebm:
Optimizes some struct layouts by ordering fields to minimize padding, and using the smallest appropriate size types.

Overall, reduces the stripped release x64 Linux binary size by ~10 KiB (less than I'd hoped)


Source/monstdat.h
```diff
@@ -75,42 +77,42 @@ typedef enum monster_resistance {
...
 
 typedef struct MonsterData {
-	Sint32 width;
-	Sint32 mImage;
+	const char *mName;
 	const char *GraphicType;
-	bool has_special;
 	const char *sndfile;
+	const char *TransFile;
+	Uint16 width;
```
AJenbo, line 84: I think 160 is about as big as they get, so potentially there is a bit more that could be cut away, but not very future proof of cause.

Source/misdat.cpp
```diff
@@ -10,182 +10,182 @@ namespace devilution {
...
-	{  MIS_ARCHLICH,      &AddFlare,                  &MI_Firebolt,       true,      1, MISR_MAGIC,     MFILE_ARCHLICH,  -1,          -1          },
-	{  MIS_BONEDEMON,     &AddFlare,                  &MI_Firebolt,       true,      1, MISR_MAGIC,     MFILE_BONEDEMON, -1,          -1          },
-	{  MIS_EXYEL2,        &AddMisexp,                 &MI_Misexp,         true,      2, MISR_NONE,      MFILE_EXYEL2,    LS_FIRIMP2,  -1          },
-	{  MIS_EXRED3,        &AddMisexp,                 &MI_Misexp,         true,      2, MISR_NONE,      MFILE_EXRED3,    LS_FIRIMP2,  -1          },
-	{  MIS_EXBL2,         &AddMisexp,                 &MI_Misexp,         true,      2, MISR_NONE,      MFILE_EXBL2,     LS_FIRIMP2,  -1          },
-	{  MIS_EXBL3,         &AddMisexp,                 &MI_Misexp,         true,      2, MISR_NONE,      MFILE_EXBL3,     LS_FIRIMP2,  -1          },
-	{  MIS_EXORA1,        &AddMisexp,                 &MI_Misexp,         true,      2, MISR_NONE,      MFILE_EXORA1,    LS_FIRIMP2,  -1          },
+	// mAddProc,                   mProc,              mName,             mDraw, mType, mResist,        mFileNum,        miSFX,       mlSFX;
```
AJenbo, line 13: Any specific reason to swap around the fields?

glebm: To reduce the size of the struct by reducing the required padding

A struct like this takes up 16 bytes

```
uint8_t a;
uint64_t b;
```

A struct like this takes up a lot less:

```
uint64_t b;
uint8_t b;
```

AJenbo (approved):

Source/misdat.cpp
AJenbo: Oh yeah, I dealt a lot with that in the early stage of this project that's basically how I read structs at this point :D

AJenbo:
Next stop PSX compatability

julealgon:
... really? The compiler does not optimize the order of the fields in memory by itself?

Yet another change that hurts the codebase in favor of "optimizing by default". Now people need to be aware of member order instead of just making it an irrelevant aspect that can be used for other purposes like readability and refactor friendliness.

We'll reach a point where touching the code anywhere will create unintended behavior changes.

glebm:
In C and C++, member order is part of the struct definition.

The compiler is not allowed to optimize it.

Generally it doesn't matter but these are used to store somewhat heavy static data (order of ~10 KiB).

glebm:
This is one of the things Rust fixes (structs can be reordered by the compiler)

glebm:
> We'll reach a point where touching the code anywhere will create unintended behavior changes.

"will reach"? :sweat_smile:

julealgon:
> Generally it doesn't matter but these particular structs are used to store somewhat heavy static data (order of ~10 KiB).

I don't understand what you are optimizing for: do you want the game to play in a microwave oven display? The more you optimize towards running on a calculator, the worse the code will become.

I honestly don't understand why everyone else is fine with this mindset.

glebm:
This change doesn't make things any less readable than they already were

> do you want the game to play in a microwave oven display?

Yes, this is the next DOOM, porting to everything that it can possibly run on.

AJenbo:
I find it an acceptable trade off, imo struct order is already somewhat abetrary. In going this Glebm has eliminated a lot of dead fields/code and put thought in to some of the arbitrary sizes that where already there. Instead of having parameters that randomly take char and sometimes int, they now consistently take an enum class with a defined size.
So him having a goal of getting the game running on 25MB memory is offset by his willingness to do the work by cleaning up garbage.

The reason compilers aren't allowed to reorder is that you can also access structs as arrays, this was how quests where done until recently. It's absolutely terrible code, but apparently part of the spec, i imagine this somewhat relates to us being able to just mem copy between files, network and memory.

PR #1418, opened 2021-04-08
:fire: remove unused parts of init_test_access

AJenbo:


Source/pfile.cpp
```diff
@@ -164,7 +164,7 @@ static HANDLE pfile_open_save_archive(DWORD save_num)
 {
 	HANDLE archive;
 
-	if (SFileOpenArchive(GetSavePath(save_num).c_str(), 0x7000, FS_PC, &archive))
+	if (SFileOpenArchive(GetSavePath(save_num).c_str(), 0, 0, &archive))
```
glebm, line 167: Doesn't removing 0x7000 change behaviour?

Here StormLib uses this flag:

https://github.com/ladislav-zezula/StormLib/blob/dc867243c099f0350229335f8e283cce2016105a/src/SFileOpenArchive.cpp#L216

Source/misdat.h
```diff
@@ -5,19 +5,104 @@
...
 #pragma once
 
-#include "missiles.h"
-
 namespace devilution {
 
-#ifdef __cplusplus
-extern "C" {
-#endif
+typedef enum missile_resistance : uint8_t {
```
glebm, line 10: `#include <cstdint>` for `std::uint8_t` or `#include <stdint.h>` for `uint8_t`.

Source/misdat.h
```diff
@@ -5,19 +5,104 @@
...
+	MISR_NONE,
+	MISR_FIRE,
+	MISR_LIGHTNING,
+	MISR_MAGIC,
+	MISR_ACID,
+} missile_resistance;
+
+typedef enum missile_graphic_id : uint8_t {
```
glebm, line 18: If we're using C++ there is no need for the `typedef` stuff on structs or enums by the way.
It's just `enum A {}; struct B {};`

Source/pfile.cpp
AJenbo: 0x7000 went to dwPriority, either this isn't implemented in storm lib or we made a mistake by implementing this our selfs in `SFileOpenFile`

glebm (dismissed):

Issue #1367, opened 2021-04-04
Items not dropping and becomming invalid when no save games are present


SoundChaser83:
**Important information**
Windows/Wine 64/32, version 1.2.0

**Describe**
Items become invalid when dropped in town on the very first game. This is caused the additional full `heroitems` file not being saved in the save game archive on Windows, which also means that migrating heros between diablo and hellfire will cause morphing on new game.

**To Reproduce**
1: delete all saves before starting devilutionx.
2: create a warrior
3: enter the church
4: leave the church
5: drop the club
6: the item will be stuck in mid air

**Expected behavior**
The item should drop to the ground, and it should be possible to pick it up.

**Screenshots**
![image](https://user-images.githubusercontent.com/204594/113957825-480b2f00-9820-11eb-9c18-56df2e0ee2fd.png)

**Workaround**
Simply create a save game (enter town go to the menu and select save) and then restart the game. New games should now work properly.

AJenbo:
It looks like `gbIsHellfire` ended up being `false` in the middle of a hellfire game.

If this happened in relation to the oil, or it was just the first hellfire item to drop after it happened isn't clear.

AJenbo:
I'm not able to replicate the catastrophic game state, despite the one item in town clearly being broken in the save. Tested both on Linux and a 64bit windows build using Wine.

One thing i do note is that when ever you return from a town portal to Leoric Tomb the light radious is significantly reduced and not fully processed.

But I think this shows that the issue stems from the game memory having become corrupt on the specific machine where this happened.

If someone on Windows could test what happens if they create a new game, go to Leoric Tomb and tp to and from town that would be appreciated.

AJenbo:
The lighting issue was also there in 1.1.0 and probably vanilla as well (needs verification).

Chance4us:
@SoundChaser83 Thank you very much for playtesting.That is very much appreciated here.
As I saw you played the barbarian for the first time.
The recommended strategy is an axe. The axe is faster and more damage with the barbarian.
However, you also need an item with a chance to hit or an axe with chance to hit.
Vitality goes up to 150 and that should almost always be built up.

Chance4us:
@AJenbo
Since you already talked about the damage caused by lightning wall / fire wall in the video, I  tried a game with the sorcerer on hell difficulty. I used lightning wall and fire wall to keep Leoric away.
In addition, I fired with Level 3 Holy Bolt spell.
[...]

PR #1423, opened 2021-04-08
loadsave.cpp: Fix pfile_write_save_file call

glebm:
Refs #1367

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -170,7 +170,7 @@ public:
 		if (m_buffer == nullptr)
 			return;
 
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Thank you for tracking this down. The idea with m_bufferPtr was that it would crop the buffer to just the amount of data that was written to it instead of whatever was allocated. I guess that didn't work out :(

glebm:
>  I guess that didn't work out :(

Well there is no code that actually does it

glebm:
It's a trivial fix, 1 sec

glebm:
Wait, a variable called `m_bufferPtr` is not actually a pointer

Perhaps this fix doesn't actually fix anything

glebm:
Yeah nevermind

PR #1425, opened 2021-04-08
Enumse

AJenbo:


Source/misdat.h
```diff
@@ -9,6 +9,12 @@
 
 namespace devilution {
 
+enum mienemy_type : uint8_t {
```
julealgon, line 12: Why name it like this? Can't we use a more meaningful name like `MissileTarget`?

julealgon, line 12: Can't we use `enum class` for new enums?

julealgon (changes requested):

glebm (approved):

Source/misdat.h
glebm: Yeah we should

One big advantage of `enum class` other than scoping is that it's not implicitly convertible to int (requires a `static_cast`)

AJenbo: It's the current name, I don't feel like renaming things atm as you know my spelling is piss poor so it both take longer and is significantly more likely to contain mistakes. Hopefully some one else will rename things in to sensible names using our chosen code style (camel case). I like your suggested name, feel free to open a PR with it.

AJenbo: sure, for new ones 🤷

PR #1426, opened 2021-04-08
Only write to the file once when saving

glebm:
I've tested this a bit but it needs to be tested further.


AJenbo (approved):
Once step closer to having sensible save game handling 🎉

PR #1430, opened 2021-04-08
add an ini option to configure gamepad deadzone

geowarin:
This adds an ini option under Controller.deadzone, read as a float.

I had some drift on my Xbox gamepads that made the use of the emulate pointer feature on the gamepad very inpractical.
So I figured, why not make this a user configurable value?

My C++ is very rusty so I'm very eager to hear your reviews.

Here are some things I am unsure about:

- Saving the deadzone in the ini file: I opted for 2 point float precision for this. This can be changed easily if you think it is necessary.
- Default deadzone value of 0.07 is untouched but this seems very low and it is likely that other xbox gamepad users are affected. Maybe we should consider making this a bit bigger, e.g: 0.15
- Maybe we should consider splitting joysticks left and right deadzones instead of having a single value?

### DIFF
--- a/3rdParty/Storm/Source/storm.h
+++ b/3rdParty/Storm/Source/storm.h
@@ -262,11 +262,13 @@ BOOL
         DWORD *pdwBpp);
 
 bool getIniBool(const char *sectionName, const char *keyName, bool defaultValue = false);
+float getIniFloat(const char *sectionName, const char *keyName, float defaultValue);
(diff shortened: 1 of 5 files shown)


julealgon:
> Maybe we should consider splitting joysticks left and right deadzones instead of having a single value?

That makes a lot of sense to me considering each analog can wear differently and end up being calibrated differently.

AJenbo:
thanks, looks good

as i remember one of mine also drifts but the receiver died during development so not sure at this point

PR #1435, opened 2021-04-09
Fix audio distortion on Windows (default to winmm)

### LINKED ISSUE #1434: Noisy audio VS2019 Debug SDL 2.0.14 Win 10 x64
Hearing noisy audio with this setup.

The audio had distortion at both high and low frequencies

A workaround that fixed the issue for me is `SDL_AUDIODRIVER=winmm`

For now, we should probably make sure we use the `winmm` driver on Windows (e.g. by calling `SDL_InitAudio("winmm")` manually).

This may be fixed in SDL HEAD by https://github.com/libsdl-org/SDL/commit/b98b5adcaea159fc6a9753f808875acf7d3ee945 though I'm not sure

glebm:
Fixes #1434

### DIFF
--- a/SourceX/display.cpp
+++ b/SourceX/display.cpp
@@ -125,6 +125,12 @@ bool SpawnWindow(const char *lpWindowName)
 	SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
 #endif
 
+#ifdef _WIN32
+	// The default WASAPI backend causes distortions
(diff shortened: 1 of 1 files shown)


AJenbo:
Reading the comments this makes a certain amount of sens.

AJenbo:
p.s. if you would at last use `:bug:` on bug fixes like this it would make it easier on me when doing release notes and cherry-picking fixes for a patch release.

PR #1438, opened 2021-04-09
Use proper bools

AJenbo:


julealgon:
❤️

Source/automap.cpp
```diff
@@ -468,7 +468,7 @@ Sint32 AmLine4;
 
 void InitAutomapOnce()
 {
-	automapflag = FALSE;
+	automapflag = true;
```
AJenbo, line 471: i fucked up my first attempt, bad replace.

PR #1442, opened 2021-04-09
Use custom IME handler on Vita

### LINKED ISSUE #1414: [Vita] Can't rename characters
Naming characters is impossible on new version. You get the default name but you cannot delete it, the input shows empty while the name of the character "Kharazim" stays, and if you add anymore letters it says "Kharazim" and then w/e you input.
Didn't happen on older versions.

isage:
Fixes #1414 
As per discussion, this implements strategy 1 (SDL_TEXTINPUT event should replace name, not append to it) and somewhat strategy 4 (except instead of SDL, setting text for IME implemented in custom handler instead).

### DIFF
--- a/SourceX/DiabloUI/diabloui.cpp
+++ b/SourceX/DiabloUI/diabloui.cpp
@@ -21,6 +21,10 @@
 // for virtual keyboard on Switch
 #include "platform/switch/keyboard.h"
 #endif
+#ifdef __vita__
+// for virtual keyboard on Vita
+#include "platform/vita/keyboard.h"
+#endif
 
 namespace devilution {
 
@@ -93,6 +97,8 @@ void UiInitList(int count, void (*fnFocus)(int value), void (*fnSelect)(int valu
 			textInputActive = true;
 #ifdef __SWITCH__
 			switch_start_text_input("", pItemUIEdit->m_value, pItemUIEdit->m_max_length, /*multiline=*/0);
+#elif defined(__vita__)
+			vita_start_text_input("", pItemUIEdit->m_value, pItemUIEdit->m_max_length);
 #else
 			SDL_StartTextInput();
 #endif
@@ -207,6 +213,12 @@ void selhero_CatToName(char *in_buf, char *out_buf, int cnt)
 {
 	std::string output = utf8_to_latin1(in_buf);
 	strncat(out_buf, output.c_str(), cnt - strlen(out_buf));
+}
+
+void selhero_SetName(char *in_buf, char *out_buf, int cnt)
+{
+	std::string output = utf8_to_latin1(in_buf);
+	strncpy(out_buf, output.c_str(), cnt);
 }
 
 bool HandleMenuAction(MenuAction menu_action)
@@ -324,7 +336,7 @@ void UiFocusNavigation(SDL_Event *event)
 #ifndef USE_SDL1
 		case SDL_TEXTINPUT:
 			if (textInputActive) {
-				selhero_CatToName(event->text.text, UiTextInput, UiTextInputLen);
+				selhero_SetName(event->text.text, UiTextInput, UiTextInputLen);
 			}
 			return;
 #endif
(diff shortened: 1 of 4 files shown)


SourceX/DiabloUI/diabloui.cpp
```diff
@@ -324,7 +336,7 @@ void UiFocusNavigation(SDL_Event *event)
 #ifndef USE_SDL1
 		case SDL_TEXTINPUT:
 			if (textInputActive) {
-				selhero_CatToName(event->text.text, UiTextInput, UiTextInputLen);
+				selhero_SetName(event->text.text, UiTextInput, UiTextInputLen);
```
glebm, line 339: This won't do the right thing if you're typing on a physical keyboard, which sends an event for each "letter"

glebm (changes requested):

SourceX/platform/vita/keyboard.h
```diff
@@ -0,0 +1,3 @@
+#pragma once
+
+void vita_start_text_input(const char *guide_text, char *initial_text, int max_length);
```
glebm, line 3: `const char *initial_text`?

SourceX/DiabloUI/diabloui.cpp
isage: Ugh. So i guess this should be done for vita only. What about switch?

glebm: Perhaps we could do something like:

```c++
if (SDL_HasScreenKeyboardSupport())
  selhero_SetName(event->text.text, UiTextInput, UiTextInputLen);
else
  selhero_CatToName(event->text.text, UiTextInput, UiTextInputLen);
```

isage: That'll probably break on ios/android/winphone, as those do have screen keyboard support, but also produce events for each key press.

glebm: In that case we have to ifdef here it seems

glebm: > What about switch?

I don't know but however it was set up before worked (at least there are no bug reports about this)

isage: It looks like this abomination https://github.com/diasurgical/devilutionX/blob/master/SourceX/platform/switch/keyboard.cpp#L55-L62
deals with prefiled text on switch.
So i've added ifdefs for both platforms and code above shouldn't be required anymore.

glebm: The Switch still currently sends individual events for each code point:

https://github.com/diasurgical/devilutionX/blob/eeccf6b5d6c8f11e9e45b93845d3a5e2aadefe5b/SourceX/platform/switch/keyboard.cpp#L68-L71

So we'll receive individual code points and they will each overwrite the text.

There is also the 3ds

isage: Oh, god dammit. Ok, i've removed switch ifdef, but that should really be fixed in switch implementation

isage: And 3ds, looks like, does OSK stuff even before name dialog. And, well, it's SDL1.
So, well, after fixing that mess in switch code, that ifdef for switch can be returned.

AJenbo (approved):

rsn8887:
Has this been tested with entering number of coins and other textinput on Vita and Switch?

I had all of this working fine on Vita and Switch a long time ago, but I guess it was broken trying to use SDL functions, which are not very good on Vita/Switch.

Anywhoo if it works now in all cases, including using physical keyboard and mouse, even better.

isage:
@rsn8887 this has nothing to do with either switch, or coins number (because in-game ui is handled elsewhere)

rsn8887:
What I am trying to say is that I had coins and name entry and all that working on Vita and Switch. Then, something broke name entering, broke coins too. So should both be fixed now. If this only fixes name entry, then the coins need to be done too. That is all.

rsn8887:
Correction: It looks like I was aware that coin amount entry didn't work on Switch, but I guess I never fixed it? So it seems it was never working correctly, neither on Switch nor on Vita.

PR #1443, opened 2021-04-09
Fix MSVC build with -DNONET=1

glebm:
Not sure why this worked on Linux without this stub


julealgon:
I read this as "DOTNET" 😆

AJenbo:
Sorry not at this point :)

PR #1445, opened 2021-04-09
Do not prefill hero name for on-screen keyboard

### LINKED ISSUE #1414: [Vita] Can't rename characters

glebm:
Refs #1414

Related: https://github.com/diasurgical/devilutionX/pull/1442

/cc @isage

### DIFF
--- a/SourceX/DiabloUI/selhero.cpp
+++ b/SourceX/DiabloUI/selhero.cpp
@@ -358,6 +358,14 @@ void selhero_ClassSelector_Focus(int value)
 
 static bool shouldPrefillHeroName()
 {
+#ifndef USE_SDL1
+	// On-screen keyboard does not allow editing existing text in SDL.
+	// Make sure the text is empty so that any name can be used.
+	// See https://github.com/diasurgical/devilutionX/issues/1414
+	if (SDL_HasScreenKeyboardSupport())
+		return false;
+#endif
+
 #if defined __3DS__
 	return false;
 #elif defined(PREFILL_PLAYER_NAME)


AJenbo:
Should we overwrite this on switch as we use it's keyboard directly, or is it better to be more vanilla and expect people to fully enter a name?

glebm:
Good point, we should probably only do this on platforms where we don't have the ability to set the initial text

isage:
The only two platforms that require that are switch and vita, afaik.
That shouldn't be required on vita after #1442
Dunno about switch, does it have same issue?

glebm:
Any SDL2 platform with OSK and no custom keyboard stuff on our side would require this

isage:
Officially, only 4 sdl platforms have OSK: android, ios, winrt and vita. Of those 4, only vita doesn't send keypresses from OSK, and has instead separate text input.

glebm:
This ifdef is not about that, it's about not initializing the value with the prefilled name, because OSK always starts with the empty string (editing not supported)

So if you didn't have this ifdef, the OSK would only be able to append to a pre-generated name. So we can't prefill the name in this case.

glebm:
This PR is incomplete in that we should *not* do this on platforms that do support the starting string for the OSK but I'm waiting for your PR before fixing that

glebm:
Ah, I think I see what you're saying: The other platforms allow sending backspace from the OSK and the actual DiabloUI input field is visible while they do that?

isage:
Yeah

glebm:
OK, closing this then :+1:

PR #1451, opened 2021-04-10
Fix gold stacking when gold stack limit is modified

austinwagner:
Equipping the Auric Amulet increases the amount gold can stack in your inventory from 5k to 10k. This works when picking gold up off the ground, but not when gaining gold through sales or attempting to drag and drop to create a stack. I've fixed this here by simply ensuring that the MaxGold variable is used when checking gold stack size instead of GOLD_MAX_LIMIT.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1382,15 +1382,15 @@ void CheckInvPaste(int pnum, int mx, int my)
 				il--;
 				gt = plr[pnum].InvList[il]._ivalue;
 				ig = plr[pnum].HoldItem._ivalue + gt;
-				if (ig <= GOLD_MAX_LIMIT) {
+				if (ig <= MaxGold) {
 					plr[pnum].InvList[il]._ivalue = ig;
 					plr[pnum]._pGold += plr[pnum].HoldItem._ivalue;
 					SetPlrHandGoldCurs(&plr[pnum].InvList[il]);
 				} else {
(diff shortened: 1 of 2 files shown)


julealgon:
When searching the codebase for `GOLD_MAX_LIMIT`, I get 34 hits:
- 1 is the define itself
- 3 others are assignments into `MaxGold`
- everything else are reads

Given the above, I'd expect 30 changes in your PR to fix gold inconsistency bugs, however it only contains 22 line changes.

Can you please take another look here? I suspect you are leaving some gold-related bugs in the codebase that we could fix in one fell swoop.

austinwagner:
I did look through all of the uses of that constant:

I didn't touch the ones inside `_DEBUG` ifdefs. It wouldn't hurt but also isn't really a bug. So if you think the cheats should also be affected I'll make that change.

There's one read in `ValidatePlayer()` which does account for Hellfire having a doubled stack limit. I didn't dig deeply enough to know if `MaxGold` could be used here, but for a basic validation I figured it wasn't necessary anyway.

`GetItemAttrs()` also reads `GOLD_MAX_LIMIT`. This function is called when spawning items and the use is to cap the randomized value when the item type is gold. I don't believe it would make sense to use `MaxGold` here, though there might be something about the purpose of this function that I'm not understanding since at the end it calls `SetPlrHandGoldCurs()`.

AJenbo:
Thanks both for making my life easier :)

PR #1454, opened 2021-04-10
:rotating_light: Fix remaning clang warnings

AJenbo:
Should try and tackle more the warnings that we silence, or enable more compiler checks next?


glebm (approved):

julealgon:
> Should try and tackle more the warnings that we silence, or enable more compiler checks next?

Would suggest first cleaning existing suppressions.

PR #1457, opened 2021-04-10
Remove top and bottom buffer borders

glebm:
They're no longer needed


AJenbo:
:+1: lets test it out for a bit and then also remove the defines

PR #1459, opened 2021-04-10
Split up OperateShrine

AJenbo:


Source/objects.cpp
```diff
@@ -3396,14 +3396,958 @@ int ItemMiscIdIdx(item_misc_id imiscid)
...
+}
+
+bool OperateShrineDivine(int pnum, int x, int y)
+{
+	if (deltaload)
+		return false;
+	if (pnum != myplr)
+		return false;
+
+	if (2 * currlevel < 7) {
```
qndel, line 3844: simplify to `currlevel < 4?`

AJenbo:
```suggestion
	if (currlevel < 4) {
```

PR #1334, opened 2021-04-03
Log multiplayer events (player joined/left and chat)

Jmgr:
This logs player events (joined, left) and the chat, as well as player errors.

Fixes #672.

PR #1464, opened 2021-04-10
Add a logger

Jmgr:
This adds a logger based on the [spdlog](https://github.com/gabime/spdlog) library. See that page for a complete list of features, but to summarize:
 * it's type safe (no more pondering if you need to put %lld or %ld or %s and no crash if you got it wrong, you just use {} everywhere)
 * it's fast
 * it supports log levels (see [this](https://stackoverflow.com/a/2031209) for an example on how to interpret each level)
 * it supports categories (I've set-up a few basic ones)
 * on non-debug builds all logging calls at levels debug and trace are not compiled in
 * on debug build the trace calls are not enabled by default (CMake configuration option)
 * every log entry is written to stdout (with colors!) as well as in a file (by default `log.txt`, but can be changed/disabled by command line). I think it could be good to keep the log file on release builds, because that could help with debugging error encountered by players.

I've also added support for precompiled headers, when available (CMake >= 3.16), to accelerate the compilation. More "stable" headers can be added to `pch.hpp` as needed (but I would not recommend adding `all.hpp`).

Note that not all `SDL_Log` calls have been replaced yet. I've created this to get some feedback, but will continue replacing those calls as needed. Please check if I got the log level right, sometimes it's not easy to see if something should be a warning or an error.


qndel:
You forgot one feature list
* it's only half as long as the entire diablo's code 🤪

AJenbo:
Could you add this via CMake or another dependency manager so we don't blot the repo?

Jmgr:
> You forgot one feature
> 
>     * it's only half as long as the entire diablo's code zany_face

That's only because I had to add spdlog. Embedding dependencies is not great...

> Could you add this via CMake or another dependency manager so we don't blot the repo?

I though all dependencies should be part of the repository? Isn't that what it currently done?

Jmgr:
One possible solution seems to be https://cmake.org/cmake/help/latest/module/ExternalProject.html Never tried, but sounds good.

AJenbo:
> One possible solution seems to be https://cmake.org/cmake/help/latest/module/ExternalProject.html Never tried, but sounds good.

You can see how libzt etc where added to cmake here:

https://github.com/diasurgical/devilutionX/commit/867dd96a73898de12e81e1d9dc4fef5cb2f53123#diff-2883d1afaaf1c9208f7b4da3d2c008c581b011d90ee73deb387ab9940ea7b6ed

Jmgr:
> > One possible solution seems to be [cmake.org/cmake/help/latest/module/ExternalProject.html](https://cmake.org/cmake/help/latest/module/ExternalProject.html) Never tried, but sounds good.
> 
> You can see how libzt etc where added to cmake here:
> 
> [867dd96#diff-2883d1afaaf1c9208f7b4da3d2c008c581b011d90ee73deb387ab9940ea7b6ed](https://github.com/diasurgical/devilutionX/commit/867dd96a73898de12e81e1d9dc4fef5cb2f53123#diff-2883d1afaaf1c9208f7b4da3d2c008c581b011d90ee73deb387ab9940ea7b6ed)

Yup, I just found `FetchContent`. Seems to work, but now there are a few compilation issues on some platforms.

AJenbo:
What are we planing to log by the way?

Jmgr:
> What are we planing to log by the way?

Warnings and errors are usually logged, but I would also include level/item/monster generation (some of them as debug). That should help with debugging, both for developers and players (you could ask players who want to report a bug to add the log file as an attachment). You could also add a way for players to report a bug from within the game itself: a shortcut or a button that would take a screenshot and upload it together with the log file.

There was an interesting talk at FOSDEM 2019 about allowing players to report bugs from within the game: https://archive.fosdem.org/2019/schedule/event/0ad_graphics/

The original idea came from https://github.com/diasurgical/devilutionX/pull/1334.

Jmgr:
I would also add very detailed logging (like each time a character attacks something) at the trace level (so disabled by default).

AJenbo:
Since SDL already has build-in logging functionality I would prefer that we make use of that. It doesn't seam that this PR integrates with that functionality?
I think all we need to do to achieve what we are looking for here is to add a file output function for the SDL logger:
https://wiki.libsdl.org/SDL_LogSetOutputFunction

The code is already sprinkled with a bunch of SDL_Log() lines both for critical, notices, and warning level issues. Though they are all being reported as "Info" atm.

CMakeLists.txt
```diff
@@ -183,6 +184,18 @@ else()
   find_package(SDL2_mixer REQUIRED)
 endif()
 
+include(FetchContent)
```
glebm, line 187: Please set this up in 3rdParty/spdlog/CMakeLists.txt

CMakeLists.txt
```diff
@@ -183,6 +184,18 @@ else()
...
+FetchContent_Declare(
+  spdlog
+  GIT_REPOSITORY https://github.com/gabime/spdlog.git
+  GIT_TAG        100f30043f33277122e0991c83845a2617172ffd
+)
+FetchContent_GetProperties(spdlog)
+if(NOT spdlog_POPULATED)
+  FetchContent_Populate(spdlog)
+  add_subdirectory(${spdlog_SOURCE_DIR} ${spdlog_BINARY_DIR})
+endif()
```
glebm, line 197: Use our `FetchContent_MakeAvailableExcludeFromAll` helper:

```suggestion
include(FetchContent_MakeAvailableExcludeFromAll)
FetchContent_MakeAvailableExcludeFromAll(spdlog)
```

CMakeLists.txt
```diff
@@ -435,21 +449,37 @@ if(GPERF)
...
-target_include_directories(devilution PUBLIC Source SourceS ${CMAKE_CURRENT_BINARY_DIR})
+target_compile_definitions(devilution PUBLIC SPDLOG_COMPILED_LIB)
+target_compile_definitions(${BIN_TARGET} PUBLIC SPDLOG_COMPILED_LIB)
+
+if(TRACE_LOGGING)
+target_compile_definitions(devilution PUBLIC TRACE_LOGGING)
+target_compile_definitions(${BIN_TARGET} PUBLIC TRACE_LOGGING)
+endif()
+
+if(${CMAKE_VERSION} VERSION_GREATER "3.16.0")
```
glebm, line 460:
```suggestion
if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16") 
```

CMakeLists.txt
```diff
@@ -435,21 +449,37 @@ if(GPERF)
   find_package(Gperftools REQUIRED)
 endif()
 
-target_include_directories(devilution PUBLIC Source SourceS ${CMAKE_CURRENT_BINARY_DIR})
+target_compile_definitions(devilution PUBLIC SPDLOG_COMPILED_LIB)
+target_compile_definitions(${BIN_TARGET} PUBLIC SPDLOG_COMPILED_LIB)
+
+if(TRACE_LOGGING)
```
glebm, line 455: Instead of doing this here, simply add `TRACE_LOGGING` to the `# Defines without value` list below

Source/appfat.cpp
```diff
@@ -11,43 +11,43 @@ namespace devilution {
 
 namespace {
 
-/** Set to true when a fatal error is encountered and the application should shut down. */
-bool terminating = false;
-/** Thread id of the last callee to FreeDlg(). */
-SDL_threadID cleanup_thread_id;
+	/** Set to true when a fatal error is encountered and the application should shut down. */
```
glebm, line 14: These formatting changes come from nested `.clang-format` files that @AJenbo hasn't had a chance to delete yet.

Please revert all the formatting changes in the PR (in general formatting changes should not be mixed with code changes -- for ease of review and better commit history).

glebm (changes requested):

Jmgr:
> Since SDL already has build-in logging functionality I would prefer that we make use of that. It doesn't seam that this PR integrates with that functionality?

But that's C and completely unsafe. We talked a bit about it on Discord and you seemed to be motivated to modernize the logging so I don't understand why the change of position on this. Why use C tools in a C++ project when you could use type-safe C++ tools instead?

Source/appfat.cpp
Jmgr: I think it would be great to solve that issue. It would be good if making a code change doesn't break the formatting because the file was not correctly formatted before.

glebm:
@Jmgr I think, importantly, this new logging does not have full platform support like SDL logging does:

For example:

https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/src/SDL_log.c#L425-L431

glebm:
SDL also already defines categories: https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/include/SDL_log.h#L66-L74

and logging levels: https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/include/SDL_log.h#L104-L110

> But that's C and completely unsafe.

It is not unsafe these days, modern compilers check arguments to printf and SDL_Log (via this attribute:  https://github.com/libsdl-org/SDL/blob/a4ddb175f1f1d832960c830191daaab7eb25638f/include/SDL_stdinc.h#L312)

Jmgr:
> @Jmgr I think, importantly, this new logging does not have full platform support like `SDL_LogOutput` does:
> 
> For example:
> 
> [libsdl-org/SDL@`c59d4dc`/src/SDL_log.c#L425-L431](https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/src/SDL_log.c#L425-L431)

For that particular example you can just create a file sink on that platform, so that's not a big issue.

Other platforms  might need a few changes, maybe with a PR upstream. But I get your point.

glebm:
> For that particular example you can just create a file sink on that platform, so that's not a big issue.

@Jmgr We support many platforms, they ship with custom SDL versions that integrate this support correctly. Moving away from that and having to support each platform ourselves is a big deal.

Jmgr:
> > For that particular example you can just create a file sink on that platform, so that's not a big issue.
> 
> @Jmgr We support many platforms, some ship with custom SDL versions that integrate this support correctly (e.g. Switch, 3DS). Moving away from that and having to support each platform ourselves is a big deal. It also makes DevilutionX harder to port to new platforms.

Yeah I understand that.

Jmgr:
> SDL also already defines categories: [libsdl-org/SDL@`c59d4dc`/include/SDL_log.h#L66-L74](https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/include/SDL_log.h#L66-L74)
> 
> and logging levels: [libsdl-org/SDL@`c59d4dc`/include/SDL_log.h#L104-L110](https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/include/SDL_log.h#L104-L110)
> 
> > But that's C and completely unsafe.
> 
> It is not unsafe these days, modern compilers check arguments to printf and SDL_Log (via this attribute: [libsdl-org/SDL@`a4ddb17`/include/SDL_stdinc.h#L312](https://github.com/libsdl-org/SDL/blob/a4ddb175f1f1d832960c830191daaab7eb25638f/include/SDL_stdinc.h#L312))

Sure, but you still have to mess with formatting and types. That's quite messy. What if you want to log a std::string? Or and std::vector of std::string? It's not extensible either. Would it be possible to at least use [fmt](https://github.com/fmtlib/fmt)?

glebm:
@Jmgr IIRC the `fmt` package requires C++17, we're still on C++11

Jmgr:
> @Jmgr IIRC the `fmt` package requires C++17, we're still on C++11

Where do you see that? On https://fmt.dev/latest/index.html it says `The library is highly portable and relies only on a small set of C++11 features`.

glebm:
Ah that's good then. Though I'm not sure it's worth the benefit of bringing in another dependency tbh for the type of logging that we do.

glebm:
> Where do you see that? 

I grepped the code for `string_view` :)

Jmgr:
> Ah that's good then. Though I'm not sure it's worth the benefit of bringing in another dependency tbh for the type of logging that we do.

That would not be only for logging, every formatting would be C++ instead of C. It would also simplify translating texts. And remove all code like `std::string("my favorite number is ") + std::to_string(42)`. And it's just three headers.

Jmgr:
> > Where do you see that?
> 
> I grepped the code for `string_view` :)

That's only because that part is a C++17 feature :)

glebm:
> That would not be only for logging

That's a good point, I think that does make it more worthwhile. Let's see what @AJenbo thinks.

galaxyhaxz:
> But that's C and completely unsafe

Well, if you're not smart enough to write good code and use a language safely, then no amount of fail-safes put in place for another language would save you...

Jmgr:
> > But that's C and completely unsafe
> 
> Well, if you're not smart enough to write good code and use a language safely, then no amount of fail-safes put in place for another language would save you...

The question is not to be smart or not, but to accept that we, as humans, are fallible. Using tools that allow a computer (in that case, a compiler) to correct us when we make a mistake can be of great help. Why deny that? That's one of the reasons programming languages have been invented, otherwise we would all write in assembly.

galaxyhaxz:
But it's been mentioned already, many modern compilers already double check variable arg functions. Is SDL_Log really unsafe enough to justify adding a whole new library what is essentially only used for temporary debugging?

Jmgr:
> But it's been mentioned already, many modern compilers already double check variable arg functions. Is SDL_Log really unsafe enough to justify adding a whole new library that is essentially only used for temporary debugging?

As I wrote before, using `fmt` would benefit all places where formatting is done, so logging, but also translated texts (would you prefer giving translators a text like `Hello %s, this item costs %d gold pieces.` or `Hello {}, this item costs {} gold pieces.`?), and basically every place where std::to_string or sprintf is used. Also, if you program in C++ and use std::string then why not use a tool that allow you insert it directly using a format?

Source/appfat.cpp
AJenbo: Very close to doing so, I think there is only one file left to deal with.

AJenbo:
> We talked a bit about it on Discord and you seemed to be motivated to modernize the logging so I don't understand why the change of position on this.

I might have glossed over a bit of the detail at the time, sorry for miss communicating that. Things are a little hectic around here atm as you can see there are lots of things people would like done sooner rather then later :D

galaxyhaxz:
> would you prefer giving translators a text like Hello %s, this item costs %d gold pieces. or Hello {}, this item costs {} gold pieces.?)

In some languages the order of things could be reversed and I find the former to be more clear as I can directly see what each type is _supposed_ to be. Regardless, just wanted to toss that out there and won't argue any further. I've never been a fan of C++, but that is not the subject at hand here and I won't derail the thread.

AJenbo:
> would you prefer giving translators a text like `Hello %s, this item costs %d gold pieces.` or `Hello {}, this item costs {} gold pieces.`?)

The first one. It indicates what content will probably be inserted into the text, and they are probably also already used to it from other translation projects. I don't remember seeing `{}` back when I worked on translations. `Hello <name>, this item costs <price> gold pieces.` would be even better

Jmgr:
> > We talked a bit about it on Discord and you seemed to be motivated to modernize the logging so I don't understand why the change of position on this.
> 
> I might have glossed over a bit of the detail at the time, sorry for miss communicating that. Things are a little hectic around here atm as you can see there are lots of things people would like done sooner rather then later :D

Yeah I can understand the situation, no worries.

AJenbo:
We had tools that would complain if we accidentally phrased a sentence in a way like `That will be %d gold for this item. Sir %s`. So that would also be an issue with the generic `{}` style.

Jmgr:
> > would you prefer giving translators a text like `Hello %s, this item costs %d gold pieces.` or `Hello {}, this item costs {} gold pieces.`?)
> 
> The first one. It indicates what content will probably be inserted into the text, and they are likely also already used to it from other translation projects. I don't remember seeing `{}` back when I worked on translations. `Hello <name>, this item costs <price> gold pieces.` would be even better

Fair enough, but <name> and such would require a lot more work, and the other options is already available.

But what about the other places `fmt` could be used?

Jmgr:
> We had tools that would complain if we accidentally phrased a sentence in a way like `That will be %d gold for this item. Sir %s`. So that would also be an issue with the generic `{}` style.

What would the tools say in that case?

AJenbo:
`bad msgstr pattern "%s"`

Jmgr:
> `bad msgstr pattern "%s"`. It can sometimes be really tedious to phrase a text so that the variables appear in the same order as in English :) Some languages that use gender and other oddities can be downright impossible as the need to know the value to use the correct gender. Polish is a good example here where I think "shield" and "sword" should have different translations. Don't even know how to solve for that :D

Yeah translation is more complicated than it looks. `fmt` could [help](https://fmt.dev/latest/syntax.html) with the order issue: you can use `That will be {0} gold for this item. Sir {1}` to reference the first and second parameter, and swap them in a translation.

glebm:
clang-tidy-module to automatically convert all `printf` calls to `fmt`:
https://github.com/mikecrowe/clang-tidy-fmt

AJenbo:
Going to close this as we seam to agree that this will need a different implementation in order to work with our goals.

PR #1465, opened 2021-04-10
`world_draw_black_tile`: Stop relying on buffer padding

glebm:
No performance difference (measured by rendering a screen full of black tiles at 640x480)

### DIFF
--- a/Source/render.cpp
+++ b/Source/render.cpp
@@ -586,33 +586,134 @@ void RenderTile(CelOutputBuffer out, int x, int y)
 	}
 }
 
+namespace {
+
+// A tile is a diamond shape consisting of 2 triangles, upper and lower.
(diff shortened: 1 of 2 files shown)


AJenbo (approved):
Looks fine, but wondering if we could just replace this with SDL_FillRect, it's not diamond shaped, it might not be as optimized, but it would be way simpler and there aren't that many of them anyway. I'm not sure weather they are drawn before the tiles that go on top, if that isn't the case then that might not work.

AJenbo:
I'm almost thinking that including a black tile sprite in devilutionx.mpq is a viable alternative.

glebm:
I think the current approach can be used as a basis for converting the rest of render.cpp, so let's go with it for now

PR #1469, opened 2021-04-11
'plr_class' enum cleanup

julealgon:
This PR converts the plr_class enum to a scoped enumeration and renames both the enum itself as well as its members.

Adjustments are made to ensure every place accessing the enum for indexing into arrays and converting into other values are handled appropriately.

### DIFF
--- a/Source/pack.cpp
+++ b/Source/pack.cpp
@@ -217,7 +217,7 @@ void UnPackPlayer(PkPlayerStruct *pPack, int pnum, bool netSync)
 	ClrPlrPath(pnum);
 	pPlayer->destAction = ACTION_NONE;
 	strcpy(pPlayer->_pName, pPack->pName);
-	pPlayer->_pClass = (plr_class)pPack->pClass;
+	pPlayer->_pClass = (HeroClass)pPack->pClass;
 	InitPlayer(pnum, true);
 	pPlayer->_pBaseStr = pPack->pBaseStr;
 	pPlayer->_pStrength = pPack->pBaseStr;
--- a/Source/pack.cpp
+++ b/Source/pack.cpp
@@ -61,7 +61,7 @@ void PackPlayer(PkPlayerStruct *pPack, int pnum, bool manashield)
 	pPack->targx = pPlayer->_ptargx;
 	pPack->targy = pPlayer->_ptargy;
 	strcpy(pPack->pName, pPlayer->_pName);
-	pPack->pClass = pPlayer->_pClass;
+	pPack->pClass = static_cast<Sint8>(pPlayer->_pClass);
 	pPack->pBaseStr = pPlayer->_pBaseStr;
 	pPack->pBaseMag = pPlayer->_pBaseMag;
 	pPack->pBaseDex = pPlayer->_pBaseDex;
@@ -233,7 +233,7 @@ void UnPackPlayer(PkPlayerStruct *pPack, int pnum, bool netSync)
 	pPlayer->_pGold = SwapLE32(pPack->pGold);
 	pPlayer->_pMaxHPBase = SwapLE32(pPack->pMaxHPBase);
 	pPlayer->_pHPBase = SwapLE32(pPack->pHPBase);
-	pPlayer->_pBaseToBlk = ToBlkTbl[pPlayer->_pClass];
+	pPlayer->_pBaseToBlk = ToBlkTbl[static_cast<std::size_t>(pPlayer->_pClass)];
 	if (!netSync)
 		if ((int)(pPlayer->_pHPBase & 0xFFFFFFC0) < 64)
 			pPlayer->_pHPBase = 64;
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -42,7 +42,7 @@ int plrxoff2[9] = { 0, 1, 0, 1, 2, 0, 1, 2, 2 };
 /** Specifies the Y-coordinate delta from a player, used for instanced when casting resurrect. */
 int plryoff2[9] = { 0, 0, 1, 1, 0, 2, 2, 1, 2 };
 /** Specifies the frame of each animation for which an action is triggered, for each player class. */
-char PlrGFXAnimLens[NUM_CLASSES][11] = {
+char PlrGFXAnimLens[static_cast<std::size_t>(plr_class::NUM_CLASSES)][11] = {
 	{ 10, 16, 8, 2, 20, 20, 6, 20, 8, 9, 14 },
 	{ 8, 18, 8, 4, 20, 16, 7, 20, 8, 10, 12 },
 	{ 8, 16, 8, 6, 20, 12, 8, 20, 8, 12, 8 },
@@ -51,7 +51,7 @@ char PlrGFXAnimLens[NUM_CLASSES][11] = {
 	{ 10, 16, 8, 2, 20, 20, 6, 20, 8, 9, 14 },
 };
 /** Maps from player class to player velocity. */
-int PWVel[NUM_CLASSES][3] = {
+int PWVel[static_cast<std::size_t>(plr_class::NUM_CLASSES)][3] = {
 	{ 2048, 1024, 512 },
 	{ 2048, 1024, 512 },
 	{ 2048, 1024, 512 },
@@ -60,7 +60,7 @@ int PWVel[NUM_CLASSES][3] = {
 	{ 2048, 1024, 512 },
 };
 /** Total number of frames in walk animation. */
-int AnimLenFromClass[NUM_CLASSES] = {
+int AnimLenFromClass[static_cast<std::size_t>(plr_class::NUM_CLASSES)] = {
 	8,
 	8,
 	8,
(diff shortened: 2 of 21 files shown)


AJenbo (dismissed):

Source/player.h
```diff
@@ -29,14 +29,14 @@ enum {
...
-	PC_BARBARIAN,
-	NUM_CLASSES
+enum class HeroClass : uint8_t {
+	Warrior,
+	Rogue,
+	Sorcerer,
+	Monk,
+	Bard,
+	Barbarian,
+	ClassCount
```
glebm, line 39: This doesn't have to happen in this PR, but it'd be nice if we moved the `ClassCount` member out of the num itself, e.g. into a `constexpr std::size_t kHeroClassCount = 7`

Having it within the enum means we can't benefit from the compiler's `switch` exhaustiveness check in switch statements.

One way to avoid having to manually specify the number is to define a `MAX = Barbarian` member and do `KHeroClassCount = HeroClass::MAX + 1`

glebm (approved):

AJenbo:
You forgot to update the tests :)

Source/player.h
julealgon: The external variable looks terrible, but the extra value inside the enum looks even worse so I agree. I actually thought about it but decided not to do it as part of this PR.

 > One way to avoid having to manually specify the number is to define a MAX = Barbarian member and do KHeroClassCount = HeroClass::MAX + 1

Adding a "MAX" entry, even if it is an alias, is also pretty bad IMHO. Suddenly, consumers could start using `MAX` instead of `Barbarian` which makes zero sense.

What about just defining it as:
```cpp
 kHeroClassCount = HeroClass::Barbarian + 1
```

BTW, what does that capital "K" in the name?

julealgon:
> You forgot to update the tests :)

How weird... I used VS's refactoring to rename it. For some reason it didn't find the ones in the test? Let me check it.

julealgon:
Tests fixed now. I have no idea why they are completely ignored by Visual Studio (both on the build step, as well as through the refactoring steps)... that's far from ideal.

Source/player.h
AJenbo: this best live right next to the enum then so that another hero is added people will notice that this should be changed to  `kHeroClassCount = HeroClass::Newhero + 1`.

may i instead suggest that we get the size by counting the length of the arrays thathold the hero names, or similar arrays? https://github.com/diasurgical/devilutionX/blob/master/Source/player.cpp#L173

AJenbo:
i havent figured out how to setup the tests in MSVC if you do feel free to comit the needed changes to the msvc json file.

the testswhere recently moved maybe that is why it missed them

SourceX/DiabloUI/diabloui.cpp
```diff
@@ -454,19 +454,19 @@ void LoadHeros()
 	LoadArt("ui_art\\heros.pcx", &ArtHero);
 
 	const int portraitHeight = 76;
-	int portraitOrder[NUM_CLASSES + 1] = { 0, 1, 2, 2, 1, 0, 3 };
+	int portraitOrder[static_cast<std::size_t>(HeroClass::ClassCount) + 1] = { 0, 1, 2, 2, 1, 0, 3 };
```
AJenbo, line 457: How come this now requires us to cast the enum when using them as indexes?

glebm: `enum class` is not implicitly convertible to integral types

AJenbo: sad, can we make it so :)

Jmgr: That's a feature, it's to prevent accidental conversions to/from int.

Source/player.h
Jmgr: > BTW, what does that capital "K" in the name?

I think some people prefix their constants with k. Like in "konstants". Because c was already taken to prefix classes. Hungarian notation...

SourceX/DiabloUI/diabloui.cpp
julealgon: It's one of the unfortunate things about indexing with them into a raw array. The `std::map` avoids this but... well... we already discussed that option 😛

AJenbo: > That's a feature, it's to prevent accidental conversions to/from int.

Wait but that is exactly what we use them for :D
Originally they were all written as `#define ATTRIB_STR 0` :D

Source/player.h
AJenbo: > Hungarian notation...

Pretty sure "konstant" is the danish word for it :thinking:

julealgon:
Rebased to remove conflicts.

SourceX/DiabloUI/diabloui.cpp
julealgon: I think the core here is "accidental". When you add the `static_cast` you are being explicit about the intention of using it as a number.

AJenbo:
```cpp
enum : permissive <with::count> {
}
```

Source/player.h
julealgon: > this best live right next to the enum then so that another hero is added people will notice that this should be changed to kHeroClassCount = HeroClass::Newhero + 1.
> 
> may i instead suggest that we get the size by counting the length of the arrays thathold the hero names, or similar arrays? https://github.com/diasurgical/devilutionX/blob/master/Source/player.cpp#L173

Hadn't seen this comment yet. As discussed, proposal for a more coupled approach (more coupled being better, in this case) is here : https://github.com/diasurgical/devilutionX/pull/1474/commits

SourceX/DiabloUI/diabloui.cpp
julealgon: What about `std::unordere...`... ah, forget it 😆

Source/player.h
AJenbo: I guess that's way you where confused about it in the chat :)

SourceX/DiabloUI/diabloui.cpp
Jmgr: Where is the discussion about using a map?

julealgon: Discord

Source/pack.cpp
```diff
@@ -61,7 +61,7 @@ void PackPlayer(PkPlayerStruct *pPack, int pnum, bool manashield)
 	pPack->targx = pPlayer->_ptargx;
 	pPack->targy = pPlayer->_ptargy;
 	strcpy(pPack->pName, pPlayer->_pName);
-	pPack->pClass = pPlayer->_pClass;
+	pPack->pClass = static_cast<Sint8>(pPlayer->_pClass);
```
AJenbo, line 64: isn't this unsigned?

Source/player.cpp
```diff
@@ -624,9 +624,9 @@ void SetPlrAnims(int pnum)
 }
 
 /**
- * @param c plr_classes value
+ * @param c the hero class.
```
AJenbo, line 627:
```suggestion
 * @param c The hero class.

```

Source/pack.cpp
```diff
@@ -233,7 +233,7 @@ void UnPackPlayer(PkPlayerStruct *pPack, int pnum, bool netSync)
 	pPlayer->_pGold = SwapLE32(pPack->pGold);
 	pPlayer->_pMaxHPBase = SwapLE32(pPack->pMaxHPBase);
 	pPlayer->_pHPBase = SwapLE32(pPack->pHPBase);
-	pPlayer->_pBaseToBlk = ToBlkTbl[pPlayer->_pClass];
+	pPlayer->_pBaseToBlk = ToBlkTbl[static_cast<std::size_t>(pPlayer->_pClass)];
```
AJenbo, line 236: Would it make sens for us to move some of this to functions or switches to avoid all the casting?

julealgon: It certainly would, however i consider that separate refactoring. Do you want to do it now, before the enum change?

Source/pack.cpp
julealgon: No:

https://github.com/diasurgical/devilutionX/blob/ced74f76f92bb00064c5c4ffc813fcca31715a70/Source/pack.h#L39

Source/player.cpp
julealgon: Fixed

Source/pack.cpp
AJenbo: Oh, well guess it doesn't matter if we support 127 classes or 255 :)

Source/pack.cpp
AJenbo: I do to, but if you didn't agree it would make me hesitant to add the casts in the first place :)

AJenbo (approved):
Now we just need to rebase and get it building :)

Source/pack.cpp
julealgon: I believe most of these table-like fetching will become functions later, similar to what I did earlier for max stats.

PR #1471, opened 2021-04-11
:recycle: Refactor talk to towners

AJenbo:


julealgon:
Suggest a slight commit rewording:

From: 
- `Re factor talkt to towners`

To: 
- `Refactor talk to towners`

AJenbo:
Out of brainpower, too many events happened to night, good night.

Source/monster.cpp
```diff
@@ -5603,15 +5602,17 @@ void TalktoMonster(int i)
 	pnum = Monst->_menemy;
 	Monst->_mmode = MM_TALK;
 	if (Monst->_mAi == AI_SNOTSPIL || Monst->_mAi == AI_LACHDAN) {
-		if (QuestStatus(Q_LTBANNER) && quests[Q_LTBANNER]._qvar1 == 2 && PlrHasItem(pnum, IDI_BANNER, &itm)) {
-			RemoveInvItem(pnum, itm);
+		itm = plr[pnum].GetItemIndex(IDI_BANNER);
```
julealgon, line 5605: I'd prefer to keep this in the condition so that it is short-circuited and not retrieved when not needed.

It obviously requires a change in interface. See comment above (perhaps rename it to `bool TryGetItemIndex(ref itemIndex)`)

Source/monster.cpp
```diff
@@ -5603,15 +5602,17 @@ void TalktoMonster(int i)
...
+		if (QuestStatus(Q_LTBANNER) && quests[Q_LTBANNER]._qvar1 == 2 && itm != -1) {
+			RemoveInvItem(&plr[pnum], itm);
 			quests[Q_LTBANNER]._qactive = QUEST_DONE;
 			Monst->mtalkmsg = TEXT_BANNER12;
 			Monst->_mgoal = MGOAL_INQUIRING;
 		}
 		if (QuestStatus(Q_VEIL) && Monst->mtalkmsg >= TEXT_VEIL9) {
-			if (PlrHasItem(pnum, IDI_GLDNELIX, &itm)) {
-				RemoveInvItem(pnum, itm);
+			itm = plr[pnum].GetItemIndex(IDI_GLDNELIX);
```
julealgon, line 5613: This is very similar to the logic above, however we are doing it differently.

Would suggest doing the same checks the same way for consistency and readability (i.e. combine the `if`s and use the item as part of the condition).

Source/objects.cpp
```diff
@@ -3318,15 +3318,13 @@ void OperateL3Door(int pnum, int i, bool sendflag)
...
-	BYTE *mem;
-	int iv;
-
 	if (numitems >= MAXITEMS) {
 		return;
 	}
 
-	if (object[i]._oVar6 != 3 && PlrHasItem(pnum, IDI_BLDSTONE, &iv) != NULL) {
-		RemoveInvItem(pnum, iv);
+	int iv = plr[pnum].GetItemIndex(IDI_BLDSTONE);
```
julealgon, line 3325: Same as the other one. This does not short-circuit and will retrieve the item when not needed.

Source/player.cpp
```diff
@@ -179,6 +179,16 @@ const char *const ClassPathTbl[] = {
 	"Warrior",
 };
 
+int PlayerStruct::GetItemIndex(int idx) const
```
julealgon, line 182:
```suggestion
bool PlayerStruct::TryGetItem(int idx, ItemStruct* item) const
```

Source/player.h
```diff
@@ -305,18 +305,25 @@ struct PlayerStruct {
...
+	 * @return The inventory index of the item, or -1 if not found.
+	 */
+	int GetItemIndex(int idx) const;
+
 	/**
 	 * @brief Gets the base value of the player's specified attribute.
 	 * @param attribute The attribute to retrieve the base value for
 	 * @return The base value for the requested attribute.
-	*/
+	 */
```
julealgon, line 319: I believe this reformatting is incorrect. Standard doxygen has it in the beggining, like this:

image

Source/towners.cpp
```diff
@@ -136,7 +136,7 @@ namespace {
 /** Specifies the active sound effect ID for interacting with cows. */
 _sfx_id CowPlaying = SFX_NONE;
 
-void CowSFX(int pnum)
+void CowSFX(PlayerStruct *player)
```
julealgon, line 139:
```suggestion
void CowSFX(const PlayerStruct &player)
```

Source/towners.cpp
```diff
@@ -503,18 +503,15 @@ void FreeTownerGFX()
...
-
 	if (i == -1)
 		return;
 
-	if (towner[i]._tbtcnt) {
-		p = towner[i]._tTalkingToPlayer;
-		dx = abs(towner[i]._tx - plr[p]._px);
-		dy = abs(towner[i]._ty - plr[p]._py);
+	if (towners[i]._tbtcnt) {
+		PlayerStruct *player = towners[i]._tTalkingToPlayer;
```
julealgon, line 510:
```suggestion
		const PlayerStruct &player = towners[i]._tTalkingToPlayer;
```

Source/towners.cpp
```diff
@@ -683,624 +680,693 @@ void ProcessTowners()
...
+void TownerTalk(int first)
 {
 	sgdwCowClicks = 0;
 	sgnCowMsg = 0;
 	storeflag = true;
 	InitQTextMsg(first);
 }
 
-void TalkToTowner(int p, int t)
+void TalkToTavern(PlayerStruct *player, TownerStruct *towner)
```
julealgon, line 714:
```suggestion
void TalkToTavern(const PlayerStruct &player, TownerStruct *towner)
```

Source/towners.cpp
```diff
@@ -683,624 +680,693 @@ void ProcessTowners()
...
+				InitQTextMsg(TEXT_BANNER3);
+				towner->_tMsgSaid = true;
 			}
 		}
-		if (!qtextflag) {
-			TownerTalk(TEXT_GRISWOLD1, t);
-			if (storeflag) {
-				StartStore(STORE_SMITH);
-			}
+	}
```
julealgon, line 774: Section above this is really hard to review right now.

Source/towners.cpp
```diff
@@ -683,624 +680,693 @@ void ProcessTowners()
...
+	if (qtextflag)
+		return;
+	TownerTalk(TEXT_OGDEN1);
+	if (storeflag) {
+		StartStore(STORE_TAVERN);
+	}
+}
+
+void TalkToDeadguy(PlayerStruct *player, TownerStruct *towner)
```
julealgon, line 783:
```suggestion
void TalkToDeadguy(coonst PlayerStruct &player, TownerStruct *towner)
```

julealgon, line 783: Seems weird to me to have the towner name in the function name, yet take a `TownerStruct` as parameter. What if someone calls this method with a towner other than the deadguy? 

I think this is a design smell.

Source/towners.cpp
```diff
@@ -683,624 +680,693 @@ void ProcessTowners()
...
+		return;
+	TownerTalk(TEXT_STORY1);
+	if (storeflag) {
+		StartStore(STORE_STORY);
+	}
+}
+
+void TalkToCow(PlayerStruct *player)
+{
+}
```
julealgon, line 1072: Why have an empty function?

Source/towners.h
```diff
@@ -52,17 +52,16 @@ struct TownerStruct {
 	bool _tSelFlag;
 	bool _tMsgSaid;
 	int8_t _tAnimOrder;
-	int8_t _tTalkingToPlayer;
+	PlayerStruct *_tTalkingToPlayer;
```
julealgon, line 55: Awesome change here.

julealgon (changes requested):
The main refactor is too big for me to reason about it. Would strongly suggest thinking about ways to split the work otherwise I won't be able to approve this (and you'll be on your own).

AJenbo:
Yeah it's still in draft mode, i ran out of juice before i could finish it, just made i compile and pushed so i could go to bed 📦 

Still nice with some erarly feedback for when i wake up.

Source/towners.cpp
AJenbo: Will delete

Source/towners.cpp
AJenbo: It's a type, there could potentially be multiple dead guy towners. We could mabe handle this as a towner-ai, to mirror monsters even more.

Source/inv.cpp
```diff
@@ -2012,15 +2010,14 @@ void CheckQuestItem(int pnum)
 	}
 	if (plr[pnum].HoldItem.IDidx == IDI_NOTE1 || plr[pnum].HoldItem.IDidx == IDI_NOTE2 || plr[pnum].HoldItem.IDidx == IDI_NOTE3) {
 		int mask, idx, item_num;
-		int n1, n2, n3;
 		ItemStruct tmp;
 		mask = 0;
 		idx = plr[pnum].HoldItem.IDidx;
-		if (PlrHasItem(pnum, IDI_NOTE1, &n1) || idx == IDI_NOTE1)
+		if (plr[pnum].GetItemIndex(IDI_NOTE1) != -1 || idx == IDI_NOTE1)
```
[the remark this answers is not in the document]
AJenbo: The return was not a bool, it was a pointer to the item or null

Source/inv.cpp
```diff
@@ -2586,8 +2577,7 @@ void RemoveScroll(int pnum)
 		if (!plr[pnum].InvList[i].isEmpty()
 		    && (plr[pnum].InvList[i]._iMiscId == IMISC_SCROLL || plr[pnum].InvList[i]._iMiscId == IMISC_SCROLLT)
 		    && plr[pnum].InvList[i]._iSpell == plr[pnum]._pRSpell) {
-			RemoveInvItem(pnum, i);
-			CalcPlrScrolls(pnum);
```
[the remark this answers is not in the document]
AJenbo: RemoveInvItem calls CalcPlrScrolls at the end

Source/items.h
```diff
@@ -366,7 +367,7 @@ bool IsUniqueAvailable(int i);
 void InitItemGFX();
 void InitItems();
 void CalcPlrItemVals(int p, bool Loadgfx);
-void CalcPlrScrolls(int p);
+void CalcPlrScrolls(struct PlayerStruct *player);
```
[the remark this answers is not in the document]
AJenbo: Gets the compiler angry since both headers depend on each other.

Source/monster.cpp
AJenbo: Maybe it would be and idea to make index optional as it's not always used. Or maybe we should really have two functions?

Source/player.h
AJenbo: No
image

Source/towners.cpp
julealgon: > there could potentially be multiple dead guy towners

This seems a bit farfetched to me considering what we currently have. Right now, looking from the outside, it would make more sense for these methods to _not_ take a `TownerStruct` argument and instead each method pick the correct towner from the global towner array.

Obviously that's _also_ less than ideal... this looks like a proper case of polimorphism to me where each towner implementation would do something different on the same method (or something like that).

Source/monster.cpp
julealgon: I proposed a `bool TryGetItem(ItemStruct *)` in another comment. It is more general purpose, it requires no " != -1" checks, and you can change the logic that relies on indexes today (remove item from inventory by index) to rely on the item reference instead).

Source/items.h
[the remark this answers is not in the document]
julealgon: Wow..... what the

Source/monster.cpp
AJenbo: im not sure how i would delete by reference 😅 set its type to empty and then scan the whole inventory for empty items?

Source/towners.cpp
AJenbo: there are 3 towners wo all have the cow type, also Wirt's id doesnt align with his type.

I dontsee an issue with my suggestion to turn this in to towner AIs?

Source/monster.cpp
julealgon: > im not sure how i would delete by reference

It's not 100% clear to me as well, but it seems more intuitive to go with something like that. I'd have to experiment a bit with it myself as well.

> set its type to empty and then scan the whole inventory for empty items?

Maybe... perhaps @glebm has a suggestion here?

Source/player.h
julealgon: This will be a problem... Visual Studio Doxygen format generates it the way I shared :S

Source/monster.cpp
glebm: std::remove works with both iterators for C++ containers and pointers for C arrays

AJenbo: problem here is setting the inventory grid to zero for where the item was located. doing a search for it probably isent the end of the world, but ill have to look at it

PR #1473, opened 2021-04-11
CMake: Replace SDL2 modules with possibly better ones

glebm:
Update to SDL2 finder modules from https://github.com/aminosbh/sdl2-cmake-modules

Yesterday I tried building in Visual Studio and the SDL2 dlls were not being copied into the build directory. Hopefully these modules resolve the issue (needs testing)


qndel:
wait, are you saying it will automatically copy SDL dlls? :O I've been doing it manually like a peon and thought it should be like this :D

glebm:
@qndel Good CMake find modules copy automatically, I'm not sure how good this one.

Please let me know if it works for you!

AJenbo:
We need to make a fork that is the definitive version of this find script :D

qndel:
image
No difference for me, I still have to put dlls into build folders manually

glebm:
Right, looks like even this CMake module doesn't implement it

For reference, this is how it should be done https://github.com/mosra/magnum/commit/52b7c26022b2e5133f84b616a4d22e9e50e90a90

AJenbo:
We do it when packaging the build asset/package, could hat be reused?

glebm:
I'm not sure; Since I do not use Windows except for when debugging Windows-specific issues, I'll give up on this one for now.

PR #1474, opened 2021-04-11
♻️ Move 'HeroClass::ClassCount' to external tracking template

julealgon:
This eliminates the problem where the compiler complains we are not handling all enumeration values by moving the "hack" counting member into somewhere else external to the enum.

### DIFF
--- a/DiabloUI/diabloui.h
+++ b/DiabloUI/diabloui.h
@@ -34,7 +34,7 @@ struct _uiheroinfo {
 	struct _uiheroinfo *next;
 	char name[16];
 	Uint16 level;
-	plr_class heroclass;
+	HeroClass heroclass;
 	Uint8 herorank;
 	Uint16 strength;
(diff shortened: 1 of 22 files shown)


julealgon:
Depends on #1469

AJenbo:
would prefer getting the size from hero names as suggested in #1469 so we dont have to do all this casting

julealgon:
@AJenbo check last commit again for the approach were we don't need to define the template specialization for each enum.

It uses a special `LAST` member in the enum.

AJenbo:
Damn, so many enum PRs going on I'm getting super confused what is what here :D

AJenbo (approved):

AJenbo:
Merged manually via https://github.com/diasurgical/devilutionX/commit/8cf972e8439d8fffd13141a58d06128a9707f857

PR #1475, opened 2021-04-11
Fixed data type of quest _qmsg in save files.

### LINKED ISSUE #1463: Crash on selecting quest in quests window
**Important information**
Windows 10 x64 / DevilutionX version 1.2.0 i386 / official build from github release page

**Describe**
Crash on selecting quest in quests window

**To Reproduce**
1. Press 'q' button to open quests window
2. Click on any quest (or press enter)
3. See crash

**Expected behavior**
Game should not crash.

**Additional context**
Savegame: [single_1.zip](https://github.com/diasurgical/devilutionX/files/6290557/single_1.zip)

StephenCWills:
This resolves #1463

FYI, the savegame attached to the issue works in the original game. This must be because the original game is reading _qmsg as a `Uint8` regardless of how DevilutionX saves it. After making this change, the same savegame works fine in DevilutionX as well.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -739,7 +739,7 @@ static void LoadQuest(LoadHelper *file, int i)
 		file->skip(2); // Alignment
 		pQuest->_qmsg = file->nextLE<Sint32>();
 	} else {
-		pQuest->_qmsg = file->nextLE<Sint8>();
+		pQuest->_qmsg = file->nextLE<Uint8>();
 	}
 	pQuest->_qvar1 = file->nextLE<Uint8>();
 	pQuest->_qvar2 = file->nextLE<Uint8>();
@@ -1690,7 +1690,7 @@ static void SaveQuest(SaveHelper *file, int i)
 		file->skip(2); // Alignment
 		file->writeLE<Sint32>(pQuest->_qmsg);
 	} else {
-		file->writeLE<Sint8>(pQuest->_qmsg);
+		file->writeLE<Uint8>(pQuest->_qmsg);
 	}
 	file->writeLE<Uint8>(pQuest->_qvar1);
 	file->writeLE<Uint8>(pQuest->_qvar2);


Trihedraf (approved):

qndel:
:O

qndel:
Omg thanks for finding, it's because _qmsg simply doesn't fit in Sint8, Even Uint8 looks wrong to me, vanilla enums go up to 266 ...

StephenCWills:
I simply assumed that `_qmsg` is `Uint8` in Diablo and was promoted to `Sint32` in Hellfire to fit more data for the additional quests. The saves work in vanilla, so if `_qmsg` is any bigger, it must encroach on the data for other variable, like `_qvar1`. I can do some more testing tomorrow.

qndel:
At this point I think it's just pure luck no enum above 255 is passed into it in vanilla :D

AJenbo:
in that case we need a bit more code here ho handle 0xFF as -1

AJenbo:
Forget that, -1 is towneres way of thinking they don't have anything to say and will never enter the save game, despite being equal to `TEXT_BOOK31`. conversly monsters use 0 which would be `TEXT_KING1` if not for some special handeling there. And as Qndel pointed out Diablos text enums goes past what can even be held by uint8 `TEXT_INTRO      = 0x102`, what an absolute disaster :D

StephenCWills:
One thing I will say is that the devilution codebase suggests the Uint8 data type is 100% correct. The original code just memcpy's the struct to the savefile, and the struct uses unsigned char. If anything, it seems the Hellfire data type could be changed to `Uint32`, although the enum values are well below the upper limit of an `Sint32`.

https://github.com/diasurgical/devilution/blob/master/structs.h#L1043

Also, after looking at the enum myself, perhaps there is no need for additional testing. The biggest value for quest text appears to be TEXT_ZHAR2, which is well under the limit of a `Uint8`. Still, even if there technically is no error, I certainly do agree that it's a disaster.

AJenbo:
> One thing I will say is that the devilution codebase suggests the Uint8 data type is 100% correct. The original code just memcpy's the struct to the savefile, and the struct uses `unsigned char`. If anything, it seems the Hellfire data type could be changed to `Uint32`, although the enum values are well below the upper limit of an `Sint32`.

All correct, but the conditions are also flexible enough for use to define it as Sint32 which allows us to define -1 as the invalid option and actually apply the type to the codebase and there by get better IDE and compiler feedback.

Issue #1472, opened 2021-04-11
Crash after modifying diablo.ini on v1.2.0


VenKamikaze:
**Important information**
ArchLinux kernel 5.11.11
Cinnamon DE
Devilutionx 1.2.0 built myself from source tag/1.2.0

**Describe**
Changed diablo.ini to enable windowed mode, and doubled width/height, turned on experience bar and enemy health bar.

I can run devilutionx and it brings up the menu, but when I now attempt to start a game in normal mode with my character, it brings up a red box with no text, and clicking the button then bombs out the game.

Prior to changing diablo.ini I was able to start a game.

**To Reproduce**
1. Modify diablo.ini to attached version.
2. Start devilutionx 1.2.0
3. Attempt to start a new single player game as any character, new or old.

**Expected behavior**
Game begins in town.

**Screenshots**
![image](https://user-images.githubusercontent.com/2378905/114289215-3bebcf80-9ab5-11eb-8821-6d96773eaf63.png)

**Additional context**
[diablo.ini.txt](https://github.com/diasurgical/devilutionX/files/6291025/diablo.ini.txt)
Note: diablo.ini already had the duplicated entries for width/height etc, so I changed both.

**Console output:**
There is a line mentioning a font that couldn't be found which might be related to the lack of text in the red box? I've included the console output as I think it could be relevant...

INFO: SFileOpenFile: Not found: ui_art\hero0.pcx
INFO: Failed to load image meta
INFO: SFileOpenFile: Not found: ui_art\hero1.pcx
INFO: Failed to load image meta
INFO: SFileOpenFile: Not found: ui_art\hero2.pcx
INFO: Failed to load image meta
INFO: SFileOpenFile: Not found: ui_art\hero3.pcx
INFO: Failed to load image meta
INFO: SFileOpenFile: Not found: ui_art\hero4.pcx
INFO: Failed to load image meta
INFO: SFileOpenFile: Not found: ui_art\hero5.pcx
INFO: Failed to load image meta
INFO: SFileOpenFile: Not found: ui_art\hero6.pcx
INFO: Failed to load image meta
INFO: Unhandled SDL event: SDL_AUDIODEVICEADDED 0
INFO: Unhandled SDL event: SDL_AUDIODEVICEADDED 0
INFO: Unhandled SDL event: SDL_KEYMAPCHANGED 0
INFO: SFileOpenFile: Not found: data\healthbox.pcx
INFO: Failed to load image meta
[...]

PR #1476, opened 2021-04-11
Copy devilutionx.mpq to the build directory when performing CMake. #1472

VenKamikaze:
As per (non) issue #1472  - devilutionx.mpq was not automatically copied across when performing 'cmake ..' in the build folder.

This just adds the copy of devilutionx.mpq when performing a cmake.


AJenbo (approved):

julealgon (approved):
Nice! Thanks for raising this. It will make development a bit more intuitive.

PR #1477, opened 2021-04-11
'attribute_id' enum cleanup

julealgon:
This PR converts the `attribute_id` enum to a scoped enumeration and renames both the enum itself as well as its members.

Adjustments are made to ensure every place accessing the enum for indexing into arrays, converting into other values and iterating over all the attributes are handled appropriately.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1505,25 +1505,25 @@ void DrawChr(CelOutputBuffer out)
 
 	col = COL_WHITE;
 	sprintf(chrstr, "%i", plr[myplr]._pBaseStr);
-	if (plr[myplr].GetMaximumAttributeValue(attribute_id::ATTRIB_STR) == plr[myplr]._pBaseStr)
+	if (plr[myplr].GetMaximumAttributeValue(CharacterAttribute::Strength) == plr[myplr]._pBaseStr)
 		col = COL_GOLD;
 	ADD_PlrStringXY(out, 95, 155, 126, chrstr, col);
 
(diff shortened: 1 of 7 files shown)


julealgon:
Depends on #1474

Source/control.cpp
```diff
@@ -1675,33 +1675,33 @@ int CapStatPointsToAdd(int remainingStatPoints, const PlayerStruct &player, attr
...
+		chrbtn[static_cast<size_t>(i)] = false;
+		if (MouseX >= ChrBtnsRect[static_cast<size_t>(i)].x
+			&& MouseX <= ChrBtnsRect[static_cast<size_t>(i)].x + ChrBtnsRect[static_cast<size_t>(i)].w
+			&& MouseY >= ChrBtnsRect[static_cast<size_t>(i)].y
+			&& MouseY <= ChrBtnsRect[static_cast<size_t>(i)].y + ChrBtnsRect[static_cast<size_t>(i)].h) {
 			PlayerStruct &player = plr[myplr];
 			int statPointsToAdd = 1;
 			if (addAllStatPoints)
-				statPointsToAdd = CapStatPointsToAdd(player._pStatPts, player, (attribute_id)i);
+				statPointsToAdd = CapStatPointsToAdd(player._pStatPts, player, (CharacterAttribute)i);
```
AJenbo, line 1690: It's `i` actually a `CharacterAttribute` type now, so this cast is now redundant?

julealgon: Yep, fixed.

Issue #1478, opened 2021-04-11
Griswold selling items over 140k in vanilla


qndel:
**Important information**
Windows, 1.2.0 + highlighting, https://github.com/diasurgical/devilutionX/commit/5f5dc1af28274c47dc6f76e8e289166b6b0d0ca9

**Describe**
Griswold sells items above 140k

**To Reproduce**
Load save, check Griswold

**Expected behavior**
Max value should be 140k like in the original

![image](https://user-images.githubusercontent.com/14297035/114296613-9af91680-9aac-11eb-93ce-fa798822f0a4.png)

https://www.reddit.com/r/Devilution/comments/mnzzoi/did_the_devilutionx_mod_increase_griswolds_gold/
Found by this guy on reddit, save also from him
He says he didn't enable hellfire, save extension is also sv = vanilla
[single_1.zip](https://github.com/diasurgical/devilutionX/files/6291361/single_1.zip)

PR #1479, opened 2021-04-11
fix vanilla premium items going over 140k

qndel:
resolves https://github.com/diasurgical/devilutionX/issues/1478
turns out wirt had the same bug so fixed it too + changed the stat ifs into max
+wirt price bug
resolves https://github.com/diasurgical/devilutionX/issues/1480

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4631,22 +4631,11 @@ static void SpawnOnePremium(int i, int plvl, int myplr)
 	bool keepgoing = false;
 	ItemStruct holditem = items[0];
 
-	int strength = plr[myplr].GetMaximumAttributeValue(ATTRIB_STR);
-	if (strength < plr[myplr]._pStrength) {
-		strength = plr[myplr]._pStrength;
-	}
+	int strength = std::max(plr[myplr].GetMaximumAttributeValue(ATTRIB_STR), plr[myplr]._pStrength);
+	int dexterity = std::max(plr[myplr].GetMaximumAttributeValue(ATTRIB_DEX), plr[myplr]._pDexterity);
+	int magic = std::max(plr[myplr].GetMaximumAttributeValue(ATTRIB_MAG), plr[myplr]._pMagic);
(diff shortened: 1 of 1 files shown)


Source/items.cpp
```diff
@@ -4664,8 +4655,10 @@ static void SpawnOnePremium(int i, int plvl, int myplr)
 		GetItemBonus(0, itype, plvl >> 1, plvl, true, !gbIsHellfire);
 
 		if (!gbIsHellfire) {
-			if (items[0]._iIvalue > 140000)
+			if (items[0]._iIvalue > 140000) {
+				keepgoing = true; // prevent breaking the do/while loop too early by failing hellfire's condition in while
```
qndel, line 4659: while this is a relatively ugly hack, I really think we should stop refactoring till we find and fix 1.2.0 issues :D

AJenbo: Are you saying that to get us to focus on looking in to the bug? We are not going to include the refactoring in 1.2.1 if that's your concern.

Source/items.cpp
```diff
@@ -4994,8 +4977,10 @@ void SpawnBoy(int lvl)
 			GetItemBonus(0, itype, lvl, 2 * lvl, true, true);
 
 			if (!gbIsHellfire) {
-				if (items[0]._iIvalue > 140000)
+				if (items[0]._iIvalue > 140000) {
```
AJenbo, line 4980: Shoudn't this be corrected also?
```suggestion
				if (items[0]._iIvalue > 90000) {
```

Source/items.cpp
qndel: oh, didn't know :D

Source/items.cpp
qndel: yes but I left it for another PR as it requires going through this function and figuring out if it matches the original :D and I wasn't feeling like it atm ;p

AJenbo (approved):

PR #1481, opened 2021-04-11
Enable compiler checks for printf-like functions

glebm:


AJenbo:
Nice, didn't know that was an option.

PR #1483, opened 2021-04-11
Fix hellfire save games corrupting the game state

AJenbo:
Fixes https://github.com/diasurgical/devilutionX/issues/1367

### DIFF
--- a/Source/pfile.cpp
+++ b/Source/pfile.cpp
@@ -151,9 +151,13 @@ static HANDLE pfile_open_save_archive(DWORD save_num)
 	return NULL;
 }
 
-static void pfile_SFileCloseArchive(HANDLE hsArchive)
+static void pfile_SFileCloseArchive(HANDLE *hsArchive)
 {
-	SFileCloseArchive(hsArchive);
+	if (*hsArchive == NULL)
+		return;
+
+	SFileCloseArchive(*hsArchive);
+	*hsArchive = NULL;
 }
 
 void pfile_write_hero()
@@ -225,6 +229,7 @@ bool pfile_ui_set_hero_infos(bool (*ui_add_hero_info)(_uiheroinfo *))
 
 				UnPackPlayer(&pkplr, 0, false);
 
+				pfile_SFileCloseArchive(&archive);
 				LoadHeroItems(&plr[0]);
 				RemoveEmptyInventory(0);
 				CalcPlrInv(0, false);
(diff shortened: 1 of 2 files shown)


Source/pfile.cpp
```diff
@@ -153,7 +153,8 @@ static HANDLE pfile_open_save_archive(DWORD save_num)
 
 static void pfile_SFileCloseArchive(HANDLE hsArchive)
 {
-	SFileCloseArchive(hsArchive);
+	if (hsArchive)
```
glebm, line 156: Why do we call this function if the archive isn't open? I think we should `assert(hsArchive != NULL)` here instead

Source/pfile.cpp
```diff
@@ -225,6 +226,8 @@ bool pfile_ui_set_hero_infos(bool (*ui_add_hero_info)(_uiheroinfo *))
 
 				UnPackPlayer(&pkplr, 0, false);
 
+				pfile_SFileCloseArchive(archive);
```
glebm, line 229: Let's change the argument to `pfile_SFileCloseArchive` to `HANDLE *` and make `pfile_SFileCloseArchive` set it to NULL instead

Source/pfile.cpp
```diff
@@ -225,6 +226,8 @@ bool pfile_ui_set_hero_infos(bool (*ui_add_hero_info)(_uiheroinfo *))
 
 				UnPackPlayer(&pkplr, 0, false);
 
+				pfile_SFileCloseArchive(archive);
+				archive = NULL;
 				LoadHeroItems(&plr[0]);
 				RemoveEmptyInventory(0);
 				CalcPlrInv(0, false);
```
glebm, line 233: Why is there another call `pfile_SFileCloseArchive` on line 238?
If it is needed, it should be removed from there and moved to a more appropriate place

AJenbo: If unpacking was a success or not.

qndel: Because LoadHeroItems opens the archive again and failed earier because it required a closed archive to open it 🤪

glebm: Instead of calling `close` unconditionally, it should only be called if `open` was called

AJenbo: I think i will save that for a later refactor. To much global stuff going on for me to feel safe about shifting stuff around for a hotfix.

Source/pfile.cpp
AJenbo: We need to be able to safely call it multiple times, or more carefully track the state.
Or I don't know how to do this atm, feel free to send a PR that does this nicer.

glebm (approved):

PR #1484, opened 2021-04-11
SDL1: Backport SDL2 logging

glebm:
1. Backports SDL2 logging.
2. Adds a `--verbose` flag.
3. As an example, adds logging priority and category to a few log calls.

### DIFF
--- a/SourceX/miniwin/misc_msg.cpp
+++ b/SourceX/miniwin/misc_msg.cpp
@@ -264,7 +264,7 @@ bool false_avail(const char *name, int value)
 {
 #ifndef __vita__
 	// Logging on Vita is slow due slow IO, so disable spamming unhandled events to log
-	SDL_Log("Unhandled SDL event: %s %d", name, value);
+	SDL_LogDebug(SDL_LOG_CATEGORY_APPLICATION, "Unhandled SDL event: %s %d", name, value);
 #endif
 	return true;
 }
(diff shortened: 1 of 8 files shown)


SourceX/miniwin/misc_msg.cpp
```diff
@@ -264,7 +264,7 @@ bool false_avail(const char *name, int value)
 {
 #ifndef __vita__
 	// Logging on Vita is slow due slow IO, so disable spamming unhandled events to log
-	SDL_Log("Unhandled SDL event: %s %d", name, value);
+	SDL_LogDebug(SDL_LOG_CATEGORY_APPLICATION, "Unhandled SDL event: %s %d", name, value);
```
AJenbo, line 267: The requirement to add categories when using the other log levels is the primary reason I haven't been making use of them.

Would it make sense for us to make some short hand functions?

`SDL_LogDebug("Unhandled SDL event: %s %d", name, value);`

glebm: We can make shorthand functions separately without the `SDL_` prefix.
This is out-of-scope for this PR, which is mainly concerned with backporting.

AJenbo: Agreed, and sounds good :+1:

AJenbo (approved):

PR #1501, opened 2021-04-12
Fix Search spell ignoring clvl of player 0

thebigMuh:
AddSearch() used > instead of >=, which lead to it ignoring the character level of player id 0 when calculating the spell duration.


AJenbo:
Wow, so this was double broken :o

Could you also fix this issue for `MI_LightningWallC` and `MI_Blodboil` as they are also effected by this very same issue.

Source/missiles.cpp
```diff
@@ -1973,7 +1973,7 @@ void AddSearch(Sint32 mi, Sint32 sx, Sint32 sy, Sint32 dx, Sint32 dy, Sint32 mid
 	missile[mi]._miVar8 = 0;
 	AutoMapShowItems = true;
 	int lvl = 2;
-	if (id > 0)
+	if (id >= 0)
```
AJenbo, line 1976:
```suggestion
	if (id > -1)

```

PR #1507, opened 2021-04-13
Tag flag enums with __attribute__((flag_enum))

glebm:
> This attribute can be added to an enumerator to signal to the compiler that it is intended to be used as a flag type. This will cause the compiler to assume that the range of the type includes all of the values that you can get by manipulating bits of the enumerator when issuing warnings.

https://clang.llvm.org/docs/AttributeReference.html#flag-enum

### DIFF
--- a/Source/effects.h
+++ b/Source/effects.h
@@ -7,6 +7,7 @@
 
 #include <stdint.h>
 
+#include "../defs.h"
 #include "sound.h"
 
 namespace devilution {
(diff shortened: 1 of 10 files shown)


AJenbo:
I feal this might be a false positive;

```
/Users/runner/work/devilutionX/devilutionX/Source/items.h:132:2: warning: enumeration value 'CF_LEVEL' is out of range of flags in enumeration type 'icreateinfo_flag' [-Wflag-enum]
        CF_LEVEL        = (1 << 6) - 1,
        ^
```

the change didnt seam to yeld anything else

AJenbo (approved):

glebm:
Closing for now, might revisit some day

PR #1509, opened 2021-04-13
Consolidate logic for Mana Shield and player damage.

### LINKED ISSUE #1111: [Vanilla bug] Barbarian rage skill
Referring to: https://diablo2.diablowiki.net/Barbarian_(Diablo_I)

The Hellfire Barbarian has a skill called Rage. This skill is a short term buff that boosts his strength, vitality, and dexterity for 12 seconds. After that time he enters a 12 second lethargy phase, when those same attributes are reduced by the same amount. Afterwards his vitality is supposed to suffer a permanent penalty, however due to a bug in Hellfire, clicking on any item in his inventory will reset the vitality loss.

The whole thing looks like this.
[output.mkv.zip](https://github.com/diasurgical/devilutionX/files/6093624/output.mkv.zip)

It doesn't matter if the belt or the inventory is clicked to restore the health.

### LINKED ISSUE #1353: Mana shield makes player unable to be stunned under certain conditions
Another trick used by speedrunners in the vanilla game. If the player has their current life at or below their clvl and they have mana shield active, hit recovery is skipped entirely whenever being hit.

StephenCWills:
This resolves #1353
This resolves #1111

Based on my testing, the hit-recovery bug does not affect PvP.

I edited a character's stats to make testing easier. Here is a save file.
[single_0.sv.zip](https://github.com/diasurgical/devilutionX/files/6301584/single_0.sv.zip)

Missles can be tested against skeleton archers on dlvl 2. Head northwest to reach the stairs.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -927,14 +927,6 @@ void CalcPlrItemVals(int p, bool Loadgfx)
 		plr[p]._pgfxnum = g;
 	}
 
-	for (i = 0; i < nummissiles; i++) {
-		mi = missileactive[i];
-		if (missile[mi]._mitype == MIS_MANASHIELD && missile[mi]._misource == p) {
-			missile[mi]._miVar1 = plr[p]._pHitPoints;
-			missile[mi]._miVar2 = plr[p]._pHPBase;
-			break;
-		}
-	}
 	if (plr[p].InvBody[INVLOC_AMULET].isEmpty() || plr[p].InvBody[INVLOC_AMULET].IDidx != IDI_AURIC) {
 		int half = MaxGold;
 		MaxGold = GOLD_MAX_LIMIT;
--- a/Source/misdat.cpp
+++ b/Source/misdat.cpp
@@ -26,7 +26,7 @@ MissileData missiledata[] = {
 	{  &AddTown,                   &MI_Town,           MIS_TOWN,          true,      1, MISR_MAGIC,     MFILE_PORTAL,    LS_SENTINEL, LS_ELEMENTL },
 	{  &AddFlash,                  &MI_Flash,          MIS_FLASH,         true,      1, MISR_MAGIC,     MFILE_BLUEXFR,   LS_NOVA,     LS_ELECIMP1 },
 	{  &AddFlash2,                 &MI_Flash2,         MIS_FLASH2,        true,      1, MISR_MAGIC,     MFILE_BLUEXBK,   SFX_NONE,    SFX_NONE    },
-	{  &AddManashield,             &MI_SetManashield,  MIS_MANASHIELD,    false,     1, MISR_MAGIC,     MFILE_MANASHLD,  LS_MSHIELD,  SFX_NONE    },
+	{  &AddManashield,             &MI_Manashield,     MIS_MANASHIELD,    false,     1, MISR_MAGIC,     MFILE_MANASHLD,  LS_MSHIELD,  SFX_NONE    },
 	{  &AddFiremove,               &MI_Firemove,       MIS_FIREMOVE,      true,      1, MISR_FIRE,      MFILE_FIREWAL,   SFX_NONE,    SFX_NONE    },
 	{  &AddChain,                  &MI_Chain,          MIS_CHAIN,         true,      1, MISR_LIGHTNING, MFILE_LGHNING,   LS_LNING1,   LS_ELECIMP1 },
 	{  NULL,                       NULL,               MIS_SENTINAL,      true,      1, MISR_LIGHTNING, MFILE_LGHNING,   SFX_NONE,    SFX_NONE    },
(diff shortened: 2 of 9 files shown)


AJenbo:
Looking just at the diff im still not sure why tjis was happening, could you give a summery of what the cause of the error was?

NiteKat:
Looking at the diff and from what I know about the bug, it's because you lose life first with mana shield, and then mana shield refunds the life and removes mana. So if you have 5 life say, and get hit, you'll take 5 life of damage... now your life is at 0, so SyncPlrKill will be called, and StartPlrHit will NEVER be called. StartPlrHit is where the stun check is. In the end your character doesn't die though, because mana shield refunds the life back.

StephenCWills:
@NiteKat is correct. If you look at `SyncPlrKill()`, it actually checks all the missles to find a mana shield that might prevent the player's death and, if so, skips calling `StartPlayerKill()`.
https://github.com/diasurgical/devilutionX/blob/master/Source/player.cpp#L2006

If the player does happen to die in spite of the mana shield, the `MI_ManaShield()` should call `SyncPlrKill()` again after marking the mana shield for deletion. But the `MI_ManaShield()` function doesn't actually know how much damage the player received from any individual attack or whether the attack had knockback. So calling `StartPlrHit()` or handling knockback at that time isn't possible. I simply added an additional check to determine whether `StartPlayerKill()` was actually invoked before skipping `StartPlrHit()` logic.

StephenCWills:
Also potentially worth noting, this does have the effect of putting the player into the hit-recovery animation and playing the sound effect in the same frame as the player's death animation and sound effect. This does not normally happen when the player is not using Mana Shield, where the player will instantly transition to the death animation without ever entering the hit-recovery state. I considered attempting a more complex solution to fix the inconsistency, but I ultimately decided against it because it actually is consistent with how the player transitions to the death state in the PvP logic. I feel it's more likely that the behavior when not using Mana Shield is simply the result of an optimization to skip unnecessary animation state changes.

NiteKat:
I remember reading this bug got introduced by Blizz during a patch, so I think they saw the hit stun and death starting on the same frame being a "bug" and this is how they fixed it, but ended up introducing the bug of never getting hit stunned with mana shield and low life. I don't have a reference off hand for that info.

StephenCWills:
I haven't been able to find any indication on whether or when this was introduced by reviewing patch notes, Jarulf's Guide, or Google searches. On the contrary, I found at least one source that claims the bug has been around as long as Diablo 1, though it's just hearsay on a forum. There were a bunch of fixes for Mana Shield in 1.07, but nothing that stood out to me as being relevant to this bug.

Perhaps it's also worth noting that my findings in terms of the cause of this bug are not consistent with the description in Jarulf's Guide. Jarulf's Guide indicates that the Mana Shield logic puts the player back in standing mode if they were previously in dying mode, but as far as I could tell, the player never enters dying mode in the first place. I playtested this change with the attached savegame so I'm confident this change fixes the issue.

NiteKat:
I noticed the same thing when looking at the official patch notes. 1.07 had a lot of Mana Shield changes, but couldn't find diddly squat on this particular issue. 😆

AJenbo:
Wouldn't a better fix to check for a mana shield at the time of getting hit and not ever subtracting any health?

StephenCWills:
Good point. That was my first thought when I started, but I got a little intimidated when I discovered that there is no reference from the player to the existence of the Mana Shield. Having said that, `SyncPlayerKill()` already has the code to search for the Mana Shield missile associated with the player so it shouldn't be too bad. I can certainly give it a shot.

StephenCWills:
I just pushed a new commit based on comments from @AJenbo. This new version was tested against the save file attached to this pull request and also with Blood Star, Bone Spirit, and Constricting Ring. Here are some additional behavioral changes as compared to the game's original logic.

* If the player receives damage from two sources in the same frame, the 33% reduction applies to each amount individually. This can produce slightly different results due to truncation errors.
* If the player receives healing in the same frame as they receive damage, the original code would have applied both to HP, causing some or all of that damage to bypass Mana Shield. This code will apply the damage to Mana Shield and the healing to HP separately.
* If the player takes damage that exceeds their current Mana, only the damage that applies to Mana is subject to the 33% reduction. The rest of the damage is readjusted to the original amount before being applied to HP.
* The player will no longer take damage through Mana Shield when Infravision wears off or during transitions of the Rage skill.

As a bit of a side-effect, this now also resolves #1111.

qndel (approved):
tenor (3)

StephenCWills:
Having slept on it, I realize now that I forgot to remove the code at the end of `CalcPlrItemVals()` that keeps the missile variables up to date.

@qndel I see this has already been approved, but do you mind if I add a little cleanup commit to this pull request?

qndel:
> Having slept on it, I realize now that I forgot to remove the code at the end of `CalcPlrItemVals()` that keeps the missile variables up to date.
> 
> @qndel I see this has already been approved, but do you mind if I add a little cleanup commit to this pull request?

Go on :D

StephenCWills:
Done

qndel (approved):
Looks good!

qndel (changes requested):

AJenbo (approved):

PR #1517, opened 2021-04-13
Minor code cleanup in inv.cpp

FluffyQuack:
Slightly altered RemoveInvItem() so a mostly identical code snippet could be replaced with a call to that function. Also added some documentation to the function.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1682,28 +1682,7 @@ void CheckInvCut(int pnum, int mx, int my, bool automaticMove)
 			}
 
 			if (!automaticMove || automaticallyMoved) {
-				for (i = 0; i < NUM_INV_GRID_ELEM; i++) {
-					if (player.InvGrid[i] == iv || player.InvGrid[i] == -iv) {
-						player.InvGrid[i] = 0;
-					}
-				}
-
-				iv--;
-
-				player._pNumInv--;
-
-				if (player._pNumInv > 0 && player._pNumInv != iv) {
-					player.InvList[iv] = player.InvList[player._pNumInv];
-
-					for (j = 0; j < NUM_INV_GRID_ELEM; j++) {
-						if (player.InvGrid[j] == player._pNumInv + 1) {
-							player.InvGrid[j] = iv + 1;
-						}
-						if (player.InvGrid[j] == -(player._pNumInv + 1)) {
-							player.InvGrid[j] = -iv - 1;
-						}
-					}
-				}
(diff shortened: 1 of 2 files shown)


AJenbo (approved):
Bool parameters like this is general not good practice, but duplicate code is worse, so ok for now

Source/inv.cpp
```diff
@@ -1814,6 +1794,7 @@ void RemoveInvItem(int pnum, int iv)
 	iv--;
 	plr[pnum]._pNumInv--;
 
+	//If the item at the end of inventory array isn't the one we removed, we need to swap its position in the array with the removed item
```
julealgon, line 1797: This is a bad comment. It explains _what_ is being done, which someone could just look at the code to discover, instead of explaining _why_.

The why is that the list needs to be defragmented so that the N first items in it are the "valid" items that can be iterated via `pNumInv`

FluffyQuack: I personally found the "why" to be pretty obvious once I understood what the code snippet was doing, which is why I wrote it like that.

Quick rant: I think there's too much arguing about nitpicking rather than writing code. There's so much stuff in the Diablo code which can be fixed. Confusing code which could be re-written or documented. Duplicated code. Ambiguous names for global variables giving you no hint of what they actually do.

julealgon: > I personally found the "why" to be pretty obvious once I understood what the code snippet was doing, which is why I wrote it like that.

Comments are not to "explain what the code is doing" if by reading the code you see what it is doing. That's just redundant. If the code was not clear enough originally, it needs to be refactored, not have comments added to it.

> Quick rant: I think there's too much arguing about nitpicking rather than writing code.

I assume you are directing this at me specifically due to my comments. I also assume you are not used to working on a shared project with other people. You are complaining about something extremely usual IMHO.

> There's so much stuff in the Diablo code which can be fixed. Confusing code which could be re-written or documented. Duplicated code. Ambiguous names for global variables giving you no hint of what they actually do.

All those are true. What is your point?

Iniquitatis: > Quick rant: I think there's too much arguing about nitpicking rather than writing code.

Nitpicking is necessary at this stage of development. I'm pretty sure that sturdy foundation needs to be built before introducing any new big features.

AJenbo: Minor cleanups aren't really big features.

I think we can gradually improve things without it being a big problem. The comment might not be fantastic, but it's better than what wasn't there. And anyone can rephrase it later.

PR #1518, opened 2021-04-13
Current state of stash implementation

AJenbo:
Copy stashback.pcx to data/ subdirectory of wherever directory your devilutionx.mpq file is in.


thebigMuh:
Uh. This was not intended to end up in a pull request, I just pushed my changes online so people can have a look.

I already found and fixed a few issues with the version that is in here. So definitely don't consider this for actually pulling.

AJenbo:
> Uh. This was not intended to end up in a pull request, I just pushed my changes online so people can have a look.
> 
> I already found and fixed a few issues with the version that is in here. So definitely don't consider this for actually pulling.

This is for me the best way to keep track of such examples (it's a draft PR), until it's either in a mergable state or we get time to use it as a basis for an new implementation.

AJenbo:
Closing since we have an implementation of stash in master now :)

PR #1520, opened 2021-04-14
Exposes player number from multi_send_packet

yuripourre:
Exposes player number from multi_send_packet.

It also adds playerNumbers for functions:
- NetSendCmdLoc
- NetSendLoPri
- NetSendHiPri
- Movement

With this PR, is possible to move a specific player (or more than one) using the gamepad logic (after a few tweaks).

You will notice that some functions are using playerId and others are using myplr, it was done on purpose. I am trying to create a convention here, myplr is the default value for player one so if the call chain is not complete, myplr is being used. When all calls were changed to expose playerId, myplr will/should be replaced.

### DIFF
--- a/Source/trigs.cpp
+++ b/Source/trigs.cpp
@@ -847,7 +847,7 @@ void CheckTriggers()
 		switch (trigs[i]._tmsg) {
 		case WM_DIABNEXTLVL:
 			if (gbIsSpawn && currlevel >= 2) {
-				NetSendCmdLoc(true, CMD_WALKXY, plr[myplr]._px, plr[myplr]._py + 1);
+				NetSendCmdLoc(myplr, true, CMD_WALKXY, plr[myplr]._px, plr[myplr]._py + 1);
 				PlaySFX(PS_WARR18);
 				InitDiabloMsg(EMSG_NOT_IN_SHAREWARE);
 			} else {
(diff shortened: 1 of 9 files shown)


SourceX/controls/plrctrls.cpp
```diff
@@ -800,26 +800,26 @@ bool IsPathBlocked(int x, int y, int dir)
 	return !PosOkPlayer(myplr, d1x, d1y) && !PosOkPlayer(myplr, d2x, d2y);
 }
 
-void WalkInDir(AxisDirection dir)
+void WalkInDir(int playerNumber, AxisDirection dir)
```
AJenbo, line 803: i thought we agreed to call it playerId :)

Source/trigs.cpp
```diff
@@ -903,7 +903,7 @@ void CheckTriggers()
 					}
 
 					InitDiabloMsg(abortflag);
-					NetSendCmdLoc(true, CMD_WALKXY, x, y);
+					NetSendCmdLoc(myplr, true, CMD_WALKXY, x, y);
```
AJenbo, line 906: @galaxyhaxz do you know how psx dealt with this?

SourceX/controls/plrctrls.cpp
yuripourre: Oops, forgot about it. Will fix today

Source/trigs.cpp
galaxyhaxz: Since they didn't need networking, they got rid of sending packets and instead it just directly tries a packet always using `myplr` as a hardcoded id. In the `ProcessPlayers` function, it loops through twice setting `myplr` to both 0 then 1.

Issue #793, opened 2020-09-05
Scroll removal behavior is different when using as items vs using as spell


julealgon:
While working on #789, I noticed that the behavior of scrolls is different if you use them as an item (belt hotkey, or right-click in inventory) vs using them as a spell (selecting on speed book, then right clicking).

| Direct use (instant consumption) | Spell use (consume on hit) |
| --- | --- |
| ![item-use-instant-consumption](https://user-images.githubusercontent.com/461579/92314582-24dd8000-efb0-11ea-8605-d3d630fee8ef.gif) | ![spell-cast-delayed-consumption](https://user-images.githubusercontent.com/461579/92314583-27d87080-efb0-11ea-8dfc-a1241b3dfd30.gif) |
| Clicked scroll is immediately removed from inventory on the same frame as the click happens &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | A scroll of the same spell is removed from inventory/belt after the spell is cast, on the same frame that mana is reduced for normal spells (after hit-frame, on missile fired) |

[...]

Issue #792, opened 2020-09-05
Scrolls and charges not consumed correctly if swapped before hit-frame


julealgon:
When using a scroll from the spell position (speedbook -> select scroll -> right click), if the user casts the scroll spell, and then quickly changes to a different spell, the scroll is not consumed.

The same behavior happens for charges in a staff.

| Casting scroll without consuming it | Casting charge without consuming it | Casting and consuming different scroll |
| --- | --- | --- |
| ![cast-scroll-not-consumed](https://user-images.githubusercontent.com/461579/92314288-1725fb80-efac-11ea-895d-842f65ecde07.gif) | ![cast-charge-not-consumed](https://user-images.githubusercontent.com/461579/92314290-1d1bdc80-efac-11ea-9f4c-7cab52aac4bc.gif) | ![different-scroll-consumed](https://user-images.githubusercontent.com/461579/92314439-cca57e80-efad-11ea-9b35-3de593abcf13.gif) |

This occurs because scroll removal and charge deduction are done based on the spell selection at the time it executes, that is, it only removes the scroll and charges for the spell corresponding to the selection at the bottom right corner (the right button action). 
- For scrolls, the first scroll matching the spell selection is removed. Lookup searches from inventory, then from belt
- For charges, the logic checks if the currently equiped staff's spell matches the current selection, and only deducts a charge if the spell matches

All this happens inside the `UseMana` function. This function executes right after the "missile is created" (weird for Identify, but there you go... pretty much everything is a "missile" in this game) which in turn happens right at the hit-frame (the animation frame that causes the action to trigger).

I believe the obvious issue here is that the "link" used between the initial cast and the action is really weak (it can easily change between the initial frame and cast frame). Instead of relying on the current selection, it should perhaps save/transfer item information from one frame to the other, and then act on the specific item.

From my understanding, this would require significant changes. I'm not entirely sure if we can use the `pVar` values to add some data to try and help find the correct item, but that's something I thought about.

AJenbo:
pVar and friends are scary, it's basically global temp vars that may be used by any code for anything at any time. If you can understand their usage well enough to do this safely I'm not against that approach, but I at least can't provide much insight in to it.

mgpat-gm:
I'm adding the link so it's pinged in the upstream vanilla bugtracker. https://github.com/diasurgical/devilution/issues/64 ?

PR #1523, opened 2021-04-14
Fix Golem HP overflow when cast by player with negative mana.

### LINKED ISSUE #1521: Monsters don't target Golem
**Important information**
Windows 10 / Latest version (acc1578c) / Compiled using VS2019

**Describe**
Enemy monsters don't target the Golem and will instead go to great lengths to get around him and attack the player.

**To Reproduce**
1. Go to the dungeon
2. Cast a Golem
3. Follow him to an enemy monster
4. Attempt to force the enemy to attack the Golem

**Expected behavior**
The enemy attacks the Golem if it's closer. This is how the enemies in the original game behave.

### LINKED ISSUE #1522: Golem HP overflows when cast with negative mana
**Describe**
When a Golem is cast by a player with negative mana, the Golem's HP overflows to give him an astronomically high number of hitpoints.

**To Reproduce**
1. Download [single_0.zip](https://github.com/diasurgical/devilutionX/files/6307876/single_0.zip)
2. Load the savegame
3. Enter the dungeon
4. Cast the Golem from the scroll

**Expected behavior**
I suppose the game should treat the player as if they have zero mana. In this case, the Golem should spawn with 10 HP at slvl 1. However, when testing, take note of the following behavior as described by Jarulf's Guide.

> When casting a spell from a scroll or staff, it will have the same slvl as the slvl you know the spell (including any modifications from items). If you don’t know the spell it will be cast at slvl 1 (plus any modifications from items). There is a bug, however. If you cast a spell from a scroll and currently have another spell hot keyed, it will use the slvl of the hot keyed spell instead of the spell you are casting. This is true even if you don’t know the spell.

**Screenshots**
image

StephenCWills:
It seems that if a Golem spawns with 0 HP, it does not immediately die and so it's still invincible. Turns out the simplest solution was to simply ensure it has at least 1 HP.

This resolves #1521
This resolves #1522

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -5621,6 +5621,12 @@ void SpawnGolum(int i, int x, int y, int mi)
 {
 	assurance((DWORD)i < MAXMONSTERS, i);
 
+	// Set lower limits on spell level and max mana
+	// so that the Golem doesn't spawn with negative HP
+	const Sint32 minMispllvl = 1;
+	const Sint32 minPMaxMana = 0;
+	Sint32 mispllvl = std::max(missile[mi]._mispllvl, minMispllvl);
+	Sint32 pMaxMana = std::max(plr[i]._pMaxMana, minMispllvl);
 	dMonster[x][y] = i + 1;
 	monster[i]._mx = x;
 	monster[i]._my = y;
@@ -5629,12 +5635,12 @@ void SpawnGolum(int i, int x, int y, int mi)
 	monster[i]._moldx = x;
 	monster[i]._moldy = y;
 	monster[i]._pathcount = 0;
-	monster[i]._mmaxhp = 2 * (320 * missile[mi]._mispllvl + plr[i]._pMaxMana / 3);
+	monster[i]._mmaxhp = 2 * (320 * mispllvl + pMaxMana / 3);
 	monster[i]._mhitpoints = monster[i]._mmaxhp;
 	monster[i].mArmorClass = 25;
-	monster[i].mHit = 5 * (missile[mi]._mispllvl + 8) + 2 * plr[i]._pLevel;
-	monster[i].mMinDamage = 2 * (missile[mi]._mispllvl + 4);
-	monster[i].mMaxDamage = 2 * (missile[mi]._mispllvl + 8);
+	monster[i].mHit = 5 * (mispllvl + 8) + 2 * plr[i]._pLevel;
+	monster[i].mMinDamage = 2 * (mispllvl + 4);
+	monster[i].mMaxDamage = 2 * (mispllvl + 8);
 	monster[i]._mFlags |= MFLAG_GOLEM;
 	M_StartSpStand(i, 0);
 	M_Enemy(i);


Source/monster.cpp
```diff
@@ -5630,6 +5630,9 @@ void SpawnGolum(int i, int x, int y, int mi)
 	monster[i]._moldy = y;
 	monster[i]._pathcount = 0;
 	monster[i]._mmaxhp = 2 * (320 * missile[mi]._mispllvl + plr[i]._pMaxMana / 3);
```
yuripourre, line 5632: Maybe `2 * (320 * missile[mi]._mispllvl + std::min(plr[i]._pMaxMana, 0) / 3);` ?

yuripourre: If I had to guess the original intent, seems that the mana should be a kind of a bonus for the HP. I know that would change the gameplay but makes more sense in my opinion.

FitzRoyX: Yeah, just glancing the formula, the intended bare minimum life is 640ish. Casting with negative mana should simply negate the life bonus mana gives, not result in a golem with 0hp.

I guess the secondary question is why would a player ever have negative mana? Shouldn't the floor on mana pool be 0?

galaxyhaxz: It's an exploit chain, where some bugs that let you cast spells before the mana updates, putting you below zero mana, lead to this being possible :P

StephenCWills: That was my first iteration of this fix, but you'll need to add a lower limit for `_mispllvl` as well because of some other bugs that allow casting at slvl 0.
`2 * (320 * std::max(missile[mi]._mispllvl, 1) + std::max(plr[i]._pMaxMana, 0) / 3);`

This also suggests that we should apply that lower limit to all of the Golem's other stats as well.

StephenCWills: As for the exploit chain, that's not really the case here since we're dealing with max mana. However, it's possible that the user can legitimately reach negative mana by lowering their mana with shrines and then removing an item with +MAG.

qndel: yes but it's current mana, the formula has max mana

StephenCWills: Ideally, max mana would be limited to a minimum of 0 and slvl would be limited to a minimum of 1 and this code wouldn't even have a bug. I personally think that the guard for negative values is a good compromise.

AJenbo: i like you std::max solution much better, then you don't evened the guard

StephenCWills: That's fine. I can switch it over. Should we apply the slvl limit to all the Golem's stats or just HP?

AJenbo: Yeah, not just golem stuff either.

StephenCWills: My understanding is that the main reason slvl can be 0 is because of inconsistencies similar to those described in #792 and #793. It's easy to put in a guard here, but fixing all those inconsistencies is a much taller order. Should we close this pull request and focus our efforts more broadly?

AJenbo: I would rather fix a systemic issue then patch each consequence individually. But good work debugging this :)

StephenCWills: Sounds good. I will keep that in mind as I submit pull requests going forward.

PR #1524, opened 2021-04-14
Consolidated buy and sell functions of Smith, Witch and Healer

joewis:
Introducing S_Scroll, S_StartBuy, S_BuyEnter, S_BuItem function and switch between the respective Item Arrays rather than duplicating the functions for each store. Removed the duplicate functions.
Wirt is not included as his item is not stored in an array (yet).

This  approach works for sell functionality as well, I leave that for a separate commit.

### DIFF
--- a/Source/stores.cpp
+++ b/Source/stores.cpp
@@ -86,6 +86,14 @@ _speech_id gossipstart;
 /** End of possible gossip dialogs for current store */
 _speech_id gossipend;
 
+/** Pointer to the current shops item list */
+Item *storeItem;
+int PinnedItemCount;
+int storeItemCount;
+talk_id activeStore;
+talk_id buyStore;
+talk_id sellStore;
+
 /** Maps from towner IDs to NPC names. */
 const char *const TownerNames[] = {
 	N_("Griswold"),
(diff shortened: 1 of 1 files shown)


Source/stores.cpp
```diff
@@ -1242,9 +1137,15 @@ void S_SmithEnter()
 		StartStore(STORE_GOSSIP);
 		break;
 	case 12:
+        buyStore = STORE_SBUY;
+        storeItem = smithitem;
```
AJenbo, line 1141: we use tab indentation, run clang-foramt

AJenbo (dismissed):
A good first step in cleaning up this file. But please apply the correct code style.

joewis:
Submitted for sell functions. 
I'll apply clang-format at the end - it appears to cause lots of whitespace changes.

glebm:
@joewis Please only apply clang-format to change lines btw

Source/stores.cpp
```diff
@@ -362,38 +362,38 @@ bool SellOk(int i)
...
-            rv = false;
-        if (pI->_itype == ITYPE_STAFF && (!gbIsHellfire || pI->_iSpell != SPL_NULL))
-            rv = true;
-        if (pI->IDidx >= IDI_FIRSTQUEST && pI->IDidx <= IDI_LASTQUEST)
-            rv = false;
-        return rv;
-	}
-
-    return true;
+	if (pI->isEmpty())
```
glebm, line 365: Please only apply clang-format to the lines you've changed, not existing code

```bash
clang-format-diff --help
```

glebm (dismissed):

Source/stores.cpp
glebm: Nevermind, was looking at the previous diff.

joewis:
Patching Wirt into the same store logic proves more tricky than expected, of all reasons, because the item in the buy dialogue is offset by 1 item. I leave him be for now.

Source/stores.cpp
```diff
@@ -204,152 +209,150 @@ void StoreAutoPlace()
...
-	int i;
+	storenumh = 0;
+	for (int i = 0; !storeItem[i].isEmpty() && i < storeItemCount; i++)
+		storenumh++;
 
 	stextsize = true;
 	stextscrl = true;
 	stextsval = 0;
-	sprintf(tempstr, "I have these items for sale:             Your gold: %i", plr[myplr]._pGold);
+	//sprintf(tempstr, "I have these items for sale:             Your gold: %i", plr[myplr]._pGold);
```
AJenbo, line 259: disabled?

joewis: Yes...   Meant to delete this.
I've moved this into S_Smith/Witch/HealerEnter, to support Griswold's different message for premium items.

AJenbo: It would be best if you could make each commit mergeable so that we could merge this gradually.

Source/stores.cpp
```diff
@@ -204,152 +209,150 @@ void StoreAutoPlace()
...
+	AddSText(0, 16, true, "Sell items", COL_WHITE, true);
+	AddSText(0, 18, true, "Repair items", COL_WHITE, true);
+	AddSText(0, 20, true, "Leave the shop", COL_WHITE, true);
+	AddSLine(5);
+	storenumh = 20;
+	activeStore = STORE_SMITH;
 }
 
-bool SmithSellOk(int i)
+bool SellOk(int i)
```
AJenbo, line 355: This looks like it would just cause the witch to buy the items that the smith would normally buy

AJenbo: Alright I see what is going on now. Probably would be better to set a discriminator function for each store. Rather then having one function that is aware of all stores.

joewis: Yes agree... For readability. I would prefer to redo this whole refactoring.

Source/items.cpp
```diff
@@ -4575,8 +4566,6 @@ void SpawnWitch(int lvl)
 
 		item._iCreateInfo = lvl | CF_WITCH;
 		item._iIdentified = true;
-		WitchBookLevel(item);
-		item._iStatFlag = StoreStatOk(item);
```
AJenbo, line 4579: Moving this to when the vendor is accessed makes the code more generic and also doesn't require part of it to be hacked in to unpacking of the player struct.

AJenbo:
@obligaron I fixed some issues when rebasing and cleaned up some related code with how stores are refreshed. I actually think it fixes a vanilla bug where after drinking enough exiles to use an item it would still appear as red in the store, but haven't confirmed if that was the case.

I may have introduced some new issues related to the line number logic since `stextsel = ` would be `14`/`16`/`18` for different stores that visually appears the same. I don't quite understand why and I think it doesn't really matter as that part should just be discarded in the next rewrite of this code.

I'll try and also get the boy shop merged with the other vendors so that it's all combined.

Source/stores.cpp
```diff
@@ -307,191 +315,200 @@ void StoreAutoPlace()
...
 
-	storenumh = 0;
-	for (int i = 0; !smithitem[i].isEmpty(); i++) {
 		storenumh++;
 	}
+	if (storenumh == 0) {
+		return false;
+	}
+
+	AddSText(0, 1, tempstr, UiFlags::ColorWhitegold | UiFlags::AlignCenter, false);
```
obligaron, line 383: We need to assign `tempstr` here.
(for example move `sprintf(tempstr, "I have these items for sale: Your gold: %i", MyPlayer->_pGold);`)
Cause it can be override elsewhere.
And we can reenter `StartBuy` after an item is bought.

Result (see title `8`):
grafik

Source/stores.cpp
```diff
@@ -2382,7 +1973,7 @@ void StartStore(talk_id s)
 			break;
 		}
 		if (hasAnyItems)
-			StartSmithBuy();
+			StartBuy();
```
obligaron, line 1976: `StartBuy` already checks if items are available.
So could we remove `hasAnyItems` and change this to `if (!StartBuy())`?

Source/stores.cpp
```diff
@@ -2402,10 +1994,7 @@ void StartStore(talk_id s)
 		break;
 	case STORE_WBUY:
 		if (storenumh > 0)
```
obligaron, line 1996: Do we need this check at all?
Can the witch have no items to sale?

If no: remove the check
if yes: do the same as blacksmith.

Source/stores.cpp
```diff
@@ -1580,18 +1320,19 @@ void StoreSellItem()
 	}
 }
 
-void SmithSellEnter()
+void SellEnter()
 {
 	if (stextsel == BackButtonLine()) {
-		StartStore(STORE_SMITH);
+		StartStore(stextshold);
 		stextsel = 16;
```
obligaron, line 1327: Going back with the `Back` button didn't work correctly.
Example:
I tried to sell a item, but didn't have anything for the current store.
When I pressed enter, i expected tor return to the previous menu.
But instead I was asked if I want to sell a (not existing) item.

```suggestion
		StoreESC();
```

grafik

obligaron (commented):
Didn't finished reviewing yet but wanted to share the first comments.
Before this pr I didn't looked at the store logic. But now I see why the old code needs refactoring. :worried:
Good to see that someone works on this. 👍🙂

Source/stores.cpp
AJenbo: no, she has infinite mana

Source/items.cpp
joewis: I believe this is done already... ?

joewis:
Adding to the list: Cain now crashes with segfault when attempting to identify items.

Source/items.cpp
AJenbo: Correct i extracted this part in to its own PR which has been merged

PR #1528, opened 2021-04-14
Rewrote two identical code snippets in inv.cpp into one function.

FluffyQuack:
A lot of stuff can be cleaned up and rewritten in CheckInvPaste() (well, besides the rest of the Diablo code, that is), but I figured this is maybe a small step towards that. I rewrote 2 identical code snippets into one function.

I'm not happy about not rewriting how this craziness is calculated:
```
yy = 10 * (ii / 10 - ((sy - 1) >> 1));
xx = (ii % 10 - ((sx - 1) >> 1));
```

I'm not sure how to re-write it (it calculates the topleft position of an item using mouse position and item size), and also, it has to be calculated the exact same as other code earlier in the same function.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1430,7 +1430,8 @@ void CheckInvPaste(int pnum, int mx, int my)
 			}
 			ii = r - SLOTXY_INV_FIRST;
 
-			//Calculate topleft position of item for InvGrid and then add item to InvGrid
+			// Calculate top-left position of item for InvGrid and then add item to InvGrid
+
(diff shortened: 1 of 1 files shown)


Source/inv.cpp
```diff
@@ -1424,22 +1429,15 @@ void CheckInvPaste(int pnum, int mx, int my)
 				}
 			}
 			ii = r - SLOTXY_INV_FIRST;
+
+			//Calculate topleft position of item for InvGrid and then add item to InvGrid
```
AJenbo, line 1433:
```suggestion
			// Calculate top-left position of item for InvGrid and then add item to InvGrid

```

Issue #1071, opened 2021-02-28
Golem is less powerfull after loading a save game


Chance4us:
If you create a golem against diablo, then he can defeat the enemy. 
If you then first save the game during the fight, then leave it and then reload it, the golem no longer hits and can therefore no longer defeat the opponent.

How to reproduce: Just load the attached savegame and see that the golem can not defeat the opponent.
Create a new golem and see that he will defeat Diablo.

Expected behavior:
Even after relaunching of the application and reloading the saved game, the golem should hit the opponent.

[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6056646/single_0.hsv.zip)

Chance4us:
Addendum after testing. It's not a vanilla bug.

Chance4us:
Excuse me. After testing it more closely, I found that it is probably a vanilla bug after all.

If you load the following savegame, you will find that the golem never dies, that Diablo never dies either and that the player never dies either. The player does not need to restore health with potions and the game will run forever. 
I don't know if it's because the character is overpowered and can block well.
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6056716/single_0.hsv.zip)

AJenbo:
This is just poor balancing in D1. You have 130 dex and your block chances ends up as 100% against Diablo. If you try on Nightmare he should be able to hit you still as you block chance would be 89%, and just 59% on Hell.

note in Vanilla you block chance would be 25% lower after loading a save game, leaving you with just 94% block chance in this scenario. Here It would take Diablo about 5 minutes to kill you if you just stand still. But this is due to a bug in vanilla.

AJenbo:
Got confused by the multiple descriptions of issues here. Golem not dying is because Diablo isn't hitting him because he is focusing on the hero, the hero not dying is because he has a 100% block chance against Diablo.

StephenCWills:
I added some debug code and confirmed that the Golem's chance to hit goes to zero when loading the savegame.

Before load:
![image](https://user-images.githubusercontent.com/9203145/114645869-373b3d00-9ca8-11eb-804f-1dc377725056.png)

After load:
![image](https://user-images.githubusercontent.com/9203145/114645881-3dc9b480-9ca8-11eb-92f7-532b02b5ac38.png)

---

I believe I also confirmed this is not a vanilla bug by adding the same code to a debug build of devilution.

Before load:
[...]

PR #1529, opened 2021-04-14
fix loading golem

qndel:
Was it just an oversight that the values were limited to signed char while the variable is unsigned char? My golem had problems while having 133 cth :D 
Fixes https://github.com/diasurgical/devilutionX/issues/1071


AJenbo (approved):

AJenbo:
good catch

PR #1532, opened 2021-04-14
:truck: Move defines to there proper files

AJenbo:


Source/diablo.cpp
```diff
@@ -14,6 +14,13 @@
 
 namespace devilution {
 
+#ifndef DEFAULT_WIDTH
+#define DEFAULT_WIDTH 640
+#endif
+#ifndef DEFAULT_HEIGHT
+#define DEFAULT_HEIGHT 480
+#endif
```
julealgon, line 22: Shouldn't these be in `diablo.h` instead?

Also, why the `#ifndef` around them? Can we remove?

Source/encrypt.cpp
```diff
@@ -16,7 +16,7 @@ void Decrypt(DWORD *castBlock, DWORD size, DWORD key)
 
 	seed = 0xEEEEEEEE;
 	for (i = 0; i < (size >> 2); i++) {
-		DWORD t = SwapLE32(*castBlock);
+		DWORD t = SDL_SwapLE32(*castBlock);
```
julealgon, line 19: This looks unrelated to moving macros?

Source/items.h
```diff
@@ -11,6 +11,12 @@
 
 namespace devilution {
 
+#define MAXRESIST 75
```
julealgon, line 14: This should be moved to `player.h`.

Source/loadsave.cpp
```diff
@@ -1746,6 +1746,10 @@ void SaveHeroItems(PlayerStruct *pPlayer)
 	SaveItems(&file, pPlayer->SpdList, MAXBELTITEMS);
 }
 
+// 256 kilobytes + 3 bytes (demo leftover) for file magic (262147)
+// final game uses 4-byte magic instead of 3
+#define FILEBUFF ((256 * 1024) + 3)
```
julealgon, line 1751: Isn't it common to have the defines closer to the top of the file?

Source/monster.cpp
```diff
@@ -11,6 +11,12 @@
 
 namespace devilution {
 
+#define NIGHTMARE_TO_HIT_BONUS 85
+#define HELL_TO_HIT_BONUS 120
+
+#define NIGHTMARE_AC_BONUS 50
+#define HELL_AC_BONUS 80
```
julealgon, line 18: Same as the other one, I always thought constants and defines were to be placed in the header, not the cpp. Was I incorrect?

Source/objects.h
```diff
@@ -9,6 +9,8 @@
 
 namespace devilution {
 
+#define MAX_LVLS 24
```
julealgon, line 12: Isn't this level related? Why objects.h?

Source/quests.cpp
```diff
@@ -652,6 +653,8 @@ void SetReturnLvlPos()
 		ReturnLvl = quests[Q_SCHAMB]._qlevel;
 		ReturnLvlT = DTYPE_CATACOMBS;
 		break;
+	case SL_MAZE:
+		break;
```
julealgon, line 657: Unrelated

defs.h
```diff
@@ -4,30 +4,9 @@
...
-#define GMENU_ENABLED			0x80000000
-
-// must be unsigned to generate unsigned comparisons with pnum
-#define MAX_PLRS				4
-
-#define MAX_CHARACTERS			99
-#define MAX_LVLS				24
-#define MAX_LVLMTYPES			24
 #define MAX_SPELLS				52
 #define MAX_SPELL_LEVEL			15
```
julealgon, line 8: Why didn't we move some of these?

julealgon (changes requested):

Source/diablo.cpp
AJenbo: its only used in diablo.pp, some devices overwrite it because of there screen size and not having hardware capable of upscaling.

Source/encrypt.cpp
AJenbo: it removed an old dev macro that doesn't make sens in devx.

Source/loadsave.cpp
AJenbo: if its only used in a specific function i think it makes more sens to have it next to it

Source/monster.cpp
AJenbo: same

Source/objects.h
AJenbo: it appears to only be used to define placment of shrines.

Source/quests.cpp
AJenbo: yeah, just got rid of a warning to make it cleare to see if i caused any issues

defs.h
AJenbo: more complicated to move, save them for round 2

Source/dead.h
```diff
@@ -5,8 +5,12 @@
  */
 #pragma once
 
+#include <SDL.h>
```
glebm, line 8: Is this just for the `Uint8`? Would probably be better to switch that `Uint8` to `std::uint8_t` instead. Anyway, it's a minor thing.

glebm (approved):

Source/dead.h
AJenbo: Yeah, the good news is that this makes it clear what files this should be cleaned up in :) I decided to do so at a later point.

Source/objects.h
julealgon: Shouldn't it still be moved to a level-related header though? I think what matters most is the semantic of the value, not its use.

AJenbo: We could rename it SHRINE_END_LEVEL if that would help.

PR #1536, opened 2021-04-15
♻️ Unify most valuable player item logic on new method

julealgon:
This PR moves and centralizes the logic for checking the player's most valuable item (used in store calculations) as a new method in the `PlayerStruct` struct.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -375,180 +375,6 @@ static bool IsSuffixValidForItemType(int i, int flgs)
 	return (flgs & PLIType) != 0;
 }
 
-int get_ring_max_value(int i)
-{
-	int j, res;
-
-	res = 0;
-	for (j = 0; j < NUM_INVLOC; j++) {
-		if (plr[i].InvBody[j]._iClass != ICLASS_NONE && plr[i].InvBody[j]._itype == ITYPE_RING && res < plr[i].InvBody[j]._iIvalue)
-			res = plr[i].InvBody[j]._iIvalue;
-	}
-	for (j = 0; j < NUM_INV_GRID_ELEM; j++) {
-		if (plr[i].InvList[j]._iClass != ICLASS_NONE && plr[i].InvList[j]._itype == ITYPE_RING && res < plr[i].InvList[j]._iIvalue)
-			res = plr[i].InvList[j]._iIvalue;
-	}
-
-	return res;
-}
-
-int get_bow_max_value(int i)
-{
-	int j, res;
-
-	res = 0;
-	for (j = 0; j < NUM_INVLOC; j++) {
-		if (plr[i].InvBody[j]._iClass != ICLASS_NONE && plr[i].InvBody[j]._itype == ITYPE_BOW && res < plr[i].InvBody[j]._iIvalue)
-			res = plr[i].InvBody[j]._iIvalue;
-	}
-	for (j = 0; j < NUM_INV_GRID_ELEM; j++) {
-		if (plr[i].InvList[j]._iClass != ICLASS_NONE && plr[i].InvList[j]._itype == ITYPE_BOW && res < plr[i].InvList[j]._iIvalue)
-			res = plr[i].InvList[j]._iIvalue;
-	}
-
-	return res;
-}
-
-int get_staff_max_value(int i)
-{
-	int j, res;
-
-	res = 0;
-	for (j = 0; j < NUM_INVLOC; j++) {
-		if (plr[i].InvBody[j]._iClass != ICLASS_NONE && plr[i].InvBody[j]._itype == ITYPE_STAFF && res < plr[i].InvBody[j]._iIvalue)
-			res = plr[i].InvBody[j]._iIvalue;
-	}
-	for (j = 0; j < NUM_INV_GRID_ELEM; j++) {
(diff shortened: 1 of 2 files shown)


Source/player.h
```diff
@@ -313,6 +313,38 @@ struct PlayerStruct {
...
+	 * @return The most valuable item out of all the player's items that match the given predicate, or 'nullptr' in case no
+	 * matching items were found.
+	 */
+	template <typename TPredicate>
+	const ItemStruct *GetMostValuableItem(TPredicate itemPredicate) const
+	{
+		auto getMostValuableItem = [](const ItemStruct *begin,
+		                               const ItemStruct *end,
+		                               TPredicate itemPredicate,
+		                               const ItemStruct *&mostValuableItem) {
```
glebm, line 328: To make this more readable:
Replace the `const ItemStruct *&mostValuableItem` argument with `const ItemStruct *mostValuableItem` and return the item instead.

Source/player.h
```diff
@@ -313,6 +313,38 @@ struct PlayerStruct {
...
+	 * @param itemPredicate The predicate used to match the items.
+	 * @return The most valuable item out of all the player's items that match the given predicate, or 'nullptr' in case no
+	 * matching items were found.
+	 */
+	template <typename TPredicate>
+	const ItemStruct *GetMostValuableItem(TPredicate itemPredicate) const
+	{
+		auto getMostValuableItem = [](const ItemStruct *begin,
+		                               const ItemStruct *end,
+		                               TPredicate itemPredicate,
```
glebm, line 327: To avoid a copy of the lambda this should be `const TPredicate &itemPredicate`.

But even better is to simply capture it: `[&itemPredicate]`

Source/player.h
```diff
@@ -313,6 +313,38 @@ struct PlayerStruct {
...
+	/**
+	 * @brief Gets the most valuable item out of all the player's items that match the given predicate.
+	 * @param itemPredicate The predicate used to match the items.
+	 * @return The most valuable item out of all the player's items that match the given predicate, or 'nullptr' in case no
+	 * matching items were found.
+	 */
+	template <typename TPredicate>
+	const ItemStruct *GetMostValuableItem(TPredicate itemPredicate) const
+	{
+		auto getMostValuableItem = [](const ItemStruct *begin,
```
glebm, line 325: `const auto`

Source/items.cpp
```diff
@@ -4656,36 +4482,32 @@ static void SpawnOnePremium(int i, int plvl, int myplr)
...
-			ivalue = get_staff_max_value(myplr);
-			break;
 		case ITYPE_RING:
-			ivalue = get_ring_max_value(myplr);
-			break;
-		case ITYPE_AMULET:
-			ivalue = get_amulet_max_value(myplr);
+		case ITYPE_AMULET: {
+			auto mostValuablePlayerItem = plr[myplr].GetMostValuableItem(
+			    [](const ItemStruct &item) { return item._itype == items[0]._itype; });
```
glebm, line 4506: 1. This is different from the block below which uses `itemType`
2. This lambda does not capture `items` so it does not have access to it

glebm (changes requested):

Source/items.cpp
```diff
@@ -4656,36 +4482,32 @@ static void SpawnOnePremium(int i, int plvl, int myplr)
 		switch (items[0]._itype) {
 		case ITYPE_LARMOR:
 		case ITYPE_MARMOR:
-		case ITYPE_HARMOR:
-			ivalue = get_armor_max_value(myplr);
+		case ITYPE_HARMOR: {
+			auto mostValuablePlayerArmor = plr[myplr].GetMostValuableItem(
```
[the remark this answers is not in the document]
glebm: clang-format should already be supported by your IDE https://devblogs.microsoft.com/cppblog/clangformat-support-in-visual-studio-2017-15-7-preview-1/

Source/items.cpp
AJenbo: items is the global list of items in a level

Source/player.h
julealgon: I initially made this function return the item, however then it became convoluted to "find the most valuable item among belt, inventory and equipment".

By passing a writable pointer, I'm able to "accumulate" the greatest one from all 3 by just calling them in sequence.

I guess you want me to pass the reference to the "maximum thus far" but still return it? I'll take a look.

Source/player.h
julealgon: > But even better is to simply capture it: [&itemPredicate]

I was avoiding capturing any variables since you said a lambda with captures was not zero-cost.

Source/items.cpp
julealgon: > 1. This is different from the block below which uses itemType

Yes it is. They were different already before this change.

Source/items.cpp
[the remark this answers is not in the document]
julealgon: clang-format is not working properly for me. I suspect it could be conflicting with .editorconfig. I manually formatted the indents for now.

AJenbo (approved):

glebm (approved):

glebm:
> I was avoiding capturing any variables since you said a lambda with captures was not zero-cost.

zero-cost only in the sense that it is no different from a regular function that's not a lambda.
Capturing a variable makes the lambda more like a struct with a field and a method.

PR #1542, opened 2021-04-15
Add an option to autosort the inventory

Jmgr:
This add an option to auto sort the inventory using a rectangular bin-packing solver.

This is just a POC PR for now, it lacks of polish and generalization to it could be used in other places. It also only works for items that are grabbed from the ground, not the ones brought in shops (but that should not be difficult to add). I will also add an option to enable this feature (I suppose it will be disabled by default).

There a many algorithms that can be used, as well as multiple options to choose from when bin-packing (see https://github.com/juj/RectangleBinPack/blob/master/MaxRectsBinPack.h#L35-L39). Well' need to experiment a bit.

### DIFF
--- a/3rdParty/RectangleBinPack/CMakeLists.txt
+++ b/3rdParty/RectangleBinPack/CMakeLists.txt
@@ -1,0 +1,14 @@
+include(FetchContent_MakeAvailableExcludeFromAll)
+
+if(DEVILUTIONX_STATIC_RECTANGLEBINPACK)
+  set(BUILD_SHARED_LIBS OFF)
+else()
+  set(BUILD_SHARED_LIBS ON)
+endif()
+include(FetchContent)
+FetchContent_Declare(rectanglebinpack
+    GIT_REPOSITORY https://github.com/juj/RectangleBinPack.git
+    GIT_TAG 34833639923ddff3db2b050f56837ef60a625658
+)
+FetchContent_MakeAvailableExcludeFromAll(rectanglebinpack)
+set(rectanglebinpack_SOURCE_DIR ${rectanglebinpack_SOURCE_DIR} PARENT_SCOPE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,6 +165,8 @@ if(NOT NONET)
   endif()
 endif()
 
+add_subdirectory(3rdParty/RectangleBinPack)
+
 if(USE_SDL1)
   find_package(SDL REQUIRED)
   find_package(SDL_ttf REQUIRED)
@@ -447,13 +449,14 @@ if(GPERF)
   find_package(Gperftools REQUIRED)
 endif()
 
-target_include_directories(devilution PUBLIC Source SourceS ${CMAKE_CURRENT_BINARY_DIR})
+target_include_directories(devilution PUBLIC Source SourceS ${CMAKE_CURRENT_BINARY_DIR} ${rectanglebinpack_SOURCE_DIR})
 target_include_directories(${BIN_TARGET} PRIVATE
   SourceS
   SourceX
   3rdParty/asio/include
   3rdParty/Radon/Radon/include
-  3rdParty/libsmacker)
+  3rdParty/libsmacker
+  ${rectanglebinpack_SOURCE_DIR})
 
 if(N3DS)
   target_link_libraries(devilution PUBLIC
@@ -470,7 +473,8 @@ target_link_libraries(${BIN_TARGET} PRIVATE
   PKWare
   StormLib
   smacker
-  Radon)
+  Radon
+  RectangleBinPack)
 
 if(NOT NONET)
   target_link_libraries(${BIN_TARGET} PRIVATE sodium)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,8 @@ option(DEVILUTIONX_SYSTEM_LIBSODIUM "Use system-provided libsodium" ON)
 cmake_dependent_option(DEVILUTIONX_STATIC_LIBSODIUM "Link static libsodium" OFF
                        "DEVILUTIONX_SYSTEM_LIBSODIUM AND NOT DIST" ON)
 
+option(DEVILUTIONX_STATIC_RECTANGLEBINPACK "Link static RectangleBinPack" ON)
+
 if(NOT VERSION_NUM)
   include(CMake/git.cmake)
   get_git_tag(VERSION_NUM)
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -5,6 +5,7 @@
  */
 #include "all.h"
 #include "options.h"
+#include "MaxRectsBinPack.h"
 
 namespace devilution {
 
@@ -877,76 +878,68 @@ bool AutoEquipEnabled(const PlayerStruct &player, const ItemStruct &item)
  */
 bool AutoPlaceItemInInventory(int playerNumber, const ItemStruct &item, bool persistItem)
 {
+	rbp::MaxRectsBinPack binPack(10, 4, false); // TODO: Cache this
+
 	InvXY itemSize = GetInventorySize(item);
-	bool done = false;
 
-	if (itemSize.X == 1 && itemSize.Y == 1) {
-		for (int i = 30; i <= 39 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
+	// Add the new item to the input rects.
+	std::vector<rbp::RectSize> input {
+		rbp::RectSize{itemSize.X, itemSize.Y}
+	};
+
+	// Add all current item to the input rects.
+	for (int i = 0; i < plr[playerNumber]._pNumInv; i++) {
+		const auto &item = plr[playerNumber].InvList[i];
+		auto size =  GetInventorySize(item);
+		input.emplace_back(rbp::RectSize{size.X, size.Y});
+	}
+
+	std::vector<rbp::Rect> output;
+
+	binPack.Insert(input, output, rbp::MaxRectsBinPack::RectBestShortSideFit);
+
+	if(output.size() != plr[playerNumber]._pNumInv+1) {
+		// Unable to place item: inventory is full.
+		return false;
+	}
+
+	if(persistItem) {
+		// Empty the item grid.
+		for (int i = 0; i < NUM_INV_GRID_ELEM; i++) {
+			plr[playerNumber].InvGrid[i] = 0;
 		}
-
-		for (int i = 20; i <= 29 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
+		for (int i = 0; i < plr[playerNumber]._pNumInv; ++i) {
+			auto &item = plr[playerNumber].InvList[i];
+			auto size =  GetInventorySize(item);
+			bool placed = false;
+			// Look for a place to put the item.
+			for (auto it = output.begin(); it != output.end();) {
+				if (it->width == size.X && it->height == size.Y) {
+					item._ix = it->x;
+					item._iy = it->y;
+					int slot = it->x % 10 + it->y * 10;
+					AddItemToInvGrid(playerNumber, slot, i+1, size.X, size.Y);
+					it = output.erase(it);
+					placed = true;
+					break;
+				} else {
+					++it;
+				}
+			}
+			if(!placed) {
+				assert(false && "failed to place item");  // TODO
+			}
 		}
-
-		for (int i = 10; i <= 19 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-
-		for (int i = 0; i <= 9 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
+		assert(output.size() == 1); // TODO
+		
+		int slot = output[0].x % 10 + output[0].y * 10;
+		auto placed = AutoPlaceItemInInventorySlot(playerNumber, slot, plr[playerNumber].HoldItem, persistItem);
+		if(!placed) {
+			assert(placed && "failed to place item"); // TODO
 		}
 	}
 
-	if (itemSize.X == 1 && itemSize.Y == 2) {
-		for (int i = 29; i >= 20 && !done; i--) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-
-		for (int i = 9; i >= 0 && !done; i--) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-
-		for (int i = 19; i >= 10 && !done; i--) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-	}
-
-	if (itemSize.X == 1 && itemSize.Y == 3) {
-		for (int i = 0; i < 20 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-	}
-
-	if (itemSize.X == 2 && itemSize.Y == 2) {
-		for (int i = 0; i < 10 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, AP2x2Tbl[i], item, persistItem);
-		}
-
-		for (int i = 21; i < 29 && !done; i += 2) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-
-		for (int i = 1; i < 9 && !done; i += 2) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-
-		for (int i = 10; i < 19 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-	}
-
-	if (itemSize.X == 2 && itemSize.Y == 3) {
-		for (int i = 0; i < 9 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-
-		for (int i = 10; i < 19 && !done; i++) {
-			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
-		}
-	}
-
-	return done;
+	return true;
 }
 
 /**
@@ -989,7 +982,7 @@ bool AutoPlaceItemInInventorySlot(int playerNumber, int slotIndex, const ItemStr
 		yy += 10;
 	}
 	if (done && persistItem) {
-		plr[playerNumber].InvList[plr[playerNumber]._pNumInv] = plr[playerNumber].HoldItem;
+		plr[playerNumber].InvList[plr[playerNumber]._pNumInv] = item;//plr[playerNumber].HoldItem;
 		plr[playerNumber]._pNumInv++;
 
 		AddItemToInvGrid(playerNumber, slotIndex, plr[playerNumber]._pNumInv, itemSize.X, itemSize.Y);


AJenbo:
Does this work by preplanning likey layout or by reshuffling the inventory as you pick up more items?

Jmgr:
> Does this work by preplanning likey layout or by reshuffling the inventory as you pick up more items?

This reshuffles after each new item. We should profile this, but knowing the size of the inventory and the small number of items I'm fairly confident this should not be an issue.

AJenbo:
Ok, I would prefer to start off with something that simply has a strategy for how to add items in the first place, and then only tries reshuffling if nessesery.

Jmgr:
I've just tested a release build on my computer, sorting a full inventory of mixed items takes around 40 microseconds. That's like 0.000040 seconds.

AJenbo:
> I've just tested a release build on my computer, sorting a full inventory of mixed items takes around 40 microseconds. That's like 0.000040 seconds.

Sounds fast enough :)

Jmgr:
> Ok, I would prefer to start off with something that simply has a strategy for how to add items in the first place, and then only tries reshuffling if nessesery.

The advantage of activating it all the time (if the player decides to use it of course) is that the same item will always go at the same place given a similar configuration, there is no randomness here. If you just add new items on the first free slot and only sort when there is none then all items will be moved at once, which might seem odd to the player.

qndel:
tbh this is cool 😮

AJenbo:
> > Ok, I would prefer to start off with something that simply has a strategy for how to add items in the first place, and then only tries reshuffling if nessesery.
> 
> The advantage of activating it all the time (if the player decides to use it of course) is that the same item will always go at the same place given a similar configuration, there is no randomness here. If you just add new items on the first free slot and only sort when there is none then all items will be moved at once, which might seem odd to the player.

Ok, I was thinking that the user might have some preferred layout and it should try not messing with that.

Jmgr:
> > > Ok, I would prefer to start off with something that simply has a strategy for how to add items in the first place, and then only tries reshuffling if nessesery.
> > 
> > 
> > The advantage of activating it all the time (if the player decides to use it of course) is that the same item will always go at the same place given a similar configuration, there is no randomness here. If you just add new items on the first free slot and only sort when there is none then all items will be moved at once, which might seem odd to the player.
> 
> Ok, I was thinking that the user might have some preferred layout and it should try not messing with that.

That makes sense, we can certainly provide some options on how the items should be organized though. The bin packing library provides some [options](https://github.com/juj/RectangleBinPack/blob/master/MaxRectsBinPack.h?rgh-link-date=2021-04-15T18%3A34%3A01Z#L35-L39) already. In the current configuration it tends to put the largest items on the top left, and the smaller ones at the bottom and on the right:
image

Jmgr:
We can also easily group similar items of the same size, like putting all the gold together, all the potions, etc.

AJenbo:
That sounds like a good idea.

Here is how I usually organize things to try and avoid having to resort to thing:

image

1x1 in the bottom row, once the bottom row is filled continue from the right, filling columns instead of rows
1x3 / 2x3 in the top row
1x2 / 2x2 starting from the right, filling columns instead of rows

image

Following these 3 rules should avoid the need to reshuffle items. Except in situations where a dagger has made the right side uneven or the user has done some manual ordering.

julealgon:
I'm personally not a fan of this as it automates too much and the game loses an important aspect to it.

What I'm in favor of (and have on my to-do notes) is to change the default iteration sequence used to place items in the inventory. The standard sequence promotes clutter unnecessarily and could be adjusted accordingly to take item sizes more into account. One easy change that would dramatically improve inventory traversal is by changing the 1x1 item iteration sequence to be vertical-first instead of horizontal-first on lines 2, 3 and 4 and 5 of the inventory (keep horizontal-first on the first line since the game doesn't have Yx4 items).

That should make the need to reorganize the inventory much rarer.

Jmgr:
> I'm personally not a fan of this as it automates too much and the game loses an important aspect to it.

That's why it's an option. I personally dislike inventory management in any game so it's all subjective.

> What I'm in favor of (and have on my to-do notes) is to change the default iteration sequence used to place items in the inventory. The standard sequence promotes clutter unnecessarily and could be adjusted accordingly to take item sizes more into account. One easy change that would dramatically improve inventory traversal is by changing the 1x1 item iteration sequence to be vertical-first instead of horizontal-first on lines 2, 3 and 4 and 5 of the inventory (keep horizontal-first on the first line since the game doesn't have Yx4 items).
> 
> That should make the need to reorganize the inventory much rarer.
> 
> EDIT: I hadn't read @AJenbo 's post above before writing this. The idea is very similar.

> Following these 3 rules should avoid the need to reshuffle items. Except in situations where a dagger has made the right side uneven or the user has done some manual ordering. But that can probably be worked around as well.

The good news is that the MaxRectsBinPack algorithm (with the current settings, I'll post some pictures to illustrate what the different settings do) does mostly what both of you want: it puts bigger items on the top left and then organizes the smaller ones "moving" to the bottom and on the right.

The issue I have with both your solutions is that they are not generic and will need to be custom tailored for new use-cases, like the stash, or when a mod adds a new item with a new size (or allows rotating items, or allows a bigger inventory, etc.).

What I would suggest is to add that as an option: the player could choose what kind of management algorithm they want. That could be explained in the documentation with a few pictures.

julealgon:
> That's why it's an option. I personally dislike inventory management in any game so it's all subjective.

My point is that such dramatic changes should not be part of DevilutionX, not even as an option.

> The issue I have with both your solutions is that they are not generic and will need to be custom tailored for new use-cases, like the stash, or when a mod adds a new item with a new size (or allows rotating items, or allows a bigger inventory, etc.).

It is not a problem now since those features do not exist. YAGNI.

> What I would suggest is to add that as an option: the player could choose what kind of management algorithm they want. That could be explained in the documentation with a few pictures.

I have thought about this, but in a much simpler way: allow the user to specify the item collection order explicitly, cell by cell. The grid is small enough to allow for that.

malvarenga123:
A stash would be a dramatic change to Diablo, by that logic DevilutionX should not have it, even as an option.
Your own suggestion of manual blocking would be a dramatic change to the core mechanics of the game.

I'm with @Jmgr here, inventory tetris is boring, we have enough of that in Diablo 2 and it always leads to heated debates.
If a stash is implemented, a sorting button would be great. That's precisely the reason it has been implemented in other games like Grim Dawn and Torchlight 2.

julealgon:
@malvarenga123 
> A stash would be a dramatic change to Diablo, by that logic DevilutionX should not have it, even as an option.

And I'm personally not in favor of one.

> Your own suggestion of manual blocking would be a dramatic change to the core mechanics of the game.

I see that as a QOL change, since you can still get the behavior by manually unequipping the shield for example.

> inventory tetris is boring, we have enough of that in Diablo 2 and it always leads to heated debates.

Inventory tetris is a core mechanic in the game. Some people find it boring (like yourself) but that doesn't mean it should be removed. I don't care about Diablo 2 discussions (nor do I know which debates you are referring to). The Diablo 3 team also thought inventory tetris was boring.

> If a stash is implemented, a sorting button would be great. 

Depends on how the stash is implemented. Inventory-style stash is _not_ the only possible implementation of a stash.

> That's precisely the reason it has been implemented in other games like Grim Dawn and Torchlight 2.

"Because it was implemented in game X and Y" is not a good reason for implementing it IMHO.

Jmgr:
> My point is that such dramatic changes should not be part of DevilutionX, not even as an option.

For me this change belongs to QoL. But again this is subjective so I can really say much else here.

> It is not a problem now since those features do not exist. YAGNI.

Why not building something reasonably future proof while being simple (literally a dozen lines of code) rather than a specific and complicated solution?

> I have thought about this, but in a much simpler way: allow the user to specify the item collection order explicitly, cell by cell. The grid is small enough to allow for that.

I don't understand what you mean here.

> Inventory tetris is a core mechanic in the game.

Or it's juts because at the time they created the game they didn't have the time/money/resources to implement auto-sorting. Who knows. We could poll players and ask for their opinion, after all they will play the game in the end. Deciding those things between software developers is usually not great.

> "Because it was implemented in game X and Y" is not a good reason for implementing it IMHO.

Why not take inspiration from commercial games that have received the attention of complete game development teams like game designers and ergonomics specialists? Why only consider that our use-case as game developers is necessarily the right one? Why reinvent the wheel? If something works, use it.

malvarenga123:
>I see that as a QOL change, since you can still get the behavior by manually unequipping the shield for example.
I also see the inventory sort as a QOL change, since you can keep manually organizing your inventory if you want.

See how that works both ways?

galaxyhaxz:
Question: why can't there be a fork of devilutionX on diasurgical that mods the game to the heart's desire alongside the main repo which keeps the game clean and only applies bugfixes? This way, you don't have to sort through what is QoL enough or not.

malvarenga123:
@galaxyhaxz I was under the impression that this was the case, but take a look at Ajenbo's response:
https://github.com/Manuel-K/devilutionX-QOL-patches/issues/11#issuecomment-716154872

julealgon:
@Jmgr 
> For me this change belongs to QoL. But again this is subjective so I can't really say much else here.

True. It's not really possible to define a precise boundary for what is QOL and what isn't I believe. At the end of the day, we have to go with our gut feeling of what makes more sense for the game, and Ajenbo has the final say as the repo owner. IMHO, just outright blindly accepting every option imaginable into the repo will make it too complex to maintain in the long run so I tend to be very cautious about what to add and what to not add. That's just me personally though. and my take does not really matter at the end of the day.

---

> Why not build something reasonably future proof while being simple (literally a dozen lines of code) rather than a specific and complicated solution?

This is just a general design philosophy: you are never certain of what the future will actually need, so you wait until there is a real need to then design a solution and have it introduced. Before the problem exists, you focus on what you have currently. It is extremely rare that you'll predict with 100% accuracy what you'll need next, so it's usually just not worth it.

Also, I'm not sure what you mean by "complicated solution". I'm not proposing any automated inventory sorting at all.

---

> I don't understand what you mean here.

Items are added to the inventory using a specific sequence of slots. You can see this sequence for yourself on `inv.cpp`'s `AutoPlaceItemInInventory`:
https://github.com/diasurgical/devilutionX/blob/2302ee27aa0ecf13752ba596f9ec221f89f0b8a4/Source/inv.cpp#L649-L721

Notice how it uses a slightly different approach depending on the size of the item.

What I mean is that _that_ logic can be improved in such a way as to minimize the need to manually "reorganize" the inventory. For example, for 1x1 items, the game will attempt to place them with horizontal priority over vertical. This is fine for the bottom-most line of the inventory (due to the abundance of 1x3 and 2x3 items in the game), however it is really bad for the other lines: the fact that it keeps adding items moving right causes other item to then not fit, which in turn forces the player to manually adjust the items again.

If we simply change the _order_ that the inventory cells are inspected when getting items, we can optimize it for the average item sizes in the game. For 1x1 items to not block bigger items, we want them to be populated in a single line horizontally first, but then populated vertically once the bottom line is already filled. This is just a change to iteration order of the cells.

You could optimize this further to take 2x2 items into account as well, perhaps have the bottom row only go as far as 8 slots, and then start going vertical for 1x1 items, allowing space for 2 2x2 items at the right-most corner of the inventory.

Same goes for other item sizes.

What I mean by "allowing one to specify this" is exactly that: let the user configure his preferences for how the game puts items into the inventory. If he wants the top line as 1x1 items, instead of the bottom line, he could configure it that way. If he wants the inventory to be filled from left to right, instead of right to left, he could also make that work.

This is _not_ the same as auto-sorting. There is no auto-sorting involved here, its about _the initial placement_ of the items.

---

> Or it's juts because at the time they created the game they didn't have the time/money/resources to implement auto-sorting. Who knows. 

I think you answered this yourself: who knows?

---

> We could poll players and ask for their opinion, after all they will play the game in the end. Deciding those things between software developers is usually not great.

Disagree strongly on this mindset. You usually don't create great stuff by "polling users". Users usually _do not know what they want_. Instead, you think about something that users trully want (but don't know they do) and implement that. Making games is not a democracy where people pick and chose what they want the game to have.

---

> Why not take inspiration from commercial games that have received the attention of complete game development teams including game designers and ergonomics specialists? 

I think you misunderstood what I said. I said that "just the fact that game X uses something" is not a reason by itself to justify doing it. You have to look at the underlying reasons to do so. Game X also did it for some reason, and _that's_ what should be the focus of the analysis. I never said you can't take inspiration on other games.

---

> Why only consider that our use-case as game developers is necessarily the right one? 

See above. Users generally do _not_ know what they want. They think they know, but they don't.

---

> Why reinvent the wheel? If something works, use it.

If the underlying reason for using the wheel matches with what one wants, then sure. Otherwise you might want to use a maglev system, or a flying method, or a treadmill mechanism, instead of wheels, if that makes more sense on your context.

---

BTW, I feel like it would be best for us to continue these discussions in an issue, not on a PR. Would be cool if you could move your proposal to an issue.

Jmgr:
> This is just a general design philosophy: you are never certain of what the future will actually need, so you wait until there is a real need to then design a solution and have it introduced. Before the problem exists, you focus on what you have currently. It is extremely rare that you'll predict with 100% accuracy what you'll need next, so it's usually just not worth it.

Well there is a need from at least 3 people that have commented here. ;)

> Also, I'm not sure what you mean by "complicated solution". I'm not proposing any automated inventory sorting at all.

No, but you propose a placement strategy that will work for this inventory and those items sizes, but break if you add just one line of inventory space or another item shape.

> you can see this sequence for yourself

Yes that is the piece of code I replaced. It's very copy and pasty. Keeping the same algorithm it could certainly be improved, but that's not what this PR is about.

> What I mean is that that logic can be improved in such a way as to minimize the need to manually "reorganize" the inventory.

I understand now, thanks for the explanation.

> Users usually do not know what they want.

At least 3 people seem to want to sort their inventory here.

> I said that "just the fact that game X uses something" is not a reason by itself to justify doing it.

:+1: 

> BTW, I feel like it would be best for us to continue these discussions in an issue, not on a PR. Would be cool if you could move your proposal to an issue.

Why? Has this PR been refused? @AJenbo?

AJenbo:
> Why? Has this PR been refused? @AJenbo?

On GitHub there is barely a distinction between an issue, a discussion, or a pull request, they all use the same sequential id, and you can actually access an issue via a PR like URL and vice versa. So as long as it is a draft I think it works as a POC/RFC. Some people work better with building prototypes, others better with writing specifications, both approaches are valid IMO.

https://github.com/diasurgical/devilutionX/issues/1542

Jmgr:
> On GitHub there is barely a distinction between an issue, a discussion, or a pull request, they all use the same sequential id, and you can actually access an issue via a PR like URL and vice versa. So as long as it is a draft I think it works as a POC/RFC. Some people work better with building prototypes, others better with writing specifications, both approaches are valid IMO.
> 
> `https://github.com/diasurgical/devilutionX/issues/1542`

Yes, well, I prefer creating a PoC/draft until I know that a concept has been reasonably well accepted (or at least the maintainers said that it could be merged once finished). Spending multiple hours on something before seeing it refused is always a bit sad.

By the way, I'm happy to work on this PR this week-end, but I would need to know first if it has a chance of being merged in the end. I opened this PR because I have used rectangular bin packing in the past and I believe it works well for that use-case, but I understand that at least you, @AJenbo, are not convinced of it's usefulness and would prefer a custom tailored algorithm similar to the one that is already there. So since you are the person choosing what gets in or not I'm asking you. ;)

AJenbo:
> Question: why can't there be a fork of devilutionX on diasurgical that mods the game to the heart's desire alongside the main repo which keeps the game clean and only applies bugfixes? This way, you don't have to sort through what is QoL enough or not.

Should it just have auto-merging of any PR made against it? We can probably set that up, but I don't think it would result in anything playable...
Managing multiple repos is an overhead, syncinging between them more so, imagine the conflict after having accepted every QoL known to man in to DevilutionX-qol and then merging in the latest DevilutionX where we rewrote the codebase :boom:. We would also still have to decide that is a mod or a QoL.

galaxyhaxz:
> Should it just have auto-merging of any PR made against it? We can probably set that up, but I don't think it would result in anything playable...

Inevitably it would be more work to maintain two repos, so it's understandable. It seems now that as things have matured many people have got involved in the project over the past 6 months (that and people being at home more because of covid). Therefore everyone is trying to push their own ideas into the game, and it's getting messier and messier to have a ton of different features but all of them have to be optional.

I think a community "mod" would suffice if someone is willing to do the work of syncing in the latest changes from the devilutionX master.

AJenbo:
I do like having an auto sorting option, it does however feel like a lot to pull in a +2600 line dependency for. Especially with dependency management being such a mess in C++
Would have to look a bit more at it, though at a glance it looks pretty good. If it was used in more aspects of the code that would make me more inclined to include it, for example, something critical like bin packing textures etc.

FitzRoyX:
That was my reaction when I first saw the pull. Adds a lot of code and libraries just to defeat a mechanic that 90% of players arent bothered by. Meanwhile, everyone complains about storage space and town walk speed.

Jmgr:
I don't understand this issue with dependencies. I feel many C++ developers prefer to re-write code rather than use existing one. I think this stems from the lack or uniformly used tools, but now with CMake being the most used used one by far this seems a bit odd. In Go everyone just uses the dependencies they need and that works very well. So I'm not convinced by the dependency argument, especially because if you want you can just copy and paste the 1 or 2 headers/source files from the MaxRects algorithm into the project.

> Would have to look a bit more at it, though at a glance it looks pretty good.

I'll leave that PR open but work on something else then. Not sure I get motivated to work again on that at a later date, but let's see.

> Adds a lot of code and libraries just to defeat a mechanic that 90% of players arent bothered by.

Where does this 90% comes from? How many players have you asked?

Iniquitatis:
@Jmgr, dependencies? Pfft!
(You wouldn't want to know how many times I reinvented the wheel... :D)

AJenbo:
> Meanwhile, everyone complains about ... town walk speed.

still withe both run in tow and game speed now allowing you to move at a 5x speed?

AJenbo:
> I don't understand this issue with dependencies. I feel many C++ developers prefer to re-write code rather than use existing one. I think this stems from the lack or uniformly used tools

it is still very non-trivial and basic compared to what I'm used to
- getting dependencies to install steams to be the most common stumbling block for new developers that wants to build the code
- we had to write a custom cmake helper
- there in no version or comparability checks
- it locks us even tighter to cmake
- half the systems we port to the communities refuse to help because they don't like cmake
- We still haven't setup caching for dependencies on the build servers
- the long jevity of projects on github is case by case, so we either need to fork or have it be a optional dependency
- upgrading a dependency is a manual process and there is no overview of what has new versions and there change logs
- for some systems you don't actually know if a dependency was correctly linked before you run the application on that platform

> with CMake being the most used used one by far this seems a bit odd. In Go everyone just uses the dependencies they need and that works very well.

im going to go out on a limb and say cmake is probably used in 20% of cases tops, handedly being beaten by projects that have not figured out to setup any form of dependency managment. Go just have one option that comes ready by default.

> if you want you can just copy and paste the 1 or 2 headers/source files from the MaxRects algorithm into the project.

I dont, its an even worse way to handle dependencies. I hate our 3rdparty folder.

> Adds a lot of code and libraries just to defeat a mechanic that 90% of players arent bothered by.

its less then 100 lines and only add one, pretty trivial if not for the dependancy situation. and i think a lot of users would actually welcome this.

mwpow3ll:
As someone that just recently re-played Diablo 1 using DevilutionX 1.2.0 and 1.2.1 and beat Normal, Nightmare, and Hell; I think that @julealgon has the more 'safe', better maintainable idea that keeps the heart of Diablo inventory tetris there while also likely minimizing 95% of the headaches without 3rd party addition or causing issues further down the line. See below:

Diablo / Diablo 2 veterans like myself, we constantly carry a ton of potions and feed them into the belt manually while walking or fighting. The solution by @Jmgr would auto sort the inventory every time and could be jarring to the user in the heat of the moment if they find that some items are drastically changed around.

Jmgr:
From the various discussions here I can see that while there is certainly some interest for this feature, this PR is not going to convince people. So I'm closing this PR (see https://github.com/diasurgical/devilutionX/pull/1876 for an alternative implementation).

Maybe when we have a stash and/or items of various shapes (via mods) this might be useful to someone.

PR #1543, opened 2021-04-15
Invcleanup part 2

qndel:
As an addition to the cleanup, this fixes ICOL_WHITE to be a proper white color instead of yellow as before,
Adds "Item Outlines: In the dungeon, the color drawn to outline items on the floor now match the type of item (e.g. Gold outline = unique item)."
from https://github.com/diasurgical/devilutionX/issues/1136 qol list (it was in infernity too ;))
Outlines on the ground and in inventory are showing item rarity now


AJenbo (approved):

Source/inv.cpp
```diff
@@ -205,7 +205,7 @@ void DrawInv(CelOutputBuffer out)
 				screen_x += frame_width == INV_SLOT_SIZE_PX ? 14 : 0;
 				screen_y += InvItemHeight[frame] == (3 * INV_SLOT_SIZE_PX) ? 0 : -14;
 			} else if (slot == INVLOC_HAND_RIGHT) {
-				screen_x = frame_width == INV_SLOT_SIZE_PX ? 13 : 1;
+				screen_x += frame_width == INV_SLOT_SIZE_PX ? 13 : 1;
```
julealgon, line 208: Guys... this is not cleanup. You are making it much harder to track things by conflating refactoring and reformatting with actual behavior changes (in this case a bug fix).

AJenbo: For some context, @qndel forgot to mark the PR as a draft and so it go merged early. This is fixing a bug in the previous PR. https://github.com/diasurgical/devilutionX/pull/1306

But yeah this was all a non-ideal situation.

PR #1546, opened 2021-04-15
Fix for a rare legacy bug causing network desync

FluffyQuack:
If a player tries to swap sword+shield with a two-handed weapon without sufficient inventory space it leads to a network desync (what's in the right hand gets de-equipped on other clients, but not for local client).

Video example: https://www.youtube.com/watch?v=2GtUv-m8FoI

The cause was the game always sending a network packet for de-equipping the right-hand item before even checking if the local player has space to do the weapon swap.

(Uh, never mind the shield position looking weird. That's an unrelated issue caused by another commit.)

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1113,7 +1113,6 @@ void CheckInvPaste(int pnum, int mx, int my)
 		cn = SwapItem(&plr[pnum].InvBody[INVLOC_HAND_RIGHT], &plr[pnum].HoldItem);
 		break;
 	case ILOC_TWOHAND:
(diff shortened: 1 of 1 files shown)


AJenbo:
Nice find

julealgon:
May I ask why you are using a cow emoji on all your changes? Can we not do that? The emojis are there to help understanding what a commit does, not for any other reason.

AJenbo:
My guess is that it's there to remind us to have fun while working on the project and not be too serious about it, this is probably especially true for people who aren't necessarily here to hone their skills as professional developers. Only a few of us use the emojis at all. I personally do not find the cow a distraction.

PR #1549, opened 2021-04-15
Restructure files

AJenbo:
Combine the 3 Source{*} folders into a single one.

Files SourceX/S/* becomes SourceX/S/utils or SourceX/S/storm, folders are moved as is.


CMakeLists.txt
```diff
@@ -666,18 +652,17 @@ if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
...
   target_compile_options(${BIN_TARGET} PRIVATE -Wall -Wextra -Wno-unused-parameter)
 
   # For ARM and other default unsigned char platforms
-  target_compile_options(devilution PRIVATE -fsigned-char)
   target_compile_options(${BIN_TARGET} PRIVATE -fsigned-char)
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
-  target_compile_options(devilution PRIVATE "/W0")
+  target_compile_options(${BIN_TARGET} PRIVATE "/W0")
```
julealgon, line 665: _Really_  hard to review this file...

Source/miniwin/miniwin.h
```diff
@@ -2,8 +2,22 @@
...
+#include <ctype.h>
+#include <math.h>
+#include "utils/stdarg_compat.h"
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#include "storm/storm_full.h"
```
julealgon, line 14: I don't understand this part. How could moving the files cause you to have to add all these headers?

Source/miniwin/miniwin.h
```diff
@@ -2,8 +2,22 @@
...
+
 namespace devilution {
 
+#ifndef MAX_PATH
+#define MAX_PATH 260
+#endif
```
julealgon, line 20: ?

Source/storm/storm.cpp
```diff
@@ -15,16 +16,16 @@
...
-#include "stubs.h"
+#include "utils/display.h"
+#include "utils/paths.h"
+#include "utils/sdl_compat.h"
+#include "utils/stubs.h"
 
 namespace devilution {
 
-DWORD nLastError = 0;
+uint32_t nLastError = 0;
```
julealgon, line 28: Unrelated?

julealgon (commented):
Can't really approve this myself as it touches on some things I do not fully understand, My approval would mean nothing.

Source/miniwin/miniwin.h
AJenbo: misc.h was combined with miniwin.h. minwin.h used to be a all.h for the minimal implementation of the Windows API used by Diablo, but at this point only misc.h was left and so one was never used without the other and made better sense as a combined file. This also makes it more clear where we need to clean things up in order to get rid of the last parts of miniwin.

Source/storm/storm.cpp
AJenbo: Much easier than to try and deal with the nested relation between storm.h and minwin.h. It was either this, have it not compile, or add a hack.

galaxyhaxz:
Great work in unifying the source tree Anders. I think this will help make it a lot more clear and organized now. Diablo's original design of having 100 files in one folder just isn't good design by modern standards. Then again, non of it was :P

CMakeLists.txt
glebm: This is OK, now there is just one target

glebm (approved):

Source/miniwin/miniwin.h
AJenbo: The max length of a file path as defined in WinAPI on Windows 95. This may be already defined on Windows platforms and should only be conditionally defined.

Again this comes from the merging of `SourceS/miniwin/misc.h` and `SourceS/miniwin.h`.

Hope that answers you questionmark.

AJenbo:
> Great work in unifying the source tree Anders. I think this will help make it a lot more clear and organized now. Diablo's original design of having 100 files in one folder just isn't good design by modern standards. Then again, non of it was :P

Yeah I'm hoping it will lead to less confusion for new developers looking at the code, it should also make it easier to analyse and run various tools on it :)

CMakeLists.txt
julealgon: I wasn't specifically talking about this particular change, but the file as a whole. Should've added a file-level comment, sorry for the confusion.

Source/storm/storm.cpp
AJenbo: ... and it's gone.

https://github.com/diasurgical/devilutionX/commit/e1278cba7b611e6ba130ccbebcff40746e73f65a#diff-7501f2d1b1864ababb342df86d326f82d9f92f4bb34b56b8ed91ac75c87b1a0a

PR #1551, opened 2021-04-16
Fix Storm error code on Windows

glebm:
`DVL_ERROR_HANDLE_EOF` was defined incorrectly on Windows.

These error codes are used and returned by Storm. Moves them from `miniwin.h` to `storm.h`.


AJenbo (approved):
P.s. it's always annoyed me that the bug is symbolized by a grub :+1:

PR #1555, opened 2021-04-16
bugfix pfile_flush_W() not clearing tables

Eider-McDuck:
With the new Archive class, pfile_flush_W() currently does nothing, since the archive is regarded as already closed. However, it should clear the hash and block tables.

pfile_flush_W() is called when leaving a game. If at that time the tables are not cleared, the next character loaded will use the same tables, resulting in crashes.

Example to reproduce: Have 2 vanilla savegames, join a game with the first, then leave, then join a game with the second, then drop an item on the floor.

### DIFF
--- a/Source/mpqapi.cpp
+++ b/Source/mpqapi.cpp
@@ -223,28 +223,32 @@ struct Archive {
 
 	bool Close(bool clear_tables = true)
 	{
-		if (!stream.IsOpen())
-			return true;
+		bool result = true;
+		if (stream.IsOpen()) {
 #ifdef _DEBUG
-		SDL_Log("Closing %s", name.c_str());
+			SDL_Log("Closing %s", name.c_str());
 #endif
 
-		bool result = true;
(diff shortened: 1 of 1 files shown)


glebm (commented):
In both places where `pfile_flush_W` is called, `pfile_write_hero` is called before it in `if (gbIsMultiplayer)`

Perhaps we could remove `pfile_flush_W` and instead add a `clear_tables` argument to `pfile_write_hero`?

Then we wouldn't need to change the `Close` function at all.

glebm:
@Eider-McDuck Thanks for the investigation and this PR!
Here is an alternative fix that I think simplifies things a bit: https://github.com/diasurgical/devilutionX/pull/1556, can you please give it it a try?

Eider-McDuck:
In the initalizer list you have a typo, you are initalizing clear_tables_ with itself. After changing that, it solves the bug. I like this solution.

PR #1556, opened 2021-04-16
🐞 pfile: Fix stale tables after leaving a game

glebm:
In multiplayer, the hash and block tables are kept around between the saves (presumably to improve performance).

When leaving the game, they should be cleared.

Clearing of the tables is handled by `Archive::Close` in `mpqapi.cpp`.

After #1426, which moved the archive closing inside `pfile_write_hero`, the tables were no longer cleared on `pfile_flush_W` (because the archive was already closed when it was called).

To fix the issue, this commit adds a `clear_tables` argument to `pfile_write_hero` and removes the `pfile_flush_W` function.

Bug reported and root cause found by @Eider-McDuck:

> Example to reproduce: Have 2 vanilla savegames, join a game with the first,
> then leave, then join a game with the second, then drop an item on the floor.

Initial proposed fix by @Eider-McDuck: #1555

### DIFF
--- a/Source/pfile.cpp
+++ b/Source/pfile.cpp
@@ -220,8 +220,9 @@ static void pfile_SFileCloseArchive(HANDLE *hsArchive)
 	*hsArchive = NULL;
 }
 
-PFileScopedArchiveWriter::PFileScopedArchiveWriter()
+PFileScopedArchiveWriter::PFileScopedArchiveWriter(bool clear_tables)
     : save_num_(pfile_get_save_num_from_name(plr[myplr]._pName))
+    , clear_tables_(clear_tables)
 {
 	if (!pfile_open_archive(save_num_))
 		app_fatal("Failed to open player archive for writing.");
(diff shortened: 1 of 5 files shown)


Eider-McDuck:
In the initalizer list you have a typo, you are initalizing clear_tables_ with itself. After changing that, it solves the bug. I like this solution.

Source/pfile.cpp
```diff
@@ -220,21 +220,22 @@ static void pfile_SFileCloseArchive(HANDLE *hsArchive)
 	*hsArchive = NULL;
 }
 
-PFileScopedArchiveWriter::PFileScopedArchiveWriter()
+PFileScopedArchiveWriter::PFileScopedArchiveWriter(bool clear_tables)
     : save_num_(pfile_get_save_num_from_name(plr[myplr]._pName))
+    , clear_tables_(clear_tables_)
```
AJenbo, line 225:
```suggestion
    , clear_tables_(clear_tables)

```

glebm:
Well spotted, thanks!

AJenbo (approved):

PR #1560, opened 2021-04-17
gamenu.cpp: Move most functions to an anonymous namespace

glebm:


AJenbo (approved):
I accept your improvement and substitute my own

PR #1562, opened 2021-04-17
Remove stdarg_compat.h header

glebm:
It was only used in one place, which did not use va_list.

Looking at the linked LLVM issue (https://reviews.llvm.org/D51265), my guess was whatever this was meant to work around was an issue in DevilutionX code at the time.


AJenbo:
miniwin is included in a lot of places, but lets see what happens.

AJenbo:
its possible that it was needed when we where compiling for 32bit mac using an older compiler, but dosent seam to cause any issues on even on my 10.13 system

PR #1564, opened 2021-04-17
fancy stairs2

qndel:
Much cleaner implementation than the previous attempt + at the moment I think recoloring is the only sane way as stairs are built from varying number of segments (1x1 for church/catas, hell = 3x2, caves stairs up = 1x2) so patterns look weird, same with attempting to display some symbol next to it

image
image
image

Current color code: 
red = stairs down
blue = stairs up
white = warp to town
standard color = quest stairs (chamber of bone only?)

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -13,6 +13,7 @@
 #include "setmaps.h"
 #include "utils/ui_fwd.h"
 #include "trigs.h"
+#include "options.h"
 
 namespace devilution {
 
@@ -115,24 +116,26 @@ void DrawAutomapTile(CelOutputBuffer out, Sint32 sx, Sint32 sy, Uint16 automap_t
 		DrawLineTo(out, sx - AmLine16, sy - AmLine8, sx + AmLine16, sy + AmLine8, COLOR_BRIGHT);
 		DrawLineTo(out, sx - AmLine16 - AmLine8, sy - AmLine4, sx + AmLine8, sy + AmLine8 + AmLine4, COLOR_BRIGHT);
 		DrawLineTo(out, sx - AmLine32, sy, sx, sy + AmLine16, COLOR_BRIGHT);
-		int radius = 2;
-		if (leveltype == DTYPE_HELL)
-			radius = 5; //hell stairs are bigger, detect all segments
-		for (int j = 0; j < numtrigs; j++) {
-			if (abs(trigs[j]._tx - ((dunx << 1) + 16)) <= radius && abs(trigs[j]._ty - ((duny << 1) + 16)) <= radius) {
-				switch (trigs[j]._tmsg) {
-				case WM_DIABTWARPUP:
-					DrawEntranceWarp(out, sx, sy);
-					break;
-				case WM_DIABNEXTLVL:
-					DrawEntranceNext(out, sx, sy);
-					break;
-				case WM_DIABPREVLVL:
-					DrawEntrancePrev(out, sx, sy);
-					break;
-				default:
-					SDL_Log("UNKNOWN STAIRS: %d", trigs[j]._tmsg);
-					break;
+		if(sgOptions.Gameplay.bRecolorStairs){
+			int radius = 2;
+			if (leveltype == DTYPE_HELL)
+				radius = 5; //hell stairs are bigger, detect all segments
+			for (int j = 0; j < numtrigs; j++) {
+				if (abs(trigs[j]._tx - ((dunx << 1) + 16)) <= radius && abs(trigs[j]._ty - ((duny << 1) + 16)) <= radius) {
+					switch (trigs[j]._tmsg) {
+					case WM_DIABTWARPUP:
+						DrawEntranceWarp(out, sx, sy);
+						break;
+					case WM_DIABNEXTLVL:
+						DrawEntranceNext(out, sx, sy);
+						break;
+					case WM_DIABPREVLVL:
+						DrawEntrancePrev(out, sx, sy);
+						break;
+					default:
+						SDL_Log("UNKNOWN STAIRS: %d", trigs[j]._tmsg);
+						break;
+					}
 				}
 			}
 		}
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -12,6 +12,7 @@
 #include "player.h"
 #include "setmaps.h"
 #include "utils/ui_fwd.h"
+#include "trigs.h"
 
 namespace devilution {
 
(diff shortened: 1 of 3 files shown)


AJenbo:
image

galaxyhaxz:
Diablo doesn't like the color green :P

qndel:
Added it as an option in ini, turned off by default, I'm seriously not going to waste more time on this, if people don't like the colors/whatever, they are free to modify them.

FitzRoyX:
Can we see ajenbos idea scaled to hell size stairs? Im not currently a huge fan of the automap going full rainbow.

While the direction wasnt pointing directly up or down, I thought it was fairly clear and didnt deviate from the original style much.

qndel:
> Can we see ajenbos idea scaled to hell size stairs? Im not currently a huge fan of the automap going full rainbow. For one, multiplayer arrows stand out better if theyre the only colored elements.
> 
> While the direction wasnt pointing directly up or down, I thought it was fairly clear and didnt deviate from the original style much.
> 
> ps. I really wonder why they thought stairs needed to be drawn as steps. Arrows of any sort are better.

Feel free to do the math required for his idea, I'm certainly not going to 🤪

Source/automap.cpp
```diff
@@ -45,10 +47,46 @@ static Sint32 AutoMapY;
 #define MAPFLAG_DIRT 0x40
 #define MAPFLAG_STAIRS 0x80
 
+/**
+ * @brief Renders a shape for entrance to next level at the specified screen coordinates.
+ */
+static void DrawEntranceNext(CelOutputBuffer out, Sint32 sx, Sint32 sy)
```
obligaron, line 53: The 4x `DrawLineTo` Sourcecode seams to be always identical.
Could we pass the color as Parameter to the method?
Rename it to DrawEntrace and use it everywhere (also at the original location).

Anyway, I like the changes. 🙂

qndel: Sure, but i kind of wanted it to be a placeholder for defining custom shapes later if we manage to come to some agreement 😅

NikoVP:
Maybe add a town/archbishop lazarus portals to the map and poisoned water supply/leorics chamber to map passage  maybe with a recolored version of door

qndel:
> Maybe add a town/archbishop lazarus portals to the map and poisoned water supply/leorics chamber to map passage maybe with a recolored version of door

it's not easy to do and goes outside the scope of this PR ;)

julealgon:
Adding new colored elements really clashes with the original style imho, so I'd prefer for it to not be color-based.

Besides, the colors still provide zero semantic value: you can't tell which is which intuitively.

qndel:
> Adding new colored elements really clashes with the original style imho, so I'd prefer for it to not be color-based.
> 
> Besides, the colors still provide zero semantic value: you can't tell which is which intuitively.
let me know when you find colors that mean up/down/warp to town/quest stairs 👍

julealgon:
> let me know when you find colors that mean up/down/warp to town/quest stairs 👍

That's exactly why i think using colors is not the way to go. "Finding the color" doesn't make sense to me, it's the wrong solution to the problem.

qndel:
That's why it's a toggle, for me the colors are fine. Red = danger = go deeper into the dungeon, blue = gamer girl bathwater = safe = go up ...

julealgon:
I just think this solution doesn't scale properly to having more stairs on the map, and it will also conflict with implementing a different color for unexplored sections on the map when the map revealing shrine is activated.

galaxyhaxz:
I know ya'll are probably tired of suggestions at this point. But I think this makes it self-explanatory and it'd be very easy to implement.
Untitled

qndel:
> I know ya'll are probably tired of suggestions at this point. But I think this makes it self-explanatory and it'd be very easy to implement.
> Untitled

aaand how will you deal with multitile stairs like caves/hell? : )

galaxyhaxz:
You'd have to draw it on the middle piece. :P

julealgon:
What about drawing a fancier shape, something like this (up stairs example):
image

(this is the extent of my paint skills @AJenbo )

qndel:
as I've said, feel free to do the math for any fancy shape you want, I'm not doing it

galaxyhaxz:
I don't necessarily think the colors idea is bad, though I'd probably do like entrance staircase stays neutral color and the "revealed" staircases for the next level/set level are different colors. Gamer guurrrl bath water could be used for set levels, since they contain good things :D

julealgon:
Just do this, but in ultra low res 😆 
image

FitzRoyX:
> I know ya'll are probably tired of suggestions at this point. But I think this makes it self-explanatory and it'd be very easy to implement.
> Untitled

I'm down for any variety of arrow options. I don't care about the town stairs or the sublevels, I just want to differentiate up and down.

AJenbo:
> I know ya'll are probably tired of suggestions at this point. But I think this makes it self-explanatory and it'd be very easy to implement.
> Untitled

I get the A for assending, but what does the V stand for?

galaxyhaxz:
> I get the A for assending, but what does the V stand for?

It stands for Vacant. As in, go down to the next level and make it vacant :P

AJenbo:
> Just do this, but in ultra low res laughing
> image

image

julealgon:
> > Just do this, but in ultra low res laughing
> > image
> 
> image

I'm honestly fine with that.

AJenbo:
image

julealgon:
That looks pretty darn good.

galaxyhaxz:
Great idea :P Did you get the code working for it, or is that a microsoft paint mockup? :P

AJenbo:
> Great idea :P Did you get the code working for it, or is that a microsoft paint mockup? :P

I don't have a license for MS Paint, used GIMP instead.

image

FitzRoyX:
Hell versions of ajenbos stairs. I think I did the sizes right
hellstairsfinalup2

malvarenga123:
What would be the symbol to the Chamber of Bone? And would there be any distinction to up to town stairs and up to upper level stairs?

FitzRoyX:
> What would be the symbol to the Chamber of Bone? And would there be any distinction to up to town stairs and up to upper level stairs?

In vanilla, chamber is surrounded by pillars and town stairs are right next to the other up stairs.

julealgon:
> ...and town stairs are right next to the other up stairs

This is incorrect @FitzRoyX . Town stairs and up level stairs spawn randomly on the map.

Issue #1566, opened 2021-04-17
Holy Shrine teleports out of bounds with Phasing or crashes the game


serpinio:
**Important information**
Using latest DevilutionX 1.2.1 on Switch and PC with Diablo and Hellfire MPQs.

**Description**
Discovered a Holy Shrine - "Wherever you go, there you are", that casts Phasing. If used, depending on the client platform it either crashes the game (Switch, firmware 10.2.0) or teleports the character out of bounds (PC, Win7/10).

**To Reproduce**
1. Load save
2. Use nearby shrine

**Expected behavior**
Phasing should only happen within the bounds of the level.

**Screenshots**
PC version screens (before, after) and the save file attached.

![before](https://user-images.githubusercontent.com/82693390/115101742-60312b80-9f4f-11eb-85cc-b3d09eae9dc8.jpg)
![after](https://user-images.githubusercontent.com/82693390/115101744-62938580-9f4f-11eb-89e8-ff2e067700c8.jpg)

[single_0_holy_shrine_out_of_bounds.zip](https://github.com/diasurgical/devilutionX/files/6328624/single_0_holy_shrine_out_of_bounds.zip)

FitzRoyX:
Shoulda been called Wandering Shrine smh

qndel:
![image](https://user-images.githubusercontent.com/14297035/115111433-a2269580-9f80-11eb-8670-3b7a308edc04.png)
That's the save you sent, are you sure you sent the correct one?

qndel:
```cpp
void AddRndTeleport(int mi, int sx, int sy, int dx, int dy, int midir, char mienemy, int id, int dam)
{
	int pn, r1, r2, nTries;

	nTries = 0;
	do {
		nTries++;
		if (nTries > 500) {
			r1 = 0;
			r2 = 0;
			break; //BUGFIX: warps player to 0/0 in hellfire, change to return or use 1.09's version of the code
		}
		```
		this is the cause ;)

serpinio:
> 
> 
> ![image](https://user-images.githubusercontent.com/14297035/115111433-a2269580-9f80-11eb-8670-3b7a308edc04.png)
> That's the save you sent, are you sure you sent the correct one?

That's certainly _not_ what I sent. Re-sending it again just in case.

[single_0_shrine_crash.zip](https://github.com/diasurgical/devilutionX/files/6329156/single_0_shrine_crash.zip)

[...]

PR #1568, opened 2021-04-17
fix phasing

qndel:
fixes https://github.com/diasurgical/devilutionX/issues/1566
Increasing number of tries isn't a proper solution because in the situation from https://github.com/diasurgical/devilutionX/issues/1566 there's simply no valid position (notice that the room is in the corner of the map and pretty narrow) so the only reasonable thing to do is just do nothing ;)

+I consider returning early to be fair - doesn't consume mana in case of a failure


AJenbo:
Would it make sense to map out the valid positions and then just do a single random call to pick one of them instead of doing 500 failed attempts?

That should make the spell exec time constant instead of variable.

Chance4us:
Wouldn't it be good to merge this important bugfix? Also #1583.

PR #1570, opened 2021-04-17
Allow (C++14 || C++17)

Jmgr:
Will this work? Let's see :)


glebm:
While it works on CI, some toolchains do not yet support C++17.

Notably:

1. The older OpenDingux toolchain (used for the RG350 and GKD350h ports, though we'll switch to OpenDingux Beta at some point)
2. The Xbox SDK (https://github.com/XboxDev/nxdk) - not ported yet

Jmgr:
Can we at least use C14 then?

glebm:
I think both the NDK and the older OD toolchains are C++11

Jmgr:
I have just tested with NXDK, adding `-std=c++14` to `NXDK_CXXFLAGS` and compiling the hello++ sample works. Will have a look at OpenDingux.

Jmgr:
Where do you see that NXDK doesn't support C++17 @glebm? I've just tried compiling a sample project while adding `-std=c++17` to the build flags and it builds without any issue.

Jmgr:
And OpenDingux compiles without any issue with `mipsel-linux-g++ main.cpp -std=c++17 -o test.dge`. Am I missing something?

galaxyhaxz:
Don't you have to actually use a feature from a specific standard for it to fail? If DevilutionX is only using C++ <= 11 features then it wouldn't cause issues no?

Jmgr:
No, if a compiler accepts that flag then it should implement all the features included in the corresponding standard. That is true for Clang and GCC. And MSVC 2019 [supports](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-160) all C++17 features.

AJenbo:
Cool that you are looking into this.

glebm:
> No, if a compiler accepts that flag then it should implement all the features included in the corresponding standard. That is true for Clang and GCC. And MSVC 2019 supports all C++17 features.

That's not how it works. For example, GCC 5 will accept `-std=c++17` but is missing a few features.

C++17 is only fully supported from GCC 7 http://www.gnu.org/software/gcc/projects/cxx-status.html#cxx17

I'd be OK with upgrading the RG350 buildroot (https://github.com/diasurgical/devilutionX/pull/1577) and dropping support for GKD350h, but I don't know about nxsdk and simply passing `-std=c++17` is not sufficient tyo check.

glebm:
I guess we could ignore nxsdk given that there isn't actually a port for it yet
Adding C++17 support isn't that hard once you have full C++11 support

glebm (approved):
Let's go straight for 17

AJenbo:
@glebm I'll let you merge this :)

Jmgr:
> That's not how it works, unfortunately. For example, GCC 5 accepts `-std=c++17` but is missing a few features. C++17 is only fully supported from GCC 7 http://www.gnu.org/software/gcc/projects/cxx-status.html#cxx17

My bad, forgot about that. :facepalm:

AJenbo:
Wasn't quite incontroversial, but you got one 😅

julealgon:
> but you got one

And this is pretty huge... lots of useful stuff on 14 and 17

AJenbo:
Yeah really nice job investigating, upgrading and deprecating :)

Jmgr:
:fist:

PR #1571, opened 2021-04-17
String format improvements

Jmgr:
This PR adds the [fmt](https://github.com/fmtlib/fmt) library as a dependency and some examples to fuel a most certainly heated discussion (Hell's daemons would certainly approve!).

We have already partially discussed this on the late [logger PR](https://github.com/diasurgical/devilutionX/pull/1464), but this time I would like to focus on the benefits of using a C++ formatting library to replace the C-style *printf functions.

The issue I have with *printf functions is that they are not type safe (this can be [mitigated](https://github.com/diasurgical/devilutionX/pull/1481) somewhat) and not extensible. You can't add new types, and C++ standard types (std::string, std::vector, etc.) won't work on them.

Then comes `fmt`, which has a few interesting features:
- simple positional arguments without the need to specify a type (like `I would like to have {} gold for that item please.`)
- named arguments, nice for translated strings (like `This item costs {gold} gold.`)
- uses a very complete [format](https://fmt.dev/latest/syntax.html)
- allows user-defined types like enums, classes, etc.
- high performance, [faster](https://github.com/fmtlib/fmt#speed-tests) than the C *printf functions
- small (minimum config is only 3 headers)

I have added some examples in this PR, and a small logger that wraps the SDL logger functions.


Source/DiabloUI/selgame.cpp
```diff
@@ -412,10 +414,11 @@ static bool IsGameCompatible(GameData *data)
 	if (data->programid != GAME_ID) {
 		UiSelOkDialog(title, "The host is running a different game than you.", false);
 	} else {
-		char msg[64];
-		sprintf(msg, "Your version %s does not match the host %d.%d.%d.", PROJECT_VERSION, PROJECT_VERSION_MAJOR, PROJECT_VERSION_MINOR, PROJECT_VERSION_PATCH);
-
-		UiSelOkDialog(title, msg, false);
+		// An example of using a format with references. Note the references that allow a localized string to change the order of the inserted parameters.
+		// Note that named insertions would probably be better here, but this is just to provide an example.
+		auto msg = fmt::format("Your version {0} does not match the host {1}.{2}.{3}.", PROJECT_VERSION, PROJECT_VERSION_MAJOR, PROJECT_VERSION_MINOR, PROJECT_VERSION_PATCH);
```
julealgon, line 419: This is _so_ much better. No more fixed array shenanigans.

CMakeLists.txt
```diff
@@ -145,7 +149,7 @@ if(PIE)
   set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 endif()
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
julealgon, line 152: This is a part of another PR to migrate to C++14/17, right?

Source/utils/format.hpp
```diff
@@ -0,0 +1,57 @@
+#pragma once
+
+#include "../player.h"
+
+#include <fmt/format.h>
+
+namespace fmt {
+
+template <>
+struct formatter<devilution::PLR_MODE> : formatter<string_view> {
```
julealgon, line 10: This is a very nice way of providing an extension point.

Source/utils/log.hpp
```diff
@@ -0,0 +1,91 @@
...
+    Info = SDL_LOG_PRIORITY_INFO,
+    Warn = SDL_LOG_PRIORITY_WARN,
+    Error = SDL_LOG_PRIORITY_ERROR,
+    Critical = SDL_LOG_PRIORITY_CRITICAL,
+};
+
+template <typename... Args>
+void Log(const char *fmt, Args &&... args)
+{
+	auto str = fmt::format(fmt, std::forward<Args>(args)...);
```
julealgon, line 38: Interesting to see an example of perfect forwarding as I've learned about this yesterday 😆

julealgon (commented):
Really cool stuff.

This post seem to indicate something like fmt will be available natively on C++20 though:
https://stackoverflow.com/a/57128991/1946412

Is that the case? Would we just then change `fmt::` to `std::` when/if migration to C++20+ happens?

Source/DiabloUI/selgame.cpp
galaxyhaxz: How would this work if you wanted to pad a number? E.g.
`printf("The percent is %02d", x);`

Iniquitatis: @galaxyhaxz, `fmt::print("The percent is {:02d}", x);`
See: https://fmt.dev/latest/syntax.html#format-examples

Jmgr:
> Is that the case? Would we just then change `fmt::` to `std::` when/if migration to C++20+ happens?

Yes, at least partially. I haven't looked into details though, but yeah.

CMakeLists.txt
Jmgr: This seems to be required when you add a custom type, I haven't managed to get that part of the code to compile using only C++11.

Source/utils/log.hpp
Jmgr: Yeah it's not the most intuitive part of C++...

Jmgr:
Some issues with the Switch/Vita SDKs not having `fileno`. It seems to have been voluntarily removed: https://github.com/vitasdk/newlib/blob/34ca70545302f63dcb15ffd3ea63894a8d224028/newlib/libc/include/stdio.h#L696 with `FIXME: must initialize stdio first, use fn`. :thinking:

Iniquitatis:
> Is that the case? Would we just then change fmt:: to std:: when/if migration to C++20+ happens?

> Yes, at least partially. I haven't looked into details though, but yeah.

@julealgon, @Jmgr, if compatibility with C++20 STL is mandatory, you could just restrict usage of fmt to the compatible feature subset. There are not too many differences anyway.
https://en.cppreference.com/w/cpp/utility/format

Jmgr:
> > Is that the case? Would we just then change fmt:: to std:: when/if migration to C++20+ happens?
> 
> > Yes, at least partially. I haven't looked into details though, but yeah.
> 
> @julealgon, @Jmgr, if compatibility with C++20 STL is mandatory, you could just restrict usage of fmt to the compatible feature subset. There are not too many differences anyway.
> [en.cppreference.com/w/cpp/utility/format](https://en.cppreference.com/w/cpp/utility/format)

Sure. The differences are listed [there](https://fmt.dev/dev/api.html#compatibility-with-c-20-std-format). Not sure what `Width calculation doesn’t use grapheme clusterization` but that doesn't sound critical to me, and `Chrono formatting doesn’t support C++20 date types` either.

Jmgr:
PR created [upstream](https://github.com/vitasdk/newlib/pull/51). In the meantime I suppose we could use my [fork](https://github.com/Jmgr/newlib) of vitasdk? We seem to be using a Docker container for the tests though, so I'll have to see how to generate one.

CMakeLists.txt
```diff
@@ -43,6 +43,10 @@ option(DEVILUTIONX_SYSTEM_LIBSODIUM "Use system-provided libsodium" ON)
 cmake_dependent_option(DEVILUTIONX_STATIC_LIBSODIUM "Link static libsodium" OFF
                        "DEVILUTIONX_SYSTEM_LIBSODIUM AND NOT DIST" ON)
 
+option(DEVILUTIONX_SYSTEM_LIBFMT "Use system-provided libfmt" OFF)
```
glebm, line 46: This should probably default to `ON`

* Provided by `libfmt-dev` on Ubuntu and Debian
* Provided by `fmt` on Mac (`brew install fmt`)

Source/utils/log.hpp
```diff
@@ -0,0 +1,91 @@
...
+	SDL_Log("%s", str.c_str());
+}
+
+template <typename... Args>
+void LogVerbose(LogCategory category, const char *fmt, Args &&... args)
```
glebm, line 43: Can we default to `Application` throughout?

Source/init.cpp
```diff
@@ -160,6 +161,9 @@ void init_archives()
 	paths.push_back("/usr/local/share/diasurgical/devilutionx/");
 #endif
 
+	// An example of logging a vector of strings.
+	Log("Init: paths = {}", paths);
```
glebm, line 165: This is already logged below with `SDL_LogVerbose`; it shouldn't be logged twice.

CMakeLists.txt
```diff
@@ -165,6 +169,13 @@ if(NOT NONET)
   endif()
 endif()
 
+if(DEVILUTIONX_SYSTEM_LIBFMT)
+  set(fmt_USE_STATIC_LIBS ${DEVILUTIONX_STATIC_LIBFMT})
+  find_package(fmt REQUIRED)
```
glebm, line 174: For this to work, we need to provide a finder module at `CMake/Findfmt.cmake`
Example https://github.com/ceph/ceph/blob/master/cmake/modules/Findfmt.cmake

CMakeLists.txt
```diff
@@ -43,6 +43,10 @@ option(DEVILUTIONX_SYSTEM_LIBSODIUM "Use system-provided libsodium" ON)
 cmake_dependent_option(DEVILUTIONX_STATIC_LIBSODIUM "Link static libsodium" OFF
                        "DEVILUTIONX_SYSTEM_LIBSODIUM AND NOT DIST" ON)
 
+option(DEVILUTIONX_SYSTEM_LIBFMT "Use system-provided libfmt" OFF)
+cmake_dependent_option(DEVILUTIONX_STATIC_LIBFMT "Link static libfmt" OFF
+                       "DEVILUTIONX_STATIC_LIBFMT AND NOT DIST" ON)
```
glebm, line 48:
```suggestion
                       "DEVILUTIONX_SYSTEM_LIBFMT AND NOT DIST" ON)

```

glebm (changes requested):

CMakeLists.txt
Jmgr: The issue is that the `libfmt-dev` package is [quite old](https://packages.ubuntu.com/search?keywords=libfmt-dev) on Ubuntu 18.04 (4.0.0). The latest version is 7.1.3. There have been some breaking changes since then. Ubuntu 18.04 is still supported until April 2023.

What we could do is check the version that is available on the system, and if its < 7.0.0 then download `fmt` using CMake?

glebm: > What we could do is check the version that is available on the system, and if its < 7.0.0 then download fmt using CMake?

This sounds good. This can be done by replacing `find_package(fmt REQUIRED)` with `find_package(fmt 7.0.0)` and checking `fmt_FOUND`.

Source/utils/log.hpp
Jmgr: Sure.

Source/init.cpp
Jmgr: Yes this was only an example of logging a std::vector.

3rdParty/libfmt/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+  if(DEVILUTIONX_STATIC_LIBFMT)
+    set(BUILD_SHARED_LIBS OFF)
+  else()
+    set(BUILD_SHARED_LIBS ON)
+  endif()
+  include(FetchContent)
+  FetchContent_Declare(fmt
+      GIT_REPOSITORY https://github.com/fmtlib/fmt.git
+      GIT_TAG 7bdf0628b1276379886c7f6dda2cef2b3b374f0b
```
glebm, line 14: Looks like fmt does not use git submodules. In that case, it's a lot faster to download a zip archive:

```suggestion
        URL https://github.com/fmtlib/fmt/archive/7bdf0628b1276379886c7f6dda2cef2b3b374f0b.zip
        URL_HASH MD5=70e0a0beda4bdac7debe60d380e1d802
```

glebm (changes requested):

CMakeLists.txt
Jmgr: That one is distributed under the LGPL license. I suppose that would be an issue? Looking for another one. A bit odd that `fmt` doesn't provide one.

glebm: `fmt` does provide a cmake config but not all distributions install it

```
$ dpkg -L libfmt-dev | grep cmake
/usr/lib/x86_64-linux-gnu/cmake
/usr/lib/x86_64-linux-gnu/cmake/fmt
/usr/lib/x86_64-linux-gnu/cmake/fmt/fmt-config-version.cmake
/usr/lib/x86_64-linux-gnu/cmake/fmt/fmt-config.cmake
/usr/lib/x86_64-linux-gnu/cmake/fmt/fmt-targets-none.cmake
/usr/lib/x86_64-linux-gnu/cmake/fmt/fmt-targets.cmake
```

Both Debian/Ubuntu and buildroot fmt packages provide these files, so perhaps it's OK to not worry about Findfmt

AJenbo:
Fails on Vita :(

Jmgr:
> Fails on Vita :(

Yeah, looks like a bug in the vitasdk, see https://github.com/diasurgical/devilutionX/pull/1571#issuecomment-821966738 and https://github.com/diasurgical/devilutionX/pull/1571#issuecomment-821985079.

Jmgr:
My plan on this is to first fix the build on Vita, rebase on master, replace the log call, then replace all sprintf calls. Alternatively I can also just add the `fmt` dependency and the log function wrappers in this PR, and then create another PR for the replacements. Comments welcome :)

AJenbo:
I would suggest saving the `sprintf` stuff for a second round.

Jmgr:
Sounds good. Just `fmt` and the log calls then.

AJenbo:
Oh wow, that's a bit of an effort for fixing this :D Hope you can upstream the Vita fix :)

Jmgr:
> Oh wow, that's a bit of an effort for fixing this :D Hope you can upstream the Vita fix :)

I hope so too, but we might have to fork and create our own Docker image at some point. Let's see.

I've created my own image using the [buildscripts](https://github.com/vitasdk/buildscripts) repository, but it does not include any library like SDL2. The official Docker image does though. I've created an [issue](https://github.com/vitasdk/buildscripts/issues/92) to ask how to do that.

Source/utils/format.hpp
```diff
@@ -0,0 +1,57 @@
...
+	template <typename FormatContext>
+	auto format(devilution::PLR_MODE mode, FormatContext &ctx)
+	{
+		string_view name = "unknown";
+		switch (mode) {
+		case devilution::PM_STAND:
```
glebm, line 16: If you move this to a `static std::string_view GetPlrModeName(devilution::PLR_MODE mode)` the `switch` can be simplified:

```c++
static std::string_view GetPlrModeName(devilution::PLR_MODE mode) {
  switch (mode) {
    case devilution::PM_STAND: return "stand";
    case devilution::PM_WALK: return "walk (N, NW, or NE)";
    ...
    default: return "unknown";
  }
}
```

Jmgr: Indeed. I this should go next to the enum declaration as it can be used in other places. That should also be `contrexpr`. Not sure about the name though. Maybe migrate the enums to `enum class` and then have one overloaded function called `toString`.

Jmgr: I've created https://github.com/diasurgical/devilutionX/pull/1674 for this, but since `string_view` is missing from the Amiga cosstools I will merge that one after this PR. This PR will only include `fmt` and the logging functions update.

glebm (approved):

glebm:
Missing CI stuff:

1. Brewfile should be updated to add `fmt` so that we use the "system" dependency there.
2. debian-testing builds should `apt-get install libfmt-dev` for the same reason https://packages.debian.org/sid/libfmt-dev

AJenbo (approved):

AJenbo:
Nice timing, going to be good to have this with basic translation support also being ready.

AJenbo:
Merged manually, was a bit lazy about it so the cleanup/fix might not have been merged with the one they correct :shrug: 
https://github.com/diasurgical/devilutionX/commit/f3a401b5f679c665dc8b2bdf5ec3d2296fae8c07
https://github.com/diasurgical/devilutionX/commit/af0ea6fdfc667216b9a5f863c56e24a97094e708
https://github.com/diasurgical/devilutionX/commit/bc8f586893d49c25b9653bca611f1ad161a989ff
https://github.com/diasurgical/devilutionX/commit/340d111963ea68fa67af08be2aea0f7a0b199f78
https://github.com/diasurgical/devilutionX/commit/ef0ca794bc24f5439a3466db2ec54f03956e4ff0

PR #1574, opened 2021-04-17
Add option to warn about Black Death zombies

Eider-McDuck:


julealgon:
Can you please provide more details about this change? Pictures of how it looks like, a description, etc.

julealgon:
Additionally, what if i want to be warned of illusion weavers as well? What if i want to be warned of the dogs boss pack?

qndel:
> Can you please provide more details about this change? Pictures of how it looks like, a description, etc.

just a text rofl

FitzRoyX:
Isn't seeing them warning enough? Slowest monsters in the game.

Tremain84:
A simple change would be to make Black Deaths take away 1 Vitality per hit. It is still a permanent effect and punishing players but it can be "cured" with level ups and/or elexir of vitality.
This would make Black Deaths still very dangerous, but it doesn't completely ruins your character, if you manages to hit you.

qndel:
> Isn't seeing them warning enough? Slowest monsters in the game.

people usually don't know it takes 1 hit from them to ruin your character when they see them for the first time ...

galaxyhaxz:
Yep, it was a poor design choice imo. If you get hit enough it'd lower you life to unusable levels once you hit the vitality stat cap.

Eider-McDuck:
This does the following: If you enter a level with Black Death Zombies, this simply displays an additional line "Black Death" under the Name, Password and Level lines, also in the same color, so it is unintrusive.

I see two alternatives:
1. A suitable icon, maybe a skull, displayed in the top right corner instead of the text, or
2. If the game_init_info seed of a new game is computed to result in Black Death Zombies appearing in any level, a new seed is tried until they do not appear. In single player this seems to be a bit complicated.

galaxyhaxz:
> If the game_init_info seed of a new game is computed to result in Black Death Zombies appearing in any level, a new seed is tried until they do not appear. In single player this seems to be a bit complicated.

At that rate why not just add an option in the INI file to disable black death's from removing life and instead remove a stat point?

Eider-McDuck:
Well, one could argue that the behaviour of removing life, but not the corresponding vitality points, was actually a bug (the same for the shrines removing mana without the corresponding magic points). In that case, we could add an option to "fix" the bug.

julealgon:
This is a mod change.

Eider-McDuck:
The same could be said about fixing the stunlock bug etc. But I also see your point, which is why I think a nice warning is the right way to go.

But I am also open to being convinced that removing life without vitality is actually a bug, I think actually many players saw it that way.

FitzRoyX:
The warrior can equip a bow or learn some basic spells to deal with them. I agree that it's stupid to fuck over melee classes with monsters like this, but... the problem isn't knowing whether they've spawned, it's that newbie's don't know what they can do. It would be better to edit monster's mouseover description with "Permanently reduces your life".

qndel:
> The warrior can equip a bow or learn some basic spells to deal with them. I agree that it's stupid to fuck over melee classes with monsters like this, but... the problem isn't knowing whether they've spawned, it's that newbies don't know what they can do. It would be better to edit monster's mouseover description with "Permanently reduces your life".

Not sure if it's like this in vanilla because I mostly play crucible, but they are slow enough you can land a hit or two and move back without them having a chance to attack you so it should be possible to deal with them as melee too

Eider-McDuck:
Cleaner version: In the automap now the warning is displayed as "Level: 3 (Black Death)", which looks cleaner. Moreover, the mouseover information of black death zombies is displayed in red and has an additional line "Drains maximum life".

Eider-McDuck:
bd

AJenbo:
I like the added text in there description

julealgon:
> I like the added text in there description

There is not enough room for an extra line in the monster information in a generic manner.

If a monster has resistances, immunites, and the player has killed enough of them to show their HP, then the view already contains the maximum of 5 lines.

I do not agree with special casing black deaths at all here.

Also, the fact that the description is red is inconsistent.

Eider-McDuck:
Yes, that is indeed a problem. I am also not convinced about the red. I removed these changes for now, but left the unintrusive warning in the level display.

PR #1575, opened 2021-04-17
Prevent player from queuing up a spell which would cause mana to drop below 0

FluffyQuack:
Did a brief test in SP and MP and this seems to work fine.

The check if the player hasn't finished casting a spell yet is very ugly, though.

This is one potential way the player could still result in negative health with this solution:
- Get hit while mana shield is active after queueing up spell (mana check only happens once)

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -3898,7 +3898,7 @@ void CheckPlrSpell()
 	switch (plr[myplr]._pRSplType) {
 	case RSPLTYPE_SKILL:
 	case RSPLTYPE_SPELL:
-		addflag = CheckSpell(myplr, rspell, plr[myplr]._pRSplType, false);
+		addflag = CheckSpell(myplr, rspell, plr[myplr]._pRSplType, false, true);
 		break;
 	case RSPLTYPE_SCROLL:
 		addflag = UseScroll();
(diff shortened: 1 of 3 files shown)


galaxyhaxz:
This was one of the worst bugs in the game. Great job tackling it! It's also easy to double cast a spell when a monster dies, and the player will cast again thinking there is a valid target.

FluffyQuack:
> This was one of the worst bugs in the game. Great job tackling it! It's also easy to double cast a spell when a monster dies, and the player will cast again thinking there is a valid target.

This PR is more of a bandaid than a full fix as there's at least one scenario where you could still result in negative mana (julealgon pointed that out and I added it to the PR text).

Fixing player doing a queued up attack against a dead monster would definitely be nice to fix. I've added the ability to hold down the button to attack in my own fork, and then it's especially easy to keep doing extra, redundant attacks as enemies die.

julealgon:
> It's also easy to double cast a spell when a monster dies, and the player will cast again thinking there is a valid target.

> Fixing player doing a queued up attack against a dead monster would definitely be nice to fix.

@FluffyQuack and @galaxyhaxz , FYI
https://github.com/diasurgical/devilutionX/pull/962

Issue #1582, opened 2021-04-18
Using caves town entrance to go to lvl 8


qndel:
**Important information**
Doesn't matter, vanilla bug

**Describe**
http://www.lurkerlounge.com/forums/thread-12422.html
If you enter the caves via the cave mouth entrance near Wirt, you can run the mouse cursor over the stairs on level 9, and a small spot on the right side of the stairs will read "up to level 8." If you left click on that spot your character will begin walking towards the level 8 stairs (even if you can't see them on the map), unless the way is blocked by lava, grates, or some other obstacle. If you teleport onto that spot, your character will teleport all the way to the level 8 stairs and go up!

**To Reproduce**
1. Go to caves using town entrance
2. Cast teleport on a specific tile on the right side of the stairs (which says "up to level 8")
3. Say hello to level 8

**Expected behavior**
This shouldn't happen

**Screenshots**
![caveswarp](https://user-images.githubusercontent.com/14297035/115143481-e466da00-a047-11eb-9a66-fad69bb7ea25.gif)

Chance4us:
At this point I very often had the feeling that something was wrong with the description.

julealgon:
I noticed this in one stream (don't remember who was playing) where i saw a description change in the stairs for a brief moment, and had taken note to look into what happened there.

Very nice that you figured the bug out.

sumyunggai:
I don't know if it adds anything, as there's a PR to address this, but attached is a savegame showing this bug. I built from source as of early AM East Coast US, May 20. Win-x64.

![cavebug](https://user-images.githubusercontent.com/39627597/119046988-59049f80-b98b-11eb-8ea5-25243c235239.gif)

[single_1_caveBug.zip](https://github.com/diasurgical/devilutionX/files/6518653/single_1_caveBug.zip)

qndel:
thanks but the fix is complete and the cause of the bug is known, no need for a savegame : P

PR #1583, opened 2021-04-18
fix cave stairs

qndel:
Fixes https://github.com/diasurgical/devilutionX/issues/1582
dPiece 182 was the cause of that problem.
It occurs both in warp to town and level up and only warp to town was checking distance.

```cpp
/** Specifies the dungeon piece IDs which constitute stairways leading up from the caves. */
int L3UpList[] = { 170, 171, 172, 173, 174, 175, 176, 177, 178, 179, 180, 181, 182, 183, -1 };
```
Whenever you put the mouse over dPiece 182 in stairs to town, the code matches that dPiece with "level up" trigger and replaces cursmx/cursmy without any check - that's why teleporting/clicking actually targetted the stairs up

While this could be applied to all entrances, I've only used it for caves one because that's the only one people reported problems with. And at the moment I'm not confident enough this wouldn't change the behavior of hell stairs which are 3x2 (I needed a 5 radius check for drawing all hell stairs tiles on automap ...)

### DIFF
--- a/Source/trigs.cpp
+++ b/Source/trigs.cpp
@@ -565,9 +565,13 @@ bool ForceL3Trig()
 				sprintf(infostr, "Up to level %i", currlevel - 1);
 				for (j = 0; j < numtrigs; j++) {
 					if (trigs[j]._tmsg == WM_DIABPREVLVL) {
-						cursmx = trigs[j]._tx;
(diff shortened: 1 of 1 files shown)


Chance4us:
Wouldn't it be good to merge this important bugfix? Also #1568.

PR #1587, opened 2021-04-18
🧹 C++17: Use std::make_unique

glebm:


AJenbo (approved):
Anything that reduces the number of meta words is ok by me :+1:

julealgon (commented):
Nice.

Issue #1589, opened 2021-04-18
Golems fight each other


qndel:
**Important information**
Vanilla bug

**Describe**
Golems fight each other

**To Reproduce**
1. Get 2 players to summon 2 golems
2. Watch them fight

**Expected behavior**
Golems should either ignore each other or respect friendly fire settings

PR #1590, opened 2021-04-18
Stops golems from fighting each other

qndel:
Fixes https://github.com/diasurgical/devilutionX/issues/1589


AJenbo:
Shoudn't this only take effect if frindly mode = true?

qndel:
nah, it's useless for pvp anyway as golems never target players

PR #1592, opened 2021-04-18
:truck: Turns inventory into class

yuripourre:
This PR moves all logic related to inventory to a class, this will help a lot modders that want to have custom UI, since would be possible to override drawing methods and hover methods.

This is needed for couch-coop.


AJenbo (dismissed):
Looks good, but considering that I'm a bit tipsy atm I would prefer that @julealgon or sober me give it a sound look.

Source/inv.h
```diff
@@ -78,54 +78,80 @@ extern bool invflag;
...
-bool AutoPlaceItemInInventory(int playerNumber, const ItemStruct &item, bool persistItem = false);
-bool AutoPlaceItemInInventorySlot(int playerNumber, int slotIndex, const ItemStruct &item, bool persistItem);
-bool AutoPlaceItemInBelt(int playerNumber, const ItemStruct &item, bool persistItem = false);
-bool GoldAutoPlace(int pnum);
-void CheckInvSwap(int pnum, BYTE bLoc, int idx, WORD wCI, int seed, bool bId, uint32_t dwBuff);
-void inv_update_rem_item(int pnum, BYTE iv);
+class Inventory {
+
+private:
+	bool AutoEquip(int playerNumber, const ItemStruct &item, inv_body_loc bodyLocation, bool persistItem);
```
glebm, line 84: Usually private methods are declared after the public methods

Source/inv.h
```diff
@@ -78,54 +78,80 @@ extern bool invflag;
...
+	void CheckInvCut(int pnum, int mx, int my, bool automaticMove);
+	void CheckInvPaste(int pnum, int mx, int my);
+	void CheckQuestItem(int pnum);
+	void CleanupItems(ItemStruct *item, int ii);
+	bool FitsInBeltSlot(const ItemStruct &item);
+	int SwapItem(ItemStruct *a, ItemStruct *b);
+	bool WeaponAutoPlace(int pnum);
+
+public:
+	void FreeInvGFX();
```
glebm, line 99: Let's remove `Inv` and `inv_` in method names?

glebm (changes requested):

Source/inv.h
yuripourre: Thanks for the suggestion, I only kept `UseInvItem` because it clashes with `items#UseItem`

yuripourre:
I messed with the rebase, please do not merge.

Source/inv.h
```diff
@@ -78,54 +78,81 @@ extern bool invflag;
 extern bool drawsbarflag;
 extern const InvXY InvRect[73];
 
-void FreeInvGFX();
-void InitInv();
+class Inventory {
```
Jmgr, line 81: This class should be `final`.

Source/inv.h
```diff
@@ -78,54 +78,81 @@ extern bool invflag;
 extern bool drawsbarflag;
 extern const InvXY InvRect[73];
 
-void FreeInvGFX();
-void InitInv();
+class Inventory {
+
+public:
+	void FreeGFX();
```
Jmgr, line 84: All the functions should be static since they don't change the class' instance.

Source/inv.h
Jmgr: And since it is only a container for functions its constructor should probably be deleted as well.

yuripourre: @Jmgr I have plans to add multiple implementations of this class. Do you think it should be final until then?

Source/inv.h
yuripourre: I thought about that but I think it would make derived classes more complex.

Perhaps, should I use this approach?

https://en.m.wikipedia.org/wiki/Curiously_recurring_template_pattern

The whole idea here is to make the UI more flexible (I plan other PRs), to allow couch coop.

```
Inventory inventory = isCouchCoop ? new CouhCoopInventory() : new Inventory();
```

I am open to suggestions, I just wanted to make the changes simple because couch coop requires lots of changes.

Source/inv.h
yuripourre: @Jmgr sorry, I didn't get what constructor are you talking about. `Init()`?

AJenbo: Could move the player inventory members to this class and attach it to the player. That would also help to clean up the player structure and possibly allow us to simplify some of the functions here.

yuripourre: @AJenbo can we merge this and I  do that in another PR? I am not sure when I will have the time to do so.

Jmgr: > @Jmgr I have plans to add multiple implementations of this class. Do you think it should be final until then?

So this class will be inherited? Then no, don't mark it as final, but do set the [destructor as virtual](https://stackoverflow.com/questions/461203/when-to-use-virtual-destructors).

Jmgr: > @Jmgr sorry, I didn't get what constructor are you talking about. `Init()`?

I meant that if the class only contains static functions and no state then constructing it makes no sense. But seeing that you intent on using inheritance that it's fine. But why use an `Init()` function instead of a constructor?

Source/inv.h
Jmgr: So the idea here is to allow to override certain inventory functions? In that case don't make the functions static, no. I'm not a huge fan of inheritance in general, but it all depends on what you want to do so it's difficult to advise now.

Source/inv.h
yuripourre: The method was originally `InitInv()` I just renamed it. I didn't convert to a constructor because the `*inventory` is being initialized still in the Main Menu and the `Init()` is only called just before the in-game screen (so it load the resources).

Source/control.cpp
```diff
@@ -2046,21 +2046,21 @@ void control_remove_gold(int pnum, int gold_index)
 		if (plr[pnum].InvList[gi]._ivalue > 0)
 			SetGoldCurs(pnum, gi);
 		else
-			RemoveInvItem(pnum, gi);
+			inventory->RemoveItem(pnum, gi);
```
julealgon, line 2049: Ok so it's not only the UI but the logic also? Then, I disagree with the way this is being handled.

The inventory is part of a player's hero, not a completely separate entity. I don't think it makes sense to have a "context-less" inventory class and then pass the player number to all calls.

Instead, this would make much more sense:
```cpp
plr[pnum].Inventory.RemoveItem(gi);
```
```

julealgon (changes requested):
I dislike the approach being taken here and, since changing it will affect a lot of the code, I'll stop reviewing for now.

See comments.

Source/control.cpp
AJenbo: If the UI gets its own class you can pass `plr[pnum].Inventory` to the UI as well.

yuripourre:
Thanks for the reviews, sorry for wasting your time. I don't have the time/energy to work on that right now, so I am closing it. Feel free to work over it or reopen.

AJenbo:
alright, hopefully thinks have been broken up a bit already once you have some more time, take care :)

Issue #1576, opened 2021-04-17
Multiplayer desyncs related to player position


FluffyQuack:
Desync cause 1:
Video example: https://www.youtube.com/watch?v=wDXCDKLnpNY

When moving after attacking, the movement becomes different on other clients (the character eventually catches up to the correct position, but it is wrong for a few seconds). The above video is recorded using DevX v1.2.1, but I also tested this using the Diablo build from GOG and the behaviour is identical there.

The cause is this: When clicking to walk, the game calls MakePlrPath() which does pathfinding to figure out a good path for the player, and it sets the _ptarx and _ptary values to be where the player ends up at the end of the pathfinding. However, when you end an attack, the game calls StartStand() which calls FixPlayerLocation() which sets _ptarx and _ptary to be the same as the player position.

The local player keeps sending player position values (including _ptarx and _ptary) to other clients so other clients and update a desynced location, but since the _ptarx and _ptary values are wrong, it ends up asking other clients to move the player to the wrong location.

Here is one way to fix this desync: https://github.com/diasurgical/devilutionX/pull/1593

Desync cause 2:
Video example: https://www.youtube.com/watch?v=eO6Ictu0Rns

This happens if you cancel pathfinding with an attack (it doesn't happen every single time. I think it happens if you attack at a certain part of the movement). And if you make sure to move immediately after the attack (which I do in the above video), then the second client won't even see the attack happen.

I didn't look too deeply into this, but I think _ptarx and _ptary are related once again. I tried a brute force fix where I set _ptarx  and _ptary to the correct values immediately before sending a packet with an attack command, and that almost entirely fixed the issue (it would still happen, but only very rarely).

Desync cause 3:
If two or more players enter a dungeon at the same time, there's a chance they'll try to spawn in the same position, and thus be in different positions on each client.

AJenbo:
I'm not sure, but this is probably one of the things that has a butterfly effect in the game.

Nice work making a good test case to make this a manageable issue to work on.

FluffyQuack:
I tried to debug this a bit, and here's my progress:

- Player 1 clicks to move after attacking and MakePlrPath() ends up being called once (with the correct information).
- Player 2 get the packet from player 1 and calls MakePlrPath() once (with the correct information), but then it also keeps calling MakePlrPath() in multi_process_network_packets() and most of the time it's with the player's old position.
[...]

PR #1593, opened 2021-04-18
Fix to network desyncs related to player movement

FluffyQuack:
This fixes the first desync issue listed here: https://github.com/diasurgical/devilutionX/issues/1576

Video example of the issue: https://www.youtube.com/watch?v=wDXCDKLnpNY

The cause of the problem is the _ptarx and _ptary values (these correspond to where the player is pathfinding to) are sometimes set incorrectly. The game keeps sending these values (and player position) to other clients so they can ensure the player position is the same across all clients, but sometimes the values are set incorrectly which makes the local player tell other clients to move the player to the wrong position.

For instance, this desync is caused by the following sequence of code:

- Player is in the middle of the attack animation.
- Player clicks to move, which makes the game pathfind to a new location, and it sets _ptarx and _ptary to where the player is pathfinding to.
- Game waits for player to finish attacking before starting to move.
- Player finishes the attack animation, this calls PlayerStand() which then calls FixPlayerLocation() which resets _ptarx and _ptary to the player's current position.
- Player starts sending the new (wrong) _ptarx and _ptary to other clients, and other clients keep pathfinding the player to the wrong location.

My fix re-calculates the _ptarx and _ptary values each time the player starts walking towards a new tile when following pathfinding. Based on brief testing, this fixes the issue entirely and I haven't seen any downsides.

A better fix would probably be to prevent the game from resetting _ptarx and _ptary to the wrong value after finishing an attack.

### DIFF
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -114,6 +114,33 @@ static BYTE *multi_recv_packet(TBuffer *pBuf, BYTE *body, DWORD *size)
 
 static void NetRecvPlrData(TPkt *pkt)
 {
+	/*
+	* We calculate _ptargx and _ptargy here and prepare for it to be sent to other clients.
+	* If px/py and ptarx/ptary don't match on other clients, then they'll move our
+	* local player to that location.
+	*/
+	{
+		int realTargetX = plr[myplr]._pfutx;
+		int realTargetY = plr[myplr]._pfuty;
+		for (int i = 0; i < MAX_PATH_LENGTH; i++) {
+			if (plr[myplr].walkpath[i] == WALK_NONE)
+				break;
+
+			if (plr[myplr].walkpath[i] == WALK_N || plr[myplr].walkpath[i] == WALK_NW || plr[myplr].walkpath[i] == WALK_W)
+				realTargetX--;
+			else if (plr[myplr].walkpath[i] == WALK_E || plr[myplr].walkpath[i] == WALK_SE || plr[myplr].walkpath[i] == WALK_S)
+				realTargetX++;
+
+			if (plr[myplr].walkpath[i] == WALK_N || plr[myplr].walkpath[i] == WALK_NE || plr[myplr].walkpath[i] == WALK_E)
+				realTargetY--;
+			else if (plr[myplr].walkpath[i] == WALK_S || plr[myplr].walkpath[i] == WALK_SW || plr[myplr].walkpath[i] == WALK_W)
+				realTargetY++;
+		}
+
+		plr[myplr]._ptargx = realTargetX;
+		plr[myplr]._ptargy = realTargetY;
+	}
+
 	pkt->hdr.wCheck = LOAD_BE32("\0\0ip");
 	pkt->hdr.px = plr[myplr]._px;
 	pkt->hdr.py = plr[myplr]._py;
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -114,32 +114,14 @@ static BYTE *multi_recv_packet(TBuffer *pBuf, BYTE *body, DWORD *size)
 
 static void NetRecvPlrData(TPkt *pkt)
 {
-	/*
-	* We calculate _ptargx and _ptargy here and prepare for it to be sent to other clients.
-	* If px/py and ptarx/ptary don't match on other clients, then they'll move our local player to that location.
-	*/
-	{
-		// clang-format off
-		int directionOffsetX[8] = {  0,-1, 1, 0,-1, 1, 1,-1 };
-		int directionOffsetY[8] = { -1, 0, 0, 1,-1,-1, 1, 1 };
-		// clang-format on
-		plr[myplr]._ptargx = plr[myplr]._pfutx;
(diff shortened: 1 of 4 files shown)


FluffyQuack:
I've got a better idea for how to implement this fix, and I've changed this PR to be an early version of that.

ptargx/ptargy are only ever used for 2 things in the code:

- They keep being sent to other players so they can fix the position of our local player.
- It's used for in the "holding left mouse button to move" code to check if we're highlighting a different tile than the one we're moving to.

Since the game does such a bad job correctly updating those two variables (and that's why it's so easy to cause desyncs with player movement), why not just calculate the values from scratch based on the player's pathfinding each time they're used? That also means we can delete a lot of code where the game tries to set them to various values based on player states, which would make it easier to overhaul code related to this in the long run.

I recorded a video showcasing how easily you can break network sync with basic player movement (I do 3 things in the video: queue a move while attacking; cancel movement with attacking and then moving again; and simply clicking to move back and forth), and how it looks with this fix applied: https://www.youtube.com/watch?v=I1ONHhJ2xg0

AJenbo (dismissed):

Source/multi.cpp
```diff
@@ -114,6 +114,25 @@ static BYTE *multi_recv_packet(TBuffer *pBuf, BYTE *body, DWORD *size)
 
 static void NetRecvPlrData(TPkt *pkt)
 {
+	/*
+	* We calculate _ptargx and _ptargy here and prepare for it to be sent to other clients.
+	* If px/py and ptarx/ptary don't match on other clients, then they'll move our local player to that location.
+	*/
+	{
+		// clang-format off
+		int directionOffsetX[8] = {  0,-1, 1, 0,-1, 1, 1,-1 };
```
AJenbo, line 123: Would be nice if you could make your editor apply the clang-format rules.

FluffyQuack: I turned it off there because the arrays became hard to read when you can't see the XY values side-by-side.

FluffyQuack:
Okay, I've done some changes and I think this is the best implementation for this fix.

I've simplified how the _ptargx/_ptargy values are used. Besides initialization, they're now only being updated in one place, which makes it easier to understand how the values work.

The values were also used for input (checking if mouse is targeting a tile player is already moving to), but it was enough to replace that check.

I like this implementation because I think it's easy to work from. If anyone wants to actually try implementing a netcode which is sync thanks to fully deterministic gameplay simulation, then the _ptargx/_ptargy values could just be removed entirely. Or someone could try to find ways to correct desyncs (that would be easier to implement, but it wouldn't fully fix the desyncs).

AJenbo:
Have you considered simply not allowing playermovment?

FluffyQuack:
> Have you considered simply not allowing playermovment?

Could fix all desyncs by setting max player count to 1.

Source/player.h
```diff
@@ -160,8 +160,14 @@ struct PlayerStruct {
...
 	Sint32 _pfuty;   // Future tile Y-position of player. Set at start of walking animation
-	Sint32 _ptargx;  // Target tile X-position for player movment. Set during pathfinding
-	Sint32 _ptargy;  // Target tile Y-position for player movment. Set during pathfinding
+
+	/*
+	* These values store the target position for non-local players during network play.
+	* This is needed because player positions drift over time due to desyncs, so we have
+	* clients try to pathfind non-local players to this position.
+	*/
+	Sint32 _ptargx, _ptargy;
```
AJenbo, line 169: I don't think this is going to jive with clang-tidy, though it makes good sense. Maybe we should instead use a point type, that would greatly simplify all the structures as well.

Source/multi.cpp
```diff
@@ -114,11 +114,14 @@ static BYTE *multi_recv_packet(TBuffer *pBuf, BYTE *body, DWORD *size)
 
 static void NetRecvPlrData(TPkt *pkt)
 {
+	Sint32 targetX, targetY;
```
AJenbo, line 117: Oh the irony, I just finished a PR that changes this to `int`.

Source/player.cpp
```diff
@@ -3131,6 +3129,24 @@ bool PM_DoNewLvl(int pnum)
 	return false;
 }
 
+void GetPlayerTargetPosition(int playerNumber, Sint32 *x, Sint32 *y)
```
AJenbo, line 3132:
```suggestion
void UpdatePlayerTargetPosition(PlayerStruct &player)
```

Source/player.h
FluffyQuack: Should I just change it back to this?

```
Sint32 _ptargx;
Sint32 _ptargy;
```

AJenbo: No, lets try it out and see what happens once I get to this file :)

AJenbo:
@FluffyQuack I merged my clean up and rebased your code on top of it, so merged manually via this commit: https://github.com/diasurgical/devilutionX/commit/4f417d2fdcb2ed32f55d51bc25b5525b9c74f888

I made it into a member function and used a range so we don't even have to bother with what types any of this is, hope you like the new style :) And feel free to ask questions if you don't get some of it.

PR #1595, opened 2021-04-18
Fix for small AC malus of less than 1 resulting in a bonus of 1 AC

thebigMuh:
This one is described in Jarulf's guide, page 24 bottom: 

> There is a minimum increase of 1 in AC. That is, even if the percentage will give an increase to AC less than one, it will be increased by at least one. Due to a bug, any decrease in AC less than 1 will be transformed into a positive increase by 1.

The code just checked if (due to rounding) the AC bonus would be == 0, and if it was, it would increase it to 1. This change makes sure that a negative bonus results in -1 for that case.

It also adds a math::Sign helper for calculating the signum of a value, without branching.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -16,6 +16,7 @@
 #include "options.h"
 #include "stores.h"
 #include "utils/language.h"
+#include "utils/math.h"
 
 namespace devilution {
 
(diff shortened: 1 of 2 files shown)


AJenbo (approved):
Well worth the effort :D

PR #1597, opened 2021-04-18
Adding fancier XP bar

thebigMuh:
Resubmitting from a different branch, since my old PR got royally messed up by mit git experiments.

Functionality is unchanged, but I split the XP bar out into its own file, and created a qol/common file as well for holding common functions.


glebm (dismissed):

qndel:
What's in qol.h at this point? 😆

thebigMuh:
> What's in qol.h at this point? 😆

Auto gold pickup, and monster health bar.
I'll happily split that out into separate files as well, especially as I have more QoL stuff planned/done already.

Source/qol/xpbar.cpp
```diff
@@ -0,0 +1,150 @@
...
+	SetPixel(out, x, y + 2, gradient[idx]);
+	SetPixel(out, x, y + 3, gradient[idx / 2]);
+}
+
+} // namespace
+
+void InitXPBar()
+{
+	if (sgOptions.Gameplay.bExperienceBar) {
+		LoadMaskedArt("data\\xpbar.pcx", &xpbarArt, 1, 1);
```
AJenbo, line 49: Should throw an error screen if file is not found (same as health bar)

AJenbo (changes requested):
Much better indeed, only thing is that it's missing the error message in case you are missing the MPQ/pcx file.

Source/qol/xpbar.cpp
thebigMuh: Done

PR #1598, opened 2021-04-19
Fix init_test_access MPQ flag

glebm:
`MPQ_FLAG_READ_ONLY` -> `MPQ_OPEN_READ_ONLY`.

The former is a flag that is set on an MPQ that was open in read-only
mode. To request opening in read-only mode, `MPQ_OPEN_READ_ONLY` should
be used instead.


AJenbo:
ouch

PR #1602, opened 2021-04-19
🎨 Apply agreeable part of clang-tidy to appfat and automap

AJenbo:
The good news, this marks the first point in the process where I need to adjust the clang-tidy config before progressing.

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -4,6 +4,7 @@
  * Implementation of the in-game map overlay.
  */
 #include "automap.h"
+
 #include "control.h"
 #include "inv.h"
 #include "miniwin/miniwin.h"
@@ -20,7 +21,7 @@ namespace {
  * Maps from tile_id to automap type.
  * BUGFIX: only the first 256 elements are ever read
  */
-Uint16 automaptype[512];
+uint16_t automaptype[512];
 
 static Sint32 AutoMapX;
 static Sint32 AutoMapY;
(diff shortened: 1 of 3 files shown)


glebm (approved):

Source/automap.cpp
```diff
@@ -230,22 +230,22 @@ void DrawAutomapTile(CelOutputBuffer out, Sint32 sx, Sint32 sy, Uint16 automap_t
 	}
 }
 
-void DrawAutomapItem(CelOutputBuffer out, Sint32 x, Sint32 y, Uint8 color)
+void DrawAutomapItem(const CelOutputBuffer &out, int x, int y, uint8_t color)
 {
-	Sint32 x1 = x - AmLine32 / 2;
-	Sint32 y1 = y - AmLine16 / 2;
-	Sint32 x2 = x1 + AmLine64 / 2;
-	Sint32 y2 = y1 + AmLine32 / 2;
```
qndel, line 238: When do we use int and when Sint32?

glebm: Generally `int` unless the exact width is important

AJenbo: `int`: when 16 bit might be is enough and 32 bit fore sure is enough, and 64 bit is defiantly enough.

`Sint32`: Only if we are eating some lower part of SDL that would rather prefer we spoke it's dialect.

int32_t: when only 32bit will do.

PR #1603, opened 2021-04-19
🧹 DiabloUI: Clean up text render cache

glebm:
Cleans up the handling of text render cache for `UiText` and `UiButton`.
This also happens to fix a memory leak (`FreeCache` was never called).


AJenbo (approved):
Also it's a reduction in lines of code

Issue #1611, opened 2021-04-19
Sorcerer - wrong mana cost of spells


Psojed:
DevilutionX 1.2.1 Sorcerers have incorrect mana cost for all spells.

It's caused by yet another difference between Diablo and Hellfire versions of the game - Hellfire patch 1.01 introduced a flat 50% reduction to ALL spell mana costs for the Sorcerer class. It seems this mana cost reduction is applied even if you play devilutionX without Hellfire enabled.

**To Reproduce**
1. Create a new sorcerer in vanilla Diablo and in devilutionX.
2. Compare spell costs of spells between versions.

**Expected behavior**
DevilutionX shouldn't apply the mana cost reduction when Hellfire mode isn't enabled.

**Screenshots**
Here's a spellbook comparison:
![sorc_spellstest_d109](https://user-images.githubusercontent.com/12875358/115218079-746b5900-a106-11eb-9e44-23ae249a4478.jpg)
![sorc_spellstest_dX121](https://user-images.githubusercontent.com/12875358/115218084-76351c80-a106-11eb-8fd0-19fb86942f62.jpg)

And here's the lvl 1 sorcerer character with all spells at level 1 used for the screenshots above:
[single_4.zip](https://github.com/diasurgical/devilutionX/files/6335028/single_4.zip)

**Additional context**
Apparently this should be a simple fix in spells.cpp.

julealgon:
I'd rather see this have a single value throughout. Would be one less useless hellfire check.

PR #1613, opened 2021-04-19
 Apply Sorcerer's mana cost reduction bonus only in Hellfire

Psojed:
fix for https://github.com/diasurgical/devilutionX/issues/1611 now with better order of checks


Eider-McDuck (approved):
I compared this with devilution and it seems that this is logically correct. I also tested it.

PR #1615, opened 2021-04-19
class sounds cleanup

qndel:
image


Source/effects.h
```diff
@@ -1065,6 +1065,8 @@ struct TSFX {
 extern int sfxdelay;
 extern _sfx_id sfxdnum;
 
+_sfx_id GetSoundByClass(int playerID, int16_t soundID);
+void PlaySoundByClass(int playerID, int16_t soundID, bool noRepeat = false);
```
AJenbo, line 1069: No reason for this to be exactly 16bit
```suggestion
_sfx_id GetSoundByClass(int playerID, int soundID);
void PlaySoundByClass(int playerID, int soundID, bool noRepeat = false);
```

qndel: doesn't hurt and matches the sound enum type : P

Source/effects.cpp
```diff
@@ -1061,6 +1061,29 @@ TSFX sgSFX[] = {
 	// clang-format on
 };
 
+_sfx_id herosounds[enum_size<HeroClass>::value][102] = {
```
julealgon, line 1064:
```suggestion
_sfx_id heroSounds[enum_size<HeroClass>::value][102] = {
```

Source/effects.cpp
```diff
@@ -1061,6 +1061,29 @@ TSFX sgSFX[] = {
...
+	{ PS_WARR1, PS_WARR2, PS_WARR3, PS_WARR4, PS_WARR5, PS_WARR6, PS_WARR7, PS_WARR8, PS_WARR9, PS_WARR10, PS_WARR11, PS_WARR12, PS_WARR13, PS_WARR14, PS_WARR15, PS_WARR16, PS_WARR17, PS_WARR18, PS_WARR19, PS_WARR20, PS_WARR21, PS_WARR22, PS_WARR23, PS_WARR24, PS_WARR25, PS_WARR26, PS_WARR27, PS_WARR28, PS_WARR29, PS_WARR30, PS_WARR31, PS_WARR32, PS_WARR33, PS_WARR34, PS_WARR35, PS_WARR36, PS_WARR37, PS_WARR38, PS_WARR39, PS_WARR40, PS_WARR41, PS_WARR42, PS_WARR43, PS_WARR44, PS_WARR45, PS_WARR46, PS_WARR47, PS_WARR48, PS_WARR49, PS_WARR50, PS_WARR51, PS_WARR52, PS_WARR53, PS_WARR54, PS_WARR55, PS_WARR56, PS_WARR57, PS_WARR58, PS_WARR59, PS_WARR60, PS_WARR61, PS_WARR62, PS_WARR63, PS_WARR64, PS_WARR65, PS_WARR66, PS_WARR67, PS_WARR68, PS_WARR69, PS_WARR70, PS_WARR71, PS_WARR72, PS_WARR73, PS_WARR74, PS_WARR75, PS_WARR76, PS_WARR77, PS_WARR78, PS_WARR79, PS_WARR80, PS_WARR81, PS_WARR82, PS_WARR83, PS_WARR84, PS_WARR85, PS_WARR86, PS_WARR87, PS_WARR88, PS_WARR89, PS_WARR90, PS_WARR91, PS_WARR92, PS_WARR93, PS_WARR94, PS_WARR95, PS_WARR96B, PS_WARR97, PS_WARR98, PS_WARR99, PS_WARR100, PS_WARR101, PS_WARR102 },
+};
+
+
+_sfx_id GetSoundByClass(int playerID, int soundID)
```
julealgon, line 1074:
```suggestion
_sfx_id GetHeroSound(int playerId, int soundId)
```
"ByClass" is an implementation detail that does not need to be in the function name.

julealgon, line 1074: Please consider making this a `PlayerStruct` method.

julealgon, line 1074: Seems like this should never be exposed. Can we make it a local lambda or an anonymous function?

Exposing it would violate "tell, don't ask".

Source/effects.cpp
```diff
@@ -1061,6 +1061,29 @@ TSFX sgSFX[] = {
...
+{
+	return herosounds[static_cast<int>(plr[playerID]._pClass)][soundID - 1];
```
julealgon, line 1076:
```suggestion
	return herosounds[static_cast<size_t>(plr[playerID]._pClass)][soundID - 1];
```
For consistency with other places where the value is used as an indexer.

Source/player.cpp
```diff
@@ -1772,19 +1760,7 @@ StartPlayerKill(int pnum, int earflag)
...
-	} else if (plr[pnum]._pClass == HeroClass::Sorcerer) {
-		PlaySfxLoc(PS_MAGE71, p->_px, p->_py);
-	} else if (plr[pnum]._pClass == HeroClass::Monk) {
-		PlaySfxLoc(PS_MONK71, p->_px, p->_py);
-	} else if (plr[pnum]._pClass == HeroClass::Bard) {
-		PlaySfxLoc(PS_ROGUE71, p->_px, p->_py);
-	} else if (plr[pnum]._pClass == HeroClass::Barbarian) {
-		PlaySfxLoc(PS_WARR71, p->_px, p->_py);
-	}
+	PlaySfxLoc(GetSoundByClass(pnum, 71), p->_px, p->_py);
```
julealgon, line 1763: Would be nice to fix the issues in a separate change instead of hiding them inside a cleanup PR.

julealgon (changes requested):
A much needed simplification. Thanks for taking this up.

AJenbo:
Regarding the missing monk72.wav, PS_MAGE72 was used in code instead, but a better solution would probably be to insert it in the sfx table in effects.cpp:
```
	{ sfx_MONK,                  "Sfx\\Monk\\Monk71.wav",       nullptr },
	{ sfx_MONK,                  "Sfx\\Sorceror\\Mage72.wav",   nullptr },
```
This keeps the code cleaner imo. Mage72.wav actually doesn't sound like the sorcerer either :shrug:

Source/effects.cpp
```diff
@@ -1061,6 +1061,29 @@ TSFX sgSFX[] = {
...
+}
+
+void PlaySoundByClass(int playerID, int soundID, bool noRepeat /* = false */)
```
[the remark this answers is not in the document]
AJenbo: I'm also happy with you just removing this as it seems more of an inconsistency, the game should always check if the player is already speaking before starting a new speach.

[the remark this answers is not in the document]
julealgon: Hmm.... you sure of that @AJenbo ? Won't this cause some situations to not work properly, like very fast attacks, or taking lots of hits simultaneously?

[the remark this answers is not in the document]
AJenbo: SFXs (pain, attach, not more room/mana, invalid spell etc) are already not permitted to play multiple times simultaneously.

For streams (speech that is more of a one-off type), the default is to interrupt the current sound, the only time you can actually trigger this before the end of the speech is by the dead guy.

AJenbo:
:fire: Merged manually https://github.com/diasurgical/devilutionX/commit/ec74917cb4b02a5e487a1156455dcb6c8afb21fc

galaxyhaxz:
You can also rewrite the array to be in a vertical layout and export an enum for each voice. Here's how I handled it a few months back:
```c
#ifdef EFFECTS_DATA
static int sgVoice[][NUM_CLASSES] = {
#else
enum {
#endif
	// 1 book: chamber of bone
	// 2 horizon's warning (unused)
	// 3 horizon's spell (unused)
	// 4 found horizon's book (unused)
	// 5 chanting spell (unused)
	// 6 Grimspike (unused)
	// 7 Grimspike summoned (unused)
	VOICE(PS_WARR8, PS_ROGUE8, PS_MAGE8, VOICE_AVENGED),
	VOICE(PS_WARR9, PS_ROGUE9, PS_MAGE9, VOICE_RIPFRIEND),
	// 10 book: valor
	// 11 book: blind
	// 12 book: blood war
	VOICE(PS_WARR13, PS_ROGUE13, PS_MAGE13, VOICE_CANTUSE),
	VOICE(PS_WARR14, PS_ROGUE14, PS_MAGE14, VOICE_NOROOM),
(...)
	PlaySFX(GetVoice(myplr, VOICE_LEORIC));
```

PR #1622, opened 2021-04-19
Replace spaces with tabs

Chance4us:


AJenbo:
Thanks, but the formatting is handled automatically by clang-format.

Tabs for indentation, spaces for alignment. Also, why move the `||` to the back? That isn't how it is done anywhere else and will cause clang-format to online this it.

Iniquitatis:
We're using `clang-format` for such stuff, aren't we?

PR #1619, opened 2021-04-19
clang-tidy cleanups

AJenbo:

PR #1623, opened 2021-04-19
♻️ Fully apply clang-tidy rules to a few files

AJenbo:


Source/control.cpp
```diff
@@ -213,33 +213,33 @@ char SpellITbl[] = {
...
-	PANBTN_SENDMSG,
-	PANBTN_FRIENDLY,
+	kPanbtnCharinfo,
+	kPanbtnQlog,
+	kPanbtnAutomap,
+	kPanbtnMainmenu,
+	kPanbtnInventory,
+	kPanbtnSpellbook,
+	kPanbtnSendmsg,
+	kPanbtnFriendly,
```
julealgon, line 223: This convention makes zero sense to me for enums (the `k` prefix).

Source/control.cpp
```diff
@@ -213,33 +213,33 @@ char SpellITbl[] = {
...
-	{  87,  91, 33, 32, true  }, // chat button
-	{ 527,  91, 33, 32, true  }, // friendly fire button
+	{   9,   9, 71, 19 }, // char button
+	{   9,  35, 71, 19 }, // quests button
+	{   9,  75, 71, 19 }, // map button
+	{   9, 101, 71, 19 }, // menu button
+	{ 560,   9, 71, 19 }, // inv button
+	{ 560,  35, 71, 19 }, // spells button
+	{  87,  91, 33, 32 }, // chat button
+	{ 527,  91, 33, 32 }, // friendly fire button
```
julealgon, line 236: Why the change here? What was the last parameter before?

Source/control.cpp
```diff
@@ -280,20 +280,18 @@ spell_id SpellPages[6][7] = {
  * @param nCel Index of the cel frame to draw. 0 based.
  * @param w Width of the frame.
  */
-static void DrawSpellCel(CelOutputBuffer out, int xp, int yp, BYTE *Trans, int nCel, int w)
+static void DrawSpellCel(const CelOutputBuffer &out, int xp, int yp, BYTE *trans, int nCel, int w)
```
julealgon, line 283: `const &` change seems unrelated to me.

Source/control.cpp
```diff
@@ -468,7 +467,7 @@ void DrawSpellList(CelOutputBuffer out)
 					else
 						sprintf(tempstr, "%i Scrolls", v);
 					AddPanelString(tempstr, true);
-					break;
+				} break;
```
julealgon, line 470: I've never see a `break` outside of the scope like this. Is this standard?

Source/control.cpp
```diff
@@ -1273,29 +1226,26 @@ static void CPrintString(CelOutputBuffer out, int y, const char *str, bool cente
 	}
 }
 
-static void PrintInfo(CelOutputBuffer out)
+static void PrintInfo(const CelOutputBuffer &out)
 {
-	int yo, lo, i;
+	if (talkflag)
```
julealgon, line 1231: Change to early return also unrelated?

Source/control.cpp
AJenbo: It's unused, it made clang-tidy grumpy and this was better handled as an `SDL_rect` anyway so that was the easy way to deal with the issue.

Source/control.cpp
AJenbo: Suppose, but it made it much easier to reason about the scope of the variables, clang-tidy doesn't like lines of uninitialized variables so had to distribute them.

Source/control.cpp
AJenbo: It's not something I came up with by self afaik. IMO It makes it clear that this is the final break and that the break is part of the switch and not something inside the block.

Personally I think that once your case grows so complex that it needs to define variables then it's time to create a sub-function for it. The good news here is that the style is also easily searchable compared to the one where it is on the inside.

Source/control.cpp
AJenbo: You will have to take that up with clang-tidy then.

Source/control.cpp
AJenbo: See if you like this better: https://github.com/diasurgical/devilutionX/compare/6ebb279ca993d2e75805826f24a6f4ad87437d8d..3eb7f46feab5dea4e9bbcfc2d7ade243f6536fbd

julealgon: Yes please.

Source/control.cpp
julealgon: Fair enough. I was just honestly curious as I've really never seen it that way. Totally agree on the sub-function argument.

AJenbo: We have accumulated 8 so far, but more might be revealed once the cleanups are done (5 were found in this PR).

Source/pack.h
```diff
@@ -13,66 +15,66 @@ namespace devilution {
...
-	Uint16 iCreateInfo;
-	Uint16 idx;
-	Uint8 bId;
-	Uint8 bDur;
-	Uint8 bMDur;
-	Uint8 bCh;
-	Uint8 bMCh;
-	Uint16 wValue;
-	Uint32 dwBuff;
+	int32_t iSeed;
```
julealgon, line 18: Why make it signed now?

Source/pack.h
```diff
@@ -13,66 +15,66 @@ namespace devilution {
...
+	uint16_t iCreateInfo;
+	uint16_t idx;
+	uint8_t bId;
+	uint8_t bDur;
+	uint8_t bMDur;
+	uint8_t bCh;
+	uint8_t bMCh;
+	uint16_t wValue;
+	int32_t dwBuff;
```
julealgon, line 27: Same here.

Source/pack.h
```diff
@@ -13,66 +15,66 @@ namespace devilution {
...
-	Sint8 destAction;
-	Sint8 destParam1;
-	Sint8 destParam2;
-	Uint8 plrlevel;
-	Uint8 px;
-	Uint8 py;
-	Uint8 targx;
-	Uint8 targy;
+	int32_t dwLowDateTime;
+	int32_t dwHighDateTime;
```
julealgon, line 32: And here (won't comment more, seems to be quite a few).

Source/player.h
```diff
@@ -151,177 +151,177 @@ struct PlayerStruct {
...
+	uint8_t pLvlLoad;
 	bool pBattleNet;
 	bool pManaShield;
-	Uint8 pDungMsgs2;
+	uint8_t pDungMsgs2;
 	bool pOriginalCathedral;
-	Uint16 wReflections;
-	uint32_t pDiabloKillLevel;
+	uint16_t wReflections;
+	uint8_t pDiabloKillLevel;
```
julealgon, line 313: Changed by mistake?

Source/player.h
```diff
@@ -151,177 +151,177 @@ struct PlayerStruct {
...
-	Uint8 *_pNData;
-	Uint8 *_pWData;
-	Uint8 *_pAData;
-	Uint8 *_pLData;
-	Uint8 *_pFData;
-	Uint8 *_pTData;
-	Uint8 *_pHData;
-	Uint8 *_pDData;
-	Uint8 *_pBData;
+	uint8_t pDamAcFlags;
```
julealgon, line 315: Same here. Drastic change. Was it on purpose?

Source/pack.h
AJenbo: Yeah, I never understand why you comment on every instance of general issues.

glebm:
What command did you use to apply these?

Source/player.h
AJenbo: No, this annoyance happened as part of cleaning up the logic that manipulates it. Suppose it could have gone in as a separate commit.

Source/pack.h
julealgon: To make sure the person is aware of all instances. It is very common when I perform reviews for the person to only fix the problem on that specific line and forget other similar issues.

AJenbo:
@glebm for this one
```
run-clang-tidy-12 -p build 'Source/appfat.cpp' -fix
run-clang-tidy-12 -p build 'Source/automap.cpp' -fix
run-clang-tidy-12 -p build 'Source/capture.cpp' -fix
run-clang-tidy-12 -p build 'Source/control.cpp' -fix
```

For https://github.com/diasurgical/devilutionX/pull/1619 I used:
`run-clang-tidy-12 -checks='-*,modernize-*' -p build 'Source' -fix`

Source/.clang-tidy
```diff
@@ -54,14 +54,16 @@ CheckOptions:
   - { key: readability-identifier-naming.ClassCase,               value: CamelCase  }
   - { key: readability-identifier-naming.StructCase,              value: CamelCase  }
```
glebm, line 55: 1. The Google Style Guide comment no longer applies.
2. `HeaderFilterRegex` above needs to be adjusted (no more SXT directories)

Source/appfat.cpp
```diff
@@ -15,9 +15,9 @@ namespace devilution {
 namespace {
 
 /** Set to true when a fatal error is encountered and the application should shut down. */
-bool terminating = false;
+bool Terminating = false;
```
glebm, line 18: `GlobalVariableCase,      value: aNy_CasE`, so why did this change?

Source/pack.h
AJenbo: Might be from the order I rand the replace commands in. Could thing you noticed.

glebm:
> For #1619 I used:
> run-clang-tidy-12 -checks='-*,modernize-*' -p build 'Source' -fix

@AJenbo 

1. If you specify `-checks` manually, it ignores the config, which is why it applied the trailing return type check (due to `modernize-*`) even though we disable it.
2. You may also want to pass `-format` so that changed lines are reformatted automatically according to our clang-format.

Source/player.h
AJenbo: Fixed.

Source/.clang-tidy
AJenbo: SXT?

Source/appfat.cpp
AJenbo: I applied the rules to `appfat.cpp` without issues, only when getting to the later files did I need to permit "aNy_CasE" for global names. Once everything else has been applied I intent to switch it back to `CamelCase` and apply it globally to the project.

AJenbo:
Oh that's helpful i was running format manually after having run tidy.

Source/.clang-tidy
AJenbo: https://github.com/diasurgical/devilutionX/compare/d68a9735555e942e271285cdfc33448964485639..c6ce60ead4df5f16b24b190117aa7a502aafc470

glebm: Perhaps `(Source|test)`?

Source/automap.cpp
```diff
@@ -516,15 +516,15 @@ void InitAutomap()
...
+		AutomapTypes[i] = b1 + (b2 << 8);
 	}
 
 	mem_free_dbg(pAFile);
 	memset(automapview, 0, sizeof(automapview));
 
 	for (int y = 0; y < MAXDUNY; y++) {
-		for (auto &dFlag : dFlags)
-			dFlag[y] &= ~BFLAG_EXPLORED;
+		for (int x = 0; x < MAXDUNX; x++)
```
AJenbo, line 526: @glebm do you know how we could avoid it formatting 2D arrays into partially using ranges?

Source/pack.h
AJenbo: Ok, I usually update my first comment to mention that they need to go over it and fine the other instances.

Source/.clang-tidy
AJenbo: Sure
https://github.com/diasurgical/devilutionX/compare/c6ce60ead4df5f16b24b190117aa7a502aafc470..41065095ff753d4d7cd73ea860a0c4d218b0fb57

Source/automap.cpp
glebm: Can be disabled with a magic comment:
a) At the end of the line `// NOLINT(modernize-loop-convert)`
b) On the line before`// NOLINTNEXTLINE(modernize-loop-convert)`

Source/objects.cpp
```diff
@@ -4073,8 +4073,8 @@ bool OperateShrineSecluded(int pnum)
 		return true;
 
 	for (int yy = 0; yy < DMAXY; yy++) {
```
glebm, line 4075: This can be an `std::fill_n(automapview, DMAXX * DMAXY)`

AJenbo: How about the ones in loadsave.cpp? They need to happen in the same order as the data is packed in to the save game.

Source/control.cpp
Iniquitatis: Should be replaced by enum class anyway.

Source/control.cpp
Iniquitatis: Yeah, recently I've adopted the same convention for personal projects. Pretty convenient I should say.

Source/appfat.cpp
Iniquitatis: Maybe also add some prefix to the globals?

Source/capture.cpp
```diff
@@ -118,7 +118,7 @@ static bool CapturePix(CelOutputBuffer buf, std::ofstream *out)
 	int height = buf.h();
 	BYTE *pBuffer = (BYTE *)DiabloAllocPtr(2 * width);
 	BYTE *pixels = buf.begin();
-	while (height--) {
+	while ((height--) != 0) {
```
thebigMuh, line 121: Tiny nitpick - I usually prefer > and < to != for cases like this.
Don't think clang tidy can make the connection between "decreasing loop variable" and "check exit condition with > instead of !=".
There are many such loops in the code, I'd use > for all of them.

Source/control.cpp
```diff
@@ -1225,12 +1185,9 @@ void FreeControlPan()
...
 bool control_WriteStringToBuffer(BYTE *str)
 {
-	int k;
-	BYTE ichar;
-
-	k = 0;
-	while (*str) {
-		ichar = gbFontTransTbl[*str];
+	int k = 0;
+	while (*str != 0) {
```
thebigMuh, line 1189: while (*str != '\0') for checking if the null terminator has been reached?
(Many cases of this, not just this one)

Source/control.cpp
```diff
@@ -1777,13 +1716,13 @@ void DrawDurIcon(CelOutputBuffer out)
...
 
-	x = PANEL_X + PANEL_WIDTH - 32 - 16;
+	int x = PANEL_X + PANEL_WIDTH - 32 - 16;
 	if (!hasRoomUnderPanels) {
 		if (invflag || sbookflag)
 			x -= SPANEL_WIDTH - (gnScreenWidth - PANEL_WIDTH) / 2;
 	}
 
-	p = &plr[myplr];
+	PlayerStruct *p = &plr[myplr];
```
thebigMuh, line 1725: Don't see a reason why this is a pointer inster of a reference?
The review won't let me mark the entire block for some reason, but I think this should be preferred:

const PlayerStruct &p = plr[myplr];
x = DrawDurIcon4Item(out, p.InvBody[INVLOC_HEAD], x, 4);
x = DrawDurIcon4Item(out, p.InvBody[INVLOC_CHEST], x, 3);
etc...

Source/control.cpp
```diff
@@ -1934,7 +1865,7 @@ void DrawSpellBook(CelOutputBuffer out)
 				} else {
 					sprintf(tempstr, "Spell Level %i", lvl);
 				}
-				break;
+			} break;
 			}
```
thebigMuh, line 1869: Can't help myself; this looks super strange.

Source/player.h
```diff
@@ -151,177 +151,177 @@ struct PlayerStruct {
 	int8_t walkpath[MAX_PATH_LENGTH];
 	bool plractive;
 	action_id destAction;
-	Sint32 destParam1;
-	Sint32 destParam2;
+	int destParam1;
```
thebigMuh, line 154: Use 'int32_t' in structures used for serialization, to express that the width here matters a lot?

Source/automap.cpp
```diff
@@ -410,7 +410,7 @@ uint16_t GetAutomapType(int x, int y, bool view)
 		return 0;
 	}
 
-	uint16_t rv = automaptype[(BYTE)dungeon[x][y]];
+	uint16_t rv = AutomapTypes[(BYTE)dungeon[x][y]];
```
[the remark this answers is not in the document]
AJenbo: feel free to look into that :)

Source/control.cpp
AJenbo: why?

Source/player.h
AJenbo: Only cmd_*, PKPlayerStruct and PKItemStruct are serialized.

Source/automap.cpp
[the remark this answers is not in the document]
thebigMuh: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-cstyle-cast.html ?

Source/control.cpp
thebigMuh: It doesn't need to be reseated, it can't ever be a nullptr that you would want to handle, and all access is read only. All of that speaks for making it a const&.

Source/player.h
thebigMuh: Ah, I didn't mean 'directly serialized'. I meant that in the sense of "There's code somewhere that reads/writes this value with the expectation of it being a certain size".

I haven't tried, but what happens if you e.g. use writeLE<Uint8>(playerStruct.destParam1)? Will it complain about the narrowing conversion?

Source/control.cpp
AJenbo: reseated as in you cannot reassign it to a different instance during the lifetime of the function?

AJenbo: I like how this sounds, but let's keep it for later as an actual refactor. Who knows maybe there is even a way to configure it in clang-tidy.

thebigMuh: Yeah. There's a bit of a base rule of "Make const& if possible, make * only if necessary".
Plus, typing "." is shorter than "->" :P

Source/objects.cpp
AJenbo: Went with `std::fill(&automapview[0][0], &automapview[DMAXX][DMAXY], true);` just to feel like I'm a free spirite.

Source/capture.cpp
AJenbo: Turns out i'm not a fan of while: `for (int height = buf.h(); height > 0; height--) {`

AJenbo: While (no pun intended) not strictly out of scope, I will probably miss most of these. But feel free to apply the change afterward. Or point it out using GitHub suggestions so I only have to click "accept".

PR #1624, opened 2021-04-20
🎨 DiabloUI: Replace `dynamic_cast` with `static_cast`

glebm:
The type is known here so we do not need to pay the overhead of a `dynamic_cast`.

Follow-up to https://github.com/diasurgical/devilutionX/commit/94f385a46b3c15fd0f5b2cf3ceb8e66df6e75bd5

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -95,7 +95,7 @@ void UiInitList(int count, void (*fnFocus)(int value), void (*fnSelect)(int valu
 	textInputActive = false;
 	for (auto &item : items) {
 		if (item->m_type == UI_EDIT) {
-			auto *pItemUIEdit = dynamic_cast<UiEdit *>(item);
+			auto *pItemUIEdit = static_cast<UiEdit *>(item);
 			SDL_SetTextInputRect(&item->m_rect);
 			textInputActive = true;
 #ifdef __SWITCH__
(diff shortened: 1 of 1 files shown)


AJenbo:
I'll probably never learn when one is more appropriate then the other :shrug:

glebm:
Don't use `dynamic_cast` ever - it's a runtime check using RTTI (which I'd love to disable completely along with exceptions but Source/dvlnet uses both)

glebm:
https://embeddedartistry.com/blog/2017/03/15/c-casting-or-oh-no-they-broke-malloc/

AJenbo:
I just trusted clang-tidy knew what it was doing :shrug:

glebm:
Well, clang-tidy doesn't know if static_cast is safe here but we do

AJenbo:
> Well, clang-tidy doesn't know if static_cast is safe here but we do

easy withe the generalizations there 😅

julealgon:
> (which I'd love to disable completely along with exceptions but Source/dvlnet uses both)

@glebm , could you share a reference on the pitfalls/problems of exceptions in C++?

glebm:
A couple of articles:

1. https://medium.com/swlh/why-some-people-go-exception-less-6138f322fce
2. http://www.furidamu.org/blog/2017/01/28/error-handling-with-statusor/

There are different opinions on whether exceptions should be used. The 2 opinions result in 2 different coding styles.

My opinion is that exceptions are action-at-a-distance and explicit error handling should be used instead (explicit control flow).

Regardless of that, some hobbyist OSs, such as Serenity, do not implement exceptions at all.

julealgon:
Thanks for the links @glebm 

> My opinion is that exceptions are action-at-a-distance and explicit error handling should be used instead (explicit control flow).

I see you really appreciate directly traceable code. You raised similar concerns over event-driven behavior. 

The exception thing is also a common discussion topic in C#, where the recommendation is to use exceptions for _trully_ exceptional things. There is even a known code smell which is "use of exception as control flow".

Are you sure you are not being a bit extreme by just _never_ using exceptions? If they are used for actual exceptional behavior, they work really well. 

Anyways, to continue in chat maybe.

PR #1629, opened 2021-04-20
🐞 Fix hang in drlg_l4

### LINKED ISSUE #1614: Can't load 13 level
**Important information**
Linux x64 / DevilutionX version 1.2.1 / download**Describe**
Can't move down to 13 level. game freeze at 1/3 splash screen load

**To Reproduce**
1. download save
2. try to move to 13 level

**Expected behavior**
Load 13 level successful

[single_0.zip](https://github.com/diasurgical/devilutionX/files/6335921/single_0.zip)

glebm:
Not sure whether this is the correct fix but this change fixes both the hang and the heap-buffer-overflow.

:warning: For all I know, this may generate a non-traversible layout, or a different layout from vanilla. Needs careful review and testing.

There are 2 commits here that both fix the particular save provided in #1629.
They probably behave differently in some scenarios and I don't know which one is the correct fix (if any).

Fixes #1614

### DIFF
--- a/Source/drlg_l4.cpp
+++ b/Source/drlg_l4.cpp
@@ -1004,12 +1004,7 @@ void UShape()
 			}
 			if (dung[i][j] == 1) {
 				// BUGFIX: check that i + 1 < 20 and j + 1 < 20 (fixed)
-				if (i + 1 < 20 && j + 1 < 20
-				    && dung[i][j + 1] == 1 && dung[i + 1][j + 1] == 0) {
-					hallok[j] = true;
-				} else {
-					hallok[j] = false;
-				}
+				hallok[j] = i == 19 || j == 19 || (dung[i][j + 1] == 1 && dung[i + 1][j + 1] == 0);
 				i = 0;
 			}
 		}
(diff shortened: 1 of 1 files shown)


glebm:
There is 2 commits here that both fix the particular save provided in #1629.
They probably behave differently in some scenarios and I don't know which one is the correct fix if any.

Source/drlg_l4.cpp
```diff
@@ -1003,12 +1003,7 @@ static void uShape()
 			}
 			if (dung[i][j] == 1) {
 				// BUGFIX: check that i + 1 < 20 and j + 1 < 20 (fixed)
```
thebigMuh, line 1005: Are these BUGFIX comments still relevant after your changes?

AJenbo: They were helpful in avoiding a revert after merging with upstream when there was a big refactor. But this is becoming much less relevant and I will probably strip them after the next merge.

AJenbo (commented):
Only 1/6 of the tested levels where affected by the change, the one that was was not totally changed, but still fairly dramatically affected.

Considering that the affected map didn't hang previously I'm not sure this is a viable solution.

Before:
image
After:
image

glebm:
Is this using the first commit or the second one?
Perhaps one of them is less invasive

AJenbo:
this was using both, ill test them individually tomorrow

glebm:
This is very much out of date but we should probably still retest the first commit only as a follow-up

PR #1630, opened 2021-04-20
More updated readmes

Psojed:


AJenbo:
Oh wow thanks for this, that's a huge help :)

PR #1631, opened 2021-04-20
🔥 Get rid of _ptargx/y

AJenbo:


Source/player.cpp
```diff
@@ -224,22 +225,22 @@ int PlayerStruct::GetMaximumAttributeValue(CharacterAttribute attribute) const
 	return MaxStats[static_cast<std::size_t>(_pClass)][static_cast<std::size_t>(attribute)];
 }
 
-void PlayerStruct::UpdateTargetPosition()
+SDL_Point PlayerStruct::GetTargetPosition() const
```
julealgon, line 228: Should we make this:
```suggestion
const SDL_Point& PlayerStruct::GetTargetPosition() const
```

glebm: The general guidance is that it's better to return small types (< 4 ints) by value, as the cost of copying for these is much smaller than the cost of memory indirection

AJenbo: I'm unsure how mem management works with something like that?

AJenbo: > The general rule of thumb is that it's better to return small types (e.g <= 16 bytes) by value, as the cost of copying for these is much smaller than the cost of memory indirection

That was my thinking when I went with the current implementation :D Well by small I was thinking 8 bytes is defiantly small on x86_64 :D

glebm: Note that in this case guaranteed RVO (return-value-optimization) applies, so the size of the return type doesn't even matter.

You're right that returning by reference here would crash upon first use (pointer to stack variable).

Source/player.cpp
```diff
@@ -224,22 +225,22 @@ int PlayerStruct::GetMaximumAttributeValue(CharacterAttribute attribute) const
 	return MaxStats[static_cast<std::size_t>(_pClass)][static_cast<std::size_t>(attribute)];
 }
 
-void PlayerStruct::UpdateTargetPosition()
+SDL_Point PlayerStruct::GetTargetPosition() const
 {
 	// clang-format off
 	int directionOffsetX[8] = {  0,-1, 1, 0,-1, 1, 1,-1 };
```
glebm, line 231: While we're here, I'd make these constexpr

Source/debug.cpp
```diff
@@ -145,7 +147,8 @@ void PrintDebugPlayer(bool bNextPlayer)
 		NetSendCmdString(1 << myplr, dstr);
 		sprintf(dstr, "  Lvl = %i : Change = %i", plr[dbgplr].plrlevel, plr[dbgplr]._pLvlChanging);
 		NetSendCmdString(1 << myplr, dstr);
-		sprintf(dstr, "  x = %i, y = %i : tx = %i, ty = %i", plr[dbgplr]._px, plr[dbgplr]._py, plr[dbgplr]._ptargx, plr[dbgplr]._ptargy);
+		SDL_Point target = plr[dbgplr].GetTargetPosition();
```
glebm, line 150:
```suggestion
		const SDL_Point target = plr[dbgplr].GetTargetPosition();

```

glebm (approved):

Source/player.cpp
```diff
@@ -224,22 +225,22 @@ int PlayerStruct::GetMaximumAttributeValue(CharacterAttribute attribute) const
...
 {
 	// clang-format off
-	int directionOffsetX[8] = {  0,-1, 1, 0,-1, 1, 1,-1 };
-	int directionOffsetY[8] = { -1, 0, 0, 1,-1,-1, 1, 1 };
+	constexpr int directionOffsetX[8] = {  0,-1, 1, 0,-1, 1, 1,-1 };
+	constexpr int directionOffsetY[8] = { -1, 0, 0, 1,-1,-1, 1, 1 };
 	// clang-format on
-	_ptargx = _pfutx;
-	_ptargy = _pfuty;
+	SDL_Point target = { _pfutx, _pfuty };
```
glebm, line 234:
```suggestion
	SDL_Point target { _pfutx, _pfuty };

```

Source/track.cpp
```diff
@@ -28,8 +29,8 @@ void track_process()
 	if (plr[myplr]._pVar8 <= 6 && plr[myplr]._pmode != PM_STAND)
 		return;
 
-	plr[myplr].UpdateTargetPosition();
-	if (cursmx != plr[myplr]._ptargx || cursmy != plr[myplr]._ptargy) {
+	SDL_Point target = plr[myplr].GetTargetPosition();
```
glebm, line 32:
```suggestion
	const SDL_Point target = plr[myplr].GetTargetPosition();

```

Source/player.cpp
julealgon: If you don't mind me asking, what is the name of this "pattern/feature"? I've seen it mentioned in a few places saying something like "no assignments".

PR #1635, opened 2021-04-21
fix secluded crash

qndel:
Clicking secluded shrine crashed the game 🙄

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -4032,7 +4032,7 @@ bool OperateShrineSecluded(int pnum)
 	if (pnum != myplr)
 		return true;
 
-	std::fill(&automapview[0][0], &automapview[DMAXX][DMAXY], true);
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -4032,7 +4032,7 @@ bool OperateShrineSecluded(int pnum)
 	if (pnum != myplr)
 		return true;
 
-	std::fill(&automapview[0][0], &automapview[DMAXX][DMAXY], true);
+	std::fill(&automapview[0][0], &automapview[0][0] + sizeof(automapview), true);
```
julealgon, line 4035: Wouldn't this be a simpler fix?
```suggestion
	std::fill(&automapview[0][0], &automapview[DMAXX - 1][DMAXY - 1], true);
```

qndel: nah, to hell with these -1's 😅

julealgon: Using `sizeof` seems a lot more hacky to me.

PR #1637, opened 2021-04-21
Tidy

AJenbo:


Source/DiabloUI/selgame.cpp
```diff
@@ -448,7 +448,7 @@ void selgame_Password_Select(int value)
 	m_game_data->bTheoQuest = sgOptions.Gameplay.bTheoQuest;
 	m_game_data->bCowQuest = sgOptions.Gameplay.bCowQuest;
 
-	if (SNetCreateGame(nullptr, selgame_Password, nullptr, 0, (char *)m_game_data, sizeof(GameData), MAX_PLRS, nullptr, nullptr, gdwPlayerId)) {
+	if (SNetCreateGame(nullptr, selgame_Password, (char *)m_game_data, sizeof(m_game_data), gdwPlayerId)) {
```
thebigMuh, line 451: reinterpret_cast instead of c-style cast

Source/DiabloUI/selgame.cpp
```diff
@@ -421,11 +421,11 @@ static bool IsGameCompatible(GameData *data)
 	return false;
 }
 
-void selgame_Password_Select(int value)
+void selgame_Password_Select(int /*value*/)
```
thebigMuh, line 424: Did it really complain about an unused function parameter having a name? That's not a very helpful rule, IMHO.

Is is for some reason not possible to remove this parameter if it's unused?

Source/DiabloUI/selok.cpp
```diff
@@ -34,7 +34,7 @@ void selok_Free()
 	vecSelOkDialog.clear();
 }
 
-void selok_Select(int value)
+void selok_Select(int /*value*/)
```
Jmgr, line 37: I'm not sure I see why that the point here is, I image it's to remove the warning about an unused argument. Commented code is usually bad though (especially using C-style), and I'd rather disable that warning of debatable usefulness...

Source/DiabloUI/diabloui.cpp
```diff
@@ -632,17 +632,17 @@ void UiAddBackground(std::vector<UiItemBase *> *vecDialog)
 {
 	if (ArtBackgroundWidescreen.surface != nullptr) {
 		SDL_Rect rectw = { 0, UI_OFFSET_Y, 0, 0 };
-		vecDialog->push_back(new UiImage(&ArtBackgroundWidescreen, /*animated=*/false, /*frame=*/0, rectw, UIS_CENTER));
+		vecDialog->push_back(new UiImage(&ArtBackgroundWidescreen, /*bAnimated=*/false, /*iFrame=*/0, rectw, UIS_CENTER));
```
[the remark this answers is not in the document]
AJenbo: Sure is, it suggested it in the last batch, but I didn't understand the difference so left it out :shrug:

Source/diablo.cpp
```diff
@@ -1547,7 +1544,7 @@ static void GetMousePos(LPARAM lParam)
 	MouseY = (short)((lParam >> 16) & 0xffff);
 }
 
-void DisableInputWndProc(UINT uMsg, WPARAM wParam, LPARAM lParam)
+void DisableInputWndProc(UINT uMsg, WPARAM /*wParam*/, LPARAM lParam)
```
thebigMuh, line 1547: This is a really annoying rule :D I suggest removing it.

Source/DiabloUI/selok.cpp
AJenbo: This is how clang-tidy automatically deals with that issue yeah. I could manually delete the name instead...

Source/DiabloUI/selgame.cpp
AJenbo: How about we save that for when we run the rules that deal with casting :)

Source/drlg_l1.cpp
```diff
@@ -1710,32 +1710,32 @@ static void L5AddWall()
 	for (j = 0; j < DMAXY; j++) {
 		for (i = 0; i < DMAXX; i++) {
 			if (!L5dflags[i][j]) {
-				if (dungeon[i][j] == 3 && random_(0, 100) < 100) {
+				if (dungeon[i][j] == 3 && GenerateRnd(100) < 100) {
```
thebigMuh, line 1713: The GenerateRnd(100) < 100 seems pretty pointless, as GenerateRnd(100 will only generate numbers from 0 - 99, unless I'm mistaken? There's a lot of those calls.

AJenbo: You are correct...

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
Jmgr: Basically `emplace_back` allows you to specify the constructor's arguments directly, and is generally more powerful. I'd use that by default and `push_back` in some special cases.

Source/DiabloUI/selok.cpp
Jmgr: Ah that's a bit annoying because you can't see the argument name anymore. Why not disable that warning?

Source/misdat.cpp
```diff
@@ -99,8 +99,8 @@ MissileData missiledata[] = {
 	{  &AddManaTrap,               &MI_Dummy,          MIS_MANATRAP,      false,     1, MISR_NONE,      MFILE_NONE,      IS_CAST7,    SFX_NONE    },
 	{  &AddRing,                   &MI_LightningRing,  MIS_LIGHTRING,     false,     1, MISR_LIGHTNING, MFILE_LGHNING,   SFX_NONE,    SFX_NONE    },
 	{  &AddSearch,                 &MI_Search,         MIS_SEARCH,        false,     1, MISR_NONE,      MFILE_NONE,      SFX_NONE,    SFX_NONE    },
-	{  &AddFlashBack,              &MI_FlashBack,      MIS_FLASHBK,       true,      1, MISR_MAGIC,     MFILE_BLUEXBK,   SFX_NONE,    SFX_NONE    },
-	{  &AddFlashFront,             &MI_FlashFront,     MIS_FLASHFR,       true,      1, MISR_MAGIC,     MFILE_BLUEXFR,   SFX_NONE,    LS_ELECIMP1 },
+	{  nullptr,                    nullptr,            MIS_FLASHBK,       false,     1, MISR_MAGIC,     MFILE_BLUEXBK,   SFX_NONE,    SFX_NONE    },
+	{  nullptr,                    nullptr,            MIS_FLASHFR,       false,     1, MISR_MAGIC,     MFILE_BLUEXFR,   SFX_NONE,    LS_ELECIMP1 },
```
Jmgr, line 103: I'm not sure what's going on here?

Source/init.cpp
```diff
@@ -229,7 +229,7 @@ void init_create_window()
 #endif
 }
 
-void MainWndProc(UINT Msg, WPARAM wParam, LPARAM lParam)
+void MainWndProc(UINT Msg)
```
thebigMuh, line 232: Can you just change the signature of the window procedure? *headscratch*

Source/misdat.cpp
thebigMuh, line 103: I guess you removed those on purpose?

thebigMuh (dismissed):
👍

Source/missiles.cpp
```diff
@@ -1482,14 +1482,14 @@ void AddBerserk(Sint32 mi, Sint32 sx, Sint32 sy, Sint32 dx, Sint32 dy, Sint32 mi
...
 										auto slvl = static_cast<double>(GetSpellLevel(id, SPL_BERSERK));
 										monster[dm]._mFlags |= MFLAG_BERSERK | MFLAG_GOLEM;
-										monster[dm].mMinDamage = ((double)(random_(145, 10) + 20) / 100 - -1) * (double)monster[dm].mMinDamage + slvl;
-										monster[dm].mMaxDamage = ((double)(random_(145, 10) + 20) / 100 - -1) * (double)monster[dm].mMaxDamage + slvl;
-										monster[dm].mMinDamage2 = ((double)(random_(145, 10) + 20) / 100 - -1) * (double)monster[dm].mMinDamage2 + slvl;
-										monster[dm].mMaxDamage2 = ((double)(random_(145, 10) + 20) / 100 - -1) * (double)monster[dm].mMaxDamage2 + slvl;
+										monster[dm].mMinDamage = ((double)(GenerateRnd(10) + 20) / 100 - -1) * (double)monster[dm].mMinDamage + slvl;
+										monster[dm].mMaxDamage = ((double)(GenerateRnd(10) + 20) / 100 - -1) * (double)monster[dm].mMaxDamage + slvl;
+										monster[dm].mMinDamage2 = ((double)(GenerateRnd(10) + 20) / 100 - -1) * (double)monster[dm].mMinDamage2 + slvl;
+										monster[dm].mMaxDamage2 = ((double)(GenerateRnd(10) + 20) / 100 - -1) * (double)monster[dm].mMaxDamage2 + slvl;
```
thebigMuh, line 1492: Teh wat?

(double)(GenerateRnd(10) + 20) / 100 - -1) * (double)monster[dm].mMinDamage + slvl;

- The "- -1" is... +1
- (rnd(10) + 20 / 100) + 1 is essentially generating a random number between 1.2 and 1.29
- The doubles are freaking me out

So essentially the same can be had with:

(120 + GenerateRnd(10)) * monster[dm].mMinDamage / 100 + slvl;

At least assuming that mMinDamage never becomes greater than ~16 mio.

Maybe that's outside the scope of this refactor... should make a note to go over the doubles that are littered everywhere, they are not really needed.

In any case, the casts should be static_cast<double>

Source/missiles.cpp
```diff
@@ -3035,12 +2994,12 @@ void AddNova(Sint32 mi, Sint32 sx, Sint32 sy, Sint32 dx, Sint32 dy, Sint32 midir
 		if (mienemy == TARGET_MONSTERS)
 			UseMana(id, SPL_NOVA);
 	} else {
-		missile[mi]._midam = ((DWORD)currlevel / 2) + random_(66, 3) + random_(66, 3) + random_(66, 3);
+		missile[mi]._midam = ((DWORD)currlevel / 2) + GenerateRnd(3) + GenerateRnd(3) + GenerateRnd(3);
```
thebigMuh, line 2997: Very randomly, DWORD.
Should be a static_cast<int> I suppose.

Source/multi.cpp
```diff
@@ -446,7 +446,7 @@ static void multi_process_tmsgs()
 	int cnt;
 	TPkt pkt;
 
-	while ((cnt = tmsg_get((BYTE *)&pkt, 512)) != 0) {
+	while ((cnt = tmsg_get((BYTE *)&pkt)) != 0) {
```
thebigMuh, line 449: reinterpret_cast

Source/multi.cpp
```diff
@@ -818,7 +795,7 @@ bool multi_init_single(GameData *gameData)
 	}
 
 	unused = 0;
-	if (!SNetCreateGame("local", "local", "local", 0, (char *)&sgGameInitInfo, sizeof(sgGameInitInfo), 1, "local", "local", &unused)) {
+	if (!SNetCreateGame("local", "local", (char *)&sgGameInitInfo, sizeof(sgGameInitInfo), &unused)) {
```
thebigMuh, line 798: They really liked using all kinds of BYTE and char and unsigned char pointers instead of void* for 'pointer to unknown type'.

reinterpret_cast

Source/themes.cpp
```diff
@@ -450,9 +450,9 @@ void InitThemes()
 		for (int i = 0; i < themeCount; i++) {
 			if (themes[i].ttype == THEME_NONE) {
 				themes[i].ttval = themeLoc[i].ttval;
-				theme_id j = ThemeGood[random_(0, 4)];
+				theme_id j = ThemeGood[GenerateRnd(4)];
 				while (!SpecialThemeFit(i, j)) {
-					j = (theme_id)random_(0, 17);
+					j = (theme_id)GenerateRnd(17);
```
thebigMuh, line 455: static_cast

Source/themes.cpp
```diff
@@ -425,9 +425,9 @@ void InitThemes()
 		for (size_t i = 0; i < 256 && numthemes < MAXTHEMES; i++) {
 			if (CheckThemeRoom(i)) {
 				themes[numthemes].ttval = i;
-				theme_id j = ThemeGood[random_(0, 4)];
+				theme_id j = ThemeGood[GenerateRnd(4)];
 				while (!SpecialThemeFit(numthemes, j)) {
-					j = (theme_id)random_(0, 17);
+					j = (theme_id)GenerateRnd(17);
```
thebigMuh, line 430: static_cast

Source/misdat.cpp
AJenbo: deleting 4 unused functions but keeping the spell indexes aligned.

Source/init.cpp
AJenbo: Why not :shrug: 

It's derived and called as a normal function since we aren't Windows-only anymore :)

Source/misdat.cpp
AJenbo: Yeah, unused.

Source/multi.cpp
AJenbo: Not sure if they did char or void. Can be hard to tell.

Source/drlg_l1.cpp
qndel: I'm against removing first argument as without it you basically can't debug rng calls, it's used to track which part of the code called it

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
AJenbo: Will come in a later run, so skipping for now.

Source/missiles.cpp
AJenbo: Please do.

Source/DiabloUI/dialogs.cpp
```diff
@@ -283,7 +285,7 @@ void UiOkDialog(const char *text, const char *caption, bool error, std::vector<U
 
 void UiErrorOkDialog(const char *text, const char *caption, std::vector<UiItemBase *> renderBehind)
```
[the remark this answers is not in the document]
AJenbo: Any chance you would be up for making that change?

Source/diablo.cpp
AJenbo: skipped for now

glebm (approved):

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
thebigMuh: The type being pushed here is a pointer. You could only emplace_back it by passing in a pointer value.

emplace_back is meant for in-place constructing a complex type in the vector, e.g. if you had a std::vector<UiImage>;
push_back also supports move semantics, so unless your type is immovable for some reason, there is no benefit to calling vec.emplace_back(a, b, c); over vec.push_back(SomeType{ a, b, c });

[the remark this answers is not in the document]
AJenbo: Good news this was the next check.

Source/DiabloUI/dialogs.cpp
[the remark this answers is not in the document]
AJenbo: And this was the next rule ... https://github.com/diasurgical/devilutionX/commit/d392f7050942385badad8f1b066e936ff0201ec2

PR #1638, opened 2021-04-21
:art: readability-convert-member-functions-to-static

AJenbo:


glebm (approved):

Source/controls/devices/game_controller.h
```diff
@@ -26,10 +26,10 @@ class GameController {
 	ControllerButton ToControllerButton(const SDL_Event &event);
 
 	bool IsPressed(ControllerButton button) const;
-	bool ProcessAxisMotion(const SDL_Event &event);
+	static bool ProcessAxisMotion(const SDL_Event &event);
 
 private:
-	SDL_GameControllerButton ToSdlGameControllerButton(ControllerButton button) const;
+	static SDL_GameControllerButton ToSdlGameControllerButton(ControllerButton button);
```
thebigMuh, line 32: Private statics are a bit strange. Unless there's a need to call them from the header, or they access a private member of the class, you could just make this not a member function at all, and instead put it in an anonymous namespace in the .cpp file.

AJenbo: Robots are a bit strange, I can live with that.

PR #1639, opened 2021-04-21
:art: Add constant prefix to DiabloUI

AJenbo:


Jmgr (commented):
Not a big fan of this one. Why does it matter that it is a constant?

Source/DiabloUI/selhero.cpp
```diff
@@ -482,7 +482,7 @@ const char *SelheroGenerateName(HeroClass heroClass)
 
 	int iRand = rand() % 10;
 
-	return KNames[static_cast<std::size_t>(heroClass) % 6][iRand];
+	return kKNames[static_cast<std::size_t>(heroClass) % 6][iRand];
```
thebigMuh, line 485: Uiui, just one k short of KKK 😲

thebigMuh (approved):

Jmgr:
> Not a big fan of this one. Why does it matter that it is a constant?

I would also keep the first upper case letter for types, not variables, constants or functions, but that's more a style question.

AJenbo:
> > Not a big fan of this one. Why does it matter that it is a constant?
> 
> I would also keep the first upper case letter for types, not variables, constants or functions, but that's more a style question.

Using it for globals only, lets solve it by getting rid of them :D

Source/DiabloUI/selhero.cpp
```diff
@@ -32,7 +32,7 @@ namespace {
 std::size_t selhero_SaveCount = 0;
 _uiheroinfo selhero_heros[MAX_CHARACTERS];
 _uiheroinfo selhero_heroInfo;
-const size_t KMaxViewportItems = 6;
+const size_t kKMaxViewportItems = 6;
```
julealgon, line 35: Wasn't the previous `K` here already intended to be the hungarian notation one?

Source/DiabloUI/selhero.cpp
```diff
@@ -399,7 +399,7 @@ void SelheroLoadSelect(int value)
 
 const char *SelheroGenerateName(HeroClass heroClass)
 {
-	static const char *const KNames[6][10] = {
+	static const char *const kKNames[6][10] = {
```
julealgon, line 402: Same here.

PR #1640, opened 2021-04-22
GetSFXLength: Respect streaming flag

glebm:
We've discovered that `Mix_LoadWAV_RW` actually loads the whole file in memory so this does not fix the streaming but is good to fix anyway

### DIFF
--- a/Source/effects.cpp
+++ b/Source/effects.cpp
@@ -10,6 +10,13 @@
 #include "sound.h"
 
 namespace devilution {
+namespace {
+#ifndef DISABLE_STREAMING_SOUNDS
+constexpr bool AllowStreaming = true;
+#else
+constexpr bool AllowStreaming = false;
+#endif
+} // namespace
 
 int sfxdelay;
 _sfx_id sfxdnum = SFX_NONE;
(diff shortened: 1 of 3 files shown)


Source/effects.cpp
```diff
@@ -1413,8 +1413,14 @@ void effects_play_sound(const char *snd_file)
 
 int GetSFXLength(int nSFX)
 {
+#ifndef DISABLE_STREAMING_SOUNDS
+	constexpr bool kAllowStreaming = true;
```
AJenbo, line 1417: Plz respect the code style, it's only on commit old :D

```suggestion
	constexpr bool allowStreaming = true;
```

glebm: Done

Source/effects.cpp
```diff
@@ -10,6 +10,13 @@
 #include "sound.h"
 
 namespace devilution {
+namespace {
+#ifndef DISABLE_STREAMING_SOUNDS
+constexpr bool AllowStreaming = true;
+#else
+constexpr bool AllowStreaming = false;
+#endif
```
julealgon, line 18: Is there any way to simplify this? Any way to do something like:
```cpp
constexpr bool AllowStreaming = TO_BOOL(DISABLE_STREAMING_SOUNDS);
```
?

glebm: I can't think of a way that wouldn't be "clever"

AJenbo (approved):

Issue #614, opened 2020-02-14
proposal: change save archive format from MPQ to JSON


mewmew:
This proposal is about updating the character save archive format (i.e. `single_1.sv` and `multi_1.sv`) from MPQ archives to JSON format. The benefit of this is better debug-ability, likely easier modability and extensibility (e.g. store @qndel pixel light sources) of the save file format. The downside is that with added extensibility, it is likely that mods may add custom fields of their own, making the save formats between mods incompatible.

The envisioned process is as follows:

1. Develop a tool for converting from the old to the new save archive format (MPQ -> JSON).
2. Implement support for loading these JSON files, and support loading both the old and the new format.
3. (optional) deprecate the old format by applying automatic conversion?
4. ....
5. profit.

mewmew:
Copy from https://github.com/sanctuary/djavul/issues/3#issue-565558540:

# Old format (MPQ archive)

Save files of Diablo 1 are stored as MPQ archives, and their contents differ based on game mode; *Single Player* or *Multi Player* game.

## Single Player Save Archive

Save archive file name: `single_NN.sv` (MPQ archive)

Save archive file contents:

* `hero`: hero information
* `game`: game state information
* `templNN`: temporary state of dungeon level *NN*
* `tempsNN`: temporary state of set level *NN*
* `permlNN`: permanent state of dungeon level *NN*
* `permsNN`: permanent state of set level *NN*

## Multi Player Save Archive

Save archive file name: `multi_NN.sv` (MPQ archive)

Save archive file contents:

* `hero`: hero information

# New format (directory with JSON files)

## Single Player Save Archive

Save archive directory name: `single_NNN.sv` (file system directory)

Save archive directory contents:

* `hero.json`: hero information
* `game.json`: game state information
* `templNN.json`: temporary state of dungeon level *NN*
* `tempsNN.json`: temporary state of set level *NN*
* `permlNN.json`: permanent state of dungeon level *NN*
* `permsNN.json`: permanent state of set level *NN*

## Multi Player Save Archive

Save archive directory name: `multi_NNN.sv` (file system directory)

Save archive directory contents:

* `hero.json`: hero information

julealgon:
Do we intent to provide any sort of protection against tampering on these files? I think the idea of using MPQ was also to offer some protection against people just manipulating stuff arbitrarily.

I'm not entirely sure myself if that's a good idea, just wondering if you put any thought into it.

julealgon:
[...]

PR #1641, opened 2021-04-22
Fix multiple projectile collisions causing player to get hit through block

### LINKED ISSUE #1531: Blocking spells activates block animation and still deals damage
[single_0.zip](https://github.com/diasurgical/devilutionX/files/6313226/single_0.zip)

We tested using a single succubus enemy type shooting their Blood Stars. Sometimes the spell is blocked completely, but sometimes the spell gets blocked but still deals damage. The same issue is present in vanilla Hellfire.

Hellfire 1.0.1: [https://www.youtube.com/watch?v=8l9SDX6BPjQ](https://www.youtube.com/watch?v=8l9SDX6BPjQ) (thx Maxpire)
devilutionX 1.2.1 Hellfire: [https://cdn.discordapp.com/attachments/764970679564501052/831966291337543680/snow_witch_blocking.mp4](https://cdn.discordapp.com/attachments/764970679564501052/831966291337543680/snow_witch_blocking.mp4)

_In the second video you can see instances of "damage taken" sound triggering, but me not taking damage, that is most likely because I'm wearing 64% magic resistance and items with -5 damage taken._

The expected behaviour is that every blocked Blood Star should be fully blocked and deal no damage, regardless of damage reductions.

---

PS: Attached is my save from devilutionX 1.2.1 Hellfire in a position ready to test.

StephenCWills:
This change renders blocked projectiles harmless by introducing a new target enemy type that hits no one. This also includes a fix for the "disappearing Blood Star" issue that occurs when the game attempts to rotate a blocked Blood Star missile.

Additional notes:
* A missile will only be rendered harmless on block if that missile would also have been deleted on impact with the player. I believe this includes most, if not all, blockable projectiles.
* This adds the `TARGET_NONE` enum value that is not supported by vanilla. If a save with a blocked missile is loaded into vanilla Hellfire, I believe it will be treated the same as `TARGET_PLAYERS`.
* Tested against Blood Star and Lightning (which is strangely not actually a projectile and is not blockable). This might need more testing against various projectiles. If this change looks acceptable, suggestions for projectiles to test would be appreciated.
* I recommend ignoring whitespace when viewing the diff.

This resolves #1531
This resolves https://github.com/diasurgical/devilutionX/issues/1672

### DIFF
--- a/Source/misdat.h
+++ b/Source/misdat.h
@@ -15,6 +15,7 @@ enum mienemy_type : uint8_t {
 	TARGET_MONSTERS,
 	TARGET_PLAYERS,
 	TARGET_BOTH,
+	TARGET_NONE,
 };
 
 enum missile_resistance : uint8_t {
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1141,6 +1141,9 @@ void SetMissAnim(int mi, int animtype)
 
 void SetMissDir(int mi, int dir)
 {
+	if (!misfiledata[missile[mi]._miAnimType].mAnimData[dir]) {
+		return;
+	}
 	missile[mi]._mimfnum = dir;
 	SetMissAnim(mi, missile[mi]._miAnimType);
 }
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -951,142 +951,153 @@ void CheckMissileCol(int i, int mindam, int maxdam, bool shift, int mx, int my, 
 		return;
 	if (my >= MAXDUNY || my < 0)
 		return;
-	if (missile[i]._micaster != TARGET_BOTH && missile[i]._misource != -1) {
-		if (missile[i]._micaster == TARGET_MONSTERS) {
-			if (dMonster[mx][my] > 0) {
-				if (MonsterMHit(
+	if (missile[i]._micaster != TARGET_NONE) {
+		if (missile[i]._micaster != TARGET_BOTH && missile[i]._misource != -1) {
+			if (missile[i]._micaster == TARGET_MONSTERS) {
+				if (dMonster[mx][my] > 0) {
+					if (MonsterMHit(
+					        missile[i]._misource,
+					        dMonster[mx][my] - 1,
+					        mindam,
+					        maxdam,
+					        missile[i]._midist,
+					        missile[i]._mitype,
+					        shift)) {
+						if (!nodel)
+							missile[i]._mirange = 0;
+						missile[i]._miHitFlag = true;
+					}
+				} else {
+					if (dMonster[mx][my] < 0
+					    && monster[-(dMonster[mx][my] + 1)]._mmode == MM_STONE
+					    && MonsterMHit(
+					        missile[i]._misource,
+					        -(dMonster[mx][my] + 1),
+					        mindam,
+					        maxdam,
+					        missile[i]._midist,
+					        missile[i]._mitype,
+					        shift)) {
+						if (!nodel)
+							missile[i]._mirange = 0;
+						missile[i]._miHitFlag = true;
+					}
+				}
+				if (dPlayer[mx][my] > 0
+				    && dPlayer[mx][my] - 1 != missile[i]._misource
+				    && Plr2PlrMHit(
 				        missile[i]._misource,
-				        dMonster[mx][my] - 1,
+				        dPlayer[mx][my] - 1,
 				        mindam,
 				        maxdam,
 				        missile[i]._midist,
 				        missile[i]._mitype,
-				        shift)) {
-					if (!nodel)
+				        shift,
+				        &blocked)) {
+					if (gbIsHellfire && blocked) {
+						if (!nodel) {
+							missile[i]._micaster = TARGET_NONE;
+						}
+						dir = missile[i]._mimfnum + (GenerateRnd(2) != 0 ? 1 : -1);
+						mAnimFAmt = misfiledata[missile[i]._miAnimType].mAnimFAmt;
+						if (dir < 0)
+							dir = mAnimFAmt - 1;
+						else if (dir > mAnimFAmt)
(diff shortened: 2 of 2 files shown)


StephenCWills:
FYI, rebased to rerun the tests.

StephenCWills:
The two latest commits tune up the logic for edge cases associated with Nova, Flamewave, Elemental, and Bone Spirit. This also fixes the case where Elemental and Bone Spirit missiles can hit an enemy twice consecutively if they attempt to change direction on the enemy's location.

Psojed:
I tested on Psychorb projectiles with monk, can confirm it works - I no longer take damage from blocking their projectiles.

Most commonly encountered projectiles would be:
Arrows
Spit Attacks (Spitting Terror dogs in Catacombs/Caves)
Magma Balls (Magma Demons in Caves)
Firebolts (from Bats in the Nest)
Fireballs (Advocates in Hell)
Blood Stars
Magic Attacks (Psychorbs in Nest, Liches in Crypt, Bone Demons in Crypt)

Psojed:
I went to test some more, and I crashed.
Using my Monk, started a New Game on Normal difficulty - walked into Catacombs, appcrash on loading screen. Next I tried the same thing, I could enter Catacombs normally, I went back to town again to try Hive, and I crashed while walking in town between Cata and Hive.
Using Win_x64 build: https://app.circleci.com/pipelines/github/diasurgical/devilutionX/4806/workflows/9a460be3-4950-445e-913a-c8e639984afd/jobs/33336/artifacts

StephenCWills:
@Psojed Thanks for testing! I don't know of anything in this change that would cause an appcrash, but I did notice some recent changes in master caused my multiplayer testing to become a bit unstable. When I have time, I'll try to reproduce and see if I can figure out what's going on.

StephenCWills:
Rebased to include bugfix from 4b4d757.

@Psojed This bug caused items to drop on whichever space the cursor was hovering over. The app would crash whenever I attempted to drop more than one item on the same space. I wasn't able to reproduce the app crash on loading screens, but maybe it's related. It could also be something that only manifests in release builds so I'll do a bit more testing.

Source/misdat.h
```diff
@@ -15,6 +15,7 @@ enum mienemy_type : uint8_t {
 	TARGET_MONSTERS,
 	TARGET_PLAYERS,
 	TARGET_BOTH,
+	TARGET_NONE,
```
obligaron, line 18: I have a question about save-game compatibility:
We introduce a new enum-Value that vanilla doesn't know about.

How much do we care?

1. It must be 100% compatible. We shouldn't introduce new variables, (enum) values and we should fill all variables vanilla uses, even if DevilutionX don't use them.
2. As long as vanilla can load it, don't crash and works normal it's okay. only test it also in vanilla.
3. We only need to read vanilla saves but don't need to write compatible saves.

There are perhaps more fine granular steps between this three options.

I think this pr is in the second category and I saw other pr's with changes in this category. 
For the Animation Distribution Logic the sourcecode could also be simplified if we allowed changes with category 2.
So I'm really interested in your opinions. ☺️ 

If this discussion is out of scope of this pr, I can also open a separate issue or we can discuss it on discord. 🙂

Psojed:
I tried using the latest master branch and I'm getting the appcrash there aswell.
I crash once upon entering catacombs for the first time after installing the latest build, then I can't reproduce the crash anymore. But I can crash the game every time I enter a dungeon and use ESC -> Quit Game. I only seem to crash on my Monk character though, a newly created Warrior didn't crash when I quit the game in the cathedral.

Source/misdat.h
StephenCWills: Disclaimer: I have no official word from the maintainers to back up this claim, but here are my thoughts on the matter.

IMO, save file compatibility is mostly important for testing purposes. It's most convenient if the game produces save files that are directly compatible with vanilla so your category 1 and 2 should both be fine. Having said that, I noticed an effort from over a year ago to use JSON for save files instead of vanilla's MPQ format (#614). This effort included a converter to ensure compatibility. This suggests that category 3 is still okay so long as there's a conversion step to enable testing.

Perhaps it's worth considering to mark changes like this one with a comment in the code, signaling that this enum value is not supported by vanilla. At the same time, I can't really make a solid case for why doing so would be at all valuable. It's already documented in this PR, and we can always reference the devilution codebase for comparison, so this is probably good enough.

AJenbo: I think we are at level 2 for compatibility. Once we get automated testing we might even move the dial to level 3.

The render fixes we have dropped the flag that tracks if things are sticking through the wall, so for a few frames, things might poke through the wall if you load a DeivlutionX save in vanilla. But gameplay doesn't diverge. I'm ok with the same in terms of the animation logic. If the animation has a little visual hiccup it's ok, but preferably the game logic should result in the same state (the attack should still take the same time).

StephenCWills:
@Psojed Rebased to include bugfix for 118d480. This should fix your crashing issue.

StephenCWills:
I played through Hives and Crypt a few times with the attached savegame, and I must say that Monk feels quite a lot better with this fixed. It was very obnoxious to be constantly placed into blockstun only to get hit by nearly every attack anyway.

StephenCWills:
The only weird thing I noticed while playing through Caves and Hell is that a blocked Fireball's explosion is also rendered harmless. I was expecting to get hit by that when standing next to a wall, but I realized that the explosion uses the same missile as the projectile for its collision detection. That's actually also true of a few other missiles, like the elemental damage in `MI_LArrow()`.

This was a good attempt, but I think it would probably be a good idea to start over with a bit of code cleanup around missile collision and then try again. Still, this version works pretty well so I'll just convert this to a draft for now.

StephenCWills:
With the recent fix to missile collision logic, I guess this PR has become entirely irrelevant.

PR #1644, opened 2021-04-22
Tidy up bools in Source/*/*.cpp

AJenbo:


Source/controls/game_controls.cpp
```diff
@@ -26,7 +26,7 @@ DWORD TranslateControllerButtonToKey(ControllerButton controllerButton)
 	case ControllerButton_BUTTON_A: // Bottom button
 		return questlog ? DVL_VK_SPACE : DVL_VK_ESCAPE;
 	case ControllerButton_BUTTON_B: // Right button
-		return sgpCurrentMenu || stextflag || questlog ? DVL_VK_RETURN : DVL_VK_SPACE;
+		return sgpCurrentMenu != nullptr || stextflag != STORE_NONE || questlog ? DVL_VK_RETURN : DVL_VK_SPACE;
```
qndel, line 29: I think some parentheses would help here :P

Source/utils/display.cpp
```diff
@@ -68,7 +68,7 @@ void AdjustToScreenGeometry(int width, int height)
 	gnScreenHeight = height;
 
 	borderRight = 64;
-	if (gnScreenWidth % 4) {
+	if ((gnScreenWidth % 4) != 0) {
 		// The buffer needs to be devisable by 4 for the engine to blit correctly
```
thebigMuh, line 72: divisible :P

Jmgr:
What is the reasoning for adding the extra verbosity to the bool conditions? Would most C++ developers not now that a non-nullptr pointer is checked with `if(pointer) {}`? Also for flags it's a bit cumbersome to write `(flags & myFlag) == myFlag`. I know C# loves this but I'm not convinced.

Source/drlg_l1.cpp
```diff
@@ -1416,9 +1416,9 @@ static void L5firstRoom()
...
 
-		VR1 = GenerateRnd(2);
-		VR2 = GenerateRnd(2);
-		VR3 = GenerateRnd(2);
+		VR1 = (GenerateRnd(2) != 0);
+		VR2 = (GenerateRnd(2) != 0);
+		VR3 = (GenerateRnd(2) != 0);
 
 		if (VR1 + VR3 <= 1)
 			VR2 = true;
```
thebigMuh, line 1424: This is a bit of a weird one.

VR* is a bool. So yes, definitely makes sense to assign comparison result (true/false) to them.

But a few lines below, they are again used like arithmetic types, and added together.

I'd change ```VR1 + VR3 <= 1``` to ```!(VR1 && VR2)```

Source/DiabloUI/ttf_render_wrapped.cpp
```diff
@@ -18,7 +18,7 @@ namespace {
 
 SDL_bool CharacterIsDelimiter(char c, const char *delimiters)
 {
-	while (*delimiters) {
+	while (*delimiters != 0) {
```
thebigMuh, line 21: Guess that could be a ```!= '\0'```

thebigMuh:
> What is the reasoning for adding the extra verbosity to the bool conditions? Would most C++ developers not now that a non-nullptr pointer is checked with `if(pointer) {}`? Also for flags it's a bit cumbersome to write `(flags & myFlag) == myFlag`. I know C# loves this but I'm not convinced.

I agree. I've run into issues where abusing bools as ints and vice versa has caused problems, but never so far have I seen a bug because someone used if (!ptr) ...

But to be honest, as long as the coding styleguide doesn't make my fingers fall off or eyes burn out, I'll just do the clang-tidy / clang-format before submitting a PR, and move on.

Source/utils/display.cpp
AJenbo: Took me a while to figure out what you where talking about, why not use the suggestion feature :D
```suggestion
		// The buffer needs to be divisible by 4 for the engine to blit correctly
```

Source/DiabloUI/ttf_render_wrapped.cpp
AJenbo: It should
```suggestion
	while (*delimiters != '\0') {
```

Source/drlg_l1.cpp
AJenbo: The original code didn't use proper bool types so they were sometimes "flexible" about the usage. Now at least we have a chance to correct some of this. Nice catch.

Source/utils/display.cpp
thebigMuh: One has to be aware of the feature to use the feature!

Now that I'm aware of it, I'll use it :)

AJenbo:
> What is the reasoning for adding the extra verbosity to the bool conditions? Would most C++ developers not now that a non-nullptr pointer is checked with `if(pointer) {}`? Also for flags it's a bit cumbersome to write `(flags & myFlag) == myFlag`. I know C# loves this but I'm not convinced.

I appreciate you spelling "know" as "now", it makes me fell less alone.
---
If `pointer` is called `cow` it might be hard to spot that it's a pointer, and we already decided not to call it `pCow`. It's a bit hazy atm but we defiantly decided that it would be best to write it as `if (pointer != NULL)` when doing the bin exact cleanups if this was just a general trend or to deal with specific issues is a bit unclear, but suffice to say most of the code should already be following this style and this is simply taking care of the leftovers using tooling.
---
We have encountered this situation before:
```
int flags = myFlag;
int allFlag = yourFlag | myFlag;
if (flags & allFlag)
```

thebigMuh (dismissed):

Source/drlg_l1.cpp
```diff
@@ -1416,9 +1416,9 @@ static void L5firstRoom()
...
 		ye = DMAXY - 1;
 
-		VR1 = GenerateRnd(2);
-		VR2 = GenerateRnd(2);
-		VR3 = GenerateRnd(2);
+		VR1 = (GenerateRnd(2) != 0);
+		VR2 = (GenerateRnd(2) != 0);
+		VR3 = (GenerateRnd(2) != 0);
 
 		if (VR1 + VR3 <= 1)
```
AJenbo, line 1423:
```suggestion
		if (!VR1 || !VR3)
```

Source/player.h
```diff
@@ -9,13 +9,13 @@
...
 
 #include "diablo.h"
-#include "enum_traits.h"
 #include "gendung.h"
 #include "items.h"
 #include "multi.h"
 #include "spelldat.h"
 #include "path.h"
 #include "interfac.h"
+#include "uils/enum_traits.h"
```
AJenbo, line 18:
```suggestion
#include "utils/enum_traits.h"
```

Jmgr:
> I appreciate you spelling "know" as "now", it makes me fell less alone.

Oops. :sweat_smile: 

> If `pointer` is called `cow` it might be hard to spot that it's a pointer, and we already decided not to call it `pCow`.

Why do you need to spot it's a pointer?

> It's a bit hazy atm but we defiantly decided that it would be best to write it as `if (pointer != NULL)` when doing the bin exact cleanups if this was just a general trend or to deal with specific issues is a bit unclear, but suffice to say most of the code should already be following this style and this is simply taking care of the leftovers using tooling.

Ok, but does that also apply to smart pointers?
Like:
```
std::unique_ptr<Cow> cow = [...]
if(cow.get() != nullptr) {}
```
How is that clearer than
```
std::unique_ptr<Cow> cow = [...]
if(cow) {}
```

> 
> We have encountered this situation before:
> 
> ```c++
> int flags = myFlag;
> int allFlag = yourFlag | myFlag;
> if (flags & allFlag) // did we want all flags or any flags?
> ```

We should probably be using std::bitset here, but that's for another PR.

AJenbo:
Not cleaner, but can help catch some issues. Maybe we can change it once things are better. We don't have many smart pointers, and I don't see your example in the code, do you have a specific one?

The rule being applied here is https://clang.llvm.org/extra/clang-tidy/checks/readability-implicit-bool-conversion.html

Jmgr:
> Not cleaner, but can help catch some issues. Maybe we can change it once things are better. We don't have many smart pointers, and I don't see your example in the code, do you have a specific one?

Indeed we don't, as of now, but we should. I would even say that modern C++ never uses raw pointers except to interact with C functions or legacy interfaces. That makes the code safer, because smart pointers explicitly display ownership and take care of freeing the memory. But that's for another PR.

PR #1645, opened 2021-04-22
stop shrines and black deaths from ruining characters

qndel:
While they'd fit the game if it was meant to be played in a single playthrough or hardcore, it clashes horribly with multiplayer/multiple difficulties, which were added later - they offer replayability, yet crap like black death / shrines which ruin the replayability still exists (keep clicking random shrines and your character will end up being unable to use any spell! yay!)

This PR makes the punishment even a bit harder for some classes (if they gain more than 1 hp per vit) but in a way that doesn't ruin characters.
Each black death hit removes 1 vitality and Ornate/Sacred/Fascinating shrines remove 10% of your magic.

Keep in mind that it was usually new players that were victims of black deaths/shrines just because they didn't know about them.
Nothing better than learning your character got ruined without even knowing huh?
Old players would never use such a shrine / get hit by black death, which is yet another design flaw in my opinion.
These -10% max mana shrines effectively make goat shrines and cauldrons useless, because who's going to take a gamble?
"Oh boy, a random shrine, only 5% chance to get my character permanently ruined but I might gain some gold or a potion!" 🤪

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -2216,30 +2216,8 @@ void M_TryH2HHit(int i, int pnum, int Hit, int MinDam, int MaxDam)
 		return;
 	}
 	if (monster[i].MType->mtype == MT_YZOMBIE && pnum == myplr) {
-		cur_ms_num = -1;
-		for (j = 0; j < nummissiles; j++) {
-			misnum = missileactive[j];
-			if (missile[misnum]._mitype != MIS_MANASHIELD)
-				continue;
-			if (missile[misnum]._misource == pnum)
-				cur_ms_num = misnum;
-		}
-		if (plr[pnum]._pMaxHP > 64) {
-			if (plr[pnum]._pMaxHPBase > 64) {
-				plr[pnum]._pMaxHP -= 64;
-				if (plr[pnum]._pHitPoints > plr[pnum]._pMaxHP) {
-					plr[pnum]._pHitPoints = plr[pnum]._pMaxHP;
-					if (cur_ms_num >= 0)
-						missile[cur_ms_num]._miVar1 = plr[pnum]._pHitPoints;
-				}
-				plr[pnum]._pMaxHPBase -= 64;
-				if (plr[pnum]._pHPBase > plr[pnum]._pMaxHPBase) {
-					plr[pnum]._pHPBase = plr[pnum]._pMaxHPBase;
-					if (cur_ms_num >= 0)
-						missile[cur_ms_num]._miVar2 = plr[pnum]._pHPBase;
-				}
-			}
-		}
+		if (plr[myplr].GetBaseAttributeValue(CharacterAttribute::Vitality) > 0)
+			ModifyPlrVit(myplr, -1);
 	}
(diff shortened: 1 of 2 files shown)


NikoVP:
I know some hardcore fans will be against this feature, IMHO this is a great help to the casual and unsuspecting player. Given that these rogue like mechanics are so unequally distributed (only 1 monster does this HP reduction is found on 3-6 levels of the game and nowhere else, or the shrines being distinguishable only in the cathedral ), doesn't add that much gameplay diversity. An option to disable these gameplay mechanics actually allows to use shrines everywhere and to bravely whak all the enemies, without unsuspecting surprises. Of course this is great as long as it doesn't create other technical issues in multiplayer. Once again great work and thanks!

qndel:
> Of course this is great as long as it doesn't create other technical issues in multiplayer.

Considering stat reducing shrines are in multiplayer, I'd expect reducing stats to work fine in general = there shouldn't be any problems with this.

malvarenga123:
I like the approach you've taken, but I think it should be optional.

qndel:
and how u gonna sync it in multiplayer if it's optional 🙄

Source/monster.cpp
```diff
@@ -2216,30 +2216,8 @@ void M_TryH2HHit(int i, int pnum, int Hit, int MinDam, int MaxDam)
...
-				}
-				plr[pnum]._pMaxHPBase -= 64;
-				if (plr[pnum]._pHPBase > plr[pnum]._pMaxHPBase) {
-					plr[pnum]._pHPBase = plr[pnum]._pMaxHPBase;
-					if (cur_ms_num >= 0)
-						missile[cur_ms_num]._miVar2 = plr[pnum]._pHPBase;
-				}
-			}
-		}
+		if (plr[myplr].GetBaseAttributeValue(CharacterAttribute::Vitality) > 0)
```
galaxyhaxz, line 2219: IMHO, you should have it so it won't reduce the player's stat (vitality, magic etc.) below the minimum starting point when creating a character.

malvarenga123:
> and how u gonna sync it in multiplayer if it's optional 
> I'm tired of all these toggles and shit, duping fix isn't a toggle and it ruins speedrunning ...

But you're not fixing anything here. It was a concious option of the devs to have Black Deaths and shrines ruining your character. Just because you don't like it, doesn't mean it should change.

As @galaxyhaxz proposed on the thread about inventory sorting, maybe it would be a better idea to have a fork where all changes could go regardless of vanilla D1. I think @pionere already has a fork with plenty of his changes.

galaxyhaxz:
>  It was a concious decision of the devs to have Black Deaths and shrines ruining your character.

If you read Condor's 1994 Diablo pitch, the original design was to make death permanent. Nada, save file deletes. As mentioned, end game wasn't really thought of so permanent penalties didn't matter. You'd play the game, kill diablo, and call it quits. Multi/difficulties were added so late, and max level went up from 30 to 50 to reflect this. But overall the game is still based around single playthroughs, unlike D2 which boasts more replayability.

PR #1646, opened 2021-04-22
♻️ Clean up bool logic in engine.cpp

AJenbo:


Source/engine.cpp
```diff
@@ -217,72 +193,64 @@ void CelBlitLightSafeTo(CelOutputBuffer out, int sx, int sy, BYTE *pRLEBytes, in
...
-	for (; src != &pRLEBytes[nDataSize]; dst -= out.pitch() + w, shift = (shift + 1) & 1) {
-		for (i = w; i;) {
-			width = *src++;
-			if (!(width & 0x80)) {
-				i -= width;
+
+	BYTE *src = pRLEBytes;
+	BYTE *dst = out.at(sx, sy);
+	BYTE *tbl = &pLightTbl[light_table_index * 256];
+	BYTE shift = ((BYTE)(size_t)dst & 1) != 0;
```
thebigMuh, line 209: Took me a bit to grok that.

This checks if the destination pointer is pointing to an even address, i.e. 16 bit aligned.

To do so, it first casts the pointer to size_t, then immediately to BYTE, then it masks out the lowest bit, and if the result is not 0 (so 1), then the expression returns true - which is immediately converted into a BYTE again, for some later use.

```suggestion
	BYTE shift = ((BYTE)(size_t)dst & 1) != 0;
	bool shift = (static_cast<uintptr_t>(dst) % 2) != 0;
```

Source/engine.cpp
```diff
@@ -217,72 +193,64 @@ void CelBlitLightSafeTo(CelOutputBuffer out, int sx, int sy, BYTE *pRLEBytes, in
...
+
+	for (; src != &pRLEBytes[nDataSize]; dst -= out.pitch() + nWidth, shift = (shift + 1) & 1) {
```
thebigMuh, line 211: If you change shift to a bool, you can do here just:

```suggestion
	for (; src != &pRLEBytes[nDataSize]; dst -= out.pitch() + nWidth, shift = !shift) {

```

Source/engine.cpp
```diff
@@ -668,19 +631,16 @@ int32_t GenerateRnd(int32_t v)
...
-	if (pdwFileLen)
+	if (pdwFileLen != nullptr)
 		*pdwFileLen = fileLen;
 
-	if (!fileLen)
+	if (fileLen == 0)
 		app_fatal("Zero length SFILE:\n%s", pszName);
 
-	buf = (BYTE *)DiabloAllocPtr(fileLen);
+	BYTE *buf = (BYTE *)DiabloAllocPtr(fileLen);
```
thebigMuh, line 643: static_cast

AJenbo: in a nother PR sure

thebigMuh (dismissed):
You pulled GetDirection and GetDirection8 into one function.

AFAICT they do the same thing, figuring out which of 8 cardinal directions come closest to a vector of arbitrary orientation.

GetDirection8 used a 5:2 ratio for finding this out, and GetDirection used a 2:1 ratio (now also 5:2).
AFAIK, Diablo's isometric view uses a x:y ratio of 2:1, so that somehow makes inherently more sense to me.

Are you sure you can just merge the two without causing some weird issues?

AJenbo:
`GetDirection` does not operate in screen space but tile space. The difference is simply one of accuracy, not perspective.

thebigMuh:
> `GetDirection` does not operate in screen space but tile space. The difference is simply one of accuracy, not perspective. It's an approximation of `tan(22.5)` and 5:2 comes closer to that than 2:1.

Ah I get it, so the 22.5 is because it's bisecting the angle between the axes (0/90/180/270) and the half-axes (45/135/...).

Well, you could go all out and use 99/41 instead of 5/2, which is off by only 0.000072

Source/engine.cpp
AJenbo: `error: invalid ‘static_cast’ from type ‘devilution::BYTE*’ {aka ‘unsigned char*’}`

AJenbo: Did a c-style cast for now `(size_t)dst`, but applied your other suggestions. Then we can make it in to a proper C++ style cast later :)

thebigMuh: Woops, reinterpret_cast

AJenbo: Yeah, now I don't feel bad for not understanding when to use them either...

PR #1651, opened 2021-04-23
Switch from SDL_mixer to SDL_audiolib

glebm:
SDL_mixer can only stream a single music track. SDL_audiolib has unlimited streams.

With this change, we finally have:

* Streaming sounds beyond just the music (respecting sfx_STREAM sound effect flag).
* Unlimited channels.
* Unified interface for all audio: SVid, Music, and Sound Effects.
* diablo.ini and build-time audio options:
  * Number of channels: can go mono if your device only has one speaker anyway
  * Output sample rate and resampling quality: can set them high on a beefy PC

Also fixes: Broken Lazarus audio.

Status: Complete :heavy_check_mark: 

* [x] GCC 6 compatibility for Amiga. https://github.com/realnc/SDL_audiolib/pull/16
* [x] Does not build in VS2019 due to the use of alternative operators. https://github.com/realnc/SDL_audiolib/issues/15
* [x] SDL1 support. https://github.com/realnc/SDL_audiolib/pull/12
* [x] Equivalent to `Mix_SetPanning`. https://github.com/realnc/SDL_audiolib/issues/6
* [x] CMake fixes:
  * [x] (A) Find libraries without pkg-config. https://github.com/realnc/SDL_audiolib/issues/7
  * (B) Alternatively, add pkg-config support where it is missing
    * [x] Fix pkg-config support on Vita. #1669
    * [x] Fix pkg-config support on mingw. #1669
    * [x] Fix pkg-config support for Switch. #1700 #1702
    * Use pkg-config via vcpkg. https://github.com/microsoft/vcpkg/issues/17481
  * [x] Fix generated headers: https://github.com/realnc/SDL_audiolib/pull/5
  * [x] Enable policy to allow easy usage with `FetchContent`. https://github.com/realnc/SDL_audiolib/pull/3
* [x] Remove the need for `libsndfile` dependency.
  * There are no `libsndfile` packages for 3DS and the Switch. Thankfully, @realnc has now added a bundled WAV decoder into `SDL_audiolib`, eliminating the need for `libsndfile`.
* [x] Occasional assert failure (crash) at exit in `SDL_audiolib`. https://github.com/realnc/SDL_audiolib/pull/20


Packaging/windows/mingw-prep.sh
```diff
@@ -7,9 +7,10 @@ wget https://www.libsdl.org/release/SDL2-devel-2.0.14-mingw.tar.gz
...
 tar -xzf SDL2_ttf-devel-2.0.15-mingw.tar.gz
-wget https://www.libsdl.org/projects/SDL_mixer/release/SDL2_mixer-devel-2.0.4-mingw.tar.gz
-tar -xzf SDL2_mixer-devel-2.0.4-mingw.tar.gz
+wget https://repo.msys2.org/mingw/i686/mingw-w64-i686-libsndfile-1.0.31-1-any.pkg.tar.zst
+tar -I zstd -xvf mingw-w64-i686-libsndfile-1.0.31-1-any.pkg.tar.zst
 wget https://github.com/jedisct1/libsodium/releases/download/1.0.18-RELEASE/libsodium-1.0.18-mingw.tar.gz
 tar -xzf libsodium-1.0.18-mingw.tar.gz --no-same-owner
 sudo cp -r libsodium-win32/* /usr/i686-w64-mingw32
 sudo cp -r SDL2*/i686-w64-mingw32 /usr
+sudo cp -r mingw64/* /usr
```
Trihedraf, line 16:
```suggestion
sudo cp -r mingw32/* /usr
```

the folder inside this archive is mingw32

.circleci/config.yml
```diff
@@ -65,7 +65,7 @@ jobs:
     steps:
       - checkout
       - run: apt-get update -y
-      - run: apt-get install -y cmake gcc-mingw-w64-i686 g++-mingw-w64-i686 wget git sudo
+      - run: apt-get install -y cmake gcc-mingw-w64-i686 g++-mingw-w64-i686 wget git sudo zstd
```
Trihedraf, line 68:
```suggestion
      - run: apt-get install -y cmake gcc-mingw-w64-i686 g++-mingw-w64-i686 pkg-config-mingw-w64-i686 wget git sudo zstd
```

Looks like it's looking for pkg-config, I think it would be this package

.circleci/config.yml
```diff
@@ -78,7 +78,7 @@ jobs:
     steps:
       - checkout
       - run: apt-get update -y
-      - run: apt-get install -y cmake gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 wget git sudo
+      - run: apt-get install -y cmake gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 wget git sudo zstd
```
Trihedraf, line 81:
```suggestion
      - run: apt-get install -y cmake gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 pkg-config-mingw-w64-x86-64 wget git sudo zstd
```

.circleci/config.yml
glebm: Thanks, I've tried adding them in [`71249d2` (#1651)](https://github.com/diasurgical/devilutionX/pull/1651/commits/71249d233c52d45099c1cced1a1759eca9d26b07) but ultimately SDL_audiolib may need to be fixed to not require pkg-config https://github.com/realnc/SDL_audiolib/issues/7

Trihedraf: Well it fixed that requirement, but now it's not finding "required package sdl2".

glebm: Strange, since `sdl2.pc` is there:

```
$ find . | grep pc                                                                   
./x86_64-w64-mingw32/lib/pkgconfig/sdl2.pc
./i686-w64-mingw32/lib/pkgconfig/sdl2.pc
```

glebm: We probably need to set the path to the pkgconfig binary

/usr/bin/i686-w64-mingw32-pkg-config
/usr/bin/x86_64-w64-mingw32-pkg-config

glebm: That worked but now it's:

```
--   Package 'flac', required by 'sndfile', not found
```

Looks like this sndfile is compiled with support for FLAC and OGG/Vorbis which means 3 more packages :(

glebm: All good now, hitting another issue but it's unrelated to this
https://github.com/realnc/SDL_audiolib/issues/8

pgielda:
Not sure if that is easily doable but it would be great to allow to compile devilutionx without sound support (NOSOUND, similar to e.g. NONET) so that the dependency is not a hard requirement.

glebm:
@pgielda Yeah, we'll have something like eventually but it's out-of-scope for this change, which simply swaps one library for another.

Source/utils/soundsample.cpp
```diff
@@ -41,16 +41,20 @@ void SoundSample::Play(int lVolume, int lPan, int channel)
 	if (!stream_.has_value())
 		return;
 
-	stream_->setVolume(static_cast<float>(std::pow(10.0, lVolume / 2000.0)));
+	if (lPan == 0) {
```
glebm, line 44: @realnc @thebigMuh Can you please review this?

Source/utils/soundsample.cpp
```diff
@@ -16,90 +21,77 @@ namespace devilution {
...
-		SDL_Log("Too few channels, skipping sound");
-		return;
+	stream_->setVolume(static_cast<float>(std::pow(10.0, lVolume / 2000.0)));
+	if (lPan == 0) {
+		stream_->setStereoPosition(0);
+	} else {
+		stream_->setStereoPosition(
+		    std::copysign(
+		        std::pow(10.F, static_cast<float>(-std::fabs(lPan) / 2000.F)),
+		        static_cast<float>(lPan)));
```
glebm, line 51: @realnc @thebigMuh Can you please review this?

thebigMuh, line 51: There's a '1 - ' missing, otherwise it will actually pan samples towards the center the further to the left/right the sound source is.

```suggestion
		    std::copysign(
		        1.0f - std::pow(10.F, static_cast<float>(-std::fabs(lPan) / 2000.F)),
		        static_cast<float>(lPan)));

```

Source/utils/soundsample.h
```diff
@@ -13,11 +16,20 @@ class SoundSample final {
...
+	 * @param dwBytes Length of buffer
+	 * @return 0 on success, -1 otherwise
+	 */
+	int SetChunk(std::unique_ptr<std::byte[]> file_data, DWORD dwBytes);
+
 	int GetLength();
 
 private:
-	Mix_Chunk *chunk;
+	std::unique_ptr<std::byte[]> file_data_;
```
thebigMuh, line 31: I'd use a std::vector<std::byte> for this. No particular reason other than its's easier to handle and understand than a unique_ptr of bytes, and you don't have to lug around the size separately.

Source/utils/soundsample.h
```diff
@@ -13,11 +16,20 @@ class SoundSample final {
...
+	std::optional<Aulib::Stream> stream_;
```
thebigMuh, line 32: Not sure why you're using an optional instead of a unique_ptr here?

glebm: unique_ptr<x> => x is stored on heap
optional<x> => x is stored on stack

Source/utils/soundsample.h
glebm: the size is known in advance and never changes, vector is wrong / overkill / miscommunicates intent for this

Source/utils/soundsample.h
thebigMuh: Well, the optional is stored in this class, so unless this class is stored on the stack, it will just be stored on the heap with the rest of it.

I'm reading bottom to top here, so maybe the stack placement is still coming, but at least right now all SoundSamples in the game get newed on the heap in sound_file_load

glebm: >  it will just be stored on the heap with the rest of it.

It's about indirection and intent. When it comes to indirection, there is no practical difference here, but I like to avoid it when it isn't necessary.

Source/sound.cpp
```diff
@@ -240,11 +256,10 @@ void music_start(uint8_t nTrack)
 				return;
 			}
 
-			Mix_VolumeMusic(MIX_MAX_VOLUME - MIX_MAX_VOLUME * sgOptions.Audio.nMusicVolume / VOLUME_MIN);
-			if (Mix_PlayMusic(music, -1) < 0) {
-				SDL_Log("Mix_PlayMusic: %s", Mix_GetError());
-				Mix_FreeMusic(music);
-				music = nullptr;
+			music->setVolume(1.0F - 1.0F * sgOptions.Audio.nMusicVolume / VOLUME_MIN);
```
thebigMuh, line 259: static_cast<float>? Another one a few lines below.

Writing of float literals seems to randomly switch between using uppercase F and lowercase f, and just a '.' or '.0'

Source/options.h
```diff
@@ -25,6 +25,15 @@ struct AudioOptions {
...
 	bool bAutoEquipSound;
+
+	/** @brief Output sample rate (Hz) */
+	Uint32 nSampleRate;
+	/** @brief The number of output channels (1 or 2) */
+	Uint8 nChannels;
+	/** @brief Buffer size (number of frames per channel) */
+	Uint32 nBufferSize;
+	/** @brief Quality of the resampler, from 0 (lowest) to 10 (highest) */
+	Uint8 nResamplingQuality;
```
thebigMuh, line 36: Ajenbo is currently busy replacing these with uint32_t and uint8_t.

thebigMuh (dismissed):
Can't say anything about the build scripts, not something I know about.

Source/options.h
AJenbo: Well only in areas where it's not directly interacting with SDL, I think it should be correct at this point. At least to my understanding.

glebm:
The only remaining issue is a crash on exit but we can get it fixed at some point after the merge.

glebm:
Figured out the crash at exit and sent a PR too

AJenbo:
This makes me nervous, but on the plus side we don't have to wait for SDL_Mixer to be ported to Xbox...

glebm:
Updated with the crash-at-exit fix and rebased. This is ready to go :rocket:

AJenbo:
In relation to mods, how difficult would it enable support for OGG on major platforms at a later point?

3rdParty/SDL_audiolib/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+
+# No need for the libsamplerate resampler:
+set(USE_RESAMP_SRC OFF)
+
+# No need for the SOX resampler:
+set(USE_RESAMP_SOXR OFF)
+
+# We do not need any of the audio formats except WAV:
+set(USE_DEC_DRWAV ON)
+set(USE_DEC_DRFLAC OFF)
```
AJenbo, line 17: Oh, that's pretty easy...

glebm: DRFLAC is bundled so it'd be trivial to enable if needed

For OGG, we'd need to add a dependency on the ogg/vorbis libraries.
That's not difficult either.

thebigMuh (dismissed):

Source/sound.cpp
```diff
@@ -227,9 +242,11 @@ void music_start(uint8_t nTrack)
 			if (musicRw == nullptr)
 				ErrSdl();
 #endif
-			music = Mix_LoadMUSType_RW(musicRw, MUS_NONE, /*freesrc=*/1);
-			if (music == nullptr) {
-				Log("Mix_LoadMUSType_RW: {}", Mix_GetError());
+			music.emplace(musicRw, std::make_unique<Aulib::DecoderDrwav>(),
+			    std::make_unique<Aulib::ResamplerSpeex>(sgOptions.Audio.nResamplingQuality), /*closeRw=*/true);
+			if (!music->open()) {
+				SDL_LogError(SDL_LOG_CATEGORY_AUDIO, "Aulib::Stream::open (from music_start): %s", SDL_GetError());
```
AJenbo, line 248:
```suggestion
				LogError(Audio, "Aulib::Stream::open (from music_start): %s", SDL_GetError());
```

glebm: Done

Source/utils/push_aulib_decoder.cpp
```diff
@@ -0,0 +1,102 @@
...
+	std::memcpy(item.data.get(), data, size * sizeof(data[0]));
+	item.len = size;
+	item.pos = item.data.get();
+	queue_.push(std::move(item));
+}
+
+void PushAulibDecoder::PushSamples(const std::int8_t *data, unsigned size) noexcept
+{
+	AudioQueueItem item;
+	item.data.reset(new std::int16_t[size]);
```
AJenbo, line 25: Technically everything else uses `int16_t` and not `std::int16_t` :shrug:

Source/sound.cpp
AJenbo: I think there are lots of them.

glebm: Don't think I've missed, have I?

Source/utils/push_aulib_decoder.h
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	[[nodiscard]] int getRate() const override
+	{
+		return sampleRate_;
+	}
+
+	bool rewind() override;
+	[[nodiscard]] std::chrono::microseconds duration() const override;
+	bool seekToTime(std::chrono::microseconds pos) override;
```
AJenbo, line 45: Fancy

Source/utils/soundsample.cpp
```diff
@@ -112,21 +103,9 @@ int SoundSample::SetChunk(BYTE *fileData, DWORD dwBytes)
...
-
-	int bytePerSample = 2;
-	if (format == AUDIO_U8 || format == AUDIO_S8) {
-		bytePerSample = 1;
-	}
-
-	uint64_t ms = 1000;                             // milliseconds, 64bit to avoid overflow when multiplied by alen
-	int bps = frequency * channels * bytePerSample; // bytes per second
-	return (int)(chunk->alen * ms / bps);
+	return std::chrono::duration_cast<std::chrono::milliseconds>(stream_->duration()).count();
```
AJenbo, line 108: Lol this is just unfair :D

AJenbo (approved):

Source/sound.cpp
AJenbo: Oh, i thought you just changed that one, maybe stale diff on my end.

PR #1652, opened 2021-04-23
♻️ Clean up direction related code

AJenbo:
this is a lot easier to read if you set the diff to ignore white space

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1868,23 +1868,21 @@ bool CanPut(int x, int y)
 
 bool TryInvPut()
 {
-	int dir;
-
 	if (numitems >= MAXITEMS)
 		return false;
 
-	dir = GetDirection(plr[myplr]._px, plr[myplr]._py, cursmx, cursmy);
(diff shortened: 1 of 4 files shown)


Source/inv.cpp
```diff
@@ -2003,7 +2003,7 @@ int SyncPutItem(int pnum, int x, int y, int idx, WORD icreateinfo, int iseed, in
 	if (!PutItem(pnum, x, y))
 		return -1;
 
-	CanPut(x, y);
+	assert(CanPut(x, y));
```
thebigMuh, line 2006: That will omit the call in release builds. Pull the call outside, capture the return value, and assert on that. (Unless you really want to have CanPut only called in debug builds).

thebigMuh (approved):

Source/inv.cpp
AJenbo: Yeah, it should only be in debug, this clearly comes from an `assert()` in the original code since it matches the assert in the corresponding local function and `CanPut()` has no side effects.

Issue #1657, opened 2021-04-23
Distant sound sources aren't panned correctly


FluffyQuack:
Example: https://www.youtube.com/watch?v=rYuxCazj9KU

When the warrior is far away, you can hear the sound is suddenly played on both channels rather than left.

I'm not sure why this happens. I tested it in original Diablo, and panning seems to work just fine there, so I think this issue is unique to DevilutionX.

julealgon:
@soundchaser83 FYI

PR #1660, opened 2021-04-23
Fix for "Distant sound sources aren't panned correctly"

thebigMuh:
This fixes exclusively the panning bug, where after a certain distance is exceeded, sound samples get played without panning.
Fixes https://github.com/diasurgical/devilutionX/issues/1657


AJenbo:
Nice work

PR #1667, opened 2021-04-24
🧹 GetPlrGFXSize: unordered_set -> direct test

glebm:
The use of `unordered_set` here is a bit gratuitous and doesn't do much for readability.


AJenbo (approved):
Sure this works too :)

PR #1670, opened 2021-04-24
♻️Cleanup player temp variables

AJenbo:


AJenbo:
@MBeijer do you know if `std::clamp` is not available on Amiga?

```
/root/repo/Source/automap.cpp: In function 'void devilution::{anonymous}::SearchAutomapItem(const devilution::CelOutputBuffer&)':
/root/repo/Source/automap.cpp:260:21: error: 'clamp' is not a member of 'std'
  const int startX = std::clamp(x - 8, 0, MAXDUNX);
                     ^~~
```

Source/engine.h
```diff
@@ -27,6 +27,16 @@
 
 namespace devilution {
 
+#if defined(__clang__) || !defined(__GNUC__) || __GNUC__ > 7 || (__GNUC__ == 7 && __GNUC_MINOR__ < 0)
```
glebm, line 30: We should probably use a feature testing macro here:

```suggestion
#if defined(__cpp_lib_clamp)
```

https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros

glebm:
I'm afraid this isn't a scalable strategy. For example, SDL_aulib uses `std::clamp`.
We need Amiga to actually support C++17.

Source/engine.h
AJenbo: oh wow was hoping something like that existed, but couldn't find anything

PR #1673, opened 2021-04-24
Enable all warnings for MSVC

AJenbo:


thebigMuh (approved):
YES

AJenbo:
@thebigMuh MSVC is spewing a lot of warnings so I ended up only increasing it by 1 level, we can increase it again once those are solved :)

PR #1674, opened 2021-04-24
Add functions to get a string representation from enum entries

Jmgr:
This is useful for debugging (as logging an enum value will display a string instead of a number in https://github.com/diasurgical/devilutionX/pull/1571, so this PR is a prerequisite) but also for exposing those enums to a scripting engine later.

Note that I have not considered flags (we should probably migrate to std::bitset) and unnamed enums (they should probably be named).


glebm:
These should probably all be marked `[[allow_unused]]` so that the compiler and clang-tidy do not warn about unused functions

Jmgr:
> These should probably all be marked `[[allow_unused]]` so that the compiler and clang-tidy do not warn about unused functions

You mean [maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused)?

glebm:
1. If this is for debug logging, isn't it better for strings to be the same as the enum keys?

2. What do you think about putting them in a separate file and checking in the code that generates them?

3. We probably don't need to log all of these. For example, stuff like DiabloUI enums?

glebm:
It might be neater to adjust the code generator to put the key and the value on the same line. And we can exclude the generated file from `clang-format`.

Jmgr:
>     1. If this is for debug logging, isn't it better for strings to be the same as the enum keys?

The issue is that some enum entries are difficult to understand. For instance `UITYPE_GRISWOLD` is actually `Griswold's Edge`. Having a human readable string would also help with mods.

> 
>     2. What do you think about putting them in a separate file and checking in the code that generates them?

No because the generation is still a partially manual process, you have to copy and paste the enum and then do the same with the result. The small tool I have written has some issues with comments, some enums are aliases, and you have to manually fix a few things here and there. So I think the functions should stay close to the enum so people that change the enum can also update it.

>        Having to update them manually is error-prone and a maintenance burden.

I offered multiple reliable solutions (see the Discord) but they were all refused, so maintenance burden it is. ;)

> 
>     3. We probably don't need to log all of these. For example, stuff like DiabloUI enums? Though if it's in a separate file then I guess it doesn't matter.

Why not? You never debug the DiabloUI? You also don't expect mods to ever change it?

> It might be neater to adjust the code generator to put the key and the value on the same line. And we can exclude the generated file from clang-format.

As I wrote before, the process is partially manual, so I can't just run it on all files and watch the result. I would prefer having clang-tidy automatically applying that style to those functions to be honest.

qndel:
>The issue is that some enum entries are difficult to understand. For instance UITYPE_GRISWOLD is actually Griswold's Edge. Having a human readable string would also help with mods.

Why not rename to human readable enums? :D

Jmgr:
> > The issue is that some enum entries are difficult to understand. For instance UITYPE_GRISWOLD is actually Griswold's Edge. Having a human readable string would also help with mods.
> 
> Why not rename to human readable enums? :D

Sure, but that would be in another PR I think, one step at a time :)

glebm:
> The issue is that some enum entries are difficult to understand. For instance UITYPE_GRISWOLD is actually Griswold's Edge. Having a human readable string would also help with mods.

This is going about it backwards. If an enum key is badly named, renaming the enum key is better.

> I offered multiple reliable solutions (see the Discord) but they were all refused, so maintenance burden it is. ;)

The 2 solutions I've seen are:

1. better_enum - rejected because it breaks compiler switch exhaustiveness check (that's really bad).
2. magic_enum - rejected because it relies on compiler extensions and we do have ports that use other compilers.

Looking at this PR, I'm starting to reconsider `magic_enum` but I think we need to first understand how it behaves on compilers that are not supported; e.g. we can try Embarcadero C++Builder

glebm:
It looks like `magic_enum` simply returns empty string for the name on unsupported compilers:

https://github.com/Neargye/magic_enum/blob/5b282dd4e86a7874427760e6ce6d1f684b9f1b30/include/magic_enum.hpp#L322

Jmgr:
> It looks like `magic_enum` simply returns empty string for the name on unsupported compilers:
> 
> [Neargye/magic_enum@`5b282dd`/include/magic_enum.hpp#L322](https://github.com/Neargye/magic_enum/blob/5b282dd4e86a7874427760e6ce6d1f684b9f1b30/include/magic_enum.hpp#L322)

That's not idea, it would be better if it returned the enum's entry numerical value at least.

glebm:
That should be possible in principle but might be a little tricky, as it would require compile-time int-to-string conversion to make a static `string_view`

Jmgr:
Not sure what will be chosen in the end, but the consensus seems to be that this PR's contents are not useful, so closing.

AJenbo:
I defiantly see pros with it, but not feeling it outweighs the cons, sorry :( Thanks for the effort, it's good to have considered the options.

Issue #1672, opened 2021-04-24
Chest trap in catacombs hitting multiple times depending on what direction it's opened from (elemental bug?)


Taohn:
**Important information**
Windows 10 Pro 64 / DevilutionX version 1.2.1 /  downloaded release

**Describe**
It's a small chest with a fire arrow trap (the one that launches from the chest itself rather than the wall). Normally it deals 10-15 damage or so with a single projectile, but around half the time the character will be hit multiple times in quick succession (several of which are blocked), dying almost instantly. It can happen from any direction but seems to do so most often from the position in the attached save.  I've had the same thing happen with nova traps ([single_1.zip](https://github.com/diasurgical/devilutionX/files/6368993/single_1.zip)), but that at least makes some kind of sense since the projectiles could be overlapping. 

Similar problem referenced [here ](https://www.lurkerlounge.com/forums/thread-12422.html) under "FIRE/LIGHTNING SUFFIX ISSUES" 

**To Reproduce**

1. Load save [single_1.zip](https://github.com/diasurgical/devilutionX/files/6368986/single_1.zip) 
2. Click on nearby chest (reload if death doesn't occur; it will within a few tries at most)

**Expected behavior**
For character to be hit once rather than multiple times.

qndel:
Thanks!

Taohn:
Forgot to add, this is Hellfire.

qndel:
Found a small issue with your save :D was wondering why the instakill was seemingly random, turns out there are hiddens on the level and they are always active = call rng for movement, that's why you don't always get killed by that chest (And I was starting to think devilutionx's rng broke :D )

qndel:
![image](https://user-images.githubusercontent.com/14297035/115955912-de667280-a4f9-11eb-87bf-2af586e09f0f.png)
[single_31.zip](https://github.com/diasurgical/devilutionX/files/6369415/single_31.zip)
This save always gets the instakill (I killed all monsters on the level)
I marked spots where character gets instakilled on the picture, maybe it's related to arrow hitting a wall and exploding and character being caught in the explosion multiple times?

qndel:
@StephenCWills actually I got a wild theory https://github.com/diasurgical/devilutionX/pull/1641
The instakill only happens in this situation:
1. you block the missile
2. you are standing next to a wall/something blocking.

So you block and deflect the missile, it flies further, hits the wall, you are still recovering from the hit = can't block again, get hit repeatedly from the explosion = dead.

[...]

PR #1675, opened 2021-04-24
fix trap multihit

qndel:
If you ignore whitespaces, I only added this check to monster/player hits by traps
```cpp
if (!missile[i]._miHitFlag) { // traps can hit only once
```
This will still make you get hit once after blocking in hellfire
Together with https://github.com/diasurgical/devilutionX/pull/1641 it should fully fix the trap multihit issue

Resolves https://github.com/diasurgical/devilutionX/issues/1672

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1043,50 +1043,52 @@ void CheckMissileCol(int i, int mindam, int maxdam, bool shift, int mx, int my, 
 			}
 		}
 	} else {
-		if (dMonster[mx][my] > 0) {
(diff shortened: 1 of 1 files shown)


StephenCWills:
With this fix, you'll get hit by the arrow but not the elemental damage.

qndel:
> With this fix, you'll get hit by the arrow but not the elemental damage.

didn't even know the damage was split like that lol

qndel:
This fix also stops trap firewall damage after just 1 hit :(

PR #1676, opened 2021-04-24
Language support via lightweight MO parser

john-tornblom:
This pull request adds initial multilingual support by using a lightweight MO parser that should be usable on most platforms. The choice of language is configured in diablo.ini, e.g., for Swedish: 
```
[Language] 
Code=sv
```
PO-files are kept in the `Translations` folder under the project root dir, and can me compiled to MO files by appending the option `-DUSE_GETTEXT=1` to cmake. 

Note that there is currently no concept of domains, all translations must reside in a single file for a given language. Furthermore, I did not investigate how plural forms should be dealt with. Hopefully, these limitations are negligible for our purposes.


AJenbo:
we will probably need plural support, i don't think we have a real need for domains. in any case if this builds then we have a huge improvement to the translation situation.

AJenbo:
```
  language.cpp
C:\projects\devilutionx\Source\language.cpp(91): error C2131: expression did not evaluate to a constant [C:\projects\devilutionx\devilutionx.vcxproj]
  C:\projects\devilutionx\Source\language.cpp(91): note: failure was caused by a read of a variable outside its lifetime
  C:\projects\devilutionx\Source\language.cpp(91): note: see usage of 'head'
C:\projects\devilutionx\Source\language.cpp(100): error C2131: expression did not evaluate to a constant [C:\projects\devilutionx\devilutionx.vcxproj]
  C:\projects\devilutionx\Source\language.cpp(100): note: failure was caused by a read of a variable outside its lifetime
  C:\projects\devilutionx\Source\language.cpp(100): note: see usage of 'head'
```

CMakeLists.txt
```diff
@@ -245,6 +246,7 @@ set(devilutionx_SRCS
   Source/inv.cpp
   Source/itemdat.cpp
   Source/items.cpp
+  Source/language.cpp
```
AJenbo, line 249: move to utils

Source/utils/language.cpp
```diff
@@ -0,0 +1,129 @@
...
+	if (it == map.end()) {
+		return key;
+	}
+
+	return it->second;
+}
+
+void LanguageInitialize()
+{
+	auto path = GetLangPath() + "./" + sgOptions.Language.szCode + ".gmo";
```
AJenbo, line 69: why does it look for files with .gmo instead of .mo?

john-tornblom: The CMake wrapper around gettext gives them this name. The g in .gmo stands for GNU, and is briefly mentioned at https://www.gnu.org/software/gettext/manual/html_node/Files.html

As far as I understand it, a GMO file is essentially the same thing as a MO file, but they are produced by GNU gettext. I have not yet figured out if there are practical differences. In any case, the parser was implemented based on the description from https://www.gnu.org/software/gettext/manual/html_node/MO-Files.html, which I assume reflects GMO files correctly.

Source/utils/paths.h
```diff
@@ -7,12 +7,14 @@ namespace devilution {
 const std::string &GetBasePath();
 const std::string &GetPrefPath();
 const std::string &GetConfigPath();
+const std::string &GetLangPath();
 const std::string &GetTtfPath();
 const std::string &GetTtfName();
-
+
```
AJenbo, line 13: ?
```suggestion

```

AJenbo (approved):

AJenbo:
Rebased and merged via https://github.com/diasurgical/devilutionX/commit/51e7a169a634e1db37de7cc9adde0f25e0371d20 https://github.com/diasurgical/devilutionX/commit/04269dedcd2aa26ad83483762074d8d95b72dfe2 https://github.com/diasurgical/devilutionX/commit/b4d05034990d327c0d1dc260b9e7b1025e716991 https://github.com/diasurgical/devilutionX/commit/b86743db25a14ab20e16f8e0fbf0604b91e89181

AJenbo:
Played around with it a bit, it mostly works, but there are a few things that needs to be ironed out:
Lots of warnings about buffer-overflow:
image

AJenbo:
Some texts do not appear translated:
image
I'm guessing this could be related to the memory leakage...

john-tornblom:
I think it might be an issue with null-termination, I will submit a patch soon. I also noticed the charset for the Swedish translation was changed to utf-8. Was this intentional? I'll add a parser for the metadata aswell, so we can query the charset and do conversions if needed, but I guess linking with iconv may be problematic on some platforms?

AJenbo:
yeah we need to process UTF8 content in order to support multiple languages in a sane way.

This means we need to expand the text rendering functions to handle this. The plan is to create font files for each Unicode SMB block https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane
So we needa function that will allow us to look this up, idk if this means converting strings to UTF-16 at runtime or how exactly to implement the lookups. Let me know if you would be interested in helping with this as well.

PR #1678, opened 2021-04-24
fix queuing operating object when clicking during attack

qndel:
Fixes multiple object interactions after an attack if you click while attacking
crazydoors


Source/player.cpp
```diff
@@ -3504,8 +3504,6 @@ void CheckNewPath(int pnum)
 				if (object[i]._oBreak == 1) {
 					d = GetDirection(plr[pnum].position.current.x, plr[pnum].position.current.y, object[i]._ox, object[i]._oy);
 					StartAttack(pnum, d);
-				} else {
```
julealgon, line 3507: What happens if we just remove the entire `ACTION_OPERATE` block? Seems like it only exists for melee attacks (for some weird reason) even though you can be on melee while casting or while firing a bow.

qndel: well I guess it wouldn't allow you to break barrels fast

julealgon: Oh is breaking barrels an operate instead of an attack?! Interesting! So _that's_ why this was here?

julealgon (approved):

Source/player.cpp
qndel: well, it performs a check for breaking and barrels are the only breakable object in the game iirc : P

PR #1680, opened 2021-04-24
Remove "AnimWidth2" expect for missiles

obligaron:
I looked in generalization the animation distribution logic and I asked myself "what is Width2 and should I care?".
Mostly the answer is we shouldn't care.
AnimWidth2 seems to be only needed for rendering and can be calculated only there, so the GameLogic don't need to care about.
This reduces complexity and memory size.

But there is a exception to this rule.
it's not possible for missiles, cause in hellfire there are some missiles that have not the normal offset (for example MFILE_REFLECT, MFILE_ARCHLICH or MFILE_BONEDEMON).
But I didn't looked into this in detail.

I submitted this as draft to discuss if we should take this, even if missiles are handled differently.

### DIFF
--- a/Source/dead.cpp
+++ b/Source/dead.cpp
@@ -30,7 +30,6 @@ void InitDead()
 				dead[nd]._deadData[d] = Monsters[i].Anims[MA_DEATH].Data[d];
 			dead[nd]._deadFrame = Monsters[i].Anims[MA_DEATH].Frames;
 			dead[nd]._deadWidth = Monsters[i].width;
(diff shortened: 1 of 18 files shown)


AJenbo (dismissed):

AJenbo:
let me know when it's ready :)

obligaron:
If you are okay that missiles can't get rid of width2 this is ready 🙂 
I also tested that it's still get loaded correctly in vanilla (with a new created hero in devilutionx).

AJenbo:
Removing most of the garbage still removes most of the garbage

PR #1681, opened 2021-04-24
Modernize deprecated headers

Jmgr:
Mostly [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/modernize-deprecated-headers.html#modernize-deprecated-headers) but also a few manual fixes (including a duplicate inclusion).


AJenbo (approved):
Nice, had noticed there might be something not idea about some of the includes but didn't take the time to look in to it :)

PR #1687, opened 2021-04-25
Parse language metadata

john-tornblom:
This pull request adds language metadata lookup capabilities, e.g., to check which character set a translation is encoded in. There is also a fix that ensures translations are null-terminated.


john-tornblom:
This is broken, will fix when I get a couple of minutes

john-tornblom:
Now metadata parsing works as intended

AJenbo (approved):

AJenbo (approved):

PR #1688, opened 2021-04-25
Enable Unittests in Visual Studio

obligaron:
I couldn't run UnitTests in Visual Studio.
This are the changes I needed to enable this.
If there is another way or this is already supported and didn't find it, just tell me.
Notes:

- Visual Studio need testadapter for google test installed.
- At my local machine I run in some problems with fmt, but I didn't included this cause it's another topic and perhaps I do something wrong. :wink:

grafik


AJenbo (approved):

glebm:
> At my local machine I run in some problems with fmt, but I didn't included this cause it's another topic and perhaps I do something wrong.

The fmt CMake rules were incorrect, #1691 includes the fixes

Issue #1686, opened 2021-04-25
Enable more text for translation


AJenbo:
In order to expose the game text to translation, it has to be encapsulated in either `_()`. We need to go over all strings in the application and do this. Probably there is a tool out there that can help automate this task, so it might be worth it to do a bit of research first.

Note that when text is static it cannot be directly translated and it then has to be wrapped with the fake macro `N_()` and at its first usage the variable should be wrapped with the actual translate function `_()`.

When submitting a PR for this try to break it up into a few commits so that it's more manageable to review.

An example of this work can be found here:
https://github.com/diasurgical/devilutionX/commit/dbfa9c27a54856b97a1955ca44cd31f7cfffccca

john-tornblom:
One such tool can be found here: https://gist.github.com/john-tornblom/5a9db5a3ea7b34890858517b3be86b5e

The tool will transform all string literals in the form "x" in the global scope to N_("x"), and locally scoped literals to _("x"). Make sure you build devilutionX first, the script depends on build/compile_commands.json.

The script requires python3 with clang bindings. In Ubuntu:
   sudo apt-get install python3-clang

john-tornblom:
I would run it myself, but I'm not that well acquainted with the devilution source code, so I cannot determinate which strings should be excluded from translation (hard-coded file names are easy, but there seems to be other less obvious once)

john-tornblom:
I'll go through those that seem obvious to me and submit a PR for review

NikoVP:
Reading this I presume not all of the text is ready for translation? What is missing? The Hellfire content and the recently transcribed missing quests only or there is more?

AJenbo:
> Reading this I presume not all of the text is ready for translation? What is missing? The Hellfire content and the recently transcribed missing quests only or there is more?

only menus, shops and towners are currently exposed.

AJenbo:
> I'll go through those that seem obvious to me and submit a PR for review

Thanks that will make it easier for me to do the remaning

AJenbo:
@john-tornblom thanks :)

PR #1689, opened 2021-04-25
Swedish text from psx release

john-tornblom:
In-game alignments etc has not been validated


AJenbo:
Thanks, but we can't distribute the original assets, you would need to either make a script that people can use or create your own translation.

john-tornblom:
Alright, no problem. I'll see what I can wrap up. I noticed that the strings in textdat.cpp are not passed through N_(), Does this have something to do with the extra " |" at the end of each string?

AJenbo:
No, simply this task not having been finished: https://github.com/diasurgical/devilutionX/issues/1686

PR #1693, opened 2021-04-25
Remove globals/pointers from paths.h

Jmgr:
This removes globals and pointers from `paths.h`. It introduces a static class to do so.

Multiple options/alternatives here:
- skip the static, use an instantiable class and pass it around (my preferred option, as we will certainly use more and more classes this will allow us to see what class depends on what; plus the singleton pattern is usually not seen as a good practice) -- the issue here is that some static functions need those paths, so implementing this would imply removing their static status (which would probably be a good thing, but is a bit more involved so I'd rather hear your opinions on that).
- only use static variables and static functions (not a fan, as it's not much better than using globals)
- use `Get` instead of `Instance` to access the instance, shorter but less explicit

### DIFF
--- a/Source/utils/paths.cpp
+++ b/Source/utils/paths.cpp
@@ -23,29 +23,31 @@
 
 namespace devilution {
 
+namespace paths {
+
 namespace {
 
-std::string *basePath = nullptr;
-std::string *prefPath = nullptr;
-std::string *configPath = nullptr;
-std::string *langPath = nullptr;
-std::string *ttfPath = nullptr;
-std::string *ttfName = nullptr;
+std::optional<std::string> basePath;
+std::optional<std::string> prefPath;
+std::optional<std::string> configPath;
+std::optional<std::string> langPath;
+std::optional<std::string> ttfPath;
+std::optional<std::string> ttfName;
 
-void AddTrailingSlash(std::string *path)
+void AddTrailingSlash(std::string &path)
 {
 #ifdef _WIN32
-	if (!path->empty() && path->back() != '\\')
-		*path += '\\';
+	if (!path.empty() && path.back() != '\\')
+		path += '\\';
 #else
-	if (!path->empty() && path->back() != '/')
-		*path += '/';
+	if (!path.empty() && path.back() != '/')
+		path += '/';
 #endif
 }
 
-std::string *FromSDL(char *s)
+std::string FromSDL(char *s)
 {
-	auto *result = new std::string(s != nullptr ? s : "");
+	std::string result = (s != nullptr ? s : "");
 	if (s != nullptr) {
 		SDL_free(s);
 	} else {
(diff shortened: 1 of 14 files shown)


glebm:
A singleton class is no different from globals, just more verbose

Jmgr:
> A singleton class is no different from globals, just more verbose.
> If you want to group these functions together, a namespace is a cleaner option.

Not really, as it allows you to automatically run initialization code the first time it is used. Here the paths are fetched only once, instead of being individually checked. The idea here was not only to group, but to also use that feature.

Jmgr:
Fair enough, I'm ok to convert that as free functions in namespaces if there is a strong feeling against using a class here. Not too happy about the singleton either, and passing an instance is inconvenient because of some call sites.

AJenbo (dismissed):

glebm:
Doesn't this change the behaviour somewhat? Previously, a path could be set to an empty string via the command-line argument; now it can't be anymore.

Perhaps the globals should all be `std::optional<std::string>`?

Source/utils/paths.cpp
```diff
@@ -1,6 +1,8 @@
 #include "utils/paths.h"
 #include "utils/log.hpp"
 
+#include <optional>
+
 #include <SDL.h>
```
glebm, line 6: Fix include order:

1. corresponding header
2. stdlib
3. system dependencies
4. our code

```suggestion
#include "utils/paths.h"

#include <optional>

#include <SDL.h>

#include "utils/log.hpp"
```

glebm:
Looks like `.value()` is only available from macOS 10.14 but you can use `*` instead

Source/utils/paths.cpp
```diff
@@ -55,98 +60,89 @@ std::string *FromSDL(char *s)
...
 #endif
+	}
 	return *basePath;
 }
 
-const std::string &GetPrefPath()
+const std::string &PrefPath()
 {
-	if (prefPath == nullptr)
+	if(!prefPath.has_value())
```
glebm, line 78: Please apply `clang-format` to the file

glebm (changes requested):

Source/utils/paths.cpp
```diff
@@ -1,6 +1,11 @@
 #include "utils/paths.h"
 
-#include <optional>
+#if !defined(__clang__) && defined(__GNUC__) && __GNUC__ < 7
```
glebm, line 3: Not thrilled about this but if we're gonna do it we should use the `#if defined(__cpp_lib_optional)` feature detection macro.
https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros

glebm (changes requested):

Source/utils/paths.cpp
Jmgr: That doesn't seem to work, `__cpp_lib_optional` is not defined only -after- including `<optional>`, so that won't work if that header is not present.

glebm: Ah, make sense. Let's use `__has_include` then; something like this:

```c++
#ifdef __has_include
#if defined(__cplusplus) && __cplusplus >= 201606L && __has_include(<optional>)
#define HAVE_STD_OPTIONAL
#elif __has_include(<experimental/optional>)
#define HAVE_EXPERIMENTAL_OPTIONAL
#endif
#endif

#if defined(HAVE_STD_OPTIONAL)
#include <optional>
#elif defined(HAVE_EXPERIMENTAL_OPTIONAL)
#include <experimental/optional>
#define optional experimental::optional
#else
#error "Missing support for <optional> or <experimental/optional>
#endif 
```

and please place into a separate header

Jmgr: Ah nice, I didn't know about `__has_include`. Fancy.

AJenbo: Now I feel silly for not remember reading about that yesterday...

Jmgr: That code doesn't work well with MSVC though: https://ci.appveyor.com/project/AJenbo/devilutionx/builds/38864222 It does support `<optional>`.

Source/utils/stdcompat/optional.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#ifdef _MSC_VER
```
glebm, line 3: Perhaps we should update MSVC version on CI instead

`__cplusplus` is correctly defined as of Visual Studio 2017 version 15.7 Preview 3

https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

Jmgr: `__has_include` is also only supported since MSVC 2017 15.3.

glebm: Good news: AppVeyor already has a recent enough version!
What we're missing is the `/Zc:__cplusplus` compilation flag, so let's add it instead

> You need to compile with the /Zc:__cplusplus switch to see the updated value of the __cplusplus macro. We tried updating the macro by default and discovered that a lot of code doesn’t compile correctly when we change the value of __cplusplus. We’ll continue to require use of the /Zc:__cplusplus switch for all minor versions of MSVC in the 19.xx family.

Jmgr: Trying that.

AJenbo: I'm fine with doing the switch. It's free and 2017 is already so recent that it's not gaining us any street cred.

glebm:
Looks good! Can you please squash it into 2 commits:
One adding the header and adjusting the C++ flags, and another one with everything else

Source/utils/stdcompat/optional.hpp
Jmgr: Nope: `Command line error D8021: invalid numeric argument '/W0 /Zc:__cplusplus'`.
Am I doing something wrong: `target_compile_options(${BIN_TARGET} PRIVATE "/W0 /Zc:__cplusplus")`?

glebm (dismissed):

PR #1695, opened 2021-04-25
Refactor Animation Informations

obligaron:
This pr doesn't change any logic (including if conditions).

Content:

- Move Fields and Logic to new `AnimationInfos` class (every move is in a separate commit and not coupled with other changes)
- Rename Variables (in separate commits)
- Remove not needed local variables (in separate commit)

Notes:

- In the future AnimationInfos should be included in other Types (Monsters, Object, Missiles). But this is outside the scope of this pr.
- `Width` is not included cause in other types (Monsters) `Width` doesn't exist as member. That's also the reason why NewPlrAnim still exists.

### DIFF
--- a/Source/player.h
+++ b/Source/player.h
@@ -398,7 +398,6 @@ void FreePlayerGFX(int pnum);
  */
 void NewPlrAnim(int pnum, BYTE *pData, int numberOfFrames, int delayLen, int width, AnimationDistributionParams params = AnimationDistributionParams::None, int numSkippedFrames = 0, int distributeFramesBeforeFrame = 0);
 void SetPlrAnims(int pnum);
-void ProcessPlayerAnimation(int pnum);
 void CreatePlayer(int pnum, HeroClass c);
 int CalcStatDiff(int pnum);
 #ifdef _DEBUG
--- a/Source/player.h
+++ b/Source/player.h
@@ -415,12 +415,6 @@ enum class AnimationDistributionParams : uint8_t {
 void NewPlrAnim(int pnum, BYTE *Peq, int numFrames, int Delay, int width, AnimationDistributionParams params = AnimationDistributionParams::None, int numSkippedFrames = 0, int distributeFramesBeforeFrame = 0);
 void SetPlrAnims(int pnum);
 void ProcessPlayerAnimation(int pnum);
-/**
- * @brief Calculates the Frame to use for the Animation rendering
- * @param pPlayer Player
- * @return The Frame to use for rendering
- */
-int GetFrameToUseForPlayerRendering(const PlayerStruct *pPlayer);
 void CreatePlayer(int pnum, HeroClass c);
 int CalcStatDiff(int pnum);
 #ifdef _DEBUG
--- a/Source/player.h
+++ b/Source/player.h
@@ -16,6 +16,7 @@
 #include "path.h"
 #include "interfac.h"
 #include "utils/enum_traits.h"
+#include "engine/animationinfo.h"
 
 namespace devilution {
 
@@ -160,24 +161,11 @@ struct PlayerStruct {
 	ActorPosition position;
 	direction _pdir; // Direction faced by player (direction enum)
 	int _pgfxnum;    // Bitmask indicating what variant of the sprite the player is using. Lower byte define weapon (anim_weapon_id) and higher values define armour (starting with anim_armor_id)
-	uint8_t *_pAnimData;
-	int _pAnimDelay; // Tick length of each frame in the current animation
-	int _pAnimCnt;   // Increases by one each game tick, counting how close we are to _pAnimDelay
-	int _pAnimLen;   // Number of frames in current animation
-	int _pAnimFrame; // Current frame of animation.
+	/*
+	* @brief Contains Information for current Animation
+	*/
+	AnimationInfo AnimInfo;
 	int _pAnimWidth;
-	/*
-	* @brief Specifies how many animations-fractions are displayed between two gameticks. this can be > 0, if animations are skipped or < 0 if the same animation is shown in multiple times (delay specified).
-	*/
-	float _pAnimGameTickModifier;
-	/*
-	* @brief Number of GameTicks after the current animation sequence started
-	*/
-	int _pAnimGameTicksSinceSequenceStarted;
-	/*
-	* @brief Animation Frames that will be adjusted for the skipped Frames/GameTicks
-	*/
-	int _pAnimRelevantAnimationFramesForDistributing;
 	int _plid;
 	int _pvid;
 	spell_id _pSpell;
--- a/Source/player.h
+++ b/Source/player.h
@@ -386,33 +386,17 @@ void InitPlrGFXMem(int pnum);
 void FreePlayerGFX(int pnum);
 
 /**
- * @brief Specifies what special logics are applied for a Animation
- */
-enum class AnimationDistributionParams : uint8_t {
-	None = 0,
-	/*
-	* @brief ProcessAnimation will be called after NewPlrAnim (in same GameTick as NewPlrAnim)
-	*/
-	ProcessAnimationPending,
-	/*
-	* @brief Delay of last Frame is ignored (for example, because only Frame and not delay is checked in game_logic)
-	*/
-	SkipsDelayOfLastFrame,
-};
-
-/**
  * @brief Sets the new Player Animation with all relevant information for rendering
-
  * @param pnum Player Id
- * @param Peq Pointer to Animation Data
- * @param numFrames Number of Frames in Animation
- * @param Delay Delay after each Animation sequence
+ * @param pData Pointer to Animation Data
+ * @param numberOfFrames Number of Frames in Animation
+ * @param delayLen Delay after each Animation sequence
  * @param width Width of sprite
  * @param params Specifies what special logics are applied to this Animation
  * @param numSkippedFrames Number of Frames that will be skipped (for example with modifier "faster attack")
  * @param distributeFramesBeforeFrame Distribute the numSkippedFrames only before this frame
  */
-void NewPlrAnim(int pnum, BYTE *Peq, int numFrames, int Delay, int width, AnimationDistributionParams params = AnimationDistributionParams::None, int numSkippedFrames = 0, int distributeFramesBeforeFrame = 0);
+void NewPlrAnim(int pnum, BYTE *pData, int numberOfFrames, int delayLen, int width, AnimationDistributionParams params = AnimationDistributionParams::None, int numSkippedFrames = 0, int distributeFramesBeforeFrame = 0);
 void SetPlrAnims(int pnum);
 void ProcessPlayerAnimation(int pnum);
 void CreatePlayer(int pnum, HeroClass c);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,6 +343,7 @@ set(devilutionx_SRCS
   Source/controls/modifier_hints.cpp
   Source/controls/plrctrls.cpp
   Source/controls/touch.cpp
+  Source/engine/animationinfo.cpp
   Source/qol/autopickup.cpp
   Source/qol/common.cpp
   Source/qol/monhealthbar.cpp
@@ -453,7 +454,7 @@ if(RUN_TESTS)
     test/scrollrt_test.cpp
     test/stores_test.cpp
     test/writehero_test.cpp
-    test/animationinformation_test.cpp)
+    test/animationinfo_test.cpp)
 endif()
 
 add_executable(${BIN_TARGET} WIN32 MACOSX_BUNDLE ${devilutionx_SRCS})
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -5,7 +5,47 @@
  */
 
 #include "animationinfo.h"
+#include "appfat.h"
+#include "nthread.h"
+#include "utils/log.hpp"
 
 namespace devilution {
+
+int AnimationInfo::GetFrameToUseForRendering()
+{
+	// Normal logic is used,
+	// - if no frame-skipping is required and so we have exactly one Animationframe per GameTick
+	// or
+	// - if we load from a savegame where the new variables are not stored (we don't want to break savegame compatiblity because of smoother rendering of one animation)
+	int relevantAnimationFramesForDistributing = RelevantFramesForDistributing;
+	if (relevantAnimationFramesForDistributing <= 0)
+		return CurrentFrame;
+
+	if (CurrentFrame > relevantAnimationFramesForDistributing)
+		return CurrentFrame;
+
+	assert(GameTicksSinceSequenceStarted >= 0);
+
+	float progressToNextGameTick = gfProgressToNextGameTick;
+
+	// we don't use the processed game ticks alone but also the fragtion of the next game tick (if a rendering happens between game ticks). This helps to smooth the animations.
+	float totalGameTicksForCurrentAnimationSequence = progressToNextGameTick + (float)GameTicksSinceSequenceStarted;
+
+	// 1 added for rounding reasons. float to int cast always truncate.
+	int absoluteAnimationFrame = 1 + (int)(totalGameTicksForCurrentAnimationSequence * GameTickModifier);
+	if (absoluteAnimationFrame > relevantAnimationFramesForDistributing) {
+		// this can happen if we are at the last frame and the next game tick is due (nthread_GetProgressToNextGameTick returns 1.0f)
+		if (absoluteAnimationFrame > (relevantAnimationFramesForDistributing + 1)) {
+			// we should never have +2 frames even if next game tick is due
+			Log("GetFrameToUseForRendering: Calculated an invalid Animation Frame (Calculated {} MaxFrame {})", absoluteAnimationFrame, relevantAnimationFramesForDistributing);
+		}
+		return relevantAnimationFramesForDistributing;
+	}
+	if (absoluteAnimationFrame <= 0) {
+		Log("GetFrameToUseForRendering: Calculated an invalid Animation Frame (Calculated {})", absoluteAnimationFrame);
+		return 1;
+	}
+	return absoluteAnimationFrame;
+}
 
 } // namespace devilution
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -81,7 +81,7 @@ void AnimationInfo::SetNewAnimation(uint8_t *pData, int numberOfFrames, int dela
 
 		if (params == AnimationDistributionParams::ProcessAnimationPending) {
 			// If ProcessAnimation will be called after SetNewAnimation (in same GameTick as NewPlrAnim), we increment the Animation-Counter.
-			// If no delay is specified, this will result in complete skipped frame (see ProcessPlayerAnimation).
+			// If no delay is specified, this will result in complete skipped frame (see ProcessAnimation).
 			// But if we have a delay specified, this would only result in a reduced time the first frame is shown (one skipped delay).
 			// Because of that, we only the remove one GameTick from the time the Animation is shown
 			relevantAnimationGameTicksWithSkipping -= 1;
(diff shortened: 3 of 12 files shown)


Source/animationinfos.cpp
```diff
@@ -5,7 +5,47 @@
  */
 
 #include "animationinfos.h"
+#include "appfat.h"
+#include "nthread.h"
+#include "utils/log.hpp"
 
 namespace devilution {
 
+int AnimationInfos::GetFrameToUseForRendering()
```
julealgon: This looks like a pure function. Can we make it `const`?

Source/animationinfos.cpp
```diff
@@ -48,4 +48,78 @@ int AnimationInfos::GetFrameToUseForRendering()
 	return absoluteAnimationFrame;
 }
 
+void AnimationInfos::SetNewAnimation(uint8_t *Peq, int numFrames, int Delay, AnimationDistributionParams params /*= AnimationDistributionParams::None*/, int numSkippedFrames /*= 0*/, int distributeFramesBeforeFrame /*= 0*/)
```
julealgon, line 48: Parameter names should not be PascalCase I think?

julealgon, line 48: What does "Peq" mean?

julealgon, line 48: `params` is a very bad argument name.

Source/animationinfos.cpp
```diff
@@ -48,4 +48,78 @@ int AnimationInfos::GetFrameToUseForRendering()
...
+{
+	pData = Peq;
+	FrameLen = numFrames;
```
julealgon, line 51: This is a very jarring name clash. When extracting the function, we should rename the arguments appropriately imho.

Source/animationinfos.h
```diff
@@ -9,6 +9,21 @@
 
 namespace devilution {
 
+/**
+ * @brief Specifies what special logics are applied for a Animation
+ */
+enum class AnimationDistributionParams : uint8_t {
+	None = 0,
```
julealgon, line 16: Should we just not initialize to 0? Isn't this taken care of automatically?

Source/animationinfos.h
```diff
@@ -9,6 +9,21 @@
 
 namespace devilution {
 
+/**
+ * @brief Specifies what special logics are applied for a Animation
+ */
+enum class AnimationDistributionParams : uint8_t {
```
julealgon, line 15: Any specific reason to have the underlying type be `uint8_t` instead of just `int`?

Source/animationinfos.h
```diff
@@ -41,6 +56,18 @@ class AnimationInfos {
 	 */
 	int GetFrameToUseForRendering();
 
+	/**
+	 * @brief Sets the new Animation with all relevant information for rendering
+
```
julealgon, line 61: Empty line without the * at the front seems non-standard to me.

Source/animationinfos.cpp
```diff
@@ -122,4 +122,18 @@ void AnimationInfos::SetNewAnimation(uint8_t *Peq, int numFrames, int Delay, Ani
...
+	DelayCounter++;
+	GameTicksSinceSequenceStarted++;
+	if (DelayCounter > DelayLen) {
+		DelayCounter = 0;
+		CurrentFrame++;
+		if (CurrentFrame > FrameLen) {
+			CurrentFrame = 1;
+			GameTicksSinceSequenceStarted = 0;
+		}
+	}
```
julealgon, line 133: Would it be possible to use clamp/max/min logic to simplify this block?

test/animationinfos_test.cpp
```diff
@@ -1,7 +1,7 @@
 #include <gtest/gtest.h>
 
 #include "nthread.h"
-#include "player.h"
+#include "animationinfos.h"
```
julealgon, line 4: This looks a bit hard to read for me... should we consider naming it `animation_info` instead?

test/animationinfos_test.cpp
```diff
@@ -53,29 +53,29 @@ struct RenderingData : TestData {
 void RunAnimationTest(int numFrames, int delay, AnimationDistributionParams params, int numSkippedFrames, int distributeFramesBeforeFrame, const std::vector<TestData *> &vecTestData)
 {
 	const int pnum = 0;
-	PlayerStruct *pPlayer = &plr[pnum];
-	NewPlrAnim(pnum, nullptr, numFrames, delay, 128, params, numSkippedFrames, distributeFramesBeforeFrame);
+	AnimationInfos animInfo;
```
julealgon, line 56: Should this be:
```cpp
	AnimationInfos animInfo = plr[pnum].AnimInfos;
```
?

If so, wouldn't that also be bad, since we'd be creating a whole copy of the entire struct in the stack? Shouldn't we be using `const AnimationInfos&` (for immutable) or `AnimationInfos*` (for mutable)?

Source/animationinfos.cpp
```diff
@@ -11,35 +11,32 @@
 
 namespace devilution {
 
-int AnimationInfos::GetFrameToUseForRendering()
+int AnimationInfos::GetFrameToUseForRendering() const
```
julealgon, line 14: This should be done in the commit where the function is moved.

julealgon (changes requested):
The commit split is appreciated.

Source/animationinfos.cpp
```diff
@@ -0,0 +1,11 @@
+/**
+ * @file animationinfos.cpp
+ *
+ * Contains the core animation informations and related logic
```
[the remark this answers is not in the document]
AJenbo: "information" is uncountable, so there is no plural form.

Source/player.h
```diff
@@ -16,6 +16,7 @@
 #include "path.h"
 #include "interfac.h"
 #include "utils/enum_traits.h"
+#include "animationinfos.h"
```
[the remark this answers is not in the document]
AJenbo: We do, clang-tidy will apply the rule but haven't applied the other rules. I suggest we just ignore it for now as it will be corrected automatically anyway.

Source/animationinfos.h
```diff
@@ -17,35 +17,35 @@ class AnimationInfos {
...
 	/*
 	* @brief Number of GameTicks after the current animation sequence started
 	*/
-	int _pAnimGameTicksSinceSequenceStarted;
+	int GameTicksSinceSequenceStarted;
 	/*
 	* @brief Animation Frames that will be adjusted for the skipped Frames/GameTicks
 	*/
-	int _pAnimRelevantAnimationFramesForDistributing;
+	int RelevantFramesForDistributing;
```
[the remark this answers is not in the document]
AJenbo: Coul you make the follow-up comments as a reply to your first comment, or even better edit the first one so that it's both in context and doesn't ballon into a wall of individual comments.

Source/animationinfos.cpp
AJenbo:
```suggestion
void AnimationInfos::SetNewAnimation(uint8_t *peq, int numFrames, int delay, animationDistributionParams params /*= AnimationDistributionParams::None*/, int numSkippedFrames /*= 0*/, int distributeFramesBeforeFrame /*= 0*/)
```

Source/animationinfos.cpp
AJenbo: @julealgon you can mark multiple lines when commenting, that way it's more clear to see what you refer to, especially on the overview page.

test/animationinfos_test.cpp
AJenbo: I suggest we create an engine folder and put it there, then just call it `engine/animation.h`. Then as we refactor `engine.cpp/h`, `scrollrt.cpp/h` and `engine.cpp/h` we can split it up into other files in that folder.

Source/animationinfos.h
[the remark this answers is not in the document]
julealgon: I could, but I actually commented separately on purpose so that each specific point can be discussed in isolation as they are independent points.

Consider it "creating separate threads" if you will.

Source/animationinfos.cpp
julealgon: Ah indeed, forgot about that. Will try to remember it for future comments :S

Source/animationinfos.h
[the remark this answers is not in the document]
AJenbo: Ok, I'm just not quite sure what the "Also" is referring to.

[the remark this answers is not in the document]
julealgon: > I'm just not quite sure what the "Also" is referring to.

Removed.

Source/animationinfos.cpp
obligaron: I restrained the moved-commits, so they don't include anything expect the move.
And const is a small change, but a change.
So I hope it's okay if I don't rewrite the history here. 🙂

julealgon: > And const is a small change, but a change.

The need for const was the move, so they should be done in the same change.

> So I hope it's okay if I don't rewrite the history here. 🙂

You'll have to anyways to fix the other issues.

obligaron: Do you want me to rewrite the history for every change? 🤔 
I would append them as new commits, so they are easier to review and you don't have to look at all changes again.
If you like to squash this pr. I have no problem with that.

Source/animationinfos.cpp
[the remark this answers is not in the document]
obligaron: thx. fixed 🙂

Source/animationinfos.h
```diff
@@ -0,0 +1,51 @@
...
+#pragma once
+
+#include <stdint.h>
+
+namespace devilution {
+
+/*
+* @brief Contains the core animation informations and related logic
+*/
+class AnimationInfos {
```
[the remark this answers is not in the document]
obligaron: thx. fixed 🙂

Source/player.h
```diff
@@ -160,24 +161,11 @@ struct PlayerStruct {
...
 	/*
-	* @brief Number of GameTicks after the current animation sequence started
+	* @brief Contains Informations for current Animation
 	*/
-	int _pAnimGameTicksSinceSequenceStarted;
-	/*
-	* @brief Animation Frames that will be adjusted for the skipped Frames/GameTicks
-	*/
-	int _pAnimRelevantAnimationFramesForDistributing;
+	AnimationInfos AnimInfos;
```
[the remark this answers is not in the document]
obligaron: "Anim" was used before and I stayed with it, cause this variable is used in many places and I dislike "AnimationInfo AnimationInfo" 😉
But if you like I can rename it. What do you think?

Source/animationinfos.h
```diff
@@ -17,35 +17,35 @@ class AnimationInfos {
 	/*
 	* @brief Pointer to Animation Data
 	*/
-	uint8_t *_pAnimData;
+	uint8_t *pData;
```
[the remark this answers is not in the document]
obligaron: I stayed with p, cause Data is not that meaningful and in this case it helps a little bit to understand that it doesn't own the Data?
But if you want I can change it. What do you think?

Source/animationinfos.cpp
julealgon: > Do you want me to rewrite the history for every change? 🤔

That's the best way to preserve a clean history, so yes.

> I would append them as new commits, so they are easier to review and you don't have to look at all changes again.

That would completely defeat three purpose of your split.

> If you like to squash this pr. I have no problem with that.

That's up to Ajenbo. I'd only squash if the history was messed up, which is not the case here. The way you split the changes is not only helpful for reviewing purposes, but for checking the history later too.

Source/animationinfos.h
```diff
@@ -17,35 +17,35 @@ class AnimationInfos {
...
-	* @brief Increases by one each game tick, counting how close we are to _pAnimDelay
+	* @brief Increases by one each game tick, counting how close we are to DelayLen
 	*/
-	int _pAnimCnt;
+	int DelayCounter;
 	/*
 	* @brief Number of frames in current animation
 	*/
-	int _pAnimLen;
+	int FrameLen;
```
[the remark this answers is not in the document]
obligaron: Renamed to `NumberOfFrames` 🙂

Source/animationinfos.h
```diff
@@ -0,0 +1,51 @@
...
+* @brief Contains the core animation informations and related logic
+*/
+class AnimationInfos {
+public:
+	/*
+	* @brief Pointer to Animation Data
+	*/
+	uint8_t *_pAnimData;
+	/*
+	* @brief Tick length of each frame in the current animation
```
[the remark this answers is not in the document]
obligaron: This was the original description and yes it's misleading. :grin:
I renamed it. Could you take a look?

What do you mean with last? It's not in the comment/this line.

Source/animationinfos.h
```diff
@@ -17,35 +17,35 @@ class AnimationInfos {
...
 	/*
 	* @brief Number of frames in current animation
 	*/
-	int _pAnimLen;
+	int FrameLen;
 	/*
 	* @brief Current frame of animation
 	*/
-	int _pAnimFrame;
+	int CurrentFrame;
```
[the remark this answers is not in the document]
obligaron: I would stay with the shorter name, cause it's an int so I hope everyone thinks it's references to the current Frame and in c++ I would think that Index starts with 0 and in the Animation-Logic this is not the case. :upside_down_face:

Source/animationinfos.h
```diff
@@ -17,35 +17,35 @@ class AnimationInfos {
...
 	/*
 	* @brief Specifies how many animations-fractions are displayed between two gameticks. this can be > 0, if animations are skipped or < 0 if the same animation is shown in multiple times (delay specified).
 	*/
-	float _pAnimGameTickModifier;
+	float GameTickModifier;
 	/*
 	* @brief Number of GameTicks after the current animation sequence started
```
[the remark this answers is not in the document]
obligaron: I would stay with gametick as it is more clear.
Also I thought you don't like abbreviations? 😉

test/animationinfos_test.cpp
obligaron: I moved it to new engine-folder but stayed with `AnimationInfo` cause
- It doesn't contain/own the Animation-Metadata (Gfx, Number-of-Frames for all Animations) so I would like to distinguish both.
- I would like to name the file identically to the containing class.

But if you want I can rename only the file or both. Just say so.

Source/animationinfos.cpp
obligaron: Fixed.
Do you want to also adjust `NewPlrAnim`?
That was the original location.

obligaron: Renamed

Source/animationinfos.cpp
obligaron: Fixed 🙂

Source/animationinfos.h
obligaron: Fixed

test/animationinfos_test.cpp
obligaron: This UnitTest only thest `AnimationInfo`.
And `AnimationInfo` will be used in other places too.
Thats why I dropped using the `PlayerStruct`.

Source/animationinfos.cpp
AJenbo: Yeah we are in the process of unifying the style used in the project, sorry for the current mess.

Source/animationinfos.cpp
obligaron: I'm not that familiar with std::clamp but I would say no, cause

- we don't have a upper or lower bound that we can't exceed. If we got the maximum value we start from the beginning (lowest value).
- this helps with debugging as you can set a breakpoint if we get a new frame
- the logic could be advanced/changed if we expand the usage of `AnimationInfo` to other types. So perhaps we could do this in a later cleanup.

Source/animationinfos.cpp
AJenbo: You can also prefix the commits with "fixup" then they will be combined when merging.

Source/animationinfos.cpp
obligaron: No Problem. fixed 🙂

Source/animationinfos.h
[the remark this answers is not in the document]
AJenbo: Tick is generally used in the application when referring to the core game logic progression.

Source/animationinfos.cpp
obligaron: Nice. 👍  Didn't know about --fixup.
I'm still learning to work with git.
In the next fixup I will try it out. Thanks 🙂

Source/animationinfos.h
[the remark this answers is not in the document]
obligaron: The member doesn't specify how many frames are to distribute.
It specifies which Animation Frames are to be adjust (normal shortened) to compensate the skipped Frames.

Would something like `FramesToModifieForSkippedFrames` or `FramesToAdjustForSkippedFrames` help?
See also [animationinfo.cpp#66](https://github.com/diasurgical/devilutionX/pull/1695/files#diff-7cb2d4eccfa51b37e493533222462f44601d98dddc7dde5bf784be0edef6577aR63)

Source/animationinfos.h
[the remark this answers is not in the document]
obligaron: I found some other occurrences of `game tick` in existing comments (for example `int _mirange; // Time to live for the missile in game ticks, oncs 0 the missile will be marked for deletion via _miDelFlag`). That's why I stayed with that.
Should i only change the variables from `GameTick` to `Tick` or also adjust all comments (change `game tick` to `tick`)?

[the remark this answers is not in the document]
AJenbo: Adjusting the var name is plenty fine. Documentation is another thing. It's where you would look if you are unsure what kind of tick it is :)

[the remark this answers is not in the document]
obligaron: 👍 
I changed GameTick to Tick and adjusted comments from "GameTick" to "game tick" 🙂

Source/player.h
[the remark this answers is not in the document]
julealgon: I'd suggest simply `Animation` for the field name in this case.

Source/animationinfos.h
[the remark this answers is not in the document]
julealgon: I expect a variable named "CurrentFrame" to be of some high abstraction type named "XFrame", or "YFrameData" or similar, and not a raw int. 

An int is not a frame, its a number.

What you have here is an indexer into an array of frames, and that's where the rename recommendation comes from: make the variable name better show it's intent.

Source/animationinfos.h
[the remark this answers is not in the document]
julealgon: > It specifies which Animation Frames are to be adjust (normal shortened) to compensate the skipped Frames.

But that sounds like a list to me. How is it a single int then? If it specifies "which frames", it would need to be a list of frames no?

test/animationinfos_test.cpp
julealgon: > It doesn't contain/own the Animation-Metadata (Gfx, Number-of-Frames for all Animations) so I would like to distinguish both.
> I would like to name the file identically to the containing class.

The relationship between header and class in C++ is not necessarily 1 to 1. Usually, you have a header that groups a certain number of elements related to the same "area". In that sense, naming the header just `animation.h` (or `animation.hpp`... whatever is the current standard) is acceptable to me.

I'd totally agree with you if this was in another context, like C#, where it is a best practice to have one file per class.

Source/animationinfos.cpp
julealgon: > You can also prefix the commits with "fixup" then they will be combined when merging.

What?! I've never heard of this as well! Is that a feature of Github specifically @AJenbo , or does it work on raw git too?

AJenbo (approved):

Source/animationinfos.h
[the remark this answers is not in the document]
obligaron: It implies Frame 1 to this specified frame. With that we have a list. 😉 
In theory it could also be frame 4 to 6 or something like this.
But in Diablo I don't think we ever need this. So I didn't complicate the logic.

Source/animationinfos.cpp
AJenbo: Don't have anything to do with GitHub. I thought I already mentioned it to you in the chat :)

To apply the fixups you just run:
`git rebase origin/master --autosquash -i`

julealgon: I use fixup all the time while rebasing, but never knew you could also use by modifying the commits themselves. Very cool.

AJenbo:
Ok, I have to admit that was a lot harder to rebase than I had expected, mostly because of all the scattered renames of files and variables (information -> infos -> info).

In the future try to apply the review changes to the relevant commits if possible. It's not relevant to see the intermediate commits once merged, in the future, it will just be confusing to see things renamed multiple times over a couple of days. Regarding review diff, GitHub gives a convenient diff from last time the PR was viewed, both as a notification but also in the history of the overview page (click on force-pushed) so that isn't really a factor.

 One mistake I did was rebase on the latest master, so the diff actually looks a bit drastic as it includes all commits that have happened on master since this was started, but the only actual diff here is correcting some bad indentation in the test comments and correcting the file documentation which still had "AnimationInfos" as the names.

obligaron:
Thank you and sorry for the inconvenience. :cry:
I will try to do it and I hope --fixup will help with that in the future.

AJenbo:
Well we are all learning and adjusting here :)

PR #1697, opened 2021-04-25
All strings translated

AJenbo:
This is not intended for being merged but just as a reference for going over the rest of the strings with a fine comb.


Source/control.cpp
```diff
@@ -1426,7 +1426,7 @@ void DrawChr(const CelOutputBuffer &out)
 	} else {
 		maxdam += plr[myplr]._pDamageMod;
 	}
-	sprintf(chrstr, "%i-%i", mindam, maxdam);
+	sprintf(chrstr, _("%i-%i"), mindam, maxdam);
```
AJenbo, line 1429: Not sure about this, but I guess we will get an error report if needed.

Source/control.cpp
```diff
@@ -1437,7 +1437,7 @@ void DrawChr(const CelOutputBuffer &out)
 	else
 		col = COL_BLUE;
 	if (plr[myplr]._pMagResist < MAXRESIST) {
-		sprintf(chrstr, "%i%%", plr[myplr]._pMagResist);
+		sprintf(chrstr, _("%i%%"), plr[myplr]._pMagResist);
```
AJenbo, line 1440: Probably not needed either.

Source/utils/language.cpp
```diff
@@ -64,15 +64,15 @@ bool parse_metadata(char *data)
 	char *ptr = data;
 	bool utf8 = false;
 
-	while (ptr && (delim = strstr(ptr, ":"))) {
+	while (ptr && (delim = strstr(ptr, _(":")))) {
```
AJenbo, line 67: Ha ha, I love the idea of the translation file parse being translated :joy:

Source/control.cpp
altiereslima: #: Source/control.cpp:1916

altiereslima: ADD_PlrStringXY(out, 366, 121, 600, "you want to remove?", COL_GOLD);

AJenbo: Thanks, i have added it.

PR #1699, opened 2021-04-25
Clean up missile logic

StephenCWills:
I'm trying to break this down into bite-sized chunks. This first one encapsulates player blocking logic into a single function. I'll be adding more so stay tuned.

EDIT: I realized I could do a bit better. I rearranged the logic in `PlayerMHit()` and `Plr2PlrMHit()` first, before introducing the new functions. This helped me find a couple mistakes, and should help clean up the diffs a bit.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -684,6 +684,104 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, int t, bool 
 	return false;
 }
 
+bool HitsPlayer(int playerID, int missileSourceID, int missileTargets, int missileType, int distanceTraveled)
+{
+	if (plr[playerID]._pHitPoints >> 6 <= 0) {
+		return false;
+	}
+
+	if (plr[playerID]._pInvincible) {
+		return false;
+	}
+
+	bool isPhysicalAttack = missiledata[missileType].mType == 0;
+	if (plr[playerID]._pSpellFlags & 1 && isPhysicalAttack) {
+		return false;
+	}
+
+	int chanceToHit = 0;
+	if (isPhysicalAttack) {
+		int totalArmorClass = plr[playerID]._pIAC + plr[playerID]._pIBonusAC + plr[playerID]._pDexterity / 5;
+		if (missileTargets == TARGET_MONSTERS) {
+			int classBonus = 0;
+			if (plr[missileSourceID]._pClass == HeroClass::Rogue)
+				classBonus = 20;
+			if (plr[missileSourceID]._pClass == HeroClass::Warrior)
+				classBonus = 10;
+			if (plr[missileSourceID]._pClass == HeroClass::Bard)
+				classBonus = 10;
+
+			chanceToHit = 50
+			    + plr[missileSourceID]._pLevel
+			    + plr[missileSourceID]._pDexterity
+			    + plr[missileSourceID]._pIBonusToHit
+			    + classBonus
+			    - totalArmorClass
+			    - distanceTraveled * distanceTraveled * 2;
+		}
+		if (missileTargets == TARGET_PLAYERS) {
+			chanceToHit = 30
+			    + monster[missileSourceID].mHit
+			    + monster[missileSourceID].mLevel * 2
+			    - plr[playerID]._pLevel * 2
+			    - totalArmorClass
+			    - distanceTraveled * 2;
+		}
+		if (missileTargets == TARGET_BOTH) {
+			chanceToHit = 100
+			    - totalArmorClass
+			    - distanceTraveled * 2;
+		}
+	} else {
+		if (missileTargets == TARGET_MONSTERS) {
+			int classBonus = 0;
+			if (plr[missileSourceID]._pClass == HeroClass::Sorcerer)
+				classBonus = 20;
+			if (plr[missileSourceID]._pClass == HeroClass::Bard)
+				classBonus = 10;
+
+			chanceToHit = 50
+			    + plr[missileSourceID]._pMagic
+			    + classBonus
+			    - plr[playerID]._pLevel * 2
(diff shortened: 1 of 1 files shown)


AJenbo:
Breaking this into chunks seems like a good idea, else it's rather complicated to verify.

StephenCWills:
> Breaking this into chunks seems like a good idea, else it's rather complicated to verify.

Yeah, I realized it as I got part of the way in. I was targeting an end result as something like...

```
if (CollidesWithPlayer() && HitsPlayer() && PlayerBlocks())
    DoDamageToPlayer();
```

At that point, we could fiddle with the formula to handle special cases in the respective missile processing functions. The problem was that the change was getting to be too big, and I wasn't sure if I'd be able to verify that I hadn't significantly changed gameplay so I'm trying to keep it manageable.

Source/missiles.cpp
```diff
@@ -904,6 +888,26 @@ bool Plr2PlrMHit(int pnum, int p, int mindam, int maxdam, int dist, int mtype, b
...
+			break;
+		case MISR_MAGIC:
+		case MISR_ACID:
+			resper = plr[p]._pMagResist;
+			break;
+		default:
+			resper = 0;
+			break;
+		}
+		if (resper <= 0 && blkper < blk) {
```
StephenCWills: I found an interesting discrepancy here. Players can't block PvP spells if they have nonzero resistance. I assume this has something to do with the fact that Hellfire didn't officially support multiplayer, but I kept the behavior the same so I could highlight this fact.

AJenbo: Feel free to unify if you like.

Source/missiles.cpp
```diff
@@ -861,48 +895,24 @@ bool PlayerMHit(int pnum, int m, int dist, int mind, int maxd, int mtype, bool s
...
-				if (m == -1)
-					if ((plr[pnum]._pIFlags & ISPL_ABSHALFTRAP) != 0)
-						dam /= 2;
-				dam += (plr[pnum]._pIGetHit << 6);
-			} else {
-				dam = mind + GenerateRnd(maxd - mind + 1);
-				if (m == -1)
-					if ((plr[pnum]._pIFlags & ISPL_ABSHALFTRAP) != 0)
-						dam /= 2;
-				dam += plr[pnum]._pIGetHit;
```
StephenCWills, line 790: This part was bothering me. Sometimes `_pIGetHit` is shifted, sometimes it isn't. The same flag that determines whether a missile collision is blockable also determines whether this value (and the min/max damage passed in) should be shifted by this logic. Missiles that do not get shifted include acid puddles, firewall, lightning wall, lightning, flash, inferno, and the explosion from Elemental.

I assumed that the value stored in `_pIGetHit` is normally unshifted, and so the unshifted value would do a negligible amount of damage (1/64th of the normal damage). Because of that, I did not include the unshifted adjustment in the cleaned up logic. However, I do also notice that many of these missiles are multi-hit missiles so the small damage adjustment could add up. Still, I believe that few of these damage sources will hit you 64 times before you either move away, stop getting hit, or die.

If anyone has an opinion, I would appreciate some input.

StephenCWills: In thinking about it a little bit more, if you have Stormshield's +4 Damage from Enemies, an acid puddle or firewall could end up doing +80 damage per second with the shifted value, which seems problematic. I wonder if maybe these attacks should apply the damage adjustment per second. Or in the case of lightning and inferno, use the duration of the missile as the divisor.

StephenCWills: I put the unshifted adjustment back in for now. I still think it's wrong to arbitrarily apply 1/64th of the damage adjustment to all these missiles. In particular, it makes no sense for the explosion from Elemental. Please chime in if you have any thoughts.

StephenCWills:
I have a feeling that none of this matters for the missile collision logic, but I feel the need to document these; just in case. This is all the places where `_micaster` is not one of the `mienemy_type` enum values.
* [monster.cpp#L3022](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/monster.cpp#L3022)
* [monster.cpp#L3105](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/monster.cpp#L3105)
* [monster.cpp#L3293](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/monster.cpp#L3293)
* [monster.cpp#L4043](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/monster.cpp#L4043)
* [objects.cpp#L3577-L3587](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/objects.cpp#L3577-L3587)
* [objects.cpp#L3747-L3757](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/objects.cpp#L3747-L3757)
* [objects.cpp#L3877](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/objects.cpp#L3877)
* [objects.cpp#L4697-L4707](https://github.com/diasurgical/devilutionX/blob/eb4fbdc30246fac46c67736892ee3e19fffbcd90/Source/objects.cpp#L4697-L4707)

AJenbo:
Well this project has been all about documentation so type away :)

StephenCWills:
Honestly, I got completely sidetracked by 3DS work, and given all the cleanup that's been going on across the codebase, I probably just need to start over again. I'll be closing this for now. Hopefully I'll find time to work on this again.

PR #1703, opened 2021-04-25
🧹 Remove SDL2 finder

glebm:
SDL2 ships with CMake config files, so a finder file should not be necessary for it.

Had to add a workaround for mingw that we'll be able to remove with the next release of SDL2


AJenbo:
The less the merrier. This will probably help the Fedora crowd.

glebm:
The `libstdc++` issue does not seem related

glebm:
Ah, I see that it may indeed be related

PR #1709, opened 2021-04-26
🔨 CMake: Fix SDL2 on older Debian derivates

glebm:
Older Debian and derivatives come with sdl2-config.cmake file that does
not define any targets.

One such sdl2-config.cmake looks like this:

```cmake
set(prefix "/usr")
set(exec_prefix "${prefix}")
set(libdir "${prefix}/lib/x86_64-linux-gnu")
set(SDL2_PREFIX "/usr")
set(SDL2_EXEC_PREFIX "/usr")
set(SDL2_INCLUDE_DIRS "${prefix}/include/SDL2")
set(SDL2_LIBRARIES " -lSDL2")
string(STRIP "${SDL2_LIBRARIES}" SDL2_LIBRARIES)
```

Works around this by defining an `SDL2::SDL2` target ourselves.

Removes SDL2::SDLmain because we do not use the SDL_main function and do not need it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,14 +202,20 @@ if(USE_SDL1)
 else()
   find_package(SDL2 REQUIRED)
   if(TARGET SDL2::SDL2)
-    set(SDL2_TARGET SDL2::SDL2)
+    set(SDL2_MAIN SDL2::SDL2main)
   elseif(TARGET SDL2::SDL2-static)
     # On some distros, such as vitasdk, only the SDL2::SDL2-static target is available.
-    set(SDL2_TARGET SDL2::SDL2-static)
     # Alias to SDL2::SDL2 because some finder scripts may refer to SDL2::SDL2.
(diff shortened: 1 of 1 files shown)


AJenbo:
afaik SDL_main is needed for Android, don't think that changes anything here, but just good to note anyway :)

glebm:
Looks like this was also needed on the mingw builds; fixed

AJenbo:
This doesn't exactly make CmakeList.txt smaller though.

glebm:
We've deleted hundreds of lines by removing the finder, still a big win

AJenbo:
Yeah, I agree, but multiple files with distinct content are easier to reason about in some aspects since you can ignore large parts of it.

glebm:
@AJenbo Not in this case, as that file should not have existed in the first place. It tried to manually find SDL headers and such, where this is something provided with the distribution.

We also eliminated system specific hacks in there which were a maintenance burden when porting.

PR #1710, opened 2021-04-26
Split of QoL features into separate files

thebigMuh:
No real changes done to the content, just splitting the two remaining features from qol.cpp (monster health bar and auto gold pickup) out into their own files.


AJenbo:
With all the other changes going it's refreshing with non-complicated change for once :D

glebm:
Looks like these files were checked in with `\n` line endings. We actually use `\r\n`, as configured in `.editorconfig`. It'd be great if you could set up your editor to respect it.

For VS Code, you can use this extension: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

thebigMuh:
> Looks like these files were checked in with `\n` line endings. We actually use `\r\n`, as configured in `.editorconfig`. It'd be great if you could set up your editor to respect it.
> 
> For VS Code, you can use this extension: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

That's very weird. I'm actually on Windows, and I just tried adding a fresh file to the project, and it defaults to CRLF. No idea why those two files are different ¯\_(ツ)_/¯

PR #1711, opened 2021-04-26
Remove AnimLenFromClass

obligaron:
A small cleanup.
`AnimLenFromClass` was a constant (always 8) and defines the same as the values in `PlrGFXAnimLens` (Animation Length).
`_pWFrames` is initialized with `PlrGFXAnimLens`, so just use this.


AJenbo:
While nice to be able to have variable walk cycles, even better than defining them in a table would be to have them defined with the animation itself. This makes that easier to implement :+1:

PR #1714, opened 2021-04-26
Adding more const&

thebigMuh:
Fixed CelOutputBuffer::pitch() to be const, and then spread the constant love around.


AJenbo:
`const love` it's the new C++ idiom to replace the unreliable `unconditional love`

PR #1715, opened 2021-04-26
Add french translation

mmaret:


mmaret:
Still a lot to translate (~50%) but may be it's time for a first review ?

AJenbo:
Looks good so far :)

image

Headers and all look correct, but can't comment on the French :)

Diablo is about 75% of the strings, and Hellfire makes up the remaining 25%

If no native speakers turn up to review it here I'll still merge it once the next series of translation features are ready.

PR #1717, opened 2021-04-26
⚙️ Show FPS option

glebm:
On some platforms, editing command-line flags is difficult.
This adds a `Show FPS` option to the ini file.

If set, FPS is shown even without the `-f` flag.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -482,6 +482,7 @@ static void SaveOptions()
 	setIniInt("Graphics", "Gamma Correction", sgOptions.Graphics.nGammaCorrection);
 	setIniInt("Graphics", "Color Cycling", sgOptions.Graphics.bColorCycling);
 	setIniInt("Graphics", "FPS Limiter", sgOptions.Graphics.bFPSLimit);
+	setIniInt("Graphics", "Show FPS", sgOptions.Graphics.bShowFPS);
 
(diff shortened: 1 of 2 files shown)


Trihedraf (approved):

AJenbo (approved):

Source/diablo.cpp
```diff
@@ -616,6 +618,9 @@ static void diablo_init_screen()
 
 static void diablo_init()
 {
+	if (sgOptions.Graphics.bShowFPS)
+		EnableFrameCount();
```
julealgon, line 622: Why is the `-f` flag handling not in this same place? Shouldn't it be?

It seems bad that we have 2 flags for the same behavior, and they are enabled in different places.

AJenbo: Flags are parsed before reading the settings.

julealgon: > Flags are parsed before reading the settings.

Still, can't we setup the FPS display in the same place? I assume we currently have 2 places in the code that call `EnableFrameCount`, which seems bad to me.

If we have the same setting exposed as a cmdline flag, and as an ini flag, those should be treated as different sources for the same setting and the enablement of it should be centralized.

julealgon: I think we should rethink this. Having the same behavior enabled in different places is a bug waiting to happen.

PR #1719, opened 2021-04-27
🔨 ⚙️  CMake: Add DEFAULT_BLENDED_TRANSPARENCY option

glebm:


AJenbo:
Do you see a meaningful difference in performance? I was considering completely removing the old transparency as I haven't see anyone complain about it. Even hardcore diablo players often do not notice it and rg300 doesn't seem to care one way or the other.

glebm:
It seems to save 0.5 MiB RAM but maybe I'm imagining things?

glebm:
The lookup table should only be 64 KiB; it's definitely variance or similar :|

AJenbo:
> It seems to save 0.5 MiB RAM but maybe I'm imagining things / measuring wrong / variance?

Hmm didn't think of memory usage, but it should hopefully not use more then 64KiB

glebm:
Oh wow but the loop iterates 256 x 256 x 256 = 16.7 million times at most

glebm:
Nevermind then

PR #1725, opened 2021-04-27
Add portable.txt check which will set paths for save and config to ./

Trihedraf:
This PR adds a check for the existence of portable.txt in the devilutionx executable folder. If found it sets the paths for save and config to ./


AJenbo:
creating a shortcut with `--save-dir .` would give you the same and also just requires one file in the folder. portable.txt dosent appear to be a norm either.

alternatives to this pr, look for savegames or ini and do a write test.

PR #1727, opened 2021-04-27
Remove libs linked by SDL2 target from lib list

isage:
Now that we use SDL2 cmake target for linking, we can drop some libs from link list, because they are automatically linked by SDL2 target import.


AJenbo:
Much nicer

PR #1728, opened 2021-04-27
More translation

mmaret:
Use translation for some in game items.

This may need to look at each translation to make sure they fit in the required space.

If we do not want to translate thoses items, we could modify devilutionx.pot template to remove associated string


AJenbo (approved):

AJenbo:
thanks, the reason this hasn't been done yet is that it couldn't be automated and exposing the text to translators was higher priority then waiting for someone to go over all cases :)

PR #1733, opened 2021-04-27
Slight cleanup of dead{.h,.cpp}

vlolteanu:
The only outright intrusive change is the use of an std::array for the animation data.

### DIFF
--- a/Source/dead.cpp
+++ b/Source/dead.cpp
@@ -17,25 +17,22 @@ int8_t stonendx;
 
 void InitDead()
 {
-	int8_t mtypes[MAXMONSTERS];
-
-	for (int8_t &mtype : mtypes)
-		mtype = 0;
+	int8_t mtypes[MAXMONSTERS] = {};
 
 	int8_t nd = 0;
 
 	for (int i = 0; i < nummtypes; i++) {
-		if (mtypes[Monsters[i].mtype] == 0) {
-			for (int d = 0; d < 8; d++)
-				dead[nd]._deadData[d] = Monsters[i].Anims[MA_DEATH].Data[d];
-			dead[nd]._deadFrame = Monsters[i].Anims[MA_DEATH].Frames;
-			dead[nd]._deadWidth = Monsters[i].width;
-			dead[nd]._deadtrans = 0;
-			nd++;
+		if (mtypes[Monsters[i].mtype] != 0)
+			continue;
 
-			Monsters[i].mdeadval = nd;
-			mtypes[Monsters[i].mtype] = nd;
-		}
+		dead[nd]._deadData = Monsters[i].Anims[MA_DEATH].Data;
+		dead[nd]._deadFrame = Monsters[i].Anims[MA_DEATH].Frames;
+		dead[nd]._deadWidth = Monsters[i].width;
+		dead[nd]._deadtrans = 0;
+		nd++;
+
+		Monsters[i].mdeadval = nd;
+		mtypes[Monsters[i].mtype] = nd;
 	}
 
 	for (auto &d : dead[nd]._deadData)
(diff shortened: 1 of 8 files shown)


AJenbo (approved):
Thanks

Source/dead.cpp
```diff
@@ -17,25 +17,22 @@ int8_t stonendx;
...
-			dead[nd]._deadWidth = Monsters[i].width;
-			dead[nd]._deadtrans = 0;
-			nd++;
+		if (mtypes[Monsters[i].mtype] != 0)
+			continue;
 
-			Monsters[i].mdeadval = nd;
-			mtypes[Monsters[i].mtype] = nd;
-		}
+		dead[nd]._deadData = Monsters[i].Anims[MA_DEATH].Data;
```
julealgon, line 28: Is this really equivalent to the old 8 position loop?

Source/dead.cpp
```diff
@@ -72,21 +68,21 @@ void InitDead()
 	assert(nd <= MAXDEAD);
 }
 
-void AddDead(int dx, int dy, int8_t dv, direction ddir)
+void AddDead(Point loc, int8_t dv, direction ddir)
```
julealgon, line 71: Can we rename the parameter to something more descriptive, like `tileLocation` or `tilePosition` or even `tileCoordinate`?

FAKE EDIT: `tilePosition` seems like the best candidate considering our naming for similar values on the `PlayerStruct`.

Source/dead.h
```diff
@@ -5,16 +5,17 @@
...
 
+#include <array>
 #include <cstdint>
 
 #include "engine.h"
 
 namespace devilution {
 
-#define MAXDEAD 31
+static constexpr unsigned MAXDEAD = 31;
```
julealgon, line 15: We should not be using ALL_CAPS for these types of constants.

Source/dead.h
```diff
@@ -5,16 +5,17 @@
...
 
 struct DeadStruct {
-	uint8_t *_deadData[8];
+	std::array<uint8_t *, 8> _deadData;
```
julealgon, line 18: Oh, so this is what enabled the copy on assignment huh.

julealgon (changes requested):

Source/dead.cpp
AJenbo: yes, since the type was changed.

Source/dead.h
AJenbo: Please resolve the comments when the question isn't relevant anymore.

julealgon: I cannot resolve my comments. Maybe I don't have enough permissions?

AJenbo: ok, will look into that.

Source/dead.h
AJenbo:
```suggestion
static constexpr unsigned MaxDead = 31;
```

Source/dead.cpp
AJenbo:
```suggestion
void AddDead(Point tilePosition, int8_t dv, direction ddir)
```

Source/dead.h
galaxyhaxz: one word a.k.a `define` isn't 133t enough in 2021. Instead we need
`static constexpr unsigned uint8_t signed halfsigned` :P

Source/dead.cpp
vlolteanu: Done

Source/dead.h
vlolteanu: Done.

@galaxyhaxz Heh, that's true. However,  if you want to keep everything in the devilution namespace, you have to ditch almost all of your macros.

galaxyhaxz: @vlolteanu Ah, I'm in the process of learning C++ so wasn't sure the exact reason for these changes. Thanks for clarifying :)

vlolteanu:
This is slightly off-topic, but the array used for the animation data shows up in a lot of other places. Maybe it should be typedef-ed, e.g. using AnimationData = std::array<uint8_t, 8>, or something like that.

AJenbo:
Sounds like an idea

PR #1735, opened 2021-04-27
Don't crash on string formatting errors, and be more verbose

### LINKED ISSUE #1734: Startup crash
The game can not be launched at all since commit 1b7e0d2cb3a01c126d5d2d45f10ce7616d8ac984.

The error message after the crash in the Linux terminal is as follows:

terminate called after throwing an instance of 'fmt::v7::format_error'
  what():  string pointer is null

Jmgr:
This catches any format exception thrown by `fmt::format` so that the program doesn't crash, and logs an error message with some information on what the error is.

We try to insert nullptr char* into a format string at some point in the code, which is not great. This PR should help debugging that issue.

Fixes #1734.

### DIFF
--- a/Source/utils/log.hpp
+++ b/Source/utils/log.hpp
@@ -25,25 +25,40 @@ enum class LogCategory {
 constexpr auto defaultCategory = LogCategory::Application;
 
 enum class LogPriority {
-    Verbose = SDL_LOG_PRIORITY_VERBOSE,
-    Debug = SDL_LOG_PRIORITY_DEBUG,
-    Info = SDL_LOG_PRIORITY_INFO,
-    Warn = SDL_LOG_PRIORITY_WARN,
-    Error = SDL_LOG_PRIORITY_ERROR,
-    Critical = SDL_LOG_PRIORITY_CRITICAL,
+	Verbose = SDL_LOG_PRIORITY_VERBOSE,
+	Debug = SDL_LOG_PRIORITY_DEBUG,
+	Info = SDL_LOG_PRIORITY_INFO,
+	Warn = SDL_LOG_PRIORITY_WARN,
+	Error = SDL_LOG_PRIORITY_ERROR,
+	Critical = SDL_LOG_PRIORITY_CRITICAL,
 };
+
+namespace detail {
+
+template <typename... Args>
+std::string format(const char *fmt, Args &&... args)
+{
+	try {
+		return fmt::format(fmt, std::forward<Args>(args)...);
+	} catch (const fmt::format_error &e) {
+		SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "Format error, fmt: %s, error: %s", fmt ? fmt : "nullptr", e.what());
+		return "error";
+	}
+}
+
+} // namespace detail
 
 template <typename... Args>
 void Log(const char *fmt, Args &&... args)
 {
-	auto str = fmt::format(fmt, std::forward<Args>(args)...);
+	auto str = detail::format(fmt, std::forward<Args>(args)...);
 	SDL_Log("%s", str.c_str());
 }
 
 template <typename... Args>
 void LogVerbose(LogCategory category, const char *fmt, Args &&... args)
 {
-	auto str = fmt::format(fmt, std::forward<Args>(args)...);
+	auto str = detail::format(fmt, std::forward<Args>(args)...);
 	SDL_LogVerbose(static_cast<int>(category), "%s", str.c_str());
 }
 
@@ -56,7 +71,7 @@ void LogVerbose(const char *fmt, Args &&... args)
 template <typename... Args>
 void LogDebug(LogCategory category, const char *fmt, Args &&... args)
 {
-	auto str = fmt::format(fmt, std::forward<Args>(args)...);
+	auto str = detail::format(fmt, std::forward<Args>(args)...);
 	SDL_LogDebug(static_cast<int>(category), "%s", str.c_str());
 }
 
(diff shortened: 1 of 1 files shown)


AJenbo:
Wait, fmt makes use of exceptions :(

Source/utils/log.hpp
```diff
@@ -25,25 +25,40 @@ enum class LogCategory {
...
 };
 
+namespace detail {
+
+template <typename... Args>
+std::string format(const char *fmt, Args &&... args)
+{
+	try {
+		return fmt::format(fmt, std::forward<Args>(args)...);
+	} catch (const fmt::format_error &e) {
```
julealgon, line 43: Won't this cause us to mask real issues that need to be fixed ASAP? Is there a real situation where it would make sense to ignore these errors? Aren't they all actual issues?

Not a big fan of swallowing exceptions in this case if the above is true.

AJenbo: I would probably go with `app_fatal()` instead of logging. I'm pretty sure that the case we are doing this for should never happen, so would be good to just get the tests to report it back, and most won't be looking in the log.

AJenbo:
On a side note could you check that we can build with exceptions disabled. Requiring exceptions is preventing me from building on my powerpc using newer compilers:
https://github.com/fmtlib/fmt/commit/8b76e970621cdd2b278fb3e751a368ee417e7627

Source/utils/log.hpp
Jmgr: > Wait, fmt makes use of exceptions :(

Well exceptions are the [official](https://isocpp.org/wiki/faq/exceptions#why-exceptions) way of signaling errors in C++, so they are quite a common sight, even if you don't like them :stuck_out_tongue: 

> Won't this cause us to mask real issues that need to be fixed ASAP? Is there a real situation where it would make sense to ignore these errors? Aren't they all actual issues?
> Not a big fan of swallowing exceptions in this case if the above is true.

> I would probably go with app_fatal() instead of logging. I'm pretty sure that the case we are doing this for should never happen, so would be good to just get the tests to report it back, and most won't be looking in the log.

I agree that this is an important issue. I propose to log it anyway but also to call `app_fatal()` so if someone sends a log file to us it will be easier to diagnose the issue. Generally logging more is good.

> On a side note could you check that we can build with exceptions disabled. Requiring exceptions is preventing me from building on my powerpc using newer compilers:
fmtlib/fmt@8b76e97

I will use the ugly FMT macros then.

I didn't know there were systems that didn't support exceptions. That is very odd. Do you know why?

Jmgr:
> Wait, fmt makes use of exceptions :(

Exceptions are the [official](https://isocpp.org/wiki/faq/exceptions#why-exceptions) way of signaling errors in C++, so it's quite a common sight :stuck_out_tongue: 



> On a side note could you check that we can build with exceptions disabled. Requiring exceptions is preventing me from building on my powerpc using newer compilers:
> [fmtlib/fmt@8b76e97](https://github.com/fmtlib/fmt/commit/8b76e970621cdd2b278fb3e751a368ee417e7627)

I will use the ugly FMT macros then.

I didn't know there were systems that didn't support exceptions. That is very odd. Do you know why?

Jmgr:
> > On a side note could you check that we can build with exceptions disabled. Requiring exceptions is preventing me from building on my powerpc using newer compilers:
> > [fmtlib/fmt@8b76e97](https://github.com/fmtlib/fmt/commit/8b76e970621cdd2b278fb3e751a368ee417e7627)
> 
> I will use the ugly FMT macros then.
> 
> I didn't know there were systems that didn't support exceptions. That is very odd. Do you know why?

Also there are already 6 try/catch blocks in dvlnet. Aren't they already an issue?

Source/utils/log.hpp
AJenbo: I think maybe the easy route is to make `UiOkDialog()` always log, not just when there is no display.

AJenbo:
> Also there are already 6 try/catch blocks in dvlnet. Aren't they already an issue?

Yes they are, some ports have had to patch them out. Haven't gotten around to dealing with them yet.

AJenbo:
> > On a side note could you check that we can build with exceptions disabled. Requiring exceptions is preventing me from building on my powerpc using newer compilers:
> > [fmtlib/fmt@8b76e97](https://github.com/fmtlib/fmt/commit/8b76e970621cdd2b278fb3e751a368ee417e7627)
> 
> I will use the ugly FMT macros then.
> 
> I didn't know there were systems that didn't support exceptions. That is very odd. Do you know why?

afaik they are both complicated to implement and can hurt performance since more things have to be tracked in case of an exception.

Jmgr:
> > > On a side note could you check that we can build with exceptions disabled. Requiring exceptions is preventing me from building on my powerpc using newer compilers:
> > > [fmtlib/fmt@8b76e97](https://github.com/fmtlib/fmt/commit/8b76e970621cdd2b278fb3e751a368ee417e7627)
> > 
> > 
> > I will use the ugly FMT macros then.
> > I didn't know there were systems that didn't support exceptions. That is very odd. Do you know why?
> 
> afaik they are both complicated to implement and can hurt performance since more things have to be tracked in case of an exception.

Yes, throwing exceptions is costly, but they are really designed for exceptional things, like a missing resource file, or a null pointer. Things that you rarely encounter, so the impact in performance is very occasional. Well, at least with C++17 we can return multiple values very easily, so that and std::optional could be enough.

Chance4us:
Hi @all. How can I build with exceptions disabled?

AJenbo:
> Hi @ ALL. How can I build with exceptions disabled?

`-fno-exceptions`

If you are wanting to do this for performance you probably won't see any real difference as things are.

AJenbo:
On the problems with exceptions: https://www.youtube.com/watch?v=ARYP83yNAWk

PR #1736, opened 2021-04-27
Check for diablo.ini in PWD, if found use that and save files into ./ for portable mode.

Trihedraf:


Source/utils/paths.cpp
```diff
@@ -75,7 +76,10 @@ const std::string &PrefPath()
 {
 	if (!prefPath)
 		prefPath = FromSDL(SDL_GetPrefPath("diasurgical", "devilution"));
-	return *prefPath;
+	if (FileExists("diablo.ini")) {
+		prefPath = "save/";
```
AJenbo, line 80: Why is it adding `save` to the path, wouldn't existing save games normally be in the same folder?

Trihedraf: with the zero tier stuff it creates so much crap with the saves so it looks a lot nicer to me to save it in a save folder.

julealgon (changes requested):
Totally against this idea. It all sounds like a big hack to me. IMHO, we are already doing the correct thing to do and shouldn't make any changes to it.

Trihedraf:
We already check for the mpqs in multiple places. Why can’t we do that for the ini and saves so that it can be setup to run out of a single contained folder without having to use to shortcuts or scripts.

AJenbo:
I knew looking for multiple files in multiple folder would be a slippery slope :D

The reason not to do things like that is that it's actually caused an issue with how the game loads assets directly from disk (as opposed to an MPQ). It makes things more complicated and complicated things have bugs. So that has to be weighted with the benefit to the user. But quite a lot of Windows users are surprised when they dump a new exe in their existing diablo folder and it doesn't load files from the same location, so there are some reasonable benefits to attempting to figure out the users intent.

Source/utils/paths.cpp
Trihedraf: But if we need to set it to be in the ./ so that it matches with Vanilla, I can switch it.

AJenbo: To me adding save defeated the purpose of this PR. But we can and should move the ZeroTier files to a sub-folder.

Trihedraf:
```suggestion
		prefPath = "./";

```

galaxyhaxz:
Great work @Trihedraf! I've been saying this should have been a thing for a long time now. It's especially useful as now you can have DevilutionX on a flash drive and play it at school or another persons computer where you may not have access/permission to get into the hidden user directory.

AJenbo:
You could already do that just with a shortcut instead of an ini, this just makes it detect if it's in an original installation with write access.

galaxyhaxz:
> You could already do that just with a shortcut instead of an ini, this just makes it detect if it's in an original installation with write access.

More streamlined and easier now though ;)

Source/utils/paths.cpp
```diff
@@ -75,7 +76,10 @@ const std::string &PrefPath()
 {
 	if (!prefPath)
 		prefPath = FromSDL(SDL_GetPrefPath("diasurgical", "devilution"));
-	return *prefPath;
+	if (FileExists("diablo.ini")) {
+		prefPath = "./";
+	}
```
AJenbo, line 81: Looks like this path doesn't care if the folder is writable or not. Maybe move the check earlier so that it can influence both paths and only has to run once. But also make sure that if the path is given specifically by the user then we do not run this check.

Trihedraf: I rewrote the save one so it checks for write access and if the user set the path.

But I couldn't figure out where exactly I would run it earlier, because they both return to static and const. Which those can't be written twice can they?

Source/utils/paths.cpp
```diff
@@ -73,8 +74,18 @@ const std::string &BasePath()
 
 const std::string &PrefPath()
 {
-	if (!prefPath)
+	if (!prefPath){
```
AJenbo, line 77: Run `clang-format`

malvarenga123:
It's as you say @AJenbo, it's quite annoying to have those files in different places. This makes it much easier to run have multiple versions of the exe in the same folder.

Source/storm/storm.cpp
```diff
@@ -36,8 +37,24 @@ std::string *SBasePath = nullptr;
 
 radon::File &getIni()
 {
-	static radon::File ini(paths::ConfigPath() + "diablo.ini");
-	return ini;
+	if (FileExists("diablo.ini")) {
+		const std::string bdwtpath = paths::BasePath() + "BaseDirWriteTest.foo";
```
glebm, line 41: This should be an `IsWriteableDir` method in file_util

glebm (changes requested):

Source/storm/storm.cpp
Trihedraf: Done.

Source/utils/paths.cpp
Trihedraf: Got it. Done.

Source/utils/file_util.cpp
```diff
@@ -39,6 +39,20 @@ bool FileExists(const char *path)
...
+	struct stat stats;
+	if (stat(path, &stats) == 0) {
+		if (stats.st_mode & 2) {
+			return true;
+		} else {
+			return false;
+		}
+	} else {
+		return false;
+	}
```
AJenbo, line 53: Some useless else and nested ifs:
```suggestion
	if (stat(path, &stats) == 0 && stats.st_mode & 2) {
		return true;
	}
	
	return false;
```

Source/utils/paths.cpp
```diff
@@ -73,15 +74,25 @@ const std::string &BasePath()
...
 const std::string &PrefPath()
 {
-	if (!prefPath)
+	if (!prefPath) {
 		prefPath = FromSDL(SDL_GetPrefPath("diasurgical", "devilution"));
+		if (IsWriteable("diablo.ini")) {
+			if (IsWriteable("./")) {
+				prefPath = "./";
+			}
+		}
```
AJenbo, line 83:
```suggestion
		if (IsWriteable("diablo.ini") && IsWriteable("./")) {
			prefPath = "./";
		}

```

AJenbo:
Nice, this is now clean enough that it's not an issue to include, compared to the benifit.

Source/utils/file_util.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include <algorithm>
 #include <string>
+#include <sys/stat.h>
```
glebm, line 6: I'm surprised this works on all platforms that we support!

AJenbo: Positively surprised I assume :D

Trihedraf:
@AJenbo is there something that we are waiting for on this one?

AJenbo:
Not much, I just need time to test it :)

There has been a lot to see to as of late, so haven't had as much time to review and merge PRs as I would have liked, but will probably get to this one this week.

Source/utils/file_util.cpp
```diff
@@ -80,6 +80,12 @@ bool FileExists(const char *path)
 #endif
 }
 
+bool IsWriteable(const char *path)
+{
+	struct stat stats;
+	return stat(path, &stats) == 0 && (stats.st_mode & 2) != 0;
```
glebm, line 86: Does this work with non-ASCII paths on Windows?

AJenbo:
@Trihedraf could you adjust this to match the latest code for checking readable paths?
https://github.com/diasurgical/devilutionX/pull/1929

Source/utils/file_util.cpp
glebm: Please move the `<sys/stat.h>` include back into the `#if _POSIX_C_SOURCE >= 200112L` block

Source/utils/file_util.cpp
```diff
@@ -73,6 +73,27 @@ bool FileExists(const char *path)
...
+bool IsWriteable(const char *path)
+#if defined(_WIN64) || defined(_WIN32)
+{
+	const auto pathUtf16 = ToWideChar(path);
+	if (pathUtf16 == nullptr) {
+		LogError("UTF-8 -> UTF-16 conversion error code {}", ::GetLastError());
+		return false;
+	}
+	uint32_t attr = ::GetFileAttributesW(&pathUtf16[0]);
+	if (attr != INVALID_FILE_ATTRIBUTES && attr & FILE_ATTRIBUTE_READONLY) {
```
glebm, line 85:
```suggestion
	if (attr != INVALID_FILE_ATTRIBUTES && (attr & FILE_ATTRIBUTE_READONLY) != 0) {
```

Source/utils/file_util.cpp
```diff
@@ -73,6 +73,27 @@ bool FileExists(const char *path)
...
+		return false;
+	}
+	return true;
+}
+#else
+{
+	struct stat stats;
+	return stat(path, &stats) == 0 && (stats.st_mode & 2) != 0;
```
glebm, line 93: 1. What is this `2`? Should it be something like this instead? The values of `st_mode` are platform-specific.

  ```suggestion
	  return stat(path, &stats) == 0 && (stats.st_mode & S_IWUSR) != 0;
  ```

2. On platforms that support `access`, we should use it instead:

   ```c++
   #elif _POSIX_C_SOURCE >= 200112L || defined(_BSD_SOURCE) || defined(__APPLE__)
	   return ::access(path, W_OK) == 0;
   #else
   ```

Source/utils/file_util.cpp
```diff
@@ -73,6 +73,27 @@ bool FileExists(const char *path)
...
+	uint32_t attr = ::GetFileAttributesW(&pathUtf16[0]);
+	if (attr != INVALID_FILE_ATTRIBUTES && (attr & FILE_ATTRIBUTE_READONLY) != 0) {
+		return false;
+	}
+	return true;
+#elif _POSIX_C_SOURCE >= 200112L || defined(_BSD_SOURCE) || defined(__APPLE__)
+	return ::access(path, W_OK) == 0;
+#else
+	struct stat stats;
+	return stat(path, &stats) == 0 && (stats.st_mode & W_OK) != 0;
```
glebm, line 93: From what I can tell, `W_OK` is an `::access` flag, not a `stat` flag?

Anyway, I think this `#else` clause is no longer necessary and can be removed. `access` is available whenever `stat` is available, they're both POSIX https://pubs.opengroup.org/onlinepubs/9699919799/functions/access.html

Trihedraf: Removed

Source/utils/file_util.h
```diff
@@ -10,6 +10,7 @@
 namespace devilution {
 
 bool FileExists(const char *path);
+bool IsWriteable(const char *path);
```
glebm, line 13: Perhaps we should rename this to `FileExistsAndIsWriteable` to better reflect what this checks for?

Source/utils/file_util.cpp
```diff
@@ -73,6 +73,24 @@ bool FileExists(const char *path)
...
+	const auto pathUtf16 = ToWideChar(path);
+	if (pathUtf16 == nullptr) {
+		LogError("UTF-8 -> UTF-16 conversion error code {}", ::GetLastError());
+		return false;
+	}
+	uint32_t attr = ::GetFileAttributesW(&pathUtf16[0]);
+	if (attr != INVALID_FILE_ATTRIBUTES && (attr & FILE_ATTRIBUTE_READONLY) != 0) {
+		return false;
+	}
+	return true;
```
glebm, line 88: Alternative, `<io.h>` on Windows has a `_waccess` function:

```suggestion
	return _waccess(&pathUtf16[0], 2) != -1);
```

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/access-waccess?view=msvc-160

Not sure which is better

qqkookie:
How about checking $PWD\SAVE\diablo.ini first and then checking $PWD\diablo.ini next? I dislike multiple save files scattered in game directory itself. Please user separate subdirectory for diablo.ini and save files and keep the game install directory clean and tidy.

Trihedraf:
I originally had it put all the saves in a ./save folder, but @AJenbo said to take it out.

AJenbo:
Hopefully this won't cause to many issues for players that extracted DevilutionX to there original diablo folder but have there save games in the current pref path.

Source/utils/paths.cpp
```diff
@@ -69,15 +70,23 @@ const std::string &BasePath()
 
 const std::string &PrefPath()
 {
-	if (!prefPath)
+	if (!prefPath) {
 		prefPath = FromSDL(SDL_GetPrefPath("diasurgical", "devilution"));
+		if (FileExistsAndIsWriteable("diablo.ini")) {
+			prefPath = "./";
```
AJenbo, line 76: Apparently this doesn't work on GCC 5.5

Trihedraf: Sent PR to fix #1984

PR #1737, opened 2021-04-28
🎉 render.cpp: Do not render out-of-bounds

glebm:
This is part of the work to allow us to eliminate buffer padding.

As this is a hotspot, we have 4 separate functions for each non-square primitive, resulting in quite a bit of code:

1. Unclipped ("Full")
2. Vertical-only clip
3. Vertical + Left clip
4. Vertical + Right clip

FPS:
* 640x480: 1420 -> 1530
* 1280x960: 365 -> 380

### DIFF
--- a/Source/render.cpp
+++ b/Source/render.cpp
@@ -5,28 +5,136 @@
  */
 #include "render.h"
 
+#include <algorithm>
 #include <climits>
+#include <cstdint>
 
 #include "lighting.h"
 #include "options.h"
 
 namespace devilution {
 
-#define NO_OVERDRAW
-
 namespace {
 
-enum {
-	RT_SQUARE,
-	RT_TRANSPARENT,
-	RT_LTRIANGLE,
-	RT_RTRIANGLE,
-	RT_LTRAPEZOID,
-	RT_RTRAPEZOID
+/**
+ * Tile type.
+ *
+ * The tile type determines data encoding and the shape.
+ *
+ * Each tile type has its own encoding but they all encode data in the order
+ * of bottom-to-top (bottom row first).
+ */
+enum class TileType {
+	/**
+	 * 🮆 A 32x32 square. Stored as an array of pixels.
+	 */
+	Square,
+
+	/**
+	 * 🮆 A 32x32 square with transparency. RLE encoded.
+	 *
+	 * Each run starts with an int8_t value.
+	 * If positive, it is followed by this many pixels.
+	 * If negative, it indicates `-value` fully transparent pixels, which are omitted.
+	 *
+	 * Runs do not cross row boundaries.
+	 */
+	TransparentSquare,
+
+	/**
+	 *🭮 Left-pointing 32x31 triangle. Encoded as 31 varying-width rows with 2 padding bytes before every even row.
+	 *
+	 * The smallest rows (bottom and top) are 2px wide, the largest row is 16px wide (middle row).
+	 *
+	 * Encoding:
+	 * for i in [0, 30]:
+	 * - 2 unused bytes if i is even
+	 * - row (only the pixels within the triangle)
+	 */
+	LeftTriangle,
+
+	/**
+	 * 🭬Right-pointing 32x31 triangle.  Encoded as 31 varying-width rows with 2 padding bytes after every even row.
+	 *
+	 * The smallest rows (bottom and top) are 2px wide, the largest row is 16px wide (middle row).
+	 *
+	 * Encoding:
+	 * for i in [0, 30]:
+	 * - row (only the pixels within the triangle)
+	 * - 2 unused bytes if i is even
+	 */
+	RightTriangle,
+
+	/**
+	 * 🭓 Left-pointing 32x32 trapezoid: a 32x16 rectangle and the 16x16 bottom part of `LeftTriangle`.
+	 *
+	 * Begins with triangle part, which uses the `LeftTriangle` encoding,
+	 * and is followed by a flat array of pixels for the top rectangle part.
+	 */
+	LeftTrapezoid,
+
+	/**
+	 * 🭞 Right-pointing 32x32 trapezoid: 32x16 rectangle and the 16x16 bottom part of `RightTriangle`.
+	 *
+	 * Begins with the triangle part, which uses the `RightTriangle` encoding,
+	 * and is followed by a flat array of pixels for the top rectangle part.
+	 */
+	RightTrapezoid,
 };
 
+/** Width of a tile rendering primitive. */
+constexpr std::int_fast16_t Width = TILE_WIDTH / 2;
+
+/** Height of a tile rendering primitive (except triangles). */
+constexpr std::int_fast16_t Height = TILE_HEIGHT;
+
+/** Height of the lower triangle of a triangular or a trapezoid tile. */
+constexpr std::int_fast16_t LowerHeight = TILE_HEIGHT / 2;
+
+/** Height of the upper triangle of a triangular tile. */
+constexpr std::int_fast16_t TriangleUpperHeight = TILE_HEIGHT / 2 - 1;
+
+/** Height of the upper rectangle of a trapezoid tile. */
+constexpr std::int_fast16_t TrapezoidUpperHeight = TILE_HEIGHT / 2;
+
+constexpr std::int_fast16_t TriangleHeight = LowerHeight + TriangleUpperHeight;
+
+/** For triangles, for each pixel drawn vertically, this many pixels are drawn horizontally. */
+constexpr std::int_fast16_t XStep = 2;
+
+std::int_fast16_t GetTileHeight(TileType tile)
+{
+	if (tile == TileType::LeftTriangle || tile == TileType::RightTriangle)
+		return TriangleHeight;
+	return Height;
+}
+
+// Debugging variables
+// #define DEBUG_RENDER_COLOR
+// #define DEBUG_RENDER_OFFSET_X 5
+// #define DEBUG_RENDER_OFFSET_Y 5
+
+#ifdef DEBUG_RENDER_COLOR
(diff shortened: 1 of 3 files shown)


AJenbo:
Is that the top row of the town portal load screen still showing in your screen shot?

AJenbo:
Also why is the wall missing where you cursor is pointing?

glebm:
The screenshot is with padding removed; this does not happen with padding.

> Is that the top row of the town portal load screen still showing in your screen shot?
> Also why is the wall missing where you cursor is pointing?

I'm guessing this could be related to the yet unported Cel functions but not sure

AJenbo:
> I'm guessing this could be related to the yet unported Cel/Cl2 functions but not sure

Looks like the clipping is off by 1, the world looks like it might only be rendered to 800x479, maybe even 798x478 but a bit hard to tell.

glebm:
@AJenbo Do you mean the white line on top? That line is there for me without this PR and even in menu, I don't think it's related

glebm:
Fixed the black tile bug

glebm:
Rebased

Source/render.h
```diff
@@ -23,10 +23,10 @@ namespace devilution {
  * @param x Target buffer coordinate
  * @param y Target buffer coordinate
  */
-void RenderTile(CelOutputBuffer out, int x, int y);
+void RenderTile(const CelOutputBuffer &out, int x, int y);
 
 /**
- * @brief Render a black tile
+ * @brief Render a black 64x31 tile ◆
```
AJenbo, line 29: Emoji everything!

glebm:
Converted to draft as apparently it's possible to make it glitch -- waiting for save to reproduce

StephenCWills:
I noticed some glitches at the right and bottom edges of the screen on 3DS when walking through town at 800x480 resolution. I can provide a save, but it shouldn't be necessary as the character I used was a newly created character.

glebm:
Ah, the glitches are there only when using SDL1. Reproduced.

glebm:
Glitches now fixed

StephenCWills:
I'm still getting the same glitches on 3DS with 2016177. Let me see if I can capture a video or something.

glebm:
@StephenCWills Are you sure? Pretty sure it should be fixed now :eyes: 

By the way you can uncomment `// #define DEBUG_RENDER_COLOR` and get the debug view

StephenCWills:
I couldn't figure out video capture, but this screenshot shows it pretty clearly.

image

I had to tweak my build change backlight settings so I could see the option to take the screenshot. Therefore, I'm quite certain this is 2016177. I'll turn `DEBUG_RENDER_COLOR` on and get another screenshot for you shortly.

glebm:
Can't reproduce on PC at all :thinking: 

image
image

StephenCWills:
Yeah, I also tried to reproduce it on Linux to make it easier to capture screenshots or video, but alas...

StephenCWills:
Here it is with the debugging on.

image

glebm:
@StephenCWills Can you please try again?

This is the salient change:

```diff
 	const auto skipRestOfTheLine = [&src](std::int_fast16_t remainingWidth) {
 		while (remainingWidth > 0) {
-			const auto v = static_cast<std::int_fast8_t>(*src++);
+			const auto v = static_cast<std::int8_t>(*src++);
```

StephenCWills:
Looks like that fixed it.

image

Source/render.cpp
```diff
@@ -422,306 +539,834 @@ inline void DoRenderLine(BYTE *dst, BYTE *src, size_t n, BYTE *tbl, DWORD mask)
...
+		result.upperTop = clip.top;
+		result.lowerBottom = clip.bottom;
+		result.lowerTop = result.upperBottom = 0;
+	}
+	return result;
+}
+
+std::size_t CalculateTriangleSourceSkipLowerBottom(std::int_fast16_t numLines)
+{
+	return XStep * numLines * (numLines + 1) / 2 + 2 * ((numLines + 1) / 2);
```
AJenbo, line 754: Can be simplified as:
```suggestion
	return XStep * numLines * (numLines + 1) / (numLines + 3);
```

Source/render.cpp
```diff
@@ -422,306 +539,834 @@ inline void DoRenderLine(BYTE *dst, BYTE *src, size_t n, BYTE *tbl, DWORD mask)
...
+}
+
+std::size_t CalculateTriangleSourceSkipUpperBottom(std::int_fast16_t numLines)
+{
+	return 2 * TriangleUpperHeight * numLines - numLines * (numLines - 1) + 2 * ((numLines + 1) / 2);
```
AJenbo, line 759: Don't know if this is supposed to do some rounding magic but could also be written as
```suggestion
	return 2 * TriangleUpperHeight * numLines - numLines * numLines + 2 * numLines + 1;
```

Source/render.cpp
```diff
@@ -422,306 +539,834 @@ inline void DoRenderLine(BYTE *dst, BYTE *src, size_t n, BYTE *tbl, DWORD mask)
...
+}
+
+void RenderLeftTriangleFull(BYTE *dst, int dstPitch, BYTE *src, const DWORD *mask, const BYTE *tbl)
```
AJenbo, line 762: Any reason to keep using the WinAPI style types in new code?

Source/render.cpp
```diff
@@ -422,306 +539,834 @@ inline void DoRenderLine(BYTE *dst, BYTE *src, size_t n, BYTE *tbl, DWORD mask)
...
+		src += width + 2 * (i % 2);
+	}
+	src += CalculateTriangleSourceSkipUpperBottom(clipY.upperBottom);
+	const auto upperMax = TriangleUpperHeight - clipY.upperTop;
+	for (auto i = 1 + clipY.upperBottom; i <= upperMax; ++i, dst -= dstPitch, --mask) {
+		const auto width = Width - XStep * i;
+		if (width > clipLeft)
+			RenderLine(dst, src + clipLeft, width - clipLeft, tbl, (*mask) << clipLeft);
+		else
+			break;
```
AJenbo, line 928: This feels missleading
```suggestion
		if (width <= clipLeft)
			break;

		RenderLine(dst, src + clipLeft, width - clipLeft, tbl, (*mask) << clipLeft);
```

Source/render.cpp
```diff
@@ -422,306 +539,834 @@ inline void DoRenderLine(BYTE *dst, BYTE *src, size_t n, BYTE *tbl, DWORD mask)
...
+	const auto clipY = CalculateDiamondClipY(clip);
+	const auto clipRight = clip.right;
+	src += CalculateTriangleSourceSkipLowerBottom(clipY.lowerBottom);
+	const auto lowerMax = LowerHeight - clipY.lowerTop;
+	for (auto i = 1 + clipY.lowerBottom; i <= lowerMax; ++i, dst -= dstPitch, --mask) {
+		const auto width = XStep * i;
+		const auto skip = Width - width < clipRight ? clipRight - (Width - width) : 0;
+		if (width > skip)
+			RenderLine(dst, src, width - skip, tbl, *mask);
+		src += width + 2 * (i % 2);
```
AJenbo, line 944: Preferably the happy path should be readable with same level if-statements collapsed.

```suggestion
		if (width <= skip)
			break;

		RenderLine(dst, src, width - skip, tbl, *mask);
		src += width + 2 * (i % 2);
```

Source/render.cpp
```diff
@@ -422,306 +539,834 @@ inline void DoRenderLine(BYTE *dst, BYTE *src, size_t n, BYTE *tbl, DWORD mask)
...
+	const BYTE *tbl = &pLightTbl[256 * light_table_index];
+	const auto *pFrameTable = reinterpret_cast<const std::uint32_t *>(pDungeonCels.get());
+	BYTE *src = &pDungeonCels[SDL_SwapLE32(pFrameTable[level_cel_block & 0xFFF])];
+	BYTE *dst = out.at(static_cast<int>(x + clip.left), static_cast<int>(y - clip.bottom));
+	const auto dstPitch = out.pitch();
+	mask -= clip.bottom;
+
+	switch (tile) {
+	case TileType::Square:
+		RenderSquare(dst, dstPitch, src, mask, tbl, clip);
```
AJenbo, line 1324: Didn't you just rename my "square" map function to diamond :)

AJenbo (commented):
Looks good. I can't help but think that this could be drastically simplified by prerendering all tiles in to simple rectangles. I doubt it would even have a negative performance effect if RLE was also applied, and the reduction of special conditions would probably help performance. However doing so would require touching multiple components (render, scrollrt, level load).

galaxyhaxz:
>  all tiles in to simple rectangles. I doubt it would even have a negative performance effect if RLE was also applied

In my own tests, converting the micro tiles to regular RLE cel files actually gives a performance boost. So, their micro-optimization not only didn't work, it actually regressed xD

Source/render.cpp
glebm: This is rounding magic, it return 2 2 4 4 6 6 8 8 etc

Source/render.cpp
glebm: Same rounding magic as above

Source/render.cpp
glebm: Your "square" is a diamond, this is an actual square. See the TileType enum documentation

Source/render.cpp
glebm: Done

Source/render.cpp
glebm: Done

Source/render.cpp
glebm: Done

Source/render.cpp
AJenbo: ok :)

PR #1739, opened 2021-04-28
Operators for Point: +, -, +=, -=

vlolteanu:


Source/engine.h
```diff
@@ -60,6 +60,26 @@ enum direction : uint8_t {
...
+	}
+
+	Point operator -(const Point &other)
+	{
+		return {x - other.x, y - other.y};
+	}
+
+	void operator +=(const Point &other)
+	{
+		*this = *this + other;
```
glebm, line 76: nit: Usually it's implemented the other way around (`+` in terms of `+=` and copy)

Source/engine.h
```diff
@@ -60,6 +60,26 @@ enum direction : uint8_t {
...
+	{
+		return {x + other.x, y + other.y};
+	}
+
+	Point operator -(const Point &other)
+	{
+		return {x - other.x, y - other.y};
+	}
+
+	void operator +=(const Point &other)
```
glebm, line 74: Operators `+=` and `-=` should return `Point &` (`return *this`)

Source/engine.h
```diff
@@ -60,6 +60,26 @@ enum direction : uint8_t {
 struct Point {
 	int x;
 	int y;
+
+	Point operator +(const Point &other)
```
glebm, line 64: From http://en.cppreference.com/w/cpp/language/operators:

> Binary operators are typically implemented as non-members to maintain symmetry

Source/engine.h
vlolteanu: Yikes! Thanks for catching this.

Source/engine.h
vlolteanu: It does make it look nicer, even though this struct is trivial.

Source/engine.h
vlolteanu: I see no point doing it the other way around in this case. (Pun intended.)

glebm: There is no advantage to either approach in this case, so please use the common approach for consistency. There is no need for this one to be special.

Source/engine.h
```diff
@@ -60,6 +60,28 @@ enum direction : uint8_t {
 struct Point {
 	int x;
 	int y;
+
+	friend Point operator +(const Point &a, const Point &b)
+	{
+		return {a.x + b.x, a.y + b.y};
+	}
```
glebm, line 67:
```suggestion
	friend Point operator +(Point a, const Point &b)
	{
		a += b;
		return a;
	}
```

etc

glebm (approved):

PR #1744, opened 2021-04-28
Replace LOAD_LE32 & LOAD_BE32 with signedness- and endian-agnostic equivalents

vlolteanu:


Source/engine.h
```diff
@@ -107,14 +107,34 @@ inline BYTE *CelGetFrameStart(BYTE *pCelBuff, int nCel)
...
-#define LOAD_BE32(b) (((DWORD)(b)[0] << 24) | ((DWORD)(b)[1] << 16) | ((DWORD)(b)[2] << 8) | (DWORD)(b)[3])
+constexpr uint32_t LoadLE32(const void *v)
+{
+	const uint8_t *b = reinterpret_cast<const uint8_t *>(v);
+
+#if BYTE_ORDER == LITTLE_ENDIAN
+	return ((uint32_t)(b)[3] << 24) | ((uint32_t)(b)[2] << 16) | ((uint32_t)(b)[1] << 8) | (uint32_t)(b)[0];
+#else /* BIG ENDIAN */
+	return ((uint32_t)(b)[0] << 24) | ((uint32_t)(b)[1] << 16) | ((uint32_t)(b)[2] << 8) | (uint32_t)(b)[3];
+#endif
```
AJenbo, line 118: How about using SDL_SwapBE32/SDL_SwapLE32 here?

vlolteanu: You could do that, but the result won't be known at compile time. GAME_ID in diablo.h depends on these compile-time swaps.

vlolteanu:
Ah, clang doesn't like reinterpret_cast in constexpr functions. GCC is fine with that.

I'll have to do it some other way.

AJenbo:
Looks like MSVC is having the same issue.

AJenbo (approved):

PR #1748, opened 2021-04-28
Fix for rare level generation bug on Crypt 1 that causes missing story book braziers

### LINKED ISSUE #1515: Frequent crashes in the crypt
[crypt_crash.zip](https://github.com/diasurgical/devilutionX/files/6305286/crypt_crash.zip)

**Important information**
Version: DevilutionX 1.2.1 AppImage from the release page
OS: Ubuntu 20.10

**Describe**
The game randomly crashes while in the crypt with the following output `“./devilutionx-linux-x86_64.appi…” terminated by signal SIGSEGV (Address boundary error)`. Sometimes immediately upon loading, sometimes while interacting with the dungeon or just walking.

**To Reproduce**
I could not distill a specific set of steps to reproduce the issue. However, just trying to play the game from the attached save should cause a crash within 30 seconds.

**Expected behavior**
Crypt is playable without issues.

thebigMuh:
Fixes #1515
- Crypt candles (braziers) around the story book were occasionally missing
- Marked OFILE_CANDLE2 to load for all levels 1 through 12, so it is always loaded for crypt as well
- Made search for object graphics a bit more secure, as in the bugged case it would run off past the end of an array


qndel (approved):

AJenbo:
Considering that we have at least 3 bug reports for it I'm not sure exactly how rare this is.

PR #1750, opened 2021-04-29
Use std::string_view for towners' names

vlolteanu:


vlolteanu:
Unfortunately, I can't access CircleCi without giving it access to some private repos (or creating a fake account). Why is the build failing on amigaos?

vlolteanu:
Let me guess... Amiga's GCC doesn't have string_view? Maybe it's in std::experimental.

AJenbo:
Sorry didn't see you request, will try to post the error next time if any.

Source/towners.h
```diff
@@ -1,76 +1,77 @@
-/**
```
glebm, line 1: Please configure your editor to respect line endings in `.editorconfig`

Source/control.cpp
```diff
@@ -1290,7 +1290,7 @@ void DrawInfoBox(const CelOutputBuffer &out)
 					PrintMonstHistory(monster[pcursmonst].MType->mtype);
 				}
 			} else if (pcursitem == -1) {
-				strcpy(infostr, towners[pcursmonst]._tName);
+				strcpy(infostr, towners[pcursmonst]._tName.data());
```
glebm, line 1293: string views are not guaranteed to be zero-terminated (even though this one is), so you must provide the length explicitly via `strncpy`.

Source/towners.h
```diff
@@ -1,76 +1,77 @@
...
-} // namespace devilution
+/**
+ * @file towners.h
+ *
+ * Interface of functionality for loading and spawning towners.
+ */
+#pragma once
+
+#include <cstdint>
+#include <string_view>
```
glebm, line 9: This won't work on older compilers. See `utils/stdcompat/optional.hpp` for an example of how to make this work.

glebm (changes requested):

Source/towners.h
vlolteanu: Github messed up this file when I resolved a conflict in the browser.

Source/control.cpp
```diff
@@ -1290,7 +1290,8 @@ void DrawInfoBox(const CelOutputBuffer &out)
 					PrintMonstHistory(monster[pcursmonst].MType->mtype);
 				}
 			} else if (pcursitem == -1) {
-				strcpy(infostr, towners[pcursmonst]._tName);
+				std::string_view townerName = towners[pcursmonst]._tName;
+				strncpy(infostr, townerName.data(), townerName.length());
```
glebm, line 1294: `strncpy` does not append `\0`, this must be done manually:

```suggestion
				strncpy(infostr, townerName.data(), townerName.length());
				infostr[townerName.length()] = '\0';
```

Source/towners.cpp
```diff
@@ -414,7 +414,7 @@ void InitFarmer()
 	}
 	towners[numtowners]._tNFrames = 15;
 	NewTownerAnim(numtowners, towners[numtowners]._tNAnim[DIR_S], towners[numtowners]._tNFrames, 3);
-	strcpy(towners[numtowners]._tName, _("Lester the farmer"));
+	towners[numtowners]._tName = {_("Lester the farmer")};
```
glebm, line 417: `string_view` is implicitly convertible from `const char *`, so the curly braces aren't necessary (throughout)

Source/utils/stdcompat/string_view.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#ifdef __has_include
+#if defined(__cplusplus) && __cplusplus >= 201703L && __has_include(<string_view>)
```
glebm, line 4: According to https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros, the correct value is 201606L

```suggestion
#if defined(__cplusplus) && __cplusplus >= 201606L && __has_include(<string_view>)
```

vlolteanu: It's 201703L in the stl header.  Maybe a comment would be worthwhile here. We risk breaking a build by using the lower value.

glebm: image

glebm (approved):

glebm (dismissed):

glebm:
Looks like this messes with the implementation from `fmt`

Options:

1. `using string_view = std::experimental::string_view` or `std::string_view` in the `devilution` namespace.
2. `using string_view = fmt::string_view`.

vlolteanu:
How about this instead?

namespace std {
using string_view = experimental::string_view;
}

glebm:
@vlolteanu User code is not allowed to define anything in the `std` namespace

glebm (approved):

vlolteanu:
> @vlolteanu User code is not allowed to define anything in the `std` namespace

That's a bit surprising. You can specialize anything in std and essentially taint std. Or is that disallowed as well?

glebm:
> That's a bit surprising. You can specialize anything in std and essentially taint std. Or is that disallowed as well?

Disallowed and usually a compiler error too

glebm:
Error from Amiga:

>      std::string_view townerName = towners[pcursmonst]._tName;

Source/control.cpp
```diff
@@ -1290,7 +1290,9 @@ void DrawInfoBox(const CelOutputBuffer &out)
 					PrintMonstHistory(monster[pcursmonst].MType->mtype);
 				}
 			} else if (pcursitem == -1) {
-				strcpy(infostr, towners[pcursmonst]._tName);
+				std::string_view townerName = towners[pcursmonst]._tName;
```
glebm, line 1293:
```suggestion
				string_view townerName = towners[pcursmonst]._tName;
```

PR #1758, opened 2021-04-29
Fix missing sounds

AJenbo:
When playing the same sound twice with in a short duration then the
sound it self, the secound request would go unfulfilled. The solution is
to simply rewind the stream pointer before requesting a playback.


mmaret:
Is it not a "feature" ? Like you don't want hear it twice when too close ?

AJenbo:
No, the game has a limit on repeating sounds twice within 80ms, this is different. It causes bow hots faster than 300ms to only play a sound half the time. It's a regression caused by the move from SDL_mixer to SDL_audiolib.

glebm (approved):

PR #1762, opened 2021-04-29
Fix memleak in CapturePix

vlolteanu:
If there's some error when writing the screenshot to the file, the buffer does not get freed.

(As a fun fact, this is literally the first hit that my IDE yielded when searching for DiabloAllocPtr.)

I was conservative with this commit, and used the DiabloAllocPtr and mem_free_dbg macros. Going forward, should we still use them in cases where we can make memory management automatic? (The network code, for example, does not use them.)

### DIFF
--- a/Source/capture.cpp
+++ b/Source/capture.cpp
@@ -116,7 +116,7 @@ static BYTE *CaptureEnc(BYTE *src, BYTE *dst, int width)
 static bool CapturePix(const CelOutputBuffer &buf, std::ofstream *out)
 {
 	int width = buf.w();
-	std::unique_ptr<BYTE, DiabloDeleter> pBuffer {(BYTE *)DiabloAllocPtr(2 * width)};
+	DiabloUniqPtr<BYTE> pBuffer = DiabloMakeUnique<BYTE>(2 * width);
 	BYTE *pixels = buf.begin();
 	for (int height = buf.h(); height > 0; height--) {
 		const BYTE *pBufferEnd = CaptureEnc(pixels, pBuffer.get(), width);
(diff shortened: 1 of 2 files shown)


glebm:
Do we actually want a `DiabloUniqPtr` type? Perhaps we should just use `std::unique_ptr`, and use proper tools to track allocations when needed.

AJenbo:
Agreed, throw it away

vlolteanu:
Done.

vlolteanu:
DiabloMakeUnique might still be useful for e.g. TMsg, where memory is allocated beyond a struct's boundary, expecting some more data to follow.

Source/engine.h
```diff
@@ -134,6 +135,18 @@ struct ActorPosition {
 #define mem_free_dbg(PTR) \
 	std::free(PTR)
 
+template <typename T>
+std::unique_ptr<T> DiabloMakeUnique(size_t size)
```
glebm, line 139: Unused, please remove

Source/capture.cpp
```diff
@@ -116,16 +116,15 @@ static BYTE *CaptureEnc(BYTE *src, BYTE *dst, int width)
 static bool CapturePix(const CelOutputBuffer &buf, std::ofstream *out)
 {
 	int width = buf.w();
-	BYTE *pBuffer = (BYTE *)DiabloAllocPtr(2 * width);
+	std::unique_ptr<BYTE []> pBuffer = std::make_unique<BYTE[]>(2 * width);
```
glebm, line 119:
```suggestion
	auto pBuffer = std::make_unique<BYTE[]>(2 * width);
```

glebm (approved):

AJenbo (approved):

AJenbo:
> I was conservative with this commit

12 commits... I'll just squash this one any way.

PR #1763, opened 2021-04-29
Update SDL_audiolib to fix mono->stereo conversion

glebm:
https://github.com/realnc/SDL_audiolib/commit/0fa26cf4e48ed21e696e60ffc88a586ced9fd6fe


AJenbo:
odd i just got dome angry liker errors when trying the same thing, maybe i should have cleared my cache

glebm:
The cache shouldn't need to be clear manually; perhaps you didn't set `URL_HASH MD5` correctly?

AJenbo:
Most likely i partially deleted the cached content, but not all of it.

glebm:
@AJenbo ah, yeah, looks like it cached the old build

PR #1765, opened 2021-04-29
Add fmt format for translations

Jmgr:
This adds `fmt`'s format support for translated texts.

In its current state this is very experimental, it does compile but some texts are corrupted. The main issue lies with the menus.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3037,7 +3037,7 @@ void GetItemStr(int i)
 			infoclr = UIS_GOLD;
 	} else {
 		int nGold = items[i]._ivalue;
-		sprintf(infostr, ngettext("%i gold piece", "%i gold pieces", nGold), nGold);
+		strcpy(infostr, fmt::format(ngettext("{:d} gold piece", "{:d} gold pieces", nGold), nGold).c_str());
 	}
 }
 
@@ -3464,9 +3464,9 @@ void PrintItemPower(char plidx, ItemStruct *x)
 		break;
 	case IPL_SPLLVLADD:
 		if (x->_iSplLvlAdd > 0)
-			sprintf(tempstr, ngettext("spells are increased %i level", "spells are increased %i levels", x->_iSplLvlAdd), x->_iSplLvlAdd);
+			strcpy(tempstr, fmt::format(ngettext("spells are increased {:d} level", "spells are increased {:d} levels", x->_iSplLvlAdd), x->_iSplLvlAdd).c_str());
 		else if (x->_iSplLvlAdd < 0)
-			sprintf(tempstr, ngettext("spells are decreased %i level", "spells are decreased %i levels", -x->_iSplLvlAdd), -x->_iSplLvlAdd);
+			strcpy(tempstr, fmt::format(ngettext("spells are decreased {:d} level", "spells are decreased {:d} levels", -x->_iSplLvlAdd), -x->_iSplLvlAdd).c_str());
 		else if (x->_iSplLvlAdd == 0)
 			strcpy(tempstr, _("spell levels unchanged (?)"));
 		break;
@@ -3474,7 +3474,7 @@ void PrintItemPower(char plidx, ItemStruct *x)
 		strcpy(tempstr, _("Extra charges"));
 		break;
 	case IPL_SPELL:
-		sprintf(tempstr, ngettext("%i %s charge", "%i %s charges", x->_iMaxCharges), x->_iMaxCharges, _(spelldata[x->_iSpell].sNameText));
+		strcpy(tempstr, fmt::format(ngettext("{:d} {:s} charge", "{:d} {:s} charges", x->_iMaxCharges), x->_iMaxCharges, _(spelldata[x->_iSpell].sNameText)).c_str());
 		break;
 	case IPL_FIREDAM:
 		if (x->_iFMinDam == x->_iFMaxDam)
@@ -3617,7 +3617,7 @@ void PrintItemPower(char plidx, ItemStruct *x)
 		strcpy(tempstr, _("fast block"));
 		break;
 	case IPL_DAMMOD:
-		sprintf(tempstr, ngettext("adds %i point to damage", "adds %i points to damage", x->_iPLDamMod), x->_iPLDamMod);
+		strcpy(tempstr, fmt::format(ngettext("adds {:d} point to damage", "adds {:d} points to damage", x->_iPLDamMod), x->_iPLDamMod).c_str());
 		break;
 	case IPL_RNDARROWVEL:
 		strcpy(tempstr, _("fires random speed arrows"));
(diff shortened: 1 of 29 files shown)


AJenbo:
3ds:```

/root/repo/Source/DiabloUI/selhero.cpp:298:17: error: cannot convert 'std::string' {aka 'std::__cxx11::basic_string<char>'} to 'const char*'
  298 |  ctr_vkbdInput(_("Enter Hero name.."), SelheroGenerateName(selhero_heroInfo.heroclass), selhero_heroInfo.name);
      |                ~^~~~~~~~~~~~~~~~~~~~~
      |                 |
      |                 std::string {aka std::__cxx11::basic_string<char>}
```

File: Source/DiabloUI/dialogs.cpp [hunk 1]
```diff
@@ -168,7 +168,7 @@ void Init(const char *text, const char *caption, bool error, bool renderBehind)
 		vecOkDialog.push_back(new UiText(text, rect2, UIS_CENTER));
 
 		SDL_Rect rect3 = { (Sint16)(PANEL_LEFT + 265), (Sint16)(UI_OFFSET_Y + 265), SML_BUTTON_WIDTH, SML_BUTTON_HEIGHT };
-		vecOkDialog.push_back(new UiButton(&SmlButton, _("OK"), &DialogActionOK, rect3, 0));
+		vecOkDialog.push_back(new UiButton(&SmlButton, _("OK").c_str(), &DialogActionOK, rect3, 0));
```
glebm, line 171: The underscore returns an `std::string` now, which is destroyed after this call, so it points to deleted memory.

Source/utils/language.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace language {
+
+void Initialize();
+std::string Translate(const std::string &key);
+
+} // namespace language
+
+template <typename... Args>
+std::string _(const std::string &key, Args &&... args)
```
glebm, line 23: This signature means a new string is created for every `_` call. We probably don't want it like this, the `_` should only return the translated but not yet interpolated format string (like it did before).

Source/DiabloUI/dialogs.cpp
Jmgr: In that case, yes, `UiButton` should probably store `m_text` as a `std::string`.

There are probably other cases like this, and they are non-trivial to detect, which makes this a bit complicated.

glebm: As I mentioned in the other comment, `_` should only return the key, not the formatted string. In which case the return type can be `const std::string &` or `const char *`.

Source/utils/language.hpp
Jmgr: Since there are arguments that will be passed to `fmt::format` there will be a new `std::string` created in any case. The inline version of `_` that does not take arguments should not though (but I'll need to double-check that this version is actually called).

This form is only to simplify having translations with arguments, replacing
```
fmt::format(_("You have {} gold"), gold)
```
with
```
_("You have {} gold", gold)
```

But both forms will always allocate a new `std::string` because of the format anyway so I don't see the issue.

glebm: Many strings do not accept an argument but having a different return type for the no-arg version would be somewhat confusing.

If a string does not accept an argument, you end up copying it needlessly.

Moreover, even if a string does accept an argument, there is no longer a way to store just the key without creating a new string.

Jmgr: Fair enough, I'll remove that format integration from `_` then.

Source/items.cpp
```diff
@@ -3428,51 +3430,51 @@ void PrintItemPower(char plidx, ItemStruct *x)
...
+		strcpy(tempstr, fmt::format(_("to hit: {:+}%, {:+}% damage"), x->_iPLToHit, x->_iPLDam).c_str());
 		break;
 	case IPL_ACP:
 	case IPL_ACP_CURSE:
-		sprintf(tempstr, _("%+i%% armor"), x->_iPLAC);
+		strcpy(tempstr, fmt::format(_("{:+}% armor"), x->_iPLAC).c_str());
 		break;
 	case IPL_SETAC:
-		sprintf(tempstr, _("armor class: %i"), x->_iAC);
+		strcpy(tempstr, fmt::format(_("armor class: {}"), x->_iAC).c_str());
```
AJenbo, line 3448: One consideration I hadn't made is that this actually removes some context for the trasnlators as they now cannot see if a string or a number is being inserted. We might want to explicitly use `{:d}` when ever the value is numeric.

Translations/hr.po
```diff
@@ -1203,12 +1203,12 @@ msgstr ""
 
 #: Source/stores.cpp:1129
 #, c-format
-msgid "Talk to %s"
+msgid "Talk to {}"
 msgstr ""
 
 #: Source/stores.cpp:1133
 #, c-format
```
glebm, line 1210: Is `c-format` still the correct thing here?

AJenbo: Not sure, also not sure how xgettext determines this :thinking:

Source/items.cpp
Jmgr: Ah, that is unfortunate. For this PR I have applied a series of regex replacements followed by manual fixes, that means that the initial information (`%i`, `%d`, etc.) has been lost. To replace `%d` with `{:d}` for instance I have to either redo the whole process, or look at the diff and replace `{}` with `{:d}` at the right places.

I think for translations the best thing to do would be to use named insertions, like `{gold}` instead of using `{}` or `{:d}`, but that's a manual operation and a lot of tedious work.

Translations/hr.po
Jmgr: > Is `c-format` still the correct thing here?

I don't think so. poedit removed that when I save a test file with translations using {}.

Source/items.cpp
AJenbo: Yeah I think naming should be done more selective as it appears to require a few other manual change an probably isn't necessary in most cases. Would you be willing to apply the regex to change `%i `to `{:d}`? Else I would do it manually when merging.

Jmgr: > Yeah I think naming should be done more selective as it appears to require a few other manual change an probably isn't necessary in most cases.

Makes sense.

> Would you be willing to apply the regex to change %i to {:d}? Else I would do it manually when merging.

I have squashed multiple commits (one per major step) into one before pushing, but I can maybe come back to the state prior to that rebase using the git reflog. That would prevent me from having to restart whole PR from scratch. I'll try that.

AJenbo: Thanks, reflog can be a life saver :)

Jmgr: Should be better now. I haven't rebased/squashed the commits for now in case something needs updating between one of the steps, but can do once the PR is accepted.

AJenbo (approved):

AJenbo:
@Jmgr congratulations on migrating thing to fmt :) Now we just need to migrate things to std::string :D

Jmgr:
> @Jmgr congratulations on migrating thing to fmt :) Now we just need to migrate things to std::string :D

Thank! Yup, that will be fun :)

AJenbo:
I moved all text rendering to a common code that currently takes `const char*`, I'm hoping that will make the migration a lot easier then before where there where basically 15 different text renders.

PR #1766, opened 2021-04-30
Load/save cleanup

vlolteanu:


glebm (approved):

Source/loadsave.cpp
```diff
@@ -128,25 +128,20 @@ class LoadHelper {
...
 
 public:
 	SaveHelper(const char *szFileName, size_t bufferLen)
+		: m_szFileName(szFileName)
+		, m_buffer(std::make_unique<uint8_t[]>(codec_get_encoded_len(bufferLen)))
+		, m_capacity(bufferLen)
 	{
-		m_szFileName = szFileName;
-		m_capacity = bufferLen;
-		m_buffer = DiabloAllocPtr(codec_get_encoded_len(m_capacity));
```
AJenbo, line 149: What is the benifit of doing the assignment in the signature instead of the body?

vlolteanu: In terms of practicality, none.

AJenbo: typical C++ to have an alternative syntax for things

AJenbo:
looks like this is causing master to fail on switch, is that something you could look into?

vlolteanu:
What do you mean by master failing on switch?

AJenbo:
the CI failed to build the Nintendo switch version https://app.circleci.com/pipelines/github/diasurgical/devilutionX/5283/workflows/54f8113e-8c71-4fe8-8940-c5fb807a7708/jobs/38100

but turns out it was an issue with the server, everything is fine

PR #1767, opened 2021-04-30
Cleanup of doom.cpp

vlolteanu:


Source/doom.cpp
```diff
@@ -44,48 +42,32 @@ int doom_get_frame_from_time()
...
+	} else {
+		doom_close();
 	}
 }
 
 void doom_close()
 {
 	doomflag = false;
-	doom_cleanup();
+	doomCel.reset();
```
glebm, line 70: nit: I find that this reads a bit better:

```suggestion
	doomCel = nullptr;
```

glebm (dismissed):

Source/doom.cpp
```diff
@@ -11,27 +11,25 @@ namespace devilution {
 
 int doom_quest_time;
 int doom_stars_drawn;
-BYTE *pDoomCel;
+std::unique_ptr<uint8_t[]> doomCel;
```
AJenbo, line 14: Globals should be CamelCase

```suggestion
std::unique_ptr<uint8_t[]> DoomCel;
```

vlolteanu: Done

Source/doom.cpp
```diff
@@ -9,10 +9,9 @@
...
 namespace devilution {
 
-int doom_quest_time;
-int doom_stars_drawn;
-std::unique_ptr<uint8_t[]> doomCel;
-bool doomflag;
+int DoomQuestTime;
+std::unique_ptr<uint8_t[]> DoomCel;
+bool DoomFlag;
 int DoomQuestState;
```
AJenbo: thanks, this will make the global cleanup seam less drastic :D

Source/doom.cpp
AJenbo: would you apply this also?

vlolteanu: Sure, why not?

Source/doom.cpp
vlolteanu: No problem.

PR #1768, opened 2021-04-30
🐞 Fix memory leaks in utils/language.{h,cpp}

glebm:
This is a minimal refactoring to fix the memory leaks which have made reading debug output impossible.


Source/utils/language.cpp
```diff
@@ -1,217 +1,187 @@
...
+	// Read strings described by entries
+	for (uint32_t i = 0; i < head.nb_mappings; i++) {
+		std::vector<char> key;
+		std::vector<char> value;
+		if (read_entry(fp, &src[i], key) && read_entry(fp, &dst[i], value)) {
+			if (key.empty()) {
+				utf8 = parse_metadata(value.data());
+			} else {
+				if (utf8) {
+					map[utf8_to_latin1(key.data())] = utf8_to_latin1(value.data());
```
AJenbo, line 180: We don't need to decode the key

AJenbo:
@Jmgr this should make your life easier

Source/utils/language.cpp
glebm: Done

PR #1772, opened 2021-04-30
Get rid of mem management macros from storm_sdl_rw.cpp

vlolteanu:


vlolteanu:
No need to leave the dunce cap of DiabloAllocPtr on. Outside of patching SDL, there's no way to use unique_ptr here.

glebm (approved):

PR #1774, opened 2021-04-30
Get rid of manual memory management for catacomb generation

vlolteanu:


AJenbo:
GCC 6.5 appears unhappy

```
/root/repo/Source/drlg_l2.cpp: In function 'void devilution::ConnectHall(const devilution::HALLNODE&)':
/root/repo/Source/drlg_l2.cpp:2106:7: error: expected unqualified-id before '[' token
  auto [nX1, nY1, nX2, nY2, nHd] = node;
       ^
```

AJenbo (approved):

PR #1779, opened 2021-04-30
Adds `MeasureCelWidth` to quickly measure CEL width

glebm:
This could be optimized further for CELs with frame headers,
where the width can be obtained directly from the header

https://github.com/savagesteel/d1-file-formats/blob/master/PC-Mac/CEL.md#42-cel-frame-data

Untested but hopefully works (perhaps with minor fixes)
For use in @qndel's Alt highlighting PR

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -35,7 +35,40 @@ const uint32_t RndMult = 0x015A4E35;
 
 namespace {
 
-constexpr std::uint8_t CelEol = 0x80;
+/**
+ * Cel frame data encoding bytes.
+ * See https://github.com/savagesteel/d1-file-formats/blob/master/PC-Mac/CEL.md#42-cel-frame-data
+ */
+
+/** [0, 0x7E]: followed by this many pixels. Ends the line. */
+constexpr std::uint8_t CelPixelsEolMax = 0x7E;
+
+/** 0x7F: followed by 128 (0x7F) pixels. Does not end the line. */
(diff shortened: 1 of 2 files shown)


Source/engine.cpp
```diff
@@ -49,6 +82,32 @@ BYTE *GetLightTable(char light) {
 
 } // namespace
 
+std::uint_fast16_t MeasureCelWidth(BYTE *pCelBuf) {
```
AJenbo, line 85: Considering that we have all the cel widths hardcoded in the items them selfs, is this measuring the non-transparent part of the graphics, and not the cell itself?

AJenbo: On a side note, this should be run when loading the item graphics and assigned to a new item property so that it doesn't run each frame.

glebm: Yeah, we'll need to know the dimensions anyway when doing clipped rendering

glebm: > is this measuring the non-transparent part of the graphics, and not the cell itself?

Pretty sure this is measuring the Cel width itself.

AJenbo: I think what @qndel wanted as a function that measure the bounding box of the graphics in the cell. Some items can look quite off-center because of where in the sprite the graphics appear.

glebm: Yeah I've understood that now. Such a function can be easily written, and @qndel's hack doesn't quite do that actually.

Issue #1755, opened 2021-04-29
Bard's speed values are different from vanilla


GrayFace:
**Important information**
Operating system Win 7 64 / DevilutionX version 1.2.1 / downloaded release

**Describe**
See https://github.com/diasurgical/devilutionX/discussions/1723
Bard's speed with Sword and Bow are the same as Warrior instead of Rogue.
This makes Bard pretty much the strongest melee fighter, at least in single player where you don't grind items and can't hope to find "King's XXX of Haste" attributes combo.

PR #1780, opened 2021-04-30
ADL: Fix calculation of fast/faster/fastest combination (Bard)

obligaron:
https://github.com/diasurgical/devilution/blob/2bf6eeb125a540e83ec838d6ab9181f2b26d0983/Source/player.cpp#L3206-3221

For Bard it's possible to wield two weapons.
That mean's it's also possible that two attack modifiers are present at the same time.
In this case the skipped frame calculation doesn't calculate the skipped frames correctly, cause the combined skipping of two attack modifiers can skip the frame that triggers the next skipping.
This is the same behavior as the combined hit recovery modifiers (zen mode).

Example:

- Bard with two swords, one with fast the other with faster attack modifier
- `PM_DoAttack` at Frame 3 advances two Frames (one for Fast Attack and one for Faster Attack.) So the Frame is 5.
- Then ProcessAnimation advances the Frame again. So now the Frame is 6.
- In the next game tick `PM_DoAttack` does see Frame 6 so the check for `frame == 5` for Faster Attack is skipped and so is the skipping skipped. 😁

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -1464,14 +1464,14 @@ void StartAttack(int pnum, direction d)
 	}
 
 	int skippedAnimationFrames = 0;
-	if ((plr[pnum]._pIFlags & ISPL_FASTATTACK) != 0) {
-		skippedAnimationFrames += 1;
-	}
 	if ((plr[pnum]._pIFlags & ISPL_FASTERATTACK) != 0) {
-		skippedAnimationFrames += 2;
-	}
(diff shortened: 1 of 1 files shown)


AJenbo:
Now that animation distribution no longer is tired the logic maybe we should simplify this to more cleanly cut the length of the animation rather than trying to skipping individual frames.

obligaron:
> 
> 
> Is this solving #1755 ?

No but I will prepare another PR for that topic (as far as I have seen this bug is not related to ADL).

julealgon:
What does "ADL" stand for?

PR #1781, opened 2021-04-30
ADL: Advance skipped Frames in SetNewAnimation and remove logic from PM_XYZ

obligaron:
Before the game tried to distribute the skipped Frames across the game ticks/frames.
This resulted in checks like `if (plr[pnum]._pIFlags & ISPL_FASTRECOVER && frame == 3) {`.

This job is now done by the Animation Distribution Logic.
But ADL needs to know in advance how long an animation is shown (time represented as game ticks).

Simplified example with an Animation with 8 Frames and GameTick = 50ms:

- Animation with 8 Frames takes 8 * 50 ms = 400ms
- Same Animation with 2 skipped Frames takes 6 * 50ms = 300ms
  - The original logic removed 2 frames (we missed you nice content). The resulted in the 300ms shown Animation. So each Frame was constantly shown 50 ms (with or without skipping).
  - ADL also skips the 2 game ticks but it doesn't remove the frames. It shortens the frame. The calculation is frame time = 300ms / 8. So each frame is now shown 37,5ms and not 50ms. So we got the same time but all frames are shown! So every bit of nice content stays. 🙂 

But the result is that we have the logic to calculate the skipped frames in two places: the original game logic and ADL. This is not a good design and error prone.

With this pr this is changing.

Notes:

- The original skipping logic is removed.
- For the game logic the skipping now happens always at the start of the Animation. But as far as I have seen no one expect the skipping logic checks frames before the animation hits or is finished.
- This changes introduces a minor breaking change with vanilla.
   - Conditions (all must be true)
      - Skipping is involved (for example faster hit recovery)
      - You save the game while the animation is shown (for example directly after the animation started)
   - Logic/Problem
      - ADL adds the skipping directly at start (for example advances CurrentFrame from 2 to 4)
      - If now loaded in vanilla the old logic also adds the skipping. So for **one** Animation (the saved) you can be faster.
    -  In DevilutionX this can't happen cause we don't apply the logic again. Only the rendered Frame can differ a little, cause the save game logic doesn't contain the new variables from ADL that contains the distribution variables.

### DIFF
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -64,7 +64,7 @@ void AnimationInfo::SetNewAnimation(byte *pData, int numberOfFrames, int delayLe
 
 	this->pData = pData;
 	NumberOfFrames = numberOfFrames;
-	CurrentFrame = 1;
+	CurrentFrame = 1 + numSkippedFrames;
 	DelayCounter = 0;
 	DelayLen = delayLen;
 	TicksSinceSequenceStarted = 0;
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2808,26 +2808,13 @@ bool PlrHitObj(int pnum, int mx, int my)
 
 bool PM_DoAttack(int pnum)
 {
-	int frame, dx, dy, m;
+	int dx, dy, m;
 	bool didhit = false;
 
 	if ((DWORD)pnum >= MAX_PLRS) {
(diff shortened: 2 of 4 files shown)


AJenbo:
I'm ok with the save game compatibility change here. We already diverge from Hellfire in the same way as it skips the end of the animation and doesn't try to distribute the frames.

AJenbo (approved):

PR #1783, opened 2021-04-30
Place LigthTbl in bss

vlolteanu:
LigthTbl is always in use while playing. Sparing 7K of RAM while in the menu (and presumably about to start playing) is pointless.


AJenbo (approved):

vlolteanu:
This was fixed in some other PR.

PR #1784, opened 2021-04-30
Translation quickfix

mmaret:
c-format string are checked by poedit
 "hit steals 5% life" could be translated into "vole 5% de vie". and poedit would like to make sure that %d would be filled with a int

### DIFF
--- a/Source/controls/modifier_hints.cpp
+++ b/Source/controls/modifier_hints.cpp
@@ -30,14 +30,14 @@ int SpaceWidth()
 struct CircleMenuHint {
 	CircleMenuHint(bool isDpad, const char *top, const char *right, const char *bottom, const char *left)
 	    : is_dpad(isDpad)
-	    , top(_(top))
-	    , top_w(CalculateTextWidth(_(top)))
-	    , right(_(right))
-	    , right_w(CalculateTextWidth(_(right)))
-	    , bottom(_(bottom))
-	    , bottom_w(CalculateTextWidth(_(bottom)))
-	    , left(_(left))
-	    , left_w(CalculateTextWidth(_(left)))
+	    , top(top)
+	    , top_w(CalculateTextWidth(top))
+	    , right(right)
+	    , right_w(CalculateTextWidth(right))
+	    , bottom(bottom)
+	    , bottom_w(CalculateTextWidth(bottom))
+	    , left(left)
+	    , left_w(CalculateTextWidth(left))
 	    , x_mid(left_w + SpaceWidth() * 2.5)
 	{
 	}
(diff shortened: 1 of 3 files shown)


Translations/devilutionx.pot
```diff
@@ -3652,7 +3652,6 @@ msgid "knocks target back"
 msgstr ""
 
 #: Source/items.cpp:3619
-#, c-format
```
AJenbo, line 3655: Was this autogenerated or did you manually remove it? If it was done manually then this will just return the next time we update the translation :(

Source/stores.cpp
```diff
@@ -161,14 +161,14 @@ void PrintStoreItem(ItemStruct *x, int l, text_color iclr)
 		if (x->_iSufPower != -1) {
 			PrintItemPower(x->_iSufPower, x);
 			if (sstr[0])
-				strcat(sstr, _(",  "));
+				strcat(sstr, ",  ");
```
AJenbo, line 164: For now, I think it's best to keep them as some languages may not use comma to separate a series of nouns. In the belong run we should not use `strcat` either and instead have something more expressive like `"{power}, {power}"`;

Source/controls/modifier_hints.cpp
```diff
@@ -114,8 +114,8 @@ void DrawStartModifierMenu(const CelOutputBuffer &out)
 {
 	if (!start_modifier_active)
 		return;
-	static const CircleMenuHint dPad(/*is_dpad=*/true, /*top=*/N_("Menu"), /*right=*/N_("Inv"), /*bottom=*/N_("Map"), /*left=*/N_("Char"));
-	static const CircleMenuHint buttons(/*is_dpad=*/false, /*top=*/"", /*right=*/"", /*bottom=*/N_("Spells"), /*left=*/N_("Quests"));
+	static const CircleMenuHint dPad(/*is_dpad=*/true, /*top=*/_("Menu"), /*right=*/_("Inv"), /*bottom=*/_("Map"), /*left=*/_("Char"));
+	static const CircleMenuHint buttons(/*is_dpad=*/false, /*top=*/"", /*right=*/"", /*bottom=*/_("Spells"), /*left=*/_("Quests"));
```
AJenbo, line 118: If this works then we need to remove the additional translation from `Source/controls/modifier_hints.cpp::33-40`

AJenbo (changes requested):
Thanks but some of it needs some adjustment.

Translations/devilutionx.pot
mmaret: Ok done in items.cpp that will propagate it into devilution.pot

Source/stores.cpp
mmaret: Fixed

Source/controls/modifier_hints.cpp
mmaret: Yes it does :)
I've done those change before the add of the translation line33-40.
I've re-check and the hints are properly tranlated into the bellow screenshot

Capture d’écran du 2021-05-01 14-29-31

Translations/devilutionx.pot
AJenbo: perfect

AJenbo (approved):

PR #1787, opened 2021-05-01
Add a demo recorder and replayer

Jmgr:
This adds a demo recorder and replayer PoC. Is it based on frame count and not on time, so should be resilient to CPU lag (and manages level loading).

Start/stop recording by pressing the right control key, and start/stop replaying using the right shift key.

This is missing quite a few features (like writing to file, saving random seeds, etc), so this is only to test if this can reliably record and replay gameplay. Having a record/replay system could help us to automatically detect crashes, regressions and desync bugs. Players could also record a demo to report a bug or just for fun.

See https://github.com/diasurgical/devilutionX/pull/904 for another implementation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -359,6 +359,7 @@ set(devilutionx_SRCS
   Source/utils/language.cpp
   Source/utils/paths.cpp
   Source/utils/thread.cpp
+  Source/utils/recorder.cpp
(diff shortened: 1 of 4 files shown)


AJenbo:
Closing as the alternate implementation now works and is more feature complete, sadly not as clean as this one :)

PR #1788, opened 2021-05-01
Added the ability to permanently cache a file

vlolteanu:
This patch lets you permanently cache the contents of any data file, with the caveat that the returned buffer is const.

There are multiple instances where this might be useful, like e.g. when loading mouse cursors.

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -11,6 +11,7 @@
  * - Video playback
  */
 
+#include <unordered_map>
 #include "lighting.h"
 #include "movie.h"
 #include "options.h"
@@ -654,6 +655,56 @@ BYTE *LoadFileInMem(const char *pszName, DWORD *pdwFileLen)
 	SFileCloseFile(file);
 
 	return buf;
+}
+
+/**
+ * @brief Load a file in to a buffer
+ * @param pszName Path of file
+ * @param pdwFileLen Will be set to file size if non-NULL
+ * @return Buffer with content of file
+ */
+std::unique_ptr<uint8_t[]> LoadFileInMemSafe(const char *pszName, uint32_t *pdwFileLen)
+{
+	HANDLE file;
+	SFileOpenFile(pszName, &file);
+	int fileLen = SFileGetFileSize(file, nullptr);
+
+	if (pdwFileLen != nullptr)
+		*pdwFileLen = fileLen;
+
+	if (fileLen == 0)
+		app_fatal("Zero length SFILE:\n%s", pszName);
+
+	auto buf = std::make_unique<uint8_t[]>(fileLen);
+
+	SFileReadFile(file, buf.get(), fileLen, nullptr, nullptr);
+	SFileCloseFile(file);
+
+	return buf;
+}
+
+const uint8_t *CacheFile(const char *name, uint32_t *pdwFileLen)
+{
+	using Key = std::string;
+	using Value = std::pair<std::unique_ptr<uint8_t[]>, uint32_t>;
+
+	static std::unordered_map<Key, Value> cachedFiles;
+
+	auto it = cachedFiles.find(name);
+
+	if (it != cachedFiles.end()) {
+		uint32_t size;
+		auto buf = LoadFileInMemSafe(name, &size);
+
+		Key key = std::string(name);
+		Value value = {std::move(buf), size};
+
+		it = cachedFiles.insert(std::make_pair<Key, Value>(std::move(key), std::move(value))).first;
+	}
+
+	if (pdwFileLen != nullptr)
+		*pdwFileLen = it->second.second;
+	return it->second.first.get();
 }
 
 /**
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -17,6 +17,7 @@
 #include <cstdlib>
 #include <SDL.h>
 #include <cstdint>
+#include <memory>
 
 #ifdef USE_SDL1
 #include "utils/sdl2_to_1_2_backports.h"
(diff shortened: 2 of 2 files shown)


AJenbo:
I don't really see the benefit/use of this. Most OSs will cache file content in ram if there is enough space, most assets are loaded into memory on game start (including the cursor) and stayed there till shutdown. The rest is deliberately streamed or specifically loaded and unloaded as needed to manage ram usage.

vlolteanu:
I see your point; given how much RAM we have nowadays, I won't be surprised if almost all of the mpq is loaded into the buffer cache after a few hours of playing.

AJenbo:
When playing from cd it's pretty obvious that this is happening :D

vlolteanu:
As for the benefits, there are a couple:
 * Programatically, you won't have to write init/cleanup code for the files whose content is always needed (only uint8_t someGlobalCelFile = CacheFile(....))
 * Some file contents are freed over-zealously; this API would be a stop-gap before a proper cleanup. Loading them again from the buffer cache entails at least a decompression and a memory copy

AJenbo:
I don't think we have any noticeable performance or other issues. So let's instead just clean up the areas where things aren't handled ideally.

vlolteanu:
We've still got loading screens, and this is a 20+-year old game. As an extreme measure, we could pre-load and pre-compute everything and the RAM hit would be less annoying that the loading screens themselves.

AJenbo:
The load screens animation takes 1.45 sec alone. For me, load time is about 2 sec 
When measuring what is taking the time I get the following:

```
114ms CreateLevel
 20ms InitMissileGFX 
 20ms LoadLevel 
 19ms InitPlayerGFX
 18ms LoadLvlGFX 
  5ms InitObjectGFX 
  2ms HoldThemeRooms 
  1ms MakeLightTable 
  1ms SetDungeonMicros 
```

Only a total of 62ms is spend loading assets. The vast majority is spent generating the level layout (118ms) and then 20ms is spend loading the level state from the scratch file. So even in the best-case scenario caching files in RAM would at best reduce load time by 40% (82ms). Which is probably not even perceivable.

Caching files in RAM could also significantly hurt some of the low spec systems that we support.

Lets identify an actual problem before we start trying to solve it.

vlolteanu:
That's a bit off. Imagine what 118ms of contemporary CPU time would mean in 1996.

Would it surprise you to know that CreateLevel also calls LoadFileInMem?

AJenbo:
It wouldn't, no. But it's probably not reading a lot of data. If you feel it's important then feel free to post some more exact measurements.

vlolteanu:
Nope, sorry.

AJenbo:
No need to feel sorry, it just seamed you where arguing strongly for this.

vlolteanu:
I meant that I gave up trying to convince you. I still have a strong opinion on the matter.

glebm:
If we want to keep something around, we can do so directly.
That said, peak RAM usage is important for supporting older devices.
For newer devices, decompression time is 62ms which doesn't really matter.

glebm:
One optimization that would benefit both would be to only unload/reload the diff between assets in 2 levels. Currently I think we unload everything first and the reload it.

vlolteanu:
I wasn't arguing to keep everything in memory. Certain stuff invariably gets loaded, and I wouldn't optimize RAM usage in the menu or in Tristram.

A proper cleanup would indeed be the end-goal. The intention for this was for it to be a stop-gap.

vlolteanu:
On the point of caching files even more aggressively, being greedy with RAM usage is not an issue on modern machines, but even small blips in performance are noticeable by users.

As for supporting older systems, we could have a compile-time option to disable or change how aggressively file caching is carried out.  A RPi with 2GB of RAM is under 50 EUR, and the older machines that you're referencing are outliers.

glebm:
> and the older machines that you're referencing are outliers.

We aim to support a wide variety of devices. Some of these are handheld devices, such as PSP, Vita, 3DS, etc. I'm currently trying to make DevilutionX run on a device with 25 MiB free RAM.

Regardless, rather than doing a "stop-gap" solution for a non-existent problem, a clean fix would be much better.

Besides, this cache won't fix the load screens at all. As @Ajenbo mentioned, the "long" load screens happen not due to actual loading, but due to the fact that loading happens on the same thread as animation and always takes at least 1 second (0.5s fade-in and 0.5s fadeout).

vlolteanu:
This is getting really off-topic, and it was a topic that I wanted to quit arguing.

The ultimate point of all of this is that loading screens can be done away with entirely on modern systems. The fade in/out is arbitrary and can be removed. Levels can be pre-computed at game start, and assets can be loaded and kept in memory.

In the absence of a loading screen, 100+ms of unresponsiveness is noticeable.

As for going straight for a clean solution, it simply takes time. Try referencing all the data loaded from files via unique_ptr::get and you'll see what I mean.

glebm:
> The ultimate point of all of this is that loading screens can be done away with entirely on modern systems. The fade in/out is arbitrary and can be removed. Levels can be pre-computed at game start, and assets can be loaded and kept in memory.

Absolutely right. This is where the biggest 1 second win would be.

> Try referencing all the data loaded from files via unique_ptr::get and you'll see what I mean.

I actually did this a few hours ago as part of https://github.com/diasurgical/devilutionX/commit/1b8b62a8c20d7e5115345c3aea4041331a84311c :)

AJenbo:
In DiabloUI id did an implementation of the fade-in/out that isn't blocking, if we can apply that to the load screen it would be a big win to the load speed.

> As for going straight for a clean solution, it simply takes time.

No arguing there, been doing this almost daily for 2 and half years just to get to the current state :)

PR #1789, opened 2021-05-01
Fix sound volume/panning attenuation

### LINKED ISSUE #1632: Changing volume in options reduces the effect of sound volume changing based on source distance
All sound effects change their volume based on the distance to the player. This is easy to test in multiplayer by having another player walk past the local client. The closer the player is, the louder the footsteps.

However, if you reduce the volume slider in options, this effect is reduced. If you change sound volume to the lowest setting where sound is still audible, everything will be at the same volume no matter how far away the sound source is.

I tested this in both original Diablo and DevilutionX 1.2.0, and the behaviour is the same in both.

I think the bug is caused by `static int CapVolume(int volume)` in sound.cpp

thebigMuh:
Fixes #1632 

The panning and distance attenuation calculations were very strange and nonsensical.
- Volume doesn't abruptly change from 'sorta half quiet' to 'muted on the last notch of the slider.
- Volume sliders are logarithmic, which feels way nicer than linear attenuation.
- Panning was completely hosed, and caused strange cutoff issues. You could hear only about 25 tiles left/right of your character before the sound volume snapped to zero, but about 100 tiles directly up/down.
- Volume sliders now have 64 steps (instead of 16), so you can actually have some reasonably fine control over volume. Each notch equals a bit more than 1.5 dB.

### DIFF
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -94,11 +94,11 @@ struct Point {
 	}
 
 	/**
-	 * @brief Computes fast approximate distance between this and other, using only integer arithmetic, with less than ~5% error
+	 * @brief Fast approximate distance between two point, using only integer arithmetic, with less than ~5% error
 	 * @param other Pointer to which we want the distance
 	 * @return Magnitude of vector this -> other
 	*/
-	int ApproxDistance(const Point &other) const
+	int ApproxDistance(Point other) const
 	{
 		int min;
 		int max;
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -98,7 +98,8 @@ struct Point {
 
 	 * @param other Pointer to which we want the distance
 	 * @return Magnitude of vector this -> other
-	*/
+	 */
+
 	int ApproxDistance(Point other) const
 	{
 		int min;
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -95,7 +95,6 @@ struct Point {
 
 	/**
 	 * @brief Fast approximate distance between two points, using only integer arithmetic, with less than ~5% error
-
 	 * @param other Pointer to which we want the distance
 	 * @return Magnitude of vector this -> other
 	 */
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -17,6 +17,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <memory>
+#include <tuple>
 #include <utility>
 
 #include <SDL.h>
@@ -90,6 +91,25 @@ struct Point {
 	{
 		a -= b;
 		return a;
+	}
+
+	/**
+	 * @brief Computes fast approximate distance between this and other, using only integer arithmetic, with less than ~5% error
+	 * @param other Pointer to which we want the distance
+	 * @return Magnitude of vector this -> other
+	*/
+	int ApproxDistance(const Point &other) const
+	{
+		int min;
+		int max;
+
+		std::tie(min, max) = std::minmax(std::abs(other.x - x), std::abs(other.y - y));
+
+		int approx = max * 1007 + min * 441;
+		if (max < (min * 16))
+			approx -= max * 40;
+
+		return (approx + 512) / 1024;
 	}
 };
 
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -94,7 +94,8 @@ struct Point {
 	}
 
 	/**
-	 * @brief Fast approximate distance between two point, using only integer arithmetic, with less than ~5% error
+	 * @brief Fast approximate distance between two points, using only integer arithmetic, with less than ~5% error
+
 	 * @param other Pointer to which we want the distance
 	 * @return Magnitude of vector this -> other
 	*/
--- a/Source/utils/soundsample.cpp
+++ b/Source/utils/soundsample.cpp
@@ -25,9 +25,11 @@ namespace {
 
 constexpr float LogBase = 10.0f;
 
-// Scaling factor for attenuating volume.
-// Picked so that a volume change of -10 dB results in half perceived loudness.
-// VolumeScale = -1000 / log(0.5)
+/**
+ * Scaling factor for attenuating volume.
+ * Picked so that a volume change of -10 dB results in half perceived loudness.
+ * VolumeScale = -1000 / log(0.5)
+ */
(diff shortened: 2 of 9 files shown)


Source/engine.h
```diff
@@ -96,6 +97,25 @@ struct Point {
 		a -= b;
 		return a;
 	}
+
+	/**
+	 * @brief Computes fast approximate distance between this and other, using only integer arithmetic, with less than ~5% error
```
AJenbo, line 102:
```suggestion
	 * @brief Fast approximate distance between two points, using only integer arithmetic, with less than ~5% error
```

AJenbo (changes requested):
The tests need to be updated as they are currently failing with the new changes.

Source/engine.h
```diff
@@ -96,6 +97,25 @@ struct Point {
 		a -= b;
 		return a;
 	}
+
+	/**
+	 * @brief Computes fast approximate distance between this and other, using only integer arithmetic, with less than ~5% error
+	 * @param other Pointer to which we want the distance
+	 * @return Magnitude of vector this -> other
+	*/
+	int ApproxDistance(const Point &other) const
```
glebm, line 106: nit: `Point` is tiny, pass it by value.

julealgon: Is there any overhead associated with passing it as a reference?

glebm: Yes, sometimes. See the generated code: https://godbolt.org/z/4ojed9f1x

glebm: Basically, a reference is effectively a pointer. So you're passing an address. The function needs to go to memory to fetch the values.

When you're passing a small value, it is passed in registers, the function has the values right away.

Source/engine.h
```diff
@@ -91,6 +92,25 @@ struct Point {
 		a -= b;
 		return a;
 	}
+
+	/**
+	 * @brief Fast approximate distance between two point, using only integer arithmetic, with less than ~5% error
```
AJenbo, line 97:
```suggestion
	 * @brief Fast approximate distance between two points, using only integer arithmetic, with less than ~5% error

```

Source/engine.h
```diff
@@ -91,6 +92,25 @@ struct Point {
 		a -= b;
 		return a;
 	}
+
+	/**
+	 * @brief Fast approximate distance between two point, using only integer arithmetic, with less than ~5% error
+	 * @param other Pointer to which we want the distance
+	 * @return Magnitude of vector this -> other
+	*/
```
AJenbo, line 100:
```suggestion
	 */

```

Source/utils/soundsample.cpp
```diff
@@ -16,10 +16,48 @@
...
 
 namespace devilution {
 
+namespace {
+
+constexpr float LogBase = 10.0f;
+
+// Scaling factor for attenuating volume.
+// Picked so that a volume change of -10 dB results in half perceived loudness.
+// VolumeScale = -1000 / log(0.5)
```
AJenbo, line 30: We use Doxygen for documentation
```suggestion
/**
 * Scaling factor for attenuating volume.
 * Picked so that a volume change of -10 dB results in half perceived loudness.
 * VolumeScale = -1000 / log(0.5)
 */
```

Source/utils/soundsample.cpp
```diff
@@ -16,10 +16,48 @@
...
+constexpr float VolumeScale = 3321.9281f;
+
+// Min and max volume range, in millibel.
+// -100 dB (muted) to 0 dB (max. loudness).
```
AJenbo, line 34:
```suggestion
/**
 * Min and max volume range, in millibel.
 * -100 dB (muted) to 0 dB (max. loudness).
/*
```

Source/utils/soundsample.cpp
```diff
@@ -16,10 +16,48 @@
...
+constexpr float MillibelMin = -10000.0f;
+constexpr float MillibelMax = 0.0f;
+
+// Stereo separation factor for left/right speaker panning. Lower values increase separation, moving sounds further left/right, while higher values will pull sounds more towards the middle, reducing separation.
+// Current value is tuned to have ~2:1 mix for sounds that happen on the edge of a 640x480 screen.
```
AJenbo, line 39: Break up extremely long lines, if they do not fit on screen they are hard to read...
```suggestion
/**
 * Stereo separation factor for left/right speaker panning. Lower values increase separation, moving
 * sounds further left/right, while higher values will pull sounds more towards the middle, reducing separation.
 * Current value is tuned to have ~2:1 mix for sounds that happen on the edge of a 640x480 screen.
 */
```

Source/utils/soundsample.cpp
```diff
@@ -16,10 +16,48 @@
...
+constexpr float StereoSeparation = 6000.0f;
+
+float PanLogToLinear(int logPan)
+{
+	if (logPan == 0)
+		return 0;
+
+	return copysign(1.0f - std::pow(LogBase, static_cast<float>(-std::abs(logPan)) / StereoSeparation), static_cast<float>(logPan));
```
AJenbo, line 47:
```suggestion
	auto factor = std::pow(LogBase, static_cast<float>(-std::abs(logPan)) / StereoSeparation);
	
	return copysign(1.0f - factor, static_cast<float>(logPan));
```

AJenbo (changes requested):
Just some readability and code format, other then that it looks good 👍

Source/engine.h
```diff
@@ -91,6 +92,27 @@ struct Point {
 		a -= b;
 		return a;
 	}
+
+	/**
+	 * @brief Fast approximate distance between two points, using only integer arithmetic, with less than ~5% error
+
```
AJenbo, line 98:
```suggestion
```

StephenCWills:
This change was alarming! My sliders were so far down from before that when I pulled this change in, I couldn't hear anything and thought the audio was broken.

thebigMuh:
> This change was alarming! My sliders were so far down from before that when I pulled this change in, I couldn't hear anything and thought the audio was broken.

Sorry, side effect of the logarithmic sliders.

Hopefully the entire slider now has 'meaning', whereas before I found that the bottom few notches packed huge changes in perceived volume, while the top 50% didn't change very much at all.

PR #1791, opened 2021-05-01
Suport for plural translation forms

AJenbo:
Also noticed that a couple of elements were using the translation before the language had been loaded.


john-tornblom:
If there is a need, I can write an generic interpreter for the gettext plural language, the grammar is pretty simple (see http://git.savannah.gnu.org/cgit/gettext.git/tree/gettext-runtime/intl/plural.y)

Hopefully I can get away with a simple recursive decent parser.

AJenbo:
Currently all languages that we have are covered, but ight be good to have one that can handle all future translations

john-tornblom:
Alright, lets see how the complexity of the current implementation grows as more translations are added, and let me know if things become difficult to maintain.

AJenbo:
@john-tornblom if you could look in to automatically detecting the default translation based on the system language that would be a big help :)

john-tornblom:
Since the preferred user locale is so dependent on the operating system being used, I think the best approach is to use some well-established API. This was actually added to SDL recently (version 2.0.14), see the function SDL_GetPreferredLocales(). Would that be sufficient?

AJenbo:
Yeah, that would cover us on Windows and some other systems. We will need a fallback for older versions of SDL but that can just return `en` for now.

If we want to go a bit further we would pull in https://icculus.org/~aspirin/findlocale/ as the fallback for when SDL 2.0.14 isen't being linked.

john-tornblom:
POSIX and windows should not be a problem to add without library support, but vita, switch, 3ds is more difficult.

AJenbo:
Lets do it on a best effort basis :)

PR #1792, opened 2021-05-01
Fix Bard incorrect number of sword animation frames

### LINKED ISSUE #1755: Bard's speed values are different from vanilla

obligaron:
Fixes #1755

Before the fix:

Class | Weapon | Hitframe | Number of Frames
-- | -- | -- | --
Bard | Sword | 10 | 10
Bard | Bow | 11 | 12
Warrior | Sword | 9 | 16
Warrior | Bow | 11 | 16
Rogue | Sword | 10 | 18
Rogue | Bow | 7 | 12

You can see that the Number of Frames differ for Bard and Rogue but not the Hitframe (the Frame where the Attack hits and after that the Animation can be canceled or next Attack can be executed).
So it's not overpowered but the animation is cropped.
The fix is to restore the original 18 Number of Frames for Sword Attacks.

Notes:

- The thing that I don't understand is why this didn't occur in vanilla Hellfire. It could mean it was fixed in another way or the source in [devilution](https://github.com/diasurgical/devilution/blob/2bf6eeb125a540e83ec838d6ab9181f2b26d0983/Source/player.cpp#L674) is wrong.
- The animation for Bard with Bow is not nice. Cause it fires at Frame 10 and at Frame 10 the Bow is already lowered. So this doesn't look good. But this is another issue.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -674,8 +674,6 @@ void SetPlrAnims(int pnum)
 		} else if (gn == ANIM_ID_STAFF) {
 			plr[pnum]._pAFrames = 16;
 			plr[pnum]._pAFNum = 11;
-		} else if (gn == ANIM_ID_SWORD_SHIELD || gn == ANIM_ID_SWORD) {
-			plr[pnum]._pAFrames = 10;
 		}
 	} else if (pc == HeroClass::Barbarian) {
 		if (gn == ANIM_ID_AXE) {


galaxyhaxz:
The devilution source (and thus DevilutionX) is wrong. It should be 
`plr[pnum]._pAFNum = 10;`
 instead of
`plr[pnum]._pAFrames = 10;`

obligaron:
> 
> 
> The devilution source (and thus DevilutionX) is wrong. It should be
> `plr[pnum]._pAFNum = 10;`
> instead of
> `plr[pnum]._pAFrames = 10;`

I also think so. 🙂 
But even `plr[pnum]._pAFNum = 10;` is pointless, cause `_pAFNum` is the default value in `PlrGFXAnimLens` for Bard so I just removed it 😉

galaxyhaxz:
>  for Bard so I just removed it wink

Yep, your fix achieves the same thing. The vanilla hellfire code however still sets it to 10. They were planning on making graphics for the bard and barbarian but never did.

AJenbo (approved):

GrayFace:
> So it's not overpowered but the animation is cropped.

Maybe there is 1 extra frame required to cancel animation and go for next swing? Because it definitely feels like Bard swings as fast as Warrior, not as Rogue.

obligaron:
> 
> Maybe there is 1 extra frame required to cancel animation and go for next swing? Because it definitely feels like Bard swings as fast as Warrior, not as Rogue.

I think you are right. 👍 
It would be 1 Frame faster, cause before the next attack would start at Animation Frame 11 (Hit Frame + 1). Now it would start at Animation Frame 10.
Sorry for the mistake on my side.

Issue #1782, opened 2021-04-30
Keyboard hit "P" - "Pause" stops music playing


Chance4us:
To reproduce just hit "P"

Chance4us:
Caused by commit acee2ef14ccad779d22554a7df6881877ea50a4b

AJenbo:
This is caused by sound_stop() now also stopping music.

Chance4us:
> Could you plz test if it now also interrupts dialog (like "the sanctity of the place has been fouled") and if that was the case in vanilla?

Yes. I will report.

Chance4us:
It interrupted in fact a dialog "It is hot down here".
The dialogue does not continue to play when you exit the pause mode.

PR #1793, opened 2021-05-01
Fix "Pause" stops music playing

Chance4us:
Fix for Keyboard hit "P" - "Pause" stops music playing #1782
sound_stop makes no sense when the game is paused.
In vanilla the music does not stop as well.


AJenbo (changes requested):
The problem is that sound_stop() now causes much to top. Pausing the game should stop any ongoing sound effects.

PR #1795, opened 2021-05-01
Enable ADL for Items

obligaron:
Contributes to #838

This uses `AnimationInfo` in `ItemStruct`.

Notes:

- This change shows the first Frame of Flipping that wasn't shown previous.
- Magic Rock uses normal/old Logic cause it's the only item that didn't stop the Animation with the last Frame, so ADL doesn't help here anyway.

<details><summary>Example Video (5% game speed)</summary>

## devilutionX 1.21

https://user-images.githubusercontent.com/25415264/119418493-6bb70580-bcf8-11eb-86c6-5a45696df657.mp4

## PR

https://user-images.githubusercontent.com/25415264/119418507-71145000-bcf8-11eb-9362-23cae37c8c6f.mp4

</details>

### DIFF
--- a/Source/qol/itemlabels.cpp
+++ b/Source/qol/itemlabels.cpp
@@ -74,7 +74,7 @@ void AddItemToLabelQueue(int id, int x, int y)
 	nameWidth += marginX * 2;
 	int index = ItemCAnimTbl[it->_iCurs];
 	if (!labelCenterOffsets[index]) {
-		std::pair<int, int> itemBounds = MeasureSolidHorizontalBounds(*it->_iAnimData, it->_iAnimFrame);
+		std::pair<int, int> itemBounds = MeasureSolidHorizontalBounds(*it->AnimInfo.pCelSprite, it->AnimInfo.CurrentFrame);
 		labelCenterOffsets[index].emplace((itemBounds.first + itemBounds.second) / 2);
 	}
 
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -627,7 +627,7 @@ static void DrawItem(const CelOutputBuffer &out, int x, int y, int sx, int sy, b
 		return;
 	}
 
-	int nCel = pItem->AnimInfo.CurrentFrame;
+	int nCel = pItem->AnimInfo.GetFrameToUseForRendering();
 	int frames = SDL_SwapLE32(*(DWORD *)cel->Data());
 	if (nCel < 1 || frames > 50 || nCel > frames) {
 		Log("Draw \"{}\" Item 1: frame {} of {}, item type=={}", pItem->_iIName, nCel, frames, pItem->_itype);
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -621,13 +621,13 @@ static void DrawItem(const CelOutputBuffer &out, int x, int y, int sx, int sy, b
 	if (pItem->_iPostDraw == pre)
 		return;
 
-	auto *cel = pItem->_iAnimData;
+	auto *cel = pItem->AnimInfo.pCelSprite;
 	if (cel == nullptr) {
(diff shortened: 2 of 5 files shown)


obligaron:
After #1796 was merged this conflicts with the usage of `CelSprite`.
Currently `PlayerStruct` still uses old `uint8_t *` for Animation Data and `ItemStruct` uses new `CelSprite`.
But `AnimationInfo` should be used in both places.

Possible Solutions:

1. Convert `PlayerStruct` to `CelSprite`
2. Add temporary both `uint8_t *` and `CelSprite` to `AnimationInfo`
3. Revert `ItemStruct` to `uint8_t *`

I'm not a fan for going backwards. That's why 3 is not a good option.
Option 1 needs some thinking, cause we can't store the `unique_ptr` easily in `PlayerStruct` cause currently `PlayerStruct` is `memset`.
Option 2 is a little bit confusing to read. But this would only be a temporary solution anyway.

Note: This topic also arises if we advance the usage of `AnimationInfo` to monsters/missiles and if they didn't convert to `CelSprite` yet.

If the usage of `CelSprite` in `PlayerStruct` takes some time, I would go with the temporary option 2. This would also decouple `CelSprite` and `AnimationInfo` work.
@glebm What do you think is the best approach?

glebm:
@obligaron I think approach 1 is best. Avoiding `memset` will mean setting each field individually, it can be done by adding default field values `int myfield = 0` and then assigning the entire struct `plr = {}`.

obligaron:
I'm okay with this. 🙂 
Until `PlayerStruct` is converted to `CelSprite` I will change this PR to draft, so no one thinks it's ready to review/merged. 😉

AJenbo:
Thanks for the positive attitude, as you might have noticed we are in a phase of heavy restructuring atm :o

But from what I can see we are progressing nicely towards much better handling of the sprites, which hopefully will make your improvements easier (sans the rebasing).

Source/qol/itemlabels.cpp
```diff
@@ -74,7 +74,7 @@ void AddItemToLabelQueue(int id, int x, int y)
 	nameWidth += marginX * 2;
 	int index = ItemCAnimTbl[it->_iCurs];
 	if (!labelCenterOffsets[index]) {
-		std::pair<int, int> itemBounds = MeasureSolidHorizontalBounds(*it->_iAnimData, it->_iAnimFrame);
+		std::pair<int, int> itemBounds = MeasureSolidHorizontalBounds(*it->AnimInfo.pCelSprite, it->AnimInfo.CurrentFrame);
```
AJenbo, line 77: This should always be the last frame of the animation.

Source/scrollrt.cpp
```diff
@@ -640,7 +640,7 @@ static void DrawItem(const CelOutputBuffer &out, int x, int y, int sx, int sy, b
 		CelBlitOutlineTo(out, GetOutlineColor(*pItem, false), position, *cel, nCel);
 	}
 	CelClippedDrawLightTo(out, position, *cel, nCel);
-	if (pItem->_iAnimFrame == pItem->_iAnimLen || pItem->_iCurs == ICURS_MAGIC_ROCK)
+	if (pItem->AnimInfo.CurrentFrame == pItem->AnimInfo.NumberOfFrames  || pItem->_iCurs == ICURS_MAGIC_ROCK)
```
AJenbo, line 643:
```suggestion
	if (pItem->AnimInfo.CurrentFrame == pItem->AnimInfo.NumberOfFrames || pItem->_iCurs == ICURS_MAGIC_ROCK)
```

AJenbo (approved):

Source/scrollrt.cpp
obligaron: fixed 🙂

Source/qol/itemlabels.cpp
AJenbo: it is, and that's how it was, just would reduce the sprite scanning a bit.

PR #1796, opened 2021-05-01
♻️  Store CEL width with the sprite itself

glebm:
Instead of passing the CEL sprite width when drawing, store the CEL width at load time in the new `CelSprite` struct.

Implemented for most sprites except towners, missiles, or monsters.

### DIFF
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -14,9 +14,11 @@
 
 #include <algorithm>
 #include <cstddef>
-#include <cstdlib>
-#include <SDL.h>
 #include <cstdint>
+#include <cstdlib>
+#include <memory>
+
+#include <SDL.h>
 
 #ifdef USE_SDL1
 #include "utils/sdl2_to_1_2_backports.h"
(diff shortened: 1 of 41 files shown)


Source/engine.h
```diff
@@ -305,83 +312,134 @@ struct CelOutputBuffer {
...
+	CelSprite &operator=(CelSprite &&) noexcept = default;
+
+	[[nodiscard]] const BYTE *Data() const
+	{
+		return data_ptr_;
+	}
+
+	[[nodiscard]] int Width(std::size_t frame = 1) const
+	{
+		return widths_ == nullptr ? width_ : widths_[frame];
```
AJenbo, line 357: All frames have the same width, so i don't see the reason for this?

glebm: That's not true, see e.g. InvItemWidth

Source/items.cpp
```diff
@@ -2253,9 +2257,9 @@ void SetupItem(int i)
 	int it;
 
 	it = ItemCAnimTbl[items[i]._iCurs];
-	items[i]._iAnimData = itemanims[it];
+	items[i]._iAnimData = &*itemanims[it];
 	items[i]._iAnimLen = ItemAnimLs[it];
-	items[i]._iAnimWidth = 96;
+	items[i]._iAnimWidth = ItemAnimWidth;
```
AJenbo, line 2262: Isn't it possible to get this from _iAnimData now?

glebm: Maybe it is; I've kept this part of the refactoring to the minimum though

Source/engine.h
AJenbo: :(

i guess its only true for cl2?

Source/towners.cpp
```diff
@@ -374,14 +376,14 @@ void InitCows()
...
+	pCowCels = LoadFileInMem("Towners\\Animals\\Cow.CEL");
 	for (i = 0; i < 3; i++) {
 		x = TownCowX[i];
 		y = TownCowY[i];
 		dir = TownCowDir[i];
 		InitTownerInfo(numtowners, 128, false, TOWN_COW, x, y, -1);
-		towners[numtowners]._tNData = pCowCels;
-		SetTownerGPtrs(towners[numtowners]._tNData, towners[numtowners]._tNAnim);
+		towners[numtowners]._tNData = nullptr;
+		SetTownerGPtrs(pCowCels.get(), towners[numtowners]._tNAnim);
```
AJenbo, line 386: how come _tNData can now be null?

Source/engine.h
glebm: I'm not sure

Source/towners.cpp
glebm: `_tNData` is the "owned" data, it's not used directly other than for cleanups
All cows share the same data, not owned by individual cows

AJenbo: probably this needs some refactoring at some point

PR #1798, opened 2021-05-01
🐞 Support multiple playback of the same sound

### LINKED ISSUE #1794: Multiple sound effects at the same time do not play to the end
Multiple sound effects at the same time do not play to the end.

How to reproduce:
For example, sort the potions in the belt or drink from a fountain (Blood Fountain or Purifying Spring).
You will notice that the sounds produced by the mouse clicks do not play to the end.

Expected behavior:
Sound effects should play to the end like in vanilla.

glebm:
Fixes #1794


AJenbo (commented):
I guess we should then remove the rewind call as this fixes the issue properly https://github.com/diasurgical/devilutionX/blob/f884ba5340fbff8e14ff19dc5860cf022ab23740/Source/utils/soundsample.cpp#L54

AJenbo:
GCC 6.5 issue
```
/root/repo/build/_deps/fmt-src/include/fmt/format.h:903:29: note: you can work around this by removing the initializer
/root/repo/Source/sound.cpp: In function 'std::unique_ptr<devilution::TSnd> devilution::sound_file_load(const char*, bool)':
/root/repo/Source/sound.cpp:182:58: error: no matching function for call to 'devilution::SoundSample::SetChunk(std::remove_reference<std::unique_ptr<unsigned char [], std::default_delete<unsigned char []> >&>::type, devilution::DWORD&)'
   error = snd->DSB.SetChunk(std::move(wave_file), dwBytes);
                                                          ^

```

AJenbo:
Clang 12 issue:
```
/Applications/Xcode_12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/memory:3705:9: note: candidate template ignored: requirement 'is_lvalue_reference<std::__1::default_delete<unsigned char []> >::value' was not satisfied [with _Yp = unsigned char [], _Dp = std::__1::default_delete<unsigned char []>]
        shared_ptr(unique_ptr<_Yp, _Dp>&&,
        ^
```

glebm:
Compiler issues fixed, `rewind` call removed

Source/sound.cpp
```diff
@@ -68,6 +71,16 @@ void CleanupMusic()
 #endif
 }
 
+std::vector<std::unique_ptr<SoundSample>> duplicateSounds;
+SoundSample *DuplicateSound(const SoundSample &sound) {
+	auto duplicate = std::make_unique<SoundSample>();
+	if (duplicate->DuplicateFrom(sound) != 0)
```
julealgon, line 77: Isn't there a nicer way to achieve this using copy constructors? Having a method that copies seems like a code smell to me, as it allows for the instance to be used in the meantime (and potentially used incorrectly).

This to me sounds like temporal coupling, similar to having "Initialize" methods.

Source/sound.cpp
```diff
@@ -109,6 +122,19 @@ static int CapVolume(int volume)
...
 }
 
+void ClearDuplicateSounds() {
+	duplicateSounds.clear();
+}
+
+void CleanupFinishedDuplicateSounds()
+{
+	duplicateSounds.erase(
+	    std::remove_if(duplicateSounds.begin(), duplicateSounds.end(), [](const std::unique_ptr<SoundSample> &sound) {
```
julealgon, line 132: Can't we use `auto` for lambda arguments now?

Source/utils/stdcompat/shared_ptr_array.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <cstddef>
+#include <memory>
+
+namespace devilution {
+
+// Apple Clang 12 has a buggy implementation that fails to compile `std::shared_ptr<T[]>(new T[size])`.
+#if __cplusplus >= 201611L && (!defined(__clang_major__) || __clang_major__ >= 13)
+template <typename T>
+using SharedPtrArray = std::shared_ptr<T[]>;
```
julealgon, line 11: This name is incredibly misleading to me. IMHO, it indicates "an array of shared pointers", but it is a "shared pointer to an array".

I suggest it be renamed for clarity:
```suggestion
using ArraySharedPtr = std::shared_ptr<T[]>;
```

Source/utils/stdcompat/shared_ptr_array.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <cstddef>
+#include <memory>
+
+namespace devilution {
+
+// Apple Clang 12 has a buggy implementation that fails to compile `std::shared_ptr<T[]>(new T[size])`.
+#if __cplusplus >= 201611L && (!defined(__clang_major__) || __clang_major__ >= 13)
```
julealgon, line 9: Oh my god at this line.....

Source/sound.cpp
glebm: Good idea, will do

Source/sound.cpp
glebm: No, because we don't really have C++17 due to still supporting Amiga (and also GCC 5 for now)

Source/sound.cpp
glebm: Actually, this would require quite a bit of refactoring because the construction can fail (meaning we can't just use a copy constructor -- as we don't use exceptions). Leaving for later.

Source/utils/stdcompat/shared_ptr_array.hpp
glebm: Is this a constructive comment or should I just brush it off?

Source/utils/stdcompat/shared_ptr_array.hpp
glebm: Done

Source/sound.cpp
glebm: Actually turns out `auto` lambda arguments are supported even in GCC-5 :smile:

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/7bdfb0655c82d97a9802d850bda2ae25b16fc588

PR #1802, opened 2021-05-01
🧹 Fix more `-Wformat` warnings

glebm:


AJenbo:
Maybe we can change app_fatal to not do any formatting once fmt has been applied to the translation functions.

PR #1811, opened 2021-05-01
Update Italian localization

M0Rf30:
As in title.
Regards


AJenbo:
Nice, especially appreciate you taking care of all the strange contractions, even as a non-Italian speaker I could tell that it was pretty bad :D

PR #1812, opened 2021-05-01
♻️  Replace `SBmpLoadImage` with a simpler method

glebm:
The new approach only makes 2 read calls and does not do any seeks.

It is also less general and is internal to `art.cpp`, allowing us to simplify it somewhat.


AJenbo:
This was never meant to live as long as it did :+1:

PR #1813, opened 2021-05-01
🐞 Fix OOB strcpy in towners.cpp

glebm:
"Ogden el propietario de la taberna" is 35 bytes and did not fit in the previous 32-byte limit.

### DIFF
--- a/Source/towners.h
+++ b/Source/towners.h
@@ -15,6 +15,7 @@
 namespace devilution {
 
 #define NUM_TOWNERS 16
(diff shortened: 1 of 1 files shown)


AJenbo:
Od querido

AJenbo:
Isn't this going to happen a lot? I suppose the propper fix involves using `std::string` everywhere?

glebm:
There is a PR changing the towner name to `std::string_view`, which will also fix the issue.

PR #1814, opened 2021-05-01
🧹 `--lang-dir`: Document and minor cleanup

glebm:


AJenbo:
I have been wondering if this should be exposed or removed. Since they are our files, there isn't much of a point in having them in unexpected locations.

glebm:
That's a good point

PR #1815, opened 2021-05-01
🐞 items.cpp: Fix nullopt access

glebm:
A better fix would probably be to not call this function in the first place before the graphics are loaded. @thebigMuh


AJenbo (approved):

thebigMuh:
To be honest, I'd turn it around and instead initialize the item graphics at startup instead of when loading into the map. Because unless you just want to stare at the menu screen, at some point those will get loaded.

glebm:
@thebigMuh I agree that it would be a better fix, please feel free to give it a go

PR #1816, opened 2021-05-01
♻️  Remove `_iAnimWidth`

glebm:
No longer needed as of #1796

We could probably go further and not expose `ItemAnimWidth`,
but it'd be dangerous because sometimes `_iAnimData` is null.


glebm:
@thebigMuh Your fix to items would help here as well

PR #1820, opened 2021-05-02
Use smart pointers for SVid stuff

vlolteanu:


AJenbo (dismissed):

Source/utils/sdl_ptrs.h
```diff
@@ -11,28 +11,18 @@
...
- * @brief Deletes the SDL surface using `SDL_FreeSurface`.
- */
-struct SDLSurfaceDeleter {
-	void operator()(SDL_Surface *surface) const
-	{
-		SDL_FreeSurface(surface);
-	}
-};
-
-using SDLSurfaceUniquePtr = std::unique_ptr<SDL_Surface, SDLSurfaceDeleter>;
```
glebm, line 23: I much prefer separate names like before.

If you put it all in the same class, it couples things somewhat unnecessarily (always have to include the whole `<SDL.h>`) and it is unclear whether the correct deleter has been implemented without looking at the source.

Source/utils/sdl_ptrs.h
```diff
@@ -11,28 +11,18 @@
...
  */
 template <typename T>
 struct SDLFreeDeleter {
-	static_assert(!std::is_same<T, SDL_Surface>::value,
-	    "SDL_Surface should use SDLSurfaceUniquePtr instead.");
-
 	void operator()(T *obj) const
 	{
-		SDL_free(obj);
+		if constexpr (std::is_same<T, SDL_Surface>::value)
```
glebm, line 20: Please avoid `if constexpr` for now as we do not have it on GCC5 and GCC6.2 
You can use template specialization but as I'd prefer separate classes

Source/utils/sdl_ptrs.h
vlolteanu: I prefer the new approach; the old one was confusing. Looking at the SDL docs, there are only a handful  of SDL_Free* functions, and we can support them exhaustively. The API isn't going to change until SDL3.

As far as having to include all of SDL, you still need it if you want the asserts from the old SDLFreeDeleter.

PR #1821, opened 2021-05-02
🐞 Clean up and fix inventory sprite handling

glebm:
@thebigMuh Can you please test this or post a save I can test with?


Source/cursor.cpp
```diff
@@ -133,19 +140,34 @@ void FreeCursor()
...
 }
 
+const CelSprite &GetInvItemSprite(int i)
+{
+	return i < InvItems1Size ? *pCursCels : *pCursCels2;
+}
+
+int GetInvItemFrame(int i)
+{
+	return i < InvItems1Size ? i : i - InvItems1Size;
```
thebigMuh, line 150: This _used_ to be an adjustment of -= 179. Which would actually skip the first entry in the hellfire array portion?

AJenbo: Save game for testing https://github.com/diasurgical/devilutionX/files/5373942/hsv.zip

glebm: Changed and added a leading zero to the array.
The weird thing is that the frame number seems to start at 1 instead of 0

glebm: @AJenbo Thanks for the test file, seems to work :+1:

thebigMuh (approved):

Source/cursor.cpp
```diff
@@ -133,19 +142,34 @@ void FreeCursor()
...
 	icursW28 = icursW / 28;
 	icursH28 = icursH / 28;
 }
 
 void NewCursor(int i)
 {
 	pcurs = i;
-	cursW = InvItemWidth[i];
-	cursH = InvItemHeight[i];
+	std::tie(cursW, cursH) = GetInvItemSize(i);
```
AJenbo, line 172: We should probably consider making some sort of structure for the cursor, there is also cursx/y afaik.

AJenbo (approved):

PR #1823, opened 2021-05-02
💨 Do not zero-initalize arrays that we write to

glebm:
`std::make_unique<T[]>(size)` always zero-initalizes the array.

C++20 has `std::make_unique_for_overwrite` to avoid that, while older C++ versions can use the approach applied here.

### DIFF
--- a/Source/DiabloUI/art.cpp
+++ b/Source/DiabloUI/art.cpp
@@ -40,7 +40,7 @@ bool LoadPcxPixelsAndPalette(HANDLE handle, int width, int height, std::uint8_t 
 
 	// We read 1 extra byte because it delimits the palette.
 	const unsigned readSize = pixelDataSize + (has256ColorPalette ? PcxPaletteSize : 0);
-	std::unique_ptr<BYTE[]> fileBuffer = std::make_unique<BYTE[]>(readSize);
+	std::unique_ptr<BYTE[]> fileBuffer { new BYTE[readSize] };
 	if (!SFileReadFile(handle, fileBuffer.get(), readSize, nullptr, nullptr)) {
 		return false;
 	}
--- a/Source/capture.cpp
+++ b/Source/capture.cpp
@@ -116,7 +116,7 @@ static BYTE *CaptureEnc(BYTE *src, BYTE *dst, int width)
 static bool CapturePix(const CelOutputBuffer &buf, std::ofstream *out)
 {
 	int width = buf.w();
-	auto pBuffer = std::make_unique<BYTE[]>(2 * width);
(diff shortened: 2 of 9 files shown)


AJenbo (approved):

AJenbo:
@glebm @isage any idea why this is causing the Vita to build die?
```

ninja: job terminated due to signal 11: cd /root/repo/build && /home/user/vitasdk/bin/vita-elf-create /root/repo/build/devilutionx /root/repo/build/devilutionx.velf
vita-elf-create: Cannot allocate 2488 bytes for SCE data at end of segment 0; segment 1 overlaps
```

isage:
> @glebm @isage any idea why this is causing the Vita to build die?
> 
> ```
> 
> ninja: job terminated due to signal 11: cd /root/repo/build && /home/user/vitasdk/bin/vita-elf-create /root/repo/build/devilutionx /root/repo/build/devilutionx.velf
> vita-elf-create: Cannot allocate 2488 bytes for SCE data at end of segment 0; segment 1 overlaps
> ```

Honestly, no idea. Can you try building with/without LTO?

AJenbo:
Hmm, I guess never mind the very next build succeeded :shrug: Just odd that this commit failed twice in a row even after rebasing.

Source/DiabloUI/art.cpp
```diff
@@ -40,7 +40,7 @@ bool LoadPcxPixelsAndPalette(HANDLE handle, int width, int height, std::uint8_t
 
 	// We read 1 extra byte because it delimits the palette.
 	const unsigned readSize = pixelDataSize + (has256ColorPalette ? PcxPaletteSize : 0);
-	std::unique_ptr<BYTE[]> fileBuffer = std::make_unique<BYTE[]>(readSize);
+	std::unique_ptr<BYTE[]> fileBuffer { new BYTE[readSize] };
```
julealgon, line 43: Why though? What is the cost associated with the zero initialization? Is it enough to drive the code into an unusual initialization scheme? Every `unique_ptr` is initialized using `make_unique` and suddenly there are now cases where it isn't.

Is this really worth it?

glebm: Yes, sometimes there is a cost associated with zero initialized. We're filling buffers here, it's fairly low-level code.

glebm: If you'd like to make it nicer, you could backport `make_unique_for_overwrite` from C++20

glebm: There is an example implementation of it here:
https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

PR #1824, opened 2021-05-02
🧹 Remove `DiabloAllocPtr` and friends

glebm:


AJenbo:
🪦DiabloAllocPtr

Source/sound.cpp
```diff
@@ -43,7 +43,7 @@ void LoadMusic(HANDLE handle)
 	SDL_RWops *musicRw = SFileRw_FromStormHandle(handle);
 #else
 	int bytestoread = SFileGetFileSize(handle, 0);
-	musicBuffer = (char *)DiabloAllocPtr(bytestoread);
+	musicBuffer = new char[bytestoread];
```
AJenbo, line 46: if you need something todo then feel free to apply `std::byte` to things like we talked about :)

AJenbo (dismissed):

Source/sound.cpp
glebm: `std::byte` isn't available on GCC 5/6 :(

AJenbo (approved):

AJenbo:
No real need to rebase with no conflicts, or are you stress testing the CI :)

glebm:
@AJenbo I saw a message about the conflict with language.cpp on GitHub. Then I rebased and there wer no conflicts :thinking:

glebm:
Oh turns out I was looking at a different PR https://github.com/diasurgical/devilutionX/pull/1823

Source/player.cpp
```diff
@@ -490,31 +490,31 @@ void InitPlrGFXMem(int pnum)
...
-	player._pTData = DiabloAllocPtr(GetPlrGFXSize(c, "QM"));
+	player._pTData = new std::uint8_t[GetPlrGFXSize(c, "QM")];
 
 	// DEATH
-	player._pDData = DiabloAllocPtr(GetPlrGFXSize(c, "DT"));
+	player._pDData = new std::uint8_t[GetPlrGFXSize(c, "DT")];
 
 	// BLOCK
-	player._pBData = DiabloAllocPtr(GetPlrGFXSize(c, "BL"));
+	player._pBData = new std::uint8_t[GetPlrGFXSize(c, "BL")];
```
vlolteanu, line 517: I'd be partial to using std::unique_ptr. If/when this gets refactored into classes, we can use RAII to free them.

Source/sound.cpp
```diff
@@ -59,7 +59,7 @@ void CleanupMusic()
 		sgnMusicTrack = NUM_MUSIC;
 #ifdef DISABLE_STREAMING_MUSIC
 	if (musicBuffer != nullptr) {
-		mem_free_dbg(musicBuffer);
+		delete[] musicBuffer;
 		musicBuffer = nullptr;
```
vlolteanu, line 63: Ditto on the unique_ptr stuff.

Source/player.cpp
glebm: This struct is being memset to 0 so we can't use non-POD types in it yet

Source/sound.cpp
glebm: There isn't much difference with this one as we're still managing it manually

Source/sound.cpp
AJenbo: `#define std::byte std::uint8_t` :shrug: 

or we could `typedef` our own `byte` type based on compiler ability?.

The current situation is a bit of a confusing mess with `uint8_t`, `BYTE`, and `char` being used in various part of the code :( Don't feel any pressure though, if you don't to anything I will at some point :)

Source/player.cpp
vlolteanu: You can do in-place construction after memset.

glebm: memset on non-POD types is UB

Source/sound.cpp
vlolteanu: It's still cleaner.

Source/player.cpp
vlolteanu: You can use a a separate variable for the memset, then do in-place construction and return the actual struct.

vlolteanu: Although if it were me, it just initialize every one of those ints to 0.

vlolteanu: Further, the destructor would free the pointers for you.

Source/sound.cpp
AJenbo: I would find it a bit confusing

Source/player.cpp
glebm: I'd definitely prefer just adding member initializers to each field and then doing `plr = {}` instead of memset. That memset is not a hotspot and does not require the placement new dance.

This came up in #1795 as well; wanna take this one?

vlolteanu: I'm very busy this week, but if nobody does it the mean time, I'll do it.

PR #1827, opened 2021-05-02
Automap cleanups

AJenbo:


Source/appfat.cpp
```diff
@@ -29,7 +29,7 @@ void MsgBox(const char *pszFmt, va_list va)
 {
 	char text[256];
 
-	vsnprintf(text, 256, pszFmt, va);
+	vsnprintf(text, sizeof(text), pszFmt, va);
```
Jmgr, line 32: Why not use a `std::string` directly?

Jmgr: Ah, I suppose this would be out of scope for this PR?

Source/automap.cpp
```diff
@@ -291,114 +249,119 @@ void SearchAutomapItem(const CelOutputBuffer &out)
...
+	DrawLineTo(out, base, point, playerColor);
+	DrawLineTo(out, point, left, playerColor);
+	DrawLineTo(out, point, right, playerColor);
 }
 
 /**
  * @brief Returns the automap shape at the given coordinate.
  */
-uint16_t GetAutomapType(int x, int y, bool view)
+uint16_t GetAutomapType(Point Macro, bool view)
```
Jmgr, line 335: Why `Macro` and not `macro`? Also, what's a `macro`?

Source/automap.cpp
```diff
@@ -663,70 +625,69 @@ void DrawAutomap(const CelOutputBuffer &out)
 	DrawAutomapText(out);
 }
 
-void SetAutomapView(int x, int y)
+void SetAutomapView(Point tile)
 {
-	int xx = (x - 16) / 2;
-	int yy = (y - 16) / 2;
+	Point Macro { (tile.x - 16) / 2, (tile.y - 16) / 2 };
```
Jmgr, line 630: Again this `Macro`.

Source/engine.cpp
```diff
@@ -548,26 +549,26 @@ void CelBlitOutlineTo(const CelOutputBuffer &out, BYTE col, int sx, int sy, cons
...
 	float ix = dx / (float)steps;
 	float iy = dy / (float)steps;
-	float sx = x0;
-	float sy = y0;
+	float sx = a.x;
+	float sy = a.y;
 
 	for (int i = 0; i <= steps; i++, sx += ix, sy += iy) {
-		SetPixel(out, sx, sy, color_index);
+		SetPixel(out, { (int)sx, (int)sy }, color_index);
```
Jmgr, line 571: C-style cast

Source/engine.h
```diff
@@ -515,11 +515,10 @@ void CelBlitOutlineTo(const CelOutputBuffer &out, BYTE col, int sx, int sy, cons
 /**
  * @brief Set the value of a single pixel in the back buffer, checks bounds
  * @param out Target buffer
- * @param sx Target buffer coordinate
- * @param sy Target buffer coordinate
+ * @param point Target buffer coordinate
  * @param col Color index from current palette
  */
-void SetPixel(const CelOutputBuffer &out, int sx, int sy, BYTE col);
+void SetPixel(const CelOutputBuffer &out, Point point, BYTE col);
```
Jmgr, line 521: `Point position`?

Source/qol/xpbar.cpp
```diff
@@ -34,11 +34,11 @@ void DrawBar(const CelOutputBuffer &out, int x, int y, int width, const ColorGra
 	FastDrawHorizLine(out, x, y + 3, width, gradient[gradient.size() / 2 - 1]);
 }
 
-void DrawEndCap(const CelOutputBuffer &out, int x, int y, int idx, const ColorGradient &gradient)
+void DrawEndCap(const CelOutputBuffer &out, Point point, int idx, const ColorGradient &gradient)
```
Jmgr, line 37: I would pass `Point` as a const reference, since its size is > than sizeof(int).

Source/qol/xpbar.cpp
```diff
@@ -101,7 +101,7 @@ void DrawXPBar(const CelOutputBuffer &out)
 	DrawBar(out, xPos, yPos, fullBar, SILVER_GRADIENT);
 
 	// End pixels appear gradually
-	DrawEndCap(out, xPos + fullBar, yPos, fade, SILVER_GRADIENT);
+	DrawEndCap(out, { xPos + (int)fullBar, yPos }, fade, SILVER_GRADIENT);
```
Jmgr, line 104: C-style cast

Source/appfat.cpp
AJenbo: Just did it quickly to avoid some warnings when trying out CLion :)

Hopefully, this makes some tools better able to reason about this code, and maybe it will even make refactoring easier :o

Source/automap.cpp
AJenbo: macro tiles are tiles constructed out of 2x2 tile and is the scale that the automap operates on.

AJenbo: Renamed it to map, that should make more sens maybe.

Source/engine.cpp
AJenbo: Who can remember how to write the C++ style ones any way :)

Source/automap.cpp
Jmgr: `Point map`? So it's not a position? I'm a bit confused.

Jmgr: Ah so it's the coordinates of a macro?

Source/engine.cpp
Jmgr: I usually go with `static_cast` first as it's the safest. Then `reinterpret_cast`/`const_cast`. I almost never use `dynamic_cast`. (I also often write `static_cats`)

AJenbo: My older cat is pretty static at this point...

Source/automap.cpp
AJenbo: > `Point map`? So it's not a position? I'm a bit confused.

Being an x,y struct makes it automatically a position in my head. During usage (`map.x, map.y`) it makes sense imo?

Jmgr: Hm. A `Point` is not always a position. I don't find it obvious, but it's subjective. I tend to prefer longer and more explicit variables, like `mapPosition` or `mapPos`.

AJenbo: The automap context helps a bit. But yeah defiantly subjective.

AJenbo: But should objectivly be a step up from just `x, y` I think :)

Source/automap.h
```diff
@@ -79,7 +78,7 @@ void DrawAutomap(const CelOutputBuffer &out);
 /**
  * @brief Marks the given coordinate as within view on the automap.
  */
-void SetAutomapView(int x, int y);
+void SetAutomapView(const Point tile);
```
glebm, line 81: no need for `const` when passing by value generally -- it is up to the function what to do with the copied argument

Jmgr: This should be a const reference I think, not a const.

glebm: Well the general guidance is to pass tiny objects by value

Jmgr: Yes, well for me anything > sizeof(int) (in this case, two ints) is not tiny. But not strong feeling about this.

AJenbo: On 64bit systems `sizeof(int) * 2` is probably a more reasonable measure as anything smaller takes less than the address for it would.

Jmgr: Not sure I understand, on 64 bit systems `sizeof(int) == 64`, which is the same as `sizeof(void*)`. But yeah the common rule is that anything bigger than a pointer should be passed by const references. Or `sizeof(Point) == sizeof(void*) * 2` on any system.

AJenbo: You probably mean `sizeof(int) == 8`, but at least on my system (Linux x86_64 GCC) `sizeof(int) == 4`.

Passes:
`static_assert(sizeof(Point) == sizeof(void*));`
Fails:
`static_assert(sizeof(Point) == sizeof(void*) * 2);`

Jmgr: > You probably mean `sizeof(int) == 8`, but at least on my system (Linux x86_64 GCC) `sizeof(int) == 4`.

Ah yes sorry, I confused `long` with `int`.

PR #1830, opened 2021-05-02
🔨 CMake: Add ThreadSanitizer via `-DTSAN=ON -DASAN=OFF`

glebm:
It immediately detects some data races


AJenbo:
Well that's good, as you where able to fix it :)

AJenbo (approved):

glebm:
There are more thread safety issues, this only fixes some of them.

AJenbo:
I get that :)

It usually takes a few days to even trigger all the issues, let alone weed all of them out,

PR #1834, opened 2021-05-02
Move 3 of the Linux Builds to GitHub actions

Trihedraf:


.github/workflows/Linux_x86_64_SDL1.yml
```diff
@@ -0,0 +1,40 @@
+name: Linux_x86_64_SDL1
+
+on: [push, pull_request]
```
AJenbo, line 3: Is there a way where we can have it only trigger once for the PR and once for the merged commit in master?

The current rule makes it trigger twice on PR and once for master.

.github/workflows/Linux_x86.yml
```diff
@@ -0,0 +1,46 @@
...
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Create Build Environment
+      run: sudo dpkg --add-architecture i386 && sudo apt-get update -y && sudo apt-get install -y cmake file g++-multilib git libfmt-dev:i386 libsdl2-dev:i386 libsdl2-ttf-dev:i386 libsodium-dev:i386 rpm wget
```
glebm, line 15:
```suggestion
      run: >
        sudo dpkg --add-architecture i386 &&
        sudo apt-get update -y &&
        sudo apt-get install -y cmake file g++-multilib git libfmt-dev:i386 libsdl2-dev:i386 libsdl2-ttf-dev:i386 libsodium-dev:i386 rpm wget
```

.github/workflows/Linux_x86_64_SDL1.yml
Trihedraf: I just copied the one from the Mac build. It was like that. And when I looked at the pr from when it was merged in and you had him change it to that so that pushes and pr's would get builds

AJenbo: yeah i know, it was not ideal, but better then when it only checked after merging. Running to builds instead of one inst a huge deal, but now it would result in 8 instead of 4 which will probably cause things to queue up so we should fix it now :)

.github/workflows/Linux_x86_64.yml
```diff
@@ -0,0 +1,48 @@
+name: Linux_x86_64
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
```
AJenbo, line 7: its important that linux_x86_64 is build on debian:stretch-backports as that provides libs old enough give the releases broad comparability across the Linux ecosystem. I only just now realize that you moved it to debian:testing on CircleCI already, but we will need to revert that. Sorry for not realizing this during the previous pr.

Trihedraf: Only the x64, not the x86 or sdl1?

glebm: Perhaps that should come as a separate PR following this one?

AJenbo: yeah my only concern there is that its easier to revert as is...

AJenbo: > Only the x64, not the x86 or sdl1?

linux_x86_64 is the only one that we produce releases for, x86 and sdl1 are more for self build targets or crossbuild from modern compilers, so its fine to run them on recent distros.

Trihedraf: #1846

Trihedraf: I just tried to go back to debian:stretch-backports, but libfmt-dev doesn't exist in the debian stable or stretch-backports repos.

glebm: That shouldn't be an issue, we build fmt from source automatically if it doesn't exist

PR #1838, opened 2021-05-02
🧹 CMake: Try removing freetype and png

glebm:
These SDL2_ttf dependencies should be picked up automatically by pkg-config


glebm:
Doesn't work :(

AJenbo:
im kinda hoping that we can remove SDL_ttf in 1.3.0 :D

Issue #1833, opened 2021-05-02
TSAN: Data race in SoundSample


glebm:
```
WARNING: ThreadSanitizer: data race (pid=709102)
  Write of size 8 at 0x7b9000420000 by main thread:
    #0 operator delete[](void*) ../../../../src/libsanitizer/tsan/tsan_new_delete.cpp:132 (libtsan.so.0+0x8e8f8)
    #1 std::enable_if<std::is_convertible<float (*) [], float (*) []>::value, void>::type std::default_delete<float []>::operator()<float>(float*) const /usr/include/c++/10/bits/unique_ptr.h:120 (devilutionx+0xcc6d7a)
    #2 std::unique_ptr<Aulib::Resampler, std::default_delete<Aulib::Resampler> >::~unique_ptr() /usr/include/c++/10/bits/unique_ptr.h:361 (devilutionx+0xc7062c)
    #3 Aulib::Stream_priv::~Stream_priv() /home/gleb/repos/glebm/SDL_audiolib/src/stream_p.cpp:38 (devilutionx+0xccfa6c)
    #4 std::__uniq_ptr_impl<Aulib::Stream, std::default_delete<Aulib::Stream> >::reset(Aulib::Stream*) /usr/include/c++/10/bits/unique_ptr.h:182 (devilutionx+0xca0dd0)
    #5 std::unique_ptr<Aulib::Stream, std::default_delete<Aulib::Stream> >::reset(Aulib::Stream*) /usr/include/c++/10/bits/unique_ptr.h:456 (devilutionx+0xca08be)
    #6 std::unique_ptr<Aulib::Stream, std::default_delete<Aulib::Stream> >::operator=(decltype(nullptr)) /usr/include/c++/10/bits/unique_ptr.h:397 (devilutionx+0xc9fa4c)
    #7 devilution::SoundSample::Release() ../Source/utils/soundsample.cpp:25 (devilutionx+0xc9d4b4)
    #8 devilution::TSnd::~TSnd() ../Source/sound.cpp:191 (devilutionx+0xc86bb0)
    #9 std::default_delete<devilution::TSnd>::operator()(devilution::TSnd*) const /usr/include/c++/10/bits/unique_ptr.h:85 (devilutionx+0x927be9)
[...]

PR #1840, opened 2021-05-02
🐞 Synchronize SFileReadFile access

glebm:
StormLib read function is not thread-safe: https://github.com/ladislav-zezula/StormLib/issues/175

This fixes static in the streaming audio on the 3DS, and probably everywhere else

Thanks @StephenCWills for the investigation!

### DIFF
--- a/CMake/ctr/n3ds_defs.cmake
+++ b/CMake/ctr/n3ds_defs.cmake
@@ -2,11 +2,6 @@
 set(NONET ON)
 set(USE_SDL1 ON)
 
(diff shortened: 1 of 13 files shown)


AJenbo (approved):
really exciting, im guessing there is also a chance this was the cause of the audio drops on windows

AJenbo:
should we fully drop the option not to stream?

glebm:
> should we fully drop the option not to stream?

Let's wait a few more weeks because there's still #1833

PR #1841, opened 2021-05-02
3DS improvements

StephenCWills:
* Fixes the TTF font path for the cia package so the ttf file does not need to be deployed separately.
* Improves performance by setting bpp to 8.
* Reduces the SystemMode setting to 80MB so the cia-installed version will run on the old 3DS. At 96MB, the game simply crashes on startup.


CMake/ctr/n3ds_defs.cmake
```diff
@@ -20,9 +20,10 @@ find_package(PNG REQUIRED)
```
glebm, line 20: Can also remove these:

```
find_package(Tremor REQUIRED)
find_package(OGG REQUIRED)
find_package(MIKMOD REQUIRED)
find_package(MAD REQUIRED)
```

glebm (approved):

AJenbo (approved):

PR #1846, opened 2021-05-03
More CircleCi <-> GitHub Actions updates

Trihedraf:


AJenbo:
Nice this looks like a good solution to the problem with building the x86_64 linux version on an older distro as well as balancing the build services.

PR #1848, opened 2021-05-03
🎉 New map renderer

glebm:
Uses integer math only: This speeds up the rendering and eliminates some zoom artifacts.

Improves player indicator look -- it's now symmetric and more legible.

FPS on my desktop (gains will be larger on devices with slow FPU):

* 640x480: 1390 -> 1410
* 1200x800: 432 -> 435

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,7 +272,6 @@ add_library(PKWare STATIC
 set(devilutionx_SRCS
   Source/appfat.cpp
   Source/automap.cpp
-  Source/automap_render.cpp
   Source/capture.cpp
   Source/codec.cpp
   Source/control.cpp
@@ -352,6 +351,7 @@ set(devilutionx_SRCS
   Source/controls/touch.cpp
   Source/controls/keymapper.cpp
   Source/engine/animationinfo.cpp
+  Source/engine/render/automap_render.cpp
   Source/qol/autopickup.cpp
   Source/qol/common.cpp
   Source/qol/monhealthbar.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,6 +272,7 @@ add_library(PKWare STATIC
 set(devilutionx_SRCS
   Source/appfat.cpp
   Source/automap.cpp
+  Source/automap_render.cpp
   Source/capture.cpp
   Source/codec.cpp
   Source/control.cpp
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -7,6 +7,7 @@
 
 #include <algorithm>
 
+#include "automap_render.hpp"
 #include "control.h"
 #include "inv.h"
 #include "miniwin/miniwin.h"
@@ -53,35 +54,30 @@ enum MapFlags : uint8_t {
 	// clang-format on
 };
 
-void DrawSquare(const CelOutputBuffer &out, Point center, uint8_t color)
+void DrawDiamond(const CelOutputBuffer &out, Point center, uint8_t color)
 {
 	const Point left { center.x - AmLine16, center.y };
 	const Point top { center.x, center.y - AmLine8 };
-	const Point right { center.x + AmLine16, center.y };
 	const Point bottom { center.x, center.y + AmLine8 };
 
-	DrawLineTo(out, top, left, color);
-	DrawLineTo(out, top, right, color);
-	DrawLineTo(out, bottom, left, color);
-	DrawLineTo(out, bottom, right, color);
+	DrawMapLineNE(out, left, AmLine8, color);
+	DrawMapLineSE(out, left, AmLine8, color);
+	DrawMapLineSE(out, top, AmLine8, color);
+	DrawMapLineNE(out, bottom, AmLine8, color);
 }
 
 void DrawMapVerticalDoor(const CelOutputBuffer &out, Point center)
 {
-	const Point offset { center.x - AmLine16, center.y - AmLine8 };
-
-	DrawLineTo(out, { center.x + AmLine16, offset.y }, { center.x + AmLine8, offset.y + AmLine4 }, MapColorsDim);
-	DrawLineTo(out, { offset.x, center.y + AmLine8 }, { offset.x + AmLine8, center.y + AmLine8 - AmLine4 }, MapColorsDim);
-	DrawSquare(out, center, MapColorsBright);
+	DrawMapLineNE(out, { center.x + AmLine8, center.y - AmLine4 }, AmLine4, MapColorsDim);
+	DrawMapLineNE(out, { center.x - AmLine16, center.y + AmLine8 }, AmLine4, MapColorsDim);
+	DrawDiamond(out, center, MapColorsBright);
 }
 
 void DrawMapHorizontalDoor(const CelOutputBuffer &out, Point center)
 {
-	const Point offset { center.x + AmLine16, center.y - AmLine8 };
-
-	DrawLineTo(out, { center.x - AmLine16, offset.y }, { center.x - AmLine16 + AmLine8, offset.y + AmLine4 }, MapColorsDim);
-	DrawLineTo(out, { offset.x, center.y + AmLine8 }, { offset.x - AmLine8, center.y + AmLine8 - AmLine4 }, MapColorsDim);
-	DrawSquare(out, center, MapColorsBright);
+	DrawMapLineSE(out, { center.x - AmLine16, center.y - AmLine8 }, AmLine4, MapColorsDim);
+	DrawMapLineSE(out, { center.x + AmLine8, center.y + AmLine4 }, AmLine4, MapColorsDim);
+	DrawDiamond(out, center, MapColorsBright);
 }
 
 /**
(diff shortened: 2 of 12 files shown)


AJenbo:
Feel free to try and implement fancy stairs if you feel like it :)

image

glebm:
These will be straightforward to implement but I'm not interested in implementing more bells and whistles atm :)

qndel:
Magnificent!

AJenbo:
image

qndel:
wobble
maybe it's a good moment to fix the wobbly mouse when walking southwest/southeast ? :P

glebm:
@qndel The wobbly mouse would be a separate fix, it's not related to how things are rendered but rather to where they're render. Thanks for the animation!

Source/engine.cpp
```diff
@@ -557,18 +558,44 @@ void SetPixel(const CelOutputBuffer &out, Point position, BYTE col)
...
+	}
+	if (from.x + width > out.w())
+		width = (from.x + width) - out.w();
+	return UnsafeDrawHorizontalLine(out, from, width, colorIndex);
+}
+
+void UnsafeDrawHorizontalLine(const CelOutputBuffer &out, Point from, int width, std::uint8_t colorIndex)
+{
+	std::memset(out.at(from.x, from.y), colorIndex, width);
+}
```
thebigMuh, line 577: There's already pretty much the same code in the QoL features (in common.h/cpp I believe).

Maybe farm out both of these into a common location?

julealgon:
Can you provide some before/after comparison via screenshot and/or gif @glebm ?

Source/qol/common.cpp
```diff
@@ -25,16 +25,12 @@ int GetTextWidth(const char *s)
...
 }
 
 void FastDrawVertLine(const CelOutputBuffer &out, int x, int y, int height, Uint8 col)
 {
-	BYTE *p = out.at(x, y);
-	for (int j = 0; j < height; j++) {
-		*p = col;
-		p += out.pitch();
-	}
+	UnsafeDrawVerticalLine(out, {x, y}, height, col);
```
thebigMuh, line 33: I think you can kill that and just directly use the UnsafeDraw* functions.

glebm: Yeah I'll do it at the very end, don't want to have conflicts in case QoL stuff gets changed in the meantime

glebm:
@julealgon Qndel's gif is "after"

julealgon:
@glebm 
> @julealgon Qndel's gif is the "after"

Having just the after, it's hard to tell what you are doing. Qndel's gif is also of really bad quality and is filtering the graphics to a blurry mess.

Would be nice if we had a proper unfiltered comparison.

FitzRoyX:
The one on the left is the old one. The other two are the new ones, first is up/down and last is left/right.

arrow

glebm:
Ready for review

CMakeLists.txt
```diff
@@ -272,6 +272,7 @@ add_library(PKWare STATIC
 set(devilutionx_SRCS
   Source/appfat.cpp
   Source/automap.cpp
+  Source/automap_render.cpp
```
AJenbo, line 275: Should this live in `engine/` with the game world render?

Source/automap.cpp
```diff
@@ -111,10 +107,13 @@ void DrawAutomapTile(const CelOutputBuffer &out, Point center, uint16_t automapT
...
-		DrawLineTo(out, { center.x - AmLine8, center.y - AmLine8 - AmLine4 }, { center.x + AmLine8 + AmLine16, center.y + AmLine4 }, MapColorsBright);
-		DrawLineTo(out, { center.x - AmLine16, center.y - AmLine8 }, { center.x + AmLine16, center.y + AmLine8 }, MapColorsBright);
-		DrawLineTo(out, { center.x - AmLine16 - AmLine8, center.y - AmLine4 }, { center.x + AmLine8, center.y + AmLine8 + AmLine4 }, MapColorsBright);
-		DrawLineTo(out, { center.x - AmLine32, center.y }, { center.x, center.y + AmLine16 }, MapColorsBright);
+		constexpr int NumStairSteps = 4;
+		const Point offset = { -AmLine8, AmLine4 };
+		Point p = { center.x - AmLine8 , center.y - AmLine8 - AmLine4 };
+		for (int i = 0; i < NumStairSteps; ++i) {
+			DrawMapLineSE(out, p, AmLine16, MapColorsBright);
+			p += offset;
```
AJenbo, line 115: Good to see otheres take advantage of the Point type :)

CMakeLists.txt
glebm: I'm not sure what `engine.h` is currently -- it seems to do quite a few things besides rendering.

Can I leave the location up to you for a follow-up refactoring? :)

AJenbo: not engine.h, engine/ (the folder)

Source/automap_render.cpp
```diff
@@ -0,0 +1,84 @@
...
+		from.x += static_cast<int>(DirX);
+		SetPixel(out, from, colorIndex);
+		from.x += static_cast<int>(DirX);
+		from.y += static_cast<int>(DirY);
+	}
+	SetPixel(out, from, colorIndex);
+}
+
+template <DirectionX DirX, DirectionY DirY>
+void DrawMapLine2(const CelOutputBuffer &out, Point from, int width, std::uint8_t colorIndex)
```
AJenbo, line 30: its a bit hard to understand what the difference between the regular and `2` functions are. from looking here im guessing its because you need to alternate at have lines line up when drawing them in blocks?

CMakeLists.txt
AJenbo: also it dosnt seam to be strictly tied to thee auto map, so how about `Source/engine/render/lines.cpp`

AJenbo (commented):
looks good, i have made a more concrete suggestion regarding the file name of the new file (src root is messy enough as is), hope that works for naming.

if you could also add a comment on when one would use DrawMapLine2 over DrawMapLine that would be apriciated.

Source/automap_render.hpp
```diff
@@ -0,0 +1,77 @@
...
+void DrawMapLineNW2(const CelOutputBuffer &out, Point from, int width, std::uint8_t colorIndex);
+
+/**
+ * @brief Draw a line in the target buffer from the given point towards south west at an `atan(1/2)` angle.
+ *
+ * Draws 2 vertical pixels for each horizontal step, then an additional one where it draws 1 pixel.
+ *
+ * The end point is at `{ from.x - (width + 1), from.y + 2 * width }`.
+ */
+void DrawMapLineSW2(const CelOutputBuffer &out, Point from, int width, std::uint8_t colorIndex);
```
AJenbo, line 75: ok, looks like the `2` functions are limited by `width` where he others are limited by `height`. though deceptively `2` is actually `width + 1`?

Source/automap_render.cpp
thebigMuh: The '2' function increments y twice per x increment, and the 'plain' function increments x twice per y increment. 

Maybe call them "Steep" and "Flat" or "Shallow"?

galaxyhaxz:
> Uses integer math only: This speeds up the rendering and eliminates some zoom artifacts.

Just curious about performance. Have you tried comparing performance of having multiple custom line functions vs simply having the single DDA algorithm with integer math? I use the DDA bit shift in my project and it renders the automap 2.5x faster than using floating point math.

Source/automap_render.hpp
glebm: These functions are automap-specific, they're not general line-drawing functions.
The extra "tip" pixel is needed to reproduce the results we had with the float renderer.

The two sets of functions accept the smaller dimension, this makes the API easier to use for map drawing.

CMakeLists.txt
glebm: It is strictly tied to the automap; Why have `engine/render` and not just `render/`? Do we render anything outside of the "engine" (whatever the "engine" means)?

glebm:
> Have you tried comparing performance of having multiple custom line functions vs simply having the single DDA algorithm with integer math?

I have not compared the performance of just the line rendering functions but I expect the custom line functions to be significantly faster than DDA

Source/automap_render.cpp
glebm: Done, the "2" functions are now "Steep", thanks for the suggestion!

glebm:
> if you could also add a comment on when one would use DrawMapLine2 over DrawMapLine that would be apriciated.

Added file-level documentation to the header

CMakeLists.txt
AJenbo: > The core functionality typically provided by a game engine may include a rendering engine ("renderer") for 2D or 3D graphics, a physics engine or collision detection (and collision response), sound, scripting, animation, artificial intelligence, networking, streaming, memory management, threading, localization support, scene graph, and video support for cinematics.

Basically anything that doesn't distinguish the game itself.

We could have all of this in the base Source folder, but it's a way for us to segregate the "complicated stuff" that some one just looking to modify the game-play probably has no need to look at, and which is allowed to be a bit more complicated then the rest of the code-base.

AJenbo (approved):

galaxyhaxz:
> I have not compared the performance of just the line rendering functions but I expect the custom line functions to be significantly faster than DDA

In this case since the lines are so small, and predetermined angles in 8 directions, I think this is the case.

http://www.edepot.com/algorithm.html
Using DDA with integer math and an unrolled loop beats even the fastest line algorithm here which is already faster than Diablo's original algorithm and claimed to hold a record lol.

PR #1853, opened 2021-05-03
♻️Typed file loading

AJenbo:
Tthis gives us the option to specify what type a file should be loaded
as, avoidng the need to case it and does some automatic checks on the
fitness of the data, while making the process simpler.

If no type is given then the type will be set to std::byte which limit
what operations can be performed on the data.

### DIFF
--- a/Source/drlg_l2.cpp
+++ b/Source/drlg_l2.cpp
@@ -1843,15 +1843,15 @@ static void DRLG_LoadL2SP()
 	setloadflag = false;
 
 	if (QuestStatus(Q_BLIND)) {
-		pSetPiece = LoadFileInMem("Levels\\L2Data\\Blind1.DUN");
-		pSetPiece[26] = 154;  // Close outer wall
-		pSetPiece[200] = 154; // Close outer wall
+		pSetPiece = LoadFileInMem<uint16_t>("Levels\\L2Data\\Blind1.DUN");
+		pSetPiece[13] = SDL_SwapLE16(154);  // Close outer wall
+		pSetPiece[100] = SDL_SwapLE16(154); // Close outer wall
 		setloadflag = true;
 	} else if (QuestStatus(Q_BLOOD)) {
-		pSetPiece = LoadFileInMem("Levels\\L2Data\\Blood1.DUN");
+		pSetPiece = LoadFileInMem<uint16_t>("Levels\\L2Data\\Blood1.DUN");
 		setloadflag = true;
 	} else if (QuestStatus(Q_SCHAMB)) {
-		pSetPiece = LoadFileInMem("Levels\\L2Data\\Bonestr2.DUN");
+		pSetPiece = LoadFileInMem<uint16_t>("Levels\\L2Data\\Bonestr2.DUN");
 		setloadflag = true;
 	}
 }
@@ -1863,28 +1863,25 @@ static void DRLG_FreeL2SP()
 
 static void DRLG_L2SetRoom(int rx1, int ry1)
 {
-	int rw, rh, i, j;
-	BYTE *sp;
-
-	rw = pSetPiece[0];
-	rh = pSetPiece[2];
+	int width = SDL_SwapLE16(pSetPiece[0]);
+	int height = SDL_SwapLE16(pSetPiece[1]);
 
 	setpc_x = rx1;
 	setpc_y = ry1;
-	setpc_w = rw;
-	setpc_h = rh;
+	setpc_w = width;
+	setpc_h = height;
 
-	sp = &pSetPiece[4];
+	uint16_t *tileLayer = &pSetPiece[2];
 
-	for (j = 0; j < rh; j++) {
-		for (i = 0; i < rw; i++) {
-			if (*sp != 0) {
-				dungeon[i + rx1][j + ry1] = *sp;
+	for (int j = 0; j < height; j++) {
+		for (int i = 0; i < width; i++) {
+			uint8_t tileId = SDL_SwapLE16(tileLayer[j * width + i]);
+			if (tileId != 0) {
+				dungeon[i + rx1][j + ry1] = tileId;
 				dflags[i + rx1][j + ry1] |= DLRG_PROTECTED;
 			} else {
 				dungeon[i + rx1][j + ry1] = 3;
 			}
-			sp += 2;
 		}
 	}
 }
@@ -3227,7 +3224,7 @@ static void DRLG_InitL2Vals()
 	}
 }
 
-static void LoadL2DungeonData(BYTE *pLevelMap)
+static void LoadL2DungeonData(const uint16_t *dunData)
 {
 	InitDungeon();
 	DRLG_InitTrans();
@@ -3239,21 +3236,21 @@ static void LoadL2DungeonData(BYTE *pLevelMap)
 		}
 	}
 
-	BYTE *lm = pLevelMap;
-	int rw = *lm;
-	lm += 2;
-	int rh = *lm;
-	lm += 2;
+	int width = SDL_SwapLE16(dunData[0]);
+	int height = SDL_SwapLE16(dunData[1]);
 
-	for (int j = 0; j < rh; j++) {
-		for (int i = 0; i < rw; i++) {
-			if (*lm != 0) {
-				dungeon[i][j] = *lm;
+	const uint16_t *tileLayer = &dunData[2];
+
+	for (int j = 0; j < height; j++) {
+		for (int i = 0; i < width; i++) {
+			uint8_t tileId = SDL_SwapLE16(*tileLayer);
+			tileLayer++;
+			if (tileId != 0) {
+				dungeon[i][j] = tileId;
 				dflags[i][j] |= DLRG_PROTECTED;
 			} else {
 				dungeon[i][j] = 3;
 			}
-			lm += 2;
 		}
 	}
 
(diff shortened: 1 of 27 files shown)


Source/drlg_l2.cpp
```diff
@@ -1843,15 +1843,15 @@ static void DRLG_LoadL2SP()
 	setloadflag = false;
 
 	if (QuestStatus(Q_BLIND)) {
-		pSetPiece = LoadFileInMem("Levels\\L2Data\\Blind1.DUN");
-		pSetPiece[26] = 154;  // Close outer wall
-		pSetPiece[200] = 154; // Close outer wall
+		pSetPiece = LoadFileInMem<uint16_t>("Levels\\L2Data\\Blind1.DUN");
+		pSetPiece[13] = 154;  // Close outer wall
+		pSetPiece[100] = 154; // Close outer wall
```
AJenbo, line 1848: Need to use LoadLE32()

Source/engine.h
```diff
@@ -598,7 +604,40 @@ void SetRndSeed(int32_t s);
...
+
+	if ((fileLen % sizeof(T)) != 0)
+		app_fatal("File size does not align with type\n%s", path);
+
+	if (elements != nullptr)
+		*elements = fileLen / sizeof(T);
+
+	std::unique_ptr<T[]> buf { new T[fileLen / sizeof(T)] };
+
+	LoadFileData(path, (byte *)buf.get(), fileLen);
```
glebm, line 636:
```suggestion
	LoadFileData(path, reinterpret_cast<byte *>(buf.get()), fileLen);
```

Source/engine.h
```diff
@@ -598,7 +604,40 @@ void SetRndSeed(int32_t s);
...
 int32_t GenerateRnd(int32_t v);
-std::unique_ptr<BYTE[]> LoadFileInMem(const char *pszName, DWORD *pdwFileLen = nullptr);
+
+size_t GetFileSize(const char *pszName);
+void LoadFileData(const char *pszName, byte *buffer, size_t fileLen);
+
+template <typename T, std::size_t N>
+void LoadFileInMem(const char *path, std::array<T, N> &data)
+{
+	LoadFileData(path, (byte *)&data, N * sizeof(T));
```
glebm, line 614:
```suggestion
	LoadFileData(path, reinterpret_cast<byte *>(&data), N * sizeof(T));
```

Source/engine.h
```diff
@@ -598,7 +604,40 @@ void SetRndSeed(int32_t s);
 int32_t AdvanceRndSeed();
 int32_t GetRndSeed();
 int32_t GenerateRnd(int32_t v);
-std::unique_ptr<BYTE[]> LoadFileInMem(const char *pszName, DWORD *pdwFileLen = nullptr);
+
+size_t GetFileSize(const char *pszName);
+void LoadFileData(const char *pszName, byte *buffer, size_t fileLen);
```
glebm, line 609: This `fileLen` is really `bufferSize` in the current implementation

glebm (dismissed):

Source/engine.h
AJenbo: I went with c-style casting in the places where it needs to be cleaned up and further propagated, just so that it's obvious that it needs to be looked at later.

AJenbo: .... but this isn't such a case :D

Source/quests.cpp
```diff
@@ -378,137 +378,118 @@ void DrawSkelKing(int q, int x, int y)
...
-			dungeon[i][j] = v;
-			sp += 2;
+	setpc_w = width;
+	setpc_h = height;
+
+	uint16_t *tileLayer = &dunData[2];
+
+	for (int j = y; j < y + height; j++) {
+		for (int i = x; i < x + width; i++) {
+			uint8_t tileId = SDL_SwapLE16(tileLayer[j * width + i]);
```
thebigMuh, line 395: There's a huge number of SDL_SwapLE16 calls, which return a UInt16, which then get assigned to uint8_t.

Looking at the old code, this should work, but if you raise the warning level those will all spam type narrowing warnings like crazy.

Since this seems to be such a common operation, maybe add a function or macro that does the 'read uint16 -> swap endian -> cast to uint8' in one step, and add a comment why this is valid and needed. tile_cast or something similar.

Source/palette.cpp
```diff
@@ -120,20 +120,22 @@ static void GenerateBlendedLookupTable(SDL_Color *palette, int skipFrom, int ski
...
 	assert(pszFileName);
 
-	SFileOpenFile(pszFileName, &pBuf);
-	SFileReadFileThreadSafe(pBuf, (char *)PalData, sizeof(PalData));
-	SFileCloseFile(pBuf);
+	struct Color {
+		uint8_t r;
+		uint8_t g;
+		uint8_t b;
+	};
```
thebigMuh, line 129: You could add this to engine.h alongside Point.

Source/gendung.cpp
```diff
@@ -125,62 +110,43 @@ char dSpecial[MAXDUNX][MAXDUNY];
...
+	std::unique_ptr<byte[]> pSBFile = LoadLevelSOLData(tileCount);
+
+	for (unsigned i = 0; i < tileCount; i++) {
+		byte bv = pSBFile[i];
+		nSolidTable[i + 1] = (bv & byte { 0x01 }) != byte { 0 };
+		nBlockTable[i + 1] = (bv & byte { 0x02 }) != byte { 0 };
+		nMissileTable[i + 1] = (bv & byte { 0x04 }) != byte { 0 };
+		nTransTable[i + 1] = (bv & byte { 0x08 }) != byte { 0 };
+		nTrapTable[i + 1] = (bv & byte { 0x80 }) != byte { 0 };
+		block_lvid[i + 1] = static_cast<uint8_t>((bv & byte { 0x70 }) >> 4);
```
thebigMuh, line 149: Since you're casting at the end anyway:

```suggestion
		uint8_t bv = static_cast<uint8_t>(pSBFile[i]);
		nSolidTable[i + 1] = (bv & 0x01) != 0;
		nBlockTable[i + 1] = (bv & 0x02) != 0;
		nMissileTable[i + 1] = (bv & 0x04) != 0;
		nTransTable[i + 1] = (bv & 0x08) != 0;
		nTrapTable[i + 1] = (bv & 0x80) != 0;
		block_lvid[i + 1] = (bv & 0x70) >> 4;
```

Source/engine.h
```diff
@@ -598,7 +605,40 @@ void SetRndSeed(int32_t s);
...
+template <typename T, std::size_t N>
+void LoadFileInMem(const char *path, std::array<T, N> &data)
+{
+	LoadFileData(path, reinterpret_cast<byte *>(&data), N * sizeof(T));
+}
+
+/**
+ * @brief Load a file in to a buffer
+ * @param path Path of file
+ * @param pdwFileLen Number of
```
thebigMuh, line 621: Guess this should be something like
```suggestion
 * @param elements Number of T elements read
```

thebigMuh:
So for some reason all of these files contain a looong list of 16 bit values, and only 8 bits are ever used. Guess we'll have to live with that.

You could get around the swaps by implementing the tile_cast (or whatever would be an appropriate name) I suggested, and instead of doing a cast, implementing it differently for LE and BE platforms. This is one way to do it, not sure if it meets your prettiness standards :P

```
uint8_t tile_cast(uint16_t v)
{
    union naughty {
        uint16_t word;
        struct {
#if SDL_BYTEORDER == SDL_LIL_ENDIAN
            uint8_t lo;
            uint8_t hi;
#else
            uint8_t hi;
            uint8_t lo;
#endif
        };
    }:

    return naughty{ v }.lo;
}
```

Source/palette.cpp
julealgon: Shouldn't the fields be UpperCase too as per our convention?

I'd also suggest renaming to `Red`, `Green` and `Blue`.

Jmgr: > Shouldn't the fields be UpperCase too as per our convention?

Where do you see that? Why?

By the way, we are [supposed](https://github.com/diasurgical/devilutionX/wiki/Code-Style) to follow the [Webkit](https://webkit.org/code-style-guidelines/) one, but we don't: class variables (not fields by the way, that's C#) should be prefixed by `m_`, no capitals on functions, etc.

Jmgr: Maybe that Wiki would need an update @AJenbo? Or we should just follow that convention? The Webkit one is quite classic for C++. ^

AJenbo: this is what we have atm:

https://github.com/diasurgical/devilutionX/blob/master/Source/.clang-tidy#L62

took a breake from applying it as there have been a lot of prs building up and people gt grumpy from having to rebase

Source/gendung.cpp
AJenbo: in that case it would be better to just load the file as uint8_t and avoid any casting.

AJenbo: really wish bit-wise operations could be done against hex values. my main argument for doing it way i did is that we do not have a numeric value before doing the bit shift

Source/palette.cpp
thebigMuh: When you've had to type m_Green for the 50000th time, you really wish you'd just named the fields "r", "g" and "b".

Source/gendung.cpp
thebigMuh: You can do bitwise operations with hex values no problem - std::byte is the weirdo here.

It's implemented as enum class explicitly to disallow any usual integer propagation, and it has only certain operators implemented (the bitwise ones, and only for operating with other byte values).

The hex values you typed in are integer literals, so to do the bitwise operation, the std::byte would have to be promoted to integer first - which is prohibited.

thebigMuh: Oh - adding operators that allow bitwise operations (or really pretty much any operation) with other data types on std::byte isn't miracle work, but it softens up the strict typing even more. At that point, why not just use BYTE/unsigned char/uint8_t.

Source/palette.cpp
Jmgr: > When you've had to type m_Green for the 50000th time, you really wish you'd just named the fields "r", "g" and "b".

The `m_` prefix is only for private class variables. For cases like these I think these should be public and without getter/setter, and lowercase.

Source/quests.cpp
AJenbo: This is a bit of a general issue, so I would rather that it be dealt with at another point.

Also instead of the proposal here I would be more inclined to changing the internal size to match the format, it would use about 3KiB more ram but we would have support for much later tilsets and maps.

PR #1859, opened 2021-05-04
Controller Improvements

stefanmielke:
I have been playing the PSOne port and find the controller to have a few differences from this port, so I made a few changes that are in the PSOne port but are not present here that improve the way the game feels when using a controller.

Changes made:
* Do not move to attack if the player has a melee weapon equipped and the monster is not close enough to get hit
  * The player will keep its current direction and just attack
  * It's preferred when you are waiting for enemies to come to you, so you can just keep attacking and the player will hit the enemies as they change position
* Try to follow walls when moving
  * Eases navigation across the map

PS: I plan to make a few more changes for the controller to feel better, but I'm trying to understand how open you guys are to these types of changes before that.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -15,6 +15,7 @@
 #include "inv.h"
 #include "minitext.h"
 #include "missiles.h"
+#include "options.h"
 #include "stores.h"
 #include "towners.h"
 #include "trigs.h"
@@ -462,18 +463,22 @@ void Interact()
 		NetSendCmdLocParam1(true, CMD_TALKXY, towners[pcursmonst].position, pcursmonst);
 	} else if (pcursmonst != -1) {
 		if (plr[myplr]._pwtype != WT_RANGED || CanTalkToMonst(pcursmonst)) {
-			int distx = abs(plr[myplr].position.tile.x - monster[pcursmonst].position.future.x);
-			int disty = abs(plr[myplr].position.tile.y - monster[pcursmonst].position.future.y);
-			if (distx > 1 || disty > 1) {
-				// attack on the current direction
-				const int x = plr[myplr].position.future.x;
-				const int y = plr[myplr].position.future.y;
-				const int dx = x + Offsets[plr[myplr]._pdir][0];
-				const int dy = y + Offsets[plr[myplr]._pdir][1];
-				
-				NetSendCmdLoc(myplr, true, CMD_SATTACKXY, { dx, dy });
-			} else {
+			if (sgOptions.Controller.bMoveToInteract) {
 				NetSendCmdParam1(true, CMD_ATTACKID, pcursmonst);
+			} else {
+				int distx = abs(plr[myplr].position.tile.x - monster[pcursmonst].position.future.x);
+				int disty = abs(plr[myplr].position.tile.y - monster[pcursmonst].position.future.y);
+				if (distx > 1 || disty > 1) {
+					// attack on the current direction
+					const int x = plr[myplr].position.future.x;
+					const int y = plr[myplr].position.future.y;
+					const int dx = x + Offsets[plr[myplr]._pdir][0];
+					const int dy = y + Offsets[plr[myplr]._pdir][1];
+
+					NetSendCmdLoc(myplr, true, CMD_SATTACKXY, { dx, dy });
+				} else {
+					NetSendCmdParam1(true, CMD_ATTACKID, pcursmonst);
+				}
 			}
 		} else {
 			NetSendCmdParam1(true, CMD_RATTACKID, pcursmonst);
@@ -481,7 +486,7 @@ void Interact()
 	} else if (leveltype != DTYPE_TOWN) {
 		if (pcursplr != -1 && !gbFriendlyMode) {
 			NetSendCmdParam1(true, plr[myplr]._pwtype == WT_RANGED ? CMD_RATTACKPID : CMD_ATTACKPID, pcursplr);
-		} else {
+		} else if (!sgOptions.Controller.bMoveToInteract) {
 			// attack on the current direction
 			const int x = plr[myplr].position.future.x;
 			const int y = plr[myplr].position.future.y;
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -33,6 +33,9 @@ static const direction FaceDir[3][3] = {
 	{ DIR_W, DIR_NW, DIR_SW },  // LEFT
 	{ DIR_E, DIR_NE, DIR_SE },  // RIGHT
 };
+static const AxisDirection AxisDir[2][2] = {
+	{  }
+};
 static const int Offsets[8][2] = {
 	{ 1, 1 },   // DIR_S
 	{ 0, 1 },   // DIR_SW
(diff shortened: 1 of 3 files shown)


galaxyhaxz:
Another difference with the PS1 port and this is probably the best change, is that the player moves in pixel increments. You can press the dpad for a second and the player will move 2-3 pixels then stop. The way it works on PC is you move one tile at a time, which makes movement very clunky. This feature is a bit tricky to implement though.

BartyMae:
It'd be nice if the first one was optional. Not following enemies is convenient in some circumstances, but I'm pretty sure I'd want to shoot myself in the face if I have to manually try to follow enemies that run away while using an analogue stick with a melee character.

stefanmielke:
I thought on doing that at first (maybe having a setting for the minimum distance), but in the end I decided not to mostly based on my time playing warrior on the PSOne port and some of the time spent on this one.

Is there anyone else here that actively plays using a controller? Would be nice to gather more opinions on this.

AJenbo:
@glebm might have an opinion

glebm:
I'd need to play test this to be sure

stefanmielke:
Is there any actions you guys want me to do? I can put that as a config if you think we can move forward with this and then come back to this later (also ok with waiting if we want to come up with something that's more solid in the end).

Do you guys have a playtest group or something to test new ideas out? I couldn't find many reviews of the controls on youtube videos, just one (playing on an n3ds) that complained that we should lower the value to go after the monster automatically.

glebm:
@stefanmielke Not in particular, but I think a few people who use controllers should playtest this before we merge it in. If it's simply better, don't think we need a setting

Source/controls/plrctrls.cpp
```diff
@@ -445,12 +462,32 @@ void Interact()
 		NetSendCmdLocParam1(true, CMD_TALKXY, towners[pcursmonst].position, pcursmonst);
 	} else if (pcursmonst != -1) {
 		if (plr[myplr]._pwtype != WT_RANGED || CanTalkToMonst(pcursmonst)) {
```
glebm, line 464: Not sure about this `CanTalkToMonst(pcursmonst)` change, perhaps we should still walk to the monster?

stefanmielke:
sounds good to me!

I'll wait for this to be playtested then, thanks!

Source/controls/plrctrls.cpp
```diff
@@ -445,12 +462,32 @@ void Interact()
...
+				NetSendCmdLoc(myplr, true, CMD_SATTACKXY, { dx, dy });
+			} else {
+				NetSendCmdParam1(true, CMD_ATTACKID, pcursmonst);
+			}
 		} else {
 			NetSendCmdParam1(true, CMD_RATTACKID, pcursmonst);
 		}
 	} else if (leveltype != DTYPE_TOWN && pcursplr != -1 && !gbFriendlyMode) {
 		NetSendCmdParam1(true, plr[myplr]._pwtype == WT_RANGED ? CMD_RATTACKPID : CMD_ATTACKPID, pcursplr);
+	} else if (leveltype != DTYPE_TOWN) {
```
glebm, line 483: Both this and the previous branch check for `DTYPE_TOWN`, let's move them into the same branch?

Source/controls/plrctrls.cpp
```diff
@@ -810,7 +830,69 @@ bool IsPathBlocked(int x, int y, int dir)
 	return !PosOkPlayer(myplr, d1x, d1y) && !PosOkPlayer(myplr, d2x, d2y);
 }
 
-void WalkInDir(int playerId, AxisDirection dir)
+direction GetPathNotBlocked(int x, int y, int dir)
```
glebm, line 833: `GetNonBlockedDirection`? Since it doesn't return a path but just a direction

Source/controls/plrctrls.cpp
```diff
@@ -810,7 +830,69 @@ bool IsPathBlocked(int x, int y, int dir)
...
+	}
+
+	if (PosOkPlayer(playerId, dx, dy) && IsPathBlocked(x, y, dir)) {
+		return; // Don't start backtrack around obstacles
+	}
+
+	NetSendCmdLoc(playerId, true, CMD_WALKXY, { dx, dy });
+}
+
+void WalkInDir(int playerId, const AxisDirection dir)
```
glebm, line 895: Please rename one of these functions, since they do slightly different things

Source/controls/plrctrls.cpp
stefanmielke: My reasoning here is based on games that include controller support (including D3 and D2R).

On these games the user controls the character more directly (as if they are the character), and not just command them to do things (which is what happens when using a mouse on these games).

AJenbo:
I think this might be something that should be an option. Unless we can find a good way to switch between them as with shift+click on mouse+keyboard controls

stefanmielke:
@glebm and @AJenbo, fixed some of the problems and also added the "move to interact" as an option in the ini file (we can remove later if we are fine with the approach or once we do "shift+click" for the controller).

StephenCWills:
I just playtested this for a little bit on 3DS so here's my initial assessment.

#### Follow walls when moving
Having the game be a little more forgiving against walls was a bit of a blessing and a curse. It was easier to move along walls so I didn't feel like I was constantly getting stuck on nothing, but it was also a bit harder to maneuver my character through doors and arches because he'd be just as likely to try to move along the wall if I wasn't accurate with the analog stick. It would be better if he could get stuck on corners.

Also potentially worth noting, this movement logic does not appear to work on the Sarcophagi in Cathedral.

Ultimately, I'd say this change is an improvement, but could also still be improved upon. I might like it better if I got stuck on corners instead of trying to pick one of two near-equally appropriate directions.

#### bMoveToInteract
It was a bit awkward because the character doesn't turn to look at his target when swinging until the enemy is right up in his face. This meant I was frequently swinging at nothing and then would suddenly turn around to swing at whatever just approached.

As for my personal opinion, I don't think I would ever play with this option on. Currently, even without this option, you can briefly move the analog stick after the last swing to avoid chasing down the next target. If I just wait for the enemy to approach, then start swinging, and finally use the analog stick to cancel when they're dead, I get the same effect without also eliminating the option to automatically chase down monsters.

This would definitely be better if it could be incorporated as a button combination, similar to Shift+Click on mouse and keyboard, or even as a separate button or a toggle. I think this option would be most comfortable on L2 and/or R2 (even though o3DS doesn't have them 😦). If we're reluctant to change the default controller layout, maybe this could be like an "extended action" that is not mapped by default, but could be manually mapped to a button to replace the default action.

AJenbo:
> Try to follow walls when moving 

I don't think i understand how this is suppose to help, it's already annoying that you are locked in to the tile grid, if the player is additionally snapping to the 4 directions of walls it sounds like you would have and even less responsive experience.

kevixm:
I have a suggestion for controller usage.  I have been playing a ton over the last week (and this project is amazing btw) but I find casting spells is a little cumbersome on controller especially with no hot keys.  Would it be possible to use the right analog stick as 4 (or better yet 8) hockey’s for spells.  This would make playing a mage so much better!  

I guess this could also be done using the dpad but I am not sure how many people use the dpad for movement.  

Just a thought.  I have no coding experience though so honestly have no idea how tough this would be :)

AJenbo:
> I have a suggestion for controller usage. I have been playing a ton over the last week (and this project is amazing btw) but I find casting spells is a little cumbersome on controller especially with no hot keys. Would it be possible to use the right analog stick as 4 (or better yet 8) hockey’s for spells. This would make playing a mage so much better!

Hold select to use the 4 right buttons as spell hotkeys

https://github.com/diasurgical/devilutionX/wiki/Game-Controller-Scheme

kevixm:
> > I have a suggestion for controller usage. I have been playing a ton over the last week (and this project is amazing btw) but I find casting spells is a little cumbersome on controller especially with no hot keys. Would it be possible to use the right analog stick as 4 (or better yet 8) hockey’s for spells. This would make playing a mage so much better!
> 
> Hold select to use the 4 right buttons as spell hotkeys
> 
> https://github.com/diasurgical/devilutionX/wiki/Game-Controller-Scheme

My nimbus controller doesn’t have select. Just a menu button here.  Good to know though :)

AJenbo:
Maybe you could map L2 for it :/

PR #1863, opened 2021-05-04
Correct plurals

hiperiondev:


AJenbo (approved):

AJenbo:
When i set it to view untranslated strings first the following still appears untranslated:

image

julealgon:
Would be nice if we could be a bit more specific with PR/commit titles.

PR #1864, opened 2021-05-04
Change Maelstorm to Maelstrom, bunch of enhancements for Italian

M0Rf30:
This fix a vanilla typo error for the Monster named Maelstorm.
Looking around, the most popular and right term is Maelstrom 
This is a monster, not a whirlpool I know :grin: 

This also brings many enhancements for Italian localization.


M0Rf30:
Another suggestion could be to split the text of `Source/help.cpp`. This is a so tedious and long piece of text to translate.
My idea is to split the strings on || or ||| characters. These separator work as text markers. But you know better than me...

AJenbo:
Since much of this text is duplicated between the shareware end retail version we do intent to split it up based on that at the very least.

AJenbo (approved):

AJenbo:
Looks like you still have 3 translations with errors:
image
(missing spaces at start and en of line)

julealgon:
I appreciate the `Maelstorm` to `Maelstrom` change. I also raised this potential issue while looking at the pt-BR translation.

galaxyhaxz:
I never noticed the `Maelstorm` typo before. Given all the other monsters of the same type have the word storm in them (Red Storm, Storm Lord etc.) I thought this was intentional. It does look to be however, that Maelstrom is the proper English spelling of a word meaning whirlpool/violent.

If we look at the PSX port, you can see they changed it to translate in all other languages, but kept the typo in English:
```
[STR_MAELSTORM]
Eng="Maelstorm"
Fre="Malstrom"
Ger="Maelstrom"
Swe="Malström"
Jap="メイルシュトロム"
```

PR #1869, opened 2021-05-05
ADL: Remove animation skipping on queued player attacks

### LINKED ISSUE #779: Remove animation skipping on player attacks

obligaron:
Fixes #779

Attack animations can be canceled after the hit frame, if another attack is queued.
The animation of the first attack is truncated.

<Details>
<summary>Example: Warrior with Sword has hit frame 9 and animation length 16.</summary>
Game Tick | Action | Animation Frame
-- | -- | --
Start | StartAttack (first) | 1
Start | ProcessAnimation | 2
1 | ProcessAnimation | 3
2 | ProcessAnimation | 4
3 | ProcessAnimation | 5
4 | ProcessAnimation | 6
5 | ProcessAnimation | 7
6 | ProcessAnimation | 8
7 | ProcessAnimation | 9
8 | ProcessAnimation | 10
10 | StartAttack (second) | 1
10 | ProcessAnimation | 2
11 | ProcessAnimation | 3
12 | ProcessAnimation | 4
13 | ProcessAnimation | 5
14 | ProcessAnimation | 6
… | … | …
</Details>

The PR adds the truncated Frames to the beginning of the repeated Attack.
To achieve this without changing balance we have to reduce the time a frame is shown (Animation Distribution).
With the existing logic this was easy to integrate.

<Details>
<summary>Example Videos (Master vs PR)</summary>

### Master
https://user-images.githubusercontent.com/25415264/117100047-01ebb180-ad73-11eb-973e-6d1e9b9f6db7.mp4

### PR
https://user-images.githubusercontent.com/25415264/117100089-25166100-ad73-11eb-8f1c-d47c0e4008f7.mp4

### Master
https://user-images.githubusercontent.com/25415264/117100166-598a1d00-ad73-11eb-89ba-c9cf2bdaaabc.mp4

### PR
https://user-images.githubusercontent.com/25415264/117100177-60b12b00-ad73-11eb-9c0f-f1ad5ec5ac4e.mp4

### Master
https://user-images.githubusercontent.com/25415264/117100209-7aeb0900-ad73-11eb-8041-85319c93d020.mp4

### PR

https://user-images.githubusercontent.com/25415264/117100544-488ddb80-ad74-11eb-8b94-7a062c2ae77f.mp4

### Master
https://user-images.githubusercontent.com/25415264/117100250-90603300-ad73-11eb-9193-fa3066920187.mp4

### PR
https://user-images.githubusercontent.com/25415264/117100261-95bd7d80-ad73-11eb-9cad-22cbdb0ec459.mp4

### Master

https://user-images.githubusercontent.com/25415264/117100570-5a6f7e80-ad74-11eb-854e-1c5e4d43b23f.mp4

### PR

https://user-images.githubusercontent.com/25415264/117100591-65c2aa00-ad74-11eb-9667-31bb3986c094.mp4

</Details>

Notes:

- Only Animation change, no Balance change
- Make fast attacks look really fast :wink:
- If wanted this feature can be toggled with an option


FitzRoyX:
Cool stuff, the axe and sword anims are dramatically better. Really makes the originals look broken.

sumyunggai:
This looks really good, the animations always did look kinda off.. just got used to them over the years.

Source/engine/animationinfo.h
```diff
@@ -6,24 +6,43 @@
...
+inline AnimationDistributionFlags operator|(AnimationDistributionFlags a, AnimationDistributionFlags b)
+{
+	return static_cast<AnimationDistributionFlags>(static_cast<std::underlying_type_t<AnimationDistributionFlags>>(a) | static_cast<std::underlying_type_t<AnimationDistributionFlags>>(b));
+}
+inline AnimationDistributionFlags& operator|=(AnimationDistributionFlags& a, AnimationDistributionFlags b)
+{
+	a = a | b;
+	return a;
+}
+
```
AJenbo, line 45: can this be made more general so that it applies to other bitflags as well?

Source/engine/animationinfo.h
```diff
@@ -85,6 +104,10 @@ class AnimationInfo {
 	* @brief Animation Frames that will be adjusted for the skipped Frames/game ticks
 	*/
 	int RelevantFramesForDistributing;
+	/*
+	* @brief Animation Frames that wasn't shown from previous Animation
+	*/
```
AJenbo, line 109: doxygen
```suggestion
	/**
	 * @brief Animation Frames that wasn't shown from previous Animation
	 */
```

Source/engine/animationinfo.h
```diff
@@ -6,24 +6,43 @@
 #pragma once
```
AJenbo, line 6: line ending should be crlf, please run clang-format

julealgon:
This looks incredible.

Source/engine/animationinfo.h
obligaron: Yes for sure! And I would love it. 👍 

That's also why I started the discussion in discord before.
At that time I looked at different implementations, for example:

- https://github.com/grisumbras/enum-flags
  - Doesn't include much overhead
  - Works well with memset (currently needed)
  - Doesn't need a special named last enum value
  - Does this work well with GCC5/6?
  - How would we include this? In 3rdParty folder?
- https://codereview.stackexchange.com/questions/96146/c-flagset-typesafe-usage-of-enumeration-as-bitset-bitmask
  - Doesn't allow bit-flags (even with cast) that aren't allowed
  - Needs to change most Flags from Flag value (1 << 2) to Bit Index (2)
  - Doesn't work for the moment, cause we still use memset (will hopefully change soon)
- and searched for BitfieldOptions but didn't found anything 😢 

My favorite currently would be grisumbras enum-flags. But I thought this topic alone is enough for a own pr, so I tried to reduce the size and focus the topic of this pr.

As I said would love this to be generic.

Possible next steps:

1. Discuss enum flags with this pr
2. Discuss enum flags in another pr/issue
3. If you like grisumbras enum-flags in 3rdParty folder I would create a follow up pr that would introduce this.
4. Something I'm missing 🙂

AJenbo: Well if that's your intent then i suggest you use regular old enum for this PR and then we can have a look at the alternatives in a follow up that can then be applied generally.

If the same 13 lines of boiler plate have to be applied manually to each bitflag enum, then I'm not much for it. But if it can be defined in a generic class then it seams perfect.

enum-flags seams to make use of macros which i'm not a fan of.

obligaron: Converted to plain enum. 😉

Source/engine/animationinfo.h
obligaron: Thanks. 🙂
I fixed the two new introduces comments.

There are other comments that contains the same error (introduces by me in previous pr's).
Should I fix them with this pr or do you prefer a separate pr?

btw: git commit --fixup is really handy. Thanks again for the tip 🙂

AJenbo: Lets clean up existing issues seperatly for now since this PR is a bout ready

AJenbo (approved):

PR #1875, opened 2021-05-05
🎉 Clip some CEL drawing functions

glebm:
Implements clipping for all CEL drawing functions except stippled and outline.

FPS in town in this location at 640x480: 1690 -> 1715

screenshot


Source/engine.h
```diff
@@ -202,7 +202,7 @@ struct FrameHeader {
 	uint16_t row128;
 };
 
-inline const uint8_t *CelGetFrameClipped(const byte *pCelBuff, int nCel, int *nDataSize)
+inline const byte *CelGetFrameClipped(const byte *pCelBuff, int nCel, int *nDataSize)
```
AJenbo, line 205: The reason this was returning a uint8_t pointer was that the pointer from `CelGetFrameClipped` is always to RLE data, where `CelGetFrame` may be to either RLE data or a frame header. Not saying you need to change it back, just making sure you know why i did it that way :)

glebm: I see, still I think `byte` is the best type here. RLE data bytes mean different things, and it's handy to treat control bytes as signed sometimes, while color index bytes are always unsigned.

AJenbo (approved):

PR #1876, opened 2021-05-05
Improve inventory filling algorithm

Aneurin:
I've seen some discussions about auto-sorting/packing the inventory, which seems a bit contentious. This PR just ~~adds an option to tweak~~ tweaks the slot placement for picked up items, without making deeper changes.

The big differences are in the gold, which goes like this:
image
And other 1x1 items, which go like this:
image
In other words, they try to fill the bottom row from opposite ends, and once it's full they then both start working up in columns starting from the right.

2-height item placement is also changed slightly - the details for that are in the commit log. 3-height item placement is unchanged.

Not sure if anyone else is interested in this and perhaps it might be impossible to reach a consensus on the "best" filling algorithm, but this is how I want it to work so I thought it would be worth posting.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -779,8 +779,13 @@ bool GoldAutoPlace(int pnum)
 		plr[pnum]._pGold = CalculateGold(pnum);
 	}
 
-	for (int i = 39; i >= 0 && !done; i--) {
+	for (int i = 39; i >= 30 && !done; i--) {
 		done = GoldAutoPlaceInInventorySlot(pnum, i);
+	}
+	for (int x = 9; x >= 0 && !done; x--) {
+		for (int y = 2; y >= 0 && !done; y--) {
+			done = GoldAutoPlaceInInventorySlot(pnum, 10 * y + x);
+		}
 	}
 
 	return done;
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -780,26 +780,35 @@ bool GoldAutoPlace(int pnum)
 	}
 
 	for (int i = 39; i >= 0 && !done; i--) {
-		int yy = 10 * (i / 10);
-		int xx = i % 10;
-		if (plr[pnum].InvGrid[xx + yy] == 0) {
-			int ii = plr[pnum]._pNumInv;
-			plr[pnum].InvList[ii] = plr[pnum].HoldItem;
-			plr[pnum]._pNumInv = plr[pnum]._pNumInv + 1;
-			plr[pnum].InvGrid[xx + yy] = plr[pnum]._pNumInv;
-			GetPlrHandSeed(&plr[pnum].InvList[ii]);
-			int gold = plr[pnum].HoldItem._ivalue;
-			if (gold > MaxGold) {
-				gold -= MaxGold;
-				plr[pnum].HoldItem._ivalue = gold;
-				GetPlrHandSeed(&plr[pnum].HoldItem);
-				plr[pnum].InvList[ii]._ivalue = MaxGold;
-			} else {
-				plr[pnum].HoldItem._ivalue = 0;
-				done = true;
-				plr[pnum]._pGold = CalculateGold(pnum);
-				NewCursor(CURSOR_HAND);
-			}
+		done = GoldAutoPlaceInInventorySlot(pnum, i);
+	}
+
+	return done;
+}
+
+bool GoldAutoPlaceInInventorySlot(int pnum, int slotIndex)
+{
+	bool done = false;
+
+	int yy = 10 * (slotIndex / 10);
+	int xx = slotIndex % 10;
+	if (plr[pnum].InvGrid[xx + yy] == 0) {
+		int ii = plr[pnum]._pNumInv;
+		plr[pnum].InvList[ii] = plr[pnum].HoldItem;
+		plr[pnum]._pNumInv = plr[pnum]._pNumInv + 1;
+		plr[pnum].InvGrid[xx + yy] = plr[pnum]._pNumInv;
+		GetPlrHandSeed(&plr[pnum].InvList[ii]);
+		int gold = plr[pnum].HoldItem._ivalue;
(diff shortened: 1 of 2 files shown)


julealgon:
I think this is an obvious, net improvement over vanilla so would recommend doing it without any toggles whatsoever: make it the default behavior.

I really doubt anyone would be bothered (geniunely) by this as automatic placement of items is very inconsequential, it just causes issues due to the ordering chosen.

Source/inv.cpp
```diff
@@ -704,6 +708,65 @@ bool AutoPlaceItemInInventory(int playerNumber, const ItemStruct &item, bool per
...
+		}
+		for (int x = 9; x >= 0 && !done; x--) {
+			for (int y = 2; y >= 0 && !done; y--) {
+				done = AutoPlaceItemInInventorySlot(playerNumber, 10 * y + x, item, persistItem);
+			}
+		}
+	}
+
+	if (itemSize.Y == 2) {
+		for (int x = 10 - itemSize.X; x >= 0 && !done; x -= itemSize.Y) {
```
AJenbo, line 737: item y is known to be 2, probably should have been x. but also jumping a full item width may skip possibly good spots

```
xoox
xoox
```
A book can fit in the middle but if i move by the size of a book both options are half filled.

```suggestion
		for (int x = 10 - itemSize.X; x >= 0 && !done; x--) {
```

Source/inv.cpp
```diff
@@ -704,6 +708,65 @@ bool AutoPlaceItemInInventory(int playerNumber, const ItemStruct &item, bool per
...
+				done = AutoPlaceItemInInventorySlot(playerNumber, 10 * y + x, item, persistItem);
+			}
+		}
+		if (itemSize.X == 2) {
+			for (int x = 7; x >= 0 && !done; x -= 2) {
+				for (int y = 0; y < 3 && !done; y++) {
+					done = AutoPlaceItemInInventorySlot(playerNumber, 10 * y + x, item, persistItem);
+				}
+			}
+		}
```
AJenbo, line 748: this doesn't seam useful, all x2 items should be served fine by y3 or y2 paths

Source/diablo.cpp
```diff
@@ -534,6 +534,7 @@ static void SaveOptions()
 	setIniInt("Game", "Auto Equip Helms", sgOptions.Gameplay.bAutoEquipHelms);
 	setIniInt("Game", "Auto Equip Shields", sgOptions.Gameplay.bAutoEquipShields);
 	setIniInt("Game", "Auto Equip Jewelry", sgOptions.Gameplay.bAutoEquipJewelry);
+	setIniInt("Game", "Improved Inventory Placement", sgOptions.Gameplay.bImprovedInventoryPlacement);
```
AJenbo, line 537: replace the original instead

Source/inv.cpp
```diff
@@ -704,6 +708,65 @@ bool AutoPlaceItemInInventory(int playerNumber, const ItemStruct &item, bool per
...
+ * algorithm that should typically provide better packing than AutoPlaceItemInInventory().
+ * @param playerNumber The player number on whose inventory will be checked.
+ * @param item The item to be checked.
+ * @param persistItem Pass 'True' to actually place the item in the inventory. The default is 'False'.
+ * @return 'True' in case the item can be placed on the player's inventory and 'False' otherwise.
+ */
+bool AutoPlaceItemInInventoryAlternate(int playerNumber, const ItemStruct &item, bool persistItem)
+{
+	InvXY itemSize = GetInventorySize(item);
+	bool done = false;
```
AJenbo, line 723: `return true:` on success instead of tracking the state of done all over the function

Source/inv.cpp
```diff
@@ -704,6 +708,65 @@ bool AutoPlaceItemInInventory(int playerNumber, const ItemStruct &item, bool per
...
+		for (int i = 0; i < 9 && !done; i++) {
+			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
+		}
+
+		for (int i = 10; i < 19 && !done; i++) {
+			done = AutoPlaceItemInInventorySlot(playerNumber, i, item, persistItem);
+		}
+	}
+
+	return done;
```
AJenbo, line 767:
```suggestion
	return false;
```

Source/inv.cpp
Aneurin: Yes, that should have been X, thanks.

As to why it's `x -= itemSize.X`: I did actually write it as just `x--` the first time but I played with that for half an hour or so and decided I didn't like the result.

The original placement algorithm for 2x2 items starts out by trying a set list of preferred positions before falling back to searching one square at a time, and without that it's very easy to end up with two rows of books that are misaligned by one. I don't *think* that misalignment ever results in technically worse packing, but I also don't think it results in technically *better* packing either (because you just end up with a 1x1 or 1x2 gap in a different place), so I decided to go with the version that looks neater visually.

This does lead to a bit of special casing for 2x2 items, so I didn't quite succeed in my original plan to make 2x2 and 1x2 items use the same placement, but it does at least have less special-casing than the original.

Source/inv.cpp
Aneurin: This is the counterpart to the `x -= itemSize.X` issue, which is that this placement first attempts to align 2x2 objects by placing them in even columns, and only if it didn't succeed will it then try again using odd-column alignment, to prevent the `xoox` problem you mentioned above.

Source/inv.cpp
Aneurin: Does the updated version look any better to you? It doesn't return immediately but does return immediately at the end of each section now.

The only obvious way I can see to eliminate this extra `done` variable would be to replace all lines like
```
done = AutoPlaceItemInInventorySlot(...);
```
with
```
if (AutoPlaceItemInInventorySlot(...) {
    return true;
}
```
and I'm not convinced that's an improvement. I'll change it to do that if that's what you're meaning though?

AJenbo (approved):

PR #1878, opened 2021-05-06
ADL: Cleanup comments & clang-format

obligaron:
Content:

- Cleaned-up all comments with fixed Doxygen Style
- run clang-format on animationinfo.h/.cpp and animationinfo_test.cpp

@AJenbo I'm not sure how to correctly check the line ending.
Is it enough if I run clang-format / check in Visual Studio or Notepad++ the Line-Ending settings before I commit?
Sorry for this perhaps silly question. I just don't want to repeat this error anymore. :pensive:

### DIFF
--- a/Source/engine/animationinfo.h
+++ b/Source/engine/animationinfo.h
@@ -17,13 +17,13 @@ namespace devilution {
  */
 enum AnimationDistributionFlags : uint8_t {
 	None = 0,
-	/*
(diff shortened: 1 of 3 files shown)


AJenbo:
Running `clang-format` should be enough. Maybe you have something setup in git to do automatic line ending conversion?

AJenbo:
I have corrected all renaming instances of bad line endings (using unix2dos) and ran clang-format on all files.

PR #1886, opened 2021-05-07
3ds sdl updates

StephenCWills:
With the latest changes to `devkitPro/3ds-sdl`, it is now possible to remove the call to `gspWaitForVBlank()` to improve performance. With this change, I am able to run DevilutionX very close to 20 fps on the old 3DS.

In addition, SDL now maps the D-pad as a hat rather than four separate buttons. The controller mapping has been updated to reflect the changes.


AJenbo (approved):
Nice, an we still have a couple of optimizations planed, so maybe we will get there even without any special changes.

PR #1890, opened 2021-05-07
Some more NPC clean up and animation fixes

AJenbo:


glebm:
`vector` doesn't seem to be appropriate here -- this is static data, let's not copy it to the heap

How about just setting a pointer to a static C-array instead?

AJenbo:
C-array would work. I tried std::array but it wasn't happy about it.
Still how can we get size for that?

Iniquitatis:
Why it (compiler, I assume?) wasn't happy? The only thing I noticed, which could make compiler complain, is a usage of `std::vector::clear`. Could be replaced by `std::array::fill(0)`.

AJenbo:
> Why it (compiler, I assume?) wasn't happy?

It wasn't really the compiiler, the issue is how to get the size of the array at runtime. Originally it was done by scanning for an item with the value -1, but that wasn't a very good solution, so first i used a vector, but @glebm suggested just using a static array with a sizeof() instead so that there wouldn't be any runtime allocation.

Iniquitatis:
Checking static array length by `sizeof(x) / sizeof(underlying_type)` will give you the same result as the `std::array::size`.
So, uhh... Either that or storing a separate variable for the real element (`!= -1`) count.

Or maybe I just don't understand the issue.

AJenbo:
I don't think you do :) The original used a fixed size multi array `char AnimOrder[6][148]` so doing so would have just proceed 148 in all cases.

glebm:
To clarify, the array sizes are different for different towners

Iniquitatis:
Yep, I've read the code.
But `sizeof()` of `char[148]` (doesn't matter if it contained in another array) is still 148 bytes, which in this case is the amount of all array elements, not just non-empty ones. So, I can't really imagine how `sizeof()` helps when a dynamic element count is needed. You will still need to process all 148 elements, or break a loop on first `-1`.
In the situation where heap allocation is undesirable, I'd just use two variables: `char blah[MAX_ELEMENTS]`* and `some_int blahCount`.

&ast; -- or `std::array<char, MAX_ELEMENTS>`

Or I just need to get some sleep and sorry for being stupid. :D

AJenbo:
> Yep, I've read the code.

Did you read both the old and the new code? You seam to be either confused about the old, or suggesting the things we actually ended up doing already :)

Iniquitatis:
D'oh!
Yes, I haven't read the current code, only diff from this PR.
Apologies for wasting time, I was just surprised and thought that you've somehow overlooked some basic stuff. 😳

PR #1894, opened 2021-05-07
🌐 Fix punctuation in Italian localization

M0Rf30:
As in title
Regards


AJenbo:
Thanks working on é support right now :)

PR #1897, opened 2021-05-08
Generic text rendering

AJenbo:
This should gradually replace all the direct rendering of game texts
throughout the code. The interface is made to closely mirror that of the
art fonts as that is what will eventually be used for rendering Unicode
fonts both in the menus and in-game.

Performance before 1082fps and after 1142fps
(my FPS isn't super stable so your mileage may vary)

### DIFF
--- a/Source/engine/render/automap_render.cpp
+++ b/Source/engine/render/automap_render.cpp
@@ -1,3 +1,8 @@
+/**
+ * @file automap_render.cpp
+ *
+ * Line drawing routines for the automap.
+ */
 #include "engine/render/automap_render.hpp"
 
 namespace devilution {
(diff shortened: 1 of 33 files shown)


Source/engine/render/automap_render.cpp
```diff
@@ -1,3 +1,8 @@
+/**
+ * @file automap_render.cpp
+ *
+ * Sprite animation handeling.
```
glebm, line 4: This comment looks wrong. Moreover, why even have file comments for `cpp` files when there is a corresponding header?

AJenbo: Doxygen doesn't pickup files unless they have a header :/

AJenbo: I suppose we could see this as a feature and only pickup public header functions variables etc?

glebm: Yeah, I think it's a feature -- the stuff in cpp files is implementation details

glebm (approved):

PR #1899, opened 2021-05-08
Set 3DS scaling behavior based on FitToScreen setting

StephenCWills:
Setting `Fit to Screen=1` in the ini file will force the display to stretch to fill the top screen.
Setting `Fit to Screen=0` in the ini file will instead keep the aspect ratio and scale to fit width or height as appropriate based on the resolution.

Because the 3DS display has to scale to fit 400x240 output, these are really the only two options. Between `Fullscreen`, `Upscale`, and `Fit to Screen`, I wasn't sure which one best described this behavior. I settled on `Fit to Screen` since that's what it feels like I'm doing by forcing the display to stretch and fill in the available space.


Source/utils/display.cpp
```diff
@@ -33,6 +33,17 @@ Uint16 gnViewportHeight;
 Uint16 borderRight;
 
 #ifdef USE_SDL1
+#ifdef __3DS__
+uint32_t Get3DSScalingFlag(int width, int height)
```
glebm, line 37: This should probably live in `Source/platform/ctr/display.hpp`

CMake/ctr/n3ds_defs.cmake
```diff
@@ -15,7 +15,6 @@ set(TTF_FONT_DIR \"romfs:/\")
 
 #Force scaling, for now..
```
glebm, line 16: I'm guessing this comment is no longer true?

Source/utils/display.cpp
StephenCWills: Do you mean it should be declared there? Or should it be an inline function?

glebm: You can define it either `inline` in the header or in a `cpp` file
It is only called once so it doesn't matter much which one it is

CMake/ctr/n3ds_defs.cmake
StephenCWills: That's correct. Thanks for catching that.

Source/utils/display.h
```diff
@@ -31,6 +31,17 @@ void SetVideoModeToPrimary(bool fullscreen, int width, int height);
 
 bool IsFullScreen();
 
+#ifdef __3DS__
```
glebm, line 34: My comment was about moving this to a platform-specific file, e.g. Source/platform/**ctr**/display.hpp

Source/utils/display.cpp
StephenCWills: Oh, I didn't read that carefully enough, and I was a little confused. I thought you just wanted it moved to the `Source/utils/display.h` header file.

Source/utils/display.h
StephenCWills: I think I got it right this time.

glebm (approved):

PR #1900, opened 2021-05-08
🎉 CEL rendering: Clip outline and stippled transparency

glebm:
Clips `CelBlitOutlineTo` and `CelBlitLightTransSafeTo` to the output buffer.

The templated implementation of the outline clipping means that this compiles to many case-specific functions with near-branchless inner loops. This increases x64 binary size by 92 KiB. We may lower some of these template parameters to values in the future (this is easier to do than the other way around) after benchmarking.

No performance difference from the outline clipping, with stippled transparency 1400 -> 1420 in this scene:

image


Iniquitatis:
Beautiful

AJenbo (approved):

PR #1901, opened 2021-05-08
Reenabling Elemental spell for Hellfire

thebigMuh:


NikoVP:
Great news!

galaxyhaxz:
Hmm, Elemental should never have been disabled in the first place, unless this was an accident carried over from the spawn build which removed many spells.

thebigMuh:
> Hmm, Elemental should never have been disabled in the first place, unless this was an accident carried over from the spawn build which removed many spells.

Yeah I talked to AJenbo about this, and he thinks it may have been a mistake introduced during a merge.

AJenbo:
could you fix this upstream?

Source/spells.cpp
```diff
@@ -405,15 +396,6 @@ int GetSpellStaffLevel(spell_id s)
 	if (!gbIsHellfire && s > SPL_LASTDIABLO)
 		return -1;
 
-	if (gbIsHellfire) {
-		switch (s) {
-		case SPL_ELEMENT:
```
AJenbo, line 410: please add it to the spawn exception

AJenbo (approved):

Source/spells.cpp
thebigMuh: Did so. I tried finding a comprehensive list of changes between standard Diablo and the Spawn install, to check for other differences, but wasn't really able to find anything. Do you maybe have a link?

PR #1905, opened 2021-05-09
♻️ file_util: Cleanup Windows UTF-8-to-16 conversion

glebm:


AJenbo:
this doesn't by any chance fix the issue with save paths that contain unicode symbols?

AJenbo (approved):

glebm:
I don't think so, this is just deduplication + unique_ptr + validation via `MB_ERR_INVALID_CHARS`

PR #1906, opened 2021-05-09
Add listfile for devilutionx.mpq

StephenCWills:
Tested with MPQ Editor. Is the filename okay?


Packaging/resources/devilutionx.mpq.txt
```diff
@@ -0,0 +1,10 @@
+data\health.pcx
+data\healthbox.pcx
+data\resistance.pcx
+data\xpbar.pcx
+Gendata\Cutportlw.pcx
+Gendata\Cutportrw.pcx
```
AJenbo, line 6: despite calling with upper case all files are looked up in lowercase on fs

```suggestion
gendata\cutportlw.pcx
gendata\cutportrw.pcx
```

StephenCWills: I made this change last night without really thinking about it, but perhaps I should have asked. MPQ Editor actually has no trouble finding the files when using a listfile with mixed casing. Is this just an implementation detail for MPQ Editor to make the filenames case insensitive, or is that how the MPQ format works in general? Wouldn't it be better to use the mixed-case names if it doesn't cause problems for MPQ readers?

AJenbo:
We should just add this inside the MPQ so that it is automatically loaded by mpq tools.

StephenCWills:
> We should just add this inside the MPQ so that it is automatically loaded by mpq tools.

I thought so too, but MPQ Editor wasn't letting me do that for some reason. I can try to fiddle with it some more to see if I can make it work.

StephenCWills:
I guess it wasn't letting me add the listfile because I had Diablo 1 compatibility selected. I suppose it doesn't have to be Diablo 1 compatible, though, so long as it's compatible with StormLib, right?

AJenbo (approved):

StephenCWills:
> Really odd that it would reject the file for that reason.

I guess it's because listfiles were not included in MPQ archives until StarCraft, although I don't believe it would break compatibility to enable its inclusion.

Anyway, the latest commit provides a brand new mpq file with a version 1 header, PKWARE compression, and no encryption. The file size is much smaller, so that's nice. It adds `(listfile)` and `(attributes)` so now the listfile is embedded in the archive itself. IMO, `(attributes)` doesn't seem that useful, but MPQ Editor put it in there automatically so I just left it in.

FYI, I was going to use ZLIB compression because that's the default in MPQ Editor, but quickly discovered that we had disabled ZLIB in StormLib unless compiled with the `FULL` option.

PR #1907, opened 2021-05-09
Put CharisSILB.ttf into devilutionx.mpq

StephenCWills:
Puts the TTF font file into devilutionx.mpq.

I refactored game initialization logic to load devilutionx.mpq before creating the game window. That's because errors that occur after the game window is created use the font to display the error message. This way, it could attempt to open the font file early and display an error message using the generic error dialog if it doesn't work.

When the game is actually loading a font, it still checks for the file in the usual locations before explicitly loading `CharisSILB.ttf` from devilutionx.mpq. This allows the font to be overridden using `TTF_FONT_DIR` and `TTF_FONT_NAME`.

In order to put the font path into the listfile for the mpq, this PR is based on #1906.


StephenCWills:
I forgot to mention, this was tested on Windows and 3DS. It seems that 3DS does not have a generic error dialog and simply crashes without an error message if devilutionx.mpq is not found. For 3DS specifically, this shouldn't really be an issue with the recent changes to romfs, but some of the other platforms should probably be tested.

Source/init.cpp
```diff
@@ -165,6 +164,27 @@ void init_archives()
...
+{
+	std::vector<std::string> paths;
+	get_mpq_search_paths(paths);
+
+	devilutionx_mpq = init_test_access(paths, "devilutionx.mpq");
+	HANDLE fh = nullptr;
+	if (!SFileOpenFile("font\\CharisSILB.ttf", &fh)) {
+		app_fatal("%s", _("Failed to load font resources.\n"
+		                  "\n"
+		                  "Make sure devilutionx.mpq is in the game folder and that it is up to date."));
```
AJenbo, line 179: This message will be hard to read with no font loaded :)

StephenCWills: I used the following code as a model, assuming it would be similarly difficult to see the error message if the game window couldn't be created.
https://github.com/diasurgical/devilutionX/blob/792817048306a63f599be5efd7ffef1e68b93a47/Source/init.cpp#L233

Here is how it looks on Windows.
image

As I mentioned above, 3DS simply crashes without displaying an error message. I am curious about what happens on Linux.

AJenbo: Right that's how it looks if a window has not yet been created, guess this is a bit ealier in the process then it used to be.

StephenCWills: Yeah, this is the complication I was referring to on Discord. I had to load devilutionx.mpq early so the error message would be visible. If we can move away from TTF, this would no longer be necessary. That's why I think we should hold off on merging until we make a firm decision regarding TTF fonts.

AJenbo: Well do we not have the same problem if using bitmap fonts? Since they would also need to be loaded from the file or diabdat.mpq? In fact we should probably adjust the dialog system to use the OS GUI when no font is currently loaded to avoid the "strange red empty dialog" issue.

galaxyhaxz: Have a small error case font converted to a byte array and embedded directly in the executable? :P

StephenCWills: Good point. I suppose you're right. And yeah, I guess loading early feels like a bit of a hack compared to using the OS GUI.

AJenbo: I think OS GUI might only be available with SDL2 though so we should take that in to consideration as well.

StephenCWills:
I simplified by removing the code to refactor initialization. In the latest version, the only change I made to init.cpp is to load devilutionx.mpq first so the font would be available if diabdat.mpq fails to load.

Source/init.cpp
StephenCWills: I believe this discussion is resolved with #1922.

StephenCWills:
I thought it might be a good time to dust the cobwebs off this now that the Android build is merged into master.

AJenbo:
Yeah it may make sens depending on when we feel the release of android should happen.

AJenbo:
TTF is no more :)

AJenbo:
Still let me just grab some stuff from this one..

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/e5d32da8a2a9174af29e22293ddeba738993c69f and https://github.com/diasurgical/devilutionX/commit/8f5060750ce3477a1fc7baf0aa1b86f2063b8c68 :)

PR #1908, opened 2021-05-09
🧹 towners.cpp: Avoid allocations for animOrder

glebm:
There is no need for `animOrder` to be a vector -- it's static data.

We don't have `std::span` yet, so opted for pointer + size.


AJenbo (approved):

AJenbo:
Look ok

PR #1909, opened 2021-05-09
Add std::span backport

glebm:
Adds [std::span](https://en.cppreference.com/w/cpp/container/span) backport from https://github.com/martinmoene/span-lite/tree/465909f2340187dd0da687081d0489a20b247f6f

It's a small header-only library, added directly instead of via `FetchContent` to reduce the amount of downloads during CMake configure and to make it easier for the distro maintainers.


AJenbo:
Idk, this seams a bit excessive.

PR #1912, opened 2021-05-09
🐞 Fix Windows UTF-8-to-16 conversion

glebm:
Follow-up to #1905


AJenbo:
Is it me or does this code keep growing with each clean up :)

Trihedraf:
@glebm Does stat not handle Unicode properly because it works on Windows?

Issue #1642, opened 2021-04-22
Opening the chests leads to disappear potions


Chance4us:
Opening the chests sometimes leads to disappear the potions from the belt.
I can only remember that in Hellfire the potions are sometimes manipulated when opening chests.

How to reproduce:
Unfortunately not simply reproducible as it happens randomly.
Simply open a randomly affected chest and many potions will disappear from the belt.
No sound effect is played as usually happens when opening trapped chests.

Expected behavior:
Potions should not disappear from belt when opening chests.

johndoe71rus:
this is trap in chests

Chance4us:
> this is trap in chests

It is normal that no sound effect is played for this trap?

NikoVP:
There should be a burpy or breaking glass sound, if I remember correctly.

AJenbo:
burp = drain mana
glass = remove potions

NikoVP:
The break potions does not always remove them, but also lowers the grade full to small.

Psojed:
Can confirm, this is a trap from Hellfire expansion. It plays the glass breaking sound, and works the same way in DevilutionX 1.2.1.
If you want to test this further, I recommend using a Rogue as she can see trapped chests.

Chance4us:
Finally, I see that some participants sometimes like to do bashing without trying to understand the issue.
But to get ahead just try to open the box next to you and hear the sound.
Savegame:
[chest.zip](https://github.com/diasurgical/devilutionX/files/6447459/chest.zip)

AJenbo:
Looks like this might be a vanilla bug that only happens if you only have potions in your belt.

Chance4us:
Thanks a lot for fixing.

PR #1915, opened 2021-05-09
Fix breaking glass sound is missing for small life/mana potions

obligaron:
Fixex #1642
If we break something, we should break and don't continue if nothing happens 🤣 

Thanks @Chance4us for the nice save and reporting 🙂


AJenbo:
The case for IMISC_REJUV is wrong as well it should be randomly creating a mana or heal, but is always overwritten as heal. Would you be willing to correct that also :)

obligaron:
> 
> 
> The case for IMISC_REJUV is wrong as well it should be randomly creating a mana or heal, but is always overwritten as heal. Would you be willing to correct that also :)

Sure and Done 🙂
Good that Change4us hero had enough money for the test :grin:

Issue #747, opened 2020-06-05
Support for non-ASCII symbols in file path


blacksparrow123:
Currently, if you put deilutionx directory into folder with non-ASCII in its name, devilutionx.exe won't be able to find diabdat.mpq, even if it's present.

Error: "Read-Only directory error"

Same goes for \AppData\Roaming\diasurgical\devilution.

My user name is written in cyrillic symbols and because of it, i can't start the game.
I'm just getting "Unable to write" error on start.

P.S. OS: Windows 10.

Issue #1246, opened 2021-03-22
ini: Switch from Radon to simpleini


glebm:
This ini library preserves comments and allows us to set them programmaticaly: https://github.com/brofield/simpleini

julealgon:
I'm confused by this. Didn't we just merge the change to toml11? What is simpleini?

glebm:
The toml11 approach didn't pan out (the issue and the draft PR are closed without merging)

PR #1920, opened 2021-05-09
Allow non-ASCII symbols in data/save-directory

obligaron:
Contributes to #747

## Background

Windows can per default only access UTF8-Paths with wchar_t function overloads.

References:
- https://stackoverflow.com/questions/396567/is-there-a-standard-way-to-do-an-fopen-with-a-unicode-string-file-path
- https://stackoverflow.com/questions/821873/how-to-open-an-stdfstream-ofstream-or-ifstream-with-a-unicode-filename

## Main Changes (on windows):

- Use `fstream`-Constructor with `wchar_t*` overload
- Definie Unicode for StormLib so that a `wchar_t*` (`TCHAR*`)  `SFileOpenArchive` is present

## Restrictions

Currently config-dir is not supported, cause it's not possible to pass a File-Handle or Stream to `Radon`.
But the game will start anyway, only the ini file is ignored (not read from or written to).

## Tested

- Open Game
- Play Intro
- Create Hero
- Save Hero
- Load Hero
- Delete Hero
- Read only check

## Alternative

An alternative could be https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/use-utf8-code-page
This also works (tested with master and changed manifest) and it's much cleaner. But the main problem is that it's only supported since Windows 10 Version 1903.
If we want to go down this route, I can also open a PR with the adjusted Manifest-File.

### DIFF
--- a/Source/utils/log.hpp
+++ b/Source/utils/log.hpp
@@ -11,7 +11,7 @@
 namespace devilution {
 
 // Local definition to fix compilation issue due to header conflict.
-void app_fatal(const char *pszFmt, ...);
+[[noreturn]] void app_fatal(const char *pszFmt, ...);
 
 enum class LogCategory {
 	Application = SDL_LOG_CATEGORY_APPLICATION,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,6 +265,11 @@ add_library(StormLib STATIC
   3rdParty/StormLib/src/SFileOpenFileEx.cpp
   3rdParty/StormLib/src/SFileReadFile.cpp)
 
+if(WIN32)
+# Enable Unicode for StormLib wchar_t* file APIs
+target_compile_definitions(StormLib PRIVATE -DUNICODE -D_UNICODE)
+endif()
+
 add_library(PKWare STATIC
   3rdParty/PKWare/explode.cpp
   3rdParty/PKWare/implode.cpp)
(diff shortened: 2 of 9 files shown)


AJenbo:
the alternative seams much more sane, but also windows 10, feels like a steep requirement

AJenbo:
do you have an idea for how to solve the ini issue or should we wait untill implementing https://github.com/diasurgical/devilutionX/issues/1246 ?

obligaron:
> 
> 
> do you have an idea for how to solve the ini issue or should we wait untill implementing #1246 ?

We would need to change the implementation of `Radon` (upstream or local).
simpleini already seems to [support](https://github.com/brofield/simpleini/blob/7bca74f6535a37846162383e52071f380c99a43a/SimpleIni.h#L96) reading/writing from a stream.
So I would wait until #1246 is implemented.

AJenbo (approved):

Source/utils/log.hpp
```diff
@@ -11,7 +11,7 @@
 namespace devilution {
 
 // Local definition to fix compilation issue due to header conflict.
-void app_fatal(const char *pszFmt, ...);
+[[noreturn]] void app_fatal(const char *pszFmt, ...);
```
AJenbo, line 14: thanks

AJenbo:
hmm i get "read only" now when trying to run the game on a Mac

PR #1921, opened 2021-05-10
Fix loading of infravision palette transformation

thebigMuh:
...instead of actual unique monster transformation.

Fixes #1868 which made all uniques look positively sunburnt:

image


AJenbo:
thanks, was having a bad copy paste that day 😅

PR #1922, opened 2021-05-10
Use OS GUI to display error message if font fails to load

StephenCWills:
This turned out to be much simpler than I anticipated. The routine that displays the error dialog was already designed to fall back on the OS GUI if an error occurs while displaying the dialog so all I had to do was call it recursively.

On platforms using SDL1, it uses `SDL_Log()` instead of displaying the error message.
https://github.com/diasurgical/devilutionX/blob/73a90502de8bab62f55b9ebe5ce3a615dcb32f22/Source/utils/sdl2_to_1_2_backports.h#L126-L133

This presents opportunities for platform-specific improvements.


AJenbo (approved):
Did try to make it a bit sensible :)

julealgon:
Nice! This will help so much with future bug reports.

PR #1932, opened 2021-05-11
Custom message box for 3DS

StephenCWills:
Custom implementation of `SDL_ShowSimpleMessageBox()` using features to display console output. Thanks to the romfs changes, it would be extremely unlikely for an error dialog to ever fall back on this function, but I feel it's still good to have a fallback that doesn't just silently "crash to desktop".

### DIFF
--- a/Source/platform/ctr/messagebox.cpp
+++ b/Source/platform/ctr/messagebox.cpp
@@ -1,0 +1,29 @@
+#include <3ds.h>
+#include <SDL.h>
+#include <fmt/core.h>
+#include "utils/sdl2_to_1_2_backports.h"
+
+int SDL_ShowSimpleMessageBox(Uint32 flags,
+    const char *title,
+    const char *message,
+    SDL_Surface *window)
+{
+	if (SDL_ShowCursor(SDL_DISABLE) <= -1)
+		SDL_Log("%s", SDL_GetError());
+
+	bool init = !gspHasGpuRight();
+	auto text = fmt::format("{}\n\n{}", title, message);
+
+	if (init)
+		gfxInitDefault();
+
+	errorConf error;
+	errorInit(&error, ERROR_TEXT, CFG_LANGUAGE_EN);
+	errorText(&error, text.c_str());
+	errorDisp(&error);
+
+	if (init)
+		gfxExit();
+
+	return 0;
+}
--- a/Source/utils/sdl2_to_1_2_backports.h
+++ b/Source/utils/sdl2_to_1_2_backports.h
@@ -123,6 +123,12 @@ enum {
 	// clang-format on
 };
 
+#ifdef __3DS__
+int SDL_ShowSimpleMessageBox(Uint32 flags,
+    const char *title,
+    const char *message,
+    SDL_Surface *window);
+#else
 inline int SDL_ShowSimpleMessageBox(Uint32 flags,
     const char *title,
     const char *message,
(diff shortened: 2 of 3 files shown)


Source/platform/ctr/messagebox.cpp
```diff
@@ -0,0 +1,50 @@
...
+		SDL_Log("%s", SDL_GetError());
+
+	devilution::SetVideoMode(320, 240, 0, SDL_CONSOLETOP | SDL_BOTTOMSCR);
+	DisplayMessage(flags, title, message);
+	SDL_Log("Press any button to continue");
+
+	int successCode = WaitForJoystickButton();
+	int width = devilution::gnScreenWidth;
+	int height = devilution::gnScreenHeight;
+	devilution::SetVideoModeToPrimary(false, width, height);
```
AJenbo, line 48: Why isn't this code in the `devilution` namespace?

MrHuu:
Any reason not to use the platform specific error applet? 

https://github.com/diasurgical/devilutionX/blob/8eb9c930fe1744f9976d4fd319c0b0bdab993d49/Source/platform/ctr/system.cpp#L52-L55

Source/platform/ctr/messagebox.cpp
StephenCWills: It's an SDL2 backport. I can move all the logic to the devilution namespace if you like, but I believe the `SDL_ShowSimpleMessageBox()` function itself has to be defined outside the namespace. Correct me if I'm wrong.

StephenCWills:
> Any reason not to use the platform specific error applet?

Only my own ignorance. The simple solution was staring me in the face, and I still did it the hard way. 😩

Source/platform/ctr/messagebox.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+int SDL_ShowSimpleMessageBox(Uint32 flags,
+    const char *title,
+    const char *message,
+    SDL_Surface *window)
+{
+	if (SDL_ShowCursor(SDL_DISABLE) <= -1)
+		SDL_Log("%s", SDL_GetError());
+
+	devilution::SetVideoMode(320, 240, 0, SDL_CONSOLETOP | SDL_BOTTOMSCR);
```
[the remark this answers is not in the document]
AJenbo: I'm assuming that `SDL_CONSOLETOP` means to put it on the top screen?

And combined with SDL_BOTTOMSCR it would put it on the bottom screen.

But isn't that what SDL_CONSOLEBOTTOM does (put the console on the bottom screen)?

MrHuu:
Just one caveat when using the error applet.
You have to make sure graphics are initialized before calling `errorDisp`, otherwise you'll get an arm exception.

Nice job btw, on the recent commits towards the 3ds port. Much appreciated.

Source/platform/ctr/messagebox.cpp
StephenCWills: This new version with the error applet doesn't reference the devilution namespace so this should be resolved.

Source/utils/sdl2_to_1_2_backports.h
```diff
@@ -123,6 +123,12 @@ enum {
 	// clang-format on
 };
 
+#ifdef __3DS__
+int SDL_ShowSimpleMessageBox(Uint32 flags,
```
glebm, line 127:
```suggestion
/** Defined in Source/platform/ctr/messagebox.cpp */
int SDL_ShowSimpleMessageBox(Uint32 flags,
```

PR #1934, opened 2021-05-11
🎉 Final clipping fix + buffer padding removal

glebm:
CL2 outline is now clipped. As this is the final clipping fix, we remove the buffer padding.

After this we can start work on switching to hardware surfaces + double-buffering on platforms where we want to use 8-bit surfaces.

### DIFF
--- a/Source/engine/render/cl2_render.cpp
+++ b/Source/engine/render/cl2_render.cpp
@@ -21,15 +21,13 @@ namespace {
  *    indicates a fill-N command.
  */
 
-constexpr std::uint8_t MaxCl2Width = 65;
-
 constexpr bool IsCl2Opaque(std::uint8_t control)
 {
 	constexpr std::uint8_t Cl2OpaqueMin = 0x80;
 	return control >= Cl2OpaqueMin;
 }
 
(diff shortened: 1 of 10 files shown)


Source/engine/render/cl2_render.cpp
```diff
@@ -407,34 +322,350 @@ void Cl2BlitLightSafe(const CelOutputBuffer &out, int sx, int sy, const byte *pR
...
-			auto width = static_cast<std::int8_t>(*dst++);
+			auto v = static_cast<std::uint8_t>(*dst++);
 			nDataSize--;
 			assert(nDataSize >= 0);
-			if (width < 0) {
-				width = -width;
-				if (width > MaxCl2Width) {
-					nDataSize--;
-					assert(nDataSize >= 0);
+			if (!IsCl2Opaque(v)) continue;
```
AJenbo, line 656: is this allowed by the code style?

glebm: Re-formatted the whole while

AJenbo:
godd to see all the boarders go, maybe we can even simplify some position logic in various parts now

PR #1938, opened 2021-05-11
DrawLightRed cleanup

qndel:
no reason to pass 1 as a function argument as you can't really get many different results by passing different values - 2 = grey (probably using stone curse trn?), 4+ = black. Only 1 is used in game so might as well hardcode it.


AJenbo (dismissed):

Source/scrollrt.cpp
```diff
@@ -188,17 +188,13 @@ static void scrollrt_draw_cursor_item(const CelOutputBuffer &out)
 
 	const auto &sprite = GetInvItemSprite(pcurs);
 	const int frame = GetInvItemFrame(pcurs);
+	void (*drawingFun)(const CelOutputBuffer &, int, int, const CelSprite &, int) = &CelClippedDrawSafeTo;
```
julealgon, line 191: This is extremely ugly to me.... having to define the entire signature parameters like that...

Is there no other way, perhaps using `auto`?

Source/inv.cpp
```diff
@@ -230,11 +232,8 @@ void DrawInv(const CelOutputBuffer &out)
 				CelBlitOutlineTo(out, GetOutlineColor(plr[myplr].InvBody[slot], true), RIGHT_PANEL_X + screen_x, screen_y, cel, celFrame, false);
 			}
 
-			if (plr[myplr].InvBody[slot]._iStatFlag) {
-				CelClippedDrawTo(out, RIGHT_PANEL_X + screen_x, screen_y, cel, celFrame);
-			} else {
-				CelDrawLightRedTo(out, RIGHT_PANEL_X + screen_x, screen_y, cel, celFrame, 1);
-			}
+			drawingFun = (plr[myplr].InvBody[slot]._iStatFlag ? &CelClippedDrawTo : &CelDrawLightRedTo);
+			drawingFun(out, RIGHT_PANEL_X + screen_x, screen_y, cel, celFrame);
```
glebm, line 236: It looks like the function call is mostly controlled by the `_iStatFlag`.

Instead of assigning function pointers, let's add a wrapper:

```c++
void CelDrawWrapper(bool red, const CelOutputBuffer &out, int x, int y, const CelSprite &cel, int frame)
{
	if (red) {
		CelDrawLightRedTo(out, x, y, cel, celFrame);
	} else {
		CelClippedDrawTo(out, x, y, cel, celFrame);
	}
}
```

This is a bit more readable and easier to trace (reason about).

Just need a better name than `CelDrawWrapper`.

See comment in scrollrt.cpp for how it can be used there.

Source/scrollrt.cpp
```diff
@@ -188,17 +188,13 @@ static void scrollrt_draw_cursor_item(const CelOutputBuffer &out)
 
 	const auto &sprite = GetInvItemSprite(pcurs);
 	const int frame = GetInvItemFrame(pcurs);
+	void (*drawingFun)(const CelOutputBuffer &, int, int, const CelSprite &, int) = &CelClippedDrawSafeTo;
 	if (pcurs >= CURSOR_FIRSTITEM) {
 		const auto &heldItem = plr[myplr].HoldItem;
```
glebm, line 193: If you follow my wrapper function suggestion, this would become:

```c++
	const auto &heldItem = plr[myplr].HoldItem;
	CelBlitOutlineTo(out, GetOutlineColor(heldItem, true), MouseX, MouseY + cursH - 1, sprite, frame, false);
}
CelDrawWrapper(!heldItem._iStatFlag, out, MouseX, MouseY + cursH - 1, sprite, frame);
```

Source/scrollrt.cpp
qndel: no

AJenbo (approved):

PR #1939, opened 2021-05-11
Remove PlayerStruct.actionFrame and enable ADL for Walking/Run

obligaron:
## Content

- Removes `PlayerStruct.actionFrame`
- Adjusts Walking Animation to start correctly with frame 1 and ends with frame 8
- Shows skipped frames for running Animation

<Details>
<Summary>Example Videos</Summary>

### V1.2.1
https://user-images.githubusercontent.com/25415264/117880996-018a7380-b2a9-11eb-8d07-d1376664902b.mp4

### PR
https://user-images.githubusercontent.com/25415264/117880970-fa636580-b2a8-11eb-92cd-1e0f6ecca901.mp4

### V1.2.1
https://user-images.githubusercontent.com/25415264/117881533-9d1be400-b2a9-11eb-9e6d-ba74dcadf939.mp4

### PR
https://user-images.githubusercontent.com/25415264/117881304-647c0a80-b2a9-11eb-8ca1-13e33d2f56b4.mp4

</Details>

## Notes

- Similar to #1781 this introduces a minor breaking change (one animation)
- It would be good if someone checks the change in tracks.cpp. I tested walking with held left mouse.

## Teaser

This pr enables Animation Distribution Logic and plays animations in the correct order and don't skip any frames (when running in town).
But we still don't update position (of the player and the camera). So this is stuttering a little.
This pr is the ground work to fix this.
Now we now correctly how much progress we have to reach the next tile (independent of the game ticks) and this is important to decouple updating the position from the game ticks (same as ADL).
Currently I have a prototyp based on this pr. See Example Video of Prototyp. 😉 

<Details>
<Summary>Example Video of Prototyp (5% Gamespeed)</Summary>

https://user-images.githubusercontent.com/25415264/117882710-e7519500-b2aa-11eb-9f45-758b88de8bf4.mp4

Note: You can see that the position is updated even if we don't show a new animation frame.

</Details>

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -421,7 +421,8 @@ static void LoadPlayer(LoadHelper *file, int p)
 	pPlayer->_pVar5 = file->nextLE<int32_t>();
 	pPlayer->position.offset2.x = file->nextLE<int32_t>();
 	pPlayer->position.offset2.y = file->nextLE<int32_t>();
-	pPlayer->actionFrame = file->nextLE<int32_t>();
+	// Skip actionFrame
+	file->skip(4);
 	for (uint8_t i = 0; i < giNumberOfLevels; i++)
 		pPlayer->_pLvlVisited[i] = file->nextBool8();
 	for (uint8_t i = 0; i < giNumberOfLevels; i++)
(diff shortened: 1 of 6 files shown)


NikoVP:
Great work! I hope the smoothing can somehow be applied to terrain to achieve true 60fps scrolling!

obligaron:
> 
> 
> Great work! I hope the smoothing can somehow be applied to terrain to achieve true 60fps scrolling!

Thanks 🙂 
I think as first step this would only be applied to player and the camera.
After that we have to first convert monsters and missiles to `AnimationInfo` (ADL) and after that it should be possible to use the same logic for them.
So it will not happen tomorrow but the important thing is we have a plan and solution. And this solution should work for pc and for old devices that can't render 60fps constantly.

AJenbo (approved):

AJenbo:
Applying similar logic to camera and player position would be fantastic. Really glad you are tackling all of this and cool that you found a way to do so in a piecemeal fashion :+1:

PR #1945, opened 2021-05-12
🔨 Fix Switch CI build

glebm:
The image was updated from Debian Stretch to Debian Buster in https://github.com/devkitPro/docker/commit/491845c3c3a2c9deff5ff1ad05152732bebb2d49

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,9 +37,6 @@ jobs:
     working_directory: ~/repo
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


AJenbo:
Nice, that can probably only be a good thing.

StephenCWills:
Does this also apply to the 3DS? I feel like it's having the same problem now.
https://app.circleci.com/pipelines/github/diasurgical/devilutionX/6324/workflows/ceacef5f-8a4d-40dc-8554-a95b18b76e73/jobs/45498

PR #1946, opened 2021-05-12
End of Poisoned Water Supply no longer ignores Blended Transparency setting

StephenCWills:
Because of the frequent palette changes that change the color of water at the end of Poisoned Water Supply, the blended lookup table get regenerated every frame for 32 frames. This logic wasn't checking the `bBlendedTransparancy` flag, which caused the framerate to plummet at the end of this quest on 3DS.


AJenbo:
Just in time for when we make blended transparancy the only option :D

PR #1948, opened 2021-05-12
🎉 SDL1: An option to render directly to video mem

glebm:
Now that we no longer have a buffer border, we can render directly to
the video memory if the output surface is 8-bit and double buffering is
enabled.

Also adds a flag to force this even when double buffering is disabled,
because some systems emulate 8-bit mode with something else, so the
output is always buffered.

This required a tweak to the cursor to make sure that we clear it after
`SDL_Flip`.


Source/scrollrt.cpp
```diff
@@ -143,7 +143,7 @@ static void BlitCursor(BYTE *dst, int dst_pitch, BYTE *src, int src_pitch)
 /**
  * @brief Remove the cursor from the buffer
  */
-static void scrollrt_draw_cursor_back_buffer(const CelOutputBuffer &out)
+static void ClearCursor(const CelOutputBuffer &out)
```
AJenbo, line 146: :+1:

Source/scrollrt.cpp
```diff
@@ -1550,18 +1551,19 @@ void DrawAndBlit()
...
 
 	DrawMain(hgt, ddsdesc, drawhpflag, drawmanaflag, drawsbarflag, drawbtnflag);
 
+	RenderPresent();
+
 	lock_buf(0);
-	scrollrt_draw_cursor_back_buffer(GlobalBackBuffer());
+	ClearCursor(GlobalBackBuffer());
 	unlock_buf(0);
-	RenderPresent();
```
AJenbo, line 1564: Did you test this both on SDL1 and SDL2?

glebm: Yeah, on SDL2 we clear the cursor from `pal_surface` (which is always a separate surface there) so the order doesn't matter

glebm: One issue is that if we're double-buffering in direct rendering mode, we're unable to clear the cursor from the back buffer -- but we can look into this later as the direct rendering feature is still experimental

Source/scrollrt.cpp
glebm: Oops, there is already a ClearCursor

Source/scrollrt.cpp
glebm: Fixed

Source/scrollrt.cpp
glebm: Fixed

AJenbo: that's good as well :)

PR #1951, opened 2021-05-13
Optimize blended transparency calculations at the end of Poisoned Water Supply

StephenCWills:
This significantly reduces the amount of work that has to be done when updating the blended transparency lookup at the end of the Poisoned Water Supply quest. Here is a summary of the changes that were made.

* Refactor part of `GenerateBlendedLookupTable()` into `FindBestMatchForColor()`. This way, we can pick and choose which blended pixel combinations we update in `palette_update_quest_palette()`.
* Disable call to `GenerateBlendedLookupTable()` in `LoadPalette()` when loading the clean water palette at the end of the quest. This prevents the blended transparency lookup from immediately turning blue. This was being overwritten in `palette_update_quest_palette()` anyway.
* Disable color cycling while water color is changing. With color cycling disabled, we can copy colors from the clean water palette to the logical palette one at a time. Cycled colors were being overwritten in `palette_update_quest_palette()` anyway.
* Update blended transparency for only the updated color and black.

Created as a draft PR for two reasons.

1. I hate the `extern int WaterDone;` in `CycleColors()`, especially considering how that value is being passed in to `palette_update_quest_palette()`. I'd like to search for a cleaner solution.
2. I hadn't imagined I'd be able to optimize this down to updating just a couple values in the blended transparency lookup. It might be worth considering updating blended pixels for all combinations that include the modified color. Doing so would provide more flexibility in the future in case we want to introduce fancy UI elements with transparent pixels that are not black.

### DIFF
--- a/Source/palette.cpp
+++ b/Source/palette.cpp
@@ -65,6 +65,26 @@ void palette_init()
 	InitPalette();
 }
 
+static Uint8 FindBestMatchForColor(SDL_Color *palette, SDL_Color color, int skipFrom, int skipTo)
+{
+	Uint8 best;
+	Uint32 bestDiff = SDL_MAX_UINT32;
+	for (int i = 0; i < 256; i++) {
+		if (i >= skipFrom && i <= skipTo)
+			continue;
+		int diffr = palette[i].r - color.r;
+		int diffg = palette[i].g - color.g;
+		int diffb = palette[i].b - color.b;
+		Uint32 diff = diffr * diffr + diffg * diffg + diffb * diffb;
+
+		if (bestDiff > diff) {
+			best = i;
+			bestDiff = diff;
+		}
+	}
+	return best;
+}
+
 /**
  * @brief Generate lookup table for transparency
  *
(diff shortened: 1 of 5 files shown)


Source/palette.cpp
```diff
@@ -271,6 +278,10 @@ void PaletteFadeOut(int fr)
  */
 static void CycleColors(int from, int to)
 {
+	extern int WaterDone;
+	if (WaterDone > 0)
+		return;
```
AJenbo, line 283: Instead of modifying a general function to have quest specific exception, either modify palette_update_caves or even better make a copy of it specifically for this quest. Maybe it can even handle more of the quest specific events there.

StephenCWills: Excellent suggestion. This new version is indeed much cleaner.

StephenCWills:
I think the code looks good now, but I'd still like to test on o3DS with blended transparency to see how much of an impact it would have to update all blended pixels for the updated color instead of limiting to just black. So there may be one more update coming.

AJenbo (approved):

StephenCWills:
The 3DS didn't even skip a beat. This should be ready for review and merge.

glebm (approved):

Source/palette.cpp
```diff
@@ -65,6 +65,26 @@ void palette_init()
 	InitPalette();
 }
 
+static Uint8 FindBestMatchForColor(SDL_Color *palette, SDL_Color &color, int skipFrom, int skipTo)
```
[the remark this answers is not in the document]
StephenCWills: Thanks! I realized `SDL_Color` is so small, I may as well just pass by value.

glebm (approved):

AJenbo (approved):

PR #1954, opened 2021-05-13
Put unit tests in a separate binary

### LINKED ISSUE #1836: 🔨  Tests should be in a separate binary target
Currently the tests are built into the same binary target, which means you either have tests enabled, or have a working binary

They should ideally be in a separate binary target

obligaron:
Fixes #1836 

Notes:

- Main source files will be compiled to a static library
- Compilation time isn't increased for unit tests as they use the same static library as the main executable
- Configuration in CMakeLists.txt isn't doubled cause of the static library
- Currently the unit tests aren't build per default. If wanted this can changed easily, but I wasn't sure if other platforms (for example 3DS) supports running with unit tests.
- I'm by far no CMake expert and I only tested this on Windows. :worried: So if this is the wrong approach or something is missing please state it.

@glebm It would be nice if you could take a look 🙂


AJenbo:
This fails to link under Linux and so the unit test fails in CI

3ds seams to have failed because of a random network error so can probably be ignored.

obligaron:
But it seems to only fail to link the coverage helper methods. :thinking:
It seems that the coverage is also added directly to the lib.

That's not good if we want to use the same lib for the unit test and the main executable.
If this is correct perhaps we should add a separate switch for the coverage?
This would allow to use the same lib for both (unit test and main executable) if we don't want code coverage.

AJenbo:
As long as we can still generate coverage when needed I'm fine with it.

CMakeLists.txt
```diff
@@ -507,7 +512,9 @@ if(RUN_TESTS)
     test/animationinfo_test.cpp)
 endif()
 
-add_executable(${BIN_TARGET} WIN32 MACOSX_BUNDLE ${devilutionx_SRCS})
+add_library(${LIB_TARGET} STATIC ${devilutionxlib_SRCS})
```
glebm, line 515: No need for LIB_TARGET and TEST_TARGET
We only have BIN_TARGET because for some platforms we have to produce an executable with a different name

CMakeLists.txt
```diff
@@ -296,7 +296,7 @@ add_library(PKWare STATIC
   3rdParty/PKWare/implode.cpp)
 target_include_directories(PKWare PUBLIC 3rdParty/PKWare)
 
-set(devilutionx_SRCS
+set(devilutionxlib_SRCS
```
glebm, line 299: Please use `libdevilutionx` for the library target as it is more conventional

glebm:
If we need to link the library with different flags for the test target, we can change it from a `STATIC` library to an `INTERFACE` library

CMakeLists.txt
```diff
@@ -535,13 +540,14 @@ if(RUN_TESTS)
   include(CTest)
   include(GoogleTest)
   find_package(GTest REQUIRED)
-  include_directories(${GTEST_INCLUDE_DIRS})
-  add_definitions(-DRUN_TESTS)
+  include_directories(devilutionxTests ${GTEST_INCLUDE_DIRS})
```
glebm, line 543: `target_include_directories`?

CMakeLists.txt
```diff
@@ -535,13 +540,14 @@ if(RUN_TESTS)
...
-  include_directories(${GTEST_INCLUDE_DIRS})
-  add_definitions(-DRUN_TESTS)
+  include_directories(devilutionxTests ${GTEST_INCLUDE_DIRS})
   if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
-    add_definitions(-fprofile-arcs -ftest-coverage)
+    target_compile_definitions(devilutionxTests -fprofile-arcs -ftest-coverage)
   endif()
-  target_link_libraries(${BIN_TARGET} PRIVATE -fprofile-arcs ${GTEST_LIBRARIES})
-  gtest_add_tests(${BIN_TARGET} "" AUTO)
+  add_executable(devilutionxTests WIN32 MACOSX_BUNDLE ${devilutionxtest_SRCS})
```
glebm, line 547: How about `devilutionx-tests`?

CMakeLists.txt
obligaron: Done 🙂

CMakeLists.txt
obligaron: Done 🙂

CMakeLists.txt
obligaron: Sorry that was a mistake. Fixed now 🙂

CMakeLists.txt
obligaron: Sure. Done 🙂

glebm (approved):

CMakeLists.txt
```diff
@@ -532,16 +537,17 @@ file(GENERATE OUTPUT ${CONFIG_PATH} CONTENT
...
+  add_executable(devilutionx-tests WIN32 MACOSX_BUNDLE ${devilutionxtest_SRCS})
   include(CTest)
   include(GoogleTest)
   find_package(GTest REQUIRED)
-  include_directories(${GTEST_INCLUDE_DIRS})
-  add_definitions(-DRUN_TESTS)
+  target_include_directories(devilutionx-tests PRIVATE ${GTEST_INCLUDE_DIRS})
   if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
-    add_definitions(-fprofile-arcs -ftest-coverage)
+    target_compile_definitions(devilutionx-tests PRIVATE -fprofile-arcs -ftest-coverage)
```
glebm, line 546: This looks wrong, and it was wrong before. It's defining a macro constant `-fprofile-arcs` etc.

Perhaps `target_compile_options`?

obligaron: 👍  Let's give it a try. I changed it.

glebm (dismissed):

test/main.cpp
```diff
@@ -0,0 +1,11 @@
+#include <SDL.h>
```
glebm, line 1: Is this include needed?

obligaron: Interesting.
Before it was needed cause SDL redefined main but now it isn't.
So just removed it 🙂

CMakeLists.txt
obligaron: This works now. Thanks 🙂

obligaron:
Yeah code coverage results are back :smile:
Per default it's not enabled (see previous comment) and can now be toggled with `ENABLE_CODECOVERAGE`.

glebm:
Awesome! :tada:

glebm (approved):

AJenbo (approved):

Issue #1962, opened 2021-05-13
"Dead Pixels" when trying to equip item


stefanmielke:
**Important information**
Master branch on 4cd1fd9b727e5ec059b447d674fa7e825e90c4a2

**Describe**
When trying to equip an item in the exact center of the slot it doesn't allow it to equip. Moving one pixel off center and it equips as expected.

**To Reproduce**
1. Create a new game as rogue
2. Click on the equipped weapon
3. Move the mouse cursor so that the equipment is in the exact center (it will not do anything)
4. Move the mouse cursor one pixel on any direction (it will equip correctly)

**Additional context**
_Originally posted by @glebm in https://github.com/diasurgical/devilutionX/issues/1885#issuecomment-840274498_

PR #1885, opened 2021-05-07
Controller Inventory Improvements

stefanmielke:
This PR is to bring inventory movement improvements when using a controller.

This should include fixes for #511

Here is a list of the changes:

* Removed code to find mouse position when using controller and mouse at the same time (too many issues when moving the cursor). Now it uses the last known slot selected with the controller.
* Only allow inventory movement to valid slots (eg.: rings can only go to ring slots).
* Warp from one side of the inventory to the other.
* Adjust mouse position on slots.
* Make the use and spell button equip items like shift click does.
* Move items to belt when using the spell button (use button still consumes the item when applicable).
* Move object as a whole item, not allowing movements that are not valid.
* Increased repeat interval to decrease 'false' moves (from 100ms to 150ms).
* Last 2 slots in the inventory now move to the last belt slot when moving down.
* Changed a lot of the code to try to make it more clear, but it is still a long function.

Known issues:

* When swapping items, even if the slot is not a valid place to put it, it keeps the item in it.
* When moving out of an item larger than 1x1, it moves to the top row of that item. It was made this way to not cause issues when picking up and dropping items.

Future enhancements:

* Option to 'save' cursor slot when reopening the inventory screen. Currently it uses the same 'slot' variable as other screens (like the character screen).

Remarks:

* I just fetched from latest, so combat is broken on this branch.

PR #1963, opened 2021-05-13
Adding offset to enable items being able to be equipped on lower resolutions

stefanmielke:
Fixes problem found by @glebm in https://github.com/diasurgical/devilutionX/issues/1885#issuecomment-840274498.

_Did not fix the underlying cause. Opened https://github.com/diasurgical/devilutionX/issues/1962 to fix that._

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -492,7 +492,7 @@ void AttrIncBtnSnap(AxisDirection dir)
 
 coords InvGetEquipSlotCoord(const inv_body_loc inv_slot)
 {
-	coords result { RIGHT_PANEL, 0 };
+	coords result { RIGHT_PANEL - 1, -1 };
 	result.x -= (icursW28 - 1) * (INV_SLOT_SIZE_PX / 2);
 	switch (inv_slot) {
 	case INVLOC_HEAD:


AJenbo (approved):

AJenbo:
fixing the underlying issue would be better, but since it is in the controller code I can accept this for now.

stefanmielke:
> fixing the underlying issue would be better, but since it is in the controller code I can accept this for now.

I did try to find the code that was causing the issue, but was unable to find where it is :(

glebm:
By the way, debugging this project with VS Code is very easy (no configuration required).
In VS Code, press Ctrl + Shift + P, then "CMake: Debug" to launch a debugger session

stefanmielke:
> By the way, debugging this project with VS Code is very easy (no configuration required).
> In VS Code, press Ctrl + Shift + P (which brings up the action menu), then "CMake: Debug" to launch a debugger session

I'm using CLion for Linux/Windows and VSCode only for PSP (CLion doesn't have a way for me to use psp-make instead). The problem was more that I tried to find a method with a name that would do it, but no breakpoint hit there (mostly tried on inv.cpp, but I'm not sure if that's the correct place).

glebm:
You could trace it down all the way from the mouse click handler

PR #1966, opened 2021-05-13
"Dead Pixels" when trying to equip item

### LINKED ISSUE #1962: "Dead Pixels" when trying to equip item

stefanmielke:
One pixel was not being calculated on the X axis when trying to find for an inventory slot to put the item.

Closes #1962 and should fix controller not being able to equip when on low resolutions.

@AJenbo @glebm

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -869,7 +869,7 @@ void CheckInvPaste(int pnum, int mx, int my)
 			yo = PANEL_TOP;
 		}
 
-		if (i >= InvRect[r].X + xo && i < InvRect[r].X + xo + INV_SLOT_SIZE_PX) {
+		if (i >= InvRect[r].X + xo && i <= InvRect[r].X + xo + INV_SLOT_SIZE_PX) {
 			if (j >= InvRect[r].Y + yo - INV_SLOT_SIZE_PX - 1 && j < InvRect[r].Y + yo) {
 				done = true;
 				r--;


glebm (approved):

Source/inv.cpp
```diff
@@ -869,7 +869,7 @@ void CheckInvPaste(int pnum, int mx, int my)
 			yo = PANEL_TOP;
 		}
 
-		if (i >= InvRect[r].X + xo && i < InvRect[r].X + xo + INV_SLOT_SIZE_PX) {
+		if (i >= InvRect[r].X + xo && i <= InvRect[r].X + xo + INV_SLOT_SIZE_PX) {
 			if (j >= InvRect[r].Y + yo - INV_SLOT_SIZE_PX - 1 && j < InvRect[r].Y + yo) {
```
glebm, line 873: Should the same be done for the Y coordinate?

glebm:
So this allows the middle of the cursor to be 1px beyond the right edge of the tile?
Seems weird that this is the fix but I don't really know how this works :shrug:

Source/inv.cpp
stefanmielke: I couldn't reproduce the same problem after applying the fix for X. I could do a few more tests if we want to be sure.

stefanmielke:
> So this allows the middle of the cursor to be 1px beyond the right edge of the tile?
> Seems weird that this is the fix but I don't really know how this works shrug

Basically mouseX was 1005, but it was only checking up to 1004 on the first slot, and starting from 1006 on the one right next to it.

glebm (commented):
Perhaps a slightly cleaner fix might be to change the calculation of the middle pixel of the cursor to:

```c++
i = mx + (icursW - 1) / 2;
j = my + (icursH - 1) / 2;
```

Imagine a 2px cursor. We could consider either the left or the right pixel as the "middle".

Previously it was the right pixel, with this change it'd be the left pixel

glebm:
> Basically mouseX was 1005, but it was only checking up to 1004 on the first slot, and starting from 1006 on the one right next to it. My guess is that it was discarding the border.

Ah, nevermind my previous comment then!

Source/inv.cpp
AJenbo: im gussing its already offset via -1

PR #1967, opened 2021-05-14
Cleanup items.cpp

vlolteanu:
Marking it as draft for the moment.

The changes were all mechanical, and hence, error-prone when all done at once. An extra pair of eyes would be really needed here.


vlolteanu:
- [ ] bc9d3b8c (HEAD -> refactor-items, dev/refactor-items) Cleanup SpawnBoy
- [ ] 02486cc9 Consolidate vendor inventory sorting
- [ ] 1a178cee Cleanup {Set,Put}ItemRecord
- [ ] ebd7fac6 Cleanup GetItemRecord
- [ ] 02420b4b Cleanup CreateMagicItem & friends
- [ ] 7382af1f Cleanup CreateSpellBook
- [ ] b80f8610 Cleanup RecalcStoreStats
- [ ] 5538c940 Cleanup RecreateHealerItem
- [ ] 281690b7 Cleanup SpawnHealer
- [ ] af93bda0 Cleanup SortHealer
- [ ] c54bdd07 Cleanup SpawnWitch
- [ ] ad41173a Cleanup WitchBookLevel
- [ ] b564df85 Cleanup SortWitch
- [ ] acc91a6d Cleanup WitchItemOk
- [ ] c889be87 Cleanup SpawnPremium
- [ ] 4f0d4915 Cleanup PremiumItemOk
- [ ] 9ca583c6 Cleanup SpawnSmith
- [ ] 88314c3d Fix DoRecharge
- [ ] 67777e8d Replace BubbleSwapItem with std:swap
- [ ] 0d732fac Consolidate RndFooItem
- [ ] 7330ce03 Cleanup SmithItemOk
- [ ] 48a36798 Cleanup StoreStatOk
- [ ] 5227d741 Cleanup DoOil
- [ ] 18c8c42c Cleanup OilItem
- [ ] a7ab3b59 Cleanup DoRecharge
- [ ] 8372a758 Cleanup GetItemStr
- [ ] f20718c8 Cleanup ProcessItems
- [ ] cf5f8275 Cleanup SpawnMapOfDoom, SpawnRuneBomb, SpawnTheodore
- [ ] a033bbe3 Cleanup SpawnRewardItem
- [ ] 0ee0fdf4 Cleanup SpawnQuestItem
- [ ] d3a03895 Cleanup items_427ABA
- [ ] 20f8651d Cleanup items_427A72
- [ ] 977e9edd Cleanup CreateTypeItem
- [ ] 904eeb7c Cleanup CreateRndUseful
- [ ] e8cf1ecb Cleanup CreateRndItem
- [ ] 21a9c7ec Cleanup SetupBaseItem
- [ ] ea939ba9 Cleanup SpawnItem
- [ ] e3b1aab5 Cleanup CheckUnique
- [ ] 98e764a4 Cleanup RndTypeItems
- [ ] 8199083f Clean RndAllItems
- [ ] b5c044f4 Cleanup RndUItem
- [ ] 60fca386 Cleanup GetItemPower
- [ ] 183f9223 Cleanup GetStaffPower
- [ ] d27098af Cleanup GetSuperItemSpace
- [ ] de0e8f27 Cleanup GetItemSpace
- [ ] 9d0d12a9 GetSuperItemLoc cleanup

vlolteanu:
The vendors' inventories are now sorted using std::sort. The items are sorted by item type, so this change of algorithm may  produce a slightly different item order from vanilla.

AJenbo:
> The vendors' inventories are now sorted using std::sort. The items are sorted by item type, so this change of algorithm may produce a slightly different item order from vanilla.

best to keep such changes separate from general cleanups so that thy can be merged independently.

qndel:
+ sorting griswold's premium items is a big nope as they are sorted by lvl

vlolteanu:
> +sorting griswold's premium items is a big nope as they are sorted by lvl

To be clear, the only functions that I touched did bubble sort based on item type. There can be multiple objects of the same type, and std::sort uses some combination of quicksort and something else when the vector is small, so the order might be different.

There was only one function related to Grisword's inventory, and it did perform bubble sort based on item type, so i guess that the code for the premium store resides somewhere else.

vlolteanu:
Is there a bubble sort in the STL?

AJenbo:
> > +sorting griswold's premium items is a big nope as they are sorted by lvl
> 
> To be clear, the only functions that I touched did bubble sort based on item type. There can be multiple objects of the same type, and std::sort uses some combination of quicksort and something else when the vector is small, so the order might be different.
> 
> There was only one function related to Grisword's inventory, and it did perform bubble sort based on item type, so i guess that the code for the premium store resides somewhere else.

If you are saying that the difference is the order of two items that would be considered equal by both sorting algorithms then i guess it's ok. Initially you made it sound like you added sorting where none had existed before.

StephenCWills:
Must have been the basic items, then. Just from my experience playing the game, that one does seem to be sorted by item type.

vlolteanu:
@AJenbo Yes, indeed. This is a conservative cleanup, and I was being straightforward about what has intentionally changed behavior-wise.

Source/objects.cpp
```diff
@@ -3810,11 +3806,11 @@ bool OperateShrineDivine(int pnum, int x, int y)
 		return false;
 
 	if (currlevel < 4) {
-		CreateTypeItem(x, y, false, ITYPE_MISC, IMISC_FULLMANA, false, true);
-		CreateTypeItem(x, y, false, ITYPE_MISC, IMISC_FULLHEAL, false, true);
+		CreateTypeItem({x, y}, false, ITYPE_MISC, IMISC_FULLMANA, false, true);
```
AJenbo, line 3809: Could you run clang-format to make sure that the code follow the project code style.

Source/items.h
```diff
@@ -387,27 +387,27 @@ void SetPlrHandGoldCurs(ItemStruct *h);
...
+void SpawnQuestItem(int itemid, Point pos, int randarea, int selflag);
 void SpawnRock();
-void SpawnRewardItem(int itemid, int xx, int yy);
-void SpawnMapOfDoom(int xx, int yy);
-void SpawnRuneBomb(int xx, int yy);
-void SpawnTheodore(int xx, int yy);
+void SpawnRewardItem(int itemid, Point pos);
+void SpawnMapOfDoom(Point pos);
+void SpawnRuneBomb(Point pos);
+void SpawnTheodore(Point pos);
```
AJenbo, line 410: Try avoiding abbreviation, it's not done in other places where Point is being used.

Source/items.cpp
```diff
@@ -4140,98 +4124,77 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
...
-		}
-		j--;
-	}
+	std::sort(items, items + count, cmp);
 }
 
 void SpawnSmith(int lvl)
 {
-	int i, iCnt, idata;
+	constexpr int Unsorted = 0;
```
AJenbo, line 4196: Maybe:
```suggestion
constexpr int PinnedItemCount = 0;
```

Source/items.cpp
```diff
@@ -4140,98 +4124,77 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
...
-	*b = h;
-}
-
-void SortSmith()
+void SortVendor(ItemStruct *items)
 {
-	int j, k;
-	bool sorted;
+	int count = 1;
+	while (!items[count].isEmpty())
```
AJenbo, line 4184: Are we sure this won't just run off the tracks if the vendor is full?

Source/items.cpp
```diff
@@ -3221,6 +3205,7 @@ static bool OilItem(ItemStruct *x, PlayerStruct *p)
 		break;
 	case IMISC_OILBSMTH:
 		if (x->_iMaxDur != 255) {
+			int dur;
```
AJenbo, line 3208: Avoid uninitialized values
```suggestion
			int dur = x->_iMaxDur + 1;
```
(a few of the lines below can then use `dur` instead of `x->_iMaxDur`)

Source/items.cpp
```diff
@@ -2795,7 +2785,7 @@ void items_427ABA(Point position)
 {
 	PkItemStruct PkSItem;
 
-	if (CornerStone.activated || position.x == 0 || position.y == 0) {
+	if (CornerStone.activated || position == Point{0, 0}) {
```
AJenbo, line 2788: This effectively change it from or to and, is this is intended?

Source/items.cpp
```diff
@@ -1373,40 +1374,41 @@ int AllocateItem()
...
+			ret.y = pos.y + j;
+			for (int i = -k; i <= k; i++) {
+				ret.x = i + pos.x;
+				if (ItemSpaceOk(ret.x, ret.y)) {
+					return ret;
 				}
 			}
 		}
 	}
+	return ret;
```
AJenbo, line 1411: This would probably be a fail condition. But not sure how we should express it, app_fatal seams a bit drastic.

Source/items.cpp
```diff
@@ -1342,19 +1343,19 @@ static bool GetItemSpace(int x, int y, int8_t inum)
...
+		yy++;
+		if (yy == 3)
+			yy = 0;
 	}
 
-	xx += x - 1;
-	yy += y - 1;
+	xx += pos.x - 1;
+	yy += pos.y - 1;
 	items[inum].position = { xx, yy };
```
AJenbo, line 1359: Would this make sens?
```suggestion
	items[inum].position = pos + Point { xx + 1, yy - 1 };
```

AJenbo (changes requested):
Looks good, but there where a couple of notes mostly related to code style.

Source/items.cpp
vlolteanu: It was a mistake. Fixed.

Source/items.cpp
vlolteanu: Agreed.

Source/items.cpp
vlolteanu: Maybe an assert?

AJenbo:
Looks like the test is failing, several items are morphing. Do you need any help with getting feedback from the test?

vlolteanu:
Please do. Or better yet, tell me how to run the tests myself. (I set RUN_TESTS to ON in CMakeLists.txt, but  that apparently didn't do anything.)

AJenbo:
If you are on linux or WSL:
```bash
apt-get install -y cmake curl g++ git lcov libgtest-dev libfmt-dev libsdl2-dev libsdl2-ttf-dev libsodium-dev
cmake .. -DRUN_TESTS=ON
make -j 4
./devilutionx
```
I usually create a `build-test` folder and run it from there.

I don't know what would be needed to do it on VC or Mac

Source/items.cpp
vlolteanu: That would break the logical flow of the code, even though the end result is still the same. I refactored it even further.

Source/items.cpp
```diff
@@ -4140,98 +4124,87 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
...
-	for (i = 1; AllItemsList[i].iLoc != ILOC_INVALID; i++) {
+	int ri = 0;
+	for (int i = 1; AllItemsList[i].iLoc != ILOC_INVALID; i++) {
 		if (!IsItemAvailable(i))
 			continue;
+		if (AllItemsList[i].iRnd == IDROP_NEVER)
+			continue;
+		if (!Ok(i))
+			continue;
+		if (AllItemsList[i].iMinMLvl < minlvl || AllItemsList[i].iMinMLvl > maxlvl)
```
AJenbo, line 4164: This seams a bit backwards. Items define what range they work with and evaluate against the given parameters. But here you are comparing `iMinMlvl` to a range as if it was the items level. It also looks like `minlvl` is always `0`, so this would probably make more sens as:
```suggestion
		if (AllItemsList[i].iMinMLvl > lvl)
```

Source/items.cpp
AJenbo: That might work, but it's a bit tricky since it's a state that is easily manipulated by the player.

AJenbo:
Looks like you found the issue :+1: Did you also manage to get the tests running? What system are you developing on?

Source/items.cpp
vlolteanu: SpawnPremium cares about the minimum level. Adria etc. don't.

AJenbo: Ok, so this is in preparation of further cleanups?

vlolteanu:
Yes, it was a mechanical error in RndVendorItem. "-DRUN_TESTS=ON" did the trick.

I'm working on Linux.

Source/items.cpp
vlolteanu: I've already done the cleanup: SpawnPremium also uses RndVendorItem now.

Source/items.cpp
vlolteanu: Agreed. Maybe the end-result is that some other item on the ground is overwritten.

Source/items.cpp
AJenbo: Arh, missed that detail :)

AJenbo (approved):

julealgon:
50 commits on a single PR seems excessive to me, and the fact that each commit is just "Cleanup X" is really, _really_ bad imho. We should be specific on what we are doing, like "Use Point struct for X/Y/Z" instead of "Cleanup".

vlolteanu:
I guess I'll open 50 PRs next time :)

I frankly expect this to get squashed.

I agree that the commits that did make a function use Point could have been worded differently, but it's a current policy to use Point wherever possible, so I'd count that as cleanup. Aside from that, all other Cleanup Foo commits were literally nothing more than cleanups of Foo, no matter how you slice it.

AJenbo:
> I frankly expect this to get squashed.

Oh, guess I don't have to bother with any rebasing then :+1:

AJenbo:
@vlolteanu you got an achievement:
image

The code now lives up to 3/10 out of ISO 25010 specifications for maintainable code.

vlolteanu:
Haha that was unexpected.

If I understand that correctly, this PR pushed devilutionX over the edge into 3/10, right?

AJenbo:
Correct, you removed the last 21 lines of duplication for us to hit 95% unique code. When we forked from Devilution it was at 89%

AJenbo:
The next mile stone is a lot further in to the future. Most likely it would be reducing the amount of code in functions with many parameters. This clean up did make a small jump on that graph as well, with some code in the 5-6 parameters tier dropping to 3-4 instead, probably due to `Point` being used instead of an `x`, `y` pair. But we will need to do 120x better to tick that box fully.

Splitting up long fictions (like what I did for `TalkToTowner()` and `OperateShrine()`) is probably what is going to make the biggest advances over all.

I don't know ho much access you guys have to the list, but if you like I can generate issues for some of the functions that are listed as prime candidates for refactoring.

Here's an examples of an issue generated for the 4 basic categories.
Long functions:
https://github.com/diasurgical/devilutionX/issues/1995
Complex functions:
https://github.com/diasurgical/devilutionX/issues/1996
Duplicate code:
https://github.com/diasurgical/devilutionX/issues/1997
Large interfaces:
https://github.com/diasurgical/devilutionX/issues/1998

There are also more advanced lists relating to code coupling. Dealing with that is usually much more complicated and it also tends to get worse as we work on the basic issues and code sharing inevitably leads to more coupling. In my experience dealing with this comes much later when the general issues are mostly death with and you can better start to reason about application structure.

PR #1968, opened 2021-05-14
fix typo

qndel:


Source/engine/render/cel_render.cpp
```diff
@@ -656,7 +656,7 @@ void CelDrawLightRedTo(const CelOutputBuffer &out, int sx, int sy, const CelSpri
 
 void CelDrawItem(bool usable, const CelOutputBuffer &out, int x, int y, const CelSprite &cel, int frame)
 {
-	if (usable) {
+	if (!usable) {
```
julealgon, line 659: This is not a typo though.

AJenbo: Maybe under a strict definition, fortunately he did change the commit message when merging: https://github.com/diasurgical/devilutionX/commit/610d5319bdbd7439d82e04c7f814e70ade61979c

PR #1969, opened 2021-05-14
♻️Use common text rendering for store menus

AJenbo:
This also corrects some minor alignment issues, like the margin being wider on the right then on the left:

image


glebm (approved):
Nice cleanup :+1:

PR #1974, opened 2021-05-14
Don't end the game after killing diablo

qndel:
He drops a soulstone now, using it triggers the ending cinematic
inspired by belzebub ;)

In itemdat.cpp I only added a line for soulstone before the last one but also reformatted it so labels aligned with stuff again

Can only use it on dlvl 16 and if diablo is dead.
Will trigger ending cinematic and quit the game for everyone on dlvl 16 at that time


julealgon:
Interesting idea! 
- Can you provide screenshots of what the soulstone looks like? 
- Is it intuitive that you can "use" it (does it have the "right click to use" hint)? 
- Does the soulstone go to the belt as well? 
- What happens if a different player takes the soulstone and uses it (say, for example, you host a game, kill Diablo, and then someone joins the game and takes it). 
- What happens if you take the soulstone, save the game, and then create a new game? Does it stay in your inventory? If so, what happens when you try to use it then? What happens when you get a new one from killing another instance of Diablo?

What exactly do we hope to achieve with this change? Being able to take the items in the ground before finishing the game? Anything else?

AJenbo:
> Can you provide screenshots of what the soulstone looks like?

reuses blood stone from valor quest

julealgon:
> reuses blood stone from valor quest

Ah that's a shame. I was hoping there was some unused soulstone asset in the game files.

FitzRoyX:
At the very least, it'll make it easier for modders to do the obvious (give Diablo loot drops, and the best ones at that).

Diablo has no corpse sprites, though, so there's some slight jank introduced from exploding him away (the movie shows him having a corpse). I suppose you could just have him spray blood on repeat until the soulstone is used, that'd be funny.

julealgon:
> Diablo has no corpse sprites, though

Then I don't see how we are going to make this work?

AJenbo:
Since this is inspired by Belzebub I'm guessing this might be helpful for visualization:
https://youtu.be/nkCxZ8pW_TU?t=1234

qndel:
> At the very least, it'll make it easier for modders to do the obvious (give Diablo loot drops, and the best ones at that).
> 
> Diablo has no corpse sprites, though, so there's some slight jank introduced from exploding him away (the movie shows him having a corpse). I suppose you could just have him spray blood on repeat until the soulstone is used, that'd be funny.

diablo drops loot ...

qndel:
> Interesting idea!
> 
> * Can you provide screenshots of what the soulstone looks like?
> * Is it intuitive that you can "use" it (does it have the "right click to use" hint)?
> * Does the soulstone go to the belt as well?
> * What happens if a different player takes the soulstone and uses it (say, for example, you host a game, kill Diablo, and then someone joins the game and takes it).
> * What happens if you take the soulstone, save the game, and then create a new game? Does it stay in your inventory? If so, what happens when you try to use it then? What happens when you get a new one from killing another instance of Diablo?
> 
> What exactly do we hope to achieve with this change? Being able to take the items in the ground before finishing the game? Anything else?

Soulstone is only usable if you are on dlvl 16 and diablo is dead.
If used not on dlvl 16, player says "that will not work here", if it's used on dlvl 16 while diablo is alive, player will say "no"
soulstone

julealgon:
The text in the item is immersion-breaking to me. I'd rather have the standard generic "right click to use" than a text that talks about 4th wall breaking, non-game stuff.

Also not fond of the idea of being able to carry the soulstone out of a game and into another one: it should be destroyed on new game IMHO.

Lastly, I'm not really fond of the idea of one player controlling the other's by using the stone and causing everyone to suddenly exit the game, though I don't have an immediate suggestion for an alternative.

Chance4us:
@qndel 
Would it be possible to just give the soul stone a different color so that it doesn't look exactly the same as the one from the Arcains Valor quest? By manipulating the color palette, for example, or something else.

qndel:
> @qndel
> Would it be possible to just give the soul stone a different color so that it doesn't look exactly the same as the one from the Arcains Valor quest? By manipulating the color palette, for example, or something else.

yup, but a PR for that would be 3x bigger than soulstone one ;)

FitzRoyX:
> Lastly, I'm not really fond of the idea of one player controlling the other's by using the stone and causing everyone to suddenly exit the game, though I don't have an immediate suggestion for an alternative.

For multiplayer in general, I feel like all quest-related items should not be pickup-able by anyone other than the host.

Chance4us:
> yup, but a PR for that would be 3x bigger than soulstone one ;)

Then let's do it 3x bigger, it would make the soul stone unique and special.

FitzRoyX:
> > yup, but a PR for that would be 3x bigger than soulstone one ;)
> 
> Then let's do it 3x bigger, it would make the soul stone unique and special.

Bloodstones, rubies (crafting mod), and diablo's soulstone have to be red for obvious reasons. There are only 3 bloodstones and unless you literally skip the last part of valor they're not going to be sharing the same inventory space. It's not a big deal, there won't be any confusion from this.

StephenCWills:
> Then let's do it 3x bigger, it would make the soul stone unique and special.

I'll also chime in here to say that I think if we're going to bother making this PR significantly bigger, we may as well go all the way and add assets to make the soulstone match the way it looks in the cutscene. Changing it from red only serves to differentiate it, but it feels to me like a whole lot of work for nothing if it doesn't actually look right.

Chance4us:
> I'll also chime in here to say that I think if we're going to bother making this PR significantly bigger, we may as well go all the way and add assets to make the soulstone match the way it looks in the cutscene. Changing it from red only serves to differentiate it, but it feels to me like a whole lot of work for nothing if it doesn't actually look right.

I agree. We could add assets or think of another solution as well. The suggestion to change the color was just one of the many possible ideas. I agree with better solutions. We also don't need to immediately make the soul stone look different.
This is also more of a luxury problem, I already think the current soul stone is already great.

PR #1979, opened 2021-05-15
Update 3rdParty/asio

glebm:
Updates ASIO to the current master, https://github.com/chriskohlhoff/asio/tree/57577c6db46a4e2de5351af2b185bf52696699a9

The only DevilutionX patch was Haiku support which has been upstreamed since.


qndel:
Doesn't solve the double error messages issue : P

AJenbo:
@qndel can you send your fix to https://github.com/chriskohlhoff/asio as a pr there :)

AJenbo:
@glebm how about pulling this in via CMake, it isn't even needed for all targets.

glebm:
@AJenbo I don't have time right now, someone else is welcome to

PR #559, opened 2020-01-28
Pixellight

qndel:
image

lighttest2

image

Fully working now :)

Done:
`support multiple colors`
`player light`
`missiles`
`static lights`
`fix static lights jumping around`
`deal with light being cut off on lower resolutions`
`fix first load have no static lights`
`stop the light system from affecting main menu etc.`
`make the fadein/fadeout when loading/going to/from main menu work properly`
`create a global table for missile color (and radius?)`
`add custom colors to missiles`
`deal with cursor having opaque rectangle around during fades`
`rewrite red on death/pause to work with pixel light`
`make light work during lag`
`stop quest area and normal level lights from interfering`
`save static lights with character`
`reset static light table between games`
`fix unique monster light jump around`
`fix light staying after unique monster teleports` (vanilla bug) - kind of fixed, even if broken, will try to follow the monster : P
`add light in town`
To do:
color explosion of missiles properly
~~`optimize performance`~~ Impossible till we rewrite all graphics to 32bit

PR #1987, opened 2021-05-16
ADL: Decouple positioning while walking from game logic

obligaron:
With this change the position of the player is now calculated independent from the game logic.
The already present information from `AnimationInfo` (ADL) is used for this.

## Restriction
Before this pr the player and camera position was updated 8 times for each walked tile.
With ADL we can calculate the offset relative to the animation fraction (0.0 to 1.0).
As this is a floating variable this can calculate many small steps and so positions updates.
But currently the rendering logic restricts this.

The current game logic operates on the pixel definitions of the original resolution.
For example the walking offset from one tile to another (to the east) are 64 pixels in x-axis.
As the current rendering logic is based on the game logic it's interface is also based on pixel basis.
That mean's currently we can calculate that we would be on pixel 8.4 but can't pass it to the rendering.
We have to squeeze the value in an integer and lose the precision.

This lose of precision is not dramatic if you walk in direct lines. cause we still have 64 pixels and means we have 64 different update positions. That's enough for smooth scrolling.
But in case of diagonal walking we only have 16 pixels at the y-axis. Compared to the original it's twice as much (8 -> 16) and better but still visible (see examples).

To lift this restriction the rendering logic must be adjusted to pass more fine granular positions.
The good thing is, this can happen at separate independent steps.

### Version 1.21x

https://user-images.githubusercontent.com/25415264/118393062-431a7600-b63d-11eb-9795-5ac4e5a5bc21.mp4

### PR

https://user-images.githubusercontent.com/25415264/118392923-64c72d80-b63c-11eb-96dd-952d6aac9f17.mp4

### Version 1.21x

https://user-images.githubusercontent.com/25415264/118392854-fda97900-b63b-11eb-84cb-cb6fce1b9b28.mp4

### PR

https://user-images.githubusercontent.com/25415264/118392859-07cb7780-b63c-11eb-8dde-4979296d9f95.mp4

</details>

### DIFF
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -52,6 +52,23 @@ int AnimationInfo::GetFrameToUseForRendering() const
 	return absoluteAnimationFrame;
 }
 
+float AnimationInfo::GetAnimationProgress() const
+{
+	if (RelevantFramesForDistributing <= 0) {
+		// This logic is used if animation distrubtion is not active (see GetFrameToUseForRendering).
+		// In this case the variables calculated with animation distribution are not initialized and we have to calculate them on the fly with the given informations.
+		float ticksPerFrame = (DelayLen + 1);
+		float totalTicksForCurrentAnimationSequence = gfProgressToNextGameTick + (float)CurrentFrame + (DelayCounter / ticksPerFrame);
+		float fAnimationFraction = totalTicksForCurrentAnimationSequence / ((float)NumberOfFrames * ticksPerFrame);
+		return fAnimationFraction;
+	}
+
+	float totalTicksForCurrentAnimationSequence = gfProgressToNextGameTick + (float)TicksSinceSequenceStarted;
+	float fProgressInAnimationFrames = totalTicksForCurrentAnimationSequence * TickModifier;
+	float fAnimationFraction = fProgressInAnimationFrames / (float)NumberOfFrames;
+	return fAnimationFraction;
+}
+
 void AnimationInfo::SetNewAnimation(byte *pData, int numberOfFrames, int delayLen, AnimationDistributionFlags flags /*= AnimationDistributionFlags::None*/, int numSkippedFrames /*= 0*/, int distributeFramesBeforeFrame /*= 0*/)
 {
 	if ((flags & AnimationDistributionFlags::RepeatedAction) == AnimationDistributionFlags::RepeatedAction && distributeFramesBeforeFrame != 0 && NumberOfFrames == numberOfFrames && CurrentFrame >= distributeFramesBeforeFrame && CurrentFrame != NumberOfFrames) {
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -124,6 +124,36 @@ const char *const szPlrModeAssert[] = {
 	"quitting"
 };
 
+Point GetOffsetForWalking(const AnimationInfo &animationInfo, const direction dir, bool cameraMode /*= false*/)
+{
+	// clang-format off
+	//                                  DIR_S,        DIR_SW,       DIR_W,	       DIR_NW,        DIR_N,        DIR_NE,        DIR_E,        DIR_SE,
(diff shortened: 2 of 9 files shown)


Source/engine/animationinfo.cpp
```diff
@@ -52,6 +52,23 @@ int AnimationInfo::GetFrameToUseForRendering() const
...
+		float ticksPerFrame = (DelayLen + 1);
+		float totalTicksForCurrentAnimationSequence = gfProgressToNextGameTick + (float)CurrentFrame + (DelayCounter / ticksPerFrame);
+		float fAnimationFraction = totalTicksForCurrentAnimationSequence / ((float)NumberOfFrames * ticksPerFrame);
+		return fAnimationFraction;
+	} else {
+		float totalTicksForCurrentAnimationSequence = gfProgressToNextGameTick + (float)TicksSinceSequenceStarted;
+		float fProgressInAnimationFrames = totalTicksForCurrentAnimationSequence * TickModifier;
+		float fAnimationFraction = fProgressInAnimationFrames / (float)NumberOfFrames;
+		return fAnimationFraction;
+	}
```
AJenbo, line 69:
```suggestion
	}


	float totalTicksForCurrentAnimationSequence = gfProgressToNextGameTick + (float)TicksSinceSequenceStarted;
	float fProgressInAnimationFrames = totalTicksForCurrentAnimationSequence * TickModifier;
	float fAnimationFraction = fProgressInAnimationFrames / (float)NumberOfFrames;
	return fAnimationFraction;
```

Source/scrollrt.cpp
```diff
@@ -677,8 +715,12 @@ static void DrawPlayerHelper(const CelOutputBuffer &out, int x, int y, int sx, i
 	}
 
 	PlayerStruct *pPlayer = &plr[p];
-	int px = sx + pPlayer->position.offset.x - CalculateWidth2(pPlayer->_pAnimWidth);
-	int py = sy + pPlayer->position.offset.y;
+	Point offset = ScrollInfo.offset;
+	if (pPlayer->IsWalking()) {
+		offset = GetOffsetForWalking(pPlayer->AnimInfo, pPlayer->_pdir);
```
AJenbo, line 720: this also needs to be fed into `CheckCursMove()` for mouse clicks to happen correctly.

Source/cursor.cpp::278

AJenbo:
This really plays and feels fantastic, grate work.

One funny thing is that since this is a gradual process you get to see the monsters move around at the lower update rate, it feels like they are stuck on old hardware :D

We might should also consider updating the light positions as a result of the render event rather then game ticks, but I think that can happen via another PR, not necessarily tied to ADL changes.

Source/engine/animationinfo.cpp
obligaron: Fixed 🙂

Source/scrollrt.cpp
obligaron: Done. 🙂 
I checked that monsters are highlighted while walking.

obligaron:
> This really plays and feels fantastic, grate work.
> 

Thanks for the feedback. This is really motivating. 😊 

> One funny thing is that since this is a gradual process you get to see the monsters move around at the lower update rate, it feels like they are stuck on old hardware :D
> 

🤣 The monsters will get faster too! Prepare yourself! :dagger::shield:

> We might should also consider updating the light positions as a result of the render event rather then game ticks, but I think that can happen via another PR, not necessarily tied to ADL changes.

The lightning part would be interesting. But i never looked at the logic.
I only saw qndel pixellight pr #559 that looked good or FluffyQuacks [lightmapping](https://www.youtube.com/watch?v=teFeNOrIAR8) also looks good. I think that's a part where the game could be improved a lot. 👍

AJenbo (approved):

julealgon:
This feels incredible. It might be the most impactful change to the game thus far in terms of smoothness in my opinion along with the 60FPS cursor movement. Fantastic work @obligaron !

Chance4us:
Really fantastic graphical revolution.
By the way, I noticed the following too and I keep my fingers crossed that the monsters will also become smooth.

> One funny thing is that since this is a gradual process you get to see the monsters move around at the lower update rate, it
> feels like they are stuck on old hardware :D

julealgon:
@Chance4us 
> and I keep my fingers crossed that the monsters will also become smooth.

They will 😉

PR #2002, opened 2021-05-17
Fix gkd350h build

glebm:
Works around a GCC bug on some platforms:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97044

### DIFF
--- a/CMake/gkd350h_defs.cmake
+++ b/CMake/gkd350h_defs.cmake
@@ -1,6 +1,10 @@
 set(NONET ON)
 set(USE_SDL1 ON)
 
+# Do not warn about unknown attributes, such as [[nodiscard]].
+# As this build uses an older compiler, there are lots of them.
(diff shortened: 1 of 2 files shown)


CMake/gkd350h_defs.cmake
```diff
@@ -1,6 +1,10 @@
 set(NONET ON)
 set(USE_SDL1 ON)
 
+# Do not warn about unknown attributes, such as [[nodiscard]].
+# As this build uses an older compiler, there are lots of them.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-attributes")
```
AJenbo, line 6: Maybe also apply this to the amiga builds as its also spewing out warnings about unknown attributes.

AJenbo: https://github.com/diasurgical/devilutionX/pull/2009

PR #2004, opened 2021-05-17
Fix movement with hold mouse button and stand animation

### LINKED ISSUE #1993: Delays in movement from standstill while click is held
**Important information**
latest build

**Describe**
Holding click at your character's feet to stop creates a 1 second delay before movement can start again. Julealgon suspects that the delay is equal to the idle animation cycle completing. This appears to be a byproduct of recent animation changes by @obligaron.

**To Reproduce**
1. Click and hold character's tile
2. Move cursor away from tile

**Expected behavior**
Movement away from current tile should be instant when click is held

obligaron:
Fixes #1993

Condition before my changes:
[`if (plr[myplr].actionFrame <= 6 && plr[myplr]._pmode != PM_STAND)`](https://github.com/diasurgical/devilutionX/blob/23cec61f155961918460fe5d896e686ed4249e0c/Source/track.cpp#L29)
Note: `actionFrame` was only used for walking and spell casting. In all other cases it was 0.

Buggy condition:
`if (plr[myplr].AnimInfo.GetFrameToUseForRendering() <= 6 || (!plr[myplr].IsWalking() && plr[myplr]._pmode != PM_STAND))`
This returns also if we are at the first 6 frames of the stand animation (the delay mentioned in #1993). So we don't track the changed mouse cursor in that case => Bug. 😉 

This is the new condition:
`if (player._pmode != PM_STAND && !(player.IsWalking() && player.AnimInfo.GetFrameToUseForRendering() >= 6))`
We check if we don't stand or are at the last walking animation frames.

### DIFF
--- a/Source/track.cpp
+++ b/Source/track.cpp
@@ -26,10 +26,12 @@ void track_process()
 	if (cursmx < 0 || cursmx >= MAXDUNX - 1 || cursmy < 0 || cursmy >= MAXDUNY - 1)
 		return;
 
-	if (plr[myplr].AnimInfo.GetFrameToUseForRendering() <= 6 || (!plr[myplr].IsWalking() && plr[myplr]._pmode != PM_STAND))
+	const auto &player = plr[myplr];
+
+	if (player.AnimInfo.GetFrameToUseForRendering() <= 6 || (!player.IsWalking() && player._pmode != PM_STAND))
 		return;
 
-	const Point target = plr[myplr].GetTargetPosition();
+	const Point target = player.GetTargetPosition();
 	if (cursmx != target.x || cursmy != target.y) {
 		Uint32 tick = SDL_GetTicks();
 		if ((int)(tick - sgdwLastWalk) >= gnTickDelay * 6) {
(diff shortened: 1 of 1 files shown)


AJenbo:
Shoudn't it be ` > 6`?

Source/track.cpp
```diff
@@ -26,10 +26,12 @@ void track_process()
 	if (cursmx < 0 || cursmx >= MAXDUNX - 1 || cursmy < 0 || cursmy >= MAXDUNY - 1)
 		return;
 
-	if (plr[myplr].AnimInfo.GetFrameToUseForRendering() <= 6 || (!plr[myplr].IsWalking() && plr[myplr]._pmode != PM_STAND))
+	auto &player = plr[myplr];
+
+	if (player._pmode != PM_STAND && !(player.IsWalking() && player.AnimInfo.GetFrameToUseForRendering() >= 6))
```
AJenbo, line 31: The inverse of `<= 6` is `> 6`
```suggestion
	if (player._pmode != PM_STAND && !(player.IsWalking() && player.AnimInfo.GetFrameToUseForRendering() > 6))
```

obligaron: thanks. fixed 🙂

PR #2014, opened 2021-05-19
Fix camera jerking when changing direction on controller

### LINKED ISSUE #2011: Camera/character jerks when playing with controller
All you need to do to reproduce is change direction mid-stride using controller. This definitely started after the recent ADL changes, but given that it can't be reproduced with the mouse controls, the fix for this may end up being unrelated to ADL.

Here is a video of the issue.

https://user-images.githubusercontent.com/9203145/118727673-1e6efb80-b801-11eb-8665-959aeae9dde6.mp4

Here is a "before" video for comparison, in case it's relevant. Something I always noticed with controller is how the direction on the map updates immediately based on controller inputs despite the character still walking in the same direction.

https://user-images.githubusercontent.com/9203145/118728614-7b1ee600-b802-11eb-8a52-045184cedc9a.mp4

StephenCWills:
Digging into past commits, I found the reason why the logic for gamepad controls was fiddling with the player's direction. It allows the player to change the direction their character is facing even if their movement is blocked. This is especially important because it's what enables the player to change targets when their character can't otherwise move.

I considered allowing players to change direction at any time so long as they aren't walking. Some additional testing revealed there were situations where the player can change targets in the middle of an attack. It seems reasonable to want to change the character's target in the middle of a series of queued attacks, but it's important to wait for the hit frame so the player doesn't end up hitting an enemy they didn't swing at. Furthermore, it's easy to imagine that there are some states in which the player probably shouldn't be allowed to change direction, such as blocking and hit recovery.

This resolves #2011

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1078,6 +1078,19 @@ bool IsPathBlocked(int x, int y, int dir)
 	return !PosOkPlayer(myplr, d1x, d1y) && !PosOkPlayer(myplr, d2x, d2y);
 }
 
+bool CanChangeDirection(const PlayerStruct &player)
+{
+	if (player._pmode == PM_STAND)
+		return true;
+	if (player._pmode == PM_ATTACK && player.AnimInfo.CurrentFrame > player._pAFNum)
+		return true;
+	if (player._pmode == PM_RATTACK && player.AnimInfo.CurrentFrame > player._pAFNum)
+		return true;
+	if (player._pmode == PM_SPELL && player.AnimInfo.CurrentFrame > player._pSFNum)
+		return true;
+	return false;
+}
+
 void WalkInDir(int playerId, AxisDirection dir)
 {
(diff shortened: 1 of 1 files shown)


obligaron:
Could we avoid setting `_pdir` entirely? :thinking:
Cause it should always be in sync with the current animation (what we render).

Note: I don't own a controller for testing. :sweat: and thanks for the fix 🙂

StephenCWills:
> Could we avoid setting `_pdir` entirely? 🤔
> Cause it should always be in sync with the current animation (what we render).
> 
> Note: I don't own a controller for testing. 😓 and thanks for the fix 🙂

I explained in the PR why setting `_pdir` is necessary. The solution there does still update `_pdir` prematurely during attack animations, but it doesn't appear to cause any significant problems. Perhaps I could look into updating targeting logic so it doesn't depend on `_pdir`, and that may prove to be a better solution. I think it will still be necessary to update `_pdir` during standing animations, but this seems to update the way the player is facing instantaneously anyway.

PR #2017, opened 2021-05-19
🌐 Italian localization enhancements

M0Rf30:
Hereby some fixes here and there
Regards


AJenbo:
Thanks, note that there will be a few further updates to the text in the coming week. We will try to apply the same changes to the translations, but it is always good to have a speaker go over it 👍

PR #2023, opened 2021-05-20
Get ASIO via CMake, plus options for disabling TCP and ZeroTier individually

### LINKED ISSUE #1982: Get ASIO via CMake
To make it easier to cache the dependency and have less in repo dependencies it would be good to move AISO out and instead get it via CMake like we do most other dependencies.

This will also make it easy to upgrade with out causing thousands of lines of change.

See:
https://github.com/diasurgical/devilutionX/pull/1979

StephenCWills:
Should be pretty self-explanatory. The only weird thing I encountered was the fact that `shlwapi` wasn't explicitly linked to `libdevilutionx` and was instead getting it indirectly from `zt-static`. This was actually already causing problems for `NONET` builds on Windows, which is probably just something no one ever does.

These could be split into separate PRs, however there is somewhat of a dependency between these two commits. If you'd like me to split them, let me know and I'll hold onto the second commit until the first one gets merged.

This resolves #1982


AJenbo:
Wellcome to the negative commit diff club

AJenbo (approved):

PR #2026, opened 2021-05-20
Draw Reflect Icon similiar to Manashield

### LINKED ISSUE #1989: Stuttery Reflect overlay while moving
**Important information**
Windows 10, DevilutionX latest commit 55fa72d1

**Describe**
Rendering of Mana Shield and Reflect overlays are inconsistent: while Mana Shield properly follows the hero using the same update frequency and position as the character, Reflect appears to be using a different logic that results in jittery motion.

**To Reproduce**
1. Cast Mana Shield
2. Cast Reflect
3. Move and observe the difference in motion

(hero with both spells for testing)
- [single_7.zip](https://github.com/diasurgical/devilutionX/files/6490015/single_7.zip)

**Expected behavior**
Reflect should be rendered using the same approach used for Mana Shield, resulting in a smooth motion.

**Screenshots**
ReflectManaShieldOverlayInconsistency

(It actually looks worse the higher the framerate is. This gif makes it look "less bad" than it is)

**Additional context**
Discovered while testing #1971. They might be related in some form.

Tested on original Hellfire and cannot repro the issue there. I assume it's related to the increase in FPS in DevilutionX?

obligaron:
Fix #1989 

<details><summary>Examples</summary>

### Master

ReflectVisiblePlayerNotVisibleMaster

ReflectPlayerNotVisibleOtherMaster

### PR

ReflectPlayerNotVisiblePR

ReflectPlayerNotVisibleOtherPR

### Master

ReflectBrightnessMaster

### PR

ReflectBrightnessPR

ReflectBrightnessPR2

</details>

Notes:

- Resolves stutter
- Resolves reflect icon is drawn even if player is not visible
- Resolves reflect icon is rendered too bright even if player is far away
- Old position updating logic isn't removed


AJenbo:
Why not remove the old position code?

AJenbo (approved):

obligaron:
> 
> 
> Why not remove the old position code?

too lazy for today 😁 

We could remove it but I would need to check if there are some vanilla compatibility issues or not (I think there are none).
The same goes for the player positioning logic. They should both be cleaned up later, if the changes prove to be good and correct.
We did the same with the animation skipping logic. First test, later remove.
But if you want I can include the removing in this pr. 🙂

obligaron:
The original position logic for mana shield is also intact? 🧐 
Should it also be removed? 🤔

FitzRoyX:
Would it be better to name these functions BuffDebuffIndicators instead of PlayerIcons? Just considering the possibility of mods adding auras and other things that don't necessarily go above the player's head or look like static icons.

AJenbo:
> Would it be better to name these functions BuffDebuffIndicators instead of PlayerIcons? Just considering the possibility of mods adding auras and other things that don't necessarily go above the player's head or look like static icons.

Doing so would be premature, we cannot know what code may or may not be used for in the future. It may well require changes to do this, and it could  turn out that this function isn't suitable for doing so at all.

AJenbo:
> The original position logic for mana shield is also intact? monocle_face
> Should it also be removed? thinking

Yeah seams like a good time to do so :)

obligaron:
old positioning logic is removed 😉

AJenbo (approved):

julealgon:
I didn't know about the brightness issue, nice! This is such a great refactor.

PR #2029, opened 2021-05-20
Changing Oil of Death application

thebigMuh:
See companion discussion #2022 

This changes Oil of Death as follows:
- The base functionality remains as previously - if MaxDmg - MinDmg < 30, it will raise MinDmg by 1 and MaxDmg by 2.
- It allows applying Oil of Death even when MaxDmg - MinDmg >= 30. In that case it will calculate the damage increase as an upgrade, bumping MinDmg and MaxDmg by 1.
- Regardless of how many Oil of Sharpness or Weird Shrines have been used, the total number of applications remains the same.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3199,12 +3199,10 @@ static bool OilItem(ItemStruct *x, PlayerStruct *p)
 		if (x->_iMaxDam - x->_iMinDam < 30)
 			x->_iMaxDam++;
 
-		auto &tempItem = items[MAXITEMS];
-		tempItem = {};
-		RecreateItem(MAXITEMS, x->IDidx, x->_iCreateInfo, x->_iSeed, x->_iIvalue, (x->dwBuff & CF_HELLFIRE) != 0);
+		const auto &orgItem = AllItemsList[x->IDidx];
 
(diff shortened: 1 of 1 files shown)


thebigMuh:
Btw - I'm not sure if I actually have to call RecreateItem, or if it would be enough to lookup the item using its ID in the AllItemsList, and fetch the min/max damage from there. The current implementation is the 'better safe than sorry' version.

AJenbo:
> Btw - I'm not sure if I actually have to call RecreateItem, or if it would be enough to lookup the item using its ID in the AllItemsList, and fetch the min/max damage from there. The current implementation is the 'better safe than sorry' version.

AllItemsList is a lot safer then recreating the item. If the item was migrated from hellfire to diablo i would morph.

thebigMuh:
> AllItemsList is a lot safer then recreating the item. If the item was migrated from hellfire to diablo i would morph.

Well, the oils are Hellfire only, so that shouldn't be an issue. But I'm changing it nonetheless, it's far simpler code that way.

strich:
Hi @thebigMuh, just checking in on the status of this PR. Are you still planning to get it in? If so you'll need to rebase it before a review can be done.

PR #2034, opened 2021-05-22
🧹 A couple of minor cleanups

glebm:


Source/DiabloUI/errorart.cpp
```diff
@@ -2,7 +2,7 @@
 
 namespace devilution {
 
-const unsigned char btnData[] = {
+const std::uint8_t btnData[] = {
```
julealgon, line 5: Nitpick: this is not a BYTE -> uint8 conversion

Source/dx.cpp
```diff
@@ -185,8 +185,10 @@ void dx_cleanup()
 	pal_surface = nullptr;
 	SDL_FreePalette(palette);
 	SDL_FreeSurface(renderer_texture_surface);
+#ifndef USE_SDL1
```
julealgon, line 188: Why? Unrelated?

Source/utils/stubs.h
```diff
@@ -1,23 +1,21 @@
...
-#include <cstdlib>
+#include "utils/log.hpp"
 
-#define UNIMPLEMENTED()                                                         \
-	do {                                                                        \
-		SDL_Log("UNIMPLEMENTED: %s @ %s:%i", __FUNCTION__, __FILE__, __LINE__); \
-		abort();                                                                \
+#define UNIMPLEMENTED()                                                                        \
+	do {                                                                                       \
+		::devilution::LogDebug("UNIMPLEMENTED: {} @ {}:{}", __FUNCTION__, __FILE__, __LINE__); \
```
julealgon, line 7: Again, this seems completely unrelated to Byte -> uint8 conversion cleanup to me.

Source/items.h
```diff
@@ -348,6 +349,20 @@ struct ItemStruct {
 			return false;
 		}
 	}
+
+	UiFlags getTextColor() const
```
julealgon, line 353: I know this is a simple "move" refactor, but this whole method seems like a tell-dont-ask violation to me. Shouldn't we just have a "renderToText" function of some sort that does the rendering and then make the color decision an implementation detail?

One could even argue that having display logic in the item class is an SRP violation as well. Ideally, logic and display concerns should not be in the same class in the first place. Have you thought about creating a different abstraction to do put this logic in?

Source/items.h
```diff
@@ -348,6 +349,20 @@ struct ItemStruct {
...
+		if (!_iStatFlag)
+			return UIS_RED;
+		switch (_iMagical) {
+		case ITEM_QUALITY_MAGIC:
+			return UIS_BLUE;
+		case ITEM_QUALITY_UNIQUE:
+			return UIS_GOLD;
+		default:
+			return UIS_SILVER;
+		}
```
julealgon, line 364: Random question: is there no pattern matching proposal for C++? This would look so much more clean:
```csharp
return item switch
{
    { _iStatFlag: false } => UIS_RED,
    { _iMagical: ITEM_QUALITY_MAGIC } => UIS_BLUE,
    { _iMagical: ITEM_QUALITY_UNIQUE } => UIS_GOLD,
    _ => UIS_SILVER,    
};
```

Source/items.h
julealgon, line 353: Aren't we using PascalCase for method names?

julealgon, line 353: Can you add the doxygen comments for the method?

julealgon, line 353: Why is this implemented in the header instead of in the .cpp? Can you move the implementation and leave only the prototype here?

julealgon (changes requested):

Source/dx.cpp
glebm: See commit message

Source/items.h
glebm: I thought about it but it's extra work -- for a later refactor. It's a minor thing too because these colors are the same throughout all Diablo games and should be consistent throughout the UI.

Source/items.h
glebm: Yes there is a proposal like that

Source/items.h
glebm: Because it's easier for the compiler to inline (e.g. even without LTO) and this function should be inlined.

glebm: Can you provide some text for the comment? Not sure what you want to be clarified by a comment.

Source/DiabloUI/errorart.cpp
glebm: It is part of the same cleanup.

Source/items.h
glebm: `ClassMethodCase` is not defined in `.clang-tidy`.
There is `isJewelry` and `isHelm` in the same class, so I followed that.

Source/utils/stubs.h
glebm: See commit message - cyclical include chain without this change (and we want each commit to be buildable for easier bisects)

Source/items.h
julealgon: I was under the impression we were adding comments even to obvious method for consistency/convention. If not, feel free to disregard.

julealgon: I find these selective rules really, _really_ bad. I'll leave it at that.

Source/utils/stubs.h
julealgon: Wouldn't it be possible to make these changes before the byte change, then? The fact they are "needed" doesn't preclude them being done _prior_ as their own commit IMHO.

Source/items.h
julealgon: Can we look into this @AJenbo ? Again, the earlier we decide on the format, the less changes there will be.

AJenbo: Not to obvious stuff I think. Though that can be subjective.

Source/utils/stubs.h
glebm: Yeah it'd certainly be cleaner in its own commit but extra work for little gain -- it's not difficult to review this PR.

AJenbo (approved):
Sold, can't wait to rebase my current refactor on this...

PR #2039, opened 2021-05-23
Hero speech function improvements

julealgon:
This PR introduces a new enum to index into hero speech as well as fixes a typo on "speach" throughout the codebase.

The enum members will be named in a subsequent change.


AJenbo:
Didn't we decide to rename it to `Say()`? That's why I haven't bothered applying the correct spelling to it :)

julealgon:
> Didn't we decide to rename it to `Say()` or something like that? That's why I haven't bothered applying the correct spelling to it :)

I actually completely forgot about that. I can rename now as part of this.

Source/effects.h
```diff
@@ -12,6 +12,112 @@
 
 namespace devilution {
 
+enum class HeroSpeech {
+	_1,
```
glebm, line 16: Well this made me look up which identifiers are reserved in C++: https://en.cppreference.com/w/cpp/language/identifiers

Let's replace the `_` with `N` or `I` to avoid raising eyebrows?

Source/player.h
```diff
@@ -369,20 +369,19 @@ struct PlayerStruct {
...
+	 * @brief Says a speech line.
+	 * @todo BUGFIX Prevent more then one speech to be played at a time (reject new requests).
 	 */
-	void PlaySpeach(int speachId) const;
+	void Say(HeroSpeech speechId) const;
 	/**
-	 * @brief Play a speach file after a given delay.
-	 * @param delay Multiple of 20ms waith before starting the speach
+	 * @brief Says a speech line after a given delay.
+	 * @param delay Multiple of 20ms waith before starting the speech
```
glebm, line 378:
```suggestion
	 * @param delay Multiple of 20ms wait before starting the speech

```

glebm (approved):

Source/player.h
AJenbo: Also it's 50ms not 20...

glebm (approved):

PR #2043, opened 2021-05-23
Leverage `CelSprite` for `PlayerStruct`

obligaron:
This was more annoying then I initially thought. :tired_face:

Notes:

- Only refactoring. No behavior or animation changes.
- `CelSprite` constructor requires both the animation data (byte) and the animation width. But in `PlayerStruct` width was set later (`SetPlrAnims`) and separated from the animation data loading (`LoadPlrGFX`). So the logic had to be adjusted.
- Currently the `CelSprite` constructor with `unique_ptr` is not used, cause I don't know yet how the multiple animation groups (one group per direction) in a cl2 file are handled correctly with `CelSprite`. But this can be changed easily (in this pr or following).
- Should unblock #1795
- Further refactoring is possible but this PR is big enough. 😉


Source/player.cpp
```diff
@@ -561,67 +577,28 @@ static DWORD GetPlrGFXSize(HeroClass c, const char *szCel)
...
-	player._pDData = new byte[GetPlrGFXSize(c, "DT")];
-
-	// BLOCK
-	player._pBData = new byte[GetPlrGFXSize(c, "BL")];
-
-	player._pGFXLoad = 0;
+	player.AnimInfo.pCelSprite = nullptr;
+	for (auto &animData : player.AnimationData) {
+		for (auto &celSprite : animData.CelSpritesForDirections)
+			celSprite.reset();
```
AJenbo, line 585: GCC 6.3 is failing on this:
```
/root/repo/Source/player.cpp: In function 'void devilution::ResetPlayerGFX(devilution::PlayerStruct&)':
/root/repo/Source/player.cpp:585:14: error: 'class std::experimental::fundamentals_v1::optional<devilution::CelSprite>' has no member named 'reset'; did you mean '_M_reset'?
    celSprite.reset()
```

obligaron: I'm working on it atm 😉

obligaron: now it compiles 🙂

julealgon:
PR title looks very misleading to me. It gives the impression that we are "replacing" `PlayerStruct` with `CelSprite` or something along those lines which doesn't make a lot of sense.

I'd recommend renaming to avoid confusion. Perhaps (if I get the meaning right) this would make sense:
- `Leverage 'CelSprite' for 'PlayerStruct'`

obligaron:
> I'd recommend renaming to avoid confusion. Perhaps (if I get the meaning right) this would make sense:
> 
>     * `Leverage 'CelSprite' for 'PlayerStruct'`

Renamed 😉

Source/engine.cpp
```diff
@@ -228,7 +228,8 @@ int32_t GenerateRnd(int32_t v)
 size_t GetFileSize(const char *pszName)
 {
 	HANDLE file;
-	SFileOpenFile(pszName, &file);
+	if (!SFileOpenFile(pszName, &file))
```
glebm, line 231: Are we now passing filenames for files that do not exist?

glebm:
Worth noting: The previous code allocated each type of graphics once to the maximum size of the graphic for a given player class. Then, when switching graphics, there was no allocation as the data was simply read into the pre-allocated buffer.

I'm not sure whether this has any performance implications for slow devices (e.g. a slight delay when switching armor?). Probably not, but worth noting.

Source/player.cpp
```diff
@@ -367,27 +367,74 @@ bool PlayerStruct::IsWalking() const
...
+	default:
+		Log("GetAnimationWidth: Unkown graphic {}", graphic);
+		return 96;
 	}
 }
 
-void LoadPlrGFX(int pnum, player_graphic gfxflag)
+void SetPlayerGPtrs(const char *path, std::unique_ptr<byte[]> &data, std::array<std::optional<CelSprite>, 8> &anim, int width)
+{
+	data = std::move(LoadFileInMem(path));
```
glebm, line 425: Set `data = nullptr;` first to reduce the maximum memory usage.

glebm: No need for `std::move` here, it's an rvalue

obligaron: Done 🙂

Source/engine.cpp
obligaron: No this is only needed for the unittests.
They now also try to load the animations, cause it's now also part of `NewPlrAnim`.
And without this conditions they would crash later.

If you want I could change the condition to also check `gbQuietMode`.

Source/loadsave.cpp
```diff
@@ -1418,32 +1429,33 @@ static void SavePlayer(SaveHelper *file, int p)
...
 	file->writeLE<int32_t>(pPlayer->_pAFrames);
-	file->writeLE<int32_t>(pPlayer->_pAWidth);
+	file->writeLE<int32_t>(pPlayer->GetAnimationWidth(player_graphic::Attack));
 	file->writeLE<int32_t>(pPlayer->_pAFNum);
 	file->skip(4 * 8); // Skip pointers _pLAnim
 	file->skip(4 * 8); // Skip pointers _pFAnim
 	file->skip(4 * 8); // Skip pointers _pTAnim
 	file->writeLE<int32_t>(pPlayer->_pSFrames);
-	file->writeLE<int32_t>(pPlayer->_pSWidth);
+	file->writeLE<int32_t>(pPlayer->GetAnimationWidth(player_graphic::Magic));
```
AJenbo, line 1448: Do we even need to write these values? I'm thinking the original will set them as it loads the various animations.

Source/player.cpp
```diff
@@ -367,27 +367,75 @@ bool PlayerStruct::IsWalking() const
 	}
 }
 
-void SetPlayerGPtrs(byte *pData, byte **pAnim)
+void PlayerStruct::Reset()
 {
-	int i;
+	// Create empty default initialized PlayerStruct on heap to avoid excessive stack usage
+	auto emptyPlayer = std::make_unique<PlayerStruct>();
+	*this = std::move(*emptyPlayer);
```
glebm, line 374: No need for a variable here?

```suggestion
	*this = std::make_unique<PlayerStruct>();
```

AJenbo:
Nice clean up by the way :)

Source/loadsave.cpp
AJenbo: Checked and thease values are overwritten by LoadGameLevel() right after LoadPlayer() is called so we can just skip them.

Source/player.cpp
obligaron: `error C2679: binary '=': no operator found which takes a right-hand operand of type 'std::unique_ptr<devilution::PlayerStruct,std::default_delete<devilution::PlayerStruct>>' (or there is no acceptable conversion)`

`*this = *std::make_unique<PlayerStruct>();`:
I get `error C2280: 'devilution::PlayerStruct &devilution::PlayerStruct::operator =(const devilution::PlayerStruct &)': attempting to reference a deleted function`

`*this = std::move(*std::make_unique<PlayerStruct>());` compiles
But this looks confusing :hushed:.

Should we define a normal copy constructor?

Source/loadsave.cpp
obligaron: Thanks for testing 🙂 
I removed the entire `GetAnimationWidth` as the only caller was `LoadPlrGFX`.

Source/engine.cpp
AJenbo: That might be a good idea in the rare cases where people have a corrupt diabdat.mpq

obligaron: I added a app_fatal call that would hopefully help to find this cases easier.

AJenbo (approved):

PR #2046, opened 2021-05-24
Direction + Point logic refactoring

julealgon:
This PR introduces a new addition operator to `Point` instances allowing one to sum them with a `Direction`. This eliminates several manual computations of direction + points, simplifying the code considerably.

### DIFF
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -91,6 +91,34 @@ struct Point {
 		return *this;
 	}
 
+	Point &operator+=(Direction direction)
+	{
+		constexpr auto toPoint = [](Direction direction) -> Point {
+			switch (direction) {
+			case DIR_S:
+				return { 1, 1 };
+			case DIR_SW:
+				return { 0, 1 };
+			case DIR_W:
+				return { -1, 1 };
+			case DIR_NW:
+				return { -1, 0 };
+			case DIR_N:
+				return { -1, -1 };
+			case DIR_NE:
+				return { 0, -1 };
+			case DIR_E:
+				return { 1, -1 };
+			case DIR_SE:
+				return { 1, 0 };
+			case DIR_OMNI:
+				return { 0, 0 };
+			}
+		};
+
+		return (*this) += toPoint(direction);
+	}
+
 	Point &operator-=(const Point &other)
 	{
 		x -= other.x;
@@ -108,6 +136,12 @@ struct Point {
 	friend Point operator+(Point a, const Point &b)
 	{
 		a += b;
+		return a;
+	}
+
+	friend Point operator+(Point a, Direction direction)
+	{
+		a += direction;
 		return a;
 	}
 
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -58,7 +58,7 @@ constexpr const T &clamp(const T &x, const T &lower, const T &upper)
 }
 #endif
 
-enum direction : uint8_t {
+enum Direction : uint8_t {
 	DIR_S,
 	DIR_SW,
 	DIR_W,
@@ -463,7 +463,7 @@ void DrawHalfTransparentRectTo(const CelOutputBuffer &out, int sx, int sy, int w
  * @param destination Tile coordinate
  * @return A value from the direction enum
  */
-direction GetDirection(Point start, Point destination);
+Direction GetDirection(Point start, Point destination);
 
 /**
  * @brief Calculate Width2 from the orginal Width
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1808,17 +1808,17 @@ bool TryInvPut()
 	if (numitems >= MAXITEMS)
 		return false;
 
-	direction dir = GetDirection(plr[myplr].position.tile, { cursmx, cursmy });
+	Direction dir = GetDirection(plr[myplr].position.tile, { cursmx, cursmy });
 	if (CanPut(plr[myplr].position.tile.x + offset_x[dir], plr[myplr].position.tile.y + offset_y[dir])) {
 		return true;
 	}
 
-	direction dirLeft = left[dir];
+	Direction dirLeft = left[dir];
 	if (CanPut(plr[myplr].position.tile.x + offset_x[dirLeft], plr[myplr].position.tile.y + offset_y[dirLeft])) {
 		return true;
 	}
 
-	direction dirRight = right[dir];
+	Direction dirRight = right[dir];
 	if (CanPut(plr[myplr].position.tile.x + offset_x[dirRight], plr[myplr].position.tile.y + offset_y[dirRight])) {
 		return true;
 	}
@@ -1844,7 +1844,7 @@ static bool PutItem(int pnum, Point &position)
 
 	auto relativePosition = position - plr[pnum].position.tile;
 
-	direction d = GetDirection(plr[pnum].position.tile, position);
+	Direction d = GetDirection(plr[pnum].position.tile, position);
 
 	if (abs(relativePosition.x) > 1 || abs(relativePosition.y) > 1) {
 		position.x = plr[pnum].position.tile.x + offset_x[d];
@@ -1853,13 +1853,13 @@ static bool PutItem(int pnum, Point &position)
 	if (CanPut(position.x, position.y))
 		return true;
 
-	direction dLeft = left[d];
+	Direction dLeft = left[d];
 	position.x = plr[pnum].position.tile.x + offset_x[dLeft];
 	position.y = plr[pnum].position.tile.y + offset_y[dLeft];
 	if (CanPut(position.x, position.y))
 		return true;
 
-	direction dRight = right[d];
+	Direction dRight = right[d];
 	position.x = plr[pnum].position.tile.x + offset_x[dRight];
 	position.y = plr[pnum].position.tile.y + offset_y[dRight];
 	if (CanPut(position.x, position.y))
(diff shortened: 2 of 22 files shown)


julealgon:
Looks like some tests failed. I'll double check.

Source/engine.h
```diff
@@ -91,6 +91,38 @@ struct Point {
 		return *this;
 	}
 
+	Point &operator+=(const Direction &direction)
```
glebm, line 94: Pass `direction` by value here and below -- it's tiny.

Source/engine.h
```diff
@@ -91,6 +91,38 @@ struct Point {
 		return *this;
 	}
 
+	Point &operator+=(const Direction &direction)
+	{
+		static auto toPoint = [](const Direction &direction) -> Point {
```
glebm, line 96: 1. `constexpr auto`
2. Pass `direction` by value.

Source/engine.h
```diff
@@ -91,6 +91,38 @@ struct Point {
...
+			case DIR_OMNI:
+				return { 0, 0 };
+			}
+		};
+
+		auto point = toPoint(direction);
+
+		x += point.x;
+		y += point.y;
+		return *this;
```
glebm, line 123:
```suggestion
		return (*this) += toPoint(direction);
```

glebm (dismissed):

Source/inv.cpp
```diff
@@ -1808,18 +1808,21 @@ bool TryInvPut()
...
+	Direction dir = GetDirection(plr[myplr].position.tile, { cursmx, cursmy });
+	Point position = plr[myplr].position.tile + dir;
+	if (CanPut(position.x, position.y)) {
 		return true;
 	}
 
-	direction dirLeft = left[dir];
-	if (CanPut(plr[myplr].position.tile.x + offset_x[dirLeft], plr[myplr].position.tile.y + offset_y[dirLeft])) {
+	Direction dirLeft = left[dir];
+	position = plr[myplr].position.tile + dirLeft;
```
AJenbo, line 1818: Do we even need the variable at this point?
```suggestion
	position = plr[myplr].position.tile + left[dir];
```

Source/inv.cpp
```diff
@@ -1808,18 +1808,21 @@ bool TryInvPut()
...
+	if (CanPut(position.x, position.y)) {
 		return true;
 	}
 
-	direction dirRight = right[dir];
-	if (CanPut(plr[myplr].position.tile.x + offset_x[dirRight], plr[myplr].position.tile.y + offset_y[dirRight])) {
+	Direction dirRight = right[dir];
+	position = plr[myplr].position.tile + dirRight;
```
AJenbo, line 1824:
```suggestion
	position = plr[myplr].position.tile + right[dir];
```

Also compared to the original this has come a long way :D
https://github.com/diasurgical/devilution/blob/f1b85a5cd3751363d248f7a2a460d527f69c2a32/Source/inv.cpp#L2378

Source/inv.cpp
```diff
@@ -1844,24 +1847,21 @@ static bool PutItem(int pnum, Point &position)
...
+		position = plr[pnum].position.tile + d;
 	}
 	if (CanPut(position.x, position.y))
 		return true;
 
-	direction dLeft = left[d];
-	position.x = plr[pnum].position.tile.x + offset_x[dLeft];
-	position.y = plr[pnum].position.tile.y + offset_y[dLeft];
+	Direction dLeft = left[d];
+	position = plr[pnum].position.tile + dLeft;
```
AJenbo, line 1859:
```suggestion
	position = plr[pnum].position.tile + left[d];
```

Source/inv.cpp
```diff
@@ -1844,24 +1847,21 @@ static bool PutItem(int pnum, Point &position)
...
 	if (CanPut(position.x, position.y))
 		return true;
 
-	direction dRight = right[d];
-	position.x = plr[pnum].position.tile.x + offset_x[dRight];
-	position.y = plr[pnum].position.tile.y + offset_y[dRight];
+	Direction dRight = right[d];
+	position = plr[pnum].position.tile + dRight;
```
AJenbo, line 1864:
```suggestion
	position = plr[pnum].position.tile + right[d];
```

Source/monster.cpp
```diff
@@ -4689,41 +4691,39 @@ void FreeMonsters()
...
+		if (SolidLoc(futurePosition.x + 1, futurePosition.y) || dFlags[futurePosition.x + 1][futurePosition.y] & BFLAG_MONSTLR)
 			return false;
 	} else if (mdir == DIR_N) {
-		if (SolidLoc(fx + 1, fy) || SolidLoc(fx, fy + 1))
+		if (SolidLoc(futurePosition.x + 1, futurePosition.y) || SolidLoc(futurePosition.x, futurePosition.y + 1))
 			return false;
 	} else if (mdir == DIR_S)
-		if (SolidLoc(fx - 1, fy) || SolidLoc(fx, fy - 1))
+		if (SolidLoc(futurePosition.x - 1, futurePosition.y) || SolidLoc(futurePosition.x, futurePosition.y - 1))
 			return false;
```
AJenbo, line 4714: Would this be doable as `futurePosition + mdir` now?

Source/player.cpp
```diff
@@ -1224,28 +1224,25 @@ bool SolidLoc(int x, int y)
...
 	if ((DWORD)pnum >= MAX_PLRS) {
 		app_fatal("PlrDirOK: illegal player %i", pnum);
 	}
 	auto &player = plr[pnum];
 
-	px = player.position.tile.x + offset_x[dir];
-	py = player.position.tile.y + offset_y[dir];
-
-	if (px < 0 || !dPiece[px][py] || !PosOkPlayer(pnum, px, py)) {
+	Point position = player.position.tile + static_cast<Direction>(dir);
```
AJenbo, line 1234: Wouldn't it be better to redefine `PlrDirOK` as taking `Direction`?

Source/player.cpp
```diff
@@ -1909,7 +1906,8 @@ StartPlayerKill(int pnum, int earflag)
 						i = NUM_INVLOC;
 						while (i--) {
 							pdd = (i + player._pdir) & 7;
-							PlrDeadItem(player, pi, offset_x[pdd], offset_y[pdd]);
+							Point offset = Point { 0, 0 } + static_cast<Direction>(pdd);
+							PlrDeadItem(player, pi, offset.x, offset.y);
```
AJenbo, line 1910: In case this isn't obvious this is the same as calling `right[pdir]` `i` times. I wonder if the is a cleaner way we could write this part.

Source/player.cpp
```diff
@@ -2869,15 +2868,18 @@ bool PM_DoAttack(int pnum)
 		                    || (player.InvBody[INVLOC_HAND_LEFT]._itype == ITYPE_SWORD && player.InvBody[INVLOC_HAND_LEFT]._iLoc == ILOC_TWOHAND)
 		                    || (player.InvBody[INVLOC_HAND_RIGHT]._itype == ITYPE_SWORD && player.InvBody[INVLOC_HAND_RIGHT]._iLoc == ILOC_TWOHAND))
 		                && !(player.InvBody[INVLOC_HAND_LEFT]._itype == ITYPE_SHIELD || player.InvBody[INVLOC_HAND_RIGHT]._itype == ITYPE_SHIELD))))) {
-			dx = player.position.tile.x + offset_x[(player._pdir + 1) % 8];
-			dy = player.position.tile.y + offset_y[(player._pdir + 1) % 8];
+			Point position = player.position.tile + static_cast<Direction>((player._pdir + 1) % 8);
```
AJenbo, line 2871:
```suggestion
			Point position = player.position.tile + right[player._pdir];
```

Source/player.cpp
```diff
@@ -2869,15 +2868,18 @@ bool PM_DoAttack(int pnum)
...
+			dy = position.y;
 			m = ((dMonster[dx][dy] > 0) ? dMonster[dx][dy] : -dMonster[dx][dy]) - 1;
 			if (dMonster[dx][dy] != 0 && !CanTalkToMonst(m) && monster[m].position.old.x == dx && monster[m].position.old.y == dy) {
 				if (PlrHitMonst(-pnum, m))
 					didhit = true;
 			}
-			dx = player.position.tile.x + offset_x[(player._pdir + 7) % 8];
-			dy = player.position.tile.y + offset_y[(player._pdir + 7) % 8];
+
+			position = player.position.tile + static_cast<Direction>((player._pdir + 7) % 8);
```
AJenbo, line 2880:
```suggestion
			position = player.position.tile + left[player._pdir];
```

Source/monster.cpp
julealgon: I'm touching this function in a subsequent PR so this will be applied there.

AJenbo (changes requested):
Would be nice to remove the now redundant variables and use `left`/`right` instead of cryptic logic with cats.

Other then that this look good.

Source/player.cpp
julealgon: For sure. I didn't touch it yet as it was outside the scope of this one. The whole player file is getting massive refactoring in a next PR (already on the way).

Source/player.cpp
julealgon: I'll revisit these in a subsequent PR where we go through every function in the player file.

AJenbo (approved):
On the condition that `PlrDirOK` is redefined in your next PR :)

PR #2048, opened 2021-05-24
Leverage 'Point' in Lighting and Player functions

julealgon:
This PR replaces raw int coordinates passed separately to various functions with `Point` instances.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -3767,37 +3767,6 @@ void MakePlrPath(int pnum, int xx, int yy, bool endspace)
 
 	if (!endspace) {
 		path--;
-
-		switch (player.walkpath[path]) {
-		case WALK_NE:
-			yy++;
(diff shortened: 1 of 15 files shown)


julealgon:
Depends on #2046 being merged first.

Source/player.cpp
```diff
@@ -1224,53 +1224,50 @@ bool SolidLoc(int x, int y)
 
 bool PlrDirOK(int pnum, int dir)
```
AJenbo, line 1225:
```suggestion
bool PlrDirOK(int pnum, Direction dir)
```

julealgon: https://github.com/diasurgical/devilutionX/pull/2050

glebm (approved):

AJenbo (approved):

PR #2051, opened 2021-05-24
Switch ini library from Radon to simpleini

### LINKED ISSUE #1246: ini: Switch from Radon to simpleini

obligaron:
Fix #1246

Notes:
- Works also with unicode paths in windows
- Someone should test this on linux/macos
- Currently bool variables are read per `getIniBool` but stored with `setIniInt`. We should think of unifying this. (not part of this pr)
- Switch to build-in methods for reading bool/long/double. This can result in different output. For example before `deadzone=0.07` (`"%.2f"`) after `deadzone=0.070000` (`"%f"`).  If this is not desired I can revert this change.
- `SimpleIni` supports spaces between key und value `key = value`. Currently the old style is used `key=value`
- `SimpleIni` supports comments. We could consider adding comments from [DevilutionX diablo.ini configuration guide](https://github.com/diasurgical/devilutionX/wiki/DevilutionX-diablo.ini-configuration-guide) directly in the ini at creation. (not part of this pr)
   Possible example:
   ```
   ...
   Sound Volume=0
   
   #0=Max Volume, -1600=No Music; Adjust music volume.
   Music Volume=0
   ...
   ```


qndel:
`Works also with unicode paths in windows` this alone is worth merging it for me :D

AJenbo (approved):

AJenbo:
read/write could probably be even further cleaned up

PR #2054, opened 2021-05-25
Leverage 'Point' in more places

julealgon:
This PR leverages the `Point` struct in a few more functions from player, monster and some other files that are impacted by those changes like function pointers in path.


Source/monster.cpp
```diff
@@ -946,23 +946,23 @@ void PlaceQuestMonsters()
 
 		if (QuestStatus(Q_LTBANNER)) {
 			auto dunData = LoadFileInMem<uint16_t>("Levels\\L1Data\\Banner1.DUN");
-			SetMapMonsters(dunData.get(), 2 * setpc_x, 2 * setpc_y);
+			SetMapMonsters(dunData.get(), Point { setpc_x, setpc_y } * 2);
```
obligaron, line 949: This uses the Point operator*(float) or?
If that is the case this could result in different result (cause of rounding/truncation) or?
Same on other occurrences.

Source/player.cpp
```diff
@@ -1053,7 +1053,7 @@ void InitPlayer(int pnum, bool FirstTime)
 				player.position.tile = { ViewX, ViewY };
 			}
 		} else {
-			for (i = 0; i < 8 && !PosOkPlayer(pnum, plrxoff2[i] + player.position.tile.x, plryoff2[i] + player.position.tile.y); i++)
+			for (i = 0; i < 8 && !PosOkPlayer(pnum, Point { plrxoff2[i], plryoff2[i] } + player.position.tile); i++)
```
obligaron, line 1056: Could the plr(x/y)off2 array also be converted to `Point` (in this or future pr)?

Source/monster.cpp
julealgon: Rounding/truncation with a flat 2? Would that really be possible?

Are you suggesting I should add an overload for the operator with ints?

Source/player.cpp
julealgon: I'm converting a lot of stuff. Eventually everything that _can_ use a `Point`, will. I'm just not doing everything on the same change to avoid huge PRs that stay open for great lengths of time and cause conflicts etc.

Doing incrementally avoids that.

Source/monster.cpp
obligaron: I would add a new overload.
There are also devices without FPU.

Source/monster.cpp
```diff
@@ -4698,19 +4698,19 @@ bool DirOK(int i, Direction mdir)
 
 	commitment((DWORD)i < MAXMONSTERS, i);
 	Point futurePosition = monster[i].position.tile + mdir;
```
AJenbo, line 4700:
```suggestion
	Point position = monster[i].position.tile;
	Point futurePosition = position + mdir;
```

Source/monster.cpp
```diff
@@ -4698,19 +4698,19 @@ bool DirOK(int i, Direction mdir)
 
 	commitment((DWORD)i < MAXMONSTERS, i);
 	Point futurePosition = monster[i].position.tile + mdir;
-	if (futurePosition.y < 0 || futurePosition.y >= MAXDUNY || futurePosition.x < 0 || futurePosition.x >= MAXDUNX || !PosOkMonst(i, futurePosition.x, futurePosition.y))
+	if (futurePosition.y < 0 || futurePosition.y >= MAXDUNY || futurePosition.x < 0 || futurePosition.x >= MAXDUNX || !PosOkMonst(i, futurePosition))
 		return false;
 	if (mdir == DIR_E) {
-		if (SolidLoc(futurePosition.x, futurePosition.y + 1) || dFlags[futurePosition.x][futurePosition.y + 1] & BFLAG_MONSTLR)
+		if (SolidLoc(futurePosition + Point { 0, 1 }) || dFlags[futurePosition.x][futurePosition.y + 1] & BFLAG_MONSTLR)
```
AJenbo, line 4704: Thease are checking if the player is diagonally blocked by a monster so it would actually make sens to use the DIR_* for checking the relative position.

If `monster` tries to move `DIR_E`, check for monster at `DIR_E+DIR_SW`. But even better would be to not add mdir and the using an illogically direction that subtraction it.

```suggestion
		if (SolidLoc(position + DIR_SE) || dFlags[position.x + 1][futurePosition.y] & BFLAG_MONSTLR)
```

Source/monster.cpp
```diff
@@ -4698,19 +4698,19 @@ bool DirOK(int i, Direction mdir)
...
-	if (futurePosition.y < 0 || futurePosition.y >= MAXDUNY || futurePosition.x < 0 || futurePosition.x >= MAXDUNX || !PosOkMonst(i, futurePosition.x, futurePosition.y))
+	if (futurePosition.y < 0 || futurePosition.y >= MAXDUNY || futurePosition.x < 0 || futurePosition.x >= MAXDUNX || !PosOkMonst(i, futurePosition))
 		return false;
 	if (mdir == DIR_E) {
-		if (SolidLoc(futurePosition.x, futurePosition.y + 1) || dFlags[futurePosition.x][futurePosition.y + 1] & BFLAG_MONSTLR)
+		if (SolidLoc(futurePosition + Point { 0, 1 }) || dFlags[futurePosition.x][futurePosition.y + 1] & BFLAG_MONSTLR)
 			return false;
 	} else if (mdir == DIR_W) {
-		if (SolidLoc(futurePosition.x + 1, futurePosition.y) || dFlags[futurePosition.x + 1][futurePosition.y] & BFLAG_MONSTLR)
+		if (SolidLoc(futurePosition + Point { 1, 0 }) || dFlags[futurePosition.x + 1][futurePosition.y] & BFLAG_MONSTLR)
```
AJenbo, line 4707:
```suggestion
		if (SolidLoc(position + DIR_SW) || dFlags[position.x][position.y + 1] & BFLAG_MONSTLR)
```

Source/player.cpp
obligaron: That's good and I like that approach. 👍 
I'm aware of this and I saw also other places that aren't converted (for example `ActivateSpawn`) and didn't comment.
I comment on this cause,
- both users of the array are modified by this pr (changed to `Point`).
- this kind of arrays can be forgotten.
- this is only a question/reminder. 😉

Source/monster.cpp
```diff
@@ -4698,19 +4698,19 @@ bool DirOK(int i, Direction mdir)
...
 			return false;
 	} else if (mdir == DIR_N) {
-		if (SolidLoc(futurePosition.x + 1, futurePosition.y) || SolidLoc(futurePosition.x, futurePosition.y + 1))
+		if (SolidLoc(futurePosition + Point { 1, 0 }) || SolidLoc(futurePosition + Point { 0, 1 }))
```
AJenbo, line 4710:
```suggestion
		if (SolidLoc(position + DIR_NE) || SolidLoc(position + DIR_NW))
```

Source/monster.cpp
```diff
@@ -4698,19 +4698,19 @@ bool DirOK(int i, Direction mdir)
...
 			return false;
 	} else if (mdir == DIR_S)
-		if (SolidLoc(futurePosition.x - 1, futurePosition.y) || SolidLoc(futurePosition.x, futurePosition.y - 1))
+		if (SolidLoc(futurePosition + Point { -1, 0 }) || SolidLoc(futurePosition + Point { 0, -1 }))
```
AJenbo, line 4713:
```suggestion
		if (SolidLoc(position + DIR_SW) || SolidLoc(position + DIR_SE))
```

AJenbo (changes requested):

Source/player.cpp
julealgon: Yeah I'm aware that this leaves some places "dangling" for a while but it's better than tackling everything in one go IMHO.

Source/player.cpp
```diff
@@ -1120,13 +1120,13 @@ void InitMultiView()
 	ViewY = myPlayer.position.tile.y;
 }
 
-bool SolidLoc(int x, int y)
+bool SolidLoc(Point position)
 {
-	if (x < 0 || y < 0 || x >= MAXDUNX || y >= MAXDUNY) {
+	if (position.x < 0 || position.y < 0 || position.x >= MAXDUNX || position.y >= MAXDUNY) {
```
AJenbo, line 1125: Making a function for checking for level OOB would be a big help to a lot of code as well and probably catch a handful of bugs (we already fixed a lot of issues like that).

julealgon: For sure. Ideally, we want to limit operating on the individual coordinates of these points as much as possible, and these bound checks could easily be done using rectangle-like operations. I'll look into this at some point in the near future when more of the code is properly using points. We also need some Matrix2d abstractions that can be indexed by points.

Source/player.cpp
```diff
@@ -1139,17 +1139,17 @@ bool PlrDirOK(int pnum, Direction dir)
 	auto &player = plr[pnum];
 
 	Point position = player.position.tile + dir;
```
AJenbo, line 1141:
```suggestion
	Point position = player.position.tile;
	Point futurePosition = position + dir;
```

Source/player.cpp
```diff
@@ -1139,17 +1139,17 @@ bool PlrDirOK(int pnum, Direction dir)
 	auto &player = plr[pnum];
 
 	Point position = player.position.tile + dir;
-	if (position.x < 0 || !dPiece[position.x][position.y] || !PosOkPlayer(pnum, position.x, position.y)) {
+	if (position.x < 0 || !dPiece[position.x][position.y] || !PosOkPlayer(pnum, position)) {
```
AJenbo, line 1142:
```suggestion
	if (futurePosition.x < 0 || !dPiece[futurePosition.x][futurePosition.y] || !PosOkPlayer(pnum, futurePosition)) {
```

Source/player.cpp
```diff
@@ -1139,17 +1139,17 @@ bool PlrDirOK(int pnum, Direction dir)
...
 	Point position = player.position.tile + dir;
-	if (position.x < 0 || !dPiece[position.x][position.y] || !PosOkPlayer(pnum, position.x, position.y)) {
+	if (position.x < 0 || !dPiece[position.x][position.y] || !PosOkPlayer(pnum, position)) {
 		return false;
 	}
 
 	isOk = true;
 	if (dir == DIR_E) {
-		isOk = !SolidLoc(position.x, position.y + 1) && !(dFlags[position.x][position.y + 1] & BFLAG_PLAYERLR);
+		isOk = !SolidLoc(position + Point { 0, 1 }) && !(dFlags[position.x][position.y + 1] & BFLAG_PLAYERLR);
```
AJenbo, line 1148:
```suggestion
		isOk = !SolidLoc(position + DIR_SE) && !(dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR);
```

Source/monster.cpp
julealgon: I was sure these additions were direction-based but I didn't spend any time looking into the logic itself: this refactor was more of a "dumb" replacement and then I'd go back to these and refactor the logic.

Thanks for explaining it, I'll apply now in this case.

Source/player.cpp
```diff
@@ -1139,17 +1139,17 @@ bool PlrDirOK(int pnum, Direction dir)
...
 	}
 
 	if (isOk && dir == DIR_W) {
-		isOk = !SolidLoc(position.x + 1, position.y) && !(dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR);
+		isOk = !SolidLoc(position + Point { 1, 0 }) && !(dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR);
```
AJenbo, line 1152:
```suggestion
		isOk = !SolidLoc(position + DIR_SW) && !(dFlags[position.x][position.y + 1] & BFLAG_PLAYERLR);
```

AJenbo (changes requested):

Source/spells.cpp
```diff
@@ -212,45 +212,44 @@ void CastSpell(int id, int spl, int sx, int sy, int dx, int dy, int spllvl)
...
 	bool done;
+	Point newPosition = {};
 
 	if (plr[pnum].plrlevel == currlevel) {
 		for (i = 0; i < 8; i++) {
-			nx = plr[pnum].position.tile.x + plrxoff2[i];
-			ny = plr[pnum].position.tile.y + plryoff2[i];
-
-			if (PosOkPlayer(pnum, nx, ny)) {
+			newPosition = plr[pnum].position.tile + Point { plrxoff2[i], plryoff2[i] };
```
AJenbo, line 222: Was merging `plrxoff2` + `plryoff2` in to a single `Point` array already mentioned?

julealgon: Obligaron mentioned it. We'll get there.

Source/player.cpp
AJenbo: Not sure why you marked this as resolved but didn't apply it, but applied the exact same change to the monster version of this function.

AJenbo (commented):
Why correct DirOK but not PlrDirOK?

Source/player.cpp
julealgon: That was me having a brain fart. I thought the comments were duplicated. Fixed now.

julealgon:
> Why correct DirOK but not PlrDirOK?

Just me being dumb. Thought the comments were duplicated (due to how similar they were). Didn't realize the logic was similar in the 2 functions (again, this was a dumb replacement, so I didn't spent much time looking at the logic in each place).

Please take another look.

AJenbo (approved):

AJenbo (approved):

PR #2055, opened 2021-05-25
Player by reference part 3

AJenbo:
Removed unused function `GetPlrGFXClass`.
Fixed impossible cases of `playerId == -1` in bloodboil code.
Moved `CalcPlrStaff` next to `CalcPlrSpell` in `player.cpp`
Plus the usual stuff that I found while plowing over the code.

Ignoring white-space will reduce the diff by about 150 lines.

Just 13 more files to go over.


julealgon (changes requested):
Stopped in the middle. I wish you could break your changes a bit more and not do everything at the same time... this would be a bit slower but would lead to less bugs in the future.

Some of the logic change diffs are really, _really_ hard to follow with other stuff going on at the same time.

Anyways... left one comment on an issue I found.

AJenbo:
> I wish you could break your changes a bit more and not do everything at the same time... this would be a bit slower but would lead to less bugs in the future.

Understandable, but not easy because of the order that this is happening in. I would have to commit, refactor and rebase before I could resume which would be a lot slower.

> Some of the logic change diffs are really, _really_ hard to follow with other stuff going on at the same time.

Ignoring white space should help a lot, changes to indentation unfortunately makes the diff on GitHub miss align quite a bit when white space is not ignored.

Source/control.cpp
```diff
@@ -313,10 +313,9 @@ static void DrawSpell(const CelOutputBuffer &out)
 	if (currlevel == 0 && st != RSPLTYPE_INVALID && !spelldata[spl].sTownSpell)
 		st = RSPLTYPE_INVALID;
 	SetSpellTrans(st);
-	if (spl != SPL_INVALID)
-		DrawSpellCel(out, PANEL_X + 565, PANEL_Y + 119, *pSpellCels, SpellITbl[spl]);
-	else
-		DrawSpellCel(out, PANEL_X + 565, PANEL_Y + 119, *pSpellCels, 27);
+
+	int nCel = (spl != SPL_INVALID) ? SpellITbl[spl] : 27;
```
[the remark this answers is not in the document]
AJenbo: I actually like this better. The expected condition comes first and you can more or less ignore the rest of the line.

Source/missiles.cpp
```diff
@@ -2904,23 +2900,24 @@ void AddNova(int mi, int sx, int sy, int dx, int dy, int midir, int8_t mienemy,
 
 void AddBlodboil(int mi, int sx, int sy, int dx, int dy, int midir, int8_t mienemy, int id, int dam)
 {
-	if (id == -1 || plr[id]._pSpellFlags & 6 || plr[id]._pHitPoints <= plr[id]._pLevel << 6) {
+	auto &player = plr[id];
+
+	if (id == -1 || player._pSpellFlags & 6 || player._pHitPoints <= player._pLevel << 6) {
```
StephenCWills, line 2905: We should probably check for `id == -1` before attempting to access `plr[id]`.

Source/missiles.cpp
```diff
@@ -3036,18 +3033,8 @@ void AddCbolt(int mi, int sx, int sy, int dx, int dy, int midir, int8_t micaster
...
-		} else {
-			missile[mi]._mirnd = GenerateRnd(15) + 1;
-			missile[mi]._midam = GenerateRnd(plr[id]._pMagic / 4) + 1;
-		}
-	} else {
-		missile[mi]._mirnd = GenerateRnd(15) + 1;
-		missile[mi]._midam = 15;
-	}
+	missile[mi]._mirnd = GenerateRnd(15) + 1;
+	missile[mi]._midam = (micaster == 0) ? (GenerateRnd(plr[id]._pMagic / 4) + 1) : 15;
```
StephenCWills, line 3037: What the heck was this nonsense? Nice cleanup. 👍

Source/missiles.cpp
```diff
@@ -3895,109 +3880,112 @@ void MI_Reflect(int i)
...
+		return;
+
+	missile[i]._miDelFlag = true;
+	PlaySfxLoc(IS_CAST7, plr[missile[i]._miVar1].position.tile.x, plr[missile[i]._miVar1].position.tile.y);
+	AutoMapShowItems = false;
 }
 
 void MI_LightningWallC(int i)
 {
+	int dp, tx, ty;
```
StephenCWills, line 3955: Normally you join declaration and assignment. Why did you separate them?

Source/missiles.cpp
```diff
@@ -4678,33 +4667,37 @@ void MI_Fireman(int i)
 
 void MI_FirewallC(int i)
 {
+	int dp, tx, ty;
```
StephenCWills, line 4670: Same as Lightning Wall. These declarations were separated.

StephenCWills:
Made it to the end. Changing indentation does make it more difficult to parse. Ignoring whitespace helps in some ways and hurts in others. However, I think this difficulty is fairly unavoidable. You could try to break cleanup into several smaller steps to ease the burden on reviewers, but it would require multiple passes through the codebase and/or tons of commits to sift through. It really is more efficient to just rewrite the whole function to fix multiple problems at once so you won't have to revisit. The potential for errors introduced by the rewrite is part of what makes the change worth reviewing.

Source/missiles.cpp
```diff
@@ -592,135 +587,145 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, int t, bool
...
+
+	if (hit >= hper)
+		return false;
+
 #ifdef _DEBUG
-	if (hit < hper || debug_mode_key_inverted_v || debug_mode_dollar_sign) {
-#else
-	if (hit < hper) {
+	if (!debug_mode_key_inverted_v && !debug_mode_dollar_sign)
+		return false;
```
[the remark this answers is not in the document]
AJenbo: Arh, right. This one has always been a bit of a pain. Oh well at least now it shouldn't confuse the editors regarding indentation despite having to retain an `#else`

Source/missiles.cpp
AJenbo: To make the left/right growth code as identical as possible and to indicate that it needs some proper refactoring.

Source/items.cpp
```diff
@@ -3917,41 +3907,43 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
...
 	int l, j;
 
+	auto &player = plr[p];
+
 	switch (Mid) {
 	case IMISC_HEAL:
 	case IMISC_FOOD:
-		j = plr[p]._pMaxHP >> 8;
+		j = player._pMaxHP >> 8;
 		l = ((j >> 1) + GenerateRnd(j)) << 6;
```
thebigMuh, line 3916:
```suggestion
		l = GenerateRnd(player._pMaxHP / 4) + player._pMaxHP / 8;
```

Source/items.cpp
```diff
@@ -3917,41 +3907,43 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
...
-		plr[p]._pHitPoints = plr[p]._pMaxHP;
-		plr[p]._pHPBase = plr[p]._pMaxHPBase;
+		player._pHitPoints = player._pMaxHP;
+		player._pHPBase = player._pMaxHPBase;
 		drawhpflag = true;
 		break;
 	case IMISC_MANA:
-		j = plr[p]._pMaxMana >> 8;
+		j = player._pMaxMana >> 8;
 		l = ((j >> 1) + GenerateRnd(j)) << 6;
```
thebigMuh, line 3932:
```suggestion
		l = GenerateRnd(player._pMaxMana / 4) + player._pMaxMana / 8;
```

Source/missiles.cpp
AJenbo: Have it done, will send it as the next one. Will be a nice small on in comparison :D

Source/items.cpp
```diff
@@ -3972,89 +3964,89 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
...
-			plr[p]._pHPBase = plr[p]._pMaxHPBase;
+			player._pHitPoints = player._pMaxHP;
+			player._pHPBase = player._pMaxHPBase;
 			drawhpflag = true;
 		}
 		break;
 	case IMISC_REJUV:
-		j = plr[p]._pMaxHP >> 8;
+		j = player._pMaxHP >> 8;
 		l = ((j / 2) + GenerateRnd(j)) << 6;
```
thebigMuh, line 3974: Ah someone changed the >> 1 to a / 2 here.
Still the same as above:
```suggestion
		l = GenerateRnd(player._pMaxHP / 4) + player._pMaxHP / 8;
```

Source/items.cpp
```diff
@@ -3972,89 +3964,89 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
...
+		if (player._pClass == HeroClass::Rogue)
 			l += l / 2;
-		plr[p]._pHitPoints = std::min(plr[p]._pHitPoints + l, plr[p]._pMaxHP);
-		plr[p]._pHPBase = std::min(plr[p]._pHPBase + l, plr[p]._pMaxHPBase);
+		player._pHitPoints = std::min(player._pHitPoints + l, player._pMaxHP);
+		player._pHPBase = std::min(player._pHPBase + l, player._pMaxHPBase);
 		drawhpflag = true;
-		j = plr[p]._pMaxMana >> 8;
+		j = player._pMaxMana >> 8;
 		l = ((j / 2) + GenerateRnd(j)) << 6;
```
thebigMuh, line 3983:
```suggestion
		l = GenerateRnd(player._pMaxMana / 4) + player._pMaxMana / 8;
```

Source/items.cpp
```diff
@@ -4272,9 +4266,11 @@ static void SpawnOnePremium(int i, int plvl, int myplr)
...
-	int dexterity = std::max(plr[myplr].GetMaximumAttributeValue(CharacterAttribute::Dexterity), plr[myplr]._pDexterity);
-	int magic = std::max(plr[myplr].GetMaximumAttributeValue(CharacterAttribute::Magic), plr[myplr]._pMagic);
+	auto &myPlayer = plr[myplr];
+
+	int strength = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Strength), myPlayer._pStrength);
+	int dexterity = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Dexterity), myPlayer._pDexterity);
+	int magic = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Magic), myPlayer._pMagic);
 	strength *= 1.2;
 	dexterity *= 1.2;
 	magic *= 1.2;
```
thebigMuh, line 4276: The random doubles that appear every now and then bother me.
```suggestion
	strength += strength / 5;
	dexterity += dexterity / 5;
	magic += magic / 5;
```

Source/items.cpp
```diff
@@ -4528,16 +4524,16 @@ int RndBoyItem(int lvl)
...
-	int magic = std::max(plr[myplr].GetMaximumAttributeValue(CharacterAttribute::Magic), plr[myplr]._pMagic);
+	auto &myPlayer = plr[myplr];
+
+	HeroClass pc = myPlayer._pClass;
+	int strength = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Strength), myPlayer._pStrength);
+	int dexterity = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Dexterity), myPlayer._pDexterity);
+	int magic = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Magic), myPlayer._pMagic);
 	strength *= 1.2;
 	dexterity *= 1.2;
 	magic *= 1.2;
```
thebigMuh, line 4539:
```suggestion
	strength += strength / 5;
	dexterity += dexterity / 5;
	magic += magic / 5;
```

Source/missiles.cpp
AJenbo: Looks like this one also cannot happen as blood boil is only cast by the player (`PM_DoSpell()`) and so always have an id.

thebigMuh (dismissed):
I wonder how many bugs slip through the cracks with these huge sweeping changes...

StephenCWills (approved):

Iniquitatis:
Off-topic:
@thebigMuh As much as I'm also concerned about this, I'm honestly failing to see Diablo as some sort of super-complex game with deep mechanics. The current codebase (which is derived from the original one) is unnecessarily convoluted, so cleaner code = easier to fix bugs and replicate desired behavior.

julealgon:
@StephenCWills 
> It really is more efficient to just rewrite the whole function to fix multiple problems at once so you won't have to revisit. 

It might be more efficient for the person making the change, but it is much less efficient for people reviewing, reading the code later, and debugging why something is not working on that function when looking at the history.

> The potential for errors introduced by the rewrite is part of what makes the change worth reviewing.

I think you have it the wrong way there: incremental refactoring is _vastly_ safer than making every single adjustment at the same time. You increase the risk of inadvertently changing things, you make it harder for others to spot the same mistake, etc.

---

@thebigMuh 
> I wonder how many bugs slip through the cracks with these huge sweeping changes...

Some will invariably do, it has happened in the recent past several times. The more stuff we touch simultaneously, the more exponential the chance for bugs will become.

---

@Iniquitatis 
> Off-topic:
@thebigMuh As much as I'm also concerned about this, I'm honestly failing to see Diablo as some sort of super-complex game with deep mechanics. The current codebase (which is derived from the original one) is unnecessarily convoluted, so cleaner code = easier to fix bugs and replicate desired behavior.

I don't think anyone is advocating for "not making it cleaner": that really is not the point of the discussion. The point (at least from my perspective) is that, by conflating multiple improvement categories in the same PRs and commits, you:
- increase the risk of introducing bugs, since there are move moving parts at any given time, and it is harder to review so consequently easier to miss problems. When you combine moving parts, they become exponentially more complex, not linearly.
- make it harder to read the change history on the project later. You see a commit stating "change to references for PlayerStruct" however there are other changes in that commit that are unrelated to that. You may end up skipping the commit by it's title, and miss something important. This is especially relevant when trying to fix bugs.

I know I insist a lot on this front, but I do it because I honestly think the codebase would be healthier that way. Yes, it needs more diligence, patience and time to do it properly, but the positives outweigh the negatives IMHO.


--- 

@AJenbo 
> Edit: I could split it up by function if that would help.

It would be an improvement, but not my main concern. Even if it is "per-function", if there are lots of different changes in it, it is still less than ideal. I'd prefer if you could focus on "one problem at a time". So, "convert to use early return" is a problem, "use references" is another, "move variable declarations closer to usage" another, "simplify the logic of X, Y Z" is yet another.

StephenCWills:
> It might be more efficient for the person making the change, but it is much less efficient for people reviewing, reading the code later, and debugging why something is not working on that function when looking at the history.

It's more efficient overall. Your suggestion of "convert to use early return," for instance, is not the type of thing that's easy to just do a search for and find all instances of it. Your comment about needing more diligence, patience, and time is perhaps a little bit of an understatement. For every little issue you identify as you review the codebase, you'd have to scour the entire codebase for each instance of it to fix them all so you can pack it up in a neat commit for your reviewers, and then your reviewers would have to review every little change independently on the same functions over and over. And since you're touching large parts of the codebase over and over, you'll frequently conflict with other PRs and changes as you're not the only one working on the codebase at any given time.

Even just as a reviewer, I would actually much prefer a full rewrite of individual functions over incremental changes to every function across the entire codebase. However, I recognize that some things are okay to do all at once and others are good to do as you go. I honestly don't think there's anything particularly wrong with this PR except maybe that it mixed a large sweeping change with a handful of rewrites that could have perhaps been organized into separate commits. However, it was pretty obvious which parts of the code were in need of more attention than others.

> I think you have it the wrong way there: incremental refactoring is _vastly_ safer than making every single adjustment at the same time. You increase the risk of inadvertently changing things, you make it harder for others to spot the same mistake, etc.

I'm not sure what you mean. Obviously I can't argue that incremental refactoring would be safer, and in that respect I do agree with what you're saying. However, I'm not clear on why that means I have it the wrong way.

AJenbo:
> > Edit: I could split it up by function if that would help.
> 
> It would be an improvement, but not my main concern. Even if it is "per-function", if there are lots of different changes in it, it is still less than ideal. I'd prefer if you could focus on "one problem at a time". So, "convert to use early return" is a problem, "use references" is another, "move variable declarations closer to usage" another, "simplify the logic of X, Y Z" is yet another.

How about I do a rebase with a pure player by reference followed by everything that i noticed when doing the other part.
The big thing for me is that I have to do all does changes once I apply clang-tidy, and that makes the clang-tidy commits pretty unruly, so this is an attempt at spreading it out a bit.

julealgon:
@AJenbo 
> How about I do a rebase with a pure player by reference followed by everything that i noticed when doing the other part.

If you can do that, it would be great. But again... I don't want you to waste even more time on conceptually trivial changes like this. The comments were more in regards to future changes. I've already mentioned this a few times by now but you still opted to combine the changes consciously, so I'll not insist anymore: instead, I'd just opt to not review some of the changes going forward.

Just keep in mind that, if you were to send these changes in a more focused manner, they'd all have probably been merged by now. By conflating changes, you are actually making your own life worse here, as evidenced by new conflicts that cropped.

> The big thing for me is that I have to do all does changes once I apply clang-tidy, and that makes the clang-tidy commits pretty unruly, so this is an attempt at spreading it out a bit.

One suggestion I can make is to avoid adding multiple clang-tidy rules at the same time and then trying to apply all of them in one go. I've cleaned up projects in the past with 5 thousand warnings by applying fixes on a per category basis and it has worked very well. Sounds like something similar could be done here: for every clang tidy new rule, make sure the entire project follows that rule, and then activate it. Do the same steps for every rule and every change will become drastically simpler and more specific.

Not sure if the boat has already sailed for such approach, though. In which case, you could manually disable the rules, enable one, refactor, enable another, refactor, and do this until all rules are enabled again. I know... considerably more work, but only because this wasn't done from the beginning.

AJenbo:
> Just keep in mind that, if you were to send these changes in a more focused manner, they'd all have probably been merged by now.

Well that is actually totally op to me.

AJenbo:
> > The big thing for me is that I have to do all does changes once I apply clang-tidy, and that makes the clang-tidy commits pretty unruly, so this is an attempt at spreading it out a bit.
> 
> One suggestion I can make is to avoid adding multiple clang-tidy rules at the same time and then trying to apply all of them in one go. I've cleaned up projects in the past with 5 thousand warnings by applying fixes on a per category basis and it has worked very well. Sounds like something similar could be done here: for every clang tidy new rule, make sure the entire project follows that rule, and then activate it. Do the same steps for every rule and every change will become drastically simpler and more specific.
> 
> Not sure if the boat has already sailed for such approach, though. In which case, you could manually disable the rules, enable one, refactor, enable another, refactor, and do this until all rules are enabled again. I know... considerably more work, but only because this wasn't done from the beginning.

I have looked at that option as well, there just isn't a way that makes this simple easy and quick.

julealgon:
@StephenCWills 
> It's more efficient overall.

I feel like the fact that this hasn't been merged yet is a very clearly indication that it is not. Smaller and more focused changes are reviewed much faster, and have dramatically reduced chance of conflicts because of that.

> Your suggestion of "convert to use early return," for instance, is not the type of thing that's easy to just do a search for and find all instances of it. Your comment about needing more diligence, patience, and time is perhaps a little bit of an understatement. For every little issue you identify as you review the codebase, you'd have to scour the entire codebase for each instance of it to fix them all so you can pack it up in a neat commit for your reviewers, and then your reviewers would have to review every little change independently on the same functions over and over. 

Ideally, fixing a single category of problems once would be best, indeed, but as you said, it's not always possible. With early returns, I totally agree that going through the entire code base would be too time consuming and potentially generate a gigantic change, so you can do that piecemeal. Solving "per category" can still be done in parts of the code as long as you make that clear in the commit message. For example, "Employ early-return on player functions" is a perfectly valid commit that still solves a category of issues, but scoped to a particular section of the code that is still manageable.

The diligence/patience part was more in regards to being aware of this split beforehand when doing the actual refactoring work. Sometimes, you end up finding extra things to fix that impact your current change and you can opt to stop the current refactor (stash for example), do the other cleanup, and then continue with your previous work. This requires a bit more effort on the person doing the change, obviously.

Sorry I didn't make that clear before.

> And since you're touching large parts of the codebase over and over, you'll frequently conflict with other PRs and changes as you're not the only one working on the codebase at any given time.

The chance for conflicts depend not only on how much code is touched, but on how long the change stays stale. An ultra simple change that spans the entire code base is _still_ easy to review and fast to merge, so the chance of conflicts is reduced. Also, since the change is simple, only very few lines would be touched on every file/method/etc, so the chance of conflits due to the "amount of code" is _also_ reduced there.

Again, I can only use examples that have worked well for me, so here is another one: I just made a change in one very big project of ours to change the namespaces of every class in it. This ended up touching upwards of 2 thousand .cs files. However, the change was still trivial to review. What I did was to separate this change in a set of smaller namespace renames, starting with inner folders first, and going to outside ones later. Each change ended up touching at most 180 files, which was still manageable to review as namespace renames were basically a couple of lines on each .cs file. Since the changes were reviewed fast and thus merged fast, there were no conflicts with work being done by other people.

> Even just as a reviewer, I would actually much prefer a full rewrite of individual functions over incremental changes to every function across the entire codebase.

Ok, fair enough. Not everything is 100% objective. We'll just agree to disagree on this one. I find it much easier to review a "simple change" that is "applied many times" (or many simple changes in sequence) vs a "complex one" applied fewer times.

AJenbo:
> > It's more efficient overall.
> 
> I feel like the fact that this hasn't been merged yet is a very clear indication that it is not. 

Part 2 was even larger and recived less reviews, so I don't think you can generalize like that.

StephenCWills:
> > It's more efficient overall.
> 
> I feel like the fact that this hasn't been merged yet is a very clear indication that it is not. Smaller and more focused changes are reviewed much faster, and have dramatically reduced chance of conflicts because of that.

I thought the reason this wasn't merged is because it had nearly immediate conflicts with one of obligaron's PRs so AJenbo offered to defer the merge. I don't think it was being held up by reviewers.

> Ok, fair enough. Not everything is 100% objective. We'll just agree to disagree on this one. I find it much easier to review a "simple change" that is "applied many times" (or many simple changes in sequence) vs a "complex one" applied fewer times.

I get what you're saying from the perspective of a style change or a rename. If the intent is a simple cleanup without changing behavior, then it's easier for the reviewer to just double-check all instances of the change for equivalence. And certainly, taken individually, any simple change will always be easier to review than a complex one.

However, as a general rule, it's better to focus on the correctness of the functionality encapsulated by the abstraction than the equivalence of the change itself, whether making the change or reviewing. As an example of this, take AJenbo's changes to `AddBloodboil()`, in which he had to check usage to determine whether `-1` was ever used as input to the function. This had a direct effect on the structure of the code he was attempting to modify because the variable declaration for the player struct couldn't be placed in the middle of the conditional. If you ever have to conceptualize the encapsulated functionality to determine the correctness of a change, it hardly matters how simple or complex the change is because the total effort spent is not limited to the change being made. Breaking a rewrite into many smaller changes is only going to increase the workload on everyone.

Now, that's not to say you should always scrutinize every change to make sure it was correct in context of the function it was made in. The vast majority of my review was just me double-checking for equivalence, and those certainly did represent the easiest (and most tedious) parts. If I were to offer any advice on how this PR should have been broken up, I would say that any function that required more than just a variable declaration and a bunch of replacements should have had its own commit. That would split the changes up so I could have raced through the equivalence checks, followed up by methodically moving through the correctness checks one-by-one. As it stands, I was continually interrupted from the process of reviewing the simple checks by having to review the more complex ones littered throughout the codebase.

---

As for the other parts of your reply, I do not disagree. I appreciate your willingness to clarify and share examples from your experience.

Issue #2062, opened 2021-05-25
Pause is displayed smaller and not centered


Chance4us:
Pause is displayed smaller and not centered since commit 6acb49bea4cc3a1ea77bad73bfa7a831e7ae3fbf.

![pause](https://user-images.githubusercontent.com/45902449/119563455-db37fe00-bda7-11eb-8c2b-16de73c0d87f.png)

julealgon:
Wait, isn't the screen supposed to be red when paused as well?

Chance4us:
> Wait, isn't the screen supposed to be red when paused as well?

It is supposed to be red only in dungeon. In town it is not supposed to be red. Only the text is no longer entirely correct.

qndel:
![image](https://user-images.githubusercontent.com/14297035/119569909-b21b6b80-bdaf-11eb-9a6b-cd7bf7c40577.png)
devilutionx 640x480

julealgon:
> It is supposed to be red only in dungeon. In town it is not supposed to be red.

Interesting. I never noticed that detail. Thanks for clarifying @Chance4us .

Chance4us:
@julealgon Don't mention it. You are welcome!

PR #2063, opened 2021-05-25
fix pause font and centering

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2062

### DIFF
--- a/Source/gmenu.cpp
+++ b/Source/gmenu.cpp
@@ -39,7 +39,7 @@ void gmenu_draw_pause(const CelOutputBuffer &out)
 		RedBack(out);
 	if (sgpCurrentMenu == nullptr) {
 		light_table_index = 0;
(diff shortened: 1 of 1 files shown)


Source/gmenu.cpp
```diff
@@ -39,7 +39,7 @@ void gmenu_draw_pause(const CelOutputBuffer &out)
 		RedBack(out);
 	if (sgpCurrentMenu == nullptr) {
 		light_table_index = 0;
-		DrawString(out, _("Pause"), { PANEL_LEFT + 252, 176, 0, 0 }, UIS_MED, 2);
+		DrawString(out, _("Pause"), { PANEL_LEFT + 252, PANEL_TOP / 2, 0, 0 }, UIS_HUGE, 2);
```
julealgon, line 42: I wonder when this regressed... it must've been a long time ago right?

qndel: looks like it has always been like this, only centered horizontally but not vertically, I think when people saw it near top of the screen on different resolutions, they assumed it was a feature

PR #2064, opened 2021-05-25
margins and tweaks

qndel:
old vs new
oldnewhighlight
+pixel perfect label targetting
+fixes labels appearing red for no reason

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1156,10 +1156,7 @@ void DrawInfoBox(const CelOutputBuffer &out)
 				strcpy(infostr, myPlayer.HoldItem._iIName);
 			else
 				strcpy(infostr, myPlayer.HoldItem._iName);
-			if (myPlayer.HoldItem._iMagical == ITEM_QUALITY_MAGIC)
-				infoclr = UIS_BLUE;
-			if (myPlayer.HoldItem._iMagical == ITEM_QUALITY_UNIQUE)
-				infoclr = UIS_GOLD;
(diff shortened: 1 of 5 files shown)


Chance4us:
By the way, the item label color in dungeon is always red, but it doesn't seem to be a highlight related issue.
It seems to be more related to the getTextColor function.

qndel:
rofl, looks like item stat flag is false till you pick the item up, that's why items in town are mostly while, because there are no monsters to drop them there, while any loot dropped by monster will be red in dungeon>.>

Chance4us:
> rofl, looks like item stat flag is false till you pick the item up, that's why items in town are mostly while, because there are no monsters to drop them there, while any loot dropped by monster will be red in dungeon>.>

Yes, stat flag behaves erratically. I have workarounded the problem by the following code.

			int txt_color = UIS_SILVER;
			if (itm._iMagical == ITEM_QUALITY_MAGIC) txt_color = UIS_BLUE;
			if (itm._iMagical == ITEM_QUALITY_UNIQUE) txt_color = UIS_GOLD;
			if (itm._iStatFlag == false && ((itm._iMagical == ITEM_QUALITY_MAGIC) || (itm._iMagical == ITEM_QUALITY_UNIQUE))) txt_color = UIS_RED;
DrawString(out, label.text.c_str(), { label.pos.x, label.pos.y, label.width, height }, txt_color);

Chance4us:
Now is is perfect.

AJenbo (approved):

AJenbo:
And it even has nice commits

PR #2066, opened 2021-05-26
assign seed to quest items + rng cleanup

qndel:
@AJenbo I know you wanted to get rid of the whole searching for items on the ground by seed, but since you haven't done that yet, it's a simple fix we can remove later and makes picking bloodstones less annoying

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -458,7 +458,6 @@ void AddInitItems()
 		dItem[x][y] = ii + 1;
 
 		items[ii]._iSeed = AdvanceRndSeed();
-		SetRndSeed(items[ii]._iSeed);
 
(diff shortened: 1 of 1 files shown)


AJenbo:
Isn't this going to affect RNG?

qndel:
how?

julealgon:
What exactly is this "fixing"? Shouldn't we reserve that tag for actual bugs? Is this a bug?

AJenbo:
That you sometimes can't pick up unique quest items like the blood stones. Context was in discord and should have been put in the issue.

julealgon:
> That you sometimes can't pick up unique quest items like the blood stones. 

Ohh ok. Felt like we were just changing something without affecting behavior. Cool.

Issue #2035, opened 2021-05-22
🧹 Move `byte` from `engine.h` to `utils/stdcompat/cstddef.hpp`


glebm:
The `byte` type is currently defined in `engine.h`, which means we're pulling this header into lots of places often just for the `byte` type.

Like other backwards-compat types, it should be defined in an `stdcompat` header, `utils/stdcompat/cstddef.hpp` in this case.

PR #2067, opened 2021-05-27
Move byte from engine.h

kalebealvs:


kalebealvs:
This is related to issue https://github.com/diasurgical/devilutionX/issues/2035.
Sorry but I wasn't able to link it at the time the pull request was created.

AJenbo:
Thanks :)

Source/utils/stdcompat/cstddef.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#ifdef __has_include
+#if __has_include(<cstddef>)
```
glebm, line 4: The `<cstddef>` include is always present, no need to test for it

Source/utils/stdcompat/cstddef.hpp
```diff
@@ -0,0 +1,16 @@
...
+#include <cstddef>
+#if defined(__cplusplus) && __cplusplus >= 201703L
+namespace devilution {
+    using byte = std::byte;
```
glebm, line 8: We do not indent within a namespace, please use clang-format in the future

kalebealvs: I submitted the changes from the comments to my fork. Should I update it here?

AJenbo: Yes you need to make a new pull request since I already merged the original proposal.

Source/encrypt.cpp
```diff
@@ -4,10 +4,9 @@
  * Implementation of functions for compression and decompressing MPQ data.
  */
 #include <memory>
+#include <SDL.h>
```
glebm, line 7: This include should not be necessary here, what is it for?

AJenbo: `SDL_SwapLE32` is being utilized.

PR #2072, opened 2021-05-27
Replace all Auto Equip options with one option.

qqkookie:
Replace all Auto Equip options with one Auto Equip Items option.

New Option: Auto Equip Items to replace all Auto Equip XXX flags in `Diablo.ini` config file.
Old:
```
Auto Equip Weapons=1
Auto Equip Armor=0
Auto Equip Helms=1
Auto Equip Shields=1
Auto Equip Jewelry=0
```
New:
```
Auto Equip Items=Weapon, Helm, Shield
```
Valid item names are `Weapon, Armor, Helm, Shield, Jewelry` (all singular words) in any order, case ignored.
Currently, old Auto Equip XXX flags and source code are preserved for transition, but will be removed eventually later times.

Thank you for motivation from @julealgon on discussion thread #1804.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -537,6 +537,7 @@ static void SaveOptions()
 	setIniInt("Game", "Enemy Health Bar", sgOptions.Gameplay.bEnemyHealthBar);
 	setIniInt("Game", "Auto Gold Pickup", sgOptions.Gameplay.bAutoGoldPickup);
 	setIniInt("Game", "Adria Refills Mana", sgOptions.Gameplay.bAdriaRefillsMana);
+	setIniValue("Game", "Auto Equip Items", sgOptions.Gameplay.szAutoEquipItems);
 	setIniInt("Game", "Auto Equip Weapons", sgOptions.Gameplay.bAutoEquipWeapons);
 	setIniInt("Game", "Auto Equip Armor", sgOptions.Gameplay.bAutoEquipArmor);
 	setIniInt("Game", "Auto Equip Helms", sgOptions.Gameplay.bAutoEquipHelms);
@@ -627,6 +628,12 @@ static void LoadOptions()
 	sgOptions.Gameplay.bAutoEquipHelms = getIniBool("Game", "Auto Equip Helms", false);
 	sgOptions.Gameplay.bAutoEquipShields = getIniBool("Game", "Auto Equip Shields", false);
 	sgOptions.Gameplay.bAutoEquipJewelry = getIniBool("Game", "Auto Equip Jewelry", false);
+	getIniValue("Game", "Auto Equip Items", sgOptions.Gameplay.szAutoEquipItems, 99, "Weapon");
+	sgOptions.Gameplay.bAutoEquipWeapons |= IniFound(sgOptions.Gameplay.szAutoEquipItems, "Weapon");
+	sgOptions.Gameplay.bAutoEquipArmor |= IniFound(sgOptions.Gameplay.szAutoEquipItems, "Armor");
+	sgOptions.Gameplay.bAutoEquipHelms |= IniFound(sgOptions.Gameplay.szAutoEquipItems, "Helm");
+	sgOptions.Gameplay.bAutoEquipShields |= IniFound(sgOptions.Gameplay.szAutoEquipItems, "Shield");
+	sgOptions.Gameplay.bAutoEquipJewelry |= IniFound(sgOptions.Gameplay.szAutoEquipItems, "Jewelry");
 	sgOptions.Gameplay.bRandomizeQuests = getIniBool("Game", "Randomize Quests", true);
 	sgOptions.Gameplay.bShowMonsterType = getIniBool("Game", "Show Monster Type", false);
 	sgOptions.Gameplay.bDisableCripplingShrines = getIniBool("Game", "Disable Crippling Shrines", false);
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -591,6 +591,14 @@ bool AutoEquip(int playerNumber, const ItemStruct &item, bool persistItem)
  */
 bool AutoEquipEnabled(const PlayerStruct &player, const ItemStruct &item)
 {
+	/*
+	// Use this when Gameplay.bAutoEquipXXX flags are removed.
+	return ((item.isWeapon() && player._pClass != HeroClass::Monk && IniFound(sgOptions.Gameplay.szAutoEquipItems, "Weapon"))
+		|| (item.isArmor() && IniFound(sgOptions.Gameplay.szAutoEquipItems, "Armor"))
+		|| (item.isHelm() && IniFound(sgOptions.Gameplay.szAutoEquipItems, "Helm"))
+		|| (item.isShield() && IniFound(sgOptions.Gameplay.szAutoEquipItems, "Shield"))
+		|| (item.isJewelry() && IniFound(sgOptions.Gameplay.szAutoEquipItems, "Jewelry")));
+	*/
 	if (item.isWeapon()) {
 		// Monk can use unarmed attack as an encouraged option, thus we do not automatically equip weapons on him so as to not
 		// annoy players who prefer that playstyle.
--- a/Source/options.h
+++ b/Source/options.h
@@ -92,6 +92,8 @@ struct GameplayOptions {
 	bool bAutoGoldPickup;
 	/** @brief Recover mana when talking to Adria. */
 	bool bAdriaRefillsMana;
+	/** @brief Automatically attempt to equip items when picking them up. */
+	char szAutoEquipItems[100];
 	/** @brief Automatically attempt to equip weapon-type items when picking them up. */
 	bool bAutoEquipWeapons;
 	/** @brief Automatically attempt to equip armor-type items when picking them up. */
--- a/Source/storm/storm.cpp
+++ b/Source/storm/storm.cpp
@@ -243,8 +243,8 @@ const char *strcasestr(const char *haystack, const char *needle)
 			return NULL;
 
 		for (pp = needle + 1, qq = look + 1; tolower(*pp) == tolower(*qq); pp++, qq++) {
-		    if (!*pp)	// match found!
-			return (look);
+			if (!*pp)	// match found!
+				return (look);
 		}
 	}
 	return NULL;
--- a/Source/storm/storm.cpp
+++ b/Source/storm/storm.cpp
@@ -216,13 +216,14 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
 }
 
 /**
- * @brief Find item in comma-sparated list
+ * @brief Check item exists in comma-sparated list
  */
 bool IniFound(const char *list, const char *item)
 {
 	const char *look = strcasestr(list, item);
 	if (!look)
 		return false;
+
 	look += strlen(item);
 	while (*look == ' ')
 		look++;
@@ -235,19 +236,20 @@ bool IniFound(const char *list, const char *item)
 const char *strcasestr(const char *haystack, const char *needle)
 {
 	const char *look, *pp, *qq;
+	assert(haystack && needle);
 
 	for (look = haystack; *look; look++) {
 		while (*look && (tolower(*look) != tolower(*needle)))
 			look++;
 		if (!*look)
-			return NULL;
+			return nullptr;
 
-		for (pp = needle + 1, qq = look + 1; tolower(*pp) == tolower(*qq); pp++, qq++) {
+		for (pp = needle+1, qq = look+1; tolower(*pp) == tolower(*qq); pp++, qq++) {
 			if (!*pp)	// match found!
 				return (look);
 		}
 	}
-	return NULL;
+	return nullptr;
 }
 
 DWORD SErrGetLastError()
--- a/Source/storm/storm.cpp
+++ b/Source/storm/storm.cpp
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
 	getIni().SetDoubleValue(keyname, valuename, value);
 }
 
+/**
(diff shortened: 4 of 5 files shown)


Source/inv.cpp
```diff
@@ -591,6 +591,14 @@ bool AutoEquip(int playerNumber, const ItemStruct &item, bool persistItem)
  */
 bool AutoEquipEnabled(const PlayerStruct &player, const ItemStruct &item)
 {
+	/*
+	// Use this when Gameplay.bAutoEquipXXX flags are removed.
```
glebm, line 595: Please remove commented code. I also do not think the flags should be removed.

Source/diablo.cpp
```diff
@@ -627,6 +628,12 @@ static void LoadOptions()
 	sgOptions.Gameplay.bAutoEquipHelms = getIniBool("Game", "Auto Equip Helms", false);
 	sgOptions.Gameplay.bAutoEquipShields = getIniBool("Game", "Auto Equip Shields", false);
 	sgOptions.Gameplay.bAutoEquipJewelry = getIniBool("Game", "Auto Equip Jewelry", false);
+	getIniValue("Game", "Auto Equip Items", sgOptions.Gameplay.szAutoEquipItems, 99, "Weapon");
+	sgOptions.Gameplay.bAutoEquipWeapons |= IniFound(sgOptions.Gameplay.szAutoEquipItems, "Weapon");
```
glebm, line 632: Perhaps there should be an `All` shorthand as well?

Source/options.h
```diff
@@ -92,6 +92,8 @@ struct GameplayOptions {
 	bool bAutoGoldPickup;
 	/** @brief Recover mana when talking to Adria. */
 	bool bAdriaRefillsMana;
+	/** @brief Automatically attempt to equip items when picking them up. */
+	char szAutoEquipItems[100];
```
glebm, line 96: `std::string`?

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
 	getIni().SetDoubleValue(keyname, valuename, value);
 }
 
+/**
+ * @brief Check item exists in comma-sparated list
```
glebm, line 219:
```suggestion
 * @brief Check item exists in comma-separated list
```

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
 	getIni().SetDoubleValue(keyname, valuename, value);
 }
 
+/**
+ * @brief Check item exists in comma-sparated list
+ */
+bool IniFound(const char *list, const char *item)
```
glebm, line 221: 1. There is no need for this to be case-insensitive.
2. Instead of calling `IniFound` for every item, it'd be better to loop over the comma-separated items once.

Source/storm/storm.h
```diff
@@ -265,6 +265,9 @@ int getIniInt(const char *keyname, const char *valuename, int defaultValue);
 void setIniInt(const char *keyname, const char *valuename, int value);
 void setIniFloat(const char *keyname, const char *valuename, float value);
 
+bool IniFound(const char *list, const char *item);
```
glebm, line 268: There is no need for these to be public.

glebm (changes requested):

glebm:
You can cherry-pick this commit for string splitting (and add whitespace skipping): https://github.com/glebm/devilutionX/commit/e97944c773f6dc92dc7643800ec8f17d6cd8dc27

Source/diablo.cpp
qqkookie: Yes, `x |= y` is shorthand of  `x = x | y`.

Source/options.h
qqkookie: I tried initially, but  get/setIniValue() prefers char []. And most other stings in options structures use char [], not std::string.

Source/inv.cpp
qqkookie: The very purpose of this change is to remove these flags from `Diablo.ini` and `struct GamePlayOption` eventually.

Source/storm/storm.h
qqkookie: Hmmm.. if it is not public, inv.cpp, diablo.cpp will get undefined function error.

Source/storm/storm.cpp
qqkookie: Got it.

Source/storm/storm.cpp
qqkookie: Case-intensively is not must but it helps. 
For long list like spell names, looping is must. But five items are not worth it.

AJenbo:
Two downsides to this are:
- This changes the ini compared to what was in 1.2 so will be a surprise to anyone that had already set it and upgraded.
- Typing out the names of the options is a lot harder and not very easy to figure out just looking at the default ini. I for one probably couldn't even spell the options correctly.

If this change should happen the in-game options menu should at the very least come first.

Source/inv.cpp
qndel: the very purpose of git is not having to leave commented code

Source/diablo.cpp
```diff
@@ -537,6 +537,7 @@ static void SaveOptions()
 	setIniInt("Game", "Enemy Health Bar", sgOptions.Gameplay.bEnemyHealthBar);
 	setIniInt("Game", "Auto Gold Pickup", sgOptions.Gameplay.bAutoGoldPickup);
 	setIniInt("Game", "Adria Refills Mana", sgOptions.Gameplay.bAdriaRefillsMana);
+	setIniValue("Game", "Auto Equip Items", sgOptions.Gameplay.szAutoEquipItems);
```
julealgon, line 540: I really, _really_ dislike the fact we are using strings here. 

I think it is a bad idea to do this from an architecture standpoint, even if it makes the ini simpler. I think without proper, decent support for enum flags (or bitsets) to be mapped directly, this is not worth the trouble: it will simplify the ini, but make the code vastly worse.

Source/diablo.cpp
julealgon, line 632: I also really don't like the fact that we have "one setting in the model used to set another setting in it". This seems like a design smell. The setting should be a single value (flags enum or bitset).

Source/storm/storm.cpp
julealgon, line 221: `IniFound` is not a good function name IMHO. If the intention is to have something that checks whether or not something is in the ini file, I'd suggest `IsInIni` or `ExistsInIni` or even `ExistsInConfiguration` (abstract the fact we use "ini" and not something else).

EDIT: Looking at the implementation, I see _zero_  mentions of INI or even configuration for that matter. This seems to be a general-purpose "search on a list"-type of method. In that case, the function name is even worse here, introducing unnecessary coupling to the config system where there is none.

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
 	getIni().SetDoubleValue(keyname, valuename, value);
 }
 
+/**
+ * @brief Check item exists in comma-sparated list
+ */
+bool IniFound(const char *list, const char *item)
+{
+	const char *look = strcasestr(list, item);
```
julealgon, line 223: What does `look` mean?

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+	look += strlen(item);
+	while (*look == ' ')
+		look++;
+	return (bool) strchr(",;:", *look); // delimters including '\0'
+}
+
+/**
+ * @brief Find needle pattern in haystack, ignoring case
+ */
+const char *strcasestr(const char *haystack, const char *needle)
```
julealgon, line 236: Why is this all lower case?

julealgon, line 236: Not a fan of the "let's use there informal phrases/expressions in code" thing. Please rename the arguments to something  more appropriate.

julealgon, line 236: Can't we use `string`s for new code?

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+{
+	const char *look, *pp, *qq;
+	assert(haystack && needle);
+
+	for ( look = haystack; *look; look++ ) {
```
julealgon, line 241: Format your code. There should be no space between the `(` and the members inside the `for`.

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+		if (tolower(*look) == tolower(*needle)) {
+			pp = needle + 1; qq = look + 1;
+			while (*pp && (tolower(*pp) == tolower(*qq)))
+				pp++, qq++;
+			if (!*pp)
+				return look;	// match found!
+		}
```
julealgon, line 248: I cannot believe there is no such method already in stdlib. Is there really not?

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+	const char *look, *pp, *qq;
+	assert(haystack && needle);
+
+	for ( look = haystack; *look; look++ ) {
+		if (tolower(*look) == tolower(*needle)) {
+			pp = needle + 1; qq = look + 1;
+			while (*pp && (tolower(*pp) == tolower(*qq)))
+				pp++, qq++;
+			if (!*pp)
+				return look;	// match found!
```
julealgon, line 247: Prefer adding comments on top of the affected line, not to the side. Goes for other occurrences as well.

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+ * @brief Find needle pattern in haystack, ignoring case
+ */
+const char *strcasestr(const char *haystack, const char *needle)
+{
+	const char *look, *pp, *qq;
+	assert(haystack && needle);
+
+	for ( look = haystack; *look; look++ ) {
+		if (tolower(*look) == tolower(*needle)) {
+			pp = needle + 1; qq = look + 1;
```
julealgon, line 243: Please don't do these "multiple statements per line" non-standard thing.

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+		look++;
+	return (bool) strchr(",;:", *look); // delimters including '\0'
+}
+
+/**
+ * @brief Find needle pattern in haystack, ignoring case
+ */
+const char *strcasestr(const char *haystack, const char *needle)
+{
+	const char *look, *pp, *qq;
```
julealgon, line 238: What is `pp` and `qq`? Can we please name these variables better?

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+	assert(haystack && needle);
+
+	for ( look = haystack; *look; look++ ) {
+		if (tolower(*look) == tolower(*needle)) {
+			pp = needle + 1; qq = look + 1;
+			while (*pp && (tolower(*pp) == tolower(*qq)))
```
julealgon, line 244: We already have a `tolower(x) == tolower(y)` check above. Can we do this in a way where we don't have to repeat the same check twice?

Source/storm/storm.cpp
```diff
@@ -215,6 +215,41 @@ void setIniFloat(const char *keyname, const char *valuename, float value)
...
+				pp++, qq++;
+			if (!*pp)
```
julealgon, line 246: I think we have agreed to not rely on C's "everything is a bool" thing and do explicit checks.

Source/storm/storm.h
```diff
@@ -265,6 +265,9 @@ int getIniInt(const char *keyname, const char *valuename, int defaultValue);
 void setIniInt(const char *keyname, const char *valuename, int value);
 void setIniFloat(const char *keyname, const char *valuename, float value);
 
+bool IniFound(const char *list, const char *item);
+const char *strcasestr(const char *haystack, const char *needle);
```
julealgon, line 269: Why are these methods part of `storm.h`?

julealgon (changes requested):
While I obviously agree that it would be better to be able to define the values in a unified way, using a single property with enum-flags-like behavior, the way this is achieved here just introduces a mess and too many hardcodes for it to be useful.

Until we have a decent enum-flags approach that is general purpose, and doesn't require hardcoding values, and all these custom string functions, I'm 100% against touching this.

Without config validation, this will just make it harder to reason about the changes, as users will not be able to tell what values are possible and won't get any feedback when wrong values are passed in.

AJenbo:
@julealgon since you seam to fundamentally not agree with this change I would suggest that it isn't helping much to nit pick it for code style issue. Maybe just a general comment regarding checking the project code style before submitting another rewrite. Making a large amount (5+) of comments can often feel overwhelming, especially to people that are new to contributing to the project.

AJenbo:
@qqkookie you can find our code style guide here: https://github.com/diasurgical/devilutionX/wiki/Code-Style

Not everything is defined in it and not all code comply with it but we are actively working to improve that. There is also a cang-tidy definition, but we are currently in the process of applying this so the code base so it's currently not advisable to run this tool automatically.

Source/storm/storm.cpp
qndel: because copy/pasted from the internet :p

Source/storm/storm.cpp
qndel: also putting a needle next to pp makes me uncomfortable

PR #2073, opened 2021-05-27
Using Point for Control functions

kalebealvs:
Honestly, I'm not good with naming schemes. I tried to get the sense from the previous code coordinates naming and some of the contexts, but I might have understood something wrong, so if you have any suggestions, I'm all "ears¨.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -226,14 +226,13 @@ spell_id SpellPages[6][7] = {
 /**
  * Draw spell cell onto the given buffer.
  * @param out Output buffer
- * @param xp Buffer coordinate
- * @param yp Buffer coordinate
+ * @param position Buffer coordinates
  * @param cel The CEL sprite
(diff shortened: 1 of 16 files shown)


Source/control.cpp
```diff
@@ -538,10 +543,10 @@ void ClearPanel()
 	pinfoflag = false;
 }
 
-void DrawPanelBox(const CelOutputBuffer &out, int x, int y, int w, int h, int sx, int sy)
+void DrawPanelBox(const CelOutputBuffer &out, Point sourcePosition, int w, int h, Point targetPosition)
```
AJenbo, line 546: We might want to use a SDL_rect here since it has width and height also.
I think `CelOutputBuffer` even understands SDL_rect.

AJenbo (approved):

Source/control.cpp
kalebealvs: It understands Point but no SDL_Rect.

PR #2074, opened 2021-05-28
use upstream devkitPro cmake machinery

WinterMute:
We finally expended a fair bit of sweat, blood, tears and swearing at inanimate objects and produced our own cmake machinery then used it to build switch devilutionx. This is the result. What we've gone for is an approach where running the appropriate <triplet>-cmake does the right thing rather than just providing a toolchain file. In the case of switch that will be /opt/devkitpro/portlibs/switch/bin/aarch64-none-elf-cmake which you can see in the .circleci changes. The packages we've added for this can be seen at https://github.com/devkitPro/pacman-packages/tree/master/cmake

There are still some issues that should probably be figured out. The current cmake setup appears to somehow set NONET ON for switch but I only seem to be able to get that to work when defined externally. Can someone explain why this might be?

[This section of the patch](https://github.com/diasurgical/devilutionX/compare/master...WinterMute:devkitPro-cmake?expand=1#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR836-R843) should actually be completely unnecessary and is in fact something that could potentially cause future issues. This should really be covered by [FindSDL2_ttf.cmake](https://github.com/diasurgical/devilutionX/blob/master/CMake/FindSDL2_ttf.cmake) which uses pkg-config but then discards the information obtained that way & uses a subset to find just libSDL2_ttf.a. I'm still not familiar enough with cmake to fix that myself unfortunately but doing so should theoretically help with the other platforms where libraries are being added manually and workarounds are being used to compensate for not just using pkg-config when available.


glebm:
Great! Looking at https://github.com/devkitPro/pacman-packages/blob/master/cmake/devkita64/devkitA64.cmake, it should do this for pkg-config to work:

```cmake
set(PKG_CONFIG_EXECUTABLE ${TOOL_PREFIX}pkg-config)
```

CMakeLists.txt
```diff
@@ -29,6 +29,14 @@ option(RUN_TESTS "Build and run tests" OFF)
 option(ENABLE_CODECOVERAGE "Instrument code for code coverage (only enabled with RUN_TESTS)" OFF)
 option(USE_GETTEXT "Build translation files using gettext" OFF)
 
+if(NINTENDO_SWITCH)
```
glebm, line 32: This should be in the same place where it previously was (just before `if(VITA)`)

WinterMute:
> Great! Looking at https://github.com/devkitPro/pacman-packages/blob/master/cmake/devkita64/devkitA64.cmake, it should do this for pkg-config to work:
> 
> ```cmake
> set(PKG_CONFIG_EXECUTABLE ${TOOL_PREFIX}pkg-config)
> ```

That's done here https://github.com/devkitPro/pacman-packages/blob/master/cmake/switch/Switch.cmake#L21-L25. devkitA64.cmake is a common base for bare-metal aarch64 platforms, not a final toolchain file.

glebm:
Ah, I see! That duplicates some of CMake's `FindPkgConfig` functionality, which also checks for the binary's existence and issues a warning. Instead of using `find_program` and issuing a warning there, it should just set the path. That's a minor nit anyway.

CMakeLists.txt
WinterMute: This was part of my fiddling to try and get this to work identically to the current setup which sets NONET ON [here](https://github.com/diasurgical/devilutionX/blob/master/CMake/switch/switch_defs.cmake#L1) and not externally as I've had to do with this. I guess it's because the current setup includes this file via the toolchain file and my changes mean it's included via CMakeLists.txt?

glebm:
>  CMake/switch/switch_defs.cmake → switch_defs.cmake 

Why was this moved to top-level? All CMake stuff should be in `CMake/`.

CMakeLists.txt
glebm: I don't think the current setup includes it via the toolchain file!

glebm: I think the issue may have been introduced in the commit that added the `DISABLE_TCP` option.

glebm: Hmm, nevermind that, the `DISABLE_TCP` stuff looks OK.

Anyway, please move this block where it was before and we can investigate the NONET stuff separately after merging this.

glebm (approved):

glebm:
Thanks for adding official upstream support for CMake and cleaning this up here!

CMakeLists.txt
AJenbo: preferably we would just get libsodium/zt compiling on devkitpro and then not have to disable network :)

PR #2076, opened 2021-05-28
♻️ Split getTextColor(bool) into 2 functions

glebm:
This is a bit more readable than using a bool argument.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1156,7 +1156,7 @@ void DrawInfoBox(const CelOutputBuffer &out)
 				strcpy(infostr, myPlayer.HoldItem._iIName);
 			else
 				strcpy(infostr, myPlayer.HoldItem._iName);
-			infoclr = myPlayer.HoldItem.getTextColor(true);
+			infoclr = myPlayer.HoldItem.getTextColor();
 		}
 	} else {
 		if (pcursitem != -1)
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -2001,7 +2001,7 @@ char CheckInvHLight()
 		int nGold = pi->_ivalue;
 		strcpy(infostr, fmt::format(ngettext("{:d} gold piece", "{:d} gold pieces", nGold), nGold).c_str());
 	} else {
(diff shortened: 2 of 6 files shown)


AJenbo (approved):

julealgon:
> This is a bit more readable than using a bool argument.

I disagree. Your version makes the code more misleading. People calling `getTextColor` might wonder why it is not returning the proper color for when the item cannot be equiped.

Decentralizing the behavior (by splitting it) is a bad idea IMHO.

AJenbo:
Boolean parameters are almost by definition violating SRP. It also often makes the code harder to read. Consider this:
```cpp
if (hovering) {
   color = item.getColor(true);
}
```
What even does the boolean do? You could easily think it was for getting a highlight color since the code surrounding it dealing with a hover state?

julealgon:
> Boolean parameters are almost by definition violating SRP.

I agree with that part, but I think the change in this particular case transfers too much responsibility to the caller: responsibility that should be inside the abstraction, not outside. Perhaps the solution was to get rid of the exposed condition in the first place and do it in a different way?

This is why I mentioned earlier that the current abstraction is bad. Instead of having a method that returns the color, we should think about a way that encapsulates the whole behavior of writing the item and have this logic as a detail of that.

Either the item itself should have some sort of "writeTo()" method (not perfect, but better than current state), or we should delegate the printing behavior to a separate class that handles how items are displayed, and these color decisions would be hidden from callers in there.

> It also often makes the code harder to read. Consider this:

Yeah, I know boolean arguments are bad and our current version of C++ doesn't allow for named arguments. Usually, either you name the argument explicitly, or the IDE does it, or you create a variable to provide meaning to the boolean value and use that.

TL;DR, I think this solves the problem in a poor way that actually introduces other downsides.

AJenbo:
> Either the item itself should have some sort of "writeTo()".

 I think that would be even worse.

> we should delegate the printing behavior to a separate class that handles how items are displayed, and these color decisions would be hidden from callers in there.

Yeah, basically `UI.infobox.display(item);`

PR #2079, opened 2021-05-28
:fire: Remove done boolean from LineClear

AJenbo:
Alternately I could change the wile condition to `while (position != endPoint && (position == startPoint || Clear(entity, position))) {`

P.s. the two paths in the big if-else are identical except for the operational axis. I wonder if we could do something clever to merge that. Some thing like this maybe?
```cpp
bool LineClear(bool (*Clear)(int, Point), int entity, Point startPoint, Point endPoint)
{
	int d;
	int incD, dincD, dincH;

	Point position = startPoint;

	int d1 = endPoint.x - position.x;
	int d2 = endPoint.y - position.y;
	bool horizontal = abs(d1) > abs(d2);
	if (horizontal)
		std::swap(d1, d2);
	if (d1 < 0) {
		std::swap(position, endPoint);
		d1 = -d1;
		d2 = -d2;
	}
	if (d2 > 0) {
		d = 2 * d2 - d1;
		dincD = 2 * d2;
		dincH = 2 * (d2 - d1);
		incD = 1;
	} else {
		d = 2 * d2 + d1;
		dincD = 2 * d2;
		dincH = 2 * (d1 + d2);
		incD = -1;
	}
	while (position != endPoint && (position == startPoint || Clear(entity, position))) {
		if ((d <= 0) ^ (incD < 0)) {
			d += dincD;
		} else {
			d += dincH;
			if (horizontal)
				position.y += incD;
			else
				position.x += incD;
		}
		if (horizontal)
			position.x++;
		else
			position.y++;
	}

	return position == endPoint;
}
```

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4758,69 +4758,43 @@ bool LineClearMissile(Point startPoint, Point endPoint)
 
 bool LineClear(bool (*Clear)(int, Point), int entity, Point startPoint, Point endPoint)
 {
-	int d;
-	int xincD, yincD, dincD, dincH;
-	bool done = false;
-
 	Point position = startPoint;
 
-	int dx = endPoint.x - position.x;
-	int dy = endPoint.y - position.y;
-	if (abs(dx) > abs(dy)) {
-		if (dx < 0) {
-			std::swap(position, endPoint);
-			dx = -dx;
-			dy = -dy;
-		}
-		if (dy > 0) {
-			d = 2 * dy - dx;
-			dincD = 2 * dy;
-			dincH = 2 * (dy - dx);
-			yincD = 1;
-		} else {
-			d = 2 * dy + dx;
-			dincD = 2 * dy;
-			dincH = 2 * (dx + dy);
-			yincD = -1;
-		}
-		while (position != endPoint) {
-			if ((d <= 0) ^ (yincD < 0)) {
-				d += dincD;
-			} else {
-				d += dincH;
-				position.y += yincD;
-			}
-			position.x++;
-			done = position != startPoint && !Clear(entity, position);
-		}
-	} else {
-		if (dy < 0) {
-			std::swap(position, endPoint);
-			dy = -dy;
-			dx = -dx;
-		}
-		if (dx > 0) {
-			d = 2 * dx - dy;
-			dincD = 2 * dx;
-			dincH = 2 * (dx - dy);
-			xincD = 1;
-		} else {
-			d = 2 * dx + dy;
-			dincD = 2 * dx;
-			dincH = 2 * (dy + dx);
-			xincD = -1;
-		}
-		while (!done && position != endPoint) {
-			if ((d <= 0) ^ (xincD < 0)) {
-				d += dincD;
-			} else {
-				d += dincH;
-				position.x += xincD;
-			}
-			position.y++;
-			done = position != startPoint && !Clear(entity, position);
-		}
+	int d1 = endPoint.x - position.x;
+	int d2 = endPoint.y - position.y;
+
+	bool horizontal = abs(d1) > abs(d2);
+
+	if (!horizontal)
+		std::swap(d1, d2);
+
+	if (d1 < 0) {
+		std::swap(position, endPoint);
+		d1 = -d1;
+		d2 = -d2;
 	}
+
+	int incD = (d2 > 0) ? 1 : -1;
+	int d = 2 * d2 - (d1 * incD);
+	int dincD = 2 * d2;
+	int dincH = 2 * (d2 - (d1 * incD));
+
+	while (position != endPoint && (position == startPoint || Clear(entity, position))) {
+		if ((d <= 0) ^ (incD < 0)) {
+			d += dincD;
+		} else {
+			d += dincH;
+			if (!horizontal)
+				position.x += incD;
+			else
+				position.y += incD;
+		}
+		if (horizontal)
+			position.x++;
+		else
+			position.y++;
+	}
+
 	return position == endPoint;
 }
 


qndel:
what about
```cpp
bool LineClear(bool (*Clear)(int, Point), int entity, Point startPoint, Point endPoint)
{
	int d;
	int incD, dincD, dincH;

	Point position = startPoint;

	int d1 = endPoint.x - position.x;
	int d2 = endPoint.y - position.y;
	bool horizontal = abs(d1) > abs(d2);
	if (horizontal)
		std::swap(d1, d2);
	if (d1 < 0) {
		std::swap(position, endPoint);
		d1 = -d1;
		d2 = -d2;
	}
	if (d2 > 0)
		incD = 1;
	else
		incD = -1;
	d = 2 * d2 - (d1 * incD);
	dincD = 2 * d2;
	dincH = 2 * (d2 - (d1 * incD));
	while (position != endPoint && (position == startPoint || Clear(entity, position))) {
		int &coord = (horizontal ? position.x : position.y);
		if ((d <= 0) ^ (incD < 0)) {
			d += dincD;
		} else {
			d += dincH;
			coord += incD;
		}
		coord++;
	}

	return position == endPoint;
}
```

AJenbo:
@qndel nice, then we can even move all the declerations to where the assignments:
```
bool LineClear(bool (*Clear)(int, Point), int entity, Point startPoint, Point endPoint)
{
	Point position = startPoint;

	int d1 = endPoint.x - position.x;
	int d2 = endPoint.y - position.y;
	bool horizontal = abs(d1) > abs(d2);
	if (horizontal)
		std::swap(d1, d2);
	if (d1 < 0) {
		std::swap(position, endPoint);
		d1 = -d1;
		d2 = -d2;
	}
	int incD = 1;
	if (d2 <= 0)
		incD = -1;
	int d = 2 * d2 - (d1 * incD);
	int dincD = 2 * d2;
	int dincH = 2 * (d2 - (d1 * incD));
	while (position != endPoint && (position == startPoint || Clear(entity, position))) {
		int &coord = (horizontal ? position.x : position.y);
		if ((d <= 0) ^ (incD < 0)) {
			d += dincD;
		} else {
			d += dincH;
			coord += incD;
		}
		coord++;
	}

	return position == endPoint;
}
```

julealgon:
Small comment before I review (after dinner): I noticed you are using `:fire:` here instead of `🔥`. Did you have issues again with the unicode characters?

AJenbo:
> Small comment before I review (after dinner): I noticed you are using `:fire:` here instead of `🔥`. Did you have issues again with the unicode characters?

Typing :fire: is faster the Googling the emoji. Don't know if there is a fancy way to enter it, but not important to me.

AJenbo:
Just play-tested this, and we defiantly broke something, got men will just run around you and not attack :(

Source/monster.cpp
```diff
@@ -4758,69 +4758,43 @@ bool LineClearMissile(Point startPoint, Point endPoint)
...
-			dincD = 2 * dy;
-			dincH = 2 * (dy - dx);
-			yincD = 1;
+	int d1 = endPoint.x - position.x;
+	int d2 = endPoint.y - position.y;
+
+	bool horizontal = abs(d1) > abs(d2);
+
+	if (!horizontal)
+		std::swap(d1, d2);
```
julealgon, line 4769: I have no idea why this happens. Can you add comments to the method to explain what it does? Or rename it to something more specific?

Source/monster.cpp
```diff
@@ -4758,69 +4758,43 @@ bool LineClearMissile(Point startPoint, Point endPoint)
...
+
+	if (d1 < 0) {
+		std::swap(position, endPoint);
+		d1 = -d1;
+		d2 = -d2;
```
julealgon, line 4774: Can we define an `abs` overload for `Point` to do this?

Source/monster.cpp
```diff
@@ -4758,69 +4758,43 @@ bool LineClearMissile(Point startPoint, Point endPoint)
...
+	}
+
+	int incD = (d2 > 0) ? 1 : -1;
+	int d = 2 * d2 - (d1 * incD);
+	int dincD = 2 * d2;
+	int dincH = 2 * (d2 - (d1 * incD));
```
julealgon, line 4780: These variable names are very poor IMHO. Can we rename them to be more specific?

Source/monster.cpp
```diff
@@ -4758,69 +4758,43 @@ bool LineClearMissile(Point startPoint, Point endPoint)
...
+
+	while (position != endPoint && (position == startPoint || Clear(entity, position))) {
```
julealgon, line 4782: What is the expectation of `Clear`? Shouldn't we rename it to `IsClear`?

Source/monster.cpp
```diff
@@ -4758,69 +4758,43 @@ bool LineClearMissile(Point startPoint, Point endPoint)
 
 bool LineClear(bool (*Clear)(int, Point), int entity, Point startPoint, Point endPoint)
 {
-	int d;
-	int xincD, yincD, dincD, dincH;
-	bool done = false;
-
 	Point position = startPoint;
```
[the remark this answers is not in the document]
AJenbo: No, just like the original code we need to compare check if `position` is still the same as `startPoint` once we hit the end of the function (`x/yorg` and `x/y1` in the original).

Source/monster.cpp
AJenbo: It's swapping the x/y axis in order to move down instead of to the right in the grid.

Source/monster.cpp
AJenbo: we already have negation, `position = -position` but d1 and d2 represents the grid more so then the point so I don't think the would work well redefined as a point.

Source/monster.cpp
AJenbo: Do you have any suggestions?

Source/monster.cpp
AJenbo: It checks the individual points in the line. If we wanted to follow the same naming as the function then it should be called `PointClear`, else we should also rename LineClear to `IsLineClear` and the same for all the related functions that follow the same naming. But I think such a change belongs in it's own PR so we don't clutter this with multiple files and case conflict hell for this very isolated PR.

Source/monster.cpp
julealgon: You misunderstood my comment. I'm saying I don't understand what the whole function is doing, so it's hard to evaluate the logic as a whole (including these swaps).

It's hard to provide feedback if the initial logic was not clear.

Source/monster.cpp
julealgon: I don't know what the method does and I don't know what the poor names mean, so no, not for now. I was hoping you would know of better names since you seem to know what's going on here.

Source/monster.cpp
AJenbo: It checks to see if there is a line free of obstacles between two point. Often used when a monster checks to see if it can shoot at the player, but also for some of the AI route planning logic.

Source/monster.cpp
AJenbo: If I could I would have. The internals is still a bit cryptic, but at least now it's not repeated 4 times.

Source/monster.cpp
julealgon: Ah ok. Do you think we can include comments to the method as part of this change? Or on a previous commit?

julealgon: BTW, it puzzles me why we do any swaps of anything if the purpose is to check for a clear line?

AJenbo: > Ah ok. Do you think we can include comments to the method as part of this change? Or on a previous commit?

Defiantly, I just though the functions porous was obvious :joy:

AJenbo: > BTW, it puzzles me why we do any swaps of anything if the purpose is to check for a clear line?

The loop expects to walk in a positive direction with most movement being horizontal. This limits it to 1/4 of all possible directions. So in order to be able to do arbitrary moves we do up to two swaps (the original had one unrolled and an `if-else`) to flip the coordinate system in a way that fits the way the loop works.

AJenbo: image

AJenbo:
Diving deeper in to this function makes me think it's absolutely terrible. When moving in a negative direction it doesn't check the start or end position, but when moving in a positive direction it checks the end position.

It also doesn't align with the path that arrows actually travel; leading to scenarios where the monster thinks it can hit you but the arrow just hits a wall instead.

PR #2084, opened 2021-05-30
Partial rng replace

qndel:
Replaces crappy vanilla rng with a decent one for everything that doesn't break save compatibility (At least I hope so :D)

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -12,6 +12,7 @@
  */
 
 #include <array>
+#include <random>
 
 #include "engine/render/common_impl.h"
 #include "lighting.h"
@@ -35,6 +36,8 @@ const uint32_t RndInc = 1;
  * Specifies the multiplier used in the Borland C/C++ pseudo-random number generator algorithm.
  */
 const uint32_t RndMult = 0x015A4E35;
+
+std::mt19937 rng;
 
 CelSprite LoadCel(const char *pszName, int width)
 {
@@ -246,6 +249,24 @@ int32_t GenerateRnd(int32_t v)
 	if (v < 0xFFFF)
 		return (AdvanceRndSeed() >> 16) % v;
 	return AdvanceRndSeed() % v;
+}
+
+void SetRndSeed_rngv2(int32_t s)
+{
+	rng.seed(s);
+}
+
+int32_t GetRndSeed_rngv2()
+{
+	std::uniform_int_distribution<int32_t> dist(0, INT_MAX);
+	return dist(rng);
+}
+
+int32_t GenerateRnd_rngv2(int32_t v)
+{
+	if (v <= 0)
+		return 0;
+	return GetRndSeed_rngv2() % v;
 }
 
 size_t GetFileSize(const char *pszName)
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -252,22 +252,22 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
-void SetRndSeed_rngv2(int32_t s)
+void SetRndSeedV2(int32_t s)
 {
 	rng.seed(s);
 }
 
-int32_t GetRndSeed_rngv2()
+int32_t GetRndSeedV2()
 {
 	std::uniform_int_distribution<int32_t> dist(0, INT_MAX);
 	return dist(rng);
 }
 
-int32_t GenerateRnd_rngv2(int32_t v)
+int32_t GenerateRndV2(int32_t v)
 {
 	if (v <= 0)
 		return 0;
-	return GetRndSeed_rngv2() % v;
+	return GetRndSeedV2() % v;
 }
 
 size_t GetFileSize(const char *pszName)
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -13,6 +13,7 @@
 
 #include <array>
 #include <random>
+#include <climits>
 
 #include "engine/render/common_impl.h"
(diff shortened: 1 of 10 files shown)


Source/engine.cpp
```diff
@@ -248,6 +252,24 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
+void SetRndSeed_rngv2(int32_t s)
```
AJenbo, line 255: `SetRndSeed_rngv2` isn't proper camel case and a bit redundant. Could cal it `SetRndSeedV2` or `SetStrongRndSeed`

qndel: well it's fixable with 1 global replace, my goal was a name that was unique enough so a search would return only my stuff instead of 100k stuff from asio/whatever xD

AJenbo: > well it's fixable with 1 global replace

Do it :)

AJenbo:
Is this algo guaranteed to give the same sequence when setting the same seed?

ephphatha:
[std::mt19937 is a default set of arguments for the mersenne twister engine](https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine), `rng` is guaranteed to give the same sequence when (re-)initialised with the same seed. I believe that distributions are guaranteed to map the same generated number to the same output within their interval but I can't find documentation covering this specifically.

Running a quick test shows that calling the distribution with different params doesn't affect the generator, and a uniform_int_distribution appears to be consistent for the same generator values.

In the following output all three lines were generated with the default mt19937 seed and a distribution having a default interval of [0, 100].
The top line used the interval [333, 999] for positions 0, 3, 6, 9, 12, 15, 18
The middle line used the same interval as the top line for positions 0, 6, 12, 18 and the same interval as the bottom line for positions 3, 9, 15
The bottom line used the interval [1, 4] for positions 0, 3, 6, 9, 12, 15, 18
```
547, 20, 50,647, 63, 43,616, 39, 83,950, 86, 22,635, 91, 15,589, 89, 56,650, 32
  1, 20, 50,647, 63, 43,  2, 39, 83,950, 86, 22,  1, 91, 15,589, 89, 56,  2, 32
  1, 20, 50,  2, 63, 43,  2, 39, 83,  4, 86, 22,  1, 91, 15,  1, 89, 56,  2, 32
```

The same numbers were generated in the same order, and mapped to the same output for a given interval even when the distribution params varied between runs (cols 3-6 for example).

AJenbo:
My biggest worry is that there could be a difference between two std implementations, but sounds like it's part of the spec to be stable.

Source/engine.cpp
```diff
@@ -248,6 +252,24 @@ int32_t GenerateRnd(int32_t v)
...
 }
 
+void SetRndSeedV2(int32_t s)
+{
+	rng.seed(s);
+}
+
+int32_t GetRndSeedV2()
+{
+	std::uniform_int_distribution<int32_t> dist(0, INT_MAX);
```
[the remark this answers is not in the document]
qndel: dunno, I tested and the results were off by 1

qndel:
> Is this algo guaranteed to give the same sequence when setting the same seed?

obviously

qndel:
vanilla rng
```
monobit_test                             0.1016892493688818 PASS
frequency_within_block_test              0.5427869036748679 PASS
runs_test                                0.20869832039571723 PASS
longest_run_ones_in_a_block_test         0.10379179874654554 PASS
binary_matrix_rank_test                  0.10733785915603897 PASS
dft_test                                 0.16373315942288413 PASS
non_overlapping_template_matching_test   0.9999981915288942 PASS
overlapping_template_matching_test       0.2136566153526672 PASS
maurers_universal_test                   0.004284665889349451 FAIL
linear_complexity_test                   0.7170447464996982 PASS
serial_test                              0.28304219315039814 PASS
approximate_entropy_test                 0.3002676702843548 PASS
cumulative_sums_test                     0.08849918274824375 PASS
random_excursion_test                    0.029870664339639433 PASS
random_excursion_variant_test            0.08634782098366245 PASS
```

qndel:
my rng
```
monobit_test                             0.2581142844998201 PASS
frequency_within_block_test              0.6608354754963154 PASS
runs_test                                0.8426423556097362 PASS
longest_run_ones_in_a_block_test         0.4422302667466412 PASS
binary_matrix_rank_test                  0.5712732704344778 PASS
dft_test                                 0.5626456741682362 PASS
non_overlapping_template_matching_test   1.0000000299879452 PASS
overlapping_template_matching_test       0.26038389890445107 PASS
maurers_universal_test                   0.13866138941713624 PASS
linear_complexity_test                   0.2408765957644154 PASS
serial_test                              0.1545838268001897 PASS
approximate_entropy_test                 0.38857399727239755 PASS
cumulative_sums_test                     0.17274721942550308 PASS
random_excursion_test                    0.291628919843869  PASS
random_excursion_variant_test            0.13470504984007847 PASS
```

qndel:
tested with https://github.com/dj-on-github/sp800_22_tests

Source/engine.cpp
```diff
@@ -248,6 +252,24 @@ int32_t GenerateRnd(int32_t v)
...
+	return dist(rng);
+}
+
+int32_t GenerateRndV2(int32_t v)
+{
+	if (v <= 0)
+		return 0;
+	return GetRndSeedV2() % v;
```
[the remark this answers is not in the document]
qndel: it's not the goal of this PR

PR #2087, opened 2021-05-30
added path for CharisSILB.ttf on android

KonstantineLazarenko:
Hello,
I noticed that simply the pathing is hard coded for many things including the paths of preferred directories. The font which you are wanting to load to android - because it is located in the assets directory it will the path can be referenced form there . 

Another thing to mention, I tried to use #elif for preprocessor and it wasn't recognized .  So I made another #ifdef __ANDROID__ statement.

### DIFF
--- a/SourceX/DiabloUI/fonts.cpp
+++ b/SourceX/DiabloUI/fonts.cpp
@@ -68,10 +68,6 @@ void LoadTtfFont()
 	if (!FileExists(ttf_font_path.c_str())) {
 		ttf_font_path = "/usr/share/fonts/truetype/" + GetTtfName();
 	}
-#elif defined(__ANDROID__)
-	if (!FileExists(ttf_font_path.c_str())) {
-		ttf_font_path = GetTtfName();
-	}
 #endif
 	font = TTF_OpenFont(ttf_font_path.c_str(), 17);
 	if (font == NULL) {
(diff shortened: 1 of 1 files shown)


AJenbo:
`#ifdef __linux__` is probably also true on Android. Instead it should be changed to:

`#if defined(__linux__) && !defined(__ANDROID__)`

The default font path can be set in the build system via `TTF_FONT_DIR`

KonstantineLazarenko:
Ok. I this seems to be corrected.

SourceX/DiabloUI/fonts.cpp
```diff
@@ -64,10 +64,14 @@ void LoadTtfFont()
...
 	std::string ttf_font_path = GetTtfPath() + GetTtfName();
-#ifdef __linux__
+#if defined(__linux__) && !defined(__ANDROID__)
 	if (!FileExists(ttf_font_path.c_str())) {
 		ttf_font_path = "/usr/share/fonts/truetype/" + GetTtfName();
 	}
+#elif defined(__ANDROID__)
+	if (!FileExists(ttf_font_path.c_str())) {
+		ttf_font_path = GetTtfName();
+	}
```
AJenbo, line 74: Set `TTF_FONT_DIR` in CMake instead.
```suggestion
```

PR #2094, opened 2021-05-30
♻️ Leverage 'Point' struct on inventory functions

julealgon:
This PR replaces usages of individual X and Y coordinates in `inv.h`/`inv.cpp` functions with the `Point` structure.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1723,19 +1723,19 @@ void SyncGetItem(Point position, int idx, uint16_t ci, int iseed)
 	assert(FindGetItem(idx, ci, iseed) == -1);
 }
 
-bool CanPut(int x, int y)
+bool CanPut(Point position)
 {
-	if (dItem[x][y] != 0)
+	if (dItem[position.x][position.y] != 0)
 		return false;
-	if (nSolidTable[dPiece[x][y]])
+	if (nSolidTable[dPiece[position.x][position.y]])
 		return false;
 
-	if (dObject[x][y] != 0) {
-		if (object[dObject[x][y] > 0 ? dObject[x][y] - 1 : -(dObject[x][y] + 1)]._oSolidFlag)
+	if (dObject[position.x][position.y] != 0) {
+		if (object[dObject[position.x][position.y] > 0 ? dObject[position.x][position.y] - 1 : -(dObject[position.x][position.y] + 1)]._oSolidFlag)
 			return false;
 	}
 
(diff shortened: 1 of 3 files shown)


Source/inv.cpp
```diff
@@ -1228,10 +1228,10 @@ void CheckInvCut(int pnum, int mx, int my, bool automaticMove)
 		}
 
 		// check which inventory rectangle the mouse is in, if any
-		if (mx >= InvRect[r].X + xo
-		    && mx < InvRect[r].X + xo + (INV_SLOT_SIZE_PX + 1)
-		    && my >= InvRect[r].Y + yo - (INV_SLOT_SIZE_PX + 1)
-		    && my < InvRect[r].Y + yo) {
+		if (cursorPosition.x >= InvRect[r].X + xo
```
AJenbo, line 1231: On a side note, we would probably replace `InvXY` with `Point`

Source/inv.cpp
```diff
@@ -1430,9 +1430,9 @@ void inv_update_rem_item(int pnum, BYTE iv)
 void CheckInvItem(bool isShiftHeld)
 {
 	if (pcurs >= CURSOR_FIRSTITEM) {
-		CheckInvPaste(myplr, MouseX, MouseY);
+		CheckInvPaste(myplr, { MouseX, MouseY });
```
AJenbo, line 1433: Looking forwared to you converting `MouseX/Y` and `ViewX/Y` with `Point`.

Source/inv.cpp
```diff
@@ -1766,22 +1766,19 @@ bool TryInvPut()
...
 	}
 
-	position = myPlayer.position.tile + right[dir];
-	if (CanPut(position.x, position.y)) {
+	if (CanPut(myPlayer.position.tile + right[dir])) {
 		return true;
 	}
 
-	return CanPut(myPlayer.position.tile.x, myPlayer.position.tile.y);
+	return CanPut(myPlayer.position.tile);
```
AJenbo, line 1781: :tada:

AJenbo (approved):

Source/inv.cpp
julealgon: Agreed.

Source/inv.cpp
julealgon: Yep.... once the more basic replacements are done, I'll start looking into these shared globals and arrays.

PR #2096, opened 2021-05-30
♻️ Leverage 'Point' struct on items functions

julealgon:
This PR replaces usages of individual X and Y coordinates in `items.h`/`items.cpp` functions with the `Point` structure.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1397,20 +1397,19 @@ static void GetSuperItemSpace(Point position, int8_t inum)
 
 Point GetSuperItemLoc(Point position)
 {
-	Point ret;
-
 	for (int k = 1; k < 50; k++) {
 		for (int j = -k; j <= k; j++) {
-			ret.y = position.y + j;
 			for (int i = -k; i <= k; i++) {
-				ret.x = i + position.x;
-				if (ItemSpaceOk(ret)) {
-					return ret;
+				Point offset = { i, j };
+				Point positionToCheck = position + offset;
+				if (ItemSpaceOk(positionToCheck)) {
+					return positionToCheck;
 				}
 			}
 		}
 	}
-	return ret; // TODO handle no space for dropping items
+
+	return { 0, 0 }; // TODO handle no space for dropping items
 }
 
 void CalcItemValue(int i)
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -2880,10 +2880,8 @@ void SpawnRock()
 
 	int ii = AllocateItem();
 
-	int xx = object[oi].position.x;
-	int yy = object[oi].position.y;
-	items[ii].position = { xx, yy };
-	dItem[xx][items[ii].position.y] = ii + 1;
+	items[ii].position = object[oi].position;
+	dItem[object[oi].position.x][object[oi].position.y] = ii + 1;
 	int curlv = items_get_currlevel();
 	GetItemAttrs(ii, IDI_ROCK, curlv);
 	SetupItem(ii);
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1377,17 +1377,18 @@ int AllocateItem()
 
 static void GetSuperItemSpace(Point position, int8_t inum)
 {
-	if (GetItemSpace(position, inum))
+	Point positionToCheck = position;
+	if (GetItemSpace(positionToCheck, inum))
 		return;
 	for (int k = 2; k < 50; k++) {
 		for (int j = -k; j <= k; j++) {
-			int yy = position.y + j;
 			for (int i = -k; i <= k; i++) {
-				int xx = i + position.x;
-				if (!ItemSpaceOk({ xx, yy }))
+				Point offset = { i, j };
+				positionToCheck = position + offset;
+				if (!ItemSpaceOk(positionToCheck))
 					continue;
-				items[inum].position = { xx, yy };
-				dItem[xx][yy] = inum + 1;
+				items[inum].position = positionToCheck;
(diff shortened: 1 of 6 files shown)


julealgon:
@AJenbo please consider _not_ squashing this.

Source/items.cpp
```diff
@@ -420,42 +420,47 @@ void InitItemGFX()
...
+	if ((dFlags[position.x][position.y] & BFLAG_POPULATED) != 0)
 		return false;
-	if (nSolidTable[dPiece[xp][yp]])
+	if (nSolidTable[dPiece[position.x][position.y]])
 		return false;
 
 	return true;
 }
 
+Point GenerateValidRandomItemPosition()
```
AJenbo, line 441: I would also say that Generate is the wrong verb for getting an enum value, it would more be fore generating some sort of sequence. Also the negation of Valid is Invalid which I don't see as a possible option for an alternative function, so maybe Available instead. Lastly Random appears to just be an implementation detail.

Maybe `GetAvalibleItemPosition()`,  `GetAvaliblePosition()`,  `GetAvalibleGroundPosition()`?

julealgon: > Generate is the wrong verb for getting an enum value

"enum value"? Did you mean something else?

> Also the negation of Valid is Invalid which I don't see as a possible option for an alternative function, so maybe Available instead. 

Fair. Will change to `Available`.

> Lastly Random appears to just be an implementation detail.

Also fair, good point.

> `GetAvaliblePosition()`, `GetAvalibleGroundPosition()`

Those would be too general and thus misleading. The implementation is specific to items, there are other similar ones for monsters and objects (which we can combine later). I'll go with your first suggestion.

AJenbo: Actually this seams some what linked to level generation and so the random calls are crucial, so maybe `GetRandomAvalibleItemPosition`?

Source/items.cpp
```diff
@@ -420,42 +420,47 @@ void InitItemGFX()
...
 		return false;
 
 	return true;
 }
 
+Point GetAvailableItemPosition()
+{
+	Point position = {};
+	do {
+		position = Point { GenerateRnd(80), GenerateRnd(80) } + Point { 16, 16 };
```
AJenbo, line 445: Since the 16 represents the level boarder (levels are 112x112 where only the inner 80x80 tiles are playable this looks more readable to me:
```suggestion
		position = Point { 16 + GenerateRnd(80), 16 + GenerateRnd(80) };
```

Source/items.cpp
julealgon: Renamed again.

Source/items.cpp
```diff
@@ -1376,17 +1377,18 @@ int AllocateItem()
...
+	if (GetItemSpace(positionToCheck, inum))
 		return;
 	for (int k = 2; k < 50; k++) {
 		for (int j = -k; j <= k; j++) {
-			int yy = position.y + j;
 			for (int i = -k; i <= k; i++) {
-				int xx = i + position.x;
-				if (!ItemSpaceOk(xx, yy))
+				Point offset = { i, j };
+				positionToCheck = position + offset;
```
AJenbo, line 1387: In other cases you doesn't seam to do temp variables, any reason it's different here?
```suggestion
				positionToCheck = position + Point { i, j };
```

Source/items.cpp
```diff
@@ -1395,20 +1397,19 @@ static void GetSuperItemSpace(Point position, int8_t inum)
...
-
 	for (int k = 1; k < 50; k++) {
 		for (int j = -k; j <= k; j++) {
-			ret.y = position.y + j;
 			for (int i = -k; i <= k; i++) {
-				ret.x = i + position.x;
-				if (ItemSpaceOk(ret.x, ret.y)) {
-					return ret;
+				Point offset = { i, j };
+				Point positionToCheck = position + offset;
```
AJenbo, line 1404: Same as 1387
```suggestion
				Point positionToCheck = position + Point { i, j };
```

AJenbo (commented):
Really nice with all the `SpawnUnique(DIR_*)` in towners.cpp.

Source/items.cpp
julealgon: I felt it was more readable in this case to declare the offset.

Source/items.cpp
julealgon: I think this is arbitrary. Why do you think it's more readable? It's the opposite to me: the "{ 16, 16 }" represents the border as a single entity instead of being spread out as separate components in each coordinate. It even helps with extra refactoring later to isolate it as a "border" member somewhere.

Source/items.cpp
julealgon: Same reasoning.

Source/items.cpp
AJenbo: Ok.

On a side note it would be nice if we at some point generalize the code that searches 50 tiles since it's copied in a few places.

Source/items.cpp
AJenbo: The part i like is that it's first since we are offsetting top and left, not rigth and bottom.

So this would work equally well:
```suggestion
		position = Point { 16, 16 } + Point { GenerateRnd(80), GenerateRnd(80) };
```

julealgon: That sounds sensible. Will change.

Source/items.cpp
julealgon: Yeah I noticed it was duplicated while working on this... definitely needs some generalization. Perhaps a circular iterator for tiles where you specify how many tiles you want or the "radius" and it iterates over the tiles until those are reached.

Issue #2028, opened 2021-05-20
[hellfire] Misaligned lava tiles


Chance4us:
Since the beginning of the Hellfire support the lava tiles have been misaligned.
I tested down to the commit a1281d8b5459625ea47678d77d4d97cac7bb1f2d of Jan 23, 2021
and there it was also.

Comparison vanilla vs. devilutionx.
![vanilla](https://user-images.githubusercontent.com/45902449/119058268-524b4c00-b9ce-11eb-9f6a-fbd840468b56.png)
![dx](https://user-images.githubusercontent.com/45902449/119058277-58d9c380-b9ce-11eb-9b29-cfad1b0ef068.png)

Savegame:
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6519081/single_0.hsv.zip)

No big deal, just graphically in need of improvement.
By the way, they are everywhere.

julealgon:
Please consider taking screenshots using the same resolution when comparing graphics.

FitzRoyX:
Why are the placements of things (like sarcophagi) so different between the two as well? Is this the result of map generation fixes in devx?

AJenbo:
I think there is a recent bug that has objects off by one

qndel:
@Chance4us 
![image](https://user-images.githubusercontent.com/14297035/120190734-4b67d900-c219-11eb-987c-9ad42815e1a8.png)
any idea what's up with the tile marked green? Is it a flaw in the original dungeon generation?

nickgavran:
Just to shed some light on these "weird" tiles that look broken, they're actually not. They are designed so you cannot step on them. I don't know if they go unused in regular hellfire.

Here's a link to the full Crypt tileset, wherein you can see all of the tiles posted above: https://www.spriters-resource.com/pc_computer/diablodiablohellfire/sheet/63724/

qndel:
So the question is if we want to stay the same as original hellfire 🤷

AJenbo:
> So the question is if we want to stay the same as original hellfire 🤷

yes plz :D

Chance4us:
> @Chance4us
> ![image](https://user-images.githubusercontent.com/14297035/120190734-4b67d900-c219-11eb-987c-9ad42815e1a8.png)
> any idea what's up with the tile marked green? Is it a flaw in the original dungeon generation?

It looks strange, but it's the same in vanilla.

PR #2100, opened 2021-05-31
crypt fix

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2028
You have no idea what I had to do to get to this point ... 😅

### DIFF
--- a/Source/drlg_l1.cpp
+++ b/Source/drlg_l1.cpp
@@ -1950,17 +1950,21 @@ void drlg_l1_crypt_rndset(const BYTE *miniset, int rndper)
 			int kk = sw * sh + 2;
 			if (miniset[kk] >= 84 && miniset[kk] <= 100 && found) {
 				// BUGFIX: accesses to dungeon can go out of bounds (fixed)
-				// BUGFIX: Comparisons vs 100 should use same tile as comparisons vs 84 (fixed)
-				if (sx > 0 && dungeon[sx - 1][sy] >= 84 && dungeon[sx - 1][sy] <= 100) {
+				// BUGFIX: Comparisons vs 100 should use same tile as comparisons vs 84 - NOT A BUG - "fixing" this breaks crypt
+
+				auto comparison_with_bounds_check = [](Point p1, Point p2) {
+					return (p1.x >= 0 && p1.x < DMAXX && p1.y >= 0 && p1.y < DMAXY) && (p2.x >= 0 && p2.x < DMAXX && p2.y >= 0 && p2.y < DMAXY) && (dungeon[p1.x][p1.y] >= 84 && dungeon[p2.x][p2.y] <= 100);
+				};
+				if (comparison_with_bounds_check({ sx - 1, sy }, { sx - 1, sy })) {
 					found = false;
 				}
-				if (sx < DMAXX - 1 && dungeon[sx + 1][sy] >= 84 && dungeon[sx + 1][sy] <= 100) {
+				if (comparison_with_bounds_check({ sx + 1, sy }, { sx - 1, sy })) {
 					found = false;
 				}
-				if (sy < DMAXY - 1 && dungeon[sx][sy + 1] >= 84 && dungeon[sx][sy + 1] <= 100) {
+				if (comparison_with_bounds_check({ sx, sy + 1 }, { sx - 1, sy })) {
 					found = false;
 				}
-				if (sy > 0 && dungeon[sx][sy - 1] >= 84 && dungeon[sx][sy - 1] <= 100) {
+				if (comparison_with_bounds_check({ sx, sy - 1 }, { sx - 1, sy })) {
 					found = false;
 				}
 			}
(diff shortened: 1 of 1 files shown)


Chance4us:
Madness, I don't even know how you can solve that. Truly unbelievable.

Chance4us:
qndel, if I tell you a story, you won't believe me. I took the fix for a walk through the level to see how well it works. And guess what, I've come across a few of these broken tiles again.
But this time these are also included in the original.
So I would say that there is still a vanilla bug and the only question is whether we want to fix it.

vanilla1
[vanilla1.zip](https://github.com/diasurgical/devilutionX/files/6571302/vanilla1.zip)
vanilla2
[vanilla2.zip](https://github.com/diasurgical/devilutionX/files/6571306/vanilla2.zip)

Both are on the same floor.

Chance4us:
The fix works 100% and should definitely be merged even though the vanilla bug is there.

julealgon:
What bug are you talking about @Chance4us ? What is the problem with those tiles in the last screenshots?

nickgavran:
These tiles aren't bugged. They are part of the tileset, I provided a link in #2028 where you can see all Crypt tiles, and they are very clearly present there. The tiles might look bugged, but they are graphically designed that way.

Here's the link again: https://www.spriters-resource.com/pc_computer/diablodiablohellfire/sheet/63724/

Chance4us:
> What bug are you talking about @Chance4us ? What is the problem with those tiles in the last screenshots?

Yes, the tiles like in #2028.

nickgavran:
> > What bug are you talking about @Chance4us ? What is the problem with those tiles in the last screenshots?
> 
> Yes, the tiles like in #2028.
> 
> @nickgavran, Does this have to be like this?
> vanilla1

@Chance4us Yes, the tile is designed like that. Here's an excerpt from the link I posted above:
yes

Chance4us:
Then all bugs are fixed, sorry for causing confusion.

Chance4us:
@julealgon , @qndel, @AJenbo, @nickgavran So the tiles are as they should be.

julealgon:
@Chance4us and honestly... they don't even look bugged. They are just cracked tiles showing the lava beneath. If you look at the original ones, they are kinda like a lid with open parts that allow you to see the lava, but this one is just a set where some of the tiles are in poor condition and with broken parts. It makes sense that they have hard edges (as they are actual tiles) and the fact you can't step on them (due to the lava).

Chance4us:
> @Chance4us and honestly... they don't even look bugged. They are just cracked tiles showing the lava beneath. If you look at the original ones, they are kinda like a lid with open parts that allow you to see the lava, but this one is just a set where some of the tiles are in poor condition and with broken parts. It makes sense that they have hard edges (as they are actual tiles) and the fact you can't step on them (due to the lava).

In fact, if you describe it as well as you do now and I look carefully, then you are 100% right.

AJenbo:
> These tiles aren't bugged. They are part of the tileset, I provided a link in #2028 where you can see all Crypt tiles, and they are very clearly present there. The tiles might look bugged, but they are graphically designed that way.
> 
> Here's the link again: https://www.spriters-resource.com/pc_computer/diablodiablohellfire/sheet/63724/

It's the placement that is the issue, like with missing shadows etc.

AJenbo:
tbf they could have done a better job with that one :)

Source/drlg_l1.cpp
```diff
@@ -1950,17 +1950,21 @@ void drlg_l1_crypt_rndset(const BYTE *miniset, int rndper)
 			int kk = sw * sh + 2;
 			if (miniset[kk] >= 84 && miniset[kk] <= 100 && found) {
 				// BUGFIX: accesses to dungeon can go out of bounds (fixed)
-				// BUGFIX: Comparisons vs 100 should use same tile as comparisons vs 84 (fixed)
-				if (sx > 0 && dungeon[sx - 1][sy] >= 84 && dungeon[sx - 1][sy] <= 100) {
+				// BUGFIX: Comparisons vs 100 should use same tile as comparisons vs 84 - NOT A BUG - "fixing" this breaks crypt
+
+				auto comparison_with_bounds_check = [](Point p1, Point p2) {
+					return (p1.x >= 0 && p1.x < DMAXX && p1.y >= 0 && p1.y < DMAXY) && (p2.x >= 0 && p2.x < DMAXX && p2.y >= 0 && p2.y < DMAXY) && (dungeon[p1.x][p1.y] >= 84 && dungeon[p2.x][p2.y] <= 100);
+				};
```
AJenbo, line 1957: How about splitting this as a separate function instead of a lambda, doesn't seam that we really get any benefits from it being a lambda.

julealgon: The lambda is less visible than a private method, so it is preferred if no one needs access to it outside the parent function IMHO.

Source/drlg_l1.cpp
```diff
@@ -1950,17 +1950,21 @@ void drlg_l1_crypt_rndset(const BYTE *miniset, int rndper)
 			int kk = sw * sh + 2;
 			if (miniset[kk] >= 84 && miniset[kk] <= 100 && found) {
 				// BUGFIX: accesses to dungeon can go out of bounds (fixed)
-				// BUGFIX: Comparisons vs 100 should use same tile as comparisons vs 84 (fixed)
-				if (sx > 0 && dungeon[sx - 1][sy] >= 84 && dungeon[sx - 1][sy] <= 100) {
+				// BUGFIX: Comparisons vs 100 should use same tile as comparisons vs 84 - NOT A BUG - "fixing" this breaks crypt
+
+				auto comparison_with_bounds_check = [](Point p1, Point p2) {
```
glebm, line 1955:
```suggestion
				constexpr auto comparison_with_bounds_check = [](Point p1, Point p2) {
```

PR #2101, opened 2021-05-31
restore build/gitkeep and tweak automap shadow

qndel:


Chance4us:
Pretty cool expression.
I didn't know that the addition "position + Point{ 0, 1 }" is possible like this.

AJenbo:
Adding `Points` to gether is defined here: https://github.com/diasurgical/devilutionX/blob/master/Source/engine.h#L80 there are a lot of other nice things that you can do with it as well :)

@Chance4us try not to delete the whole build folder, or at least not commit `.gitkeep` next time :)

PR #2103, opened 2021-05-31
Fix 3DS CircleCI build

StephenCWills:
This adjusts the cmake commands for the CircleCI build for 3DS to match some of the other builds that didn't fail after `.gitkeep` was removed from the build folder. This makes the 3DS build a bit more robust and works around a CircleCI issue where restoring `.gitkeep` does not fix the build.


AJenbo:
thansk, sorry about .gitkeep forgot about the secound time i saw the PR

PR #2104, opened 2021-05-31
Add required libsdl2-mixer-2.0-0 to ubuntu dependencies

naggie:
I found it was necessary, at least with Ubuntu 21.04.

Thanks for the great project.


AJenbo:
Hmm I guess that is a bit of an odd situation but for 1.3.0 SDL_mixer won't be required since we switched to SDL_audiolib after the release of 1.2.1. The readme shipped with 1.2.1 should correctly reflect the need for installing SDL_mixer.

naggie:
Oh ok no problem -- thanks!

PR #2105, opened 2021-05-31
Replace ASIO with SDL_net

StephenCWills:
Here I was experimenting with replacing ASIO with SDL_net to see if I could get multiplayer working on 3DS. If you mock an empty `net/if.h` header and tweak the settings to turn off NONET and ZeroTier, you can get this to build on 3DS. To actually join a game, you also have to add some keyboard support to some of the menus. However, it crashes when attempting to join a game so there are still some things to iron out.

FYI, 3DS is the whole reason I went through the headache of adding the `3rdParty` build script and the `DEVILUTIONX_SYSTEM_SDL_NET` option. Windows and Linux can both use system dependencies for SDL_net, but 3DS doesn't have one. That's also why the `3rdParty` build script is only designed to work for *nix build environments.

I have tested the build on Windows, Linux, and 3DS. I have done minimal in-game testing on Windows with clients on localhost. This is by no means a comprehensive test, but it does seem to work for this simple case.

---

**Notes:**
* SDL_net has no API call for `setsockopt()`. I think it does have some logic for setting nodelay and reuseaddress on some platforms, but I'm not familiar with what the logic is.
* `SDLNet_TCP_Accept()` doesn't block, but it does generate a debug message when it fails. I think it would be better to use `SDLNet_CheckSockets()`, to avoid generating countless debug messages.
* SDL_net only supports IPv4 and can only bind to 0.0.0.0. This renders the `Bind Address` option in the INI obsolete so I removed it.
* ~I should have also mentioned, the CI checks are going to fail for platforms that build with TCP because there's a new dependency. The build scripts will need to be updated as well.~

### DIFF
--- a/CMake/FindSDL2_net.cmake
+++ b/CMake/FindSDL2_net.cmake
@@ -1,0 +1,197 @@
+# - Find SDL2_net
+# Find the SDL2 headers and libraries
+#
+#  SDL2::SDL2_net - Imported target
+#
+#  SDL2_net_FOUND - True if SDL2_net was found.
+#  SDL2_net_DYNAMIC - If we found a DLL version of SDL2_net
+#
+# Modified for SDL2_net of FindSDL2.cmake
+# Original Author:
+# 2015 Ryan Pavlik <ryan.pavlik@gmail.com> <abiryan@ryand.net>
+#
+# Copyright Sensics, Inc. 2015.
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE_1_0.txt or copy at
+# http://www.boost.org/LICENSE_1_0.txt)
+
+# Set up architectures (for windows) and prefixes (for mingw builds)
+if(WIN32)
+	if(MINGW)
+		include(MinGWSearchPathExtras OPTIONAL)
+		if(MINGWSEARCH_TARGET_TRIPLE)
+			set(SDL2_net_PREFIX ${MINGWSEARCH_TARGET_TRIPLE})
+		endif()
+	endif()
+	if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+		set(SDL2_net_LIB_PATH_SUFFIX lib/x64)
+		if(NOT MSVC AND NOT SDL2_net_PREFIX)
+			set(SDL2_net_PREFIX x86_64-w64-mingw32)
+		endif()
+	else()
+		set(SDL2_net_LIB_PATH_SUFFIX lib/x86)
+		if(NOT MSVC AND NOT SDL2_net_PREFIX)
+			set(SDL2_net_PREFIX i686-w64-mingw32)
+		endif()
+	endif()
+endif()
+
+if(SDL2_net_PREFIX)
+	set(SDL2_net_ORIGPREFIXPATH ${CMAKE_PREFIX_PATH})
+	if(SDL2_net_ROOT_DIR)
+		list(APPEND CMAKE_PREFIX_PATH "${SDL2_net_ROOT_DIR}")
+	endif()
+	if(CMAKE_PREFIX_PATH)
+		foreach(_prefix ${CMAKE_PREFIX_PATH})
+			list(APPEND CMAKE_PREFIX_PATH "${_prefix}/${SDL2_net_PREFIX}")
+		endforeach()
+	endif()
+	if(MINGWSEARCH_PREFIXES)
+		list(APPEND CMAKE_PREFIX_PATH ${MINGWSEARCH_PREFIXES})
+	endif()
+endif()
+
+# Invoke pkgconfig for hints
+find_package(PkgConfig QUIET)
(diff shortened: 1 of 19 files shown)


AJenbo:
did you try an disable libsodium to see if that was causing issues?

StephenCWills:
> did you try an disable libsodium to see if that was causing issues?

No, I haven't tried disabling it. I have, however, been tracing it with logs. The issue is that 3DS doesn't have `/dev/urandom` or `/dev/random`, which causes libsodium to call a `sodium_misuse()` function that crashes the application. Sodium is going to need a [custom RNG provider](https://libsodium.gitbook.io/doc/advanced/custom_rng) for 3DS, and possibly also for other platforms.

I'm currently working on one that uses the [PS service](https://libctru.devkitpro.org/ps_8h.html#a535dc015b8a14fcf5993640db964451d) and will try again once that's complete.

StephenCWills:
So the RNG provider was just problem number 1.

Problem number 2 was the 64 MB of memory allocated by Argon2. I guess it determined that was okay based on the memlimit parameter passed in at:

https://github.com/diasurgical/devilutionX/blob/2c03da35161f2b65c9e406c7f20acfaca759e571/Source/dvlnet/packet.cpp#L227-L232

We'll have to use a smaller value for platforms with memory constraints. There is a `crypto_pwhash_argon2id_MEMLIMIT_MIN` macro that indicates 8 KB is the absolute minimum required by the algorithm. I went ahead and used that for testing.

Problem number 3 is that SDL_net is returning an error with the message "Couldn't create socket". It's gotten late for me so I'll have to resume this later.

StephenCWills:
Just a little 3DS progress update. Sockets in 3DS are provided by yet another service that needs initialization. This one is called the SOC service. After initializing that, I was able to create sockets, but I'm getting some strange behavior that's still making it impossible to join any multiplayer games.

When attempting to create a game, `SDLNet_TCP_Open()` attempts to set up a listener socket with `socket()`, `setsockopt(SO_REUSEADDR)`, `bind()`, and `listen()`. Despite the fact that this is exactly the same process used by the [FTPD](https://github.com/mtheall/ftpd) service I use all the time to deploy builds, SDL_net fails on the call to `listen()` with `EBADF` (bad file number).

When attempting to join a game, SDL_net is able to connect to the remote system, but unable to receive data. The first call to `SDLNet_TCP_Recv()` gives me "error: read 0 bytes from server". I'm not sure yet if that's actually returning a zero or an error code, but it represents an error either way. The good news is, I was able to force some specific errors to occur so it is working to some extent. Joining localhost gives me an error because nothing is listening on that port. A non-existent host or IP returns a different error.

And finally, joining a host that is blocked by a firewall will cause the game to hang. It seems we'll need to do something special to make sure the socket is nonblocking.
https://github.com/mtheall/ftpd/blob/0665777a85c949c618dbe10d19091ae597957cb2/source/socket.cpp#L217-L233

CMake/FindSDL2_net.cmake
```diff
@@ -0,0 +1,197 @@
...
+		foreach(_prefix ${CMAKE_PREFIX_PATH})
+			list(APPEND CMAKE_PREFIX_PATH "${_prefix}/${SDL2_net_PREFIX}")
+		endforeach()
+	endif()
+	if(MINGWSEARCH_PREFIXES)
+		list(APPEND CMAKE_PREFIX_PATH ${MINGWSEARCH_PREFIXES})
+	endif()
+endif()
+
+# Invoke pkgconfig for hints
```
glebm, line 54: Looks like this was based on the SDL_ttf version that we have but that version is somewhat broken.
If SDL_net is found via pkg-config, it should not be used for hints but directly

```cmake
pkg_check_modules(SDL2_net QUIET SDL2_net IMPORTED_TARGET)
```

This will create a `PkgConfig::SDL2_net` imported target with the correct link flags and include directories.

Source/dvlnet/tcp_client.cpp
```diff
@@ -11,21 +12,14 @@
...
-		auto port = sgOptions.Network.nPort;
-		local_server = std::make_unique<tcp_server>(ioc, addrstr, port, passwd);
-		return join(local_server->localhost_self(), passwd);
-	} catch (std::system_error &e) {
-		SDL_SetError("%s", e.what());
-		return -1;
-	}
+	auto port = sgOptions.Network.nPort;
+	local_server = std::make_unique<tcp_server>(port, passwd);
+	return join(std::string("localhost"), passwd);
```
glebm, line 22:
```suggestion
	return join("localhost", passwd);
```

glebm:
Looks like `SDL_net` does not support IPv6 https://github.com/libsdl-org/SDL_net/issues/28

StephenCWills:
I think this code could still be useful so I'll keep the branch around, but I agree with glebm that SDL_net is really underwhelming.

AJenbo:
Yeah, it didn't sound as awesome as i was hoping :(

PR #2106, opened 2021-06-01
Point API improvements

vlolteanu:
I've also applied them non-exhaustively in the source code.

### DIFF
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -202,6 +202,13 @@ struct Point {
 
 		return offset.x + offset.y;
 	}
+
+	int WalkingDistance(Point other) const
+	{
+		Point offset = abs(*this - other);
+
+		return std::max(offset.x, offset.y);
+	}
 };
 
 struct Size {
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -67,24 +67,24 @@ struct Point {
 	int x;
 	int y;
 
-	bool operator==(const Point &other) const
+	constexpr bool operator==(const Point &other) const
 	{
 		return x == other.x && y == other.y;
 	}
 
-	bool operator!=(const Point &other) const
+	constexpr bool operator!=(const Point &other) const
 	{
 		return !(*this == other);
 	}
 
-	Point &operator+=(const Point &other)
+	constexpr Point &operator+=(const Point &other)
 	{
 		x += other.x;
 		y += other.y;
 		return *this;
 	}
 
-	Point &operator+=(Direction direction)
+	constexpr Point &operator+=(Direction direction)
 	{
 		constexpr auto toPoint = [](Direction direction) -> Point {
 			switch (direction) {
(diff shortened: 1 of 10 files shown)


vlolteanu:
There are also several instances where max(abs(x), abs(y)) is computed. I can't think of an appropriate name for that right now.

julealgon:
Can you fixup the last commit into the first? No reason to have a partial version in a commit if you already know the full implementation upfront.

Source/engine.h
```diff
@@ -197,6 +197,18 @@ struct Point {
 
 		return (approx + 512) / 1024;
 	}
+
+	Point Absolute()
```
julealgon: From what I've gathered, usually libraries override the `abs` function to do this. I've seen multiple libraries from STD library do it, for example "complex numbers" have an `abs` method.

I'd recommend following the same approach and using the same name here.

EDIT: Of course, it would also change into a static method instead of an instance one.

julealgon (changes requested):

julealgon:
> There are also several instances where max(abs(x), abs(y)) is computed. I can't think of an appropriate name for that right now.

Do you know why that is used? Perhaps we can abstract it more than that. Sounds like the intention might be "what quadrant is this in" or "which axis does it align with", which are more basic issues than just computing the "biggest coordinate value".

Perhaps even some sort of "project into cardinal direction" sort of thing could be used, depending on the purpose.

AJenbo:
For a triangle the longest leg is called the height, but that doesn't really seem like a good name.

julealgon:
Additionally, it might be worth considering whether or not we need to declare a separate `Vector2D` class that actually represents a vector and not a point. For example, "projecting a point into one axis" does not make sense, but projecting a vector does.

vlolteanu:
> Do you know why that is used? Perhaps we can abstract it more than that. Sounds like the intention might be "what quadrant is this in" or "which axis does it align with", which are more basic issues than just computing the "biggest coordinate value".

Most of that stuff is in monster.cpp and is ostensibly labeled "dist". It's as if we're computing the shortest walking distance. (AFAIK, in Diablo, walking costs the same, even if you're going diagonally.)

Maybe we should call it "NaiveDistance" or "WalkingDistance".

There are some other pieces of code where they're doing stuff like "if (abs(x1 - x2) <= k && abs(y1 - y2) <= k)", essentially checking that the naive distance is k or less.

vlolteanu:
It seems clang doesn't think std::abs is constexpr, and the C++ standard doesn't mandate it yet. Maybe a devilution::constexpr is in order?

Source/missiles.cpp
```diff
@@ -4637,35 +4635,30 @@ void MI_Rhino(int i)
...
-	}
-	if ((bx != ax || by != ay) && ((missile[i]._miVar1 & 1 && (abs(ax - cx) >= 4 || abs(ay - cy) >= 4)) || missile[i]._miVar2 > 1) && PosOkMonst(missile[i]._misource, { ax, ay })) {
-		MissToMonst(i, { ax, ay });
+	int src = missile[i]._misource;
+	Point b = missile[i].position.tile;
+	int enemy = monster[src]._menemy;
+	Point c = [&]() {
+		if ((monster[src]._mFlags & MFLAG_TARGETS_MONSTER) == 0)
+			return plr[enemy].position.tile;
+		return monster[enemy].position.tile;
```
AJenbo, line 4650:
```suggestion
	Point c = monster[src]._mFlags & MFLAG_TARGETS_MONSTER) == 0 ? plr[enemy].position.tile : monster[enemy].position.tile;
```

AJenbo (dismissed):

glebm:
> There are also several instances where max(abs(x), abs(y)) is computed. I can't think of an appropriate name for that right now.

This is L₁ norm, or Manhattan length.

PR #2115, opened 2021-06-03
Introduce Direction16 as enum

vlolteanu:
Use Point wherever possible
Make some functions static (not exhaustive)
Reduce excessive indentation (not exhaustive)
Declare variables as needed (not exhaustive)


Source/missiles.cpp
```diff
@@ -368,26 +365,24 @@ void DeleteMissile(int mi, int i)
...
-void PutMissile(int i)
+static void PutMissile(int i)
 {
-	int x, y;
-
-	x = missile[i].position.tile.x;
-	y = missile[i].position.tile.y;
+	int x = missile[i].position.tile.x;
+	int y = missile[i].position.tile.y;
 	if (x <= 0 || y <= 0 || x >= MAXDUNX || y >= MAXDUNY)
```
ephphatha, line 386: This could use `!InDungeonBounds()` as well.

Source/missiles.cpp
```diff
@@ -936,12 +930,15 @@ bool Plr2PlrMHit(int pnum, int p, int mindam, int maxdam, int dist, int mtype, b
...
 	int oi;
 	bool blocked;
 	int dir, mAnimFAmt;
 
+	int mx = position.x;
+	int my = position.y;
+
 	if (i >= MAXMISSILES || i < 0)
 		return;
 	if (mx >= MAXDUNX || mx < 0)
```
ephphatha, line 944: The checks on `mx, my` could potentially use `!InDungeonBounds()`, if it weren't for the <0 check which doesn't quite match...

ephphatha (changes requested):
This review is intended as a suggestion, mainly from what I've picked up from my limited exposure to the codebase. I'm open to comments as I'm sure there are points where my assumptions about existing conventions are incorrect.

There were a few places using C-style casts (mainly for `Direction`s) that I skipped over. My personal preference would be to use static_cast or reinterpret_cast to communicate intent but as I don't see a convention in the style guide I only suggested one change to document "odd" use of a function parameter.

I've also suggested a few places where I think an enum value makes sense instead of using magic numbers, and a couple of opportunities to remove/refactor temporary variables.

The diff for missiles.cpp is large enough that I didn't suggest refactoring any additional functions but there are low-hanging fruit such as `MoveMisslePos()` which could be refactored in a similar fashion.

Source/spells.cpp
```diff
@@ -250,7 +250,7 @@ void DoResurrect(int pnum, int rid)
 	int hp;
 
 	if ((char)rid != -1) {
-		AddMissile(plr[rid].position.tile.x, plr[rid].position.tile.y, plr[rid].position.tile.x, plr[rid].position.tile.y, 0, MIS_RESURRECTBEAM, TARGET_MONSTERS, pnum, 0, 0);
+		AddMissile(plr[rid].position.tile, plr[rid].position.tile, (Direction)0, MIS_RESURRECTBEAM, TARGET_MONSTERS, pnum, 0, 0);
```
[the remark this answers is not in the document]
vlolteanu: We could do that, but it obfuscates the fact that the direction doesn't matter in this case.

Maybe a new constant, like DIR_DEFAULT is in order?

Source/quests.cpp
```diff
@@ -230,7 +230,7 @@ void CheckQuests()
 	    && quests[Q_BETRAYER]._qvar2 == 4) {
 		rportx = 35;
 		rporty = 32;
```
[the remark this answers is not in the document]
vlolteanu: That would be beyond the scope of this PR. I consciously didn't touch quests.cpp or other files beyond the point of conforming to the new API in missiles.h.

Source/monster.cpp
```diff
@@ -3109,7 +3105,7 @@ void MAI_Bat(int i)
 		Monst->_mgoal = MGOAL_RETREAT;
 		Monst->_mgoalvar1 = 0;
 		if (Monst->MType->mtype == MT_FAMILIAR) {
-			AddMissile(Monst->enemyPosition.x, Monst->enemyPosition.y, Monst->enemyPosition.x + 1, 0, -1, MIS_LIGHTNING, TARGET_PLAYERS, i, GenerateRnd(10) + 1, 0);
+			AddMissile(Monst->enemyPosition, { Monst->enemyPosition.x + 1, 0 }, (Direction)-1, MIS_LIGHTNING, TARGET_PLAYERS, i, GenerateRnd(10) + 1, 0);
```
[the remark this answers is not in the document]
vlolteanu: Great observation. I'll take another look.

Source/missiles.cpp
```diff
@@ -4982,65 +4889,53 @@ void MI_Hbolt(int i)
...
-			CheckMissileCol(i, dam, dam, true, cx - 1, cy, true);
-		if (!CheckBlock(px, py, cx - 1, cy + 1))
-			CheckMissileCol(i, dam, dam, true, cx - 1, cy + 1, true);
-		if (!CheckBlock(px, py, cx - 1, cy - 1))
-			CheckMissileCol(i, dam, dam, true, cx - 1, cy - 1, true);
+		if (!CheckBlock(p, c))
+			CheckMissileCol(i, dam, dam, true, c, true);
+
+		constexpr Point Offsets[] = { { 0, 1 }, { 0, -1 }, { 1, 0 }, { 1, -1 }, { 1, 1 }, { -1, 0 }, { -1, 1 }, { -1, -1 } };
+		for (Point offset : Offsets) {
```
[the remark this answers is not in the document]
vlolteanu: I actually wanted to do something along the lines of "for (i = -1; i <= 1; i++) for (j = -1; j <= 1; j++)", but I was afraid of introducing side-effects due to changing the order in which everything is processed.

Source/missiles.h
```diff
@@ -92,90 +92,90 @@ void InitMissileGFX();
...
+void AddFlamec(int mi, Point src, Point dst, Direction midir, int8_t mienemy, int id, int dam);
+void AddCbolt(int mi, Point src, Point dst, Direction midir, int8_t micaster, int id, int dam);
+void AddHbolt(int mi, Point src, Point dst, Direction midir, int8_t micaster, int id, int dam);
+void AddResurrect(int mi, Point src, Point dst, Direction midir, int8_t mienemy, int id, int dam);
+void AddResurrectBeam(int mi, Point src, Point dst, Direction midir, int8_t mienemy, int id, int dam);
+void AddTelekinesis(int mi, Point src, Point dst, Direction midir, int8_t mienemy, int id, int dam);
+void AddBoneSpirit(int mi, Point src, Point dst, Direction midir, int8_t mienemy, int id, int dam);
+void AddRportal(int mi, Point src, Point dst, Direction midir, int8_t mienemy, int id, int dam);
+void AddDiabApoca(int mi, Point src, Point dst, Direction midir, int8_t mienemy, int id, int dam);
+int AddMissile(Point src, Point dst, Direction midir, int mitype, int8_t micaster, int id, int midam, int spllvl);
```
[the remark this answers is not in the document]
vlolteanu: This warrants rethinking the API. Perhaps std::variant<monostate, Direction, int> would be a cleaner replacement. (I still think it's ugly.)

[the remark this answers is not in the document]
vlolteanu: Or maybe we could just add a new int parameter aside from the Direction one.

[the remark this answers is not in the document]
AJenbo: SetMissDir() actually takes a sprite animation index so it's proper value depends on what sprite is being used.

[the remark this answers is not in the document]
AJenbo: I'm thinking that the function needs to be split up into multiple variants with a few common sub functions.
I would suggest doing it in a separate PR, so maybe revert the type for now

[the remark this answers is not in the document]
vlolteanu: I took another look and it seems that it's indexed by a "Direction16". It's not an explicit enum (yet!) that also takes south-by-southwest etc. into consideration (hence there are twice as many constants).

Source/spells.cpp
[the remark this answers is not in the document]
ephphatha: I defaulted to DIR_S as this constant is the only value carrying the same meaning for base Diablo directions and Hellfire directions. Given the type of this function now implies it takes a direction it seemed more appropriate to provide an explicit direction. Especially since `AddMissile()` does pass the value to `SetMissDir()` as if it was a direction even though it often doesn't end up getting used.

That said, like you point out we can work out which AddProc is actually getting called based on the missile type so it would be reasonable to flag it as being unused. I wouldn't want to add anything to the Direction enum in its current form though. Perhaps the outcome of https://github.com/diasurgical/devilutionX/pull/2115#discussion_r650710739 could be applied here and similar call sites?

(Sorry for the lack of context, I went from bottom to top when reviewing and forgot to update this comment.)

Source/missiles.cpp
```diff
@@ -4911,22 +4819,21 @@ void MI_Flamec(int i)
...
 void MI_Cbolt(int i)
 {
-	int md;
 	int bpath[16] = { -1, 0, 1, -1, 0, 1, -1, -1, 0, 0, 1, 1, 0, 1, -1, 0 };
 
 	missile[i]._mirange--;
 	if (missile[i]._miAnimType != MFILE_LGHNING) {
 		if (missile[i]._miVar3 == 0) {
-			md = (missile[i]._miVar2 + bpath[missile[i]._mirnd]) & 7;
+			Direction md = (Direction)((missile[i]._miVar2 + bpath[missile[i]._mirnd]) & 7);
```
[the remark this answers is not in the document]
vlolteanu: GCC is fine with this, but clang doesn't like it. I reverted this one.

Source/missiles.h
```diff
@@ -34,6 +34,38 @@ struct MissilePosition {
 	Point traveled;
 };
 
+
+/*
+ *      W  sW  SW   Sw  S
```
glebm, line 39: South is pointing up? What's going on here?

vlolteanu: Those are actually the x and y axes. If something is moving along the Y axis, it's going SW on-screen.

vlolteanu:
I've reverted midir from AddMissiles back to int. A clean solution would probably entail splitting missliedata into multiple tables based on how midir is interpreted (and probably other criteria, as well).

Source/missiles.h
ephphatha: As @AJenbo mentioned the order these values are declared matches the layout of missile sprite sheets (e.g. https://www.spriters-resource.com/pc_computer/diablodiablohellfire/sheet/97919/). With an isometric grid DIR16_S is pointing directly to the bottom of the screen and corresponds to movement along both x and y axes (+1, +1).

Source/quests.cpp
```diff
@@ -217,7 +217,7 @@ void CheckQuests()
 		quests[Q_BETRAYER].position.y = 2 * quests[Q_BETRAYER].position.y + 16;
 		rportx = quests[Q_BETRAYER].position.x;
 		rporty = quests[Q_BETRAYER].position.y;
```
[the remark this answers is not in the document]
vlolteanu: Agreed, but not as part of this PR.

Source/missiles.h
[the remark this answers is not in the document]
vlolteanu: > m thinking that the function needs to be split up into multiple variants with a few common sub functions.
> I would suggest doing it in a separate PR, so maybe revert the type for now

I did exactly that.

Source/missiles.cpp
```diff
@@ -1699,30 +1685,28 @@ void AddRuneExplosion(int mi, int sx, int sy, int dx, int dy, int midir, int8_t
...
-	missile[mi]._mlid = AddLight({ sx, sy }, 8);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { -1, -1 }, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { 0, -1 }, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { 1, -1 }, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { -1, 0 }, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { 1, 0 }, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { -1, 1 }, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { 0, 1 }, true);
+		CheckMissileCol(mi, dmg, dmg, false, missile[mi].position.tile + Point { 1, 1 }, true);
```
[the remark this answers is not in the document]
vlolteanu: Yes. Done.

Source/missiles.cpp
```diff
@@ -4911,22 +4799,21 @@ void MI_Flamec(int i)
...
 void MI_Cbolt(int i)
 {
-	int md;
 	int bpath[16] = { -1, 0, 1, -1, 0, 1, -1, -1, 0, 0, 1, 1, 0, 1, -1, 0 };
 
 	missile[i]._mirange--;
 	if (missile[i]._miAnimType != MFILE_LGHNING) {
 		if (missile[i]._miVar3 == 0) {
-			md = (missile[i]._miVar2 + bpath[missile[i]._mirnd]) & 7;
+			Direction md = (Direction)((missile[i]._miVar2 + bpath[missile[i]._mirnd]) & 7);
```
AJenbo, line 4807: Here is another way this could be written with out the need for a lot of casting and bit operations:

```suggestion
Direction md = static_cast<Direction>(missile[i]._miVar2);
switch (missile[i]._mirnd) {
	case 0:
	case 3:
	case 6:
	case 7:
	case 14:
		md = left[md];
		break;
	case 2:
	case 5:
	case 10:
	case 11:
	case 13:
		md = right[md];
		break;
}
```

vlolteanu (commented):
Done.

Source/missiles.cpp
AJenbo: I don't understand what `bpath` refers to?

Maybe a subfunction with a name like `ApplyRandomDirection()` could be better?

AJenbo: ... hadn't seen your change, yeah that is nicer :)

Source/missiles.cpp
```diff
@@ -3818,39 +3818,18 @@ void MI_FireRing(int i)
 		if (nMissileTable[dp] || missile[i]._miVar8)
 			missile[i]._miVar8 = 1;
 		else
-			AddMissile({ tx, ty }, { tx, ty }, 0, MIS_FIREWALL, TARGET_BOTH, src, dmg, missile[i]._mispllvl);
+			AddMissile({ tx, ty }, { tx, ty }, 0, type, TARGET_BOTH, src, dmg, missile[i]._mispllvl);
 	}
 }
 
-void MI_LightningRing(int i)
+void MI_FireRing(int i)
```
AJenbo, line 3826: It would have been best if you had made this change while the two functions where still identical.

vlolteanu: I'll just squash and force push.

AJenbo: acceptable

Source/missiles.cpp
```diff
@@ -4015,11 +4005,21 @@ void MI_Lightctrl(int i)
...
+			    missile[i]._micaster,
+			    missile[i]._misource,
+			    dam,
+			    missile[i]._mispllvl);
 		}
+		missile[i]._miVar1 = missile[i].position.tile.x;
+		missile[i]._miVar2 = missile[i].position.tile.y;
 	}
-	if (missile[i]._mirange == 0 || mx <= 0 || my <= 0 || mx >= MAXDUNX || my > MAXDUNY) {
+	if (missile[i]._mirange == 0 || mx == 0 || my == 0) {
```
vlolteanu, line 4022: If the checks I've deleted aren't redundant, then accessing dPiece[mx][my] is undefined behavior.

vlolteanu: In other words, is there a bug here?

qndel: technically yes but actually no (xD) as even if accessing dPiece[mx][my] is undefined behavior, the condition after it checks for the bounds so nothing scary happens but stilll ... maybe move the bounds check in front of the dpiece?

vlolteanu: There's "int pn = dPiece[mx][my]", and then some pn-related stuff right at the top of the function.

ephphatha: It looks like the x/y == 0 checks are to ensure that the missile can't move out of bounds as a workaround for the out of order/missing bounds check before indexing into dPiece you've pointed out. ProcessMissiles() checks if the missile is still in bounds (using a buggy check of `>= MAXDUNX/y - 1` where either `> MAXDUNX - 1` or `>= MAXDUNX` would've been appropriate) before calling this function, the missile position should be in the range (0,0) to (MAXDUNX-2, MAXDUNY-2) as a result. The call to GetMissilePos causes the position to be recalculated based on the start location and the distance travelled so far, but I *think* this results in the same position value.

vlolteanu: The point I was making is that dPiece is MAXDUNX x MAXDUNY in size, and mx and my are used to access one of its elements. For it not to be UB, mx and my must be in [0, MAXDUNX/Y -1].

mx and my are never mutated throughout the function, and the bounds checking is done at the very end, i.e. too late to prevent any potential bugs when accessing dPiece if the values are out of range.

I don't know if we can assume that the values supplied ([mx, my] == initial value of missile[i].position.tile) are always within bounds. Hence, the late bounds check tipped me off that there might be some bug lurking here.

ephphatha: Apologies, I didn't explain that particularly well, I was trying to say that the check at the end of that missile procedure guarantees that the missile does not move out of bounds next time it is called. However I couldn't be certain about the initial conditions. Looking into it further the flow for this type of missile is:

1. AddMissile() initialises the missile position and start location with whatever values were passed in. For MIS_LIGHTCTRL missiles this is only called from MI_Chain with either the current position of a valid missile (or the casting entity) or a position that has been checked to fall within (0, MAXDUNX), (0, MAXDUNY) - any valid dungeon tile except row 0 or col 0
2. AddLightctrl() sets the initial velocity (where x is in the range [-32, 32] and y is in the range [-16, 16])
3. ProcessMissiles() checks the missile position is "valid", meaning x is within [0, MAXDUNX-2]and y is within [0, MAXDUNY-2]
    * Missiles spawned in or which travelled to the last row/column of the dungeon are deleted
4. MI_Lightctrl() updates how far the missile has travelled using travelled += velocity and GetMissilePos()
    * due to the way the velocity is set initially this moves the missile a number of subpixels, this appears to result in movement 1 tile in the x direction and 0.5 tiles in the y direction relative to the last update
    * The missile cannot move past MAXDUNX/Y due to ProcessMissiles killing missiles in the last row/col. The asserts before indexing into dPiece should never be triggered
5. MI_Lightctrl() performs the missile logic, then flags the missile for deletion if it moved into row/col 0
    * This prevents missiles which have velocity {-32, 0} from resulting in a read out of dpiece[-1][y] in the next update

It looks like there is a potential bug if a player (or monster?) casts chain lightning to the -x direction while standing in col 0 (or in the -y direction while standing in row 0). I haven't checked the map generation code to see if it's possible for a dungeon layer to generate with an empty border tile, and I haven't checked any of the other missile types for similar bugs.

NiteKat: > I haven't checked the map generation code to see if it's possible for a dungeon layer to generate with an empty border tile
 
I have seen screenshots of a few seeds where dungeon generation lead outside of the bounds (I think into negative Y), and it causes a hall of mirrors effect on the area that is open. I can't remember if there are bad effects to missiles or not, I'm trying to locate information on the seeds as it's been a while since I've seen it. If I can find it, I'll either update this or add a new comment with it.

Source/missiles.cpp
```diff
@@ -276,6 +276,13 @@ int GetSpellLevel(int id, spell_id sn)
 	return result;
 }
 
+constexpr Direction16 Direction16Flip(Direction16 x, Direction16 pivot)
+{
+	unsigned ret = (2 * pivot + 16 - x) % 16;
+
+	return (Direction16)ret;
```
glebm, line 283: Use C++ casts instead of C casts (`static_cast<Direction16>(...)`)

AJenbo:
Everything from this PR except for the `GetDirection16` clean up has now been merged. Since Direction16 is only used as a return type of `GetDirection16` and it imply ends up as an index for the sprite frames I don't think it really adds anything value. Maybe if we could find some more use for it, but I'm not sure that is even really possible. At least not without further refactoring of how missiles are added.

AJenbo:
On second though the value it adds is that the function no longer returns magic numbers with lots of inline comments as to what they mean.

PR #2118, opened 2021-06-04
Compiling ASIO for 3DS

StephenCWills:
FINALLY, I have a version that can play Multiplayer with a Windows client, and presumably other clients.

Notes:
* Initializing the SOC service requires allocation of a 1 MB memory-aligned buffer. I'm currently doing this at startup, but it isn't necessary for Single Player games so it might be nice if this could be done as-needed.
* This PR includes a modification to the `opslimit` and `memlimit` parameters used in the libsodium algorithm for key derivation. That makes this version incompatible with prior versions so this version cannot join games hosted on older versions and vice-versa.
* Hosting games is noticeably slower than joining games on o3DS. In cross-play situations, it would probably be best if the 3DS does not host the game.

### DIFF
--- a/Source/DiabloUI/selgame.cpp
+++ b/Source/DiabloUI/selgame.cpp
@@ -182,7 +182,7 @@ void selgame_GameSelection_Select(int value)
 		vecSelGameDialog.push_back(new UiArtText(_("Enter address"), rect4, UIS_CENTER | UIS_BIG));
 
 		SDL_Rect rect5 = { (Sint16)(PANEL_LEFT + 305), (Sint16)(UI_OFFSET_Y + 314), 285, 33 };
-		vecSelGameDialog.push_back(new UiEdit(selgame_Ip, 128, rect5, UIS_MED | UIS_GOLD));
+		vecSelGameDialog.push_back(new UiEdit(_("Enter address"), selgame_Ip, 128, rect5, UIS_MED | UIS_GOLD));
 
 		SDL_Rect rect6 = { (Sint16)(PANEL_LEFT + 299), (Sint16)(UI_OFFSET_Y + 427), 140, 35 };
 		vecSelGameDialog.push_back(new UiArtTextButton(_("OK"), &UiFocusNavigationSelect, rect6, UIS_CENTER | UIS_VCENTER | UIS_BIG | UIS_GOLD));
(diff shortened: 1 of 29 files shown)


StephenCWills:
Looks like this latest version actually builds on CI. I was able to successfully create and join TCP games. However, when my Windows client tried to connect, it got stuck on "Waiting for game data", and when 3DS tried to join, it got dropped immediately like with [SDL_net](https://github.com/diasurgical/devilutionX/pull/2105). I guess the next step either way is to figure out how `multi_handle_delta()` works.

StephenCWills:
FYI, I just force pushed a small update to ASIO that eliminates the need for the `socket.h` override. I also rebased it against the CircleCI fix.

StephenCWills:
The latest version can play Multiplayer!

AJenbo:
Nice :D

StephenCWills:
The headers I added were in need of clang-format so I just pushed that up as well.

AJenbo (approved):

Source/DiabloUI/selgame.cpp
```diff
@@ -178,6 +182,10 @@ void selgame_GameSelection_Select(int value)
 	case 1: {
 		strncpy(title, _("Join TCP Games"), sizeof(title) - 1);
 
+#if defined __3DS__
+		ctr_vkbdInput(_("Enter address"), selgame_Ip, selgame_Ip);
```
AJenbo, line 186: Is it not possible to move this to `UiInitList()` where the SDL, Switch and Vita implementation lives?

AJenbo (changes requested):
Really nice, pretty impressive that you got it all connected :+1: 

Just one comment regarding hard coding multiple calls for bringing up the keyboard rather mirroring all other platforms.

Source/DiabloUI/selgame.cpp
StephenCWills: So, the hint text was somewhat necessary because the keyboard applet blocks the current thread. Without that, it's difficult to determine what it is you're supposed to be entering via the keyboard. I ended up adding hint text to the `UiEdit` class and deferring 3DS keyboard input until after render and fade-in. I went ahead and applied the hint text to the Switch and Vita function calls as well.

AJenbo (approved):

AJenbo:
Pretty cool to have network ported to a platform like this

PR #2119, opened 2021-06-04
cleanup

qndel:
Replaces 10 with INV_ROW_SLOT_SIZE whenever needed

Gets rid of
```cpp
int yy = 10 * (slotIndex / 10);
int xx = slotIndex % 10;
```
when they were used as `xx + yy` which is equal to `slotIndex` 😆 

Merge split loops trying to iterate separate inventory rows into one

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -217,10 +217,10 @@ void DrawInv(const CelOutputBuffer &out)
 
 			// calc item offsets for weapons smaller than 2x3 slots
 			if (slot == INVLOC_HAND_LEFT) {
-				screenX += frameW == INV_SLOT_SIZE_PX ? INV_SLOT_SIZE_PX : 0;
+				screenX += frameW == INV_SLOT_SIZE_PX ? INV_SLOT_HALF_SIZE_PX : 0;
 				screenY += frameH == (3 * INV_SLOT_SIZE_PX) ? 0 : -INV_SLOT_HALF_SIZE_PX;
 			} else if (slot == INVLOC_HAND_RIGHT) {
-				screenX += frameW == INV_SLOT_SIZE_PX ? (INV_SLOT_SIZE_PX - 1) : 1;
+				screenX += frameW == INV_SLOT_SIZE_PX ? (INV_SLOT_HALF_SIZE_PX - 1) : 1;
 				screenY += frameH == 3 * INV_SLOT_SIZE_PX ? 0 : -INV_SLOT_HALF_SIZE_PX;
 			}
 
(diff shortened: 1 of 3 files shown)


Source/inv.cpp
```diff
@@ -817,17 +817,19 @@ void CheckInvPaste(int pnum, Point cursorPosition)
...
+				j -= INV_SLOT_HALF_SIZE_PX;
 		}
-		if (r == SLOTXY_INV_LAST && (sy & 1) == 0)
-			j += 14;
+		if (r == SLOTXY_BELT_FIRST && (sy & 1) == 0)
+			j += INV_SLOT_HALF_SIZE_PX;
+
+		if (done) // found a slot
+			break;
 	}
```
AJenbo, line 833: `for`-loop already checks for `done`.

```suggestion
```

Source/inv.cpp
```diff
@@ -243,7 +243,7 @@ void DrawInv(const CelOutputBuffer &out)
 						light_table_index = 0;
 						cel_transparency_active = true;
 
-						const int dstX = RIGHT_PANEL_X + slotPos[INVLOC_HAND_RIGHT].x + (frameW == INV_SLOT_SIZE_PX ? 13 : -1);
+						const int dstX = RIGHT_PANEL_X + slotPos[INVLOC_HAND_RIGHT].x + (frameW == INV_SLOT_SIZE_PX ? (INV_SLOT_SIZE_PX - 1) : -1);
```
AJenbo, line 246:
```suggestion
						const int dstX = RIGHT_PANEL_X + slotPos[INVLOC_HAND_RIGHT].x + (frameW == INV_SLOT_SIZE_PX ? INV_SLOT_SIZE_PX : 0) - 1;

```

Source/inv.cpp
qndel: yes but I don't want it there, I guess I forgot to remove

Source/inv.cpp
```diff
@@ -623,62 +623,11 @@ bool AutoEquipEnabled(const PlayerStruct &player, const ItemStruct &item)
...
-		}
-
-		for (int i = 10; i < 19; i++) {
-			if (AutoPlaceItemInInventorySlot(player, i, item, persistItem))
-				return true;
-		}
-		return false;
+	for (int i = 0; i < NUM_INV_GRID_ELEM; i++) {
+		if (AutoPlaceItemInInventorySlot(player, i, item, persistItem))
+			return true;
```
julealgon, line 628: Doesn't this change the whole order where items are added? I think this was touched recently to enforce a given order.

#1876 

@Aneurin FYI

qndel: so just reverse the loop

julealgon: > so just reverse the loop

I don't think it's as simple as that (it _might_ be... but I think there is more to it).

Aneurin: Yeah this definitely looks like it would change the inventory filling algorithm. It's not just a question of whether to go forward or backwards, but also needs to consider the size of the item in order to group the right things together, plus gold and 1x1 items want to fill by columns once the bottom row is full.

qndel: > Yeah this definitely looks like it would change the inventory filling algorithm. It's not just a question of whether to go forward or backwards, but also needs to consider the size of the item in order to group the right things together, plus gold and 1x1 items want to fill by columns once the bottom row is full.

there is 0 consistency in your algorithm, it skips valid slots, 2x2 items are being placed from the right side, everything else from the left, what's going on xD

AJenbo: ree the pr and issue for further descriptions on how and why

Issue #2108, opened 2021-06-01
Magic rock anomalous alt highlightning flicker


Chance4us:
The magic rock description flickers at regular intervals.
This is probably because the item is permanently animated.

How it looks like:
[output.mkv.zip](https://github.com/diasurgical/devilutionX/files/6580137/output.mkv.zip)

Savegame:
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6580139/single_0.hsv.zip)

AJenbo:
magic rock is animated, so probably that is why this happens.

nickgavran:
I spoke with qndel about this a few weeks ago, when he was messing around with alt-highlighting. He is aware of the issue, and it does indeed happen because the magic rock is animated.

Chance4us:
If the flickering of the magic rock cannot be resolved or if the effort is too great, it can stay as it is.

julealgon:
> If the flickering of the magic rock cannot be resolved or if the effort is too great, it can stay as it is.

Keep in mind that we want to make sure game systems work consistently regardless of how many occurrences there are in vanilla. Even if the magic rock is the only affected item in the game, it is still important to fix it since we want modders to be able to introduce their own animated items if they want to do so.

This will most surely be fixed.

qndel:
I literally forgot about this :D

Chance4us:
> I literally forgot about this :D

You probably expected that I would find it anyway. I seem somehow particularly talented in finding bugs.

PR #2122, opened 2021-06-05
fix magic rock highlight

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2108

### DIFF
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -640,7 +640,7 @@ static void DrawItem(const CelOutputBuffer &out, int x, int y, int sx, int sy, b
 		CelBlitOutlineTo(out, GetOutlineColor(*pItem, false), position, *cel, nCel);
 	}
 	CelClippedDrawLightTo(out, position, *cel, nCel);
-	if (pItem->_iAnimFrame == pItem->_iAnimLen)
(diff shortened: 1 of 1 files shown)


Source/scrollrt.cpp
```diff
@@ -640,7 +640,7 @@ static void DrawItem(const CelOutputBuffer &out, int x, int y, int sx, int sy, b
 		CelBlitOutlineTo(out, GetOutlineColor(*pItem, false), position, *cel, nCel);
 	}
 	CelClippedDrawLightTo(out, position, *cel, nCel);
-	if (pItem->_iAnimFrame == pItem->_iAnimLen)
+	if (pItem->_iAnimFrame == pItem->_iAnimLen || pItem->_iCurs == ICURS_MAGIC_ROCK)
```
julealgon, line 643: This seems like a bad way to fix it IMHO. Why is the magic rock special? What if someone introduces a new animated object? Will they also have to list the object here?

qndel: yes

AJenbo: Unfortunately the magic rock not having a drop animation and instead having a idle animation special to that item (see `ProcessItems()`). Of cause refactoring it in to a property that could be used more generally and would be nice, but fixing the issue under the current para dime is imo not degrading code quality.

PR #2124, opened 2021-06-05
fix mac builds

qndel:
latest asio changes break mac builds, this is a working asio build + my fix for double windows errors

### DIFF
--- a/3rdParty/asio/CMakeLists.txt
+++ b/3rdParty/asio/CMakeLists.txt
@@ -2,8 +2,8 @@ include(FetchContent_MakeAvailableExcludeFromAll)
 
 include(FetchContent)
 FetchContent_Declare(asio
-    URL https://github.com/chriskohlhoff/asio/archive/a41c0b5299115264d2740ae0f19fb88e6124d90d.zip
-    URL_HASH MD5=16bc6ece908bb67f6a67e016fb79dbcf
+    URL https://github.com/qndel/asio/archive/c196c1b70b0503f59745d63bce4b9df914e00d53.zip
+    URL_HASH MD5=d1706a64a32abafaaafa2bc3022d1878
 )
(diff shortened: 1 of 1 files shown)


AJenbo:
kinda depressing... ps maybe we should fork it on diasurgical and not your private account :)

qndel:
> kinda depressing... ps maybe we should fork it on diasurgical and not your private account :)

up to you, atm I just wanted to get the mac build to work 😅

StephenCWills:
We may end up needing our own fork anyway because 3DS doesn't support the `pipe()` system call. Switching the `select_interrupter` implementation to the `socket_select_interrupter` is the simplest solution, but for some reason they added the following preprocessor logic to both `socket_select_interrupter.hpp` and `socket_select_interrupter.ipp` so we can't just override this one header.
https://github.com/chriskohlhoff/asio/blob/a41c0b5299115264d2740ae0f19fb88e6124d90d/asio/include/asio/detail/select_interrupter.hpp#L33-L39

@qndel Unless you want to manage all the PRs for the fork by yourself, it might be a good idea to put it on diasurgical.

qndel:
> We may end up needing our own fork anyway because 3DS doesn't support the `pipe()` system call. Switching the `select_interrupter` implementation to the `socket_select_interrupter` is the simplest solution, but for some reason they added the following preprocessor logic to both `socket_select_interrupter.hpp` and `socket_select_interrupter.ipp` so we can't just override this one header.
> https://github.com/chriskohlhoff/asio/blob/a41c0b5299115264d2740ae0f19fb88e6124d90d/asio/include/asio/detail/select_interrupter.hpp#L33-L39
> 
> @qndel Unless you want to manage all the PRs for the fork by yourself, it might be a good idea to put it on diasurgical.

good point lol!

AJenbo:
https://github.com/diasurgical/asio

Issue #2126, opened 2021-06-05
[vanilla] Thaumaturgic shrine doesn't close all chests


Chance4us:
As already known, there is a "Thaumaturgic shrine" in the game and it has the following effect.
a) Refills all chests on level with new contents.
b) Message: What once was opened now is closed

Unfortunately, it doesn't close all chests and I've actually known this issue for more than two decades and I'm probably not the only one. But we have had to fix so many worse bugs that I have withheld it until now. 

How to reproduce.
1. Load attached game
2. Touch the shrine next to you
3. Follow the mana potions
4. Notice that the chest next to the last mana potion is not closed

![screen00](https://user-images.githubusercontent.com/45902449/120900771-0556af00-c637-11eb-9fae-ddb766cca79a.png)
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6602980/single_0.hsv.zip)

qndel:
Making a potion path leading to stuff should be the new bug reporting standard <3

PR #2129, opened 2021-06-05
fix thaumaturgic

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2126
Looks like there are 2 sets of chests and thaumaturgic only handled one

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -3653,7 +3653,10 @@ bool OperateShrineThaumaturgic(int pnum)
 		assert((DWORD)v1 < MAXOBJECTS);
 		if ((object[v1]._otype == OBJ_CHEST1
 		        || object[v1]._otype == OBJ_CHEST2
-		        || object[v1]._otype == OBJ_CHEST3)
+		        || object[v1]._otype == OBJ_CHEST3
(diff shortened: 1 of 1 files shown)


AJenbo:
Arh it doesn't close trapped chests.

Chance4us:
> Arh it doesn't close trapped chests.

I checked again. It was actually a trapped chest.
Then is the question whether it is a bug.

AJenbo:
> > Arh it doesn't close trapped chests.
> 
> I checked again. It was actually a trapped chest.
> Then is the question whether it is a bug.

This PR fixes it.

Chance4us:
> This PR fixes it.

I just mean, could it be that the developers wanted it that way.

qndel:
> > This PR fixes it.
> 
> I just mean, could it be that the developers wanted it that way.

Looks more like they forgot they had a second set of chests to me 🤷

Chance4us:
So I don't mind if they are closed also. That also seems more logical if all chests are closed.

julealgon (approved):

Chance4us:
Thank you for fixing this everlasting issue.

PR #2130, opened 2021-06-06
Leverage 'Size' and 'Point' in inventory-related functions

julealgon:
This PR updates inventory-related functions to use `Size` and `Point` when possible in favor of raw values or `std::pair` instances.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -810,8 +810,7 @@ void CheckInvPaste(int pnum, Point cursorPosition)
 	SetICursor(player.HoldItem._iCurs + CURSOR_FIRSTITEM);
 	int i = cursorPosition.x + (IsHardwareCursor() ? 0 : (icursW / 2));
 	int j = cursorPosition.y + (IsHardwareCursor() ? 0 : (icursH / 2));
-	int sx = icursW28;
-	int sy = icursH28;
+	Size itemSize { icursW28, icursW28 };
 	bool done = false;
 	int r = 0;
 	for (; r < NUM_XY_SLOTS && !done; r++) {
@@ -829,12 +828,12 @@ void CheckInvPaste(int pnum, Point cursorPosition)
 			}
 		}
 		if (r == SLOTXY_CHEST_LAST) {
-			if ((sx & 1) == 0)
+			if ((itemSize.width & 1) == 0)
 				i -= 14;
-			if ((sy & 1) == 0)
+			if ((itemSize.height & 1) == 0)
 				j -= 14;
 		}
-		if (r == SLOTXY_INV_LAST && (sy & 1) == 0)
+		if (r == SLOTXY_INV_LAST && (itemSize.height & 1) == 0)
 			j += 14;
 	}
 	if (!done)
@@ -865,7 +864,7 @@ void CheckInvPaste(int pnum, Point cursorPosition)
 		done = true;
 	}
 	if (player.HoldItem._iLoc == ILOC_UNEQUIPABLE && il == ILOC_BELT) {
-		if (sx == 1 && sy == 1) {
+		if (itemSize == Size { 1, 1 }) {
 			done = true;
 			if (!AllItemsList[player.HoldItem.IDidx].iUsable)
 				done = false;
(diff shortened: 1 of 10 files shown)


Source/inv.cpp
```diff
@@ -159,12 +159,12 @@ void InitInv()
 	drawsbarflag = false;
 }
 
-static void InvDrawSlotBack(const CelOutputBuffer &out, int x, int y, int w, int h)
+static void InvDrawSlotBack(const CelOutputBuffer &out, Point position, Size size)
```
AJenbo, line 162: Do you plan to switch this to `Rect slot` at a later point and why not now?

Source/inv.cpp
```diff
@@ -207,21 +207,19 @@ void DrawInv(const CelOutputBuffer &out)
...
-			int frameH;
-			std::tie(frameW, frameH) = GetInvItemSize(frame);
+			auto frameSize = GetInvItemSize(frame);
 
 			// calc item offsets for weapons smaller than 2x3 slots
 			if (slot == INVLOC_HAND_LEFT) {
-				screenX += frameW == INV_SLOT_SIZE_PX ? 14 : 0;
-				screenY += frameH == (3 * INV_SLOT_SIZE_PX) ? 0 : -14;
+				screenX += frameSize.Width == InventorySlotSizeInPixels.Width ? 14 : 0;
+				screenY += frameSize.Height == (3 * InventorySlotSizeInPixels.Height) ? 0 : -14;
```
AJenbo, line 219: I know you didn't introduce it but `3*0 == 0` either something is wrong here or the multiplication is redundant.

Source/inv.cpp
```diff
@@ -207,21 +207,19 @@ void DrawInv(const CelOutputBuffer &out)
...
 			} else if (slot == INVLOC_HAND_RIGHT) {
-				screenX += frameW == INV_SLOT_SIZE_PX ? 13 : 1;
-				screenY += frameH == 3 * INV_SLOT_SIZE_PX ? 0 : -14;
+				screenX += frameSize.Width == InventorySlotSizeInPixels.Width ? 13 : 1;
+				screenY += frameSize.Height == 3 * InventorySlotSizeInPixels.Height ? 0 : -14;
```
AJenbo, line 222:
```suggestion
				screenY += frameSize.Height == (3 * InventorySlotSizeInPixels.Height) ? 0 : -14;

```

Source/inv.cpp
julealgon: Yep. All Point+Size instances or even Point+some constant will invariably turn into rectangles.

Source/inv.cpp
julealgon: Why was this not raising a warning in the old code?

Source/inv.cpp
julealgon: As for why, i like doing things incrementally. Check the next PR for rectangle stuff

Source/inv.cpp
AJenbo: It's probably not considered ambitious by GCC. Maybe VS would have complained if we ran it at a higher level of pedanticism.

Source/inv.cpp
AJenbo: Not sure an in between pos+size stage makes it more incremental, it just feels like more churn, but let me look a bit at #2131 and maybe that makes it more clear/digestible.

AJenbo: The many commits with rewrites of the same code makes it hard for me to see what the end result is for the various parts of the code. Could you at least add the commits needed for making Post+Size in to Rect to this PR (but not Contains or the changes to the mouse related code) so that I can see it in a combined view.

julealgon: > The many commits with rewrites of the same code makes it hard for me to see what the end result is for the various parts of the code.

There is no such thing as "end result". This same code will be touched upon again in the future with further refactoring. You can check the total result of the PR by just clicking the "Files changed" tab if you don't think the incremental view is worthwhile.

If you want to see the result of the application of this plus #2131, the latter currently contains this change in it (it will be rebased out once this is merged) so you can just look into the "Files changed" tab of #2131 instead.

> Could you at least add the commits needed for making Post+Size in to Rect to this PR (but not Contains or the changes to the mouse related code)

That doesn't make sense since we are not using `Rectangle` in here though. Rectangle logic starts being used on #2131 .

AJenbo: > There is no such thing as "end result".

Sure, was the best way I could figure to phrase it at the time. Maybe "series target" would be a better one.

> You can check the total result of the PR by just clicking the "Files changed" tab

I know. But I'm asking for a PR that has More then #2130 and less then #2131, there is no tab for that. I think we already concluded that we review things a bit differently. If you can 
add c64942b276f7b23508e4684176d7c2db53a35431, 3f918c6e9bcaf61a970e7109ae31d0c7466e744c, 89793cef1916d957a00620dc10f79f9adcf71e14 and a7a26602e126408397b390f23ac5c364a0011442 to this PR it would make reviewing this series a lot easier for me. If not then I'll have to do it manually later at a later time when time permits.

julealgon: Commits added.

julealgon:
@AJenbo is there anything else holding you back on this one?

AJenbo (dismissed):
Simply haven't had as much time on my hands lately

AJenbo (dismissed):

AJenbo (approved):

PR #2131, opened 2021-06-06
Refactor and use 'Rectangle' logic in some places

julealgon:
This PR leverages `Rectangle` by introducing a new `Contains` method for bounds checking in some places.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -963,33 +963,30 @@ void HotSpellMove(AxisDirection dir)
 		}
 	}
 
-	int x = speedspellscoords[spbslot].x;
-	int y = speedspellscoords[spbslot].y;
+	Point newMousePosition = speedspellscoords[spbslot];
 
 	if (dir.x == AxisDirectionX_LEFT) {
 		if (spbslot < speedspellcount - 1) {
-			x = speedspellscoords[spbslot + 1].x;
-			y = speedspellscoords[spbslot + 1].y;
+			newMousePosition = speedspellscoords[spbslot + 1];
 		}
 	} else if (dir.x == AxisDirectionX_RIGHT) {
 		if (spbslot > 0) {
-			x = speedspellscoords[spbslot - 1].x;
-			y = speedspellscoords[spbslot - 1].y;
+			newMousePosition = speedspellscoords[spbslot - 1];
 		}
 	}
 
 	if (dir.y == AxisDirectionY_UP) {
-		if (HSExists({ x, y - SPLICONLENGTH })) {
-			y -= SPLICONLENGTH;
+		if (HSExists(newMousePosition - Point { 0, SPLICONLENGTH })) {
+			newMousePosition.y -= SPLICONLENGTH;
 		}
 	} else if (dir.y == AxisDirectionY_DOWN) {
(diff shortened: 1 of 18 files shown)


julealgon:
Depends on #2130

Source/engine.h
```diff
@@ -226,8 +226,7 @@ struct Size {
 
 struct Rectangle {
 	Point Position;
-	int w;
-	int h;
+	Size Size;
 };
```
AJenbo: I don't think building the rectangle struct out of a `Point` and `Size` is ideal. x, y, width and height seam inherent to a rectangle. It doesn't seam to benefit any specific areas of the code either, but if it would help I think a method for getting the origin point from a rectangle might make sens instead.

julealgon: I'd disagree. I think building on top of `Point` and `Size` makes the most sense by far here. We could, optionally, expose other properties of the rectangle as we see fit, but I didn't feel the need to do so at this point.

"x, y" is a `Point`, and "width and height" is a `Size`. There is no sensible reason to downgrade the abstraction and use their properties directly in a rectangle.

This is vaguelly (not that vaguelly) based on `System.Drawing.Rectangle` from here:
https://docs.microsoft.com/en-us/dotnet/api/system.drawing.rectangle?view=net-5.0

ephphatha: I like the idea of composing a rect from those properties (instead of using multiple inheritance) but it does make it a bit verbose when trying to access specific dimensions of a Rectangle. Using .Net as an example, `x` and `y` properties exist as aliases for `Location.x` and `Location.y` respectively. Having accessors available in devilutionx::Rectangle would be nice as a shorter way to refer to these values.

Something like the following?
```cpp
constexpr int &x() { return Position.x; }
constexpr int &y() { return Position.y; }
constexpr int &height() { return Size.height; }
constexpr int &width() { return Size.width; }
```

julealgon: I thought about creating these but didn't feel an immediate need to. I'll add them later today.

AJenbo:
I think this should be ready for a rebase, would be nice to land it as other PRs are refering to it and could benefit from the rectangle logic.

julealgon:
@AJenbo rebased.

Source/engine.h
julealgon: This PR ended up being about using rectangle in a few places (instead of adding it) as it was added in a different PR. Will send a separate PR later to add these shortcuts.

ephphatha (approved):
Looks good :)

Source/controls/plrctrls.cpp
```diff
@@ -654,7 +651,7 @@ void InvMove(AxisDirection dir)
 	char itemInvId;
 	auto itemSize = GetItemSizeOnSlot(slot, itemInvId);
 
-	Point mousePos { MouseX, MouseY };
+	Point mousePos { MousePosition.x, MousePosition.y };
```
AJenbo, line 654:
```suggestion
	Point mousePos = MousePosition;
```

AJenbo (approved):

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/f621655a93b88f610c59e65991e35784fa8994a8

PR #2137, opened 2021-06-09
Item Highlight transparency

yuripourre:
This PR just tweaks the item highlight appearance a little, increasing margin/padding and turning the rectangle "transparent" when the mouse is over it.

Proposed Highlight:
highlight-pr

Original Highlight:
highlight-original

Edit: Ok, it's not really transparent but I still think it looks better than the current opaque rectangle.

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -87,34 +87,34 @@ void UnsafeDrawVerticalLine(const CelOutputBuffer &out, Point from, int height, 
 	}
 }
 
-static void DrawHalfTransparentBlendedRectTo(const CelOutputBuffer &out, int sx, int sy, int width, int height)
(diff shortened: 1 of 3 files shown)


FitzRoyX:
Hard to see a difference because, well, diablo is so dark. However, what I like about qndel's padding choice is that it keeps obstruction to a minimum when there are lots of items bunched together. I actually think there is an opportunity here for smaller fonts like press start 2p to reduce it even more, too bad we have no ttf support to experiment with.

qndel:
Did you just casually ignore `// going above 13 scatters labels of items that are next to each other` ? :D

julealgon:
Won't go into the other points raised, but the proposed padding looks a bit uneven to me. Would look better if it had similar distances in all directions or at least similar distances on left/right and on top/bottom pairs (top and bottom visibly differ in the proposal).

Issue #905, opened 2020-10-30
German Translation


Baracudus:
I could translate the whole game to native german if desired. I searched the forum for a existing translation file and found this italian "dexolutionx-it.po" file (attached) in this forum as a base to start with. Is there a "official" way to translate the whole project or is the german language done by this time?

Does the game support german characters like "ß" and "ö", "ä", "ü"?

Would like to contribute, please let me know how to do it "the right way". I there a "guide"  or something I missed? Thanks!
[devilutionx.zip](https://github.com/diasurgical/devilutionX/files/5462036/devilutionx.zip)

AJenbo:
Hi, translation support is still in the early stage. I believe I have the basics working but still need to tweak the build scripts and mark all the strings for translation. This effort can be tracked here: https://github.com/diasurgical/devilutionX/pull/902

ß, ö, ä, and ü are part of the Latin-1 Supplement block in Unicode, which we aim to support in the first version that will include translations. Possibly utilizing the font found here as a basis:
https://github.com/diasurgical/devilution/issues/32#issuecomment-402322326

The PO file should be UTF8 encoded. Since I haven't marked all the text in the application yet, the Itialien translation would be the best starting position at this point.

There won't be a guide before things are in a more coherent state. Preferably one where you can check your work ingame. For an overall status of this you can follow this issue: https://github.com/diasurgical/devilutionX/issues/66

ChaosMarc:
Keep in mind that there is a multilang (german included) PSX Version of Diablo. There are german patches floating around (https://www.gog.com/forum/general_de/sammelthread_zu_deutschmods_und_sprachpatches_fur_spiele_die_hier_nur_auf_englisch_erhaltlich_sin/post2038). With a little search, or the right tooling you might find the extracted strings or an iso to extract them yourself. this way you don't need to translate everything by hand

Baracudus:
> Keep in mind that there is a multilang (german included) PSX Version of Diablo. 

Yea, I know this Version! I will compare strings etc before posting anything here when I'm not 100% sure what the exact meaning in english was for the artist (this is special to the extended conversations in Tritram) Also, some things make simply no sense in german. Like the english "of". (Sometimes). In german, this can be translated as "von", "vom", "der", "die", "das", "welches", "welchem",...etc... 

Exaple: 
"King's sword of haste" --> König's Schwert >>**der**<< Eile
"Book of Teleport" --> "Buch >>**des**<< Teleport**s**

[...]

PR #2138, opened 2021-06-09
German translation by Baracudus

ChaosMarc:
I opened this PR so that all people who want to help with the German translation can pitch in. I've just added 3 translations of my own for now, but will add more every now and then when I find the time. For submissions please create PRs to my branch and I will merge them asap.

As all who had contact with the German language know, it's a bit complicated when it comes to genders and the endings of adjectives. This is most noticeable with item prefixes, e.g. if there was a prefix "Small" there would be more than one German translation depending on the item's gender (female: "Kleine", male: "Kleiner", neutral: "Kleines"). At the moment gendered translations are not supported so I'm proposing to use the base form of the adjectives (e.g. "Klein" which is almost always wrong) and marking the translation with "Needs work / Fuzzy" (optionally the different adjectives can be added as a comment like I did with "Bent" but I'm not sure if it's worth it). For the handling of the "of" in suffixes see [Translations/glossary.md](https://github.com/diasurgical/devilutionX/blob/master/Translations/glossary.md).

### DIFF
--- a/Translations/de.po
+++ b/Translations/de.po
@@ -2391,9 +2391,6 @@ msgstr ""
 msgid "Useless"
 msgstr ""
 
-# Verbogenes
-# Verbogener
-# Verbogene
 #: Source/itemdat.cpp:205
 #, fuzzy
 msgid "Bent"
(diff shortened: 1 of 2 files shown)


AJenbo:
Note that any translation marked as fuzzy will not appear in game.

joewis:
Translation to German while preserving the tone will be a challenge.

"Runter" sounds a bit funny to my German ears (rhymes with "kunter" and that's more associated with Pipi Longstocking).
Perhaps "Hinab in die Hölle" and "Hinab zu Diablo" sounds better, as well as "Hinauf in die Stadt" instead of "Rauf".

Also "Gillian die Barfrau" may be more fitting than Kellnerin.

ChaosMarc:
I was thinking the exact same things but those are the "official" translations used in the PSX Version. Not sure if we want to use them at all or go or own way. Some of the PSX translations are horribile. "Bent" was translated with "Pfuschen" instead of "Verbogen" 😂

AJenbo:
We don't maintin other bugs either

joewis:
> I was thinking the exact same things but those are the "official" translations used in the PSX Version. Not sure if we want to use them at all or go or own way. Some of the PSX translations are horribile. "Bent" was translated with "Pfuschen" instead of "Verbogen" 😂

Sounds like that translation was verpfuscht...

Baracudus:
Update in the Main thread!

ChaosMarc:
I have pushed the unmodified translation provided by @Baracudus from [here](https://github.com/diasurgical/devilutionX/issues/905#issuecomment-886198499) overwriting my previous changes. so we can start fresh with his nearly complete version

Issue #2140, opened 2021-06-10
Torn note reconstruction incorrect


Chance4us:
Since commit 6ce8f1375184d60afcdf2ac47fdb02610d140509 the torn note reconstruction is incorrect.

How to reproduce:
1. Load attached savegame
2. Collect the fragments of the torn note (1+2+3)
3. See the result

Normal behavior:
1. Torn note
![torn](https://user-images.githubusercontent.com/45902449/121594086-e728ee80-ca3c-11eb-8b9c-2ce0a33b6d3a.png)
2. Reconstructed note
![reconstructed](https://user-images.githubusercontent.com/45902449/121594179-00ca3600-ca3d-11eb-8ff5-16cc2d416472.png)

Unusual behavior:
Reconstructed note + Torn note 3
![unusual](https://user-images.githubusercontent.com/45902449/121594279-235c4f00-ca3d-11eb-9dd1-609545efb4db.png)

Savegame:
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6634300/single_0.hsv.zip)

The cause is unclear, just discovered and bisected.

AJenbo:
> See the result

It's not really clear what the incorrect behavior is. Does it always leave you with node 3 or does that depend on the order you pick them up?

Chance4us:
> > See the result
> 
> It's not really clear what the incorrect behavior is. Does it always leave you with node 3 or does that depend on the order you pick them up?

Good point. Just a few minutes. I will test right away.

Chance4us:
Abbreviations legend: rn = reconstructed note, 1 = rn 1, 2 = rn 2, 3 = rn 3
Results according to the collection sequence:
1,2,3 = rn + 2
1,3,2 = rn + 3
2,1,3 = default/normal
2,3,1 = rn + 3
3,1,2 = default/normal
3,2,1 = default/normal
Vanilla: Always default/normal

Chance4us:
The bug can be found in the function CheckNaKrulNotes.
If you restore the original "switch (idx) {" statement, everything works normally again.

PR #2141, opened 2021-06-12
fix reconstructed note + HasItem/RemoveInvItem combo rewritten into TryRemoveInvItemById

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2140

From what I understand, you cannot just save the indexes of 3 items like this and then remove them one by one, as removing an item might affect other indexes, so you have to perform the check again to obtain a valid index

Actually this was a good opportunity to rewrite that ugly HasItem / RemoveInvItem combo, the code looks much cleaner imo
I moved `TryRemoveInvItemById` to separate conditions to make it more visible that this check actually performs an action so it has to be last. Otherwise an if could fail but you'd remove an item!

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1478,39 +1478,24 @@ static void CheckBookLevel(PlayerStruct &player)
 static void CheckNaKrulNotes(PlayerStruct &player)
 {
 	int idx = player.HoldItem.IDidx;
+	_item_indexes notes[] = { IDI_NOTE1, IDI_NOTE2, IDI_NOTE3 };
 
 	if (idx != IDI_NOTE1 && idx != IDI_NOTE2 && idx != IDI_NOTE3) {
 		return;
 	}
 
-	int n1;
-	if (idx != IDI_NOTE1 && !player.HasItem(IDI_NOTE1, &n1)) {
-		return;
-	}
-
-	int n2;
-	if (idx != IDI_NOTE2 && !player.HasItem(IDI_NOTE2, &n2)) {
-		return;
-	}
-
-	int n3;
-	if (idx != IDI_NOTE3 && !player.HasItem(IDI_NOTE3, &n3)) {
-		return;
+	for (auto note : notes) {
+		if (idx != note && !player.HasItem(note)) {
+			return; // the player doesn't have all notes
+		}
 	}
 
 	plr[myplr].Say(HeroSpeech::JustWhatIWasLookingFor, 10);
 
-	if (idx != IDI_NOTE1) {
-		player.HasItem(IDI_NOTE1, &n1);
-		player.RemoveInvItem(n1);
-	}
-	if (idx != IDI_NOTE2) {
-		player.HasItem(IDI_NOTE2, &n2);
-		player.RemoveInvItem(n2);
-	}
-	if (idx != IDI_NOTE3) {
-		player.HasItem(IDI_NOTE3, &n3);
-		player.RemoveInvItem(n3);
+	for (auto note : notes) {
+		if (idx != note) {
+			player.TryRemoveInvItemById(note);
+		}
 	}
 
 	int itemNum = itemactive[0];
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1501,12 +1501,15 @@ static void CheckNaKrulNotes(PlayerStruct &player)
 	plr[myplr].Say(HeroSpeech::JustWhatIWasLookingFor, 10);
 
 	if (idx != IDI_NOTE1) {
+		player.HasItem(IDI_NOTE1, &n1);
 		player.RemoveInvItem(n1);
 	}
 	if (idx != IDI_NOTE2) {
+		player.HasItem(IDI_NOTE2, &n2);
 		player.RemoveInvItem(n2);
 	}
 	if (idx != IDI_NOTE3) {
+		player.HasItem(IDI_NOTE3, &n3);
 		player.RemoveInvItem(n3);
 	}
 
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -265,6 +265,16 @@ void PlayerStruct::RemoveInvItem(int iv, bool calcScrolls)
 		CalcScrolls();
 }
 
+bool PlayerStruct::TryRemoveInvItemById(int item)
+{
+	int idx;
+	if (HasItem(item, &idx)) {
+		RemoveInvItem(idx);
+		return true;
+	}
+	return false;
+}
+
 void PlayerStruct::RemoveSpdBarItem(int iv)
 {
 	SpdList[iv]._itype = ITYPE_NONE;
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -515,7 +515,7 @@ void InitPlayerGFX(int pnum)
 		player._pgfxnum = 0;
 		LoadPlrGFX(player, player_graphic::Death);
 	} else {
-		for (int i = 0; i < enum_size<player_graphic>::value; i++) {
+		for (size_t i = 0; i < enum_size<player_graphic>::value; i++) {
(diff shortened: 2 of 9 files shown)


Chance4us:
> From what I understand, you cannot just save the indexes of 3 items like this and then remove them one by one, as removing an item might affect other indexes, so you have to perform the check again to obtain a valid index

So the solution is even more elegant.

qndel:
> > From what I understand, you cannot just save the indexes of 3 items like this and then remove them one by one, as removing an item might affect other indexes, so you have to perform the check again to obtain a valid index
> 
> So the solution is even more elegant.

now it really is ;)

AJenbo (changes requested):
Looks good, but please run cleang-format, indentation are all over the place.

Source/inv.cpp
```diff
@@ -1478,39 +1478,24 @@ static void CheckBookLevel(PlayerStruct &player)
...
-	int n2;
-	if (idx != IDI_NOTE2 && !player.HasItem(IDI_NOTE2, &n2)) {
-		return;
-	}
-
-	int n3;
-	if (idx != IDI_NOTE3 && !player.HasItem(IDI_NOTE3, &n3)) {
-		return;
+	for (auto note : notes) {
+		if (idx != note && !player.HasItem(note)) { // the player doesn't have all notes
```
julealgon, line 1488: Please consider moving comments to above the line instead of having it at the side like this

Source/inv.cpp
```diff
@@ -1478,39 +1478,24 @@ static void CheckBookLevel(PlayerStruct &player)
 static void CheckNaKrulNotes(PlayerStruct &player)
 {
 	int idx = player.HoldItem.IDidx;
+	_item_indexes notes[] = { IDI_NOTE1, IDI_NOTE2, IDI_NOTE3 };
```
julealgon, line 1481: This can probably be `static`/`constexpr`

Source/player.cpp
```diff
@@ -265,6 +265,16 @@ void PlayerStruct::RemoveInvItem(int iv, bool calcScrolls)
 		CalcScrolls();
 }
 
+bool PlayerStruct::FindAndRemoveInvItem(int item)
```
julealgon, line 268: I think `int item` is very misleading. I'd expect an element named `item` to be of type `ItemStruct` instead.

Can we rename it to be more specific? `itemId` maybe?

Source/player.h
```diff
@@ -306,12 +306,18 @@ struct PlayerStruct {
...
      * @param iv invList index of item to be removed
      * @param calcScrolls If true, CalcScrolls() gets called after removing item
      */
 	void RemoveInvItem(int iv, bool calcScrolls = true);
 
+	/**
+     * @brief Remove an item from player inventory and return true if the player has the item, return false otherwise
+     * @param item IDidx of item to be removed
+     */
+
```
julealgon, line 318: Extra blank line.

Source/towners.cpp
```diff
@@ -354,14 +354,15 @@ void TalkToBarOwner(PlayerStruct &player, TownerStruct &barOwner)
...
-			if (quests[Q_LTBANNER]._qvar2 == 1 && player.HasItem(IDI_BANNER, &i)) {
-				quests[Q_LTBANNER]._qactive = QUEST_DONE;
-				quests[Q_LTBANNER]._qvar1 = 3;
-				player.RemoveInvItem(i);
-				SpawnUnique(UITEM_HARCREST, barOwner.position + DIR_SW);
-				InitQTextMsg(TEXT_BANNER3);
-				return;
+
+			if (quests[Q_LTBANNER]._qvar2 == 1) {
+				if(player.FindAndRemoveInvItem(IDI_BANNER)){
```
julealgon, line 359: Why not combine the `if`s with an `&&`? This is creating more nesting for no reason.

I see you did this in quite a few places in this change, so the diff would be drastically reduced if you just combined them.

julealgon (changes requested):
@AJenbo please don't squash this one if possible.

FitzRoyX:
The combination thing looks super weird to me. Possession is a status, removing an item is a consequence. You're if checking a consequence.

julealgon:
> Possession is a status, removing an item is a consequence. You're if checking a consequence.

Those are both true, and they are common. It is very usual for a "Remove" method to return `true` in case it was able to remove something and `false` otherwise.

I'd even say that the `FindAnd...` part of the method name is redundant here.

However, I'd be fine with a rename to `TryRemoveInvItem` as well.

AJenbo:
> @AJenbo please don't squash this one if possible.

Of cause not, the commits are sensible and do different things.

FitzRoyX:
> Those are both true, and they are common.

Well, yeah, they're both true when you handpick the places where removal is done. If it was common for a trumpet noise to playback every time an item was returned, would that warrant a new combo as well? If (player.HasItemThenRemoveItemAndPlayTrumpet())

Source/inv.cpp
qndel: I'm not an expert when it comes to static/constexpr but I don't see anything that could be evaluated at compile time since idx is dynamic 🤷

Source/player.cpp
qndel: no, I literally copied it from the function above lol

Source/towners.cpp
qndel: Because I want to emphasize this has to be the last check, otherwise you might end up with removing the item and not meeting all conditions

julealgon: Well... ands already are short circuiting so you have to be aware of the order anyways.

I think this is just a bad change: enforcing order should not cause `if`s to be broken up like this IMHO.

Source/inv.cpp
julealgon: While `idx` is, the list itself is static, so it should work just fine.

Source/player.cpp
julealgon: Screw the function above 🤣 

So much of the code needs work that we can't just blindly copy something and consider that as good IMHO. New code should follow stricter standards.

julealgon:
> Well, yeah, they're both true when you handpick the places where removal is done. If it was common for a trumpet noise to playback every time an item was returned, would that warrant a new combo as well? If (player.HasItemThenRemoveItemAndPlayTrumpet())

@FitzRoyX I think the analogy there is poor. To remove an item, you have to find it on the player, its part of the removal process. Playing trumpets is a completely unrelated operation to removing the item so ideally it should be done outside. Having said that, if we played something when removing the item, it _could_ be implemented by having the logic inside the remove method, but there would be no need to indicate that in the method's name, so even then the argument is flawed.

Your idea of separation of concerns is correct though... in purely functional programming style, you'd not conflate the two. However in object oriented code this kind of thing is quite common and acceptable.

Additionally, note that it is also fine to have `player.TryRemoveItem(itemId)` _and_ `player.HasItem(itemId)` at the same time. They are just abstractions to make callers simpler, even if they share the same internal logic for actually finding the item.

Lastly, keep in mind that forcing consumers to do `itemIndex = player.HasItem(x); player.RemoveItem(itemIndex);` is a very clear violation of the ["tell, don't ask" principle](https://martinfowler.com/bliki/TellDontAsk.html) too: you don't want to coordinate the tasks on an abstraction like that, but instead tell it what to do and let it do the logic internally.

Hope this sheds some light on why Qndel's change here is actually desirable.

julealgon:
@AJenbo 
> Of cause not, the commits are sensible and do different things.

Well... not anymore 🤣

AJenbo:
> @AJenbo
> 
> > Of cause not, the commits are sensible and do different things.
> 
> Well... not anymore rofl

Was going to write the exact same message :)

Chance4us:
Very interesting and amusing discussion.
Thanks for fixing.

PR #2143, opened 2021-06-12
attempt to fix 3ds build

qndel:


AJenbo:
nice try

qndel:
I don't know enough to attempt anything else so leaving it alone ;d

PR #2144, opened 2021-06-12
Use '_item_indexes' in the 'ItemStruct' model and make item conversion code typesafe

julealgon:
This PR is an initial change to allow us to use the `_item_indexes` enum directly in a few places instead of relying on raw 'int' values.

It replaces the raw int in the `ItemStruct` and then makes a few changes to ensure places that populate and read that value (packing and unpacking for now) are updated to handle the value as a proper enum.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1087,7 +1087,7 @@ void SetPlrHandItem(ItemStruct *h, int idata)
 	h->_iPrePower = IPL_INVALID;
 	h->_iSufPower = IPL_INVALID;
 	h->_iMagical = ITEM_QUALITY_NORMAL;
-	h->IDidx = idata;
+	h->IDidx = static_cast<_item_indexes>(idata);
 	if (gbIsHellfire)
 		h->dwBuff |= CF_HELLFIRE;
 }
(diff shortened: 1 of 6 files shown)


Source/items.cpp
```diff
@@ -1087,7 +1087,7 @@ void SetPlrHandItem(ItemStruct *h, int idata)
 	h->_iPrePower = IPL_INVALID;
 	h->_iSufPower = IPL_INVALID;
 	h->_iMagical = ITEM_QUALITY_NORMAL;
-	h->IDidx = idata;
```
AJenbo, line 1090: Looks like it's only 3 cases where `SetPlrHandItem()` isn't already being used with the enum so why not just type `idata` as `_item_indexes`?

AJenbo (approved):
I suppose you intent to fully apply it in a follow up?

Source/items.cpp
julealgon: I will rename all parameters and variables in all usages once i remake the enum.

glebm (approved):

glebm:
I think this may have introduced a bug around in save game handling:

```
../Source/loadsave.cpp: In function ‘void devilution::SaveItem(devilution::{anonymous}::SaveHelper*, devilution::ItemStruct*)’:
../Source/loadsave.cpp:1269:10: warning: comparison is always false due to limited range of data type [-Wtype-limits]
 1269 |  if (idx == -1) {
      |      ~~~~^~~~~
```

AJenbo:
Isn't this more that the enum is missing a define for `IDI_NONE = -1`, there are about 150 values missing from this enum. But the generated machine code should still work correctly.

glebm:
Ah, could be

julealgon:
> `if (idx == -1)`

@glebm 

This is probably a bit more consistent now if anything. The enum is unsigned, so -1 will now be 0xFFFF, which is actually used in other places too.

Since the enum was uint8 earlier, 2 distinct values were being used to represent the same thing. Now with uint16 they have the exact same value.

We'll be adding all missing entries to this enum in a future change and getting rid of these hardcoded values.

PR #2145, opened 2021-06-12
Fix 3DS CI build

StephenCWills:
The latest 3DS packages from devkitpro have capitalized the name of the CMake toolchain file and moved it to a `cmake` subdirectory. It also changed the CMake platform variable to `NINTENDO_3DS` instead of `N3DS`.

Furthermore, their docker image has the latest packages for the moment. The maintainer has requested we do not update packages on our CI container and instead request updates to the docker images if we need them. See: https://github.com/devkitPro/docker/issues/17

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,14 +46,12 @@ jobs:
     working_directory: ~/repo
     steps:
       - checkout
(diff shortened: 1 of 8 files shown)


AJenbo:
Thanks :)

PR #2151, opened 2021-06-13
Clean android build

StephenCWills:
* Upgrades SDL2 to 2.0.14
* Replaces `prepare.sh` with CMake `FetchContent` API
* Eliminates custom `CMakeLists.txt` files for SDL2 and SDL2_ttf
* Enables SDL_audiolib for audio support

FYI, Android Studio is no longer able to resolve JNI functions, perhaps because the SDL2 source code isn't fetched into the project's source tree. However, the project still builds and runs fine.


AJenbo:
Very nice, looks like this solves the renaming build issues and should make the Android port merge-able, only consideration would be to also have a CI integration, but with how things looks for this platform I'm thinking about doing that as a later PR.

PR #2152, opened 2021-06-13
Ensure AnimationInfo is reset when a game is loaded

### LINKED ISSUE #2149: Items often in the air when return to town
I returned to town and there was the cleaver in the air.
Unfortunately, I was unable to reproduce it.
After reloading the game, it was back on the floor. 

cleaver
I'm really sorry I don't have any useful information. I can only attach the savegame when it already looks normal.
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6644125/single_0.hsv.zip)

obligaron:
Fixes #2149

When loading a game not all members of AnimationInfo gets set explicitly.
But we should ensure they are all set to the default.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -217,6 +217,7 @@ static void LoadItemData(LoadHelper *file, ItemStruct *pItem)
 	pItem->position.y = file->nextLE<int32_t>();
 	pItem->_iAnimFlag = file->nextBool32();
 	file->skip(4); // Skip pointer _iAnimData
+	pItem->AnimInfo = {};
 	pItem->AnimInfo.NumberOfFrames = file->nextLE<int32_t>();
 	pItem->AnimInfo.CurrentFrame = file->nextLE<int32_t>();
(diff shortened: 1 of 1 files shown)


Source/loadsave.cpp
```diff
@@ -337,6 +338,7 @@ static void LoadPlayer(LoadHelper *file, int p)
 	file->skip(4); // Unused
 	player._pgfxnum = file->nextLE<int32_t>();
 	file->skip(4); // Skip pointer pData
+	player.AnimInfo = {};
```
julealgon, line 341: This explains the weird behavior @qndel mentioned not being able to repro with the character, right?

julealgon (approved):

Source/loadsave.cpp
sumyunggai: I can check if this fixes it for me as well.

sumyunggai: Fixed my issues

Chance4us:
@obligaron , @julealgon , @sumyunggai , @AJenbo , @qndel Thanks for checking/fixing.

Issue #1684, opened 2021-04-25
Replace "int x, int y" arguments with "Point point"


AJenbo:
A lot of functions take two ints as the coordinates for their operations, this is both verbose and error-prone. Instead we should use the Point structure, see an example of this type of cleanup done for the `GetDirection()`
https://github.com/diasurgical/devilutionX/commit/f77c52941afd73d4e734c0696d082218cd04c3a7

Try to split the changes up in one function per commit to making it easy to review and less prone to typos.

ephphatha:
I've been having a go at this refactoring this arvo and found a couple of spots I'm unsure about.

In objects.cpp the Open/Operate*Door functions copy the working object position to `int xp, yp` locals however there's no direct writes to those variables or to the object position. It appears that objects are positioned when a level loads (or is generated?) and don't move afterward, though given the object array is a global it's hard to be confident. Are there any side effects (e.g. in `NetSync*()`) that result in objects moving or are the local variables `xp, yp` intended as aliases? There are a few places where uses of a local are mixed with a direct reference to the object which is confusing me. For example: https://github.com/diasurgical/devilutionX/blob/6580dc6fb15ad850d96b70c1d07000250e55bf39/Source/objects.cpp#L2480

Also looking at the level generation code I was initially thinking of leaving them using x/y params as they are always used as array indexes for 2D arrays. If it's preferred to refactor those functions anyway, should the functions dealing with x/y/width/height params be updated to take a `RECT32` (or `SDL_Rect`)?

AJenbo:
I think we should use `RECT32` (though it should be renamed) and only use `SDL_Rect` when directly interacting with SDL. We can add an option for `RECT32` to be castable to  `SDL_Rect`, if that becomes useful.

`PlaySfxLoc(IS_DOORCLOS, xp, object[oi].position.y); ` appears to just be messy code that should have been cleaned up as `PlaySfxLoc(IS_DOORCLOS, xp, yp); `

It would be good if we had a better option for how to handle using indexes in to 2d arrays. Potentially we could use a getter/setter method instead of accessing the array directly, but I'm not sure that's much of an improvement.

PR #2154, opened 2021-06-14
Refactor DrawCircleMenuHint, DrawString to use devilutionX::Point

ephphatha:
This function calls through to DrawString which is effectively a wrapper around SDL drawing functions, so that function is currently taking an SDL_Rect. I've used devilutionx::Point instead of SDL_Point as the operator overloads allow calculating the specific points used for each hint position. Using the project class also let me add an explicit conversion function to go from Point to SDL_Rect in a way that suits the needs of this function. This probably could be done as a friend function for SDL_Point but I wasn't sure if adding to third party library classes is encouraged or avoided for this project.

This is part of a bunch of refactoring I've done for issue #1684, I'm splitting them into separate pull requests to minimise conflicts with other refactoring efforts such as pull requests #2115, #2130, #2131.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -324,15 +324,13 @@ static void DrawSpell(const CelOutputBuffer &out)
 
 static void PrintSBookHotkey(const CelOutputBuffer &out, Point position, const std::string &text)
 {
-	int &x = position.x;
-	int &y = position.y;
-	x -= GetLineWidth(text.c_str()) + 5;
-	x += SPLICONLENGTH;
-	y += 17;
-	y -= SPLICONLENGTH;
+	// Align the hot key text with the top-right corner of the spell icon
+	position += { SPLICONLENGTH - (GetLineWidth(text.c_str()) + 5), 17 - SPLICONLENGTH };
 
-	DrawString(out, text.c_str(), { x - 1, y + 1, 0, 0 }, UIS_BLACK);
-	DrawString(out, text.c_str(), { x + 0, y + 0, 0, 0 }, UIS_SILVER);
+	// Draw a drop shadow below and to the left of the text
+	DrawString(out, text.c_str(), position + Point { -1, 1 }, UIS_BLACK);
+	// Then draw the text over the top
+	DrawString(out, text.c_str(), position, UIS_SILVER);
 }
 
 void DrawSpellList(const CelOutputBuffer &out)
@@ -1104,7 +1102,7 @@ static void PrintInfo(const CelOutputBuffer &out)
 	if (talkflag)
 		return;
 
-	SDL_Rect line { PANEL_X + 177, PANEL_Y + LineOffsets[pnumlines][0], 288, 0 };
+	Rectangle line { PANEL_X + 177, PANEL_Y + LineOffsets[pnumlines][0], 288, 0 };
 
 	int yo = 0;
 	int lo = 1;
@@ -1115,7 +1113,7 @@ static void PrintInfo(const CelOutputBuffer &out)
 	}
 
 	for (int i = 0; i < pnumlines; i++) {
-		line.y = PANEL_Y + LineOffsets[pnumlines - lo][i + yo];
+		line.position.y = PANEL_Y + LineOffsets[pnumlines - lo][i + yo];
 		DrawString(out, panelstr[i], line, infoclr | UIS_CENTER | UIS_FIT_SPACING, 2);
 	}
 }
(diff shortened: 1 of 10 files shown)


Source/engine.h
```diff
@@ -68,6 +68,16 @@ struct Point {
 	int x;
 	int y;
 
+	/**
+	 * @brief Converts a point to an SDL_Rect with 0 width/height
```
julealgon, line 72: This is... weird to me. A rect with 0 width and height shouldn't even be a valid rect.

ephphatha: DrawString uses the SDL_Rect type as a dual use positioning and clipping rectangle, rectangles with size 0 effectively mean "don't clip, just position". This conversation was added as a hopefully temporary convenience instead of refactoring that function as part of this change and blowing out the diff.

julealgon: I'd rather we refactored that first and never had this in the first place.

ephphatha: I'll add an overload of DrawString that takes a point instead of a rect when I get home, that will remove the only use of this conversion.

Source/engine/render/text_render.cpp
```diff
@@ -275,6 +275,11 @@ void WordWrapGameString(char *text, size_t width, GameFontTables size, int spaci
 	}
 }
 
+int DrawString(const CelOutputBuffer &out, const char *text, const Point &position, uint16_t flags, int spacing, int lineHeight, bool drawTextCursor)
```
AJenbo, line 278: Now I feel like there is a ton of places that should start using this :D

ephphatha: So it turns out default initialisers and implicitly defined constructors combine to make `{1, 2}` a valid initialiser for both `Point` and `SDL_Rect`. In the case of `SDL_Rect` the values populate x and y, leaving w and h to get default-initialised (with the value 0). This means that with `DrawString(..., Point, ...)` and `DrawString(..., SDL_Rect, ...)` both available, a call such as `DrawString(..., {2, 4}, ...)` is ambiguous as it could be interpreted as `DrawString(..., SDL_Rect{2, 4/*, 0, 0*/}, ...)` or `DrawString(..., Point{2, 4}, ...)`. Using a previously defined variable of type `SDL_Rect` or `Point` (or a type with an implicit conversion to only one of those types) is unambiguous.

If only the `SDL_Rect` version of the function was available and there was an implicit conversion from `Point` to `SDL_Rect` then the behaviour would be transparent to callers, but this can be confusing. With the explicit conversion function from the previous commit callers need to do `static__cast<SDL_Rect>(Point)` which at least cuts down on the confusion a bit, but looks ugly.

Looking at the callers, DrawString is called in three "modes"; with a position only (width/height of 0), with a position and a width to force specific wrapping (height 0), or with a full bounding rect. Thankfully, only the second case is ever called with hardcoded x/y values, every case using purely a position is better represented with a `Point` variable declared earlier in the calling function.

With that in mind I'll update this pull request after refactoring the "position only" call sites to use Points, I think this will be reasonable.

ephphatha: I've updated this request with every instance where DrawString was called with only x,y params changed to use the Point version of the function.

glebm, line 278: Normally this kind of function would be defined in the header, so that the compiler can elide the call even without LTO

glebm (approved):

Source/engine/render/text_render.cpp
```diff
@@ -275,6 +275,11 @@ void WordWrapGameString(char *text, size_t width, GameFontTables size, int spaci
 	}
 }
 
+int DrawString(const CelOutputBuffer &out, const char *text, const Point &position, uint16_t flags, int spacing, int lineHeight, bool drawTextCursor)
+{
+	return DrawString(out, text, { position.x, position.y, 0, 0 }, flags, spacing, lineHeight, drawTextCursor);
```
AJenbo, line 280: I would suggest moving the logic for what to do when w/h == 0 out of DrawString<Rect> and in to DrawString<Point>.

Source/engine/render/text_render.cpp
ephphatha: Done :)

AJenbo (dismissed):

Source/engine/render/text_render.cpp
ephphatha: I've moved the handling of width 0 out of the DrawString(SDL_Rect) version. This changes the behaviour of that function when called with an SDL_Rect of width 0 so I've updated the comment to document this and replaced the one caller that was using that behaviour with an appropriately sized rectangle.

Changing the handling of height 0 will need another overload introduced, either splitting size into its own argument or having a signature for width only. It'd make sense to use the work done by @julealgon in #2130 and #2131 for this.

AJenbo:
I think something went wrong with the rebase :/

ephphatha:
With the changes to engine.h from pull request #2182 this has merge conflicts and the github web interface did a pretty awful job of resolving it. I'll have a  proper look at this later in the week.

ephphatha:
I've reordered the history to reduce the number of commits and distinct files changed in each commit. Due to moving the handling of width 0 rects into the point overload of DrawString from the end of the history to the beginning this will cause rendering issues unless the rest of the commits are included as well.

ephphatha:
Need to revise this now that the Rectangle and Size structs are available.

AJenbo:
If you can get it to build then I think it would be ready for review :)
image

ephphatha:
I ended up flattening the history for a few reasons. The change in behaviour for DrawString(Point) vs DrawString(SDL_Rect) required updating callers to use specific versions, which was relatively easy when this pull request was first opened. An initialiser list suitable for constructing an SDL_Rect was ambiguous as it could also construct a Point so I updated the call sites for Points at the time. There was one call site using the SDL_Rect version from an initialiser list which was messy but workable. Then when #2182 was merged the constants used by that call site were replaced with a Size struct. As part of the same PR some of the other call sites also were updated to have Size/Rectangle structs available.

It seemed wasteful to pull the properties out purely to convert to an SDL_Rect object so changing DrawString((SDL_Rect) to use Rectangle appeared a more suitable solution. This meant that every caller needed to be updated and since they couldn't be split squashing to a single commit seemed like the best approach.

Source/control.cpp
```diff
@@ -1700,22 +1698,39 @@ void DrawGoldSplit(const CelOutputBuffer &out, int amount)
...
 	tempstr[0] = '\0';
 	if (amount > 0) {
-		sprintf(tempstr, "%u", amount);
+		// snprintf ensures that the destination buffer ends in a null character.
+		snprintf(tempstr, bufferSize, "%u", amount);
 	}
-	DrawString(out, tempstr, { dialogX + 37, 140, 0, 0 }, UIS_SILVER, 1, -1, true);
+	// Even a ten digit amount of gold only takes up about half a line. There's no need to wrap or clip text here so we
+	// use the Point form of DrawString.
+	DrawString(out, tempstr, Point{ dialogX + 37, 140 }, UIS_SILVER, 1, -1, true);
```
AJenbo, line 1733: Looks like this needs a call from clang-format

Source/engine/render/text_render.hpp
```diff
@@ -36,7 +37,51 @@ void InitText();
...
+ * @param lineHeight Allows overriding the default line height, useful for multi-line strings.
+ * @param drawTextCursor Whether to draw an animated cursor sprite at the end of the text (default is to display nothing).
+ * @return The number of characters rendered, including characters "drawn" outside the buffer.
+*/
+int DrawString(const CelOutputBuffer &out, const char *text, const Rectangle &rect, uint16_t flags = 0, int spacing = 1, int lineHeight = -1, bool drawTextCursor = false);
+
+/**
+ * @brief Draws a line of text at the given position relative to the origin of the output buffer.
+ *
+ * This method is provided as a convenience to pass through to DrawString(..., SDL_Rect, ...) when no explicit
```
AJenbo, line 66:
```suggestion
 * This method is provided as a convenience to pass through to DrawString(..., Rectangle, ...) when no explicit

```

AJenbo:
Sorry reviewing is a bit slow atm, a bit exhausted after the last couple of days.

Source/inv.cpp
```diff
@@ -330,14 +330,9 @@ void DrawInvBelt(const CelOutputBuffer &out)
...
-			sprintf(tempstr, "%i", i + 1);
-			SDL_Rect rect {
-				InvRect[i + SLOTXY_BELT_FIRST].x + PANEL_X + InventorySlotSizeInPixels.width - GetLineWidth(tempstr),
-				InvRect[i + SLOTXY_BELT_FIRST].y + PANEL_Y - 1,
-				0,
-				0
-			};
-			DrawString(out, tempstr, rect, UIS_SILVER);
+			snprintf(tempstr, sizeof(tempstr) / sizeof(*tempstr), "%i", i + 1);
+			// FIXME: Replace this with Rectangle(Point, Size) once https://github.com/diasurgical/devilutionX/pull/2131 is merged
```
AJenbo, line 334: It's a bit unclear to me if this has already happened or not, could you point me to the specific commit where the dependant change happens?

ephphatha: That comment is out of date, I should've removed it 😅

ephphatha:
No worries if it takes a bit of time to review, I re-requested because I saw that it had been done for a few other PRs so figured I should follow convention :)

Source/inv.cpp
AJenbo: Alright, that helps motivate me to review a bit more :)

glebm (approved):

PR #2155, opened 2021-06-14
Refactor plrctrls functions to use Point and Direction types

ephphatha:
As part of this I moved the toPoint lambda to a class function on Point (and renamed it since Point::toPoint read a little off 😂 ). Ideally this would be an explicit conversion function on Direction but as that's an enum as far as I know it's not supported in c++.

I've skipped over GetItemSizeOnSlot as this is being updated in pull request #2130.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -385,7 +385,7 @@ void FindTrigger()
 	for (int i = 0; i < nummissiles; i++) {
 		int mi = missileactive[i];
 		if (missile[mi]._mitype == MIS_TOWN || missile[mi]._mitype == MIS_RPORTAL) {
-			const int newDdistance = GetDistance(missile[mi].position.tile, 2);
+			const int newDistance = GetDistance(missile[mi].position.tile, 2);
 			if (newDdistance == 0)
 				continue;
 			if (pcursmissile != -1 && distance < newDdistance)
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1086,7 +1086,7 @@ void WalkInDir(int playerId, AxisDirection dir)
 	}
 
 	const Direction pdir = FaceDir[static_cast<std::size_t>(dir.x)][static_cast<std::size_t>(dir.y)];
-	const auto delta{ future + pdir };
+	const auto delta = player.position.future + pdir;
 
 	if (CanChangeDirection(player))
 		player._pdir = pdir;
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -49,19 +49,18 @@ int slot = SLOTXY_INV_FIRST;
 
 /**
  * Number of angles to turn to face the coordinate
- * @param x Tile coordinates
- * @param y Tile coordinates
+ * @param destination Tile coordinates
  * @return -1 == down
  */
-int GetRotaryDistance(int x, int y)
+int GetRotaryDistance(Point destination)
 {
 	auto &myPlayer = plr[myplr];
 
-	if (myPlayer.position.future.x == x && myPlayer.position.future.y == y)
+	if (myPlayer.position.future == destination)
 		return -1;
 
 	int d1 = myPlayer._pdir;
-	int d2 = GetDirection(myPlayer.position.future, { x, y });
+	int d2 = GetDirection(myPlayer.position.future, destination);
 
 	int d = abs(d1 - d2);
 	if (d > 4)
@@ -119,7 +118,7 @@ void FindItemOrObject()
 	int my = plr[myplr].position.future.y;
 	int rotations = 5;
 
-	// As the player can not stand on the edge fo the map this is safe from OOB
+	// As the player can not stand on the edge of the map this is safe from OOB
 	for (int xx = -1; xx < 2; xx++) {
 		for (int yy = -1; yy < 2; yy++) {
 			if (dItem[mx + xx][my + yy] <= 0)
@@ -128,7 +127,7 @@ void FindItemOrObject()
 			if (items[i].isEmpty()
 			    || items[i]._iSelFlag == 0)
 				continue;
-			int newRotations = GetRotaryDistance(mx + xx, my + yy);
+			int newRotations = GetRotaryDistance({ mx + xx, my + yy });
 			if (rotations < newRotations)
 				continue;
 			if (xx != 0 && yy != 0 && GetDistance(mx + xx, my + yy, 1) == 0)
@@ -152,7 +151,7 @@ void FindItemOrObject()
 				continue;
 			if (xx == 0 && yy == 0 && object[o]._oDoorFlag)
 				continue; // Ignore doorway so we don't get stuck behind barrels
-			int newRotations = GetRotaryDistance(mx + xx, my + yy);
+			int newRotations = GetRotaryDistance({ mx + xx, my + yy });
 			if (rotations < newRotations)
 				continue;
 			if (xx != 0 && yy != 0 && GetDistance(mx + xx, my + yy, 1) == 0)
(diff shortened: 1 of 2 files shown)


Source/engine/point.hpp
```diff
@@ -131,6 +155,20 @@ struct Point {
...
+	 *
+	 * In practice it is likely that ApproxDistance gives the same result, especially for nearby points.
+	 * @param other Point to which we want the distance
+	 * @return Exact magnitude of vector this -> other
+	*/
+	int ExactDistance(Point other) const
+	{
+		auto vector = *this - other; //No need to call abs() as we square the values anyway
+
+		return sqrt(vector.x * vector.x + vector.y * vector.y);
```
AJenbo, line 169: `#include <algorithm>`

ephphatha: <algorithm> is already included by stdcompat/algorithm.hpp, I thought stdcompat/abs.hpp was including <cmath> but that has <cstdlib> instead.

Source/engine/point.hpp
```diff
@@ -10,6 +12,30 @@ struct Point {
 	int x;
 	int y;
 
+	static constexpr Point fromDirection(Direction direction)
```
AJenbo, line 15: Can't this then replace the the identical switch in `+=(Direction direction)` on line 58?

Source/controls/plrctrls.cpp
```diff
@@ -102,24 +100,23 @@ int GetDistance(int dx, int dy, int maxDistance)
...
-
-	return sqrt(a * a + b * b);
+	return plr[myplr].position.future.ExactDistance(destination);
 }
 
 void FindItemOrObject()
 {
+	// The loops in this function are iterating through tiles offset from the player position. This
+	// could be accomplished by looping over the values in the Direction enum and making use of
+	// Point math instead of nested loops from [-1, 1].
```
AJenbo, line 114: Doxygen:
```suggestion
/**
 * @todo The loops in this function are iterating through tiles offset from the player position. This
 * could be accomplished by looping over the values in the Direction enum and making use of
 * Point math instead of nested loops from [-1, 1].
 */
void FindItemOrObject()
{
```

Source/controls/plrctrls.cpp
```diff
@@ -389,18 +383,16 @@ void FindTrigger()
 	for (int i = 0; i < nummissiles; i++) {
 		int mi = missileactive[i];
 		if (missile[mi]._mitype == MIS_TOWN || missile[mi]._mitype == MIS_RPORTAL) {
-			int mix = missile[mi].position.tile.x;
-			int miy = missile[mi].position.tile.y;
-			const int newDdistance = GetDistance(mix, miy, 2);
+			const int newDdistance = GetDistance(missile[mi].position.tile, 2);
```
AJenbo, line 386:
```suggestion
			const int newDistance = GetDistance(missile[mi].position.tile, 2);
```

Source/controls/plrctrls.cpp
```diff
@@ -941,13 +933,13 @@ void InvMove(AxisDirection dir)
...
 {
 	for (int r = 0; r < speedspellcount; r++) {
-		if (x >= speedspellscoords[r].x - SPLICONLENGTH / 2
-		    && x < speedspellscoords[r].x + SPLICONLENGTH / 2
-		    && y >= speedspellscoords[r].y - SPLICONLENGTH / 2
-		    && y < speedspellscoords[r].y + SPLICONLENGTH / 2) {
+		if (target.x >= speedspellscoords[r].x - SPLICONLENGTH / 2
+		    && target.x < speedspellscoords[r].x + SPLICONLENGTH / 2
+		    && target.y >= speedspellscoords[r].y - SPLICONLENGTH / 2
+		    && target.y < speedspellscoords[r].y + SPLICONLENGTH / 2) {
```
AJenbo, line 942: Rectangle with a check for `Rectangle.Contains(target)` would be a nice opportunity for cleaning this :)

Source/controls/plrctrls.cpp
```diff
@@ -1068,15 +1049,13 @@ bool IsPathBlocked(int x, int y, int dir)
 		return false;
 	}
 
-	d1x = x + Offsets[d1][0];
-	d1y = y + Offsets[d1][1];
-	d2x = x + Offsets[d2][0];
-	d2y = y + Offsets[d2][1];
+	auto leftStep{ position + d1 };
+	auto rightStep{ position + d2 };
```
AJenbo, line 1053: I have a strong feeling this dien't have clang-format applied to it.

Source/controls/plrctrls.cpp
```diff
@@ -1096,26 +1075,24 @@ void WalkInDir(int playerId, AxisDirection dir)
 {
 	auto &player = plr[playerId];
 
-	const int x = player.position.future.x;
-	const int y = player.position.future.y;
+	const auto future{ player.position.future };
```
AJenbo, line 1078: Why not use `player.position.future` directly like is other parts of the code?

Source/controls/plrctrls.cpp
```diff
@@ -1096,26 +1075,24 @@ void WalkInDir(int playerId, AxisDirection dir)
...
 		if (sgbControllerActive && player.walkpath[0] != WALK_NONE && player.destAction == ACTION_NONE)
-			NetSendCmdLoc(playerId, true, CMD_WALKXY, { x, y }); // Stop walking
+			NetSendCmdLoc(playerId, true, CMD_WALKXY, future); // Stop walking
 		return;
 	}
 
 	const Direction pdir = FaceDir[static_cast<std::size_t>(dir.x)][static_cast<std::size_t>(dir.y)];
-	const int dx = x + Offsets[pdir][0];
-	const int dy = y + Offsets[pdir][1];
+	const auto delta{ future + pdir };
```
AJenbo, line 1087: Isn't the style normally:
```suggestion
	const auto delta = future + pdir;
```

Source/controls/plrctrls.cpp
```diff
@@ -1442,8 +1419,9 @@ void UpdateSpellTarget()
 	if (plr[myplr]._pRSpell == SPL_TELEPORT)
 		range = 4;
 
-	cursmx = plr[myplr].position.future.x + Offsets[plr[myplr]._pdir][0] * range;
-	cursmy = plr[myplr].position.future.y + Offsets[plr[myplr]._pdir][1] * range;
+	auto cursm{ plr[myplr].position.future + Point::fromDirection(plr[myplr]._pdir) * range };
```
AJenbo, line 1422:
```suggestion
	auto cursm = plr[myplr].position.future + Point::fromDirection(plr[myplr]._pdir) * range;
```

AJenbo (changes requested):
Good to have this cleaned up, but there seams to be a few style issues and some duplicate code that could be removed.

Source/engine/point.hpp
ephphatha: Yes, that was part of the original PR but I didn't catch all the dropped changes after the rebase. I'll update it tomorrow, getting late now where I am.

Source/controls/plrctrls.cpp
AJenbo: Oh, maybe I should have made it more clear that this is a general issues with the var name. Looks like it was actually there before your changes.

AJenbo (approved):

Source/controls/plrctrls.cpp
ephphatha: Agreed, though I think it would be more appropriate and easier to review in a separate PR. There are multiple places in this file using Chebyshev distance style checks like this which isn't appropriate given the integer coordinate system for UI elements and the inability to represent the middle of a region with even dimensions.

The checks can be rewritten to `target.WalkingDistance(speedspellscoords[r]) < SPLICONLENGTH` with a bit of refactoring in the callers which is even simpler than Rectangle.Contains, however that obscures the intent so I think the more "wasteful" approach of construction a temporary rectangle would be clearer.

Source/controls/plrctrls.cpp
```diff
@@ -107,11 +107,13 @@ int GetDistanceRanged(Point destination)
 	return plr[myplr].position.future.ExactDistance(destination);
 }
 
+/**
+ * @todo The loops in this function are iterating through tiles offset from the player position. This
```
julealgon, line 111: Instead of leaving `todo`s in the code, I'd suggest creating issues in github and documenting there instead.

TODO comments are best used for temporary work, while you are still coding the change, and should ideally not be merged.

If we add every refactoring opportunity as a TODO comment, the code will be absolutely littered with them in no time.

Issue #2089, opened 2021-05-30
Android required fixes in progress


KonstantineLazarenko:
This is a list of problem needed to be fixed before Android builds will work correctly.

- [x] SDL_GetPrefPath isn't returning the correct directory - it is returning the path to the APK directory which is owned by root.
- [x] SDL_GetBasePath is returns null on Android so should not be used in init_archives(). Instead we need a way to get the external storage directory, the path here may vary on different devices so some research is required.
- [x] The app should request access to the external storage so that we can load DIABDAT.MPQ from there.
- [x] Touch events are not currently recognized on Android. This might have been disabled when Vita support was merged.
- [x] FreeType is currently committed to the git tree in https://github.com/diasurgical/devilutionX-android/tree/master/freetype. Instead it should be build from the folder included in SDL_ttf.
- [x] Integrate SDL_audiolib into project.
- [x] prepare.sh (downloads SDL src) should be ported to CMake
- [x] SDL_audiolib needs to be told where to find the SDL src.

AJenbo:
Build error for SDL_audiolib when building after rebasing on master:

```
/home/ajenbo/code/devilutionX/CMakeLists.txt : C/C++ debug|x86 : CMake Error at /home/ajenbo/code/devilutionX/android-project/app/.cxx/cmake/debug/x86/_deps/sdl_audiolib-src/CMakeLists.txt:207 (find_package):
  By not providing "FindSDL2.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "SDL2", but
  CMake did not find one.

  Could not find a package configuration file provided by "SDL2" with any of
  the following names:

    SDL2Config.cmake
    sdl2-config.cmake

  Add the installation prefix of "SDL2" to CMAKE_PREFIX_PATH or set
  "SDL2_DIR" to a directory containing one of the above files.  If "SDL2"
  provides a separate development package or SDK, be sure it has been
  installed.

/home/ajenbo/code/devilutionX/CMakeLists.txt : C/C++ debug|x86 : Configuration failed.
executing external native build for cmake /home/ajenbo/code/devilutionX/CMakeLists.txt
/home/ajenbo/code/devilutionX/CMakeLists.txt : C/C++ debug|x86 : CMake Deprecation Warning:
  The 'cmake-server(7)' is deprecated.  Please port clients to use the
  'cmake-file-api(7)' instead.
```

PR #2156, opened 2021-06-14
Android buildsystem

AJenbo:
Resolves: #2089

We finally have a build setup for android that doesn't require manual steps :tada:


Source/init.cpp
```diff
@@ -139,20 +139,29 @@ void init_archives()
...
 	paths.emplace_back("/usr/share/diasurgical/devilutionx/");
 	paths.emplace_back("/usr/local/share/diasurgical/devilutionx/");
-#endif
-#ifdef __3DS__
+#elif defined(__3DS__)
 	paths.emplace_back("romfs:/");
 #endif
 
 	paths.emplace_back(""); // PWD
+	SDL_LogSetAllPriority(SDL_LOG_PRIORITY_VERBOSE);
```
glebm, line 164: Left-over debugging stuff?

```suggestion
```

Source/utils/display.cpp
```diff
@@ -127,7 +127,7 @@ bool SpawnWindow(const char *lpWindowName)
 	scePowerSetArmClockFrequency(444);
 #endif
 
-#if SDL_VERSION_ATLEAST(2, 0, 6)
+#if SDL_VERSION_ATLEAST(2, 0, 6) && !defined(__ANDROID__)
```
glebm, line 130: I wonder why we disable these in the first place

android-project/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
```diff
@@ -0,0 +1,650 @@
...
+        mManager = manager;
+        mDevice = device;
+        mDeviceId = mManager.getDeviceIDForIdentifier(getIdentifier());
+        mIsRegistered = false;
+        mIsChromebook = mManager.getContext().getPackageManager().hasSystemFeature("org.chromium.arc.device_management");
+        mOperations = new LinkedList<GattOperation>();
+        mHandler = new Handler(Looper.getMainLooper());
+
+        mGatt = connectGatt();
+        // final HIDDeviceBLESteamController finalThis = this;
```
glebm, line 169: Let's not commit commented code

android-project/app/src/main/java/org/libsdl/app/HIDDeviceManager.java
```diff
@@ -0,0 +1,680 @@
...
+
+                disconnectBluetoothDevice(device);
+            }
+        }
+    };
+
+    private HIDDeviceManager(final Context context) {
+        mContext = context;
+
+/*
```
glebm, line 107: Commented code

android-project/app/src/main/java/org/libsdl/app/SDLAudioManager.java
```diff
@@ -0,0 +1,390 @@
...
+                }
+                break;
+            default:
+                Log.v(TAG, "Requested " + desiredChannels + " channels, getting stereo");
+                desiredChannels = 2;
+                channelConfig = AudioFormat.CHANNEL_OUT_STEREO;
+                break;
+            }
+
+/*
```
glebm, line 134: Commented code

glebm (commented):
What's up with thousands of lines of Java here? The commented code indicates that they come from somewhere else originally

Source/utils/display.cpp
AJenbo: Vita using the touch as buttons

android-project/app/src/main/java/org/libsdl/app/HIDDeviceBLESteamController.java
AJenbo: `android-project/app/src/main/java/org/libsdl` is SDL bridge code. Everything in https://github.com/diasurgical/devilutionX/commit/b8fa575b5d99514c8a36bc0006320ef87f1c025d is taken directly from SDL 2.0.14, sorry if that wasn't clear. I would prefer to touch it as little as possible to make it easier to see what need to be updated when we update SDL lib in the future.

So maybe send this comment to https://github.com/libsdl-org/SDL :)

StephenCWills:
> What's up with thousands of lines of Java here? The commented code indicates that they come from somewhere else originally

Most of that code comes from SDL.
See: https://github.com/libsdl-org/SDL/tree/main/android-project/app/src/main/java/org/libsdl/app

The most notable exception is `DevilutionXSDLActivity.java`, in which we subclass `SDLActivity` to check for Android permissions and search for the MPQ file.

android-project/app/src/main/java/org/libsdl/app/HIDDeviceManager.java
StephenCWills: This one is actually a modification of the original code from SDL. I noticed it while updating SDL to 2.0.14 and left it in, but I'm not sure why it was commented.

AJenbo:
> What's up with thousands of lines of Java here? The commented code indicates that they come from somewhere else originally

Android can only run Java/Kotlin apps, so for a C++ app you have a Java wrapper that handles all IO between the two. This all the Java code here is taken directly from the SDL sample app: https://github.com/libsdl-org/SDL/tree/main/android-project

Unfortunately I don't think there is a better way to handle this then to commit it to the repo.

Source/utils/display.cpp
glebm: Perhaps `&& defined(__vita__)` is better than `&& !defined(__ANDROID)`?

glebm:
We already `add_subdirectory` with SDL2 code, shouldn't we use the files directly from there instead of copying them here?

Source/init.cpp
```diff
@@ -139,20 +139,29 @@ void init_archives()
 {
 	std::vector<std::string> paths;
 	paths.reserve(5);
+#ifdef __ANDROID__
+	paths.push_back(std::string(getenv("EXTERNAL_STORAGE")) + "/");
```
StephenCWills, line 143: Would it perhaps be better to place the data files in a `devilutionx` subfolder or something?

AJenbo: > We already `add_subdirectory` with SDL2 code, shouldn't we use the files directly from there instead of copying them here?

Probably, but my Java/Android skills aren't that strong.

AJenbo:
> We already `add_subdirectory` with SDL2 code, shouldn't we use the files directly from there instead of copying them here?

My Java/Android skills aren't that strong.

android-project/app/src/main/java/org/libsdl/app/HIDDeviceManager.java
AJenbo: I think @kkszysiu was either having issues with the APP crashing, or simply wanted to avoid depending on `hidapi` in order to make the app more lean.

https://github.com/libsdl-org/SDL/blob/98a966d1c20afc0abd27d8cd449810893104a908/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L167

vs

https://github.com/diasurgical/devilutionX/pull/2156/commits/93bbb3422bf95183c9ee9cd0c2e6081e6da84df5#diff-73b87fd6134eecce53d1d1af5c25cfa6d3bfbc6d87c50e2cf238ec17c27605bfR7

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -0,0 +1,136 @@
...
+				.setMessage(message)
+				.show();
+
+		if (dataFileFound) {
+			triggerPulse(); // Start full game as soon as we have file access
+		}
+
+		try {
+			Looper.loop();
+		} catch (RuntimeException ignored) {
```
StephenCWills, line 89: I actually hate this. I'm sure there is a better way to exit the looper.

StephenCWills:
> We already `add_subdirectory` with SDL2 code, shouldn't we use the files directly from there instead of copying them here?

I don't believe that it's an option. At the very least, the build.gradle files come from SDL, and those represent the entry point for the build. Also, I believe they are meant to be modified, and they are used to indicate where CMakeLists.txt is located.

https://github.com/diasurgical/devilutionX/blob/45f6ec4345c18f506cba952af9775d7895ae3187/android-project/app/build.gradle#L38-L43

If CMake brings in the SDL code, even if it copies the necessary files from SDL to the appropriate locations, we may have a complex chicken-and-egg problem on our hands.

You can find more info from SDL documentation. Their instructions do involve copying the `android-project` folder from their repo.
https://github.com/libsdl-org/SDL/blob/main/docs/README-android.md

Source/init.cpp
StephenCWills: So you're saying we should defer changing it until we can settle on a well-defined pattern?

glebm:
Fair enough. I guess we'll just need to "rebase" these files when updating SDL versions.

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
AJenbo: Yeah I hate it as well, but wasn't able to find anything better :(

The thing it's trying to do is pause the application until the user closes the dialog. The proper solution is to make custom gui/configure stage and not use the alert dialog, but I also feel strongly that something like that should get out of the way asap and preferably never show if the user correctly added the MPQ files before even launching the app the first time.

AJenbo:
OpenMW appears to have gone the same route: https://github.com/xyzz/openmw-android/tree/master/app/src/main/java/org/libsdl/app

StephenCWills (dismissed):
I've reviewed all the changes and just ran a successful build against this branch so I think it's ready.

Source/init.cpp
AJenbo: That's my current thinking. But I will give it a quick look before merging the PR.

StephenCWills:
> OpenMW appears to have gone the same route: https://github.com/xyzz/openmw-android/tree/master/app/src/main/java/org/libsdl/app

One more thing that's worth noting along the same lines. OpenMW also modifies SDL's Java code. For example, here are some additional JNI calls that were added to implement the on-screen controls.

https://github.com/xyzz/openmw-android/blob/62694d7ce3db97e558455e9db0a4b4b3d6518d2c/app/src/main/java/org/libsdl/app/SDLActivity.java#L801-L805

AJenbo:
I always knew Java was pretty heavy on boiler plate but 6k lines is ridiculous :D

StephenCWills:
I believe that boilerplate heaviness one of the major arguments in favor of Kotlin. But, to be fair, it looks like a lot of that code is for SDL features that need platform support like audio and controllers.

StephenCWills (approved):

Source/init.cpp
AJenbo: Looks like the app adds `/devilutionx/` to the path, I think that is reasonable so I have added that for now. If there are any issues with this we should still have time to adjust :)

PR #2159, opened 2021-06-14
Refactor sound related functions to use Point

ephphatha:
This is used in quite a few places in objects.cpp and had some very confusing invocations in the various Operate*Door functions due to the inconsistent use of temporary/alias variables. I've included some refactoring of these functions to try and make it clearer while I was in the area.

This may conflict with pull request #2115, would be worth approving that request before this one so I can fix any conflicts I cause.

I plan to open a separate pull request for ObjSetMicro and other related functions in objects.cpp, I've picked these changes out to try keep the diffs manageable.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -2444,335 +2444,337 @@ void RedoPlayerVision()
  * @param doorPos Map tile where the door is in its closed position
  * @return true if the door is free to be closed, false if anything is blocking it
 */
-static inline bool isDoorClear(const Point &doorPos)
+static inline bool isDoorClear(const Point &doorPosition)
 {
-	return dDead[doorPos.x][doorPos.y] == 0 && dMonster[doorPos.x][doorPos.y] == 0 && dItem[doorPos.x][doorPos.y] == 0;
+	return dDead[doorPosition.x][doorPosition.y] == 0
+		&& dMonster[doorPosition.x][doorPosition.y] == 0
+		&& dItem[doorPosition.x][doorPosition.y] == 0;
 }
 
 void OperateL1RDoor(int pnum, int oi, bool sendflag)
 {
-	const Point &objPos = object[oi].position;
+	ObjectStruct &door = object[oi];
 
-	if (object[oi]._oVar4 == 2) {
+	if (door._oVar4 == 2) {
 		if (!deltaload)
-			PlaySfxLoc(IS_DOORCLOS, objPos);
+			PlaySfxLoc(IS_DOORCLOS, door.position);
 		return;
 	}
 
-	if (object[oi]._oVar4 == 0) {
+	if (door._oVar4 == 0) {
 		if (pnum == myplr && sendflag)
 			NetSendCmdParam1(true, CMD_OPENDOOR, oi);
 		if (currlevel < 21) {
 			if (!deltaload)
-				PlaySfxLoc(IS_DOOROPEN, objPos);
-			ObjSetMicro(objPos.x, objPos.y, 395);
+				PlaySfxLoc(IS_DOOROPEN, door.position);
+			ObjSetMicro(door.position.x, door.position.y, 395);
 		} else {
 			if (!deltaload)
-				PlaySfxLoc(IS_CROPEN, objPos);
-			ObjSetMicro(objPos.x, objPos.y, 209);
+				PlaySfxLoc(IS_CROPEN, door.position);
+			ObjSetMicro(door.position.x, door.position.y, 209);
 		}
 		if (currlevel < 17) {
-			dSpecial[objPos.x][objPos.y] = 8;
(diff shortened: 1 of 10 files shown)


Source/objects.cpp
```diff
@@ -2603,23 +2612,21 @@ void OperateL1LDoor(int pnum, int oi, bool sendflag)
 
 void OperateL2RDoor(int pnum, int oi, bool sendflag)
 {
-	int xp, yp;
-	bool dok;
+	const Point &objPos = object[oi].position;
```
AJenbo, line 2615: We are trying to avoid using abbreviations, so either `objectPosition` or `objPosition` would be preferred.

ephphatha: Done :). as `objectPosition` isn't much shorter than `object[oi].position` I used a reference to the current object and called it `door` instead to hopefully make it more clear what we're operating on. This also let me make these functions consistently use the reference instead of mixing indexes into the objects array and the position property reference.

AJenbo:
Wow that's bad timing, it's going to need another rebase :)

Source/objects.cpp
```diff
@@ -2437,352 +2437,345 @@ void RedoPlayerVision()
...
+{
+	return dDead[doorPosition.x][doorPosition.y] == 0
+		&& dMonster[doorPosition.x][doorPosition.y] == 0
+		&& dItem[doorPosition.x][doorPosition.y] == 0;
+}
+
 void OperateL1RDoor(int pnum, int oi, bool sendflag)
 {
-	int xp, yp;
+	ObjectStruct &door = object[oi]; // Non const ref as flags are updated in this function
```
AJenbo, line 2457: Idk, this comment seams pretty obvious, how would you open a door by only inspecting it?
```suggestion
	ObjectStruct &door = object[oi];
```

Source/objects.cpp
```diff
@@ -2436,352 +2436,345 @@ void RedoPlayerVision()
...
 	} else {
-		object[oi]._oVar4 = 2;
+		door._oVar4 = 2;
 	}
 }
 
 void OperateL1LDoor(int pnum, int oi, bool sendflag)
 {
-	int xp, yp;
+	ObjectStruct &door = object[oi]; // Non const ref as flags are updated in this function
```
AJenbo, line 2534:
```suggestion
	ObjectStruct &door = object[oi];
```

Source/objects.cpp
```diff
@@ -2436,352 +2436,345 @@ void RedoPlayerVision()
...
-		object[oi]._oVar4 = 2;
+		door._oVar4 = 2;
 	}
 }
 
 void OperateL2RDoor(int pnum, int oi, bool sendflag)
 {
-	int xp, yp;
-	bool dok;
+	ObjectStruct &door = object[oi]; // Non const ref as flags are updated in this function
```
AJenbo, line 2615:
```suggestion
	ObjectStruct &door = object[oi];
```

Source/objects.cpp
```diff
@@ -2436,352 +2436,345 @@ void RedoPlayerVision()
...
-		object[oi]._oVar4 = 2;
+		door._oVar4 = 2;
 	}
 }
 
 void OperateL2LDoor(int pnum, int oi, bool sendflag)
 {
-	int xp, yp;
-	bool dok;
+	ObjectStruct &door = object[oi]; // Non const ref as flags are updated in this function
```
AJenbo, line 2658:
```suggestion
	ObjectStruct &door = object[oi];
```

Source/objects.cpp
```diff
@@ -2436,352 +2436,345 @@ void RedoPlayerVision()
...
-		object[oi]._oVar4 = 2;
+		door._oVar4 = 2;
 	}
 }
 
 void OperateL3RDoor(int pnum, int oi, bool sendflag)
 {
-	int xp, yp;
-	bool dok;
+	ObjectStruct &door = object[oi]; // Non const ref as flags are updated in this function
```
AJenbo, line 2701:
```suggestion
	ObjectStruct &door = object[oi];
```

Source/objects.cpp
```diff
@@ -2436,352 +2436,345 @@ void RedoPlayerVision()
...
-		object[oi]._oVar4 = 2;
+		door._oVar4 = 2;
 	}
 }
 
 void OperateL3LDoor(int pnum, int oi, bool sendflag)
 {
-	int xp, yp;
-	bool dok;
+	ObjectStruct &door = object[oi]; // Non const ref as flags are updated in this function
```
AJenbo, line 2742:
```suggestion
	ObjectStruct &door = object[oi];
```

AJenbo (approved):
Nice work, the sound functions where on of the things I was most looking forward to having the Point cleanup applied. Also nice work on the door functions.

Issue #2158, opened 2021-06-14
[enhancement proposal] Interacting with the door frames


Chance4us:
Well, what does the topic mean?
Anyone who has played the game for several years or maybe decades will know that in some unfavorable situations the doors cannot be opened.

As an example, below is the picture and savegame for the caves.
To try out the problem, just load the attached game and try to open the door without killing the monsters. You will notice that you are not able to select the door frame and open the door.

Of course, it's not a problem for every hero. Because the rogue or the sorcerer can shoot the monsters from the distance. In the case of the warrior or barbarian, however, it can look very hopeless to open the door.

In situation below, the problem can still be circumvented because the room is spacious,
but there can be no way out if the room is small and full of monsters.

So what could be a possible solution to the problem? Maybe it makes sense to activate the focus on the door frame by pressing the ALT key. Or maybe you have even better ideas?

![door](https://user-images.githubusercontent.com/45902449/121864744-867c0900-ccfd-11eb-8a23-ff229b7ef1b1.png)

Savegame:
[door.zip](https://github.com/diasurgical/devilutionX/files/6646808/door.zip)

AJenbo:
You can if you use a gamepad :D

qndel:
How about extending alt highlighting to objects? ;)

ChaosMarc:
I would love that change as it would also highlight chests being hidden by walls

qndel:
![blockdoor](https://user-images.githubusercontent.com/14297035/121880678-951eec00-cd0e-11eb-9179-b77875abf0de.gif)

Chance4us:
Instead of "potato obj" maybe better "open / close" description ;-)

FitzRoyX:
Even with the monsters gone, it feels overly difficult to select that gate. When you mouse over the top third of it, nothing happens.

qndel:
can you stop comparing stuff to belzebub which doesn't have a single line of code in common with diablo? 🙄 oh geez, 100% different codebase, they don't have the same bug, who would've thought ...

FitzRoyX:
It's just an example of the gate behaving how you would expect with all parts selectable. Good info considering everyone thus far is assuming it's just because the monsters are in front of it?

julealgon:
> When you mouse over the top third of it, nothing happens.

This is bad. I agree it should be fixed regardless of how it works in Belzebub. I think the same thing happens with clicking on stairs, which can be very spotty as well.

julealgon:
[...]

PR #2160, opened 2021-06-14
Object highlight

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2158

Most of the changes are from getting rid of the keyword `item` as labels became more generic right now (supporting both items and objects)
I've done file renaming in second commit so you can view only the first one to see the actual changes


Chance4us:
That works really wonderful. Thanks a lot.

sumyunggai:
@qndel I'm going to be the fly in the ointment on this one; as one of the nice things about Diablo is wonderfully old-school lack of visual noise in the gameplay screen. 

Since there is no image attached to this PR I don't exactly know your implementation, are you highlighting objects (not items) and writing labels over them? If so, I would argue that's too much visual noise and instead advocate for solely highlighting non-item objects for which we can interact.

Honestly, I'm not sure highlighting doors in this way is even appropriate. It's ugly, quite frankly.

StephenCWills:
> Since there is no image attached to this PR I don't exactly know your implementation, are you highlighting objects (not items) and writing labels over them? If so, I would argue that's too much visual noise and instead advocate for solely highlighting non-item objects for which we can interact.

I imagine this is an extension of the alt-highlighting feature for items. If you aren't holding Alt, you don't see it. So there wouldn't normally be any visual clutter, similar to how the map overlay doesn't appear if you don't press Tab.

Are you perhaps worried about what happens when there are a bunch of items around a door? In that case, I'm pretty sure the labels adjust themselves so they don't overlap.

qndel:
image

sumyunggai:
Thanks @qndel. @StephenCWills and @AJenbo I am considering that the use case for ALT-highlight will predominantly be to find items, not doors or chests/sarcophagi/bookcases/shrines/lecturns. In that dominant use case, the additional visual clutter from the labels for all non-items will actually impede the visual clarity provided by the highlight feature _for the purpose of finding items_.

Fundamentally, the question is: just because we can implement a feature, should we? I am not convinced that labels (or highlighting, to be honest) is needed for for non-items.

AJenbo:
@sumyunggai how would you handle opening doors like this or find chests behind cave walls?

sumyunggai:
> 
> 
> @sumyunggai how would you handle opening doors like this or find chests behind cave walls? Imo highlighting deals with the stuff that it's hard to locate stuff in the game with the mouse.

Sorry to answer a question with a question, but is there a check to see if an object is visible or obscured? If so, I would highlight, not label, the obscured items and make the highlight appear through the obscuring object with a 1-pixel shadow border around the highlight (1 pixel per side) to deal with the contrast of the highlight and whatever tile art is obscuring the item.

qndel:
> > @sumyunggai how would you handle opening doors like this or find chests behind cave walls? Imo highlighting deals with the stuff that it's hard to locate stuff in the game with the mouse.
> 
> Sorry to answer a question with a question, but is there a check to see if an object is visible or obscured? If so, I would highlight, not label, the obscured objects and make the highlight appear through the obscuring object with a 1-pixel shadow border around the highlight (1 pixel per side) to deal with the contrast of the highlight and whatever tile art is obscuring the object.

There's no such a thing

sumyunggai:
there must be a probe at least on the player, it's how transparency is handled when behind tile art right?

StephenCWills:
> there must be a probe at least on the player, it's how transparency is handled when behind tile art right?

I'm pretty sure the player is not visible through Cave walls. Also, I'm fairly certain transparency isn't handled by identifying objects that overlap, but rather by blending a transparent object's color with the color of the pixel underneath.

AJenbo:
> there must be a probe at least on the player, it's how transparency is handled when behind tile art right?

It's simply done by room, if you stand in a room all it's lower walls will be rendered transparent. But the engine doesn't know if the wall is tall enough to overlap an item or not. Identifying this would require rendering items to a G-buffer and probe it for every wall pixel being rendered. This would both be complicated hard and extremely slow with how things are currently.

StephenCWills:
> It's simply done by room, if you stand in a room all it's lower walls will be rendered transparent. But the engine doesn't know if the wall is tall enough to overlap an item or not. Identifying this would require rendering items to a G-buffer and probe it for every wall pixel being rendered. This would both be complicated hard and extremely slow with how things are currently.

I think I understand the line of thinking now. What if we draw the pixels for object/item highlights to a separate buffer during rendering and then add logic to tile rendering to check for pixels in the highlight buffer to render instead of tile pixels?

Chance4us:
I think it's not about visual optics at all, but about a possibility to open a door that cannot be opened.
@qndel's solution is already most simple and effective imho. For those who don't like this, I suggest to add a disable option to the ini file.

StephenCWills:
> I think it's not about visual optics at all, but about a possibility to open a door that cannot be opened.
> @qndel's solution is already most simple and effective imho.

Right, but alt-highlighting was introduced in the first place for finding items that had dropped behind walls or in other places that are difficult to locate. In context of the solution provided by this PR, the two issues have become linked and are therefore both relevant to the discussion here.

> For those who don't like this, I suggest to add a disable option to the ini file.

I would argue that the issue of clicking on an obscured door needs to be fixed, and that fix shouldn't be toggleable regardless. If we plan on toggling object highlights via the INI, perhaps we shouldn't be using alt-highlighting as the solution to this problem in the first place.

Chance4us:
I noticed that now all objects are highlighted, including chests, fountains, barrels, slain heros.
It would be sufficient if only the items and doors are highlighted. The other objects do not need to be highlighted.

AJenbo:
They do if they can be obscured by walls

Chance4us:
I have nothing against that, only the spring shouldn't be labeled four times.
spring
[spring.zip](https://github.com/diasurgical/devilutionX/files/6651453/spring.zip)

julealgon:
> I think it's not about visual optics at all, but about a possibility to open a door that cannot be opened.

Not a fan of the fact that the fix for this situation is behind a toggleable, mostly visual (until now) thing.

A suggestion perhaps we could explore is giving priority to the door if one is behind the cursor, so that it would be selected even if there is a monster on top of most of its sprite.

This wouldn't mess much with being able to attack the monster since you can still hover over its actual tile, or use shift, or cast behind it.

Chance4us:
If "Disable Crippling Shrines" is activated and the hero is approaching a cauldron, the items in inventory and belt change the color to red.
[output.mkv.zip](https://github.com/diasurgical/devilutionX/files/6681298/output.mkv.zip)

qndel:
> If "Disable Crippling Shrines" is activated and the hero is approaching a cauldron, the items in inventory and belt change the color to red.
> [output.mkv.zip](https://github.com/diasurgical/devilutionX/files/6681298/output.mkv.zip)

only with object highlighting?

Chance4us:
@qndel Yes in fact. I absolutely have this impression.
You can also double-check with the following savegame. I don't want to say anything inapplicable.
[cauldron.zip](https://github.com/diasurgical/devilutionX/files/6682021/cauldron.zip)

qndel:
> @qndel Yes in fact. I absolutely have this impression.
> You can also double-check with the following savegame. I don't want to say anything inapplicable.
> [cauldron.zip](https://github.com/diasurgical/devilutionX/files/6682021/cauldron.zip)

Thanks I think I know what's going on

qndel:
image
labels will have the same color as the object now :)

Chance4us:
I tried again and the colors are great, but now the objects could no longer be selected on the text field and rect.

qndel:
> I tried again and the colors are great, but now the objects could no longer be selected on the text field and rect.

what do you mean? selecting objects with label works fine for me

Chance4us:
Sorry, that only happens in my devilutionx version, which I cobbled together myself.
Since I'm not a good programmer, it's hardly surprising.
As for the thing you described, I'll test as soon as there is free time.

Chance4us:
@qndel 
I've played through almost an entire game now and haven't seen any problems.

Because not everyone likes this feature, I would like to suggest that we separate the keyboard shortcuts as follows.
1. Highlight only items when "left alt" pressed, or not highlight if "invertHighlightToggle" for items is true.
2. Highlight only objects when "right alt" pressed, or not highlight if "invertHighlightToggle" for objects is true.
3. "left ctrl" for toggle between "always highlight" and "never highlight" for items
4. "right ctrl" for toggle between "always highlight" and "never highlight" for objects

AJenbo:
Use the keymapper instead of hardcoding keys

Chance4us:
> Use the keymapper instead of hardcoding keys

I generally meant that item highlight should use other buttons than object highlight.
Because if I were to remap the buttons for item highlight with the keymapper, the same buttons would apply to the object highlight. Another advantage would be that item highlight could be switched on and off separately from object highlight.

qndel:
> > Use the keymapper instead of hardcoding keys
> 
> I generally meant that item highlight should use other buttons than object highlight.
> Because if I were to remap the buttons for item highlight with the keymapper, the same buttons would apply to the object highlight. Another advantage would be that item highlight could be switched on and off separately from object highlight.
> The configuration above is an example of the standard buttons we could use.

im not splitting it

Chance4us:
> im not splitting it

Should i try this when i have a little more time?

sumyunggai:
> 
> 
> > Use the keymapper instead of hardcoding keys
> 
> I generally meant that item highlight should use other buttons than object highlight.
> Because if I were to remap the buttons for item highlight with the keymapper, the same buttons would apply to the object highlight. Another advantage would be that item highlight could be switched on and off separately from object highlight.
> The configuration above is an example of the standard buttons we could use.

I think this is a good, balanced approach to help preserve the aesthetics of the game.

julealgon:
I tend to agree with the split in functionality proposed by @Chance4us as well. If players want, they could bind all of them to the same key, but this should ideally be optional.

I for one think object highlight looks bad, for example, and would probably not use it.

FitzRoyX:
Dunno about anyone else, but most of my visibility problems are with small items, not objects. I wouldn't personally want labels for big objects that are already highly visible like fountains/bookshelves/shrines/sarcos. I think I'd also rather have an outline than a text label for objects. This isn't to poop on the feature, it does have some value to someone but I definitely think it should be an ini boolean and not married to item highlighting.

qndel:
> Dunno about anyone else, but most of my visibility problems are with small items, not objects. I wouldn't personally want labels for big objects that are already highly visible like fountains/bookshelves/shrines/sarcos. I think I'd also rather have an outline than a text label for objects. This isn't to poop on the feature, it does have some value to someone but I definitely think it should be an ini boolean and not married to item highlighting.

The entire point of this feature is being able to click doors blocked by monsters, this won't be achieved with just "highlighting the outline" + what about objects hidden by walls in caves etc. ? We don't know what's visible and what's not

Chance4us:
> The entire point of this feature is being able to click doors blocked by monsters, this won't be achieved with just "highlighting the outline" + what about objects hidden by walls in caves etc. ? We don't know what's visible and what's not

I agree here. I think we can find a compromise.

PR #2163, opened 2021-06-14
Vita: resolution changing support

isage:
Fixes https://github.com/diasurgical/devilutionX/discussions/2162


Source/diablo.cpp
```diff
@@ -587,20 +585,22 @@ static void LoadOptions()
...
+#elif defined(__vita__)
+	sgOptions.Graphics.bUpscale = true;
 #else
 	sgOptions.Graphics.bUpscale = false;
 #endif
+#ifndef __vita__
+	sgOptions.Graphics.bFullscreen = getIniBool("Graphics", "Fullscreen", true);
 	sgOptions.Graphics.bFitToScreen = getIniBool("Graphics", "Fit to Screen", true);
+#else
+	sgOptions.Graphics.bFitToScreen = true;
```
AJenbo, line 601: Why not leave this as configurable? Same with bUpscale.

isage: That currently leads to stretched image instead of black bars.
For upscale: as far as i understood from wiki, turning upscale off will lead to trying to set window/fullscreen resolution to the internal one, which is impossible on vita right now

AJenbo: Hmm yeah I think under SDL2 it would actually give you the same internal scaling as the native one. For SDL1 you should get black bars on all sides.

AJenbo (approved):

isage:
Closed in favor of #2175

PR #2164, opened 2021-06-14
Leverage `CelSprite` for `MonsterStruct`

obligaron:
Notes:

- Similar to #2043
- Only refactoring. No behavior or animation changes.
- First Step to get ADL for Monsters

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2497,11 +2497,12 @@ void InitMissileAnimationFromMonster(MissileStruct &mis, int midir, const Monste
 	const AnimStruct &anim = mon.MType->Anims[graphic];
 	mis._mimfnum = midir;
 	mis._miAnimFlags = 0;
-	mis._miAnimData = anim.Data[midir];
+	auto& celSprite = *anim.CelSpritesForDirections[midir];
+	mis._miAnimData = celSprite.Data();
 	mis._miAnimDelay = anim.Rate;
 	mis._miAnimLen = anim.Frames;
-	mis._miAnimWidth = mon.MType->width;
-	mis._miAnimWidth2 = CalculateWidth2(mon.MType->width);
+	mis._miAnimWidth = celSprite.Width();
+	mis._miAnimWidth2 = CalculateWidth2(celSprite.Width());
 	mis._miAnimAdd = 1;
 	mis._miVar1 = 0;
 	mis._miVar2 = 0;
@@ -4646,7 +4647,7 @@ void MI_Fireman(int i)
 	if (!PosOkMissile(0, b) || (j > 0 && !(missile[i]._miVar1 & 1))) {
 		missile[i].position.velocity *= -1;
 		missile[i]._mimfnum = opposite[missile[i]._mimfnum];
-		missile[i]._miAnimData = monster[src].MType->Anims[MA_WALK].Data[missile[i]._mimfnum];
+		missile[i]._miAnimData = monster[src].MType->Anims[MA_WALK].CelSpritesForDirections[missile[i]._mimfnum]->Data();
 		missile[i]._miVar2++;
 		if (j > 0)
 			missile[i]._miVar1 |= 1;
@@ -5192,7 +5193,7 @@ void missiles_process_charge()
 				else
 					anim = &mon->Anims[MA_WALK];
 			}
-			missile[mi]._miAnimData = anim->Data[mis->_mimfnum];
+			missile[mi]._miAnimData = anim->CelSpritesForDirections[mis->_mimfnum]->Data();
 		}
 	}
 }
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2491,64 +2491,54 @@ void AddMetlHit(int mi, int sx, int sy, int dx, int dy, int midir, int8_t mienem
 	missile[mi]._mirange = missile[mi]._miAnimLen;
 }
 
+namespace {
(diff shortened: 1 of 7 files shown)


Source/monster.h
```diff
@@ -193,6 +193,11 @@ struct MonsterStruct { // note: missing field _mAFNum
 	const char *mName;
 	CMonster *MType;
 	const MonsterData *MData;
+
+	/*
```
AJenbo, line 197:
```suggestion
	/**
```

obligaron: fixed 🙂

Source/monster.h
```diff
@@ -148,7 +148,7 @@ struct MonsterStruct { // note: missing field _mAFNum
 	int _menemy;
 	/** Usually correspond's to the enemy's future position */
 	Point enemyPosition;
-	byte *_mAnimData;
+	CelSprite *_mAnimData;
```
ephphatha, line 151: I see this is declared as a pointer to avoid creating a copy of a CelSprite whenever the animation changes. It looks like making this a `std::reference_wrapper<std::optional<CelSprite>>` type would save having to constantly get and then take the address of the optional, while also allowing the removal of the two uses of `std::optional::has_value` in monster.cpp

Source/monster.h
```diff
@@ -193,6 +193,11 @@ struct MonsterStruct { // note: missing field _mAFNum
 	const char *mName;
 	CMonster *MType;
 	const MonsterData *MData;
+
+	/**
+	 * @brief Check thats the correct stand Animation is loaded. This is needed if direction is changed (monster stands and looks to player).
```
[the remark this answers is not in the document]
obligaron: fixed 🙂

Source/missiles.cpp
```diff
@@ -2533,22 +2541,11 @@ void AddFireman(int mi, int sx, int sy, int dx, int dy, int midir, int8_t mienem
 
 	anim = &monster[id].MType->Anims[MA_WALK];
```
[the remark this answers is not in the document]
obligaron: fixed, but i hesitated cause it's not directly related to this pr. But I did it anyway 😉

Source/missiles.cpp
```diff
@@ -2491,6 +2491,25 @@ void AddMetlHit(int mi, int sx, int sy, int dx, int dy, int midir, int8_t mienem
 	missile[mi]._mirange = missile[mi]._miAnimLen;
 }
 
+namespace {
+void InitMissileAnimationFromMonster(MissileStruct &mis, int midir, const MonsterStruct &mon, const AnimStruct *anim)
```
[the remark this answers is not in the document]
obligaron: Instead I passed the graphic-id. Thats even more simple. Take a look. 🙂

Source/monster.h
obligaron: The design was introduced for items (see pr #1796). I follow that design.
This pr only unify monster implementation with items and players (That is a prerequisite for ADL).

Personally I would stick with a `shared_ptr` instead of a optional. But I don't have a strong opinion in that case.

Anyway I would rather not change this as part of this pr as this affects also items and players and that means it's outside of the scope of this. But that doesn't mean it can't be changed later. 🙂

Source/missiles.cpp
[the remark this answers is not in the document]
ephphatha: Mainly mentioned this variable as it was only used in the function you created to pull out shared code. With the change to use an animation ID it's been completely removed anyway so that's great 😀

Source/monster.h
glebm: `shared_ptr` is a code smell in C++ most of the time -- every object should have an owner, it is rare that there really isn't a clear one.

Source/scrollrt.cpp
```diff
@@ -690,10 +690,12 @@ static void DrawMonsterHelper(const CelOutputBuffer &out, int x, int y, int oy,
 		return;
 	}
 
-	px = sx + pMonster->position.offset.x - CalculateWidth2(pMonster->MType->width);
+	CelSprite *pCelSprite = pMonster->_mAnimData;
```
glebm, line 693:
```suggestion
	const CelSprite &pCelSprite = *pMonster->_mAnimData;
```

Source/scrollrt.cpp
```diff
@@ -315,7 +315,7 @@ static void DrawMonster(const CelOutputBuffer &out, int x, int y, int mx, int my
 	}
 
 	int nCel = monster[m]._mAnimFrame;
-	auto frameTable = reinterpret_cast<const uint32_t *>(monster[m]._mAnimData);
+	auto frameTable = reinterpret_cast<const uint32_t *>(monster[m]._mAnimData->Data());
 	int frames = SDL_SwapLE32(frameTable[0]);
```
glebm, line 319: Unrelated to this PR but it's worth highlighting a bug here -- `Data` is not guaranteed to be 4-byte aligned so this is illegal (e.g. it can cause a crash on MIPS CPUs)

Source/scrollrt.cpp
obligaron: fixed 🙂

Source/scrollrt.cpp
obligaron: This logic is used in multiple places to validate the `CelSprite`.
It would be great if we have method that would perform this validations and also fixes this bug.

glebm (approved):

AJenbo (approved):

PR #2166, opened 2021-06-15
♻️ Use '_item_indexes' instead of 'int' where possible

julealgon:
This PR leverages the `_item_indexes` enum in a multitude of places where raw `int`s were being used to represent item IDs.

This is a preparation for converting the enum into an enum class and adding missing entries.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -322,7 +322,7 @@ int premiumLvlAddHellfire[] = {
 	// clang-format on
 };
 
-bool IsItemAvailable(int i)
+bool IsItemAvailable(_item_indexes i)
 {
 	if (gbIsSpawn) {
 		if (i >= 62 && i <= 71)
@@ -2260,7 +2260,7 @@ int RndItem(int m)
 
 	ri = 0;
 	for (i = 0; AllItemsList[i].iLoc != ILOC_INVALID; i++) {
-		if (!IsItemAvailable(i))
+		if (!IsItemAvailable(static_cast<_item_indexes>(i)))
 			continue;
 
 		if (AllItemsList[i].iRnd == IDROP_DOUBLE && monster[m].mLevel >= AllItemsList[i].iMinMLvl
@@ -2294,7 +2294,7 @@ _item_indexes RndUItem(int m)
 	int curlv = items_get_currlevel();
 	int ri = 0;
 	for (int i = 0; AllItemsList[i].iLoc != ILOC_INVALID; i++) {
-		if (!IsItemAvailable(i))
+		if (!IsItemAvailable(static_cast<_item_indexes>(i)))
 			continue;
 
 		okflag = true;
@@ -2336,7 +2336,7 @@ _item_indexes RndAllItems()
 	int curlv = items_get_currlevel();
 	int ri = 0;
 	for (int i = 0; AllItemsList[i].iLoc != ILOC_INVALID; i++) {
-		if (!IsItemAvailable(i))
+		if (!IsItemAvailable(static_cast<_item_indexes>(i)))
 			continue;
 
 		if (AllItemsList[i].iRnd != IDROP_NEVER && 2 * curlv >= AllItemsList[i].iMinMLvl && ri < 512) {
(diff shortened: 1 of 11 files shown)


Source/items.cpp
```diff
@@ -2323,20 +2323,20 @@ int RndUItem(int m)
 		}
 	}
 
-	return ril[GenerateRnd(ri)];
+	return static_cast<_item_indexes>(ril[GenerateRnd(ri)]);
```
AJenbo, line 2326: Make the array use the item ID type as well

Source/items.cpp
```diff
@@ -4153,12 +4155,12 @@ int RndVendorItem(int minlvl, int maxlvl)
...
 
-	return ril[GenerateRnd(ri)] + 1;
+	return static_cast<_item_indexes>(ril[GenerateRnd(ri)] + 1);
 }
 
-int RndSmithItem(int lvl)
+_item_indexes RndSmithItem(int lvl)
 {
-	return RndVendorItem<SmithItemOk, true>(0, lvl);
+	return static_cast<_item_indexes>(RndVendorItem<SmithItemOk, true>(0, lvl) - 1);
```
AJenbo, line 4163: Is it not possible to have this return the enum? How about a struct that can also represent the book that would normally be represented via negative numbers

AJenbo (changes requested):
I would prefer that the needed values where added before converting thing so that we do not have a commit that adds a lot of temp casts that will subsequently be removed

julealgon:
> I would prefer that the needed values where added before converting thing so that we do not have a commit that adds a lot of temp casts that will subsequently be removed

If we add the values beforehand, we'd have to follow the existing naming convention which would then cause all values to be renamed again when converting to a `enum class`, and that would be vastly more painful IMHO. I did it this way exactly to reduce the number of changes needed during the conversion.

Source/items.cpp
julealgon: Not really, because of the weird subtraction. Even if `RndVendorItem` returns the enum (which it does) the subtraction forces we to convert again.

Source/items.cpp
julealgon: Not sure this would be a lot better, as we'll just move the `static_cast` to the indexer.

I think it would be better to come back to this once we change how `AllItems` array is accessed (not via an array, preferably).

AJenbo:
Can't we start with the change to `enum class` then?

julealgon:
> Can't we start with the change to enum class then?

The change to enum class, if done first, would force potentially hundreds of `static_cast<enum>` calls, since `int`s would not be accepted anymore. I think we should actually leave it to last in this case.

AJenbo:
Ok, i think I will wait with merging until more of the series is done as I can't really see what direction it's heading atm.

AJenbo:
Lots of casting is hard to deal with and and feels a bit fragile, it's not very reader friendly either. It just feels a bit unfinished in this stage. Changing the type of all variables and adding the minimum necessary entries should be enough for it to feel cohesive. Renaming, changing the type and additional entries can come later if you prefer.

julealgon:
> Lots of casting is hard to deal with and and feels a bit fragile, it's not very reader friendly either. It just feels a bit unfinished in this stage. Changing the type of all variables and adding the minimum necessary entries should be enough for it to feel cohesive.

Fair enough... I'll see what I can do. As I said, I don't think there is a way to eliminate the casts considering the logic itself is incrementing values to find items, unless we completely change the underlying logic.

I guess I can come up with another refactoring that will help with this, to stop using raw arrays for `AllItems`

Source/msg.h
```diff
@@ -210,7 +210,7 @@ struct TCmdGItem {
 	uint8_t bLevel;
 	uint8_t x;
 	uint8_t y;
-	uint16_t wIndx;
+	_item_indexes wIndx;
```
vlolteanu, line 213: These structs get passed over the network as-is. Here it pays to have exact control over the fields' sizes, irrespective of what else is happening in the codebase. I'd leave the standard ints in place.

AJenbo: _item_indexes is set to be backed up by uint16_t, using _item_indexes is actually better as it means that mods would only have to adjust the size of _item_indexes if they add more items to the game

vlolteanu: Do we want to support multiplayer:
 * across different versions of DevilutionX? (Probably not.) If so, we can't afford to change the message format.
 * across different architectures? If so, we need to do byte swaps to keep the endianness consistent. Then the enum loses its meaning in this context.

AJenbo: The game already checks the version and refuses to join if the versions do not match.

Network endianness is being worked on, still needs some testing:
https://github.com/diasurgical/devilutionX/tree/endian

julealgon:
Abandoning this.

PR #2168, opened 2021-06-15
Android - move version to build.gradle

licaon-kter:


AJenbo:
Would it be possible to get the version string from CMake?

licaon-kter:
To enable autoupdates for F-Droid a static defined versionCode/Name in build.gradle work best.

From CMake how?

Benefits? Dunno, just that most "modern" apps moved them to build.gradle, not sure if Android Studio will raise some alerts and ask to move them too.

StephenCWills:
> From CMake how?

CMake uses the `VERSION_NUM` variable for the version number (major.minor.patch syntax) as well as the `VERSION_SUFFIX` to identify the commit hash for debug builds. If there were a way to export CMake variables to Gradle, and if the versionName could be supplied by the exported variables, then we could easily use the same versioning logic we use on other platforms. Ideally, versionCode could be computed from `PROJECT_VERSION_MAJOR`, `PROJECT_VERSION_MINOR`, and `PROJECT_VERSION_PATCH` so that it always increases when the version number does (binary-encoded decimal?).

If we can't export CMake variables or if we can't use exported variables to set the version number in Gradle, an alternative approach could be to replicate the CMake versioning logic in Gradle and use that to generate `versionName` and `versionCode`.

https://github.com/diasurgical/devilutionX/blob/3f7ad14306220e9b6f97b07a213a4db21bf9c66b/CMakeLists.txt#L78-L88
https://github.com/diasurgical/devilutionX/blob/3f7ad14306220e9b6f97b07a213a4db21bf9c66b/CMake/git.cmake#L1-L17

licaon-kter:
FYI, ATM, F-Droid won't run any functions/calculations in the checkupdate phase, ideally that would be something to be run in a clean VM that we'd destroy ofter, for safety reasons, but we're not there yet.

StephenCWills:
> FYI, ATM, F-Droid won't run any functions/calculations in the checkupdate phase, ideally that would be something to be run in a clean VM that we'd destroy after, for safety reasons, but we're not there yet.

I'm sorry, I don't know anything about F-Droid. Does that checkupdate phase read the version straight out of the Gradle script rather than retrieving it from a compiled APK?

licaon-kter:
F-Droid builds everything from source, so we must have this info before doing anything.

Again, it's a limitation of the automated checkupdate, also given then pace of releases around here I don't see an issue to have versionName/Code updated and Tag a new version right after.

We won't build each commit, not a CI, so no need for hash info and whatnot.

AJenbo:
> given then pace of releases around here I don't see an issue to have versionName/Code updated and Tag a new version right after.

If you are saying that it's slow, then that is more the reason to have it automatically propagated from the single source of truth instead of having to manually remember to update it in multiple locations for different ports. It's easy to forget or miss.

AJenbo (approved):

StephenCWills:
> F-Droid builds everything from source, so we must have this info before doing anything.

Sorry, but I'm still not snapping in. How does F-Droid obtain the source code? Do we have to upload it, or can it use Git? Does it get the version number by parsing the Gradle script or does it build the APK and get the version number from there? Basically, why should we have to hardcode it in the source tree if it ends up in the APK either way?

> Again, it's a limitation of the automated checkupdate, also given then pace of releases around here I don't see an issue to have versionName/Code updated and Tag a new version right after.

IMO, this isn't necessarily about the pace of releases, but rather the process for releasing. AFAIK, we have no formal process so we'll be scrambling to update documentation and create all the install packages for all our different platforms manually. All other platforms get the version number automatically via CMake by querying Git tags so it would be nice if F-Droid could do the same.

licaon-kter:
@AJenbo not slow, but rar(er) :)


@StephenCWills `checkupdate` clones, looks at a the latest Tag and tries to find a newer (than what's available) versionCode&Name. If it finds it, it adds a new build block to the recipe, for this new version

When the main buildserver runs, it looks if a new version is in the recipe, if so, clones, checksout the Tag or commit hash, thes builds.

If the single source of version truth is not a parseable string, not necessarily in build gradle eg. It can parse any file to extract it, but it's created dynamically via some function, for the moment, it can't automatically pick it up.

As with other apps this needs me/you/fans to either open a MR with thu added version block or at least open an issue (on top of the current many :)).

I'm following & manually updating at least OMW_Nightly (stable is on autoupdate), I could follow this too, for now.

AJenbo:
> When the main buildserver runs, it looks if a new version is in the recipe, if so, clones, checks out the Tag or commit hash, then builds it.

This sounds like it could easily pick up a commit that is part of a preperation for a release and not the release it self.

> If the single source of version truth is not a parseable string, not necessarily in build gradle eg. It can parse any file to extract it, but it's created dynamically via some function, for the moment, it can't automatically pick it up.

The version is tagged in git, this is the SSOT, if F-Droid was able to check git tags it should both be faster and more accurate.

> As with other apps this needs me/you/fans to either open a MR with thu added version block or at least open an issue (on top of the current many :)).

Well, thank you for doing this :)

licaon-kter:
> Well, thank you for doing this :)

Can this be merged in the mean time? :)

> This sounds like it could easily pick up a commit that is part of a preperation for a release and not the release it self.
> The version is tagged in git, this is the SSOT, if F-Droid was able to check git tags it should both be faster and more accurate.

As said above, Tags are used, the issue is that once a Tag is checked out, it needs to find the corresponding info about the new version, which Code? which Name?

StephenCWills (approved):

AJenbo:
> As said above, Tags are used, the issue is that once a Tag is checked out, it needs to find the corresponding info about the new version, which Code? which Name?

Yeah we probably wouldn't get around the version code easily, unless we just had it auto increment. It's a bit odd that Android has this it makes it impossible to have feature branches, or at the very least it makes it rather hard (you could partition the version code so that 10300 - 10399 was reserved for bugfixes for 1.3.x)

licaon-kter:
Yes, you need to be careful around versionCode as it needs to be _always incrementing_, so now it's 18 but you'll need to jump to 10300 on the next release if you plan to bugfix older versions. Note that F-Droid builds latest available so we tend not to pickup any of those if a newer (bigger) version has been built.

You're thinking more akin to an app in a Linux distro :)

PR #2171, opened 2021-06-15
Android - Enable multiplayer

licaon-kter:


qndel:
is it multiplayer if libsodium is off and zerotier is disabled?

StephenCWills:
> is it multiplayer if libsodium is off and zerotier is disabled?

System libsodium is off. It means the build script will download the source and add it to the build process.

licaon-kter:
Note that this was as advised by @AJenbo and works in my test build via my local fdroid build server. :)

licaon-kter:
Is zerotier FOSS? Can I build this?

AJenbo:
> Is zerotier FOSS? Can I build this?

I belive so, but it spat some errors at me so moved on to better things.

android-project/app/src/main/AndroidManifest.xml
```diff
@@ -35,6 +35,9 @@
     <uses-permission android:name="android.permission.BLUETOOTH" />
     <!-- Allow access to the vibrator -->
     <uses-permission android:name="android.permission.VIBRATE" />
+    <!-- Allow access to the network for multiplayer -->
+    <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
```
AJenbo, line 40: Since we do not make use of this API i think it's premature to request access to it. I'm not sure it would be relevant either since it is not the only way for a device to join a game.

licaon-kter: True, that was more for completeness.

tsunamistate:
> Is zerotier FOSS? Can I build this?

Kind of? But not really
https://github.com/zerotier/ZeroTierOne/blob/master/LICENSE.txt

licaon-kter:
@tsunamistate if it's not FSF and/or OSI approved...I don't touch it, meh

PR #2175, opened 2021-06-15
Re-enable graphical settings on PSVita

isage:
Continuation of #2163
(still need to wait until SDL2 package is updated in vitasdk)


Source/diablo.cpp
```diff
@@ -507,14 +507,12 @@ static void SaveOptions()
...
-#ifndef __vita__
 	setIniInt("Graphics", "Width", sgOptions.Graphics.nWidth);
 	setIniInt("Graphics", "Height", sgOptions.Graphics.nHeight);
-#endif
-	setIniInt("Graphics", "Fullscreen", sgOptions.Graphics.bFullscreen);
 #ifndef __vita__
-	setIniInt("Graphics", "Upscale", sgOptions.Graphics.bUpscale);
+	setIniInt("Graphics", "Fullscreen", sgOptions.Graphics.bFullscreen);
 #endif
+	setIniInt("Graphics", "Upscale", sgOptions.Graphics.bUpscale);
```
AJenbo, line 515: Looks like this should only be exposed when `#if !defined(USE_SDL1)` like in loadOptions

PR #2178, opened 2021-06-16
Clip UI draw calls

glebm:
Here it is at 480x480:

Screenshot at 480x480

Not particularly useful but allows for easier experimentation without crashing.

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -36,6 +36,52 @@ const uint32_t RndInc = 1;
  */
 const uint32_t RndMult = 0x015A4E35;
 
+namespace {
+
+template <bool SkipColorIndexZero>
+void BufferBlit(const CelOutputBuffer &src, SDL_Rect srcRect, const CelOutputBuffer &dst, Point dstPosition)
+{
+	// We do not use `SDL_BlitSurface` here because the palettes may be different objects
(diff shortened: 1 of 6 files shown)


AJenbo:
> Not particularly useful but allows for easier experimentation without crashing.

This will perfectly solve the issue when running it on a smart watch

image

Source/engine.cpp
```diff
@@ -36,6 +36,52 @@ const uint32_t RndInc = 1;
  */
 const uint32_t RndMult = 0x015A4E35;
 
+namespace {
+
+template <bool SkipColorIndexZero>
+void BufferBlit(const CelOutputBuffer &src, SDL_Rect srcRect, const CelOutputBuffer &dst, Point dstPosition)
+{
+	// We do not use `SDL_BlitSurface` here because the palettes may be different objects
```
AJenbo, line 44: Would it make sens to check if they use the same palette and then use SDL?

glebm: Maybe as an optimization but I think this is unlikely to make much difference. We'd also need to fiddle around with COLOR_KEY for the SkipColorIndexZero version, and that's a bit trickier.

AJenbo (approved):

PR #2180, opened 2021-06-16
🐞 Fix off-by-one in `DrawCursor`

### LINKED ISSUE #2179: Artefacts when club is above the bottom panel
Seems like a somewhat recent regression:

screenshot

Strangely enough, this doesn't seem to happen with the sword and shield

glebm:
The correct size is +2px for the outline, because the outline is on both sides.

Fixes #2179

### DIFF
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -207,10 +207,10 @@ static void DrawCursor(const CelOutputBuffer &out)
 		return;
 
 	sgdwCursX = std::max(MouseX - 1, 0);
-	sgdwCursWdt = MouseX < 0 ? cursW + MouseX + 1 : std::min(MouseX + cursW + 1, out.w()) - MouseX;
+	sgdwCursWdt = MouseX < 0 ? cursW + MouseX + 2 : std::min(MouseX + cursW + 2, out.w()) - MouseX;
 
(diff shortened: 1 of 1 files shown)


Source/scrollrt.cpp
```diff
@@ -207,10 +207,10 @@ static void DrawCursor(const CelOutputBuffer &out)
 		return;
 
 	sgdwCursX = std::max(MouseX - 1, 0);
-	sgdwCursWdt = MouseX < 0 ? cursW + MouseX + 1 : std::min(MouseX + cursW + 1, out.w()) - MouseX;
+	sgdwCursWdt = MouseX < 0 ? cursW + MouseX + 2 : std::min(MouseX + cursW + 2, out.w()) - MouseX;
 
 	sgdwCursY = std::max(MouseY - 1, 0);
-	sgdwCursHgt = MouseY < 0 ? cursH + MouseY + 1 : std::min(MouseY + cursH + 1, out.h()) - MouseY;
+	sgdwCursHgt = MouseY < 0 ? cursH + MouseY + 2 : std::min(MouseY + cursH + 2, out.h()) - MouseY;
```
julealgon, line 213: Would be nice to extract a constant with a good name to represent what this "2" is.

PR #2182, opened 2021-06-16
🚚 engine.h: Extract some code to separate files

glebm:
`engine.h` is getting quite bloated. Moves this code to their own files.


julealgon:
I'm a bit confused by this proposal. Didn't you say it is common in C++ to have multiple classes/structs/etc in the same file since they are in the same "translation unit" or something? 

Now we are using "it's too bloated" criteria for splitting?

Can you elaborate on this a little bit please? I actually added some of the types myself to `engine.h` on purpose with that assumption in mind, instead of creating new files as it would be much more standard for me coming from C#.

glebm:
Suggested reading about translation units: https://www.toptal.com/c-plus-plus/c-plus-plus-understanding-compilation

What you basically want is functions defined in `cpp` files (for `hpp` it doesn't matter) to be able to inline into each other where it makes sense. There is nothing like that here.

vlolteanu:
It's also worth getting rid of these from monsters.{hpp,cpp} and turning them into constexpr functions in direction.hpp:

`extern Direction left[8];
extern Direction right[8];
extern Direction opposite[8];`

glebm:
> It's also worth getting rid of these from monsters.{hpp,cpp} and turning them into constexpr functions in direction.hpp:

Sounds like these should just be `constexpr` in the direction header. Out of scope for this PR though.

glebm:
@AJenbo Now there are conflicts :(

AJenbo:
Sorry, my mom visited this weekend...

AJenbo (approved):

AJenbo (approved):

AJenbo (dismissed):

AJenbo:
@glebm easy enough to resolve, feel free to merge when build passes if you agree with the rebase

glebm:
This needed another rebase, should be good to merge now

AJenbo (approved):

PR #2184, opened 2021-06-16
Update ASIO to diasurgical fork

StephenCWills:
This update fixes crashes on Windows when playing TCP and enables 3DS support in ASIO.


AJenbo:
p.s. tar.gz is often smaller

PR #2187, opened 2021-06-17
Android release fixes

AJenbo:


CMakeLists.txt
```diff
@@ -196,6 +196,10 @@ if(ANDROID)
   set(DISABLE_ZERO_TIER ON)
   set(TTF_FONT_DIR \"\")
 
+  if(BINARY_RELEASE OR CMAKE_BUILD_TYPE STREQUAL "Release")
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -flto=full")
```
AJenbo, line 200: Without this we get an error stating: `Optimization level must be between 0 and 3` when doing a release build.

I don't know if this is relevant for other targets or if there is a better way to specify it :shrug:

StephenCWills: Maybe [add_compile_options()](https://cmake.org/cmake/help/latest/command/add_compile_options.html#command:add_compile_options)?

Also, it might be about time to add an `android_defs.cmake` for all these custom settings.

glebm: Interesting, where did you find the `-flto=full` thing? Most compilers these days do thin LTO.

glebm (changes requested):
Looks good, though at this point it'd be nice to move the Android definitions to `android_defs.cmake`, similar to how we do this for all the other platforms.

CMakeLists.txt
AJenbo: https://stackoverflow.com/questions/56238019/linker-error-when-enabling-link-time-optimization-in-ndk

licaon-kter:
While you're at it, is `-DANDROID_APP_PLATFORM=android-21` really needed? I saw some warning that gradle plugin already uses platform as defined by minSDK

CMakeLists.txt
glebm: Let's add a comment with a link to the NDK issue? https://github.com/android/ndk/issues/721

AJenbo:
> While you're at it, is `-DANDROID_APP_PLATFORM=android-21` really needed? I saw some warning that gradle plugin already uses platform as defined by minSDK

Looks like it isn't needed.

glebm (approved):

PR #2189, opened 2021-06-18
🎉 Hardware cursor (SDL2-only)

glebm:
Disabled by default because of these issues:

1. When clicking on inventory item, it briefly appears a bit shifted (in the wrong coordinates).

   This issue can happen with software cursor as well, but is a lot more obvious with the hardware cursor.

2. Cursor is scaled with nearest-neighbour scaling, which may look a bit different from how the rest of the graphics are scaled.

See also previous attempt: https://github.com/diasurgical/devilutionX/pull/955 by @viciious

### DIFF
--- a/Source/DiabloUI/credits.cpp
+++ b/Source/DiabloUI/credits.cpp
@@ -10,6 +10,7 @@
 #include "DiabloUI/support_lines.h"
 #include "control.h"
 #include "controls/menu_controls.h"
+#include "hwcursor.hpp"
 #include "utils/display.h"
 #include "utils/sdl_compat.h"
 #include "utils/sdl_ptrs.h"
@@ -198,6 +199,9 @@ bool TextDialog()
 {
 	CreditsRenderer creditsRenderer;
 	bool endMenu = false;
+
+	if (IsHardwareCursorEnabled())
+		SetHardwareCursorVisible(false);
 
 	SDL_Event event;
 	do {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,6 +336,7 @@ set(libdevilutionx_SRCS
   Source/gendung.cpp
   Source/gmenu.cpp
   Source/help.cpp
+  Source/hwcursor.cpp
   Source/init.cpp
   Source/interfac.cpp
   Source/inv.cpp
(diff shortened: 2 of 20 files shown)


ephphatha:
Works fine on Windows 10 using VS 2021. Inventory items flicker/get drawn in the wrong position for a frame in master so I don't think that was introduced by this change.

julealgon:
> When clicking on inventory item, it briefly appears a bit shifted (in the wrong coordinates).

This is a very old, known issue, that exists (but is harder to see) in vanilla as well.

viciious:
> 2. Cursor switches back and forth between the default cursor while the palette is fading (Ubuntu X11).
Confirmed on windows 10 as well.

viciious:
In my implementation the cursor was hidden during the palette fades.

glebm:
Flickering issue fixed

glebm:
> This is a very old, known issue, that exists (but is harder to see) in vanilla as well.

Thanks for confirming. It is indeed very easy to see with the hardware cursor because it updates at the monitor refresh rate (144 Hz in my case so it's reproducible every time)

Source/DiabloUI/credits.cpp
```diff
@@ -199,6 +200,10 @@ bool TextDialog()
 	CreditsRenderer creditsRenderer;
 	bool endMenu = false;
 
+	bool wasCursorVisible;
+	if (IsHardwareCursorEnabled())
+		wasCursorVisible = SetHardwareCursorVisible(false);
```
AJenbo, line 205: How come we aren't able to use `ShouldShowCursor()` in situations like this?

glebm: This is the main menu code, `ShouldShowCursor` is game-specific, though it could probably be improved in the future.

AJenbo: Then I'm pretty sure this can simply be replaced with `!sgbControllerActive`

glebm: Yeah I guess, let me change that

glebm: Turned out to be even simpler than that and fixed some subtle bugs too. Done.

AJenbo (approved):

julealgon:
What's the reason for not having this as the default cursor again? Is there any advantage to the software-based cursor?

AJenbo:
As is the hardware cursor always uses NN sampling despite what the game is using, some systems appear to have issues with large cursors such as armors.

PR #2190, opened 2021-06-18
[switch] Switch on networking

AJenbo:


julealgon:
The '[switch] Switch on' gave me a hard time 🤣 Ultra minor, but would suggest renaming to "[Switch] Turn on networking"

AJenbo:
Expected it to crash and burn so didn't put much effort in to the title :)

PR #2196, opened 2021-06-19
fix descriptions for zerotier

qndel:


AJenbo (approved):
It would be better to track the connection name and use fmt to add it to a generic string.

AJenbo:
Replaced by https://github.com/diasurgical/devilutionX/pull/3102

PR #2201, opened 2021-06-20
Add tests for LCG random engine

ephphatha:
This documents the chosen LCG parameters and the default mapping function used in random number generation.

GetLCGEngineState() is a helper needed for testing because AdvanceRndSeed combines a distribution with a call to progress the engine. I'll split those eventually but this at least shows the current behaviour and should flag any bugs introduced by future changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -510,6 +510,7 @@ if(RUN_TESTS)
     test/missiles_test.cpp
     test/pack_test.cpp
     test/player_test.cpp
+    test/random_test.cpp
     test/scrollrt_test.cpp
     test/stores_test.cpp
     test/writehero_test.cpp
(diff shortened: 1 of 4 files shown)


AJenbo (approved):
Random 🙃

qndel:
tbh do we need extra tests for rng? If rng failed, item tests would fail

AJenbo:
It's nice to have a seperate test for the individual high value components. Even if items fail we don't know if it is broken RNG or broken `RecreateItem` code, with a test for the RNG code we can now easily spot where the failure may ly

ephphatha:
The current implementation relies on undefined behaviour `abs(int32::min)` and implementation defined behaviour `unsigned to signed conversion for values greater than int32::max`.

A C++11 compliant compiler can freely change that behaviour which would cause builds of devilutionx to be incompatible with the original game or builds with other compilers that happen to match expected behaviour. These tests immediately flag divergences.

I'll be opening a few more pull requests to expand the test suite and remove relying on undefined/implementation defined behaviour.

PR #2205, opened 2021-06-21
Expand test cases for Diablo RNG to cover UB/IB

ephphatha:
These tests cases cover all instances of undefined behaviour and implementation defined behaviour that I could see in the current `AdvanceRngSeed()` and `GenerateRnd(int)` functions. I've added/updated comments to refer to the relevant sections in the spec (based off the latest working drafts since I'm too cheap to pay for the official standards 😂 ).

### DIFF
--- a/test/random_test.cpp
+++ b/test/random_test.cpp
@@ -2,7 +2,7 @@
 
 #include "engine.h"
 
-using namespace devilution;
+namespace devilution {
 
 // These tests use ASSERT_EQ as the PRNG is expected to depend on state from the last call, so one failing assertion
 // means the rest of the results can't be trusted.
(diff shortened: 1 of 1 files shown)


test/random_test.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "engine.h"
 
-using namespace devilution;
+namespace devilution {
```
AJenbo, line 5: Why change the namespace compared to all other tests?

ephphatha: I mentioned this in the comment for that commit, but this is recommended for use with the GTest framework and required if you want to use [FRIEND_TEST](http://google.github.io/googletest/reference/testing.html#FRIEND_TEST)/test fixtures for classes with private members. Given the nature of an RNG it's likely that this would occur if a class is created for the RNG implementation.

AJenbo (approved):

AJenbo (dismissed):

AJenbo:
Feel free to amend previous commits when making corrections :)

PR #2209, opened 2021-06-21
Refactored "StartWalk(...)" 

matheusgomes28:
Refactored the `StartWalk(...)` function into different functions and templated some with the `Direction` enum.

As I've noticed, most of the params of the old `StartWalk(...)` functions were tightly coupled with the `Direction` parameter. For this reason, I added the `std::array<DirectionSettings, N>` to hold those tightly coupled dependencies and resolve them at compile time. Hopefully, we have gained some runtime efficiency there.

I understand I may not have named the functions correctly, so please feel free to suggest better names for them. These functions are:

- HandleWalkMode
- WalkDownwards
- WalkUpwards
- WalkSides

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -28,6 +28,187 @@
 
 namespace devilution {
 
+namespace {
+
+struct DirectionSettings {
+	Direction dir;
+	Point tileAdd;
+	Point offset;
+	Point map;
+	_scroll_direction scrollDir;
+	PLR_MODE walkMode;
+	void (*walkModeHandler)(int, Point, const DirectionSettings &);
+};
+
+void PM_ChangeLightOff(PlayerStruct &player)
+{
+	if (player._plid == NO_LIGHT)
+		return;
+
+	const LightListStruct *l = &LightList[player._plid];
+	int x = 2 * player.position.offset.y + player.position.offset.x;
+	int y = 2 * player.position.offset.y - player.position.offset.x;
+
+	x = (x / 8) * (x < 0 ? 1 : -1);
+	y = (y / 8) * (y < 0 ? 1 : -1);
+	int lx = x + (l->position.tile.x * 8);
+	int ly = y + (l->position.tile.y * 8);
+	int offx = l->position.offset.x + (l->position.tile.x * 8);
+	int offy = l->position.offset.y + (l->position.tile.y * 8);
+
+	if (abs(lx - offx) < 3 && abs(ly - offy) < 3)
+		return;
+
+	ChangeLightOff(player._plid, { x, y });
+}
+
+void WalkUpwards(int pnum, Point vel, const DirectionSettings &walkParams)
+{
+	auto &player = plr[pnum];
+	dPlayer[player.position.future.x][player.position.future.y] = -(pnum + 1);
+	player.position.temp = walkParams.tileAdd;
+}
+
+void WalkDownwards(int pnum, Point vel, const DirectionSettings &walkParams)
+{
+	auto &player = plr[pnum];
+	dPlayer[player.position.tile.x][player.position.tile.y] = -(pnum + 1);
+	player.position.temp = player.position.tile;
+	player.position.tile = player.position.future; // Move player to the next tile to maintain correct render order
+	dPlayer[player.position.tile.x][player.position.tile.y] = pnum + 1;
+	ChangeLightXY(player._plid, player.position.tile);
+	PM_ChangeLightOff(player);
+}
+
+void WalkSides(int pnum, Point vel, const DirectionSettings &walkParams)
+{
+	auto &player = plr[pnum];
+
+	Point const nextPosition = walkParams.map + player.position.tile;
+
+	dPlayer[player.position.tile.x][player.position.tile.y] = -(pnum + 1);
+	dPlayer[player.position.future.x][player.position.future.y] = -(pnum + 1);
+	player._pVar4 = nextPosition.x;
+	player._pVar5 = nextPosition.y;
+	dFlags[nextPosition.x][nextPosition.y] |= BFLAG_PLAYERLR;
+
+	if (leveltype != DTYPE_TOWN) {
+		ChangeLightXY(player._plid, nextPosition);
+		PM_ChangeLightOff(player);
+	}
+
+	player.position.temp = player.position.future;
+}
+
+static constexpr std::array<const DirectionSettings, 8> directionSettings { {
+	// clang-format off
+	{ DIR_S,  {  1,  1 }, {   0, -32 }, { 0, 0 }, SDIR_S,  PM_WALK2, WalkDownwards },
+	{ DIR_SW, {  0,  1 }, {  32, -16 }, { 0, 0 }, SDIR_SW, PM_WALK2, WalkDownwards },
+	{ DIR_W,  { -1,  1 }, {  32, -16 }, { 0, 1 }, SDIR_W,  PM_WALK3, WalkSides     },
+	{ DIR_NW, { -1,  0 }, {   0,   0 }, { 0, 0 }, SDIR_NW, PM_WALK,  WalkUpwards   },
+	{ DIR_N,  { -1, -1 }, {   0,   0 }, { 0, 0 }, SDIR_N,  PM_WALK,  WalkUpwards   },
+	{ DIR_NE, {  0, -1 }, {   0,   0 }, { 0, 0 }, SDIR_NE, PM_WALK,  WalkUpwards   },
+	{ DIR_E,  {  1, -1 }, { -32, -16 }, { 1, 0 }, SDIR_E,  PM_WALK3, WalkSides     },
+	{ DIR_SE, {  1,  0 }, { -32, -16 }, { 0, 0 }, SDIR_SE, PM_WALK2, WalkDownwards }
+	// clang-format on
+} };
+
+void ScrollViewPort(const PlayerStruct &player, _scroll_direction dir)
+{
+	ScrollInfo.tile.x = player.position.tile.x - ViewX;
+	ScrollInfo.tile.y = player.position.tile.y - ViewY;
+
+	if (zoomflag) {
+		if (abs(ScrollInfo.tile.x) >= 3 || abs(ScrollInfo.tile.y) >= 3) {
+			ScrollInfo._sdir = SDIR_NONE;
+		} else {
+			ScrollInfo._sdir = dir;
+		}
+	} else if (abs(ScrollInfo.tile.x) >= 2 || abs(ScrollInfo.tile.y) >= 2) {
+		ScrollInfo._sdir = SDIR_NONE;
+	} else {
+		ScrollInfo._sdir = dir;
+	}
+}
+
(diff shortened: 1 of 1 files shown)


AJenbo:
Linked issue: https://github.com/diasurgical/devilutionX/issues/1998

> * HandleWalkMode

`ScrollViewport`

> * WalkDownwards
> * WalkUpwards
> * WalkSides

Use cardinal cordinates: WalkSouth(ern), WalkNorth(ern), WalkHorizontal

Source/player.cpp
```diff
@@ -26,6 +26,196 @@
 #include "utils/language.h"
 #include "utils/log.hpp"
 
+namespace {
+using namespace devilution;
```
glebm, line 30: Simply place `namespace {` inside `namespace devilution {` instead

Source/player.cpp
```diff
@@ -26,6 +26,196 @@
...
+	}
+
+	if (dir == DIR_W) {
+		return !SolidLoc(position + DIR_SW) && !(dFlags[position.x][position.y + 1] & BFLAG_PLAYERLR);
+	}
+
+	return true;
+}
+
+template <Direction dir>
```
glebm, line 163: I don't think it makes sense to lift this to a type parameter here. This function is not a hotspot so increased code size is likely to decrease performance. Simply pass as argument.

Source/player.cpp
```diff
@@ -26,6 +26,196 @@
...
+		skippedFrames = 2;
+	if (pmWillBeCalled)
+		skippedFrames += 1;
+	NewPlrAnim(player, player_graphic::Walk, dir, player._pWFrames, 0, AnimationDistributionFlags::ProcessAnimationPending, skippedFrames);
+}
+
+/**
+ * @brief Start moving a player to a new tile
+ */
+template <Direction Dir>
```
glebm, line 202: Same, lower from type to argument

matheusgomes28 (commented):
I still have some [suggestions to address in my fork's PR here](https://github.com/matheusgomes28/devilutionX/pull/2). Please don't merge until I've deleted this comment (apologies for the inconvenience).

Source/player.cpp
glebm: >  this code actually produces a smaller executable (by a few bytes) than the master branch.

This could be due to other changes in this code, please measure just the argument vs non-type template parameter change.

> We just won't be able to make proper use of the `constexpr` array data.

We'll still be able to use it, it'll just be stored contiguously instead of an element in each generated function.

matheusgomes28: > This could be due to other changes in this code, please measure just the argument vs non-type template parameter change.

Indeed, that's what I was trying to say. If the binary is bigger it could also be due to other changes (I've added a few functions, so if any of them didn't get properly optimised then the binary would end up bigger as well). I'll make the change and measure the results this afternoon after work, I'll let you know the results!

> We'll still be able to use it, it'll just be stored contiguously instead of an element in each generated function.
Agreed, it's currently contiguously stored in the data section of the binary too, but there's no need for runtime access with the template. I doubt that the runtime access of that array would "choke" the runtime, I just thought it would be a nice little optimistion. I can change that if the parameter implementation proves to be better :smile:

AJenbo: But also keep in min that what you are doing here is improve code readability :)

Source/player.cpp
matheusgomes28: Sorted

Source/player.cpp
```diff
@@ -83,17 +84,16 @@ void WalkSides(int pnum, Point vel, DirectionSettings const &walkParams)
 {
 	auto &player = plr[pnum];
 
-	int const x = walkParams.map.x + player.position.tile.x;
-	int const y = walkParams.map.y + player.position.tile.y;
+	Point const nextPosition = walkParams.map + player.position.tile;
```
AJenbo, line 87: Don't think we have const before the type any where else.
```suggestion
	const Point nextPosition = walkParams.map + player.position.tile;
```

Source/player.cpp
```diff
@@ -28,6 +28,189 @@
...
+			ScrollInfo._sdir = SDIR_NONE;
+		} else {
+			ScrollInfo._sdir = dir;
+		}
+	} else if (abs(ScrollInfo.tile.x) >= 2 || abs(ScrollInfo.tile.y) >= 2) {
+		ScrollInfo._sdir = SDIR_NONE;
+	} else {
+		ScrollInfo._sdir = dir;
+	}
+}
```
AJenbo, line 134: Looking at this again I think it would be better to rewrite it as:
```suggestion
void ScrollViewPort(_scroll_direction dir)
{
	auto const &myPlayer = plr[myplr];

	ScrollInfo.tile.x = myPlayer.position.tile.x - ViewX;
	ScrollInfo.tile.y = myPlayer.position.tile.y - ViewY;

	int distance = 2;
	if (zoomflag) {
        distance++;
	}
	ScrollInfo._sdir = SDIR_NONE;
	if (abs(ScrollInfo.tile.x) < distance && abs(ScrollInfo.tile.y) < distance) {
		ScrollInfo._sdir = dir;
	}
}
```

And then change the call site to:
```cpp
if (pnum == myplr) {
    ScrollViewPort(directionSettings[Dir].scrollDir);
}
```

Source/player.cpp
```diff
@@ -28,6 +28,189 @@
...
+
+static constexpr std::array<DirectionSettings const, 8> directionSettings { { { DIR_S, { 1, 1 }, { 0, -32 }, { 0, 0 }, SDIR_S, PM_WALK2, WalkDownwards },
+	{ DIR_SW, { 0, 1 }, { 32, -16 }, { 0, 0 }, SDIR_SW, PM_WALK2, WalkDownwards },
+	{ DIR_W, { -1, 1 }, { 32, -16 }, { 0, 1 }, SDIR_W, PM_WALK3, WalkSides },
+	{ DIR_NW, { -1, 0 }, { 0, 0 }, { 0, 0 }, SDIR_NW, PM_WALK, WalkUpwards },
+	{ DIR_N, { -1, -1 }, { 0, 0 }, { 0, 0 }, SDIR_N, PM_WALK, WalkUpwards },
+	{ DIR_NE, { 0, -1 }, { 0, 0 }, { 0, 0 }, SDIR_NE, PM_WALK, WalkUpwards },
+	{ DIR_E, { 1, -1 }, { -32, -16 }, { 1, 0 }, SDIR_E, PM_WALK3, WalkSides },
+	{ DIR_SE, { 1, 0 }, { -32, -16 }, { 0, 0 }, SDIR_SE, PM_WALK2, WalkDownwards } } };
```
AJenbo, line 110: Seam Clang is struggling with this one, might as well give it the full manual treatment:
```suggestion
static constexpr std::array<DirectionSettings const, 8> directionSettings { {
	// clang-format off
	{ DIR_S,  {  1,  1 }, {   0, -32 }, { 0, 0 }, SDIR_S,  PM_WALK2, WalkDownwards },
	{ DIR_SW, {  0,  1 }, {  32, -16 }, { 0, 0 }, SDIR_SW, PM_WALK2, WalkDownwards },
	{ DIR_W,  { -1,  1 }, {  32, -16 }, { 0, 1 }, SDIR_W,  PM_WALK3, WalkSides },
	{ DIR_NW, { -1,  0 }, {   0,   0 }, { 0, 0 }, SDIR_NW, PM_WALK,  WalkUpwards },
	{ DIR_N,  { -1, -1 }, {   0,   0 }, { 0, 0 }, SDIR_N,  PM_WALK,  WalkUpwards },
	{ DIR_NE, {  0, -1 }, {   0,   0 }, { 0, 0 }, SDIR_NE, PM_WALK,  WalkUpwards },
	{ DIR_E,  {  1, -1 }, { -32, -16 }, { 1, 0 }, SDIR_E,  PM_WALK3, WalkSides },
	{ DIR_SE, {  1,  0 }, { -32, -16 }, { 0, 0 }, SDIR_SE, PM_WALK2, WalkDownwards }
	// clang-format on
} };
```

Source/player.cpp
matheusgomes28: @AJenbo @glebm Apologies for the delay. I did some building yesterday and here are the results.

```
GCC:
	Templated:   4,172,472
	No Template: 4,172,504
 
Clang (Release O2):
	Template:    7,178,408
	No Template: 7,174,328
 
Clang (Release Os):
	Template:    7,178,408
	No Template: 7,174,328
 
Clang (Release Oz):
	Template:    7,178,408
	No Template: 7,174,328
 
MSVC (Release):
	Template:    2,987,008
	No template: 2,985,472
```

These results were taken from the branch in this PR and "start-walk-refactor-no-template" branch I've created that removes the params. As we see, GCC probably does a better job at figuring out the common bits of the template functions and perhaps not inlining them when they are called, so the size is a little smaller. Cannot say the same for MSVC and Clang.

Although binary size is not a clear indicator of how efficient your code is, it seems like smaller size is preferred here so I'll remove the templates.

Also, I have no idea why Clang bloats the executable so much, to the point that the executable is much larger than the ones generated by GCC and Windows. Hence the reason why I tried different optimisation flags. I guess if you really care about binary size, don't compile DevilutionX with Clang :smile:

Source/player.cpp
matheusgomes28: My bad, I'm an `east const` user. I'll change them to `west cont`s, but we might wanna consider `east const` for[ its benefits](https://isocpp.org/wiki/faq/const-correctness#const-ref-alt).

Source/player.cpp
matheusgomes28: Sorted

Source/player.cpp
matheusgomes28: Noticed that too, it made it borderline unreadable! Fixed.

Source/player.cpp
matheusgomes28: I think I've sorted all of the consts out.

Source/player.cpp
matheusgomes28: @glebm, no more templates. Let me know if there's anything else.

AJenbo: Thanks, really interesting. The GCC vs MSVC size is also interesting, but I'm guessing that is down to extra dependencies that comes with MSVC.

AJenbo:
@matheusgomes28 nice work, i picked out another function that need a similar treatment, https://github.com/diasurgical/devilutionX/issues/2211 if you are interested, or you could move thing in player.cpp to the anon name spaces like we talked about :)

AJenbo:
P.s. if you write `Resolves #1998` in the PR message then the issue will acutomatically get closed when merged.

matheusgomes28:
@AJenbo I'll pick #2211 this week! My bad, didn't realise I didn't tag the issue.

glebm:
@matheusgomes28 Thanks for the investigation with the template. Smaller size is preferred here because this function is not a hotspot at all. We do use the template approach in hotspots, for example in https://github.com/diasurgical/devilutionX/tree/master/Source/engine/render

julealgon:
> My bad, didn't realise I didn't tag the issue.

@matheusgomes28 , you can still do it, just edit the PR description.

PR #2213, opened 2021-06-22
🔥 Remove redundant MI_Lightctrl code

AJenbo:


Source/missiles.cpp
```diff
@@ -3992,7 +3990,7 @@ void MI_Lightctrl(int i)
 		missile[i]._miVar1 = missile[i].position.tile.x;
 		missile[i]._miVar2 = missile[i].position.tile.y;
 	}
-	if (missile[i]._mirange == 0 || mx == 0 || my == 0) {
+	if (missile[i]._mirange == 0) {
```
AJenbo, line 3993: levels always have a 16 tile border so it shouldn't be able to reach 0,0 to begin with, not that this was original checking for the full map, but was reduced in https://github.com/diasurgical/devilutionX/pull/2115 as most of the check was logically impossible as well.

ephphatha: Could you add asserts at line 3942 checking that mx/my are >= 0? This will match the existing checks for missiles outside the dungeon upper limits and they get compiled out of non debug builds anyway.

Source/missiles.cpp
```diff
@@ -3944,12 +3944,10 @@ void MI_Lightctrl(int i)
 	int pn = dPiece[mx][my];
 	assert((DWORD)pn <= MAXTILES);
 
-	if (missile[i]._misource == -1) {
-		if ((Point { mx, my } != missile[i].position.start) && nMissileTable[pn]) {
+	if (missile[i]._misource != -1 || Point { mx, my } != missile[i].position.start) {
```
StephenCWills, line 3947: Curious that `int p = missile[i]._misource;` is used in the damage calculation above, but `p` is never used again in the whole function.

StephenCWills (dismissed):

Source/missiles.cpp
AJenbo: Done

Source/missiles.cpp
AJenbo: Fixed

PR #2215, opened 2021-06-23
♻️ Clean up `cel_header.hpp`

glebm:
1. Rename `CelGetFrameStart` to `CelGetFrame`, in line with the other 2 functions with the same name and load the `uint32_t` safely.
2. Remove redundant `FrameHeader`, simply use `LoadLE16`.
3. Document all the functions.

### DIFF
--- a/Source/utils/endian.hpp
+++ b/Source/utils/endian.hpp
@@ -5,6 +5,14 @@
 namespace devilution {
 
 template <typename T>
+constexpr std::uint16_t LoadLE16(const T *b)
+{
+	static_assert(sizeof(T) == 1, "invalid argument");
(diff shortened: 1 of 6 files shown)


Source/utils/endian.hpp
```diff
@@ -4,6 +4,14 @@
 
 namespace devilution {
 
+template <typename T>
+constexpr std::uint16_t LoadLE16(const T *b)
```
AJenbo, line 8: Could I request a `LoadBE16` be added and used for in `NetRecvPlrData`?

AJenbo (approved):

Source/utils/endian.hpp
glebm: Sure but that's for a separate PR

PR #2222, opened 2021-06-24
Cleanups

AJenbo:
About ½ of the diff is indentation so defiantly ignore white space on this one.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -315,11 +315,9 @@ static void DrawSpell(const CelOutputBuffer &out)
 	if (currlevel == 0 && st != RSPLTYPE_INVALID && !spelldata[spl].sTownSpell)
 		st = RSPLTYPE_INVALID;
 	SetSpellTrans(st);
+	const int nCel = (spl != SPL_INVALID) ? SpellITbl[spl] : 27;
 	const Point position { PANEL_X + 565, PANEL_Y + 119 };
-	if (spl != SPL_INVALID)
-		DrawSpellCel(out, position, *pSpellCels, SpellITbl[spl]);
-	else
-		DrawSpellCel(out, position, *pSpellCels, 27);
+	DrawSpellCel(out, position, *pSpellCels, nCel);
 }
 
 static void PrintSBookHotkey(const CelOutputBuffer &out, Point position, const std::string &text)
--- a/Source/spells.cpp
+++ b/Source/spells.cpp
@@ -154,29 +154,26 @@ void EnsureValidReadiedSpell(PlayerStruct &player)
 
 bool CheckSpell(int id, spell_id sn, spell_type st, bool manaonly)
 {
-	bool result;
-
 #ifdef _DEBUG
 	if (debug_mode_key_inverted_v)
 		return true;
 #endif
 
-	result = true;
 	if (!manaonly && pcurs != CURSOR_HAND) {
-		result = false;
-	} else {
-		if (st != RSPLTYPE_SKILL) {
-			if (GetSpellLevel(id, sn) <= 0) {
-				result = false;
-			} else {
-				auto &player = plr[id];
-
-				return player._pMana >= GetManaAmount(player, sn);
-			}
-		}
+		return false;
 	}
 
-	return result;
+	if (st == RSPLTYPE_SKILL) {
+		return true;
+	}
+
+	if (GetSpellLevel(id, sn) <= 0) {
+		return false;
+	}
+
+	auto &player = plr[id];
+
+	return player._pMana >= GetManaAmount(player, sn);
 }
 
 void CastSpell(int id, int spl, int sx, int sy, int dx, int dy, int spllvl)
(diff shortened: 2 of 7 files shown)


Source/spells.cpp
```diff
@@ -296,46 +293,37 @@ void DoResurrect(int pnum, int rid)
...
-		if (player._pClass == HeroClass::Warrior || player._pClass == HeroClass::Barbarian) {
-			hp *= 2;
-		} else if (player._pClass == HeroClass::Rogue || player._pClass == HeroClass::Bard) {
-			hp += hp / 2;
-		} else if (player._pClass == HeroClass::Monk) {
-			hp *= 3;
-		}
-
-		target._pHitPoints += hp;
+	if (rid == -1 || (target._pHitPoints >> 6) <= 0) {
```
StephenCWills, line 303: This function is called from `On_HEALOTHER` which passes in a `uint16_t` as the `rid` parameter. This means `rid` should never be -1, but AFAIK it could be after casting to `char`. Either this change introduces potential errors or it's entirely unnecessary.

StephenCWills (changes requested):
I just tested Heal Other. `TryOnCurs()` passes `pcursplr` which has an `int8_t` value of -1 if no player is targeted. `NetSendCmdParam1()` accepts that value as `uint16_t` which causes an implicit conversion to 65535. When the packet is processed, `On_HEALOTHER()` receives that value of 65535 which then gets passed into `DoHealOther()` as-is. So the value does need to be cast to `char` before the comparison to -1.

Source/spells.cpp
AJenbo: The signature casts it to an `int`, isn't that enough?

Source/control.cpp
```diff
@@ -315,11 +315,9 @@ static void DrawSpell(const CelOutputBuffer &out)
 	if (currlevel == 0 && st != RSPLTYPE_INVALID && !spelldata[spl].sTownSpell)
 		st = RSPLTYPE_INVALID;
 	SetSpellTrans(st);
+	int nCel = (spl != SPL_INVALID) ? SpellITbl[spl] : 27;
```
glebm, line 318:
```suggestion
	const int nCel = (spl != SPL_INVALID) ? SpellITbl[spl] : 27;
```

Source/spells.cpp
ephphatha: `DoHealOther` should take `int pnum, uint rid`. Converting from `uint16_t` to `int` is implementation defined. The usual implementation for x86 is simply sign-extending the MSB and reinterpreting as the target type, so this will preserve -1 (really `std::numeric_limits<uint16_t>::max()` in this context) as -1.

Assuming you change the function signature you then avoid IB. The cast to char is unnecessary in both cases. Regarding the comparison with -1, conversion of a signed value to an unsigned value is well defined but you may get warnings from your compiler if it's implicit. If so, wrapping -1 in a cast to uint16_t would supress it but the better solution would be to define a constant along the lines of `constexpr uint16_t NO_TARGET = -1`

Source/spells.cpp
```diff
@@ -296,46 +293,37 @@ void DoResurrect(int pnum, int rid)
...
 void DoHealOther(int pnum, int rid)
 {
-	int i, j, hp;
-
 	if (pnum == myplr) {
 		NewCursor(CURSOR_HAND);
 	}
 
 	auto &player = plr[pnum];
 	auto &target = plr[rid];
```
ephphatha, line 301: `rid` should be checked before using it to index into the `plr` array. This causes OOB reads if an invalid network packet is received.

ephphatha (changes requested):

Source/spells.cpp
AJenbo: Doing so would also get rid of the `uint16 -1` madness.

Source/spells.cpp
StephenCWills: @ephphatha FYI, I did my testing using MSVC. I thought you might be curious.

Source/spells.cpp
```diff
@@ -294,48 +291,46 @@ void DoResurrect(int pnum, int rid)
...
+void DoHealOther(int pnum, uint16_t rid)
 {
-	int i, j, hp;
-
 	if (pnum == myplr) {
 		NewCursor(CURSOR_HAND);
 	}
 
+	if ((DWORD)pnum >= MAX_PLRS) {
+		app_fatal("CreatePlayer: illegal player %i", pnum);
```
StephenCWills, line 301: This should be `DoHealOther` instead of `CreatePlayer`. Having said that, I'm not sure that crashing is an appropriate response to receiving a bad packet. Seems like that could invite some foul play.

Source/missiles.cpp
```diff
@@ -52,12 +52,8 @@ void GetDamageAmt(int i, int *mind, int *maxd)
 			*mind += *mind / 2;
 		}
 		*maxd = 10;
-		for (int k = 0; k < myPlayer._pLevel; k++) {
-			*maxd += 4;
-		}
-		for (int k = 0; k < sl; k++) {
-			*maxd += 6;
-		}
+		*maxd += 4 * myPlayer._pLevel;
```
StephenCWills, line 55: These changes come from your other PR, right? Did you mean to include these here?

AJenbo: fixed

StephenCWills (approved):

PR #2223, opened 2021-06-24
Cleanup missiles

AJenbo:
About 1/3 of the changes are simply indentation so might want to play with the "ignore white space" option, also i split the most intrusive changes in to a second commit.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -3810,40 +3810,53 @@ void MI_Search(int i)
 	AutoMapShowItems = false;
 }
 
+bool GrowWall(int playerId, Point position, Point target, missile_id type, int spellLevel, int damage)
+{
+	int dp = dPiece[position.x][position.y];
+	assert(dp <= MAXTILES && dp >= 0);
+
+	if (nMissileTable[dp] || !InDungeonBounds(target)) {
+		return false;
+	}
+
+	AddMissile(position, position, plr[playerId]._pdir, type, TARGET_BOTH, playerId, damage, spellLevel);
+	return true;
+}
+
 void MI_LightningWallC(int i)
 {
-	int tx, ty;
-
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -562,7 +552,7 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, int t, bool
 	if (monster[m]._mmode == MM_CHARGE)
 		return false;
 
-	mor = monster[m].mMagicRes;
+	uint8_t mor = monster[m].mMagicRes;
```
StephenCWills, line 555: Isn't `mMagicRes` a uint16_t?

Source/missiles.cpp
```diff
@@ -1552,7 +1545,7 @@ void AddStealPotions(int mi, Point src, Point dst, int midir, int8_t mienemy, in
 					SetPlrHandItem(&player.HoldItem, ii);
 					GetPlrHandSeed(&player.HoldItem);
 					player.HoldItem._iStatFlag = true;
-					player.SpdList[si] = plr[pnum].HoldItem;
+					player.SpdList[si] = player.HoldItem;
```
StephenCWills, line 1548: Nice catch. 👍

StephenCWills (changes requested):
I only had the one question about `mMagicRes`. Everything else looks good to me.

Source/missiles.cpp
AJenbo:
```
MonsterStruct: uint16_t
UniqMonstStruct: uint16_t
MonsterData: uint8_t
```

It represents `monster_resistance` (`uint8_t`) as a bitflag.

ephphatha (approved):

StephenCWills (approved):

PR #2224, opened 2021-06-24
♻️Clean up items

AJenbo:


Source/items.cpp
```diff
@@ -3889,14 +3889,13 @@ void UseItem(int p, item_misc_id Mid, spell_id spl)
 	switch (Mid) {
 	case IMISC_HEAL:
 	case IMISC_FOOD:
-		j = player._pMaxHP >> 8;
-		l = ((j >> 1) + GenerateRnd(j)) << 6;
+		l = GenerateRnd(player._pMaxHP / 4) + player._pMaxHP / 8;
```
ephphatha, line 3892: Is this intended to be included in this PR? This makes heals more granular, previously the low bits 5 bits would always be 0. It makes better use of the fixed point math anyway.

Source/items.cpp
```diff
@@ -4247,9 +4243,9 @@ static void SpawnOnePremium(int i, int plvl, int myplr)
 	int strength = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Strength), myPlayer._pStrength);
 	int dexterity = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Dexterity), myPlayer._pDexterity);
 	int magic = std::max(myPlayer.GetMaximumAttributeValue(CharacterAttribute::Magic), myPlayer._pMagic);
-	strength *= 1.2;
-	dexterity *= 1.2;
-	magic *= 1.2;
+	strength += strength / 5;
```
ephphatha, line 4246: Was there a particular reason for changing this from float multiplication to int division? I don't think int division causes any loss of precision, mainly commenting as I'm curious.

AJenbo: The result is the same, but it avoid floats which isn't ideal on some systems. It was done after a comment by @thebigMuh https://github.com/diasurgical/devilutionX/pull/2055#discussion_r639114212

Source/items.cpp
AJenbo: https://github.com/diasurgical/devilutionX/pull/2055#discussion_r639110686
Good point about the 5 bits, the compiler should be fully capable of converting it to shifts though.

ephphatha: Yeah, not worried about the divisions here. I *think* player._pMaxHP is always a value less than `2^15  - 1`, so using the current RNG with a higher value isn't going to alter the behaviour much.

ephphatha (approved):

PR #2225, opened 2021-06-24
♻️Simplify loops in to mulitplications

AJenbo:


Source/missiles.cpp
```diff
@@ -166,13 +162,9 @@ void GetDamageAmt(int i, int *mind, int *maxd)
 		break;
 	case SPL_APOCA:
 		*mind = 0;
-		for (int k = 0; k < myPlayer._pLevel; k++) {
-			*mind += 1;
-		}
+		*mind += myPlayer._pLevel;
```
StephenCWills, line 165: Instead of initializing to zero and adding, should we just set the value?

StephenCWills (changes requested):
I suppose it doesn't just apply to Apocalypse. We can consider consolidating all of these to look more like a damage calculation formula. Honestly, I don't have a particular preference so use your judgment. However, I do think there's no sense initializing to zero and then using `+=`.

AJenbo:
> consider consolidating all of these to look more like a damage calculation formula

Gave all the calculations and overhaul, hope it's to your liking :)

Source/missiles.cpp
```diff
@@ -42,40 +74,21 @@ void GetDamageAmt(int i, int *mind, int *maxd)
...
-		} else if (myPlayer._pClass == HeroClass::Rogue || myPlayer._pClass == HeroClass::Bard) {
-			*maxd += *maxd / 2;
-		}
-		*mind = -1;
-		*maxd = -1;
+	case SPL_HEAL:
+	case SPL_HEALOTHER:
+		/// BUGFIX: healing calculation is unused
+		*mind += AddClassHealingBonus(myPlayer._pLevel + sl + 1, myPlayer._pClass) - 1;
+		*maxd += AddClassHealingBonus((4 * myPlayer._pLevel) + (6 * sl) + 10, myPlayer._pClass) - 1;
```
StephenCWills, line 83:
```suggestion
		*mind = AddClassHealingBonus(myPlayer._pLevel + sl + 1, myPlayer._pClass) - 1;
		*maxd = AddClassHealingBonus((4 * myPlayer._pLevel) + (6 * sl) + 10, myPlayer._pClass) - 1;
```

StephenCWills (approved):
Fantastic work! It looks great. 😄

Source/missiles.cpp
```diff
@@ -30,6 +30,38 @@ int numchains;
...
+	if (heroClass == HeroClass::Warrior || heroClass == HeroClass::Monk || heroClass == HeroClass::Barbarian) {
+		return hp * 2;
+	}
+
+	if (heroClass == HeroClass::Rogue || heroClass == HeroClass::Bard) {
+		return hp + hp / 2;
+	}
+
+	return hp;
+}
```
julealgon, line 44: This method would be vastly more readable with a switch IMHO.

Source/missiles.cpp
```diff
@@ -30,6 +30,38 @@ int numchains;
...
 const int CrawlNum[19] = { 0, 3, 12, 45, 94, 159, 240, 337, 450, 579, 724, 885, 1062, 1255, 1464, 1689, 1930, 2187, 2460 };
 
+int AddClassHealingBonus(int hp, HeroClass heroClass)
+{
+	if (heroClass == HeroClass::Warrior || heroClass == HeroClass::Monk || heroClass == HeroClass::Barbarian) {
+		return hp * 2;
+	}
+
+	if (heroClass == HeroClass::Rogue || heroClass == HeroClass::Bard) {
+		return hp + hp / 2;
```
julealgon, line 40: I'd suggest using `hp * 1.5` here so that all statements look similar.

PR #2229, opened 2021-06-26
♻️Clean up implicit bool conversions

AJenbo:
Sorry for the boring parts of this, mostly just generated by clang-tidy, but I also did some manual changes to satisfy it in a way that didn't turn in to a casting nightmare and a kaleidoscope of parentheses. The most interesting part is finally splitting ini settings in to options.cpp

### DIFF
--- a/Source/DiabloUI/selgame.cpp
+++ b/Source/DiabloUI/selgame.cpp
@@ -447,9 +447,9 @@ void selgame_Password_Select(int value)
 
 	m_game_data->nDifficulty = nDifficulty;
 	m_game_data->nTickRate = nTickRate;
-	m_game_data->bRunInTown = sgOptions.Gameplay.bRunInTown;
-	m_game_data->bTheoQuest = sgOptions.Gameplay.bTheoQuest;
-	m_game_data->bCowQuest = sgOptions.Gameplay.bCowQuest;
+	m_game_data->bRunInTown = sgOptions.Gameplay.bRunInTown ? 1 : 0;
+	m_game_data->bTheoQuest = sgOptions.Gameplay.bTheoQuest ? 1 : 0;
+	m_game_data->bCowQuest = sgOptions.Gameplay.bCowQuest ? 1 : 0;
 
 	if (SNetCreateGame(nullptr, selgame_Password, (char *)m_game_data, sizeof(*m_game_data), gdwPlayerId)) {
 		UiInitList_clear();
--- a/Source/cursor.cpp
+++ b/Source/cursor.cpp
@@ -334,11 +334,11 @@ void CheckCursMove()
 	ShiftGrid(&mx, &my, -columns / 2, -lrow / 2);
 
 	// Align grid
-	if ((columns & 1) == 0 && (lrow & 1) == 0) {
+	if ((columns % 2) == 0 && (lrow % 2) == 0) {
 		sy += TILE_HEIGHT / 2;
-	} else if (columns & 1 && lrow & 1) {
+	} else if ((columns % 2) != 0 && (lrow % 2) != 0) {
 		sx -= TILE_WIDTH / 2;
-	} else if (columns & 1 && (lrow & 1) == 0) {
+	} else if ((columns % 2) != 0 && (lrow % 2) == 0) {
 		my++;
 	}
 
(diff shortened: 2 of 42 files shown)


Source/objects.cpp
```diff
@@ -4277,7 +4277,7 @@ bool OperateShrineMurphys(int pnum)
 	for (auto &item : plr[myplr].InvBody) {
 		if (!item.isEmpty() && GenerateRnd(3) == 0) {
 			if (item._iDurability != DUR_INDESTRUCTIBLE) {
-				if (item._iDurability) {
+				if (item._iDurability != 0) {
```
ephphatha, line 4280:
```suggestion
				if (item._iDurability > 0) {
```

Source/pack.cpp
```diff
@@ -106,12 +106,12 @@ void PackPlayer(PkPlayerStruct *pPack, const PlayerStruct &player, bool manashie
 	pPack->pDifficulty = SDL_SwapLE32(player.pDifficulty);
 	pPack->pDamAcFlags = SDL_SwapLE32(player.pDamAcFlags);
 	pPack->pDiabloKillLevel = SDL_SwapLE32(player.pDiabloKillLevel);
-	pPack->bIsHellfire = gbIsHellfire;
+	pPack->bIsHellfire = gbIsHellfire ? 1 : 0;
```
ephphatha, line 109: `bIsHellfire` is always treated as a bool, but due to the way `PkPlayerStruct` is used it needs to have a known width type. Adding a comment to `PkPlayerStruct::bIsHellfire` documenting this would be great.

Source/palette.cpp
```diff
@@ -212,7 +212,7 @@ void DecreaseGamma()
 
 int UpdateGamma(int gamma)
 {
-	if (gamma) {
+	if (gamma != 0) {
```
ephphatha, line 215:
```suggestion
	if (gamma > 0) {
```
This function is called with gamma having the value 0 or something in the range [30, 100]

Source/path.cpp
```diff
@@ -329,7 +327,7 @@ void path_set_coords(PATHNODE *pPath)
 	int i;
 
 	path_push_active_step(pPath);
-	while (gdwCurPathStep) {
+	while (gdwCurPathStep != 0) {
```
ephphatha, line 330:
```suggestion
	// while there are path nodes to check
	while (gdwCurPathStep > 0) {
```
`gdwCurPathStep` is used as an index into `pnode_tblptr` so must be positive to ensure `path_pop_active_step()` doesn't start underrunning the buffer.

Source/quests.cpp
```diff
@@ -803,7 +803,7 @@ void QuestlogESC()
 	int y, i;
 
 	y = (MouseY - 32) / 12;
-	if (numqlines) {
+	if (numqlines != 0) {
```
ephphatha, line 806: This check is unnecessary, `for(...; i < numqlines; ...)` already covers it.

ephphatha (changes requested):
The WebKit style guide calls out equaility checks with 0 as against convention. I actually prefer quality checks for flag variables, I mention that purely to give context for why I've suggested doing `> 0` comparisons for counters.

There are a few spots where it looks like you're doing unnecessary bool to int conversions using `? 1 : 0`, I'm not sure how clang flags those in it's analysis but it stood out to me from other projects I work on which flag unnecessary conversion of that form. Feel free to ignore those comments.

AJenbo:
> There are a few spots where it looks like you're doing unnecessary bool to int conversions using `? 1 : 0`, I'm not sure how clang flags those in it's analysis

clang-tidy will change: `uint16_t transport = mybool;` to `uint16_t transport = static_cast<int>(mybool);` which i find very unpalatable so I instead went for `uint16_t transport = mybool ? 1 : 0;`

Source/items.h
```diff
@@ -218,7 +218,7 @@ struct ItemStruct {
 	int16_t _iPLGetHit;
 	int16_t _iPLLight;
 	int8_t _iSplLvlAdd;
-	int8_t _iRequest;
+	bool _iRequest;
```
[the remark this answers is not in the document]
AJenbo: True, though I was thinking of delaying that for a later major renaming

Source/diablo.h
```diff
@@ -9,6 +9,7 @@
 
 #include "utils/endian.hpp"
 
+#include "controls/keymapper.hpp"
```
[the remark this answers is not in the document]
AJenbo: Maybe, but we don't really use forward declare in the codebase

Source/diablo.cpp
```diff
@@ -157,23 +136,20 @@ int arrowdebug = 0;
 #endif
 /** Specifies whether players are in non-PvP mode. */
 bool gbFriendlyMode = true;
-/** Default quick messages */
-const char *const spszMsgTbl[] = {
+const char *const spszMsgTbl[4] = {
```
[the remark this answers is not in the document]
AJenbo: It would also be nice to merge the key and value in to some sort of struct, maybe even consider something that helps it integrate easy with the keymapper but that will have to wait for a later refactoring.

Source/DiabloUI/selgame.cpp
```diff
@@ -447,9 +447,9 @@ void selgame_Password_Select(int value)
 
 	m_game_data->nDifficulty = nDifficulty;
 	m_game_data->nTickRate = nTickRate;
-	m_game_data->bRunInTown = sgOptions.Gameplay.bRunInTown;
-	m_game_data->bTheoQuest = sgOptions.Gameplay.bTheoQuest;
-	m_game_data->bCowQuest = sgOptions.Gameplay.bCowQuest;
+	m_game_data->bRunInTown = sgOptions.Gameplay.bRunInTown ? 1 : 0;
```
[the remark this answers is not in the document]
AJenbo: Given that `static_cast<uint8_t>()` is a fairly unique concept to C++ I consider it to not be very readable for new comers to the language/code base and so prefer to avoid it when more generic/shorter solutions are available. I suppose ` ? 1 : 0` could be a bit more prone to typos but I think it's acceptable.

ephphatha:
Understandable. I prefer the static_cast version myself but happy to stick to convention.

AJenbo:
> I've suggested doing `> 0` comparisons for counters

Nice, I don't think I had to mental capacity to do so when I went over things :D

Source/path.cpp
AJenbo: This code is a real horror show of what can be done with globals :)

Source/nthread.cpp
```diff
@@ -80,7 +80,7 @@ bool nthread_recv_turns(bool *pfSendAsync)
 {
 	*pfSendAsync = false;
 	sgbPacketCountdown--;
-	if (sgbPacketCountdown) {
+	if (sgbPacketCountdown != 0) {
```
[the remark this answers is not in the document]
AJenbo: The name "Countdown" is also a good indication of this :)

Source/diablo.cpp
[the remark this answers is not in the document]
AJenbo: Did both :D

Source/effects.cpp
```diff
@@ -1170,10 +1170,10 @@ static void PlaySFX_priv(TSFX *pSFX, bool loc, int x, int y)
 {
 	int lPan, lVolume;
 
-	if (plr[myplr].pLvlLoad && gbIsMultiplayer) {
+	if (plr[myplr].pLvlLoad != 0 && gbIsMultiplayer) {
 		return;
 	}
-	if (!gbSndInited || !gbSoundOn || gbBufferMsgs) {
+	if (!gbSndInited || !gbSoundOn || gbBufferMsgs != 0) {
```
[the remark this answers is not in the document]
AJenbo: I believe the b in the notation often stands for "byte" and not necessarily boolean, if nothing else it's a good argument for why we should move away from it (which we are) :)

Source/diablo.cpp
```diff
@@ -1965,11 +1770,9 @@ static void timeout_cursor(bool bTimeout)
  */
 void game_loop(bool bStartup)
 {
-	int i;
-
-	i = bStartup ? sgGameInitInfo.nTickRate * 3 : 3;
+	uint16_t i = bStartup ? sgGameInitInfo.nTickRate * 3 : 3;
 
-	while (i--) {
+	while ((i--) != 0) {
```
StephenCWills, line 1775: Perhaps this would be more readable this way?

```suggestion
	while (i > 0) {
		i--;
```

AJenbo: Indeed, i absolutely detest the other one.

AJenbo: Now that I am awake I think that this could be better written as:
```cpp
	uint16_t wait = bStartup ? sgGameInitInfo.nTickRate * 3 : 3;

	for (unsigned i = 0; i < wait; i++) {
```

AJenbo:
I almost considered this to be to trivial for a review, but your guys feed back was well worth it, thanks :)

Source/pack.h
```diff
@@ -65,6 +65,7 @@ struct PkPlayerStruct {
 	uint8_t pBattleNet;
 	uint8_t pManaShield;
 	uint8_t pDungMsgs2;
+    /** The format the charater is in, 0: Diablo, 1: Hellfire */
```
StephenCWills, line 68: Should this be tab-indented?

StephenCWills (changes requested):

Source/pack.h
AJenbo: yes

Source/cursor.cpp
```diff
@@ -334,11 +334,11 @@ void CheckCursMove()
...
 
 	// Align grid
-	if ((columns & 1) == 0 && (lrow & 1) == 0) {
+	if ((columns % 2) == 0 && (lrow % 2) != 0) {
 		sy += TILE_HEIGHT / 2;
-	} else if (columns & 1 && lrow & 1) {
+	} else if ((columns % 2) != 0 && (lrow % 2) != 0) {
 		sx -= TILE_WIDTH / 2;
-	} else if (columns & 1 && (lrow & 1) == 0) {
+	} else if ((columns % 2) != 0 && (lrow % 2) != 0) {
```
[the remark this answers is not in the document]
AJenbo: Good catch, not sure how I ended up doing that.

StephenCWills (approved):

PR #2233, opened 2021-06-26
Clean up more implicit bool conversion

AJenbo:


Source/player.cpp
```diff
@@ -1791,16 +1786,13 @@ StartPlayerKill(int pnum, int earflag)
...
-						while (i--) {
-							pdd = (i + player._pdir) & 7;
-							Point offset = Point { 0, 0 } + static_cast<Direction>(pdd);
-							PlrDeadItem(player, pi, offset.x, offset.y);
-							pi++;
+						Direction pdd = player._pdir;
+						for (int i = 0; i < NUM_INVLOC; i++) {
+							pdd = left[pdd];
+							PlrDeadItem(player, &player.InvBody[i], Point::fromDirection(pdd));
 						}
```
AJenbo, line 1796: There is a slight change to the logic here, the original would always drop the last item at position `-left[player._pdir]`, instead the new code will always drop the first item at position `+left[player._pdir]`.
I don't think the difference warrants keeping the extra logic as it not really relevant to anything.

Source/player.cpp
```diff
@@ -1791,16 +1786,13 @@ StartPlayerKill(int pnum, int earflag)
...
-						while (i--) {
-							pdd = (i + player._pdir) & 7;
-							Point offset = Point { 0, 0 } + static_cast<Direction>(pdd);
-							PlrDeadItem(player, pi, offset.x, offset.y);
-							pi++;
+						Direction pdd = player._pdir;
+						for (auto &item : player.InvBody) {
+							pdd = left[pdd];
+							PlrDeadItem(player, &item, Point::fromDirection(pdd));
 						}
```
AJenbo, line 1796: There is a slight change to the logic here, the original would always drop the last item at position `-left[player._pdir]`, instead the new code will always drop the first item at position `+left[player._pdir]`.
I don't think the difference warrants keeping the extra logic as it not really relevant to anything.

StephenCWills: Doesn't `pdd` need to be updated on each iteration?

StephenCWills (approved):

PR #2235, opened 2021-06-26
🌐 Update Italian localization

M0Rf30:
As in title,
Regards

### DIFF
--- a/Translations/it.po
+++ b/Translations/it.po
@@ -10139,15 +10139,3 @@ msgstr "Giù da Diablo"
 #: Source/trigs.cpp:722 Source/trigs.cpp:739 Source/trigs.cpp:756
 msgid "Back to Level {:d}"
 msgstr "Torna al Livello {:d}"
-
-#~ msgid "fully recover life"
-#~ msgstr "recupero totale vita"
-
-#~ msgid "recover mana"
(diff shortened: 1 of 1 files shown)


Translations/it.po
```diff
@@ -10032,105 +10037,117 @@ msgstr "Praedictum Otium."
...
+#~ msgstr "recupero totale vita"
+
+#~ msgid "recover mana"
+#~ msgstr "recupera mana"
+
+#~ msgid "fully recover mana"
+#~ msgstr "recupero totale mana"
+
+#~ msgid "fully recover life and mana"
+#~ msgstr "recupero totale vita e mana"
```
AJenbo, line 10153:
```suggestion
```

AJenbo:
Thanks :+1:

AJenbo:
Hey @M0Rf30 we ave 11 new string that needs translation if you have time :)

Not that most of them are part of the Android port where things are a bit different as it uses XML to handle it's translations template:
https://github.com/diasurgical/devilutionX/blob/master/android-project/app/src/main/res/values/strings.xml
translation: https://github.com/diasurgical/devilutionX/blob/master/android-project/app/src/main/res/values-it/strings.xml

PR #2237, opened 2021-06-26
Cleanup all implicit bool conversions

AJenbo:
With this `readability-implicit-bool-conversion` will be fully applied :)

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1334,17 +1334,17 @@ void M_Enemy(int i)
 			continue;
 		if (M_Talker(mi) && monster[mi].mtalkmsg != TEXT_NONE)
 			continue;
-		if ((Monst->_mFlags & MFLAG_GOLEM) && (monster[mi]._mFlags & MFLAG_GOLEM)) // prevent golems from fighting each other
+		if ((Monst->_mFlags & MFLAG_GOLEM) != 0 && (monster[mi]._mFlags & MFLAG_GOLEM) != 0) // prevent golems from fighting each other
 			continue;
 
 		dist = monster[mi].position.tile.WalkingDistance(Monst->position.tile);
-		if ((!(Monst->_mFlags & MFLAG_GOLEM)
-		        && !(Monst->_mFlags & MFLAG_BERSERK)
+		if (((Monst->_mFlags & MFLAG_GOLEM) == 0
+		        && (Monst->_mFlags & MFLAG_BERSERK) == 0
 		        && dist >= 2
 		        && !M_Ranged(i))
-		    || (!(Monst->_mFlags & MFLAG_GOLEM)
-		        && !(Monst->_mFlags & MFLAG_BERSERK)
-		        && !(monster[mi]._mFlags & MFLAG_GOLEM))) {
+		    || ((Monst->_mFlags & MFLAG_GOLEM) == 0
+		        && (Monst->_mFlags & MFLAG_BERSERK) == 0
+		        && (monster[mi]._mFlags & MFLAG_GOLEM) == 0)) {
 			continue;
 		}
 		sameroom = dTransVal[Monst->position.tile.x][Monst->position.tile.y] == dTransVal[monster[mi].position.tile.x][monster[mi].position.tile.y];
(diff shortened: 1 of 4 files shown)


Source/monster.cpp
```diff
@@ -2603,20 +2603,17 @@ bool M_DoDeath(int i)
...
-		if (x < 0)
-			x = -1;
-		else
-			x = x > 0;
-		ViewX += x;
+		if (monster[i].position.tile.x > ViewX) {
+			ViewX--;
+		} else if (monster[i].position.tile.x < ViewX) {
+			ViewX++;
+		}
```
StephenCWills, line 2610: Using substitution:
```
x < 0
monster[i].position.tile.x - ViewX < 0
monster[i].position.tile.x < ViewX
```

```suggestion
		if (monster[i].position.tile.x < ViewX) {
			ViewX--;
		} else if (monster[i].position.tile.x > ViewX) {
			ViewX++;
		}
```

Source/monster.cpp
```diff
@@ -2603,20 +2603,17 @@ bool M_DoDeath(int i)
...
 
-		y = monster[i].position.tile.y - ViewY;
-		if (y < 0) {
-			y = -1;
-		} else {
-			y = y > 0;
+		if (monster[i].position.tile.y > ViewY) {
+			ViewY--;
+		} else if (monster[i].position.tile.y < ViewY) {
+			ViewY++;
```
StephenCWills, line 2615:
```suggestion
		if (monster[i].position.tile.y < ViewY) {
			ViewY--;
		} else if (monster[i].position.tile.y > ViewY) {
			ViewY++;
```

Source/monster.cpp
```diff
@@ -2846,29 +2843,34 @@ bool M_DumbWalk(int i, Direction md)
...
 
-	Direction mdtemp = md;
 	bool ok = DirOK(i, md);
-	if (!ok) {
-		if (*dir) {
-			md = right[mdtemp];
-			ok = DirOK(i, md) || (md = right[right[mdtemp]], DirOK(i, md));
+
+	for (int j = 0; !ok && j < 2; j++) {
+		// Keep trying up to an including backwards
```
StephenCWills, line 2856: This looks to me like it's turning the opposite direction from the original 90-degree turn. So it's not turning the monster backwards, but rather forwards.

```suggestion
		// Keep trying by turning back toward the original direction
```

Source/monster.cpp
```diff
@@ -2846,29 +2843,34 @@ bool M_DumbWalk(int i, Direction md)
...
+		ok = DirOK(i, md);
 	}
-	if (ok) {
-		M_WalkDir(i, md);
-	} else {
-		*dir = !*dir;
-		ok = M_CallWalk(i, opposite[mdtemp]);
+
+	if (!ok) {
+		// Try 90 degrees in the oposit then desired direction
```
StephenCWills, line 2866: Just grammar and spelling here.

```suggestion
		// Try 90 degrees in the opposite than desired direction
```

StephenCWills (changes requested):

Source/monster.cpp
AJenbo: Oh you are right, I might rethink how this is done in that case.

StephenCWills (dismissed):
It looks correct now, but I wonder if you might like it better if we reduce nesting, like below. This feels a bit more straightforward to me.

```C++
bool M_RoundWalk(int i, Direction direction, int *dir)
{
	Direction turn45deg = turn(direction, *dir != 0);
	Direction turn90deg = turn(turn45deg, *dir != 0);

	if (DirOK(i, turn90deg)) {
		// Turn 90 degrees
		M_WalkDir(i, turn90deg);
		return true;
	}

	if (DirOK(i, turn45deg)) {
		// Only do a small turn
		M_WalkDir(i, turn45deg);
		return true;
	}

	if (DirOK(i, direction)) {
		// Continue straight
		M_WalkDir(i, direction);
		return true;
	}

	// Try 90 degrees in the opposite than desired direction
	*dir = (*dir == 0) ? 1 : 0;
	return M_CallWalk(i, opposite[turn90deg]);
}
```

AJenbo:
> I wonder if you might like it better if we reduce nesting

Yeah I had an inner conflict where I couldn't decide if I would avoid the repetition or nesting. You suggestion of utilizing the direction variables directly does improve both so I think that is indeed the best solution here.

PR #2240, opened 2021-06-26
renewed translation

ybrbnf2080:
I freed myself and decided to continue


julealgon:
Please be specific when creating issues so that people can know what it is about without having to click on them. In this case, please specify the language at least.

PR #2241, opened 2021-06-26
Clang-tidy: readability-uppercase-literal-suffix

AJenbo:
Suffixes are rarely used in the code and these cases do not seam to add any value, should I simply strip them instead of applying this rule?


julealgon:
> Suffixes are rarely used in the code

For this reason alone, I'd prefer we just remove the instances where there is a suffix applied instead.

PR #2245, opened 2021-06-27
fix compiling on 3ds

qndel:


AJenbo:
Thanks, also corrected the NOLINT line so this doesn't happen when running clang-tidy: https://github.com/diasurgical/devilutionX/commit/a17d96129fcbf39f7eda7c099416b33561154cd7

PR #2226, opened 2021-06-25
Update random number functions to avoid IB

ephphatha:
Most calls to set seed were using uint32_t already, there were a few variables which were better served by having their type changed from signed to unsigned. The one exception is ItemStruct::_iSeed. This is an identifier that also happens to be used as a seed occasionally, ~so when that is passed to SetLCGEngineState I cast to uint32_t manually~. I have left this doing an implicit conversion to avoid excessive changes in this PR.

~As part of this change some functions were renamed to better indicate their use. This is deliberately awkward naming as this random engine needs to maintain compatibility with the base game.~

PR #2246, opened 2021-06-27
Split RNG functions to dedicated source files

ephphatha:
Depends on #2226. Only the second commit is relevant to this request, I opened this as a separate PR to hopefully make it easier to review.

This is in preparation for documenting the existing RNG functions, how they  are used, and providing more suitable alternatives for modders/where repeatable sequences are not necessary.


AJenbo:
Ready for rebase

AJenbo (approved):

PR #2259, opened 2021-06-28
Introduce 'Displacement' abstraction

julealgon:
This PR adds a new `Displacement` abstraction to the codebase for use with deltas/offsets instead of using `Point`s for that purpose.

The name was chosen to avoid any confusion with the native `vector`, but this is conceptually the exact same: a vector with deltas for X and Y displacement.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -976,11 +976,11 @@ void HotSpellMove(AxisDirection dir)
 	}
 
 	if (dir.y == AxisDirectionY_UP) {
-		if (HSExists(newMousePosition - Point { 0, SPLICONLENGTH })) {
+		if (HSExists(newMousePosition - Displacement { 0, SPLICONLENGTH })) {
 			newMousePosition.y -= SPLICONLENGTH;
 		}
 	} else if (dir.y == AxisDirectionY_DOWN) {
-		if (HSExists(newMousePosition + Point { 0, SPLICONLENGTH })) {
+		if (HSExists(newMousePosition + Displacement { 0, SPLICONLENGTH })) {
 			newMousePosition.y += SPLICONLENGTH;
 		}
 	}
@@ -1267,8 +1267,8 @@ void HandleRightStickMotion()
 		int dx = 0;
 		int dy = 0;
 		acc.Pool(&dx, &dy, 32);
-		AutomapOffset.x += dy + dx;
-		AutomapOffset.y += dy - dx;
+		AutomapOffset.deltaX += dy + dx;
+		AutomapOffset.deltaY += dy - dx;
 		return;
 	}
 
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1417,7 +1417,7 @@ void UpdateSpellTarget()
 	if (plr[myplr]._pRSpell == SPL_TELEPORT)
 		range = 4;
 
-	auto cursm = plr[myplr].position.future + Point::fromDirection(plr[myplr]._pdir) * range;
+	auto cursm = plr[myplr].position.future + Displacement::fromDirection(plr[myplr]._pdir) * range;
 	cursmx = cursm.x;
 	cursmy = cursm.y;
 }
--- a/Source/engine/displacement.hpp
+++ b/Source/engine/displacement.hpp
@@ -82,6 +82,30 @@ struct Displacement {
 	{
 		return { abs(a.deltaX), abs(a.deltaY) };
 	}
+
+	static constexpr Displacement fromDirection(Direction direction)
+	{
+		switch (direction) {
+		case DIR_S:
+			return { 1, 1 };
+		case DIR_SW:
+			return { 0, 1 };
+		case DIR_W:
+			return { -1, 1 };
+		case DIR_NW:
+			return { -1, 0 };
+		case DIR_N:
+			return { -1, -1 };
+		case DIR_NE:
+			return { 0, -1 };
+		case DIR_E:
+			return { 1, -1 };
+		case DIR_SE:
+			return { 1, 0 };
+		default:
+			return { 0, 0 };
+		}
+	};
 };
 
 } // namespace devilution
--- a/Source/engine/displacement.hpp
+++ b/Source/engine/displacement.hpp
@@ -1,0 +1,87 @@
+#pragma once
+
+#include <cmath>
+
+#include "direction.hpp"
+#include "utils/stdcompat/abs.hpp"
+
+namespace devilution {
+
+struct Displacement {
+	int deltaX;
+	int deltaY;
+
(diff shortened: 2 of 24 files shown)


ephphatha (dismissed):
😍 makes it much clearer what the intent of certain operations is my eyes. This also highlighted one odd check in player.cpp which I couldn't see any similar instance of but that can be fixed in another PR.

dcworldwide:
Hello, is it still just a Point, with the only diff/value been purely sematic? If yes, will a variable naming convention of offset not suffice?

AJenbo:
> Hello, is it still just a Point, with the only diff/value been purely sematic? If yes, will a variable naming convention of offset not suffice?

It can also help prevent some mistakes by not allowing non nonsensical operations, and notice that the the example given by @ephphatha one is not a variable.

julealgon:
@obligaron please see if you can take a look at this change as well. I left you a message in discord about it.

Source/player.cpp
```diff
@@ -67,7 +67,7 @@ void WalkUpwards(int pnum, const DirectionSettings &walkParams)
 {
 	auto &player = plr[pnum];
 	dPlayer[player.position.future.x][player.position.future.y] = -(pnum + 1);
-	player.position.temp = walkParams.tileAdd;
+	player.position.temp = { walkParams.tileAdd.deltaX, walkParams.tileAdd.deltaY };
```
obligaron, line 70: As I said in discord. In this case `position.temp` is used as a `Displacement`.
[This](https://github.com/diasurgical/devilutionX/pull/2259/files#diff-71461699ae2f94f652e6cba2136d3c696151ba3a6c0e357eaf55438a5c212ad3R2107) is the usage of the set variable.
But in all other occurrences it's used as a point.

It's a little bit sad that the original design is not straight forward. :cry:
If we want to clarify this, we would need two different variables and we would adjust loading/saving to stay compatible with vanilla (check mode and if mode is PM_WALK use `Displacement` instead of `Point` variable).
But imho the gain is not worth the trouble.
If you want you can write a comment to explain this conversion. 🙂 

Note:
In Monster they even used another set of variables. 😕

Source/controls/plrctrls.cpp
```diff
@@ -1267,8 +1267,8 @@ void HandleRightStickMotion()
 		int dx = 0;
 		int dy = 0;
 		acc.Pool(&dx, &dy, 32);
-		AutomapOffset.x += dy + dx;
-		AutomapOffset.y += dy - dx;
+		AutomapOffset.deltaX += dy + dx;
```
AJenbo, line 1270: I guess an alternative name would be Offset, Displacment is fine though :)

Source/player.cpp
julealgon: Fair enough. Will leave this as-is then as the only exception. I'm actually happy this refactor applied smoothly to all other parts of the code, though.

julealgon:
Looks like some tests are failing. Will look into those in a few hours.

AJenbo:
> Looks like some tests are failing. Will look into those in a few hours.

Looks like you didn't update the test where the object where changed from Point to Displacment (x vs deltaX)

Source/engine/point.hpp
```diff
@@ -91,17 +68,22 @@ struct Point {
...
-		return a;
+		return { b.x - a.x, b.y - a.y };
 	}
 
 	constexpr friend Point operator-(const Point &a)
 	{
 		return { -a.x, -a.y };
 	}
 
+	constexpr friend Point operator-(Point a, Displacement displacement)
```
AJenbo, line 81: How about + for symmetry?

Source/player.cpp
```diff
@@ -1396,23 +1396,23 @@ void PM_ChangeOffset(int pnum)
...
-	px -= player.position.offset2.x >> 8;
-	py -= player.position.offset2.y >> 8;
+	px -= player.position.offset2.deltaX >> 8;
+	py -= player.position.offset2.deltaY >> 8;
 
 	if (pnum == myplr && ScrollInfo._sdir != SDIR_NONE) {
-		ScrollInfo.offset.x += px;
-		ScrollInfo.offset.y += py;
+		ScrollInfo.offset.deltaX += px;
+		ScrollInfo.offset.deltaY += py;
```
AJenbo, line 1415:
```suggestion
		ScrollInfo.offset += Displacement { px, py };
```

Source/player.cpp
```diff
@@ -1396,23 +1396,23 @@ void PM_ChangeOffset(int pnum)
...
 		player.position.offset2 += player.position.velocity;
 	}
 
-	player.position.offset = { player.position.offset2.x >> 8, player.position.offset2.y >> 8 };
+	player.position.offset = { player.position.offset2.deltaX >> 8, player.position.offset2.deltaY >> 8 };
 
-	px -= player.position.offset2.x >> 8;
-	py -= player.position.offset2.y >> 8;
+	px -= player.position.offset2.deltaX >> 8;
+	py -= player.position.offset2.deltaY >> 8;
```
AJenbo, line 1411: wonder if we should add `>>` as an operator or use `/` here instead

Source/engine/displacement.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <cmath>
+
+#include "direction.hpp"
+#include "utils/stdcompat/abs.hpp"
+
+namespace devilution {
+
+struct Displacement {
+	int deltaX;
+	int deltaY;
```
glebm, line 12: The `delta` prefix seems redundant here, it's already a `Displacement`?

Source/engine/point.hpp
ephphatha: I believe this is deliberate. A `Displacement` represents the difference between two `Points` so it is only well defined for subtraction. Combining (adding) two points is rare and I think it would be better expressed as `displacementFromOrigin = (endPoint - origin); startPoint + displacementFromOrigin;`.

Shortcuts could be provided using `Displacement::Displacement(Point)` or a similar conversion from a `Point` to a `Displacement` implied to be from the origin if it turns out to be useful.

ephphatha (dismissed):

Source/player.cpp
ephphatha: This looks like a compiler optimised division. `/` gets my vote.

julealgon: Outside the scope of the PR.

Source/engine/displacement.hpp
julealgon: I think the prefix makes sense and adds semantic to usages. The value itself should provide semantic to what it is without necessarily depending on the parent class/struct in some cases.

julealgon:
Finally passed pipeline. Apologies for the back and forth on this one.

AJenbo:
well im glad the tests where there so that missiles are mocing backwards now :D

Issue #2206, opened 2021-06-21
The current RNG implementation generates a negative number from a specific seed


ephphatha:
**Important information**
Bug present in all DevilutionX versions since the initial commit on any compiler with similar implementation defined behaviour as msvc

**Describe**
Calling `GenerateRnd(int)` with a seed of 1457187811 (either supplied as the seed from a call to `SetRndSeed(int)` immediately prior or reached naturally from calls to `AdvanceRndSeed()`) results in the next call to either `AdvanceRndSeed()` or `GenerateRndSeed(x)` (where x is any value in the range [2, INT_MAX]) generating a negative number.

In instances like this where the number is used in a switch statement none of the cases will be taken, but the program will continue as normal.
https://github.com/diasurgical/devilutionX/blob/14218080b667e2572d20cd94b34ff0f409e4f978/Source/drlg_l1.cpp#L1602

There are instances where a random number is used as an array index where the possible values are still within bounds. For example `PlaySfx` calls `RndSfx` to get a random sound from a certain category. The lowest value that could be returned is 8 (corresponding to the effect PS_LGHIT1) when called with PS_SWING (expected values are either 9 or 10, PS_SWING or PS_SWING2)
https://github.com/diasurgical/devilutionX/blob/14218080b667e2572d20cd94b34ff0f409e4f978/Source/effects.cpp#L1256

Other instances will lead to reading a negative array index. The code which looks for a free item space is one example, this code can generate a point with an x or y position of -63 (16 + -79) which is immediately used as an index into `dMonster`
https://github.com/diasurgical/devilutionX/blob/14218080b667e2572d20cd94b34ff0f409e4f978/Source/items.cpp#L457
https://github.com/diasurgical/devilutionX/blob/14218080b667e2572d20cd94b34ff0f409e4f978/Source/items.cpp#L437

**To Reproduce**
Edit a save or use breakpoints to seed the RNG before one of these call sites. If there's a save with a known seed leading to a crash when generating a dungeon that would be great, I haven't tried to craft one yet.

**Expected behavior**
I would expect non-critical calls to the RNG to generate within the expected range, so situations like `PlaySfx` should always play one of the expected sounds.

Critical calls to the RNG should mimic the same behaviour if it didn't lead to a crash. Using the first example to keep compatibility with old saves/original clients bypassing the defined cases in the switch statement is desirable.

[...]

Issue #2193, opened 2021-06-19
Picking one of two options is performed in multiple ways in the code


ephphatha:
There are a few places in the code where the intent is to pick between two options at random but due to inconsistent implementations it is non-obvious that they have the same intent.

For example, this code gets either +1 or -1 (avoiding the value 0) to use as a tile offset.
https://github.com/diasurgical/devilutionX/blob/4e6870a92fcbffa8f15b82f0a1b7b7edc4b30649/Source/monster.cpp#L2485

Other places also use +1 or -1 (avoiding the value 0), in this case as a rotation direction. This is the same intent as above but a much simpler implementation.
https://github.com/diasurgical/devilutionX/blob/4e6870a92fcbffa8f15b82f0a1b7b7edc4b30649/Source/missiles.cpp#L994

There are other places performing coin flips to determine which branch to take. These usually use a comparison of `random value in the interval [0, 1] != 0`.
https://github.com/diasurgical/devilutionX/blob/4e6870a92fcbffa8f15b82f0a1b7b7edc4b30649/Source/monster.cpp#L2836
With some exceptions, e.g.
https://github.com/diasurgical/devilutionX/blob/4e6870a92fcbffa8f15b82f0a1b7b7edc4b30649/Source/missiles.cpp#L1423

I think it would be worth having a GenerateRndBoolean function for this use case. Probably just a wrapper for `GenerateRnd(2) != 0`, preferably using the generator from #2084.

AJenbo:
If alone :

GenerateRndBoolean ? 1:-1

Else 

if(GenerateRndBoolean && otherSuff)

ephphatha:
Turns out there's a [bernoulli distribution](https://en.cppreference.com/w/cpp/numeric/random/bernoulli_distribution). This would allow replacing calls such as:
https://github.com/diasurgical/devilutionX/blob/43bc08bd63e321505ec4961a9339391f9c3bc9ac/Source/monster.cpp#L3343
with code like:
```cpp
GenerateRndBoolean(double p)
{
    return bernoulli_distribution(rng, p) //rng is a std::mersenne_twister_engine or similar Generator
}

// Get a true value 75% of the time
//GenerateRnd(4) != 0
GenerateRndBoolean(0.75)
```

AJenbo:
Not that we must call RNG identical to the original for thing like item and level generation.

julealgon:
[...]

PR #2261, opened 2021-06-28
Add alternate RNG and proposed API for versioning Random Number Generators

ephphatha:
This builds on the work done by @qndel in #2084. I have not updated any of the call sites identified in that PR as I wanted to make it obvious that this does not change existing code.

This is more of a discussion point at the moment. I was thinking that namespaces would be a good way to delineate behaviour. I've implemented "vanilla" to cover the base game logic and bugs. "randomV1" is based on #2084 and uses STL concepts to add convenience functions as discussed in #2193 and #2206.

The API design is intended to be common between versions. The next step would be to update the vanilla namespace functions to align member functions of vanilla::RandomEngine. The reason for having this class is to allow creating a RandomEngine object and passing it in to the level/item generation code (for example) to make it clearer how a given object/dungeon layout is generated.

Looking at code such as quests.cpp:InitQuests() makes me think a "vanillaSafe" namespace which fixes the negative number bug would be useful, but this would be a future enhancement iff this proposal is acceptable.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1905,7 +1905,7 @@ void LoadGameLevel(bool firstflag, lvl_entry lvldir)
 	if (pcurs > CURSOR_HAND && pcurs < CURSOR_FIRSTITEM) {
 		NewCursor(CURSOR_HAND);
 	}
-	SetRndSeed(glSeedTbl[currlevel]);
+	vanilla::SetRndSeed(glSeedTbl[currlevel]);
 	IncProgress();
 	MakeLightTable();
 	LoadLvlGFX();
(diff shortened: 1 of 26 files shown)


ephphatha:
The main changes in this PR are the addition of RandomEngine as a class in engine/random.hpp (and implementation of global functions in random.cpp), versioning namespaces, and updated test cases showing usages. I have updated all uses of RNG functions to refer to the vanilla:: namespaced versions because I don't like having unbuildable commits. I can split that if desired.

@qndel would I be able to add you as a co-author for this PR? I didn't want to imply you had signed off on this but I do want to acknowledge your work in this space.

ephphatha:
Added an example of how this could be used to make it easier to ensure code which relies on pseudo-random sequences is functioning correctly. This PR is already massive so I didn't want to make it worse by actually using the new API, but hopefully this gives context for why I think this change is worthwhile.

AJenbo:
The coverage result is pretty amazing for this PR :)
image

PR #2262, opened 2021-06-28
More init clean up

AJenbo:
... again half of the remaining, at this rate it will take infinite time to finish.

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -493,10 +493,6 @@ char GetLight(int x, int y)
 
 void DoLighting(Point position, int nRadius, int Lnum)
 {
-	int x, y, v, mult, radius_block;
-	int min_x, max_x, min_y, max_y;
-	int dist_x, dist_y, temp_x, temp_y;
-
 	int xoff = 0;
 	int yoff = 0;
 	int light_x = 0;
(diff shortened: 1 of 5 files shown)


Source/lighting.cpp
```diff
@@ -493,10 +493,6 @@ char GetLight(int x, int y)
 
 void DoLighting(Point position, int nRadius, int Lnum)
 {
-	int x, y, v, mult, radius_block;
-	int min_x, max_x, min_y, max_y;
-	int dist_x, dist_y, temp_x, temp_y;
```
julealgon, line 498: ...lol at the amount of variables. This code man...

Source/msg.cpp
```diff
@@ -804,15 +797,15 @@ void DeltaLoadLevel()
...
-						for (j = -k; j <= k && !done; j++) {
-							yy = y + j;
-							for (l = -k; l <= k && !done; l++) {
-								xx = x + l;
+					bool done = false;
+					for (int k = 1; k < 50 && !done; k++) {
+						for (int j = -k; j <= k && !done; j++) {
+							int yy = y + j;
+							for (int l = -k; l <= k && !done; l++) {
+								int xx = x + l;
```
julealgon, line 808: My god...

julealgon (approved):

Source/msg.cpp
qndel: he's not even yours! 😃

julealgon: This might be the most indented code I've ever seem in a real code actually... now that I think about it.

AJenbo: Still two less then what I have to deal with professionally.

PR #2268, opened 2021-06-30
Switch multiplayer test

StephenCWills:
Testing smaller buffer size to see if it fixes multiplayer on Switch.

### DIFF
--- a/Source/dvlnet/frame_queue.h
+++ b/Source/dvlnet/frame_queue.h
@@ -22,7 +22,7 @@ typedef uint32_t framesize_t;
 
 class frame_queue {
 public:
-	constexpr static framesize_t max_frame_size = 0xFFFF;
+	constexpr static framesize_t max_frame_size = 0x1000;
 
 private:
 	framesize_t current_size = 0;
--- a/Source/dvlnet/tcp_server.cpp
+++ b/Source/dvlnet/tcp_server.cpp
@@ -6,6 +6,7 @@
 #include <utility>
 
 #include "dvlnet/base.h"
+#include "plrmsg.h"
 #include "utils/log.hpp"
 
 namespace devilution {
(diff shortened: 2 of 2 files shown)


StephenCWills:
@rsn8887, @spitfire, or anyone else who is willing to test. This build uses a much smaller recv() buffer to ensure that IPC messages fit into the "pointer buffer" space. With this change, multiplayer works in Ryujinx without any workarounds. I'm hopeful that means this build will work on the Switch console as well.

https://48703-143324737-gh.circle-artifacts.com/0/devilutionx.nro

spitfire:
Still ends up with "unable to connect". Is there any more information I can collect? I've tried hosting the game on the PC (I've let game through Windows Firewall) and on the switch. I've connected both PC and Switch to the same router, using their respective IP addresses to connect and '1234' as a password.

AJenbo:
Could you test with two pcs, or maybe just running two instances of the game on the same pc

StephenCWills:
> Still ends up with "unable to connect". Is there any more information I can collect?

Honestly, I can't even think of anything to log to help me solve this. The error occurs at the bottom of this section of code.
https://github.com/diasurgical/devilutionX/blob/dadf418fe080b7f1d456e12d11f30f943bdd16a9/Source/dvlnet/tcp_client.cpp#L48-L70

This means the console makes it through resolving the IP/hostname and the three-way handshake and believes it's sending the `PT_JOIN_REQUEST` packet, but is unable to receive data on the socket. When this was happening to me with Ryujinx, it was because Ryujinx was trying to read into the zero-length "Receive List" buffer instead of the "Receive" buffer. By lowering the buffer size, it allows libnx to fit the buffer in the "pointer buffer" space so that it allocates space for the "Receive List" buffer instead of the "Receive" buffer. It should fail to host a game in the first place since it has to make a local connection.

I guess, since the console still fails despite of the buffer size change, there's a chance that the TCP server never received the `PT_JOIN_REQUEST` packet in the first place. That's something we could attempt to verify with log messages or Wireshark. I still would have no idea what to do about it, but I'd be able to provide a bit more information to the devkitPro developers.

StephenCWills:
The latest commit adds the following message when a player attempts to join your game.
image

The best way to test this would be to host the game on PC, then watch the screen while you attempt to join the game on Switch. If you see the message, that should tell us pretty definitively whether the `PT_JOIN_REQUEST` packet was received by the TCP server.

Windows x86: https://github.com/diasurgical/devilutionX/suites/3123071942/artifacts/71571486
Windows x64: https://github.com/diasurgical/devilutionX/suites/3123071943/artifacts/71571883

StephenCWills:
FYI, here is the link to the forum topic I created on the devkitPro forums.
https://devkitpro.org/viewtopic.php?f=42&t=9240

StephenCWills:
I just received a report that multiplayer on Switch is working in master. I suspect that the rdimon library, which was removed in #2730, was causing problems.

AJenbo:
grate work getting this running!

Issue #2267, opened 2021-06-29
dLight array is saved/loaded twice


qndel:
dLight array is saved/loaded twice

See LoadGame/SaveGameData

https://github.com/diasurgical/devilutionX/blob/master/Source/loadsave.cpp#L1194
https://github.com/diasurgical/devilutionX/blob/master/Source/loadsave.cpp#L1224

https://github.com/diasurgical/devilutionX/blob/master/Source/loadsave.cpp#L1960
https://github.com/diasurgical/devilutionX/blob/master/Source/loadsave.cpp#L1990

julealgon:
What's the underlying issue related to this double save/load? Can you detail the problem it causes a bit more?

ephphatha:
It's effectively dead code when loading/saving in a dungeon as the second read/write tramples the first. It also is redundant for saves made while in town as the light array should be all zeroes, vanilla logic loads non-zero values if they're present in the save which causes bugs/crashes with dodgy saves that devilutionX can handle gracefully.

PR #2270, opened 2021-06-30
fix duplicated dLight saving/loading

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2267

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -1189,10 +1189,7 @@ void LoadGame(bool firstflag)
 	for (bool &UniqueItemFlag : UniqueItemFlags)
 		UniqueItemFlag = file.nextBool8();
 
-	for (int j = 0; j < MAXDUNY; j++) {
-		for (int i = 0; i < MAXDUNX; i++) // NOLINT(modernize-loop-convert)
-			dLight[i][j] = file.nextLE<int8_t>();
-	}
+	file.skip(MAXDUNX * MAXDUNY); // BUGFIX: dLight array is saved/loaded twice (fixed)
 	for (int j = 0; j < MAXDUNY; j++) {
 		for (int i = 0; i < MAXDUNX; i++) // NOLINT(modernize-loop-convert)
 			dFlags[i][j] = file.nextLE<int8_t>();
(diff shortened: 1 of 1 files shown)


ephphatha:
This exposes an issue with the current save/load code 😂 . Making a save while in town with this fix results in the first file region where the light array was saved being set to the value of the uninitialised byte array backing SaveHelper (note 1). If you then load that save in a version of devilutionx without this fix applied it'll load the values into dlight due to the first load and likely crash depending on whether the value happens to be valid. Making a save in a debug build writes the value 0xCD repeatedly in this region. Old versions will load this value and it is interpreted as -51 when assigning to light_table_index
https://github.com/diasurgical/devilutionX/blob/3f69faaecfa7069c17e51c05e84f6ddc27e56814/Source/scrollrt.cpp#L596
This then causes a buffer underrun in `RenderTile`
https://github.com/diasurgical/devilutionX/blob/3f69faaecfa7069c17e51c05e84f6ddc27e56814/Source/engine/render/dun_render.cpp#L1399

1: https://github.com/diasurgical/devilutionX/commit/f33f7ae7ebb8583a2bb270b4d27b3f39281a9929#diff-eb2db6084d42c0224b56a122152915d07d944c27040e427d6193bc382ca754b3 introduced the potential for this bug, a fix that would avoid having to zero-init the entire array would be to make skip() memset the region it skips over. Given there hasn't been a release since that commit it's only going to be an issue for people running their own builds, a mitigation could be to leave the duplicate save in for a while.

qndel:
Tbh, I'd get rid of saving / loading light at all + https://github.com/diasurgical/devilutionX/pull/2110/

ephphatha (approved):
I think now that SaveHelp::skip writes 0's this is safe to merge, town expects a light level of zero for older versions so this should be backwards compatible.

AJenbo:
Won't this cause save games to load with no lights in vanilla?

ephphatha:
When loading a save made inside the dungeon dLight gets initialised from the second load (unchanged by this commit):
https://github.com/diasurgical/devilutionX/blob/e1f2fbcb9b87e95d8c283e833610de253f83b4aa/Source/loadsave.cpp#L1225-L1228

When in town the light level doesn't get applied to the tiles as far as I can tell, and even if it does the expected value for dLight values in town is 0, the entire array is memset when the level is loaded from a normal transition:
https://github.com/diasurgical/devilutionX/blob/e1f2fbcb9b87e95d8c283e833610de253f83b4aa/Source/town.cpp#L322

PR #2271, opened 2021-07-01
Clean up variable init in drlg files

AJenbo:


Source/drlg_l2.cpp
```diff
@@ -1751,16 +1747,13 @@ static void DRLG_L2PlaceRndSet(const BYTE *miniset, int rndper)
...
-	int x, y, i, j, k, rv;
-	BYTE c;
-
-	for (y = 0; y < DMAXY; y++) {
-		for (x = 0; x < DMAXX; x++) {
+	for (int y = 0; y < DMAXY; y++) {
+		for (int x = 0; x < DMAXX; x++) {
 			if ((x < nSx1 || x > nSx2) && (y < nSy1 || y > nSy2) && GenerateRnd(4) == 0) {
-				c = BTYPESL2[dungeon[x][y]];
+				int8_t c = BTYPESL2[dungeon[x][y]];
```
StephenCWills, line 1753: Should this be uint8_t?

```suggestion
				uint8_t c = BTYPESL2[dungeon[x][y]];
```

StephenCWills (approved):
I just had the one small comment. I expect it should work with int8_t, but it makes more sense to use uint8_t.

Source/drlg_l4.cpp
```diff
@@ -1129,9 +1097,15 @@ static bool L4checkRoom(int x, int y, int width, int height)
 
 static void L4roomGen(int x, int y, int w, int h, int dir)
 {
-	bool ran, ran2;
-	int width, height, rx, ry, ry2;
-	int cw, ch, cx1, cy1, cx2;
+	bool ran;
+	int width;
+	int height;
```
ephphatha, line 1102: These two declarations can be moved to line 1135

Source/drlg_l4.cpp
```diff
@@ -1129,9 +1097,15 @@ static bool L4checkRoom(int x, int y, int width, int height)
...
-	int cw, ch, cx1, cy1, cx2;
+	bool ran;
+	int width;
+	int height;
+	int rx;
+	int ry;
+	int cw;
+	int ch;
+	int cx1;
+	int cy1;
```
ephphatha, line 1108: These can move inside the if statement on line 1113

ephphatha (approved):

PR #2272, opened 2021-07-01
Clean-up uneccesary type conversions in player hpPer calculations

ephphatha:
The player HP/Mana percentage calculations are doing conversions that are unnecessary with the operands and the expected values during gameplay. Refactoring the life flask upper section for now, I'll have a go at the rest when I get more time.

This also includes explicit casts for Point/Displacement operator/ because I was planning on doing a bunch of small changes then I almost immediately hit the code in DrawLifeFlask 😂 . I can break it out if desired.

### DIFF
--- a/Source/player.h
+++ b/Source/player.h
@@ -402,6 +402,27 @@ struct PlayerStruct {
 	 * @brief Resets all Data of the current PlayerStruct
 	 */
 	void Reset();
+
+	/**
+	 * @brief Calculates the players current Hit Points as a percentage of their max HP and stores it for later reference
+	 *
+	 * The stored value is unused...
+	 * @see _pHPPer
+	 * @return The players current hit points as a percentage of their maximum (from 0 to 80%)
+	*/
+	int UpdateHitPointPercentage()
+	{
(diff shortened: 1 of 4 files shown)


Source/player.h
```diff
@@ -402,6 +402,27 @@ struct PlayerStruct {
 	 * @brief Resets all Data of the current PlayerStruct
 	 */
 	void Reset();
+
+	/**
+	 * @brief Calculates the players current Hit Points as a percentage of their max HP and stores it for later reference
+	 *
+	 * The stored value is unused...
+	 * @see _pHPPer
```
AJenbo, line 410: Is the value used by vanilla? If not then let's just skip it.

AJenbo (approved):

Source/player.h
ephphatha: I'm guessing the original intent was to refer to this value when drawing the UI and only recalculate it when the players max or current HP changes. I'd like to go back to that behaviour eventually since it doesn't really need to be recalculated every frame.

AJenbo:
Really nice that some of the MSVC warnings are getting addressed :)

PR #2275, opened 2021-07-01
Clean up missiles var init

AJenbo:


Source/missiles.cpp
```diff
@@ -4133,39 +4083,38 @@ void MI_Guardian(int i)
...
+				if (offset == Displacement { 0, 0 }) {
+					break;
+				}
+				if (previous == offset) {
 					continue;
 				}
-				sx = missile[i].position.tile.x + vCrawlTable[j][k];
-				sy = missile[i].position.tile.y + vCrawlTable[j][k + 1];
-				ex = Sentfire(i, { sx, sy });
+				ex = Sentfire(i, { position.x + offset.deltaX, position.y + offset.deltaY });
```
ephphatha, line 4101: With how this has been refactored to use `Point`/`Displacement` types I think an element-wise (Hadamard) product function on Displacement could help explain what the tests are doing.

```cpp
/**
 * @brief Computes the element-wise product of two Displacements.
 *
 * @return { a.deltaX * b.deltaX, a.deltaY * b.deltaY }
 */
constexpr friend Displacement HadamardProduct(Displacement a, Displacement b) {
    return { a.deltaX * b.deltaX, a.deltaY * b.deltaY };
}

constexpr friend Displacement HadamardProduct(Displacement a, Direction dir) {
    return a.HadamardProduct(Displacement::fromDirection(dir));
}
```

```cpp
Sentfire(i, position + offset.HadamardProduct(DIR_S))
```

It's a very unusual operation though so there's likely a way to refactor this further to avoid needing it.

Not something I'd worry about now, I'd log an issue for this suggestion if I was more confident it's even useful.

Source/missiles.cpp
```diff
@@ -4931,8 +4857,7 @@ void ProcessMissiles()
 		}
 	}
 
-	i = 0;
-	while (i < nummissiles) {
+	for (int i = 0; i < nummissiles;) {
 		if (missile[missileactive[i]]._miDelFlag) {
 			DeleteMissile(missileactive[i], i);
 			i = 0;
```
ephphatha, line 4863:
```suggestion
```
This assignment is unnecessary, the DeleteMissile function only modifies `missileactive[i]`

Source/missiles.cpp
```diff
@@ -4894,17 +4823,14 @@ void MI_Rportal(int i)
...
 		if (missile[missileactive[i]].position.tile.x < 0 || missile[missileactive[i]].position.tile.x >= MAXDUNX - 1 || missile[missileactive[i]].position.tile.y < 0 || missile[missileactive[i]].position.tile.y >= MAXDUNY - 1)
 			missile[missileactive[i]]._miDelFlag = true;
 	}
 
-	i = 0;
-	while (i < nummissiles) {
+	for (int i = 0; i < nummissiles;) {
 		if (missile[missileactive[i]]._miDelFlag) {
 			DeleteMissile(missileactive[i], i);
 			i = 0;
```
ephphatha, line 4836:
```suggestion
```
This assignment is unnecessary, the DeleteMissile function only modifies `missileactive[i]`

ephphatha (approved):
Couple of potential improvements that could be done in this PR or as a new PR. The ProcessMissile stuff seemed too small to write an issue for.

Source/missiles.cpp
```diff
@@ -2003,15 +1988,13 @@ void AddKrull(int mi, Point src, Point dst, int /*midir*/, int8_t /*mienemy*/, i
...
-	for (i = 0; i < 6; i++) {
-		k = CrawlNum[i];
-		pn = k + 2;
-		for (j = (BYTE)CrawlTable[k]; j > 0; j--) {
-			tx = dst.x + CrawlTable[pn - 1];
-			ty = dst.y + CrawlTable[pn];
+	for (int i = 0; i < 6; i++) {
+		int k = CrawlNum[i];
+		int pn = k + 2;
+		for (int j = static_cast<uint8_t>(CrawlTable[k]); j > 0; j--) {
```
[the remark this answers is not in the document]
AJenbo: It's a good point, I probably won't be applying it to everything here so creating an issue sounds like a good idea.

Source/missiles.cpp
```diff
@@ -344,26 +344,14 @@ static void PutMissile(int i)
...
-		dx = dx / 64;
-	}
-	if (dy < 0) {
-		ly = -(-dy / 8);
-		dy = -(-dy / 64);
-	} else {
-		ly = dy / 8;
-		dy = dy / 64;
-	}
+	int lx = dx / 8;
```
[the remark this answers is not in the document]
AJenbo: The original code was obsessed with using bit shifting where ever possible to optimize things:
https://github.com/diasurgical/devilution/blob/master/Source/missiles.cpp#L486

To be fair MSVC 4/5/6 where flaky with detecting this optimization (why we could tell in the first place).

We changed things to multiply instead in a previous cleanup: https://github.com/diasurgical/devilutionX/commit/ef20d1a038ba420f86ef07baabcf1bc74c0ea717

Source/missiles.cpp
```diff
@@ -890,7 +877,8 @@ void CheckMissileCol(int i, int mindam, int maxdam, bool shift, Point position,
 {
 	int oi;
 	bool blocked;
-	int dir, mAnimFAmt;
+	int dir;
+	int mAnimFAmt;
```
AJenbo, line 881: missed this

Source/missiles.cpp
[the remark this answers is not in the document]
AJenbo: done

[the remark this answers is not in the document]
ephphatha: Logged #2277 :D

Source/missiles.cpp
AJenbo: I think I will keep things as they are for now, but you are welcome to make this change.

julealgon: This is the first time I heard about Hadamard product. Interesting stuff.

I think it is a good idea to have abstractions like this to make the code simpler to reason about, so +1 from me.

PR #2276, opened 2021-07-01
Clean up objects var init

AJenbo:
This might still be a bit rough as I haven't had time to go over it myself yet. After this only monster.cpp remains.

Almost half of the changes are just indentation changes, so ignore white space is advised.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -282,25 +282,21 @@ _speech_id StoryText[3][3] = {
 
 void InitObjectGFX()
 {
-	bool fileload[56];
-	char filestr[32];
-	int i, j;
-
-	memset(fileload, 0, sizeof(fileload));
+	bool fileload[56] = {};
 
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -4645,13 +4542,13 @@ bool OperateFountains(int pnum, int i)
...
 		if (object[i]._oSelFlag == 0)
 			break;
-		prev = -1;
-		add = -1;
-		done = false;
-		cnt = 0;
+		int prev = -1;
+		int add = -1;
+		bool done = false;
+		int cnt = 0;
```
ephphatha, line 4551: Move these above the while loop on 4559? Looking at that loop it seems like `done` is unnecessary and `cnt` only exists to make it run twice... I've created https://github.com/diasurgical/devilutionX/issues/2278 for that refactoring opportunity.

ephphatha (dismissed):

PR #2282, opened 2021-07-02
Fix AddDoppelganger

AJenbo:
This was broken during a previous refactoring.

https://github.com/diasurgical/devilutionX/commit/21fdb0dcd951ab5adee57b5a212e2d86e6ac036f#diff-d46585089d315b53630b1035442b3d490d7b422cb05ceee76cf9041de3e39179L1251-L1252

The name is taken from the cleanup that was done to Devilution but hasn't been merged yet: https://github.com/diasurgical/devilution/blob/master/Source/monster.cpp#L1323

I noticed this while doing a general clean up so it's a bit tangled with with the clean up of the function it self.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1236,31 +1236,35 @@ int AddMonster(Point position, Direction dir, int mtype, bool InMap)
 	return -1;
 }
 
-void monster_43C785(int i)
+void AddDoppelganger(MonsterStruct &monster)
 {
-	int d, j, oi;
-	Point position = {};
+	if (monster.MType == nullptr) {
+		return;
+	}
 
-	if (monster[i].MType != nullptr) {
-		for (d = 0; d < 8; d++) {
-			position = monster[i].position.tile + monster[i]._mdir;
-			if (!SolidLoc(position)) {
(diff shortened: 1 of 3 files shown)


Source/monster.cpp
```diff
@@ -1236,31 +1236,35 @@ int AddMonster(Point position, Direction dir, int mtype, bool InMap)
...
-			if (!SolidLoc(position)) {
-				if (dPlayer[position.x][position.y] == 0 && dMonster[position.x][position.y] == 0) {
-					if (dObject[position.x][position.y] == 0)
-						break;
-					oi = dObject[position.x][position.y] > 0 ? dObject[position.x][position.y] - 1 : -(dObject[position.x][position.y] + 1);
-					if (!object[oi]._oSolidFlag)
-						break;
+	Point target = { 0, 0 };
+	for (int d = 0; d < 8; d++) {
+		const Point position = monster.position.tile + static_cast<Direction>(d);
```
AJenbo, line 1247: This is the offending line, instead of offsetting the monster in all possible directions it instead always put the new monster in front of the old one.

StephenCWills (approved):

PR #2287, opened 2021-07-03
SaveHelper: Write zeroes on `skip`

### LINKED ISSUE #2281: SaveHelper writes unitialised memory to the save file
**Important information**
Win10 / devilutionx from https://github.com/diasurgical/devilutionX/commit/f33f7ae7ebb8583a2bb270b4d27b3f39281a9929 

**Describe**
An optimisation to the SaveHelper constructor means that the backing buffer is left uninitialised. When skipping over unused blocks of memory using SaveHelper::skip the skipped buffer region ends up getting written to file with garbage values. If this save file is then loaded in an earlier version of devilutionx/devilution or vanilla Diablo/Hellfire that version of the game may end up reading the garbage values into areas where it expects structured data.

**To Reproduce**
Using #2270 as an example of the conditions that trigger this bug.
1. Build and run devilutionx from duplicate-dlight
2. Start a new game or load an existing save and navigate to town
3. Save the game and quit
4. Build and run devilutionx from https://github.com/diasurgical/devilutionX/commit/3f69faaecfa7069c17e51c05e84f6ddc27e56814
5. Load the saved game

In this situation the game crashes with an OOB memory read.

**Expected behavior**
The game loads and runs as normal

**Additional context**
This is more visible in debug builds due to memory watchdogs. I think this affects vanilla Diablo but due to how buggy the game already is it seems like the OS allows it to keep on trucking regardless.

glebm:
Instead of writing uninitialized memory to the file, which can lead to
issues with older game versions, we now write zeroes.

One alternative would be to have separate `skip` and `blank` functions
but I think we want the save files from the same state to be
byte-identical, so we always write zeroes.

Fixes #2281

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -5,8 +5,10 @@
  */
 #include "loadsave.h"
 
-#include <SDL.h>
 #include <climits>
+#include <cstring>
+
+#include <SDL.h>
 
 #include "automap.h"
 #include "codec.h"
@@ -156,6 +158,7 @@ public:
 
 	void skip(uint32_t len)
 	{
+		std::memset(&m_buffer_[m_cur_], 0, len);
 		m_cur_ += len;
 	}
 


ephphatha (approved):
Given the other bugs that pop up when reading back uninitialised values it's certainly safer to zero on skip :D

PR #2289, opened 2021-07-03
Fix `Rectangle::Contains` bounds check

glebm:


ephphatha (commented):
The <= conditions were implemented to match existing logic for attribute button hit tests.
https://github.com/diasurgical/devilutionX/commit/a37311ddc472c02c0584ad2b0d18ea8d90b88029

The logic in this PR is correct, but to keep the same behaviour for those call sites you'd be best to fix the off-by-one in size when declaring ChrBtnsRect[]

glebm:
Is there an off-by-one in size, or was the previous behaviour incorrect? Going to do some digging

glebm:
image

There is no off-by-1 in `ChrBtnsRect[]`, it's the previous hit test that was off-by-1.

ephphatha (approved):

julealgon:
This is an interesting one. As you noticed, I moved existing logic to create that method, and now you found out that the original logic was incorrect.

However, I think it is arguable that this fix is really "correct". By making the check for a point not consider the bottom/right border feels weird. I took a look at .NET's `System.Drawing.Rectangle`'s implementation and it does also ignore the borders.

I think this is weird. We are basically saying that "a point in the left border is in the rectangle, but a point in the right border is not". Feels very inconsistent to me.

I understand how this exclusive check simplifies tiled checks, but even then it seems wrong to me. I'd rather have "overlapping" rectangles and use check order to define which rectangle wins, or have the rectangle next to another one begin in X+1 coordinate instead of X.

glebm:
> I think this is weird. We are basically saying that "a point in the left border is in the rectangle, but a point in the right border is not". Feels very inconsistent to me.

That's not what the math says. Please try with an example to understand.

ephphatha:
I was wondering if it would be worth having an explicit constructor for `Rectangle(Point origin, Point extent)` which creates a rectangle with size large enough to encompass both points (`size = { extent.x - origin.x + 1, extent.y - origin.y + 1 }`). There are some places in the dungeon logic which check all tiles in a range xmin <= x <= xmax, ymin <= y <= ymax which could use that construct to get the same behaviour.

Making the Rectangle(Point, Point) constructor explicit will make the existing Rectangle(Point, Size) constructor the default for initialisations like:
https://github.com/diasurgical/devilutionX/blob/02d2754a68655207f3afc13e2e9a6f57ef435d97/Source/quests.cpp#L710

Cases like the Add*Objs functions can be easily converted to Rectangle(Point, Size) since they use `< max` checks:
https://github.com/diasurgical/devilutionX/blob/02d2754a68655207f3afc13e2e9a6f57ef435d97/Source/objects.cpp#L568-L581

The Rectangle(Point, Point) constructor would allow code like DRLG_AreaTrans/ListTrans to retain their existing meaning of defining a region based on two inclusive points:
https://github.com/diasurgical/devilutionX/blob/02d2754a68655207f3afc13e2e9a6f57ef435d97/Source/gendung.cpp#L356-L378
(This could use `DRLG_RectTrans({ { x1, y1 }, Point { x2, y2} });`)

And the actual RectTrans function could then be updated to operate on a rectangle using size-based logic instead of the current point to point loop.
https://github.com/diasurgical/devilutionX/blob/02d2754a68655207f3afc13e2e9a6f57ef435d97/Source/gendung.cpp#L341-L349

PR #2290, opened 2021-07-03
🚚 Rename `CelOutputBuf` to `Surface` and extract

glebm:


AJenbo (approved):

julealgon:
`Surface` seems a bit too generic to me. Maybe `DrawingSurface` would be more appropriate?

PR #2291, opened 2021-07-03
Align Life/Mana flask drawing routines

ephphatha:
These can be combined into a pair of functions now but I was gonna do that as a separate PR.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -626,9 +626,7 @@ void control_update_life_mana()
 {
 	auto &myPlayer = plr[myplr];
 
-	int maxMana = std::max(myPlayer._pMaxMana, 0);
-	int mana = std::max(myPlayer._pMana, 0);
-	myPlayer._pManaPer = maxMana != 0 ? ((double)mana / (double)maxMana * 80.0) : 0;
+	myPlayer.UpdateManaPercentage();
 	myPlayer.UpdateHitPointPercentage();
 }
 
@@ -636,16 +634,11 @@ void UpdateManaFlask(const Surface &out)
 {
 	auto &myPlayer = plr[myplr];
 
-	int maxMana = std::max(myPlayer._pMaxMana, 0);
-	int mana = std::max(myPlayer._pMana, 0);
-	int filled = maxMana != 0 ? ((double)mana / (double)maxMana * 80.0) : 0;
+	int filled = clamp(myPlayer.UpdateManaPercentage(), 0, 69);
 
-	myPlayer._pManaPer = filled;
-
-	filled = std::min(filled, 69);
-	if (filled != 69)
+	if (filled < 69)
 		DrawFlaskTop(out, { PANEL_X + 464, PANEL_Y }, pManaBuff, 16, 85 - filled);
-	if (filled != 0)
+	if (filled > 0)
 		DrawPanelBox(out, { 464, 85 - filled, 88, filled }, { PANEL_X + 464, PANEL_Y + 69 - filled });
 
 	DrawSpell(out);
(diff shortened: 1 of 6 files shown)


ephphatha:
Just noticed that the _pManaPer value isn't updating as expected... Need to fix that

Source/control.cpp
```diff
@@ -52,13 +52,19 @@ std::optional<CelSprite> pSpellCels;
 BYTE sgbNextTalkSave;
 BYTE sgbTalkSavePos;
 
+/**
+ * @brief Set if the life flask needs to be redrawn during next frame
+*/
```
AJenbo, line 57: The stars are not aligning:
```suggestion
/**
 * @brief Set if the life flask needs to be redrawn during next frame
 */
```

Source/control.cpp
```diff
@@ -578,81 +581,64 @@ static void DrawFlask(const CelOutputBuffer &out, const CelOutputBuffer &celBuf,
...
-	if (filled != 0)
-		DrawPanelBox(out, { 96, 85 - filled, 88, filled }, { 96 + PANEL_X, PANEL_Y + 69 - filled });
+	if (filled < 69)
+		DrawFlaskTop(out, { PANEL_X + offset, PANEL_Y }, sourceBuffer, 16, 85 - filled);
+	if (filled > 0)
+		DrawPanelBox(out, { offset, 85 - filled, 88, filled }, { PANEL_X + offset, PANEL_Y + 69 - filled });
 }
 
-void DrawManaFlask(const CelOutputBuffer &out)
+// the Draw*Flask* functions use precalculated value of HP/ManaPer from an earlier call to control_update_life_mana()
```
AJenbo, line 613:
```suggestion
/**
 * the Draw*Flask* functions use precalculated value of HP/ManaPer from an earlier call to control_update_life_mana()
 */
```

Source/player.h
```diff
@@ -418,11 +418,22 @@ struct PlayerStruct {
...
+			_pHPPer = clamp(_pHitPoints * 80 / _pMaxHP, 0, 80); // hp should never be greater than maxHP but just in case
 		}
 
 		return _pHPPer;
 	}
+
+	int UpdateManaPercentage()
+	{
+		if (_pMaxMana <= 0) {
+			_pManaPer = 0;
```
AJenbo, line 430: I looked at the Devilution code and _pManaPer is not relevant so we can stript it from the struct and skip it in the saves. Would be nice to get that done to clean this up as well.

ephphatha: I agree they don't need to be saved, vanilla recalculates them when setting up the game state (and on every frame) so that won't cause issues in old versions. I'd like to keep the variables in the struct though as they serve a useful purpose for the UI code. The HP and Mana values change relatively infrequently so there's no real need to redo this calculation 3 times each every frame. Ideally this is done when changing the value of `_Mana`/`_MaxMana`/`_Life`/`_MaxLife` and the draw routines can just rely on the `*Per` value being accurate

AJenbo: Go for it :)

Source/control.cpp
ephphatha: I wasn't sure if doxygen comments were appropriate here and it turns out VS picks this up instead of the header comment for the next function. I've updated the comments in the header instead to address this.

Source/control.cpp
ephphatha: Thanks for the pickup, Visual studio is a bit inconsistent about applying style rules and doesn't handle /** comments well, I'll see if I can fix it.

Source/control.cpp
AJenbo: Oh I thought this was a header comment for the next function, my bad :)

PR #2298, opened 2021-07-04
💚 Fix remaning compiler warnings for GCC and CLang

AJenbo:


Source/itemdat.h
```diff
@@ -13,7 +13,7 @@
 namespace devilution {
 
 /** @todo add missing values and apply */
-enum _item_indexes : uint16_t {
+enum _item_indexes : int16_t {
```
ephphatha, line 16: `_item_indexes` type variables are occasionally used as array indexes which makes me think it'd be best to keep this as an unsigned type, but at the same time this type is often used interchangeably with `int32_t` in functions like `IsItemAvailable()`. I've logged #2300 to cover that investigation.

ephphatha (approved):

PR #2301, opened 2021-07-04
fix: DrawMissile code doesn't handle cases where dMissile[x][y] == 0

### LINKED ISSUE #2286: DrawMissile code doesn't handle cases where dMissile[x][y] == 0
**Important information**
Operating system / DevilutionX master

**Describe**
The `dMissile[]` array uses `0` and `-1` as flag values (FREE or INVALID from what I can tell) expecting actual missile indexes to be 1-based values. This function doesn't guard against the 0 case so can read before the start of the `missile[]` array and pass garbage data to `DrawMissilePrivate()`.
https://github.com/diasurgical/devilutionX/blob/d7a86330c6e152a287aa245a1a92eea37358ce3f/Source/scrollrt.cpp#L284-L291

**To Reproduce**
1. Same reproduction as #2285 

**Expected behavior**
`DrawMissile()` should check the value is positive (`>0`) before trying to use it.

rbrishabh:
Fixes #2286 
Hey @ephphatha, how does this look?
let me know if this needs any changes! Thanks

### DIFF
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -284,7 +284,7 @@ void DrawMissile(const Surface &out, int x, int y, int sx, int sy, bool pre)
 	if ((dFlags[x][y] & BFLAG_MISSILE) == 0)
 		return;
 
-	if (dMissile[x][y] != -1) {
+	if (dMissile[x][y] > 0) {
 		m = &missile[dMissile[x][y] - 1];
 		DrawMissilePrivate(out, m, sx, sy, pre);
 		return;


ephphatha (approved):
👍

Source/scrollrt.cpp
```diff
@@ -284,7 +284,7 @@ void DrawMissile(const Surface &out, int x, int y, int sx, int sy, bool pre)
 	if ((dFlags[x][y] & BFLAG_MISSILE) == 0)
 		return;
 
-	if (dMissile[x][y] != -1) {
+	if (dMissile[x][y] > 0) {
```
julealgon, line 287: What does `0` represent? Why do we treat both -1 and 0 the same? Shouldn't it really only be `-1`? Would it make more sense to use `-1` in the places where `0` is used?

julealgon: I see most of the info is in the linked issue. Should've read that before asking.

rbrishabh: no problem!

PR #2303, opened 2021-07-04
Main menu: make title char *

vlolteanu:


AJenbo:
What's the rational behind this?

vlolteanu:
The two instances of "title" always contain stuff obtained from LanguageTranslate, or an empty string. It makes no sense to sense to strcpy an immutable string, hence const char * is appropriate.

It's similar to what I've done here: https://github.com/diasurgical/devilutionX/pull/1750

AJenbo (approved):

Source/DiabloUI/selgame.cpp
```diff
@@ -35,7 +35,7 @@ extern int provider;
 
 namespace {
 
-char title[32];
+const char *title = "";
```
julealgon, line 38: Can't we use `string` instead?

vlolteanu: That would just make extra copies. If you really want to use some something from the stl, `string_view` is the closest to what we're doing.

julealgon: Can we use `string_view` then?

AJenbo: Eventually all of it should be moved to `string`/`string_view`, it's already what is being used internally in the translation handling as well as the `fmt` string building. Most of the warnings we get in MSVC is from the use of `char*` for string handling. I guess the next step in this process would be to adapt the string rendering code to expect `std::string`, and then take out all the in between conversions as a final stage.

jmechnich:
This PR broke menu titles (see #2373).

PR #2304, opened 2021-07-04
Clean up missiles.cpp

AJenbo:
This had a few interesting casts and alignments related to the CrawlTable usage so might be best to give it a review :)


Source/missiles.cpp
```diff
@@ -3622,16 +3594,16 @@ void MI_Reflect(int i)
...
-	int k = CrawlNum[3] + 1;
+	int8_t src = missile[i]._micaster;
 	uint8_t lvl = src > 0 ? plr[src]._pLevel : currlevel;
 	int dmg = 16 * (GenerateRndSum(10, 2) + lvl + 2) / 2;
 
-	for (int j = CrawlTable[b]; j > 0; j--, k += 2) {
-		int tx = missile[i]._miVar1 + CrawlTable[k - 1];
-		int ty = missile[i]._miVar2 + CrawlTable[k];
+	int k = CrawlNum[3];
+	int ck = k + 1;
```
AJenbo, line 3603: This seams to be off by 1 compared to all other usages of `CrawlTable`.

AJenbo: Turned out to be a bug in Devilution: https://github.com/diasurgical/devilution/pull/2237

StephenCWills (approved):
This `CrawlTable` pattern is used so many times. It would be nice to encapsulate it, but it looks like that would be somewhat complicated. I'm glad to have a more homogeneous pattern.

AJenbo:
Yeah that was the idea here, tried to make even the temp var names identical to make it easier on the person wanting to do this refactoring :)

PR #2307, opened 2021-07-04
clang-tidy/Android Stuidio cleanups 5

AJenbo:


Source/quests.cpp
```diff
@@ -647,7 +647,7 @@ void ResyncQuests()
 			ObjChangeMapResync(setpc_x, setpc_y, (setpc_w / 2) + setpc_x + 2, (setpc_h / 2) + setpc_y - 2);
 			for (int i = 0; i < nobjects; i++)
 				SyncObjectAnim(objectactive[i]);
-			int tren = TransVal;
+			int8_t tren = TransVal;
```
ephphatha, line 650:
```suggestion
			auto tren = TransVal;
```
TransVal is currently a char type, given this temp is purely to save and restore the initial state of TransVal it might as well use an inferred type.

ephphatha (dismissed):

Source/quests.cpp
AJenbo: Really good point 🤔

Source/palette.cpp
```diff
@@ -291,11 +291,13 @@ void PaletteFadeOut(int fr)
  */
 static void CycleColors(int from, int to)
 {
-	SDL_Color col = system_palette[from];
-	for (int i = from; i < to; i++) {
-		system_palette[i] = system_palette[i + 1];
+	{
+		SDL_Color col = system_palette[from];
```
glebm, line 295: This function can be replaced with:

```c++
std::rotate(&system_palette[from],  &system_palette[from + 1], &system_palette[to]);
```

Source/palette.cpp
```diff
@@ -291,11 +291,13 @@ void PaletteFadeOut(int fr)
  */
 static void CycleColors(int from, int to)
 {
-	SDL_Color col = system_palette[from];
-	for (int i = from; i < to; i++) {
-		system_palette[i] = system_palette[i + 1];
+	{
```
qndel, line 294: the hell is this?

AJenbo: a anon scope to avoid overlapping of the two variables called `col`

PR #2309, opened 2021-07-05
Introduce template form of skip to represent deprecated variables

ephphatha:
Also ensuring that size_t is used consistently for size/offset types.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -163,7 +163,7 @@ public:
 	}
 
 	template<typename T>
-	constexpr void Skip(void)
+	constexpr void Skip()
 	{
 		Skip(sizeof(T));
 	}
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -1510,7 +1510,6 @@ static void SavePlayer(SaveHelper *file, int p)
 	file->WriteLE<int32_t>(player.position.offset2.deltaX);
 	file->WriteLE<int32_t>(player.position.offset2.deltaY);
 	file->Skip<int32_t>(); // Skip _pVar8
-
 	for (uint8_t i = 0; i < giNumberOfLevels; i++)
 		file->WriteLE<uint8_t>(player._pLvlVisited[i] ? 1 : 0);
 	for (uint8_t i = 0; i < giNumberOfLevels; i++)
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -1510,7 +1510,7 @@ static void SavePlayer(SaveHelper *file, int p)
 	file->WriteLE<int32_t>(player._pVar5);
 	file->WriteLE<int32_t>(player.position.offset2.deltaX);
 	file->WriteLE<int32_t>(player.position.offset2.deltaY);
-	file->WriteLE<int32_t>(0); // Write actionFrame for vanilla compatibility
+	file->Skip<int32_t>(); // Skip _pVar8
 
 	for (uint8_t i = 0; i < giNumberOfLevels; i++)
 		file->WriteLE<uint8_t>(player._pLvlVisited[i] ? 1 : 0);
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -70,7 +70,7 @@ T SwapBE(T in)
 
 class LoadHelper {
 	std::unique_ptr<byte[]> m_buffer_;
-	uint32_t m_cur_ = 0;
+	size_t m_cur_ = 0;
 	size_t m_size_;
 
 	template <class T>
@@ -93,13 +93,19 @@ public:
 		m_buffer_ = pfile_read(szFileName, &m_size_);
 	}
 
-	bool IsValid(uint32_t size = 1)
+	bool IsValid(size_t size = 1)
 	{
 		return m_buffer_ != nullptr
 		    && m_size_ >= (m_cur_ + size);
 	}
 
-	void Skip(uint32_t size)
+	template<typename T>
+	constexpr void Skip(void)
+	{
+		Skip(sizeof(T));
+	}
+
+	void Skip(size_t size)
 	{
 		m_cur_ += size;
 	}
@@ -139,8 +145,8 @@ public:
 class SaveHelper {
 	const char *m_szFileName_;
 	std::unique_ptr<byte[]> m_buffer_;
-	uint32_t m_cur_ = 0;
-	uint32_t m_capacity_;
+	size_t m_cur_ = 0;
+	size_t m_capacity_;
 
 public:
 	SaveHelper(const char *szFileName, size_t bufferLen)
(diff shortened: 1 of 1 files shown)


Source/loadsave.cpp
```diff
@@ -1666,10 +1700,10 @@ static void SaveMonster(SaveHelper *file, int i)
 	file->Skip(1); // Alignment
 	file->WriteLE<uint16_t>(pMonster->mExp);
 
-	file->WriteLE<uint8_t>(pMonster->mHit < UINT8_MAX ? pMonster->mHit : UINT8_MAX); // For backwards compatibility
+	file->WriteLE<uint8_t>(std::min<uint16_t>(pMonster->mHit, std::numeric_limits<uint8_t>::max())); // For backwards compatibility
 	file->WriteLE<uint8_t>(pMonster->mMinDamage);
 	file->WriteLE<uint8_t>(pMonster->mMaxDamage);
-	file->WriteLE<uint8_t>(pMonster->mHit2 < UINT8_MAX ? pMonster->mHit2 : UINT8_MAX); // For backwards compatibility
+	file->WriteLE<uint8_t>(std::min<uint16_t>(pMonster->mHit2, std::numeric_limits<uint8_t>::max())); // For backwards compatibility
```
ephphatha, line 1706: Not happy with this but it avoids comparisons between different types and the downcast is recognised as non-truncating in msvc at least.

Source/loadsave.cpp
```diff
@@ -1285,8 +1297,10 @@ static void SaveItem(SaveHelper *file, ItemStruct *pItem)
 	}
 
 	file->WriteLE<int32_t>(pItem->_iSeed);
-	file->WriteLE<int16_t>(pItem->_iCreateInfo);
-	file->Skip(2); // Alignment
+
+	file->WriteLE<uint16_t>(pItem->_iCreateInfo);
+	file->Skip(2); // Alignment, _iCreateInfo was written as a 4 byte LE value but only requires 2 bytes
```
AJenbo, line 1302: I don't think this is why alignment happen for structs https://en.m.wikipedia.org/wiki/Data_structure_alignment

Maybe you are not aware but the original save files would mostly just memcpy structs to the file buffer

ephphatha: This was a bit of a guess based on the comments in devilution for certain structs:
https://github.com/diasurgical/devilution/blob/e426c38adb5df05b8e2c02a607529bdc913c281b/structs.h#L241-L247

Situations like this looked more like type conversion/aliasing? compared to other instances where a group of short/char variables needed padding to get back to 4 byte alignment.

AJenbo: I think that comment is wrong, the compiler never was able to do this when we worked with it and this looks more like just regular padding.

ephphatha: Ahh, in that case I'll revert those comment changes. The other comments I updated with `packed _var, _var, ...` were also more helpers for myself along the same lines so I'll revert those as well.

Source/loadsave.cpp
```diff
@@ -1497,17 +1510,16 @@ static void SavePlayer(SaveHelper *file, int p)
 	file->WriteLE<int32_t>(player._pVar5);
 	file->WriteLE<int32_t>(player.position.offset2.deltaX);
 	file->WriteLE<int32_t>(player.position.offset2.deltaY);
-	// Write actionFrame for vanilla compatibility
-	file->WriteLE<int32_t>(0);
+	file->WriteLE<int32_t>(0); // Write actionFrame for vanilla compatibility
```
AJenbo, line 1513:
```suggestion
	file->Skip<int32_t>(); // Skip _pVar5
```

Source/loadsave.cpp
```diff
@@ -1497,17 +1510,16 @@ static void SavePlayer(SaveHelper *file, int p)
...
 	for (uint8_t i = 0; i < giNumberOfLevels; i++)
 		file->WriteLE<uint8_t>(player._pLvlVisited[i] ? 1 : 0);
 	for (uint8_t i = 0; i < giNumberOfLevels; i++)
 		file->WriteLE<uint8_t>(player._pSLvlVisited[i] ? 1 : 0); // only 10 used
 
 	file->Skip(2); // Alignment
 
-	// Write _pGFXLoad for vanilla compatibility
-	file->WriteLE<int32_t>(0);
+	file->WriteLE<int32_t>(0); // Write _pGFXLoad for vanilla compatibility
```
AJenbo, line 1522:
```suggestion
	file->Skip<int32_t>(); // Skip _pGFXLoad
```

Source/loadsave.cpp
```diff
@@ -1639,8 +1651,7 @@ static void SaveMonster(SaveHelper *file, int i)
 	file->WriteLE<int32_t>(pMonster->position.temp.y);
 	file->WriteLE<int32_t>(pMonster->position.offset2.deltaX);
 	file->WriteLE<int32_t>(pMonster->position.offset2.deltaY);
-	// Write actionFrame for vanilla compatibility
-	file->WriteLE<int32_t>(0);
+	file->WriteLE<int32_t>(0); // Write actionFrame for vanilla compatibility
```
AJenbo, line 1654:
```suggestion
	file->Skip<int32_t>(); // Skip _mVar8
```

Source/loadsave.cpp
```diff
@@ -150,13 +156,19 @@ class SaveHelper {
...
-	bool IsValid(uint32_t len = 1)
+	bool IsValid(size_t len = 1)
 	{
 		return m_buffer_ != nullptr
 		    && m_capacity_ >= (m_cur_ + len);
 	}
 
-	void Skip(uint32_t len)
+	template<typename T>
+	constexpr void Skip(void)
```
glebm, line 166: The `void` is unnecessary in C++

```suggestion
	constexpr void Skip()
```

(same above in LoadHelper)

Source/loadsave.cpp
```diff
@@ -1497,17 +1510,16 @@ static void SavePlayer(SaveHelper *file, int p)
 	file->WriteLE<int32_t>(player._pVar5);
 	file->WriteLE<int32_t>(player.position.offset2.deltaX);
 	file->WriteLE<int32_t>(player.position.offset2.deltaY);
-	// Write actionFrame for vanilla compatibility
```
qndel, line 1500: what happened to this?

Source/loadsave.cpp
```diff
@@ -1298,7 +1310,8 @@ static void SaveItem(SaveHelper *file, ItemStruct *pItem)
 	file->WriteLE<int32_t>(ItemAnimWidth);
 	// write _iAnimWidth2 for vanilla compatibility
 	file->WriteLE<int32_t>(CalculateWidth2(ItemAnimWidth));
-	file->Skip(4); // Unused since 1.02
+	file->Skip<uint32_t>(); // _delFlag, unused since 1.02
+
```
AJenbo, line 1314:
```suggestion
```

Source/loadsave.cpp
```diff
@@ -1497,17 +1510,16 @@ static void SavePlayer(SaveHelper *file, int p)
 	file->WriteLE<int32_t>(player._pVar5);
 	file->WriteLE<int32_t>(player.position.offset2.deltaX);
 	file->WriteLE<int32_t>(player.position.offset2.deltaY);
-	// Write actionFrame for vanilla compatibility
-	file->WriteLE<int32_t>(0);
+	file->Skip<int32_t>(); // Skip _pVar8
+
```
AJenbo, line 1513:
```suggestion
```

AJenbo (approved):

PR #2310, opened 2021-07-05
Fix MI_Rhino don't update dMonster correctly

obligaron:
58cddd6e changed MI_Rhino (Charging monsters) so that the old dMonster variable aren't reset correctly.
That means the game thinks the tile is occupied with the monsters.
This results in players and monsters can't move to them. If the cursor hovers over this not updated tiles, the monster information is (incorrectly) shown.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -4269,27 +4269,30 @@ void MI_Rhino(int i)
 		return;
 	}
 	GetMissilePos(i);
-	Point position = missile[i].position.tile;
-	dMonster[position.x][position.y] = 0;
+	Point prevPos = missile[i].position.tile;
+	Point newPosSnake;
+	dMonster[prevPos.x][prevPos.y] = 0;
 	if (monster[monst]._mAi == AI_SNAKE) {
 		missile[i].position.traveled += missile[i].position.velocity * 2;
 		GetMissilePos(i);
+		newPosSnake = missile[i].position.tile;
 		missile[i].position.traveled -= missile[i].position.velocity;
 	} else {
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -4269,27 +4269,30 @@ void MI_Rhino(int i)
...
+		newPosSnake = missile[i].position.tile;
 		missile[i].position.traveled -= missile[i].position.velocity;
 	} else {
 		missile[i].position.traveled += missile[i].position.velocity;
 	}
 	GetMissilePos(i);
-	if (!PosOkMonst(monst, position) || (monster[monst]._mAi == AI_SNAKE && !PosOkMonst(monst, position))) {
-		MissToMonst(i, position);
+	Point newPos = missile[i].position.tile;
+	if (!PosOkMonst(monst, newPos) || (monster[monst]._mAi == AI_SNAKE && !PosOkMonst(monst, newPosSnake))) {
```
ephphatha, line 4285: Good pickup, I should've realised the side effects in `GetMissilePos()` modify `missile[i].position.tile` and make the order of reads important -_- (plus the special case needed for AI_SNAKE checking two ticks/moves ahead)

ephphatha (approved):

Source/missiles.cpp
obligaron: That's not so easy to detected.
The Method `GetMissilePos` is really misleading.
We should change it to something like `UpdateMissilePos` (not part of this pr).

AJenbo:
Thanks I noticed something fishy was going on but didn't have time to investigate :)

PR #2311, opened 2021-07-05
Fix C4244 - type conversion warnings in MSVC

ephphatha:
Some of the c-style casts appeared to be unnecessary, they were either casting variables to the same type or forcing a narrower conversion than numeric conversion rules would have defaulted to. I've only checked this on MSVC, not sure if this was needed for a specific compiler.

### DIFF
--- a/Source/DiabloUI/ttf_render_wrapped.cpp
+++ b/Source/DiabloUI/ttf_render_wrapped.cpp
@@ -142,7 +142,7 @@ SDL_Surface *RenderUTF8_Solid_Wrapped(TTF_Font *font, const char *text, SDL_Colo
 	SDLC_SetColorKey(textbuf, 0);
 
 	// Reduced space between lines to roughly match Diablo.
-	const int lineskip = 0.7 * TTF_FontLineSkip(font);
+	const int lineskip = TTF_FontLineSkip(font) * 7 / 10; // avoids forced int > float > int conversion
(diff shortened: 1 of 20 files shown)


AJenbo:
Since it's C-style it's probably inherited from Devilution. At an early point many variables where incorrect and got cast to a few different things along the way. Often we wouldn't notice casts that where to the same type as what we had eventually corrected the variable to.

Source/DiabloUI/ttf_render_wrapped.cpp
```diff
@@ -142,7 +142,7 @@ SDL_Surface *RenderUTF8_Solid_Wrapped(TTF_Font *font, const char *text, SDL_Colo
 	SDLC_SetColorKey(textbuf, 0);
 
 	// Reduced space between lines to roughly match Diablo.
-	const int lineskip = 0.7 * TTF_FontLineSkip(font);
+	const int lineskip = TTF_FontLineSkip(font) * 7 / 10; // avoids forced int > float > int conversion
```
AJenbo, line 145: :+1:

ephphatha:
There are only 4 lines left with this warning and it's all related to the vector scaling logic in `GetMissileVel()`. Pretty sure that can be reordered to simplify it a bit but I need to spend a bit of time working through it to be sure.

AJenbo (approved):

PR #2313, opened 2021-07-05
missiles.cpp: Rename GetMissilePos/MissileVel/VileMIssPos -> UpdateMissilePos/MissileVel/VileMIssPos

obligaron:
Rename methods with "Get"-Prefix and side-effects to "Update"-Prefix.
Follow-up from #2310


julealgon (commented):
Nice. It is such a terrible bad practice to have `void GetX` methods, glad to see those gone.

Issue #2278, opened 2021-07-02
Refactor OperateFountains to clarify the stat modification logic of TEARFTN


ephphatha:
This case in OperateFountains removes a point from an attribute at random and adds a point to a different attribute at random.
https://github.com/diasurgical/devilutionX/blob/3f69faaecfa7069c17e51c05e84f6ddc27e56814/Source/objects.cpp#L4648-L4692

The stat modification logic can be split into a function, then the loop can be refactored so it's expressed as the following steps:
1. Select a random stat
2. Remove an attribute point from that stat
3. Call GenerateRnd in a loop until a different stat is selected
4. Add an attribute point to that stat

vlolteanu:
> 3\. Call GenerateRnd in a loop until a different stat is selected

Selecting a second stat can be done with a single call to `GenerateRnd` without any bias: `rnd = GenerateRnd(3); if (rnd >= prev) rnd++`. That would certainly alter the game's behavior, but would it break anything?

However, generating random stuff in a loop doesn't offer any kind of guarantees on how many iterations are required.

AJenbo:
> Selecting a second stat can be done with a single call to `GenerateRnd` without any bias: `rnd = GenerateRnd(3); if (rnd >= prev) rnd++`. That would certainly alter the game's behavior, but would it break anything?

This should be safe as it happens at game time where RNG sequence doesn't have to be locked to the original implementation.

ephphatha:
> Selecting a second stat can be done with a single call to `GenerateRnd` without any bias: `rnd = GenerateRnd(3); if (rnd >= prev) rnd++`. That would certainly alter the game's behavior, but would it break anything?

`GenerateRnd()` with a non-power of two argument has a chance at returning an unexpected result (see #2206). This could result in a point being removed from a stat and not being added to another stat. It's not a major issue though, it can be avoided by checking for negative results or by using a safer implementation like the functions provided in #2084 or #2261

ephphatha:
This was closed by #2314/#2318 :D

PR #2314, opened 2021-07-05
Refactor tear fountain logic; remove GenerateRnd loop

vlolteanu:
Addresses https://github.com/diasurgical/devilutionX/issues/2278

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -4564,33 +4564,30 @@ bool OperateFountains(int pnum, int i)
 			return false;
 		if (pnum != myplr)
 			return false;
-		int prev = -1;
-		int add = -1;
-		int cnt = 0;
-		while (true) {
-			int rnd = GenerateRnd(4);
-			if (rnd != prev) {
-				switch (rnd) {
-				case 0:
-					ModifyPlrStr(pnum, add);
-					break;
-				case 1:
-					ModifyPlrMag(pnum, add);
-					break;
-				case 2:
-					ModifyPlrDex(pnum, add);
-					break;
-				case 3:
-					ModifyPlrVit(pnum, add);
-					break;
-				}
-				prev = rnd;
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -4564,33 +4564,33 @@ bool OperateFountains(int pnum, int i)
...
-			}
-			if (cnt > 1)
+
+		int fromStat = GenerateRnd(4);
+		int toStat = [&]() {
+			int rnd = GenerateRnd(3);
+			if (rnd >= fromStat)
+				rnd++;
+			return rnd;
+		}();
```
AJenbo, line 4574: What is the lambda giving us here compared to just dowing it directly?
```suggestion
		int toStat = GenerateRnd(3);
		if (toStat >= fromStat)
			toStat++;
```

vlolteanu: Slightly better indication of intent, but otherwise, nothing.

AJenbo: Imo it adds a significant number of cryptic symbols that some new comers may not be familiar with, for something like level objects I think it's important that the code be very approachable for novices.

Source/objects.cpp
```diff
@@ -4564,33 +4564,33 @@ bool OperateFountains(int pnum, int i)
...
+
+		std::pair<int, int> alterations[] = { { fromStat, -1 }, { toStat, 1 } };
+		for (auto i : alterations) {
```
AJenbo, line 4577:
```suggestion
		for (auto alteration : alterations) {
```

AJenbo (approved):

Source/objects.cpp
vlolteanu: Alright.

I prefer the lambda because toStat >= fromStat isn't an exceptional condition.

PR #2318, opened 2021-07-05
Check for negative random values in tear fountain logic

vlolteanu:
See https://github.com/diasurgical/devilutionX/issues/2278#issuecomment-874363045

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -4566,7 +4566,7 @@ bool OperateFountains(int pnum, int i)
 			return false;
 
 		int fromStat = GenerateRnd(4);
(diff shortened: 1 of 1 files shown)


ephphatha (approved):

AJenbo:
Will the random value be in the expected range with this?

vlolteanu:
Yes. x % y == -(-x % y).

ephphatha:
Yes, GenerateRnd(3) returns -2, 0, 1, or 2, and abs handles the -2 case. There's slight bias due to the modulo done in GenerateRnd() but that needs an alternate generator to avoid.

Issue #2260, opened 2021-06-28
Refactor missiles.cpp to use Direction enum instead of magic Point/Displacement constants


ephphatha:
There are multiple places in missiles.cpp which iterate through tiles adjacent to a given location.

https://github.com/diasurgical/devilutionX/blob/5537fe748e63b06ed2e474f7dea6ebd1216fa3d3/Source/missiles.cpp#L1634
https://github.com/diasurgical/devilutionX/blob/5537fe748e63b06ed2e474f7dea6ebd1216fa3d3/Source/missiles.cpp#L3417
https://github.com/diasurgical/devilutionX/blob/5537fe748e63b06ed2e474f7dea6ebd1216fa3d3/Source/missiles.cpp#L3995
https://github.com/diasurgical/devilutionX/blob/5537fe748e63b06ed2e474f7dea6ebd1216fa3d3/Source/missiles.cpp#L4017
https://github.com/diasurgical/devilutionX/blob/5537fe748e63b06ed2e474f7dea6ebd1216fa3d3/Source/missiles.cpp#L4772

These lists could be directly replaced with a list of the appropriate `Direction` enum constants. This would retain the same behaviour as existing code while making the intent clearer.

ephphatha:
There is similar code in player.cpp and towners.cpp (probably plenty of other places). If you were feeling keen and wanted to clean up other files I'd recommend limiting PRs to a single file per commit to make it easier to review 😃

agris-codes:
Does “good first issue” mean a good issue for a newcomer to the project to tackle, rather than a good writeup of the issue by the poster?

AJenbo:
> Does “good first issue” mean a good issue for a newcomer to the project to tackle, rather than a good writeup of the issue by the poster?

yes https://goodfirstissue.dev/language/cplusplus

agris-codes:
T

> > Does “good first issue” mean a good issue for a newcomer to the project to tackle, rather than a good writeup of the issue by the poster?


> > Does “good first issue” mean a good issue for a newcomer to the project to tackle, rather than a good writeup of the issue by the poster?
> 
> yes https://goodfirstissue.dev/language/cplusplus

Thanks, I’ll tackle this issue then if the turn-around time required isn’t < 1 week.

julealgon:
I imagine some instances could just turn into:
```cpp
for (auto direction in enum_values<Direction>())
    ...
```
?

[...]

PR #2319, opened 2021-07-06
Cleanups

AJenbo:
This needs to be split up the commit first, but didn't have time for it. For now this PR is just to have some test builds to see how things stand.

This covers all files directly in `Source` and all files in the `controls` and `DiabloUI` sub folders.

Depends on https://github.com/diasurgical/devilutionX/pull/2321

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -45,7 +45,7 @@ bool InGameMenu()
 
 namespace {
 
-int slot = SLOTXY_INV_FIRST;
+int Slot = SLOTXY_INV_FIRST;
 
 /**
  * Number of angles to turn to face the coordinate
@@ -257,9 +257,9 @@ void FindMeleeTarget()
 		SearchNode node = queue.front();
 		queue.pop_front();
 
-		for (int i = 0; i < 8; i++) {
-			const int dx = node.x + pathxdir[i];
-			const int dy = node.y + pathydir[i];
+		for (auto pathDir : PathDirs) {
+			const int dx = node.x + pathDir.deltaX;
+			const int dy = node.y + pathDir.deltaY;
 
 			if (visited[dx][dy])
 				continue; // already visisted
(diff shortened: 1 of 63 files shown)


Source/path.cpp
```diff
@@ -30,8 +30,18 @@ PATHNODE *pnode_tblptr[MAXPATHNODES];
...
+	{ -1, -1 },
+	{ -1,  1 },
+	{  1, -1 },
+	{  1,  1 },
+	{ -1,  0 },
+	{  0, -1 },
+	{  1,  0 },
+	{  0,  1 },
+	// clang-format on
+};
```
ephphatha, line 44:
```suggestion
const Direction PathDirs[8] = {
	// clang-format off
	Direction::DIR_N,
	Direction::DIR_W,
	Direction::DIR_E,
	Direction::DIR_S,
	Direction::DIR_NW,
	Direction::DIR_NE,
	Direction::DIR_SE,
	Direction::DIR_SW,
	// clang-format on
};
```
This is an array of directions, it can use the Direction enum values.

Source/controls/plrctrls.cpp
```diff
@@ -257,9 +257,9 @@ void FindMeleeTarget()
 		SearchNode node = queue.front();
 		queue.pop_front();
 
-		for (int i = 0; i < 8; i++) {
-			const int dx = node.x + pathxdir[i];
-			const int dy = node.y + pathydir[i];
+		for (auto pathDir : PathDirs) {
+			const int dx = node.x + pathDir.deltaX;
+			const int dy = node.y + pathDir.deltaY;
```
ephphatha, line 262:
```suggestion
			const auto testPosition = node + pathDir;
			const int dx = testPosition.x;
			const int dy = testPosition.y;
```
To handle changing the backing type of PathDirs

ephphatha (approved):
Looks fine, only a couple of suggestions where I thought there was a mismatch between the name and purpose

Source/path.cpp
AJenbo: I went with this to avoid the issue in your next comment :)

Since you approved the PR and it was set to auto-merge it's already been merged, but you are welcome to make the change if you prefer.

Source/controls/plrctrls.cpp
AJenbo: PathNode could also be changed to be composed of a Point to allow for this directly.

Source/dvlnet/base.h
```diff
@@ -71,13 +71,13 @@ class base : public abstract_net {
 
 	std::unique_ptr<packet_factory> pktfty;
 
-	void handle_accept(packet &pkt);
-	void recv_local(packet &pkt);
-	void run_event_handler(_SNETEVENT &ev);
+	void HandleAccept(packet &pkt);
+	void RecvLocal(packet &pkt);
```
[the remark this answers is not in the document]
AJenbo: Yeah, this was just CLang correcting the style and I didn't have a good idea for a name at the time.

Source/path.cpp
ephphatha: Wasn't worth holding up this PR over, and it's easier to implement in isolation anyway :). I'll leave it for #2260 or a followup story

PR #2320, opened 2021-07-06
Cache GitHub Action build folders

AJenbo:
@glebm does it make sens to cache all of `/build` or should I limit it to `/build/_deps` or something else ?


glebm:
Yeah, makes sense to cache the whole thing 👍

glebm (approved):

Issue #1561, opened 2021-04-17
Ingame settings menu


AJenbo:
To help with managing the ever-growing list of game options we would like to provide an advanced options menu based on the Help/Shop interface.

Here is a basic mockup of what such a menu could look like:
![image](https://user-images.githubusercontent.com/204594/115096584-1a13a200-9f26-11eb-8f92-3605ea239846.png)

I have not thought of how to indicate that some options will only take effect in new games or by restarting the game, feel free to be creative or have a debate about it.

I'm thinking that this menu will be accessed from `menu` -> `options` -> `advanced` and will replace the "game speed" option.

galaxyhaxz:
Great idea, some games that require restart will usually have text that comes up when changing an option and then when leaving the menu it recreates the window context. This works perfectly fine on desktops, but doubt it would work well on game consoles.

AJenbo:
Recreating the window probably won't work for enabling all quests :)

galaxyhaxz:
Options like those should be added to the UI when creating a game IMO. Just like selecting difficulty.

AJenbo:
Unfortunately, the game creates UI isn't good for handling multiple options. Currently you have to mandatorily go over each of them one at a time before you are allowed to start the game. So for the first iteration, I think it's acceptable to have all options listed in this menu. Then later we can create a more flexible Create Game screen that supports multiple options on a single screen (proposals wellcome).

galaxyhaxz:
I've had this implemented in my own UI for almost a year now ;P
![Untitled](https://user-images.githubusercontent.com/65138215/115097357-33e5c280-9eef-11eb-9fb6-efe943c63e49.png)

Sircrashalot:
Nice. Will you also add the option for attacking with just one click? Although i loved to spam click back in the good old days... i am just too old for that now :D

julealgon:
> Will you also add the option for attacking with just one click? Although i loved to spam click back in the good old days... i am just too old for that now :D

@Sircrashalot this is not the place to ask about settings themselves. This issue is to discuss how the existing settings are displayed in-game. If you want to discuss actual settings proposals, open another issue or start a discussion.

Having said that, specifically talking about "holding mouse button to attack constantly", it is already in development in some form, so the feature is definitely being considered.

AJenbo:
An even better place is the chat https://discord.gg/YQKCAYQ

PR #2324, opened 2021-07-07
Only save diablo.ini if there are changes

obligaron:
I often change the game speed directly in the diablo.ini
But every time I leave the game my editor reminds me that the file is changed, cause we save the ini at every exit.
With this pr the ini is only saved if there are changes.

Tested:
- No .ini exists, default ini is created
- New entry doesn't exist, ini should be saved with default value

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -59,6 +59,41 @@ CSimpleIni &GetIni()
 	return ini;
 }
 
+static bool IniChanged = false;
+
+/**
+ * @brief Checks if a ini entry is changed by comparing value before and after
+ */
+class IniChangedChecker {
+public:
+	IniChangedChecker(const char *sectionName, const char *keyName)
+	{
+		this->sectionName = sectionName;
+		this->keyName = keyName;
+		std::list<CSimpleIni::Entry> values;
+		if (!GetIni().GetAllValues(sectionName, keyName, values)) {
+			// No entry found in original ini => new entry => changed
+			IniChanged = true;
+		}
+		auto value = GetIni().GetValue(sectionName, keyName);
+		if (value != nullptr)
+			oldValue = value;
+	}
+	~IniChangedChecker()
+	{
+		auto value = GetIni().GetValue(sectionName, keyName);
+		std::string newValue;
+		if (value != nullptr)
+			newValue = value;
+		if (oldValue != newValue)
+			IniChanged = true;
+	}
+private:
+	std::string oldValue;
+	const char *sectionName;
+	const char *keyName;
+};
+
 int GetIniInt(const char *keyname, const char *valuename, int defaultValue)
 {
 	return GetIni().GetLongValue(keyname, valuename, defaultValue);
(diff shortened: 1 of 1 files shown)


Source/options.cpp
```diff
@@ -59,6 +59,40 @@ CSimpleIni &GetIni()
...
+		if (value != nullptr)
+			newValue = value;
+		if (oldValue != newValue)
+			IniChanged = true;
+	}
+private:
+	std::string oldValue;
+	const char *sectionName;
+	const char *keyName;
+};
```
julealgon, line 94: The scope-based approach to this seems cool, but I'm a bit surprised by the apparent complexity of this class here.

Wouldn't it be simpler to have these checks at the model level? Say, after the values are loaded into the global settings model, have a way to tell which of the settings values have been updated, and then only save those?

Feels a bit weird to me having these checks at the actual "ini saving helpers" layer and not on the caller.

obligaron: Initially I thought the same. But the problem is that the "ini saving helpers" are not only responsible for saving the (user) changed values. They also generate the default ini (if not present). This is needed so the user can find what options he could change.

I thought about checking if the file exists. But if I did it this way, no new (default) entries would be added to the ini (for example `Hardware Cursor`).

That's why I came up with this solution.

After #1561 is resolved, we could revisit this, cause the reason for option discoverability would be gone.

julealgon: What if we distribute the default ini with the game package and stop "generating an empty one" at runtime?

I have always found it weird to generate the config in runtime. Coming from .NET, this is extremely unusual to me. Usually what happens is that the application has the configuration file as part of the project contents and this is deployed alongside the application.

obligaron: The ini that is created per default is not empty 😉 
And I think `diablo.ini` is more a settings file then classic `.config` file. Cause as you stated config files shouldn't normally be changed but distributed with the program.

Back to your question. 🙂 
The problem would be that we can't insert new entries for new options in existing ini files.

For example:

- User installed DevilutionX 1.21
- Ini is created with default values (copied ini file)
- User updates to DevilutionX 1.3
- Ini does exist, we don't copy cause that would override the users preferences => Missing `Hardware Cursor` entry in users `diablo.ini` file

AJenbo: People love to strip everything except the .exe for some reason so if we can't boot without one we would probably get complaints.

julealgon: Fair enough.

julealgon: Also, yeah, didn't mean "empty" above, I know its saved with defaults initially.

Source/options.cpp
```diff
@@ -59,6 +59,40 @@ CSimpleIni &GetIni()
 	return ini;
 }
 
+static bool IniChanged = false;
+
+/**
+ * @brief Checks if a ini entry is changed by comparing value before and after
+ */
+struct IniChangedChecker {
```
glebm, line 67: This would usually be a class, as it has private fields. Idiomatically, only things with all-public members are called `struct`.

obligaron: I changed it to `class` 🙂

obligaron: @glebm ping. Are you okay with the changes? 🙂

AJenbo (approved):

PR #2325, opened 2021-07-08
Use Point/Displacement/Rectangle structs across object and lighting functions

ephphatha:
Mostly focused on objects.cpp, every commit builds so hopefully reviewing by commit is straightforward.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -723,7 +723,22 @@ void AddChestTraps()
 	}
 }
 
-void LoadMapObjects(const char *path, int startx, int starty, int x1, int y1, int w, int h, int leveridx)
+/**
+ * @brief Used by the LoadMapObjects routines to translate from "grid" coordinates to "tile" coordinates
+ *
+ * This appears to be mainly to handle differences in map scale for certain environments where some are
+ * based on 2x2 objects and some are 1x1.
+ * 
+ * @param start Initial location in grid coordinates
+ * @param offset Distance from the grid origin
+ * @return start + offset + a constant factor of {16, 16}
+*/
+constexpr Point convertToMapPos(Point start, Displacement offset)
+{
+	return start + Displacement { 16, 16 } + offset;
+}
+
+void LoadMapObjects(const char *path, Point start, Rectangle mapRange, int leveridx)
 {
 	LoadingMapObjects = true;
 	ApplyObjectLighting = true;
@@ -745,9 +760,9 @@ void LoadMapObjects(const char *path, int startx, int starty, int x1, int y1, in
 		for (int i = 0; i < width; i++) {
 			uint8_t objectId = SDL_SwapLE16(objectLayer[j * width + i]);
 			if (objectId != 0) {
-				Point mapPos { startx + 16 + i, starty + 16 + j };
+				Point mapPos = convertToMapPos(start, { i, j });
 				AddObject(ObjTypeConv[objectId], mapPos);
-				SetObjMapRange(ObjIndex(mapPos), x1, y1, x1 + w, y1 + h, leveridx);
+				SetObjMapRange(ObjIndex(mapPos), mapRange.position.x, mapRange.position.y, mapRange.position.x + mapRange.size.width, mapRange.position.y + mapRange.size.height, leveridx);
 			}
 		}
 	}
@@ -756,7 +771,7 @@ void LoadMapObjects(const char *path, int startx, int starty, int x1, int y1, in
 	LoadingMapObjects = false;
 }
 
-void LoadMapObjs(const char *path, int startx, int starty)
+void LoadMapObjs(const char *path, Point start)
 {
 	LoadingMapObjects = true;
 	ApplyObjectLighting = true;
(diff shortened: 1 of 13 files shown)


Source/objects.cpp
```diff
@@ -725,7 +725,22 @@ void AddChestTraps()
...
+ * based on 2x2 objects and some are 1x1.
+ * 
+ * @param start Initial location in grid coordinates
+ * @param offset Distance from the grid origin
+ * @return start + offset + a constant factor of {16, 16}
+*/
+constexpr Point convertToMapPos(Point start, Displacement offset)
+{
+	return start + Displacement { 16, 16 } + offset;
+}
```
AJenbo, line 741: What's happening here is not a conversion from the MegaTiles to DungeonPiece scale. It is simply adding the 16dp border (for rendering) and displacing the map data to the given `start` location.

The 2x2 mega tile only applies to the dimensions in the .dun header and the size of the first of 4 layers (tile, monster, item, object).

Pseudo format definition for .dun:
```
typedef MegaTiles uint16_t;
typedef DungeonPiece uint16_t;
dun {
    uint16_t mtWidth;
    uint16_t mtHeight;
    MegaTiles tiles[mtWidth][mtHeight];
    DungeonPiece unused[mtWidth*2][mtHeight*2];
    DungeonPiece monsters[mtWidth*2][mtHeight*2];
    DungeonPiece objects[mtWidth*2][mtHeight*2];
    DungeonPiece transparancy[mtWidth*2][mtHeight*2];
}
```

The only time the conversion is relevant is for calculating how much data to skip to locate the `objectLayer` portion of the data.

Instead I would suggest adding the 16dp border to the start location outside the loop

AJenbo: Did a bunch of edits to try and make the comment more clear :)

ephphatha: Thanks, I wrote this a while back and had a pretty poor understanding of the reason behind the translation. Your comment makes sense. I've removed the function in the latest commit since it didn't serve much purpose and the doxygen comment was wildly inaccurate.

Source/objects.cpp
```diff
@@ -1315,24 +1317,24 @@ void AddChest(int i, int t)
 	Objects[i]._oVar2 = GenerateRnd(8);
 }
 
-void AddL2Door(int i, int x, int y, int ot)
+void AddL2Door(int i, Point position, _object_id objectType)
```
AJenbo, line 1320: Nice :+1:

Source/objects.cpp
```diff
@@ -2163,42 +2165,41 @@ void ProcessObjects()
...
+	Point megaOrigin = position * 2 + Displacement { 16, 16 };
 
-	ObjSetMicro(xx + 0, yy + 0, SDL_SwapLE16(mega.micro1) + 1);
-	ObjSetMicro(xx + 1, yy + 0, SDL_SwapLE16(mega.micro2) + 1);
-	ObjSetMicro(xx + 0, yy + 1, SDL_SwapLE16(mega.micro3) + 1);
-	ObjSetMicro(xx + 1, yy + 1, SDL_SwapLE16(mega.micro4) + 1);
+	ObjSetMicro(megaOrigin, SDL_SwapLE16(mega.micro1) + 1);
+	ObjSetMicro(megaOrigin + DIR_SE, SDL_SwapLE16(mega.micro2) + 1);
+	ObjSetMicro(megaOrigin + DIR_SW, SDL_SwapLE16(mega.micro3) + 1);
+	ObjSetMicro(megaOrigin + DIR_S, SDL_SwapLE16(mega.micro4) + 1);
```
AJenbo, line 2202: A lot easier to visualize what the pattern is now :+1:

Source/setmaps.cpp
```diff
@@ -82,37 +82,37 @@ const char *const QuestLevelNames[] = {
 	N_("Archbishop Lazarus' Lair"),
 };
 
-int ObjIndex(int x, int y)
+int ObjIndex(Point position)
 {
 	for (int i = 0; i < ActiveObjectCount; i++) {
 		int oi = ActiveObjects[i];
-		if (Objects[oi].position.x == x && Objects[oi].position.y == y)
+		if (Objects[oi].position.x == position.x && Objects[oi].position.y == position.y)
```
AJenbo, line 89:
```suggestion
		if (Objects[oi].position == position)
```

AJenbo (changes requested):
Just one line that I would prefer got written as a direct comparison between `Points` and then optionally fix all the last lines in the doxygen comments :)

Source/objects.cpp
```diff
@@ -3969,7 +3978,15 @@ bool OperateShrineTainted(int pnum)
...
 
-bool OperateShrineOily(int pnum, int x, int y)
+/**
+ * @brief Oily shrines increase the players primary stat(s) by a total of two, but spawn a
+ *        firewall near the shrine that will spread towards the player
+ * @param pnum The player that activated the shrine
+ * @param spawnPosition Start location for the firewall
+ * @return false if the current player did not activate the shrine (i.e. it's a multiplayer
+ *         game) and we bailed early to avoid doubling the effects, true otherwise.
+*/
```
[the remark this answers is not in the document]
ephphatha: I will eventually work out how to get visual studio to align the end marker properly :D, added a commit fixing this and the times I missed it in previous PRs.

AJenbo (approved):

AJenbo:
Merged with some selective squashing: https://github.com/diasurgical/devilutionX/commit/b680e3fef1cf649969b01058b138f43360db82d3

PR #2326, opened 2021-07-08
Add explicit casts/replace c-style casts for type conversions

ephphatha:
Addresses the last remaining C4244 warnings (ignoring third party code).

missiles.cpp is doing a lot of conversion between screen space and world space which really shouldn't be necessary, but I couldn't be bothered trying to unpick it now. That'll be a later job unless someone else wants to take it on.

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -919,10 +919,9 @@ void MakeLightTable()
 		for (int i = 0; i < 8; i++) {
 			for (int k = 0; k < 16; k++) {
 				for (int l = 0; l < 16; l++) {
-					double fs = (BYTE)sqrt((double)(8 * l - j) * (8 * l - j) + (8 * k - i) * (8 * k - i));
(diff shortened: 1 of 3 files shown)


Source/lighting.cpp
```diff
@@ -919,10 +919,9 @@ void MakeLightTable()
 		for (int i = 0; i < 8; i++) {
 			for (int k = 0; k < 16; k++) {
 				for (int l = 0; l < 16; l++) {
-					double fs = (BYTE)sqrt((double)(8 * l - j) * (8 * l - j) + (8 * k - i) * (8 * k - i));
-					fs += fs < 0 ? -0.5 : 0.5;
```
ephphatha, line 923: This did absolutely nothing useful. BYTE is an unsigned integral type so fs is always a positive whole number. It then gets immediately converted back to an integral type which rounds toward zero, dropping the .5 value.

AJenbo (approved):

PR #2329, opened 2021-07-08
Bump warning level on MSVC to w3

AJenbo:
All but 2 issues have been fixed for w2

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -832,7 +832,7 @@ if(NOT CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
 endif()
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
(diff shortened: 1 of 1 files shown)


AJenbo:
@ephphatha considering that you are probably the only on directly targeting these I think it would be most fair to see what you think of this change :)

ephphatha (approved):
Stops msvc warning about overriding a warning level with a lower warning level, looks good :)

AJenbo:
Actually that should be the previous commit, this should add a bunch new warnings for us to deal with

ephphatha:
master gives 249 D9025 warnings, ~150 of those are for overriding /w3 with /w2 and the rest are overriding /w3 with /w. This branch only gives the 104 D9025 warnings for overriding /w3 with /w.

The increased number of other warnings is the main point of raising the warning level so I don't see that as a problem :D. Lets us know what areas can be improved.

PR #2330, opened 2021-07-08
Enable CMP0092

AJenbo:


julealgon:
What is this?

AJenbo:
Easy to google: https://cmake.org/cmake/help/git-stage/policy/CMP0092.html

julealgon:
> Easy to google: https://cmake.org/cmake/help/git-stage/policy/CMP0092.html

I obviously didn't google it on purpose. The question was rethorical, since the PR/commit title should provide the detail without needing one to google things.

AJenbo:
That wasn't obvious from your message.

PR #2333, opened 2021-07-09
Update some door related functions to take object by reference

ephphatha:
Started out as trying to use point but then focused on removing the use of globals/passing object indexes around the various SyncDoors and DoorSet functions.

Probably easiest to go through this one by commit as well since I moved a couple of functions to be defined next to the related functions.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -5215,21 +5215,20 @@ void SyncPedistal(int i)
 void SyncL2Doors(ObjectStruct &door)
 {
 	door._oMissFlag = door._oVar4 != 0;
-	int x = door.position.x;
-	int y = door.position.y;
 	door._oSelFlag = 2;
-	if (door._otype == OBJ_L2LDOOR && door._oVar4 == 0) {
-		ObjSetMicro({ x, y }, 538);
-		dSpecial[x][y] = 0;
-	} else if (door._otype == OBJ_L2LDOOR && (door._oVar4 == 1 || door._oVar4 == 2)) {
-		ObjSetMicro({ x, y }, 13);
-		dSpecial[x][y] = 5;
-	} else if (door._otype == OBJ_L2RDOOR && door._oVar4 == 0) {
-		ObjSetMicro({ x, y }, 540);
-		dSpecial[x][y] = 0;
-	} else if (door._otype == OBJ_L2RDOOR && (door._oVar4 == 1 || door._oVar4 == 2)) {
-		ObjSetMicro({ x, y }, 17);
-		dSpecial[x][y] = 6;
+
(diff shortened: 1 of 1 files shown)


AJenbo:
> Started out as trying to use point but then focused on removing the use of globals

I don't think you are the first this happens to :)

qndel:
Maybe a good moment to split into 2 functions? Handle Crypt doors / handle rest? :D

Source/objects.cpp
```diff
@@ -2274,75 +2274,75 @@ void ObjL2Special(int x1, int y1, int x2, int y2)
 	}
 }
 
-void DoorSet(int oi, int dx, int dy)
+void DoorSet(int oi, Point position)
```
AJenbo: This one really begged for the change after the previous PR :)

Source/objects.cpp
```diff
@@ -5237,14 +5237,17 @@ void SyncL3Doors(int i)
...
-	if (Objects[i]._otype == OBJ_L3LDOOR && Objects[i]._oVar4 == 0) {
-		ObjSetMicro(Objects[i].position, 531);
-	} else if (Objects[i]._otype == OBJ_L3LDOOR && (Objects[i]._oVar4 == 1 || Objects[i]._oVar4 == 2)) {
-		ObjSetMicro(Objects[i].position, 538);
-	} else if (Objects[i]._otype == OBJ_L3RDOOR && Objects[i]._oVar4 == 0) {
-		ObjSetMicro(Objects[i].position, 534);
-	} else if (Objects[i]._otype == OBJ_L3RDOOR && (Objects[i]._oVar4 == 1 || Objects[i]._oVar4 == 2)) {
-		ObjSetMicro(Objects[i].position, 541);
+
+	bool isLeftDoor = Objects[i]._otype == _object_id::OBJ_L3LDOOR; // otherwise the door is type OBJ_L3RDOOR
```
AJenbo: I wish we had some Left/Right type instead of bool for situations like this.

AJenbo (approved):

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/74584bb02a9e47e8730f6aefdb1f84a24c589c29 with selective squashing.

Personally I find that having this many commits for changes to the same lines a bit to much since a comment you think of 4 in to the review might be taken care of in commit 12.

Source/objects.cpp
julealgon: Can you elaborate what you mean? What would the API look like? You got me curious with this.

AJenbo: not sure, but a simple `enum RelativeDirection {left: right:}` would be a step up imo.

PR #2334, opened 2021-07-09
Revert "Add explicit cast for LogicalToOutput"

qndel:
This reverts commit d3c36166dfd53532af9fbe8b3754c269d4db632a.
Caused cursor to jump to bottom of the screen after picking an item up

### DIFF
--- a/Source/utils/display.h
+++ b/Source/utils/display.h
@@ -89,8 +89,8 @@ void LogicalToOutput(T *x, T *y)
 
 	float scaleX;
 	SDL_RenderGetScale(renderer, &scaleX, NULL);
-	*x = static_cast<T>(*x * scaleX);
-	*y = static_cast<T>(*x * scaleX);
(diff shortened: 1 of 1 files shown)


Source/utils/display.h
```diff
@@ -89,8 +89,8 @@ void LogicalToOutput(T *x, T *y)
 
 	float scaleX;
 	SDL_RenderGetScale(renderer, &scaleX, NULL);
-	*x = static_cast<T>(*x * scaleX);
-	*y = static_cast<T>(*x * scaleX);
```
glebm, line 93: this should've been `*y` on the right hand side

qndel: lol!

AJenbo: Fixed :)

https://github.com/diasurgical/devilutionX/commit/46c8e550ba30ac1391d6d015a7e37a4411755f5a

PR #2336, opened 2021-07-09
move click functions out of anonymous namespace

qndel:


AJenbo:
With out anything that uses it there is not good reason for this. maybe combine it with your mouse click PR...

But probably all of this should just be moved to a file that deals with input, cursor.cpp, track.cpp or something new.

qndel:
dunno, just don't want this to be half of the diff of my PR 😩

AJenbo:
> dunno, just don't want this to be half of the diff of my PR weary

Make two commits in one PR

PR #2337, opened 2021-07-09
Pause game and mute music when game window loses focus

JoBergeron:
When playing singleplayer, when the game window loses focus, I thought it would be nice to mute the music and pause the game.

This is very useful when you're playing for 10-15 at home and a pesky Teams meeting springs up and you just want to answer without having the music playing in the back.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1658,29 +1658,29 @@ bool MinimizePaused = false;
 
 void diablo_focus_pause()
 {
-	if (!gbIsMultiplayer) {
-		if (!minimizePaused) {
-			gameWasAlreadyPaused = PauseMode != 0;
-
-			if (!gameWasAlreadyPaused) {
-				PauseMode = 2;
-				sound_stop();
-				track_repeat_walk(false);
-			}
-
-			music_mute();
-			minimizePaused = true;
-		} else {
-			if (!gameWasAlreadyPaused) {
-				PauseMode = 0;
-			}
-
-			music_unmute();
-			minimizePaused = false;
-		}
-		
-		//force_redraw = 255;
+	if (gbIsMultiplayer) {
+		return;
 	}
+
+	if (!MinimizePaused) {
+		GameWasAlreadyPaused = PauseMode != 0;
+
+		if (!GameWasAlreadyPaused) {
+			PauseMode = 2;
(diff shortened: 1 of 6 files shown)


Source/diablo.cpp
```diff
@@ -1648,10 +1648,41 @@ void diablo_pause_game()
...
+			if (!gameWasAlreadyPaused) {
+				PauseMode = 0;
+			}
+
+			music_unmute();
+			minimizePaused = false;
+		}
+		
+		//force_redraw = 255;
+	}
```
AJenbo, line 1683:
```suggestion
	if (gbIsMultiplayer) {
		return;
	}

	if (!minimizePaused) {
		gameWasAlreadyPaused = PauseMode != 0;

		if (!gameWasAlreadyPaused) {
			PauseMode = 2;
			sound_stop();
			track_repeat_walk(false);
		}

		music_mute();
		minimizePaused = true;
	} else {
		if (!gameWasAlreadyPaused) {
			PauseMode = 0;
		}

		music_unmute();
		minimizePaused = false;
	}
```

Source/diablo.cpp
```diff
@@ -1648,10 +1648,41 @@ void diablo_pause_game()
 			sound_stop();
 			track_repeat_walk(false);
 		}
+
 		force_redraw = 255;
 	}
 }
 
+bool gameWasAlreadyPaused = false;
+bool minimizePaused = false;
```
AJenbo, line 1657:
```suggestion
bool GameWasAlreadyPaused = false;
bool MinimizePaused = false;
```

AJenbo:
I can see this generally being useful, especially on phones.

AJenbo (approved):

qndel:
Love this btw :D

Source/sound.cpp
```diff
@@ -306,4 +306,16 @@ int sound_get_or_set_sound_volume(int volume)
 	return sgOptions.Audio.nSoundVolume;
 }
 
+void music_mute()
+{
+	if (music)
```
julealgon, line 311: @AJenbo weren't we using explicit checks?

AJenbo: yeah, not sure what type this is, but clang-tidy should get it if it's not a bool

PR #2344, opened 2021-07-10
Object by reference in SyncObjectAnimation functions

ephphatha:
Trying to minimise the direct references to Objects so it can eventually be treated like a vector of active objects.

I suspect SyncQSTLever would be appropriately renamed SyncQuestLever but I wasn't sure so I left it as is. SyncPedistal looked like a typo given the string used is "Pedestal of Blood".

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -5142,7 +5142,7 @@ void SyncBreakObj(int pnum, int oi)
 		BreakBarrel(pnum, oi, 0, true, false);
 }
 
-void SyncCrux(int i)
+void SyncCrux(const ObjectStruct &crux)
 {
 	bool found = true;
 	for (int j = 0; j < ActiveObjectCount; j++) {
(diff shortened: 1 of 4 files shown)


AJenbo (approved):

Source/objects.cpp
```diff
@@ -5142,54 +5142,54 @@ void SyncBreakObj(int pnum, int oi)
 		BreakBarrel(pnum, oi, 0, true, false);
 }
 
-void SyncCrux(int i)
+void SyncCrux(const ObjectStruct &crux)
```
julealgon, line 5145: I know it is unrelated to the change, but I really dislike seeing this kind of thing. An operation that is specific to a certain type of object, but then passing an `ObjectStruct` which could actually represent any type of object (and not just the one where the operation would make sense).

This is where actual OO and polymorphism seems to be the way to go. Either that, or the method should be more general and work on any object.

Source/objects.cpp
```diff
@@ -5282,37 +5282,37 @@ void SyncObjectAnim(int o)
...
 		break;
 	case OBJ_BOOK2R:
 	case OBJ_BLINDBOOK:
 	case OBJ_STEELTOME:
-		SyncQSTLever(o);
+		SyncQSTLever(object);
 		break;
 	case OBJ_PEDISTAL:
-		SyncPedistal(o);
+		SyncPedestal(object, { setpc_x, setpc_y }, setpc_w);
```
julealgon, line 5315: Passing the other things here is a bit out of scope of the PR I think. Not that I don't think its a good idea, though.

Issue #2211, opened 2021-06-22
Refactor devilution.DefaultClass::SaveItemPower(int, item_effect_type, int, int, int, int, int)


AJenbo:
I've selected [**devilution.DefaultClass::SaveItemPower(int, item_effect_type, int, int, int, int, int)**](https://github.com/diasurgical/devilutionX/blob/713abdca8fc78427070d0b2b9b7fe2422c38a8b4/Source/items.cpp#L1695-L2080) for refactoring, which is a unit of **382** lines of code and **7** parameters. Addressing this will make our codebase more maintainable and improve [Better Code Hub](https://bettercodehub.com)'s **Keep Unit Interfaces Small** guideline rating! 👍 

Here's the gist of this guideline:
- **Definition** 📖 
Limit the number of parameters per unit to at most 4.
- **Why**❓ 
Keeping the number of parameters low makes units easier to understand, test and reuse.
- **How** 🔧 
Reduce the number of parameters by grouping the power parameters into an `ItemPower` struct so the function becomes `SaveItemPower(int, ItemPower)`. `PLStruct` and `UItemStruct` should also be made to be composed of this type to make it easy to pass to the function.

You can find more info about this guideline in [Building Maintainable Software](http://shop.oreilly.com/product/0636920049159.do). 📖 
 
---- 
ℹ️ To know how many _other_ refactoring candidates need addressing to get a guideline compliant, select some by clicking on the 🔲  next to them. The risk profile below the candidates signals (✅) when it's enough! 🏁 

----
Good luck and happy coding! :shipit: :sparkles: :100:

matheusgomes28:
If no one is on this yet, I'll make a start tonight. Maybe will get a PR this weekend, depending on how much actual work I have in the upcoming days :smile:

AJenbo:
Go for it :)

PR #2345, opened 2021-07-10
Refactored parameters of "StartItemPower(...)" and code around calls to it.

matheusgomes28:
The improvements include:

1. Packed the parameters to `StartItemPower(...)` into a struct.
2. Renamed some of the parameters to make logical sense, to the best of my knowledge.
3. Added builder functions to make the required parameters from other item-related structs.

A few discussions were added by @AJenbo to my [local fork's PR here](https://github.com/matheusgomes28/devilutionX/pull/4/), make sure there are no outstanding discussions in that PR before merging (I've so far replied to all of them).

This PR implements issue #2211


AJenbo:
Nice work. I went ahead and made a couple of changes, I did this instead of the normal review process as the PR is currently blocking another one, hope you don't mind :)

- `SaveItemPowerParams` is renamed `ItemPower` and used for compositing `PLStruct` and `UItemStruct`. This avoids having to convert them in various places and simplifies the others.
- Renamed randMin/Max back to param1/2 as they can represent other values depending on the power type.
- Get item by reference instead of an index in to the global `Items` (cause I actually made a bug in reusing `i` when making the other changes)

Lastly I split the change in to two commits, one that moves the function and then one that applies all the changes, so that it is easier to see what changed in the diff.

ephphatha (approved):
looks good, I wasn't a fan of the static conversion functions but that's no longer an issue with the introduction of the ItemPower struct :)

matheusgomes28 (commented):
@ephphatha I wasn't a fan of that myself, glad it's gone 😄 

@AJenbo Very welcoming changes, looking a lot better now!

AJenbo:
Did a follow up that rosters the order of itemdat.cpp and removes the redundant values from uniques:

https://github.com/diasurgical/devilutionX/pull/2347

PR #2346, opened 2021-07-10
Object by reference for BreakCrux

ephphatha:
Really not sold on the name of the check function or even if it's necessary. Feel free to chop and change.


AJenbo (approved):

AJenbo:
Lots of good stuff here :)

PR #2347, opened 2021-07-10
Unique powers cleanup

AJenbo:
Felt inspired by the cleanup of `StartItemPower`.

Unused unique item powers are now marked correctly as `IPL_INVALID` (`-1`) instead of `IPL_TOHIT` (`0`) so that `powers` can be used in range loops instead of relying on `UINumPL` being correct.

`minVal`, `maxVal` and `multVal` are moved back to `PLStruct` and calculating item value is now no longer being done by `StartItemPower` since `UItemStruct` contains a fixed value that overwrites the calculated value.

Instead SaveItemAffix() now does this for regular magic items.

Lastly the calculation for `IPL_TOHIT_DAMP` and `IPL_TOHIT_DAMP_CURSE` is now being done in a sub-function (`CalculateToHitBonus`).

All most all changes here are covered by tests except for the handling of invalid values and rendering the unique items info box:
image


Source/items.cpp
```diff
@@ -4130,29 +4142,9 @@ void DrawUniqueInfo(const Surface &out)
...
-		PrintItemPower(uitem.powers[3].type, &curruitem);
-		DrawString(out, tempstr, rect, UIS_SILVER | UIS_CENTER);
-	}
-	if (uitem.UINumPL > 4) {
-		rect.position.y += 2 * 12;
-		PrintItemPower(uitem.powers[4].type, &curruitem);
-		DrawString(out, tempstr, rect, UIS_SILVER | UIS_CENTER);
-	}
-	if (uitem.UINumPL > 5) {
+	for (int i = 1; i < uitem.UINumPL; i++) {
```
ephphatha, line 4145: This loop can start from 0, line 4142 can be split to make the first bonus behave the same as subsequent bonuses.

ephphatha: Could potentially add an assert to document that UINumPL is at most 6.

Source/items.cpp
```diff
@@ -1093,45 +1090,58 @@ void SaveItemPower(ItemStruct &item, const ItemPower &power)
...
-		item._iPLMana -= (r2 << 6);
-		item._iPLHP += (r2 << 6);
-		break;
-	case IPL_LIFETOMANA:
-		r2 = ((Players[MyPlayerId]._pMaxHPBase >> 6) * 40 / 100);
-		item._iPLHP -= (r2 << 6);
-		item._iPLMana += (r2 << 6);
-		break;
+	case IPL_MANATOLIFE: {
+		int fraction = ((Players[MyPlayerId]._pMaxManaBase >> 6) * 50 / 100) << 6;
```
ephphatha, line 1094: `fraction` doesn't really describe the purpose of this value, but I can't think of a more appropriate name 😅

ephphatha (commented):
Rolling the first case into the loop seems like a quick win. The mana > life and life > mana variable name stood out to me while reading through the changes so I wanted to mention it but I don't think that's a blocker.

Source/items.cpp
julealgon: Maybe `portion`?

Source/items.cpp
AJenbo: Would having bool first = true; be an ok way to handle the extra offsets needed for the first power?

ephphatha: I was thinking something like:

```cpp
rect.position.y += (10 - uitem.UINumPL) * 12; // using 10 - power level instead of 12
for (int i = 0; i < uitem.UINumPL; i++) {
    rect.position.y += 2 * 12; // this allows treating the first case as normal and gets back to the expected offset
    PrintItemPower();
    DrawString();
}
```

Could also do the same thing by leaving the first increment as 12 - power and moving the inner increment to the end of the loop.

Source/items.cpp
AJenbo: Anything's better then `r2` :P The number isn't even Random :D

ephphatha (approved):

PR #2351, opened 2021-07-11
Move SetObjMapRange to a member function

ephphatha:
Due to the way this function is used it was easiest to also change ObjIndex to return a reference to the object instead of the index. Also added an explicit conversion from Size to Displacement to make one of the uses slightly cleaner to express.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -509,7 +509,7 @@ void AddBookLever(int x1, int y1, int x2, int y2, _speech_id msg)
 		AddObject(OBJ_BLOODBOOK, { xp, yp });
 	}
 	int ob = dObject[xp][yp] - 1;
-	SetObjMapRange(ob, x1, y1, x2, y2, leverid);
+	Objects[ob].SetMapRange({ x1, y1 }, { x2, y2 }, leverid);
 	SetBookMsg(ob, msg);
 	leverid++;
 	Objects[ob]._oVar6 = Objects[ob]._oAnimFrame + 1;
@@ -748,7 +748,7 @@ void LoadMapObjects(const char *path, Point start, Rectangle mapRange, int lever
 			if (objectId != 0) {
 				Point mapPos = start + Displacement { i, j };
 				AddObject(ObjTypeConv[objectId], mapPos);
-				SetObjMapRange(ObjIndex(mapPos), mapRange.position.x, mapRange.position.y, mapRange.position.x + mapRange.size.width, mapRange.position.y + mapRange.size.height, leveridx);
+				ObjectAtPosition(mapPos).SetMapRange(mapRange, leveridx);
 			}
 		}
 	}
@@ -1245,15 +1245,6 @@ void SetupObject(int i, Point position, _object_id ot)
 	Objects[i]._oPreFlag = false;
 	Objects[i]._oTrapFlag = false;
 	Objects[i]._oDoorFlag = false;
-}
-
-void SetObjMapRange(int i, int x1, int y1, int x2, int y2, int v)
-{
-	Objects[i]._oVar1 = x1;
-	Objects[i]._oVar2 = y1;
-	Objects[i]._oVar3 = x2;
-	Objects[i]._oVar4 = y2;
-	Objects[i]._oVar8 = v;
 }
 
 void SetBookMsg(int i, _speech_id msg)
--- a/Source/objects.h
+++ b/Source/objects.h
@@ -8,6 +8,7 @@
 #include <cstdint>
 
 #include "engine/point.hpp"
+#include "engine/rectangle.hpp"
 #include "itemdat.h"
 #include "objdat.h"
 #include "textdat.h"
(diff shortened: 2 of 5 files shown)


Source/setmaps.cpp
```diff
@@ -82,37 +82,37 @@ const char *const QuestLevelNames[] = {
...
-	SetObjMapRange(ObjIndex({ 27, 37 }), 8, 1, 15, 11, 3);
-	SetObjMapRange(ObjIndex({ 46, 35 }), 8, 1, 15, 11, 3);
-	SetObjMapRange(ObjIndex({ 49, 53 }), 8, 1, 15, 11, 3);
-	SetObjMapRange(ObjIndex({ 27, 53 }), 8, 1, 15, 11, 3);
+	ObjectAtPosition({ 64, 34 }).SetMapRange({ 20, 7 }, { 23, 10 }, 1);
+	ObjectAtPosition({ 64, 59 }).SetMapRange({ 20, 14 }, { 21, 16 }, 2);
+	ObjectAtPosition({ 27, 37 }).SetMapRange({ 8, 1 }, { 15, 11 }, 3);
+	ObjectAtPosition({ 46, 35 }).SetMapRange({ 8, 1 }, { 15, 11 }, 3);
+	ObjectAtPosition({ 49, 53 }).SetMapRange({ 8, 1 }, { 15, 11 }, 3);
+	ObjectAtPosition({ 27, 53 }).SetMapRange({ 8, 1 }, { 15, 11 }, 3);
```
AJenbo, line 102: Much more readable, the old stuff was just a table of numbers

Source/objects.cpp
```diff
@@ -748,7 +748,7 @@ void LoadMapObjects(const char *path, Point start, Rectangle mapRange, int lever
 			if (objectId != 0) {
 				Point mapPos = start + Displacement { i, j };
 				AddObject(ObjTypeConv[objectId], mapPos);
-				SetObjMapRange(ObjIndex(mapPos), mapRange.position.x, mapRange.position.y, mapRange.position.x + mapRange.size.width, mapRange.position.y + mapRange.size.height, leveridx);
+				ObjectAtPosition(mapPos).SetMapRange(mapRange, leveridx);
```
AJenbo, line 751: wow

Source/objects.h
```diff
@@ -49,6 +50,36 @@ struct ObjectStruct {
...
+		_oVar4 = bottomRight.y;
+		_oVar8 = v;
+	}
+
+	/**
+	 * @brief Convenience function for SetMapRange(Point, Point, int)
+	 * @param mapRange A rectangle defining the top left corner and size of the affected region
+	 * @param v Object subtype/discriminator
+	 */
+	constexpr void SetMapRange(Rectangle mapRange, int v)
```
AJenbo, line 79: I hope we can eventually make this the actual function :)

AJenbo (approved):

Source/objects.cpp
```diff
@@ -509,7 +509,7 @@ void AddBookLever(int x1, int y1, int x2, int y2, _speech_id msg)
 		AddObject(OBJ_BLOODBOOK, { xp, yp });
 	}
 	int ob = dObject[xp][yp] - 1;
-	SetObjMapRange(ob, x1, y1, x2, y2, leverid);
+	Objects[ob].SetMapRange({ x1, y1 }, { x2, y2 }, leverid);
```
julealgon, line 512: Have you considered making `SetMapRange` only take a `Rectangle`, and then creating a `Rectangle` ctor that takes the topleft and bottomright points to build it?

I think that would centralize more of the logic in the rectangle class and this could be used in other places as well, while keeping consumer APIs simpler (no need for overloads in this case).

Source/objects.h
```diff
@@ -49,6 +50,36 @@ struct ObjectStruct {
...
+	 * @param topLeft corner of the map region closest to the origin
+	 * @param bottomRight corner of the map region furthest from the origin
+	 * @param v ID/discriminator for the object type? Needs to be investigated further
+	 */
+	constexpr void SetMapRange(Point topLeft, Point bottomRight, int v)
+	{
+		_oVar1 = topLeft.x;
+		_oVar2 = topLeft.y;
+		_oVar3 = bottomRight.x;
+		_oVar4 = bottomRight.y;
```
julealgon, line 70: Are `oVar`s used for other things? Asking just in case it would be possible to unify them into a single `Rectangle` variable.

Source/objects.h
```diff
@@ -49,6 +50,36 @@ struct ObjectStruct {
...
+	 *
+	 * Some objects will cause a map region to change when a player interacts with them (e.g. Skeleton King
+	 * antechamber levers). The coordinates used for this region are based on a 40*40 grid overlaying the central
+	 * 80*80 region of the dungeon.
+	 *
+	 * @param topLeft corner of the map region closest to the origin
+	 * @param bottomRight corner of the map region furthest from the origin
+	 * @param v ID/discriminator for the object type? Needs to be investigated further
+	 */
+	constexpr void SetMapRange(Point topLeft, Point bottomRight, int v)
```
julealgon, line 65: I'd suggest suffixing the arguments with "position" for consistency and semantic value. So `topLeftPosition` and `bottomRightPosition`.

Source/objects.h
ephphatha: Unfortunately yes. They're general purpose variables for all sorts of object types and the ones at the start tend to be the most commonly used.

Source/objects.cpp
ephphatha: I was a bit hesitant about adding that constructor in this PR as it would make declarations/initialisations like `Rectangle{ { 1, 2 }, { 4, 4 } }` ambiguous, forcing the use of Size or Point explicitly in those cases. It also makes the SetMapRange logic more verbose unless we also updated the save/load code to detect when these variables are used as map coordinates as we need to translate from Point/Size to Point/Point in at least one of those places. It would make it nicer to use in consumers (such as ObjChangeMap/Resync, Obj*Special, Add*Objs) though, so I reckon it would be worthwhile eventually.

PR #2352, opened 2021-07-11
Extract common Book object initialisation logic

ephphatha:
This really is a work in progress, trying to document small chunks to the best of my understanding so this can be split into specialised object classes or components in the future.

### DIFF
--- a/Source/objects.h
+++ b/Source/objects.h
@@ -63,28 +63,64 @@ struct ObjectStruct {
 	 * antechamber levers). The coordinates used for this region are based on a 40*40 grid overlaying the central
 	 * 80*80 region of the dungeon.
 	 *
-	 * @param topLeftPosition corner of the map region closest to the origin
-	 * @param bottomRightPosition corner of the map region furthest from the origin
-	 * @param v ID/discriminator for the object type? Needs to be investigated further
+	 * @param topLeftPosition corner of the map region closest to the origin.
+	 * @param bottomRightPosition corner of the map region furthest from the origin.
 	 */
-	constexpr void SetMapRange(Point topLeftPosition, Point bottomRightPosition, int v)
+	constexpr void SetMapRange(Point topLeftPosition, Point bottomRightPosition)
 	{
 		_oVar1 = topLeftPosition.x;
 		_oVar2 = topLeftPosition.y;
 		_oVar3 = bottomRightPosition.x;
 		_oVar4 = bottomRightPosition.y;
-		_oVar8 = v;
 	}
 
 	/**
-	 * @brief Convenience function for SetMapRange(Point, Point, int)
-	 * @param mapRange A rectangle defining the top left corner and size of the affected region
-	 * @param v Object subtype/discriminator
+	 * @brief Convenience function for SetMapRange(Point, Point).
+	 * @param mapRange A rectangle defining the top left corner and size of the affected region.
 	 */
-	constexpr void SetMapRange(Rectangle mapRange, int v)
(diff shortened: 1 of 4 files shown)


Source/objects.h
```diff
@@ -58,28 +63,64 @@ struct ObjectStruct {
...
 	/**
-	 * @brief Convenience function for SetMapRange(Point, Point, int)
-	 * @param mapRange A rectangle defining the top left corner and size of the affected region
-	 * @param v Object subtype/discriminator
+	 * @brief Marks an object which was spawned from a sublevel in response to a lever activation.
+	 * @param mapRange The region which was updated to spawn this object.
+	 * @param leverID The id (*not* an object ID/index) of the lever responsible for the map change.
 	 */
-	constexpr void SetMapRange(Rectangle mapRange, int v)
+	constexpr void InitializeLoadedObject(Rectangle mapRange, int leverID)
```
ephphatha, line 118: This logic is also used for the quest objects in setmaps.cpp, I went with this name and description as the primary use seemed to be in objects.cpp:LoadMapObjects()

Source/objects.h
```diff
@@ -58,28 +63,64 @@ struct ObjectStruct {
...
+	 *
+	 * Books of this type use a generic message (see OperateSChambBook()) compared to the more specific quest books
+	 * initialized by IntializeQuestBook().
+	 *
+	 * @param mapRange The region to be updated when this object is activated.
+	*/
+	constexpr void InitializeBook(Rectangle mapRange)
+	{
+		SetMapRange(mapRange);
+		_oVar6 = _oAnimFrame + 1; // Save the frame number for the open book frame
```
ephphatha, line 97: I'm pretty sure using `_oVar6` this way is unnecessary, it's used in OperateBookLever and OperateSChambBk to determine if the player has interacted with the object, but this check can also be accomplished by looking at the quest state (done earlier in OperateBookLever) or using this as a bool value. Changing that would require adding logic when saving/loading for compatibility.

Source/setmaps.cpp
```diff
@@ -94,25 +94,30 @@ ObjectStruct &ObjectAtPosition(Point position)
...
-	ObjectAtPosition({ 46, 35 }).SetMapRange({ 8, 1 }, { 15, 11 }, 3);
-	ObjectAtPosition({ 49, 53 }).SetMapRange({ 8, 1 }, { 15, 11 }, 3);
-	ObjectAtPosition({ 27, 53 }).SetMapRange({ 8, 1 }, { 15, 11 }, 3);
+	constexpr Rectangle region1 { { 20, 7 }, { 3, 3 } };
+	ObjectAtPosition({ 64, 34 }).InitializeLoadedObject(region1, 1);
+
+	constexpr Rectangle region2 { { 20, 14 }, { 1, 2 } };
+	ObjectAtPosition({ 64, 59 }).InitializeLoadedObject(region2, 2);
+
+	constexpr Rectangle region3 { { 8, 1 }, { 7, 10 } };
```
AJenbo, line 103: would be nice if we could name them `smallSecretRoom`, `gate` and `largeSecretRoom`

AJenbo (approved):

Source/objects.h
AJenbo: I do prefer isolating the bad architecture in the save/load routine

Source/setmaps.cpp
ephphatha: Thanks heaps, I hadn't got around to working out what these actually were. I'll update the PR now.

AJenbo (approved):

Issue #2340, opened 2021-07-09
Broken level generation (catacombs and hell)


AJenbo:
While testing a clean up I noticed that we appear to have broken some level generation in a subtle way

Using this save from the original Hellfire:
https://github.com/diasurgical/devilutionX/files/5373942/hsv.zip

Hell level 15:
It would have been impossible for the hero to get to this point with out seeing the area south west of where he is standing in this screen shot.
![image](https://user-images.githubusercontent.com/204594/125144803-50288280-e11f-11eb-9b1d-6f44416f017f.png)

Catacombs level 7:
A free floating closed and open door plus two candle sticks just hanging out in the middle of a room.
![image](https://user-images.githubusercontent.com/204594/125144905-bc0aeb00-e11f-11eb-8a15-1e1e573aedaf.png)

If someone could help bisect it it would be appreciated :)

Chance4us:
In the first picture I see that the automap is defective. But when I load the game and go up to level 15, it looks normal.
Do I have to pay attention to something or what exactly differs from the normality?

AJenbo:
Odd, what version are you running? This was with the https://github.com/diasurgical/devilutionX/commit/74584bb02a9e47e8730f6aefdb1f84a24c589c29 and a debug build.

The way i went to level 15 was taking the TP and then the stairs up before diablo get to many hits in.

Chance4us:
In fact you are right. The automap looks broken and a chest is unusually close to the wall.
To be honest, the bug looks like #1506 or something related to it. I'll investigate soon.

Chance4us:
In the first picture the automap is broken by the following commit.
373f1db22edec89860eafdf4a0bb0f7199203b78 If cleanup + bugfixes (#2273)

ephphatha:
Great work tracking that down :D.

https://github.com/diasurgical/devilutionX/blob/e25d287ba97346946d94a6d6f51195fc8dea2e9b/Source/drlg_l3.cpp#L1868
The `145` case (in both checks) should be `134`.

https://github.com/diasurgical/devilutionX/blob/e25d287ba97346946d94a6d6f51195fc8dea2e9b/Source/drlg_l4.cpp#L543
This is missing case `8`

Those were the only two changes to level generation logic I could see, can't guarantee I got everything though 😅. I think this would have a flow on effect to the monster spawns, worth testing after this is fixed.

Chance4us:
The single door is an another issue.

Chance4us:
[...]

PR #2353, opened 2021-07-11
Restore dropped cases for fence fix and vertical wall checks

ephphatha:
I still haven't bought hellfire so can't test this myself using the original save from #2340, hoping someone can check this for me.


AJenbo (approved):

qndel:
Nice catch :D

Chance4us:
@AJenbo, @ephphatha, @qndel.
I am sorry to say that this problem has still not been resolved. I actually thought so and just spontaneously tested again and the problem is still there.

PR #2354, opened 2021-07-11
Fallen fear: Use _mgoalvar2 for new direction instead of _mdir

### LINKED ISSUE #2338: Dark one moves strangely when under fire
Dark one moves strangely when under fire.
I haven't investigated since when and why it happened.
I cannot add any information about whether other monsters are also affected.
In order to reproduce this movement it is important to leave a little distance and to fire from a distance.
You can see what it looks like on the following video.

https://user-images.githubusercontent.com/45902449/125136224-b35bea00-e10a-11eb-981f-51d2c1dfe168.mp4

Savegame:
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6793932/single_0.hsv.zip)

I will provide more information a bit later.

obligaron:
Fixes #2338

## Notes

- This slightly breaks save game compatibility, cause vanilla doesn't use `_mgoalvar2`. But it's only when a fallen is retreating.
- The bug is also present in vanilla. Just with a different manifestation. When you save a game with a fallen that retreated and load it again it shows the wrong animation. This is cause `SyncMonsterAnim` uses `_mdir` to load the animation. And so `_mmode` and the animation doesn't match. 😉 

## Alternative Solutions

- Introduce new Variable `RealMonsterDirection` to `MonsterStruct` 😛
- Calculate the direction from used `CelSprite` (comparing pointers) :open_mouth:

But both wouldn't fix the vanilla bug and the way `SyncMonsterAnim` is used it seams that the game expects that `_mdir` and animation always matches (this should also be the case for players etc.).

@Chance4us Thanks for reporting the issue. Could you test this PR? That would help a lot. 🙂

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -3232,7 +3232,7 @@ void MAI_Fallen(int i)
 	if (monst->_mgoal == MGOAL_RETREAT) {
 		if (monst->_mgoalvar1-- == 0) {
 			monst->_mgoal = MGOAL_NORMAL;
-			M_StartStand(i, opposite[monst->_mdir]);
+			M_StartStand(i, opposite[monst->_mgoalvar2]);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


julealgon (approved):
Seems sensible to me.

AJenbo:
Considering what _mgoalvar2 is normally used for this seven seams sensible.

PR #2356, opened 2021-07-11
:bug: WillThemeRoomFit is failing to place rooms near edge of map

AJenbo:
Broken in 1.0.0 a1b7b8522a528cc5fd309b4295bbb3aa255bea30

Fixes https://github.com/diasurgical/devilutionX/issues/2340

Additionally this now does proper bounds check before accessing the `dungeon` array.

### DIFF
--- a/Source/gendung.cpp
+++ b/Source/gendung.cpp
@@ -96,9 +96,11 @@ bool WillThemeRoomFit(int floor, int x, int y, int minSize, int maxSize, int *wi
 	int xCount = 0;
 	int yCount = 0;
 
-	// BUGFIX: change '&&' to '||' (fixed)
-	if (x > DMAXX - maxSize || y > DMAXY - maxSize) {
-		return false;
+	if (x + maxSize > DMAXX && y + maxSize > DMAXY) {
+		return false; // Original broken bounds check, avoids lower right corner
+	}
+	if (x + minSize > DMAXX || y + minSize > DMAXY) {
+		return false; // Skip definit OOB cases
 	}
 	if (!SkipThemeRoom(x, y)) {
 		return false;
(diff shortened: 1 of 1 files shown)


Source/gendung.cpp
```diff
@@ -96,8 +96,10 @@ bool WillThemeRoomFit(int floor, int x, int y, int minSize, int maxSize, int *wi
 	int xCount = 0;
 	int yCount = 0;
 
-	// BUGFIX: change '&&' to '||' (fixed)
-	if (x > DMAXX - maxSize || y > DMAXY - maxSize) {
+	if (x > DMAXX - maxSize && y > DMAXY - maxSize) {
+		return false;
+	}
+	if (x > DMAXX - minSize || y > DMAXY - minSize) {
```
julealgon, line 102: Just to make sure: is this really an "and for max, but or for min"?

AJenbo: Yes, rooms will apparently always be `maxSize` on either W or H, so if that isn't possible with out OOB the first one will do an early return. The room also has to be at least minSize in the other direction so if that isn't possible on either then we can return false as well.

It might be worth it to be more clever about it write it as:
```cpp
if (!(x + maxSize > DMAXX && y + minSize > DMAXY) && !(x + minSize > DMAXX && y + maxSize > DMAXY)) {
    return false;
}
```

julealgon: Thanks for confirming.

Source/gendung.cpp
```diff
@@ -96,8 +96,7 @@ bool WillThemeRoomFit(int floor, int x, int y, int minSize, int maxSize, int *wi
 	int xCount = 0;
 	int yCount = 0;
 
-	// BUGFIX: change '&&' to '||' (fixed)
-	if (x > DMAXX - maxSize || y > DMAXY - maxSize) {
+	if ((x + maxSize >= DMAXX || y + minSize >= DMAXY) && (x + minSize >= DMAXX && y + maxSize >= DMAXY)) {
```
qndel, line 99: why the mixed check?

AJenbo: To get as close as possible to handeling all OOB issues

ephphatha: This one confused me too. Having a comment in the code with your explanation from https://github.com/diasurgical/devilutionX/pull/2356#discussion_r667533450 would be great

qndel: yes but wasn't it supposed to be minSize x2 in the first check then maxSize x2 in the second one? I don't understand the current one

AJenbo: Hmm I thought it required one of the sides to be maxSize, but looking at this again it might be ok with minSize x minSize rooms. The original would actually reject a bunch of valid rooms if it's in the corner so we might need to keep that.

qndel (approved):
LGTM

PR #2357, opened 2021-07-11
ADL: Handle skipped Frames for Diablo Got-Hit Animation

obligaron:
Diablo is the only monster that contains skipped frames, only for the got-hit animation and only in hellfire.
Now we get the missing frames back.

<details><summary>Example Video (10% game speed)</summary>

### Master

https://user-images.githubusercontent.com/25415264/125210759-192eaa00-e2a2-11eb-984e-184e0a36d348.mp4

### PR

https://user-images.githubusercontent.com/25415264/125210761-1df35e00-e2a2-11eb-937e-4075d0490ab8.mp4

</details>

Notes:

- No behavior changes, only animations
- Golem has no Got-Hit animation, so `NewMonsterAnim` shouldn't be called with Golem and Got-Hit. That's why I think that the path that previous didn't checked for `MT_GOLEM` should be fine.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -50,6 +50,21 @@ void NewMonsterAnim(int i, AnimStruct *anim, Direction md, AnimationDistribution
 	monst->_mdir = md;
 }
 
+void StartMonsterGotHit(int monsterId)
+{
+	auto &monster = Monsters[monsterId];
+	if (monster.MType->mtype != MT_GOLEM) {
+		auto animationFlags = gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None;
+		NewMonsterAnim(monsterId, &monster.MType->Anims[MA_GOTHIT], monster._mdir, animationFlags);
+		monster._mmode = MM_GOTHIT;
+	}
+	monster.position.offset = { 0, 0 };
+	monster.position.tile = monster.position.old;
+	monster.position.future = monster.position.old;
+	M_ClearSquares(monsterId);
+	dMonster[monster.position.tile.x][monster.position.tile.y] = monsterId + 1;
+}
+
 } // namespace
 
 #define NIGHTMARE_TO_HIT_BONUS 85
@@ -1574,13 +1589,7 @@ void M_GetKnockback(int i)
 
 	M_ClearSquares(i);
 	Monsters[i].position.old += d;
-	NewMonsterAnim(i, &Monsters[i].MType->Anims[MA_GOTHIT], Monsters[i]._mdir, gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None);
-	Monsters[i]._mmode = MM_GOTHIT;
-	Monsters[i].position.offset = { 0, 0 };
-	Monsters[i].position.tile = Monsters[i].position.old;
-	Monsters[i].position.future = Monsters[i].position.tile;
-	M_ClearSquares(i);
-	dMonster[Monsters[i].position.tile.x][Monsters[i].position.tile.y] = i + 1;
+	StartMonsterGotHit(i);
 }
 
 void M_StartHit(int i, int pnum, int dam)
@@ -1608,13 +1617,7 @@ void M_StartHit(int i, int pnum, int dam)
 			Monsters[i]._mgoalvar2 = 0;
 		}
 		if (Monsters[i]._mmode != MM_STONE) {
-			NewMonsterAnim(i, &Monsters[i].MType->Anims[MA_GOTHIT], Monsters[i]._mdir, gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None);
-			Monsters[i]._mmode = MM_GOTHIT;
-			Monsters[i].position.offset = { 0, 0 };
-			Monsters[i].position.tile = Monsters[i].position.old;
-			Monsters[i].position.future = Monsters[i].position.old;
-			M_ClearSquares(i);
-			dMonster[Monsters[i].position.tile.x][Monsters[i].position.tile.y] = i + 1;
+			StartMonsterGotHit(i);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


Source/monster.cpp
```diff
@@ -50,6 +50,19 @@ void NewMonsterAnim(int i, AnimStruct *anim, Direction md, AnimationDistribution
 	monst->_mdir = md;
 }
 
+void StartMonsterGotHit(int mid)
```
julealgon, line 53: Can we take the opportunity and actually pass a `MonstStruct` reference here instead of the id?

julealgon, line 53: If the above is not possible, I'd suggest renaming to `monsterId` or `monsterIndex`.

julealgon, line 53: Also if passing the reference is not possible, consider creating a local `monster` variable that assigns the reference to avoid duplication.

AJenbo:
Even at 10% it's pretty fluid :D

Source/monster.cpp
```diff
@@ -50,6 +50,19 @@ void NewMonsterAnim(int i, AnimStruct *anim, Direction md, AnimationDistribution
...
 
+void StartMonsterGotHit(int mid)
+{
+	if (Monsters[mid].MType->mtype != MT_GOLEM) {
+		NewMonsterAnim(mid, &Monsters[mid].MType->Anims[MA_GOTHIT], Monsters[mid]._mdir, gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None);
+		Monsters[mid]._mmode = MM_GOTHIT;
+	}
+	Monsters[mid].position.offset = { 0, 0 };
+	Monsters[mid].position.tile = Monsters[mid].position.old;
+	Monsters[mid].position.future = Monsters[mid].position.old;
```
julealgon, line 62: I wonder if you considered abstracting this "sequence" of actions here as a member function on `position` itself. Perhaps something like `stop()` or similar.

EDIT: Or as a member function on the `MonstStruct` itself even (might make even more sense).

Source/monster.cpp
```diff
@@ -50,6 +50,19 @@ void NewMonsterAnim(int i, AnimStruct *anim, Direction md, AnimationDistribution
 	monst->_mdir = md;
 }
 
+void StartMonsterGotHit(int mid)
+{
+	if (Monsters[mid].MType->mtype != MT_GOLEM) {
+		NewMonsterAnim(mid, &Monsters[mid].MType->Anims[MA_GOTHIT], Monsters[mid]._mdir, gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None);
+		Monsters[mid]._mmode = MM_GOTHIT;
+	}
+	Monsters[mid].position.offset = { 0, 0 };
```
julealgon, line 60: Should we consider creating a static field `Displacement::None` or `Displacement::Zero` for readability?

Source/monster.cpp
```diff
@@ -53,7 +53,8 @@ void NewMonsterAnim(int i, AnimStruct *anim, Direction md, AnimationDistribution
 void StartMonsterGotHit(int mid)
 {
 	if (Monsters[mid].MType->mtype != MT_GOLEM) {
-		NewMonsterAnim(mid, &Monsters[mid].MType->Anims[MA_GOTHIT], Monsters[mid]._mdir, gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None);
+		int numSkippedFrames = (gbIsHellfire && Monsters[mid].MType->mtype == MT_DIABLO) ? 4 : 0;
+		NewMonsterAnim(mid, &Monsters[mid].MType->Anims[MA_GOTHIT], Monsters[mid]._mdir, gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None, numSkippedFrames);
```
julealgon, line 57: The ternary as the second to last argument here makes the line quite long and a bit hard to read. Consider extracting a variable for that value instead.

Source/monster.cpp
ephphatha: Currently Direction::DIR_OMNI converts to a Displacement of 0, 0 however I don't want to suggest using that enum value here. In this context I think a separate static for 0 magnitude displacement as you suggest makes more sense.

Source/monster.cpp
obligaron: Currently that's not possible cause `M_ClearSquares` and `dMonster` need the index.

obligaron: Renamed to `monsterId` as it was used more often in the existing code base 🙂

obligaron: done 🙂

Source/monster.cpp
obligaron: I found the same sequence at two other places (death and diablo death).
But I think this can be done later and not part of this pr.

Source/monster.cpp
obligaron: I also think this is a good idea. 🙂 
But as this is used in another places too, this should be in a separate refactoring pr.

Source/monster.cpp
obligaron: done 🙂

Source/monster.cpp
```diff
@@ -39,6 +39,19 @@
...
+void NewMonsterAnim(int i, AnimStruct *anim, Direction md, AnimationDistributionFlags flags = AnimationDistributionFlags::None, int numSkippedFrames = 0, int distributeFramesBeforeFrame = 0)
+{
+	MonsterStruct *monst = &Monsters[i];
+	auto *pCelSprite = &*anim->CelSpritesForDirections[md];
+	monst->AnimInfo.SetNewAnimation(pCelSprite, anim->Frames, anim->Rate, flags, numSkippedFrames, distributeFramesBeforeFrame);
+	monst->_mFlags &= ~(MFLAG_LOCK_ANIMATION | MFLAG_ALLOW_SPECIAL);
+	monst->_mdir = md;
+}
+
+} // namespace
```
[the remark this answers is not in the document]
obligaron: The defines aren't used, so it's not necessary.
I also wasn't sure where to insert the anonymous namespace. 😕 
It could also be done after all the globals. Perhaps before `InitMonsterTRN`?
But we could also defer this to later when we need the stuff.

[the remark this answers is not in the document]
julealgon: Yeah I know it is not "necessary" per se, but was just wondering about best practices and consistency. I guess this is more a question to @AJenbo and @glebm here. It's just that I've always seen #defines come at the very top of the file.

julealgon (approved):

Source/monster.cpp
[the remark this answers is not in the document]
obligaron: I looked at player.cpp there the anonymous namespace is directly after devilution namespace.
So I did the same. 😄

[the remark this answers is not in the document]
AJenbo: I would prefer that they be converted in to `constexpr`, in both cases I would have them at the top of the namespace.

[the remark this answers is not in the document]
AJenbo: But I haven't cleaned that file yet :o

PR #2358, opened 2021-07-11
Clean up ProcessMissiles

AJenbo:


Source/missiles.cpp
```diff
@@ -198,7 +198,7 @@ static bool CheckBlock(Point from, Point to)
 	return false;
 }
 
-static inline bool InDungeonBounds(Point position)
+static inline bool InDungeonBounds(const Point position)
```
ephphatha, line 201: Is there any improvement gained by declaring const when passing by value? I like using const reference for non-primitives, though I recall conversation on previous PRs where that was discouraged for relatively simple structs like Point.

julealgon: Yeah this seems unnecessary and also nonstandard as per the rest of the codebase. Would suggest removing.

Source/missiles.cpp
```diff
@@ -198,7 +198,7 @@ static bool CheckBlock(Point from, Point to)
 	return false;
 }
 
-static inline bool InDungeonBounds(Point position)
+static inline bool InDungeonBounds(const Point position)
 {
 	return position.x > 0 && position.x < MAXDUNX && position.y > 0 && position.y < MAXDUNY;
```
ephphatha: While this isn't an issue with the changes made in this PR this check doesn't quite match the function name. The last row/col is considered in bounds but the first row/col isn't. I would like to fix this eventually even if in practice it doesn't change anything 😅 . Bringing this up because ProcessMissiles() relies on the current behaviour to compensate for missing checks in the missile logic functions.

Source/missiles.cpp
```diff
@@ -4773,48 +4773,51 @@ void MI_Rportal(int i)
...
+
+void ProcessMissiles()
+{
+	for (int i = 0; i < ActiveMissileCount; i++) {
+		auto &missile = Missiles[ActiveMissiles[i]];
+		const auto &position = missile.position.tile;
+		dFlags[position.x][position.y] &= ~BFLAG_MISSILE;
+		dMissile[position.x][position.y] = 0;
+		if (!InDungeonBounds(position))
+			missile._miDelFlag = true;
```
ephphatha, line 4795: Again, not something you've introduced in this PR but it stands out now that you've cleaned up the code. This is a very strange check, position is assumed to be valid and used to index into the dFlags and dMissile arrays and then checked after it would cause out of bounds writes? From reading the code the only time the !InDungeonBounds check accomplishes anything is if the missile moves into row/col 0.

We should be able to remove the InDungeonBounds check here once the missile procedures handle bounds checking properly. I can log an issue to track that once I've had a bit of time to think about the wording 😄

Source/missiles.cpp
AJenbo: True, missed that it wasn't by reference.

ephphatha (approved):
Looks good, a couple of issues with the existing logic were highlighted by the cleanup but I'll try write up an issue to address them properly.

Source/missiles.cpp
AJenbo: Yeah I only wanted to make this code more readable since it smelled a bit :D

PR #2360, opened 2021-07-12
Get rid of unchecked calls to malloc

vlolteanu:
When malloc fails, it does so silently and returns NULL. Not checking the returned value causes the program to segfault sometime later.

Such calls were replaced with new, which fails right away with bad_alloc.

### DIFF
--- a/Source/dthread.cpp
+++ b/Source/dthread.cpp
@@ -81,6 +81,8 @@ void dthread_send_delta(int pnum, _cmd_id cmd, byte *pbSrc, int dwLen)
 	}
 
 	pkt = static_cast<TMegaPkt *>(std::malloc(dwLen + 20));
+	if (pkt == nullptr)
+		app_fatal("Failed to allocate memory");
 	pkt->pNext = nullptr;
 	pkt->dwSpaceLeft = pnum;
 	pkt->data[0] = static_cast<byte>(cmd);
--- a/Source/tmsg.cpp
+++ b/Source/tmsg.cpp
@@ -38,6 +38,8 @@ void tmsg_add(byte *pbMsg, uint8_t bLen)
 	TMsg **tail;
 
 	TMsg *msg = static_cast<TMsg *>(std::malloc(bLen + sizeof(*msg)));
+	if (msg == nullptr)
+		app_fatal("Failed to allocate memory");
 	msg->hdr.pNext = nullptr;
 	msg->hdr.dwTime = SDL_GetTicks() + gnTickDelay * 10;
 	msg->hdr.bLen = bLen;
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -60,6 +60,8 @@ void GetNextPacket()
 	TMegaPkt *result;
 
 	sgpCurrPkt = static_cast<TMegaPkt *>(std::malloc(sizeof(TMegaPkt)));
+	if (sgpCurrPkt == nullptr)
+		app_fatal("Failed to allocate memory");
 	sgpCurrPkt->pNext = nullptr;
 	sgpCurrPkt->dwSpaceLeft = sizeof(result->data);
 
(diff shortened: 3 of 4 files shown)


AJenbo (approved):

Source/dthread.cpp
```diff
@@ -80,7 +80,7 @@ void dthread_send_delta(int pnum, _cmd_id cmd, byte *pbSrc, int dwLen)
 		return;
 	}
 
-	pkt = static_cast<TMegaPkt *>(std::malloc(dwLen + 20));
+	pkt = reinterpret_cast<TMegaPkt *>(new byte[dwLen + 20]);
```
ephphatha, line 83: This results in calling `delete` on a pointer allocated with `new[]`, I think this is undefined behaviour? Allocating a buffer smaller than the struct size might already be in that realm anyway so it's likely no worse than the current implementation 😅

vlolteanu: I have some lingering doubts about the UB stuff as well.

I did take a look at what gcc outputs in cases like these and it seems to be correct.

vlolteanu: Might as well cast it back to byte[] and then call delete[] :laughing:

Source/tmsg.cpp
```diff
@@ -29,15 +29,15 @@ size_t tmsg_get(byte *pbMsg)
...
+	delete head;
 	return len;
 }
 
 void tmsg_add(byte *pbMsg, uint8_t bLen)
 {
 	TMsg **tail;
 
-	TMsg *msg = static_cast<TMsg *>(std::malloc(bLen + sizeof(*msg)));
+	TMsg *msg = reinterpret_cast<TMsg *>(new byte[bLen + sizeof(*msg)]);
```
glebm, line 40: This is invalid: `new byte[]` is not guaranteed to be aligned on `alignof(TMsg)`. Use `new TMsg[]` instead.

glebm: Oh I see what this is doing.. It is unfortunately a bit tricky to do corretly in C++.
You have to use both `new` and `delete` with the alignment argument: 

Something like:

```c++
new (std::align_val_t(alignof(TMsg))) byte[bLen + sizeof(TMsg)]
```

and the corresponding `::operator delete[]` with the same alignment value when destroying

This is because `malloc` is guaranteed to align to the largest primitive alignment but `new` is not

Source/dthread.cpp
glebm: Same comment as below applies to this (needs to be an aligned `new []` and a corresponding aligned `delete []`)

Source/tmsg.cpp
vlolteanu: ...and also cast back to byte[] before deleting, to be kosher and avoid any potential UB...

This is more trouble than it's worth. I'm just going to throw bad_alloc if the result is nullptr.

glebm: >  I'm just going to throw bad_alloc if the result is nullptr.

It won't be a nullptr, as it doesn't know that you're going to use it as `TMsg` from the `new byte[]` call alone. If the allocator returns a pointer with an insufficient alignment for `TMsg`, it can crash on access on some CPUs, such as: MIPS, ARM before v7, x86 if the code is auto-vectorized, etc..

vlolteanu: That's not what I meant. I kept malloc and checked the result.

Source/tmsg.cpp
```diff
@@ -38,6 +38,8 @@ void tmsg_add(byte *pbMsg, uint8_t bLen)
 	TMsg **tail;
 
 	TMsg *msg = static_cast<TMsg *>(std::malloc(bLen + sizeof(*msg)));
+	if (msg == nullptr)
+		throw std::bad_alloc();
```
AJenbo, line 42: Would prefer not to introduce exceptions to the code base, maybe we can use `app_fatal` instead?

vlolteanu: Ok, `app_fatal` it is. What message would be appropriate?

`app_fatal("Failed to allocate memory")`?

AJenbo: Sounds good 🙂

AJenbo (approved):

PR #2362, opened 2021-07-12
Refactor NewMonsterAnim and MonsterGraphic enum

obligaron:
After #2354 and #2361 I thought about refactoring monster animation logic a little bit.

Content:

- Less locations where `_mdir` is set
- Less locations that directly work with `AnimStruct`
- Scoped enum `MonsterGraphic`

... and there is much more to refactor but step for step 😉

### DIFF
--- a/Source/dead.cpp
+++ b/Source/dead.cpp
@@ -19,7 +19,7 @@ namespace {
 void InitDeadAnimationFromMonster(DeadStruct &dead, CMonster &mon)
 {
 	int i = 0;
-	auto &animData = mon.GetAnimData(MA_DEATH);
+	auto &animData = mon.GetAnimData(MonsterGraphic::Death);
 	for (const auto &celSprite : animData.CelSpritesForDirections)
 		dead.data[i++] = celSprite->Data();
 	dead.frame = animData.Frames;
--- a/Source/dead.cpp
+++ b/Source/dead.cpp
@@ -16,7 +16,7 @@ DeadStruct Dead[MaxDead];
 int8_t stonendx;
 
 namespace {
-void InitDeadAnimationFromMonster(DeadStruct &dead, CMonster &mon)
+void InitDeadAnimationFromMonster(DeadStruct &dead, const CMonster &mon)
 {
 	int i = 0;
 	auto &animData = mon.GetAnimData(MonsterGraphic::Death);
--- a/Source/dead.cpp
+++ b/Source/dead.cpp
@@ -16,13 +16,14 @@ DeadStruct Dead[MaxDead];
 int8_t stonendx;
 
 namespace {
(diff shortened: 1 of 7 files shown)


Source/monster.h
```diff
@@ -87,13 +87,13 @@ enum MON_MODE : uint8_t {
...
-	MA_GOTHIT,
-	MA_DEATH,
-	MA_SPECIAL,
+enum class MonsterGraphic {
+	Stand,
+	Walk,
+	Attack,
+	GotHit,
+	Death,
+	Special,
```
ephphatha, line 96: 😍

Source/monster.h
```diff
@@ -127,6 +127,13 @@ struct CMonster {
 	/** placeflag enum as a flags*/
 	uint8_t mPlaceFlags;
 	AnimStruct Anims[6];
+	/**
+	 * @brief Returns AnimStruct for specified graphic
```
ephphatha, line 131: Bit nitpicky but this should document the param and return despite those being pretty obvious from the name.

Source/dead.cpp
```diff
@@ -16,13 +16,14 @@ DeadStruct Dead[MaxDead];
 int8_t stonendx;
 
 namespace {
-void InitDeadAnimationFromMonster(DeadStruct &dead, const CMonster &mon)
+void InitDeadAnimationFromMonster(DeadStruct &dead, CMonster &mon)
```
ephphatha, line 19: Providing a const version of CMonster::GetAnimData should allow keeping mon as a const ref

ephphatha (changes requested):
Love it, only suggestion I hope can be implemented is adding a const form of GetAnimData to save changing the signature of InitDeadAnimationFromMonster.

Source/monster.h
obligaron: I'm not a fan of inserting comments that doesn't help anyone.
If you really want it, I can add it.

Something like this?

```
	/**
	 * @brief Returns AnimStruct for specified graphic
	 * @param graphic Specifies which AnimStruct should be returned
	 * @return AnimStruct for the graphic
	 */
```

Source/dead.cpp
obligaron: I got a error in `AnimationInfo` and thought I would need to update many places.
But after trying it, it wasn't many places that hand to be adjusted.
So I added it (last commit) 👍🙂

Source/monster.h
AJenbo: I'm fine with leaving out lines that aren't helpful

ephphatha: In that case no need to worry about this :)

ephphatha (approved):

PR #2363, opened 2021-07-12
Mingw fixes

jmechnich:
This PR addresses a compilation error I encountered while building on Debian testing/bullseye, but also streamlines the mingw-prep scripts in a way that there is only one script left (`mingw-prep.sh`) and `mingw-prep64.sh` just a symbolic link to the other one, so that the architecture is selected from the filename only and there is only one script left to be maintained.


AJenbo (approved):
Thanks

AJenbo:
Looks like this breaks the builds:
```
 Packaging/windows/mingw-prep64.sh: line 37: SUDO: unbound variable
Error: Process completed with exit code 1.
```

AJenbo (approved):

PR #2364, opened 2021-07-12
Fix MAI_Warlord direction

AJenbo:
@obligaron GCC told me you might have missed this one


obligaron (approved):
Really missed this one.
But the only problem with it is that the monster doesn't look to your direction.
I also checked the offer cases again but they look fine.

Issue #2373, opened 2021-07-13
Menu title (translation) broken in character list page


jmechnich:
**Important information**
For me, this issue appears on a MacOS arm64 native, as well as a Windows64-mingw build for the current master branch. The last release 1.2.1 was still working fine. The bug seems to have been introduced in commit 2d3ad5e3f0900c6364a8c7c9d7d4fdda7c9f55a0.

**Describe**
When navigating the menu, titles for single and multiplayer character list pages are "lagging" behind, i.e. when entering the multiplayer page, the title reads "Single Player Characters", and vice versa. Initially, the title will be empty no matter if single or multiplayer has been selected.

**To Reproduce**
1. Start Game and enter main menu.
2. Select e.g. "Single Player" and notice missing title.
3. Return to main menu and select "Single Player" again, correct title is shown.
4. Now return to main menu and select "Multi Player", navigate to character list, "Single Player Characters" is shown as title.

jmechnich:
Reverting 2d3ad5e fixed the issue for me.

PR #2375, opened 2021-07-13
DiabloUI: make UiArtText follow mutating char*

vlolteanu:
There are two instances where a UiArtText of the title is created before setting the title.

The title used to be a char[], so updating the title before the UiArtText is draws used to be sufficient. After a change introduced in https://github.com/diasurgical/devilutionX/pull/2303, the UiArtText retains the contents of title at the time of its instantiation.

This fixes https://github.com/diasurgical/devilutionX/issues/2373 as well as something similar happening for Create/Join TCP Game.

### DIFF
--- a/Source/DiabloUI/ui_item.h
+++ b/Source/DiabloUI/ui_item.h
@@ -135,15 +135,30 @@ class UiArtText : public UiItemBase {
 public:
 	UiArtText(const char *text, SDL_Rect rect, int flags = 0)
 	    : UiItemBase(rect, flags)
+	    , m_text(text)
 	{
 		m_type = UI_ART_TEXT;
-		m_text = text;
 	};
+
+	UiArtText(const char **ptext, SDL_Rect rect, int flags = 0)
+	    : UiItemBase(rect, flags)
+	    , m_ptext(ptext)
+	{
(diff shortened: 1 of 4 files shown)


AJenbo (approved):

vlolteanu:
The issue was deeper than that: the character creation screens reuse the same UiArtText, and they expect simply updating the title to do the trick.

This newer solution lets UiArtText either take a const char * (i.e. text that's set in stone) or a const char ** (i.e. string that might later point somewhere else) upon creation.

Source/DiabloUI/ui_item.h
```diff
@@ -135,15 +135,30 @@ class UiArtText : public UiItemBase {
...
+
+	UiArtText(const char **ptext, SDL_Rect rect, int flags = 0)
+	    : UiItemBase(rect, flags)
+	    , m_ptext(ptext)
 	{
 		m_type = UI_ART_TEXT;
-		m_text = text;
 	};
 
+	const char *get_text() const
```
glebm, line 150: just `text()`, it's unusual to prefix getters with `get` in C++

AJenbo (approved):

PR #2385, opened 2021-07-15
Introduce MissilePosition.StopMissile

### LINKED ISSUE #2371: Succubi Blood Stars do not appear 60fps like other missiles
**Important information**
Windows / Master

**Describe**
Blood Stars do not appear 60fps like other missiles. Loading the same savegame in fluffy's old 60fps mod yields a much smoother blood star. Only tested with succubi, have not seen player blood star.

**To Reproduce**
1. Load savegame

[single_0.zip](https://github.com/diasurgical/devilutionX/files/6806911/single_0.zip)

obligaron:
Fixes #2371

Now when a missile hits, the velocity is set to zero.
So ADL doesn't calculate a new position.
In some cases also the offset is copied (for example firebolt).
This should be okay, cause the offset isn't used (apart from rendering) for the game logic.


AJenbo (approved):
Code also looks cleaner this way :+1:

PR #2386, opened 2021-07-15
Allow heros with same name

obligaron:
## Content

- Replaces all usages of hero name to load/save games with save number
- Allows creation of a hero with same name
- In debug shows save number as information

<details><summary>Debug Sample</summary>

grafik

</details>

I did this cause it was sometimes frustrating to debug reported issues.
I copy the save game into my save folder and sometimes there is already a hero with the same name.
Before this PR it was impossible to select them correctly. It was always the first one loaded.

It would be good if someone else could test this PR before it get merged.

### DIFF
--- a/Source/DiabloUI/diabloui.h
+++ b/Source/DiabloUI/diabloui.h
@@ -53,6 +53,7 @@ struct _uidefaultstats {
 };
 
 struct _uiheroinfo {
+	uint32_t saveNumber;
 	char name[16];
 	uint8_t level;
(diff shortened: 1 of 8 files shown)


AJenbo (approved):
Thanks, this way pretty jankey even if the intent was to prevent duplicate names. Had intented for this change as part of the 1.3.0 clean ups. Going to merge it as it looks good and if there are issues we can resolve them in a follow up.

PR #2388, opened 2021-07-16
Clean up CheckNewPath target loading

AJenbo:
This tries to clean things up a bit in regards to locating the various targets from the globals, there is still a lot that could be done to clean up the function as a whole, and I'm not even sure if this is the right first step, feedback appreciated.


ephphatha (approved):
Looks good as a first step. Splitting this into different functions for interacting with objects, items, talkable monsters, and attack targets would be great for a future cleanup.

PR #2389, opened 2021-07-16
Cleanup monster related logic

AJenbo:


Source/missiles.cpp
```diff
@@ -3594,8 +3610,7 @@ void MI_LightningArrow(int i)
 		if ((mx != Missiles[i]._miVar1 || my != Missiles[i]._miVar2) && mx > 0 && my > 0 && mx < MAXDUNX && my < MAXDUNY) {
 			if (Missiles[i]._misource != -1) {
 				if (Missiles[i]._micaster == TARGET_PLAYERS
-				    && Monsters[Missiles[i]._misource].MType->mtype >= MT_STORM
-				    && Monsters[Missiles[i]._misource].MType->mtype <= MT_MAEL) {
+				    && IsAnyOf(Monsters[Missiles[i]._misource].MType->mtype, MT_STORM, MT_RSTORM, MT_STORML, MT_MAEL)) {
```
obligaron, line 3613: Enum is defined as:
```
	MT_INVALID,
	MT_RSTORM,
	MT_STORM,
	MT_STORML,
	MT_MAEL,
	MT_WINGED,
```

So we should end up with MT_STORM, MT_STORML and MT_MAEL. MT_RSTORM isn't in the range.

```suggestion
				    && IsAnyOf(Monsters[Missiles[i]._misource].MType->mtype, MT_STORM, MT_STORML, MT_MAEL)) {

```

Source/missiles.cpp
```diff
@@ -3811,8 +3826,7 @@ void MI_Lightctrl(int i)
 	    && InDungeonBounds({ mx, my })) {
 		if (id != -1) {
 			if (Missiles[i]._micaster == TARGET_PLAYERS
-			    && Monsters[id].MType->mtype >= MT_STORM
-			    && Monsters[id].MType->mtype <= MT_MAEL) {
+			    && IsAnyOf(Monsters[id].MType->mtype, MT_STORM, MT_RSTORM, MT_STORML, MT_MAEL)) {
```
obligaron, line 3829: Same as above.
```suggestion
			    && IsAnyOf(Monsters[id].MType->mtype, MT_STORM, MT_STORML, MT_MAEL)) {

```

obligaron (changes requested):
Expect the one condition it looks good.
Good cleanup 👍

Source/missiles.cpp
```diff
@@ -466,10 +467,12 @@ void CheckMissileCol(int i, int mindam, int maxdam, bool shift, Point position,
 		return;
 	if (Missiles[i]._micaster != TARGET_BOTH && Missiles[i]._misource != -1) {
 		if (Missiles[i]._micaster == TARGET_MONSTERS) {
-			if (dMonster[mx][my] > 0) {
+			int mid = dMonster[mx][my];
+			if (mid > 0) {
+				mid = mid - 1;
```
ephphatha, line 472: I think decrements are usually done with `postfix--`?

Source/missiles.cpp
```diff
@@ -557,11 +560,13 @@ void CheckMissileCol(int i, int mindam, int maxdam, bool shift, Point position,
 			}
 		}
 	} else {
-		if (dMonster[mx][my] > 0) {
+		int mid = dMonster[mx][my];
+		if (mid > 0) {
+			mid -= 1;
```
ephphatha, line 565: Inconsistent with above, for style purposes would be good to use the same code 😄

Source/missiles.cpp
```diff
@@ -2585,35 +2590,39 @@ void AddStone(int mi, Point /*src*/, Point dst, int /*midir*/, int8_t /*mienemy*
...
-				}
-			}
+			if (!InDungeonBounds({ tx, ty }))
+				continue;
+
+			int mid = dMonster[tx][ty];
+			if (mid == 0)
+				continue;
+
+			auto &monster = Monsters[mid > 0 ? mid - 1 : -(mid + 1)];
```
ephphatha, line 2608:
```suggestion
			auto &monster = Monsters[abs(mid) - 1];
```
IMO the ternary conditional doesn't make the code easy to understand when it's inlined.

Source/missiles.cpp
```diff
@@ -2585,35 +2590,39 @@ void AddStone(int mi, Point /*src*/, Point dst, int /*midir*/, int8_t /*mienemy*
...
+				continue;
+
+			if (IsAnyOf(monster._mmode, MM_FADEIN, MM_FADEOUT, MM_CHARGE))
+				continue;
+
+			found = true;
+			Missiles[mi]._miVar1 = monster._mmode;
+			Missiles[mi]._miVar2 = mid;
+			monster.Petrify();
+			i = 6;
```
ephphatha, line 2620: Putting this last is good, makes it more visible that it's being used as a double break.

Source/missiles.cpp
```diff
@@ -2645,7 +2654,7 @@ void AddGolem(int mi, Point src, Point dst, int /*midir*/, int8_t /*mienemy*/, i
 	Missiles[mi]._miVar2 = src.y;
 	Missiles[mi]._miVar4 = dst.x;
 	Missiles[mi]._miVar5 = dst.y;
-	if ((Monsters[id].position.tile.x != 1 || Monsters[id].position.tile.y != 0) && id == MyPlayerId)
+	if (Monsters[id].position.tile != Point { 1, 0 } && id == MyPlayerId)
```
ephphatha, line 2657: This is crying out for a wrapper method explaining what a golem at position 1,0 means 😂.

Source/missiles.cpp
```diff
@@ -3148,24 +3157,28 @@ void MI_Dummy(int i)
...
-							SpawnGolum(src, { tx, ty }, mi);
-							break;
-						}
-					}
-				}
+
+				if (!InDungeonBounds({ tx, ty }))
+					continue;
+
+				int dp = dPiece[tx][ty];
```
ephphatha, line 3172: As far as I can tell, `LineClearMissile` doesn't modify `dPiece` (or `tx, ty`). This declaration can move to above the conditional on line 3176.

Source/monster.cpp
```diff
@@ -319,7 +319,8 @@ void PlaceGroup(int mtype, int num, int leaderAttributes, int leaderId)
 		while (placed != 0) {
 			ActiveMonsterCount--;
 			placed--;
-			dMonster[Monsters[ActiveMonsterCount].position.tile.x][Monsters[ActiveMonsterCount].position.tile.y] = 0;
+			const Point &position = Monsters[ActiveMonsterCount].position.tile;
```
ephphatha, line 322:
```suggestion
			const auto &position = Monsters[ActiveMonsterCount].position.tile;
```
Most other places are using `auto` instead of an explicit type for these kinds of aliases.

Source/monster.cpp
```diff
@@ -5074,10 +5086,9 @@ bool SpawnSkeleton(int ii, Point position)
 		}
 	}
 
-	int dx = position.x - 1 + x2;
-	int dy = position.y - 1 + y2;
-	Direction dir = GetDirection({ dx, dy }, position);
-	ActivateSpawn(ii, dx, dy, dir);
+	Point spawn = position + Displacement { x2 - 1, y2 - 1 };
```
ephphatha, line 5089: 😑 I see what this is doing now, everything from 5072 to here is trying to find a free adjacent spot in a given Direction. This cleanup made it easy to spot at least 😅

ephphatha (changes requested):

ephphatha:
Apologies, that should've been a comment review instead of request changes. Got confused with a previous PR...

Source/missiles.cpp
AJenbo: Where are you looking that up???

```cpp
	MT_HELLBURN,
	MT_STORM,
	MT_RSTORM,
	MT_STORML,
	MT_MAEL,
	MT_BIGFALL,
```

Source/missiles.cpp
AJenbo: `GolemHoldingCell`

Source/missiles.cpp
AJenbo: Made it `nSolidTable[dPiece[tx][ty]]` like in most other use cases.

Source/monster.cpp
AJenbo: Apparently ALL other places

AJenbo:
> Apologies, that should've been a comment review instead of request changes. Got confused with a previous PR...

You had some good points

ephphatha (approved):
All good imo :)

Source/missiles.cpp
obligaron: Sorry my mistake.
I looked at `MonstConvTbl` in `monsterdat.cpp`.

AJenbo: Good to know :)

PR #2390, opened 2021-07-16
Make MonsterIsTileAccessible aware if monsters can open doors or not

AJenbo:


Source/monster.cpp
```diff
@@ -2046,10 +2046,10 @@ bool RandomWalk2(int i, Direction md)
 /**
  * @brief Check if a tile is affected by a spell we are vunerable to
  */
-bool MonsterIsTileSafe(int i, Point position)
+bool MonsterIsTileSafe(MonsterStruct &monster, Point position)
```
ephphatha, line 2049:
```suggestion
bool MonsterIsTileSafe(const MonsterStruct &monster, Point position)
```
This function doesn't (and shouldn't) modify `monster`.

Source/monster.cpp
```diff
@@ -2098,44 +2098,49 @@ bool MonsterIsTileClear(int i, Point position)
...
 
 /**
- * @brief Check that the amonster that can open doors can stand on the tile
+ * @brief If a monster can access the given tile (possibly by opening a door)
  */
 bool MonsterIsTileAccessible(int i, Point position)
 {
-	bool isdoor = false;
+	MonsterStruct *monster;
+	if (i < 0)
```
ephphatha, line 2110:
```suggestion
	if (i >= 0)
```
Otherwise this is forcing OOB reads. Even better would be to update `MonsterIsTileAccessible` to take a `const MonsterStruct &` 😀(that does require updating the pathfinding code so not suggesting it has to be done now.)

Source/monster.cpp
```diff
@@ -5008,6 +5013,9 @@ void MissToMonst(int i, Point position)
 	}
 }
 
+/**
+ * @brief Check that the given tile is avalible to the monster
```
ephphatha, line 5017:
```suggestion
 * @brief Check that the given tile is available to the monster
```
(should probably fix the typo in the function name too 🤷 )

ephphatha (changes requested):

Source/monster.cpp
AJenbo: Yeah, path finding can also be done for players and nothing, so we would need to adjust for that as well.

Issue #2366, opened 2021-07-12
2x2 items cannot be placed in the upper left corner of the inventory


Chance4us:
Video:

https://user-images.githubusercontent.com/45902449/125354707-6a06d700-e364-11eb-9c4b-a10ad06af4af.mp4

Savegame:
[magic_rock.zip](https://github.com/diasurgical/devilutionX/files/6804297/magic_rock.zip)

julealgon:
Is it exclusive to magic rock? Can you test with other items? Other 2x2 items perhaps?

Chance4us:
I can confirm that it also affects other 2x2 items tested on a book of reflect.

Chance4us:
The issue was introduced with commit fd3893190fc6822ed5676a81feab0a1e72f2452c

PR #2394, opened 2021-07-16
fix upper left inventory corner

qndel:
This reverts this part of the code to before my cleanup, 
I added some log
INFO: r == slotxy_chest_last
INFO: itemSize.width & 1
INFO: itemSize.height & 1
INFO: DONE WITH SLOT: 24
INFO: r == slotxy_chest_last
INFO: itemSize.width & 1
INFO: itemSize.height & 1
INFO: GOING OUT WITH SLOT: 25

so it seems the "working" code actually hits that if twice, I have no clue how inventory works at this point so just leaving it alone

Resolves https://github.com/diasurgical/devilutionX/issues/2366


Chance4us:
Thanks @qndel

PR #2396, opened 2021-07-16
Use bound parameters for the test function passed to LineClear

ephphatha:
Bit of a proof of concept, this can be applied to the pathfinding code as well. I'll give that a shot tomorrow.


AJenbo:
Merged via https://github.com/diasurgical/devilutionX/pull/2399

PR #2397, opened 2021-07-16
Use bound parameters for pathfinding functions operating on entities

ephphatha:
Once #2390 is merged I'll update the signature of `MonsterIsTileAccessible` to use a `MonsterStruct` reference so the call to `FindPath` in monster.cpp can use `std::cref(Monsters[i])`.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -91,7 +91,7 @@ int GetDistance(Point destination, int maxDistance)
 	}
 
 	int8_t walkpath[MAX_PATH_LENGTH];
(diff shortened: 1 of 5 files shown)


AJenbo:
Actually I think thease functions should not be chained like they currently are, the only reason MonsterIsTileClear takes a reference to a monster is to pass it to MonsterIsTileSafe. Instead the top level function should compose what it needs and MonsterIsTileClear should not even be aware that the game has monsters.

Bind may still be needed for this approach so thanks for making me read about it :D

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/pull/2399

Please see https://github.com/diasurgical/devilutionX/pull/2398 for my follow up :)

PR #2398, opened 2021-07-16
Cleanup Pathing

AJenbo:


Source/path.cpp
```diff
@@ -56,6 +57,36 @@ const Displacement PathDirs[8] = {
...
 int8_t path_directions[9] = { 5, 1, 6, 2, 0, 3, 8, 4, 7 };
 
+bool IsTileNotSolid(Point position)
+{
+	return !nSolidTable[dPiece[position.x][position.y]];
+}
+
+bool IsTileSolid(Point position)
+{
+	if (position.x < 0 || position.y < 0 || position.x >= MAXDUNX || position.y >= MAXDUNY) {
```
ephphatha, line 67: This is the proper form for an `InDungeonBounds` check. Could we extract this code to a function (and rename the missiles.cpp version until the logic issues in MissileProcs are fixed 😂 )?

This test is used in multiple places in the code as well. I haven't flagged any of them for this PR, mentioning it more to say this would be useful in more places than just these two functions.

Source/path.cpp
```diff
@@ -56,6 +57,36 @@ const Displacement PathDirs[8] = {
  */
 int8_t path_directions[9] = { 5, 1, 6, 2, 0, 3, 8, 4, 7 };
 
+bool IsTileNotSolid(Point position)
+{
+	return !nSolidTable[dPiece[position.x][position.y]];
```
ephphatha, line 62: I think this should do the same bounds checking as IsTileSolid to guard against OOB reads. Calling `!IsTileSolid` wouldn't be appropriate as that would return true for out of bounds positions, would need to replicate the check and return false.

Source/path.cpp
```diff
@@ -56,6 +57,36 @@ const Displacement PathDirs[8] = {
...
+	return nSolidTable[dPiece[position.x][position.y]];
+}
+
+/**
+ * @brief Checks the position is solid or blocked by an object
+ */
+bool IsTileWalkable(Point position, bool ignoreDoors)
+{
+	if (dObject[position.x][position.y] != 0) {
+		int oi = dObject[position.x][position.y] > 0 ? dObject[position.x][position.y] - 1 : -(dObject[position.x][position.y] + 1);
```
ephphatha, line 80:
```suggestion
		int oi = abs(dObject[position.x][position.y]) - 1;
```
Saves having to index in to the array multiple times.

Source/player.cpp
```diff
@@ -147,11 +147,11 @@ bool PlrDirOK(int pnum, Direction dir)
 	}
 
 	if (dir == DIR_E) {
-		return !SolidLoc(position + DIR_SE) && (dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR) == 0;
+		return !IsTileSolid(position + DIR_SE) && (dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR) == 0;
```
ephphatha, line 150:
```suggestion
		return IsTileNotSolid(position + DIR_SE) && (dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR) == 0;
```
`!IsTileSolid` is true for out of bounds positions. This is a pre-existing bug that theoretically could lead to OOB reads with the dFlags check.

Source/player.cpp
```diff
@@ -147,11 +147,11 @@ bool PlrDirOK(int pnum, Direction dir)
 	}
 
 	if (dir == DIR_E) {
-		return !SolidLoc(position + DIR_SE) && (dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR) == 0;
+		return !IsTileSolid(position + DIR_SE) && (dFlags[position.x + 1][position.y] & BFLAG_PLAYERLR) == 0;
 	}
 
 	if (dir == DIR_W) {
-		return !SolidLoc(position + DIR_SW) && (dFlags[position.x][position.y + 1] & BFLAG_PLAYERLR) == 0;
+		return !IsTileSolid(position + DIR_SW) && (dFlags[position.x][position.y + 1] & BFLAG_PLAYERLR) == 0;
```
ephphatha, line 154:
```suggestion
		return IsTileNotSolid(position + DIR_SW) && (dFlags[position.x][position.y + 1] & BFLAG_PLAYERLR) == 0;
```
Same issue here.

Source/objects.cpp
```diff
@@ -670,7 +670,7 @@ void AddObjTraps()
 
 			if (GenerateRnd(2) == 0) {
 				int xp = i - 1;
-				while (!nSolidTable[dPiece[xp][j]]) // BUGFIX: check if xp >= 0
+				while (IsTileNotSolid({ xp, j })) // BUGFIX: check if xp >= 0
```
ephphatha, line 673: Given this use of the function (and the comment 😀) making `IsTileNotSolid` return false for out of bounds positions would be good.

ephphatha (commented):
Using lambdas instead of `std::bind` looks good. Thanks @glebm for pointing that out, I didn't realise the issues with bind in C++.

Extracting the logic into functions has highlighted inconsistencies with the way certain tests are performed that I think would be worth changing, but nothing I'd consider a blocker.

Source/monster.cpp
```diff
@@ -2207,7 +2184,7 @@ bool AiPlanPath(int i)
 	}
 
 	bool clear = LineClear(
-	    std::bind(MonsterIsTileAvailable, i, std::placeholders::_1),
+	    [monster](Point position) { return IsTileAvailable(monster, position); },
```
glebm, line 2187: This captures `monster` by-value (copies it). Use `&monster` instead to capture by reference.

Source/path.cpp
AJenbo: Totally agree, but decided not to do so for fear I might apply it to all cases and grow the PR further

Source/path.cpp
AJenbo: Wish some one had pointed this out earlier, we should be using abs() every where for this :D

Source/path.cpp
AJenbo: I think IsTileSolid should return true for OOB, but I would like to wait with both of these changes as there is some potential they could affecting various things at this point. So prefer getting things stable and tested first.

Source/objects.cpp
AJenbo: Almost did it when I noticed this case as well, but instead i think it serves as a good reminder to do this later.

PR #2399, opened 2021-07-16
Use bound parameters for the test function passed to LineClear

AJenbo:
Rebased https://github.com/diasurgical/devilutionX/pull/2396 cause i got excited


Source/controls/plrctrls.cpp
```diff
@@ -91,7 +91,7 @@ int GetDistance(Point destination, int maxDistance)
 	}
 
 	int8_t walkpath[MAX_PATH_LENGTH];
-	int steps = FindPath(PosOkPlayer, MyPlayerId, Players[MyPlayerId].position.future.x, Players[MyPlayerId].position.future.y, destination.x, destination.y, walkpath);
+	int steps = FindPath(std::bind(PosOkPlayer, MyPlayerId, std::placeholders::_1), Players[MyPlayerId].position.future.x, Players[MyPlayerId].position.future.y, destination.x, destination.y, walkpath);
```
glebm, line 94: `std::bind` is basically obsolete and has issues. Use a lambda instead.
https://abseil.io/tips/108

AJenbo: To bad, I like the bind syntax better then the lambda syntax... but on the other hand I like lambda better as a concept...

AJenbo: Fixed via #2398, thanks for pointing it out.

AJenbo: Oh, nice using lambda also solves the issue with bind() seeing `IsTileAvailable(monster, Point)` being ambiguous so now I don't have to prefix it with `Monster` anymore :+1:

PR #2403, opened 2021-07-17
Add tests and refactor path functions to use Point/Displacement

ephphatha:
The FindPath function has some pretty basic test coverage as I don't think it really needs to be deterministic. The tests I did add were simply enough to validate I didn't change the behaviour while I was moving code around.


Source/engine/point.hpp
```diff
@@ -149,6 +152,19 @@ struct Point {
 
 		return std::max<int>(offset.deltaX, offset.deltaY);
 	}
+
+	#ifdef RUN_TESTS
```
AJenbo, line 156:
```suggestion
#ifdef RUN_TESTS
```

same with `#endif`

Source/objects.h
```diff
@@ -121,6 +121,15 @@ struct ObjectStruct {
 		SetMapRange(mapRange);
 		_oVar8 = leverID;
 	}
+
+	/**
+	 * @brief Check if this object is a door
+	 * @return True if the object is one of the door types (see _object_id)
+	 */
+	bool IsDoor() const
```
AJenbo, line 129: Nice, I was thinking this would make sens :D

Source/path.h
```diff
@@ -28,23 +28,49 @@ struct PATHNODE {
...
+const Displacement PathDirs[8] = {
+	// clang-format off
+	{ -1, -1 }, //DIR_N
+	{ -1,  1 }, //DIR_W
+	{  1, -1 }, //DIR_E
+	{  1,  1 }, //DIR_S
+	{ -1,  0 }, //DIR_NW
+	{  0, -1 }, //DIR_NE
+	{  1,  0 }, //DIR_SE
+	{  0,  1 }, //DIR_SW
```
AJenbo, line 72: Any reason not to convert this to an array of directions now?

ephphatha: FindMeleeTarget uses it in a rather annoying way, but mostly I was trying to work out if the order mattered so it could be replaced with one of the other collections of `Direction` enum values excluding DIR_OMNI. Didn't want to get too deep into that refactoring in this PR.

AJenbo (dismissed):

AJenbo:
Really enjoying the back and worth we are having with working on this part of the code :D

AJenbo:
Nice work with the test, all changes covered and a decent bump to total coverage:
image
path.cpp is now the largest well tested file in the project.

AJenbo (approved):

PR #2410, opened 2021-07-17
Convert anim_weapon_id & anim_armor_id to enum class and remove _pwtype

obligaron:


Source/items.cpp
```diff
@@ -119,11 +119,11 @@ namespace {
 
 std::optional<CelSprite> itemanims[ITEMTYPES];
 
-enum anim_armor_id : uint8_t {
+enum class anim_armor_id : uint8_t {
```
AJenbo, line 122:
```suggestion
enum class PlayerArmorGraphics : uint8_t {
```

test/writehero_test.cpp
```diff
@@ -306,7 +306,6 @@ static void AssertPlayer(PlayerStruct &player)
 	ASSERT_EQ(player._pAblSpells, 134217728);
 	ASSERT_EQ(player._pScrlSpells, 0);
 	ASSERT_EQ(player._pSpellFlags, 0);
-	ASSERT_EQ(player._pwtype, 1);
```
AJenbo, line 309: assert true UsesRangedWeapon() to maintain coverage and test your new code.

Source/items.cpp
obligaron: I like it. 🙂 

I renamed the enum to `PlayerArmorGraphic`, cause we normally use singular for enum class.
I also renamed `anim_weapon_id` to `PlayerWeaponGraphic` so that both matches.

test/writehero_test.cpp
obligaron: done 🙂

Source/items.cpp
AJenbo: Nice :)

AJenbo (approved):

PR #2415, opened 2021-07-17
Remove some more use of miniwin

AJenbo:


Source/DiabloUI/ui_item.h
```diff
@@ -340,7 +340,7 @@ class UiList : public UiItemBase {
 	int indexAt(Sint16 y) const
 	{
 		ASSERT(y >= m_rect.y);
-		const int index = (y - m_rect.y) / m_height;
+		const size_t index = (y - m_rect.y) / m_height;
```
obligaron, line 343: @AJenbo With this changes the warnings for `ui_item.h` are back. Did the changes that was done in #2414 caused some problems? 🤔

AJenbo: Yeah, GCC and CLang was spamming warnings as a result :D

Fixing it would be good, but we need to test the solution with more then one compiler apparently.

PR #2418, opened 2021-07-18
Mouse actions

AJenbo:


obligaron (approved):
Looks and plays fine and is a good improvement. 👍 

One thing let me think a little bit, but it's a edge case:
If you request a melee attack to a monster that you still need to walk to.
You start walking to the monster but the monster dies (cause of a spell, trap or another player).
Before the change we would still walk to the monster tile.
After the change we would stop, cause `player.Stop()` also clears walkpath.
This is a behavior change, but this could also be defined as a improvement. 😉 

btw: I tried to test the change in multiplayer. But I got some not so nice desyncs. But the desyns also exists with master, so I they can't be introduced with your changes.

AJenbo:
>  After the change we would stop, cause `player.Stop()` also clears walkpath.

Yes, this is intentional since the target has now become invalid. Previously you would even take a swing at the corpse once you reach it.

PR #2419, opened 2021-07-18
dthread: simplify synchronization logic

vlolteanu:
Dthread had some convoluted synchronization logic, and made use of 2 mutexes and 1 condition variable. It lends itself to a more standard producer-consumer approach.

On the consumer: Drain the message queue in a loop. Once empty, wait for a signal.

On the producer: Signal whenever something new is placed in the queue. If this occurs during `multi_send_zero_packet`, the signal is lost, but the consumer checks the queue again.

When killing the thread: if this happens during `multi_send_zero_packet`, `DthreadRunning = false` is sufficient to kill the thread. Otherwise, the thread is signaled again and eventually dies.

### DIFF
--- a/Source/dthread.cpp
+++ b/Source/dthread.cpp
@@ -4,12 +4,11 @@
  * Implementation of functions for updating game state from network commands.
  */
 
-#include <atomic>
 #include <list>
 #include <mutex>
 
 #include "nthread.h"
-#include "utils/sdl_mutex.h"
+#include "utils/sdl_cond.h"
 #include "utils/thread.h"
 
 namespace devilution {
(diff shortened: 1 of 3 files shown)


vlolteanu:
Should we get rid of event_emul? It isn't used anywhere else.

Source/dthread.cpp
```diff
@@ -77,36 +74,40 @@ void dthread_send_delta(int pnum, _cmd_id cmd, std::unique_ptr<byte[]> data, uin
...
+	if (!DthreadRunning)
 		return;
 
-	dthread_running = false;
-	SetEvent(sghWorkToDoEvent);
+	DthreadMutex->lock();
+	DthreadRunning = false;
+	InfoList.clear();
+	WorkToDo->signal();
+	DthreadMutex->unlock();
```
glebm, line 102:
```suggestion
	{
		std::lock_guard<SdlMutex> lock(*DthreadMutex);
		DthreadRunning = false;
		InfoList.clear();
		WorkToDo->signal();
	}
```

glebm (approved):

Issue #2412, opened 2021-07-17
Monsters not syncing


vlolteanu:
**Important information**
Linux / 1.2.1-0233b12b / compiled by myself

**Describe**
Monster kills on one client do not register on another client. From the other client's point of view, dead monsters go about their business.

**To Reproduce**
1. Create a multiplayer game over TCP.
2. Start a second client on the same machine.
3. Join the TCP game running on 127.0.0.1
4. Have both characters enter the cathedral.
5. Kill a monster with one of the characters
6. Switch to the other client

**Expected behavior**
The monster should also be dead on the other client.

**Additional context**
Found this while testing https://github.com/diasurgical/devilutionX/pull/2411

vlolteanu:
I've done a git bisect and apparently this commit broke it: 2bdb6caa07398dbdda5c1c4319cf6a82025c85be

PR #2420, opened 2021-07-18
Fix monster sync on kill

vlolteanu:
Fixes https://github.com/diasurgical/devilutionX/issues/2412

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4203,7 +4203,7 @@ void M_SyncStartKill(int i, Point position, int pnum)
 	assert(i >= 0 && i < MAXMONSTERS);
 	auto &monster = Monsters[i];
 
-	if (monster._mhitpoints > 0 || monster._mmode == MM_DEATH) {
+	if (monster._mhitpoints == 0 || monster._mmode == MM_DEATH) {
 		return;
 	}
 


obligaron:
The bug was introduced with #2234
Should we check for `<= 0`?

glebm:
Interestingly, elsewhere in the code there is a check like this `monster._mhitpoints >> 6 <= 0`

obligaron:
> 
> 
> Interestingly, elsewhere in the code there is a check like this `_mhitpoints >> 6 <= 0`

Not everywhere:

- [msg.cpp](https://github.com/diasurgical/devilutionX/blob/0889780923262ee8fecfe2d07c33272c58ab5bc1/Source/msg.cpp#L393)
- [msg.cpp](https://github.com/diasurgical/devilutionX/blob/0889780923262ee8fecfe2d07c33272c58ab5bc1/Source/msg.cpp#L1329)
- [missiles.cpp](https://github.com/diasurgical/devilutionX/blob/0889780923262ee8fecfe2d07c33272c58ab5bc1/Source/missiles.cpp#L4246)
- [objects.cpp](https://github.com/diasurgical/devilutionX/blob/0889780923262ee8fecfe2d07c33272c58ab5bc1/Source/objects.cpp#L4433)

But don't ask me if this is correct or not. At least it's confusing.
Anyway I think it would be good if we add a helper method to clarify this.

ephphatha:
Adding a fixed point wrapper for these values would be great. That'd allow clarifying how HP/Mana values (6 bit fractional part) are handled and would also be useful for tile offsets (15 or 16 bit fractional parts depending on the use...).

As for the hitpoints > 0 checks, most of the time this check looks only at the integer portion so a monster with .077 hp is considered dead. I think it should use the form glebm pointed out for now.

AJenbo (approved):

AJenbo:
When the monster is will it's HP is specifically set to 0, going to merge this for now and then we can clean it up later when adding a helper and do some some testing

PR #2421, opened 2021-07-18
Fix separated monster pack handling

### LINKED ISSUE #2413: Groups can get stuck in a special move stance
**Important information**
Test version https://github.com/diasurgical/devilutionX/commit/3ffedb7a6973d7a151c50587dae7eb7368d1aaeb

**Describe**
Bongo and his minion started jumping on the spot constantly never moving. If you walk next to him he will still attack.

**To Reproduce**
1. Load save game, go to level 3 and find bongo

Before issue:
[single_3.zip](https://github.com/diasurgical/devilutionX/files/6834908/single_3.zip)

After issue:
[single_3.zip](https://github.com/diasurgical/devilutionX/files/6834900/single_3.zip)
(bongo is permanently in this broken state)

obligaron:
Fixes #2413

Bug was introduced in #2379 commit a44781bd73d440628a096f4e79e4dbc27cbcc590

`packsize` got a negative overflow.
This is cause the [new check](https://github.com/diasurgical/devilutionX/pull/2379/commits/a44781bd73d440628a096f4e79e4dbc27cbcc590#diff-d46585089d315b53630b1035442b3d490d7b422cb05ceee76cf9041de3e39179R1966) can [subtract separated minions](https://github.com/diasurgical/devilutionX/pull/2379/commits/a44781bd73d440628a096f4e79e4dbc27cbcc590#diff-d46585089d315b53630b1035442b3d490d7b422cb05ceee76cf9041de3e39179R1972-R1973) multiple times from `packsize`.

First I reverted to the original logic (first commit), refactored the logic again (second commit) and last did some renaming to reduce future misunderstandings (third commit).

Notes:

- Existing saves can't be fixed by this pr, cause the wrong `packsize` is already saved.
- Now I have enough from Bongo, cause I killed his pack around 20 times.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -220,7 +220,7 @@ void InitMonster(MonsterStruct &monster, Direction rd, int mtype, Point position
 	monster.mArmorClass = monsterType.MData->mArmorClass;
 	monster.mMagicRes = monsterType.MData->mMagicRes;
 	monster.leader = 0;
-	monster.leaderflag = MonsterRelation::Individual;
+	monster.leaderRelation = LeaderRelation::None;
 	monster._mFlags = monsterType.MData->mFlags;
 	monster.mtalkmsg = TEXT_NONE;
 
@@ -361,7 +361,7 @@ void PlaceGroup(int mtype, int num, int leaderAttributes, int leaderId)
 
 				if ((leaderAttributes & 2) != 0) {
 					minion.leader = leaderId;
-					minion.leaderflag = MonsterRelation::Minion;
+					minion.leaderRelation = LeaderRelation::Leashed;
 					minion._mAi = leader._mAi;
 				}
 
@@ -1949,25 +1949,25 @@ void GroupUnity(int i)
 	assert(i >= 0 && i < MAXMONSTERS);
 	auto &monster = Monsters[i];
 
-	if (monster.leaderflag == MonsterRelation::Individual)
+	if (monster.leaderRelation == LeaderRelation::None)
 		return;
 
 	assert(monster.leader >= 0);
 
 	auto &leader = Monsters[monster.leader];
 	if (IsLineNotSolid(monster.position.tile, leader.position.future)) {
-		if (monster.leaderflag == MonsterRelation::Leader
+		if (monster.leaderRelation == LeaderRelation::Separated
 		    && monster.position.tile.WalkingDistance(leader.position.future) < 4) {
 			// Reunite the separated monster with the pack
 			leader.packsize++;
-			monster.leaderflag = MonsterRelation::Minion;
+			monster.leaderRelation = LeaderRelation::Leashed;
 		}
-	} else if (monster.leaderflag == MonsterRelation::Minion) {
+	} else if (monster.leaderRelation == LeaderRelation::Leashed) {
 		leader.packsize--;
-		monster.leaderflag = MonsterRelation::Leader;
+		monster.leaderRelation = LeaderRelation::Separated;
 	}
 
-	if (monster.leaderflag == MonsterRelation::Minion) {
+	if (monster.leaderRelation == LeaderRelation::Leashed) {
 		if (monster._msquelch > leader._msquelch) {
 			leader.position.last = monster.position.tile;
 			leader._msquelch = monster._msquelch - 1;
@@ -1983,7 +1983,7 @@ void GroupUnity(int i)
 		return;
 	for (int j = 0; j < ActiveMonsterCount; j++) {
 		auto &minion = Monsters[ActiveMonsters[j]];
-		if (minion.leaderflag != MonsterRelation::Minion || minion.leader != i)
+		if (minion.leaderRelation != LeaderRelation::Leashed || minion.leader != i)
 			continue;
 		if (monster._msquelch > minion._msquelch) {
 			minion.position.last = monster.position.tile;
@@ -2497,9 +2497,9 @@ void ScavengerAi(int i)
 	if (monster._mmode != MM_STAND)
 		return;
 	if (monster._mhitpoints < (monster._mmaxhp / 2) && monster._mgoal != MGOAL_HEALING) {
-		if (monster.leaderflag != MonsterRelation::Individual) {
+		if (monster.leaderRelation != LeaderRelation::None) {
 			Monsters[monster.leader].packsize--;
-			monster.leaderflag = MonsterRelation::Individual;
+			monster.leaderRelation = LeaderRelation::None;
 		}
 		monster._mgoal = MGOAL_HEALING;
 		monster._mgoalvar3 = 10;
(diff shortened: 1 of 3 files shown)


Source/monster.cpp
```diff
@@ -1949,50 +1949,51 @@ void GroupUnity(int i)
 	assert(i >= 0 && i < MAXMONSTERS);
 	auto &monster = Monsters[i];
 
+	if (monster.leaderflag == MonsterRelation::Individual)
+		return;
+
+	assert(monster.leader >= 0);
```
AJenbo, line 1955: arh, this explains it, I thought this function was called both for the leader and the minions. Thanks for spending the time to correct this.

Source/monster.cpp
```diff
@@ -1949,50 +1949,51 @@ void GroupUnity(int i)
...
+	if (IsLineNotSolid(monster.position.tile, leader.position.future)) {
+		if (monster.Relation == MonsterRelation::SeparatedMinion
+		    && monster.position.tile.WalkingDistance(leader.position.future) < 4) {
+			// Reunite the separated monster with the pack
+			leader.packsize++;
+			monster.Relation = MonsterRelation::Minion;
 		}
+	} else if (monster.Relation == MonsterRelation::SeparatedMinion) {
+		// Monster is already separated from pack
+	} else {
```
AJenbo, line 1967:
```suggestion
	} else if (monster.Relation == MonsterRelation::Minion) {
```

Source/monster.cpp
```diff
@@ -1949,50 +1949,51 @@ void GroupUnity(int i)
...
+			leader._mmode = MM_SATTACK;
+		}
+		return;
 	}
 
-	if (monster.leaderflag == MonsterRelation::Leader && monster._uniqtype != 0) {
-		if ((UniqMonst[monster._uniqtype - 1].mUnqAttr & 2) == 0) {
-			return;
+	if (monster._uniqtype == 0 || (UniqMonst[monster._uniqtype - 1].mUnqAttr & 2) == 0)
+		return;
```
AJenbo, line 1985: Uniques are never also minions so the code past this point would be unused afaik?

Source/monster.h
```diff
@@ -199,7 +209,7 @@ struct MonsterStruct { // note: missing field _mAFNum
 	uint16_t mMagicRes;
 	_speech_id mtalkmsg;
 	uint8_t leader;
-	MonsterRelation leaderflag;
+	MonsterRelation Relation;
```
AJenbo, line 212: Wrong casing, but also this might be better names:
```suggestion
	LeaderRelation leaderRelation;
```

Source/monster.h
```diff
@@ -115,10 +115,20 @@ enum placeflag : uint8_t {
 	// clang-format on
 };
 
+/**
+ * @brief Defines the relation of the monster to a monster pack.
+ *        If value is differnt from Individual MonsterStruct.leader must also be set
+ */
 enum class MonsterRelation : uint8_t {
 	Individual,
```
AJenbo, line 123:
```suggestion
	None,
```

Source/monster.h
```diff
@@ -115,10 +115,20 @@ enum placeflag : uint8_t {
...
+	/**
+	 * @brief Minion that sticks with the leader
+	 */
 	Minion,
-	Leader,
+	/**
+	 * @brief Minion that was separated from leader and acts individual until it reaches the leader again
+	 */
+	SeparatedMinion,
```
AJenbo, line 131:
```suggestion
	Separated,
```

Source/monster.h
```diff
@@ -115,10 +115,20 @@ enum placeflag : uint8_t {
...
+/**
+ * @brief Defines the relation of the monster to a monster pack.
+ *        If value is differnt from Individual MonsterStruct.leader must also be set
+ */
 enum class MonsterRelation : uint8_t {
 	Individual,
+	/**
+	 * @brief Minion that sticks with the leader
+	 */
 	Minion,
```
AJenbo, line 127:
```suggestion
	Leashed,
```

Source/monster.cpp
AJenbo: Hmm apparently I'm wrong about this as well in some cases...

AJenbo: Did a small hack to spawn all unique monsters and indeed nothing ever gets past this point. But I did bump in to some interesting unused unique monsters :)

Here is a few of them:
image
image
image
image

Source/monster.cpp
obligaron: Initially I thought the same 😄 
Then my debugger told me otherwise. First I thought wtf and added debug output for `leader`/`minion` flag and was a little bit surprised. 😁

Source/monster.cpp
obligaron: Good idea, thanks 🙂

Source/monster.cpp
obligaron: I also looked again at the code and think that's impossible.
So I removed the code and added an assert. 👍 

btw: sometimes I think we should add a flag in the ini to enable all the missing content. similar to bard etc. But hey we still have some versions of devilutionX that will come in the future. 😁

Source/monster.h
obligaron: Done 🙂 
I also thought about `PackRelation` but `LeaderRelation` is also good. 👍

Source/monster.h
obligaron: Done 🙂

Source/monster.h
obligaron: Done 🙂

Source/monster.h
obligaron: Done 🙂

Source/monster.h
AJenbo: It relates back to the original name so might be easier for people hew about the old one :shrug:

Source/monster.cpp
AJenbo: Good thinking with the assert.

Source/monster.cpp
```diff
@@ -1949,51 +1949,37 @@ void GroupUnity(int i)
...
-			}
-			if (monster._msquelch > minion._msquelch) {
-				minion.position.last = monster.position.tile;
-				minion._msquelch = monster._msquelch - 1;
-			}
+		if (leader._mAi == AI_GARG && (leader._mFlags & MFLAG_ALLOW_SPECIAL) != 0) {
+			leader._mFlags &= ~MFLAG_ALLOW_SPECIAL;
+			leader._mmode = MM_SATTACK;
 		}
+		return;
```
AJenbo, line 1982:
```suggestion
```

obligaron: lol fixed 🙂

AJenbo (approved):
Nice work, this code just became tons easier to read.

AJenbo:
Oh, removing the dead code also made this able to take MonsterStruct as an input :)

PR #2422, opened 2021-07-18
msg.cpp: get rid of excessive indentation

vlolteanu:


Source/msg.cpp
```diff
@@ -2003,6 +2004,26 @@ void DeltaSaveLevel()
...
+				if (CanPut({ xx, yy }))
+					return { xx, yy };
+
+			}
+		}
+	}
+
+	return position;
+}
+
```
AJenbo, line 2026: Put this in an anonymous namespace instead of marking it as static.

AJenbo (approved):

PR #2430, opened 2021-07-19
Have PosOkPlayer take PlayerStruct as input

AJenbo:


Source/player.cpp
```diff
@@ -319,16 +319,11 @@ void ScrollViewPort(const PlayerStruct &player, _scroll_direction dir)
 	}
 }
 
-bool PlrDirOK(int pnum, Direction dir)
+bool PlrDirOK(PlayerStruct &player, Direction dir)
```
ephphatha, line 322:
```suggestion
bool PlrDirOK(const PlayerStruct &player, Direction dir)
```
This can take a const reference once PosOkPlayer does.

ephphatha (changes requested):
One inverted check I noticed, otherwise looks good. Potential for a couple of methods to take const references assuming direct property reads doesn't cause issues (otherwise may need const getters)

Source/player.h
```diff
@@ -486,8 +485,8 @@ void RestartTownLvl(int pnum);
 void StartWarpLvl(int pnum, int pidx);
 void ProcessPlayers();
 void ClrPlrPath(PlayerStruct &player);
-bool PosOkPlayer(int pnum, Point position);
-void MakePlrPath(int pnum, Point targetPosition, bool endspace);
+bool PosOkPlayer(PlayerStruct &player, Point position);
```
AJenbo, line 488:
```suggestion
bool PosOkPlayer(const PlayerStruct &player, Point position);
```

ephphatha (approved):

AJenbo:
@ephphatha thanks for the review :+1:

PR #2437, opened 2021-07-21
Introduce std::thread-like class; get rid of utils/thread.{h,cpp}

vlolteanu:


AJenbo:
@vlolteanu could you share some of the thinking behind this change.

@StephenCWills could you test this on 3DS since it's a bit sensitive to having this changed?

vlolteanu:
This is a straightforward imitation of std::thread, albeit using SDL_Thread underneath. The basics apply: it can't be copied, only moved; destroying without joining and joining from the same thread are considered errors and terminate the program.

The old stuff was an imitation of the Windows API.

Functionally, I didn't change anything.

AJenbo (approved):

StephenCWills:
> could you test this on 3DS since it's a bit sensitive to having this changed?

It seems the latest master crashes at startup on 3DS so my test was inconclusive.

AJenbo:
Ok, thanks for testing. in that case I think we won't have it block this PR. Hopefully it won't cause any issues with getting 3DS back up and running.

StephenCWills:
I was able to fix the crash. I ran the test again, and there's no problem.

PR #2440, opened 2021-07-21
Refactor MisFileData

vlolteanu:
This makes MissileSpriteData a lot more legible and easy to expand, if needed. It also gets rid of manual memory management for MisFileData.

I've checked to see if the new table is equivalent to the old.


Source/misdat.h
```diff
@@ -121,19 +122,57 @@ typedef struct MissileData {
...
+	uint8_t animName;
+	uint8_t animFAmt;
+	uint32_t flags;
+	std::array<byte *, 16> animData = {};
+	std::array<uint8_t, 16> animDelay = {};
+	std::array<uint8_t, 16> animLen = {};
+	std::array<int16_t, 16> animWidth = {};
+	std::array<int16_t, 16> animWidth2 = {};
+
+	std::vector<std::unique_ptr<byte[]>> pinnedMem;
```
ephphatha, line 136: Does this need to be exposed? Could be worth changing MisFileData to a class and making this private.

vlolteanu: You're right, pinnedMem doesn't have to be exposed.

vlolteanu: The rest of them do.

Source/misdat.cpp
```diff
@@ -127,68 +129,162 @@ MissileDataStruct MissileData[] = {
...
-	{ "",         MFILE_NONE,        0, 0, { nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr }, { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }, {  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0 }, {   0,   0,   0,   0,   0,   0,   0,   0,  0,  0,  0,  0,  0,  0,  0,  0 }, {   0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0 } },
+	// name,      animName,        animFAmt, flags,                animDelay[16], animLen[16],                           animWidth[16],          animWidth2[16]
+	{ "Arrows",   MFILE_ARROWS,     1,       MFLAG_LOCK_ANIMATION,    0,            16,                                     96,                     16               },
```
AJenbo, line 133: Using `monster_flag` for the missile flags is wrong, instead a separate `enum` should be created:
```cpp
enum class MissileDataFlags {
	NoGraphics    = 1 << 0,
	NotAnimated  = 1 << 1,
}
```

Source/misdat.h
```diff
@@ -121,19 +122,59 @@ typedef struct MissileData {
...
-	uint8_t mAnimName;
-	uint8_t mAnimFAmt;
-	uint32_t mFlags;
-	byte *mAnimData[16];
-	uint8_t mAnimDelay[16];
-	uint8_t mAnimLen[16];
-	int16_t mAnimWidth[16];
-	int16_t mAnimWidth2[16];
-} MisFileDataStruct;
+struct MisFileData {
```
AJenbo, line 125:
```suggestion
struct MissileFileData {
```

Source/missiles.cpp
```diff
@@ -1263,50 +1243,32 @@ void SetMissDir(int mi, int dir)
...
 	}
 }
 
 void FreeMissiles()
 {
-	for (int mi = 0; MissileSpriteData[mi].mAnimFAmt != 0; mi++) {
-		if ((MissileSpriteData[mi].mFlags & MFLAG_HIDDEN) == 0)
-			FreeMissileGFX(mi);
+	for (int mi = 0; MissileSpriteData[mi].animFAmt != 0; mi++) {
+		if ((MissileSpriteData[mi].flags & MFLAG_HIDDEN) == 0)
```
AJenbo, line 1262: Do we even need to make this check now that `FreeGFX()` operates on a vector?

Source/misdat.cpp
```diff
@@ -127,68 +129,162 @@ MissileDataStruct MissileData[] = {
...
+}
+
+MisFileData::MisFileData(const char *name, uint8_t animName, uint8_t animFAmt, uint32_t flags,
+			 uint8_t animDelay, std::array<uint8_t, 16> animLen,
+			 std::array<int16_t, 16> animWidth, std::array<int16_t, 16> animWidth2)
+	: MisFileData(name, animName, animFAmt, flags,
+		      duplicate<uint8_t>(animDelay, animFAmt), animLen,
+		      animWidth, animWidth2)
+{
+}
```
AJenbo, line 261: While this helps to simplify the data table it ads ~70 lines of logic and only covers 5/8 of the possible permutations, meaning that some one editing the data table could easily end up with build errors for what might not be obvious to them.

Source/misdat.cpp
```diff
@@ -127,68 +129,162 @@ MissileDataStruct MissileData[] = {
...
+	{ "Firewal",  MFILE_FIREWAL,    2,       0,                       0,          { 13, 11 },                              128,                     32               },
+	{ "MagBlos",  MFILE_MAGBLOS,    1,       0,                       1,            10,                                    128,                     32               },
+	{ "Portal",   MFILE_PORTAL,     2,       0,                    {  0, 1 },       16,                                     96,                     16               },
+	{ "Bluexfr",  MFILE_BLUEXFR,    1,       0,                       0,            19,                                    160,                     48               },
+	{ "Bluexbk",  MFILE_BLUEXBK,    1,       0,                       0,            19,                                    160,                     48               },
+	{ "Manashld", MFILE_MANASHLD,   1,       MFLAG_LOCK_ANIMATION,    0,             1,                                     96,                     16               },
+	{ "Blood",    MFILE_BLOOD,      4,       0,                       0,          { 15,  8,  8,  8 },                    {  96, 128, 128, 128 }, {  16, 32, 32, 32 } },
```
AJenbo, line 143: Blood is unused, since it is the only one that doesn't use the same value for all entries in `animWidth` and `animWidth2` we could blank it and that would allow use to change the type from `std::array<int16_t, 16>` to simply `int16_t` which would achieve much of the simplification done here but with out the need for the additional constructors.

Source/misdat.cpp
```diff
@@ -127,68 +129,162 @@ MissileDataStruct MissileData[] = {
...
-	{ "",         MFILE_NONE,        0, 0, { nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr }, { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }, {  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0 }, {   0,   0,   0,   0,   0,   0,   0,   0,  0,  0,  0,  0,  0,  0,  0,  0 }, {   0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0 } },
+	// name,      animName,        animFAmt, flags,                animDelay[16], animLen[16],                           animWidth[16],          animWidth2[16]
+	{ "Arrows",   MFILE_ARROWS,     1,       MFLAG_LOCK_ANIMATION,    0,            16,                                     96,                     16               },
+	{ "Fireba",   MFILE_FIREBA,    16,       0,                       0,            14,                                     96,                     16               },
+	{ "Guard",    MFILE_GUARD,      3,       0,                       1,          { 15, 14,  3 },                           96,                     16               },
```
AJenbo, line 135: It looks like it would also be possible to split up the unusual parts of `Guard`, `Firewal`, `Portal`, `Holy`, `Acidpud`, and `Sklball` if we modefy `missiles_process_charge()` to be aware of the mapping in cases where `mis->_mimfnum > animFAmt`

```cpp
MissileExtraFileData MissileExtraSpriteData[] = {
{ "Guard2", 3, 0, 1, 14, 96, 16 },
{ "Guard3", 3, 0, 1, 14, 96, 16 },
{ "Firewal2", 2, 0, 0, 11, 128, 32 },
{ "Portal2", 2, 0, 1, 16, 96, 16 },
{ "Holy2", 16, 0, 0, 14, 96, 16 },
{ "Acidpud2", 2, 0, 0, 4, 96, 16 },
{ "Sklball9", 9, 0, 1, 8, 96, 16 },
};
```

Additionally this also would need a special version of `SetMissAnim()` to be used in the place where the 7 special missile animations are loaded using `SetMissDir()`.

But I think this is optional in relation to this PR.

AJenbo (changes requested):
- Null "Blood" to simplify the data table
- monster_flag should not be used for missiles

Source/misdat.cpp
vlolteanu: I've come up with a much simpler solution: extending std::array and adding a constructor that fills the entire array with something.

vlolteanu:
Now that the unusual format of Blood is no longer bugging us, I guess we should leave it there.

Source/misdat.cpp
```diff
@@ -127,68 +129,120 @@ MissileDataStruct MissileData[] = {
...
+		pinnedMem.reserve(1);
+	else
+		pinnedMem.reserve(animFAmt);
+
+	for (int i = animFAmt; i < 16; i++) {
+		animDelay[i] = 0;
+		animLen[i] = 0;
+		animWidth = 0;
+		animWidth2 = 0;
+	}
```
vlolteanu, line 218: Does the game ever touch entries beyond animFAmt? If not, this is not needed.

Source/misdat.cpp
vlolteanu: It takes care of all permutations without requiring extra MisFileData constructors.

Source/missiles.cpp
vlolteanu: You're right.

Source/misdat.cpp
```diff
@@ -127,68 +129,123 @@ MissileDataStruct MissileData[] = {
...
+	if ((flags & MFLAG_HIDDEN) != 0)
+		return;
+
+	char pszName[256];
+	if ((flags & MFLAG_ALLOW_SPECIAL) != 0) {
+		sprintf(pszName, "Missiles\\%s.CL2", name);
+		auto file = LoadFileInMem(pszName);
+		for (int i = 0; i < animFAmt; i++)
+			animData[i] = CelGetFrame(file.get(), i);
+		pinnedMem.push_back(std::move(file));
```
vlolteanu, line 235: It looks like this never happens.

If that's the case, animData can be turned into an array of unique_ptrs.

Source/misdat.cpp
AJenbo: It does not.

Source/utils/autofill_array.h
```diff
@@ -0,0 +1,24 @@
...
+#include <array>
+
+namespace devilution {
+
+template<typename T, std::size_t N>
+struct AutofillArray: public std::array<T, N>
+{
+	using std::array<T, N>::array;
+
+	AutofillArray(T x)
```
glebm, line 12: Single-argument constructors should be marked `explicit` https://google.github.io/styleguide/cppguide.html#Implicit_Conversions

Source/utils/autofill_array.h
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	AutofillArray()
+		: AutofillArray(0)
+	{
+	}
+
+	template<typename... Xs>
+	AutofillArray(const T &x, const Xs &... xs)
+		: std::array<T, N>({ x, (const T &)xs... })
```
glebm, line 24: Avoid C-style casts

Source/utils/autofill_array.h
vlolteanu: Not in this case. The whole point was to make MissileSpriteData legible.

glebm: You can do something like `{15}` in `MissileSpriteData`

Source/missiles.cpp
```diff
@@ -1263,51 +1243,28 @@ void SetMissDir(int mi, int dir)
...
 
 void InitMissileGFX()
 {
-	for (int mi = 0; MissileSpriteData[mi].mAnimFAmt != 0; mi++) {
+	for (int mi = 0; MissileSpriteData[mi].animFAmt != 0; mi++) {
 		if (!gbIsHellfire && mi > MFILE_SCBSEXPD)
 			break;
-		if ((MissileSpriteData[mi].mFlags & MFLAG_HIDDEN) == 0)
-			LoadMissileGFX(mi);
+		LoadMissileGFX(mi);
```
AJenbo, line 1254: dont think there is a need to keep this, just call MissileSpriteData[mi].LoadGFX();

AJenbo:
@vlolteanu do you think you will have time to address the two last comments in the next week or so? :)

vlolteanu:
> @vlolteanu do you think you will have time to address the two last comments in the next week or so? :)

Sure.

vlolteanu:
"Blood" is actually used by Blood Star. I'm not removing it.

AJenbo:
> "Blood" is actually used by Blood Star. I'm not removing it.

That's fine we moved past that, the only remaining issue is this one i believe: https://github.com/diasurgical/devilutionX/pull/2440#discussion_r676575771

I'm guessing you already handled this one, but haven't pushed yet?
https://github.com/diasurgical/devilutionX/pull/2440#discussion_r676485137

vlolteanu:
There's also the enum and the explicit initializer one, which require me to redo the table.

Source/utils/autofill_array.h
vlolteanu: Wasn't needed; the initializer list used the explicit constructor anyway. You get a bogus-sounding error if you try to put braces around it.

vlolteanu:
It's done now.

AJenbo:
@glebm would you review/approve the changes?

AJenbo (dismissed):
Glad we can finally merge this :)

AJenbo (approved):

AJenbo:
Wtf GitHub! This didn't pass CI and I was two commits from having it all resolved :D

PR #2443, opened 2021-07-21
Fix 3DS crash at startup

### LINKED ISSUE #2442: 3DS crashes at startup
Release build on 3DS crashes at startup as of 9ebdb6d. FWIW, I couldn't reproduce the crash with a debug build.

StephenCWills:
I assume this return statement was removed by mistake because most platforms do not define these `JOY_AXIS_*` variables.

This resolves #2442


AJenbo:
yeah, this is probably an auto fix, to avoid that we should change the `break;` in the switch to instead be `return true;` or use a variable to track the return value.

AJenbo (approved):
:+1:

PR #2444, opened 2021-07-22
Fix Counselor always hiding

### LINKED ISSUE #2441: Advocate/Magistrate is scared and hides in subspace
Advocate is scared and hides in subspace similar to #2429.

https://user-images.githubusercontent.com/45902449/126544383-00cddbab-5cb8-411f-bf8b-8b91582fa402.mp4

Unfortunately this time I could not determine the causing commit because the Advocate appears after the game is loaded again. I can only guess that it must be a similar problem.

By the way, there were Mages in level 15 who shoot with charged bolts, and I got the impression that someone of them was hiding there too.

It is also somehow so that it does not happen to all mages and can be experienced irregularly. I hope the information is sufficient to resolve the problem.

Savegame:
[advocate.zip](https://github.com/diasurgical/devilutionX/files/6858184/advocate.zip)

obligaron:
Fixes #2441

Bug was introduced with #2283
Thanks @Chance4us for finding the correct commit that helped a lot 🙂 

The problem was that a [branch was removed](https://github.com/diasurgical/devilutionX/pull/2283/files?diff=split&w=1#diff-d46585089d315b53630b1035442b3d490d7b422cb05ceee76cf9041de3e39179L4178-L4180), the pr restores it.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -3211,6 +3211,9 @@ void CounselorAi(int i)
 				monster._mgoal = MGOAL_NORMAL;
 				StartFadein(monster, md, true);
 			}
+		} else {
+			monster._mgoal = MGOAL_NORMAL;
+			StartFadein(monster, md, true);
 		}
 	} else if (monster._mgoal == MGOAL_NORMAL) {
 		if (abs(mx) >= 2 || abs(my) >= 2) {


AJenbo (approved):

AJenbo:
Thanks for tracking down the offending line :)

Chance4us:
Thanks for fixing

PR #2453, opened 2021-07-25
[Android] Locate external App data path and migrate data files

AJenbo:


android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -1,136 +1,152 @@
...
+					byte[] buf = new byte[1024];
+					int len;
+					while ((len = in.read(buf)) > 0) {
+						out.write(buf, 0, len);
+					}
+
+					return true;
+				}
+			}
+		} catch (IOException ignored) {}
```
StephenCWills, line 86: Is there any reason not to at least use `Log.v()` to log what happened?

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -1,136 +1,152 @@
...
+		if (newPath.exists()) {
+			if (file.canWrite()) {
+				file.delete();
+			}
+			return;
+		}
+		if (!newPath.canWrite()) {
+			return;
+		}
+		if (file.renameTo(newPath)) {
```
StephenCWills, line 105: If the rename succeeded, we shouldn't need to copy, right?

```suggestion
		if (!file.renameTo(newPath)) {

```

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -1,136 +1,152 @@
...
+
+		migrateFile(new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS) + "/diabdat.mpq"));
+		migrateFile(new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS) + "/DIABDAT.MPQ"));
+
+		migrateFile(new File("/sdcard/diabdat.mpq"));
+		migrateFile(new File("/sdcard/devilutionx/diabdat.mpq"));
+		migrateFile(new File("/sdcard/devilutionx/spawn.mpq"));
+
+		for (File internetFile : getFilesDir().listFiles()) {
+			migrateFile(internetFile);
```
StephenCWills, line 130: What does `internetFile` mean? Is this supposed to be `internalFile`?

StephenCWills (changes requested):

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -1,136 +1,152 @@
...
+
+	private void installationGuideShown() {
+		SharedPreferences.Editor editor = mPrefs.edit();
+		editor.putBoolean(shouldShowInstallationGuidePref, false);
+		editor.apply();
+	}
+
+	private void showInstallationGuide() {
+		String message = "To play the full version you must place DIABDAT.MPQ from the original game in Android"
+				+ externalDir.split("/Android")[1]
```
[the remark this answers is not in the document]
AJenbo: The path will usually be something like `/storage/Emulated/0/Android/data/org.diasurgical.devilutionx/files`, but `/storage/Emulated/<int>` or (`/storage/<UUID>`) is never exposed via the UI or when connecting via USB, instead they would be shown as `SD`, `Internal Storage`, `Phone` or `Card` (depending on the tool used). So showing the actual path would be unhelpful to most users, possibly leading them to place the files in `/storage/Emulated/0/storage/Emulated/0/Android/data/org.diasurgical.devilutionx/files` instead of the intended folder.

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -1,136 +1,155 @@
...
+				return;
+			}
+		}
+
+		migrateFile(new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS) + "/diabdat.mpq"));
+		migrateFile(new File(Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS) + "/DIABDAT.MPQ"));
+
+		migrateFile(new File("/sdcard/diabdat.mpq"));
+		migrateFile(new File("/sdcard/devilutionx/diabdat.mpq"));
+		migrateFile(new File("/sdcard/devilutionx/spawn.mpq"));
```
StephenCWills, line 130: Is there a reason why we're copying `DIABDAT.MPQ` from `Downloads`, but not from `/sdcard/devilutionx`?

AJenbo: The previous build where hard coded for `/sdcard/devilutionx/diabdat.mpq`, if the user placed the file as `DIABDAT.MPQ` it would either not have worked or the SD card was formatted as FAT32 (case insensitive).

StephenCWills: I see now. Mine was the latter. It's not copying, though, because `file.canRead()` is returning `false`. I checked permissions, and my only options under "Files and media" are "Allow access to media only" or "Deny".

AJenbo: I have adjusted the SDK version in the gradle file, please try again :)

StephenCWills: That did it.

StephenCWills (approved):

PR #2322, opened 2021-07-07
Decouple missile movement from game logic

obligaron:
## Introduction

After #2230 the monsters got faster.
Now we have to arm up! 
This PR enables us to fight back with faster missiles! 🚀 

The only downside is that the new missile technology got leaked to the monsters... so they also fight back faster. But now we all have nice new missiles. 😋 

## Content

This PR decouples missile movement from game logic.
The logic is different from normal animation distribution logic cause the progress of the movement isn't tide to a animation frame but the velocity of the missile. Also collision logic is only needed for missiles.

## Example Videos at 10% game speed

<details><summary>Firebolt</summary>

### Master

https://user-images.githubusercontent.com/25415264/124798940-40a90e00-df54-11eb-8a67-fa71c5f9815e.mp4

### PR

https://user-images.githubusercontent.com/25415264/124798960-456dc200-df54-11eb-94d9-432c6dc1135b.mp4

</details>

<details><summary>Arrow</summary>

### Master

https://user-images.githubusercontent.com/25415264/124799027-5c141900-df54-11eb-88a2-29c055af400d.mp4

### PR

https://user-images.githubusercontent.com/25415264/124799031-5e767300-df54-11eb-9931-b42eb08f091c.mp4

</details>

<details><summary>Nova</summary>

### Master

https://user-images.githubusercontent.com/25415264/124799181-8fef3e80-df54-11eb-9f22-52f825ca55c5.mp4

### PR

https://user-images.githubusercontent.com/25415264/124799197-94b3f280-df54-11eb-8004-2b6aa289592a.mp4

</details>

<details><summary>Holy Bolt</summary>

### Master

https://user-images.githubusercontent.com/25415264/124799254-a4cbd200-df54-11eb-94e4-e7e68e913cab.mp4

### PR

https://user-images.githubusercontent.com/25415264/124799265-a7c6c280-df54-11eb-86bd-32c9eb4a4079.mp4

</details>

<details><summary>Flamewave</summary>

### Master

https://user-images.githubusercontent.com/25415264/124799546-f5432f80-df54-11eb-8ebf-ec4ad58ac35a.mp4

### PR

https://user-images.githubusercontent.com/25415264/124799532-f1171200-df54-11eb-9299-24b1fdd4caf5.mp4

</details>

<details><summary>Elemental</summary>

### Master

https://user-images.githubusercontent.com/25415264/124799657-16a41b80-df55-11eb-82cb-4820401052af.mp4

### PR

[...]

PR #2230, opened 2021-06-26
Enable ADL for Monster walking and decouple positioning while walking from game logic

obligaron:
Beware! The monsters are getting faster! 👹 

Contributes to #838

<details><summary>Example Video</summary>

### Master

https://user-images.githubusercontent.com/25415264/123510936-8de8ce00-d67e-11eb-8397-fe39538104d1.mp4

### PR

https://user-images.githubusercontent.com/25415264/123510939-950fdc00-d67e-11eb-9854-fbb0638937b8.mp4

</details>

Notes:

- No behavior changes
- Similar to #1939 this introduces a minor breaking change (one animation)

PR #2457, opened 2021-07-26
ADL: Introduce preview Frame (reduce input lag feeling)

obligaron:
## Introduction

PR#2322
The [missiles](https://github.com/diasurgical/devilutionX/pull/2322) and [monsters](https://github.com/diasurgical/devilutionX/pull/2230) got faster, But our mighty hero seems sometimes slow... like his is thinking too much. This is now gone! We got instant reflexes.

## Content

This PR tries to address input lag. This was done in the same way some other (multiplayer) games handle it.
We decouple shown animation in some cases from game logic.
If the player does nothing (stand animation) and we got an input we analyse the input.
We will anticipate what animation the game logic will show for this input.
Then we will show this preview animation until the next game tick (game logic run).

## Examples

<details><summary>Example Video (1% game speed)</summary>

https://user-images.githubusercontent.com/25415264/127044184-3c0349c1-cc43-4654-9945-3d1fd492f53f.mp4

</details>

<details><summary>Example Logs (Time in ms)</summary>

Event | Start-Time | Duration
-- | -- | --
DVL_WM_LBUTTONDOWN | 7342 | 0
PreviewWalk | 7342 | 12
StartWalk | 7354 |  
DVL_WM_LBUTTONDOWN | 16278 | 0
PreviewWalk | 16278 | 24
StartWalk | 16302 |  
DVL_WM_LBUTTONDOWN | 22934 | 0
PreviewWalk | 22934 | 18
StartWalk | 22952 |  
DVL_WM_LBUTTONDOWN | 46950 | 0
PreviewWalk | 46950 | 0
StartWalk | 46950 |  
DVL_WM_LBUTTONDOWN | 52509 | 0
PreviewWalk | 52509 | 42
StartWalk | 52551 |  
DVL_WM_LBUTTONDOWN | 62139 | 0
PreviewWalk | 62139 | 13
StartWalk | 62152 |  
DVL_WM_LBUTTONDOWN | 66839 | 0
PreviewWalk | 66839 | 12
StartWalk | 66851 |  
DVL_WM_LBUTTONDOWN | 67644 | 0
PreviewWalk | 67644 | 6
StartWalk | 67650 |  
DVL_WM_LBUTTONDOWN | 68436 | 0
PreviewWalk | 68436 | 18
StartWalk | 68454 |  
DVL_WM_LBUTTONDOWN | 69149 | 0
PreviewWalk | 69149 | 1
StartWalk | 69150 |  
DVL_WM_LBUTTONDOWN | 69868 | 1
PreviewWalk | 69869 | 30
StartWalk | 69899 |  
DVL_WM_LBUTTONDOWN | 70515 | 0
PreviewWalk | 70515 | 36
StartWalk | 70551 |  
DVL_WM_LBUTTONDOWN | 71149 | 0
PreviewWalk | 71149 | 0
StartWalk | 71149 |  
DVL_WM_LBUTTONDOWN | 71898 | 0
PreviewWalk | 71898 | 6
StartWalk | 71904 |  
DVL_WM_LBUTTONDOWN | 72587 | 0
PreviewWalk | 72587 | 12
StartWalk | 72599 |  

</body>

</html>

</details>

## Notes

- No behavior changes. Only animations.
- The time until your action is displayed on screen is still limited by fps. Cause if we render too slow, we can't show something. 😉 
- We always show the latest action, cause the last action would override/set `destAction` of the player. This can also result in multiple/different animations shown (see example video). But this is unlikely to happen at normal game speed.
- The preview can be wrong. For example if one player attacks another player and the player walked to next tile at the next game tick.
- In preview we always show the first frame of the animation. The other frames should be shown with the normal game logic.
- If the player is currently executing any other task (not standing), we don't show a preview frame, cause this would interrupt the normal animation and the task will be executed _far_ later anyway.
- Handling of input must be done in `NetSendCmdXYZ` cause the messages get only iterated/parsed when the game logic runs ( `multi_process_network_packets` calls `ParseCmd` and this sets players `destAction`).
- This adds complexity to the source code. So it would be great if you could play test it. That would help to decide if it's worth it or not. 🙂

### DIFF
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -75,7 +75,7 @@ float AnimationInfo::GetAnimationProgress() const
 	return animationFraction;
 }
 
-void AnimationInfo::SetNewAnimation(const CelSprite *celSprite, int numberOfFrames, int ticksPerFrame, AnimationDistributionFlags flags /*= AnimationDistributionFlags::None*/, int numSkippedFrames /*= 0*/, int distributeFramesBeforeFrame /*= 0*/)
+void AnimationInfo::SetNewAnimation(const CelSprite *celSprite, int numberOfFrames, int ticksPerFrame, AnimationDistributionFlags flags /*= AnimationDistributionFlags::None*/, int numSkippedFrames /*= 0*/, int distributeFramesBeforeFrame /*= 0*/, float previewShownGameTickFragments /*= 0.F*/)
 {
 	if ((flags & AnimationDistributionFlags::RepeatedAction) == AnimationDistributionFlags::RepeatedAction && distributeFramesBeforeFrame != 0 && NumberOfFrames == numberOfFrames && CurrentFrame >= distributeFramesBeforeFrame && CurrentFrame != NumberOfFrames) {
 		// We showed the same Animation (for example a melee attack) before but truncated the Animation.
@@ -128,6 +128,11 @@ void AnimationInfo::SetNewAnimation(const CelSprite *celSprite, int numberOfFram
 			// This also means Rendering should never hapen with TicksSinceSequenceStarted < 0.
 			TicksSinceSequenceStarted = -1.F;
 		}
+
+		// The preview animation was shown some times (less then one game tick)
+		// So we overall have a longer time the animation is shown
+		TicksSinceSequenceStarted += previewShownGameTickFragments;
+		relevantAnimationTicksWithSkipping += previewShownGameTickFragments;
 
 		if ((flags & AnimationDistributionFlags::SkipsDelayOfLastFrame) == AnimationDistributionFlags::SkipsDelayOfLastFrame) {
 			// The logic for player/monster/... (not ProcessAnimation) only checks the frame not the delay.
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -24,7 +24,7 @@ int AnimationInfo::GetFrameToUseForRendering() const
 	if (CurrentFrame > RelevantFramesForDistributing)
 		return CurrentFrame;
 
-	float ticksSinceSequenceStarted = static_cast<float>(TicksSinceSequenceStarted);
+	float ticksSinceSequenceStarted = TicksSinceSequenceStarted;
 	if (TicksSinceSequenceStarted < 0) {
 		ticksSinceSequenceStarted = 0.0F;
 		Log("GetFrameToUseForRendering: Invalid TicksSinceSequenceStarted {}", TicksSinceSequenceStarted);
@@ -95,7 +95,7 @@ void AnimationInfo::SetNewAnimation(const CelSprite *celSprite, int numberOfFram
 	CurrentFrame = 1 + numSkippedFrames;
 	TickCounterOfCurrentFrame = 0;
 	TicksPerFrame = ticksPerFrame;
-	TicksSinceSequenceStarted = 0;
+	TicksSinceSequenceStarted = 0.F;
 	RelevantFramesForDistributing = 0;
 	TickModifier = 0.0F;
 	IsPetrified = false;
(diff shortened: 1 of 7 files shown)


julealgon:
It's not clear to me exactly what is happening in the video. Can you provide a summary? I was actually expecting 2 videos to compare the behavior there.

obligaron:
> 
> 
> It's not clear to me exactly what is happening in the video. Can you provide a summary? I was actually expecting 2 videos to compare the behavior there.

I didn't attach a comparison video cause you don't see anything on it. 😉 

<details><summary>Example Video Master 1% game speed</summary>

https://user-images.githubusercontent.com/25415264/127059610-e46f2297-64da-4415-9fdc-5a6f97f60a66.mp4

</details>

In both videos (this and the one in the pr description) I do the same. I click on the monster or click on the tiles (give walking commands).
At normal game speed the input lag is max 50ms, that's hard to see. That's why both videos are extremely slow (1% game speed) to capture the input lag.
In the master version you see ... nothing. The player doesn't react to my clicks in any way => input lag.
In the PR version you see that the player reacts to my clicks. It changes direction or change the animation. 🙂

FitzRoyX:
I tested it and couldn't really tell if there was an improvement. The only thing I noticed was that hold clicking the base of the wall makes the player rapidly blink between animations. How does this compare to a higher tickrate of 60? Correct me if I'm wrong, but input can only affect every third frame right now. Since every monitor does at least 60, the max delay right now is 48ms instead of 16ms.

qndel:
Player going crazy when holding mouse near wall is a huge nope for me ;p other than that I don't think I've noticed a difference :D

obligaron:
The wall thing should be fixed now.
Also note that the wall thing is also present in master. Just a little bit different. The animation won't update (you always see the first frame of the walk animation). 😁 

>  How does this compare to a higher tickrate of 60? Correct me if I'm wrong, but input can only affect every third frame right now. Since every monitor does at least 60, the max delay right now is 48ms instead of 16ms.

The game logic runs per default 20 times per second (this can be configured with the `speed` option in the ini).
That means every 50 ms the game logic runs.
Every input between two game logic runs don't result in a different rendering. Cause only the game logic changes what kind of animation (walking, attacking, ...) is shown.
Even classic ADL doesn't help in that regard, cause ADL only distributes already started animations. But ADL (previous to this PR) doesn't start a new animation.
btw: You can also get less then 60 fps, if your device is too slow.

> In practice, D1's slow tile-based movement seems to mask input delay due to being "locked in" when travelling between tiles on the grid. So it's not like the lag is as noticeable as it would be if your movement was totally free.

This PR don't help in that case. It only shows the new animation if you currently idle stand (stand animation is shown). Cause this is the only case where the next game tick can start a new animation and we don't crop an existing animation.

obligaron:
I tried to visualize current master (20 game ticks per second), this pr and high tickrate (60 game ticks per second):

Legend | Description
-- | --
Time | time passed
Action | action done by user
Animation | What animation is currently rendered
Game logic | What is the current game logic state

Time | Action | Master Game Logic | Master Animaton | PR Game Logic | PR Animation | High Tickrate Game Logic | High Tickrate Animation
-- | -- | -- | -- | -- | -- | -- | --
0 ms | game tick (master and high tickrate) | Stand | Stand | Stand | Stand | Stand | Stand
10 ms | clicked | Stand | Stand | Stand | Walking | Stand | Stand
16 ms | game tick (high tickrate) | Stand | Stand | Stand | Walking | Walking (1 of 24) | Walking
33 ms | game tick (high tickrate) | Stand | Stand | Stand | Walking | Walking (2 of 24) | Walking
50 ms | game tick (master and high tickrate) | Walking (1 of 8) | Walking | Walking (1 of 8) | Walking | Walking (3 of 24) | Walking
66 ms | game tick (high tickrate) | Stand | Stand | Stand | Walking | Walking (4 of 24) | Walking
83 ms | game tick (high tickrate) | Stand | Stand | Stand | Walking | Walking (5 of 24) | Walking
100 ms | game tick (master and high tickrate) | Walking (2 of 8) | Walking | Walking (2 of 8) | Walking | Walking (6 of 24) | Walking
… | … | … | … | … | … | … | …

Cause high tickrate (60 game ticks per second) ticks more often the click (the walking command) get processed earlier by the game. That's why high tickrates help to improve input lag.
But even high tickrate is still coupled with the fix time span (around 16,6 ms) between two ticks.
With the preview we don't have to change the game ticks and still got fast animation updates. But it doesn't change what the game logic does. That means that the animation only looks more responsive.

FitzRoyX:
Well, my opinion on this is known. I prefer real gains over tricks and half-measures. If it were me, I'd find a way to support 60 tickrate characters (+ camera jerk fix) in the ini. Or I would at least promise to move to it once vanilla has every improvement it can possibly get.

AJenbo:
> I prefer real gains over tricks and half-measures.

I think we disagree on what the definition of this is :roll_eyes:

julealgon:
> The game logic runs per default 20 times per second (this can be configured with the speed option in the ini).
That means every 50 ms the game logic runs.

What if we make the game update logic faster? Wouldn't that be a more sensible "fix" for the lag?

AJenbo:
60 fps logic would still be delayed on a 120 hz monitor. Fixed rates will never perfectly sync with all monitors on all systems. That's why games do not limit the local player to the logic rate and also deploy interpolation on both position and animations.

FitzRoyX:
If a game has a sub-60 tickrate, it's because they couldn't reach 60fps for it to matter (console) or because they want cheap servers that can't handle the bandwidth of a high tickrate for online play.

It doesn't matter that 60 != 120. The jump from 20 to 60 is far more perceptible. You're talking a 33ms reduction for everyone vs 41ms for people who own 120hz monitors. Doubling the tickrate to reduce 8ms for a small fraction of players? Clearly not worth it.

AJenbo:
> It doesn't matter that 60 != 120. The jump from 20 to 60 is far more perceptible. You're talking a 33ms reduction for everyone vs 41ms for people who own 120hz monitors. Doubling the tickrate to reduce 8ms for a small fraction of players? Clearly not worth it.

I never said anything to the contrary, what I'm saying is that any fixed rate solution is inherently a half-measures. I don't think "clearly not worth it" holds much meaning in the context of Devilution.

If the various event loops are properly separated you can adjust them individually without it breaking the game, they could even be exposed as game settings, but it sounds more like you are advocating for a **60fps**-logo then removing the engine limitations.

AJenbo (approved):

PR #2458, opened 2021-07-26
Fix item swapping and AnimationInfo::ChangeAnimationData

### LINKED ISSUE #2455: Camera shifts with open inventory
We have still one of the last two side effects of enabling ADL.
This has happen since commit 3ba4bac5e284626c701cd37546e7d769fcfa5212 when we fixed "Gold Viper's" death animation, see following link.
https://github.com/diasurgical/devilutionX/pull/2230#issuecomment-869012115

So it happened to us that we fixed one bug and generated a new one.

https://user-images.githubusercontent.com/45902449/126910528-7c485896-577e-477b-9dd4-7f4f40085682.mp4

Savegame:
[shifts.zip](https://github.com/diasurgical/devilutionX/files/6874433/shifts.zip)

All in all, once we've solved this and #2365 it looks very good and the side effects should be gone.

obligaron:
Fixes #2455 

Notes:

- Main problem was that we [changed](https://github.com/diasurgical/devilutionX/pull/2244) from `delayLen` to `ticksPerFrame` but the call in `items.cpp` wasn't adjusted.
- Secondary problem was that we could end up with a currentframe=0 if we swap animations in `ChangeAnimationData`. But this only surfaced cause we passed the wrong parameter (see main problem). Fixed anyway. 😉 
- I also found the issue while working at #2457 😄

### DIFF
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -7,6 +7,7 @@
 #include "animationinfo.h"
 #include "appfat.h"
 #include "nthread.h"
+#include "utils/stdcompat/algorithm.hpp"
 #include "utils/log.hpp"
 
 namespace devilution {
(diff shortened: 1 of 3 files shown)


AJenbo (approved):

Chance4us:
Many thanks.

PR #2468, opened 2021-07-29
Scoped enums for ui_item.h

ephphatha:
Bit of an experiment to see how scoped enums read when used as a collection of flags, happy for any feedback. The operator overloads are mainly to replicate the implicit conversion to/from int that is allowed with unscoped enums.

~Also provides `operator&&` as a test if a flag is set, though this does cause syntax errors if used in a conditional statement without parentheses to force correct precedence. This may be desirable :)~

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -102,7 +102,7 @@ void UiInitList(int count, void (*fnFocus)(int value), void (*fnSelect)(int valu
 #endif
 	textInputActive = false;
 	for (const auto &item : items) {
-		if (item->m_type == UI_EDIT) {
+		if (item->m_type == UiType::UI_EDIT) {
 			auto *pItemUIEdit = static_cast<UiEdit *>(item.get());
 			SDL_SetTextInputRect(&item->m_rect);
 			textInputActive = true;
@@ -808,28 +808,28 @@ void RenderItem(UiItemBase *item)
 	if (item->has_flag(UiFlags::UIS_HIDDEN))
 		return;
 	switch (item->m_type) {
-	case UI_TEXT:
+	case UiType::UI_TEXT:
 		Render(static_cast<UiText *>(item));
 		break;
-	case UI_ART_TEXT:
+	case UiType::UI_ART_TEXT:
 		Render(static_cast<UiArtText *>(item));
 		break;
-	case UI_IMAGE:
+	case UiType::UI_IMAGE:
 		Render(static_cast<UiImage *>(item));
 		break;
-	case UI_ART_TEXT_BUTTON:
+	case UiType::UI_ART_TEXT_BUTTON:
 		Render(static_cast<UiArtTextButton *>(item));
 		break;
-	case UI_BUTTON:
+	case UiType::UI_BUTTON:
 		RenderButton(static_cast<UiButton *>(item));
 		break;
-	case UI_LIST:
+	case UiType::UI_LIST:
 		Render(static_cast<UiList *>(item));
 		break;
-	case UI_SCROLLBAR:
+	case UiType::UI_SCROLLBAR:
 		Render(static_cast<UiScrollBar *>(item));
 		break;
-	case UI_EDIT:
+	case UiType::UI_EDIT:
 		Render(static_cast<UiEdit *>(item));
 		break;
 	}
@@ -914,13 +914,13 @@ bool HandleMouseEvent(const SDL_Event &event, UiItemBase *item)
 	if (item->has_any_flag(UiFlags::UIS_HIDDEN | UiFlags::UIS_DISABLED) || !IsInsideRect(event, item->m_rect))
 		return false;
 	switch (item->m_type) {
-	case UI_ART_TEXT_BUTTON:
+	case UiType::UI_ART_TEXT_BUTTON:
 		return HandleMouseEventArtTextButton(event, static_cast<UiArtTextButton *>(item));
-	case UI_BUTTON:
+	case UiType::UI_BUTTON:
 		return HandleMouseEventButton(event, static_cast<UiButton *>(item));
-	case UI_LIST:
+	case UiType::UI_LIST:
 		return HandleMouseEventList(event, static_cast<UiList *>(item));
-	case UI_SCROLLBAR:
+	case UiType::UI_SCROLLBAR:
 		return HandleMouseEventScrollBar(event, static_cast<UiScrollBar *>(item));
 	default:
 		return false;
@@ -970,7 +970,7 @@ bool UiItemMouseEvents(SDL_Event *event, const std::vector<UiItemBase *> &items)
 	if (event->type == SDL_MOUSEBUTTONUP && event->button.button == SDL_BUTTON_LEFT) {
 		scrollBarState.downArrowPressed = scrollBarState.upArrowPressed = false;
 		for (const auto &item : items) {
-			if (item->m_type == UI_BUTTON)
+			if (item->m_type == UiType::UI_BUTTON)
 				HandleGlobalMouseUpButton(static_cast<UiButton *>(item));
 		}
 	}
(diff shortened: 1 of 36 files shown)


ephphatha:
Apparently linking issues implies that this closes the issue, not just that it relates 😂

Source/DiabloUI/ui_item.h
```diff
@@ -42,36 +43,60 @@ enum UiFlags : uint16_t {
 	// clang-format on
 };
 
+inline UiFlags operator|(UiFlags lhs, UiFlags rhs)
```
obligaron, line 46: @AJenbo had a [comment](https://github.com/diasurgical/devilutionX/pull/1869#discussion_r626852934) for this kind of bitflag operators.
Perhaps this refactoring PR is a good point to address the comment.

ephphatha: Ooh, thanks for pointing that out. (I think I actually read that comment trail and then remembered nothing from it when I was looking at this 😓 )

I was looking at C++20 constraints, which would provide for something like the following (if we could even target that version of the language):
```cpp
template<typename T> requires std::is_enum_v<T>
inline T operator|(T lhs, T rhs)
{
	using V = std::underlying_type_t<T>;
	return static_cast<T>(static_cast<V>(lhs) | static_cast<V>(rhs));
}
```
Even if we could use C++20, C++ enums don't have any distinction between flags types and non-flags types like you get in other languages so that would allow for some nonsensical operations like `Direction::DIR_S | Direction::DIR_N`.

The grisumbras/enum-flags library you found in that comment looks ideal, that wrapper class provides actual friend functions and a way to mark certain enums as being used for flags.

Source/DiabloUI/diabloui.cpp
```diff
@@ -734,7 +734,7 @@ void Render(const UiArtText *uiArtText)
 void Render(const UiImage *uiImage)
 {
 	int x = uiImage->m_rect.x;
-	if ((uiImage->m_iFlags & UIS_CENTER) != 0 && uiImage->m_art != nullptr) {
+	if ((uiImage->m_iFlags && UiFlags::UIS_CENTER) && uiImage->m_art != nullptr) {
```
qndel, line 737: u sure u meant `&&`?

AJenbo: Yeah this seams highly misleading, I'm not in favor of using this overload.

julealgon: For reference, C# defines a `.HasFlag(flag)` method on enums (seems like an intuitive API to me).
https://docs.microsoft.com/en-us/dotnet/api/system.enum.hasflag?view=net-5.0

julealgon:
I think adding all the operator overloading is mostly considered a bad practice for `enum class`, since the intention is that you encourage people to _not_ rely on the underlying value of the enum.

julealgon:
Another point is that, IMHO, we should move away from flag enums and use `bitset` instead.

Source/DiabloUI/diabloui.cpp
AJenbo: `HasAnyOf(...)`

ephphatha: C++ enums aren't classes (even though scoped enums use the class keyword...) and can't have member functions, which makes it a little awkward to have functions like the C# `HasFlag` equivalent. I've added a commit replacing `operator&&` with `hasAnyOf(UiFlag, UiFlag)` as the closest approximation without a wrapper class.

I'm not sure `std::bitset` helps in this regard, testing for multiple flags requires longer syntax. I'll give it a try this evening and see how it looks.

Source/DiabloUI/ui_item.h
```diff
@@ -42,36 +43,60 @@ enum UiFlags : uint16_t {
...
+	return lhs;
+}
+
+inline UiFlags operator~(UiFlags value)
+{
+	using T = std::underlying_type_t<UiFlags>;
+	return static_cast<UiFlags>(~static_cast<T>(value));
+}
+
+inline bool hasAnyOf(UiFlags lhs, UiFlags test)
```
AJenbo, line 76: Not sure why you decided to do this with a lower case since the style guide says functions start with an upper case :)

Source/DiabloUI/ui_item.h
```diff
@@ -22,8 +22,9 @@ enum UiType : uint8_t {
 	UI_EDIT,
 };
 
-enum UiFlags : uint16_t {
+enum class UiFlags {
 	// clang-format off
+	NONE            = 0,
 	UIS_SMALL       = 1 << 0,
```
AJenbo, line 28: Since all the line with usages are changed with this PR it would be nice if you could also adjust the naming of the values at the same time :)

```suggestion
	None        = 0,
	Small       = 1 << 0,
```
(`FontSmall` is also an option)

AJenbo (commented):
Looks pretty good, but I'm not a fan of the flag functions being specific to UiFlags and not generally covers all enum based flags, but that might not be possible :/

Source/DiabloUI/ui_item.h
ephphatha: I blame writing code while eating breakfast :D. I'll update that now.

Source/DiabloUI/ui_item.h
ephphatha: Done. Used prefixes to show related flags, the Kerning flag is one I'm not happy with the name of but couldn't think of anything more suitable.

AJenbo: You just made this PR very enticing :)

ephphatha:
> Looks pretty good, but I'm not a fan of the flag functions being specific to UiFlags and not generally covers all enum based flags, but that might not be possible :/

Not really possible without using C++20 constraints or a helper. The grisumbras/enum-flags lib @obligaron pointed out looks like a decent solution. I haven't had time to play with it yet but I'm aiming to give it a shot after testing `std::bitflags`.

Attempting to use the template function from https://github.com/diasurgical/devilutionX/pull/2468#discussion_r679130374 without C++20 constraints leads to undefined behaviour as the function can be applied to non-enum types. Even with the use of constraints the best that can be done using C++ language features is that we allow the function to be applied to any enum type, not just enums that are intended for use as flags.

AJenbo (approved):

ephphatha:
Actually, I just tested `std::enable_if_t` and it looks like that can be used to replicate C++20 constraints (with a more awkward syntax).

```cpp
template <typename ET, std::enable_if_t<std::is_enum_v<ET>, bool> = true>
constexpr ET operator|(ET lhs, ET rhs)
{
	using T = std::underlying_type_t<ET>;
	return static_cast<ET>(static_cast<T>(lhs) | static_cast<T>(rhs));
}

inline UiFlags operator|=(UiFlags &lhs, UiFlags rhs)
{
	lhs = lhs | rhs;
	return lhs;
}
```

However this causes the operator| overload to be used for every enum | enum statement, even while declaring an enum... e.g. https://github.com/diasurgical/devilutionX/blob/8c188634c4a82d9b12bcd1be299989924f680842/Source/items.h#L160

So limiting the scope becomes an issue. Could still do it as part of this PR if that's desirable.

AJenbo:
It would be interesting to see and i prefer thing not changing several times as that can lead to confusion regarding what is the preferred one.

ephphatha:
Had a bit of an experiment this evening that ended up taking long enough that I only got a couple of implementations working.

[Adding template specialisation to the overloads fits in reasonably well with the current usage](https://github.com/diasurgical/devilutionX/commit/8818198b648750058469fd0c3f4b5bdbd659a8ef). The macro `use_enum_as_flags` is a convenience for declaring the appropriate template specialisation (this is similar to the ALLOW_FLAGS_FOR_ENUM macro in enum-flags).

[enum-flags annoyingly takes a bit of work due to design decisions made by the author](https://github.com/diasurgical/devilutionX/commit/9c1752486f7d641d6099c4245fb2d11ebc899023). The helpers from https://github.com/grisumbras/enum-flags/issues/31 are needed so the macro can be used inside a namespace. The big issue with that library is it intentionally avoids creating values of the enum type that aren't represented by a single enum value, forcing the use of the wrapper any time you wish to store a combination of flags (or extracting the underlying value and casting to the enum type). This requires a whole bunch more boilerplate for cases like the Ui* classes to the point I haven't bothered with a fully compiling example.

I haven't had a chance to investigate https://github.com/mrts/flag-set-cpp/blob/master/include/flag_set.hpp yet. It looks about as awkward as enum-flags so maybe it'll be largely the same if I get that working.

AJenbo:
Sounds like we should start by merging this pr as is 🙂

PR #2471, opened 2021-07-30
Remove unused LastMouseButtonTime

obligaron:
... and gone 😄


AJenbo:
Yay less use of SDL_GetTicks() :tada:

PR #2472, opened 2021-07-30
Enable VS installation on Windows with vcpkg.

ropufu:
Removed hard dependency on cigwin and specific versions of referenced libraries to enable "install" builds on Windows via Visual Studio.

**Update:** I screwed up the wording. By cygwin I meant mingw.


StephenCWills:
FYI, there is no dependency on Cygwin. The build instructions include a section on compiling for Windows via MinGW, and it uses Linux to run the build. It looks like CI is actually cross-compiling on Ubuntu for Windows so we'll need to get it working for both VS and MinGW before this change can be merged.

There are a handful of places in `CMakeLists.txt` where we're explicitly checking if the compiler is MSVC. You can use that to branch the `CPACK` logic.
https://github.com/diasurgical/devilutionX/blob/2baca9a5ce3798e0ca3cc72fb6b23ccccee9d317/CMakeLists.txt#L591

ropufu:
> there is no dependency on Cygwin

Not sure I follow you. The Windows-specific code in CPACK block (lines 958--967 and 995--1004) has **i686-w64-mingw32** in library paths; moreover those lines assume that SDL2 and SLD2_ttf live specifically in PROJECT_SOURCE_DIR, which they do not.

This PR was meant to enable "Install devilutionX" option in Visual Studio on Windows (not Linux, not WSL) to succeed. This "Install" option calls CPACK. As it currently is in the master branch, it---naturally---fails. With this PR, it succeeds (at least on my machine). And I tried to keep it as minimal as possible.

Anyways, the decision is yours.

AJenbo:
> Not sure I follow you. The Windows-specific code in CPACK block (lines 958--967 and 995--1004) has **i686-w64-mingw32** in library paths;

mingw32 != Cygwin

ropufu:
A-haha, I'm a dummy. Sorry. That's what I meant all along. Not cygwin. Mingw. Apologies.

AJenbo:
> This PR was meant to enable "Install devilutionX" option in Visual Studio on Windows (not Linux, not WSL) to succeed.

Sure, but you broke support for building the project on Linux so I won't be able to do any more releases for Windows if we merge this (I do not have Windows my self).

The CI tests the cross compiling so if you can get the PR to pass then it should work for both of us. Since I do not have Windows (which is required for VC) there is no way for me to do the opposite.

ropufu:
> You can see the build error by clicking on "Details" next to the failing build.

Thank you for that. Btw, correct me if I am wrong, I think Linux builds are just fine. It was the two Windows builds that failed. I'll see if I can address those. Again, thank you for the "Details" link explanation!

AJenbo:
> I think Linux builds are just fine. It was the two Windows builds that failed.

What's failing are the builds for Windows targets, done from Linux systems ;)

We only have one Windows based build server (they are a bit hard to come by), it's the one called "continuous-integration/appveyor/pr — AppVeyor build succeeded", that one is doing fine with the changes you did here :)

Trihedraf:
What needs to be done is to have a 3rd if for WIN32 and MSVC and add !MSVC to the other two Windows ifs and leave them as they are for the MinGW builds to work properly.

ropufu:
Thanks everybody, I know what I have to do, and will keep working to make a better PR :^)

ropufu:
> for the MinGW builds to work properly

I would like to try to make it work for all builds, not just MinGW. I understand that the current CMake works with Linux; but it does not when built from under Windows. This was my goal: to make it work uniformly across all host machines and targets. I failed, hehe. But I still want to get it to look nicer than it is now.

Trihedraf:
> > for the MinGW builds to work properly
> 
> 
> 
> I would like to try to make it work for all builds, not just MinGW. I understand that the current CMake works with Linux; but it does not when built from under Windows. This was my goal: to make it work uniformly across all host machines and targets. I failed, hehe. But I still want to get it to look nicer than it is now.

Right. You need a third IF for the Windows CPacks. Leave the two that are as they are so the mingw will continue to work and make a third one that is for MSVC so that the VS will work.

ropufu:
Ah. I see. That would be a fast fix. Thank you!! I'll try to play around with Linux hosts though. Reducing the amount of code rather than adding to it just feels so much nicer \*^_^\*

AJenbo:
P.s. you don't need to close your PR, you can simply push the new code to the same branch and the PR will get updated with the changes :) But it's also up to you how you prefer to do it.

ropufu:
Cool beans!

julealgon:
> P.s. you don't need to close your PR, you can simply push the new code to the same branch and the PR will get updated with the changes :) But it's also up to you how you prefer to do it.

@ropufu you can also mark your PR as draft while you are still figuring things out and testing the server build. This can help with notifications and to make it clear that your PR is not yet ready for review.

Once you get it passing, you can then convert from 'draft' to 'ready for review'.

ropufu:
Folks, could you please take a look at this version and criticize it? On the off-chance that it is actually okay: I dunno git or how to do squashes etc., so should I create a new PR so as to avoid my two commits and only have one with a better description?

CMakeLists.txt
```diff
@@ -940,44 +940,18 @@ if(NINTENDO_3DS)
...
-    install(FILES "${PROJECT_SOURCE_DIR}/SDL2-2.0.14/README-SDL.txt"
-      DESTINATION "LICENSE"
-    )
+  if(WIN32)
+    if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+      set(SDL2_WIN32_DLLS_DIR "${CMAKE_BINARY_DIR}")
+      set(SDL2_WIN32_LICENSES_DIR "${PROJECT_SOURCE_DIR}/Packaging/resources")
+    else()
+      set(SDL2_WIN32_DLLS_DIR "${SDL2_EXEC_PREFIX}/bin")
+      set(SDL2_WIN32_LICENSES_DIR "${SDL2_EXEC_PREFIX}/bin")
```
StephenCWills, line 949: What is `SDL_EXEC_PREFIX`? Are you sure that variable exists?

CMakeLists.txt
```diff
@@ -940,44 +940,18 @@ if(NINTENDO_3DS)
...
+    endif()
+
+    file(GLOB _SDL2_ALL_DLLS "${SDL2_EXEC_PREFIX}/bin/*.dll")
+    file(GLOB _SDL2_ALL_LICENSES "${SDL2_EXEC_PREFIX}/bin/LICENSE*.txt")
```
StephenCWills, line 953: I don't think this works for MinGW since the dlls and license files don't all come from the same directory. Anyway, here's a suggestion with variable names to match the ones you use above and below.

```suggestion
    file(GLOB SDL2_WIN32_ALL_DLLS "${SDL2_WIN32_DLLS_DIR}/*.dll")
    file(GLOB SDL2_WIN32_ALL_LICENSES "${SDL2_WIN32_LICENSES_DIR}/LICENSE*.txt")
```

StephenCWills (changes requested):
I downloaded the artifact from the Windows x64 CI build, and it doesn't have all the dlls and license files. I've provided some suggestions for the more obvious mistakes I caught, but you'll need to do even more to get it working properly for the MinGW build.

AJenbo:
> I dunno git or how to do squashes etc.

No worries, it's one click of a button for me when merging

CMakeLists.txt
ropufu: Yes. I checked it on WSL (Windows Subsystem for Linux) Ubuntu to figure out where the SDL libraries live. I can add logging to CMake so that it shows exactly which libraries it fishes out from that path (if I recall correctly, it's SDL2, SDL2_ttf, zlib1, and maybe one other, not sure).

CMakeLists.txt
ropufu: It happens that on WSL (see my comment above) they do: that directory seems to hold a "redistributable" version of SDL dll's along with licensing information. I used that directory to have the current licenses (rather than possibly outdated ones for older SDL2 builds). The ones I added to "Packaging/resources" are a specific fix for MSVC builds, since locating the licenses (at least for vcpkg) seemed to be quite a challenge.

In short---in my opinion keeping the same SDL2_EXEC_PREFIX would make a little more sense on Linux builds.

ropufu:
By the way, if the current version looks okay to everybody, before I un-draft it, could somebody please check if the resulting Linux build actually runs on Windows? Unfortunately (my bet is I mis-configured something... or it could be that my WSL is not a full-blown Ubuntu) I could not get it to complete the compilation on Linux with Windows as target so as to test it. My MSVC builds work and run just fine (and I did check CMake variables, Linux paths, etc).

AJenbo:
just grab the artifact from the build server https://github.com/diasurgical/devilutionX/suites/3379334226/artifacts/79643447

CMakeLists.txt
StephenCWills: Sorry, but your insistence that it does work does not change the reality that it didn't work on CI. Furthermore, if you don't fix the variables based on my suggestion, then it won't work properly for VS either.

CMakeLists.txt
StephenCWills: Logging is not necessary. Just check the CI artifact. Compared to the one on master, it's missing all the dlls and license files that this PR is targeting.

CMakeLists.txt
ropufu: Wow, I seem to have screwed up again, sorry. This time it's the naming. I'll update the PR shortly.

ropufu: Just a little explanation: I have two copies of devilutionX. Copy-pasting and a last-minute deciision to make names more consistent with other CMake variables was a mistake. Windows artifacts now look okay ^(. .)^

ropufu:
> 
> 
> just grab the artifact from the build server https://github.com/diasurgical/devilutionX/suites/3379334226/artifacts/79643447
> 
> what build error do you get under WSL?

ropufu:
I think Github made a wrong design/UI decision to have the red "Close with comment" button right by the green "Comment" one. I always think it's "Cancel and discard message" T_T

CMakeLists.txt
ropufu: > Sorry, but your insistence that it does work does not change the reality that it didn't work on CI

What does CI stand for?

StephenCWills: Continuous Integration

ropufu:
> what build error do you get under WSL?

Right after
`[ 54%] Building CXX object CMakeFiles/libdevilutionx.dir/Source/dx.cpp.obj`
I get a recursive include in `build/_deps/libzt-src/include/mingw-fixes/synchapi.h` which contains one line: `#include <synchapi.h>`. After that `build/_deps/mingw-std-threads-src/mingw.mutex.h` naturally fails to compile.

To setup the build process I used commands from the logs of PR checks. Specifically,
```sudo apt update && sudo apt install -y cmake gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 pkg-config-mingw-w64-x86-64 wget git sudo
sudo Packaging/windows/mingw-prep.sh
cmake -S. -Bbuild -DNIGHTLY_BUILD=ON -DCMAKE_TOOLCHAIN_FILE=../CMake/mingwcc.cmake
cmake --build build -j $(nproc) --target package
```

CMakeLists.txt
AJenbo: > Just a little explanation: I have two copies of devilutionX. Copy-pasting and a last-minute deciision to make names more consistent with other CMake variables was a mistake. Windows artifacts now look okay ^(. .)^

Happenes a lot to me as well, that's what the CI is for 🙂

CMakeLists.txt
ropufu: Both windows artifacts now look (and run) fine.

AJenbo:
Odd, I suppose you could disable zt intergration (option at top of cmake file) and see if that let's you build.

CMakeLists.txt
ropufu: I just clicked the "Resolve conversation" button in the branch above, and now it is gone. Have I done something horrible?

StephenCWills: > I just clicked the "Resolve conversation" button in the branch above, and now it is gone. Have I done something horrible?

No, it's just been collapsed. You can expand it again to read the resolved conversation by clicking `Show resolved` on the far right.

ropufu:
> Odd, I suppose you could disable zt intergration (option at top of cmake file) and see if that let's you build.

That's too complicated for @ropufu. As long as it works in your CI I'm happy :^)

StephenCWills (commented):
This is looking pretty good to me now. From what I can tell, this PR adds `libsodium-23.dll` and `LICENSE/LICENSE.zlib.txt`, and it removes `LICENSE/README-SDL.txt`. `README-SDL.txt` was originally coming from `${PROJECT_SOURCE_DIR}/SDL2-2.0.14/` so that should still be viable. You could also potentially modify the `mingw-prep.sh` script to copy that file out to the `${SDL2_EXEC_PREFIX}/bin` directory if you want them all to be in the same place.

For VS, I guess you'd need to put it into the `Packaging` folder with the rest. As you said, I don't see any straightforward way to get license files from vcpkg.

ropufu:
Right ho.

-- I thought that having a `README-SDL` file in `LICENSES` folder was a little strange, hehe. I'll get it back in there. Not through `${PROJECT_SOURCE_DIR}/SDL2-2.0.13` though. That was one of the big points of this PR: to get rid of explicit library versions.
-- As far as having an extra dll in the output, I'd say it does not hurt much. But I could add an extra clause to make sure it is *not* copied---if you think it is worth the extra line or two of code.

Finally, one question and one request. Question: now that the libraries are pulled from their original location rather than `${PROJECT_SOURCE_DIR}`, should I get rid of the copying process altogether (I guess it is in `prep` scripts)? Request: would you mark your old change request as completed, and mark your post above as a new change request?

Thank you!

StephenCWills (dismissed):

StephenCWills (changes requested):
I can't change my previous comment into a change request, but here's a modified quote.

> This PR removes `LICENSE/README-SDL.txt`, which was originally coming from `${PROJECT_SOURCE_DIR}/SDL2-2.0.14/`. You could potentially modify the mingw-prep.sh script to copy that file out to the ${SDL2_EXEC_PREFIX}/bin directory if you want them all to be in the same place.
> 
> For VS, I guess you'd need to put it into the Packaging folder with the rest. As you said, I don't see any straightforward way to get license files from vcpkg.

FYI, `README-SDL.txt` is a pretty small file with a plea to include it when distributing the SDL runtime. It also contains a link to the zlib license, which is perhaps why we didn't bother including that license file in the build script originally. You can see its contents here.
https://github.com/libsdl-org/SDL/blob/main/README-SDL.txt

As for excluding libsodium, I assume if we don't need it then it must be statically linked. Personally, I don't particularly care if it's there or not, and I'm sure most (all?) Windows users don't either. I suppose I'll defer to @AJenbo on this one.

ropufu:
Awesome. Then one more question: would it make sense to you to have `SDL2_WIN32_LICENSES_DIR` set to `"${PROJECT_SOURCE_DIR}/Packaging/resources"` for both MSVC and Linux hosts? We need them there for MSVC builds anyways...

StephenCWills:
> Awesome. Then one more question: would it make sense to you to have `SDL2_WIN32_LICENSES_DIR` set to `"${PROJECT_SOURCE_DIR}/Packaging/resources"` for both MSVC and Linux hosts? We need the licenses/readmes there for MSVC builds anyways...

Yeah, that's fine. The only problem I foresee is one of versioning. If those licenses ever change with a specific version of the library, the license file we distribute would be coming from our repo whereas the license associated with the assembly that actually gets distributed could be the modified one. However, in practice, it just means we need to keep the licenses in `Packaging/resources` in sync with the versions we download via `mingw-prep.sh` because the CI builds are the ones we distribute.

> Question: now that the libraries are pulled from their original location rather than ${PROJECT_SOURCE_DIR}, should I get rid of the copying process altogether (I guess it is in prep scripts)?

I just noticed I failed to answer this. After running my own build on Linux, I've come to realize that the "original location" you're referring to is actually the directory that the `mingw-prep.sh` script is copying to. You won't be able to obsolete those prep scripts quite so easily.

StephenCWills:
> Right after
> `[ 54%] Building CXX object CMakeFiles/libdevilutionx.dir/Source/dx.cpp.obj`
> I get a recursive include in `build/_deps/libzt-src/include/mingw-fixes/synchapi.h` which contains one line: `#include <synchapi.h>`. After that `build/_deps/mingw-std-threads-src/mingw.mutex.h` naturally fails to compile.
> 
> To setup the build process I used commands from the logs of PR checks. Specifically,
> 
> ```
> sudo Packaging/windows/mingw-prep.sh
> cmake -S. -Bbuild -DNIGHTLY_BUILD=ON -DCMAKE_TOOLCHAIN_FILE=../CMake/mingwcc.cmake
> cmake --build build -j $(nproc) --target package
> ```

Looks like that `mingw-fixes` directory was added to the diasurgical fork to handle case-sensitive filesystem issues. The recursive include occurs when the filesystem you use to run the build is not case-sensitive.

StephenCWills (dismissed):

ropufu:
@StephenCWills Thank you so much for your patience. I am sure I enjoyed this way more than you did :^)

ropufu:
In the meantime, what's with the pending **testflow** required check?

AJenbo:
> In the meantime, what's with the pending **testflow** required check?

I think you did not permit Circle CI to access your fork so it's unable to run the tests on your changes.

ropufu:
> you did not permit Circle CI to access your fork

Ugh. Sorry, how do I do that?

StephenCWills:
I think CircleCI just didn't get the memo about 88e3c89. I'm not seeing any pipeline associated with that commit. Maybe the API call to run the pipeline failed due to a transient issue.

ropufu:
If I updated the PR with some simple change, would that fix it?

StephenCWills:
> If I updated the PR with some simple change, would that fix it?

I can't say for sure, but probably yes.

StephenCWills:
It's still not working. I guess AJenbo is probably right about the permission, then.

> https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-
> 
> If you revoke OAuth access to the CircleCI app on your VCS provider it will disable the keys that we used to authorize your account. We do not receive notifications when the auth is revoked, and may still attempt to use the now revoked auth. For instance, you may encounter this when trying to submit a PR to an open source project that uses CircleCI and it does not trigger. If you encounter this issue submit a support ticket and we can resolve it for you.

Since it's talking about OAuth, you can try checking `Applications > Authorized OAuth Apps` in your GitHub personal account settings. You can also try simply logging into CircleCI using your GitHub account credentials.

StephenCWills (approved):

Trihedraf (approved):

ropufu:
I just added CircleCI to the Authorized OAuth Apps. Dunno what to do next.

P.S. I don't like the whole idea. That in order to do a PR you have to grant some app access to your account... that's strange.

AJenbo:
I would prefer if this could be made to fetch the licenses from the packages rather then having them statically in our repo. The one already there is because we also hade the content in the repo.

StephenCWills:
> I would prefer if this could be made to fetch the licenses from the packages rather then having them statically in our repo. The one already there is because we also hade the content in the repo.

I don't think we can get them from vcpkg.

```
~/projects/vcpkg$ find . -name README-SDL.txt
./buildtrees/sdl1/src/ase-1.2.15-086d8ac6c2.clean/README-SDL.txt
./buildtrees/sdl2/src/SDL2-2-62ee9e680a.clean/README-SDL.txt

~/projects/vcpkg$ find . -name LICENSE*.txt
./buildtrees/libpng/src/v1.6.37-05658d6f1a.clean/contrib/pngminus/LICENSE.txt
./buildtrees/sdl2/src/SDL2-2-62ee9e680a.clean/src/hidapi/LICENSE.txt
./downloads/tools/git-2.26.2-1-windows/LICENSE.txt
./downloads/tools/git-2.26.2-1-windows/mingw32/doc/git-credential-manager/LICENSE.txt
./downloads/tools/powershell-core-7.1.0-windows/LICENSE.txt
./LICENSE.txt
./ports/ffnvcodec/LICENSE.txt
./ports/hungarian/LICENSE.txt
./ports/irrlicht/LICENSE.txt
./ports/tensorflow-common/LICENSE.txt
./ports/vulkan/LICENSE.txt
```

AJenbo:
I suppose that is an issue with the package that VCPKG should address.

StephenCWills:
@AJenbo Maybe we could use these files instead?

```
${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/sdl2/copyright
${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/sdl2-ttf/copyright
${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/freetype/copyright
${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/libsodium/copyright
${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/fmt/copyright
${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/share/zlib/copyright
```

See: [licenses.zip](https://github.com/diasurgical/devilutionX/files/6921410/licenses.zip)

ropufu:
@StephenCWills I like the idea. Although it might be a dash more complicated since the [installation guide](https://github.com/diasurgical/devilutionX/blob/master/docs/building.md) allows for manual `SDL2`, `SDL2_mixer`, `SDL2_ttf` and `Libsodium` installation (without **vcpkg**).

AJenbo:
Yeah VC install doesn't need the licenses as it's not a form of redist

ropufu:
> Yeah VC install doesn't need the licenses as it's not a form of redist

But it could be. I mean, if we have it working, why not have it as an option? Why treat it as something different after we've come this far?

In which case I'd go with: (i) removing the manual libs option in the **building from source** guide; (ii) reverting `SDL2_WIN32_LICENSES_DIR` on Linux to `${SDL2_EXEC_PREFIX}/bin`; (iii) adding **copyright** files transfer `MSVC` clause to CMakeLists' **CPACK** as per @StephenCWills's suggestion. Does this sound reasonable?

AJenbo:
Because it's installing it, not generating an installer for distribution

AJenbo:
zip = we send it to other people, we are responsible for informing people of copyright

msvc install = you install it on your personal computer for your own usage, using the libs that you installed via vcpk = vcpk is responsible for informing you of copyright etc

ropufu:
Okay... One possible issue I can see here. Related to @StephenCWills's comment:

> Speaking of which, by fixing the Install option in Visual Studio, we should be able to add it to our AppVeyor process so we can continue to test it with each build going forward.

If you do that, then you'll have to ensure that all future developers of **devilutionX** will *not* use MSVC builds for redist's. Which seems a lot to ask. A lot more, in my opinion, than pulling copyright files from **vcpkg** in the corresponding build process. Am I missing something here? Sorry for focusing on possibly irrelevant details---I just want to get things straight in my head.

AJenbo:
Does MSVC generate a package that some one could redistribute (msi, zip, nsis etc)? As I understand it it installs the application directly on the current system?

ropufu:
I am afraid I do not know, although I assumed (not basis, pure speculation) yes. By default it does not pack it up, but I'm sure it's a pretty straightforward task to create a zip from the installation directory. I'll check to see if I can run the output on a different Windows machine. I have a couple lying around collecting dust.

AJenbo:
If it does not pack it up then I do not consider it to be building something intended for redistribution, most things (that do not employ DRM) can be copy pasted, despite the intent.

For generating an installer with MSVC you would needed to either setup the WIX CPack generator (only available on Windows, hens why we haven't employed it), or MSVC's "Installer Project" extension.

StephenCWills:
> Does MSVC generate a package that some one could redistribute (msi, zip, nsis etc)? As I understand it it installs the application directly on the current system?

It looks to me like Visual Studio is essentially just running `cmake --build build --target install`. I was able to run `cmake --build build --target package` to produce a zip file, same as the MinGW build. Furthermore, I tried running cmake the same way that AppVeyor does to produce a Visual Studio solution, and it actually produces a `PACKAGE.vcxproj` that serves the same purpose.

StephenCWills:
@AJenbo @ropufu 
Almost got it. The dlls aren't there.
https://ci.appveyor.com/project/StephenCWills/devilutionx/builds/40240507/artifacts

ropufu:
Wow, awesome! @StephenCWills, how tightly tied are you to the current master branch? I would like to play around with the CMakeLists.txt a bit to log variables on the Windows AppVeyor thing and see what is happening behind the CPACK scenes in more detail. Is creating a PR (draft) a reasonable way to achieve it?

AJenbo:
Appveyor is just a service that let's us run tests on windows, if you are running windows yourself it should be a lot faster /simpler to do so locally

StephenCWills:
> how tightly tied are you to the current master branch?

I made a branch based on the latest commit in master at the time so it should be close. I only added one commit to tweak the AppVeyor config.

https://github.com/StephenCWills/devilutionX/tree/app-veyor

> I would like to play around with the CMakeLists.txt a bit to log variables on the Windows AppVeyor thing and see what is happening behind the CPACK scenes in more detail. Is creating a PR (draft) a reasonable way to achieve it?

As AJenbo said, it was significantly faster to run the build locally. I recommend reviewing the AppVeyor config and/or build log to figure out how it works and do the same thing locally for testing. FYI, the Developer Command Prompt for VS2019 adds CMake to your PATH.

PR #2473, opened 2021-07-30
Timedemo: Regard ProcessInput and gfProgressToNextGameTick

obligaron:
## Example Replay

https://user-images.githubusercontent.com/25415264/127667604-53ef7f87-6d82-475a-a77a-e10b0fd9c37c.mp4

</details>

### `ProcessInput` was skipped in replay

Previous the replay only calculated game ticks and never called the [only rendering part](https://github.com/diasurgical/devilutionX/compare/timedemo...obligaron:timedemo?expand=1#diff-84e63ba430a13546a287665b7640118b2f9d66fb9586056bfbab8215bab319c0L842-L847) (game tick is not due). Cause `nthread_has_500ms_passed` returned always true.
That skipped the additional [call](https://github.com/diasurgical/devilutionX/compare/timedemo...obligaron:timedemo?expand=1#diff-84e63ba430a13546a287665b7640118b2f9d66fb9586056bfbab8215bab319c0L843) to `ProcessInput`.
But that changes the replay.
Cause `ProcessInput` sends new _network_ message (for example in `RepeatMouseAction` we have `NetSendCmdParam1(true, rangedAttack ? CMD_RATTACKID : CMD_ATTACKID, pcursmonst);`).

The Problem is that this _network_ messages are added to an internal `message_queue`. And this queue is then parsed in `multi_process_network_packets` and sets for example new `player.destAction`.

But if we skip the additional `ProcessInput` it will be called [too late](https://github.com/diasurgical/devilutionX/blob/2baca9a5ce3798e0ca3cc72fb6b23ccccee9d317/Source/diablo.cpp#L1212) and the game logic doesn't see the new state that otherwise `multi_process_network_packets` would set.

#### Alternative Solutions

- Reorder `ProcessInput` so it's always called before `multi_process_network_packets`.
- Disable additional rendering (and with that `ProcessInput` calls) while recording

### `gfProgressToNextGameTick` needed for game coordinate calculation

ADL and `gfProgressToNextGameTick` is only used for rendering and you could think that rendering is independent from replay logic.
That could be the case, but in our records we store SDL events. SDL events represents mouse movements, clicks etc.
And if we calculate a `game position` from the `mouse position`, we must know what we render to get the correct `game position`. For that case we use [GetOffsetForWalking](https://github.com/diasurgical/devilutionX/blob/2baca9a5ce3798e0ca3cc72fb6b23ccccee9d317/Source/cursor.cpp#L306) and this depends on `gfProgressToNextGameTick`.
That's the reason why this pr stores the current `gfProgressToNextGameTick` for every recorded sdl event.

#### Alternative Solutions

- Use network messages instead of sdl events. But that has it own problems.
- Disable ADL while recording.

## Notes

- This can be improved further. But i think it's a good step.
- While testing I got one desync with the game menu (try to quit the game). But I couldn't reproduce it with another replay and there are still some `SDL_GetTicks` left. So there is still work to do.

I hope this helps a little for future work, testing and more stable releases 🙂

### DIFF
--- a/Source/miniwin/misc_msg.cpp
+++ b/Source/miniwin/misc_msg.cpp
@@ -606,18 +606,19 @@ void CreateDemoFile(int i)
 	demoRecording << "0," << gSaveNumber << "," << gnScreenWidth << "," << gnScreenHeight << "\n";
 }
 
-void SaveDemoMessage(tagMSG *lpMsg, int tick)
+void SaveDemoMessage(tagMSG *lpMsg)
 {
-	demoRecording << tick << ",0," << lpMsg->message << "," << lpMsg->wParam << "," << lpMsg->lParam << "\n";
+	demoRecording << static_cast<uint32_t>(DemoMsgType::Message) << "," << gfProgressToNextGameTick << "," << lpMsg->message << "," << lpMsg->wParam << "," << lpMsg->lParam << "\n";
 }
 
-void PumpDemoMessage(int tick, uint32_t message, int32_t wParam, int32_t lParam)
+void PumpDemoMessage(DemoMsgType demoMsgType, uint32_t message, int32_t wParam, int32_t lParam, float progressToNextGameTick)
 {
 	demoMsg msg;
-	msg.tick = tick;
+	msg.type = demoMsgType;
 	msg.message = message;
 	msg.wParam = wParam;
 	msg.lParam = lParam;
+	msg.progressToNextGameTick = progressToNextGameTick;
 
 	demo_message_queue.push_back(msg);
 }
@@ -657,20 +658,25 @@ bool LoadDemoMessages(int i)
 		std::stringstream command(line);
 
 		std::getline(command, number, ',');
-		int tick = std::stoi(number);
+		int typeNum = std::stoi(number);
+		DemoMsgType type = static_cast<DemoMsgType>(typeNum);
 
 		std::getline(command, number, ',');
-		int type = std::stoi(number);
+		float progressToNextGameTick = std::stof(number);
 
 		switch (type) {
-		case 0:
+		case DemoMsgType::Message: {
 			std::getline(command, number, ',');
 			uint32_t message = std::stoi(number);
 			std::getline(command, number, ',');
 			int32_t wParam = std::stoi(number);
 			std::getline(command, number, ',');
 			int32_t lParam = std::stoi(number);
-			PumpDemoMessage(tick, message, wParam, lParam);
+			PumpDemoMessage(type, message, wParam, lParam, progressToNextGameTick);
+			break;
+		}
+		default:
+			PumpDemoMessage(type, 0, 0, 0, progressToNextGameTick);
 			break;
 		}
 	}
@@ -680,7 +686,7 @@ bool LoadDemoMessages(int i)
 	return true;
 }
 
-bool DemoMessage(tagMSG *lpMsg, int tick)
+bool DemoMessage(tagMSG *lpMsg)
 {
 	SDL_Event e;
 	if (SDL_PollEvent(&e)) {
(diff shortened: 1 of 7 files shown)


AJenbo:
Thank you so much for taking the time to solve this. It is going to mean the world for us to be able to use this when it comes to testing and benchmarking.

> Disable additional rendering (and with that `ProcessInput` calls) while recording

This is definatly worth considering since it is what Doom ports do. But personally I think I prefer refactoring FetchMessage, after all there is no original demo format that we are trying to stay compatible with :D

Source/miniwin/misc_msg.cpp
```diff
@@ -717,17 +724,29 @@ bool DemoMessage(tagMSG *lpMsg, int tick)
...
 
-bool FetchMessage(tagMSG *lpMsg, int tick)
+bool GetDemoRunGameLoop()
+{
+	if (demo_message_queue.empty())
+		app_fatal("Demo queue empty");
+	demoMsg dmsg = demo_message_queue.front();
+	if (dmsg.type == DemoMsgType::Message)
+		app_fatal("Unexpected Message");
+	gfProgressToNextGameTick = dmsg.progressToNextGameTick;
```
AJenbo, line 734: I actually like that we are recording the exact time of input, despite floats not being my favorite kinda data.

obligaron:
> In any event I think that we might want to accept not having a locked down demo format for a few iterations of this feature.

I also think this feature needs time to be mature and that we will break the compatibility multiple times until then. 😉 
But I think it's important that we have a first version. If we have something committed in the repo, It's much easier to improve something that is already running. That way we can also all work together and advance step for step. 

> The tricky part about this is that if we skip or disable parts of the logic we end up not testing some parts. On the other hand I think it would be preferable to abstract away the difference in using mouse vs. using a gamepad etc. And this would make for a very ideal place to record messages and replay them.

I think this is a fundamental question we have to decide soon. 👍 
Cause we have to enforce some rules to the code to make replays work.
For example something like [SDL_GetModeState()](https://github.com/diasurgical/devilutionX/blob/fe034ce70920b5f4fdaa5ff123437f3669e13403/Source/control.cpp#L1019) doesn't work with replays and should be prohibited in the game logic / abstraction.

Source/miniwin/misc_msg.cpp
obligaron: I would also like to get rid of this. :laughing:

in the future I could imagine we would replace every call to `SDL_GetTick()` with a new introduced `DiabloGetTick()`.
And it would look something like:

```
int DiabloGetTick()
{
	if (gbRunGame)
		return gameTick;
	return SDL_GetTicks();
}

int gameTick;
void UpdateGameTick()
{
	if (timedemo) {
		gameTick = GetDemoMessageAsTime();
		return;
	}
	gameTick = SDL_GetTicks();
}
```

`UpdateGameTick()` would only be called at the top of the [loop](https://github.com/diasurgical/devilutionX/blob/fe034ce70920b5f4fdaa5ff123437f3669e13403/Source/diablo.cpp#L805) in `RunGameLoop`.
This has some benefits:

- We can remove specials like in `nthread_has_500ms_passed` or `nthread_UpdateProgressToNextGameTick`
- We don't have to care about new or changed calls To `DiabloGetTick()` causes issues in the replay. Cause there is only one point in time while the loop runs, regardless how often you ask the method. 😉 
- We don't have to store floats and implementation details like `gfProgressToNextGameTick`

I didn't add this to this PR cause it can result in unexpected errors for the normal play logic and I think it's also big enough to justify it's own PR. But I think we improve the timedemo also step for step and make this a great feature. it will take some time as it was before with ADL. 😉  But together I think we can achieve this. 🙂

AJenbo: Agreed, I was thinking that maybe it could even help multiplayer if the game can communicate what the current tick is and that allow the client to better sync up with the server.

AJenbo:
One place where replays currently have a chance to diverge is with dialogs as there display lenght is tied up with the lenght of the audio, which is not sped up during faster play speeds. But we can deal with that later.

AJenbo:
> I think this is a fundamental question we have to decide soon. +1

I'm strongly leaning in the directly of abstracting this away and move to basically recording network messages. This might also help us deal with desync as network would become just another input device and could be tested on equal footing with mouse+keyboard.

PR #2481, opened 2021-08-01
Russian strings.xml for android.

AAS:
As requested, here is russian strings.xml for android.

I tried to keep "shareware" and "spawn" words intact, since none one in modern world knows what "shareware" is, with the word "spawn" things are even worse.


Iniquitatis:
@AJenbo I can't think of any suggestions for this PR, so it can be merged as is.

AJenbo:
At least we already have Cyrillic in this part of the app:

image

PR #2486, opened 2021-08-03
fix name check and limit to multi

qndel:
it didn't work at all before
this makes it work and limits to multi - like in vanilla

### DIFF
--- a/Source/DiabloUI/selhero.cpp
+++ b/Source/DiabloUI/selhero.cpp
@@ -299,17 +299,18 @@ void SelheroClassSelectorEsc()
 
 void SelheroNameSelect(int /*value*/)
 {
-	if (!UiValidPlayerName(selhero_heroInfo.name)) {
+	// only check names in multiplayer, we don't care about them in single
+	if (selhero_isMultiPlayer && !UiValidPlayerName(selhero_heroInfo.name)) { 
 		ArtBackground.Unload();
 		UiSelOkDialog(title, _("Invalid name. A name cannot contain spaces, reserved characters, or reserved words.\n"), false);
 		LoadBackgroundArt("ui_art\\selhero.pcx");
+	} else{
+		if (gfnHeroCreate(&selhero_heroInfo)) {
+			SelheroLoadSelect(1);
+			return;
+		}
+		UiErrorOkDialog(_(/* TRANSLATORS: Error Message */ "Unable to create character."), vecSelDlgItems);
 	}
-
-	if (gfnHeroCreate(&selhero_heroInfo)) {
-		SelheroLoadSelect(1);
-		return;
-	}
-	UiErrorOkDialog(_(/* TRANSLATORS: Error Message */ "Unable to create character."), vecSelDlgItems);
 
 	memset(selhero_heroInfo.name, '\0', sizeof(selhero_heroInfo.name));
 	SelheroClassSelectorSelect(0);


Source/DiabloUI/selhero.cpp
```diff
@@ -299,18 +299,19 @@ void SelheroClassSelectorEsc()
 
 void SelheroNameSelect(int /*value*/)
 {
-	if (!UiValidPlayerName(selhero_heroInfo.name)) {
+	// only check names in multiplayer, we don't care about them in single
+	if (selhero_isMultiPlayer && !UiValidPlayerName(selhero_heroInfo.name)) { 
 		ArtBackground.Unload();
 		UiSelOkDialog(title, _("Invalid name. A name cannot contain spaces, reserved characters, or reserved words.\n"), false);
 		LoadBackgroundArt("ui_art\\selhero.pcx");
+	} else{
```
AJenbo, line 307: `} else {`

Source/DiabloUI/selhero.cpp
```diff
@@ -299,18 +299,19 @@ void SelheroClassSelectorEsc()
 
 void SelheroNameSelect(int /*value*/)
 {
-	if (!UiValidPlayerName(selhero_heroInfo.name)) {
+	// only check names in multiplayer, we don't care about them in single
+	if (selhero_isMultiPlayer && !UiValidPlayerName(selhero_heroInfo.name)) {
```
julealgon, line 303: Why make this specific to multiplayer?

IMHO, this just complicates the code more for no real benefit. Ideally, we want to keep special casing to a minimum.

AJenbo:
Does this also fix the bug that made you look in to this code?

Source/DiabloUI/selhero.cpp
qndel: because that's how it was in the original and nobody cares how you name your stuff in single

qndel:
> Does this also fix the bug that made you look in to this code?
yup

AJenbo:
Yeah, just a bit to tired to grog both changes at once.

Source/DiabloUI/selhero.cpp
qndel: my formatting in visual studio died for some reason :(

Source/DiabloUI/selhero.cpp
julealgon: > because that's how it was in the original

That's a really poor reason to have something in the codebase.

> nobody cares how you name your stuff in single

Do they care in multi?

qndel: More than in single ;) and if they don't, this code is useless

AJenbo: > Do they care in multi?

yes, it's sent to there screen if you play with them.

julealgon: > > Do they care in multi?
> 
> yes, it's sent to there screen if you play with them.

I don't get it. All I see is this validation which seems completely nonsensical to me:
https://github.com/diasurgical/devilutionX/blob/ad6f9cb8e50ef351b198c48f804cf569dafb2957/Source/DiabloUI/diabloui.cpp#L573-L582

Am I missing something?

ephphatha: This is an obfuscated list of vulgar or potentially deceptive terms, the comparisons involve decrementing each character to get the real values ("benjo" -> "admin" for example).

AJenbo: They are encrypted, probably to avoid that people easily remove them via hex edit etc.

ThomasChr: Do we need to have them encrypted or should we change that?

AJenbo: it was like this originally, i don't see reason to change it.

julealgon: This is such a silly protection.

qndel: > This is such a silly protection.

still better than having raw "fuck" in your code or game's memory, also pls don't necro PRs 😢 let them die

julealgon: > still better than having raw "fuck" in your code or game's memory

Who cares though? Are we offending someone for having "fuck" in the codebase? Or worse, why does this even exist in the first place...

> ...also pls don't necro PRs 😢 let them die

It's hard to keep track of these... sometimes it takes a while.

AJenbo: "This is such a silly protection." didn't really feel like a worth while comment to dig this up for. The subject here is even off topic to the PR and i locked the discussion (apparently only applies to outsiders), use the chat for side tracks like this in the future.

julealgon: The discussion is very on-topic to me. We are basically maintaining and changing useless code here. Feel free to ignore though.

PR #2489, opened 2021-08-03
Unify character formulas

qndel:
This finally makes the to hit value shown in character panel useful - I had no idea how broken it was before.
It was basically always showing melee hit chance formula, which gains 1% per 2 dex while ranged gains 1% per 1 dex, which is a huge difference! I also made it include class bonuses and character level = it finally shows the real value that gets used in actual hit chance calculating

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1358,7 +1358,7 @@ void DrawChr(const Surface &out)
 		style = UiFlags::ColorBlue;
 	if (myPlayer._pIBonusToHit < 0)
 		style = UiFlags::ColorRed;
-	sprintf(chrstr, "%i%%", (myPlayer._pDexterity / 2) + myPlayer._pIBonusToHit + 50);
+	sprintf(chrstr, "%i%%", (myPlayer.InvBody[INVLOC_HAND_LEFT]._itype == ITYPE_BOW ? myPlayer.GetRangedToHit() : myPlayer.GetMeleeToHit()));
 	DrawString(out, chrstr, { { 258, 211 }, { 43, 0 } }, style | UiFlags::AlignCenter);
 
 	style = UiFlags::ColorSilver;
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1350,7 +1350,7 @@ void DrawChr(const Surface &out)
 		style = UiFlags::ColorBlue;
 	if (myPlayer._pIBonusAC < 0)
 		style = UiFlags::ColorRed;
-	sprintf(chrstr, "%i", myPlayer._pIBonusAC + myPlayer._pIAC + myPlayer._pDexterity / 5);
+	sprintf(chrstr, "%i", myPlayer.GetArmor());
 	DrawString(out, chrstr, { { 258, 183 }, { 43, 0 } }, style | UiFlags::AlignCenter);
 
 	style = UiFlags::ColorSilver;
--- a/Source/player.h
+++ b/Source/player.h
@@ -396,6 +396,14 @@ struct PlayerStruct {
 	void Reset();
 
 	/**
+	 * @brief Return player's armor value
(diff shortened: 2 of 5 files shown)


vlolteanu:
FWIW, the ranged chance to hit also depends on, and drops steeply with, the distance to the target. I can see why they originally displayed only the melee to-hit.

qndel:
> FWIW, the ranged chance to hit also depends on, and drops steeply with, the distance to the target. I can see why they originally displayed only the melee to-hit.

how is it related? melee hit chance also depends on enemy armor, it's just displaying a shitty value that has 0 connection with actual formula :D

vlolteanu:
> how is it related? melee hit chance also depends on enemy armor, it's just displaying a shitty value that has 0 connection with actual formula :D

That's why I prefaced my comment with "FWIW". I was making the point that actual ranged chance to hit is even more disconnected from the base number, so they didn't even bother with it.

qndel:
That's why this PR fixes all the issues and makes the number displayed in character sheet actually meaningful

NiteKat:
For these reasons, I like that they went with Attack Rating in Diablo 2 instead of a number labeled as To Hit. :)

Source/player.h
```diff
@@ -395,6 +395,49 @@ struct PlayerStruct {
 	 */
 	void Reset();
 
+	/**
+	 * @brief Return armor value displayed in character sheet
+	 */
+	int GetSheetArmor() const {
```
AJenbo, line 401: This is used mostly out side of the sheet, maybe GetBaseArmor(). GetArmorRating() or GetArmor() would be better names.

Source/player.h
```diff
@@ -395,6 +395,49 @@ struct PlayerStruct {
...
 	void Reset();
 
+	/**
+	 * @brief Return armor value displayed in character sheet
+	 */
+	int GetSheetArmor() const {
+		return _pIBonusAC + _pIAC + _pDexterity / 5;
+	}
+
+#define BASE_HIT_CHANCE 50
```
AJenbo, line 405:
```suggestion
constexpr int BaseHitChance = 50;
```

Source/player.h
qndel: probably, I wanted to show that the sheet is not lying anymore xD

Source/control.cpp
```diff
@@ -1350,15 +1350,15 @@ void DrawChr(const Surface &out)
...
+	sprintf(chrstr, "%i", myPlayer.GetSheetArmor());
 	DrawString(out, chrstr, { { 258, 183 }, { 43, 0 } }, style | UiFlags::AlignCenter);
 
 	style = UiFlags::ColorSilver;
 	if (myPlayer._pIBonusToHit > 0)
 		style = UiFlags::ColorBlue;
 	if (myPlayer._pIBonusToHit < 0)
 		style = UiFlags::ColorRed;
-	sprintf(chrstr, "%i%%", (myPlayer._pDexterity / 2) + myPlayer._pIBonusToHit + 50);
+	sprintf(chrstr, "%i%%", ( myPlayer.InvBody[INVLOC_HAND_LEFT]._itype == ITYPE_BOW ? myPlayer.GetSheetRangedToHit() : myPlayer.GetSheetMeleeToHit()));
```
AJenbo, line 1361:
```suggestion
	sprintf(chrstr, "%i%%", (myPlayer.InvBody[INVLOC_HAND_LEFT]._itype == ITYPE_BOW ? myPlayer.GetSheetRangedToHit() : myPlayer.GetSheetMeleeToHit()));

```

AJenbo (dismissed):
Really nice clean up, though I would like a slight adjustment to the naming so that it doesn't appear as if the functions are only for display :)

AJenbo (approved):

PR #2490, opened 2021-08-03
Introduce timedemo based tests

obligaron:
## Content

Now that timedemo got live, we can start to use this also in automated tests.
This PR lay the groundwork for this feature. There are also two samples tests included.

## Notes

- We shouldn't add more tests before the demo format got more stable, cause we have to redo all recorded demos.
- Recording creates now a reference save game. This save contains all changes that are done while replaying and don't override the normal save game.
- The tests work with reference saves (created after recording) that is compared to save games from the replay. If they differ something went wrong.
- The reference save game can also be used in normal timedemo replays. It also verify if the outcome is stable (logs to console).
- Tests run headless (without rendering). This improves replay speed a lot. But I discovered at least one part in rendering functions (spell selection) that is relevant for the replay. In this case I refactored it out of the rendering function.
- I encountered multiple problems where currently the run is not stable. For example [premium items](https://github.com/diasurgical/devilutionX/blob/502b2684c998949fe8f5a2169924bb9d51ad8afa/Source/stores.cpp#L2226) in stores are not demorun stable (`SDL_GetTicks` is used) and this causes different saves. Current workaround: don't go to town while recording. 😉
- Keymapper had a bug (also present in master with normal gameplay) where no default entries where created if `diablo.ini` is not present (fresh installation). But this is needed for timedemo tests, cause we don't deploy a `diablo.ini`.
- Tests are currently disabled in CI. Cause we need a `mpq` for the tests.

## Possible future work

- Make Tests CI compatible (provide a mpq)
- Improve replay format in general (for example fix use of time based functions (see store bug above) or include more infos in format (like individual key mappings).
- Check if more logic happens during rendering and refactor it out (see spell selection bug above)


mewmew:
Really incredible work @obligaron! I've been dreaming of having this feature for such a long time. Can't believe today is day that it finally gets implemented in DevilX. This should help a lot with improving test coverage throughout the engine and also make it easier for us to track regressions in future development.

Keep up the incredible work! :)

Happy summer wishes,
Robin

test/timedemo_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+	InitKeymapActions();
+	LoadOptions();
+
+	const int demoNumber = 0;
+
+	gbIsHellfire = false;
+	gbMusicOn = false;
+	gbSoundOn = false;
+	gbQuietMode = true;
```
AJenbo, line 37: Maybe we should rename this to Headless?

test/Fixtures.cmake
```diff
@@ -61,6 +61,12 @@ set(devilutionx_fixtures
   Levels/L4Data/Vile1.DUN
   Levels/L4Data/Warlord.DUN
   Levels/L4Data/Warlord2.DUN
+  timedemo/RogueLevel1to2/demo_0.dmo
+  timedemo/RogueLevel1to2/demo_0_reference_single_0.sv
+  timedemo/RogueLevel1to2/single_0.sv
+  timedemo/WarriorLevel1to2/demo_0.dmo
```
AJenbo, line 67: I still would prefer if the playback was stored inside of the save (renamed to .dmo) to create a self contained file, that would also just show as bin in git.

test/CMakeLists.txt
```diff
@@ -43,7 +44,10 @@ include(Fixtures.cmake)
 
 foreach(test_target ${tests})
   add_executable(${test_target} "${test_target}.cpp")
-  gtest_discover_tests(${test_target})
+  gtest_discover_tests(
+    ${test_target}
+    TEST_FILTER -Timedemo.*
+  )
```
AJenbo, line 50: I'm a bit split on this one. Why would we not want to run this and get coverage from it? Are the drawbacks significant?

How that DRLG and Item RNG is covered I think it might be ok to have end-to-end tests for covering the majority of the remaining code.

test/timedemo_test.cpp
obligaron: `gbHeadless`?
I think that would be a good idea. 🙂 
As part of this pr or separate?

test/Fixtures.cmake
obligaron: Do you want a unstable format to be part of the save format?
Should we discuss this in a separate issue? 🤔

test/CMakeLists.txt
obligaron: Cause we can't currently. 😢 
For a running game we need more data from the mpq.
For example `LoadLevelSOLData` or dun-files.
Also the trick with providing `pMegaTiles` doesn't work well with `LoadLvlGFX`.
But perhaps i miss something. 🙂

test/timedemo_test.cpp
AJenbo: We can should actually drop the gb notation since we updated the code style.

I would prefer it here since it's getting used a not, and specifically to turn off the graphics

test/Fixtures.cmake
AJenbo: It's just as unstable inside the archive as outside of it, and we can always remove it from the archive using an mpq tool if we want to turn it in to a normal save, although it would work event if we didn't.

test/CMakeLists.txt
AJenbo: Yeah just realized that 😅

But we could make a recording of a shareware run and that would still cover a lot of the mechanics 🤔

test/timedemo_test.cpp
obligaron: No problem. I renamed it to `HeadlessMode`. 🙂

test/Fixtures.cmake
obligaron: We can do this (have a file that contains both the save game and the demo recording).
But I think this is outside of the scope of this pr, cause this pr only uses the currently demo format and don't introduce it.
Also the topic needs a little bit more thoughts, for example: If we have a `demo1.dmo` file that also includes the save game how do we know which version the save game belongs to (diablo, hellfire, spawn)?
I think this is all possible and a good idea but someone told me I should make small pr's. 😉

test/CMakeLists.txt
obligaron: We could download the `spawn.mpq` from `devilutionx-assets` for the tests.
But I don't know how to add this to (correctly) CMake.
Can someone help with this (for example add a commit to this pr) or should we leave this to a follow-up pr?
If wanted I can record the two small tests again with the spawn version,

AJenbo: If you record the demo I will add it, but probably as part of the GitHub action script rather then CMake :)

test/Fixtures.cmake
AJenbo: image ;)

We could check the header in `game` to see which game mode the save is for.
Or we could add it as part of the header line in the recording.

test/CMakeLists.txt
obligaron: But this method would only work when running the game in ci or?
If the repo is cloned and the tests are run locally it will fail or?

AJenbo: The readme clearly states that you must have diabdat.mpq or spawn.mpq for things to work. So that is the humans respectability.

obligaron: The readme also says you can provide them with `--data-dir`, but this doesn't work for unit tests or?

AJenbo: Im not sure i follow your point?

obligaron: Sorry if I was unclear.

Following scenario
- New Developer
- Clones repro
- Compiles program
- Mpq's are not present in default search directory
- Developer uses `--data-dir` to pass a directory where the mpq's stored
- Works 👍 
- Tries to run unit tests
- Unit tests can't find mpq's
- Developer can't use `--data-dir` to pass the directory where the mpq's are stored
- Bad 😢 
- Developer copy's mpq's to bin directory manually, cause this is the only way to fix it. Or is there another way?

I thought we could make the work for the developer more easy if we always download `spawn.mpq`.

Side-node:
We could also ensure that we only load `spawn.mpq` when running in `HeadlessMode` that would make working locally consistent with github actions.

AJenbo: It wasn't so much the steps i didn't get as why it was important.

> Developer can't use --data-dir to pass the directory where the mpq's are stored

If you think this is relevant then just make the additional calls for setting this up during the test would solve it. I don't think it's important though. It's ok to expect developers (who are skilled enough to make use of tests) to also be able to put a file in a specific place. We could add it to the build readme "For tests mpq files must be in folder X" if you think it's important.

> I thought we could make the work for the developer more easy if we always download spawn.mpq.

But some of the tests wont work without diabdat.mpq anyway so i just don't see the point here.

They also already seam upset that we will download ZT dependencies by default, adding another 25MB download rather then asking them to move a file that they are already required to have probably won't make them much happier.

>  We could also ensure that we only load spawn.mpq when running in HeadlessMode that would make working locally consistent with github actions.

I don't think it's good to bind these two options to each other. I'm not asking for the current demos to be removed, I just would like it if there is at least one demo that can be run on the CI so that we can have automatic benefits from the work done here :)

test/Fixtures.cmake
obligaron: With #4799 the most painful issue (large text files in the repro) is fixed.
I hope we can get the embedded demo file later.

test/CMakeLists.txt
obligaron: I changed the following
- Recorded demos with spawn version.
- Removed filter for timedemo unit tests. Now they should also run in github actions.
- Added an assert and comment that spawn.mpq or diabdat.mpq needs to be present for the tests to run

Now the github actions need to be adjusted to download spawn.mpq.

Hope this helps to get a better code coverage. 🙂

AJenbo: This is so exciting 😀

test/timedemo_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+	ASSERT_FALSE(gbRunGame);
+	gbRunGame = false;
+	init_cleanup();
+}
+
+} // namespace
+
+TEST(Timedemo, RogueLevel1to2)
+{
+	RunTimedemo("RogueLevel1to2");
```
AJenbo, line 65: Does this demo work? Spawn does not have the Rogue and I haven't ben able to load the save either.

obligaron: I tried it locally with spaw.mpq.
In normal gameplay it crashes, cause the animations are missing.
Vut with headless mode we don't need the animations and it worked.

AJenbo: It always failed for me in headless

obligaron: Tested again, still works. 😵‍
I have no problem with removing them. They are a little bit hacky, cause rouge is normally not supported in spawn.
Anyway I'm still glad that the tests now run in github actions. 🙂

AJenbo: Once merged you can open a new PR with more tests and if they work then we can merge :D

AJenbo (approved):

PR #2494, opened 2021-08-04
SDL stuff: automatic memory management and error checking, part 1 of N

vlolteanu:
This patch introduces wrappers for a few SDL functions. These wrappers perform error checking and return the appropriate type of unique_ptr.

The wrappers introduced are not employed exhaustively yet. It also includes a refactoring of RenderUTF8_Solid_Wrapped.

The rationale is two-fold:

- Automatic error checking: there are multiple instances in the codebase where the function inside one of these wrappers is called, but no error checking is performed. If the function returns NULL, something bad will happen at a later time.
- The programmer no longer has to figure out what kind of unique_ptr to wrap the result in.

Case in point: RenderUTF8_Solid_Wrapped would return an error when the string was empty, or when some SDL-related error occurred. Its caller, DrawTTF, calls it twice, passing the same text, but only checks the 1st result, falsely assuming that either both would succeed or both would fail.

### DIFF
--- a/Source/DiabloUI/art.cpp
+++ b/Source/DiabloUI/art.cpp
@@ -8,6 +8,7 @@
 #include "utils/display.h"
 #include "utils/log.hpp"
 #include "utils/sdl_compat.h"
+#include "utils/sdl_wrap.h"
 
 namespace devilution {
 namespace {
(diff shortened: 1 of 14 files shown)


AJenbo (approved):

AJenbo (approved):
Nice work, this code was pretty complicated so not easy to make your way around.

PR #2497, opened 2021-08-04
Use string_view in DrawString and friends

vlolteanu:
str::string and char * are converted to this implicitly.


AJenbo (approved):
Thanks, needed this for what I was doing :)

vlolteanu:
No problem. Are you optimizing away the memory copies in tempstr, infostr etc., by any chance?

AJenbo:
No I'm not :)

I'm working on rendering a custom font set and using unicode/utf8 as the text encoding

PR #2498, opened 2021-08-05
Center the sub-panels (Quest, Character, Spell and Inventory)

4aron5un:
According to screen resolution to center the sub-panels

[1440x900]
1440x900_Char_Inv
1440x900_Quest_Spell

[800x600]
800x600_Char_Inv
800x600_Quest_Spell

[640x480]
640x480_Char_Inv
640x480_Quest_Spell

### DIFF
--- a/Source/control.h
+++ b/Source/control.h
@@ -27,8 +27,10 @@ namespace devilution {
 
 #define SPANEL_WIDTH 320
 #define SPANEL_HEIGHT 352
-#define RIGHT_PANEL (gnScreenWidth - SPANEL_WIDTH)
-#define RIGHT_PANEL_X RIGHT_PANEL
+#define LEFT_PANEL_X (gnScreenWidth / 4 - SPANEL_WIDTH / 2)
+#define LEFT_PANEL_Y ((gnScreenHeight - PANEL_HEIGHT - SPANEL_HEIGHT) / 2)
+#define RIGHT_PANEL_X (LEFT_PANEL_X + gnScreenWidth / 2)
+#define RIGHT_PANEL_Y LEFT_PANEL_Y
 
 extern bool drawhpflag;
 extern bool dropGoldFlag;
--- a/Source/control.h
+++ b/Source/control.h
@@ -27,9 +27,9 @@ namespace devilution {
 
 #define SPANEL_WIDTH 320
 #define SPANEL_HEIGHT 352
-#define LEFT_PANEL_X (gnScreenWidth / 4 - SPANEL_WIDTH / 2)
+#define LEFT_PANEL_X 0
 #define LEFT_PANEL_Y ((gnScreenHeight - PANEL_HEIGHT - SPANEL_HEIGHT) / 2)
-#define RIGHT_PANEL_X (LEFT_PANEL_X + gnScreenWidth / 2)
+#define RIGHT_PANEL_X (gnScreenWidth - SPANEL_WIDTH)
 #define RIGHT_PANEL_Y LEFT_PANEL_Y
 
 extern bool drawhpflag;
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -327,17 +327,17 @@ void LeftMouseDown(int wParam)
 
 	if (MousePosition.y < PANEL_TOP || MousePosition.x < PANEL_LEFT || MousePosition.x >= PANEL_LEFT + PANEL_WIDTH) {
 		if (!gmenu_is_active() && !TryIconCurs()) {
-			if (QuestLogIsOpen && MousePosition.x > 32 && MousePosition.x < 288 && MousePosition.y > 32 && MousePosition.y < 308) {
+			if (QuestLogIsOpen && MousePosition.x > LEFT_PANEL_X + 32 && MousePosition.x < LEFT_PANEL_X + 288 && MousePosition.y > LEFT_PANEL_Y + 32 && MousePosition.y < LEFT_PANEL_Y + 308) {
 				QuestlogESC();
 			} else if (qtextflag) {
 				qtextflag = false;
 				stream_stop();
-			} else if (chrflag && MousePosition.x < SPANEL_WIDTH && MousePosition.y < SPANEL_HEIGHT) {
+			} else if (chrflag && MousePosition.x > LEFT_PANEL_X && MousePosition.x < LEFT_PANEL_X + SPANEL_WIDTH && MousePosition.y > LEFT_PANEL_Y && MousePosition.y < LEFT_PANEL_Y + SPANEL_HEIGHT) {
 				CheckChrBtns();
-			} else if (invflag && MousePosition.x > RIGHT_PANEL && MousePosition.y < SPANEL_HEIGHT) {
+			} else if (invflag && MousePosition.x > RIGHT_PANEL_X && MousePosition.x < RIGHT_PANEL_X + SPANEL_WIDTH && MousePosition.y > RIGHT_PANEL_Y && MousePosition.y < RIGHT_PANEL_Y + SPANEL_HEIGHT) {
 				if (!dropGoldFlag)
 					CheckInvItem(isShiftHeld);
-			} else if (sbookflag && MousePosition.x > RIGHT_PANEL && MousePosition.y < SPANEL_HEIGHT) {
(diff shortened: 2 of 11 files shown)


4aron5un:
Centered_Panels

AJenbo:
How does this look in the more common case of 853x480? The benifit of having them to the side is that you can more reasonably play while they are open. Though this might not be ideal for touch controls.

4aron5un:
Thanks.
I got your point about it, and how about this?

Centered_Panels_Side

Attached 853x480 for your reference

[Current]
853x480_Char_Inv_Current
[Centered]
853x480_Char_Inv

[Current]
853x480_Quest_Spell_Current
[Centered]
853x480_Quest_Spell

AJenbo:
> I got your point about it, and how about this?
> 
> Centered_Panels_Side

This looks good :+1:

4aron5un:
Just to share my experience...:)

For desktop user with high resolution and large monitor, current panel is far for me.

So I suggest moving them to the visual focus area.

But for mobile phone players, the touch area is very important.

Maybe we can try this "vertical centered" version, I will commit later...:)

1920x1080_Char_Inv_Current

1920x1080_Char_Inv_Centered

Visual_Focus

4aron5un:
Vertical centered

[640x480]
640x480_Vertical_Centered


[853x480]
853x480_Vertical_Centered


[1440x900]
1440x900_Vertical_Centered


[1920x1080]
1920x1080_Vertical_Centered

AJenbo:
:+1: eventually we can add an option that allows the user to pick where they should be positioned. Doing so would also get rid of the preprocessor defines which would help to improve the code.

glebm:
This is a neat refactoring! Perhaps the refactoring can be merged separately, and then a follow-up PR adjusting the positions?

julealgon:
What about allowing to move the windows in-game, and then persist their position in the save, or in the ini (easier, and would then be global)?

As an alternative, just allowing to set their positions in the ini would be a good start.

FitzRoyX:
Yeah, the cornered panels were not ideal for 720p and up. Too much mouse travel. Bringing them down is a welcome change. I think a clamped solution where high res can enjoy the low mouse travel of "floating" panels without being obstructive to ultra-low resolutions is best.

4aron5un:
Thank you all for the great input :blush:

Please check my TODOs below:
- Separate this change into a macro definition, just like what NONET/USE_SDL1 did
- Disable this feature by default, developer can enable this feature manually
- After we have enough verification, maybe we can consider to enable it by default
- Provide several options in the .ini file, such like (0: Align to the Corner, 1: Vertically Centered, 2: Fully Centered, ...)
- And the next phase, maybe we can implement the in-game adjustment and persist the last position

julealgon:
> Thank you all for the great input 😊
> 
> Please check my TODOs below:
> 
> * Separate this change into a macro definition, just like what NONET/USE_SDL1 did

Why? I think we should avoid macros as much as possible for new code. This is a bad idea imho.

> * Disable this feature by default, developer can enable this feature manually

If the settings for changing the positions are in the .ini, there should be no need for a "master toggle" of the feature.

> * After we have enough verification, maybe we can consider to enable it by default

See above. I disagree with this.

> * Provide several options in the .ini file, such like (0: Align to the Corner, 1: Vertically Centered, 2: Fully Centered, ...)

That's an option, but it might be simpler to just allow one to set the corner(s) position in pixels instead. This would even allow people to swap the positions of the panels entirely (say, have the inventory on the left, and the character screen on the right, etc) or even have the panels one on top of the other in high resolutions.

> * And the next phase, maybe we can implement the in-game adjustment and persist the last position

Yeah I think being able to move the windows inside the game could definitely be a separate, incremental effort from this.

AJenbo:
>     * Separate this change into a macro definition, just like what NONET/USE_SDL1 did
>     * Disable this feature by default, developer can enable this feature manually
>     * After we have enough verification, maybe we can consider to enable it by default

I don't think any of that is needed.

>     * Provide several options in the .ini file, such like (0: Align to the Corner, 1: Vertically Centered, 2: Fully Centered, ...)

Sounds good as a follow up.
 
>     * And the next phase, maybe we can implement the in-game adjustment and persist the last position

I think this is overkill, but I'm open to be proven wrong.

4aron5un:
> > Thank you all for the great input blush
> > Please check my TODOs below:
> > 
> > * Separate this change into a macro definition, just like what NONET/USE_SDL1 did
> 
> Why? I think we should avoid macros as much as possible for new code. This is a bad idea imho.
> 
> > * Disable this feature by default, developer can enable this feature manually
> 
> If the settings for changing the positions are in the .ini, there should be no need for a "master toggle" of the feature.
> 
> > * After we have enough verification, maybe we can consider to enable it by default
> 
> See above. I disagree with this.

OK, got it. Just worried about my change will impact other function in the formal build.
> > * Provide several options in the .ini file, such like (0: Align to the Corner, 1: Vertically Centered, 2: Fully Centered, ...)
> 
> That's an option, but it might be simpler to just allow one to set the corner(s) position in pixels instead. This would even allow people to swap the positions of the panels entirely (say, have the inventory on the left, and the character screen on the right, etc) or even have the panels one on top of the other in high resolutions.
> 
> > * And the next phase, maybe we can implement the in-game adjustment and persist the last position
> 
> Yeah I think being able to move the windows inside the game could definitely be a separate, incremental effort from this.

Thanks for the comment. Let's have an ini version first. :smile:

4aron5un:
> > ```
> > * Separate this change into a macro definition, just like what NONET/USE_SDL1 did
> > * Disable this feature by default, developer can enable this feature manually
> > * After we have enough verification, maybe we can consider to enable it by default
> > ```
> 
> I don't think any of that is needed.
> 
> > ```
> > * Provide several options in the .ini file, such like (0: Align to the Corner, 1: Vertically Centered, 2: Fully Centered, ...)
> > ```
> 
> Sounds good as a follow up.
> 
Got it, I will implement options in .ini file.
> > ```
> > * And the next phase, maybe we can implement the in-game adjustment and persist the last position
> > ```
> 
> I think this is overkill, but I'm open to be proven wrong.

julealgon:
> OK, got it. Just worried about my change will impact other function in the formal build.

If it introduces any issues, we fix them. 

The toggles we currently have in ini are not for this purpose (fear of breaking code/being easy to revert) but for toggling between 2 desirable behavior states for the user.

Source/control.h
```diff
@@ -27,8 +27,10 @@ namespace devilution {
 
 #define SPANEL_WIDTH 320
 #define SPANEL_HEIGHT 352
-#define RIGHT_PANEL (gnScreenWidth - SPANEL_WIDTH)
-#define RIGHT_PANEL_X RIGHT_PANEL
+#define LEFT_PANEL_X 0
+#define LEFT_PANEL_Y ((gnScreenHeight - PANEL_HEIGHT - SPANEL_HEIGHT) / 2)
+#define RIGHT_PANEL_X (gnScreenWidth - SPANEL_WIDTH)
+#define RIGHT_PANEL_Y LEFT_PANEL_Y
```
AJenbo, line 33: When changing this to be configurable these should probably be replaced with a function that returns a point when given an enum that represents one of the 4 panels.

`Point GetPanelPosition(UiPanels panel);`

Source/diablo.cpp
```diff
@@ -327,17 +327,17 @@ void LeftMouseDown(int wParam)
 
 	if (MousePosition.y < PANEL_TOP || MousePosition.x < PANEL_LEFT || MousePosition.x >= PANEL_LEFT + PANEL_WIDTH) {
 		if (!gmenu_is_active() && !TryIconCurs()) {
-			if (QuestLogIsOpen && MousePosition.x > 32 && MousePosition.x < 288 && MousePosition.y > 32 && MousePosition.y < 308) {
+			if (QuestLogIsOpen && MousePosition.x > LEFT_PANEL_X + 32 && MousePosition.x < LEFT_PANEL_X + 288 && MousePosition.y > LEFT_PANEL_Y + 32 && MousePosition.y < LEFT_PANEL_Y + 308) {
```
AJenbo, line 330: Stuff like this should end up some thing like this:

```cpp
for (auto panelType : UiPanels) {
    if (!IsPanelOpen(panelType)
        continue;
        
    Rectangle panel { GetPanelPosition(panelType), { 320, 352 } };
    if (panel.Contains(MousePosition)) {
        return true;
    }
}

return false;
```

AJenbo (approved):

Source/control.h
4aron5un: I will do the refactoring for this.

Source/diablo.cpp
4aron5un: Got it.

This will become a function such like IsCursorOnPanel().

PR #2499, opened 2021-08-05
unify block formula

qndel:


Source/player.h
```diff
@@ -442,6 +442,18 @@ struct PlayerStruct {
 		return hper;
 	}
 
+	/**
+	 * @brief Return block chance
+	 * @param useLevel - indicate if player's level should be added to block chance (the only case where it isn't is blocking a trap)
```
AJenbo, line 447: Is this not a bug?

qndel: depends on how you look at it, I'd say the traps are enough of a joke already, no need to nerf them further by increasing block chance against them :D

AJenbo (approved):

PR #2501, opened 2021-08-05
Hardware Cursor: Default to ON only on SDL 2.0.12+

### LINKED ISSUE #2445: Hardware cursor not supported on Raspberry Pi 4, but still enabled
When using the joystick on a raspberry pi 4 I cannot see where the mouse pointer is when navigating to boxes in my inventory or increasing character stats. This was working not too long ago, im afraid it may have been removed.

**Describe**
I cannot see where the mouse pointer is on the screen when managing my inventory items or increasing character attribute points after leveling up.

**To Reproduce**
Use a controller when playing devilutionx, and go try to move your gold around. even better move it to the middle of the inventory close your inventory, and try to pick up the gold again. It will be difficult.

**Expected behavior**
See which box the mouse cursor is on so I know which button to press on the joystick when managing inventory or increasing character stats.

glebm:
Fixes #2445

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -157,6 +157,15 @@ void SaveIni()
 	IniChanged = false;
 }
 
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+bool HardwareCursorDefault()
+{
+	SDL_version v;
+	SDL_GetVersion(&v);
+	return SDL_VERSIONNUM(v.major, v.minor, v.patch) >= SDL_VERSIONNUM(2, 0, 12);
+}
+#endif
+
 } // namespace
 
 void SetIniValue(const char *sectionName, const char *keyName, const char *value, int len)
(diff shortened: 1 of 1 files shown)


Source/options.cpp
```diff
@@ -157,6 +157,15 @@ void SaveIni()
 	IniChanged = false;
 }
 
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+bool HardwareCursorDefault()
+{
+	SDL_version v;
+	SDL_GetVersion(&v);
+	return SDL_VERSIONNUM(v.major, v.minor, v.patch) >= SDL_VERSIONNUM(2, 0, 12);
```
AJenbo, line 165: Maybe I'm just being silly, but a safer comparison would be:
```suggestion
	return v.major > 1 && (v.major > 2 || v.minor > 0 || v.patch >= 12);
```

glebm: I think we can trust the SDL-provided macro. That's what SDL uses anyway for things like `SDL_VERSION_ATLEAST`.

AJenbo: That's at least a comfort :)

PR #2505, opened 2021-08-05
Eliminate a lot of string copies

vlolteanu:
This PR introduces StringVariant (stupid name, feel free to suggest a different one). It is a wrapper around string_view, with an optional backing string.

Various stuff, like messages that appear in the panel at the bottom of the screen, can now either be either copied or passed as a string_view, as appropriate.

### DIFF
--- a/Source/utils/string_variant.hpp
+++ b/Source/utils/string_variant.hpp
@@ -30,7 +30,18 @@ public:
 	{
 	}
 
+	explicit StringVariant(char *cstr)
+	    : str(cstr)
+	{
+	}
+
 	StringVariant &operator =(const char *cstr)
+	{
+		*this = StringVariant(cstr);
+		return *this;
+	}
+
+	StringVariant &operator =(char *cstr)
 	{
 		*this = StringVariant(cstr);
 		return *this;
--- a/Source/utils/string_variant.hpp
+++ b/Source/utils/string_variant.hpp
@@ -53,7 +53,7 @@ public:
 		return *this;
 	}
 
-	StringVariant &operator =(const std::string &str)
+	StringVariant &operator =(std::string &str)
 	{
 		*this = StringVariant(std::move(str));
 		return *this;
--- a/Source/utils/string_variant.hpp
+++ b/Source/utils/string_variant.hpp
@@ -1,0 +1,66 @@
+#pragma once
+
+#include <string>
+#include "utils/stdcompat/string_view.hpp"
+
+namespace devilution {
+
+class StringVariant
+{
+	std::string str;
+	string_view view;
+
+public:
+	StringVariant()
+	{
+	}
+
+	explicit StringVariant(string_view view)
+	    : view(view)
+	{
+	}
+
+	explicit StringVariant(std::string str)
+	    : str(std::move(str))
+	{
+	}
+
+	explicit StringVariant(const char *cstr)
+	    : view(cstr)
+	{
+	}
+
+	StringVariant &operator =(const char *cstr)
+	{
+		*this = StringVariant(cstr);
+		return *this;
+	}
+
+	StringVariant &operator =(const string_view &view)
+	{
+		*this = StringVariant(view);
+		return *this;
+	}
+
+	StringVariant &operator =(const std::string &str)
+	{
+		*this = StringVariant(std::move(str));
+		return *this;
+	}
+
+	size_t length() const
+	{
+		if (str.length() > 0)
+			return str.length();
+		return view.length();
+	}
+
+	operator string_view() const
+	{
+		if (str.length() > 0)
+			return str;
+		return view;
+	}
+};
+
+} // namespace devilution
--- a/Source/utils/string_variant.hpp
+++ b/Source/utils/string_variant.hpp
@@ -5,6 +5,12 @@
 
 namespace devilution {
 
+/**
+ * @brief Offers an interface similar to variant<string, string_view>
+ * Assignments from char * work like assignments from string, and those from const char *, like assignments from string_view.
+ * The intention is to use string and string_view almost exclusively in the codebase.
+ * The constructors and assignment operators dealing with (const) char * will go away once a critical mass is reached.
+ */
 class StringVariant
 {
 	std::string str;
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -57,7 +57,7 @@ bool talkflag;
 bool sbookflag;
 bool chrflag;
 bool drawbtnflag;
-char infostr[64];
+StringVariant infostr;
 bool panelflag;
 int initialDropGoldValue;
 bool panbtndown;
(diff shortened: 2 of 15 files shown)


julealgon:
> This PR introduces StringVariant (stupid name, feel free to suggest a different one). It is a wrapper around string_view, with an optional backing string.

Not a fan of introducing custom stuff like this. @glebm , is there another way this could be done that doesn't force us into custom types leaking everywhere?

vlolteanu:
I did consider std::variant<string, string_view> (see where the name StringVariant came from?), but it's a lot more cumbersome to use in this context. E.g. it won't implicitly convert const char * to string_view when assigning to it. It can also throw when you try to look at its contents.

vlolteanu:
> custom types leaking everywhere

There are a grand total of 3 StringVariant variables and one function returning it, with little potential for a lot more cropping up. 80% of strcpys are gone.

AJenbo:
Seams like a reasonable solution, but maybe it should have a comment that it is a helper for converting strings and should eventually be removed once all text is converted to use string/string_view?

We can then gradually remove one of its constructors and see what is left to change before it can be removed.

vlolteanu:
@AJenbo Yes, that sounds about right.

Source/utils/string_variant.hpp
```diff
@@ -0,0 +1,77 @@
+#pragma once
+
+#include <string>
+#include "utils/stdcompat/string_view.hpp"
+
+namespace devilution {
+
+class StringVariant
```
AJenbo, line 8:
```suggestion
/**
 * Helper for conversion between various string formats while we are in the process to migrating to purely using string/string_view
 */
class StringVariant
```

julealgon:
> Seams like a reasonable solution, but maybe it should have a comment that it is a helper for converting strings and should eventually be removed once all text is converted to use string/string_view?
> 
> We can then gradually remove one of its constructors and see what is left to change before it can be removed.

It's a stopgap solution, then? What exactly prevents us from just getting rid of the `char*` directly?

The problem I have with this is the tendency these types have to start being used more and more throughout the project. @vlolteanu indicates this is unlikely with this particular one, but even then I'm skeptical.

AJenbo:
> It's a stopgap solution, then? What exactly prevents us from just getting rid of the `char*` directly?

Time and effort

vlolteanu:
I think there's a misunderstanding here. This particular implementation is a stopgap (we can use std::variant later), but it's unlikely to ever go away unless you bite the bullet and make extra string copies where things are ambiguous.

To be clear:

1. Whenever the strings don't need to be owned (e.g. _("Stay a while and listen")), we use string_view. There's no reason to go beyond string_view for stuff that only ever houses or returns unowned strings.
2. Whenever the strings need to be owned, because they have to be freed later (e.g. fmt::format("You're rich! You have this much gold {:d}", gold)), we use std::string. Incidentally, looking at a std::string's contents using string_view is practical.
3. When a variable, depending on circumstances, sometimes doesn't need to own the string (e.g. _("Indestructible")), but some other times it does (("durability {}/{}", 12, 30)), is when this wrapper (or something like it) is appropriate. Looking at either one's contents can be done using string_view.
 
To get rid of (3), the only real solution is to use std::string instead of StringVariant (or std::variant<string, string_view>). Everything will still work, but we're doing extra memory copies for stuff that doesn't need to be owned.

Source/utils/string_variant.hpp
vlolteanu: This wasn't meant as a string conversion tool. (In fact, I'd argue against using it solely for this purpose.) I'll word things differently.

Source/utils/string_variant.hpp
```diff
@@ -0,0 +1,77 @@
...
+{
+	std::string str;
+	string_view view;
+
+public:
+	StringVariant()
+	{
+	}
+
+	explicit StringVariant(string_view view)
```
glebm, line 18: Rather than these constructors, I think it maybe clearer to have factory methods:

```c++
StringVariant::Owned(...);
StringVariant::View(...);
```

Otherwise it is not obvious at call sites which one is being constructed

Source/utils/string_variant.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+	StringVariant &operator =(const string_view &view)
+	{
+		*this = StringVariant(view);
+		return *this;
+	}
+
+	StringVariant &operator =(const std::string &str)
+	{
+		*this = StringVariant(std::move(str));
```
glebm, line 58: Moving a `const` does nothing

Source/utils/string_variant.hpp
```diff
@@ -0,0 +1,77 @@
...
+	    : view(cstr)
+	{
+	}
+
+	explicit StringVariant(char *cstr)
+	    : str(cstr)
+	{
+	}
+
+	StringVariant &operator =(const char *cstr)
```
glebm, line 38: See comment above about factory methods. Can remove all of these assignment operators and just have an assignment operator from `StringVariant`

Source/utils/string_variant.hpp
vlolteanu: That's even more tedious than `std::variant<string, string_view>`. At least with `variant`, you only have to worry about `const char *` and `char[]`; the others you can assign directly.

The long term plan is to get rid of using char for strings (and these constructors along with them).

glebm: Tedious is fine, code is read more than it is written!

glebm: And with factory methods like this, you wouldn't need `char *` overloads at all, just a `string_view` one.

vlolteanu: It's tedious to read, as well.

vlolteanu: Also more error prone. It's now on the programmer to find the right type of StringVariant. Made an unowned one from a std::string? The game may display garbage.

Source/utils/string_variant.hpp
vlolteanu: Good catch

AJenbo:
@vladtepesch @glebm do you guys think we can resolve the remaning issues with this PR so that we can merge it :)

@vladtepesch let me know if you would like me to resolve the conflicts that have happened as a results of other code changes.

vlolteanu:
> @vlolteanu @glebm do you guys think we can resolve the remaning issues with this PR so that we can merge it :)

The problem is that we disagree on the API. I think that this should be more like a variant, where you assign either strings or string_views, or equivalents thereof. In the rare instances (none, so far) where you know the string will outlive the StringVariant, or that the string_view will will turn invalid in the StringVariant's lifetime, you just cast to the other type. @glebm is of the opinion that you should go the extra step and explicitly say whether or not you want a string copy made for every assignment (and there are hundreds of them).
 
> @vlolteanu let me know if you would like me to resolve the conflicts that have happened as a results of other code changes.

Sure thing; go right ahead.

vlolteanu:
@glebm Have at you! :)

AJenbo:
I did a rough rebase of this. Most of the changes have now been applied via general `char *` clean ups. Mostly it is only `panelstr` that appear to be using it. So using `std::variant` feels pretty doable at this point. But I'm honestly not sure how much of a benefit it would be to avoid the string copying that is done here?

In any case I'm closing this PR as it looks like it would make more sens to do a fresh take rather then rebasing this one (preferably clean up the remaining `char *` code in separate PRs to keep controversy to a minimal change set).

PR #2513, opened 2021-08-06
Add options for side panel alignment

4aron5un:
Support 3x3 combinations
- Horizontal: left/center/right
- Vertical: top/center/bottom

Alignment


Source/options.cpp
```diff
@@ -259,6 +259,8 @@ void LoadOptions()
 	sgOptions.Gameplay.bRandomizeQuests = GetIniBool("Game", "Randomize Quests", true);
 	sgOptions.Gameplay.bShowMonsterType = GetIniBool("Game", "Show Monster Type", false);
 	sgOptions.Gameplay.bDisableCripplingShrines = GetIniBool("Game", "Disable Crippling Shrines", false);
+	sgOptions.Gameplay.nSPanelHAlign = GetIniInt("Game", "Sub Panel Horizontal Alignment", 0);
+	sgOptions.Gameplay.nSPanelVAlign = GetIniInt("Game", "Sub Panel Vertical Alignment", 0);
```
AJenbo, line 263: I think s stands for side panels, not sub panels

FitzRoyX:
Mouse travel and player obstruction makes half of these seem pointless, like offering a fork with soup. Center/center is the best tradeoff and the most aesthetically pleasing, just go with that.

AJenbo:
> Mouse travel and player obstruction makes half of these seem pointless, like offering a fork with soup. Center/center is the best tradeoff and the most aesthetically pleasing, just go with that.

The fact that we do not agree pretty much proves that having one fixed position is the wrong solution. For me left/center is the best option. Possibly it could be argued that we should only support a sub-set but then we have to pick which and might remove one that some one prefer, and the code won't be much simpler from it.

FitzRoyX:
Well at least you're not picking one from the bottom. I think you're going to realize later that moving items between a future stash and inventory will get old quickly with the panels spread out that far. The money spot is keeping them as close as possible without affecting navigation. If you can't see obstacles at all or until they're right next to you, the panels are too close together.

Source/options.cpp
4aron5un: Modified

AJenbo:
>  I think you're going to realize later that moving items between a future stash and inventory will get old quickly with the panels spread out that far.

Are you thinking that a stash would have to be a panel and that such a panel has to be where the char/quest panel is?
Some stashes work like the shops, which makes it completely irrelevant where the panels are. If a stash is made to work like the inventory then simply have it where the unique item dialog is would solve that issue. Some users might also not care to even use the stash so it's position would be irrelevant.

In any case we shouldn't limit our self's to what might be in the future.

Source/controls/plrctrls.cpp
```diff
@@ -558,7 +558,7 @@ Point InvGetEquipSlotCoordFromInvSlot(const inv_xy_slot slot)
 Point InvGetSlotCoord(int slot)
 {
 	assert(slot <= SLOTXY_INV_LAST);
-	return { InvRect[slot].x + RIGHT_PANEL_X, InvRect[slot].y + RIGHT_PANEL_Y };
+	return GetPanelPosition(UiPanels::Inventory, { InvRect[slot].x, InvRect[slot].y });
```
AJenbo, line 561:
```suggestion
	return GetPanelPosition(UiPanels::Inventory, InvRect[slot]);
```

Source/diablo.cpp
```diff
@@ -393,8 +393,8 @@ void RightMouseDown()
 		SetSpell();
 		return;
 	}
-	if (MousePosition.y <= RIGHT_PANEL_Y || MousePosition.y >= RIGHT_PANEL_Y + SPANEL_HEIGHT
-	    || ((!sbookflag || MousePosition.x <= RIGHT_PANEL_X || MousePosition.x >= RIGHT_PANEL_X + SPANEL_WIDTH)
+	if (MousePosition.y <= rectRightPanel.position.y || MousePosition.y >= rectRightPanel.position.y + SPANEL_HEIGHT
```
AJenbo, line 396:
```suggestion
	if ((!sbookflag || !rectRightPanel.Contains(MousePosition))
```

Source/control.h
```diff
@@ -56,6 +52,9 @@ extern bool panelflag;
 extern int initialDropGoldValue;
 extern bool panbtndown;
 extern bool spselflag;
+extern Rectangle rectMainPanel;
+extern Rectangle rectLeftPanel;
+extern Rectangle rectRightPanel;
```
AJenbo, line 57: Don't prefix them with there type, and globals start with upper case:

This would work:
```suggestion
extern Rectangle MainPanel;
extern Rectangle LeftPanel;
extern Rectangle RightPanel;
```

Source/diablo.cpp
```diff
@@ -393,8 +393,8 @@ void RightMouseDown()
 		SetSpell();
 		return;
 	}
-	if (MousePosition.y <= RIGHT_PANEL_Y || MousePosition.y >= RIGHT_PANEL_Y + SPANEL_HEIGHT
-	    || ((!sbookflag || MousePosition.x <= RIGHT_PANEL_X || MousePosition.x >= RIGHT_PANEL_X + SPANEL_WIDTH)
+	if (MousePosition.y <= rectRightPanel.position.y || MousePosition.y >= rectRightPanel.position.y + SPANEL_HEIGHT
+	    || ((!sbookflag || MousePosition.x <= rectRightPanel.position.x || MousePosition.x >= rectRightPanel.position.x + SPANEL_WIDTH)
```
AJenbo, line 397: Looks like the original might have had some broken logic
```suggestion
	if ((!sbookflag || !rectRightPanel.Contains(MousePosition))
```

Source/inv.cpp
```diff
@@ -1203,7 +1203,7 @@ void DrawInv(const Surface &out)
 		if (myPlayer.InvGrid[i] != 0) {
 			InvDrawSlotBack(
 			    out,
-			    InvRect[i + SLOTXY_INV_FIRST] + Displacement { RIGHT_PANEL_X, RIGHT_PANEL_Y - 1 },
+			    InvRect[i + SLOTXY_INV_FIRST] + Displacement { rectRightPanel.position.x, rectRightPanel.position.y - 1 },
```
AJenbo, line 1206: Maybe
```suggestion
			    GetPanelPosition(UiPanels::Inventory, InvRect[i + SLOTXY_INV_FIRST]) + Displacement { 0, -1 },
```

Source/scrollrt.cpp
```diff
@@ -1212,7 +1212,7 @@ void DrawView(const Surface &out, int startX, int startY)
 		DrawQuestLog(out);
 	}
 	if (!chrflag && Players[MyPlayerId]._pStatPts != 0 && !spselflag
-	    && (!QuestLogIsOpen || PANEL_Y >= LEFT_PANEL_Y + SPANEL_HEIGHT + 74 || PANEL_X >= LEFT_PANEL_X + SPANEL_WIDTH)) {
+	    && (!QuestLogIsOpen || !rectLeftPanel.Contains({ rectMainPanel.position.x, rectMainPanel.position.y - 74 }))) {
```
AJenbo, line 1215:
```suggestion
	    && (!QuestLogIsOpen || !rectLeftPanel.Contains(rectMainPanel + Displacement { 0,  -74 }))) {
```

Source/utils/display.cpp
```diff
@@ -121,6 +124,48 @@ void CalculatePreferdWindowSize(int &width, int &height)
 #endif
 }
 
+void CalculateSubPanelArea(void)
```
AJenbo, line 127: Sub -> Side, but it also does main so:
```suggestion
void CalculatePanelAreas(void)
```

Source/control.cpp
```diff
@@ -795,6 +795,41 @@ void ClearPanel()
...
+		return invflag;
+	default:
+		return false;
+	}
+}
+
+Point GetPanelPosition(UiPanels panel, Point offset)
+{
+	switch(panel)
+	{
```
AJenbo, line 819: Please run clang-format, or manually make sure to apply the code style.

Source/control.cpp
```diff
@@ -795,6 +795,41 @@ void ClearPanel()
...
+	case UiPanels::Main:
+		return { rectMainPanel.position.x + offset.x, rectMainPanel.position.y + offset.y };
```
AJenbo, line 821:
```suggestion
		return rectMainPanel.position + offset;
```

Source/control.cpp
```diff
@@ -1574,7 +1609,7 @@ void CheckChrBtns()
 			continue;
 		}
 		auto buttonId = static_cast<size_t>(attribute);
-		if (ChrBtnsRect[buttonId].Contains({ MousePosition.x - LEFT_PANEL_X, MousePosition.y - LEFT_PANEL_Y })) {
+		if (ChrBtnsRect[buttonId].Contains({ MousePosition.x - rectLeftPanel.position.x, MousePosition.y - rectLeftPanel.position.y })) {
```
AJenbo, line 1612:
```suggestion
		if (GetPanelPosition(UiPanels::Character, ChrBtnsRect[buttonId]).Contains(MousePosition)) {
```

AJenbo (changes requested):
There are a few places where you have inherited flipped logic, and a little bit of code style adjustments needed.

Other then that this is looking really good, a much needed clean up!

4aron5un:
Thanks for the review, I will check your comments.

julealgon:
While this is a definite improvement over what we currently have, I think it is still a bit less flexible than what we could go with.

For example, this does not allow one to swap the positions of the inventory and the character screen.

I'd like to see an approach that was more general purpose, as that could be extremely useful for modding the UI later.

I'm not opposed to having "named alignments" like the ones here, but the underlying mechanism should still be using raw pixel positions and also, ideally, accept those pixels as input in the .ini too.

Source/control.h
4aron5un: I will modify related part and update later.

Source/utils/display.cpp
4aron5un: I will modify related part and update later.

Source/controls/plrctrls.cpp
```diff
@@ -567,7 +567,7 @@ Point InvGetSlotCoord(int slot)
 Point BeltGetSlotCoord(int slot)
 {
 	assert(slot >= SLOTXY_BELT_FIRST && slot <= SLOTXY_BELT_LAST);
-	return { InvRect[slot].x + PANEL_LEFT, InvRect[slot].y + PANEL_TOP };
+	return GetPanelPosition(UiPanels::Main, { InvRect[slot].x, InvRect[slot].y });
```
AJenbo, line 570:
```suggestion
	return GetPanelPosition(UiPanels::Main, InvRect[slot]);
```

Source/inv.cpp
```diff
@@ -1184,12 +1184,12 @@ void DrawInv(const Surface &out)
 					if (myPlayer._pClass != HeroClass::Barbarian
 					    || (myPlayer.InvBody[slot]._itype != ITYPE_SWORD
 					        && myPlayer.InvBody[slot]._itype != ITYPE_MACE)) {
-						InvDrawSlotBack(out, { RIGHT_PANEL_X + slotPos[INVLOC_HAND_RIGHT].x, RIGHT_PANEL_Y + slotPos[INVLOC_HAND_RIGHT].y }, { slotSize[INVLOC_HAND_RIGHT].width * InventorySlotSizeInPixels.width, slotSize[INVLOC_HAND_RIGHT].height * InventorySlotSizeInPixels.height });
+						InvDrawSlotBack(out, GetPanelPosition(UiPanels::Inventory, { slotPos[INVLOC_HAND_RIGHT].x, slotPos[INVLOC_HAND_RIGHT].y }), { slotSize[INVLOC_HAND_RIGHT].width * InventorySlotSizeInPixels.width, slotSize[INVLOC_HAND_RIGHT].height * InventorySlotSizeInPixels.height });
```
AJenbo, line 1187:
```suggestion
						InvDrawSlotBack(out, GetPanelPosition(UiPanels::Inventory, slotPos[INVLOC_HAND_RIGHT]), { slotSize[INVLOC_HAND_RIGHT].width * InventorySlotSizeInPixels.width, slotSize[INVLOC_HAND_RIGHT].height * InventorySlotSizeInPixels.height });
```

Source/inv.cpp
```diff
@@ -1215,7 +1215,7 @@ void DrawInv(const Surface &out)
 
 			const auto &cel = GetInvItemSprite(frame);
 			const int celFrame = GetInvItemFrame(frame);
-			const Point position { InvRect[j + SLOTXY_INV_FIRST].x + RIGHT_PANEL_X, InvRect[j + SLOTXY_INV_FIRST].y + RIGHT_PANEL_Y - 1 };
+			const Point position = GetPanelPosition(UiPanels::Inventory, { InvRect[j + SLOTXY_INV_FIRST].x, InvRect[j + SLOTXY_INV_FIRST].y - 1 });
```
AJenbo, line 1218:
```suggestion
			const Point position = GetPanelPosition(UiPanels::Inventory, InvRect[j + SLOTXY_INV_FIRST])+ Displacement { 0, -1 };
```

AJenbo:
> While this is a definite improvement over what we currently have, I think it is still a bit less flexible than what we could go with.
> 
> For example, this does not allow one to swap the positions of the inventory and the character screen.
> 
> I'd like to see an approach that was more general purpose, as that could be extremely useful for modding the UI later.
> 
> I'm not opposed to having "named alignments" like the ones here, but the underlying mechanism should still be using raw pixel positions and also, ideally, accept those pixels as input in the .ini too.

I mostly agree, but I think that should come later, I don't see anything here that isn't moving in that direction.

`CalculatePanelAreas()` figures out where to place the panels on a pixel level, if the remaining direct usages of `rectRightPanel` and `rectLeftPanel` are replaced with `GetPanelPosition()` then it should be very trivial to swap positions etc. by adding a mapping between positions and panels in `GetPanelPosition()`.

Exposing the raw px position in the ini could be a bit to gritty for a lot of users, especially with various screen scaling going on. I think it would also be best to control the UI in relative terms as we would eventually want to have the UI scalable.

Source/utils/display.cpp
AJenbo: By the way I don't think this really belongs in `utils/display.cpp` as it is part of the game interface and not the computer abstraction. I suggest moving it to `control.cpp`

Source/diablo.cpp
```diff
@@ -393,8 +393,7 @@ void RightMouseDown()
 		SetSpell();
 		return;
 	}
-	if (MousePosition.y <= RIGHT_PANEL_Y || MousePosition.y >= RIGHT_PANEL_Y + SPANEL_HEIGHT
-	    || ((!sbookflag || MousePosition.x <= RIGHT_PANEL_X || MousePosition.x >= RIGHT_PANEL_X + SPANEL_WIDTH)
+	if ((!sbookflag || !rectRightPanel.Contains(MousePosition))
 	        && !TryIconCurs()
 	        && (pcursinvitem == -1 || !UseInvItem(MyPlayerId, pcursinvitem)))) {
```
AJenbo, line 398: This should fix the build
```suggestion
	        && (pcursinvitem == -1 || !UseInvItem(MyPlayerId, pcursinvitem))) {
```

Source/control.h
4aron5un: Modified in the last commit.

Source/control.cpp
```diff
@@ -1574,7 +1651,7 @@ void CheckChrBtns()
 			continue;
 		}
 		auto buttonId = static_cast<size_t>(attribute);
-		if (ChrBtnsRect[buttonId].Contains({ MousePosition.x - LEFT_PANEL_X, MousePosition.y - LEFT_PANEL_Y })) {
+		if (ChrBtnsRect[buttonId].Contains({ MousePosition.x - LeftPanel.position.x, MousePosition.y - LeftPanel.position.y })) {
```
AJenbo, line 1654: It seams a bit non nonsensical to move the mouse position rather then the button.

```suggestion
		Rectangle button = ChrBtnsRect[buttonId];
		button.position = GetPanelPosition(UiPanels::Character, button.position);
		if (button.Contains(MousePosition)) {
```
Same for line 1670

Also it looks like you also fully missed this adjustment in `AttrIncBtnSnap()` so it's probably not possible to level up using a gamepad atm.

Source/utils/display.cpp
```diff
@@ -253,6 +256,8 @@ bool SpawnWindow(const char *lpWindowName)
 		AdjustToScreenGeometry(width, height);
 	}
 
+	CalculatePanelAreas();
```
AJenbo, line 259: Call this from `InitControlPan()` instead.

Source/utils/display.cpp
```diff
@@ -34,6 +34,9 @@ extern SDL_Surface *renderer_texture_surface; /** defined in dx.cpp */
 Uint16 gnScreenWidth;
 Uint16 gnScreenHeight;
 Uint16 gnViewportHeight;
+Rectangle MainPanel;
+Rectangle LeftPanel;
+Rectangle RightPanel;
```
AJenbo, line 39: This also needs to live in `control.cpp` :)

Source/control.cpp
4aron5un: The gamepad issue has been fixed.

Source/utils/display.cpp
4aron5un: It make sense :smile:

AJenbo (dismissed):
I'll probably do a bit of rebasing before merging since there are a few iterations reflected in the commits.

4aron5un:
> I'll probably do a bit of rebasing before merging since there are a few iterations reflected in the commits.

Noted and thanks.

FitzRoyX:
> Are you thinking that a stash would have to be a panel and that such a panel has to be where the char/quest panel is?

Unless you plan to force a minimum resolution, you can only show two panels at once and... we need to be able to drag items between stash and inventory. If they're not both viewable at the same time, how do you do that? Clicking a tiny button with a cursored item?

> Some stashes work like the shops, which makes it completely irrelevant where the panels are.

Being forced to scroll past items you don't care about to see other items is bad design (see: shopping for staves at Adria). The inventory method for vendors is a way better use of screen real estate and would save thousands of clicks. Lists are a dead end.

> If a stash is made to work like the inventory then simply have it where the unique item dialog is would solve that issue. Some users might also not care to even use the stash so it's position would be irrelevant.

That's resolution dependent and obscures your character and the drop space around his feet.

AJenbo:
> > Are you thinking that a stash would have to be a panel and that such a panel has to be where the char/quest panel is?
> 
> Unless you plan to force a minimum resolution, you can only show two panels at once and... we need to be able to drag items between stash and inventory.

You know I don't, saying so is simply facetious. That could be handled in a smart way instead.

> If they're not both viewable at the same time, how do you do that? Clicking a tiny button with a cursored item?

Could be shift-click like we already have for quick relocating of items. Lets look for solutions rather then problems.
 
> > Some stashes work like the shops, which makes it completely irrelevant where the panels are.
> 
> Being forced to scroll past items you don't care about to see other items is bad design (see: shopping for staves at Adria). The inventory method for vendors is a way better use of screen real estate and would save thousands of clicks. Lists are a dead end.

That's your opinion but it may not be universally held. Lists can work better in some situations (gamepad for example), and there are ways to improve on how lists work.
 
> > If a stash is made to work like the inventory then simply have it where the unique item dialog is would solve that issue. Some users might also not care to even use the stash so it's position would be irrelevant.
> 
> That's resolution dependent and obscures your character and the drop space around his feet.

So is mostly everything you mentioned. I would prefer that debates like this be held in the discussion area or on the chat so we don't clutter the pull request process with options about ancillary things.

julealgon:
@FitzRoyX 
> Being forced to scroll past items you don't care about to see other items is bad design (see: shopping for staves at Adria). The inventory method for vendors is a way better use of screen real estate and would save thousands of clicks. Lists are a dead end.

We discussed this at some point in discord and, while the list approach is unpopular mostly due to "Diablo 2 is a perfect game and we should just blindly copy everything it does", this is all 100% subjective. As Ajenbo mentioned, there are ways to make lists work well too (for example, the D3 system is not terrible and uses a list).

Depending on exactly what you are viewing in "grid mode", you have to scroll past items just as much as you do on a list, arguably with a worse interface at that point. Purchasing staves in D2 is a pain too since you are forced to highlight every single staff to check their attributes, since you can't tell anything apart just by the pictures besides the base item (a useless attribute on staves, for the most part).

A list approach could potentially allow you to filter items in a way that would make searching for specific items _a lot_ easier than the grid approach would.

It is not as black and white as you imply it is.

Lastly, we don't yet have an approved stash implementation in DevilutionX and until we do, we shouldn't base any other change on assumptions of one concrete stash implementation IMHO.

FitzRoyX:
Yes, many PC players noticed that the D3 item art and interface regressed so that the devs didn't have to write a separate interface for console ports. That's just a financial decision. You may not think d2 was the roundest wheel, but I'd be careful thinking that they were way off. Moving the cursor doesn't involve clicking and the size of mouseover descriptions doesn't impact the distance to skip.

AJenbo:
Stop debating in pull requests!

AJenbo:
Merged via: https://github.com/diasurgical/devilutionX/commit/3ee90ca62945804097b2b8f2594ae7901804b92e

Changes:
Didn't include the two last commits since they came after the PR was approved, and I don't think I agree with the change.

`IsPanelOpen()` is unused and so not included (feel free to add it in a follow up with more cleanups etc, it is a nice function).

`CalculatePanelAreas()` was removed from the header as it is only used in the same file as it is declared.

4aron5un:
Thank you and sorry for the changes without further discussion after the PR  was approved due to the conversation was already locked.

AJenbo:
Yeah I understand, normally it wouldn't have been a problem :)

4aron5un:
Even though I know you already understand my design, I still share it here for others to know and maybe get better ideas. :smile:

If the screen width is less than 1280, the panel position will be the same as left/center. (The maximum panel distance is 640, which is equal to the bottom panel width.)

[853x480]
853x480

[800x600]
800x600

[1280x720]
1280x720

[1600x900]
1600x900

My idea is not only to reduce the mouse traveling, but also to the eye traveling, I can see the change of stat at the same time when changing equipment.

But without this change, users still can choose the favorite alignment with the current design.

AJenbo:
Seams like a good option, it won't affect me since I run at 1146x480 and it would take people running at high resolution in to account 👍

Mostly i just didn't want to add more parameters to the ini ☺️

4aron5un:
> Seams like a good option, it won't affect me since I run at 1146x480 and it would take people running at high resolution in to account 
> 
> Mostly i just didn't want to add more parameters to the ini

I got your point and agree with you, too many parameters make things more complicated.
Yes. my goal is trying to make it simple. :smile:
Keep the 3X3 combination just for functional complete.

AJenbo:
Feel free to make a PR that changes things to this new suggestion

AJenbo:
@4aron5un this ended up being critical for us being able to get the Android port out in time, thank you :)

PR #2514, opened 2021-08-06
Disable hardware cursor on Android by default

### LINKED ISSUE #2502: Android hardware cursor not visible when using gamepad
When using the hardware cursor on Android, the mouse cursor is not visible in the inventory, character sheet, or when freely moving the cursor using the right analog stick.

SDL on Android implements hardware cursor using the [PointerIcon](https://developer.android.com/reference/android/view/PointerIcon) class. `PointerIcon` objects can be used to set the default cursor for a `View`, but the SDK doesn't seem to provide any control over the position or visibility of the mouse cursor. From what I can tell, the `PointerIcon` is only visible when a hardware mouse is used to interact with the device.

Incidentally, these `PointerIcon` objects are stored in a hash table to be referenced by an integer ID, but the SDL implementation doesn't provide a JNI function for `SDL_FreeCursor()` to remove them from the hash table. This represents a bit of a memory leak for the application so perhaps hardware cursor on Android should be avoided entirely.

---

SDL_CreateColorCursor:
https://github.com/diasurgical/devilutionX/blob/9aee0581374858991705ecae373c20392f93646f/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1501-L1515

SDL_SetCursor:
https://github.com/diasurgical/devilutionX/blob/9aee0581374858991705ecae373c20392f93646f/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1520-L1532

glebm:
Fixes #2502

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -160,9 +160,14 @@ void SaveIni()
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 bool HardwareCursorDefault()
 {
+#ifdef __ANDROID__
+	// See https://github.com/diasurgical/devilutionX/issues/2502
+	return false;
+#else
 	SDL_version v;
 	SDL_GetVersion(&v);
 	return SDL_VERSIONNUM(v.major, v.minor, v.patch) >= SDL_VERSIONNUM(2, 0, 12);
+#endif
 }
 #endif
 


AJenbo (approved):
Thanks for handling these :)

PR #2516, opened 2021-08-06
Handle some clang-tidy issues

AJenbo:


Source/codec.cpp
```diff
@@ -24,22 +25,22 @@ struct CodecSignature {
...
 void CodecInitKey(const char *pszPassword)
 {
-	char key[136]; // last 64 bytes are the SHA1
+	byte key[136]; // last 64 bytes are the SHA1
 	uint32_t randState = 0x7058;
-	for (char &notch : key) {
+	for (auto &notch : key) {
 		randState = randState * 214013 + 2531011;
-		notch = randState >> 16; // Downcasting to char keeps the 2 least-significant bytes
+		notch = static_cast<byte>(randState >> 16); // Downcasting to char keeps the 2 least-significant bytes
```
ephphatha, line 32: This comment needs to be updated :D

Source/diablo.cpp
```diff
@@ -701,7 +701,7 @@ void GameEventHandler(uint32_t uMsg, int32_t wParam, int32_t lParam)
 		ReleaseKey(wParam);
 		return;
 	case DVL_WM_CHAR:
-		PressChar(wParam);
+		PressChar((char)wParam);
```
ephphatha, line 704:
```suggestion
		PressChar(static_cast<char>(wParam));
```

c++ style casts

AJenbo: I didn't investigate if this is the proper type or place (just pushed it back from where it was definitely wrong), so keeping it as C style to make it clear that it needs more work

Source/codec.cpp
AJenbo: Thanks for the review 🙂

```suggestion
		notch = static_cast<byte>(randState >> 16); // Downcasting to byte keeps the 2 least-significant bytes
```

ephphatha: I should've elaborated, the comment doesn't clearly  describe what the code is actually doing and doesn't explain why it's doing the operation. This is generating a fixed key at runtime from a LCG engine using the low byte of the high word. The downcast isn't saving two bytes at a time.

The key also doesn't need to be generated at random. I'd just get rid of the comment for now, can replace the entire loop with a hardcoded key later unless there's a particular need to keep generating it at run time.

AJenbo: Yeah, I don't really understand the comment so hard for me to adjust it :D

I think your follow up takes care of it, so will just merge as is now.

AJenbo:
image

image

PR #2517, opened 2021-08-06
Remove dead code

AJenbo:


qndel:
What changed in itemdat.cpp?

AJenbo:
Imaxvalue was unused, thee row was identical to the previous one as well

PR #2518, opened 2021-08-07
Build and include translations

AJenbo:
Note this does not yet include translations with the non-Android ports, if you maintain one of the ports or know how to do so then please feel free to help out :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -615,6 +615,14 @@ if (Gettext_FOUND)
       list(APPEND VITA_TRANSLATIONS_LIST "FILE" "${CMAKE_CURRENT_BINARY_DIR}/${lang}.gmo" "${lang}.gmo")
     endif()
   endforeach(lang)
+
+  if(ANDROID)
+    add_custom_target(copy_translations ALL
+      COMMAND ${CMAKE_COMMAND} -E copy ${devilutionx_TRANSLATIONS} ${DevilutionX_SOURCE_DIR}/android-project/app/src/main/assets
+      DEPENDS ${devilutionx_TRANSLATIONS})
+
+    add_dependencies(${BIN_TARGET} copy_translations)
+  endif()
 endif()
 
 target_include_directories(libdevilutionx PUBLIC
@@ -947,7 +955,7 @@ if(NINTENDO_3DS)
 
   add_custom_target(romfs_directory
     COMMAND ${CMAKE_COMMAND} -E make_directory ${APP_ROMFS})
-    
+
   add_custom_target(romfs_files
     COMMAND ${CMAKE_COMMAND} -E copy ${APP_ROMFS_FILES} ${APP_ROMFS}
     DEPENDS romfs_directory)
@@ -955,7 +963,7 @@ if(NINTENDO_3DS)
   include(Tools3DS)
   add_3dsx_target(${BIN_TARGET})
   add_cia_target(${BIN_TARGET} ${APP_RSF} ${APP_BANNER} ${APP_AUDIO})
-  
+
   get_filename_component(APP_TARGET_PREFIX ${BIN_TARGET} NAME_WE)
   add_dependencies(${APP_TARGET_PREFIX}_3dsx romfs_files)
   add_dependencies(${APP_TARGET_PREFIX}_cia romfs_files)
--- a/android-project/.gitignore
+++ b/android-project/.gitignore
@@ -12,3 +12,4 @@ build/
 /local.properties
 
 /app/src/main/assets/CharisSILB.ttf
+/app/src/main/assets/*.gmo


.gitignore
```diff
@@ -446,3 +446,4 @@ Brewfile.lock.json
 # GPerf heap profile dumps
 *.heap
 *.mo
+/android-project/app/src/main/assets/*.gmo
```
StephenCWills, line 449: The rule to ignore the TTF font file is in `android-project/.gitignore`.
https://github.com/diasurgical/devilutionX/blob/cfa3c8fa4e08e19a1976fd64cbc1472762e9d150/android-project/.gitignore#L14

CMakeLists.txt
```diff
@@ -606,10 +605,18 @@ if(GPERF)
...
-if(USE_GETTEXT)
-  find_package(Gettext)
+find_package(Gettext)
+if (Gettext_FOUND)
+  set(PO_COUNT 0)
   foreach(lang bg da de es fr hr it ko_KR pt_BR ru sv zh_CN)
     GETTEXT_PROCESS_PO_FILES(${lang} ALL PO_FILES ${CMAKE_CURRENT_LIST_DIR}/Translations/${lang}.po)
+
+    if(ANDROID)
+      MATH(EXPR PO_COUNT "${PO_COUNT}+1")
```
StephenCWills, line 615:
```suggestion
      math(EXPR PO_COUNT "${PO_COUNT}+1")
```

Source/utils/language.cpp
```diff
@@ -237,21 +237,21 @@ const char *LanguageMetadata(const char *key)
 
 bool HasTranslation(const std::string &locale)
 {
-	std::string gmoPath = paths::LangPath() + locale + ".gmo";
+	std::string gmoPath = paths::LangPath() + locale + ".mo";
 	if (FileExists(gmoPath.c_str()))
```
StephenCWills, line 241:
```suggestion
	std::string moPath = paths::LangPath() + locale + ".mo";
	if (FileExists(moPath.c_str()))
```

Source/utils/language.cpp
```diff
@@ -237,21 +237,21 @@ const char *LanguageMetadata(const char *key)
...
 bool HasTranslation(const std::string &locale)
 {
-	std::string gmoPath = paths::LangPath() + locale + ".gmo";
+	std::string gmoPath = paths::LangPath() + locale + ".mo";
 	if (FileExists(gmoPath.c_str()))
 		return true;
 
-	std::string moPath = paths::LangPath() + locale + ".mo";
+	std::string moPath = paths::LangPath() + locale + ".gmo";
 	return FileExists(moPath.c_str());
```
StephenCWills, line 245:
```suggestion
	std::string gmoPath = paths::LangPath() + locale + ".gmo";
	return FileExists(gmoPath.c_str());
```

StephenCWills (changes requested):

CMakeLists.txt
```diff
@@ -606,10 +605,18 @@ if(GPERF)
...
+if (Gettext_FOUND)
+  set(PO_COUNT 0)
   foreach(lang bg da de es fr hr it ko_KR pt_BR ru sv zh_CN zh_TW)
     GETTEXT_PROCESS_PO_FILES(${lang} ALL PO_FILES ${CMAKE_CURRENT_LIST_DIR}/Translations/${lang}.po)
+
+    if(ANDROID)
+      math(EXPR PO_COUNT "${PO_COUNT}+1")
+      add_custom_target(copy_${lang} ALL COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_BINARY_DIR}/${lang}.gmo" "${DevilutionX_SOURCE_DIR}/android-project/app/src/main/assets/${lang}.gmo")
+      add_dependencies(${BIN_TARGET} copy_${lang})
+      add_dependencies(copy_${lang} "pofiles_${PO_COUNT}")
```
glebm, line 618: What's this for? As far as I can tell this `pofiles_${PO_COUNT}` doesn't seem to do anything useful

CMakeLists.txt
```diff
@@ -606,10 +605,18 @@ if(GPERF)
...
-  find_package(Gettext)
+find_package(Gettext)
+if (Gettext_FOUND)
+  set(PO_COUNT 0)
   foreach(lang bg da de es fr hr it ko_KR pt_BR ru sv zh_CN zh_TW)
     GETTEXT_PROCESS_PO_FILES(${lang} ALL PO_FILES ${CMAKE_CURRENT_LIST_DIR}/Translations/${lang}.po)
+
+    if(ANDROID)
+      math(EXPR PO_COUNT "${PO_COUNT}+1")
+      add_custom_target(copy_${lang} ALL COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_BINARY_DIR}/${lang}.gmo" "${DevilutionX_SOURCE_DIR}/android-project/app/src/main/assets/${lang}.gmo")
```
glebm, line 616: Here is what I think is a more correct example of how to do this from StackOverflow:

```cmake
add_custom_target(copy-test-makefile ALL DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/input.txt)
add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/input.txt
                   COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/input.txt
                                                    ${CMAKE_CURRENT_BINARY_DIR}/input.txt
                   DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/input.txt)
```

(https://stackoverflow.com/a/55184207/181228)

CMakeLists.txt
AJenbo: Android studio /grade doesn't build the translation targets, so this is adding them as dependencies so that they get build with the project

glebm: It doesn't though, `"pofiles_${PO_COUNT}"` is not a target that exists. I don't think you'll need after fixing the commands per the comment below. What I think happens here is a rebuild triggers every time because it depends on this non-existing target.

AJenbo: GETTEXT_PROCESS_PO_FILES generates the pofiles_# targets, I needed the dependency to even generate the gmo files even before I started looking in to relocating them to the assets folder.
But I'll give removing it a shot just in case 🤷

isage: GETTEXT_PROCESS_PO_FILES also generates/updates `pofiles` target with dep set to all `pofiles_N`
wouldn't it be easier/better to just copy all .gmo files at once (via custom command, that depends on `pofiles` target)

StephenCWills: I like @isage's idea. How about this?

```cmake
find_package(Gettext)
if (Gettext_FOUND)
  foreach(lang bg da de es fr hr it ko_KR pt_BR ru sv zh_CN zh_TW)
    GETTEXT_PROCESS_PO_FILES(${lang} ALL PO_FILES ${CMAKE_CURRENT_LIST_DIR}/Translations/${lang}.po)
    list(APPEND libdevilutionx_TRANSLATIONS ${CMAKE_CURRENT_BINARY_DIR}/${lang}.gmo)
  endforeach(lang)
  
  if(ANDROID)
    add_custom_target(copy_translations ALL
      COMMAND ${CMAKE_COMMAND} -E copy ${libdevilutionx_TRANSLATIONS} "${DevilutionX_SOURCE_DIR}/android-project/app/src/main/assets"
      DEPENDS ${libdevilutionx_TRANSLATIONS})
      
    add_dependencies(${BIN_TARGET} copy_translations)
  endif()
endif()
```

AJenbo: > GETTEXT_PROCESS_PO_FILES also generates/updates `pofiles` target with dep set to all `pofiles_N`
> wouldn't it be easier/better to just copy all .gmo files at once (via custom command, that depends on `pofiles` target)

I would think so too from looking at FindGettext.cmake, but this is what I get if I set the target to `pofiles`:
`> Unexpected native build target pofiles. Valid values are: libdevilutionx, StormLib, pofiles_8, fmt, pofiles_7, pofiles_9, pofiles_4, pofiles_3, SDL2, pofiles_6, pofiles_5, uninstall, devilutionx, PKWare, pofiles_2, pofiles_1, SDL2_ttf, hidapi, pofiles_10, pofiles_11, SDL_audiolib, pofiles_12, pofiles_13, smacker, sodium, freetype, SDL2main`

AJenbo: > How about this?

Works nicely, lets go with that. Boy do I feel useless now :unamused:

PR #2520, opened 2021-08-07
Search Spell Improvements

vladtepesch:
Some Improvement of Devilution reduces the usefullness of the Search spell:
 - objects visibility due to increased screen size vs the automap range of the spell
 - Diablo 2 style item display by CTRL-key vs item highlighting 

I tried to add some features to make the spell more attractive to be used:

Also unopened container objects are highlighted
Also unopened container items are shown own map as darker blue tile marker.
The automap display range is increased by one every 3 spell levels.

screen00

### DIFF
--- a/Source/objects.h
+++ b/Source/objects.h
@@ -137,8 +137,8 @@ struct ObjectStruct {
 	 */
 	bool IsChest() const
 	{
-		return IsAnyOf(_otype, _object_id::OBJ_CHEST1,  _object_id::OBJ_CHEST2,  _object_id::OBJ_CHEST3,
-		                       _object_id::OBJ_TCHEST1, _object_id::OBJ_TCHEST2, _object_id::OBJ_TCHEST3, _object_id::OBJ_SIGNCHEST);
+		return IsAnyOf(_otype, _object_id::OBJ_CHEST1, _object_id::OBJ_CHEST2, _object_id::OBJ_CHEST3,
+		    _object_id::OBJ_TCHEST1, _object_id::OBJ_TCHEST2, _object_id::OBJ_TCHEST3, _object_id::OBJ_SIGNCHEST);
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


Source/objects.h
```diff
@@ -130,6 +130,38 @@ struct ObjectStruct {
...
+	 */
+	bool IsBookContainer() const
+	{
+		return IsAnyOf(_otype, _object_id::OBJ_BOOKSHELF, _object_id::OBJ_BOOKCASEL, _object_id::OBJ_BOOKCASER,
+				               _object_id::OBJ_BOOKSTAND, _object_id::OBJ_SKELBOOK);
+	}
+
+	bool isAnyContainer() const
+	{
+		return     IsChest()
```
qndel, line 156: wtf is this formatting lol

AJenbo: What @qndel is trying to say is that you should run `clang-format` on the code before committing ;)

vladtepesch: sorry - i pushed a commit with clang-formatted sources

qndel:
I wouldn't touch search at all - for people that want to shit on it there's alt highlighting, for people that don't, there's 100% vanilla search, I'm fine with it being that way

vladtepesch:
> 
> 
> I wouldn't touch search at all - for people that want to shit on it there's alt highlighting, for people that don't, there's 100% vanilla search, I'm fine with it being that way

then maybe an additional game option?
The whole point of the improvement is that currently with the alternative highlighting there is no need for the current implementation of the spell. To correct the balancing I added some additional usefulnes.

Regarding "vanilla": the spell expierenced significant changes during some hellfire patches at the time (at beginning there was no map highlight) - so what exactly does "vanilla" mean? The whole Search spell is a convinient feature. 

I would be with you if we would talk about changing fundamental spells that influences the game play.

vladtepesch:
for some reason the appveyour stepp crashes with `internal server error`. Seems to be unrelated to the the commit

vladtepesch:
I added the gameoption so it should be fine, isn't it?
I mean the alt-Highlighting is definitevly more changing and does not even has an option

vladtepesch:
if not then maybe close the PR, so that you does not got notified everytime i push to my branch after a rebase

AJenbo:
>  I mean the alt-Highlighting is definitevly more changing and does not even has an option

It still needs a bit of tweaking.

julealgon:
I like the idea of having the automap range modifiable. Would be nice to have this as a mechanic even if it is not surfaced in the game right now, for modding purposes.

Actually, I wonder if it wouldn't make even more sense to have automap range be based on light radius.

PR #2524, opened 2021-08-07
Refactor Diablo-SHA to attempt to document behaviour through code

ephphatha:
Nerd sniped by #2516. If you do want to build the key at runtime the following initial state gets the right sequence:
```cpp
std::linear_congruential_engine<uint32_t, 214013, 2531011, 0> engine(3193970784);
for (auto &notch: key)
	key = static_cast<byte>(engine() >> 16);
```

(This will need rebasing after #2516 is merged to use the correct type for the `key` array)

### DIFF
--- a/Source/codec.cpp
+++ b/Source/codec.cpp
@@ -22,8 +22,6 @@ struct CodecSignature {
 	uint16_t unused;
 };
 
-#define BlockSize 64
-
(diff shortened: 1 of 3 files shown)


AJenbo:
All yours now :)

AJenbo (approved):

PR #2526, opened 2021-08-07
Timedemo: Create reference save game ...

obligaron:
… and allow comparing of reference save game with actual timedemo save game
Extract from #2490 cause it can also help without tests if you want to ensure that your run is identically and the timedemo still works as expected.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -853,6 +853,7 @@ void DiabloParseFlags(int argc, char **argv)
 	bool timedemo = false;
 	int demoNumber = -1;
 	int recordNumber = -1;
+	bool createDemoReference = false;
 	for (int i = 1; i < argc; i++) {
 		const string_view arg = argv[i];
 		if (arg == "-h" || arg == "--help") {
@@ -893,6 +894,8 @@ void DiabloParseFlags(int argc, char **argv)
 				diablo_quit(0);
 			}
 			recordNumber = SDL_atoi(argv[++i]);
+		} else if (strcasecmp("--create-reference", argv[i]) == 0) {
+			createDemoReference = true;
 		} else if (arg == "-n") {
 			gbShowIntro = false;
 		} else if (arg == "-f") {
@@ -934,7 +937,7 @@ void DiabloParseFlags(int argc, char **argv)
 	if (demoNumber != -1)
 		demo::InitPlayBack(demoNumber, timedemo);
 	if (recordNumber != -1)
-		demo::InitRecording(recordNumber);
+		demo::InitRecording(recordNumber, createDemoReference);
 }
 
 void DiabloInitScreen()
--- a/Source/engine/demomode.cpp
+++ b/Source/engine/demomode.cpp
@@ -32,6 +32,7 @@ struct demoMsg {
 int DemoNumber = -1;
 bool Timedemo = false;
 int RecordNumber = -1;
+bool CreateDemoReference = false;
 
 std::ofstream DemoRecording;
 std::deque<demoMsg> Demo_Message_Queue;
@@ -132,9 +133,10 @@ void InitPlayBack(int demoNumber, bool timedemo)
 		diablo_quit(1);
 	}
 }
-void InitRecording(int recordNumber)
+void InitRecording(int recordNumber, bool createDemoReference)
 {
 	RecordNumber = recordNumber;
+	CreateDemoReference = createDemoReference;
 }
 void OverrideOptions()
 {
@@ -278,8 +280,11 @@ void NotifyGameLoopEnd()
 {
 	if (IsRecording()) {
 		DemoRecording.close();
+		if (CreateDemoReference)
+			pfile_write_hero_demo(RecordNumber);
 
 		RecordNumber = -1;
+		CreateDemoReference = false;
 	}
 
 	if (IsRunning()) {
@@ -287,6 +292,19 @@ void NotifyGameLoopEnd()
 		SDL_Log("%d frames, %.2f seconds: %.1f fps", LogicTick, secounds, LogicTick / secounds);
 		gbRunGameResult = false;
 		gbRunGame = false;
+
+		HeroCompareResult compareResult = pfile_compare_hero_demo(DemoNumber);
+		switch (compareResult) {
+		case HeroCompareResult::ReferenceNotFound:
+			SDL_Log("Timedemo: No final comparision cause reference is not present.");
+			break;
+		case HeroCompareResult::Same:
+			SDL_Log("Timedemo: Same outcome as inital run. :)");
+			break;
+		case HeroCompareResult::Difference:
+			SDL_Log("Timedemo: Different outcome then inital run. ;(");
+			break;
+		}
 	}
 }
 
--- a/Source/engine/demomode.h
+++ b/Source/engine/demomode.h
@@ -12,7 +12,7 @@ namespace devilution {
 namespace demo {
 
 void InitPlayBack(int demoNumber, bool timedemo);
-void InitRecording(int recordNumber);
+void InitRecording(int recordNumber, bool createDemoReference);
 void OverrideOptions();
 
 bool IsRunning();
--- a/Source/pfile.cpp
+++ b/Source/pfile.cpp
@@ -38,12 +38,16 @@ MpqWriter StashWriter;
 /** List of character names for the character selection screen. */
 char hero_names[MAX_CHARACTERS][PLR_NAME_LEN];
 
+std::string savePrefix;
+
 std::string GetSavePath(uint32_t saveNum)
 {
 	std::string path = paths::PrefPath();
 	const char *ext = ".sv";
 	if (gbIsHellfire)
 		ext = ".hsv";
+
+	path.append(savePrefix);
 
 	if (gbIsSpawn) {
 		if (!gbIsMultiplayer) {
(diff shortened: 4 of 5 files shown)


AJenbo:
I think if not only in test, then this should definitely  be debug build only

obligaron:
> 
> 
> I think if not only in test, then this should definitely be debug build only

## Testing
Release and debug configuration can have different bugs. So this is needed in release cause it should also be possible to run the tests with release configuration.

## Product

### Comparison

The comparison itself doesn't harm. If the reference save game is not present nothing happens, only one Log-Entry.

### Reference save game creation

The creation of reference save games could be annoying, cause it creates (in the future) often not needed files. So I think a separate toggle would be good (something like `--create-reference`).

### Present

This feature helps a lot to verify if the replay is really stable and I think early adapters could also be interested in this diagnostic.

### Future

Currently issues are often reported with a attached save game. This helps a lot but we still need to click around to often reproduce the issue and sometimes we simply don't click the same way the issue author did. This can mean that we sometimes can't reproduce the issue. 
With a attached timedemo this can't happen. But even with this, can you verify that you really got the same results as the one that recorded the demo file (the issue writer)? He could used a different version/commit or the timedemo is broken in someway. In this case this pr/feature helps.
We could advice to add a savegame, a timedemo and the verification save game. With this we can ensure that we got the same result as the issue author.

That's why I would like to keep the logic in release, but behind a toggle (`--create-reference`). What do you think?

AJenbo:
I agree with all of it, but reference save and comparison is a developer tool and should only be in the debug version. Demo recording and playback should be in product.

Maybe it would help of some of this code lived more isolated in a seperat file.

obligaron:
> Maybe it would help of some of this code lived more isolated in a seperat file.

I tried this with the last two commits. So take a look. 🙂 
But I'm not that happy to couple refactoring with enhancement in one PR. Cause you told me otherwise. 😜

Source/engine/demomode.h
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Contains most of the the demomode specific logic
+ */
+#pragma once
+
+#include "miniwin/miniwin.h"
+
+namespace devilution {
+
+class DemoMode {
```
glebm, line 12: Rather than a class with all-static methods, make this a `namespace demo {`.

obligaron: I wasn't sure what is the best way, so I asked @AJenbo but we both were uncertain. So we started with a `class`.
But I'm also happy with a namespace. Fixed. Take a look. 🙂

glebm: Thanks!

Source/engine/demomode.cpp
```diff
@@ -0,0 +1,312 @@
...
+}
+
+} // namespace
+
+namespace demo {
+
+int demoNumber = -1;
+bool timedemo = false;
+int recordNumber = -1;
+bool createDemoReference = false;
```
glebm, line 130: Seems like these should be in the anonymous namespace above?

obligaron: They could cause they are not exported.
But I couldn't figure out how to write:
`demo::demoNumber = demoNumber;`
with a anonymous namespace, cause both the global variable and the parameter have the same name.

Alternative I could rename the function parameter (to `demoNumb` or something else) or the global variable (to `gDemoNumber` or something else).

Any preference? 🙂

Source/diablo.cpp
```diff
@@ -933,15 +924,13 @@ void DiabloParseFlags(int argc, char **argv)
 		} else if (strcasecmp("--save-dir", argv[i]) == 0) {
 			paths::SetPrefPath(argv[++i]);
 		} else if (strcasecmp("--demo", argv[i]) == 0) {
-			demoMode = true;
-			if (!LoadDemoMessages(SDL_atoi(argv[++i]))) {
-				SDL_Log("Unable to load demo file");
-				diablo_quit(1);
-			}
+			demoNumber = SDL_atoi(argv[++i]);
```
AJenbo, line 927:
```suggestion
			demoNumber = SDL_atoi(argv[++i]);
			gbShowIntro = false;
```

obligaron: Added. 😉

offtopic: the intro skipping could be an ini option or? Cause I think this could be something that could be toggled with ingame settings in the future.

Source/diablo.cpp
```diff
@@ -933,15 +924,14 @@ void DiabloParseFlags(int argc, char **argv)
...
-			}
+			demoNumber = SDL_atoi(argv[++i]);
+			gbShowIntro = false;
 		} else if (strcasecmp("--timedemo", argv[i]) == 0) {
 			timedemo = true;
 		} else if (strcasecmp("--record", argv[i]) == 0) {
-			recordDemo = SDL_atoi(argv[++i]);
+			recordNumber = SDL_atoi(argv[++i]);
+		} else if (strcasecmp("--create-reference", argv[i]) == 0) {
+			createDemoReference = true;
```
AJenbo, line 934: I still think we should find a different an probably more automated way to achieve this, but unless I think of something by the end of this review then we can keep this for now.

Source/diablo.cpp
AJenbo: Maybe, not sure. But i noticed that it's confuses users that the video plays during playback and trying to skip it will case the replay to abort.

Source/diablo.cpp
AJenbo: How about simply always doing this for debug builds?

Source/miniwin/misc_msg.cpp
```diff
@@ -592,192 +590,15 @@ bool FetchMessage_Real(tagMSG *lpMsg)
...
-
 bool FetchMessage(tagMSG *lpMsg)
 {
-	if (demoMode) {
-		return DemoMessage(lpMsg);
-	}
+	if (demo::IsRunning())
+		return demo::FetchMessage(lpMsg);
 
 	bool available = FetchMessage_Real(lpMsg);
```
AJenbo, line 598:
```suggestion
	bool available = demo::IsRunning() ? demo::FetchMessage(lpMsg) : FetchMessage_Real(lpMsg);
```

Source/pfile.cpp
```diff
@@ -243,6 +247,42 @@ bool ArchiveContainsGame(HANDLE hsArchive)
 	return IsHeaderValid(hdr);
 }
 
+bool CompareSaves(std::string actualSavePath, std::string referenceSavePath)
```
AJenbo, line 250: I'm having a hard time seeing why this should live in pfile. Also instead of using a full blown save for comparison I think it would be better to just have a hash generated by `picosha2`

Source/pfile.cpp
```diff
@@ -34,13 +34,17 @@ namespace {
...
+std::string savePrefix;
+
 std::string GetSavePath(uint32_t saveNum)
 {
 	std::string path = paths::PrefPath();
 	const char *ext = ".sv";
 	if (gbIsHellfire)
 		ext = ".hsv";
 
+	path.append(savePrefix);
```
AJenbo, line 46: This is a bit of a hack. If instead we change it so that the demo file is independent of the original save then we could create a normal save game and compare that to our known hash with out it affecting the demo (I presume avoiding overwriting the demo-start-point-save-game is the reason for this).

Source/engine/demomode.cpp
AJenbo: They are globals, style guide says globals start with a capital :)
```suggestion
int DemoNumber = -1;
bool Timedemo = false;
int RecordNumber = -1;
bool CreateDemoReference = false;
```

Source/pfile.cpp
obligaron: This are two topics.

### Binary format on disk:
In my first version I simply tried to write the save game and compare the written file with the original. Sadly this doesn't worked. So I thought someone adds data was always changes (random data or a timestamp). Then I changed save game creation that less and less data was written. But at the end it seams that stormlib or the codec functions add random bytes.
That's why I ended up with comparing the uncompressed content of the mpq. And for that I need functions like `ReadArchive` or `GetPermSaveNames` that all live in `pfile.cpp`. 🙂 
I think the same problem also applies for hash generation.

### Hash comparsion
In the future I think this is a good thing. Cause it's decreases the needed size massively. Currently I would stick with the save game, cause it allows better diagnostic what changed (cause you can compare content) and I'm currently not sure if I compare enough data (are some files in the mpq missing?).

Source/pfile.cpp
obligaron: Your assumption is correct. 🙂 
But if you write a new save game, you aren't allowed to override the existing (regardless if this is the original or the demo starting save game or if you create the savegame only for hashcode generation). So I think this topic stays or I miss something. 😉 
If you want I can take a look to change `pfile_write_hero` so that it get's the filename passed in. Additional question: if yes in this pr or new pr? 😉

Source/miniwin/misc_msg.cpp
obligaron: Fixed 🙂

Source/diablo.cpp
obligaron: If you want I can do this. But I thought a additional parameter doesn't harm anyone.

Source/engine/demomode.cpp
obligaron: Missed this one, as I often see globals in old name. Thank you, fixed. 🙂

AJenbo:
Bibibop, Bobs your uncle. I merged the second half of this PR (the part pertaining just to cleaning up the demo code) and rebased the code related to comparing runs.

I'll come back to evaluating the comparison part and think of alternatives a bit later, probably at the start of next week. Thanks for working on this :)

AJenbo (commented):
Looks good, but maybe we should hide it behind #ifdef BUILD_TESTING so make it more ovious hot to treat it for later refactors of the save system.

Source/pfile.cpp
AJenbo: If you think you will do that in the next PR I'm ok merging as is :)

AJenbo (approved):
Feel free to merge https://github.com/diasurgical/devilutionX/pull/2526#discussion_r889590695 :)

Source/pfile.cpp
obligaron: I will check it in a next pr.
If it's too much work (rabbit hole) I will open a pr that adds `BUILD_TESTING` defines.

PR #2527, opened 2021-08-07
Introduce GetSpellListSelection and remove selection game logic from DrawSpellList

obligaron:
Extract from #2490 cause it also affects normal timedemo.

The rendered frames while recording and playback can differ (for example the recording machine has a better cpu).
That's normal and needed to ensure a smooth playback of a timedemo.
But we can skip some frames or introduce new ones.
Skipping/changes frames shouldn't result in a problem.

But atm we can change some game stats while rendering.
This PR is a example of this problem. The `DrawSpellList` changes the currently selected spell.
If we don't render the frame we don't select the spell => broken timedemo.

Note: This kind of changes decouples UI related game logic from rendering. This _could_ also help if we later introduce different renderer (for example hardware renderer).


test/control_test.cpp
```diff
@@ -4,20 +4,6 @@
...
-	pSpell = SPL_FIREBOLT;
-	pSplType = RSPLTYPE_CHARGES;
-	auto &myPlayer = Players[MyPlayerId];
-	SetSpell();
-	EXPECT_EQ(spselflag, false);
-	EXPECT_EQ(myPlayer._pRSpell, SPL_FIREBOLT);
-	EXPECT_EQ(myPlayer._pRSplType, RSPLTYPE_CHARGES);
-	EXPECT_EQ(pnumlines, 0);
-	EXPECT_EQ(force_redraw, 255);
-}
```
AJenbo, line 19: Can we replace this test with something equivalent instead of just removing it :thinking:

Source/control.cpp
```diff
@@ -563,6 +561,67 @@ void RemoveGold(int pnum, int goldIndex)
...
+bool GetSpellListSelection(spell_id &pSpell, spell_type &pSplType)
+{
+	pSpell = spell_id::SPL_INVALID;
+	pSplType = spell_type::RSPLTYPE_INVALID;
+
+	uint64_t mask;
+
+	Point location;
+	int &x = location.x;
+	int &y = location.y;
```
AJenbo, line 573: Why use a `Point`, if the `Point` is never used?

Source/control.cpp
```diff
@@ -563,6 +561,67 @@ void RemoveGold(int pnum, int goldIndex)
...
+	x = PANEL_X + 12 + SPLICONLENGTH * SPLROWICONLS;
+	y = PANEL_Y - 17;
```
AJenbo, line 575: Why use a `Point`, if the `Point` is never used?
```suggestion
	int x = PANEL_X + 12 + SPLICONLENGTH * SPLROWICONLS;
	int y = PANEL_Y - 17;
```

Source/control.cpp
```diff
@@ -563,6 +561,67 @@ void RemoveGold(int pnum, int goldIndex)
 	dropGoldValue = 0;
 }
 
+bool GetSpellListSelection(spell_id &pSpell, spell_type &pSplType)
```
AJenbo, line 564: Does this whole thing replace some other part of the code, or can it? It seams a lot of code for what used to just be `pSpell == SPL_INVALID` :)

Nice that it gets rid of some globals though :+1:

AJenbo (commented):
Looks good, but I have questions :no_mouth:

Source/control.cpp
obligaron: I simply like `Point` struct 😁 
Fixed

test/control_test.cpp
obligaron: I also thought about it.
We would need to set up the the resolution + player spell list + calculate the mouse position for this spell.

But then I saw what we (initially) test and I thought that's currently not worth it:

```
void SetSpell()
{
	spselflag = false;
	if (pSpell == SPL_INVALID) {
		return;
	}

	ClearPanel();

	myPlayer._pRSpell = pSpell;
	myPlayer._pRSplType = pSplType;

	force_redraw = 255;
}
```

One branch and three assignments.

Source/control.cpp
obligaron: It also gets you the valid spells. 😉 

One possibility could be to make a new functions that returns a list of all spells and there rendering positions.
This could be used in both rendering and selection logic.

obligaron: i tried this approach. Take a look at the new commit. 🙂

Source/control.cpp
```diff
@@ -734,12 +634,146 @@ void DrawSpellList(const Surface &out)
...
+	return spellListItems;
+}
+
+bool GetSpellListSelection(spell_id &pSpell, spell_type &pSplType)
+{
+	pSpell = spell_id::SPL_INVALID;
+	pSplType = spell_type::RSPLTYPE_INVALID;
+	auto &myPlayer = Players[MyPlayerId];
+
+	for (auto& spellListItem : GetSpellListItems()) {
```
AJenbo, line 647: clang-format?

obligaron: Fixed 🙂 

Btw:
clang-format is now used in all files or?
Should we run clang-format as validation step in CI? (Fails if clang-format would reformat the code)
Do we want this and if yes, do we have a issue for it?
Cause that would help me to avoid such issues and while reviewing you don't have about format.

AJenbo: Yeah I think it would be ok to require parsing clang-format at this point. No issue for it.

AJenbo (approved):

AJenbo:
Thanks for keeping the commits clean.

PR #2528, opened 2021-08-07
Modify the calculation of panel position

4aron5un:
By default, the auto mode, left/center combination is used, but panel distance is limited.

If the screen width is less than 1280, the panel position will be the same as left/center combination. (The maximum panel distance is 640, which is equal to the bottom panel width)

[853x480]
853x480

[800x600]
800x600

[1280x720]
1280x720

[1600x900]
1600x900

If we prefer other combinations, we can still change this option in the ini file.


AJenbo:
Went with this as THE option, nice work :)
https://github.com/diasurgical/devilutionX/commit/aef812fb5ab9f59a6142f30177ee1a392aa4a8ce

4aron5un:
It's good to see this change. 👍

PR #2531, opened 2021-08-08
some useless code removal

qndel:
CreatePlayer -> CreatePlrItems -> CalcPlrItemVals which recalculates resistances and damage modifier
Same in modify/set stat functions - they call CalcPlrInv and then redo the stuff

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2441,14 +2441,6 @@ void CreatePlayer(int playerId, HeroClass c)
 
 	player._pLevel = 1;
 
-	if (player._pClass == HeroClass::Monk) {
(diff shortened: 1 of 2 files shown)


AJenbo:
How do you feel like adding a test for CreatePlayer :)

Most of what is needed already exists: https://github.com/diasurgical/devilutionX/blob/master/test/writehero_test.cpp#L245

AJenbo:
Looks like a no, so went ahead and did it my self: https://github.com/diasurgical/devilutionX/commit/9217e8c7f07e76626d0f3ea2da2d67360fc54133

AJenbo (approved):

PR #2533, opened 2021-08-08
Fix attribute up (+) button doesn't work in special case

4aron5un:
1) Open quest panel
2) Open character panel by click level up (+) button


AJenbo (approved):
Good catch, also this is why we need helper functions instead of just copying globals every where 😐

4aron5un:
> Good catch, also this is why we need helper functions instead of just copying globals every where 😐

Yeah, I'm with you on that one.

PR #2534, opened 2021-08-08
Automap: Scoped enums and wrappers to minimise bitwise operations

ephphatha:
I ended up implementing the suggestions I made while testing if they'd actually work, so here we go :D

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -58,16 +58,16 @@ struct AutomapTile {
 	/** Additional details about the given tile */
 	enum class Flags : uint8_t {
 		// clang-format off
-		VerticalDoor     = 1 << 0,
-		HorizontalDoor   = 1 << 1,
-		VerticalArch     = 1 << 2,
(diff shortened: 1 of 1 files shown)


Source/automap.cpp
```diff
@@ -35,41 +35,48 @@ enum MapColors : uint8_t {
...
+	enum class Flags : uint8_t {
+		// clang-format off
+		VerticalDoor     = 1 << 0,
+		HorizontalDoor   = 1 << 1,
+		VerticalArch     = 1 << 2,
+		HorizontalArch   = 1 << 3,
+		VerticalGrate    = 1 << 4,
+		HorizontalGrate  = 1 << 5,
+		VerticalObject   = VerticalDoor | VerticalArch | VerticalGrate,
+		HorizontalObject = HorizontalDoor | HorizontalArch | HorizontalGrate,
```
AJenbo, line 68: Passage

AJenbo:
I learned something new, very nice 🙂

PR #2535, opened 2021-08-08
fix armor piercing affix range

qndel:
this is a pretty serious one, devilutionx made armor piercing have 2x larger stat range than the original
(example bashing 8-24, devx 16-48)

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1127,8 +1127,8 @@ void SaveItemAffix(ItemStruct &item, const PLStruct &affix)
 
 	if (!gbIsHellfire) {
 		if (power.type == IPL_TARGAC) {
-			power.param1 = 2 << power.param1;
-			power.param2 = 6 << power.param2;
+			power.param1 = 1 << power.param1;
+			power.param2 = 3 << power.param2;
 		}
 	}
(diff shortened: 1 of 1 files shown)


Source/items.cpp
```diff
@@ -1127,8 +1127,8 @@ void SaveItemAffix(ItemStruct &item, const PLStruct &affix)
 
 	if (!gbIsHellfire) {
 		if (power.type == IPL_TARGAC) {
-			power.param1 = 2 << power.param1;
-			power.param2 = 6 << power.param2;
+			power.param1 = 2 << (power.param1 - 1);
+			power.param2 = 6 << (power.param2 - 1);
```
ephphatha, line 1131:
```suggestion
			power.param1 = 1 << power.param1; // 2**param1
			power.param2 = 3 << power.param2; // 2**param2 + 2**(param2+1)
```
Halving the lhs has the same effect, saves needing to subtract as a separate operation. The comments are to try explain the operation the shifts are accomplishing.

qndel: good catch :D

ephphatha (approved):
Looks good :)

PR #2541, opened 2021-08-08
unify ranged and melee pierce

qndel:
1. Make hit chance displayed in character sheet in vanilla include armor pierce - this gives us an option to somehow check the real value of armor piercing affixes before we get a proper description for them

2. Unify hellfire's melee and ranged pierce.
Hellfire values are pretty low and indicate tiers of % armor that should be ignored
1 = 25%
2 = 50%
3 = 75% armor ignored.

Ofc hellfire didn't do such a thing for ranged, in hellfire ranged pierce even had a wrong sign, so it actually decreased hit chance by these 1/2/3%.

This makes ranged pierce follow the same rules as melee's for hellfire - ofc barbarian's passive 12.5% pierce stays only melee.

### DIFF
--- a/Source/player.h
+++ b/Source/player.h
@@ -416,7 +416,20 @@ struct PlayerStruct {
 	}
 
 	/**
+	 * @brief Return player's melee to hit value, including armor piercing
+	 */
+	int GetMeleePiercingToHit() const
+	{
+		int hper = GetMeleeToHit();
+		//in hellfire armor piercing ignores % of enemy armor instead, no way to include it here
+		if (!gbIsHellfire)
+			hper += _pIEnAc;
+		return hper;
+	}
+
+	/**
 	 * @brief Return player's ranged to hit value
+	 * @param pierceArmor - indicate if player's armor piercing should be included - true vs monsters, false vs players
 	 */
 	int GetRangedToHit() const
 	{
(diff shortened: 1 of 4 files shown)


Source/player.h
```diff
@@ -407,25 +407,33 @@ struct PlayerStruct {
...
+			hper += _pIEnAc;
 		return hper;
 	}
 
 	/**
 	 * @brief Return player's ranged to hit value
+	 * @param pierceArmor - indicate if player's armor piercing should be included - true vs monsters, false vs players
 	 */
-	int GetRangedToHit() const
+	int GetRangedToHit(bool pierceArmor = true) const
```
AJenbo, line 427: Instead of bool could we make a variant:
```cpp
int GetRangedPiercingToHit()
{
	int hper = GetRangedToHit();
	if (!gbIsHellfire)
		hper += _pIEnAc;
	return hper;
}
```

Source/player.h
```diff
@@ -454,6 +462,30 @@ struct PlayerStruct {
...
+	{
+		int tmac = monsterArmor;
+		if (gbIsHellfire && _pIEnAc > 0) {
+			int pIEnAc = _pIEnAc - 1;
+			if (pIEnAc > 0)
+				tmac >>= pIEnAc;
+			else
+				tmac -= tmac / 4;
+		}
+		if (isMelee && _pClass == HeroClass::Barbarian) {
```
AJenbo, line 480: Can we check weapon to determine if its melee?

qndel: what if someone fires an arrow and unequips the bow before the arrow hits? ;)

AJenbo (approved):

PR #2544, opened 2021-08-09
Ensure shift+click and auto-equip properly update readied spell from staff

### LINKED ISSUE #2509: Shift equip/autoequip does not load item spell
- Find a staff with charges
- Do not have any spell ready
- Equip the staff by way of shift-click in the inventory
- Notice that there still isn't a readies spell

julealgon:
This PR fixes a problem that caused auto-equipping staves from the groud or equipping them with shift+click in the inventory to not automatically ready the staff's charge spell when charges are available.

We are now making sure that the same logic that sets the readied spell applies on every equipment change situation by moving the processing logic directly into the command processing method for "equipment change", which is broadcast everytime an item is equipped or updated.

Resolves #2509


AJenbo (approved):
This could almost be approved just for reducing the size of CheckInvPaste

PR #2546, opened 2021-08-09
Auto-detect locale on PSVita.

isage:
Sadly, vitasdk locale() always returns "C".
Luckily, there's not that many supported locales on vita, so we can just provide our own mapping.


AJenbo (approved):
Nice work 🙂

Issue #2550, opened 2021-08-09
Stone Curse spell often last forever 


vladtepesch:
**Important information**

Win10 / DevilutionX based on Commit SHA-1: 5ef79ad1f236c70c5f0d589a4e7e852c954d737d Date: 06.08.2021 13:07:34/ build myself with vs2019

**Describe**
Often observe that stone curse seems to last extremely long on low spell levels.
Sometimes they wake up quite early. As far as I remember from the orignal game, the time was not nearly that long.

**To Reproduce**
1. cast stone curse on creature
2. wait looong time
3. still stare at a statue
4. wonder

**Expected behavior**
according to some online ressources, the effect should end at max spell level after 12s (--> https://diablo.fandom.com/wiki/Stone_Curse)

I can provide a save game but it should be easy to reproduce

vladtepesch:
this looks wrong - allthough I atm do not know where this does come from:
![grafik](https://user-images.githubusercontent.com/1601159/128777484-524dcf91-5403-48bf-bbe4-d872f568d200.png)

PR #2551, opened 2021-08-09
fixes stone curse forever bug caused by negative monster index

vladtepesch:
fixes https://github.com/diasurgical/devilutionX/issues/2550

the negative index was written to `_miVar2` and used later on in missile update in `MI_Stone`

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2557,8 +2557,8 @@ void AddStone(int mi, Point /*src*/, Point dst, int /*midir*/, int8_t /*mienemy*
 			int mid = dMonster[tx][ty];
 			if (mid == 0)
 				continue;
-
-			auto &monster = Monsters[abs(mid) - 1];
+			mid = abs(mid) - 1;
+			auto &monster = Monsters[mid];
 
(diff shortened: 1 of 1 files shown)


StephenCWills (approved):
This fix matches the logic in Devilution.
https://github.com/diasurgical/devilution/blob/da7ea7b593e57cee9d80d2ee13717543a727a6ba/Source/missiles.cpp#L3016-L3017

vladtepesch:
> 
> 
> This fix matches the logic in Devilution.
> https://github.com/diasurgical/devilution/blob/da7ea7b593e57cee9d80d2ee13717543a727a6ba/Source/missiles.cpp#L3016-L3017

i do not understand that you mean with this comment. Also the linked code is wrong. if `mi` is `0` then the index `-1` is calculated and used.

Sadly there are different styles of this expression over the code. 
checking for index 0 and then using `abs(m)+1` is the most consitent.

StephenCWills:
> i do not understand that you mean with this comment.

No problem. I am merely stating that the fix matches the vanilla behavior, as justification for my approval of this PR. The link to the Devilution codebase is my evidence.

PR #2552, opened 2021-08-10
3DS: Detect system language and select the appropriate translation

StephenCWills:
I can only select English, French, Spanish, and Portuguese on my console, but here's the full list of supported languages according to libctru:
* Japanese (ja)
* English (en)
* French (fr)
* German (de)
* Italian (it)
* Spanish (es)
* Simplified Chinese (zh_CN)
* Korean (ko_KR)
* Dutch (nl)
* Portuguese (pt_BR)
* Russian (ru)
* Traditional Chinese (zh_TW)

The language codes I listed aren't provided by the OS, but rather are the ones I selected based on the OS language setting. I did my best to select language codes based on the translations we have so far (e.g., `ko_KR` instead of just `ko`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -507,6 +507,7 @@ if(NINTENDO_3DS)
     Source/platform/ctr/messagebox.cpp
     Source/platform/ctr/random.cpp
     Source/platform/ctr/sockets.cpp
+    Source/platform/ctr/locale.cpp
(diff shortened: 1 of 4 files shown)


AJenbo:
so all we miss is nl and ja :o

AJenbo (approved):
very nice implementation

PR #2554, opened 2021-08-10
refactor 3 versions of FindTransparencyValues in: drlg_l1.cpp, drlg_l2.cpp, drlg_l4.cpp

### LINKED ISSUE #2348: 3 versions of FindTransparencyValues in: drlg_l1.cpp, drlg_l2.cpp, drlg_l4.cpp
I've selected for refactoring 38 lines of code which are duplicated in 3 file(s) ([1](https://github.com/diasurgical/devilutionX/blob/42e2c174acdfc1b0f22e9c4d26a1b9f9c9dde407/Source/drlg_l1.cpp#L1953-L2003), [2](https://github.com/diasurgical/devilutionX/blob/42e2c174acdfc1b0f22e9c4d26a1b9f9c9dde407/Source/drlg_l2.cpp#L2769-L2819), [3](https://github.com/diasurgical/devilutionX/blob/42e2c174acdfc1b0f22e9c4d26a1b9f9c9dde407/Source/drlg_l4.cpp#L1190-L1240)).  Addressing this will make our codebase more maintainable and improve [Better Code Hub](https://bettercodehub.com)'s **Write Code Once** guideline rating! 👍 

Here's the gist of this guideline:
- **Why**❓ 
When code is copied, bugs need to be fixed in multiple places. This is both inefficient and a source of regression bugs.
- **How** 🔧 
The 3 versions of `FindTransparencyValues` are identical except for tile ID being checked in the first line. Change the function to take this id as an argument and move the unified version to gendung.cpp to avoid the duplication of code here.

You can find more info about this guideline in [Building Maintainable Software](http://shop.oreilly.com/product/0636920049159.do). 📖 
 
---- 
ℹ️ To know how many _other_ refactoring candidates need addressing to get a guideline compliant, select some by clicking on the 🔲  next to them. The risk profile below the candidates signals (✅) when it's enough! 🏁 

----
Good luck and happy coding! :shipit: :sparkles: :100:

bcko:
fixes #2348

### DIFF
--- a/Source/gendung.cpp
+++ b/Source/gendung.cpp
@@ -549,4 +549,52 @@ void InitLevels()
 	setlevel = false;
 }
 
+void FindTransparencyValues(int tileID, int i, int j, int x, int y, int d)
+{
+	if (dTransVal[x][y] != 0 || dungeon[i][j] != tileID) {
+		if (d == 1) {
+			dTransVal[x][y] = TransVal;
+			dTransVal[x][y + 1] = TransVal;
+		}
+		if (d == 2) {
+			dTransVal[x + 1][y] = TransVal;
+			dTransVal[x + 1][y + 1] = TransVal;
+		}
+		if (d == 3) {
+			dTransVal[x][y] = TransVal;
+			dTransVal[x + 1][y] = TransVal;
+		}
+		if (d == 4) {
+			dTransVal[x][y + 1] = TransVal;
+			dTransVal[x + 1][y + 1] = TransVal;
+		}
+		if (d == 5) {
+			dTransVal[x + 1][y + 1] = TransVal;
+		}
+		if (d == 6) {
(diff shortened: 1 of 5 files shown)


Source/gendung.cpp
```diff
@@ -549,4 +549,52 @@ void InitLevels()
 	setlevel = false;
 }
 
+void FindTransparencyValues(uint8_t tileID, int i, int j, int x, int y, int d)
```
AJenbo, line 552: Probably it would have better to add the new parameter to the end instead of the start, also tileID isn't very descriptive, I'm guessing that this is the ID of the floor tile for the given tilset so a better name might have been floorID

bcko: opened a new pull request to address this issue #2555. Thank you!

PR #2555, opened 2021-08-10
fix parameter name and order in FindTransparencyValues

bcko:
- related to #2348 , #2554
- addresses https://github.com/diasurgical/devilutionX/pull/2554#discussion_r685753643

### DIFF
--- a/Source/drlg_l1.cpp
+++ b/Source/drlg_l1.cpp
@@ -1955,12 +1955,12 @@ void FillChambers()
 void FloodTransparancyValues()
 {
 	int yy = 16;
-	uint8_t tileID = 13;
+	uint8_t floorID = 13;
 	for (int j = 0; j < DMAXY; j++) {
 		int xx = 16;
 		for (int i = 0; i < DMAXX; i++) {
-			if (dungeon[i][j] == tileID && dTransVal[xx][yy] == 0) {
-				FindTransparencyValues(tileID, i, j, xx, yy, 0);
+			if (dungeon[i][j] == floorID && dTransVal[xx][yy] == 0) {
+				FindTransparencyValues(i, j, xx, yy, 0, floorID);
 				TransVal++;
 			}
 			xx += 2;
--- a/Source/drlg_l1.cpp
+++ b/Source/drlg_l1.cpp
@@ -1952,23 +1952,6 @@ void FillChambers()
 	}
 }
 
-void FloodTransparancyValues()
-{
-	int yy = 16;
-	uint8_t floorID = 13;
-	for (int j = 0; j < DMAXY; j++) {
-		int xx = 16;
-		for (int i = 0; i < DMAXX; i++) {
-			if (dungeon[i][j] == floorID && dTransVal[xx][yy] == 0) {
-				FindTransparencyValues(i, j, xx, yy, 0, floorID);
-				TransVal++;
-			}
-			xx += 2;
-		}
-		yy += 2;
-	}
-}
-
 void FixTransparency()
 {
 	int yy = 16;
@@ -2201,6 +2184,7 @@ void GenerateLevel(lvl_entry entry)
 	}
 
 	bool doneflag;
+	uint8_t floorID = 13;
 	do {
 		DRLG_InitTrans();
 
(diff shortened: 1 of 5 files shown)


ephphatha:
You've pretty much got the code to the point that all the `FloodTransparancyValues()` functions can be combined and take floorId as a parameter to that function :). That'd make this code significantly less repetitive if you felt like making that change too.

bcko:
> You've pretty much got the code to the point that all the `FloodTransparancyValues()` functions can be combined and take floorId as a parameter to that function :). That'd make this code significantly less repetitive if you felt like making that change too.

Sounds great! I will update the pull request to combine `FloodTransparancyValues()`. Thank you @ephphatha !

bcko:
Updated the pull request to refactor FloodTransparencyValues().

Source/gendung.h
```diff
@@ -248,5 +248,6 @@ void DRLG_Init_Globals();
 bool SkipThemeRoom(int x, int y);
 void InitLevels();
 void FindTransparencyValues(int i, int j, int x, int y, int d, uint8_t floorID);
```
ephphatha, line 250: This is no longer called from outside gendung.h so no longer needs to be in the header.

Source/drlg_l1.cpp
```diff
@@ -2201,6 +2184,7 @@ void GenerateLevel(lvl_entry entry)
 	}
 
 	bool doneflag;
+	uint8_t floorID = 13;
```
AJenbo, line 2187: Not much point in having a local variable for this when it just gets pass to another function

Source/gendung.cpp
```diff
@@ -597,4 +597,21 @@ void FindTransparencyValues(int i, int j, int x, int y, int d, uint8_t floorID)
 	FindTransparencyValues(i + 1, j + 1, x + 2, y + 2, 8, floorID);
 }
```
ephphatha, line 598: `FindTransparencyValues` can be moved into the anonymous namespace at the start of gendung.cpp now that it's only called from one version of `FloodTransparencyValues` :D

Source/drlg_l4.cpp
```diff
@@ -1323,6 +1306,7 @@ void GenerateLevel(lvl_entry entry)
 	constexpr int Minarea = 173;
 	int ar;
 	bool doneflag;
+	uint8_t floorID = 6;
```
ephphatha:
```suggestion
	constexpr uint8_t HellFloorID = 6;
```
This variable is not modified so can be a constexpr, also suggested a more descriptive name to differentiate the different uses of the resulting function.

Source/drlg_l2.cpp
```diff
@@ -2921,6 +2904,7 @@ void FixDoors()
 void GenerateLevel(lvl_entry entry)
 {
 	bool doneflag = false;
+	uint8_t floorID = 3;
```
ephphatha:
```suggestion
	constexpr uint8_t CatacombsFloorID = 3;
```
Same situation as the previous comment :)

AJenbo (approved):

Source/drlg_l1.cpp
bcko: Updated the pull request to remove local variable floorID in all three places. Thank you @AJenbo !

ephphatha:
```suggestion
	constexpr uint8_t CatacombsFloorID = 13;
```
Last one. Recent changes for globals/constants have been moving to PascalCase as well, so trying to keep that convention where possible.

ephphatha (commented):
Looks good, there's a couple of cleanups made possible now that you've removed the duplicates :D

Source/drlg_l1.cpp
ephphatha: I was too slow 😅 . You can ignore that part of my comments.

Source/drlg_l4.cpp
AJenbo: I think this is used in multiple places so maybe it belongs in a more general cleanup

bcko:
@ephphatha updated the pull request to address `FindTransparencyValues()` in `gendung.h` and `gendung.cpp`. Thank you!

AJenbo (approved):

ephphatha (approved):

AJenbo:
@bcko looks like you need to login to CircleCi for it to be able to run the tests on your PRs

bcko:
> @bcko looks like you need to login to CircleCi for it to be able to run the tests on your PRs

Thank you for letting me know! @AJenbo I just logged into CircleCI

image

Issue #1666, opened 2021-04-24
Cleanup manual value claming


AJenbo:
The code is littered with conditions that check that values do not go over or under the min/max range.

This is relatively easy but tedious to clean up, but doing so greatly helps the readability of the code.

An example of such cleanups in inv.cpp and a couple of other files:
https://github.com/diasurgical/devilutionX/pull/1665

mpanneton:
Hi I'm available to clean up the code. Before I start I would like to know if there is a specific branch for this issue or should I create my own?

AJenbo:
You should create your own, please send regular pull requests where you clean up one function/file at a time to make require quick and easy.

PR #2556, opened 2021-08-10
cleanup manual value clamping: spells.cpp GetManaAmount()

bcko:
related to #1666 

cleanup manual value clamping in `spells.cpp` `GetManaAmount()`
https://github.com/diasurgical/devilutionX/blob/bcd60ea0b44a9a2588746c749f7a101b1bf6621a/Source/spells.cpp#L109L156

### DIFF
--- a/Source/spells.cpp
+++ b/Source/spells.cpp
@@ -114,11 +114,7 @@ int GetManaAmount(PlayerStruct &player, spell_id sn)
 	int adj = 0;
 
 	// spell level
-	int sl = player._pSplLvl[sn] + player._pISplLvlAdd - 1;
-
-	if (sl < 0) {
-		sl = 0;
-	}
+	int sl = std::max(player._pSplLvl[sn] + player._pISplLvlAdd - 1, 0);
 
 	if (sl > 0) {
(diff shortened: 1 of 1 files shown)


Source/spells.cpp
```diff
@@ -131,15 +127,13 @@ int GetManaAmount(PlayerStruct &player, spell_id sn)
...
+		ma = std::max((spelldata[SPL_HEAL].sManaCost + 2 * player._pLevel - adj), 0);
 	} else if (spelldata[sn].sManaCost == 255) {
-		ma = ((BYTE)player._pMaxManaBase - adj);
+		ma = std::max(((BYTE)player._pMaxManaBase - adj), 0);
 	} else {
-		ma = (spelldata[sn].sManaCost - adj);
+		ma = std::max((spelldata[sn].sManaCost - adj), 0);
 	}
 
-	if (ma < 0)
```
qndel, line 141: why not leave the code above alone and just use max here?

bcko: ahh you are right, updated the pull request. Thank you @qndel !

AJenbo (approved):

PR #2559, opened 2021-08-10
diablo_quit also frees game state (and with this sound objects)

### LINKED ISSUE #2220: Crash on exit
**Important information**
Operating system Win 7 x64/ DevilutionX version 1.2.1 / official release

**Describe**
Sometimes (not often) it crashes upon pressing Alt+F4.

**To Reproduce**
1. Press Alt+F4.

**Additional context**
Here's a crash dump: [devilutionx crash.zip](https://github.com/diasurgical/devilutionX/files/6705153/devilutionx.crash.zip)

obligaron:
Fixes #2220 (at least on my pc) 🙂 

Notes:

- Sound was still playing, that was the cause of the exception (at least on my pc)
- `FreeGameMem` stops music and sound
- `FreeGameMem` is called if you exit a running game normally (in `RunGameLoop`).
- In menu only `music_stop` is called, but `FreeGameMem` includes `music_stop`.
- `FreeGameMem` can be called multiple times
- Alternative Solution would be to only stop music & sound.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1689,6 +1689,7 @@ bool StartGame(bool bNewGame, bool bSinglePlayer)
 
 void diablo_quit(int exitStatus)
 {
+	FreeGameMem();
 	DiabloDeinit();
 	exit(exitStatus);
 }


ephphatha (approved):
Works for me (windows and msvc as well though 😉 )

AJenbo (approved):

PR #2567, opened 2021-08-10
Automap dirt fixes

AJenbo:
Original:
image

master:
image

This fix:
image


Source/automap.cpp
```diff
@@ -210,13 +210,13 @@ void DrawCaveVertical(const Surface &out, Point center, AutomapTile tile)
...
 void DrawAutomapTile(const Surface &out, Point center, AutomapTile tile)
 {
-	if (tile.type == AutomapTile::Types::None)
-		return;
-
 	if (tile.HasFlag(AutomapTile::Flags::Dirt)) {
 		DrawDirt(out, center);
 	}
 
+	if (tile.type == AutomapTile::Types::None)
```
ephphatha, line 217: Is this case necessary? There are some automap tile entries with the stairs flag and type none which get skipped by this condition, if drawing those tiles is acceptable or preferred this could simplify the logic further.

ephphatha (approved):
Looks good, also raises a question whether we should be skipping the DrawStairs call but I haven't looked into the impact of changing that :D

Source/automap.cpp
AJenbo: Hmm the original probably checked that both type and flags where 0 (because both where handled as a single int16), and with the rest of the cleanups to `DrawAutomapTile()` simply dropping this check might actually be closer both in performance an behavior.

Issue #1610, opened 2021-04-19
Casting lvl 0 spell produces "not enough mana" sound


qndel:
**Important information**
Vanilla bug

**Describe**
When you cast a lvl 0 spell (you can achieve these with shrines), the character will say "Not enough mana" which is a lie

**To Reproduce**
1. Get a lvl 0 spell
2. Cast it
3. "Not enough mana"

**Expected behavior**
I'd expect a different message like "No way", not "Not enough mana" which isn't true

Better sounds:
PS_MAGE17 (No way)
PS_MAGE18 (Not a chance)
PS_MAGE19 (I would never use this) (Perfection!)
PS_MAGE28 (I cannot cast that yet)
PS_MAGE29 (That did not do anything)
PS_MAGE33 (I do not want that)
PS_MAGE34 (I do not have a spell ready)
PS_MAGE36 (I cannot do that) (YES!)
PS_MAGE37 (No!)
PS_MAGE39 (That will not work)

Oh well, seems I messed the enums (means we need a redesign ...) but the text matters ;p

AJenbo:
PS_MAGE28

nickgavran:
I like PS_MAGE36 the best

julealgon:
I'm also partial to "I cannot cast that yet".

RicoLeNecro:
How about randomly selecting one of those answers that kind of make sense each time the player attempts it? (Kind of like with the sounds of the cows in town.)

qndel:
Not worth it.

RicoLeNecro:
Well if we really have to chose a single one, my vote goes to PS_MAGE34.

malvarenga123:
+1 to PS_MAGE28

julealgon:
> Well if we really have to chose a single one, my vote goes to PS_MAGE34.

@RicoLeNecro I think that would be misleading. "Having a spell ready" in the game is having something selected in the "Readied Spell" icon (that's the actual name of the icon). If you select something there, then it means you _have_ a spell ready. 

Now, being unable to cast your readied spell is something else entirely and should be handled with a different message.

RicoLeNecro:
One could interpret a level 0 spell as never being in a ready state, but I get how someone might get confused by the character saying that.

PR #2568, opened 2021-08-10
fixes casting level 0 spell says "not enough mana"

vladtepesch:
fixes https://github.com/diasurgical/devilutionX/issues/1610
fixes casting level 0 spell says "not enough mana"-> changed return value of `CheckSpell` to enum to reflect the fail reason

Iam not sure about the Busy-Case - Could not reproduce it. If the cursor is not "Hand" (e.g. on using town portal or oil) its impossible to cast without first aborting the other spell/task.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -491,7 +491,7 @@ spell_type GetSBookTrans(spell_id ii, bool townok)
 		st = RSPLTYPE_SKILL;
 	}
 	if (st == RSPLTYPE_SPELL) {
-		if (!CheckSpell(MyPlayerId, ii, st, true)) {
+		if (CheckSpell(MyPlayerId, ii, st, true) != SpellCheckResult::Success) {
 			st = RSPLTYPE_INVALID;
 		}
 		if ((char)(myPlayer._pSplLvl[ii] + myPlayer._pISplLvlAdd) <= 0) {
@@ -668,7 +668,7 @@ void DrawSpell(const Surface &out)
 	// BUGFIX: Move the next line into the if statement to avoid OOB (SPL_INVALID is -1) (fixed)
 	if (st == RSPLTYPE_SPELL && spl != SPL_INVALID) {
 		int tlvl = myPlayer._pISplLvlAdd + myPlayer._pSplLvl[spl];
-		if (!CheckSpell(MyPlayerId, spl, st, true))
+		if (CheckSpell(MyPlayerId, spl, st, true) != SpellCheckResult::Success)
 			st = RSPLTYPE_INVALID;
 		if (tlvl <= 0)
 			st = RSPLTYPE_INVALID;
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -3575,19 +3575,17 @@ void CheckPlrSpell()
 
 	if (!addflag) {
 		if (myPlayer._pRSplType == RSPLTYPE_SPELL) {
-			HeroSpeech sentence;
 			switch (spellcheck) {
 			case SpellCheckResult::Fail_NoMana:
-				sentence = HeroSpeech::NotEnoughMana;
+				myPlayer.Say(HeroSpeech::NotEnoughMana);
 				break;
 			case SpellCheckResult::Fail_Level0:
-				sentence = HeroSpeech::ICantCastThatYet;
+				myPlayer.Say(HeroSpeech::ICantCastThatYet);
 				break;
 			default:
-				sentence = HeroSpeech::ICantDoThat;
+				myPlayer.Say(HeroSpeech::ICantDoThat);
 				break;
 			}
-			myPlayer.Say(sentence);
 			LastMouseButtonAction = MouseActionType::None;
 		}
 		return;
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -3556,11 +3556,12 @@ void CheckPlrSpell()
 				return;
 		}
 	}
-
+	SpellCheckResult spellcheck = SpellCheckResult::Success;
 	switch (myPlayer._pRSplType) {
 	case RSPLTYPE_SKILL:
 	case RSPLTYPE_SPELL:
-		addflag = CheckSpell(MyPlayerId, rspell, myPlayer._pRSplType, false);
+		spellcheck = CheckSpell(MyPlayerId, rspell, myPlayer._pRSplType, false);
+		addflag = spellcheck == SpellCheckResult::Success;
 		break;
 	case RSPLTYPE_SCROLL:
 		addflag = UseScroll();
@@ -3574,7 +3575,19 @@ void CheckPlrSpell()
 
 	if (!addflag) {
 		if (myPlayer._pRSplType == RSPLTYPE_SPELL) {
-			myPlayer.Say(HeroSpeech::NotEnoughMana);
+			HeroSpeech sentence;
+			switch (spellcheck) {
+			case SpellCheckResult::Fail_NoMana:
+				sentence = HeroSpeech::NotEnoughMana;
+				break;
+			case SpellCheckResult::Fail_Level0:
+				sentence = HeroSpeech::ICantCastThatYet;
+				break;
+			default:
+				sentence = HeroSpeech::ICantDoThat;
+				break;
+			}
+			myPlayer.Say(sentence);
 			LastMouseButtonAction = MouseActionType::None;
 		}
 		return;
--- a/Source/spells.cpp
+++ b/Source/spells.cpp
@@ -196,28 +196,31 @@ void EnsureValidReadiedSpell(PlayerStruct &player)
 	}
 }
 
-bool CheckSpell(int id, spell_id sn, spell_type st, bool manaonly)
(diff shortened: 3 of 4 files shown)


Source/control.cpp
```diff
@@ -592,7 +592,7 @@ void DrawSpell(const Surface &out)
 	// BUGFIX: Move the next line into the if statement to avoid OOB (SPL_INVALID is -1) (fixed)
 	if (st == RSPLTYPE_SPELL && spl != SPL_INVALID) {
 		int tlvl = myPlayer._pISplLvlAdd + myPlayer._pSplLvl[spl];
-		if (!CheckSpell(MyPlayerId, spl, st, true))
+		if (SpellFailReason::Success != CheckSpell(MyPlayerId, spl, st, true))
```
julealgon, line 595: Are we using the yoda checks @AJenbo ?

Source/spells.h
```diff
@@ -9,9 +9,17 @@
 
 namespace devilution {
 
+enum class SpellFailReason : uint8_t {
+	Success = 0,
+	None = 0,
+	NoMana,
+	Level0,
+	Busy
+};
```
julealgon, line 18: Can't we use a discriminated union instead of an enum to map these?

Also, I really dislike "Fail.Success" type of result (dreaded Windows API ERROR_SUCCESS). If we really need an enum here, I'd recommend something neutral as the enum name, like `SpellCheckingResult`, with a "Success" member (instead of None),

Source/spells.h
```diff
@@ -9,9 +9,17 @@
 
 namespace devilution {
 
+enum class SpellFailReason : uint8_t {
+	Success = 0,
+	None = 0,
```
julealgon, line 14: Why have 2 names for the same result? This seems like a code smell to me.

Source/control.cpp
vladtepesch: sorry, I am a bit corrupted by MISRA standard rules

Source/spells.h
vladtepesch: Yes I also hesitated changed it back and forth. checking the result of function calls feels more naturally if `checkSpell == success` instead of `checkSpell == none` but on other places `none` looks more natural.

Source/spells.h
vladtepesch: what do you mean with `discriminated union`?

Source/control.cpp
AJenbo: I was initiallys sold on the idea, but i have since found that good linters do just as good a job and doesn't hurt readability.

Source/spells.h
AJenbo: I know what it is, but not sure how it would apply here.
https://en.wikipedia.org/wiki/Tagged_union

Source/control.cpp
```diff
@@ -493,7 +493,7 @@ spell_type GetSBookTrans(spell_id ii, bool townok)
 		st = RSPLTYPE_SKILL;
 	}
 	if (st == RSPLTYPE_SPELL) {
-		if (!CheckSpell(MyPlayerId, ii, st, true)) {
+		if (SpellFailReason::Success != CheckSpell(MyPlayerId, ii, st, true)) {
```
AJenbo, line 496: This will always be true as None is always returned where Success could have. Pick one :)

vladtepesch: `none` and `success` are synonyms, so its not always true. But I guess your are right - only one should be used.
Any naming suggestions? 

@julealgon suggested changing the result enum to  `SpellCheckingResult`  
values:
- Success
- Fail_Level0
- Fail_NoMana
- Fail_Busy (does this ever happen? If yes - how to achieve this case?)

This maybe could also used for staff and scroll cast tests.
For this 
- Fail_NoCharge
- Fail_NoScroll
- Fail_WronPlace

Source/spells.h
vladtepesch: same for me - thatswhy I thought that he ment something different

Source/control.cpp
AJenbo: `FailReason == Success` doesn't make sens to me, it seams a different category.

`Result == Success/Fail` seams fine.

Source/spells.cpp
```diff
@@ -196,28 +196,31 @@ void EnsureValidReadiedSpell(PlayerStruct &player)
...
+		return SpellCheckResult::Success;
 	}
 
 	if (GetSpellLevel(id, sn) <= 0) {
-		return false;
+		return SpellCheckResult::Fail_Level0;
 	}
 
 	auto &player = Players[id];
+	if (player._pMana < GetManaAmount(player, sn)) {
```
AJenbo, line 219: mana for other players are not synced, make sure this doesn't break spell casting in multiplayer

Source/spells.cpp
```diff
@@ -196,28 +196,31 @@ void EnsureValidReadiedSpell(PlayerStruct &player)
...
 {
 #ifdef _DEBUG
 	if (debug_mode_key_inverted_v)
-		return true;
+		return SpellCheckResult::Success;
 #endif
 
 	if (!manaonly && pcurs != CURSOR_HAND) {
-		return false;
+		return SpellCheckResult::Fail_Busy;
```
AJenbo, line 207: You could probably trigger this (if you wanted to test it), by firing two fireballs and before the second on fire pick up an item in the inventory.

Source/spells.h
```diff
@@ -9,9 +9,21 @@
 
 namespace devilution {
 
+enum class SpellCheckResult : uint8_t {
+	Success = 0,
```
AJenbo, line 13: To match other enum styles
```suggestion
	Success,
```

Source/spells.h
```diff
@@ -9,9 +9,21 @@
...
+enum class SpellCheckResult : uint8_t {
+	Success = 0,
+	Fail_NoMana,
+	Fail_Level0,
+	Fail_Busy,
+	/* following values may be used as result of refactored UseStaff or UseScroll function
+	Fail_NoCharge,
+	Fail_NoScroll,
+	Fail_WrongPlace,
+	*/
```
AJenbo, line 21: Lets worry about that when/if we get there:
```suggestion
```

Source/player.cpp
```diff
@@ -3577,7 +3578,19 @@ void CheckPlrSpell()
...
+				sentence = HeroSpeech::NotEnoughMana;
+				break;
+			case SpellCheckResult::Fail_Level0:
+				sentence = HeroSpeech::ICantCastThatYet;
+				break;
+			default:
+				sentence = HeroSpeech::ICantDoThat;
+				break;
+			}
+			myPlayer.Say(sentence);
```
AJenbo, line 3593: I don't think we are actually saving anything by using the var and it compiles to the same code:
```suggestion
			switch (spellcheck) {
			case SpellCheckResult::Fail_NoMana:
				myPlayer.Say(HeroSpeech::NotEnoughMana);
				break;
			case SpellCheckResult::Fail_Level0:
				myPlayer.Say(HeroSpeech::ICantCastThatYet);
				break;
			default:
				myPlayer.Say(HeroSpeech::ICantDoThat);
				break;
			}
```

AJenbo (approved):

Source/spells.cpp
vladtepesch: in this case, maybe nothing should be said. so the buisy case should result in nothing

AJenbo:
Thanks for solving this

vladtepesch:
sorry, 

> 
> 
> Thanks for solving this

sorry. if i had knwon that you are that fast i hadn't force pushed. but it seems that my clang-format does changes on places it shouldn't and I rebased to master in beforehand

AJenbo:
??? what are you sorry for, you fixed a bug in the game and got it to a place where we can easily merge it.

vladtepesch:
> 
> 
> ??? what are you sorry for, you fixed a bug in the game and got it to a place where we can easily merge it.

;-) just because I changed the allready reviewed commit mutiple times.

But I have to thank you for your effords on maining such a old game and make it editiable. How often am I annoyed about bugs or unconviniences in devices and games that would so easy fixable if just the code would be available ^^

AJenbo:
> How often am I annoyed about bugs or unconviniences in devices and games that would so easy fixable if just the code would be available ^^

I feel you :)

Source/spells.h
julealgon: Sorry for not replying back. Discriminated unions are done using the `std::variant` type in C++, which signals a value can be one of a closed set of possible results.

AJenbo: Are you suggesting to use `HeroSpeech` as the return type?

julealgon: Not really, since the HeroSpeech to use is an implementation detail on the caller and not the responsibility of this method.

Just saying discriminated unions model this kind of scenario better than a raw enum does IMHO. For example, the case of "busy" could also return what the current action is, while the case of "no mana" could return how much mana was needed vs what the player currently has etc.

It is just a more robust way to communicate potentially different return values that is also more extensible.

PR #2570, opened 2021-08-11
Add clang-format validation to CI

obligaron:
Cause without validation new violations lurk in (see clang-format commits in this pr 😉).
This PR adds automated validation, that checks that all files under `Source` and `test` are clang-formated.
Uses [clang-format-check](https://github.com/marketplace/actions/clang-format-check) github action.


AJenbo:
Hoping this support the github error format so that we can get inline errors :)

test/picosha2.h
```diff
@@ -27,7 +27,7 @@ THE SOFTWARE.
 
 #ifndef PICOSHA2_BUFFER_SIZE_FOR_INPUT_ITERATOR
```
AJenbo, line 28: Can we exclude this file?

obligaron:
> 
> 
> Hoping this support the github error format so that we can get inline errors :)

Interesting I didn't knew a github error format exists. Just for me: what it is and where get the errors inlined? 🙂 

btw: had to reformat `player.cpp` again cause of new introduced clang-format error. So I think this pr has some value. 😁

AJenbo (approved):

PR #2574, opened 2021-08-11
Clean up multiplayer XP cap clamping

ephphatha:
Noticed this while reviewing #2572 but it got merged before I could finish the review 😅 

It may be clearer to write out the std::min call using two separate checks instead of an initialiser list. Happy to change that (or have it edited on my behalf :)).

Future cleanups are possible in this area. player._pExperience is pretty much always positive so could be an unsigned value. As far as I can tell players never lose experience (and exp is clamped to 0 at the lowest) so the addition can prevent overflows. This function is also only ever called with pnum = MyPlayerId so the first two checks are redundant.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -409,9 +409,9 @@ void LoadPlayer(LoadHelper *file, int p)
 	player._pLevel = file->NextLE<int8_t>();
 	player._pMaxLvl = file->NextLE<int8_t>();
 	file->Skip(2); // Alignment
-	player._pExperience = file->NextLE<int32_t>();
-	player._pMaxExp = file->NextLE<int32_t>();
-	player._pNextExper = file->NextLE<int32_t>();
+	player._pExperience = file->NextLE<uint32_t>();
+	file->Skip<uint32_t>();                        // Skip _pMaxExp - unused
+	player._pNextExper = file->NextLE<uint32_t>(); // This can be calculated based on pLevel (which in turn could be calculated based on pExperience)
 	player._pArmorClass = file->NextLE<int8_t>();
 	player._pMagResist = file->NextLE<int8_t>();
 	player._pFireResist = file->NextLE<int8_t>();
@@ -1081,9 +1081,9 @@ void SavePlayer(SaveHelper *file, int p)
 	file->WriteLE<int8_t>(player._pLevel);
 	file->WriteLE<int8_t>(player._pMaxLvl);
 	file->Skip(2); // Alignment
-	file->WriteLE<int32_t>(player._pExperience);
-	file->WriteLE<int32_t>(player._pMaxExp);
-	file->WriteLE<int32_t>(player._pNextExper);
+	file->WriteLE<uint32_t>(player._pExperience);
+	file->Skip<uint32_t>(); // Skip _pMaxExp
+	file->WriteLE<uint32_t>(player._pNextExper);
 	file->WriteLE<int8_t>(player._pArmorClass);
 	file->WriteLE<int8_t>(player._pMagResist);
 	file->WriteLE<int8_t>(player._pFireResist);
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -3409,7 +3409,7 @@ bool OperateShrineGlowing(int pnum)
 	auto &myPlayer = Players[MyPlayerId];
 
 	// Add 0-5 points to Magic (0.1% of the players XP)
-	ModifyPlrMag(MyPlayerId, std::min(myPlayer._pExperience / 1000, 5U));
+	ModifyPlrMag(MyPlayerId, static_cast<int>(std::min<uint32_t>(myPlayer._pExperience / 1000, 5)));
 
 	// Take 5% of the players experience to offset the bonus, unless they're very low level in which case take all their experience.
 	if (myPlayer._pExperience > 5000)
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -3408,19 +3408,17 @@ bool OperateShrineGlowing(int pnum)
 
 	auto &myPlayer = Players[MyPlayerId];
 
-	int playerXP = myPlayer._pExperience;
-	int magicGain = playerXP / 1000;
-	int xpLoss = 0;
-	if (playerXP > 5000) {
-		magicGain = 5;
-		xpLoss = static_cast<int>(playerXP * 0.95);
-	}
-	ModifyPlrMag(MyPlayerId, magicGain);
-	myPlayer._pExperience = xpLoss;
+	// Add 0-5 points to Magic (0.1% of the players XP)
+	ModifyPlrMag(MyPlayerId, std::min(myPlayer._pExperience / 1000, 5U));
 
-	if (sgOptions.Gameplay.bExperienceBar) {
+	// Take 5% of the players experience to offset the bonus, unless they're very low level in which case take all their experience.
+	if (myPlayer._pExperience > 5000)
+		myPlayer._pExperience = static_cast<uint32_t>(myPlayer._pExperience * 0.95);
+	else
+		myPlayer._pExperience = 0;
+
+	if (sgOptions.Gameplay.bExperienceBar)
 		force_redraw = 255;
-	}
 
 	CheckStats(Players[pnum]);
 
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2669,19 +2669,11 @@ void AddPlrExperience(int pnum, int lvl, int exp)
 
 	// Prevent power leveling
 	if (gbIsMultiplayer) {
-		int powerLvlCap = player._pLevel < 0 ? 0 : player._pLevel;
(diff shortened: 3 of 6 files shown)


Source/player.cpp
```diff
@@ -2669,23 +2669,15 @@ void AddPlrExperience(int pnum, int lvl, int exp)
...
-		// cap to 1/20 of current levels xp
-		if (exp >= ExpLvlsTbl[powerLvlCap] / 20) {
-			exp = ExpLvlsTbl[powerLvlCap] / 20;
-		}
-		// cap to 200 * current level
-		int expCap = 200 * powerLvlCap;
-		if (exp >= expCap) {
-			exp = expCap;
-		}
+		auto clampedPlayerLevel = std::clamp(static_cast<int>(player._pLevel), 0, 50);
```
AJenbo, line 2672: We have our own `clamp()` for a reason ;)
```suggestion
		auto clampedPlayerLevel = clamp(static_cast<int>(player._pLevel), 0, 50);
```

Source/player.cpp
```diff
@@ -2669,23 +2669,15 @@ void AddPlrExperience(int pnum, int lvl, int exp)
...
+
+		// for low level characters experience gain is capped to 1/20 of current levels xp
+		// for high level characters experience gain is capped to 200 * current level - this is a smaller value than 1/20 of the exp needed for the next level after level 5.
+		exp = std::min({ exp, /* level 0-5: */ ExpLvlsTbl[clampedPlayerLevel] / 20, /* level 6-50: */ 200 * clampedPlayerLevel });
 	}
 
 	player._pExperience += exp;
-	if ((DWORD)player._pExperience > MAXEXP) {
+	if (player._pExperience > MAXEXP || player._pExperience < 0) {
```
AJenbo, line 2680: im fine changing _pExperience to unsigned

Source/player.cpp
ephphatha: Been meaning to ask about that one, it's defined as std::clamp for c++17 which is the version devilutionx targets. Is there a platform where c++17 support is incomplete that the alternate definition handles?

AJenbo: Yes, Amiga and PowerPC Linux. They are on GCC 6.2 which only has partial C++17 support. There is also a guy somewhere who insist on Slackware which has GCC 5.5, it currently works, but I'm less concerned about it.

Source/objects.cpp
```diff
@@ -3408,15 +3408,13 @@ bool OperateShrineGlowing(int pnum)
...
-	}
-	ModifyPlrMag(MyPlayerId, magicGain);
-	myPlayer._pExperience = xpLoss;
+	// Add 0-5 points to Magic (0.1% of the players XP)
+	ModifyPlrMag(MyPlayerId, std::min(myPlayer._pExperience / 1000, 5U));
+
+	// Take 5% of the players experience to offset the bonus, unless they're very low level.
+	// Should this be >=? if someone has 5000 XP exactly they get the max bonus and lose nothing.
+	if (myPlayer._pExperience > 5000)
+		myPlayer._pExperience = static_cast<uint32_t>(myPlayer._pExperience * 0.95);
```
ephphatha, line 3417: I introduced a bug here... Fixing that now

Source/objects.cpp
```diff
@@ -3408,15 +3408,13 @@ bool OperateShrineGlowing(int pnum)
...
-		magicGain = 5;
-		xpLoss = static_cast<int>(playerXP * 0.95);
-	}
-	ModifyPlrMag(MyPlayerId, magicGain);
-	myPlayer._pExperience = xpLoss;
+	// Add 0-5 points to Magic (0.1% of the players XP)
+	ModifyPlrMag(MyPlayerId, std::min(myPlayer._pExperience / 1000, 5U));
+
+	// Take 5% of the players experience to offset the bonus, unless they're very low level.
+	// Should this be >=? if someone has 5000 XP exactly they get the max bonus and lose nothing.
```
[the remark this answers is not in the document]
ephphatha: This was something I broke and have since fixed :)

Source/objects.cpp
```diff
@@ -3408,19 +3408,17 @@ bool OperateShrineGlowing(int pnum)
...
-	int magicGain = playerXP / 1000;
-	int xpLoss = 0;
-	if (playerXP > 5000) {
-		magicGain = 5;
-		xpLoss = static_cast<int>(playerXP * 0.95);
-	}
-	ModifyPlrMag(MyPlayerId, magicGain);
-	myPlayer._pExperience = xpLoss;
+	// Add 0-5 points to Magic (0.1% of the players XP)
+	ModifyPlrMag(MyPlayerId, static_cast<int>(std::min(static_cast<uint32_t>(myPlayer._pExperience / 1000U), 5U))); // 3ds builds promote the division result to a 64 bit type... downcast manually so the template deduction works
```
AJenbo, line 3412: Missing `<algorithm>` in this file it seams.

Source/loadsave.cpp
```diff
@@ -1081,9 +1081,9 @@ void SavePlayer(SaveHelper *file, int p)
 	file->WriteLE<int8_t>(player._pLevel);
 	file->WriteLE<int8_t>(player._pMaxLvl);
 	file->Skip(2); // Alignment
-	file->WriteLE<int32_t>(player._pExperience);
-	file->WriteLE<int32_t>(player._pMaxExp);
-	file->WriteLE<int32_t>(player._pNextExper);
+	file->WriteLE<uint32_t>(player._pExperience);
+	file->Skip<uint32_t>(); // Skip _pMaxExp
```
AJenbo, line 1085: Should still be save for BC

Source/player.cpp
```diff
@@ -2665,30 +2662,26 @@ void AddPlrExperience(int pnum, int lvl, int exp)
...
 	}
 
+	constexpr uint32_t MaxExperience = 2000000000U;
+
+	const uint32_t newExperienceValue = player._pExperience + clampedExp;
+	if (newExperienceValue > MaxExperience) {
+		player._pExperience = MaxExperience;
+	} else if (newExperienceValue > player._pExperience) {
+		player._pExperience = newExperienceValue;
+	} // else experience has overflowed, could either reorder the checks to apply MAXEXP or just let the existing value stay (the current approach)
```
AJenbo, line 2683: What is `MAXEXP` :)
```suggestion
	} // else experience has overflowed, could either reorder the checks to apply MaxExperience or just let the existing value stay (the current approach)
```

Source/player.cpp
```diff
@@ -2665,30 +2662,26 @@ void AddPlrExperience(int pnum, int lvl, int exp)
...
-		}
-	}
+		const uint32_t clampedPlayerLevel = clamp(static_cast<int>(player._pLevel), 0, 50);
 
-	player._pExperience += exp;
-	if ((DWORD)player._pExperience > MAXEXP) {
-		player._pExperience = MAXEXP;
+		// for low level characters experience gain is capped to 1/20 of current levels xp
+		// for high level characters experience gain is capped to 200 * current level - this is a smaller value than 1/20 of the exp needed for the next level after level 5.
+		clampedExp = std::min({ clampedExp, /* level 0-5: */ ExpLvlsTbl[clampedPlayerLevel] / 20U, /* level 6-50: */ 200U * clampedPlayerLevel });
```
AJenbo, line 2673: wow did not know you could give min a list, neat :)

Source/player.cpp
```diff
@@ -2665,30 +2662,26 @@ void AddPlrExperience(int pnum, int lvl, int exp)
...
 	}
 
+	constexpr uint32_t MaxExperience = 2000000000U;
+
+	const uint32_t newExperienceValue = player._pExperience + clampedExp;
+	if (newExperienceValue > MaxExperience) {
+		player._pExperience = MaxExperience;
+	} else if (newExperienceValue > player._pExperience) {
+		player._pExperience = newExperienceValue;
+	} // else experience has overflowed, could either reorder the checks to apply MaxExperience or just let the existing value stay (the current approach)
```
AJenbo, line 2683: Woudn't this work for what we need?
```suggestion
	player._pExperience = clamp(newExperienceValue, player._pExperience, MaxExperience);
```

I would also think it's very unlikely for an overflow to happen here with you having to receive a minimum of `2294967296` xp from one kill in order to do so at this point.

```suggestion
	player._pExperience = std::min(newExperienceValue, MaxExperience);
```

AJenbo (changes requested):
Just one comment left, after reading all of the changes carefully :)

Sorry this is turning in to a bit of a saga :open_mouth:

Source/player.cpp
ephphatha: yeah, this was being overly cautious. The first suggestion using clamp would be fine, or this can just directly modify _pExperience again and apply the cap after += like the old code did.

ephphatha: went with the std::min and no temp variable option :D

AJenbo (approved):

AJenbo (approved):

PR #2575, opened 2021-08-11
cleanup manual value clamping: inv.cpp

bcko:
part of #1666

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -451,7 +451,7 @@ void CheckInvPaste(int pnum, Point cursorPosition)
 						done = false;
 					} else {
 						if (player.InvGrid[xx + yy] != 0) {
-							int8_t iv = std::max<uint8_t>(player.InvGrid[xx + yy], 0);
+							int8_t iv = abs(player.InvGrid[xx + yy]);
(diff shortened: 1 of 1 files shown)


Source/inv.cpp
```diff
@@ -441,23 +441,17 @@ void CheckInvPaste(int pnum, Point cursorPosition)
...
+				int xx = std::max((ii % INV_ROW_SLOT_SIZE) - ((itemSize.width - 1) / 2), 0);
 				for (i = 0; i < itemSize.width && done; i++) {
 					if (xx >= INV_ROW_SLOT_SIZE) {
 						done = false;
 					} else {
 						if (player.InvGrid[xx + yy] != 0) {
-							int8_t iv = player.InvGrid[xx + yy];
-							if (iv < 0)
-								iv = -iv;
+							int8_t iv = std::max<uint8_t>(player.InvGrid[xx + yy], 0);
```
ephphatha, line 454: this should be a call to `abs()`

bcko: updated. Thank you @ephphatha !

ephphatha (approved):
👍

Issue #2576, opened 2021-08-11
Attempting to cast a resurrect scroll on self causes a CTD


0mnicydle:
**Important information**
Windows 10  x64 / latest master / download or compile

**Describe**
Attempting to cast a resurrect scroll on self causes a CTD.

**Expected behavior**
Nothing happens, scroll is wasted, game continues.

**Additional context**
I believe I tracked it down to commit 7effa1a931e7e33d2afface5438e55ed76e3dc15

[multi_0.zip](https://github.com/diasurgical/devilutionX/files/6970118/multi_0.zip)

qndel:
![image](https://user-images.githubusercontent.com/14297035/129081919-424fefd2-6503-4a45-86bf-a2ec1cc2cf1d.png)

0mnicydle:
Just in case someone is waiting for me to respond, I can report this PR does indeed fix the crash on my end.

julealgon:
Why remove the 1.3.0 milestone @qndel ?

AJenbo:
Probably because it was a bug introduced after 1.2.1, so he might be keeping the milestone more useful once we need to create a change log.

PR #2577, opened 2021-08-11
fix resurrect

qndel:
copy/pasted the part from DoHealOther
Resolves https://github.com/diasurgical/devilutionX/issues/2576
int rid would never reach -1 because 
image

### DIFF
--- a/Source/spells.cpp
+++ b/Source/spells.cpp
@@ -246,14 +246,15 @@ void CastSpell(int id, int spl, int sx, int sy, int dx, int dy, int spllvl)
  * @param pnum player index
  * @param rid target player index
  */
-void DoResurrect(int pnum, int rid)
+void DoResurrect(int pnum, uint16_t rid)
 {
 	if (pnum == MyPlayerId) {
 		NewCursor(CURSOR_HAND);
 	}
 
-	if (rid == -1)
+	if ((DWORD)pnum >= MAX_PLRS || rid >= MAX_PLRS) {
 		return;
+	}
 
 	auto &target = Players[rid];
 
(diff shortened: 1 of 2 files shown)


Source/spells.cpp
```diff
@@ -246,14 +246,15 @@ void CastSpell(int id, int spl, int sx, int sy, int dx, int dy, int spllvl)
...
  */
-void DoResurrect(int pnum, int rid)
+void DoResurrect(int pnum, uint16_t rid)
 {
 	if (pnum == MyPlayerId) {
 		NewCursor(CURSOR_HAND);
 	}
 
-	if (rid == -1)
+	if ((DWORD)pnum >= MAX_PLRS || rid >= MAX_PLRS) {
```
julealgon, line 255: Why would either of these ever be greater than `MAX_PLRS`?

Can't we make `rid` `optional<int>` instead?

StephenCWills: Because they're parameters sent via network messages so they need to be validated regardless.

AJenbo: We have added a bit of validation for these values, but the more the better before we remove the encryption requirement and implement a game list :hear_no_evil:

AJenbo (approved):

Source/spells.cpp
AJenbo: That being said `pnum` should have been validated before this point or we would have OOBs in `ParseCmd()`

StephenCWills: FYI, this was previously discussed in #2222.

StephenCWills (approved):

Source/spells.cpp
qndel: I just copy/pasted the code from heal other, if you think it's wrong, fix heal other too ;)

PR #2580, opened 2021-08-12
cleanup manual value clamping: sdl2_to_1_2_backports.cpp

bcko:
part of #1666

### DIFF
--- a/Source/utils/sdl2_to_1_2_backports.cpp
+++ b/Source/utils/sdl2_to_1_2_backports.cpp
@@ -458,8 +458,8 @@ int SDL_BlitScaled(SDL_Surface *src, SDL_Rect *srcrect,
 
 	final_dst.x = (Sint16)SDL_floor(dst_x0 + 0.5);
 	final_dst.y = (Sint16)SDL_floor(dst_y0 + 0.5);
-	dst_w = std::max((int)SDL_floor(dst_x1 - dst_x0 + 1.5), 0);
-	dst_h = std::max((int)SDL_floor(dst_y1 - dst_y0 + 1.5), 0);
+	dst_w = std::max(static_cast<int>(SDL_floor(dst_x1 - dst_x0 + 1.5)), 0);
+	dst_h = std::max(static_cast<int>(SDL_floor(dst_y1 - dst_y0 + 1.5)), 0);
 
 	final_dst.w = static_cast<Uint16>(dst_w);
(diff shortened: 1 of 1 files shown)


Source/utils/sdl2_to_1_2_backports.cpp
```diff
@@ -449,24 +450,16 @@ int SDL_BlitScaled(SDL_Surface *src, SDL_Rect *srcrect,
 
 	final_src.x = (Sint16)SDL_floor(src_x0 + 0.5);
 	final_src.y = (Sint16)SDL_floor(src_y0 + 0.5);
-	src_w = (int)SDL_floor(src_x1 + 1 + 0.5) - (int)SDL_floor(src_x0 + 0.5);
-	src_h = (int)SDL_floor(src_y1 + 1 + 0.5) - (int)SDL_floor(src_y0 + 0.5);
-	if (src_w < 0)
-		src_w = 0;
-	if (src_h < 0)
-		src_h = 0;
+	src_w = std::max((int)SDL_floor(src_x1 + 1 + 0.5) - (int)SDL_floor(src_x0 + 0.5), 0);
```
ephphatha, line 453: These calls to sdl_floor should be cleaned up. The only time a non-integer value is relevant is when determining and using the scaling_w/h variables, so with a bit of work these casts and calls to SDL_floor could be removed.

If you're willing to go through this code it'd be great to clean up the whole function. For now just changing the cast on these lines to c++ style static_casts would be ideal.

bcko: Updated the pull request to change (int) to static_cast<int>. Thank you @ephphatha !

At the moment, I might not know enough of c++, SDL, and devilutionx codebase to properly clean up the whole function. As I gain more knowledge, I am happy to revisit so I can clean up the whole function. :)

ephphatha: you got the bottom two casts but the four on this line and the next can use static_cast as well :)

ephphatha (changes requested):

Source/utils/sdl2_to_1_2_backports.cpp
bcko: Updated the pull request. I am learning a lot from your pull request reviews. Thank you @ephphatha ! :).

ephphatha (approved):

PR #2581, opened 2021-08-12
cleanup manual value clamping: monster.cpp

bcko:
part of #1666

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -189,10 +189,7 @@ void InitMonster(MonsterStruct &monster, Direction rd, int mtype, Point position
 	}
 
 	if (!gbIsMultiplayer) {
-		monster._mmaxhp /= 2;
-		if (monster._mmaxhp < 64) {
-			monster._mmaxhp = 64;
-		}
+		monster._mmaxhp = std::max(monster._mmaxhp / 2, 64);
 	}
 
 	monster._mhitpoints = monster._mmaxhp;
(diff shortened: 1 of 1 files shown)


Source/monster.cpp
```diff
@@ -1489,9 +1489,7 @@ void MonsterAttackPlayer(int i, int pnum, int hit, int minDam, int maxDam)
 				dam = 64;
 			int mdam = dam * (GenerateRnd(10) + 20L) / 100;
 			monster._mhitpoints -= mdam;
-			dam -= mdam;
-			if (dam < 0)
-				dam = 0;
+			dam = std::max(dam - mdam, 0);
```
ephphatha, line 1492: You can also clean up the `dam += player._pIGetHit` call in the same way a few lines up :)

bcko: Updated the pull request. Thank you @ephphatha !

ephphatha: There's another case of the same thing on lines 1525-1527 (a few lines above the next change)

bcko: updated the pull request clamp another case. Thank you!

Source/monster.cpp
```diff
@@ -189,10 +189,7 @@ void InitMonster(MonsterStruct &monster, Direction rd, int mtype, Point position
 	}
 
 	if (!gbIsMultiplayer) {
```
ephphatha, line 191: To keep this consistent with other single statement conditionals (most notably the identical code in PlaceUniqueMonst around line 509) the braces should be removed here.

ephphatha (approved):
Looks good, only one comment regarding code style but it's a non-functional change.

Source/monster.cpp
bcko: ahh you are right. Updated the pull request. Thank you @ephphatha !

ephphatha (approved):

PR #2582, opened 2021-08-12
cleanup manual value clamping: lighting.cpp

bcko:
part of #1666

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -5,6 +5,8 @@
  */
 #include "lighting.h"
 
+#include <algorithm>
+
 #include "automap.h"
 #include "diablo.h"
 #include "engine/load_file.hpp"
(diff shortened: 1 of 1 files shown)


ephphatha (dismissed):

Source/lighting.cpp
```diff
@@ -629,23 +631,10 @@ void DoUnVision(Point position, int nRadius)
...
-	if (x1 < 0) {
-		x1 = 0;
-	}
-	if (x2 > MAXDUNX) {
-		x2 = MAXDUNX;
-	}
+	int y1 = std::max(position.y - nRadius, 0);
+	int y2 = std::min(position.y + nRadius, MAXDUNY);
+	int x1 = std::max(position.x - nRadius, 0);
+	int x2 = std::min(position.x + nRadius, MAXDUNX);
```
AJenbo, line 637: Reshuffled to the logical order:
```suggestion
	int x1 = std::max(position.x - nRadius, 0);
	int y1 = std::max(position.y - nRadius, 0);
	int x2 = std::min(position.x + nRadius, MAXDUNX);
	int y2 = std::min(position.y + nRadius, MAXDUNY);
```

PR #2583, opened 2021-08-12
cleanup manual value clamping: stores.cpp 2

bcko:
part of #1666

### DIFF
--- a/Source/stores.cpp
+++ b/Source/stores.cpp
@@ -2606,10 +2606,7 @@ void StorePrior()
 	PlaySFX(IS_TITLEMOV);
 	if (stextsel != -1 && stextscrl) {
 		if (stextsel == stextup) {
-			if (stextsval != 0) {
-				stextsval -= 4;
-			}
-			stextsval = std::max(stextsval, 0);
+			stextsval = std::max(stextsval - 4, 0);
 		} else {
 			stextsel = stextup;
 		}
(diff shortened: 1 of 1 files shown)


Source/stores.cpp
```diff
@@ -2608,10 +2606,10 @@ void StorePrior()
 	PlaySFX(IS_TITLEMOV);
 	if (stextsel != -1 && stextscrl) {
 		if (stextsel == stextup) {
-			if (stextsval != 0)
+			if (stextsval != 0) {
 				stextsval -= 4;
-			if (stextsval < 0)
-				stextsval = 0;
+			}
+			stextsval = std::max(stextsval, 0);
```
ephphatha, line 2612:
```suggestion
			stextsval = std::max(stextsval - 4, 0);
```
The if != 0 statement is unnecessary, the decrement can be inlined and let std::max clamp to 0 in all cases.

Webkit Code Style guidelines forbid braces on single statement conditionals as well. (I prefer braces everywhere but try keep it consistent for this project 😅 )

bcko: Updated the pull request and looking into Webkit Code Style guideline (https://webkit.org/code-style-guidelines/). Thank you @ephphatha !

ephphatha (approved):

Source/stores.cpp
AJenbo: > I prefer braces everywhere

Same :D

AJenbo: @bcko if you haven't see it we have some additional rules on top of what Webkit provides: https://github.com/diasurgical/devilutionX/wiki/Code-Style

Issue #2435, opened 2021-07-20
Reduce use of globals


AJenbo:
The following globals are being referenced a lot where it could instead have been done by reference, changing this will make functions more dependable/testable.

- [ ] `Items`
- [ ] `Lights`
- [ ] `Missiles`
- [ ] `Objects`
- [ ] `QuestData`
- [ ] `Quests`
- [ ] `VisionList`
- [x] `Players`
- [x] `Monsters`
- [x] `Towners`

PracticalMetal:
Hi, I would like to work on this, can you suggest me ways to get started?

AJenbo:
A good place to start is to go over the functions in objects.cpp and for the ones that have multiple direct usages of `Objects`, instead get it by reference at the start of the function, for example `AddL1Door` which has 6 references to `Objects`:

```cpp
void AddL1Door(int i, Point position, _object_id objectType)
{
	Objects[i]._oDoorFlag = true;
	if (objectType == _object_id::OBJ_L1LDOOR) {
		Objects[i]._oVar1 = dPiece[position.x][position.y];
		Objects[i]._oVar2 = dPiece[position.x][position.y - 1];
	} else { //_object_id::OBJ_L1RDOOR
		Objects[i]._oVar1 = dPiece[position.x][position.y];
		Objects[i]._oVar2 = dPiece[position.x - 1][position.y];
	}
	Objects[i]._oVar4 = 0;
}
```

Instead we can put it in a variable called door (if the function doesn't deal with destinct type just call it `object`) and use in place of the global:
```cpp
void AddL1Door(int i, Point position, _object_id objectType)
{
	auto &door = Objects[i];
	
	door._oDoorFlag = true;
	if (objectType == _object_id::OBJ_L1LDOOR) {
		door._oVar1 = dPiece[position.x][position.y];
		door._oVar2 = dPiece[position.x][position.y - 1];
	} else { //_object_id::OBJ_L1RDOOR
		door._oVar1 = dPiece[position.x][position.y];
		door._oVar2 = dPiece[position.x - 1][position.y];
	}
	door._oVar4 = 0;
}
```

At this point it clear to see that `i` is only used as an index in the global and we can instead take the object as a reference and push the look up to the caller:
```cpp
void AddL1Door(ObjectStruct &door, Point position, _object_id objectType)
{
	door._oDoorFlag = true;
	if (objectType == _object_id::OBJ_L1LDOOR) {
[...]

PR #2585, opened 2021-08-12
#2435 Reduce use of globals: VisionList

bcko:
part of #2435

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -1099,13 +1099,14 @@ int AddVision(Point position, int r, bool mine)
 	int vid = -1; // BUGFIX: if VisionCount >= MAXVISION behavior is undefined (fixed)
 
 	if (VisionCount < MAXVISION) {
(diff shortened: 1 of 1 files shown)


Source/lighting.cpp
```diff
@@ -1140,48 +1141,50 @@ void ChangeVisionRadius(int id, int r)
...
+			vision.oldRadius = vision._lradius;
+			vision.position.tile = position;
 			dovision = true;
 		}
 	}
 }
 
 void ProcessVisionList()
 {
+	auto &vision = VisionList;
```
AJenbo, line 1157: Aliasing the global it self doesn't really help or solve anything.

qndel: still, I like this change at it will make passing visionlist as function argument a bit easier

AJenbo: Well, it will be irrelevant once the the function is cleaned up properly I think.

AJenbo (approved):

Issue #2578, opened 2021-08-11
Theo-Quest Girl is still crying after finishing quest


vladtepesch:
**Important information**
win 10 64bit self build based on commit 0e56e52c75

**Describe**
Clicking the girl after finishing the theo quest just plays the crying sound.

**Expected behavior**
I am quite sure that it not cried after returning Theo. I tried to verify this but I was unable to get the original game running on my computers.

AJenbo:
She only has 4 lines, i'm going to assume that she probably doesn't say anything when the quest is over.

Here is the extremely hard to read bin exact version:
https://github.com/diasurgical/devilution/blob/master/Source/towners.cpp#L1387

julealgon:
It's such terrible design to have a character that you can click on but doesn't say anything..... I have no idea why they didn't record a final voice for her.

Would be nice if we could provide at least a text-only version of some kind of farewell or thank-you message for her and use that instead of doing nothing.

NikoVP:
The end of her final speech would do. "Thanks again, hero person "

AJenbo:
Yeah the hellfire towners stick out for not having any gossip or menus. I'm wondering if it wouldn't make more sens to simply make them non clickable once there quest is finished.

vladtepesch:
> Could you check if there is a similar issue with `TalkToCowFarmer` and `TalkToFarmer`?

actually another thing that made me wonder was that the Farmer allways talks about the explosion after getting the auric amulet. Afair he also reacts differently.
And another thing: the cowfamer does not wear the dress i brought to him 
But for this two Bugs I wanted to check if it is actual my own fault because I changed the Cowquest to not replace the Farmer but be both to be active - but this change seemed to be quite minor --> https://github.com/vladtepesch/devilutionX/tree/CowQuestAndFarmer  



> Yeah the hellfire towners stick out for not having any gossip or menus.
The girls quest of course was never be official, but also the official HF quests are badly integrated. At least Cain should have something to say about Na'Kroul (or however its called) and the Cornerstone

NikoVP:
Sadly they never employed the original diablo voicecast when making the expansion. Our only hope is using artificial intelligence to generate dialogs. Someone has to write the dialog, make sure it's consistent with original diablo game lore. Another setback is that there aren't many good free generators. Maybe the free trial of Overdub can be used for that https://www.descript.com/overdub. Again we are entering mod area with that.

julealgon:
[...]

PR #2589, opened 2021-08-12
No More Tears

vladtepesch:
fixes https://github.com/diasurgical/devilutionX/issues/2578

### DIFF
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -742,6 +742,9 @@ void TalkToGirl(PlayerStruct &player, TownerStruct &girl)
 	case QUEST_ACTIVE:
 		InitQTextMsg(TEXT_GIRL3);
 		return;
+	case QUEST_DONE:
+		// do nothing
+		return;
 	default:
 		PlaySFX(Texts[TEXT_GIRL1].sfxnr);
 		return;
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -743,7 +743,6 @@ void TalkToGirl(PlayerStruct &player, TownerStruct &girl)
 		InitQTextMsg(TEXT_GIRL3);
 		return;
 	case QUEST_DONE:
-		// do nothing
 		return;
 	default:
 		PlaySFX(Texts[TEXT_GIRL1].sfxnr);


Source/towners.cpp
```diff
@@ -742,6 +742,9 @@ void TalkToGirl(PlayerStruct &player, TownerStruct &girl)
 	case QUEST_ACTIVE:
 		InitQTextMsg(TEXT_GIRL3);
 		return;
+	case QUEST_DONE:
+		// do nothing
+		return;
```
julealgon, line 747: Didn't we mention about not allowing the player to select the NPC anymore? That felt like a more robust option than just doing nothing.

I'd rather see that implemented if possible.

vladtepesch: puh - are you sure? I guess this would be confusing.

julealgon: > I guess this would be confusing.

That's exactly why. People expect _something_ to happen when clicking on a highlighted element in the game. Literally every other entity interaction results in some effect, sound, or a combination of both. "Nothing happening" goes against that standard behavior and is really confusing.

On the other hand, objects which have already been interacted with and cannot produce any more results, stop being highlightable. I think extending it to NPCs would make sense here.

Again, I'm not particularly opposed to actually adding some dialog line for her, but disabling the interaction is a conceptually simpler change that avoids the confusion without having to introduce extra content to the game.

Source/towners.cpp
```diff
@@ -742,6 +742,9 @@ void TalkToGirl(PlayerStruct &player, TownerStruct &girl)
 	case QUEST_ACTIVE:
 		InitQTextMsg(TEXT_GIRL3);
 		return;
+	case QUEST_DONE:
```
AJenbo, line 745: Won't this result in the girl not saying thank you after getting the bear?

Source/towners.cpp
vladtepesch: i can do this but the changes would be more complex. should also the mouse over info be disabled?

Source/towners.cpp
vladtepesch: this saying should be triggered by the first if - condition, isn't? surprisingly the second part is not in an `else` so i would have expected that right after finishing the quest the girl shoud have started crying.

Source/towners.cpp
AJenbo: I think fixing the dialog logic would be a good, and important, first step. Improvements can come later. That way we don't do multiple things in one commit either.

Source/towners.cpp
AJenbo: Yeah that is what it looks like.

Source/towners.cpp
NikoVP: She mentions of going home, maybe completely remove her if she isn't supposed to interact.

vladtepesch: I can do this, if you wish but at first at least the farmer quest fix should be merged because it also would need the same deactivation.

Source/towners.cpp
AJenbo: Just tested it, she cries but the text is correct...

vladtepesch: which version have you tested? my fix? I also tested it and she was not crying 🤔

Source/towners.cpp
vladtepesch: letting her go home, is also a nice idea.

Source/towners.cpp
AJenbo: current master.

vladtepesch: then everything is fine

Source/towners.cpp
vladtepesch: okay i added this.
But not seeing her makes the tree looking somewhat empty ;) and feels like if the game was just startet new

vladtepesch: additionally to this the girls sprite with the teddy will not be shown anymore (as it only is loaded on init if quest is done)

so maybe I tend to roll this change back

AJenbo:
https://www.youtube.com/watch?v=CprfjfN5PRs

Source/towners.cpp
vladtepesch: so I reset the change. Imho it would be best to first just recreate the correct vanilla behaviour of not crying on Teddys return

FitzRoyX: I think this speaks more to a deficiency in the quest log, which doesn't currently have a "completed quests" section. I have taken long breaks and come back to old saves with no memory of what quests I had completed. I think I spent a long time looking for black mushroom once when I had in fact already finished it.

julealgon: > I think this speaks more to a deficiency in the quest log, which doesn't currently have a "completed quests" section.

This feels like an obviously-good idea to me. Have you considered documenting the proposal as an issue?

julealgon:
LOL at the PR rename

Source/towners.cpp
```diff
@@ -742,6 +742,9 @@ void TalkToGirl(PlayerStruct &player, TownerStruct &girl)
 	case QUEST_ACTIVE:
 		InitQTextMsg(TEXT_GIRL3);
 		return;
+	case QUEST_DONE:
+		// do nothing
+		return;
 	default:
```
AJenbo, line 748:
```suggestion
	case QUEST_DONE:
		return;
	default:
```

Issue #2590, opened 2021-08-12
Farmer in hellfire repeatedly wonders about explosion


vladtepesch:
mportant information
win 10 64bit self build based on commit 0e56e52

Describe
Clicking the farmer after getting reward always replays the last text.

Expected behavior
Originally he does nothing after finishing quest.

AJenbo:
Isn't this a duplicate of https://github.com/diasurgical/devilution/issues/2247 ?

vladtepesch:
the other one is a `devilution` bug (without X).
There is a bug in both.
here is the `case 10` missing and the `case 3` quest state assignment correct. but in the other the wrong quest state is assigned, so that the case cannot be reached (in contrast to the actual original game)

PR #2592, opened 2021-08-12
Make farmer not wondering about explosion forever 

vladtepesch:
fixes https://github.com/diasurgical/devilutionX/issues/2590

### DIFF
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -630,6 +630,9 @@ void TalkToFarmer(PlayerStruct &player, TownerStruct &farmer)
 		if (gbIsMultiplayer)
 			NetSendCmdQuest(true, Q_FARMER);
 		break;
+	case QUEST_HIVE_DONE:
+		// do nothing
+		return;
 	default:
 		InitQTextMsg(TEXT_FARMER4);
 		break;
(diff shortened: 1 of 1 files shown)


Source/towners.cpp
```diff
@@ -630,6 +630,9 @@ void TalkToFarmer(PlayerStruct &player, TownerStruct &farmer)
 		if (gbIsMultiplayer)
 			NetSendCmdQuest(true, Q_FARMER);
 		break;
+	case QUEST_HIVE_DONE:
+		// do nothing
+		return;
```
AJenbo, line 635:
```suggestion
		break;
```

StephenCWills: The only place in this whole file where `return` is used in a switch is the `TalkToGirl()` function. Personally, I would very much prefer `break` than `return` here.

vladtepesch: Oh - I mislooked here - I interpreted as adding a `break` after `return` (which is enforeced by some code styles) - Sorry for confusion. I see - reakng here would do the same

AJenbo: Sounds like a terribly redundant code style since it would be unreachable code :D

AJenbo (approved):

Issue #2579, opened 2021-08-11
Sprites are missing for some spells


tsunamistate:
**Important information**

Windows 10 MSVC 2019 self-compilation from 09cbbc30cff596c2741085683b6b1ed4a05d84c6

**Describe**
Firebolt doesn't show the fireball sprite, but the spell still illuminates map

**To Reproduce**
1. Load a save
2. Switch the spell to Firebolt
3. Cast it

**Expected behavior**

**Video**

https://user-images.githubusercontent.com/4958520/129112891-52c4f105-bcd7-44e2-9aa5-896b6143b1a7.mp4

**Save file**
[single_0.zip](https://github.com/diasurgical/devilutionX/files/6971786/single_0.zip)

AJenbo:
@vlolteanu do you know what might be the issue here?

0mnicydle:
hmm could be related to the bug and commit I just created an issue regarding here:

https://github.com/diasurgical/devilutionX/issues/2584

b76e8c060b5415699fc8b0b6b869951b45aca7ce

Not much of a magic user on my current save otherwise I might have noticed some more things.

ThomasChr:
Broke in b5d67582955be0b30a55e1ad50271ff172c42fdc

vlolteanu:
I've figured out what's going on.

Making `AutofillArray(T x)` explicit is what introduced this. There's another constructor `template <typename... Xs> AutofillArray(const T &x, const Xs &...xs)` which now takes the place of the former when `Xs` is empty. The end-result is that calling it with a single arg zeroes out all elements aside from the first.

All spells with a `animFAmt` greater than 1 are affected by this. Firebolt only animates properly if you fire southwards; the first position corresponds to this direction.

julealgon:
@0mnicydle and @tsunamistate can you please work to rename this to a more generic name to match both the behaviors you are seeing?

AJenbo:
@vlolteanu monster missiles where still not being loaded, fixed here: https://github.com/diasurgical/devilutionX/pull/2596

PR #2593, opened 2021-08-12
Fix MissileFileData arrays not getting properly filled

vlolteanu:
Fixes https://github.com/diasurgical/devilutionX/issues/2579

I couldn't find a pretty way of calling explicit constructors in a brace-enclosed initializer list.


AJenbo:
Thanks, I had just gotten to the same conclusion :)

AJenbo (approved):

StephenCWills:
FYI, the last entry in the `MissileSpriteData` array gives me a couple run-time errors at startup in Visual Studio.
`Run-Time Check Failure #3 - The variable 'ret' is being used without being initialized.`

The exception occurs at the return statement in the `maybeAutofill()` function. Presumably, the reason this occurs is that it's operating on a list of size 0 and so no value ever gets assigned to the contents of the array. As far as I can tell, it's not actually causing any issues at runtime apart from the debugger stopping on the thrown exception.

vlolteanu:
You could try doing `std::array<T, 16> ret = {};`. It won't change the end-result, but maybe it will pacify Visual Studio.

StephenCWills:
Thanks, that does appear to pacify Visual Studio. I'll submit a PR.

PR #2598, opened 2021-08-14
Demo playback improvmentts

AJenbo:
- Use `+/-` to control playback speed
- Don't warp mouse while demo is playing

Besides that I recommend disabling hardware cursor for now when playing back demos to be able to see mouse moments.

### DIFF
--- a/Source/miniwin/misc_msg.cpp
+++ b/Source/miniwin/misc_msg.cpp
@@ -703,6 +703,16 @@ bool DemoMessage(tagMSG *lpMsg)
 			timedemo = false;
 			last_tick = SDL_GetTicks();
 		}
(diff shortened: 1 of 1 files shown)


obligaron (commented):
Looks good. 👍 
But now I have to rebase #2526 again 😢 

I also thought about disabling hardware cursor while demo playback.
This could be done in [OverrideOptions](https://github.com/diasurgical/devilutionX/pull/2526/files#diff-9180dcad034212a97d5acf1d39ec2ccc73cfeb0353b6102ef73b40a34aa2b6bfR147-R156). So we don't clutter options.cpp 🙂

obligaron (approved):

AJenbo:
Yeah, sorry about that. Considered not submitting it but it's just too useful when debugging a demo.

The reason I didn't disable the hardware curser is that it won't matter if we change the demo format... I guess that might have been true for the warping as well 🤔

PR #2601, opened 2021-08-14
Add debug text commands

obligaron:
### Intro

This PR introduces something like a debug `console`.
In debug mode you can write chat text and this will be analyzed by `CheckDebugTextCommand`.
For this change it was necessary to allow chat in singleplayer (only in debug).
Change is inspired by @qndel #2128. 🙂 

<details><summary>Example Usage Video</summary>

https://user-images.githubusercontent.com/25415264/129446094-1dba7f03-29e8-45f4-a1ce-2e2dfe195ff1.mp4

</details>

### Notes
- Works only in debug and is not in production/release version.
- This PR lays the groundwork to easily add more debug commands.
- With this PR only some commands are realized. This list can be expanded in followup PRs.
- Many commands that are available per debug text commands have currently hard coded keys (for example `lvl up` command). The discussion if we want to have this hard coded keys remaining or not is not part of this pr.
- This PR introduces `warp to lvl`, `reset lvl` and `godmode` toggle.
- Includes help command 😉 
- It's also possible to use `QuickMessage` and debug text commands together. So only one click for godmode. 😜

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -4,7 +4,10 @@
  * Implementation of debug functions.
  */
 
+#ifdef _DEBUG
+
 #include "cursor.h"
+#include "debug.h"
 #include "engine/cel_sprite.hpp"
 #include "engine/load_cel.hpp"
 #include "engine/point.hpp"
@@ -15,8 +18,6 @@
 namespace devilution {
 
 std::optional<CelSprite> pSquareCel;
-
-#ifdef _DEBUG
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


AJenbo:
Reset level seams particularly useful

Source/debug.cpp
```diff
@@ -58,6 +59,122 @@ void PrintDebugMonster(int m)
...
+	TakeGoldCheat();
+	return "You are poor...";
+}
+
+std::string DebugCmdWarpToLevel(const std::string_view parameter)
+{
+	auto &myPlayer = Players[MyPlayerId];
+	auto level = atoi(parameter.data());
+	if (level < 0 || level > (gbIsHellfire ? 24 : 16))
+		return fmt::format("Level {} is not known. Do you want to write a extension mod?", level);
```
qndel, line 119:
```suggestion
		return fmt::format("Level {} is not known. Do you want to write an extension mod?", level);
```
😆

Source/debug.cpp
```diff
@@ -58,6 +59,122 @@ void PrintDebugMonster(int m)
...
+	if (myPlayer.plrlevel == level)
+		return fmt::format("I did nothing but fulfilled you wish. You are already at level {}.", level);
```
qndel, line 121:
```suggestion
		return fmt::format("I did nothing but fulfilled your wish. You are already at level {}.", level);
```

Source/debug.cpp
obligaron: fixed 🙂

Source/debug.cpp
obligaron: fixed 🙂

qndel (changes requested):
Make commands in /cmd or /cmd<space><params> format
Remove text displayed after executing commands
No space in commands themselves 
Other than that, can't wait to merge and add more 🤤

AJenbo (approved):
No comment, well other then this being awesome :)

Being a Quake fan I might have a long term goal of having this work via a drop down terminal :D

julealgon:
Cool to see this idea added in. Obvious and useful improvement.

AJenbo (approved):
I partially addressed @qndel comment.

- Commands are now simulating one word with arguments (this way they will stay the same once we make them more flexible.
- They are now closer in name and style to what Quake uses

I did not prefix them with / for 2 reason
- Quake doesn't do this
- If we move to a debug console this won't be needed any way
- We do not map / in our crappy keyboard handeling so it would be impossible to type

julealgon:
We should make this separate from ingame chat at some point BTW.

AJenbo:
Another idea we can lift from Quake is to remove the application arguments and instead allow the application to take generic console commands as an argument:

`devilutionx +god` would start the game in god mode, this also removes a bunch of specialized code and coupling.

> We should make this separate from ingame chat at some point BTW.

Yep, see my other comments regarding a Quake style console :D

PR #2602, opened 2021-08-14
#2435 cleanup parameter in QuestStatus

bcko:
This pull request removes globals (Quests and QuestData) from QuestStatus function. 

part of #2435

### DIFF
--- a/Source/quests.h
+++ b/Source/quests.h
@@ -87,7 +87,7 @@ void InitQuests();
 void InitialiseQuestPools(uint32_t seed, QuestStruct quests[]);
 void CheckQuests();
 bool ForceQuests();
-bool QuestStatus(QuestStruct &quest, QuestDataStruct &questData);
+bool QuestStatus(int i);
 void CheckQuestKill(const MonsterStruct &monster, bool sendmsg);
 void DRLG_CheckQuests(int x, int y);
 void SetReturnLvlPos();
--- a/Source/quests.h
+++ b/Source/quests.h
@@ -12,6 +12,7 @@
 #include "engine/point.hpp"
 #include "gendung.h"
 #include "monster.h"
+#include "objdat.h"
 #include "textdat.h"
 #include "utils/stdcompat/optional.hpp"
 
@@ -87,7 +88,7 @@ void InitQuests();
 void InitialiseQuestPools(uint32_t seed, QuestStruct quests[]);
 void CheckQuests();
 bool ForceQuests();
-bool QuestStatus(int i);
+bool QuestStatus(quest_id questId);
 void CheckQuestKill(const MonsterStruct &monster, bool sendmsg);
 void DRLG_CheckQuests(int x, int y);
 void SetReturnLvlPos();
--- a/Source/quests.h
+++ b/Source/quests.h
@@ -87,7 +87,7 @@ void InitQuests();
 void InitialiseQuestPools(uint32_t seed, QuestStruct quests[]);
 void CheckQuests();
 bool ForceQuests();
-bool QuestStatus(int i);
+bool QuestStatus(QuestStruct &quest, QuestDataStruct &questData);
 void CheckQuestKill(const MonsterStruct &monster, bool sendmsg);
 void DRLG_CheckQuests(int x, int y);
 void SetReturnLvlPos();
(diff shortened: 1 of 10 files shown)


Source/quests.h
```diff
@@ -87,7 +87,7 @@ void InitQuests();
 void InitialiseQuestPools(uint32_t seed, QuestStruct quests[]);
 void CheckQuests();
 bool ForceQuests();
-bool QuestStatus(int i);
+bool QuestStatus(QuestStruct &quest, QuestDataStruct &questData);
```
AJenbo, line 90: I don't think this change really improves reliability/readability since the caller now has to make sure to get the two corresponding structs and correctly use the same enum/offset twice.

Instead what might make sens would be to have QuestStatus() take the enum instead of an int:
```suggestion
bool QuestStatus(quest_id &questId);
```

AJenbo (changes requested):
This seams like the wrong solution here.

ephphatha:
There's two relevant checks as far as I can tell which could both be member functions on `QuestStruct`. One is if the quest is available in the game ( true if `!setlevel`, `this->_qactive != NOT_AVAIL`, and the current game is a single player game if the `QuestDataStruct` indicates this quest is single player only), the other is if the quest has special objects that need to be generated in the current level (true if the quest is available and `this->_qlevel == currentLevel`).

The former check is effectively static when loading a game so could be baked in when initialising the `Quests` array, the second could take `currentLevel` as a parameter and will only need to reference its own properties otherwise.

bcko:
I will update the pull request incorporate your feedback.Thank you @AJenbo and @ephphatha !

Source/quests.h
bcko: I had to change from `quest_id &questId` to `quest_id questId` because I kept getting below error message and was not sure how to resolve other than changing.

```
error: cannot bind non-const lvalue reference of type ‘devilution::quest_id&’ to an rvalue of type ‘devilution::quest_id’
 4966 |  if (QuestStatus(Q_VEIL) && monster.mtalkmsg >= TEXT_VEIL9) {
```

AJenbo: sorry that was a typo on my side, simple value shouldn't be passed by reference :)

AJenbo (approved):

bcko:
Hi @ephphatha , I have been reading your feedback multiple times and read about initialization in c++ to incorporate your feedback. However, due to my gap in knowledge, I am unsure how to incorporate your feedback. :cry:

AJenbo:
> Hi @ephphatha , I have been reading your feedback multiple times and read about initialization in c++ to incorporate your feedback. However, due to my gap in knowledge, I am unsure how to incorporate your feedback. cry

Something like this:

Source/quests.h:
```cpp
struct QuestStruct {
	uint8_t _qlevel;
...
	bool _qlog;
	
	bool IsAvalible()
	{
		if (setlevel)
			return false;
		if (currlevel != _qlevel)
			return false;
		if (_qactive == QUEST_NOTAVAIL)
			return false;
		if (gbIsMultiplayer && QuestData[_qidx].isSinglePlayerOnly)
			return false;
		return true;
	}
};
```

This would allow you to replace:

`QuestStatus(Q_BETRAYER)` with `Quests[Q_BETRAYER].IsAvalible()`, wich will make a lot of sens when cleaning up other parts of the code that repeatedly makes use of the `Quests` global.


```cpp
auto &quest = Quests[Q_BETRAYER];
if (quest.IsAvalible())
    return;
    
quest._qactive = QUEST_ACTIVE
NetSendCmdQuest(true, quest._qidx)
```

AJenbo:
Also it seams to be possible to use a range-loop in `ForceQuests()`, which would make it more robust :
```cpp
bool ForceQuests()
{
...

	for (int i = 0; i < MAXQUESTS; i++) {
		if (i != Q_BETRAYER && currlevel == Quests[i]._qlevel && Quests[i]._qslvl != 0) {
			int ql = Quests[Quests[i]._qidx]._qslvl - 1;
			int qx = Quests[i].position.x;
			int qy = Quests[i].position.y;
...
		}
	}
...
}
```
->
```cpp
bool ForceQuests()
{
...
	for (auto &quest : Quests) {
		if (quest._qidx != Q_BETRAYER && currlevel == quest._qlevel && quest._qslvl != 0) {
			int ql = quest._qslvl - 1;
			int qx = quest.position.x;
			int qy = quest.position.y;
...
		}
	}
...
}
```

bcko:
@AJenbo thank you so much for taking your time to explain in more detail! I will open a new pull request to modify what you told me. :+1:

Issue #2107, opened 2021-06-01
[Bug] Affix on magical item morphs when left in town


oli25:
MP Mode

Magical items that is left in town has different affix when you go to catacombs and comeback.

1.Leave magical item on the ground
2.Go to catacombs
3.Comback to town
4.Affix on item has changed

Works once per mag item
Tested on items bought at Griswold
Tested on nighly build from few days ago and self compiled build

julealgon:
Can you provide a save the repros the issue please?

oli25:
Bug some magical item from Gris , drop it on the ground, go to cata and comeback.

[multi_0.zip](https://github.com/diasurgical/devilutionX/files/6580338/multi_0.zip)

qndel:
Can't reproduce, bought 
![image](https://user-images.githubusercontent.com/14297035/121814493-a9f97200-cc71-11eb-9721-813aebac0aa3.png) from gris, threw on the ground, went to catas and back, item is fine
![image](https://user-images.githubusercontent.com/14297035/121814503-baa9e800-cc71-11eb-93ca-f93e5ddec164.png)

oli25:
Weird, did you try multiple times?
Gonna try it on latest build maybe it got fixed accidentally.

oli25:
It's still there, try multiple times, also maybe move around abit in catas

qndel:
I thought it always happens, thx for extra info

PR #2606, opened 2021-08-15
fix shop items morphing

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2107

After this gets merged,
all
```cpp
			Items[x]._iSeed = AdvanceRndSeed();
			SetRndSeed(Items[x]._iSeed);
```
blocks should be turned into a function


AJenbo:
@obligaron just in case you haven't noticed, this change affects game RNG meaning that old demos may not replay correctly after this point. I don't consider this an issue, just an unfortunate consequence of the bug fix.

PR #2608, opened 2021-08-15
random looking direction of cows 🐄

vladtepesch:
Just a little town gimmick. 
Cows direction is randomized on town loading.
I found it strange that they are full animated for all directions but only the same 3 are shown at any time

### DIFF
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -220,15 +220,16 @@ void InitCows(TownerStruct &towner, const TownerInit &initData)
 	towner._tAnimWidth = 128;
 	towner.animOrder = nullptr;
 	towner.animOrderSize = 0;
+	auto dir = GenerateRnd(8);
 	for (int i = 0; i < 8; i++) {
 		towner._tNAnim[i] = CelGetFrame(CowCels.get(), i);
 	}
-	NewTownerAnim(towner, towner._tNAnim[initData.dir], 12, 3);
+	NewTownerAnim(towner, towner._tNAnim[dir], 12, 3);
 	towner._tAnimFrame = GenerateRnd(11) + 1;
 	towner.name = _("Cow");
 
 	const Point position = initData.position;
-	const Point offset = position + CowOffsets[initData.dir];
+	const Point offset = position + CowOffsets[dir];
 	int index = -dMonster[position.x][position.y];
 	if (dMonster[position.x][offset.y] == 0)
 		dMonster[position.x][offset.y] = index;


AJenbo:
image

julealgon:
We could also slightly randomize their spawn positions. Have you considered that?

julealgon:
Also, question: how does this work on multiplayer? Does every player see them consistently?

yuripourre:
> We could also slightly randomize their spawn positions. Have you considered that?

@julealgon yeah, maybe having an array of possible positions shouldn't be hard since there will be no need to calculate collisions.

@vladtepesch what do you think?

vladtepesch:
> We could also slightly randomize their spawn positions. Have you considered that?

then we also could have moore  cows  🐄🐄🐄🐄🐄🐄🐄🐄🐄🐄🐄 🤣

> Also, question: how does this work on multiplayer? Does every player see them consistently?

hm - good point - if not this would possibly get the players position out of sync.
But if both players have the same initial random seed of the town, they would get identical random results and thus identical positions. but on the other side if the town seed would be the same, both players would have the same items and wouldn't then the cows always get the same directions?

> @julealgon yeah, maybe having an array of possible positions shouldn't be hard since there will be no need to calculate collisions.
> 
> @vladtepesch what do you think?

love this idea - but I get the feeling that things are going to escalate 😁

julealgon:
> love this idea - but I get the feeling that things are going to escalate 😁

Not necessarily suggesting to do that in this same change. I consider them separate things but figured I'd throw the idea here since we don't have an issue.

Also, this might be a good opportunity to generalize some of the systems behind NPC placement. If you could softcode and parameterize some of the logic while working on this it would be really great.

Allowing one to randomize the position of cows might not be very valuable, but it might be useful later for positioning other NPCs, especially those that have walk animations.

vladtepesch:
> Also, question: how does this work on multiplayer? Does every player see them consistently?

I took a look at it in multiplayer and the results are a bit strange.

I one player leaves the town and reenters the cows are different (as expected - because the direction is created randomly).
If then the other player leaves and reenters town he then as the same orientations as the other player.

So it seems that both players share the same seed (obviously) but from this point on the towns go out of sync dependend on how often the players enter the town.

I have to confess that I still do not completely understand the distribution and handling of randomness within the game. It seems quite awkwart

However in this special case it seems that even with the different blocking locations the position do not go out of sync and the impact is quite minor but possible not worth the efford.

AJenbo:
> Also, this might be a good opportunity to generalize some of the systems behind NPC placement.

It's pretty general imo:
```cpp
const TownerInit TownerInitList[] = {
	// clang-format off
	// type         position    dir     init           talk
	{ TOWN_SMITH,   { 62, 63 }, DIR_SW, InitSmith,     TalkToBlackSmith  },
	{ TOWN_HEALER,  { 55, 79 }, DIR_SE, InitHealer,    TalkToHealer      },
	{ TOWN_DEADGUY, { 24, 32 }, DIR_N,  InitTownDead,  TalkToDeadguy     },
	{ TOWN_TAVERN,  { 55, 62 }, DIR_SW, InitBarOwner,  TalkToBarOwner    },
	{ TOWN_STORY,   { 62, 71 }, DIR_S,  InitTeller,    TalkToStoryteller },
	{ TOWN_DRUNK,   { 71, 84 }, DIR_S,  InitDrunk,     TalkToDrunk       },
	{ TOWN_WITCH,   { 80, 20 }, DIR_S,  InitWitch,     TalkToWitch       },
	{ TOWN_BMAID,   { 43, 66 }, DIR_S,  InitBarmaid,   TalkToBarmaid     },
	{ TOWN_PEGBOY,  { 11, 53 }, DIR_S,  InitBoy,       TalkToBoy         },
	{ TOWN_COW,     { 58, 16 }, DIR_SW, InitCows,      TalkToCow         },
	{ TOWN_COW,     { 56, 14 }, DIR_NW, InitCows,      TalkToCow         },
	{ TOWN_COW,     { 59, 20 }, DIR_N,  InitCows,      TalkToCow         },
	{ TOWN_COWFARM, { 61, 22 }, DIR_SW, InitCowFarmer, TalkToCowFarmer   },
	{ TOWN_FARMER,  { 62, 16 }, DIR_S,  InitFarmer,    TalkToFarmer      },
	{ TOWN_GIRL,    { 77, 43 }, DIR_S,  InitGirl,      TalkToGirl        },
	// clang-format on
};
```

julealgon:
> It's pretty general imo:

Not enough to support random placement and random direction selection it seems.

vladtepesch:
> It's pretty general imo:

I was wondering a bit, why not a lot more of the info of the init-functions is in there. for the most towners the init function could be generalized if this table would contain more initialization data.



But back to the multiplayer topic:
how does the sync work and what could possibly do?

julealgon:
> I was wondering a bit, why not a lot more of the info of the init-functions is in there. for the most towners the init function could be generalized if this table would contain more initialization data.

Probably a good idea.

> how does the sync work and what could possibly be done?

Not sure how it works but this is a tough question. Should the town be "the same" for all players? Should then vendors have the same shared items? When would they refresh, then? When everybody leaves town?

Surely NPC positions should be shared even if their items are not? Maybe we can add network commands for moving NPCs?

vladtepesch:
npcs stores always were (in valilla) somewhat strange. The smith mostly had the same items. Adria always had different  - even if you save/load next to her.
But my guess is that also griswoulds store would go out of sync if one player enters town much more frequently than another.

AJenbo:
There is no sync of town except for items and quest state.

AJenbo:
Closing as there are some potential issues with this and the benefits are minimal.

PR #2610, opened 2021-08-15
optimized packing/unpacking inventory items

qndel:


AJenbo:
You will have to create a new save and verify it manually for the test to pass.

PR #2612, opened 2021-08-15
2602 Use range-based for-loop for Quests

bcko:
Use range-based for-loop for Quests based on @AJenbo 's comment https://github.com/diasurgical/devilutionX/pull/2602#issuecomment-899109164

> Also it seams to be possible to use a range-loop in `ForceQuests()`, which would make it more robust :
> 
> ```c++
> bool ForceQuests()
> {
> ...
> 
> 	for (int i = 0; i < MAXQUESTS; i++) {
> 		if (i != Q_BETRAYER && currlevel == Quests[i]._qlevel && Quests[i]._qslvl != 0) {
> 			int ql = Quests[Quests[i]._qidx]._qslvl - 1;
> 			int qx = Quests[i].position.x;
> 			int qy = Quests[i].position.y;
> ...
> 		}
> 	}
> ...
> }
> ```
> 
> ->
> 
> ```c++
> bool ForceQuests()
> {
> ...
> 	for (auto &quest : Quests) {
> 		if (quest._qidx != Q_BETRAYER && currlevel == quest._qlevel && quest._qslvl != 0) {
> 			int ql = quest._qslvl - 1;
> 			int qx = quest.position.x;
> 			int qy = quest.position.y;
> ...
> 		}
> 	}
> ...
> }
> ```

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -276,7 +276,7 @@ void DeltaImportJunk(byte *src)
 
 	int q = 0;
 	for (auto &quest : Quests) {
-		if (!QuestData[quest._idx].isSinglePlayerOnly) {
+		if (!QuestData[quest._qidx].isSinglePlayerOnly) {
 			memcpy(&sgJunk.quests[q], src, sizeof(MultiQuests));
 			src += sizeof(MultiQuests);
 			quest._qlog = sgJunk.quests[q].qlog != 0;
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -240,11 +240,11 @@ byte *DeltaExportJunk(byte *dst)
 	}
 
 	int q = 0;
-	for (int i = 0; i < MAXQUESTS; i++) {
-		if (!QuestData[i].isSinglePlayerOnly) {
-			sgJunk.quests[q].qlog = Quests[i]._qlog ? 1 : 0;
-			sgJunk.quests[q].qstate = Quests[i]._qactive;
-			sgJunk.quests[q].qvar1 = Quests[i]._qvar1;
+	for (auto &quest : Quests) {
+		if (!QuestData[q].isSinglePlayerOnly) {
+			sgJunk.quests[q].qlog = quest._qlog ? 1 : 0;
+			sgJunk.quests[q].qstate = quest._qactive;
+			sgJunk.quests[q].qvar1 = quest._qvar1;
 			memcpy(dst, &sgJunk.quests[q], sizeof(MultiQuests));
 			dst += sizeof(MultiQuests);
 			q++;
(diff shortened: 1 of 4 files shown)


Source/msg.cpp
```diff
@@ -275,13 +275,13 @@ void DeltaImportJunk(byte *src)
 	}
 
 	int q = 0;
-	for (int i = 0; i < MAXQUESTS; i++) {
-		if (!QuestData[i].isSinglePlayerOnly) {
+	for (auto &quest : Quests) {
+		if (!QuestData[q].isSinglePlayerOnly) {
```
AJenbo, line 279: Since `q` is only incremented when `QuestData[i].isSinglePlayerOnly` is `false` we can't use it as the index as it will stop to match `i` the first time `isSinglePlayerOnly` is true.

Instead you could use `quest._idx`

bcko: Updated to use `quest._idx`. Thank you!

AJenbo: Note this is present in several places in this pr

bcko: Updated the pull request to use `quest._qidx` except in `void InitQuests()` where we have not initialized `quest._qidx` yet. Thank you!

Source/quests.cpp
```diff
@@ -269,10 +269,12 @@ void InitQuests()
...
 	} else {
-		for (int i = 0; i < MAXQUESTS; i++) {
-			if (QuestData[i].isSinglePlayerOnly) {
-				Quests[i]._qactive = QUEST_NOTAVAIL;
+		int q = 0;
+		for (auto &quest : Quests) {
+			if (QuestData[q].isSinglePlayerOnly) {
+				quest._qactive = QUEST_NOTAVAIL;
 			}
+			q++;
```
AJenbo, line 277:
```suggestion
		for (auto &quest : Quests) {
			if (QuestData[quest._qidx].isSinglePlayerOnly) {
				quest._qactive = QUEST_NOTAVAIL;
			}
```

Source/quests.cpp
```diff
@@ -283,31 +285,33 @@ void InitQuests()
...
-		Quests[i]._qlvltype = QuestData[i]._qlvlt;
-		Quests[i]._qvar2 = 0;
-		Quests[i]._qmsg = QuestData[i]._qdmsg;
+		quest._qslvl = QuestData[q]._qslvl;
+		quest.position = { 0, 0 };
+		quest._qidx = q;
+		quest._qlvltype = QuestData[q]._qlvlt;
+		quest._qvar2 = 0;
+		quest._qmsg = QuestData[q]._qdmsg;
+		q++;
```
AJenbo, line 314: This loop has a continue condition so this would not be executed and would miss-align in multiplayer.

AJenbo, line 314:
```suggestion
		quest._qlvltype = QuestData[q]._qlvlt;
		quest._qvar2 = 0;
		quest._qmsg = QuestData[q]._qdmsg;
```

AJenbo, line 314: This will break in multiplayer because of the continue at the start of the loop.

AJenbo (changes requested):
The changes here will break quests in multiplayer, please adjust the loop or revert that part.

Source/quests.cpp
bcko: Updated the pull request to adjust the loop. Thank you @AJenbo !

AJenbo (approved):
Nice work. I'm not super happy about the way InitQuests() looks still, but that can be refactored later.

PR #2613, opened 2021-08-16
Hellfire town interaction fixes

AJenbo:
This fixes a bunch of minor issues with town interaction for the Hellfire content in both single and multiplayer, some are vanilla and some where introduced (deliberately lazy clean up of towner code after 1.2.1).

- Some warps triggers where active when the graphics didn't indicate so and vice versa
- Lester will now leave to get food for his cows like he says (bug in master)
- Complete Nut now swaps his suit when given the brown suit (required town reload in vanilla)
- Complete Nut now say the correct dialog lines when given the gray suit (bug in master)
- Celia now hols Theodor when he is returned (required town reload in vanilla)
- Celia will go back home after quest is done (like she says, aka not spawn on next load, like the farmer)
- Celia will now say the correct dialog lines when given Theodor (bug in master)
- Hive require level 13 in multiplayer (same as cave, opening hive requires 15 or cave)
- Crypt is now open from the the start in multiplayer sessions
- Crypt requires level 17 in multiplayer (same as hell)

Peek 2021-08-15 17-36
image

Peek 2021-08-15 18-03


StephenCWills (approved):

julealgon:
@obligaron is there a way to retain the current animation frame when changing the animations like this? I notice for instance that the Celia change from normal to with_bear above is fairly harsh since the whole animation starts over. Was wondering if we could use something like was done for the character animations when changing armor sets, where the frame is preserved for smoothness.

AJenbo:
Hmm I don't think the two animations match up, not even sure if they have the same number of frames.

julealgon:
> Hmm I don't think the two animations match up, not even sure if they have the same number of frames.

At first glance they appeared similar enough. But looking again at the gif you might be right. 

If they really don't match at least in direction, then please disregard the suggestion.

Issue #2615, opened 2021-08-16
1.2.1 release fails to build with recent libsoduim snapshots


Polynomial-C:
Hi,

this a result of Gentoo downstream bug report https://bugs.gentoo.org/791031

first of all, I don't know why our libsodium package maintainer rolled a snapshot release. Perhaps because latest libsoduim release is from 2019 and there were some security issues that require a recent libsodium to get addressed but that's pure speculation from my side.
Fact is, recent libsodium breaks build of latest devilutionx release. I cannot test git HEAD right now due to unrelated issues that I first need to fix.
This is the build issue:

```
[...]

PR #2623, opened 2021-08-17
CMake: fix linking against libsodium statically

thesamesam:
Our previous processing would throw away all but the last
entry in the found STATIC_LIBRARIES, which in newer versions
of libsodium, might be "pthread" -- hence not trying to link
against libsodium at all.

We now generate a temporary list, mangle it, then return that.

Bug: https://bugs.gentoo.org/791031
Fixes: https://github.com/diasurgical/devilutionX/issues/2615
Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/CMake/Findsodium.cmake
+++ b/CMake/Findsodium.cmake
@@ -55,26 +55,26 @@ if (UNIX OR CMAKE_SYSTEM_NAME STREQUAL "Generic" OR AMIGA)
     endif()
 
     if(sodium_USE_STATIC_LIBS)
-    if(sodium_PKG_STATIC_LIBRARIES)
+      if(sodium_PKG_STATIC_LIBRARIES)
         foreach(_libname ${sodium_PKG_STATIC_LIBRARIES})
-            if (NOT _libname MATCHES "^lib.*\\.a$") # ignore strings already ending with .a
-                list(INSERT sodium_PKG_STATIC_LIBRARIES 0 "lib${_libname}.a")
-            endif()
+          if (NOT _libname MATCHES "^lib.*\\.a$") # ignore strings already ending with .a
+            list(INSERT sodium_PKG_STATIC_LIBRARIES 0 "lib${_libname}.a")
+          endif()
         endforeach()
         list(REMOVE_DUPLICATES sodium_PKG_STATIC_LIBRARIES)
-    else()
+      else()
         # if pkgconfig for libsodium doesn't provide
         # static lib info, then override PKG_STATIC here..
-            set(sodium_PKG_STATIC_LIBRARIES libsodium.a)
-        endif()
+        set(sodium_PKG_STATIC_LIBRARIES libsodium.a)
+      endif()
 
-        set(XPREFIX sodium_PKG_STATIC)
+      set(XPREFIX sodium_PKG_STATIC)
     else()
-    if(sodium_PKG_LIBRARIES STREQUAL "")
-            set(sodium_PKG_LIBRARIES sodium)
-        endif()
+      if(sodium_PKG_LIBRARIES STREQUAL "")
+        set(sodium_PKG_LIBRARIES sodium)
+      endif()
 
-        set(XPREFIX sodium_PKG)
+      set(XPREFIX sodium_PKG)
     endif()
 
     find_path(sodium_INCLUDE_DIR sodium.h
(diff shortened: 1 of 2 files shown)


thesamesam:
re CI failure: I'm not sure if devliutionX wants to support use of system libsodium (or any libraries) on Windows, but that's not my call to make, so I'll have it fall back to the old CMake module when pkgconfig can't be found for now.

AJenbo:
Thanks for looking in to this.

> re CI failure: I'm not sure if devilutionX wants to support use of system libsodium (or any libraries) on Windows, but that's not my call to make, so I'll have it fall back to the old CMake module when pkgconfig can't be found for now.

Sounds good, 2 of the 3 Windows builds are done via cross compiling where libsodium is simply extracted to a folder in the system path.

AJenbo:
> It's not clear why the CMake module started to fail,
> but it appears that it has a few bugs in how it handles
> fallbacks, and it wasn't ever official upstream anyway

It actually was, but they dropped it rather then maintain it.
https://github.com/jedisct1/libsodium/issues/929

CMakeLists.txt
```diff
@@ -212,8 +212,19 @@ endif()
...
 if(NOT NONET)
   if(DEVILUTIONX_SYSTEM_LIBSODIUM)
-    set(sodium_USE_STATIC_LIBS ${DEVILUTIONX_STATIC_LIBSODIUM})
-    find_package(sodium REQUIRED)
+    # We really need to use pkgconfig to find libsodium.
+    # They don't seem to support the CMake module method anymore and
+    # pkgconfig is simpler anyway.
+    # https://github.com/diasurgical/devilutionX/issues/2615
+    find_package(PkgConfig)
+    if(PkgConfig_FOUND)
```
glebm, line 220: This doesn't make sense in the main CMakeLists.txt. Findsodium.cmake should be adjusted instead.
As far as I can tell, it already tries pkgconfig first but perhaps it does so incorrectly.

glebm: Note that I mean our own Findsodium.cmake here: https://github.com/diasurgical/devilutionX/blob/master/CMake/Findsodium.cmake

We do not use the upstream (now deleted) Findsodium.cmake

thesamesam: Thank for the pointers! On it.

thesamesam:
@glebm @AJenbo Thanks for your help! Thanks to the advice, I've changed tactics a bit here. I think I've now found the "real" issue - the mangling of the list of static libraries was broken.

How're we looking now?

glebm (approved):
Looks good, thank you!

escalade:
Well, this seems to have broken linking against libsodium dynamically for me :)

```
CMake Error at /storage/RetroELEC/build.RetroELEC-skylake.x86_64-9.2-devel/toolchain/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
  Could NOT find sodium (missing: sodium_LIBRARY_RELEASE
  sodium_LIBRARY_DEBUG) (found version "1.0.18")
Call Stack (most recent call first):
  /storage/toolchain/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:582 (_FPHSA_FAILURE_MESSAGE)
  CMake/Findsodium.cmake:236 (find_package_handle_standard_args)
  CMakeLists.txt:216 (find_package)
```

From the error log:
```
[2/2] Linking C executable cmTC_07440
FAILED: cmTC_07440 
: && /storage/toolchain/bin/x86_64-libreelec-linux-gnu-gcc -march=skylake -m64 -mmmx -msse -msse2 -mfpmath=sse -fomit-frame-pointer -Wall -pipe  -O2 -Wl,--sort-common -Wl,--as-needed -fuse-ld=gold -Wl,-O2 CMakeFiles/cmTC_07440.dir/src.c.o -o cmTC_07440   && :
CMakeFiles/cmTC_07440.dir/src.c.o:src.c:function main: error: undefined reference to 'pthread_create'
CMakeFiles/cmTC_07440.dir/src.c.o:src.c:function main: error: undefined reference to 'pthread_detach'
CMakeFiles/cmTC_07440.dir/src.c.o:src.c:function main: error: undefined reference to 'pthread_cancel'
CMakeFiles/cmTC_07440.dir/src.c.o:src.c:function main: error: undefined reference to 'pthread_join'
ninja: build stopped: subcommand failed.
```

AJenbo:
@escalade are you also using gentoo?

thesamesam:
@escalade would you mind filing a new bug please and we’ll dig into it there? It’s be useful to have all the logs.

I can try setup which distro you’re using if needed.

escalade:
Compiling with -DCMAKE_BUILD_TYPE=MinSizeRel rather than "Release" solved my issue. My toolchain is a heavily customized LibreELEC toolchain, which is similar to buildroot.

thesamesam:
OK. I'm not sure why this would make a difference but if you hit it again, please do file a bug with all the details and I will happily look into it, of course! Glad it is OK now.

PR #2624, opened 2021-08-17
CI: enable verbose CMake builds

thesamesam:
Signed-off-by: Sam James <sam@gentoo.org>


AJenbo:
CI permitted

AJenbo:
You might also want to do it for https://github.com/diasurgical/devilutionX/blob/master/.github/workflows/MacOSX.yml since that is the one that is currently failing in your PR.

PR #2626, opened 2021-08-18
2602 add IsAvailable in QuestStruct and remove QuestStatus()

bcko:
this pull request is related to below feedback 
@ephphatha https://github.com/diasurgical/devilutionX/pull/2602#issuecomment-899023688
@AJenbo https://github.com/diasurgical/devilutionX/pull/2602#issuecomment-899108179

### DIFF
--- a/Source/quests.cpp
+++ b/Source/quests.cpp
@@ -357,7 +357,7 @@ void CheckQuests()
 	if (gbIsSpawn)
 		return;
 
-	if (QuestStatus(Q_BETRAYER) && gbIsMultiplayer && Quests[Q_BETRAYER]._qvar1 == 2) {
+	if (Quests[Q_BETRAYER].IsAvailable() && gbIsMultiplayer && Quests[Q_BETRAYER]._qvar1 == 2) {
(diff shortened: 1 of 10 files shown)


Source/quests.cpp
```diff
@@ -829,4 +816,18 @@ void SetMultiQuest(int q, quest_state s, bool log, int v1)
...
+bool QuestStruct::IsAvailable()
+{
+	if (setlevel)
+		return false;
+	if (currlevel != _qlevel)
+		return false;
+	if (_qactive == QUEST_NOTAVAIL)
+		return false;
+	if (gbIsMultiplayer && QuestData[_qidx].isSinglePlayerOnly)
+		return false;
```
AJenbo, line 828: Reading over @ephphatha comment again I see that he also pointed out that this part could be determined when quests are initialized (or loaded), as a new parameter, which would allow IsAvailable() to be simplified a bit and moved to the header as a `cosnt` function.

This would be a decent improvement to do at a later point.

AJenbo (dismissed):

AJenbo (approved):

PR #2627, opened 2021-08-18
fix displayed save number

qndel:
image
Displays actual save number the new created character is going to have instead of 99


ThomasChr (approved):

AJenbo (approved):

julealgon:
A bit surprised at how fast this was merged without further discussion. I think it is quite arguable whether or not this makes sense (the label is not semantic enough to a new player), and the alignment in the screenshot looks really bad.

Would've been nice to at least polish it a bit more before moving forward with it imho.

StephenCWills:
@julealgon This is just a fix. I believe the save number only displays for debug builds. See: #2386

PR #2628, opened 2021-08-18
Get rid of InitHash

vlolteanu:


AJenbo (approved):
Complex but better :)

PR #2634, opened 2021-08-18
2435 refactor Missiles part 1

bcko:
- related to #2435 
- part 1 of refactor Missiles
- remove global variable `Missiles` in function `UpdateMissileVelocity`

PR #2635, opened 2021-08-19
2435 refactor missiles part 2

bcko:
related to #2435 

Based on @AJenbo comment in https://github.com/diasurgical/devilutionX/pull/2634#issuecomment-901493966

```
UpdateMissilePos()
SetMissAnim()
SetMissDir()
SetMissAnim()
MissilesFoundTarget()
UpdateVileMissPos()
AddLightningWall()
```

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -607,21 +607,22 @@ void CheckMissileCol(int i, int mindam, int maxdam, bool shift, Point position, 
 
 void SetMissAnim(int mi, int animtype)
 {
-	int dir = Missiles[mi]._mimfnum;
+	auto &missile = Missiles[mi];
+	int dir = missile._mimfnum;
 
 	if (animtype > MFILE_NONE) {
 		animtype = MFILE_NONE;
 	}
 
-	Missiles[mi]._miAnimType = animtype;
-	Missiles[mi]._miAnimFlags = MissileSpriteData[animtype].flags;
-	Missiles[mi]._miAnimData = MissileSpriteData[animtype].animData[dir].get();
-	Missiles[mi]._miAnimDelay = MissileSpriteData[animtype].animDelay[dir];
-	Missiles[mi]._miAnimLen = MissileSpriteData[animtype].animLen[dir];
-	Missiles[mi]._miAnimWidth = MissileSpriteData[animtype].animWidth;
-	Missiles[mi]._miAnimWidth2 = MissileSpriteData[animtype].animWidth2;
-	Missiles[mi]._miAnimCnt = 0;
-	Missiles[mi]._miAnimFrame = 1;
+	missile._miAnimType = animtype;
+	missile._miAnimFlags = MissileSpriteData[animtype].flags;
+	missile._miAnimData = MissileSpriteData[animtype].animData[dir].get();
+	missile._miAnimDelay = MissileSpriteData[animtype].animDelay[dir];
+	missile._miAnimLen = MissileSpriteData[animtype].animLen[dir];
+	missile._miAnimWidth = MissileSpriteData[animtype].animWidth;
+	missile._miAnimWidth2 = MissileSpriteData[animtype].animWidth2;
+	missile._miAnimCnt = 0;
+	missile._miAnimFrame = 1;
 }
 
 bool MissilesFoundTarget(int mi, Point *position, int rad)
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2935,7 +2935,6 @@ void AddDiabApoca(int mi, Point src, Point /*dst*/, int /*midir*/, int8_t mienem
 
 int AddMissile(Point src, Point dst, int midir, int mitype, int8_t micaster, int id, int midam, int spllvl)
 {
-
 	if (ActiveMissileCount >= MAXMISSILES - 1)
 		return -1;
 
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2961,7 +2961,7 @@ int AddMissile(Point src, Point dst, int midir, int mitype, int8_t micaster, int
 	ActiveMissiles[ActiveMissileCount] = mi;
 	ActiveMissileCount++;
 
-	memset(&Missiles[mi], 0, sizeof(*Missiles));
+	memset(&missile, 0, sizeof(missile));
 
 	missile._mitype = mitype;
 	missile._micaster = micaster;
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -948,8 +948,9 @@ Direction16 GetDirection16(Point p1, Point p2)
 
 void DeleteMissile(int mi, int i)
 {
-	if (Missiles[mi]._mitype == MIS_MANASHIELD) {
-		int src = Missiles[mi]._misource;
+	auto &missile = Missiles[mi];
+	if (missile._mitype == MIS_MANASHIELD) {
+		int src = missile._misource;
 		if (src == MyPlayerId)
 			NetSendCmd(true, CMD_REMSHIELD);
 		Players[src].pManaShield = false;
@@ -1222,8 +1223,9 @@ void InitMissiles()
 	if (myPlayer._pInfraFlag) {
 		for (int i = 0; i < ActiveMissileCount; ++i) {
 			int mi = ActiveMissiles[i];
-			if (Missiles[mi]._mitype == MIS_INFRA) {
-				int src = Missiles[mi]._misource;
+			auto &missile = Missiles[mi];
+			if (missile._mitype == MIS_INFRA) {
+				int src = missile._misource;
 				if (src == MyPlayerId)
 					CalcPlrItemVals(MyPlayerId, true);
 			}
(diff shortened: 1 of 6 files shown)


Source/missiles.cpp
```diff
@@ -1261,28 +1267,30 @@ void InitMissiles()
 
 void AddHiveExplosion(int mi, Point /*src*/, Point /*dst*/, int midir, int8_t mienemy, int id, int dam)
 {
+	auto &missile = Missiles[mi];
```
ephphatha, line 1270: Ordinarily I'd suggest declaring variables close to where they are being used as this matches the style guide, but in this case having it right where the function begins helps with the eventual refactoring of the `AddMissile` class of functions.

Source/missiles.cpp
```diff
@@ -1414,6 +1426,7 @@ void AddHorkSpawn(int mi, Point src, Point dst, int midir, int8_t /*mienemy*/, i
 void AddJester(int mi, Point src, Point dst, int midir, int8_t /*mienemy*/, int id, int /*dam*/)
 {
 	int spell = MIS_FIREBOLT;
+	auto &missile = Missiles[mi];
```
ephphatha, line 1429: This should be the first variable declared (move above `int spell`) since the switch statement is used to fully initialise `spell`;

Source/missiles.cpp
```diff
@@ -1605,9 +1621,10 @@ void AddWarp(int mi, Point src, Point /*dst*/, int /*midir*/, int8_t mienemy, in
 			}
 		}
 	}
-	Missiles[mi]._mirange = 2;
-	Missiles[mi]._miVar1 = 0;
-	Missiles[mi].position.tile = tile;
+	auto &missile = Missiles[mi];
```
ephphatha, line 1624: For consistency moving this to the start of the function would be good.

Source/missiles.cpp
```diff
@@ -3225,56 +3287,59 @@ void MI_Acidpud(int i)
...
+		missile._miVar2++;
 	}
 	PutMissile(i);
 }
 
 void MI_Fireball(int i)
 {
-	FireballUpdate(i, Missiles[i].position.velocity, false);
+	auto &missile = Missiles[i];
+	FireballUpdate(i, missile.position.velocity, false);
```
ephphatha, line 3319: It would be good to have FireballUpdate take `MissileStruct&` as the first argument to make this change worthwhile. The current use with a call to PutMissile at the end of that function makes it difficult to do this refactoring now though so happy for that to come later.

Source/missiles.cpp
```diff
@@ -4566,14 +4668,15 @@ void missiles_process_charge()
...
 	}
 }
 
 void ClearMissileSpot(int mi)
 {
-	dFlags[Missiles[mi].position.tile.x][Missiles[mi].position.tile.y] &= ~BFLAG_MISSILE;
-	dMissile[Missiles[mi].position.tile.x][Missiles[mi].position.tile.y] = 0;
+	auto &missile = Missiles[mi];
+	dFlags[missile.position.tile.x][missile.position.tile.y] &= ~BFLAG_MISSILE;
+	dMissile[missile.position.tile.x][missile.position.tile.y] = 0;
```
ephphatha, line 4679:
```suggestion
	auto &missileTile = Missiles[mi].position.tile;
	dFlags[missileTile.x][missileTile.y] &= ~BFLAG_MISSILE;
	dMissile[missileTile.x][missileTile.y] = 0;
```
This only needs the dungeon tile, can go deeper into the struct immediately.

The function could take a Point as the argument instead of an index as well, it's called from a context where we've already accessed the relevant item in the Missiles array.

Source/missiles.cpp
```diff
@@ -4550,6 +4651,7 @@ void missiles_process_charge()
 {
 	for (int i = 0; i < ActiveMissileCount; i++) {
 		int mi = ActiveMissiles[i];
+		auto &missile = Missiles[mi];
 		MissileStruct *mis = &Missiles[mi];
```
ephphatha, line 4655: Replace the uses of `mis->` with `missile.` and you can get rid of this pointer variable.

ephphatha (changes requested):
There's a few places which could be cleaned up further to remove duplication and/or make the code clearer.

I haven't marked every instance of the "AddMissileProc" callbacks where `MissileStruct &missile` is declared at the start/in the middle of the function, there are multiple places where it could be reordered if you wanted and there's consensus from the maintainers. I've suggested using declaration at the start of the function even though that's a deliberate violation of the style guide as I think would be helpful for future refactoring.

Source/missiles.cpp
```diff
@@ -435,154 +437,155 @@ void CheckMissileCol(int i, int mindam, int maxdam, bool shift, Point position,
 
 	int mx = position.x;
 	int my = position.y;
+	auto &missile = Missiles[i];
```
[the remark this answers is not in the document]
AJenbo: Indeed with `SetMissDir()` also taking `MissileStruct` as input `CheckMissileCol()` can be made to do so as well and the change check for `i` can be removed (the check was pointless any was as all callers utilized `i` before calling).

Source/missiles.cpp
AJenbo: Agreed, I prefer it for that reason as well, it helps to highlight what globals are being used by the function.

Source/missiles.cpp
AJenbo: Yeah, I think this should be deferred til round 4 or 5 :)

Source/portal.cpp
```diff
@@ -52,12 +52,13 @@ void AddWarpMissile(int i, int x, int y)
 	MissileData[MIS_TOWN].mlSFX = SFX_NONE;
 	dMissile[x][y] = 0;
 	mi = AddMissile({ 0, 0 }, { x, y }, 0, MIS_TOWN, TARGET_MONSTERS, i, 0, 0);
+	auto &missile = Missiles[mi];
 
 	if (mi != -1) {
```
AJenbo, line 57: First make sure that `mi` is valid.
```suggestion
	if (mi != -1) {
	auto &missile = Missiles[mi];

```

Source/missiles.cpp
```diff
@@ -2893,50 +2947,52 @@ int AddMissile(Point src, Point dst, int midir, int mitype, int8_t micaster, int
...
 	}
 
 	int mi = AvailableMissiles[0];
+	auto &missile = Missiles[mi];
 
 	AvailableMissiles[0] = AvailableMissiles[MAXMISSILES - ActiveMissileCount - 1];
 	ActiveMissiles[ActiveMissileCount] = mi;
 	ActiveMissileCount++;
 
 	memset(&Missiles[mi], 0, sizeof(*Missiles));
```
AJenbo, line 2964:
```suggestion
	memset(&missile, 0, sizeof(*Missiles));

```

Source/missiles.cpp
```diff
@@ -2877,11 +2930,12 @@ void AddDiabApoca(int mi, Point src, Point /*dst*/, int /*midir*/, int8_t mienem
 
 		AddMissile({ 0, 0 }, player.position.future, 0, MIS_BOOM2, mienemy, id, dam, 0);
 	}
-	Missiles[mi]._miDelFlag = true;
+	missile._miDelFlag = true;
 }
 
 int AddMissile(Point src, Point dst, int midir, int mitype, int8_t micaster, int id, int midam, int spllvl)
 {
+
```
AJenbo, line 2938:
```suggestion
```

Source/missiles.cpp
```diff
@@ -2591,79 +2630,84 @@ void AddFirewallC(int mi, Point src, Point dst, int midir, int8_t /*mienemy*/, i
...
 	CalcPlrItemVals(id, true);
 	force_redraw = 255;
 	player.Say(HeroSpeech::Aaaaargh);
 }
 
 void AddRepair(int mi, Point /*src*/, Point /*dst*/, int /*midir*/, int8_t /*mienemy*/, int id, int /*dam*/)
 {
-	Missiles[mi]._miDelFlag = true;
+	auto &missile = Missiles[mi];
+	missile._miDelFlag = true;
```
AJenbo, line 2710: I normally wouldn't do this chance if there was only one usage, but in the case of `void Add*(int mi,` we want to change them to take `MissileStruct` as the input, so it can server as prep for that still.

AJenbo (commented):
Big clean up here, glad you kept it to about 1000 lines as it start to get hard to review at that point :)

bcko:
@ephphatha & @AJenbo Thank you so much for all your feedback! I updated the pull request to incorporate all your feedback.

Source/missiles.cpp
bcko: @AJenbo , for this one I changed `sizeof(*Missiles)` to `sizeof(missile)` as suggested by @ephphatha in Discord development channel. 

```c++
memset(&missile, 0, sizeof(missile));
```

bcko:
> There's a few places which could be cleaned up further to remove duplication and/or make the code clearer.
> 
> I haven't marked every instance of the "AddMissileProc" callbacks where `MissileStruct &missile` is declared at the start/in the middle of the function, there are multiple places where it could be reordered if you wanted and there's consensus from the maintainers. I've suggested using declaration at the start of the function even though that's a deliberate violation of the style guide as I think would be helpful for future refactoring.

Updated the pull request to move `MissileStruct &missile` at the start of the function! thank you so much

Source/missiles.cpp
```diff
@@ -4566,14 +4662,15 @@ void missiles_process_charge()
 		} else {
 			graphic = MonsterGraphic::Walk;
 		}
-		Missiles[mi]._miAnimData = mon->GetAnimData(graphic).CelSpritesForDirections[mis->_mimfnum]->Data();
+		missile._miAnimData = mon->GetAnimData(graphic).CelSpritesForDirections[missile._mimfnum]->Data();
 	}
 }
 
 void ClearMissileSpot(int mi)
```
AJenbo, line 4669: For part 3 then:
```suggestion
void ClearMissileSpot(MissileStruct &missile)
```

AJenbo (approved):

PR #2637, opened 2021-08-19
Debugmapcmd

AJenbo:
- Clean up god mode (the debug command was actually switching god+tools)
- Fix changing to regular levels when in a quest map
- Add command for loading quest maps
- Remove arguments that are now available as commands
- Remove chat buttons, it was largely hardcoded to specific keys and not easy to understand or remember

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -103,13 +103,41 @@ std::string DebugCmdHelp(const std::string_view parameter)
 
 std::string DebugCmdGiveGoldCheat(const std::string_view parameter)
 {
-	GiveGoldCheat();
+	auto &myPlayer = Players[MyPlayerId];
+
+	for (int8_t &itemId : myPlayer.InvGrid) {
+		if (itemId != 0)
+			continue;
+
+		int ni = myPlayer._pNumInv++;
+		SetPlrHandItem(&myPlayer.InvList[ni], IDI_GOLD);
+		GetPlrHandSeed(&myPlayer.InvList[ni]);
+		myPlayer.InvList[ni]._ivalue = GOLD_MAX_LIMIT;
+		myPlayer.InvList[ni]._iCurs = ICURS_GOLD_LARGE;
+		myPlayer._pGold += GOLD_MAX_LIMIT;
+		itemId = myPlayer._pNumInv;
+	}
+
 	return "You are now rich! If only this was as easy in real life...";
 }
 
 std::string DebugCmdTakeGoldCheat(const std::string_view parameter)
 {
-	TakeGoldCheat();
+	auto &myPlayer = Players[MyPlayerId];
+
+	for (auto itemId : myPlayer.InvGrid) {
+		itemId -= 1;
+
+		if (itemId < 0)
+			continue;
+		if (myPlayer.InvList[itemId]._itype != ITYPE_GOLD)
+			continue;
+
+		myPlayer.RemoveInvItem(itemId);
+	}
+
+	myPlayer._pGold = 0;
+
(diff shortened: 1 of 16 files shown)


Source/debug.cpp
```diff
@@ -101,13 +106,41 @@ std::string DebugCmdHelp(const std::string_view parameter)
...
 	return "You are now rich! If only this was as easy in real life...";
 }
 
 std::string DebugCmdTakeGoldCheat(const std::string_view parameter)
 {
-	TakeGoldCheat();
+	auto &myPlayer = Players[MyPlayerId];
+
+	for (auto itemId : myPlayer.InvGrid) {
+		itemId -= 1;
```
qndel, line 132: how is this supposed to work?

Source/debug.cpp
```diff
@@ -137,106 +194,54 @@ std::string DebugCmdResetLevel(const std::string_view parameter)
...
 }
 
-std::string DebugCmdMaxSpellLevel(const std::string_view parameter)
+std::string DebugCmdVision(const std::string_view parameter)
 {
-	MaxSpellsCheat();
-	return "Knowledge is power.";
+	DebugVision = !DebugVision;
+	if (DebugVision)
+		return "You see as i do.";
```
qndel, line 214:
```suggestion
		return "You see as I do.";

```

Source/missiles.cpp
```diff
@@ -264,13 +267,12 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, int t, bool
...
 		return ret;
 
+	if (hit >= hper) {
 #ifdef _DEBUG
-	if (hit >= hper && !debug_mode_key_inverted_v && !debug_mode_dollar_sign)
-		return false;
-#else
-	if (hit >= hper)
-		return false;
+		if (!DebugGodMode)
```
qndel, line 272: I hate this identation so much, can't we deal with it somehow? 😅

Source/debug.cpp
```diff
@@ -260,6 +265,50 @@ void SetAllSpellsCheat()
...
+
+	return "Knowledge is power.";
+}
+
+std::vector<DebugCmdItem> DebugCmdList = {
+	{ "help", "Prints help overview or help for a specific command.", "({command})", &DebugCmdHelp },
+	{ "give gold", "Fills the inventory with gold.", "", &DebugCmdGiveGoldCheat },
+	{ "give xp", "Levels the player up (min 1 level or {levels}).", "({levels})", &DebugCmdLevelUp },
+	{ "give spells", "Add all spells to player.", "", &DebugCmdGetAllSpells },
+	{ "give spells 10", "Set spell level to 10 for all spells.", "", &DebugCmdMaxSpellLevel },
```
qndel, line 291: give spells takes priority over give spells 10, so it's impossible to execute give spells 10, merge these 2 commands into one that takes spell level as an argument

Source/missiles.cpp
AJenbo: So do I, but it's a step up from the duplicate

Source/debug.cpp
AJenbo: Good point, but it's like that currently so shouldn't stop this pr

Source/debug.cpp
AJenbo: Idk, it was like this when I found it, just moved the code

qndel (approved):
👍

obligaron:
Great to see more debug commands and the cleanup to the existing logic is nice. 👍 🙂

PR #2642, opened 2021-08-19
tweak debug spells command

qndel:
unifies 2 commands into one and lets you unlearn all spells by using the command with 0 lvl


Source/debug.cpp
```diff
@@ -288,8 +263,7 @@ std::vector<DebugCmdItem> DebugCmdList = {
 	{ "help", "Prints help overview or help for a specific command.", "({command})", &DebugCmdHelp },
 	{ "give gold", "Fills the inventory with gold.", "", &DebugCmdGiveGoldCheat },
 	{ "give xp", "Levels the player up (min 1 level or {levels}).", "({levels})", &DebugCmdLevelUp },
-	{ "give spells", "Add all spells to player.", "", &DebugCmdGetAllSpells },
-	{ "give spells 10", "Set spell level to 10 for all spells.", "", &DebugCmdMaxSpellLevel },
+	{ "set spells", "Set spell level to {level} for all spells.", "{level}", &DebugCmdSetSpellsLevel },
```
AJenbo, line 266: I would prefer that we follow the norm of `give spells`, `take spells`

Source/debug.cpp
```diff
@@ -242,44 +242,19 @@ std::string DebugCmdLevelUp(const std::string_view parameter)
...
+	int level = std::max(0, atoi(parameter.data()));
 	for (int i = SPL_FIREBOLT; i < MAX_SPELLS; i++) {
 		if (GetSpellBookLevel((spell_id)i) != -1) {
 			myPlayer._pMemSpells |= GetSpellBitmask(i);
-			myPlayer._pSplLvl[i] = 10;
+			myPlayer._pSplLvl[i] = level;
 		}
 	}
+	if (level == 0)
+		myPlayer._pMemSpells = 0;
```
AJenbo, line 257: Would prefer a separate command for removing spells to keep things more consistent.
```suggestion
			myPlayer._pSplLvl[i] += level;
		}
	}
```

Source/debug.cpp
qndel: what's the point if you can do it with one just fine?

Source/debug.cpp
qndel: I see no reason to artificially split it into 2 commands when setting spell level to 0 would be the logical way to remove spells

AJenbo: You could give +1 to all spells to test something, then take away -1 to return to the original spells to not affect potential save games or further tests.

Source/debug.cpp
AJenbo: A more generic interface to commands, making it easier to get an overview and discover options.
Some of this won't really take effect before we get a proper command prompt.

julealgon:
My understanding is that the way we are coding these, their effects won't be propagated in multiplayer.

Can we change these commands to actually send high-level commands to the game so that they are processed as if they were in-game generated, and thus properly reflect on multiplayer as well?

I'm not necessarily advocating for "cheating in multiplayer" or anything of the sort, but I think converting them into high level commands has several benefits:
- allows us to test things while in local multiplayer games
- prevents code duplication by ensuring the actual change is performed once by the engine, regardless if it is a cheat or done normally
- encourage the creation of more valid network commands that can be used for modding
- encourage parametrization of methods when needed to support the cheats, which can then also be used for modding

Making direct changes like we are doing is a bad approach, IMHO.

AJenbo:
> Can we change these commands to actually send high-level commands to the game so that they are processed as if they were in-game generated, and thus properly reflect on multiplayer as well?

Sure thing, but boy is it out of scope for this PR :D

julealgon:
> Sure thing, but boy is it out of scope for this PR :D

Sure... I meant more as something for us to think about before we have 1 billion such commands added.

The more we keep doing these "hacks", the harder will be to migrate them to something more reusable and compatible later.

> Most of this is inherited from the Diablo 1.00 debug build

I actually did not know that, thanks for clarifying. I guess it made total sense for them considering multiplayer didn't even exist (so I can imagine in the beginning there weren't even in-game commands at all) and also because they had no intention of exposing any of it to anyone (i.e. modders).

AJenbo:
Yeah new ones should be done properly, but this PR is just getting rid of one of the old ones, so no harm :)

Source/debug.cpp
```diff
@@ -242,44 +242,19 @@ std::string DebugCmdLevelUp(const std::string_view parameter)
 	return "New experience leads to new insights.";
 }
 
-std::string DebugCmdGetAllSpells(const std::string_view parameter)
-{
-	SetSpellLevelCheat(SPL_FIREBOLT, 8);
```
AJenbo, line 247: `SetSpellLevelCheat()` is unused now and should be fully removed.

qndel: oh I can actually reuse it : D

Source/debug.cpp
qndel: meh, not really, spells are capped at 15, so lets say you gave yourself 15, there's no going back :D

ThomasChr:
Made a short test: Works fine!

qndel:
@AJenbo if you still feel we'd need an extra command, feel free to add it but I see no harm in having this one be as complete as it is now

Issue #2632, opened 2021-08-18
Reflect spell is no longer active after loading the game


Chance4us:
Self-explanatory.
To reproduce:
1. Play a game
2. Read the book of reflect
3. Cast Reflect spell and save the game
4. Leave the game to stop process
5. Open the game binary and load the saved game
6. Be surprised that Reflect is no longer activated

EDIT: 4. Exit the binary is not needed, it is sufficient just to load the game while playing

ThomasChr:
could confirm on windows e1db9ba77b06500e43bc49f310323758e67ba743 - works with manashield, does not work with reflect

ThomasChr:
Hey! That does seem to work great!
Unfortunately it also leaves the reflect spell active even if you are starting a new game. And that does not seem quite right to me!

Happy litle sorcerer starts with reflect:
![grafik](https://user-images.githubusercontent.com/13014163/130191263-1c5795c3-1136-4ef2-840c-150ce7475957.png)

qndel:
fixed

ThomasChr:
checked, and it works. Haven't doubt that for a second!

PR #2643, opened 2021-08-19
fix reflect on loading

qndel:
Fixes https://github.com/diasurgical/devilutionX/issues/2632
Manashield and reflect should get a full rewrite at some point, current code is beyond garbage, that's why it needs to be patched out in seemingly random places :D

I made reflect replicate manashield logic.


AJenbo (approved):
Pretty sure I would reject most of this if it wasn't copied from the manashild code. I do have a nagging feeling that a bunch of it is however redundant and could be done with some shared function, but we can look at that later.

qndel:
Yeah that's why I'm saying it should be rewritten anyway - tbh old reflect's code seemed to work just fine, it only had 1 flaw - always removing reflect on InitMissiles() which also happened on reloading the game so I guess a quick and dirty fix would be just letting that function know if you reloaded the game, if yes, don't remove reflect :D 

Also I'm suspecting that infravision is getting removed on load as well

julealgon:
Holy shit at this change! this code man... 🤦‍♂️

qndel:
> Holy shit at this change! this code man... 🤦‍♂️

I mean old reflect probably worked just fine, it'd only require a dirty hack to stop deleting it on loading, but since manashield works fine, I just duplicated it

PR #2644, opened 2021-08-19
Use loops to spawn needed golems

AJenbo:


qndel (approved):
👍

Source/monster.cpp
```diff
@@ -3831,10 +3831,8 @@ void monster_some_crypt()
 void InitMonsters()
 {
 	if (!setlevel) {
-		AddMonster(GolemHoldingCell, DIR_S, 0, false);
-		AddMonster(GolemHoldingCell, DIR_S, 0, false);
-		AddMonster(GolemHoldingCell, DIR_S, 0, false);
-		AddMonster(GolemHoldingCell, DIR_S, 0, false);
+		for (int j = 0; j < MAX_PLRS; j++)
```
julealgon, line 3834: Can't we use something like `for (auto _ : players)` here and avoid the extra dependency with `MAX_PLRS`?

Source/monster.cpp
```diff
@@ -3897,10 +3895,9 @@ void InitMonsters()
 void SetMapMonsters(const uint16_t *dunData, Point startPosition)
 {
 	AddMonsterType(MT_GOLEM, PLACE_SPECIAL);
-	AddMonster(GolemHoldingCell, DIR_S, 0, false);
-	AddMonster(GolemHoldingCell, DIR_S, 0, false);
-	AddMonster(GolemHoldingCell, DIR_S, 0, false);
-	AddMonster(GolemHoldingCell, DIR_S, 0, false);
+	for (int k = 0; k < MAX_PLRS; k++)
```
julealgon, line 3898: The use of `j` and then `k` to represent the same thing in the same commit looks weird to me.

Source/monster.cpp
AJenbo: That seams really missleading

Source/monster.cpp
AJenbo: Fixed

Source/monster.cpp
julealgon: I don't know. I think the `MAX_PLRS` macro should not exist.

AJenbo: Not disagreeing, but that is a refactor for a later day :) Just pulled this out of another PR so that it could get closed.

PR #2646, opened 2021-08-19
Clean up effect code

AJenbo:
A follow up to https://github.com/diasurgical/devilutionX/pull/2643 for getting rid of some of the duplicate code


Source/player.cpp
```diff
@@ -3215,18 +3215,11 @@ void RemovePlrMissiles(int pnum)
...
-		if (Missiles[am]._mitype == MIS_MANASHIELD && Missiles[am]._misource == pnum) {
-			ClearMissileSpot(am);
-			DeleteMissile(am, i);
-		}
-		if (Missiles[am]._mitype == MIS_REFLECT && Missiles[am]._misource == pnum) {
-			ClearMissileSpot(am);
-			DeleteMissile(am, i);
-		}
-		if (Missiles[am]._mitype == MIS_ETHEREALIZE && Missiles[am]._misource == pnum) {
+		if (IsAnyOf(missile._mitype, MIS_MANASHIELD, MIS_REFLECT) && missile._misource == pnum) {
```
AJenbo, line 3222: Didn't add `MIS_ETHEREALIZE` as it's dead code

Source/missiles.h
```diff
@@ -88,7 +88,8 @@ enum Direction16 {
 };
 
 struct MissileStruct {
-	int _mitype; // Type of projectile (missile_id)
+	/** Type of projectile */
+	missile_id _mitype;
```
AJenbo, line 92: Can't believe how long this has gone unappealing.

Source/missiles.cpp
```diff
@@ -2933,40 +2933,35 @@ void AddDiabApoca(int mi, Point src, Point /*dst*/, int /*midir*/, int8_t mienem
...
-
-			for (int i = 0; i < ActiveMissileCount; i++) {
-				int mi = ActiveMissiles[i];
-				auto &missile = Missiles[mi];
-				if (missile._mitype == MIS_REFLECT && missile._misource == id)
-					return -1;
-			}
-		}
-	}
+	if (mitype == MIS_REFLECT && Players[id].wReflections != 0 && !CanAddAura(id, mitype))
```
qndel, line 2963:
```suggestion
	if (mitype == MIS_REFLECT && Players[id].wReflections > 0 && !CanAddAura(id, mitype))
```

qndel:
Looks good but I'm really not sold on the name "aura", it brings d2 auras to mind instantly and it just doesn't fit xD when I saw the PR name I was like "what the hell is it about?"

FitzRoyX:
Yeah an aura is generally an aoe buff for allies or an aoe debuff for enemies. These spells are just buffs.

obligaron:
in `scrollrt.cpp` we have a drawing function that renders mana shield and reflect. The function is called `DrawPlayerIcons`. Perhaps this helps for the naming. At least we should name both occurrences the same.

AJenbo:
We cant all be D2 aficionados

julealgon:
I suggest using the term "enchantment" for this kind of thing.

AJenbo:
> in `scrollrt.cpp` we have a drawing function that renders mana shield and reflect. The function is called `DrawPlayerIcons`. Perhaps this helps for the naming. At least we should name both occurrences the same.

I don't think this is nesseserily true, a player might have an icon drawn on them for other reasons.

AJenbo:
> I suggest using the term "enchantment" for this kind of thing.

Went with "effect" instead, maybe some one wants to add curses in the future :shrug:

qndel (approved):
👍

PR #2647, opened 2021-08-20
Clean up quest init

AJenbo:


Source/quests.cpp
```diff
@@ -505,7 +494,7 @@ void CheckQuestKill(const MonsterStruct &monster, bool sendmsg)
 void DRLG_CheckQuests(int x, int y)
 {
 	for (auto &quest : Quests) {
-		if (Quests[static_cast<quest_id>(quest._qidx)].IsAvailable()) {
+		if (Quests[quest._qidx].IsAvailable()) {
```
ephphatha, line 502:
```suggestion
		if (quest.IsAvailable()) 
```
No real reason to go back into the array here, _qidx is only equivalent to the index because of the way quest_ids are defined.

Source/quests.cpp
```diff
@@ -287,31 +273,34 @@ void InitQuests()
...
-					quest._qvar1 = 0;
-					quest._qlog = false;
-				}
-				initiatedQuests++;
-			} else {
+		quest._qidx = static_cast<quest_id>(q);
+		auto &questData = QuestData[q];
+		q++;
+
+		quest._qactive = QUEST_NOTAVAIL;
```
ephphatha, line 280: Probably worth bringing up the constant assignments for _qcar1, _qvar2, _qlog, position. They could potentially be made into a constructor later and it'll be easier to identify them if they're together.

Source/quests.h
```diff
@@ -51,7 +51,7 @@ struct QuestStruct {
 	dungeon_type _qlvltype;
 	Point position;
 	_setlevels _qslvl;
-	uint8_t _qidx;
+	quest_id _qidx;
```
ephphatha, line 54: Could even rename this _id or something similar, it was only ever used to index into the array once in DRLG_CheckQuests which is unnecessary now.

ephphatha (commented):
Comments from the first commit only, I'll have a look at the second commit later.

Source/quests.h
AJenbo: Yeah, most enums needs some heavy renaming so I'm half way pushing it in to next week, halfway hoping that others feel like they want to rename some of them.

Source/quests.cpp
```diff
@@ -264,54 +264,38 @@ void PrintQLString(const Surface &out, int x, int line, const char *str)
...
+		quest._qvar1 = 0;
+		quest._qvar2 = 0;
+		quest._qlog = false;
+		quest._qmsg = questData._qdmsg;
+
+		if (!gbIsMultiplayer) {
+			quest._qlevel = questData._qdlvl;
+			quest._qactive = QUEST_INIT;
+		} else if (!questData.isSinglePlayerOnly) {
+			quest._qlevel = questData._qdmultlvl;
```
ephphatha, line 293: Just noticed this in the original code, the only time this does anything useful is for Q_DIABLO where it sets qlevel to -1 instead of 15. Every other case is a no-op since `qdmultlvl == qlevel` for every other case when `isSinglePlayerOnly == false`. Is that a pre-existing bug?

AJenbo: I don't understand that to be a bug, just not very interesting given the quests in the original game

ephphatha: I see, the drlg_l4 checks involving Q_DIABLO never look at qlevel so this value isn't used for that quest.

Source/quests.cpp
```diff
@@ -123,7 +123,7 @@ void DrawButcher()
 	DRLG_RectTrans(x + 3, y + 3, x + 10, y + 10);
 }
 
-void DrawSkelKing(int q, int x, int y)
+void DrawSkelKing(quest_id q, int x, int y)
```
ephphatha, line 126: This could take a QuestStruct&, but it is also only ever called for Q_SKELKING. Could even make this constexpr given how short the function is.

Source/quests.cpp
```diff
@@ -150,7 +150,7 @@ void DrawWarLord(int x, int y)
 	}
 }
 
-void DrawSChamber(int q, int x, int y)
+void DrawSChamber(quest_id q, int x, int y)
```
ephphatha, line 153: Another opportunity for QuestStruct& (even with the known single use of Q_SCHAMB).

ephphatha (approved):
Some minor cleanups that could be included or left for a later pr, dealers choice 😄

Source/quests.cpp
AJenbo: Oh guess I missed this on during my second round.
Oh well still left some stuff so will probably still be handled in the next cleanup

AJenbo:
> Some minor cleanups that could be included or left for a later pr, dealers choice 😄

It was set to auto merge so that will have to be a follow up 😉

PR #2653, opened 2021-08-20
Rename SetDead

qndel:
Found this name in 1.08 address dump, makes much more sense :)

### DIFF
--- a/Source/dead.cpp
+++ b/Source/dead.cpp
@@ -76,7 +76,7 @@ void AddDead(Point tilePosition, int8_t dv, Direction ddir)
 	dDead[tilePosition.x][tilePosition.y] = (dv & 0x1F) + (ddir << 5);
 }
 
-void SetDead()
+void SyncUniqDead()
 {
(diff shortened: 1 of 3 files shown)


Source/dead.cpp
```diff
@@ -76,7 +76,7 @@ void AddDead(Point tilePosition, int8_t dv, Direction ddir)
 	dDead[tilePosition.x][tilePosition.y] = (dv & 0x1F) + (ddir << 5);
 }
 
-void SetDead()
+void SyncUniqDead()
```
julealgon, line 79: Can we avoid abbreviations? Is this for unique monsters? If so, can we use "Unique"?

julealgon, line 79: Also, what does "Sync" mean here? I think we should rename this to something that makes more sense/is more specific.

qndel: dunno, I literally copy/pasted the name

AJenbo: i think its related to loading them from saves

AJenbo (approved):

PR #2655, opened 2021-08-20
2435 refactor global variable Items part 1

bcko:
- related to #2435
- this pull request simply put `auto &item = Items[...omit]` at the top of functions so it is easier to refactor in subsequent pull requests.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -408,25 +408,26 @@ void AddInitItems()
 	int rnd = GenerateRnd(3) + 3;
 	for (int j = 0; j < rnd; j++) {
 		int ii = AllocateItem();
+		auto &item = Items[ii];
 
 		Point position = GetRandomAvailableItemPosition();
-		Items[ii].position = position;
+		item.position = position;
 
 		dItem[position.x][position.y] = ii + 1;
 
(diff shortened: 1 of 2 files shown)


Source/items.cpp
```diff
@@ -1492,6 +1500,7 @@ int RndTypeItems(int itype, int imid, int lvl)
 
 _unique_items CheckUnique(int i, int lvl, int uper, bool recreate)
 {
+	auto &item = Items[i];
```
AJenbo, line 1503: If there is just a single use in the function isn't not really beneficial do do this

AJenbo (approved):
A good first round for items. Maye a good thing to do next would be changing function that take the item by pointer to instead take it by reference. The benefit here is that references cannot be null and cannot be changed during the function call, but prevents some common crash issues, but also it makes the code more consistent.

Two functions i spotted where this applies are `SetPlrHandGoldCurs(ItemStruct *)` and
`SetPlrHandItem(ItemStruct *)`.
You could also start changing function to take `ItemStruct &` directly instead of the global index where it makes sens to do so (caller already has the item, function doesn't use the index for anything else)

PR #2659, opened 2021-08-21
reflect without missile!

qndel:
This gets rid of all the bugs and even allows stacking reflect charges. Inspired by berserk.
Tested in single and multi, but since it's a completely new approach, I'd appreciate some testing from you too : )
Look how much code got trashed 😀 
If this gets merged, we can think about rewriting manashield same way, as iirc it stored some variables in the missile but do we really need them after we've included manashield in the damage calculation itself?


Source/player.cpp
```diff
@@ -3221,7 +3225,7 @@ void RemovePlrMissiles(int pnum)
 		if (missile._mitype == MIS_STONE && missile._misource == pnum) {
 			Monsters[missile._miVar2]._mmode = (MON_MODE)missile._miVar1;
 		}
-		if (IsAnyOf(missile._mitype, MIS_MANASHIELD, MIS_REFLECT) && missile._misource == pnum) {
+		if (IsAnyOf(missile._mitype, MIS_MANASHIELD) && missile._misource == pnum) {
```
glebm, line 3228:
```suggestion
		if (missile._mitype == MIS_MANASHIELD && missile._misource == pnum) {
```

qndel: I knew about this but left it for future expansion ;p

Chance4us:
I've already tested a little and haven't had any problems so far.

qndel:
> I've already tested a little and haven't had any problems so far.

Can you join discord? :p It's much faster to chat than on github :D

AJenbo (approved):

Chance4us:
> > I've already tested a little and haven't had any problems so far.
> 
> Can you join discord? :p It's much faster to chat than on github :D

Yes great, i will notify you on discord.

PR #2661, opened 2021-08-21
stop gold from displaying red

qndel:
image
Without this, gold appeared red


Chance4us:
Never seen anything like this before. In what situation can this happen?

qndel:
> Never seen anything like this before. In what situation can this happen?

After you use debug command that gives gold ;)

AJenbo (approved):

PR #2662, opened 2021-08-21
New debug commands

qndel:
Refill - refills hp/mana
dropitem - tries to find an item that matches the name
dropunique - tries to find a unique item that matches the name

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4886,6 +4886,63 @@ void PutItemRecord(int nSeed, uint16_t wCI, int nIndex)
 	}
 }
 
+#ifdef _DEBUG
+std::string DebugSpawnItem(std::string itemName, bool unique)
+{
+	if (ActiveItemCount >= MAXITEMS)
+		return "No space to generate the item!";
+
+	bool onlygood = true;
+
+	int ii = AllocateItem();
+	Point pos = Players[MyPlayerId].position.tile;
+	GetSuperItemSpace(pos, ii);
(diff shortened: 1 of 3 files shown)


Source/items.cpp
```diff
@@ -4886,6 +4886,65 @@ void PutItemRecord(int nSeed, uint16_t wCI, int nIndex)
 	}
 }
 
+void DebugSpawnItem(std::string itemName, bool unique)
```
AJenbo, line 4889: Hide behind ifdef DEBUG

qndel: done

Source/items.cpp
```diff
@@ -4886,6 +4886,67 @@ void PutItemRecord(int nSeed, uint16_t wCI, int nIndex)
...
+
+	std::chrono::steady_clock::time_point begin = std::chrono::steady_clock::now();
+	MonsterStruct fake_m;
+	fake_m.MData = &MonsterData[0];
+	fake_m._uniqtype = 0;
+	for (int i = 0;; i++) {
+		std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();
+		int diff = std::chrono::duration_cast<std::chrono::milliseconds>(end - begin).count();
+		if (diff > 3000) {
+			NetSendCmdString(1 << MyPlayerId, "ITEM NOT FOUND IN 3 SECONDS!");
```
obligaron, line 4910: We should make `DebugSpawnItem` return this as `std::string`.
With this we can don't need `NetSendCmdString` here and if we later get a real console don't need update this method. 😉

AJenbo (approved):

PR #2663, opened 2021-08-21
apply enums to mTreasure

qndel:


AJenbo (approved):
Nice work, yet another example of multiple things cramed in to one variable

PR #2664, opened 2021-08-21
More debug commands

obligaron:
visit - let you visit someone in town
exit - exits the game
arrow - change arrow effect

PR #2666, opened 2021-08-21
talkto debug command

qndel:
Inspired by the command from https://github.com/diasurgical/devilutionX/pull/2664 but imo more functional :D 
"talkto x" talks to npc x, supports partial name matching


Source/debug.cpp
```diff
@@ -298,6 +307,7 @@ std::vector<DebugCmdItem> DebugCmdList = {
 	{ "refill", "Refills health and mana.", "", &DebugCmdRefillHealthMana },
 	{ "dropunique", "Attempts to generate unique item {name}.", "{name}", &DebugCmdGenerateUniqueItem },
 	{ "dropitem", "Attempts to generate item {name}.", "{name}", &DebugCmdGenerateItem },
+	{ "visit", "Interacts with a NPC whose name contains {name}.", "{name}", &DebugCmdTalkToTowner},
```
AJenbo, line 310: woudn't talkto make more sens as a name for this command?

qndel: aye

PR #2668, opened 2021-08-22
Make spelldata const

AJenbo:


Source/spelldat.h
```diff
@@ -217,6 +217,6 @@ struct SpellData {
 	int sStaffCost;
 };
 
-extern SpellData spelldata[];
+const extern SpellData spelldata[];
```
obligaron, line 220: Only a nit: in the codebase only `extern const` is used and never `const extern`

obligaron (changes requested):

obligaron:
Btw are the other debug keys used by anyone or can we remove them?

- 'd'/'D' `PrintDebugPlayer`
- 'M'/'m' `DebugMonster`
- 'R'/'r' Prints level seed informations.
- 't'/'T' Prints tile Informations.

'd' and 'r' could be easily converted to debug commands, cause they don't need cursor position.

AJenbo:
R is quite useful.

maybe T and D could be made to be continious (and not use the chat, but maybe automap for text placment)

PR #2669, opened 2021-08-22
fix hellfire/cows crash

qndel:
In debug "talkto" command

### DIFF
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -804,6 +804,29 @@ _speech_id QuestDialogTable[NUM_TOWNER_TYPES][MAXQUESTS] = {
 	// clang-format on
 };
 
+bool IsTownerPresent(_talker_id npc)
+{
+	switch (npc) {
+	case TOWN_DEADGUY:
+		if (Quests[Q_BUTCHER]._qactive == QUEST_NOTAVAIL || Quests[Q_BUTCHER]._qactive == QUEST_DONE)
+			return false;
+		break;
+	case TOWN_FARMER:
+		if (!gbIsHellfire || sgGameInitInfo.bCowQuest != 0 || Quests[Q_FARMER]._qactive == QUEST_HIVE_DONE)
+			return false;
+		break;
+	case TOWN_COWFARM:
+		if (!gbIsHellfire || sgGameInitInfo.bCowQuest == 0)
+			return false;
+		break;
+	case TOWN_GIRL:
+		if (!gbIsHellfire || sgGameInitInfo.bTheoQuest == 0 || !Players->_pLvlVisited[17] || Quests[Q_GIRL]._qactive == QUEST_DONE)
+			return false;
+		break;
+	}
+	return true;
+}
+
 void InitTowners()
 {
 	assert(CowCels == nullptr);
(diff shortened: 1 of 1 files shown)


Source/towners.cpp
```diff
@@ -898,6 +898,12 @@ bool DebugTalkToTowner(std::string targetName)
 	std::transform(targetName.begin(), targetName.end(), targetName.begin(), [](unsigned char c) { return std::tolower(c); });
 	auto &myPlayer = Players[MyPlayerId];
 	for (auto &towner : TownerInitList) {
+		// prevent going into hellfire territory without hellfire
+		if (towner.type == TOWN_COWFARM && !gbIsHellfire)
```
AJenbo, line 902: does this also affect the nut and the girl?

qndel: yes because they are behind in the table

Source/towners.cpp
```diff
@@ -898,6 +898,12 @@ bool DebugTalkToTowner(std::string targetName)
 	std::transform(targetName.begin(), targetName.end(), targetName.begin(), [](unsigned char c) { return std::tolower(c); });
 	auto &myPlayer = Players[MyPlayerId];
 	for (auto &towner : TownerInitList) {
+		// prevent going into hellfire territory without hellfire
+		if (towner.type == TOWN_COWFARM && !gbIsHellfire)
+			return false;
```
obligaron, line 903: `InitTowners` has a similar check that is even better (checks also quest state).
For example now you could talk to girl but theo quest is disabled.

Source/towners.cpp
```diff
@@ -898,6 +898,12 @@ bool DebugTalkToTowner(std::string targetName)
 	std::transform(targetName.begin(), targetName.end(), targetName.begin(), [](unsigned char c) { return std::tolower(c); });
 	auto &myPlayer = Players[MyPlayerId];
 	for (auto &towner : TownerInitList) {
+		// prevent going into hellfire territory without hellfire
+		if (towner.type == TOWN_COWFARM && !gbIsHellfire)
+			return false;
+		// cows have an init function that differs from the rest and isn't compatible with this code, skip them :(
+		if (towner.type == TOWN_COW)
+			continue;
```
obligaron, line 906: The only reason the init function is needed, is that it sets the name or?
Cause we don't need any animation (and could avoiding to load them).
We could add the name to `TownerInit` then we would avoid the init function entirely and it would also simplify existing code (cause name is not set in multiple places).

obligaron (changes requested):

Source/towners.cpp
qndel: yup

Source/towners.cpp
qndel: Meh, copying that check would duplicate a lot of code and I can't just call that function because it modifies dMonsters

obligaron: Can't [these lines](https://github.com/diasurgical/devilutionX/blob/master/Source/towners.cpp#L819-L838) be refactored in a method `IsTownerPresent` that returns a bool?

qndel: I guess :)

obligaron (approved):
I would address [this comment](https://github.com/diasurgical/devilutionX/pull/2669#discussion_r693481196). But I'm also okay with leaving it as it is.
So I'm fine with this change.
Thanks qndel 🙂

Issue #2656, opened 2021-08-21
monster.cpp:IsTileSafe breaks out of loops too early


ephphatha:
**Important information**
devliutionX master e355d03

**Describe**
The following code in IsTileSafe handles the case where multiple missiles occupy a single tile (dMissile[x][y] == -1):
https://github.com/diasurgical/devilutionX/blob/e355d0301903fac126b70ce1b6a3deda32ebd8e8/Source/monster.cpp#L2021-L2036

If a tile is occupied by both a Firewall and a Lightning Wall spell (may not be possible?) the break bails out of the loop as soon as it encounters one of those type of spell. This allows monsters which are immune to lightning but vulnerable to fire (or vv) to walk into a tile that damages them if the Lightning Wall spell is early in the active missiles list. (`lightning = true`, the loop is broken, `fire` remains set to false).

**To Reproduce**
Cast firewall and lightning wall so they overlap near a monster with only one relevant immunity.

**Expected behavior**
The tile should be considered unsafe if a monster is vulnerable to any of the spells/missiles occupying that tile.

**Additional context**
This can be fixed while keeping the intent of the original code by moving the monster attribute check before the loop and returning early if a tile contains a type of spell the monster should avoid.

For example:
```cpp
bool avoidFire = (monster.mMagicRes & IMMUNE_FIRE) == 0 || monster.MType->mtype == MT_DIABLO);
...
if (missile._miType == MT_FIREWALL && avoidFire)
    return false;
```

CodingBucko:
Alright, I've been wanting to dip my toe with open source projects and what better than contributing to Diablo, ill give it a shot!

qndel:
> Alright, I've been wanting to dip my toe with open source projects and what better than contributing to Diablo, ill give it a shot!
> 
> edit: I was looking at the wiki section and I am a bit confused about the binary exact stuff, does that pertain to any changes I make to this function?

Don't worry about bin exactness, this is a thing from the past when we were comparing stuff with original diablo, since devilutionx runs on linux and much more, these times are long gone :)

AJenbo:
> Alright, I've been wanting to dip my toe with open source projects and what better than contributing to Diablo, ill give it a shot!

Sorry I ended up fixing this since I had to rewrite the affected function for other reason. But feel free to pick another issue, there is pleanty to look at, most of it we haven't even created issues for so you can also ask for a type of issue you would like ot look in to :)

PR #2671, opened 2021-08-22
Replace dMissiles with BFLAG_MISSILE

AJenbo:
This is prep that lets us change most missile functions to use MissileStruct as the input rather then an index in a follow up PR.

@bcko once this is merged you can try and change all the missile add functions if you like :)


ephphatha (approved):
Fixes #2656

bcko (approved):
👏🎉

PR #2673, opened 2021-08-22
2435 refactor global variable Items part 2

bcko:
- related to #2435
- refactor global variable Items part 2

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -2101,12 +2101,12 @@ void PrintItemMisc(ItemStruct &item)
 	}
 }
 
-void PrintItemInfo(ItemStruct *x)
+void PrintItemInfo(ItemStruct &item)
 {
-	PrintItemMisc(*x);
-	uint8_t str = x->_iMinStr;
-	uint8_t dex = x->_iMinDex;
-	uint8_t mag = x->_iMinMag;
+	PrintItemMisc(item);
+	uint8_t str = item._iMinStr;
+	uint8_t dex = item._iMinDex;
+	uint8_t mag = item._iMinMag;
 	if (str != 0 || mag != 0 || dex != 0) {
 		strcpy(tempstr, _("Required:"));
 		if (str != 0)
@@ -4209,7 +4209,7 @@ void PrintItemDetails(ItemStruct *x)
 		ShowUniqueItemInfoBox = true;
 		curruitem = *x;
 	}
-	PrintItemInfo(x);
+	PrintItemInfo(*x);
 }
 
 void PrintItemDur(ItemStruct *x)
@@ -4249,7 +4249,7 @@ void PrintItemDur(ItemStruct *x)
 	}
 	if (x->_itype == ITYPE_RING || x->_itype == ITYPE_AMULET)
 		AddPanelString(_("Not Identified"));
-	PrintItemInfo(x);
+	PrintItemInfo(*x);
 }
 
 void UseItem(int p, item_misc_id mid, spell_id spl)
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -424,7 +424,7 @@ void AddInitItems()
 			GetItemAttrs(ii, IDI_MANA, curlv);
 
 		item._iCreateInfo = curlv | CF_PREGEN;
-		SetupItem(ii);
+		SetupItem(item);
 		item.AnimInfo.CurrentFrame = item.AnimInfo.NumberOfFrames;
 		item._iAnimFlag = false;
 		item._iSelFlag = 1;
@@ -1610,7 +1610,7 @@ void SetupAllItems(int ii, int idx, int iseed, int lvl, int uper, bool onlygood,
 			GetUniqueItem(item, (_unique_items)iseed); // uid is stored in iseed for uniques
 		}
 	}
-	SetupItem(ii);
+	SetupItem(item);
 }
 
 void SetupBaseItem(Point position, int idx, bool onlygood, bool sendmsg, bool delta)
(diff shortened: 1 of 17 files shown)


Source/scrollrt.cpp
```diff
@@ -679,30 +679,30 @@ void DrawItem(const Surface &out, int x, int y, int sx, int sy, bool pre)
 	if (bItem <= 0)
 		return;
 
-	ItemStruct *pItem = &Items[bItem - 1];
-	if (pItem->_iPostDraw == pre)
+	auto &pItem = Items[bItem - 1];
```
AJenbo, line 682: The `p` prefix is Hungarian notation (which we are moving away from) for pointer, since this is now a reference the name should instead be just `item`.

Source/items.cpp
```diff
@@ -3101,9 +3084,9 @@ int GetGoldCursor(int value)
  * @brief Update the gold cursor on the given gold item
  * @param h The item to update
  */
-void SetPlrHandGoldCurs(ItemStruct *h)
+void SetPlrHandGoldCurs(ItemStruct &h)
```
AJenbo, line 3087: Would be nice if you could rename this to `gold` since you are already here :)

Source/items.cpp
```diff
@@ -3010,45 +2995,43 @@ void CalcPlrInv(int playerId, bool loadgfx)
 	}
 }
 
-void SetPlrHandItem(ItemStruct *h, int idata)
+void SetPlrHandItem(ItemStruct &h, int idata)
 {
-	ItemDataStruct *pAllItem;
-
-	pAllItem = &AllItemsList[idata];
+	auto &pAllItem = AllItemsList[idata];
```
AJenbo, line 3000: `item` and `itemData` would be better names.

AJenbo (approved):
Nice general clean up, I made a few comments regarding some renames that would be nice to have done at the same time if you are willing :)

bcko:
> Nice general clean up, I made a few comments regarding some renames that would be nice to have done at the same time if you are willing :)

Thank you for your feedback @AJenbo ! I will update my pull request according to your comments

bcko:
Pull request updated! @AJenbo

Source/items.cpp
```diff
@@ -4477,13 +4456,13 @@ void SpawnSmith(int lvl)
 			memset(&Items[0], 0, sizeof(*Items));
 			Items[0]._iSeed = AdvanceRndSeed();
 			SetRndSeed(Items[0]._iSeed);
-			int idata = RndSmithItem(lvl) - 1;
-			GetItemAttrs(0, idata, lvl);
+			int itemData = RndSmithItem(lvl) - 1;
+			GetItemAttrs(Items[0], itemData, lvl);
 		} while (Items[0]._iIvalue > maxValue);
 		smithitem[i] = Items[0];
```
ephphatha, line 4462:
```suggestion
			GetItemAttrs(smithitem[i], itemData, lvl);
		} while (smithitem[i]._iIvalue > maxValue);
```

No need to use `Items[0]` as a temp buffer anymore :)

Source/items.cpp
```diff
@@ -4532,17 +4511,17 @@ void SpawnWitch(int lvl)
 	int j = PinnedItemCount;
 
 	memset(&Items[0], 0, sizeof(*Items));
-	GetItemAttrs(0, IDI_MANA, 1);
+	GetItemAttrs(Items[0], IDI_MANA, 1);
 	witchitem[0] = Items[0];
```
ephphatha, line 4515: Same again for these lines, witchitem[0/1/2/i/j] can all be used directly now.

Source/items.cpp
```diff
@@ -4629,8 +4608,8 @@ void SpawnBoy(int lvl)
 		Items[0]._iSeed = AdvanceRndSeed();
```
ephphatha, line 4608: There are no early returns from this point in the function, `boyitem` can be used instead of `Items[0]` without side effects.

Source/items.cpp
```diff
@@ -4728,20 +4707,20 @@ void SpawnHealer(int lvl)
 	int srnd;
 
 	memset(&Items[0], 0, sizeof(*Items));
-	GetItemAttrs(0, IDI_HEAL, 1);
+	GetItemAttrs(Items[0], IDI_HEAL, 1);
```
ephphatha, line 4710: Same thing here :)

Source/items.cpp
```diff
@@ -4771,7 +4750,7 @@ void SpawnHealer(int lvl)
 void SpawnStoreGold()
 {
 	memset(&Items[0], 0, sizeof(*Items));
-	GetItemAttrs(0, IDI_GOLD, 1);
+	GetItemAttrs(Items[0], IDI_GOLD, 1);
```
ephphatha, line 4753: And here :)

Source/items.cpp
```diff
@@ -4921,23 +4901,23 @@ std::string DebugSpawnItem(std::string itemName, bool unique)
 
 		int uper = (unique ? 15 : 1);
 
-		Point bkp = Items[ii].position;
-		memset(&Items[ii], 0, sizeof(ItemStruct));
-		Items[ii].position = bkp;
+		Point bkp = item.position;
+		memset(&item, 0, sizeof(ItemStruct));
```
ephphatha, line 4905:
```suggestion
		memset(&item, 0, sizeof(item));
```

ephphatha (approved):
There are a lot of cleanups made possible by this PR :D. I've made comments on many (but not all) spots I saw however I don't expect it to be implemented immediately, this PR is already large enough. Doing it as a part 3 will make it easier to review :)

PR #2674, opened 2021-08-22
2435 refactor missiles part 5

bcko:
- related to #2435

[Note]
- ~As @ephphatha pointed out in the Discord development channel, `Missiles[mx]._miVar3 == missile._miVar3` in `AddFireBolt()` is wrong.~ updated the pull request
- forward declaration of `struct MissileStruct;` due to dependency issue
- remove Missiles[] in PutMissile()
- remove Missiles[] in Add*()


Source/missiles.cpp
```diff
@@ -1959,7 +1933,7 @@ void AddFirebolt(int mi, Point src, Point dst, int midir, int8_t micaster, int i
 		int i;
 		for (i = 0; i < ActiveMissileCount; i++) {
 			int mx = ActiveMissiles[i];
-			if (Missiles[mx]._mitype == MIS_GUARDIAN && Missiles[mx]._misource == id && Missiles[mx]._miVar3 == mi)
+			if (Missiles[mx]._mitype == MIS_GUARDIAN && Missiles[mx]._misource == id && Missiles[mx]._miVar3 == missile._miVar3)
```
bcko, line 1936: This is wrong but I do not know 
- how best to add the internal identifier or 
- how to do it in a way without using the index/id

`Missiles[mx]._miVar3 == missile._miVar3`

AJenbo: _miVar3 is a temporary variable that is used for a variety of things depending on context. This happens in a few struct and is really bad code architecture as you can't easily tell the meaning of the value, and you can easily get in to a situation where two parts of the code both make use of the same variable for different reasons leading to unstable gameplay.

It looks to me like what is happening here is that the code is trying to identify if the firebolt was produced by a Guardian, and in that case not use player mana.


```suggestion
	auto &guardian = Missiles[mx];
	if (guardian._mitype == MIS_GUARDIAN && guardian._misource == id && guardian._miVar3 >= 0 && guardian._miVar3 < MAXMISSILES && &Missiles[guardian._miVar3] == &missile)
```

AJenbo: `guardian._miVar3` gets set to the id of the next missile just before spawning the firebolt.
In `AddFirebolt()` we need to figure out if the firebolt was spawned by a guardian. This is done by looping over all missiles looking for guardians and checking if it's _miVar3 points to our index in the global array.

Since memory lookups are a bit more sensitive then comparing indexes (integers) i also added a safeguard to check that it's with in the range of the `Missiles` array (because _miVar3 is also being used to hold other values) `guardian._miVar3 >= 0 && guardian._miVar3 < MAXMISSILES`.

`&Missiles[guardian._miVar3] == &missile` is the part that is equivalent to `guardian._miVar3 == mi`

bcko: Updated the pull request. Thank you so much for detailed explanation @AJenbo ! :+1:

AJenbo (approved):
Nice work!

Having all the add functions take MissileStruct as there input is a huge improvement to the code, I did not expect us to get here this soon!

AJenbo:
I guess next step is to see if we can do the same for all the missile logic functions `MI_*(int i)` -> `MI_*(MissileStruct &missile)`.

PR #2676, opened 2021-08-23
Clean up AddMissiles

AJenbo:


bcko (approved):
:clap: :tada:

Source/objects.cpp
```diff
@@ -2772,7 +2772,7 @@ bool OperateShrineMagical(int pnum)
 	    player.position.tile,
 	    player._pdir,
 	    MIS_MANASHIELD,
-	    -1,
+	    TARGET_PLAYERS,
```
AJenbo, line 2775: This value will appear in save games, but it only has an effect for spawning and simply needs to be `!= TARGET_MONSTER`, so won't have an effect on comparability.

Source/monster.cpp
```diff
@@ -2570,7 +2570,7 @@ void RhinoAi(int i)
 		if (dist >= 5
 		    && v < 2 * monster._mint + 43
 		    && LineClear([&monster](Point position) { return IsTileAvailable(monster, position); }, monster.position.tile, { fx, fy })) {
-			if (AddMissile(monster.position.tile, { fx, fy }, md, MIS_RHINO, monster._menemy, i, 0, 0) != -1) {
+			if (AddMissile(monster.position.tile, { fx, fy }, md, MIS_RHINO, TARGET_PLAYERS, i, 0, 0) != -1) {
```
AJenbo, line 2573: `MIS_RHINO` does not make use of the player id so no need to abuse `_micaster`

Source/missiles.cpp
```diff
@@ -3130,7 +3118,7 @@ void MI_Firebolt(int i)
 				AddMissile(missile.position.tile, { i, 0 }, missile._mimfnum, MIS_MISEXP3, missile._micaster, missile._misource, 0, 0);
 				break;
 			case MIS_BONESPIRIT:
-				SetMissDir(missile, DIR_OMNI);
+				SetMissDir(missile, 8);
```
AJenbo, line 3121: This is not a direction but the explode frame of the missile, only 6/59 of the missiles abusing the directions in this way :tada: the rest spawn an explosion missile on impact.

Source/missiles.cpp
```diff
@@ -3407,8 +3395,8 @@ void MI_FireRing(int i)
 {
 	auto &missile = Missiles[i];
 	missile._miDelFlag = true;
-	int8_t src = missile._micaster;
-	uint8_t lvl = src > 0 ? Players[src]._pLevel : currlevel;
+	int8_t src = missile._misource;
+	uint8_t lvl = missile._micaster == TARGET_MONSTERS ? Players[src]._pLevel : currlevel;
```
AJenbo, line 3399: Bad bug in hellfire, see commit message

bcko (dismissed):

Source/missiles.cpp
```diff
@@ -2686,202 +2698,167 @@ void AddDisarm(MissileStruct &missile, Point /*src*/, Point /*dst*/, int /*midir
...
 {
 	missile._miVar1 = 8;
-	missile._miVar2 = src.y - missile._miVar1;
-	missile._miVar3 = missile._miVar1 + src.y;
-	missile._miVar4 = src.x - missile._miVar1;
-	missile._miVar5 = missile._miVar1 + src.x;
+	missile._miVar2 = std::max(missile.position.start.y - 8, 1);
+	missile._miVar3 = std::min(missile.position.start.y + 8, MAXDUNY - 1);
+	missile._miVar4 = std::max(missile.position.start.x - 8, 1);
+	missile._miVar5 = std::min(missile.position.start.x + 8, MAXDUNX - 1);
```
ephphatha, line 2707: Pre-existing bug but 1 and maxdun*-1 seem mismatched, should this be 0 and maxdun*-1? 1 and maxdun*-2?

AJenbo: Yeah I noticed this as well, there is a similar issue elsewhere with the bound check in missiles, but it has no consequences for the game as the walls around the level are 16 tiles thick.

ephphatha (approved):
Nice, highlights a few opportunities to extract duplicate code and make checks more consistent but that can come later :D

AJenbo:
> Nice, highlights a few opportunities to extract duplicate code and make checks more consistent but that can come later :D

yeah the goal was just to get it to something that was possible to reason about :D

PR #2680, opened 2021-08-23
Some small missile cleanups

AJenbo:


bcko (approved):

Source/missiles.cpp
```diff
@@ -1415,7 +1412,7 @@ void AddBerserk(MissileStruct &missile, Point dst, int /*midir*/)
 
 void AddHorkSpawn(MissileStruct &missile, Point dst, int midir)
 {
-	UpdateMissileVelocity(missile, missile.position.start, dst, 8);
+	UpdateMissileVelocity(missile, dst, 8);
```
ephphatha, line 1415: This confused me for a sec but when mAddProc is called missile.position.start == missile.position.tile, so using tile instead of start (inside UpdateMissilVelocity) doesn't change behaviour :D

ephphatha (approved):

Issue #2683, opened 2021-08-23
Some spells don't deal damage anymore


Chance4us:
Torchant cannot be killed introduced with commit bea8cc9d87d04271ee2adc83f635e6c5b0aeade8.
Savegame:
[torchant.zip](https://github.com/diasurgical/devilutionX/files/7033904/torchant.zip)

ThomasChr:
I can't kill Bloodgutter.
Works on 1.2.1. Possibly the same problem.
[single_9.zip](https://github.com/diasurgical/devilutionX/files/7034151/single_9.zip)

ThomasChr:
![image](https://user-images.githubusercontent.com/13014163/130500514-a7bd88d6-0066-4fdf-b290-34e0706268a9.png)
mindam is 0, maxdam is 0, so dam will always be 0.

qndel:
The problem is not with monsters but with spells, a lot of them don't deal damage - lightning wall / lightning / nova / apoc etc.

PR #2685, opened 2021-08-23
fix missiles without dmg

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2683

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2841,6 +2841,7 @@ int AddMissile(Point src, Point dst, int midir, missile_id mitype, mienemy_type 
 	missile._mitype = mitype;
 	missile._micaster = micaster;
 	missile._misource = id;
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -2841,6 +2841,7 @@ int AddMissile(Point src, Point dst, int midir, missile_id mitype, mienemy_type
 	missile._mitype = mitype;
 	missile._micaster = micaster;
 	missile._misource = id;
+	missile._midam = midam;
```
julealgon, line 2844: Was this removed by mistake at some point?

AJenbo: Yeah https://github.comdiasurgical/devilutionX/pull/2676/files#diff-cf201edb86708433771f232de909a54b7712af8ed1e9feeadd25d149a8e9caf7L2916

PR #2686, opened 2021-08-23
rewrite manashield into not using a missile

qndel:


StephenCWills (approved):
FYI, if you load a save into vanilla, this has the interesting and somewhat comical effect of fooling the player into thinking they have Mana Shield when, in fact, they do not. That is to say, the game renders the mana shield (probably based on `pManaShield`), but it does not actually shield you from damage without the missile.

AJenbo:
> FYI, if you load a save into vanilla, this has the interesting and somewhat comical effect of fooling the player into thinking they have Mana Shield when, in fact, they do not. That is to say, the game renders the mana shield (probably based on `pManaShield`), but it does not actually shield you from damage without the missile.

That would be a bug that we should fix.

Source/missiles.cpp
```diff
@@ -2193,19 +2171,23 @@ void AddFlash2(MissileStruct &missile, Point /*dst*/, int /*midir*/)
...
-	}
 
-	missile._mirange = 48 * player._pLevel;
-	if (missile._micaster == TARGET_MONSTERS)
+	auto &player = Players[missile._misource];
+
+	if (missile._micaster == TARGET_MONSTERS) {
 		UseMana(missile._misource, SPL_MANASHIELD);
-	if (missile._misource == MyPlayerId)
+		player.pManaShield = !player.pManaShield;
```
AJenbo, line 2183: Not a fan of removing manshield, i don't see an upside and i could imagine mistake happening

StephenCWills:
> That would be a bug that we should fix.

In that case, I'm not sure how Reflect worked in vanilla, but I'd think the same issue probably applies to #2659.

AJenbo:
> In that case, I'm not sure how Reflect worked in vanilla, but I'd think the same issue probably applies to #2659.

Reflect and Manashield do not load in Hellfire, that bug was only fixed in Diablo in version 1.07

qndel:
> > In that case, I'm not sure how Reflect worked in vanilla, but I'd think the same issue probably applies to #2659.
> 
> Reflect and Manashield do not load in Hellfire, that bug was only fixed in Diablo in version 1.07

Actually manashield DOES load, only reflect doesn't

Source/items.cpp
```diff
@@ -3005,7 +3005,7 @@ void CalcPlrInv(int playerId, bool loadgfx)
 		CalcPlrBookVals(player);
 		player.CalcScrolls();
 		CalcPlrStaff(player);
-		if (playerId == MyPlayerId && currlevel == 0)
+		if (currlevel == 0)
```
AJenbo, line 3008: This seams a bit unrelated to manashield?

qndel: ye but I had some code there earlier and cleaned this useless if up

StephenCWills:
> I don't see a way to fix it and since new game gets rid of it, I wouldn't worry about it.
> Vanilla only unsets manashield flag when it finds a missile connected to it and since there's no missile, that's why it doesn't delete it and it doesn't work there. Trying to achieve any kind of compatibility would remove any purpose from this rewrite.

I'd think you could fabricate a missile and add it to the save file based on the existence of the `pManaShield` flag. However, I'm inclined to agree with the sentiment that compatibility here is not worth the effort.

Source/missiles.cpp
```diff
@@ -4465,6 +4427,15 @@ static void DeleteMissiles()
 	}
 }
 
+void ProcessManaShield()
+{
+	auto &myPlayer = Players[MyPlayerId];
+	if (myPlayer.pManaShield && myPlayer._pMana <= 0) {
+		myPlayer.pManaShield = false;
```
AJenbo, line 4434: You don't actually need this, it's done by the network command :)

PR #2691, opened 2021-08-24
items: add BUGFIX, reset item get records when resetting items

mewmew:
The item get record array tracks items being recently looted in an
effort to prevent the same item from being looted more than once.
Prior to this commit, the item get record array (and corresponding
item get record array length) variables were not cleared when
creating a new game. Therefore, the item get record array of a
previous game could remain in between games and prevent an item
from being looted (if it was looted in a previous), even if it was
never looted in the current game. In practice this almost never
shows up, since each item get record is valid for a total of 6
seconds before being cleared. So, you would either have to save
a game, quickly loot an item, when load the game and try to loot
the same item before 6 seconds pass. OR, you could use the demo
replay functionality to run test cases, and speed up execution to
run e.g. 10'000'000 logic ticks per second. Both would exhibit the
bug and prevent the item from being looted.

This commit fixes such issues by explicitly clearing the item get
record array and item get record array length variables whenever
items are initialized.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -127,8 +127,13 @@ enum class PlayerArmorGraphic : uint8_t {
 	// clang-format on
 };
 ItemStruct curruitem;
+
+/** Holds item get records, tracking items being recently looted. This is in an effort to prevent items being picked up more than once. */
 ItemGetRecordStruct itemrecord[MAXITEMS];
+
 bool itemhold[3][3];
+
+/** gnNumGetRecords specifies the number of active item get records. */
 int gnNumGetRecords;
 
 int OilLevels[] = { 1, 10, 1, 10, 4, 1, 5, 17, 1, 10 };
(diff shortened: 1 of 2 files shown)


Source/items.cpp
```diff
@@ -127,8 +127,13 @@ enum class PlayerArmorGraphic : uint8_t {
 	// clang-format on
 };
 ItemStruct curruitem;
+
+/** Holds item get records, tracking items being recently looted. This is in an effort to prevent items being picked up more than once. */
 ItemGetRecordStruct itemrecord[MAXITEMS];
+
 bool itemhold[3][3];
+
+/** gnNumGetRecords specifies the number of active item get records. */
```
mewmew, line 136:
```suggestion
/** Specifies the number of active item get records. */

```

AJenbo:
We are finding some interesting bugs now that we have demos :)

Nice to also have all the documentation with this change :+1:

AJenbo (approved):

julealgon:
Is this a vanilla issue?

mewmew:
> Is this a vanilla issue?

Yes. See https://github.com/diasurgical/devilution/pull/2251 for vanilla BUGFIX comment.

PR #2692, opened 2021-08-24
Fully type AddMissiles

AJenbo:
Use proper Direction type for AddMissile midir 

- Lightnings cast by familiars have a short range, this was communicated
by setting midir to -1. Solved by looking at the type of the owner

- Lightning and Inferno sync position with the parent missile. Solved by
copying the sync from the guardian spell


Source/monster.h
```diff
@@ -160,6 +160,8 @@ struct CMonster {
 	const MonsterDataStruct *MData;
 };
 
+struct MissileStruct;
```
ephphatha, line 163: Mixing in this forward declaration with actual definitions doesn't seem ideal, could this be moved just after the constants/defines at the start of the file? Or maybe just before the function declarations?

Source/missiles.cpp
```diff
@@ -2085,14 +2165,15 @@ void AddMisexp(MissileStruct &missile, Point dst, int /*midir*/)
 		}
 	}
 
-	missile.position.tile = Missiles[dst.x].position.tile;
-	missile.position.start = Missiles[dst.x].position.start;
-	missile.position.offset = Missiles[dst.x].position.offset;
-	missile.position.traveled = Missiles[dst.x].position.traveled;
+	int mx = FindParent(missile);
```
ephphatha, line 2168:
```suggestion
	int mx = FindParent(missile);
	assert(mx != -1); // AddMisexp will always be called with a parent associated to the missile.
```

Source/missiles.cpp
```diff
@@ -2410,34 +2491,56 @@ void AddStone(MissileStruct &missile, Point dst, int /*midir*/)
...
-	missile._miVar1 = missile.position.start.x;
-	missile._miVar2 = missile.position.start.y;
-	missile._miVar4 = dst.x;
-	missile._miVar5 = dst.y;
 	if (Monsters[missile._misource].position.tile != GolemHoldingCell && missile._misource == MyPlayerId)
 		M_StartKill(missile._misource, missile._misource);
+
 	UseMana(missile._misource, SPL_GOLEM);
+
+	int src = missile._misource;
```
ephphatha, line 2511: Might as well move this to the start of the function so it's used for the other `missile._misource` references. Could use a new name too (`playerId`?).

Source/missiles.cpp
```diff
@@ -762,6 +763,92 @@ bool CanAddEffect(const PlayerStruct &player, missile_id type)
...
+	if (mx == -1)
+		return;
+
+	auto &parent = Missiles[mx];
+
+	missile.position.offset = parent.position.offset;
+	missile.position.traveled = parent.position.traveled;
+}
+
+void SpawnLightning(MissileStruct &missile, int dam)
```
ephphatha, line 807: 🎉

ephphatha (approved):
Haven't tested it ingame so it'd be good to get another set of eyes on this, couldn't see any logic issues when I looked at it though.

Source/missiles.cpp
AJenbo: i dont get it :)

ephphatha: Glad to see a dedup of so many lines of code, that's all :D

AJenbo: lol thought it was a 🍕 emoji 😂

PR #2693, opened 2021-08-24
stop shops reroll on load

qndel:
no clue if that works, needs testing :)


Source/stores.cpp
```diff
@@ -2214,9 +2214,8 @@ void InitStores()
 
 void SetupTownStores()
 {
-	SetRndSeed(glSeedTbl[currlevel] * SDL_GetTicks());
-
 	auto &myPlayer = Players[MyPlayerId];
+	SetRndSeed(glSeedTbl[currlevel] * (myPlayer.townSeed + 1));
```
AJenbo, line 2218: why the +1 :)

julealgon:
Are you prepared for the rage this will cause? I noticed a lack of toggles and that's what immediately came to mind.

I'm all for it... but you guys know how this goes.

qndel:
> Are you prepared for the rage this will cause? I noticed a lack of toggles and that's what immediately came to mind.
> 
> I'm all for it... but you guys know how this goes.

It mostly affects speedrunners and nobody is going to speedrun devilutionx because we've disabled duping anyway

Source/stores.cpp
qndel: to still squeeze something useful out of old saves that will have this value at 0 :)

malvarenga123:
> Are you prepared for the rage this will cause? I noticed a lack of toggles and that's what immediately came to mind.
> 
> I'm all for it... but you guys know how this goes.

But is this actually a bug? If it isn't and gets changed, then it should be togglable indeed.

AJenbo:
image

julealgon:
> But is this actually a bug? If it isn't and gets changed, then it should be togglable indeed.

The criteria for adding a toggle is not "it is not a bugfix change".

PR #2701, opened 2021-08-26
Fix inaccurate gameplay with NOSOUND

AJenbo:
There where 3 calls to `GenerateRnd()` which would not get triggered if the game was compiled without sound support, which resulted in the game-play diverging from a full build.


glebm (dismissed):

AJenbo:
Reduced the impact of NOSOUND to about a dusin places.

obligaron (approved):
lgtm 🙂

AJenbo:
Thanks for reviewing :)

PR #2718, opened 2021-08-27
Update Bulgarian translation

NikoVP:
Remove fuzzy stuff + fixes. Used the latest available pot file to update the translation, but seems the last version seems to be based on a more recent version that isn't available.


AJenbo:
> Remove fuzzy stuff + fixes. Used the latest available pot file to update the translation, but seems the last version seems to be based on a more recent version that isn't available.

You can update directly from the source. The template is only updated when there are significant changes, and is only really intended for starting a new translation

PR #2726, opened 2021-08-27
Fix bugs in AddRndTeleport

AJenbo:
Phasing could end up attempting to teleport to hunreds of random
location and still fail to find a valid target and send the player in to
a wall or outside the map.

This is fixed by first finding all valid locations within range and then
picking one of them at random.

Valid targets is a 13x13 area around the player minus the inner 7x7

In total there can be no more then 120 valid tiles and on average there
should be 75 avalible.

----------

Previously an improper check was being done to see if the player was
located on the 3 teleporting pads in the level. Even if fixng this the
spell would not work correctly if the layer was standing on the pads
when casting the spell.

Simply use TARGET_BOTH to indicate that it should not be random and send
the player to the given destination

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1869,46 +1869,47 @@ void UpdateVileMissPos(MissileStruct &missile, Point dst)
 
 void AddRndTeleport(MissileStruct &missile, Point dst, Direction /*midir*/)
 {
-	int pn;
-	int r1;
-	int r2;
-
-	int nTries = 0;
-	do {
-		nTries++;
-		if (nTries > 500) {
-			r1 = 0;
-			r2 = 0;
-			break; //BUGFIX: warps player to 0/0 in hellfire, change to return or use 1.09's version of the code
-		}
-		r1 = GenerateRnd(3) + 4;
-		r2 = GenerateRnd(3) + 4;
-		if (GenerateRnd(2) == 1)
-			r1 = -r1;
-		if (GenerateRnd(2) == 1)
-			r2 = -r2;
-
-		r1 += missile.position.start.x;
-		r2 += missile.position.start.y;
-		if (r1 < MAXDUNX && r1 >= 0 && r2 < MAXDUNY && r2 >= 0) { ///BUGFIX: < MAXDUNX / < MAXDUNY (fixed)
-			pn = dPiece[r1][r2];
-		}
-	} while (nSolidTable[pn] || dObject[r1][r2] != 0 || dMonster[r1][r2] != 0);
-
 	missile._mirange = 2;
-	if (!setlevel || setlvlnum != SL_VILEBETRAYER) {
-		missile.position.tile = { r1, r2 };
-		if (missile._micaster == TARGET_MONSTERS)
-			UseMana(missile._misource, SPL_RNDTELEPORT);
-	} else {
+
+	auto &player = Players[missile._misource];
+
+	if (setlevel && setlvlnum == SL_VILEBETRAYER) {
 		int oi = dObject[dst.x][dst.y] - 1;
 		// BUGFIX: should only run magic circle check if dObject[dx][dy] is non-zero.
 		if (Objects[oi]._otype == OBJ_MCIRCLE1 || Objects[oi]._otype == OBJ_MCIRCLE2) {
 			missile.position.tile = dst;
-			if (!PosOkPlayer(Players[MyPlayerId], dst))
+			if (!PosOkPlayer(player, dst))
 				UpdateVileMissPos(missile, dst);
+			return;
 		}
 	}
+
+	std::array<Point, 13 * 13 - 7 * 7> targets;
+
+	int count = 0;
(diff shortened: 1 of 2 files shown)


Source/objects.cpp
```diff
@@ -1474,7 +1474,7 @@ void UpdateCircle(int i)
 		ObjChangeMapResync(Objects[i]._oVar1, Objects[i]._oVar2, Objects[i]._oVar3, Objects[i]._oVar4);
 		if (Quests[Q_BETRAYER]._qactive == QUEST_ACTIVE && Quests[Q_BETRAYER]._qvar1 <= 4) // BUGFIX stepping on the circle again will break the quest state (fixed)
 			Quests[Q_BETRAYER]._qvar1 = 4;
-		AddMissile(myPlayer.position.tile, { 35, 46 }, myPlayer._pdir, MIS_RNDTELEPORT, TARGET_MONSTERS, MyPlayerId, 0, 0);
+		AddMissile(myPlayer.position.tile, { 35, 46 }, myPlayer._pdir, MIS_RNDTELEPORT, TARGET_BOTH, MyPlayerId, 0, 0);
```
ephphatha, line 1477:
```suggestion
		AddMissile(myPlayer.position.tile, { 35, 46 }, DIR_S, MIS_RNDTELEPORT, TARGET_BOTH, MyPlayerId, 0, 0);
```
`MIS_RNDTELEPORT` has no animation (`_miAnimType == MFILE_NONE`) so the direction is unused. Might as well do the same as OperateShrineHoly and use a default value for this parameter.

Source/objects.cpp
```diff
@@ -2207,7 +2207,7 @@ void OperateBook(int pnum, int i)
 			}
 			if (doAddMissile) {
 				Objects[dObject[35][36] - 1]._oVar5++;
-				AddMissile(player.position.tile, { dx, dy }, player._pdir, MIS_RNDTELEPORT, TARGET_MONSTERS, pnum, 0, 0);
+				AddMissile(player.position.tile, { dx, dy }, player._pdir, MIS_RNDTELEPORT, TARGET_BOTH, pnum, 0, 0);
```
ephphatha, line 2210:
```suggestion
				AddMissile(player.position.tile, { dx, dy }, DIR_S, MIS_RNDTELEPORT, TARGET_BOTH, pnum, 0, 0);
```
Same here to match what you've done for OperateShrineHoly.

Source/missiles.cpp
```diff
@@ -1869,46 +1869,43 @@ void UpdateVileMissPos(MissileStruct &missile, Point dst)
...
+
+	if (count == 0) {
+		missile._miDelFlag = true;
+		return;
+	}
+
+	missile.position.tile = targets[std::max(GenerateRnd(count), 0)];
+
+	if (missile._micaster == TARGET_MONSTERS)
+		UseMana(missile._misource, SPL_RNDTELEPORT);
```
ephphatha, line 1908: The only time this will be true for this missile type is when AddMissile is called from CastSpell, if we can align the rest of the player cast spells we could move the UseMana call there eventually 🙂

Source/objects.cpp
```diff
@@ -3049,22 +3049,7 @@ bool OperateShrineHoly(int pnum)
...
-		xx = GenerateRnd(MAXDUNX);
-		yy = GenerateRnd(MAXDUNY);
-		lv = dPiece[xx][yy];
-		j++;
-		if (j > MAXDUNX * MAXDUNY)
-			break;
-	} while (nSolidTable[lv] || dObject[xx][yy] != 0 || dMonster[xx][yy] != 0);
-
-	AddMissile(player.position.tile, { xx, yy }, player._pdir, MIS_RNDTELEPORT, TARGET_PLAYERS, pnum, 0, 2 * leveltype);
+	AddMissile(Players[pnum].position.tile, { 0, 0 }, DIR_S, MIS_RNDTELEPORT, TARGET_PLAYERS, pnum, 0, 2 * leveltype);
```
ephphatha, line 3052: Looks like Holy Shrines were at one point intended to teleport you anywhere in the dungeon but that never worked 😂 . Given everyone expects the current behaviour of acting like phasing this is good.

Source/missiles.cpp
```diff
@@ -1869,46 +1869,43 @@ void UpdateVileMissPos(MissileStruct &missile, Point dst)
...
+			count++;
 		}
 	}
+
+	if (count == 0) {
+		missile._miDelFlag = true;
+		return;
+	}
+
+	missile.position.tile = targets[std::max<int32_t>(GenerateRnd(count), 0)];
```
ephphatha, line 1905: Surprised being explicit about the type for std::max is necessary, GenerateRnd returns int32_t, the numeric literal should get the same type ☹️

ephphatha (approved):
Looks good, a couple of minor changes possible to consistently mark unused params.

Source/missiles.cpp
AJenbo: 3ds is a bit anal about its types

PR #2727, opened 2021-08-27
Clean up points in missiles.cpp

AJenbo:


Source/missiles.cpp
```diff
@@ -1854,16 +1842,16 @@ void UpdateVileMissPos(MissileStruct &missile, Point dst)
...
-					missile.position.tile = { xx, yy };
+				missile.position.tile.x = i + dst.x;
+				if (PosOkPlayer(Players[MyPlayerId], missile.position.tile)) {
 					return;
 				}
 			}
 		}
 	}
+
 	missile.position.tile = dst;
```
ephphatha, line 1855: This assignment is unnecessary (as long as the loop doesn't modify `missile.position.tile` before the check succeeds anyway) since the same assignment is set before UpdateVileMissPos is called.

Source/missiles.cpp
```diff
@@ -1854,16 +1842,16 @@ void UpdateVileMissPos(MissileStruct &missile, Point dst)
...
 		for (int j = -k; j <= k; j++) {
-			int yy = j + dst.y;
+			missile.position.tile.y = j + dst.y;
 			for (int i = -k; i <= k; i++) {
-				int xx = i + dst.x;
-				if (PosOkPlayer(Players[MyPlayerId], { xx, yy })) {
-					missile.position.tile = { xx, yy };
+				missile.position.tile.x = i + dst.x;
+				if (PosOkPlayer(Players[MyPlayerId], missile.position.tile)) {
 					return;
```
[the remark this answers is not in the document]
ephphatha: Yeah, just using `{xx, yy}` or a temporary `Point` would be better imo.

Source/missiles.cpp
```diff
@@ -1953,20 +1941,20 @@ void AddMagmaball(MissileStruct &missile, Point dst, Direction /*midir*/)
...
-			if (0 < tx && tx < MAXDUNX && 0 < ty && ty < MAXDUNY) {
-				if (IsTileNotSolid({ tx, ty }) && dMonster[tx][ty] == 0 && dObject[tx][ty] == 0 && dPlayer[tx][ty] == 0) {
-					missile.position.tile = { tx, ty };
-					missile.position.start = { tx, ty };
-					missile._miDelFlag = false;
-					i = 6;
-					break;
-				}
+			Point target = dst + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
+			if (!InDungeonBounds(target) || !IsTileNotSolid(target))
```
ephphatha, line 1950:
```suggestion
			if (!InDungeonBounds(target) || IsTileSolid(target))
```
The double negative irks me a bit 😂. While `IsTileSolid != !IsTileNotSolid` the only exceptional case is when the point is out of bounds which gets checked first, so this should be fine.

Source/missiles.cpp
```diff
@@ -2242,19 +2231,19 @@ void AddGuardian(MissileStruct &missile, Point dst, Direction /*midir*/)
...
-					if (dMonster[tx][ty] == 0 && !nSolidTable[k] && !nMissileTable[k] && dObject[tx][ty] == 0 && (dFlags[i][j] & BFLAG_MISSILE) == 0) {
-						missile.position.tile = { tx, ty };
-						missile.position.start = { tx, ty };
-						missile._miDelFlag = false;
-						UseMana(missile._misource, SPL_GUARDIAN);
-						i = 6;
-						break;
-					}
+			Point target = dst + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
+			k = dPiece[target.x][target.y];
```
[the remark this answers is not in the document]
ephphatha: Yeah, this is reusing k for no real benefit in a way that's confusing. Might as well move the assignment inside `if(LineClearMissile()){ ... }`.

Source/missiles.cpp
```diff
@@ -2593,20 +2580,20 @@ void AddFirewallC(MissileStruct &missile, Point dst, Direction midir)
...
-						i = 6;
-						break;
-					}
+			Point target = dst + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
+			if (!InDungeonBounds(target))
+				continue;
+
+			k = dPiece[target.x][target.y];
+			if (LineClearMissile(missile.position.start, target)) {
+				if (missile.position.start != target && !nSolidTable[k] && dObject[target.x][target.y] == 0) {
```
ephphatha, line 2589:
```suggestion
			if (LineClearMissile(missile.position.start, target)) {
				if (missile.position.start != target && IsTileNotSolid(target) && dObject[target.x][target.y] == 0) {
```
Using k as an alias doesn't do anything useful here.

Source/missiles.cpp
```diff
@@ -3609,8 +3598,7 @@ void MI_Firemove(MissileStruct &missile)
 {
 	constexpr int ExpLight[14] = { 2, 3, 4, 5, 5, 6, 7, 8, 9, 10, 11, 12, 12 };
 
-	missile.position.tile.x--;
-	missile.position.tile.y--;
+	missile.position.tile -= Displacement { 1, 1 };
```
ephphatha, line 3601:
```suggestion
```
`missile.position.tile` gets set in UpdateMissilePos, so this modification does nothing before the next use. Same with the modification to `missile.position.offset.deltaY`.

Source/missiles.cpp
```diff
@@ -3628,19 +3616,18 @@ void MI_Firemove(MissileStruct &missile)
...
 		}
 	} else {
 		if (missile.var2 == 0)
 			missile._mlid = AddLight(missile.position.tile, ExpLight[0]);
 		ChangeLight(missile._mlid, missile.position.tile, ExpLight[missile.var2]);
 		missile.var2++;
 	}
-	missile.position.tile.x++;
-	missile.position.tile.y++;
+	missile.position.tile += Displacement { 1, 1 };
```
ephphatha, line 3630:
```suggestion
	missile.position.tile += DIR_S;
```
No idea why this missile is being shifted south one tile... This is being done for other missiles too, seems to always be a shift north or south.

obligaron (approved):
lgtm 🙂

PR #2728, opened 2021-08-27
Proper data structure for Minisets (Catacombs)

vlolteanu:


Source/drlg_l2.cpp
```diff
@@ -1723,50 +1727,40 @@ bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool
...
-					}
-					ii++;
+			if (sx >= nSx1 && sx <= nSx2 && sy >= nSy1 && sy <= nSy2)
+				goto next;
+			for (int yy = 0; yy < sh; yy++) {
+				for (int xx = 0; xx < sw; xx++) {
+					if (miniset.search[yy][xx] != 0 && dungeon[xx + sx][yy + sy] != miniset.search[yy][xx])
+						goto next;
+					if (dflags[xx + sx][yy + sy] != 0)
+						goto next;
```
AJenbo, line 1744: can we do this wwithout introducing goto, maybe some sub functions?

vlolteanu: I'm very partial towards using goto, especially for the purpose of simplifying the control flow of a function. However, I made sure to place it in a separate commit. If you're dead set against goto, just merge everything aside from https://github.com/diasurgical/devilutionX/pull/2728/commits/6f3586fb58c7bbe1a8b99991e65e19a1a64feb1f.

Source/drlg_l2.cpp
```diff
@@ -41,1497 +42,1504 @@ const ShadowStruct SPATSL2[2] = { { 6, 3, 0, 3, 48, 0, 50 }, { 9, 3, 0, 3, 48, 0
...
-
-	3, 0, // search
-	3, 1,
-	3, 4,
-	0, 7,
+struct Miniset {
+	Size size;
+	/* these are indexed as [y][x] */
+	unsigned char search[5][5];
+	unsigned char replace[5][5];
```
ephphatha, line 49:
```suggestion
	uint8_t search[5][5];
	uint8_t replace[5][5];
```
`search`/`replace` are only used in comparisons and assignments with `dungeon`, which is an array of uint8_t

Source/drlg_l2.cpp
```diff
@@ -1723,50 +1727,40 @@ bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool
...
-				for (int xx = 0; xx < sw && found; xx++) {
-					if (miniset[ii] != 0 && dungeon[xx + sx][yy + sy] != miniset[ii]) {
-						found = false;
-					}
-					if (dflags[xx + sx][yy + sy] != 0) {
-						found = false;
-					}
-					ii++;
+			if (sx >= nSx1 && sx <= nSx2 && sy >= nSy1 && sy <= nSy2)
+				goto next;
```
ephphatha, line 1738:
```suggestion
				continue;
```
No need to jump to the end of the current loop only to start the next iteration.

Source/drlg_l2.cpp
```diff
@@ -1723,50 +1727,40 @@ bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool
...
-			if (found && GenerateRnd(100) < rndper) {
-				for (int yy = 0; yy < sh; yy++) {
-					for (int xx = 0; xx < sw; xx++) {
-						if (miniset[kk] != 0) {
-							dungeon[xx + sx][yy + sy] = miniset[kk];
-						}
-						kk++;
-					}
+			if (GenerateRnd(100) >= rndper)
+				goto next;
```
ephphatha, line 1755:
```suggestion
				continue;
```

Source/drlg_l2.cpp
vlolteanu: I guess brevity doesn't hurt.

Source/drlg_l2.cpp
ephphatha: It's nice for breaking out of nested loops, but that only really applies to the inner checks (miniset overlaps a pre-existing set, miniset search doesn't match the full pattern). The search code in particular could be reused in PlaceMiniSet if it were a function.

Source/drlg_l2.cpp
vlolteanu: I simply did it because there were other instances of `goto next` in the function. Let's see if @AJenbo likes gotos at all.

Source/drlg_l2.cpp
```diff
@@ -1677,15 +1685,13 @@ bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool
...
 			for (int yy = 0; yy < sh && abort; yy++) {
 				for (int xx = 0; xx < sw && abort; xx++) {
-					if (miniset[ii] != 0 && dungeon[xx + sx][yy + sy] != miniset[ii])
+					if (miniset.search[yy][xx] != 0 && dungeon[xx + sx][yy + sy] != miniset.search[yy][xx])
 						abort = false;
 					if (dflags[xx + sx][yy + sy] != 0)
 						abort = false;
-					ii++;
 				}
 			}
```
ephphatha, line 1696: This is the same logic as lines 1739-1746. Would be nice extracted as a function.

Source/drlg_l2.cpp
```diff
@@ -1703,14 +1709,12 @@ bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool
...
 		for (int yy = 0; yy < sh; yy++) {
 			for (int xx = 0; xx < sw; xx++) {
-				if (miniset[ii] != 0) {
-					dungeon[xx + sx][yy + sy] = miniset[ii];
+				if (miniset.replace[yy][xx] != 0) {
+					dungeon[xx + sx][yy + sy] = miniset.replace[yy][xx];
 				}
-				ii++;
 			}
 		}
```
ephphatha, line 1719: This is the same as (the original logic for) lines 1756-1761.

Source/drlg_l2.cpp
```diff
@@ -1723,50 +1727,40 @@ bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool
...
-						if (miniset[kk] != 0) {
-							dungeon[xx + sx][yy + sy] = miniset[kk];
-						}
-						kk++;
-					}
+			if (GenerateRnd(100) >= rndper)
+				goto next;
+			for (int yy = 0; yy < sh; yy++) {
+				for (int xx = 0; xx < sw; xx++) {
+					if (miniset.search[yy][xx] != 0)
```
ephphatha, line 1758:
```suggestion
					if (miniset.replace[yy][xx] != 0)
```
The old logic was using kk (the replace index) for the check and assignment

Source/drlg_l2.cpp
AJenbo: I do not, we do not have any `goto` elsewhere in the code except for what we inherited 8 we inherited, 7 of them should be replaced with RAII and the last one is a copy of an SDL function that we keep for comparability with older versions of SDL and those do not want to modify (so that we can easily update it from upstream).

> I simply did it because there were other instances of `goto`

This is basically the problem with introducing them in the first place. They tend to breed and makes the code harder to read very quickly. Each instance is custom logic where `continue` is well defined.

julealgon:
Man.... this is a lot of changes for a single PR 😢

Source/drlg_l2.cpp
vlolteanu: > They tend to breed

I guess `goto`s have a pretty good sex life. Let's not be envious of them.

vlolteanu:
I can't think of a proper name for what this thing does:

```
for (int yy = std::max(sy - sh, 0); yy < std::min(sy + 2 * sh, DMAXY) && found; yy++) {
	for (int xx = std::max(sx - sw, 0); xx < std::min(sx + 2 * sw, DMAXX); xx++) {
		// BUGFIX: yy and xx can go out of bounds (fixed)
		if (dungeon[xx][yy] == miniset.replace[0][0]) {
			found = false;
			break;
		}
	}
}
```

AJenbo:
It's searching for the start of the mini set

vlolteanu:
I know, but the area searched seems arbitrary and asymmetrical. It bans a duplicate miniset from existing in an area the size of 4x4 such minisets. The candidate position is not centered (as in at 1.5, 1.5); it is at 1, 1.

ephphatha (approved):
Good cleanup compared to the old code, annoyingly PlaceMiniSet checking cx and cy separately means it's non-trivial to refactor `abort` the same way you've done for `found` in PlaceMiniSetRandom.

PR #2729, opened 2021-08-27
Demo: Decouple recording and playback resolution

obligaron:
With this change it is possible to play demos in a different resolution then recorded. 📹 

### Original resolution (also used for recording)

https://user-images.githubusercontent.com/25415264/131180804-681dd444-e3ea-4d2a-a628-e418b1f9402d.mp4

### Larger resolution

https://user-images.githubusercontent.com/25415264/131180840-f0fd01f6-5f59-4c99-8877-f6fc116f8e62.mp4

### Smaller resolution

https://user-images.githubusercontent.com/25415264/131180885-3e85644b-f343-473f-a1c5-0451559c0c4d.mp4

</details>

Notes:

- The difference speed of the example videos is cause I used timedemo mode. And timedemo mode tries to progress the playback as fast as possible. And cause larger resolution needs longer to render, the video is longer (see fps counter).
- This change introduces two different views. One for gameplay (used for click position calculation; uses recording solution) and one for visual/rendering (uses replay resolution). In normal gameplay both are identically. This is only needed for demo mode.
- In a follow-up pr, we could remove `PANEL_TOP`, `PANEL_LEFT`, `PANEL_X` and `PANEL_Y` and replace them with `GetMainPanel`.
- This change drives demo mode in a specific direction (demo recording contains user/click interactions). If we don't want to follow this route, this pr adds complexity that is not needed. But even if we discard this later, it can be easily removed then.

### DIFF
--- a/Source/utils/display.cpp
+++ b/Source/utils/display.cpp
@@ -36,6 +36,21 @@ Uint16 gnScreenWidth;
 Uint16 gnScreenHeight;
 Uint16 gnViewportHeight;
 
+Uint16 GetScreenWidth()
+{
+	return gnScreenWidth;
+}
+
+Uint16 GetScreenHeight()
+{
+	return gnScreenHeight;
+}
+
+Uint16 GetViewportHeight()
+{
+	return gnViewportHeight;
+}
+
 #ifdef USE_SDL1
 void SetVideoMode(int width, int height, int bpp, uint32_t flags)
 {
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -78,6 +78,19 @@ Rectangle LeftPanel;
 Rectangle RightPanel;
 std::optional<OwnedSurface> pBtmBuff;
 
+const Rectangle &GetMainPanel()
+{
+	return MainPanel;
+}
+const Rectangle &GetLeftPanel()
+{
+	return LeftPanel;
+}
+const Rectangle &GetRightPanel()
+{
+	return RightPanel;
+}
+
 extern std::array<Keymapper::ActionIndex, 4> quickSpellActionIndexes;
 
 /** Maps from attribute_id to the rectangle on screen used for attribute increment buttons. */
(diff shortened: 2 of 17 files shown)


AJenbo:
Definitionally looks nice :)

I still mostly believe in changing the recording to record commands rather then mouse/keyboard, both because it will allow recordings from game-controllers, but also because it forces us to clean up these parts of the code.

I'll have a look at the code in this PR in the next few days to see how I feel about merging it despite still feeling like the demo format should eventually change.

obligaron:
> I still mostly believe in changing the recording to record commands rather then mouse/keyboard, both because it will allow recordings from game-controllers, but also because it forces us to clean up these parts of the code.

The refactoring is a little bit weak argument, cause we can do the refactoring anyway. But I think that's a good discussion point for #2733 😉 

> I'll have a look at the code in this PR in the next few days to see how I feel about merging it despite still feeling like the demo format should eventually change.

This PR has five major changes and I think two of them are good even if we change the demo message format

- Replace global `gnScreenWidth`, `gnScreenHeight` and `gnViewportHeight` with functions (good anyway, cause less globals and a function ensures that it doesn't change anything)
- Replace define `UI_OFFSET_Y` with a function (good anyway, cause defines are ugly).
- Replace global `MainPanel`, `LeftPanel` and `RightPanel` with functions (good anyway, cause less globals and a function ensures that it doesn't change anything)
- Replace define `PANEL_TOP`, `PANEL_LEFT`, `PANEL_X`, `PANEL_Y` with functions (good anyway, cause defines are ugly and these coordinates are also redundant cause `MainPanel` has the exact same information).
- Introduce `forGamePlay` flag for these new functions to differ between gameplay and rendering.

Only the last is only specific for the current demo message format and can be easily removed if not needed (remove the parameter for the function and adjust all call-sites; doesn't take long 😉).

obligaron:
@AJenbo after #2733 was closed/decided I adjusted this pr as mentioned in my [previous comment](https://github.com/diasurgical/devilutionX/pull/2729#issuecomment-907736078).
This pr doesn't replace all globals/defines but is a start. 😉

Source/utils/display.cpp
```diff
@@ -15,6 +15,7 @@
 #include "controls/devices/joystick.h"
 #include "controls/devices/kbcontroller.h"
 #include "controls/game_controls.h"
+#include "engine/demomode.h"
```
AJenbo, line 18: Probably not needed any more?

obligaron: Yes that's true. Thanks for finding it. 🙂 Fixed

AJenbo (approved):

PR #2731, opened 2021-08-28
Use scoped enum for Direction

ephphatha:
Some very ugly uses of Direction values mean this enum has a lot of casts. I've introduced a couple of helpers to start addressing the ugliness but there's plenty of room for improvement. Mainly tried to keep this minimal since just getting it to compile required changing a huge number of files...


Source/scrollrt.cpp
```diff
@@ -836,7 +836,7 @@ void DrawDungeon(const Surface &out, int sx, int sy, int dx, int dy)
 	if (LightTableIndex < LightsMax && bDead != 0) {
 		do {
 			DeadStruct *pDeadGuy = &Dead[(bDead & 0x1F) - 1];
-			auto dd = static_cast<Direction>((bDead >> 5) & 7);
+			auto dd = (bDead >> 5) & 7; // Direction, but used as an array index
 			int px = dx - CalculateWidth2(pDeadGuy->width);
 			const byte *pCelBuff = pDeadGuy->data[dd];
```
AJenbo, line 841:
```suggestion
			int px = dx - CalculateWidth2(pDeadGuy->width);
			const byte *pCelBuff = pDeadGuy->data[(bDead >> 5) & 7];
```

ephphatha:
The linuz_x86_64 and amigaos-m68k builds use gcc 6.3 which has a bug stemming from an early C++14 revision being too strict with the definition of constexpr (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66297). If those builds can use gcc 7.2 or later instead then the GetCelSpritesForDirection helpers can be made constexpr instead of inline.

AJenbo:
> If those builds can use gcc 7.2 or later instead then the GetCelSpritesForDirection helpers can be made constexpr instead of inline.

If GCC 7.2 is avalible form Debian backports we could probably switch to that, but I haven't been able to update my PowerPC past GCC 6.2 :(

Amiga ports where working on a GCC 9 port afaik, don't know what the status is on that.

Source/player.cpp
```diff
@@ -293,14 +293,14 @@ _sfx_id herosounds[enum_size<HeroClass>::value][enum_size<HeroSpeech>::value] =
...
-	{ DIR_E,  {  1, -1 }, { -32, -16 }, { 1, 0 }, SDIR_E,  PM_WALK3, WalkSides     },
-	{ DIR_SE, {  1,  0 }, { -32, -16 }, { 0, 0 }, SDIR_SE, PM_WALK2, WalkDownwards }
+	{ Direction::South,  {  1,  1 }, {   0, -32 }, { 0, 0 }, SDIR_S,  PM_WALK2, WalkDownwards },
+	{ Direction::SouthWest, {  0,  1 }, {  32, -16 }, { 0, 0 }, SDIR_SW, PM_WALK2, WalkDownwards },
+	{ Direction::West,  { -1,  1 }, {  32, -16 }, { 0, 1 }, SDIR_W,  PM_WALK3, WalkSides     },
+	{ Direction::NorthWest, { -1,  0 }, {   0,   0 }, { 0, 0 }, SDIR_NW, PM_WALK,  WalkUpwards   },
+	{ Direction::North,  { -1, -1 }, {   0,   0 }, { 0, 0 }, SDIR_N,  PM_WALK,  WalkUpwards   },
+	{ Direction::NorthEast, {  0, -1 }, {   0,   0 }, { 0, 0 }, SDIR_NE, PM_WALK,  WalkUpwards   },
+	{ Direction::East,  {  1, -1 }, { -32, -16 }, { 1, 0 }, SDIR_E,  PM_WALK3, WalkSides     },
+	{ Direction::SouthEast, {  1,  0 }, { -32, -16 }, { 0, 0 }, SDIR_SE, PM_WALK2, WalkDownwards }
```
Iniquitatis, line 303: Broken alignment.

AJenbo:
@ephphatha sans the conflicts is this ready for review/merge?

ephphatha:
Yeah, I was hoping to review how directions were used for sprites and see what further cleanup was available there but I think that'll be best for a later PR. I'll rebase this evening and it'll be ready for review.

Source/monster.h
```diff
@@ -213,6 +219,26 @@ struct Monster { // note: missing field _mAFNum
 	CMonster *MType;
 	const MonsterData *MData;
 
+	/**
+	 * @brief Sets the current cell sprite to match the desired direction and animation sequence
+	 * @param graphic Animation sequence of interest
+	 * @param direction Desired direction the monster should be visually facing
+	*/
+	void ActivateAnimation(MonsterGraphic graphic, Direction direction)
```
obligaron, line 227: I would rename it to `ChangeAnimationData` to match the method in `AnimationInfo` (or if we dislike it, rename the version in `AnimationInfo`), cause both do more or less the same. Only `ChangeAnimationData` is even more safe (cause it regards if the number of frames differ from one animation to another).

(Same for `ActivateAnimationForCurrentDirection`)

obligaron (commented):
lgtm, only some small comments.
Thanks for working on this. 🙂

Source/missiles.cpp
```diff
@@ -167,33 +167,40 @@ void UpdateMissilePos(Missile &missile)
...
-		offset = { 1, 1 };
+	case Direction::South:
+	case Direction::SouthWest:
+	case Direction::SouthEast:
+		moveDirection = Direction::South;
 		break;
+	case Direction::NorthWest:
+	case Direction::North:
+	case Direction::NorthEast:
+	default:
```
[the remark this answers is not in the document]
ephphatha: It's not strictly necessary, I left in the explicit cases for North* directions more as documentation of the behaviour for those values. We could just use default and have the method comment document the behaviour for these cases.

Source/missiles.cpp
```diff
@@ -111,7 +111,7 @@ Monster *FindClosest(Point source, int rad)
 
 constexpr Direction16 Direction16Flip(Direction16 x, Direction16 pivot)
 {
-	unsigned ret = (2 * pivot + 16 - x) % 16;
```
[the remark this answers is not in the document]
ephphatha: The default underlying type will be (signed) `int`. Previously it was implementation defined and may or may not have been unsigned. The values assigned to the constants will be >= 0 so it ended up not making a difference to this logic.

Source/missiles.cpp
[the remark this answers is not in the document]
AJenbo: I would say it should either have all explicit cases or use default, not both :)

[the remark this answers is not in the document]
ephphatha: `default` it is then :D. I'll make that change this evening along with the other suggestions.

Source/engine/animationinfo.cpp
```diff
@@ -174,7 +174,6 @@ void AnimationInfo::ChangeAnimationData(const CelSprite *celSprite, int numberOf
 		TickModifier = 0.0F;
 	}
 	this->pCelSprite = celSprite;
-	TicksPerFrame = ticksPerFrame;
```
ephphatha, line 177: This was always a no-op, either the two values are equal or the same assignment was performed on line 171

Source/monster.cpp
```diff
@@ -4574,18 +4561,13 @@ void SyncMonsterAnim(Monster &monster)
 	case MonsterMode::Charge:
 		graphic = MonsterGraphic::Attack;
 		monster.AnimInfo.CurrentFrame = 1;
-		monster.AnimInfo.NumberOfFrames = monster.MType->GetAnimData(MonsterGraphic::Attack).Frames;
 		break;
 	default:
 		monster.AnimInfo.CurrentFrame = 1;
-		monster.AnimInfo.NumberOfFrames = monster.MType->GetAnimData(MonsterGraphic::Stand).Frames;
```
ephphatha, line 4581: `Monster::ChangeAnimationData` will update `AnimInfo.NumberOfFrames` if necessary so I didn't see a reason to duplicate that logic here.

Source/monster.cpp
```diff
@@ -4574,18 +4561,13 @@ void SyncMonsterAnim(Monster &monster)
...
 	default:
 		monster.AnimInfo.CurrentFrame = 1;
-		monster.AnimInfo.NumberOfFrames = monster.MType->GetAnimData(MonsterGraphic::Stand).Frames;
 		break;
 	}
 
-	if (monster.MType->GetAnimData(graphic).CelSpritesForDirections[mdir])
-		monster.AnimInfo.pCelSprite = &*monster.MType->GetAnimData(graphic).CelSpritesForDirections[mdir];
-	else
-		monster.AnimInfo.pCelSprite = nullptr;
```
ephphatha, line 4588: Attempting to render a monster with a `nullptr` for pCelSprite is logged as an error, but I hadn't seen that actually fire in my testing. I can restore this code in `ChangeAnimationData` if it's actually hit on some levels.

Source/towners.cpp
```diff
@@ -57,13 +57,10 @@ void InitTownerInfo(int i, const TownerData &townerData)
...
-void LoadTownerAnimations(Towner &towner, const char *path, int frames, Direction dir, int delay)
+void LoadTownerAnimations(Towner &towner, const char *path, int frames, int delay)
 {
 	towner.data = LoadFileInMem(path);
-	for (auto &animation : towner._tNAnim) {
-		animation = towner.data.get();
-	}
-	NewTownerAnim(towner, towner._tNAnim[dir], frames, delay);
+	NewTownerAnim(towner, towner.data.get(), frames, delay);
 }
```
ephphatha, line 64: Towner::_TNAnim was only ever used as an array in InitCows. For every other towner every index has the same pointer to the sprite sheet and the only read was in this function. Just referencing the data pointer directly saves some dead code and removes the implication that direction actually does something.

Source/towners.cpp
```diff
@@ -765,22 +762,22 @@ void TalkToGirl(Player &player, Towner &girl)
...
+	{ TOWN_DRUNK,   { 71, 84 }, Direction::South,     InitDrunk,     TalkToDrunk       },
+	{ TOWN_WITCH,   { 80, 20 }, Direction::South,     InitWitch,     TalkToWitch       },
+	{ TOWN_BMAID,   { 43, 66 }, Direction::South,     InitBarmaid,   TalkToBarmaid     },
+	{ TOWN_PEGBOY,  { 11, 53 }, Direction::South,     InitBoy,       TalkToBoy         },
+	{ TOWN_COW,     { 58, 16 }, Direction::SouthWest, InitCows,      TalkToCow         },
+	{ TOWN_COW,     { 56, 14 }, Direction::NorthWest, InitCows,      TalkToCow         },
+	{ TOWN_COW,     { 59, 20 }, Direction::North,     InitCows,      TalkToCow         },
+	{ TOWN_COWFARM, { 61, 22 }, Direction::SouthWest, InitCowFarmer, TalkToCowFarmer   },
+	{ TOWN_FARMER,  { 62, 16 }, Direction::South,     InitFarmer,    TalkToFarmer      },
+	{ TOWN_GIRL,    { 77, 43 }, Direction::South,     InitGirl,      TalkToGirl        },
```
ephphatha, line 780: The only time Direction values are relevant here is for `type == TOWN_COW`. I left it in for single-facing towners for consistency.

ephphatha:
@obligaron Would you mind double checking the updated `Monster::ChangeAnimationData` function? I wasn't 100% sure whether setting the celSprite pointer to nullptr (from a non-null value) was actually expected to happen. The only time I saw a nullptr value here was during level load when monsters were being initialised.

Source/missiles.h
```diff
@@ -57,34 +57,39 @@ struct MissilePosition {
...
- *              |
- *     NW ------+-----> SE
- *              |
- *     Nw       |       sE
- *              |
- *      N  Ne  NE   nE  E
+/**
+ * Represent a more fine-grained direction than the 8 value Direction enum.
+ *
+ * This is mainly used in Hellfire specific rendering code to draw sprites for "half-winds" on a 16-point compass.
```
AJenbo, line 63: This isn't really true afaik, the only place where it was specifically used in Hellfire is `AddFirenova()`. Instead it would be more accurate to say that it is exclusively used for rendering projectile missile directions.

Source/missiles.h
```diff
@@ -138,7 +143,34 @@ Direction16 GetDirection16(Point p1, Point p2);
...
+*/
+inline void SetMissDir(Missile &missile, Direction dir)
+{
+	SetMissDir(missile, static_cast<int>(dir));
+}
+
+/**
+ * @brief Sets the sprite for this missile so it matches the given Direction16
+ * @param missile this object
+ * @param dir Desired facing as a hellfire direction
```
AJenbo, line 167: Hellfire did not add this, it did add new flare types with a directional sprite which might be where this assumption comes from.
```suggestion
 * @param dir Desired facing at a 22.8125 degree resolution
```

Source/missiles.h
ephphatha, line 63:
```suggestion
 * This is used when rendering projectiles like arrows which have additional sprites for "half-winds" on a 16-point compass.
```

AJenbo: Yep that looks good :)

Source/monster.cpp
obligaron: If we change `ChangeAnimationData` to also set `pCelSprite` to `nullptr`, it wouldn't be any difference (see comment in `ChangeAnimationData).
The only case this could happen in normal game that comes to my mind is that golems don't have a Stand graphics. So they could end up with `nullptr`, but I'm not sure if this is the case.
If you want you can check it again. 🙂

obligaron (commented):
> @obligaron Would you mind double checking the updated `Monster::ChangeAnimationData` function? I wasn't 100% sure whether setting the celSprite pointer to nullptr (from a non-null value) was actually expected to happen. The only time I saw a nullptr value here was during level load when monsters were being initialised.

I checked again and I would only adjust it slightly (see relevant comment). 🙂

Source/engine/animationinfo.cpp
```diff
@@ -166,15 +166,18 @@ void AnimationInfo::ChangeAnimationData(const CelSprite *celSprite, int numberOf
 {
 	if (numberOfFrames != NumberOfFrames || ticksPerFrame != TicksPerFrame) {
 		// Ensure that the CurrentFrame is still valid and that we disable ADL cause the calculcated values (for example TickModifier) could be wrong
-		CurrentFrame = clamp(CurrentFrame, 1, numberOfFrames);
+		if (numberOfFrames >= 1)
+			CurrentFrame = clamp(CurrentFrame, 1, numberOfFrames);
+		else
+			CurrentFrame = 0;
```
ephphatha, line 172: This is necessary because `clamp(int 1, 0)` is invalid, fails assertions in debug builds and is called out as undefined behaviour for std::clamp - https://en.cppreference.com/w/cpp/algorithm/clamp

julealgon: Would it make sense to add support for this in our own abstraction? We have a polyfill for `clamp` right now right?

Alternatively, would it be possible to map this particular logic to another operation name? Semantically, what are we doing here?

AJenbo: It's a polyfil, so should act as much as the real thing as possible, eventually we should be able to just switch to the real thing without any changes needed

AJenbo:
This looks ready to merge, but I'm to intoxicated to trust myself 😕

obligaron (dismissed):
@AJenbo Have fun, but don't forget to drink some water. 🥳 And for me it looks good. 😉

Source/engine/animationinfo.cpp
ephphatha: > Alternatively, would it be possible to map this particular logic to another operation name? Semantically, what are we doing here?

I'm not familiar enough with the animation code to confidently say why this is the case, but for a animation sequence with 4 frames (0-3) when played "forwards" the animation sequence starts at frame 1 and goes to frame 3, before wrapping back to frame 1. For animations played "backwards" the animation sequence starts at frame 3 and goes to frame 0 before wrapping back around to frame 3.

https://github.com/diasurgical/devilutionX/blob/ec1850e2950f015d7814a8d9472bde471f4cb0d6/Source/engine/animationinfo.cpp#L188-L200

If all sequences started from 0 we could handle the case for numberOfFrames (really the last frame number) == 0 with the original code, `clamp(int, 0, 0)` is valid.

AJenbo (approved):

PR #2732, opened 2021-08-28
Clean up AddRune

AJenbo:
Ok I think I'm ready to move on from missiles.cpp now :D

Runes cannot be learned so will never use mana, MissilesFoundTarget might have been used in some dead code, but now AddRune is the only consumer and merging the two makes more sens as some of the input is always the same.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -610,28 +610,30 @@ void SetMissAnim(MissileStruct &missile, int animtype)
 	missile._miAnimFrame = 1;
 }
 
-bool MissilesFoundTarget(MissileStruct &missile, Point *position, int rad)
+void AddRune(MissileStruct &missile, Point dst, missile_id missileID)
 {
-	rad = std::min(rad, 19);
-	for (int i = 0; i < rad; i++) {
-		int k = CrawlNum[i];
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -610,28 +610,30 @@ void SetMissAnim(MissileStruct &missile, int animtype)
...
+				if (nSolidTable[dp] || dObject[target.x][target.y] != 0 || (dFlags[i][j] & BFLAG_MISSILE) != 0)
+					continue;
 
-			missile.position.tile = { tx, ty };
-			*position = { tx, ty };
-			return true;
+				missile.position.tile = target;
+				missile.var1 = missileID;
+				missile._mlid = AddLight(target, 8);
+				return:
```
obligaron, line 631:
```suggestion
				return;
```

obligaron (commented):
When it's compile I'm fine with it. 🙂

Source/missiles.cpp
AJenbo: Not using my regular system and this thing didn't even want to start compiling 😢

ephphatha (approved):

PR #2734, opened 2021-08-28
Clean up some players references

AJenbo:


ephphatha (changes requested):

Source/items.cpp
```diff
@@ -3769,75 +3762,59 @@ void GetItemStr(ItemStruct &item)
...
-
 	if (cii >= NUM_INVLOC)
-		pi = &Players[pnum].InvList[cii - NUM_INVLOC];
-	else
-		pi = &Players[pnum].InvBody[cii];
+		cii -= NUM_INVLOC;
 
-	pi->_iIdentified = true;
-	CalcPlrInv(pnum, true);
+	player.InvBody[cii]._iIdentified = true;
```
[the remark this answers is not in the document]
AJenbo: wow i compleatly overlooked that it was two different arrays, thanks for catching

Source/items.h
```diff
@@ -409,11 +411,14 @@ bool IsItemAvailable(int i);
...
-void CalcPlrInv(int p, bool Loadgfx);
+void CalcPlrItemVals(PlayerStruct &player, bool Loadgfx);
+void CalcPlrInv(PlayerStruct &player, bool Loadgfx);
 void SetPlrHandItem(ItemStruct &item, int itemData);
 void GetPlrHandSeed(ItemStruct *h);
-void GetGoldSeed(int pnum, ItemStruct *h);
+/**
+ * @brief Set a new unique seed value on the given item
+ */
+void GetGoldSeed(PlayerStruct &player, ItemStruct &gold);
```
obligaron, line 421: The name is misleading. Shouldn't it be `SetGoldSeed`?

```suggestion
/**
 * @brief Set a new unique seed value on the given item
 */
void SetGoldSeed(PlayerStruct &player, ItemStruct &gold);
```

Source/items.cpp
```diff
@@ -28,6 +28,7 @@
 #include "utils/language.h"
 #include "utils/math.h"
 #include "utils/stdcompat/algorithm.hpp"
+#include "player.h"
```
obligaron, line 31: nit: include at alphabetical order?

Source/items.cpp
```diff
@@ -3769,75 +3762,59 @@ void GetItemStr(ItemStruct &item)
...
 
-	if (pnum == MyPlayerId)
+	if (&player == &Players[MyPlayerId])
 		NewCursor(CURSOR_HAND);
 }
 
-void DoOil(int pnum, int cii)
+void DoOil(PlayerStruct &player, int cii)
 {
 	if (cii < NUM_INVLOC && cii != INVLOC_HEAD && (cii <= INVLOC_AMULET || cii > INVLOC_CHEST))
```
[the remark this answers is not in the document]
obligaron: I think this check is entirely not needed. 🙁 
Cause you are also not allowed to apply oil to rings or amulets in inventory.
This check is performed in `ApplyOilToItem` and is also executed when oil is applied to equipped jewellery.

obligaron (commented):
Same issues at @ephphatha and two nits, else lgtm. 😉

Source/items.cpp
AJenbo: I tried :(

obligaron: Don't give up! We believe in you!

Source/items.cpp
```diff
@@ -3769,75 +3762,59 @@ void GetItemStr(ItemStruct &item)
...
 		return;
-	auto &player = Players[pnum];
 	if (!ApplyOilToItem(player.InvBody[cii], player))
```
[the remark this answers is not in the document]
AJenbo: Probably only works cause InvBody and InvList are next to each other

obligaron (approved):

Source/items.cpp
```diff
@@ -3769,28 +3762,26 @@ void GetItemStr(ItemStruct &item)
...
+void CheckIdentify(PlayerStruct &player, int cii)
 {
 	ItemStruct *pi;
 
 	if (cii >= NUM_INVLOC)
-		pi = &Players[pnum].InvList[cii - NUM_INVLOC];
+		pi = &player.InvList[cii - NUM_INVLOC];
 	else
-		pi = &Players[pnum].InvBody[cii];
+		pi = &player.InvBody[cii];
```
ephphatha, line 3772: Nit: Other functions with this code use braces, inconsistent with the style guide for single line cases.

However given you're touching the lines anyway extracting this to a function returning an `ItemStruct&` would be great. Then we know all four functions are using the same logic and the initialisation can also be made simpler.

ephphatha (approved):
No logic issues that I could see, one nit if you felt like it :D

Source/items.cpp
AJenbo: Saw it, but decided i needed to move on and leave something for someone else :)

PR #2739, opened 2021-08-28
Attribute cleanup

qndel:


Source/control.cpp
```diff
@@ -1690,24 +1671,9 @@ void ReleaseChrBtns(bool addAllStatPoints)
...
-				NetSendCmdParam1(true, CMD_ADDDEX, statPointsToAdd);
-				myPlayer._pStatPts -= statPointsToAdd;
-				break;
-			case CharacterAttribute::Vitality:
-				NetSendCmdParam1(true, CMD_ADDVIT, statPointsToAdd);
-				myPlayer._pStatPts -= statPointsToAdd;
-				break;
-			}
+			_cmd_id cmd[] = { CMD_ADDSTR, CMD_ADDMAG, CMD_ADDDEX, CMD_ADDVIT };
+			NetSendCmdParam1(true, cmd[static_cast<int>(attribute)], statPointsToAdd);
```
julealgon, line 1675: Please avoid using the underlying value of the enum like this.

qndel: be glad I didn't do (CMD_ADDSTR + static_cast<int>(attribute)) : P

glebm: This is a common pattern in the Linux kernel. However, Linux kernel uses the enum-keyed arrays GCC/clang extension, making this robust.

In our case, it be better to extract the mapping into a function with a `switch`.

AJenbo (approved):

PR #2743, opened 2021-08-29
Disable hardware cursor for demo playback

obligaron:
This way you always see the cursor of the player and doesn't have to adjust your `diablo.ini` for that.

### DIFF
--- a/Source/engine/demomode.cpp
+++ b/Source/engine/demomode.cpp
@@ -147,6 +147,9 @@ void OverrideOptions()
 	sgOptions.Graphics.nWidth = DemoGraphicsWidth;
 	sgOptions.Graphics.nHeight = DemoGraphicsHeight;
 	sgOptions.Graphics.bFitToScreen = false;
+#if SDL_VERSION_ATLEAST(2, 0, 0)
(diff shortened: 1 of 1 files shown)


Source/engine/demomode.cpp
```diff
@@ -147,6 +147,9 @@ void OverrideOptions()
 	sgOptions.Graphics.nWidth = DemoGraphicsWidth;
 	sgOptions.Graphics.nHeight = DemoGraphicsHeight;
 	sgOptions.Graphics.bFitToScreen = false;
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+	sgOptions.Graphics.bHardwareCursor = false;
+#endif
```
AJenbo, line 152: its only true that is problematic for older versions of sdl

```suggestion
	sgOptions.Graphics.bHardwareCursor = false;
```

obligaron: This was also my first version. 😄 
But then CI reminded me with a compilation error, that `bHardwareCursor` is [not defined](https://github.com/diasurgical/devilutionX/blob/0ab1270dc9118e8ff22fb075600c68ed144ee134/Source/options.h#L66) when SDL version is < 2.

PR #2744, opened 2021-08-29
Level seed related debug commands

obligaron:
Content

- restart can now (optional) set a new seed for a level. This is a replacement for `-r` command line argument.
- new command `seedinfo` that is replacement for `r` and `R` keys
- executed debug commands gets logged to the console (if you want to expect the printed information (for example seed) longer then the info is displayed on screen)

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -6,6 +6,8 @@
 
 #ifdef _DEBUG
 
+#include <sstream>
+
 #include "debug.h"
 
 #include "automap.h"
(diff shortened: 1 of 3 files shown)


AJenbo (approved):

Source/debug.cpp
```diff
@@ -249,10 +257,21 @@ std::string DebugCmdVisitTowner(const std::string_view parameter)
...
+	std::string singleParameter;
+	if (!std::getline(paramsStream, singleParameter, ' '))
+		return "What level do you want to visit?";
+	auto level = atoi(singleParameter.c_str());
 	if (level < 0 || level > (gbIsHellfire ? 24 : 16))
 		return fmt::format("Level {} is not known. Do you want to write an extension mod?", level);
 	myPlayer._pLvlVisited[level] = false;
+
+	if (std::getline(paramsStream, singleParameter, ' ')) {
+		auto seed = atoi(singleParameter.c_str());
```
ephphatha, line 271: Can this use `stoul` from `<string>`? `atoi` returns a signed int, this uses undefined behaviour for seeds larger than int_max.

```suggestion
		uint32_t seed = static_cast<uint32_t>(std::stoul(singleParameter));
```
(The cast isn't necessary on platforms where `sizeof(uint32_t) == sizeof(unsigned long)` but just to be safe)

obligaron: Didn't thought much about it, cause the old command line parameter did the same. 😉 
I opened #2746 to fix it.
Thank you 🙂

PR #2746, opened 2021-08-29
DebugCmdResetLevel: Use stoul for parameter seed parsing

obligaron:
To avoid undefined behavior, see {comment](https://github.com/diasurgical/devilutionX/pull/2744#discussion_r697986644).
Thanks @ephphatha 🙂

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -268,7 +268,7 @@ std::string DebugCmdResetLevel(const std::string_view parameter)
 	myPlayer._pLvlVisited[level] = false;
 
 	if (std::getline(paramsStream, singleParameter, ' ')) {
-		auto seed = atoi(singleParameter.c_str());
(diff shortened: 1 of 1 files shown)


ephphatha (approved):
The only downside of stoul is it throws std::out_of_range if the value is too large to be represented as a long (atoi just returns -1 which isn't any better :/) https://tio.run/##Sy4o0E1PTv7/XzkzLzmnNCVVwSYzv7ikKDUx146LKzOvRCE3MTNPA8jQUUjOSCzS0tJUqObiLC5JsbJKzi8tUbCxUQBzEkvyMzWUDI2MTUzNzC0sDQwtLCygHCVNa2w6ikvyS3PwaSlKLSktylMwsOaq/f8fAA

Given it's a debug command I think a crash on bad input is fine.

PR #2747, opened 2021-08-29
Allow debug commands from command line (with + prefix)

obligaron:
With this you can add debug commands to the command line.
This commands will be executed when you start/load the first game.

Notes:

- `+` signal the start of a debug command in command line. It ends when we find a command that starts with a `-` in the command lines.
- No command or command parameter is allowed to start with `-`, cause this would signal a normal command line argument.
- The commands are only executed for the first started/loaded game, cause some debug commands are global toggles (like god mode).
- The commands are executed at the second game tick, cause the first game tick is special (for example the `changelevel` command would generate level with errors).

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -103,6 +103,7 @@ _monster_id DebugMonsters[10];
 int debugmonsttypes = 0;
 bool debug_mode_key_inverted_v = false;
 bool debug_mode_key_i = false;
+std::vector<std::string> DebugCmdsFromCommandLine;
 #endif
 /** Specifies whether players are in non-PvP mode. */
 bool gbFriendlyMode = true;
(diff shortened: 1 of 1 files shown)


Source/diablo.cpp
```diff
@@ -812,6 +823,10 @@ void RunGameLoop(interface_mode uMsg)
 
 void DiabloParseFlags(int argc, char **argv)
```
AJenbo, line 824: Line 819: `printInConsole(" %-20s %-30s\n", "+<internal command>", "Pass commands to the engine");`

AJenbo (changes requested):
Another big step in to having respectable tech :+1: 

Only issue, it's missing the help line.

Source/diablo.cpp
obligaron: Added 🙂

AJenbo (approved):

PR #2748, opened 2021-08-29
Add GOG installation paths to game data search paths (fixes #2709)

### LINKED ISSUE #2709: Auto detect GOG installation (PC)
Instead of having the user copy the game data, auto detect where GOG has diablo installed

There's a C header which can help implement this: https://github.com/cxong/find_steam_game
(diablo + hellfire's game id is 1412601690 )
```
#include <find_steam_game.h>

char buf[4096];
fsg_get_gog_game_path(buf, "1412601690");
printf("GOG: Diablo found at %s\n", buf);

// prints:
GOG: 1412601690 found at C:\Program Files (x86)\GOG Galaxy\Games\Diablo
```

cxong:


ephphatha:
Adding the file to 3rdParty (instead of Source/utils) will exclude it from clang-format checks. You'll need to add it as a library in CMakeLists.txt as well though.

Source/utils/find_steam_game.h
```diff
@@ -0,0 +1,262 @@
+/*
```
AJenbo, line 1: This file should go in the 3rdParty folder since it's not developed/maintained by us.

Source/init.cpp
```diff
@@ -143,6 +144,12 @@ void init_archives()
 	if (paths[0] == paths[1])
 		paths.pop_back();
 
+	char gogpath[_FSG_PATH_MAX];
+	fsg_get_gog_game_path(gogpath, "1412601690");
+	if (strlen(gogpath) > 0) {
+		paths.emplace_back(std::string(gogpath) + "/");
+		paths.emplace_back(std::string(gogpath) + "/hellfire/");
+	}
```
AJenbo, line 152: Despite this being a cross-platform lib the current version only supports GOG on Windows, and Diablo will probably never be available under other platforms either, so preferably this should be behind `#if defined(_WIN64) || defined(_WIN32)` (same with the include)

AJenbo (commented):
Looks good, thanks for settings this up. I made two comments regarding where the file should be put and also making this Windows only.

Source/init.cpp
cxong: I think this is fine because on other platforms, the library will just return an empty string and no paths will be added

glebm: Let's not link and run extra code on platforms where it won't do anything

cxong: fair enough

CMakeLists.txt
```diff
@@ -643,6 +644,7 @@ target_link_libraries(libdevilutionx PUBLIC
   PKWare
   StormLib
   smacker
+  find_steam_game
```
glebm, line 647: Please only fetch and link the dependency for Windows

CMakeLists.txt
```diff
@@ -290,6 +290,7 @@ add_library(smacker STATIC
   3rdParty/libsmacker/smacker.c)
 target_include_directories(smacker PUBLIC 3rdParty/libsmacker)
 
+add_subdirectory(3rdParty/find_steam_game)
```
AJenbo, line 293:
```suggestion
if(WIN32)
add_subdirectory(3rdParty/find_steam_game)
endif()

```

AJenbo (approved):

Issue #2733, opened 2021-08-28
Demo: Recorded Messages decission


obligaron:
# Introduction

When we record a demo, we need to record what the player does to replay it.
There are multiple alternatives what can be recorded. Each alternative has some pros and cons.
Currently UI/Click-based recording is implemented. But we should explicitly decide what alternative (and with them what pro and contra) we want.
Switching from one alternative to another is expensive (switch implementation, existing replays gets invalid, ...), so we should explicitly decide what alternative we want to be implemented.
This issue can be closed if we have a list of alternatives, a overview of the pro and cons of them and a decision what we want.

Notes to admins: Feel free to edit this post.

# Alternatives

### Description
We record players clicks, keyboard input, mouse movements, ...
The replay simulates the same interactions with the UI.

### Implementation notes
Cause the UI interactions are replayed, we need to know what UI settings (resolution, panel distance, ...) where used when recording the replay. Else the replay would get out of sync with original play (for example click on the wrong UI buttons cause the panel moved to different location).

### Pros

- You can see the original players mouse movements, See how he inspects the inventory or how he interacts with the Griswold.
- Currently implemented 😉 
- UI is also tested

### Cons

- If the UI changes, replays can get invalid (more unstable).
- Replay with different UI or different keyboard bindings is more work and adds more complexity (for example see #2729).
- Controller support needs to be added (for example controller could also uses messages).
- It's hard(er) to show different content then the original player (apm, inventory, monster statistics).
- Demo format is nearly impossible to read, tweak, fix, or adjust.
- Demo files gets bigger, cause more information (mouse movements etc.) are recorded.

### Description
We record players commands. For example move to tile xy, attack monster, pick up item, ...
The replay simulates the same commands.

### Implementation notes
This is similar to how the multiplayer work. It's possible that there are synergies between replay and multiplayer.
Cause we only records actions that have a meaning to game logic, we don't record something like mouse movements. But that also means you will never see mouse movements in the replay.

### Pros

- Doesn't depend on UI aspects and cause of this is more stable.
- Would help the engine, cause this could be a good base to abstract the difference input types (mouse+keyboard, controller, touch). (Note: this could be done even if we don't take this alternative)
- Could perhaps used also for multiplayer.
- Viewing different content (for example another resolution, apm, inventory, monster statistics) then the original player is easier.

### Cons

- You never see original players mouse movements or see how he inspects the inventory or how he interacts with Griswold. It's similar to how you see another player in multiplayer.
[...]

PR #2750, opened 2021-08-29
Demo: Fix rendering jittering

obligaron:
See [comment](https://github.com/diasurgical/devilutionX/issues/2733#issuecomment-907831791).
This could be done better, but before #2733 isn't decided. I wouldn't take that extra step. 😉


AJenbo (approved):
Nice, i was looking at doing something similar when we implemented it, but decided it wasn't worth it at the time :)

Issue #66, opened 2019-03-24
Translation support


liberodark:
Hi,

Have see patch in FR for diablo but on linux with devilutionx how to make that ?

Best Regards

liberodark:
Little other question have possibility to play in 1080 Full Screen ?

AJenbo:
Hi @liberodark 
We currently do not have the translation infrastructure in place, but if you have any suggestions we could talk about what solution would be best. I usually have used Gettext on previous software solutions. It generates a .po file for the translator to work on. There is plenty of software for editing them like PoEdit.

The game will upscale to your current resolution, so yes it will run at 1080p with black bars in the side and bilinear filtering (same as GOG.com's version). That said we are currently working on the renderer to improve the visual output, here amounts having it change the output resolution meaning you will be able to see more of the world at one time (no black bars etc).

liberodark:
Great thank you for your work !

grepwood:
@AJenbo We could acquire gamedata from around the world and do some basic text analysis. The purpose of this exercise is to find out what encoding tables were used for the gamedata.

Analogously, OpenMW and VCMI allow the user to switch between Windows encoding tables 1250 (Eastern European), 1251 (Cyrillic) and 1252 (Western European). I pitched them an idea way back ago that we could do away with manual selection. Morrowind's main data file has a header with a 256 byte comment. In the commercial releases, this comment is in the same language as the rest of the gamedata, so we can compare that 256 byte chunk against known chunks to determine the language and encoding table. This is a little resource hungry since each acknowledged language would take up 256 bytes in resource segment and the language check is O(n). But it takes away hassle from the end user.

If Diablo doesn't use UTF-8, then this is an available solution to compare found gamedata against known gamedata - either by part or checksum and size.

AJenbo:
Diablo always uses ISO_8859-1, changing this would break multiplayer and save game compatability. The only exception is the PS1 release, but we are unable to use the data from this release as it is in a mostly unknown format (and has a lower resolution and frame rate).

There are some errors in Diablo's fonts:
```
005C is / instead of \
2018 includes ‘ from Windows-1252
2019 includes ’ from Windows-1252
00B7 is ? instead of ·
00B8 is ? instead of ¸
00BC is ½ instead of ¼
00BD is ¼ instead of ½
```
(this is based on the UI font)
Extending the fonts to being Windows-1252 should be easy enough if some one is willing to render out the addition letters in a matching style. We could also fix the errors this way.
[...]

PR #2751, opened 2021-08-30
Quest entrance check refactoring

julealgon:
Replaces some low-level X/Y offset checks for quest entrance check with some higher level abstractions.

### DIFF
--- a/Source/quests.cpp
+++ b/Source/quests.cpp
@@ -86,15 +86,11 @@ int lineSpacing;
 int act2finSpacing;
 
 /**
- * Specifies a delta in X-coordinates from the quest entrance for
+ * Specifies a displacement from the quest entrance for
  * which the hover text of the cursor will be visible.
  */
-char questxoff[7] = { 0, -1, 0, -1, -2, -1, -2 };
-/**
- * Specifies a delta in Y-coordinates from the quest entrance for
- * which the hover text of the cursor will be visible.
- */
-char questyoff[7] = { 0, 0, -1, -1, -1, -2, -2 };
+Displacement questEntranceOffsets[7] = { { 0, 0 }, { -1, 0 }, { 0, -1 }, { -1, -1 }, { -2, -1 }, { -1, -2 }, { -2, -2 } };
+
 const char *const QuestTriggerNames[5] = {
 	N_(/* TRANSLATORS: Quest Map*/ "King Leoric's Tomb"),
 	N_(/* TRANSLATORS: Quest Map*/ "The Chamber of Bone"),
@@ -457,7 +453,7 @@ bool ForceQuests()
 			int ql = quest._qslvl - 1;
 
 			for (int j = 0; j < 7; j++) {
-				if (quest.position + Displacement { questxoff[j], questyoff[j] } == cursPosition) {
+				if (quest.position + questEntranceOffsets[j] == cursPosition) {
 					strcpy(infostr, fmt::format(_(/* TRANSLATORS: Used for Quest Portals. {:s} is a Map Name */ "To {:s}"), _(QuestTriggerNames[ql])).c_str());
 					cursPosition = quest.position;
 					return true;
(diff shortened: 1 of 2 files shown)


julealgon:
Depends on https://github.com/diasurgical/devilutionX/pull/2253 being merged first.

AJenbo (dismissed):

Source/quests.cpp
```diff
@@ -885,4 +873,14 @@ bool QuestStruct::IsAvailable()
 	return true;
 }
 
+bool QuestStruct::EntranceBoundaryContains(Point position) const
```
ephphatha, line 876: I think this is the same shape as used for town portal portals, potentially useful to pull this out of the QuestStruct class.

AJenbo:
```
      Start 66: QuestTest.SinglePlayerBadPools
66/98 Test #66: QuestTest.SinglePlayerBadPools ......................***Exception: SegFault  0.02 sec
Note: Google Test filter = QuestTest.SinglePlayerBadPools
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from QuestTest
[ RUN      ] QuestTest.SinglePlayerBadPools

```

julealgon:
> ```
>       Start 66: QuestTest.SinglePlayerBadPools
> 66/98 Test #66: QuestTest.SinglePlayerBadPools ......................***Exception: SegFault  0.02 sec
> Note: Google Test filter = QuestTest.SinglePlayerBadPools
> [==========] Running 1 test from 1 test suite.
> [----------] Global test environment set-up.
> [----------] 1 test from QuestTest
> [ RUN      ] QuestTest.SinglePlayerBadPools
> ```

I have no idea what is happening there. Any hints?

AJenbo:
The change is causing one of the RNG tests to crash the application as best I can tell. You would probably need to run the test locally to figure out exactly what is triggering it.

obligaron:
I run the test locally.
The generated random value of this [line](https://github.com/diasurgical/devilutionX/blob/60e27e2427e4d125b533b0855fb4116ab455209d/Source/quests.cpp#L357) is -2.
So we try to access `QuestGroup1` with a negative index => Undefined behavior. 😭 
This pr changes something that changes the memory layout, so we now don't override some random memory but get a access violation exception.

Is there a way, that we can fix this? So we don't write random memory but still stay compatible? 🤔 

@ephphatha To your notice, cause you wrote the test cases and might be interested. 🙂

ephphatha:
> Is there a way, that we can fix this? So we don't write random memory but still stay compatible? 

I'd been lazy about fixing it with the current rng code (there's a fix in #2261 along with the proposed api).

```cpp
int randomIndex = GenerateRnd(sizeof(QuestGroup1) / sizeof(int));
if (randomIndex >= 0)
	quests[QuestGroup1[randomIndex]]._qactive = QUEST_NOTAVAIL;
randomIndex = GenerateRnd(sizeof(QuestGroup2) / sizeof(int));
if (randomIndex >= 0)
	quests[QuestGroup2[randomIndex]]._qactive = QUEST_NOTAVAIL;
randomIndex = GenerateRnd(sizeof(QuestGroup3) / sizeof(int));
if (randomIndex >= 0)
	quests[QuestGroup3[randomIndex]]._qactive = QUEST_NOTAVAIL;
randomIndex = GenerateRnd(sizeof(QuestGroup4) / sizeof(int));
if (randomIndex >= 0) // always true, QuestGroup4 has two members
	quests[QuestGroup4[randomIndex]]._qactive = QUEST_NOTAVAIL;
```

julealgon:
What are the next steps here? Should I mark this as draft until the fix is made to the RNG logic?

AJenbo:
> What are the next steps here? Should I mark this as draft until the fix is made to the RNG logic?

Either that, or just leave it like this for now.

AJenbo (approved):

Source/quests.cpp
AJenbo: Indeed CheckTown and CheckRportal have the same checks

AJenbo: Plz see https://github.com/diasurgical/devilutionX/pull/2845

PR #2752, opened 2021-08-30
🚚 Rename 'PlayerStruct' to 'Player'

julealgon:
Removing the dumb "Struct" suffix on `PlayerStruct`. Benefits are fairly obvious so I won't go into much detail here.


StephenCWills (approved):
Straightforward rename with no diversions. :+1:

AJenbo (approved):

PR #2753, opened 2021-08-30
🚚 Rename 'PkPlayerStruct' to 'PlayerPack'

julealgon:


julealgon:
Depends on https://github.com/diasurgical/devilutionX/pull/2752.

AJenbo:
do thease really conflict?

StephenCWills:
> do thease really conflict?

The first commit of this PR is the commit in #2752.

AJenbo:
> The first commit of this PR is the commit in #2752.

Was hoping the commits didn't actually conflict so that I could just review and get them merged pretty quick instead of needing a back and forth between each.

Source/multi.cpp
```diff
@@ -32,7 +32,7 @@ bool gbSomebodyWonGameKludge;
 TBuffer sgHiPriBuf;
 char szPlayerDescript[128];
 uint16_t sgwPackPlrOffsetTbl[MAX_PLRS];
-PkPlayerStruct netplr[MAX_PLRS];
+PlayerPack netplr[MAX_PLRS];
```
AJenbo, line 35: How about `CompactPlayer`? Pack seams a bit ambiguous. Could also be `PackedPlayer` (swapping the words is also fine).

julealgon: I searched a bit about this before renaming and "Pack" seems fine in this context:
https://wikidiff.com/pack/package

`CompactPlayer` would be weird to me as it would indicate a "variation of player", as if it was some sort of inheritance. `PackedPlayer` would fall on this category a bit as well IMHO.

If you feel strongly about it, I can rename to whatever you want though.

AJenbo: thinking about it this is acutally used as the payload in some network packages so might just be ok as is

AJenbo (approved):

PR #2758, opened 2021-08-30
debug X toggle

qndel:
Adds a `DebugToggle` debug variable and connects it to currently unused X hotkey - this is to make debugging stuff that needs turning on/off easier, up till now I've always had to cook some code for it myself, then remove it knowing I'd need it again soon. NO MORE 🤪


AJenbo:
😐

PR #2761, opened 2021-08-31
Change monster mode enum into an enum class

julealgon:


Source/diablo.h
```diff
@@ -49,7 +49,7 @@ enum class MouseActionType : int {
 	Spell,
 	SpellMonsterTarget,
 	SpellPlayerTarget,
-	Attack,
+	MeleeAttack,
```
ephphatha, line 52: Was this intended to be part of the MonsterMode changes? Looks like a find-replace casualty

Source/monster.cpp
```diff
@@ -2341,7 +2341,7 @@ void OverlordAi(int i)
 	int v = GenerateRnd(100);
 	if (abs(mx) >= 2 || abs(my) >= 2) {
 		if ((monster._mVar2 > 20 && v < 4 * monster._mint + 20)
-		    || ((monster._mVar1 == MM_WALK || monster._mVar1 == MM_WALK2 || monster._mVar1 == MM_WALK3)
+		    || ((monster._mVar1 == static_cast<int>(MonsterMode::MoveNorthwards) || monster._mVar1 == static_cast<int>(MonsterMode::MoveSouthwards) || monster._mVar1 == static_cast<int>(MonsterMode::MoveSideways))
```
ephphatha, line 2344:
```suggestion
		    || (IsAnyOf(static_cast<MonsterMode>(monster._mVar1), MonsterMode::MoveNorthwards, MonsterMode::MoveSouthwards, MonsterMode::MoveSideways)
```
Cuts down on the explicit casts, and I reckon promoting `monster._mVar1` to `MonsterMode` is preferable to casting away from the enum type.

As a side note this check is repeated multiple times for the different AI procedures, the checks for `Var1 is a walking mode && Var2 == 0` would be useful as a member function of `MonsterStruct`. It seems to be used to check if the monster was walking before it started idling.

ephphatha (commented):
Had a proper look through, looks good apart from a couple of things that seem to be collateral (the commit only mentioned MonsterMode).

Also with the comparisons between Monster::_mVar1 and a MonsterMode value I think casting _mVar1 to MonsterMode is preferable to the reverse. I've only suggested it for the cases where multiple comparisons are performed at once since you can make it shorter and clearer with `IsAnyOf`.

Source/diablo.h
julealgon: This was a mistake. Intended to rename the equivalent entry on `MonsterMode`.

Source/monster.cpp
julealgon: Extracting a function to encapsulate that is out of scope for this. Though I agree, I'd have to check these monster vars in more detail to understand why var2 is being checked as well. Also, we'd have to decide on what to call it.

The `IsAnyOf` change has been implemented.

Source/diablo.h
julealgon: Fixed. There were other renames as well that were incorrect.

Source/missiles.cpp
```diff
@@ -1342,13 +1342,13 @@ void AddBerserk(MissileStruct &missile, Point dst, Direction /*midir*/)
 
 			if (monster._uniqtype != 0 || monster._mAi == AI_DIABLO)
 				continue;
-			if (monster._mmode == MM_FADEIN || monster._mmode == MM_FADEOUT)
+			if (monster._mmode == MonsterMode::FadeIn || monster._mmode == MonsterMode::FadeOut)
```
AJenbo, line 1345:
```suggestion
			if (IsAnyOf(monster._mmode, MonsterMode::FadeIn, MonsterMode::FadeOut))
```

Source/sync.cpp
```diff
@@ -167,12 +167,12 @@ void SyncMonster(int pnum, const TSyncMonster *p)
 	if (monster.position.future.x == p->_mx && monster.position.future.y == p->_my) {
 		return;
 	}
-	if (monster._mmode == MM_CHARGE || monster._mmode == MM_STONE) {
+	if (monster._mmode == MonsterMode::Charge || monster._mmode == MonsterMode::Petrified) {
```
AJenbo, line 170:
```suggestion
	if (IsAnyOf(monster._mmode, MonsterMode::Charge, MonsterMode::Petrified)) {
```

Source/sync.cpp
```diff
@@ -167,12 +167,12 @@ void SyncMonster(int pnum, const TSyncMonster *p)
...
 		return;
 	}
-	if (monster._mmode == MM_CHARGE || monster._mmode == MM_STONE) {
+	if (monster._mmode == MonsterMode::Charge || monster._mmode == MonsterMode::Petrified) {
 		return;
 	}
 
 	if (monster.position.tile.WalkingDistance({ p->_mx, p->_my }) <= 2) {
-		if (monster._mmode < MM_WALK || monster._mmode > MM_WALK3) {
+		if (monster._mmode < MonsterMode::MoveNorthwards || monster._mmode > MonsterMode::MoveSideways) {
```
AJenbo, line 175:
```suggestion
		if (IsNoneOf(monster._mmode, MonsterMode::MoveNorthwards, MonsterMode::MoveSouthwards, MonsterMode::MoveSideways)) {
```

Source/scrollrt.cpp
```diff
@@ -382,8 +382,8 @@ void DrawMonster(const Surface &out, int x, int y, int mx, int my, const Monster
 	int frames = SDL_SwapLE32(frameTable[0]);
 	if (nCel < 1 || frames > 50 || nCel > frames) {
 		const char *szMode = "unknown action";
-		if (monster._mmode <= 17)
-			szMode = MonsterModeNames[monster._mmode];
+		if (static_cast<int>(monster._mmode) <= 17)
```
AJenbo, line 385:
```suggestion
		if (monster._mmode == MonsterMode::Talk)
```

Source/scrollrt.cpp
```diff
@@ -382,8 +382,8 @@ void DrawMonster(const Surface &out, int x, int y, int mx, int my, const Monster
 	int frames = SDL_SwapLE32(frameTable[0]);
 	if (nCel < 1 || frames > 50 || nCel > frames) {
 		const char *szMode = "unknown action";
-		if (monster._mmode <= 17)
-			szMode = MonsterModeNames[monster._mmode];
+		if (static_cast<int>(monster._mmode) <= 17)
+			szMode = MonsterModeNames[static_cast<int>(monster._mmode)];
```
AJenbo, line 386: Could we turn this in to a switch-return function instead of and array with an enum as key?

Source/missiles.cpp
```diff
@@ -3777,15 +3777,15 @@ void MI_Stone(MissileStruct &missile)
...
+	if (monster._mmode != MonsterMode::Petrified) {
 		missile._miDelFlag = true;
 		return;
 	}
 
 	if (missile._mirange == 0) {
 		missile._miDelFlag = true;
 		if (monster._mhitpoints > 0) {
-			monster._mmode = (MON_MODE)missile.var1;
+			monster._mmode = (MonsterMode)missile.var1;
```
AJenbo, line 3788:
```suggestion
			monster._mmode = static_cast<MonsterMode>(missile.var1);
```

Source/monster.cpp
```diff
@@ -5007,24 +5007,24 @@ void decode_enemy(MonsterStruct &monster, int enemy)
 
 void MonsterStruct::CheckStandAnimationIsLoaded(Direction mdir)
 {
-	if (_mmode == MM_STAND || _mmode == MM_TALK) {
+	if (_mmode == MonsterMode::Stand || _mmode == MonsterMode::Talk) {
```
AJenbo, line 5010:
```suggestion
	if (IsAnyOf(_mmode, MonsterMode::Stand, MonsterMode::Talk)) {
```

Source/monster.cpp
```diff
@@ -4188,16 +4188,16 @@ void GolumAi(int i)
 		return;
 	}
 
-	if (golem._mmode == MM_DEATH
-	    || golem._mmode == MM_SPSTAND
-	    || (golem._mmode >= MM_WALK && golem._mmode <= MM_WALK3)) {
+	if (golem._mmode == MonsterMode::Death
+	    || golem._mmode == MonsterMode::SpecialStand
+	    || (golem._mmode >= MonsterMode::MoveNorthwards && golem._mmode <= MonsterMode::MoveSideways)) {
```
AJenbo, line 4193:
```suggestion
	if (IsAnyOf(golem._mmode, MonsterMode::Death, MonsterMode::SpecialStand, MonsterMode::MoveNorthwards, MonsterMode::MoveSouthwards, MonsterMode::MoveSideways)) {
```

Source/monster.cpp
```diff
@@ -3026,13 +3026,13 @@ void SnakeAi(int i)
...
-		if (abs(mx) < 3 && abs(my) < 3 && LineClear([&monster](Point position) { return IsTileAvailable(monster, position); }, monster.position.tile, { fx, fy }) && monster._mVar1 != MM_CHARGE) {
+		if (abs(mx) < 3 && abs(my) < 3 && LineClear([&monster](Point position) { return IsTileAvailable(monster, position); }, monster.position.tile, { fx, fy }) && monster._mVar1 != static_cast<int>(MonsterMode::Charge)) {
 			if (AddMissile(monster.position.tile, { fx, fy }, md, MIS_RHINO, TARGET_PLAYERS, i, 0, 0) != -1) {
 				PlayEffect(monster, 0);
 				dMonster[monster.position.tile.x][monster.position.tile.y] = -(i + 1);
-				monster._mmode = MM_CHARGE;
+				monster._mmode = MonsterMode::Charge;
 			}
-		} else if (monster._mVar1 == MM_DELAY || GenerateRnd(100) >= 35 - 2 * monster._mint) {
+		} else if (monster._mVar1 == static_cast<int>(MonsterMode::Delay) || GenerateRnd(100) >= 35 - 2 * monster._mint) {
```
AJenbo, line 3035:
```suggestion
		} else if (static_cast<MonsterMode>(monster._mVar1) == MonsterMode::Delay || GenerateRnd(100) >= 35 - 2 * monster._mint) {
```

Source/monster.cpp
```diff
@@ -3026,13 +3026,13 @@ void SnakeAi(int i)
 	Direction md = GetDirection(monster.position.tile, monster.position.last);
 	monster._mdir = md;
 	if (abs(mx) >= 2 || abs(my) >= 2) {
-		if (abs(mx) < 3 && abs(my) < 3 && LineClear([&monster](Point position) { return IsTileAvailable(monster, position); }, monster.position.tile, { fx, fy }) && monster._mVar1 != MM_CHARGE) {
+		if (abs(mx) < 3 && abs(my) < 3 && LineClear([&monster](Point position) { return IsTileAvailable(monster, position); }, monster.position.tile, { fx, fy }) && monster._mVar1 != static_cast<int>(MonsterMode::Charge)) {
```
AJenbo, line 3029:
```suggestion
		if (abs(mx) < 3 && abs(my) < 3 && LineClear([&monster](Point position) { return IsTileAvailable(monster, position); }, monster.position.tile, { fx, fy }) && static_cast<MonsterMode>(monster._mVar1) != MonsterMode::Charge) {

```

Source/monster.cpp
```diff
@@ -2405,7 +2405,7 @@ void SkeletonBowAi(int i)
 
 	if (abs(mx) < 4 && abs(my) < 4) {
 		if ((monster._mVar2 > 20 && v < 2 * monster._mint + 13)
-		    || ((monster._mVar1 == MM_WALK || monster._mVar1 == MM_WALK2 || monster._mVar1 == MM_WALK3)
+		    || (IsAnyOf(static_cast<MonsterMode>(monster._mVar1), MonsterMode::MoveNorthwards, MonsterMode::MoveSouthwards, MonsterMode::MoveSideways)
```
AJenbo, line 2408: Having an `IsWalking()` method like on player would help to clean up this pattern.

Source/monster.cpp
```diff
@@ -2369,13 +2369,13 @@ void SkeletonAi(int i)
...
 	if (abs(x) >= 2 || abs(y) >= 2) {
-		if (monster._mVar1 == MM_DELAY || (GenerateRnd(100) >= 35 - 4 * monster._mint)) {
+		if (monster._mVar1 == static_cast<int>(MonsterMode::Delay) || (GenerateRnd(100) >= 35 - 4 * monster._mint)) {
 			RandomWalk(i, md);
 		} else {
 			AiDelay(monster, 15 - 2 * monster._mint + GenerateRnd(10));
 		}
 	} else {
-		if (monster._mVar1 == MM_DELAY || (GenerateRnd(100) < 2 * monster._mint + 20)) {
+		if (monster._mVar1 == static_cast<int>(MonsterMode::Delay) || (GenerateRnd(100) < 2 * monster._mint + 20)) {
```
AJenbo, line 2378:
```suggestion
		if (static_cast<MonsterMode>(monster._mVar1) == MonsterMode::Delay || (GenerateRnd(100) < 2 * monster._mint + 20)) {
```

Source/monster.cpp
```diff
@@ -2369,13 +2369,13 @@ void SkeletonAi(int i)
 	Direction md = GetDirection(monster.position.tile, monster.position.last);
 	monster._mdir = md;
 	if (abs(x) >= 2 || abs(y) >= 2) {
-		if (monster._mVar1 == MM_DELAY || (GenerateRnd(100) >= 35 - 4 * monster._mint)) {
+		if (monster._mVar1 == static_cast<int>(MonsterMode::Delay) || (GenerateRnd(100) >= 35 - 4 * monster._mint)) {
```
AJenbo, line 2372:
```suggestion
		if (static_cast<MonsterMode>(monster._mVar1) == MonsterMode::Delay || (GenerateRnd(100) >= 35 - 4 * monster._mint)) {
```

Source/monster.cpp
```diff
@@ -2205,13 +2205,13 @@ void AiRanged(int i, missile_id missileType, bool special)
 		if (monster._msquelch < UINT8_MAX)
 			MonstCheckDoors(monster);
 		monster._mdir = md;
-		if (monster._mVar1 == MM_RATTACK) {
+		if (monster._mVar1 == static_cast<int>(MonsterMode::RangedAttack)) {
```
AJenbo, line 2208:
```suggestion
		if (static_cast<MonsterMode>(monster._mVar1) == MonsterMode::RangedAttack) {
```

Source/monster.cpp
```diff
@@ -3060,8 +3060,8 @@ void SnakeAi(int i)
 			AiDelay(monster, 15 - monster._mint + GenerateRnd(10));
 		}
 	} else {
-		if (monster._mVar1 == MM_DELAY
-		    || monster._mVar1 == MM_CHARGE
+		if (monster._mVar1 == static_cast<int>(MonsterMode::Delay)
+		    || monster._mVar1 == static_cast<int>(MonsterMode::Charge)
```
AJenbo, line 3064:
```suggestion
		if (IsAnyOf(static_cast<MonsterMode>(monster._mVar1), MonsterMode::Delay, MonsterMode::Charge))
```

AJenbo:
There are a few cases where we could better handle things as `MonsterMode` rather then casting it to an `int`, other then that it looks good.

Source/monster.cpp
AJenbo: I suppose `oldMode` would follow the original convention, but `previousMode` sounds better.

Source/monster.cpp
julealgon: There is already one, but it deals with the actual state of the monster. `_mVar1` and `_mVar2` are not the current state. This is the same as what @ephphatha commented previously and I already replied back.

julealgon:
> There are a few cases where we could better handle things as MonsterMode rather then casting it to an int, other then that it looks good.

So you are saying that only those casts to `int` are the blocker aspects of this change? Because you are also suggesting a variety of other refactoring changes that would apply to the old code as well and are not caused by this.

If possible, please make it clear what you consider blocker vs what would be a nice to have (I used to send 2 distinct reviews at work, one with a "needs change" and another with a "comment").

If we can't focus on only one thing at a time, every change will be a near endless stream of refactoring on top of refactoring.

AJenbo:
Required when changing enum to class enum:
- Remove C-style casts
- Do not treat it as an int
- Avoid casting it to int unless strictly needed

Optional:
- Apply IsAnyOf/IsNoneOf when only dealing with `==||==` || `!=&&!=`

That being said, I think we are capable of keeping things finite.

Source/missiles.cpp
AJenbo: optional

Source/sync.cpp
AJenbo: optional

Source/sync.cpp
AJenbo: required

Source/scrollrt.cpp
AJenbo: Optional, but greatly appreciated.

Source/missiles.cpp
AJenbo: required

Source/monster.cpp
AJenbo: Second half is required:
```suggestion
	if (golem._mmode == MonsterMode::Death
	    || golem._mmode == MonsterMode::SpecialStand
	    || IsAnyOf(golem._mmode, MonsterMode::MoveNorthwards, MonsterMode::MoveSouthwards, MonsterMode::MoveSideways)) {
```

Source/monster.cpp
AJenbo: required

Source/monster.cpp
AJenbo: required

Source/monster.cpp
AJenbo: I thought that was more about how to handle `var1`, I actually didn't really notice that it was var1 and not `MMode` when making this comment.

Source/monster.cpp
AJenbo: required

Source/monster.cpp
AJenbo: required

Source/monster.cpp
AJenbo: First half is required
```suggestion
		if (static_cast<MonsterMode>(monster._mVar1) == MonsterMode::Delay
		    || static_cast<MonsterMode>(monster._mVar1) == MonsterMode::Charge
```

Source/monster.cpp
```diff
@@ -3124,7 +3124,7 @@ void CounselorAi(int i)
 				monster._mgoal = MGOAL_RETREAT;
 				monster._mgoalvar1 = 0;
 				StartFadeout(monster, md, false);
-			} else if (monster._mVar1 == MM_DELAY
+			} else if (monster._mVar1 == static_cast<int>(MonsterMode::Delay)
```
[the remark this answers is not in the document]
AJenbo: required

Source/player.cpp
```diff
@@ -3207,7 +3207,7 @@ void RemovePlrMissiles(int pnum)
 		int am = ActiveMissiles[i];
 		auto &missile = Missiles[am];
 		if (missile._mitype == MIS_STONE && missile._misource == pnum) {
-			Monsters[missile.var2]._mmode = (MON_MODE)missile.var1;
+			Monsters[missile.var2]._mmode = (MonsterMode)missile.var1;
```
ephphatha, line 3210:
```suggestion
			Monsters[missile.var2]._mmode = static_cast<MonsterMode>(missile.var1);
```

Source/scrollrt.cpp
```diff
@@ -377,13 +356,74 @@ void DrawMonster(const Surface &out, int x, int y, int mx, int my, const Monster
...
+			return "petrified";
+
+		case MonsterMode::Heal:
+			return "healing";
+
+		case MonsterMode::Talk:
+			return "talking";
+
+		default:
+			app_fatal("Invalid monster mode.");
```
ephphatha, line 416: This is only used for the log message, might as well return "unknown action" instead of triggering a fatal error,

Source/scrollrt.cpp
```diff
@@ -377,13 +356,74 @@ void DrawMonster(const Surface &out, int x, int y, int mx, int my, const Monster
...
+
 	int nCel = monster.AnimInfo.GetFrameToUseForRendering();
 	const auto *frameTable = reinterpret_cast<const uint32_t *>(monster.AnimInfo.pCelSprite->Data());
 	int frames = SDL_SwapLE32(frameTable[0]);
 	if (nCel < 1 || frames > 50 || nCel > frames) {
 		const char *szMode = "unknown action";
-		if (monster._mmode <= 17)
-			szMode = MonsterModeNames[monster._mmode];
+		if (static_cast<int>(monster._mmode) <= 17)
+			szMode = getMonsterModeDisplayName(monster._mmode);
```
ephphatha, line 426: That way this check becomes unnecessary, szMode can be initialised directly from a call to `getMonsterModeDisplayName`

Source/monster.cpp
```diff
@@ -4188,16 +4187,14 @@ void GolumAi(int i)
 		return;
 	}
 
-	if (golem._mmode == MM_DEATH
-	    || golem._mmode == MM_SPSTAND
-	    || (golem._mmode >= MM_WALK && golem._mmode <= MM_WALK3)) {
+	if (IsAnyOf(golem._mmode, MonsterMode::Death, MonsterMode::SpecialStand) || golem.IsWalking()) {
```
ephphatha, line 4190: Good pickup :D

ephphatha (approved):
Apart from the one leftover C-style cast this looks good. There's a potential simplification for the get mode description logic but your choice if that's worth implementing.

Source/scrollrt.cpp
julealgon: It should not be possible to ever hit this since there is a check right before it. I believe stopping is the best course of action here.

Source/scrollrt.cpp
julealgon: It's not the same behavior, though. If we change the function to return something, we are now assigning it on all code paths.

I kept the behavior as before.

ephphatha: The old backing type was unsigned, -1 would be treated as a large positive value and get excluded by this check. Casting to int here allows negative values to pass this check if `_mmode` is set to some garbage value or a marker value.

szMode is already assigned a value on every code path, using the existing default value as the default case in the switch doesn't change behaviour compared to the old implementation.

AJenbo: I would be fine reducing this to :
`const char *szMode = getMonsterModeDisplayName(monster._mmode);`

Since we are already logging a bad animation state, potentially shutting down if the AI is also in a bad state seams reasonable. Funtime is likely over in either case.

This change is optional, though it get rid of a cast to int.

Source/scrollrt.cpp
AJenbo: nice work here :)

I would also have gone with `return "unknown action"` but it's an error case on top of an error case and I have never seen the first one so I don't think halting here is an issue.

AJenbo (dismissed):

AJenbo:
Thanks for cleaning up the scrollrt.cpp logic related to this :+1:

AJenbo (approved):

PR #2762, opened 2021-08-31
Avoid segfaults when using randomly generated array indexes in items.cpp

ephphatha:
The first couple of commits set up helpers for future checks. Items.cpp was one of the easier wins, drlg1.cpp is probably the worst offender so I'm trying to avoid that one as long as possible 😂

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -394,6 +394,8 @@ int ItemsGetCurrlevel()
 
 bool ItemPlace(Point position)
 {
+	if (!InDungeonBounds(position))
+		return false;
 	if (dMonster[position.x][position.y] != 0)
 		return false;
 	if (dPlayer[position.x][position.y] != 0)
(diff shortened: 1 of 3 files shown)


Source/items.cpp
```diff
@@ -1445,7 +1459,7 @@ int RndUItem(MonsterStruct *monster)
 		}
 	}
 
-	return ril[GenerateRnd(ri)];
+	return ril[GenerateNonNegativeRnd(ri)];
```
ephphatha, line 1462: This seemed like the safest way to handle this, I don't think reading invalid memory was actually useful for any generated items but it may break compatibility. That said, with the way it was working previously any unrelated code change potentially could've broken compatibility, no way to know without tests for the exceptional case.

Source/items.cpp
```diff
@@ -3214,17 +3225,21 @@ bool ItemSpaceOk(Point position)
...
+	if (!InDungeonBounds(testPosition))
 		return false;
+	oi = dObject[testPosition.x][testPosition.y];
+	if (oi != 0) {
+		if (Objects[abs(oi) - 1]._oSelFlag != 0)
+			return false;
+	}
 
 	if (dObject[position.x + 1][position.y] > 0
 	    && dObject[position.x][position.y + 1] > 0
```
ephphatha, line 3245: This set of checks only looks at positive object indexes, not negative encoded ones like the previous checks. Not sure if this is a bug or what the reasoning behind it is.

ephphatha:
Lets fix all these signed/unsigned comparisons 🙃

PR #2773, opened 2021-09-01
Debug commands: Fix level transitions

### LINKED ISSUE #2765: Dungeon items appearing in town
do changelevel 0
image

[single_33.zip](https://github.com/diasurgical/devilutionX/files/7086585/single_33.zip)

Must be dItem/Items/whatever not being cleared

obligaron:
Could fix #2765 😉 

Background:
We start the level transition with `StartNewLvl`.
But `StartNewLvl` only `PostMessage` a level transition to the message queue.
That means that the level transition happens a lot later.
But we change the `currlevel` direclty after `StartNewLvl`.
This is a problem for the save game logic.

When you transit from level to another following code is executed:
```
		IncProgress();
		if (!gbIsMultiplayer) {
			SaveLevel();
		} else {
			DeltaSaveLevel();
		}
		IncProgress();
		FreeGameMem();
		currlevel--;
		leveltype = gnLevelTypeTbl[currlevel];
		assert(myPlayer.plrlevel == currlevel);
		IncProgress();
		LoadGameLevel(false, ENTRY_PREV);
		IncProgress();
```

First is the current level saved and `SaveLevel` uses `currlevel`.
That means we set store items, monster etc. in the wrong level.

Additional fixes:

- Not available maps are can't be accessed by `map` command
- If you leave a map the normal way, it should port you back to the expected position and not (0, 0). 😉

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -177,7 +177,7 @@ std::string DebugCmdWarpToLevel(const string_view parameter)
 	if (!setlevel && myPlayer.plrlevel == level)
 		return fmt::format("I did nothing but fulfilled your wish. You are already at level {}.", level);
 
-	StartNewLvl(MyPlayerId, (level != 21) ? interface_mode::WM_DIABNEXTLVL : interface_mode::WM_DIABTWARPUP, level);
+	StartNewLvl(MyPlayerId, (level != 21) ? interface_mode::WM_DIABNEXTLVL : interface_mode::WM_DIABTOWNWARP, level);
 	return fmt::format("Welcome to level {}.", level);
 }
 
@@ -204,7 +204,7 @@ std::string DebugCmdLoadMap(const string_view parameter)
 		if (level != quest._qslvl)
 			continue;
 
-		StartNewLvl(MyPlayerId, (quest._qlevel != 21) ? interface_mode::WM_DIABNEXTLVL : interface_mode::WM_DIABTWARPUP, quest._qlevel);
+		StartNewLvl(MyPlayerId, (quest._qlevel != 21) ? interface_mode::WM_DIABNEXTLVL : interface_mode::WM_DIABTOWNWARP, quest._qlevel);
 		ProcessMessages();
 
 		setlvltype = quest._qlvltype;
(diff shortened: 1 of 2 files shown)


Source/interfac.cpp
```diff
@@ -358,6 +359,7 @@ void ShowProgress(interface_mode uMsg)
 		}
 		IncProgress();
 		FreeGameMem();
+		setlevel = false;
```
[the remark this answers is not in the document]
obligaron: `setlevel = false;` is only set in the case for `WM_DIABTWARPUP` and `WM_DIABNEXTLVL`.
But I tested again and I think `WM_DIABTWARPUP` is wrong, cause it is the command for going up from dungeon to town.
I replaced it with correct `WM_DIABTOWNWARP`.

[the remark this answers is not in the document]
AJenbo: Oh your right, it's just GitHub having some issues regarding indentation lately:

image

Source/debug.cpp
```diff
@@ -184,22 +197,23 @@ std::string DebugCmdLoadMap(const string_view parameter)
...
 	if (level < 1)
 		return fmt::format("Map id must be 1 or higher", level);
-	if (setlevel && myPlayer.plrlevel == level)
-		return fmt::format("I did nothing but fulfilled your wish. You are already at level {}.", level);
+	if (setlevel && setlvlnum == level)
+		return fmt::format("I did nothing but fulfilled your wish. You are already at mapid {}.", level);
 
 	for (auto &quest : Quests) {
-		if (level != quest._qslvl)
+		if (level != quest._qslvl || quest._qactive == quest_state::QUEST_NOTAVAIL)
```
AJenbo, line 204: Won't this result in it reporting maps for inactive quests to be reported as unknown?

I's the a good reason we should not allow the player to jump to a level from an inactive quesets when debugging?

qndel: inactive quest maps have no exit so you end up at 0,0 if you leave them manually

AJenbo: That isn't really an issue when debugging, just `changelevel 0` and your good.

obligaron: Personally I would disable the map if the quest is not active.
But I don't want to argue too much about a debug command, so I removed the commit. 😉

PR #2774, opened 2021-09-01
small debug item spawn cleanup

qndel:


AJenbo (approved):

Source/items.cpp
```diff
@@ -4901,17 +4899,14 @@ std::string DebugSpawnItem(std::string itemName, bool unique)
...
 			continue;
 
-		int uper = (unique ? 15 : 1);
-
 		Point bkp = item.position;
 		memset(&item, 0, sizeof(Item));
 		item.position = bkp;
 		memset(UniqueItemFlags, 0, sizeof(UniqueItemFlags));
-		SetupAllItems(item, idx, AdvanceRndSeed(), fake_m.mLevel, uper, onlygood, false, false);
+		SetupAllItems(item, idx, AdvanceRndSeed(), fake_m.mLevel, (unique ? 15 : 1), false, false, false);
```
julealgon, line 4909: Removing the variables makes this call harder to read. The variables provided some semantic value while just the raw values like this provide nothing.

qndel: It's a debug function, it's fine if it's a bit of voodoo

julealgon: Perhaps we could name the parameters here using the comment style that @glebm used in a few places?

PR #2780, opened 2021-09-02
Convert 'MonsterClass' enum to enum class

julealgon:


Source/monster.cpp
```diff
@@ -3481,11 +3481,11 @@ void BoneDemonAi(int i)
...
-	case MC_ANIMAL:
+	case MonsterClass::Animal:
 		return _("Animal");
-	case MC_DEMON:
+	case MonsterClass::Demon:
 		return _("Demon");
-	case MC_UNDEAD:
+	case MonsterClass::Undead:
 		return _("Undead");
 	}
```
StephenCWills, line 3490: Should we use the same error logic here as we do in `getBorderColor()`? Or at the very least, for the line that comes right after this switch block, should we avoid treating `mMonstClass` as an integer to use it in the error message?

https://github.com/diasurgical/devilutionX/blob/04deb99b93289004da22c50168c4d1d8c1993220/Source/monster.cpp#L3492

julealgon: I don't disagree but that's completely out of scope for this change.

StephenCWills: I don't know how you figure that, seeing as how "Do not treat it as an int" is one of the requirements of this change, and that line of code has literally everything to do with the one field this PR changes.

julealgon: I don't understand what you want to do here. We can't "not treat it as an integer" when the value is not defined in the enumeration. The int value is being emitted here to facilitate debugging, to see the real value and try to detect where it is coming from.

If anything, the other method should follow this one, not the other way around.

julealgon: Changed the other method to comply with this one.

StephenCWills (approved):

PR #2783, opened 2021-09-02
search cleanup

qndel:


Source/missiles.cpp
```diff
@@ -1689,10 +1689,9 @@ void AddRing(Missile &missile, Point /*dst*/, Direction /*midir*/)
 
 void AddSearch(Missile &missile, Point /*dst*/, Direction /*midir*/)
 {
-	missile.var1 = missile._misource;
```
julealgon, line 1692: Are you sure this `var1` is not read for this value at some other point after this call?

qndel: checked now :P

AJenbo (approved):

Source/missiles.cpp
```diff
@@ -3332,7 +3331,7 @@ void MI_Search(Missile &missile)
 		return;
 
 	missile._miDelFlag = true;
-	PlaySfxLoc(IS_CAST7, Players[missile.var1].position.tile);
+	PlaySfxLoc(IS_CAST7, Players[missile._misource].position.tile);
```
AJenbo, line 3334: Technically this change causes the sound to always be played for player 0 if the game is loaded back in the original game... but that just happens to be the ide of the player in single player.

AJenbo:
Likely the reason for it being written like this is that player id used to be passed as a parameter and the hellfire devs weren't aware that the property was also available on the missile.

PR #2786, opened 2021-09-02
add enums to monsters

qndel:


AJenbo (approved):

Source/monstdat.cpp
```diff
@@ -17,145 +17,145 @@ const MonsterDataStruct MonsterData[] = {
...
+// TRANSLATORS: Monster Block end
+/* MT_NAKRUL  */ { "Na-Krul",                     "Monsters\\Nkr\\Nkr%c.CL2",         "Monsters\\newsfx\\Nakrul%c%i.WAV",  nullptr,                             226,   1200, true,        true,        false,     {  2,  6, 16,  3, 16, 16 }, { 0, 0, 0, 0, 0, 0 },        60,       60,     40,   1332,   1332, AI_SKELSD,                   MFLAG_KNOCKBACK | MFLAG_SEARCH | MFLAG_CAN_OPEN_DOOR,    3,  150,      7,         40,         50,   150,      10,          40,          50,         125, MC_DEMON,    IMMUNE_MAGIC | IMMUNE_FIRE  | RESIST_LIGHTNING | IMMUNE_NULL_40 , IMMUNE_MAGIC | IMMUNE_FIRE | IMMUNE_LIGHTNING | IMMUNE_NULL_40    ,  7,                   0,        13333 },
```
julealgon, line 158: I'd like to once again ask you guys to reconsider using a map for this kind of stuff. Allowing the use of maps in the codebase would drastically improve readability in several places.

Adding commented code is just silly to me.

@glebm / @AJenbo

qndel: but it comes at a performance cost and this is free

AJenbo: @julealgon I don't see why this is relevant in this PR? It has always been the case that this list should be replaced by an easily modifiable data file rather then be hard-coded. But no one has done the work do to so yet. Implementing it using a map is fine as long as it mainly effects load time, or has a negligible effect. But this PR is simply about making the current list readable.

AJenbo (approved):

Source/monstdat.cpp
julealgon: > It has always been the case that this list should be replaced by an easily modifiable data file rather then be hard-coded. But no one has done the work do to so yet.

The same could be said for numerous other parts of the codebase (most of it needs to be redone one way or another). I don't think this is an argument for accepting or perpetuating bad code since we don't know when the work will be done.

> Implementing it using a map is fine as long as it mainly effects load time, or has a negligible effect.

This is good to hear. It is a change of heart compared to previous discussions where it was 100% prohibited.

> But this PR is simply about making the current list readable.

By introducing extra technical debt in the process.

AJenbo: > The same could be said for numerous other parts of the codebase (most of it needs to be redone one way or another).

... I think we are saying just that.

> I don't think this is an argument for accepting or perpetuating bad code since we don't know when the work will be done.

I would rather have better documented static arrays today then hold back any changes until someone submits the perfect PR, "don't let perfect stand in the way of better".

> This is good to hear. It is a change of heart compared to previous discussions where it was 100% prohibited.

I think this is based on a miss understanding

> By introducing extra technical debt in the process.

I do not agree that this is adding technical debt.

PR #2788, opened 2021-09-03
♻️ Replace 'ViewX' and 'ViewY' globals with single 'ViewPosition'

julealgon:


Source/drlg_l4.cpp
```diff
@@ -244,8 +244,7 @@ bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool
 		Quests[Q_BETRAYER].position = { sx + 1, sy + 1 };
 	}
 	if (setview) {
-		ViewX = 2 * sx + 21;
-		ViewY = 2 * sy + 22;
+		ViewPosition = { 2 * sx + 21, 2 * sy + 22 };
```
StephenCWills, line 247: This doesn't match the pattern you used for other similar lines.
```suggestion
		ViewPosition = { 21, 22 } + Displacement { sx, sy } * 2;
```

StephenCWills (changes requested):

StephenCWills (approved):

PR #2789, opened 2021-09-03
Fix NULL Cel Buffer error when blocking missiles

### LINKED ISSUE #2770: Missiles do not get properly purged when blocked
**Describe**
Blocking a missile causes it to be removed, but does not clear it from the map

**To Reproduce**
1. Fight a succubus
2. Block a flare projectile
4. See error `INFO: Draw Missile 2 type 24: NULL Cel Buffer`

**Expected behavior**
The missile references should be cleared so that the render no longer attempts to draw it

**Additional context**
This is likely a vanilla issue

StephenCWills:
Fixes the off-by-one error in the missile blocking logic that was causing Blood Star missiles to disappear with a `NULL Cel Buffer` error after being blocked. For more info, see https://github.com/diasurgical/devilutionX/issues/2770#issuecomment-910369255. Tested using the save from #1531.

This resolves #2770

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -476,7 +476,7 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
 					int mAnimFAmt = MissileSpriteData[missile._miAnimType].animFAmt;
 					if (dir < 0)
 						dir = mAnimFAmt - 1;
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Nice work figuring this one out :+1:

PR #2791, opened 2021-09-04
New game fonts

AJenbo:
**To do**
- [x] Implement TRN for Art (PCX) content, and use it for better color rendering
- [x] Adjust brightness of fonts
- [x] Adjust brightness of font bevel heightlight
- [x] Fix missing pixels in fonts
- [x] Add the fonts to the MPQ
- [x] Adjust vertical alignment of text placed using `Point`
- [x] Allow for | as the cursor symbol in main menus.
- [x] Evaluate if font16 can be replaced by font12


Source/DiabloUI/art.cpp
```diff
@@ -131,6 +131,8 @@ void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette)
 	}
 	SFileCloseFileThreadSafe(handle);
 
+	SDL_SetSurfaceRLE(&*artSurface, 1);
```
AJenbo, line 134:
```suggestion
#ifndef USE_SDL1
	SDL_SetSurfaceRLE(&*artSurface, 1);
#endif
```

Source/DiabloUI/selconn.cpp
```diff
@@ -54,27 +53,27 @@ void SelconnLoad()
...
 	SDL_Rect rect6 = { (Sint16)(PANEL_LEFT + 35), (Sint16)(UI_OFFSET_Y + 393), DESCRIPTION_WIDTH, 21 };
 	vecSelConnDlg.push_back(std::make_unique<UiArtText>(selconn_Gateway, rect6, UiFlags::AlignCenter));
 
 	SDL_Rect rect7 = { (Sint16)(PANEL_LEFT + 300), (Sint16)(UI_OFFSET_Y + 211), 295, 33 };
-	vecSelConnDlg.push_back(std::make_unique<UiArtText>(_("Select Connection"), rect7, UiFlags::AlignCenter | UiFlags::FontBig));
+	vecSelConnDlg.push_back(std::make_unique<UiArtText>(_("Select Connection"), rect7, UiFlags::AlignCenter | UiFlags::FontSize30, 3));
 
 	SDL_Rect rect8 = { (Sint16)(PANEL_LEFT + 16), (Sint16)(UI_OFFSET_Y + 427), 250, 35 };
-	vecSelConnDlg.push_back(std::make_unique<UiArtTextButton>(_("Change Gateway"), nullptr, rect8, UiFlags::AlignCenter | UiFlags::VerticalCenter | UiFlags::FontBig | UiFlags::ColorGold | UiFlags::ElementHidden));
+	vecSelConnDlg.push_back(std::make_unique<UiArtTextButton>(_("Change Gateway"), nullptr, rect8, UiFlags::AlignCenter | UiFlags::VerticalCenter | UiFlags::FontSize30 | UiFlags::ColorGold));
```
StephenCWills, line 68: This button has no action and therefore crashes the game when selected. I think it should have stayed hidden.

Source/DiabloUI/selconn.cpp
```diff
@@ -54,27 +53,27 @@ void SelconnLoad()
 	vecSelConnDlg.push_back(std::make_unique<UiArtText>(_("Requirements:"), rect3));
 
 	SDL_Rect rect4 = { (Sint16)(PANEL_LEFT + 35), (Sint16)(UI_OFFSET_Y + 275), DESCRIPTION_WIDTH, 66 };
-	vecSelConnDlg.push_back(std::make_unique<UiArtText>(selconn_Description, rect4));
+	vecSelConnDlg.push_back(std::make_unique<UiArtText>(selconn_Description, rect4, UiFlags::FontSize12, 1, 16));
 
 	SDL_Rect rect5 = { (Sint16)(PANEL_LEFT + 30), (Sint16)(UI_OFFSET_Y + 356), 220, 31 };
-	vecSelConnDlg.push_back(std::make_unique<UiArtText>(_("no gateway needed"), rect5, UiFlags::AlignCenter | UiFlags::FontMedium));
+	vecSelConnDlg.push_back(std::make_unique<UiArtText>(_("no gateway needed"), rect5, UiFlags::AlignCenter | UiFlags::FontSize24));
```
StephenCWills, line 59: This text is just a little bit too wide so the last character ends up on its own line. Also, I'm not sure why this text is necessary in the first place.

StephenCWills (changes requested):
FYI, the first one is just a question I had, but the rest of these require some action.

Source/panels/charpanel.cpp
```diff
@@ -235,87 +204,85 @@ void DrawShadowString(const Surface &out, const PanelEntry &entry)
...
 void DrawChr(const Surface &out)
 {
-	if (!CharPanelLoaded) {
-		LoadCharPanel();
-		CharPanelLoaded = true;
-	}
+	//if (!CharPanelLoaded) {
+	LoadCharPanel();
+	CharPanelLoaded = true;
+	//}
```
StephenCWills, line 274:
```suggestion
	LoadCharPanel();
```

Source/panels/charpanel.cpp
```diff
@@ -235,87 +204,85 @@ void DrawShadowString(const Surface &out, const PanelEntry &entry)
...
 		}
 		DrawShadowString(out, entry);
 	}
 
-	for (auto &gfx : PanelParts) {
-		gfx.Unload();
-	}
+	PanelBoxLeft.Unload();
+	PanelBoxMiddel.Unload();
+	PanelBoxRight.Unload();
```
StephenCWills, line 250:
```suggestion
	PanelBoxRight.Unload();

	CharPanelLoaded = true;
```

StephenCWills (changes requested):
I messed these up a moment ago so here's another review.

Source/engine/render/text_render.cpp
```diff
@@ -206,8 +102,8 @@ int GetLineWidth(string_view text, GameFontTables size, int spacing, int *charac
 		if (text[i] == '\n')
 			break;
 
-		uint8_t frame = FontFrame[size][FontIndex[static_cast<uint8_t>(text[i])]];
-		lineWidth += FontKern[size][frame] + spacing;
+		uint8_t frame = text[i] & 0xFF;
```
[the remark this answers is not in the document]
AJenbo: True, this was done early on when I expected to implement Unicode support in the same go.

Source/scrollrt.cpp
```diff
@@ -1206,21 +1206,17 @@ void DrawView(const Surface &out, Point startPosition)
...
 				char coordstr[10];
 				sprintf(coordstr, "%d:%d", dunCoords.x, dunCoords.y);
 				int textWidth = GetLineWidth(coordstr);
 				int textHeight = 12;
-				Point position = center + Displacement { -textWidth / 2, textHeight / 2 };
-				DrawString(out, coordstr, { position, { textWidth, textHeight } }, UiFlags::ColorRed);
+				Size tileSize = { TILE_WIDTH, TILE_HEIGHT };
+				if (!zoomflag)
+					tileSize *= 2;
+				DrawString(out, coordstr, { pixelCoords - Displacement { 0, tileSize.height }, tileSize }, UiFlags::ColorRed | UiFlags::AlignCenter | UiFlags::VerticalCenter);
```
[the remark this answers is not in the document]
AJenbo: Oh yeah getting rid of `textWidth` has why i started to port this to centering :D

Source/DiabloUI/selconn.cpp
AJenbo: It was for selecting a bnet server originally, we could instead use it to let the player set port, bind ip or ZeroTier network.

Source/panels/charpanel.cpp
AJenbo: This will be relevant again once i get TRN working, I have it coded but it's crashing...

StephenCWills: I'll await your further changes then.

Source/DiabloUI/selconn.cpp
StephenCWills: I just meant that I wasn't sure why we need it when we don't have a bnet server with separate gateways, but I guess you're saying that we're keeping it with thoughts of potentially repurposing it. I like those ideas you have for repurposing.

Source/engine/render/text_render.cpp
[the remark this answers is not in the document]
StephenCWills: If you think we'll need it when adding Unicode support, then perhaps we can leave it alone. Maybe keep it as a reminder since it's not hurting anything.

Source/engine/render/text_render.cpp
```diff
@@ -309,16 +234,32 @@ uint16_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
...
-		int symbolWidth = FontKern[size][frame];
+		uint8_t frame = text[i] & 0xFF;
+		int symbolWidth = FontKerns[size][frame];
 		if (text[i] == '\n' || characterPosition.x + symbolWidth > rightMargin) {
 			if (characterPosition.y + lineHeight >= bottomMargin)
 				break;
 			characterPosition.y += lineHeight;
 
-			if (HasAnyOf(flags, (UiFlags::AlignCenter | UiFlags::AlignRight | UiFlags::KerningFitSpacing)))
+			if (HasAnyOf(flags, (UiFlags::AlignCenter | UiFlags::AlignRight | UiFlags::KerningFitSpacing)) && i < text.length())
```
StephenCWills, line 262:
```suggestion
			if (HasAnyOf(flags, (UiFlags::AlignCenter | UiFlags::AlignRight | UiFlags::KerningFitSpacing)) && i + 1 < text.length())
```

Source/engine/render/text_render.hpp
```diff
@@ -56,15 +68,14 @@ void WordWrapGameString(char *text, size_t width, GameFontTables size = GameFont
...
  * @return The number of characters rendered, including characters "drawn" outside the buffer.
  */
-uint16_t DrawString(const Surface &out, string_view text, const Rectangle &rect, UiFlags flags = UiFlags::None, int spacing = 1, int lineHeight = -1, bool drawTextCursor = false);
+uint32_t DrawString(const Surface &out, string_view text, const Rectangle &rect, UiFlags flags = UiFlags::None, int spacing = 1, int lineHeight = -1);
 
 /**
  * @brief Draws a line of text at the given position relative to the origin of the output buffer.
  *
- * This method is provided as a convenience to pass through to DrawString(..., Rectangle, ...) when no explicit
+ * This method is provided as a convenience to pass through to lineHeight(..., Rectangle, ...) when no explicit
```
StephenCWills, line 78:
```suggestion
 * This method is provided as a convenience to pass through to DrawString(..., Rectangle, ...) when no explicit
```

Source/engine/render/text_render.hpp
```diff
@@ -76,14 +87,14 @@ uint16_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
  * @param spacing Additional space to add between characters.
  *                This value may be adjusted if the flag UIS_FIT_SPACING is passed in the flags parameter.
  * @param lineHeight Allows overriding the default line height, useful for multi-line strings.
- * @param drawTextCursor Whether to draw an animated cursor sprite at the end of the text (default is to display nothing).
  * @return The number of characters rendered (could be less than the string length if it wrapped past the bottom of the buffer).
  */
-inline uint16_t DrawString(const Surface &out, string_view text, const Point &position, UiFlags flags = UiFlags::None, int spacing = 1, int lineHeight = -1, bool drawTextCursor = false)
+inline void DrawString(const Surface &out, string_view text, const Point &position, UiFlags flags = UiFlags::None, int spacing = 1, int lineHeight = -1)
```
StephenCWills, line 92: Since the return value is now `void`, the `@return` comment above is no longer necessary.

StephenCWills (changes requested):
You should already be aware from our discussion on Discord, but in addition to these comments, it seems the fonts are being unloaded on level transitions.

Source/engine/render/text_render.cpp
StephenCWills: This fixes the OOB error, but is the logic correct? I understand skipping past the character to calculate `lineWidth` in the case of encountering a newline. However, when a character's width exceeds the right margin, that character gets moved down to the next line so shouldn't we then include that character in the `lineWidth` calculation?

StephenCWills: Maybe something like this.
```suggestion
			if (HasAnyOf(flags, (UiFlags::AlignCenter | UiFlags::AlignRight | UiFlags::KerningFitSpacing))) {
				int nextLineIndex = text[i] == '\n' ? i + 1 : i;
				if (nextLineIndex < text.length())
					lineWidth = GetLineWidth(&text[nextLineIndex], size, spacing, &charactersInLine);
				else
					lineWidth = 0;
			}
```

AJenbo: Good catch, I think I'll actually just simplify the logic here instead. The original problem with rendering past the edge of the box is largely alleviated by the back-buffer now being clipped so that content doesn't wrap the screen.

Source/monster.cpp
```diff
@@ -1372,6 +1372,8 @@ bool MonsterWalk(int i, MonsterMode variant)
 			dFlags[monster.position.temp.x][monster.position.temp.y] &= ~BFLAG_MONSTLR;
 			dMonster[monster.position.tile.x][monster.position.tile.y] = i + 1;
 			break;
+		default:
+			break;
```
StephenCWills, line 1376: What's the purpose of this?
```suggestion
```

StephenCWills (dismissed):

Source/engine/render/text_render.cpp
StephenCWills: With the latest change, the OOB can only occur if a center-aligned, right-aligned, or kerning-adjusted string ends in a newline. In the first place, there aren't any strings in the game that end in newlines so for now it should be okay.

Source/monster.cpp
AJenbo: Exhaustive enum case handeling, like everywhere else so that we don't get a long ass compiler warning for every enum value not handled (monster mode does not just contain walk directions)

StephenCWills (approved):

altiereslima:
I have an error.
Sem título

AJenbo:
> I have an error.

Please let us know what translation this is from?

altiereslima:
> > Eu tenho um erro.
> 
> Por favor, deixe-nos saber de que tradução é esta?

Portuguese translation

julealgon:
Why would the exact same string have an issue in the info panel but not in the health bar? Are the text rendering functions different somehow?

AJenbo:
> Portuguese translation

Thanks, turns out the name was being double translated in the info panel.

PR #2796, opened 2021-09-05
Clean up some references to global arrays

ephphatha:
Doing this as prep work to try implement a container suitable for the way Missiles/Objects/Items/Monsters are used, so the current aim is to make uses of those arrays more obviously similar.


Source/monster.cpp
```diff
@@ -4399,8 +4399,6 @@ void FreeMonsters()
 			}
 		}
 	}
-
-	FreeMissiles2();
```
AJenbo, line 4403: This will cause the game not to unload missiles that where only used on a specific floor (monster projectiles). Please restore it in some form :)

Source/misdat.cpp
```diff
@@ -248,4 +248,22 @@ void MissileFileData::LoadGFX()
 	}
 }
 
+void InitMissileGFX(bool loadHellfireGraphics)
+{
+	for (size_t mi = 0; MissileSpriteData[mi].animFAmt != 0; mi++) {
```
julealgon, line 253: While technically correct, don't you think using `size_t` here is a bit overkill? This is the type of code that I think damages readability.

Would just use `int` here as the previous code did.

Source/misdat.cpp
```diff
@@ -248,4 +248,22 @@ void MissileFileData::LoadGFX()
 	}
 }
 
+void InitMissileGFX(bool loadHellfireGraphics)
```
julealgon, line 251: I'd love to not see references to "hellfire" explicitly in the code. The fact that we are introducing yet another one here bothers me a bit.

- Would it be possible to rename this to something more generic that doesn't reference hellfire explicitly? 
- Wouldn't it be better to have 2 loading methods instead of having one with a parameter?
- Why is `MissileSpriteData[]` populated with missiles that may end up not being loadable in the first place? Can't we just iterate through the entire array here, but prevent it from having hellfire missiles added to it somewhere else?

Source/misdat.cpp
```diff
@@ -248,4 +248,22 @@ void MissileFileData::LoadGFX()
...
+	for (size_t mi = 0; MissileSpriteData[mi].animFAmt != 0; mi++) {
+		if (!loadHellfireGraphics && mi > MFILE_SCBSEXPD)
+			break;
+		if (MissileSpriteData[mi].flags == MissileDataFlags::MonsterOwned)
+			continue;
+		MissileSpriteData[mi].LoadGFX();
+	}
+}
+
+void FreeMissileGFX()
```
julealgon, line 262: The old version of the 2 variations actually had specific behaviors built-in:
- `FreeMissiles` -> `!= MissileDataFlags::MonsterOwned`
- `FreeMissiles2` ->`== MissileDataFlags::MonsterOwned` 

We seem to be combining both by just freeing all missiles here unconditionally.

Are you sure this has equivalent behavior as before?

Source/monster.cpp
julealgon: All missiles will be unloaded here alongside every other missile:
- https://github.com/diasurgical/devilutionX/pull/2796/commits/7f72da754005c826b767cf0a99eac607c620fe4d#diff-84e63ba430a13546a287665b7640118b2f9d66fb9586056bfbab8215bab319c0R1524

Source/loadsave.cpp
```diff
@@ -682,43 +682,41 @@ void LoadMissile(LoadHelper *file, Missile &missile)
...
-	pObject->_olid = file->NextLE<int32_t>();
-	pObject->_oRndSeed = file->NextLE<uint32_t>();
-	pObject->_oVar1 = file->NextLE<int32_t>();
-	pObject->_oVar2 = file->NextLE<int32_t>();
-	pObject->_oVar3 = file->NextLE<int32_t>();
-	pObject->_oVar4 = file->NextLE<int32_t>();
-	pObject->_oVar5 = file->NextLE<int32_t>();
-	pObject->_oVar6 = file->NextLE<uint32_t>();
-	pObject->bookMessage = static_cast<_speech_id>(file->NextLE<int32_t>());
-	pObject->_oVar8 = file->NextLE<int32_t>();
```
julealgon, line 721: The diff from github is absolutely terrible here. I used VS to compare:
image

Source/loadsave.cpp
```diff
@@ -1328,43 +1326,41 @@ void SaveMissile(SaveHelper *file, Missile &missile)
...
-	file->WriteLE<int32_t>(pObject->_olid);
-	file->WriteLE<uint32_t>(pObject->_oRndSeed);
-	file->WriteLE<int32_t>(pObject->_oVar1);
-	file->WriteLE<int32_t>(pObject->_oVar2);
-	file->WriteLE<int32_t>(pObject->_oVar3);
-	file->WriteLE<int32_t>(pObject->_oVar4);
-	file->WriteLE<int32_t>(pObject->_oVar5);
-	file->WriteLE<uint32_t>(pObject->_oVar6);
-	file->WriteLE<int32_t>(pObject->bookMessage);
-	file->WriteLE<int32_t>(pObject->_oVar8);
```
julealgon, line 1367: image

Source/loadsave.cpp
```diff
@@ -207,313 +207,309 @@ class SaveHelper {
...
+	player._pDFrames = file.NextLE<int32_t>();
+	file.Skip(4);     // skip _pDWidth
+	file.Skip(4 * 8); // Skip pointers _pBAnim
+	player._pBFrames = file.NextLE<int32_t>();
+	file.Skip(4); // skip _pBWidth
+
+	for (Item &item : player.InvBody)
+		LoadItemData(file, item);
+	for (Item &item : player.InvList)
+		LoadItemData(file, item);
```
julealgon, line 457: Consider leaving an empty line after these scopeless statements for readability.

Source/loadsave.cpp
```diff
@@ -1623,9 +1619,9 @@ void LoadHeroItems(Player &player)
 
 	gbIsHellfireSaveGame = file.NextBool8();
 
-	LoadMatchingItems(&file, NUM_INVLOC, player.InvBody);
-	LoadMatchingItems(&file, NUM_INV_GRID_ELEM, player.InvList);
-	LoadMatchingItems(&file, MAXBELTITEMS, player.SpdList);
+	LoadMatchingItems(file, NUM_INVLOC, player.InvBody);
+	LoadMatchingItems(file, NUM_INV_GRID_ELEM, player.InvList);
+	LoadMatchingItems(file, MAXBELTITEMS, player.SpdList);
```
julealgon, line 1615: You removed the iteration constants related to `LoadItems` above, but we are doing something very similar here.

Can't we leverage some of the iterator work created recently (to iterate on all "valid" player items) for this as well, by creating a variation of that that iterates on _all_ items?

That would combine the iteration here, and on the other loops, as well as remove all of these constants from the equation.

Source/loadsave.cpp
```diff
@@ -1852,9 +1843,12 @@ void SaveHeroItems(Player &player)
...
 
-	SaveItems(&file, player.InvBody, NUM_INVLOC);
-	SaveItems(&file, player.InvList, NUM_INV_GRID_ELEM);
-	SaveItems(&file, player.SpdList, MAXBELTITEMS);
+	for (const Item &item : player.InvBody)
+		SaveItem(file, item);
+	for (const Item &item : player.InvList)
+		SaveItem(file, item);
+	for (const Item &item : player.SpdList)
+		SaveItem(file, item);
```
julealgon, line 1851: Same on these loops. Would be nice if we could leverage an aggregate iterator.

Source/monster.cpp
AJenbo: That's game shutdown, not level transition.

Source/loadsave.cpp
AJenbo: I use Diffuse

Source/loadsave.cpp
AJenbo: From the description this looks to be prep for doing just that.

Source/misdat.cpp
AJenbo: It isn't, see my comment on this: https://github.com/diasurgical/devilutionX/pull/2796#discussion_r702488149

Source/monster.cpp
julealgon: The place I pointed to is currently the only place in the codebase where `FreeMonsters` is called.

I think you are mistaken here.

Source/misdat.cpp
julealgon: Care to elaborate on your downvote @AJenbo ? 

In particular regarding the third bullet above: why do we have entries in `MissileSpriteData` that cannot be loaded? To me, the array should only be populated with entries that can effectively be loaded, and thus the hellfire check should be happening there, and not here.

Source/monster.cpp
ephphatha: `FreeMissiles2()` had exactly one use in `FreeMonsters()`, which itself has exactly one use in `FreeGameMem()`. `FreeGameMem()` gets called on all forms of level transition, starting a new game, quitting diablo, and loading a game.

From what I could tell nothing in `FreeMonsters()` requires the non-monster missile graphics to be freed first so it's safe to unload all missile graphics in one pass.

Source/loadsave.cpp
ephphatha: Yeah, once I hit 500 changed lines I figured it was worth getting feedback on. These loops are more involved than the simple cases, and because `InvBody` et al are C-style arrays instead of STL containers using them in a function requires passing the size information as a separate parameter. I could change `LoadMatchingItems` to operate on a Player reference but I figured it would be better to leave it till we can use containers for the collections instead.

Source/misdat.cpp
ephphatha: Happy to change it back given the comparison with a `missile_graphic_id` member. The main reason I wanted to use `size_t` here was so I can see from that line that `mi` is used as an index and not stored/saved like what happens with `ActiveMissiles` and the other common use of `mi`.

Source/misdat.cpp
ephphatha: This param was introduced to extract the use of the `gbIsHellfire` global. misdat.cpp does not know of that symbol and I did not want to add another reference to a global that is not strictly necessary.

`MissileSpriteData[]` is initialised when declared instead of being dynamic, changing that is possible but a significant amount of work I really didn't want to get into here 😂

Source/misdat.cpp
ephphatha: The old code path was as follows:

```cpp
for (auto &missileData : MissileSpriteData) {
    if (missileData.flag != MonsterOwned)
        missileData.FreeGFX();
}

//Do the FreeMonsters stuff

for (auto &missileData : MissileSpriteData) {
    if (missileData.flag == MonsterOwned)
        missileData.FreeGFX();
}
```

`FreeMissile()` was always called alongside `FreeMonsters()` which then called `FreeMissile2()`

Source/misdat.cpp
AJenbo: I'm fine either way.

AJenbo (approved):

AJenbo (approved):
LGTM, but will wait with merging to see if @julealgon has anything to add.

Source/monster.cpp
AJenbo: Hmm maybe it would be worth fixing this as some point, it sound like this will mean that more data gets loaded if you play multiple levels in one sitting then if you restart between each level.

PR #2797, opened 2021-09-05
SDL stuff: Automatic memory management and error checking, part 2 of N

vlolteanu:


Source/dx.cpp
```diff
@@ -200,14 +201,14 @@ void dx_cleanup()
 	sgdwLockCount = 0;
 	MemCrit.unlock();
 
-	if (pal_surface == nullptr)
+	if (PalSurface == nullptr)
 		return;
```
vlolteanu, line 205: Is it ever the case that `PalSurface == nullptr`?

AJenbo: Yes if `SDL_GetWindowSurface` or `SDL_CreateRGBSurfaceWithFormat` fails, since we do not appear to check this.

vlolteanu: It doesn't seem like a good idea to cut `dx_cleanup` short for this reason.

I added error checking for `SDL_CreateRGBSurfaceWithFormat`; I'll do the same for `SDL_GetWindowSurface`.

AJenbo:
@vlolteanu would you be willing to rebase this?

vlolteanu:
Done.

Source/utils/sdl_wrap.h
```diff
@@ -14,22 +14,55 @@ namespace devilution {
...
 inline SDLSurfaceUniquePtr CreateRGBSurfaceWithFormat(Uint32 flags, int width, int height, int depth, Uint32 format)
 {
-	SDLSurfaceUniquePtr ret { SDL_CreateRGBSurfaceWithFormat(flags, width, height, depth, format) };
-	if (ret == nullptr)
-		ErrSdl();
+	return SDLSurfaceUniquePtr { NonNull(SDL_CreateRGBSurfaceWithFormat(flags, width, height, depth, format)) };
+}
 
-	return ret;
+inline SDLSurfaceUniquePtr CreateRGBSurfaceWithFormatFrom(void *pixels, int width, int height, int depth, int pitch, Uint32 format)
```
AJenbo, line 35: Would it be possible to do something similar for the `goto error` in `codec_decode` and `OpenMPQ`?

Source/dx.cpp
```diff
@@ -238,17 +236,14 @@ void dx_reinit()
 
 void InitPalette()
 {
-	Palette = SDL_AllocPalette(256);
-	if (Palette == nullptr) {
-		ErrSdl();
-	}
+	Palette = SDLWrap::AllocPalette(256);
```
AJenbo, line 239: I'm fine with us assuming that palettes contains 256 colors to maybe simplify this.

Source/dx.cpp
```diff
@@ -200,14 +200,12 @@ void dx_cleanup()
...
-	if (pal_surface == nullptr)
-		return;
-	SDL_FreeSurface(pal_surface);
-	pal_surface = nullptr;
-	SDL_FreePalette(Palette);
-	SDL_FreeSurface(renderer_texture_surface);
+	PalSurface = nullptr;
+	PinnedPalSurface = nullptr;
+	Palette = nullptr;
+	RendererTextureSurface = nullptr;
```
AJenbo, line 206: so much nicer :+1:

Source/utils/sdl_wrap.h
vlolteanu: I think your comment was meant for a different PR.

Source/dx.cpp
vlolteanu: Thanks!

Source/utils/sdl_wrap.h
AJenbo: No, I'm just asking if you think we could implement a unique pointer to avoid the `goto error` related code in those two function :)

vlolteanu: I took a look and the short answer is yes:

- For the SHA stuff, the clean solution is to refactor the code and get rid of sgSHA1.
- For the MPQ stuff, I've got a few more commits incoming.

AJenbo: Nice, in that case this has my vote for the next thing to tackle in this series :)

AJenbo (approved):

AJenbo (approved):

PR #2801, opened 2021-09-06
♻️ Standardize 'Direction' to 'Displacement' conversions

julealgon:
This PR changes how we usually convert from a `Direction` instance to a `Displacement` by introducing an explicit conversion constructor that is used in place of the old `fromDirection` static method. This leads to simpler and more standardized code since we currently use this approach for other types like `Size` as well.

The old method is kept to simplify the construction logic (leveraging the automatic copy constructor) but is made `private` to encourage the use of the constructor instead.


ephphatha (approved):
I'd been holding off adding this constructor until `Direction` was made a scoped enum, however there's no `Displacement(int)` constructor so there'll be no ambiguity. Looks good to me.

PR #2805, opened 2021-09-06
🚚 Rename 'TextDataStruct' struct to 'TextData'

julealgon:


AJenbo (approved):

Source/textdat.cpp
```diff
@@ -11,7 +11,7 @@ namespace devilution {
 /* todo: move text out of struct */
 
 /** Contains the data related to each speech ID. */
-const TextDataStruct Texts[] = {
+const TextData Texts[] = {
```
AJenbo, line 14: Maybe this would be better named Dialog, Gosip or Speach since test is just one of it's properties.

julealgon: Went with 'Speech' since that is more generic. 'Dialog' could be misleading since some of it comes from the narrator, and that is not really a dialog. Gossip would also limit it too much since not every interaction is considered gossip in the game.

AJenbo (approved):

PR #2810, opened 2021-09-06
[Android] Fix input not being in view for 5.1 - 7.0

AJenbo:
This is still broken in Android 7.1.1+


StephenCWills (approved):
I can confirm this doesn't change anything on my Pixel 4.

PR #2811, opened 2021-09-06
:bug: Fix teleport not working in all locations

AJenbo:
- Player cannot TP to current location
- Cannot teleport to any object location (including doors)
- Can teleport to void locations
- Cannot teleport to dead monsters

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1898,29 +1898,23 @@ void AddMagmaball(Missile &missile, Point dst, Direction /*midir*/)
 
 void AddTeleport(Missile &missile, Point dst, Direction /*midir*/)
 {
-	missile._miDelFlag = true;
-
 	for (int i = 0; i < 6; i++) {
 		int k = CrawlNum[i];
 		int ck = k + 2;
 		for (auto j = static_cast<uint8_t>(CrawlTable[k]); j > 0; j--, ck += 2) {
 			Point target = dst + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
-			if (!InDungeonBounds(target) || IsTileSolid(target))
+			if (!PosOkPlayer(Players[missile._misource], target))
 				continue;
-			if (dMonster[target.x][target.y] == 0 && dObject[target.x][target.y] == 0 && dPlayer[target.x][target.y] == 0) {
-				missile.position.tile = target;
-				missile.position.start = target;
-				missile._miDelFlag = false;
-				i = 6;
-				break;
-			}
+
+			missile.position.tile = target;
+			missile.position.start = target;
+			UseMana(missile._misource, SPL_TELEPORT);
+			missile._mirange = 2;
+			return;
 		}
 	}
 
-	if (!missile._miDelFlag) {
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -1907,7 +1907,7 @@ void AddTeleport(Missile &missile, Point dst, Direction /*midir*/)
 			Point target = dst + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
 			if (!InDungeonBounds(target) || IsTileSolid(target))
```
julealgon, line 1908: Aren't these conditions also covered by `PosOkPlayer`?

qndel: no clue, the function that had it was in anonymous namespace so I copy/pasted it

julealgon:
I see you added a lot more changes here. I'd probably have to bombard you with questions to understand some of them since I've not touched this part of the codebase much if at all... so I'll pass reviewing this one.

AJenbo:
> I see you added a lot more changes here. I'd probably have to bombard you with questions to understand some of them since I've not touched this part of the codebase much if at all... so I'll pass reviewing this one.

I'll split it, but do feel free to ask away.

Source/missiles.cpp
```diff
@@ -1898,29 +1898,23 @@ void AddMagmaball(Missile &missile, Point dst, Direction /*midir*/)
...
 		for (auto j = static_cast<uint8_t>(CrawlTable[k]); j > 0; j--, ck += 2) {
 			Point target = dst + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
-			if (!InDungeonBounds(target) || IsTileSolid(target))
+			if (!PosOkPlayer(Players[missile._misource], target))
 				continue;
-			if (dMonster[target.x][target.y] == 0 && dObject[target.x][target.y] == 0 && dPlayer[target.x][target.y] == 0) {
-				missile.position.tile = target;
-				missile.position.start = target;
-				missile._miDelFlag = false;
-				i = 6;
```
julealgon, line 1914: OMG this assignment is insane..... am I understanding this right? Was this a hacky way of "breaking out of the external loop as well"?

Source/missiles.cpp
```diff
@@ -1898,29 +1898,23 @@ void AddMagmaball(Missile &missile, Point dst, Direction /*midir*/)
...
-				break;
-			}
+
+			missile.position.tile = target;
+			missile.position.start = target;
+			UseMana(missile._misource, SPL_TELEPORT);
+			missile._mirange = 2;
```
julealgon, line 1912: Are these 2 order dependent? If not, I'd suggest swapping them so all the assignments happen before the call to reduce cognitive load.

julealgon (approved):

Source/missiles.cpp
AJenbo: They are not, but looks like  you already approved so it was auto merged :upside_down_face:

Source/missiles.cpp
AJenbo: yep, it's confused me like 50 times by now.

PR #2816, opened 2021-09-06
:bug: Fix search spell affecting all players

AJenbo:


julealgon:
We have an issue to track this, right? Can you link it to the PR?

Source/missiles.cpp
```diff
@@ -1689,7 +1689,8 @@ void AddRing(Missile &missile, Point /*dst*/, Direction /*midir*/)
 
 void AddSearch(Missile &missile, Point /*dst*/, Direction /*midir*/)
 {
-	AutoMapShowItems = true;
+	if (missile._misource == MyPlayerId)
+		AutoMapShowItems = true;
```
julealgon, line 1693: Consider leaving a blank line after this for readability.

julealgon (approved):

Source/missiles.cpp
AJenbo: Good point to late

PR #2817, opened 2021-09-06
Clean up missile code

AJenbo:


Source/missiles.cpp
```diff
@@ -1616,10 +1616,6 @@ void AddFireNova(Missile &missile, Point dst, Direction midir)
 	UpdateMissileVelocity(missile, dst, sp);
 	SetMissDir(missile, GetDirection16(missile.position.start, dst));
 	missile._mirange = 256;
-	missile.var1 = missile.position.start.x;
-	missile.var2 = missile.position.start.y;
-	missile.var4 = missile.position.start.x;
-	missile.var5 = missile.position.start.y;
```
julealgon, line 1622: Can you elaborate why we can remove this? I don't see it being compensated anywhere else in the PR.

Source/missiles.cpp
```diff
@@ -1664,7 +1660,6 @@ void AddMana(Missile &missile, Point /*dst*/, Direction /*midir*/)
 		player._pManaBase = player._pMaxManaBase;
 	UseMana(missile._misource, SPL_MANA);
 	missile._miDelFlag = true;
-	drawmanaflag = true;
```
julealgon, line 1667: I assume this is being handled by `UseMana` already and is why we can remove it from here?

Source/missiles.cpp
```diff
@@ -1928,13 +1923,9 @@ void AddLightball(Missile &missile, Point dst, Direction /*midir*/)
...
 	missile._miAnimFrame = GenerateRnd(8) + 1;
 	missile._mirange = 255;
-	if (missile._misource < 0) {
-		missile.var1 = missile.position.start.x;
-		missile.var2 = missile.position.start.y;
-	} else {
-		missile.var1 = Players[missile._misource].position.tile.x;
-		missile.var2 = Players[missile._misource].position.tile.y;
-	}
+	Point position = missile._misource < 0 ? missile.position.start : Players[missile._misource].position.tile;
```
julealgon, line 1926: Consider marking `const` and using initializer list:
```suggestion
	const Point position { missile._misource < 0 ? missile.position.start : Players[missile._misource].position.tile };
```

Source/missiles.cpp
```diff
@@ -1980,8 +1971,6 @@ void AddFireball(Missile &missile, Point dst, Direction midir)
 	missile._mirange = 256;
 	missile.var1 = missile.position.start.x;
 	missile.var2 = missile.position.start.y;
-	missile.var4 = missile.position.start.x;
-	missile.var5 = missile.position.start.y;
```
julealgon, line 1984: Similar to the other one, can you elaborate why this is a safe removal?

Source/missiles.cpp
AJenbo: Never used by `MI_FireNova`

Source/missiles.cpp
AJenbo: The spell it self doesn't affect your mana.

Source/missiles.cpp
AJenbo: Never used by `MI_Fireball`

Source/missiles.cpp
obligaron: I'm not sure. Doesn't it change mana?
I tested with `Staff of Mana` and it will not redraw the mana flask correctly.
But perhaps I miss something.

AJenbo: Your right, i was thinking this was AddManashield

PR #2819, opened 2021-09-07
Standardize 'ItemType' enum

julealgon:
This PR cleans up all usages of `ItemType` and converts it into a proper enum class.


Source/objects.cpp
```diff
@@ -3974,26 +3974,11 @@ bool OperateFountains(int pnum, int i)
...
-		weaponType = ITYPE_AXE;
-		break;
-	case ITYPE_BOW:
-		weaponType = ITYPE_BOW;
-		break;
-	case ITYPE_MACE:
-		weaponType = ITYPE_MACE;
-		break;
-	}
+	ItemType weaponType { PickRandomlyAmong({ ItemType::Sword, ItemType::Axe, ItemType::Bow, ItemType::Mace }) };
```
ephphatha, line 3981: This is a safe use of `PickRandomlyAmong` only because the list is a power of 2 in length.

test/inv_test.cpp
```diff
@@ -12,7 +12,7 @@ void set_up_scroll(Item &item, spell_id spell)
 	pcurs = CURSOR_HAND;
 	leveltype = DTYPE_CATACOMBS;
 	Players[MyPlayerId]._pRSpell = static_cast<spell_id>(spell);
-	item._itype = ITYPE_MISC;
+	item._itype = Misc;
```
ephphatha, line 15:
```suggestion
	item._itype = ItemType::Misc;
```

Even if the prefix is optional here it would be best to include it in the test files.

Source/engine/random.hpp
```diff
@@ -57,4 +58,18 @@ int32_t AdvanceRndSeed();
...
+ *
+ * @param values The values to pick from
+ * @return A random value from the 'values' list.
+ */
+template <typename T>
+const T PickRandomlyAmong(const std::initializer_list<T> &values)
+{
+	auto index = GenerateRnd(values.size());
+
+	return *(values.begin() + index);
```
ephphatha, line 72: What's the intent when `GenerateRnd()` returns a negative value? This code will attempt to read past the head of the list. Need to document the exceptional case and either clamp to 0 or recognise callers that may rely on reading out of bounds to match vanilla logic.

Source/objects.cpp
julealgon: Can you elaborate why that's the case? The documentation states it 
> Generates a random integer less than the given limit

Why would it cause a problem if the list had non-power-of-2 items?

test/inv_test.cpp
julealgon: I hate these test........... every refactor operation I do in Visual Studio doesn't include them. I'm just not used to not having the tests be part of normal compilation.

Source/engine/random.hpp
julealgon: Wait... `GenerateRnd` can return a negative value? The hell....... I'll take a look.

ephphatha: It's a rare case but that's what caused the segfault in the quest tests in #2751. See https://github.com/diasurgical/devilutionX/blob/946ae987a31c239499a59525c15df99a904ac842/Source/engine/random.hpp#L44-L58 and the comment on AdvanceRndSeed for more detail.

julealgon: If `GenerateRnd` can create a negative value, the original code would generate a `ItemType::Misc` item on a weapon rack.... is that really a thing? I've never heard of someone getting that before.

Source/objects.cpp
ephphatha: Essentially this is due to the vanilla RNG codes dodgy implementation of Borland `rand()`. It tries to convert the current random engine state to a signed value using `abs()`. This fails when the engine state is UINT_MAX, the call to `abs()` returns INT_MIN (-2^31). `GenerateRnd()` then does modulo division on that value, if the limit/`v` is a power of 2 `-2^31 % v == 0` otherwise you end up with a negative number.

Source/engine/random.hpp
julealgon: Anyways....... clamped the value to 0. Let me know if that's fine.

ephphatha: Yep, that looks good. For weapon racks the old code (and the change you made) was safe because `GenerateRnd(4)` will always return 0, 1, 2, or 3.

If for example you tried to add shields to the potential drops then `GenerateRnd(5)` can return 0, 1, 2, 3, 4, or -2. The -2 value would correspond to ItemType::None which isn't handled by the switch statement so the old code would call CreateTypeItem with an uninitialised `weaponType` variable. The first version of your PickRandomlyAmong code would try dereference `(values.begin() - 2)` which may or may not crash.

ephphatha (dismissed):
Also need to update the ItemType references in pack_test.cpp, however other than that I think it's good to go.

Source/objects.cpp
julealgon: There are a multitude of places in the codebase currently calling `GenerateRnd` with non-power-of-2 values... are _all_ of those bugs?

Because if they are, I don't really see the problem in this usage here.

ephphatha: > are all of those bugs?

Yep 😢 .

I originally commented here to highlight that this was safe because we were lucky with the number of choices. However now that you've updated PickRandomlyAmong to always return an item in the list it is safe to use no matter how many elements are in the list.

Source/engine/random.hpp
julealgon: That's pretty crazy......... I guess at least it _can_ return all of the values in the range, so the clamping works, but biases the result against position 0 in that case?

Wouldn't it be possible to just try generating another number if the value returns negative as an implementation detail on `GenerateRnd`? Or would advancing the seed in the process mess up things?

ephphatha: Clamping to zero also has the benefit of correcting a slight bias that would otherwise be introduced. Definitely do not want to reroll as that would break things that rely on a specific number of RNG calls like item generation.

ephphatha (approved):

test/inv_test.cpp
AJenbo: Then enable the tests... it will compile both the normal build and the tests then.

julealgon:
@AJenbo can we move forward with this one?

AJenbo:
image

AJenbo:
@julealgon a few other PRs where merged and caused conflicts with this one, would you be willing to rebase it?

julealgon:
> @julealgon a few other PRs where merged and caused conflicts with this one, would you be willing to rebase it?

Rebased.

ephphatha (approved):
No functional changes from the rebase that I can see, looks good to go.

The test files use right-aligned "headers" on the test datasets for every column except the first, which is a bit of an oddity.  All text columns are left-aligned, if we wanted to nit could left align the _itype header as well.

Source/engine/random.hpp
AJenbo: Wouldn't clamping it make it biased to 0?

Source/controls/plrctrls.cpp
```diff
@@ -1319,7 +1319,7 @@ void UseBeltItem(int type)
 		if ((type == BLT_HEALING && (id == IMISC_HEAL || id == IMISC_FULLHEAL || (id == IMISC_SCROLL && spellId == SPL_HEAL)))
 		    || (type == BLT_MANA && (id == IMISC_MANA || id == IMISC_FULLMANA))
 		    || id == IMISC_REJUV || id == IMISC_FULLREJUV) {
-			if (myPlayer.SpdList[i]._itype > -1) {
+			if (static_cast<int>(myPlayer.SpdList[i]._itype) > -1) {
```
AJenbo, line 1322:
```suggestion
			if (myPlayer.SpdList[i]._itype != ItemType::None) {
```

or 

```suggestion
			if (!myPlayer.SpdList[i].isEmpty()) {
```

Source/items.cpp
```diff
@@ -2781,31 +2779,31 @@ void CalcPlrItemVals(Player &player, bool loadgfx)
...
-		if (player.InvBody[INVLOC_HAND_LEFT]._itype == ITYPE_SHIELD || player.InvBody[INVLOC_HAND_RIGHT]._itype == ITYPE_SHIELD) {
-			if (player.InvBody[INVLOC_HAND_LEFT]._itype == ITYPE_SHIELD)
+		if (player.InvBody[INVLOC_HAND_LEFT]._itype == ItemType::Shield || player.InvBody[INVLOC_HAND_RIGHT]._itype == ItemType::Shield) {
+			if (player.InvBody[INVLOC_HAND_LEFT]._itype == ItemType::Shield)
 				player._pIAC -= player.InvBody[INVLOC_HAND_LEFT]._iAC / 2;
-			else if (player.InvBody[INVLOC_HAND_RIGHT]._itype == ITYPE_SHIELD)
+			else if (player.InvBody[INVLOC_HAND_RIGHT]._itype == ItemType::Shield)
 				player._pIAC -= player.InvBody[INVLOC_HAND_RIGHT]._iAC / 2;
-		} else if (player.InvBody[INVLOC_HAND_LEFT]._itype != ITYPE_STAFF && player.InvBody[INVLOC_HAND_RIGHT]._itype != ITYPE_STAFF && player.InvBody[INVLOC_HAND_LEFT]._itype != ITYPE_BOW && player.InvBody[INVLOC_HAND_RIGHT]._itype != ITYPE_BOW) {
+		} else if (player.InvBody[INVLOC_HAND_LEFT]._itype != ItemType::Staff && player.InvBody[INVLOC_HAND_RIGHT]._itype != ItemType::Staff && player.InvBody[INVLOC_HAND_LEFT]._itype != ItemType::Bow && player.InvBody[INVLOC_HAND_RIGHT]._itype != ItemType::Bow) {
```
AJenbo, line 2806:
```suggestion
		} else if (IsNoneOf(player.InvBody[INVLOC_HAND_LEFT]._itype, ItemType::Staff, ItemType::Bow) && IsNoneOf(player.InvBody[INVLOC_HAND_RIGHT]._itype, ItemType::Staff, ItemType::Bow)) {
```

AJenbo (changes requested):
Really nice cleanup, there is still once case of using the enum as an int which should be changed and a possible improvement with using `IsNoneOf()`, other then that it's ready to merge.

julealgon:
> The test files use right-aligned "headers" on the test datasets for every column except the first, which is a bit of an oddity. All text columns are left-aligned, if we wanted to nit could left align the _itype header as well.

Agreed. While I feel it is unrelated, fixed anyways.

AJenbo (approved):

PR #2820, opened 2021-09-07
🌐 Update Italian localization

M0Rf30:
As in title,
Regards.

### DIFF
--- a/Translations/devilutionx.pot
+++ b/Translations/devilutionx.pot
@@ -2,11 +2,11 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: DevilutionX\n"
-"POT-Creation-Date: 2021-08-12 11:42+0200\n"
+"POT-Creation-Date: 2021-09-07 10:25+0200\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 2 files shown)


AJenbo:
Thanks, I'll let you know when the renaming elements are translatable :)

AJenbo:
@M0Rf30 all planned features affecting the Italian translation should now be done and all translated text should show correctly in game. We plan to release in October so now would be an excellent time to work on the renaming parts of the translation :)

image

M0Rf30:
@AJenbo Amazing work. I'll refine the localization just in time for the next release, but I'm having an annoying issue with libsodium detection that breaks the software building process. I've seen that a related similar issue was previously fixed. At a first glance, reverting this [commit](https://github.com/diasurgical/devilutionX/commit/1f7f5bf3dccee957e23bdf4d976883b155d5e81a) does the magic. I'll keep you up to date

AJenbo:
Ok, You will need to let us know what system you are on for us to help with that :)

Don't forget that you can also just grab a test build from the CI

PR #2825, opened 2021-09-07
IsAnyOf/IsNoneOf enhancements

julealgon:
This PR proposes a variation of our `IsAnyOf` and `IsNoneOf` methods that leads to more readable and standard code by relying on the comparison operators to work instead of having to pass the source of the comparison as an argument.


AJenbo:
linux_x86_64 is running GNU 6.3.0, not sure if that's why it's failing but it might be.

julealgon:
@AJenbo is this what is happening here?

https://stackoverflow.com/a/57803674/1946412

If so, how do we fix it? Do I have to add one of those "polyfills" like you mentioned to cover `std::apply` as well?

AJenbo:
This is a bit advanced for me, but you can see if it works: https://github.com/diasurgical/devilutionX/blob/4a7ba9470287f803765bda20c2e29a94f89b3a8f/Source/utils/stdcompat/string_view.hpp

It does look like it will work even in GCC 5.5: https://godbolt.org/z/xIzbD8

ephphatha:
You could also use std::any_of and std::none_of (https://en.cppreference.com/w/cpp/algorithm/all_any_none_of) as they've been part of the STL since C++11. Worse syntax to use but it'll be hidden by the helper anyway.

qndel:
Personally I hate the `!= AllOf`, it makes no sense, != AnyOf would make some :D

qndel (changes requested):
Imo it'd be best if there was no != AllOf but != AnyOf, if it's not any, it 's none :P
Other than that, definitely an improvement 👍

Source/engine.h
```diff
@@ -37,35 +37,67 @@
...
+		    [&](auto... vals) { return ((lhs == vals) || ...); },
+		    rhs.values);
+	}
+
+	template <typename T>
+	[[nodiscard]] friend constexpr bool operator!=(T lhs, AnyOfHelper const &rhs) noexcept
+	{
+		return apply(
+		    [&](auto... vals) { return ((lhs != vals) && ...); },
+		    rhs.values);
```
ephphatha, line 72: This might as well be an inversion of operator==
```suggestion
		return !(lhs == rhs);
```

Source/engine.h
```diff
@@ -37,35 +37,67 @@
...
-}
+	constexpr AnyOfHelper(Args... values)
+	    : values(std::move(values)...)
+	{
+	}
 
-template <typename V, typename X, typename... Xs>
-bool IsNoneOf(const V &v, X x, Xs... xs)
+	template <typename T>
+	[[nodiscard]] friend constexpr bool operator==(T lhs, AnyOfHelper const &rhs) noexcept
```
ephphatha, line 60: nit: convention has been to use `const T &` instead of `T const &` so far

AJenbo:
I admire you persistence in trying to solve minor style issues like this :D

ephphatha:
GCC 6 has an error in their own header? This is causing the failing build on linux_x86_64:
```
/usr/include/c++/6/experimental/tuple:52:37: error: 'tuple_size' was not declared in this scope
     constexpr size_t tuple_size_v = tuple_size<_Tp>::value;
```

glebm:
Personally, this seems less readable to me, as `operator==` is being abused as a belongs-to operator.

julealgon:
> I admire you persistence in trying to solve minor style issues like this :D

I have to say I'm nearly giving up on this. It is extremely frustrating to try and have the logic run on that target since I can't properly detect things (looks like feature-test macros are incredibly unreliable unless you are on C++20 already) and when I do detect it, there are weird random errors inside files I don't own.

I'll try a couple more times here but this is really unproductive use of time it turns out.

> Personally, this seems less readable to me, as operator== is being abused as a belongs-to operator.

I don't disagree that this is abuse of the operator, but you are using that to justify it being less readable? That doesn't make much sense, since that is an implementation detail.

The new format is _far_ more intuitive to me, as it actually approaches C# pattern matching (which IMHO is quite readable):
```csharp
if (x is 10 or 20 or 30)
    DoSomething();
```

Current syntax is very misleading, since the method name implies the values are what you are checking against, but it is a static method, so it barely makes sense. If C++ had the concept of extension methods, this would be a better alternative for sure (even though I still personally prefer the `==` approach):
```cpp
if (x.IsAnyOf(10, 20, 30))
    DoSomething();
```

That would be a lot more semantic since you are saying "x is any of (a, b, c)" and not "is any of (x, a, b, c)" which is incredibly unintuitive to read and understand.

Anyways, it is looking like this won't make it.

julealgon:
> This might explain the difference of opinion :)

Do you agree with @glebm then, that the new syntax is less readable? Or did you read until you found "C#" and dismissed it?

StephenCWills:
I like how the proposed syntax differentiates the variable being matched from the collection of values it's matched against. Also, I think it reads just fine as "if thing is equal to any of these". I understand the point about abusing the `==` operator, but I feel it would be a lot more compelling if there were a high likelihood that the syntax might subvert the developer's expectations, and I don't see how it could. Overall, I'd say it's a positive change, but I'm not sure if it's worth the effort.

AJenbo:
> > This might explain the difference of opinion :)
> 
> Do you agree with @glebm then, that the new syntax is less readable? Or did you read until you found "C#" and dismissed it?

That was just meant as a light hearted message. I stand by my original statement (on Discord). I think it looks good, but I don't know enough to say if it clashes with some principle in C++. In the end I'm personally fine with either solution.

qndel:
> > > This might explain the difference of opinion :)
> > 
> > 
> > Do you agree with @glebm then, that the new syntax is less readable? Or did you read until you found "C#" and dismissed it?
> 
> That was just meant as a light hearted message. I stand by my original statement (on Discord). I think it looks good, but I don't know enough to say if it clashes with some principle in C++. In the end I'm personally fine with either solution.

Python best anyway
`if potato in [1,2,3]:`

julealgon:
https://github.com/julealgon/devilutionX/pull/3

AJenbo:
Booo
image

P.s. You forgot to remove Windows support...

PR #2828, opened 2021-09-08
Unhack hellfire objects

qndel:


AJenbo:
+5k lines, something tells me there is some deduplication needed here :)

qndel:
> +5k lines, something tells me there is some deduplication needed here :)

ye, this is just a draft, I don't think I'll ever do it, ideally we should copy/paste files like I did, merge them and cleanup slowly over time, otherwise I'm not going to deal with all the conflicts and stuff :D

AJenbo:
It looks like this would also solve: https://github.com/diasurgical/devilutionX/pull/1031

PR #2831, opened 2021-09-09
Improve collision detection for moving missiles

### LINKED ISSUE #1531: Blocking spells activates block animation and still deals damage

### LINKED ISSUE #2826: Arrows'/projectile spells' effective chance to hit influenced by projectile speed
**Important information**

https://github.com/diasurgical/devilutionX/commit/effee3f7bb7b3e3b868befb2006d3200e5e617a5

**Describe**
Slower missiles often perform a hit check multiple times as they spend time passing over a monster's tile, meaning that a projectile that should have low chance to hit gets multiple attempts to roll a hit and therefore has higher _effective_ chance to hit. This seems to be the more common case. However, faster missiles can sometimes skip a hit check altogether seemingly by never spending time on the same tile as a monster, meaning that faster missiles have lower _effective_ chance to hit, other things equal.

This can be tested by tracking the `hper` variable within the `MonsterMHit` method in missiles.cpp while running the game and shooting stuff with arrows and things like Firebolt. You can exaggerate the problem by altering the speed of arrows/firebolts to be extremely slow and by drastically reducing your (normal) hit chance, and more hit checks will be output even when firing single shots at one monster.

This seems like it might be a vanilla bug as well, but I'm unsure. Also possibly related to #1672 and #1531

**To Reproduce**
1. Get a character with bad chance to hit for magic or bows.
2. Shoot monsters.
3. Compare actual chance to successfully hit a monster with expected chance to hit.

**Expected behavior**
A projectile performs exactly one hit check against a monster it collides with. If it fails the first hit check, it continues past the initial monster and cannot attempt to hit it again. Faster projectiles should also always perform a hit check against a monster standing in the projectile's path.

**Screenshots**
sorctesting
hitchancedebug
I drastically lowered Sorcerer's hit chance in the code and shot some stuff on dlvl 1. The debug screenshot shows my chance to hit being returned on each hit check against a monster I shot at, and it also shows me tapping spacebar in between each shot. On the final shot against the Scavenger a couple tiles directly south of me, the game performed three hit checks as seen by the three 46s in a row, and ultimately landing the hit and killing it on the third hit check. This is with a normal speed Firebolt spell.

### LINKED ISSUE #2837: Arrows Bypass Target Tile at Certain Angles and Distances
**Important information**
Windows 10 / 1.2.1 / Downloaded Release

**Describe**
At certain angles and distances, your arrows will bypass the targeted tile. This can make it impossible to hit a monster without moving slightly to the side or another direction.

**To Reproduce**
1. Load attached save file.
2. Move mouse over the monster shown in the screenshot.
3. Click on it to your hearts content. Your arrows will never land.

**Expected behavior**
With no obstructions between the player and monster, the player should be able to hit them.

**Screenshots**
image

**Additional context**
If you move slightly to the right (non-isometrically, so thinking +1 in x-axis), the monster stays on screen and can be hit with arrows by clicking on them, and you can hit them pretty easily, so I do not believe this is a case of the to hit dropping to 0 from the distance.

[single_2.zip](https://github.com/diasurgical/devilutionX/files/7146605/single_2.zip)

obligaron:
Fixes #2826
Fixes #1531
Fixes #2837

Notes:

- First Commit is only refactoring as a base for the following commits
- Check if the missile skipped tiles and do also collision checking for these skipped tiles
- Prevent perform collision detection for a target more then once at a tile (that means you can also avoid a hit by missed hit chance and not only block)
- Tested with missiles 10x faster and 10x slower then normal

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -685,6 +685,7 @@ void LoadMissile(LoadHelper *file, Missile &missile)
 	missile.var6 = file->NextLE<int32_t>();
 	missile.var7 = file->NextLE<int32_t>();
 	missile.limitReached = file->NextBool32();
+	missile.lastCollisionTargetHash = 0;
 }
 
 void LoadObject(LoadHelper &file, Object &object)
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -616,6 +616,15 @@ void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, boo
 	else
 		possibleVisitTiles = prevTile.ManhattanDistance(missile.position.tile);
 
+	int16_t tileTargetHash = dMonster[missile.position.tile.x][missile.position.tile.y] ^ dPlayer[missile.position.tile.x][missile.position.tile.y];
+
+	if (possibleVisitTiles == 0) {
+		// missile didn't change the tile... check that we perform CheckMissileCol only once for any monster/player to avoid multiple hits for slow missiles
+		if (missile.lastCollisionTargetHash == tileTargetHash)
+			return;
+	}
+	// remember what target CheckMissileCol was checked against
+	missile.lastCollisionTargetHash = tileTargetHash;
 	// Did the missile skipped a tile?
 	if (possibleVisitTiles > 1) {
 		// Implementation note: If someone knows the correct math to calculate this without this step for step increase loop, I would really appreciate it.
@@ -2885,6 +2894,7 @@ int AddMissile(Point src, Point dst, Direction midir, missile_id mitype, mienemy
 	missile._miAnimType = missileData.mFileNum;
 	missile._miDrawFlag = missileData.mDraw;
 	missile._mlid = NO_LIGHT;
+	missile.lastCollisionTargetHash = 0;
 
 	if (missile._miAnimType == MFILE_NONE || MissileSpriteData[missile._miAnimType].animFAmt < 8)
 		SetMissDir(missile, 0);
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -590,10 +590,76 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
 		PlaySfxLoc(MissilesData[missile._mitype].miSFX, missile.position.tile);
 }
 
+/**
+ * @brief Could the missile collide with solid objects? (like walls or closed doors)
+ */
+bool CouldMissileCollideWithSolidObject(Point tile)
+{
+	int oid = dObject[tile.x][tile.y];
+	if (oid != 0) {
+		oid = abs(oid) - 1;
+		if (!Objects[oid]._oMissFlag)
+			return true;
+	}
+	return nMissileTable[dPiece[tile.x][tile.y]];
+}
+
 void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, bool ignoreStart, bool ifCollidesDontMoveToHitTile)
 {
+	Point prevTile = missile.position.tile;
 	missile.position.traveled += missile.position.velocity;
 	UpdateMissilePos(missile);
+
+	int possibleVisitTiles;
+	if (missile.position.velocity.deltaX == 0 || missile.position.velocity.deltaY == 0)
+		possibleVisitTiles = prevTile.WalkingDistance(missile.position.tile);
+	else
+		possibleVisitTiles = prevTile.ManhattanDistance(missile.position.tile);
+
+	// Did the missile skipped a tile?
+	if (possibleVisitTiles > 1) {
+		// Implementation note: If someone knows the correct math to calculate this without this step for step increase loop, I would really appreciate it.
+		auto incVelocity = missile.position.velocity * (0.01f / (float)(possibleVisitTiles - 1));
+		auto traveled = missile.position.traveled - missile.position.velocity;
+		do {
+			traveled += incVelocity;
+
+			// calculate in-between tile
+			int mx = traveled.deltaX >> 16;
+			int my = traveled.deltaY >> 16;
+			int dx = (mx + 2 * my) / 64;
+			int dy = (2 * my - mx) / 64;
+
+			auto tile = missile.position.start + Displacement { dx, dy };
+
+			// we are at the orginal calculated position => resume with normal logic
+			if (tile == missile.position.tile)
+				break;
+
+			// don't call CheckMissileCol more then once for a tile
+			if (prevTile == tile)
+				continue;
+			prevTile = tile;
+
+			CheckMissileCol(missile, mindam, maxdam, false, tile, false);
+
+			// Did missile hit anything?
+			if (missile._mirange != 0)
+				continue;
+
+			if ((missile._miHitFlag && MissilesData[missile._mitype].MovementDistribution == MissileMovementDistrubution::Blockable) || CouldMissileCollideWithSolidObject(tile)) {
+				missile.position.traveled = traveled;
+				if (ifCollidesDontMoveToHitTile && missile._mirange == 0) {
+					missile.position.traveled -= incVelocity;
+					UpdateMissilePos(missile);
+					missile.position.StopMissile();
+				} else {
+					UpdateMissilePos(missile);
+				}
+				return;
+			}
+		} while (true);
+	}
 	if (ignoreStart && missile.position.start == missile.position.tile)
 		return;
 	CheckMissileCol(missile, mindam, maxdam, false, missile.position.tile, false);
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -590,6 +590,20 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
 		PlaySfxLoc(MissilesData[missile._mitype].miSFX, missile.position.tile);
 }
 
+void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, bool ignoreStart, bool ifCollidesDontMoveToHitTile)
+{
+	missile.position.traveled += missile.position.velocity;
+	UpdateMissilePos(missile);
+	if (ignoreStart && missile.position.start == missile.position.tile)
+		return;
+	CheckMissileCol(missile, mindam, maxdam, false, missile.position.tile, false);
+	if (ifCollidesDontMoveToHitTile && missile._mirange == 0) {
+		missile.position.traveled -= missile.position.velocity;
+		UpdateMissilePos(missile);
+		missile.position.StopMissile();
+	}
+}
+
 void SetMissAnim(Missile &missile, int animtype)
 {
 	int dir = missile._mimfnum;
@@ -2856,8 +2870,6 @@ void MI_LArrow(Missile &missile)
 		MissilesData[missile._mitype].mResist = rst;
 	} else {
 		missile._midist++;
-		missile.position.traveled += missile.position.velocity;
-		UpdateMissilePos(missile);
 
 		int mind;
 		int maxd;
@@ -2874,16 +2886,13 @@ void MI_LArrow(Missile &missile)
 			maxd = GenerateRnd(10) + 1 + currlevel * 2;
 		}
 
-		if (missile.position.tile != missile.position.start) {
-			missile_resistance rst = MissilesData[missile._mitype].mResist;
-			MissilesData[missile._mitype].mResist = MISR_NONE;
-			CheckMissileCol(missile, mind, maxd, false, missile.position.tile, false);
-			MissilesData[missile._mitype].mResist = rst;
-		}
+		missile_resistance rst = MissilesData[missile._mitype].mResist;
+		MissilesData[missile._mitype].mResist = MISR_NONE;
+		MoveMissileAndCheckMissileCol(missile, mind, maxd, true, true);
+		MissilesData[missile._mitype].mResist = rst;
+
 		if (missile._mirange == 0) {
 			missile._mimfnum = 0;
-			missile.position.traveled -= missile.position.velocity;
-			UpdateMissilePos(missile);
 			if (missile._mitype == MIS_LARROW)
 				SetMissAnim(missile, MFILE_MINILTNG);
 			else
@@ -2909,8 +2918,6 @@ void MI_Arrow(Missile &missile)
 {
 	missile._mirange--;
 	missile._midist++;
-	missile.position.traveled += missile.position.velocity;
-	UpdateMissilePos(missile);
 	int p = missile._misource;
 
 	int mind;
@@ -2927,8 +2934,7 @@ void MI_Arrow(Missile &missile)
 		mind = currlevel;
 		maxd = 2 * currlevel;
 	}
-	if (missile.position.tile != missile.position.start)
-		CheckMissileCol(missile, mind, maxd, false, missile.position.tile, false);
+	MoveMissileAndCheckMissileCol(missile, mind, maxd, true, false);
 	if (missile._mirange == 0)
 		missile._miDelFlag = true;
 	PutMissile(missile);
@@ -2940,10 +2946,6 @@ void MI_Firebolt(Missile &missile)
 
 	missile._mirange--;
 	if (missile._mitype != MIS_BONESPIRIT || missile._mimfnum != 8) {
-		int omx = missile.position.traveled.deltaX;
-		int omy = missile.position.traveled.deltaY;
-		missile.position.traveled += missile.position.velocity;
-		UpdateMissilePos(missile);
 		int p = missile._misource;
 		if (p != -1) {
 			if (missile._micaster == TARGET_MONSTERS) {
@@ -2968,14 +2970,9 @@ void MI_Firebolt(Missile &missile)
 		} else {
 			d = currlevel + GenerateRnd(2 * currlevel);
 		}
-		if (missile.position.tile != missile.position.start) {
-			CheckMissileCol(missile, d, d, false, missile.position.tile, false);
-		}
+		MoveMissileAndCheckMissileCol(missile, d, d, true, true);
 		if (missile._mirange == 0) {
 			missile._miDelFlag = true;
-			missile.position.traveled = { omx, omy };
-			UpdateMissilePos(missile);
-			missile.position.StopMissile();
 			missile.var4 = -(AvailableMissiles[0] + 1);
 			Point dst = { 0, 0 };
 			auto dir = static_cast<Direction>(missile._mimfnum);
@@ -3042,10 +3039,8 @@ void MI_Lightball(Missile &missile)
 	int tx = missile.var1;
 	int ty = missile.var2;
 	missile._mirange--;
-	missile.position.traveled += missile.position.velocity;
-	UpdateMissilePos(missile);
 	int j = missile._mirange;
-	CheckMissileCol(missile, missile._midam, missile._midam, false, missile.position.tile, false);
+	MoveMissileAndCheckMissileCol(missile, missile._midam, missile._midam, false, false);
 	if (missile._miHitFlag)
 		missile._mirange = j;
 	int8_t obj = dObject[tx][ty];
@@ -3118,10 +3113,7 @@ void MI_Fireball(Missile &missile)
 		}
 	} else {
 		int dam = missile._midam;
-		missile.position.traveled += missile.position.velocity;
-		UpdateMissilePos(missile);
-		if (missile.position.tile != missile.position.start)
-			CheckMissileCol(missile, dam, dam, false, missile.position.tile, false);
+		MoveMissileAndCheckMissileCol(missile, dam, dam, true, false);
 		if (missile._mirange == 0) {
 			Point m = missile.position.tile;
 			ChangeLight(missile._mlid, missile.position.tile, missile._miAnimFrame);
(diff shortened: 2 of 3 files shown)


Source/missiles.cpp
```diff
@@ -590,6 +590,90 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
...
+}
+
+void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, bool ignoreStart, bool ifCollidesDontMoveToHitTile)
+{
+	Point prevTile = missile.position.tile;
+	missile.position.traveled += missile.position.velocity;
+	UpdateMissilePos(missile);
+	int walkingDistance = prevTile.WalkingDistance(missile.position.tile);
+
+	int16_t tileTargetIdentifer = dMonster[missile.position.tile.x][missile.position.tile.y] - dPlayer[missile.position.tile.x][missile.position.tile.y];
```
ephphatha, line 614: Given the way dMonster and dPlayer contain negated indexes for moving entities I think using bitwise combinations would be more straightforward than subtraction here.

XOR should be fine, could also shift the value in dMonster left 4-8 bits and mask the player index onto the low bits to avoid collisions (which is already pretty unlikely).
```suggestion
	int16_t tileTargetIdentifer = dMonster[missile.position.tile.x][missile.position.tile.y] ^ dPlayer[missile.position.tile.x][missile.position.tile.y];
```

Source/missiles.cpp
```diff
@@ -590,6 +590,90 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
...
+	if (walkingDistance == 0) {
+		// missile didn't change the tile... check that we perform CheckMissileCol only once for any monster/player to avoid multiple hits for slow missiles
+		if (missile.lastCollisionTargetIdentifer == tileTargetIdentifer)
+			return;
+	}
+	// remember what target CheckMissileCol was checked against
+	missile.lastCollisionTargetIdentifer = tileTargetIdentifer;
+	// Did the missile skipped a tile?
+	if (walkingDistance > 1) {
+		// Implementation note: If someone knows the correct math to calculate this without this step for step increase loop, I would really appreciate it.
```
ephphatha, line 625: This can be done by checking all the discrete tiles between prevTile and the current position, I'll have a go at simplifying the logic when I finish work.

(Been meaning to do the same for the animation logic as well, that's a case of working out the distance to the tile border as a proportion of the total movement in a frame then scaling the predicted movement by the result. That case can be done without a loop.)

ephphatha (commented):
Combining missile movement and collision checking seems sensisble (especially given the observed bugs 😂 ). A few comments from a quick read through the combined function, I'll have a look at the call sites when I have more time this evening.

Source/missiles.cpp
StephenCWills: Rather than unlikely, I think collisions are impossible. Can players and monsters occupy the same tile simultaneously?

StephenCWills: If the indexes are negated for moving entities, doesn't that mean we might still check collisions twice for the same entity in the same tile if the entity starts or stops moving while the missile travels through that tile? Perhaps we should use `abs()`.

Source/missiles.cpp
```diff
@@ -590,6 +590,90 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
...
+
+	if (walkingDistance == 0) {
+		// missile didn't change the tile... check that we perform CheckMissileCol only once for any monster/player to avoid multiple hits for slow missiles
+		if (missile.lastCollisionTargetIdentifer == tileTargetIdentifer)
+			return;
+	}
+	// remember what target CheckMissileCol was checked against
+	missile.lastCollisionTargetIdentifer = tileTargetIdentifer;
+	// Did the missile skipped a tile?
+	if (walkingDistance > 1) {
```
StephenCWills, line 624: If a missile passes over an adjacent tile into a diagonal tile in a single bound, this would skip the collision test for the adjacent tile.

Source/missiles.cpp
ephphatha: > Rather than unlikely, I think collisions are impossible. Can players and monsters occupy the same tile simultaneously?
> 
> EDIT: I suppose maybe you meant hash collisions. 🤦

I did mean hash collisions :D, however hash collisions are only an issue if a player occupies the same tile as the golem they control, which I don't think is possible even if both entities are moving? Given the point you bring up in the second comment about hitting a moving entity multiple times using the first non-zero absolute value sounds fine.

StephenCWills: > I did mean hash collisions :D, however hash collisions are only an issue if a player occupies the same tile as the golem they control, which I don't think is possible even if both entities are moving?

I have bad news.

https://user-images.githubusercontent.com/9203145/132784190-43ff7aec-6fd4-44ef-8e80-75481b2cb77f.mp4

StephenCWills: While scanning through the DSF Buglist just now, I found this text on the subject of sharing tiles.

> **SHARING SQUARES**
> 
> When a character teleports/phases onto a square that a monster was moving in to, they will overlap. The character will have to move off the square to hit the monster. The monster, however, can attack the same square it's standing on.
> 
> If two characters wind up occupying the same square, which can happen rarely when casting Teleport or when Resurrecting while standing over a body, the player who does not move will become invisible to both players, until he/she moves. Further, the player first on the internal list of players may be the first targeted and attacked if a monster approaches.
> 
> The player can also share a square with a Golem.

NiteKat: A player can also share a tile with a monster by spawning on that tile while a monster occupies it. The function for spawning players on a floor (such as by taking the stairs down) will put a player on a tile occupied by a monster happily. When this happens, the player cannot hit the monster until they move, but the monster is free to melee the player to death. Fun times. (if you attack the monster, your attack goes straight down on screen)

julealgon: > I have bad news.

> A player can also share a tile with a monster by spawning on that tile while a monster occupies it.

I think it is a bad idea to base new logic on what are clearly existing bugs. Instead, we should fix the bugs and prevent the situation from ever happening.

StephenCWills: > I think it is a bad idea to base new logic on what are clearly existing bugs. Instead, we should fix the bugs and prevent the situation from ever happening.

On that note, I think the hash collisions are pretty much a non-issue. Even in an imagined future state of the game where any monster could share an ID with any player, it would be extremely difficult to create a situation where the hash collision actually made a difference to this logic. Furthermore, the impact of the hash collision is negligible.

Source/missiles.cpp
```diff
@@ -590,6 +590,90 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
...
 }
 
+/**
+ * @brief Could the missile collide with solid objects? (like walls or closed doors)
+ */
+bool CouldMissileCollideWithSolidObject(Point tile)
+{
+	int oid = dObject[tile.x][tile.y];
+	if (oid != 0) {
+		oid = oid > 0 ? oid - 1 : -(oid + 1);
```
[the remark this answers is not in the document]
obligaron: Copied from the wrong location. 😁 
Fixed, thanks. 🙂

Source/missiles.h
```diff
@@ -124,6 +124,10 @@ struct Missile {
 	int var6;
 	int var7;
 	bool limitReached;
+	/**
+      * @brief For moving missiles lastCollisionTargetIdentifer contains the last entity (player or monster) that was checked in CheckMissileCol (needed to avoid multiple hits for a entity at the same tile).
+      */
+	int16_t lastCollisionTargetIdentifer;
```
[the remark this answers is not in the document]
obligaron: Renamed it to `lastCollisionTargetHash` 🙂

Source/missiles.cpp
obligaron: I'm also not sure if we should care much about it.
For me this are also bugs that should be fixed and @qndel was fast and fixed the first one already. 👍 
Do we have a issue for the other player spawning case?

But if wanted we could improve it now or in a later pr.
We could for example add two fields to `Missile`, one for player and one for monster.

NiteKat: I don't think there's an issue for players and monsters occupying the same tiles. Though, now I'm curious about other ways to possibly do it.... if I get some time this weekend, I'll try to get some save files and confirm things and get some issues open.

Source/missiles.cpp
obligaron: This would be great! Really looking forward to it. 🙂 
But I think the current version could be good enough to get this pr merged. Cause I dislike to have pr's too long open.

Source/missiles.cpp
obligaron: Totally missed this one. 😄 
I tried to fix it with checking coordinates diff and only use `WalkingDistance` if the missile flies straight in one direction (only x or y).
Hope you can look again. 🙂

Source/missiles.cpp
```diff
@@ -590,6 +590,90 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
...
+	}
+	return nMissileTable[dPiece[tile.x][tile.y]];
+}
+
+void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, bool ignoreStart, bool ifCollidesDontMoveToHitTile)
+{
+	Point prevTile = missile.position.tile;
+	missile.position.traveled += missile.position.velocity;
+	UpdateMissilePos(missile);
+	int possibleVisitTiles = (missile.position.velocity.deltaX == 0 || missile.position.velocity.deltaY == 0) ? prevTile.WalkingDistance(missile.position.tile) : abs(prevTile.x - missile.position.tile.x) + abs(prevTile.y - missile.position.tile.y);
```
StephenCWills, line 612: Good thinking on checking if the missile is flying straight. I thought I'd mention that I saw the following function when I was reviewing this PR yesterday. It looks like you're using effectively the same logic in the ternary else clause.

https://github.com/diasurgical/devilutionX/blob/ec1850e2950f015d7814a8d9472bde471f4cb0d6/Source/engine/point.hpp#L142-L147

obligaron: I also think it's the same, switched to `ManhattanDistance`.
This way it looks cleaner and less code. Thanks 🙂

Source/missiles.cpp
StephenCWills: I still think that absolute value is a good idea, but the impact is similarly negligible so I'll approve the PR regardless.

StephenCWills (dismissed):
I'm impressed by the speedy delivery of this PR. Since this does have an impact on game balance, I'd like to leave it to @AJenbo to decide whether to merge this now, but I approve of the changes regardless.

Source/missiles.cpp
obligaron: I changed it to suggested `^`, it also doesn't hurt. 🙂

StephenCWills (approved):

Source/missiles.cpp
ephphatha: Yep, I also got sidetracked yesterday and didn't get around to looking at it sorry 😅 . I'll look at both of these cases as a separate PR (though happy for someone else to take a look too).

StephenCWills: TBH, I played around with the math a bit yesterday and found it to be rather difficult to implement. I was relieved when you said you would look into it, but I'd be happy to give it another try.

Source/loadsave.cpp
```diff
@@ -685,6 +685,7 @@ void LoadMissile(LoadHelper *file, Missile &missile)
 	missile.var6 = file->NextLE<int32_t>();
 	missile.var7 = file->NextLE<int32_t>();
 	missile.limitReached = file->NextBool32();
+	missile.lastCollisionTargetHash = 0;
```
AJenbo, line 688: There are 3 bytes available from alignment, afaik they will always be zero for vanilla saves and it would allow us to retain the sate past saves.

Source/missiles.cpp
```diff
@@ -590,6 +590,90 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
...
+}
+
+void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, bool ignoreStart, bool ifCollidesDontMoveToHitTile)
+{
+	Point prevTile = missile.position.tile;
+	missile.position.traveled += missile.position.velocity;
+	UpdateMissilePos(missile);
+	int possibleVisitTiles = (missile.position.velocity.deltaX == 0 || missile.position.velocity.deltaY == 0) ? prevTile.WalkingDistance(missile.position.tile) : prevTile.ManhattanDistance(missile.position.tile);
+
+	int16_t tileTargetHash = dMonster[missile.position.tile.x][missile.position.tile.y] ^ dPlayer[missile.position.tile.x][missile.position.tile.y];
```
AJenbo, line 614: I prefer lines to be 80, and hard limited to 120 chars. This one is 212 which just barley fits on half of my ultra wide monitor in this differ with 100% zoom.

```suggestion
	Point position = missile.position.tile;
	int possibleVisitTiles = (missile.position.velocity.deltaX == 0 || missile.position.velocity.deltaY == 0) ? prevTile.WalkingDistance(position) : prevTile.ManhattanDistance(position);

	int16_t tileTargetHash = dMonster[position.x][position.y] ^ dPlayer[position.x][position.y];
```
Even with this change it's still 186, but that's a bit better. We can wait with further reducing till it's gets added as a clang rule :)

Source/missiles.cpp
StephenCWills: @ephphatha My current line of thinking is rather than working out the distance to the edge of each tile, it would probably be a lot easier to calculate the points of collision with horizontal/vertical lines that cut through the centers of the tiles. This way, you could simply iterate through the horizontal/vertical lines in the missile's path and handle corners as a special case. Plus, you shouldn't have to deal with the possibility of truncation errors putting the missile in the wrong tile during calculations like you would when straddling the edge.

Since I'm struggling with the implementation, I figured I'd share this idea in case the information helps get you there faster than me.

image

AJenbo (dismissed):
Definitely something we should merge, it solves most of the bugs caused by the dodgy collision code.

I do slightly wish that there was less "will work if x never happens", but I don't feel they are unreasonable assumptions and the alternative would probably be some major reworkings.

Since there are a few active debates still going, I will not merge it right away but anyone is free to do so if you feel it makes more sense to revise with additional PRs.

Source/missiles.cpp
obligaron: If you want I can change it to this:
```cpp
	int possibleVisitTiles;
	if (missile.position.velocity.deltaX == 0 || missile.position.velocity.deltaY == 0)
		possibleVisitTiles = prevTile.WalkingDistance(missile.position.tile);
	else
		possibleVisitTiles = prevTile.ManhattanDistance(missile.position.tile);
```
This has max 87 chars and avoids a new local variable... but now we get 4 instead of 1 line. 😉

Source/loadsave.cpp
obligaron: I'm not sure if we should use it for this atm, cause
- DevilutionX (includes Version 1.2.1) writes uninitialized data to this (was changed in #2287). So I think it would be good to have version 1.3 out so that most saves should contain zero initialized values (saved in vanilla or devilutionx 1.3).
- The effect is not that great. At maximum you get hit twice once.
- I feel like this bytes are precious and perhaps we could use them for something even better. :innocent:

Source/missiles.cpp
qndel: funny, I wanted to add manhattan distance recently :D

AJenbo: To me long lines are worse then many lines. They are very hard to see structure in and if line length is not standardized eh complexity of two function with the same line count can be vastly different.

I'm not a big fan of uninitialized variables, but seeing this I think it's the best of the 3 options.

AJenbo: > funny, I wanted to add manhattan distance recently :D

You are 3 months late :)

obligaron: Changed it. 🙂

Chance4us:
I haven't tested it for long, but the first impression was that you hit the opponents much easier. This greatly reduces the difficulty of the game. You could of course do some balancing by making the opponents stronger in a way, but it's all a balancing act between the bugs by design and the feeling of originality. In my opinion, we should therefore consider making the change unlockable via an optional switch.

StephenCWills:
> I haven't tested it for long, but the first impression was that you hit the opponents much easier. This greatly reduces the difficulty of the game.

If you are using arrows or high level spells, that would make sense. Arrows move at a speed very close to 1 tile per tick so you'd be much more likely to miss a tile that the projectile passes through than get multiple collisions in the same tile.

> You could of course do some balancing by making the opponents stronger in a way, but it's all a balancing act between the bugs by design and the feeling of originality.

I originally thought we'd need to adjust the hit chance of missiles, but I've changed my mind precisely because arrow speed is so close to one tile per tick. I'm still a little concerned that this could adversely affect low-level mages which are notoriously difficult already, but otherwise this just feels like a nasty bug.

> In my opinion, we should therefore consider making the change unlockable via an optional switch.

I assume you mean to say that the hit chance adjustment could be an optional change, not this bug. In that case, I'd also say it's worth considering.

Chance4us:
The hit chance of the arrows should be reduced and the hit chance of the rocket should be increased (see # 2890),
otherwise we are dealing with a heavily modified gameplay, which does not even come close to the original.
The way it feels now is a different game.

Chance4us:
@StephenCWills I don't mean to say that the changes are fundamentally wrong or bad.
I don't mean that in a negative way and such changes are of course welcome. 
After all, they fix bugs that were not considered or overlooked during development.
Since these actually feel a lot different from the original, I recommend combining them into one option.
Or the other way around, to summarize the vanilla peculiarities and diseases in one option and to award with the title "hit chance classic mode" for example.

StephenCWills:
> The hit chance of the arrows should be reduced and the hit chance of the missiles should be increased (see #2890),
> otherwise we are dealing with a heavily modified gameplay mechanics, which does not even come close to the original.
> The way it feels now is an another different game that wants to be perfect.

I recommend you review #2826 to gain a better understanding of what effect this PR has on projectiles. If you insist that the accuracy of arrows needs to be reduced, you're going to have to convince me with data. Spells are much trickier because the projectile speed is affected by spell level. I can appreciate that much of your work here is based on feel, and I don't want to discount that. It's just that the solution can't be as simple as "turn these two knobs and we have the original game again".

Chance4us:
I understand that. I see that from the player's side. And I'm not a good programmer at all. From here on we're slowly moving into a gray area, where other players may also say, wait a minute, the original was somehow different. What do we want to achieve with Devilutionx, a mod or the original just with more comfort and fewer annoying bugs? The players will think that we will no longer hit the enemy in the same way as before and also the other way around, my hero will be hit differently or, strangely enough, not hit.

I can't convince you with data or numbers,
but I can roughly describe the problem with the arrows.
You have to imagine you are in level 5 and surrounded by the monsters (Goat Man https://diablo-archive.fandom.com/wiki/Goat_Man_(Diablo_I)). These monsters love to run around, criss-cross, orbit the hero until they attack him directly. While they are circling the hero they are as good as impossible to hit, especially the most difficult to hit diagonally. That makes the game very tough and difficult in this situation.
The current change makes it possible to hit opponents more easily at any angle, and it makes the game feel less difficult. 

With the missiles on the other hand, I felt the opposite effect and demonstrated it with the help of the numbers. Here we are dealing with a reduced hit rate. As a player, I find the bad hit rate with the missiles worse than the problem with the arrows, because the missile flies through for no good reason and even though you aim very precisely. I don't know how but we should find a compromise, because otherwise the game feels either too easy or too difficult depending on the situation.

julealgon:
> While they are circling the hero they are as good as impossible to hit, especially the most difficult to hit diagonally. That makes the game very tough and difficult in this situation.

That was because the hit checks were completely wrong.

> The current change makes it possible to hit opponents more easily at any angle, and it makes the game feel less difficult.

There were other changes that impact gameplay as well, for example the removal of horizontal walking, which will make the game harder. Slower spells will also hit less, which makes the game harder.

Additionally, missile logic not only impacts players, but monsters too. So monsters will hit you with projectiles more often when you are moving, and less often if they fire very slow missiles (due to previous multiple hit checks).

I don't think we should try to rebalance the game. We are here to fix bugs. If someone wants a rebalance after the bugs are fixed, that should be done in a mod.

AJenbo:
> What do we want to achieve with Devilutionx, a mod or the original just with more comfort and fewer annoying bugs?

Getting hit 11 times by the same missile is very annoying, also arrows skipping some tiles making it impossible to hit is also very annoying. I do appreciate that fixing issues like this will affect game balance and as such it is something we have to carefully evaluate additionally. But I think we already started down this road by fixing the AI not being able to locate and avoid firewalls as well as firewalls dealing damage based on dungeon level instead of hero level.

Issue #2833, opened 2021-09-10
Golem can spawn on a player's tile


StephenCWills:
**To Reproduce**
Cast the Golem spell on top of a player or on an object directly northeast of a player.

**Expected behavior**
Golem should spawn on a nearby empty tile instead of the player's tile.

**Screenshots**

https://user-images.githubusercontent.com/9203145/132797058-411e180c-27d1-42af-b9fc-75cdec8ac702.mp4

**Additional context**
It seems the `AddGolem()` function uses the crawl table to locate the nearest dungeon tile that isn't occupied, but it only checks for monsters, solid tiles, and objects. The crawl table checks southwest first if the initial tile is occupied so that's how I did it in the video, but it seems you can literally just cast at the player's feet and the golem will spawn there.

This bug is actually documented on lurkerlounge, although their description suggests you need to be standing next to an enemy. I suppose that just keeps the Golem from moving right away.

> http://www.lurkerlounge.com/forums/thread-12422.html
> If, while standing adjacent to an enemy, you cast a golem with your cursor on yourself, the golem will appear on your tile. Your golem takes no damage with this bug.

julealgon:
Damn this one was quick 🤣

PR #2834, opened 2021-09-10
fix golem spawning on top of player

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2833

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2377,7 +2377,7 @@ void AddGolem(Missile &missile, Point dst, Direction /*midir*/)
 				if (!LineClearMissile(missile.position.start, target))
 					continue;
 
-				if (dMonster[target.x][target.y] != 0 || nSolidTable[dPiece[target.x][target.y]] || dObject[target.x][target.y] != 0)
+				if (dMonster[target.x][target.y] != 0 || nSolidTable[dPiece[target.x][target.y]] || dObject[target.x][target.y] != 0 || dPlayer[target.x][target.y] != 0)
 					continue;
 
 				SpawnGolem(playerId, target, missile);


StephenCWills (approved):

AJenbo:
Could also use the MonsterPosOk function to allow them on non solid objects and doors :shrug:

Source/missiles.cpp
```diff
@@ -2377,7 +2377,7 @@ void AddGolem(Missile &missile, Point dst, Direction /*midir*/)
 				if (!LineClearMissile(missile.position.start, target))
 					continue;
 
-				if (dMonster[target.x][target.y] != 0 || nSolidTable[dPiece[target.x][target.y]] || dObject[target.x][target.y] != 0)
+				if (dMonster[target.x][target.y] != 0 || nSolidTable[dPiece[target.x][target.y]] || dObject[target.x][target.y] != 0 || dPlayer[target.x][target.y] != 0)
```
julealgon, line 2380: Don't we have a "tile is free" function somewhere we could reuse here?

StephenCWills: Maybe [CanPlaceMonster()](https://github.com/diasurgical/devilutionX/blob/0d3bf15601ab2a45464c72ec115673f7833acdc5/Source/monster.cpp#L265) or [IsTileAvailable()](https://github.com/diasurgical/devilutionX/blob/0d3bf15601ab2a45464c72ec115673f7833acdc5/Source/monster.cpp#L2015). I don't see a `MonsterPosOk()` function anywhere.

* The former checks dungeon bounds, `dMonster`, `dPlayer`, `dFlags` for `BFLAG_VISIBLE` and `BFLAG_POPULATED`, and `nSolidTable`, but not objects.
* The latter checks `dMonster`, `dPlayer`, `dObject`, and `nSolidTable`, but doesn't check dungeon bounds or any `dFlags`.

AJenbo: Making functions public is allow, yeah I didn't really remember the name but there is definitely a function like this that gets used during navigation

Issue #2735, opened 2021-08-28
Move documentation from implementations to declarations


AJenbo:
**Describe**
A lot of documentation has been added to the implementation in .cpp files, but the proper place for it is in the header files (.h).

**Steps**
1. Find functions with documentation comments
2. Cut the comment
3. Insert it above the function definition in the .h file

**Additional context**
![image](https://user-images.githubusercontent.com/204594/131220345-ae1235c8-024e-46a5-bca1-aa3c81c482ae.png)
![image](https://user-images.githubusercontent.com/204594/131220367-359de284-5b03-49b0-bc8b-3e9a2efdf2ac.png)

qndel:
What about functions that aren't in headers? ;)

AJenbo:
They are locale so there descriptions stay in the .cpp file, they are easy to spot as they are all in the anonymous `namespace {}`.

bcko:
I will work on this.

Sid672:
@bcko if u are not working on this. I would like to contribute to this issue. 
@AJenbo can I work on this?

bcko:
@Sid672 sure, you can also reference #2740

AJenbo:
>  @AJenbo can I work on this?

Please :) There is plenty of it.

Sid672:
@AJenbo  created a pull request https://github.com/diasurgical/devilutionX/pull/2767 after moved documentation of town.cpp to town.h. Please review it.

Sid672:
Thanks, @AJenbo  for explaining everything. I start working on other files too.

Sid672:
@AJenbo I think all files are completed. If any file is remaining please tell me.

PR #2835, opened 2021-09-10
Move documentation from implementations to declarations #2735

Sid672:
setmaps.cpp doc is moved to setmaps.h  please review it.


ephphatha:
Please squash this commit history, there is no need to have 23 commits which result in 3 lines moving from one file to another.

Sid672:
> Please squash this commit history, there is no need to have 23 commits which result in 3 lines moving from one file to another.

okay.

PR #2843, opened 2021-09-11
Ensure scrolls and charges are cleared using correct spell after use

### LINKED ISSUE #792: Scrolls and charges not consumed correctly if swapped before hit-frame

julealgon:
This PR changes the variable we rely on to match the scroll or staff spell when removing a scroll or a staff charge that was just used via right clicking. 

Instead of picking the spell from the "readied spell" value (the spell currently selected), we pick the spell being cast by the player. This ensures the player cannot change the spell into something else right after casting and either causing wrong scrolls to be removed or no scrolls/charges removed at all.

Fixes #792

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1911,7 +1911,7 @@ void RemoveScroll(Player &player)
 	for (int i = 0; i < player._pNumInv; i++) {
 		if (!player.InvList[i].isEmpty()
 		    && (player.InvList[i]._iMiscId == IMISC_SCROLL || player.InvList[i]._iMiscId == IMISC_SCROLLT)
-		    && player.InvList[i]._iSpell == player._pRSpell) {
+		    && player.InvList[i]._iSpell == player._pSpell) {
 			player.RemoveInvItem(i);
 			player.CalcScrolls();
 			return;
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1960,7 +1960,7 @@ void UseStaffCharge(Player &player)
 {
 	auto &staff = player.InvBody[INVLOC_HAND_LEFT];
 
-	if (!CanUseStaff(staff, player._pRSpell))
+	if (!CanUseStaff(staff, player._pSpell))
 		return;
 
 	staff._iCharges--;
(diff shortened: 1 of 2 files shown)


Source/inv.cpp
```diff
@@ -1911,7 +1911,7 @@ void RemoveScroll(Player &player)
 	for (int i = 0; i < player._pNumInv; i++) {
 		if (!player.InvList[i].isEmpty()
 		    && (player.InvList[i]._iMiscId == IMISC_SCROLL || player.InvList[i]._iMiscId == IMISC_SCROLLT)
-		    && player.InvList[i]._iSpell == player._pRSpell) {
+		    && player.InvList[i]._iSpell == player._pSpell) {
```
AJenbo, line 1914: Why only for the belt and not the inv?

AJenbo (approved):

AJenbo (approved):

PR #2847, opened 2021-09-12
tiledata debug command

qndel:
lets you display values from many internal arrays (works best combined with grid)
For example dMonster in action
dMonster


AJenbo (approved):

AJenbo:
Looks very useful for debugging

PR #2852, opened 2021-09-12
Improve missile collision interpolation logic

StephenCWills:
The idea behind this is to efficiently traverse all the tiles in a missile's path by interpolating its position in half-tile increments along one axis. The following image provides a visual representation of the approach. Black lines represent the edges of tiles, red lines represent the half-tile increments, and orange dots represent the interpolated positions of the missile used to check for collisions.

image

---

I've been testing extensively using the savegame from #2837, and I am continually finding edge cases that break this logic. There is an issue with the way missiles travel through tile space due to a distortion along the diagonals caused by integer truncation. I find it too frustrating to continue my attempts to work around that issue and so I think it would be better to fix missile paths first and then revisit this logic.

For posterity, here is a visual representation of the space distortion. The center of the largest tile is the starting position of the missile.

image


Source/missiles.cpp
```diff
@@ -629,47 +743,9 @@ void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, boo
 	missile.lastCollisionTargetHash = tileTargetHash;
 	// Did the missile skipped a tile?
```
qndel, line 744:
```suggestion
	// Did the missile skip a tile?
```

StephenCWills:
I decided to revisit this now that the missile space distortion is fixed. I thought about how to simplify the logic so I wouldn't have to use so many lambdas and decided to just compute `incVelocity` a little differently. I think, given the amount of precision in these computations, the cumulative error accrued within one frame isn't going to cause any changes to missile behavior.

That said, I wasn't expecting this to cause the timedemo to fail. I thought that if anything changed, it would be extremely rare cases where a missile stops on a corner between tiles, but it seems I was wrong. I added some debug logs to the code and compared the output from the timedemo between master and this branch. It seems that master has a bit of trouble when straddling the line between two tiles that causes it to alternate collision detection a little bit. You can see it happen a couple times in the following diff. The good news is that this PR seems to fix that issue, but the bad news is that we need yet another update to the timedemo. 😩 

```diff
diff --git a/Source/missiles.cpp b/Source/missiles.cpp
index 626b539bc..e4450a6bb 100644
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -24,6 +24,7 @@
 #include "lighting.h"
 #include "monster.h"
 #include "spells.h"
+#include "utils/log.hpp"
 #include "utils/str_cat.hpp"
 
 namespace devilution {
@@ -528,6 +529,7 @@ bool MoveMissile(Missile &missile, tl::function_ref<bool(Point)> checkTile, bool
 void MoveMissileAndCheckMissileCol(Missile &missile, int mindam, int maxdam, bool ignoreStart, bool ifCollidesDontMoveToHitTile)
 {
 	auto checkTile = [&](Point tile) {
+		LogDebug("[SCW] checkTile({}, {}, {})", missile._mirange, tile.x, tile.y);
 		if (ignoreStart && missile.position.start == tile)
 			return true;
 
```

image

AJenbo:
I guess that gives me an excuse to play the game some more

PR #2853, opened 2021-09-12
Move documentation from implementations to declarations.

Sid672:
Functional documentation  of pack.cpp moved to pack.h


AJenbo (approved):

qndel:
This is such a waste of time to keep merging changes in one file, can't you do a bit more in a single PR? Technically it should all be done in 1 commit because it's work for 30 min max.

Sid672:
Thanks for the advice @qndel. I will try to complete this in a single PR. All these things are new to me, so I am doing it in multiple PR.

PR #2855, opened 2021-09-12
mpqapi: saner handling of hash/block tables

vlolteanu:
The rationale for this is two-fold:

- `std::vector` is used to store the tables;  `new` and `delete` are gone
- Caching of the tables was brittle: basically, the programmer basically promised that the same archive would be loaded next time, and that was that. This PR checks the name of the file; if it doesn't match, the cache is disregarded.

### DIFF
--- a/Source/mpqapi.cpp
+++ b/Source/mpqapi.cpp
@@ -32,6 +32,32 @@ namespace devilution {
 
 namespace {
 
+struct FileHeader {
+	uint32_t signature;
+	int headersize;
+	uint32_t filesize;
+	uint16_t version;
+	int16_t sectorsizeid;
+	int hashoffset;
+	int blockoffset;
+	int hashcount;
+	int blockcount;
+	uint8_t pad[72];
+};
+
+struct HashEntry {
+	uint32_t hashcheck[2];
+	uint32_t lcid;
+	int32_t block;
+};
+
+struct BlockEntry {
+	uint32_t offset;
+	uint32_t sizealloc;
+	uint32_t sizefile;
+	uint32_t flags;
+};
+
 // Validates that a Type is of a particular size and that its alignment is <= the size of the type.
 // Done with templates so that error messages include actual size.
 template <std::size_t A, std::size_t B>
@@ -49,8 +75,8 @@ struct CheckSize : AssertEq<sizeof(T), S>, AssertLte<alignof(T), sizeof(T)> {
 // Check sizes and alignments of the structs that we decrypt and encrypt.
 // The decryption algorithm treats them as a stream of 32-bit uints, so the
 // sizes must be exact as there cannot be any padding.
-static_assert(CheckSize<_HASHENTRY, 4 * 4>::value, "sizeof(_HASHENTRY) == 4 * 4 && alignof(_HASHENTRY) <= 4 * 4 not satisfied");
-static_assert(CheckSize<_BLOCKENTRY, 4 * 4>::value, "sizeof(_BLOCKENTRY) == 4 * 4 && alignof(_BLOCKENTRY) <= 4 * 4 not satisfied");
+static_assert(CheckSize<HashEntry, 4 * 4>::value, "sizeof(HashEntry) == 4 * 4 && alignof(HashEntry) <= 4 * 4 not satisfied");
+static_assert(CheckSize<BlockEntry, 4 * 4>::value, "sizeof(BlockEntry) == 4 * 4 && alignof(BlockEntry) <= 4 * 4 not satisfied");
 
 const char *DirToString(std::ios::seekdir dir)
 {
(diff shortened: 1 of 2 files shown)


AJenbo:
Haven't looked at this, but i'm pretty excited :D

Source/mpqapi.cpp
```diff
@@ -385,47 +419,45 @@ void AllocBlock(uint32_t blockOffset, uint32_t blockSize)
 
 int FindFreeBlock(uint32_t size, uint32_t *blockSize)
 {
-	int result;
-
-	_BLOCKENTRY *pBlockTbl = cur_archive.sgpBlockTbl;
-	for (int i = 0; i < INDEX_ENTRIES; i++, pBlockTbl++) {
-		if (pBlockTbl->offset == 0)
+	for (int i = 0; i < INDEX_ENTRIES; i++) {
```
ephphatha, line 422: This can use a range based loop
```suggestion
	for (BlockEntry &block : cur_archive.blockTbl) {
```

`i` is not used in the loop beyond the first index, `block` is always dereferenced so doesn't need to be a pointer.

Source/mpqapi.cpp
```diff
@@ -614,32 +646,35 @@ bool mpqapi_has_file(const char *pszName)
...
+	} else {
+		Cache = {};
 		if (!cur_archive.exists) {
 			InitDefaultMpqHeader(&cur_archive, &fhdr);
 		} else if (!ReadMPQHeader(&cur_archive, &fhdr)) {
 			goto on_error;
 		}
-		cur_archive.sgpBlockTbl = new _BLOCKENTRY[BlockEntrySize / sizeof(_BLOCKENTRY)];
-		std::memset(cur_archive.sgpBlockTbl, 0, BlockEntrySize);
+		cur_archive.blockTbl.resize(INDEX_ENTRIES);
```
ephphatha, line 665: This always uses a static size? Can blockTbl be made a std::array instead of a vector?

vlolteanu: It can be made a `std::array`, but if you want to move the tables quickly to/from the cache (rather than copying them), you'd have to go the extra step and make it a `std::unique_ptr<std::array>>`. This would mess up the syntax when accessing the tables.

vlolteanu:
Making this a draft for the moment. I've got a few more commits coming.

vlolteanu:
I've noticed problems w.r.t. error handling, or rather, lack thereof. Both `mpqapi_write_file` and `mpqapi_flush_and_close` can return false is case of error, but the result is always ignored. Both errors have something to do with writing stuff to the save file.

Should Diablo outright crash when something like this happens?

Source/mpqapi.cpp
glebm: Rather than a vector for a fixed-size array, use a `unique_ptr<T[]>`. It is specialised for square brackets access.

StephenCWills:
For one thing, the `FStreamWrapper` class uses a `CheckError()` function to log the errors when they occur so they're not entirely ignored. I imagine that in the worst case, a failure to write could corrupt a save file. That's not something that would necessarily prevent the player from continuing their game so ignoring it isn't going to cause any immediate problems.

On the other hand, the game doesn't produce any log files so the log messages would get lost in most cases. I believe a failure to write is typically an indication of either a full disk, a permissions issue, or a hardware failure. Such errors should be rare, should typically manifest early on, and often require the user to take action.

I'd say crashing is probably better than doing nothing, but given how unlikely these errors probably are, I don't feel very strongly one way or the other. Just my two cents.

Source/mpqapi.cpp
vlolteanu: What would the advantage of using `unique_ptr<T[]>` over `vector` be? We lose the ability to do range-based for loops.

glebm: A vector is not the appropriate type for a fixed-size container. While it offers the nicest syntax among the options in stdlib, that alone not a good reason to use the wrong abstraction. In this case, `unique_ptr<array>` is also a good option.

vlolteanu: I really don't see how `unique_ptr<T[]>` is superior to `vector` in this case, when expressing and enforcing the container size is the main issue:

- `vector` is resizable when it ideally shouldn't be.
- `unique_ptr<T[]>` has indeterminate size, and can be replaced with an array of a different size.

vlolteanu:
Avoiding corrupt saved games would, in essence, require atomic modifications to the MPQs. It would probably be done as follows:

- Create a backup before the first modification of an MPQ, and call `fsync` (I'm not sure if there's a portable version). Crash if the backup can't be made. (Newly-created MPQs are exempt from this.)
- Before closing a modified MPQ, call `fsync`. Delete the backup if successful. Crash if the backup can't be deleted.
- When attempting to open an MPQ with an existing backup, attempt to restore the backup and call `fsync`. Delete the backup if successful, crash otherwise.

Source/mpqapi.cpp
glebm: Sounds like `unique_ptr<array>` is the most appropriate type here, despite somewhat awkward access syntax

Issue #2707, opened 2021-08-26
ZeroTier gamelist UI


AJenbo:
**Describe**
Via ZeroTier it is possible to get a list of other running games. The API for this is already implemented, but the commit needs to be rebased as some of the relevant files have been moved and or altered sinc. Additionally the UI side of things is missing. The games should simply be added to the list found under `Multi Player`->`ZeroTier`->`Hero`, selecting one does the same as selecting `Join Game` and entering game name + password. Additionally `Join Game` should be renamed to `Join Private Game`

**Expected behavior**
If a player creates a game without entering a password it should appear on the list for others.

**Screenshots**
![image](https://user-images.githubusercontent.com/204594/130930899-013d88dd-f489-410d-b669-cd8b3adb54f0.png)

**Additional context**
The relevant commit for pulling a list of active games from ZeroTier: https://github.com/diasurgical/devilutionX/pull/2706/commits/b7eed948b0637147c7c33a4c8be54948e4c2c918

Add games here -> https://github.com/diasurgical/devilutionX/blob/master/Source/DiabloUI/selgame.cpp#L90

On a related note it would be good to fix the title so that it says ZeroTier instead of TCP, see: https://github.com/diasurgical/devilutionX/pull/2196

Issue #2641, opened 2021-08-19
Passwordless multiplayer games


vlolteanu:
In 99% of cases, TCP games entail either being in the same LAN or setting up a VPN. It seems to me that passwords are useless in most cases, yet they're mandated.

It would be nice to:

- be able to setup passwordless multiplayer games and
- not be prompted for a password when joining a passwordless multiplayer game.

AJenbo:
Apparently I have never played with 99% of people, which I suppose is believable...

I am willing to remove the password / encryption requirement, now that most OOB issues are fixed, if we validate the packages (hack that values are within the expected range).

It would go nicely with game lists using zerotier.

StephenCWills:
I'm not sure there's much to be gained by adding this. Do you simply dislike typing passwords?

vlolteanu:
I've played vanilla over modem a long time ago.

As for DevilutionX, I touched the multiplayer code on several occasions, and I had to test it somehow, so I ended up playing with myself (not in the lewd sense).

WRT playing in LANs/VPNs, unless there's the possibility of an asshole sibling/visitor joining and ruining your fun, passwords are just an annoyance.

vlolteanu:
> Apparently I have never played with 99% of people, which I suppose is believable...

P.S. I really like this kind of pushback against hyperbole :)

AJenbo:
> I'm not sure there's much to be gained by adding this. Do you simply dislike typing passwords?

Games with no passwords are requried in two cases:
- Discoverable games on ZeroTier
- joining games on devices with no keyboard (RG350, implementing out own controller keyboard is an alternative)

julealgon:
Another obvious improvement IMHO. Leaves it up to the host as to whether he wants his game to be completely public or not.

StephenCWills:
I still haven't tried playing with ZeroTier, but it doesn't look like the ZeroTier code uses the packet factory and, by extension, doesn't use libsodium or passwords. So I assumed this would only apply to TCP games. Does ZeroTier currently require a password?

The game's not completely public either way. You still have to share your IP/hostname (and port) so the password can be shared at the same time. AFAIK, this is literally only saving you the effort of typing the letter 'a' into the password prompt.

The same goes for RG350. Without the keyboard, there's no way to type the IP/hostname so removing the password requirement doesn't really help you in lieu of implementing a controller keyboard.

julealgon:
[...]

PR #2856, opened 2021-09-12
🐛 Prevent frame-perfect manipulation of spell being cast

### LINKED ISSUE #2844: Frame-perfect exploit when casting scrolls and staff charges
**Important information**

Vanilla issue. Happens on any DevilutionX build.

**Describe**
The issue described in #792 was caused by incorrectly using `pRSpell` as the variable to match the spells during scroll and staff charge removal. This meant a player could use the pre-hitframe frames in the spell animation to quickly swap from the selection he used before, to a different spell, which would then be used in the scroll/staff charge cleanup.

I've found a variation of that bug where, even with the correct `pSpell` being used, the player is _still_ able to influence the scroll/charge cleanup logic by _immediately_ swapping to a different spell while casting. Instead of having the entire window from cast to hit-frame as before, the window is now what appears to be a single frame, similar to how item dupping was exploited.

I'm currently not sure exactly what is creating this vulnerability but it is a different bug than #792, which is why I'm creating this new issue.

**To Reproduce**
1. Select scroll or staff charge as the readied spell
2. Right click to cast
3. _Immediately_ swap to a spell from a different source than the cast spell (this is important, see below)
4. The spell will be cast, but the scroll won't be removed/staff won't have its charges deducted

**Expected behavior**
The scroll should be removed/staff charge should be deducted according to the precise spell used for the cast and should not be influenced by the readied spell after-the-fact.

**Screenshots**
See #792 as the end result is the same.

**Additional context**
Discovered this while playing around with #792 after chatting with @soundchaser83 about it. Originally I believed the issue was related only to the window between cast and hit frame, but after experimenting more with it it became clear that we were seeing 2 distinct behaviors: I was having a much harder time reproing 792 with a scroll in my belt vs the same scenario with a scroll in my inventory.

From my experiments, what I believe is happening is the inverse of 792: I am somehow "casting a healing scroll", but switching to "firebolt" _before_ the cast takes place. So I get a healing cast from the scroll, ~~but mana is consumed as if I cast a firebolt and the scroll is not cleared~~ this is not correct: mana consumption is actually coming from the _healing spell_. Game is treating the scroll as if it was a learned spell.

I'm not sure if this is related to the fact that "casting" sends a network command, but "switching the readied spell" doesn't, meaning the switch happens instantaneously, while the cast is delayed via the messaging system somehow.

FAKE EDIT: I believe I now fully understand where the exploit is coming from. When the "cast" command is sent (in this case a "CMD_SPELLXY"), the tile coordinates and the spell ID are sent as arguments to the command, _but the spell_type **is not**_. When `CMD_SPELLXY` is processed, the "current spell type" is just read from the player casting the spell:

https://github.com/diasurgical/devilutionX/blob/ce40344ae9f9d457736114358036b707191ff4d3/Source/msg.cpp#L957-L978

->

https://github.com/diasurgical/devilutionX/blob/ce40344ae9f9d457736114358036b707191ff4d3/Source/msg.cpp#L970

This means it is possible to perform a cast and, before the command is processed, swap the readied spell to a spell of a different source (say, instead of a scroll, a spell from the spellbook), which _also changes the spell type on the player_, thus influencing the command. 

From there, the modified spell type is then used in other places, _including scroll/staff removal logic in `UseMana`_:

https://github.com/diasurgical/devilutionX/blob/940a120bf92f44c39229681444150f7fa1866c6b/Source/spells.cpp#L164-L184

This explains why this frame-perfect trick does not work when swapping among the same "category". So, trying to cast a Guardian but exploit the swap to use Firebolt mana will not work since both are of the "spell" category.

julealgon:
This PR changes all spell commands where the spell type was previously being inferred to explicitly pass this value at creation time.

This completely eliminates any possibility of manipulating the spell type between cast and execution and causing undefined behavior (like not consuming scrolls, staff changes, using incorrect mana values, etc).

Fixes #2844

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1096,17 +1096,18 @@ DWORD OnRangedAttackPlayer(TCmd *pCmd, Player &player)
 
 DWORD OnSpellMonster(TCmd *pCmd, Player &player)
 {
-	auto *p = (TCmdParam3 *)pCmd;
+	auto *p = (TCmdParam4 *)pCmd;
 
 	if (gbBufferMsgs != 1 && currlevel == player.plrlevel) {
 		auto spell = static_cast<spell_id>(p->wParam2);
+		auto spellType = static_cast<spell_type>(p->wParam3);
 		if (currlevel != 0 || spelldata[spell].sTownSpell) {
 			ClrPlrPath(player);
 			player.destAction = ACTION_SPELLMON;
 			player.destParam1 = p->wParam1;
-			player.destParam2 = p->wParam3;
+			player.destParam2 = p->wParam4;
 			player._pSpell = spell;
-			player._pSplType = player._pRSplType;
+			player._pSplType = spellType;
 			player._pSplFrom = 0;
 		} else {
 			PlayerMessageFormat(fmt::format(_("{:s} has cast an illegal spell."), player._pName).c_str());
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -2295,6 +2295,21 @@ void NetSendCmdParam3(bool bHiPri, _cmd_id bCmd, uint16_t wParam1, uint16_t wPar
 		NetSendLoPri(MyPlayerId, (byte *)&cmd, sizeof(cmd));
 }
 
+void NetSendCmdParam4(bool bHiPri, _cmd_id bCmd, uint16_t wParam1, uint16_t wParam2, uint16_t wParam3, uint16_t wParam4)
+{
+	TCmdParam4 cmd;
+
+	cmd.bCmd = bCmd;
+	cmd.wParam1 = wParam1;
+	cmd.wParam2 = wParam2;
+	cmd.wParam3 = wParam3;
+	cmd.wParam4 = wParam4;
+	if (bHiPri)
+		NetSendHiPri(MyPlayerId, (byte *)&cmd, sizeof(cmd));
+	else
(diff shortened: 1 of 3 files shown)


Source/msg.cpp
```diff
@@ -1094,17 +1096,18 @@ DWORD OnRangedAttackPlayer(TCmd *pCmd, Player &player)
 
 DWORD OnSpellMonster(TCmd *pCmd, Player &player)
 {
-	auto *p = (TCmdParam3 *)pCmd;
+	auto *p = (TCmdParam4 *)pCmd;
 
 	if (gbBufferMsgs != 1 && currlevel == player.plrlevel) {
 		auto spell = static_cast<spell_id>(p->wParam2);
+		auto spellType = static_cast<spell_type>(p->wParam3);
```
AJenbo, line 1103: I think pulling out the parameters at the start of the process like this gives us a good place to perform input validation in preparation for https://github.com/diasurgical/devilutionX/issues/2641 and https://github.com/diasurgical/devilutionX/issues/2707 :+1:

Source/msg.h
```diff
@@ -165,6 +165,16 @@ struct TCmdLocParam3 {
 	uint16_t wParam3;
 };
 
+struct TCmdLocParam4 {
```
AJenbo, line 168: Wow, are you sure the tubes will handle sending 11 bytes in a single message :o

AJenbo (approved):
Nice work, this feels like the very first step in making the netcode sane.

PR #2857, opened 2021-09-13
Make CreateFileStream return optional, rather than unique_ptr

vlolteanu:
It saves one memory allocation.

### DIFF
--- a/Source/mpqapi.cpp
+++ b/Source/mpqapi.cpp
@@ -96,12 +96,12 @@ public:
 
 	void Close()
 	{
-		s_ = nullptr;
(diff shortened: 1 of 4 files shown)


glebm:
Older GCC had a bug where stream wasn't moveable. Looks like it’s OK at least on CI.

vlolteanu:
Stuff doesn't have to be copyable/movable in C++17 for you to be able to return it. Are we using some older C++ standard for other platforms?

glebm:
Yeah, these older GCC versions aren’t fully cpp17

glebm (approved):

PR #2858, opened 2021-09-13
Move documentation from implementations to declarations.

Sid672:
Please check monster.cpp and monster.h because the file contains multiple functional documentation for a single function. So I add both in a single one.


Source/monster.cpp
```diff
@@ -1998,9 +1998,6 @@ bool IsTileSafe(const Monster &monster, Point position)
 	return true;
 }
 
-/**
- * @brief Check that the given tile is not currently blocked
- */
 bool IsTileAvailable(Point position)
```
ephphatha, line 2001: This is not the same function as `IsTileAvailable(Monster, Point)`. This function checks if a given position is free of monsters or players, then calls `IsTileWalkable(Point)` to check that the position is also free of objects and is a dungeon floor tile. It's declared here and not the header as this function isn't exposed to other cpp files.

Short story, this comment needs to be restored.

Source/monster.h
```diff
@@ -300,6 +300,11 @@ void PrintMonstHistory(int mt);
 void PrintUniqueHistory();
 void PlayEffect(Monster &monster, int mode);
 void MissToMonst(Missile &missile, Point position);
+
+/**
+ * @brief Check that the given tile is not currently blocked
+ * @brief Check that the given tile is available to the monster
```
ephphatha, line 306: There must be only one `@brief` annotation per doxygen comment. This is combining the comments for two different functions, you'll need to restore the `IsTileAvailable(Point)` comment and leave the `IsTileAvailable(Monster, Point)` version here.

Sid672: okay

AJenbo (approved):

Issue #2848, opened 2021-09-12
Chat doesn't wrap lines anymore


qndel:
![image](https://user-images.githubusercontent.com/14297035/132979862-c2c9e0b7-7c69-4cb8-ad16-f46419460577.png)

Broken in commit 7169882b1f3f12dc56e4a2d5ffa5e727f041fb40

PR #2862, opened 2021-09-13
Fix OOB error in DrawString

StephenCWills:
See https://github.com/diasurgical/devilutionX/pull/2791#discussion_r705485943

With #2848 having been merged, this fix has become relevant again. I wouldn't worry so much about it except there's a good chance translators will run into issues when trying different words and abbreviations to get things to fit. So here's the fix.


AJenbo (approved):
Thanks, that saved me a bit of time :)

PR #2866, opened 2021-09-14
Fix missiles don't hit horizontal walking players/monsters

obligaron:
## Content
In Master and Vanilla missiles can't hit players/monsters that walk horizontal.
This is cause while horizontal walking the player/monster _disappears_ from `dPlayers`/`dMonster`.
Both arrays contains what player/monster is located at a specific tile.
If a player/monster moves he is between two tiles and both are marked as occupied.
But only one location result in a missile hit. In the game engine this is represented as a positive Id (can be hit) and a negative number (only needed to mark the tile as occupied, so that no other player/monster can walk on this tile).

When the player/monster walks horizontal both tiles get a negative id, so you the player/monster can't be hit by a missile.
This PR changes this that also in horizontal walk one tile is used with a positive id (can be hit).

## Sample Video
This videos are created with extra long horizontal walking animations. This effect was only used for the video and is not part of this pr. 😉 
The shown Id's are presented by @qndel debug command `tiledata` #2847 🙂 

<details><summary>Master</summary>

https://user-images.githubusercontent.com/25415264/133337123-2e2ecc33-fa35-4b37-9571-288140870a9b.mp4

</details>

<details><summary>PR</summary>

https://user-images.githubusercontent.com/25415264/133337143-bc68d0ad-1a64-4433-966a-7cf02c49556f.mp4

</details>

## Notes

- This PR includes a breaking change that only affects a saved currently active horizontal walking and will go away after the walking animation is finished.
   - From vanilla to PR: The player/monster will not be drawn, cause the special logic for drawing is removed.
   - From PR to vanilla: The player/monster will be drawn at incorrect location.
- This PR affects game balance and PvP 😉 
- BFLAG_PLAYERLR and BFLAG_MONSTLR was used for [PlrDirOK](https://github.com/diasurgical/devilutionX/compare/master...obligaron:moveside?expand=1#diff-71461699ae2f94f652e6cba2136d3c696151ba3a6c0e357eaf55438a5c212ad3L334-L340) and [(Monster)DirOK](https://github.com/diasurgical/devilutionX/compare/master...obligaron:moveside?expand=1#diff-d46585089d315b53630b1035442b3d490d7b422cb05ceee76cf9041de3e39179L4403-L4408). I removed this specials cause I didn't hit them and they seem suspicious. Cause monster don't check horizontal walking players and vice versa. But perhaps I miss something. So please take a careful look.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -416,7 +416,7 @@ void LoadPlayer(LoadHelper &file, Player &player)
 	player.position.temp.y = file.NextLE<int32_t>();
 	player.tempDirection = static_cast<Direction>(file.NextLE<int32_t>());
 	player._pVar4 = file.NextLE<int32_t>();
-	player._pVar5 = file.NextLE<int32_t>();
+	file.Skip(4); // skip _pVar5, was used for storing position of a tile which should have its BFLAG_PLAYERLR flag removed after walking
 	player.position.offset2.deltaX = file.NextLE<int32_t>();
 	player.position.offset2.deltaY = file.NextLE<int32_t>();
 	file.Skip(4); // Skip actionFrame
@@ -1091,7 +1091,7 @@ void SavePlayer(SaveHelper &file, const Player &player)
 	file.WriteLE<int32_t>(player.position.temp.y);
 	file.WriteLE<int32_t>(static_cast<int32_t>(player.tempDirection));
 	file.WriteLE<int32_t>(player._pVar4);
-	file.WriteLE<int32_t>(player._pVar5);
+	file.Skip<int32_t>(); // skip _pVar5, was used for storing position of a tile which should have its BFLAG_PLAYERLR flag removed after walking
 	file.WriteLE<int32_t>(player.position.offset2.deltaX);
 	file.WriteLE<int32_t>(player.position.offset2.deltaY);
 	file.Skip<int32_t>(); // Skip _pVar8
@@ -1772,7 +1772,7 @@ void LoadGame(bool firstflag)
 	}
 	for (int j = 0; j < MAXDUNY; j++) {
 		for (int i = 0; i < MAXDUNX; i++) // NOLINT(modernize-loop-convert)
-			dFlags[i][j] = file.NextLE<int8_t>();
+			dFlags[i][j] = file.NextLE<int8_t>() & ~BFLAG_PLAYERLR;
 	}
 	for (int j = 0; j < MAXDUNY; j++) {
 		for (int i = 0; i < MAXDUNX; i++) // NOLINT(modernize-loop-convert)
@@ -2159,7 +2159,7 @@ void LoadLevel()
 
 	for (int j = 0; j < MAXDUNY; j++) {
 		for (int i = 0; i < MAXDUNX; i++) // NOLINT(modernize-loop-convert)
-			dFlags[i][j] = file.NextLE<int8_t>();
+			dFlags[i][j] = file.NextLE<int8_t>() & ~BFLAG_PLAYERLR;
 	}
 	for (int j = 0; j < MAXDUNY; j++) {
 		for (int i = 0; i < MAXDUNX; i++) // NOLINT(modernize-loop-convert)
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -415,7 +415,7 @@ void LoadPlayer(LoadHelper &file, Player &player)
 	player.position.temp.x = file.NextLE<int32_t>();
 	player.position.temp.y = file.NextLE<int32_t>();
 	player.tempDirection = static_cast<Direction>(file.NextLE<int32_t>());
-	player._pVar4 = file.NextLE<int32_t>();
+	player.spellLevel = file.NextLE<int32_t>();
 	file.Skip(4); // skip _pVar5, was used for storing position of a tile which should have its BFLAG_PLAYERLR flag removed after walking
 	player.position.offset2.deltaX = file.NextLE<int32_t>();
 	player.position.offset2.deltaY = file.NextLE<int32_t>();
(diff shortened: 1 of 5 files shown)


Source/loadsave.cpp
```diff
@@ -1090,8 +1090,8 @@ void SavePlayer(SaveHelper &file, const Player &player)
 	file.WriteLE<int32_t>(player.position.temp.x);
 	file.WriteLE<int32_t>(player.position.temp.y);
 	file.WriteLE<int32_t>(static_cast<int32_t>(player.tempDirection));
-	file.WriteLE<int32_t>(player._pVar4);
-	file.WriteLE<int32_t>(player._pVar5);
+	file.WriteLE<int32_t>(player.spellLevel);
+	file.WriteLE<int32_t>(0); // skip _pVar5, was used for storing position of a tile which should have its BFLAG_PLAYERLR flag removed after walking
```
AJenbo, line 1094:
```suggestion
	file.Skip<int32_t>(); // skip _pVar5, was used for storing position of a tile which should have its BFLAG_PLAYERLR flag removed after walking
```

Source/scrollrt.cpp
```diff
@@ -1416,7 +1404,7 @@ Displacement GetOffsetForWalking(const AnimationInfo &animationInfo, const Direc
 {
 	// clang-format off
 	//                                           South,        SouthWest,    West,         NorthWest,    North,        NorthEast,     East,         SouthEast,
-	constexpr Displacement StartOffset[8]    = { {   0, -32 }, {  32, -16 }, {  32, -16 }, {   0,   0 }, {   0,   0 }, {  0,    0 },  { -32, -16 }, { -32, -16 } };
+	constexpr Displacement StartOffset[8]    = { {   0, -32 }, {  32, -16 }, {  64,   0 }, {   0,   0 }, {   0,   0 }, {  0,    0 },  { -64,   0 }, { -32, -16 } };
 	constexpr Displacement MovingOffset[8]   = { {   0,  32 }, { -32,  16 }, { -64,   0 }, { -32, -16 }, {   0, -32 }, {  32, -16 },  {  64,   0 }, {  32,  16 } };
```
AJenbo, line 1408: Feels like a really good sign that the offers are not symmetrical.

Seams this function could now be reduced to something like:
```cpp
Displacement GetOffsetForWalking(const AnimationInfo &animationInfo, const Direction dir, bool cameraMode /*= false*/)
{
	//                                         South,     SouthWest,   West,       NorthWest,    North,      NorthEast,   East,      SouthEast,
	constexpr Displacement MovingOffset[8] = { { 0, 32 }, { -32, 16 }, { -64, 0 }, { -32, -16 }, { 0, -32 }, { 32, -16 }, { 64, 0 }, { 32, 16 } };

	float fAnimationProgress = animationInfo.GetAnimationProgress();

	if (cameraMode) {
		fAnimationProgress = -fAnimationProgress;
	} else {
		fAnimationProgress -= 1.F;
	}

	return MovingOffset[static_cast<size_t>(dir)] * fAnimationProgress;
}
```

Source/loadsave.cpp
obligaron: Fixed. Thanks 🙂

Source/scrollrt.cpp
obligaron: I tested the code and it doesn't work with NorthWest, North and NorthEast.
The problem is that in these directions the logic which tile has the positive or negative in `dPlayer`/`dMonster` is different.
Normally the starting tile is negative and the target tile is positive.
But in NorthWest, North and NorthEast the logic is reversed. 😕 
I think this could be unified, but I also think this is outside of the scope of this pr cause it also affects/alters missile hit logic (another tile would be hit tested).

NiteKat: Just wondering, could this difference in behavior be related to how it's easier to kite monsters in certain directions? Like when moving southward, if you get hit stunned, you are moved to the destination tile automatically, but if you are moving northward and get hit stunned, you move back to the source tile?

qndel:
Could you leave renaming _pVar4 into spellLevel for a separate PR? Then it'd be easier to see changes only related to horizontal walk :) Great job as usual!

Source/scrollrt.cpp
qndel: yes, that's another thing to fix after horizontal walk

obligaron:
> 
> 
> Could you leave renaming _pVar4 into spellLevel for a separate PR? Then it'd be easier to see changes only related to horizontal walk :) Great job as usual!

Thanks 🙂 

I tried to separate rename and fix in the commits.
The main change/fix is the first commit. The renaming of `_pVar4` to `spellLevel` was done in a separate commit.
Do you think that's enough?

Source/player.cpp
```diff
@@ -730,8 +726,8 @@ bool DoWalk(int pnum, int variant)
 			break;
 		case PM_WALK3:
 			dPlayer[player.position.tile.x][player.position.tile.y] = 0;
-			dFlags[player._pVar4][player._pVar5] &= ~BFLAG_PLAYERLR;
 			player.position.tile = player.position.temp;
+			// Don't delete this (in devilutionX) unncessary assignment. It's needed to stay compatible with vanilla saves.
```
AJenbo, line 730:
```suggestion
			// dPlayer is set here for backwards compatability, without it the player would be invisible is save is loaded in older version
```

Source/monster.cpp
```diff
@@ -1356,7 +1355,7 @@ bool MonsterWalk(int i, MonsterMode variant)
 		case MonsterMode::MoveSideways:
 			dMonster[monster.position.tile.x][monster.position.tile.y] = 0;
 			monster.position.tile = { monster._mVar1, monster._mVar2 };
-			dFlags[monster.position.temp.x][monster.position.temp.y] &= ~BFLAG_MONSTLR;
+			// Don't delete this (in devilutionX) unncessary assignment. It's needed to stay compatible with vanilla saves.
```
AJenbo, line 1358: I feel the comment would be better if it described what the consequences is

```suggestion
			// dMonster is set here for backwards comparability, without it the monster would be invisible is save is loaded in older version
```

AJenbo (changes requested):
All looks good, except the comments, did i get it right?

Source/player.cpp
qndel: Please edit `comparability` to `compatibility`

obligaron:
> All looks good, except the comments, did i get it right?

Not 100% I think.
The problem is without these lines, you will get invisible players/monsters in devilutionX.
Cause devilutionX wouldn't set 1 in `dPlayer`/`dMonster`.
But I tried to clarify the comment, take a look. 🙂

PR #2867, opened 2021-09-14
Fix existing missiles check for Guardian/Townportal/Rune

### LINKED ISSUE #2864: Guardian's positioning no longer works properly
Commit 7f6e171b4b51e832e3fbbfb1bee46508d24ed212 has introduced that guardians are always created on the same position (tile). This means that if you do not move the mouse and cast some guardians, only one guardian appears. In vanilla the guardians are always positioned on a tile next to the neighbor. The issue seems also to be similar to #2833.

To reproduce cast guardians to the same position with the attached game.
[guardians.zip](https://github.com/diasurgical/devilutionX/files/7163838/guardians.zip)

obligaron:
Fixes #2864

Was caused by a copy&past error in 7f6e171b4b51e832e3fbbfb1bee46508d24ed212.
Instead of using target x/y the crawling loop variables were used.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -712,7 +712,7 @@ void AddRune(Missile &missile, Point dst, missile_id missileID)
 					continue;
 
 				int dp = dPiece[target.x][target.y];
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Ouch, thanks :)

PR #2873, opened 2021-09-16
LoadFile: Avoid opening the file twice

glebm:
Previously we were passing the path to `GetFileSize`, which resulted in opening and closing the file twice.


Source/player.cpp
```diff
@@ -398,12 +398,14 @@ void StartWalk(int pnum, Displacement vel, Direction dir, bool pmWillBeCalled)
...
 {
 	data = nullptr;
+#ifndef RUN_TESTS
 	data = LoadFileInMem(path);
 
 	for (int i = 0; i < 8; i++) {
 		byte *pCelStart = CelGetFrame(data.get(), i);
 		anim[i].emplace(pCelStart, width);
 	}
+#endif
```
AJenbo, line 408: Not a fan of test exceptions, what happens currently with this, and does this PR change that?

glebm: This only worked before by accident, I think.
The LoadFileInMem function previously returned an empty array in test mode, which was also invalid here because of `GetCelFrame`.

AJenbo: We should probably see if we can run the test with the shareware data, ill try to make a mental note of this.

AJenbo (approved):

PR #2878, opened 2021-09-16
fix saves bricked by spawn command

qndel:
If you used "spawn" command and spawned a monster type that wasn't originally on the level, saving with this monster wouldn't let you load after restarting the game basically bricking your save. With this "fix" all debug monsters change into zombies anyway, but it's just to allow loading the save at all.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4524,6 +4524,13 @@ bool LineClear(const std::function<bool(Point)> &clear, Point startPoint, Point 
 void SyncMonsterAnim(Monster &monster)
 {
 	monster.MType = &LevelMonsterTypes[monster._mMTidx];
+#ifdef _DEBUG
+	// fix for saves with debug monsters having type originally not on the level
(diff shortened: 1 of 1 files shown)


julealgon:
> With this "fix" all debug monsters change into zombies anyway

Why? Can't we honor the original monster while loading?

qndel:
> > With this "fix" all debug monsters change into zombies anyway
> 
> Why? Can't we honor the original monster while loading?

no clue, feel free to investigate, outside of scope of this PR, this only unbricks

Source/monster.cpp
```diff
@@ -4524,6 +4524,11 @@ bool LineClear(const std::function<bool(Point)> &clear, Point startPoint, Point
 void SyncMonsterAnim(Monster &monster)
 {
 	monster.MType = &LevelMonsterTypes[monster._mMTidx];
+	// fix for saves with debug monsters having type originally not on the level
```
AJenbo, line 4527: if def debug?

AJenbo (approved):

Issue #2463, opened 2021-07-27
Charging monsters ends up in an invalid state on level change


Chance4us:
Unfortunately, I have to report a worrying bug this time because I didn't really know what to look for by bisecting.
As for the history, I can tell that I killed the monsters in the affected level with the Guardians.
When I reentered the level again sometime I noticed the following.

https://user-images.githubusercontent.com/45902449/127213986-8d0d77aa-439c-4052-895c-93c7305ab259.mp4

It is namely a "Frost Charger" that should actually be dead, but information about this is displayed as if he were still alive.
I now hope that one of you has some idea how to find the bug.

Savegame:
[frost_charger.zip](https://github.com/diasurgical/devilutionX/files/6888084/frost_charger.zip)

By the way, is there actually a possibility that this is related to #2441 or #2429 and that other types of monsters could also be affected?

AJenbo:
@obligaron do you think this could be related to the issue we had with MI_Rhino recently?

obligaron:
Yeah it looks kind of similar to #2310
#2310 doesn't fix corrupted game state.
So the question is when / with what version was the level cleared?

And it could also be something different, that we didn't discovered yet. 🤔

Chance4us:
> So the question is when / with what version was the level cleared?

It was played with the following slightly different version.
https://github.com/Chance4us/devilutionX/tree/testing

AJenbo:
Specifically https://github.com/Chance4us/devilutionX/commit/464ba0aaeff5b37ee8bb96f5c7f0f8c5ab0bc7c6 ?

NiteKat:
Does Telekenesis fix the monster?

Chance4us:
> Specifically [Chance4us@464ba0a](https://github.com/Chance4us/devilutionX/commit/464ba0aaeff5b37ee8bb96f5c7f0f8c5ab0bc7c6) ?

It is @qndel's object highlighting with resolved merge conflicts.
https://github.com/diasurgical/devilutionX/pull/2160

Chance4us:
> Does Telekenesis fix the monster?

Unfortunately, I couldn't try it because the hero lacks this spell.

FitzRoyX:
Why one and not the others? Was it the only one to die mid-charge or something?

Chance4us:
As I looked at #2310, unpredictable bugs often arise from cleanups or code style changes that are actually expected to improve.
For example for expected performance improvements or for greater clarity of the code.

[...]

PR #2879, opened 2021-09-16
fix stuck chargers

qndel:
Fixes stuck chargers when you enter the level. In multiplayer it triggers only when you are the only person on the level.
Resolves https://github.com/diasurgical/devilutionX/issues/2463

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1140,14 +1140,14 @@ void CreateLevel(lvl_entry lvldir)
 void UnstuckChargers()
 {
 	if (gbIsMultiplayer) {
-		for (int i = 0; i < MAX_PLRS; i++) {
-			if (!Players[i].plractive)
+		for (auto &player : Players) {
+			if (!player.plractive)
 				continue;
-			if (Players[i]._pLvlChanging)
+			if (player._pLvlChanging)
 				continue;
-			if (Players[i].plrlevel != MyPlayer->plrlevel)
+			if (player.plrlevel != MyPlayer->plrlevel)
 				continue;
-			if (i == MyPlayerId)
+			if (&player == MyPlayer)
 				continue;
 			return;
 		}
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1137,6 +1137,28 @@ void CreateLevel(lvl_entry lvldir)
 	}
 }
 
+void UnstuckChargers()
+{
+	if (gbIsMultiplayer) {
+		for (int i = 0; i < MAX_PLRS; i++) {
+			if (!Players[i].plractive)
+				continue;
+			if (Players[i]._pLvlChanging)
+				continue;
+			if (Players[i].plrlevel != MyPlayer->plrlevel)
+				continue;
+			if (i == MyPlayerId)
+				continue;
+			return;
+		}
+	}
+	for (int i = 0; i < ActiveMonsterCount; i++) {
+		auto &monster = Monsters[ActiveMonsters[i]];
+		if (monster._mmode == MonsterMode::Charge)
+			monster._mmode = MonsterMode::Stand;
+	}
+}
+
 void UpdateMonsterLights()
 {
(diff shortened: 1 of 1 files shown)


Source/diablo.cpp
```diff
@@ -1137,6 +1137,28 @@ void CreateLevel(lvl_entry lvldir)
 	}
 }
 
+void UnstuckChargers()
```
julealgon, line 1140: I look at this method name and I already don't like it. Looks like a crazy hack to me.

obligaron: I think at least it's a short term solution.
It also fixes old vanilla saves or saves from previous deviltuionX versions.

Long term there could be other solutions like remodeling how the charging works (avoid missiles for it) or also save/load missiles when changing levels or ... but that's more complex and to stay savegame compatible with vanilla is not easy.

So I think this pr is currently fine. 🙂

Source/diablo.cpp
```diff
@@ -1137,6 +1137,28 @@ void CreateLevel(lvl_entry lvldir)
...
+{
+	if (gbIsMultiplayer) {
+		for (int i = 0; i < MAX_PLRS; i++) {
+			if (!Players[i].plractive)
+				continue;
+			if (Players[i]._pLvlChanging)
+				continue;
+			if (Players[i].plrlevel != MyPlayer->plrlevel)
+				continue;
+			if (i == MyPlayerId)
```
AJenbo, line 1150:
```suggestion
		for (auto &player : Players) {
			if (!player.plractive)
				continue;
			if (player._pLvlChanging)
				continue;
			if (player.plrlevel != MyPlayer->plrlevel)
				continue;
			if (&player == MyPlayer)
```

AJenbo (approved):

Source/diablo.cpp
julealgon: I hope it doesn't become one of those "multi-year short-term solutions" that I'm so used to seeing everywhere.

AJenbo: That probably depends on how many/what other unsolved issues are left in a year.

obligaron: > I hope it doesn't become one of those "multi-year short-term solutions" that I'm so used to seeing everywhere.

I think that's possible. 😉 
But this PR doesn't span across the codebase or hinders future development or something like this.

julealgon: > But this PR doesn't span across the codebase or hinders future development or something like this.

It doesn't need to. If we keep introducing hacks to the codebase, even if they are all very localized and don't impact other things, there will reach a point where the code becomes a mess. These things pile up over time.

Anyways, I'll stop ranting.

AJenbo: Better fixes are welcome, until that happens we have to be pragmatic

julealgon: > Better fixes are welcome, until that happens we have to be pragmatic

You make it sound like this is critical enough to warrant a hack, which is definitely not the case (IMHO). It is an extremely rare vanilla bug.

In my view, it would be more productive in the long term to pursue proper solutions instead of settling for hacks.

AJenbo: I saw it happen just yesterday (after this pr was created) on souncaser83's stream, he didn't notice him self as he was looking elsewhere. I think it might be more common then we would think.

julealgon: > I think it might be more common then we would think.

I can see that being the case in speedruns where you are moving up/down stairs a lot with monsters around you. Maybe I didn't see it as much since I play more "casually", clearing floors and whatnot.

Even if that's the case, it is still largely inconsequential.

My overall point was that hacks should be weighed against how critical it is the thing that they are hacking around. If not critical enough, we should consider not applying the hack and living with the minor issue for a bit longer until a decent fix is made.

AJenbo: I don't think we disagree much here, we just don't agree on what weights to insert in to the Bayesian model.

PR #2882, opened 2021-09-16
Add new debug commands, remove old hotkeys and update debug.md

obligaron:
## Content

- Remove `-m` debug command line parameter. Was used for spawning monsters but now we have `spawn` debug command. 😉 
- Remove `-^` debug command line parameter. Was only used for scrolling with shift. Now on per default in debug (hope that's okay).
- Rename `debug_mode_key_i` to `DebugDisableNetworkTimeout`.
- Introduce `iteminfo` debug command to show item infos. Replaces `ItemInfo` keybinding and now also works for inventory and held items.
- Introduce `questinfo` debug command that replaces `PrintDebugQuest` keybinding and now allows to specify what quest to show instead of needing to iterate all previous quests.
- Remove `-t` debug hotkey. Was used for showing tile related info but now we have `tiledata` debug command. 😉 
- Introduce `playerinfo` debug command to show player infos. Replaces `d` hotke. Thios also allows now to specify what player to show instead of needing to iterate all previous players (even the not active ones).
- Reworked `debug.md` to match current debug capacities.
- More code in `debug.cpp` and less debug related code elsewhere. 🙂

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -34,6 +34,7 @@ bool DebugGodMode = false;
 bool DebugVision = false;
 bool DebugGrid = false;
 std::unordered_map<int, Point> DebugCoordsMap;
+bool DebugScrollViewEnabled = false;
 
 namespace {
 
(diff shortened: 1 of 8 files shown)


AJenbo:
> scrolling with shift
could you add a command to toggle, its really annoying when on :)

obligaron:
> could you add a command to toggle, its really annoying when on :)

Done 🙂 

Out of curiosity: do you often hold shift while ingame?

AJenbo:
> Out of curiosity: do you often hold shift while ingame?

Shift attack at the edge of the screen is pretty common as an archer

AJenbo (approved):

obligaron:
> 
> 
> > Out of curiosity: do you often hold shift while ingame?
> 
> Shift attack at the edge of the screen is pretty common as an archer

How could i forget... I think i should go to bed. :upside_down_face:

Issue #2586, opened 2021-08-12
Make custom iterators for activ entity lists


AJenbo:
There are a lot of loops that we are not able to convert to ranges as they only loop over the active entities rather then all entities.

```cpp
	for (int i = 0; i < ActiveMissileCount; i++) {
		int mx = ActiveMissiles[i];
		if (Missiles[mx]._mitype == MIS_TOWN && missile._misource == id)
			Missiles[mx]._mirange = 0;
	}
```

This also means that we will still be indexing in to the global array at least once.

To solve this we can create custom iterators:

```cpp
	for (auto missile : ActiveMissileIterator) {
		if (missile._mitype == MIS_TOWN && missile._misource == id)
			missile._mirange = 0;
	}
```

https://internalpointers.com/post/writing-custom-iterators-modern-cpp

Some times it may also be useful with iterators that can take a custom range:

```cpp
	int start = gbIsMultiplayer ? 3 : 2;
	int end = gbIsHellfire ? 10 : 8;
	for (auto healitem : ItemIterator(&Healitems[start], &Healitems[end])) {
		healitem._iCreateInfo = lvl | CF_HEALER;
		healitem._iIdentified = true;
		healitem._iStatFlag = StoreStatOk(&healitem[i]);
	}
```

This issue builds on the changes done in https://github.com/diasurgical/devilutionX/issues/2435 but does not strictly require it.

AJenbo:
As long as it doesn't break multiplayer, savegames or performance that sounds good :)

ephphatha:
I had a bit of a play with using STL containers to wrap these collections and it looks like the default concepts aren't appropriate due to various constraints implied (or enforced in C++20).

std::vector and std::array both imply elements will be stored in contiguous memory regions and C++20 actually enforces this by requiring their iterators to implement the `contiguous_iterator` concept. This gets broken by the way diablo/devilutionx uses a backing buffer to avoid moving objects around in memory.

std::deque is close, but that container is geared more towards use like a stack/queue instead of the type of use in diablo/devilutionx (always appending, removal from anywhere). Plus I don't think we can force it to use a single fixed backing array or control how it allocates and deallocates memory.

I've started implementing an extremely restricted vector-like container to see if that'll work for these use cases, but happy for someone else to have a go at it too since it'll probably take me a while.

AJenbo:
see https://github.com/diasurgical/devilutionX/blob/master/Source/inv_iterators.hpp for examples of how to implement this

PR #2892, opened 2021-09-18
Refactor: Use item iterators for scroll checks

glebm:
Refs #2586

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -5,6 +5,7 @@
  */
 #include "control.h"
 
+#include <algorithm>
 #include <array>
 #include <cstddef>
 
@@ -24,6 +25,7 @@
 #include "gamemenu.h"
 #include "init.h"
 #include "inv.h"
+#include "inv_iterators.hpp"
 #include "lighting.h"
 #include "minitext.h"
 #include "missiles.h"
@@ -688,6 +690,7 @@ void DrawSpellList(const Surface &out)
 	auto &myPlayer = Players[MyPlayerId];
 
 	for (auto &spellListItem : GetSpellListItems()) {
+		const spell_id spellId = spellListItem.id;
 		spell_type transType = spellListItem.type;
 		int spellLevel = 0;
 		const SpellData &spellDataItem = spelldata[static_cast<size_t>(spellListItem.id)];
@@ -701,7 +704,7 @@ void DrawSpellList(const Surface &out)
 		}
 
 		SetSpellTrans(transType);
-		DrawSpellCel(out, spellListItem.location, *pSpellCels, SpellITbl[static_cast<size_t>(spellListItem.id)]);
+		DrawSpellCel(out, spellListItem.location, *pSpellCels, SpellITbl[static_cast<size_t>(spellId)]);
 
 		if (!spellListItem.isSelected)
 			continue;
@@ -714,7 +717,7 @@ void DrawSpellList(const Surface &out)
 		case RSPLTYPE_SPELL:
 			DrawSpellCel(out, spellListItem.location, *pSpellCels, SPLICONLAST + 4);
 			strcpy(infostr, fmt::format(_("{:s} Spell"), pgettext("spell", spellDataItem.sNameText)).c_str());
-			if (spellListItem.id == SPL_HBOLT) {
+			if (spellId == SPL_HBOLT) {
 				strcpy(tempstr, _("Damages undead only"));
 				AddPanelString(tempstr);
 			}
@@ -727,23 +730,11 @@ void DrawSpellList(const Surface &out)
 		case RSPLTYPE_SCROLL: {
 			DrawSpellCel(out, spellListItem.location, *pSpellCels, SPLICONLAST + 1);
 			strcpy(infostr, fmt::format(_("Scroll of {:s}"), pgettext("spell", spellDataItem.sNameText)).c_str());
-			int v = 0;
-			for (int t = 0; t < myPlayer._pNumInv; t++) {
-				if (!myPlayer.InvList[t].isEmpty()
-				    && (myPlayer.InvList[t]._iMiscId == IMISC_SCROLL || myPlayer.InvList[t]._iMiscId == IMISC_SCROLLT)
-				    && myPlayer.InvList[t]._iSpell == spellListItem.id) {
-					v++;
-				}
-			}
-			for (auto &item : myPlayer.SpdList) {
-				if (!item.isEmpty()
-				    && (item._iMiscId == IMISC_SCROLL || item._iMiscId == IMISC_SCROLLT)
-				    && item._iSpell == spellListItem.id) {
-					v++;
-				}
-			}
-			strcpy(tempstr, fmt::format(ngettext("{:d} Scroll", "{:d} Scrolls", v), v).c_str());
-			AddPanelString(tempstr);
+			const InventoryAndBeltPlayerItemsRange items { myPlayer };
+			const int scrollCount = std::count_if(items.begin(), items.end(), [spellId](const Item &item) {
+				return item.IsScrollOf(spellId);
+			});
+			strcpy(tempstr, fmt::format(ngettext("{:d} Scroll", "{:d} Scrolls", scrollCount), scrollCount).c_str());
 		} break;
 		case RSPLTYPE_CHARGES: {
 			DrawSpellCel(out, spellListItem.location, *pSpellCels, SPLICONLAST + 2);
@@ -756,7 +747,7 @@ void DrawSpellList(const Surface &out)
 			break;
 		}
 		for (int t = 0; t < 4; t++) {
-			if (myPlayer._pSplHotKey[t] == spellListItem.id && myPlayer._pSplTHotKey[t] == spellListItem.type) {
+			if (myPlayer._pSplHotKey[t] == spellId && myPlayer._pSplTHotKey[t] == spellListItem.type) {
 				auto hotkeyName = keymapper.KeyNameForAction(quickSpellActionIndexes[t]);
 				PrintSBookHotkey(out, spellListItem.location, hotkeyName);
 				strcpy(tempstr, fmt::format(_("Spell Hotkey {:s}"), hotkeyName.c_str()).c_str());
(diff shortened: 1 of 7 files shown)


Source/control.cpp
```diff
@@ -6,6 +6,7 @@
 #include "control.h"
 
 #include <array>
+#include <algorithm>
```
julealgon, line 9: "al" comes before "ar"

Source/control.cpp
```diff
@@ -684,6 +686,7 @@ void DrawSpellList(const Surface &out)
 	auto &myPlayer = Players[MyPlayerId];
 
 	for (auto &spellListItem : GetSpellListItems()) {
+		const spell_id spell = spellListItem.id;
```
julealgon, line 689: Can we rename the variable to `spellId`?

Source/control.cpp
```diff
@@ -723,22 +726,11 @@ void DrawSpellList(const Surface &out)
...
-				    && item._iSpell == spellListItem.id) {
-					v++;
-				}
-			}
-			strcpy(tempstr, fmt::format(ngettext("{:d} Scroll", "{:d} Scrolls", v), v).c_str());
+			const InventoryAndBeltPlayerItemsRange items { myPlayer };
+			const int count = std::count_if(items.begin(), items.end(), [spell](const Item &item) {
+				return item.IsScrollOf(spell);
+			});
+			strcpy(tempstr, fmt::format(ngettext("{:d} Scroll", "{:d} Scrolls", count), count).c_str());
```
julealgon, line 733: Really nice cleanup.

Source/control.cpp
```diff
@@ -1180,22 +1172,11 @@ void CheckPanelInfo()
...
-					if (!item.isEmpty()
-					    && (item._iMiscId == IMISC_SCROLL || item._iMiscId == IMISC_SCROLLT)
-					    && item._iSpell == v) {
-						s++;
-					}
-				}
-				strcpy(tempstr, fmt::format(ngettext("{:d} Scroll", "{:d} Scrolls", s), s).c_str());
+				const InventoryAndBeltPlayerItemsRange items { myPlayer };
+				const int count = std::count_if(items.begin(), items.end(), [v](const Item &item) {
+					return item.IsScrollOf(v);
```
julealgon, line 1177: Can we rename `v` to `spellId` as well so that the changes in this PR are consistent and readable?

Source/control.cpp
```diff
@@ -723,22 +726,11 @@ void DrawSpellList(const Surface &out)
...
-			for (auto &item : myPlayer.SpdList) {
-				if (!item.isEmpty()
-				    && (item._iMiscId == IMISC_SCROLL || item._iMiscId == IMISC_SCROLLT)
-				    && item._iSpell == spellListItem.id) {
-					v++;
-				}
-			}
-			strcpy(tempstr, fmt::format(ngettext("{:d} Scroll", "{:d} Scrolls", v), v).c_str());
+			const InventoryAndBeltPlayerItemsRange items { myPlayer };
+			const int count = std::count_if(items.begin(), items.end(), [spell](const Item &item) {
```
julealgon, line 730: `count` is a bit generic for a name. Can we rename to something more semantic, like `scrollCount` maybe?

Source/inv.cpp
```diff
@@ -1932,27 +1933,16 @@ bool UseScroll()
 	if (pcurs != CURSOR_HAND)
 		return false;
 
-	auto &myPlayer = Players[MyPlayerId];
+	Player &myPlayer = Players[MyPlayerId];
```
julealgon, line 1936: Why did you change from `auto` to `Player` here?

Source/inv_iterators.hpp
```diff
@@ -19,7 +19,7 @@ class ItemsContainerRange {
 	class Iterator {
 	public:
 		using iterator_category = std::forward_iterator_tag;
-		using difference_type = void;
+		using difference_type = int;
```
julealgon, line 22: Can you elaborate why this is being changed? Was it wrong? Is it needed for the new iterator for some reason?

Source/inv_iterators.hpp
```diff
@@ -261,6 +261,37 @@ class BeltPlayerItemsRange {
...
+		});
+	}
+
+	[[nodiscard]] ItemsContainerListRange::Iterator end() const // NOLINT(readability-identifier-naming)
+	{
+		auto result = ItemsContainerListRange::Iterator({
+		    InventoryPlayerItemsRange(*player_).end(),
+		    BeltPlayerItemsRange(*player_).end(),
+		});
+		return result;
```
julealgon, line 288: Can't we just return the iterator directly like we did in line 276? Would be good if for consistency alone.

Source/inv_iterators.hpp
```diff
@@ -261,6 +261,37 @@ class BeltPlayerItemsRange {
...
+	}
+
+private:
+	Player *player_;
```
julealgon, line 292: Why does the field name end in an `_`? Shouldn't the underscore be _before_ the name to represent private data instead of _after_ it?

Source/items.h
```diff
@@ -357,6 +357,16 @@ struct Item {
 		}
 	}
 
+	[[nodiscard]] bool IsScroll() const
```
julealgon, line 360: Do we really need this `[[nodiscard]]`? What does it do?

Source/items.h
```diff
@@ -357,6 +357,16 @@ struct Item {
 		}
 	}
 
+	[[nodiscard]] bool IsScroll() const
+	{
+		return _iMiscId == IMISC_SCROLL || _iMiscId == IMISC_SCROLLT;
```
julealgon, line 362: Please use the `IsAnyOf` method.

Source/items.h
```diff
@@ -357,6 +357,16 @@ struct Item {
 		}
 	}
 
+	[[nodiscard]] bool IsScroll() const
+	{
+		return _iMiscId == IMISC_SCROLL || _iMiscId == IMISC_SCROLLT;
+	}
+
+	[[nodiscard]] bool IsScrollOf(spell_id spell) const
```
julealgon, line 365: Again would recommend renaming to `spellId`.

Source/player.cpp
```diff
@@ -1973,17 +1974,9 @@ void CheckCheatStats(Player &player)
...
-		}
-	}
-
-	for (auto &item : SpdList) {
-		if (!item.isEmpty() && (item._iMiscId == IMISC_SCROLL || item._iMiscId == IMISC_SCROLLT)) {
-			if (item._iStatFlag)
-				_pScrlSpells |= GetSpellBitmask(item._iSpell);
+	for (Item &item : InventoryAndBeltPlayerItemsRange { *this }) {
+		if (item.IsScroll() && item._iStatFlag) {
+			_pScrlSpells |= GetSpellBitmask(item._iSpell);
```
julealgon, line 1979: Can't we use something from algorithm to do this without an explicit foreach too (like we did for counting)?

Source/control.cpp
glebm: Done

Source/control.cpp
glebm: Done

Source/control.cpp
glebm: Done

Source/control.cpp
glebm: Done

Source/inv.cpp
glebm: On `auto`, from https://google.github.io/styleguide/cppguide.html#Type_deduction:

> Use type deduction only if it makes the code clearer to readers who aren't familiar with the project, or if it makes the code safer. Do not use it merely to avoid the inconvenience of writing an explicit type.

Source/inv_iterators.hpp
glebm: `std::count_if` needs (and returns) `difference_type`

Source/inv_iterators.hpp
glebm: Done

Source/items.h
glebm: Done

Source/inv_iterators.hpp
glebm: After: https://github.com/diasurgical/devilutionX/blob/7771a08d033335e8857687dd0da258280908d345/Source/.clang-tidy#L70

Source/items.h
glebm: Errors if the return value is not used by the caller. Recommended for getter methods by clang-tidy.

Source/items.h
glebm: `IsAnyOf` seems to be a bit of an overkill in this case.

Source/player.cpp
glebm: It is possible with `std::accumulate` but is less readable because we also filter and there isn't an `accumulate_if`

Source/inv.cpp
julealgon: Well that's Google's opinion right? Are we following _that_ standard for `auto`? I don't think so.

@AJenbo , can we enforce proper use of `auto` to avoid these discussions? Current codebase doesn't follow the rule @glebm just shared.

Source/items.h
julealgon: Interesting. I guess it would make sense to omit it only when there are out arguments and the user might potentially ignore the `bool` return and stuff like that.

Thanks for clarifying.

Source/items.h
julealgon: It's not a matter of being overkill, but of consistency.

Seems like @AJenbo himself marked as resolved... I don't get it :(

Source/inv.cpp
StephenCWills: It may not make the code clearer, but I also don't think it makes it any less clear. I would prefer `auto` here simply for consistency. All the cleanups recently have been using it in this same way.

julealgon: > All the cleanups recently have been using it in this same way.

That's the most important aspect of my question: we've been doing tons of cleanup that includes `auto` uses like the one here, and now we are undoing some of it.

If we keep doing and undoing it that would be really confusing.

glebm:
Whoa, rebasing is a bit of work :|

glebm:
Can we merge this? Looks like all the discussions have been resolved.

AJenbo:
Yes, it just needs a review :)

Source/items.h
AJenbo: Found it not important/optional, wanted a better overview

Source/player.cpp
ephphatha: This could use a custom accumulator function, something like (untested):

```cpp
const InventoryAndBeltPlayerItemsRange items { *this };
_pScrlSpells = std::accumulate(items.begin(), items.end(), 0, [](uint64_t mask, const Item &item) {
    if (item.IsScroll() && item._iStatFlag) {
        mask |= GetSpellBitmask(item._iSpell);
    }
    return mask;
});
```

Don't think it's worth it for this case unless you really want to use `<algorithm>` concepts.

ephphatha (approved):
Looks good to me, left a comment on the accumulator discussion but the current implementation is good imo.

PR #2895, opened 2021-09-19
Fix opening Diablos chamber with only one lever

### LINKED ISSUE #1352: Opening Diablo's chamber with only one lever
This is a common trick abused by speedrunners regularly: flipping only one of the levers in the far west/final lever room of dlvl 16, portalling back to town, and coming back through the portal will open Diablo's chamber without needing to touch any other levers. I don't know if it's a bug or by design, but you can also use telekinesis to flip both levers in the final lever room and Diablo's chamber will open without needing to flip the first two levers in the northeastern wing of the level.

obligaron:
Fixes #1352

Similar [code](https://github.com/diasurgical/devilutionX/blob/master/Source/objects.cpp#L4224) exists for cruxes in Skeleton King's Lair.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -4236,6 +4236,9 @@ void SyncLever(const Object &lever)
 	if (lever._oSelFlag != 0)
 		return;
 
+	if (currlevel == 16 && !AreAllLeversActivated(lever._oVar8))
+		return;
+
 	ObjChangeMap(lever._oVar1, lever._oVar2, lever._oVar3, lever._oVar4);
 }
 
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -2138,27 +2138,33 @@ void OperateL1Door(int pnum, int i, bool sendflag)
 		OperateL1RDoor(pnum, i, sendflag);
 }
 
+bool AreAllLeversActivated(int leverId)
+{
+	for (int j = 0; j < ActiveObjectCount; j++) {
+		int oi = ActiveObjects[j];
+		if (Objects[oi]._otype == OBJ_SWITCHSKL
+		    && Objects[oi]._oVar8 == leverId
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -2138,35 +2138,41 @@ void OperateL1Door(int pnum, int i, bool sendflag)
...
 void OperateLever(int pnum, int i)
 {
-	if (Objects[i]._oSelFlag == 0) {
+	Object &object = Objects[i];
+	if (object._oSelFlag == 0) {
 		return;
 	}
 
 	if (!deltaload)
 		PlaySfxLoc(IS_LEVER, Objects[i].position);
```
AJenbo, line 2162:
```suggestion
		PlaySfxLoc(IS_LEVER, object.position);
```

AJenbo:
nice work :+1:

Source/objects.cpp
obligaron: Done 🙂

julealgon:
Doesn't this fix contradict the discussion we had in the issue? The main issue was not the fact that "only the last lever opened Diablo's chamber" I think, but that you were able to exploit the game by bypassing the other levers and being able to open the final chamber by operating a single lever in the last room instead of 2.

From what I'm gathering here, this fix will force players to operate _all_ levers in floor 16, right?

obligaron:
You only need to operate the two levers in the last room.
So it works like wanted in the issue.
At least I tested it with only one room. 😄 

But this is not easy to see, cause we have great variable names like `_oVar8`. 😉 
This `_oVar8` holds (only for levers 😕) a lever Id.
Each room uses a different lever id. But for all levers in the same room the id is the same.
And only if all levers with the **same id** are activated the corresponding walls/room should open.

The bug was, that this check was only done when operating the lever but not when loading the level (through load save or going up to town and back).

Issue #2894, opened 2021-09-19
Griswold's magic rock speech plays from quest panel by default


AJenbo:
1. Get at least one quest
2. Complete all active quests
3. Open the quest panel and hit enter
4. hear speech

**Expected behavior**
No speech should play if there are no pending quests

**Screenshots**
![image](https://user-images.githubusercontent.com/204594/133913446-57f3a9ae-844c-4894-b3bc-088bbb7a66e4.png)

PR #2896, opened 2021-09-19
fix quest panel

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2894
Index was -1 when pressing enter on a list that contained only done quests


glebm (approved):

AJenbo:
thanks

PR #2902, opened 2021-09-20
Add 'Damage' abstraction

julealgon:
This PR introduces a new `Damage` struct to abstract damage values (min/max vs constant) and randomization logic in a single place. All methods and models that had a pair of `int` values now leverage a `Damage` instance in its place.


Source/missiles.cpp
```diff
@@ -2843,53 +2835,39 @@ void MI_LArrow(Missile &missile)
...
-			CheckMissileCol(missile, mind, maxd, false, missile.position.tile, true);
+			CheckMissileCol(missile, damage, false, missile.position.tile, true);
 		}
 		MissilesData[missile._mitype].mResist = rst;
 	} else {
 		missile._midist++;
 
-		int mind;
-		int maxd;
+		Damage damage { 0 };
```
ephphatha, line 2857: This should be able to use the default constructor, explicit initialisation here is unnecessary since it will always be overwritten

Source/monster.cpp
```diff
@@ -223,11 +221,11 @@ void InitMonster(Monster &monster, Direction rd, int mtype, Point position)
 		monster.mLevel += 15;
 		monster.mExp = 2 * (monster.mExp + 1000);
 		monster.mHit += NIGHTMARE_TO_HIT_BONUS;
-		monster.mMinDamage = 2 * (monster.mMinDamage + 2);
-		monster.mMaxDamage = 2 * (monster.mMaxDamage + 2);
+		monster.mDamage += { 2 };
+		monster.mDamage *= 2;
```
ephphatha, line 225: I'm assuming this was split onto two lines due to only having operator overloads for `Damage, int` and not `int, Damage`, this should work:
```suggestion
		monster.mDamage = (monster.mDamage + 2) * 2;
```

`operator+` and the `Damage(int)` constructor should promote the int in addition... if not adding an explicit overload for `+ int` would be worth it given it has a lot of uses

Source/player.cpp
```diff
@@ -916,12 +916,10 @@ bool PlrHitMonst(int pnum, int m)
 	}
 
 	if ((player._pIFlags & ISPL_FIREDAM) != 0 && (player._pIFlags & ISPL_LIGHTDAM) != 0) {
-		int midam = player._pIFMinDam + GenerateRnd(player._pIFMaxDam - player._pIFMinDam);
+		int midam = player._pIFDamage.GetValue();
```
ephphatha, line 919: This isn't equivalent (old code was `rand(max - min)`, this is `rand(max-min+1)`. The old code looked like an error anyway, does `_pIFDamage.maxValue` need to be adjusted to compensate or are we fine with this change?

Source/player.cpp
```diff
@@ -1304,7 +1300,7 @@ bool DoRangeAttack(int pnum)
 			mistype = MIS_LARROW;
 		}
 		if ((player._pIFlags & ISPL_FIRE_ARROWS) != 0 && (player._pIFlags & ISPL_LIGHT_ARROWS) != 0) {
-			dmg = player._pIFMinDam + GenerateRnd(player._pIFMaxDam - player._pIFMinDam);
+			dmg = player._pIFDamage.GetValue();
```
ephphatha, line 1303: Same here

Source/damage.hpp
```diff
@@ -0,0 +1,99 @@
+#pragma once
+
+#include "engine/random.hpp"
+
+namespace devilution {
+
+struct Damage {
```
ephphatha, line 7: I was going to suggest giving this struct a more generic name so it could be used for other pairs/ranges, however the only example I could find was min/max AC on ItemData... Might still be worth it even with the sole non-damage use?

Source/player.cpp
julealgon: Check commit history. There is one commit that specifically fixes this. I know it's not equivalent, but that was on purpose.

Source/damage.hpp
julealgon: I think it should potetially _rely_ on a more generic range struct, not be replaced by it.

There is semantic value in having a type dedicated to represent damage and nothing else, as well as damange-specific methods.

ephphatha:
Sorry for the brevity in these comments, you were pushing changes so github kept asking me to refresh before I could continue writing 😓 . Hopefully it makes sense.

Overall I like the abstraction. I see most of the logic was unchanged, mainly pointed out a couple of things where it could trip up modders or someone not as familiar with the existing code.

Source/monster.cpp
julealgon: I didn't want to repeat the value like that so the use of the 2 operations in sequence was on purpose.

Source/damage.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+	constexpr Damage(int minValue, int maxValue)
+	    : minValue(minValue)
+	    , maxValue(maxValue)
+	{
+	}
+
+	[[nodiscard]] bool IsFixed() const
+	{
+		return minValue == maxValue;
```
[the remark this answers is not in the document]
julealgon: Undefined behavior basically. We don't have a real use-case for it today unless its a misuse of the type (there is one place that uses the minDamage in a switch case to set missile type, i.e. a misuse of damage in the first place).

Source/panels/charpanel.cpp
```diff
@@ -90,8 +90,8 @@ std::pair<int, int> GetDamage()
 	} else {
 		damageMod += MyPlayer->_pDamageMod;
 	}
-	int mindam = MyPlayer->_pIMinDam + MyPlayer->_pIBonusDam * MyPlayer->_pIMinDam / 100 + damageMod;
-	int maxdam = MyPlayer->_pIMaxDam + MyPlayer->_pIBonusDam * MyPlayer->_pIMaxDam / 100 + damageMod;
+	int mindam = MyPlayer->_pIDamage.minValue + MyPlayer->_pIBonusDam * MyPlayer->_pIDamage.minValue / 100 + damageMod;
+	int maxdam = MyPlayer->_pIDamage.maxValue + MyPlayer->_pIBonusDam * MyPlayer->_pIDamage.maxValue / 100 + damageMod;
 	return { mindam, maxdam };
```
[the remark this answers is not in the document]
julealgon: That would require refactoring the method to return `Damage` too. I can look into it tomorrow.

Source/objects.cpp
```diff
@@ -2715,9 +2713,9 @@ bool OperateShrineWeird(int pnum)
 	auto &player = Players[pnum];
 
 	if (!player.InvBody[INVLOC_HAND_LEFT].isEmpty() && player.InvBody[INVLOC_HAND_LEFT]._itype != ItemType::Shield)
-		player.InvBody[INVLOC_HAND_LEFT]._iMaxDam++;
+		player.InvBody[INVLOC_HAND_LEFT]._iDamage += { 0, 1 };
```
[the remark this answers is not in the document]
julealgon: I disagree here. IMHO we should keep the operation at the highest level possible.

Source/player.cpp
ephphatha: Ahh, I should've checked that (once github let me). Thanks for clarifying :)

Source/objects.cpp
```diff
@@ -2684,9 +2683,8 @@ bool OperateShrineGloomy(int pnum)
 		case ItemType::Bow:
 		case ItemType::Mace:
 		case ItemType::Staff:
-			item._iMaxDam--;
-			if (item._iMaxDam < item._iMinDam)
-				item._iMaxDam = item._iMinDam;
+			if (!item._iDamage.IsFixed())
+				item._iDamage -= { 0, 1 };
```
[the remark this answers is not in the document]
julealgon: It changes the logic to avoid "setting a value and undoing the operation".

Previously, this was decrementing max damage, and if it became less than min damage (only possible if the values were equal prior to the decrement, or "fixed") it would set maxDamage to minDamage, basically undoing the operation.

The new logic is more sane and does the same thing: if the damage is already fixed, there is no need decrementing max component and then undoing it. If it is _not_ fixed, then it will always be able to decrement max damage without creating the problem.

julealgon:
> Sorry for the brevity in these comments, you were pushing changes so github kept asking me to refresh before I could continue writing 😓 . Hopefully it makes sense.

No problem, I appreciate the review :)

I'm trying to make these damn tests pass now..... will have to continue tomorrow since its a bit late here and I have to go to bed :(

julealgon:
> AJenbo added this to the 1.4.0 milestone now

😞

Source/panels/charpanel.cpp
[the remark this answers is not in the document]
ephphatha: Gah, didn't notice this was returning `std::pair<int, int>`. In that case a temporary would still save some repetition but using `Damage` for this method might be more suitable.

AJenbo:
> > AJenbo added this to the 1.4.0 milestone now
> 
> disappointed

Sorry, the number of remaining mandatory features for 1.3.0 is getting low at this point and the deadline is also getting tighter. I would rather have an early 1.3.0 release then miss the October deadline. So that basically means a freeze on major refactoring for the next 1-5 weeks.

If I had known earlier I would probably have send out a warning but the last puzzle piece just fell in to place over the night.

Source/damage.hpp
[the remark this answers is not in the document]
StephenCWills: https://github.com/diasurgical/devilutionX/issues/1672#issuecomment-826102213

julealgon:
> So that basically means a freeze on major refactoring for the next 1-5 weeks.

I think this is far from major refactoring, but I guess there is no point in pursuing it if we are going to freeze it for up to 5 weeks. 

See you guys in 5 weeks then.

AJenbo:
> > So that basically means a freeze on major refactoring for the next 1-5 weeks.
> 
> I think this is far from major refactoring, but I guess there is no point in pursuing it if we are going to freeze it for up to 5 weeks.
> 
> See you guys in 5 weeks then.

Ok, let me define it as anything over 100 lines that doesn't directly contribute to the binary release.

Hopefully that will limit the churn and we can get the release out quickly. If it works out well then in 3 weeks we can merge a bunch of nice cleanups that are all nice and ready, which out much conflict.

julealgon:
https://github.com/julealgon/devilutionX/pull/4

PR #2904, opened 2021-09-20
Unicode font rendering

AJenbo:
This adds full support for the first 1279 Unicode symbols:
- Latin Extended-A
- Latin Extended-B
- IPA Extensions
- Spacing Modifier Letters
- Combining Diacritical Marks
- Greek and Coptic
- Cyrillic

Which is needed for the following existing translations:
- Russian
- Bulgarian
- Polish

image


Source/engine/render/text_render.cpp
```diff
@@ -166,40 +184,48 @@ int AdjustSpacingToFitHorizontally(int &lineWidth, int maxSpacing, int character
...
-			}
+		if (lineWidth - spacing <= width) {
+			continue; // String is still within the limit, continue to the next symbol
 		}
 
-		if (j == lineStart) { // Single word longer than width
-			if (i == textLength)
-				break;
-			j = i;
+		if (lastKnownSpaceAt == -1) {
```
AJenbo, line 221:
```suggestion
		if (lastKnownSpaceAt == -1) { // Single word longer than width
```

NikoVP:
The correct list of blocks is:

- Basic Latin
- Latin-1 Supplement
- Latin Extended-A
- Latin Extended-B
- Greek and Coptic
- Cyrillic

Which allows display of most languages based on Latin, Cyrillic and Greek

Source/help.cpp
```diff
@@ -101,7 +101,7 @@ std::vector<std::string> HelpTextLines;
 void InitHelp()
 {
 	HelpFlag = false;
-	char tempString[512];
```
vlolteanu, line 104: There might be other character arrays that need to be doubled in size to accommodate the larger characters. I'd search for `strcpy` and `strcat`. Some of them, like `tempstr` seem large enough as is; others I'm not sure about.

Source/DiabloUI/diabloui.cpp
```diff
@@ -226,15 +225,13 @@ void UiFocusPageDown()
...
 }
 
 #ifdef __vita__
 void selhero_SetName(char *in_buf, char *out_buf, int cnt)
 {
-	std::string output = utf8_to_latin1(in_buf);
-	strncpy(out_buf, output.c_str(), cnt);
+	strncpy(out_buf, in_buf, cnt);
 }
 #endif
```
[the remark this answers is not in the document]
AJenbo: You get a decode error and the string stops rendering at that point. So usually it should be acceptable.

Source/help.cpp
AJenbo: I would prefer to do that in a later update, and maybe search them out by testing (ASAN reports them at runtime)

Source/engine/render/text_render.cpp
```diff
@@ -132,19 +154,38 @@ void UnloadFont(GameFontTables size, text_color color)
...
+	for (; *textData != '\0'; i++) {
+		textData = utf8_decode(textData, &next, &error);
+		if (error || next == '\n')
 			break;
 
-		uint8_t frame = text[i] & 0xFF;
-		lineWidth += FontKerns[size][frame] + spacing;
+		uint8_t frame = next & 0xFF;
+		uint32_t unicodeRow = next >> 8;
+		if (unicodeRow != currentUnicodeRow || kerning == nullptr) {
```
StephenCWills, line 180: > Run-Time Check Failure #3 - The variable 'currentUnicodeRow' is being used without being initialized.

image

Source/engine/render/text_render.cpp
```diff
@@ -166,40 +207,56 @@ int AdjustSpacingToFitHorizontally(int &lineWidth, int maxSpacing, int character
...
+			lastKnownSpaceAt = -1;
 			lineWidth = 0;
 			continue;
 		}
 
-		uint8_t frame = text[i] & 0xFF;
-		lineWidth += FontKerns[size][frame] + spacing;
+		uint8_t frame = next & 0xFF;
+		uint32_t unicodeRow = next >> 8;
+		if (unicodeRow != currentUnicodeRow || kerning == nullptr) {
```
StephenCWills, line 234: > Run-Time Check Failure #3 - The variable 'currentUnicodeRow' is being used without being initialized.

Source/engine/render/text_render.cpp
```diff
@@ -240,52 +297,62 @@ uint32_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
...
+	uint32_t next;
+	uint32_t currentUnicodeRow;
+	int error;
+	for (; *textData != '\0'; previousPosition = textData) {
+		textData = utf8_decode(textData, &next, &error);
+		if (error)
+			break;
+
+		uint32_t unicodeRow = next >> 8;
+		if (unicodeRow != currentUnicodeRow || font == nullptr) {
```
StephenCWills, line 317: > Run-Time Check Failure #3 - The variable 'currentUnicodeRow' is being used without being initialized.

StephenCWills (commented):
There were a couple more of those, apparently.

Source/engine/render/text_render.cpp
AJenbo: :+1: fixed

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
StephenCWills: It seems that backspace will screw up multi-byte characters. To make this hero, I typed `Ā<backspace>erika`.

image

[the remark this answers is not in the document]
AJenbo: Right that would cause a decoding issue with the very first character, since it only deletes half of it. 

Shouldn't be to hard to fix.

Source/engine/render/text_render.cpp
```diff
@@ -92,14 +92,38 @@ text_color GetColorFromFlags(UiFlags flags)
...
+	uint32_t fontId = (color << 24) | (size << 16) | row;
+
+	auto hotFont = Fonts.find(fontId);
+	if (hotFont != Fonts.end()) {
+		return hotFont->second.get();
+	}
 
 	char path[32];
-	sprintf(path, "fonts\\%i-00.pcx", FontSizes[size]);
+	sprintf(path, "fonts\\%i-%02d.pcx", FontSizes[size], row);
```
StephenCWills, line 124: `LoadFileInMem()` triggers a fatal error if I type a character that's not supported.

image

Source/engine/render/text_render.cpp
```diff
@@ -20,14 +20,14 @@
...
+#include "utils/utf8.h"
 
 namespace devilution {
 
 namespace {
 
 std::unordered_map<uint32_t, std::unique_ptr<Art>> Fonts;
-
-std::array<std::array<uint8_t, 256>, 5> FontKerns;
+std::unordered_map<uint32_t, std::unique_ptr<std::array<uint8_t, 256>>> FontKerns;
```
vlolteanu, line 30: You can do away with the `unique_ptr`, and just store the `array`.

Source/engine/render/text_render.cpp
```diff
@@ -92,14 +92,38 @@ text_color GetColorFromFlags(UiFlags flags)
...
+	}
+
+	char path[32];
+	sprintf(path, "fonts\\%i-%02d.bin", FontSizes[size], row);
+
+	auto kerning = std::make_unique<std::array<uint8_t, 256>>();
+
+	LoadFileInMem(path, kerning.get());
+
+	return FontKerns.insert(make_pair(fontId, move(kerning))).first->second.get();
```
vlolteanu, line 111: After ditching the `unique_ptr`, you can do something like:
```
auto *kerning = &FontKerns[fontId];
LoadFileInMem(path, kerning);
return kerning;
```

Source/engine/render/text_render.cpp
```diff
@@ -109,19 +133,17 @@ void LoadFont(GameFontTables size, text_color color)
 		LoadMaskedArt(path, font.get(), 256, 1);
 	}
 
-	uint32_t fontId = (color << 24) | (size << 16);
-	Fonts.insert(make_pair(fontId, move(font)));
-
-	sprintf(path, "fonts\\%i-00.bin", FontSizes[size]);
-	LoadFileInMem(path, FontKerns[size]);
+	return Fonts.insert(make_pair(fontId, move(font))).first->second.get();
```
vlolteanu, line 136: Same as with `FornKerns`: no need for the `unique_ptr`.

Source/engine/render/text_render.cpp
AJenbo: yeah, plan is to add the rest, but maybe we should simply skip any such missing symbols :thinking:

Source/engine/render/text_render.cpp
AJenbo: wont we leak when unloading things?

StephenCWills: I guess so long as we don't store pointers to dynamically allocated memory in the `unordered_map`, then it shouldn't be a problem. In other words, we could start leaking if we used `std::unordered_map<uint32_t, Art*>` as opposed to `std::unordered_map<uint32_t, Art>`.

AJenbo:
> The same applies to ingame chat's 😉

Same what? I think you made your comments in the wrong way, I have no context and can't reply to them

AJenbo:
> We could write ¿ 😁

Yes, that's the point, do you se an issue or are you exited about it?

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
StephenCWills: Actually, backspace seems to work fine for the in-game chat so I intentionally didn't mention it.

StephenCWills:
> Same what? I think you made your comments in the wrong way, I have no context and can't reply to them

It looks like his replies got added to his review so they show up in two places. I have no idea why GitHub would allow this. It's just confusing.

obligaron:
Yes I replied/added to existing conversations.
And interestingly I can't close this duplicates. 😕

Source/engine/render/text_render.cpp
obligaron: @AJenbo I would vote for writing a error character. A questions mark or a box or something.
The program shouldn't stop executing or simply print nothing.
That could also help to diagnostic future errors.

Source/engine/render/text_render.cpp
```diff
@@ -109,19 +133,17 @@ void LoadFont(GameFontTables size, text_color color)
...
 {
-	uint32_t fontId = (color << 24) | (size << 16);
+	uint32_t fontStyle = (color << 24) | (size << 16);
 
 	for (auto font = Fonts.begin(); font != Fonts.end();) {
-		if ((font->first & 0xFFFF0000) == fontId) {
+		if ((font->first & 0xFFFF0000) == fontStyle) {
 			Fonts.erase(font++);
 		} else {
 			font++;
```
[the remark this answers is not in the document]
StephenCWills: I don't disagree with this suggestion, but just because I found it interesting.

> https://en.cppreference.com/w/cpp/container/unordered_map/erase
> References and iterators to the erased elements are invalidated. Other iterators and references are not invalidated.

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
StephenCWills: It looks like backspace works for the in-game chat because we're using the return value from `DrawString()` to force a null character onto the "end" of it.

https://github.com/diasurgical/devilutionX/blob/3426a3d76f11a5973c50915eb3fa3185508be7bf/Source/control.cpp#L1689-L1690

This works because the rendering stops when it encounters an error so the null character overwrites the broken character. Incidentally, this is also the only place in the whole codebase where the return value of `DrawString()` is used.

StephenCWills (changes requested):
Since you said backspace would be easy, and because it's an easy issue for the player to stumble on, I think that should be fixed before merging. Everything else that was mentioned can be handled in future PRs so I'd consider those optional.

Source/engine/render/text_render.cpp
AJenbo: I guess that might be safe to do considering that UTF-8 is self synchronizing. But I would go for `?` as that is more canonical and simpler to add to the code.

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
AJenbo: Yeah I did look in to getting rid of this, but it's not super simple to do, we would need a function that does almost the same to calculate the cut off point. Maybe something could be done via splitting DrawString() up in to sub routines, but atm. I don't think it's worth it.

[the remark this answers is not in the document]
StephenCWills: Based on what I know of UTF-8, I think backspace could be done with...

```c++
std::size_t index = strlen(TalkMessage) - 1;
while (index > 0 && (TalkMessage[index] & 0xC0) == 0x80)
	index--;
if (index >= 0)
	TalkMessage[index] = '\0';
```

This assumes there are no broken characters in the string. If there are broken characters, then you might see some weird behavior, but nothing catastrophic.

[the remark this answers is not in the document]
AJenbo: In case it wasn't clear my previous comment was in regards to `DrawString()` returning a count for what will fit in a box. A potential solution for that would be to do a `GetLineWidt(TalkMessage) < 3 * width` check in `control_new_text()` and prevent the text from being added in the first place.

As for a back space function i simply decode the string and return the position of the last found symbol (see latest commit). Your method should work as well and is a bit simpler, but might be a bit more cryptic to read, let me know if you think it would be better to use it.

Source/engine/render/text_render.cpp
AJenbo: I need up to a 1024 arrays of 256 bytes depending on the language and chat content. But in most cases 4 arrays of 256 is all that is needed so having it in a variable size container allows me to save up to 0.2M of ram.

AJenbo: Took me a few tries to figure out how this was suppose to work, but once i figured it out the code is much cleaner, so thanks :)

Source/DiabloUI/diabloui.cpp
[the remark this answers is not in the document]
StephenCWills: Ahh, I get it now. I had been thinking that surely backspace wasn't the only reason for using that return value.

As for how you handled backspace, your function should work better than my suggestion.

Source/utils/utf8.h
```diff
@@ -64,23 +63,23 @@ utf8_decode(const unsigned char *buf, uint32_t *c, int *e)
 	*e ^= 0x2a; // top two bits of each tail byte correct?
 	*e >>= shifte[len];
 
-	return next;
+	return reinterpret_cast<const char *>(next);
 }
 
-inline std::string utf8_to_latin1(const char *in)
+inline int FindLastUft8Symbols(const char *text)
```
StephenCWills, line 69:
```suggestion
inline int FindLastUtf8Symbols(const char *text)
```

Source/control.cpp
```diff
@@ -1839,9 +1823,7 @@ bool control_presskeys(int vkey)
 	} else if (vkey == DVL_VK_RETURN) {
 		ControlPressEnter();
 	} else if (vkey == DVL_VK_BACK) {
-		std::size_t len = strlen(TalkMessage);
-		if (len > 0)
-			TalkMessage[len - 1] = '\0';
+		TalkMessage[FindLastUft8Symbols(TalkMessage)] = '\0';
```
StephenCWills, line 1826:
```suggestion
		TalkMessage[FindLastUtf8Symbols(TalkMessage)] = '\0';
```

Source/DiabloUI/diabloui.cpp
```diff
@@ -328,10 +331,7 @@ void UiFocusNavigation(SDL_Event *event)
 #endif
 			case SDLK_BACKSPACE:
 			case SDLK_LEFT: {
-				int nameLen = strlen(UiTextInput);
-				if (nameLen > 0) {
-					UiTextInput[nameLen - 1] = '\0';
-				}
+				UiTextInput[FindLastUft8Symbols(UiTextInput)] = '\0';
```
StephenCWills, line 334:
```suggestion
				UiTextInput[FindLastUtf8Symbols(UiTextInput)] = '\0';
```

StephenCWills (changes requested):
Just barely noticed this typo as I was finishing my review (Uft8 vs Utf8). Apart from this, I think it's ready for merge.

PR #2905, opened 2021-09-20
🌐 Update Italian localization

M0Rf30:
As in the title.
I also post some notes here that may be useful, but I think you already know.
Some buttons compress longer words. An example:
The word "CHAR", which stands for character, is immediately translated to "Personaggio". But I don't like the idea of ​​using the abbreviation PG to accomodate the button aspect. I had thought of the Italian equivalent of Profile ("Profilo"). Before disrupting the translation, I'm waiting for your feedback. For the rest, as always, a splendid job. Thanks for supporting accented letters :tada:

### DIFF
--- a/Translations/it.po
+++ b/Translations/it.po
@@ -1,8 +1,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: DevilutionX\n"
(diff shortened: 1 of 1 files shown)


AJenbo:
"Profilo" sounds like a good option.

AJenbo (approved):

PR #2906, opened 2021-09-21
Fix item out of bounds hack

### LINKED ISSUE #1356: Corrupt items causing overvflow
**Important information**
Windows 10 / Vanilla

**Describe**
Glitch item

**To Reproduce**
1. Get someone on battle.net to dupe you one

**Expected behavior**
It shouldn't exist

**Screenshots**
unknown

**Additional context**
Causes insane character damage (only works in PVE)
10 STR required to wear
Says it raises Vitality, but unable to right click it to consume

Uses Optic Amulet as base item, despite using Map of the Stars graphic (idx: 0A)
Has a seed of Demical: -1 / Hexadecimal: FFFFFFFFFFFFFFFF
Has a Creation Flag of 0
[single_0.zip](https://github.com/diasurgical/devilutionX/files/6253914/single_0.zip)

AJenbo:
Fixes #1356

image


Source/pack.cpp
```diff
@@ -152,7 +148,7 @@ void UnPackItem(const ItemPack *is, Item *id, bool isHellfire)
 		idx = RemapItemIdxFromDiablo(idx);
 	}
 
-	if (!IsItemAvailable(idx)) {
+	if (idx == IDI_NONE || idx > IDI_LAST || !IsItemAvailable(idx)) {
```
glebm, line 151:
```suggestion
	if (idx < 0 || idx > IDI_LAST || !IsItemAvailable(idx)) {
```

qndel: what about <= IDI_NONE? : P

qndel (changes requested):
Move idx check to IsItemAvailable function

Source/pack.cpp
AJenbo: Though about it, but it doesn't really make sens given its context

Source/loadsave.cpp
```diff
@@ -1442,14 +1442,14 @@ const int HellfireItemSaveSize = 372;
 
 void RemoveInvalidItem(Item &item)
 {
-	bool isInvalid = !IsItemAvailable(item.IDidx) || !IsUniqueAvailable(item._iUid);
+	bool isInvalid = item.IDidx > IDI_LAST || !IsItemAvailable(item.IDidx) || !IsUniqueAvailable(item._iUid);
```
ephphatha, line 1445: This doesn't use the same check for negative values as `UnPackItem` does.

Source/itemdat.h
```diff
@@ -13,7 +13,7 @@
 namespace devilution {
 
 /** @todo add missing values and apply */
-enum _item_indexes : int16_t {
+enum _item_indexes : int16_t { // TODO defines all indexes in AllItemsList
```
ephphatha, line 16: Are negative values expected to correspond to valid Item IDs? If not using an unsigned type for the backing type would save having to do `<0` checks, letting that be handled by `>= LAST`

AJenbo: -1 == no item

Source/loadsave.cpp
AJenbo: Yeah I got a suggestion for changing the other one, but didn't get around to change this one manually yet :)

Source/pack.cpp
ephphatha: From what I could see `IsItemAvailable` is used mostly in the context of a loop over `AllItemsList` so the index (ID) is assumed to be within the range of 0-166 (`AllItemsList[167]` is a marker for the end of the list).

This and the use in `RemoveInvalidItem` break that assumption. I'd like to see this check moved into the function as well since to me it makes no sense than an invalid item is available.

Source/itemdat.h
ephphatha: -1 isn't considered a valid item though, `AllItemsList[-1]` is an out of bounds access. Looking at the way `Item.IDidx` is used it seems like `Item._itype == None` is used as the guard, then `IDidx` is used as a straight index into that array.

qndel (approved):
👍

PR #2911, opened 2021-09-21
Use SDL Textinput for ingame chat

### LINKED ISSUE #1195: Ingame chat hardcoded to US layout
Hello there,

This is probably related to issue #1026 but since I typically play multiplayer in the same room as my sister, I hadn't tried text chat yet until yesterday.

So, in case that makes any significant difference, I'm using DevilutionX v1.1.0 (from devilutionx-windows-i386.zip) on Windows 10 x64, v20H2 and up to date with the "French (France)" keyboard layout (see https://en.wikipedia.org/wiki/AZERTY#/media/File:KB_France.svg for reference)

I've noticed that while typing text in the chat field during a multiplayer game, while the alpha layout may seem to be correctly recognized because typing "AZERTY" does show as "AZERTY" and not "QWERTY", some keys seem to not be recognized at all, or not as the right character. By that I mean that either no character appears when trying to type them, or it's not the correct character that appears.

Here's a list (might not be completely exhaustive)

Numpad:
- Numpad numbers, Numpad "/", numpad "*" and numpad "." are not recognized
- Numpad "+" is replaced with "="
- Numpad "-" is correctly recognized

On the alpha part of the keyboard:
- "<" and ">"  (between left shift and the "w" key) are not recognized
- on the alpha pad, "," and ";" are correctly recognized, but holding shift while using these should produce  "?" and ".", yet "<" and ":" appear instead
- the key with "=" and "+" (on the left of the backspace key) works fine, but the key one position further left doesn't do anything
- the keys with "^", "$", "ù", "*", ":" and "!" do not work
- the keys above AZERTY… are recognized as numbers unless Shift is held, which is opposite of the normal behavior on french keyboards, and capslock has no effect on that
- Assuming we'd accept the weirdness about inverted "Shift" key behaviour for those keys, then when holding Shift and typing on the keys above AZERTY… from 1 to 0, we should get the following: &é"'(-è_çà but instead we get the same result as if we were typing on an US keyboard, that is !@#$%^&*()

I suppose I should also point out that the official Diablo game, at least when patched in 1.09, did not have that issue.

Also, the issue only happens when trying to use chat in game, but every key is recognized as it should when entering the name for a new character or when setting the password before creating a multiplayer game.

obligaron:
Fixes #1195

Works only with SDL2.

### DIFF
--- a/Source/miniwin/misc_msg.cpp
+++ b/Source/miniwin/misc_msg.cpp
@@ -23,6 +23,7 @@
 #include "utils/log.hpp"
 #include "utils/sdl_compat.h"
 #include "utils/stubs.h"
+#include "utils/utf8.h"
 
 #ifdef __vita__
 #include "platform/vita/touch.h"
@@ -532,8 +533,15 @@ bool FetchMessage_Real(tagMSG *lpMsg)
 		return FalseAvail("SDL_KEYMAPCHANGED", 0);
 #endif
 	case SDL_TEXTEDITING:
+		if (gbRunGame)
+			break;
 		return FalseAvail("SDL_TEXTEDITING", e.edit.length);
 	case SDL_TEXTINPUT:
+		if (gbRunGame) {
+			std::string output = utf8_to_latin1(e.text.text);
+			control_new_text(output);
+			break;
+		}
 		return FalseAvail("SDL_TEXTINPUT", e.text.windowID);
 	case SDL_WINDOWEVENT:
 		switch (e.window.event) {
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1772,6 +1772,10 @@ void control_type_message()
 		return;
 
 	talkflag = true;
+	int x = PANEL_LEFT + 200;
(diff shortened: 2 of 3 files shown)


Source/miniwin/misc_msg.cpp
```diff
@@ -532,8 +533,15 @@ bool FetchMessage_Real(tagMSG *lpMsg)
 		return FalseAvail("SDL_KEYMAPCHANGED", 0);
 #endif
 	case SDL_TEXTEDITING:
+		if (gbRunGame)
+			break;
 		return FalseAvail("SDL_TEXTEDITING", e.edit.length);
 	case SDL_TEXTINPUT:
+		if (gbRunGame) {
+			std::string output = utf8_to_latin1(e.text.text);
```
AJenbo, line 541: Oh no now i have to remove another `utf8_to_latin1` in my PR :D

AJenbo (approved):
Nice, good timing too :)

yuripourre:
Amazing!

PR #2912, opened 2021-09-21
Hardware rendering for the virtual gamepad

StephenCWills:
This PR renders the gamepad using the SDL rendering API instead of `SDL_BlitScaled()` to the output surface.

This should solve the following rendering issues...
* Dissatisfactory scaling algorithms in `SDL_BlitScaled()`
* Downscaling to the the game's resolution before upscaling to the display resolution

---

Here is a particularly extreme comparison using the following settings on my 1920x1200 display.

```ini
[Graphics]
Width=640
Height=480
Fullscreen=1
Upscale=1
Fit to Screen=1
Scaling Quality=2
Integer Scaling=0
```

**SDL_BlitScaled:**
image

**HW render:**
image

### DIFF
--- a/Source/DiabloUI/art.h
+++ b/Source/DiabloUI/art.h
@@ -13,6 +13,10 @@ struct Art {
 	int frame_height;
 	unsigned int palette_version;
 
+#ifndef USE_SDL1
+	SDLTextureUniquePtr texture;
+#endif
+
 	Art()
 	{
 		surface = nullptr;
@@ -20,6 +24,10 @@ struct Art {
 		logical_width = 0;
 		frame_height = 0; // logical frame height (before scaling)
 		palette_version = 0;
+
+#ifndef USE_SDL1
+		texture = nullptr;
+#endif
 	}
 
 	int w() const
@@ -35,6 +43,10 @@ struct Art {
 	void Unload()
 	{
 		surface = nullptr;
+
+#ifndef USE_SDL1
+		texture = nullptr;
+#endif
 	}
 };
 
--- a/Source/dx.cpp
+++ b/Source/dx.cpp
@@ -7,6 +7,7 @@
 
 #include <SDL.h>
 
+#include "controls/touch/renderers.h"
 #include "engine.h"
 #include "options.h"
 #include "storm/storm.h"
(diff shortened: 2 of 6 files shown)


StephenCWills:
@NikoVP FYI

Source/options.cpp
```diff
@@ -229,8 +229,10 @@ void LoadOptions()
 #else
 	sgOptions.Graphics.bFullscreen = true;
 #endif
-#if !defined(USE_SDL1)
+#if !defined(VIRTUAL_GAMEPAD) && !defined(USE_SDL1)
 	sgOptions.Graphics.bUpscale = GetIniBool("Graphics", "Upscale", true);
+#elif defined(VIRTUAL_GAMEPAD) && !defined(USE_SDL1)
+	sgOptions.Graphics.bUpscale = true;
```
AJenbo, line 235: I'm a bit unsure if we should force this on just because it might have a negative effect on a partially related but optional feature :thinking:

Source/dx.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include <SDL.h>
 
+#include "controls/touch/renderers.h"
```
AJenbo, line 10: #ifdef VIRTUAL_GAMEPAD

StephenCWills: The header itself checks the precompiler flag so I figured it wouldn't be necessary.

Source/options.cpp
StephenCWills: I'm forcing it on because the render API isn't used unless the upscale flag is set.
https://github.com/diasurgical/devilutionX/blob/6dbe4340458bfea947b7b52772252ec715dd7052/Source/utils/display.cpp#L220-L231

If I don't force it on, then I have to fall back on `SDL_BlitScaled()`, which is tricky because of how I have to shim the rendering call into `RenderPresent()`. Or, I suppose, we could simply not render the virtual gamepad.

StephenCWills: After thinking about it for a bit, I decided it wasn't all that tricky after all.

AJenbo: Yeah I was thinking to opt for not rendering it. Some one with a crappy Android gamepad (emulating touch events) might even see it as a feature :joy: :joystick:

Source/DiabloUI/art.h
```diff
@@ -13,13 +13,21 @@ struct Art {
 	int frame_height;
 	unsigned int palette_version;
 
+#ifndef USE_SDL1
+	SDLTextureUniquePtr texture;
+#endif
```
AJenbo, line 18: If nothing else, this is interesting, lets see where it leads.

AJenbo (approved):

NikoVP:
@StephenCWills  will try to work on the remaining buttons this or next week but currently have my hands occupied with personal stuff. Will get back to it as soon as I can get some free time.

PR #2914, opened 2021-09-21
Fix Fedora build command

### LINKED ISSUE #896: Build from source failed on Fedora 32 with error "/usr/bin/ld: cannot find -lstdc++"

yuripourre:
This is the right way to fix the "/usr/bin/ld: cannot find -lstdc++" error on Fedora.

Tested on Fedora 33

This PR can be closed: #1040

Closes #896


AJenbo (approved):

AJenbo:
Nice work

PR #2918, opened 2021-09-21
Update french translation

mmaret:
Some fuzzy translation are still remaining after this

### DIFF
--- a/Translations/fr.po
+++ b/Translations/fr.po
@@ -211,7 +211,7 @@ msgstr "Remerciements à"
 
 #: Source/DiabloUI/credits_lines.cpp:200
 msgid "In memory of"
-msgstr "À la mémoire de "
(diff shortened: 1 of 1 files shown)


Translations/fr.po
```diff
@@ -6448,81 +6245,65 @@ msgstr ""
...
-msgstr "Bâton de Lazare"
+msgstr "L'Antre de l’Archevêque Lazarus"
 
 #: Source/spelldat.cpp:16
-#, fuzzy
-#| msgid "Fire "
 msgctxt "spell"
 msgid "Firebolt"
-msgstr "Feu "
+msgstr "[sort] Feu "
```
AJenbo, line 6276: Why is all of this prefixed with [sort]?

mmaret: Because I get fooled by my tool :D

PR #2919, opened 2021-09-21
Clean up quests.cpp global naming

AJenbo:


qndel (approved):
Now this finally makes some sense ;)

PR #2921, opened 2021-09-22
Fix missing switch cases

AJenbo:


ephphatha (approved):
Given it just got duplicated might as well get this one approved :)

PR #2924, opened 2021-09-22
Remove uses of Items[] as a temporary for other Item variables

ephphatha:
This was a holdover from when the Item creation/initialisation functions operated on an index into the global Items array. Now that they all operate on Item& arguments there was no need to use Items[0] as a temporary space.

As far as I can tell this is every explicit reference to Items[0], but I can't be sure since Visual Studio seems to like breaking its search indexes on my machine...

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -2565,14 +2565,12 @@ void InitItemGFX()
 
 void InitItems()
 {
-	memset(&Items[0], 0, sizeof(*Items));
-	GetItemAttrs(Items[0], IDI_GOLD, 1);
-	golditem = Items[0];
+	golditem = {};
+	GetItemAttrs(golditem, IDI_GOLD, 1);
 	golditem._iStatFlag = true;
 	ActiveItemCount = 0;
 
-	for (int i = 0; i < MAXITEMS; i++) {
-		auto &item = Items[i];
+	for (auto &item : Items) {
 		item._itype = ItemType::None;
 		item.position = { 0, 0 };
 		item._iAnimFlag = false;
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4430,9 +4430,6 @@ void SpawnSmith(int lvl)
 {
 	constexpr int PinnedItemCount = 0;
 
-	Item holditem;
-	holditem = Items[0];
-
 	int maxValue = 140000;
 	int maxItems = 20;
 	if (gbIsHellfire) {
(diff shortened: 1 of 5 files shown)


Source/items.cpp
```diff
@@ -4471,27 +4464,26 @@ void SpawnPremium(int pnum)
...
-			// Discard first 2 items and shift next 3
-			std::move(&premiumitems[2], &premiumitems[4] + 1, &premiumitems[0]);
-			SpawnOnePremium(3, premiumlevel + premiumlvladd[3], pnum);
-			premiumitems[4] = premiumitems[5];
-			SpawnOnePremium(5, premiumlevel + premiumlvladd[5], pnum);
+			// Discard first 2 items and shift the rest of the list up
+			std::move(premiumitems.begin() + 2, premiumitems.end(), premiumitems.begin()); // This could use premiumitems.begin() + 6 for the second arg but no harm doing the whole range anyway.
+			// Roll new items for the last 2 places
+			SpawnOnePremium(premiumitems[4], premiumlevel + premiumlvladd[4], pnum);
+			SpawnOnePremium(premiumitems[5], premiumlevel + premiumlvladd[5], pnum);
```
ephphatha, line 4486: This looked like a bug, the offsets used after the move were nonsensical. I can split this out to a separate PR if we need to keep the old logic for some reason, but I can't see it breaking anything.

AJenbo:
is the same doable for Items[max] ?

ephphatha:
Despite having the unittests configuration active VS wasn't actually building tests -_-. I've updated those methods and PackItem to use references to match what was done for UnPackItem while I was removing the use of Items[MAXITEMS] as a temporary.

Not strictly necessary and it's inflated the number of lines changed, but figured it was best to be consistent.

StephenCWills:
> As far as I can tell this is every explicit reference to Items[0], but I can't be sure since Visual Studio seems to like breaking its search indexes on my machine...

It's not just your machine. 😩 
`git clean -fdx .vs`

Source/items.cpp
AJenbo: What would the you expect the consequence of the bug to be, is there anyway for us to test it?

ephphatha: The logic here is a bit confusing, just realised this isn't a correct fix either.

It appears that this was an attempt to avoid rerolling the shop premium items if the player levels up while in town. It's probably not necessary with towners using a fixed seed but doesn't hurt to keep in.

The behaviour for Diablo gameplay was correct (but somewhat wasteful) in that it'd take the item offsets of `-1, -1, 0, 0, 1, 2` and shift it to `0, 0, 1, X, X, 2`, then fix the garbage at positions 3/4 and ends up with `0, 0, 1, 1, 2, 3`. This change introduces a bug and results in `0, 0, 1, 2*, 2, 3`, position 3 ends up one level higher than it should be.

Hellfire logic was slightly off. Starting with offsets `-1, -1, -1, 0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 3, 3` it discards the below player level items and ends up with `0, 0, 0, 0*, 1, 1, 1, 1*, 2, 2, X, X, 3, 3` after the move. At this point positions 3 and 7 are one level lower than they would have been with a fresh shop. It then tried to fix the garbage/end of the list and gives a final list of `0, 0, 0, 0*, 1, 1, 1, 1*, 2, 2, 2, 3, 3^, 3*, 4`. Position 3 and 7 didn't get rerolled so are still a level lower than they should be. Position 12 gets rerolled when it should've been copied from 14, position 13 gets copied from 14 when it should've been rerolled so the resulting item is one level lower than it should be.

This change gives an outcome of `0, 0, 0, 0*, 1, 1, 1, 1*, 2, 2, 3*, 3, 3^, 4, 4`. It doesn't fix positions 3 and 7 getting the wrong level (need to discard the first 4 items to make that right), but also results in one of the +3 level items taking the place of an item that should have been +2 and a new +3 level item being rolled (like the old code). A proper fix needs to insert two more +2 level items, keep the previous +3 level items, and reroll the last two places as +4 level items.

The other thing is this doesn't really need to be done as a loop, but evidently my brain isn't working well enough to worry about that refactoring at the moment 😅. I'll fix the logic properly and pull it into a separate PR.

I did see the discord message you copied on the other PR as well, I'm mainly opening these make it visible and hopefully avoid someone duplicating the effort unnecessarily. I don't expect any of these changes to make it into 1.3.

ephphatha: Split into https://github.com/diasurgical/devilutionX/pull/2938 - Ill remove that commit from this PR

AJenbo: > I'm mainly opening these make it visible and hopefully avoid someone duplicating the effort unnecessarily. I don't expect any of these changes to make it into 1.3.

Excellent, I just don't want you to feel like the changes are unwanted/ignored :)

AJenbo (approved):

PR #2926, opened 2021-09-22
Fix an assert at the wrong place

mmaret:
It seems that the assert try to protect an out of bound access to the
Players array, but it legitime to access the Monsters Array with the
same index before

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4156,7 +4156,6 @@ void M_WalkDir(int i, Direction md)
 
 void GolumAi(int i)
 {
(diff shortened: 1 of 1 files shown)


qndel:
No.

mmaret:
Then, Can I share a savegame that raise this assert ?

[save.zip](https://github.com/diasurgical/devilutionX/files/7212081/save.zip)
Does it occur because the savegame were created with a previous buggy version of the game ?

AJenbo:
yes, and sine the asserts are only there for debug builds it won't break the save game for release. But I do welcome a fix to how old save games are loaded if you have a good solution.

PR #2927, opened 2021-09-22
Drop unique tweak

qndel:
Makes all uniques droppable with "dropu" command in singleplayer. Optimizes for speed.


AJenbo:
Please don't mege your on pr when they are both controversial, and unreviewed. It's also really bad timing for game changes right before a release.

AJenbo:
Ok, I see that most of this is related to the debug command, still not to excited about the self merge in this case.

obligaron (commented):
The split in two methods is a nice cleanup. 👍 
Overall looks good to me, only some edge cases.
But hey it's a debug command and a really helpful one. 😉

PR #540, opened 2020-01-15
Add fix to allow all unique items to drop

galaxyhaxz:
This adds an optional fix for the unique item generation bug, as well as a header file to toggle on/off optional fixes. I know that some people will want a pure vanilla experience, so it's disabled by default.

As an FYI, the bug involves unique items where more than one item of the same base type exists. The game will always pick the last occurring item in multi player, making it impossible to find any item but that. In single player the game records which uniques have already dropped so it is possible, but it starts from bottom to top making it extremely difficult since you would have to find 3-4+ uniques of the same base in the **_same_** game.

How this should be fixed can be argued, but interestingly there is an unused `random` call in the vanilla code. I fixed it by making the game use this call, so it picks one at random rather than the last one. Below is a screenshot of me finding both Lightforge and Constricting Ring in multiplayer:
DIABLO_20200115_024042

PR #2930, opened 2021-09-23
loot system fixes and tweaks

### LINKED ISSUE #7109: [Issue Report]: Unique rings and amulets drop are always the same.
### Operating System

Windows x64

### DevilutionX version

1.5.2

### Describe

Every drop of unique rings and amulets are always the same. Acolyte's Amulet and Gladiators Ring. Quest rings and amulets are correct. I have stacks of the Acolyte's Amulet and Gladiators Ring.

### To Reproduce

Kill unique bosses, find a unique ring or amulet and it is always the same.

### Expected Behavior

Maybe one of the many other rings or amulets?

### Additional context

_No response_

qndel:
Part 2 of the "oneliners that flip the game upside down"
Now required because of https://github.com/diasurgical/devilutionX/pull/2763
Turns out the previous loot fix actually makes some uniques unfindable as they were only findable because of that bug which we fixed. This will apply to all uniques dropped in devilutionx and they might morph into a different unique of the same base in vanilla - In my opinion that's acceptable as there's literally no way we can keep it compatible so there's no reason to even attempt to.
I think that along with #2763 this puts Diablo's loot in a nice state and I'd even consider the game almost playable ;)

- Needs a fat changelog entry!

#edit
Added 1 way compatibility - uniques from vanilla won't morph, but devx uniques will morph in vanilla - there's simply no way around that :P

It's worth considering if we wouldn't want to improve the rng for regular magic items as well, since this PR is introducing a way to tell the difference between vanilla items and devilutionx ones.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1480,6 +1480,7 @@ void CheckInvSwap(Player &player, inv_body_loc bLoc, int idx, uint16_t wCI, int 
 {
 	auto &item = Items[MAXITEMS];
 	memset(&item, 0, sizeof(item));
+	item.dwBuff = dwBuff;
 	RecreateItem(item, idx, wCI, seed, 0, (dwBuff & CF_HELLFIRE) != 0);
 
 	player.HoldItem = item;
@@ -1785,6 +1786,7 @@ int SyncPutItem(Player &player, Point position, int idx, uint16_t icreateinfo, i
 	if (idx == IDI_EAR) {
 		RecreateEar(item, icreateinfo, iseed, id, dur, mdur, ch, mch, ivalue, ibuff);
 	} else {
+		item.dwBuff = ibuff;
 		RecreateItem(item, idx, icreateinfo, iseed, ivalue, (ibuff & CF_HELLFIRE) != 0);
 		if (id != 0)
 			item._iIdentified = true;
@@ -1798,7 +1800,6 @@ int SyncPutItem(Player &player, Point position, int idx, uint16_t icreateinfo, i
 		item._iMinMag = minMag;
 		item._iMinDex = minDex;
 		item._iAC = ac;
-		item.dwBuff = ibuff;
 	}
 
 	item.position = position;
--- a/Source/items.h
+++ b/Source/items.h
@@ -165,6 +165,7 @@ enum icreateinfo_flag {
 enum icreateinfo_flag2 {
 	// clang-format off
 	CF_HELLFIRE = 1,
+	CF_NEW_UNIQ = 2,
 	// clang-format on
 };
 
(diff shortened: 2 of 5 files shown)


julealgon (dismissed):
Scary. I like it.

galaxyhaxz (dismissed):
LGTM. I tested against the PR I made 2 years ago (https://github.com/diasurgical/devilutionX/pull/540) and they both generate the exact same results. I still prefer my cleanup however as it makes the logic a bit more understandable.

Source/items.h
```diff
@@ -165,6 +165,7 @@ enum icreateinfo_flag {
 enum icreateinfo_flag2 {
 	// clang-format off
 	CF_HELLFIRE = 1,
+	CF_NEW_UNIQ = 2,
```
obligaron, line 168: Some thoughts...

I like the idea of adding a flag for the new behavior. We could also think a little ahead,

Previous we had discussions about the crap... sub-optimal random generator of vanilla.
We also also have PR's open that replaces them (#2084 and #2261).
But both PR's don't touch item generators, cause this would break compatibility.

But with a similar flag (or the same?) like you did, we could change this.
With this we would stay compatible with vanilla (vanilla to devilutionX) but still create new (better) items.
The drawback would be that the new items would morph if loaded in vanilla (like the uniques).

This could be outside of this scope of this pr. I wanted to mention it as food for thoughts. And cause if we combine this we could save one flag. 😁

qndel: yeah but I don't think it's the best time to mess around with all magic items :P uniques are enough for now + we got a whole uint32 to use, we don't have to save space for now : P

Source/inv.cpp
```diff
@@ -1476,6 +1476,7 @@ void CheckInvSwap(Player &player, BYTE bLoc, int idx, uint16_t wCI, int seed, bo
 {
 	auto &item = Items[MAXITEMS];
 	memset(&item, 0, sizeof(item));
+	item.dwBuff = dwBuff;
 	RecreateItem(item, idx, wCI, seed, 0, (dwBuff & CF_HELLFIRE) != 0);
```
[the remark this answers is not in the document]
qndel: no idea : P

qndel:
@AJenbo if this goes into 1.4.0, I believe https://github.com/diasurgical/devilutionX/pull/2763 should be reverted for 1.3.0. They really need to go together imo.

AJenbo:
Could you do so and then add it to this PR?

obligaron:
Additional Question: if we change the item generation in a 1 way compatible way. Should we also use a better random number generator for the new item generator? It's not backwards compatible anyway.

qndel:
> Additional Question: if we change the item generation in a 1 way compatible way. Should we also use a better random number generator for the new item generator? It's not backwards compatible anyway.

Can't see why not

PR #2932, opened 2021-09-23
Remove AvailableItems array, maintain ActiveItems instead

ephphatha:
This PR refactors Items to remove the redundant AvailableItems array, essentially splitting the ActiveElements (ActiveItems) array into a section for live entities and a section for "free" memory. (This is prep work for #2586). Edit: I thought this had already been done for Missiles for some reason, the same approach can be used there.

I also updated the save/load from file code to decouple the indexes in the save file from the position in the Items array, this allows us to keep save compatibility with vanilla which uses AvailableItems as a list of free indexes. The only real difficulty there was ensuring that dItems maps from the saved values to the appropriate runtime values.

The Item sync code for DeltaLoadLevel originally tried to delete items based on position in the replay/sync data instead of the actual location in the current client for some reason, this PR updates that code to only attempt to delete client side items that match the expected identifiers. The old logic would potentially leave items visible for one player if they were on a different floor and another player picks up an item, this change handles that situation fine. I am not sure if there are any other situations where desync would be possible.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1084,11 +1084,12 @@ void OpenCrypt()
 		NetSendCmdQuest(true, quest);
 }
 
-void CleanupItems(Item *item, int ii)
+void CleanupItems(int ii)
 {
-	dItem[item->position.x][item->position.y] = 0;
+	auto &item = Items[ii];
+	dItem[item.position.x][item.position.y] = 0;
 
-	if (currlevel == 21 && item->position == CornerStone.position) {
+	if (currlevel == 21 && item.position == CornerStone.position) {
 		CornerStone.item._itype = ItemType::None;
 		CornerStone.item._iSelFlag = 0;
 		CornerStone.item.position = { 0, 0 };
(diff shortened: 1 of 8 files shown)


AJenbo:
I'm guessing you might not have seen this in the chat :)

**Announcement:** I consider development to have shifted in to release mode up until 1.3.0 is out, which will be in no more then 5 weeks, but the sooner the better.

This means that changes that are not directly contributing to the 1.3.0 milestone will not be merged till after 1.3.0 has been released. This is done to avoid introducing any new bugs up until the release as well as help focus resources on solving any outstanding features or bugs that still block the release of 1.3.0. This way we can hopefully have quick and stable release and soon resume the work on cleaning up the code etc.

Current work list:
- Polish Androd port
- Polish ZeroTier intergration
- Validate network messages
- Make password/encryption optional
- Game list via ZeroTier
- Finish various translations
- Convert Unifont to PCX
- Test
- Fix bugs

Optional:
- ZeroTier support on Android
- Game list via devilutionx.net for TCP
- ZeroTier support on Switch

Additionally see https://github.com/diasurgical/devilutionX/milestone/5

AJenbo (approved):

AJenbo (approved):

PR #2933, opened 2021-09-24
Update portuguese translation

yuripourre:
This PR updates the pt_BR translation

The words on the char panel were all turned into lower case to keep it consistent (some words simply don't fit otherwise).

There is one word that doesn't fit which is the translation for Armor Class (_"Armadura"_).

I tried to improve some weapons and monsters' names, I hope someone better skilled could improve it in the future.


AJenbo:
image

AJenbo (approved):

PR #2935, opened 2021-09-24
Add vcpkg.json to the project

StephenCWills:
After installing and integrating vcpkg, CMake will use vcpkg.json to automatically determine dependencies for the project and place them in `${CMAKE_CURRENT_BINARY_DIR}/vcpkg_installed` during the configure step.

This behavior requires the `VCPKG_MANIFEST_MODE` CMake variable to be set, which will happen automatically for new builds during the configure step, based on the presence of the vcpkg.json file. In other words, users who already have build directories set up would have to delete their CMakeCache.txt files to make use of this.

By default, packages appear to be cached in `%localappdata%\vcpkg\archives` so packages don't need to be redownloaded or built once installed. This means that Debug and Release builds can share the same packages, and it's safe to delete build directories without taking a debilitating hit to build performance on the next build.

Translations and tests are included as features which can be optionally turned on by modifying the `VCPKG_MANIFEST_FEATURES` CMake variable. I added a bit of logic to CMakeLists.txt to automatically turn those on based on the `BUILD_TRANSLATIONS` and `RUN_TESTS` options. The former needed to be optional because it takes about an hour to install the gettext[tools] package from vcpkg. The latter ensures that the gtest dependency gets downloaded automatically when targeting the x64-Debug-UnitTests or x86-Debug-UnitTests CMake configurations.

---

See https://github.com/microsoft/vcpkg/blob/master/docs/users/manifests.md for more detailed information about vcpkg manifests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,11 @@ mark_as_advanced(DISABLE_STREAMING_SOUNDS)
 option(STREAM_ALL_AUDIO "Stream all the audio. For extremely RAM-constrained platforms.")
 mark_as_advanced(STREAM_ALL_AUDIO)
 
+# The gettext[tools] package takes a very long time to install
+if(CMAKE_TOOLCHAIN_FILE MATCHES "vcpkg.cmake$")
+  option(USE_GETTEXT_FROM_VCPKG "Add vcpkg dependency for gettext[tools] for compiling translations" OFF)
+endif()
+
 RELEASE_OPTION(CPACK "Configure CPack")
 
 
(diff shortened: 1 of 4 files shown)


StephenCWills:
It looks like CI is building the translations using gettext provided by Git for Windows. Perhaps `BUILD_TRANSLATIONS` isn't the best name for the CMake option. Maybe something like `USE_GETTEXT_FROM_VCPKG`?

julealgon:
> ...it takes about an hour to install the gettext[tools] package from vcpkg

What?! Is this really the case? That's insane!

tsunamistate:
@julealgon that's how long it takes to compile `gettext` from source :)
Lots of small processes launched by `msys`, plus AV (like Windows Defender) may slow the compilation process to a crawl

StephenCWills:
> What?! Is this really the case? That's insane!

Yeah, the reason my upgrade took two hours the other day is because I had installed both `gettext[tools]:x64-windows` and `gettext[tools]:x86-windows` a little while back when I was messing around trying to get translations to build. I verified it again on my laptop last night just to make sure it wasn't some messed up configuration on my desktop, and it also took an hour. I think it must be downloading the source code to build the whole tool suite.

On that note, there's actually an option to set up binary caching so developers on a team can share binaries to speed this kind of thing up. Since it does take so unbelievably long, it might be worth looking into setting up CI to cache some binaries for this project. That said, the CI is currently already using prebuilt binaries from Git for Windows, and I'd assume users are free to use those as well so that may just be a waste of CI resources.

AJenbo (approved):

PR #2938, opened 2021-09-25
Reuse items of the same level when rerolling Griswold's premiums after level up

ephphatha:
When the player levels up while in town the next time Griswold refreshes his premium item stock he'll keep the same items if they are still an appropriate level for the player.

When playing with Hellfire logic the old code ended up offering items of a lower level than intended in certain situations, this change ensures that the item levels match what would have been generated if the shop was doing a full refresh.

To test this you need to level up while in town then buy an item from Griswold's premiums, the shop only refreshes when you buy an item or leave town. Leaving town resets the town seed which changes all the items.

You should see the first four items disappear as the first three are too low level for the new player level, and the fourth is excess to needs. Four new items will be generated near the end of the list; two at +2 item level relative to the player (the fifth and sixth last items) and two at +3 item level relative to the player (the last two items).

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -312,38 +312,6 @@ _sfx_id ItemDropSnds[] = {
 	IS_FLARM,
 	IS_FLARM,
 };
-/** Maps from Griswold premium item number to a quality level delta as added to the base quality level. */
-int premiumlvladd[] = {
-	// clang-format off
-	-1,
-	-1,
-	 0,
-	 0,
-	 1,
-	 2,
-	// clang-format on
-};
-/** Maps from Griswold premium item number to a quality level delta as added to the base quality level. */
-int premiumLvlAddHellfire[] = {
-	// clang-format off
-	-1,
-	-1,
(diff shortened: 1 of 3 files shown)


Source/items.cpp
```diff
@@ -4461,6 +4429,35 @@ void SpawnSmith(int lvl)
 
 void SpawnPremium(int pnum)
 {
+	/** Maps from Griswold premium item number to a quality level delta as added to the base quality level. */
+	int premiumlvladd[] = {
```
ephphatha, line 4433:
```suggestion
	constexpr int premiumlvladd[] = {
```

ephphatha:
Looking at this there is one relatively simple change that could be made, moving the `if (numpremium < maxItems) {...}` block below the while loop lets the loop shift empty items off the front of the list in some cases. I thought this would change the items that get rolled at certain character levels but it seems to make no difference? haven't included it in this PR anyway since I was looking at rerolling items with a single pass if possible

PR #2943, opened 2021-09-25
Initial clean up of code related to creating items

ephphatha:
This is part of the ongoing cleanup for #2586. #2932 has effectively got `DeleteItem` to the point where it can be replaced by an STL container's `erase` method, the aim with this branch is to get `AllocateItem` ready to be replaced with a `push_back` equivalent.

To make it easier to reason about how items are spawned and dropped on the ground I've started working through the code responsible for creating items in general. This set of commits is mainly cleanups for items.cpp, using `const Item &` for functions which do not modify the argument and renaming some initialisation functions to match their purpose better.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -549,7 +549,7 @@ void RespawnDeadItem(Item *itm, Point target)
 
 	Items[ii] = *itm;
 	Items[ii].position = target;
-	RespawnItem(&Items[ii], true);
+	RespawnItem(Items[ii], true);
 
 	itm->_itype = ItemType::None;
 }
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -582,10 +582,8 @@ void DeadItem(Player &player, Item *itm, Displacement direction)
 	}
 }
 
-int DropGold(int pnum, int amount, bool skipFullStacks)
+int DropGold(Player &player, int amount, bool skipFullStacks)
 {
-	auto &player = Players[pnum];
-
 	for (int i = 0; i < player._pNumInv && amount > 0; i++) {
 		auto &item = player.InvList[i];
 
@@ -593,22 +591,20 @@ int DropGold(int pnum, int amount, bool skipFullStacks)
 			continue;
 
 		if (amount < item._ivalue) {
+			Item gold {};
+			InitializeItem(gold, IDI_GOLD);
+			SetGoldSeed(player, gold);
+
+			gold._ivalue = amount;
 			item._ivalue -= amount;
-			InitializeItem(player.HoldItem, IDI_GOLD);
-			SetGoldSeed(player, player.HoldItem);
-			SetPlrHandGoldCurs(player.HoldItem);
-			player.HoldItem._ivalue = amount;
-			DeadItem(player, &player.HoldItem, { 0, 0 });
+			SetPlrHandGoldCurs(gold);
+			DeadItem(player, &gold, { 0, 0 });
 			return 0;
 		}
 
 		amount -= item._ivalue;
+		DeadItem(player, &item, { 0, 0 });
 		player.RemoveInvItem(i);
-		InitializeItem(player.HoldItem, IDI_GOLD);
-		SetGoldSeed(player, player.HoldItem);
-		SetPlrHandGoldCurs(player.HoldItem);
-		player.HoldItem._ivalue = item._ivalue;
-		DeadItem(player, &player.HoldItem, { 0, 0 });
 		i = -1;
 	}
 
@@ -622,13 +618,12 @@ void DropHalfPlayersGold(int pnum)
 	}
 	auto &player = Players[pnum];
 
-	int hGold = player._pGold / 2;
+	int remainingGold = DropGold(player, player._pGold / 2, true);
+	if (remainingGold > 0) {
+		DropGold(player, remainingGold, false);
+	}
 
-	hGold = DropGold(pnum, hGold, true);
-	if (hGold > 0)
-		DropGold(pnum, hGold, false);
-
-	player._pGold -= hGold;
+	player._pGold /= 2;
 }
 
 void InitLevelChange(int pnum)
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -538,7 +538,7 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
 	player.spellLevel = GetSpellLevel(pnum, player._pSpell);
 }
 
-void RespawnDeadItem(Item *itm, Point target)
+void RespawnDeadItem(Item &itm, Point target)
 {
 	if (ActiveItemCount >= MAXITEMS)
 		return;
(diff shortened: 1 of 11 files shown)


AJenbo:
Something tells me I won't be able to just lean back once 1.3.0 is out :D

Source/player.cpp
```diff
@@ -538,31 +538,29 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
...
 {
-	if (itm->isEmpty())
+	if (itm.isEmpty())
 		return;
 
 	Point target = player.position.tile + direction;
 	if (direction != Displacement { 0, 0 } && ItemSpaceOk(target)) {
 		RespawnDeadItem(itm, target);
-		player.HoldItem = *itm;
+		player.HoldItem._itype = ItemType::None;
```
AJenbo, line 563:
```suggestion
		itm._itype = ItemType::None;
```

Source/player.cpp
```diff
@@ -573,7 +571,7 @@ void DeadItem(Player &player, Item *itm, Displacement direction)
 				Point next = player.position.tile + Displacement { i, j };
 				if (ItemSpaceOk(next)) {
 					RespawnDeadItem(itm, next);
-					player.HoldItem = *itm;
+					player.HoldItem._itype = ItemType::None;
```
AJenbo, line 574: This is making an assumption that HoldItem is what we recived
```suggestion
					itm._itype = ItemType::None;
```

AJenbo: It's also possible that this should instead be moved to the caller.

ephphatha: Yeah, I'm looking at this a bit more carefully since it didn't outright conflict with any changes since it was originally opened, but a lot of the lines have lost context and ended up significantly different than intended.

Source/player.cpp
ephphatha: I ended up rolling this back into RespawnDeadItem, it's meant to indicate that the item has been moved but because it's trivially copyable clang-tidy discourages actually using std::move :D

ephphatha: And back to the caller 🙃 . Tried to tidy things up so it doesn't always assume that the item being dropped needs to be invalidated, many of the call sites are temporaries anyway.

ephphatha:
Think I've finally got this branch as simple as I can while doing a couple of the remaining cleanups from when it was originally opened. Some of them got addressed in other PRs so it got to shrink a bit :D

Source/player.cpp
```diff
@@ -3214,12 +3207,14 @@ void StripTopGold(Player &player)
...
-				SetPlrHandItem(player.HoldItem, 0);
+				InitializeItem(player.HoldItem, IDI_GOLD);
 				SetGoldSeed(player, player.HoldItem);
 				player.HoldItem._ivalue = val;
 				SetPlrHandGoldCurs(player.HoldItem);
-				if (!GoldAutoPlace(player))
-					DeadItem(player, &player.HoldItem, { 0, 0 });
+				if (!GoldAutoPlace(player)) {
+					DeadItem(player, player.HoldItem, { 0, 0 });
+					player.HoldItem._itype == ItemType::None;
```
AJenbo, line 3216: This gets overwritten by line 3222 so not really needed. But eventually we should probably not use `player.HoldItem` as the temp item.

ephphatha: Not entirely sure why I didn't remove the use of `player.HoldItem` here, `DeadItem` no longer uses holditem as a temporary so I can create a new item instance instead

Source/player.cpp
```diff
@@ -547,23 +547,20 @@ void RespawnDeadItem(Item *itm, Point target)
 
 	dItem[target.x][target.y] = ii + 1;
 
-	Items[ii] = *itm;
+	Items[ii] = itm;
 	Items[ii].position = target;
-	RespawnItem(&Items[ii], true);
-
-	itm->_itype = ItemType::None;
+	RespawnItem(Items[ii], true);
```
AJenbo, line 552:
```suggestion
	RespawnItem(Items[ii], true);
	NetSendCmdPItem(false, CMD_RESPAWNITEM, target, Items[ii]);
```

Source/player.cpp
```diff
@@ -547,23 +547,20 @@ void RespawnDeadItem(Item *itm, Point target)
...
-	if (itm->isEmpty())
+	if (itm.isEmpty())
 		return;
 
 	Point target = player.position.tile + direction;
 	if (direction != Displacement { 0, 0 } && ItemSpaceOk(target)) {
 		RespawnDeadItem(itm, target);
-		player.HoldItem = *itm;
-		NetSendCmdPItem(false, CMD_RESPAWNITEM, target, player.HoldItem);
+		NetSendCmdPItem(false, CMD_RESPAWNITEM, target, itm);
```
AJenbo, line 563:
```suggestion
```

Source/player.cpp
```diff
@@ -573,42 +570,37 @@ void DeadItem(Player &player, Item *itm, Displacement direction)
 				Point next = player.position.tile + Displacement { i, j };
 				if (ItemSpaceOk(next)) {
 					RespawnDeadItem(itm, next);
-					player.HoldItem = *itm;
-					NetSendCmdPItem(false, CMD_RESPAWNITEM, next, player.HoldItem);
+					NetSendCmdPItem(false, CMD_RESPAWNITEM, next, itm);
```
AJenbo, line 573:
```suggestion
```

Source/player.cpp
```diff
@@ -547,23 +547,20 @@ void RespawnDeadItem(Item *itm, Point target)
 
 	dItem[target.x][target.y] = ii + 1;
 
-	Items[ii] = *itm;
+	Items[ii] = itm;
 	Items[ii].position = target;
-	RespawnItem(&Items[ii], true);
-
-	itm->_itype = ItemType::None;
```
AJenbo, line 554: It feels like luck that the original code actually worked :D

Could you move `NetSendCmdPItem` in here, I think that would make the code a lot more sensible and less fragile.

Source/player.cpp
```diff
@@ -3156,7 +3147,8 @@ StartPlayerKill(int pnum, int earflag)
 			drawhpflag = true;
 
 			if (pcurs >= CURSOR_FIRSTITEM) {
-				DeadItem(player, &player.HoldItem, { 0, 0 });
+				DeadItem(player, player.HoldItem, { 0, 0 });
+				player.HoldItem._itype = ItemType::None;
 				NewCursor(CURSOR_HAND);
```
AJenbo, line 3152: This actually highlights an issue, if a player is holding an item in the cursor on death the item will get deleted :o

We should probably try and drop the item first, but if you prefer we can wait with that till a later PR

Source/player.cpp
```diff
@@ -622,13 +614,12 @@ void DropHalfPlayersGold(int pnum)
...
-	hGold = DropGold(pnum, hGold, true);
-	if (hGold > 0)
-		DropGold(pnum, hGold, false);
+	int remainingGold = DropGold(player, player._pGold / 2, true);
+	if (remainingGold > 0) {
+		DropGold(player, remainingGold, false);
+	}
 
-	player._pGold -= hGold;
+	player._pGold /= 2;
```
AJenbo, line 622: Looks like you fixed an issue with how remaining gold was calculated here :)

Source/player.cpp
ephphatha: Because the auto-pickup gold code used holditem as a temporary as well -_-. I'll include that in the next push

AJenbo (approved):
Finally finished reviewing this :sweat: 

There are a couple of things that could be further improved with relatively little effort, but things looks good already :+1: 

I'll give you a bit to decide if you want to make the few suggested adjustments :)

ephphatha:
I'll do that as a separate PR, cleaning up the use of holdItem in gold auto-pickup is pretty involved and the deaditem deduplication is also worth splitting.

Source/player.cpp
ephphatha: yeah, the sum of gold in the inventory value was getting set to a value that didn't actually represent what the player was holding as hGold was being reused to determine how much additional gold to drop from the second pass. I don't think it mattered much in practice as the value is only used in shops and players can only die in the dungeon (normally :D).

PR #2947, opened 2021-09-25
[Virtual Gamepad] Add buttons for using potions from belt

StephenCWills:
Adds a couple more buttons to match the functionality of L1/R1 on a physical controller.

Thanks to the hardware rendering, it's easier to scale the potion graphics to fit in the gamepad buttons without it looking like trash. Here are some sample screenshots.

image

image

image

### DIFF
--- a/Source/controls/game_controls.cpp
+++ b/Source/controls/game_controls.cpp
@@ -126,6 +126,16 @@ bool GetGameAction(const SDL_Event &event, ControllerButtonEvent ctrlEvent, Game
 				*action = GameActionSendKey { DVL_VK_SPACE, false };
 			return true;
 		}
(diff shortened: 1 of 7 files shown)


AJenbo (approved):

yuripourre:
@StephenCWills Nice addition! Have you considered making the belt touchable?

StephenCWills:
> Have you considered making the belt touchable?

It's already touchable, but it only performs a mouse hover so you can examine the item using the info panel. I actually hadn't considered a direct tap-to-use interaction for the belt, but I would certainly like to improve the general touch interactions in areas of the screen outside the virtual gamepad. This is a good suggestion. Thanks!

PR #2950, opened 2021-09-25
Better text alignment in spell book

AJenbo:
image


ephphatha (approved):
LGTM :)

julealgon:
This one is a bit subjective to me. The previous layout (align on top instead of center and with less spacing between lines) actually lead to more separate entries (in general) and more readability IMHO.

This change here would go better with a separator between spells (say, a gold line).

AJenbo:
Some things that make it less subjective
- For some languages the letters would overlap
- Top entry has less space but only two lines

PR #2952, opened 2021-09-26
Show install guide to users that only run shareware

AJenbo:


android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -110,7 +110,11 @@ private void migrateFile(File file) {
 		if (!file.exists() || !file.canRead()) {
 			return;
 		}
+
 		File newPath = new File(externalDir + "/" + file.getName());
+		if (file.getName().equals("spawn.mpq"))
+			newPath = new File(externalDir + "/" + file.getName() + "-temp");
```
StephenCWills, line 116: If the file is read-only, it will be copied as `spawn.mpq-temp`, but then subsequently renamed to `spawn.mpq`. The next time the activity is created, this file would be copied again as `spawn.mpq-temp`. This is probably not a big deal, but just FYI.

StephenCWills (approved):

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
AJenbo: Looked at it for a bit and I think that since the chance of it being read only is very low (how did the old app put it there in the first place?), combined with this being removed by the time 1.4.0 rolls around I'm not to worried about this potentially eating 50mb of disk space.

PR #2953, opened 2021-09-26
Fix: Holding mouse when shift is held shouldn't lock to a target & Trigger Text is incorrectly shown

### LINKED ISSUE #2695: Shift clicking locks on original position of first target
**Describe**
When shift clicking on a monster/item/object the player will keep attack in the direction of the same tile.

**To Reproduce**
1. Go to dungeon
2. Fine an item or an object
3. Point to the item/object
4. Hold shift+click
5. Move the mouse around
6. Notice that you keep attacking in the same direction.

**Expected behavior**
The high light should not be locked when shift was activated, and attack should track the mouse movement.

### LINKED ISSUE #2859: Trigger text takes priority over item name
When mouse button is pressed down, trigger text is taking priority over item name as seen on the gif
triggertext

obligaron:
Fixes #2695
Fixes #2859

Notes:

- When the mouse is held cursor position only get updated if we previous didn't select an monster/object/item/... This was implemented to track a target. But this target tracking should activate if shift is pressed. So adjust the logic to only track if we want to.
- Shift+Spell needed more adjustment cause `CheckPlrSpell` sets `LastMouseButtonAction`. `CheckPlrSpell` is also used in initial mouse click (`RightMouseDown`) and for repeating mouse actions (`RepeatMouseAction`). Cause of this we must ensure that it regards shift/doesn't override `LastMouseButtonAction` when we are in `RepeatMouseAction`.
- The entire logic of this tracking target should only be performed if we have we want to track any target (is not `MouseActionType::None`). This fixes the inventory issue, where `CheckTrigForce` was called when mousebutton where held.
- Please test this. Tracking logic is ... special. 😉 

@FluffyQuack @AJenbo It would be nice if you both could take a look. 🙂

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1441,7 +1441,7 @@ void PerformSpellAction()
 	}
 
 	UpdateSpellTarget();
-	CheckPlrSpell();
+	CheckPlrSpell(false);
 }
 
 void CtrlUseInvItem()
--- a/Source/cursor.cpp
+++ b/Source/cursor.cpp
@@ -330,7 +330,7 @@ void CheckCursMove()
 	my = clamp(my, 0, MAXDUNY - 1);
 
 	// While holding the button down we should retain target (but potentially lose it if it dies, goes out of view, etc)
-	if (sgbMouseDown != CLICK_NONE && pcursinvitem == -1) {
+	if (sgbMouseDown != CLICK_NONE && pcursinvitem == -1 && IsNoneOf(LastMouseButtonAction, MouseActionType::Attack)) {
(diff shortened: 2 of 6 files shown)


FluffyQuack:
Those changes look good to me.

julealgon:
> Tracking logic is ... special.

😞

AJenbo (approved):

AJenbo:
@obligaron nice work, I think https://github.com/diasurgical/devilutionX/issues/2962 might be easy for you to tackle since it also involves this part of the code.

obligaron:
Yes same location, so opened #2964 🙂

PR #2954, opened 2021-09-26
Fix melee friendly fire when sgOptions.Gameplay.bFriendlyFire is set

### LINKED ISSUE #2772: Shift Click Melee Attacks Hurt Players While Player Friendly
**Important information**
Windows 10 Pro / DevilutionX 1.2.1 / Downloaded release

**Describe**
If you shift click to melee attack while player friendly, your attack can hit another player.

**To Reproduce**
1. Enter a multiplayer game with at least two players.
2. Go to dungeon level 1.
3. Position the players next to each other, and while player friendly hold shift and click on the other player.
4. Do this enough times and you'll see that you can damage the player.

**Expected behavior**
Melee attacks should NEVER be able to hurt another player while you are player friendly.

obligaron:
Fixes #2772

Notes:

- `gbFriendlyMode` toggles friendly mode (the UI button)
- `sgGameInitInfo.bFriendlyFire` added the possibility to also disable missile friendly fire when `gbFriendlyMode` is active.
- This was introduced with #1104 but the check was also (incorrectly) applied to melee attacks. For melee we don't need to check `sgGameInitInfo.bFriendlyFire` cause it's not a missile.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -1210,7 +1210,7 @@ bool DoAttack(int pnum)
 				m = -(dMonster[dx][dy] + 1);
 			}
 			didhit = PlrHitMonst(pnum, m);
-		} else if (dPlayer[dx][dy] != 0 && (!gbFriendlyMode || sgGameInitInfo.bFriendlyFire != 0)) {
+		} else if (dPlayer[dx][dy] != 0 && !gbFriendlyMode) {
 			BYTE p = dPlayer[dx][dy];
 			if (dPlayer[dx][dy] > 0) {
 				p = dPlayer[dx][dy] - 1;


AJenbo:
You know what, now that I think of it I changed melee to be consistent with how missiles work :D

obligaron:
But that's not what the wiki states the ini switch should do:

> [Friendly Fire=1](https://github.com/diasurgical/devilutionX/wiki/DevilutionX-diablo.ini-configuration-guide#friendly-fire1)
> 
> 0=disabled, 1=enabled; Allow arrow/spell damage between players in multiplayer even when the friendly mode is on.

And second if `sgGameInitInfo.bFriendlyFire` should be the same for melee and missiles, it would be a setting that enable/disable the capacity to toggle friendly fire. And if this is the case the UI button is useless and should be removed (and some other locations in code need to be updated). 😉

PR #2955, opened 2021-09-26
Draw belt item number for unusable scrolls

### LINKED ISSUE #856: Inconsistent behavior with hotkeys and belt items that cannot be used due to unmet requirements
When an item in the belt becomes red, i.e. it cannot be used due to requirements, the game refrains from drawing the number hotkey in the belt:

image

However, when pressing the hotkey (`3` in this case) the game will still try to execute it.

Even though red items cannot be moved to the belt by default, this situation can happen when the scroll is added to the belt in a usable state, and then equipment is removed making the player have less stats than needed to use it (thanks @qndel ) 

## Expected Behavior

One of:
1. The game should not move forward with the hotkey and do nothing when the player presses `3`
2. The game should still render the number hotkey even for items that cannot be currently used

obligaron:
Fixes #856

Initially I didn't want to fix this, cause it's a good first issue for newcomers. But someone added it to 1.3.0 milestone and time is running out. So now one less issue. 😁

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1245,7 +1245,6 @@ void DrawInvBelt(const Surface &out)
 		CelDrawItem(myPlayer.SpdList[i], out, position, cel, celFrame);
 
 		if (AllItemsList[myPlayer.SpdList[i].IDidx].iUsable
-		    && myPlayer.SpdList[i]._iStatFlag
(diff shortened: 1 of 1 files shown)


AJenbo:
Yeah everything on that list is free game at this point :)

AJenbo (approved):

PR #2958, opened 2021-09-26
Add PlayerId param to methods needed by couch-coop

yuripourre:


Source/control.cpp
```diff
@@ -467,9 +467,9 @@ void PrintSBookStr(const Surface &out, Point position, const char *text)
 	DrawString(out, text, { GetPanelPosition(UiPanels::Spell, { SPLICONLENGTH + position.x, position.y }), { 222, 0 } }, UiFlags::ColorWhite);
 }
 
-spell_type GetSBookTrans(spell_id ii, bool townok)
+spell_type GetSBookTrans(spell_id ii, bool townok, int playerId)
 {
-	auto &myPlayer = Players[MyPlayerId];
+	auto &myPlayer = Players[playerId];
```
AJenbo, line 472: Rename `myPlayer` to `player`

Source/control.cpp
```diff
@@ -864,37 +864,37 @@ void DrawPanelBox(const Surface &out, SDL_Rect srcRect, Point targetPosition)
 	out.BlitFrom(*pBtmBuff, srcRect, targetPosition);
 }
 
-void DrawLifeFlaskUpper(const Surface &out)
+void DrawLifeFlaskUpper(const Surface &out, int playerId)
```
AJenbo, line 867: Considering that this UI isn't multi user friendly I don't see the point in making it multi user capable?

AJenbo (changes requested):
I'm not sure all the changes here make sens, rather they just change any occurrence of MyPlayerId, and without changing the name of related variable.

I would suggest initially only changing what is needed to gain individual control of players, and then deal with the UI portion at a later point.

Source/control.cpp
yuripourre: Thanks!

Source/control.cpp
yuripourre: Right, I will convert the PR into draft, this is more like a POC to me. I am still creating a strategy to have the couch-coop working. I think I will be working on this until I have it fully working (including UI), then I will slice it into smaller manageable PRs.

Thanks for reviewing

AJenbo:
I think the preferred way to move forward with this is to start making player specific operations members of the player struct so that they operate with in the context of the given player. This will also naturally happen in a gradual (but a bit slow atm) fashion.

The UI is starting to be more dynamic which should make some of the work on coop support easier.

yuripourre:
@AJenbo this approach would be ideal, for sure. I see the code improving over time, I have a really good time watching the progress. I will probably try to implement it again in the future.

Issue #2071, opened 2021-05-27
Korean translation


AJenbo:
I have noticed that we have a fairly big Korean user base, but no one has started work on a Korean translation.

We still haven't started work on the necessary fonts, but it might still be a good time to work on it as there is a lot of text to cover.

If you are interested in working on this please see:
https://github.com/diasurgical/devilutionX/wiki/Translation-Guide

And maybe contact us via the chat (see readme) if you have any questions.

AJenbo:
@british-choi @sana2dang I have heard that you guys have worked on similar translations for other games, is this something that you might be interested in working on?

qqkookie:
I volunteer unless other suitable people are not interested. Is it right to translate the Translation/devilutionx.pot?

british-choi:
> @british-choi @sana2dang I have heard that you guys have worked on similar translations for other games, is this something that you might be interested in working on?

Sorry, I am busy now. so I will support other developers.

AJenbo:
@british-choi thanks for the reply :)

@qqkookie that would be great, yes that is the translation template.

qqkookie:
I worked on Korean translation for three days.  And current progress is about 4-5% of all quest/gossip  messages, so I guess  that it will take a month or so. long way to go.

AJenbo:
> I worked on Korean translation for three days. And current progress is about 4-5% of all quest/gossip messages, so I guess that it will take a month or so. long way to go.

Thanks for the update. Feel free to submit the changes gradually, that way others can follow the progress and give feedback. Maybe some will even help out with other parts of it.

nsm53project:
I have Korean translation data of Diablo that I translated 10 years ago(HaHa). I want to provide it soon.

The word order of Korean and English is different. I wonder how the combination of item names works in devilutionX.

AJenbo:
We do provide an option for simple swapping of prefix and item name, but for more advanced rules we will eventually add support for generating names via lua script.

nsm53project:
When will you start working on the necessary fonts? All characters are displayed as '???'.

AJenbo:
We are already looking in to way of generating the needed fonts, but it won't land till after Extended Lating and Cyrillic, which I hope will be done this month.

uwodb:
> When will you start working on the necessary fonts? All characters are displayed as '???'.

hangul font? I remember a question I asked 2 years ago :) I'll comment on your tistory,

AJenbo:
[...]

PR #2961, opened 2021-09-27
#2071 Korean translation part 6

bcko:
- part of #2071
- also contains @british-choi 's translation suggestion https://github.com/diasurgical/devilutionX/issues/2071#issuecomment-927532627


AJenbo:
Looks like you are making nice progress atm, I think I will leave this open for a bit so that you can add more lines to it :)

PR #2967, opened 2021-09-28
Korean translation 01

cholovesdata:
Translated 32 vocabs into Korean


AJenbo:
Thanks, I have merged it with the other updates to the Korean translation.

PR #2968, opened 2021-09-28
Validate CMD_SYNCDATA messages

AJenbo:
Checks that the CMD_SYNCDATA message won't cause OOB or CTD. CMD_SYNCDATA looks to be one of the more complicated messages and some of the changes here also lay the ground work for validating more messages so that we can more safely allow password-less games.

There are also some cleanups included since it helped me better understand the code and avoid the need for null pointer checks.

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -77,7 +77,7 @@ void FreePackets()
 
 void PrePacket()
 {
-	uint8_t playerId = -1;
+	uint8_t playerId = std::numeric_limits<uint8_t>::max();
 	for (TMegaPkt &pkt : MegaPktList) {
 		byte *data = pkt.data;
 		size_t spaceLeft = sizeof(pkt.data);
@@ -100,7 +100,17 @@ void PrePacket()
 				continue;
 			}
 
-			uint32_t pktSize = ParseCmd(playerId, (TCmd *)data);
+			if (playerId >= MAX_PLRS) {
+				Log("Missing source of network message");
+				return;
+			}
+
+			uint32_t size = ParseCmd(playerId, (TCmd *)data);
+			if (size == 0) {
+				Log("Discarding bad network message");
+				return;
+			}
+			uint32_t pktSize = size;
 			data += pktSize;
 			spaceLeft -= pktSize;
 		}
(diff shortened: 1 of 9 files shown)


Source/msg.cpp
```diff
@@ -89,6 +89,10 @@ void PrePacket()
 				data += sizeof(*cmd);
 				spaceLeft -= sizeof(*cmd);
 				playerId = cmd->bPlr;
+				if (playerId > MAX_PLRS) {
```
ephphatha, line 92:
```suggestion
				if (playerId >= MAX_PLRS) {
```

ParseCmd doesn't validate playerId before trying to use it as an index into Players

Source/msg.cpp
```diff
@@ -89,6 +89,10 @@ void PrePacket()
 				data += sizeof(*cmd);
 				spaceLeft -= sizeof(*cmd);
 				playerId = cmd->bPlr;
+				if (playerId > MAX_PLRS) {
+					Log("Invalid source of network message");
+					playerId = std::numeric_limits<uint8_t>::max();
```
qndel, line 94: why? playerId is invalid already, no need to make it invalid again

Source/msg.cpp
```diff
@@ -100,7 +104,17 @@ void PrePacket()
 				continue;
 			}
 
-			uint32_t pktSize = ParseCmd(playerId, (TCmd *)data);
+			if (playerId == std::numeric_limits<uint8_t>::max()) {
```
qndel, line 107: `playerId > MAX_PLRS` is more flexible and saves you that assignment in the change above

Source/sync.cpp
```diff
@@ -235,37 +281,40 @@ uint32_t sync_all_monsters(const byte *pbBuf, uint32_t dwMaxLen)
...
-		if (currlevel == pHdr->bLevel) {
-			SyncMonster(pnum, (TSyncMonster *)pbBuf);
+	assert(header.wLen % sizeof(TSyncMonster) == 0);
+
+	int monsterCount = header.wLen / sizeof(TSyncMonster);
+	const auto *monsterSyncs = reinterpret_cast<const TSyncMonster *>(pCmd + sizeof(header));
+
+	for (int i = 0; i < monsterCount; i++) {
+		uint8_t level = header.bLevel;
+		if (level >= NUMLEVELS)
```
qndel, line 304: why is this check in a loop? can't you just check it once before the loop?

qndel (approved):
🎉

AJenbo:
Thanks for the review 🙂 expect more of this stuff next week 😮

PR #2969, opened 2021-09-28
Use struct inheritance for network messages

AJenbo:


Source/msg.h
```diff
@@ -135,98 +135,62 @@ struct TCmd {
...
-	uint16_t wParam3;
+struct TCmdParam4 : TCmdParam3 {
 	uint16_t wParam4;
 };
 
-struct TCmdGolem {
-	_cmd_id bCmd;
-	uint8_t _mx;
-	uint8_t _my;
+struct TCmdGolem : TCmdLoc {
```
glebm, line 175: There are no guarantees about the memory layout here because inherited structs are no longer POD.
We have to be very careful with how we unpack these. Definitely can't just cast them from a network byte stream.

AJenbo: Sounds like it's best to just close this PR and leave it as is :)

julealgon:
Not a big fan of using inheritance for field sharing alone. I tend to consider this a code smell and overall poor use of inheritance. Semantically, it is also misleading since we are saying that "a Cmd4 _is_ a Cmd3" which is just not true.

AJenbo:
The main drive for this was sharing the `cmd` parameter, since it controls which other struct it is blindly cast to.

PR #2978, opened 2021-09-28
Ctrl Drop

### LINKED ISSUE #1358: Drop items from inventory/belt/equipment with CTRL+CLICK
## Description

Dropping multiple items from the inventory is currently a painful process, as one needs to move the cursor back and forth between the inventory screen and the play area, while keeping track of occupied space on the ground.

We can introduce a new shortcut to make this process less problematic, similar to how shift+click works today for moving items around.

## Details

- CTRL+Click on an item in the belt, inventory or equipment should immediately attempt to drop that item to the ground.
- Equipment restrictions should be consistent: if unequipping while attacking or moving is in place, the same should hold while attempting to drop it
- If all 9 tiles around the player are already filled with an item, the item should not be dropped (consistency with standard dropping behavior).
- Add info about command in help screen (similar to shift actions)
- Logic for dropping items should be consistent with the one used on gamepads (rely on the same function/behavior for consistency)
- Dropping special items should automatically target their object (cathedral map, rune bomb, etc)

Behavior should be standard: no need for feature flag

yuripourre:
In order to avoid clashing with the auto equip feature, auto equip will always take precedence

Closes #1358

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -404,7 +404,7 @@ void ReleaseKey(int vkey)
 		CaptureScreen();
 	if (vkey == DVL_VK_MENU || vkey == DVL_VK_LMENU || vkey == DVL_VK_RMENU)
 		AltPressed(false);
-	if (vkey == DVL_VK_CONTROL || vkey == DVL_VK_LCONTROL || vkey == DVL_VK_RCONTROL)
+	if (vkey == DVL_VK_RCONTROL)
 		ToggleItemLabelHighlight();
 }
 
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -308,6 +308,7 @@ void LeftMouseDown(int wParam)
 	}
 
 	bool isShiftHeld = (wParam & DVL_MK_SHIFT) != 0;
+	bool isCtrlHeld = (wParam & DVL_MK_CTRL) != 0;
 
 	if (!MainPanel.Contains(MousePosition)) {
 		if (!gmenu_is_active() && !TryIconCurs()) {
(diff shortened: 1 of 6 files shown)


Source/diablo.cpp
```diff
@@ -403,7 +404,7 @@ void ReleaseKey(int vkey)
 		CaptureScreen();
 	if (vkey == DVL_VK_MENU || vkey == DVL_VK_LMENU || vkey == DVL_VK_RMENU)
 		AltPressed(false);
-	if (vkey == DVL_VK_CONTROL || vkey == DVL_VK_LCONTROL || vkey == DVL_VK_RCONTROL)
+	if (vkey == DVL_VK_RCONTROL)
```
AJenbo, line 407: Is this change needed?

yuripourre: @AJenbo no, it's not needed although is very annoying dropping items while item highlight toggles.

yuripourre: I will revert so someone else can take the decision of what to do about it in a future PR.

AJenbo (approved):

Source/diablo.cpp
AJenbo: Highlight should be moved to a keymapper option any way

AJenbo:
Thanks, was worried this feature might not have made it :)

AJenbo:
I found it 70% annoying and 20% not configurable.

qndel:
Bad move imo, people ask for it in d2, it's added in tchernobog, it's just useful 🙄

AJenbo:
Yeah there should be an option for it, sometimes it's useful

PR #2966, opened 2021-09-28
Generate devilutionx.mpq

AJenbo:
This is probably not the best implementation but it's a working draft at the very least :) Feedback is extremely welcome.

Next step would be:
- adding translations to assets
- removing ttf,
- having platforms that provide there own container prefer that over the MPQ (Android APK)

PR #2979, opened 2021-09-28
Replace TTF with PCX font

AJenbo:
- Faster
- smaller (50kb vs 1600kb)
- less dependencies
- less code
- unicode support (the ttf font we used only had symbols for latin + cyrillic)
- Also fixes a bunch of edge cases for the error dialog

This also now lets the error dialog fall back to SDL dialogs in case no font was loaded (you still can get a blank error dialog if only the dialog font is missing, but it now prints to the console in addition).

Note that this contains a baked version of the font, so visually there are no changes.

P.s. I failed to remove `SDL_ttf` from `Packaging/windows/mingw-prep.sh` as it then starts complaining that it is unable to link zlib, I'm not really sure why that is happening or what needs to be added in it's place.

Build time for Mac, 3DS, and Switch seams to have been significantly reduced.


Source/DiabloUI/dialogs.cpp
```diff
@@ -166,56 +165,50 @@ void Init(const char *text, const char *caption, bool error, bool renderBehind)
...
 	if (!renderBehind) {
+		ArtBackground.Unload();
 		LoadBackgroundArt("ui_art\\black.pcx");
 		if (ArtBackground.surface == nullptr) {
 			LoadFallbackPalette();
+			if (SDL_ShowCursor(SDL_ENABLE) <= -1)
+				Log("{}", SDL_GetError());
 		}
 	}
-	SetFadeLevel(256);
```
glebm, line 194: Why are there changes to the fade level and background art in this file?

glebm:
This is very cool! Perhaps #2966 should go in first.

Source/DiabloUI/dialogs.cpp
AJenbo: I fixes a bunch of edge cases for the error dialog, they where more evident when working on the font code as they mostly manifest when there is a failure early on.

Specifically the fade level should always start out black as each dialog screen will slowly fade to full bright. The line here was causing the screen to flicker when the error log appeared.

glebm (approved):

Source/DiabloUI/dialogs.cpp
```diff
@@ -260,28 +255,33 @@ void UiOkDialog(const char *text, const char *caption, bool error, const std::ve
 {
 	static bool inDialog = false;
 
+	if (caption != nullptr) {
+		Log("{}", text);
+		LogError("{}", caption);
```
glebm, line 260:
```suggestion
		LogError("{}\n{}", caption, text);
```

julealgon:
Question: isn't TTF "more standard" when it comes to font/font rendering than using PCX?

At first glance, this feels like we are going backwards in extensibility and standards.

I don't know much about fonts though... asking to make sure.

AJenbo:
PCX font isn't even a standard. But TTF isn't serving us well, we can't rely on the OS and instead have to carry a large set of dependencies just to have basic features, and we still miss simple things like fallback fonts, or access to OS installed fonts.
We would have to use tools to manually merge several fonts in to an UBER font of several MB to provide support for multiple languages, and some systems are already struggling with the size of the font we where using, which only provided Latin and Cyrillic letters.

TTF is intended for outline fonts, but the game has a textured font, which led us to abandon using TTF for the in-game font. We don't have unilateral support for colored TTF fonts, and despite extensive searching we haven't been able to find free tools that would let us create one. 

Also it's exceedingly hard to find a font that provides good hints for rendering low resolution monochrome symbols, the latest version of the font we have dropped it and become almost unreadable if updated.

Editing fonts require specialized tools which very few people have experience with, letalone the expertise to produce high quality fonts.

Dropping TTF dependencies makes the game more portable, by only requiring a widely supported standard; pixels.

julealgon:
Thanks for the really nice summary @AJenbo . This looks rough..... I didn't know it could be this hard to work with fonts in a standardized manner. Figured this would've been "solved" by now.

PR #2982, opened 2021-09-29
Delete manual debian pakaging script

AJenbo:
Why remove this:
- It's replaced by the CPACK generated .deb
- It's a lost of files to update each time we make a release
- I have been receiving emails for months about this being broken for the Launchpad PPA, afaik it's stuck on 1.0.1, no one cares to fix it.

@Vitexus @muzena Speak Now or Forever Hold Your Peace :church:


muzena:
@AJenbo 
I tried to fix PPA build.
Problem with Launcpad building is that don't support downloading needed dependencies,
so I add manually asio, fmt, libpng, libzt, SDL2_image, SDL_audiolib, simpleini dependencies with -DFETCHCONTENT_SOURCE_DIR flag in [rules file](https://github.com/muzena/devilutionX/blob/fix-debian_versioning/Packaging/debian/rules), but building again failed
This is [build log](https://launchpadlibrarian.net/560854219/buildlog_ubuntu-impish-amd64.devilutionx_1.2.1+git3909-d00bd478-202109291454~ubuntu21.10.1_BUILDING.txt.gz)

AJenbo:
@muzena thanks for looking in to it, would it maybe be possible for us to change the PPA to be a binery repo where we push builds to from an external build service, rather then having thing build on Launchpad since it appears to be a really poor fit for what we are doing?

AJenbo:
`source_package.cmake` looks like a better solution for getting things on launchpad: https://robocomp-rajath.readthedocs.io/en/latest/packaging/

glebm:
@muzena Looks like the issue is with libzt.
Note that libzt's repo needs to be cloned **recursively** for the build to work (a GitHub-provided zip won't work, only a git clone will).

You also do not need to install the system version of libzt. DevilutionX currently requires and only uses our vendored patched version.

AJenbo:
we now produce a tarball that includes version and dependecies

PR #2984, opened 2021-09-29
Different map colors

qndel:
+ will allow adding seeing map explored by other players in multiplayer in the future


julealgon:
Add some pictures :)

qndel:
explorer

julealgon:
@yuripourre FYI (since you were looking into this earlier)

yuripourre:
@julealgon thanks for mentioning me!

AJenbo:
>     * (how many times have I mentioned that I hate anonymous namespaces?)

Enough that it is predictable and tedious

FitzRoyX:
This is a good use of color on the map. But for what reason do we need to differentiate between "shrine" and "others"? I guess I don't see the purpose of indicating what external force is acting on the map.

julealgon:
> But for what reason do we need to differentiate between "shrine" and "others"? I guess I don't see the purpose of indicating what external force is acting on the map.

I'd probably agree. I don't immediately see a point to knowing the source of any given automap tile. The main purpose was to differentiate "what I have personally seen" from "what is being told me but I haven't visited yet" which was the annoying part to me.

Source/automap.h
```diff
@@ -14,10 +14,21 @@
 
 namespace devilution {
 
+enum MapExplorationType : uint8_t {
```
julealgon, line 17: Why is this not an enum class?

Source/automap.cpp
```diff
@@ -518,7 +535,7 @@ std::unique_ptr<AutomapTile[]> LoadAutomapData(size_t &tileCount)
 } // namespace
 
 bool AutomapActive;
-bool AutomapView[DMAXX][DMAXY];
+uint8_t AutomapView[DMAXX][DMAXY];
```
julealgon, line 538: This should be typed to the enum instead of `uint8_t`

Source/automap.cpp
```diff
@@ -96,161 +96,178 @@ void DrawDiamond(const Surface &out, Point center, uint8_t color)
...
+
+	switch (explorationType) {
+	case MapExplorationType::MAP_EXP_SHRINE:
+		colorDim = PAL16_BLUE + 8;
+		colorBright = PAL8_BLUE;
+		break;
+	case MapExplorationType::MAP_EXP_SELF:
+	case MapExplorationType::MAP_EXP_NONE:
+		break;
+	}
```
julealgon, line 228: I'd recommend extracting this whole switch into a function and returning the colors from there.

Right now we have 2 different places setting the colors in here and in lines 216/217. We should centralize that in the switch.

Source/automap.cpp
```diff
@@ -96,161 +96,178 @@ void DrawDiamond(const Surface &out, Point center, uint8_t color)
...
 /**
  * @brief Renders the given automap shape at the specified screen coordinates.
  */
-void DrawAutomapTile(const Surface &out, Point center, AutomapTile tile)
+void DrawAutomapTile(const Surface &out, Point center, Point map)
 {
+	AutomapTile tile = GetAutomapTypeView(map);
+	uint8_t colorBright = MapColorsBright;
+	uint8_t colorDim = MapColorsDim;
+	MapExplorationType explorationType = static_cast<MapExplorationType>(AutomapView[map.x][map.y]);
```
julealgon, line 218: No casting needed if this is an enum class and the array is typed to it.

Source/automap.cpp
```diff
@@ -96,161 +96,178 @@ void DrawDiamond(const Surface &out, Point center, uint8_t color)
...
 	if (tile.HasFlag(AutomapTile::Flags::HorizontalDoor)) {
-		DrawMapVerticalDoor(out, { center.x + AmLine16, center.y + AmLine8 });
+		DrawMapVerticalDoor(out, { center.x + AmLine16, center.y + AmLine8 }, colorBright, colorDim);
 	} else {
-		DrawMapLineNE(out, { center.x, center.y + AmLine16 }, AmLine16, MapColorsDim);
+		DrawMapLineNE(out, { center.x, center.y + AmLine16 }, AmLine16, colorDim);
 	}
 }
 
+AutomapTile GetAutomapTypeView(Point map);
```
julealgon, line 208: This is a bit crazy to me.

Source/automap.cpp
```diff
@@ -707,15 +724,21 @@ void DrawAutomap(const Surface &out)
 	DrawAutomapText(out);
 }
 
-void SetAutomapView(Point position)
+void UpdateAutomapExplorer(Point map, MapExplorationType explorer)
+{
+	if (AutomapView[map.x][map.y] < explorer)
+		AutomapView[map.x][map.y] = explorer;
```
julealgon, line 730: I think we should be more explicit about this "hierarchy" of values you are implicitly creating here with the comparisons.

This feels a bit too magical to me.

Source/automap.h
```diff
@@ -14,10 +14,21 @@
...
+	/** unexplored map tile */
+	MAP_EXP_NONE,
+	/** map explored by a shrine */
+	MAP_EXP_SHRINE,
+	/** map tile explored by someone else in multiplayer */
+	MAP_EXP_OTHERS,
+	/** map tile explored by current player */
+	MAP_EXP_SELF,
+};
```
julealgon, line 26: I think we should rethink this a bit.

First, I'd rename it to something like `AutomapTileExplorationState`.

Then I think we should consider different names for the entries, and having it as a flags enum:
- `Uncharted` for the default value. This makes sense to me considering we are talking about a map
- `Charted` to indicate if the tile is known to the player (the shrine effect would open up tiles in this state)
- `Explored` to mean a tile that has been visited

If later on we wanna keep track of "who explored", I think this should be converted into a discriminated union using `variant<T>` and adding a field to the `Explored` state to point to the player who found it.

Source/automap.h
```diff
@@ -77,10 +88,15 @@ void AutomapZoomOut();
  */
 void DrawAutomap(const Surface &out);
 
+/**
+ * @brief Updates automap explorer at point if value is higher than existing.
+ */
+void UpdateAutomapExplorer(Point map, MapExplorationType explorer);
```
julealgon, line 94: A `map` is not a `Point`. I suggest renaming the argument:
Also suggest renaming the method to reflect the enum:
```suggestion
void UpdateAutomapExplorationState(Point mapPosition, AutomapTileExplorationState explorationState);
```

Source/automap.h
qndel: because I'm not going to cast this shit everywhere

Source/automap.cpp
qndel: then casting is needed in 100 other places

Source/automap.cpp
qndel: why? higher "explorer" takes priority = 
none < shrine < other players in multi < self
to provide proper map coloring when you explore map with a different color already

Source/automap.cpp
qndel: nah, it's good, this is the only function that needs the colors, no need to artifically create a new one then mess around with returning 2 values and other stuff

Source/automap.h
qndel: I see no reason to separate it like this 🤷 using current system it's easy to add new exploration types and colors

Source/automap.cpp
qndel: Either that or move 2 functions :D

Source/automap.h
qndel: Your point is invalid
image
xD

qndel:
multiexplorer

yuripourre:
@qndel the multiplayer feature is so cool!

qndel:
> @qndel the multiplayer feature is so cool!
last unused remnants of infernity :D

NiteKat:
> > But for what reason do we need to differentiate between "shrine" and "others"? I guess I don't see the purpose of indicating what external force is acting on the map.
> 
> I'd probably agree. I don't immediately see a point to knowing the source of any given automap tile. The main purpose was to differentiate "what I have personally seen" from "what is being told me but I haven't visited yet" which was the annoying part to me.

The difference I see is what was given to me via shrine probably has monsters and chests/barrels to investigate. What was given to me by other players is probably clear already and doesn't have those things.

Source/automap.h
julealgon: I don't see any reason to cast this ever though?

qndel: Saving/Loading, comparing if the value is higher for overwriting etc.

obligaron:
> The difference I see is what was given to me via shrine probably has monsters and chests/barrels to investigate. What was given to me by other players is probably clear already and doesn't have those things.

I agree. Currently the PR uses light blue for shrine and red for multiplayer?
If yes, I would vote for swapping the colors. Cause the shrine "explored" are dangerous => red and my friend should have cleared everything => light blue. 😉

qndel:
> > The difference I see is what was given to me via shrine probably has monsters and chests/barrels to investigate. What was given to me by other players is probably clear already and doesn't have those things.
> 
> I agree. Currently the PR uses light blue for shrine and red for multiplayer? If yes, I would vote for swapping the colors. Cause the shrine "explored" are dangerous => red and my friend should have cleared everything => light blue. 😉

While I agree with this approach, the color from other players exploring the map will be the one you see much more often than the "explore all map" shrine, or at least I think so, and red has a better contrast 🤷

ThomasChr:
grafik
I don't see any garbage commits :)

PR #2985, opened 2021-09-30
Prevent ZeroTier race condition during application shutdown

### LINKED ISSUE #2197: zerotier error
**Important information**
latest - https://github.com/diasurgical/devilutionX/commit/b478d8295c41de0475d925488016d3e08552b39b

**Describe**
no idea
image

**To Reproduce**
enter zerotier menu
select hero
select create game
select any difficulty
go back
select join game
spam esc till game quits 
ERROR

**Expected behavior**
No error

StephenCWills:
Removes the call to `zts_stop()` based on this excerpt from the libzt project README.

> At the end of your program or when no more network activity is anticipated, the user application can shut down the service with `zts_stop()`. However, it is safe to leave the service running in the background indefinitely as it doesn't consume much memory or CPU while at idle. `zts_stop()` is a non-blocking call and will itself issue a series of events indicating that various aspects of the ZeroTier service have successfully shut down.

This resolves #2197

### DIFF
--- a/Source/dvlnet/zerotier_native.cpp
+++ b/Source/dvlnet/zerotier_native.cpp
@@ -25,7 +25,6 @@ static constexpr uint64_t ZtNetwork = 0xaf78bf943649eb12;
 
 static std::atomic_bool zt_network_ready(false);
 static std::atomic_bool zt_node_online(false);
-static std::atomic_bool zt_started(false);
 static std::atomic_bool zt_joined(false);
 
 static void Callback(struct zts_callback_msg *msg)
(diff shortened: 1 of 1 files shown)


AJenbo (approved):

AJenbo:
If anywhere this should have been in `DiabloDeinit()` since `atexit()` can be problematic.

StephenCWills:
> If anywhere this should have been in `DiabloDeinit()` since `atexit()` can be problematic.

Based on the excerpt from the README, I think the most appropriate place to call `zts_stop()` would be when we exit the ZeroTier menu in the UI. That way, it wouldn't be running in the background if you switch to Single Player. But it also sounds like there isn't much benefit to doing that. I think if we can get ZT running on 3DS, then it might be worth looking into `zts_stop()` again.

PR #2986, opened 2021-09-30
Fix quater damage in Hellfire

AJenbo:
Quater damage is a negative value and so would be dealing damage based on garbage values...


StephenCWills:
Should we also move the bounds check up so it doesn't try to read `player._pLevel` from a bad memory location?

julealgon (approved):

StephenCWills (approved):

PR #2989, opened 2021-09-30
Generate MPQ during build phase

AJenbo:


CMakeLists.txt
```diff
@@ -762,12 +762,12 @@ set(devilutionx_assets
...
 if(SMPQ)
-  file(REMOVE "Packaging/resources/devilutionx.mpq")
-  execute_process(
+  add_custom_target(build_mpq ALL
     WORKING_DIRECTORY "${DevilutionX_SOURCE_DIR}/Packaging/resources/assets"
-    COMMAND ${SMPQ} -M 1 -C PKWARE -c "../devilutionx.mpq" ${devilutionx_assets})
-  # Copy devilutionx.mpq to the build directory so the game can run from the build directory
-  file(COPY "Packaging/resources/devilutionx.mpq" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+    COMMAND ${CMAKE_COMMAND} -E remove -f "${CMAKE_CURRENT_BINARY_DIR}/devilutionx.mpq"
+    COMMAND ${SMPQ} -M 1 -C PKWARE -c "${CMAKE_CURRENT_BINARY_DIR}/devilutionx.mpq" ${devilutionx_assets}
```
glebm, line 768: 1. I believe you need to list `${devilutionx_assets}` as dependencies so that this target is rebuilt whenever the assets change.
2. `"${CMAKE_CURRENT_BINARY_DIR}/devilutionx.mpq"` is the output, not a byproduct.

AJenbo: Will it also check if the actually assets changed or only if files where added/removed?

glebm: It will check the timestamps. From [the docs](https://cmake.org/cmake/help/latest/command/add_custom_command.html#command:add_custom_command):

> In makefile terms this creates a new target in the following form:
>
> ```
> OUTPUT: MAIN_DEPENDENCY DEPENDS
>         COMMAND
> ```

glebm: See here for the canonical example(s)

https://cmake.org/cmake/help/latest/command/add_custom_command.html#examples-generating-files

StephenCWills: FYI, the canonical example is probably sufficient, but you can also refer to the ffmpeg command used by the 3DS build.

https://github.com/diasurgical/devilutionX/blob/ff9affa5137f8731465b42ae45342d414363ef32/CMakeLists.txt#L1077-L1080

isage (approved):

Issue #2991, opened 2021-09-30
Player can end up without light


qndel:
![image](https://user-images.githubusercontent.com/14297035/135419131-586b0058-9abe-440a-b781-afbc909888a5.png)
Introduced by https://github.com/diasurgical/devilutionX/commit/6788d6a351d1a009a770a36f77690d07a2c9e7fb
Player 1 goes to lvl 1
Player 2 enters the game
Player 2 enters lvl 1
Player 2 steals light from player 1 on player 1's screen till player 1 moves

Tested both with regular walking and `changelevel 1`

Important thing - player 2 has to enter the game after player 1 entered the dungeon, if they are both in town at the same time and enter dungeon, the bug won't happen = there's some info that player 1 doesn't get from player because of not being in town when player 2 joins.
Doesn't happen in vanilla

PR #2992, opened 2021-09-30
fix light stealing in multi

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/2991
OnPlayerJoinLevel was setting light of other players to -1 too late, StartStand happened earlier and it updated light, which was 0 = light stealing.


AJenbo:
It would be nice if you could at least give me time enough to get off of work before merging with no review

qndel:
> It would be nice if you could at least give me time enough to get off of work before merging with no review

Was simple enough and I tracked the cause of this bug down to my knockback light fix so was no reason to hold back :P

PR #2993, opened 2021-09-30
WordWrapString: Wrap on punctuation

glebm:
Useful for Chinese and Japanese, which do not have spaces between words, nor whitespace after punctuation.

### DIFF
--- a/Source/DiabloUI/selconn.cpp
+++ b/Source/DiabloUI/selconn.cpp
@@ -112,7 +112,8 @@ void SelconnFocus(int value)
 	}
 
 	strncpy(selconn_MaxPlayers, fmt::format(_("Players Supported: {:d}"), players).c_str(), sizeof(selconn_MaxPlayers));
-	WordWrapString(selconn_Description, DESCRIPTION_WIDTH);
+	const std::string wrapped = WordWrapString(selconn_Description, DESCRIPTION_WIDTH);
+	strncpy(selconn_Description, wrapped.data(), sizeof(selconn_Description) - 1);
 }
 
 void SelconnSelect(int value)
--- a/Source/error.cpp
+++ b/Source/error.cpp
@@ -35,8 +35,7 @@ void InitNextLines()
 	char tempstr[1536]; // Longest test is about 768 chars * 2 for unicode
 	strcpy(tempstr, message.data());
 
-	WordWrapString(tempstr, LineWidth, GameFont12, 1);
-	const string_view paragraphs = tempstr;
+	const std::string paragraphs = WordWrapString(tempstr, LineWidth, GameFont12, 1);
 
 	size_t previous = 0;
 	while (true) {
--- a/Source/panels/charpanel.cpp
+++ b/Source/panels/charpanel.cpp
@@ -199,16 +199,19 @@ void DrawPanelField(const Surface &out, Point pos, int len)
 
 void DrawShadowString(const Surface &out, const PanelEntry &entry)
 {
-	if (entry.label == "")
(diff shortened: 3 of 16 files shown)


Source/DiabloUI/selconn.cpp
```diff
@@ -112,7 +112,8 @@ void SelconnFocus(int value)
 	}
 
 	strncpy(selconn_MaxPlayers, fmt::format(_("Players Supported: {:d}"), players).c_str(), sizeof(selconn_MaxPlayers));
```
AJenbo, line 114: Shoudn't we be able to remove this string copy and instead use the string directly as the input?

Source/error.cpp
```diff
@@ -35,8 +35,8 @@ void InitNextLines()
 	char tempstr[1536]; // Longest test is about 768 chars * 2 for unicode
 	strcpy(tempstr, message.data());
 
-	WordWrapString(tempstr, LineWidth, GameFont12, 1);
-	const string_view paragraphs = tempstr;
+	const std::string wrapped = WordWrapString(tempstr, LineWidth, GameFont12, 1);
+	const string_view paragraphs = wrapped;
```
AJenbo, line 39: Is there a need for a conversion to string_view here?

Source/panels/charpanel.cpp
```diff
@@ -199,16 +199,19 @@ void DrawPanelField(const Surface &out, Point pos, int len)
...
-	std::string text(buffer);
+	constexpr int Spacing = 0;
+	const std::string &textStr = LanguageTranslate(entry.label.c_str());
+	string_view text;
+	std::string wrapped;
+	if (entry.labelLength > 0) {
+		wrapped = WordWrapString(textStr, entry.labelLength, GameFont12, Spacing);
+		text = wrapped;
+	} else {
+		text = textStr;
```
AJenbo, line 213: I don't think labels with 0 length is allowed /used

AJenbo (commented):
Looks good, sorry for my bit ignorant review but wasn't in a position to gather much context 😅

Source/error.cpp
glebm: Yes, so that the `substring` call below is zero-copy.

Source/DiabloUI/selconn.cpp
glebm: Not sure what you mean exactly but sounds like something that should go into a separate refactoring.

Source/error.cpp
glebm: Although `TextLines` is an array of `std::string`, so it doesn't matter

Source/panels/charpanel.cpp
glebm: From the code it seems length 0 means "Do not wrap". Sounds like this change should be made separately though.

Source/error.cpp
glebm: Done

Source/DiabloUI/selconn.cpp
AJenbo: Arh my mistake, I thought this was the text that was being wordwrapped on the next line. But it's being copied to `selconn_MaxPlayers` not `selconn_Description`.

Source/panels/charpanel.cpp
AJenbo: Right, I can see now that you added the `else` as a substitution for the `strcpy`, not as a new new condition.

AJenbo (approved):
Nice, thanks for tackling this :+1:

PR #2997, opened 2021-09-30
Enhance ctrl + drop

### LINKED ISSUE #2988: Ctrl+drop improvment
- Ctrl+clicking and item makes the mouse jump.
- When there is no more room the player should say `WhereWouldIPutThis`

yuripourre:
Resolve #2988


Source/inv.cpp
```diff
@@ -888,7 +888,12 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 	}
 
 	if (dropItem) {
-		TryDropItem();
+		if (TryDropItem()) {
+			// Reset cursor position
+			SetCursorPos(cursorPosition);
```
julealgon, line 893: This looks hackish to me. Why does `TryDropItem` change the cursor in the first place?

We should avoid having to "revert" things and instead "just not do them in the first place". Otherwise, the code becomes harder and harder to reason about.

Source/inv.cpp
```diff
@@ -888,7 +888,12 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 	}
 
 	if (dropItem) {
-		TryDropItem();
+		if (TryDropItem()) {
+			// Reset cursor position
+			SetCursorPos(cursorPosition);
+		} else {
+			player.Say(HeroSpeech::WhereWouldIPutThis);
```
julealgon, line 895: Shouldn't this go inside the `TryDropItem` method instead?

Where else is that method being used today? Do we emit the same sound in there?

julealgon: Should we consider randomizing this with something like `ICantDoThat`?

Source/inv.cpp
yuripourre: TryDropItem doesn't move the cursor, auto equip does. I didn't want to change too much, but I am open to suggestions.

Source/inv.cpp
yuripourre: > Shouldn't this go inside the `TryDropItem` method instead?
> 
> Where else is that method being used today? Do we emit the same sound in there?

@julealgon this method is called in a few places, but currently, only misc_msg.cpp `BlurInventory()` plays the audio if the result is false.

Source/inv.cpp
yuripourre: @julealgon I probably could check `if (!IsHardwareCursor() && ! dropItem) {` but I prefer having the dropItem feature handled in a single place, instead of spreading the flag.

julealgon: This is bad.... IMHO, we need to rethink this to avoid the cursor change altogether. I'll take a closer look later. Ideally, we'd want to split InvCut into multiple methods, one for each action.

Source/inv.cpp
julealgon: It seems wrong to me that we are not (apparently) consistently playing the same audio in the same situations. Maybe @AJenbo can comment here since (I think?) `TryDrop` was created for gamepad controls initially.

AJenbo: Yes, make it always say the line when it fails.

Source/inv.cpp
obligaron: I agree with @julealgon
We shouldn't change cursPosition at all. I think that is also not great in the current controller support.

To advance we could (as a first step):

- Add a parameter (for example `Point dropLocation`) to `DropItemBeforeTrig` and `TryInvPut`. And use this instead of `cursPosition`.
- Move `TryDropItem` to `inv.cpp`, cause now it is not only used for controller support.

PR #2998, opened 2021-10-01
Restore original Warrior death sound

StephenCWills:
One of the changes to player.cpp in ec74917 unified the logic to play the death sound by switching from `sfx/misc/dead.wav` to `sfx/warrior/warior71.wav`, but this changes the sound that gets played when the Warrior dies. This change represents a compromise, keeping the logic unified by introducing a new index in the `herosounds` array that plays with the `HeroSpeech::Death` enum value.


julealgon (approved):
Very clean approach.

Source/effects.h
```diff
@@ -116,7 +116,8 @@ enum class HeroSpeech {
 	MaybeItsLockedFromTheInside,
 	LooksLikeItsRustedShut,
 	MaybeTheresAnotherWay,
-	LAST = MaybeTheresAnotherWay
+	Death,
```
AJenbo, line 119: One issue here, all enums here are verbatim the content of the audio, and the player does not say "death" when calling this one 🙂


Preferably transcribe the death sound the warrior makes here.

StephenCWills: I had considered that. For some reason, I was thinking that it wasn't applicable to this sound effect since the sound clip doesn't match up between the classes, but it occurs to me now that these are all based on the Warrior sound clips anyway.

AJenbo (approved):

PR #3004, opened 2021-10-01
CI: Manually build & install SMPQ on Vita CI

glebm:


AJenbo:
At some point it might be worth spinning this of in to it's own repo where we can do re release build that can then be fetched by the Vita builder, to avoid having to build it multiple times, and maybe to the benefit of others

glebm:
> At some point it might be worth spinning this of in to it's own repo where we can do re release build that can then be fetched by the Vita builder, to avoid having to build it multiple times, and maybe to the benefit of others

That is literally what a package is 😂

AJenbo:
Well i felt it was more polite to write a few more words then that :)

AJenbo:
looks like success :taco:

AJenbo (approved):
Thank! Now we get to enjoy auto-merges again :)

PR #3006, opened 2021-10-01
pixel perfect npc selection

qndel:
Original
originalselection

New
pixelperfectselection

### DIFF
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -740,13 +740,17 @@ void DrawMonsterHelper(const Surface &out, Point tilePosition, Point targetBuffe
 
 	if (leveltype == DTYPE_TOWN) {
 		auto &towner = Towners[mi];
+		assert(towner._tAnimData);
 		int px = targetBufferPosition.x - CalculateWidth2(towner._tAnimWidth);
 		const Point position { px, targetBufferPosition.y };
-		if (mi == pcursmonst) {
-			CelBlitOutlineTo(out, 166, position, CelSprite(towner._tAnimData, towner._tAnimWidth), towner._tAnimFrame);
-		}
-		assert(towner._tAnimData);
+		selectionState = selection_state::SELECTION_INIT;
 		CelClippedDrawTo(out, position, CelSprite(towner._tAnimData, towner._tAnimWidth), towner._tAnimFrame);
+		if (selectionState == selection_state::SELECTION_DISPLAY) {
+			pcursmonst = mi;
+			CelBlitOutlineTo(out, 166, position, CelSprite(towner._tAnimData, towner._tAnimWidth), towner._tAnimFrame);
+			CelClippedDrawTo(out, position, CelSprite(towner._tAnimData, towner._tAnimWidth), towner._tAnimFrame);
+		}
+		selectionState = selection_state::SELECTION_NONE;
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


julealgon:
Is there any particular reason we can't extend this to all entities, like objects and monsters, as well?

Or was it just an attempt to limit the scope of the change for now and add it incrementally to other parts of the game?

Source/scrollrt.cpp
```diff
@@ -740,13 +740,17 @@ void DrawMonsterHelper(const Surface &out, Point tilePosition, Point targetBuffe
 
 	if (leveltype == DTYPE_TOWN) {
 		auto &towner = Towners[mi];
+		assert(towner._tAnimData);
 		int px = targetBufferPosition.x - CalculateWidth2(towner._tAnimWidth);
 		const Point position { px, targetBufferPosition.y };
-		if (mi == pcursmonst) {
+		checkSelection = true;
+		CelClippedDrawTo(out, position, CelSprite(towner._tAnimData, towner._tAnimWidth), towner._tAnimFrame);
+		if (!checkSelection) {
```
StephenCWills, line 748: Why does `false` mean the entity is selected? I feel this is highly unintuitive.

qndel: I know :D but if I start with true and only change it if there's no selection, there's no guarantee that line memcpy function got called at all ... and atm while the names aren't intuitive, I can be sure it works properly

qndel:
> Is there any particular reason we can't extend this to all entities, like objects and monsters, as well?
> 
> Or was it just an attempt to limit the scope of the change for now and add it incrementally to other parts of the game?



> Is there any particular reason we can't extend this to all entities, like objects and monsters, as well?
> 
> Or was it just an attempt to limit the scope of the change for now and add it incrementally to other parts of the game?

Sure we can but when it comes to aiming at monsters in the middle of combat, we are entering an area where some people might've had years to get used to the original behavior

Basically I've added it only for NPCs because atm can't see the benefit of doing it for other things - if I added it for items, I bet people would have a lot of fun trying to click that ring ;)

Source/scrollrt.cpp
StephenCWills: It looks to me like it should still work if you start with false and set it to true instead.

StephenCWills:
> I think doing it for objects could be nice too - can even add some error margin to avoid requiring pixel perfect precision on lets say doors (remember cave doors got transparent parts in the middle)

I've also seen complaints about shrines.

qndel:
> > I think doing it for objects could be nice too - can even add some error margin to avoid requiring pixel perfect precision on lets say doors (remember cave doors got transparent parts in the middle)
> 
> I've also seen complaints about shrines.

Looks like pixel selection for objects is problematic because they respond to light and when far away, they are almost completely black, even at medium range it's hard to select, so I'd have to go with allowing selecting shadows :S

qndel:
I managed to add a pixel perfect selection for objects but it requires drawing them twice

AJenbo:
> I managed to add a pixel perfect selection for objects but it requires drawing them twice

My only issue with this PR is that it relies on the render for this logic. Instead we should have a function that can do hit detection on a given sprite frame, independent of rendering.
Tying rendering to hit detection would become a problem when implementing a hardware render where memory and logic can't be shared in a direct way.

AJenbo:
:coffin:

PR #3008, opened 2021-10-02
Validate more network messages

AJenbo:
- Mark network messages as `const`
- Validate coordinates for `OnWalk` and `OnGotoGetItem` to avoid OOB
- Use C++ in `OnAddStrength`, `OnAddMagic`, `OnAddDexterity`, and `OnAddVitality`

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -168,7 +168,7 @@ int WaitForTurns()
 	return 100 * sgbDeltaChunks / MAX_CHUNKS;
 }
 
-byte *DeltaExportItem(byte *dst, TCmdPItem *src)
+byte *DeltaExportItem(byte *dst, const TCmdPItem *src)
 {
 	for (int i = 0; i < MAXITEMS; i++, src++) {
 		if (src->bCmd == CMD_INVALID) {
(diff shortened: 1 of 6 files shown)


Source/multi.cpp
```diff
@@ -463,49 +465,49 @@ bool InitMulti(GameData *gameData)
...
 
-void multi_msg_add(byte *pbMsg, BYTE bLen)
+void multi_msg_add(const byte *data, size_t size)
 {
-	if (pbMsg != nullptr && bLen != 0) {
-		tmsg_add(pbMsg, bLen);
+	if (data != nullptr && size != 0) {
+		tmsg_add(data, size);
 	}
 }
```
vlolteanu, line 473: I think we can get rid of the checks, or make them asserts. They're never false.

Source/msg.cpp
```diff
@@ -614,81 +614,83 @@ void NetSendCmdExtra(TCmdGItem *p)
...
 
-DWORD OnWalk(TCmd *pCmd, Player &player)
+DWORD OnWalk(const TCmd *pCmd, Player &player)
 {
-	auto *p = (TCmdLoc *)pCmd;
+	const auto &message = *reinterpret_cast<const TCmdLoc *>(pCmd);
+	const Point position { message.x, message.y };
 
-	if (gbBufferMsgs != 1 && currlevel == player.plrlevel) {
+	if (gbBufferMsgs != 1 && currlevel == player.plrlevel && InDungeonBounds(position)) {
```
vlolteanu, line 622: If a message is invalid, I'd argue in favor of disconnecting the the other player, if hosting, or leaving the game, otherwise. Ignoring it simply sweeps a potential bigger problem under the rug, be it a bug, a corrupted/altered network packet potentially desyncing everything, a malicious client, or anything in between.

Source/msg.cpp
```diff
@@ -614,81 +614,83 @@ void NetSendCmdExtra(TCmdGItem *p)
...
-DWORD OnGotoGetItem(TCmd *pCmd, Player &player)
+DWORD OnGotoGetItem(const TCmd *pCmd, Player &player)
 {
-	auto *p = (TCmdLocParam1 *)pCmd;
+	const auto &message = *reinterpret_cast<const TCmdLocParam1 *>(pCmd);
+	const Point position { message.x, message.y };
 
-	if (gbBufferMsgs != 1 && currlevel == player.plrlevel) {
-		MakePlrPath(player, { p->x, p->y }, false);
+	if (gbBufferMsgs != 1 && currlevel == player.plrlevel && InDungeonBounds(position)) {
```
vlolteanu, line 684: Ditto for disconnecting citing desynchronization.

StephenCWills (approved):

Source/multi.cpp
AJenbo: Good catch, I'll do this as a follow up PR where I'll shave it down a bit to make it more clear that it will never be null.

Source/msg.cpp
AJenbo: From a network perspective it's probably safe enough to draw the conclusion, but considering that the engine often plays fast and loose with the bounds I would be a bit nervous that making it that unforgiving could result in some false positive disconnects. Maybe we can look in to it further after the release of 1.3.0.

Source/multi.cpp
AJenbo: Here we go: https://github.com/diasurgical/devilutionX/pull/3009

PR #3011, opened 2021-10-03
Validate simple network messages

AJenbo:


Source/msg.cpp
```diff
@@ -1537,9 +1548,9 @@ DWORD OnPlayerJoinLevel(const TCmd *pCmd, int pnum)
 		EventPlrMsg(fmt::format(_("Player '{:s}' (level {:d}) just joined the game"), player._pName, player._pLevel).c_str());
 	}
 
-	if (player.plractive && MyPlayerId != pnum) {
-		player.position.tile = { p->x, p->y };
-		player.plrlevel = p->wParam1;
+	if (player.plractive && MyPlayerId != pnum && message.wParam1 < MAXCHARLEVEL && InDungeonBounds(position)) {
```
obligaron, line 1551: This could leave the game in a inconsistent state cause the activated the player but didn't add assign him to any level. (For example what content has `player.plrlevel` and `player.position.tile`. Can they could contain uninitialized variables with negative values? If yes what would happen later int the game logic?

Source/msg.cpp
```diff
@@ -898,44 +900,47 @@ DWORD OnRespawnItem(const TCmd *pCmd, int pnum)
 
 DWORD OnAttackTile(const TCmd *pCmd, Player &player)
 {
-	auto *p = (TCmdLoc *)pCmd;
+	const auto &message = *reinterpret_cast<const TCmdLoc *>(pCmd);
```
obligaron, line 903: Can these `reinterpret_cast` result in reading out of bounds?
Example: A bad client sends a message with id `CMD_GOTOAGETITEM` but only send sizeof(TCmd) + 1 byte.
What will happen?
Bonus question: could this also occur in `HandleAllPackets`?

AJenbo: Yes that seams likely. But it is only reads (`const`), and we are now validating that read content. So it wouldn't let you cause anything you couldn't simply do by providing a valid size and distinct values.

I'm unsure if `multi_process_network_packets()` currently does sufficient validation to prevent an OOB read in `HandleAllPackets()`, but `ProcessTmsgs()` should be using it safely at this point.

Source/msg.cpp
AJenbo: Moved the check up so that it bails before announcing the player.

Source/msg.cpp
obligaron: If you read uncommitted memory couldn't you get a read access violation? And this could turn down the process?
I know this is unlikely and shouldn't be security issue cause at worst the app crashes, just wanted to mention it. 🙂

Source/msg.cpp
```diff
@@ -1520,26 +1530,32 @@ DWORD OnSendPlayerInfo(const TCmd *pCmd, int pnum)
...
 
 	if (gbBufferMsgs == 1) {
-		SendPacket(pnum, p, sizeof(*p));
-		return sizeof(*p);
+		SendPacket(pnum, &message, sizeof(message));
+		return sizeof(message);
+	}
+
+	int playerLevel = message.wParam1;
+	if (playerLevel >= MAXCHARLEVEL || !InDungeonBounds(position)) {
```
obligaron, line 1542:
```suggestion
	if (playerLevel > (gbIsHellfire ? 24 : 16) || !InDungeonBounds(position)) {
```

Overlooked this in the first review. 😝 
`plrlevel` corresponds to dungeon `currlevel`.
`_pLevel` is the character level.

Source/msg.cpp
AJenbo: Didn't mention it as it's unlikely :)

For 1.3.0 the main concern is security, but being able to ruin peoples fun isn't grate either. I have thought of ways to prevent it, but it would require some restructuring so I'm going to wait til after all the validation is done with seeing how i feel about it.

Source/msg.cpp
AJenbo: Fixed

obligaron (approved):

PR #3012, opened 2021-10-03
Validate spell networkmessages

AJenbo:


Source/msg.cpp
```diff
@@ -1104,92 +1122,128 @@ DWORD OnRangedAttackPlayer(const TCmd *pCmd, Player &player)
...
+	auto spell = static_cast<spell_id>(message.wParam2);
+	if (currlevel == 0 && !spelldata[spell].sTownSpell) {
+		LogError(_("{:s} has cast an illegal spell."), player._pName);
+		return sizeof(message);
+	}
+
+	ClrPlrPath(player);
+	player.destAction = ACTION_SPELLMON;
+	player.destParam1 = message.wParam1;
+	player.destParam2 = message.wParam4;
```
obligaron, line 1147: Should we validate spellLevel ?
Cause `ScaleSpellEffect` could overflow.
(I didn't check if a negative damage has a negative effect. Just wanted to mention it.)

AJenbo: I feel that for now it's only necessary to validate engine integration and not game rules. Trying to prevent chats this way is probably largely ineffective, complicated, and risks introducing game play bugs.

Other then the gameplay issue of healing monsters I don't think there is much consequence of overflowing the spell level.

obligaron: For example (not the case in the game): the range of nova could depend on spell level and to check the range there is a array int RangeInTilesForSpellLevel[20]; This could overflow.
I'm not aware of any case there something like this could happen. Just wanted to mention that I didn't checked all places where a too great value could cause a (security) problem.
So I'm okay with it. 🙂

obligaron (approved):

Source/msg.cpp
AJenbo: Gave it a second look and it all looks good :)

PR #3015, opened 2021-10-03
shorter & better MeasureSolidHorizontalBounds

qndel:
Generates same results for all items as the old one = I think it works ;)


glebm:
The previous version did an early return, this one doesn't 
It is shorter, but better? 🤔

qndel:
> The previous version did an early return, this one doesn't It is shorter, but better? 🤔

Early return means it might not properly for sprites with a fully transparent line in the middle :P

glebm (approved):

PR #3021, opened 2021-10-04
Fix OnSpellTile check

obligaron:
`wParam3` is the spell level, not the direction (like in `CMD_SPELLXYD`).
What great network messages. 😛

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -979,8 +979,6 @@ DWORD OnSpellTile(const TCmd *pCmd, Player &player)
 		return sizeof(message);
 	if (message.wParam2 > RSPLTYPE_INVALID)
 		return sizeof(message);
-	if (message.wParam3 > static_cast<int>(Direction::SouthEast))
-		return sizeof(message);
 
 	auto spell = static_cast<spell_id>(message.wParam1);
 	if (currlevel == 0 && !spelldata[spell].sTownSpell) {
@@ -992,7 +990,7 @@ DWORD OnSpellTile(const TCmd *pCmd, Player &player)
 	player.destAction = ACTION_SPELL;
 	player.destParam1 = position.x;
 	player.destParam2 = position.y;
-	player.destParam3 = static_cast<Direction>(message.wParam3);
+	player.destParam4 = message.wParam3;
 	player._pSpell = spell;
 	player._pSplType = static_cast<spell_type>(message.wParam2);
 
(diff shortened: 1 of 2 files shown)


julealgon:
> What great network messages. 😛

Yeah.... honestly, this is the result of laziness: instead of creating dedicated message models, one for each command (with their typed, specific values), a "generic X-number-of-arguments" type was created and reused for many different commands.

I'd recommend we change this at some point to dedicated classes.

Source/msg.cpp
```diff
@@ -979,8 +979,6 @@ DWORD OnSpellTile(const TCmd *pCmd, Player &player)
 		return sizeof(message);
 	if (message.wParam2 > RSPLTYPE_INVALID)
 		return sizeof(message);
-	if (message.wParam3 > static_cast<int>(Direction::SouthEast))
-		return sizeof(message);
```
julealgon, line 983: Are we not going to add any validation to the spell level? It seems it should have some basic range checks at least (`> 0` and  `< cap`)?

julealgon: Just saw this discussion:
- https://github.com/diasurgical/devilutionX/pull/3012#discussion_r720858392

I guess it resolves my comment.

AJenbo: I think the max level is 20 + 2, but I don't feel confident enough to apply this restriction 😅

AJenbo:
In that case we probably shoudn't cast it to a direction either....

obligaron:
> I'd recommend we change this at some point to dedicated classes.

I think this is a good idea.
But first I would go even further and change how multiplayer works.

Currently we have so many specials (sync data, ai, ...).
The initial state between sever and client should be shared with the same format as the savegame loading/saving.
After that server and client should only need to exchange what the player does (could be the same command used for demo recording; see command-based solution in #2733).

I think this could reduce the code needed for multiplayer significantly.

Source/msg.cpp
julealgon: What about `< 0`? Can spells with negative level be cast normally in the game somehow?

obligaron: `wParam3` is unsigned. So we shouldn't get negative numbers.

AJenbo:
@obligaron I did some aditional adjustments to make it a bit more sane, does it look ok to you?

obligaron:
> @obligaron I did some aditional adjustments to make it a bit more sane, does it look ok to you?

I think this is good. 🙂 
Just thought about going a bit further and swap `destParam3` and `destParam4`, cause `destParam3` is only used if destParam4 is also used. But this is really minor.

Also to note, this is a small minor breaking change. Cause now queued spell attacks could behave differently (save was written with `destParam3` and we now use `destParam4`). But I think this is negligible.

Issue #1467, opened 2021-04-10
Auto gold pickup playing sound but not picking up gold pile


SoundChaser83:
https://www.twitch.tv/soundchaser83/clip/TiredCrackyInternOSsloth-EK5NDHxUFex-y8LY?filter=clips&range=30d&sort=time

The above clip shows the issue occurring live on stream. Every time I take a step next to a valid tile where gold pickup is possible, the pickup sound plays but the gold stays on the ground. I had to manually pick up the gold like normal.

julealgon:
@SoundChaser83 could you provide the exact version you are using? Is it the official release, or an older build shared by @AJenbo ?

AJenbo:
I didn't make any old builds :( Probably grabbed it from NiteKat who got it from CircleCI :shrug:

SoundChaser83:
@julealgon This was on the main 1.2.0 released version.

Issue #1360, opened 2021-04-04
Add item pickup sounds


julealgon:
## Description

Picking up items from the ground in vanilla does not produce any sounds. Some games have started to introduce a pickup sound to provide better feedback to the player (i.e. Diablo 2).

Currently, picking up gold automatically when Auto Gold Pickup is enabled, produces a pickup sound. This should be expanded to all items.

## Details
- We can use the same "inventory pick" sound to use as the sound for grabing items on the ground.
- Ensure the same logic runs for gold pickup (move current gold pickup sound to a central place)
- Pickup sound should happen as the character actually picks up the item, not before network commands are sent

New flag should be added for this. Disabled by default.

Issue #1355, opened 2021-04-04
Gold Piles Next to Trigger with Auto Pickup Playing Audio when Not Picked up


NiteKat:
**Important information**
Windows 10 / 184fbde / Downloaded from circleci

**Describe**
If you drop gold next to the trigger to take you to a previous or next floor, and then step onto that trigger, the audio for picking up the gold with auto gold pickup will play, but the gold will not be picked up.

**To Reproduce**
1. Drop gold (or have a monster drop gold) on a tile that is adjacent to a trigger tile.
2. Step onto the trigger tile with the gold still there.

**Expected behavior**
Either no audio should play and the gold remains on the ground or the gold is picked up.

AJenbo:
Hmm not easy to fix, the game fires events in funny ways. We play the sound when we request the gold to be picked up. But its fire and forget so we won't know if it actually happens or not.

julealgon:
> We play the sound when we request the gold to be picked up.

We should be able to play it once it is picked up. That's what I wanted to do to introduce item pickup sounds.

AJenbo:
Ok, fix the mana usage bug while you are at it :)

PR #3022, opened 2021-10-04
Add item pickup sounds

burned42:
Fixes https://github.com/diasurgical/devilutionX/issues/1360

Should also fix https://github.com/diasurgical/devilutionX/issues/1467 and https://github.com/diasurgical/devilutionX/issues/1355 as the sound is now played when adding the item to the inventory.

One drawback is that when the new option 'Item Pickup Sound' is disabled, the gold pickup sound is not played anymore as it was playing before when auto gold pickup was enabled. If you don't like this change the previous behavior can also be kept with a slightly complex if statement.

I kept the previous sound for picking up gold, but we could also drop this special treatment and use the default chink sound for gold.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1610,6 +1610,14 @@ void AutoGetItem(int pnum, Item *item, int ii)
 	}
 
 	if (done) {
+		if (sgOptions.Audio.bItemPickupSound && pnum == MyPlayerId) {
+			if (player.HoldItem._itype == ItemType::Gold) {
+				PlaySFX(IS_IGRAB);
+			} else {
+				PlaySFX(ItemInvSnds[ItemCAnimTbl[item->_iCurs]]);
+			}
+		}
+
 		CleanupItems(&Items[ii], ii);
 		return;
 	}
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -216,6 +216,7 @@ void LoadOptions()
 	sgOptions.Audio.nMusicVolume = GetIniInt("Audio", "Music Volume", VOLUME_MAX);
 	sgOptions.Audio.bWalkingSound = GetIniBool("Audio", "Walking Sound", true);
 	sgOptions.Audio.bAutoEquipSound = GetIniBool("Audio", "Auto Equip Sound", false);
+	sgOptions.Audio.bItemPickupSound = GetIniBool("Audio", "Item Pickup Sound", false);
 
 	sgOptions.Audio.nSampleRate = GetIniInt("Audio", "Sample Rate", DEFAULT_AUDIO_SAMPLE_RATE);
 	sgOptions.Audio.nChannels = GetIniInt("Audio", "Channels", DEFAULT_AUDIO_CHANNELS);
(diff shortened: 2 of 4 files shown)


Source/inv.cpp
```diff
@@ -1610,6 +1610,14 @@ void AutoGetItem(int pnum, Item *item, int ii)
 	}
 
 	if (done) {
+		if (sgOptions.Audio.bItemPickupSound && pnum == MyPlayerId) {
+			if (player.HoldItem._itype == ItemType::Gold) {
+				PlaySFX(IS_IGRAB);
+			} else {
+				PlaySFX(ItemInvSnds[ItemCAnimTbl[item->_iCurs]]);
```
julealgon, line 1617: We are not supposed to play _equip_ sounds on pickup unless they are automatically equipped. It should always play `IS_IGRAB` instead.

burned42: @julealgon oh ok, sure, I adjusted the changeset in fa71dc10ce90d6740691e37e22d05934308356a9

Source/inv.cpp
```diff
@@ -1601,6 +1602,10 @@ void AutoGetItem(int pnum, Item *item, int ii)
 		}
 	} else {
 		done = AutoEquipEnabled(player, player.HoldItem) && AutoEquip(pnum, player.HoldItem);
+		if (done) {
+			pickupSound = ItemInvSnds[ItemCAnimTbl[item->_iCurs]];
```
julealgon, line 1606: I think you misunderstood my comment earlier?

Autoequip sounds _already_ play in a separate place. When _picking_ items up, we should play only `IS_IGRAB`

burned42: > I think you misunderstood my comment earlier?

@julealgon yes indeed, I misunderstood that one, sorry.

I now see that at [inv.cpp#331](https://github.com/diasurgical/devilutionX/blob/a17b0bf947a1a60649de3111868837db2b140f18/Source/inv.cpp#L331) the auto equip sounds are played, so playing those again doesn't make sense of course :see_no_evil: 

In d78e7b4bcbe251c34f9103bd055d38d2d88fc96a I adjusted the code to just play `IS_IGRAB`. I noticed that when picking up an item which gets auto-equipped, both sounds are playing, the item sound and `IS_IGRAB`. Therefore I also adjusted the code to only play the sound when the item wasn't auto-equipped.

ThomasChr (approved):
LGTM

julealgon:
> @julealgon Pretty sure we don‘t want to merge this before 1.3 because of the feature freeze (on the other hand it will close three issues and it‘s not complicated).

Hmm not sure about that. As far as I know, there is no "feature freeze", but a "complex change freeze". Simple stuff is still being accepted.

> But can we/you at least accept it 

LGTM. 

> ...and also add the hacktoberfest-accepted tag?
> Thruth is I praised our devX Repo for beeing the best repo for Hacktoberfest PRs!

I have no idea what you are talking about there.

ThomasChr:
@julealgon Every year in October, Github, Digitalocean and some others make Hacktober. Everyone can register there and if he can make four pull request in participaiting repositorys he get‘s some swag (shirt, stickers). DevilutionX does participate (it has the hacktoberfest topic).
A PR does only count if it's either merged or it has the hacktoberfest-accepted label.

Hacktoberfest is a great oppertunity for repositorys to get newbies into helping. But on the other hand it‘s sometimes a lot of work because the PRs can be of low quality.

I‘m one of the guys who thinks that every newbie should get a chance and the bar to participating should be low and I‘m happy that devilutionX does participate in hacktober.

Also I‘m happy to help everyone who wants to get some PRs into devilutionX. Yes, mostly those PRs are nothing fancy, but I‘m thinking that some of these newbies will stick around because of how friendly they where treated. And maybe they will bring great things into devX at some point!

julealgon:
Thanks for explaining @ThomasChr . I personally don't care about that (I don't see how we are in any way blocking newbies from coming and helping normally), but it seems ajenbo got you covered already :)

PR #3026, opened 2021-10-04
Fix ctrl + drop garbage

### LINKED ISSUE #3024: Ctlr + Left Clicking Empty Belt Slots Generates Fake Items
**Important information**
Windows 10 / master / Compiled 403f507d myself

**Describe**
If you ctrl + left click an empty slot in the belt, the game will drop a fake item that looks like the item that was previously in that slot. You cannot pick up the item, but clicking it multiple times will eventually cause the fake item to disappear. Sometimes, this causes you to need to click the real item that dropped from that slot multiple times to pick it up.

**To Reproduce**
1. Create a new character.
2. Use ctrl + left click to drop one of your starting healing potions.
3. ctrl + left click that same belt slot to drop as many "fake" healing potions as your heart desires.

**Expected behavior**
ctrl + left clicking an empty belt slot should do nothing.

yuripourre:
Resolve #3024

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -887,7 +887,7 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 		}
 	}
 
-	if (dropItem) {
+	if (dropItem && holdItem._itype != ItemType::None) {
 		TryDropItem();
 	}
 }


Source/inv.cpp
```diff
@@ -887,7 +887,7 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 		}
 	}
 
-	if (dropItem) {
+	if (dropItem && holdItem._itype != ItemType::None) {
```
ephphatha, line 890:
```suggestion
	if (dropItem && !holdItem.isEmpty()) {
```

The Item struct provides a convenience method for performing this check

yuripourre: Thank you @ephphatha I didn't notice.

Source/inv.cpp
```diff
@@ -887,7 +887,7 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 		}
 	}
 
-	if (dropItem) {
+	if (dropItem && !holdItem.isEmpty()) {
```
julealgon, line 890: Once again I wonder: wouldn't it make more sense for the empty check to be inside the `TryDropItem` function instead?

qndel: @yuripourre can you apply the suggestion? :P

PR #3048, opened 2021-10-06
Basic emojis support

AJenbo:
image

Based on https://github.com/googlefonts/noto-emoji (same license as the CharisSILB)

This will prevent Android users from crashing the game by using emojis in chat :sweat_smile:


AJenbo:
The Android emoj font has a lot of gradients which makes its conversion to the game colors problematic. I think  Twemoji would probably be a better fit here.

AJenbo:
Fixed via https://github.com/diasurgical/devilutionX/commit/2d3574db7068cfd94f849ddc85801f2a902f488f

image

PR #3050, opened 2021-10-07
Remove pointless increments of life/mana on level up

### LINKED ISSUE #3049: Single player characters gain an extra fraction of a point in mana/hp per level
@StephenCWills pointed out in discord that single player characters have a fairly pointless increment to hp/mana in `NxtPlrLevel` https://github.com/diasurgical/devilutionX/blob/31ba8e2bc68f4efab5ba4cd7c7dc911d288fc719/Source/player.cpp#L2583-L2590

This results in a level 50 character having an extra ~0.77 points of hp/mana (even barbarians, which start at 0 mana and should gain 0 each level).

ephphatha:
fixes #3049

No visible impact on gameplay since this never leads to a full point difference in mana or hp, but this at least removes two references to the global `gbIsMultiplayer`

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2581,9 +2581,7 @@ void NextPlrLevel(int pnum)
 	player._pNextExper = ExpLvlsTbl[player._pLevel];
 
 	int hp = player._pClass == HeroClass::Sorcerer ? 64 : 128;
-	if (!gbIsMultiplayer) {
-		hp++;
-	}
+
 	player._pMaxHP += hp;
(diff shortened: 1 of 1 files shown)


ephphatha:
@nitekat pointed out this is mentioned in [The DSF Buglist](http://www.lurkerlounge.com/forums/thread-12422.html) under Miscellaneous Bugs > Single Player Level Bonuses. The description there doesn't reflect that Sorcerers gain 2 points of mana instead of 1 (128/64 instead of 64/64), but the increment is always 1/64 for all classes.

PR #3061, opened 2021-10-08
Remember last selected hero

obligaron:
Let the game remember the last selected hero

### Master

https://user-images.githubusercontent.com/25415264/136593714-10d5696b-6c7b-4474-a18a-0b4cada5f9cf.mp4

### PR

https://user-images.githubusercontent.com/25415264/136593722-dbf4c448-c40c-48dc-a20d-a8da836aab1c.mp4

</details>

Notes:

- Stores the last used save game
- Differs between Diablo/Hellfire and Single-/Multiplayer

### DIFF
--- a/Source/menu.cpp
+++ b/Source/menu.cpp
@@ -86,11 +86,14 @@ bool DummyGetHeroInfo(_uiheroinfo * /*pInfo*/)
 
 bool mainmenu_select_hero_dialog(GameData *gameData)
 {
+	uint32_t *pSaveNumberFromOptions = nullptr;
 	_selhero_selections dlgresult = SELHERO_NEW_DUNGEON;
 	if (demo::IsRunning()) {
 		pfile_ui_set_hero_infos(DummyGetHeroInfo);
 		gbLoadGame = true;
 	} else if (!gbIsMultiplayer) {
+		pSaveNumberFromOptions = gbIsHellfire ? &sgOptions.Hellfire.lastSinglePlayerHero : &sgOptions.Diablo.lastSinglePlayerHero;
+		gSaveNumber = *pSaveNumberFromOptions;
 		UiSelHeroSingDialog(
 		    pfile_ui_set_hero_infos,
 		    pfile_ui_save_create,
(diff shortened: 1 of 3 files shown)


julealgon:
Very cool idea. I think this deserves a mention on release notes. Adding the "enhancement" tag.

julealgon:
Have you considered ordering the list of heroes by "last played time" instead of doing something like this? I feel that would be a more common/expected solution to this problem and wouldn't need any cursor manipulation.

frrsth:
I don't think we should change the sort order at all. Because if we change the order and someone doesn't remember the order they created the heroes and they are trying to copy just one of their saves they wouldn't know which one to copy. Unless we change the names of the saves to the heroes name, then the order could be whatever.

qndel:
> I don't think we should change the sort order at all. Because if we change the order and someone doesn't remember the order they created the heroes and they are trying to copy just one of their saves they wouldn't know which one to copy. Unless we change the names of the saves to the heroes name, then the order could be whatever.

Debug mode lets you see save numbers in character selection + it's pretty easy to know which save has which number - load a character, save, see which save has most recent "date modified"

AJenbo:
>  list are not sorted by "last modified/accessed" for example a file system or in git issue

https://github.com/diasurgical/devilutionX/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

that said i prefer the list to be stable

obligaron:
Sorry I wasn't clear enough, I meant by default. 😉

julealgon:
> I think a list that is sorted by "last played time" makes the game look unstable.

How so? 

> Normally list are not sorted by "last modified/accessed" for example a file system or in git issue/pr/code. But this is also preference.

Most "recently access/used" lists are ordered by access date. This is how I operate in Visual Studio every day for example, the last solutions I've opened appear first.

Games on my Nintendo Switch dashboard are also shown "most recently played" first.

I think comparing the game to a file system is missing the point: the game is not a "file explorer".

> Another thing we could do, is change the sort of the list. Currently it's sorted by savenumber descending. That's kind of arbitrary. Imho sort by alphabetical -> savenumber or class -> alphabetical -> savenumber would be better.

This to me just complicates things more. The main reason I suggested ordering in most recently played order is that you actually don't need to persist cursor positions.

Also, what happens today with this change if someone adds a file to the list, or otherwise change the save files? Will the cursor point to an arbitrary save now? Most recently played would prevent that issue too.

obligaron:
In most cases (expect the visual studio solution example 😉 ), I prefer a stable list and I think @qndel has the same feeling. But this personal preference. My analogy with file system and github was that this is the default for many systems (but can be changed by personal settings. So I think there is no clear right or wrong.

We don't persist cursor position. We persist last used save game number. So even if change the saves (add new ones, delete existing etc.). The selection will not change. 🙂

PR #3065, opened 2021-10-09
Fix broken links and replace http links with their https version

burned42:
I noticed that some links in this repo are using http instead of https, so I tried replacing them with their https version. While doing that I also noticed some broken links and tried to fix those.

### Changes
- 115e9de18d8e57dd2b4cf546e3d3913613448c6f I did adjust http links to https where possible (i.e. where there's a working ssl certificate via https / no error from the browser)
- cc1eca55292bcb8600e155780f335c35a343eac2 I adjusted links to libsdl to point to github, some where http links and while adjusting those, I changed them to point to github as they forward there anyway
- 5dd46478c770770cc685b1194f1aae42f4c0820f I replaced some dead links with a link to their snapshot in the wayback machine
- 816f8299f1c2a16446d10fec5fd4a48a8125be18 I replaced links regarding the platform GKD350h where a release file is linked that was last built in version 1.0.1 of devilutionX, so I pointed the links to that version

### Things I didn't change
- I didn't touch the 3rd party stuff as I guess those just get copied from somewhere else and may get updated (automatically?)
- I didn't adjust the links for xml namespaces and such as I don't know if that might break anything
- Unfortunately I couldn't manage to find a working version of https://github.com/sanctuary/graphics/blob/master/inventory.png (see https://github.com/diasurgical/devilutionX/blob/master/Source/inv.h#L39) so that link remains broken for now

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -104,7 +104,7 @@ sudo apt install devilutionx
 
 <details><summary>GKD350h</summary>
 
-- Copy [devilutionx-gkd350h.opk](https://github.com/diasurgical/devilutionX/releases/latest/download/devilutionx-gkd350h.opk) to `/media/data/apps` or `/media/sdcard/apps/`.
+- Copy [devilutionx-gkd350h.opk](https://github.com/diasurgical/devilutionX/releases/download/1.0.1/devilutionx-gkd350h.opk) to `/media/data/apps` or `/media/sdcard/apps/`.
 - Copy the MPQ files to `/usr/local/home/.local/share/diasurgical/devilution/`
 
 </details>
--- a/docs/manual/platforms/gkd350h.md
+++ b/docs/manual/platforms/gkd350h.md
@@ -4,7 +4,7 @@
 
 ## Installation
 
-1. Copy [devilutionx-gkd350h.opk](https://github.com/diasurgical/devilutionX/releases/latest/download/devilutionx-gkd350h.opk) to `/media/data/apps` or `/media/sdcard/apps/`.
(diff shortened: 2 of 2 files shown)


burned42:
@AJenbo sure, I didn't notice that `android-project/` is also something that's synced from somewhere else, I reverted my changes in that directory in 5d495d834eda68e82b4c06a44c59c86878c82b74

I also rebased on latest master and did a force push.

AJenbo:
Thanks, I amended the previous commits and dropped the revert (easy now that you made the actual edits) to avoid unnecessary churn after merging.

AJenbo:
@burned42 here is the missing image:

image

I'm not sure it adds a lot of value, so maybe we should just remove the link?

burned42:
> @burned42 here is the missing image:
> 
> ...
> 
> I'm not sure it adds a lot of value, so maybe we should just remove the link?

@AJenbo I'm impressed that you found it, I really did my best to search for it :+1: 

I guess if you never had the game running and you're just looking at the code, this image could be helpful. However I don't know if that's to be expected for someone looking at the source code of the game :grin: 

But TBH I expected the image to maybe include some indices or something, explaining the numbers used as identifiers for the slots in https://github.com/diasurgical/devilutionX/blob/master/Source/inv.h#L39.

Should I open a PR to remove the link?

AJenbo:
> Should I open a PR to remove the link?

I think that would be the thing to do her 🙂
The is a textual version of the image somewhere else in the code, but I don\t remember exactly where. We could add a `@see` reference to that instead

PR #3072, opened 2021-10-09
Replace broken link with @see annotation to text representation of the linked image

burned42:
@AJenbo as discussed in https://github.com/diasurgical/devilutionX/pull/3065#issuecomment-939322597

I found the text representation in the code that you mentioned and replaced the broken link with a reference to it. I'm not sure how best to reference it as a specific link to one line will probably break in future then the file gets changed but the comment doesn't, so I just described where to find it. If you have a better idea I'm open for suggestions.

There also was a broken `@see` annotation at the mentioned comment (probably also referencing the same image) so I removed that, too.

### DIFF
--- a/Source/inv.h
+++ b/Source/inv.h
@@ -37,7 +37,7 @@ enum inv_item : int8_t {
 
 /**
  * identifiers for each of the inventory squares
- * @see inv.cpp (comment block at InvRect[] initialization)
+ * @see #InvRect
  */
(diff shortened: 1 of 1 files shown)


burned42:
> `@see` should follow these rules: https://www.doxygen.nl/manual/autolink.html

@AJenbo please see d901db6a6a494a638341f2dd2fad9b0d837d483a

PR #3076, opened 2021-10-09
Support building without smpq

obligaron:
Notes:
- If smpq isn't available, assets will be copied to `${CMAKE_CURRENT_BINARY_DIR}/assets`.
- App search exe.path + `/assets` folder as last fallback
- I didn't wanted to change file search logic too much in this pr. But I thinks this should be adjusted in a follow-up pr, cause this can be improved. For example: On windows with gog installed the last loaded mpq is `hfopt2.mpq`. The last loaded mpq sets `SBasePath` and `SBasePath` is the only location where overridden files are searched. So per default `gog-diablo-folder/hellfire` is searched for overridden files. 😉 Note this wasn't the case before e5d32da8a2a9174af29e22293ddeba738993c69f.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -135,6 +135,7 @@ Make sure to install the `C++ CMake tools for Windows` and `Windows SDK` compone
 Install vcpkg following the instructions from https://github.com/microsoft/vcpkg#quick-start-windows.
 Don't forget to perform _user-wide integration_ step for additional convenience.
 
+### If you want to build the devilutionX.mpq File (optional)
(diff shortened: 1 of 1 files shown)


licaon-kter:
Testing on Android, looks like it doesn't want to read from
`/Android/data/org.diasurgical.devilutionx/files/assets/` or
`/data/data/org.diasurgical.devilutionx/files/assets/` or
`/data/data/org.diasurgical.devilutionx/files/` 

but just 
`/Android/data/org.diasurgical.devilutionx/files/` (fonts/..., data/... etc)

Is that expected per https://github.com/diasurgical/devilutionX/blob/e53960eb0c009e9b2a60e96a805fd4a88dddc416/Source/storm/storm.cpp#L132-L138?

PR #3083, opened 2021-10-10
Vita: pack raw assets if smpq isn't available

isage:
Honestly, i noticed no difference in loading between mpq and unpacked assets.
So, if you want to, i can drop smpq from circleci build.


isage:
> Any size difference? It looks like you are copying all assets, including the PNGs for the SDL2 touch interface. Not a real problem but just wanted to make sure you realized it's not necessary to copy all files :)

Non-mpq build is around 900kb bigger. An oh, yeah, didn't know about assets. But it's just easier to copy them all instead of file-by-file.

AJenbo:
As long as 900kb doesn't sound scary to you :)

Does it have any form of compression? When adding full fonts the data will probably grow to about 30mb, but just 10mb compressed in an MPQ.

isage:
VPKs are zip with default deflate compression.
Well, we'll see. I'm okay with 30mb too. Maybe by that time smpq package will appear in alpine)

isage:
CircleCI artifact is actually somehow smaller than previous one (with mpq) and my local build.

PR #3092, opened 2021-10-11
Prevent Spell Targeting Players with Friendly Mode enabled

kphoenix137:
Prevents players from targeting each other with spells in the same way that players are prevented from targeting each other with weapons when Friendly Mode is enabled


obligaron (approved):
lgtm.
Checked `Heal Other`, cause it's also a spell and you should "target" another player.
But `Heal Other` uses a selection cursor after the spell is casted, so this pr doesn't break anything.

PR #3096, opened 2021-10-11
optimize SyncInitPlrPos

qndel:
Doesn't check same tiles multiple times anymore.
Optimization for full 50 range run : 166649 iterations -> 9801


qndel:
wut

Issue #3103, opened 2021-10-12
Item drops on cursor position while in a different dlvl


StephenCWills:
If an item is dropped by Player A when Player B is in a different dlvl, Player B's client will see the item dropped on Player A's cursor position instead of the actual drop location chosen by Player A's client.

**To reproduce**
1. Create a TCP game on Client 1
2. Join the TCP game on Client 2
3. Travel to dlvl 1 on Client 2
4. Attempt to drop an item across the river in Client 1 -- the item will drop next to the player
5. Travel back to town on Client 2 -- the item will appear to be on the other side of the river

---

**Client 1 (Player A)**
![image](https://user-images.githubusercontent.com/9203145/136878795-aa96c11a-a4e4-4428-a958-dae0a1e84d96.png)

**Client 2 (Player B)**
![image](https://user-images.githubusercontent.com/9203145/136878710-ab68033e-cf06-49e0-88cd-b0dc5f91170d.png)

PR #3104, opened 2021-10-12
Enable opening Hive and Crypt using gamepad

### LINKED ISSUE #2540: Using the rune bomb creates a duplicate in multiplayer
When one player drops the rune  bomb on the hive entrance it will activate correctly for both players. But other players will also see the bomb drop as a normal items and can pick it up.

StephenCWills:
* Introduces `OpensHive(Point position)` and `OpensGrave(Point position)` functions in town.cpp
* Update logic in `InvPutItem()` to use these new functions
* Check for Rune Bomb using `IDI_RUNEBOMB` instead of `ICURS_RUNE_BOMB`
* Add additional logic to `SyncPutItem()` to prevent erroneous item drops in Multiplayer when Hive or Crypt are opened
* Add logic to plrctrls.cpp `TryDropItem()` to check if the player is next to Hive or Crypt when dropping the Rune Bomb or Cathedral Map respectively

This resolves #2540 (Assuming all players are in town when the item drops; otherwise, see #3103)

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -18,6 +18,7 @@
 #include "minitext.h"
 #include "missiles.h"
 #include "stores.h"
+#include "town.h"
 #include "towners.h"
 #include "trigs.h"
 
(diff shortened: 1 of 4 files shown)


AJenbo:
Oh nice, was going to try and tackle this one next so that I could continue my warrior play test, but now I can skip right to play testing :D

AJenbo:
P.s. could we maybe also make it so that if you click "use" on the item and you are standing in range of the matching entrance it also triggers?

PR #3118, opened 2021-10-13
:bug: Don't follow stuck leader

### LINKED ISSUE #2232: Enemies don't attack
For some unknown reason, the enemies don't attack and just stand around.
It looks strange, although the attached savegame also behaves in the same kind in vanilla. 

screen01
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/6720047/single_0.hsv.zip)

AJenbo:
Fixes #2232

This causes minions to evaluate if there leader is stuck before continuing to follow it blindly. It is not uncommon for the leader to be temporary stuck for a split second, meaning that this change will cause some slight behavioral changes to groups in general where they will now prefer to move forward if their leader has yet to make up his mind. So groups are more prone to splitting up, however usually they will almost instantly regroup as the leader is rarely far behind.

For that reason I would prefer if we can have some play testing of this PR before it is merged, @Chance4us do you have time to do so?

Before:
stuck

After:
free

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1923,7 +1923,8 @@ void GroupUnity(Monster &monster)
 	assert(monster._uniqtype == 0);
 
 	auto &leader = Monsters[monster.leader];
-	if (IsLineNotSolid(monster.position.tile, leader.position.future)) {
+	bool leaderIsStuck = leader._mmode == MonsterMode::Stand && !DirOK(monster.leader, monster._mdir);
+	if (!leaderIsStuck && IsLineNotSolid(monster.position.tile, leader.position.future)) {
 		if (monster.leaderRelation == LeaderRelation::Separated
 		    && monster.position.tile.WalkingDistance(leader.position.future) < 4) {
 			// Reunite the separated monster with the pack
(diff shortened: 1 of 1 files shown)


Chance4us:
> For that reason I would prefer if we can have some play testing of this PR before it is merged, @Chance4us do you have time to do so?

Yes, as soon as I find trouble free time

ephphatha (dismissed):
sensible variable name and Chance4us confirms the behaviour is improved, looks good :D

Source/monster.cpp
```diff
@@ -4368,7 +4369,7 @@ void ProcessMonsters()
 				raflag = MonsterTalk(monster);
 				break;
 			}
-			if (raflag) {
+			if (raflag || monster._mmode == MonsterMode::Stand) {
```
obligaron, line 4372: Why do we check `Stand` here?
I ask cause gargoyles don't use stand when sitting on the ground and I'm unsure if we need to address this also here.

AJenbo: To see if they are attempting to move but unable to do so. For gargoyles, they are prohibited from sleeping when loading a group i believe

PR #3123, opened 2021-10-13
Add mWhoHit Indicators

NiteKat:
Adds mWhoHit indicator graphics to the monster health bar. This will allow players to know who has tagged a monster for experience. Together with https://github.com/diasurgical/devilutionX/pull/3114 should provide accurate information to players. Colors used match the colors used for minimap arrows.

Would be very useful for ironman style runs with multiple players, and may be of interest to casual runners if trying to not out level the other player(s).

Example screenshots:
image

image


NiteKat:
> The look misaligned to me: image
> 
> Please double check.

If you mean from the left of the bar, it's using the same offset value as resistances, and the bottom tips are slightly on the bar by choice (like how the resistances protrude into the health bar slightly).

julealgon:
Also, as discussed in discord, these colors look really subpar to me for use as player colors... Usually, player colors in games are easily distinguishable from one another so they can be quickly associated to each player.

I think we should consider changing these colors first, before this change takes place.

julealgon:
Can you share an example with all colors set, for reference and discussion?

NiteKat:
> Can you share an example with all colors set, for reference and discussion?

I edited the original PR screenshots with corrected graphics (need to update the PR with the new graphics), and here is a screenshot with all indicators lit up.

image

NiteKat:
> > Either way, I think discussion around the colors used (outside of saying to use a single color) should go in a different issue/comment section?
> 
> I'd disagree. I think introducing this change with these colors increases confusion and is a bad idea, so the colors should be changed prior to it being merged.
> 
> To me, the current colors are a blocker for this.

Then I guess we'll just have to disagree on this one. Changing the colors associated with the players is **_not_** in the scope of this pull request. I will instead take your response as meaning "we should use a single color for now for this feature" and wait for more feedback before making such a change to it.

AJenbo:
> you are making a problem worse by exposing it on other parts of the game

A stamens like this could hold true for building shared map based on un-synced vision range. But in this case it is trivial to the point where it can be ignored.

FitzRoyX:
Can we see a playthrough video of the problem this is trying to address? I guess I can't visualize this being useful, it seems like you'd die more being worried about killing monsters too quickly in battle.

AJenbo:
I fixed the code style, feel free to run clang-format for the next PR ;)

AJenbo (approved):

PR #3124, opened 2021-10-13
Introduce template specialisation for Enum as flags

ephphatha:
This allows marking an enum as being a collection of flags and supporting basic bitwise combinations/tests to see if a flag is set.


AJenbo:
At least the code style is correct :)

ephphatha:
Turns out scoped enums require strict types 😅

AJenbo:
They are really annoying to work with :(

Issue #2959, opened 2021-09-27
Feedback on current touch controls


AJenbo:
After playing the game a bit on my phone using the new touch controls I though I would gather some of my thoughts regarding how they feel.

1. They work, the game is actually enjoyable to play

2 The inventory is often obscured by the action buttons, maybe if we instead moved them to the center of the screen it would be easier to interact with them. This should be easy after the recent refactoring that allows us to relocate them. Only thing that might need a check is the unique item popup since it might not be aware that there isn't room for all 3 panels despite the screen width being there.

3, main panel buttons are hard to click even on a 21:9 screen, we should consider having seperate buttons, at least for the main 3 (char, inv, menu), probably at the top of the screen.

4. Drinking potions on the right side might not be ideal, during fights you are already mashing the attack button on the right side, but you aren't really doing anything on the left side so maybe move the potion buttons there. Alternatively we could make the belt touch event "use" when the inventory isn't open.

5. The buttons sometimes do not update to match the current action, for example if an item drops next to you it gets highlighted, but the button action is still blank. Maybe we need to share more logic between how the game decides the current button action.

6. I know that some icons are still missing but I think it might have been overlooked that the top action button will make you enter dungeons when near one. I think this is a helpful one as it's sometimes hard to place yourself on the trigger. Maybe a stair icon would work here?

7. I fear that the inventory isn't intuitive and that I was mostly able to figure it out because I know how it works with a regular controller, this might get better as more button icons are added.

8. The row wrapping when you hit the end of the inventory (with dpad navigation) makes it seam like the cursor is just jumping around uncontrollably, this is probably exasperated by the high repeat rate.
If realizing that the joy pad can be used as a dpad inventory navigation with single clicks becomes way easier then trying to hold down any direction. 

9. It's possible to close the inventory while holding an item with out dropping it.

10. We should enable all the "auto" options by default when building with virutal controls

StephenCWills:
> I know that some icons are still missing but I think it might have been overlooked that the top action button will make you enter dungeons when near one. I think this is a helpful one as it's sometimes hard to place yourself on the trigger. Maybe a stair icon would work here?

Actually, I added a comment to remind me of that one. :+1:

https://github.com/diasurgical/devilutionX/blob/16d98d48c3333e747e508be309e39f7d32b17709/Source/controls/touch/renderers.cpp#L348-L358

[...]

PR #3133, opened 2021-10-14
[Gamepad] Don't treat special cursors as items

StephenCWills:
Fixes the issue where gamepad users can't apply oils, repair skill, or recharge staff skill to equipped gear.

See: https://github.com/diasurgical/devilutionX/issues/2959#issuecomment-938949242


AJenbo:
By the way I think I figured out why the inventory is sometimes behaving erratically. I believe we made it so that the cursor will skip empty spots when the cursor isn't the empty hand, but this only makes sens if it's on of the cursors that is suppose to be applied to an item.

PR #3134, opened 2021-10-15
[Virtual Gamepad] Closing inventory drops item

StephenCWills:
This PR ensures that the player cannot close the inventory while holding an item.

* Prevents main panel interactions when an item is held
* Use `GameActionType_TOGGLE_INVENTORY` when pressing the gamepad's cancel button from the inventory so the item will automatically be dropped when the inventory is closed

---

Based on feedback from #2959
> It's possible to close the inventory while holding an item with out dropping it.


glebm:
I often drop items by closing the inventory with an actual gamepad on purpose

PR #3135, opened 2021-10-15
Use scoped enums for DungeonFlags/BFLAG

ephphatha:
This was an unnamed enum previously with plenty of chances for confusion given it's intended for use with the `dFlags` array but not the `dflags` array.

~~Uses #3124, feel free to ignore the first commit. Can drop the last commit too.~~ Stripped out the irrelevant commits now.

I've pulled InDungeonBounds into the header and made it constexpr as it's a very simple check that gets reused a lot in the if conditions that were then touched by this refactoring. I'm hoping all compilers we target will be clever enough to optimise this check out given it's usually performed at the start of the calling function anyway.

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -567,7 +567,7 @@ void InitAutomap()
 
 	for (auto &column : dFlags)
 		for (auto &dFlag : column)
-			dFlag &= ~BFLAG_EXPLORED;
+			dFlag &= ~DungeonFlag::Explored;
 }
 
 void StartAutomap()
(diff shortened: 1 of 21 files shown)


AJenbo (approved):

AJenbo:
Needs a rebase :(

ephphatha:
> Needs a rebase :(

Done :). One of those non-conflicting conflicts that git couldn't handle automatically.

AJenbo (approved):

PR #3141, opened 2021-10-15
Load full width kerning programmatically

AJenbo:
A lot of rows in Unicode uses full width for all symbols, generating the
data from code saves us almost 200kb of data files.

### DIFF
--- a/Source/engine/render/text_render.cpp
+++ b/Source/engine/render/text_render.cpp
@@ -29,7 +29,7 @@ namespace {
 std::unordered_map<uint32_t, Art> Fonts;
 std::unordered_map<uint32_t, std::array<uint8_t, 256>> FontKerns;
 std::array<int, 6> FontSizes = { 12, 24, 30, 42, 46, 22 };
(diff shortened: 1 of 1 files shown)


Chance4us:
@julealgon I would ask you to express yourself a little less roughly. It was just a typo and @AJenbo does a very good job. You are doing a very good job as well and we all want to be on friendly terms with each other.

Chance4us:
If you feel offended I want to apologize to you. I am not angry with you.
It's not about whether you've done something wrong, but how the other person feels about what and how has been said.
And there are certain manners in a dialogue. But expanding the topic here is the wrong place.

All in all, however, the following applies:
1. bilateral aggressive behavior destroys the mutual respect - every time it is shown

There are unspoken rules when dealing with one another:
2. Show respect for the other and always treat him / her in a polite and friendly manner

3. If mutual respect slowly but gradually erodes, this has negative consequences to
- the synergy effect
- the well-being of each individual
- friendship

PR #3148, opened 2021-10-16
Switch to release for Android test builds

AJenbo:
Phones will not always allow the user to install a debug package, this should make it a lot easier to install the latest test build

### DIFF
--- a/.github/workflows/Android.yml
+++ b/.github/workflows/Android.yml
@@ -44,4 +44,4 @@ jobs:
       uses: actions/upload-artifact@v2
       with:
         name: devilutionx.apk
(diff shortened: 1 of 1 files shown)


StephenCWills:
> Phones will now always allow the user to install a debug package...

I was confused at first, but you meant "not" instead of "now", didn't you?

AJenbo:
> EDIT: I just noticed you did run two builds, and the first one was 30 minutes so 20 minutes might be the new norm.

Hmm, let see how this works out and maybe revert back to Debug, or switch LTO off for test builds.

AJenbo:
Looks like this indeed did blow the build time from 7 min to 20 :(

PR #3155, opened 2021-10-17
[gamepad] Make horizontal inventory navigation stable

AJenbo:
To try and make navigation a bit faster in the inventory we had implemented jumping by item instead of by tile, but the way it was done made for some erratic move patterns that made it very hard to navigate.

The main issue was that the cursor would jump to the center of the item visually, but logically it would be at the top left.

Another issues was that that the cursor wraps around the end of the inventory, while this allows for faster jumping between the ends of the inventory it can be disorienting since the cursor can move pretty fast at times when a direction is held.

The new logic allows the cursor to be at the tile directly where the item was entered from, and will simply jump to the first tile that is not part of the current item when moving across the grid.

This make all movement bi directional which should hopefully make navigation feel more natural.

Before:
image
image
(passing an item pushes the cursor to the top of that item when leaving)
image
(inventory wraps, belt does not)
image
(passing an item pushes the cursor to the left when leaving)

After:
image
image

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -554,50 +554,48 @@ Point InvGetEquipSlotCoordFromInvSlot(const inv_xy_slot slot)
 }
 
 /**
- * Get coordinates for a given inventory slot (for belt use BeltGetSlotCoord)
+ * Get coordinates for a given slot
  */
-Point InvGetSlotCoord(int slot)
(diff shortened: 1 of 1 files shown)


AJenbo:
I didn't actually look much in to any of the logic related to movement when holding an item, so there may well be some edge cases there that should be looked at.
And yeah this will be part of a series of changes to the logic found here :)

PR #3157, opened 2021-10-17
Update Italian localization

M0Rf30:
As in the title.
Regards


M0Rf30:
> Thanks. Could you also check the app translation since we changed shareware to demo.

Looks good to me.

Issue #3154, opened 2021-10-17
Maps can create 0 gp treasure piles


agris-codes:
**Important information**
Win10 / modified source build from ~1 day ago, x64

**Describe**
In the catacombs, I came across a room with gold all over the floor. One of the piles of gold was 0 gp. 

**To Reproduce**
Load the attached save game and observe. [single_42.zip](https://github.com/diasurgical/devilutionX/files/7358457/single_42.zip)

**Expected behavior**
If a 0 gp gold pile is rolled, it is not placed. Conversely, it may be rounded up to a minimum of 1 gp. 

**Screenshots**
![0 gp](https://user-images.githubusercontent.com/39627597/137605613-333779cb-b8e7-486f-abdf-eff6da4d281f.jpg)

**Additional context**
I've never seen this happen before, with a lot of hours clocked in vanilla D1. AJenbo thinks this is a "bug in the original game where RNG can return negative numbers, it could be that the gold amount is set to less 4, and then capped to 0."

AJenbo:
@Chance4us could you test if this is a vanilla issue?

agris-codes:
> 
> 
> @Chance4us could you test if this is a vanilla issue?

I just loaded it in Diablo v1.09 and the 0 gp pile is still there.

Chance4us:
The 0 gp piles are actually generated in the master. I'll bisect it as soon as there is free time.

qndel:
![image](https://user-images.githubusercontent.com/14297035/137627448-4f58b477-f3cc-4159-92e6-e7ad23055a08.png)
Checked on 1.2.1, seems the whole map layout changed

Chance4us:
>  so just enter the quick entrance to test faster

Is the "quick entrance" some kind of debug feature?

PR #3158, opened 2021-10-17
fix 0 gold piles

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/3154
Applying the vanilla label because while I haven't seen 0 tiles there, it might just be super rare and according to the code it could happen and we saw it in devilutionx

### DIFF
--- a/Source/themes.cpp
+++ b/Source/themes.cpp
@@ -656,16 +656,17 @@ void Theme_SkelRoom(int t)
  */
 void Theme_Treasure(int t)
 {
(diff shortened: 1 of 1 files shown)


AJenbo:
Nice find, though I would prefer not also having some over optimization snug in there too 😉

qndel:
> Nice find, though I would prefer not also having some over optimization snug in there too 😉

It's the first and last time in my life I'm looking at that function so might as well make a proper if ;)

PR #3160, opened 2021-10-18
Gamepad inventory improvements

AJenbo:
- You can now bout use the mouse/virtual mouse/touch to navigate and the gamepad navigation will pick up from the new position instead of resetting to the previous position.

- Remove wrapping from body slots when holding a weapon/shield/ring (it's only ever one move any way, so nothing was made easier by it)

- When not relevant the cursor is no longer shown when using a touch screen


AJenbo:
> The last-minute edits on #3155 introduced an error

Oh thanks, this was the last issue I needed to fix, I had narrowed it down to that function before passing out last night, so this was really helpful to wake up to :D

PR #3170, opened 2021-10-19
Startup game (diablo or hellfire) can be selected ingame.

obligaron:
With #2748 in most cases you start with hellfire. So often we got questions how to start diablo, and we pointed them to command line parameters.
With this change you can select this ingame.

https://user-images.githubusercontent.com/25415264/137993132-b5bd33ae-f00f-445b-99f4-ac3554b0cec8.mp4

</details>

Notes:

- Selection is only shown if `hellfire.mpq` is present and `--diablo` is not specified. Cause no hellfire no choice. 😉 
- Option is saved in `diablo.ini` (under `[Hellfire] StartUpGameOption`).
- Your selection is remember. For example: If you last selection was `Diablo` the selection will be the default, so only one enter to press to continue. 😉
- If you select a `always` option, you can only change the option in the ini (until #1561 is implemented).


NiteKat:
A settings page in the main menu would be awesome for almost anything that can be toggled via the INI file.

obligaron:
@AJenbo 
I updated the PR and adjusted the logic to your requests.

Here is a sample video:

https://user-images.githubusercontent.com/25415264/138183808-b4b752cf-8845-44de-84b3-59e7edb2bfdf.mp4

Note:
With this changes, we could encounter more problems.
Cause previous we only changed the game shortly after `forceDiablo` and `forceSpawn` was executed.
Now we could load a game and switch the mode later.
If we don't reinitialize correctly, this could result in undefined behavior / bugs.

AJenbo:
>  If we don't reinitialize correctly, this could result in undefined behavior / bugs.

Worst case scenario the user can resolve it by restarting the game.

obligaron:
> Worst case scenario the user can resolve it by restarting the game.

I also think so. Just wanted to mention it, cause it could now introduce new bugs and we now have to ask "did you change the game before the bug occurred?" 😉 

> P.s. I forgot to mention that if hardware cursor is enabled this will cause it be become white when switching game mode 😅

Fixed in #3191 🙂

qndel:
> Two issues with [f51768a](https://github.com/diasurgical/devilutionX/commit/f51768a567300a63a3ddddb27377eb3a1c1a6df8) on Linux amd64
> 
> * if switched to Shareware trying to Play Intro just shows the usual message, maybe, if I have the full game, the actual Intro can be played :)
> * go to Extras -> Support -> press ESC -> goes back to the menu -> press ESC -> goes back to Support(???) instead of Main menu

Open an issue for it?
+ save behavior for intro, ESC doesn't go back but plays intro again

PR #3176, opened 2021-10-20
remember items bought from shops

### LINKED ISSUE #3172: Adria restocks shop after save/load
**Important information**

Probably any version after the Adria change to prevent save/load from randomizing her shop again.

**Describe**
Adria's shop state doesn't save when saving the game. This means you can buy a book from Adria, save the game and reload, and that same book will then be back in her shop again. This can be done indefinitely.

**To Reproduce**
1. Buy a finite supply item from Adria.
2. Save game.
3. Reload and check Adria's shop.
4. Adria's shop should be restored to the way it was when you first entered town before you last saved the game, restocking any finite supply items you previously bought.

**Expected behavior**
Saving the game should save exactly what Adria's shop state is at the time you save, not what her shop state was when her new shop was rolled upon first entering town.

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/3172

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1794,6 +1794,7 @@ void DisableInputWndProc(uint32_t uMsg, int32_t /*wParam*/, int32_t lParam)
 
 void LoadGameLevel(bool firstflag, lvl_entry lvldir)
 {
+	ResetBoughtItems();
 	music_stop();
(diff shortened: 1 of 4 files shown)


AJenbo:
Nice, I guess having this fixed before release is a good idea :D

AJenbo:
I'm this is starting to become a bit unintuitive and we may as well store the shop items the same way as is already done for the smith prim items to keep things simple.

Issue #3183, opened 2021-10-20
Hell tile with invisible collision


agris-codes:
**Important information**
Win10 / compiled x64 myself, almost-vanilla build from ~1 week ago

**Describe**
It appears that this Hell tile with a body has collision set incorrectly, such that missiles impact it and the player can't walk there.

**To Reproduce**
Load this save, observe missiles impacting mid-air. 
[single_8 - hell_tile.zip](https://github.com/diasurgical/devilutionX/files/7383670/single_8.-.hell_tile.zip)

**Expected behavior**
No collision. 

**Screenshots**
DevilutionX
![DX_Hell](https://user-images.githubusercontent.com/39627597/138155602-c3940945-1391-4117-8970-490cea6ee8c1.jpg)

Vanilla 1.09b
![v109b_Hell](https://user-images.githubusercontent.com/39627597/138155607-d60b1141-bd50-4ee0-9eb7-865b73799056.png)

qndel:
Turns out these bodies are "objects" and they are so weird that to end up aligned with the wall, they have to be placed 1 tile in front of it and got collision set to true

PR #3185, opened 2021-10-20
fix collision of bodies in butcher chamber/hell

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/3183


ephphatha (approved):
Looks good to me, was able to walk past (through) the tortured souls/bodies on the wall in hell.

These objects/sprites appear to be aligned to the left of a tile so this only really impacts bodies on the northwest wall of a room, bodies on the northeast wall are placed inside the wall so the collision doesn't matter.

qndel:
> Looks good to me, was able to walk past (through) the tortured souls/bodies on the wall in hell.
> 
> These objects/sprites appear to be aligned to the left of a tile so this only really impacts bodies on the northwest wall of a room, bodies on the northeast wall are placed inside the wall so the collision doesn't matter.

Easiest way to test most of them is butcher's chamber :)

PR #3188, opened 2021-10-20
Enable ADPCM and Huffman compression

AJenbo:
This allows for huge savings in the size of dubbing packs, from 397MB to down to just 134MB.

Support for this is often also expected as it was added to Diablo 1.09 as a result of storm.dll being updated, so this should also provide better compatibility with existing user content.


AJenbo:
Kinda need this for testing testing automatic fetching of content on android, so merging since it passes CI...

Issue #3187, opened 2021-10-20
Scrolls don't work if spell has 8+ level


UmmonPwr:
**Important information**
OS: Win10
devilutionX version: x86 / both own build and the latest test build from GitHub. I did not check x64 builds.
Issue could have been introduced somewhere at the beginning of September. At least my build from September 15th already have this bug.

**Describe**
Town portal is not opened when casted from scroll.

**To Reproduce**
When right clicking on town portal scroll in inventory it disappears, cursor changes but no portal can be opened.
When casting town portal from spells it is working OK.

UmmonPwr:
[single_0.zip](https://github.com/diasurgical/devilutionX/files/7386004/single_0.zip)
Attached my savegame, just in case.

PR #3193, opened 2021-10-21
fix scrolls

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/3187


AJenbo:
Probably this property should never have been marked as a direction since it is used for other things :(

PR #3199, opened 2021-10-21
Allow to customize shown intro in diablo.ini

obligaron:
Extend `Intro` option in `diablo.ini` with:

- **2**: No Intro  (equivalent to `-n` command line parameter)
- **3**: Show Title Dialog only (I like the title dialog 😉)

With this another alternative to command line arguments is present. This is relevant for platforms that don't support them (for example android).

Note: existing values (0 = no cinematic and 1 = initial show cinematic once (at first start)) stay the same for compatibility.

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -565,6 +565,18 @@ string_view OptionCategoryBase::GetDescription() const
 
 StartUpOptions::StartUpOptions()
     : OptionCategoryBase("StartUp", N_("Start Up"), N_("Start Up Settings"))
+    , diabloIntro("Diablo Intro", OptionEntryFlags::OnlyDiablo, N_("Intro"), N_("Shown Intro cinematic."), StartUpIntro::Once,
+          {
+              { StartUpIntro::Off, N_("OFF") },
+              // Once is missing, cause we want to hide it from UI-Settings.
+              { StartUpIntro::On, N_("ON") },
+          })
+    , hellfireIntro("Hellfire Intro", OptionEntryFlags::OnlyHellfire, N_("Intro"), N_("Shown Intro cinematic."), StartUpIntro::Once,
+          {
+              { StartUpIntro::Off, N_("OFF") },
+              // Once is missing, cause we want to hide it from UI-Settings.
+              { StartUpIntro::On, N_("ON") },
+          })
     , splash("Splash", OptionEntryFlags::None, N_("Splash"), N_("Shown splash screen."), StartUpSplash::LogoAndTitleDialog,
           {
               { StartUpSplash::LogoAndTitleDialog, N_("Logo and Title Screen") },
@@ -576,32 +588,28 @@ StartUpOptions::StartUpOptions()
 std::vector<OptionEntryBase *> StartUpOptions::GetEntries()
 {
 	return {
+		&diabloIntro,
+		&hellfireIntro,
 		&splash,
 	};
 }
 
 DiabloOptions::DiabloOptions()
     : OptionCategoryBase("Diablo", N_("Diablo"), N_("Diablo specific Settings"))
-    , intro("Intro", OptionEntryFlags::OnlyDiablo, N_("Intro"), N_("Enable/disable Intro cinematic."), true)
 {
 }
 std::vector<OptionEntryBase *> DiabloOptions::GetEntries()
 {
-	return {
-		&intro,
-	};
+	return {};
 }
 
 HellfireOptions::HellfireOptions()
     : OptionCategoryBase("Hellfire", N_("Hellfire"), N_("Hellfire specific Settings"))
-    , intro("Intro", OptionEntryFlags::OnlyHellfire, N_("Intro"), N_("Enable/disable Intro cinematic."), true)
 {
 }
 std::vector<OptionEntryBase *> HellfireOptions::GetEntries()
 {
-	return {
-		&intro,
-	};
+	return {};
 }
 
 AudioOptions::AudioOptions()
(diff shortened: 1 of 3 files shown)


obligaron:
Rebased on new settings menu:

grafik

Source/options.cpp
```diff
@@ -581,24 +581,38 @@ string_view OptionCategoryBase::GetDescription() const
 DiabloOptions::DiabloOptions()
     : OptionCategoryBase("Diablo", N_("Diablo"), N_("Diablo specific Settings"))
     , intro("Intro", OptionEntryFlags::OnlyDiablo, N_("Intro"), N_("Enable/disable Intro cinematic."), true)
+    , splash("Splash", OptionEntryFlags::OnlyDiablo, N_("Splash"), N_("Shown splash screen at startup."), StartUpSplash::LogoAndTitleDialog,
+          {
+              { StartUpSplash::LogoAndTitleDialog, N_("Logo and Title Dialog") },
+              { StartUpSplash::TitleDialog, N_("Title Dialog") },
```
AJenbo, line 587:
```suggestion
              { StartUpSplash::LogoAndTitleDialog, N_("Logo and Title Screen") },
              { StartUpSplash::TitleDialog, N_("Title Screen") },
```

Source/options.cpp
```diff
@@ -581,24 +581,38 @@ string_view OptionCategoryBase::GetDescription() const
...
 
 HellfireOptions::HellfireOptions()
     : OptionCategoryBase("Hellfire", N_("Hellfire"), N_("Hellfire specific Settings"))
     , intro("Intro", OptionEntryFlags::OnlyHellfire, N_("Intro"), N_("Enable/disable Intro cinematic."), true)
+    , splash("Splash", OptionEntryFlags::OnlyHellfire, N_("Splash"), N_("Shown splash screen at startup."), StartUpSplash::LogoAndTitleDialog,
+          {
+              { StartUpSplash::LogoAndTitleDialog, N_("Logo and Title Dialog") },
+              { StartUpSplash::TitleDialog, N_("Title Dialog") },
+              { StartUpSplash::None, N_("None") },
+          })
```
AJenbo, line 608: I don't feel there is a need for this per game option, the one in `sgOptions.Diablo.splash` will suffice.

Source/options.cpp
obligaron: Fixed 🙂

Source/options.cpp
obligaron: This mimic what we currently have for the `intro` option.

I have no problem with changing it, only two questions:
- Should we let it stay in `Diablo` section or move it somewhere else? Cause I thought that the `Diablo` and `Hellfire` settings are specific for the corresponding game.
- Should we also move/combine the `intro` setting for diablo and hellfire?

AJenbo: its fine to have it in the Diablo section, else we would have to create a new one (which is also fine)

intos are different an automatically change setting, leave them seperate

obligaron: Currently intro is [not automatically](https://github.com/diasurgical/devilutionX/pull/3437#issuecomment-964590894) changed.
> Intro changed to on/off option (default on) and will be on until it's switched in the settings menu
Did I misinterpreted the comments ([Nr1](https://github.com/diasurgical/devilutionX/pull/3437#issuecomment-962715359) and [Nr2](https://github.com/diasurgical/devilutionX/pull/3437#issuecomment-962760168)) wrong?

What do you think about a new category "Startup" (or another name) that combines intro, splash, (later) game mode (diablo or hellfire) and perhaps also fullversion and shareware? Then we wouldn't show the diablo or hellfire category in the settings menu at all.

AJenbo: I wasn't very specific and was more in general agreement, but not necessarily on all details of the comment. Sorry for not giving more details :)

Some thing like this could work
0: off
1: play once
2: always on

But I doubt anyone would use 2, so that is optional imo

> What do you think about a new category "Startup" (or another name) that combines intro, splash, (later) game mode (diablo or hellfire) and perhaps also fullversion and shareware? Then we wouldn't show the diablo or hellfire category in the settings menu at all.

Sounds good.

obligaron: I moved intro and splash to a new `StartUp`-Category that don't make a difference between diablo and hellfire.
Intro has now a "Once"-Option (Isn't visible in UI-Settings. My thought was, that if a user want's to explicitly enable it, the user always wants to enable it and not only once; can change it if wanted).
Diablo and Hellfire category is not visible in UI-Settings.

Source/options.cpp
```diff
@@ -579,28 +584,53 @@ string_view OptionCategoryBase::GetDescription() const
...
+    : OptionCategoryBase("StartUp", N_("Start Up"), N_("Start Up Settings"))
+    , diabloIntro("Diablo Intro", OptionEntryFlags::OnlyDiablo, N_("Intro"), N_("Shown Intro cinematic."), StartUpIntro::Once,
+          {
+              { StartUpIntro::None, N_("None") },
+              // Once is missing, cause we want to hide it from UI-Settings.
+              { StartUpIntro::Always, N_("Always") },
+          })
+    , hellfireIntro("Hellfire Intro", OptionEntryFlags::OnlyHellfire, N_("Intro"), N_("Shown Intro cinematic."), StartUpIntro::Once,
+          {
+              { StartUpIntro::None, N_("None") },
```
AJenbo, line 597:
```suggestion
              { StartUpIntro::None, N_("Never") },
```

Might want to change the enum also

AJenbo: No/Yes would also work

obligaron: I changed to OFF/ON, cause we use it already and so there are less text to translate. 🙂

AJenbo (approved):

PR #3211, opened 2021-10-22
add megatiles to tiledata

qndel:
image
`tiledata dflags`


Source/scrollrt.cpp
```diff
@@ -1206,9 +1206,15 @@ void DrawView(const Surface &out, Point startPosition)
 		// force redrawing or debug stuff stays on panel on 640x480 resolution
 		force_redraw = 255;
 		char debugGridTextBuffer[10];
+		bool megaTiles = IsDebugGridInMegatiles();
```
AJenbo, line 1209: It's starting to introducing a few debug features in this file that aren't left out for release builds, so it would be good if we could clean things up a bit in the near future.

AJenbo (approved):

PR #3212, opened 2021-10-22
Also look for translation files in `BasePath`

glebm:
This makes sure translations are loaded even when launched as `build/devilutionx`.


AJenbo:
`SFileOpenRw()`, but then we need to also move the files to `./assets` / add them to the mpq

PR #3249, opened 2021-10-25
Translate remaining french

mmaret:
There is 2 translations that may need particular attention.
* Dunsel
It seems to be a Joke that mean "Useless Captain" or something like that. But just translating it into "Inutile" loose the Joke.
So I kept Dunsel.

* New Ability (NW)
I don't get what "NW" was refering too. So I just kept it as it.

### DIFF
--- a/Translations/fr.po
+++ b/Translations/fr.po
@@ -66,7 +66,7 @@ msgstr "Illustration"
 
 #: Source/DiabloUI/credits_lines.cpp:48
 msgid "Technical Artwork"
-msgstr ""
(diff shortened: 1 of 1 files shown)


AJenbo:
@mmaret @Phalcon22 just wanted you guys to know that your work is being appreciated by players :)

> Absolument génial, en français, je le trouve un peux sombre

mmaret:
Thx for the feedback!

Phalcon22:
Thank you! I'm really glad our work makes some people happy ^^

PR #3263, opened 2021-10-25
Fix ZeroTier crash when handling invalid packet

StephenCWills:
This fixes the crash on ZeroTier when processing an invalid packet.

### DIFF
--- a/Source/dvlnet/base_protocol.h
+++ b/Source/dvlnet/base_protocol.h
@@ -262,7 +262,8 @@ void base_protocol<P>::recv_ingame(packet &pkt, endpoint sender)
 		return;
 	} else if (pkt.Source() >= MAX_PLRS) {
 		// normal packets
-		ABORT();
+		LogDebug("Invalid packet: packet source ({}) >= MAX_PLRS", pkt.Source());
+		return;
 	}
 	connected_table[pkt.Source()] = true;
 	peers[pkt.Source()] = sender;


ukos-git:
I am getting a lot of these Errors and also see ZT crashing from time. What exactly is `MAX_PLRS`?
<details>
  <summary>Invalid packet: packet source ({}) >= MAX_PLRS</summary>

```
DEBUG: Invalid packet: packet source (205) >= MAX_PLRS
DEBUG: Invalid packet: packet source (146) >= MAX_PLRS
DEBUG: Invalid packet: packet source (9) >= MAX_PLRS
DEBUG: Invalid packet: packet source (175) >= MAX_PLRS
DEBUG: Invalid packet: packet source (13) >= MAX_PLRS
DEBUG: Invalid packet: packet source (23) >= MAX_PLRS
DEBUG: Invalid packet: packet source (72) >= MAX_PLRS
DEBUG: Invalid packet: packet source (189) >= MAX_PLRS
DEBUG: Invalid packet: packet source (246) >= MAX_PLRS
DEBUG: Invalid packet: packet source (197) >= MAX_PLRS
DEBUG: Invalid packet: packet source (102) >= MAX_PLRS
DEBUG: Invalid packet: packet source (49) >= MAX_PLRS
DEBUG: Invalid packet: packet source (247) >= MAX_PLRS
```

</details>

yuripourre:
@ukos-git MAX_PLRS is a constant that has the number of players allowed per game (locked at 4)

ukos-git:
so this means, the packages indicate a player count of 247 in my game. Sounds like a poor data protocol.

StephenCWills:
> so this means, the packages indicate a player count of 247 in my game. Sounds like a poor data protocol.

It means you received a packet from player 247, which is invalid since there can only be four players in your game. I imagine the player ID is being interpreted as an 8-bit value, meaning the number can fall anywhere in the range of 0 to 255. It sounds like your client is simply processing invalid packets. I'm not sure where the invalid packets could be coming from.

Rather than posting in a PR that was merged long ago, I'd recommend creating an issue and linking to the relevant PR.

ukos-git:
You are right, I will make an issue. 
But rather than fixing the root cause of the crash you just converted it to a debug message in this PR. I am thankful for the debug output, though, as the current release 1.3 will simply crash.

AJenbo:
Malicious actors sending non nonsensical packages isn't something you can simply fix.

The issue you are having may well not be what this PR targeted in the first place.

StephenCWills:
> But rather than fixing the root cause of the crash you just converted it to a debug message in this PR.

We actually did fix the root cause at the time. This change just prevents the client from immediately killing itself when it receives a bad packet, which is clearly undesirable behavior.

> I am thankful for the debug output, though, as the current release 1.3 will simply crash.

This change is in 1.3. If 1.3 is crashing without displaying debug output, that would be relevant information to include in the issue report.

PR #3287, opened 2021-10-26
Apply suggested changes by Android Studio

tsunamistate:
* Upgrade Gradle plugin to 7.0.3
* Upgrade `compileSdkVersion` and `targetSdkVersion` to 30. Version 31 move is not as seamless as I though and requires code changes, that I don't know how to do 😞
* Migrate from `aaptOptions` to `androidResources`: [noCompress still works, more here](https://developer.android.com/reference/tools/gradle-api/7.0/com/android/build/api/dsl/AndroidResources?hl=en)
* Remove semicolons since they are unnecessary and applied inconsistently


tsunamistate:
Happy belated Halloween!
Removing draft since it's November now

AJenbo (approved):

Issue #3219, opened 2021-10-23
[Hellfire] Blocked arrows always rotate south


StephenCWills:
In Hellfire, when an arrow is blocked, it is always rotated south as it continues on past the player that blocked it.

**To Reproduce**
1. Load the following save: [single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/7401987/single_0.hsv.zip)
2. Walk forward and start drinking from the Blood Fountain
3. Observe how every arrow that the Monk blocks gets rotated toward the south

**Expected behavior**
In Hellfire, when projectiles are blocked, the game picks a random number to determine whether to rotate the missile left or right. The same should happen for arrows.

**Additional context**
The following logic attempts to rotate the missile in a random (adjacent) direction.

https://github.com/diasurgical/devilutionX/blob/b32e234578a393154f2a1e7191e50d5cc8e509c4/Source/missiles.cpp#L471-L479

Since Firebolt has a changing animation, the `animData` array consists of sixteen different animations for all the different angles a Firebolt can fly in. The `SetMissDir()` function simply selects the appropriate animation from the `animData` array based on the direction passed into that function.

Arrows, on the other hand, have a static animation. Instead of having sixteen animations each with one frame, they were consolidated into one animation with sixteen frames. Instead, their direction is set by explicitly selecting the appropriate frame in `AddArrow()` with the following line of code.

https://github.com/diasurgical/devilutionX/blob/b32e234578a393154f2a1e7191e50d5cc8e509c4/Source/missiles.cpp#L1833

From a quick search, it appears to me as if arrows are the only projectile to which this special case applies.

PR #3296, opened 2021-10-26
Use a temporary seed for RNG when operating fountains

StephenCWills:
* Based on @qndel's suggestion, implement a class for RAII to temporarily override the current seed
* Use a temporary seed in `OperateFountains()` so that using the fountain doesn't overwrite engine RNG

I noticed when testing #3219 that spamming the Blood Fountain caused all the randomness in monster attack frequency, hit rate, and block chance to go away. It seems that Fountain of Tears captures a seed to provide predictable behavior when used regardless of the current state of the engine's RNG, but the call to `SetRndSeed()` doesn't discriminate based on the type of fountain. Even if it did discriminate, using a Fountain of Tears would overwrite RNG state. It makes more sense to use a temporary seed here.

### DIFF
--- a/Source/engine/random.cpp
+++ b/Source/engine/random.cpp
@@ -47,4 +47,15 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
+TemporarySeed::TemporarySeed(uint32_t seed)
+{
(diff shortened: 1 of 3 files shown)


tsunamistate:
This should be closed since it was superseded by #3314

PR #3310, opened 2021-10-28
Introduce a replacement/abstraction for CrawlTable for single tile searches

ephphatha:
With the way CrawlTable and CrawlNum are defined callers needed to do some awkward iteration and manually construct displacements. This also meant that the logic performed when a valid target was found was mixed in with the actual search code since often there were two levels of loop to break out of.

I've added a function to abstract this table, using an interface similar to the pathfinding logic. I've also placed it in path.h/cpp as it seemed most appropriate given the existing use.

There are other functions which search for multiple tiles, this could be supported using a custom iterator but I wanted to focus on the simple case for now.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2071,32 +2071,40 @@ void AddWeapexp(Missile &missile, Point dst, Direction /*midir*/)
 
 void AddTown(Missile &missile, Point dst, Direction /*midir*/)
 {
-	Point target = dst;
-	if (currlevel != 0) {
-		missile._miDelFlag = true;
-		for (int i = 0; i < 6; i++) {
-			int k = CrawlNum[i];
-			int ck = k + 2;
-			for (auto j = static_cast<uint8_t>(CrawlTable[k]); j > 0; j--, ck += 2) {
-				target = dst + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
-				if (!InDungeonBounds(target))
-					continue;
-
-				int dp = dPiece[target.x][target.y];
-				if (!TileContainsMissile(target) && !nSolidTable[dp] && !nMissileTable[dp] && dObject[target.x][target.y] == 0 && dPlayer[target.x][target.y] == 0) {
-					if (!CheckIfTrig(target)) {
-						missile.position.tile = target;
-						missile.position.start = target;
-						missile._miDelFlag = false;
-						i = 6;
-						break;
-					}
-				}
-			}
-		}
+	if (currlevel == 0) {
+		missile.position.tile = dst;
+		missile.position.start = dst;
 	} else {
-		missile.position.tile = target;
-		missile.position.start = target;
+		std::optional<Point> targetPosition = FindClosestValidPosition(
+		    [](Point target) {
+			    if (!InDungeonBounds(target)) {
+				    return false;
+			    }
+			    if (dObject[target.x][target.y] != 0) {
+				    return false;
+			    }
+			    if (dPlayer[target.x][target.y] != 0) {
+				    return false;
+			    }
+			    if (TileContainsMissile(target)) {
+				    return false;
+			    }
+
+			    int dp = dPiece[target.x][target.y];
+			    if (nSolidTable[dp] || nMissileTable[dp]) {
+				    return false;
+			    }
+			    return !CheckIfTrig(target);
+		    },
+		    dst, 0, 5);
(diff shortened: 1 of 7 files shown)


Source/path.cpp
```diff
@@ -5,13 +5,426 @@
...
+#include <array>
+
 #include "gendung.h"
 #include "objects.h"
 
 namespace devilution {
 
 namespace {
 
+constexpr std::array<Displacement, 1369> CrawlOffsets {
```
AJenbo, line 17: Extracting the counts was a nice improvement, but I would be nice if we could also make the bounds explicit instead of them just being arbitrary offsets in to a huge array.

Also what is left to be converted before we can fully drop the old array?

ephphatha: The corners are the only exceptional case. If it wasn't slightly rounded we could just use the loop logic everywhere.

FireRing uses the rounding to make the effect look nice, all other uses of CrawlTable are used to find targets so squaring off the corners shouldn't cause visible impact. That would be StealPotions (used to select a target player), ChainLightning (used to select monsters to target, the visual effect is part of the missile movement logic) and the debug spawn monsters command (used to find free spaces).

Having a think about it, I can probably update the loop logic to inset the corners in the same way as the crawl table currently does so we could remove all the hardcoded values. This would mean that offsets +/-50, +/-50 are never checked but given searches that large are unlikely to have a completely occupied search space it shouldn't be a massive issue.

ephphatha:
Introduced a bug when I was removing the hardcoded offsets so I've expanded the tests a bit to cover the special cases.

The unoptimised loop this logic replaces only went to 49 anyway so I think putting a cap of 50 is reasonable to include the original search space + a bit extra.

AJenbo (approved):

AJenbo:
Looks really good, since there is a lot of changes I would like a second pare of eyes to review it as well. Maybe @obligaron would be willing to do so :)

Source/missiles.cpp
```diff
@@ -2075,32 +2069,40 @@ void AddWeapexp(Missile &missile, Point dst, Direction /*midir*/)
...
+			    return !CheckIfTrig(target);
+		    },
+		    dst, 0, 5);
+
+		if (targetPosition) {
+			missile.position.tile = *targetPosition;
+			missile.position.start = *targetPosition;
+			missile._miDelFlag = false;
+		} else {
+			missile._miDelFlag = true;
```
obligaron, line 2104: nit:
Should we update/set `missile.position.tile` here?
In the old logic `dst` was used for `NetSendCmdLocParam3` and now we use `missile.position.tile`. I'm not sure if both are the same.
nit cause I'm not sure if we ever hit this case or if it's relevant, cause `_miDelFlag` is set to `true`.

obligaron (approved):
lgtm 👍 
thank you for the nice clean up 🙂

Source/missiles.cpp
ephphatha: CMD_ACTIVATEPORTAL is only sent if _miDelFlag is false, in which case missile.position.tile has been set to target. PutMissile doesn't modify position (but may set delFlag to true) so I felt it was safe (and more descriptive) to show that the portal is being spawned where the missile is currently located.

Source/missiles.cpp
```diff
@@ -83,22 +83,18 @@ bool CheckBlock(Point from, Point to)
...
-		for (auto j = static_cast<uint8_t>(CrawlTable[k]); j > 0; j--, ck += 2) {
-			Point target = source + Displacement { CrawlTable[ck - 1], CrawlTable[ck] };
-			if (!InDungeonBounds(target))
-				continue;
+	std::optional<Point> monsterPosition = FindClosestValidPosition(
+	    [&source](Point target) {
+		    // search for a monster with clear line of sight
+		    return InDungeonBounds(target) && dMonster[target.x][target.y] > 0 && !CheckBlock(source, target);
+	    },
+	    source, 1, rad);
```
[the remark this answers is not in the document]
ephphatha: The check here looked to be only for ensuring the read into CrawlNum and CrawlTable didn't go out of bounds, given the new function handles larger radiuses I let it pass through (it would've been more correct to use 18 to match the old behaviour exactly).

It did seem useful to have the search radius as a parameter for this FindClosestMonster function and it didn't look critical that we match the behaviour exactly here, but I'm happy to update the call sites to use 18 instead.

PR #3319, opened 2021-10-28
Add Ukrainian translation (Main Menu)

tsunamistate:
This adds Ukrainian TL to CMake and Translations folder.
TL'd Main Menu to start with.
Reviews very welcome!

Conscious choices (which I may regret):
* Rouge as "Пройдисвіт". Maybe it should be genderized ("Пройдисвітка")?
* Multiplayer as "Гра з Друзями". Transliterating "Multiplayer" really irks me, I only did it in Russian TL because I couldn't think of an alternative. Maybe "Мережева Гра" is better, but I don't know what is the precedent in Ukrainian localizations.
* Omitting "Singleplayer" TL altogether except in Main Menu. This word has no precedent except for transliterating it, AFAIK, so "New Single Player Hero" becomes just "Новий Герой".
* "Hero" and "Character" are TL'd as "Герой" in both cases.
* "Delete" as "Стерти". Any other meaning was too long in UI.

For reviewers, start from second commit, to isolate newly added strings.

### DIFF
--- a/android-project/app/src/main/res/values-uk/strings.xml
+++ b/android-project/app/src/main/res/values-uk/strings.xml
@@ -1,0 +1,14 @@
+<resources>
+	<string name="app_name" translatable="false">DevilutionX</string>
+	<!-- Strings used for fragments for navigation -->
+
+	<string name="guide_full_installation">Для того щоб зіграти в повну версію, Вам треба помістити файл з оригінальної гри DIABDAT.MPQ в папку Android/data/org.diasurgical.devilutionx/files.<br/>\n<br/>\nЯкщо у вас немає повної версії, то її можна купити в <a href="https://www.gog.com/game/diablo">GOG.com.</a></string>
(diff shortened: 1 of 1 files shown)


SourceCodeDeleted:
So does this mean there is a Ukrainian version @tsunamistate ?

AJenbo:
Yes, it's 5% complete

tsunamistate:
@SourceCodeDeleted there's no existing Ukrainian TL, if that's what you're asking. This is a new initiative by me. Hopefully I'll make it to 1.4.0.
More games need to be in Ukrainian, since you can count official localizations on one hand!

SourceCodeDeleted:
More games need to be in Ukrainian, since you can count official localizations on one hand!

I agree ) 

I started on the russian one , but didn't.do anything with it when I didn't think we would have genders correct for words. However this is no longer the case so I think we can make this quickly.

PR #3324, opened 2021-10-29
Use helper function for Object by position lookups

ephphatha:
This already existed but had limited usage (and also only handled known positions). This PR expands the functionality and adds helpers to allow using position based tests.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -396,11 +396,11 @@ bool ItemPlace(Point position)
 		return false;
 	if (dItem[position.x][position.y] != 0)
 		return false;
-	if (dObject[position.x][position.y] != 0)
+	if (IsObjectAtPosition(position))
 		return false;
 	if (TileContainsSetPiece(position))
 		return false;
-	if (nSolidTable[dPiece[position.x][position.y]])
+	if (IsTileSolid(position))
 		return false;
 
 	return true;
@@ -3106,44 +3106,31 @@ void CreatePlrItems(int playerId)
 
 bool ItemSpaceOk(Point position)
 {
-	if (!InDungeonBounds(position))
+	if (!InDungeonBounds(position)) {
 		return false;
-
-	if (dMonster[position.x][position.y] != 0)
-		return false;
-
-	if (dPlayer[position.x][position.y] != 0)
-		return false;
-
-	if (dItem[position.x][position.y] != 0)
-		return false;
-
-	if (dObject[position.x][position.y] != 0) {
-		int oi = abs(dObject[position.x][position.y]) - 1;
-		if (Objects[oi]._oSolidFlag)
-			return false;
 	}
 
-	Point south = position + Direction::South;
-	if (InDungeonBounds(south)) {
-		int objectId = dObject[south.x][south.y];
-		if (objectId != 0 && Objects[abs(objectId) - 1]._oSelFlag != 0)
-			return false;
+	if (IsTileSolid(position)) {
+		return false;
 	}
 
-	Point southEast = position + Direction::SouthEast;
-	Point southWest = position + Direction::SouthWest;
-	if (InDungeonBounds(southEast) && InDungeonBounds(southWest)) {
-		int objectIdSE = dObject[southEast.x][southEast.y];
-		int objectIdSW = dObject[southWest.x][southWest.y];
-		if (objectIdSE > 0 && objectIdSW > 0) {
-			if (Objects[objectIdSE - 1]._oSelFlag != 0 && Objects[objectIdSW - 1]._oSelFlag != 0) {
-				return false;
-			}
-		}
+	if (dItem[position.x][position.y] != 0) {
+		return false;
 	}
 
-	return IsTileNotSolid(position);
+	if (dMonster[position.x][position.y] != 0) {
+		return false;
+	}
+
+	if (dPlayer[position.x][position.y] != 0) {
+		return false;
+	}
+
+	if (IsItemBlockingObjectAtPosition(position)) {
+		return false;
+	}
+
+	return true;
 }
 
 int AllocateItem()
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -547,38 +547,19 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
 			}
 		}
 	}
-	if (dObject[mx][my] != 0) {
-		Object &object = Objects[abs(dObject[mx][my]) - 1];
-		if (!object._oMissFlag) {
-			if (object.IsBreakable()) {
-				BreakObject(-1, object);
-			}
-			if (!nodel)
-				missile._mirange = 0;
-			missile._miHitFlag = false;
+	if (IsMissileBlockedByTile({ mx, my })) {
+		Object *object = ObjectAtPosition({ mx, my });
+		if (object != nullptr && object->IsBreakable()) {
+			BreakObject(-1, *object);
 		}
-	}
-	if (nMissileTable[dPiece[mx][my]]) {
+
 		if (!nodel)
 			missile._mirange = 0;
 		missile._miHitFlag = false;
 	}
+
 	if (missile._mirange == 0 && MissilesData[missile._mitype].miSFX != -1)
 		PlaySfxLoc(MissilesData[missile._mitype].miSFX, missile.position.tile);
-}
-
-/**
- * @brief Could the missile collide with solid objects? (like walls or closed doors)
- */
-bool CouldMissileCollideWithSolidObject(Point tile)
-{
-	int oid = dObject[tile.x][tile.y];
-	if (oid != 0) {
-		oid = abs(oid) - 1;
-		if (!Objects[oid]._oMissFlag)
-			return true;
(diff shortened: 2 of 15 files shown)


ephphatha:
I think I was waiting for #3336 to be ready before I marked this one as ready but I've been putting off making the reverse iterator compile in that PR... Might as well get this one in progress now.

Source/objects.cpp
```diff
@@ -4349,6 +4350,60 @@ bool Object::IsDisabled() const
...
+		if (otherDoor != nullptr && object->_oSelFlag != 0 && otherDoor->_oSelFlag != 0) {
+			// open double tile door
+			return true;
+		}
+	}
+
+	return false;
+}
+
+bool IsMissileBlockingObjectAtPosition(Point position)
```
AJenbo, line 4400:
```suggestion
bool IsMissilesBlockedByObjectAtPosition(Point position)
```

ephphatha: This makes me think this function would be better defined in missiles.cpp. I'll update and move that definition.

Source/themes.cpp
```diff
@@ -712,10 +719,10 @@ void Theme_Library(int t)
...
 				AddObject(OBJ_BOOKSTAND, { xp, yp });
-				if (GenerateRnd(2 * librnd[leveltype - 1]) != 0 && dObject[xp][yp] != 0) { /// BUGFIX: check dObject[xp][yp] was populated by AddObject (fixed)
-					int oi = dObject[xp][yp] - 1;
-					Objects[oi]._oSelFlag = 0;
-					Objects[oi]._oAnimFrame += 2;
+				Object *bookstand = ObjectAtPosition({ xp, yp });
+				if (GenerateRnd(2 * librnd[leveltype - 1]) != 0 && bookstand != nullptr) { /// BUGFIX: check bookstand was actually created (fixed)
+					bookstand->_oSelFlag = 0;
+					bookstand->_oAnimFrame += 2;
 				}
```
AJenbo, line 726:
```suggestion
				if (GenerateRnd(2 * librnd[leveltype - 1]) != 0) {
					Object *bookstand = ObjectAtPosition({ xp, yp });
					if (bookstand != nullptr) {
						bookstand->_oSelFlag = 0;
						bookstand->_oAnimFrame += 2;
					}
				}
```

Source/objects.cpp
ephphatha: I ended up going with the name `IsMissileBlockedByTile`, renaming the existing `CouldMissileCollide` function. Every time a position was checked for an object with `missFlag` the code also checked `nMissileTable` so there was no point keeping them separate.

Source/missiles.cpp
```diff
@@ -547,40 +547,21 @@ void CheckMissileCol(Missile &missile, int mindam, int maxdam, bool shift, Point
...
+		if (object != nullptr && object->IsBreakable()) {
+			BreakObject(-1, *object);
 		}
-	}
-	if (nMissileTable[dPiece[mx][my]]) {
+
 		if (!nodel)
 			missile._mirange = 0;
 		missile._miHitFlag = false;
 	}
```
ephphatha, line 559: This isn't strictly equivalent to the old code, but I don't think there are any naturally occurring cases where a breakable object that allows missiles to pass through is on a tile which blocks missiles... that'd have to be something like a barrel inside a wall or set piece.

Source/objects.cpp
```diff
@@ -4350,6 +4351,53 @@ bool Object::IsDisabled() const
...
+	if (object != nullptr && object->_oSolidFlag) {
+		// solid object
+		return true;
+	}
+
+	object = ObjectAtPosition(position + Direction::South);
+	if (object != nullptr && object->_oSelFlag != 0) {
+		// open single tile door
+		return true;
+	}
```
AJenbo, line 4387: Previously it was possible to drop items in open doors, wouldn't this change that?

Source/objects.cpp
```diff
@@ -4350,6 +4351,53 @@ bool Object::IsDisabled() const
...
+
+	object = ObjectAtPosition(position + Direction::SouthEast, false);
+	if (object != nullptr) {
+		Object *otherDoor = ObjectAtPosition(position + Direction::SouthWest, false);
+		if (otherDoor != nullptr && object->_oSelFlag != 0 && otherDoor->_oSelFlag != 0) {
+			// open double tile door
```
AJenbo, line 4393: does this exist?

Source/objects.cpp
```diff
@@ -4350,6 +4351,53 @@ bool Object::IsDisabled() const
...
+
+	// nothing at this position, return a nullptr
+	return nullptr;
+}
+
+bool IsObjectAtPosition(Point position)
+{
+	return ObjectAtPosition(position) != nullptr;
+}
+
```
AJenbo, line 4374: trivial enough to live in the header?

Source/objects.cpp
```diff
@@ -587,39 +591,41 @@ void AddObjTraps()
...
 
 				if (!CanPlaceWallTrap(i, yp) || j - yp <= 1)
 					continue;
 
 				AddObject(OBJ_TRAPR, { i, yp });
-				int8_t oiTrap = dObject[i][yp] - 1;
-				Objects[oiTrap]._oVar1 = i;
-				Objects[oiTrap]._oVar2 = j;
-				Objects[oi]._oTrapFlag = true;
+				trapObject = ObjectAtPosition({ i, yp });
```
AJenbo, line 621: not required, but returning the object from `AddObject` would probably make sens

Source/themes.cpp
```diff
@@ -181,16 +182,22 @@ bool TFit_GoatShrine(int t)
 bool CheckThemeObj3(int xp, int yp, int t, int f)
 {
 	for (int i = 0; i < 9; i++) {
-		if (xp + trm3x[i] < 0 || yp + trm3y[i] < 0)
+		Point testPosition = Point { xp, yp } + Displacement { trm3x[i], trm3y[i] };
```
AJenbo, line 185: `trm3x`/`trm3y` are just begging to become an array of directions :)

Source/items.cpp
```diff
@@ -3106,44 +3106,31 @@ void CreatePlrItems(int playerId)
...
-	Point southEast = position + Direction::SouthEast;
-	Point southWest = position + Direction::SouthWest;
-	if (InDungeonBounds(southEast) && InDungeonBounds(southWest)) {
-		int objectIdSE = dObject[southEast.x][southEast.y];
-		int objectIdSW = dObject[southWest.x][southWest.y];
-		if (objectIdSE > 0 && objectIdSW > 0) {
-			if (Objects[objectIdSE - 1]._oSelFlag != 0 && Objects[objectIdSW - 1]._oSelFlag != 0) {
-				return false;
-			}
-		}
```
AJenbo, line 3143: Is all of this unessesery because the dObject will contain the negative id at the base position anyway?

AJenbo:
just a couple of questions and suggestions :)

Source/themes.cpp
ephphatha: That was the original approach in #3336 :D (then I overcomplicated it with the rectangle iterator, I'll get that working eventually)

Source/objects.cpp
ephphatha: `AddObject` needs work, it currently partially initialises an `Object` instance while also placing it on the map. I'd rather see it only place the object (similar to `container::push_back`) and have the caller fully initialise the object instead of having to do it in two stages like this.

Source/objects.cpp
ephphatha: This doesn't change the behaviour, you can still drop an item in an open door. You can't drop an item in the tile the door opens into (whether it's open or closed) which appears to be what this check is looking at, though I'm not sure why the existing code was looking at _oSelFlag...

ephphatha: And after testing this I realised I got the context for this completely wrong. These checks are looking for interactive objects (breakable barrels and unopened containers). I'll update the comments.

Source/objects.cpp
ephphatha: This check is actually looking for two objects along a horizontal diagonal (relative to the screen). The player isn't allowed to drop an item on the tile behind an interactive object, and also can't drop on the tile between/behind two interactive objects.

Source/items.cpp
ephphatha: Rolled into `IsItemBlockingObjectAtPosition`. This code was identical to the checks in inv.cpp `CanPut`. The `considerLargeObjects` flag is set to false for this check to replicate the positive index only behaviour.

Source/objects.cpp
ephphatha: Hmm, having said that this could be considered an analogue to `container::emplace_back` which returns a reference. I'll add that to the list for a future PR :)

Source/objects.cpp
AJenbo: _oSelFlag controls how many tiles the object be selectable in. So yeah it makes sens that it would prevent you from dropping items behind the area where you would instead be only able to height light the object and not the item.

AJenbo (approved):

PR #3328, opened 2021-10-30
Refactor DrawObject to avoid repeated Objects lookup

ephphatha:
The check if the object position is offset is not strictly necessary as a displacement of {0, 0} remains the same after the world/screen transform.

(Splitting this out from #3324 to make that PR a little more focused)


obligaron (approved):
lgtm 🙂

PR #3331, opened 2021-10-30
render town outside borders

qndel:
image

### DIFF
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -970,10 +970,10 @@ Point FillPattern(Point tilePosition)
 		doInit = false;
 		std::mt19937 BetterRng;
 		BetterRng.seed(666);
-		std::uniform_int_distribution<int32_t> dist(0, INT_MAX);
+		std::uniform_int_distribution<int> dist(0, patterns.size() - 1);
 		for (int x = 0; x < MAXDUNX; x++) {
 			for (int y = 0; y < MAXDUNY; y++) {
-				patternRandomizer[x][y] = dist(BetterRng) % patterns.size();
+				patternRandomizer[x][y] = dist(BetterRng);
 			}
 		}
 	}
--- a/Source/scrollrt.cpp
+++ b/Source/scrollrt.cpp
@@ -3,6 +3,8 @@
  *
  * Implementation of functionality for rendering the dungeons, monsters and calling other render routines.
  */
+#include <climits>
+#include <random>
 
 #include "automap.h"
 #include "controls/touch/renderers.h"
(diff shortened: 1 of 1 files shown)


Chance4us:
Oh that's great. Let's place some monsters and quests over the river.

AJenbo:
Open world!

qndel:
> Oh that's great. Let's place some monsters and quests over the river.

that area is not walkable, it's fake :)

NiteKat:
> > Oh that's great. Let's place some monsters and quests over the river.
> 
> that area is not walkable, it's fake :)

I say this is proof enough that this is a good enough PR and accomplishes the job it's meant to. 😆

Source/scrollrt.cpp
```diff
@@ -916,6 +916,93 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
...
+		else if (IsAnyOf(tilePosition.x, 3, 63, 83))
+			pos = { 69, 22 - abs(tilePosition.y % 3) };
+		else
+			pos = FillPattern(tilePosition);
+	}
+
+	//bottom side
+	else if (tilePosition.y > 95) {
+		//sand patch
+		if (tilePosition.y == -1 && tilePosition.x >= 53 && tilePosition.x <= 60) {
```
ephphatha, line 979: This is dead code, the `tilePosition.y == -1` condition will never be true and there's no sand patch on that side that needs to be closed.

Source/scrollrt.cpp
```diff
@@ -916,6 +916,93 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
...
+		else
+			pos = FillPattern(tilePosition);
+	}
+
+	//left side
+	else if (tilePosition.x < 0)
+		pos = FillPattern(tilePosition);
+
+	//right side
+	else if (tilePosition.x > 95)
```
ephphatha, line 1000: There's a large sand patch here that should be closed off to match what you're doing for the top side

Source/scrollrt.cpp
```diff
@@ -916,6 +916,93 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
 	}
 }
 
+bool randomize = true;
+#define PATTERNWIDTH 3
+#define PATTERNHEIGHT 5
+int patternRandomizer[MAXDUNX][MAXDUNY];
```
ephphatha, line 922: These can be static/constexpr variables inside FillPattern

ephphatha (commented):
Makes it look much nicer at larger resolutions :)

Using positions instead of tile IDs is unfortunate, but I see why that's necessary given the way DrawFloor currently works.

I noticed a lot of the conditions in RenderTownBorders were selecting the same set of tiles so I had a go at collapsing the common conditions, https://github.com/qndel/devilutionX/commit/96e87a2692543be03555bfde06a05245dc574b2f for the diff if it's of use.

Source/scrollrt.cpp
qndel: > This is dead code, the `tilePosition.y == -1` condition will never be true and there's no sand patch on that side that needs to be closed.

good point, seems I copy/pasted too much and just ignored it : D

Source/scrollrt.cpp
```diff
@@ -916,6 +918,139 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
...
+	static bool doInit = true;
+
+	if (doInit) {
+		doInit = false;
+		std::mt19937 BetterRng;
+		BetterRng.seed(666);
+		std::uniform_int_distribution<int32_t> dist(0, INT_MAX);
+		for (int x = 0; x < MAXDUNX; x++) {
+			for (int y = 0; y < MAXDUNY; y++) {
+				patternRandomizer[x][y] = dist(BetterRng) % patterns.size();
```
ephphatha, line 976: Can use the distribution to get a value in the expected range directly instead of needing to use mod division. This also removes the need to `#import <climits>`
```suggestion
		std::uniform_int_distribution<int> dist(0, patterns.size() - 1);
		for (int x = 0; x < MAXDUNX; x++) {
			for (int y = 0; y < MAXDUNY; y++) {
				patternRandomizer[x][y] = dist(BetterRng);
```

Source/scrollrt.cpp
```diff
@@ -916,6 +918,139 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
...
+	constexpr int PATTERNHEIGHT = 5;
+	constexpr std::array<Point, 6> patterns { { { 51, 27 }, { 23, 57 }, { 54, 34 }, { 79, 46 }, { 41, 32 }, { 52, 21 } } };
+
+	static int patternRandomizer[MAXDUNX][MAXDUNY] {};
+	static bool doInit = true;
+
+	if (doInit) {
+		doInit = false;
+		std::mt19937 BetterRng;
+		BetterRng.seed(666);
```
ephphatha, line 972: Using a constant seed means the same pattern will be generated each run of the game, previously this would change each time the game was launched?

If you wanted it to be random on load can use `std::random_device` to init the engine
```suggestion
		std::random_device rd;
		std::mt19937 BetterRng(rd());
```

qndel: I wanted it to be the same

Source/scrollrt.cpp
```diff
@@ -3,6 +3,8 @@
  *
  * Implementation of functionality for rendering the dungeons, monsters and calling other render routines.
  */
+#include <climits>
```
qndel, line 6:
```suggestion
```

Source/scrollrt.cpp
```diff
@@ -916,6 +917,139 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
...
+}
+
+/**
+ * @brief Render some semi random patterns beyond town borders instead of black void
+ */
+
+void RenderTownBorders(const Surface &out, Point tilePosition, Point targetBufferPosition, bool isDungeon)
+{
+	if (leveltype != DTYPE_TOWN)
+		return;
```
ephphatha, line 1036: This blocks the cases where `isDungeon == true`

Source/scrollrt.cpp
```diff
@@ -916,6 +917,139 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
...
+	patternBaseTile.x /= PATTERNWIDTH;
+	patternBaseTile.y /= PATTERNHEIGHT;
+
+	Point patternPos = CheckPredefinedPattern(tilePosition);
+	if (patternPos != Point { 0, 0 }) {
+		if (InDungeonBounds(patternBaseTile)) {
+			patternRandomizer[patternBaseTile.x][patternBaseTile.y] = -1;
+		}
+		return patternPos;
+	}
```
ephphatha, line 998: CheckPredefinedPattern deal with the special cases for town so the border looks right, this condition can be skipped entirely for dungeon layers.

The code which sets a cell to -1 can also be removed, you're returning patternPos here so the cell in patternRandomizer is not read (with the current logic). If you did end up skipping this block but you kept that logic in then if a save game is loaded in a dungeon certain areas will draw with a different pattern up until the player visits town and returns to the dungeon, which probably isn't intended.

Source/scrollrt.cpp
qndel: no

Source/scrollrt.cpp
qndel: this code isn't supposed to run for dungeons

Source/scrollrt.cpp
```diff
@@ -916,6 +917,139 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
...
+
+/**
+ * @brief Render some semi random patterns beyond town borders instead of black void
+ */
+
+void RenderTownBorders(const Surface &out, Point tilePosition, Point targetBufferPosition, bool isDungeon)
+{
+	if (leveltype != DTYPE_TOWN)
+		return;
+	static bool firstRender = true;
```
glebm, line 1037: Let's do this without this static bool, e.g. by having a separate function for the stuff inside `if (firstRender)`?

qndel: this part of code is basically me giving up trying to do it the normal way and going with a hacky approach instead 😅

PR #3336, opened 2021-10-31
Use Point/Displacements when searching the area centered on a tile

ephphatha:
This addresses a todo comment, though not in the way that was originally intended as we no longer use Direction::Omni

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -161,16 +161,12 @@ void FindItemOrObject()
 	}
 
 	for (Point targetPosition : searchArea) {
-		int8_t objectId = abs(dObject[targetPosition.x][targetPosition.y]) - 1;
-		if (objectId < 0) {
+		Object *object = ObjectAtPosition(targetPosition);
+		if (object == nullptr || object->_oSelFlag == 0) {
+			// No object or non-interactive object
 			continue;
 		}
-
-		Object &object = Objects[objectId];
-		if (object._oSelFlag == 0) {
-			continue;
-		}
-		if (targetPosition == futurePosition && object._oDoorFlag) {
+		if (targetPosition == futurePosition && object->_oDoorFlag) {
 			continue; // Ignore doorway so we don't get stuck behind barrels
 		}
 
@@ -182,12 +178,12 @@ void FindItemOrObject()
 			// Don't check the tile we're leaving if the player is walking
 			continue;
 		}
-		if (object.IsDisabled()) {
+		if (object->IsDisabled()) {
 			continue;
 		}
 
 		rotations = newRotations;
-		pcursobj = objectId;
+		pcursobj = object->GetId();
 		cursPosition = targetPosition;
 	}
 }
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -13,6 +13,7 @@
 #include "cursor.h"
 #include "doom.h"
 #include "engine/point.hpp"
+#include "engine/points_in_rectangle_range.hpp"
 #include "gmenu.h"
 #include "help.h"
 #include "inv.h"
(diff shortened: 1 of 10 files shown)


Source/controls/plrctrls.cpp
```diff
@@ -122,61 +122,80 @@ int GetDistanceRanged(Point destination)
...
-/**
- * @todo The loops in this function are iterating through tiles offset from the player position. This
- * could be accomplished by looping over the values in the Direction enum and making use of
- * Point math instead of nested loops from [-1, 1].
- */
 void FindItemOrObject()
 {
-	int mx = Players[MyPlayerId].position.future.x;
-	int my = Players[MyPlayerId].position.future.y;
+	// This search uses row-major order, compared to CheckThemeObj3 using col-major :(
```
AJenbo, line 127: Feel free to change this one

ephphatha:
To elaborate on the last commit, in vanilla Diablo/Hellfire opening a trapped chest fires the missile toward the player closest to the left/bottom of the trapped object. In the following screenshot the warrior would be targeted. With the early break, the rogue gets targeted instead. I thought it was a minor change to avoid looping unnecessarily in multiplayer games but it can be ignored pretty easily.

image

The other behavioural change relates to moving with a controller which was added in devX anyway, this means that the search prefers the item above/to the right of the player when they move between two items or interactable objects.

Source/path.h
```diff
@@ -5,14 +5,26 @@
...
+/** Specifies a 3x3 search area (including the starting tile). */
+constexpr std::array<Displacement, 9> SearchOffsets3x3 {
+	{
+	    // clang-format off
+		    { -1, -1 }, { 0, -1 }, { 1, -1 },
+		    { -1,  0 }, { 0,  0 }, { 1,  0 },
+		    { -1,  1 }, { 0,  1 }, { 1,  1 },
+	    // clang-format on
+	}
+};
```
julealgon, line 27: Can't this be done using a 3x3 rectangle?

I understand using individual points for more "complex" shapes (like quest entry points) since we don't have a general purpose `Region` abstraction, but for these simpler areas, it feels to me there is no reason to go with individual displacements in place of a `Rectangle` instance that is applied at tile space.

ephphatha: Probably, I'll see what it looks like at some point over the weekend.

ephphatha:
@julealgon Took a fair bit of boilerplate but the last commit adds a wrapper to allow iterating over all points (tiles) covered by a rectangle. I added some constructors to make it easier to define a square centered on a point however the range class works with arbitrary rectangles.

Would definitely appreciate more eyes on the iterator since it's the first time I've attempted something which doesn't return an lvalue.

Source/engine/size.hpp
```diff
@@ -6,6 +6,20 @@ struct Size {
...
+	    : width(width)
+	    , height(height)
+	{
+	}
+
+	explicit constexpr Size(int size)
+	    : width(size)
+	    , height(size)
+	{
+	}
```
AJenbo, line 21: Do we finally have a solution the the rectangle vs square problem of programming :o

ephphatha:
Given I've introduced (now fixed) an off by one error I'll add some tests for this iterator after work. It'll help document the ordering requirements needed for use during dungeon generation.

Source/engine/points_in_rectangle_range.hpp
```diff
@@ -0,0 +1,195 @@
+#pragma once
+
+#include "point.hpp"
```
glebm, line 3:
```suggestion
#include <iterator>

#include "point.hpp"
```

Source/engine/points_in_rectangle_range.hpp
```diff
@@ -0,0 +1,195 @@
...
+		PointsInRectangleIterator(Rectangle region, int index = 0)
+		    : region(region)
+		    , index(index)
+		{
+		}
+
+		value_type operator*() const
+		{
+			// Default to row-major iteration e.g. {0, 0}, {1, 0}, {2, 0}, {0, 1}, {1, 1}, ...
+			return region.position + Displacement { index % region.size.width, index / region.size.width };
```
glebm, line 27: Not important for small rectangles, but if we start using this in larger rectangles, it'd be nice to avoid a modulo and an integer division on every iteration (can simply keep 2 indices instead of 1).

glebm (changes requested):

glebm (dismissed):
Approval for the C++ iterator stuff

Source/engine/points_in_rectangle_range.hpp
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+public:
+	using iterator_category = std::random_access_iterator_tag;
+	using difference_type = int;
+	using value_type = Point;
+	using pointer = const value_type *;
+	using reference = value_type &&;
+
+	virtual reference operator*() const = 0;
```
glebm, line 31: Let's avoid virtual methods, they're slower to call. Simply remove this line, remove `override` throughout.

ephphatha: I've split this into a separate branch for now while I try work out the syntax required to get it building on GCC. The only reason I had this virtual method was to make the base class abstract, but that's not necessary given the base class doesn't fully define any of the required concepts for an iterator.

ephphatha:
Finally got around to fixing this on GCC. MSVC appears to use inheritance for iterators with common logic but from what I can see GCC really does not like this approach. There are also footguns with certain operators which return a copy, they need to be redefined to ensure that derived class don't revert back to the base class when those operators are used. I ended up just splitting the definitions and redefining all operators so that there was no chance for one iterator/range to morph into the other.

There are additional uses I haven't included in this PR (`trm5x/y` in `TFit_Obj5`, loops over x/y like `DoUnlight`) since I didn't want to make it too annoying to review, if this is an acceptable approach I'll use it more often throughout the code.

Source/objects.cpp
```diff
@@ -4822,16 +4823,19 @@ void OperateTrap(Object &trap)
...
-		}
-	}
+
 	if (!deltaload) {
+		// default to firing at the trigger object
+		Point target = trigger.position;
+
+		PointsInRectangleRange searchArea { Rectangle { target, 1 } };
+		// look for a player near the trigger (using a reverse search to match vanilla behaviour)
+		auto foundPosition = std::find_if(searchArea.crbegin(), searchArea.crend(), [](Point testPosition) { return InDungeonBounds(testPosition) && dPlayer[testPosition.x][testPosition.y] != 0; });
```
AJenbo, line 4833: Would it make sens to have `InDungeonBounds()` be part of `PointsInRectangleRange`?

Source/controls/plrctrls.cpp
```diff
@@ -122,61 +123,71 @@ int GetDistanceRanged(Point destination)
...
-			rotations = newRotations;
-			pcursitem = i;
-			cursPosition = Point { mx, my } + Displacement { xx, yy };
+	auto searchArea = PointsInRectangleRangeColMajor { Rectangle { futurePosition, 1 } };
+
+	for (Point targetPosition : searchArea) {
+		// As the player can not stand on the edge of the map this is safe from OOB
+		int8_t itemId = dItem[targetPosition.x][targetPosition.y] - 1;
+		if (itemId < 0) {
+			// there shouldn't be any multi-tile objects, but just in case only considering positive indexes here
```
AJenbo, line 137: Isn't sarcophagus such an object?

Source/controls/plrctrls.cpp
```diff
@@ -122,61 +123,71 @@ int GetDistanceRanged(Point destination)
...
 	}
 
-	if (leveltype == DTYPE_TOWN || pcursitem != -1)
+	if (leveltype == DTYPE_TOWN || pcursitem != -1) {
 		return; // Don't look for objects in town
+	}
 
-	for (int xx = -1; xx < 2; xx++) {
-		for (int yy = -1; yy < 2; yy++) {
-			if (dObject[mx + xx][my + yy] == 0)
```
AJenbo, line 162: This feels like the change that makes this iterator pay off :+1:

Source/controls/plrctrls.cpp
ephphatha: This comment should say "...multi-tile (dropped) items...". No idea why I wrote object here.

Source/objects.cpp
ephphatha: I don't think that would be appropriate for this iterator, it's more concerned with providing a way to access all discrete Points that are contained within a Rectangle. If we wanted to restrict creating rectangles to only in bounds areas then that'd be best enforced in the Rectangle class (but I wouldn't want to do that either). For this case, I'd rather use a wrapper `IsPlayerAtPosition` similar to the `IsObjectAtPosition` helper that is proposed in #3324

Source/controls/plrctrls.cpp
ephphatha, line 137:
```suggestion
			// there shouldn't be any items that occupy multiple ground tiles, but just in case only considering positive indexes here

```

Source/controls/plrctrls.cpp
ephphatha: Haha, and the sad part is I want to define a replacement for this use case (and the first step of dropping an item) which looks at adjacent points relative to a given direction (basically the order in #3530 as an iterator).

AJenbo:
Hm going to need a rebase if you don't mind :)

PR #3337, opened 2021-10-31
Move objectIsDisabled to a member function

ephphatha:
While it does reference a global configuration option it's mainly concerned with the properties of the object itself, so makes sense (at least I think it does :D) to be a member of the object class.


qndel (approved):
👍

PR #3347, opened 2021-10-31
Update BG translation

NikoVP:
Update Bg translation


qndel:
Why do you open and close ...

PR #3348, opened 2021-10-31
Update Bg translation

NikoVP:


AJenbo:
You really should look in to resetting your branch before updating translation so that we don't end up with conflicts and broken code like has happened at this point.

I'm sure @qndel can help you. Note that resetting will remove any changes you have made, so it should be done before you start you next round of work, and after the previous one was merged.

NikoVP:
I'll try doing that later today if I get the time

PR #3349, opened 2021-10-31
Disable shift-clicking of items you don't have stats to wear

joewis:
Ensure people who shift-click items they don't have the stats for don't end up empty-handed.


Source/inv.cpp
```diff
@@ -829,7 +829,7 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 			if (automaticMove) {
 				if (CanBePlacedOnBelt(holdItem)) {
 					automaticallyMoved = AutoPlaceItemInBelt(player, holdItem, true);
-				} else {
+				} else if (CanEquip (holdItem)) {
```
AJenbo, line 832: Don't forget to run clang-format
```suggestion
				} else if (CanEquip(holdItem)) {
```

PR #3354, opened 2021-11-01
Replace StormLib with libmpq

### LINKED ISSUE #3223: Try switching to libmpq
As of #3180 and #3222, we've completely isolated MPQ reading logic. We already use our own code for writing mpqs.

There are a few issues with StormLib:

1. It is not thread-safe, requiring a lock around open, read, and close, which is not ideal for streaming music.
2. It is not very portable, requiring lots of platform-specific ifdefs internally in the past.
3. It mmaps the whole archive, requiring loads of virtual memory (was a problem on the SerenityOS 32-bit port).

https://github.com/ge0rg/libmpq does not seem to have any of these issues. Its source code is also somewhat more readable.

However, it is quite low-level: the only way to stream files out of it is to operate directly on blocks.

It would be worth investigating if we can write an `SDL_RWops` wrapper around the block API (the current StormLib wrapper is at https://github.com/diasurgical/devilutionX/blob/master/Source/storm/storm_sdl_rw.cpp).

Adding the dependency via CMake was fairly straightforward: https://github.com/glebm/devilutionX/commits/libmpq

glebm:
libmpq is a much simpler alternative to StormLib for reading MPQ archives.

We use our own fork of libmpq: https://github.com/diasurgical/libmpq

Impact:

* DevilutionX is now a lot more portable. Unlike StormLib, libmpq only  has platform-specific code for Windows.
* Locks around file access **removed** (instead we duplicate the file descriptor for streamed audio only).
* RAM usage is **300 KiB** lower than StormLib.
* Stripped release linux_x86_64 binary is **32 KiB** smaller.
* Amiga build now shows an error dialog with no text instead of crashing.  :shrug: 

Fixes #3223.


Source/game_assets.cpp
```diff
@@ -0,0 +1,76 @@
+#include "game_assets.hpp"
```
AJenbo, line 1: Should this live in `/engine`?

glebm: Done

Source/engine/game_assets.cpp
```diff
@@ -0,0 +1,76 @@
...
+	const auto at = [=](std::optional<MpqArchive> &src) -> bool {
+		if (src && src->GetFileNumber(fileHash, *fileNumber)) {
+			*archive = &(*src);
+			return true;
+		}
+		return false;
+	};
+
+	return at(font_mpq) || at(lang_mpq) || at(devilutionx_mpq)
+	    || (gbIsHellfire && (at(hfvoice_mpq) || at(hfmusic_mpq) || at(hfbarb_mpq) || at(hfbard_mpq) || at(hfmonk_mpq) || at(hellfire_mpq))) || at(spawn_mpq) || at(diabdat_mpq);
```
AJenbo, line 29: I think it would be best if we could instead have an ordered map rather then hardcode the order, and set the priority with `LoadMPQ("spawn.mpq", 1000)`, which appears to be how things where originally handled.

This is the original priority values:
diabdat.mpq/spawn.mpq 1000
hellfire.mpq 8000
hfmonk.mpq 8100
hfbard.mpq 8110
hfbarb.mpq 8120
hfmusic.mpq 8200
hfvoice.mpq 8500

(high number are checked fierst)

Suggested values for  our MPQ files
lang_mpq 10200
font.mpq 10100
devilutionx_mpq 10000

Also I think we should add a check for user.mpq to allow for custom content (previously people would usually name it patch_rt.mpq, or use the two unused Hellfire names hfopt1.mpq hfopt2.mpq)

I'm ok with this not being done as part of this PR, just thought it was a good time to bring it up :)

Source/engine/game_assets.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+} // namespace
+
+SDL_RWops *OpenGameAsset(const char *filename, bool threadsafe)
```
AJenbo, line 34: For some reason I prefer to not have `Game` in this name, it some how seams either redundant or too specific: `OpenAsset()`, what do you think?

AJenbo (commented):
Lots of nice improvements from making this switch, I just have two comments/questions in relation to the changes :)

Source/engine/game_assets.cpp
glebm: Renamed to OpenAsset

Source/engine/game_assets.cpp
glebm: That could be an option, although I'm not sure priorities are better than a simple list but is out-of-scope for this PR

AJenbo: The benefit of priorities is that it allows us to load some mpq's early (fonts), but prioritize them at a different order then they where loaded.

glebm: Sure but that could still just be a list. The drawback of priorities is that their effects are non-local.

AJenbo (approved):

PR #3357, opened 2021-11-01
Add helpers for checking object type

ephphatha:
Comments get a bit repetitive, I'm happy to cut them down to a single line.

Also made the IsAnyOf/IsNoneOf helpers constexpr here since they're simple functions, allows making all the helpers constexpr which should minimise the impact of introducing a wrapper.

### DIFF
--- a/Source/engine.h
+++ b/Source/engine.h
@@ -45,25 +45,25 @@
 namespace devilution {
 
 template <typename V, typename X>
-bool IsAnyOf(const V &v, X x)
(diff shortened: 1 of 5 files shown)


AJenbo (approved):
rebase

AJenbo (approved):

AJenbo:
P.s. I would use the term `check for` instead of `test for` since the latter constantly makes me think you are writing new unit tests :)

AJenbo (approved):

PR #3363, opened 2021-11-02
Optional functionality to auto-pickup potions

joewis:
Searching for and picking up potions disturbs immersion, especially when playing on mobile.
Adding options in diablo.ini to auto-pickup the specified number of (full) healing/mana/reju potions.


Source/qol/autopickup.cpp
```diff
@@ -57,4 +74,42 @@ void AutoGoldPickup(int pnum)
...
+                                NetSendCmdGItem(true, CMD_REQUESTAGITEM, pnum, pnum, itemIndex);
+                                item._iRequest = true;
+                        }
+                        if (sgOptions.Gameplay.nFullRejuPotionPickup > numMiscItemsInInv(IMISC_FULLREJUV) && item._iMiscId == IMISC_FULLREJUV) {
+                                NetSendCmdGItem(true, CMD_REQUESTAGITEM, pnum, pnum, itemIndex);
+                                item._iRequest = true;
+                        }
+			}
+                }
+         }
```
AJenbo: I can see why clang-format would be upset with you for this :)

AJenbo:
@joewis what IDE do you use, most have some way to integrate with clang-format at this point, or you can run it from a command prompt.

Source/qol/autopickup.cpp
```diff
@@ -30,6 +30,23 @@ bool HasRoomForGold()
...
+			numItems++;
+		}
+	}
+
+	for (int i = 0; i < MAXBELTITEMS; i++) {
+		if (!Players[MyPlayerId].SpdList[i].isEmpty() && Players[MyPlayerId].SpdList[i]._iMiscId == iMiscId) {
+			numItems++;
+		}
+	}
+	return numItems;
```
ephphatha, line 47: This can use the `InventoryAndBeltPlayerItemsRange` iterator glebm added in #2892
```suggestion
	int numItems = 0;
	for (const auto &item : InventoryAndBeltPlayerItemsRange { Players[MyPlayerId] }) {
		if (item._iMiscId == iMiscId) {
			numItems++;
		}
	}
	return numItems;
```

It handles empty spaces, lets you combine the loops too :)

Source/qol/autopickup.cpp
```diff
@@ -57,4 +74,41 @@ void AutoGoldPickup(int pnum)
...
 
+void AutoItemPickup(int pnum)
+{
+	for (auto pathDir : PathDirs) {
+		Point tile = Players[pnum].position.tile + pathDir;
+		if (dItem[tile.x][tile.y] != 0) {
+			int itemIndex = dItem[tile.x][tile.y] - 1;
+			auto &item = Items[itemIndex];
+			if (AutoPlaceItemInInventory(Players[pnum], item, false) || AutoPlaceItemInBelt(Players[pnum], item, false)) {
+				if (sgOptions.Gameplay.nHealPotionPickup > numMiscItemsInInv(IMISC_HEAL) && item._iMiscId == IMISC_HEAL) {
```
ephphatha, line 85: I don't think it'll make much of a difference in release builds but it's nice to have the simple condition first:
```suggestion
				if (item._iMiscId == IMISC_HEAL && sgOptions.Gameplay.nHealPotionPickup > numMiscItemsInInv(IMISC_HEAL)) {
```

lets the compiler short circuit and avoid running the loop in `numMiscItemsInInv` if it's not smart enough to recognise that optimisation.

Source/qol/autopickup.cpp
ephphatha: Also just noticed the AutoPickup loop is looking at items on the ground near Players[pnum], while this function is looking only at Players[MyPlayerId]. I think it could desync if a remote player walks near a tile containing a health potion and their client sees they're above the threshold, but the local player doesn't have any health potions in their inventory? Could make this a function on the Player struct so it looks at `this` instead of the global `MyPlayer`

Source/qol/autopickup.cpp
joewis: I trust my compiler in this case ;)

ephphatha: Turns out the built in local AND/OR operators enforce sequence ordering. We don't override operator&&(bool, bool) so this will always run the loop before checking the item is actually the correct type.

> Every value computation and side effect of the first (left) argument of the built-in logical AND operator && and the built-in logical OR operator || is sequenced before every value computation and side effect of the second (right) argument.
https://en.cppreference.com/w/cpp/language/eval_order

joewis: Seriously.... OK I'll swap it around.

julealgon:
Just wanted to comment on how I really dislike the special casing going on here.

IMHO, we should not specialize for potions at all, but generalize the auto pickup behavior currently used for gold, to allow customizing it for any item.

Otherwise, the code will keep getting more and more complex and it will be much harder to refactor out of it later.

joewis:
> Just wanted to comment on how I really dislike the special casing going on here.
> 
> IMHO, we should not specialize for potions at all, but generalize the auto pickup behavior currently used for gold, to allow customizing it for any item.
> 
> Otherwise, the code will keep getting more and more complex and it will be much harder to refactor out of it later.

Do you mean to suggest to merge this into 1 function? It would be rather straight forward, and also avoid running the loop once for gold and once more for other items.

julealgon:
> Do you mean to suggest to merge this into 1 function?

Not only that, but changing how even the toggle works, so we have a single place to define "items which will be auto picked up", you know?

Instead of having a special flag for gold, and another special flag for potions, and in the future other special flags for other things, etc.

The idea is to combine all the logic into a single "allow selection of items which will be picked up automatically" feature, and handle it all generally.

joewis:
That is intentionally not like that. If you keep picking up potions regardless how many you're already carrying, it will get annoying. 
I want to be able to define which potions I pick up, and how many.

vlolteanu:
> That is intentionally not like that. If you keep picking up potions regardless how many you're already carrying, it will get annoying. I want to be able to define which potions I pick up, and how many.

This is slightly offtopic, but the same applies for gold auto-pickup, albeit to a lesser extent. In hell difficulty, money piles up really quickly. If you hoard it on the ground in town, auto-pickup becomes counter-productive.

I'd argue disabling auto-pickup (for both gold and pots) is useful if:

- The stuff has been intentionally dropped. I don't see a way for this to persist when saving the game unless we modify the file format (and break backward compatibility), so this might be a bit unrealistic.
- The stuff is on the ground in town (implying it has been intentionally dropped)

julealgon:
> That is intentionally not like that. If you keep picking up potions regardless how many you're already carrying, it will get annoying.
I want to be able to define which potions I pick up, and how many.

@joewis seems to me it can still be generalized nicely, the only difference is that the mechanism needs to support an optional threshold check for the auto-pickup action: in case of gold, it is unbounded, in case of potions, you can check how many you already have, etc.

The "which potion" part is also fine: that's just "a different item". 

With a general auto-pickup system, you could configure:
- auto-pickup gold, always
- auto-pickup health potions, up to 5 + belt only
- auto-pickup mana potions, up to 2 + belt only
- auto-pickup healing scrolls, up to 3
- etc

I hope that makes sense.

joewis:
> - auto-pickup gold, always
> - auto-pickup health potions, up to 5 + belt only
> - auto-pickup mana potions, up to 2 + belt only
> - auto-pickup healing scrolls, up to 3
> - etc
> 

For gold might be nice to have a configurable limit too, e.g. don't pick up anymore when carrying >= 50k. "Always pick-up" can be achieved by setting that limit too 40x5000, allowing the whole inventory to be filled with gold.

For health, mana and reju potions the functionality is already as you described.

For scrolls - where to stop? Maybe someone also wants to pickup up to 10 Nova scrolls but is not interested in healing.

Picking up rings and amulets on mobile is also quite challenging, and may be a candidate for autopickup.

I think it would be nice to pickup and autoequip one-of-everything, e.g. after dying and walking over the dropped items. On PC with mouse that may be part of the challenge, but on mobile it's hopeless.

ephphatha:
Got around to playtesting this a bit and multiplayer games force all players to auto-pickup based on the highest limit among whoever's connected.

The warrior has a limit of 0 health potions set:

https://user-images.githubusercontent.com/357684/142745068-25c69ebd-138b-43cb-bee5-a5157c5de883.mp4

When I join with the sorcerer (having a limit of 5 set in that clients config file) the warrior gets forced to pick up potions beyond its limit. This ends up causing a desync as well:

https://user-images.githubusercontent.com/357684/142745112-bfaccc51-2e8b-4783-8641-c156fd3052cc.mp4

Unifying the functions should help as AutoGoldPickup has guards to ensure it only runs for the local player (and not in town).

joewis:
> Got around to playtesting this a bit and multiplayer games force all players to auto-pickup based on the highest limit among whoever's connected.
> 
> Unifying the functions should help as AutoGoldPickup has guards to ensure it only runs for the local player (and not in town).

This should be fixed now. I have also added an option to enable/disable auto pickup in town.

Source/qol/autopickup.cpp
```diff
@@ -30,31 +31,73 @@ bool HasRoomForGold()
...
 		Point tile = Players[pnum].position.tile + pathDir;
 		if (dItem[tile.x][tile.y] != 0) {
 			int itemIndex = dItem[tile.x][tile.y] - 1;
 			auto &item = Items[itemIndex];
-			if (item._itype == ItemType::Gold) {
+			if (hasRoomForGold && item._itype == ItemType::Gold) {
 				NetSendCmdGItem(true, CMD_REQUESTAGITEM, pnum, pnum, itemIndex);
 				item._iRequest = true;
 			}
+			if (AutoPlaceItemInInventory(Players[pnum], item, false) || AutoPlaceItemInBelt(Players[pnum], item, false)) {
```
ephphatha, line 65: It'd be worth having a check for item._itype == ItemType::Misc here. It's mainly for clarity but also avoids running the inventory size calcs for items that will never be picked up automatically. _iMiscId is occasionally populated for non-misc items, though this currently doesn't impact the functionality as potion/elixir subtypes aren't reused.
```suggestion
			if (item._itype == ItemType::Misc && (AutoPlaceItemInInventory(Players[pnum], item, false) || AutoPlaceItemInBelt(Players[pnum], item, false))) {
```

ephphatha (dismissed):
Behaves much nicer in multiplayer games now that the guard for pnum == myplayerid is used consistently. Defaulting to off is good, given how many autopickup options are now available it's potentially worth having a dedicated section in the config😂

joewis:
I've added the check for iType Misc - good to avoid extra checks when the outcome is certain.

Source/qol/autopickup.cpp
```diff
@@ -30,31 +31,73 @@ bool HasRoomForGold()
...
 		Point tile = Players[pnum].position.tile + pathDir;
 		if (dItem[tile.x][tile.y] != 0) {
 			int itemIndex = dItem[tile.x][tile.y] - 1;
 			auto &item = Items[itemIndex];
-			if (item._itype == ItemType::Gold) {
+			if (hasRoomForGold && item._itype == ItemType::Gold) {
 				NetSendCmdGItem(true, CMD_REQUESTAGITEM, pnum, pnum, itemIndex);
 				item._iRequest = true;
 			}
+			if (item._itype == ItemType::Misc && AutoPlaceItemInInventory(Players[pnum], item, false) || AutoPlaceItemInBelt(Players[pnum], item, false)) {
```
ephphatha, line 65: logical `&&` has higher precedence than `||`, as written this will evaluate to `(type == misc && placeInInventory) || placeInBelt`

```suggestion
			if (item._itype == ItemType::Misc && (AutoPlaceItemInInventory(Players[pnum], item, false) || AutoPlaceItemInBelt(Players[pnum], item, false))) {
```

joewis: ohh... right.

Source/qol/autopickup.cpp
```diff
@@ -30,31 +31,73 @@ bool HasRoomForGold()
...
 
+	bool hasRoomForGold = HasRoomForGold();
+
 	for (auto pathDir : PathDirs) {
 		Point tile = Players[pnum].position.tile + pathDir;
 		if (dItem[tile.x][tile.y] != 0) {
 			int itemIndex = dItem[tile.x][tile.y] - 1;
 			auto &item = Items[itemIndex];
-			if (item._itype == ItemType::Gold) {
+			if (hasRoomForGold && item._itype == ItemType::Gold) {
```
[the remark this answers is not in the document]
joewis: @ephphatha Thanks for the review! I've gone through the code again and included your suggestions.

ephphatha (dismissed):
Looks good :).

Defaulting to off like the other optional features makes it easy enough to ignore if the player doesn't want to use it. I think the choices presented are a sufficient to make it useful without overloading the player (or making the implementation overly complicated).

Source/qol/autopickup.cpp
```diff
@@ -30,31 +31,73 @@ bool HasRoomForGold()
...
 
+int NumMiscItemsInInv(int iMiscId)
+{
+	int numItems = 0;
+	for (const auto &item : InventoryAndBeltPlayerItemsRange(Players[MyPlayerId])) {
+		if (item._iMiscId == iMiscId) {
+			numItems++;
+		}
+	}
+	return numItems;
```
glebm, line 42: Add `#include <algorithm>` and:

```suggestion
        InventoryAndBeltPlayerItemsRange items{Players[MyPlayerId]};
	return std::count_if(items.begin(), items.end(), [](const Item &item) { return item._iMiscId == iMiscId; });
```

Source/qol/autopickup.cpp
```diff
@@ -30,31 +31,73 @@ bool HasRoomForGold()
...
+					break;
+				case IMISC_ELIXSTR:
+				case IMISC_ELIXMAG:
+				case IMISC_ELIXDEX:
+				case IMISC_ELIXVIT:
+					doPickup = *sgOptions.Gameplay.autoElixirPickup;
+					break;
+				default:
+					break;
+				}
```
glebm, line 94: This can be a bit more concise if extracted into a function (so you can do `return` instead of assignment + `break`)

Source/qol/autopickup.cpp
joewis: OK that's neat, I wasn't aware of function templates.

Source/qol/autopickup.cpp
joewis: OK, I've done that.

glebm (approved):

PR #3370, opened 2021-11-03
Fix speedbook navigation on gamepad

StephenCWills:
This fixes gamepad navigation after opening the speedbook with anything other than `GameActionType_TOGGLE_QUICK_SPELL_MENU`.

---

The gamepad code that toggles the speedbook called a function named `StoreSpellCoords()` in order to compute the location of each spell in the speedbook. However, if the speedbook was opened via any other method, such as the `S` key or by clicking on the spell icon in the main panel, the speedbook coordinates would not be updated and gamepad navigation would not work properly. This was most obvious when using touch controls which doesn't make use of `GameActionType_TOGGLE_QUICK_SPELL_MENU` to open the speedbook.

This PR rewrites the `HotSpellMove()` function to use `GetSpellListItems()` instead, which is the function that provides information for rendering the speedbook.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -618,65 +618,6 @@ void RemoveGold(Player &player, int goldIndex)
 	dropGoldValue = 0;
 }
 
-struct SpellListItem {
-	Point location;
-	spell_type type;
-	spell_id id;
-	bool isSelected;
-};
-
-std::vector<SpellListItem> GetSpellListItems()
-{
-	std::vector<SpellListItem> spellListItems;
-
-	uint64_t mask;
-
-	int x = PANEL_X + 12 + SPLICONLENGTH * SPLROWICONLS;
-	int y = PANEL_Y - 17;
-
-	for (int i = RSPLTYPE_SKILL; i < RSPLTYPE_INVALID; i++) {
-		auto &myPlayer = Players[MyPlayerId];
-		switch ((spell_type)i) {
-		case RSPLTYPE_SKILL:
-			mask = myPlayer._pAblSpells;
-			break;
-		case RSPLTYPE_SPELL:
-			mask = myPlayer._pMemSpells;
-			break;
-		case RSPLTYPE_SCROLL:
-			mask = myPlayer._pScrlSpells;
-			break;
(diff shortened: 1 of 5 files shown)


Source/controls/plrctrls.cpp
```diff
@@ -1014,64 +1012,62 @@ void InvMove(AxisDirection dir)
...
-		if (spbslot < speedspellcount - 1) {
-			newMousePosition = speedspellscoords[spbslot + 1];
-		}
-	} else if (dir.x == AxisDirectionX_RIGHT) {
-		if (spbslot > 0) {
-			newMousePosition = speedspellscoords[spbslot - 1];
-		}
-	}
+		for (size_t i = 0; i < spellListItems.size(); i++) {
+			auto index = i;
```
glebm, line 1040:
```suggestion
			const size_t index = i;
```

Source/controls/plrctrls.cpp
```diff
@@ -1014,64 +1012,62 @@ void InvMove(AxisDirection dir)
...
+		Point center = spellListItem.location + Displacement { SPLICONLENGTH / 2, -SPLICONLENGTH / 2 };
+		int distance = MousePosition.ManhattanDistance(center);
+		if (distance < shortestDistance) {
+			position = center;
+			shortestDistance = distance;
 		}
 	}
 
-	Point newMousePosition = speedspellscoords[spbslot];
+	auto search = [&](AxisDirection dir, bool searchForward) {
```
glebm, line 1035:
```suggestion
	const auto search = [&](AxisDirection dir, bool searchForward) {
```

Source/control.cpp
```diff
@@ -817,6 +758,58 @@ void DrawSpellList(const Surface &out)
 	}
 }
 
+std::vector<SpellListItem> GetSpellListItems()
+{
+	std::vector<SpellListItem> spellListItems;
```
glebm, line 763: I know you've simply moved an existing function but it'd be good to clean this up to avoid allocating a vector on every frame (in a separate PR).

glebm (approved):

PR #3381, opened 2021-11-03
Update Russian translation

Elidevin:
correct "game speed" translation


SourceCodeDeleted:
I think this is good. Скорость игры makes more sense .

I am ok with it. But let @AJenbo decide because I have no authority here.

AJenbo:
Merged since there is another update in the works, and  one approval

PR #3392, opened 2021-11-05
Enable SDL1 build option for Windows

StephenCWills:
* Move SDL dependencies to features in vcpkg.json so the appropriate feature can be selected based on CMake options
* Add Windows backports for `SDL_GetBasePath()` and `SDL_GetPrefPath()`
* Replace `SDL_Rect` initializers with calls to `MakeSdlRect()`
* Disable winmm audio driver selection for SDL1


Source/utils/sdl2_to_1_2_backports.cpp
```diff
@@ -477,6 +485,218 @@ int SDL_BlitScaled(SDL_Surface *src, SDL_Rect *srcrect,
...
+	path[i + 1] = '\0'; /* chop off filename. */
+
+	retval = WIN_StringToUTF8(path);
+	SDL_free(path);
+
+	return retval;
+}
+
+char *
+SDL_GetPrefPath(const char *org, const char *app)
```
AJenbo, line 610:
```suggestion
char *SDL_GetPrefPath(const char *org, const char *app)
```

AJenbo:
Are we finally getting our Windows 95 port :D

Issue #3400, opened 2021-11-06
Shareware Hellfire monk crashes game


JohnsterID:
**Important information**
Windows 7 / DevilutionX 1.3.0 release

**Describe**
Loading monk on spawn hellfire crashes the game.
Test on 1.3.0 release after reading https://github.com/diasurgical/devilutionX/issues/3217

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Single Player
2. Select Monk character
3. Play

**Expected behaviour**
Loading warrior on spawn hellfire works.

[procdump](https://github.com/diasurgical/devilutionX/files/7489927/devilutionx.exe_211106_155329.zip)

StephenCWills:
I am unable to reproduce on Windows 10.

* Did you use devilutionx-windows-i386.zip or devilutionx-windows-x86_64.zip?
* Is this a crash-to-desktop or was there an error message?
* If there was an error message, can you share it?

---

> Testing after #3217

What does this mean? You did test the release version from https://github.com/diasurgical/devilutionX/releases/tag/1.3.0, correct?

JohnsterID:
devilutionx-windows-x86_64.zip for release 1.3.0 used.

Crash to desktop with no error message so I attempted to use procdump.

I edited the issue to clarify why I was mentioning that previous issue. The concept sounded interesting.

Tested devilutionx-windows-i386.zip for 1.3.0 crash to desktop as well.

SYMBOL_NAME:  sdl2!SDL_DYNAPI_entry+71848
MODULE_NAME: SDL2
IMAGE_NAME:  SDL2.dll
FAILURE_BUCKET_ID:  BREAKPOINT_80000003_SDL2.dll!SDL_DYNAPI_entry

qndel:
Can you list all mpq files your devilutionx has access to?

JohnsterID:
spawn plus hellfire
devilutionx.mpq SHA1: 97729C97F14D81EA5897A3DD398C251F76032133
hellfire.mpq SHA1: 4852A68A70F7F55A28084281A96E95C2027C5137
hfmonk.mpq SHA1: 47F44845803B05FA243C4733995B08CBAA0DBE37
hfmusic.mpq SHA1: 5D64D26798738C503B0AA1651C1534FF5488291C
hfvoice.mpq SHA1: 04E7FE3A5509013CBF22D0937817E2775226993E
[...]

PR #3401, opened 2021-11-06
fix hellfire shareware monk

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/3400


glebm:
Is it correct to set it to blank, or should we try to use the sorcerer sound if diabdat.mpq is present?

qndel:
> Is it correct to set it to blank, or should we try to use the sorcerer sound if diabdat.mpq is present?

No clue, in original hellfire it was blank

glebm (approved):

qndel:
Also looks like I was the one that changed it
https://github.com/diasurgical/devilutionX/commit/ec74917cb4b02a5e487a1156455dcb6c8afb21fc
And I don't remember any particular reason for it :D 
+ it doesn't seem like we are using that sound anywhere, so I'd blank it 🤷

PR #3406, opened 2021-11-06
remove ttf references and dependencies

frrsth:
Since we no longer use SDL-TTF. I have removed all the references and dependencies to ttf


AJenbo (approved):
Thanks for updating the docs :)

PR #3414, opened 2021-11-07
DrawString: Stop copying the input string

glebm:
Switch to a state-machine UTF-8 decoder from the branchless one.
This allows us to avoid copying the string on every `DrawString` call.


AJenbo:
I did think that branchless might not be so amazing if it requires padding, besides error handling being done on each call on the outside :)

AJenbo (approved):

PR #3418, opened 2021-11-07
debug fps command

qndel:


glebm (approved):

AJenbo:
I wonder what the 4th option for flipping this bool will be :thinking:

PR #3426, opened 2021-11-07
Use object reference (and self-id lookup) for BreakBarrel

ephphatha:
Theoretically we can set the ID as an attribute when the object is spawned but I haven't checked how the Objects array is used yet to confirm it's not getting memset at some point after allocation. The GetId method just uses the index (from the dObject cell value) for now.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -4125,81 +4125,86 @@ void BreakCrux(Object &crux)
 	ObjChangeMap(crux._oVar1, crux._oVar2, crux._oVar3, crux._oVar4);
 }
 
-void BreakBarrel(int pnum, int i, int dam, bool forcebreak, bool sendmsg)
+void BreakBarrel(int pnum, Object &barrel, int dam, bool forcebreak, bool sendmsg)
 {
-	if (Objects[i]._oSelFlag == 0)
+	if (barrel._oSelFlag == 0)
 		return;
 	if (forcebreak) {
-		Objects[i]._oVar1 = 0;
+		barrel._oVar1 = 0;
 	} else {
-		Objects[i]._oVar1 -= dam;
-		if (pnum != MyPlayerId && Objects[i]._oVar1 <= 0)
-			Objects[i]._oVar1 = 1;
+		barrel._oVar1 -= dam;
+		if (pnum != MyPlayerId && barrel._oVar1 <= 0)
+			barrel._oVar1 = 1;
 	}
-	if (Objects[i]._oVar1 > 0) {
+	if (barrel._oVar1 > 0) {
 		if (deltaload)
 			return;
 
-		PlaySfxLoc(IS_IBOW, Objects[i].position);
+		PlaySfxLoc(IS_IBOW, barrel.position);
 		return;
 	}
 
-	Objects[i]._oVar1 = 0;
-	Objects[i]._oAnimFlag = 1;
-	Objects[i]._oAnimFrame = 1;
-	Objects[i]._oAnimDelay = 1;
-	Objects[i]._oSolidFlag = false;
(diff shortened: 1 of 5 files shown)


Source/player.cpp
```diff
@@ -1143,16 +1143,15 @@ bool PlrHitPlr(int pnum, int8_t p)
...
-	} else {
-		oi = -dObject[mx][my] - 1;
+	int oi = abs(dObject[mx][my]) - 1;
+	if (oi < 0) {
+		return false;
 	}
 
-	if (Objects[oi]._oBreak == 1) {
-		BreakObject(pnum, oi);
+	Object targetObject = Objects[oi];
```
ephphatha, line 1151:
```suggestion
	Object &targetObject = Objects[oi];
```

Source/objects.cpp
```diff
@@ -4125,81 +4125,86 @@ void BreakCrux(Object &crux)
...
 					PlayerMHit(dPlayer[xp][yp] - 1, nullptr, 0, 8, 16, MIS_FIREBOLT, false, 0, &unused);
-				if (dObject[xp][yp] > 0) {
-					int oi = dObject[xp][yp] - 1;
-					if (Objects[oi]._otype == OBJ_BARRELEX && Objects[oi]._oBreak != -1)
-						BreakBarrel(pnum, oi, dam, true, sendmsg);
+				}
+				int oi = dObject[xp][yp] - 1;
+				if (oi >= 0) {
+					Object &adjacentObject = Objects[oi];
+					if (adjacentObject._otype == _object_id::OBJ_BARRELEX && !adjacentObject.IsBroken()) {
```
AJenbo, line 4182: Unless _object_id was turned in to a scoped enum I prefer that it uses the old style, if it was then it needs to be renamed :)

Source/objects.cpp
```diff
@@ -4325,6 +4330,11 @@ void SyncL3Doors(Object &door)
 
 } // namespace
 
+unsigned int Object::GetId() const
+{
+	return abs(dObject[position.x][position.y]) - 1;
```
AJenbo, line 4335: Uh, interesting solution. I think it works here for now, but probably will need a change in the long run.

Source/objects.cpp
ephphatha: It hasn't been scoped (yet 😉 ), I've prefixed the enum name purely to make that refactoring easier in the future.

Source/objects.cpp
ephphatha: Definitely. This is ... something that works for now.

I'd much rather write a property when the object is created as this allows creating objects from the level deltas based on what the host has generated, instead of having to have the object generation perfectly synced. But that requires going through the way objects are created to ensure we don't blow away any attributes before they get fully initialised.

Source/player.cpp
```diff
@@ -1143,16 +1143,15 @@ bool PlrHitPlr(int pnum, int8_t p)
...
-	} else {
-		oi = -dObject[mx][my] - 1;
+	int oi = abs(dObject[mx][my]) - 1;
+	if (oi < 0) {
+		return false;
 	}
 
-	if (Objects[oi]._oBreak == 1) {
-		BreakObject(pnum, oi);
+	Object &targetObject = Objects[oi];
```
AJenbo, line 1151: This only get called if the position is known to have a positive id. Might as well have it take `Object &targetObject` and make the call `PlrHitObj(pnum, Objects[dObject[dx][dy] - 1])`

Source/objects.cpp
AJenbo: Will take your word for it :)

AJenbo (approved):

PR #3428, opened 2021-11-07
Update libzt

glebm:
Includes https://github.com/diasurgical/libzt/commit/b2be9882771116fcfd4ad918f36de8587324d9e7, which is a fix for repo maintainers.

### DIFF
--- a/3rdParty/libzt/CMakeLists.txt
+++ b/3rdParty/libzt/CMakeLists.txt
@@ -3,7 +3,7 @@ include(FetchContent_MakeAvailableExcludeFromAll)
 include(FetchContent)
 FetchContent_Declare(libzt
   GIT_REPOSITORY https://github.com/diasurgical/libzt.git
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Should we add it to the 1.3 branch?

glebm:
@AJenbo Yeah that's a good idea

AJenbo:
Done, as well as some of the other changes you did that I would consider relevant for a potential 1.3.1. My current thinking is that 1.4.0 will have the following features:
- Touch support on all SDL2 platforms
- Fix voice pack support on 3DS/Switch
- Settings menus
- UI dynamically adapts to current input device
- Improved Gamepad UI
- Configurable Gampad layout

If it doesn't materialize before end of December 1.3.1 is pushed out at that point, and then 1.4.0 lands whenever.

1.3.1 would be:
- latest translations
- all platforms build from same commit
- bug-fixes from master that are trivial to pluck

PR #3435, opened 2021-11-07
Clean up id normalization

AJenbo:
Been wanting to do this for a while now :)


ephphatha (approved):
Makes it easier to pick out uses of dObject/dPlayer/dMonster as well for helpers like #3324 :)

PR #3437, opened 2021-11-07
Ingame settings menu - first version

obligaron:
Contributes to #1561

https://user-images.githubusercontent.com/25415264/141010145-198bec57-e73b-4124-b151-7779ad8703d8.mp4

<details><summary>Obsolete 1. Version</summary>

https://user-images.githubusercontent.com/25415264/140665713-efad7363-693c-4a15-844d-65755363565b.mp4

</details>

</details>

## Content

- Add `OptionEntryBase` as an abstraction for an option/setting
  - Needed for UI (name, description)
  - Encapsulates reading and writing from `ini`
  - Custom actions (see `SetValueChangedCallback`) are possible when changing options (for example we need to modify the window if grap input is changed)
  - Flags are supported to describe if a option can be changed and/or is visible (for example `Theo Quest` is only available in Hellfire)
  - `OptionEntryListBase` allows generic Lists (for example we could show a list of translated languages based on `mo`/`gmo ` files)
- Add `OptionCategoryBase` as an abstraction for categories. Is used to browse the categories and for `ini` reading/writing
- Improves diabloui
  - ViewPort Logic is now part of `diabloui.cpp` and not of the user
  - `UiListItem` can have individual flags (for example a different color)
  - `UiListItem` can be disabled
  - `UiListItem` can be paired together (`NeedsNextElement` is used for the settings value)
- ~~Small bugfix in `DrawString` for center or right aligned values and auto line breaks~~
- Generic settings menu that works with `OptionCategory` and `OptionEntry` (see example video)

## Notes

- This pr is big enough. So some fine-tuning and missing features should be included in follow-up pr's
  - UX isn't optimal atm (for example a background would be nice or some spacing's can be adjusted)
  - Text and description could be improved
  - Many options are currently not implemented
- Key mapper is currently completely missing
- With the abstraction `OptionCategory` and `OptionEntry` it should be (relatively) easy to implement a settings menu that works while be in a single- or multiplayer game (not part of this pr)

I hope this will be the first step for a settings menu, so everyone can discover all the nice little extras that devilutionX contains. 🙂


AJenbo:
Should we make it so Theo/Cow Quest only show when in Hellfire mode?

obligaron:
> Should we make it so Theo/Cow Quest only show when in Hellfire mode?

This is currently the case, cause both are specified with `OptionEntryFlags::OnlyHellfire` 🙂

FitzRoyX:
This is too much menu nesting and it is confusing that settings would be nested behind "extras" in the first place. That word just isn't associated with configuration and I know of no game menu that does this.

I would:
-rename "extras" to "settings"
-replace the "replay / switch to" actions to booleans "hellfire: on/off" and "play intro: on/off"
-proceed to list out all the other settings and just have (no sub-sub-menus).

AJenbo:
@obligaron if you can come up with a clear separator for sections then what you showed me, then I think @FitzRoyX's suggestion will work well. Except that we need to move Support back in to the main menu. Dropping Replay Intro is unfortunate, but acceptable.

julealgon:
As always, I'd strongly suggest splitting this PR into smaller changes. Each bullet in the description sounds like a valid standalone change to me, especially the "improvements" and "bug fixes" there.

As for UI-specific feedback, I believe it is considered a bad practice to have elements and their values distributed vertically like this without any spacing between. It is the classic checkbox/radiobutton issue, where it becomes hard to tell if the setting value applies to the top or bottom element, as well as making it overall less readable. Based on that, I'd suggest we try to find a way to keep the setting value in the same line as the setting itself, or at least introduce visible spacing between each setting (obviously easier, but I think the former is preferred).

The description being in the bottom is weird, but interesting. I'm personally fine with it even though I think it is less than ideal. The nicer version would be to show it alongside the setting itself, only when it is highlighted. I do understand that is harder to do, however.

Another question I'd raise is if this is intended to be accessed only from the main menu, and not from the in-game menu. It would be nice if it was accessible from both, but right now it is a bit unintuitive that some settings wouldn't persist on a per-character basis. So in that sense, it is better for it to only show in the main menu right now.

Having said that, if I were to chose the preferred order of execution here, I'd first introduce support for per-character settings persistence (so that each character could have their own configurations for some of these), and only then introduce this menu and the in-game menu, at the same time ideally.

Source/options.h
```diff
@@ -14,18 +16,180 @@ enum class StartUpGameOption {
...
+class OptionEntryEnum : public OptionEntryEnumBase {
+public:
+	OptionEntryEnum(string_view key, OptionEntryFlags flags, string_view name, string_view description, T defaultValue, std::initializer_list<std::pair<T, string_view>> entries)
+	    : OptionEntryEnumBase(key, flags, name, description, static_cast<int>(defaultValue))
+	{
+		for (auto entry : entries) {
+			AddEntry(static_cast<int>(entry.first), entry.second);
+		}
+	}
+	T GetValue()
```
glebm, line 154: For wrapper types it is idiomatic in C++ to override `operator*`, i.e.:

```suggestion
	T operator*()
```

Also, shouldn't this method be `const`?

julealgon: Would using something like this make sense for DevilutionX?
- https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nodiscard.html

glebm: Yeah, and we have it enabled already I think

glebm: It doesn't trigger because the function is not marked `const`

obligaron:
For me this is pr is the first step for the settings menu.
I wanted to create an abstraction that can be the basis for a settings menu while in menu or in a single/multiplayer game.
And this pr introduces this abstraction and I would welcome feedback for it.

But the problem with an abstraction that no one uses is that it's hard to argue if it's good or bad. That's why I created a first version of the settings menu but tried to separate every change in a separate clean commit.
It's by not perfect and while developing the first version I discovered multiple limitations with the current diablo menu ui (for example I tried to put the name and value (on/off) in one line, that's currently not supported).

My hope with this pr is that if we have the ground work running, we can advance independent and step for step with settings.
We can add future pr's with new settings. Meanwhile someone else can tweak the settings menu. Another person could implement a settings menu that runs while in a game. The next person could tackle the keymapper (currently not part of `Options` class).
So this pr is the start and not the end for ingame settings. 😉 

Anyway I tried to address the feedback and updated the layout (see new example video in initial post):
- Settings menu is now in the root menu
- No sub-menu for settings categories
- For now I left the "Switch to XYZ"-options untouched but moved them to the settings menu. (for example we don't have a shareware option in the ini and I didn't want to introduced it as the pr is big enough already)
- Intro changed to on/off option (default on) and will be on until it's switch in the settings menu
- Removed replay intro option

I hope you like the changes. 🙂

Source/options.h
obligaron: Missed this method. On the other GetXYZ-Methods I used `const` and `[[nodiscard]]`.
Fixed 🙂

julealgon:
> Anyway I tried to address the feedback and updated the layout (see new example video in initial post):

New version seems massively improved already. Pretty cool.

FitzRoyX:
I would...
1. Unify the intro setting. Disabling it, and then disabling it again once hellfire is activated doesn't make a lot of sense to me. I don't know why the ini splits this out either or makes hf its own category.
2. Rename "Grab Input" to "Confine Cursor"
3. Remove the hellfire class/quest activations. I'd rather condense the settings and make room for things people are actually divided about.
4. Does the Support entry need to be there? Is a readme and google not enough?

NikoVP:
Here is a quick mockup on how you can reuse assets to achieve look consistent with the rest of the menu. I used selgame.pcx, probably a new horizontal directional buttons might be needed. Was too lazy to change the white text to in-game one but you get the gist.
image

FitzRoyX:
I agree that "go back" shouldn't need scrolled to, but a left panel would restrict the width of the option names. I also think the description text just increases translator workload for little benefit. We should consider what julealgon said about making on/off the same line to double the number of options visible.
141615369-2c948fc3-6700-425a-9639-254506238a13dd

AJenbo:
> We should consider what julealgon said about making on/off the same line to double the number of options visible.

Thats a current limitation of the menu system, that we intent to resolve in a follow up. So as to not over complicate the change set.

Source/DiabloUI/settingsmenu.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+		optionDescription[0] = '\0';
+
+		const char *titleText = _("Settings");
+		vecDialog.push_back(std::make_unique<UiArtText>(titleText, MakeSdlRect(PANEL_LEFT, UI_OFFSET_Y + 161, PANEL_WIDTH, 35), UiFlags::FontSize30 | UiFlags::ColorUiSilver | UiFlags::AlignCenter, 8));
+		vecDialog.push_back(std::make_unique<UiScrollbar>(&ArtScrollBarBackground, &ArtScrollBarThumb, &ArtScrollBarArrow, MakeSdlRect(rectList.position.x + rectList.size.width + 5, rectList.position.y, 25, rectList.size.height)));
+		vecDialog.push_back(std::make_unique<UiArtText>(optionDescription, MakeSdlRect(rectDescription), UiFlags::FontSize12 | UiFlags::ColorUiSilverDark | UiFlags::AlignCenter, 1));
+
+		if (diabdat_mpq && hellfire_mpq)
+			vecDialogItems.push_back(std::make_unique<UiListItem>(gbIsHellfire ? _("Switch to Diablo") : _("Switch to Hellfire"), static_cast<int>(SpecialMenuEntry::SwitchGame), UiFlags::ColorUiGold));
```
AJenbo, line 117: Probably change this to either "Game Mode: Hellfire", "Hellfire Expansion: On" "Hellfire: Enabled" or something along that line.

AJenbo (approved):
@glebm would you give this a quick look just to see if there is something that should be changed before or after (since it's the first in a series) merging?

glebm:
This looks good to me overall. I forgot how DiabloUI scrolling bits work in detail but I assume @obligaron has tested it where it's used and will fix regressions if they're found.

The C++ bits look good. :heavy_check_mark: 

:rocket:

glebm (approved):

Issue #3162, opened 2021-10-18
Better CJK support


AJenbo:
This is for tracking issues noticed with displaying the Chinese translation, some of these may affect other languages as well.

- [ ] https://github.com/diasurgical/devilutionX/issues/3548
- [ ] https://github.com/diasurgical/devilutionX/issues/3549
- [ ] https://github.com/diasurgical/devilutionX/issues/3538
- [x] All symbols in the small fonts have been left aligned, but CJK symbols should be centered as they are not variable width
- [x] Item prefix blue string is not exposed to translators, change it to fmt to allow translators to reorder it as well (`{:s} {:s}` -> `{1}{0}` = "Golden Plate" -> "PlateGolden")
- [x] Line breaks can still be limited for languages that rarely have separator symbols (` `, `.`, `?`, `!`, `　`, `，`, `、`, `。`, `？`, and `！`). https://github.com/diasurgical/devilutionX/pull/3438 https://github.com/diasurgical/devilutionX/pull/3528 https://github.com/diasurgical/devilutionX/pull/3541
- [x] Store / talk text overlap. https://github.com/diasurgical/devilutionX/pull/3461
- [x] https://github.com/diasurgical/devilutionX/issues/3547 https://github.com/diasurgical/devilutionX/pull/3584
- [x] https://github.com/diasurgical/devilutionX/issues/3603
- [x] https://github.com/diasurgical/devilutionX/issues/3621
- [x] Re render font in high quality
- [x] Add texture to small font

![image](https://user-images.githubusercontent.com/204594/137707438-ef6d2166-d27e-4560-8901-30f05354f338.png)

![image](https://user-images.githubusercontent.com/204594/137707417-c3929868-9d65-4767-bfc0-0d00f9bf4d0d.png)

AJenbo:
Please see https://github.com/diasurgical/devilutionX/pull/3171 for the chosen solution to long lines not being split automatically.

tytannial:
Quest review progress

Diablo Quest
- [ ] Poisoned Water Supply
- [x] Skeleton King
- [x] The Butcher
- [ ] Ogden's Sign
- [ ] Gharbad the Weak
- [ ] Magic Rock
- [x] Arkaine's Valor
- [ ] Halls of the Blind
- [ ] Zhar the Mad
- [ ] Black Mushroom
- [ ] Anvil of Fury
- [x] Warlord of Blood
- [x] Lachdanan
- [x] Chamber of Bone
[...]

PR #3438, opened 2021-11-08
Automatically insert ZWSP between words in Chinese translation files

glebm:
Adds a tool can insert ZWSP between words in Chinese .po files using this model: https://tfhub.dev/google/zh_segmentation/1

We will use this tool to be able to word wrap Chinese translations at runtime without bundling a segmenter.

Doing this offline also allows us to use a segmenter that would otherwise be too slow for runtime.

The tool takes ~15s to run and depends on Python, TensorFlow, and downloads a model file as needed.
It is too heavy to run automatically at build time, so it will only be run manually for now:

    tools/zh_segmenter/segment.py --input_path Translations/zh_CN.po
    tools/zh_segmenter/segment.py --input_path Translations/zh_TW.po

image

Refs #3162
/cc @tytannial

### DIFF
--- a/Source/engine/render/text_render.cpp
+++ b/Source/engine/render/text_render.cpp
@@ -327,6 +327,7 @@ std::string WordWrapString(string_view text, size_t width, GameFontTables size, 
 		processedEnd = remaining.data();
 		lastBreakablePos = -1;
 		lineWidth = 0;
+		nextCodepoint = !remaining.empty() ? DecodeFirstUtf8CodePoint(remaining, &nextCodepointLen) : U'\0';
 	} while (!remaining.empty() && remaining[0] != '\0');
 	output.append(processedEnd, remaining.data());
 	return output;
--- a/Source/engine/render/text_render.cpp
+++ b/Source/engine/render/text_render.cpp
@@ -288,13 +288,15 @@ std::string WordWrapString(string_view text, size_t width, GameFontTables size, 
 			continue;
 		}
 
-		uint8_t frame = codepoint & 0xFF;
-		uint32_t unicodeRow = codepoint >> 8;
-		if (unicodeRow != currentUnicodeRow || kerning == nullptr) {
-			kerning = LoadFontKerning(size, unicodeRow);
-			currentUnicodeRow = unicodeRow;
+		if (codepoint != ZWSP) {
+			uint8_t frame = codepoint & 0xFF;
+			uint32_t unicodeRow = codepoint >> 8;
+			if (unicodeRow != currentUnicodeRow || kerning == nullptr) {
+				kerning = LoadFontKerning(size, unicodeRow);
+				currentUnicodeRow = unicodeRow;
+			}
+			lineWidth += (*kerning)[frame] + spacing;
 		}
-		lineWidth += (*kerning)[frame] + spacing;
 
 		const bool isWhitespace = IsWhitespace(codepoint);
 		if (isWhitespace || IsBreakAllowed(codepoint, nextCodepoint)) {
--- a/Source/engine/render/text_render.cpp
+++ b/Source/engine/render/text_render.cpp
@@ -398,8 +398,8 @@ uint32_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
 
 			if (HasAnyOf(flags, (UiFlags::AlignCenter | UiFlags::AlignRight))) {
 				lineWidth = (*kerning)[frame];
-				if (text[0] != '\0')
-					lineWidth += spacing + GetLineWidth(text, size, spacing);
+				if (!remaining.empty())
+					lineWidth += spacing + GetLineWidth(remaining, size, spacing);
 			}
 
 			if (HasAnyOf(flags, UiFlags::AlignCenter))
--- a/tools/zh_segmenter/segment.py
+++ b/tools/zh_segmenter/segment.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python
+import argparse
+import pathlib
+import os
+import re
+import sys
+
+import tensorflow_text
+
+_DISALLOW_SEGMENTATION = re.compile(
+	r'(?:\{[^\}]*\}|\\[a-z]|[\w/-](?::[\w/-]?)?|[.,;?!=+/#]|\s)+'.encode())
+
+_ZWSP = "\u200B"
+
+_MODEL_HANDLE = "https://tfhub.dev/google/zh_segmentation/1"
+
+default_input_path = pathlib.Path(__file__).resolve(
+).parent.parent.parent.joinpath("Translations/zh_CN.po")
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--input_path", default=default_input_path,
+                    help="Path to the input .po file")
+parser.add_argument("--output_path", help="Output path (default: in-place)")
+parser.add_argument("--separator", default=_ZWSP,
+                    help="Separator to use between segments")
+parser.add_argument("--debug",
+                    help="If this flag is given, segments the given string, joins with \"|\", and prints it.")
+args = parser.parse_args()
+
+_SEPARATOR = args.separator.encode()
+_SEGMENTER = tensorflow_text.HubModuleTokenizer(_MODEL_HANDLE)
+
+
+def _RunSegmenter(text, separator):
+	"""Runs the segmenter and produces a separator-joined string."""
+	if not text:
+		return []
+	text = _RemoveAllMarkers(text)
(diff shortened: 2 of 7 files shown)


glebm:
@tytannial Can you please play-test this a bit?

tytannial:
> @tytannial Can you please play-test this a bit?

I will test it after work. Thanks.

Source/engine/render/text_render.cpp
```diff
@@ -288,13 +288,17 @@ std::string WordWrapString(string_view text, size_t width, GameFontTables size,
...
 		}
 
-		uint8_t frame = codepoint & 0xFF;
-		uint32_t unicodeRow = codepoint >> 8;
-		if (unicodeRow != currentUnicodeRow || kerning == nullptr) {
-			kerning = LoadFontKerning(size, unicodeRow);
-			currentUnicodeRow = unicodeRow;
+		if (codepoint == ZWSP) {
+			lineWidth += spacing;
+		} else {
```
AJenbo, line 293: ZWSP should not add spacing.

```suggestion
		if (codepoint != ZWSP) {
```

tools/zh_segmenter/segment.py
```diff
@@ -0,0 +1,126 @@
...
+import pathlib
+import os
+import re
+import sys
+
+import tensorflow_text
+
+_DISALLOW_SEGMENTATION = re.compile(r'(?:\{[^\}]*\}|\\[a-z]|\s)+'.encode())
+
+_ZWSP = "\u200B"
```
AJenbo, line 12: Should we strip ZWSP before running the model, or is it able to cope with previously inserted break points?

tools/zh_segmenter/segment.py
```diff
@@ -0,0 +1,126 @@
...
+		if cur != m.start():
+			result.extend(_RunSegmenter(text[cur:m.start()]))
+		result.append(text[m.start():m.end()])
+		cur = m.end(0)
+	if cur != len(text):
+		result.extend(_RunSegmenter(text[cur:]))
+	return result
+
+
+# We handle these at runtime, no need to mark it.
```
AJenbo, line 70: I suppose alternatively we could simplify it so that this is also expected to be done at build time, thought that might not be idea for testing/Japanese?

AJenbo:
@nsm53project is similar processing needed for Korean or does the current methods work well?

glebm:
@AJenbo Korean has spaces, no processing is needed/

tools/zh_segmenter/segment.py
glebm: We do this in `_RemoveSegmentationMarkers`.

Source/engine/render/text_render.cpp
glebm: Hmm, it doesn't look good with this change.

AJenbo:
Yeah, I just forgot what frequency they where at (sentence or word), but they are by word so no issue there.

AJenbo:
@glebm do you know if there is a model for Japanese?

tools/zh_segmenter/segment.py
glebm: We could but let's keep it as is for now, can change later if we want to after thinking about it.

glebm:
@AJenbo Korean is per-word. There are models out there for Japanese, something we can tackle in the future once there is a Japanese translation effort.

Source/engine/render/text_render.cpp
glebm: Odd, I guess spacing is needed somehow:

image

tytannial:
Did they release a Japan edition on ps1?

Source/engine/render/text_render.cpp
glebm: Anyway, this is clearly the correct change, so I've applied it.
The text going all the way to the edge bug must be elsewhere.

glebm:
Yeah looks like there was a Japanese PS1 release https://www.youtube.com/watch?v=8FcJi7PeWoU

glebm:
It word-wraps in the middle of the words in the official Japanese PS1 release :laughing: 
image

AJenbo:
> @AJenbo Korean is per-word.

Yeah that's what I was saying :)

> There are models out there for Japanese, something we can tackle in the future once there is a Japanese translation effort.

Nice.

Source/engine/render/text_render.cpp
AJenbo: I might actually have forgotten to adjust the CJK width after we did the final render of the font, I'll have a look sometime this week.

glebm: OK, sounds like it's not a blocker for merging this! :+1:

tytannial:
I think.....we should just use that text break line way they publish on ps1. In Chinese, you can change line anywhere, except the punctuation mark. Comma, point, ellipsis, etc should not appear at start. 😶

Not perfect but M&B 2 bannerlord use those rule.

tytannial:
If that confused, there is a clear doc, about the line break.
https://stackoverflow.com/questions/44742217/whats-the-difference-between-word-break-normal-and-word-break-keepall
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break 🤣

1. normal: Follow normal line break rules, that is to break line at the space between words. So even if the last words goes out of bounds of the container, sentence won't go to next line till next word comes. This will be done for all text even for CJK characters (Chinese, Japanese, Korean and derivatives)
2. break-all: Break at the characters if out of bounds, that means a word itself will be broken and taken to the second line. So suppose ALongWord is going out of bounds at the AL then break-all will make the rest ongWord go to the second line. This will not be done for the CJK characters.
3. keep-all: Break by normal line rules except for the CJK characters. It's like normal except in case of CJK it won't break at all (neither by line rules as done by normal nor at characters as done by break-all)

glebm:
Fair enough, I wanted to do better but perhaps my effort is misguided :sweat_smile:

glebm:
I guessed that the acceptance of mid-word line breaks in Chinese and Japanese came initially from software limitations.

I thought that breaking between words would result in text that is easier to read and is aesthetically more pleasing, as in the example here https://github.com/google/budou

Am I completely wrong on this one? Is this just what people are used to, and it looks odd otherwise?

nsm53project:
@AJenbo Korean is no problem. and it would be good to reduce the character spacing in the next release.

nsm53project:
image

ファーンハム、可哀想に。あの暗黒の日、ラザルスと共に修道院に入っていった不幸な人々のことを思い出すよ。あの男は生きて戻りはしたが、勇気も正気もあの闇の中に置いてきたのじゃな。今では酒がないと心が安まらないようじゃ。いつもぶつぶつ言っているが中には真実が込められていることもあるぞ。

In Japanese, some characters overlap. 

[ja_JP.zip](https://github.com/diasurgical/devilutionX/files/7495053/ja_JP.zip)
(Cain only...)

glebm:
I've spoken to another Chinese friend, who has a background in linguistics, and he isn't sure which approach is better either.

Given that this is already implemented and is an improvement over the current situation, I suggest we go with this for now.

nsm53project:
image

image

Chinese and Japanese do not break lines based on words. CJ games have limits usually on the number of characters or pixels-per-line.

But we know an easy(?) solution: Enter key :)

怀特的故事是一个可怕而悲惨的故事。他
从他母亲的怀抱中被带到迷宫里，被那些
挥舞邪恶长矛的邪恶小恶魔们拖进了迷宫。
那天还有许多孩子被带走，包括李奥瑞克
国王的儿子。宫殿的骑士们下楼去了，但
再也没有回来。铁匠找到了那个男孩，但
那是在那些邪恶的野兽开始折磨他以获取
他们的虐待狂般的快乐之后

ファーンハム、可哀想に。あの暗黒の日、
ラザルスと共に修道院に入っていった不
幸な人々のことを思い出すよ。あの男は
生きて戻りはしたが、勇気も正気もあの
闇の中に置いてきたのじゃな。今では酒
がないと心が安まらないようじゃ。いつ
もぶつぶつ言っているが中には真実が込
められていることもあるぞ。

Source/engine/render/text_render.cpp
glebm: This was fixed by #3457

AJenbo:
> But we know an easy(?) solution: Enter key :)

The problem with that is that it makes editing tedious and if we ever make adjustments to the font or way the text is displayed then it breaks things, not to mention it won't work if it's displayed in two different locations.

AJenbo (approved):
This seams to do a good job and alleviates manual work

glebm:
In the original screenshot, word boundaries happen to line up with the edges.

Here is a few screenshots where word-based segmentation actually makes a difference:

image

image

image

Issue #3423, opened 2021-11-07
Segault on devilution::UpdatePalette in recent code


vanfanel:
**Important information**
Operating system / DevilutionX version / did you download a release or did you compile yourself?

Debian 11, latest DevilutionX GIT code, I built DevilutionX from code

**Describe**
Engine segfaults when loading the in-game graphics (or so it seems) as soon as the loading progress bar is about to get to the rightmost end.

I built a debug version to help diagnose the cause and do a complete and (I hope) helpful report.
This is what I get from a GDB session:

```
Thread 1 "devilutionx" received signal SIGSEGV, Segmentation fault.
devilution::UpdatePalette (art=0x5555d492e8 <devilution::(anonymous namespace)::PanelFull>, output=0x0)
    at /root/src/devilutionX/Source/DiabloUI/art_draw.cpp:12
12              if (art->surface->format->BitsPerPixel != 8)
(gdb) p art
$1 = (devilution::Art *) 0x5555d492e8 <devilution::(anonymous namespace)::PanelFull>
(gdb) p art->surface
$2 = std::unique_ptr<SDL_Surface> = {get() = 0x0}
(gdb) bt
#0  devilution::UpdatePalette (art=0x5555d492e8 <devilution::(anonymous namespace)::PanelFull>, output=0x0)
    at /root/src/devilutionX/Source/DiabloUI/art_draw.cpp:12
#1  0x00000055557e2a10 in devilution::LoadCharPanel () at /root/src/devilutionX/Source/panels/charpanel.cpp:252
#2  0x00000055555f70ec in devilution::InitControlPan () at /root/src/devilutionX/Source/control.cpp:955
#3  0x000000555561ce90 in devilution::LoadGameLevel (firstflag=true, lvldir=devilution::ENTRY_MAIN)
    at /root/src/devilutionX/Source/diablo.cpp:2026
#4  0x0000005555666a64 in devilution::ShowProgress (uMsg=devilution::WM_DIABNEWGAME) at /root/src/devilutionX/Source/interfac.cpp:259
#5  0x0000005555615748 in devilution::(anonymous namespace)::StartGame (uMsg=devilution::WM_DIABNEWGAME)
    at /root/src/devilutionX/Source/diablo.cpp:157
#6  0x000000555561752c in devilution::(anonymous namespace)::RunGameLoop (uMsg=devilution::WM_DIABNEWGAME)
    at /root/src/devilutionX/Source/diablo.cpp:689
[...]

PR #3439, opened 2021-11-08
show error in LoadArt

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/3423

### DIFF
--- a/Source/DiabloUI/art.cpp
+++ b/Source/DiabloUI/art.cpp
@@ -102,7 +102,7 @@ Uint32 GetPcxSdlPixelFormat(unsigned bpp)
 
 } // namespace
 
-void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette, const std::array<uint8_t, 256> *colorMapping)
+void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette, const std::array<uint8_t, 256> *colorMapping, bool showError)
 {
 	if (art == nullptr || art->surface != nullptr)
 		return;
@@ -114,6 +114,8 @@ void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette, con
 	std::uint8_t bpp;
 	SDL_RWops *handle = OpenAsset(pszFile);
 	if (handle == nullptr) {
+		if (showError)
+			ErrDlg("OpenAsset failed", pszFile, __FILE__, __LINE__);
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


Source/DiabloUI/diabloui.cpp
```diff
@@ -510,7 +510,7 @@ void LoadHeros()
 		Art portrait;
 		char portraitPath[18];
 		sprintf(portraitPath, "ui_art\\hero%i.pcx", i);
-		LoadArt(portraitPath, &portrait);
+		LoadArt(portraitPath, &portrait, 1, nullptr, nullptr, false);
```
obligaron, line 513: This is the only case where the `showError` parameter is used.
How can `LoadArt` ever succeed in this case (`heroXYZ.pcx` exists)?
Is this only possible when loading a custom assets/mpqs?

Source/DiabloUI/art.cpp
```diff
@@ -102,7 +102,7 @@ Uint32 GetPcxSdlPixelFormat(unsigned bpp)
 
 } // namespace
 
-void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette, const std::array<uint8_t, 256> *colorMapping)
+void LoadArt(const char *pszFile, Art *art, int frames, SDL_Color *pPalette, const std::array<uint8_t, 256> *colorMapping, bool showError)
```
obligaron, line 105: We only have one call site that passes `showError` = `false`.
Could we remove this callsite (see other comment) or should we add a `ExistsAsset` method instead of adding this to `LoadArt`?

obligaron (commented):
Could non existing assets also cause errors in `LoadPNG` or `SVidPlayBegin`?

Source/DiabloUI/diabloui.cpp
qndel: Yes. This is for loading optional assets so you don't want to throw an error when they don't exist.

Source/DiabloUI/art.cpp
qndel: up to you : P

obligaron: If no one uses it => remove it, cause less code is better code. 😜 
But I'm [not sure](https://i2.wp.com/scifiinterfaces.com/wp-content/uploads/2018/10/Idiocracy_ippa05.png?resize=768%2C432&ssl=1) if no one uses it.

obligaron: @AJenbo I looked in the git history and it seems like you added the logic in `LoadHeros`. What do you think?

AJenbo: I would prefer to keep it as I am working on content for it to load :sweat_smile:

obligaron: I think we should introduce something like a `ExistsAsset` method and crash if any error happens in `LoadArt` that didn't create a surface.
We do the same thing for sounds (see `sound_file_load`) and I think that helps to diagnose odd setups easier.

PR #3440, opened 2021-11-08
Add VS configuration for building in WSL using GCC

ephphatha:
This is mainly to help me test code builds under GCC faster than pushing to the remote and letting the build server do it. If someone without WSL installed selects the build target they'll get an error in their IDE, but no harmful impact other than that. The variables should be populated for any WSL2 installation. No idea about WSL1.

The whitespace in this file is showing significant changes because visual studio regenerates the whole file when a change is made, I didn't think it was worth reverting since it'll happen again next time someone makes a change (seems it defaults to spaces...)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -48,7 +48,7 @@ end_of_line = crlf
 [AppRun]
 end_of_line = lf
 
-[CMakeLists.txt]
+[{CMakeLists.txt,CMakeSettings.json}]
 indent_style = space
 indent_size = 2
 end_of_line = crlf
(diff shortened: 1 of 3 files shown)


glebm:
Please update .editorconfig with an entry for CMakeSettings.json corresponding to what VS generates.

.editorconfig
```diff
@@ -48,7 +48,7 @@ end_of_line = crlf
 [AppRun]
 end_of_line = lf
 
-[CMakeLists.txt]
+[{CMakeLists.txt,CMakeSettings.json}]
```
glebm, line 51: Hmm, last time I tried, multiple patterns are not supported by all the tools. hope this works in vs code

glebm (approved):

AJenbo (approved):

PR #3446, opened 2021-11-08
OpenDingux: Use assets directly

glebm:
Saves ~750 KiB.

retrofw: 4.9 MiB -> 4.1 MiB
rg350: 5.4 MiB -> 4.7 MiB


AJenbo (approved):

AJenbo:
Nice :) Should also improve loading/file access I suppose.

PR #3453, opened 2021-11-09
Ukrainian: TL support strings and some status bar msg

tsunamistate:
Continuing my Ukrainian TL when time permits, reviews welcome

Known issues:

* On debug builds, going to a new game or "Support" menu hits an assert

TL notes:

* I am fully committing to TL "Spell" as "Чари"
* IMO, "Rogue" should stay as "Пройдисвітка" because of DnD class origins


AJenbo:
I think the error is fixed in the latest master

Issue #3444, opened 2021-11-08
Black screen on Nintendo 2DS


3F2A8C4AE9:
**Important information**
This was using the provided download from the Releases page, v1.3.0. I used the .cia version to install the app onto my 2DS. I originally tried this on a 2DS that was not fully up to date system-wise, but updating it had no impact on performance.

**Describe**
When you start the app, you can hear the audio of the Blizzard logo and if you interact with the game, you can hear the menu and the Tristram music and your character's footsteps/etc. But there is simply no video output regardless of settings in the diablo.ini file. 

The game screen does appear when you have to name your character, but it disappears again once you're finished. When you hit the Home button, you then can see the screen as it pulls away, but it disappears when you close the Home Menu once more. Putting the system to sleep will also show the game screen for a split second when waking. 

As I had a spare system lying around, I installed it the same way onto an Old 3DS, and the game works as expected. 

**To Reproduce**
1.) Install the game as a .cia through FBI onto a non-New 2DS. 
2.) Edit diablo.ini to disable the Hellfire intro (issue #3445).
3.) Attempt to open the game. 
4.) Experience black screen, but still can interact with the game. 

**Expected behavior**
The expected behavior is for the game to display anything on the top screen. 

**Screenshots**
The amusing thing is that even with the black screen, taking screenshots through Luma works as expected. Screenshots show the game screen in full despite nothing showing on the screen during play. 

**Additional context**
I've nothing further to add.

StephenCWills:
I wonder if this has something to do with shutting the code to shut the backlight off on the bottom screen.

https://github.com/diasurgical/devilutionX/blob/69e7e7bc0d25ff1b294582bf5836318fad610977/Source/platform/ctr/system.cpp#L77-L78

StephenCWills:
@3F2A8C4AE9 Can you please test this build to see if it works better?
https://70551-143324737-gh.circle-artifacts.com/0/devilutionx.cia

3F2A8C4AE9:
> 
> 
> @3F2A8C4AE9 Can you please test this build to see if it works better? https://70551-143324737-gh.circle-artifacts.com/0/devilutionx.cia

Hey, sorry it took me a while to get back.  I gave this new version a spin. 

[...]

PR #3455, opened 2021-11-09
Enable backlight on bottom screen

StephenCWills:
For testing on 2DS (#3444)


StephenCWills:
Superseded by #3477

For the moment, this is still the best build to use on o2DS because it's close to 1.3.0 and does not suffer from crashes due to audio race conditions.

PR #3458, opened 2021-11-10
Remove 3dsx from docs and builds

StephenCWills:
Plus some additional cleanup.

Was anyone using the build.sh script? I think it missed a few of the recent updates to build dependencies. If no one's using it, I'd rather not maintain it.


AJenbo (approved):
:toilet:

MrHuu:
This stops non-hacked 3ds's from being able to run devilutionx, is this intended?
Was the .3dsx format causing issues? If so, anything i could help with?

It hurts to see support for the homebrew format being removed.

AJenbo:
Mostly it was confusing users, and us :D

I'm happy with it being restored based on what you say here. But we should probably mention in the ReadMe why one would choose one image over the other. If for no other reason then stopping us from deleting it again down the line :)

@StephenCWills was there an issue regarding how the 3dsx image loads files?

MrHuu:
The .3dsx format is able to be launched with just a flashcard on an original (read: not hacked) 3ds.
While the .cia format is Nintendo's proprietary format used for downloadable content. Which can only be installed on 3ds's running a custom firmware.

While most users probably prefer the .cia format for convenience, other users rely on the .3dsx format.

StephenCWills:
> This stops non-hacked 3ds's from being able to run devilutionx, is this intended?

No, that wasn't the intent. I thought you needed to be able to run hbmenu to use 3dsx so I was completely unaware that it could be used to play on a non-hacked console.

> Was the .3dsx format causing issues? If so, anything i could help with?

The problem is memory usage. Shortly before 1.3.0 was released, I lowered the `SystemMode` in `template.rsf` down to 64MB to improve startup time. However, after the release we started receiving reports from o3DS and o2DS users that the game would crash during the intro video so we had to increase it to 80MB again. Because the `template.rsf` file only applies to the cia binary, the crashes would still occur when running the game using the 3dsx binary. Perhaps this is not a problem for New 3DS models.

If you know of a way to access extended memory so the game doesn't crash on old models it would become a non-issue. Otherwise, we could reenable the 3dsx format and warn o3DS users away from it.

Also, I'm currently working to see if I can implement libsmackerdec to replace libsmacker so we could stream the intro video from the SD card instead of loading the whole thing into memory. If that works out, then o3DS users should be able to use the 3dsx format again without issue.

MrHuu:
I don't think it's possible to select the ```SystemMode``` when launching the .3dsx. It would depend on the entrypoint / exploit used to run the homebrew launcher. In some cases without required privileges i believe.
Unfortunately, my knowledge is limited on this subject.

At some point the homebrew launcher was able to launch a .3dsx piggybacking an official (.cia) application with the required NCCH (.rsf) definitions.
Not really sure how this is handled at this date.

From these o2ds/o3ds crash reports, is it by any chance known if they were running a CFW or an other *hax entrypoint?
Are the known cases from both old and new 3ds, or just old 3ds?
Were these crashes consistent, and would i be able to reliably reproduce the issue?

I never really had any issues using the .3dsx as opposed to the .cia. But this is on a New3DS with CFW.
If there's a way to keep on providing the .3dsx format, maybe even with known limitations, i would personally vote for that.
Even if we can't access the required memory, we should try to catch the error and provide a proper error message on unsupported devices.

StephenCWills:
> From these o2ds/o3ds crash reports, is it by any chance known if they were running a CFW or an other *hax entrypoint?

The initial crash reports came from users who installed via the cia. I reproduced them on my o3DS after reenabling the intro video. I ran the 3dsx using hbmenu to verify that it suffered from the same issue. It's my understanding that all parties involved were using CFW.

> Are the known cases from both old and new 3ds, or just old 3ds?

The reports were from o3DS and o2DS users. My guess is that because the New 3DS models support the `SystemModeExt` directive, they have access to more memory regardless of the context.

> Were these crashes consistent, and would i be able to reliably reproduce the issue?

Yes, you should be able to reproduce if you use o3DS. You might be able to reproduce on a New 3DS as well if you create a build using the `Legacy` option for `SystemModeExt`, but that's just my speculation.

Here are the CI builds from the commit before just I increased `SystemMode` to 80MB (3f85fda). FYI, these should not be affected by the audio race conditions that are currently causing crashes in the latest master.

https://69642-143324737-gh.circle-artifacts.com/0/devilutionx.3dsx
https://69642-143324737-gh.circle-artifacts.com/0/devilutionx.cia

MrHuu:
Great, thanks!

A quick test using the provided builds confirms the issue.
Both .3dsx and .cia exit (no crash) after showing the Blizzard North intro on an old3DS.

On a new3DS however, both builds are running and i'm able to get in-game seemingly without issues.

Both 3ds's used are running a custom firmware and share a similar setup.
I fail to understand why it does run on a new3DS, and under the same circumstances (homebrew environment ran from cfw) does not on the old3DS.

I'll have to build from source to be able to make a comparison between the two devices and see what resources are available in both cases.

PR #3459, opened 2021-11-10
Update platform manual for 3DS

StephenCWills:
I found some more docs.

### DIFF
--- a/docs/manual/platforms/n3ds.md
+++ b/docs/manual/platforms/n3ds.md
@@ -1,22 +1,8 @@
 # DevilutionX (Diablo 1) for Nintendo 3DS
-*New3DS only
 
 ## Installation
 
-Start by downloading [devilutionx-3ds.zip](https://github.com/diasurgical/devilutionX/releases/latest/download/devilutionx-3ds.zip).
-
-<details><summary>.3dsx installation</summary>
-
-#### Install DevilutionX: Diablo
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Should this just be merged with the general install instructions?

StephenCWills:
> Should this just be merged with the general install instructions?

Let's see... It does have a 3DS-tailored version of the [Game Controller Scheme](https://github.com/diasurgical/devilutionX/wiki/Game-Controller-Scheme) page on the wiki, and it describes usage for the touchscreen FWIW. Oddly enough, it seems that the build instructions link to this document, but the installation instructions don't.

I guess we should merge with the general install instructions for the sake of consistency. In fact, we can drop that section and maybe link back to the general install instructions page or vice-versa. However, rather than dropping this manual entirely, we might consider fleshing this out a bit in other areas. I think it was @stefanmielke who once suggested we could add tips for boosting performance on o3DS models, and I think it's a good idea.

> The file name also implies that it's for the new 3DS and not original.

I think n3ds is actually Nintendo 3DS. This is an abbreviation [used by devkitPro's SDL branch](https://github.com/devkitPro/SDL/tree/3ds-sdl-1.2/src/audio/n3ds). Admittedly, using the term o3DS does tend to confuse things so I suppose we should probably rename it anyway.

Issue #3464, opened 2021-11-10
Changed item variable types might cause issues - for example broken hidden shrine


qndel:
Caused by https://github.com/diasurgical/devilutionX/commit/461da3cda0f2d13d7a927365e6460f76ee013ac9
So is in 1.2.1 and 1.3.0.

I can see all kinds of issues caused by this change - earlier durability got capped to 1-255 manually and it was fine, now there's a lot of code that assumes the type is int

for example
```cpp
	case IPL_DUR_CURSE:
		item._iMaxDur -= r * item._iMaxDur / 100;
		item._iMaxDur = std::max<uint8_t>(item._iMaxDur, 1);
```

Hidden shrine:
```cpp
			player.InvBody[r]._iDurability -= 20;
			player.InvBody[r]._iMaxDur -= 20;
			if (player.InvBody[r]._iDurability <= 0)
				player.InvBody[r]._iDurability = 1;
			if (player.InvBody[r]._iMaxDur <= 0)
				player.InvBody[r]._iMaxDur = 1;
```

earlier adding 10 dura and subtracting 20 was fine because the type was int, now low durability items will get flipped into 250ish dura and items with durability around 250 will get flipped into low durability after adding 10.

All variables changed by mentioned commit should be checked for such a behavior.

`#edit`
checked changes from that commit - seems only durability has issues

SourceCodeDeleted:
I am suprised it was changed away from int .

glebm:
>  uint8_t is supposed to by only to 255 no?

So is durability and max durability in Diablo, but, as @qndel points out, the code does not do arithmetic on it correctly, requiring an intermediate large value.

qndel:
> > uint8_t is supposed to by only to 255 no?
> 
> So is durability and max durability in Diablo, but, as @qndel points out, the code does not do arithmetic on it correctly, requiring an intermediate large value.

We could stay with uint8_t and rewrite all the math to stay within bounds but imo it's simply not worth it when we can just switch back to int :D

PR #3467, opened 2021-11-10
make durability int again

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/3464


glebm:
It'd be nice if fixed this where it's used (as described in the linked issue) as I really want to keep things only as small as necessary.

glebm (approved):

PR #3475, opened 2021-11-11
Always draw hotkeys

vlolteanu:
This makes hotkeys always show up in the spell bar, as well as for the selected spell. It mimics DIablo 2.

We could also add an INI option for this.

### DIFF
--- a/Source/panels/spell_list.cpp
+++ b/Source/panels/spell_list.cpp
@@ -81,6 +81,17 @@ bool GetSpellListSelection(spell_id &pSpell, spell_type &pSplType)
 	return false;
 }
 
+std::optional<std::string> GetHotkeyName(spell_id spellId, spell_type spellType)
+{
+	auto &myPlayer = Players[MyPlayerId];
+	for (int t = 0; t < 4; t++) {
+		if (myPlayer._pSplHotKey[t] != spellId || myPlayer._pSplTHotKey[t] != spellType)
+			continue;
+		return keymapper.KeyNameForAction(quickSpellActionIndexes[t]);
+	}
+	return {};
+}
+
 } // namespace
 
 void DrawSpell(const Surface &out)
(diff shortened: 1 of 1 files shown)


vlolteanu:
screen11

Source/control.cpp
```diff
@@ -648,6 +648,21 @@ bool IsChatAvailable()
 #endif
 }
 
+namespace {
```
obligaron, line 651: `control.cpp` already has a anonymous namespace.
Should we move `GetHotkeyName` inside of the existing one?

obligaron (approved):
lgtm.

My personal preference would be to not draw the hotkey in `DrawSpell` (active spell at buttom right), cause i don't think that it will help much. But you can discard that.

Another location where we could draw it would be the spell book.

Source/control.cpp
```diff
@@ -747,13 +771,9 @@ void DrawSpellList(const Surface &out)
...
 		}
-		for (int t = 0; t < 4; t++) {
-			if (myPlayer._pSplHotKey[t] == spellId && myPlayer._pSplTHotKey[t] == spellListItem.type) {
-				auto hotkeyName = keymapper.KeyNameForAction(quickSpellActionIndexes[t]);
-				PrintSBookHotkey(out, spellListItem.location, hotkeyName);
-				strcpy(tempstr, fmt::format(_("Spell Hotkey {:s}"), hotkeyName.c_str()).c_str());
-				AddPanelString(tempstr);
-			}
+		if (hotkeyName) {
+			strcpy(tempstr, fmt::format(_("Spell Hotkey {:s}"), hotkeyName->c_str()).c_str());
```
glebm, line 775:
```suggestion
			strcpy(tempstr, fmt::format(_("Spell Hotkey {:s}"), *hotkeyName).c_str());
```

Source/control.cpp
vlolteanu: You're right.

glebm:
Conflicts are caused because moved the spell stuff out of `control.c` into separate files in https://github.com/diasurgical/devilutionX/pull/3585

AJenbo (approved):

AJenbo:
Just noticed that the short cut is not displayed if the spell is disabled (firebolt in town)

vlolteanu:
> Just noticed that the short cut is not displayed if the spell is disabled (firebolt in town)

Is the spell type set to `RSPLTYPE_INVALID` while in town?

AJenbo:
idk

PR #3478, opened 2021-11-12
Update Appveyor image to VS2022

tsunamistate:


AJenbo:
A few new warnings regarding libzt, why not :shrug:

PR #3480, opened 2021-11-12
Create ja.po

bubio:
added translation to Japanese!

It still needs to be tweaked, but it's mostly translated.

There may be a problem with the handling of line breaks, but the lines are displayed overlapping each other.

image

### DIFF
--- a/Translations/ja.po
+++ b/Translations/ja.po
@@ -1,0 +1,10603 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: DevilutionX\n"
+"POT-Creation-Date: 2021-10-31 23:17+0100\n"
+"MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


AJenbo:
> https://drive.google.com/file/d/1eCVuDsdoU7K2IvVvu8bndzujE446pSrZ/view?usp=sharing

What is this and why does it want my Google account :o

bubio:
Can I access this?

https://drive.google.com/file/d/1eCVuDsdoU7K2IvVvu8bndzujE446pSrZ/view?usp=drivesdk

AJenbo:
You can just post it directly in the comments instead of using Google drive and links

glebm:
あざっす！Don't worry about line breaks and overlap, I'll work on that.

PR #3482, opened 2021-11-13
UK TL: Status bar, CMD options, Help screen

tsunamistate:
@SourceCodeDeleted please review when you have the time

Scrolling help screen past Automap hits xstring assert "cannot remove prefix longer than total size" on debug builds

I have changed "Гра з друзями" to "Мережева гра" because it sounds better, and looking at some Ukrainian localizations, this is the precedent.

### DIFF
--- a/Translations/uk.po
+++ b/Translations/uk.po
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: DevilutionX\n"
-"POT-Creation-Date: 2021-11-09 13:09+0200\n"
+"POT-Creation-Date: 2021-11-11 15:51+0200\n"
 "PO-Revision-Date: \n"
 "Last-Translator: Oleksandr Kalko (tsunami_state) <tsunami.wave@ukr.net>\n"
 "Language-Team: \n"
@@ -391,8 +391,8 @@ msgstr "Клієнт-Сервер (TCP)"
 msgid "Loopback"
 msgstr "Петльова адреса"
 
-#: Source/DiabloUI/selconn.cpp:53 Source/DiabloUI/selgame.cpp:484
-#: Source/DiabloUI/selgame.cpp:505
+#: Source/DiabloUI/selconn.cpp:53 Source/DiabloUI/selgame.cpp:489
+#: Source/DiabloUI/selgame.cpp:510
 msgid "Multi Player Game"
 msgstr "Гра з Друзями"
 
@@ -1000,7 +1000,7 @@ msgstr ""
 msgid "Hotkey: 's'"
 msgstr ""
 
-#: Source/control.cpp:1367 Source/inv.cpp:1992 Source/items.cpp:3739
+#: Source/control.cpp:1367 Source/inv.cpp:1979 Source/items.cpp:3729
 msgid "{:d} gold piece"
 msgid_plural "{:d} gold pieces"
 msgstr[0] ""
@@ -2036,7 +2036,7 @@ msgid "Quilted Armor"
 msgstr ""
 
 #: Source/itemdat.cpp:74 Source/itemdat.cpp:75 Source/itemdat.cpp:76
-#: Source/itemdat.cpp:77 Source/objects.cpp:5423
+#: Source/itemdat.cpp:77 Source/objects.cpp:5433
 msgid "Armor"
 msgstr ""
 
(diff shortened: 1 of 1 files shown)


Translations/uk.po
```diff
@@ -1116,192 +1116,198 @@ msgstr "А тепер ПОМРИ!"
...
 
 #: Source/error.cpp:65
 msgid "No pause in multiplayer"
-msgstr ""
+msgstr "Пауза недоступна в мережевій грі"
 
 #: Source/error.cpp:66
 msgid "Loading..."
-msgstr ""
+msgstr "Завантажуємо..."
```
SourceCodeDeleted, line 1306: Is there a reason you don't  want to say завантажується instead? I mean to me it seems slightly weird to hear we are loading when we can say , it is loading.

Translations/uk.po
```diff
@@ -1116,192 +1116,198 @@ msgstr "А тепер ПОМРИ!"
...
 
 #: Source/error.cpp:67
 msgid "Saving..."
-msgstr ""
+msgstr "Зберігаємо..."
```
SourceCodeDeleted, line 1310: The same here, 

зберігається seems more correct imho

Translations/uk.po
```diff
@@ -1579,120 +1579,128 @@ msgstr "Пауза"
...
 
 #: Source/help.cpp:40
 msgid "1 - 8: Use Belt item"
-msgstr ""
+msgstr "1 - 8: Використати предмет з Поясу"
 
 #: Source/help.cpp:41
 msgid "F5, F6, F7, F8:     Set hotkey for skill or spell"
-msgstr ""
+msgstr "F5, F6, F7, F8:     Встановити гарячу клавішу для чар або таланту"
```
SourceCodeDeleted, line 1642: How are we translating skill? Talant? 

I think навичка is better, but not a big problem.

Translations/uk.po
```diff
@@ -1702,20 +1710,27 @@ msgid ""
...
 "You can select a specific amount of gold to drop by right-clicking on a pile "
 "of gold in your inventory."
 msgstr ""
+"Ви можете визначити точну кількість золота для викидання кліком правої "
+"кнопки миші на купі золота в інвентарі."
 
 #: Source/help.cpp:72
 msgid "$Skills & Spells:"
-msgstr ""
+msgstr "$Таланти і Чари:"
```
SourceCodeDeleted, line 1733: навички та чари ? 

how do you think about this?

SourceCodeDeleted:
ok I added some comments. I don't think we should say "we" МИ in this game when we describe actions perfoemed, to me it seems weird. See comments above. 

Other than this, looks good.

Translations/uk.po
tsunamistate: Disagree on "Навичка", since it's a habit, not a skill
"Талант" is closer to game's meaning of "Skill" since it's an innate skill of a character that they can use immediately.

Translations/uk.po
tsunamistate: As I said above, I disagree with "Навички", but "та" instead of "i" is better

Translations/uk.po
tsunamistate: OK, done

Translations/uk.po
tsunamistate: Done

tsunamistate:
@SourceCodeDeleted thank you for reviewing. While I disagree with 1 suggestion, others are good, and I applied them. Will it be OK to ping you every time I complete a PR for review?
We should chat on Discord (I am Tsunami State#0013), since I need help on sections where I'm stuck

@AJenbo good to merge!

SourceCodeDeleted:
@tsunamistate  yes, this is fine. Which particular part did you disagree with ?

tsunamistate:
The "Skill" TL from "Талант" то "Навичка". Навичка is a habit, not a skill

SourceCodeDeleted:
> The "Skill" TL from "Талант" то "Навичка". Навичка is a habit, not a skill

No, this isn't correct.
Look here for example. 

" За тиждень чи місяць ніхто не стане програмістом. Потрібен час, щоб вивчити теорію та досконало оволодіти практичними **навичками**."

Талант is more like a inherent skill or ability like the word Talent in English. Навичка is more of a skill which you develop.  

If you are looking for the word Habit then звичка is the word you are looking for.

tsunamistate:
@SourceCodeDeleted let's continue this in Discord to avoid GitHub email spam on closed commit :)

PR #3489, opened 2021-11-13
improve dropping items

### LINKED ISSUE #3514: Can drop items inside houses
image
While it's not a big issue with vanilla's item drop algorithm as you can simply pick the item up, it ruins an algorithm that assumes all tiles around an item are reachable :(

### LINKED ISSUE #4124: Desync when killing a player with 18+ gold stacks
**Important information**
Windows / DevilutionX 3b3419d / compiled MSVC 2022

**Describe**
If the host of a multiplayer game kills another player with a melee attack the logic for dropping the slain players items causes the host to see only 9 stacks/items appear on the ground. The host generates the expected messages and all other players see the correct state, and if the host leaves the level then returns they also see the correct state when the level delta messages are played back.

**To Reproduce**
1. Join a game with two clients.
2. Use `give gold` on the second player or ensure they have 18 or more full stacks of gold
3. Enter the dungeon with both players
4. Have the host kill the second player with a melee attack
5. Desync is visible at this point
6. Have the host leave the level then re-enter
7. Desync is corrected

**Expected behavior**
Items to appear the same on all clients the first time.

**Screenshots**
Step 5 desync:
2022-03-06T142103_devilutionx

Step 7 resync:
2022-03-06T142136_devilutionx
(One of the gold piles on the stairs got picked up by the warrior due to auto-gold pickup)

**Additional context**
This appears to be due to the logic that stops dropping items when the local player has an item at their feet. Dead items usually ignore the distance logic and allow dropping further away than adjacent tiles however they also shouldn't drop on a tile where a living player is standing, notice the dead player in the screenshots does not see an item at the feet of the living player.

qndel:
Allows dropping items further than only tiles adjacent to the character and prevents them from dropping in unreachable places.
@Chance4us can you test? : P

Resolves https://github.com/diasurgical/devilutionX/issues/3514 in a super ugly way 🙃 
Fixes https://github.com/diasurgical/devilutionX/issues/4124
```cpp
bool CanPut(Point position)
{
	// TODO: remove after collision of houses has been fixed - see https://github.com/diasurgical/devilutionX/issues/3514
	// clang-format off
	if (IsAnyOf(position,
		Point { 53, 45 }, Point { 46, 45 }, Point { 46, 44 }, // house near portal location
		Point { 27, 16 }, Point { 29, 18 }, Point { 27, 27 }, // cathedral
		Point { 38, 67 }, Point { 38, 66 }, Point { 41, 67 }, Point { 41, 64 }, Point { 40, 64 }, // gillian's house
		Point { 71, 81 }, Point { 68, 81 }, Point { 68, 80 }, Point { 70, 78 }, Point { 71, 78 }, // house near farnham
		Point { 73, 71 }, // house near spawn
		Point { 60, 61 }, Point { 60, 60 }, Point { 62, 59 }, Point { 62, 58 }, // griswold's house
		Point { 9, 45 } // stones around wirt
	))
	// clang-format on
		return false;
```

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -7,6 +7,7 @@
 
 #include <algorithm>
 #include <fmt/format.h>
+#include <queue>
 
 #include "DiabloUI/ui_flags.hpp"
 #include "controls/plrctrls.h"
@@ -1097,6 +1098,37 @@ void CleanupItems(int ii)
 	}
 }
 
+std::optional<Point> FindClosestValidItemPosition(Point startingPosition)
+{
+	if (CanPut(startingPosition)) {
+		return startingPosition;
+	}
+	if (dItem[startingPosition.x][startingPosition.y] == 0)
+		return {};
+
+	bool checkedTiles[MAXDUNX][MAXDUNY] = { false };
+	std::queue<Point> tilesToCheck;
+	tilesToCheck.push(startingPosition);
+	while (!tilesToCheck.empty()) {
+		Point currentPosition = tilesToCheck.front();
+		tilesToCheck.pop();
+		for (auto dir : PathDirs) {
+			Point newPosition = currentPosition + dir;
+			if (InDungeonBounds(newPosition)) {
+				if (!checkedTiles[newPosition.x][newPosition.y]) {
+					checkedTiles[newPosition.x][newPosition.y] = true;
+					if (CanPut(newPosition))
+						return newPosition;
+					if (dItem[newPosition.x][newPosition.y] != 0)
+						tilesToCheck.push(newPosition);
+				}
+			}
+		}
+	}
+
+	return {};
+}
+
 bool PutItem(Player &player, Point &position)
 {
 	if (ActiveItemCount >= MAXITEMS)
@@ -1107,24 +1139,8 @@ bool PutItem(Player &player, Point &position)
 	if (position.WalkingDistance(player.position.tile) > 1) {
 		position = player.position.tile + d;
 	}
-	if (CanPut(position))
-		return true;
 
-	position = player.position.tile + Left(d);
-	if (CanPut(position))
-		return true;
-
-	position = player.position.tile + Right(d);
-	if (CanPut(position))
-		return true;
-
-	position = player.position.tile + Left(Left(d));
-	if (CanPut(position))
-		return true;
-
-	position = player.position.tile + Right(Right(d));
-	if (CanPut(position))
-		return true;
+	std::optional<Point> itemPosition = FindClosestValidItemPosition(position);
 
 	position = player.position.tile + Left(Left(Left(d)));
 	if (CanPut(position))
@@ -1755,7 +1771,28 @@ void SyncGetItem(Point position, int32_t iseed, _item_indexes idx, uint16_t ci)
 
 bool CanPut(Point position)
 {
+	// TODO: remove after collision of houses has been fixed - see https://github.com/diasurgical/devilutionX/issues/3514
+	if (IsAnyOf(position,
+	        Point { 53, 45 }, Point { 46, 45 }, Point { 46, 44 },                                     // house near portal location
+	        Point { 27, 16 }, Point { 29, 18 }, Point { 27, 27 },                                     // cathedral
+	        Point { 38, 67 }, Point { 38, 66 }, Point { 41, 67 }, Point { 41, 64 }, Point { 40, 64 }, // gillian's house
+	        Point { 71, 81 }, Point { 68, 81 }, Point { 68, 80 }, Point { 70, 78 }, Point { 71, 78 }, // house near farnham
+	        Point { 73, 71 },                                                                         // house near spawn
+	        Point { 60, 61 }, Point { 60, 60 }, Point { 62, 59 }, Point { 62, 58 },                   // griswold's house
+	        Point { 9, 45 }                                                                           // stones around wirt
+	        )) {
+		return false;
+	}
+
 	if (!InDungeonBounds(position)) {
+		return false;
+	}
+
+	if (!InDungeonBounds(position + Displacement { 1, 1 })) {
+		return false;
+	}
+
+	if (dItem[position.x][position.y] != 0) {
 		return false;
 	}
 
@@ -1789,21 +1826,13 @@ bool TryInvPut()
 		return false;
 
 	auto &myPlayer = Players[MyPlayerId];
-
 	Direction dir = GetDirection(myPlayer.position.tile, cursPosition);
-	if (CanPut(myPlayer.position.tile + dir)) {
+	std::optional<Point> itemPosition = FindClosestValidItemPosition(myPlayer.position.tile + dir);
+
+	if (itemPosition) {
 		return true;
 	}
-
-	if (CanPut(myPlayer.position.tile + Left(dir))) {
-		return true;
-	}
-
-	if (CanPut(myPlayer.position.tile + Right(dir))) {
-		return true;
-	}
-
-	return CanPut(myPlayer.position.tile);
+	return false;
 }
 
 int InvPutItem(Player &player, Point position)
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1139,23 +1139,25 @@ bool PutItem(Player &player, Point &position)
 	if (position.WalkingDistance(player.position.tile) > 1) {
 		position = player.position.tile + d;
 	}
-
-	std::optional<Point> itemPosition = FindClosestValidItemPosition(position);
-
-	position = player.position.tile + Left(Left(Left(d)));
 	if (CanPut(position))
 		return true;
 
-	position = player.position.tile + Right(Right(Right(d)));
+	position = player.position.tile + Left(d);
 	if (CanPut(position))
 		return true;
 
-	position = player.position.tile + Opposite(d);
+	position = player.position.tile + Right(d);
 	if (CanPut(position))
 		return true;
 
-	position = player.position.tile;
-	return CanPut(position);
+	std::optional<Point> itemPosition = FindClosestValidPosition(CanPut, player.position.tile, 1, 50);
+
+	if (itemPosition) {
+		position = itemPosition.value();
+		return true;
+	}
+
+	return false;
 }
 
 bool CanUseStaff(Item &staff, spell_id spell)
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1098,37 +1098,6 @@ void CleanupItems(int ii)
 	}
 }
 
-std::optional<Point> FindClosestValidItemPosition(Point startingPosition)
-{
-	if (CanPut(startingPosition)) {
-		return startingPosition;
-	}
-	if (dItem[startingPosition.x][startingPosition.y] == 0)
-		return {};
-
-	bool checkedTiles[MAXDUNX][MAXDUNY] = { false };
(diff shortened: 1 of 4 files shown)


Chance4us:
> Allows dropping items further than only tiles adjacent to the character and prevents them from dropping in unreachable places. @Chance4us can you test? : P

I tested but can't decide.

What has always irritated me in general with this logic is the fact that if I aim at the top, then the item is dropped at the bottom. Or if I aim to the right, the item may end up on the left or on the right. The behavior is generally somewhat unpredictable. That was always kind of strange in this game.

I think the best thing to do would be to drop the item where the mouse pointer is, paying attention to whether this tile is free. But I don't know if it's that easy to do.

qndel:
Then nothing would stop you from accidentally dropping your stuff in unreachable places.

Chance4us:
> Then nothing would stop you from accidentally dropping your stuff in unreachable places.

Yes, we should merge it quickly, because it is an improvement in any case.

ephphatha:
The adjacency check currently bails if the search area contains a tile that isn't adjacent to an item, this causes the search to fail far quicker than necessary. I think this could be changed to continue searching if any tile in the current search radius contains an item.

2021-11-14T121039_devilutionx
In this screenshot the tile under the cursor is not adjacent to an item so I can't drop anything more, but removing the `if (adjacentFails)` check lets me continue dropping up/left/over the bridge.

Source/inv.cpp
```diff
@@ -1109,6 +1109,70 @@ void CleanupItems(Item *item, int ii)
...
+				Point pos = startingPosition + Displacement { xx, yy };
+				can_put_result res = CanPutAdjacent(pos);
+				if (res == CANPUT_OK)
+					return pos;
+				if (res == CANPUT_NO_ADJACENT_ITEMS)
+					adjacentFails++;
+			}
+		}
+		if (adjacentFails != 0)
+			break;
```
ephphatha, line 1170:
```suggestion
		bool continueFloodFill = false;
		for (int yy = -range; yy <= range; yy++) {
			// optimization to only check edges of the square
			// makes sure we don't check same tiles multiple times
			int step = range * 2;
			if (yy == -range || yy == range)
				step = 1;
			for (int xx = -range; xx <= range; xx += step) {
				Point pos = startingPosition + Displacement { xx, yy };
				can_put_result res = CanPutAdjacent(pos);
				if (res == CANPUT_OK)
					return pos;
				if (dItem[pos.x][pos.y] != 0)
					continueFloodFill = true;
			}
		}
		if (!continueFloodFill)
			break;
```
This allows items to continue to be dropped as long as the adjacency check has a chance to succeed on the next largest radius. Also lets you remove the can_put_result enum and have CanPutAdjacent return true/false.

AJenbo:
> The adjacency check currently bails if the search area contains a tile that isn't adjacent to an item, this causes the search to fail far quicker than necessary. I think this could be changed to continue searching if any tile in the current search radius contains an item.
> 
> 2021-11-14T121039_devilutionx In this screenshot the tile under the cursor is not adjacent to an item so I can't drop anything more, but removing the `if (adjacentFails)` check lets me continue dropping up/left/over the bridge.

Woudn't that let you drop things in areas that are impossible to reach or on the other side of walls?

ephphatha:
> Woudn't that let you drop things in areas that are impossible to reach or on the other side of walls?

Dropping on the other side of walls is only possible if there's an unbroken line of items to the space. This could potentially be an unpathable line given the item adjacency check counts diagonals and player pathing requires clear spaces on both sides of a diagonal for it to be a valid path. Fixing that would be feasible by only checking orthogonal (non-diagonal) tiles in the item adjacency code.

https://user-images.githubusercontent.com/357684/141664550-73450d0c-26f2-44d3-b233-313c6f328005.mp4
(with just the continueFloodFill check)

https://user-images.githubusercontent.com/357684/141664635-d4cb76f6-8c8c-4229-8fb0-930da72846b1.mp4
(with continueFloodFill and CanPutAdjacent updated to only check `{ Direction::NorthEast, Direction::NorthWest, Direction::SouthEast, Direction::SouthWest }`)

qndel:
How's this then? 
image

Chance4us:
> How's this then? image

Looks nice. I would also like to own that much gold.

qndel:
> Looks nice. I would also like to own that much gold.

`give gold` in debug build ;)

FitzRoyX:
Here's a place you can drop items in vanilla that makes them irretrievable. Pro griefer strat, just throw quest items over there. This PR does seem to fix that.

spot1

qndel:
@FitzRoyX how the hell do you drop items there in vanilla? These tiles aren't adjacent to character o.o

FitzRoyX:
Oh my bad. I assumed devx was vanilla in this regard since it's based on RE code. Was this behavior already altered once before?

qndel:
> Oh my bad. I assumed devx was vanilla in this regard since it's based on RE code. Was this behavior already altered once before?

yes ... that's exactly the point of this PR, to make the altered behavior usable (can't drop in unreachable places)

AJenbo:
Could have even tested in 1.3.0...

Chance4us:
@qndel, @AJenbo 
I have now tested the new behavior (335b0b934922172fb60779749e4e27b864c3d45f) over a longer period of time.
Due to the unpredictable dropping position, IMO the new logic is worse than vanilla, even if you can drop more items.
Unless the positioning could be made more logical and predictable. So the items should drop either in the north / east / west / south on the side of the mouse pointer direction.

qndel:
> @qndel, @AJenbo I have now tested the new behavior ([335b0b9](https://github.com/diasurgical/devilutionX/commit/335b0b934922172fb60779749e4e27b864c3d45f)) over a longer period of time. Due to the unpredictable dropping position, IMO the new logic is worse than vanilla, even if you can drop more items. Unless the positioning could be made more logical and predictable. So the items should drop either in the north / east / west / south on the side of the mouse pointer direction.

335b0b9 is not related to this PR

Chance4us:
> [335b0b9](https://github.com/diasurgical/devilutionX/commit/335b0b934922172fb60779749e4e27b864c3d45f) is not related to this PR

That is clear. I'll put the comment there.

qndel:
@Chance4us what about this one? is it fine?

Chance4us:
> @Chance4us what about this one? is it fine?

Where do I have to look? Is there a new PR? I probably missed something, didn't I?

qndel:
> > @Chance4us what about this one? is it fine?
> 
> Where do I have to look? Is there a new PR? I probably missed something, didn't I?

This PR lol

Chance4us:
> > > @Chance4us what about this one? is it fine?
> > 
> > 
> > Where do I have to look? Is there a new PR? I probably missed something, didn't I?
> 
> This PR lol

The aforementioned PR behaves almost like vanilla. You cannot drop more than 9 items. But the positioning is better. It feels very good. But if you want to drop more item, it won't be enough for us.

qndel:
Can we stop discussing other PRs here? Just tell me what you think about my one lol, I've changed it many times, have you tested the last version? Is it good?

Chance4us:
> Can we stop discussing other PRs here? Just tell me what you think about my one lol, I've changed it many times, have you tested the last version? Is it good?

Just tested. Feels pretty good.

qndel:
> > Can we stop discussing other PRs here? Just tell me what you think about my one lol, I've changed it many times, have you tested the last version? Is it good?
> 
> Just tested. Feels pretty good.

Thanks :)

Source/inv.cpp
```diff
@@ -1109,6 +1110,37 @@ void CleanupItems(Item *item, int ii)
...
+	if (dItem[startingPosition.x][startingPosition.y] == 0)
+		return {};
+
+	bool checkedTiles[MAXDUNX][MAXDUNY] = { false };
+	std::queue<Point> tilesToCheck;
+	tilesToCheck.push(startingPosition);
+	while (!tilesToCheck.empty()) {
+		Point currentPosition = tilesToCheck.front();
+		tilesToCheck.pop();
+		for (auto dir : PathDirs) {
```
ephphatha, line 1127: Had a play around with this and it turns out it's possible to drop items through stairs when diagonals are included -_-
https://user-images.githubusercontent.com/357684/147747604-624bd7d2-f4ca-4aa6-8612-665bb54ac8b9.mp4

I did a [rebase of this on current master](https://github.com/ephphatha/devilutionX/tree/qndel/fix_dropping_items) to test some changes, changing this line to only look at non-diagonal adjacent tiles avoids that issue.

```suggestion
		for (auto dir : { Direction::NorthEast, Direction::NorthWest, Direction::SouthEast, Direction::SouthWest }) {
```

ephphatha (commented):
Even though this search needs to run twice each time an item is dropped, on my machine there's no noticeable lag. It'd be interesting seeing how this performs on handhelds/low power devices. I don't think it's a significant amount of extra work/space, the checkTiles array clocks in around 12kb and expected worst case for tilesToCheck seems to be around 40 queued positions after dropping a full inventory of gold.

Assuming no other surprising corner cases pop up like the stairs thing I reckon this is good after a rebase.

Source/inv.cpp
AJenbo: Well that is actually a general issue with the stairs, you can also see monsters on the other side of it etc.

ephphatha: That's similar, in this case I'm mainly thinking of it from the way players path around solid tiles. Players can't walk diagonally between two tiles unless all four tiles that meet at that corner are non-solid (path_solid_pieces in path.cpp). This prevents players walking through stairs or between rocks.
2021-12-30T233408_devilutionx

https://user-images.githubusercontent.com/357684/147752603-113cc65a-e0e0-4973-9d14-c571ce1dcd29.mp4
 Players are allowed to walk between two monsters/objects in the same formation though (visible at the end of that capture).

Having this check progress along north/south/east/west leads to item trails through gaps the player can't path through, only checking the four suggested directions avoids this issue and doesn't significantly impact dropping items near objects/corners.

Source/inv.cpp
```diff
@@ -1109,6 +1110,37 @@ void CleanupItems(Item *item, int ii)
 	}
 }
 
+std::optional<Point> FindClosestValidItemPosition(Point startingPosition)
+{
+	if (CanPut(startingPosition)) {
+		return startingPosition;
+	}
+	if (dItem[startingPosition.x][startingPosition.y] == 0)
+		return {};
```
ephphatha, line 1119: Another thing I noticed while testing... This check stops the player dropping items inside houses (which is great :D), but also stops ctrl+click dropping items if you're standing on a tile above an object (something where `IsItemBlockingObjectAtPosition` returns true). Relatively minor issue since the player can just move to a different tile, break/open the object, or drop with the mouse targeting a valid tile.

qndel:
@Chance4us could you test again? :)

Source/inv.cpp
qndel: well for me it resulted in a significantly worse shape of the item pool

qndel:
The only issue I've noticed so far was that sometimes I got a gold pile I couldn't pick up but seemed to only happen with auric amulet (was testing in multi - 2 players, both with auric amulet) so worth checking if this doesn't also happen with current item drop algorithm.

ephphatha:
Can confirm this fixes #4124

Chance4us:
The items almost always fall on the position where you aim the mouse pointer. It feels really intuitive.

PR #3491, opened 2021-11-13
Upgrade Android builds to SDK 31

AJenbo:
@tsunamistate I think the issue you where hitting was probably with the deprecated migration code which has now been removed.

### DIFF
--- a/3rdParty/SDL2/CMakeLists.txt
+++ b/3rdParty/SDL2/CMakeLists.txt
@@ -14,8 +14,8 @@ endif()
 include(FetchContent_MakeAvailableExcludeFromAll)
 include(FetchContent)
 FetchContent_Declare(SDL2
-    URL https://github.com/libsdl-org/SDL/archive/2e9821423a237a1206e3c09020778faacfe430be.tar.gz
-    URL_HASH MD5=5311423cceaea04fa1e00ae21050b012
+    URL https://github.com/libsdl-org/SDL/archive/refs/tags/release-2.0.18.tar.gz
+    URL_HASH MD5=647e2890385e5f13b9aeee2e27fa26e7
 )
 FetchContent_MakeAvailableExcludeFromAll(SDL2)
 
--- a/android-project/app/build.gradle
+++ b/android-project/app/build.gradle
@@ -7,13 +7,13 @@ if (buildAsApplication) {
 }
 
 android {
-	compileSdkVersion 30
+	compileSdkVersion 31
 	defaultConfig {
 		if (buildAsApplication) {
 			applicationId "org.diasurgical.devilutionx"
 		}
 		minSdkVersion 18
-		targetSdkVersion 30
+		targetSdkVersion 31
 		versionCode 20
 		versionName "1.3.0"
 		externalNativeBuild {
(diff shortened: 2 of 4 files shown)


StephenCWills:
It seems that SDL requires the `BLUETOOTH_CONNECT` permission on Android 12 devices or the game will crash at startup. This permission requires a runtime request so it's not enough to simply declare the permission in `AndroidManifest.xml`. I've submitted an issue to SDL: https://github.com/libsdl-org/SDL/issues/4952.

android-project/app/src/main/res/xml/backup_descriptor.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <exclude
+        domain="sharedpref"
+        path="hidapi.xml" />
+</full-backup-content>
```
StephenCWills, line 6: This is interesting. Do you know how this backup feature works?

I read that it backs up several folders, including the external storage directory, and that backups are limited to 25 MB per user of the app. However, I don't know what happens when a backup exceeds 25 MB, as ours inevitably would even with just spawn.mpq. It would be great if save files and INI settings could be restored upon reinstall so it would be a bit disappointing if the whole backup process fails because of the MPQ files.

AJenbo: Not really, after updating to 31 Android Studio highlighted the `android:allowBackup="true"` line in `AndroidManifest.xml` and said that it could generate the now needed `backup_descriptor.xml`.

Looking at the docs the path only supports exact match, so we can't do `*.sv` :(

I guess the solution is to have it be `.` and exclude all known MPQ files.

StephenCWills: Hrm, you're right, and of course it's case-sensitive, too. It feels absurd to me that they don't support wildcards, but I guess they can justify keeping it simple because they allow overriding the Auto Backup process with a BackupAgent written in Java.

Excluding the known MPQ files should be good enough for us, though. 👍

AJenbo: I split this up in to two PRs so that we can get backup working before we do an upgrade to SDK 31

android-project/app/src/main/java/org/libsdl/app/HIDDeviceManager.java
```diff
@@ -584,7 +554,14 @@ public boolean openDevice(int deviceID) {
         if (usbDevice != null && !mUsbManager.hasPermission(usbDevice)) {
             HIDDeviceOpenPending(deviceID);
             try {
-                mUsbManager.requestPermission(usbDevice, PendingIntent.getBroadcast(mContext, 0, new Intent(HIDDeviceManager.ACTION_USB_PERMISSION), 0));
+                final int FLAG_MUTABLE = 0x02000000; // PendingIntent.FLAG_MUTABLE, but don't require SDK 31
+                int flags;
```
qndel, line 558: why not flags = 0 and get rid of else? ;d

AJenbo: I just copied the files from upstream 😜

StephenCWills:
@AJenbo I'm not sure if you were missing a few of the latest commits or what, but there were a few things missing from the Java files that I went ahead and merged in. I also updated the SDL version in the 3rdParty folder to get the latest changes there. After that, I was able to get this working on my phone again.

Also, I'm not sure what happened in SDLActivity.java. Seems like my merge made a bunch of unintended changes there so I'll need to check it again.

StephenCWills (approved):
Looks good to me, and it works on my phone. So cool to see my INI and save files come back after an uninstall.

PR #3492, opened 2021-11-13
3ds C-stick

StephenCWills:
This seems like it should enable the C-stick on New 3DS.
https://github.com/devkitPro/SDL/blob/a9a6d223b4ceb1618053f9e1e6ef5afbbf4f7335/src/joystick/n3ds/SDL_sysjoystick.c#L67-L82

I have only tested on Old 3DS so all I really know is that this seems to have no effect if you don't have a C-stick. I don't have a Circle Pad Pro so I can't test that either. Furthermore, having not tried this on New 3DS myself, I have no idea if the controls feel good or if they're even useful. It might be good if we could get some feedback before merging, or we can simply trust that this shouldn't break anything too badly.


AJenbo:
We could also call on @MrHuu or merge it to get more feedback.

StephenCWills:
It's probably a good idea to mention once again that there is still an upstream bug in the 3DS SDL audio driver that is causing race conditions in SDL_audiolib and crashing the game. Currently, it's recommended to test by [disabling Sound and Music Volume in diablo.ini](https://github.com/diasurgical/devilutionX/wiki/DevilutionX-diablo.ini-configuration-guide#audio) or by rebasing onto the 1.3 branch.

CMake/ctr/n3ds_defs.cmake
```diff
@@ -22,9 +22,11 @@ set(SDL1_VIDEO_MODE_BPP 8)
 set(DEFAULT_WIDTH 800)
 set(DEFAULT_HEIGHT 480)
 
-#SDL Joystick axis mapping (circle-pad)
+#SDL Joystick axis mapping (circle-pad/C-stick)
 set(JOY_AXIS_LEFTX 0)
 set(JOY_AXIS_LEFTY 1)
+set(JOY_AXIS_RIGHTX 3)
+set(JOY_AXIS_RIGHTY 4)
```
MrHuu, line 29: These should be changed to:
```cmake
set(JOY_AXIS_RIGHTX 2)
set(JOY_AXIS_RIGHTY 3)
```

Issue #3510, opened 2021-11-14
Item seed collisions can cause items to be picked up other than the target


ephphatha:
**Important information**
Win 11 / DevilutionX master / compiled

**Describe**
The way vanilla Diablo/Hellfire handle picking up items means they always do a search over the list of dropped items, even when the coordinates provided in the RequestGetItem message are valid. This can result in a different item getting picked up, leaving the target item on the ground and un-interactable (as it has been flagged as requested).

**To Reproduce**
This is easiest to reproduce with gold.
1. Use `give gold` to fill your inventory with gold
2. Drop it all onto the ground
3. Start picking up items
4. Pay attention to which gold pile disappears off the ground, eventually it'll be one away from where you clicked.
5. Try pick up the pile you originally clicked, the player character will path to the item but will not collect it from the ground.

**Expected behavior**
Picking up items should pick up the one that was targeted.

**Screenshots**
https://user-images.githubusercontent.com/357684/141663183-203e0766-e3e2-4194-a812-73b6c2296afd.mp4
Keep an eye on item 105 (the gold pile left of the rock) and item 60 (the gold pile right of the rock).

**Additional context**
This is a combination of issues. Item seeds for gold are frequently rerolled which with the dodgy default RNG means collisions are more likely, and the logic for picking up an item ignores the position data that gets sent with the message.

AJenbo:
This has also caused other issues like quest items and "unique" items not reacting to being picked up.

PR #3513, opened 2021-11-14
Check the item at the target position before searching all items

ephphatha:
Quick and dirty fix for #3510. I was gonna look at this more if #2932 got merged as this is a bit of an odd use case of the dItems array, this PR is not as clean as it could be.

### DIFF
--- a/Source/items.h
+++ b/Source/items.h
@@ -367,6 +367,11 @@ struct Item {
 		return IsScroll() && _iSpell == spellId;
 	}
 
+	[[nodiscard]] bool KeyAttributesMatch(int32_t seed, _item_indexes itemIndex, uint16_t createInfo)
+	{
+		return _iSeed == seed && IDidx == itemIndex && _iCreateInfo == createInfo;
+	}
+
 	UiFlags getTextColor() const
 	{
 		switch (_iMagical) {
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -707,7 +707,18 @@ DWORD OnRequestGetItem(const TCmd *pCmd, Player &player)
 	if (gbBufferMsgs != 1 && IOwnLevel(player.plrlevel) && IsGItemValid(message)) {
 		const Point position { message.x, message.y };
 		if (GetItemRecord(message.dwSeed, message.wCI, message.wIndx)) {
-			int ii = FindGetItem(message.wIndx, message.wCI, message.dwSeed);
+			int ii = -1;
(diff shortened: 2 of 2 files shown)


Source/items.h
```diff
@@ -367,6 +367,11 @@ struct Item {
 		return IsScroll() && _iSpell == spellId;
 	}
 
+	[[nodiscard]] bool KeyAttributesMatch(int32_t seed, _item_indexes itemIndex, uint16_t createInfo)
```
AJenbo, line 370: Could we also use this in FindGetItem?

Source/msg.cpp
```diff
@@ -707,7 +707,18 @@ DWORD OnRequestGetItem(const TCmd *pCmd, Player &player)
...
+			int ii = -1;
+			if (InDungeonBounds(position)) {
+				ii = abs(dItem[position.x][position.y]) - 1;
+				if (ii >= 0 && !Items[ii].KeyAttributesMatch(message.dwSeed, static_cast<_item_indexes>(message.wIndx), message.wCI)) {
+					ii = -1;
+				}
+			}
+
+			if (ii == -1) { // No item at the target position or the key attributes don't match, so try find a matching item.
+				ii = FindGetItem(message.wIndx, message.wCI, message.dwSeed);
```
AJenbo, line 719: Is there a valid case where this could happen?

Source/items.h
ephphatha: Yes, also in SyncGetItem and DeltaAddItem. I'd been avoiding updating those call sites because only DeltaAddItem has a "valid" reason to use it and it's to do an equality check, which should use a method wrapping this check and other conditions.

Source/msg.cpp
ephphatha: In the current code, its more likely if a remote player in a multiplayer game is dropping items on a different floor to the local player due to the way item deltas are recorded and processed. #2932 updates the code that "plays back" the history of actions to handle dropped items a bit better which should hopefully make this check unnecessary.

Potentially could still be an issue with lag though, but in that case it might be safe to just fail to pick up an item and let the game stay desynced.

AJenbo (approved):
A bit dirty, but solves some issues. Ok for now, hope we don't forget to clean it up later.

Source/msg.cpp
AJenbo: Damn you have so many clean ups going, I'm having a hard time keeping up :D

PR #3518, opened 2021-11-14
Implement virtual gamepad activation/deactivation

### LINKED ISSUE #3375: [touch] joystick can get stuck when switching to a menu
> However, I have met a bug that if you were dead when pushing the direction panel, the menu will automatically moving so that the user can’t push any key.

I have experienced this my self on one occasion, i was holding the joystick in a direction and entered a menu where the joystick isn't active, the cursor would then constantly move up making it impossible to navigate the menu.

StephenCWills:
This PR explicitly activates and deactivates the virtual gamepad in the same functions that load and unload the gamepad art. The `isActive` Boolean value is used to determine whether the game should read virtual gamepad state. This prevents it from interfering in UI menus where you cannot see or interact with the gamepad. Deactivation also resets the state of the d-pad and buttons so they don't retain invalid state upon reactivation.

This resolves #3375

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,6 +239,10 @@ if(PIE)
   set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 endif()
 
+if(USE_SDL1)
+  set(VIRTUAL_GAMEPAD OFF)
+endif()
+
 if(NONET)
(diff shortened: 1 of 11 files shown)


AJenbo (approved):
Sounds like a good way to make sure the visual and logic state matches each other.

Also nice clean up of the flags.

PR #3544, opened 2021-11-19
Increased stack size to 128KB

aviy15:


Source/platform/ctr/system.cpp
```diff
@@ -9,11 +9,9 @@
 
 using namespace devilution;
 
-// Increase stack size for recursion in FindTransparencyValues()
-// 128 KB supports around 500 levels of recursion
-// Default stack size on 3DS is only 32 KB
+// Increased stack size to 128KB to allow higher levels of recursion and avoid game crashes
 extern "C" {
-u32 __stacksize__ = 128 * 1024;
+u32 __stacksize__ = 4 * 128 * 1024;
```
StephenCWills, line 14:
```suggestion
u32 __stacksize__ = 512 * 1024;
```

Source/platform/ctr/system.cpp
```diff
@@ -9,11 +9,9 @@
 
 using namespace devilution;
 
-// Increase stack size for recursion in FindTransparencyValues()
-// 128 KB supports around 500 levels of recursion
-// Default stack size on 3DS is only 32 KB
+// Increased stack size to 128KB to allow higher levels of recursion and avoid game crashes
```
StephenCWills, line 12: What is the purpose of removing the extra details in the comment? Also, you have increased the stack size to 512 KB, not 128 KB.

AJenbo:
`8 * 128 * 1024` = 1MiB

Why? The comment is wrong, and it would probably be better written as `1024 * 1024`.

StephenCWills (changes requested):
Can you provide some context for why you made this change? Have you tested this?

qndel:
I love PRs from new guys with 0 explanation lol ...

PR #3554, opened 2021-11-20
Increase some DiabloUI buffer sizes

### LINKED ISSUE #3551: Crashes when selecting Hell difficulty level.
**Important information**
Operating system: macOS 12.0.1（21A559）
DevilutionX version : #3550 self compile

**Describe**
I don't know if this is the right place to write this, but recently I've been getting frequent crashes when running in Japanese.

When you hover the cursor over Difficulty Hell.
When you choose to delete a character.
etc...

**To Reproduce**
1. diablo.ini  Code=ja
2. Select 'Single Player'
3. Select 'New Hero'
4. Select Any Class
5. Enter Name
6. Select Difficulty Hell
7. Crash!

**Additional context**
The location is the same, the DecodeFirstUtf8CodePoint function.
Log file is attached.

[CrashLog_Mac_Japanese.txt](https://github.com/diasurgical/devilutionX/files/7573992/CrashLog_Mac_Japanese.txt)

glebm:
Fixes #3551

### DIFF
--- a/Source/DiabloUI/selgame.cpp
+++ b/Source/DiabloUI/selgame.cpp
@@ -18,7 +18,7 @@ namespace devilution {
 char selgame_Label[32];
 char selgame_Ip[129] = "";
 char selgame_Password[16] = "";
-char selgame_Description[256];
+char selgame_Description[512];
 bool selgame_enteringGame;
 int selgame_selectedGame;
 bool selgame_endMenu;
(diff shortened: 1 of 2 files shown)


bubio:
https://github.com/diasurgical/devilutionX/issues/3551#issuecomment-974645031

AJenbo (approved):
This makes sens.

Previously Bulgarian has been one of the languages that uses most bytes per string. Owing to a combination of most symbols being Cyclic (2 bytes per symbol), and being a very verbose language.

Japanese uses 3 bytes per symbol, so despite it not being particularly verbose it quickly grow in byte size.

Chinese and Korean also use 3 bytes per symbol, but are pretty condense. So that is why they haven't been triggering these issues.

I'll run threw the game a few times with the Chinese translation enabled to see if I can trigger any more issues.

PR #3566, opened 2021-11-21
Clean up item name generation

### LINKED ISSUE #2071: Korean translation

AJenbo:
Resolves #2071

This allows translators to grater control over how item names are generated. It should also make the code easier to read and simpler to hook in to once/if we add scripting to the name generation.

@nsm53project has mentioned this being required for Korean to generate proper names since they need it to be `{prefix}{suffix}{item}`.

Most of the changes is just syncing the translation files ;) Note that Russian and Ukrainian have been left out as they are currently being edited in other PRs.

### DIFF
--- a/Source/DiabloUI/selhero.cpp
+++ b/Source/DiabloUI/selhero.cpp
@@ -16,6 +16,7 @@
 #include "options.h"
 #include "pfile.h"
 #include "utils/language.h"
+#include "utils/utf8.hpp"
 #include <menu.h>
 
 namespace devilution {
@@ -138,12 +139,12 @@ void SelheroListFocus(int value)
 	}
 
 	SELHERO_DIALOG_HERO_IMG->m_frame = static_cast<int>(enum_size<HeroClass>::value);
-	strncpy(textStats[0], "--", sizeof(textStats[0]) - 1);
-	strncpy(textStats[1], "--", sizeof(textStats[1]) - 1);
-	strncpy(textStats[2], "--", sizeof(textStats[2]) - 1);
-	strncpy(textStats[3], "--", sizeof(textStats[3]) - 1);
-	strncpy(textStats[4], "--", sizeof(textStats[4]) - 1);
-	strncpy(textStats[5], "--", sizeof(textStats[5]) - 1);
+	CopyUtf8(textStats[0], "--", sizeof(textStats[0]));
+	CopyUtf8(textStats[1], "--", sizeof(textStats[1]));
+	CopyUtf8(textStats[2], "--", sizeof(textStats[2]));
+	CopyUtf8(textStats[3], "--", sizeof(textStats[3]));
+	CopyUtf8(textStats[4], "--", sizeof(textStats[4]));
+	CopyUtf8(textStats[5], "--", sizeof(textStats[5]));
 	SELLIST_DIALOG_DELETE_BUTTON->m_iFlags = baseFlags | UiFlags::ColorUiSilver | UiFlags::ElementDisabled;
 	selhero_deleteEnabled = false;
 }
@@ -269,7 +270,7 @@ void SelheroClassSelectorSelect(int value)
 	title = selhero_isMultiPlayer ? _("New Multi Player Hero") : _("New Single Player Hero");
 	memset(selhero_heroInfo.name, '\0', sizeof(selhero_heroInfo.name));
 	if (ShouldPrefillHeroName())
-		strncpy(selhero_heroInfo.name, SelheroGenerateName(selhero_heroInfo.heroclass), sizeof(selhero_heroInfo.name) - 1);
+		CopyUtf8(selhero_heroInfo.name, SelheroGenerateName(selhero_heroInfo.heroclass), sizeof(selhero_heroInfo.name));
 	vecSelDlgItems.clear();
 	SDL_Rect rect1 = { (Sint16)(PANEL_LEFT + 264), (Sint16)(UI_OFFSET_Y + 211), 320, 33 };
 	vecSelDlgItems.push_back(std::make_unique<UiArtText>(_("Enter Name"), rect1, UiFlags::AlignCenter | UiFlags::FontSize30 | UiFlags::ColorUiSilver, 3));
@@ -596,11 +597,11 @@ static void UiSelHeroDialog(
 			char dialogTitle[128];
 			char dialogText[256];
 			if (selhero_isMultiPlayer) {
-				strncpy(dialogTitle, _("Delete Multi Player Hero"), sizeof(dialogTitle) - 1);
+				CopyUtf8(dialogTitle, _("Delete Multi Player Hero"), sizeof(dialogTitle));
 			} else {
-				strncpy(dialogTitle, _("Delete Single Player Hero"), sizeof(dialogTitle) - 1);
+				CopyUtf8(dialogTitle, _("Delete Single Player Hero"), sizeof(dialogTitle));
 			}
-			strncpy(dialogText, fmt::format(_("Are you sure you want to delete the character \"{:s}\"?"), selhero_heroInfo.name).c_str(), sizeof(dialogText));
+			CopyUtf8(dialogText, fmt::format(_("Are you sure you want to delete the character \"{:s}\"?"), selhero_heroInfo.name), sizeof(dialogText));
 
 			if (UiSelHeroYesNoDialog(dialogTitle, dialogText))
 				fnremove(&selhero_heroInfo);
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -67,6 +67,7 @@
 #include "utils/console.h"
 #include "utils/language.h"
 #include "utils/paths.h"
+#include "utils/utf8.hpp"
 
 #ifdef __vita__
 #include "platform/vita/touch.h"
@@ -897,7 +898,7 @@ void DiabloInitScreen()
 void SetApplicationVersions()
 {
 	snprintf(gszProductName, sizeof(gszProductName) / sizeof(char), "%s v%s", PROJECT_NAME, PROJECT_VERSION);
-	strncpy(gszVersionNumber, fmt::format(_("version {:s}"), PROJECT_VERSION).c_str(), sizeof(gszVersionNumber) / sizeof(char));
+	CopyUtf8(gszVersionNumber, fmt::format(_("version {:s}"), PROJECT_VERSION), sizeof(gszVersionNumber) / sizeof(char));
 }
 
 void DiabloInit()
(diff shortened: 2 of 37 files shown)


Source/items.cpp
```diff
@@ -1171,27 +1171,61 @@ void GetStaffPower(Item &item, int lvl, int bs, bool onlygood)
...
+
 	CalcItemValue(item);
 }
 
+namespace {
+
+std::string GenerateMagicItemName(const string_view &baseNamel, int preidx, int sufidx)
+{
+	if (preidx != -1 && sufidx != -1) {
+		const char *fmt = _(/* TRANSLATORS: Constructs item names. Format: {Prefix} {Item} of {Spell}. Example: King's Long Sword of the Whale */ "{0} {1} of {2}");
```
ephphatha, line 1214:
```suggestion
		const char *fmt = _(/* TRANSLATORS: Constructs item names. Format: {Prefix} {Item} of {Suffix}. Example: King's Long Sword of the Whale */ "{0} {1} of {2}");
```
Guessing this was copied from the staff suffix TL note.

AJenbo: Thanks :D

Source/items.cpp
```diff
@@ -1171,27 +1171,61 @@ void GetStaffPower(Item &item, int lvl, int bs, bool onlygood)
...
+std::string GenerateMagicItemName(const string_view &baseNamel, int preidx, int sufidx)
+{
+	if (preidx != -1 && sufidx != -1) {
+		const char *fmt = _(/* TRANSLATORS: Constructs item names. Format: {Prefix} {Item} of {Suffix}. Example: King's Long Sword of the Whale */ "{0} {1} of {2}");
+		return fmt::format(fmt, _(ItemPrefixes[preidx].PLName), baseNamel, _(ItemSuffixes[sufidx].PLName));
+	} else if (preidx != -1) {
+		const char *fmt = _(/* TRANSLATORS: Constructs item names. Format: {Prefix} {Item}. Example: King's Long Sword */ "{0} {1}");
+		return fmt::format(fmt, _(ItemPrefixes[preidx].PLName), baseNamel);
+	} else if (sufidx != -1) {
+		const char *fmt = _(/* TRANSLATORS: Constructs item names. Format: {Item} of {Spell}. Example: Long Sword of the Whale */ "{0} of {1}");
```
ephphatha, line 1220:
```suggestion
		const char *fmt = _(/* TRANSLATORS: Constructs item names. Format: {Item} of {Suffix}. Example: Long Sword of the Whale */ "{0} of {1}");
```

AJenbo: ... bed time

qndel:
With this PR I can't even enter character selection menu :/

ephphatha:
pretty much every use of strncpy needs to be made aware of multibyte characters so it doesn't split a character. This appears to work, but I'm not 100% sure the logic is correct (and whether we need to call `setlocale` with a utf8 locale code...)

```cpp
void mbstrncpy(char *dest, const string_view source, size_t bytes)
{
	size_t len = 0;
	while (true) {
		int charLength = mblen(source.data() + len, bytes - len);
		if (charLength <= 0) {
			break;
		}
		if (len + charLength >= bytes) {
			break;
		}
		len += charLength;
	}
	strncpy(dest, source.data(), std::min(len, bytes - 1));
	dest[bytes] = '\0';
}

// used similar to strncpy_s, no need for the caller to append the null terminator
mbstrncpy(item._iName, u8"地 回 器 在 生 堂 地 回 器 在 生 堂 地 回器在生", sizeof(item._iName));
```

qndel:
[single_30.zip](https://github.com/diasurgical/devilutionX/files/7576146/single_30.zip)
save that crashes my char list

glebm:
We could have a simple function like this one:

```c++
string_view Utf8Truncate(string_view str, std::size_t len)
{
	if (str.size() > len) {
		str.remove_suffix(str.size() - len);
		while (!str.empty() && !IsTrailUtf8CodeUnit(str.back())) str.remove_suffix(1);
	}
	return str;
}
```

glebm:
It can then be used like this:

```c++
void mbstrncpy(char *dest, string_view source, size_t bytes)
{
  source = TruncateUtf8(source, bytes - 1);
  strcpy(dest, source.data(), source.size());
  dest[source.size()] = '\0';
}
```

Source/items.cpp
```diff
@@ -1171,27 +1171,61 @@ void GetStaffPower(Item &item, int lvl, int bs, bool onlygood)
...
 			SaveItemAffix(item, ItemPrefixes[preidx]);
 			item._iPrePower = ItemPrefixes[preidx].power.type;
 		}
 	}
-	if (!StringInPanel(item._iIName)) {
-		strcpy(item._iIName, _(AllItemsList[item.IDidx].iSName));
-		if (preidx != -1) {
-			strncpy(item._iIName, fmt::format(/* TRANSLATORS: Constructs item names. Format: {Prefix} {Item}. Example: King's Sword */ _("{0} {1}"), _(ItemPrefixes[preidx].PLName), item._iIName).c_str(), sizeof(item._iIName) - 1);
+
+	const char *baseName = _(AllItemsList[item.IDidx].iName);
```
glebm, line 1180: Unrelated but perhaps we should have a function that returns a `const std::string &` instead of a `const char*`.
Using `const char *` means lengths are computed multiple times.

AJenbo: Yeah, was disappointed to find this out when I started compiling (had written it assuming `std::string` at first). But I think this should be part of another clean up where we start migrating things to `std::string`/`string_view` in general.

AJenbo:
> [single_30.zip](https://github.com/diasurgical/devilutionX/files/7576146/single_30.zip) save that crashes my char list

Check if it works now :)

image

AJenbo:
@ephphatha @glebm I took a shot a making use of UTF8 aware `strncpy` across code base, let me know what you think.

qndel:
> > [single_30.zip](https://github.com/diasurgical/devilutionX/files/7576146/single_30.zip) save that crashes my char list
> 
> Check if it works now :)
> 
> image

Works for me now.

glebm:
Looks good, though I think a lot of these should just crash instead of truncating.
Otherwise there is a good chance we won't notice that a buffer is too small.

Source/DiabloUI/selhero.cpp
```diff
@@ -138,12 +139,12 @@ void SelheroListFocus(int value)
...
-	strncpy(textStats[2], "--", sizeof(textStats[2]) - 1);
-	strncpy(textStats[3], "--", sizeof(textStats[3]) - 1);
-	strncpy(textStats[4], "--", sizeof(textStats[4]) - 1);
-	strncpy(textStats[5], "--", sizeof(textStats[5]) - 1);
+	CopyUtf8(textStats[0], "--", sizeof(*textStats));
+	CopyUtf8(textStats[1], "--", sizeof(*textStats));
+	CopyUtf8(textStats[2], "--", sizeof(*textStats));
+	CopyUtf8(textStats[3], "--", sizeof(*textStats));
+	CopyUtf8(textStats[4], "--", sizeof(*textStats));
+	CopyUtf8(textStats[5], "--", sizeof(*textStats));
```
ephphatha, line 147:
```suggestion
	CopyUtf8(textStats[0], "--", sizeof(textStats[0]));
	CopyUtf8(textStats[1], "--", sizeof(textStats[1]));
	CopyUtf8(textStats[2], "--", sizeof(textStats[2]));
	CopyUtf8(textStats[3], "--", sizeof(textStats[3]));
	CopyUtf8(textStats[4], "--", sizeof(textStats[4]));
	CopyUtf8(textStats[5], "--", sizeof(textStats[5]));
```
Might as well keep the specific array references even though this is copying a static/non-translated string into a buffer we know is large enough.

Source/diablo.cpp
```diff
@@ -1447,11 +1448,10 @@ void InitKeymapActions()
...
 		    };
-		    strncpy(pszStr, fmt::format(_(/* TRANSLATORS: {:s} means: Character Name, Game Version, Game Difficulty. */
-		                                    "{:s}, version = {:s}, mode = {:s}"),
-		                        PROJECT_NAME, PROJECT_VERSION, difficulties[sgGameInitInfo.nDifficulty])
-		                        .c_str(),
-		        MAX_SEND_STR_LEN - 1);
+		    CopyUtf8(pszStr, fmt::format(_(/* TRANSLATORS: {:s} means: Character Name, Game Version, Game Difficulty. */
+		                                     "{:s}, version = {:s}, mode = {:s}"),
+		                         PROJECT_NAME, PROJECT_VERSION, difficulties[sgGameInitInfo.nDifficulty]),
+		        MAX_SEND_STR_LEN);
```
ephphatha, line 1454:
```suggestion
		        sizeof(pszStr));
```

Source/pack.cpp
```diff
@@ -222,8 +223,7 @@ bool UnPackPlayer(const PlayerPack *pPack, Player &player, bool netSync)
 	ClrPlrPath(player);
 	player.destAction = ACTION_NONE;
 
-	strncpy(player._pName, pPack->pName, PLR_NAME_LEN - 1);
-	player._pName[PLR_NAME_LEN - 1] = '\0';
+	CopyUtf8(player._pName, pPack->pName, PLR_NAME_LEN);
```
ephphatha, line 226:
```suggestion
	CopyUtf8(player._pName, pPack->pName, sizeof(player._pName));
```

Source/pfile.cpp
```diff
@@ -337,13 +337,11 @@ bool pfile_ui_save_create(_uiheroinfo *heroinfo)
 	giNumberOfLevels = gbIsHellfire ? 25 : 17;
 
 	archive.RemoveHashEntries(GetFileName);
-	strncpy(hero_names[saveNum], heroinfo->name, PLR_NAME_LEN);
-	hero_names[saveNum][PLR_NAME_LEN - 1] = '\0';
+	CopyUtf8(hero_names[saveNum], heroinfo->name, PLR_NAME_LEN);
```
ephphatha, line 340: Looks like it's valid to call sizeof with a runtime argument in this case (guessing since both array dimensions are known), using the destination buffer size directly seems appropriate.
```suggestion
	CopyUtf8(hero_names[saveNum], heroinfo->name, sizeof(hero_names[saveNum]));
```

Source/utils/utf8.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <cstring>
```
ephphatha, line 4: This is only for the `size_t` typedef? `<cstddef>` carries less baggage given most of these functions deal with `string` derivatives.

ephphatha:
I should've put those in a single review sorry, didn't realise github would make it look so garbage doing single comments. Most of those are nits to try minimise the use of defines so aren't blockers. I haven't tested it but the implementation looks good.

Source/utils/utf8.hpp
AJenbo: `std::nemcpy`

AJenbo:
> I should've put those in a single review sorry, didn't realise github would make it look so garbage doing single comments. Most of those are nits to try minimise the use of defines so aren't blockers. I haven't tested it but the implementation looks good.

My notification settings makes it come out pretty much the same so don't worry about it :)

Source/utils/utf8.hpp
AJenbo: ... but i moved that to the cpp file

PR #3568, opened 2021-11-21
Return Utf8DecodeError when passed an incomplete code point

ephphatha:
Also ensure that len is initialised and set properly, otherwise it leads to a call to remove_prefix with garbage.

It doesn't seem useful or valid to return an incomplete code point from this function. To match the behaviour of `mblen` we could have an "incomplete code point" return value, but this at least handles truncated strings better than before.

### DIFF
--- a/Source/utils/utf8.cpp
+++ b/Source/utils/utf8.cpp
@@ -21,7 +21,8 @@ char32_t DecodeFirstUtf8CodePoint(string_view input, uint8_t *len)
 			return Utf8DecodeError;
 		}
 	}
-	return codepoint;
+	*len = input.size();
+	return Utf8DecodeError;
 }
 
 } // namespace devilution


Source/utils/utf8.hpp
```diff
@@ -23,7 +23,7 @@ char32_t DecodeFirstUtf8CodePoint(string_view input, uint8_t *len);
  */
 inline char32_t ConsumeFirstUtf8CodePoint(string_view *input)
 {
-	uint8_t len;
+	uint8_t len = 0;
```
glebm, line 26: `len` is now always set by `DecodeFirstUtf8CodePoint`, so perhaps the initialization here isn't needed?

ephphatha: Yeah, clang wasn't warning about this before presumably because the function took a pointer. This was a first attempt while I was working out what the actual cause of the failure was. Of course now clang isn't warning about unused intialisation value... So not sure whether it's safer to leave it or remove it.

glebm: Well, we shouldn't initialize it unless necessary, and it doesn't seem necessary!

glebm (approved):

Issue #3546, opened 2021-11-20
RFC: Embedding color switching in text


AJenbo:
This is currently already done to some extend for the help menu where $ means the line should be rendered red.

We currently support 5 colors, white, red, blue, gold and black. Being able to color a single line in multiple colors can have several benefits:

- It would allow a better looking options menu
- It can be helpful for gamepad hints (ⓍⓎⒶⒷ)
- Some people enjoy having lots of colors in there names or chat messages

I suggest that we the text render look for ⬜🟥🟦🟨⬛, and instead of rendering it, switch the current color when encountered so that the remainder of the text will be rendered with the given color rather then what was given via flags. 

Example:
`DrawString(_("Press 🟦Ⓧ⬜ to cast spell"))`
`DrawString(_("Run in town: ⬜On"), UiFlags::ColorGold)`

The potential down side is that you would not be able to use the 5 symbols. An alternative would be to only do it if given in a combination with a control character like U+2061 (FUNCTION APPLICATION) or U+91 (PRIVATE USE ONE). The down side being that they are harder to enter in to the text.

AJenbo:
@qndel @glebm @obligaron please comment on this proposal :)

qndel:
Yeah sounds good

glebm:
Perhaps we should handle this in the code instead, e.g.:

```c++
DrawStringWithColors(_("Press {} to cast spell"), {"Ⓧ", UiFlags::ColorBlue});
DrawStringWithColors(_("Run in town: {}"), {_("On"), UiFlags::ColorWhite}, UiFlags::ColorGold));
```
We probably only need to support `{}` at least initially.

AJenbo:
Sure if you can figure out how to do that in a good way :)

glebm:
I think I have :tada:  #3576

AJenbo:
> Some people enjoy having lots of colors in there names or chat messages

To bad it doesn't allow for the user text part, oh well.

glebm:
Mixing translations with formatting is what I wanted to avoid with the submitted PR.

Markup for user input can be done as a separate feature!

PR #3576, opened 2021-11-22
Add `DrawStringWithColors`

glebm:
A way to color parts of the string differently while keeping the color information outside of the string contents.

This is an alternative to #3546.

Example:

```c++
DrawStringWithColors(out, "Test {1} {0}", {{"blue", UiFlags::ColorBlue}, {"red", UiFlags::ColorRed}}, ...);
```

image


glebm:
Added support for integers

AJenbo (approved):

PR #3578, opened 2021-11-22
[3DS] Increase buffer size for libmpq file pointers

StephenCWills:
This change significantly improves load times for MPQ assets on my o3DS console.


glebm:
Great work! :tada: 

Note that on (most?) other platforms this is not necessary because the default buffer size is already equal to the partition block size. It appears that the 3ds uses a non-conformant libc that defaults to unbuffered I/O for files.

glebm (approved):

AJenbo:
> This change significantly improves load times for MPQ assets on my o3DS console.

Do you have numbers :)

StephenCWills:
> Do you have numbers :)

The difference is so staggeringly significant that I didn't really bother taking any hard before/after measurements. All I know is that before this change, loading time for the intro video was a whopping 60 seconds. After this change, it's maybe around 2-3 seconds.

But it's not just the intro video. All loading screens are noticeably faster with this change.

AJenbo:
cool i was just wondering if it was 60 vs 45, or some thing more appropriate :)

PR #3581, opened 2021-11-23
Update Korean Translation

nsm53project:
Sync translation files
Weapons


nsm53project:
image

There are some problems caused by the difference between Korean and English, but {prefix}{suffix}{item} is very good :D

AJenbo (approved):

Issue #3412, opened 2021-11-07
Building without network access


jpalus:
For distribution packaged software it is desirable to do compilation in an isolated environment without network access (reproducible builds, no risk of downloading malicious code etc). In latest devilutionx version (1.3.0) that is no longer possible since ie asio must be downloaded. Therefore would it be possible to provide "full" source tarball as a part of release including all dependencies that are otherwise downloaded?

AJenbo:
The needed dependencies depend on the target platform. If this is desirable for you I suggest having the system that downloads the code in the first place also download the dependencies before pushing it to your build server.

StephenCWills:
> In latest devilutionx version (1.3.0) that is no longer possible since ie asio must be downloaded.

FYI, it's not necessary to download ASIO during the build process.
`cmake -DFETCHCONTENT_SOURCE_DIR_ASIO=/path/to/asio ...`

glebm:
First, set this option:

```
-DFETCHCONTENT_FULLY_DISCONNECTED=ON
```

This will make sure none of the FetchContent calls do any network access.

You can then specify every dependency similar to how it is done here:

https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/ports/devilutionx/devilutionx.mk

jpalus:
While I'm interested in particular platform it does not change the fact that I believe all sources which cannot be satisfied by system library should be included in source release. So I wouldn't mind including SDL2 in a source tarball because Android needs it. Not up to me though. Thanks for the tips with FETCHCONTENT_* vars -- I will start maintaining dependencies manually.

AJenbo:
Including them in git will affect performance during development, if you develop a script for generating a full tarball you are welcome to open a PR with it. It's worth noting that we do checksum validation on dependencies so reproducible builds are assured as the build will fail if they change.

glebm:
@jpalus DevilutionX now provides a full source tarball that includes all fetchcontent deps and devilutionx.mpq.
This makes packaging it a lot easier, e.g. https://github.com/batocera-linux/batocera.linux/pull/5146

PR #3386, opened 2021-11-04
Add CMake option to use system SDL_image

StephenCWills:
* Adds `DEVILUTIONX_SYSTEM_SDL_IMAGE` to specify whether to search for a system library
* Adds `sdl2-image` dependency to vcpkg for Windows builds
* Adds `FindSDL2_image.cmake` to utilize `PkgConfig` for locating libSDL2_image on Linux

Works for Linux SDL1/SDL2, Windows SDL2, and possibly a few other platforms.

This resolves #3384

PR #3589, opened 2021-11-24
A tool to build a self-contained source tarball

glebm:
The does some pruning of the dependencies' files to avoid a 100 MiB+ tarball.
Even with that, the final tarball size is 17 MiB with xz, 25 MiB with gzip.

Building from the resulting tarball does not require `smpq` (`devilutionx.mpq` is included).
The version number is also set in the tarball itself instead of relying on git (which does not work with tarballs).

See the file documentation at the top of `tools/make_src_dist.py` for more information.

This tarball will be produced on release via GitHub CI.
The GitHub workflow has been tested using https://github.com/nektos/act:

```shell
act -W .github/workflows/dist_src_release.yml -j make_src_dist
```

Refs #3412 /cc @jpalus


.github/workflows/dist_src_release.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    - name: Build
+      working-directory: ${{github.workspace}}
+      run: tools/make_src_dist.py
+
+    - name: Upload-Package
+      if: ${{ !env.ACT }}
+      uses: actions/upload-artifact@v2
+      with:
+        name: devilutionx.tar.xz
```
AJenbo, line 30:
```suggestion
        name: devilutionx-src.tar.xz
```

AJenbo (commented):
I wonder how stable tis will be and how generic the tarball is (will it build on android and windows?)

glebm:
> will it build on android and windows?

Good point. I've removed `FETCHCONTENT_FULLY_DISCONNECTED`, so now this tarball will build on Android and Windows (although it will fetch some dependencies there).

As to how generic it is, this tarball should work everywhere except Android and Windows.

Stability-wise, if we change the dependencies that must be vendored, we'll have to update the Python script.

AJenbo (approved):
I'm fine having this as long as it is maintained :)

glebm:
@AJenbo I maintain packages for 2 Linux distributions so no worries there, looking forward to simplifying those packages for the next release

jpalus:
@glebm Thanks a lot for driving this! It really makes build process way more straightforward. Just a few minor nits from my side after using first such dist from 1.3.0 release:
1. It would nice if the source distribution followed standard practice of putting sources into single top level directory `<name>-<version>`
2. Related to above it would nice if source distribution would be versioned so ie `devilution-src-1.3.0.tar.xz` (might be worth considering for other artifacts too btw)
3. Note that source distribution attached to 1.3.0 release is not actually created from 1.3.0 tag as it seems to include #3386

AJenbo:
2. this would prevent `latest` links on github from working

AJenbo:
@Vitexus @muzena you might find this relevant for the build systems you worked on

muzena:
@glebm 
I tried to generate tar archive with command:

`sudo act -W .github/workflows/src_dist_release.yml -j make_src_dist`

In terminal output it seems all good, in the end it says that tar archive is generated in path:

`/home/mint/aaaaaaaa/devilutionX/build-src-dist/devilutionx-1.3.0-ac152589.tar.xz`

But there's no exist  "build-src-dist" dir with devilutionx-1.3.0-ac152589.tar.xz archive.

glebm:
@muzena That directory is in the Docker container, not on your local machine.

You can just run `tools/make_src_dist.py`

muzena:
@glebm 
Is all dependencies are included or it still need to download some packages?
Like fmt, see [log file](https://launchpadlibrarian.net/571567061/buildlog_ubuntu-focal-amd64.devilutionx_1.3.0~focal_BUILDING.txt.gz)

glebm:
@muzena All dependencies *that must be vendored* are included. For fmt you should just use the existing package

AJenbo:
`sudo apt install libfmt-dev`

Vitexus:
> `sudo apt install libfmt-dev`

https://github.com/Vitexus/devilutionX/blob/98aaec15f84f63ec9a8b2a81383d48baff934aaf/debian/control#L5

PR #3600, opened 2021-11-24
Update Japanese Translations(Only Segmentation)

bubio:
- Apply segment_en.py

It's easy to forget this...

You might want to incorporate that into your build process.


glebm:
> You might want to incorporate that into your build process.

Well I'd rather we don't bring the dependency on the segmenters into the build process. While a locale is being translated, re-segmentation is frequent, but after that we won't need to do it that often. We will run the segmenter before release as well.

bubio:
Thanks for the reply.
I understand!

PR #3605, opened 2021-11-24
Clean up SpawnWitch

AJenbo:
This should hopefully make it more clear what this function is doing.

Instead of several small loops (some unrolled) that works on part of the `witchitem[]` one big loop that iterates over all items in now used instead. And each body is now an if-block that checks to see if the loop is in the range that it acts on.

There are 2 functional changes to the function:
- The 3 reserved items are now assigned a seed value. This is to simplify how the store works (instead of assigning it on purchase).
- `itemCount` in Diablo mode was 5 higher then it should have been since we added Hellfire.

Note: There might also have been a bug in vanilla where only 18 out of the 20 witch items would be used, or it was lazily reduced to 18 on purpose with out resizing the array.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4540,7 +4540,6 @@ void SpawnWitch(int lvl)
 			if (i < PinnedItemCount + MaxPinnedBookCount && bookCount < pinnedBookCount) {
 				int bookType = PinnedBookTypes[i - PinnedItemCount];
 				if (lvl >= AllItemsList[bookType].iMinMLvl) {
-
 					item._iSeed = AdvanceRndSeed();
 					SetRndSeed(item._iSeed);
 					GetItemAttrs(item, bookType, lvl);
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4514,77 +4514,71 @@ void SpawnPremium(int pnum)
 void SpawnWitch(int lvl)
 {
 	constexpr int PinnedItemCount = 3;
+	constexpr std::array<int, PinnedItemCount> PinnedItemTypes = { IDI_MANA, IDI_FULLMANA, IDI_PORTAL };
+	constexpr int MaxPinnedBookCount = 4;
+	constexpr std::array<int, MaxPinnedBookCount> PinnedBookTypes = { 114, 115, 116, 117 };
 
-	int j = PinnedItemCount;
+	int bookCount = 0;
+	const int pinnedBookCount = gbIsHellfire ? GenerateRnd(MaxPinnedBookCount) : 0;
+	const int reservedItems = gbIsHellfire ? 10 : 17;
+	const int itemCount = GenerateRnd(WITCH_ITEMS - reservedItems) + 10;
+	const int maxValue = gbIsHellfire ? 200000 : 140000;
 
-	witchitem[0] = {};
-	GetItemAttrs(witchitem[0], IDI_MANA, 1);
-	witchitem[0]._iCreateInfo = lvl;
-	witchitem[0]._iStatFlag = true;
-	witchitem[1] = {};
-	GetItemAttrs(witchitem[1], IDI_FULLMANA, 1);
-	witchitem[1]._iCreateInfo = lvl;
-	witchitem[1]._iStatFlag = true;
-	witchitem[2] = {};
-	GetItemAttrs(witchitem[2], IDI_PORTAL, 1);
-	witchitem[2]._iCreateInfo = lvl;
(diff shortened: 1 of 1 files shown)


Source/items.cpp
```diff
@@ -529,17 +529,17 @@ void CalcPlrItemMin(Player &player)
 	}
 }
 
-void WitchBookLevel(int ii)
+void WitchBookLevel(Item &item)
 {
-	if (witchitem[ii]._iMiscId != IMISC_BOOK)
+	if (item._iMiscId != IMISC_BOOK)
```
ephphatha, line 534: Given we bail out of the function if the item passed in is not a book we can use a more descriptive variable name. I went with `bookItem` in #2924 but `book` would be fine, no need to be repetitive :D

Source/items.cpp
```diff
@@ -4519,79 +4519,69 @@ void SpawnPremium(int pnum)
 void SpawnWitch(int lvl)
 {
 	constexpr int PinnedItemCount = 3;
+	constexpr std::array<int, PinnedItemCount> PinnedItemTypes = { IDI_MANA, IDI_FULLMANA, IDI_PORTAL };
```
ephphatha, line 4522: `GetItemAttrs` taking `int` instead of `_item_indexes` as the second param is not ideal, short of refactoring that function to take an ItemData refence I'd prefer to cast a member of this array to `int` explicitly when it is used than do the cast implicitly when declaring the array.

```suggestion
	constexpr std::array<_item_indexes, PinnedItemCount> PinnedItemTypes = { IDI_MANA, IDI_FULLMANA, IDI_PORTAL };
```

Same for PinnedBookTypes (even though IDI_BOOK2/8/14/20 haven't been defined) and the use in SpawnHealer

ephphatha (commented):
Still need to go through it properly but a couple notes from a quick skim

Source/items.cpp
```diff
@@ -4519,79 +4519,69 @@ void SpawnPremium(int pnum)
...
-		reservedItems = 10;
-
-		int books = GenerateRnd(4);
-		for (int i = 114, bCnt = 0; i <= 117 && bCnt < books; ++i) {
-			if (!WitchItemOk(i))
-				continue;
-			if (lvl < AllItemsList[i].iMinMLvl)
+		if (gbIsHellfire && i < PinnedItemCount + MaxPinnedBookCount && bookCount < pinnedBookCount) {
+			int bookType = PinnedBookTypes[i - PinnedItemCount];
+			if (lvl < AllItemsList[bookType].iMinMLvl)
```
ephphatha, line 4546: This function gets called with a `lvl` value clamped to the range 6, 16. The fourth book has a `iMinMLvl` of 20, so will never spawn.

With the way this condition is structured this PR will end up generating fewer items for low level characters (leaving a garbage in slots 4, 5, 6). Continuing the loop here leaves the current item as a zero-initialised `ItemType::Misc` instead of setting it to `ItemType::None` or generating a random item.

image

AJenbo: Should be fixed now

Source/items.cpp
AJenbo: Merged

Source/items.cpp
AJenbo: I can add a commit that implements all item types across the code base or do it as a follow up PR.

ephphatha: Eh, might as well do it properly* in a follow up.

*for a given definition of properly :D

Source/items.cpp
```diff
@@ -4514,77 +4514,71 @@ void SpawnPremium(int pnum)
...
-
-			bookItem = {};
-			bookItem._iSeed = AdvanceRndSeed();
-			SetRndSeed(bookItem._iSeed);
-			AdvanceRndSeed();
+		if (gbIsHellfire) {
+			if (i < PinnedItemCount + MaxPinnedBookCount && bookCount < pinnedBookCount) {
+				int bookType = PinnedBookTypes[i - PinnedItemCount];
+				if (lvl >= AllItemsList[bookType].iMinMLvl) {
+
```
ephphatha, line 4543:
```suggestion

				if (lvl >= AllItemsList[bookType].iMinMLvl) {
```
Surprised clang-format didn't complain about this

ephphatha (commented):
Beyond the formatting nit looks good to me, the combined loops are clear and given they get run when town loads there's no need to worry about performance. (Consider this approval, just wanna give a chance to reorder the whitespace before auto-merge kicks in)

Issue #3567, opened 2021-11-21
List of zerotier games overflows


ThomasChr:
**Describe**
Ajenbo noticed in discord that the list of available zerotier games does overflow

**To Reproduce**
Have enough open zerotier games, then go to the zerotier network menu

**Expected behavior**
The list should be scrollable

**Screenshots**
![E364B3C1-2E00-4AB2-BD62-22BFDD4F4606](https://user-images.githubusercontent.com/13014163/142756634-77ed20c7-3a59-45da-91d7-08e2c7a4e88e.png)

obligaron:
#3437 contains diablouI enhancements (caller doesn't need to managed viewport logic and scroll-bar simplifications) that could help with fixing this.

PR #3606, opened 2021-11-24
Pass viewportSize to UiList constructor and calculate height from it

### LINKED ISSUE #3595: Crash in character selection menu after clicking "Cancel"
**Important information**
Compiled debug https://github.com/diasurgical/devilutionX/commit/7b82c52da9f67916af962600b610796f325e88cc

**Describe**
Crash

**To Reproduce**
1. Go to character selection menu
2. Press cancel
3. Crash

**Expected behavior**
Should work fine

**Screenshots**
image

Maybe only happens above a certain number of characters

obligaron:
Fixes #3595

`UiList` constructed it's height from itemcount * `item_height`. With #3437 we now pass more then 6 item to the list and so the height is calculated wrong.
This PR pass `viewportSize` directly to UiList (instead of `UiInitList`) and calculate `UiList` height from it.

Helps also #3567 (scroll bar is still missing)

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -95,11 +95,11 @@ void AdjustListOffset(std::size_t itemIndex)
 
 } // namespace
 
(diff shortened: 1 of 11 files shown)


AJenbo (approved):
I noticed that this seamed to be a left over from when the lists where static constexpr structs.

PR #3609, opened 2021-11-25
Add comments so clang-format doesn't complain about eol space

ephphatha:


ephphatha:
Must've done that PR a while ago, it still had the bad indent from visual studio as well 😅

glebm (approved):

PR #3614, opened 2021-11-26
GitHub CI: Automatically add files to the release

glebm:
Not sure if this will works but would be nice

Uses https://github.com/ncipollo/release-action

### DIFF
--- a/.github/workflows/src_dist_release.yml
+++ b/.github/workflows/src_dist_release.yml
@@ -21,11 +21,22 @@ jobs:
 
     - name: Build
       working-directory: ${{github.workspace}}
-      run: tools/make_src_dist.py
+      run: tools/make_src_dist.py && mv build-src-dist/devilutionx-*.tar.xz devilutionx-src.tar.xz
 
     - name: Upload-Package
       if: ${{ !env.ACT }}
       uses: actions/upload-artifact@v2
       with:
         name: devilutionx-src.tar.xz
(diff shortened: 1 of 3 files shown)


.github/workflows/src_dist_release.yml
```diff
@@ -21,11 +21,22 @@ jobs:
...
+    - name: Update Release
+      if: startsWith(github.ref, 'refs/tags/') && ${{ !env.ACT }}
+      uses: ncipollo/release-action@v1
+      with:
+        artifacts: "devilutionx-src.tar.xz"
+        allowUpdates: true
+        omitBody: true
+        omitName: true
+        omitPrereleaseDuringUpdate: true
+        token: ${{ secrets.GITHUB_TOKEN }}
```
AJenbo, line 42: Do you know if this is already in place?

glebm: Ah, looks like we need to set permissions as well, one moment https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

glebm: Nevermind, we're using permissive defaults so probably don't need to set permissions
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

The token is always there btw

AJenbo (approved):

PR #3615, opened 2021-11-26
.github/workflows/src_dist_release.yml: Fix python

glebm:
Seems to default to Python 2 on Ubuntu 20.04, whereas we use Python 3

(previously failed at https://github.com/diasurgical/devilutionX/runs/4337688960?check_suite_focus=true)


AJenbo:
> we use Python 3

Good to know :)

PR #3619, opened 2021-11-27
Settings menu: Set line height for description 

### LINKED ISSUE #3603: [cjk] Setting description lines overlap in Japanese
image

glebm:
Before | After
---|---
settings-ja-before | settings-ja-new

Non-CJK is also improved:

Before | After
---|---
settings-en-before | settings-en-new

Fixes #3603


Source/DiabloUI/settingsmenu.cpp
```diff
@@ -124,7 +124,7 @@ void UiSettingsMenu()
 		const char *titleText = _("Settings");
 		vecDialog.push_back(std::make_unique<UiArtText>(titleText, MakeSdlRect(PANEL_LEFT, UI_OFFSET_Y + 161, PANEL_WIDTH, 35), UiFlags::FontSize30 | UiFlags::ColorUiSilver | UiFlags::AlignCenter, 8));
 		vecDialog.push_back(std::make_unique<UiScrollbar>(&ArtScrollBarBackground, &ArtScrollBarThumb, &ArtScrollBarArrow, MakeSdlRect(rectList.position.x + rectList.size.width + 5, rectList.position.y, 25, rectList.size.height)));
-		vecDialog.push_back(std::make_unique<UiArtText>(optionDescription, MakeSdlRect(rectDescription), UiFlags::FontSize12 | UiFlags::ColorUiSilverDark | UiFlags::AlignCenter, 1));
+		vecDialog.push_back(std::make_unique<UiArtText>(optionDescription, MakeSdlRect(rectDescription), UiFlags::FontSize12 | UiFlags::ColorUiSilverDark | UiFlags::AlignCenter, 1, IsSmallFontTall() ? 22 : 18));
```
obligaron, line 127: Sometimes we display 3 lines of description text.
With height 18 the last line will be near to the border.
To avoid this we could reduce the height to 16 or move the start of the description text up.

## line height 18
grafik 
## line height 16
grafik

obligaron:
Another question:
On my local machine the linending for `settingsmenu` is display as `CRLF`. But it is wrong in the repo.

Is this a misconfiguration on my side? Can I avoid this in the future with a different (git?) configuration?
Can we enable a clang-format check for it and hope that clang-format in the ci would detect it?

StephenCWills:
> Is this a misconfiguration on my side? Can I avoid this in the future with a different (git?) configuration?

It might have something to do with `git config core.autocrlf`.
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

Source/DiabloUI/settingsmenu.cpp
glebm: I've moved the description up by 4px and it now fits:

image

obligaron (approved):

obligaron:
> It might have something to do with `git config core.autocrlf`. https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

That would explain the difference. Thanks. 🙂 
Should we add `.gitattribute` to the repo to avoid this (for all contributors) in the future?

PR #3620, opened 2021-11-27
Fix chat messages overlap in CJK

### LINKED ISSUE #3548: [cjk] Chat messages overlap
image

It's worth nothing that I didn't enter the text here, rather I just hard coded a long line of Chinese, given the 80 byte limit (26 Chinese symbols). In reality the input box probably won't accept more then two lines, so it should be fine to increase the line height.

glebm:
Fixes #3548


AJenbo:
Does this both the input and the display of the messages?

glebm:
Just the display I think

AJenbo (approved):

PR #3634, opened 2021-11-28
Settings - Add language selection

obligaron:
## Notes

- Language list is at the moment hard coded. This could be improved in the future (searching for  *.mo/gmo files).
- No validation is currently applied (could be added later; for example check if language files really exists)
- UI needs to be recreated, cause we need to reload all texts (this will also be needed if we change resolution)
- I'm unsure if the language setting ("Languages") and the languages list ("English", "Spanish", ...) should be translatable. At the moment they aren't
- Didn't test polish voice pack
- UI could be improved (not part of this pr; for example we could show a sub-menu with the list entries)

## Sample Video

https://user-images.githubusercontent.com/25415264/143745625-a7269988-b7c7-4a15-ad5f-98d7567d8fee.mp4

### DIFF
--- a/Source/init.cpp
+++ b/Source/init.cpp
@@ -83,30 +83,7 @@ std::optional<MpqArchive> LoadMPQ(const std::vector<std::string> &paths, const c
 	return std::nullopt;
 }
 
-} // namespace
-
-void init_cleanup()
-{
-	if (gbIsMultiplayer && gbRunGame) {
-		pfile_write_hero(/*writeGameData=*/false, /*clearTables=*/true);
-	}
-
-	spawn_mpq = std::nullopt;
-	diabdat_mpq = std::nullopt;
-	hellfire_mpq = std::nullopt;
-	hfmonk_mpq = std::nullopt;
-	hfbard_mpq = std::nullopt;
-	hfbarb_mpq = std::nullopt;
-	hfmusic_mpq = std::nullopt;
-	hfvoice_mpq = std::nullopt;
-	lang_mpq = std::nullopt;
-	font_mpq = std::nullopt;
-	devilutionx_mpq = std::nullopt;
-
-	NetClose();
-}
-
-void init_archives()
+std::vector<std::string> GetMPQSearchPaths()
 {
 	std::vector<std::string> paths;
 	paths.push_back(paths::BasePath());
@@ -145,12 +122,11 @@ void init_archives()
 		LogVerbose("MPQ search paths:{}", message);
 	}
 
-#ifndef __ANDROID__
-	// Load devilutionx.mpq first to get the font file for error messages
-	devilutionx_mpq = LoadMPQ(paths, "devilutionx.mpq");
-#endif
-	font_mpq = LoadMPQ(paths, "fonts.mpq"); // Extra fonts
+	return paths;
+}
 
+void init_language_archives(const std::vector<std::string> &paths)
+{
 	if (strcasecmp("en", sgOptions.Language.szCode) != 0 || strlen(sgOptions.Language.szCode) != 2) {
 		char langMpqName[10] = {};
 		CopyUtf8(langMpqName, sgOptions.Language.szCode, sizeof(langMpqName));
@@ -158,6 +134,42 @@ void init_archives()
 		strncat(langMpqName, ".mpq", sizeof(langMpqName) - strlen(langMpqName) - 1);
 		lang_mpq = LoadMPQ(paths, langMpqName);
 	}
+}
+
+} // namespace
+
+void init_cleanup()
+{
+	if (gbIsMultiplayer && gbRunGame) {
+		pfile_write_hero(/*writeGameData=*/false, /*clearTables=*/true);
+	}
+
+	spawn_mpq = std::nullopt;
+	diabdat_mpq = std::nullopt;
+	hellfire_mpq = std::nullopt;
+	hfmonk_mpq = std::nullopt;
+	hfbard_mpq = std::nullopt;
+	hfbarb_mpq = std::nullopt;
+	hfmusic_mpq = std::nullopt;
+	hfvoice_mpq = std::nullopt;
+	lang_mpq = std::nullopt;
+	font_mpq = std::nullopt;
+	devilutionx_mpq = std::nullopt;
+
+	NetClose();
+}
+
+void init_archives()
+{
+	auto paths = GetMPQSearchPaths();
+
+#ifndef __ANDROID__
+	// Load devilutionx.mpq first to get the font file for error messages
+	devilutionx_mpq = LoadMPQ(paths, "devilutionx.mpq");
+#endif
+	font_mpq = LoadMPQ(paths, "fonts.mpq"); // Extra fonts
+
+	init_language_archives(paths);
 
 	diabdat_mpq = LoadMPQ(paths, "DIABDAT.MPQ");
 	if (!diabdat_mpq) {
@@ -197,6 +209,12 @@ void init_archives()
 		UiErrorOkDialog(_("Some Hellfire MPQs are missing"), _("Not all Hellfire MPQs were found.\nPlease copy all the hf*.mpq files."));
 		app_fatal(nullptr);
 	}
+}
+
+void init_language_archives()
+{
+	lang_mpq = std::nullopt;
+	init_language_archives(GetMPQSearchPaths());
 }
 
 void init_create_window()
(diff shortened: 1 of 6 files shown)


Source/options.cpp
```diff
@@ -712,13 +652,152 @@ std::vector<OptionEntryBase *> ChatOptions::GetEntries()
...
+	GetIniValue(category.data(), key.data(), szCode, sizeof(szCode), locale.c_str());
+}
+void OptionEntryLanguageCode::SaveToIni(string_view category) const
+{
+	SetIniValue(category.data(), key.data(), szCode);
+}
+
+void OptionEntryLanguageCode::CheckLanguagesAreInitialized() const
+{
+	if (languages.size() != 0)
```
glebm, line 732:
```suggestion
	if (!languages.empty())
```

Source/options.cpp
```diff
@@ -712,13 +652,152 @@ std::vector<OptionEntryBase *> ChatOptions::GetEntries()
...
+	std::string locale = std::string(vita_locales[language]);
+	sceAppUtilShutdown();
+#elif defined(__3DS__)
+	std::string locale = n3ds::GetLocale();
+#else
+	std::string locale = std::locale("").name().substr(0, 5);
+#endif
+
+	locale = locale.substr(0, 5);
+	LogVerbose("Prefered locale: {}", locale);
```
glebm, line 714:
```suggestion
	LogVerbose("Preferred locale: {}", locale);
```

Source/init.cpp
```diff
@@ -145,19 +122,55 @@ void init_archives()
...
+{
+	auto paths = GetMPQSearchPaths();
+
+#ifndef __ANDROID__
+	// Load devilutionx.mpq first to get the font file for error messages
+	devilutionx_mpq = LoadMPQ(paths, "devilutionx.mpq");
+#endif
+	font_mpq = LoadMPQ(paths, "fonts.mpq"); // Extra fonts
+
+	init_language_archives();
```
glebm, line 173: We already have `paths` here so let's just:

```suggestion
	init_language_archives(paths);
```

Source/init.cpp
```diff
@@ -145,19 +122,55 @@ void init_archives()
...
-#ifndef __ANDROID__
-	// Load devilutionx.mpq first to get the font file for error messages
-	devilutionx_mpq = LoadMPQ(paths, "devilutionx.mpq");
-#endif
-	font_mpq = LoadMPQ(paths, "fonts.mpq"); // Extra fonts
+	return paths;
+}
 
-	if (strcasecmp("en", sgOptions.Language.szCode) != 0 || strlen(sgOptions.Language.szCode) != 2) {
+void init_language_archives(std::vector<std::string> &paths)
```
glebm, line 128:
```suggestion
void init_language_archives(const std::vector<std::string> &paths)
```

Source/init.cpp
```diff
@@ -199,6 +212,13 @@ void init_archives()
 	}
 }
 
+void init_language_archives()
+{
+	lang_mpq = std::nullopt;
+	auto paths = GetMPQSearchPaths();
+	init_language_archives(paths);
```
glebm, line 219: Once the argument has been made const, we can simply do this thanks to reference lifetime extension.

```suggestion
	init_language_archives(GetMPQSearchPaths());
```

Source/utils/language.cpp
```diff
@@ -283,13 +283,15 @@ bool HasTranslation(const std::string &locale)
 
 bool IsSmallFontTall()
 {
-	string_view code(sgOptions.Language.szCode, 2);
+	string_view code = *sgOptions.Language.code;
```
glebm, line 286: We should be looking only at the prefix here, not the entire code:

```suggestion
	string_view code = (*sgOptions.Language.code).substr(0, 2);
```

glebm (changes requested):

Source/options.cpp
```diff
@@ -712,13 +652,152 @@ std::vector<OptionEntryBase *> ChatOptions::GetEntries()
...
+}
+void OptionEntryLanguageCode::SetActiveListIndex(size_t index)
+{
+	strcpy(szCode, languages[index].first.c_str());
+	NotifyValueChanged();
+}
+
+string_view OptionEntryLanguageCode::operator*() const
+{
+	return szCode;
```
glebm, line 788: Please move this to the header so that the compiler can inline it even without LTO. We access this often via `IsSmallFontTall`.

Source/options.cpp
obligaron: Fixed

Source/options.cpp
obligaron: Fixed

Source/init.cpp
obligaron: That was a careless mistake.  😞
I added the `init_language_archives` overload explicitly for this use case.

Source/init.cpp
obligaron: Fixed

Source/init.cpp
obligaron: Fixed

Source/utils/language.cpp
obligaron: Fixed

Source/options.cpp
obligaron: I will open a new pr for the `operator*()` in other OptionEntryBase based classes. 🙂

glebm (approved):
Looks good! Should "Language" perhaps be the first item in the settings menu? Seems like the most important setting if the game starts in the wrong language.

Source/options.cpp
```diff
@@ -712,13 +652,147 @@ std::vector<OptionEntryBase *> ChatOptions::GetEntries()
...
+	languages.push_back({ "de", "German" });
+	languages.push_back({ "en", "English" });
+	languages.push_back({ "es", "Spanish" });
+	languages.push_back({ "fr", "French" });
+	languages.push_back({ "ja", "Japanese" });
+	languages.push_back({ "hr", "Croatian" });
+	languages.push_back({ "it", "Italian" });
+	languages.push_back({ "ko_KR", "Korean" });
+	languages.push_back({ "pl", "Polish" });
+	languages.push_back({ "pt_BR", "Portuguese" });
```
glebm, line 748: Perhaps Brazilian Portuguese?

obligaron: I don't know the background and how far they differ.
We also use region specific extensions for `ko_KR` and `ro_RO` should we also adjust the text there?
Perhaps @AJenbo or the translator could clarify what they would expect?

glebm: Brazilian Portuguese is different enough from Portuguese that usually there is a separate translation for it.
Portuguese speakers need to be aware that this is `pt_BR` or else they'll report bugs :) /cc @julealgon 

This is not the case with `ko_KR` or `ro_RO`

obligaron:
>Should "Language" perhaps be the first item in the settings menu?

This would be no problem, only change the order in `Options::GetCategories`.
But perhaps we should defer this when we have most options implemented and then we can reoder them all? 🤔

glebm:
>  But perhaps we should defer this when we have most options implemented and then we can reoder them all?

Sure, I don't mind either way

Source/init.cpp
AJenbo: wasent there a bug in windows regarding this?

Source/options.cpp
AJenbo: The Korean governments differ in there policies on language. This coupled with the low cultural exchange rate appears to have caused the languages to start to drift apart. So the `ko_KR` distinction makes sens here. However considering the sanctions, there is little reason to carry out this distinction in the user interface.

The `_RO` seams unnecessary and should be removed for better auto detection.

obligaron: I changed to `Portuguese (Brazil)`. 🙂

AJenbo (approved):

Source/options.cpp
julealgon: Good callout by @glebm . They are actually vastly different from one another so the distinction is indeed the way to go.

Sometimes in local apps we have "Portugues" (Portuguese) and "Portugues de Portugal" (Portugal Portuguese) options. That's a more common way we refer to "Portugues" in general over here. However, that's only the case when the information is already presented in pt-BR in the first place. For neutral/english interfaces such as this it makes sense to indicate the country in both options.

A pt-BR translated version of this menu could present "Portugues" (unqualified) and "Portugues de Portugal" as options and it would not be confusing to Brazilian folks.

PR #3635, opened 2021-11-28
Add an option to build SDL2 from source

glebm:
Replaces the Android-specific version of this.
We'll need it for the iOS build as well.

Also switches to the 2.0.16 release version now that it's out.


AJenbo (approved):
Thanks :)

PR #3640, opened 2021-11-28
Updated Czech translation

Srandista:
Updated Czech translation with story lines (used already existing translations from Adam Rambousek [Diablo 1] and René "Raptor" Spišák [Hellfire])

### DIFF
--- a/Translations/cs.po
+++ b/Translations/cs.po
@@ -321,26 +321,20 @@ msgid "OK"
 msgstr "OK"
 
 #: Source/DiabloUI/extrasmenu.cpp:41 Source/DiabloUI/selstart.cpp:44
-#, fuzzy
-#| msgid "Down to Diablo"
 msgid "Switch to Diablo"
-msgstr "Dolů do Hnízda úrovně {:d}"
+msgstr ""
 
 #: Source/DiabloUI/extrasmenu.cpp:41
-#, fuzzy
-#| msgid "Exit Hellfire"
 msgid "Switch to Hellfire"
-msgstr "Ukonči Hellfire"
+msgstr ""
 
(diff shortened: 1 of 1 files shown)


Translations/cs.po
```diff
@@ -3647,24 +3633,20 @@ msgstr "Olej Nepropustnosti"
...
-msgstr "{:s} {:s}"
+msgstr ""
 
 #. TRANSLATORS: Constructs item names. Format: {Prefix} {Item} of {Spell}. Example: King's War Staff of Firewall
 #: Source/items.cpp:1193
-#, fuzzy
 msgctxt "spell"
 msgid "{0} {1} of {2}"
-msgstr "{:s} {:s} {:s}"
+msgstr ""
```
NikoVP, line 3644:
```suggestion
msgstr "{0} {1} of {2}"
```

AJenbo:
Thank you :)

AJenbo:
@Psojed @Srandista if either of you have time to update the translation now would be a good time to make it for the next release :)

There are about 250 lines that needs an update.

Psojed:
I'll take a look :)

But on the subject of this commit.. @Srandista do you have permission from the original authors? Or did you just "borrow" these translations from the internet?

Srandista:
@Psojed As stated in OP, I did used already existing traslations. I think it's clearly stated, I take no credit for the transition itself and credited both original authors immediately.

Psojed:
@Srandista I'm not asking where are these translations from. I'm asking, if you had premission from the original authors to use their work.

Srandista:
Yes, I have e-mail confirmation from both authors. Anything else?

PR #3642, opened 2021-11-28
Settings: Add Game Mode and Shareware and remove specials in settingsmenu

obligaron:
## Content

- Add Game Mode to Settings
- Add Shareware to Settings
- Remove Special Actions `Switch to Diablo/Hellfire` and `Switch to Fullgame/Shareware`
- Unify `gbIsSpawn` and `gbSpawned`

## Screenshot

grafik

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -235,6 +235,11 @@ void OptionLanguageCodeChanged()
 	init_language_archives();
 }
 
+void OptionGameModeChanged()
+{
+	gbIsHellfire = *sgOptions.StartUp.gameMode == StartUpGameMode::Hellfire;
+}
+
 } // namespace
 
 void SetIniValue(const char *sectionName, const char *keyName, const char *value, int len)
@@ -272,7 +277,6 @@ void LoadOptions()
 	sgOptions.Diablo.lastMultiplayerHero = GetIniInt("Diablo", "LastMultiplayerHero", 0);
 	sgOptions.Hellfire.lastSinglePlayerHero = GetIniInt("Hellfire", "LastSinglePlayerHero", 0);
 	sgOptions.Hellfire.lastMultiplayerHero = GetIniInt("Hellfire", "LastMultiplayerHero", 0);
-	sgOptions.Hellfire.startUpGameOption = static_cast<StartUpGameOption>(GetIniInt("Hellfire", "StartUpGameOption", static_cast<int>(StartUpGameOption::None)));
 	GetIniValue("Hellfire", "SItem", sgOptions.Hellfire.szItem, sizeof(sgOptions.Hellfire.szItem), "");
 
 	sgOptions.Audio.nSoundVolume = GetIniInt("Audio", "Sound Volume", VOLUME_MAX);
@@ -343,7 +347,6 @@ void SaveOptions()
 	SetIniValue("Hellfire", "SItem", sgOptions.Hellfire.szItem);
 	SetIniValue("Hellfire", "LastSinglePlayerHero", sgOptions.Hellfire.lastSinglePlayerHero);
 	SetIniValue("Hellfire", "LastMultiplayerHero", sgOptions.Hellfire.lastMultiplayerHero);
-	SetIniValue("Hellfire", "StartUpGameOption", static_cast<int>(sgOptions.Hellfire.startUpGameOption));
 
 	SetIniValue("Audio", "Sound Volume", sgOptions.Audio.nSoundVolume);
 	SetIniValue("Audio", "Music Volume", sgOptions.Audio.nMusicVolume);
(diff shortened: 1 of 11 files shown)


Source/options.cpp
```diff
@@ -235,6 +235,16 @@ void OptionLanguageCodeChanged()
 	init_language_archives();
 }
 
+void OptionGameModeChanged()
+{
+	gbIsHellfire = *sgOptions.StartUp.gameMode == StartUpGameMode::Hellfire;
+}
+
+void OptionSharwareChanged()
```
glebm, line 243:
```suggestion
void OptionSharewareChanged()
```

glebm:
Do the `--diablo`, `--hellfire`, and `--spawn` still work with this? Just want to make sure

Source/options.cpp
qndel: image

obligaron: Method name was only compressed 🙄 😁 
(Fixed)

StephenCWills:
> Do the `--diablo`, `--hellfire`, and `--spawn` arguments still work with this? Just want to make sure

I just did a bit of testing, and they work about as well as you'd suspect.

* The arguments override whatever the current INI settings are at startup
* The settings menu shows you whatever the game mode is in the INI
* If you use the settings menu to change the game mode, whatever you select will override the arguments
* When you exit the game, it changes INI state only if you used the settings menu to change it

I haven't reviewed the code, but I like this change because the previously defined special actions did not update the INI, and I found it to be pretty annoying.

Source/options.cpp
```diff
@@ -505,6 +513,13 @@ string_view OptionCategoryBase::GetDescription() const
 
 StartUpOptions::StartUpOptions()
     : OptionCategoryBase("StartUp", N_("Start Up"), N_("Start Up Settings"))
+    , gameMode("Game", OptionEntryFlags::NeedHellfireMpq | OptionEntryFlags::RecreateUI, N_("Game Mode"), N_("Play Diablo or Hellfire."), StartUpGameMode::Ask,
+          {
+              { StartUpGameMode::Diablo, N_("Diablo") },
+              // Ask is missing, cause we want to hide it from UI-Settings.
+              { StartUpGameMode::Hellfire, N_("Hellfire") },
+          })
+    , shareware("Shareware", OptionEntryFlags::NeedDiabloMpq | OptionEntryFlags::RecreateUI, N_("Restrict to Shareware"), N_("Makes the game compatible with the shareware. Allows to play multiplayer games with friends that don't own a diablo copy."), false)
```
StephenCWills, line 522: We started using the term "demo" for Spawn in the Android build so I think it's also appropriate here. Also, Diablo should be capitalized. Other than that, this is just a gentle suggestion to reword this description.

```suggestion
    , shareware("Shareware", OptionEntryFlags::NeedDiabloMpq | OptionEntryFlags::RecreateUI, N_("Restrict to Shareware"), N_("Makes the game compatible with the demo. Enables multiplayer with friends who don't own a full copy of Diablo."), false)
```

StephenCWills (changes requested):

Source/options.cpp
obligaron: The new description is a improvement, so changed the description,. 😉 

But your comment let me reflect how we should name the "not full version":
- `shareware`
- `spawn`
- `demo`

Currently we use all three. Do we want to use only one term?

`shareware` is used in the UI and (in this pr) in `ini` and some option code
`spawn` is used in mpq, code and arguments (`--spawn`)
`demo` is used in android build and description

StephenCWills: The Android build currently uses the phrase:
> Alternatively you can download the spawn version, to play the demo portion of the game.

So we use both `spawn` and `demo` in the same sentence. It seems our usage suggests that all three are valid terms for referring to that version of the game. This is largely up to personal opinion, but some brief notes about the three terms.

* Spawn is the term used in the file names for the exe and the mpq. It is perhaps the most accurate term to use for describing this version of the game to those who are familiar with it. However, it's something of a made-up term and doesn't really convey much useful information to anyone who doesn't already know what it is.
* Shareware is perhaps the best term to convey the idea that this version of the game is free to play. It typically refers to closed-source-but-freely-available software so it may not be all that accurate to use it in the context of DevilutionX. However, Blizzard did use this term in the background art for the main menu so it's probably here to stay. This term may not be as recognizable to players who are not as familiar with personal computers.
* Demo is perhaps the best term to convey the idea that you will only be experiencing a subset of the full game's content. I would say this is the most broadly understood of the three terms since it's used in a variety of contexts.

I think it's fairly safe to default to `demo` in most contexts if you're having trouble figuring out which one to use.

StephenCWills (approved):

PR #3645, opened 2021-11-29
Try load the translation in the users language on windows

ephphatha:
Lets see if this builds on CI...

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -24,8 +24,6 @@
 
 #ifdef _WIN32
 // clang-format off
-// Need to explicitly define the minimum supported windows version for mingw builds
-#define WINVER 0x0600
 #include <windows.h>
 #include <winnls.h>
 // clang-format on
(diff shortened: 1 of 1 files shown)


Source/options.cpp
```diff
@@ -728,6 +735,37 @@ void OptionEntryLanguageCode::LoadFromIni(string_view category)
...
+
+#if WINVER >= 0x0600
+	WCHAR localeBuffer[LOCALE_NAME_MAX_LENGTH];
+	if (GetUserDefaultLocaleName(localeBuffer, LOCALE_NAME_MAX_LENGTH) != 0) {
+		// The user default locale could be loaded, we need to convert from WIN32's default of UTF16 to UTF8
+		char utf8Buffer[12] {};
+		// We only handle 5 character locales (lang2-region2), so don't bother reading past that. This does leave the resulting string unterminated but the buffer was zero initialised anyway.
+		WideCharToMultiByte(CP_UTF8, 0, localeBuffer, 5, utf8Buffer, 12, nullptr, nullptr);
+
+		// GetUserDefaultLocaleName could return an ISO 639-2/T string (three letter language code) or even an arbitrary custom locale, however we only handle 639-1 (two letter language code) locale names when checking the fallback language.
```
AJenbo, line 749: Should we map ISO 639-2 since cutting alpha-3 to 2 chars won't give the correct langauge in many cases?

ephphatha: It's not strictly necessary for the current supported languages. alpha3 codes are only returned if no alpha2 code exists for the given language. This would be an issue if we got a translation for a language that has no alpha2 representation (e.g. Hawaiian). In that case the code consuming locale needs to be updated to remove the hardcoded substr lengths.

AJenbo (approved):

PR #3648, opened 2021-11-30
CMake: Improve translation deps handling

glebm:
Instead of using the `GETTEXT_PROCESS_PO_FILES` function, do it ourselves in a cleaner way.

This fixes a number of issues:

1. Translations and assets are now built whenever the binary is built, without causing the binary to be rebuilt.
2. Translations are only copied to `build/assets` (previously they were also copied to `build/`).
3. This should now be compatible with the new XCode (@bubio who hit this issue).
4. Visual Studio and VS Code now generate the assets when running/debugging the binary (@qndel who hit this issue).

The main new trick we're using is to depend both on the command output file and the associated custom target.

The translation target has no outputs, so the targets that depend on it are not rebuilt when it changes.
Depending on the command output file as well triggers the rebuild correctly (e.g. `devilutionx.mpq` is rebuilt when a translation file changes), while depending on the target at the same time ensures that we do not violate the following CMake rule:

>  Do not list the output in more than one independent target that may build in parallel or the two instances of the rule may conflict (instead use add_custom_target to drive the command and make the other targets depend on that one).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -724,21 +724,41 @@ if(USE_GETTEXT_FROM_VCPKG)
 endif()
 find_package(Gettext)
 if (Gettext_FOUND)
+  file(MAKE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/assets")
   foreach(lang ${devilutionx_langs})
-    GETTEXT_PROCESS_PO_FILES(${lang} ALL PO_FILES ${CMAKE_CURRENT_LIST_DIR}/Translations/${lang}.po)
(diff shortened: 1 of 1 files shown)


glebm:
@AJenbo I've noticed we only copy the translation files to the `android-project` assets directory but not the other asset files. Is that handled elsewhere?

AJenbo:
Assets are copied here for Android:
https://github.com/diasurgical/devilutionX/blob/master/CMake/android_defs.cmake#L15

AJenbo (approved):

glebm:
@AJenbo Oh, we need to fix that as well. `file(COPY` only runs at configure time, so the assets aren't copied on change without reconfiguring.

PR #3649, opened 2021-11-30
Remove the build directory from the repo

glebm:
There is no benefit to having it anymore.

The minimum CMake version we require supports the `cmake -S. -Bbuild` invocation which creates the directory. This is also the command we now use in the build instructions.


tsunamistate:
I was worried that `build` folder will not be created using Visual Studio via File - Open - CMake (the method recommended in the docs).
However, folder `build/{arch}-{configuration}` (e.g. build/x64-Debug) is created automatically with fresh checkout of this PR.
Approved from Windows side 👍

StephenCWills (approved):

AJenbo:
More special argument to remember i suppose :shrug:

glebm:
Let go of the old ways! No need to `cd` anymore!

StephenCWills:
There's no real reason to let go. You can still use `cd` if you `mkdir` first. IMO, the parameters are better for the automated scripts and the instructions.

AJenbo:
True :)

Issue #3329, opened 2021-10-30
Crash in SDL_GameController during video playback 


AJenbo:
This is happening on at least 5 phones, it could possibly be due to a corrupted spawn.mpq from a bad download, so if we can replicate it that way I think we can simply close the issue.

```
libSDL2.so (SDL_GameControllerGetBindForButton_REAL)
libSDL2.so (SDL_GameControllerHasButton_REAL)
libdevilutionx.so (devilution::ProcessControllerMotion(SDL_Event const&, devilution::ControllerButtonEvent))
libdevilutionx.so (devilution::FetchMessage_Real(devilution::tagMSG*))
libdevilutionx.so (devilution::play_movie(char const*, bool))
libdevilutionx.so (devilution::DiabloMain(int, char**))
```

StephenCWills:
I've tried everything I can think of, but I can't seem to replicate this.

PR #3662, opened 2021-12-02
[Android] Clear memory on exit with System.exit(0)

### LINKED ISSUE #3553: Android: Game crashes every other startup.
Device: Xiaomi Black Shark 2 Pro (128GB + 8GB RAM)
Gamepad: Razer Kishi for Android (Xbox Variant)
OS: Android 10
DevilutionX Version: v1.3.0

Description: When quitting the game, there is a high chance that the app will instacrash next time you try to run it. Trying again after said crash will start the game normally. Thankfully, settings and saves aren't affected by this issue.

To Reproduce:
1) Start DevilutionX normally.
2) Quit the game.
3) Attempt to start DevilutionX again, and note how it fails to start, going to a black screen briefly before crashing, with no error message.
4) Try to start to game one more time, and it'll boot properly now.

Expected Behavior: The game would boot up every time. without fail.

Screenshots: None.

Additional Context: None.

StephenCWills:
It turns out that #3329 and #3553 are the same issue.

I was completely unable to reproduce these errors on my Pixel 4, but the Pixel 2 AVD running Android 10 reproduced them quite easily.

As @AJenbo pointed out (https://github.com/diasurgical/devilutionX/issues/3553#issuecomment-974666734), `System.exit(0)` does not seem to do anything. The game still shows up in the task switcher and will restart when selected from there. However, it seems that under the covers the OS will reclaim heap data without reclaiming or reinitializing static data if you do not use it. This causes the static `controllers_` vector to maintain a reference to a deleted `GameController` instance on restart.

This change also has the nice added benefit of detaching the Android Studio debugger on exit.

This resolves #3553

### DIFF
--- a/android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
+++ b/android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
@@ -26,6 +26,7 @@ import java.util.Objects;
 
 public class DevilutionXSDLActivity extends SDLActivity {
 	private String externalDir;
+	private boolean noExit;
 
 	protected void onCreate(Bundle savedInstanceState) {
 		// windowSoftInputMode=adjustPan stopped working
@@ -49,7 +50,20 @@ public class DevilutionXSDLActivity extends SDLActivity {
 		if (missingGameData()) {
 			Intent intent = new Intent(this, DataActivity.class);
 			startActivity(intent);
+			noExit = true;
 			this.finish();
+		}
+	}
+
+	/**
+	 * When the user exits the game, use System.exit(0)
+	 * to clear memory and prevent errors on restart
+	 */
+	protected void onDestroy() {
+		super.onDestroy();
+
+		if (!noExit) {
+			System.exit(0);
 		}
 	}
 


AJenbo (approved):

pionere:
I do not know much about Android, but using System.exit(0) is very much advised against.
How about something like:
```
protected void onDestroy() {
		super.onDestroy();

		if (!isFinishing()) {
			finish();
		}
	}
```

StephenCWills:
> I do not know much about Android, but using System.exit(0) is very much advised against.

In actuality, the community seems to be rather split on the subject, with lots of vocal advocates both for the use of `System.exit(0)` *in specific situations* and against the use of it altogether. From what I've been able to tell, it doesn't seem like much of anybody has a good grasp on why they feel a particular way about this issue. On the one side, it just seems like some people who personally had success with using `System.exit(0)` end up with a vague sense that the OS is holding onto some portion of memory for the application and maintain that there is no other way programmatically to force the OS to clear it. On the other side, it seems like everybody's just parroting the vague notion that you should just let the OS do what it's designed to do lest you be forced to deal with the consequences.

One of the top arguments on Google advising against using `System.exit(0)` essentially amounts to "it doesn't do what you'd expect" in that it doesn't actually terminate the app. If you have multiple activities in the "back stack," it only terminates one of them, which is similar to the behavior of `finish()`--although it does apparently terminate the JVM process running that activity which certainly explains why the static data gets cleared. One of the suggestions I read was to call `finishAffinity()` before calling `System.exit(0)` to first ensure that all activities are finished and then to terminate the JVM process. This should obviously be unnecessary if you don't have any activities left in the back stack, which is the case for DevilutionX once the `DevilutionXSDLActivity` is finished.

I've also seen one report that `System.exit(0)` may have some adverse effects related to runtime permissions and maybe caching. This makes some sense since you're basically having the OS forcefully clear that cached memory. I haven't seen any particular issues when it comes to DevilutionX, but it would probably be prudent to keep this in mind if we add any more Java code to the mix.

> How about something like:
> 
> ```
> protected void onDestroy() {
> 		super.onDestroy();
> 
> 		if (!isFinishing()) {
> 			finish();
> 		}
> 	}
> ```

That doesn't make much sense since, at least in the case of the issue being solved here, we end up in `onDestroy()` as a direct result of having already called `finish()`. My testing reveals that this suggestion does not resolve the issue.

pionere:
> That doesn't make much sense since, at least in the case of the issue being solved here, we end up in onDestroy() as a direct result of having already called finish().

I think it makes perfect sense. In the documentation they explicitly say to call isFinishing() to test if the OS is completely closing the application or just tries to leave it in the background. If what you say is true, then there would be no point of calling isFinishing() in [onDestroy()](https://developer.android.com/reference/android/app/Activity#onDestroy()) at all.

> My testing reveals that this suggestion does not resolve the issue.

Well, then that pretty much settles the case...

StephenCWills:
> I think it makes perfect sense. In the documentation they explicitly say to call isFinishing() to test if the OS is completely closing the application or just tries to leave it in the background. If what you say is true, then there would be no point of calling isFinishing() in [onDestroy()](https://developer.android.com/reference/android/app/Activity#onDestroy()) at all.

I meant that it doesn't make sense in the context of this issue. The issue occurs because when `nativeRunMain()` completes, the `SDLMain` class calls `finish()` on the `SDLActivity`, which is what leads to the call to `onDestroy()` and subsequently the error when the app is relaunched. From that perspective, it stands to reason that calling `finish()` again shouldn't resolve the issue.

pionere:
My only concern is that I think calling `System.exit(0)` is akin to pressing the reset button. Sure it works, but most probably there is an underlying problem which should be fixed instead. E.g. maybe that `controllers_` vector should be cleared/freed on exit?

StephenCWills:
> My only concern is that I think calling `System.exit(0)` is akin to pressing the reset button. Sure it works, but most probably there is an underlying problem which should be fixed instead. E.g. maybe that `controllers_` vector should be cleared/freed on exit?

The OS should be cleaning up process memory when the application shuts down. At least, that's the expectation we have on every other platform, and on most platforms it would be a serious OS bug if it didn't. What we've done here is (arguably) something like a workaround for an Android NDK bug. Surely Google agrees or they wouldn't have fixed it after Android 10.

We certainly could attempt to clear out the `controllers_` collection on exit, and that would (probably) solve this particular error. However, I absolutely do not trust that this quirk isn't going to cause some other more obscure issue, and we don't necessarily have control over all the static data that is used by third-party dependencies (think ZeroTier). I think it's perfectly reasonable to force the OS to do what we need it to so long as it doesn't cause any adverse effects. If we notice some bad behavior as a result of this, we can consider other options.

pionere:
It is your choice...
https://github.com/pionere/devilutionX/commit/9c8c7d1f898e9a8f41ac8e3855677423b78935fe (+ https://github.com/pionere/devilutionX/commit/52e7656a618f6d89d6280faddeacd4d6f8bfdbac)

glebm:
From what I understand, the issue is as follows:

1. Java has `LoadLibrary` but no `UnloadLibrary`.
2. The Java process is cached so the `.so` stays in memory.

I think a robust way to solve this would be to have another `.so` (e.g. `devilutionx_loader.so`), that can load and unload `devilutionx.so` in response to the Java application lifecycle.

> Surely Google agrees or they wouldn't have fixed it after Android 10.

I'm very curious about this in light of the above. Do you know how this was fixed?

StephenCWills:
> I'm very curious about this in light of the above. Do you know how this was fixed?

I wish I did. All I know is that I couldn't reproduce the issue on Android 12, and a quick review on Google Play Console just now reveals that all the crash reports are for Android 10 and 11.

StephenCWills:
> I think a robust way to solve this would be to have another `.so` (e.g. `devilutionx_manager.so`) that can load and unload `devilutionx.so` in response to the Java application lifecycle.

I looked into this and found that SDL was already invoking the main method of our application using `dlopen()` and `dlsym()`.
https://github.com/libsdl-org/SDL/blob/b7a0200aab8088e04800d33cc88fee384df0d78c/src/core/android/SDL_android.c#L723-L740

`dlclose()` is called after `SDL_main()` returns but before `nativeRunMain()` returns.
https://github.com/libsdl-org/SDL/blob/b7a0200aab8088e04800d33cc88fee384df0d78c/src/core/android/SDL_android.c#L776-L790

So all we should have to do to test this theory is make sure that the Java code doesn't invoke `SDL.loadLibrary("devilutionx")`.
https://github.com/StephenCWills/devilutionX/commit/c4bd6db40f7ff4c250c01e7330482efd5aef17c5

Even this did not resolve the error. Maybe I did something wrong. Feel free to run your own tests against that branch or provide additional suggestions.

PR #3663, opened 2021-12-02
Handle multiple language preferences on windows

ephphatha:
This PR looks for all region-specific locales according to user preferences to find a matching translation. The last commit updates the fallback behaviour to look for non-regional translations after the last regional variation of a given language.

For example, a user preference of `en_AU, fr_FR, en_US` will search for translations in the order `en_AU, fr_FR, fr, en_US, en`. `en_AU` is the first preference so will be checked first, but because another english regional variation is in the list to be checked in the future we do not look for non-regional translations yet. `fr_FR` is immediately followed by a check for a non-regional `fr` translation as no other regional variations were specified. Finally `en_US` then `en` are checked being the last regional variation of english in the list of languages.

To ensure that cases where `en_US` or `en` are higher in the user preferences than other languages with translation files `HasTranslation` includes an explicit check for these codes. Without this these languages would never be picked over another existing translation since there should never be a .mo/.gmo file for those codes.

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -911,6 +911,25 @@ void OptionEntryLanguageCode::LoadFromIni(string_view category)
 		}
 	}
 
+	std::vector<std::string> fallbackLocales;
+	fallbackLocales.reserve(locales.size());
+
+	for (const auto &locale : locales) {
+		std::string neutralLocale = locale.substr(0, locale.find('_'));
+		if (std::find(fallbackLocales.cbegin(), fallbackLocales.cend(), neutralLocale) == fallbackLocales.cend()) {
+			fallbackLocales.push_back(neutralLocale);
+		}
+	}
+
+	for (const auto &fallbackLocale : fallbackLocales) {
+		LogVerbose("Trying to load fallback translation: {}", fallbackLocale);
+		if (HasTranslation(fallbackLocale)) {
+			LogVerbose("Found matching fallback locale: {}", fallbackLocale);
+			CopyUtf8(szCode, fallbackLocale, sizeof(szCode));
+			return;
+		}
+	}
+
 	LogVerbose("No suitable translation found");
 	strcpy(szCode, "en");
 }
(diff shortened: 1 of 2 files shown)


Source/utils/language.cpp
```diff
@@ -270,15 +270,19 @@ const std::string &LanguageTranslate(const char *key)
...
 bool HasTranslation(const std::string &locale)
 {
-	for (const char *ext : { ".mo", ".gmo" }) {
-		SDL_RWops *rw = OpenAsset((locale + ext).c_str());
+	if (locale == "en") {
+		return true;
+	}
+
+	constexpr std::array<char *, 2> Extensions { ".mo", ".gmo" };
+	return std::any_of(Extensions.cbegin(), Extensions.cend(), [locale](const std::string &extension) {
```
ephphatha, line 278: [readability-use-anyofallof](https://clang.llvm.org/extra/clang-tidy/checks/readability-use-anyofallof.html) is enabled so I updated this while I was in the area. Personally I'm not a fan but it's one less squiggly line 🙃

Source/options.cpp
```diff
@@ -839,56 +848,82 @@ void OptionEntryLanguageCode::LoadFromIni(string_view category)
...
-		if (!HasTranslation(locale)) {
-			locale = "en";
+	// Insert non-regional locale codes after the last regional variation so we fallback to neutral translations if no regional translation exists that meets user preferences.
+	for (auto localeIter = locales.rbegin(); localeIter != locales.rend(); localeIter++) {
+		auto regionSeparator = localeIter->find('_');
+		if (regionSeparator != std::string::npos) {
+			std::string neutralLocale = localeIter->substr(0, regionSeparator);
+			if (std::find(locales.rbegin(), localeIter, neutralLocale) == localeIter) {
+				localeIter = std::make_reverse_iterator(locales.insert(localeIter.base(), neutralLocale));
+			}
```
ephphatha, line 910: This assumes that user locales will always be regional variations, but if a user does have a non-regional locale earlier in the list then at worst we're duplicating the check one extra time.

AJenbo (approved):
This seams to follow how other tools prioritize picking a matching language :+1: 

I do think that maybe some of all of this logic should be moved out of options.cpp, some in to platform specific files and some in to the language file. Maybe, but that can wait and isn't really related anyway.

PR #3664, opened 2021-12-03
Quick cast

vlolteanu:
This patch lets you cast spells with the push of a hotkey, rather than swapping out the readied spell.

Something similar has recently been introduced to D2R.

### DIFF
--- a/Source/controls/plrctrls.h
+++ b/Source/controls/plrctrls.h
@@ -3,6 +3,8 @@
 
 #include <cstdint>
 
+#include "player.h"
+
 namespace devilution {
 
 typedef enum belt_item_type : uint8_t {
@@ -45,6 +47,7 @@ bool TryDropItem();
 void InvalidateInventorySlot();
 void FocusOnInventory();
 void PerformSpellAction();
+void QuickCast(int slot);
 
 extern int speedspellcount;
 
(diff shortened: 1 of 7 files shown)


Source/controls/plrctrls.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 // Controller actions implementation
 
+#include "player.h"
```
glebm, line 4: Our includes should come after:

1. Standard library
2. Other libraries
2. Ours

glebm:
The C++ bits looks good to me 👍

Source/player.h
```diff
@@ -631,7 +631,7 @@ void ClrPlrPath(Player &player);
 bool PosOkPlayer(const Player &player, Point position);
 void MakePlrPath(Player &player, Point targetPosition, bool endspace);
 void CalcPlrStaff(Player &player);
-void CheckPlrSpell(bool isShiftHeld);
+void CheckPlrSpell(bool isShiftHeld, spell_id spellID = Players[MyPlayerId]._pRSpell, spell_type spellType = Players[MyPlayerId]._pRSplType);
```
AJenbo, line 634: I did not expect it to be possible to set this complex of a default :o

```suggestion
void CheckPlrSpell(bool isShiftHeld, spell_id spellID = MyPlayer->_pRSpell, spell_type spellType = MyPlayer->_pRSplType);
```

Source/controls/plrctrls.h
vlolteanu: I know. It turns out clang-format places them in this order them if you don't add a blank line in between.

AJenbo (approved):

PR #3667, opened 2021-12-04
Add text to translate

bubio:
- Fixed that  "Define what language to use in game." is not included in the translation.

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -701,7 +701,7 @@ std::vector<OptionEntryBase *> ChatOptions::GetEntries()
 }
 
 OptionEntryLanguageCode::OptionEntryLanguageCode()
-    : OptionEntryListBase("Code", OptionEntryFlags::CantChangeInGame | OptionEntryFlags::RecreateUI, "Language", "Define what language to use in game")
+    : OptionEntryListBase("Code", OptionEntryFlags::CantChangeInGame | OptionEntryFlags::RecreateUI, "Language", N_("Define what language to use in game."))
 {
 }
 void OptionEntryLanguageCode::LoadFromIni(string_view category)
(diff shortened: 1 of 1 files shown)


Source/options.cpp
```diff
@@ -701,7 +701,7 @@ std::vector<OptionEntryBase *> ChatOptions::GetEntries()
 }
 
 OptionEntryLanguageCode::OptionEntryLanguageCode()
-    : OptionEntryListBase("Code", OptionEntryFlags::CantChangeInGame | OptionEntryFlags::RecreateUI, "Language", "Define what language to use in game")
+    : OptionEntryListBase("Code", OptionEntryFlags::CantChangeInGame | OptionEntryFlags::RecreateUI, "Language", N_("Define what language to use in game."))
```
glebm, line 704: Let's do this for `"Language"` as well?

glebm:
I see you keep using the master branch :sweat_smile: 

image

It'd be much easier for you to send each PR from a separate branch.

Source/options.cpp
bubio: "Language" was already there, so we did not apply it.

bubio:
> I see you keep using the master branch 😅
> 
> image
> 
> It'd be much easier for you to send each PR from a separate branch.

ズバリ、そうでしょう！！😁

I tried that, but it left me with a commit list, so I decided to reset it.

```
git fetch --all
git rebase upstream/master
git push --force
```

Source/options.cpp
AJenbo: If the other line is removed it would cause the string to no longer be translated, so best to have it marked explicitly.

```suggestion
    : OptionEntryListBase("Code", OptionEntryFlags::CantChangeInGame | OptionEntryFlags::RecreateUI, N_("Language"), N_("Define what language to use in game."))
```

PR #3670, opened 2021-12-04
Settingsmenu: Clean up UI-Resources before switching options that require RecreateUI

### LINKED ISSUE #3668: Crashes when changing language settings.(macOS)
**Important information**
macOS 12.0.1（21A559）
Self Build: e22bdda1b9977ed442f8172d755a957b63c67c22

**Describe**
Crashes when changing language settings.

**To Reproduce**
1. Click on 'SETTINGS'
2. Click on 'LANGUAGE'
3. Crash

**Expected behavior**
Language setting will be changed and the menu will be redrawn.

**Additional context**
It worked fine on Windows(x64) and Linux.
Crash log is attached.
[Issue_3668.txt](https://github.com/diasurgical/devilutionX/files/7653618/Issue_3668.txt)

obligaron:
Fixes #3668 (hopefully)

`UiPollAndRender` calls `UiHandleEvents`
`UiHandleEvents` calls `ItemSelected`. We change the language, mark the UI to be recreated.
But after that `UiPollAndRender` calls `UiRenderItems` once. And then we could access freed memory.

This pr avoid this by do the cleanup first (with `UiInitList_clear`), change settings and then recreates the UI and settingsmenu after that.

I don't own a mac, so can't verify it.
@bubio could you test this pr to verify if this fixes your crash on macOS?

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -563,6 +563,8 @@ void LoadUiGFX()
 	LoadHeros();
 }
 
+} // namespace
+
 void UnloadUiGFX()
 {
 	ArtHero.Unload();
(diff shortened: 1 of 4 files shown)


bubio:
I tested it.
There was no problem.😀

AJenbo (approved):

PR #688, opened 2020-04-05
Bug on Windows: game crashes when the PC wakes up from sleep

tyomitch:
If I close and reopen my laptop during the game, it crashes with

> SDL Error
> LockRect(): INVALIDCALL
> The error occurred at: ..\..\SourceX\dx.cpp line 264

I have absolutely no experience with SDL, but I've managed to put
together a workaround. Improvements are welcome.

PR #3673, opened 2021-12-04
Settings: Add graphic settings that resizes the window

obligaron:
Notes:
- Add the possibility to resize the game and switch between window and fullscreen mode.
- Add the possibility to recreate upscale renderer (similar to #688 but not exactly the same).
- Added the follow settings to settings menu
  - Fullscreen
  - Fit to Screen
  - Upscale
  - Integer Scaling
  - Vertical Sync
- Fixed Scaling Quality doesn't instantly update when changed in settingsmenu,
- Settings are only active if supported (for example upscale is not supported in sdl 1).
- Added new VS configuration for SDL 1 on windows (based on the work done in #3392).
- Tested with SDL 2 and SDL 1 on windows.
- Is the groundwork for resolution changes,

### DIFF
--- a/Source/utils/display.cpp
+++ b/Source/utils/display.cpp
@@ -51,6 +51,67 @@ Uint16 GetViewportHeight()
 	return gnViewportHeight;
 }
 
+namespace {
+
+#ifndef USE_SDL1
+void CalculatePreferdWindowSize(int &width, int &height)
+{
+	SDL_DisplayMode mode;
+	if (SDL_GetDesktopDisplayMode(0, &mode) != 0) {
+		ErrSdl();
+	}
+
+	if (!sgOptions.Graphics.bIntegerScaling) {
+		float wFactor = (float)mode.w / width;
+		float hFactor = (float)mode.h / height;
+
+		if (wFactor > hFactor) {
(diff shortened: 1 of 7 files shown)


AJenbo (dismissed):

AJenbo:
Probably we should also call `ResizeWindow()` when exiting full screen to make the window the proper size and not leave it semi maximized. But that isn't directly related to this change.

obligaron:
> 

This is currently the case:
`fullscreen.SetValueChangedCallback(ResizeWindow);`

I had the same thought and tested the maximize case, 🙂

Source/utils/display.cpp
```diff
@@ -51,6 +51,67 @@ Uint16 GetViewportHeight()
...
+	gnScreenHeight = windowSize.height;
+
+	gnViewportHeight = gnScreenHeight;
+	if (gnScreenWidth <= PANEL_WIDTH) {
+		// Part of the screen is fully obscured by the UI
+		gnViewportHeight -= PANEL_HEIGHT;
+	}
+}
+
+Size GetPreferedWindowSize()
```
glebm, line 101: typo: Preferred

PR #3674, opened 2021-12-05
[mac] Add game data to container

AJenbo:
This should add the assets directly to the application container for both macOS and iOS. I did have to create dummy translation files in order to add them to the exec target, don't know if there is a better way to do that?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -656,100 +656,6 @@ if(RUN_TESTS)
     test/animationinfo_test.cpp)
 endif()
 
-add_library(libdevilutionx OBJECT ${libdevilutionx_SRCS})
-if (ANDROID)
-  add_library(${BIN_TARGET} SHARED Source/main.cpp)
-else()
-  add_executable(${BIN_TARGET}
-    WIN32
-    MACOSX_BUNDLE
-    Source/main.cpp
-    Source/devilutionx.exe.manifest
-    Packaging/macOS/AppIcon.icns
-    Packaging/windows/devilutionx.rc)
-endif()
-target_link_libraries(${BIN_TARGET} PRIVATE libdevilutionx)
-
-# Use file GENERATE instead of configure_file because configure_file
-# does not support generator expressions.
-get_property(is_multi_config GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
-if(is_multi_config)
-  set(CONFIG_PATH $<CONFIG>/config.h)
-  target_include_directories(libdevilutionx PUBLIC ${CMAKE_BINARY_DIR}/$<CONFIG>)
-else()
-  set(CONFIG_PATH config.h)
-endif()
-file(GENERATE OUTPUT ${CONFIG_PATH} CONTENT
-"#pragma once
-#define PROJECT_NAME \"${PROJECT_NAME}\"
-#define PROJECT_VERSION \"${PROJECT_VERSION}${VERSION_SUFFIX}\"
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -855,29 +761,156 @@ set(devilutionx_assets
...
   add_custom_command(
     COMMENT "Copying ${asset_file}"
     OUTPUT "${dst}"
     DEPENDS "${src}"
     COMMAND ${CMAKE_COMMAND} -E copy "${src}" "${dst}"
     VERBATIM)
+
+  if(APPLE)
+    list(APPEND BIN_TARGET_SRCS ${dst})
+    string(REGEX MATCH "(.*)\/" subpath ${asset_file})
```
glebm, line 794: Let's not strip the `assets/` prefix? Then no code changes will be required?

CMakeLists.txt
```diff
@@ -855,29 +761,156 @@ set(devilutionx_assets
...
+  list(APPEND Gettext_ROOT ${CMAKE_CURRENT_BINARY_DIR}/vcpkg_installed/${VCPKG_TARGET_TRIPLET}/tools/gettext/bin)
+endif()
+find_package(Gettext)
 if (Gettext_FOUND)
   foreach(lang ${devilutionx_langs})
     list(APPEND DEVILUTIONX_MPQ_FILES "${lang}.gmo")
+
+	if(APPLE)
+      set(dst "${DEVILUTIONX_ASSETS_OUTPUT_DIRECTORY}/${lang}.gmo")
+      file(TOUCH ${dst})
```
glebm, line 813: Whoa. I think we can simply add a dependency on lang targets to `BIN_TARGET` for mac/ios

```cmake
add_dependency(${BIN_TARGET} ${devilutionx_lang_targets} ${devilutionx_lang_files})
```

CMakeLists.txt
```diff
@@ -855,29 +761,156 @@ set(devilutionx_assets
...
     DEPENDS "${src}"
     COMMAND ${CMAKE_COMMAND} -E copy "${src}" "${dst}"
     VERBATIM)
+
+  if(APPLE)
+    list(APPEND BIN_TARGET_SRCS ${dst})
+    string(REGEX MATCH "(.*)\/" subpath ${asset_file})
+    set_source_files_properties(${dst} PROPERTIES
+      MACOSX_PACKAGE_LOCATION "Resources/${subpath}"
+      XCODE_EXPLICIT_FILE_TYPE compiled)
```
glebm, line 797: On Mac, might as well not copy the file at all but set this property on `${src}` instead (without XCODE_EXPLICIT_FILE_TYPE compiled).

CMakeLists.txt
AJenbo: I was unable to get it working with out adding them to the exec, but couldn't do so with out them existing at config time. I do realize that this is terrible solution :)

CMakeLists.txt
AJenbo: This is stripping the filename, not assets, so that we know the assets subfolder. On Mac/iOS the path is resources and not assets. But the target is only know when generating the installer. Xcode can actualy take the folder as the thing to copy, but CMake appears to be limited to taking files.

CMakeLists.txt
AJenbo: If we do not set XCODE_EXPLICIT_FILE_TYPE selecting the file in Xcode wil try to load it as text which can freeze Xcode.

AJenbo:
Hopefully your suggestions will automatically do all the things that I was figuring out how to do manually :)

PR #3675, opened 2021-12-05
Handle screen orientation

AJenbo:
Some devices may report the screen orientation in portrait mode, but we need it in landscape mode since that is what we will be configuring.

Also simplify the calculations for integer scaling (and flip the case/order).


obligaron (approved):
lgtm 🙂

AJenbo:
This appears to be required for all iOS devices and at least 1 Android device :D

PR #3676, opened 2021-12-05
macOS/iOS: Add resources to the bundle

glebm:
alternative to #3674

I have no way of testing this, so it may require further tweaking

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -744,6 +744,14 @@ if (Gettext_FOUND)
     list(APPEND devilutionx_lang_targets "${_lang_target}")
     list(APPEND devilutionx_lang_files "${_gmo_file}")
 
+    if(APPLE)
+      set_source_files_properties("${_gmo_file}" PROPERTIES
+        MACOSX_PACKAGE_LOCATION Resources
+        XCODE_EXPLICIT_FILE_TYPE compiled)
+      add_dependencies(${BIN_TARGET} "${_lang_target}")
+      target_sources(${BIN_TARGET} PRIVATE "${_gmo_file}")
+    endif()
+
     if(VITA)
       list(APPEND VITA_TRANSLATIONS_LIST "FILE" "${_gmo_file}" "assets/${lang}.gmo")
     endif()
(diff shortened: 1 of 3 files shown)


Source/engine/assets.cpp
```diff
@@ -63,8 +63,8 @@ SDL_RWops *OpenAsset(const char *filename, bool threadsafe)
 	if ((rwops = SDL_RWFromFile(path.c_str(), "rb")) != nullptr)
 		return rwops;
 
-#ifdef __ANDROID__
-	// On Android, fall back to the APK's assets.
+#if defined(__ANDROID__) || defined(__APPLE__)
+	// On Android and iOS, fall back to the APK/Bundle's assets.
```
AJenbo, line 67: This applies to all of Apples OSs (macOS, tvOS, watchOS and iOS), and hopefully we can also use this code path for other systems that has a build in package format.
```suggestion
	// On fall back to the bundled assets on supported systems.
```

CMakeLists.txt
```diff
@@ -744,6 +744,14 @@ if (Gettext_FOUND)
     list(APPEND devilutionx_lang_targets "${_lang_target}")
     list(APPEND devilutionx_lang_files "${_gmo_file}")
 
+    if(APPLE)
+      set_source_files_properties("${_gmo_file}" PROPERTIES
+        MACOSX_PACKAGE_LOCATION "Resources/assets/${lang}.gmo"
+        XCODE_EXPLICIT_FILE_TYPE compiled)
+      add_dependencies(${BIN_TARGET} "${_gmo_file}" "${_lang_target}")
+      target_sources(${BIN_TARGET} PRIVATE "${_gmo_file}")
```
AJenbo, line 752:
```

CMake Error at CMakeLists.txt:751 (add_dependencies):
  The dependency target
  "/Users/runner/work/devilutionX/devilutionX/build/assets/bg.gmo" of target
  "devilutionx" does not exist.
```

glebm: Can you please try again?

glebm: I've just force-pushed a fix to `MACOSX_PACKAGE_LOCATION` (it's a directory, not a filename)

CMakeLists.txt
```diff
@@ -855,43 +863,61 @@ set(devilutionx_assets
...
-    WORKING_DIRECTORY "${DEVILUTIONX_ASSETS_OUTPUT_DIRECTORY}"
-    DEPENDS ${DEVILUTIONX_OUTPUT_ASSETS_FILES} ${devilutionx_lang_targets} ${devilutionx_lang_files}
-    VERBATIM)
-  add_custom_target(devilutionx_mpq DEPENDS "${DEVILUTIONX_MPQ}")
-  add_dependencies(${BIN_TARGET} devilutionx_mpq)
+if(APPLE)
+  foreach(asset_file ${devilutionx_assets})
+    set(src "${CMAKE_CURRENT_SOURCE_DIR}/Packaging/resources/assets/${asset_file}")
+    set_source_files_properties("${src}" PROPERTIES
+      MACOSX_PACKAGE_LOCATION Resources/assets
```
AJenbo, line 879: This isn't going to work since the game will be looking for the files in the subfoders `data`, `ui_art`, `fonts`, `gendata`, not `assets`.

glebm: Oh, I finally understand what your regexp was doing

AJenbo: I can see it being confusing with android having them in `./asset`s, apple using `./resources` an we have them in `./resources/assets` :D

glebm: OK this should be good now, please give it a try!

AJenbo: Oh I didn't think get_filename_component would work with out a file existing in the given path :o

CMakeLists.txt
```diff
@@ -855,43 +863,62 @@ set(devilutionx_assets
...
-    DEPENDS ${DEVILUTIONX_OUTPUT_ASSETS_FILES} ${devilutionx_lang_targets} ${devilutionx_lang_files}
-    VERBATIM)
-  add_custom_target(devilutionx_mpq DEPENDS "${DEVILUTIONX_MPQ}")
-  add_dependencies(${BIN_TARGET} devilutionx_mpq)
+if(APPLE)
+  foreach(asset_file ${devilutionx_assets})
+    set(src "${CMAKE_CURRENT_SOURCE_DIR}/Packaging/resources/assets/${asset_file}")
+    get_filename_component(_asset_dir "${asset_file}" DIRECTORY)
+    set_source_files_properties("${src}" PROPERTIES
+      MACOSX_PACKAGE_LOCATION "Resources/assets/${_asset_dir}"
```
AJenbo, line 880: Close:
```suggestion
      MACOSX_PACKAGE_LOCATION "Resources/${_asset_dir}"
```

CMakeLists.txt
```diff
@@ -744,6 +744,14 @@ if (Gettext_FOUND)
     list(APPEND devilutionx_lang_targets "${_lang_target}")
     list(APPEND devilutionx_lang_files "${_gmo_file}")
 
+    if(APPLE)
+      set_source_files_properties("${_gmo_file}" PROPERTIES
+        MACOSX_PACKAGE_LOCATION Resources/assets
```
AJenbo, line 749:
```suggestion
        MACOSX_PACKAGE_LOCATION Resources
```

CMakeLists.txt
glebm: That would result in assets/assets/
See the build log here, the paths are already correct: https://github.com/diasurgical/devilutionX/runs/4423597286?check_suite_focus=true

CMakeLists.txt
glebm: That would result in Resources/en.gmo instead of Resources/assets/en.gmo

CMakeLists.txt
glebm: Ah, nvm, we *don't* want to place them under `assets/`

glebm: Fixed

AJenbo: > That would result in assets/assets/

Is it currently assets/assets/assets/?

The log says that they are in `Resources/assets/data/boxleftend.pcx`, but they should be in `Resources/data/boxleftend.pcx`

PR #3678, opened 2021-12-05
Correct DPI on devices that apply extra scaling

AJenbo:
on iOS the windows size is reported in a size that matches the DPI of the original iPhone rather then the actual DPI. Getting the render size should give us the correct scale for all devices.

Hopefully this also works correctly on the iPhone 8+ which adds another layer of craziness by rendering at a higher resolution then the actually pixel count and then reports a fraction of that as it's windows size...
https://www.paintcodeapp.com/news/ultimate-guide-to-iphone-resolutions

### DIFF
--- a/Source/controls/touch/gamepad.cpp
+++ b/Source/controls/touch/gamepad.cpp
@@ -62,7 +62,11 @@ void InitializeVirtualGamepad()
 	if (SDL_GetDisplayDPI(displayIndex, nullptr, &hdpi, &vdpi) == 0) {
 		int clientWidth;
 		int clientHeight;
-		SDL_GetRendererOutputSize(renderer, &clientWidth, &clientHeight);
+		if (renderer != nullptr)
+			SDL_GetRendererOutputSize(renderer, &clientWidth, &clientHeight);
+		else
+			SDL_GetWindowSize(ghMainWnd, &clientWidth, &clientHeight);
+
 		hdpi *= static_cast<float>(gnScreenWidth) / clientWidth;
 		vdpi *= static_cast<float>(gnScreenHeight) / clientHeight;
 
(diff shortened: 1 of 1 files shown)


Source/controls/touch/gamepad.cpp
```diff
@@ -62,7 +62,7 @@ void InitializeVirtualGamepad()
 	if (SDL_GetDisplayDPI(displayIndex, nullptr, &hdpi, &vdpi) == 0) {
 		int clientWidth;
 		int clientHeight;
-		SDL_GetWindowSize(ghMainWnd, &clientWidth, &clientHeight);
+		SDL_GetRendererOutputSize(renderer, &clientWidth, &clientHeight);
```
StephenCWills, line 65: `renderer` could be null if `Upscale=0`.

```suggestion
		if (renderer != nullptr)
			SDL_GetRendererOutputSize(renderer, &clientWidth, &clientHeight);
		else
			SDL_GetWindowSize(ghMainWnd, &clientWidth, &clientHeight);

```

StephenCWills (changes requested):

Source/controls/touch/gamepad.cpp
AJenbo: Oh, I thought the renderer was was required for drawing the touch UI :)

StephenCWills: Nah, I was going require upscaling since touch controls don't work on SDL1 anyway. But when you asked me a question about making the virtual gamepad depend on a tangentially related setting, I gave it a bit more thought and discovered a relatively easy way to make it work with `SDL_BlitScaled()`. It looks awful, of course, but it functions.

PR #3679, opened 2021-12-05
Settingsmenu: Add list browsing

obligaron:
For option entries with type `OptionEntryType::List` we now show a browsing with all selected values for this option.
This avoids to reset the UI every time we browse to the desired value. When adding resolution selection this is especially important (not part of this pr).

Sample Video:

https://user-images.githubusercontent.com/25415264/144754148-9e5308ca-1084-4164-8cfc-2d764f306e03.mp4


AJenbo (approved):
Some day we should rewrite all of this in to standard widgets :)

PR #3684, opened 2021-12-06
Use distro zlib package for MinGW instead of SDL-ttf zlib dll

### LINKED ISSUE #3037: Remove SDL_ttf dependency from mingw builds
Currently doing so results in the zlib target missing and thus the build fails, so some adjustments are needed, see: https://github.com/diasurgical/devilutionX/pull/3019

StephenCWills:
Modifies the MinGW build scripts and instructions to use the `libz-mingw-w64-dev` package instead of the zlib dll provided by SDL-ttf. I also added `-DDEVILUTIONX_SYSTEM_BZIP2=OFF` to the build instructions since I couldn't find a package for it.

This resolves #3037

### DIFF
--- a/.github/workflows/Windows_x64.yml
+++ b/.github/workflows/Windows_x64.yml
@@ -22,13 +22,6 @@ jobs:
         sudo apt install -y cmake gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 pkg-config-mingw-w64-x86-64 libz-mingw-w64-dev dpkg-dev wget git sudo smpq &&
         sudo Packaging/windows/mingw-prep64.sh
 
-    - name: Cache CMake build folder
-      uses: actions/cache@v2
-      with:
-        path: build
-        key: windows-x86_64-cmake-v2-${{ github.sha }}
-        restore-keys: windows-x86_64-cmake-v2-
-
     - name: Configure CMake
       shell: bash
       working-directory: ${{github.workspace}}
--- a/.github/workflows/Windows_x64.yml
+++ b/.github/workflows/Windows_x64.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Create Build Environment
       run: >
         sudo apt update &&
-        sudo apt install -y cmake gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 pkg-config-mingw-w64-x86-64 dpkg-dev wget git sudo smpq &&
+        sudo apt install -y cmake gcc-mingw-w64-x86-64 g++-mingw-w64-x86-64 pkg-config-mingw-w64-x86-64 libz-mingw-w64-dev dpkg-dev wget git sudo smpq &&
         sudo Packaging/windows/mingw-prep64.sh
 
     - name: Cache CMake build folder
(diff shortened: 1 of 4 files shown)


StephenCWills:
I suspect this will only work properly if we can get pkg-config to work on the CI runners. Here are a couple relevant bug reports that might help...

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=967969
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=930492

Trihedraf:
I just pushed that change to my repo and it ran on the ci just fine. [log](https://github.com/Trihedraf/devilutionX/actions/runs/1543872618). Maybe we can try re-triggering it here?

StephenCWills:
> I just pushed that change to my repo and it ran on the ci just fine. [log](https://github.com/Trihedraf/devilutionX/actions/runs/1543872618). Maybe we can try re-triggering it here?

Looks like it's the same error as last time. Maybe the cached build folder is causing this to fail.

Trihedraf:
Must be. It doesn't even look like an error that has to do with the change.

StephenCWills:
I was able to reproduce the error on a local Ubuntu instance from a clean working directory by checking out the parent commit, following the CI build steps, clearing the MinGW packages from the system, checking out this PR, and following the build steps again. My CMakeCache.txt file has a number of references to zlib1.dll which would likely need to be purged or updated.

The same system has no problems if I don't use the parent commit and just build this from scratch. This basically confirms that the CI errors are due to the cached build folder.

@glebm You mentioned recently you were able to detect and upgrade from a previous CMake cache (https://github.com/diasurgical/devilutionX/pull/3691#issuecomment-987289185). Would that approach be applicable here?

glebm:
@StephenCWills My approach in that PR is quite specific to the PR, as it detect an invalid combination of state (missing system dependency with `DEVILUTIONX_*_SYSTEM=ON`) that was possible before the PR.

The issue here might be caused by one of the bugs resolved by #3691 (not defaulting to `STATIC` when `DIST` is `ON`)).
Can you please try passing `-DDEVILUTIONX_STATIC_BZIP2=ON`?

maoa3:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

StephenCWills:
> My approach in that PR is quite specific to the PR...

I thought that might be the case, but no harm in asking I hope.

> The issue here might be caused by one of the bugs resolved by #3691 (not defaulting to `STATIC` when `DIST` is `ON`)). Can you please try passing `-DDEVILUTIONX_STATIC_BZIP2=ON`?

Static bzip2 doesn't appear to make any difference. I'm quite certain the issue is due a mismatch between the name of the zlib static library before and after this change.

From CMakeCache.txt:
```txt
ZLIB_LIBRARY_RELEASE:FILEPATH=/usr/x86_64-w64-mingw32/bin/zlib1.dll
png_static_LIB_DEPENDS:STATIC=general;/usr/x86_64-w64-mingw32/bin/zlib1.dll;
FIND_PACKAGE_MESSAGE_DETAILS_ZLIB:INTERNAL=[/usr/x86_64-w64-mingw32/bin/zlib1.dll][/usr/x86_64-w64-mingw32/include][v1.2.11()]
```

After removing those entries, the build is successful:
`-- Found ZLIB: /usr/x86_64-w64-mingw32/lib/libz.dll.a (found version "1.2.11")`

I'm not sure what would be an appropriate thing to do here.

AJenbo:
Hmm the Windows builds appear to take roughly the same amount of time with and with out caching so I don't mind just leaving it out.

PR #3688, opened 2021-12-06
Settings: add resolution option

obligaron:
This pr allows to change the resolution ingame.

Notes:
- List with resolutions if populated by monitor resolutions (`SDL_ListModes` / `SDL_GetDisplayMode`).
- if `ini` contains a resolution that is not part of the monitor resolutions it is preserved.
- Default resolution (640x480) is always added explicitly.
- Tested with SDL 2 and SDL 1 on windows.

grafik

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -299,8 +299,6 @@ void LoadOptions()
 	sgOptions.Audio.nBufferSize = GetIniInt("Audio", "Buffer Size", DEFAULT_AUDIO_BUFFER_SIZE);
 	sgOptions.Audio.nResamplingQuality = GetIniInt("Audio", "Resampling Quality", DEFAULT_AUDIO_RESAMPLING_QUALITY);
 
-	sgOptions.Graphics.nWidth = GetIniInt("Graphics", "Width", DEFAULT_WIDTH);
-	sgOptions.Graphics.nHeight = GetIniInt("Graphics", "Height", DEFAULT_HEIGHT);
 	sgOptions.Graphics.nGammaCorrection = GetIniInt("Graphics", "Gamma Correction", 100);
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	sgOptions.Graphics.bHardwareCursor = GetIniBool("Graphics", "Hardware Cursor", HardwareCursorDefault());
@@ -354,8 +352,6 @@ void SaveOptions()
 	SetIniValue("Audio", "Channels", sgOptions.Audio.nChannels);
 	SetIniValue("Audio", "Buffer Size", sgOptions.Audio.nBufferSize);
 	SetIniValue("Audio", "Resampling Quality", sgOptions.Audio.nResamplingQuality);
-	SetIniValue("Graphics", "Width", sgOptions.Graphics.nWidth);
-	SetIniValue("Graphics", "Height", sgOptions.Graphics.nHeight);
 	SetIniValue("Graphics", "Gamma Correction", sgOptions.Graphics.nGammaCorrection);
 #if SDL_VERSION_ATLEAST(2, 0, 0)
 	SetIniValue("Graphics", "Hardware Cursor", sgOptions.Graphics.bHardwareCursor);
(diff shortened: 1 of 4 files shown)


Source/options.h
```diff
@@ -216,6 +216,31 @@ class OptionEntryLanguageCode : public OptionEntryListBase {
...
+	[[nodiscard]] virtual size_t GetActiveListIndex() const override;
+	virtual void SetActiveListIndex(size_t index) override;
+
+	Size operator*() const
+	{
+		return size;
+	}
+
+private:
+	/** @brief Language code (IETF) for text. */
```
AJenbo, line 237:
```suggestion
	/** @brief View size. */
```

ephphatha (commented):
For some reason none of the graphic settings that call `ReinitializeRenderer` play nice on my system (Windows 11), ends up with a stuck frame and the display doesn't update. It also affects fullscreen toggling so not something introduced in this PR.

Source/options.h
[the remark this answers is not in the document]
AJenbo: Also the language codes we use are not in IETF :P

Source/options.cpp
```diff
@@ -572,6 +568,90 @@ std::vector<OptionEntryBase *> AudioOptions::GetEntries()
...
+		SDL_DisplayMode mode;
+		if (SDL_GetDisplayMode(0, i, &mode) != 0) {
+			ErrSdl();
+		}
+		sizes.emplace_back(Size { mode.w, mode.h });
+	}
+#endif
+
+	// Ensures that the ini specified resolution is present in resolution list even if it doesn't match a monitor resolution (for example if played in window mode)
+	sizes.emplace_back(this->size);
```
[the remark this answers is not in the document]
obligaron: Fixed 🙂

Source/options.cpp
```diff
@@ -572,6 +568,90 @@ std::vector<OptionEntryBase *> AudioOptions::GetEntries()
...
+		return;
+
+	std::vector<Size> sizes;
+
+	// Add resolutions
+#ifdef USE_SDL1
+	auto *modes = SDL_ListModes(nullptr, SDL_FULLSCREEN | SDL_HWPALETTE);
+	if (modes != nullptr) {
+		for (size_t i = 0; modes[i] != nullptr; i++) {
+			sizes.emplace_back(Size { modes[i]->w, modes[i]->h });
```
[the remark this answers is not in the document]
obligaron: Didn't changed for the moment. Is that okay?

Source/options.h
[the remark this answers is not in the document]
obligaron: Yeah copy&paste. 😢 
Fixed.

Source/options.h
```diff
@@ -216,6 +216,31 @@ class OptionEntryLanguageCode : public OptionEntryListBase {
...
+public:
+	OptionEntryResolution();
+
+	void LoadFromIni(string_view category) override;
+	void SaveToIni(string_view category) const override;
+
+	[[nodiscard]] virtual size_t GetListSize() const override;
+	[[nodiscard]] virtual string_view GetListDescription(size_t index) const override;
+	[[nodiscard]] virtual size_t GetActiveListIndex() const override;
+	virtual void SetActiveListIndex(size_t index) override;
```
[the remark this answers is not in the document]
obligaron: Fixed and opened #3693 🙂

Source/options.h
obligaron: Fixed 🙂

obligaron:
> For some reason none of the graphic settings that call `ReinitializeRenderer` play nice on my system (Windows 11), ends up with a stuck frame and the display doesn't update. It also affects fullscreen toggling so not something introduced in this PR.

Interesting. On my windows 10 machine (amd cpu&graphic) it works fine.
Does this also happen for scaling quality? Asking cause this is the option that has the least impact (only changes a hint).
Does this system still works (do you hear a sound when you move the cursor up/down)?

Source/options.cpp
```diff
@@ -572,6 +568,90 @@ std::vector<OptionEntryBase *> AudioOptions::GetEntries()
...
+		sizes.emplace_back(Size { mode.w, mode.h });
+	}
+#endif
+
+	// Ensures that the ini specified resolution is present in resolution list even if it doesn't match a monitor resolution (for example if played in window mode)
+	sizes.push_back(this->size);
+	// Ensures that the vanilla/default resolution is always present
+	sizes.emplace_back(Size { DEFAULT_WIDTH, DEFAULT_HEIGHT });
+
+	// Sort by width then by height
```
AJenbo, line 616: Might be better to do it by height first since most designated resolutions by height alone (480p 720p 1080p 1440p etc)

obligaron: I tried to mimic how my os (windows 10) and starcraft 2 presents me the resolution list:
grafik grafik
There it is sorted by width first and then by height.

But I don't have a strong opinion on this, if you want I can change it. 🙂

AJenbo: Damn now I want to play StarCraft, good thing it works properly via Wine do I don't have to RE and port it first :D

Source/options.cpp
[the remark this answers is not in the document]
glebm: It's fine with `push_back(Size {})` IMO. Note that either one will generate the same assembly.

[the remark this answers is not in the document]
glebm: Please have a look at this article regarding emplace vs push_back in general https://abseil.io/tips/112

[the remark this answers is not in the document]
obligaron: Thanks that made it clear. 🙂

ephphatha:
> > For some reason none of the graphic settings that call `ReinitializeRenderer` play nice on my system (Windows 11), ends up with a stuck frame and the display doesn't update. It also affects fullscreen toggling so not something introduced in this PR.
> 
> Interesting. On my windows 10 machine (amd cpu&graphic) it works fine. Does this also happen for scaling quality? Asking cause this is the option that has the least impact (only changes a hint). Does this system still works (do you hear a sound when you move the cursor up/down)?

This is on a ryzen + 1080 system. Sound still plays and the game runs as normal (with no visual feedback). The menu is still navigable with the keyboard (I can quit out of the game or toggle options if I remember where they are). And this is only an issue if Upscale is On. If I turn that setting off the new resolution renders fine. I can also launch devX with Upscale Off, change it to On and it'll work until I change resolution/fullscreen mode or toggle upscale off then on again in the same session. I can always turn upscale off to get it to render properly.

Source/options.cpp
AJenbo: I guess if it was me designing the world i would group them by aspect as well

Source/options.cpp
```diff
@@ -572,6 +568,90 @@ std::vector<OptionEntryBase *> AudioOptions::GetEntries()
...
+	if (!resolutions.empty())
+		return;
+
+	std::vector<Size> sizes;
+
+	// Add resolutions
+#ifdef USE_SDL1
+	auto *modes = SDL_ListModes(nullptr, SDL_FULLSCREEN | SDL_HWPALETTE);
+	if (modes != nullptr) {
+		for (size_t i = 0; modes[i] != nullptr; i++) {
```
StephenCWills, line 596: This crashes on 3DS. We're dealing with an.. *interesting* error code.

https://www.libsdl.org/release/SDL-1.2.15/docs/html/sdllistmodes.html
```c++
/* Check if our resolution is restricted */
if(modes == (SDL_Rect **)-1){
  printf("All resolutions available.\n");
}
```

StephenCWills (changes requested):

StephenCWills:
On my Windows 10 system (NVIDIA), I'm getting severe issues any time I select a resolution that's larger than the initial resolution when the game was launched. Here's a screenshot taken by going from 640x480 to 768x480. FYI, the same thing happens if `Fit to Screen` is initially off and I turn it on.

image

Source/options.cpp
obligaron: Very interesting.
I adjusted the code. Could you take a look again?

StephenCWills: This works. With this change, I can toggle between 800x480 and 640x480. Changing the resolution absolutely destroys the quality of the graphics for some reason, but that's probably only tangentially related to this PR, like the `PalSurface` issue. It looks like scaling artifacts.

image

AJenbo: It might be performing the scaling early in software instead of late in hardware.

AJenbo (approved):

PR #3691, opened 2021-12-06
CMake: Overhaul dependency options

glebm:
Changes how dependency options are handled and moves them out of `CMakeLists.txt`.

Major changes:

1. The static default now takes `TARGET_SUPPORTS_SHARED_LIBS` into account
2. The `DIST` setting no longer affects the static default.
3. The static option is now visible for system dependencies as well (it was previously hidden by `cmake_dependent_option`).
4. The auto-detection mechanism for `libfmt`, `SDL_image`, and `libpng` only applies if the setting has not been set.

When using non-system dependencies, the default should always be static, as there are 2 main use cases for these:

1. Dependencies that we must currently vendor because our forks have significant changes that are not yet merged/released upstream.
2. Platforms that have no system packages, such as Android and iOS.

In both cases, static linking is the most appropriate default.


glebm:
@AJenbo I think this is good to go now but the build cache is causing the build to fail.
We now **require** a system libfmt if `DEVILUTIONX_SYSTEM_LIBFMT` is `ON`, whereas the previous version did auto-detection in that case.

AJenbo:
> the build cache is causing the build to fail.

There are two options, add `-v2` to the cache key of the build job definitions. Or add an env that we can change each time we need the build folder to be cleared between two version.

I prefer the first for a couple of reason:
- it's simple
- it follows the change

glebm:
@AJenbo I think I've figured out a way to detect and upgrade previous inconsistent CMake cache.
This is also good because developers' caches will keep working. A few weeks after merging we can remove the upgrade code.

3rdParty/SDL_image/CMakeLists.txt
```diff
@@ -1,18 +1,24 @@
-option(DEVILUTIONX_SYSTEM_LIBPNG "Use system-provided libpng" ON)
-cmake_dependent_option(DEVILUTIONX_STATIC_LIBPNG "Link static libpng" OFF
-                      "DEVILUTIONX_SYSTEM_LIBPNG AND NOT DIST" ON)
+include(dependency_options)
 
-if(DEVILUTIONX_SYSTEM_LIBPNG)
+# In older versions of DevilutionX, libpng could be built from
```
AJenbo, line 3:
```suggestion
# In previous versions of DevilutionX, libpng could be built from
```

AJenbo:
just 1 to go :)

CMake/DevilutionX_Dependencies.cmake
```diff
@@ -0,0 +1,179 @@
...
+if(DEVILUTIONX_SYSTEM_BZIP2)
+  find_package(BZip2 REQUIRED)
+else()
+  add_subdirectory(3rdParty/bzip2)
+endif()
+
+add_subdirectory(3rdParty/libsmackerdec)
+
+if(WIN32)
+add_subdirectory(3rdParty/find_steam_game)
```
AJenbo, line 154:
```suggestion
  add_subdirectory(3rdParty/find_steam_game)
```

glebm:
> just 1 to go :)

Android was tricky. Turns out the `ANDROID` variable is only defined after the `project` call. However, `CMAKE_SYSTEM_NAME STREQUAL Android` is always available.

glebm:
Gave up on including platform definitions before defining the options, should be OK now

glebm:
Well, this was the source of my troubles:

```cmake
# Platform definitions can override options and we want `cmake_dependent_option` to see the effects,
# so ideally we would include Platforms.cmake before definining the options.
#
# However, `Platforms` require `project` to have been called (to get access to toolchain defs),
# but `project` must be called after `VcPkgManifestFeatures`, and `VcPkgManifestFeatures` need
# to be after the options.
```

glebm:
Also we now link SDL2 on Android statically, resulting in a 1.1 MiB smaller APK

AJenbo:
Doesn't look relevant :)

PR #3696, opened 2021-12-06
Parallel build of iOS port

AJenbo:
Lets see if this works now.


Trihedraf:
Don't know if it's related, but it won't install on the altstore saying ipa provided no bundle, can't install

AJenbo:
should be fixed now

PR #3699, opened 2021-12-07
Update iOS.yml

AJenbo:


AJenbo:
@Trihedraf this fixes the ipa, thanks for mentioning it

PR #3702, opened 2021-12-07
IsWallSpell

qndel:


Source/spells.h
```diff
@@ -16,6 +16,7 @@ enum class SpellCheckResult : uint8_t {
 	Fail_Busy,
 };
 
+bool IsWallSpell(const spell_id spl);
```
glebm, line 19: nit: There is no need to specify `const` here. It is passed by value (copied) and what the function does with the copy is its own business.

glebm (approved):

Source/spells.h
AJenbo:
```suggestion
bool IsWallSpell(spell_id spl);
```

Source/spells.cpp
```diff
@@ -108,6 +108,11 @@ void PlacePlayer(int pnum)
 
 } // namespace
 
+bool IsWallSpell(const spell_id spl)
```
AJenbo, line 111:
```suggestion
bool IsWallSpell(spell_id spl)
```

AJenbo (approved):

Source/spells.h
qndel: right : D damn I wanted to use const for the first time

PR #3704, opened 2021-12-07
CMake: Organize files 

glebm:
1. Platform and toolchain files are now all under `platforms/`, with a single `platforms/${platform}.cmake` per platform.

2. Custom functions/macros are under `functions/`.

3. Finder modules are in `finders/`.


AJenbo:
Maybe move cpp/h from CMake/switch to Source/platform/switch ?

AJenbo (approved):

AJenbo:
image

Issue #225, opened 2019-08-07
Could NOT find Sodium (missing: sodium_LIBRARY_RELEASE sodium_LIBRARY_DEBUG)


Polynomial-C:
Since commit 1f7f5bf3dccee957e23bdf4d976883b155d5e81 I cannot build the game anymore on Linux. Reverting this commit "fixes" the issue.

OrangBuangan:
Can you tell more about your setup?

Polynomial-C:
What information do you want to know? It's a 64bit Gentoo Linux but I'm trying to build a 32bit executable and all dependencies have the 32bit parts installed as well.

Polynomial-C:
Alright, here's some information about this system:
```
# emerge --info libsodium libsdl sdl2-mixer sdl2-ttf 
Portage 2.3.71 (python 3.6.9-final-0, default/linux/amd64/17.1/desktop/plasma, gcc-8.3.0, glibc-2.29-r2, 4.19.65 x86_64)
=================================================================
                         System Settings
=================================================================
System uname: Linux-4.19.65-x86_64-AMD_Opteron-tm-_Processor_6328-with-gentoo-2.6
KiB Mem:   131999304 total, 100710164 free
KiB Swap:          0 total,         0 free
Timestamp of repository gentoo: Thu, 08 Aug 2019 06:00:01 +0000
Head commit of repository gentoo: 7330a6332370d71843f98b290e166f2294958257
Timestamp of repository poly-c: Thu, 08 Aug 2019 11:05:02 +0000
sh dash 0.5.10.2
ld GNU ld (Gentoo 2.32 p2) 2.32.0
app-shells/bash:          5.0_p7::gentoo
dev-java/java-config:     2.2.0-r4::gentoo
dev-lang/perl:            5.30.0::gentoo
dev-lang/python:          2.7.16::gentoo, 3.6.9::gentoo, 3.7.4-r1::gentoo
dev-util/cmake:           3.15.1::gentoo
sys-apps/baselayout:      2.6-r1::poly-c
sys-apps/openrc:          0.41.2::gentoo
sys-apps/sandbox:         2.18::gentoo
sys-devel/autoconf:       2.13-r1::gentoo, 2.69-r4::gentoo
sys-devel/automake:       1.13.4-r2::gentoo, 1.16.1-r1::gentoo
sys-devel/binutils:       2.32-r1::gentoo
sys-devel/gcc:            8.3.0-r1::gentoo
[...]

PR #3709, opened 2021-12-08
Remove NIGHTLY_BUILD and BINARY_RELEASE options

StephenCWills:
Removing these options is pretty easy since all the code that explicitly checked them was already checking `CMAKE_BUILD_TYPE` as well. The main thing was updating build scripts and docs to reflect the change.

I also took the liberty of decoupling the `Release`/`RelWithDebInfo` build types from the `DIST` and `CPACK` options. Recent changes have rendered the `DIST` option all but useless so I updated the description to reflect (as best I can figure) [what it actually does](https://github.com/StephenCWills/devilutionX/blob/f102b766d8389f729db9ab2aca99e94d2142f707/CMakeLists.txt#L961-L967). I also reviewed all usages of `NIGHTLY_BUILD` and `BINARY_RELEASE` to determine which ones should retain the `CPACK` flag and manually added `-DCPACK` as appropriate.

Heavily inspired by https://github.com/diasurgical/devilutionX/issues/225#issuecomment-963406590


StephenCWills:
Hmm, it didn't fix this, though. Converting to draft...

image

StephenCWills:
The libsodium dll is dynamically linked with the `libgcc_s_seh-1.dll` so I added `-DDEVILUTIONX_STATIC_LIBSODIUM=ON` to the MinGW build scripts. I also had a hell of a time trying to get libpng to statically link to zlib, and eventually decided to simply force it by deleting `libz.dll.a` once I ran out of ideas.

AJenbo:
LTGM, but would like @glebm to take a quick look he line for deleting the dll.a, at and see if it makes sens:

https://github.com/diasurgical/devilutionX/pull/3709/files#diff-f508c675d41bbae3ad403b6eaac2989a0d2ad5c14c7a06221bf82fc3ccdd6225R23

glebm (dismissed):
The zlib.dll.a seems like an acceptable workaround for now. I'll look into it a bit more at some point.

CMakeLists.txt
```diff
@@ -40,9 +40,7 @@ option(GPERF "Build with GPerfTools profiler" OFF)
 cmake_dependent_option(GPERF_HEAP_FIRST_GAME_ITERATION "Save heap profile of the first game iteration" OFF "GPERF" OFF)
 option(DISABLE_LTO "Disable link-time optimization (by default enabled in release mode)" OFF)
 option(PIE "Generate position-independent code" OFF)
-option(DIST "Dynamically link only glibc and SDL2" OFF)
-option(BINARY_RELEASE "Enable options for binary release" OFF)
-option(NIGHTLY_BUILD "Enable options for nightly build" OFF)
+option(DIST "Create app bundle for MacOSX/iOS" OFF)
```
glebm, line 43: Should this perhaps be ON by default?
Also, is it used only in 1 place? How about removing this as an option entirely?

AJenbo: It only applies for macOS, but requires sudo ... maybe we can switch it to a CPACK check instead?

StephenCWills: Does it make sense to check CPACK there when the script sets CPACK just two lines after?

https://github.com/StephenCWills/devilutionX/blob/f102b766d8389f729db9ab2aca99e94d2142f707/CMakeLists.txt#L970

StephenCWills: > It only applies for macOS, but requires sudo...

Actually, can you elaborate on this a bit? That code appears to run for both macOS and iOS. Does it simply do nothing for iOS?

Also, since it requires sudo, do we want to leave it as an option? Based on the documentation, it looks like this command makes the bundle standalone and portable to other devices. Maybe we can just rename it.

```suggestion
option(MACOSX_STANDALONE_APP_BUNDLE "Generate a portable app bundle to use on other devices (requires sudo)" OFF)
```

AJenbo: That's probably a good place to start :) Lets, make that change and then merge this.

StephenCWills: I renamed the variable and double-checked MacOSX.yml. It wasn't using `BINARY_RELEASE` or `CMAKE_BUILD_TYPE` but rather setting the `BUILD_TYPE` environment variable so I missed it the first time. Also, it occurs to me that maybe `-DCPACK=ON` is unnecessary since it is defined as a `RELEASE_OPTION()`.

AJenbo (approved):

PR #3710, opened 2021-12-08
add variable for display name and set to DevilutionX

Trihedraf:
Currently the name on the iOS app is set to devilutionx. I think it looks much better with the captials as DevilutionX.


AJenbo (approved):
Thanks though about it at some point but got distracted 😅

glebm:
@AJenbo The checks seems stuck :thinking:

PR #3711, opened 2021-12-08
Implement Discord rich presence

heinermann:
Rebased and refactored from original PR: https://github.com/diasurgical/devilutionX/pull/968 . See that PR for extending Discord integration beyond this PR.

## Features
- Discord rich presence, showing the following:
  - "In Menu" when in menus.
  - Line 1: Character level+class, dungeon type+level
  - Line 2: Difficulty
  - Line 3: Elapsed time in game instance/menu (elapsed time resets when entering new menu or new game instance)
  - Character image asset showing the character, equipment, and weapons being wielded the same as in Diablo. (108/150 assets used)

## Notes
~~- Disabled by default, need conditions for enabling so it only works on PC systems~~

## Differences from old PR
- New clang-format shenanigans.
- Anything related to Discord game invitations/sharing removed.
- Utilize string inputs from elsewhere, where possible.
- Favour new-age stuff like std::array.
- Use fmt and translations.
- Fix a bug where being in-menu wouldn't register correctly.

## Screenshots
2021-12-07 10_19_35-🛠starcraft-tools - Discord
2021-12-07 10_19_52-🛠starcraft-tools - Discord
2021-12-07 10_20_26-🛠starcraft-tools - Discord

### DIFF
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -133,18 +133,13 @@ if(NOT NONET AND NOT DISABLE_ZERO_TIER)
 endif()
 
 if(DISCORD_INTEGRATION)
-  message("Retrieving Discord SDK")
   FetchContent_Declare(discordsrc
     URL https://dl-game-sdk.discordapp.net/2.5.6/discord_game_sdk.zip
     URL_HASH MD5=f86f15957cc9fbf04e3db10462027d58
     SOURCE_DIR 3rdParty/Discord
   )
 
-  FetchContent_Populate(discordsrc)
-
-  if(NOT discordsrc_POPULATED)
-    message(FATAL_ERROR "Failed to retrieve Discord SDK.")
-  endif()
+  FetchContent_MakeAvailableExcludeFromAll(discordsrc)
 
   file(GLOB discord_SRCS ${discordsrc_SOURCE_DIR}/cpp/*.cpp)
   add_library(discord STATIC ${discord_SRCS})
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -131,3 +131,35 @@ endif()
 if(NOT NONET AND NOT DISABLE_ZERO_TIER)
   add_subdirectory(3rdParty/libzt)
 endif()
+
+if(DISCORD_INTEGRATION)
+  message("Retrieving Discord SDK")
+  FetchContent_Declare(discordsrc
+    URL https://dl-game-sdk.discordapp.net/2.5.6/discord_game_sdk.zip
+    URL_HASH MD5=f86f15957cc9fbf04e3db10462027d58
+    SOURCE_DIR 3rdParty/Discord
+  )
+
+  FetchContent_Populate(discordsrc)
+
+  if(NOT discordsrc_POPULATED)
+    message(FATAL_ERROR "Failed to retrieve Discord SDK.")
+  endif()
+
+  file(GLOB discord_SRCS ${discordsrc_SOURCE_DIR}/cpp/*.cpp)
+  add_library(discord STATIC ${discord_SRCS})
+
+  target_include_directories(discord INTERFACE "${discordsrc_SOURCE_DIR}/..")
+  if (CMAKE_SIZEOF_VOID_P EQUAL 4)
+    set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86")
+  else()
+    set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86_64")
+  endif()
+
+  find_library(DISCORD_LIB discord_game_sdk${CMAKE_SHARED_LIBRARY_SUFFIX} ${DISCORD_LIB_DIR})
+
+  add_library(discord_game_sdk STATIC IMPORTED)
+  set_property(TARGET discord_game_sdk PROPERTY IMPORTED_LOCATION ${DISCORD_LIB})
+
+  file(COPY "${DISCORD_LIB_DIR}/discord_game_sdk${CMAKE_SHARED_LIBRARY_SUFFIX}" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+endif()
--- a/CMake/Dependencies.cmake
+++ b/CMake/Dependencies.cmake
@@ -133,28 +133,5 @@ if(NOT NONET AND NOT DISABLE_ZERO_TIER)
 endif()
 
 if(DISCORD_INTEGRATION)
-  FetchContent_Declare(discordsrc
-    URL https://dl-game-sdk.discordapp.net/2.5.6/discord_game_sdk.zip
-    URL_HASH MD5=f86f15957cc9fbf04e3db10462027d58
-    SOURCE_DIR 3rdParty/Discord
-  )
-
-  FetchContent_MakeAvailableExcludeFromAll(discordsrc)
-
-  file(GLOB discord_SRCS ${discordsrc_SOURCE_DIR}/cpp/*.cpp)
-  add_library(discord STATIC ${discord_SRCS})
-
-  target_include_directories(discord INTERFACE "${discordsrc_SOURCE_DIR}/..")
-  if (CMAKE_SIZEOF_VOID_P EQUAL 4)
-    set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86")
-  else()
-    set(DISCORD_LIB_DIR "${discordsrc_SOURCE_DIR}/lib/x86_64")
-  endif()
-
-  find_library(DISCORD_LIB discord_game_sdk${CMAKE_SHARED_LIBRARY_SUFFIX} ${DISCORD_LIB_DIR})
-
-  add_library(discord_game_sdk STATIC IMPORTED)
-  set_property(TARGET discord_game_sdk PROPERTY IMPORTED_LOCATION ${DISCORD_LIB})
-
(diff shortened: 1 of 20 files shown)


CMake/Dependencies.cmake
```diff
@@ -151,3 +151,35 @@ endif()
 if(NOT NONET AND NOT DISABLE_ZERO_TIER)
   add_subdirectory(3rdParty/libzt)
 endif()
+
+if(DISCORD_INTEGRATION)
+  message("Retrieving Discord SDK")
```
glebm, line 156: Please move this to `3rdParty/discord` and use `FetchContent_MakeAvailableExcludeFromAll`, similar to how we handle other fetched dependencies. No need for any `message` calls btw. `MakeAvailable` will fail with a fatal error if the download fails. Download logging can be turned on globally by the user with `-DFETCHCONTENT_QUIET=OFF`.

Source/discord/discord.h
```diff
@@ -0,0 +1,63 @@
+#pragma once
+
+#ifdef DISCORD
+
+#include <cstdint>
+#include <string>
+
+#include "../gendung.h"
+#include "../player.h"
```
glebm, line 9:
```suggestion
#include "gendung.h"
#include "player.h"
```

Source/discord/discord.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <sstream>
+#include <string>
+#include <tuple>
+
+#include <fmt/format.h>
+
+#include "../gendung.h"
+#include "../panels/charpanel.hpp"
+#include "../player.h"
+#include "../utils/language.h"
```
glebm, line 19:
```suggestion
#include "gendung.h"
#include "panels/charpanel.hpp"
#include "player.h"
#include "utils/language.h"
```

glebm (changes requested):

Source/discord/discord.cpp
```diff
@@ -0,0 +1,176 @@
...
+// App ID used when someone launches classic Diablo
+constexpr discord::ClientId DiscordDiabloAppId = 496571953147150354;
+
+// App ID used for DevilutionX's Diablo
+constexpr discord::ClientId DiscordDevilutionxAppId = 795760213524742205;
+
+namespace devilution {
+
+namespace {
+constexpr std::array<const char *, 3> DifficultyStrs = { N_("Normal"), N_("Nightmare"), N_("Hell") };
```
ephphatha, line 30: This is only used inside `GetStateString`, can be moved inside that function.

Source/discord/discord.cpp
```diff
@@ -0,0 +1,176 @@
...
+constexpr std::array<const char *, DTYPE_LAST + 1> DungeonStrs = { N_("Town"), N_("Cathedral"), N_("Catacombs"), N_("Caves"), N_("Hell"), N_("Nest"), N_("Crypt") };
```
ephphatha, line 31: Same here for `GetLocationString`

Source/discord/discord.cpp
```diff
@@ -0,0 +1,176 @@
...
+{
+	static DiscordManager instance {};
+	return instance;
+}
+
+std::string DiscordManager::GetLocationString() const
+{
+	const char *dungeonStr = _(/* TRANSLATORS: type of dungeon (i.e. Cathedral, Caves)*/ "None");
+	if (dungeon_area_ != DTYPE_NONE) {
+		dungeonStr = _(DungeonStrs.at(dungeon_area_));
```
ephphatha, line 52: `std::out_of_range` is never caught, might as well use `operator[]` in this case. Also given `dungeon_level_` is an enum type using explicit casts here will handle potential future changes to scoped enums.

```suggestion
		dungeonStr = _(DungeonStrs[static_cast<std::size_t>(dungeon_area_)]);
```

This suggestion can be applied to the other uses of `at()` with no exception handling, and the uses of enum values with arithmetic operations.

Source/discord/discord.cpp
```diff
@@ -0,0 +1,176 @@
...
+		discord::Activity activity = {};
+		activity.SetName(gszProductName);
+		activity.SetState(GetStateString().c_str());
+		activity.SetDetails(GetDetailString().c_str());
+		activity.SetInstance(true);
+
+		activity.GetTimestamps().SetStart(start_time_);
+
+		// Set image assets
+		activity.GetAssets().SetLargeImage(GetPlayerAssetString().c_str());
```
ephphatha, line 131: `Players[MyPlayerId].AnimationData[static_cast<std::size_t>(player_graphic::Stand)]` should be initialised with the stand sprite representing the current player character class and loadout, does Discord's API allow using the sprite data directly instead of reloading it?

Source/discord/discord.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace discord {
+class Core;
+} // namespace discord
+
+namespace devilution {
+
+class DiscordManager {
+public:
+	DiscordManager();
```
ephphatha, line 19: I'm assuming this is meant to be a singleton class, if so should probably make the constructor private. That said, does it need to be defined as a class? Only `StartGame`, `UpdateGame`, and `UpdateMenu` are used by calling code, they don't need to know anything else about the implementation. Could move the implementation inside discord.cpp with non-extern globals. random.hpp/cpp is a partial example of this approach.

Source/discord/discord.cpp
heinermann: I'd rather have `at()` throw an uncaught exception that the unhandled exception filter would then report on. `operator[]` can create hard to diagnose bugs by allowing buffer overflows without incident. For example if someone is modding or the capability to add more dungeon types becomes natively supported through json or something, but this particular file is forgotten. Given that kind of risk and no drawback I don't see a benefit to using `operator[]` over `at()` in this piece.

Source/discord/discord.cpp
heinermann: It looks like `AnimationData` is for actual image data rather than the lookup string.

Discord can only use pre-uploaded sprites using string keys as lookup. I'm using the same strings that Diablo uses for this key. There are 108 character graphics (9 wpns * 3 armours * 4 characters), 1 icon, and 10 currently unused transition backgrounds, out of 300 available asset slots in the developer portal.

Source/discord/discord.h
heinermann: Yeah that's fair, it's left over from back when I was using it as a class in `diablo.cpp` over the singleton.

Source/discord/discord.cpp
ephphatha: That makes sense. Also explains why the string is lowercase when the mpq filenames are uppercase.

Source/discord/discord.cpp
AJenbo: Also I think global exception filters are only available on Windows?

heinermann: > Also I think global exception filters are only available on Windows?

There's a standard library equivalent but if everyone favours `operator[]` I'll change it.

CMake/Dependencies.cmake
heinermann: Having an issue with moving it to `3rdParty/discord`. It's failing to discover `discord_game_sdk.dll.lib` and instead looking for `discord_game_sdk.lib`, but doesn't do that when all that is in `Dependencies.cmake`. Will investigate tomorrow.

heinermann:
- Having issue with CMake and discord sdk lib location noted above.
- Still need to conditionally enable Discord rich presence for Windows/Mac/Linux PCs.

heinermann:
```cmake
message("LIB_DIR ${DISCORD_LIB_DIR}")
find_library(DISCORD_LIB discord_game_sdk ${DISCORD_LIB_DIR})
message("LIB ${DISCORD_LIB}")
```

```
1> [CMake] LIB_DIR C:/Users/heine/diablo/devilutionX/build/x64-Release/3rdParty/discord/Discord/lib/x86_64
1> [CMake] LIB C:/Users/heine/diablo/devilutionX/build/x64-Release/3rdParty/Discord/lib/x86_64/discord_game_sdk.dll.lib
```

Anyone know why CMake is bugged and returning the wrong path for something that should be dead obvious?

heinermann:
x86 builds are not functioning --

```
---------------------------
devilutionx.exe - Entry Point Not Found
---------------------------
The procedure entry point f‰^0jF0PQFPèJzÿÿƒÄ‹F<…T could not be located in the dynamic link library C:\Users\heine\Downloads\devilutionx_x86.zip\devilutionx\devilutionx.exe. 
---------------------------
OK   
---------------------------
```

heinermann:
Need to include quest levels like "Poisoned Water Supply".

heinermann:
> Can two assets be displayed at once? For example the big tileset blocks can be used as a background while the characters are layered on top.

No, but you can have a separate tiny icon in the bottom right. i.e.
2021-12-13 11_47_53-Discord Developer Portal — My Applications
 
Since there are over a hundred slots remaining another idea could be character level there.

heinermann:
Verified that the Github actions built Windows binaries are working properly now. Just need to add support for quest levels for the location field.

I also need someone to check the Mac and Linux binaries.

obligaron:
> No, but you can have a separate tiny icon in the bottom right.

We could use this for showing the game mode as icon (Diablo/Hellfire/Shareware).

AJenbo:
https://github.com/diasurgical/devilutionX/blob/master/Packaging/resources/hellfire.png

heinermann:
2021-12-14 11_15_32-general - Discord
2021-12-14 11_16_18-general - Discord

So playing Hellfire vs Diablo mode. Hovering the small icon shows the version.

glebm (approved):

PR #3718, opened 2021-12-08
Fixed issue on ios when playing with a bluetooth mouse

JoBergeron:
Mouse clicks would also send a touch event so, for example,  when clicking to leave a shop, the character would also move to where the same position where the "Leave shop" menu item was.


Source/utils/display.cpp
```diff
@@ -158,6 +158,11 @@ bool SpawnWindow(const char *lpWindowName)
 #if SDL_VERSION_ATLEAST(2, 0, 6) && defined(__vita__)
 	SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
 #endif
+    
+#if SDL_VERSION_ATLEAST(2, 0, 6) && defined(__IPHONEOS__)
+    SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
+    SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");
```
AJenbo, line 164: Use tab indentation... or clang-format :)

JoBergeron: I knew I forgot something!

Source/utils/display.cpp
```diff
@@ -160,8 +160,8 @@ bool SpawnWindow(const char *lpWindowName)
 #endif
     
 #if SDL_VERSION_ATLEAST(2, 0, 6) && defined(__IPHONEOS__)
-    SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
-    SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");
+	SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
+	SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");
```
AJenbo, line 164: @StephenCWills would it make sens to disable this in general? Since we specifically handle both I don't see a point in either.

StephenCWills: We don't use mouse touch events at all. Honestly, I wasn't even sure what context they'd be generated in so maybe the hint is off by default on most platforms.

However, touch mouse events are used in diabloui. I'd expect that you'd be unable to navigate the menus with the touchscreen on iOS after making this change. Can someone check that?

JoBergeron: I tested that. I can confirm that you can still navigate the menus with the touchscreen

StephenCWills: I tested this on Android without the check for `defined(__IPHONEOS__)`. I am completely unable to skip videos and the splash screen. I am also unable to navigate menus. I wonder how this could work on iOS but not Android.

JoBergeron: My bad, the SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0"); is not needed

AJenbo: The iPad mouse is really odd, it's works more like a touch emulator then a mouse, but it does have a movable pointer.

JoBergeron: SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0"); is needed for the mouse click not to send both a mouse click event and a touch event.

StephenCWills: Indeed, without `SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");`, everything works fine on Android. I'm not sure if Android was using touch mouse events in the first place, but my mouse still works as expected so it shouldn't hurt anything to use `SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");` on all platforms.

Source/utils/display.cpp
```diff
@@ -159,6 +159,10 @@ bool SpawnWindow(const char *lpWindowName)
 	SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
 #endif
 
+#if SDL_VERSION_ATLEAST(2, 0, 6) && defined(__IPHONEOS__)
+	SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");
+#endif
```
AJenbo, line 164:
```suggestion
SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");
```

AJenbo (approved):

StephenCWills:
@AJenbo Looks like `SDL_HINT_MOUSE_TOUCH_EVENTS` was introduced in SDL 2.0.10.
https://github.com/libsdl-org/SDL/commit/e41576188d17fd09c95777d665f6c4532574f8ac

JoBergeron:
My bad, I will fix it

StephenCWills:
> My bad, I will fix it

Not really, you did fine. AJenbo has already merged this PR and fixed the version check in master so we should be good now.

JoBergeron:
> > My bad, I will fix it
> 
> Not really, you did fine. AJenbo has already merged this PR and fixed the version check in master so we should be good now.

oh ok. Well I created a PR to check for the version anyways

Source/utils/display.cpp
StephenCWills: > Without this change, try leaving a shop using a mouse on android. You'll see you character move as well

To satisfy my own curiosity, I did this test on both Android and Windows. Android does indeed generate mouse touch events so it's good to disable this hint. Windows doesn't generate mouse touch events, but I think it's more correct to indiscriminately disable them anyway.

AJenbo:
I fucked it up twice :face_in_clouds:

PR #3720, opened 2021-12-09
Added SDL version check before SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");

JoBergeron:


StephenCWills:
Thanks for the fix, but AJenbo already cleaned up his merge so this change has already been implemented (94fa772). And especially considering I can't figure out what clang-format is complaining about, I think it's safer to just close this PR.

Also, just some potentially helpful advice. The commit log here is a bit confusing. You can fix that by rebasing or resetting your master branch onto diasurgical's, however I highly recommend creating new branches for your PRs.

JoBergeron:
ok! Still trying to get the hang of Git, thanks for the pointers

PR #3722, opened 2021-12-09
chatlog 2.0

qndel:
image
After right margin change:
image

Disadvantages vs old ver:
No spinner
No scrollbar

Improvements:
`Chat History (Messages: {:d})` turns red if there's a new message and you haven't scrolled to top yet
page up / page down scroll to top / bottom of the chatlog
Clock in upper left corner - that's why I allow opening chat log in single player, it takes care of a request to have an ingame clock :D

### DIFF
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -19,6 +19,7 @@
 #include "options.h"
 #include "pfile.h"
 #include "plrmsg.h"
+#include "qol/chatlog.h"
 #include "storm/storm_net.hpp"
 #include "sync.h"
 #include "tmsg.h"
@@ -761,6 +762,11 @@ bool NetInit(bool bSinglePlayer)
 	if (!SNetGetGameInfo(GAMEINFO_PASSWORD, szPlayerDescript, 128))
 		nthread_terminate_game("SNetGetGameInfo2");
 	PublicGame = DvlNet_IsPublicGame();
+
+	auto &myPlayer = Players[MyPlayerId];
+	// separator for marking messages from a different game
+	AddMessageToChatLog(_("New Game"), nullptr, UiFlags::ColorRed);
+	AddMessageToChatLog(fmt::format(_("Player '{:s}' (level {:d}) just joined the game"), myPlayer._pName, myPlayer._pLevel));
 
 	return true;
 }
--- a/Source/qol/chatlog.cpp
+++ b/Source/qol/chatlog.cpp
@@ -130,7 +130,7 @@ void AddMessageToChatLog(const std::string &message, Player *player, UiFlags fla
 		ChatLogLines.emplace_back(MultiColoredText { "{0} - {1}", { { timestamp, UiFlags::ColorRed }, { playerInfo, nameColor } } });
 	}
 
-	int diff = ChatLogLines.size() - oldSize;
+	unsigned int diff = ChatLogLines.size() - oldSize;
 	// only autoscroll when on top of the log
 	if (SkipLines != 0) {
 		SkipLines += diff;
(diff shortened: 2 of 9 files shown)


glebm:
> Instead of current crappy multiline design, use different colors in 1 line

This is easy, just use DrawStringWithColors

qndel:
Rebased, merge or close pls, I won't rebase again xD

glebm (approved):

obligaron (approved):
lgtm 🙂 

Only one note: You can also open the chat log in singleplayer. You will also see the chat message "new game" and "player joined" in the chat log. But this is a interesting feature, cause now you can look up how long you played. 😛

qndel:
> Clock in upper left corner - that's why I allow opening chat log in single player, it takes care of a request to have an ingame clock

that's why I wrote `Clock in upper left corner - that's why I allow opening chat log in single player, it takes care of a request to have an ingame clock` ;)

qndel:
> lgtm 🙂
> 
> Only one note: You can also open the chat log in singleplayer. You will also see the chat message "new game" and "player joined" in the chat log. But this is a interesting feature, cause now you can look up how long you played. 😛

That's why I wrote `Clock in upper left corner - that's why I allow opening chat log in single player, it takes care of a request to have an ingame clock` :P

Source/qol/chatlog.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include "engine.h"
+#include "player.h"
+
+namespace devilution {
+
+extern bool ChatLogFlag;
+
+void ToggleChatLog();
+void AddMessageToChatLog(const std::string &message, Player *player = nullptr, UiFlags flags = UiFlags::ColorWhite);
```
AJenbo, line 16: Maybe use a string_view instead of string so that you don't have to create a string from const char* when calling this function.

Source/multi.cpp
```diff
@@ -762,6 +763,11 @@ bool NetInit(bool bSinglePlayer)
 		nthread_terminate_game("SNetGetGameInfo2");
 	PublicGame = DvlNet_IsPublicGame();
 
+	auto &myPlayer = Players[MyPlayerId];
+	// separator for marking messages from a different game
+	AddMessageToChatLog(_("New Game"), nullptr, UiFlags::ColorRed);
+	AddMessageToChatLog(fmt::format(_("Player '{:s}' (level {:d}) just joined the game"), myPlayer._pName, myPlayer._pLevel));
```
AJenbo, line 769: It feels like it would make more sens to clear the log between each game, that way it also won't grow to much over time.

Source/qol/chatlog.cpp
```diff
@@ -0,0 +1,209 @@
...
+	const int sx = PANEL_X + PaddingLeft;
+	const int sy = UI_OFFSET_Y;
+
+	DrawString(out, fmt::format(_("Chat History (Messages: {:d})"), MessageCounter),
+	    { { sx, sy + PaddingTop + blankLineHeight }, { ContentTextWidth, lineHeight } },
+	    (UnreadFlag ? UiFlags::ColorRed : UiFlags::ColorWhitegold) | UiFlags::AlignCenter);
+
+	char timestamp[64];
+	time_t tm = time(nullptr);
+	sprintf(timestamp, "%02d:%02d:%02d", localtime(&tm)->tm_hour, localtime(&tm)->tm_min, localtime(&tm)->tm_sec);
```
AJenbo, line 161: Could we use a std::string and fmt for this?

Source/qol/chatlog.h
qndel: Then I have to do 
image
in AddMessageToChatLog anyway

Source/qol/chatlog.cpp
qndel: aye

PR #3726, opened 2021-12-10
Use the latest (stable) debian image for linux_x86_64 build

ephphatha:
Bumps the image to debian:buster with g++ 10 instead of debian:stretch which uses g++ 6. g++ versions < 7 have bugs around the handling of constexpr class members (which causes failures such as #3719).

All tools appear to work the same (minor versions are more recent but no breaking changes that I could see). AppImage assumes `file` is present on the system so this needs to be installed in the buster image.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,14 +2,12 @@ version: 2
 jobs:
   linux_x86_64:
     docker:
-      - image: debian:stretch-backports
+      - image: debian:latest
     working_directory: ~/repo
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


AJenbo:
this would effectively bump the minimal supported system level. Considering that my PowePC system is currently stuck on GCC 6.3 this would mean I would have no way of testing BE support

ephphatha:
Ahh, ok, I wasn't sure why such an old version was being targeted (especially since the amigaos docker image is using debian:testing as a base...). Looks like there's no upgrade for that image which keeps m68 compatibility so I'll have to add workarounds for missing functionality anyway, meaning there's not much reason to upgrade this image now.

PR #3730, opened 2021-12-11
Clean up check for borken monster position in DirOK

AJenbo:
This check was needed back when monsters current position could appear as a negative index. Since it has been fixed this check should no longer be needed.


ephphatha (approved):
I think this is ok, the old `minion.position.future == Point {x, y}` check looks like it was only counting leashed minions which were standing still anyway. There should be no issues caused by north moving minions being out of range of the leader based on the tile they're leaving.

PR #3733, opened 2021-12-11
Check that monsters aren't placed in doors

### LINKED ISSUE #2700: Monster on same tile as a door
**Important information**
Linux on commit d8a0211da864a9195311f01af19797652e3a5814

**Describe**
I'm not quite sure if the monster spawned exactly there, but I've never been at the door to my left. But there seems to be a monster occupying the same tile as the (closed) door. You can hit it with a firebolt and if you come too close it can hit you through the closed door without opening it.

**To Reproduce**
Open the savegame, walk to the left. Please don't kill my hero!

**Expected behavior**
Either it's behind the closed door or in front of the closed door. Both things at the same time sound to me like Schrödinger's cat.

**Screenshots**
image
image
[single_9.zip](https://github.com/diasurgical/devilutionX/files/7049359/single_9.zip)

obligaron:
Fixes #2700

Notes:
- I verified that the monsters exactly spawn in doors (disabled monster AI for test).
- Doors are not present in `nSolidTable`, cause you can open them => not solid 😉 
- Monsters are placed (`InitMonsters`) before objects get generated (`InitObjects`).
  That means we can't check `dObject` if a door exists, cause it's not yet initialized.
  As alternative we use `dPiece` cause it also contains the information that a door exists (and is used for object creation).
- Monsters are placed on different locations, if they would spawn in a door. That can result in a different "monster layout" then in vanilla. Visited dungeons in old save games are not affected (only initial level generation).
- An alternative fix to this pr could be to generate opened doors (instead of closed doors), if a monster is placed in them. This would avoid the different "monster layout".

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -266,6 +266,29 @@ bool CanPlaceMonster(int xp, int yp)
 		return false;
 	}
 
+	if (!setlevel) {
+		// Avoid to spawn a monster in a door
+		// Monsters are placed (InitMonsters) before objects get generated (InitObjects).
+		// That means we can't check dObject if a door exists, cause it's not yet initialized.
+		// As alternative we use dPiece cause it also contains the information that a door exists (and is used for object creation).
+		int pn = dPiece[xp][yp];
+		switch (leveltype) {
+		case DTYPE_CATHEDRAL:
+		case DTYPE_CATACOMBS:
+			if (IsAnyOf(pn, 13, 541, 17, 542))
+				return false;
+			break;
+		case DTYPE_CAVES:
+			if (IsAnyOf(pn, 531, 534))
+				return false;
+			break;
+		case DTYPE_CRYPT:
+			if (IsAnyOf(pn, 77, 80))
+				return false;
+			break;
+		}
+	}
+
 	return !IsTileSolid({ xp, yp });
 }
 


AJenbo:
We already generate different monster and object layout which is acceptable since they are loaded from save games.

How about we simply place objects before monster and then check for solid objects when placing monsters? That would also avoid them spawning on top of other solid objects like book stands or barrels.

obligaron:
Can monsters spawn on solid objects like book stands?
I thought the check in `IsTileSolid` would prevent this, cause `IsTileSolid` checks `dPiece` and that's initialized when placing monsters. 🤔 

But we could initialize objects before monsters. That's a good alternative. Then we could check if a door is on the tile and stop monster spawning. This is clearly more intuitive then checking `dPIce` ids.
I hesitated to do this, cause it's a much bigger change. But if you want I can change the pr. 🙂

AJenbo:
Is it a big change to the code, or just a big change to how things work? I think it's the proper solution, since if more objects are added they would need to take them in to consideration also.

> Can monsters spawn on solid objects like book stands?
> I thought the check in `IsTileSolid` would prevent this, cause `IsTileSolid` checks `dPiece` and that's initialized when placing monsters. 🤔

If objects haven't been placed yet how can IsTileSolid check for them?

obligaron:
> Is it a big change to the code, or just a big change to how things work? I think it's the proper solution, since if more objects are added they would need to take them in to consideration also.

How things work. It's only swapping [`InitObjects`](https://github.com/diasurgical/devilutionX/blob/aee8ed3951af2cb635a7e2ce3b6ebe0ce782ee67/Source/diablo.cpp#L1921) with [`InitMonsters`](https://github.com/diasurgical/devilutionX/blob/aee8ed3951af2cb635a7e2ce3b6ebe0ce782ee67/Source/diablo.cpp#L1918).

> If objects haven't been placed yet how can IsTileSolid check for them?

When dungeons gets generated `dPiece` (and `nSolidTable`) is filled.
The dungeon generator also generate the objects and add the door-specific piece id's to `dPiece`.
This happens before `InitMonsters` and `InitObjects`.

`IsTileSolid` checks `dPiece` if there is a solid object (with the help of `nSolidTable`).
My thought was that a book stand would be solid and be contained in `nSolidTable`.

When `InitObjects` runs, ti's checks `dPiece` what objects should be generated (added to `dObject` and `Objects`).

The problem is that door's aren't considered solid, cause you can walk through them. When you open them. 😉 

But I think you are right that the same problem could occur with barrels. You can also walk through them after they are destroyed. 🤔

obligaron:
Okay checked again.
Barrels don't get generated in dungeon generation. They get generated in `InitObjects` and checks that no monster is spawned on barrel location.
So there should never be a barrel and a monster on the same location (same is true for other objects, see uses of `RndLocOk`).

AJenbo:
> IsTileSolid checks dPiece if there is a solid object (with the help of nSolidTable).

`nSolidTable` does not hold information regarding solid object, only dungeon tiles. So `IsTileSolid` says nothing about if a solid object will be placed there at a later point.

The reason monsters mostly do not appear inside objects is that when placing objects `RndLocOk` is usually (true for barrels) called which checks if a monster was placed in that position previously.

When placing monsters I think `IsTileOccupied` is the correct function to make the check.

Reading the code I think that a monster could potentially get stuck in a book stand as they are placed after monsters with no call to `RndLocOk`, during `CreateThemeRooms`.

PR #3735, opened 2021-12-11
Call SyncPortals() only once for a level load

### LINKED ISSUE #3734: Incorrect Town portal position shifts
Of course, I isolated the issue so that it could be fixed more quickly.
This has occurred since the commit 9964bd9e495314e2d466eeb6a0c11d3c12c963a.
With this commit, an important bug was fixed, but a new one was also introduced.

1. The bug was discovered accidentally
2. It is unclear whether further issues could have arisen in connection with the commit
3. To reproduce, download the attached game and go up to town or down the next level and return
4. You should notice that the portal changes its position a little every time you return
5. As soon as the portal reaches a wall, it does one step backwards
6. savegame attached
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/7696960/single_0.hsv.zip)

https://user-images.githubusercontent.com/45902449/145676267-18a99b41-a215-4fbe-80f9-7e212b21a8ce.mp4

obligaron:
Fixes #3734

Notes:
- `SyncPortals` was called twice (once in `LoadLevel` and once in `LoadGameLevel`)
- That means `AddTown` was also called twice. The first call creates a town portal. The second call sees this town portal (`TileContainsMissile` is true) and places the second town portal near the first town portal. But a player is only allowed to have one town portal. That's why the first town portal gets deleted (in the video in the pr you can see it for a short period).
- Fix is to call `SyncPortals` only once.
- 9964bd9e495314e2d466eeb6a0c11d3c12c963ae only worked, cause the check for existing missiles/town portals was broken before this commit/bugfix.
- Alternative fix could be to check if new to place town portal is on the same location as a existing town portal from the same player and allow this special case.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -2271,7 +2271,6 @@ void LoadLevel()
 		AutomapZoomReset();
 		ResyncQuests();
 		RedoMissileFlags();
-		SyncPortals();
 		UpdateLighting = true;
 	}
 


ephphatha (approved):
Works fine in single/multi games now :D

Chance4us:
Thanks ;-)

AJenbo:
Does save games also work :)

ephphatha:
> Does save games also work :)

Portals only seem to shift on level transition in master, loading from a save didn't cause the bug.

PR #3736, opened 2021-12-11
MpqWriter: Cleanups and renames

glebm:
I've spent some time understanding how the `MpqWriter` works and cleaned it up a bit and renamed some variables for clarity.

1. Fixes all signedness and clang-tidy warnings.
2. Renames variables, structs, and methods for clarity.
3. Marks structs used for I/O as packed.
4. Adds comments in a few places.
5. Eliminates recursion from `MpqWriter::AllocBlock` (a clang-tidy warning).

### DIFF
--- a/Source/mpq/mpq_writer.cpp
+++ b/Source/mpq/mpq_writer.cpp
@@ -15,46 +15,70 @@
 
 namespace devilution {
 
-#define INDEX_ENTRIES 2048
-
 namespace {
 
 // Validates that a Type is of a particular size and that its alignment is <= the size of the type.
 // Done with templates so that error messages include actual size.
-template <std::size_t A, std::size_t B>
+template <size_t A, size_t B>
 struct AssertEq : std::true_type {
 	static_assert(A == B, "A == B not satisfied");
 };
-template <std::size_t A, std::size_t B>
+template <size_t A, size_t B>
 struct AssertLte : std::true_type {
 	static_assert(A <= B, "A <= B not satisfied");
 };
-template <typename T, std::size_t S>
+template <typename T, size_t S>
 struct CheckSize : AssertEq<sizeof(T), S>, AssertLte<alignof(T), sizeof(T)> {
 };
 
 // Check sizes and alignments of the structs that we decrypt and encrypt.
 // The decryption algorithm treats them as a stream of 32-bit uints, so the
 // sizes must be exact as there cannot be any padding.
-static_assert(CheckSize<_HASHENTRY, 4 * 4>::value, "sizeof(_HASHENTRY) == 4 * 4 && alignof(_HASHENTRY) <= 4 * 4 not satisfied");
-static_assert(CheckSize<_BLOCKENTRY, 4 * 4>::value, "sizeof(_BLOCKENTRY) == 4 * 4 && alignof(_BLOCKENTRY) <= 4 * 4 not satisfied");
+static_assert(CheckSize<MpqHashEntry, static_cast<size_t>(4 * 4)>::value, "sizeof(MpqHashEntry) == 4 * 4 && alignof(MpqHashEntry) <= 4 * 4 not satisfied");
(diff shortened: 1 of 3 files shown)


Source/mpq/mpq_writer.cpp
```diff
@@ -193,18 +215,12 @@ bool MpqWriter::ReadMPQHeader(_FILEHEADER *hdr)
...
-	for (int i = 0; i < INDEX_ENTRIES; i++, blockEntry++) {
-		if (blockEntry->offset != 0)
-			continue;
-		if (blockEntry->sizealloc != 0)
-			continue;
-		if (blockEntry->flags != 0)
-			continue;
-		if (blockEntry->sizefile != 0)
+	for (unsigned i = 0; i < BlockEntriesCount; ++i, ++blockEntry) {
+		if (blockEntry->offset != 0 || blockEntry->flags != 0 || blockEntry->unpackedSize != 0 || blockEntry->packedSize != 0)
```
AJenbo, line 223: This is becoming a bit wide, I think I preferred the multi line version

Source/mpq/mpq_writer.cpp
```diff
@@ -218,111 +234,102 @@ _BLOCKENTRY *MpqWriter::NewBlock(int *blockIndex)
 
 void MpqWriter::AllocBlock(uint32_t blockOffset, uint32_t blockSize)
 {
-	_BLOCKENTRY *block;
-	int i;
-
+	MpqBlockEntry *block;
+expanded:
```
AJenbo, line 238: What does this do?

StephenCWills: It's the label for the gotos. I guess the function always returns after the recursion so there's no need to allocate the extra stack space if we simulate the tail recursion by instead jumping back to the beginning of the function.

Wouldn't this still be pretty easy to implement by surrounding the for loop in a while loop? Perhaps we should avoid the gotos here.

glebm: I'm not sure how to implement it with a while loop without an extra boolean. It seemed pretty confusing when I tried it that way. `goto` is sometimes (but rarely so) the simplest solution.

Source/mpq/mpq_writer.cpp
glebm: It was a bit odd to have these checks separate. I've now extracted this into a separate function!

Source/mpq/mpq_writer.cpp
StephenCWills: You would need another bool for the while loop, but it would essentially take the place of the label. Conceptually, it's already pretty difficult to understand what's going on in this loop, and I don't think the while loop really adds any confusion compared to the goto. I'd argue that the curly braces and nesting should help to see the structure of the nested loop here, even if it doesn't really help to clarify the intent.

```c++
	MpqBlockEntry *block;
	bool expanded = true;
	while (expanded) {
		expanded = false;
		block = blockTable_;
		for (unsigned i = BlockEntriesCount; i-- != 0; ++block) {
			// Expand to adjacent blocks.
			if (!IsAllocatedUnusedBlock(block))
				continue;
			if (block->offset + block->packedSize == blockOffset) {
				blockOffset = block->offset;
				blockSize += block->packedSize;
				memset(block, 0, sizeof(MpqBlockEntry));
				expanded = true;
				break;
			}
			if (blockOffset + blockSize == block->offset) {
				blockSize += block->packedSize;
				memset(block, 0, sizeof(MpqBlockEntry));
				expanded = true;
				break;
			}
		}
	}
```

AJenbo: > I'd argue that the curly braces and nesting should help to see the structure of the nested loop

I much agree with this. Goto is easy to miss and break.

Other then that I'm happy with this PR at this point 👍

glebm: I find it a lot more readable with `goto` but I guess I'm more used to `goto` from reading lots of lower-level performance-optimized C code. Done.

AJenbo (approved):

PR #3737, opened 2021-12-11
storm_svid: Cleanup video rendering code

glebm:
1. Removes adjustments to `orig/system_palette`. They're not needed because videos have a separate render loop.  For an in-game movie, the caller sets the palette after the movie.

2. Splits some of `SVidPlayContinue` code into functions (`UpdatePalette` and `BlitFrame`).
3. Adds a separate config variable for SDL1 SVid video mode. E.g. on rg99 there isn't enough video RAM for double-buffering
   the main game but we can do it for the videos.

Fixes videos on rg99.


StephenCWills (approved):
I like the code changes. Also, I tested on 3DS and Windows and didn't encounter any issues.

glebm:
@StephenCWills Thanks for testing!

AJenbo (approved):

PR #3738, opened 2021-12-11
Fix page down/up don't wrap in settingsmenu

### LINKED ISSUE #3723: Page down/up in DiabloUI should not wrap
Unlike the arrow keys, it is really odd for page up/down to wrap past the start/end.

obligaron:
Fixes #3723

@glebm please take a look 🙂


glebm:
Thanks! Looks good to me

glebm (approved):

PR #3740, opened 2021-12-11
Fix Knockback checks wrong position when a monster is moving

### LINKED ISSUE #3527: Two enemies occupy the same tile
**Important information**
Ubuntu MATE 20.04 / DevilutionX 1.30 / AppImage / Hellfire game

**Describe**
I managed to get two enemies to occupy the same tile directly behind an opened door.
On Level 6 I opened a door with dozens of "Unseen" enemies in it. I stood in front of the door and already killed one or two oncoming enemies holding the shift key, when another enemy stepped into the door and was immediately knocked back by my weapon (or maybe he turned around and tried to walk away, I'm not absolutely sure what happened).
This seemed to have put this enemy on the same tile as another one that followed right behind him. Since those two "merged" enemies cannot move, and they are also blocking the tile behind the door, no other enemies were able to approach me. They all just stand there waiting.
The merged enemies are flickering with each of them being seen in the foreground for a few milliseconds. One of the them is facing me while the other enemy is facing the other way.

**To Reproduce**
I don't even know if this can be reproduced. I managed to trigger it once and only by accident.

**Expected behavior**
Enemies shouldn't be able to get blocked by accidentally occupying the same tile. When the surrounding tiles are all blocked, then knocking back should probably fail.

**Screenshots**
I have attached a screenshot (sorry for the size, this is because my desktop has 4K resolution).
Unfortunately there is not much to be seen on the screenshot, besides the blocked Unseen standing there. They are not able to move.

**Additional context**
Add any other context about the problem here
devilutionx
.

obligaron:
Fixes #3527 (hopefully)

I couldn't reproduce the issue locally. So this pr is only a educated guess.

In `M_GetKnockback`:
`DirOk` previous checked if the dir + `monster.position.tile` is free.
When the check is successful the monster will be set on dir + `monster.position.old` (see end of `M_GetKnockback` and `StartMonsterGotHit`).
This can only differ if the monster is currently moving. In #3527 the monster was perhaps retreating.

This pr fixes changes the check to dir + `monster.position.old` is free.
Also we don't forbid knock back if the monster is part of a pack.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -3540,6 +3540,26 @@ void (*AiProc[])(int i) = {
 	/*AI_BONEDEMON*/ &BoneDemonAi
 };
 
+bool IsRelativeMoveOK(const Monster &monster, Point position, Direction mdir)
+{
+	Point futurePosition = position + mdir;
+	if (!InDungeonBounds(futurePosition) || !IsTileAvailable(monster, futurePosition))
+		return false;
+	if (mdir == Direction::East) {
+		if (IsTileSolid(position + Direction::SouthEast))
+			return false;
+	} else if (mdir == Direction::West) {
+		if (IsTileSolid(position + Direction::SouthWest))
+			return false;
+	} else if (mdir == Direction::North) {
+		if (IsTileSolid(position + Direction::NorthEast) || IsTileSolid(position + Direction::NorthWest))
+			return false;
+	} else if (mdir == Direction::South)
+		if (IsTileSolid(position + Direction::SouthWest) || IsTileSolid(position + Direction::SouthEast))
+			return false;
+	return true;
+}
+
 } // namespace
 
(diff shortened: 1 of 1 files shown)


Source/monster.cpp
```diff
@@ -3540,6 +3540,26 @@ void (*AiProc[])(int i) = {
 	/*AI_BONEDEMON*/ &BoneDemonAi
 };
 
+bool DirOK(const Monster &monster, Point position, Direction mdir)
```
AJenbo, line 3543: Reuse of the `DirOK` name is a bit confusing to me :)

```suggestion
bool IsRelativeMoveOK(const Monster &monster, Point position, Direction mdir)
```

obligaron: Renamed 🙂

AJenbo (approved):

PR #3741, opened 2021-12-11
Update Swedish Translation.

SimonCeder:
Back to 100%.


AJenbo:
Yay nice work :D

I got some positive feedback regarding the translation for halls of the blind already :)

PR #3743, opened 2021-12-11
`make_src_dist`: Add a top-level directory

glebm:
Addresses a request by @jpalus to place the archive contents into a top-level directory. (https://github.com/diasurgical/devilutionX/pull/3589#issuecomment-981084124)

The directory looks like this: `devilutionx-src-1.3.0-7154522bf` (or  `devilutionx-src-1.3.0` for a tagged commit).

@jpalus Is this what you meant?


jpalus:
@glebm Thanks for taking care of it! If I understand correctly for official releases directory will be called devilutionx-src-1.3.0 is that correct? If that's the case then it's perfectly fine by me. Additional git hash (supposedly for nightly releases?) does not look that great but only because I'm lazy :) (I prefer to just update version and skip searching git hash associated with this release which btw some projects do).

glebm:
@jpalus Yeah that's exactly how it'll be. Note that in the release the file itself will be named `devilutionx-src.tar.xz` (so that /latest/ links can work), but the archive will have the version number

jpalus:
@glebm Thanks for confirmation. That's fine by me for archive name, though I would have preferred it to include version number (I'd argue that most of source distribution consumers are not much interested in /latest/ links since all of them version their target packages, but perhaps it's just my bias :) ).

AJenbo (approved):

PR #3745, opened 2021-12-12
(Partially) deduplicate code used to check if a player can equip an item

ephphatha:
`CheckItemStats`, `ItemMinStats`, and `StoreStatsOk` all did effectively the same thing. The `item.statFlag` variable itself can be replaced by a check against the local player but with the way this check is used it's sometimes hard to define what is the local player (especially when building the save game list). I've done a very limited dedup/cleanup here to avoid breaking things by trying to remove runtime use of `statFlag` completely.


AJenbo (approved):
Only thing I don't like about this is that it will cause a conflict with what i'm working

PR #3746, opened 2021-12-12
Overhaul tests

glebm:
1. Adds a `libdevilution_so` target when tests are enabled.
2. Each test file is now a separate binary target linked against `libdevilutionx_so` (can now run tests in parallel).
3. Tests are now defined in a separate `test/CMakeLists.txt` file.
4. Building the tests is now controlled by the standard `BUILD_TESTING` option (defined by CTest).
5. Tests are now built by default.
6. On CI, test errors are now reported.

Visual Studio test explorer now looks like this:

image


glebm:
Ah, if I just run the binary I see an ASAN error:

```
../test/pack_test.cpp:720:2: runtime error: reference binding to misaligned address 0x7ffd9deb3e1d for type 'const short unsigned int', which requires 2 byte alignment
0x7ffd9deb3e1d: note: pointer points here
 00 00 00 00 e8 03 00  00 00 00 9d fd 7f 00 00  d0 06 00 00 20 60 00 00  80 3e eb 9d fd 7f 00 00  00
             ^ 
/usr/include/gtest/gtest.h:1545:11: runtime error: load of misaligned address 0x7ffd9deb3e1d for type 'const short unsigned int', which requires 2 byte alignment
0x7ffd9deb3e1d: note: pointer points here
 00 00 00 00 e8 03 00  00 00 00 9d fd 7f 00 00  d0 06 00 00 20 60 00 00  80 3e eb 9d fd 7f 00 00  00
```

glebm:
Ah, I think this is due to the missing font kerning

Source/utils/attributes.h
```diff
@@ -31,3 +31,14 @@
 #else
 #define DVL_ATTRIBUTE_HOT
 #endif
+
+// Any global data used by tests must be marked with `DVL_API_FOR_TEST`.
```
AJenbo, line 35: are we running in to issues that makes this necessary?

glebm: Now that we link libdevilutionx to tests as a shared library, all global data must be marked this way for MSVC to export it.
For functions and classes we set the `WINDOWS_EXPORT_ALL_SYMBOLS` property but it doesn't work for global variables.

The alternative would be to link libdevilutionx statically into each test binary but that is much slower and would take lots of disk space.

AJenbo: > Now that we link libdevilutionx to each test binary as a shared library

Well this is the part I don't understand why is needed :)

glebm: > The alternative would be to link libdevilutionx statically into each test binary but that is much slower and would take lots of disk space.

^ basically this

ephphatha (commented):
Purely from the perspective of someone annoyed with visual studio test explorer not picking up new test suites I really like the split test binaries approach. Having the test explorer display tests with less nesting is a nice side effect. Builds don't appear to be any faster (or slower) so the overall impact appears minimal. The awkwardness needed when referencing globals in tests provides a bit of incentive to refactor code and avoid globals :D

Auto-merge is on so I'm not gonna submit this as approval 😂 but it looks good to me.

AJenbo:
> Purely from the perspective of someone annoyed with visual studio test explorer not picking up new test suites I really like the split test binaries approach. Having the test explorer display tests with less nesting is a nice side effect.

That's good to hear.

> The awkwardness needed when referencing globals in tests provides a bit of incentive to refactor code and avoid globals :D

I was just thinking the same thing, I guess that is a good enough justification :joy: 

> This also paves the way forward for splitting DevilutionX itself into separate libraries that can be tested independently.

Still a bit fuzzy to me, but I also can't help getting excited by the sound of it :P

Source/utils/paths.cpp
```diff
@@ -100,7 +100,7 @@ const std::string &ConfigPath()
 const std::string &AssetsPath()
 {
 	if (!assetsPath)
-		assetsPath.emplace(AppPath() + "assets/");
+		assetsPath.emplace(BasePath() + "assets/");
```
AJenbo, line 103: The intent here is to load it from the build folder. Changing this to use BasePath will cause the game to fail to launch if a recent version of the devilutionx.mpq ins't in the search paths.

glebm:
Rebased

Source/utils/paths.cpp
glebm: Actually `BasePath` was the same as `AppPath` unless set via CLI flags. Removed `AppPath` in https://github.com/diasurgical/devilutionX/pull/3770

AJenbo: Yes, that is the point. It should be the path of the exe even when the CLI flag is given.

glebm: But why? We do not know that `assets/` is next to the exe
`--data-dir` is mostly used by distros that place the data files in the custom likes

glebm: > if a recent version of the devilutionx.mpq ins't in the search paths.

devilutionx.mpq is not looked up via this. It is only used for `assets/`

AJenbo: We do, it's put there by the build process. `assets/` is only used when running the application from an IDE and not having smpq installed. It is not intended for use by the end user.

AJenbo: I think you are confusing direct file access (which uses just a relative path) and `assets/` which each other. `assets/` is purely for developers. I guess this also explains why you where confused regarding the path where files should be placed for the iOS container.

AJenbo (approved):

PR #3747, opened 2021-12-12
3rdParty/libpng: SKIP_INSTALL_ALL

glebm:
We do not need to create any install targets when using libpng from source.

/cc @bubio @StephenCWills This may fix the iOS `EFFECTIVE_PLATFORM_NAME` error


StephenCWills (approved):
Seems like a good idea to me.

PR #3749, opened 2021-12-12
Initialize golems before objects

### LINKED ISSUE #3748: Golems not initialized in player indexes
**Important information**
Latest master, debug build (for asserts)

**Describe**
The game fails to initialize golems in the expected player indexes in the monster array.

**To Reproduce**
1. Create a debug build
2. Start a new game
3. Enter the cathedral

**Expected behavior**
Golems should be initialized to indexes 0-3 during `InitMonsters()`.

**Screenshots**
image

**Additional context**
The issue seems to be caused by the fact that monsters can be spawned by `InitObjects()`, for instance when skeletons appear from barrels and sarcophagi.

Introduced by #3744

StephenCWills:
This resolves #3748

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1915,6 +1915,7 @@ void LoadGameLevel(bool firstflag, lvl_entry lvldir)
 			if (firstflag || lvldir == ENTRY_LOAD || !myPlayer._pLvlVisited[currlevel] || gbIsMultiplayer) {
 				HoldThemeRooms();
 				[[maybe_unused]] uint32_t mid1Seed = GetLCGEngineState();
+				InitGolems();
 				InitObjects();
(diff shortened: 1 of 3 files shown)


obligaron:
`InitMonsters` has 3 call sites.
I think we need to adjust all 3. :-/

AJenbo (dismissed):

StephenCWills:
> `InitMonsters` has 3 call sites. I think we need to adjust all 3. :-/

Good catch, thanks. Although, it makes me wonder whether we really need to initialize golems in town. 🤔

obligaron:
I also wasn't sure about town. But better be save then sorry? 😛 
Perhaps in the future we don't need that golems are in the first monster slots.
For example when we want to support more then 4 players in multiplayer.

AJenbo (approved):

PR #3750, opened 2021-12-12
[Android] Fix error when turning on upscaling in the settings menu

### LINKED ISSUE #3708: [android] Multiple renderers when reenabling upscaling
Switch upscalling off, then on:

image

StephenCWills:
This is a hack that manipulates the window data that SDL uses internally to track the renderer associated with the window. It's the only way I could find to make toggling the upscale option work on Android.

**Other things I tried:**
* Simply destroy the internal renderer before creating our renderer. Unfortunately, the SDL internals also keep track of the "window texture" so there's no way to force it to recreate this renderer without a similar (but probably worse) hack.
* Recreate the whole window. This freezes up the display, and I couldn't figure out how to get the display to start updating again after recreating the window.

This resolves #3708

### DIFF
--- a/Source/utils/display.cpp
+++ b/Source/utils/display.cpp
@@ -254,6 +254,11 @@ void ReinitializeRenderer()
 	Size windowSize = { current.current_w, current.current_h };
 	AdjustToScreenGeometry(windowSize);
 #else
+	// SDL headers don't expose this key so this is taken from SDL source code
+	// https://github.com/libsdl-org/SDL/blob/release-2.0.18/src/render/SDL_render.c#L47
+	constexpr char SDL_WINDOWRENDERDATA[] = "_SDL_WindowRenderData";
+	static SDL_Renderer *sdlInternalRenderer = nullptr;
+
 	if (texture)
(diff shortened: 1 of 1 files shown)


AJenbo:
I guess the alternative is to not allow switching off upscaling on Android

StephenCWills:
> I guess the alternative is to not allow switching off upscaling on Android

I considered that too. Android may not be the only platform with this issue, but even so.. I guess we can just lump them all together as platforms that can't toggle upscale in-game.

glebm:
Perhaps this should be fixed in SDL?

StephenCWills:
> Perhaps this should be fixed in SDL?

I'm going back and forth on this, however I think it would be best to simply disable/hide the setting on Android. Essentially, I question the value of providing the option to disable upscaling on a system that is just going to use the render API under the covers anyway. It should be possible to simply fiddle with the other settings to get an equivalent result. FYI, the actual effect of `Upscale=0` on my phone is the same as changing the game's resolution to match that of my phone's screen.

PR #3751, opened 2021-12-12
Use std::string in PrintItemPower

AJenbo:
- Mark item as `const`, call it `item` instead of `x` and pass it by reference
- Prep work for store.cpp rewrite
- xgettext is no longer confused about the string format
- Code is simpler as string copying is implicit and break replaced by direct returns
- Removes use of `tempstr` global
- Avoid improper use of `PrintItemPower` via `[[nodiscard]]`


Source/items.cpp
```diff
@@ -3791,292 +3791,226 @@ void DoOil(Player &player, int cii)
...
-		if ((x->_iFlags & ISPL_STEALMANA_3) != 0)
-			/*xgettext:no-c-format*/ strcpy(tempstr, _("hit steals 3% mana"));
-		if ((x->_iFlags & ISPL_STEALMANA_5) != 0)
-			/*xgettext:no-c-format*/ strcpy(tempstr, _("hit steals 5% mana"));
-		break;
+		if ((item._iFlags & ISPL_STEALMANA_3) != 0)
+			return _("hit steals 3% mana");
+		if ((item._iFlags & ISPL_STEALMANA_5) != 0)
+			return _("hit steals 5% mana");
+		return _("Another ability (NW)");
```
ephphatha, line 3929: Previously tempstr was left with the result of the last successful call. Even though in practice this case should never be hit could we use either `""` or a more descriptive string for these cases?
```suggestion
		return _("hit steals mana");
```

If you wanted to put a number on it "hit steals 0% mana" is accurate 😂

Source/items.cpp
```diff
@@ -3791,292 +3791,226 @@ void DoOil(Player &player, int cii)
...
-		if ((x->_iFlags & ISPL_STEALLIFE_3) != 0)
-			/*xgettext:no-c-format*/ strcpy(tempstr, _("hit steals 3% life"));
-		if ((x->_iFlags & ISPL_STEALLIFE_5) != 0)
-			/*xgettext:no-c-format*/ strcpy(tempstr, _("hit steals 5% life"));
-		break;
+		if ((item._iFlags & ISPL_STEALLIFE_3) != 0)
+			return _("hit steals 3% life");
+		if ((item._iFlags & ISPL_STEALLIFE_5) != 0)
+			return _("hit steals 5% life");
+		return _("Another ability (NW)");
```
ephphatha, line 3935:
```suggestion
		return _("hit steals life");
```

Source/items.cpp
AJenbo: I think I'll go with `""`. No need for adding new strings to be translated if they are never used :)
```suggestion
		return "";
```

Source/items.cpp
AJenbo:
```suggestion
		return "";
```

PR #3752, opened 2021-12-12
Perform item requirement check when entering vendors

AJenbo:
More prep work for rewriting stores.cpp

Vendor item stat checks were being performed at some really odd times:
- When reading a hero from a save game (select hero menu etc.)
- When generating the vendor items
- Reparing, identifying, recharging, applying oil, buying, resurrect, changing equipment
- Joining/starting a game, entering town
- Applying stats

Instead of all that craziness simply calculate it when opening the vendor item list.

### DIFF
--- a/Source/stores.cpp
+++ b/Source/stores.cpp
@@ -365,7 +365,11 @@ void StartSmithBuy()
 	AddItemListBackButton();
 
 	storenumh = 0;
-	for (int i = 0; !smithitem[i].isEmpty(); i++) {
+	for (Item &item : smithitem) {
+		if (item.isEmpty())
+			continue;
+
+		item._iStatFlag = MyPlayer->CanUseItem(item);
 		storenumh++;
 	}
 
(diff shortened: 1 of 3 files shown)


glebm (dismissed):

Source/stores.cpp
```diff
@@ -365,7 +365,11 @@ void StartSmithBuy()
 	AddItemListBackButton();
 
 	storenumh = 0;
-	for (int i = 0; !smithitem[i].isEmpty(); i++) {
+	for (Item &item : smithitem) {
+		if (item.isEmpty())
+			continue;
```
ephphatha, line 370: With all of these loops previously they would `break;` on the first empty item (which doesn't matter in practice since the arrays are sorted so there are no holes). Did you want to restore that behaviour or always run through the whole list?

AJenbo: I think it's simpler to process the whole list.

The idea is to treat all arrays the same (`premiumitems` isn't sorted) and not have special requirements, like needing an empty item at the end to avoid OOB. Things should work out much the same as before since it will skip all empty items.

ephphatha (approved):

PR #2934, opened 2021-09-24
Update Croatian translation

muzena:

PR #3758, opened 2021-12-14
Prepend version suffix with dash character in source tarball

StephenCWills:
Based on feedback from https://github.com/diasurgical/devilutionX/pull/2934#issuecomment-993789379. FYI, this merely looks correct to me, but I haven't tested this at all.


StephenCWills:
I just had another thought. Seems to me that it's probably more intuitive to define `VERSION_SUFFIX` without the dash. Unless there's some expectation that one should be allowed to use a different character to separate the version number from suffix, we could make changes to `CMakeLists.txt` instead.

AJenbo:
That sounds sensible, also it should be omitted when building a release

muzena:
I use this for [PPA git repo](https://launchpad.net/~devilutionx/+archive/ubuntu/dev), I made this [script](https://github.com/muzena/devilutionX/blob/debian-manually/Packaging/debian/devilutionx-git-upload) and add it to cron job to run once day or I start it manually since Launchpad PPA build do not accept cloning or downloading dependencies outside Launchpad

PR #3759, opened 2021-12-14
Improve CMake logic for version suffix

### LINKED ISSUE #3382: Android release builds contain git hash in version
It appears that Android CMake includes the commit has despite build a release build, for 1.3.0 i had to edit the CMake files to "fix" the version number.

StephenCWills:
This introduces a `PROJECT_VERSION_WITH_SUFFIX` variable which includes the version suffix only for Debug builds.

* Dash character is no longer prepended to the `VERSION_SUFFIX` variable, but is instead added to the `PROJECT_VERSION_WITH_SUFFIX` variable as needed.
* If `VERSION_NUM` is predefined, but `VERSION_SUFFIX` is not, force a generic value of `debug` as the `VERSION_SUFFIX`. This allows us to append the version suffix to debug builds more consistently, making it clear when a build is a debug build even if we can't get a commit hash.
* Modified the generator expression to apply the version suffix to `CMAKE_BUILD_TYPE=Debug` only. `MinSizeRel` and `RelWithDebInfo` will no longer have a version suffix.

Based on feedback from https://github.com/diasurgical/devilutionX/pull/2934#issuecomment-993789379

This replaces #3758
This resolves #3382

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,10 +99,13 @@ if(NOT VERSION_NUM)
   if (NOT "${VERSION_NUM}" STREQUAL "")
     string(REGEX MATCH "([0-9]+\\.[0-9]+\\.[0-9]+)" VERSION_NUM ${VERSION_NUM} )
   endif()
-  get_git_commit_hash(GIT_COMMIT_HASH)
   if(NOT VERSION_SUFFIX)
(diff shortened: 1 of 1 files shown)


AJenbo:
I think the issue might be here actually: https://github.com/diasurgical/devilutionX/blob/master/tools/make_src_dist.py#L165

AJenbo (approved):
Looks like i'm just tired

PR #3762, opened 2021-12-15
CMake: Move assets to Assets.cmake

glebm:


AJenbo (approved):

AJenbo:
I wonder if we can do something similar to make VC less confused about where to add new files to the project.

glebm:
VC?

AJenbo:
Visual Studio

PR #3764, opened 2021-12-15
Fixed an issue that caused the app to run as an iPhone app on iPad.

bubio:
- To run as an iPad app, the UIDeviceFamily key is required in Info.plist.

https://discord.com/channels/518540764754608128/518541192993046562/920602213901025300

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -16,7 +16,7 @@ Note that ```pkg-config``` is an optional dependency for finding libsodium, alth
 
 ### Installing dependencies on Debian and Ubuntu
 ```
-sudo apt-get install cmake g++ libsdl2-dev libsodium-dev libpng-dev libbz2-dev libgtest-dev libgmock-dev
+sudo apt-get install cmake g++ libsdl2-dev libsodium-dev libpng-dev libbz2-dev libgtest-dev libgmock-dev libsdl2-image-dev libfmt-dev
 ```
 ### If you want to build the translations (optional)
 ```


AJenbo:
Could you check if it still works when removing https://github.com/diasurgical/devilutionX/blob/master/CMakeLists.txt#L755

bubio:
> Could you check if it still works when removing https://github.com/diasurgical/devilutionX/blob/master/CMakeLists.txt#L755

It is better not to delete it.
Since Xcode is used for the simulator, Xcode will rewrite it into an iPhone app (the UIDeviceFamily will be only 1).

Trihedraf (dismissed):

AJenbo:
We might also want to set `UIStatusBarHidden true`

Trihedraf:
> We might also want to set `UIStatusBarHidden true`

On my iPhone and iPad it is already hidden, but if there is some other scenario that makes it not be, then we should.

AJenbo:
> On my iPhone and iPad it is already hidden, but if there is some other scenario that makes it not be, then we should.

I think the way it currently works is depricated for the next version of the iOS

Packaging/apple/Info.plist
```diff
@@ -48,7 +48,31 @@
...
+	<key>UISupportedInterfaceOrientations~ipad</key>
+	<array>
+		<string>UIInterfaceOrientationLandscapeLeft</string>
+		<string>UIInterfaceOrientationLandscapeRight</string>
+	</array>
+	<key>UISupportedInterfaceOrientations~iphone</key>
+	<array>
+		<string>UIInterfaceOrientationLandscapeLeft</string>
+		<string>UIInterfaceOrientationLandscapeRight</string>
+	</array>
```
Trihedraf, line 74:
```suggestion
```
This is all handled already by SDL_SetHint https://github.com/diasurgical/devilutionX/blob/98cd3613a840b93ea2aaf78fc12bfebd231a72ce/Source/utils/display.cpp#L178

bubio: I think it should be set in Info.plist as well.

Trihedraf: I think redundant settings are a waste of lines. But it's up to @AJenbo really.

bubio: If it is not listed in Info.plist, it will always start with UIInterfaceOrientationLandscapeRight.

Trihedraf: Oh well, that's a valid reason, carry on. 👍

Trihedraf (approved):

Packaging/apple/Info.plist
AJenbo: that might fix some of the startup flickering

AJenbo:
Could one of you test with the latest master and see if the app becomes unresponsive after changing resolution?

Trihedraf:
> Could one of you test with the latest master and see if the app becomes unresponsive after changing resolution?

The first time I click on it it doesn't show the selection of resolution menu. It switches to native resolution. Then it acts a little weird, but if I click it again it switches back to 640x480 and becomes unresponsive.

bubio:
> Could one of you test with the latest master and see if the app becomes unresponsive after changing resolution?

Are you saying that the plist settings are related to this problem?

AJenbo:
im not really sure what the cause is.

AJenbo:
@bubio could you create this PR again, I haven't had time to test an merge yet :(

PR #3766, opened 2021-12-16
Another stab at implementing a Stash

AJenbo:
Here is @kphoenix137 work on implementing a stash for DevilutionX. It still needs a bit of work to be fully implemented and there are probably something in the branch that are not related to the stash and should be excluded from the change set.

Afaik he would appreciate a helping hand if anyone wants to join in on the process :) See https://github.com/diasurgical/devilutionX/pull/3768 for the latest rebase of this PR.


AJenbo:
https://github.com/diasurgical/devilutionX/pull/3836

PR #3770, opened 2021-12-16
paths.cpp: Remove `AppPath`

glebm:
It was redundant.

### DIFF
--- a/Source/utils/paths.cpp
+++ b/Source/utils/paths.cpp
@@ -20,7 +20,6 @@ namespace paths {
 
 namespace {
 
-std::optional<std::string> appPath;
 std::optional<std::string> basePath;
 std::optional<std::string> prefPath;
 std::optional<std::string> configPath;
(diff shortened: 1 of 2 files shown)


StephenCWills (approved):

glebm:
@AJenbo looks like "testflow" is stuck

AJenbo:
This breaks the scenario where someone has there MPQs in a non default folder specified by `--data-dir` and do not have `smpq` installed :(

@obligaron will this be an issue for you?

glebm:
@AJenbo It makes that scenario more sensible IMO because `assets/` folder will be looked for in the data directory instead of next to the binary.

glebm:
We could add an explicit `--assets-dir` flag though, if there is a scenario I'm not thinking of.

glebm:
@AJenbo Sorry for merging before getting feedback btw. I saw that it got stuck and put it on auto-merge. But then it got unstuck

StephenCWills:
Oops, I didn't think about the case where `BasePath` is set independently. I remember now why this distinction was needed.

AJenbo:
> @AJenbo Sorry for merging before getting feedback btw. I saw that it got stuck and put it on auto-merge. But then it got unstuck

No worries :)

PR #3772, opened 2021-12-16
AssetsPath: Default to exe dir

glebm:
Also adds `SetAssetsPath`, for use by tests.

See discussion in #3770


AJenbo (approved):

obligaron:
With this pr I can open a game even `devilutionx.mpq` is not present and `--data-dir` is set to another folder.
So the developer experience without `smpq` installed is good. 👍

PR #3774, opened 2021-12-16
Emscripten SDK support

glebm:
Adds support for the Emscripten SDK via CMake.

The resulting wasm file is only 3 MiB:

```
$ du -sh build-em/devilutionx.wasm 
3.0M	build-em/devilutionx.wasm
```

The libraries are as follows:

```
-- 📚 SDL2: Emscripten system library via USE_SDL=2;USE_PTHREADS=1
-- 📚 SDL_image: Emscripten system library via USE_SDL_IMAGE=2;SDL2_IMAGE_FORMATS='["png"]'
-- 📚 SDL_audiolib: static library from source
-- 📚 libsodium: static library from source
-- 📚 libfmt: static library from source
-- 📚 bzip2: static library from source
-- 📚 zlib: Emscripten system library via USE_ZLIB=1
```


glebm:
Let's merge this -- it doesn't affect other platforms while making experimenting easier

Source/utils/paths.cpp
```diff
@@ -90,8 +90,12 @@ const std::string &ConfigPath()
 
 const std::string &AssetsPath()
 {
+#if __EMSCRIPTEN__
+	assetsPath.emplace("assets/");
```
obligaron, line 94: Doesn't this add `assets/` every time?
```suggestion
	if (!assetsPath)
		assetsPath.emplace("assets/");
```

glebm: Ah good catch! Fixed

AJenbo (commented):
Do we expect TCP multiplayer to work?

glebm:
No idea but it builds 🤷‍♂️😅

CMake/functions/emscripten_system_library.cmake
```diff
@@ -0,0 +1,18 @@
...
+#
+# Arguments:
+#   LIB_NAME: a human-readable library name.
+#   TARGET_NAME: the library target name
+#   ...ARGN: Emscript flags.
+#
+# Example:
+#   emscripten_system_library("SDL2_image" SDL2::SDL2_image USE_SDL_IMAGE=2 "SDL2_IMAGE_FORMATS='[\"png\"]'")
+function(emscripten_system_library LIB_NAME TARGET_NAME)
+  set(_lib_target "EMSCRIPTEN_${LIB_NAME}")
```
StephenCWills, line 11: This variable doesn't appear to be used.

glebm: Fixed

StephenCWills (approved):

PR #3775, opened 2021-12-17
Prefer DX11 renderer if running on windows with vsync active

ephphatha:
This works around a bug in DX9 on Win11.


ephphatha:
I don't have access to a system running a previous version of windows to test this, would be interesting to see how it runs on a system without DX11 support (early Vista or XP)

obligaron:
Tested on windows 10 (changed resolution and vsync option). Still works fine. 🙂

AJenbo (approved):

PR #3779, opened 2021-12-17
settingsmenu: In list selection show option description

obligaron:
Master | PR
--- | ---
grafik | grafik

### DIFF
--- a/Source/DiabloUI/settingsmenu.cpp
+++ b/Source/DiabloUI/settingsmenu.cpp
@@ -7,6 +7,7 @@
 #include "hwcursor.hpp"
 #include "options.h"
 #include "utils/language.h"
+#include "utils/utf8.hpp"
 
 namespace devilution {
 namespace {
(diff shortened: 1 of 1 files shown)


Source/DiabloUI/settingsmenu.cpp
```diff
@@ -73,16 +73,24 @@ void GoBackOneMenuLevel()
 	shownMenu = ShownMenuType::Settings;
 }
 
+void UpdateDescription(OptionEntryBase *pOption)
+{
+	auto paragraphs = WordWrapString(pOption->GetDescription(), rectDescription.size.width, GameFont12, 1);
+	strncpy(optionDescription, paragraphs.c_str(), sizeof(optionDescription));
```
ephphatha, line 79: GetDescription returns a translated string of potentially multibyte utf8 codepoints, might as well use CopyUtf8 here to be safe.
```suggestion
	CopyUtf8(optionDescription, paragraphs, sizeof(optionDescription));
```

obligaron: Fixed 🙂

Source/DiabloUI/settingsmenu.cpp
```diff
@@ -73,16 +74,24 @@ void GoBackOneMenuLevel()
 	shownMenu = ShownMenuType::Settings;
 }
 
+void UpdateDescription(OptionEntryBase *pOption)
+{
+	auto paragraphs = WordWrapString(pOption->GetDescription(), rectDescription.size.width, GameFont12, 1);
+	CopyUtf8(optionDescription, paragraphs, sizeof(optionDescription));
+}
```
ephphatha, line 81: This function can take `const OptionEntryBase &option` as the argument. Polymorphic calls still work on references even if it was using a virtual method.

ephphatha (approved):
Functions fine, good to have context while changing a setting (especially for those of us with dodgy short term memory 😂 )

Source/DiabloUI/settingsmenu.cpp
obligaron: Changed 🙂

ephphatha (approved):

PR #3780, opened 2021-12-17
.gitattributes: Do not let git change line endings

glebm:


Source/engine/demomode.cpp
```diff
@@ -1,301 +1,295 @@
-/**
- * @file animationinfo.h
- *
- * Contains most of the the demomode specific logic
- */
-
```
AJenbo, line 6: what happened to the header?

glebm: Third commit -- this header did not belong here as it's a cpp file (the name was wrong anyway)

AJenbo (approved):

Issue #3776, opened 2021-12-17
Refactor UI+game to a single main loop


glebm:
If we ever want to port to a platform where we're not in control of the main loop, such as Emscripten, we need to refactor how we handle rendering and events first.

Currently, each "screen" in the menu has its own main loop, and the game has yet another main loop.

Chance4us:
> If we ever want to port to a platform where we're not in control of the main loop, such as Emscripten, we need to refactor how we handle rendering and events first.
> 
> Currently, each "screen" in the menu has its own main loop, and the game has yet another main loop.

That's right,  unfortunately it is like that, and as far as I have seen there are several "rendering surface" confusions.

AJenbo:
Here is a map of the menu flow:

![image](https://user-images.githubusercontent.com/204594/216949444-ce17df38-00b7-4216-87cc-95157876e1d2.png)

glebm:
The hero screen can also have an error if there are more than the maximum number of characters

AJenbo:
I'm starting this by restructuring the menus in to groups with sub panels.

Panels in a group will share the same background and loads/frees a common set of resources. Only when navigating from one group to another will assets be loaded/freed, this is also when the fade transition happens.
It's possible to jump to an arbitrary of any group (mainmenu -> hero/list -> hero/load -> multi/difficulty -> game) 

I'm going to try and wrangle the various properties that the menus interact with in a common place, rather then have them scattered around the game and various panels.

credits/-
error/-
mainmenu/-
ok/-
progress/-
title/-
yesno/-

hero/class
hero/create
hero/list
hero/load
hero/name

hub/chat
hub/create
hub/frinds
hub/join

multi/difficulty
multi/join
multi/list
multi/provider

options/groups
options/list
options/bind
options/enum

This probably won't get us directly to a well organized/structured system in one go, but I think it will at least be a step in the right direction

PR #3782, opened 2021-12-17
RFC: Unify some event handling and rendering loops

glebm:
As outlined in #3776, we want to have a single main loop in order to clean up the UI code and be compatible with Emscripten.

This is my attempt at this that:

1. Adds a library for handling the main loop stuff (`Source/main_loop`).
2. Migrates a few main loops to the new approach: startup, movie playback, and main menu to the single loop.

@AJenbo @obligaron @StephenCWills


julealgon:
What does "RFC" mean in this context?

AJenbo:
https://en.wikipedia.org/wiki/Request_for_Comments

obligaron:
It's a good idea and I think we can improve the approach.

Logically we have multiple "dialogs"/"menus".
They are not created randomly and you can't jump from one to another.
They are build as a stack.
If you leave the top most you get handler before it.

Startup: Main Menu -> Title -> Video (Intro) -> Video(Logo)
InGame: Main Menu -> GameHandler
InGameVideo: Main Menu -> GameHandler -> Video(Diablo Death)

For this we need to add a `Activated` / `Deactivated` callback to `MainLoopHandler`, cause a menu can be activated multiple times.

Special case is the singleplayer / multiplayer selection.
But even there we could:
Multiplayer: Main Menu -> Hero Selection
(Escape pressed, closes Hero selection and creates "new" dialog multiplayer selection)
Multiplayer: Main Menu -> Connection selection

I experimented with that approach and I think it feels more natural, see the following [branch](https://github.com/obligaron/devilutionX/commits/loop-handler2).

If you dislike the idea, feel free to discard it. 😉

glebm:
In your branch you've removed the factories and instead replaced construct/destructor with Activate/Deactivate.

One issue with that is that destruction is now manual, so it's possible to keep the resources around accidentally.

Perhaps we can have 2 classes for each "screen"? That's somewhere between your approach and mine.
So we push the "manager" class on the stack, the "manager" has a "const View &View()` method, and the "view" class is the one that receives events. The "manager" constructs and deconstructs the "View" based on activation.

Another issue in your branch is that `Activated` is only called on the next loop iteration instead of immediately.

obligaron:
> One issue with that is that destruction is now manual, making it possible to keep the resources around accidentally (we've had bugs like this in the past, e.g. #1380).

Just to make clear:
It's not manual in the sense that the implementation of a `MainLoopHandler` doesn't need to call these methods.
It's manual in the sense of that you need to ensure that you free your memory in `Deactivated` instead of destructor.
(added one small bugfix where `Deactivated` is sometimes called twice)

> Perhaps we can have 2 classes for each "screen"? That's somewhere between your approach and mine. So we push the "manager" class on the stack, the "manager" has a "view" instance in it and the "view" instance is the one that receives events. The "manager" constructs and deconstructs the "View" based on activation.

That could be possible and makes this cleaner.
The only downside I see is that it's also make it more complicated.
Cause now you have to pass the parameter for the "screen" (for example the movie to play) twice or the view needs a pointer to the manager that contains the parameter.

> Another issue in your branch is that `Activated` is only called on the next loop iteration instead of immediately.

This is needed, cause at startup we add Mainmenu, Title-Dialog, Movie(Intro) and Movie(Intro).
But only Movie(Intro) is activated at the beginning. The other `MainLoopHandler` are only activated when they are shown.
Else all would be enabled at the same time or twice activated/deactivated.

glebm:
> It's manual in the sense of that you need to ensure that you free your memory

Yeah that's exactly what I meant by manual memory management. That is the dangerous part that caused issues in the past.

>  Cause now you have to pass the parameter for the "screen" (for example the movie to play) twice or the view needs a pointer to the manager that contains the parameter.

Yeah I think it makes sense for the view to be able to access the manager's data.

> This is needed, cause at startup we add Mainmenu, Title-Dialog, Movie(Intro) and Movie(Intro).

I think it should be possible to figure out a way to not wait for the next loop iteration (e.g. in this PR activation is instant).

obligaron:
> I think it should be possible to figure out a way to not wait for the next loop iteration (e.g. in this PR activation is instant).

Do you mean with instant that we update the active `MainLoopHandler` inside of `RunMainLoopIteration`?
This would allow to render the new content a little bit earlier.
If yes, I tried this with the following [commit](https://github.com/obligaron/devilutionX/commit/34d5c2c6ea667c04e24468aada099ad4b93c2fa3).

glebm:
@obligaron I think I quite like your version once the manual memory is fixed.
Hopefully we can have something like this once 1.4.0 is released and then parallelize the work on migrating individual screens

ephphatha:
> Logically we have multiple "dialogs"/"menus". They are not created randomly and you can't jump from one to another. They are build as a stack. If you leave the top most you get handler before it.
> 
> Startup: Main Menu -> Title -> Video (Intro) -> Video(Logo)  
> InGame: Main Menu -> GameHandler  
> InGameVideo: Main Menu -> GameHandler -> Video(Diablo Death)

I tend to model menu transitions as a state machine personally, as this allows transitioning from one submenu to another directly (if there's a path where that makes sense) instead of having to traverse back up the stack and down another flow. This allows modelling the way the vanilla menu code works more directly. The current implementation is a state machine using function pointers to define transitions between states (and triggering alternate flows with global variables -_-).

Mixing state machines and stacks is fine for the uses which make sense in one form or the other. It's even possible to use a state machine to implement a stack (by transitioning to a state which transitions back to its caller) or vv.

This PR appears to be using the `HandlersBeingCreated` vector as a stack implementing a state machine, which I think could be simplified a bit given the way certain transitions work.

I'm not personally a fan of the way MainMenuDialog (and TitleDialog) are named to begin with, it's a bit of a misnomer given there are actual dialogs in the form of the Dialog class/appFatal and friends. I think this would be more appropriate as just MainMenu/TitleScreen.

Given the MainMenuDialog isn't really a dialog I don't think it's appropriate to use a stack in the way this PR does to handle transitions for this class. It looks like MainMenuSelected becomes a wrapper state that exists because MainMenuDialog can't transition to different targets? Moving those calls in to `MainMenuDialog::Select` or tying the functions to the list items in the constructor would remove the need for this wrapper class. It looks like this is the way MovieMainLoopHandler/TitleDialog work.

Recreating the handler at the start of the `MainMenuSelected()` constructor is a bit of a weird way to loop (assuming I understand the purpose of that code correctly), is this purely so attract mode goes back to the main menu? I still don't think I understand this code and I've read it multiple times now.

Having said that it's certainly an improvement on the current code in my eyes. Splitting the event handling and render logic is a good idea, and deduplicating some redundant loops certainly doesn't hurt. Only real downside in my mind is the complicated way MainMenuSelected loops in this PR.

obligaron:
@ephphatha I didn't wanted to push commits in glebms pr. That's why i created a own [branch](https://github.com/obligaron/devilutionX/commits/loop-handler2), In this branch the `HandlersBeingCreated` and `MainMenuSelectedHandler` is removed and replaced with the mentioned stack. 🙂
Sorry this was confusing as we kept discussing in this pr but with the [other branch](https://github.com/obligaron/devilutionX/commits/loop-handler2) in mind.

glebm:
This is a lot easier to implement now that we go rid of all the indirection in event handling.
Going to close this PR for now, hopefully someone can give it another go in the future.

AJenbo:
I had a bit of time to look at all of this finally :sweat_smile: 

Splitting rendering and event handeling from each other is probably a good idea no matter which approach is taken.

Another issue i find is that some menu pages feels like they are really two merged into one in an attempt to avoid code duplication (this is probably all my fault). So separering that so we at least have a 1:1 between logical pages and code units would probably be the best place to start with detangling things.

The other general issue is that the code has at this point gone through so many people, ideas, stages and features that it has bits and pieces of ever design pattern imaginable. The original often used function pointers and callbacks, bit that is mostly gone now replaced by globals like the rest of the game. But still not very intergrated with the rest.

I know one thing that made the system hard for me to grasp is how you return to the menu after a game is done. My intuition was that the would start the menu back up, but in reality it never really went a way (this also contributed to the issue with when to free game ressources, and when to load then back in again).
Formalizing this as a stack would probably make this a lot easier to understand. But there are a few instances where I don't think it fully fits:

### Quitting from in game
Here we do not actually return to the previous element on the stack. Instead we have to have them either know about the global "exit" state or flush them before coming back to them.

### Starting a game via commands or invites:
Here we might entirely skip the menus, dialogs and intros and instead load directly into a game. But when ending the game we might want to return to the normal menu. But then how do we make sure the stack was correctly populated in order for us to later be able to to "back" to the menu?

### Endgame video
At the end of the game 3 videos are shown and then you are returned to the menu. It feels odd that they are placed on top of the game, yet when they are done we do not return to the game, instead the game needs to instantly shut down so that we drop to the menu. It's probably possible to remove the game first by doing some shuffling.

----

Another good step on the road to a cleaner menu system I think would be to make them less of a driver in how to manage things (network, start a game, initializing various things) would be to demote them to be just one of a couple of ways to configure some structures in the engine (similar to how it's able to change game settings, but not the only way to do so).

I think in the end i favor a state machine design where there isn't an explicit direction, but esc/back is just another way to move to a different panel. I'm hoping that doing so would also make it much cleaner when to load and unload resources.
My biggest concern would be how to implement modals and how to maintain highlighted items on lists. On the other hand it's not something that the current system does a particularly good job of (esc in the password screen and you will end up on the speed list with a different option then you picked), and having a consistent pattern to follow will probably make it more likely that we account for the right things.

glebm:
> I think in the end i favor a state machine design where there isn't an explicit direction, but esc/back is just another way to move to a different panel.

After reading the above, I agree.
For handling the esc/back and list selection, we could still have an optional stack but not as the primary mechanism for state transitions.

AJenbo:
Well in that case:
image

AJenbo:
Map updated with labels and a couple of corrections

glebm:
Perhaps we should track it in the open issue (https://github.com/diasurgical/devilutionX/issues/3776) rather than the closed PR for discoverability?

PR #3785, opened 2021-12-17
Disable recreation of the rendere on Apple systems

AJenbo:
Turns out the issue is the same on MacOS 10.3 (intel GPU), I'm assuming it will be the same for iOS and friends.

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -67,10 +67,10 @@ namespace devilution {
 
 namespace {
 
-#ifdef __ANDROID__
-constexpr OptionEntryFlags InvisibleOnAndroid = OptionEntryFlags::Invisible;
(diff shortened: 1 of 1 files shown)


obligaron:
I would make it invisible as we did it previous with android.
Else we need a ifdef in:
- Header
- Constructor
- `GetEntries()`

This would also allow to override it in the `ini`.

StephenCWills:
> Else we need a ifdef in:
> 
> * Header
> * Constructor
> * `GetEntries()`

Not to mention all the places we use the variable, which could be significantly more error prone since they're all if statements.

AJenbo:
Yeah this didn't turn out as nice as I had hoped :( Please check again.

obligaron (approved):

StephenCWills (approved):

PR #3790, opened 2021-12-17
Discard portrait resolutions

AJenbo:


Trihedraf:
Would it be possible to instead be able to rotate to vertical for the portrait resolutions? I like being able to play games without having to flip my phone on it's side.

AJenbo:
Maybe at some distant time in the future, but right now everything is assuming that the display is in landscape mode meaning that setting the resolution doesn't work as it's actually not supported. When that time comes we probably won't even need the portrait resolution as this is pretty unique to iOS and everything else expects you to swap them your self, so it would all be handled by generic code at that point.

StephenCWills (approved):

Source/utils/display.cpp
```diff
@@ -110,8 +110,32 @@ Size GetPreferredWindowSize()
 	AdjustToScreenGeometry(windowSize);
 	return windowSize;
 }
+
 } // namespace
 
+float GetDpiScalingFactor()
```
obligaron, line 116: This method confuses me a little bit.
Does iOS / `SDL_GetDisplayMode` report wrong resolutions?

Could this [otherwise](https://discourse.libsdl.org/t/ios-full-screen-resolution-issue/27178) be fixed?

StephenCWills: Given the peculiarity of this code, I assumed this must have been tested for correctness, but I missed the error you pointed out above so now I'm pretty skeptical. Where did this function come from?

AJenbo: It's based on how we handle iOS resolutions in the touch UI and when spawning the render. Both had to be adjusted to handle iOS correctly as it has this stupid concept of "Retina resolution", basically all resolutions re reported as if the DPI was that of the original iPhone and a multiplier factor then has be added to figure out the true nature of the display. To make things worse there are even some iPhones that additionally renders at a higher resolution then the screen and then down scales it when outputting, in order to have clean devisions for there multiplyer.

The fact that SDL chose to follow this maddness and didn't just abstract it away for us boggles my mind.

p.s. to get a buffer with the full size of the screen (resolution, and widescreen) we must provide a flag when creating the window, set a special value in the application manifest and make a (black) splash screen in the same dimensions as the real screen. Else the application will be limited to the same resolution as the original iPhone even when running on an iPad.

Regarding the bug in the function, I do the testing via a remote connection to a Mac that runs the emulator and then write the same code on my main machine and submit it a PR. So it's a bit more error prone then usual. Additionally I just got sick (covid test pending) so struggling a bit to focus.

Sorry for the low quality code, long text and notifications. Thanks you for reviewing and feed back :)

AJenbo: > Could this [otherwise](https://discourse.libsdl.org/t/ios-full-screen-resolution-issue/27178) be fixed?

We already do all of this, but that is just for window creation. Almost all functions lye about the resolution, except for `SDL_GL_Drawable()` and `SDL_RendereInfo()`.

Source/utils/display.cpp
```diff
@@ -110,8 +110,32 @@ Size GetPreferredWindowSize()
...
 }
+
 } // namespace
 
+float GetDpiScalingFactor()
+{
+#ifdef USE_SDL1
+	return 1.0F;
+#else
+	if (renderer != nullptr)
```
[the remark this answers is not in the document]
AJenbo: Thanks it now works as expected

image

image

Source/utils/display.cpp
StephenCWills: > Regarding the bug in the function, I do the testing via a remote connection to a Mac that runs the emulator and then write the same code on my main machine and submit it a PR. So it's a bit more error prone then usual. Additionally I just got sick (covid test pending) so struggling a bit to focus.

In other words, you determined the need for this function during your testing on the Mac. So it was tested for correctness, but the bug slipped in while submitting the PR.

I fully expect it will work fine because the multiplier should be `1.0f` either way, but let me test one more time on Android with the bug fixed and then I'll resubmit approval.

AJenbo: Correct, and appreciated :)

StephenCWills (approved):

Source/utils/display.cpp
obligaron: > The fact that SDL chose to follow this maddness and didn't just abstract it away for us boggles my mind.

Me too. That's why i asked. 😄 

> p.s. to get a buffer with the full size of the screen (resolution, and widescreen) we must provide a flag when creating the window, set a special value in the application manifest and make a (black) splash screen in the same dimensions as the real screen. Else the application will be limited to the same resolution as the original iPhone even when running on an iPad.

W.T.F.

> Sorry for the low quality code, long text and notifications. Thanks you for reviewing and feed back :)

No problem. There are many things to do and I'm impressed how you manage all of this. ❤️

AJenbo: Yes the resolution shenanigans appears to be exclusive to Apple devices (I haven't had a chance to test a recent Mac, and haven't tried on watchOS or tvOS. I decided to implement the solution in general in case any thing else decides to copy them.

There is one Android review that seams to indicate that at least some Android devices will report there resolutions in portrait mode instead of landscape mode.

obligaron (dismissed):

Source/utils/display.cpp
StephenCWills: @obligaron FYI, you might want to refer to the SDL documentation for `SDL_GetWindowSize()`. It doesn't exactly clarify anything definitively, but it does at least suggest that the `SDL_GetRendererOutputSize()` function was introduced exclusively to deal with Apple's high-dpi quirks. The wording leaves things open for other platforms that might want to mimic this nonsense, but I doubt we can expect much variation in how all these functions work.

https://wiki.libsdl.org/SDL_GetWindowSize
> The window size in screen coordinates may differ from the size in pixels, if the window was created with `SDL_WINDOW_ALLOW_HIGHDPI` on a platform with high-dpi support (e.g. iOS or macOS). Use `SDL_GL_GetDrawableSize()`, `SDL_Vulkan_GetDrawableSize()`, or `SDL_GetRendererOutputSize()` to get the real client area size in pixels.

Also, speaking from my own experience with rendering the virtual gamepad, `SDL_GetRendererOutputSize()` should return exactly the same metrics as `SDL_GetWindowSize()` on most platforms. I found that Windows systems have their own brand of high-dpi quirks, but even so they don't differentiate between window size and renderer output size from what I was able to tell.

obligaron: Now it makes more sense to me. I only developed on windows, so I'm not that familiar with the other environments. Thanks for clarifying. 🙂

AJenbo: On a side note `SDL_GL_GetDrawableSize()`, `SDL_Vulkan_GetDrawableSize()` aren't the best choices here since the render might be Metal in which case both would just error :)

PR #3795, opened 2021-12-19
Use span filling algorithm for FillTransparencyValues

### LINKED ISSUE #3543: Improve `FindTransparencyValues` to use less stack
From @StephenCWills in #3540:

> [single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/7568499/single_0.hsv.zip) causes the game to crash on 3DS when entering dungeon level 13, due to the amount of recursion in `FindTransparencyValues()`.
> 
> If you're not familiar with the function, it seems to be using a flood fill algorithm to mark contiguous regions of floor, as well as the adjacent and surrounding non-floor spaces, with the same "transparency value". Based on the name of it, I would guess this is used to determine which walls should be rendered transparent when the player enters a room. However, it also seems to be used in a few other places, such as when determining whether a monster is in the same room as another entity.
> 
> My testing reveals that the function reached about 470 levels of recursion. If my estimate is right, 128 KB should barely cover this. I guess the big question is whether this savegame represents an extreme or whether we should expect to encounter even deeper recursion in another dungeon layout.
> 
> Here are some thoughts I had while messing with this.
> * Add some additional stack space to be safe, but I'm not sure how much.
> * Rewrite the function to use heap space instead.
> * Select a different algorithm with lower memory overhead.
>   * https://en.wikipedia.org/wiki/Flood_fill

This option from Wikipedia seems like a good algorithm:

> The final, combined-scan-and-fill span filler was then published in 1990 [...]

Faster and uses a lot less space.

StephenCWills:
One peculiar thing about the original algorithm was that it traversed megatile coordinates to fill the corresponding tile coordinates. A careful review of that code reveals that the pattern in which it filled the corresponding tiles resulted in the floor being completely filled, plus all the immediately surrounding non-floor tiles. Rather than carrying that strategy over to the rewrite, I concluded that it would be simpler to have the fill algorithm traverse the floor tiles and then fill in the surrounding non-floor tiles as it went.

Additional notes:
* The algorithm as written on Wikipedia scans left but not right. However, tweaking the initial seed is enough to fix it.
* The lambdas help to make the function resemble the Wikipedia pseudocode pretty closely, but I did rename the variables to hopefully help clarify what the algorithm is doing.
* I had to add some code to check diagonals after all. Details are in the linked issue.

This resolves #3543

### DIFF
--- a/Source/gendung.cpp
+++ b/Source/gendung.cpp
@@ -3,6 +3,8 @@
  *
  * Implementation of general dungeon generation code.
  */
+#include <stack>
+
 #include "gendung.h"
 
 #include "engine/load_file.hpp"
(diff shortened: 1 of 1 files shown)


AJenbo (approved):

AJenbo:
I guess this should be upstreamed to Wikipedia :thinking:

Issue #3778, opened 2021-12-17
Broken test execution on VS2022 since #3746


tsunamistate:
**Important information**
Windows 10 / 0d44d955e5f56fa938e524b55d32b39c317f5ed7 / self-compile

**Describe**
Since #3746 was merged, test execution is now part of build process for all configurations (why?).
Test execution fails with the following logs:
```
[190/200] Linking CXX executable test\dead_test.exe
  FAILED: test/dead_test.exe test/dead_test[1]_tests.cmake D:/src/devilutionX/build/x64-Debug/test/dead_test[1]_tests.cmake 
[...]

PR #3797, opened 2021-12-19
Fix compiling tests with sdl 1

obligaron:
See [comment](https://github.com/diasurgical/devilutionX/issues/3778#issuecomment-997288942).
Introduced in #3777.


ephphatha (approved):
Building master with SDL1 fails, this branch works :)

Issue #3791, opened 2021-12-18
[touch] Changing resolution does not recalculate controler position


AJenbo:

obligaron:
```
#ifdef VIRTUAL_GAMEPAD
	InitializeVirtualGamepad();
#endif
```

If we include this at the buttom of  `ResizeWindow` fix the issue?
I don't have a android development environment ready, so can't test.

Issue #3788, opened 2021-12-17
[ios] Mouse emulation dies after changing graphics settings


AJenbo:
The menu becomes unresponsive after changing one of the following graphics options:
  - [ ] Resolution
  - [ ] Fit to screen
  - [ ] scaling quality
  - [x] integer scaling
  - [ ] vertical sync

This happens even if scaling quality is isolated to the following, but only about 1/3 of the time:
```cpp
void ReinitializeTexture()
{
	if (texture)
		texture.reset();

	auto quality = fmt::format("{}", static_cast<int>(*sgOptions.Graphics.scaleQuality));
	SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, quality.c_str());

	texture = SDLWrap::CreateTexture(renderer, SDL_PIXELFORMAT_RGB888, SDL_TEXTUREACCESS_STREAMING, gnScreenWidth, gnScreenHeight);
}
```

Printing out the events shows that we still get touch events, but no longer any mouse emulation. Which is actually in line with what we requested. But when we had mouse emulation enabled from the start we would receive 2x the number of expected mouse events.

bubio:
https://github.com/diasurgical/devilutionX/blob/74a77666b9676057822f26e4b3a494197ab6e8fc/Source/utils/display.cpp#L331

When I run SDL_CreateRenderer, the touch doesn't work anymore.
But if you change the settings on the keyboard, you're good to go.

bubio:
https://github.com/libsdl-org/SDL/blob/11ae43ca167476070225d493a642fe6e5bbb96da/src/video/uikit/SDL_uikitview.m#L287

Before releasing the finger, the Window, etc. will be rebuilt, so touchesEnded will not occur and the internal state of SDL will be wrong.

AJenbo:
Maybe this should be reported as an issue to libsdl?

bubio:
https://github.com/diasurgical/devilutionX/blob/74a77666b9676057822f26e4b3a494197ab6e8fc/Source/DiabloUI/diabloui.cpp#L924

```cpp
if (event.type != SDL_MOUSEBUTTONUP || event.button.button != SDL_BUTTON_LEFT)
```

I think this problem can be fixed by changing the timing for enabling touch to when the finger is released.

AJenbo:
i dont think i follow, maybe you can ma a test buid?

bubio:
> i dont think i follow, maybe you can ma a test buid?

[...]

PR #3803, opened 2021-12-20
Refactor graphic option callbacks

### LINKED ISSUE #3697: Setting Fullscreen to ON does not return game to full screen
**Important information**
master, self build, linux, radeon

**Describe**

**To Reproduce**
1. Go to Settings
2. Switch Fullscreen off
3. Switch Fullscreen on

**Additional context**
alt+enter still works correctly for setting the game back to fullscreen

Several other graphics options will force the game to ext fullscreen mode.

Scaling quality works as expected.

AJenbo:
Fixes #3697 #3791

Only the relevant parts are reset when changing graphics options.

- Turning fullscreen on in settings doesn't flicker on Mac
- `-x` has been removed
- `alt+enter` changes the setting instead of being temporary
- Allow systems with no GPU to use CPU based rendering (this was an issue for [SerenityOS](https://github.com/SerenityOS/serenity) when @awesomekling [ported DeviutionX](https://www.youtube.com/watch?v=ZOzZ8R4gphE) to it, despite the CPU being powerful enough). Even if the CPU isn't powerful enough it should allow users to switch off upscalling in settings for better performance, so win win.
- When exiting fullscreen mode the window is correctly resized instead of being semi maximized
- Flipping interger scalling no longer break input on iOS (#3788)
- Remove stripped transparancy
- #3791
- Fixes window not updating after resizing on linux


StephenCWills (changes requested):
Something is definitely not right with integer scaling. It doesn't really seem like it was working correctly before this PR either, but with this PR I'm able to cause graphical glitches and crashes with certain combinations of integer scaling and fullscreen changes that weren't happening before. For instance, the following process never seems to fail to cause such issues.

1. Start game in fullscreen
2. Enter settings menu and change Integer Scaling
3. Exit the settings menu and enter a new single player game
4. Use Alt+Enter to toggle fullscreen

Source/utils/display.cpp
```diff
@@ -278,16 +300,13 @@ void ReinitializeRenderer()
 	Size windowSize = { current.current_w, current.current_h };
 	AdjustToScreenGeometry(windowSize);
 #else
-	if (texture)
-		texture.reset();
```
StephenCWills, line 282: Better add these back. I got a failed assertion when changing the resolution in the settings menu. It may seem redundant to do it both here and in `ReinitializeTexture()`, but the texture does need to be destroyed before the renderer to avoid errors when we inevitably attempt to destroy the texture later on.

StephenCWills:
This is pretty minor, but using Alt+Enter while in the settings menu doesn't update the value displayed next to the `Fullscreen` setting.

Source/options.cpp
```diff
@@ -792,14 +784,14 @@ GraphicsOptions::GraphicsOptions()
...
+	fullscreen.SetValueChangedCallback(SetFullscreenMode);
 #if !defined(USE_SDL1) || defined(__3DS__)
 	fitToScreen.SetValueChangedCallback(ResizeWindow);
 #endif
 #ifndef USE_SDL1
 	upscale.SetValueChangedCallback(ResizeWindow);
-	scaleQuality.SetValueChangedCallback(ReinitializeRenderer);
-	integerScaling.SetValueChangedCallback(ResizeWindow);
+	scaleQuality.SetValueChangedCallback(ReinitializeTexture);
+	integerScaling.SetValueChangedCallback(ReinitializeIntegerScale);
```
obligaron, line 794: Doesn't integer scaling also affect the potential window size?
(see `CalculatePreferdWindowSize`/ `ResizeWindow` / `SetFullscreenMode`)

Source/utils/display.cpp
```diff
@@ -337,6 +350,34 @@ void ReinitializeRenderer()
...
+		ErrSdl();
+	}
+	if (renderer != nullptr && !*sgOptions.Graphics.fullscreen) {
+		SDL_RestoreWindow(ghMainWnd); // Avoid window being maximized before resizing
+		Size windowSize = GetPreferredWindowSize();
+		SDL_SetWindowSize(ghMainWnd, windowSize.width, windowSize.height);
+	}
+#endif
+	force_redraw = 255;
+}
```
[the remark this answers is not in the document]
StephenCWills: I don't think so. Rendering coordinates should only be affected by changes to the logical size. It's easy enough to test, though, so I'll double-check.

Source/options.cpp
StephenCWills: You hit the nail on the head. The changes to window size are actually the cause of the problem here. Just take them out and integer scaling works entirely as expected for me. Also, `dx_resize()` is no longer necessary.

Source/utils/display.cpp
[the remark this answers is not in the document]
StephenCWills: Yep, I've confirmed that so long as the logical size of the renderer does not change, we don't need to call `InitializeVirtualGamepad()`.

Source/options.cpp
StephenCWills: @AJenbo FYI, it looks to me like `CalculatePreferdWindowSize()` was trying to adjust the resolution to the largest integer multiple of the requested resolution. However, that changes the game's resolution by updating `gnScreenWidth` and `gnScreenHeight` to that integer multiple. I think instead the desired effect is to set the logical size of the renderer to the requested resolution and the window size to the integer multiple in order to get the maximum scale possible for that display mode.

If you like, I can work on a commit for this PR to fix everything that came up based on my testing, and then you can rebase however you like.

AJenbo: Thanks that would be appreciated, haven't been able to do much at all today.

Source/utils/display.cpp
```diff
@@ -268,6 +269,27 @@ bool SpawnWindow(const char *lpWindowName)
...
+
+	texture = SDLWrap::CreateTexture(renderer, SDL_PIXELFORMAT_RGB888, SDL_TEXTUREACCESS_STREAMING, gnScreenWidth, gnScreenHeight);
+}
+
+void ReinitializeIntegerScale()
+{
+	if (SDL_RenderSetIntegerScale(renderer, *sgOptions.Graphics.integerScaling ? SDL_TRUE : SDL_FALSE) < 0) {
+		ErrSdl();
+	}
+	dx_resize();
```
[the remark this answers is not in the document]
StephenCWills: The latest commit merely updates the window size without updating the render logical size. Since the logical size doesn't change, there is no need to reinitialize the texture or the global back buffer.

Source/utils/display.cpp
StephenCWills: I slipped this in with the integer scaling fixes.

Source/options.cpp
StephenCWills: Alright, I fixed all the issues I could find. I'm submitting my approval since AJenbo can't do it himself, but it would be good if someone else reviews my changes before merging.

StephenCWills (dismissed):

StephenCWills:
I thought about this a bit more, and I realize now that the integer scaling logic that updated the game's resolution was probably put in place to eliminate the black borders around the viewport. It actually makes some sense to do it this way when Fit to Screen is on, but we'll have to call `ResizeWindow()` if we intend to change the resolution like that.

AJenbo:
oh yeah, forgot about that.

StephenCWills:
Okay, that was actually significantly less complicated. 😅

StephenCWills (approved):

AJenbo:
Alright went over the changes and fixed two more issues that we introduced :)
- crash when switching integer scaling if upscale is off
- switching upscaling didn't flip resiable window on/off

I think that if either of you are good with the latest changes then we can merge it :)

StephenCWills (approved):

PR #3815, opened 2021-12-21
CMake: Move libdevilutionx to Source/

glebm:


AJenbo (approved):
Just need to move the platform specific packaging out of the main file now :)

glebm:
Yeah, ideally we'd also move all of the platform-specific stuff into `CMake/platforms`
I have an idea on how to do that

PR #3817, opened 2021-12-21
CMake: Simpler Threads workaround for the 3DS

glebm:


StephenCWills (approved):
`find_package()` is so confusing. I guess it's up to the module to produce an error if the package is not found, so the `REQUIRED` flag doesn't really matter if your module doesn't call `find_package_handle_standard_args()`.

Issue #3792, opened 2021-12-18
Do not show hardware cursor options when not supported


AJenbo:
Calling `HardwareCursorDefault()` will let you know if it's supported or not.

https://github.com/diasurgical/devilutionX/blob/master/Source/options.cpp#L230

Additionally we can avoid [this issue](https://github.com/diasurgical/devilutionX/issues/3488) if we simply ignore the ini file when `HardwareCursorDefault()` returns `false`.

obligaron:
I'm not familiar which platform can't support a mouse and have a SDL2 build (cause on SDL1 hardware cursor is not shown/compiled).

- iOS?
- Switch?
- Vita?

On android you can have a (bluetooth) mouse.

StephenCWills:
Android hardware cursor has some issues: #2502

AJenbo:
iOS will simply emit an error to the console when trying to set a hardware cursor. And the feature did not work in general before SDL 2.0.12.

Calling `HardwareCursorDefault()` will let you know if it's supported or not.

https://github.com/diasurgical/devilutionX/blob/master/Source/options.cpp#L230

Additionally we can avoid [this issue](https://github.com/diasurgical/devilutionX/issues/3488) if we simply ignore the ini file when `HardwareCursorDefault()` returns `false`.

obligaron:
Reread #2502

> I'm thinking that the cursor smoothness is less important on Android given that it will most likely be moved using a gamepad (virtual) and rarely an actual mouse. If some one does play it using a mouse on an Android device they are probably enough of a power user that they can change the ini?

Is this not a argument that we should show the option on android (with default off)?
Would this leave only iOS (`TARGET_OS_IPHONE`)?

glebm:
Android can support a hardware cursor but only when a physical mouse is plugged in.

It'd probably make sense to extract a separate function, something like this:

```c++
bool HardwareCursorSupported()
{
	SDL_version v;
	SDL_GetVersion(&v);
	return SDL_VERSIONNUM(v.major, v.minor, v.patch) >= SDL_VERSIONNUM(2, 0, 12);
}
```

obligaron:
Added #3799

It gets only hidden based on the SDL version.
Also iOS also has [mouse cursor support](https://www.youtube.com/watch?v=iH9QZAzWGiA)?
So I thought that iOS should also allow it. It's only deactivated per default, like Android.

AJenbo:
> Reread #2502
> 
[...]

PR #3825, opened 2021-12-22
Touch support on all SDL2 platforms

### LINKED ISSUE #3824: Touch support on all SDL2 platforms
- Track what input type is currently in use
- Hide touch interface when other input types are active
- Enable use of touch controls on all SDL2 platforms

AJenbo:
Fixes #3824

The cursor used for (virtual) gamepad is now fully internal which should help on platforms that do not support mouse warping (Wayland, something with no mouse)

On the flip side when emulating the mouse using a gamepad it actually moves the real mouse.

I have tried to remove as much of the old mess as possible and keep everything isolated, and confined to plrctrl.cpp

Switching input mode is now mostly even driven rather then relying on boolean being set in various places scattered.

Fixed a OOB issue

Do not use hardware cursor when it's known not to be supported (even if set to true in ini) see https://github.com/diasurgical/devilutionX/issues/3792.


AJenbo:
@glebm do you have any hesitations regarding adding the touch assets to SDL1 despite the code not utilizing them? At this point we are talking 411KB of png files. I could see that grow to about 1MB.

Source/miniwin/misc_msg.cpp
```diff
@@ -273,6 +273,35 @@ bool FalseAvail(const char *name, int value)
 	return true;
 }
 
+ControlTypes GetInputTypeFromEvent(SDL_Event &event)
```
AJenbo, line 276: Todo skip if gamepad is simulating mouse or similar shenanigans

Source/control.h
```diff
@@ -32,6 +33,15 @@ namespace devilution {
 #define SPANEL_WIDTH 320
 #define SPANEL_HEIGHT 352
 
+enum class ControlTypes : uint8_t {
+	None,
+	KeyboardAndMouse,
```
AJenbo, line 38: Should keyboard be considered separate from mouse? Maybe someone thinks that touch + keyboard is amazing?

CMake/Assets.cmake
```diff
@@ -142,16 +142,12 @@ set(devilutionx_assets
   ui_art/creditsw.pcx
   ui_art/hf_titlew.pcx
   ui_art/mainmenuw.pcx
-  ui_art/supportw.pcx)
-
-if(VIRTUAL_GAMEPAD)
```
glebm, line 147: `if(NOT USE_SDL1)`?

glebm: > do you have any hesitations regarding adding the touch assets to SDL1

Yeah, considering that we use SDL1 on some very resource-constrained platforms, let's not add unnecessary stuff to it, even though this is just the MPQ data, dwarfed by `diabdat.mpq` itself.

Source/CMakeLists.txt
```diff
@@ -82,6 +82,9 @@ set(libdevilutionx_SRCS
   controls/modifier_hints.cpp
   controls/plrctrls.cpp
   controls/keymapper.cpp
+  controls/touch/event_handlers.cpp
```
glebm, line 85: `if(NOT USE_SDL1)`?

Source/control.h
glebm: Devices like the new RG552 have a physical Gamepad + a touch screen

AJenbo: This is the same for Switch (and there are also some Android devices). 3DS and Vita also has a similar situation except the touch surface isn't directly mapped to the screen. There is nothing about this change that should cause issues for such platforms compared to previously. But we might want to do add individual enhancements to these platforms to take additional advantage of there unique interface, similarly to what is currently done for Vita (something I'm not currently taking in to account and why this is a draft).

Source/CMakeLists.txt
AJenbo: I went with a in-file guard instead since that is also what we do for `game_controller.cpp` and some other files.

Source/control.h
AJenbo: The vita unique touch handling is now taken in to account.

AJenbo:
@glebm are you able to spot the change that is causing Switch to produce a link error for this pr?

glebm:
Not anything obvious. Try disabling libzt and using non-system sdl image?

StephenCWills:
`IMG_LoadPNG_RW()` calls `IMG_Init()` which calls `IMG_InitWEBP()` which has a section that looks like this.

https://github.com/libsdl-org/SDL_image/blob/a31da6b890e857ee0ab089a13cc9a74601062283/IMG_webp.c#L67-L89
```c++
#if WEBP_DECODER_ABI_VERSION < 0x0100
        FUNCTION_LOADER(WebPGetFeaturesInternal, VP8StatusCode (*) (const uint8_t *data, uint32_t data_size, WebPBitstreamFeatures* const features, int decoder_abi_version))
        ...
#else
        FUNCTION_LOADER(WebPGetFeaturesInternal, VP8StatusCode (*) (const uint8_t *data, size_t data_size, WebPBitstreamFeatures* features, int decoder_abi_version))
        ...
#endif
```

Adding `#include "utils/png.h"` to any file would probably have caused that linker error. devkitPro has a libwebp package for Switch so we can either add `libwebp` as a dependency for the Switch build or else take @glebm's suggestion to use non-system sdl-image which excludes webp support by redefining `IMG_Init()`.

Source/controls/touch/gamepad.cpp
```diff
@@ -1,5 +1,7 @@
 #include <SDL.h>
 
+#ifndef USE_SDL1
```
StephenCWills, line 3: Why not just put this at the very top of the file?

StephenCWills:
@AJenbo FYI, I seem to have made a mistake when testing https://github.com/diasurgical/devilutionX/pull/3803#discussion_r772610640. `SetFullscreenMode()` does need to call `InitializeVirtualGamepad()`. You might consider slipping a fix into this PR. Otherwise, I can submit an issue and we can take care of it after this is merged.

Source/controls/touch/gamepad.cpp
AJenbo: Removed this and went with only conditionally including things instead

Source/controls/plrctrls.cpp
```diff
@@ -1249,16 +1250,104 @@ struct RightStickAccumulator {
...
+			if (IsHardwareCursor())
+				SetHardwareCursor(CursorInfo::UnknownCursor());
+		} else {
+			ResetCursor();
+		}
+#endif
+
+		CalculatePanelAreas();
+	}
+} // namespace
```
StephenCWills, line 1346:
```suggestion
}
```

AJenbo: Done :)

ephphatha: [llvm-namespace-comment](https://clang.llvm.org/extra/clang-tidy/checks/llvm-namespace-comment.html) is enabled in the clang-tidy config at the moment, this gets flagged (as a warning) by VS when it runs the clang-tidy checks

AJenbo: This was the end of a function where the namespace curly brace had ended up being reused.

StephenCWills (dismissed):
I just found one little typo in a comment. Otherwise, I'm liking what I'm seeing, and this works great on my touchscreen laptop as I switch between all the different input types.

AJenbo:
Nice! I have been testing it by plugging in a USB-C to A hub in my Phone and then attaching a USB mouse, keyboard and Gamepad in the hub. A bit clunky but It seamed to work there also :D

PR #3827, opened 2021-12-22
Load MPQ archives before reading options so translations can be detected

### LINKED ISSUE #3821: Language code from .INI ignored on start
**Important information**
Linux amd64, built 3e0a295dd36ddb16099324d4128c825a740c7ed2

**To Reproduce**
1. Start game
2. Go to Settings
3. Select a non-English language
4. See UI language changed
5. Exit game
6. Start game again

**Expected behavior**
UI is in selected language

**Actual behavior**
.INI has `Code=de` (or whatever) but the game starts in English, Settings shows English selected

ephphatha:
This is a result of checking that the translation specified in diablo.ini actually exists:
https://github.com/diasurgical/devilutionX/blob/cacf55bb34c1c314d2b39868826f63f732cbfdf6/Source/options.cpp#L932-L937

I found the check useful when I was switching between branches with different translations available, but another fix for # 3821 could be to always assume the ini value is correct and bail early if the value exists (remove the HasTranslation call in that block).

For now I think it's safe to load the MPQs before the ini, there's no circular dependency (yet 😂 )

fixes #3821

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -935,9 +935,6 @@ void DiabloInit()
 	init_create_window();
 	was_window_init = true;
 
-	init_archives();
-	was_archives_init = true;
-
 	if (forceSpawn || *sgOptions.StartUp.shareware)
 		gbIsSpawn = true;
 	if (forceDiablo || *sgOptions.StartUp.gameMode == StartUpGameMode::Diablo)
@@ -1618,6 +1615,10 @@ int DiabloMain(int argc, char **argv)
 
 	DiabloParseFlags(argc, argv);
 	InitKeymapActions();
+
+	init_archives();
+	was_archives_init = true;
+
 	LoadOptions();
 
 	DiabloInit();


ephphatha:
Loading the MPQs first allows the automatic language detection to actually work on first run for a release build (with packaged devilutionx.mpq and no assets directory), guess this reordering is necessary anyway.

licaon-kter:
> actually work on first run for a release build (with packaged devilutionx.mpq and no assets directory),

Was seeing this yes, but thought it normal, thanks for fixing

glebm:
Should the `LanguageInitialize` call in that function be moved down as well?
Or perhaps we should load just the `devilutionx.mpq` archive early?

glebm:
Ah, nevermind, `DiabloParseFlags` has to run before `init_archives` in case the data directory is specified

glebm (approved):

AJenbo:
We should probably drop translations from the CLI help screen since we can't load translations that early. We wouldn't be the first app to work like that.

PR #3830, opened 2021-12-22
Draft: adding birds

matheusgomes28:
Attempting to add birds, failing miserably so far ...


matheusgomes28:
I know this is failing the pipeline and crashes in town... just adding this here for visibility in a thread.

Source/engine/assets.cpp
```diff
@@ -31,6 +31,11 @@ bool OpenMpqFile(const char *filename, MpqArchive **archive, uint32_t *fileNumbe
 
 } // namespace
 
+SDL_RWops* OpenAssetFromDisk(const char* filename)
```
AJenbo, line 34: Fyi the game will already look on disk before looking in the MPQs

It looks relatively to the mpq files, and also in ./assets next to the exe.

Source/towners.cpp
```diff
@@ -76,7 +77,7 @@ void InitSmith(Towner &towner, const TownerData &townerData)
 	towner.animOrder = AnimOrder;
 	towner.animOrderSize = sizeof(AnimOrder);
 	LoadTownerAnimations(towner, "Towners\\Smith\\SmithN.CEL", 16, 3);
-	towner.name = _("Griswold the Blacksmith");
+	towner.name = _("Griswold the Penis");
```
AJenbo, line 80:
```suggestion
	towner.name = _("Griswold the Blacksmith");
```

matheusgomes28: This wasn't meant to make it here! Big mb

Source/towners.cpp
```diff
@@ -233,6 +234,21 @@ void InitCows(Towner &towner, const TownerData &townerData)
...
+{
+	towner._tAnimWidth = 1;
+	static const uint8_t AnimOrder[] = {
+		// clang-format off
+		1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 
+		// clang-format on
+	};
+	towner.animOrder = AnimOrder;
+	towner.animOrderSize = 10;
+	towner.data = LoadFileInMemFromDisk(R"(D:\development\tmp\pngs2cel\new_bird.cel)");
```
glebm, line 247: Add it to `Packaging/resources/assets/data` in the repo and to `CMake/Assets.cmake`, and then use `LoadFileInMem`

Source/engine/assets.cpp
matheusgomes28: Thanks, I did try to load from disk before with an absolute path, but it didn't work too well on Windows. As you and @glebm mentioned, I'm going to add it as an "asset" and add to CMake as well.

Source/towners.cpp
matheusgomes28: Cheers, sounds like a much more reasonable solution.

Source/engine/assets.cpp
matheusgomes28: I added the file as an asset now, should illustrate the crash a bit better! It eventually crashes in the load little endian functions, probably as a result of me setting up the animation completely incorrect... I'll be doing some more digging into how the cel/cl2 files are used and hopefully get something working haha

Source/towners.cpp
```diff
@@ -824,6 +842,7 @@ void InitTowners()
 	assert(CowCels == nullptr);
 
 	CowCels = LoadFileInMem("Towners\\Animals\\Cow.CEL");
+	BirdCels = LoadFileInMem(R"(data\bird.cl2)");
```
qndel, line 845: what's this R lol

AJenbo: A Raw string where no escape sequences are active, because windows uses backslashes for the folder devider this avoid the constant need for escaping it.
`"windows\\file\\systems\\are\\silly.txt"`
vs
`R"windows\file\systems\are\silly.txt"`

heinermann: Forward slashes are valid in the windows filesystem but MPQs will use backslashes in the name which directly get translated to a hash.

PR #3841, opened 2021-12-25
Update Italian localization

M0Rf30:
It also includes some little fixes here and there.
Happy Xmas :tada:

### DIFF
--- a/Translations/it.po
+++ b/Translations/it.po
@@ -1,8 +1,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: DevilutionX\n"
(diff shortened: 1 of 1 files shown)


AJenbo (approved):

AJenbo:
happy holidays :)

yuripourre:
@M0Rf30 hey, I just noticed that if you change **accuracy** to _precisione_ instead of _accuratezza_, the translation will be the same for both **accuracy** and **precision** suffixes.

M0Rf30:
@yuripourre I'll review it in the next iteration. thanks for your catch

PR #3843, opened 2021-12-26
Fix Event Messages showing white instead of gold

kphoenix137:


AJenbo:
This is still your old branch from before the change was merged. You should update your master to be in sync with upstream and then make a new branch from there (or make one directly from upstream/master).

AJenbo:
Merged via https://github.com/diasurgical/devilutionX/commit/1c25cb4bf05c2a1309886ed81d7fc3341ff50450

PR #3853, opened 2021-12-28
Stash

kphoenix137:
image

image

Thease issues will be addressed in later PRs:

- [ ] If a player doesn't have the requirements of an item it is placed on the stash tinted as red, if another player opens the stash it will be still red. The opposite is also true, items show colored even if the player doesn't have the requirements
- [ ] It's not possible to use or cast spells on items in the stash
- [ ] Gamepad is untested and probably very broken
- [ ] It's possible to open both gold input panels at the same time


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
AJenbo, line 1: add .vscode to .gitignore and delete this file.

Source/stash.h
```diff
@@ -62,9 +62,9 @@ enum stash_xy_slot : uint8_t {
 struct StashStruct {
 	void RemoveStashItem(int iv, bool calcScrolls);
 
-	Item StashList[100];
-	int8_t StashGrid[NUM_STASH_GRID_ELEM];
-	int _pNumStash;
+	std::vector<Item> StashList;
+	std::vector<int8_t> StashGrid;
+	int NumStash;
```
AJenbo, line 67: Only globals start with upper case:
```suggestion
	std::vector<Item> stashList;
	std::vector<int8_t> stashGrid;
	int numStash;
```

Source/pfile.cpp
```diff
@@ -64,6 +64,30 @@ std::string GetSavePath(uint32_t saveNum)
...
+		}
+	} else {
+		if (!gbIsMultiplayer) {
+			path.append("single_stash");
+		} else {
+			path.append("multi_stash");
+		}
+	}
+	path.append(ext);
+	return path;
```
AJenbo, line 88:
```suggestion
    `return "stash.sv";`
```

Create multiple files insides of the archive instead.

kphoenix137: I have to disagree with this idea; For the sake of moving files around, this might not be ideal for some players. Perhaps they would want to delete their Hellfire Stash for some reason, but maintain their Diablo stash. I think this could be comparable to packing all Diablo, Shareware, Hellfire SP and MP files into one big file.

Source/loadsave.cpp
```diff
@@ -1776,6 +1783,44 @@ void LoadHeroItems(Player &player)
...
+{
+	const char *filename;
+	if (!gbIsMultiplayer)
+		filename = "spstashitems";
+	else
+		filename = "mpstashitems";
+
+	Stash.gold = 0;
+	Stash.stashGrids.clear();
+	Stash.stashList.clear();
```
AJenbo, line 1798: @StephenCWills ?
```suggestion
	Stash = {};
```

Source/stash.cpp
```diff
@@ -0,0 +1,839 @@
...
+					itemId = iv + 1;
+				}
+				if (itemId == -(stashList.size())) {
+					itemId = -(iv + 1);
+				}
+			}
+		}
+	}
+	stashList.pop_back();
+}
```
ephphatha, line 837:
```suggestion
	if (stashList.empty()) {
		return;
	}
	
	// If the item at the end of stash array isn't the one we removed, we need to swap its position in the array with the removed item
	std::size_t lastItemIndex = stashList.size() - 1;
	if (lastItemIndex != iv) {
		stashList[iv] = stashList[lastItemIndex];

		for (auto &pair : Stash.stashGrids) {
			auto &grid = pair.second;
			for (auto &itemId : grid) {
				if (itemId == lastItemIndex + 1) {
					itemId = iv + 1;
				}
				if (itemId == -(lastItemIndex + 1)) {
					itemId = -(iv + 1);
				}
			}
		}
	}
	stashList.pop_back();
}
```

Calling pop_back on an empty vector is undefined behaviour. Also might as well capture the intent of the vector::size() call in a named variable.

Source/stash.h
```diff
@@ -0,0 +1,57 @@
...
+	// clang-format off
+	STASH_SLOT_FIRST = 0,
+	STASH_SLOT_LAST  = 99,
+	STASH_SLOT_NUM = 100
+	// clang-format on
+};
+
+struct StashStruct {
+	void RemoveStashItem(int8_t iv);
+	std::map<int, std::array<int8_t, STASH_SLOT_NUM>> stashGrids;
```
ephphatha, line 33:
```suggestion
	std::map<int, std::array<std::ptrdiff_t, STASH_SLOT_NUM>> stashGrids;
```
or std::size_t if the logic is refactored to not rely on negative values would allow more items to be stored.

Source/stash.cpp
```diff
@@ -0,0 +1,839 @@
...
+	    && item._iStatFlag
+	    && AllItemsList[item.IDidx].iUsable;
+}
+
+int SwapStashItem(Item *a, Item *b)
+{
+	std::swap(*a, *b);
+
+	return b->_iCurs + CURSOR_FIRSTITEM;
+}
```
ephphatha, line 317: Any reason these need to be reimplemented instead of using the inv.cpp definitions?

ephphatha (commented):
I know this is still a draft, but had a read through the implementation and there were a few things which stood out as potential improvements (some minor, some not 😓 )

Source/stash.cpp
```diff
@@ -0,0 +1,839 @@
...
+	iv++;
+
+	// Iterate through stashGrid and remove every reference to item
+	for (int8_t &itemId : Stash.stashGrids[StashPage]) {
+		if (itemId == iv || itemId == -iv) {
+			itemId = 0;
+		}
+	}
+
+	iv--;
```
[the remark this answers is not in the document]
AJenbo: @ephphatha this is actually still like this in the inventory logic :)

Source/stash.cpp
```diff
@@ -0,0 +1,839 @@
...
+		    && MousePosition.y >= StashRect[r].y + yo - (InventorySlotSizeInPixels.height + 1)
+		    && MousePosition.y < StashRect[r].y + yo) {
+			break;
+		}
+	}
+
+	if (r >= STASH_SLOT_NUM)
+		return -1;
+
+	int8_t rv = -1;
```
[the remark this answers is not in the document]
AJenbo: Won't that cause an issue as long as `-1` is used to indicate a special case?

Source/stash.cpp
```diff
@@ -0,0 +1,839 @@
...
+}
+
+void DrawStash(const Surface &out)
+{
+	Point stashButton;
+	CelDrawTo(out, GetPanelPosition(UiPanels::Stash, { 0, 351 }), *StashPanelArt, 1);
+	for (int penis = 0; penis < 5; penis++) {
+		stashButton = StashButtonCelRect[penis];
+		if (StashButtons[penis]) {
+			switch (penis) {
```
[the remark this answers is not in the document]
AJenbo: Not much to show here either with `penis < 5`...

Source/stash.cpp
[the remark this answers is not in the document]
ephphatha: using size_t(-1) as a return value for a function returning an index is well defined (used in a lot of the string functions via `npos` for example) - https://en.cppreference.com/w/cpp/string/basic_string/npos

Conversion rules mean this is the same as declaring the max value for the given type. Callers can either check the returned index is != -1 or that the returned index is < the container size.

AJenbo:
@ephphatha I addressed a chunk of the issues you pointed out and did some additional cleanups. But I am way to tired atm to be sure I did it right, or attempt changing the array in to a 2D 10x10, or get rid of the regative index :cold_sweat:  If you could give the changes a thorough look that would be appreciated :)

ephphatha (commented):
Haven't had a chance to sit down and play it properly but it looks good apart from the broken additional missiles loading. Few more comments and a potential bug, didn't go too deep as I'm aware it's still being worked on :)

Source/controls/plrctrls.cpp
```diff
@@ -717,6 +739,25 @@ int FindClosestInventorySlot(Point mousePos)
...
+
+	for (int y = 0; y < 10; y++) {
+		for (int x = 0; x < 10; x++) {
+			int distance = mousePos.ManhattanDistance(GetStashSlotCoord({ x, y }));
+			if (distance < shortestDistance) {
+				shortestDistance = distance;
+				bestSlot = { x, y };
+			}
+		}
+	}
```
ephphatha, line 756: Loops like this which use `x,y` as a point can use `PointsInRectangleRange`
```suggestion
	for (auto point : PointsInRectangleRange{{0, 0}, {10, 10}}) {
		int distance = mousePos.ManhattanDistance(GetStashSlotCoord(point));
		if (distance < shortestDistance) {
			shortestDistance = distance;
			bestSlot = point;
		}
	}
```

could even potentially use `std::min_element` but given the derived attribute it doesn't make the logic clearer.

Source/stash.cpp
```diff
@@ -0,0 +1,618 @@
...
+Point FindSlotUnderCursor(Point cursorPosition)
+{
+	for (int y = 0; y < 10; y++) {
+		for (int x = 0; x < 10; x++) {
+			Rectangle cell {
+				GetPanelPosition(UiPanels::Stash, StashCellPosition({ x, y })),
+				InventorySlotSizeInPixels
+			};
+
+			if (cell.Contains({ cursorPosition })) {
```
ephphatha, line 130:
```suggestion
			if (cell.Contains(cursorPosition)) {
```

Source/stash.cpp
```diff
@@ -0,0 +1,618 @@
...
+	holdItem._itype = ItemType::None;
+
+	bool automaticallyMoved = false;
+	bool automaticallyEquipped = false;
+
+	int16_t ii = Stash.stashGrids[Stash.page][slot.x][slot.y];
+	if (ii != 0) {
+		int iv = abs(ii) - 1;
+
+		holdItem = Stash.stashList[iv - 1];
```
ephphatha, line 244:
```suggestion
		holdItem = Stash.stashList[iv];
```

already subtracted 1 on line 242

Source/stash.cpp
```diff
@@ -0,0 +1,618 @@
...
+		if (automaticMove) {
+			if (CanBePlacedOnBelt(holdItem)) {
+				automaticallyMoved = AutoPlaceItemInBelt(player, holdItem, true);
+			} else {
+				automaticallyMoved = automaticallyEquipped = AutoEquip(MyPlayerId, holdItem);
+			}
+		}
+
+		if (!automaticMove || automaticallyMoved) {
+			Stash.RemoveStashItem(iv - 1);
```
ephphatha, line 254:
```suggestion
			Stash.RemoveStashItem(iv);
```
same here

Source/stash.cpp
```diff
@@ -0,0 +1,618 @@
...
+	SDL_StartTextInput();
+}
+
+} // namespace
+
+constexpr Point StashCellPosition(Point slot)
+{
+	constexpr int StashNextCell = INV_SLOT_SIZE_PX + 1; // spacing between each cell
+
+	return slot * StashNextCell + Displacement { 17, 76 };
```
ephphatha, line 313:
```suggestion
	return slot * StashNextCell + Displacement { 17, 47 };
```
This displacement causes the top row of cells to be inaccessible, 47 is `76 - StashNextCell` but might not be correct either.

Source/stash.cpp
```diff
@@ -0,0 +1,618 @@
...
+		CheckStashCut(MousePosition, isShiftHeld, isCtrlHeld);
+	}
+	if (gbIsMultiplayer) {
+		sfile_write_stash();
+	}
+}
+
+int16_t CheckStashHLight(Point mousePosition)
+{
+	Point slot = { -1, -1 };
```
ephphatha, line 474: `-1,-1` is used fairly frequently as a magic value, would be worth defining as a constant/constexpr value somewhere.

ephphatha (commented):
Haven't had time to give it a proper review, only one real issue when playtesting though

AJenbo:
@ephphatha much apriciated :)

Source/control.cpp
```diff
@@ -1069,6 +1116,28 @@ void control_drop_gold(char vkey)
...
+
+	if (myPlayer._pHitPoints >> 6 <= 0) {
+		CloseGoldWithdraw();
+		withdrawGoldValue = 0;
+		return;
+	}
+
+	if (vkey == DVL_VK_RETURN) {
+		if (withdrawGoldValue > 0)
+			WithdrawGold(myPlayer);
```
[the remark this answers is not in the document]
AJenbo: Nicely figured out :)

Source/loadsave.cpp
```diff
@@ -1557,7 +1564,7 @@ void SaveAdditionalMissiles()
 
 void LoadAdditionalMissiles()
 {
-	LoadHelper file("additionalMissiles");
+	LoadHelper file(OpenStashArchive(), "additionalMissiles");
```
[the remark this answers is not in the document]
AJenbo: Ha ha, yeah that was an ovious typo :D

Source/stash.cpp
AJenbo: Thanks that saved me some debugging :)

Source/controls/plrctrls.cpp
AJenbo: There are probably lots of opotunities like this all around the codebase :thinking:

Source/stash.cpp
AJenbo: 76 was assuming the position relevant for the draw function rather then the logical one.

Source/stash.cpp
```diff
@@ -0,0 +1,839 @@
...
+		if (Stash.stashGrids[StashPage][i] != 0) {
+			StashDrawSlotBack(
+			    out,
+			    GetPanelPosition(UiPanels::Stash, StashRect[i + STASH_SLOT_FIRST]) + Displacement { 0, -1 },
+			    InventorySlotSizeInPixels);
+		}
+	}
+
+	for (int i = 0; i < STASH_SLOT_NUM; i++) {
+		if (Stash.stashGrids[StashPage][i] > 0) { // first slot of an item
```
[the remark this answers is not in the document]
AJenbo: Nicely spotted, there where a few things that had to slide in to place for me before I could see it :)

Source/stash.h
AJenbo: I have gone for `uint16_t` to be sure that it matches with the save format.

Source/controls/plrctrls.cpp
```diff
@@ -670,6 +685,17 @@ int FindFirstSlotOnItem(int8_t itemInvId)
 	return -1;
 }
 
+int FindFirstStashSlotOnItem(int8_t itemInvId)
```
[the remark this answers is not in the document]
AJenbo: Updated to `uint16_t` to match the type it's being compared to.

Source/stash.cpp
AJenbo: I added `constexpr Point InvalidStashPoint { -1, -1 };` in stash.h

Source/stash.cpp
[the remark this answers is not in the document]
AJenbo: Thanks for explaining.

Source/loadsave.cpp
```diff
@@ -93,9 +95,12 @@ class LoadHelper {
 	}
 
 public:
-	LoadHelper(const char *szFileName)
+	LoadHelper(std::optional<MpqArchive> archive, const char *szFileName)
```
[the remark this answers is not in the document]
AJenbo: I'm not super happy about it, but I feel resolving it would require a lot of general refactoring to how save games are handled. So going to leave it like this for now.

Source/controls/plrctrls.cpp
```diff
@@ -1530,6 +1562,40 @@ void PerformPrimaryAction()
 				Slot = jumpSlot;
 				SetCursorPos(mousePos);
 			}
+		} else if (IsStashOpen && GetLeftPanel().Contains(MousePosition)) {
+			Point stashSlot = (ActiveStashSlot != InvalidStashPoint) ? ActiveStashSlot : FindClosestStashSlot(MousePosition);
+
+			// Find any item occupying a slot that is currently under the cursor
+			uint16_t itemUnderCursor = [](Point stashSlot) -> uint16_t {
```
glebm, line 1569: nit: Probably makes more sense as a `uint16_t FindStashItemUnderCursor(Point stashSlot)` function rather than a self-evaluating lambda.

Source/qol/stash.cpp
```diff
@@ -0,0 +1,609 @@
...
+	for (auto slot : PointsInRectangleRange({ { 0, 0 }, { 10, 10 } })) {
+		if (Stash.stashGrids[Stash.page][slot.x][slot.y] == 0) {
+			continue; // Not item in the given slot
+		}
+
+		InvDrawSlotBack(out, GetStashSlotCoord(slot) + offset, InventorySlotSizeInPixels);
+
+		uint16_t itemId = Stash.stashGrids[Stash.page][slot.x][slot.y] - 1;
+		if (Stash.stashList[itemId].position == slot) {
+			itemIds.push_back(itemId);
```
glebm, line 356: This `itemIds` vector is only used once below. Rather than creating a vector on every frame, simply move the body of the loop below in here.

AJenbo: The background must first be drawn before the items are drawn on top. The vector is to save us from iterating over all slots twice and instead remember which ones contains items from the first loop.

glebm: Ah ok, though I bet scanning twice is faster than allocating.

AJenbo: Alright, I can revert it :)

Source/controls/plrctrls.cpp
AJenbo: Being an almost exact copy of the block above it I think I will save this for a later refactor to minimize diff in this PR.

Source/control.cpp
```diff
@@ -383,8 +378,32 @@ void RemoveGold(Player &player, int goldIndex)
...
+	}
+	if (ControlMode == ControlTypes::VirtualGamepad) {
+		return false;
+	}
+	if (stextflag != STORE_NONE || IsStashOpen) {
+		return false;
+	}
+	if (QuestLogIsOpen && GetLeftPanel().Contains(GetMainPanel().position + Displacement { 0, -74 })) {
+		return false;
+	}
```
StephenCWills, line 394: In this case, does it make more sense to treat stash the same as store text or the quest log?

Source/qol/stash.cpp
```diff
@@ -0,0 +1,610 @@
...
+		PrintItemDetails(item);
+	} else {
+		strcpy(infostr, item._iName);
+		PrintItemDur(item);
+	}
+
+	return ii;
+}
+
+bool UseStashItem(uint16_t c)
```
StephenCWills, line 437: This is only used by `CtrlUseStashItem()`, which is itself unused.

Source/controls/plrctrls.cpp
```diff
@@ -1496,7 +1528,7 @@ void PerformPrimaryAction()
 		if (pcurs > CURSOR_HAND && pcurs < CURSOR_FIRSTITEM) {
 			TryIconCurs();
 			NewCursor(CURSOR_HAND);
-		} else {
+		} else if (pcursinvitem != -1) {
```
StephenCWills, line 1531: It seems this change makes it impossible to place items in the inventory using gamepad.

AJenbo: Oops. Fixed :)

Source/qol/stash.cpp
AJenbo: Yeah, it's going to be needed once we implement using items directly from the stash.

Source/controls/plrctrls.cpp
```diff
@@ -1687,6 +1755,24 @@ void CtrlUseInvItem()
 	}
 }
 
+void CtrlUseStashItem()
```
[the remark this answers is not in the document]
AJenbo: It's intended for later use once gamepad controls are fully implemented. I could leave it out but feel it would be just extra effort to do so.

Source/control.cpp
AJenbo: Since it is accessed via the store and you can't move with it open I went with treating it more as a store. I would also like to use it as a basis for an eventual graphical store.

StephenCWills (approved):

obligaron:
Note for the following issue:

>  It's not possible to use or cast spells on items in the stash

I think we should consider not to add the feature to **cast scrolls** from the stash (at least for the 1.4 release).
If we add this feature this would increase the complexity to fix #793 a fair bit and would invalidate some of the work done in #4074.

AJenbo:
Mostly it's regarding casting identify and repair on items that are in the stash. Currently you simply loos the identify scroll.

obligaron:
Oh sorry I read the issue wrong. 😞
I have no problem with applying spells (identify or repair) on stash items.

My thoughts are about using scrolls (`UseStashItem`) that are placed in the stash (right click on a scroll in stash).
Cause in #4074 we don't remove a used scroll immediately. That means we remove the scroll when the cast begins (not when clicking the item). If we support stash item scrolls to be used directly, we would need to remember that we casted the spell from the stash and remove the scroll from the stash instead of the inventory.
To fix double cast, #4074 also validates when casting if the scroll still existing. This check (`UseScroll`) would also need to check the stash. 😕

PR #3856, opened 2021-12-29
Use Object references when initialising objects from template and second init for crypts/doors

ephphatha:
Doors really threw me in this one. They involve setting a lot of other state and it's not really clear what it does or whether it's necessary. I ended up combining catacombs/caves as they are identical apart from the piece number used and whether the workaround for what I assume is rendering issues caused by overlapping tiles is needed on the level type.

Does this approach look reasonable? I can drop the last commit and leave the init functions duplicating a bit of code if it makes it clearer to read.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1123,38 +1123,38 @@ void ObjSetMicro(Point position, int pn)
 	}
 }
 
-void AddL1Door(Object &door)
+void InitializeL1Door(Object &door)
 {
-	door._oDoorFlag = true;
+	door.InitializeDoor();
+	door._oVar1 = dPiece[door.position.x][door.position.y];
 	if (door._otype == _object_id::OBJ_L1LDOOR) {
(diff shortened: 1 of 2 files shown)


Source/objects.cpp
```diff
@@ -4717,18 +4720,19 @@ void AddObject(_object_id objType, Point objPos)
...
-		AddL1Door(oi, objPos, objType);
+		InitializeL1Door(object);
 		break;
 	case OBJ_L2LDOOR:
 	case OBJ_L2RDOOR:
-		AddL2Door(oi, objPos, objType);
-		break;
+		// If a catacombs door happens to overlap an arch then clear the arch tile to prevent weird rendering
+		dSpecial[object.position.x][object.position.y] = 0;
+		// intentional fall-through
```
AJenbo, line 4729: It's kinda funny that this feature of switches is so commonly a mistake that it is customary to write a comment when it's not 😅

AJenbo (approved):

AJenbo (approved):

PR #3859, opened 2021-12-29
Replace en_US code with en when searching for user preferred languages

### LINKED ISSUE #3858: Users with US English as a preferred language see "en_US" (as well as "English") in the settings menu
**Important information**
Windows 11 / DevX since https://github.com/diasurgical/devilutionX/commit/a1652bd5d6ea087f02c691dc7b27d1670ec9cd0a / compiled

**Describe**
Returning true for "en_US" in `HasTranslation` doesn't match the expected codes for English as presented in the options menu. Users will see their language set to "en_US" instead of "English" which may be confusing.

**To Reproduce**
1. Launch the game with diablo.ini configured with [Language] Code=en_US
2. Navigate to the ingame language settings menu
3. See the language is displayed as "en_US"
4. Change the language, see that there are options for "English" and "en_US"

**Expected behavior**
en_US should not be an option, only "English" should be available

**Screenshots**
2021-12-30T100445_devilutionx

**Additional context**
This is only an issue for US english, other languages with translations should appear using their posix code if they haven't been added to the list of options yet.

ephphatha:
This allows the settings menu to display the expected value

fixes #3858

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -1030,7 +1030,27 @@ void OptionEntryLanguageCode::LoadFromIni(string_view category)
 	locales.emplace_back(std::locale("").name().substr(0, 5));
 #endif
 
-	// Insert non-regional locale codes after the last regional variation so we fallback to neutral translations if no regional translation exists that meets user preferences.
+	// So that the correct language is shown in the settings menu for users with US english set as a preferred language
+	//  we need to replace the "en_US" locale code with the neutral string "en" as expected by the available options
+	auto usEnglishLocale = std::find(locales.begin(), locales.end(), "en_US");
+	if (usEnglishLocale != locales.end()) {
+		auto neutralEnglishLocale = std::find(locales.begin(), locales.end(), "en");
+		if (usEnglishLocale < neutralEnglishLocale) {
+			// Neutral locale either isn't user-preferred or is after the us english variation in preferences, so we
+			//  can update this entry
+			*usEnglishLocale = "en";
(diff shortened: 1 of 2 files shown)


AJenbo (approved):
Seams a bit overkill, would it not be simpler to have a list of aliases?

ephphatha:
Maybe? I think this is the only case where we'd ever have two different codes correspond to the same language entry. Most of the complexity comes from removing the duplicate entry, this could be implemented as the following instead:

```cpp
std::replace(locales.begin(), locales.end(), "en_US", "en");
```

...I think that's a better implementation anyway since it doesn't matter if there are duplicate entries in the locales list after the first occurrence of "en", the search code will bail out early.

Source/options.cpp
```diff
@@ -1030,7 +1030,27 @@ void OptionEntryLanguageCode::LoadFromIni(string_view category)
...
+			*usEnglishLocale = "en";
+			if (neutralEnglishLocale != locales.end()) {
+				// And remove the now duplicate neutral variant (if it exists)
+				locales.erase(neutralEnglishLocale);
+			}
+		} else {
+			// somehow there's already a neutral locale earlier than the regional variation, remove the en_US entry
+			locales.erase(usEnglishLocale);
+		}
+	}
```
ephphatha, line 1050:
```suggestion
	std::replace(locales.begin(), locales.end(), std::string { "en_US" }, std::string { "en" });
```

AJenbo (approved):

PR #3862, opened 2021-12-30
Move platform specific locale lookup code to platform directory

ephphatha:
As AJenbo mentioned in https://github.com/diasurgical/devilutionX/pull/3663#pullrequestreview-831049564 options.cpp isn't really the best place for platform specific language lookup code to live. Given `n3ds::GetLocale()` already existed in platform/ctr/locale.cpp I figure creating a matching platform/locale.cpp as a generic wrapper is suitable.

I was also playing with an approach where each platform provides a definition for `GetLocales()`, however this requires having a "default" platform library or a way to have multiple definitions for a function and some method to pick the specialisation at compile time... seemed overly complex compared to just using the `#ifdef` chain.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -141,7 +141,8 @@ set(libdevilutionx_SRCS
   dvlnet/packet.cpp
   storm/storm_net.cpp
   storm/storm_svid.cpp
-  miniwin/misc_msg.cpp)
+  miniwin/misc_msg.cpp
+  platform/locale.cpp)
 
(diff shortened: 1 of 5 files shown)


AJenbo (dismissed):
Thanks, i started messing with it but never got fare. I might have stuck it in utilities, but this works too :)

ephphatha:
Finding all sorts of derived includes 😂 . Vita build should work now 🤞

AJenbo (approved):

PR #3863, opened 2021-12-30
Fix Elemental Arrow Behavior

kphoenix137:
Fixes collision detection for the fire/lightning sprite that causes element base damage on collision with players and monsters, and removed Damage Mod on items from changing elemental damage.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -263,7 +263,7 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 
 	const auto &player = Players[pnum];
 
-	if (MissilesData[t].mType == 0) {
+	if (MissilesData[t].mType == 0 && mir == MISR_NONE) {
 		dam = player._pIBonusDamMod + dam * player._pIBonusDam / 100 + dam;
 		if (player._pClass == HeroClass::Rogue)
 			dam += player._pDamageMod;
@@ -384,7 +384,7 @@ bool Plr2PlrMHit(int pnum, int p, int mindam, int maxdam, int dist, missile_id m
 		dam = target._pHitPoints / 3;
 	} else {
 		dam = mindam + GenerateRnd(maxdam - mindam + 1);
-		if (MissilesData[mtype].mType == 0)
+		if (MissilesData[mtype].mType == 0 && MissilesData[mtype].mResist == MISR_NONE)
 			dam += player._pIBonusDamMod + player._pDamageMod + dam * player._pIBonusDam / 100;
 		if (!shift)
 			dam <<= 6;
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -262,7 +262,7 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 
 	const auto &player = Players[pnum];
 
-	if (MissilesData[t].mType == 0) {
+	if (MissilesData[t].mType == 0 && mir == 0) {
```
StephenCWills, line 265:
```suggestion
	if (MissilesData[t].mType == 0 && mir == MISR_NONE) {
```

Source/missiles.cpp
```diff
@@ -383,7 +383,7 @@ bool Plr2PlrMHit(int pnum, int p, int mindam, int maxdam, int dist, missile_id m
 		dam = target._pHitPoints / 3;
 	} else {
 		dam = mindam + GenerateRnd(maxdam - mindam + 1);
-		if (MissilesData[mtype].mType == 0)
+		if (MissilesData[mtype].mType == 0 && resper == 0)
```
StephenCWills, line 386: Do we want to apply bonus damage when the player has no resistance to the element? Maybe we can use `mir` again?
```suggestion
		if (MissilesData[mtype].mType == 0 && MissilesData[mtype].mResist == MISR_NONE)
```

Source/missiles.cpp
```diff
@@ -2862,7 +2862,7 @@ void MI_LArrow(Missile &missile)
 
 		missile_resistance rst = MissilesData[missile._mitype].mResist;
 		MissilesData[missile._mitype].mResist = MISR_NONE;
-		MoveMissileAndCheckMissileCol(missile, mind, maxd, true, true);
+		MoveMissileAndCheckMissileCol(missile, mind, maxd, true, false);
```
StephenCWills, line 2865: The only thing that worries me at this point is how this change might exacerbate #1672.

StephenCWills (approved):

Source/missiles.cpp
kphoenix137: It will make it worse. If the logic is consistent between the two, that means trap elemental arrows haven't been dealing elemental damage most of the time, and now they will.

Source/missiles.cpp
```diff
@@ -30,6 +30,7 @@ int AvailableMissiles[MAXMISSILES];
 Missile Missiles[MAXMISSILES];
 int ActiveMissileCount;
 bool MissilePreFlag;
+bool collisionFlag;
```
StephenCWills, line 33: Don't use a global variable to hold state for a single missile. If multiple elemental arrows are active, they'll be vying for control over the state of the global variable.

It should be much simpler if we apply the elemental damage to the same target at the point where they take the physical damage, in which case we wouldn't need to use any additional state or spread damage out across multiple frames.

StephenCWills (changes requested):

Source/missiles.cpp
kphoenix137: The one downside to this though is that these arrows' elemental damage wouldn't be able to bypass a block animation.  From a PvE perspective, this is inconsequential and likely beneficial, however this is a very helpful PvP feature

StephenCWills: That's not true, actually. There are two ways to bypass the block animation. The first is to use the `shift` flag to bypass the block check, like how Firewall or Acid Puddles do it. The second way is to apply the damage after the player has already started blocking.

In Diablo, if the player blocks the arrow, they won't be able to block the elemental damage regardless of whether it occurs in the same frame. If the player gets hit by the arrow, they will have a chance to block the elemental damage regardless of whether it occurs in the same frame. In either case, the frame in which the damage occurs doesn't matter.

In Hellfire, blocked projectiles don't get deleted so the arrow would pass right through and the elemental damage would never be applied. If the player gets hit by the projectile, they'd have a chance of blocking the elemental damage the same as in Diablo. Once again, the frame in which the damage occurs doesn't matter.

kphoenix137: Okay, sounds like the way to achieve the desired effect then is to apply elemental damage immediately after the arrow is blocked to allow the elemental damage to still hit?

kphoenix137: > Don't use a global variable to hold state for a single missile. If multiple elemental arrows are active, they'll be vying for control over the state of the global variable.
> 
> It should be much simpler if we apply the elemental damage to the same target at the point where they take the physical damage, in which case we wouldn't need to use any additional state or spread damage out across multiple frames.

Removed bool

kphoenix137: I think the collision needs to be rewritten before we can check for collision in the same tick twice, since the second animation isn't moved to the target tile until the next tick.

StephenCWills (dismissed):

Source/missiles.cpp
```diff
@@ -2810,69 +2810,66 @@ int AddMissile(Point src, Point dst, Direction midir, missile_id mitype, mienemy
...
 {
-	missile._mirange--;
+	int mind;
+	int maxd;
+	int eMind;
+	int eMaxd;
 	int p = missile._misource;
+	missile_graphic_id eAnim;
+	missile_resistance eRst;
+	auto &player = Players[p];
```
StephenCWills, line 2820: Try to see if you can move these declarations closer to where they're used. In particular, `Players[p]` might be accessing OOB memory because `p` can be `-1`.

Source/missiles.cpp
```diff
@@ -2810,69 +2810,66 @@ int AddMissile(Point src, Point dst, Direction midir, missile_id mitype, mienemy
...
+				eMaxd = player._pIFMaxDam;
+				eAnim = MFILE_MAGBLOS;
+				eRst = MISR_FIRE;
+				break;
+			}
+			SetMissAnim(missile, eAnim);
+			if (p == -1) {
+				eMind = GenerateRnd(10) + 1 + currlevel;
+				eMaxd = GenerateRnd(10) + 1 + currlevel * 2;
+			}
```
StephenCWills, line 2869: The same point can be made here. This was a good idea to try and clean things up, but if `p == -1`, you just initialized `eMind` and `eMaxd` using an invalid player.

StephenCWills (changes requested):

Source/missiles.cpp
kphoenix137: Ah yes! I'll fix that tonight

StephenCWills (dismissed):

StephenCWills (approved):

PR #3864, opened 2021-12-30
Implement 3XDAMVDEM Logic for Arrows

kphoenix137:
The Deadly Hunter bow contains the 3XDAMVDEM affix power which isn't present in missiles.cpp, preventing the affix from doing anything altogether. This fix adds the logic from player.cpp in a consistent way with how damage gets handled in melee combat.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -275,10 +275,8 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 	if (resist)
 		dam >>= 2;
 
-	if (monster.MData->mMonstClass == MonsterClass::Demon) {
-		if ((player._pIFlags & ISPL_3XDAMVDEM) != 0) {
+	if (monster.MData->mMonstClass == MonsterClass::Demon && (player._pIFlags & ISPL_3XDAMVDEM) != 0) {
 			dam *= 3;
-		}
 	}
 
 	if (pnum == MyPlayerId)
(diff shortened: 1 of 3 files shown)


Source/missiles.cpp
```diff
@@ -275,6 +275,10 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 	if (resist)
 		dam >>= 2;
 
+	if (monster.MData->mMonstClass == MonsterClass::Demon && (player._pIFlags & ISPL_3XDAMVDEM) != 0) {
+			dam *= 3;
```
AJenbo, line 279:
```suggestion
		dam *= 3;
```

ikonomov:
This is clearly a bug, but if this bow was doing three times the total char damage it will be far too powerful, totally breaking the balance in the game.  I propose another fix, to simply remove the IPL_3XDAMVDEM affix from its properties and assume that the bug is not implementing the damage, but removing the affix from the item once they decided that the extra damage should not be implemented.

kphoenix137:
> This is clearly a bug, but if this bow was doing three times the total char damage it will be far too powerful, totally breaking the balance in the game.  I propose another fix, to simply remove the IPL_3XDAMVDEM affix from its properties and assume that the bug is not implementing the damage, but removing the affix from the item once they decided that the extra damage should not be implemented.

The scope of DevilutionX isn't to change the balance of the game, but to fix but and improve the game engine

ikonomov:
So let's assume the bug is having the affix.

ikonomov:
I want Deadly Hunter's bug to be fixed, because it should be fixed.  I just pointed out that there are 2 ways to fix it, and one doesn't involve breaking the balance, while the other one does, so when deciding which way to fix it, it might be worth considering.

AJenbo (dismissed):

StephenCWills:
> Afaik this will cause all missiles shot by the player 3x, it should be limited to only arrows.

The `MissilesData[t].mType == 0` condition should only apply to arrows so that should already be handled.

AJenbo (approved):

PR #3865, opened 2021-12-30
Implement unused Stormshield graphic

kphoenix137:
screen68
Note: This may not update the graphic for character files already used in DevilutionX


kphoenix137:
Note: This may not update the graphic for character files already used in DevilutionX

qndel:
think it could be a good moment to replace all these magic numbers with enums

kphoenix137:
> think it could be a good moment to replace all these magic numbers with enums

It would make the table look extremely messy since INVCURS doesn't always fall in the same column

qndel:
> > think it could be a good moment to replace all these magic numbers with enums
> 
> It would make the table look extremely messy since INVCURS doesn't always fall in the same column

so adjust it with spaces? :D

ephphatha (approved):
image
Same with the war bow, old versions use the graphic assigned when the item was dropped so this doesn't break anything (this screenshot was taken on devX master).

FitzRoyX:
The odd thing about this graphic is how it appears to be from a 3d render while the others are hand drawn. It's also at a different angle. So it may have gone unused for consistency reasons.

ikonomov:
> The odd thing about this graphic is how it appears to be from a 3d render while the others are hand drawn. It's also at a different angle. So it may have gone unused for consistency reasons.

I agree.  The reason they chose not to use this graphic seems to be a valid one.

AJenbo:
Closing based on feedback and my own feelings.

PR #3866, opened 2021-12-30
Implement unused Long War Bow graphic

kphoenix137:
screen69
Note: This may not update the graphic for character files already used in DevilutionX


qndel:
You actually haven't done anything.
You replaced `ICURS_LONG_WAR_BOW` (was 119) with `ICURS_LONG_BATTLE_BOW`
and you added `ICURS_LONG_BATTLE_BOW             = 119,`

Chance4us:
Thank you, it would be a great added value.

ephphatha:
One of the item pack test cases needs to be updated to handle the changed graphic for war bows, I think it's "Emerald Bow of burning" on line 328 of pack_test.cpp

kphoenix137:
> 
Thanks! Was wondering how to fix that

ephphatha (approved):
2022-01-03T120612_devilutionx
The left bow is a long war bow dropped in a save before this change, the right bow is a long war bow dropped with this change (the centre is a long battle bow).

This actually works nicely going backwards, old versions load the new graphic for that item. I think it's good to include :)

PR #3867, opened 2021-12-30
Implement Difficulty in Automap

kphoenix137:
screen70
Adds difficulty display as well as capitalizing Game and Password.

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -18,6 +18,7 @@
 #include "utils/language.h"
 #include "utils/stdcompat/algorithm.hpp"
 #include "utils/ui_fwd.h"
+#include "utils/utf8.hpp"
 
 namespace devilution {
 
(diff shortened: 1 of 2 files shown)


AJenbo:
Lets remove it from the version string

Source/automap.cpp
```diff
@@ -509,7 +509,20 @@ void DrawAutomapText(const Surface &out)
...
+	case DIFF_NORMAL:
+		strcpy(desc, _("Difficulty: Normal"));
+		break;
+	case DIFF_NIGHTMARE:
+		strcpy(desc, _("Difficulty: Nightmare"));
+		break;
+	case DIFF_HELL:
+		strcpy(desc, _("Difficulty: Hell"));
+		break;
 	}
```
ephphatha, line 524: Pulling the constant "Difficulty: " text into a format string would allow reusing existing translations for this context, this approach is used in diablo.cpp:1488 (KeyMapper game info action) as well as others.

```suggestion
	string_view difficulty;
	switch (sgGameInitInfo.nDifficulty) {
	case DIFF_NORMAL:
		difficulty = _("Normal");
		break;
	case DIFF_NIGHTMARE:
		difficulty = _("Nightmare");
		break;
	case DIFF_HELL:
		difficulty = _("Hell");
		break;
	}
	CopyUtf8(desc, fmt::format(
	    _(/* TRANSLATORS: {:s} means: Game Difficulty. */ "Difficulty: {:s}"),
	    difficulty
	), sizeof(desc));
```

(untested, might not compile as-is)

kphoenix137:
> Lets remove it from the version string

Done

AJenbo (approved):

PR #3870, opened 2021-12-31
Implement Level Names

kphoenix137:
Replaces default numerical level names with relevant text strings that represent the current tileset. Changes the names in both the triggers and automap text. "Town" text has been modified to no longer be hardcoded. Town Portals will now show you where they lead to from the town side only.
screen74
screen75


Source/trigs.cpp
```diff
@@ -57,6 +57,34 @@ int L6TWarpUpList[] = { 79, 80, 81, 82, 83, 84, 85, 86, 87, 88, 89, 90, 91, 92,
 int L6UpList[] = { 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, -1 };
 int L6DownList[] = { 57, 58, 59, 60, 61, 62, 63, 64, -1 };
 
+const char *LvlNames[] = {
+	"Town",
+	"Dungeon 1",
+	"Dungeon 2",
+	"Dungeon 3",
+	"Dungeon 4",
```
AJenbo, line 65: Cathedral

kphoenix137: Used Dungeon based off the text that displays for the cathedral trigger for consistency. Still change?

kphoenix137: > Cathedral

And how do you feel about renaming town to tristram?

ephphatha: I reckon Cathedral here does read better (I think you were going that way based off the discussion in discord anyway, but adding another vote for :P)

AJenbo: That trigger probably refers to all levels rather then the first one.

Source/cursor.cpp
```diff
@@ -216,7 +216,10 @@ void CheckTown()
 				trigflag = true;
 				ClearPanel();
 				strcpy(infostr, _("Town Portal"));
-				strcpy(tempstr, fmt::format(_("from {:s}"), Players[missile._misource]._pName).c_str());
+				if (currlevel == 0)
+					strcpy(tempstr, fmt::format(_("to {:s} from {:s}"), LvlNames[Portals[i].level], Players[missile._misource]._pName).c_str());
+				else
+					strcpy(tempstr, fmt::format(_("from {:s}"), Players[missile._misource]._pName).c_str());
```
obligaron, line 222: Portals can also lead to quest levels (see `Portal.setlevel`).

```suggestion
				if (currlevel != 0)
					CopyUtf8(tempstr, fmt::format(_("from {:s}"), Players[missile._misource]._pName), sizeof(infostr));				
				else if (Portals[i].setlevel)
					CopyUtf8(tempstr, fmt::format(_("to {:s} from {:s}"), _(QuestLevelNames[Portals[i].level]), Players[missile._misource]._pName), sizeof(infostr));
				else
					CopyUtf8(tempstr, fmt::format(_("to {:s} from {:s}"), _(LvlNames[Portals[i].level]), Players[missile._misource]._pName), sizeof(infostr));
```

Source/cursor.cpp
```diff
@@ -237,7 +240,7 @@ void CheckRportal()
 				if (!setlevel)
 					strcpy(tempstr, _("The Unholy Altar"));
 				else
-					strcpy(tempstr, _("level 15"));
+					strcpy(tempstr, LvlNames[15]);
```
obligaron, line 243: Translation support 🙂 

```suggestion
					strcpy(tempstr, _(LvlNames[15]));
```

Source/trigs.cpp
```diff
@@ -58,6 +58,34 @@ int L6TWarpUpList[] = { 79, 80, 81, 82, 83, 84, 85, 86, 87, 88, 89, 90, 91, 92,
 int L6UpList[] = { 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, -1 };
 int L6DownList[] = { 57, 58, 59, 60, 61, 62, 63, 64, -1 };
 
+const char *LvlNames[] = {
```
obligaron, line 61: All names should be marked with `N_()` macro for translation support.

Source/trigs.cpp
```diff
@@ -409,10 +437,7 @@ bool ForceL1Trig()
 	if (currlevel < 17) {
 		for (int i = 0; L1UpList[i] != -1; i++) {
 			if (dPiece[cursPosition.x][cursPosition.y] == L1UpList[i]) {
-				if (currlevel > 1)
-					strcpy(infostr, fmt::format(_("Up to level {:d}"), currlevel - 1).c_str());
-				else
-					strcpy(infostr, _("Up to town"));
+				strcpy(infostr, fmt::format(_("Up to {:s}"), LvlNames[currlevel - 1]).c_str());
```
obligaron, line 440:
```suggestion
				CopyUtf8(infostr, fmt::format(_("Up to {:s}"), _(LvlNames[currlevel - 1])), sizeof(infostr));
```

(and more places)

obligaron (commented):
Overall I like the changes and the adjusted names.

But we can simply and unify this even more. 
Imho we shouldn't add `LvlNames`.
We could modify (and move) `discord_manager::GetLocationString()`. The difference is that we would pass a setlevel and level into the function instead of using the `tracked_data`.
That would also reduce the number of texts that need to be translated and would ensure that we use the same text in game and discord.

Source/trigs.cpp
```diff
@@ -58,6 +58,34 @@ int L6TWarpUpList[] = { 79, 80, 81, 82, 83, 84, 85, 86, 87, 88, 89, 90, 91, 92,
...
+	"Diablo",
+	"Nest 1",
+	"Nest 2",
+	"Nest 3",
+	"Nest 4",
+	"Crypt 1",
+	"Crypt 2",
+	"Crypt 3",
+	"Crypt 4",
+};
```
julealgon, line 87: Starting the count back in 1 for each tileset is a pretty big change in design/purpose. The old number was supposed to represent overall depth and this is completely lost with this change. The restarting number makes each tileset "disconnected" with the others, it loses the "sequence" aspect. That could be fine for hive and crypt, but it really goes against the original non-Hellfire design in Diablo.

Additionally, I'm really not a fan of having a hardcoded array of names like this that includes the number.. I'd rather have a function that generates the name based on the depth, and tileset names.

julealgon (changes requested):

AJenbo:
Some controversy and there are a few unresolved deficiencies and merge conflicts. Better for this to be done as separate cleanups, fixes, and changes.

PR #3872, opened 2021-12-31
Enable Dungeon Messages in Multiplayer

kphoenix137:
Enables the Dungeon Message audio for entering levels for the first time in Multiplayer.


AJenbo (approved):

ephphatha:
Removing references to globals is always a good thing :D. One thing to note is that `pDungMsgs` isn't carried over between games so these voice lines will play once *per lobby*. This matches what happens for single player characters when you select new game anyway, single player games loaded from a save are the exceptional case.

Would you like to rebase this branch to resolve the conflict? Otherwise I can do it in the web interface.

kphoenix137:
> Removing references to globals is always a good thing :D. One thing to note is that `pDungMsgs` isn't carried over between games so these voice lines will play once _per lobby_. This matches what happens for single player characters when you select new game anyway, single player games loaded from a save are the exceptional case.
> 
> Would you like to rebase this branch to resolve the conflict? Otherwise I can do it in the web interface.

This is intentional. Would it be more ideal to have it performed once per player?

ephphatha:
Nah, what you've got in this PR is good. Mainly wanted to mention it so if someone complains about the sound playing more than once per character we can say we were aware 😂

AJenbo:
Don't forget to resolve the conflict :)

ephphatha (approved):

PR #3874, opened 2022-01-01
Implement Shrine Descriptions

kphoenix137:
Added Shrine descriptions to the pop-up messages that appear after using a shrine, so the player understand what the Shrine did. Side effects were not included (Shrines casting damaging spells against the player). Implemented Shrine descriptions this way to avoid breaking immersion and difficulty included with the player learning the game and taking risks.


Source/error.cpp
```diff
@@ -66,49 +66,49 @@ const char *const MsgStrings[] = {
...
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Blessed by a benevolent companion!"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "The hands of men may be guided by fate"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Strength is bolstered by heavenly faith"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "The essence of life flows from within"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "The way is made clear when viewed from above"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Salvation comes at the cost of wisdom"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Mysteries are revealed in the light of reason"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Those who are last may yet be first"),
-	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Generosity brings its own rewards"),
+	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Crimson and Azure become as the sun (All potions converted to Rejuvination)"),
```
yuripourre, line 81:
```suggestion
	N_(/* TRANSLATORS: Shrine Text. Keep atmospheric. :) */ "Crimson and Azure become as the sun (All potions converted to Rejuvenation)"),

```

AJenbo:
Instead of editing the original line add it as a separate line.

kphoenix137:
> Instead of editing the original line add it as a separate line. Also I think this should be optional.

Sounds good, I will work on doing both

FitzRoyX:
IMO this option could replace the "disable crippling shrines" option.

Issue #3641, opened 2021-11-28
Find best matching language


AJenbo:
Currently if a user's system language is set to `fr_CA` it will automatically match the `fr` translation after failing to locate a Canadian specific dialect. But for `pt_PT` it will fall back to `en_US` instead of matching `pt_BR`, which would be a better solution in most cases.

Since scanning for files isn't to portable and we already need a curated list, it might be good to change `HasTranslation()` to simply use the list from https://github.com/diasurgical/devilutionX/blob/master/Source/options.cpp#L758

This way we can simply match the string lengths (first 5, then 2) when looking for matches.

ephphatha:
Having a play around with windows internationalisation and the win32 api provides a function to retrieve the list of supported languages for the system, maybe other platforms have similar functionality? on my system I have it configured for `en-AU`, `en-GB`, `en-US` and could reasonably expect a fallback to `en` if no other translation exists.

The current approach of using the default language if the regional variation doesn't exist seems appropriate as long as we ensure that a neutral translation is provided for all languages where we have a regional translation.

AJenbo:
I don't think such a thing as a neutral language variant exists. But we could have a default if more then one transformation exists.

julealgon:
> But for pt_PT it will fall back to en_US instead of matching pt_BR, which would be a better solution in most cases.

Are you sure about that? How did you conclude this?

I'd personally not want to fallback to pt-PT if pt-BR was missing (the opposite situation), and I don't think that's common at all (I might be wrong here).

AJenbo:
@julealgon do you have you system language set to `pt` or `en`?

10 base languages most commonly spoken in regions with low English proficiency, based on number from the EF:
Arabic, Chinese, French, Japanese, Portuguese, Punjabi, Russian, Spanish, Tamil, Turkic

Additionally from what I have read and the people I have spoken with it seams that most variants of are intelligible to the speakers of other variants. Brazilian is supposedly one of them most derived variants.

julealgon:
>@julealgon do you have you system language set to pt or en?

My personal laptop is unfortunately setup with pt-BR as it is a single language OEM Windows install. I don't understand how relevant the question is however. What are you alluding to?

> 10 base languages most commonly spoken in regions with low English proficiency, based on number from the EF:
Arabic, Chinese, French, Japanese, Portuguese, Punjabi, Russian, Spanish, Tamil, Turkic

Again, I'm not getting your point here.

[...]

PR #3875, opened 2022-01-01
Get user preferred languages on apple devices

ephphatha:
Adapts the sample code provided by @bubio in #3641 so GetLocales() returns a list of POSIX language codes on apple devices (hopefully... I don't have an environment set up to cross-compile this, lets see if the build works).

Co-authored-by: Bubio <bubio66@gmail.com>

### DIFF
--- a/Source/platform/locale.cpp
+++ b/Source/platform/locale.cpp
@@ -15,13 +15,42 @@
 #include <windows.h>
 #include <winnls.h>
 // clang-format on
-
-#include "utils/stdcompat/algorithm.hpp"
+#elif defined(__APPLE__) and defined(USE_COREFOUNDATION)
+#include <CoreFoundation/CoreFoundation.h>
 #else
 #include <locale>
 #endif
 
+#include "utils/stdcompat/algorithm.hpp"
+#include "utils/stdcompat/string_view.hpp"
+
 namespace devilution {
+namespace {
+
+std::string IetfToPosix(string_view langCode)
+{
+	/*
+	 * Handle special case for simplified/traditional Chinese. IETF/BCP-47 specifies that only the script should be
+	 *  used to discriminate languages when the region doesn't add additional value. For chinese scripts zh-Hans is
+	 *  preferred over zh-Hans-CN (but platforms may include the region identifier anyway). POSIX locales don't use
+	 *  script in the same way so we need to convert these back to lang_region.
+	 */
+	if (langCode.substr(0, 7) == "zh-Hans") {
+		return "zh_CN";
+	}
+	if (langCode.substr(0, 7) == "zh-Hant") {
+		return "zh_TW";
+	}
+
+	std::string posixLangCode { langCode };
+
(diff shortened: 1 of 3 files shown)


Source/platform/locale.cpp
```diff
@@ -15,13 +15,42 @@
...
+
+std::string IetfToPosix(string_view langCode)
+{
+	/*
+	 * Handle special case for simplified/traditional Chinese. IETF/BCP-47 specifies that only the script should be
+	 *  used to discriminate languages when the region doesn't add additional value. For chinese scripts zh-Hans is
+	 *  preferred over zh-Hans-CN (but platforms may include the region identifier anyway). POSIX locales don't use
+	 *  script in the same way so we need to convert these back to lang_region.
+	 */
+	if (langCode.substr(0, 7) == "zh-Hans") {
```
AJenbo, line 38: starts_with

ephphatha: That was introduced in C++20 unfortunately, doesn't look like we have a polyfill

AJenbo (dismissed):

AJenbo (approved):

bubio:
If locales is
    "en-JP",
    "ja-JP",
then the selected one is
   "ja-JP"
is the correct answer?

AJenbo:
No, it should result in ja_JP

ephphatha:
If you run with verbose logging active, what output do you get for the log message in options.cpp:935? (The line "Found {} user preferred locales")

bubio:
> If you run with verbose logging active, what output do you get for the log message in options.cpp:935? (The line "Found {} user preferred locales")

I don't know how to do it.

ephphatha:
Ahh, ok. With the order of configured languages on your system English/US English is prioritised below Japanese (and Korean... sorta). Is this your usual language order preference? If you got rid of the English (US) and English (Australia) entries you'd end up with the neutral english fallback happening almost immediately.

The search order is:
1. English (Japan) - not found, no fallback as a different english variant is provided further down the list.
2. Japanese (Japan) - redundant region specifier, not found with this tag.
3. Japanese - "neutral" fallback code, found. *search stops here*
4. Korean (Japan) - not found
5. Korean - "neutral" fallback code, this would not find a match as the provided translation is for Korean (Korea)
6. English - this was English (US), was transformed by the special case handling for the base language. This would be the language selected if step 3 wasn't a match.
7. Portuguese (Brazil) - this would be selected if both step 3 and 6 weren't matches.
8. Portuguese - "neutral" fallback code, not found
9. Chinese (Simplified) - next match if 3,6,7 failed
10. English (Australia) - not found
11. English - "neutral" fallback, duplicate of 6 because of the special handling for English (US).
12. Chinese (Traditional) - next match if 3,6,7,9 failed
13. Chinese - "neutral" fallback

bubio:
Thank you very much.
I understand now.
 macOS 12
 iPad OS 15.2
I have confirmed that it works fine.


Usually, it's just ja_JP :)
For testing purposes, I set it up.

PR #3876, opened 2022-01-01
move FPS display down due to new display of difficulty

Trihedraf:
screen00
screen01


AJenbo (approved):

Trihedraf:
@AJenbo looks like it is stuck

PR #3880, opened 2022-01-01
Actualize Spanish translation

hiperiondev:
- Shorten texts to accommodate the settings menu
- Actualize translation from source


AJenbo (approved):

AJenbo:
@hiperiondev @Mad-Soft if either of you have time to update the translation now would be a good time to make it for the next release :)

There are about 15 lines that needs an update.

PR #3882, opened 2022-01-01
Create variable MSOURCE_TRAP

kphoenix137:
Creates a variable MSOURCE_TRAP defined as -1 to use for checks with _misource.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -30,7 +30,7 @@ int AvailableMissiles[MAXMISSILES];
 Missile Missiles[MAXMISSILES];
 int ActiveMissileCount;
 bool MissilePreFlag;
-constexpr int MSOURCE_TRAP = -1;
+constexpr int MSOURCE_OBJECT = -1;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -1301,7 +1302,7 @@ void AddFireRune(Missile &missile, Point dst, Direction /*midir*/)
 void AddLightningRune(Missile &missile, Point dst, Direction /*midir*/)
 {
 	int id = missile._misource;
-	int lvl = (id > -1) ? Players[id]._pLevel : 0;
+	int lvl = (id > MSOURCE_TRAP) ? Players[id]._pLevel : 0;
```
AJenbo, line 1305:
```suggestion
	int lvl = id != MSOURCE_TRAP ? Players[id]._pLevel : 0;
```

Source/missiles.cpp
```diff
@@ -30,6 +30,7 @@ int AvailableMissiles[MAXMISSILES];
 Missile Missiles[MAXMISSILES];
 int ActiveMissileCount;
 bool MissilePreFlag;
+constexpr int MSOURCE_TRAP = -1;
```
AJenbo, line 33: Maybe `MSOURCE_OBJECT` or `MSOURCE_WORLD` would be better names like you suggested.

But also this should be used where the value is being set in the first place, not just for the check.

kphoenix137: Got the first part. Not sure what you mean for the second part. What am I missing?

AJenbo:
Good clean up, but I liked @qndel take on it better, thanks to both of you :)

PR #3883, opened 2022-01-01
Cleanup missiles.cpp

kphoenix137:
Cleans up missiles.cpp variables


Source/missiles.cpp
```diff
@@ -2812,15 +2812,16 @@ void MI_LArrow(Missile &missile)
 {
 	missile._mirange--;
 	int p = missile._misource;
+	auto &player = Players[p];
```
ephphatha, line 2815: In this case you want to put the declaration of `player` inside each if statement where it gets used. If _misource is -1 this creates a reference to an invalid area of memory currently.

AJenbo (approved):

Issue #3834, opened 2021-12-23
In the settings menu not all texts are displayed correctly in translation


hiperiondev:
Some texts are shown below the scroll bar, another by the top and others are clipped to the next line.
Tested on Spanish translation.

AJenbo:
Any chance you could post a screen shot of this :)

hiperiondev:
![1](https://user-images.githubusercontent.com/17483817/147185719-04b46879-a871-4775-afdd-181a6ffe804b.png)

![2](https://user-images.githubusercontent.com/17483817/147185734-2e7c95b7-141d-4272-8ce1-189b684f10ed.png)

![3](https://user-images.githubusercontent.com/17483817/147185739-dcde80c2-5c8f-4448-98ac-1923d19a1044.png)

hiperiondev:
I understand that the texts can be shortened but I think it is a bug that behaves differently in each case

ephphatha:
Screenshots 1 and 2 show the same behaviour, options entries are only allocated a single line so when they wrap they overflow the next line. There's a blank line as padding between sections which allows the line to be visible in shot 1, shot 2 ends up getting overdrawn by the next line (you can see the O character behind ap in "Lanzamiento Rapido").

Shot 3 is an interesting case, the way words are wrapped at the moment relies on them being drawn fully outside the containing rectangle. If characters at the end of a line overlap the in bounds area by even a pixel then they don't get wrapped.
![2021-12-28T233623_devilutionx](https://user-images.githubusercontent.com/357684/147567348-8076710c-5215-4e8a-8b98-406d39296428.png)
The red box is the containing rectangle for settings menu items, the S at the start of the line and M at the end of the line are drawn partially out of bounds.

The left boundary isn't calculated properly either, characters at the start of a line are free to wander left as far as necessary as if the text fit on a single line.
![2021-12-28T234018_devilutionx](https://user-images.githubusercontent.com/357684/147567327-f8a06689-dea7-44f6-9560-4e24774a4a21.png)

hiperiondev:
> Screenshots 1 and 2 show the same behaviour.

Not exactly the same, I think. In the first screen "Titulo" is wrapped after "Tit", not overlapped in any letter nor missing. In the second screen "Apagado" miss the last letter.

ephphatha:
![image](https://user-images.githubusercontent.com/357684/147861524-6f43ef21-0488-44eb-b54c-f89cc54caa89.png)
[...]

PR #3884, opened 2022-01-01
Use native script for translated language names

### LINKED ISSUE #3881: Languages can't be translated
Language names are not found in the translation po file and cannot be translated.

Languages

ephphatha:
Without fonts.mpq:
2022-01-02T205847_devilutionx

With fonts.mpq:
2022-01-02T210050_devilutionx

Requires fonts.mpq in order to display options with extended scripts like korean/japanese/chinese. ~Also not sure the strings chosen are appropriate in all cases, e.g. I think "Limba română" literally translated means "language of romania"? "Român" might be better there.~

fixes #3881 

~Also note the line for Portugese (Brazil) is affected by the bug described in #3834. It's long enough that if wrapping was calculating the correct bounds it'd take up two lines. I'd rather fix that bug than shorten these lines as a workaround.~

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -957,25 +957,25 @@ void OptionEntryLanguageCode::CheckLanguagesAreInitialized() const
 		return;
 
 	// Add well-known supported languages
-	languages.emplace_back("bg", "Bulgarian");
-	languages.emplace_back("cs", "Czech");
-	languages.emplace_back("da", "Danish");
-	languages.emplace_back("de", "German");
+	languages.emplace_back("bg", "Български - Bulgarian");
+	languages.emplace_back("cs", "Čeština - Czech");
+	languages.emplace_back("da", "Dansk - Danish");
+	languages.emplace_back("de", "Deutsch - German");
 	languages.emplace_back("en", "English");
-	languages.emplace_back("es", "Spanish");
-	languages.emplace_back("fr", "French");
-	languages.emplace_back("ja", "Japanese");
-	languages.emplace_back("hr", "Croatian");
-	languages.emplace_back("it", "Italian");
-	languages.emplace_back("ko_KR", "Korean");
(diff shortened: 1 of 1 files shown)


AJenbo:
How about checking if fonts.mpq is loaded and either leaving out the languages that require it, or writing them in english. I think I would prefer leaving them out so that users don't end up with a blank screen.

and for the other names I think it would be better to just have them in there native language and avoid the english one.

AJenbo:
@licaon-kter can you check the name for Romanian here?

licaon-kter:
@ephphatha `Română` is the right word

NikoVP:
Is it possible to implement loading a image instead of a text string for Asian languages? You could have their names and description (along with the prompt to install the fonts.mpq) as images in the main mpq. Would save space and would work without having the fonts files?

ephphatha:
> Is it possible to implement loading a image instead of a text string for Asian languages?

The way this list is built and rendered would make it difficult to use an image instead for those options. Possible, but a bit too much effort for me at the moment 😓 

> How about checking if fonts.mpq is loaded and either leaving out the languages that require it, or writing them in english. I think I would prefer leaving them out so that users don't end up with a blank screen.

The options menu is able to access the `font_mpq` global so I've gone with skipping these languages if it's not loaded. If someone sets their language as ko_KR and they don't have fonts.mpq then it'll still be visible in the list (as "ko_KR") because of the fallback, but they've already got broken rendering at that point so I think that's acceptable given it displays what the current setting is.

> and for the other names I think it would be better to just have them in there native language and avoid the english one.

Done 😄 . I also noticed Russian and Ukrainian strings translated to "Russian language"/"Ukrainian language", so I cut them down to match Romanian.

AJenbo (approved):
Nice work on this 👍

PR #3889, opened 2022-01-03
More QuickSpell Hotkeys

Drakker:
- Added 12 more QuickSpell hotkeys for a total of 16.
- Preserve compatibility with single player in LoadPlayer/SavePlayer by saving/loading only 4 hotkeys.
- The number of QuickSpells is hard coded now, but it's easy to change. The LoadHotkeys()/SaveHotkeys() functions already handles a different number of saved hotkeys by reading/writing the number of hotkeys at the beginning of the "hotkeys" file.

Note: when loading older save files, the LoadHotkeys() function will read offset data/garbage, in my tests this doesn't cause problems, but I might just have been lucky, this probably needs some more testing.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1449,12 +1449,12 @@ void InitKeymapActions()
 		    CanPlayerTakeAction,
 		    i + 1);
 	}
-	for (int i = 0; i < 4; ++i) {
+	for (size_t i = 0; i < NumHotkeys; ++i) {
 		sgOptions.Keymapper.AddAction(
 		    "QuickSpell{}",
 		    N_("Quick spell {}"),
 		    N_("Hotkey for skill or spell."),
-		    DVL_VK_F5 + i,
+		    i < 4 ? DVL_VK_F5 + i : DVL_VK_INVALID,
 		    [i]() {
 			    if (spselflag) {
 				    SetSpeedSpell(i);
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -372,11 +372,13 @@ void LoadPlayer(LoadHelper &file, Player &player)
 	player._pSpellFlags = static_cast<SpellFlag>(file.NextLE<uint8_t>());
 	file.Skip(3); // Alignment
 
-	for (auto &spell : player._pSplHotKey)
-		spell = static_cast<spell_id>(file.NextLE<int32_t>());
-
-	for (auto &spellType : player._pSplTHotKey)
-		spellType = static_cast<spell_type>(file.NextLE<int8_t>());
+	// Extra hotkeys: to keep single player save compatibility, read only 4 hotkeys here, rely on LoadHotkeys for the rest
+	for (size_t i = 0; i < 4; i++) {
+		player._pSplHotKey[i] = static_cast<spell_id>(file.NextLE<int32_t>());
+	}
+	for (size_t i = 0; i < 4; i++) {
+		player._pSplTHotKey[i] = static_cast<spell_type>(file.NextLE<uint8_t>());
+	}
 
 	file.Skip<int32_t>(); // Skip _pwtype
 	player._pBlockFlag = file.NextBool8();
@@ -1093,11 +1095,13 @@ void SavePlayer(SaveHelper &file, const Player &player)
 	file.WriteLE<uint8_t>(static_cast<uint8_t>(player._pSpellFlags));
 	file.Skip(3); // Alignment
 
-	for (auto &spellId : player._pSplHotKey)
-		file.WriteLE<int32_t>(spellId);
-
-	for (auto &spellType : player._pSplTHotKey)
-		file.WriteLE<int8_t>(spellType);
+	// Extra hotkeys: to keep single player save compatibility, write only 4 hotkeys here, rely on SaveHotkeys for the rest
+	for (size_t i = 0; i < 4; i++) {
+		file.WriteLE<int32_t>(player._pSplHotKey[i]);
+	}
+	for (size_t i = 0; i < 4; i++) {
+		file.WriteLE<uint8_t>(player._pSplTHotKey[i]);
+	}
 
 	file.WriteLE<int32_t>(player.UsesRangedWeapon() ? 1 : 0);
 	file.WriteLE<uint8_t>(player._pBlockFlag ? 1 : 0);
@@ -1720,6 +1724,13 @@ bool IsHeaderValid(uint32_t magicNumber)
 	return false;
 }
 
+// Returns the size of the hotkeys file with the number of hotkeys passed and if a header with the number of hotkeys is present in the file
+size_t HotkeysSize(size_t nHotkeys = NumHotkeys)
+{
+	//     header            spells                         spell types                    active spell      active spell type
+	return sizeof(uint8_t) + (nHotkeys * sizeof(int32_t)) + (nHotkeys * sizeof(uint8_t)) + sizeof(int32_t) + sizeof(uint8_t);
+}
+
 void LoadHotkeys()
 {
 	LoadHelper file(OpenSaveArchive(gSaveNumber), "hotkeys");
@@ -1727,32 +1738,59 @@ void LoadHotkeys()
 		return;
 
 	auto &myPlayer = Players[MyPlayerId];
+	size_t nHotkeys = 4; // Defaults to old save format number
 
-	for (auto &spellId : myPlayer._pSplHotKey) {
-		spellId = static_cast<spell_id>(file.NextLE<int32_t>());
+	// Refill the spell arrays with no selection
+	std::fill(myPlayer._pSplHotKey, myPlayer._pSplHotKey + NumHotkeys, SPL_INVALID);
+	std::fill(myPlayer._pSplTHotKey, myPlayer._pSplTHotKey + NumHotkeys, RSPLTYPE_INVALID);
+
+	// Checking if the save file has the old format with only 4 hotkeys and no header
+	if (file.IsValid(HotkeysSize(nHotkeys))) {
+		// The file contains a header byte and at least 4 entries, so we can assume it's a new format save
+		nHotkeys = file.NextLE<uint8_t>();
 	}
-	for (auto &spellType : myPlayer._pSplTHotKey) {
-		spellType = static_cast<spell_type>(file.NextLE<int8_t>());
+
+	// Read all hotkeys in the file
+	for (size_t i = 0; i < nHotkeys; i++) {
+		// Do not load hotkeys past the size of the spell types array, discard the rest
+		if (i < NumHotkeys) {
+			myPlayer._pSplHotKey[i] = static_cast<spell_id>(file.NextLE<int32_t>());
+		} else {
+			file.Skip<int32_t>();
+		}
 	}
+	for (size_t i = 0; i < nHotkeys; i++) {
+		// Do not load hotkeys past the size of the spells array, discard the rest
+		if (i < NumHotkeys) {
+			myPlayer._pSplTHotKey[i] = static_cast<spell_type>(file.NextLE<uint8_t>());
+		} else {
+			file.Skip<uint8_t>();
+		}
+	}
+
+	// Load the selected spell last
 	myPlayer._pRSpell = static_cast<spell_id>(file.NextLE<int32_t>());
-	myPlayer._pRSplType = static_cast<spell_type>(file.NextLE<int8_t>());
+	myPlayer._pRSplType = static_cast<spell_type>(file.NextLE<uint8_t>());
 }
 
 void SaveHotkeys()
 {
 	auto &myPlayer = Players[MyPlayerId];
 
-	const size_t nHotkeyTypes = sizeof(myPlayer._pSplHotKey) / sizeof(myPlayer._pSplHotKey[0]);
-	const size_t nHotkeySpells = sizeof(myPlayer._pSplTHotKey) / sizeof(myPlayer._pSplTHotKey[0]);
+	SaveHelper file(CurrentSaveArchive(), "hotkeys", HotkeysSize());
 
-	SaveHelper file(CurrentSaveArchive(), "hotkeys", (nHotkeyTypes * 4) + nHotkeySpells + 4 + 1);
+	// Write the number of spell hotkeys
+	file.WriteLE<uint8_t>(static_cast<uint8_t>(NumHotkeys));
 
+	// Write the spell hotkeys
 	for (auto &spellId : myPlayer._pSplHotKey) {
 		file.WriteLE<int32_t>(spellId);
 	}
 	for (auto &spellType : myPlayer._pSplTHotKey) {
 		file.WriteLE<uint8_t>(spellType);
 	}
+
+	// Write the selected spell last
 	file.WriteLE<int32_t>(myPlayer._pRSpell);
 	file.WriteLE<uint8_t>(myPlayer._pRSplType);
 }
--- a/Source/player.h
+++ b/Source/player.h
@@ -34,6 +34,7 @@ namespace devilution {
 #define MAX_SPELL_LEVEL 15
 #define PLR_NAME_LEN 32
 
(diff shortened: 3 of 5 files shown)


yuripourre:
@Drakker since you are changing it, you could extract a constant instead of using `16`.

ephphatha (commented):
Given the hotkey spell ID and spell type are saved as two separate arrays yuripourre's suggestion of adding a constant is worthwhile, you'd be able to use that for the loops in diablo.cpp and spell_list.cpp (as they use values from both arrays it's more appropriate to use the constant than relying on `.size()` of one) and when declaring the arrays in the player struct.

Source/loadsave.cpp
```diff
@@ -363,11 +363,13 @@ void LoadPlayer(LoadHelper &file, Player &player)
 	player._pSpellFlags = file.NextLE<uint8_t>();
 	file.Skip(3); // Alignment
 
-	for (auto &spell : player._pSplHotKey)
-		spell = static_cast<spell_id>(file.NextLE<int32_t>());
-
-	for (auto &spellType : player._pSplTHotKey)
-		spellType = static_cast<spell_type>(file.NextLE<int8_t>());
+	// Extra hotkeys: to keep single player save compatibility, read only 4 hotkeys here, rely on LoadHotkeys for the rest
+	for (uint8_t i = 0; i < 4; i++) {
```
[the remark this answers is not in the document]
Drakker: Thanks for all the great feedback! I haven't done much C since the late 90s on DOS, so I don't know much about all the "new" stuff and the standard library, this is really helpful.

About the constant for the size, I thought about adding one, but I didn't find any place in the code that looked like a central repository of constants that is automatically included in other files. Where would be the best place to add it?

Thanks!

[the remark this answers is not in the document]
AJenbo: We have migrated a few things to C++ as well so that is probably also going to be different from what you are used to :)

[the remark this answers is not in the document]
ephphatha: > About the constant for the size, I thought about adding one, but I didn't find any place in the code that looked like a central repository of constants that is automatically included in other files. Where would be the best place to add it?

In this case I think it would be appropriate to define a `static constexpr` variable inside the player class (in player.h). Something like:

```cpp
// ...
	static constexpr size_t HotKeyCount = 16;
	spell_id _pSplHotKey[Player::HotKeyCount];
	spell_type _pSplTHotKey[Player::HotKeyCount];
// ...
```

I don't know if GCC6 will allow scoping the declaration like this though, it doesn't handle constexpr members in classes particularly well. If that fails to compile moving the declaration with the other constants defined at the top of this file would be fine.

Source/loadsave.cpp
```diff
@@ -1685,11 +1689,26 @@ void LoadHotkeys()
 
 	auto &myPlayer = Players[MyPlayerId];
 
-	for (auto &spellId : myPlayer._pSplHotKey) {
-		spellId = static_cast<spell_id>(file.NextLE<int32_t>());
+	const size_t nHotkeyTypes = sizeof(myPlayer._pSplHotKey) / sizeof(myPlayer._pSplHotKey[0]);
+	const size_t nHotkeySpells = sizeof(myPlayer._pSplTHotKey) / sizeof(myPlayer._pSplTHotKey[0]);
```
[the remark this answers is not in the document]
Drakker: It's a standard array in the struct, are those std::array 100% backward compatible with normal c-style arrays? Could something break if I change it?

I could also use the #define NUM_HOTKEYS from player.h I guess.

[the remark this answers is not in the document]
ephphatha: > It's a standard array in the struct, are those std::array 100% backward compatible with normal c-style arrays? Could something break if I change it?

They're effectively the same, with the way these arrays are used `std::array` and c-style arrays are interchangeable.

> I could also use the #define NUM_HOTKEYS from player.h I guess.

I can't see an existing define under that name? If you're adding a new constant constexpr declarations (added in C++11) are preferred for this sort of use.

Source/player.h
```diff
@@ -33,6 +33,7 @@ namespace devilution {
 #define MAX_SPELL_LEVEL 15
 #define PLR_NAME_LEN 32
 
+static constexpr uint8_t NUM_HOTKEYS = 16;
```
ephphatha, line 36: [readability-identifier-naming](https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html), devilutionX uses `CamelCase` for constexpr variables.

The type is used in comparison with size_t variables and is used to define the size of arrays, `size_t` is more appropriate than `uint8_t`
```suggestion
static constexpr size_t NumHotkeys = 16;
```

Source/loadsave.cpp
```diff
@@ -1677,20 +1681,43 @@ bool IsHeaderValid(uint32_t magicNumber)
...
 	auto &myPlayer = Players[MyPlayerId];
+	size_t nHotkeys = 4;
 
-	for (auto &spellId : myPlayer._pSplHotKey) {
-		spellId = static_cast<spell_id>(file.NextLE<int32_t>());
+	// Checking if the save file has the old format with only 4 hotkeys
+	if (!file.IsValid(HotkeysSize(4,0))) {
+		// If it doesn't match exactly the size expected for 4 hotkeys, assume we have a new style format and read the number of hotkeys from the beginning of the file 
+		nHotkeys = file.NextLE<uint8_t>();
 	}
```
ephphatha: FileHelper::IsValid() returns true if the file is >= the provided size. This check doesn't function as intended since it expects a strict equals comparison.

You'll need to structure it so the check is positive:
```suggestion
	if (file.IsValid(HotkeysSize(4, true))) {
		// The file contains a header byte and and at least 4 entries, so we can assume it's a new format save
		nHotkeys = file.NextLE<uint8_t>();
	}
```
This also removes the need to have the header flag parameter in `HotkeysSize()` since it's set to true in both call sites.

Eventually we could drop support for saves made in devilutionX before this PR is merged, so after the next release this could be inverted for an early return again.
```cpp
	if (!file.IsValid(HotkeysSize(4, true))) {
		// Missing header byte or empty hotkeys file, probably an old save.
		return;
	}

	nHotkeys = file.NextLE<uint8_t>();
```

Source/loadsave.cpp
```diff
@@ -1677,20 +1681,43 @@ bool IsHeaderValid(uint32_t magicNumber)
...
+		}
 	}
+	for (size_t i = 0; i < nHotkeys; i++) {
+		// Do not load hotkeys past the size of the spells array, discard the rest
+		if (i < NUM_HOTKEYS) {
+			myPlayer._pSplTHotKey[i] = static_cast<spell_type>(file.NextLE<int8_t>());
+		} else { 
+			file.Skip<int8_t>();
+		}
+	}
```
ephphatha: If `nHotkeys < NUM_HOTKEYS` then `_pSplHotKey`/`_pSplTHotKey` will have uninitialised data at the end of the arrays (either garbage or whatever hotkeys were set in the last save game). Should fill `_pSplHotKey` with `SPL_INVALID` like what happens in `CreatePlayer` since that array seems to be the first checked in use.

Source/loadsave.cpp
```diff
@@ -1677,20 +1681,43 @@ bool IsHeaderValid(uint32_t magicNumber)
 	return false;
 }
 
+// Returns the size of the hotkeys file with the number of hotkeys passed and if a header with the number of hotkeys is present in the file
+size_t HotkeysSize(size_t nHotkeys=NUM_HOTKEYS, bool header=1) {
+	return (sizeof(uint8_t)*header) + (nHotkeys * sizeof(int32_t)) + (nHotkeys * sizeof(int8_t) + sizeof(int32_t) + sizeof(int8_t));
```
ephphatha: Integer promotion allows implicit conversion from `bool` to `int` but this isn't great for readability/maintainability. devilutionX uses both [modernize-use-bool-literals](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-bool-literals.html) and [readability-implicit-bool-conversion](https://clang.llvm.org/extra/clang-tidy/checks/readability-implicit-bool-conversion.html) which this use of `header` violates.

Source/loadsave.cpp
Drakker: Ok, I did that for _pSplHotKey, but as far as I can see, _pSplTHotKey is never initialized, though when looking around quickly, there are checks in panels/spell_list.cpp that seem to deal with invalid values.

Source/loadsave.cpp
Drakker: Ok, noted, removed it as per your previous suggestion.

Drakker:
Alright, this looks like its ready for a second round of review/improvements. I left the type of array alone in the player struct, otherwise I did everything that was mentioned.

I did tests saving/loading alternatively from this branch and the official 1.3 release on Linux and it works both ways every time, I am now confident that loading old save files will not cause problems.

The only thing I can see that may cause problems is this:

diablo.cpp line 1402:
	for (int i = 0; i < NumHotkeys; ++i) {
		    "QuickSpell{}",
		    N_("Quick spell {}"),
		    N_("Hotkey for skill or spell."),
		    **DVL_VK_F5 + i**,

It's starting at F5 and will go on to F21, which likely doesn't exist, so its binding random stuff. Later it will load the config file and rebind all of those, but if the config file doesn't contain those entries, I guess the random binds will stay. I suppose I could give it an array of keys (F5,F6,F7,F8,W,E,R,T,Y,U,A,D,H,J,B,N) of which F5-F8 seem to be the default keys, and some letters that are available as far as I can see in diablo.cpp. Would that cause other problems?

Thanks again for the great feedback!

ephphatha:
Function keys up to F24 are defined, they just aren't mapped on most keyboards 😄 . Some OSs let you send them with Shift+F1 for F13 and continuing from there. Apparently this used to work in Windows 7 but doesn't since Win 8 ☹️, I also haven't been able to get it to work on a laptop running Ubuntu 21.04.

ephphatha:
I did notice a conflict when playing around with this though. We currently have quick messages bound to F9-F12. This ends up adding a duplicate event handler for those keys that triggers both the quick message action and the spell action.

Drakker:
Ok, I'll add this array, is it worth it to make it longer than 16 entries just in case someone changes the number of keys? I could add F13-F24 to pad it at the end.

obligaron:
Not all actions have a default key, for example `StopHero`.
I think we don't need to add a default-key for the new QuickSpell hotkeys:

```cpp
		    "QuickSpell{}",
		    N_("Quick spell {}"),
		    N_("Hotkey for skill or spell."),
		    i < 4 ? DVL_VK_F5 + i : DVL_VK_INVALID,
```

Drakker:
I added obligaron's suggested fixes and cleaned up the commits. All my tests work fine with this build and I can seamlessly play with the same saves in the official 1.3 release on Linux. I don't see any other potential problems (but I'm new to this code base... so its not worth much). I removed the draft status from the pull request.

Thanks everyone for your time and guidance!

Source/loadsave.cpp
```diff
@@ -1677,39 +1681,73 @@ bool IsHeaderValid(uint32_t magicNumber)
...
 
-	for (auto &spellId : myPlayer._pSplHotKey) {
-		spellId = static_cast<spell_id>(file.NextLE<int32_t>());
+	// Refill the spell arrays with no selection
+	std::fill(myPlayer._pSplHotKey, myPlayer._pSplHotKey + NumHotkeys, SPL_INVALID);
+	std::fill(myPlayer._pSplTHotKey, myPlayer._pSplTHotKey + NumHotkeys, RSPLTYPE_INVALID);
+
+	// Checking if the save file has the old format with only 4 hotkeys and no header
+	if (file.IsValid(HotkeysSize(nHotkeys))) {
+		// The file contains a header byte and and at least 4 entries, so we can assume it's a new format save
```
ephphatha, line 1706:
```suggestion
		// The file contains a header byte and at least 4 entries, so we can assume it's a new format save
```
My typo 😅

Source/player.h
```diff
@@ -33,6 +33,7 @@ namespace devilution {
 #define MAX_SPELL_LEVEL 15
 #define PLR_NAME_LEN 32
 
+static constexpr size_t NumHotkeys = 16;
```
ephphatha, line 36: `static` in this context doesn't do anything.
```suggestion
constexpr size_t NumHotkeys = 16;
```

Source/loadsave.cpp
```diff
@@ -363,11 +363,13 @@ void LoadPlayer(LoadHelper &file, Player &player)
...
-		spell = static_cast<spell_id>(file.NextLE<int32_t>());
-
-	for (auto &spellType : player._pSplTHotKey)
-		spellType = static_cast<spell_type>(file.NextLE<int8_t>());
+	// Extra hotkeys: to keep single player save compatibility, read only 4 hotkeys here, rely on LoadHotkeys for the rest
+	for (size_t i = 0; i < 4; i++) {
+		player._pSplHotKey[i] = static_cast<spell_id>(file.NextLE<int32_t>());
+	}
+	for (size_t i = 0; i < 4; i++) {
+		player._pSplTHotKey[i] = static_cast<spell_type>(file.NextLE<int8_t>());
```
ephphatha, line 371: This set of reads (and the matching writes in `SavePlayer`) can use `uint8_t` to match the updates you made when loading the hotkeys file. uint8_t is more suitable given the cast to an enum and the values used, good to update it when you're in the area anyway :D

Source/diablo.cpp
```diff
@@ -1399,12 +1399,12 @@ void InitKeymapActions()
...
 		    i + 1);
 	}
-	for (int i = 0; i < 4; ++i) {
+	for (size_t i = 0; i < NumHotkeys; ++i) {
 		sgOptions.Keymapper.AddAction(
 		    "QuickSpell{}",
 		    N_("Quick spell {}"),
 		    N_("Hotkey for skill or spell."),
-		    DVL_VK_F5 + i,
+				i < 4 ? DVL_VK_F5 + i : DVL_VK_INVALID,
```
ephphatha, line 1407: clang-format is complaining about this line being indented too much.

ephphatha (dismissed):
Apart from the nits this looks good to me, functional without breaking compatibility. The keymapper section of settings is getting very long but that's not a knock against this PR 😂

Source/loadsave.cpp
Drakker: The player save function writes a int8_t there, LoadHotkeys() has uint8_t, but its separate from the old style saves. I assume that the int8_t here should stay for compatibility.

The LoadHotkeys and SaveHotkeys were another thing, SaveHotkeys was writing a uint8_t, so I made the LoadHotkeys function match it because all existing saves have a uint8_t because of that.

ephphatha: `spell_type` is defined with an explicit backing type of `uint8_t`, reading and writing to the main save file should really use the same type. It's not a required change because we read an 8 bit value and the expected range is 0-4 (there's no sign extension issues), however it'd be nice to have to make the relationship clearer.

AJenbo:
@Drakker very nice work, sorry I didn't have time to go over this and get it merged before now.

AJenbo (approved):

Drakker:
@AJenbo Thanks, no problem, I'm just happy it made it in. I started playing with the automap, when I have more time I'll see if I can get something worth pulling in, like showing lava areas in the caves.

AJenbo:
A town map would also be nice :)

Drakker:
That can be arranged, I suppose it could even be hard coded to make it look really good, as the town never changes.

AJenbo:
Town is a bit different from other levels as it does not have a .amp file which defines the the mapping between the 2x2 files and the automap idiograms

Drakker:
Yeah, I could just load a pre-defined town dataset that's already in the automap "format" and that's hand crafted. But looking at the map code, it could probably use a full rewrite and have more tiles, nothing prevents making a map that looks more like Diablo 2 with more interesting tiles when appropriate (buildings in town, lava and fences in the caves, etc.).

PR #3890, opened 2022-01-04
Update libzt to the latest version

### LINKED ISSUE #3700: Update libzt from upstream
We currently have a few workaround in CMake files, such as having to link the shared libzt library on Android.
Ideally, we'd switch to the upstream libzt but they haven't merged all of our patches yet.

Perhaps we can rebase our stuff on the current upstream?

StephenCWills:
Testing for this update included Windows built using MSVC, Windows built using MinGW, Debian Linux, and Android built using Android Studio on Windows. I submitted a handful of issues and PRs to libzt for the troubles I encountered when testing those builds. In the meantime, I've also incorporated fixes for those troubles in the diasurgical forks for libzt, ZeroTierOne, and lwip-contrib.

The only changes I made that I have not submitted back to libzt are diasurgical/libzt@02a6e96 and diasurgical/libzt@e707330. I imagine the former could be made into an option, like `ANDROID_DISABLE_JAVA`. From what I can tell, the latter seems to be related to the TODO comment above that change as well as the following comment in `zerotier_native.h` so I just left the hardcoded value in.

https://github.com/diasurgical/devilutionX/blob/2e0a73e967ddbc88b805be518357d97e862338b1/Source/dvlnet/zerotier_native.h#L10-L15

The biggest change I did not include was part of diasurgical/libzt@e774a9a. It had to do with replacing `pthread_t` with `std::thread` and the use of a condition variable for shutdown instead of pipes. I couldn't determine what that change was for or, by extension, if it's still necessary.

Nearly all my testing was done using my personal ZT network, to avoid the potential for causing issues with 1.3.0 players. I haven't tried testing this version with the release version.

This resolves #3700


AJenbo (approved):
Tested on macOS and was able to discover  game, and create one. i didnt get to test game play, but we know that it is able to send and receive since i got a message saying im to low level to join

AJenbo (approved):

glebm (approved):

PR #3893, opened 2022-01-04
Translation validation

### LINKED ISSUE #3360: Make script for detecting fmt errors in translations
Currently we have an unfortunate issue with the Spanish and French users of the Android port as they will some times crash because of a translation error, and we cannot push a new update for a while. To avoid this in the future we should not terminate the application in case of invalid fmt string.

https://github.com/diasurgical/devilutionX/blob/4274cc6f1fff3191ff17babb6f2d77274d44e4d9/Source/utils/log.hpp#L52

AJenbo:
Fixes #3360

Not the most advanced thing, but it was able to catch a bunch of issues in the translations:
image

### DIFF
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -1,0 +1,25 @@
+name: Validate translations
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types: [ opened, synchronize ]
+
+jobs:
+  fmt-check:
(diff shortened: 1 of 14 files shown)


ephphatha (approved):
The checking script and CI build look good :)

My french is pretty rusty but the fr.po "{:d} Scroll(s)" line looks correct, it includes a quantity so there's no need for an indefinite article.

PR #3895, opened 2022-01-04
Rename Korean/Romanian translation files so they are used in fallback…

### LINKED ISSUE #3641: Find best matching language

ephphatha:
… resolution

fixes #3641

### DIFF
--- a/CMake/Assets.cmake
+++ b/CMake/Assets.cmake
@@ -2,7 +2,7 @@ if(NOT DEFINED DEVILUTIONX_ASSETS_OUTPUT_DIRECTORY)
   set(DEVILUTIONX_ASSETS_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/assets")
 endif()
 
-set(devilutionx_langs bg cs da de es fr hr it ja ko_KR pl pt_BR ro_RO ru uk sv zh_CN zh_TW)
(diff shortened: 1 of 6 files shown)


ephphatha:
The po files have one changed line each, `"Language: ko_KR\n"` > `"Language: ko\n"` (and ro_RO > ro)

AJenbo:
> The po files have one changed line each, `"Language: ko_KR\n"` > `"Language: ko\n"` (and ro_RO > ro)

Sure, if you ignore that all the line endings where changed from LF to CRLF :D

AJenbo (approved):

ephphatha:
Thanks for that pickup, I think I've got tortoisediff assuming native style line endings while visual studio is set up to use source line endings so if I edit the diff directly I need to remember to set the line ending explicitly -_-

AJenbo:
Thanks for making the change :)

PR #3896, opened 2022-01-05
Mana Shield 1.07 bugfix revision

kphoenix137:
This is a bugfix for the vanilla 1.06 and earlier Mana Shield logic. In 1.07 Blizzard North "fixed" the Mana Shield logic since the damage reduction would decrease in effectiveness with increased spell level instead of increase. As a fix, they set the damage reduction to a flat value regardless of spell level. Originally, damage reduction started at 1/3 and capped at 1/21 at spell level 7. This fix reverses the order and gives 1/21 damage reduction at spell level 1, and caps at 1/3 damage reduction at spell level 7.

End result damage reduction values (based on 100 incoming damage):
SLvl 0 = 0
SLvl 1 = 4
SLvl 2 = 5
SLvl 3 = 6
SLvl 4 = 8
SLvl 5 = 11
SLvl 6 = 16
SLvl 7+ = 33

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2104,6 +2104,12 @@ void Player::Reset()
 	*this = std::move(*emptyPlayer);
 }
 
+int Player::GetManaShieldDamageReduction()
+{
+	constexpr int8_t Max = 7;
+	return 24 - std::min(_pSplLvl[SPL_MANASHIELD], Max) * 3;
+}
+
 void Player::RestorePartialLife()
 {
 	int wholeHitpoints = _pMaxHP >> 6;
@@ -3229,7 +3235,7 @@ void ApplyPlrDamage(int pnum, int dam, int minHP /*= 0*/, int frac /*= 0*/, int 
 	if (totalDamage > 0 && player.pManaShield) {
 		int8_t manaShieldLevel = player._pSplLvl[SPL_MANASHIELD];
 		if (manaShieldLevel > 0) {
-			totalDamage += totalDamage / -3;
+			totalDamage += totalDamage / -player.GetManaShieldDamageReduction();
 		}
 		if (pnum == MyPlayerId)
 			drawmanaflag = true;
@@ -3240,7 +3246,7 @@ void ApplyPlrDamage(int pnum, int dam, int minHP /*= 0*/, int frac /*= 0*/, int 
 		} else {
 			totalDamage -= player._pMana;
 			if (manaShieldLevel > 0) {
-				totalDamage += totalDamage / 2;
+				totalDamage += totalDamage / (player.GetManaShieldDamageReduction() - 1);
 			}
 			player._pMana = 0;
 			player._pManaBase = player._pMaxManaBase - player._pMaxMana;
--- a/Source/player.h
+++ b/Source/player.h
@@ -528,6 +528,13 @@ struct Player {
 	}
 
 	/**
+	 * @brief Return reciprocal of the factor for calculating damage reduction due to Mana Shield.
+	 *
+	 * Valid only for players with Mana Shield spell level greater than zero.
+	 */
+	int GetManaShieldDamageReduction();
+
+	/**
 	 * @brief Return monster armor value after including player's armor piercing % (hellfire only)
 	 * @param monsterArmor - monster armor before applying % armor pierce
 	 * @param isMelee - indicates if it's melee or ranged combat


StephenCWills:
FYI, if the damage exceeds the amount of mana the player has left, the total damage is reduced by the mana amount and the remainder is restored to the unreduced amount assuming a 33% reduction. You'll need to adjust this logic too.

https://github.com/diasurgical/devilutionX/blob/9b32c8d54110ad1a2610d741a41ae3eceeaf658e/Source/player.cpp#L3053-L3056

kphoenix137:
> 

Got it :)

Source/player.cpp
```diff
@@ -3039,8 +3039,12 @@ void ApplyPlrDamage(int pnum, int dam, int minHP /*= 0*/, int frac /*= 0*/, int
 	int totalDamage = (dam << 6) + frac;
 	if (totalDamage > 0 && player.pManaShield) {
 		int8_t manaShieldLevel = player._pSplLvl[SPL_MANASHIELD];
+		int dmgReduction = 24;
```
glebm, line 3042: It'd be good to extract this to a function, so that we can later display it in the spell book.

Source/player.cpp
```diff
@@ -3032,6 +3032,17 @@ void StripTopGold(Player &player)
...
 
+int GetManaShieldDamageReduction(int pnum)
+{
+	auto &player = Players[pnum];
+	int8_t manaShieldLevel = player._pSplLvl[SPL_MANASHIELD];
+	int dmgReduction = 24;
+	for (int i = 0; i < manaShieldLevel && i < 7; i++) {
+		dmgReduction -= 3;
+	}
+	return dmgReduction;
```
glebm, line 3043:
```suggestion
	return 24 - std::min(Players[pnum]._pSplLvl[SPL_MANASHIELD], 6) * 3;
```

Source/player.cpp
```diff
@@ -3033,6 +3033,12 @@ void StripTopGold(Player &player)
 	player.HoldItem = tmpItem;
 }
 
+int GetManaShieldDamageReduction(int pnum)
+{
+	int8_t max = 7;
+	return 24 - std::min(Players[pnum]._pSplLvl[SPL_MANASHIELD], max) * 3;
```
StephenCWills, line 3039:
```suggestion
	constexpr int8_t Max = 7;
	return 24 - std::min(Players[pnum]._pSplLvl[SPL_MANASHIELD], Max) * 3;
```

StephenCWills (dismissed):

Source/player.cpp
```diff
@@ -3033,6 +3033,12 @@ void StripTopGold(Player &player)
 	player.HoldItem = tmpItem;
 }
 
+int GetManaShieldDamageReduction(int pnum)
+{
+	constexpr int8_t Max = 7;
+	return 24 - std::min(Players[pnum]._pSplLvl[SPL_MANASHIELD], Max) * 3;
+}
+
```
ephphatha, line 3041: This should be a member function on the `Player` class, (or at least take a `const Player &` argument).
```suggestion
int Player::GetManaShieldDamageReduction()
{
	constexpr int8_t Max = 7;
	return 24 - std::min(_pSplLvl[SPL_MANASHIELD], Max) * 3;
}
```

It'd then be used in ApplyPlrDamage like `player.GetManaShieldDamageReduction()`

kphoenix137: This method didn't work, but I tried directly adding the function definition into the player struct

StephenCWills (approved):
Squashed and rebased, applied clang format, moved the `GetManaShieldDamageReduction()` function definition, and added documentation comments.

AJenbo (dismissed):
In case anyone was curious what this looked like (Hellfire 1.01 vs Diablo 1.09):
https://github.com/diasurgical/devilution/blob/de6154094d0bc24dcb20e5b8ab479336dd61544c/Source/missiles.cpp#L4876

Source/player.h
```diff
@@ -527,6 +527,17 @@ struct Player {
...
 
+	/**
+	 * @brief Return reciprocal of the factor for calculating damage reduction due to Mana Shield.
+	 *
+	 * Valid only for players with Mana Shield spell level greater than zero.
+	 */
+	int GetManaShieldDamageReduction()
+	{
+		constexpr int8_t Max = 7;
+		return 24 - std::min(_pSplLvl[SPL_MANASHIELD], Max) * 3;
```
AJenbo, line 538: MSVC is having none of this:
image

AJenbo: Probably missing `#include <algorithm>`

StephenCWills: I guess it would make sense to move the definition to player.cpp, then.

StephenCWills (approved):

Chance4us:
Does this bug fix apply to both games diablo and hellfire?

qndel:
`This is a bugfix for the vanilla 1.06 and earlier Mana Shield logic.` which means it worked fine in vanilla 1.09 and was broken in hellfire because hellfire was based on 1.04

AJenbo:
For DevilutionX the behavior has always been that of 1.09 (flat bonus for all levels), also for Hellfire since it applied all the 1.04-1.09 fixes to Hellfire. From 1.4.0 DevilutionX will instead use the fixed original intent of manashield improving with spell level 1-7 in both games.

PR #3897, opened 2022-01-05
Use static libzt on Android

glebm:
Made possible by a recent sync of libzt to upstream


StephenCWills:
Unfortunately, I think they force the shared library on Android, which would explain why the linker can't find the static one.
https://github.com/diasurgical/libzt/blob/a9974991b031fd8f84c4bcfd93f129b3938ed8c8/CMakeLists.txt#L187-L195

glebm:
Ah :(

PR #3898, opened 2022-01-05
Fix Stat Adjustment

kphoenix137:
This fixes a problem related to shrines. When a player with a base Magic or Vitality stat of 0 has their respective stat reduced below 0, they lose Mana/Life, and their stat is corrected to 0 without refunding them the lost Mana/Life. This will prevent stats from being modified to a value less than 0 in the first place.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -3601,12 +3601,7 @@ void ModifyPlrStr(int p, int l)
 	auto &player = Players[p];
 
 	int max = player.GetMaximumAttributeValue(CharacterAttribute::Strength);
-	if (player._pBaseStr + l > max) {
-		l = max - player._pBaseStr;
-	}
-	if (player._pBaseStr + l < 0) {
-		l = 0;
-	}
+	l = clamp(l, 0, max - player._pBaseStr);
 
 	player._pStrength += l;
 	player._pBaseStr += l;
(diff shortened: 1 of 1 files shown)


Source/player.cpp
```diff
@@ -3603,6 +3603,8 @@ void ModifyPlrStr(int p, int l)
 	int max = player.GetMaximumAttributeValue(CharacterAttribute::Strength);
 	if (player._pBaseStr + l > max) {
 		l = max - player._pBaseStr;
+	} else if (player._pBaseStr + l < 0) {
```
glebm: Looks like a good candidate for `clamp` from https://github.com/diasurgical/devilutionX/blob/master/Source/utils/stdcompat/algorithm.hpp

kphoenix137: Having a hard time figuring out how to incorporate this

glebm: Perhaps like this:

```c++
l = clamp(l, 0, max - player._pBaseStr);
```

kphoenix137: Actually after looking at this, it would prevent stats from being reduced by 1 regardless of base stat

Source/player.cpp
```diff
@@ -3601,12 +3601,7 @@ void ModifyPlrStr(int p, int l)
 	auto &player = Players[p];
 
 	int max = player.GetMaximumAttributeValue(CharacterAttribute::Strength);
-	if (player._pBaseStr + l > max) {
-		l = max - player._pBaseStr;
-	}
-	if (player._pBaseStr + l < 0) {
-		l = 0;
-	}
+	l = clamp(l, 0, max - player._pBaseStr);
```
glebm, line 3604: Might as well inline `max` now

kphoenix137: `inline int GetMaxStat(Player &player, CharacterAttribute attribute)
{
	return player.GetMaximumAttributeValue(attribute);
}`
Like this?

Source/player.cpp
```diff
@@ -3680,9 +3671,7 @@ void ModifyPlrVit(int p, int l)
 	auto &player = Players[p];
 
 	int max = player.GetMaximumAttributeValue(CharacterAttribute::Vitality);
```
glebm, line 3673: Unused variable?

glebm (approved):

PR #3901, opened 2022-01-06
Reset game info after failing to join game

StephenCWills:
Prior to this change, game init info would only be initialized at the same time the multiplayer menu is initialized. Joining a game will overwrite game init info, but if you fail to join the game, you will never leave the multiplayer menu and so game init info would never be reinitialized. This can lead to some really unhappy situations if you create a new game right after failing to join one.


StephenCWills:
AppVeyor had a fit.

qndel:
is this a vanilla bug?

AJenbo:
Probably not, maybe if you join a LAN game with a different version.

ephphatha (approved):
The way the `sgGameInitInfo` global is used makes it really hard to reason about where it gets modified and when it needs to be reset. I think this approach is fine, I couldn't find a code path that left garbage in that variable in my testing. I wish there was a cleaner way to clear it before it gets used instead of after it gets corrupted though ☹️

PR #3902, opened 2022-01-06
Menu text revision

kphoenix137:
Reorders and renames items on the Escape Menu, and removes "Restart In Town" with a replacement mechanic.

- Prevents players from accidentally leaving/quitting the game while either in the dying animation, or already dead.
- "New Game" renamed to "Exit Game" to be more accurate.
- Being dead now only allows the player to use the Escape/Menu key/button to Load Game in SP and Restart In Town in MP, as a safety mechanism.

SP:
image

image

MP:
image

image

Note: This PR includes 30-c0.clx, which is used for the gamepad icons for size 30 font, which previously didn't exist. It's simply a copy of the existing gamepad icon clx which exists for smaller text, so the buttons are not any visually larger. It works fine, however a future PR will be a good idea to provide larger button icons that are more visually cohesive.

This PR also does not add 30-c0.pcx to https://github.com/diasurgical/devilutionx-assets


Trihedraf:
I think we should still have an exit button without saving.

kphoenix137:
> I think we should still have an exit button without saving.

I will update the original post to show most recent changes

kphoenix137:
Rebased and cleaned up. No longer removes the red screen from death this time, and no longer displays lost gold. Fixed a problem where the game tried to load a save even if a save didn't exist. Players can no longer use the escape menu during the death sequence before the death message appears, preventing accidental saving.

kphoenix137:
Ready for review

FitzRoyX:
> * "New Game" renamed to "Exit Game" to be more accurate.

"Exit" and "Quit" are practically synonyms, this reads even less clear than before to me. If you were going to rename "New Game" to anything, it should be "Other Game" or "Select Game". Though it hardly seems worth disrupting 20 years of muscle memory for a slightly better word.

"Options" was probably between save and load to prevent accidental clicks by spacing them out. I don't think that's worth undoing either.

The change that turns escape into something else while dead isn't terrible, but it is a more elaborate way of solving the problem vs just letting escape do the same thing all the time and graying out new game.

glebm:
"Exit Game" is indeed a synonym of "Quit Game". Perhaps "Exit to Main Menu"?

ChaosMarc:
What a about "Exit Game" and "Exit to Desktop" or "Exit Diablo/Hellfire"?

kphoenix137:
> "Options" was probably between save and load to prevent accidental clicks by spacing them out. I don't think that's worth undoing either.

Kind of like "Restart in Town" was sandwiched between "New Game" and "Quit Diablo" to encourage accidental clicks that make the player lose all their items or SP progress, as a hardcore feature.

kphoenix137:
I'll need to address non-M&K control schemes, as well as the fact you can still physically click the menu button to bring up the menu while on the death screen.

FitzRoyX:
> Kind of like "Restart in Town" was sandwiched between "New Game" and "Quit Diablo" to encourage accidental clicks that make the player lose all their items, as a hardcore feature since Multiplayer is supposed to be more difficult than Single Player.

Well we all know MP was a rushed feature. The arrangement in MP looks like a simple carryover of SP with "Save" going away and "Restart" taking the place of "Load".

> "Exit Game" is indeed a synonym of "Quit Game". Perhaps "Exit to Main Menu"?

I actually like that "New Game" takes you back to the charselect screen and not title screen. You're more likely to want to load a different game than change modes or settings, so it is effectively a click saver. You can still retreat back from the char screen if you really want.

kphoenix137:
> > Kind of like "Restart in Town" was sandwiched between "New Game" and "Quit Diablo" to encourage accidental clicks that make the player lose all their items, as a hardcore feature since Multiplayer is supposed to be more difficult than Single Player.
> 
> Well we all know MP was a rushed feature. The arrangement in MP looks like a simple carryover of SP with "Save" going away and "Restart" taking the place of "Load". Just put Restart where Save used to be then.
> 
Making a lot of extrapolations here


> > "Exit Game" is indeed a synonym of "Quit Game". Perhaps "Exit to Main Menu"?
> 
> I actually like that "New Game" takes you back to the charselect screen and not title screen. You're more likely to want to load a different game than change modes or settings, so it is effectively a click saver. You can still retreat back from the char screen if you really want.

I didn't change the functionality of "New Game". It still goes back to character select. Anything prior to seeing Tristram is considered the main menu.

Source/control.cpp
```diff
@@ -1414,6 +1415,34 @@ void RedBack(const Surface &out)
...
+	text = _("You have died");
+	DrawString(out, text, linePosition, largeTextOptions);
+	linePosition.y += verticalPadding;
+	if (!gbIsMultiplayer) {
+		if (gbValidSaveFile)
+			text = _("Press ESC to load last save.");
+		else
+			text = _("Press ESC to exit game.");
+	} else {
+		text = _("Press ESC to continue.");
```
AJenbo, line 1441:
```suggestion
		text = _("Press ESC to restart in town.");
```
```suggestion
		text = _("Press ESC to return to town.");
```

FitzRoyX:
> "Exit to Main Menu" still goes back to character select. Anything prior to seeing Tristram is considered the main menu. I think glebm hit the nail on the head with most accurate description of what this option does. "New Game" doesn't create a new game

So the main folder of an application can also be any of its subfolders? Doesnt seem right that nothing is a submenu.

"New" does get used as "different" in English, as in "let me go put on a new shirt". It's not perfect but it's a short word that avoids the exit/quit/leave conflict.

AJenbo:
@kphoenix137 what's missing here?

kphoenix137:
> Right, need any help with that :)

The file is ready to go, although currently not consistent with the existing button graphics (Remember I was also going to make a smaller version of my new button graphics, but for a different PR). I will work on getting that ready on Thursday. Feel free to bug me if I forget.

AJenbo:
Thanks for the update :)

kphoenix137:
Modified original post for clarity about the clx file.

PR #3910, opened 2022-01-06
Settingsmenu: Handle long text

### LINKED ISSUE #3834: In the settings menu not all texts are displayed correctly in translation

obligaron:
Fixes #3834

Notes:
- If an option is too long to display, it will be manually separated in two lines. Similar how the first version of the settings menu handled this.
- After each change to a option this need to be rechecked, cause the value description text can be longer or shorter.
- `UiFlags::NeedsNextElement` didn't work correctly when a item is preselected or with the changed up/down logic. Didn't occur in master, cause in master no one uses `UiFlags::NeedsNextElement`

<details><summary>Sample Video</summary>

https://user-images.githubusercontent.com/25415264/148422581-ba118c30-65a4-40b7-a792-006a3e49ddef.mp4

</details>

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -142,6 +142,8 @@ void UiInitList(void (*fnFocus)(int value), void (*fnSelect)(int value), void (*
 			SelectedItemMax = std::max(uiList->m_vecItems.size() - 1, static_cast<size_t>(0));
 			ListViewportSize = uiList->viewportSize;
 			gUiList = uiList;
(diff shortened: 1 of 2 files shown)


AJenbo (approved):
An acceptable solution, LGTM

obligaron:
Ideally this should be handled in diablo ui and not in the menus. This would help all menus and not only the settings.
Something that could be revisit later. 🙂

PR #3911, opened 2022-01-06
codec.cpp: Fix alignment and endianness issues

glebm:


AJenbo (approved):
What issues was this causing?

glebm:
It wasn't causing any issues on the platforms I can test on but:

1. On big-endian systems, the signature was read and written with the wrong endianness.
2. Unaligned reads are UB.

Can't test a big-endian system since the Amiga version doesn't work at the moment.

PR #3912, opened 2022-01-06
Make Diablo Unique

kphoenix137:
Makes Diablo into a Unique Monster while maintaining his usual stats. Mlvl now scales the same was as unique monsters do, resulting in D1 mlvl 30/60/90 and Hellfire mlvl 45/75/105. This was left in place since the mlvl did not scale as you would expect it to (Normal Leoric has a huge mlvl gap with Normal Diablo, however Hell Leoric has a small level gap with Hell Diablo for example. I believe Hellfire attempted to adjust this issue by increasing the base mlvl of Diablo, however the scaling is still strange..)

This change would work well in conjunction with a change that stops the game from ending upon killing Diablo, in order to better utilize the increased Unique Item drop rate.

image


Source/monster.cpp
```diff
@@ -586,6 +595,137 @@ void PlaceUniqueMonst(int uniqindex, int miniontype, int bosspacksize)
 	}
 }
 
+void PlaceDiablo(int x, int y)
```
glebm, line 598: Is this code copy-pasted from somewhere? Would be nice to dedupe

kphoenix137: The code in here is copy pasted from PlaceUniqueMonst, but has been modified as to not alter Diablo's stats. I also use the x and y coordinates so the game knows where to place him. In PlaceUniqueMonst, no coordinates get provided to this function. Any ideas for a better workaround?

ikonomov:
I am against this change for a mlvl 45, because in Diablo this would allow getting a few items that are not possible to get otherwise.  I think it is a great idea to add some meaningful reward for killing D if he is changed to a unique mlvl 30.  It is very much a QoL improvement, but in MP we could possibly justify it if we are not forced with the "End" cinematic anymore.  All in all, I think both are great improvements, but we can be a little more sensitive to the change so it doesn't have an affect on the balance of the game, introducing some unobtainable items.

kphoenix137:
Gonna close this PR. Too much is affected by doing this, resulting in a rebalance. These changes could be negated however it would be confusing the the player when Diablo doesn't behave like a unique monsters based on mlvl with experience given, and item drops.

PR #3913, opened 2022-01-07
Cleanup Diablo-SHA and codec.cpp

glebm:
1. Switches to 32-bit based calculation throughout. This is faster and less error-prone as we only byte-swap once when reading and writing.
2. Removes global state from Diablo-SHA implementation.


AJenbo (approved):
Looks a lot nicer :+1:

PR #3917, opened 2022-01-07
Apocrypha fixes

StephenCWills:
For review/discussion. Fixes from @SoundChaser83's Apocrypha mod.

> ... it looks like Hellbats and Torchants do absurdly low damage because their fireballs aren't reading damage from the monster data table, and also all monster melee/missile hits do up to + or - 63/64ths points of damage over and above their usual damage range because of a parentheses/order of operations issue.

> Regarding that Torchant/Hellbat fireball fix, those things will hit like an absolute truck if you go through with that fix, so it may be surprising to people.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1177,7 +1177,7 @@ bool PlayerMHit(int pnum, Monster *monster, int dist, int mind, int maxd, missil
 		dam = player._pHitPoints / 3;
 	} else {
 		if (!shift) {
-			dam = (mind << 6) + GenerateRnd((maxd - mind + 1) << 6);
+			dam = (mind << 6) + GenerateRnd(((maxd - mind) << 6) + 1);
 			if (monster == nullptr)
 				if ((player._pIFlags & ISPL_ABSHALFTRAP) != 0)
 					dam /= 2;
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -3093,7 +3093,8 @@ void MI_Fireball(Missile &missile)
 			AddUnLight(missile._mlid);
 		}
 	} else {
-		int dam = missile._midam;
+		auto &monster = Monsters[id];
+		int dam = (missile._micaster == TARGET_MONSTERS) ? missile._midam : monster.mMinDamage + GenerateRnd(monster.mMaxDamage - monster.mMinDamage + 1);
 		MoveMissileAndCheckMissileCol(missile, dam, dam, true, false);
 		if (missile._mirange == 0) {
 			Point m = missile.position.tile;
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -3093,9 +3093,15 @@ void MI_Fireball(Missile &missile)
 			AddUnLight(missile._mlid);
 		}
 	} else {
-		auto &monster = Monsters[id];
-		int dam = (missile._micaster == TARGET_MONSTERS) ? missile._midam : monster.mMinDamage + GenerateRnd(monster.mMaxDamage - monster.mMinDamage + 1);
-		MoveMissileAndCheckMissileCol(missile, dam, dam, true, false);
+		int minDam = missile._midam;
+		int maxDam = missile._midam;
+
+		if (missile._micaster != TARGET_MONSTERS) {
+			auto &monster = Monsters[id];
+			minDam = monster.mMinDamage;
+			maxDam = monster.mMaxDamage;
+		}
+		MoveMissileAndCheckMissileCol(missile, minDam, maxDam, true, false);
 		if (missile._mirange == 0) {
 			Point m = missile.position.tile;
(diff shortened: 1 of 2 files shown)


Source/monster.cpp
```diff
@@ -1504,7 +1504,7 @@ void MonsterAttackPlayer(int i, int pnum, int hit, int minDam, int maxDam)
 			}
 		}
 	}
-	int dam = (minDam << 6) + GenerateRnd((maxDam - minDam + 1) << 6);
+	int dam = (minDam << 6) + (GenerateRnd(maxDam - minDam + 1) << 6);
```
AJenbo, line 1507:
```suggestion
	int dam = (minDam + (GenerateRnd(maxDam - minDam + 1)) << 6;
```

Source/missiles.cpp
```diff
@@ -1177,7 +1177,7 @@ bool PlayerMHit(int pnum, Monster *monster, int dist, int mind, int maxd, missil
 		dam = player._pHitPoints / 3;
 	} else {
 		if (!shift) {
-			dam = (mind << 6) + GenerateRnd((maxd - mind + 1) << 6);
+			dam = (mind << 6) + (GenerateRnd(maxd - mind + 1) << 6);
```
AJenbo, line 1180:
```suggestion
			dam = (mind + (GenerateRnd(maxd - mind + 1)) << 6;

```

ephphatha:
With the monster damage rolls I think having the shift applied to the argument of `GenerateRnd()` is intentional. Using Skeletons as an example, they have a damage range of 1-2. In the current code they can apply fractional damage anywhere within that range with the 6 digit fixed point numbers used for health. With this PR they will always do either exactly 1 damage or exactly 2 damage, nothing in between.

Source/monster.cpp
ephphatha:
```suggestion
	int dam = (minDam << 6) + GenerateRnd(((maxDam - minDam) << 6) + 1);
```
is probably more in line with the original intent, this allows a monster to deal up to max damage without the extra fractional component

Source/missiles.cpp
```diff
@@ -3093,7 +3093,8 @@ void MI_Fireball(Missile &missile)
 			AddUnLight(missile._mlid);
 		}
 	} else {
-		int dam = missile._midam;
+		auto &monster = Monsters[id];
+		int dam = (missile._micaster == TARGET_MONSTERS) ? missile._midam : monster.mMinDamage + GenerateRnd(monster.mMaxDamage - monster.mMinDamage + 1);
 		MoveMissileAndCheckMissileCol(missile, dam, dam, true, false);
```
ephphatha, line 3098:
```suggestion
		int minDam = missile._midam;
		int maxDam = missile._midam;
		
		if (missile._micaster != TARGET_MONSTERS) {
			minDam = monster.mMinDamage;
			maxDam = monster.mMaxDamage;
		}
		MoveMissileAndCheckMissileCol(missile, minDam, maxDam, true, false);
```

This lets the damage roll be applied in PlayerMHit, letting the duplicate code be removed from here.

AJenbo:
> With the monster damage rolls I think having the shift applied to the argument of `GenerateRnd()` is intentional. Using Skeletons as an example, they have a damage range of 1-2. In the current code they can apply fractional damage anywhere within that range with the 6 digit fixed point numbers used for health. With this PR they will always do either exactly 1 damage or exactly 2 damage, nothing in between.

That explains why there are two shift operations

Source/missiles.cpp
StephenCWills: Unless you also want the variables renamed here, this looks to me like it should be resolved.

AJenbo:
Line 3112: `CheckMissileCol(missile, minDam, maxDam, false, m + shift, true);`

Source/missiles.cpp
StephenCWills: Hmm, apparently it reuses the same damage roll for the splash damage. I guess it shouldn't make much of a difference to just have it reroll damage in each tile.

AJenbo (dismissed):

Source/missiles.cpp
```diff
@@ -3093,16 +3093,23 @@ void MI_Fireball(Missile &missile)
 			AddUnLight(missile._mlid);
 		}
 	} else {
-		int dam = missile._midam;
-		MoveMissileAndCheckMissileCol(missile, dam, dam, true, false);
+		auto &monster = Monsters[id];
+		int minDam = missile._midam;
+		int maxDam = missile._midam;
+
+		if (missile._micaster != TARGET_MONSTERS) {
```
AJenbo, line 3100:
```suggestion
		int minDam = missile._midam;
		int maxDam = missile._midam;

		if (missile._micaster != TARGET_MONSTERS) {
			auto &monster = Monsters[id];
```

AJenbo (approved):
So many co authors on this one :o

StephenCWills:
I feel like I'm getting all the credit for everyone else's work.

PR #3922, opened 2022-01-07
Split loading of MPQs to handle translations and voice packs

### LINKED ISSUE #3919: Language audio patches not loaded
**Important information**
master branch compiled on ubuntu 20.04 (x86_64)

**Describe**
I have used my own patch_rt.mpq with Swedish audio in the past. Recently, the mechanism to load patch_rt.mpq was removed, in favor of language-specific patches, e.g., sv.mpq. Hence, I renamed patch_rt.mpq to sv.mpq, but archives are initialized before user options, so the language code is set to the emtpy string and a file called ".mpq" is loaded instead.

**To Reproduce**
1. launch with --verbose flag
2. Note the log message "VERBOSE: Missing: .mpq"

**Expected behavior**
Language patches are loaded (and unloaded) when the language option changes (either during init, or when a user changes the language from the Settings menu)

**Additional context**
Quick workaround: name the file ".mpq" instead of "patch_rt.mpq"

ephphatha:
Remember when I said there's no dependency from the mpq loading code on options? I was blind 😂 

fixes #3919

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1629,10 +1629,17 @@ int DiabloMain(int argc, char **argv)
 	DiabloParseFlags(argc, argv);
 	InitKeymapActions();
 
-	init_archives();
+	// Need to ensure devilutionx.mpq (and fonts.mpq if available) are loaded before attempting to read translation settings
+	LoadCoreArchives();
 	was_archives_init = true;
 
+	// Read settings including translation next. This code will use the presence of fonts.mpq and look for assets in devx.mpq
 	LoadOptions();
+	// Then look for a voice pack file based on the selected translation
+	LoadLanguageArchive();
+
+	// Finally load game data
+	LoadGameArchives();
 
 	DiabloInit();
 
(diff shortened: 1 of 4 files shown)


Source/diablo.cpp
```diff
@@ -1629,10 +1629,17 @@ int DiabloMain(int argc, char **argv)
 	DiabloParseFlags(argc, argv);
 	InitKeymapActions();
 
-	init_archives();
+	// Need to ensure devilutionx.mpq (and fonts.mpq if available) are loaded before attempting to read translation settings
+	LoadCoreArchives();
 	was_archives_init = true;
 
+	// Read settings including translation next. This code will use the presence of fonts.mpq and look for assets in devx.mpq
```
AJenbo, line 1636:
```suggestion
	// Read settings including translation next. This will use the presence of fonts.mpq and look for assets in devilutionx.mpq
```

AJenbo (approved):

PR #3924, opened 2022-01-08
Dedupe code which determines how much life/mana to restore for potions

### LINKED ISSUE #3923: Rejuvenation potions heal Monk/Bard ~2/3 as much as mana/health potions
**Important information**
Win 11 / DevilutionX master / compiled

**Describe**
In items.cpp UseItem there's some code which determines how much health (or mana) is restored when the player uses one of the partial heal potions. Depending on the class after the random heal amount is determined it gets scaled up by either 100% (warrior, barb), 50% (rogue, monk, bard), or 0% (sorc).
https://github.com/diasurgical/devilutionX/blob/074f191be80db73b1158dd137f7bee4c767e48ce/Source/items.cpp#L4066-L4077

Rejuvenation potions use almost identical code however they don't include Monk/Bard when checking if the heal amount should be boosted by 50%.
https://github.com/diasurgical/devilutionX/blob/074f191be80db73b1158dd137f7bee4c767e48ce/Source/items.cpp#L4125-L4146

**Expected behavior**
The same logic should be used for both potion types, rejuvenation potions should restore the same amount of life/mana as if the player had used one of each potion.

ephphatha:
This is very similar to the code used when casting heal other, except for how the amount to heal is chosen. Could potentially reuse it there as well with a bit of work.

fixes #3923

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4064,39 +4064,27 @@ void UseItem(int pnum, item_misc_id mid, spell_id spl)
 
 	switch (mid) {
 	case IMISC_HEAL:
-	case IMISC_FOOD: {
-		int j = player._pMaxHP >> 8;
-		int l = ((j / 2) + GenerateRnd(j)) << 6;
-		if (IsAnyOf(player._pClass, HeroClass::Warrior, HeroClass::Barbarian))
-			l *= 2;
-		if (IsAnyOf(player._pClass, HeroClass::Rogue, HeroClass::Monk, HeroClass::Bard))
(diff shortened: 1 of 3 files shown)


AJenbo (approved):

julealgon:
Just wanted to mention that I initially thought this was a pure refactoring change (as per title/commit), however this introduces the behavior change indicated in the issue with rejuvenation potions.

Would be nice if in the future these kinds of changes were made at least using 2 commits: one for the fix, and another for unifying the logic.

PR #3926, opened 2022-01-08
Update Italian localization

M0Rf30:
As in the title.
Regards


AJenbo:
Nice, I was just wondering how long until you would react when I updated the translations this morning :D

PR #3928, opened 2022-01-08
Fix character level up to 50

kphoenix137:
The character does not receive life, mana, and skill points when leveling up to 50 because the NextPlrLevel function is skipped upon reached the needed experience to level up to 50. This change allows the player to receive the benefits of leveling up, but still prevents the player from leveling up to 51.

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -530,6 +530,14 @@ std::string DebugCmdLevelUp(const string_view parameter)
 	int levels = std::max(1, atoi(parameter.data()));
 	for (int i = 0; i < levels; i++)
 		NetSendCmd(true, CMD_CHEAT_EXPERIENCE);
+	return "Faster than FireIceTalon!";
+}
+
+std::string DebugCmdGiveExperience(const string_view parameter)
+{
+	int experience = std::max(1, atoi(parameter.data()));
+	Player &myPlayer = *MyPlayer;
+	AddPlrExperience(myPlayer, myPlayer._pLevel, experience);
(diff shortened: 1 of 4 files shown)


julealgon:
A bit surprised this doesn't link to any vanilla bug issue. Do we have this behavior documented anywhere?

kphoenix137:
> A bit surprised this doesn't link to any vanilla bug issue. Do we have this behavior documented anywhere?
Unsure if this behavior was intended or not, but this is vanilla logic. It appears as if this logic was implemented to force players back to level 50 when they attempted to cheat their character level to anything above 50. Unfortunately, they implemented it in such a way where it prevents players from receiving any of the level up benefits upon reaching level 50.

AJenbo:
I belive this issue might also be addressed in this PR. It's helpful to have it rebased an reperated in to it's own :+1: 

https://github.com/diasurgical/devilutionX/pull/1626

ephphatha:
> A bit surprised this doesn't link to any vanilla bug issue. Do we have this behavior documented anywhere?

This is mentioned in the [DSF Buglist thread](http://www.lurkerlounge.com/forums/thread-12422.html)

> LEVEL 50 CROSSOVER PROBLEMS
>
> Any partners in a game where a player levels up to 50 will still see the player as level 49 in that game.
>
> A character does not receive a hit point or mana point bonus on leveling to 50.

ikonomov:
https://github.com/diasurgical/devilutionX/pull/5782 seems a bit cleaner to me.

PR #3931, opened 2022-01-09
Prevent items from breaking

kphoenix137:
Prevents items from being deleted when they hit 0 durability. Keeps the item equipped, but turns it red and removes bonuses. Unable to be re-equipped until it is repaired. Item will not lose durability if already at 0 durability and the player continues to take damage. Toggleable in the settings menu/ini file. Logic for turning items red persists if toggled off in case the player already has a 0 durability item in their inventory when turning this feature off, or if another player drops such an item in a multiplayer game and the player with the option toggled off picks it up.

image

image

image

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -496,7 +496,8 @@ void CalcSelfItems(Player &player)
 
 			if (currstr >= equipment._iMinStr
 			    && currmag >= equipment._iMinMag
-			    && currdex >= equipment._iMinDex)
+			    && currdex >= equipment._iMinDex
+			    && ((equipment._iClass == ICLASS_MISC || equipment._iClass == ICLASS_GOLD || equipment._iClass == ICLASS_QUEST) || equipment._iDurability != 0))
 				continue;
 
 			changeflag = true;
@@ -512,7 +513,7 @@ void CalcSelfItems(Player &player)
 
 void CalcPlrItemMin(Player &player)
 {
-	for (Item &item : InventoryAndBeltPlayerItemsRange { player }) {
+	for (Item &item : PlayerItemsRange { player }) {
 		item._iStatFlag = player.CanUseItem(item);
 	}
 }
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1697,6 +1697,8 @@ void ItemDoppel()
 
 void RepairItem(Item &item, int lvl)
 {
+	auto &myPlayer = Players[MyPlayerId];
+
 	if (item._iDurability == item._iMaxDur) {
 		return;
 	}
@@ -1717,6 +1719,7 @@ void RepairItem(Item &item, int lvl)
 	} while (rep + item._iDurability < item._iMaxDur);
 
 	item._iDurability = std::min<int>(item._iDurability + rep, item._iMaxDur);
+	CalcPlrInv(myPlayer, true)
 }
 
 void RechargeItem(Item &item, int r)
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1719,7 +1719,7 @@ void RepairItem(Item &item, int lvl)
 	} while (rep + item._iDurability < item._iMaxDur);
 
 	item._iDurability = std::min<int>(item._iDurability + rep, item._iMaxDur);
-	CalcPlrInv(myPlayer, true)
+	CalcPlrInv(myPlayer, true);
 }
 
 void RechargeItem(Item &item, int r)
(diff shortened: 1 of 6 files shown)


Source/player.h
```diff
@@ -333,7 +333,8 @@ struct Player {
 	{
 		return _pStrength >= item._iMinStr
 		    && _pMagic >= item._iMinMag
-		    && _pDexterity >= item._iMinDex;
+		    && _pDexterity >= item._iMinDex
+		    && ((item._iClass == ICLASS_MISC || item._iClass == ICLASS_GOLD || item._iClass == ICLASS_QUEST) || item._iDurability != 0);
```
ephphatha, line 337:
```suggestion
		    && (IsAnyOf(item._iClass, ICLASS_MISC, ICLASS_GOLD, ICLASS_QUEST) || item._iDurability != 0);
```

Source/stores.cpp
```diff
@@ -1658,6 +1658,7 @@ void SmithRepairItem()
 	}
 
 	myPlayer.InvList[i]._iDurability = myPlayer.InvList[i]._iMaxDur;
+	CalcPlrInv(myPlayer, true);
```
ephphatha, line 1661: I don't think this works as expected? There's an early return a few lines up if an equipped item is repaired, this would only run if a backpack item is repaired.

Source/player.cpp
```diff
@@ -787,10 +786,9 @@ bool DamageWeapon(int pnum, int durrnd)
 			return false;
 		}
 
-		player.InvBody[INVLOC_HAND_RIGHT]._iDurability--;
+		if (player.InvBody[INVLOC_HAND_RIGHT]._iDurability != 0)
+			player.InvBody[INVLOC_HAND_RIGHT]._iDurability--;
 		if (player.InvBody[INVLOC_HAND_RIGHT]._iDurability == 0) {
-			NetSendCmdDelItem(true, INVLOC_HAND_RIGHT);
-			player.InvBody[INVLOC_HAND_RIGHT]._itype = ItemType::None;
 			CalcPlrInv(player, true);
```
ephphatha, line 792: If you wanted to make this an optional feature you could restore the `DelItem` call (and setting the relevant item to None) but gate it behind an if check. Something like:
```cpp
if (durability == 0) {
    if (!sgOptions.allowZeroDurabilityItems) {
        NetSendCmdDelItem(...);
        player.InvBody[<location>]._itype = ItemType::None;
    }
    CalcPlrInv(...);
    return true
}

File: Source/items.cpp [hunk 4]
```diff
@@ -496,7 +496,8 @@ void CalcSelfItems(Player &player)
 
 			if (currstr >= equipment._iMinStr
 			    && currmag >= equipment._iMinMag
-			    && currdex >= equipment._iMinDex)
+			    && currdex >= equipment._iMinDex
+			    && ((equipment._iClass == ICLASS_MISC || equipment._iClass == ICLASS_GOLD || equipment._iClass == ICLASS_QUEST) || equipment._iDurability != 0))
```
ephphatha, line 500: This could use the `IsAnyOf` template like the check in `Player::CanUseItem`. I reckon it would be worth moving this to a method of the Item class that takes the stats to use in the comparison as arguments now that the check is more complicated.

Something like (using unsigned given stats should be non-negative, int is probably more appropriate in the current codebase 😂 ):
```cpp
Item::IsUsable(unsigned strength, unsigned magic, unsigned dexterity) const
{
	return strength >= _iMinStr
		    && magic >= _iMinMag
		    && dexterity >= _iMinDex
		    && (IsAnyOf(_iClass, ICLASS_MISC, ICLASS_GOLD, ICLASS_QUEST) || _iDurability != 0);
}
```

Lets you simplify this use to `if(equipment.IsUsable(currstr, currmag, currdex))` and Player::CanUseItem to
```cpp
Player::CanUseItem(const Item &item) const
{
	return item.IsUsable(_pStrength, _pMagic, _pDexterity);
}

**Comment by ephphatha** [comment]
I'm not 100% sure why the PM_DoGotHit test is failing either, the logic should be the same as before since the default behaviour is to break 0 durability items. It's segfaulting while trying to acquire a lock on storm_net_mutex in SNetSendMessage?

File: Source/items.cpp (hunk 4 above)
Comment (julealgon) [reply Source/items.cpp:500]: May I suggest extracting the last section further into a `Item::IsBroken` method? Then the original would read:
```cpp
Item::IsUsable(unsigned strength, unsigned magic, unsigned dexterity) const
{
	return strength >= _iMinStr
		    && magic >= _iMinMag
		    && dexterity >= _iMinDex
		    && !IsBroken();
}
```

PR #3932, opened 2022-01-09
Keymapper: Add more Actions

### LINKED ISSUE #2832: Keymapping request -> Item pickup "Alt" key
Requesting an option in "diablo.ini" for remapping the "Item pickup key" 

image

obligaron:
Fixes #2832

Notes:
- Adds Item Highlighting (Default key ALT & RCONTROL)
- Adds Screenshot (Default key PRINT)
- Adds Pause (Default key PAUSE)
- Adds Automap Toggle (Default key TAB)
- Adds Hide info Screen (Default key SPACE)
- With this PR all "normal" actions are mappable. Only menu and store Interactions, Automap movement and "ESC" is still hard coded. But imho these are fine.
- Still missing is mapping the mouse based actions. But this is another topic/issue.
- Also not implemented are modifier keys `SHIFT` or `ALT`.

Open question: currently left and right control are different keys for the keymapper, cause Item Highlighting only used right control. But the alt keys are combined, cause Item Highlighting also used to do this. Should we stay with this old logic or separate all keys or combine all?

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -404,12 +404,6 @@ bool PressSysKey(int wParam)
 
 void ReleaseKey(int vkey)
 {
-	if (vkey == DVL_VK_SNAPSHOT)
-		CaptureScreen();
-	if (vkey == DVL_VK_MENU || vkey == DVL_VK_LMENU || vkey == DVL_VK_RMENU)
-		AltPressed(false);
-	if (vkey == DVL_VK_RCONTROL)
-		ToggleItemLabelHighlight();
 	if (sgnTimeoutCurs != CURSOR_NONE || dropGoldFlag)
 		return;
 	sgOptions.Keymapper.KeyReleased(vkey);
@@ -436,9 +430,6 @@ void PressKey(int vkey)
 		return;
 	}
 
-	if (vkey == DVL_VK_MENU || vkey == DVL_VK_LMENU || vkey == DVL_VK_RMENU)
-		AltPressed(true);
-
 	if (MyPlayerIsDead) {
 		if (sgnTimeoutCurs != CURSOR_NONE) {
 			return;
@@ -463,10 +454,6 @@ void PressKey(int vkey)
 	}
 
 	if (sgnTimeoutCurs != CURSOR_NONE || dropGoldFlag) {
-		return;
-	}
-	if (vkey == DVL_VK_PAUSE) {
-		diablo_pause_game();
 		return;
 	}
 
@@ -524,20 +511,6 @@ void PressKey(int vkey)
 		if (AutomapActive && !talkflag) {
 			AutomapRight();
 		}
-	} else if (vkey == DVL_VK_TAB) {
-		DoAutoMap();
-	} else if (vkey == DVL_VK_SPACE) {
-		ClosePanels();
-		HelpFlag = false;
-		spselflag = false;
-		if (qtextflag && leveltype == DTYPE_TOWN) {
-			qtextflag = false;
-			stream_stop();
-		}
-		AutomapActive = false;
-		CancelCurrentDiabloMsg();
-		gamemenu_off();
-		doom_close();
 	}
 }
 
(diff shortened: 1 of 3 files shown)


AJenbo:
> Open question: currently left and right control are different keys for the keymapper, cause Item Highlighting only used right control. But the alt keys are combined, cause Item Highlighting also used to do this. Should we stay with this old logic or separate all keys or combine all?

which ever looks cleaner in code, if someone prefers using the right alt they now have the option to remap it

Source/diablo.cpp
```diff
@@ -1424,73 +1413,124 @@ void InitKeymapActions()
...
+	    CanPlayerTakeAction);
 	sgOptions.Keymapper.AddAction(
 	    "QuickSave",
 	    N_("Quick save"),
 	    N_("Saves the game."),
 	    DVL_VK_F2,
 	    [] { gamemenu_save_game(false); },
-	    [&]() { return !gbIsMultiplayer && !IsPlayerDead(); });
+	    nullptr,
+	    [&]() { return !gbIsMultiplayer && CanPlayerTakeAction(); });
```
AJenbo, line 1425: is there an issue with letting the player save when the game i paused?

Source/diablo.cpp
```diff
@@ -1424,73 +1413,124 @@ void InitKeymapActions()
...
-	    [&]() { return !gbIsMultiplayer && gbValidSaveFile && stextflag == STORE_NONE; });
+	    nullptr,
+	    [&]() { return !gbIsMultiplayer && gbValidSaveFile && stextflag == STORE_NONE && IsGameRunning(); });
 	sgOptions.Keymapper.AddAction(
 	    "QuitGame",
 	    N_("Quit game"),
 	    N_("Closes the game."),
 	    DVL_VK_INVALID,
-	    [] { gamemenu_quit_game(false); });
+	    CanPlayerTakeAction);
```
AJenbo, line 1439: this looks wrong?

obligaron: Copy&Past error. Sorry.
I also changed it (from master) that it also allows quieting when pause is active.

Source/diablo.cpp
obligaron: It isn't allowed in master and I didn't changed it.
But i tested it, when you save the game it gets unpaused. But else it seem to work.
Related note: you currently can't load a game when the game is paused (in master and this pr).
Thoughts?

obligaron:
> which ever looks cleaner in code, if someone prefers using the right alt they now have the option to remap it

I think it's more cleaner to distinguish between them. Take a look. 🙂 

For now I didn't include `DVL_VK_MENU`/`SDLK_MENU` cause I couldn't test it.
What keyboard key is this or is it platform specific (for example apple)?

Source/diablo.cpp
AJenbo: I think we will keep it the way you did it here. I could see wanting to save when it's paused, but I also feel that it shouldn't be possible to load, and that just seams inconsistent on my part :)

Source/diablo.cpp
AJenbo: >  I also changed it (from master) that it also allows quieting when pause is active.

Small changes like that can go both way, since most of the restrictions comes from the original didn't allow you to access the menu during this state. But that doesn't mean that shortcuts should abide by the same restrictions as they do not depend on the menu state.

AJenbo:
I belive it's this key:
image

I don't know why our comment says it's the alt key.

qndel:
https://www.howtogeek.com/425623/what-is-the-menu-key-for-and-how-to-remap-it/
`The menu key opens a context menu for your current application. It’s basically like right-clicking on your selection in the application`

AJenbo:
> https://www.howtogeek.com/425623/what-is-the-menu-key-for-and-how-to-remap-it/ `The menu key opens a context menu for your current application. It’s basically like right-clicking on your selection in the application`

No that's called the app_key in the api, and alt is called menu.

Source/diablo.cpp
```diff
@@ -1500,28 +1540,44 @@ void InitKeymapActions()
...
+	    nullptr,
+	    CanPlayerTakeAction);
 	sgOptions.Keymapper.AddAction(
 	    "IncreaseGamma",
 	    N_("Increase Gamma"),
 	    N_("Increase screen brightness."),
 	    'F',
 	    IncreaseGamma,
-	    [&]() { return !IsPlayerDead(); });
+	    CanPlayerTakeAction);
```
ephphatha, line 1565:
```suggestion
	    nullptr,
	    CanPlayerTakeAction);
```

Otherwise CanPlayerTakeAction is firing on key release?

PR #3933, opened 2022-01-09
Update bg.po

NikoVP:
Updates and fixes to the translation


AJenbo:
Nice :)

Translations/bg.po
```diff
@@ -1,11 +1,7 @@
-# Translation of DevilutionX to Bulgarian
-# Nikolay Popov <nikolai.popovz@gmail.com>, 2021.
-# @Krezzin, 2021.
-#
```
AJenbo, line 4: Any reason you removed the credits?

NikoVP: No, I overwrote the file. with a version I already have been working on, and forgot to include them. Will update it now to include them.

PR #3935, opened 2022-01-10
Work around MSVC+CMake+LTO bug

glebm:
Disables tests by default in MSVC LTO configurations. Refs #3778

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,13 @@ else()
 endif()
 set(PROJECT_VERSION_WITH_SUFFIX "${PROJECT_VERSION}$<$<CONFIG:Debug>:-${VERSION_SUFFIX}>")
 
+if(MSVC AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT DISABLE_LTO)
+  # Work around MSVC + CMake bug when LTO is enabled.
+  # See https://github.com/diasurgical/devilutionX/issues/3778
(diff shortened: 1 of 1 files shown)


ephphatha:
On my system MSVC isn't set until after the project call, using that variable this early doesn't get picked and tests are built no matter what config is chosen. I can't find documentation calling out whether this is expected behaviour...

Also release builds that include zerotier are broken due to a gcc flag being passed in to MSVC in libzt:
https://github.com/diasurgical/libzt/blob/a9974991b031fd8f84c4bcfd93f129b3938ed8c8/CMakeLists.txt#L286-L287

`-w` gets translated to `/w` to supress all warnings, but the `-wno-everything` flag is treated as `/W<int>` which fails to parse.

StephenCWills:
> Also release builds that include zerotier are broken due to a gcc flag being passed in to MSVC in libzt: https://github.com/diasurgical/libzt/blob/a9974991b031fd8f84c4bcfd93f129b3938ed8c8/CMakeLists.txt#L286-L287
> 
> `-w` gets translated to `/w` to supress all warnings, but the `-wno-everything` flag is treated as `/W<int>` which fails to parse.

😩 I thought that AppVeyor would catch something like that, but it looks like it's using `RelWithDebInfo`.

glebm:
Pushed a new fix, tested in VS2022, PTAL.

ephphatha (approved):
Looks good to me, release builds succeed and don't build tests.

PR #3946, opened 2022-01-14
Xbox One/Series support

fwannmacher:
Added support for Xbox One/Series.


CMakeSettings.json
```diff
@@ -15,7 +15,9 @@
           "value": "True",
           "type": "BOOL"
         }
-      ]
+      ],
+      "cmakeToolchain": "C:/Users/Felipe Wannmacher/source/repos/vcpkg/scripts/buildsystems/vcpkg.cmake",
```
AJenbo, line 19: This probably won't work for most users :)

fwannmacher: oops xD

tools/build_uwp_app.bat
```diff
@@ -0,0 +1,20 @@
+call VsDevCmd.bat
+
+mkdir ..\build
+cd ..\build
+
+git clone https://github.com/libsdl-org/SDL.git
+git -C SDL checkout b424665e0899769b200231ba943353a5fee1b6b6
```
AJenbo, line 7: Can we update our SDL dependency and let CMake handle this instead?
https://github.com/diasurgical/devilutionX/blob/master/3rdParty/SDL2/CMakeLists.txt#L17

fwannmacher: I have to say that I'm not an CMake expert. So I don't know because we need to use a specific project inside their repo (WinRT). So I did it this way so we can build the lib before doing the CMake stuff and then just link the lib itself.

Other than that we need some specific parameters for that WinRT project which is pretty outdated. `/p:PlatformToolset=v143;TargetPlatformVersion=10.0.22000.0;TargetPlatformMinVersion=10.0.14393.0`

CMake/Dependencies.cmake
```diff
@@ -23,7 +38,6 @@ else()
         # Aliasing local targets is not supported on CMake < 3.18, so make it global.
         set_target_properties(SDL2::SDL2-static PROPERTIES IMPORTED_GLOBAL TRUE)
       endif()
-      add_library(SDL2::SDL2 ALIAS SDL2::SDL2-static)
```
glebm, line 26: Why remove this alias? Most of the CMake changes seem to come from removing this line.

fwannmacher: because both SDL_image and SDL_audiolib have such approach:
```
if(TARGET SDL2::SDL2)
    set(SDL_LIBRARIES SDL2::SDL2)
elseif(TARGET SDL2::SDL2-static)
    set(SDL_LIBRARIES SDL2::SDL2-static)
```
without this change - keeping the alias - those libraries will try to link with the dynamic SDL2 instead of the static one.

glebm: If `SDL2::SDL2` is aliased to `SDL2::SDL2-static`, there is no dynamic SDL2 to begin with

fwannmacher: sorry, i saw the comment and i thought it was related to the the SDL2 `CMakeLists.txt` change
```
if(TARGET SDL2::SDL2-static)
  add_library(SDL2::SDL2 ALIAS SDL2-static)
endif()
```
i guess i have removed it during tests and forgot to add it back.

fwannmacher: ok, after reviewing the changes i remembered why it was removed. as it have removed the alias from the SDL2 CMakeLists i talked above the alias you are pointing is not needed anymore. and i can guarantee that there stills exits static linking as the xbox port is all static linked :)

fwannmacher: the issue was that we had both SDL2::SDL2 and SDL2::SDL2-static set up and it would be a problem for building SDL_image and SDL_audiolib as i mentioned before. so this way both static and dynamic linking should work properly.

glebm: The way SDL2 is set up when using `add_subdirectory` is that only one of the `SDL2::SDL2` and `SDL2::SDL2-static` is available. When we add `SDL2` as an alias to `SDL2-static`, it does not mean that the library will be linked dynamically.

We do have other platforms, such as Vita and Switch, that are fully statically linked.

uwp-project/src/SDL_winrt_main_NonXAML.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+	args->Handled = true;
+}
+
+void onInitialized()
+{
+	Windows::UI::Core::SystemNavigationManager::GetForCurrentView()->BackRequested += ref new Windows::Foundation::EventHandler<Windows::UI::Core::BackRequestedEventArgs^>(OnBackRequested);
+	
+	// workaround untill new config is released
+	std::string controllerMapping = ",*,a:b1,b:b0,x:b3,y:b2,back:b6,start:b7,leftstick:b8,rightstick:b9,leftshoulder:b4,rightshoulder:b5,dpup:b10,dpdown:b12,dpleft:b13,dpright:b11,leftx:a1,lefty:a0~,rightx:a3,righty:a2~,lefttrigger:a4,righttrigger:a5,platform:WinRT";
```
glebm, line 65: Why is this necessary? Doesn't SDL already handle controller mappings?

CMake/Dependencies.cmake
```diff
@@ -42,11 +59,21 @@ else()
 endif()
 
 add_library(DevilutionX::SDL INTERFACE IMPORTED GLOBAL)
-if(USE_SDL1)
-  target_link_libraries(DevilutionX::SDL INTERFACE ${SDL_LIBRARY})
-  target_compile_definitions(DevilutionX::SDL INTERFACE USE_SDL1)
+if(UWP_LIB)
```
glebm, line 62: Perhaps this can be simplified by simply not setting or unsetting `SDL2_MAIN` in case of `UWP_LIB`?

uwp-project/src/SDL_winrt_main_NonXAML.cpp
fwannmacher: The issue is that we need the controller GUID and Xbox has a bunch of different GUIDs. But it will removed after the new controller mapping is done.

tools/build_uwp_app.bat
glebm: We can probably make this work by passing an appropriate value of [`-DCMAKE_GENERATOR_TOOLSET=...`](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html#visual-studio-toolset-selection)

This would mean DevilutionX itself will also be compiled with this old toolset but that is probably OK

fwannmacher: I have tried that and I ended up with a ton off errors related to the windows runtime using wchar* and placing trying to use char* instead.

uwp-project/src/SDL_winrt_main_NonXAML.cpp
fwannmacher: And I'm doing it for swapping A - B X - Y

tools/build_uwp_app.bat
fwannmacher: like i said before i'm not a CMake expert so if you find a better way of doing that it would be great but i can only think about 2 options:
1 - keep this as it is
2 - keep a precompiled SDL2 lib in the repo

AJenbo (commented):
Looks pretty good at this point :) But I wish we didn't have to include all the images. Doesn't the Xbox have a GPU capable of resizing textures?

fwannmacher:
> Looks pretty good at this point :) But I wish we didn't have to include all the images. Doesn't the Xbox have a GPU capable of resizing textures?

They are used for the OS and store not for the game itself so i guess we can remove a good amount of them.

fwannmacher:
i removed all images but the recommended scale which is `200`.

glebm:
It's not perfect but it's a good start, and we can improve the CMake stuff after the merge. LGTM on the CMake stuff.

Please run `optipng -o9` on all the images.

uwp-project/devilutionx.vcxproj.filters
```diff
@@ -0,0 +1,393 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
glebm, line 1: Does this define which files to package with the app bundle? Ideally, in the future, we'd generate this file via CMake to not have to maintain the list of assets in 2 places.

CMake/Dependencies.cmake
```diff
@@ -3,6 +3,21 @@
 include(functions/dependency_options)
 include(functions/emscripten_system_library)
 
+if(EMSCRIPTEN)
+  emscripten_system_library("zlib" ZLIB::ZLIB USE_ZLIB=1)
+elseif(UWP_LIB)
+  add_subdirectory(3rdParty/zlib)
+  set(ZLIB_LIBRARY "${CMAKE_CURRENT_BINARY_DIR}/3dsParty/zlib")
+  get_property(ZLIB_INCLUDE_DIR TARGET ZLIB::ZLIB PROPERTY INCLUDE_DIRECTORIES)
```
glebm, line 11: This should be set in `3rdParty/zlib/CMakeLists.txt` (`set(... PARENT_SCOPE)`)

CMake/Dependencies.cmake
```diff
@@ -3,6 +3,21 @@
 include(functions/dependency_options)
 include(functions/emscripten_system_library)
 
+if(EMSCRIPTEN)
+  emscripten_system_library("zlib" ZLIB::ZLIB USE_ZLIB=1)
```
glebm, line 7: I don't think we need `zlib` on emscripten. I'm guessing you need it here because of the precompiled SDL2?

tools/build_uwp_app.bat
glebm: OK, let's keep this as is for now and look into improving it in the future

CMake/Dependencies.cmake
glebm: Fine to keep it as is for now

docs/manual/platforms/xbox-one.md
```diff
@@ -0,0 +1,40 @@
+# DevilutionX (Diablo 1) for Nintendo Switch
```
AJenbo, line 1: ms xb

docs/manual/platforms/xbox-one.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Installation
+
+1. Download and unzip [devilutionx-xbox-one.zip](https://github.com/diasurgical/devilutionX/releases/latest/download/devilutionx-xbox-one.zip).
+2. In Xbox Device Portal click on `Add` button located at `Home/My games & apps`
+3. In the first popup - called `Deploy or Install Application` choose `devilutionX.*<version>*.appx` and then click `Next`
```
AJenbo, line 7:
```suggestion
3. In the first popup - called `Deploy or Install Application` choose `devilutionX..appx` and then click `Next`
```

docs/manual/platforms/xbox-one.md
```diff
@@ -0,0 +1,40 @@
...
+4. In the second popup - called `Choose any necessary dependencies` choose `Microsoft.VCLibs.\*.appx` and then click `Start`
+5. In your PC go to the shared folder `WindowsApps` and find the devilutionX folder which should be something like `<random string>\_<version>\_random string<>`
+6. Copy `diabdat.mpq` from your CD (or GoG install folder) to it.
+7. In the Xbox press `View` on devilutionX and select `Details` and there change the type from `App` to `Game`.
```
AJenbo, line 11:
```suggestion
7. In the Xbox press `View` on DevilutionX and select `Details` and there change the type from `App` to `Game`.
```

docs/manual/platforms/xbox-one.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Usage
+
+Launch `devilutionX` from the Dev Mode home.
```
AJenbo, line 15:
```suggestion
Launch `DevilutionX` from the Dev Mode home.
```

AJenbo:
Do you think its possible to run the build on appveyor?

uwp-project/devilutionx.vcxproj.filters
fwannmacher: `.filter` only organize them when the project is opened on visual studio. the actual list of files is in the `.vcxproj` file

CMake/Dependencies.cmake
fwannmacher: it was already present in the CMake file and i just moved to the beginning. and i need the zlib as there are no package managers for the build as vcpkg was causing incompatibilities with its packages

fwannmacher:
> Do you think its possible to run the build on appveyor?

i do think so https://www.appveyor.com/docs/windows-images-software/#visual-studio-2022

AJenbo:
If you can figure that out it would be much appreciated :) You can see our existing script here: https://github.com/diasurgical/devilutionX/blob/master/appveyor.yml

uwp-project/devilutionx.vcxproj.filters
AJenbo: Was this generated in some way? Normally we generate the projects via CMake, (even for Android the file list is picked up from there).

fwannmacher: i have no idea how to do it with cmake. i'll have to check

fwannmacher:
> If you can figure that out it would be much appreciated :) You can see our existing script here: https://github.com/diasurgical/devilutionX/blob/master/appveyor.yml

i do think this is it https://github.com/diasurgical/devilutionX/pull/3946/commits/681ce04a5a3260f6f252740a04a5774de99662d1

uwp-project/devilutionx.vcxproj.filters
fwannmacher: What I suggest is doing something like the way I have changed the app version in the manifest using PowerShell. I'll do it later and push.

docs/manual/platforms/xbox-one.md
```diff
@@ -0,0 +1,40 @@
+# DevilutionX (Diablo 1) for Microsoft Xbox One/Series
+
+## Installation
+
+1. Download and unzip [devilutionx-xbox-one.zip](https://github.com/diasurgical/devilutionX/releases/latest/download/devilutionx-xbox-one.zip).
+2. In Xbox Device Portal click on `Add` button located at `Home/My games & apps`
+3. In the first popup - called `Deploy or Install Application` choose `devilutionX..appx` and then click `Next`
+4. In the second popup - called `Choose any necessary dependencies` choose `Microsoft.VCLibs..appx` and then click `Start`
```
AJenbo, line 8: Sorry, slipped up with my suggestion here :)

```suggestion
3. In the first popup - called `Deploy or Install Application` choose `devilutionX.appx` and then click `Next`
4. In the second popup - called `Choose any necessary dependencies` choose `Microsoft.VCLibs.appx` and then click `Start`
```

AJenbo:
> i do think this is it [681ce04](https://github.com/diasurgical/devilutionX/commit/681ce04a5a3260f6f252740a04a5774de99662d1)

Hm looks like I have to setup a second project for the same repo, or the script has to be woven in to the other one using a build matrix configuration. I have setup a secound project on appveyor for now, lets see if it triggers on the next update.

uwp-project/devilutionx.vcxproj.filters
AJenbo: Recent versions of Visual Studio has CMake support build in where you can simple open the CMakeLists.txt file as the project file. But you can also generate a sln file via CMake:
`cmake -G "Visual Studio 17 2022" .` Not sure if that is what you where asking :)

fwannmacher: I thought you were asking about the assets :) no it wasn't generated. I created the solution on visual studio and configured it by hand.

fwannmacher: there is a way of adding files to the project using wildcards. so i changed the `vcxproj` file for looking into `build/assets` and linking all the files there. this way any changes on assets will be reflected in the build.

uwp-project/devilutionx.vcxproj
```diff
@@ -0,0 +1,140 @@
...
+    <Image Include="Assets\WideTile.scale-200.png" />
+  </ItemGroup>
+  <ItemGroup>
+    <AppxManifest Include="Package.appxmanifest">
+      <SubType>Designer</SubType>
+    </AppxManifest>
+    <None Include="..\build\assets\*\*.*">
+      <DeploymentContent>true</DeploymentContent>
+      <Link>Assets\%(RecursiveDir)%(Filename)%(Extension)</Link>
+    </None>
```
fwannmacher, line 128: this does the trick of adding the assets

fwannmacher: TBH i don't think that there is a need for generating this file with cmake as it is pretty generic already.

uwp-project/devilutionx.vcxproj.filters
AJenbo: Looks a whole lot better now :)

AJenbo:
@glebm I think this is ready for merging, do you still have unresolved parts or can the conversations be closed?

glebm (approved):

fwannmacher:
> @fwannmacher thanks for the work with getting this up to speed +1
> 
> I noticed that the appveyor job didn't trigger, could you try and implement it as a build matrix of the regular job as a follow up PR?

TBH i'm not used with appveyor so if you know someone more used to it i think that would be better.

AJenbo:
Unfortunately not, this is the only time I have used it since I don't have windows I don't often get involved with development on the platform.

fwannmacher:
> Unfortunately not, this is the only time I have used it since I don't have windows I don't often get involved with development on the platform.

Ok... I'll try to find it out

fwannmacher:
https://github.com/diasurgical/devilutionX/pull/3959

PR #3950, opened 2022-01-15
Fix walk stuttering with Preview Frame

obligaron:
Fixes a reported issue from @FitzRoyX

> Another issue is that obligaron's prediction algo thing is causing some slight wierd abruptness when coming to a stop walking horizontally. It feels like the camera is shifting, I don't know how to explain it

In #2457 (last commit) the time the preview frame was shown was added to the animation distribution to smooth the animation better. But this resulted in a small stuttering, cause `GetOffsetForWalking`/`GetAnimationProgress` is also based on this time.
Workaround is to not add preview frame time for the walking animation. Also fixes some casting warnings.

### DIFF
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -59,17 +59,17 @@ int AnimationInfo::GetFrameToUseForRendering() const
 
 float AnimationInfo::GetAnimationProgress() const
 {
-	int ticksSinceSequenceStarted = TicksSinceSequenceStarted;
+	float ticksSinceSequenceStarted = TicksSinceSequenceStarted;
 	float tickModifier = TickModifier;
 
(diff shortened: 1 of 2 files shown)


ephphatha (approved):
Changes look reasonable, it's really hard to tell whether I'm getting the same stuttering when walking on master as what FitzRoyX reported though 😅. I noticed both master and this branch are a bit jittery when walking diagonally or north/south but this appears to be due to the way the offset is translated to screen space and isn't related to the animation preview. Walking horizontally is buttery smooth even at 1 game speed.

AJenbo (approved):

obligaron:
You are right. I also found this limitation while working on initial smooth walking (in #1987 see initial comment).
It could be fixed with a new/enhanced renderer.
But it's also not a great show-stopper.

PR #3951, opened 2022-01-15
Keymapper: Add middle/x1/x2 mouse button support

obligaron:
Adds support in the keymapper to bind and use the additional mouse buttons
- middle mouse button
- x1 button (extra button 1)
- x2 button (extra button 2)

They can be bind like any keyboard button and so be used for any action (like item highlighting).


AJenbo (approved):
Neat, I was thinking we could do gamepad in a similar fashion, but not sure if we should refactor before or after such a change.

obligaron:
I also think so. 👍 
But I don't own a controller or some of the other supported controller-based devices.
That's why I don't work on controller support. 😕

PR #3953, opened 2022-01-15
Show correct death animation

kphoenix137:
This fix revises the death animation, so that a death resulting in no equipped items being dropped will allow your death animation to match your current armor tier, instead of defaulting to light armor. If your equipped items are dropped, you will continue to use the light armor animation.


AJenbo (approved):

qndel:
Causes warrior's body to disappear

PR #3954, opened 2022-01-15
Fix death animation frames for Warrior/Barbarian

kphoenix137:
Fixes the animation frames for medium and heavy armor Warrior.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2436,6 +2436,7 @@ void SetPlrAnims(Player &player)
 	player._pSFNum = PlrGFXAnimLens[static_cast<std::size_t>(pc)][10];
 
 	auto gn = static_cast<PlayerWeaponGraphic>(player._pgfxnum & 0xF);
+	auto agn = player._pgfxnum & ~0xF;
 	if (pc == HeroClass::Warrior) {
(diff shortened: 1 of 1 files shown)


Source/player.cpp
```diff
@@ -2436,6 +2436,7 @@ void SetPlrAnims(Player &player)
 	player._pSFNum = PlrGFXAnimLens[static_cast<std::size_t>(pc)][10];
 
 	auto gn = static_cast<PlayerWeaponGraphic>(player._pgfxnum & 0xF);
+	auto armorGraphicIndex = player._pgfxnum & ~0xF;
```
AJenbo, line 2439: `int` should only be used if the type is clear to the human also.

AJenbo (approved):

julealgon:
@kphoenix137 could you elaborate what the problem was here? What was wrong with the previous frames?

kphoenix137:
> @kphoenix137 could you elaborate what the problem was here? What was wrong with the previous frames?

Warrior death animation for medium and heavy armor is 15 frames. All other death animations in the game are 20 frames. This wasn't accounted for in the PR to add death animations for medium and heavy armor for all classes

PR #3957, opened 2022-01-16
Remove missile/sprite limit

### LINKED ISSUE #768: Remove or enhance sprite limit
**Describe**
In Diablo I the sprite limit of the game is 255. This bothers me a lot since it will impact your late game casters quite a bit. I.e. if you use Chain Lightning on level 15 a lot of gaps appear in the bolts and it does less damage.
[See here](https://diablo.fandom.com/wiki/Chain_Lightning_(Diablo_I)
)
**To Reproduce**
1. Take a sorc with firewall.
2. Cast it all over the place.
3. At some point, there will be no more wall if you cast it.

**Expected behavior**
A sprite limit of 2048 would be much harder to notice in casual play.

**Additional context**
Tested on 1.01 I don't know if a fix would break the game.

obligaron:
Fixes #768

## PR

https://user-images.githubusercontent.com/25415264/149675676-175df2c0-1ab8-498a-8c4b-4cdd78f2f2ae.mp4

## Vanilla

https://user-images.githubusercontent.com/25415264/149675685-49e35c77-0ee3-451d-95a5-25603c5bb16a.mp4

</details>

Notes:
- Technically the limit is not "removed" but only enhanced from 125 to 2147483648. But for the dungeon size this is the same as unlimited? 😜 
- Change is backwards-compatible (see Example Videos). This is achieved by adding the first 125 the normal way and the rest is added to a new `addtionalMissiles` "file" in the save game. Vanilla ignores `addtionalMissiles` and only loads the first 125 missiles.
- `AddMissile` checks from 0 upwards if a missile id is free. If we think this is performance wise not a good move, please state it.
- First 3 commits are needed refactorings for this pr. If we want we can split them from this pr.
- It would be nice if this could be tested by others also. 🙂

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -35,7 +35,7 @@ namespace devilution {
 
 ControlTypes ControlMode = ControlTypes::None;
 int pcurstrig = -1;
-int pcursmissile = -1;
+Missile *pcursmissile = nullptr;
 quest_id pcursquest = Q_INVALID;
 
 /**
@@ -405,26 +405,24 @@ void FindTrigger()
 	if (pcursitem != -1 || pcursobj != -1)
 		return; // Prefer showing items/objects over triggers (use of cursm* conflicts)
 
-	for (int i = 0; i < ActiveMissileCount; i++) {
-		int mi = ActiveMissiles[i];
-		auto &missile = Missiles[mi];
+	for (auto &missile : Missiles) {
 		if (missile._mitype == MIS_TOWN || missile._mitype == MIS_RPORTAL) {
 			const int newDistance = GetDistance(missile.position.tile, 2);
 			if (newDistance == 0)
 				continue;
-			if (pcursmissile != -1 && distance < newDistance)
+			if (pcursmissile != nullptr && distance < newDistance)
 				continue;
 			const int newRotations = GetRotaryDistance(missile.position.tile);
-			if (pcursmissile != -1 && distance == newDistance && rotations < newRotations)
+			if (pcursmissile != nullptr && distance == newDistance && rotations < newRotations)
 				continue;
 			cursPosition = missile.position.tile;
-			pcursmissile = mi;
+			pcursmissile = &missile;
 			distance = newDistance;
 			rotations = newRotations;
 		}
 	}
 
-	if (pcursmissile == -1) {
+	if (pcursmissile == nullptr) {
 		for (int i = 0; i < numtrigs; i++) {
 			int tx = trigs[i].position.x;
 			int ty = trigs[i].position.y;
@@ -1443,7 +1441,7 @@ void plrctrls_after_check_curs_move()
 	pcursmonst = -1;
 	pcursitem = -1;
 	pcursobj = -1;
-	pcursmissile = -1;
+	pcursmissile = nullptr;
 	pcurstrig = -1;
 	pcursquest = Q_INVALID;
 	cursPosition = { -1, -1 };
@@ -1710,8 +1708,8 @@ void PerformSecondaryAction()
 		NetSendCmdLocParam1(true, CMD_OPOBJXY, cursPosition, pcursobj);
 	} else {
 		auto &myPlayer = Players[MyPlayerId];
-		if (pcursmissile != -1) {
-			MakePlrPath(myPlayer, Missiles[pcursmissile].position.tile, true);
+		if (pcursmissile != nullptr) {
+			MakePlrPath(myPlayer, pcursmissile->position.tile, true);
 			myPlayer.destAction = ACTION_WALK;
 		} else if (pcurstrig != -1) {
 			MakePlrPath(myPlayer, trigs[pcurstrig].position, true);
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -38,6 +38,8 @@ uint8_t giNumberOfLevels;
 
 namespace {
 
+constexpr size_t MaxMissilesForSaveGame = 125;
+
 uint8_t giNumberQuests;
 uint8_t giNumberOfSmithPremiumItems;
 
@@ -166,9 +168,9 @@ public:
 	}
 
 	template <typename T>
-	constexpr void Skip()
+	constexpr void Skip(size_t count = 1)
 	{
-		Skip(sizeof(T));
+		Skip(sizeof(T) * count);
 	}
 
 	void Skip(size_t len)
@@ -664,8 +666,9 @@ void SyncPackSize(Monster &leader)
 	}
 }
 
-void LoadMissile(LoadHelper *file, Missile &missile)
+void LoadMissile(LoadHelper *file)
 {
+	Missile missile = {};
 	missile._mitype = static_cast<missile_id>(file->NextLE<int32_t>());
 	missile.position.tile.x = file->NextLE<int32_t>();
 	missile.position.tile.y = file->NextLE<int32_t>();
@@ -712,6 +715,7 @@ void LoadMissile(LoadHelper *file, Missile &missile)
 	missile.var7 = file->NextLE<int32_t>();
 	missile.limitReached = file->NextBool32();
 	missile.lastCollisionTargetHash = 0;
+	Missiles.push_back(missile);
 }
 
 void LoadObject(LoadHelper &file, Object &object)
@@ -1810,8 +1814,8 @@ void LoadGame(bool firstflag)
 
 	ViewPosition = { viewX, viewY };
 	ActiveMonsterCount = tmpNummonsters;
-	ActiveMissileCount = tmpNummissiles;
 	ActiveObjectCount = tmpNobjects;
+
 
 	for (int &monstkill : MonsterKillCounts)
 		monstkill = file.NextBE<int32_t>();
@@ -1823,12 +1827,12 @@ void LoadGame(bool firstflag)
 			LoadMonster(&file, Monsters[ActiveMonsters[i]]);
 		for (int i = 0; i < ActiveMonsterCount; i++)
 			SyncPackSize(Monsters[ActiveMonsters[i]]);
-		for (int &missileId : ActiveMissiles)
-			missileId = file.NextLE<int8_t>();
-		for (int &missileId : AvailableMissiles)
-			missileId = file.NextLE<int8_t>();
-		for (int i = 0; i < ActiveMissileCount; i++)
-			LoadMissile(&file, Missiles[ActiveMissiles[i]]);
+		// Skip ActiveMissiles
+		file.Skip<int8_t>(MaxMissilesForSaveGame);
+		// Skip AvailableMissiles
+		file.Skip<int8_t>(MaxMissilesForSaveGame);
+		for (int i = 0; i < tmpNummissiles; i++)
+			LoadMissile(&file);
 		for (int &objectId : ActiveObjects)
 			objectId = file.NextLE<int8_t>();
 		for (int &objectId : AvailableObjects)
@@ -1990,7 +1994,7 @@ void SaveGameData()
 	file.WriteLE<uint8_t>(chrflag ? 1 : 0);
 	file.WriteBE<int32_t>(ActiveMonsterCount);
 	file.WriteBE<int32_t>(ActiveItemCount);
-	file.WriteBE<int32_t>(ActiveMissileCount);
+	file.WriteBE<int32_t>(std::min(Missiles.size(), MaxMissilesForSaveGame));
 	file.WriteBE<int32_t>(ActiveObjectCount);
 
 	for (uint8_t i = 0; i < giNumberOfLevels; i++) {
@@ -2014,12 +2018,21 @@ void SaveGameData()
 			file.WriteBE<int32_t>(monsterId);
 		for (int i = 0; i < ActiveMonsterCount; i++)
 			SaveMonster(&file, Monsters[ActiveMonsters[i]]);
-		for (int missileId : ActiveMissiles)
-			file.WriteLE<int8_t>(missileId);
-		for (int missileId : AvailableMissiles)
-			file.WriteLE<int8_t>(missileId);
-		for (int i = 0; i < ActiveMissileCount; i++)
-			SaveMissile(&file, Missiles[ActiveMissiles[i]]);
+		// Write ActiveMissiles
+		for (int8_t activeMissile = 0; activeMissile < MaxMissilesForSaveGame; activeMissile++)
+			file.WriteLE<int8_t>(activeMissile);
+		// Write AvailableMissiles
+		for (size_t avaiableMissile = Missiles.size(); avaiableMissile < MaxMissilesForSaveGame; avaiableMissile++)
+			file.WriteLE<int8_t>(avaiableMissile);
+		file.Skip<int8_t>(std::min(Missiles.size(), MaxMissilesForSaveGame));
+		// Write Missile Data
+		size_t wroteMissiles = 0;
+		for (auto &missile : Missiles) {
+			SaveMissile(&file, missile);
+			wroteMissiles += 1;
+			if (wroteMissiles == MaxMissilesForSaveGame)
+				break;
+		}
 		for (int objectId : ActiveObjects)
 			file.WriteLE<int8_t>(objectId);
 		for (int objectId : AvailableObjects)
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -1532,6 +1532,46 @@ void SaveDroppedItemLocations(SaveHelper &file, const std::unordered_map<uint8_t
 	}
 }
 
+constexpr uint32_t VersionAdditionalMissiles = 0;
+
+void SaveAdditionalMissiles()
+{
+	constexpr size_t BytesWrittenBySaveMissile = 180;
+	size_t missileCountAdditional = (Missiles.size() > MaxMissilesForSaveGame) ? Missiles.size() - MaxMissilesForSaveGame : 0;
+	SaveHelper file("additionalMissiles", sizeof(uint32_t) + sizeof(int32_t) + (missileCountAdditional * BytesWrittenBySaveMissile));
+
+	file.WriteLE<uint32_t>(VersionAdditionalMissiles);
+	file.WriteLE<uint32_t>(missileCountAdditional);
+
+	size_t wroteMissiles = 0;
+	for (auto &missile : Missiles) {
+		wroteMissiles += 1;
+		if (wroteMissiles >= MaxMissilesForSaveGame) {
+			SaveMissile(&file, missile);
+		}
+	}
+}
+
+void LoadAdditionalMissiles()
+{
+	LoadHelper file("additionalMissiles");
+
+	if (!file.IsValid()) {
+		// no addtional Missiles saved
+		return;
+	}
+
+	auto loadedVersion = file.NextLE<uint32_t>();
+	if (loadedVersion > VersionAdditionalMissiles) {
+		// unknown version
+		return;
+	}
+	uint32_t missileCountAdditional = file.NextLE<uint32_t>();
+	for (int i = 0; i < missileCountAdditional; i++) {
+		LoadMissile(&file);
+	}
+}
+
 const int DiabloItemSaveSize = 368;
 const int HellfireItemSaveSize = 372;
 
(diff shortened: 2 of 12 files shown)


Source/missiles.cpp
```diff
@@ -1266,17 +1212,7 @@ void InitMissiles()
...
-	for (int i = 0; i < MAXMISSILES; i++) {
-		AvailableMissiles[i] = i;
-		ActiveMissiles[i] = 0;
-	}
-	numchains = 0;
-	for (auto &link : chain) {
-		link.idx = -1;
-		link._mitype = MIS_ARROW;
-		link._mirange = 0;
-	}
```
AJenbo, line 1279: Where chains even utilized anywhere?

Source/missiles.cpp
```diff
@@ -2766,17 +2691,20 @@ void AddDiabApoca(Missile &missile, Point /*dst*/, Direction /*midir*/)
...
-	if (ActiveMissileCount >= MAXMISSILES - 1)
-		return -1;
+	constexpr int32_t MaxMissiles = std::numeric_limits<int32_t>::max();
 
-	int mi = AvailableMissiles[0];
-	auto &missile = Missiles[mi];
+	int32_t mi = 0;
+	while (true) {
+		if (Missiles.find(mi) == Missiles.end())
+			break;
```
AJenbo, line 2701: Isn't this the same?
```suggestion
	while (Missiles.find(mi) != Missiles.end())
```

AJenbo: We could probably get rid of this loop if we track the lowest available number (increase by 1 each time `.insert()` is called and `global = min(key, global)` each time `.erase()` is called).

Source/loadsave.cpp
```diff
@@ -2010,16 +2068,35 @@ void SaveGameData()
...
+		size_t wroteMissiles = 0;
+		for (auto &missileWithId : Missiles) {
+			if (missileWithId.first < MaxMissilesForSaveGame) {
+				file.WriteLE<int8_t>(missileWithId.first);
+				wroteMissiles += 1;
+			}
+		}
+		for (; wroteMissiles < MaxMissilesForSaveGame; wroteMissiles++) {
+			file.WriteLE<int8_t>(0);
+		}
```
AJenbo, line 2093:
```suggestion
		file.Skip<int8_t>(MaxMissilesForSaveGame - wroteMissiles);
```

Source/loadsave.cpp
```diff
@@ -1528,6 +1530,45 @@ void SaveDroppedItemLocations(SaveHelper &file, const std::unordered_map<uint8_t
 	}
 }
 
+constexpr uint32_t VersionAdditionalMissiles = 0;
+
+void SaveAdditionalMissiles(int missileCountAdditional)
+{
+	SaveHelper file("additionalMissiles", sizeof(uint32_t) + sizeof(int32_t) + (missileCountAdditional * 180));
```
AJenbo, line 1537: 180?

Source/loadsave.cpp
```diff
@@ -1528,6 +1530,45 @@ void SaveDroppedItemLocations(SaveHelper &file, const std::unordered_map<uint8_t
 	}
 }
 
+constexpr uint32_t VersionAdditionalMissiles = 0;
+
+void SaveAdditionalMissiles(int missileCountAdditional)
```
AJenbo, line 1535: What's the reason you went with only saving additional missiles to the expansion file?

Source/missiles.h
```diff
@@ -135,10 +128,7 @@ struct Missile {
 	int16_t lastCollisionTargetHash;
 };
 
-extern Missile Missiles[MAXMISSILES];
-extern int AvailableMissiles[MAXMISSILES];
-extern int ActiveMissiles[MAXMISSILES];
-extern int ActiveMissileCount;
+extern std::unordered_map<int32_t, Missile> Missiles;
```
AJenbo, line 131: Should we maybe `reserve()` 125 to avoid rehashing during runtime for any normal range? @glebm might prefer it one way or the other depending on what hardware he targets :)

Source/controls/plrctrls.cpp
```diff
@@ -410,7 +409,7 @@ void FindTrigger()
 			if (pcursmissile != -1 && distance == newDistance && rotations < newRotations)
 				continue;
 			cursPosition = missile.position.tile;
-			pcursmissile = mi;
+			pcursmissile = missileWithId.first;
```
ephphatha, line 412: `pcursmissile` could be a pointer or an iterator into the container to save needing to reference the key here (and remove the need to do a lookup when the value is actually used).

Source/loadsave.cpp
```diff
@@ -1810,25 +1851,29 @@ void LoadGame(bool firstflag)
...
 	ViewPosition = { viewX, viewY };
 	ActiveMonsterCount = tmpNummonsters;
-	ActiveMissileCount = tmpNummissiles;
 	ActiveObjectCount = tmpNobjects;
 
 	for (int &monstkill : MonsterKillCounts)
 		monstkill = file.NextBE<int32_t>();
 
 	if (leveltype != DTYPE_TOWN) {
+		std::vector<int8_t> tmpActiveMissiles(MaxMissilesForSaveGame);
```
ephphatha, line 1860: This can be a `std::array`, it's never resized. Strictly speaking I don't think it's needed at all though. Missile IDs are very rarely used, if you clean up the few points where ID comparison/lookup is performed you can skip this completely and just insert from the start of the `Missiles` container.

Source/loadsave.cpp
```diff
@@ -1990,7 +2038,17 @@ void SaveGameData()
...
-	file.WriteBE<int32_t>(ActiveMissileCount);
+	int missileCountOldFormat = 0;
+	int missileCountAdditional = 0;
+	if (leveltype != DTYPE_TOWN) {
+		for (auto &missileWithId : Missiles) {
+			if (missileWithId.first < MaxMissilesForSaveGame)
+				missileCountOldFormat += 1;
+			else
+				missileCountAdditional += 1;
+		}
```
ephphatha, line 2049: This potentially leaves the vanilla save file with far less than 125 active missiles. For example if 150 get spawned at once and the 120 oldest are deleted (leaving 30 active with IDs 121-150), the vanilla save file contains 5 active missiles, the additional file contains 25.

If you handle this like we currently do with `Objects` you can set up the vanilla save file in a way that active missiles start from index 0 and go in ascending order. Then you can just write the first 125 missiles here and whatever overflow there is in the additional file.

Also there's a potential bug with town portal missiles and the corresponding light source. I think it'll end up with an orphaned light source if the corresponding missile gets written to the additional save file, but lights are buggy in general 😅

Source/missiles.cpp
```diff
@@ -2804,7 +2732,9 @@ int AddMissile(Point src, Point dst, Direction midir, missile_id mitype, mienemy
 		PlaySfxLoc(missileData.mlSFX, missile.position.start);
 	}
 
+	pParentMissile = pParent;
 	missileData.mAddProc(missile, dst, midir);
+	pParentMissile = nullptr;
```
ephphatha, line 2737: I'm personally not a fan of introducing a global for what should be a function parameter. The `AddProc` signatures have changed a few times already so I think it'd be fine to add a new param back in. Could add an `if (missileData.mAddProc == AddMisexp)` or similar check and pulling out the logic which refers to `parent` if you really wanted to avoid changing the function signatures for all the `AddProc`s, wouldn't be my preferred approach though.

Source/portal.cpp
```diff
@@ -113,16 +113,15 @@ bool PortalOnLevel(int i)
...
+	for (auto &missileWithId : Missiles) {
+		auto &missile = missileWithId.second;
 		if (missile._mitype == MIS_TOWN && missile._misource == id) {
 			dFlags[missile.position.tile.x][missile.position.tile.y] &= ~DungeonFlag::Missile;
 
 			if (Portals[id].level != 0)
 				AddUnLight(missile._mlid);
 
-			DeleteMissile(i);
+			DeleteMissile(missileWithId.first);
```
ephphatha, line 124: This is the only use of `DeleteMissile`. This should use `Missiles.erase(iterator)` directly since it's looping over the collection.

Source/missiles.cpp
```diff
@@ -2804,7 +2732,9 @@ int AddMissile(Point src, Point dst, Direction midir, missile_id mitype, mienemy
 		PlaySfxLoc(missileData.mlSFX, missile.position.start);
 	}
 
+	pParentMissile = pParent;
 	missileData.mAddProc(missile, dst, midir);
+	pParentMissile = nullptr;
 
 	return mi;
```
ephphatha, line 2739: The only time the returned index is actually used in a lookup is in portal.cpp `AddWarpMissile`. If this function returns a pointer to the missile object you remove the need to store an ID with the missile and can remove the whole loop to find the first available ID. Can also change the data structure to a vector/linked list instead of a map.

Source/missiles.cpp
```diff
@@ -4194,11 +4120,11 @@ void MI_Rportal(Missile &missile)
...
-		if (Missiles[ActiveMissiles[i]]._miDelFlag) {
-			DeleteMissile(i);
+	for (auto it = Missiles.begin(); it != Missiles.end();) {
+		if (it->second._miDelFlag) {
+			it = Missiles.erase(it);
 		} else {
-			i++;
+			++it;
 		}
 	}
```
ephphatha, line 4129: This could use `std::remove_if(Missiles.begin(), Missiles.end(), [](Missile &it) { return it._miDelFlag });` (change the lambda as needed based on the container type 😄 )

Source/portal.cpp
ephphatha: Also just realised after looking at the above, this is another case where `std::remove_if` could be used. The current code doesn't return early so it matches the current logic and means you don't need to manage the iterator manually.

Source/missiles.cpp
obligaron: I changed the loop. 🙂 

For getting rid of it. I think we can remember where we can start the loop. But don't get rid of it.
Example:
Missiles 1,2,3 exists.
Missile 1 gets deleted.
global = 1

Now we insert two Missiles
Missile 1
and after that the next would be Missile 4.
But to get the index 4 we need the loop.

Source/loadsave.cpp
obligaron: I needed to adjust `SaveHelper::Skip` (similar to `LoadHelper`), cause previous it didn't have a count parameter.

Source/loadsave.cpp
obligaron: `SaveMissile` writes 180 bytes 🙂

Source/missiles.h
obligaron: Would be possible. 🙂

Source/loadsave.cpp
obligaron: Cause this simplifies loading. 😉 
We don't need to check if the new missiles are present and ignore the original missiles if they exists.

Also we don't write the same content twice.

Source/missiles.cpp
obligaron: `std::remove_if` [doesn't work](https://stackoverflow.com/questions/9210014/remove-elements-from-an-unordered-map-fulfilling-a-predicate) with `unordered_map`. 😢
And I'm not sure if [`erase_if`](https://en.cppreference.com/w/cpp/container/unordered_map/erase_if) is supported on all platforms (c++ 20 needed).

Source/loadsave.cpp
obligaron: I used a `std::vector` so it's allocated on heap.
Cause some platforms have small stack size (like 3ds, see #3540).
But perhaps this is premature? 🤔

We could remove `tmpActiveMissiles`, if we don't want to have a stable id anymore.

Source/missiles.cpp
obligaron: I agree, but just didn't have enough motivation to update all `Add` Methods.
If you want I can go the extra mile.

Source/missiles.cpp
obligaron: For devilutionX it would work. I was surprised that the missile id's aren't stored in other variables anymore. And we got rid of `dMissiles`.

But I hesitate to do this change:
- We would store pointers to `std::vector`. But these pointers are not stable if the vector is resized (elements removed or capacity is insufficient). For example your purposed `pcursmissile`. It's more fragile.
- Mod's can't use the id to store in other variables. For example if you want to add a mod feature to teleport to the last fired missile of a player.

On the other side we could simplify some logic.

Source/loadsave.cpp
StephenCWills: It looks like this is just 125 bytes. As long as it's under 32 KB and does no recursion, it should be fine.

Source/loadsave.cpp
obligaron: You are right if we don't want to support stable id's (see [discussion](https://github.com/diasurgical/devilutionX/pull/3957#discussion_r786291804)).
If we want to support stable id's this is one downside of it. 😉

Source/portal.cpp
obligaron: I inlined `Missiles.erase` and made a `break`/early return if we found a town portal for the user.
Cause there is only one possible town portal per player.

Source/loadsave.cpp
obligaron: I changed it to normal stack arrays. 🙂

Source/missiles.cpp
ephphatha: You'd need to use `std:::list` or `std::forward_list` for an STL container that doesn't invalidate iterators on resize/removal. The old array backed structure had the advantage of never moving a missile object once it had been allocated a space in the array (though it did change the order elements were accessed from the ActiveMissiles indirect lookup).

As for mods, if using a data structure where iterators aren't invalidated when adding/removing elements to the container they can still use a stored iterator at runtime. Saving and loading needs to translate from the iterator to the index using `std:::distance` but this is feasible with a bit of code isolated in the save/load functions.

Source/missiles.cpp
ephphatha: Didn't realise the implementation of `std::remove_if` uses move/resize 😓 . This would require a different datatype to make use of that helper. The two list types provide their own specialisation of `remove_if` which would be suitable if that container gets the pick.

Source/loadsave.cpp
AJenbo: Could you add it as a constant so that we know what it is :)

Source/loadsave.cpp
AJenbo: I was going to just overwrite the content loaded from the original file :)

Source/missiles.cpp
AJenbo: I don't think there is a need for preserving the IDs for potential mods. I would be more worried about how we would do a server/client based network protocol, but that is also just to fare removed to be a worry at this stage.

Source/missiles.cpp
AJenbo: Good point, it's probably just complicating things more then it's helping so lets leave it as is.

glebm:
A pointer-stable way of doing it without using maps is via an object pool. Alternatively, a vector of unique_ptrs is also a (slower) option.

Source/missiles.cpp
obligaron: I changed it to `std::list`. 😉

Source/missiles.h
obligaron: I think with `std::list` this isn't needed anymore or?

Source/controls/plrctrls.cpp
obligaron: I changed it to a pointer. 😉

Source/loadsave.cpp
obligaron: Fixed. 😉

Source/missiles.cpp
obligaron: I changed to `remove_if` on `std::list`.

Source/missiles.cpp
obligaron: I added a new `AddMissileParameter` struct that passes the parent to all `Add` Methods.
Now I'm tired. 😩

Source/loadsave.cpp
obligaron: Done. 😉

Source/loadsave.cpp
```diff
@@ -2014,12 +2063,25 @@ void SaveGameData()
...
+		file.Skip<int8_t>((MaxMissilesForSaveGame > Missiles.size()) ? Missiles.size() : MaxMissilesForSaveGame);
+		// Write Missile Data
+		size_t wroteMissiles = 0;
+		for (auto &missile : Missiles) {
+			SaveMissile(&file, missile);
+			wroteMissiles += 1;
+			if (wroteMissiles == MaxMissilesForSaveGame)
+				break;
+		}
```
ephphatha, line 2084: This could use `std::for_each_n` maybe?
```suggestion
		std::for_each_n(Missiles.begin(), std::min(Missiles.size(), MaxMissilesForSaveGame), [&file] {
			SaveMissile(&file, missile);
		})
```

Source/loadsave.cpp
```diff
@@ -2014,12 +2063,25 @@ void SaveGameData()
...
-		for (int missileId : AvailableMissiles)
-			file.WriteLE<int8_t>(missileId);
-		for (int i = 0; i < ActiveMissileCount; i++)
-			SaveMissile(&file, Missiles[ActiveMissiles[i]]);
+		// Write ActiveMissiles
+		for (size_t activeMissile = 0; activeMissile < std::min(MaxMissilesForSaveGame, Missiles.size()); activeMissile++) {
+			file.WriteLE<int8_t>(activeMissile);
+		}
+		if (MaxMissilesForSaveGame > Missiles.size())
+			file.Skip<int8_t>(MaxMissilesForSaveGame - Missiles.size());
```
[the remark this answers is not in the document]
obligaron: Changed 😉

Source/loadsave.cpp
```diff
@@ -2014,12 +2063,25 @@ void SaveGameData()
...
+		// Write AvailableMissiles
+		for (size_t avaiableMissile = Missiles.size(); avaiableMissile < MaxMissilesForSaveGame; avaiableMissile++) {
+			file.WriteLE<int8_t>(avaiableMissile);
+		}
+		file.Skip<int8_t>((MaxMissilesForSaveGame > Missiles.size()) ? Missiles.size() : MaxMissilesForSaveGame);
```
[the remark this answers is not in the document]
obligaron: I would stay with the original. I think it's more readable. 🤔

Source/loadsave.cpp
```diff
@@ -1810,25 +1853,29 @@ void LoadGame(bool firstflag)
...
+		int8_t tmpActiveMissiles[MaxMissilesForSaveGame];
+
 		for (int &monsterId : ActiveMonsters)
 			monsterId = file.NextBE<int32_t>();
 		for (int i = 0; i < ActiveMonsterCount; i++)
 			LoadMonster(&file, Monsters[ActiveMonsters[i]]);
 		for (int i = 0; i < ActiveMonsterCount; i++)
 			SyncPackSize(Monsters[ActiveMonsters[i]]);
-		for (int &missileId : ActiveMissiles)
+		for (int8_t &missileId : tmpActiveMissiles)
```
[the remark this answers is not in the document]
obligaron: Changed 🙂

Source/loadsave.cpp
obligaron: We will see if the old gcc can handle it 😁

obligaron: Sadly not 😢 
> /root/repo/Source/loadsave.cpp:2072:3: error: 'for_each_n' is not a member of 'std'

Changed back.

AJenbo: Thanks for testing :)

Source/loadsave.cpp
```diff
@@ -2014,12 +2061,21 @@ void SaveGameData()
...
-			file.WriteLE<int8_t>(missileId);
-		for (int i = 0; i < ActiveMissileCount; i++)
-			SaveMissile(&file, Missiles[ActiveMissiles[i]]);
+		// Write ActiveMissiles
+		for (int8_t activeMissile = 0; activeMissile < MaxMissilesForSaveGame; activeMissile++)
+			file.WriteLE<int8_t>(activeMissile);
+		// Write AvailableMissiles
+		for (size_t avaiableMissile = Missiles.size(); avaiableMissile < MaxMissilesForSaveGame; avaiableMissile++)
+			file.WriteLE<int8_t>(avaiableMissile);
+		file.Skip<int8_t>((MaxMissilesForSaveGame > Missiles.size()) ? Missiles.size() : MaxMissilesForSaveGame);
```
AJenbo, line 2070:
```suggestion
		file.Skip<int8_t>(std::min(Missiles.size(), MaxMissilesForSaveGame));
```

AJenbo (commented):
Thanks for taking the time to clean all of this up. The missile code looks so much better then the original at this point :)

I only noticed one line that could be improved when going over the new version.

Source/loadsave.cpp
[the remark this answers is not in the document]
AJenbo: Agree, except you can use `std::min()` for the second line (see new comment) :)

Source/loadsave.cpp
obligaron: Changed. 🙂

AJenbo (approved):

AJenbo (approved):

AJenbo:
Ugh I have to wait 5 minutes before this gets merged :(

PR #3958, opened 2022-01-17
Soulstone

kphoenix137:
This PR borrows from qndel's PR #1974. This implements a Soulstone dropped by Diablo on death. Killing Diablo no longer ends the game. Upon killing Diablo, he will bleed for a short time and an apocalpyse missile will shroud him momentarily, and he will drop a Soulstone. The player will then say, "Vengeance is mine!". If the game is Single Player, all other monsters will die as they normally do. In Multiplayer, they will remain alive.

The Soulstone can be right clicked to trigger the End Game cinematic for the player who used it. If the player attempts to use it outside of level 16, the player will say, "That won't work here". If the player attempts to use it in level 16 while Diablo is alive, the player will say, "I can't use this, yet". Attempting to use the Soulstone and failing will not remove it from the inventory.

https://www.youtube.com/watch?v=renjD2-8isM

> Interesting idea!
> 
>     * Can you provide screenshots of what the soulstone looks like?
> 
soulstonebig

>     * Is it intuitive that you can "use" it (does it have the "right click to use" hint)?
> 
image

>     * Does the soulstone go to the belt as well?
> 
Intentionally goes to the belt, since the item is right-clickable.

>     * What happens if a different player takes the soulstone and uses it (say, for example, you host a game, kill Diablo, and then someone joins the game and takes it).
> 
The other player will trigger their own end game cinematic and leave the game (of no loss or benefit to anyone, unless we somehow end up utilizing `pDiabloKillLevel`, since we no longer have battle.net). All other players will remain in the game.

>     * What happens if you take the soulstone, save the game, and then create a new game? Does it stay in your inventory? If so, what happens when you try to use it then? What happens when you get a new one from killing another instance of Diablo?
 
The player will be unable to consume the item unless they are currently in level 16 and Diablo quest has been completed in that game.

> What exactly do we hope to achieve with this change? Being able to take the items in the ground before finishing the game? Anything else?

This dungeon level is the go-to for PVPers due to the dungeon layout being ideal. Being forced out of the game with the current functionality is not ideal, since players often will leave items in town. There have been times in my experience where items have been lost because all players were in level 16, and someone killed Diablo unexpectedly.


Source/multi.cpp
```diff
@@ -181,6 +181,7 @@ void ParseTurn(int pnum, uint32_t turn)
 void PlayerLeftMsg(int pnum, bool left)
 {
 	auto &player = Players[pnum];
+	auto &myPlayer = Players[MyPlayerId];
```
yuripourre, line 184: Unused variable?

Source/items.cpp
```diff
@@ -2030,6 +2251,12 @@ void PrintItemMisc(const Item &item)
 		strcpy(tempstr, _("Right-click to view"));
 		AddPanelString(tempstr);
 	}
+	if (item._iMiscId == IMISC_SOULSTONE) {
+		strcpy(tempstr, _("Contains Diablo's essence"));
```
yuripourre, line 2255: Maybe the description could give a hint for the player to only use it at level 16. Otherwise, it could surprise players.

kphoenix137: Removing the part that kills players. Going to make the item unusable entirely unless the player is on level 16 and the Diablo quest has been completed.

yuripourre: I like that. Maybe it could play something like:
```
myPlayer.Say(HeroSpeech::ICantCastThatHere);
```

kphoenix137:
There appears to be an issue with the function that kills monsters when Diablo dies in multiplayer games, where the monsters desync between players. I'm going to remove that functionality. I also noticed that only the host player was able to pick up the Soulstone. Not sure if this is due to a multiplayer bug or not.

Source/items.cpp
kphoenix137: Selected "That won't work here" for levels other than 16 and "I can't use this, yet" for level 16 when the quest hasn't been completed. These voice lines seemed more relevant.

yuripourre: @kphoenix137 perfect!

qndel:
I'd suggest extracting the part that adds new enums into a separate PR

julealgon:
Allowing one to use the soulstone anywhere will create a disconnect with the cinematic, which shows the player next to Diablo's corpse.

At the same time, trying to restrict in code where the item can be used would be very convoluted and would be hard to explain to the player (e.g. that they can only use the item on dlvl 16).

Ideally, we should instead provide a different mechanic that already forces the player to be in the correct level. With that in mind, I suggest _we allow looting diablo's corpse explicitly_, which then would trigger the cinematic.

I'm aware we don't have animations/sprites for that, but we could, for example, use one of the towner corpses (the ones that drop gold usually, or the slain hero one) for this purpose: after the explosion of Diablo's corpse, we could add one of those bodies on the ground, with a custom text (maybe "Albrecht's body" or something) and have it be "lootable". Once the player clicks on the body, the cinematic triggers.

There would be no soulstone item, in this scenario.

Note that this would be separate from any potential item drops from Diablo, of course.

AJenbo:
> With that in mind, I suggest _we allow looting diablo's corpse explicitly_, which then would trigger the cinematic.

he doesn't have one

Chance4us:
> 

Will this beautiful thing be revised?
I assume the solution for the multiplayer problem with picking up the soul stone could be solved similar to PR #4026?

yuripourre:
@Chance4us right? I know this solution is not perfect but is better than the current behavior, I would like to have some time to explore hell properly after killing diablo.

kphoenix137:
> > 
> 
> Will this beautiful thing be revised?
> I assume the solution for the multiplayer problem with picking up the soul stone could be solved similar to PR #4026? 

There will be a new PR that doesn't add a new item. Instead we will allow a client option that allows the player to prevent the endgame sequence from occurring at all for themselves only, then they will continue gameplay.

Chance4us:
Many thanks for your effort.

PR #3960, opened 2022-01-18
msg: add documentation of cmd_id enum

mewmew:
document all the things

### DIFF
--- a/Source/msg.h
+++ b/Source/msg.h
@@ -20,63 +20,312 @@ namespace devilution {
 #define MAXMULTIQUESTS 10
 
 enum _cmd_id : uint8_t {
+	// Player mode standing.
+	//
+	// body (TCmd)
 	CMD_STAND,
+	// Walk to location.
+	//
+	// body (TCmdLoc)
 	CMD_WALKXY,
+	// Acknowledge receive of player info.
+	//
+	// body (TCmdPlrInfoHdr)
 	CMD_ACK_PLRINFO,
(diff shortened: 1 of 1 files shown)


mewmew:
Note, Hellfire related `cmd` enums are not documented.

Also, there were two enums present in Diablo v1.09b which seem to be missing from `msg.h`? Namely these:

```c
	// Cast spell from speedbook.
	//
	// body (TCmdParam1):
	//    int16_t spell_id
	CMD_SBSPELL,

	// Remove mana shield of player (invoked at end of timer; 48 ticks per clvl).
	//
	// body (TCmd)
	CMD_ENDSHIELD,
```

galaxyhaxz:
I think just about all of it is spot on. Two things to note:
`CMD_SBSPELL` is unused, the code for casting spells from the speedbook was removed after the demo but the network command and function still remain.
`CMD_REQUESTAGITEM` any command with `A` in front of the item means "auto" as in use the auto-place function for placing in the inventory, rather than cursor. :)

mewmew:
> CMD_REQUESTAGITEM any command with A in front of the item means "auto" as in use the auto-place function for placing in the inventory, rather than cursor. :)

Yeah :) I choose to use "loot" for looting an item to (and auto-placing in) inventory, and "lift" to lift an item to hand (the `HoldItem` cursor).

AJenbo:
Nice :)

PR #3961, opened 2022-01-18
Clean up window focus handlers

### LINKED ISSUE #3119: :bug: paused multiplayer game
**Describe**
Multiplayer games can end up being paused locally. This seams to happen if the game is set to a paused state before the multiplayer game is started.

**To Reproduce**
1. start a multiplayer game
2. Before the game loads switch to a different application
3. Switch back to the game after load has compleated
4. See that the game is paused...

**Expected behavior**
Multiplayer games should not be paused ever.

**Extra**
You can make the game unpause by hitting esc

StephenCWills:
* Mute cinematics when the window is not in focus
* Mute music/cinematic audio when menu state transitions while the window is not in focus
* Fix the error that causes multiplayer games to start paused
* Add SDL1 window focus logic

This resolves #3119


ephphatha (approved):
LGTM, audio still plays in multiplayer games when the window doesn't have focus which is desirable given they don't pause.

PR #3964, opened 2022-01-19
Handle runtime_errors thrown by std::locale constructor

ephphatha:
Appears some implementations don't support unknown locales and throw errors instead, causing the language detection to error and not complete.

Fixes a crash reported by Linerd in discord (by working around a potential misconfigured environment).


AJenbo (approved):
On boy or first try catch :cold_sweat:

StephenCWills:
There's a bunch of them in the net code.

AJenbo:
Most can be avoided with NONET :)

PR #3972, opened 2022-01-21
Proposed fix for compression bug

AJenbo:
@mewmew does this look correct?

I would have suspected that ASAN would have already reported an issue for us, maybe we corrected it in another way?

### DIFF
--- a/3rdParty/PKWare/pkware.h
+++ b/3rdParty/PKWare/pkware.h
@@ -72,7 +72,7 @@ typedef struct
     unsigned short phash_offs[0x2204];      // 49D0: Table of offsets for each PAIR_HASH
 } TCmpStruct;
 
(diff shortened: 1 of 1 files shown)


mewmew:
> I would have suspected that ASAN would have already reported an issue for us, maybe we corrected it in another way?

It would only be exposed by specifically crafted network packets (e.g. by an attacker), not by regular data generated by the game.

mewmew:
> @mewmew does this look correct?

It looks like the DevilX one was never broken :)

The PKWARE headers I was looking at had hardcoded values, while DevilX infers them using `sizeof` (which is the correct way).

Thus, we can close this PR.

AJenbo:
Thanks for notifying :)

PR #2965, opened 2021-09-27
Alternative Gamepad Scheme (8 buttons joystick)

yuripourre:
diablo-snes-diagram

**Combos:**

Hold L + A = Health Potion
Hold L + X = Mana Potion

Hold R + X = Primary Spell (F5) and Cast
Hold R + Y = Secondary Spell (F6)
Hold R + B = Tertiary Spell (F7)

PR #3975, opened 2022-01-21
Uwp controller schema

fwannmacher:
I've added custom controls to Xbox and I think it it would be a good practice doing this type of controls customization instead of doing it through CMake as it allows more specific changes.


AJenbo (commented):
Is it intentional that Source/controls/platforms/uwp.cpp is empty?

Individual schemes is something we want. But we should avoid having it be a compile time option. Players on Mac who uses an Xbox controller won't be able to benefit from this.

Source/controls/modifier_hints.cpp
```diff
@@ -184,21 +184,29 @@ void DrawStartModifierMenu(const Surface &out)
...
+
+#ifdef __UWP__
+	static const CircleMenuHint DPad(/*isDpad=*/true, /*top=*/_("Quests"), /*right=*/_("Inv"), /*bottom=*/_("Spells"), /*left=*/_("Char"));
+	DrawCircleMenuHint(out, DPad, { PANEL_LEFT + CircleMarginX, PANEL_TOP - CircleTop });
+#else
 	static const CircleMenuHint DPad(/*isDpad=*/true, /*top=*/_("Menu"), /*right=*/_("Inv"), /*bottom=*/_("Map"), /*left=*/_("Char"));
 	static const CircleMenuHint Buttons(/*isDpad=*/false, /*top=*/"", /*right=*/"", /*bottom=*/_("Spells"), /*left=*/_("Quests"));
 	DrawCircleMenuHint(out, DPad, { PANEL_LEFT + CircleMarginX, PANEL_TOP - CircleTop });
 	DrawCircleMenuHint(out, Buttons, { PANEL_LEFT + PANEL_WIDTH - Buttons.Width() - CircleMarginX, PANEL_TOP - CircleTop });
+#endif
```
AJenbo, line 196: The hints should look at the mapping rather then be hardcoded

fwannmacher:
> Is it intentional that Source/controls/platforms/uwp.cpp is empty?
> 
> Individual schemes is something we want. But we should avoid having it be a compile time option. Players on Mac who uses an Xbox controller won't be able to benefit from this.

no it shouldn't :-p it is fixed now

fwannmacher:
> Is it intentional that Source/controls/platforms/uwp.cpp is empty?
> 
> Individual schemes is something we want. But we should avoid having it be a compile time option. Players on Mac who uses an Xbox controller won't be able to benefit from this.

regarding making the controls available for other platforms i think we would have to do deeper change in the way controllers are mapped :thinking:

Source/controls/modifier_hints.cpp
fwannmacher: I have changed the way it is done. now game_controls.cpp or any other platform - like uwp - can create their own lists

docs/manual/platforms/xbox-one.md
```diff
@@ -1,40 +1,41 @@
...
+## Controls
+
+- Left analog or D-Pad: move hero
+- A: attack nearby enemies, talk to townspeople and merchants, pickup/place items in the inventory, OK while in main menu
+- B: select spell, back while in menus
+- X: pickup items, open nearby chests and doors, use item in the inventory
+- Y: cast spell, delete character while in main menu
+- LB: use health item from belt
+- RB: use mana potion from belt
+- LT + ←: character sheet
```
glebm, line 26: A completely different control scheme and implementation only for UWP seems rather hard to maintain.

fwannmacher: Another solution would keeping the current implementation for limited systems like vita and 3ds and call the one I'm adding "full range" and use it for everything else. So PC, Mac, Xbox and switch would use this configuration as those have access to all needed buttons

fwannmacher: https://github.com/diasurgical/devilutionX/pull/3975/commits/a0a5f39cb86d8f5467a0d6e100b2a7d0ce9df7df
:point_up: this does the trick

StephenCWills:
@fwannmacher Ideally, controller mappings should be data driven, customizations for platforms could be changed by providing default mappings for the platform, and users would be able to modify their personal mappings based on their preference. That's the direction I'd like to see the project move in, so if it were up to me, I would hesitate to accept a PR that focuses on a specific platform's controls and introduces a lot of code in the process.

Can you explain why these changes should be necessary at this time? Is it entirely because the controls are unintuitive for Xbox players, or is there more to it?

fwannmacher:
> @fwannmacher Ideally, controller mappings should be data driven, customizations for platforms could be changed by providing default mappings for the platform, and users would be able to modify their personal mappings based on their preference. That's the direction I'd like to see the project move in, so if it were up to me, I would hesitate to accept a PR that focuses on a specific platform's controls and introduces a lot of code in the process.
> 
> Can you explain why these changes should be necessary at this time? Is it entirely because the controls are unintuitive for Xbox players, or is there more to it?

Yes, it isn't intuitive. Even more when confirm and cancel are "Nintendo like" with confirming with the right face button. Other than that using start and select as modifiers is awkward when we would be able to do it with the triggers. And one more thing is that I'm adding a button for highlighting items.

I understand your point and I agree that we should have configurable controls but in other hand few console games offer that. And I don't think the changes will be discarded as they can be used as the "default" configuration.

CMake/platforms/windows.cmake
```diff
@@ -20,3 +20,5 @@ endif()
 if(MINGW_CROSS)
   list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/mingw")
 endif()
+
+set(FULL_RANGE_CONTROLLER ON)
```
StephenCWills, line 24: Your control scheme requires an LT/L2 button on the controller in order to access all in-game functions, correct? I don't think we should be making the assumption at build time that LT/L2 will be available on controllers used by Windows, Linux, and Mac players.

fwannmacher: I don't see them playing with a controller at all xD but if they do that why wouldn't they use a Xbox/PS controller? Even 8bitdo controllers have L2/R2.

StephenCWills: My understanding is that the control scheme that was developed for DevilutionX was designed specifically to support 12-button gamepads. I don't see a compelling reason to restrict platforms that support those types of controllers.

I'm especially uncompelled if your the justification is "why wouldn't they use a different controller"? It doesn't really matter why they wouldn't, but rather whether there exists a player that wouldn't. And the only answer I can give you is "maybe yes".

StephenCWills: Also potentially relevant to this discussion. #2965

fwannmacher: TBH my initial idea was to only add this controller config to the Xbox port but I was criticized but doing that. So I have added the option for doing it to any system you want through CMAKE. I also agree with the idea that we need a major change and allow customization by the user but until there we are not allowing Xbox player to have a better controller scheme. That is my opinion regarding all this discussion

fwannmacher: I have an idea for having multiple predefined control schemes. I'll push it as soon as I get it done.

AJenbo: I know @NiteKat at least has a gamepad with very few buttons connected to his computer :D

Source/controls/devices/game_controller_factory.cpp
```diff
@@ -0,0 +1,20 @@
...
+#if defined(DUAL_STICK_4_SHOULDERS_GAME_CONTROLLER)
+	return new DualStick4ShouldersGameController();
+#elif defined(OPTIONS_SET_GAME_CONTROLLER)
+	return nullptr;
+#else
+	return new DefaultGameController();
+#endif
+}
+
+} // namespace devilution
```
fwannmacher, line 20: with this it would be possible setting both predefined controllers during compile or having a list of predefined controllers which could be selected on settings.
it still needs support for SDL1 and the implementation for getting a controller from a list won't be implemented in this PR

fwannmacher:
@AJenbo @StephenCWills @glebm basically i kept the game controls the same for SDL1 but for SDL2 i have used polymorphism for handling different controllers configurations where we can have both predefined controllers set during compilation (which is useful for consoles) and the possibility of selecting controllers from a list in the settings (which is useful for PC)
There is no implementation for selecting from a list yet and TBH i'm not willing to implement it :-p so anyone can have fun and add it as the base for support it is there.
Also the new way controllers are handled makes it easier to even creating a full customizable controller.
WDYT?

fwannmacher:
No opinions?

yuripourre:
@fwannmacher tbh, I really like the idea of using polymorphism to handle controllers. I tried to do it once but didn't work well. I am not sure what others think about it but looks good to me.

AJenbo:
@fwannmacher sorry I haven't responded yet, I have been really tied up this past week. I'll try to get to it soon.

AJenbo:
> @AJenbo @StephenCWills @glebm basically i kept the game controls the same for SDL1 but for SDL2 i have used polymorphism for handling different controllers configurations where we can have both predefined controllers set during compilation (which is useful for consoles) and the possibility of selecting controllers from a list in the settings (which is useful for PC) There is no implementation for selecting from a list yet and TBH i'm not willing to implement it :-p so anyone can have fun and add it as the base for support it is there. Also the new way controllers are handled makes it easier to even creating a full customizable controller. WDYT?

This sounds good. I don't really have experience with polymorphism and since we are also close to the 1.4.0 release and I would prefer not to rock the boat I'm unfortunately going to pros pone this to 1.5.0

fwannmacher:
> This sounds good. I don't really have experience with polymorphism and since we are also close to the 1.4.0 release and I would prefer not to rock the boat I'm unfortunately going to pros pone this to 1.5.0

No problem. TBH I have some ideas which will require a refactoring of the controller handling as a whole. So I think I'll create a new branch and as soon as I get something done I'll create a new PR.

fwannmacher:
Sorry for the delay but I had no time lately. Anyways this week I may have time for getting back to this :)

glebm:
Thanks to @StephenCWills efforts, gamepad controls are now fully customizable via the config.
Might be a good time to give this another go in a data-driven way!

PR #3982, opened 2022-01-22
Missiles tidy-up/max-size check

ephphatha:
Addresses a few warnings about truncating conversions between types, unused variables, and potentially allows changing the container type without throwing as many errors.

### DIFF
--- a/Source/portal.cpp
+++ b/Source/portal.cpp
@@ -51,11 +51,13 @@ void AddWarpMissile(int i, Point position)
 {
 	MissilesData[MIS_TOWN].mlSFX = SFX_NONE;
 
-	auto &missile = AddMissile({ 0, 0 }, position, Direction::South, MIS_TOWN, TARGET_MONSTERS, i, 0, 0);
-	SetMissDir(missile, 1);
+	auto *missile = AddMissile({ 0, 0 }, position, Direction::South, MIS_TOWN, TARGET_MONSTERS, i, 0, 0);
+	if (missile != nullptr) {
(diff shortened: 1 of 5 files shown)


Source/portal.cpp
```diff
@@ -51,11 +51,13 @@ void AddWarpMissile(int i, Point position)
 {
 	MissilesData[MIS_TOWN].mlSFX = SFX_NONE;
 
-	auto &missile = AddMissile({ 0, 0 }, position, Direction::South, MIS_TOWN, TARGET_MONSTERS, i, 0, 0);
-	SetMissDir(missile, 1);
+	auto *missile = AddMissile({ 0, 0 }, position, Direction::South, MIS_TOWN, TARGET_MONSTERS, i, 0, 0);
+	if (missile != nullptr) {
```
AJenbo, line 55: There also used to be 3 checks in https://github.com/diasurgical/devilutionX/commit/234a053863d41ce79b97f325c1ec66fbc8806496#diff-d46585089d315b53630b1035442b3d490d7b422cb05ceee76cf9041de3e39179L2561

ephphatha: I added those checks back plus one more for `MonsterRangedSpecialAttack` since it has a sound effect associated with the missile.

There are other potential places where a failed missile spawn could potentially be handled in objects.cpp and quests.cpp (trap noise plays even when the trap missile doesn't get created, quest state advances if the portal fails to spawn for Archbishop Lazarus) which I haven't touched because they were already unhandled.

AJenbo (approved):

PR #3983, opened 2022-01-23
Build option to remove Quit Game (save options when changed)

### LINKED ISSUE #3940: iOS app doesn't shut down on exit
On my iPad, DevilutionX app does not close on Quit, leaving a black screen and requiring a manual close

StephenCWills:
This PR saves options closer to when they're modified and removes the Quit Game option from the 3DS, Vita, Switch, Android, and iOS.

---

Besides the settings menu, the following areas of the game can be used to modify INI settings.

**In-game Menu:**
- sgOptions.Gameplay.nTickRate
- sgOptions.Graphics.nGammaCorrection
- sgOptions.Audio.nSoundVolume
- sgOptions.Audio.nMusicVolume

**Cornerstone:**
- sgOptions.Hellfire.szItem

**Alt+Enter:**
- sgOptions.Graphics.fullscreen

**Splash:**
- sgOptions.StartUp.diabloIntro
- sgOptions.StartUp.hellfireIntro

**Game selection at startup:**
- sgOptions.StartUp.gameMode

The intent here is to add calls to `SaveOptions()` in the following places.

* At startup after displaying the splash screen
* When closing the in-game menu
* When saving the game in Single Player after the Cornerstone is used
* After changing fullscreen mode using Alt+Enter

---

This resolves #3940


StephenCWills:
I'm a little concerned about the Cornerstone, actually. I guess you could get the Cornerstone item to persist without actually saving the game by simply opening and closing the in-game menu.

ephphatha:
Bit of a corner case but if you set `[StartUp] Game=0` and leave `*Intro=0 (or 2)` then the choose game mode prompt comes up each time you launch the game until you force the settings to save via some other change. Also happens if `Intro = 1` and you force quit before the intro movie finishes playing, though in that case letting the intro movie finish or skipping it causes the settings to be written after intro gets set to 0.

(3980 has the exact opposite problem by saving the game select but not the skip intro setting 😂 )

StephenCWills:
I added an additional call to `SaveOptions()` for the game selection. I thought the one for the splash screen would catch that one too, but I was blind to all the conditionals around it.

Also, I tested Cornerstone, and it does indeed work just fine (in the context of this PR 😓).

ephphatha (dismissed):
All settings persist as far as I can tell (didn't check cornerstone but given your testing with 3984 I'm happy you've covered it :D). `NOEXIT` looks good as well, that could potentially even be exposed for desktop builds similar to NOSOUND/NONET since Alt+F4 works for quitting in testing.

Source/DiabloUI/mainmenu.cpp
```diff
@@ -21,12 +21,14 @@ void UiMainMenuSelect(int value)
 
 void MainmenuEsc()
 {
+#ifndef NOEXIT
```
AJenbo, line 24: Woudn't it make more sens to not add this to the menu rather then making it a no-op?

PR #3987, opened 2022-01-24
Sync quest items in multiplayer games

### LINKED ISSUE #1437: Some Quest Items Can Be Behave Bizarrely in Multiplayer with Multiple Players
**Important information**
Windows 10 LTSC 2019 x64, DevilutionX (v1.2.0-5F5DCIAF) with Hellfire, self-compiled, multiplayer.

**Description**
While playing with a friend, we noticed that quest items can be destroyed by the game if you pick them up with the wrong player, seemingly mostly tied to who is the host. For example, having a non-host player talk to the Farmer to spawn the Auric Amulet will drop the Auric Amulet on the ground, and if the non-host player (who does not already have an Auric Amulet) attempts to pick it up, nothing will happen for a few moments, and then suddenly the Auric Amulet will disappear from the game entirely. This is not limited to simply the Auric Amulet - for example, the Rune Bomb, Cathedral Map, and Torn Notes seemed to have the same problem.

**Expected behavior**
Ideally, all players should be able to pick these quest items up and use them.

ephphatha:
Fixes #1437

I think this is working acceptably, I was able to see quest items drop for all players in a multiplayer game no matter who triggered the item to spawn and also see the item if a player joined late. I didn't notice monster loot or chest loot getting duplicated, I was worried that might need more care to prevent. I would definitely appreciate more eyes to make sure I haven't missed anything.

Main thing was sending the DROPITEM message when quest rewards are spawned, and refactoring the Sync*Item code so it doesn't assume that a player is always dropping an item from hand.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1588,12 +1588,12 @@ bool TryDropItem()
 
 	if (currlevel == 0) {
 		if (UseItemOpensHive(myPlayer.HoldItem, myPlayer.position.tile)) {
-			NetSendCmdPItem(true, CMD_PUTITEM, { 79, 61 });
+			NetSendCmdPItem(true, CMD_PUTITEM, { 79, 61 }, myPlayer.HoldItem);
 			NewCursor(CURSOR_HAND);
 			return true;
 		}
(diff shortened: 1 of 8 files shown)


AJenbo (approved):
Nooks good, nice with the deduplication as well :)

PR #3993, opened 2022-01-25
changemp/hp commands

qndel:
Commands for manipulating hp/mana


AJenbo (approved):
I could add some comments about minor ways this is a bit odd, but it's a debug command :shrug:

PR #4003, opened 2022-01-27
Deadly Hunter's damage vs. demons

ikonomov:
I'm proposing this PR for the respect of the great people that work tirelessly on this project, as I suspect they won't be able to stomach making the change themselves.  A bug fix for forgetting to edit out the unused affix of an item.  I don't like this "fix" myself, but now that we know of the bug, there is no way to unknow it.  Although this is not the ideal solution that we might want to always have, it might still be A solution that we can allow ourselves to make in service to the game.


ikonomov:
Perhaps choosing not to implement this affix for bows was actually intentional.  Undead and Animal 50% extra damage only applies for melee weapons.  They could have easily implemented -50% damage to undead and +50% damage to animals with missiles if they chose to, but they didn't.  I think the same logic can be extended for Demons.

FitzRoyX:
I think the proper fix is to make the suffix work on missiles rather than remove it.

PR #4019, opened 2022-02-01
Enhanced attack using controllers

fwannmacher:
As discussed [here](https://github.com/diasurgical/devilutionX/discussions/4016)

For now the character won't stop moving when the attack is released as we need more opinions about it.


AJenbo:
I think this PR is ok things considered, what do you think @glebm ?

glebm:
Yeah didn't play test it but the code seems fine

glebm (approved):

PR #4034, opened 2022-02-06
Fix dungeon reroll - Initialize special Room coordinates

### LINKED ISSUE #1506: Level Reroll Bug Reproducible
**Important information**
Windows 10 / 1.2.0 / Release Copy

**Describe**
Attached save file was actually created using Vanilla 1.09, but when loaded with DevilutionX 1.2.0 (Vanilla not Hellfire), you can also get the level to reroll. Floor 5 rerolls on this save if you follow the steps below.

**To Reproduce**
1. Load the save, you will start on correct floor 5.
2. Go to floor 6.
5. Go back to floor 5, it is now rerolled.

**Expected behavior**
Level 5 should always load it's original layout.

**Screenshots**
image

**Additional context**
Screenshot is of my AI playing vanilla when it ran into this bug, you can see from the ASCII output the level of what is being seen and what the AI remembers does not make sense, and there are torches in the wall. In addition, the only way to get level 5 back to normal, that I have found, is to close the game, open it back up, then load the save. Going to floor 4 doesn't fix it, floor 6 doesn't fix it, saving and loading on floor 5 or other floors doesn't fix it. Only thing I didn't try was repeating the steps to produce the situation a second time.

Save file: 
[single_2.zip](https://github.com/diasurgical/devilutionX/files/6300806/single_2.zip)

obligaron:
Fixes #1506

`CreateRoom` [remembers](https://github.com/diasurgical/devilutionX/blob/2d6c602ef54cd84b6287a3b8af16bfc40183c761/Source/drlg_l2.cpp#L1990-L1995) the room coordinates in some variables.
This only happens if the room is [reserved for a quest](https://github.com/diasurgical/devilutionX/blob/2d6c602ef54cd84b6287a3b8af16bfc40183c761/Source/drlg_l2.cpp#L2688-L2714).
It doesn't happen if the level has no active quest.

The problem is that these variables are never reset when generating the level. They only get set when space is reserved for a quest set piece.
in issue #1506 Arkaine's Valor is not available, so level 5 has no quest set piece.
But level 6 has a quest room for Chamber of Bone.
That means level 6 sets these variables and when loading level 5 again you get another generated level, cause (in this case) [PlaceMiniSet](https://github.com/diasurgical/devilutionX/blob/2d6c602ef54cd84b6287a3b8af16bfc40183c761/Source/drlg_l2.cpp#L1698-L17009) behaves different when these variables are set.
That's also explains why the layout is also rerolled when you load the game again. But if you restart the game everything is fine (variables aren't set).

Side-note: When randomize quests is disabled this only happens on level 8, cause all other levels have a quest set pieces (zhar gets place in a normal theme room).

### DIFF
--- a/Source/drlg_l2.cpp
+++ b/Source/drlg_l2.cpp
@@ -2899,6 +2899,10 @@ void GenerateLevel(lvl_entry entry)
 	bool doneflag = false;
 	while (!doneflag) {
 		nRoomCnt = 0;
+		nSx1 = 0;
+		nSy1 = 0;
+		nSx2 = 0;
+		nSy2 = 0;
 		InitDungeonFlags();
 		DRLG_InitTrans();
 		if (!CreateDungeon()) {


qndel:
image

AJenbo (approved):

AJenbo:
Stellar work tracking this down :)

Chance4us:
A whole bunch of thank you @obligaron for fixing this weird thing.

obligaron:
> image

You don't know what I had to sacrifice for these powers... 😶 ... :stuck_out_tongue_winking_eye:

> Stellar work tracking this down :)

Thanks. The debug session was ... interesting 😁 

> A whole bunch of thank you @obligaron for fixing this weird thing.

Thanks. 🙂

galaxyhaxz:
Ahaha. I fixed this a little over a year ago, and I'm fairly sure a _certain someone_ didn't want the bugfix because they were afraid it would break dungeons in vanilla saves.

Anyway, now that the bug is fixed you can delete Blizzard's existing hack to fix it. In Patch 1.02 they added some code that tries to regenerate the dungeon if there is no set level in order to initialize the set piece coords. The problem is that their fix didn't address the situation with Arkaine on level 5.

Below is the code you can delete. I'd also recommend initializing them to `-1` instead of `0` for good measure. This is because the game will still think (0,0) is a set piece and skip it for substitutions (though in theory that tile should always be dirt anyway).

Capture

obligaron:
@galaxyhaxz Sorry, I didn't knew that there is already a bugfix for the issue. 😞

I completely missed the old bugfix in `CreateL2Dungeon`. Thanks for pointing that out. 🙂 
As you said this is code is now (in master) not needed / doesn't help.
From my understand it would still be needed for dungeon compatibility or?
I tried to address this in #4050. It would be nice if you could take a look. 🙂

galaxyhaxz:
Nah it's not your fault don't worry about it. I posted the fix on discord last year and we had a discussion about it, but the fix never ended up getting pushed.

The bugfix Blizzard did was indeed tricky to spot. It makes no sense to create the dungeon twice for the same level right? Well, some sources confirmed it was a lousy bugfix attempt lol.

It's not needed to retain compatibility. What the code does is essentially tries to initialize the set piece coords in the sequence the levels were generated but it doesn't work right. You can just delete the code, initializing them to -1 in either the main function or `GenerateLevel` like you did originally will fix all cases.

There only compatibility breakage with vanilla saves (that I know of) is this:

1) Player makes game in DevilutionX, goes down to level 6 and saves.
2) Player transfers save to vanilla and loads it.
3) Player walks up to level 5

In this case, DevilutionX will have generated level 5 properly because it initialized the set piece the first time. Vanilla however, will use the set piece coords from Bone Chamber and thus the dungeon will be different.

Now keep in mind that the vanilla bugfix already DIDNT work in many cases, so it's worth fixing IMO.

AJenbo:
Yeah, since vanilla isn't even compatible with it self here, there is no reason to try and secound guess it, there will always be a case where it's wrong.

obligaron:
I adjusted the pr 😉

galaxyhaxz:
Also if you're wondering how I discovered that code was garbage, look at the PSX source code:
https://github.com/diabpsx/skeleton/blob/master/JAP_1998_05_29/DIABPSX/SOURCE/DRLG_L2.CPP#L3562

They fixed it the same way, by initialized them to -1. That port never ceases to amaze

Who made Diablo? Gary Liddon.... XD

PR #4038, opened 2022-02-06
Clean up owned/unowned CelSprite ambiguity

glebm:
Makes `CelSprite` unowned and adds a new `OwnedCelSprite` class for owned sprites.

This clarifies ownership and makes the code cleaner in a number of places.

Additionally, because the `CelSprite` class is now tiny (1 less pointer), we can pass it by-value instead of by-reference, removing a pointer indirection in the rendering functions.

Based on top of #4031

### DIFF
--- a/Source/items.h
+++ b/Source/items.h
@@ -173,74 +173,74 @@ constexpr int ItemAnimWidth = 96;
 
 struct Item {
 	/** Randomly generated identifier */
-	int32_t _iSeed;
-	uint16_t _iCreateInfo;
-	enum ItemType _itype;
-	Point position;
-	bool _iAnimFlag;
+	int32_t _iSeed = 0;
+	uint16_t _iCreateInfo = 0;
+	enum ItemType _itype = ItemType::None;
+	Point position = { 0, 0 };
+	bool _iAnimFlag = false;
 	/*
 	 * @brief Contains Information for current Animation
 	 */
 	AnimationInfo AnimInfo;
-	bool _iDelFlag; // set when item is flagged for deletion, deprecated in 1.02
-	uint8_t _iSelFlag;
-	bool _iPostDraw;
-	bool _iIdentified;
-	item_quality _iMagical;
-	char _iName[64];
-	char _iIName[64];
-	enum item_equip_type _iLoc;
-	enum item_class _iClass;
-	uint8_t _iCurs;
-	int _ivalue;
-	int _iIvalue;
-	uint8_t _iMinDam;
-	uint8_t _iMaxDam;
-	int16_t _iAC;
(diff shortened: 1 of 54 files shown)


Source/scrollrt.cpp
```diff
@@ -852,10 +852,9 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
 			int px = targetBufferPosition.x - CalculateWidth2(pDeadGuy->width);
 			const byte *pCelBuff = pDeadGuy->data[(bDead >> 5) & 7];
 			assert(pCelBuff != nullptr);
-			const auto *frameTable = reinterpret_cast<const uint32_t *>(pCelBuff);
-			int frames = SDL_SwapLE32(frameTable[0]);
+			const uint32_t frames = LoadLE32(pCelBuff);
 			int nCel = pDeadGuy->frame;
-			if (nCel < 1 || frames > 50 || nCel > frames) {
+			if (nCel < 1 || frames > 50 || nCel > static_cast<int>(frames)) {
```
AJenbo, line 857: Any reason we can't make frame/nCel unsigned?

Source/items.h
```diff
@@ -173,74 +173,74 @@ constexpr int ItemAnimWidth = 96;
 
 struct Item {
 	/** Randomly generated identifier */
-	int32_t _iSeed;
-	uint16_t _iCreateInfo;
-	enum ItemType _itype;
-	Point position;
-	bool _iAnimFlag;
+	int32_t _iSeed = 0;
```
AJenbo, line 176: Nice with all default :+1:

AJenbo:
Alright https://github.com/diasurgical/devilutionX/pull/4031 is merged, time for this to get a rebase :)

glebm:
Rebased!

Source/scrollrt.cpp
glebm: I'm not sure, we probably could but best leave this for another PR

Source/missiles.h
```diff
@@ -97,8 +97,8 @@ struct Missile {
 	const byte *_miAnimData;
 	int _miAnimDelay; // Tick length of each frame in the current animation
 	int _miAnimLen;   // Number of frames in current animation
-	int _miAnimWidth;
-	int _miAnimWidth2;
+	uint16_t _miAnimWidth;
+	int16_t _miAnimWidth2;
```
AJenbo, line 101: Why does this need to be signed, isn't it just half of the other one? Or is it an offset?

glebm: Not entirely sure what this represents but it is negative sometimes

AJenbo (approved):

PR #4039, opened 2022-02-07
Clean up player.cpp and player.h

kphoenix137:
Reduces amount of code in SetPlrAnims function and is more easily moddable.


ikonomov:
I hardly know anything about C++ programming, but the new code seems more confusing to follow than the old one.

DakkJaniels:
I think this meets the intent of what you are trying to do, while keeping the original array in place.
https://github.com/kphoenix137/devilutionX/pull/3/

DakkJaniels:
I still think you should leave the original array as is and get rid of the `default` in your new arrays so they can make the existing `PlayerWeaponGraphics` enum.

ikonomov:
The updated version seems easier to follow now.

Source/player.cpp
```diff
@@ -2447,68 +2447,27 @@ void NewPlrAnim(Player &player, player_graphic graphic, Direction dir, int numbe
...
+		player._pSFrames = PlrGFXAnimLens[static_cast<std::size_t>(pc)][5];
+		player._pDFrames = PlrGFXAnimLens[static_cast<std::size_t>(pc)][4];
+		player._pBFrames = PlrGFXAnimLens[static_cast<std::size_t>(pc)][3];
+		player._pAFNum = PlrGFXAFNum[static_cast<std::size_t>(pc)][static_cast<std::size_t>(gn)];
 	}
-	player._pSFNum = PlrGFXAnimLens[static_cast<std::size_t>(pc)][8];
+	player._pSFNum = PlrGFXAnimLens[static_cast<std::size_t>(pc)][10];
 
-	auto gn = static_cast<PlayerWeaponGraphic>(player._pgfxnum & 0xF);
 	int armorGraphicIndex = player._pgfxnum & ~0xF;
```
DakkJaniels, line 2477: I would move this to line 2451, to be consistent.

qndel:
Nice! Could you add class names in comments too? Like 
```cpp
const int PlrGFXAFrames[enum_size<HeroClass>::value][10] = {
	// Unarmed, UnarmedShield, Sword, SwordShield, Bow, Axe, Mace, MaceShield, Staff
/*monk       */	{  16,      16,            16,    16,          16,  20,  16,   16,         16 },
/*barbarian */	{  18,      18,            18,    18,          12,  22,  18,   18,         16 },
/* dunno     */	{  20,      16,            16,    16,          20,  24,  16,   16,         16 },
	```

File: Source/player.cpp [hunk 2]
```diff
@@ -164,15 +164,44 @@ struct DirectionSettings {
 	void (*walkModeHandler)(int, const DirectionSettings &);
 };
 
+// clang-format off
```
StephenCWills, line 167: Based on [other areas of the codebase](https://github.com/diasurgical/devilutionX/blob/32d39c65246b17fbf6b89d4a914beb04f02ee7e8/Source/drlg_l1.cpp#L141-L156), these clang-format comments should be inside the array initializers. I assume this avoids the possibility of slipping formatting errors into the left-hand side of the variable definition.

For example...

```c++
/** Specifies the frame of each animation for which an action is triggered, for each player class. */
const int PlrGFXAnimLens[enum_size<HeroClass>::value][11] = {
	// clang-format off
	{ 10, 16,  8,  2, 20, 20,  6, 20,  8,  9, 14 },
	{  8, 18,  8,  4, 20, 16,  7, 20,  8, 10, 12 },
	{  8, 16,  8,  6, 20, 12,  8, 20,  8, 12,  8 },
	{  8, 16,  8,  3, 20, 18,  6, 20,  8, 12, 13 },
	{  8, 18,  8,  4, 20, 16,  7, 20,  8, 10, 12 },
	{ 10, 16,  8,  2, 20, 20,  6, 20,  8,  9, 14 },
	// clang-format on
};
```

Source/player.cpp
qndel: then you have a mess where you don't see instantly where each variable is being used.

Source/player.cpp
```diff
@@ -163,45 +163,41 @@ struct DirectionSettings {
 	PLR_MODE walkMode;
 	void (*walkModeHandler)(int, const DirectionSettings &);
 };
-
-// clang-format off
 /** Specifies the frame of each animation for which an action is triggered, for each player class. */
 const int PlrGFXAnimLens[enum_size<HeroClass>::value][11] = {
```
DakkJaniels, line 167: you should pull in the comments I added here so that people know what this stuff is. The language might need to be modified, but at least it's something.

qndel: what comments?

kphoenix137: He submitted a PR to my branch with the changes

DakkJaniels: I cleaned it up, you just have to pull it now

Source/player.h
```diff
@@ -288,14 +288,23 @@ struct Player {
 	 * @brief Contains Data (Sprites) for the different Animations
 	 */
 	std::array<PlayerAnimationData, enum_size<player_graphic>::value> AnimationData;
+	/* Nothing Frames */
```
qndel, line 291: nothing frames? ...

kphoenix137: Yes., hence NFrames

Source/player.cpp
```diff
@@ -49,53 +49,17 @@ int plrxoff2[9] = { 0, 1, 0, 1, 2, 0, 1, 2, 2 };
...
-	30,
-	20,
-	15,
-	25,
-	20,
-	40,
-};
-/** Maps from player_class to starting stat in magic. */
-int MagicTbl[enum_size<HeroClass>::value] = {
+/** Maps from player_class to starting stat. */
```
qndel, line 52: 0 info about class and stats, everything accessed just by raw index, burn it with fire 🤮 
add comments with stat / class name + maybe use some enum for accessing the array?  
0 = STAT_STR or something

qndel (changes requested):

Source/player.h
qndel: aaand what do they do?

kphoenix137: They're when the hero is doing nothing :pukaczu:

qndel: that's the shittiest name ever, that's idle frames :S

Source/player.cpp
kphoenix137: how's that?

Source/player.h
AJenbo: Neutral

kphoenix137:
Not sure why tests are failing

kphoenix137:
Superceded by #5763

PR #4045, opened 2022-02-08
Fix Diablo's HP

### LINKED ISSUE #4043: Diablo HP in Diablo
Untitled-1
As discussed earlier and confirmed by staphen the HP for Diablo is incorrectly defined in two places by -=2000 instead of /=2 which results in Diablo having less HP in Diablo.

StephenCWills:
This restores Diablo to his original HP values of 833 for single player and 1666 for multiplayer in non-Hellfire games.

This resolves #4043

---

It was discovered by @ikonomov and reported on Discord that there are a couple places in the code where Diablo's HP is arbitrarily reduced by 2000 when it seemingly should have instead been reduced by half in `InitMonsterGFX()` and `PrintMonstHistory()`. These code blocks can be traced back to f05ae4c which doesn't really provide much context about where they came from. From what I can tell, it seems like a math error so that was my assumption when making these changes.

Note that the error by itself didn't actually modify Diablo's HP, which was still hardcoded in `InitMonster()` at the time. Eventually, the hardcoded logic was replaced in #1486. The HP value was adjusted as though it was working from the full Hellfire HP value of 3333, but it was actually working from 1333, granting Diablo a measly 333.25 HP in single player and 666.5 HP in multiplayer.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -175,11 +175,12 @@ void InitMonster(Monster &monster, Direction rd, int mtype, Point position)
 	monster.AnimInfo.CurrentFrame = GenerateRnd(monster.AnimInfo.NumberOfFrames - 1) + 1;
 
 	monster.mLevel = monster.MData->mLevel;
-	monster._mmaxhp = (monster.MType->mMinHP + GenerateRnd(monster.MType->mMaxHP - monster.MType->mMinHP + 1)) << 6;
+	int maxhp = monster.MData->mMinHP + GenerateRnd(monster.MData->mMaxHP - monster.MData->mMinHP + 1);
 	if (monster.MType->mtype == MT_DIABLO && !gbIsHellfire) {
-		monster._mmaxhp /= 2;
+		maxhp /= 2;
 		monster.mLevel -= 15;
 	}
+	monster._mmaxhp = maxhp << 6;
 
 	if (!gbIsMultiplayer)
 		monster._mmaxhp = std::max(monster._mmaxhp / 2, 64);
@@ -3736,13 +3737,6 @@ void InitMonsterGFX(int monst)
 		}
 	}
 
-	monster.mMinHP = monsterData.mMinHP;
-	monster.mMaxHP = monsterData.mMaxHP;
-	if (!gbIsHellfire && mtype == MT_DIABLO) {
-		monster.mMinHP -= 2000;
-		monster.mMaxHP -= 2000;
-	}
-	monster.mAFNum = monsterData.mAFNum;
 	monster.MData = &monsterData;
 
 	if (monsterData.has_trans) {
@@ -4633,8 +4627,8 @@ void PrintMonstHistory(int mt)
 		int minHP = MonstersData[mt].mMinHP;
 		int maxHP = MonstersData[mt].mMaxHP;
 		if (!gbIsHellfire && mt == MT_DIABLO) {
-			minHP -= 2000;
-			maxHP -= 2000;
+			minHP /= 2;
+			maxHP /= 2;
 		}
 		if (!gbIsMultiplayer) {
 			minHP /= 2;
--- a/Source/monster.h
+++ b/Source/monster.h
@@ -158,9 +158,6 @@ struct CMonster {
 		return Anims[static_cast<int>(graphic)];
 	}
 	std::unique_ptr<TSnd> Snds[4][2];
-	uint16_t mMinHP;
-	uint16_t mMaxHP;
-	uint8_t mAFNum;
 	int8_t mdeadval;
 	const MonsterData *MData;
 };


AJenbo:
Diablo clearly has the 3333 HP from Hellfire in the table. During spawn his HP needs to be reduced if the expansion is not active:

https://github.com/diasurgical/devilutionX/blob/master/Source/monstdat.cpp#L130

Source/monster.cpp
```diff
@@ -3726,13 +3726,6 @@ void InitMonsterGFX(int monst)
 		LevelMonsterTypes[monst].Anims[anim].Rate = MonstersData[mtype].Rate[anim];
 	}
 
-	LevelMonsterTypes[monst].mMinHP = MonstersData[mtype].mMinHP;
```
DakkJaniels, line 3729: why is 3729, 3730 and 3735 being deleted?

Source/monster.h
```diff
@@ -157,9 +157,6 @@ struct CMonster {
 		return Anims[static_cast<int>(graphic)];
 	}
 	std::unique_ptr<TSnd> Snds[4][2];
-	uint16_t mMinHP;
```
DakkJaniels, line 160: why are these lines being deleted from the struct?

Source/monster.cpp
StephenCWills: Because all it really does is precalculate Diablo's HP while initializing monster graphics, which represents an unnecessary level of misdirection that enabled this error to exist in the first place.

Source/monster.h
StephenCWills: The HP variables were only being used in one place and were almost always just a direct copy of the HP data in the `MonstersData` array. After making this change, those variables are no longer used. `mAFNum` is simply unused so I removed it as well.

StephenCWills:
> Diablo clearly has the 3333 HP from Hellfire in the table. During spawn his HP needs to be reduced if the expansion is not active:

No one is disputing that. This PR does reduce his HP to the correct amount.

DakkJaniels:
Per Devilution and Jarulf's Guide, in Vanilla Diablo, Diablo had 1666 hp:
image
image

It seems like this is just a graphical error though - the actual hp is set in line 173, which correctly divides by 2 (although if we want to be pedantic, not exactly, since the hp value is bit shifted to the left, Diablo's hp is set to 1666.5 rather than 1666 if it were an int that was truncated when divided by 2 - I'd say that's pretty close though).

StephenCWills:
@DakkJaniels Your base assumption is that I don't fully understand why the variables are there whereas you somehow do. I find this to be a bit insulting given how thorough I was in my investigation of this issue. I'm aware that you have no way of knowing how thorough I had been, but you should have just said what you think those variables are for, and then I wouldn't feel compelled to nitpick your proposed approach before you've even implemented it.

First of all, `InitMonsterGFX()` is loading CEL and CL2 data for monster animations and the missiles they may have to produce, plus TRNs for palette transformations. It's something we don't do in `MonstersData` because it would be wasteful of RAM. It makes sense to use the `CMonster` struct in this place to house that data when the level is loaded.

Likewise, `InitMonsterSND()` is loading audio data from sound files. The same rationale applies to this because it's fundamentally the same thing. These functions are loading assets from disk so it makes sense to do it as needed when loading the dungeon level.

Copying all the data from `MonsterData` structs into `CMonster` structs when loading the level, however, would serve no real purpose. At best, you might be able to convince me that precalculating Diablo's HP would avoid us duplicating the logic in `PrintMonstHistory()`, but you'd also have to search through `LevelMonsterTypes` to find the appropriate `CMonster` instance so you're not really simplifying the code. I would argue that a better approach to deduplication would be to introduce member functions in the `MonsterData` struct that calculate HP from the initial HP values, a strategy that would be both less intrusive and less wasteful.

BTW, `MType` and `MData` are pointer variables in the `Monster` struct. They are not tables.

DakkJaniels:
@StephenCWills I apologize for my language and the implication. Let me just make the PR to better explain what I mean, because I didn't mean it exactly as I wrote it, since yes, there would be many things that would not need to be copied over exactly. I appreciate your feedback though, and I do like your idea about the member functions though, I'll mess around with it.
To clarify though, this is more what I was thinking:
image
image
And then other information that needs to be modified, based on difficulty/gametype/playercount, could have gone in InitMonsterType. Having the member function might be an overall better idea though.

Regarding your BTW: what's the difference between a pointer variable that points to an struct, and a struct? Unless you are just objecting to my colloquial use of "table" to refer to the information stored in the struct.

StephenCWills:
> And then other information that needs to be modified, based on difficulty/gametype/playercount, could have gone in InitMonsterType. Having the member function might be an overall better idea though.

Hm, that's actually a good point. I glossed over the fact that you mentioned additional discrepancies between play modes as I was too focused on this one discrepancy in Diablo's HP. Although I do think this is probably a good idea after all, I'd say it falls outside the scope of this PR. The bug needs to be fixed regardless of whatever additional cleanups or improvements we decide to apply later.

> Regarding your BTW: what's the difference between a pointer variable that points to an struct, and a struct? Unless you are just objecting to my colloquial use of "table" to refer to the information stored in the struct.

I assumed you meant to refer to the `MonstersData` array and the `LevelMonsterTypes` array--the initializers for which are formatted as tables--rather than the `MData` and `MType` pointer variables. If you meant it as a synonym for struct, then I completely misunderstood you.

DakkJaniels:
> > And then other information that needs to be modified, based on difficulty/gametype/playercount, could have gone in InitMonsterType. Having the member function might be an overall better idea though.
> 
> Hm, that's actually a good point. I glossed over the fact that you mentioned additional discrepancies between play modes as I was too focused on this one discrepancy in Diablo's HP. Although I do think this is probably a good idea after all, I'd say it falls outside the scope of this PR. The bug needs to be fixed regardless of whatever additional cleanups or improvements we decide to apply later.

I agree - I was just looking at the other changes that were outside the scope of that one change (switching to MData and deleting the other stuff that didn't make sense in the GFX function) and then started down the rabbit hole of looking at what might make sense overall. Regardless, I'll try and put something together, separate from this PR.

> 
> > Regarding your BTW: what's the difference between a pointer variable that points to an struct, and a struct? Unless you are just objecting to my colloquial use of "table" to refer to the information stored in the struct.
> 
> I assumed you meant to refer to the `MonstersData` array--the initializer for which is formatted as a table--and the `LevelMonsterTypes` array rather than the `MData` and `MType` pointer variables. If you meant it as a synonym for struct, then I completely misunderstood you.

Sorry about that, I'm not a professional developer, so sometimes I don't use words correctly haha. I meant what MData and MType ultimately point to (their respective structs), and the stuff that needs to be modified should be held in what is being pointed to by MType.

AJenbo:
> > Diablo clearly has the 3333 HP from Hellfire in the table. During spawn his HP needs to be reduced if the expansion is not active:
> 
> No one is disputing that. This PR does reduce his HP to the correct amount.

This was a reference to this:

> The HP value was adjusted as though it was working from the full Hellfire HP value of 3333, but it was actually working from 1333,

But I probably didn't quite understand exactly what is being referred to there.

AJenbo (approved):
Nice work. The monster init code is such a mess. I got close to writing a test for it, but there is a lot to test there and I'm tired :sleeping:

Issue #4044, opened 2022-02-08
damages target’s armor affix working incorrectly for Ring of Engagement in Hellfire


ikonomov:
https://github.com/diasurgical/devilutionX/blob/32d39c65246b17fbf6b89d4a914beb04f02ee7e8/Source/items.cpp#L1084
The above code generates the power for the affix "damages target’s armor" only for magic items, but not for unique items.  The affix functions differently in both games https://github.com/diasurgical/devilutionX/blob/32d39c65246b17fbf6b89d4a914beb04f02ee7e8/Source/items.cpp#L940
Because of this the Ring of Engagement has a much higher power than it should in Hellfire.  Instead of having IPL_TARGAC 2, 2 it has 4,12.

PR #4047, opened 2022-02-08
fix armor pierce on uniques and display affix values

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/4044
And displays the actual armor pierce value when not in hellfire


AJenbo:
Isn't this a duplicate of what is happening here? 

https://github.com/diasurgical/devilutionX/blob/master/Source/items.cpp#L1090

qndel:
> Isn't this a duplicate of what is happening here?
> 
> https://github.com/diasurgical/devilutionX/blob/32d39c65246b17fbf6b89d4a914beb04f02ee7e8/Source/items.cpp#L1090

yeah, the old one should be removed I think, will test later :D

AJenbo:
needs testing = draft

Source/items.cpp
```diff
@@ -710,8 +710,15 @@ int CalculateToHitBonus(int level)
 	}
 }
 
-int SaveItemPower(Item &item, const ItemPower &power)
+int SaveItemPower(Item &item, ItemPower &power)
 {
+	if (!gbIsHellfire) {
+		if (power.type == IPL_TARGAC) {
+			power.param1 = 1 << power.param1;
```
DakkJaniels, line 717: I know these bitshifts were here before, but looking at piercing, puncturing, and bashing, this results in 2-8, 4-16, and 6-24, which doesn't match the original values.
image

qndel: quite the opposite
3 << 1 = 6
3 << 2 = 12
3 << 3 = 24

DakkJaniels: oops sorry about that, saw the parameters backwards ( `param1 << 1' etc ) by accident.

StephenCWills (approved):
I know this is still in draft, but I tried generating a few rings and it seems to work properly.

PR #4049, opened 2022-02-08
Unify monster statistic adjustments

DakkJaniels:
InitMonsterType function added, moved modifications based on vanilla/hellfire, single/multiplayer, and difficulty to that function and removed from InitMonsters. Other changes required, still WIP.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -167,11 +167,8 @@ void InitMonster(Monster &monster, Direction rd, int mtype, Point position)
 	monster.ChangeAnimationData(MonsterGraphic::Stand);
 	monster.AnimInfo.TickCounterOfCurrentFrame = GenerateRnd(monster.AnimInfo.TicksPerFrame - 1);
 	monster.AnimInfo.CurrentFrame = GenerateRnd(monster.AnimInfo.NumberOfFrames - 1) + 1;
-
 	monster.mLevel = monster.MType->mLevel;
-	
 	monster._mhitpoints = (monster.MType->mMinHP + GenerateRnd(monster.MType->mMaxHP - monster.MType->mMinHP + 1)) << 6;
-
 	monster._mAi = monster.MData->mAi;
 	monster._mint = monster.MData->mInt;
 	monster._mgoal = MGOAL_NORMAL;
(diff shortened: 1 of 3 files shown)


AJenbo:
/root/repo/Source/monster.cpp:3691:11: error: redeclaration of 'int hp_modifier'
       int hp_modifier;

DakkJaniels:
> /root/repo/Source/monster.cpp:3691:11: error: redeclaration of 'int hp_modifier' int hp_modifier;

Thanks missed this. I also want to change over PrintMonstHistory before I think it's "done"

Source/monster.cpp
```diff
@@ -3691,6 +3650,75 @@ void GetLevelMTypes()
...
+	LevelMonsterTypes[monst].mMaxHP = MonstersData[mtype].mMaxHP;
+	if (!gbIsHellfire && mtype == MT_DIABLO) {
+		LevelMonsterTypes[monst].mMinHP /= 2;
+		LevelMonsterTypes[monst].mMaxHP /= 2;
+		LevelMonsterTypes[monst].mLevel -= 15;
+	}
+	
+	if (!gbIsMultiplayer)
+		LevelMonsterTypes[monst].mMinHP = std::max(LevelMonsterTypes[monst].mMinHP / 2, 1);
+		LevelMonsterTypes[monst].mMinHP = std::max(LevelMonsterTypes[monst].mMaxHP / 2, 1);
```
StephenCWills, line 3667: This looks wrong. You'll also need to fix a bunch of erroneous indentation that seems to have been caused by this.

```suggestion
	if (!gbIsMultiplayer) {
		LevelMonsterTypes[monst].mMinHP = std::max(LevelMonsterTypes[monst].mMinHP / 2, 1);
		LevelMonsterTypes[monst].mMinHP = std::max(LevelMonsterTypes[monst].mMaxHP / 2, 1);
	}
```

Source/monster.cpp
```diff
@@ -4625,39 +4691,14 @@ void PrintMonstHistory(int mt)
...
-		}
-		if (sgGameInitInfo.nDifficulty == DIFF_NIGHTMARE) {
-			minHP = 3 * minHP + hpBonusNightmare;
-			maxHP = 3 * maxHP + hpBonusNightmare;
-		} else if (sgGameInitInfo.nDifficulty == DIFF_HELL) {
-			minHP = 4 * minHP + hpBonusHell;
-			maxHP = 4 * maxHP + hpBonusHell;
-		}
+		int minHP = LevelMonsterTypes[mt].mMinHP;
+		int maxHP = LevelMonsterTypes[mt].mMaxHP;
```
StephenCWills, line 4695: I think you have to look up the data by type rather than using the type as an index directly.
`if (LevelMonsterTypes[index].mtype == mt) { /* This is the one */ }`

PR #4050, opened 2022-02-08
Adjust dungeon reroll bugfix

obligaron:
Notes:
- Removes old vanilla bugfix

### DIFF
--- a/Source/drlg_l2.cpp
+++ b/Source/drlg_l2.cpp
@@ -2899,10 +2899,6 @@ void GenerateLevel(lvl_entry entry)
 	bool doneflag = false;
 	while (!doneflag) {
 		nRoomCnt = 0;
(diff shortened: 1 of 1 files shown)


galaxyhaxz:
LGTM. You can initialize them in either the place you had originally or now. They only need to be initialized once however, since the game will initialize them each time in the while loop if the quest has spawned.

Also I think I should point out, the original bug usually only affects substitutions. However in rare cases, if the set piece coords from the existing level land where the stairs should be on the other level, the game WILL not be able to place stairs down. If stairs can't be placed, it rerolls the whole dungeon.

Just an explanation for how the original bug worked. Documented last year thanks to @NiteKat for posting the issue (:

AJenbo (approved):

PR #4055, opened 2022-02-10
remove InitLightMax

qndel:


Source/lighting.cpp
```diff
@@ -20,7 +20,7 @@ int VisionId;
 Light Lights[MAXLIGHTS];
 uint8_t ActiveLights[MAXLIGHTS];
 int ActiveLightCount;
-char LightsMax;
+const char LightsMax = 15;
```
AJenbo, line 23:
```suggestion
constexpr char LightsMax = 15;
```

You can probably also move it to the header.

AJenbo (approved):

PR #4056, opened 2022-02-10
TRN rewrite + bugfix

qndel:
Old TRN system was pretty poo - the game could only handle 8 unique monsters with it - now unlimited.
I've also noticed a bug while doing the rewrite - TRNs for missiles never worked, except for rhino.
I think they forgot to plug the system for the rest of the missiles.
trnbeforefix
trnafterfix
Here's bloodlust before and after fix. (notice blue color baked into first frames of the animation)
The goal of this PR is to make future modding easier - adding custom TRNs to missiles etc.

Deathspit:
deathspittrn

Plaguewrath:
image

image

Screenshot from Crucible as a bonus:
image

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -848,6 +848,12 @@ void MakeLightTable()
 		tbl += 240;
 	}
 
+	LoadFileInMem("PlrGFX\\Infra.TRN", tbl, 256);
+	tbl += 256;
+
+	LoadFileInMem("PlrGFX\\Stone.TRN", tbl, 256);
+	tbl += 256;
+
 	for (int i = 0; i < 8; i++) {
 		for (uint8_t col = 226; col < 239; col++) {
 			if (i != 0 || col != 226) {
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -848,12 +848,6 @@ void MakeLightTable()
 		tbl += 240;
 	}
 
-	LoadFileInMem("PlrGFX\\Infra.TRN", tbl, 256);
-	tbl += 256;
-
-	LoadFileInMem("PlrGFX\\Stone.TRN", tbl, 256);
-	tbl += 256;
-
 	for (int i = 0; i < 8; i++) {
 		for (uint8_t col = 226; col < 239; col++) {
 			if (i != 0 || col != 226) {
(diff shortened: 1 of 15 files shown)


Source/scrollrt.cpp
```diff
@@ -318,9 +319,10 @@ void DrawMissilePrivate(const Surface &out, const Missile &missile, Point target
 
 	const Point missileRenderPosition { targetBufferPosition + missile.position.offsetForRendering - Displacement { missile._miAnimWidth2, 0 } };
 	CelSprite cel { missile._miAnimData, missile._miAnimWidth };
-	if (missile._miUniqTrans != 0)
-		Cl2DrawLightTbl(out, missileRenderPosition.x, missileRenderPosition.y, cel, missile._miAnimFrame, missile._miUniqTrans + 3);
-	else if (missile._miLightFlag)
+	if (missile._miUniqTrans != 0) {
+		uint8_t *trn = GetUniqueMonsterTRN(UniqueMonstersData[Monsters[missile._misource]._uniqtype - 1].mTrnName);
```
glebm, line 323: This does a string map lookup in the render loop. Can't we load these once at initialization time and get rid of the map altogether?

Source/lighting.cpp
```diff
@@ -849,12 +849,6 @@ void MakeLightTable()
 		tbl += 240;
 	}
 
-	LoadFileInMem("PlrGFX\\Infra.TRN", tbl, 256);
```
glebm, line 852: The previous approach seems better -- no allocations or string map lookups

Source/scrollrt.cpp
qndel: will try :P

Source/lighting.cpp
qndel: yes but it used the lighttable, the goal of this PR was to make TRNs usable by name and not only by some voodoo index

glebm: Extract the voodoo index into a constant?

Source/scrollrt.cpp
qndel: Added TRNs to monsters (at least I hope so) @glebm

Source/lighting.cpp
qndel: Restored + added functions to get these TRNs

glebm (approved):

PR #4057, opened 2022-02-11
The Sharp prefix bug from vanilla

ikonomov:
The Sharp prefix has the wrong PLOk value set to false preventing it from being dropped.


ikonomov (commented):
added a space to preserve column format

qndel:
While this is a proper fix, it can cause existing items to morph, don't think it's worth it.

ikonomov:
Opps, I did not consider this.  Yeah, definitely not worth it

obligaron:
@qndel could this be added to #2930 ? 😁

qndel:
> @qndel could this be added to #2930 ? 😁

not really because that PR doesn't cause items to morph duh :D

obligaron:
My impression of #2930 was that we add a new (vanilla incompatible) item generation. There we don't need to care about morphing, cause it's a new algorithm.
The vanilla compatibility comes with a new flag that decides if the old or new logic is used.
That's why I thought we could add it to the new algorithm without breaking (one way) compatibility with vanilla.
But perhaps I missed something.

PR #4060, opened 2022-02-13
Unique Monster Variable Rename

DakkJaniels:
Data member `customHitpoints` actually modifies the `mHit` and `mHit2` of the monster (its to hit rating) not the hitpoints of the monster (which would be _mmaxhp or _mhitpoints). Confirmed this reflects the code in Devilution as well:
image


AJenbo (dismissed):

qndel:
@DakkJaniels while we are at it, could you fix the alignment of the header and the rest of the table? ;)

DakkJaniels:
> @DakkJaniels while we are at it, could you fix the alignment of the header and the rest of the table? ;)

fixed and spread them out a little more. the commas are slightly different for mMagicRes, but not going to move that around on each line...

qndel:
> > @DakkJaniels while we are at it, could you fix the alignment of the header and the rest of the table? ;)
> 
> fixed and spread them out a little more. the commas are slightly different for mMagicRes, but not going to move that around on each line...

then don't expect it to get merged lol
why didn't you just fix the header?

DakkJaniels:
> then don't expect it to get merged lol
why didn't you just fix the header?

You said fix the header *and the rest of the table*.  If you can explain how to make the 15 character header fit when the field below is only 9 characters (without expanding the column) I will do it.

Regardless, I see the comma thing was introduced as part of adjusting the field widths, and that the resistance column was an easy fix by adjusting the alignment for that field.  If you don't want to merge it after this fix, oh well 🤷🏽‍♂️

qndel:
> > then don't expect it to get merged lol
> > why didn't you just fix the header?
> 
> You said fix the header _and the rest of the table_. If you can explain how to make the 15 character header fit when the field below is only 9 characters (without expanding the column) I will do it.
> 
> Regardless, I see the comma thing was introduced as part of adjusting the field widths, and that the resistance column was an easy fix by adjusting the alignment for that field. If you don't want to merge it after this fix, oh well 🤷🏽‍♂️

I don't because it's a mess, look at whitespaces

DakkJaniels:
So having the header name overlap an adjacent column is preferable to increasing the width of the field? I'm not sure what you are referring to in my latest commit, unless you are referring to trailing white space, in which case I can easily remove that.

qndel:
> So having the header name overlap an adjacent column is preferable to increasing the width of the field? I'm not sure what you are referring to in my latest commit, unless you are referring to trailing white space, in which case I can easily remove that.

Then easily do that, duh. We don't add random whitespaces for no reason.

DakkJaniels:
> Then easily do that, duh. We don't add random whitespaces for no reason.

Then just say that because it's hard to read your mind and know exactly what whitespaces you are referring to. I didn't notice the trailing white space that was introduced as part of adjusting the field widths. If I had, I would have removed it. Now I have removed it.

qndel:
> > Then easily do that, duh. We don't add random whitespaces for no reason.
> 
> Then just say that because it's hard to read your mind and know exactly what whitespaces you were referring to. I didn't notice the trailing white space that was introduced as part of adjusting the field widths. If I had, I would have removed it. Now I have removed it.

That's not on me to point out but on you to check your diff xD
"customToHit,customArmorClass" mind adding a space between these and to the table? ;)

DakkJaniels:
fixed

PR #4064, opened 2022-02-14
Revised Attack Speed power

kphoenix137:
image

This changes the affix power for attack speed modifiers. Instead of set frame reductions for each affix, instead they increase attack speed by adding 10%, 20%, 30% and 30% (Haste). This is intended to be a solution to the original issue with attack speed modifiers that could not be reimplemented (would cause up to 4 reduces attack frames which would end up completely unbalanced on classes using weapons they are already fast with). This solution increases attack speed more fairly for all classes. Fast attacks such as Sword of Haste on Warrior and Bow of swiftness on Rogue are unable to become faster than they are in vanilla. This mainly affects the slowest animations, such as axe on Sorcerer.

The main takeaway is that this will affect early gameplay by giving you Readiness which will actually do something, if you have a low base speed for that weapon type.

Warrior:
melee = unchanged behavior from vanilla
axe = unchanged from vanilla
bow = readiness reduces 1 frame, swiftness reduces 1 frame
staff = readiness reduces 1 frame, swiftness reduces 1 frame, speed reduces 2 frames, haste reduces 2 frames

Rogue:
melee = unchanged from vanilla
axe = readiness reduces 1 frame, swiftness reduces 2 frames, speed reduces 3 frames, haste reduces 3 frames
bow = unchanged from vanilla
staff = readiness reduces 1 frame, swiftness reduces 1 frame, speed reduces 2 frames, haste reduces 2 frames

Sorcerer:
melee = readiness reduces 1 frame, swiftness reduces 2 frames, speed reduces 2 frames, haste reduces 3 frames
axe = readiness reduces 1 frame, swiftness reduces 2 frames, speed reduces 3 frames, haste reduces 4 frames
bow = readiness reduces 1 frame, swiftness reduces 2 frames
staff = readiness reduces 1 frame, swiftness reduces 2 frames, speed reduces 3 frames, haste reduces 3 frames

Hellfire Characters:
Bard:
melee = unchanged from vanilla
axe = readiness reduces 1 frame, swiftness reduces 2 frames, speed reduces 3 frames, haste reduces 3 frames
bow = readiness reduces 1 frame, swiftness reduces 1 frame
staff = readiness reduces 1 frame, swiftness reduces 1 frame, speed reduces 2 frames, haste reduces 2 frames

Barbarian:
melee = swiftness reduces 1 frame, haste reduces 2 frames
axe = swiftness reduces 1 frame, haste reduces 2 frames
bow = readiness reduces 1 frame, swiftness reduces 1 frame
staff = readiness reduces 1 frame, swiftness reduces 1 frame, speed reduces 2 frames, haste reduces 2 frames

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3881,7 +3881,7 @@ void DoOil(Player &player, int cii)
 		if ((item._iFlags & ISPL_FASTERATTACK) != 0)
 			return _("+30% increased attack speed");
 		if ((item._iFlags & ISPL_FASTESTATTACK) != 0)
-			return _("+40% increased attack speed");
+			return _("+30% increased attack speed");
 		return _("Another ability (NW)");
 	case IPL_FASTRECOVER:
 		if ((item._iFlags & ISPL_FASTRECOVER) != 0)
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3875,13 +3875,13 @@ void DoOil(Player &player, int cii)
 		return _("penetrates target's armor");
 	case IPL_FASTATTACK:
 		if ((item._iFlags & ISPL_QUICKATTACK) != 0)
-			return _("quick attack");
+			return _("+10% increased attack speed");
 		if ((item._iFlags & ISPL_FASTATTACK) != 0)
-			return _("fast attack");
+			return _("+20% increased attack speed");
 		if ((item._iFlags & ISPL_FASTERATTACK) != 0)
-			return _("faster attack");
+			return _("+30% increased attack speed");
 		if ((item._iFlags & ISPL_FASTESTATTACK) != 0)
-			return _("fastest attack");
+			return _("+40% increased attack speed");
 		return _("Another ability (NW)");
 	case IPL_FASTRECOVER:
 		if ((item._iFlags & ISPL_FASTRECOVER) != 0)
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -447,7 +447,7 @@ void StartAttack(int pnum, Direction d)
 		int increasePercent = 0;
 
 		if ((player._pIFlags & ISPL_FASTESTATTACK) != 0) {
-			increasePercent = 40;
+			increasePercent = 30;
 		} else if ((player._pIFlags & ISPL_FASTERATTACK) != 0) {
 			increasePercent = 30;
 		} else if ((player._pIFlags & ISPL_FASTATTACK) != 0) {
@@ -456,9 +456,7 @@ void StartAttack(int pnum, Direction d)
 			increasePercent = 10;
 		}
 
-		if (increasePercent >= ((1000 / (player._pAFNum - 4)) * player._pAFNum - 1000) / 10) {
-			skippedAnimationFrames = 4;
-		} else if (increasePercent >= ((1000 / (player._pAFNum - 3)) * player._pAFNum - 1000) / 10) {
+		if (increasePercent >= ((1000 / (player._pAFNum - 3)) * player._pAFNum - 1000) / 10) {
 			skippedAnimationFrames = 3;
 		} else if (increasePercent >= ((1000 / (player._pAFNum - 2)) * player._pAFNum - 1000) / 10) {
 			skippedAnimationFrames = 2;
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -442,18 +442,19 @@ void StartAttack(int pnum, Direction d)
 	}
 
 	int skippedAnimationFrames = 0;
-	int increasePercent = 0;
-	if ((player._pIFlags & ISPL_FASTESTATTACK) != 0) {
-		increasePercent = 40;
-	} else if ((player._pIFlags & ISPL_FASTERATTACK) != 0) {
-		increasePercent = 30;
-	} else if ((player._pIFlags & ISPL_FASTATTACK) != 0) {
-		increasePercent = 20;
-	} else if ((player._pIFlags & ISPL_QUICKATTACK) != 0) {
-		increasePercent = 10;
-	}
 
 	if ((player._pIFlags & ISPL_QUICKATTACK) != 0 || (player._pIFlags & ISPL_FASTATTACK) != 0 || (player._pIFlags & ISPL_FASTERATTACK) != 0 || (player._pIFlags & ISPL_FASTESTATTACK) != 0) {
+		int increasePercent = 0;
+
+		if ((player._pIFlags & ISPL_FASTESTATTACK) != 0) {
+			increasePercent = 40;
+		} else if ((player._pIFlags & ISPL_FASTERATTACK) != 0) {
+			increasePercent = 30;
+		} else if ((player._pIFlags & ISPL_FASTATTACK) != 0) {
+			increasePercent = 20;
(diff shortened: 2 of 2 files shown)


DakkJaniels:
Wouldn't this be better in a mod than pulling into master, since you are modifying how the affix works from how it originally worked?

kphoenix137:
> Wouldn't this be better in a mod than pulling into master, since you are modifying how the affix works from how it originally worked?

Perhaps. In the original, 2 of the speed affixes are fully non-functional, which is a problem that was never able to be properly addressed in any way- this is my proposed solution.

DakkJaniels:
Isn't only 1 affix non-functional, and the other is just the same as the one below it (speed = haste)? So wouldn't a better "fix",  that meets the original intent/logic of the game to just stop the non-functional (readiness) one from spawning, unless you are playing Hellfire, in which case it should only spawn on bows (although since their are cursed items, one could also argue to just keep readiness in game, since it does "something", i.e., stops you from getting a different suffix). I agree that it would be nice if the affixes all did something different from one another, but this changes the result of how each affix works in some cases.

kphoenix137:
> Isn't only 1 affix non-functional, and the other is just the same as the one below it (speed = haste)? So wouldn't a better "fix",  that meets the original intent/logic of the game to just stop the non-functional (readiness) one from spawning, unless you are playing Hellfire, in which case it should only spawn on bows (although since their are cursed items, one could also argue to just keep readiness in game, since it does "something", i.e., stops you from getting a different suffix). I agree that it would be nice if the affixes all did something different from one another, but this changes the result of how each affix works in some cases.

Yes that's correct. I mentioned in Discord the best course of action for the base game might be just a rename of the panel text for the affixes as follows:
Quick Attack -> (Empty)
Fast Attack -> Fast Attack
Faster Attack -> Faster Attack
Fastest Attack -> Faster Attack

I think we should a special case for bows in Hellfire to indicate an increase to arrow velocity instead of attack speed increase

DakkJaniels:
I meant stop it from dropping (i.e., being selected as an affix), but yes, just keeping it and renaming the panel text would make sense and be in line with the original combat mechanics of the game.

kphoenix137:
> I meant stop it from dropping (i.e., being selected as an affix), but yes, just keeping it and renaming the panel text would make sense and be in line with the original combat mechanics of the game.

That would be a reasonable approach. What about haste

DakkJaniels:
> > I meant stop it from dropping (i.e., being selected as an affix), but yes, just keeping it and renaming the panel text would make sense and be in line with the original combat mechanics of the game.
> 
> That would be a reasonable approach. What about haste? Removing that as a possible drop would effectively cut your odds of getting a 2 frame skip weapon in half, unless we just rename haste to speed and have 2 speed affixes with different qlvls

I said "removing" was only applicable to readiness. I would keep haste regardless and rename the panel description accordingly.

ikonomov:
If this behavior went unchanged in vanilla after 7 patches 5 years after it was released, I think we can assume that they were content with the result.  1.07 even says Haste has been fixed, whatever that meant.  They must have been aware of the issue and fixed it or didn't fix it as they saw appropriate for the gameplay that they were after.  Obviously if the game worked at 60 ticks per second instead of 20 we wouldn't be having this problem.  Your proposal would still not fix Readiness, as it would not give faster attack speed universally, just a random increase in % rate of fire that sometimes would round down to no change, sometimes it would have an affect.  As bad as having it not function in the original at all, I think this proposed solution is not an improvement.

kphoenix137:
> If this behavior went unchanged in vanilla after 7 patches 5 years after it was released, I think we can assume that they were content with the result.  1.07 even says Haste has been fixed, whatever that meant.  They must have been aware of the issue and fixed it or didn't fix it as they saw appropriate for the gameplay that they were after.  Obviously if the game worked at 60 ticks per second instead of 20 we wouldn't be having this problem.  Your proposal would still not fix Readiness, as it would not give faster attack speed universally, just a random increase in % rate of fire that sometimes would round down to no change, sometimes it would have an affect.  As bad as having it not function in the original at all, I think this proposed solution is not an improvement.

They were also content with all the other stuff, like the dupe bug, or endless other problems in base game. As far as the default functionality goes, making each one deduct a different amount of frames as a standard would break the balance of the game  making fast classes insanely fast with haste. There is no good solution using default behavior, other than removal of haste completely, and readiness (except not removing readiness on bows in hellfire). The diablo 2 way of doing attack speed would have been perfect even with 20 tps, however I doubt they were willing to put very much effort in with these patches. For the most part they were focusing on game breaking bugs. The goal of this PR is not to universally deduct a certain amount of frames per affix, that is how vanilla works which perpetuates the issue of, for example, barbarian getting a 33% speed boost using axe of haste, where a sorcerer would get only a 14% boost with axe of haste, less than double. The goal is to give the same boost to characters regardless of their AFNum. Unfortunately this poses the issue of rebalance.

DakkJaniels:
I agree with what you are saying, in terms of the game, but:
> The goal of this PR is not to universally deduct a certain amount of frames per affix, that is how vanilla works which perpetuates the issue of, for example, barbarian getting a 33% speed boost using axe of haste, where a sorcerer would get only a 14% boost with axe of haste, less than double. The goal is to give the same boost to characters regardless of their AFNum. Unfortunately this poses the issue of rebalance.

As you note, there is a balance issue here.  My understanding of DevilutionX is that the "classic" mechanics and experience are still supposed to be maintained, it just has "engine improvements, bugfixes and optional quality of life features" added to it from Devilution. I agree that this is probably a fine idea if you want to change the original Diablo experience, but I don't know that game mechanics changes are appropriate to pull into this project.

kphoenix137:
> I agree with what you are saying, in terms of the game, but:
> > The goal of this PR is not to universally deduct a certain amount of frames per affix, that is how vanilla works which perpetuates the issue of, for example, barbarian getting a 33% speed boost using axe of haste, where a sorcerer would get only a 14% boost with axe of haste, less than double. The goal is to give the same boost to characters regardless of their AFNum. Unfortunately this poses the issue of rebalance.
> 
> As you note, there is a balance issue here.  My understanding of DevilutionX is that the "classic" mechanics and experience are still supposed to be maintained, it just has "engine improvements, bugfixes and optional quality of life features" added to it from Devilution. I agree that this is probably a fine idea if you want to change the original Diablo experience, but I don't know that game mechanics changes are appropriate to pull into this project.

The only reason this PR is still open is in case we collaboratively could come up with a better solution to the problem. I don't believe the PR as it remains at the moment should be merged. It'll remain a draft until then

DakkJaniels:
In terms of maintaining the original experience, I see two options (which were mentioned above):

- Option 1: Adjust the panel descriptions to more accurately portray what is happening:

 In Diablo:
- [ ] Readiness: "No change in attack speed"
- [ ] Swiftness: "Faster Attack Speed"
- [ ] Speed/Haste: "Fastest Attack Speed" (or use Fast and Faster... doesn't really matter)

In Hellfire:
- [ ] Readiness on bows: "fast arrow speed"
- [ ] Swiftness on bows: "faster arrow" (or use Faster and Fastest... again doesn't really matter too much)
- [ ] Same as Diablo for other weapons

- Option 2: In addition to the panel updates above, do not allow the readiness affix to drop while playing Diablo, or to drop on non-bow weapons in Hellfire. Keep the affix around to maintain compatibility. This change does modify the original drop chances, however, so I can see why this might not want to be done.

Issue #2724, opened 2021-08-27
Port inventory synchronization from Pionere's fork


AJenbo:
**Describe**
@pionere has a lot of interesting in his port, one of the major parts is inventories syncronization for multiplayer

This is mainly done via the two following commits: https://github.com/diasurgical/devilutionX/commit/ee39ebabd354bf963141a8e73620ceb7de63e255 - https://github.com/diasurgical/devilutionX/commit/5baf6a6dc99dd106754cdb3fe0a2bb84609f8139

Porting this is complicated a bit by large part of the fork still using an holder iteration of the code and a different code style, but it should still be reasonably to do do so.

**Steps**
1. Go over the two commits and get a good feel for the semantic changes, and separate it from the style changes.
2. Start applying the same changes to the code
3. Run two clients on your local system and run a TCP game where you test that the items are being synced (you could do this by adding some player debug info).
4. Create a pull request with the backported code

**Expected behavior**
It might be a good idea to also test this on the fork before starting the process to get a good feel for how it works and now to verify it.

obligaron:
I think with #5217 this is obsolete/done.

Issue #2681, opened 2021-08-23
Desync in spellcasting


qndel:
Oldest commit I got the bug on is https://github.com/diasurgical/devilutionX/commit/f2d858af4b15e82f71dc393471d4461afb2c5451 for now (not related to the bug, just oldest one I've checked atm)

![allspellsbroke](https://user-images.githubusercontent.com/14297035/130465572-4a65863f-5730-4427-8b72-9f8f487a7504.gif)

To reproduce: just get 2 players, 1 casts once, 2 sees doublecast

julealgon:
Could this be related to the "hold button" feature you added?

AJenbo:
probably not

qndel:
ab8afa7cd7e83b7df295cac7956cd6358f068728 is the first bad commit
commit ab8afa7cd7e83b7df295cac7956cd6358f068728
Author: Anders Jenbo <anders@jenbo.dk>
Date:   Sun Jul 25 05:57:42 2021 +0200

    Fix edge cases in mouse action tracking code (#2452)

    Fixes #2431

    Also:
    - Track spell targets
    - Merge repeat walk in to the new repeat action code
    - Avoid time based repeats (use action frame instead)


Needs verifying, I've been bisecting on different game speeds when I didn't know I could only get the bug on slowest

obligaron:
I looked a bit at the problem and want to share what I discovered so far.

- When you click the right mouse button `RightMouseDown` is called. `RightMouseDown` calls `CheckPlrSpell` and `CheckPlrSpell` calls `NetSendCmdLocParam3` with `CMD_SPELLXY`. No `player.destAction` is updated.
- When the next rendering happens  `ProcessInput` is called and `ProcessInput` calls `RepeatMouseAction`. `RepeatMouseAction` calls `CheckPlrSpell`, because it thinks the player has new new command (`player.destAction` was not set). Then `CheckPlrSpell` calls `NetSendCmdLocParam3` with `CMD_SPELLXY`. No `player.destAction` is updated.
  Note: This step can happen multiple times depending on the rendering speed / vsync and when right mouse button (relative to next game tick) is pressed.
- When a gametick is due `multi_process_network_packets` is called. `multi_process_network_packets` calls `ParseCmd` and `ParseCmd` calls `OnSpellTile`. Now `player.destAction` is updated to `ACTION_SPELL`.
[...]

PR #4074, opened 2022-02-19
Fix double casting bug

### LINKED ISSUE #789: Fix double-casting bug
When casting a spell or using a scroll through the spellbook, players are allowed to cast 2 times in rapid succession, resulting in 2 distinct bugs:
1. Scrolls can be used twice (even though they are consumable, single-use items)
2. Spells can be cast without sufficient mana

At first, I thought this was related to animation skipping (#779), in particular this section on `player.cpp` which interrupts the spell casting animation and starts it again if the user is quickly casting:

https://github.com/diasurgical/devilutionX/blob/9e1428dd0b7738c0c305ffaa0b7e7882736c94f0/Source/player.cpp#L3256-L3272

However, even without this skipping in place, it is still possible to perform double casting and get to negative mana (or use a scroll multiple times):

| Spells | Scrolls |
| --- | --- |
| double-casting-spell | double-casting-scroll |

I'm sure this is a single-frame type of exploit, but wasn't able to find what exactly is causing it yet.

### LINKED ISSUE #793: Scroll removal behavior is different when using as items vs using as spell

obligaron:
Fixes https://www.youtube.com/watch?v=GFFmOEVU3tM
Fixes #789
Fixes #793

## Notes
- _Fix double casting bug - use separated info for casted and queued spells_ address the possibility to swap casted and queued spell infos (spell level, spell type, etc.)
- _Checks conditions for spell again, cause initial check was done when spell was queued_ address #789
- _Store scroll location in SpellCastInfo.spellFrom and remove the scroll only after the spell is casted_ address #793
- This could result in slightly more desyncs, cause we now validate that a spell can be casted and this could differ from client to client (for example different mana or inventory). But we still have #2681 so I think this doesn't add much.
- I'm unsure if we should add this to 1.4 or give it more time in 1.5.

## Todo

- [ ] Needs inventory sync to work correctly with scrolls #2724


AJenbo:
Would be nice to have in 1.4.0, but would also be nice to have 1.4.0 asap :)

StephenCWills (commented):
I would feel more comfortable with this change if we were to sync player data more aggressively first.

Source/player.cpp
```diff
@@ -514,26 +514,44 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
 		return;
 	}
 
+	// Checks conditions for spell again, cause initial check was done when spell was queued and the parameters could be changed meanwhile
+	bool isValid = true;
+	switch (player.queuedSpell.spellType) {
+	case RSPLTYPE_SKILL:
+	case RSPLTYPE_SPELL:
+		isValid = CheckSpell(MyPlayerId, player.queuedSpell.spellId, player.queuedSpell.spellType, true) == SpellCheckResult::Success;
```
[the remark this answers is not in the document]
obligaron: Thanks for catching this one. 🙂
This was a copy&past error from `CheckPlrSpell`.

Source/player.cpp
```diff
@@ -514,26 +514,44 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
...
 
+	// Checks conditions for spell again, cause initial check was done when spell was queued and the parameters could be changed meanwhile
+	bool isValid = true;
+	switch (player.queuedSpell.spellType) {
+	case RSPLTYPE_SKILL:
+	case RSPLTYPE_SPELL:
+		isValid = CheckSpell(MyPlayerId, player.queuedSpell.spellId, player.queuedSpell.spellType, true) == SpellCheckResult::Success;
+		break;
+	case RSPLTYPE_SCROLL:
+		isValid = UseScroll(player, player.queuedSpell.spellId);
```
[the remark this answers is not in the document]
obligaron: You are right. 🙂 That's why I wrote in the initial comment that we can have slightly more desync.
For reference #2681 happens much more often.
Another way to mitigate this would be to resolve #2724.

Personally I wouldn't set this as a show stopper.
The damage is still synced. The player location is updated frequently.
More important is that the player that generates the events can't cheat/abuse the bug anymore.

Source/player.cpp
```diff
@@ -514,26 +514,44 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
...
+		break;
+	case RSPLTYPE_CHARGES:
+		isValid = UseStaff(player, player.queuedSpell.spellId);
```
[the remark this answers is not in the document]
obligaron: Does `CMD_CHANGEPLRITEMS` also set the current charges of the staff correctly?
If not this could also result in a desync, cause Client A check's correctly that there are not enough charges and forbids the cast. Client B sees more charges and still casts.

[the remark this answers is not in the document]
StephenCWills: It doesn't. Since the staff is regenerated from the item seed, it will actually just give it the charges it would have at spawn. Also, the `UseStaff()` function doesn't reduce staff charges. That happens in `UseMana()` when it calls `UseStaffCharge()`, but `UseMana()` doesn't do anything if the casting player isn't the local player so staff charges aren't tracked accurately at all.

But I don't consider that to be much of an issue. Assuming Client A is the caster in your scenario, they won't send the network message in the first place since they don't have charges left to cast. That type of desync could only happen as a result of queuing the spell on Client A and then cancelling it as a result of this logic, which will only happen once at the point when the staff is depleted.

Source/player.cpp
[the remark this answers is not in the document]
StephenCWills: The reason I consider this issue to be pretty significant is because it's guaranteed to affect all casting from scrolls when playing brand new characters, and it's also likely to affect nearly all casting from scrolls except Town Portal. The only argument I can think of to downplay this issue is the fact that most players probably don't use scrolls very often, if at all. But for players who do, I don't believe they'll typically be toting them around at the start of every game they play.

Source/player.cpp
```diff
@@ -515,26 +515,44 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
 		return;
 	}
 
+	// Checks conditions for spell again, cause initial check was done when spell was queued and the parameters could be changed meanwhile
+	bool isValid = true;
+	switch (player.queuedSpell.spellType) {
+	case RSPLTYPE_SKILL:
+	case RSPLTYPE_SPELL:
+		isValid = CheckSpell(pnum, player.queuedSpell.spellId, player.queuedSpell.spellType, true) == SpellCheckResult::Success;
```
StephenCWills, line 523: With the previous issue out of the way, I have one last issue to report here. Mana isn't synced between clients, so if a player joins a game with 0 mana, their spells will always fail on remote clients for the duration of the game. However, this one should be easy to solve by syncing mana in `TPktHdr` structs the way we do for HP.

Source/player.cpp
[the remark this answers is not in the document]
obligaron: Currently we need to decide between:
1. Double-cast bugs:
   I have only one scroll, but I can cast it twice.
   I have only enough mana for one cast, but I can cast twice.
   Affects gameplay / balance.
   Happens in single- **and** multiplayer.
   Reported in Issue #789.
   Present in current master.
3. **Visual** desync in multiplayer when a scroll is not synced or mana is not synced correctly
   The player A doesn't **see** Player B casting.
   Happens only in multiplayer.
   Resulting damage is still synced.
   Multiplayer still works.
   Will be introduced with this pr.

Ideally we shouldn't have any of these two bugs.
When #2861 is fixed this should be possible.
But if i have to decide I would go with the **visual** multiplayer-only bug.
That's why I think this pr is a improvement and would prefer to merge it. 🙂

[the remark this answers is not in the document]
AJenbo: Why does that require that the game works on Mac from before 2010?

AJenbo:
My preference here is that we keep the vanilla bug until we can fix them with out breaking new things. Changing behavior (even if it's arguably better) could easily make people nervous that things are more broken then they appear.

Source/player.cpp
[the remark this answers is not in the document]
obligaron: Without old mac support everything is nothing. 😛 
Yeah sorry i meant #2724.

obligaron:
When strictly following this logic we need to revert holding mouse button feature, cause it introduced/worsen #2681. 😉

For a compromise: We add support to sync mana between players. Then only scrolls are affected. Casting scrolls happens much less then spells and scrolls are often used for thing that doesn't affect other players (like identify). Would this be okay? Otherwise I think we can close this pr for the near future?

AJenbo:
> When strictly following this logic we need to revert holding mouse button feature, cause it introduced/worsen #2681. wink

It wasn't know at the time when it got merged :)

AJenbo:
> For a compromise: We add support to sync mana between players. Then only scrolls are affected. Casting scrolls happens much less then spells and scrolls are often used for thing that doesn't affect other players (like identify). Would this be okay? Otherwise I think we can close this pr for the near future?

Sounds ok, but I'm a bit tired at this point so my logic isn't 100%

Source/player.cpp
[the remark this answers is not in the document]
StephenCWills: Framing the fact that remote clients do not see the spell as being merely a visual desync is a bit overzealous. The spell gets cast on the local client, but does not get cast on the remote client. Try casting an slvl 15 Fire Wall from a scroll using this PR and let me know if you still think the issue is merely a visual desync.

Furthermore, there's nothing the player can do to get around this. If two new players are roaming through the dungeon together and they happen upon a Scroll of Chain Lightning, only one of them will actually be able to try out the spell. With this PR, the other will also inevitably not see what happened.

Let's also consider the implications of the double-casting bug. Sure, I can cast an Identify scroll twice in quick succession, but that doesn't mean I can use it to identify two items. I can cast an extra Fireball from a scroll, which means I can hit up to 8 or 9 monsters (but more likely 0-3 monsters) that I guess I shouldn't have been able to hit with Fireball. Maybe I can get two Stone Curses, occurring simultaneously, when I should have only been able to get one. These issues don't seem very impactful, and I'm grasping at straws. If there's a major implication of this bug I am not seeing that could justify fast-tracking this fix, please correct me.

I do not dispute that #2724 should help to resolve this issue, but I don't think that's an argument for merging this now. Instead, I think it makes more sense to sync player inventories first, then revisit this PR.

Source/player.cpp
```diff
@@ -515,26 +515,44 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
...
+		break;
+	case RSPLTYPE_SCROLL:
+		isValid = UseScroll(player, player.queuedSpell.spellId);
+		break;
+	case RSPLTYPE_CHARGES:
+		isValid = UseStaff(player, player.queuedSpell.spellId);
+		break;
+	}
+	if (!isValid)
+		return;
```
StephenCWills, line 533: It seems that runes are completely broken in this PR. I can't even put one down.

Source/player.cpp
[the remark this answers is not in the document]
obligaron: > Framing the fact that remote clients do not see the spell as being merely a visual desync is a bit overzealous. The spell gets cast on the local client, but does not get cast on the remote client. Try casting an slvl 15 Fire Wall from a scroll using this PR and let me know if you still think the issue is merely a visual desync.

This following happens in the fire wall example:

- Player 1 cast a fire wall on monsters.
- Player 2 doesn't see the fire wall
- Monster dies on Player 1 client. Monster damage/kill gets synced.  (`CheckMissileCol` -> `MonsterMHit` -> `M_StartHit`/`M_StartKill`).
- Player 2 sees monster dying, cause of the synced damage/kill.

That means the monster still dies/take damage but you don't see it => that's why I concluded this is a only visual desync. But perhaps I miss some cases? 🤔

(Note if this is not only a visual desync #2681 is even more serious/critical then I thought, Cause there happens the same problem. Player 1 casts but Player 2 doesn't "see" it.)

> Furthermore, there's nothing the player can do to get around this. If two new players are roaming through the dungeon together and they happen upon a Scroll of Chain Lightning, only one of them will actually be able to try out the spell. With this PR, the other will also inevitably not see what happened.

You are correct Player 2 doesn't see the cast and this is a regression this pr would introdcue. But also in master only one player can "try out"/cast the spell or do I miss something?

> Let's also consider the implications of the double-casting bug. Sure, I can cast an Identify scroll twice in quick succession, but that doesn't mean I can use it to identify two items. I can cast an extra Fireball from a scroll, which means I can hit up to 8 or 9 monsters (but more likely 0-3 monsters) that I guess I shouldn't have been able to hit with Fireball. Maybe I can get two Stone Curses, occurring simultaneously, when I should have only been able to get one. These issues don't seem very impactful, and I'm grasping at straws. If there's a major implication of this bug I am not seeing that could justify fast-tracking this fix, please correct me.

For me it's some sort of cheating. You can do thing's you aren't allowed to do.

You can
- cast a scroll multiple times
- cast a spell/scroll/staff charge with higher spell level
- cast a spell without the needed mana

See the linked video in the description.

> I do not dispute that #2724 should help to resolve this issue, but I don't think that's an argument for merging this now. Instead, I think it makes more sense to sync player inventories first, then revisit this PR.

I think this should go in, even if #2724 isn't merged. That doesn't mean it needs to be in 1.4 release. I think this needs time, cause it changes some core mechanics and need well testing; even in singleplayer.

Perhaps we can work together to get this done. You are much more versed in regards to multiplayer then me. So perhaps you can work on #2724. Then we can get both in 1.5 release; a fixed bug and a nicely synchronized inventory. I would really appreciate that. 🙂

Source/player.cpp
obligaron: Thank you very much for catching this one. 👍 🙂 
I should really test more in hellfire. 😢
Hopefully with the last change this should be fixed.

Source/player.cpp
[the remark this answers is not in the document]
StephenCWills: > That means the monster still dies/take damage but you don't see it => that's why I concluded this is a only visual desync. But perhaps I miss some cases? 🤔

There's two important cases you missed, I think. First, monster AI will avoid flaming tiles. On Player 1's client, the monster may be stuck behind a Fire Wall, whereas on Player 2's client, the monster will walk right on through. In my testing, I found the monsters eventually just teleported to the other side of the Fire Wall on Player 1's client. Therefore, this issue can sabotage some early-game strategies, although admittedly early-game Fire Wall doesn't last very long.

Second, players can also get hurt by walking onto flaming tiles. In my testing, Player 1 thought Player 2 was taking damage whereas Player 2 thought he was standing on a normal tile. This situation can ultimately resolve itself okay, except that Player 2 probably should not have been able to just brazenly walk through the flaming tiles without taking damage. Also Player 1 won't be able to walk through even though Player 2 won't have any idea why.

> Perhaps we can work together to get this done. You are much more versed in regards to multiplayer then me. So perhaps you can work on #2724. Then we can get both in 1.5 release; a fixed bug and a nicely synchronized inventory. I would really appreciate that. 🙂

FYI, my comments here are just me sharing my thoughts after having reviewed this PR to determine if I think it should be included in 1.4. As you might imagine, I was thinking it should be probably be postponed so that these comments can be addressed. I certainly don't mind working on #2724 to see that this change makes it into 1.5.

Source/player.cpp
StephenCWills: I also don't usually test in Hellfire, and it's bitten me a few times.

It does appear that the issue is fixed, although this has the same issue as scrolls. In this case, however, Player 1 was able to kill Player 2 with a Greater Rune of Fire that Player 2 didn't even know had been placed, let alone triggered. 😓

Source/player.cpp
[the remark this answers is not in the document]
StephenCWills: > You are correct Player 2 doesn't see the cast and this is a regression this pr would introdcue. But also in master only one player can "try out"/cast the spell or do I miss something?

I forgot to respond to this. You are correct that only one player can try out the spell, but both players should be able to see the result. A Barbarian playing with a Sorcerer, for instance, will likely be unable to use the scroll but would be able to see the effect if he watched the Sorcerer cast it. I wouldn't call it much of a shared experience for the two players otherwise.

Source/player.cpp
obligaron: Yeah same issues and same benefits (double "casting"/placing a rune is fixed). 🙂

Source/player.cpp
obligaron: I opened #4207 for this. 🙂

Source/player.cpp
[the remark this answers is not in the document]
obligaron: > There's two important cases you missed, I think. First, monster AI will avoid flaming tiles. On Player 1's client, the monster may be stuck behind a Fire Wall, whereas on Player 2's client, the monster will walk right on through. In my testing, I found the monsters eventually just teleported to the other side of the Fire Wall on Player 1's client. Therefore, this issue can sabotage some early-game strategies, although admittedly early-game Fire Wall doesn't last very long.
> 
> Second, players can also get hurt by walking onto flaming tiles. In my testing, Player 1 thought Player 2 was taking damage whereas Player 2 thought he was standing on a normal tile. This situation can ultimately resolve itself okay, except that Player 2 probably should not have been able to just brazenly walk through the flaming tiles without taking damage. Also Player 1 won't be able to walk through even though Player 2 won't have any idea why.

I think the most problematic spell is fire wall or? For the other spells it's mostly cosmetic/visual. 🤔 
And don't get me wrong, I would really like to get both issues (cheating/abuse & visual/sync) correct.

> FYI, my comments here are just me sharing my thoughts after having reviewed this PR to determine if I think it should be included in 1.4. As you might imagine, I was thinking it should be probably be postponed so that these comments can be addressed. I certainly don't mind working on #2724 to see that this change makes it into 1.5.

It would be a great if you can help to get #2724 forward. ❤️ 
And yes the 1.4 release is out of scope for this pr. Personally I would like to get this pr in early 1.5 version, so we can play test it a long time.

AJenbo:
On a bit of a side note, after working on getting quest items to drop correctly in multiplayer I did find it concerning that the inventory isn't synced :)

AJenbo:
Could we add a `- [] TODO` to this PR regarding it depending on inventory being synced?

Source/player.h
```diff
@@ -243,9 +252,10 @@ struct Player {
 	float progressToNextGameTickWhenPreviewWasSet;
 	int _plid;
 	int _pvid;
-	spell_id _pSpell;
-	spell_type _pSplType;
-	int8_t _pSplFrom; // TODO Create enum
+	/* @brief next queued spell */
+	SpellCastInfo queuedSpell;
+	/* @brief the spell that is currently casted */
+	SpellCastInfo castedSpell;
```
AJenbo, line 258:
```suggestion
	SpellCastInfo currentSpell;
```

AJenbo: Oh, I might need to read more of the code to understand it better :D

How about "inflightSpell"?

obligaron: In some way the queued spell is also in-flight or? 🤔 

The pr tries to distinguish between the spell (what spell, level, etc.) that is currently performed by the player (`_pmode` = `PM_SPELL`) and what is the next spell (`destAction` = `ACTION_SPELL`/`ACTION_SPELLWALL`/`ACTION_SPELLMON`/`ACTION_SPELLPLR`).
In master/vanilla for both the same variables were used and this introduced the double casting bug.

AJenbo: no, it's pending on the runway

obligaron: Renamed to `executedSpell`.
Hope this is better. 🙂

AJenbo: yeah that works :)

Source/inv.cpp
```diff
@@ -1897,15 +1897,45 @@ int8_t CheckInvHLight()
...
+	if (!item->isEmpty() && isCurrentSpell(*item)) {
+		if (itemSlot <= INVITEM_INV_LAST)
+			player.RemoveInvItem(static_cast<int>(itemIndex));
+		else
+			player.RemoveSpdBarItem(itemIndex);
+		return true;
+	}
+
+	// Didn't find it at the selected slot, take the first one we find
+	// This path is always used when the scroll is consomed via spell selection
```
AJenbo, line 1929:
```suggestion
	// This path is always used when the scroll is consumed via spell selection
```

Source/player.h
```diff
@@ -210,6 +210,15 @@ struct PlayerAnimationData {
 	}
 };
 
+struct SpellCastInfo {
+	spell_id spellId;
+	spell_type spellType;
+	/* @brief Inventory location for scrolls */
+	int8_t spellFrom;
+	/* @brief Used for spell level */
+	int spellLevel;
```
AJenbo, line 219: spell level is fairly limited in range (0-22 afaik), so probably can use a much smaller size type.

obligaron: That's true and i think we should do this. 🙂 
But I wouldn't include this in this pr.
The pr is big enough and in master `spellLevel` is also an `int`.

Source/inv.cpp
obligaron: Thanks. Fixed. 🙂

Source/missiles.cpp
```diff
@@ -1094,21 +1095,25 @@ void AddLightningRune(Missile &missile, const AddMissileParameter &parameter)
 	int dmg = 16 * (GenerateRndSum(10, 2) + lvl + 2);
 	missile._midam = dmg;
 	AddRune(missile, parameter.dst, MIS_LIGHTWALL);
+	UseMana(Players[missile._misource], SPL_RUNELIGHT);
```
AJenbo, line 1098: Runes don't use mana

obligaron: The function `UseMana` is also called (in master) with `_pSplType` set to `RSPLTYPE_SCROLL` (`RemoveCurrentSpellScroll`) and `RSPLTYPE_CHARGES` (`UseStaffCharge`).
I reused it for runes, cause they are also considered "scrolls". 😕 

But yes the name is misleading (in master and in this pr even more).
Do you have a idea for a new name?

AJenbo: Ok i thought that might be the case. How about `ConsumeSpell()`?

Source/inv.cpp
```diff
@@ -1914,22 +1944,16 @@ void UseStaffCharge(Player &player)
...
-	if (!CanUseStaff(staff, player._pSpell))
+	if (!CanUseStaff(staff, player.executedSpell.spellId))
 		return;
 
 	staff._iCharges--;
 	CalcPlrStaff(player);
 }
 
-bool UseStaff(const spell_id spell)
+bool UseStaff(Player &player, spell_id spellId)
```
AJenbo, line 1954: Since it's just a shortcut now:
```suggestion
bool CanUseStaff(Player &player, spell_id spellId)
```

AJenbo (approved):

Source/inv.cpp
```diff
@@ -1897,15 +1897,45 @@ int8_t CheckInvHLight()
 	return rv;
 }
 
-bool UseScroll(const spell_id spell)
+bool RemoveCurrentSpellScroll(Player &player)
```
AJenbo, line 1900: A bit of an odd name, maybe
```suggestion
bool RemoveActiveScroll(Player &player)
```

qndel: why not RemoveUsedScroll? :P

Source/missiles.cpp
obligaron: 👍 
I changed it to `ConsumeSpell` 🙂

Source/inv.cpp
obligaron: I renamed it to `CanUseStaff`. 🙂 
I also renamed `UseScroll` to `CanUseScroll`, cause the function only checks if the scroll is usable.

Source/inv.cpp
obligaron: I tried `ConsumeScroll`, cause that aligns well with `ConsumeSpell` and `ConsumeStaffCharge`.
I hope you like it. 🙂

PR #4076, opened 2022-02-20
TRNs for different classes  + TRN debug command

qndel:
Created a way to add TRNs for classes + debug command to test TRNs
Made loading TRNs optional so we can just add TRN files later to make it work

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -20,6 +20,8 @@
 #include "engine/load_file.hpp"
 #include "engine/points_in_rectangle_range.hpp"
 #include "engine/random.hpp"
+#include "engine/render/clx_render.hpp"
+#include "engine/trn.hpp"
 #include "engine/world_tile.hpp"
 #include "gamemenu.h"
 #include "help.h"
@@ -2237,6 +2239,10 @@ void LoadPlrGFX(Player &player, player_graphic graphic)
 	*fmt::format_to(pszName, FMT_COMPILE(R"(PlrGFX\{0}\{1}\{1}{2}.CL2)"), path, string_view(prefix, 3), szCel) = 0;
 	const uint16_t animationWidth = GetPlayerSpriteWidth(cls, graphic, animWeaponId);
 	animationData.sprites = LoadCl2Sheet(pszName, animationWidth);
+	std::optional<std::array<uint8_t, 256>> trn = GetClassTRN(player);
+	if (trn) {
+		ClxApplyTrans(*animationData.sprites, trn->data());
+	}
 }
 
 void InitPlayerGFX(Player &player)
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -42,6 +42,7 @@ bool DebugVision = false;
 bool DebugGrid = false;
 std::unordered_map<int, Point> DebugCoordsMap;
 bool DebugScrollViewEnabled = false;
+std::string debugTRN;
 
 namespace {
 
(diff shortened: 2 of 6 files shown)


qndel:
@glebm Wondering if https://github.com/diasurgical/devilutionX/pull/4076/commits/6c6bb906f80874eaaaeabbe64209275a49482463 is fine, I'm not a template expert 😅 not sure if there won't be any conflict between 
```cpp
template <typename T, bool isOptional = false>
```
and
```cpp
template <typename T, std::size_t N>
```
since bool can be treated as size and the opposite.

I'm just hoping 
```cpp
template <typename T, std::size_t N>
void LoadFileInMem(const char *path, std::array<T, N> &data)
```
arguments for that function are distinguishable enough that there won't be any confusion
Do you have any idea how to handle it in a more elegant way? Ideally some type like bool that doesn't convert to a number :D

glebm:
It's fine but perhaps let's add a `LoadFileInMem(SDL_RWops *, ...)` instead?

Source/player.cpp
```diff
@@ -364,16 +366,20 @@ void StartWalk(int pnum, Displacement vel, Direction dir, bool pmWillBeCalled)
...
 	if (data == nullptr && gbQuietMode)
 		return;
 
-	const byte *directionFrames[8];
+	byte *directionFrames[8];
 	CelGetDirectionFrames(data.get(), directionFrames);
+	auto trn = GetClassTRN(player);
 	for (size_t i = 0; i < 8; i++) {
+		if (trn != nullptr) {
+			Cl2ApplyTrans(directionFrames[i], reinterpret_cast<std::array<uint8_t, 256> &>(trn[0]), (int)directionFrames[i][0]);
```
AJenbo, line 381: Don't use c-style casting

Source/player.cpp
```diff
@@ -364,16 +366,20 @@ void StartWalk(int pnum, Displacement vel, Direction dir, bool pmWillBeCalled)
...
 	if (data == nullptr && gbQuietMode)
 		return;
 
-	const byte *directionFrames[8];
+	byte *directionFrames[8];
 	CelGetDirectionFrames(data.get(), directionFrames);
+	auto trn = GetClassTRN(player);
 	for (size_t i = 0; i < 8; i++) {
+		if (trn != nullptr) {
+			Cl2ApplyTrans(directionFrames[i], reinterpret_cast<std::array<uint8_t, 256> &>(trn[0]), static_cast<int>(directionFrames[i][0]));
```
glebm, line 381: Previously we didn't have a `Cl2ApplyTrans` here for warrior/rogue/etc.
Why apply these now? For modding?

glebm (approved):

Source/player.cpp
glebm: Nevermind, I've now read the PR description :smile:

Source/player.cpp
```diff
@@ -364,16 +366,20 @@ void StartWalk(int pnum, Displacement vel, Direction dir, bool pmWillBeCalled)
...
 {
 	data = nullptr;
 	data = LoadFileInMem(path);
 	if (data == nullptr && gbQuietMode)
 		return;
 
-	const byte *directionFrames[8];
+	byte *directionFrames[8];
 	CelGetDirectionFrames(data.get(), directionFrames);
+	auto trn = GetClassTRN(player);
```
glebm, line 378:
```suggestion
	std::unique_ptr<uint8_t[]> trn = GetClassTRN(player);
```

Source/player.cpp
glebm, line 381: This is a very unsafe cast!

Let's make `GetClassTRN` return an `std::array<uint8_t, 256>` instead.

glebm (changes requested):

glebm (dismissed):

glebm (approved):

PR #4092, opened 2022-02-23
add spawning of unique monsters to debug command

DakkJaniels:
modifications to DebugCmdSpawnMonster function to enable spawning of unique monsters. Also made a slight modification to PlaceUniqueMonst to enable this. I added a Point parameter to the function with a default value. If the function is called with the default value, it will operate as it normally does, however if it is called and has a different value for the Point, it will instead spawn a unique monster at that location and skip spawning a pack with it.

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -568,25 +568,33 @@ std::string DebugCmdSpawnMonster(const string_view parameter)
 
 	std::transform(name.begin(), name.end(), name.begin(), [](unsigned char c) { return std::tolower(c); });
 
+	/* check normal monster list */
 	int mtype = -1;
-	for (int i = 0; i < 138; i++) {
+	int uniqidx = -1;
+	for (int i = 0; i < NUM_MTYPES; i++) {
 		auto mondata = MonstersData[i];
 		std::string monsterName(mondata.mName);
 		std::transform(monsterName.begin(), monsterName.end(), monsterName.begin(), [](unsigned char c) { return std::tolower(c); });
+		/* if match is not found, check next */
 		if (monsterName.find(name) == std::string::npos)
 			continue;
+		/* match was found */
 		mtype = i;
 		break;
 	}
 
+	/* check unique monster list */
 	if (mtype == -1) {
-		for (int i = 0; i < 100; i++) {
+		for (int i = 0; UniqueMonstersData[i].mtype != MT_INVALID; i++) {
 			auto mondata = UniqueMonstersData[i];
 			std::string monsterName(mondata.mName);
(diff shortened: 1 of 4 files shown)


DakkJaniels:
Note: actual lines changed is very low - I had to move PlaceUniqueMonst outside of the generic namespace so that I could call it from the debug command. Diff for PlaceUniqueMonst is the following:
```diff
+void PlaceUniqueMonst(int uniqindex, int miniontype, int bosspacksize, Point pos /* 0, 0*/)
{
	auto &monster = Monsters[ActiveMonsterCount];
	const auto &uniqueMonsterData = UniqueMonstersData[uniqindex];

	int uniqtype;
	for (uniqtype = 0; uniqtype < LevelMonsterTypeCount; uniqtype++) {
		if (LevelMonsterTypes[uniqtype].mtype == uniqueMonsterData.mtype) {
			break;
		}
	}

	int count = 0;
+	int xp = pos.x;
+	int yp = pos.y;
+	bool skipGroupPlacement = false; /* skips the placement of debug monsters */

+	if (xp == 0 && yp == 0){
		while (true) {
			xp = GenerateRnd(80) + 16;
// skip to end of if block at line 3458
+     } else {
+		monster._udeadval = 1;
+		skipGroupPlacement = true;
+	}
// skip to line 3562
+	/* skip group placement if placing debug monsters */
+	if (!skipGroupPlacement && uniqueMonsterData.monsterPack != UniqueMonsterPack::None) {
		PlaceGroup(miniontype, bosspacksize, uniqueMonsterData.monsterPack, ActiveMonsterCount - 1);
	}
```

Source/debug.cpp
```diff
@@ -568,33 +568,41 @@ std::string DebugCmdSpawnMonster(const string_view parameter)
...
+		for (int i = 0; UniqueMonstersData[i].mtype != MT_INVALID; i++) {
 			auto mondata = UniqueMonstersData[i];
 			std::string monsterName(mondata.mName);
 			std::transform(monsterName.begin(), monsterName.end(), monsterName.begin(), [](unsigned char c) { return std::tolower(c); });
+			/* if a match is not found, check next */
 			if (monsterName.find(name) == std::string::npos)
 				continue;
-			mtype = mondata.mtype;
+			/* match was found */
+			mtype = UniqueMonstersData[i].mtype;
```
qndel, line 596:
```suggestion
			mtype = mondata.mtype;
```

qndel:
Best to avoid having debug code modify normal functions :) Replaced with https://github.com/diasurgical/devilutionX/pull/4094

PR #4094, opened 2022-02-23
fix spawning unique monsters

qndel:
allows spawning unique monsters
makes sound work for spawned monsters
improves name matching - earlier first partial match would take priority over a full match

Split into two commands to avoid a situation where having a match in unique monsters will always be prioritized 
(skeleton would spawn skeleton king instead of a skeleton)


StephenCWills (approved):

DakkJaniels:
I don't get what the big deal is with loading an mtype into the array? If there is no space available, just overwrite the last space like you are doing.

obligaron (approved):

PR #4095, opened 2022-02-24
modify SpawnOnePremium in items.dat for readability

DakkJaniels:
I think this is a little bit easier to follow, cleanly splits the vanilla vs. Hellfire portions and removes the "keepgoing" variable  - loop continues until count is reached or you find a valid item.


Source/items.cpp
```diff
@@ -2169,10 +2169,7 @@ void SpawnOnePremium(Item &premiumItem, int plvl, int playerId)
 
 	plvl = clamp(plvl, 1, 30);
 
-	int count = 0;
-
-	do {
-		keepGoing = false;
+	for (int count = 0; count < 150; count++) {
```
obligaron, line 2172: Question: Is this the same as master? And does it matter?
In master and diablo (`gbIsHellfire` = false) the loop could iterates more then 150 times if `premiumItem._iIvalue > 140000` is always true.

DakkJaniels: Good catch, just checked Devilution. I guess hypothetically you could be stuck in the loop forever if you got very unlucky? lol

Could separate the increment out of the loop declaration and just increment as part of the hellfire side - or set count to 0 when entering the Diablo side (the former seems less "hacky" to me). 

I guess the question is how long you could get realistically stuck in the loop, and does it make sense to have a count limitation in general, even if it didn't work like that originally? Any thoughts here?

kphoenix137: > Good catch, just checked Devilution. I guess hypothetically you could be stuck in the loop forever if you got very unlucky? lol
> 
> Could separate the increment out of the loop declaration and just increment as part of the hellfire side - or set count to 0 when entering the Diablo side (the former seems less "hacky" to me).
> 
> I guess the question is how long you could get realistically stuck in the loop, and does it make sense to have a count limitation in general, even if it didn't work like that originally? Any thoughts here?

This is true, in fact you can get stuck in the loading screen if this function never resolves. I don't think this is ideal.

Source/items.cpp
```diff
@@ -2181,55 +2182,52 @@ void SpawnOnePremium(Item &premiumItem, int plvl, int playerId)
...
+				itemValue = mostValuablePlayerItem == nullptr ? 0 : mostValuablePlayerItem->_iIvalue;
+				break;
+			}
+			default:
+				itemValue = 0;
+				break;
+			}
+			itemValue = itemValue * 4 / 5; // avoids forced int > float > int conversion
+			if (premiumItem._iIvalue <= 200000
+			    && premiumItem._iMinStr <= strength
```
ephphatha, line 2223: The `value <= 200000` check really only exists to make sure the player can potentially afford to buy the item (as that's how much they can carry with a full inventory of full gold stacks). Now that the stash has been implemented we could allow higher value items to be generated if you wanted. This would be a behaviour change from vanilla Hellfire so would need to be called out as an intentional change.
```suggestion
			if (premiumItem._iMinStr <= strength
```

Source/items.cpp
```diff
@@ -2169,10 +2169,11 @@ void SpawnOnePremium(Item &premiumItem, int plvl, int playerId)
 
 	plvl = clamp(plvl, 1, 30);
 
-	int count = 0;
+	int maxCount = 65000;
+	if (gbIsHellfire)
+		maxCount = 150;
```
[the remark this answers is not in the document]
ephphatha: I also think this will lead to items worth more than the cap being offered by Griswold where the game would otherwise have locked up, to avoid reopening #1478 need to accept this possibility and maybe call it out in the change notes?

DakkJaniels:
Comments all implemented, except for removing the gold cap on Hellfire items.

kphoenix137:
Is 65000 iterations necessary? I thought the game seed starts repeating itself far sooner than that

StephenCWills:
It's impossible to predict whether or when you will start seeing repeat seeds. The limit is somewhat arbitrary because it's probably not possible to find a sequence of 65k seeds that would all produce items that cannot be sold.

Source/items.cpp
DakkJaniels: Also, I believe changing the value thresholds would cause item morphing.

StephenCWills: The seed gets captured at the beginning of the loop for each item that gets generated, whether it fails or succeeds the threshold check. So changing the thresholds shouldn't cause any morphing.

DakkJaniels: oh right 👍🏽

Source/items.cpp
```diff
@@ -1941,10 +1938,13 @@ void SpawnOnePremium(Item &premiumItem, int plvl, const Player &player)
 
 	plvl = clamp(plvl, 1, 30);
 
-	int count = 0;
+	int maxCount = 150;
+	if (!gbIsHellfire) {
+		// Diablo logic could lead to an infinite loop if a suitable item can never be generated, using an arbitrary cutoff here to let the game continue.
+		maxCount = 65000;
```
AJenbo, line 1944: This equates to roughly 200ms on my 3700x, that's probably plenty fine. But still can't help but be nervous about putting an arbitrary cutoff in there :sweat_smile:

DakkJaniels: Could use max int if that makes you feel better 😄 No one is likely to let the game sit there for hours waiting for it to load town.

AJenbo: I think 2000000 would make me calm, cause that would be enough that the game would timeout during multiplayer so there is no way that is an acceptable value :D

StephenCWills: Would it be better to keep the infinite loop?

```c++
const bool unlimited = !gbIsHellfire;
for (int count = 0; unlimited || count < maxCount; count++) {
	...
}
```

AJenbo: If that change was applied then I think we can merge it and look in to applying a limit at a later point.

PR #4096, opened 2022-02-24
Search assets next to mpq the same way we search for mpqs

### LINKED ISSUE #4007: Language reverts to english after restarting the game
**Important information**
Windows 10 / compiled master https://github.com/diasurgical/devilutionX/commit/9536927301dafd33008bef6b4d5b5131541db5bc

**Describe**
Game seems to reset to english after restart

**To Reproduce**
Set the language to something other than english (either by editing diablo.ini or using ingame menu)
Restart game
See english

**Expected behavior**

obligaron:
Fixes #4007

## Notes
- Previous we searched for assets next to the mpq only for the **last** mpq we loaded
- On windows this was often the hf-mpqs that are found at the gog installation
- But for the language the case was special. The (language) settings are loaded after `devilutionx.mpq`. So if we didn't find `devilutionx.mpq` the old search path wasn't set. See @StephenCWills [good analysis](https://github.com/diasurgical/devilutionX/issues/4007#issuecomment-1048145649).
- With this change we search for overridden assets everywhere where we also search for mpqs.
- I think we should revisit our assets searching strategy. For example I would prefer loading /assets folder before mpqs and I wouldn't load from current working directory (mpqs and overridden files). But this is another topic.

### DIFF
--- a/Source/engine/assets.cpp
+++ b/Source/engine/assets.cpp
@@ -43,8 +43,9 @@ SDL_RWops *OpenAsset(const char *filename, bool threadsafe)
 
 	// Files next to the MPQ archives override MPQ contents.
 	SDL_RWops *rwops;
-	if (paths::MpqDir()) {
-		const std::string path = *paths::MpqDir() + relativePath;
+	const auto &searchPaths = paths::GetMPQSearchPaths();
+	for (const auto &mpqDir : searchPaths) {
+		const std::string path = mpqDir + relativePath;
 		// Avoid spamming DEBUG messages if the file does not exist.
 		if ((FileExists(path.c_str())) && (rwops = SDL_RWFromFile(path.c_str(), "rb")) != nullptr) {
 			LogVerbose("Loaded MPQ file override: {}", path);
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1725,6 +1725,9 @@ int DiabloMain(int argc, char **argv)
 	DiabloParseFlags(argc, argv);
 	InitKeymapActions();
 
+	// Initialize search paths for assets (mpq and relative files)
+	InitMPQSearchPaths();
+
 	// Need to ensure devilutionx.mpq (and fonts.mpq if available) are loaded before attempting to read translation settings
 	LoadCoreArchives();
 	was_archives_init = true;
--- a/Source/init.cpp
+++ b/Source/init.cpp
@@ -17,6 +17,7 @@
 #include "engine/assets.hpp"
 #include "mpq/mpq_reader.hpp"
 #include "options.h"
+#include "path.h"
 #include "pfile.h"
 #include "utils/language.h"
 #include "utils/log.hpp"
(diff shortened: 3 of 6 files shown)


AJenbo:
Isn't this going to result in a large penality on load time as every file will be probed for in multiple directories before the MPQ is checked. This is one of the motivations for only looking in the ./assets folder if the files aren't found in an MPQ.

obligaron:
From game start to a loaded singleplayer game, we got around 560 calls to `OpenAsset` and on my windows machine around 2800 checks for overridden files (search paths x `OpenAsset` calls).
With the old logic there are only 560 checks for overridden files. But the old logic was a little bit wired, cause it uses the path of the last loaded mpq and not a fix path.

Alternative fix could be probing only in `paths::BasePath`. But I wasn't sure how much we want to change the search logic before 1.4 is out.
Or a more complete revisit of the searching strategy (for examples see comment above).

obligaron:
From game start to a loaded singleplayer game, we got around 560 calls to `OpenAsset` and on my windows machine around 2800 checks for overridden files (search paths x `OpenAsset` calls).
With the old logic there are only 560 checks for overridden files. But the old logic was a little bit wired, cause it uses the path of the last loaded mpq and not a fix path.

Alternative fix could be probing only in `paths::BasePath`. But I wasn't sure how much we want to change the search logic before 1.4 is out.
Or a more complete revisit of the searching strategy (for examples see comment above).

AJenbo:
I'll give it a think, the current situation is defiantly a bit problematic :)

xGnoSiSx:
Apologies for jumping in here,

Isn't it better to just create a list of files to be overriden before loading anything? Then as you start loading the MPQs you don't have to do any other calls if you don't find an asset - because if it isn't in the list, you won't find it anyway.

AJenbo:
How would we know what files the user or a mod would like to overwrite??

Source/engine/assets.cpp
```diff
@@ -43,8 +43,9 @@ SDL_RWops *OpenAsset(const char *filename, bool threadsafe)
 
 	// Files next to the MPQ archives override MPQ contents.
 	SDL_RWops *rwops;
-	if (paths::MpqDir()) {
-		const std::string path = *paths::MpqDir() + relativePath;
+	const auto &searchPaths = paths::GetMPQSearchPaths();
+	for (const auto &mpqDir : searchPaths) {
```
glebm, line 47: This calls `OpenMpqFile(filename, ...)` multiple times with the same arguments.

glebm (changes requested):

Source/engine/assets.cpp
obligaron: Could you elaborate?
The `OpenMpqFile` call is outside the `searchPaths` loop, so it should only be called once (as in master).

glebm: Oh, sorry, misread the code. Nevermind

xGnoSiSx:
> How would we know what files the user or a mod would like to overwrite??

I assume that files aimed to be considered to be overriden are placed in a seperate directory/ies like assets.
You would scan those directories at start and build a list of files.

So if there is just 2 files in them, then that's just 2 overrides and you don't have to do anything extra about the rest of the thousands of files in the MPQs.

AJenbo:
I don't think this would work well, it could potentially hand the application for a long time if things are place in a folder with a lot of files. And there isn't really a good way of knowing what files may be relevant before hand.

obligaron:
Another problem is that the behavior is os specific.
For example linux is case sensitive but windows is case insensitive.

AJenbo:
Closed since I think we should instead figure out a more restrained solution that makes the path predictable. I would suggest we look at Skyrim and Quake 3 for inspiration.

PR #4100, opened 2022-02-27
Don't restart music if the same track is played

### LINKED ISSUE #799: Avoid restarting music on same-tileset dungeon level transition
This one is a suggestion for what I believe should be a simple change somewhere in the level loading logic. When moving from one floor to the next _on the same tileset_, I propose we do not stop playing the music. Currently, the music stops for the loading screen, and then starts over from the beginning after it finishes.

If we could check for the floor transition and validate that it is the same tileset as the previous level, we could keep the music going without interruption to provide a more seamless experience.

I also believe keeping the music playing is slightly more immersive.

IIRC, the "The Hell" MOD implemented something along these lines. I haven't played a lot of other mods so it might be the case for others as well.

Obviously, this being an improvement is my personal take. If the team agrees this is an interesting idea, I could look into it at some point. I have not looked into the implementation for this one yet.

obligaron:
Fixes #799

## Notes
- When the same track is played on both levels the music isn't restarted.
- When a different track is needed the music will change when the loading screen is shown (the music will stop after 20% finished). This feels a little bit like fade out / in.
- The music will not stop when quick-loading (if the saved level uses the same track).
- Music in menus continues playing.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -184,6 +184,7 @@ void FreeGame()
 	FreeDebugGFX();
 #endif
 	FreeGameMem();
+	music_stop();
 }
 
 bool ProcessInput()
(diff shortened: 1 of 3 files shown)


AJenbo (approved):

AJenbo:
Afaik that music stopping music in the menu was introduced by the refactoring of the menu between 1.04 and 1.09.
Hard to say if this was deliberate or just an accident of that refactoring. It does seam most people prefer the music playing all the time, it if pretty good after all :)

PR #4101, opened 2022-02-28
Fix duplicate affix names (Crimson and Fine)

kphoenix137:
Reverts affix name for Crimson back to Burgundy and Fine back to Sturdy (Taken from Diablo 1 Beta and Diablo 2).

https://github.com/diasurgical/devilutionX/pull/4101#issuecomment-1363224094


ikonomov:
If this was changed in the release version of the game, first we know from the beta it was Burgundy and then changed to Crimson, we can safely assume that this change is not a bug or an oversight, but intentional.  I think the intent of the designers should be respected regardless of how we feel about this change that makes two affixes have the same name.  Another example of two affixes sharing the same name is Fine, which exists both for weapons and for armor.

glebm:
> we can safely assume that this change is not a bug or an oversight

I don't think we can assume that. DIablo's development was rushed, didn't use version control, and stuff like this could've easily regressed due to somebody copy pasting the wrong line/file.

glebm:
> we can safely assume that this change is not a bug or an oversight

I don't think we can assume that. DIablo's development was rushed, didn't use version control, and stuff like this could've easily regressed due to somebody copy pasting the wrong line/file.

AJenbo:
They did actually use version control (MS Visual SourceSafe afaik), but sparingly so :)

qndel:
> They did actually use version control (MS Visual SourceSafe afaik), but sparingly so :)

The name is pretty ironic :D

FitzRoyX:
> If this was changed in the release version of the game, first we know from the beta it was Burgundy and then changed to Crimson, we can safely assume that this change is not a bug or an oversight, but intentional.

While many of kphoenix's prs are mods, this isn't one of them. The affix tables have several obvious errors that have zero chance of being intentional. The only problem with this one was that we needed to know the missing word to fix it. Thankfully, we do, based on two sources.

galaxyhaxz:
I looked into this for my own project and figured I'd share some of the findings here. **TLDR I agree with this change**, supporting evidence:
- In the pre-release demo it was called `Burgandy` (typo)
- In the battle.net beta it was fixed to `Burgundy`
- In Diablo 2 (Classic), they used `Red/Crimson/Burgundy/Garnet/Ruby`. Expansion Added Russet too.

As @FitzRoyX pointed out, the affix tables have obvious errors and you can see them by diffing the demo/beta/final tables.

A few more include:
- `King's` added in last minute and being level 28 when the weaker tier is 40
- `Sharp` is marked as cursed so shops can't sell it.
- `Fine` is duplicated for both armor/weapon. The armor version was called `Sturdy` in the beta, and also in Diablo 2.

Several other affixes in the demo/beta also got removed but added back in Diablo 2. For example, `of vileness` (Prevent Monster Heal) which while the affix was removed in the final game, is still fully coded in albeit bugged to trigger by Fire Arrows. `of vampires` was called `the vampire` in the beta and also in Diablo 2. The list goes on, really...

galaxyhaxz:
And a follow-up to the above regarding Diablo's development: The battle.net beta was released november 9 1996, less than 2 months before release. Diablo was originally going to be a single player game but Condor was going bankrupt and Blizzard decided to fund it on two condition: it had to have multiplayer and had to be real time.

This is very obvious when you look at all the planned features in the demo that didn't make it into the game. David Brevik stated in an interview he wanted homing arrows but they didn't have time, so it was added in Diablo 2 as `Guided Arrow`. It is very VERY VERY blatantly obvious that sometime between the pre-release demo and battle.net beta, the developers stopped working on new content (i.e. finishing those unused affixes) and went full force into just getting multiplayer to work to meet their deadline.

If Diablo had another 6 months in its development, I'm sure the changes they would have made would largely prove my theories right (:

AJenbo (approved):

ikonomov:
Currently Diablo 2 has no Burgundy, as far as I can tell in the expansion Russet was added and Burgundy was removed.  There must have been a reason why that name kept being changed.

Sturdy probably didn't get used because there's another affix called sturdiness.  Yes, to be exact sturdiness differs from sturdy, but it's close enough.

The point being, there is as much evidence about why they changed those affixes to what they are in the final version as much as there is about why they should have named them to what is being proposed.  This also bothers me and in fact I've renamed the affixes to Russet and Stout in my mod, but that's in my mod.  So I agree that both of these affixes should have had different names, but sadly we can't and shouldn't change King's and Sharp and I think neither should we change Fine (armors) and Crimson (31-40).

galaxyhaxz:
> but sadly we can't and shouldn't change King's and Sharp

I fixed both. Bugs are bugs. Rushed development is rushed. It's a 25 year old game and needs to be brought into the modern age.

>  There must have been a reason why that name kept being changed.

In the case of Diablo II, they reduced the overall number of single resistance affixes. There was "Green/Blue/Red/Orange" albeit unused and a few others they were enabled/changed.

> Sturdy probably didn't get used because there's another affix called sturdiness

True, but: In Diablo II there are such items as "Berserker's Berserker Axe of Berserk". You can't win all name combinations across all languages :P

PR #4102, opened 2022-02-28
Fix Attack Speed Affix Panel Strings

kphoenix137:
Changes Readiness & Swiftness to "fires quick speed arrows" & "fires fast speed arrows", on bow items in Hellfire.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3586,7 +3586,7 @@ bool DoOil(Player &player, int cii)
 		}
 		if (HasAnyOf(item._iFlags, ItemSpecialEffect::FastAttack)) {
 			if (gbIsHellfire && item._itype == ItemType::Bow)
-				return _("fires fast speed arrows");
+				return _("fires fast arrows");
 			return _("fast attack");
 		}
 		if (HasAnyOf(item._iFlags, ItemSpecialEffect::FasterAttack))
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3580,8 +3580,12 @@ bool DoOil(Player &player, int cii)
 		return _("penetrates target's armor");
 	case IPL_FASTATTACK:
 		if (HasAnyOf(item._iFlags, ItemSpecialEffect::QuickAttack))
+			if (gbIsHellfire && item._itype == ItemType::Bow)
+				return _("fires quick speed arrows");
 			return _("quick attack");
 		if (HasAnyOf(item._iFlags, ItemSpecialEffect::FastAttack))
+			if (gbIsHellfire && item._itype == ItemType::Bow)
+				return _("fires fast speed arrows");
 			return _("fast attack");
 		if (HasAnyOf(item._iFlags, ItemSpecialEffect::FasterAttack))
 			return _("faster attack");
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3579,14 +3579,16 @@ bool DoOil(Player &player, int cii)
 	case IPL_TARGAC:
 		return _("penetrates target's armor");
 	case IPL_FASTATTACK:
-		if (HasAnyOf(item._iFlags, ItemSpecialEffect::QuickAttack))
+		if (HasAnyOf(item._iFlags, ItemSpecialEffect::QuickAttack)) {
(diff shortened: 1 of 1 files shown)


Source/items.cpp
```diff
@@ -4013,7 +4016,8 @@ void PrintItemDetails(const Item &item)
 		AddPanelString(PrintItemPower(item._iPrePower, item));
 	}
 	if (item._iSufPower != -1) {
-		AddPanelString(PrintItemPower(item._iSufPower, item));
+		if ((item._itype == ItemType::Bow && gbIsHellfire) || (item._iFlags & ISPL_QUICKATTACK) == 0)
```
DakkJaniels, line 4019: Why are you only including the suffix power in the panel string under these specific cases?

kphoenix137: So it works as intended

DakkJaniels: Fair enough - I read something backwards so it wasn't making sense to me. Why not just blank out quick attack under PrintItemPower, since you are doing all that logic there anyway? Then return the string properly formatted for whatever version of the game you are using.

kphoenix137: Because we need to prevent a blank line from being added

DakkJaniels: This one probably could have stayed closed

DakkJaniels:
This one probably could have stayed closed

kphoenix137:
> This one probably could have stayed closed

With the latest change to attack speed modifiers, yes

kphoenix137:
@AJenbo I would like to apologize, I completely forgot the contents of this PR. This will actually be a useful PR for Hellfire for displaying different Info Strings for bows that use fast attack speed in Hellfire. I just need to remove the info display changes for melee items. Can you reopen this?

AJenbo:
sure, you should be able to as well

kphoenix137:
```
	case IPL_RNDARROWVEL:		return _("fires random speed arrows");```
Using this as a reference, what would be the best way of representing the increased arrow velocities in hellfire? We only have quick attack and fast attack flags. Here are some possible options:
"Fires quick speed arrows"
"Fires fast speed arrows"

"Fires faster speed arrows"
"Fires fastest speed arrows"

"Quick arrow speed"
"Fast arrow speed"

"Faster arrow speed"
"Fastest arrow speed"

"Fast arrow speed"
"Faster arrow speed"

**Comment by AJenbo** [comment]
Sounds good to me

**Comment by kphoenix137** [comment]
> Sounds good to me

Changes are finalized, I'll just need to rebase and it will be ready for review.

File: Source/items.cpp [hunk 2]
```diff
@@ -3565,10 +3565,16 @@ bool DoOil(Player &player, int cii)
 	case IPL_TARGAC:
 		return _("penetrates target's armor");
 	case IPL_FASTATTACK:
-		if (HasAnyOf(item._iFlags, ItemSpecialEffect::QuickAttack))
+		if (HasAnyOf(item._iFlags, ItemSpecialEffect::QuickAttack)) {
+			if (gbIsHellfire && item._itype == ItemType::Bow)
+				return _("fires quick speed arrows");
```
AJenbo, line 3570:
```suggestion
				return _("fires quick arrows");
```

Source/items.cpp
```diff
@@ -3565,10 +3565,16 @@ bool DoOil(Player &player, int cii)
...
-		if (HasAnyOf(item._iFlags, ItemSpecialEffect::QuickAttack))
+		if (HasAnyOf(item._iFlags, ItemSpecialEffect::QuickAttack)) {
+			if (gbIsHellfire && item._itype == ItemType::Bow)
+				return _("fires quick speed arrows");
 			return _("quick attack");
-		if (HasAnyOf(item._iFlags, ItemSpecialEffect::FastAttack))
+		}
+		if (HasAnyOf(item._iFlags, ItemSpecialEffect::FastAttack)) {
+			if (gbIsHellfire && item._itype == ItemType::Bow)
+				return _("fires fast speed arrows");
```
AJenbo, line 3575:
```suggestion
				return _("fires fast arrows");
```

PR #4109, opened 2022-03-01
DoHealOther converted to use GenerateRndSum

ikonomov:
DoHealOther spell function converted to use GenerateRndSum as the other spells

### DIFF
--- a/Source/spells.cpp
+++ b/Source/spells.cpp
@@ -309,9 +309,10 @@ void DoHealOther(int pnum, uint16_t rid)
 		return;
 	}
 
-	int hp = (GenerateRnd(10) + 1) << 6;
-	hp += GenerateRndSum(4, player._pLevel) + player._pLevel << 6;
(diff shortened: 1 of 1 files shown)


StephenCWills:
`GenerateRndSum()` is defined in the anonymous namespace in `missiles.cpp`, so it won't be available to `spells.cpp`. I would recommend moving the function to `engine/random.cpp`, then declare it in `engine/random.hpp` so it can be used in both `missiles.cpp` and `spells.cpp`.

AJenbo:
I think he still just edits things on the github website rather then use git locally.

ikonomov:
> I think he still just edits things on the github website rather then use git locally.

I tried using Git in Windows and found it more intimidating :(  I guess I have to be more persistent, will give it another shot.

AJenbo:
Did you try using GitHub Desktop or similar GUI tools?

PR #4110, opened 2022-03-02
Use Staff Spell level for Staff Recharge

kphoenix137:
Uses the Staff Spell level instead of Book Level. Using Book level results in some spells using -1 in the RNG, resulting in staves losing 50% of charges when recharged at 0 charges consistently, when the spell does not have a book counterpart. Performed testing with a clvl 50 Sorcerer. The results were very similar between changed and unchanged, however recharging a Staff of Resurrect was far less penalizing, and recharging a Staff of Apocalypse was slightly less penalizing. Tests were done in Diablo not Hellfire.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4536,7 +4536,7 @@ void RechargeItem(Item &item, Player &player)
 	if (item._iCharges == item._iMaxCharges)
 		return;
 
-	int r = GetSpellBookLevel(item._iSpell);
+	int r = GetSpellStaffLevel(item._iSpell);
 	r = GenerateRnd(player._pLevel / r) + 1;
 
 	do {


qndel:
How about 
image

DakkJaniels:
Seems more like a mod change than a bug fix. There is nothing wrong with how the code works, originally or now. Also not sure Diablo needs to be made easier, it's easy enough as it is.

StephenCWills (approved):
To me, this change just makes sense. The intent seems to be that the higher the level of spell on the staff, the higher the complexity for recharging should be, resulting in fewer charges restored per iteration and more total charges lost as a result. A value of -1 doesn't make much sense and only incidentally makes the spell as difficult as possible to recharge.

The spell staff level is a natural choice since the recharge staff skill applies to staves and not books. Given the very low impact of this change, I don't really see a compelling reason to preserve the inexplicable mismatch.

AJenbo (approved):

PR #4116, opened 2022-03-05
Filter out ZeroTier packets that do not come from an expected source

StephenCWills:
This change should reduce ZT errors due to erroneously interpreting random network traffic as an in-game packet. In cases where the player index is included as the packet source, this merely validates that the endpoint of the client who sent the packet matches the one that we would send packets to when communicating with the player at that index. FYI, this means that packets cannot be sent on behalf of another player, but that sort of delegation shouldn't be necessary since all players should have a direct channel to communicate between each other.

I believe this should help to address crashes similar to the one reported in #3263.

### DIFF
--- a/Source/dvlnet/base_protocol.h
+++ b/Source/dvlnet/base_protocol.h
@@ -288,6 +288,9 @@ void base_protocol<P>::recv_ingame(packet &pkt, endpoint sender)
 		// normal packets
 		LogDebug("Invalid packet: packet source ({}) >= MAX_PLRS", pkt.Source());
 		return;
+	} else if (sender != peers[pkt.Source()]) {
+		LogDebug("Invalid packet: packet source ({}) received from unrecognized endpoint", pkt.Source());
+		return;
 	}
 	connected_table[pkt.Source()] = true;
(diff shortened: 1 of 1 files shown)


AJenbo:
Since this is done in the base layer it would also affect TCP networks, won't that cause issues there as the host is broadcasting all other players info?

StephenCWills:
No, this doesn't affect TCP at all. The `tcp_client` class derives directly from `base`. The `base_protocol` class is actually not used as a base class, but would instead be more appropriately named as the `zerotier_client` class.

AJenbo:
Confusing :)

AJenbo (approved):

ephphatha:
Immediately after this check is an assignment `peers[pkt.Source()] = sender` which is now a no-op, this value is set when a player joins a game in progress thanks to the `PT_CONNECT` broadcast message in `handle_join_request`. Can remove that assignment.

StephenCWills:
`PT_CONNECT` is not a broadcast message. It's only sent to the player who joined. I think that line of code should still be necessary for peers that did not handle `PT_JOIN_REQUEST`.

PR #4117, opened 2022-03-05
monstdat cleanup

kphoenix137:


qndel:
I wouldn't go that far

kphoenix137:
> I wouldn't go that far

?

Source/monstdat.h
```diff
@@ -222,9 +209,6 @@ enum _monster_id : int16_t {
 	MT_BTBLACK,
 	MT_RBLACK,
 	MT_UNRAV,
-	MT_HOLOWONE,
-	MT_PAINMSTR,
-	MT_REALWEAV,
```
AJenbo, line 227: Deleting thease will shift the ids which is probably going to break a ton of things?

kphoenix137: I tested it out, no crashes or issues as far as I can tell.

AJenbo: Did you load a single player savegame?

kphoenix137: Just teleported through all crypted, killed na krul. all monsters maintained the correct data and names

AJenbo: I would also expect this to break loading some quest levels.

kphoenix137: > Did you load a single player savegame?

no, let me test

kphoenix137: Just tested, no issues

kphoenix137: > I would also expect this to break loading some quest levels.

I will test

kphoenix137: No crash in Leoric's Tomb. Should I test more levels?

qndel: aren't most of these monsters fully functional? Removing working stuff like this isn't beneficial at all

kphoenix137:
I'm going to create an enum class for MonsterID so this pr won't be needed

PR #4118, opened 2022-03-05
_pSpellFlag enums

DakkJaniels:
Added enums for the spell flags that are used. Used class format as discussed in comments.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -1047,7 +1047,7 @@ bool PlrHitPlr(int pnum, int8_t p)
 		return false;
 	}
 
-	if ((target._pSpellFlags & SF_ETHEREALIZE) != 0) {
+	if (HasAnyOf(target._pSpellFlags, SpellFlag::Etherealize)) {
 		return false;
 	}
 
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -1047,7 +1047,7 @@ bool PlrHitPlr(int pnum, int8_t p)
 		return false;
 	}
 
-	if ((target._pSpellFlags & 1) != 0) {
+	if ((target._pSpellFlags & SF_ETHEREALIZE) != 0) {
 		return false;
 	}
 
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2640,7 +2640,7 @@ void CreatePlayer(int playerId, HeroClass c)
 		spellLevel = 0;
 	}
 
-	player._pSpellFlags = 0;
+	player._pSpellFlags = SpellFlag::None;
 
 	if (player._pClass == HeroClass::Sorcerer) {
 		player._pSplLvl[SPL_FIREBOLT] = 2;
--- a/Source/player.h
+++ b/Source/player.h
@@ -780,5 +780,4 @@ extern int DexterityTbl[enum_size<HeroClass>::value];
 extern int VitalityTbl[enum_size<HeroClass>::value];
 extern uint32_t ExpLvlsTbl[MAXCHARLEVEL + 1];
 
-
 } // namespace devilution
--- a/Source/player.h
+++ b/Source/player.h
@@ -155,6 +155,7 @@ enum class SpellFlag : uint8_t {
 	// bits 3-7 are unused
 	// clang-format on
 };
+use_enum_as_flags(SpellFlag);
 
 /** Maps from armor animation to letter used in graphic files. */
 constexpr std::array<char, 4> ArmourChar = {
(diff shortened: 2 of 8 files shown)


Source/player.h
```diff
@@ -147,6 +147,19 @@ enum action_id : int8_t {
...
+enum spell_flag : uint8_t {
+	// clang-format off
+	SF_ETHEREALIZE      = 1 << 0,
+	SF_RAGE_ON          = 1 << 1,
+	SF_RAGE_COOLDOWN    = 1 << 2,
+	SF_NOTHING3         = 1 << 3,
+	SF_NOTHING4         = 1 << 4,
+	SF_NOTHING5         = 1 << 5,
+	SF_NOTHING6         = 1 << 6,
+	SF_NOTHING7         = 1 << 7,
```
AJenbo, line 159: There isn't much point to these as they are at the end.
```suggestion
```

Source/player.cpp
```diff
@@ -1047,7 +1047,7 @@ bool PlrHitPlr(int pnum, int8_t p)
 		return false;
 	}
 
-	if ((target._pSpellFlags & 1) != 0) {
+	if ((target._pSpellFlags & SF_ETHEREALIZE) != 0) {
```
AJenbo, line 1050: This makes it clear what need to be remove :D

Source/player.h
```diff
@@ -147,6 +147,19 @@ enum action_id : int8_t {
...
+	// clang-format on
+};
```
ephphatha, line 161: New enums should be scoped (use `enum class` in this case) and use TitleCase for the enum and values (also don't need to define every value in an enum when they're unused, a comment is fine).
```suggestion
enum class SpellFlag : uint8_t {
	// clang-format off
	Etherealize  = 1 << 0,
	RageActive   = 1 << 1,
	RageCooldown = 1 << 2,
	// bits 3-7 are unused
	// clang-format on
};
```

Uses of the values will need to be updated to `SpellFlag::Etherealize` instead of `SF_ETHEREALIZE`.

`Player::_pSpellFlags` should also use this enum as the backing type, then you can use the `use_enum_as_flags` macro to make `HasAnyOf`/`HasNoneOf` (from utils/enum_traits.h) available so tests can be expressed as `if(HasAnyOf(target._pSpellFlags, SpellFlag::Etherealize)) { ... }`

Source/player.cpp
DakkJaniels: haha yeah that's what I figured 😄

Source/player.h
DakkJaniels: I'll update accordingly, thanks!

DakkJaniels: @ephphatha do you still set and clear the flags the same way, or is there a different way to do that?

AJenbo:
Looks like you also need to update loadsave.cpp:
`player._pSpellFlags = file.NextLE<uint8_t>();`
`player._pSpellFlags = static_cast<SpellFlag>(file.NextLE<uint8_t>());`

DakkJaniels:
> Looks like you also need to update loadsave.cpp: `player._pSpellFlags = file.NextLE<uint8_t>();` `player._pSpellFlags = static_cast<SpellFlag>(file.NextLE<uint8_t>());`
> 
> `file.WriteLE<uint8_t>(player._pSpellFlags);` `file.WriteLE<uint8_t>(static_cast<uint8_t>(player._pSpellFlags));`

yeah I just realized that and made those changes, but I'm still having a problem building it. Let me upload and see if it it fails the tests here too...

AJenbo:
I forget if there is a limit about enum classs that prevents you from using them for bit wise operations (unless you cast them first). Or if that was another usage where we generally avoid it. If that is the case then i suggest reverting to just a simple enum.

Source/player.h
ephphatha: Can still use bitwise operators to set and clear flags, there are template specialisations defined for `&`, `|`, `~` that become available when using the `use_enum_as_flags` macro.

ephphatha:
The helper is being used already, bitwise operators are available :D. The current error is assigning a literal 0 in player.cpp `CreatePlayer`, need to define a `SpellFlags::None` value.

DakkJaniels:
> The helper is being used already, bitwise operators are available :D. The current error is assigning a literal 0 in player.cpp `CreatePlayer`, need to define a `SpellFlags::None` value.

Ah... makes sense. I'll fix it. Thanks!

Source/player.h
```diff
@@ -769,4 +780,5 @@ extern int DexterityTbl[enum_size<HeroClass>::value];
 extern int VitalityTbl[enum_size<HeroClass>::value];
 extern uint32_t ExpLvlsTbl[MAXCHARLEVEL + 1];
 
+
```
ephphatha, line 783:
```suggestion
```
Sneaky whitespace 😄

AJenbo:
```suggestion
```

AJenbo: Lol you beat me by 9 sec :)

AJenbo:
@DakkJaniels thanks for taking this all the way :)

AJenbo (approved):

ephphatha (approved):

Issue #4098, opened 2022-02-25
Multiplayer sync problems


kphoenix137:
**Important information**
Windows 10 / c6abd740 / compiled myself
**Describe**
In both ZeroTier and TCP:
Duplicate items drop in the dungeon, with the second one unable to be picked up for a brief time, and sometimes can be picked up later, or be picked up by the other person.
Sometimes items drop and are unable to be picked up by either player, and then vanish.
Frequent hourglass which is not seen in vanilla.

**To Reproduce**
Unable to reproduce, other than by standard gameplay co-op MP.

**Expected behavior**
Hourglass should occur at same rate as vanilla (which is rare) or less frequently.

**Screenshots**
[If applicable, add screenshots to help explain your problem.](https://www.youtube.com/watch?v=aDNQJ6g6CuE)

AJenbo:
This seams to be listing several issues, some I believe are vanilla, some can depend on your specific setup and there is little detail regarding the environment or what can affect it in various ways. As such it's going to be hard to work on this as an issue.

Also the video is over 8 minutes long, please post time codes for the relevant parts of it.

AJenbo:
Both issues should now be resolved.

PR #4122, opened 2022-03-06
Apply sequence number to net turns

StephenCWills:
This PR attempts to resolve "frequent hourglass which is not seen in vanilla" reported as part of #4098.

The idea behind this PR is to modify the network provider to prevent "turn desync" that occurs due to the unpredictable timing of packet arrival on each client. I plan to follow up with a much more detailed description of what's happening and what the PR does to fix it. It's a big, long writeup so I thought it might be good to break it into separate posts this time.

FYI, I am a little worried that these changes may prevent players with very high latencies from joining each other's games because the guest's first turn may get skipped by the host. I'm thinking about a couple ways to fix that, but I'm not entirely sure if it's really a problem just yet.

### DIFF
--- a/Source/dvlnet/base.cpp
+++ b/Source/dvlnet/base.cpp
@@ -41,7 +41,7 @@ void base::HandleAccept(packet &pkt)
 	}
 	if (pkt.Cookie() == cookie_self) {
 		plr_self = pkt.NewPlayer();
-		connected_table[plr_self] = true;
+		Connect(plr_self);
 	}
 	if (game_init_info != pkt.Info()) {
 		if (pkt.Info().size() != sizeof(GameData)) {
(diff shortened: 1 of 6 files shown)


StephenCWills:
Turns are processed in groups, where one turn from every client is added to the group and processed simultaneously. One group is processed every 8 frames, and the hourglass appears whenever a client attempts to process a turn from a remote client before it has been received over the network.

Between two clients, the desync we're trying to solve looks something like this.

#### Client 1:
* [Frame 0] Local turn: 52, Remote turn: 53
* [Frame 8] Local turn: 53, Remote turn: 54
* [Frame 16] Local turn: 54, Remote turn: 55
* [Frame 32] Local turn: 55, Remote turn: 56
* [Frame 48] Local turn: 56, Remote turn: 57

#### Client 2:
* [Frame 0] Local turn: 53, Remote turn: 53
* [Frame 8] Local turn: 54, Remote turn: 54
* [Frame 16] Local turn: 55, Remote turn: 55
* [Frame 32] Local turn: 56, Remote turn: 56
* [Frame 48] Local turn: 57, Remote turn: 57

In this example, you can see that the turns are being grouped differently by each client. In other words, the two clients do not process client 1's turns at the same time, relative to the frame number. For the record, client 2 is using the correct grouping in this scenario.

StephenCWills:
This PR applies a sequence number to every turn that acts as a sort of "group ID". Using the sequence number, every client can agree on which turns should be grouped and processed simultaneously. There are three parts to making this work.

When a player joins a game, their client must first receive a turn from another client to learn what sequence number to apply to the first turn they send. Because much of the logic surrounding turns is handled by the game logic on strict 8-frame cycles, the network layer has to queue the first local turn without sending it, then proceed to apply the sequence number and send it as soon as the first remote turn arrives. However, there is one edge case. If the client creates a game, there will be no remote client from which to receive a sequence number so it also needs to ensure that it doesn't get stuck waiting for one.

When a client processes a turn, it needs to make sure all the remote turns' sequence numbers match the next local turn's sequence number. Therefore, a client may skip processing a remote turn if its sequence number indicates that it should have already been processed. Furthermore, if a client receives a remote turn with a sequence number that indicates it should be processed in the future, the client will fill in zeros for the remote client's turns until that sequence number is reached.

StephenCWills:
Now let's take a look at why the desync is an issue. For this, take the following two considerations into account.

* A client cannot process a remote turn until it has been sent by the remote client and subsequently received by the local client.
* A client does not send a local turn to remote clients until the frame after processing the previous turn.

Given these, the desync creates a strict order of processing between the two clients that shouldn't normally exist. If you look at the desync example I gave previously, client 1 must process turn 52 before it can send 53 so that client 2 can process turn 53. Client 2 must process 53 before it can send 54 so that client 1 can process turn 53. If you keep extending this sequence, it's possible to assert the exact sequence in which turns were processed over time between the two clients. In this example, the order in which clients must process their local turns looks like this.

* Client 1 processes 52
* Client 2 processes 53
* Client 1 processes 53
* Client 2 processes 54
* Client 1 processes 54
* etc...

Based on the information above, we can map out the exact sequence of events that must occur during the 8 frames between processing turns 52 and 53 on client 1. Keep in mind that each of these steps depends directly on the previous one.

* Client 1 processes 52
* Client 1 sends 53
* Client 2 receives 53
* Client 2 processes 53
* Client 2 sends 54
* Client 1 receives 54
* Client 1 processes 53

The key point here is that the turn packet must be transmitted between each turn that gets processed. Because turns are "scheduled" based on the fact that they are necessarily processed every 8 frames, the actual latency requirement depends on the difference in time between client 1 processing and client 2 processing. The extra dependencies that enforce this strict sequencing introduce delays that lead to more frequent hourglass occurrences. This is only exacerbated when more than two clients are involved.

If you go through the same exercise in a scenario where there is no desync, you should find that there is no way to assert the exact order in which the two clients process their turns. The only constraint is that a client must receive the remote turn 53 before it can process local turn 53, meaning that the remote client definitely processed turn 52 and may or may not have processed 53. In practice, the order in which turns get processed should still be very consistent--because of the 8-frame timing--but also largely irrelevant. The hourglass should only appear if a change in processing delays or network latencies causes a packet to arrive much later than usual.

StephenCWills:
Rebased to include 3b3419d for the purposes of testing on ZeroTier.

AJenbo (dismissed):
This sounds reasonable, but not easy to give good feedback on based. @Eider-McDuck might have some insight but not sure.

From my perspective I think the best way to evaluate this is to merge it so that it can be play tested and we can then get some feedback on things from there.

AJenbo (approved):

PR #4123, opened 2022-03-06
switch affix_item_type bit flag enum to class format

DakkJaniels:
Same as that spellflag enum, switched over the affix_item_type to the same format.


Source/items.cpp
```diff
@@ -1201,7 +1201,7 @@ void GetItemPower(Item &item, int minlvl, int maxlvl, affix_item_type flgs, bool
 				continue;
 			if (onlygood && !ItemPrefixes[j].PLOk)
 				continue;
-			if (flgs == PLT_STAFF && ItemPrefixes[j].power.type == IPL_CHARGES)
+			if (HasAnyOf(flgs,AffixItemType::Staff) && ItemPrefixes[j].power.type == IPL_CHARGES)
```
AJenbo, line 1204:
```suggestion
			if (HasAnyOf(flgs, AffixItemType::Staff) && ItemPrefixes[j].power.type == IPL_CHARGES)
```

DakkJaniels: yup just noticed it. I wish Visual Studio would fix that before it uploads the commit...

Source/itemdat.cpp
```diff
@@ -226,94 +226,94 @@ const PLStruct ItemPrefixes[] = {
 	// clang-format off
 	// PLName,              { type,                 param1, param2 }, PLMinLvl, PLIType,                                                         PLGOE,    PLDouble, PLOk,  minVal, maxVal, multVal
```
AJenbo, line 227: Probably the header needs some adjustment now?

DakkJaniels: missed that - fixed

AJenbo (approved):

PR #4125, opened 2022-03-06
Dedupe network call when dropping items on death

ephphatha:
This lets us use `std::move` to indicate that the data is discarded after use in these functions, the one time we reuse the variable (when dropping equipped items in `StartPlayerKill`) it's "reinitialised" by setting `_itype` to indicate an empty item.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -536,7 +536,7 @@ void StartSpell(int pnum, Direction d, int cx, int cy)
 	player.spellLevel = GetSpellLevel(pnum, player._pSpell);
 }
 
-void RespawnDeadItem(Item &itm, Point target)
+void RespawnDeadItem(Item &&itm, Point target)
 {
 	if (ActiveItemCount >= MAXITEMS)
 		return;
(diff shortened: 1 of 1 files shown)


Source/player.cpp
```diff
@@ -3143,7 +3142,7 @@ StartPlayerKill(int pnum, int earflag)
 			drawhpflag = true;
 
 			if (pcurs >= CURSOR_FIRSTITEM) {
-				DeadItem(player, player.HoldItem, { 0, 0 });
+				DeadItem(player, std::move(player.HoldItem), { 0, 0 });
```
AJenbo, line 3145: I'm a bit confused by this, is it safe to treat a global like this?

ephphatha: In this case it's fine, as far as I understand `std::move` lets the compiler treat the variable with less concern about what happens to the source memory. For msvc after this function resolves the variable contains garbage. The next line then sets the `_itype` member to a value that means `player.HoldItem.isEmpty()` returns true, and most other normal uses of the HoldItem variable use this check to see if there is an item held. I say most, because some checks use `pcurs` instead which gets set by the call to `NewCursor()` two lines down.

AJenbo: Thanks for the explanation, with the other cleanup this is also reduced to this case which is the more clean of the two that where there previously.

PR #4129, opened 2022-03-06
Item special effect enum

DakkJaniels:
Same as other two PR's. Switched all the Item Special Effects bit flags to use the class enum format.


DakkJaniels:
I guess Amiga doesn't like swapping SDL_SwapLE32 unless I cast it each way? Let's find out...

Source/items.h
```diff
@@ -182,65 +182,65 @@ struct Item {
 	 * @brief Contains Information for current Animation
 	 */
 	AnimationInfo AnimInfo;
-	bool _iDelFlag = false; // set when item is flagged for deletion, deprecated in 1.02
```
ephphatha, line 185: The default values for these members are to help ensure items are default-initialised appropriately, where possible they should be retained.

DakkJaniels: weird... I didn't make any of those changes. Is that a clang format thing?

AJenbo: It was a recent addition so more likely it was an improper merge/rebase when you created the branch. clang format is always applied so It wouldn't have been possible to add the defaults in the first place if that was the case.

DakkJaniels: ah ok - I'll fix it

Source/utils/enum_traits.h
```diff
@@ -106,6 +106,12 @@ constexpr bool HasAnyOf(EnumType lhs, EnumType test)
 	return (lhs & test) != static_cast<EnumType>(0); // Some flags enums may not use a None value outside this check so we don't require an EnumType::None definition here.
 }
 
+template <typename EnumType, std::enable_if_t<std::is_enum<EnumType>::value && is_flags_enum<EnumType>::value, bool> = true>
+constexpr bool HasAllOf(EnumType lhs, EnumType test)
+{
+	return (lhs & test) == test; // Some flags enums may not use a None value outside this check so we don't require an EnumType::None definition here.
```
ephphatha, line 112:
```suggestion
	return (lhs & test) == test;
```
That comment is only really relevant for the anyof check

test/pack_test.cpp
```diff
@@ -234,98 +234,98 @@ const ItemPack PackedDiabloItems[] = {
...
+	//_iIName,                        _itype,                _iClass, _iCurs, _iIvalue, _iMinDam, _iMaxDam, _iAC,                                    _iFlags, _iMiscId, _iSpell, _iCharges, _iMaxCharges, _iDurability, _iMaxDur, _iPLDam, _iPLToHit, _iPLAC, _iPLStr, _iPLMag, _iPLDex, _iPLVit, _iPLFR, _iPLLR, _iPLMR, _iPLMana, _iPLHP, _iPLDamMod, _iPLGetHit, _iPLLight, _iSplLvlAdd, _iUid, _iFMinDam, _iFMaxDam, _iLMinDam, _iLMaxDam, _iPrePower, _iSufPower, _iMinStr, _iMinMag, _iMinDex, IDidx );
+	{ "Amber Helm of harmony",        ItemType::Helm,              2,     98,    21100,        0,        0,   11,    static_cast<ItemSpecialEffect>(8388608),        0,       0,         0,            0,           60,       60,       0,         0,      0,       0,       0,       0,       0,     18,     18,     18,        0,      0,          0,          0,         0,           0,     0,         0,         0,         0,         0,         11,         59,       50,        0,        0,    53 },
```
ephphatha, line 238: These values are combinations of flags, should be able to compose them from the named enum values.

test/player_test.cpp
```diff
@@ -29,13 +29,13 @@ int RunBlockTest(int frames, int flags)
...
-#define BALHAR (ISPL_FASTRECOVER | ISPL_FASTESTRECOVER)
-#define STAHAR (ISPL_FASTERRECOVER | ISPL_FASTESTRECOVER)
-#define ZEN (ISPL_FASTRECOVER | ISPL_FASTERRECOVER | ISPL_FASTESTRECOVER)
+#define BAL static_cast<uint32_t>(ItemSpecialEffect::FastHitRecovery)
+#define STA static_cast<uint32_t>(ItemSpecialEffect::FasterHitRecovery)
+#define HAR static_cast<uint32_t>(ItemSpecialEffect::FastestHitRecovery)
+#define BALSTA static_cast<uint32_t>(ItemSpecialEffect::FastHitRecovery | ItemSpecialEffect::FasterHitRecovery)
+#define BALHAR static_cast<uint32_t>(ItemSpecialEffect::FastHitRecovery | ItemSpecialEffect::FastestHitRecovery)
+#define STAHAR static_cast<uint32_t>(ItemSpecialEffect::FasterHitRecovery | ItemSpecialEffect::FastestHitRecovery)
+#define ZEN static_cast<uint32_t>(ItemSpecialEffect::FastHitRecovery | ItemSpecialEffect::FasterHitRecovery | ItemSpecialEffect::FastestHitRecovery)
```
ephphatha, line 38: Could be worth changing these to constexpr values and making `BlockData` a collection of structs instead of a 2d int array. That'd save having to put the cast in these defines and then invert it inside `RunBlockTest`, could take `ItemSpecialEffect` as an arg there.

```cpp
struct BlockTestCase {
    int expectedRecoveryFrame;
    int maxRecoveryFrame;
    ItemSpecialEffect itemFlags;
}

BlockTestCase BlockData[] = { ... };
```

I wouldn't say it's necessary for this PR unless you wanted to have a go, definitely good for a followup in the near future.

Source/items.h
```diff
@@ -197,7 +197,7 @@ struct Item {
 	uint8_t _iMinDam = 0;
 	uint8_t _iMaxDam = 0;
 	int16_t _iAC = 0;
-	uint32_t _iFlags = 0; // item_special_effect
+	ItemSpecialEffect _iFlags = ItemSpecialEffect::None; // ItemSpecialEffect flags
```
ephphatha, line 200:
```suggestion
	ItemSpecialEffect _iFlags = ItemSpecialEffect::None;
```
This comment doesn't add any value now that you've given the member a proper type :)

ephphatha (commented):
Looks good, the only real hangup I've got is about the use of magic numbers in the pack/unpack tests. Expressing those as flag combinations would be ideal.

Source/items.h
```diff
@@ -240,7 +240,7 @@ struct Item {
 	bool _iStatFlag = false;
 	_item_indexes IDidx = IDI_NONE;
 	uint32_t dwBuff = 0;
-	uint32_t _iDamAcFlags = 0;
+	ItemSpecialEffectHf _iDamAcFlags = ItemSpecialEffectHf::None;
```
glebm, line 243: If this is a *set* of flags then this is not the correct type for it.
`std::underlying_type<ItemSpecialEffectHf>::type` is more appropriate.

Typing a set as the enum type breaks static analysis of switch statements.

test/pack_test.cpp
DakkJaniels: Yeah, I was a bit lazy on this one, but also was concerned about breaking the table for a few items with a bunch of values. This keeps it a little more compact. If we really want it converted to the actual flag values, I guess I can script it.

Source/items.h
DakkJaniels: Can you clarify what I'm supposed to do here? Not familiar with this one, and it looks like it doesn't like it when I rename do the variable declarations accordingly: 
image

ephphatha: I was under the impression that using the enum as the type was suitable for flags enums given the expected values fall within the expected range of the enum (or the underlying type is fixed, which bypasses the issue entirely). Using a flag enum in a switch statement seems wrong anyway, it'd only test for an exact value instead of whether a given flag is set?

Using the underlying type directly makes it more difficult to assign/set/remove flags but that could be worked around without much drama if we wanted to go that way.

glebm: I suppose it's OK to use the enum class directly if it's verbose otherwise. This does mean it is not possible to distinguish a single value from a set based on the type alone, but I guess `flags` is there in the name so it's not a major issue in practice.

DakkJaniels: Sounds good. I'll leave as is.

test/pack_test.cpp
ephphatha: Looks like only Griswold's Edge and Empyrean Band use more than one flag, giving them named constants would let the table be a list of single flag values otherwise. That'd work fine.

DakkJaniels: done, although there were 4 more in the other table too 🙂

test/player_test.cpp
DakkJaniels: Done

test/pack_test.cpp
```diff
@@ -232,100 +232,103 @@ const ItemPack PackedDiabloItems[] = {
 	// clang-format on
 };
 
+#define EmpyreanBandSpecialEffect (ItemSpecialEffect::FastHitRecovery | ItemSpecialEffect::HalfTrapDamage)
+#define GrisworldEdgeSpecialEffect (ItemSpecialEffect::FireDamage | ItemSpecialEffect::Knockback | ItemSpecialEffect::FastAttack)
```
ephphatha, line 236:
```suggestion
constexpr ItemSpecialEffect EmpyreanBandSpecialEffect = ItemSpecialEffect::FastHitRecovery | ItemSpecialEffect::HalfTrapDamage;
constexpr ItemSpecialEffect GrisworldEdgeSpecialEffect = ItemSpecialEffect::FireDamage | ItemSpecialEffect::Knockback | ItemSpecialEffect::FastAttack;
```

test/pack_test.cpp
```diff
@@ -559,97 +562,102 @@ const ItemPack PackedHellfireItems[] = {
 	// clang-format on
 };
 
+#define GnatStingSpecialEffect (ItemSpecialEffect::MultipleArrows | ItemSpecialEffect::QuickAttack)
+#define ThunderclapSpecialEffect (ItemSpecialEffect::FireDamage | ItemSpecialEffect::LightningDamage)
+#define FlambeauSpecialEffect (ItemSpecialEffect::FireArrows | ItemSpecialEffect::LightningArrows)
+#define BlitzenSpecialEffect (ItemSpecialEffect::FireArrows | ItemSpecialEffect::LightningArrows)
```
ephphatha, line 568:
```suggestion
constexpr ItemSpecialEffect GnatStingSpecialEffect = ItemSpecialEffect::MultipleArrows | ItemSpecialEffect::QuickAttack;
constexpr ItemSpecialEffect ThunderclapSpecialEffect = ItemSpecialEffect::FireDamage | ItemSpecialEffect::LightningDamage;
constexpr ItemSpecialEffect ExplosiveArrows = ItemSpecialEffect::FireArrows | ItemSpecialEffect::LightningArrows;
```

Could even rename the constant for thunderclap to "ExplosiveDamage".

test/player_test.cpp
```diff
@@ -8,13 +8,13 @@ namespace devilution {
...
 
-int RunBlockTest(int frames, int flags)
+int RunBlockTest(int frames, ItemSpecialEffect flags)
 {
 	int pnum = 0;
 	auto &player = Players[pnum];
 
 	player._pHFrames = frames;
-	player._pIFlags = flags;
+	player._pIFlags = static_cast<ItemSpecialEffect>(flags);
```
ephphatha, line 17: this cast is unnecessary now, flags is already of type `ItemSpecialEffect`
```suggestion
	player._pIFlags = flags;
```

ephphatha (dismissed):
Everything I was even slightly hesitant about has been addressed, all good from my perspective 😄

Source/items.h
```diff
@@ -182,6 +182,7 @@ struct Item {
 	 * @brief Contains Information for current Animation
 	 */
 	AnimationInfo AnimInfo;
+
```
qndel, line 185: extra new line

PR #4145, opened 2022-03-09
Expand monster resists

DakkJaniels:
Rewrote monster resistances for expandability by mods in the future:

1. Monster resist struct has individual int8's for each resistance, allowing any value from 0-127 for magic, fire, and lightning, while acid is restricted to multiplies of 16. This is to maintain save compatibility, which required grabbing two adjacent free spots. If save compatibility doesn't matter, the 4 resistances could fit into a 32-bit space - or alternatively, maybe the monster resistances could be restored from the MonstersData table when loading a save file - let me know what you think).
2.  Added member functions to the monster resist struct to help with the conditional logic checks on resistances/immunities
3. Expanded out the monster table so you could have different resistances for Normal and Nightmare. For this PR, the normal and nightmare values are the same per the original game.

Note that IMMUNE_NULL_40 was not included as part of this PR since it isn't used anywhere else in the game, as far as I know. I included IMMUNE_ACID since there is a logic check on that with missiles in Hellfire.


Source/missiles.cpp
```diff
@@ -200,22 +200,24 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
...
+	} else if (mor.isFireResistant() && mir == MISR_FIRE) {
+		resist = mor.getFireResist();
+	} else if (mor.isLightningResistant() && mir == MISR_LIGHTNING) {
+		resist = mor.getLightningResist();
+	}
 
-	if (gbIsHellfire && t == MIS_HBOLT && (monster.MType->mtype == MT_DIABLO || monster.MType->mtype == MT_BONEDEMN))
-		resist = true;
+	if (gbIsHellfire && t == MIS_HBOLT && (monster.MType->mtype == MT_DIABLO || monster.MType->mtype == MT_BONEDEMN)) {
+		resist = 75;
```
yuripourre, line 219:
```suggestion
		resist = MAX_RESIST;
```

DakkJaniels: I don't think the intent was that Diablo and the Bone Demon have resistance against holy bolt equal to the maximum resistance the player can have. If we don't want a "magic number" here, it should probably be called something else so those two things aren't tied together.

yuripourre: It was only a suggestion, I didn't go too deep into the change you are proposing. If doesn't make sense, please ignore my comment.

DakkJaniels: No worries, I was just thinking, I shouldn't have said intent since it's not like they intended for all of this stuff to be modded when they first made Diablo anyway haha - but more from the perspective of it I was modding things, I wouldn't necessarily want those two things linked together, since I might not realize changing `MAX_RESIST` would effect those two things. Regardless, thanks for taking a look 🙂

yuripourre: Yes, I totally got your argument and agree. No hard feelings.

qndel:
`. This also includes switching MonsterAI to an enumerated class because I forgot to making a new branch when I started work on this.`

how about no ... lol

DakkJaniels:
If that is really a big deal, I will undo it.

qndel:
> If that is really a big deal, I will undo it.

If you dump multiple things into a single PR like this ,expect it to stay in review forever ;)

DakkJaniels:
> If you dump multiple things into a single PR like this ,expect it to stay in review forever ;)

No problem, fixed! Thank you for the insight 😄

Source/monstdat.h
```diff
@@ -61,17 +61,299 @@ enum class MonsterClass : uint8_t {
...
+	 */
+	bool isAcidImmune() const
+	{
+		return this->Acid >= 100;
+	}
+
+	/**
+	 * @brief Sets all monster resistances to 0.
+	 */
+	void setToZero()
```
qndel, line 190: name doesn't explain anything (should contain the word resistances or something)

DakkJaniels: Considering you have to be calling it from something directly related to resistances, I disagree that this is entirely unclear, however I will modify accordingly. Thank you!

qndel: nvm, I didn't see it was used for resistances only :D a bit weird design

Source/missiles.cpp
```diff
@@ -200,22 +200,24 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
...
 
-	uint8_t mor = monster.mMagicRes;
+	MonsterResists mor = monster.mResists;
 	missile_resistance mir = MissilesData[t].mResist;
 
-	if (((mor & IMMUNE_MAGIC) != 0 && mir == MISR_MAGIC)
-	    || ((mor & IMMUNE_FIRE) != 0 && mir == MISR_FIRE)
-	    || ((mor & IMMUNE_LIGHTNING) != 0 && mir == MISR_LIGHTNING)
-	    || ((mor & IMMUNE_ACID) != 0 && mir == MISR_ACID))
+	if ((mor.isMagicImmune() && mir == MISR_MAGIC) || (mor.isFireImmune() && mir == MISR_FIRE) || (mor.isLightningImmune() && mir == MISR_LIGHTNING) || (mor.isAcidImmune() && mir == MISR_ACID)) {
```
AJenbo, line 206: Rather then having several  functions I think it would be a lot cleaner to have an element enum and have `Resist::isImmune()` take it as a parameter. Else anyone who want to expand on the type will have to chance all lines like this down and add to them.

Source/missiles.cpp
```diff
@@ -200,22 +200,24 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
...
-	    || ((mor & RESIST_FIRE) != 0 && mir == MISR_FIRE)
-	    || ((mor & RESIST_LIGHTNING) != 0 && mir == MISR_LIGHTNING))
-		resist = true;
+	if (mor.isMagicResistant() && mir == MISR_MAGIC) {
+		resist = mor.getMagicResist();
+	} else if (mor.isFireResistant() && mir == MISR_FIRE) {
+		resist = mor.getFireResist();
+	} else if (mor.isLightningResistant() && mir == MISR_LIGHTNING) {
+		resist = mor.getLightningResist();
+	}
```
AJenbo, line 216: I don't see a reason to check before getting?

```suggestion
	resist = mor.getMagicResist(mir);
```

Also, try avoiding acronyms, especially custom ones. I have no idea what mor mean, other them mom in Danish.

Source/missiles.cpp
```diff
@@ -267,8 +269,10 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 
 	if (!shift)
 		dam <<= 6;
-	if (resist)
-		dam >>= 2;
+
+	assert(dam < (0x7FFFFFFF >> 6));
```
AJenbo, line 273: Probably use `std::numeric_limits<int>::max()`

Source/missiles.cpp
```diff
@@ -970,16 +974,16 @@ bool MonsterTrapHit(int m, int mindam, int maxdam, int dist, missile_id t, bool
...
 
 	missile_resistance mir = MissilesData[t].mResist;
-	int mor = monster.mMagicRes;
-	if (((mor & IMMUNE_MAGIC) != 0 && mir == MISR_MAGIC)
-	    || ((mor & IMMUNE_FIRE) != 0 && mir == MISR_FIRE)
-	    || ((mor & IMMUNE_LIGHTNING) != 0 && mir == MISR_LIGHTNING)) {
+	MonsterResists mor = monster.mResists;
+	if ((mor.isMagicImmune() && mir == MISR_MAGIC)
+	    || (mor.isFireImmune() && mir == MISR_FIRE)
+	    || (mor.isLightningImmune() && mir == MISR_LIGHTNING)) {
```
AJenbo, line 980: Maybe we can even avoid the distinction between resistance and immunity.
```suggestion
	if (mor.getResist(mir) >= 100) {
```

Source/missiles.cpp
DakkJaniels: I assume this is comment would require the element enum comment above to be implemented? I agree that is cleaner, I can implement.

Source/missiles.cpp
DakkJaniels: agreed, I didn't realize that was a thing at the time 🙂

Source/missiles.cpp
DakkJaniels: Yeah I think that makes sense, I'll modify.

Source/missiles.cpp
```diff
@@ -267,8 +269,10 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 
 	if (!shift)
 		dam <<= 6;
-	if (resist)
-		dam >>= 2;
+
+	assert(dam < (0x7FFFFFFF >> 6));
+	if (resist > 0)
+		dam = dam * (100 - resist) / 100;
```
AJenbo, line 275: I find this a bit less convoluted:
```suggestion
		dam -= dam * resist / 100;
```

Source/missiles.cpp
DakkJaniels: Also, even though you edited it out, I think mor was monster resists, and mir was missile resist? I'm not really sure, since they are original, but I can change them to something a bit more descriptive.

Source/missiles.cpp
DakkJaniels: I don't think those are equivalent? If you have 50 damage at 75% resist, the original would produce 12 damage, but I think the revised version would result in 13 damage (please check my math, I could be wrong)? I was trying to keep the result of 75% resist the same.

Source/monstdat.h
```diff
@@ -61,17 +61,299 @@ enum class MonsterClass : uint8_t {
...
+	}
+
+	/**
+	 * @brief Returns if monster has any immunities to Magic, Fire, or Lightning.
+	 * @return True if monster has non-acid immunities.
+	 */
+	bool hasImmunities() const
+	{
+		return this->Magic >= 100 || this->Fire >= 100 || this->Lightning >= 100;
+	}
```
AJenbo, line 113: This seams specific (ignores a specific element type) to the UI and probably shouldn't be part of the general class.

Source/monstdat.h
```diff
@@ -61,17 +61,299 @@ enum class MonsterClass : uint8_t {
...
+	uint8_t getLightningResist()
+	{
+		return this->Lightning;
+	}
+	uint8_t getAcidResist()
+	{
+		return this->Acid;
+	}
+
```
AJenbo, line 88: ... why actually have getters for the properties?

Source/monstdat.h
```diff
@@ -61,17 +61,299 @@ enum class MonsterClass : uint8_t {
...
+	/**
+	 * @brief Sets all monster resistances to 0.
+	 */
+	void setResistsToZero()
+	{
+		this->Magic = 0;
+		this->Fire = 0;
+		this->Lightning = 0;
+		this->Acid = 0;
+	}
```
AJenbo, line 196: This appears to be unique to Na-Krull, so probably just set them there.

Source/monstdat.h
```diff
@@ -61,17 +61,299 @@ enum class MonsterClass : uint8_t {
...
+	 * @brief Converts the monster resists to save info.
+	 * @return 32 bit int for saving purposes.
+	 */
+	uint32_t getSaveData()
+	{
+		/**
+		 * Save Structure (low to high):
+		 * 8 bits - original Diablo save data.
+		 * 3 bits - acid data.
+		 * 7 biits - lightning data.
```
AJenbo, line 208:
```suggestion
		 * 7 bits - lightning data.
```

Source/missiles.cpp
AJenbo: Yeah, I realized that it wasn't added by you so didn't want to direct it at you. But you are more then welcome to change it to something readable :)

Issue #4135, opened 2022-03-06
Extreme camera shaking after Diablo's death


Chance4us:
Everybody can experience extreme camera shaking after Diablo's death.
To reproduce, load the attached game and move the hero to the north direction.
This is also a vanilla issue.

https://user-images.githubusercontent.com/45902449/156942740-e0367611-2ae5-493c-815f-2903b3926c08.mp4

Savegame:
[shaking.zip](https://github.com/diasurgical/devilutionX/files/8193532/shaking.zip)

qndel:
That's because diablo's death disables movement but that lock gets removed when you load the game = you aren't supposed to move when he's dying :D

qndel:
No need for [vanilla] in title, we have labels for that ;)

qndel:
> I don't think [Chance4us](https://github.com/Chance4us) has the option to apply labels

that's what we are for 🤪

PR #4149, opened 2022-03-09
stop shaking camera if dead diablo loaded from save.

DakkJaniels:
Issue #4135


Source/loadsave.cpp
```diff
@@ -1974,6 +1974,11 @@ void LoadGame(bool firstflag)
 	}
 
 	gbIsHellfireSaveGame = gbIsHellfire;
+	if (Quests[Q_DIABLO]._qactive == QUEST_DONE) {
+		PlaySFX(USFX_DIABLOD);
+		gbProcessPlayers = false;
+		myPlayer._pmode = PM_QUIT;
```
obligaron, line 1980: Is this needed?
I thought `PrepDoEnding` will set this and then the game ends? 🤔

DakkJaniels: the game still ends, but the camera is shaking when the game is loaded. This stops it from shaking.

obligaron: I read the issue. My point was: is the line `myPlayer._pmode = PM_QUIT;` necessary or not.
Sorry if I didn't clarify good enough. 😢 

The program flow is this:
1. In [`DiabloDeath`](https://github.com/diasurgical/devilutionX/blob/a66ca446950bceb25ac31ee268dfc8e2ec9259ce/Source/monster.cpp#L955) `gbProcessPlayers` is set to false, but `PM_QUIT` isn't set.
2. Then the players saves the game.
3. `PM_QUIT` is later set in [`PrepDoEnding`](https://github.com/diasurgical/devilutionX/blob/a66ca446950bceb25ac31ee268dfc8e2ec9259ce/Source/monster.cpp#L4151) (or when starting a new game `GamemenuNewGame`).
4. After this the video starts and the game ends.

If we load the save game, we need to restore the game state to the save game state (as it was in 2.). That means we should only need to set `gbProcessPlayers` to false and don't need to set `PM_QUIT` at all. Because even when loading the save game `PrepDoEnding` later sets `PM_QUIT`.
That's why I raised the question. But perhaps I miss something.

DakkJaniels: If I recall correctly, without setting PM_QUIT at this point, the screen was still shaking on load.

DakkJaniels: > Because even when loading the save game PrepDoEnding later sets PM_QUIT.

Are you sure it will call `PrepDoEnding()` on load? I only see it called in two places, on monster death (`MonsterDeath` in monster.cpp, which already occurred by the point of save) and in `ShowProgress` in interfac.cpp which requires `gbSomebodyWonGameKludge == true` but it sets it to false right after running `PrepDoEnding()`

StephenCWills: `gbProcessPlayers = false` prevents `ProcessPlayers()` from being called, which is what freezes the character's animation. The pmode doesn't really matter at that point. My testing suggests that `PM_QUIT` is indeed not necessary.

`MonsterDeath()` is the function that gets called in `ProcessMonsters()` when the monster's mmode is `MonsterMode::Death`. Diablo's death animation, which is still playing at time of save/load in this scenario, would indeed lead to `PrepDoEnding()` on load.

DakkJaniels: I could have sworn I did all this without PM_QUIT to start with, and only added it after it didn't work. Let me run it again...

DakkJaniels: Without PM_QUIT being set, the issue (#4135) is not resolved, the screen still shakes (tbf it's more like rubberbands, just using the same language as the issue). Just ran it and confirmed.

DakkJaniels: Video:

https://user-images.githubusercontent.com/6080734/163580265-b08386c7-7237-42df-913c-5112a7f5e11c.mp4

StephenCWills: I see. This is a discrepancy between `Load Game` from the main menu vs the in-game `Load Game` menu option. In the former, `RunGameLoop()` invokes `StartGame()` which leads to `LoadGame()` and then sets `gbProcessPlayers = true` just a few lines later. `PM_QUIT` is a workaround that forces the player into a weird animation state.

DakkJaniels: Yeah - looking through it a bit more, I see that `PM_QUIT` should not have been set yet (I assume it's something like 7 seconds, in normal speed, after Diablo's death that `PrepDoEnding()` is called?) I'm having trouble following what is actually holding the player and screen in place when Diablo is dead under normal conditions.

StephenCWills: > I assume it's something like 7 seconds, in normal speed, after Diablo's death that PrepDoEnding() is called?

You got it.

> I'm having trouble following what is actually holding the player and screen in place when Diablo is dead under normal conditions.

The player is being held in place by the fact that `ProcessPlayers()` is not called during Diablo's death animation (the aforementioned `gbProcessPlayers = false`). The fact that the player can't do anything is also what holds the camera in place.

The reason `ProcessPlayers()` causes the camera to rubber band during Diablo's death animation because the player's walking animation updates the camera position based on the player's position.

https://github.com/diasurgical/devilutionX/blob/aab970dcdedf9a9ee900e19d227c223a2ddf1fd5/Source/player.cpp#L704-L707

And `ProcessMonsters()` causes the camera to pan toward Diablo's location at a rate of one tile per game tick.

https://github.com/diasurgical/devilutionX/blob/aab970dcdedf9a9ee900e19d227c223a2ddf1fd5/Source/monster.cpp#L1709-L1720

DakkJaniels: Thanks for the help/insight @StephenCWills and @obligaron. I moved the fix to `RunGameLoop` so the `PM_QUIT` workaround isn't needed.

obligaron:
As @StephenCWills [pointed out](https://github.com/diasurgical/devilutionX/pull/4149#discussion_r851291416) the game handles loading from main menu different then in-game menu / quick-load.
So it seems we need to apply this fix at two places.
Your original place in `LoadGame(bool firstflag)` when `firstflag` is false and the current location. 😕

DakkJaniels:
I like this a bit better now to minimize duplicating code - let me know if you think the function should go some place else.

DakkJaniels:
I don't know why amiga is failing, I don't think it has anything to do with the code? 
image

AJenbo:
@DakkJaniels outdated dependencies. Please update your branch against master.

Source/diablo.cpp
```diff
@@ -2364,4 +2364,14 @@ void diablo_color_cyc_logic()
 	}
 }
 
+bool isDiabloAlive(bool isMultiplayer, bool playSFX)
+{
+	if (Quests[Q_DIABLO]._qactive == QUEST_DONE && !isMultiplayer) {
```
obligaron, line 2369:
```suggestion
	if (Quests[Q_DIABLO]._qactive == QUEST_DONE && !gbIsMultiplayer) {
```

Source/diablo.cpp
```diff
@@ -2364,4 +2364,14 @@ void diablo_color_cyc_logic()
 	}
 }
 
+bool isDiabloAlive(bool isMultiplayer, bool playSFX)
```
obligaron, line 2367:
```suggestion
bool IsDiabloAlive(bool playSFX)
```

Source/diablo.cpp
```diff
@@ -726,7 +726,7 @@ void RunGameLoop(interface_mode uMsg)
 	saveProc = SetWindowProc(GameEventHandler);
 	run_delta_info();
 	gbRunGame = true;
-	gbProcessPlayers = true;
+	gbProcessPlayers = isDiabloAlive(gbIsMultiplayer, false);
```
obligaron, line 729:
```suggestion
	gbProcessPlayers = IsDiabloAlive(true);
```

Source/diablo.h
```diff
@@ -97,6 +97,7 @@ void diablo_focus_unpause();
 bool PressEscKey();
 void DisableInputWndProc(uint32_t uMsg, int32_t wParam, int32_t lParam);
 void LoadGameLevel(bool firstflag, lvl_entry lvldir);
+bool isDiabloAlive(bool isMultiplayer, bool playSFX);
```
obligaron, line 100:
```suggestion
bool IsDiabloAlive(bool playSFX);
```

Source/loadsave.cpp
```diff
@@ -2053,7 +2053,7 @@ void LoadGame(bool firstflag)
 	missiles_process_charge();
 	RedoMissileFlags();
 	NewCursor(CURSOR_HAND);
-	gbProcessPlayers = true;
+	gbProcessPlayers = isDiabloAlive(gbIsMultiplayer, true);
```
obligaron, line 2056:
```suggestion
	gbProcessPlayers = IsDiabloAlive(gbIsMultiplayer, !firstflag);
```

obligaron (commented):
Some nits, for example functions should start with upper-case.

obligaron:
I thought again about the function name `IsDiabloDead`.
Perhaps we should rename it to `InitProcessPlayers` and set `gbProcessPlayers` directly in the function? 🤔

obligaron (approved):
Let's get this in. 🙂 
The changed function name was only a nit.

Source/diablo.cpp
```diff
@@ -2364,4 +2364,14 @@ void diablo_color_cyc_logic()
...
 }
 
+bool IsDiabloAlive(bool playSFX)
+{
+	if (Quests[Q_DIABLO]._qactive == QUEST_DONE && !gbIsMultiplayer) {
+		if (playSFX)
+			PlaySFX(USFX_DIABLOD);
+		return false;
+	} else
+		return true;
```
AJenbo, line 2374:
```suggestion
	}

	 return true;
```

Source/diablo.cpp
```diff
@@ -2364,4 +2364,15 @@ void diablo_color_cyc_logic()
...
 
+bool IsDiabloAlive(bool playSFX)
+{
+	if (Quests[Q_DIABLO]._qactive == QUEST_DONE && !gbIsMultiplayer) {
+		if (playSFX)
+			PlaySFX(USFX_DIABLOD);
+		return false;
+	}
+
+	 return true;
```
AJenbo, line 2375:
```suggestion
	return true;
```

PR #4152, opened 2022-03-11
Players[MyPlayerId] -> MyPlayer

qndel:


AJenbo (approved):

Source/player.cpp
```diff
@@ -36,7 +36,7 @@
 namespace devilution {
 
 int MyPlayerId;
-Player *MyPlayer;
+Player *MyPlayer = &Players[MyPlayerId];
```
ephphatha, line 39: This is good, globals are zero-initialised so this ends up equivalent to `MyPlayer = &Players[0]`. I had an attempt at this a while back where I didn't initialise MyPlayer and tried to find where it was referenced before a game actually started only to get annoyed at shop initialisation code running when selecting a save game... This at least keeps the current behaviour where the code just uses player 0.

qndel: I know? xD

AJenbo (approved):

PR #4156, opened 2022-03-13
prevent items from going above 255 durability

qndel:


julealgon:
Suggestion: create a value type called `Durability` and handle these concerns transparently inside the type with operator overloads. The same type could also abstract the "indestructible" concept more cleanly.

TL;DR: primitive obsession is bad.

qndel:
> Suggestion: create a value type called `Durability` and handle these concerns transparently inside the type with operator overloads. The same type could also abstract the "indestructible" concept more cleanly.
> 
> TL;DR: primitive obsession is bad.

Feel free to make a PR ;)

PR #4160, opened 2022-03-15
Clean up inventory code

AJenbo:
These are the general cleanups that we ended up doing while cleaning up the Stash code

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -610,9 +610,9 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 		return;
 	}
 
-	if (dropGoldFlag) {
+	if (IsDropGoldOpen) {
 		CloseGoldDrop();
-		dropGoldValue = 0;
+		DropGoldValue = 0;
 	}
 
 	bool done = false;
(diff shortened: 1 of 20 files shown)


Source/player.cpp
```diff
@@ -1937,16 +1937,13 @@ bool Player::HasItem(int item, int *idx) const
 
 void Player::RemoveInvItem(int iv, bool calcScrolls)
 {
-	iv++;
-
 	// Iterate through invGrid and remove every reference to item
 	for (int8_t &itemId : InvGrid) {
-		if (itemId == iv || itemId == -iv) {
+		if (abs(itemId) - 1 == iv) {
```
glebm, line 1942: Is the `- 1` intentional?

Source/inv.cpp
```diff
@@ -2050,6 +2042,19 @@ bool UseStaff()
 	return CanUseStaff(myPlayer.InvBody[INVLOC_HAND_LEFT], myPlayer._pRSpell);
 }
 
+const Item &GetInventoryItem(int location)
```
ephphatha, line 2045: I think this function can take a player reference to make its purpose a bit clearer. This would match the way `UseInvItem` is called.

ephphatha (dismissed):
GetOutlineColor could potentially use a rename to make it clear that the function returns an index into the palette array but that's way out of scope of this PR.

PR #4162, opened 2022-03-15
Update action for source tarball

tsunamistate:
* Use apt-get instead of apt, since apt has a warning
* Add Gtest dependency since it's now required

@glebm please review


AJenbo (approved):
I dont see a need for switching to apt-get, since we live in the future we know that the interface did not change. I would suggest using a newer ubuntu image instead

glebm (approved):

PR #4164, opened 2022-03-16
Leave healer

AJenbo:
When the healing option was removed from the menu the leave option was moved up. This is evidently not the right change since all stores have the leave option at at least the 18th position even when they have less options. It just happened that the healer had an option for all other lines.


StephenCWills:
Hope you don't mind me adding a commit here. Leaving his Buy menu will place the cursor on Leave Healer's Home, which is no longer valid since you just moved it. However, I believe this was an error in the first place because the Receive Healing option was removed so I changed the values to 14 instead of 18.

StephenCWills (approved):

PR #4165, opened 2022-03-16
Adjust gradle task dependencies to ensure assets are included in the APK

StephenCWills:
This resolves the issue with the Android test build that's been reported a handful of times on Discord.

### DIFF
--- a/android-project/app/build.gradle
+++ b/android-project/app/build.gradle
@@ -29,6 +29,10 @@ android {
 			proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
 		}
 	}
+	applicationVariants.all { variant ->
+		tasks["merge${variant.name.capitalize()}Assets"]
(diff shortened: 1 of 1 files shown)


android-project/app/build.gradle
```diff
@@ -29,6 +29,10 @@ android {
 			proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
 		}
 	}
+	applicationVariants.all { variant ->
+		tasks["merge${variant.name.capitalize()}Assets"]
+			.dependsOn(externalNativeBuildDebug)
```
StephenCWills, line 34: Actually.. I wonder if this needs to be `externalNativeBuild${variant.name}` or something...

StephenCWills:
image

AJenbo:
I wonder if it would be worth it to upstream this fix?
https://github.com/libsdl-org/SDL/blob/main/android-project/app/build.gradle

PR #4166, opened 2022-03-16
Use apt-get instead of apt in CI

tsunamistate:
Usage of `apt-get` instead of `apt` applied everywhere else.

For reasoning, see https://manpages.debian.org/buster/apt/apt.8#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS


AJenbo (approved):
Sure why not

PR #4168, opened 2022-03-17
[gamepad] Align cursor after using item

AJenbo:


obligaron (approved):

AJenbo:
That you for the review :)

PR #4169, opened 2022-03-18
Only load stash on New Game and Load Game

StephenCWills:
This filters calls to `LoadStash()` so items get loaded from disk only on New Game and Load Game.

The game loads the stash from disk whenever the player changes levels, which causes some problems for Single Player. Kudos go to AJenbo for catching this so quick after merging.


AJenbo (approved):
Thanks :)

PR #4170, opened 2022-03-18
Remove use of tempstr

AJenbo:
Avoid buffer overflow, text cropping and useless string copying.


Source/qol/xpbar.cpp
```diff
@@ -45,6 +44,27 @@ void DrawEndCap(const Surface &out, Point point, int idx, const ColorGradient &g
...
+		mlength = 3;
+	out.append(number.substr(0, mlength));
+	for (int i = mlength; i < length; i += 3) {
+		out.append(_(/* TRANSLATORS: Thousands separator */ ","));
+		out.append(number.substr(i, 3));
+	}
+
+	return out;
+}
+
```
ephphatha, line 67: Bit of an aside but this can be replaced with `fmt::format("{:L}", n)` if the global locale is configured properly (e.g. by adding `std::locale::global(std::locale(""));` into `InitXpBar()`).

Source/stores.cpp
```diff
@@ -201,11 +202,13 @@ void AddSTextVal(int y, int val)
...
 
-void AddSText(int x, int y, const char *str, UiFlags flags, bool sel)
+void AddSText(int x, int y, string_view text, UiFlags flags, bool sel)
 {
 	stext[y]._sx = x;
 	stext[y]._syoff = 0;
-	strcpy(stext[y]._sstr, str);
+	int len = std::min<int>(text.length(), 127);
+	strncpy(stext[y]._sstr, text.data(), len);
+	stext[y]._sstr[len] = '\0';
```
ephphatha, line 211:
```suggestion
	CopyUtf8(stext[y]._sstr, text, sizeof(stext[y]._sstr));
```

(after including `utils/utf8.hpp`)

Source/stores.cpp
```diff
@@ -973,33 +952,35 @@ void StoreConfirm()
 	AddSTextVal(8, item._iIvalue);
 	PrintStoreItem(item, 9, itemColor);
 
+	const char *prompt;
```
ephphatha, line 955:
```suggestion
	string_view prompt;
```

A fatal exception is thrown if the default case is hit so this is overly defensive, but this ensures the default `string_view` constructor is used giving a 0 length string instead of potentially calling AddSText with a garbage pointer.

AJenbo: Nice, we should get rid of as much `char *` as possible.

Source/stores.cpp
AJenbo: Thanks. I had a feeling I forgot this in a couple of places, but was to tired to properly go over it again :)

Source/qol/xpbar.cpp
AJenbo: I looked at it and tried it, but that requires the locale be present on the system, and also that it is properly supported which isn't the case for a few systems (Windows doesn't properly support `locale`). And we support setting the language to `zh_TW` with out having system wide zh_TW, which results in a crash when using `"{:L}"`. So i begrudgingly ended up making my own implementation :)

Source/monster.cpp
```diff
@@ -4679,36 +4678,34 @@ void PrintMonstHistory(int mt)
...
 				if ((res & RESIST_FIRE) != 0)
-					strcat(tempstr, _("Fire "));
+					resists.append(_("Fire "));
 				if ((res & RESIST_LIGHTNING) != 0)
-					strcat(tempstr, _("Lightning "));
-				string_view str { tempstr };
-				str.remove_suffix(str.size() - FindLastUtf8Symbols(str));
+					resists.append(_("Lightning "));
+				string_view str { resists };
+				str.remove_suffix(str.length() - FindLastUtf8Symbols(str));
```
ephphatha, line 4697: This is to remove the trailing space right? Would be easier to use space as a prefix on the resist type strings:

``` cpp
std::string resists = _("Resists:");
if ((res & RES_MAGIC) != 0)
    resists.append(_(" Magic"));
...
AddPanelString(resists);
```
(picking this up because the call to `string_view::size()` was changed here but not on line 4709)

AJenbo: Could even do `esists.append(" ").append(_("Magic"));` so not to generate new translation strings for this.

AJenbo: ... except that some languages do not sure regular spaces for separation :face_in_clouds:

ephphatha: Yeah, I remember you telling me that for another translation string 😄 . Figured this way using the space as a prefix lets the translations drop it if they don't need word separators. I'm happy to do that refactor as another PR anyway, for this one the main concern is just making the `size()`/`length()` calls consistent.

AJenbo: Fixed, also replaced a bunch of `strcpy()` with `CopyUTF8()`, hope it's not to distracting in the middle of a review :)

ephphatha (approved):
Auto-merge goooooo. (Possibly a couple of spots where some minutia could potentially mean those specific instances are slower than the old code but I don't see any obvious pitfalls)

PR #4172, opened 2022-03-18
Update _iStatFlag for stash items

obligaron:
Fixes a issue mentioned in stash pr #3853

> If a player doesn't have the requirements of an item it is placed on the stash tinted as red, if another player opens the stash it will be still red. The opposite is also true, items show colored even if the player doesn't have the requirements

This pr updates `_iStatFlag` for the active page when opening the stash or changing the active page.
This pr also reduces the number of places where max stash pages are defined.

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -548,6 +548,13 @@ void StashStruct::SetPage(int newPage)
 	page = clamp(newPage, 0, CountStashPages - 1);
 }
 
+void StashStruct::RefreshItemStatFlags()
+{
+	for (auto &item : Stash.stashList) {
+		item._iStatFlag = MyPlayer->CanUseItem(item);
+	}
+}
+
 void WithdrawGoldKeyPress(char vkey)
 {
 	auto &myPlayer = Players[MyPlayerId];
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -28,6 +28,8 @@ int WithdrawGoldValue;
 
 namespace {
 
+constexpr int CountStashPages = 50;
+
 int InitialWithdrawGoldValue;
 
 /** Contains mappings for the buttons in the stash (2 navigation buttons, withdraw gold buttons, 2 navigation buttons) */
@@ -109,7 +111,7 @@ void CheckStashPaste(Point cursorPosition)
 	// Check that no more then 1 item is replaced by the move
 	uint16_t it = 0;
 	for (auto point : PointsInRectangleRange({ { 0, 0 }, itemSize })) {
-		uint16_t iv = Stash.stashGrids[Stash.page][firstSlot.x + point.x][firstSlot.y + point.y];
+		uint16_t iv = Stash.stashGrids[Stash.GetPage()][firstSlot.x + point.x][firstSlot.y + point.y];
 		if (iv == 0 || it == iv)
 			continue;
 		if (it == 0) {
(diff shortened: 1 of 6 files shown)


Source/qol/stash.cpp
```diff
@@ -538,6 +537,21 @@ void StashStruct::RemoveStashItem(uint16_t iv)
 	Stash.dirty = true;
 }
 
+void StashStruct::SetPage(int newPage)
+{
+	page = clamp(newPage, 0, 49);
```
qndel, line 542: I think it'd be good to extract the 49 to a separate variable for visibility and change it to 99 :)

obligaron: I generalized `CountStashPages` and used it here and for `AutoPlaceItemInStash`.
I would like to separate the change to 100 pages from this bugfix-pr.
But I'm not against more pages. 😁

Source/stores.cpp
```diff
@@ -2169,6 +2169,8 @@ void BarmaidEnter()
 	case 14:
 		stextflag = STORE_NONE;
 		IsStashOpen = true;
+		// Forces the page to be refreshed (are items valid to use ...)
+		Stash.SetPage(Stash.GetPage());
```
ephphatha, line 2173: Having a `Stash::RefreshItemStatFlags()` (or similar) method would be good here instead of a weird looking call to set the page to the current page :D

obligaron: Thanks. I added the method. 🙂

obligaron:
With the last change I changed that all items are updated, not only the visible one.
This should help for a text based stash version.

Another thing I changed is that now `_iStatFlag` is also updated when the player changes gears. This was missing in the initial version.

Source/qol/stash.cpp
```diff
@@ -542,6 +543,19 @@ void StashStruct::RemoveStashItem(uint16_t iv)
...
 }
 
+void StashStruct::SetPage(int newPage)
+{
+	page = clamp(newPage, 0, CountStashPages - 1);
+}
+
+void StashStruct::RefreshItemStatFlags()
+{
+	// Iterate through active page and update if a item is usable
```
ephphatha, line 553: Was tempted to suggest changing this loop but you were a step ahead :D. Only thing I'd be hesitant about is if potential slowdown when opening the stash if someone ends up with more than a few hundred items in the stash. It's a pretty fast operation anyway so I don't think this'll be noticeable even with a full stash.

```suggestion
```
The doxygen comment on the method is more accurate than this inline comment, no value keeping this one.

obligaron: Gone 😉

AJenbo (approved):
Much better now that it's not part of SetPage :+1:

Source/qol/stash.cpp
```diff
@@ -423,7 +424,7 @@ uint16_t CheckStashHLight(Point mousePosition)
 
 	ClearPanel();
 
-	uint16_t itemId = abs(Stash.stashGrids[Stash.page][slot.x][slot.y]);
+	uint16_t itemId = abs(Stash.stashGrids[Stash.GetPage()][slot.x][slot.y]);
```
glebm, line 427: This calls `abs` on a `uint16_t`. Redundant or a bug?

AJenbo: Redundant, leftover from back when they where `int8_t`

AJenbo: Will be fixed once this is merged: https://github.com/diasurgical/devilutionX/pull/4239/files#diff-fcb318e528e35c32e1bb412094800c059f6bea7a86003f15541e06bde8116fc0R415

PR #4173, opened 2022-03-18
Allow to swap items with ctrl+click between inventory and stash

obligaron:
With this pr we can fill the brand new stash faster. 😁

## Notes
- If inventory is full the item is still dropped on the ground.
- If stash (unlikely) is full the item is still dropped on the ground.
- If a stash page is full the item will be placed on the next stash page. If we are on the last stash page the algorithm starts with page 1 again.
- Needs to be rebased if #4172 is merged before (cause of the usage of `Stash.page`).

## Example Video
https://user-images.githubusercontent.com/25415264/159096311-ae1f41e4-8e8b-43c9-95eb-d3619f46b2c8.mp4

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -895,8 +895,13 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 		}
 	}
 
-	if (dropItem) {
-		TryDropItem();
+	if (dropItem && !holdItem.isEmpty()) {
+		if (IsStashOpen && AutoPlaceItemInStash(player, holdItem, true)) {
+			holdItem._itype = ItemType::None;
+			NewCursor(CURSOR_HAND);
+		} else {
+			TryDropItem();
+		}
 	}
 }
 
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -158,19 +158,6 @@ void AddItemToInvGrid(Player &player, int invGridIndex, int invListIndex, Size i
 }
 
 /**
- * @brief Gets the size, in inventory cells, of the given item.
- * @param item The item whose size is to be determined.
- * @return The size, in inventory cells, of the item.
- */
-Size GetInventorySize(const Item &item)
-{
-	int itemSizeIndex = item._iCurs + CURSOR_FIRSTITEM;
-	auto size = GetInvItemSize(itemSizeIndex);
-
-	return { size.width / InventorySlotSizeInPixels.width, size.height / InventorySlotSizeInPixels.height };
-}
-
-/**
  * @brief Checks whether the given item can fit in a belt slot (i.e. the item's size in inventory cells is 1x1).
  * @param item The item to be checked.
  * @return 'True' in case the item can fit a belt slot and 'False' otherwise.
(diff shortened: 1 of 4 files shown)


yuripourre:
Really nice addition! This is QoL!

Source/inv.cpp
```diff
@@ -908,8 +895,13 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 		}
 	}
 
-	if (dropItem) {
-		TryDropItem();
+	if (dropItem && !player.HoldItem.isEmpty()) {
+		if (IsStashOpen && AutoPlaceItemInStash(player, player.HoldItem, true)) {
+			player.HoldItem._itype = ItemType::None;
```
AJenbo, line 900: They are both a reference to the same, but maybe better for consistency :)
```suggestion
	if (dropItem && !holdItem.isEmpty()) {
		if (IsStashOpen && AutoPlaceItemInStash(player, holdItem, true)) {
			holdItem._itype = ItemType::None;

```

Source/qol/stash.cpp
```diff
@@ -228,8 +228,13 @@ void CheckStashCut(Point cursorPosition, bool automaticMove, bool dropItem)
 		}
 	}
 
-	if (dropItem) {
-		TryDropItem();
+	if (dropItem && !player.HoldItem.isEmpty()) {
+		if (invflag && AutoPlaceItemInInventory(player, player.HoldItem, true)) {
+			player.HoldItem._itype = ItemType::None;
```
AJenbo, line 233:
```suggestion
	if (dropItem && !holdItem.isEmpty()) {
		if (invflag && AutoPlaceItemInInventory(player, holdItem, true)) {
			holdItem._itype = ItemType::None;
```

Source/qol/stash.cpp
```diff
@@ -607,4 +612,51 @@ void GoldWithdrawNewText(string_view text)
 	}
 }
 
+bool AutoPlaceItemInStash(Player &player, const Item &item, bool persistItem)
```
AJenbo, line 615: Nice, this would also work nicely with a potential text (like the stores) version of the stash :)

Source/qol/stash.cpp
```diff
@@ -607,4 +612,51 @@ void GoldWithdrawNewText(string_view text)
...
+			bool isSpaceFree = true;
+			for (auto itemPoint : PointsInRectangleRange({ { 0, 0 }, itemSize })) {
+				uint16_t iv = Stash.stashGrids[pageIndex][stashPosition.x + itemPoint.x][stashPosition.y + itemPoint.y];
+				if (iv != 0) {
+					isSpaceFree = false;
+					break;
+				}
+			}
+			if (!isSpaceFree)
+				continue;
```
AJenbo, line 647: Some of this could probably be shared with other parts of the code.

AJenbo (approved):

AJenbo:
Nice, this is a big help for getting us ready for the release :)

I'm ok with merging as is, but still left a few minor comments. I will leave it up to you if you would prefer to have it merged or make the adjustments.

Source/inv.cpp
obligaron: Fixed 🙂

Source/qol/stash.cpp
obligaron: Fixed 🙂

Source/qol/stash.cpp
obligaron: I thought about reusing it in `CheckStashPaste`.
But `CheckStashPaste` works differently. if we find exactly one item at this space this item is swapped with the pasted item.
This isn't behavior isn't desired in `AutoPlaceItemInStash`.
That's why I didn't combine them. After that I didn't investigate further.

obligaron:
> I'm ok with merging as is, but still left a few minor comments. I will leave it up to you if you would prefer to have it merged or make the adjustments.

if we have good feedback, we should address it. 🙂

AJenbo (approved):

PR #4176, opened 2022-03-19
Allow identify/oil/repair/recharge on stash items

obligaron:
Fixes a issue mentioned in stash pr #3853

> It's not possible to use or cast spells on items in the stash

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1843,9 +1843,14 @@ bool TryIconCurs()
 	}
 
 	if (pcurs == CURSOR_OIL) {
+		bool changeCursor = true;
 		if (pcursinvitem != -1)
-			DoOil(myPlayer, pcursinvitem);
-		else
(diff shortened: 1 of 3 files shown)


Source/diablo.cpp
```diff
@@ -1812,8 +1812,11 @@ bool TryIconCurs()
 	if (pcurs == CURSOR_IDENTIFY) {
 		if (pcursinvitem != -1)
 			CheckIdentify(myPlayer, pcursinvitem);
-		else
-			NewCursor(CURSOR_HAND);
+		else if (pcursstashitem != uint16_t(-1)) {
+			Item &item = Stash.stashList[pcursstashitem];
+			item._iIdentified = true;
+		}
+		NewCursor(CURSOR_HAND);
```
AJenbo, line 1819: I guess we can remove this from `CheckIdentify()` now if we always do it after calling it :)

AJenbo (approved):
Looks good, nice with the clean ups :+1:

PR #4177, opened 2022-03-19
Allow to consume items from stash

obligaron:
With this pr it is possible to use/consume items from stash (for example a potion).

Notes
- I tried to simplify `RightMouseDown` and removed a redundant call to `UseInvItem`. i would appreciate if someone could double check this.
- For now i disabled the use/consume of scrolls from stash. This would make fixing double cast bug harder. For more details see this [discussion](https://github.com/diasurgical/devilutionX/pull/3853#issuecomment-1072957281).

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -397,15 +397,16 @@ void RightMouseDown(bool isShiftHeld)
 		SetSpell();
 		return;
 	}
-	if ((!sbookflag || !GetRightPanel().Contains(MousePosition))
-	    && !TryIconCurs()
-	    && (pcursinvitem == -1 || !UseInvItem(MyPlayerId, pcursinvitem))) {
-		if (pcurs == CURSOR_HAND) {
-			if (pcursinvitem == -1 || !UseInvItem(MyPlayerId, pcursinvitem))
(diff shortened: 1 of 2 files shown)


Source/diablo.cpp
```diff
@@ -397,15 +397,18 @@ void RightMouseDown(bool isShiftHeld)
...
+	if (sbookflag && GetRightPanel().Contains(MousePosition))
+		return;
+	if (TryIconCurs())
+		return;
+	if (pcursinvitem != -1 && UseInvItem(MyPlayerId, pcursinvitem))
+		return;
+	if (pcursstashitem != uint16_t(-1) && UseStashItem(pcursstashitem))
+		return;
+	if (pcurs == CURSOR_HAND) {
+		CheckPlrSpell(isShiftHeld);
```
AJenbo, line 409: From what I can tell the old code actually never managed to call this since the first call to `UseInvItem()` would mean that the condition guarding it would always be false. Do you think this fixes an issue in the game?

AJenbo (approved):
The original logic is really messy :(

After staring at your code for an extended period of time I think it should be good. I'll merge it now and do some extended play testing later today.

PR #4178, opened 2022-03-19
Remove the use of HoldItem in AutoGetItem

ephphatha:
I also found a really stupid feature that can be supported with auto-pickup which was only prevented by a minor bug with the gold pickup code, so the last commit addresses that bug. Now you can pick up items while walking with the inventory open and a held item :D

https://user-images.githubusercontent.com/357684/159118640-2a7fb953-13ff-4ee1-861e-a4458e2ae8f9.mp4

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -867,7 +867,7 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 		}
 
 		CalcPlrInv(player, true);
-		CheckItemStats(player);
+		holdItem._iStatFlag = player.CanUseItem(holdItem);
 
 		if (pnum == MyPlayerId) {
 			if (automaticallyEquipped) {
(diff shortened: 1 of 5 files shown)


Source/inv.cpp
```diff
@@ -1636,48 +1628,44 @@ void InvGetItem(int pnum, int ii)
...
-void AutoGetItem(int pnum, Item *item, int ii)
+void AutoGetItem(int pnum, Item *itemPointer, int ii)
 {
-	bool done;
-	bool autoEquipped = false;
-
-	if (pcurs != CURSOR_HAND) {
-		return;
-	}
+	Item &item = *itemPointer;
```
AJenbo, line 1633: Why not take the item by reference as the parameter? There only appears to be 3 usage and it can never be null.

AJenbo (approved):
Nice work, this code was really unnecessarily convoluted :D

Source/inv.cpp
ephphatha: To make `CleanupItem` behave like an `erase` function it should operator on an iterator, which would mean that this function would be best to take an iterator as well. So for now I was leaving the function taking a pointer until I get around to that refactoring.

ephphatha: Hmm, actually that wasn't the reason exactly. CleanupItem and DeleteItem are related but CleanupItem does a bit extra to handle cornerstone items. Probably gonna end up combining the functions since they're both used for picking up an item from the ground so really should be the same thing.

PR #4180, opened 2022-03-19
Ctrl+click stash item swap: if there is no room left, don't drop the item on the ground

obligaron:
Instead play `HeroSpeech::IHaveNoRoom` and put the item in the hand/cursor.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -902,7 +902,7 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 				holdItem._itype = ItemType::None;
 				NewCursor(CURSOR_HAND);
 			} else {
-				player.SaySpecific(HeroSpeech::IHaveNoRoom);
+				player.SaySpecific(HeroSpeech::WhereWouldIPutThis);
 			}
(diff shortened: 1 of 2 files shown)


Source/inv.cpp
```diff
@@ -897,9 +897,13 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
...
 	if (dropItem && !holdItem.isEmpty()) {
-		if (IsStashOpen && AutoPlaceItemInStash(player, holdItem, true)) {
-			holdItem._itype = ItemType::None;
-			NewCursor(CURSOR_HAND);
+		if (IsStashOpen) {
+			if (AutoPlaceItemInStash(player, holdItem, true)) {
+				holdItem._itype = ItemType::None;
+				NewCursor(CURSOR_HAND);
+			} else {
+				player.SaySpecific(HeroSpeech::IHaveNoRoom);
```
AJenbo, line 905: For variation, also "I" can be taken as referring to the hero, and not the stash which is Gillians :)
```suggestion
				player.SaySpecific(HeroSpeech::WhereWouldIPutThis);
```

PR #4183, opened 2022-03-19
dont allow opening gold withdraw with item in cursor

qndel:
Earlier withdrawing gold while having an item in cursor would destroy the item.

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -246,6 +246,9 @@ void StartGoldWithdraw()
 {
 	CloseGoldDrop();
 
+	if (pcurs != CURSOR_HAND)
+		return;
+
 	InitialWithdrawGoldValue = Stash.gold;
 
 	if (talkflag)


AJenbo:
The thinking is to not make use of the cursor item for doing the transfer.

qndel:
> The thinking is to not make use of the cursor item for doing the transfer.

I'd block withdrawing gold while having an item in cursor anyway

obligaron:
I'm fine with disallowing it. We currently don't allow dropping gold directly from stash to inventory and how often do you withdraw gold when you have a item in the cursor? 🤔 

We also need to adjust these two lines in `diablo.cpp`
https://github.com/diasurgical/devilutionX/blob/a2f85ab8949011f7c288ab073d4b974791ecde99/Source/diablo.cpp#L340
https://github.com/diasurgical/devilutionX/blob/a2f85ab8949011f7c288ab073d4b974791ecde99/Source/diablo.cpp#L337
Else you can grab a item from stash when inventory gold dialog is open and vice versa.
This can also result in a lost item.

AJenbo:
Gold withdrawal no longer makes use of `HoldItem` so this should no longer be an issue.l

PR #4189, opened 2022-03-20
Only attempt to spawn loot for the player who delivered the killing blow

### LINKED ISSUE #4187: Loot drops 2x for the player that does not trigger the drop
In multiplayer if a player triggers a loot drop (open chest/kill monster) the other players will see the loot drop twice and sometimes in different locations causing duplicate objects that cannot be picked up.

This issue has been traced down to this commit: https://github.com/diasurgical/devilutionX/commit/eeb9ac3303d890d77e122949280d990e1fe8644d

To reproduce:
- Player 1 kills a monster
- Monster drops 5 gold
- Player 2 sees two stacks of 5 gold drop

To reproduce:
- Player 1 breaks a barrel
- A potion drops
- Player 2 hears two potions drop (there might be two flip animations happening in the same spot as well)

ephphatha:
This prevents non-host players seeing multiple item drops when the host kills a monster.

fixes #4187

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1104,7 +1104,10 @@ void StartMonsterDeath(int i, int pnum, bool sendmsg)
 	MonsterKillCounts[monster.MType->mtype]++;
 	monster._mhitpoints = 0;
 	SetRndSeed(monster._mRndSeed);
-	SpawnLoot(monster, sendmsg);
+	if (pnum == MyPlayerId) {
+		// prevent double-spawning loot, only the player who delivered the killing blow should send a drop item message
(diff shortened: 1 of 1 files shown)


AJenbo:
isn't the `sendmsg` flag the same as `pnum == MyPlayerId`? Also this wouldn't explain/solve 2x loot from chests.

AJenbo:
@ephphatha I would think this might be the proper solution?

https://github.com/diasurgical/devilutionX/pull/4191

ephphatha:
Didn't realise chest loot was getting doubled too, hadn't seen that happen myself :/. That PR looks more comprehensive so I'll close this one

Issue #4182, opened 2022-03-19
Multiplayer Playtest issues


AJenbo:
- [x] #4184
- [x] #4187
- [x] When auto equiping items the other player will sometimes not see that it has been equiped
- [x] If network dies for player 1, player 2 is kicked and the game disapears
- [x] #4186
- [x] Sometimes a monster will be alive for player 2 but dead for player 1, player 2 is not able to target them. If player 2 changes levels the monster is corrected to being dead. Monster HP some times appear to stutter, this might be caused by syncing monster HP, and appears to maybe only be happening when both players are hitting the same monster.
- [x] The vegetable line is played when entering level 16
- [x] #4252
- [ ] Repair menu gives wrong text ("No requirements") when no items are repairable (can't reproduce)

AJenbo:
- When auto equiping items the other player will not see that it has been equiped
- Sometimes player 2 can't pick items up (this might be the origina hellfire loot bug)
- Player 2 can't pickup gray suit (might be related to previous issue)
- When casting heal other there is no way to canel it if you don't have anyone to heal

StephenCWills:
> Multiplayer: Loot drops 2x for the player that does not trigger the drop

> Sometimes player 2 can't pick items up (this might be the original hellfire loot bug)

I believe these two issues are related. Also, these were previously reported in #4098 with a video. I bisected the loot drop 2x issue to eeb9ac3.

AJenbo:
- If network dies for player 1, player 2 is kicked and the game disapears

AJenbo:
- When selecting Join in ZeroTier it says "Join TCP game", description says "enter IP or Hostname", should be "game id"

AJenbo:
- Some times a monster will a live for player 2 but deat for player 1, player 2 is not able to target them

StephenCWills:
> If network dies for player 1, player 2 is kicked and the game disapears

This appears to be intentional behavior.

https://github.com/diasurgical/devilutionX/blob/8424ad4c6d816de7da30991500b809363d9900c9/Source/multi.cpp#L283-L293

* `bGroupPlayers` is the number of active players
* `bGroupCount` is the number of inactive players
* `nLowestActive` is the lowest player index of all the active players
* `nLowestPlayer` is the lowest player index of all the connected players

Assuming a 2-player game, player 1 and player 2 are both connected. From player 2's perspective, player 1 is no longer active.

* `bGroupPlayers = 1`
* `bGroupCount = 1`
[...]

Issue #3115, opened 2021-10-13
Monster hp not synced properly


qndel:
Player 1 damages a monster (you can disable monster regen for easier testing)
Player 2 enters the level and sees the monster with full hp.

Might be a vanilla bug 🤷

PR #4190, opened 2022-03-20
Avoid resetting monster health and mWhoHit

AJenbo:
As observed in https://github.com/diasurgical/devilutionX/issues/4182 monsters health appears to get stuck and jitter in current master. This is probably caused by https://github.com/diasurgical/devilutionX/commit/b00feb4c3d85fca3fa7cf2cf1975a1629744b35a where monsters health is synced on every tick, which may well cancel out most damage dealt to the monster by a ranged player (since the closest player's data will overwrite the more distant one).

It also appeared that mWhoHit has not taking the ranged player in to account which would again be explained by `mWhoHit` being taken from the melee player. Instead this is now ored on to the local state.

### DIFF
--- a/Source/sync.cpp
+++ b/Source/sync.cpp
@@ -197,8 +197,7 @@ void SyncMonster(int pnum, const TSyncMonster &monsterSync)
 	}
 
 	decode_enemy(monster, enemyId);
-	monster.mWhoHit = monsterSync.mWhoHit;
-	monster._mhitpoints = monsterSync._mhitpoints;
+	monster.mWhoHit |= monsterSync.mWhoHit;
 }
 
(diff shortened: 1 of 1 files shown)


obligaron (approved):
lgtm

As far as I understand this doesn't regress #3115, cause we still use the synced hitpoints when loading a level (`DeltaLoadLevel`).
`SyncMonster` is only called when we are on the same level as the (synced) monster.

AJenbo:
> lgtm
> 
> As far as I understand this doesn't regress #3115, cause we still use the synced hitpoints when loading a level (`DeltaLoadLevel`). `SyncMonster` is only called when we are on the same level as the (synced) monster.

Correct :)

PR #4191, opened 2022-03-20
Only spawn loot once

AJenbo:
Alternate solution to #4187

1. https://github.com/diasurgical/devilutionX/commit/eeb9ac3303d890d77e122949280d990e1fe8644d turned out to cause loot to drop twice
2. Since quest items are not per-generated like other loot it has to be recreated on the other end, similar to inventory items
3. Quest items are not always dropped next to the player so skip `CanPut()`
4. `_iSelFlag` is `2` instead of `1` for quest items, so take that into account when re-spawning the item.

This does not solve Torn Notes as they use `SpawnQuestItem()` (also used by most single player quests), but it should be easy to make the same adjustments to it once we feel confident about how this works. As well the Torn Notes are not critical for accessing parts of the game.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3346,7 +3346,7 @@ void SpawnQuestItem(int itemid, Point position, int randarea, int selflag)
 	}
 }
 
-void SpawnRewardItem(int itemid, Point position)
+void SpawnRewardItem(int itemid, Point position, bool sendmsg)
 {
 	if (ActiveItemCount >= MAXITEMS)
 		return;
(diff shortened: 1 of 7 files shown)


AJenbo:
@ephphatha I tested this a bit and the issue is still happening frequently. clearly seen from barrels and monster drops.

- Player 1 kills a monster
- Monster drops 5 gold
- Player 2 sees two stacks of 5 gold drop

AJenbo:
I think maybe reverting https://github.com/diasurgical/devilutionX/pull/3987 and replacing it with this PR may solve the issue.

ephphatha (approved):
I'm not getting doubled loot from chests or monsters either as the level controller or a secondary player.

I do get some desyncs in position though, think it's best to try solve that in a later PR.

Source/items.cpp
```diff
@@ -3362,25 +3362,26 @@ void SpawnRewardItem(int itemid, Point position)
 	item._iSelFlag = 2;
 	item._iPostDraw = true;
 	item._iIdentified = true;
+	GenerateNewSeed(item);
 
-	if (gbIsMultiplayer) {
-		NetSendCmdPItem(false, CMD_DROPITEM, item.position, item);
+	if (sendmsg) {
+		NetSendCmdPItem(true, CMD_RESPAWNITEM, item.position, item);
```
AJenbo, line 3368: As an alternative to using `CMD_RESPAWNITEM` we could create a new command (`CMD_SPAWNITEM`) so that we have less special have handling in side the function. This also avoids reverting most of the changes. See https://github.com/diasurgical/devilutionX/pull/4198 for that option.

PR #4194, opened 2022-03-20
Use a viewport size of 1 when displaying a list of actions for the ok…

ephphatha:
… dialog

Having a size 0 viewport made the action invisible, forcing the player to use the Esc action.

master
2022-03-20T143147_devilutionx

fixed
2022-03-20T143243_devilutionx


AJenbo:
Ah thanks I just noticed that as well but was going to solve the item spawning before reporting it :D

AJenbo (approved):

PR #4196, opened 2022-03-20
Use string_view for UiArtTextButton, UiButton, UiEdit hint

ephphatha:
These are the items that are always called with a string instance owned by the current translation dictionary. ~UiEdit was included even though hints are never referenced...~ Found what UiEdit hints are used for, that mess is being pulled out for a later attempt

`string_view` can be created from c-strings so this works fine with current master and `std::string &` as returned by `LanguageTranslate` in #4174

### DIFF
--- a/Source/DiabloUI/ui_item.h
+++ b/Source/DiabloUI/ui_item.h
@@ -209,7 +209,7 @@ class UiArtTextButton : public UiItemBase {
 public:
 	using Callback = void (*)();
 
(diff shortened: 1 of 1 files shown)


AJenbo (approved):

AJenbo:
Nice, doesn't looks as scary as I thought it might :)

AJenbo:
... but also it didn't fix as much as it though :)

PR #4198, opened 2022-03-20
Only spawn loot once

AJenbo:
Alternative to #4191


ephphatha (approved):
I like this one a bit better though part of that might've been because I wasn't getting any desyncs with item positions when testing 😅

AJenbo:
Glad I did both :)

PR #4199, opened 2022-03-20
UK TL: Reach over 75% of translated entries

tsunamistate:
Yesterday, I only wanted to TL strings related to stash.
But then I had decided to pony up the cash for POEdit Pro licence. Machine translations from POEdit Pro provided the basis to start tackling the most fearsome beast - Shrine and dialogue text.

That helped me to reach 76% of translated entries in Ukrainian language, which I believe gets it inclusion in future release?

Note to self — DevilutionX ignores em-dashes

### DIFF
--- a/Translations/uk.po
+++ b/Translations/uk.po
@@ -5,7 +5,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: DevilutionX\n"
-"POT-Creation-Date: 2022-03-07 17:13+0200\n"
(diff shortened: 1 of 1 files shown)


AJenbo:
> Yesterday, I only wanted to TL strings related to stash.

:smile: A lot of people are really excited by the stash.

> That helped me to reach 76% of translated entries in Ukrainian language, which I believe gets it inclusion in future release?

Indeed this is fantastic :D Pretty crazy timing to I had just been looking at what the status was of the various translations.

PR #4200, opened 2022-03-20
Change infostr from char[128] to std::string

AJenbo:
I decided to do one last update to string handling before 1.4.0. @obligaron hope you like `std::string` as much as `CopyUtf8()` :)

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -8,6 +8,7 @@
 #include <algorithm>
 #include <array>
 #include <cstddef>
+#include <string>
 
 #include <fmt/format.h>
 
@@ -73,7 +74,7 @@ bool talkflag;
 bool sbookflag;
 bool chrflag;
 bool drawbtnflag;
-char infostr[128];
+std::string InfoString;
 bool panelflag;
 int initialDropGoldValue;
 bool panbtndown;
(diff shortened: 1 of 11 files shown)


Source/control.cpp
```diff
@@ -567,7 +568,7 @@ void InitControlPan()
 		buttonEnabled = false;
 	chrbtnactive = false;
 	pDurIcons = LoadCel("Items\\DurIcons.CEL", 32);
-	strcpy(infostr, "");
+	InfoString = "";
```
ephphatha, line 571: There's a few places you can use `string::clear()` instead of copy-assigning from an empty string. If it weren't for the calls to fmt I'd be tempted to use `string_view` for `InfoString` and leave these assignments as is since it works better with a view.
```suggestion
	InfoString.clear();
```

Source/control.cpp
```diff
@@ -889,19 +890,19 @@ void DrawInfoBox(const Surface &out)
 					PrintMonstHistory(monster.MType->mtype);
 				}
 			} else if (pcursitem == -1) {
-				CopyUtf8(infostr, Towners[pcursmonst].name, sizeof(infostr));
+				InfoString = std::string(Towners[pcursmonst].name);
```
ephphatha, line 893: The move constructor should work without the explicit constructor call, is this failing on GCC6?

AJenbo: yes :(

obligaron (approved):
lgtm 🙂 

`std::string` is even better then `CopyUtf8` 😉

PR #4201, opened 2022-03-20
Use string_view for UiEdit (and platform specific code)

ephphatha:
This is a bit of an experiment, hoping the automated builds flag any problems.

### DIFF
--- a/Source/platform/switch/keyboard.cpp
+++ b/Source/platform/switch/keyboard.cpp
@@ -7,7 +7,7 @@
 
 #include "utils/utf8.hpp"
 
-static void switch_keyboard_get(const char *guide_text, char *initial_text, int max_len, int multiline, char *buf)
+static void switch_keyboard_get(devilution::string_view guide_text, devilution::string_view initial_text, char *buf, unsigned buf_len)
 {
 	Result rc = 0;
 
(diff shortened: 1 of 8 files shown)


ephphatha:
😮 it builds! Hoping someone with the appropriate hardware can test creating a new hero or joining a multiplayer game, especially on 3DS. Should behave exactly the same as before.

AJenbo:
@StephenCWills @isage are you guy avalible for testing this on 3DS and Vita?

Source/platform/switch/keyboard.cpp
```diff
@@ -17,10 +17,25 @@ static void switch_keyboard_get(const char *guide_text, char *initial_text, int
...
+			memcpy(buffer, guide_text.data(), guide_text.size());
+			buffer[guide_text.size()] = '\0';
+			swkbdConfigSetGuideText(&kbd, buffer);
+		}
+		if (!initial_text.empty()) {
+			char buffer[initial_text.size() + 1];
+			memcpy(buffer, initial_text.data(), initial_text.size());
+			buffer[initial_text.size()] = '\0';
+			swkbdConfigSetInitialText(&kbd, buffer);
+		}
```
ephphatha, line 35: This could potentially just use `std::string`?
```suggestion
		// swkbConfigSetGuide/InitialText both copy the input string. They expect a null terminated c-string but we're
		//  getting a string view which may be a substring of a larger std::string/c-string, so we copy to a temporary
		//  null-terminated string first for safety
		if (!guide_text.empty()) {
			std::string textCopy { guide_text };
			swkbdConfigSetGuideText(&kbd, textCopy.c_str());
		}
		if (!initial_text.empty()) {
			std::string textCopy { initial_text };
			swkbdConfigSetInitialText(&kbd, textCopy.c_str());
		}
```

isage:
> @StephenCWills @isage are you guy avalible for testing this on 3DS and Vita?

Yup. I'll test a little later

StephenCWills:
On 3DS, the hint text seems to work fine, but the initial text does not. The on-screen keyboard acts as if the initial text is empty.

isage:
Seems to work fine on Vita

ephphatha:
> On 3DS, the hint text seems to work fine, but the initial text does not. The on-screen keyboard acts as if the initial text is empty.
> 
> EDIT: It seems just a glance at the libctru code should probably help explain why. https://github.com/devkitPro/libctru/blob/ebb53051886cfe966f1aa27ff3dd6689176de483/libctru/source/applets/swkbd.c#L92-L106

Can't believe I missed that, especially after checking the switch code for the same pitfall. Using a separate buffer for the initial text now to ensure it stays valid until the input call.

StephenCWills:
Looks good now on 3DS.

AJenbo (approved):

AJenbo:
Yay now we can save 3 words in the menu code :joy:

PR #4207, opened 2022-03-20
Sync mana between clients

obligaron:
With this change mana is synchronized between clients/players.
Should help to get #4074 along, see this [comment](https://github.com/diasurgical/devilutionX/pull/4074#discussion_r830529747).
Tested with modified `control.cpp` (when selecting another player mana is shown instead of hp; not included in this pr).
Works even with debug `fill` command. 😉


StephenCWills (approved):

AJenbo:
The more I understand how the network code works the crazier I think it is :D

PR #4210, opened 2022-03-21
Do not exit game on network timeout

StephenCWills:
This removes the logic that attempted to determine which multiplayer client should be the one to resume a game session in the event of a network interruption. Instead, every client will individually determine which players need to be kicked so that the game session can resume locally.

After a bunch of testing and brainstorming with @AJenbo, we came to the conclusion that there's no compelling reason to try so hard to make sure that only one game session remains active in the event of a network disruption. On the contrary, it's not difficult to imagine legitimate scenarios where doing so would be undesirable for the entire party.

For more context regarding what this is all about, see https://github.com/diasurgical/devilutionX/issues/4182#issuecomment-1073078627.

### DIFF
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -257,42 +257,7 @@ void BeginTimeout()
 		return;
 	}
 
-	int nLowestActive = -1;
-	int nLowestPlayer = -1;
-	uint8_t bGroupPlayers = 0;
-	uint8_t bGroupCount = 0;
-	for (int i = 0; i < MAX_PLRS; i++) {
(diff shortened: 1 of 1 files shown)


AJenbo:
During my play test the host computer crashed, causing the other play to conclude that they where to blame so they kicked them selfs from the session leaving, meaning there session was lost (including all none-stashed equipment).

@StephenCWills 's changes leaves it up to the players to decide how to continue the game, which is arguably the right thing to do in nearly all situations.

AJenbo (approved):

PR #4215, opened 2022-03-21
Unify gold handling (withdraw any amount from stash)

AJenbo:
- Always create gold from template
- Allow withdrawing any amount from stash
- Fill inventory in the same pattern when getting gold from ground/shop/stash

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -369,11 +369,7 @@ void RemoveGold(Player &player, int goldIndex)
 		SetPlrHandGoldCurs(player.InvList[gi]);
 	else
 		player.RemoveInvItem(gi);
-	InitializeItem(player.HoldItem, IDI_GOLD);
-	SetGoldSeed(player, player.HoldItem);
-	player.HoldItem._ivalue = dropGoldValue;
-	player.HoldItem._iStatFlag = true;
-	ControlSetGoldCurs(player.HoldItem);
+	MakeGoldStack(player.HoldItem, dropGoldValue);
 	player._pGold = CalculateGold(player);
 	dropGoldValue = 0;
 }
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1556,7 +1556,7 @@ int AddGoldToInventory(Player &player, int value)
 		value = CreateGoldItemInInventorySlot(player, i, value);
 	}
 
-	// Remaning inventory in colums, bottom to top, right to left
+	// Remaining inventory in columns, bottom to top, right to left
 	for (int x = 9; x >= 0 && value > 0; x--) {
 		for (int y = 2; y >= 0 && value > 0; y--) {
 			value = CreateGoldItemInInventorySlot(player, 10 * y + x, value);
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -571,11 +571,11 @@ void CheckInvPaste(int pnum, Point cursorPosition)
 				cn = SwapItem(player.InvList[invIndex], player.HoldItem);
 				if (player.HoldItem._itype == ItemType::Gold)
 					player._pGold = CalculateGold(player);
-				for (auto &itemId : player.InvGrid) {
-					if (itemId == it)
-						itemId = 0;
-					if (itemId == -it)
-						itemId = 0;
+				for (auto &itemIndex : player.InvGrid) {
+					if (itemIndex == it)
+						itemIndex = 0;
+					if (itemIndex == -it)
+						itemIndex = 0;
 				}
 			}
 			int ii = r - SLOTXY_INV_FIRST;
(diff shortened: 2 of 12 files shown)


Source/items.h
```diff
@@ -413,6 +413,8 @@ struct CornerStoneStruct {
 
 struct Player;
 
+/** Temporary item used to generate gold piles by various function */
```
qndel, line 416:
```suggestion
/** Temporary item used to generate gold piles by various functions */
```

Source/qol/stash.cpp
```diff
@@ -264,11 +264,7 @@ void StartGoldWithdraw()
...
 void WithdrawGold(Player &player, int amount)
 {
-	InitializeItem(player.HoldItem, IDI_GOLD);
-	SetGoldSeed(player, player.HoldItem);
-	player.HoldItem._ivalue = amount;
-	player.HoldItem._iStatFlag = true;
-	ControlSetGoldCurs(player.HoldItem);
+	AddGoldToInventory(player, amount);
 	Stash.gold -= amount;
 	Stash.dirty = true;
```
ephphatha, line 269: This assumes the player has room for the full amount they requested to withdraw. The return value from `AddGoldToInventory` is ignored, this should be added back to `Stash.gold` or removed from the value that is then subtracted from `Stash.gold`

AJenbo: The input is limited to only accept the maximum amount that the user can fit in there inventory so that can never happen.

It's clunky with use of a Global, but that's cause it was copied from how SplitGold was done in the original game :roll_eyes:

Source/items.cpp
```diff
@@ -2901,15 +2874,15 @@ void CreatePlrItems(int playerId)
 		break;
 	}
 
-	auto &startingGold = player.InvList[player._pNumInv];
+	GenerateNewSeed(GoldTemplate);
+	GoldTemplate._ivalue = 100;
+	GoldTemplate._iCurs = ICURS_GOLD_SMALL;
+
+	player.InvList[player._pNumInv] = GoldTemplate;
```
ephphatha, line 2881: This is modifying a global object used in loads of other places, if you're going to copy based on the template would be better to make the modifications to the `InvList` instance.

Source/control.cpp
```diff
@@ -370,7 +370,7 @@ void RemoveGold(Player &player, int goldIndex)
 	else
 		player.RemoveInvItem(gi);
 	InitializeItem(player.HoldItem, IDI_GOLD);
```
ephphatha, line 372: You've removed every use of `InitializeItem(..., IDI_GOLD)` except for this one. TBH my preference would be to keep these calls instead of copying from `GoldTemplate` but consistency one way or the other would be better than mixing.

Source/items.cpp
```diff
@@ -4336,11 +4309,11 @@ void SpawnHealer(int lvl)
...
 }
 
-void SpawnStoreGold()
+void InitGoldTemplate()
 {
-	golditem = {};
-	GetItemAttrs(golditem, IDI_GOLD, 1);
-	golditem._iStatFlag = true;
+	GoldTemplate = {};
+	GetItemAttrs(GoldTemplate, IDI_GOLD, 1);
```
ephphatha, line 4315: I think this should be `InitializeItem`,`GetItemAttrs` does an RNG call to set the gold value which never ends up being relevant for this template (I'd argue its more predictable having a value of 0).

Source/qol/stash.cpp
ephphatha: Ahh, I see. I kinda like the behaviour of allowing the user to attempt to withdraw more than they can hold and leaving the excess behind, it saves having to think when you've got a partially filled inventory since there's no indicator of how much you're allowed to withdraw.

I did find a bug with this implementation though 😄 
image

Source/qol/stash.cpp
```diff
@@ -248,7 +248,7 @@ void StartGoldWithdraw()
 {
 	CloseGoldDrop();
 
-	InitialWithdrawGoldValue = Stash.gold;
+	InitialWithdrawGoldValue = RoomForGold();
```
ephphatha, line 251:
```suggestion
	InitialWithdrawGoldValue = std::min(RoomForGold(), Stash.gold);
```
or just `Stash.gold` depending on the decision from the other comment trail :D

Source/inv.cpp
```diff
@@ -1519,42 +1511,72 @@ bool AutoPlaceItemInInventorySlot(Player &player, int slotIndex, const Item &ite
...
+			continue;
+		}
+		if (itemIndex == 0) {
+			amount += MaxGold;
+			continue;
+		}
 
-	for (int i = 0; i < player._pNumInv && !done; i++) {
-		if (player.InvList[i]._itype != ItemType::Gold)
+		Item &goldItem = MyPlayer->InvList[itemIndex];
```
qndel, line 1526:
```suggestion
		Item &goldItem = MyPlayer->InvList[itemIndex - 1];
```

Source/qol/stash.cpp
AJenbo: I don't disagree, but also do not want to refactor this with out also improving the other one. Simply adding the check here would possibly cause a lot of work for each key press, and atm I'm looking for the shortest path to a release.

Source/control.cpp
AJenbo: I prefer that my self, but Constancy was most important part and I wast sure if there would be a performance benefit. So wanted some feedback first :)

Source/objects.cpp
```diff
@@ -3089,7 +3089,7 @@ bool OperateShrineSpiritual(int pnum)
 	for (int8_t &gridItem : player.InvGrid) {
 		if (gridItem == 0) {
 			int r = 5 * leveltype + GenerateRnd(10 * leveltype);
-			uint8_t t = player._pNumInv;
+			Item &invItem = player.InvList[player._pNumInv];
 			player.InvList[t] = GoldTemplate;
```
qndel, line 3093: u forgot to change these lines ;p

AJenbo: They where not included in the suggestio, but I will do it manually once I'm at a computer.

Source/control.cpp
AJenbo: Alright I have made a helper function for generating gold stacks, I think this is both cleaner and probably even more optimal performance wise then copying the template.

Source/inv.cpp
```diff
@@ -1519,42 +1508,72 @@ bool AutoPlaceItemInInventorySlot(Player &player, int slotIndex, const Item &ite
...
-		done = GoldAutoPlaceInInventorySlot(player, i, goldStack);
+	// Last row right to left
+	for (int i = 39; i >= 30 && value > 0; i--) {
+		value = CreateGoldItemInInventorySlot(player, i, value);
 	}
-	for (int x = 9; x >= 0 && !done; x--) {
-		for (int y = 2; y >= 0 && !done; y--) {
-			done = GoldAutoPlaceInInventorySlot(player, 10 * y + x, goldStack);
+
+	// Remaning inventory in colums, bottom to top, right to left
```
qndel, line 1559:
```suggestion
	// Remaining inventory in columns, bottom to top, right to left

```

qndel (approved):

PR #4221, opened 2022-03-23
Zero-based frame indexing

glebm:
Index frames starting at 0 instead of 1.

This makes CEL sprites consistent with PCX sprites and removes the need for a redundant `0` at the start of every widths array.

Note: This or #4228 needs to be adjusted when the other one has been merged.

/cc @ephphatha


Source/cursor.cpp
```diff
@@ -147,26 +145,27 @@ void FreeCursor()
...
 
 Size GetInvItemSize(int cursId)
 {
-	if (cursId >= InvItems1Size)
-		return { InvItemWidth2[cursId - (InvItems1Size - 1)], InvItemHeight2[cursId - (InvItems1Size - 1)] };
-	return { InvItemWidth1[cursId], InvItemHeight1[cursId] };
+	const int i = cursId - 1;
+	if (i >= InvItems1Size)
+		return { InvItemWidth2[i - (InvItems1Size - 1)], InvItemHeight2[i - (InvItems1Size - 1)] };
+	return { InvItemWidth1[i], InvItemHeight1[i] };
```
ephphatha, line 163: I think the `(InvItem1Size - 1)` value is now off by 1, it was compensating for the 0 size entry that's been removed from the start of that list. Rune of stone is being reported as a 2x2 item in this branch as `ICURS_RUNE_OF_STONE + CURS_FIRST_ITEM - 1 - (179 - 1) == 30`, which I think is meant to be the bovine suit's size.

Source/cursor.cpp
```diff
@@ -147,26 +145,27 @@ void FreeCursor()
...
 }
 
 void SetICursor(int cursId)
 {
-	icursSize = GetInvItemSize(cursId);
+	icursSize = cursId == 0 ? Size { 0, 0 } : GetInvItemSize(cursId);
```
ephphatha, line 168:
```suggestion
	icursSize = cursId == CURSOR_NONE ? Size { 0, 0 } : GetInvItemSize(cursId);
```

Source/cursor.cpp
```diff
@@ -178,7 +177,7 @@ void ResetCursor()
 void NewCursor(int cursId)
 {
 	pcurs = cursId;
-	cursSize = GetInvItemSize(cursId);
+	cursSize = cursId == 0 ? Size { 0, 0 } : GetInvItemSize(cursId);
```
ephphatha, line 180:
```suggestion
	cursSize = cursId == CURSOR_NONE ? Size { 0, 0 } : GetInvItemSize(cursId);
```

ephphatha (commented):
Won't have time to do a proper review till the weekend, a few comments from a quick test. It would be good to have some tests added to cursor_test.cpp for the boundary cases like Bovine Suit, Rune of Stone, Gold, Short Battle Bow, etc.

Source/cursor.cpp
```diff
@@ -58,7 +58,6 @@ const uint16_t InvItemWidth1[] = {
 	2 * 28, 2 * 28, 2 * 28, 2 * 28, 2 * 28, 2 * 28, 2 * 28, 2 * 28,
 };
 const uint16_t InvItemWidth2[] = {
-	0,
```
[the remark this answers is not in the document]
glebm: Very true but I'd rather do that cleanup in a separate PR

Source/cursor.cpp
glebm: Thanks! Fixed, I think?

Source/cursor.cpp
```diff
@@ -31,13 +31,13 @@ namespace {
 /** Cursor images CEL */
 std::optional<OwnedCelSprite> pCursCels;
 std::optional<OwnedCelSprite> pCursCels2;
-constexpr uint16_t InvItems1Size = 180;
+constexpr uint16_t InvItems1Size = 179;
```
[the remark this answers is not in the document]
glebm: Done (replaced with sizeof)

Source/cursor.cpp
```diff
@@ -178,7 +177,7 @@ void ResetCursor()
 void NewCursor(int cursId)
 {
 	pcurs = cursId;
-	cursSize = GetInvItemSize(cursId);
+	cursSize = cursId == CURSOR_NONE ? Size { 0, 0 } : GetInvItemSize(cursId);
```
ephphatha, line 180: Given the check is fairly simple and the function currently returns both inventory item sizes and cursor sizes you could move the `cursId == CURSOR_NONE` check inside `GetInvItemSize` to save repeating it here. Most of the callers of that function are using either a known item cursor or have checked that the cursor is not `CURSOR_NONE` but I don't think it'd have much impact being run for all call sites.

ephphatha (approved):
No sense holding up the auto-merge for a minor nit, it'd be worth going over this to make cursor-related enums scoped types anyway so can do cleanup with that work.

Source/cursor.cpp
glebm: It just seems odd to allow passing a `CURSOR_NONE` into this function, as it's not a valid value for an item cursor, but I'll leave it up to you if you want to refactor this further!

PR #4222, opened 2022-03-23
An option to RLE-compress masked art

glebm:
Saves about 200 KiB RAM in dungeon at a large performance cost.
Almost never a good idea, except for severely RAM-constrained devices (e.g. RG99).

Another option to explore in the future is to convert fonts to CEL at load time. That could be both smaller and faster than SDL RLE.


AJenbo:
> Another option to explore in the future is to convert fonts to CEL at load time. That could be both smaller and faster than SDL RLE.

Yeah I think something like this would be good, then we can have the same render path and load PNG and other formats with out over complicating things.

It would also be nice to do the same with CEL vs CL2.

One thing though, do we actually know that CEL-RLE renders faster then SDL (without RLE)?

glebm:
> One thing though, do we actually know that CEL-RLE renders faster then SDL (without RLE)?

I don't!

AJenbo (approved):

PR #4226, opened 2022-03-24
Drop item in hand before closing stash

### LINKED ISSUE #4205: closing stash while holding an item
There is a possible scenario where the player can lose an item, being forced to exit the game: 
1. select and hold an item with the cursor from the stash
2. close the stash using ESC (or hitting C, Q, I or B) key but:
  A. inventory is full and can't place the item there
  B. tiles around the character are occupied with other items

AJenbo:
Fixes #4205

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -772,41 +772,53 @@ void CheckBtnUp()
 		switch (i) {
 		case PanelButtonCharinfo:
 			QuestLogIsOpen = false;
-			IsStashOpen = false;
-			chrflag = !chrflag;
+			if (BlurStash()) {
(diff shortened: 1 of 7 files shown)


obligaron:
I'm not sure if we should go this route. This adds many new branches.
Can't #3489 also fix #4205 ?

While testing I also discovered other examples where you can also loose a item:
- Don't have free tiles on the ground. Full inventory. Remove a item that needs 50 strength, you have 45 strength and the item adds +10 strength. => Item is lost.
- Multiplayer: Hold a item in the cursor then close the game => item is lost.

PR #4227, opened 2022-03-24
UK TL: Reach 85% of translated content

tsunamistate:
Man, is POEdit Pro useful!

I don't look forward to those last 15% as they have dumb portmanteau monster names like Warpskull or Goretongue…

### DIFF
--- a/Translations/uk.po
+++ b/Translations/uk.po
@@ -5,7 +5,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: DevilutionX\n"
-"POT-Creation-Date: 2022-03-23 13:41+0200\n"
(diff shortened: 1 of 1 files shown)


AJenbo:
> Man, is POEdit Pro useful!

Nice I haven't had much chance to use it my self

> I don't look forward to those last 15% as they have dumb portmanteau monster names like Warpskull or Goretongue…

Yeah, the names of monsters and items are always a bit tricky to handle.

I'm planning to intergrate em-dash and friends in core later today :)

AJenbo:
Did a quick test with the new translation and proper symbol spacing for the extended punctuation symbols. It looked ok for that I saw, but it would be good if you can test things out to see if it looks correct.

PR #4228, opened 2022-03-24
Convert fonts and transparent menu animations to CEL at load-time

glebm:
Reduces RAM usage in-game by 200 KiB with no performance drop, perhaps even an improvement.

The menu graphics change does not affect RAM usage in-game but it does reduce allocator pressure.


glebm:
Rebased

ephphatha (commented):
A couple of minor nits otherwise looks good on pc. No noticeable performance impact or improvement (un?)fortunately, have to see how it goes on more resource constrained platforms.

I was also a little confused by the param order for the AppendCel* functions having the output buffer last. That's more consistent for that space so seems fine after reading the code a bit better 😅

glebm:
Thanks for review! Addressed the comments

Source/DiabloUI/title.cpp
```diff
@@ -13,19 +14,21 @@ std::vector<std::unique_ptr<UiItemBase>> vecTitleScreen;
 void TitleLoad()
 {
 	if (gbIsHellfire) {
+		// This is a 2.4 MiB PCX file without transparency (4.6 MiB as an SDL surface).
```
AJenbo, line 17: Use cl2?

Source/engine/render/cel_render.hpp
```diff
@@ -20,6 +20,18 @@ namespace devilution {
...
 
+/**
+ * @brief Apply the color swaps to a CEL sprite
+ *
+ * Only works on CEL files with frame headers.
+ *
+ * @param p CEL buffer
+ * @param translation Palette translation table
+ * @param numFrames Number of frames in the CEL file.
+ * @param hasFrameHeaders Whether the CEL has frame headers.
```
AJenbo, line 31: It doesn't have these params and the description contradicts the last one.

Source/DiabloUI/title.cpp
glebm: We don't have a CL2 loader, only a CEL one, which won't help with the size here.
We can render PCX directly in a follow-up PR (see https://github.com/diasurgical/devilutionX/issues/4301)

Source/engine/render/cel_render.hpp
glebm: Fixed

Source/DiabloUI/title.cpp
AJenbo: > We don't have a CL2 loader, only a CEL one, which won't help with the size here.

I'm not sure I follow. At some point we didn't have a CEL loader either.

> We can render PCX directly in a follow-up PR (see https://github.com/diasurgical/devilutionX/issues/4301)

Should we try and load CL2 as PCX to simplify things in the rendere?

glebm: In PCX, colors <= 0xBF can be encoded as a single byte, while other colors always require a run-length prefix, which means applying a palette transform requires reallocation.

For PCX, we don't have a renderer yet but we'd only need a simple one that doesn't handle clipping, as we'd only use it for the backgrounds.

Loading PCX as CL2 is extra work/allocation and is not easier than implementing a non-clipping PCX renderer.

AJenbo: > This is out-of-scope for this PR though.

Of cause not, I was just wondering about possible future steps :)

AJenbo (approved):

PR #4230, opened 2022-03-25
Invalidate removed inventory items

AJenbo:


Source/player.cpp
```diff
@@ -1937,34 +1937,35 @@ bool Player::HasItem(int item, int *idx) const
...
 
-	iv--;
+	InvList[iv]._itype = ItemType::None;
+
 	_pNumInv--;
 
 	// If the item at the end of inventory array isn't the one we removed, we need to swap its position in the array with the removed item
 	if (_pNumInv > 0 && _pNumInv != iv) {
 		InvList[iv] = InvList[_pNumInv];
+		InvList[_pNumInv]._itype = ItemType::None;
```
mewmew, line 1954: Looks good. I'd add a `memset(&InvList[_pNumInv], 0, sizeof(ItemStruct))` prior to this line, just to clear more garbage data from memory. And do the same on the above `._itype = ItemType::None;` line.

mewmew (commented):
LGMT, with minor comments.

Source/player.cpp
AJenbo: DevilutionX is C++: `InvList[_pNumInv] = {};` ;)

AJenbo: Going to opt out of zeroing just to keep thing simple for now.

PR #4231, opened 2022-03-26
Remove `(un)lock_buf`

glebm:
We do not seem to render from multiple threads, so these calls are unnecessary.


StephenCWills (approved):
dthread and nthread are the only threads other than the main thread. Those only deal with network messages so only the main thread should be using these buffers.

StephenCWills:
Oops, I didn't notice it was set to auto-merge. I shouldn't have approved it.

AJenbo:
It's probably ok :)

PR #4238, opened 2022-03-26
[gamepad] Fix managing belt

AJenbo:
(broken when adding stash)


obligaron (commented):
lgtm, but I'm not familiar with the controller logic.

I can only think about one case where we still have a difference between this pr and pre-stash logic. If inventory is open and the cursor is not placed on the inventory or belt (main panel) we still don't execute the path but previous we did. But if this is a problem or not I'm unsure.

StephenCWills (approved):
> I can only think about one case where we still have a difference between this pr and pre-stash logic. If inventory is open and the cursor is not placed on the inventory or belt (main panel) we still don't execute the path but previous we did. But if this is a problem or not I'm unsure.

I did a bit of testing on the pre-stash code, and this logic doesn't do anything when the mouse isn't over one of these panels anyway.

PR #4237, opened 2022-03-26
[gamepad] Avoid OOB on empty belt

AJenbo:

PR #4239, opened 2022-03-26
Add gamepad support for stash

### LINKED ISSUE #4211: [gamepad] Stash navigation
The stash is currently not usable using a gamepad

L1 and L2 for switching pages, and L1+L2 (if possible) for withdrawing gold

image

image

AJenbo:
Resolves #4211 and an issue with the cursor not resetting correctly in the inventory (see `InvGetEquipSlotCoordFromInvSlot()`)

- [x] Cursor is misaligned when jumping from inventory stash (drops by one row)
- [x] Cursor is misaligned when jumping between stash and inventory (if there is an item larger then 1x1 at the destination)
- [x] Cannot jump between stash and inventory if edge item is 2 in width
- [x] Empty cursor is misaligned in stash grid (pointer is at upper left instead of center)
- [x] Miss aligned when jumping to belt from stash (lower left instead of center)
- [x] L1/L2 for changing page
- [x] SELECT for withdrawing gold
- [x] Secondary action should transfer items between stash and inventory

Optional:
- [ ] Center cursor when hovering items in stash (same as inventory)
- [ ] It's possible to move in to the wrong equipment slots when jumping from stash while holding an item

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -60,6 +60,7 @@ namespace {
 int Slot = SLOTXY_INV_FIRST;
 Point ActiveStashSlot = InvalidStashPoint;
 int PreviousInventoryColumn = -1;
+bool BeltReturnsToStash = false;
 
 const Direction FaceDir[3][3] = {
 	// NONE             UP                DOWN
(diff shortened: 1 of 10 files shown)


Source/inv.cpp
```diff
@@ -1174,7 +1157,8 @@ bool CanBePlacedOnBelt(const Item &item)
 {
 	return FitsInBeltSlot(item)
 	    && item._itype != ItemType::Gold
-	    && item._iStatFlag
+	    && MyPlayer->GetItemLocation(item) == ILOC_UNEQUIPABLE
+	    && MyPlayer->CanUseItem(item)
```
AJenbo, line 1161: I'm not sure this is needed, copied it from `CheckInvPaste()` for now.

Source/controls/plrctrls.cpp
```diff
@@ -1508,18 +1646,16 @@ void plrctrls_after_game_logic()
...
+		auto &myPlayer = Players[MyPlayerId];
+		const int id = AllItemsList[myPlayer.SpdList[i].IDidx].iMiscId;
+		const int spellId = AllItemsList[myPlayer.SpdList[i].IDidx].iSpell;
+		if ((type == BLT_HEALING && (id == IMISC_HEAL || id == IMISC_FULLHEAL || (id == IMISC_SCROLL && spellId == SPL_HEAL)))
+		    || (type == BLT_MANA && (id == IMISC_MANA || id == IMISC_FULLMANA))
+		    || id == IMISC_REJUV || id == IMISC_FULLREJUV) {
+			if (!myPlayer.SpdList[i].isEmpty()) {
+				UseInvItem(MyPlayerId, INVITEM_BELT_FIRST + i);
+				break;
+			}
```
StephenCWills, line 1658: This reverts #4237

StephenCWills (changes requested):

Source/inv.cpp
StephenCWills: `CanUseItem()` is for scrolls and runes. The `ILOC_UNEQUIPABLE` thing appears to be superfluous, unless there's a usable ring or amulet in the game that I don't know about. 🤷

AJenbo: `CanUseItem()` should also be unessesery if `_iStatFlag` is up to date, but we are also moving away from it, so might as well use `CanUseItem()` for safety, even if the current way I did it is a bit ugly :)

StephenCWills (approved):

PR #4243, opened 2022-03-27
Fix multiplayer equipment synchronization

StephenCWills:
From #4182
> When auto equiping items the other player will sometimes not see that it has been equiped

This is caused by the fact that `NetSendCmdChItem()` uses `myPlayer.HoldItem` to populate the network message with item creation info. When auto-equipping items, the local client places the item in the appropriate equipment slot directly from the list of items on the ground. The network message ends up using whatever data is left over from the last item the player was holding.

The idea here is to change `NetSendCmdChItem()` to use data from `player.InvBody` instead. This means the item must be equipped locally before attempting to send the network message. The `ChangeEquipment()` function was introduced to ensure that the network message is only sent under the appropriate conditions. Since some of the equip logic also uses `NetSendCmdDelItem()` and those network messages are quite similar, I also introduced `RemoveEquipment()` as an additional cleanup.

---

I also found an unreported issue while testing. When using shift+click on one-handed weapons or shields in the player's inventory, the game would force the shield into the player's right hand before attempting to determine which hand the inventory item should be placed in. This swap wasn't being synchronized between clients, causing some confusion on the remote client. Based on how the logic was determining which hand the inventory item should be placed in, it doesn't appear that the swap was really necessary so I simply removed it.

I also went ahead and modified an edge case in the two-handed weapon logic to explicitly send `CMD_DELPLRITEMS` for the equipment in the player's right hand. It didn't seem like it was entirely necessary because the remote client would figure it out when receiving `CMD_CHANGEPLRITEMS` for the two-handed weapon, but this behavior matches the logic in `CheckInvPaste()`.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -262,6 +262,15 @@ bool CanEquip(Player &player, const Item &item, inv_body_loc bodyLocation)
 	}
 }
 
+void ChangeEquipment(Player &player, inv_body_loc bodyLocation, const Item &item)
+{
+	player.InvBody[bodyLocation] = item;
+
+	if (&player == MyPlayer) {
+		NetSendCmdChItem(false, bodyLocation);
+	}
+}
+
 bool AutoEquip(int playerId, const Item &item, inv_body_loc bodyLocation, bool persistItem)
 {
 	auto &player = Players[playerId];
@@ -271,13 +280,12 @@ bool AutoEquip(int playerId, const Item &item, inv_body_loc bodyLocation, bool p
 	}
 
 	if (persistItem) {
-		player.InvBody[bodyLocation] = item;
+		ChangeEquipment(player, bodyLocation, item);
 
 		if (*sgOptions.Audio.autoEquipSound && playerId == MyPlayerId) {
 			PlaySFX(ItemInvSnds[ItemCAnimTbl[item._iCurs]]);
 		}
 
-		NetSendCmdChItem(false, bodyLocation);
 		CalcPlrInv(player, true);
 	}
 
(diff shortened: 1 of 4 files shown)


Source/inv.cpp
```diff
@@ -432,65 +440,77 @@ void CheckInvPaste(int pnum, Point cursorPosition)
...
-					NetSendCmdChItem(false, INVLOC_HAND_LEFT);
-					player.InvBody[INVLOC_HAND_LEFT] = player.HoldItem;
+					ChangeEquipment(player, INVLOC_HAND_LEFT, player.HoldItem);
 				} else {
-					NetSendCmdChItem(false, INVLOC_HAND_RIGHT);
-					cn = SwapItem(player.InvBody[INVLOC_HAND_RIGHT], player.HoldItem);
+					Item previouslyEquippedItem = player.InvBody[INVLOC_HAND_RIGHT];
+					ChangeEquipment(player, INVLOC_HAND_RIGHT, player.HoldItem);
+					player.HoldItem = previouslyEquippedItem;
+					cn = previouslyEquippedItem._iCurs + CURSOR_FIRSTITEM;
```
qndel, line 461: these 4 lines repeat a lot, maybe worth making some function for that?

StephenCWills: I do think that was the idea behind `SwapItem()`, but I found the pattern significantly more confusing because of the return value. Rather than introducing a function for these four lines, I'd rather see a cleanup of `CheckInvPaste()` that deduplicates all the nested conditionals in the `ILOC_ONEHAND` case, which should clear up basically all of this duplication as well. I had to resist the urge to do that cleanup in this PR, as I figured it was more of a 1.5.0 thing.

Perhaps I'll submit another PR based on top of this one so you can see what I'm talking about.

AJenbo (approved):
The two handed logic is super complicated to reason about 😵‍💫

Source/inv.cpp
AJenbo: I had the same though but also came to the same conclusion

StephenCWills: FYI, #4246

Issue #4217, opened 2022-03-22
Cursor jumps to bottom right corner of screen when selecting item in inventory 


DrGonzo564:
**Important information**
Mac OS 12.3 / DevilutionX 1.4b / Compiled (Artifact came from https://github.com/diasurgical/devilutionX/pull/4212 )

**Describe**
When selecting an item from my inventory the cursor will jump to the bottom right corner of the screen 

**To Reproduce**
1. Go to inventory
2. Click on item
3. Cursor jumps to bottom right corner of screen

**Additional context**
5K iMac Radeon Pro 580 maybe issue because high DPI monitor 27" ??

https://user-images.githubusercontent.com/82344993/159406491-2656668b-e310-4b1e-ba6c-11bfba9e0369.mov

bubio:
https://github.com/diasurgical/devilutionX/blob/af59f7e254b18ea47d98112b5e64778e46d97923/Source/utils/display.h#L93

There seems to be a problem with the scale value that can be obtained with the SDL_RenderGetScale function.

PR #4244, opened 2022-03-27
DPI Scale is now included in the calculation.

bubio:
Fixed issue #4217

Tested on:
MacBook Pro 2017 / macOS 12.3 / Retina
Note PC / Zorin OS / Non-Retina


AJenbo:
Nicely done, I wasn't sure this was going to happen in time for the release considering I don't have the needed hardware :)

PR #4245, opened 2022-03-27
Close gold withdraw when closing stash

AJenbo:


StephenCWills (approved):
Makes me feel like maybe we should introduce a `CloseStash()` function similar to the `CloseInventory()` function so these two lines will always go together, but I'll leave that up to you.

PR #4246, opened 2022-03-27
Clean up logic for 1-handed weapons in CheckInvPaste()

StephenCWills:
It _should_ be this simple, but the original logic had a lot of branches so it's not really clear. Since this is just a cleanup, I would recommend pushing this to 1.5.0, unless someone feels significantly more confident about this than I do.


ephphatha (approved):
Testing this one is pretty easy :D, swapping weapons seems to work exactly the same in master as in this branch.

PR #4249, opened 2022-03-27
Remove vegtable dialog

AJenbo:
It's kinda silly and may have been deliberately disabled

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -3950,10 +3950,6 @@ void PlayDungMsgs()
 	} else if (currlevel == 13 && !myPlayer._pLvlVisited[13] && (myPlayer.pDungMsgs & DungMsgHell) == 0) {
 		myPlayer.Say(HeroSpeech::IMustBeGettingClose, 40);
 		myPlayer.pDungMsgs |= DungMsgHell;
-	} else if (currlevel == 16 && !myPlayer._pLvlVisited[16] && (myPlayer.pDungMsgs & DungMsgDiablo) == 0) {
-		sfxdelay = 40;
-		sfxdnum = PS_DIABLVLINT;
-		myPlayer.pDungMsgs |= DungMsgDiablo;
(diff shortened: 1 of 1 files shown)


qndel:
it's not silly, it's classic! + not everyone knows what it says backwards

StephenCWills:
I'm not sure I understand the reasoning behind the conclusion that this may have been deliberately disabled. [You said](https://github.com/diasurgical/devilution/pull/454#issuecomment-436787042) it was broken in vanilla v1.0.0, so were you thinking they deliberately didn't fix it in any of their patches? Maybe it was just a low-priority bug that few ever noticed and no one ever bothered to resolve.

qndel:
Or like galaxy wrote 
> "Funny enough, I never realized this was broken, because I always use the command line to start at level 16 directly, therefore 15 is still "never visited"."

Maybe it was the exact same case with them.

AJenbo:
> it's not silly, it's classic! + not everyone knows what it says backwards

I think it's silly not even knowing what it is saying :shrug:

AJenbo:
> [You said](https://github.com/diasurgical/devilution/pull/454#issuecomment-436787042) it was broken in vanilla v1.0.0, so were you thinking they deliberately didn't fix it in any of their patches?

Like I said in that quote, it was never functional in any version. My thinking is that they deliberately changed the code to be non function before release in order to disable it.

qndel:
> > [You said](https://github.com/diasurgical/devilution/pull/454#issuecomment-436787042) it was broken in vanilla v1.0.0, so were you thinking they deliberately didn't fix it in any of their patches?
> 
> Like I said in that quote, it was never functional in any version. My thinking is that they deliberately changed the code to be non function before release in order to disable it.

Even they weren't that special to leave code in and just break it on purpose xD

AJenbo:
🥕 🪥

PR #4253, opened 2022-03-28
Fix cow quest getting stuck in multiplayer

### LINKED ISSUE #4184: Hellfire Nut thinks he already gave the rune
To replicate
- enable cow quest
- start multiplayer game with level 24 hero
- enter level 8
- talk to the nut in town

He will reply with "what are you wasting time for...", but has not yet given you the bomb meaning the quest/hive is now inaccessible.

AJenbo:
Fixes #4184

This turned out to be a vanilla hellfire bug. When loading level 8 the quest would be set as active, like for most other quests. But this quests should only be set active when the player is handed the rune bomb by the Nut in town. This would lead the logic to think the Nut had already given you the bomb and so the quest would be stuck in this state.

The solution is quite simple, simply do not enable this quest during loading of level 8.

Also I removed the specific dialog line for `QUEST_ACTIVE` as it is the same as the default dialog.

### DIFF
--- a/Source/quests.cpp
+++ b/Source/quests.cpp
@@ -658,12 +658,6 @@ void ResyncMPQuests()
 		nakrulQuest._qactive = QUEST_ACTIVE;
 		NetSendCmdQuest(true, nakrulQuest);
 	}
-
-	auto &cowQuest = Quests[Q_JERSEY];
-	if (cowQuest._qactive == QUEST_INIT && currlevel == cowQuest._qlevel - 1) {
-		cowQuest._qactive = QUEST_ACTIVE;
-		NetSendCmdQuest(true, cowQuest);
-	}
 }
 
 void ResyncQuests()
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -671,9 +671,6 @@ void TalkToCowFarmer(Player &player, Towner &cowFarmer)
 		if (gbIsMultiplayer)
 			NetSendCmdQuest(true, quest);
 		break;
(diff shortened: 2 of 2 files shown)


Source/quests.cpp
```diff
@@ -658,12 +658,6 @@ void ResyncMPQuests()
 		nakrulQuest._qactive = QUEST_ACTIVE;
 		NetSendCmdQuest(true, nakrulQuest);
 	}
-
-	auto &cowQuest = Quests[Q_JERSEY];
-	if (cowQuest._qactive == QUEST_INIT && currlevel == cowQuest._qlevel - 1) {
-		cowQuest._qactive = QUEST_ACTIVE;
```
ephphatha, line 664: Potentially this was meant to be set to QUEST_HIVE_ACTIVE to indicate the bomb can be spawned? The intent might've been to skip the tease states and just jump straight into the quest.

AJenbo: I think it's more likely that they simply copied this section for all the Hellfire quests and never really tested multiplayer, since all Hellfire quests are horribly broken in multiplayer and no other quest is fast forwarded by this section of the code.

ephphatha (approved):

Source/quests.cpp
AJenbo: On a side note the Jersey quest also uses _qactive to track the quest state rater then using _qvar1/2 meaning that it is the only quest that has _qactive values other then inactive/active/completed.

PR #4254, opened 2022-03-28
Update zlib dependency to 1.2.12

AJenbo:
They deleted the 1.2.11 src from there site, a bit agressive. Let's hope noting major changed :/

### DIFF
--- a/3rdParty/zlib/CMakeLists.txt
+++ b/3rdParty/zlib/CMakeLists.txt
@@ -2,8 +2,8 @@ include(functions/FetchContent_MakeAvailableExcludeFromAll)
 
 include(FetchContent)
 FetchContent_Declare(zlib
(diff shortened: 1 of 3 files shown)


StephenCWills (approved):

tsunamistate:
From zlib website:

Version 1.2.12 has these key improvements over 1.2.11:

-  Fix a deflate bug when using the Z_FIXED strategy that can result in out-of-bound accesses.
-  Fix a deflate bug when the window is full in deflate_stored().
-  Speed up CRC-32 computations by a factor of 1.5 to 3.
-  Use the hardware CRC-32 instruction on ARMv8 processors.
-  Speed up crc32_combine() with powers of x tables.
-  Add crc32_combine_gen() and crc32_combine_op() for fast combines. 

Due to the bug fixes, any installations of 1.2.11 should be replaced with 1.2.12.

It's just a bugfix release, no breaking changes

PR #4256, opened 2022-03-29
Cancel heal other/resurrect on right-click

### LINKED ISSUE #4252: There is no way to cancel heal other/resurrect
We have made changes that makes heal other/resurrect only trigger when a player is targeted (to avoid wasting the spell). But this sometimes leaves the player in a bad situation if there are no valid targets on screen.

A simple solution that still serves to not waste the spell would be to have right click cancel the spell. Which makes sense since that is also how you activate it to begin with.

ephphatha:
fixes #4252

> We have made changes that makes heal other/resurrect only trigger when a player is targeted (to avoid wasting the spell).

The cast spell code fires in `TryIconCurs()` for both left and right clicks (and cursor driven gamepad actions with the inventory open). In the current code these spells are "cast" with no target, then when the local client receives the message the cursor is changed back to the hand only if there was a valid target. In a roundabout way this leaves the spell readied unless the user does something else which triggers a cursor change (like interacting with an inventory cell).

This change restricts casting resurrect/heal other only with a valid target. This ends up making left mouse clicks progress through the fallback cases that allow moving with a readied spell, right clicks eventually fall through to reset the cursor and "unready" the spell. The mana is spent (or the scroll is consumed) as soon as the spell is readied which makes it harder to restore the source, so I haven't bothered to try track and undo that state change.

### DIFF
--- a/Source/player.h
+++ b/Source/player.h
@@ -22,6 +22,7 @@
 #include "spelldat.h"
 #include "utils/attributes.h"
 #include "utils/enum_traits.h"
+#include "utils/stdcompat/algorithm.hpp"
 
 namespace devilution {
 
@@ -554,6 +555,20 @@ struct Player {
 	 * Valid only for players with Mana Shield spell level greater than zero.
 	 */
 	int GetManaShieldDamageReduction();
+
+	/**
+	 * @brief Gets the effective spell level for the player, considering item bonuses
+	 * @param spell spell_id enum member identifying the spell
+	 * @return effective spell level
+	 */
+	int GetSpellLevel(spell_id spell) const
+	{
+		if (spell == SPL_INVALID || static_cast<std::size_t>(spell) >= sizeof(_pSplLvl)) {
+			return 0;
+		}
+
+		return std::max<int8_t>(_pISplLvlAdd + _pSplLvl[static_cast<std::size_t>(spell)], 0);
+	}
 
 	/**
 	 * @brief Return monster armor value after including player's armor piercing % (hellfire only)
--- a/Source/spells.cpp
+++ b/Source/spells.cpp
@@ -247,10 +247,6 @@ void CastSpell(int id, spell_id spl, int sx, int sy, int dx, int dy, int spllvl)
 
 void DoResurrect(int pnum, uint16_t rid)
 {
-	if (pnum == MyPlayerId) {
-		NewCursor(CURSOR_HAND);
-	}
-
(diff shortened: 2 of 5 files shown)


Source/spells.cpp
```diff
@@ -293,42 +289,32 @@ void DoResurrect(int pnum, uint16_t rid)
...
-	} else if (player._pClass == HeroClass::Monk) {
+	} else if (caster._pClass == HeroClass::Monk) {
 		hp *= 3;
 	}
 
 	target._pHitPoints = std::min(target._pHitPoints + hp, target._pMaxHP);
 	target._pHPBase = std::min(target._pHPBase + hp, target._pMaxHPBase);
 
-	if (rid == MyPlayerId) {
+	if (std::addressof(target) == MyPlayer) {
```
AJenbo, line 317: Is there a good reason to start using `std::addressof` here?

Source/player.h
```diff
@@ -555,6 +556,20 @@ struct Player {
...
 	int GetManaShieldDamageReduction();
 
+	/**
+	 * @brief Gets the effective spell level for the player, considering item bonuses
+	 * @param spell spell_id enum member identifying the spell
+	 * @return effective spell level
+	 */
+	int GetSpellLevel(spell_id spell) const
+	{
+		if (spell == SPL_INVALID || static_cast<std::size_t>(spell) >= sizeof(_pSplLvl)) {
```
AJenbo, line 566: I'm not sure we really need this check when already limiting it to `spell_id`, we could possibly define `_pSplLvl` as `_pSplLvl[SPL_LAST]` instead.

```suggestion
		if (spell == SPL_INVALID) {
```

ephphatha: With the way enums are handled in C++ any 6 bit value is valid for spell_id and because the backing type is specified potentially any 8 bit value could be passed in to this function without triggering undefined behaviour. This is overly defensive for the current use case given it's called with a known value, it's more for future handling of loaded hotkeys given there's no validation performed on the way in.

Source/spells.cpp
ephphatha: I actually thought we'd been using this already 😅 . There's no real need given there's no `Player::operator&` overload so either way works.

AJenbo:
```suggestion
	if (&target == MyPlayer) {
```

AJenbo:
Nice, that solves the last of the release blocking bugs :)

PR #4257, opened 2022-03-29
Add OptionEntry for alternative nest art

obligaron:
… and remove command line parameter `--nestart`.

This enables platforms that can't use command line parameter (for example android) to use the alternative nest art.
It's also easier to access cause it's now part of the settings menu.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -101,7 +101,6 @@ bool gbLoadGame;
 bool cineflag;
 int force_redraw;
 int PauseMode;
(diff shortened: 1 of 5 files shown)


AJenbo:
For documentation :)

Normal level 1
image
Alternative level 1
image

Normal level 2
image
Alternative level 2
image

Normal level 3
image
Alternative level 3
image

Normal level 4
image
Alternative level 4
image

AJenbo (approved):

PR #4259, opened 2022-03-29
[Android] Add shortcut to import game data

### LINKED ISSUE #3990: [Android] Option to import the MPQ file via Scoped Storage
A lot of users are having a hard time figuring out how to navigate to and access the app folder on Android (thanks Google), to help users more easily transfer the data we could have a button on the installation screen for opening a file picker that will let them navigate to where they might have put the file and allow them to select it. Then we can copy the file using Java (which has access to scoped storage) so that it is available for the game.

https://developer.android.com/about/versions/11/privacy/storage

Additionally it would be good to have this accessible as a short cut (see https://github.com/diasurgical/devilutionX/issues/3197)

StephenCWills:
Adds an "Import Data" shortcut that opens the file browser to select files that will be copied into the data folder.

image

This is functional, but it's missing an icon and is rather lacking in user feedback.

This resolves #3990


StephenCWills:
I fixed a bunch of whitespace and removed some debugging code I forgot about. Also, FYI, tested on Pixel 4 (Android 12).

AJenbo:
outline_drive_file_move_rtl_black_48

android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
```diff
@@ -0,0 +1,94 @@
...
+			InputStream inputStream = null;
+			OutputStream outputStream = null;
+
+			try {
+				ContentResolver contentResolver = getContentResolver();
+				inputStream = contentResolver.openInputStream(fileUri);
+				outputStream = new FileOutputStream(externalFilesPath);
+
+				// Transfer bytes from in to out
+				byte[] buf = new byte[1024];
```
glebm, line 79: nit: 4096 is the default block size on most filesystems these days

AJenbo (dismissed):

StephenCWills:
image

PR #4262, opened 2022-03-30
Update zh_CN.po

yjfyy:
Update translation

### DIFF
--- a/Translations/zh_CN.po
+++ b/Translations/zh_CN.po
@@ -884,7 +884,7 @@ msgstr "墓穴: {:d} 层"
 
 #: Source/automap.cpp:509 Source/items.cpp:2034
 msgid "Level: {:d}"
-msgstr "等级: {:d}"
+msgstr "层数: {:d}"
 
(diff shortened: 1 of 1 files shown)


AJenbo:
@tytannial @4aron5un could one of you review this? Also if you have time to update the translation that would be appreciated as we are about to make a new release :)

tytannial:
> @tytannial @4aron5un could one of you review this? Also if you have time to update the translation that would be appreciated as we are about to make a new release :)

I will take time for review :D

tytannial:
I problem....#: Source/automap.cpp:509 Source/items.cpp:2034
Will this text effect the show on items? if the text on item should be "等级" means how much level you need.
The text on map, means which dungeon "level" you are currently at..."层数"
We use completely different word to explain that.....

AJenbo:
> Will this text effect the show on items? if the text on item should be "等级" means how much level you need.

The text appears on ears that you have gotten for killing another player.
image

tytannial:
Yes.....but this text also be used on automap...for display current dungeon level text tip...🤣

Issue #4206, opened 2022-03-20
hardware cursor outline


ikonomov:
This has been discussed, but I couldn't find it reported here. 

There is a brown outline around the cursor when hardware cursor is enabled.

glebm:
Platform / GPU / Resolution / Scale?

qndel:
> Platform / GPU / Resolution / Scale?

it doesn't need anything special, I have it too

qndel:
[diablo.zip](https://github.com/diasurgical/devilutionX/files/8355628/diablo.zip)
my config

glebm:
> it doesn't need anything special, I have it too with default settings - think all it needs is hardware cursor on.

I can't reproduce this though

AJenbo:
![20220326_154233](https://user-images.githubusercontent.com/204594/160244569-e7dac877-682f-4674-ae17-e16f6b11d7c5.jpg)

AJenbo:
It's clearly visible in town. You of cause also need to be up scaling. The screenshot here is taken at 3440x1440 (3x).

AJenbo:
![16483060427933383947881098899763](https://user-images.githubusercontent.com/204594/160244708-bd19053b-06d9-42ea-bf3c-8be9369220dd.jpg)

In the dungeon it's not as easy to see since the color matches pretty well with the rest of the cursor

glebm:
Thanks, reproduced

glebm:
Probably a bug in `BilinearScale32` around transparent pixels

AJenbo:
I'm thinking it's bleeding in a color from the palette for the transparent area, and maybe if we set it to black the result would be better, but not sure how that part works at all.

NiteKat:
I noticed this while streaming yesterday, was pretty noticable in town and in catacombs. 1.5.0, Windows, using OBS and hardware cursor.

PR #4263, opened 2022-03-30
Fix underflow in `BilinearScale32`

glebm:
Fixes funky hardware cursor colors during fade transitions.

Refs #4206


ephphatha (approved):
Fixes the flicker on the cursor just before it becomes fully transparent during a level transition, assuming it's the same for any other instance it occurs but that was the only one I could trigger on master

PR #4274, opened 2022-03-31
Remove calls to NewCursor() from store functions

StephenCWills:
@ikonomov reported this issue on Discord.
> in Hellfire single player when I try to buy an item from Pepin the gold is spent, but no item is transferred to the inventory
> and at Gris and Adria when I buy an item it appears in the inventory, but when I try to interact with the newly bought item it disapears

The culprit here is the code introduced into `NewCursor()` by #4239. Or perhaps it would be more accurate to say it's the stores' abuse of the `Player::HoldItem` variable. Regardless, the calls to `NewCursor()` in each of the `*BuyEnter()` functions was clearing the item type of the hold item, causing the item to disappear after it was copied from there into the player's inventory.

I suspect that the reason the stores were setting the cursor in the first place was to populate the `icurseSize28` variable so it could be used to determine whether the player has room for the item in their inventory. Nowadays, the `AutoPlaceItemInInventory()` function uses the `GetInventorySize()` function which computes the size from the item's cursor data without having to actually update the cursor. Removing the `NewCursor()` calls resolves the bug and doesn't seem to have any adverse effects.

### DIFF
--- a/Source/stores.cpp
+++ b/Source/stores.cpp
@@ -1400,7 +1400,6 @@ void SmithBuyEnter()
 	}
 
 	myPlayer.HoldItem = smithitem[idx];
-	NewCursor(myPlayer.HoldItem._iCurs + CURSOR_FIRSTITEM);
 
 	bool done = AutoEquipEnabled(myPlayer, myPlayer.HoldItem) && AutoEquip(MyPlayerId, myPlayer.HoldItem, false);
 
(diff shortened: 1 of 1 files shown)


StephenCWills:
Actually, it occurred to me that I should probably go ahead and test buying an item when there's no space in my inventory. It seems like it doesn't stop me anymore, so maybe there are adverse effects.

AJenbo (approved):
Hopefully we can at some point make an even tighter coupling between the cursor and HoldItem so as to make it's usage simpler and consistent :)

PR #4276, opened 2022-03-31
Don't use HoldItem in CMD_PUTITEM handler

StephenCWills:
Similar to the issue mentioned in #4274, dropping items on the ground causes the hold item to lose its item type. In this case, the game tries to use the hold item in the handler for the `CMD_PUTITEM` network message, but the cursor was cleared in the previous frame when generating the network message.

Of course, the game only uses the hold item on the local client because remote clients would not have access to that data. Instead, remote clients must recreate the item using the item information included in the network message, which was copied from the hold item when generating the network message. The idea behind this change is to go ahead and recreate the item on the local client as well instead of abusing the hold item as swap space.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1836,14 +1836,14 @@ bool TryInvPut()
 	return CanPut(myPlayer.position.tile);
 }
 
-int InvPutItem(Player &player, Point position)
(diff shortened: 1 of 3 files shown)


Source/inv.cpp
```diff
@@ -1855,14 +1855,34 @@ int InvPutItem(Player &player, Point position)
 	assert(CanPut(position));
 
 	int ii = AllocateItem();
+	auto &item = Items[ii];
```
ephphatha, line 1858: The rest of this procedure is now functionally identical to SyncDropItem (except for the NewCursor call which really should be moved out) - could dedupe the logic by calling SyncDropItem directly.

ephphatha (approved):

AJenbo (approved):

PR #4282, opened 2022-03-31
German Translation

ChaosMarc:
Just opening this, so my progress is shown. There are still 55 translations to go.

### DIFF
--- a/Translations/de.po
+++ b/Translations/de.po
@@ -197,7 +197,7 @@ msgstr "Verkauf International"
 
 #: Source/DiabloUI/credits_lines.cpp:158
 msgid "U.S. Sales"
-msgstr "Verkauf U.S."
+msgstr "US-Verkäufe"
 
 #: Source/DiabloUI/credits_lines.cpp:161
 msgid "Manufacturing"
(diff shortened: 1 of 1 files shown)


AJenbo:
Two errors found:

Belt item {} != Gürtelgegenstand
{:s} {:s} != {:d} {:s}

ChaosMarc:
Ok, I'm done for now. I have some questions though:

image
this was not changed by me, but it looks wrong :D should I use a not so often used abbreviation "Geschw." or should this be fixed by making the menu multiline compatible aka a bug ticket?

image
it seems like I found a row length / word combination that collides with the selection spinner. again: should I "fix" this in the translation or is this worth a bug ticket?

AJenbo:
>  this was not changed by me, but it looks wrong :D should I use a not so often used abbreviation "Geschw." or should this be fixed by making the menu multiline compatible aka a bug ticket?

We cant really extend the menu any further since it's already taking up all the screen space as the lowest resolution and doesn't have a scroll feature.

AJenbo:
>  it seems like I found a row length / word combination that collides with the selection spinner. again: should I "fix" this in the translation or is this worth a bug ticket?

See if you can't find a shorter way to phrase it. It's ok if it's not exhaustive, people can read the description at the bottom if they aren't sure what it does.

ChaosMarc:
image
I've abbreviated all usages of "automatisch" (automatic) with "autom." for the menu items to be consistent. the description still uses the unabbreviated word.

with this change I think this PR can be merged

AJenbo:

ChaosMarc:
Sorry I'm really busy at the moment. I will try and squeeze it in if I can, but no promises 😇

AJenbo:
Thanks, also I got this suggestion for an update to it if you can evaluate it: https://github.com/diasurgical/devilutionX/pull/6256

(just one line)

ChaosMarc:
😂 I already did 😇

PR #4287, opened 2022-03-31
Add helpers to change stash pages

ephphatha:
The first commit addresses a bug I introduced in #4265 and #4281 where the stash page counter can underflow allowing navigation from page 1 directly to page 50. Splitting the other helpers to a separate PR since they're independent.

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -536,6 +536,7 @@ void StashStruct::RemoveStashItem(uint16_t iv)
 void StashStruct::SetPage(unsigned newPage)
 {
 	page = std::min(newPage, LastStashPage);
+	dirty = true;
 }
 
 void StashStruct::NextPage(unsigned offset)
@@ -545,6 +546,7 @@ void StashStruct::NextPage(unsigned offset)
 	} else {
 		page = LastStashPage;
 	}
+	dirty = true;
 }
 
 void StashStruct::PreviousPage(unsigned offset)
@@ -554,6 +556,7 @@ void StashStruct::PreviousPage(unsigned offset)
 	} else {
 		page = LastStashPage;
 	}
+	dirty = true;
 }
 
 void StashStruct::RefreshItemStatFlags()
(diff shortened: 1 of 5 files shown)


Source/qol/stash.cpp
```diff
@@ -537,6 +533,20 @@ void StashStruct::SetPage(unsigned newPage)
 	page = std::min(newPage, CountStashPages - 1);
 }
 
+void StashStruct::NextPage(unsigned offset)
+{
+	SetPage(page + offset);
```
obligaron, line 538: Is this the last caller to `SetPage`?
If yes, should we inline `SetPage` and remove the function?

ephphatha: How's this version look? I updated prev/nextpage to be more defensive and realised that we can save/load the current stash page to make use of this function. It's entirely backwards compatible with the current stash format, can be dropped if desired.

obligaron: lgtm 🙂 

For saving the active page:
I was also thinking of adding this to the normal settings/options (`diablo.ini`).
But it can also be saved along the normal stash items.
If we want to save it with the stash, doesn't we need to set the `dirty` flag on the stash?

ephphatha: I think it's best to save it with the stash archive itself, purely because there's a single player stash and a multiplayer stash which contain different items. Presumably the player will expect to end up on the last used page in each context.

Good point about the dirty flag, I hadn't picked up on that and was relying on modifications like adding/removing an item to trigger saving a page. I think it makes sense to mark dirty on page change directly so it gets saved even with no other changes.

obligaron: When adding to `diablo.ini` we would have needed multiple entries (diablo sp, diablo mp, hellfire sp, hellfire mp).
The only pro I can think about, is that this setting wouldn't be reverted when loading a game in sp. But even this can be good or bad.

obligaron (approved):

AJenbo (approved):

PR #4288, opened 2022-03-31
Czech Translation update for 1.4.0

Psojed:


AJenbo:

PR #4291, opened 2022-04-01
Monks do not automatically equip shields

ikonomov:
Monks suffer a penalty to unarmed damage (1/2) whenever they have a shield or a weapon equipped so likely players that prefer this playstyle wouldn't want to have both weapons and shields automatically equipped.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1402,9 +1402,7 @@ bool AutoEquip(int playerId, const Item &item, bool persistItem)
 bool AutoEquipEnabled(const Player &player, const Item &item)
 {
 	if (item.isWeapon()) {
-		// Monk can use unarmed attack as an encouraged option, thus we do not automatically equip weapons on him so as to not
-		// annoy players who prefer that playstyle.
-		return player._pClass != HeroClass::Monk && *sgOptions.Gameplay.autoEquipWeapons;
+		return *sgOptions.Gameplay.autoEquipWeapons;
 	}
 
 	if (item.isArmor()) {
@@ -1416,9 +1414,7 @@ bool AutoEquipEnabled(const Player &player, const Item &item)
 	}
 
 	if (item.isShield()) {
-		// Monk can use unarmed attack as an encouraged option, thus we do not automatically equip shields on him so as to not
-		// annoy players who prefer that playstyle.
-		return player._pClass != HeroClass::Monk && *sgOptions.Gameplay.autoEquipShields;
+		return *sgOptions.Gameplay.autoEquipShields;
 	}
 
 	if (item.isJewelry()) {
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1416,7 +1416,9 @@ bool AutoEquipEnabled(const Player &player, const Item &item)
 	}
 
 	if (item.isShield()) {
-		return *sgOptions.Gameplay.autoEquipShields;
+		// Monk can use unarmed attack as an encouraged option, thus we do not automatically equip shields on him so as to not
+		// annoy players who prefer that playstyle.
+		return player._pClass != HeroClass::Monk && *sgOptions.Gameplay.autoEquipShields;
 	}
 
 	if (item.isJewelry()) {


ephphatha (dismissed):
Not really related to this PR, I noticed the weapon check doesn't allow for Staves to be auto-equipped but that doesn't appear to cause a negative impact to monk gameplay (damageMod is the higher value for both staff and empty hands). Should that check allow the monk to auto-equip staff items?

ikonomov:
Also monks suffer a penalty to armor depending on its type, so maybe that check can be added as well...

AJenbo:
This is why we originally ended up making each type configurable. If we additionally also make it automatically handle things then it make the individual configurations redundant, or robs people of the ability to set there preference.

ikonomov:
This makes sense, but I think for a Monk in particular weapons and shields should be automatically disabled from being picked up, because an unarmed monk is a viable and for some people preferable playing style.  And unarmed in this context means having no shield equipped so I think if weapons have an automatic check for a Monk it makes sense to have the same for shields.  If we only rely on the the manual configuration then it could necessitate having to switch those configurations each time a player decides to play another class.  For armor I agree with you and I think it is best left to the manual configuration alone, because unlike weapons and shields, monks can wear some types of armor with some or no penalty so I think it would be safe to assume that most people will have an armor equipped.  I mentioned armor because I wanted to bring all details to the discussion.  So my vote goes for weapons and shield having automatic check for a Monk, while anything else does not.

ikonomov:
In any case, it is a minor thing either way.  But if we wanted to, we can even add armor to the automatic check for a Monk by allowing the type of armor that gives no penalty, so light and medium unique armor would pass the check.

StephenCWills:
The simple and dismissive answer is, players who prefer to utilize unarmed characters in their playstyles should probably just turn off the auto-equip setting. It's not that big of a deal to go into the inventory and equip the weapon you just picked up, so it's very unlikely that such a player will have a significantly worse experience for it.

From a UX perspective, it doesn't seem like a good idea to me to implement arbitrarily selected exceptions to the behavior defined by an explicit setting. I can understand the rationale behind the selection from a design perspective, but the player will have to learn from experience that their auto-equip weapon setting doesn't apply to Monks. Instead, I think it's much better for players to discover the advantage of playing Monk unarmed and make the conscious decision to turn off auto-equip.

ikonomov:
I agree with your argument and I think your point is absolutely valid, but if we accept it, then I think the weapon check that is currently in place should be removed so that we are consistent in our approach.

StephenCWills:
Sorry if this is obtuse, but I'm not sure what you're talking about. What weapon check are you referring to?

ikonomov:
https://github.com/diasurgical/devilutionX/blob/889cc04f95f8ee1289201c8dec288d7f4a2f5ea9/Source/inv.cpp#L1407

ikonomov:
So the point is, I think we should either remove this check for a Monk or add another for shields (and maybe for light/medium unique armors).

StephenCWills:
Huh, I play with auto-equip disabled across the board so I didn't realize there was already an exception. I would agree that it should be removed.

AJenbo:
I would like to have all the auto options merged in to one enum with about 4 options (none, weapon, managed, all). That would clean up the settings and allow for this (managed) logic.

ikonomov:
I've edited out the Monk exception for weapons, but not sure how to do AJenbo's proposal.  It would be nice to have it this way though, it would simplify the settings, which is nice.

StephenCWills (approved):

AJenbo:
Ok, maybe better to write a ticket about it, this pr isn't really going in any particular direction

julealgon:
Just wanted to share my 2c on this one. While I agree that having the hardcoded monk-specific checks is bad, I don't think the proper solution here is to just remove them per se. Instead, what needs to happen is that the auto-equip settings should be "class-scoped", meaning, you should be able to define _different settings for different classes_. Then, we modify the default values for the monk to disallow auto-equipping and remove the hardcodes.

For this to work in a satisfactory manner, you'd also want to expand the auto-equip settings from just the item type, to a proper "filter" where you can specify more granular properties of the items. 

TL;DR, it should be possible at some point to specify the following via config alone:
- "don't auto-equip bows with the warrior"
- "don't auto-equip any weapon with the monk"
- "don't auto-equip daggers with any characters" 
- "don't auto-equip shields with the bard"
- "don't auto-equip another 1-handed weapon on a barbarian already wielding a 1-handed weapon"
- etc

Additionally, I also think an even more advanced system would allow you to override and persist these settings _per specific character_. Settings like this would follow a hierarchy to define the priority of application:
1. if setting exists for specific character, use that. otherwise
2. if setting exists for the character class, use that. otherwise
3. if setting exists globally (what we have today basically), use that. otherwise
4. use default values

PR #4292, opened 2022-04-01
Update bg.po

NikoVP:
Update to latest master


AJenbo:
There aren't any new translations in this PR (well there are a few that are marked as needing review/fuzzy). There is also a conflict and I think it's about time that you maybe reset the branch as it's pretty tangled at this point with 48 old commits in it :) Let me know if you need any help with this.

PR #4298, opened 2022-04-02
Fix #4251 macOS mouse grab issue

### LINKED ISSUE #4251: Cursor confined to 640x480??? Top right. Corrects by switching app then back to DevX
MacOS 12.3 / DevilutionX 1.4 Test Build #6717 / Download

**Describe**
When I launch the game my cursor is stuck in the top right of the screen in what looks like maybe a 640x480 box?

**To Reproduce**
1. Launch DevilutionX
2. Cursor is stuck in top right box/

**Additional context**
When I switch out of the app and then back to DevX it fixes the problem.

https://user-images.githubusercontent.com/82344993/160358054-56df02f5-a312-4bec-8e22-0fc50409a11c.mov

bubio:
I think this is fundamentally an SDL issue.
However, I think this is fine, since the problem can be avoided by making the process the same as the options menu.

Tested on:
MacBook Pro 2017 / macOS 12.3 / Retina
Note PC / Zorin OS / Non-Retina

### DIFF
--- a/Source/utils/display.cpp
+++ b/Source/utils/display.cpp
@@ -243,11 +243,11 @@ bool SpawnWindow(const char *lpWindowName)
 		flags |= SDL_WINDOW_FULLSCREEN;
 	}
 
-	if (*sgOptions.Gameplay.grabInput) {
-		flags |= SDL_WINDOW_INPUT_GRABBED;
-	}
-
 	ghMainWnd = SDL_CreateWindow(lpWindowName, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, windowSize.width, windowSize.height, flags);
+    
+    if (ghMainWnd != nullptr)
+        SDL_SetWindowGrab(ghMainWnd, *sgOptions.Gameplay.grabInput ? SDL_TRUE : SDL_FALSE);
+    
 #endif
 	if (ghMainWnd == nullptr) {
 		ErrSdl();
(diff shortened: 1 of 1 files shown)


Source/utils/display.cpp
```diff
@@ -243,11 +243,11 @@ bool SpawnWindow(const char *lpWindowName)
 		flags |= SDL_WINDOW_FULLSCREEN;
 	}
 
-	if (*sgOptions.Gameplay.grabInput) {
-		flags |= SDL_WINDOW_INPUT_GRABBED;
-	}
-
 	ghMainWnd = SDL_CreateWindow(lpWindowName, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, windowSize.width, windowSize.height, flags);
+
+	if (ghMainWnd != nullptr)
```
glebm, line 248: Please add a comment pointing to the SDL issue: https://github.com/libsdl-org/SDL/issues/962

bubio: Added comment.

Source/utils/display.cpp
```diff
@@ -243,11 +243,13 @@ bool SpawnWindow(const char *lpWindowName)
 		flags |= SDL_WINDOW_FULLSCREEN;
 	}
 
-	if (*sgOptions.Gameplay.grabInput) {
-		flags |= SDL_WINDOW_INPUT_GRABBED;
-	}
-
 	ghMainWnd = SDL_CreateWindow(lpWindowName, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, windowSize.width, windowSize.height, flags);
+
+    // Note: https://github.com/libsdl-org/SDL/issues/962
```
glebm, line 248: Looks like these 2 comment lines are indented with spaces instead of a tab, causing `clang-format` check to fail

bubio: Corrected comment.

glebm (approved):

AJenbo (commented):
nice work figuring this out :+1:

PR #4300, opened 2022-04-02
Use truthy test instead of ifdef for TARGET_OS_IPHONE

### LINKED ISSUE #4299: Full screen switching option menu does not appear in macOS
**Important information**
macOS 12.3 / self build

**Describe**
Full screen switching option menu does not appear in macOS.

issue_4299_1

ephphatha:
This constant is defined as 1 when building for iphone, 0 for MacOS, and undefined on other platforms.

fixes #4299

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -55,7 +55,7 @@ constexpr OptionEntryFlags OnlyIfNoImplicitRenderer = OptionEntryFlags::Invisibl
 constexpr OptionEntryFlags OnlyIfNoImplicitRenderer = OptionEntryFlags::None;
 #endif
 
-#if defined(__ANDROID__) || defined(TARGET_OS_IPHONE)
+#if defined(__ANDROID__) || (defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE == 1)
 constexpr OptionEntryFlags OnlyIfSupportsWindowed = OptionEntryFlags::Invisible;
 #else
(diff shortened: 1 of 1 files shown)


glebm (dismissed):

Source/options.cpp
```diff
@@ -55,7 +55,7 @@ constexpr OptionEntryFlags OnlyIfNoImplicitRenderer = OptionEntryFlags::Invisibl
 constexpr OptionEntryFlags OnlyIfNoImplicitRenderer = OptionEntryFlags::None;
 #endif
 
-#if defined(__ANDROID__) || defined(TARGET_OS_IPHONE)
+#if defined(__ANDROID__) || TARGET_OS_IPHONE
```
glebm, line 58: > Identifiers that are not macros, which are all considered to be the number zero. This allows you to write #if MACRO instead of #ifdef MACRO, if you know that MACRO, when defined, will always have a nonzero value. Function-like macros used without their function call parentheses are also treated as zero. 

This is a nice shortcut but will cause warnings with certain compiler flags.

> In some contexts this shortcut is undesirable. The '-Wundef' option causes GCC to warn whenever it encounters an identifier which is not a macro in an #if.

Perhaps it'd be better to test for `(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE == 1)`?

ephphatha: Good point, I'm not sure if we have -Wundef on (the MSVC equivalent isn't active) but it's trivial to avoid.

glebm (approved):

PR #4304, opened 2022-04-02
Remove use of HoldItem from stores.cpp

AJenbo:


Source/stores.cpp
```diff
@@ -1890,43 +1867,43 @@ void StorytellerIdentifyItem()
...
+			BoyBuyItem(item);
 			break;
 		case STORE_HBUY:
-			HealerBuyItem();
+			HealerBuyItem(item);
 			break;
 		case STORE_SIDENTIFY:
-			StorytellerIdentifyItem();
+			StorytellerIdentifyItem(item);
 			StartStore(STORE_IDSHOW);
```
ephphatha, line 1903: StorytellerIdentifyItem doesn't really need the item as an argument, it's more so StartStore(STORE_IDSHOW) knows which item was identified to display to the player. It could take the selected index or cost.

Source/stores.cpp
```diff
@@ -1890,43 +1867,43 @@ void StorytellerIdentifyItem()
 	} else {
 		myPlayer.InvList[idx]._iIdentified = true;
 	}
-	myPlayer.HoldItem._iIdentified = true;
-	TakePlrsMoney(myPlayer.HoldItem._iIvalue);
+	item._iIdentified = true;
```
ephphatha, line 1870: This is only so that the item can be shown to the user in STORE_IDSHOW. Copying the identified item into StoreItem here would work, StoreItem isn't really relevant before this point.

Source/stores.cpp
```diff
@@ -1699,25 +1683,28 @@ void WitchSellEnter()
...
 /**
  * @brief Recharges an item in the player's inventory or body in the witch.
  */
-void WitchRechargeItem()
+void WitchRechargeItem(int price)
 {
-	auto &myPlayer = Players[MyPlayerId];
-
-	TakePlrsMoney(myPlayer.HoldItem._iIvalue);
+	TakePlrsMoney(price);
```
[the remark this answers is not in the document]
AJenbo: I get what you are saying, but as things are it's not super clear to me when the items value is manipulated to hold a temporary store exchange value versus the actual items value. I feel that I would be running a risk of introducing bugs by making assumptions regarding `StoreItem == storehold[idx]` at this point and would rather do so when addressing the whole `stextlhold`, `stextsel`, `stextvhold`, `stextsval`, `storehold` mess. Also `BoyBuyItem()` does not operate with `idx` and this way they get to have a similar interface.

The idea behind having Repair/Recharge/Identify take the value was that in the future the shop prices would be separate from `item._iIvalue` since manipulating the items makes things fragile and confusing. But I can make them take `Item &item` as the input instead if that makes a difference to you.

Source/stores.cpp
AJenbo: Sure, but doing so would require rewriting most of this function. Can we save that for an actual rewrite of the stores?

Source/stores.cpp
AJenbo: If I only give it the price i have to set it as identified directly here, or in `StartStorytellerIdentifyShow()` which seams like the wrong place to do so. Keeping it like this feels like it keeps it clear what to refactor with out things getting more misleading then they where.

Source/stores.cpp
```diff
@@ -1387,7 +1383,7 @@ void SmithBuyItem()
 		}
 		smithitem[idx]._itype = ItemType::None;
 	}
-	CalcPlrInv(myPlayer, true);
+	CalcPlrInv(Players[MyPlayerId], true);
```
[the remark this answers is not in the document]
AJenbo: Done :+1:

AJenbo:
Maybe I'm to tired to deal with the logic in this file. But generally I would prefer to not mess around in it to much as part of removing `HoldItem` from it. You are welcome to add a commit on top of it if you like and maybe that will make more sens to me that way. But there is also this one that needs to be rebased on top of this: https://github.com/diasurgical/devilutionX/pull/1524 + a more or less full rewrite to go on top (local branch not pushed).

Source/stores.cpp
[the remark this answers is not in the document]
ephphatha: > I get what you are saying, but as things are it's not super clear to me when the items value is manipulated to hold a temporary store exchange value versus the actual items value.

The _ivalue/_iIvalue variables in storehold are always the cost of action/vendor offer (either 1/4 the item's known value or the cost for repairing/recharging/identifying the item). Your point about avoiding introducing bugs is good though, I brought it up here because of the added parameter but it's easy enough to change later.

> The idea behind having Repair/Recharge/Identify take the value was that in the future the shop prices would be separate from `item._iIvalue` since manipulating the items makes things fragile and confusing. But I can make them take `Item &item` as the input instead if that makes a difference to you.

I would definitely not want this to take an item reference as the sole param 😂 . What you've got now is good enough, functions the same (at least from my limited testing) and solves the original purpose.

ephphatha (approved):

PR #4306, opened 2022-04-03
Avoid using HoldItem in AddStealPotions

AJenbo:


Source/missiles.cpp
```diff
@@ -1420,10 +1420,7 @@ void AddStealPotions(Missile &missile, const AddMissileParameter & /*parameter*/
 					}
 				}
 				if (ii != -1) {
-					InitializeItem(player.HoldItem, ii);
-					GenerateNewSeed(player.HoldItem);
-					player.HoldItem._iStatFlag = true;
-					player.SpdList[si] = player.HoldItem;
+					InitializeItem(player.SpdList[si], ii);
```
ephphatha, line 1423:
```suggestion
					InitializeItem(player.SpdList[si], ii);
					player.SpdList[si]._iStatFlag = true;
```
`InitialiseItem()` uses the default constructor for `Item` which sets `_iStatFlag` to false.

PR #4307, opened 2022-04-03
Avoid destroying held item from save game

AJenbo:


StephenCWills:
I can't actually save the game when I'm holding an item. Can you provide steps to reproduce?

AJenbo:
Probably not, it might actually just be redundant to save the HoldItem, but maybe it wasn't at some point :shrug:

StephenCWills:
FYI, I was trying to figure out whether you'd need to add `if (player.HoldItem._itype == ItemType::None)`, but it seems that the code is forcing `CURSOR_HAND` anyway.

https://github.com/diasurgical/devilutionX/blob/0c6f130e5d8237596d3e21f15dcae83ef24e0623/Source/loadsave.cpp#L2055

obligaron:
> I can't actually save the game when I'm holding an item. Can you provide steps to reproduce?

This can happen in multiplayer, cause the autosave mechanic kicks in if you close the window.
But sadly `PackPlayer` doesn't save the hold item. So there are not much possibilities.

PR #4308, opened 2022-04-03
Avoid using HeldItem in CreatePlrItems

AJenbo:


Source/items.cpp
```diff
@@ -2843,9 +2843,12 @@ void CreatePlrItems(int playerId)
...
 
-		InitializeItem(player.HoldItem, IDI_WARRCLUB);
-		GenerateNewSeed(player.HoldItem);
-		AutoPlaceItemInInventory(player, player.HoldItem, true);
+		InitializeItem(player.InvList[player._pNumInv], IDI_WARRCLUB);
+		GenerateNewSeed(player.InvList[player._pNumInv]);
+		player._pNumInv++;
+		player.InvGrid[0] = -player._pNumInv;
+		player.InvGrid[10] = -player._pNumInv;
+		player.InvGrid[20] = player._pNumInv;
```
ephphatha, line 2851: `AddItemToInventoryGrid()` is in the anonymous namespace in inv.cpp, perhaps that should be exposed in the header? Or use `AutoPlaceItemInInventorySlot(player, 0, club, true)` with a local item variable.

AJenbo: I linked this one cause it didn't create a temporary item, and creating variables in a switch is always a pain. But here you go :)

`AddItemToInvGrid()` requires item size so that goes down a path I don't think makes things look any nicer.

ephphatha (approved):

PR #4310, opened 2022-04-03
Avoid using HoldItem in OperateShrineEldritch

AJenbo:

PR #4312, opened 2022-04-03
Use HoldItem for determining if an item is held

AJenbo:
This cleans up a bit of code and solves a few edge cases where an item
could be lost, the game be unresponsive, or miss fire an event during
lag cause of the cursor not reflecting the currently held item.

Depends on:
- https://github.com/diasurgical/devilutionX/pull/4304
- https://github.com/diasurgical/devilutionX/pull/4306
- https://github.com/diasurgical/devilutionX/pull/4308
- https://github.com/diasurgical/devilutionX/pull/4309
- https://github.com/diasurgical/devilutionX/pull/4310

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -582,7 +582,7 @@ void CheckInvPaste(int pnum, Point cursorPosition)
 	CalcPlrInv(player, true);
 	if (pnum == MyPlayerId) {
 		if (cn == CURSOR_HAND && !IsHardwareCursor())
-			SetCursorPos(MousePosition + Displacement(cursSize / 2));
+			SetCursorPos(MousePosition + Displacement { itemSize * INV_SLOT_HALF_SIZE_PX });
 		NewCursor(cn);
 	}
 }
@@ -856,6 +856,7 @@ void CheckInvCut(int pnum, Point cursorPosition, bool automaticMove, bool dropIt
 				NewCursor(holdItem._iCurs + CURSOR_FIRSTITEM);
 				if (!IsHardwareCursor() && !dropItem) {
 					// For a hardware cursor, we set the "hot point" to the center of the item instead.
+					Size cursSize = GetInvItemSize(holdItem._iCurs + CURSOR_FIRSTITEM);
 					SetCursorPos(cursorPosition - Displacement(cursSize / 2));
 				}
 			}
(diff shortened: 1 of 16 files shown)


ephphatha:
Would you mind rebasing this (and marking it ready for review)? I've been refactoring uses of `NewCursor()` so that function doesn't modify `MyPlayer->HoldItem` which ends up overlapping with a lot of the checks this PR touches. Don't want to duplicate effort if it can be avoided :D

AJenbo:
I'm considering changing moving HoldItem out of the Player struct.
```cpp
Cursor {
    Point position; // cursPosition
    Point size; // cursSize
    cursor_id pointer; // pcurs
private:
    Item item; // MyPlayer->HoldItem;
}

/** Set item and item size, */
void Cursor::setItem(const Item &item);
/** Set item and item size, */
void Cursor::clearItem();
bool Cursor::hasItem();
```

Possibly also have a reference to the current graphics to avoid having to look that each time it renderes.
I think it could simplify a lot of the code.

Let me know if you prefer that done as part of this or later :)

Source/inv.cpp
```diff
@@ -581,8 +581,10 @@ void CheckInvPaste(int pnum, Point cursorPosition)
 	}
 	CalcPlrInv(player, true);
 	if (pnum == MyPlayerId) {
-		if (cn == CURSOR_HAND && !IsHardwareCursor())
+		if (cn == CURSOR_HAND && !IsHardwareCursor()) {
+			Size cursSize = GetInvItemSize(MyPlayer->HoldItem._iCurs + CURSOR_FIRSTITEM);
 			SetCursorPos(MousePosition + Displacement(cursSize / 2));
```
ephphatha, line 586: This relies on HoldItem staying valid after being emptied. Using `itemSize * INV_SLOT_HALF_SIZE_PX` here has the same effect (or create a variable like hotPixelOffset in stash.cpp).

ephphatha:
> Let me know if you prefer that done as part of this or later :)

Honestly that seems like a decent approach. I was a bit torn about whether holdItem should stay part of the player struct as it can be useful to think of it as another item space per player (especially for potential couch co-op :D). I think that would be more straightforward and better support multiple cursors anyway. I would prefer it as a separate PR as this one is good to go apart from that comment/nit.

ephphatha (approved):

PR #4316, opened 2022-04-03
set statflag after recreating potion

ephphatha:
I was lazy with the review and didn't flag both cases.


AJenbo:
I was to tired to notice :sweat_smile:

PR #4318, opened 2022-04-03
Ensure Reflect + Thorn don't kill a monster twice

### LINKED ISSUE #3869: Na-Krul loses twice
It seems to be a newly introduced problem that I was unfortunately not prepared for. That's why I haven't been able to determine since when it happened yet. Anyway, as you can see, Na-Krul lost double the number of items when he was killed. It was very generous of him, but certainly not in the spirit of the inventor. The causative PR has yet to be determined.
items
Savegame:
[single_1.hsv.zip](https://github.com/diasurgical/devilutionX/files/7794817/single_1.hsv.zip)

EDIT:
These items seem to be duped for an unknown reason. They duped themselves, or it was the golem who killed Na-Krul.

obligaron:
Fixes #3869

Notes
- When a monster hits (`MonsterAttackPlayer`) first reflect is calculated. Reflect can kill a monster (`CheckReflect` -> `M_StartKill` -> `StartMonsterDeath`).
- After this thorn (`Attacker takes x Damage`) is calculated. Thorn can also kill a monster (`M_StartKill` -> `StartMonsterDeath`).
- If reflect already killed the monster (and set it's hp to less then 1). Then thorn also adds damage, sees the monster has hp less then 1 and also triggers the kill. When this happens we have the double kill.
- This is not specific to Na-Krul. It can happen with any monster.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1414,7 +1414,8 @@ void MonsterAttackPlayer(int i, int pnum, int hit, int minDam, int maxDam)
 			CheckReflect(i, pnum, dam);
 		ApplyPlrDamage(pnum, 0, 0, dam);
 	}
-	if ((player._pIFlags & ISPL_THORNS) != 0) {
+	// Reflect can also kill a monster, so make sure the monster is still alive
+	if ((player._pIFlags & ISPL_THORNS) != 0 && monster._mmode != MonsterMode::Death) {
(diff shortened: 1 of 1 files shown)


qndel (approved):
Nice :)

AJenbo:
nice work, bugs like thease are stressful to me :)

PR #4321, opened 2022-04-03
Mouse/Gamepad fixes

### LINKED ISSUE #4242: [gamepad] Inventory cursor missing
Initially when utilizing the inventory the cursor can be missing or dark and stuck. This can be solved by moving the cursor. Probably replicating this may require the game to run in windowed mode and the cursor to start offscreen.

It also appears that the cursor does not snap to the first slot when opening the inventory, this may be the cause of the issue.

(this may or may not be related to the stash-gamepad branch)

glebm:
Introduces a `ControlDevice` global which is distinct from `ControlMode` in that it is set to `Gamepad` even when simulating a mouse.

This allows us to avoid a number of edge cases related to mode changes.

Fixes #4242

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -4,6 +4,10 @@
 #include <cstdint>
 #include <list>
 
+#ifdef USE_SDL1
+#include "utils/sdl2_to_1_2_backports.h"
+#endif
+
 #include "automap.h"
 #include "control.h"
 #include "controls/controller_motion.h"
@@ -30,12 +34,14 @@
 #include "stores.h"
 #include "towners.h"
 #include "trigs.h"
+#include "utils/log.hpp"
 
 #define SPLICONLENGTH 56
 
 namespace devilution {
 
 ControlTypes ControlMode = ControlTypes::None;
+ControlTypes ControlDevice = ControlTypes::None;
 int pcurstrig = -1;
 Missile *pcursmissile = nullptr;
 quest_id pcursquest = Q_INVALID;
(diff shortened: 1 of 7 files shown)


glebm:
One thing I'm not sure about is whether we should call `NextMouseButtonClickEventIsSimulated` in touch.cpp

Source/controls/plrctrls.cpp
```diff
@@ -1491,12 +1497,51 @@ bool ContinueSimulatingMouse(const SDL_Event &event, const ControllerButtonEvent
 	return false;
 }
 
+bool IsNextMouseButtonClickEventIsSimulated = false;
```
AJenbo, line 1500:
```suggestion
bool IsNextMouseButtonClickEventSimulated = false;

```

AJenbo:
> One thing I'm not sure about is whether we should call `NextMouseButtonClickEventIsSimulated` in touch.cpp

I don't think so, it's not emulating the mouse in the same way i believe. I can test things out a bit later, is there something specific i should look out for?

glebm:
@AJenbo Look out for issues around switching between mouse and touch. I suggest we merge it and fix `touch.cpp` later if needed, so that it's tested a bit more widely on master.

AJenbo:
8 hours ago I sat down to test this, and then i started getting notifications ...

AJenbo:
Ok, i did notice one issue when testing. When using a mouse on Android picking up and placing items in the inventory will have the mouse jump to the top and bottom of the object instead of staying in the center. But this turns out to also be an issue on master, so :shrug:

The same issue does not happen on Linux both with Hardware and Software Mouse.

AJenbo (approved):

glebm:
I'm guessing that's probably an SDL issue

AJenbo:
Probably, I just replicated it on 1.3.0 from the play store, so in any case it's non blocking :)

PR #4325, opened 2022-04-03
Fix some more warnings

glebm:


Source/DiabloUI/settingsmenu.cpp
```diff
@@ -142,7 +142,8 @@ bool ChangeOptionValue(OptionEntryBase *pOption, size_t listIndex)
 
 void ItemSelected(int value)
 {
-	auto &vecItem = vecDialogItems[value];
+	const auto index = static_cast<size_t>(value);
```
ephphatha, line 145: Was wondering why not change the function signature, but then I realised every menu is driven by function pointers 😂 .

Another thing I noticed is the inconsistent use of `std::size_t` vs `size_t` in the current code, is the typedef in the `std` namespace is preferred for c++? I've been trying to use that where I remember.

Source/options.h
```diff
@@ -555,13 +557,13 @@ struct KeymapperOptions : OptionCategoryBase {
...
 	private:
-		Action(string_view key, string_view name, string_view description, int defaultKey, std::function<void()> actionPressed, std::function<void()> actionReleased, std::function<bool()> enable, int index);
+		Action(string_view key, string_view name, string_view description, int defaultKey, std::function<void()> actionPressed, std::function<void()> actionReleased, std::function<bool()> enable, size_t index);
 		int defaultKey;
 		std::function<void()> actionPressed;
 		std::function<void()> actionReleased;
 		std::function<bool()> enable;
 		int boundKey = DVL_VK_INVALID;
-		int dynamicIndex;
+		size_t dynamicIndex;
```
ephphatha, line 566: This is not really an index, it's more a numeric discriminator for options with the same name. `unsigned` would be suitable here.

Source/options.h
```diff
@@ -573,7 +575,10 @@ struct KeymapperOptions : OptionCategoryBase {
 
 	void AddAction(
 	    string_view key, string_view name, string_view description, int defaultKey,
-	    std::function<void()> actionPressed, std::function<void()> actionReleased = nullptr, std::function<bool()> enable = nullptr, int index = -1);
+	    std::function<void()> actionPressed,
+	    std::function<void()> actionReleased = nullptr,
+	    std::function<bool()> enable = nullptr,
+	    size_t index = std::numeric_limits<size_t>::max());
```
ephphatha, line 581: This is a non-zero value every time it's relevant, could use 0 as the default/unused marker.

Source/DiabloUI/settingsmenu.cpp
glebm: It was the case that `std::size_t` was preferred but in C++20 `size_t` is guaranteed to be defined in `<cstdef>` (and `uint_8` in `<cstdint>`) without the `std` prefix.

This may change yet again with C++23 stdlib modules. 😅

glebm:
@ephphatha Thanks for the review, all comments addressed!

Issue #4181, opened 2022-03-19
[PROPOSAL] Come back on main menu on "Quit Game" click, instead of exit


M0Rf30:
**Describe**
Actually when you click on "Quit game", during the game, it completely exit from the execution.
Wouldn't be better to come back to main menu and avoid launching the executable again?
In this way you will quit devilutionx only with "Exit Diablo" button, to be more clear

qndel:
That's how "new game" button works xD

M0Rf30:
So we have two button with the same logic? Anyway, I honestly think about how other games approach this behaviour. Basically they avoid to "break" the experience. Many of them also repeatly ask for confirmation on exit.

They want the player to stay in.

qndel:
From what I understand you want to convert the function of "exit game" button to what "new game" is currently doing?

M0Rf30:
I would like to come back to "single player, multiplayer......" menu. so if new game will bring you to character screen. exit game will bring you to the main menu. where you can choose about options, game mode .... or completely quit diablo. so two click to completely exit, instead of one

AJenbo:
Since exiting the application is an issue on mobiles I think this is a good over all change.
We can get rid of the "new game" option. I don't think we are loosing much by having to press two extra buttons to quit the game.

qndel:
> Since exiting the application is an issue on mobiles I think this is a good over all change. We can get rid of the "new game" option. I don't think we are loosing much by having to press two extra buttons to quit the game.

I can counter that with "I don't think we are gaining much - could've achieved the same with two extra button presses earlier" ;)

M0Rf30:
> > Since exiting the application is an issue on mobiles I think this is a good over all change. We can get rid of the "new game" option. I don't think we are loosing much by having to press two extra buttons to quit the game.
> 
> I can counter that with "I don't think we are gaining much - could've achieved the same with two extra button presses earlier" ;)

Sorry but I don't get it. could you elaborate on this? What do you mean?

qndel:
New game -> esc = 
![image](https://user-images.githubusercontent.com/14297035/159249102-61011a8f-70c0-4b2c-b104-fef61c57243b.png)
this menu

So having 2 buttons performing very similar tasks is pointless, personally I like being able to quit the game fast.

AJenbo:
[...]

PR #4331, opened 2022-04-03
Make 'quit game' return to the main menu when NOEXIT is set

AJenbo:
This partially implements #4181 but only for systems where NOEXIT is set. This should avoid at least some confusion for users that are unsure how to get back to the main menu from the game menu.

### DIFF
--- a/Source/gamemenu.cpp
+++ b/Source/gamemenu.cpp
@@ -31,39 +31,35 @@ void GamemenuSpeed(bool bActivate);
 /** Contains the game menu items of the single player menu. */
 TMenuItem sgSingleMenu[] = {
 	// clang-format off
-    // dwFlags,      pszStr,              fnMenu
-	{ GMENU_ENABLED, N_("Save Game"),     &gamemenu_save_game  },
-	{ GMENU_ENABLED, N_("Options"),       &GamemenuOptions    },
-	{ GMENU_ENABLED, N_("New Game"),      &GamemenuNewGame   },
-	{ GMENU_ENABLED, N_("Load Game"),     &gamemenu_load_game  },
-#ifndef NOEXIT
-	{ GMENU_ENABLED, N_("Quit Game"),     &gamemenu_quit_game  },
-#endif
-	{ GMENU_ENABLED, nullptr,              nullptr             }
+	// dwFlags,      pszStr,          fnMenu
+	{ GMENU_ENABLED, N_("Save Game"), &gamemenu_save_game },
+	{ GMENU_ENABLED, N_("Options"),   &GamemenuOptions    },
+	{ GMENU_ENABLED, N_("New Game"),  &GamemenuNewGame    },
+	{ GMENU_ENABLED, N_("Load Game"), &gamemenu_load_game },
+	{ GMENU_ENABLED, N_("Quit Game"), &gamemenu_quit_game },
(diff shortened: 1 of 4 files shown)


Source/gamemenu.cpp
```diff
@@ -31,39 +31,35 @@ void GamemenuSpeed(bool bActivate);
...
-    // dwFlags,      pszStr,              fnMenu
-	{ GMENU_ENABLED, N_("Save Game"),     &gamemenu_save_game  },
-	{ GMENU_ENABLED, N_("Options"),       &GamemenuOptions    },
-	{ GMENU_ENABLED, N_("New Game"),      &GamemenuNewGame   },
-	{ GMENU_ENABLED, N_("Load Game"),     &gamemenu_load_game  },
-#ifndef NOEXIT
-	{ GMENU_ENABLED, N_("Quit Game"),     &gamemenu_quit_game  },
-#endif
-	{ GMENU_ENABLED, nullptr,              nullptr             }
+    // dwFlags,      pszStr,          fnMenu
```
qndel, line 34: if you decided to align this stuff, please align the header too ;)

Source/gamemenu.cpp
```diff
@@ -31,39 +31,35 @@ void GamemenuSpeed(bool bActivate);
...
+	{ GMENU_ENABLED, N_("New Game"),  &GamemenuNewGame    },
+	{ GMENU_ENABLED, N_("Load Game"), &gamemenu_load_game },
+	{ GMENU_ENABLED, N_("Quit Game"), &gamemenu_quit_game },
+	{ GMENU_ENABLED, nullptr,         nullptr             }
 	// clang-format on
 };
 /** Contains the game menu items of the multi player menu. */
 TMenuItem sgMultiMenu[] = {
 	// clang-format off
     // dwFlags,      pszStr,                fnMenu
```
qndel, line 46: same here

Source/gamemenu.cpp
```diff
@@ -31,39 +31,35 @@ void GamemenuSpeed(bool bActivate);
...
-	{ GMENU_ENABLED, N_("Quit Game"),       &gamemenu_quit_game    },
-#endif
-	{ GMENU_ENABLED, nullptr,                nullptr               },
+	{ GMENU_ENABLED, N_("Quit Game"),       &gamemenu_quit_game  },
+	{ GMENU_ENABLED, nullptr,               nullptr              },
 	// clang-format on
 };
 TMenuItem sgOptionsMenu[] = {
 	// clang-format off
     // dwFlags,                     pszStr,              fnMenu
```
qndel, line 56: same here

StephenCWills:
I like this idea. I found that selecting Quit Game with a touchscreen causes a click event on the main menu as well, forcing me into the Support menu option every time.

Source/gamemenu.cpp
AJenbo: I actually did, but didn't spot that someone ad used spaces instead of tab for the indentation.

Source/gamemenu.cpp
AJenbo: Thanks I know how to search

qndel (approved):

AJenbo:
> I like this idea. I found that selecting Quit Game with a touchscreen causes a click event on the main menu as well, forcing me into the Support menu option every time.

Not sure what is happening there, but lets look in to it later :)

PR #4332, opened 2022-04-03
[Gamepad] Allow casting Town Portal, Teleport, and Guardian scrolls from inventory

### LINKED ISSUE #4273: [gamepad] Enable use of targeted spells from the inventory
A few spells can not be triggered from the inventory, there isn't any specific reason for this in most cases as they work without an  active target when using a gamepad.

https://github.com/diasurgical/devilutionX/pull/4266/files#diff-be3ecbe1ead4062ccd317dfeb0428b38adf3d54377059e602229c75bfc052da8R1813

StephenCWills:
This resolves #4273

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1768,7 +1768,7 @@ bool SpellHasActorTarget()
 	return pcursplr != -1 || pcursmonst != -1;
 }
 
-void UpdateSpellTarget()
+void UpdateSpellTarget(spell_id spell)
 {
 	if (SpellHasActorTarget())
 		return;
@@ -1778,7 +1778,7 @@ void UpdateSpellTarget()
 
 	auto &myPlayer = Players[MyPlayerId];
 
-	int range = myPlayer._pRSpell == SPL_TELEPORT ? 4 : 1;
+	int range = spell == SPL_TELEPORT ? 4 : 1;
 
 	cursPosition = myPlayer.position.future + Displacement(myPlayer._pdir) * range;
 }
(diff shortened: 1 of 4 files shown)


Source/spells.cpp
```diff
@@ -113,6 +113,16 @@ bool IsWallSpell(spell_id spl)
...
 }
 
+bool TargetsMonster(spell_id id)
+{
+	return id == SPL_FIREBALL
+	    || id == SPL_FIREWALL
+	    || id == SPL_FLAME
+	    || id == SPL_LIGHTNING
+	    || id == SPL_STONE
+	    || id == SPL_WAVE;
```
AJenbo, line 123: This is what took the longest to review, it's so confusing that fire `SPL_FIREBALL` targets monsters, but `SPL_FIREBOLT` does not :D

AJenbo (approved):
This will make support happy :smiling_face_with_tear:

Source/controls/plrctrls.cpp
```diff
@@ -1861,39 +1861,52 @@ void PerformSpellAction()
...
-	if (pcursstashitem == uint16_t(-1))
+	if (pcursstashitem == uint16_t(-1)) {
 		return;
+	}
 
 	const Item &item = Stash.stashList[pcursstashitem];
-	if (item.IsScroll() && spelldata[item._iSpell].sTargeted) {
-		return;
+	if (item.IsScroll()) {
+		if (TargetsMonster(item._iSpell)) {
```
AJenbo, line 1904: Technically this won't happen as they are not allowed in town, but i like the consistency :+1:

Source/spells.cpp
StephenCWills: Hmm, that's a good point. I hope this doesn't get used for anything other than scrolls...

PR #4336, opened 2022-04-04
increase max stash pages from 50 to 100

qndel:
I feel if I don't do it, nobody will 😅  50 is a weird number to have as a max especially considering we can handle more than that. 100 is a nice round number :)

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -28,7 +28,7 @@ int WithdrawGoldValue;
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


Trihedraf (approved):

obligaron:
I'm against this change. The number 100 is arbitrary in context of diablo!
We need a number with context... we should use 666. 😈 
With 666 pages we have max. 66600 items and we can store 65535 items so... okay forget it. 😜😁

PR #4345, opened 2022-04-05
Prefer when using VSync with no upscaling on Windows

### LINKED ISSUE #3806: Display freezing on Windows when turning Upscale off with VSync on
**Important information**
Windows 10 / DevilutionX master / compiled through VS2019

**Describe**
With both VSync and Upscale set to On, changing Upscale to Off causes the display to freeze.

**To Reproduce**
1. Go to Settings > Graphics
2. Ensure Vertical Sync and Upscale are both on
3. Turn Upscale off
4. The display will freeze on the last displayed frame

The game still responds to keyboard input and can be exited/have other settings changed. Turning Upscale back on will cause the display to update again. However, once the D3D11 renderer has been selected even once by turning both VSync and Upscale on, I can no longer get the display to update when Upscale is off.

**Expected behavior**
Turning Upscale off does not freeze the display.

**Additional context**
Plagiarized almost verbatim from #3706 because the symptoms are really similar, but that's almost 100% coincidence. I didn't want to reopen the other issue since the cause is fundamentally different. I'm sorry if this seems really familiar.

Introduced by #3775

AJenbo:
Fixes #3806

@obligaron @StephenCWills Could I get one of you to test this since I don't have Windows :/

Regarding always using OpenGL on Windows, that might not be the best choice since we have the option of using higher quality upscaling when DirectX is being used.

### DIFF
--- a/Source/utils/display.cpp
+++ b/Source/utils/display.cpp
@@ -323,10 +323,10 @@ void ReinitializeRenderer()
 
 #ifdef _WIN32
 		// On Windows 11 the directx9 VSYNC timer doesn't get recreated properly, see https://github.com/libsdl-org/SDL/issues/5099
-		// Attempt to use the directx11 driver instead if we have vsync active.
+		// Attempt to use the opengl driver instead if we have vsync active.
 		const char *const renderHint = SDL_GetHint(SDL_HINT_RENDER_DRIVER);
-		if ((rendererFlags & SDL_RENDERER_PRESENTVSYNC) != 0 && SDL_SetHint(SDL_HINT_RENDER_DRIVER, "direct3d11") != SDL_TRUE) {
-			Log("Error when trying to set hint for direct3d11, using default render driver");
+		if ((rendererFlags & SDL_RENDERER_PRESENTVSYNC) != 0 && SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl") != SDL_TRUE) {
+			Log("Error when trying to set hint for opengl, using default render driver");
 		}
 #endif
 


obligaron:
I tried the change.
The positive first, I can now change VSync and Upscale freely.

But I encountered two issues while testing
1. It looks like two windows are created on start-up. Initial one is destroyed when calling `SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl")`. After that the new window is created. (Checked that both have different window handles).
2. When first running the app, the second window creation seem to fail. But after debugging once I couldn't reproduce the issue anymore.

As alternative fix: Perhaps we could avoid setting VSync when recreating the renderer? VSync would be enabled after the next restart. On windows machines normally we should be fast enough that this doesn't make a big difference. 🤔

StephenCWills:
It seems this is what's causing the second window to be created.
https://github.com/libsdl-org/SDL/blob/5613a5603245f600b36fe327c11c67f0e1ef2cc3/src/render/opengl/SDL_render_gl.c#L1742-L1754

If VSync is on at startup, the window is recreated when the first renderer is created.
If VSync is off at startup, the window is recreated when VSync is turned on from the Settings menu.

FYI, despite the fact that we added code to save the previous render hint and attempt to restore it, we don't actually ever change the renderer back to Direct3D unless the game is restarted.
https://github.com/libsdl-org/SDL/issues/5120

It is actually possible to switch back to the default renderer by using a driver name that doesn't exist, such as an empty string.

```diff
@@ -324,8 +324,11 @@ void ReinitializeRenderer()
 #ifdef _WIN32
 		// On Windows 11 the directx9 VSYNC timer doesn't get recreated properly, see https://github.com/libsdl-org/SDL/issues/5099
 		// Attempt to use the opengl driver instead if we have vsync active.
-		const char *const renderHint = SDL_GetHint(SDL_HINT_RENDER_DRIVER);
+		const char *renderHint = SDL_GetHint(SDL_HINT_RENDER_DRIVER);
+		if (renderHint == nullptr) {
+			renderHint = "";
+		}
 		if ((rendererFlags & SDL_RENDERER_PRESENTVSYNC) != 0 && SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl") != SDL_TRUE) {
 			Log("Error when trying to set hint for opengl, using default render driver");
 		}
 #endif
 ```

Based on my testing, switching back and forth between renderers does not force the window to be recreated so long as that `SDL_WINDOW_OPENGL` flag is defined. Therefore, we should be able to add that flag on window creation regardless of what renderer we end up using. However, I'm not sure if that has any adverse effects.

```diff
@@ -242,5 +242,8 @@ bool SpawnWindow(const char *lpWindowName)
 	} else if (*sgOptions.Graphics.fullscreen) {
 		flags |= SDL_WINDOW_FULLSCREEN;
 	}
+#ifdef _WIN32
+	flags |= SDL_WINDOW_OPENGL;
+#endif
 
 	ghMainWnd = SDL_CreateWindow(lpWindowName, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, windowSize.width, windowSize.height, flags);
```

StephenCWills:
It seems this is what's causing the second window to be created.
https://github.com/libsdl-org/SDL/blob/5613a5603245f600b36fe327c11c67f0e1ef2cc3/src/render/opengl/SDL_render_gl.c#L1742-L1754

If VSync is on at startup, the window is recreated when the first renderer is created.
If VSync is off at startup, the window is recreated when VSync is turned on from the Settings menu.

FYI, despite the fact that we added code to save the previous render hint and attempt to restore it, we don't actually ever change the renderer back to Direct3D unless the game is restarted.
https://github.com/libsdl-org/SDL/issues/5120

It is actually possible to switch back to the default renderer by using a driver name that doesn't exist, such as an empty string.

```diff
@@ -324,8 +324,11 @@ void ReinitializeRenderer()
 #ifdef _WIN32
 		// On Windows 11 the directx9 VSYNC timer doesn't get recreated properly, see https://github.com/libsdl-org/SDL/issues/5099
 		// Attempt to use the opengl driver instead if we have vsync active.
-		const char *const renderHint = SDL_GetHint(SDL_HINT_RENDER_DRIVER);
+		const char *renderHint = SDL_GetHint(SDL_HINT_RENDER_DRIVER);
+		if (renderHint == nullptr) {
+			renderHint = "";
+		}
 		if ((rendererFlags & SDL_RENDERER_PRESENTVSYNC) != 0 && SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl") != SDL_TRUE) {
 			Log("Error when trying to set hint for opengl, using default render driver");
 		}
 #endif
 ```

Based on my testing, switching back and forth between renderers does not force the window to be recreated so long as that `SDL_WINDOW_OPENGL` flag is defined. Therefore, we should be able to add that flag on window creation regardless of what renderer we end up using. However, I'm not sure if that has any adverse effects.

```diff
@@ -242,5 +242,8 @@ bool SpawnWindow(const char *lpWindowName)
 	} else if (*sgOptions.Graphics.fullscreen) {
 		flags |= SDL_WINDOW_FULLSCREEN;
 	}
+#ifdef _WIN32
+	flags |= SDL_WINDOW_OPENGL;
+#endif
 
 	ghMainWnd = SDL_CreateWindow(lpWindowName, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, windowSize.width, windowSize.height, flags);
```

StephenCWills:
I dug into the `SDL_WINDOW_OPENGL` flag a bit and quickly found this code.
https://github.com/libsdl-org/SDL/blob/5613a5603245f600b36fe327c11c67f0e1ef2cc3/src/video/SDL_video.c#L1513-L1523

Also, if you look just a few lines down from there, you'll see that the Vulkan renderer is not compatible with an OpenGL window. So if we add that flag to window creation, I'd expect that the main issues we might encounter would be...

* Cards or drivers that do not support OpenGL
* Failure to load the OpenGL library
* Users who set the `SDL_HINT_RENDER_DRIVER` environment variable so they can force Vulkan instead

TBH I'm not very well versed in graphics cards and the features they commonly support. However, these all seem like fairly low risks to me. My general impression is that OpenGL is pretty much universally supported and that overriding SDL environment variables is really uncommon.

glebm:
> My general impression is that OpenGL is pretty much universally supported

A lot of mobile GPUs only support GLES, not sure if it's a concern on Windows though.

PR #4351, opened 2022-04-06
Fix a typo in Hellfire book text

glebm:
Fixes a typo in "Journal: The tirade"

A tab instead of space meant that this was rendered as "arefull"


AJenbo (approved):
Thanks for also syncing the translations

PR #4353, opened 2022-04-07
Update zhCN translation

tytannial:
More translation. Fix fuzzy.


AJenbo:

tytannial:
> There's going to be and update to DevilutionX soon so now would be a good time to update the translation if you have time 

There have been too many things to do at work recently, so I don't have time to do localization work at the moment. :(
I hope to be free from busyness next month. XD

AJenbo:
ok thanks for letting me know

Issue #4357, opened 2022-04-07
Cursor changes to Windows default


ikonomov:
The cursor changes to Windows default when moving along the edges of the screen even when at Full Screen.  Enabling Grab Input in the .ini does not fix the problem.

qndel:
works fine for me, can u record and post your config and specs?

ikonomov:
https://user-images.githubusercontent.com/5624187/162244702-49480eda-cbe1-4cf9-a3f0-f57c5b8dd2f9.mp4

ikonomov:
this is with the default settings, deleted previous .ini and this is straight out of the build

glebm:
That bright line at the top of the screen is suspicious. Is full screen really full screen?

ikonomov:
I don't see a bright line on my screen.  It is really at full screen.  The problem it turned out is caused by windows scaling.  I normally have Windows scaling at 125%, when I set it to 100% this problem disappears.  Another way to resolve it I just found out is to set this inside the compatibility file properties for devilutionx.exe

![image](https://user-images.githubusercontent.com/5624187/162248531-90994190-5f01-4799-8d83-12028af0d3a6.png)

qndel:
Works fine with 125% scaling and these settings for me (I haven't changed anything, these were default)
![image](https://user-images.githubusercontent.com/14297035/162249193-73cdb1b6-3fc0-41c5-a629-6656911f0252.png)

Anyway not a devilutionx issue

ikonomov:
It turns out, the problem is caused by using a compiled .exe from GitHub's test builds.  If I use a compiled .exe by Visual Studio I don't have this problem.

AJenbo:
> It turns out, the problem is caused by using a compiled .exe from GitHub's test builds. If I use a compiled .exe by Visual Studio I don't have this problem.

Which of the text builds are you refering to, there are 3, and one of them is build with Visual Studio.

If this is an issue with mingw builds then it's an issue since it would affect release. But it sounds more likely that this is related to the version of SDL being used.

qndel:
@AJenbo 
If I build master with visual studio, I don't have this issue.
if I run https://github.com/diasurgical/devilutionX/actions/runs/2108354471 I can see windows cursor

ikonomov:
I might be wrong but pretty sure it's not the SDL version, but the compiler itself.

AJenbo:
[...]

PR #4361, opened 2022-04-08
Fix application manifest in MinGW builds

### LINKED ISSUE #4358: Hardware cursor for items changes image size
Possibly related to https://github.com/diasurgical/devilutionX/issues/4357.  When running with the build from GitHub with hardware cursor set to ON for items when picking up items they are downsized while holding them with the cursor.  Running with a compiled .exe from Visual Studio doesn't have this behavior and there is no resizing of the item images.  It appears this is resolution dependent, it happens for me at 1280x960 resolution, but likely can happen at other resolutions as well and at some lower resolutions it doesn't happen.

https://user-images.githubusercontent.com/5624187/162288523-98be034b-6cd2-4893-a07b-f5807e472e44.mp4

StephenCWills:
The resource file needs to reference the manifest file, and the manifest file had some errors.

This resolves #4358


AJenbo (approved):
Thanks for digging in to this :)

ikonomov:
... and it also resolves https://github.com/diasurgical/devilutionX/issues/4357 :)  Thank you staphen for fixing this!

Issue #4334, opened 2022-04-03
[touch] Selecting Quit Game opens into the Support menu


AJenbo:
I found that selecting Quit Game with a touchscreen causes a click event on the main menu as well, forcing me into the Support menu option every time.

_Originally posted by @StephenCWills in https://github.com/diasurgical/devilutionX/issues/4331#issuecomment-1086934643_

StephenCWills:
Looks like this is happening because of #3852. Finger down triggers a call to `gmenu_left_mouse(true)` which exits the game, and then finger up triggers the mouse interaction with the main menu. I confirmed this by holding my finger down on the `Quit Game` button without releasing, which didn't trigger the main menu.

Technically this is also happening for the `New Game` option, but the menu screen you end up on has no menu items under the cursor in that region of the screen. Therefore, tapping `New Game` won't trigger any main menu options, but dragging before you release might.

AJenbo:
From looking at other apps I think the correct solution here is to make the ingame menu only trigger on fingerUp as well.

bubio:
Which platform does this occur on?
The 'Quit Game' menu doesn't appear on iOS/Android, does it?

StephenCWills:
I tested on Android. #4331 adds the `Quit Game` option back, but it goes to the main menu instead of closing the app.

AJenbo:
The one in the main menu does not, but the one ingame does

bubio:
I made the latest test build and it shows up.
This should really be enabled by SDL_MOUSEBUTTONONDOWN -> SDL_MOUSEBUTTONUP. I'm not sure how to implement that.🤔

PR #4363, opened 2022-04-08
FIX: Selecting Quit Game opens into the Support menu #4334

bubio:
-  Enabled MOUSEDOWN->MOUSEUP transition as same as UiButton.


.python-version
```diff
@@ -0,0 +1 @@
+3.9.10
```
AJenbo, line 1: What is this for?

bubio: Oops, that's not necessary.

Issue #4365, opened 2022-04-08
[Vita] Huffman decompression failing because of pinter manipulation


AJenbo:
Currently the PS Vita port is unable to use the Polish voice pack as it uses ADPCM+Huffman compression.

https://pastebin.com/rgyH1F9z

ldr     r3, [r3, #0], but r3 is 0x7da849bf (and memory range starts from 0x81000000)

there are a bunch of blocks like this 
```
        /* check if previous item is a pointer. */
        if (PTR_INT(temp) <= 0) {
            temp = PTR_NOT(temp);
        } else {
            temp += (hi - hi->next->prev);
        }
```

but then there is also
```
        /* check if pointer is valid. */
        if (PTR_INT(ht->item305C) < 0) {
            p_prev += (struct huffman_tree_item_s *)pp_item - (*pp_item)->prev;
        } else {
            p_prev += PTR_INT(ht->item305C);
        }
```

PR #4366, opened 2022-04-08
Support MP3 playback

glebm:
May help with #4365 in conjunction with https://github.com/diasurgical/devilutionx-assets/pull/1


AJenbo (approved):
Lets see how this perform on 3DS before merging.

AJenbo:
Looks like the slow down on 3DS isn't related to this so should be fine.

PR #4367, opened 2022-04-08
Use romfs on Switch for devilutionx.mpq

StephenCWills:
Packs devilutionx.mpq into the nro so it won't have to be distributed separately.

Tested with Ryujinx.

Note that we will probably encounter issues similar to 3DS and MacOSX if the user has a leftover version of devilutionx.mpq on their SD card.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -300,6 +300,21 @@ if(APPLE)
 endif()
 
 if(NINTENDO_SWITCH)
+  set(APP_ROMFS       "${CMAKE_BINARY_DIR}/romfs")
+
+  add_custom_target(romfs_directory
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${APP_ROMFS})
+
+  add_custom_target(romfs_files
+    COMMAND ${CMAKE_COMMAND} -E copy ${DEVILUTIONX_MPQ} ${APP_ROMFS}
+    DEPENDS ${DEVILUTIONX_MPQ})
+
+  # the romfs_directory target will create this directory at build time,
+  # but the nx_create_nro function requires it to also exist at configure time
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -300,17 +300,35 @@ if(APPLE)
 endif()
 
 if(NINTENDO_SWITCH)
+  set(APP_ROMFS       "${CMAKE_BINARY_DIR}/romfs")
+
+  add_custom_target(romfs_directory
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${APP_ROMFS})
+
+  add_custom_target(romfs_files
+    COMMAND ${CMAKE_COMMAND} -E copy ${DEVILUTIONX_MPQ} ${APP_ROMFS}
```
glebm, line 309: Is romfs already compressed? If so, it might be better to copy the files directly rather than copying the MPQ, similar to how we handle Vita

```
set(BUILD_ASSETS_MPQ OFF)
```

Target: `devilutionx_copied_assets`

StephenCWills: On 3DS it's not compressed. I can run a test on Switch to confirm.

glebm (approved):

CMakeLists.txt
AJenbo: In that case we can also stop loading devilutionx.mpq on that platform and avoid the version issue that way.

glebm: Oh nevermind, looks like romfs doesn't even support compression https://www.kernel.org/doc/html/latest/filesystems/romfs.html

AJenbo: The difference is between a 15 and 20MB NRO file, is that meaning full on something like the Switch, considering it avoids some confusion for users? The download should be about the same size.

AJenbo (approved):

Issue #4368, opened 2022-04-08
[SDL1] - START + D-PAD-R does not open inventory


jimmiejoe:
Downloaded the 3ds (85962) commit [2450adf]

The video on the 2DS FTR-001, now works, but the intro stutters.
The START + D-PAD-R does not open the inventory.
The START + R does open the inventory.

Expected behavior is that both should open the inventory.

jimmiejoe:
START + D-PAD-U works on 2DS
START + D-PAD-L works on 2DS
START + D-PAD-D works on 2DS
START + L works on 2DS
SELECT + Y works on 2DS
SELECT + B works on 2DS

Not sure what other mouse emulations there are on the 2DS.

glebm:
Does START + D-PAD-L work?

jimmiejoe:
Ok. I just figured out how to get to the correct file.
Pulled 3ds (86019) commit [https://github.com/diasurgical/devilutionX/commit/999fce97826d25436e13afb5fe7f9f2a0788ffaf]
The START + D-PAD-R still does not ope the Inventory, but other emulations work.
I also check the SELECT + D-PAD, but none of the emulation works, but the menu does show. The cursor jumps to the edge of the screen. In addition, you can not bring the menu back up with SELECT unless the character moves at least 1 step. Then you can get the SELECT menu to come up.

jimmiejoe:
Pulled 3ds (86024) commit [4c2fa76]
Same as 3ds (86019) the emulation clicks is not working on the 2DS.

jimmiejoe:
Yes

glebm:
> In addition, you can not bring the menu back up with SELECT unless the character moves at least 1 step. 

The rest sound like bugs but this one intended for D-Pad mouse emulation. It leaves mouse mode on until you press another button, e.g. so that you can use SELECT + L to click.

glebm:
Are you sure you've actually downloaded the correct version?
E.g. from here https://app.circleci.com/pipelines/github/diasurgical/devilutionX/12617/workflows/7405a61c-0372-4d19-9567-cde884c59769/jobs/86040/artifacts

(the current 3ds build appears to be quite slow, this is being investigated separately)

AJenbo:
[...]

PR #4371, opened 2022-04-09
Controller mouse emulation fixes

### LINKED ISSUE #4369: SELECT + D-Pad mouse emulation clicks broken on master
1. Clicks don't work
2. The hint menu is visible during mouse emulation

glebm:
1. Do not interrupt mouse mode on virtual clicks.
2. Handle virtual clicks directly instead of sending an SDL event.
3. Fix D-Pad mouse emulation state handling.
4. Hides the modifier hints during D-Pad mouse emulation.

Fixes #4369

@jimmiejoe, can you please check if this helps with #4368?

### DIFF
--- a/Source/controls/controller_motion.cpp
+++ b/Source/controls/controller_motion.cpp
@@ -12,8 +12,11 @@
 #include "controls/plrctrls.h"
 #include "controls/touch/gamepad.h"
 #include "options.h"
+#include "utils/log.hpp"
 
 namespace devilution {
+
+bool SimulatingMouseWithSelectAndDPad;
 
 namespace {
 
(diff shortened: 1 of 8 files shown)


AJenbo (approved):
Big fan of not ending SDL events for this :+1:

AJenbo (approved):

jimmiejoe:
I pulled down 3ds (86012) commit [56af849].
The START + D-PAD-R still does not ope the Inventory, but other emulations work.

glebm:
> Big fan of not ending SDL events for this +1

@AJenbo Yeah, we pushed an SDL event in the past because we used the real cursor position. Now that we always use emulated cursor for this, there is no longer a need for pushing an SDL event.

PR #4375, opened 2022-04-09
Set initial overwrite path to PrefPath

AJenbo:
As a partial fix for #4007

### DIFF
--- a/Source/init.cpp
+++ b/Source/init.cpp
@@ -151,6 +151,8 @@ void init_cleanup()
 
 void LoadCoreArchives()
 {
+	paths::SetMpqDir(paths::PrefPath());
+
 	auto paths = GetMPQSearchPaths();
 
 #if !defined(__ANDROID__) && !defined(__APPLE__)


obligaron (changes requested):
Sadly this is not enough. 😢 

Following test environment:
- Windows with gog version installed. The mpq are automatically located and loaded from there.
- Set `--config-dir` to another directory (or could be not set, then the executable location is used). Translation is placed in this directory.

Program flow in `DiabloMain`:
- `LoadOptions` gets called. With this pr the initial language check in `OptionEntryLanguageCode::LoadFromIni` works (fails in master, cause `MpqDir` was not set). This is good.
- But then `LoadGameArchives` loads the mpqs from the gog installation and changes `MpqDir`.
- `DiabloInit` calls `LanguageInitialize` and it sees the new `MpqDir` and fails to load the translation. Now we have a not loaded translation but different value in `sgOptions.Language.code`.

As a alternative fix, that can also be discussed, I opened #4377.

PR #4376, opened 2022-04-10
Update magic requirement for stashed books

### LINKED ISSUE #4360: Stash books update when reading a book
I'm using the newest master as of 4/7 and I found a small issue with the Stash books not updating when a book is read.  So if there are 2 books of Healing for example, one requiring magic 15 and the other magic 20, and the character has magic 17, if I read one of the books the other book does not turn red until it is picked up.

StephenCWills:
Adds stash items to the list of items updated in the `CalcPlrBookVals()` function.

This resolves #4360

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -539,12 +539,12 @@ uint8_t CalculateMagicRequirementForBook(spell_id spell, int8_t spellLevel)
 
 void CalcPlrBookVals(Player &player)
 {
-	auto processItem = [](Player &plr, Item &item) {
+	auto processItem = [](Player &player, Item &item) {
 		if (item._itype == ItemType::Misc && item._iMiscId == IMISC_BOOK) {
 			spell_id spell = item._iSpell;
-			int8_t spellLevel = plr._pSplLvl[spell];
+			int8_t spellLevel = player._pSplLvl[spell];
 			item._iMinMag = CalculateMagicRequirementForBook(spell, spellLevel);
-			item._iStatFlag = plr.CanUseItem(item);
+			item._iStatFlag = player.CanUseItem(item);
 		}
 	};
 	for (Item &item : InventoryPlayerItemsRange { player }) {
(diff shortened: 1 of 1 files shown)


Source/items.cpp
```diff
@@ -521,22 +521,38 @@ void CalcPlrItemMin(Player &player)
...
-				if (item._iMinMag + 20 * item._iMinMag / 100 > 255) {
-					item._iMinMag = 255;
-					spellLevel = 0;
-				}
-			}
-			item._iStatFlag = player.CanUseItem(item);
+			spell_id spell = item._iSpell;
+			int8_t spellLevel = plr._pSplLvl[spell];
+			item._iMinMag = CalculateMagicRequirementForBook(spell, spellLevel);
+			item._iStatFlag = plr.CanUseItem(item);
```
AJenbo, line 547: This would be the only place in the code where we write `plr` instead of `player`

Source/items.cpp
```diff
@@ -521,22 +521,38 @@ void CalcPlrItemMin(Player &player)
...
+		}
+	}
+
+	return magicRequirement;
+}
+
 void CalcPlrBookVals(Player &player)
 {
-	for (Item &item : InventoryPlayerItemsRange { player }) {
+	auto processItem = [](Player &plr, Item &item) {
```
AJenbo, line 542: This would be the only place in the code where we write `plr` instead of `player`
```suggestion
	auto processItem = [](Player &player, Item &item) {
```

Source/items.cpp
AJenbo, line 547:
```suggestion
			spell_id spell = item._iSpell;
			int8_t spellLevel = player._pSplLvl[spell];
			item._iMinMag = CalculateMagicRequirementForBook(spell, spellLevel);
			item._iStatFlag = player.CanUseItem(item);
```

AJenbo (approved):

Source/items.cpp
StephenCWills: In C#, Visual Studio would have yelled at me for using the same variable name. I've been conditioned. 😓

PR #4379, opened 2022-04-10
Don't hide the automap in Help and Chat Log

### LINKED ISSUE #4355: Chat Log hides automap
Currently opening the Chat Log hides the Automap, which is necessary, but it would be nice to have the game remember the state of the Automap when Chat Log was opened so that Automap is restored when Chat Log is closed.

glebm:
This was necessary in the past due to stippled transparency but we use blended transparency now.

screenshot

Fixes #4355


qndel (approved):
Yup, looks good enough :)

ikonomov:
Definitely! :)

PR #4380, opened 2022-04-10
Update SDL_audiolib from upstream 

glebm:
Includes updates of drwav and drmp3 to the latest versions

### DIFF
--- a/3rdParty/SDL_audiolib/CMakeLists.txt
+++ b/3rdParty/SDL_audiolib/CMakeLists.txt
@@ -31,6 +31,6 @@ set(USE_DEC_DRMP3 ON)
 
 include(FetchContent)
 FetchContent_Declare(SDL_audiolib
-  URL https://github.com/realnc/SDL_audiolib/archive/aa79660eba4467a44f9dcaecf26b0f0a000abfd7.tar.gz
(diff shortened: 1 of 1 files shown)


AJenbo:
Are there any specific fixes or improvements that affect us here? I see a bit of fixing of warnings and compiler support, but nothing that jumps out at me.

glebm:
My motivation was this entry in the mp3 [changelog](https://github.com/mackron/dr_libs/blob/15f37e3ab01654c1a3bc98cff2a9ca64e8296fa9/dr_mp3.h#L4474-L4475):

```
v0.6.31 - 2021-08-22
  - Fix a bug when loading from memory.
```

AJenbo:
Yeah saw that as well, but assumed we might not be affected since the audio is playing :)

AJenbo:
I'm a bit hesitant to upgrade such a critical component this close to a release. Do you have an option to test it on Windows a bit?

glebm:
As I understand it may be skipping a frame (?) but that would be very difficult to notice

@AJenbo I'll test on Windows!

AJenbo (approved):

glebm:
Tested on Windows and Linux with English, Polish, and Russian

PR #4383, opened 2022-04-10
Quick cast fixes

### LINKED ISSUE #4269: Quick cast does not work with a gamepad
- Enable quick cast in settings
- Start a game
- Bind a spell to the first quick spell hotkey
- Press Select+A button on the game pad
- See hero twitch, but not cast the expected spell

StephenCWills:
When quick casting from gamepad, update the cursor position so the player can quick cast spells in front of them without targeting a player or monster.

When quick casting from staff charges and scrolls, these are now explicitly validated using the spell that was cast rather than `Player::_pRSpell`.

This resolves #4269


obligaron (approved):
lgtm 🙂 

The only downside is that I have to rebase #4074 again. 😜

Issue #3706, opened 2021-12-08
Display freezing on Windows when changing graphics settings with VSync and Upscale On


ephphatha:
**Important information**
Windows 11 / DevilutionX master / compiled through VS2022

**Describe**
With both VSync and Upscale set to On, changing another graphics setting that requires reinitialising the renderer causes the display to freeze.

**To Reproduce**
1. Go to Settings > Graphics
2. Ensure Vertical Sync and Upscale are both On
3. Change resolution, toggle fullscreen, toggle fit to screen, change scaling quality, or toggle integer scaling
4. The display will freeze on the last displayed frame

The game still responds to keyboard input and can be exited/have other settings changed. Turning any of Vertical Sync or Upscale Off will cause the display to update immediately.

**Expected behavior**
Settings changes take effect immediately without freezing

**Additional context**
Unsure if this is a driver issue. I'm on a AMD + Nvidia system, unfortunately I don't have another system I can easily test on.

obligaron:
@ephphatha can you post your `[Graphics]` settings in the ini? Perhaps then I can reproduce the issue.

ephphatha:
Sure, current ini looks like this:
```ini
[Graphics]
Width=1280
Height=720
Fullscreen=0
Upscale=1
Fit to Screen=1
Scaling Quality=2
Integer Scaling=1
Vertical Sync=1
Blended Transparency=1
Gamma Correction=95
Color Cycling=1
Hardware Cursor=1
Hardware Cursor For Items=0
Hardware Cursor Maximum Size=128
FPS Limiter=1
Show FPS=1
```

I can reproduce the issue with a default config (deleting every key and launching devX). This gives me the following settings:
```ini
[Graphics]
Width=640
Height=480
Fullscreen=1
Fit to Screen=1
Upscale=1
Scaling Quality=2
Integer Scaling=0
Vertical Sync=1
Blended Transparency=1
Color Cycling=1
FPS Limiter=1
Show FPS=1
Gamma Correction=100
Hardware Cursor=1
Hardware Cursor For Items=0
Hardware Cursor Maximum Size=128
```

ephphatha:
I was able to reproduce this with a very simple SDL2 program, it's only after the second time initialising a renderer with the SDL_RENDERER_PRESENTVSYNC flag set that causes the issue. No errors get reported though... so not sure why it's failing to update after that point. https://gist.github.com/ephphatha/d0e68a79f38e8762145bdc47a16f8483

[...]

PR #4385, opened 2022-04-10
[Windows] Recreate the whole window if the renderer would otherwise freeze

### LINKED ISSUE #3806: Display freezing on Windows when turning Upscale off with VSync on

StephenCWills:
How about the nuclear option? Inspiration taken from the OpenGL renderer.

Given how d3d11 essentially "poisons" the window so other renderers can't use it, I don't see how the d3d11 renderer and the Upscale option can coexist without this. I expect this should also allow us to use the default renderer instead of setting the render hint.

I'm not sure if recreating the window will actually solve the original issue in #3706 as I don't have Windows 11. Therefore, it would be helpful if someone who is able to reproduce that issue could test this. If we can't wait for that, it shouldn't hurt to go ahead and unconditionally set the render hint to request d3d11.

This resolves #3806

### DIFF
--- a/Source/utils/display.cpp
+++ b/Source/utils/display.cpp
@@ -87,6 +87,47 @@ void CalculatePreferredWindowSize(int &width, int &height)
 		height = mode.h * width / mode.w;
 	}
 }
+
+void FreeRenderer()
+{
+#ifdef _WIN32
+	bool wasD3D9 = false;
+	bool wasD3D11 = false;
+	if (renderer != nullptr) {
+		SDL_RendererInfo previousRendererInfo;
+		SDL_GetRendererInfo(renderer, &previousRendererInfo);
+		wasD3D9 = (std::string_view(previousRendererInfo.name) == "direct3d");
+		wasD3D11 = (std::string_view(previousRendererInfo.name) == "direct3d11");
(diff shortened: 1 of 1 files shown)


obligaron (approved):
lgtm 🙂 

Tested on windows 10.

PR #4388, opened 2022-04-10
Avoid recreating items in PutItem locally

### LINKED ISSUE #4386: Converted Hellfire items can morph in single player
> It is regenerating the entire item from the seed. It is a bit worrisome that this also happens in single player as it could cause converted items to morph.

FYI, this is due to one of the `HoldItem` fixes (#4276). If we need to avoid regenerating the item, then I guess we'll need to revert this and add swap space for `CMD_PUTITEM`.

_Originally posted by @StephenCWills in https://github.com/diasurgical/devilutionX/issues/4382#issuecomment-1094303011_

StephenCWills:
This basically reverts #4276, but introduces a new variable to take the place of `Player::HoldItem` to be used when processing the message locally. This only requires space for one item if you assume that the player won't be able to drop multiple items in one frame. If we want to support dropping more than one item in one frame, we'd have to use something like `std::deque<Item>`.

This resolves #4386

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1837,14 +1837,14 @@ bool TryInvPut()
 	return CanPut(myPlayer.position.tile);
 }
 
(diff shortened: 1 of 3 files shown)


AJenbo:
Still better then the original and we have planed solutions that won't require `std::deque<Item>` :+1:

AJenbo (approved):

PR #4390, opened 2022-04-10
[touch] Spend stat points when clicking the buttons

### LINKED ISSUE #4387: [touch] Spend stat points when clicking the buttons
The way the button for applying stats currently work is inconsistent from how other buttons behave with touch controls.
Rather then functioning as a button it behaves like the cursor for handling inventory items. Touching the button only moves the cursor to that position, you then have to additionally click the main action button to apply stat points.

The reason for it's current behavior was to make it harder to accidentally spend stat points on the wrong category. But there is little to no visual feedback or other indications that this is how it works. A few users have expressed confusion regarding how to apply stat points thinking that the feature is broken.

I think the risk and consequence of people accidentally hitting the wrong stat is small compared to the benefit of having this work more intuitively (even if some have been conditioned with how it currently behaves). Users that wants to be extra careful with how the spend stat points should instead use the dpad to position the cursor.

AJenbo:
Fixes #4387

### DIFF
--- a/Source/controls/touch/event_handlers.cpp
+++ b/Source/controls/touch/event_handlers.cpp
@@ -95,6 +95,17 @@ void HandleBottomPanelInteraction(const SDL_Event &event)
 	}
 }
 
+void HandleCharacterPanelInteraction(const SDL_Event &event)
+{
+	if (!chrflag)
+		return;
+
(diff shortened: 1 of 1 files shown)


StephenCWills (approved):
Works better than I expected.

PR #4407, opened 2022-04-11
Reduce debug log output (changed to verbose)

glebm:
Many of the debug messages are no longer relevant as they come from a fairly stable modules.

Reduces the priority level of these messages from Debug to Verbose.


AJenbo:
THANK YOU! :D

AJenbo (approved):

AJenbo:
Slightly relevant. When running a game via ZT the output is extremely spammy

PR #4408, opened 2022-04-11
update swedish lines

SimonCeder:
@AJenbo Thank you for the reminder!


AJenbo (approved):
Thanks for updating, this was right in the nick of time :)

AJenbo:

PR #4411, opened 2022-04-11
Fix an OOB in CEL/CL2 outline rendering

### LINKED ISSUE #4405: Game crashes when holding buttons on gamepad
**Important information**
Nintendo 3DS (and to some extent, Windows)
Latest master, with some controller patches applied (recently added PRs)
Compiled myself

**Describe**
On 3DS, the game seems to crash periodically when just playing normally in the dungeon. Several attempts to reproduce this on Windows led me to discover memory access violations when exiting the game. Further testing seems to suggest this has something to do with holding down the primary action button on gamepad to attack nearby enemies while simultaneously using the left stick to change targets or cancel the player's movement.

**To Reproduce**
1. Go to the dungeon
2. Hold down the `A` button on your gamepad
3. Use the left stick to wander around the level searching for monsters to interact with
4. Continue to move the left stick around while fighting the monsters
5. Exit the game

**Expected behavior**
The player kills all the monsters, then exits the game.

**Additional context**
I had very mixed results when trying to reproduce this, perhaps largely because I was blindly trying to figure out the cause. It seems like I was able to consistently reproduce this issue by clearing around 50%-75% of dlvl 1 in Single Player on maximum game speed with a godmode Warrior.

A large majority of my testing was done against SDL1, however I did encounter one similar access violation when testing SDL2 on Windows before posting this issue.

glebm:
This OOB happened when rendering a sprite outline where the sprite it is exactly off-screen (touching the border but not visible) on top/bottom while also being only partly off-screen on the left or right.

Fixes #4405

### DIFF
--- a/Source/engine/render/cel_render.cpp
+++ b/Source/engine/render/cel_render.cpp
@@ -436,9 +436,17 @@ void RenderCelOutlineClippedXY(const Surface &out, Point position, const byte *s
 
 	if (position.y == dstHeight) {
 		// After-bottom line - can only draw north.
-		src = RenderCelOutlineRowClipped<SkipColorIndexZero, /*North=*/true, /*West=*/false, /*South=*/false, /*East=*/false,
-		    /*ClipWidth=*/true>(
-		    out, position, src, clipX, color);
+		if (position.x <= 0) {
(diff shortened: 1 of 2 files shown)


AJenbo (approved):
Magic

PR #4412, opened 2022-04-12
Update Korean Translation

nsm53project:


AJenbo (approved):

AJenbo:
100%, nice work :)

AJenbo:

PR #4414, opened 2022-04-12
[gamepad] Reuse repeat logic from mouse

### LINKED ISSUE #4406: Controller primary action repeat fires a bit too frequently
It's kinda hard to open a door with the A button. It opens and closes in the short time I'm pressing the button

AJenbo:
This reverts a large part of https://github.com/diasurgical/devilutionX/pull/4019 and instead reuses the mouse's logic for repeating the last action.

Fixes #4406

This still doesn't add the logic to the touch interface, but that should be relatively trivial at this point I think.

### DIFF
--- a/Source/controls/game_controls.cpp
+++ b/Source/controls/game_controls.cpp
@@ -15,7 +15,6 @@
 #include "doom.h"
 #include "gmenu.h"
 #include "options.h"
-#include "plrctrls.h"
 #include "qol/stash.h"
 #include "stores.h"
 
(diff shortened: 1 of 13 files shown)


Source/controls/game_controls.cpp
```diff
@@ -15,16 +15,15 @@
...
 #include "stores.h"
 
 namespace devilution {
 
 bool start_modifier_active = false;
 bool select_modifier_active = false;
-const ControllerButton ControllerButton_ATTACK = ControllerButton_BUTTON_B;
-const ControllerButton ControllerButton_CAST_SPELL = ControllerButton_BUTTON_X;
+const ControllerButton ControllerButtonPrimery = ControllerButton_BUTTON_B;
+const ControllerButton ControllerButtonSecoundery = ControllerButton_BUTTON_X;
```
StephenCWills, line 26: It was a pain to add a review comment for every usage. It's probably better to just use an IDE to rename anyway. 😓 
```suggestion
const ControllerButton ControllerButtonPrimary = ControllerButton_BUTTON_B;
const ControllerButton ControllerButtonSecondary = ControllerButton_BUTTON_X;
```

StephenCWills (changes requested):

Source/controls/game_controls.cpp
AJenbo: Yeah, don't do that I know how to search and replace :D

StephenCWills (approved):

PR #4425, opened 2022-04-14
Fix conversion of C-string literal to `char*`

glebm:


AJenbo (commented):
I don't get it :)

glebm:
You can't pass a `const char *` (the type of the C-string literal) to a function that accepts a `char *`. It is non-standard behaviour and there is a compiler warning about it.

Anyway, I've rewritten this function to use a `string_view`

AJenbo (approved):
Arh I didn't realize it didn't take a `const` since it simply compares. I think the right solution is to correct the function :+1:

glebm:
It actually had this snippet previously, where it did modify the string:

```c++
	for (unsigned i = 0; i < strlen(expression); i++) {
		if (expression[i] == ';') {
			expression[i] = '\0';
			break;
		}
	}
```

PR #4429, opened 2022-04-14
Remove use of SetICursor and related globals from CheckStashPaste

ephphatha:
Cleaned up the logic for `FindSlotUnderCursor` while I was at it, the code was difficult to read and ended up being simple to express entirely in `CheckStashPaste` instead.


AJenbo (approved):

AJenbo (approved):

AJenbo:
Just 4 usages left and all in inv.cpp :)

PR #4432, opened 2022-04-14
Add zoom on start ini option

ikonomov:
Adding in "zoom on start" option allows starting a game zoomed in without the need to press the Z button for the people that prefer to play this way.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -150,6 +150,11 @@ bool was_ui_init = false;
 void StartGame(interface_mode uMsg)
 {
 	zoomflag = true;
+	if (*sgOptions.Graphics.zoomStart) {
+		zoomflag = false;
+	} else {
+		zoomflag = true;
+	}
 	CalcViewportGeometry();
 	cineflag = false;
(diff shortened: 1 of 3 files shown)


Source/diablo.cpp
```diff
@@ -150,6 +150,11 @@ bool was_ui_init = false;
 void StartGame(interface_mode uMsg)
 {
 	zoomflag = true;
+	if (*sgOptions.Graphics.zoomStart) {
+		zoomflag = false;
+	} else {
+		zoomflag = true;
+	}
```
obligaron, line 157: I think we could implement this [similar](https://github.com/diasurgical/devilutionX/pull/4509#discussion_r907658554) to #4509.

We could replace all occurrences of `zoomflag` with `sgOptions.Graphics.zoom`.
(And revert the condition. I don't understand why `zoomflag` = `true` means no zoom 😕).

With this
- we don't need this code
- if zoom is pressed the state would be remembered
- it's still accessible via settings for players without a keyboard
- it works similar to item highlighting

What do you think? 🙂

PR #4463, opened 2022-04-17
Address type conversion/truncation warnings in LoadPcxPixelsAndPalette

ephphatha:
SDL_RWsize is documented as returning any negative value on error (even though in practice errors are always indicated by a value of -1) so allow for that instead of only checking for unknown size.

There were also a few unnecessary arithmetic operations being done to calculate the size to read, straightforward to simplify.

SDL_CreateRGBSurfaceWithFormat expects width and height to be positive values < 2^16. Could get away with using unsigned if the SDL API did but it doesn't so stick with signed types even when it doesn't really fit the use.

bufferPitch is an oddity, ultimately it's only used to determine a pointer offset so making the param ptrdiff_t as it's more appropriate than size_t.

### DIFF
--- a/Source/DiabloUI/art.cpp
+++ b/Source/DiabloUI/art.cpp
@@ -15,7 +15,6 @@ namespace devilution {
 namespace {
 constexpr size_t PcxHeaderSize = 128;
 constexpr unsigned NumPaletteColors = 256;
-constexpr unsigned PcxPaletteSize = 1 + NumPaletteColors * 3;
 
 bool LoadPcxMeta(SDL_RWops *handle, int &width, int &height, std::uint8_t &bpp)
 {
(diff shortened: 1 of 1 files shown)


AJenbo (approved):

glebm:
Guess I'll rebase https://github.com/diasurgical/devilutionX/pull/4228

Source/DiabloUI/art.cpp
```diff
@@ -30,23 +29,23 @@ bool LoadPcxMeta(SDL_RWops *handle, int &width, int &height, std::uint8_t &bpp)
 }
 
 bool LoadPcxPixelsAndPalette(SDL_RWops *handle, int width, int height, std::uint8_t bpp,
-    uint8_t *buffer, std::size_t bufferPitch, SDL_Color *palette)
+    uint8_t *buffer, std::ptrdiff_t bufferPitch, SDL_Color *palette)
```
glebm, line 32: `bufferPitch` cannot be negative, it's not a difference, `ptrdiff_t` doesn't seem appropriate here.

ephphatha: It's called with `SDL_Surface::pitch` which is currently int, I agree an unsigned type would be more appropriate (and width/height would also be better as unsigned types) but given the use is tied to SDL types I figured this was the best fit for the current caller.

PR #4466, opened 2022-04-17
Play gold sound effect when depositing in stash

ephphatha:
This adds a bit of audible feedback to help the player know that the action had an impact (along with the cursor changing and the count increasing).

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -88,6 +88,7 @@ void CheckStashPaste(Point cursorPosition)
 
 	if (player.HoldItem._itype == ItemType::Gold) {
 		Stash.gold += player.HoldItem._ivalue;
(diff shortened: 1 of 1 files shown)


obligaron (dismissed):
The sound is also played (in master) when depositing with ctrl+click  So this pr make this also more consistent. 🙂 

Related question: Should we also play the sound when withdrawing gold?

ephphatha:
Good point, withdrawing gold goes straight into the inventory instead of ending up in the held item slot (e.g. when splitting a stack). So it'd be appropriate to play a sound then too.

Selling items to a shop also places gold straight into the inventory but it already plays a sound for interacting with a menu item, no need to layer anything on top there.

AJenbo (approved):

PR #4467, opened 2022-04-17
(3DS) Correct C-stick axis mapping

MrHuu:
C-Stick mapping is incorrect for the 3ds platform.

Currently left/right is not functional,
up/down moves the cursor left/right.

This corrects the behavior.

Congrats on the new release!


StephenCWills (approved):
Thanks for testing and the PR. I feel like you mentioned this somewhere before, and it must have slipped my mind. If so, I'm sorry about that.

PR #4478, opened 2022-04-17
Remove Lights[]

BCullen2201:


Source/lighting.cpp
```diff
@@ -1007,18 +1013,20 @@ void ProcessLightList()
...
+			if (lights._lunflag) {
+				DoUnLight(lights.position.old.x, lights.position.old.y, lights.oldRadius);
+				lights._lunflag = false;
 			}
 		}
 		for (int i = 0; i < ActiveLightCount; i++) {
 			int j = ActiveLights[i];
-			if (!Lights[j]._ldel) {
-				DoLighting(Lights[j].position.tile, Lights[j]._lradius, j);
+			auto &lights = Lights[j];
```
yuripourre, line 1027: Maybe `auto &light = Lights[j];` ?

BCullen2201: makes sense

Source/lighting.cpp
```diff
@@ -941,7 +942,8 @@ void AddUnLight(int i)
 		return;
 	}
 
-	Lights[i]._ldel = true;
+	auto &lights = Lights[i];
+	lights._ldel = true;
```
AJenbo, line 946: If it's just a single use there is no need to place it in another variable.
```suggestion
	Lights[i]._ldel = true;
```

Source/lighting.cpp
```diff
@@ -921,14 +921,15 @@ int AddLight(Point position, int r)
 	}
 
 	lid = NO_LIGHT;
+	auto &lights = Lights[lid];
 
 	if (ActiveLightCount < MAXLIGHTS) {
 		lid = ActiveLights[ActiveLightCount++];
```
AJenbo, line 927: If you make the reference before knowing the index (`lid`), then you will be operating on the wrong reference later on.

Also the value should only be declared where in the same scope (`{}`) so that it can't be used in the wrong place by mistake.
```suggestion

	if (ActiveLightCount < MAXLIGHTS) {
		lid = ActiveLights[ActiveLightCount++];
		Light &lights = Lights[lid];
```
(one last thing here, we decided that `auto` should preferably only by used in places where the type also appears on the right side)

BCullen2201: Got it!

Source/lighting.cpp
```diff
@@ -977,10 +980,11 @@ void ChangeLightOffset(int i, Point position)
...
 	}
 
-	Lights[i]._lunflag = true;
-	Lights[i].position.old = Lights[i].position.tile;
-	Lights[i].oldRadius = Lights[i]._lradius;
-	Lights[i].position.offset = position;
+	auto &lights = Lights[i];
+	lights._lunflag = true;
+	lights.position.old = Lights[i].position.tile;
+	lights.oldRadius = Lights[i]._lradius;
```
AJenbo, line 986: Two things I noticed is that you generally have called the reference `lights`, but since it's a reference to a single element from the `Lights` array it should be called `light`. I saw you corrected it in once place, but this was the case for all the changes.

You also missed a couple of usages of `Lights[i]` on the right side, like here and on line 971-972.

I corrected this in the rest of the PR.

Other than that it looks good.

AJenbo (approved):

PR #4479, opened 2022-04-17
Throttle send player commands to avoid desyncs

obligaron:
Workaround for #2681

Notes:
- This implements the [mentioned](https://github.com/diasurgical/devilutionX/issues/2681#issuecomment-932789249) workaround:
   >Avoid sending multiple network messages in RepeatMouseAction (for one game tick).
- When the player should take a new action the logic checks if this command was the last sent action in this game ticks. If yes this is a redundant action and can be discarded. This helps to reduce the impact of #2681 significantly. For a detailed analysis see #2681.
- For local players this shouldn't make a difference, cause game ticks and network messages are always aligned.
- The logic is implemented in the `NetSendCmd` functions to work with normal clicks, repeated mouse actions, touch and gamepads.
- Hopefully this workaround can be removed in future version that's why it is implemented in a way that the code can be removed easily.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -2337,6 +2337,7 @@ void game_loop(bool bStartup)
 		}
 		TimeoutCursor(false);
 		GameLogic();
+		ClearLastSendPlayerCmd();
 
 		if (!gbRunGame || !gbIsMultiplayer || demo::IsRunning() || demo::IsRecording() || !nthread_has_500ms_passed())
 			break;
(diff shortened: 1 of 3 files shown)


StephenCWills:
Given the commands listed in `WasPlayerCmdAlreadyRequested()`, should we also modify `NetSendCmdLocParam4()`, `NetSendCmdParam4()`, and `NetSendCmdParam3()`? These appear to be used for `CMD_SPELLID`, `CMD_TSPELLID`, `CMD_SPELLPID`, `CMD_TSPELLPID`, and `CMD_SPELLXYD`.

obligaron:
Thanks. Fixed. 🙂

StephenCWills (approved):

PR #4481, opened 2022-04-18
Update fr.po

Krabator:
Typos and Quests
French way to display the items description :)

### DIFF
--- a/Translations/fr.po
+++ b/Translations/fr.po
@@ -394,11 +394,11 @@ msgstr "Changer de Passerelle"
 #: Source/DiabloUI/selconn.cpp:108
 msgid "All computers must be connected to a TCP-compatible network."
 msgstr ""
-"Tous les ordinateurs doivent être connectées à un réseau compatible TCP."
+"Tous les ordinateurs doivent être connectés à un réseau compatible TCP."
 
 #: Source/DiabloUI/selconn.cpp:112
 msgid "All computers must be connected to the internet."
-msgstr "Toutes les ordinateurs doivent être connectées à internet."
+msgstr "Tous les ordinateurs doivent être connectés à internet."
 
 #: Source/DiabloUI/selconn.cpp:116
 msgid "Play by yourself with no network exposure."
@@ -571,8 +571,8 @@ msgid ""
 "This is where a starting character should begin the quest to defeat Diablo."
 msgstr ""
 "Difficulté Normale\n"
-"C'est là qu'un personnage débutant devrait commencer sa quête pour vaincre "
-"Diablo."
+"C'est le mode avec lequel un personnage débutant devrait commencer sa quête "
+"pour vaincre Diablo."
 
 #: Source/DiabloUI/selgame.cpp:365
 msgid ""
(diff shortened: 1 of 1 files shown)


AJenbo:
@mmaret @Phalcon22 are one of you available to review this?

Phalcon22Epita:
Just finished reviewing it. I find it to be a great improvement over the previous translation, it is really meticulous.

I have a few things I didn't like but it is just my point of view:
-Many times when something of the form "something of something" is translated it becomes "something - something" for example "Short Staff Of Mana" becomes "Short Staff - Mana" (equivalent in French). I find it a little confusing personnally.
-Long Sword has been translated to Rapier. I don't think the game sprite looks like a Rapier but I can be mistaking.
-The separator for thousands in French is a whitespace, but it was removed here which makes the XP points hard to read.
-Finally the poem Halls of the Blind has been changed but for me it is not an improvement just a different way deal with this translation. Translating the poem is very hard and some choices must be made, the meaning, the shape etc... It is a different way to translate it, I don't know which one is better. Though I think I personally prefer the the old version.

Krabator:
I agree with you on the something of something problem, but I have an issue
Staff of Apocalypse is translated by : "Bâton d'Apocalypse" ou "Bâton de l'Apocalypse" whatever...
With the present tool I only have the possibility to use "item name" + "link work (ex : "de")" + "Spell name"
For my exemple : "Bâton de Apocalypse"... and it's not French
same for "Livre de Inferno"...
I have think about that and fianlly think that it's better to put a "-" separator and keep the translation as much French as possible 

For the item translation, I have copied as much of the D2 french translatio nas possible... in order to keep consistency between the D1 translation and the D2 official translation... I don't like some of the offical choices but i think it's better to keep consistency
source : https://diablo2.judgehype.com/objets-normaux-epees/
you'll see  long sword --> Rapière

For the poem ^^ you are right it's a matter of choice

AJenbo:
I guess if you fix " -The separator for thousands in French is a whitespace, but it was removed here which makes the XP points hard to read." then it's good enough to merge :)

Translations/fr.po
```diff
@@ -2067,7 +2069,7 @@ msgstr "Epée Courte"
 
 #: Source/itemdat.cpp:55 Source/itemdat.cpp:124
 msgid "Buckler"
-msgstr "Bocle"
+msgstr "Écu"
```
mmaret, line 2072: The Wiki page of the buckler (https://en.wikipedia.org/wiki/Buckler) redirect to the "bocle" when looking for the french page.

The "Bocle" or "Buckler" is supposed to be small and round. An ècu could be lager with some painting.
I cannot check how is it in the game right now.

Translations/fr.po
```diff
@@ -2303,19 +2305,19 @@ msgstr "Armure"
...
+msgstr "Armure en Cuir Bouilli"
 
 #: Source/itemdat.cpp:114
 msgid "Studded Leather Armor"
 msgstr "Armure en Cuir Clouté"
 
 #: Source/itemdat.cpp:115
 msgid "Ring Mail"
-msgstr "Broigne"
+msgstr "Cotte de Mailles"
```
mmaret, line 2320: Broigne was too specific ?

Translations/fr.po
```diff
@@ -2548,23 +2550,23 @@ msgstr "Hache"
 
 #: Source/itemdat.cpp:185
 msgid "Large Axe"
-msgstr "Hache Large"
+msgstr "Grosse Hache"
 
 #: Source/itemdat.cpp:186
 msgid "Broad Axe"
-msgstr "Doloire"
+msgstr "Hache Large"
```
mmaret, line 2557: Doloire was too specific ?

mmaret:
I have not finish the review, but it seems like a great improvement to me.
I also just saw your point about the stuff name consistency with d2 and it make sense to mee too

Translations/fr.po
mmaret: :) I'm far from it too !

AJenbo:
If it's not obviously wrong I think following what the other games in the series use is a good idea.

It looks like there are still 3 words that have comments that needs to be resolved on way or the other and then this is ready to be merged?

mmaret:
It's ok for me, you can merge it like that

PR #4485, opened 2022-04-18
ps4: fix crash, minor cleanup

Cpasjuste:


Source/platform/locale.cpp
```diff
@@ -170,6 +170,8 @@ std::vector<std::string> GetLocales()
 	}
 
 	CFRelease(languages);
+#elif PS4
+    locales.emplace_back("en_US");
```
ephphatha, line 174: An empty list is valid as a return value from GetLocales(), you can let the default language be loaded without constructing a temporary string.

```suggestion
    // std::locale is unavailable on PS4 so we just let the fallback language get loaded
```

Comment is a guess at what the cause was, feel free to make it more accurate :D

Cpasjuste: Thanks :)

Source/platform/ps4/sys/endian.h
```diff
@@ -0,0 +1,10 @@
+//
+// Created by cpasjuste on 18/04/22.
+//
+
+#ifndef DEVILUTIONX_ENDIAN_H
+#define DEVILUTIONX_ENDIAN_H
+
+#include <endian.h>
```
glebm, line 8: Perhaps this should simply be fixed in https://github.com/diasurgical/libmpq?

john-tornblom: OrbisOS is based on freebsd, which normally includes <machine/endian.h>. This is also where the official sony sdk puts this file. OpenOrbis uses a modified version of musl, which assumes a linux kernel (and thus puts endian.h directly under /usr/include). This issue will occur on many other projects too, so its best fixed in upstream musl repo if you ask me.

john-tornblom:
I believe the sce_sys and sce_module stuff included in this PR is superfluous.

Source/platform/ps4/sys/endian.h
Cpasjuste: Hi guy,

I'm not planning to support the PS4 port, I just helped a bit on request :) Feel free to do any modifications !

john-tornblom: @Cpasjuste, it is much appreciated! I added my comments to inform reviewers of the situations so they can make an informed decision about the PR. Another note worth mentioning is that progress on openorbis-musl is somewhat sporadic, so perhaps it makes sense to merge this.

Cpasjuste: Yeah no problem mate :)
They just need to know that I won't be able to put more time in it anytime soon. But yes:
- libmpq should be patched (and "platform/ps4" removed), I didn't notice it was a "devix" repo.
- some should verify if pkg modules are really needed (it's possible that they aren't needed anymore since I made some changes to the PS4 "pacbrew" toolchain package manager.

PR #4486, opened 2022-04-18
Refactor TryDropItem to reduce use of globals

ephphatha:
The way TryInvPut and DropItemBeforeTrig interacted was fragile, don't think this would've worked as expected at the best of times. Looks like it was added as part of controller support so guessing this isn't vanilla behaviour anyway.


AJenbo (approved):

AJenbo:
Hmm builds broke after merge

AJenbo:
Arh, the function was renamed in another PR that was merged in between :)

PR #4490, opened 2022-04-18
Clear held item when putting in an empty belt slot

ephphatha:
Missed this one in #4487, which ended up introducing a dupe

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -560,7 +560,8 @@ void CheckInvPaste(Player &player, Point cursorPosition)
 	case ILOC_BELT: {
 		int ii = r - SLOTXY_BELT_FIRST;
 		if (player.SpdList[ii].isEmpty()) {
-			player.SpdList[ii] = player.HoldItem;
+			player.SpdList[ii] = std::move(player.HoldItem);
+			player.HoldItem.Clear();
 		} else {
 			std::swap(player.SpdList[ii], player.HoldItem);
 			if (player.HoldItem._itype == ItemType::Gold)


Source/inv.cpp
```diff
@@ -560,7 +560,8 @@ void CheckInvPaste(Player &player, Point cursorPosition)
 	case ILOC_BELT: {
 		int ii = r - SLOTXY_BELT_FIRST;
 		if (player.SpdList[ii].isEmpty()) {
-			player.SpdList[ii] = player.HoldItem;
+			player.SpdList[ii] = std::move(player.HoldItem);
+			player.HoldItem.Clear();
```
glebm, line 564: Use after move is a bit odd, perhaps `player.HoldItem = {}` or simply:

```c++
std::swap(player.SpdList[ii], player.HoldItem);
```

ephphatha: I based this on the behaviour of std::string as given in the example at https://en.cppreference.com/w/cpp/utility/move, `Item::Clear()` is intended to put an item into a known state from an unspecified state.

All uses of `Item::Clear()` could use `item = {}` instead to initialise the whole object but most of the time it's not really necessary since the object will be copy initialised when it is next used.

ephphatha: Looking into it more I think there's a way to define a move constructor and move assignment operator taking a non-const xvalue (`Item::Item(Item &&)`/`Item &operator=(Item &&)` that could clear the moved from item, but I'm not sure how to define these without explicitly initialising every member of the struct individually... Would be great since it would remove the need to call clear in cases like this.

ephphatha: @glebm how about something like https://github.com/ephphatha/devilutionX/commit/50f94351c6f936e36f97a5d8841ce0da17076046 ? I believe the temporary gets optimised into whatever is most efficient (a copy or a move) so this should be effectively the same once compiled with less thought required at the call site.

glebm: Oh, this was a minor nitpick, sorry to send you down a rabbit hole.

There is nothing wrong with the code, except it may look bugprone to linters and casual readers. https://clang.llvm.org/extra/clang-tidy/checks/bugprone-use-after-move.html

Feel free to keep as-is.

> Would be great since it would remove the need to call clear in cases like this.

That's not how move operators usually behave so it's fine as-is.

glebm (approved):

PR #4492, opened 2022-04-19
Add hint to avoid bugprone-use-after-move warnings

ephphatha:
Makes clang (and GCC) consider item.Clear() a reinitialisation of a moved variable - see https://clang.llvm.org/extra/clang-tidy/checks/bugprone-use-after-move.html#reinitialization

The cpp.hint file is something that MSVC/Visual Studio use to provide better intellisense hints. It needs to be in a parent directory of the file which uses the macro (not where it's defined). It only really needs to indicate whether a macro starts/ends a block, the contents of decorators like DVL_API_FOR_TEST doesn't actually matter in that file.


glebm:
Nice!

glebm (approved):

PR #4493, opened 2022-04-19
Add helper for moving an item

ephphatha:
Used when moving from one persistent variable to another where we want the old variable to be marked empty. Defining as a function instead of move constructor/assignment operator as we only really need to mark the source for xvalues. Detecting that in the constructor/assignment would be needlessly complicated.

### DIFF
--- a/Source/items.h
+++ b/Source/items.h
@@ -243,6 +243,16 @@ struct Item {
 	ItemSpecialEffectHf _iDamAcFlags = ItemSpecialEffectHf::None;
 
 	/**
+	 * @brief Clears this item and returns the old value
+	 */
+	Item pop() &
+	{
+		Item temp = std::move(*this);
+		Clear();
+		return temp;
+	}
+
+	/**
 	 * @brief Resets the item so isEmpty() returns true without needing to reinitialise the whole object
 	 */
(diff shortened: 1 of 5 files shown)


Source/items.h
```diff
@@ -242,6 +242,13 @@ struct Item {
 	uint32_t dwBuff = 0;
 	uint32_t _iDamAcFlags = 0;
 
+	friend Item Move(Item &source)
```
glebm, line 245: Perhaps this could just be a method?

ephphatha: That'd work, declaring a friend method was based on recommendations for defining generic swap functions (so ADL can find the specialisation) but because I used the DevX code style of PascalCase it wouldn't actually be useful in that way :D.

How about calling it Pop() to sounds like STL container pop_front/pop_back methods? In my mind `DoThing(item.Pop())` is easier to guess at the intent than `DoThing(item.Move())`

AJenbo: PascalCase is for globals, members should be camelCase :P

ephphatha: Welp, got a few things to tidy up now :D

AJenbo (approved):

Issue #4469, opened 2022-04-17
To-do list before submission to Flathub


tsunamistate:
- [x] Releases need to provide fully vendored source tarball. Flathub does provide dependencies via [shared-modules](https://github.com/flathub/shared-modules), but they don't have any that are needed for this project. Hell, they don't even have SDL2, only SDL1; Solved via https://github.com/diasurgical/devilutionX/pull/4473
- [x] Flatpak demands that resource files (icons, metadata, `.desktop` files) are renamed to app ID. `devilutionx.desktop` is renamed via `rename-desktop`, need to figure out how to rename icons and Hellfire `.desktop` file. See here: https://github.com/flatpak/flatpak/issues/750; Solved via https://github.com/diasurgical/devilutionX/pull/4468/commits/c2698f8a3588b3da2f5964b9de6844c16018e75a and https://github.com/diasurgical/devilutionX/pull/4468/commits/b51dee16c0638226bb00055621ec09d3ce48dbaf
- [x] Build probably creates unnecessary files, this is handled [through cleanup section of manifest](https://docs.flatpak.org/en/latest/manifests.html#cleanup). Need to figure out what needs to be cleaned up; Nothing to clean up, props to @AJenbo, @glebm and @StephenCWills who worked on CMake install logic
- [x] Copy `devilutionx.mpq` to `XDG_DATA_HOME/diagurgical/devilution` at install or run time; No need to copy, fixed by https://github.com/diasurgical/devilutionX/pull/5306
- [ ] Import dialogue to avoid manually copying MPQs to sandbox. See https://github.com/diasurgical/devilutionX/issues/4470, not required
- [ ] Add Discord integration for x86/64 (?). Discord SDK was added via https://github.com/diasurgical/devilutionX/pull/4430, but flatpak-builder fails to find it when building with `-DDISCORD_INTEGRATION=ON`. IMO, this shouldn't be a priority for first release on Flathub
- [x] Add Appstream metadata file, handle its installation to standard location via CMake. Metadata format described here: https://www.freedesktop.org/software/appstream/docs/; Flatpak requirements for metadata here: https://docs.flatpak.org/en/latest/conventions.html#appdata-files; https://github.com/diasurgical/devilutionX/pull/4497

Pimmelton:
Bump.  12 months later.  I suggest the remaining two items (MPQ import dialog) and Discord integration be removed as blockers for Flathub submission.  

[...]

PR #4497, opened 2022-04-19
Add Appstream metainfo

tsunamistate:
This is a required file for Flathub submission, see #4469 
More info on what Appstream is can be seen here: https://www.freedesktop.org/software/appstream/docs/

This file needs to be installed as  `/usr/share/metainfo/org.diasurgical.DevilutionX.metainfo.xml`
Optionally, this file can be localized. From Appstream docs:

AppStream MetaInfo files are recognized by `xgettext`, so you just need to add `org.diasurgical.DevilutionX` to your `POTFILES.in` to generate templates for all translatable elements in the file. When building the software, you can then instruct your build system to merge in the translations into the new file before installing the translated file as `/usr/share/metainfo/org.diasurgical.DevilutionX.metainfo.xml`. 

@glebm could you please look into CMake logic to install, and optionally localize this file?


tsunamistate:
Rebased and added (untested) logic to install the metainfo file to appropriate location
This is my first time messing with CMake, hopefully it doesn't break anything

Still no idea how to rig up localization, which would be preferable

glebm:
Sorry, I’m away with only an iPad until May 6th

tsunamistate:
@glebm it's OK, I will wait

Packaging/nix/devilutionx.metainfo.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.diasurgical.DevilutionX</id>
+  
+  <name>DevilutionX</name>
```
glebm, line 5: Let's just localize this file inline? Seems simpler than doing it via gettext and xml merging.

Packaging/nix/org.diasurgical.DevilutionX.yml
```diff
@@ -4,13 +4,16 @@ runtime-version: '21.08'
...
+rename-icon: devilutionx
+
 finish-args:
     - --socket=fallback-x11
     - --socket=wayland
     - --device=dri
     - --share=network
+    - --share-ipc
     - --socket=pulseaudio
-    - --filesystem=xdg-config
```
glebm, line 13: Is this not needed?

glebm (approved):

Packaging/nix/org.diasurgical.DevilutionX.yml
tsunamistate: We already have write access in `~/.var/app/$FLATPAK_ID`
Changing the path of diablo.ini from `~/.var/app/$FLATPAK_ID/data/diasurgical/devilution`, which is `XDG_DATA_HOME` to `~/.var/app/$FLATPAK_ID/config`, which is `XDG_CONFIG_HOME` does not require additional permissions

Packaging/nix/devilutionx.metainfo.xml
tsunamistate: I don't know what you're talking about, but whatever works :)

Packaging/nix/devilutionx.metainfo.xml
```diff
@@ -0,0 +1,39 @@
...
+    <control>pointing</control>
+    <control>keyboard</control>
+    <control>touch</control>
+    <control>gamepad</control>
+  </supports>
+  
+  <description>
+    <p>
+      DevilutionX is a source port of Diablo and Hellfire that strives to make it simple to run the game while providing engine improvements, bugfixes, and some optional quality of life features. This includes support for modern operating systems and higher resolutions, controller and touch control support, stash, optional experience and enemy health bars, etc.
+    </p>
```
tsunamistate, line 21: Thinking more about it, I think "reimplementation" is more appropriate for project description.
We're not a "source port" since Diablo's source was never released.

And "build" is not appropriate since the scope of the project went FAR beyond just compatibility with modern OS's.

AJenbo:
@tsunamistate what is our `POTFILES.in` :)

Packaging/nix/devilutionx.metainfo.xml
qndel: It didn't have to be released if we recreated it 100% ...

tsunamistate: And that's why "reimplementation" is a good word
Since Activision-Blizzard-King never released the source code, you decompiled and reimplemented it yourself

Remember: this text will be seen in stores like Gnome Software Center and KDE Discover, so semantics are important

qndel: whatever, I've never even heard about flathub / these other things before xD

AJenbo: We could align the text probably the text should align between snap craft, Google play, Github, Xbox store and switch homebrew store

tsunamistate: It would be better to use "reimplementation" everywhere

Trihedraf: It is definitely a source port. The source was binary exact and then ported to other systems. It didn't start out as something that kind of matched vanilla enough that it kind of worked with the D1 assets.

I strongly disagree with changing it away from source port

tsunamistate: @Trihedraf source port needs source code to be released.
GZDoom is a source port, FreeSpace 2 Source Code Project is a source port. They took the code that was released and changed it with new features.

Diablo never had a source code release, Devilution is a decompilation. Decompilation =/= source code release, even if the code is binary exact, and definitely doesn't save us from litigation (see fate of re3, GTA3 decompilation).

If you disagree that strongly, let's move discussion to Discord

tsunamistate:
Rebased to master
Since the "Source port -> Reimplementation" change really ruffled some feathers, I have reverted it.

tsunamistate:
If everything is okay, let's get this merged please :)

AJenbo (approved):

PR #4500, opened 2022-04-20
Validate remote players in multiplayer games

StephenCWills:
This change rejects `PT_MESSAGE` packets from players that have not yet sent a valid player info packet to the local client. This prevents situations where an unannounced player can send `PT_MESSAGE` packets to other players, affecting their game without them knowing the rogue client was present.

The conditions for player validation were lifted from `UnpackPlayer()` because it's used by `recv_plrinfo()` to decode the player info data sent by the remote client. In addition, we also validate the player's name against empty string since this was recently confirmed to cause issues.

A new function was added to time out players that do not send their player info within 15 seconds of joining a game session. This prevents potentially malicious players from lurking in games for extended periods of time, which otherwise would have enabled them to listen in on game chat and prevent legitimate players from joining.

### DIFF
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -26,6 +26,7 @@
 #include "utils/endian.hpp"
 #include "utils/language.h"
 #include "utils/stdcompat/cstddef.hpp"
+#include "utils/stdcompat/string_view.hpp"
 
 namespace devilution {
 
@@ -56,6 +57,7 @@ bool PublicGame;
 BYTE gbDeltaSender;
 bool sgbNetInited;
 uint32_t player_state[MAX_PLRS];
+Uint32 playerInfoTimers[MAX_PLRS];
 
 /**
  * Contains the set of supported event types supported by the multiplayer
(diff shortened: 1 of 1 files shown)


Source/multi.cpp
```diff
@@ -133,6 +135,38 @@ void NetReceivePlayerData(TPkt *pkt)
...
 
+bool IsNetPlayerValid(const Player &player)
+{
+	return player._pLevel >= 1
+	    && player._pLevel <= MAXCHARLEVEL
+	    && static_cast<uint8_t>(player._pClass) < enum_size<HeroClass>::value
+	    && player.plrlevel < NUMLEVELS
+	    && player.pDifficulty <= DIFF_LAST
+	    && InDungeonBounds(player.position.tile)
+	    && string_view(player._pName) != "";
```
AJenbo, line 146:
```suggestion
	    && !string_view(player._pName).empty();
```

AJenbo:
When testing I am no longer able to rejoin a game after leaving it, before restarting the client.

AJenbo:
Ok, looks like that can already happen if you where the host and left the game. But it's just not super constant. The progress bar just hangs when trying to connect.

AJenbo (approved):

PR #4506, opened 2022-04-22
Map Start+b to toggle item highlighting

beelzebielsk:


AJenbo:
Alooks like a good option. But it should also have an icon in the circle menu:

image

Are you up for making an icon or would you like help with that part?

beelzebielsk:
I'd need help with the icon. I'm open to changing it to Start+A (that's the bottom button, right?)

AJenbo:
I don't have an xbox or nintendo controller :shrug: 

Here is an image of a Switch controller
image

I'l work on an icon.

AJenbo:
How about this?

image

[hinticons.zip](https://github.com/diasurgical/devilutionX/files/8544110/hinticons.zip)

beelzebielsk:
Thinking on it again, my suggestion breaks immersion. Other game icons show something that could exist in the game world, like a map. Though, item labels are also kinda immersion breaking. But I took a crack at what I had in mind anyway, just in case it would help. I chose an axe 'cause I think it is recognizable at a low resolution, and the name is short.

image

AJenbo:
Having text in icons isn't a good idea since they will look out of place in other languages and will lead to some confusion.

Icons also do require a bit of training but once the meaning is learned it quick to recall (why does a grail open the quest log). For now I think we can just tick with this icon and then they can be overhauled at later by some one with better UX/Graphics skills :)

PR #4509, opened 2022-04-23
show item labels toggle in menu

Trihedraf:
Makes it so controller and mobile can have item labels on. Also convenient for not having to press right control every time you start the game.

image
Screenshot on iPhone which previously was very annoying to enable

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -986,6 +986,7 @@ GameplayOptions::GameplayOptions()
     , autoEquipJewelry("Auto Equip Jewelry", OptionEntryFlags::None, N_("Auto Equip Jewelry"), N_("Jewelry will be automatically equipped on pickup or purchase if enabled."), false)
     , randomizeQuests("Randomize Quests", OptionEntryFlags::CantChangeInGame, N_("Randomize Quests"), N_("Randomly selecting available quests for new games."), true)
     , showMonsterType("Show Monster Type", OptionEntryFlags::None, N_("Show Monster Type"), N_("Hovering over a monster will display the type of monster in the description box in the UI."), false)
+    , showItemLabels("Show Item Labels", OptionEntryFlags::None, N_("Show Item Labels"), N_("Show labels for items on the ground when enabled."), false)
     , autoRefillBelt("Auto Refill Belt", OptionEntryFlags::None, N_("Auto Refill Belt"), N_("Refill belt from inventory when belt item is consumed."), false)
     , disableCripplingShrines("Disable Crippling Shrines", OptionEntryFlags::None, N_("Disable Crippling Shrines"), N_("When enabled Cauldrons, Fascinating Shrines, Goat Shrines, Ornate Shrines and Sacred Shrines are not able to be clicked on and labeled as disabled."), false)
     , quickCast("Quick Cast", OptionEntryFlags::None, N_("Quick Cast"), N_("Spell hotkeys instantly cast the spell, rather than switching the readied spell."), false)
@@ -1016,6 +1017,7 @@ std::vector<OptionEntryBase *> GameplayOptions::GetEntries()
 		&experienceBar,
 		&enemyHealthBar,
 		&showMonsterType,
+		&showItemLabels,
 		&disableCripplingShrines,
 		&quickCast,
 		&autoRefillBelt,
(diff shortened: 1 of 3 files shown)


Source/options.h
```diff
@@ -467,6 +467,8 @@ struct GameplayOptions : OptionCategoryBase {
 	OptionEntryBoolean randomizeQuests;
 	/** @brief Indicates whether or not monster type (Animal, Demon, Undead) is shown along with other monster information. */
 	OptionEntryBoolean showMonsterType;
+	/** @brief Displays Item Labels for Items on the Ground.  */
```
qndel, line 470:
```suggestion
	/** @brief Displays item labels for items on the ground.  */
```

Source/options.cpp
```diff
@@ -845,6 +845,7 @@ GameplayOptions::GameplayOptions()
     , autoEquipJewelry("Auto Equip Jewelry", OptionEntryFlags::None, N_("Auto Equip Jewelry"), N_("Jewelry will be automatically equipped on pickup or purchase if enabled."), false)
     , randomizeQuests("Randomize Quests", OptionEntryFlags::CantChangeInGame, N_("Randomize Quests"), N_("Randomly selecting available quests for new games."), true)
     , showMonsterType("Show Monster Type", OptionEntryFlags::None, N_("Show Monster Type"), N_("Hovering over a monster will display the type of monster in the description box in the UI."), false)
+    , showItemLabels("Show Item Labels", OptionEntryFlags::None, N_("Show Item Labels"), N_("Enables Item Labels for Items on the Ground."), false)
```
qndel, line 848:
```suggestion
    , showItemLabels("Show Item Labels", OptionEntryFlags::None, N_("Show Item Labels"), N_("Enables item labels for items on the ground."), false)
```

Source/diablo.cpp
```diff
@@ -148,6 +148,9 @@ bool was_ui_init = false;
 void StartGame(interface_mode uMsg)
 {
 	zoomflag = true;
+	if (*sgOptions.Gameplay.showItemLabels && !IsHighlightingLabelsEnabled()) {
+		ToggleItemLabelHighlight();
+	}
```
obligaron, line 153: Could we replace all occurrences of `invertHighlightToggle` with `sgOptions.Gameplay.showItemLabels`?

With this
- we don't need this code
- if right control is pressed the state would be remembered
- it's still accessible via settings for players without a keyboard

What do you think? 🙂

Trihedraf: It's a good idea. But I can't figure out how to actually set the sgOptions.Gameplay.showItemLabels. For ToggleItemLabelHighlight if you do ```*sgOptions.Gameplay.showItemLabels = !* sgOptions.Gameplay.showItemLabels``` it complains and I tried a bunch of different ways, but it always gives an error. I think that I need to find out how to use the ini saving function somehow in there as the way to change it but that is above my head.

obligaron: `sgOptions.Gameplay.showItemLabels.SetValue(!*sgOptions.Gameplay.showItemLabels)` should work 🙂

Trihedraf: That worked thanks. Has been pushed with the changes.

Source/options.cpp
```diff
@@ -986,6 +986,7 @@ GameplayOptions::GameplayOptions()
     , autoEquipJewelry("Auto Equip Jewelry", OptionEntryFlags::None, N_("Auto Equip Jewelry"), N_("Jewelry will be automatically equipped on pickup or purchase if enabled."), false)
     , randomizeQuests("Randomize Quests", OptionEntryFlags::CantChangeInGame, N_("Randomize Quests"), N_("Randomly selecting available quests for new games."), true)
     , showMonsterType("Show Monster Type", OptionEntryFlags::None, N_("Show Monster Type"), N_("Hovering over a monster will display the type of monster in the description box in the UI."), false)
+    , showItemLabels("Show Item Labels", OptionEntryFlags::None, N_("Show Item Labels"), N_("Enables item labels for items on the ground."), false)
```
AJenbo, line 989:
```suggestion
    , showItemLabels("Show Item Labels", OptionEntryFlags::None, N_("Show Item Labels"), N_("Show labels for items on the ground by default."), false)

```

AJenbo:
With an update to the description i think this is good to go.

Source/options.cpp
Trihedraf: Should it be by default though? Since if you hit right control in game it will switch it off. I would say, "when enabled." ?

obligaron: What about "Permanent show labels for items on the ground."

In keymapper it is:
```cpp
	    "Toggle Item Highlighting",
	    N_("Toggle item highlighting"),
	    N_("Permanent show/hide items on ground."),
```

AJenbo:
Nicely done :)

Source/options.cpp
AJenbo: Doesn't alt switch it of temp?

PR #4510, opened 2022-04-24
Fix viewport size of UiList on settings menu

### LINKED ISSUE #4476: [android] Crash in settings menu from mouse handeling
This looks similar to the issue in the hero menu, but I haven't been able to replicate it yet.

```
backtrace:
libc.so (abort+166)
libdevilutionx.so (offset 0x1000) (devilution::(anonymous namespace)::HandleMouseEvent(SDL_Event const&, devilution::UiItemBase*) (.llvm.7318811127165588370)) (SourceCode: libgcc2.c)
libdevilutionx.so (offset 0x1000) (devilution::UiFocusNavigation(SDL_Event*))
libdevilutionx.so (offset 0x1000) (devilution::UiPollAndRender(std::__ndk1::function<bool (SDL_Event&)>))
libdevilutionx.so (offset 0x1000) (devilution::UiSettingsMenu())
```

StephenCWills:
Fixes a crash when clicking in a specific region on certain submenus within the settings menu.

The viewport size for the `UiList` was hardcoded to 8 (`rectList.size.height / 26`), meaning that the calculated size of that list was too large for the actual number of entries when viewing submenus with fewer than 6 options.

This resolves #4476

### DIFF
--- a/Source/DiabloUI/settingsmenu.cpp
+++ b/Source/DiabloUI/settingsmenu.cpp
@@ -347,7 +347,8 @@ void UiSettingsMenu()
 		vecDialogItems.push_back(std::make_unique<UiListItem>("", static_cast<int>(SpecialMenuEntry::None), UiFlags::ElementDisabled));
 		vecDialogItems.push_back(std::make_unique<UiListItem>(_("Previous Menu"), static_cast<int>(SpecialMenuEntry::PreviousMenu), UiFlags::ColorUiGold));
 
-		vecDialog.push_back(std::make_unique<UiList>(vecDialogItems, rectList.size.height / 26, rectList.position.x, rectList.position.y, rectList.size.width, 26, UiFlags::FontSize24 | UiFlags::AlignCenter));
+		size_t maxViewportSize = rectList.size.height / 26;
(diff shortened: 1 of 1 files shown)


Source/DiabloUI/settingsmenu.cpp
```diff
@@ -347,7 +347,8 @@ void UiSettingsMenu()
 		vecDialogItems.push_back(std::make_unique<UiListItem>("", static_cast<int>(SpecialMenuEntry::None), UiFlags::ElementDisabled));
 		vecDialogItems.push_back(std::make_unique<UiListItem>(_("Previous Menu"), static_cast<int>(SpecialMenuEntry::PreviousMenu), UiFlags::ColorUiGold));
 
-		vecDialog.push_back(std::make_unique<UiList>(vecDialogItems, rectList.size.height / 26, rectList.position.x, rectList.position.y, rectList.size.width, 26, UiFlags::FontSize24 | UiFlags::AlignCenter));
+		size_t maxViewportSize = rectList.size.height / 26;
+		vecDialog.push_back(std::make_unique<UiList>(vecDialogItems, std::min<size_t>(maxViewportSize, vecDialogItems.size()), rectList.position.x, rectList.position.y, rectList.size.width, 26, UiFlags::FontSize24 | UiFlags::AlignCenter));
```
ephphatha, line 351: There's one other call that creates a UiList with fewer than the viewportSize items in selhero.cpp, could potentially put the std::min call/logic in the constructor for UiList itself.

ephphatha (approved):

AJenbo:
Picked the alternative (https://github.com/diasurgical/devilutionX/pull/4511) instead so that we can hopefully stop chasing this :D

PR #4511, opened 2022-04-24
Fix viewport size of UiList in constructor

### LINKED ISSUE #4476: [android] Crash in settings menu from mouse handeling

StephenCWills:
Fixes a crash that occurs when clicking in a specific region on `UiList` elements that are made larger than their contents by resizing them automatically based on their contents.

Based on the following suggestion: https://github.com/diasurgical/devilutionX/pull/4510#discussion_r857054873
Alternative to #4510
This resolves #4476

### DIFF
--- a/Source/DiabloUI/selhero.cpp
+++ b/Source/DiabloUI/selhero.cpp
@@ -484,7 +484,7 @@ void selhero_List_Init()
 	}
 	vecSelHeroDlgItems.push_back(std::make_unique<UiListItem>(_("New Hero").c_str(), static_cast<int>(selhero_SaveCount)));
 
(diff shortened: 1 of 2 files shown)


AJenbo:
Considering that I think there is maybe more cases of this from what I'm seeing in the crash logs I much like this solution :)

PR #4525, opened 2022-04-28
HP bar tweaks and fixes

qndel:
leosteal

This handles some hp/hp bar stuff
1. fixes vanilla bug where monsters could reach HP within a single regen tick above their max (was necessary because monsters regenerating over their max would show as having more than 1 health bar with this PR which was obviously not the goal)
2. fixes devx hp bar - it was inaccurate earlier because of using wrong width - was showing full bar when monster actually had 95% HP
3. People usually thought something was broken when Leo stole hp from them and they damaged him and his HP bar didn't budge - this solves the issue by borrowing the "boss hp bar" mechanic from some games - a number appears saying how many hp bars the boss actually has.

Recolored extra health bars blue 
image

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4311,6 +4311,7 @@ void ProcessMonsters()
 			} else {
 				monster._mhitpoints += monster.mLevel;
 			}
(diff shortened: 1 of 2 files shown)


StephenCWills (dismissed):
I see nothing but improvements in this PR.

StephenCWills (approved):

Issue #4524, opened 2022-04-28
Wrong displacement on death


qndel:
**Important information**
Operating system / DevilutionX version / did you download a release or did you compile yourself?
Compiled master - https://github.com/diasurgical/devilutionX/commit/0bbb7a1572220587cd212b5e1bb319b21aac4a44

**Describe**
Weird displacement while dying

**To Reproduce**
Go die
Watch your character

**Expected behavior**
Don't move on death

**Screenshots**
![deathdisplace](https://user-images.githubusercontent.com/14297035/165767382-4d74cff6-f50b-43e7-ac8d-2264111d9c2a.gif)

qndel:
Bisected to 
commit ac0bd7bb582b27cdd41813c589e5c4c5a026bead
Author: obligaron <obligaron@live.com>
Date:   Sun Jul 25 10:29:17 2021 +0200
    DrawPlayer: Calculate Width2 later

PR #4528, opened 2022-04-29
fix displacement on death

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/4524
```cpp
const Displacement center { CalculateWidth2(player.AnimInfo.celSprite ? player.AnimInfo.celSprite->Width() : 96), 0 };
```
This part got moved to DrawPlayer already and removed from DrawPlayerHelper but not from DrawDeadPlayer which executed it twice while drawing a dead player.


obligaron (approved):
lgtm
thanks 🙂

Issue #1874, opened 2021-05-05
Shrines need to effect stats & equipment randomly from the start of a game.


DrGonzo564:
**Important information**
Mac OS 10.15.7 / DevilutionX 1.2.1 / Downloaded Release

**Describe**
Due to a design flaw in how the game engine handles the random effects in shrines, Hidden Shrines will tend to reduce the durability of the item in the right slot (left hand) more often. Mysterious Shrines will tend to raise Dexterity more often than other stats. Enchanted shrines will tend to reduce Chain Lightning more often than the other spells. This effect becomes less noticeable the more you play in a single gaming session, i.e. the shrine effects will become more 'random.' Namely, the "random seed" stored with ANY shrine is not initialized properly, and the shrine uses whatever happens to be in that memory location. At startup this memory will mostly be "empty", which is read as zero, which then causes the non-random behavior. After the game is played for a while, the memory is much less likely to be zero, and therefore tends to approach randomness.
(I copy/pasted this from https://www.lurkerlounge.com/forums/thread-12422.html)

**Expected behaviour**
They should be randomized from the start of a new game.

**Additional context**
Example my Warrior is lvl 40 and my spells are all high level 9-15 but CL is lvl 4 and my left hand shield only has 10 durability due to this bug.

AJenbo:
To Reproduce and Screenshots sections are misleading, please fill them out or remove them

julealgon:
I actually thought the vanilla behavior was because "the random number generator was bad" (@qndel ) but this points to something else entirely. Seems like a very sensible thing to fix.

qndel:
This is caused by "the random number generator is bad"

AJenbo:
The guide is somewhere between wrong and misleading on this point:
https://github.com/diasurgical/devilutionX/blob/master/Source/objects.cpp#L3432

julealgon:
Thanks for clarifying @qndel / @AJenbo .

qndel:
Turns out the cause is different
Found by @mewmew 
lshrine and rshrine will use whatever seed value is present in that object, which may be a garbage value from a previous object present in the same memory location
if the object at that object index was a door, a candle or some other object that didn't need a seed, then the shrine object at the next dlvl would also have a zero seed = that's what is causing the bias

StephenCWills:
[...]

PR #4529, opened 2022-04-30
fix shrine rng

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/1874

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1219,6 +1219,7 @@ void AddBarrel(Object &barrel)
 
 void AddShrine(int i)
 {
+	Objects[i]._oRndSeed = AdvanceRndSeed();
 	bool slist[NumberOfShrineTypes];
 
 	Objects[i]._oPreFlag = true;


Apocalypse612:
Just tested this, and my chain lightning is now leveling from Enchanted Shrines. Glorious!

StephenCWills:
This change will cause dungeon generation to diverge between mid2 and mid3.

Before:
image

After:
image

qndel:
```cpp
				[[maybe_unused]] uint32_t mid2Seed = GetLCGEngineState();
				IncProgress();
				InitMonsters();
				InitItems();
				if (currlevel < 17)
					CreateThemeRooms();
				IncProgress();
				[[maybe_unused]] uint32_t mid3Seed = GetLCGEngineState();
				InitMissiles();
				InitCorpses();
```
Shrines affect CreateThemeRooms so only theme rooms would be affected
Do you think I should move assigning seed to them after theme rooms finish generating?

StephenCWills:
Hmm.. Yeah, I think it would be nice if we could rely on the seed to generate the same theme rooms in vanilla in case we need to test something specific. Assigning the seed at the end would certainly solve that problem.

StephenCWills (approved):

ikonomov:
This is absolutely fantastic.

AJenbo (approved):
Rolled back to the original proposal as it turns out not to be an issue :+1:

PR #4534, opened 2022-05-02
minor alignment adjustment

ikonomov:
It is still not perfect, but I think that's how it was intended to be aligned and then probably got misaligned after some change.  The alignment looks slightly different in GitHub and Visual Studio, I made this edit according to how it looks in Visual Studio.


StephenCWills (dismissed):
The difference in alignment is due to GitHub rendering tabs as 8 spaces whereas Visual Studio renders them as 4 spaces. If you're so inclined, you can change it in your GitHub account settings.

qndel:
This still looks bad ...

qndel (changes requested):
This still isn't aligned properly

StephenCWills:
@qndel What's the issue? It looks okay to me. Maybe the first column header could be shifted left, but it's not really a problem visually.

image

qndel:
image

StephenCWills:
Your alignment is off by 4 spaces because of the way GitHub renders tabs. Are you saying the tabs need to be removed?

qndel:
That too, I don't think we are using tabs anywhere else for alignment?

ikonomov:
However you guys think looks best.  I simply noticed that it was way off currently.

AJenbo:
tab should only be used for indentation, never for alignment

StephenCWills (approved):

PR #4536, opened 2022-05-03
Check equipment/stats of other players

qndel:
inspectdemo
Automatically deactivates after opening stash or entering dungeon

::update
doesn't deactivate after entering dungeon anymore since there's a command for it now so don't have to worry about having to run back to farnham = whatever ;)

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -407,10 +407,45 @@ std::string TextCmdArenaPot(const string_view parameter)
 	return ret;
 }
 
+std::string TextCmdInspect(const string_view parameter)
+{
+	std::string ret;
+	if (!gbIsMultiplayer) {
+		StrAppend(ret, _("Inspecting only supported in multiplayer."));
+		return ret;
+	}
+
+	if (parameter.empty()) {
+		StrAppend(ret, _("Stopped inspecting players."));
+		InspectPlayer = MyPlayer;
+		return ret;
+	}
+
+	std::string param { parameter.data() };
+	std::transform(param.begin(), param.end(), param.begin(), [](unsigned char c) { return std::tolower(c); });
+	for (auto &player : Players) {
+		std::string playerName { player._pName };
+		std::transform(playerName.begin(), playerName.end(), playerName.begin(), [](unsigned char c) { return std::tolower(c); });
+		if (playerName.find(param) != std::string::npos) {
+			InspectPlayer = &player;
+			StrAppend(ret, _("Inspecting player: "));
+			StrAppend(ret, player._pName);
+			OpenCharPanel();
+			if (!sbookflag)
+				invflag = true;
+			RedrawEverything();
+			return ret;
+		}
+	}
+	StrAppend(ret, _("No players found with such a name"));
+	return ret;
+}
+
 std::vector<TextCmdItem> TextCmdList = {
 	{ N_("/help"), N_("Prints help overview or help for a specific command."), N_("({command})"), &TextCmdHelp },
 	{ N_("/arena"), N_("Enter a PvP Arena."), N_("{arena-number}"), &TextCmdArena },
 	{ N_("/arenapot"), N_("Gives Arena Potions."), N_("{number}"), &TextCmdArenaPot },
+	{ N_("/inspect"), N_("Inspects stats and equipment of another player."), N_("{player name}"), &TextCmdInspect },
 };
 
 bool CheckTextCommand(const string_view text)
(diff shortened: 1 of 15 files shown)


obligaron:
This pr would help with testing multiplayer quests. 🙂 
And now we have command support (since #4599) that means we could use this with `/inspect [PlayerName]` or `/inspect [PlayerNumber]` or something similar.

AJenbo:
I think that would work

Source/cursor.cpp
```diff
@@ -243,7 +243,7 @@ void FreeHalfSizeItemSprites()
 
 void DrawItem(const Item &item, const Surface &out, Point position, ClxSprite clx)
 {
-	const bool usable = item._iStatFlag;
+	const bool usable = MyPlayer == InspectPlayer ? item._iStatFlag : InspectPlayer->CanUseItem(item);
```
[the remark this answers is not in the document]
AJenbo: Hm that kinda sounds like an issue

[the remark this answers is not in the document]
qndel: that's why I made the comment :P

Source/multi.cpp
obligaron, line 1: Could we add the following code in `PlayerLeftMsg`:

```
	if (&player == InspectPlayer)
		InspectPlayer = MyPlayer;
```

This would prevent seeing the stats & items of a left player and avoid undefined behavior.

Source/diablo.cpp
```diff
@@ -407,7 +407,7 @@ void RightMouseDown(bool isShiftHeld)
 {
 	LastMouseButtonAction = MouseActionType::None;
 
-	if (gmenu_is_active() || sgnTimeoutCurs != CURSOR_NONE || PauseMode == 2 || MyPlayer->_pInvincible) {
+	if (gmenu_is_active() || sgnTimeoutCurs != CURSOR_NONE || PauseMode == 2 || MyPlayer->_pInvincible || MyPlayer != InspectPlayer) {
```
obligaron, line 410: Why check here `InspectPlayer`?
The check could be done in `UseInvItem` (similar to `CheckInvItem`).
This would also help with enabling this feature for gamepad & touch users.

Perhaps @StephenCWills could help with pointing out where other checks are needed for gamepad & touch support.

Source/control.cpp
```diff
@@ -383,9 +383,40 @@ std::string TextCmdArena(const string_view parameter)
...
+	}
+	StrAppend(ret, _("No players found with such a name"));
+	return ret;
+}
+
 std::vector<TextCmdItem> TextCmdList = {
 	{ N_("/help"), N_("Prints help overview or help for a specific command."), N_("({command})"), &TextCmdHelp },
-	{ N_("/arena"), N_("Enter a PvP Arena."), N_("{arena-number}"), &TextCmdArena }
+	{ N_("/arena"), N_("Enter a PvP Arena."), N_("{arena-number}"), &TextCmdArena },
+	{ N_("/inspect"), N_("Inspects stats and equipment."), N_("{player name}"), &TextCmdInspect },
```
obligaron, line 419: should we add a `/stopinspect` command?
Using `/inspect` without parameter to stop inspect wasn't that clear (at least to me).

obligaron (commented):
Could we also add spellbook and belt support to the pr?
It would be nice to see to spells of other players and the items in the belt.
🙂

Source/control.cpp
qndel: that'd be an overkill imo, you can always use it with your own name so 2 ways to stop inspecting already

qndel:
> Could we also add spellbook and belt support to the pr? It would be nice to see to spells of other players and the items in the belt. 🙂

I thought belt was shared already?

Source/cursor.cpp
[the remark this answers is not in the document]
StephenCWills: Well, the stat flag is not being synced at all. It's not even part of the struct we use to sync item data.

https://github.com/diasurgical/devilutionX/blob/2296bf5625d76569405bff716ac929bec6b5ee69/Source/msg.h#L549-L566

I don't think it's an issue that it's not synced, but perhaps we should be calling `item.updateRequiredStatsCacheForPlayer(player)` on remote player's inventories. That said, it does feel a bit weird to me to use `_iStatFlag` on remote player inventories since we could just as easily want to see `item.updateRequiredStatsCacheForPlayer(*MyPlayer)` in some contexts, like what we do for shop menus. At least `InspectPlayer->CanUseItem(item)` is less ambiguous than simply reading `item._iStatFlag`.

obligaron:
> > Could we also add spellbook and belt support to the pr? It would be nice to see to spells of other players and the items in the belt. 🙂
> 
> I thought belt was shared already?

No `DrawBelt` only uses `MyPlayer` and not `InspectPlayer`. 😉

Source/multi.cpp
obligaron: I added the code to `PlayerLeftMsg` and tried it. Seems to work 😄

Source/diablo.cpp
obligaron: I tried to move the check later down the path so it should also work for gamepad and touch controls.
But I couldn't test it. :-/

AJenbo (approved):

PR #4544, opened 2022-05-07
Settings: Add Resampler

glebm:
Makes the resampler algorithm configurable from the settings menu.

Also changes the default resampler back to Speex (configurable via the `DEVILUTIONX_DEFAULT_RESAMPLER` CMake var)


malvarenga123:
Hi @glebm, just want to point out that SoX has a resampler library that many consider to be of very high quality: http://sox.sourceforge.net/SoX/Resampling
Granted, I'm not sure it's usable, as is, in a project like this.

glebm:
SDL_audiolib does have a SoX resampler interface, it is possible to add it as an option in the future.

AJenbo (approved):

PR #4550, opened 2022-05-08
Vanilla bug: Fix Lazarus quest level resetting

galaxyhaxz:
This is a rather fun one: once you receive the Laz quest and go back down to level 15, the quest state is set to 3. When entering Laz's chamber it increases once you pop the seals and again after Laz finishes speaking. No problem right?

Well the devs didn't consider that you can portal out of Laz's chamber; and then walk back down to level 15 again, which sets the quest state back to 3. Going through the red portal would put you on the teleport pad thus fixing the quest state, but if you go back to town and through the original town portal instead, the dungeon will reset back to its initial state.


AJenbo (approved):
Thanks :)

Issue #4301, opened 2022-04-02
Render PCX directly


glebm:
We can reduce RAM usage / allocator pressure by rendering certain PCX graphics directly rather than converting them.

The largest file is `hf_logo1.pcx`. It is a 2.4 MiB PCX file without transparency (4.6 MiB as an SDL surface).

As we do not have to worry about clipping these, the rendering code can be quite simple.

obligaron:
@glebm is this done with #4551? 🤔

AJenbo:
Done and redon

PR #4551, opened 2022-05-08
Direct PCX rendering

glebm:
The backgrounds can be quite large, for example the Hellfire title background is a 2.4 MiB PCX file.

Previously, we converted all background to SDL surfaces.
This required extra memory, e.g. the 2.4 MiB Hellfire title background uses 4.6 MiB as an SDL surface.

Changes the background to render directly from PCX instead to reduce the allocator pressure.

Refs #4301


glebm:
Oof, that's a lot of conflicts

AJenbo:
yikes, sorry about that :(

glebm:
Rebased

Source/engine/load_pcx.hpp
```diff
@@ -0,0 +1,15 @@
...
+#include <cstdint>
+
+#include <SDL.h>
+
+#include "engine/pcx_sprite.hpp"
+#include "utils/stdcompat/optional.hpp"
+
+namespace devilution {
+
+std::optional<OwnedPcxSprite> LoadPcxAsset(const char *path, std::optional<uint8_t> transparentColor = std::nullopt, SDL_Color *outPalette = nullptr);
```
AJenbo, line 12: `LoadCel` isn't called `LoadCelAsset()`, isn't this inconstant with naming this function `LoadPcxAsset`?

glebm: True, though I'd rather rename `LoadCel` to `LoadCelAsset`, to indicate that it's loaded via our assets mechanism.

AJenbo (approved):

Source/DiabloUI/diabloui.cpp
```diff
@@ -865,33 +888,36 @@ void Render(const UiList *uiList)
 
 void Render(const UiScrollbar *uiSb)
 {
+	const Surface out = Surface(DiabloUiSurface());
```
AJenbo, line 891: @glebm was it intentional that this one is a copy while all other usages are references?

`const Surface &out = Surface(DiabloUiSurface());`

glebm: It's an unowned surface, the other ones only work thanks to reference lifetime extension and should probably also be non-references (though it doesn't make a difference)

PR #4552, opened 2022-05-08
Overhaul translation fetching

glebm:
1. Do not modify the map after loading. Instead, return string views (guaranteed to be null-terminated here) from look up functions and return the key directly if not found.
2. Use an `unordered_map` instead of `map` where heterogenous lookup is available (C++20). Saves a bit of RAM (~50 KiB) and improves lookup performance.
3. Introduces a `StringOrView` type which allows us to avoid unnecessary copying of translation strings in some cases.

Heterogenous `unordered_map` requires bumping `CXX_STANDARD` to 20, which in turn requires wrapping non-constexpr `fmt::format` calls with `fmt::runtime`.


AJenbo (approved):

glebm:
Needs quite a bit more work to:
1. Ensure heterogenous lookup (`string_view` key).
2. Avoid inserting unknown keys into the map.

glebm:
Rebased as there were some conflicts

AJenbo:
```
C:\projects\devilutionx\Source\objects.cpp(3936): message : see reference to function template instantiation 'const T devilution::PickRandomlyAmong<devilution::ItemType>(const std::initializer_list<devilution::ItemType> &)' being compiled [C:\projects\devilutionx\Source\libdevilutionx.vcxproj]
          with
          [
              T=devilution::ItemType
          ]
  options.cpp

```

glebm:
amigaos-m68k

```
/root/repo/Source/diablo.cpp:1792:5: internal compiler error: Segmentation fault
 int DiabloMain(int argc, char **argv)
```

uhh

glebm:
Fixed the MSVC error. amiga compiler segfault remains a mystery

glebm:
Renaming `log_internal::SafeFormat` back to `detail::format` fixed the amiga build. Oh well 🤷

AJenbo:
> amigaos-m68k
> 
> ```
> /root/repo/Source/diablo.cpp:1792:5: internal compiler error: Segmentation fault
>  int DiabloMain(int argc, char **argv)
> ```
> 
> uhh


was having the same issue withe a fairly trivial refactor of a loop the other day. i guess something ubrelated has gotten the compiler in to an unstable place

AJenbo (commented):
The other changes now build correctly so maybe we are on stable grounds again.

glebm:
Rebased

AJenbo (approved):

PR #4562, opened 2022-05-10
Clean up use of currlevel

AJenbo:
Hellfire set the wrong spawn position when entering crypt level 1:
image

This is fixed by this PR

### DIFF
--- a/Source/drlg_l1.cpp
+++ b/Source/drlg_l1.cpp
@@ -759,7 +759,7 @@ void ApplyShadowsPatterns()
 	}
 }
 
-int PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool setview, int noquad)
+bool PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool setview, int noquad)
 {
 	int sx;
 	int sy;
@@ -830,7 +830,7 @@ int PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool s
 					}
 				}
 				if (++found > 4000)
-					return -1;
+					return false;
 			}
 		}
 
@@ -859,14 +859,7 @@ int PlaceMiniSet(const BYTE *miniset, int tmin, int tmax, int cx, int cy, bool s
 		ViewPosition = Point { 19, 20 } + Displacement { sx, sy } * 2;
 	}
 
-	if (sx < cx && sy < cy)
-		return 0;
-	if (sx > cx && sy < cy)
-		return 1;
-	if (sx < cx && sy > cy)
-		return 2;
-
-	return 3;
+	return true;
 }
 
 void PlaceMiniSetRandom(const BYTE *miniset, int rndper)
@@ -1988,32 +1981,32 @@ void FixTransparency()
 
 void FixDirtTiles()
 {
-	if (currlevel < 21) {
-		for (int j = 0; j < DMAXY - 1; j++) {
-			for (int i = 0; i < DMAXX - 1; i++) {
-				if (dungeon[i][j] == 21 && dungeon[i + 1][j] != 19) {
-					dungeon[i][j] = 202;
-				}
-				if (dungeon[i][j] == 19 && dungeon[i + 1][j] != 19) {
-					dungeon[i][j] = 200;
-				}
-				if (dungeon[i][j] == 24 && dungeon[i + 1][j] != 19) {
-					dungeon[i][j] = 205;
-				}
-				if (dungeon[i][j] == 18 && dungeon[i][j + 1] != 18) {
-					dungeon[i][j] = 199;
-				}
-				if (dungeon[i][j] == 21 && dungeon[i][j + 1] != 18) {
-					dungeon[i][j] = 202;
-				}
-				if (dungeon[i][j] == 23 && dungeon[i][j + 1] != 18) {
-					dungeon[i][j] = 204;
-				}
+	for (int j = 0; j < DMAXY - 1; j++) {
+		for (int i = 0; i < DMAXX - 1; i++) {
+			if (dungeon[i][j] == 21 && dungeon[i + 1][j] != 19) {
+				dungeon[i][j] = 202;
+			}
+			if (dungeon[i][j] == 19 && dungeon[i + 1][j] != 19) {
+				dungeon[i][j] = 200;
+			}
+			if (dungeon[i][j] == 24 && dungeon[i + 1][j] != 19) {
+				dungeon[i][j] = 205;
+			}
+			if (dungeon[i][j] == 18 && dungeon[i][j + 1] != 18) {
+				dungeon[i][j] = 199;
+			}
+			if (dungeon[i][j] == 21 && dungeon[i][j + 1] != 18) {
+				dungeon[i][j] = 202;
+			}
+			if (dungeon[i][j] == 23 && dungeon[i][j + 1] != 18) {
+				dungeon[i][j] = 204;
 			}
 		}
-		return;
 	}
+}
 
+void FixCryptDirtTiles()
+{
 	for (int j = 0; j < DMAXY - 1; j++) {
 		for (int i = 0; i < DMAXX - 1; i++) {
 			if (dungeon[i][j] == 19)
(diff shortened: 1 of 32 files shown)


Source/drlg_l1.cpp
```diff
@@ -2168,6 +2168,130 @@ void CryptPatternGroup7(int rndper)
...
+			ViewPosition.y++;
+		} else {
+			if (PlaceMiniSet(L5STAIRSUPHF, 1, 1, 0, 0, true, -1) < 0)
+				doneflag = true;
+			if (currlevel != 24 && PlaceMiniSet(L5STAIRSDOWN, 1, 1, 0, 0, false, -1) < 0) {
+				doneflag = true;
+			}
+			ViewPosition.y++;
+		}
+	} else if (entry == ENTRY_PREV) {
```
AJenbo, line 2253: Previously this might actually have produced incorrect Crypt layouts for games that where converted from D1 to Hellfire (DevilutionX only). I see this as a self-bugfix despite being a possible break in comparability :shrug:

Source/drlg_l1.cpp
```diff
@@ -2168,6 +2168,130 @@ void CryptPatternGroup7(int rndper)
...
+		if (Players[MyPlayerId].pOriginalCathedral) {
+			if (PlaceMiniSet(L5STAIRSUP, 1, 1, 0, 0, true, -1) < 0)
+				return true;
+			if (PlaceMiniSet(STAIRSDOWN, 1, 1, 0, 0, false, -1) < 0)
+				return true;
+		} else {
+			if (PlaceMiniSet(STAIRSUP, 1, 1, 0, 0, true, -1) < 0)
+				doneflag = true;
+			if (PlaceMiniSet(STAIRSDOWN, 1, 1, 0, 0, false, -1) < 0)
+				doneflag = true;
```
ephphatha, line 2208: When `PlaceMiniSet` is called with `tmax - tmin == 0` it calls the global rng exactly twice. This could use an early return like:
```suggestion
			if (PlaceMiniSet(STAIRSUP, 1, 1, 0, 0, true, -1) < 0) {
				// Skip trying to place the down stairs since we've already failed to generate the level, but we still need to advance the RNG state twice to match vanilla dungeon generation.
				AdvanceRndSeed();
				AdvanceRndSeed();
				return true;
			}
			if (PlaceMiniSet(STAIRSDOWN, 1, 1, 0, 0, false, -1) < 0)
				return true;
```

ephphatha (commented):
Would be good to get another set of eyes over the dungeon generation code. Breaking up those top level if statements into separate functions is good and looks like it hasn't changed anything from a read through, I just can't be confident I haven't missed anything 😅

Source/drlg_l1.cpp
```diff
@@ -2168,6 +2168,130 @@ void CryptPatternGroup7(int rndper)
...
+			if (PlaceMiniSet(STAIRSUP, 1, 1, 0, 0, true, -1) < 0)
+				return true;
+		} else {
+			if (PlaceMiniSet(STAIRSUP, 1, 1, 0, 0, false, -1) < 0)
+				doneflag = true;
+			if (entry == ENTRY_PREV) {
+				ViewPosition = Point { 20, 28 } + Displacement { setpc_x, setpc_y } * 2;
+			} else {
+				ViewPosition.y--;
+			}
```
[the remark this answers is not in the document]
AJenbo: Yeah, I decided to wait with it til later not to risk breaking anything, and hopefully have written some unit tests first

Source/drlg_l1.cpp
```diff
@@ -2199,110 +2322,7 @@ void GenerateLevel(lvl_entry entry)
...
-				if (PlaceMiniSet(L5STAIRSUPHF, 1, 1, 0, 0, true, -1) < 0)
-					doneflag = false;
-				if (currlevel != 24) {
-					if (PlaceMiniSet(L5STAIRSDOWN, 1, 1, 0, 0, false, -1) < 0)
-						doneflag = false;
-				}
-			}
-		}
-	} while (!doneflag);
+	} while (PlaceMiniSets(entry));
```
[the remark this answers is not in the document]
AJenbo: I had actually misunderstood this loop as continually adding to the level until there was no more room for set piece. That also makes refactoring it in to doing early returns much safer, thanks for figuring that out :+1:

Source/drlg_l1.cpp
AJenbo: There are also two entrances that are programmed incorrectly in the original Hellfire, one gets overwritten, the other case can never trigger, so there was never any real consequence.

AJenbo:
@ephphatha thanks for pointing out the intention of the `GenerateLevel()` loop. I think `Place*MiniSets()` actually turned out to be sensible now :)

Source/drlg_l4.cpp
```diff
@@ -1283,11 +1283,49 @@ void GeneralFix()
 	}
 }
 
+bool PlaceMiniSets(lvl_entry entry)
```
ephphatha, line 1286: each version of this function can be called something like `PlaceStairs`, bit more descriptive since they all deal with a specific set of minisets.

AJenbo: Damn, thats exactly what I called it when i started on drlg_l4, but decided to rename it back to `PlaceMiniSets` :D

AJenbo: Fixed and rebased

AJenbo:
Do you have a save for that?

Source/drlg_l1.cpp
```diff
@@ -2168,6 +2161,62 @@ void CryptPatternGroup7(int rndper)
 	PlaceMiniSetRandom(CryptPattern8, rndper);
 }
 
+bool PlaceCathedralStairs(lvl_entry entry)
+{
+	// Place poison water entrance
+	if (Quests[Q_PWATER].IsAvailable()) {
+		if (!PlaceMiniSet(PWATERIN, 1, 1, 0, 0, entry == ENTRY_RTNLVL, -1))
+			return false;
```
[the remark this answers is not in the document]
AJenbo: No worries I will just rewrite it again. Also nice catch :)

Sure would have been nice if they had put extra effort in to writing the level generation code well considering how important it is.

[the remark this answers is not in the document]
AJenbo: Alright, looks like only Cathedral/Crypt was affected. From them I removed the early return, except in the case that we are placing the stairs up in the none-Hellfire variant (this doesn't look as bad as it sounds I think :D ).

Diff from before: https://github.com/diasurgical/devilutionX/compare/7377ab32c8a7a430eede4ae24ec0fdb226fa77a0..8d486cc490910259838158851edb65dad47e4db4

AJenbo:
Ugh, something is wrong about Catacombs as well:

image

AJenbo:
Ok, fixed, but can someone explain to my why this:
```cpp
bool doneflag;
do {
	doneflag = PlaceStairs(entry);
} while (!doneflag);
```

is not the same as this:
```cpp
do {
} while (!PlaceStairs(entry));
```

Source/drlg_l1.cpp
```diff
@@ -2168,6 +2161,69 @@ void CryptPatternGroup7(int rndper)
...
+bool PlaceCathedralStairs(lvl_entry entry)
+{
+	bool success = true;
+
+	// Place poison water entrance
+	if (Quests[Q_PWATER].IsAvailable()) {
+		if (!PlaceMiniSet(PWATERIN, 1, 1, 0, 0, entry == ENTRY_RTNLVL, -1))
+			success = false;
+		if (entry == ENTRY_RTNLVL)
+			ViewPosition.y--;
```
AJenbo, line 2173: The original actually sets the wrong position for quest level entrances. Instead `GetReturnLvlPos()` will overwrite it with the position where the player was when triggering the level translation.

We could get rid of one of them. Any thoughts?

Source/drlg_l2.cpp
```diff
@@ -2880,35 +2905,8 @@ void GenerateLevel(lvl_entry entry)
...
-				doneflag = PlaceMiniSet(DSTAIRS, 1, 1, -1, -1, false);
-				if (doneflag && currlevel == 5) {
-					doneflag = PlaceMiniSet(WARPSTAIRS, 1, 1, -1, -1, true);
-				}
-			}
-			ViewPosition.y -= 2;
-		}
-	}
+		doneflag = PlaceStairs(entry);
+	} while (!doneflag);
```
ephphatha, line 2909: You could even do this as `while (true) { if (PlaceStairs()) { break; } }` :D

would remove the need for `doneFlag`

qndel: why not for(;;)
:D

ephphatha (commented):
Ran through each level type and didn't spot any differences between master/this branch. The code is much nicer to read as well. Consider this approval, just wanted to give obligaron/qndel a chance to check before auto-merge kicks in.

Source/drlg_l2.cpp
AJenbo: @ephphatha done, I also flattened the one for caves.

Source/drlg_l1.cpp
AJenbo: I have corrected it to give the correct position for PosionedWater, but I realize that it will also other adjustments for the other quests as most a positioned using .dun sections, except for laz portal.

obligaron (approved):
lgtm 🙂

AJenbo:
Feels good to get to this point. I think we should be in good shape for handling custom Hellfire levels at this point :)

PR #4565, opened 2022-05-11
printInConsole: Add string_view support

glebm:
Also print `\r\n` instead of `\n` on Windows.

This paves the way forward for returning a `string_view` from `_(...)`.

### DIFF
--- a/Source/utils/console.cpp
+++ b/Source/utils/console.cpp
@@ -3,6 +3,7 @@
 #if defined(_WIN64) || defined(_WIN32)
 #include <cstddef>
 #include <cstdio>
+#include <string>
 
 // Suppress definitions of `min` and `max` macros by <windows.h>:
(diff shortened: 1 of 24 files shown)


Source/utils/console.cpp
```diff
@@ -24,34 +25,45 @@ HANDLE GetStderrHandle()
...
+void printInConsole(string_view str)
+{
+	OutputDebugString(std::string(str).c_str());
+	WriteToStderr(str);
+}
+
+void printNewlineInConsole()
+{
+	OutputDebugString("\r\n");
+	WriteToStderr("\r\n");
```
AJenbo, line 47: i guess this might fix the line break issue on windows

AJenbo (approved):
I dont think translations are actually working for the console output so we cold also just drop that part. Feel free to merge

PR #4567, opened 2022-05-12
Add debug command for testing custom levels

AJenbo:
Load a custom level `map path type x y`. Example `map Levels\L3Data\Foulwatr 3 31 83` will load the poison water level with the cave tiles and spawn position of 31x83
image

`map levels\l1data\vile1 6 35 36` will load you in to Lazarus' chamber but using the Crypt theme (partially compatible as it is based on Cathedral)
image

`map levels\l1data\vile1 4 35 36` also just for fun, Lazarus' chamber with Hell tiles
image

`map Levels\L3Data\Foulwatr 5 31 83` Poison water with the Hive tiles:
image
(note that loading a level with the incorrect theme is likely to result in crashes)

`map Levels\L1Data\Lv1MazeB 1 20 50` loads the unused maze level:
image

Be aware that you need to start a new/reload game between loading each level as else the game will try to transfer object and monster states from the previous level you had loaded as the test level.

A handy way to use this is to bind it to one of the quick messages so that you can load your test level with the press of a button.

Note that level palette will be a random one based on the selected tileset and no triggers will be setup.

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -32,6 +32,7 @@
 
 namespace devilution {
 
+std::string TestMapPath;
 std::optional<OwnedCelSprite> pSquareCel;
 bool DebugToggle = false;
 bool DebugGodMode = false;
@@ -218,7 +219,7 @@ std::string DebugCmdWarpToLevel(const string_view parameter)
 	return fmt::format("Welcome to level {}.", level);
 }
 
-std::string DebugCmdLoadMap(const string_view parameter)
+std::string DebugCmdLoadQuestMap(const string_view parameter)
 {
 	if (parameter.empty()) {
 		std::string ret = "What mapid do you want to visit?";
@@ -250,6 +251,47 @@ std::string DebugCmdLoadMap(const string_view parameter)
 	}
 
 	return fmt::format("Mapid {} is not known. Do you want to write a mod?", level);
+}
+
+std::string DebugCmdLoadMap(const string_view parameter)
+{
+	TestMapPath.clear();
+	int mapType = 0;
+	Point spawn = {};
+
+	std::stringstream paramsStream(parameter.data());
+	int count = 0;
+	for (std::string tmp; std::getline(paramsStream, tmp, ' ');) {
+		switch (count) {
+		case 0:
+			TestMapPath = fmt::format("{:s}.dun", tmp);
+			break;
+		case 1:
+			mapType = atoi(tmp.c_str());
+			break;
+		case 2:
+			spawn.x = atoi(tmp.c_str());
+			break;
+		case 3:
+			spawn.y = atoi(tmp.c_str());
+			break;
+		}
+		count++;
+	}
+
+	if (TestMapPath.empty() || mapType < DTYPE_CATHEDRAL || mapType > DTYPE_LAST || !InDungeonBounds(spawn))
+		return "Directions not understood";
+
+	ReturnLvlPosition = ViewPosition;
+	ReturnLevel = currlevel;
+	ReturnLevelType = leveltype;
+
+	setlvltype = static_cast<dungeon_type>(mapType);
+	ViewPosition = spawn;
+
+	StartNewLvl(MyPlayerId, WM_DIABSETLVL, SL_NONE);
+
+	return "Welcome to this unique place.";
 }
 
 std::unordered_map<string_view, _talker_id> TownerShortNameToTownerId = {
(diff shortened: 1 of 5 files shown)


Source/setmaps.cpp
```diff
@@ -5,9 +5,15 @@
  */
 #include "setmaps.h"
 
+#include <fmt/format.h>
```
obligaron, line 8: Why is fmt needed?

Source/debug.cpp
```diff
@@ -252,6 +253,38 @@ std::string DebugCmdLoadMap(const string_view parameter)
...
+		case 3:
+			ViewPosition.y = atoi(tmp.c_str());
+			break;
+		}
+		count++;
+	}
+
+	StartNewLvl(MyPlayerId, WM_DIABSETLVL, SL_NONE);
+
+	return fmt::format("Welcome to {}.", "Some level");
```
obligaron, line 285:
```suggestion
	return "Welcome to this unique place.");
```

Source/debug.cpp
```diff
@@ -252,6 +253,38 @@ std::string DebugCmdLoadMap(const string_view parameter)
...
 }
 
+std::string DebugCmdLoadMap(const string_view parameter)
+{
+	ReturnLvlPosition = ViewPosition;
+	ReturnLevel = currlevel;
+	ReturnLevelType = leveltype;
+
+	std::stringstream paramsStream(parameter.data());
+	std::string name;
```
obligaron, line 263: `name` is not used or?

Source/debug.cpp
```diff
@@ -252,6 +253,38 @@ std::string DebugCmdLoadMap(const string_view parameter)
...
+			ViewPosition.x = atoi(tmp.c_str());
+			break;
+		case 3:
+			ViewPosition.y = atoi(tmp.c_str());
+			break;
+		}
+		count++;
+	}
+
+	StartNewLvl(MyPlayerId, WM_DIABSETLVL, SL_NONE);
```
obligaron, line 283: Should we validate the command and not execute the debug command if the parameters are wrong?
(And don't change the globals (`setlvltype`, `ViewPosition`, ...)?)
For example wrong parameter count, not existing file.
(I know it's only a debug command, only asking 😉)

Source/objects.cpp
```diff
@@ -5195,8 +5180,11 @@ void ObjChangeMap(int x1, int y1, int x2, int y2)
 		ObjL2Special(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
 		AddL2Objs(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
 	}
+	if (leveltype == DTYPE_CAVES) {
+		AddL3Objs(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
+	}
 	if (leveltype == DTYPE_CRYPT) {
-		AddL5Objs(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
+		AddCryptObjects(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
```
obligaron, line 5187: Old `AddL5Objs` and `AddCryptObjects` aren't identically (checks different piece IDs).
Is this no problem, cause the method was never executed or in hellfire the piece's aren't used?
(Perhaps I miss something, I'm no map expert 😉)

Source/objects.cpp
```diff
@@ -5195,8 +5180,11 @@ void ObjChangeMap(int x1, int y1, int x2, int y2)
 		ObjL2Special(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
 		AddL2Objs(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
 	}
+	if (leveltype == DTYPE_CAVES) {
+		AddL3Objs(2 * x1 + 16, 2 * y1 + 16, 2 * x2 + 17, 2 * y2 + 17);
```
obligaron, line 5184: This doesn't change behavior, cause in vanilla caves never uses this function.
Is my assumption correct?

obligaron (commented):
Some small nit's and a bunch of non-expert questions, else looks good to me.
I think this will help a lot when creating custom maps. 👍 
Thanks AJenbo 🙂

Source/setmaps.cpp
AJenbo: Leftover from when implementing it
```suggestion
```

Source/objects.cpp
AJenbo: `AddL5Objs` was a very rudimentary split from `AddL5Objs`, `AddCryptObjects` was a fully "optimized" version. Both where added by me last week when splitting crypt logic from cathedral, but at two different points of the call chain.

Source/objects.cpp
AJenbo: Poisoned water is the only custom vanilla caves map. And it doesn't change it layout. The portion of code is also only relevant if the map changes to add fence doors to an area, not so likely, but good to have just in case :)

Source/debug.cpp
AJenbo: Probably more left over :sweat_smile:

Source/debug.cpp
AJenbo: This would be nice, but it was also where I explicitly decided I wasn't going to put more time in to this command :)

Maybe some one else will :sweat_smile:

AJenbo: Fine! I threw some basic validation at it :)

AJenbo:
Thanks for the review, I think I have addressed the things you pointed out :)

obligaron (approved):
lgtm 🙂

PR #4571, opened 2022-05-13
Clean up object references

AJenbo:
Combined with https://github.com/diasurgical/devilutionX/pull/4152 this leaves us with about 200 uses of `Players` and 100 uses of `Objects`

- [x] Doors are broken when loaded from a save (appears to be wrong animation frame)
- [ ] Test object sync in multiplayer
- [ ] Check if `SyncOpObject(MyPlayerId)` will cause issues
- [x] Torches outside poison water are missing
- [ ] Moving saves back to previous versions (vanilla) laves doors permanently blocking
- [ ] Update function docs
- [ ] check for more const ref opotunities


Source/objects.cpp
```diff
@@ -5255,66 +5265,69 @@ int ItemMiscIdIdx(item_misc_id imiscid)
...
 	case OBJ_L1RDOOR:
 		if (teleFlag) {
-			if (Objects[i]._otype == OBJ_L1LDOOR)
-				OperateL1LDoor(pnum, i, true);
-			if (Objects[i]._otype == OBJ_L1RDOOR)
-				OperateL1RDoor(pnum, i, true);
+			if (object._otype == OBJ_L1LDOOR)
+				OperateL1LDoor(&player, i, true);
+			if (object._otype == OBJ_L1RDOOR)
+				OperateL1RDoor(&player, i, true);
```
ephphatha, line 5279: These calls could use `sendflag` as declared a few lines up instead of passing through the player reference.

ephphatha (commented):
Gonna split this because github keeps forcing me to refresh -_-

AJenbo:
> Gonna split this because github keeps forcing me to refresh -_-

Yikes, I was hoping I made it before you started reviewing :cold_sweat: Sorry for that :) I'm done now by the way :angel:

Source/items.cpp
```diff
@@ -1623,21 +1623,23 @@ void SpawnRock()
...
 		return;
 
-	int oi;
-	bool ostand = false;
-	for (int i = 0; i < ActiveObjectCount && !ostand; i++) {
-		oi = ActiveObjects[i];
-		ostand = Objects[oi]._otype == OBJ_STAND;
+	Object *stand = nullptr;
+	for (int i = 0; i < ActiveObjectCount; i++) {
+		Object &object = Objects[ActiveObjects[i]];
```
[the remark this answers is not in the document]
AJenbo: Thanks, I didn't spend a lot of time evaluating if that was the case or not :+1:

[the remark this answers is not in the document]
AJenbo: Marked the two you pointed out here. I'll go over more of them when my brain recovers.

Source/objects.cpp
```diff
@@ -3654,163 +3624,167 @@ void OperateShrine(int pnum, int i, _sfx_id sType)
...
-	CalcPlrInv(Players[pnum], true);
+	CalcPlrInv(player, true);
 	force_redraw = 255;
 
 	if (pnum == MyPlayerId)
 		NetSendCmdParam2(false, CMD_PLROPOBJ, pnum, i);
 }
 
-void OperateSkelBook(int pnum, int i, bool sendmsg)
+void OperateSkelBook(Player &player, int i, bool sendmsg)
```
ephphatha, line 3747: These operate bookcase functions don't use the player instance beyond checking if a network message should be sent to say the object was interacted with. I haven't looked at the usage to see if the existing `sendmsg` parameter is appropriate, for these cases it's probably worth leaving the argument as `pnum` or making it a new bool instead of a player reference.

Source/objects.cpp
```diff
@@ -5220,26 +5206,29 @@ void ObjChangeMapResync(int x1, int y1, int x2, int y2)
...
 		return;
 	}
-	int trapdisper = 2 * Players[pnum]._pDexterity - 5 * currlevel;
+	int trapdisper = 2 * player._pDexterity - 5 * currlevel;
 	if (GenerateRnd(100) > trapdisper) {
 		return;
 	}
 	for (int j = 0; j < ActiveObjectCount; j++) {
 		Object &trap = Objects[ActiveObjects[j]];
 		if (trap.IsTrap() && dObject[trap._oVar1][trap._oVar2] - 1 == i) {
```
ephphatha, line 5225:
```suggestion
		if (trap.IsTrap() && ObjectAtPosition({trap._oVar1, trap._oVar2}) == &target) {
```

Would let you take an object reference in this function.

Source/objects.cpp
```diff
@@ -5220,26 +5206,29 @@ void ObjChangeMapResync(int x1, int y1, int x2, int y2)
 	}
 }
 
-void TryDisarm(int pnum, int i)
+void TryDisarm(Player &player, int i)
```
ephphatha, line 5209:
```suggestion
void TryDisarm(const Player &player, int i)
```
(the object reference would need to be non-const)

Source/player.cpp
```diff
@@ -1016,7 +996,7 @@ bool PlrHitMonst(int pnum, int m, bool adjacentDamage = false)
 	return true;
 }
 
-bool PlrHitPlr(int pnum, int8_t p)
+bool PlrHitPlr(Player &attacker, int8_t p)
```
ephphatha, line 999: Unlike missile attacks this method allows for lifesteal effects and modifies the attacker, worth having a comment explaining why a non-const reference is used here.

Source/scrollrt.cpp
```diff
@@ -459,13 +459,13 @@ void DrawMonster(const Surface &out, Point tilePosition, Point targetBufferPosit
 /**
  * @brief Helper for rendering a specific player icon (Mana Shield or Reflect)
  */
-void DrawPlayerIconHelper(const Surface &out, int pnum, missile_graphic_id missileGraphicId, Point position, bool lighting)
+void DrawPlayerIconHelper(const Surface &out, Player &player, missile_graphic_id missileGraphicId, Point position, bool lighting)
```
ephphatha, line 462: Another case where the player reference isn't used, though this does make the calling function signature nicer. Maybe have a helper like the following so this function can take a bool?

```cpp
bool Player::isLocalPlayer() const { return this == MyPlayer; }
```

Source/scrollrt.cpp
```diff
@@ -521,7 +517,7 @@ void DrawPlayer(const Surface &out, int pnum, Point tilePosition, Point targetBu
 	}
 
 	if (!sprite) {
-		Log("Drawing player {} \"{}\": NULL CelSprite", pnum, player._pName);
+		Log("Drawing player \"{}\": NULL CelSprite", player._pName);
```
ephphatha, line 520: For this error message the player index could potentially still be useful? `std::distance(Players, &player)` could work as a substitute for `pnum` so it's still easy to identify which player had dodgy sprite data.

AJenbo: Any reason not to use std::distance when ever we need the ID for a network messages or save state?

ephphatha: For objects and items 99% of the time they're referenced based on position so the concept of a network id isn't really relevant. std::distance would work for the cases where the code currently uses the array index, I'd rather completely change over to position based lookups instead.

Source/objects.cpp
AJenbo: Looks like this is a safe assumption, it can only be called from ParseCmd where `sendmsg` isn't clearly the same condition. And that should afaik always be coming from a different player id.

Source/objects.cpp
```diff
@@ -5120,50 +4450,10 @@ void RedoPlayerVision()
...
-				int dpy = abs(Objects[oi].position.y - my);
-				if (dpx == 1 && dpy <= 1 && Objects[oi]._otype == OBJ_L5LDOOR)
-					OperateL5LDoor(MyPlayerId, oi, true);
-				if (dpx <= 1 && dpy == 1 && Objects[oi]._otype == OBJ_L5RDOOR)
-					OperateL5RDoor(MyPlayerId, oi, true);
-			}
+	for (Point tile : PointsInRectangleRange { Rectangle { monster.position.tile, 1 } }) {
+		Object *object = ObjectAtPosition(tile);
+		if (object != nullptr && object->IsDoor() && object->_oVar4 == DOOR_CLOSED) {
+			OperateDoor(std::distance(Objects, object), true);
```
ephphatha, line 4456:
```suggestion
			OperateDoor(object.getId(), true);
```

This is a nit more than anything, mainly so that we can use the uses of that method to tell where to update the network code later. (I'd like to change the commands to use `NetSendCmd2(CMD_OPENDOOR, door.position.x, door.position.y)` etc for objects since the layout should be identical :D)

ephphatha (commented):
Looks great, sorry it took a few days to get through it. I need to find some time to actually play through with these changes but I like the refactoring. there are a few things I noticed that I'll save for a followup, these comments are mainly about functions that were changed in this PR that could potentially be cleaned up slightly.

Source/objects.cpp
AJenbo: Sounds good

Source/objects.h
```diff
@@ -17,8 +17,18 @@
...
+
 #define MAXOBJECTS 127
 
+enum {
+	// clang-format off
+	DOOR_CLOSED  =  0,
+	DOOR_OPEN    =  1,
+	DOOR_BLOCKED =  2,
+	// clang-format on
+};
```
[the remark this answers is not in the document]
AJenbo: Yep 🙂

AJenbo:
No worries, I didn't have as much time to dedicate to it myself it turned out. Won't be at a computer until Friday actually.

AJenbo:
Replaced by https://github.com/diasurgical/devilutionX/pull/5169 and all the hard work from @ephphatha @canassa and others.

PR #4572, opened 2022-05-14
Fix condition for set pieces in AddL2Torches()

StephenCWills:
Fixes an issue reported by NickGavran on Discord.

> image
> Is it normal that the halls of the blind can spawn with torches on its walls? Asking about vanilla and about devilx (Screenshot 
> was taken from a random stream I stumbled upon playing Devilx) 
> These torches are not there in the .dun file
> Watching several vanilla playthroughs of this, they do not produce any torches whatsoever 

The condition seems to have been mistakenly flipped by 099fad5.


qndel (approved):

AJenbo:
I was watching the same stream and made a metal note to check it later, nice catch :D

AJenbo (approved):

PR #4573, opened 2022-05-14
Refill smith premium items in a single pass

ephphatha:
Got deadset on implementing this so here we go.

When loading a game from the main menu SmithPremium gets called with premiumlevel == 1 and numpremiums == 0. For a level 50 character in a Hellfire game it ended up running through the item generation code 15 + 3 * 49 times (to initialise the array then rolling 3 new items for every character level beyond the first). This code rolls at most 15 items.

Annoyingly loading a game from a save (i.e. load game, not new game) the premium items get overwritten by whatever was in the save file so this is still being performed unnecessarily in those cases.

~Also looks like there might've been a bug with the way items were generated in old saves? I have a level 7 character which has items of ilvl 5 in the premium items list. The lowest item level for that character level should be 6. This ends up causing more items than expected to get replaced when I buy something after loading that save file. It's 3am though so not gonna debug that more at the moment.~ edit: bitten by the dodgy reroll/level up code used by Hellfire logic, see Stephen's comment for details.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -314,38 +314,6 @@ _sfx_id ItemDropSnds[] = {
 	IS_FLARM,
 	IS_FLARM,
 };
-/** Maps from Griswold premium item number to a quality level delta as added to the base quality level. */
-int premiumlvladd[] = {
-	// clang-format off
-	-1,
-	-1,
-	 0,
-	 0,
-	 1,
-	 2,
-	// clang-format on
-};
-/** Maps from Griswold premium item number to a quality level delta as added to the base quality level. */
-int premiumLvlAddHellfire[] = {
-	// clang-format off
-	-1,
-	-1,
-	-1,
-	 0,
-	 0,
-	 0,
-	 0,
-	 1,
-	 1,
-	 1,
-	 1,
-	 2,
-	 2,
-	 3,
-	 3,
-	// clang-format on
-};
-
 bool IsPrefixValidForItemType(int i, AffixItemType flgs)
 {
 	AffixItemType itemTypes = ItemPrefixes[i].PLIType;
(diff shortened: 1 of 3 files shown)


ephphatha:
Based on #2938 because of the use of std::array for premiumitems, the first two commits are largely obsoleted by the third.

StephenCWills:
> Also looks like there might've been a bug with the way items were generated in old saves? I have a level 7 character which has items of ilvl 5 in the premium items list. The lowest item level for that character level should be 6. This ends up causing more items than expected to get replaced when I buy something after loading that save file. It's 3am though so not gonna debug that more at the moment.

I had noticed this once before. For some reason, it seems the Hellfire devs chose to sacrifice the appropriate spread of ilvls (based on `premiumLvlAddHellfire`) in order to generate exactly three new items at `clvl+1`, `clvl+2`, and `clvl+3` when the player levels up. This mixes up the ilvls in the table a bit depending on what clvl your character is. For better or worse, if you adhere strictly to the `premiumLvlAddHellfire` table or Jarulf's Guide, you won't get the same spread of ilvls as the original algorithm.

You'll notice below that at clvl2, the algorithm ends up with too many ilvl 1 items. As a result, as early as clvl 3, the list includes a stale `clvl-2` item. I've marked these entries with `!`, and all newly generated items with `*`.

_Note, I left the ilvl 0 entries in the table to demonstrate they're being generated at `clvl-1`, even though they get clamped to ilvl 1. Those entries are removed at clvl 2, so they're otherwise irrelevant._

```
    clvl1  clvl2  clvl3  ...
 0      0*     1      1!
 1      0*     1      2
 2      0*     1      2
 3      1*     1!     2
 4      1*     2      2
 5      1*     2      3
 6      1*     2      3
 7      2*     2      3
 8      2*     3      4
 9      2*     3      4
10      2*     3*     4*
11      3*     4      4
12      3*     4*     5*
13      4*     4      5
14      4*     5*     6*
```

ephphatha:
> > Also looks like there might've been a bug with the way items were generated in old saves? I have a level 7 character which has items of ilvl 5 in the premium items list. The lowest item level for that character level should be 6. This ends up causing more items than expected to get replaced when I buy something after loading that save file. It's 3am though so not gonna debug that more at the moment.
> 
> I had noticed this once before. For some reason, it seems the Hellfire devs chose to sacrifice the appropriate spread of ilvls (based on `premiumLvlAddHellfire`) in order to generate exactly three new items at `clvl+1`, `clvl+2`, and `clvl+3` when the player levels up. This mixes up the ilvls in the table a bit depending on what clvl your character is. For better or worse, if you adhere strictly to the `premiumLvlAddHellfire` table or Jarulf's Guide, you won't get the same spread of ilvls as the original algorithm.

I should've remembered that 😂, [it was the whole reason](https://github.com/diasurgical/devilutionX/pull/2924#discussion_r715964267) I opened #2938

Discarding only three items per level up instead of 4 throws it out fairly significantly for characters level 6-31 (though the garbage starts from level 2). Using my clvl7 character as an example you end up with a state of:

> ```
>       clvl1  clvl2  clvl3  clvl4  clvl5  clvl6  clvl7
>  0 -1     1*     1      1!     2!     3!     4!     5!
>  1 -1     1*     1      2      2!     3!     4!     5!
>  2 -1     1*     1      2      3      4      4!     5!
>  3 +0     1*     1!     2!     3!     4!     5!     6!
>  4 +0     1*     2      2!     3!     4!     5!     6!
>  5 +0     1*     2      3      4      4!     5!     6!
>  6 +0     1*     2      3      4      5      6      7
>  7 +1     2*     2!     3!     4!     5!     6!     7!
>  8 +1     2*     3      4      4!     5!     6!     7!
>  9 +1     2*     3      4      5      6      7      8
> 10 +1     2*     3*     4*     5*     6*     7*     8*
> 11 +2     3*     4      4!     5!     6!     7!     8!
> 12 +2     3*     4*     5*     6*     7*     8*     9*
> 13 +3     4*     4!     5!     6!     7!     8!     9!
> 14 +3     4*     5*     6*     7*     8*     9*    10*
> ```

After level 6 only five items are at the correct level for their position, the three that were newly created and two that end up being shifted to an appropriate location by chance.

The reason I think this is a bug is because buying an item causes the item at that position to get rerolled with the correct level. So buying out the store without levelling up will eventually bring everything back into line. Also the diablo logic correctly handles level ups (though it is much simpler) so items will always be of the appropriate level.

Source/items.cpp
```diff
@@ -4074,35 +4042,80 @@ void SpawnSmith(int lvl)
...
-		if (gbIsHellfire) {
-			// Discard first 3 items and shift next 10
-			std::move(&premiumitems[3], &premiumitems[12] + 1, &premiumitems[0]);
-			SpawnOnePremium(premiumitems[10], premiumlevel + premiumLvlAddHellfire[10], pnum);
-			premiumitems[11] = premiumitems[13];
-			SpawnOnePremium(premiumitems[12], premiumlevel + premiumLvlAddHellfire[12], pnum);
-			premiumitems[13] = premiumitems[14];
-			SpawnOnePremium(premiumitems[14], premiumlevel + premiumLvlAddHellfire[14], pnum);
+
+		if (!premiumItem->isEmpty() && (premiumItem->_iCreateInfo & CF_LEVEL) == targetLevel) {
```
ephphatha, line 4102: At high character levels (31-50) the old logic would discard three items on level up to make the store offer a bit of new stock. This change only replaces items in response to purchases at those levels. The old behaviour could be restored by using the following:

```suggestion
		if (nextReusableItem != premiumEnd && !premiumItem->isEmpty() && (premiumItem->_iCreateInfo & CF_LEVEL) == targetLevel) {
```

ephphatha:
Another approach could be to change the offsets used for Hellfire from `[-1, -1, -1, 0, 0, 0, 0, 1, 1, 1, 1, 2, 2, 3, 3]` to `[-2, -2, -2, -1, -1, -1, 0, 0, 0, 1, 1, 1, 2, 2, 3]`. This would make the refilling a purchased item code match the current behaviour on level up.

PR #4574, opened 2022-05-14
slide belt items down when consumed

NickTholin:
If multiple items of the same type are on the belt and one is consumed, then the one further down the belt is used.  This change is in tandem with the AutoRefillBelt from inventory feature.  It's purpose is to reduce the amount of hunting for the correct hotkey on the belt so that items can be used hitting the same key.  It may help address the shock of no more potions available 
https://github.com/diasurgical/devilutionX/pull/983#issuecomment-765237425
Consider the case where the belt is full of full healing potion and the inventory is now out.  Player uses belt slot1, causing slot 8 to be used. Player can notice belt depleting, but can keep finger near the same hotkey, instead of hitting slot1 only to notice too late that it's empty.

beltslide2
beltslide3

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -2045,6 +2045,20 @@ bool UseInvItem(int pnum, int cii)
 				break;
 			}
 		}
+
+		// If speedlist item is not inventory, use same item at the end of the speedlist if exists.
+		if (speedlist && *sgOptions.Gameplay.autoRefillBelt) {
+			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > c; i--) {
+				Item &candidate = player.SpdList[i];
+
+				if (!candidate.isEmpty() && candidate._iMiscId == item->_iMiscId && candidate._iSpell == item->_iSpell) {
+					c = i;
+					item = &candidate;
+					break;
+				}
+			}
+		}
+
 	}
 
 	constexpr int SpeechDelay = 10;


Source/inv.cpp
```diff
@@ -2045,6 +2045,20 @@ bool UseInvItem(int pnum, int cii)
 				break;
 			}
 		}
+
+		// If speedlist item is not inventory, use same item at the end of the speedlist if exists.
+		if (speedlist) {
+			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > cii - INVITEM_BELT_FIRST && *sgOptions.Gameplay.autoRefillBelt; i--) {
+				Item candidate = player.SpdList[i];
+
+				if (candidate._iMiscId == item->_iMiscId && candidate._iSpell == item->_iSpell && candidate._itype != ItemType::None) {
```
ephphatha, line 2054:
```suggestion
				if (!candidate.isEmpty() && candidate._iMiscId == item->_iMiscId && candidate._iSpell == item->_iSpell) {
```

The helper makes it a bit clearer at a glance what the purpose of that check is.

Source/inv.cpp
```diff
@@ -2045,6 +2045,20 @@ bool UseInvItem(int pnum, int cii)
 				break;
 			}
 		}
+
+		// If speedlist item is not inventory, use same item at the end of the speedlist if exists.
+		if (speedlist) {
+			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > cii - INVITEM_BELT_FIRST && *sgOptions.Gameplay.autoRefillBelt; i--) {
```
ephphatha, line 2051: The auto-refill condition doesn't really belong in the for loop conditions in either loop. In this case you've already got a wrapping if statement that is a better place for the check, it could even be moved further up and guard both loops.
```suggestion
		if (speedlist && *sgOptions.Gameplay.autoRefillBelt) {
			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > cii - INVITEM_BELT_FIRST; i--) {
```

Pretty sure you could also handle the case where someone has a belt full of health pots and uses hotkey 3. Currently this will not use the potions in slot 1 and 2, you could run through the entire belt range (in reverse order like you're currently doing) skipping the current index.
```suggestion
		if (speedlist && *sgOptions.Gameplay.autoRefillBelt) {
			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i >= 0; i--) {
				if (i == c) { continue; }
```

Source/inv.cpp
```diff
@@ -2045,6 +2045,20 @@ bool UseInvItem(int pnum, int cii)
...
 		}
+
+		// If speedlist item is not inventory, use same item at the end of the speedlist if exists.
+		if (speedlist) {
+			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > cii - INVITEM_BELT_FIRST && *sgOptions.Gameplay.autoRefillBelt; i--) {
+				Item candidate = player.SpdList[i];
+
+				if (candidate._iMiscId == item->_iMiscId && candidate._iSpell == item->_iSpell && candidate._itype != ItemType::None) {
+					c = i;
+					item = &player.SpdList[c];
```
ephphatha, line 2056:
```suggestion
					item = &candidate;
```

No need to use the array index again here.

Source/inv.cpp
qndel, line 2051:
```suggestion
			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > c && *sgOptions.Gameplay.autoRefillBelt; i--) {
```

Source/inv.cpp
```diff
@@ -2045,6 +2045,20 @@ bool UseInvItem(int pnum, int cii)
 				break;
 			}
 		}
+
+		// If speedlist item is not inventory, use same item at the end of the speedlist if exists.
+		if (speedlist) {
+			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > cii - INVITEM_BELT_FIRST && *sgOptions.Gameplay.autoRefillBelt; i--) {
+				Item candidate = player.SpdList[i];
```
qndel, line 2052:
```suggestion
				Item& candidate = player.SpdList[i];
```
no need to copy

Source/inv.cpp
NickTholin: Thanks, I agree that `autoRefillBelt` check shouldn't be in loop.  I moved it but must not have pushed before the PR.

I didn't loop around to the beginning of the belt intentionally.  My thought process was that this feature would give illusion that belt items will slide down to auto fill the slot used.  I think if someone used a similar item further in the belt that it would be intentional and may not want the items to "slide" to the right.  What's your thought on this?

Source/inv.cpp
NickTholin: Nice.  I thought that checking the item type was a bit clunky.  I looked for another property to try and didn't see one, I should have looked for functions.

Source/inv.cpp
ephphatha: eh, I'm not fussed either way. personally I'd prefer using the item I'm mashing last but I wouldn't go to the effort of trying to make belt items drain to the centre for visual effect. I'm sure others would prefer the implementation you currently have.

Source/inv.cpp
```diff
@@ -2045,6 +2045,20 @@ bool UseInvItem(int pnum, int cii)
 				break;
 			}
 		}
+
+		// If speedlist item is not inventory, use same item at the end of the speedlist if exists.
+		if (speedlist && *sgOptions.Gameplay.autoRefillBelt) {
+			for (int i = INVITEM_BELT_LAST - INVITEM_BELT_FIRST; i > c; i--) {
+				Item& candidate = player.SpdList[i];
```
qndel, line 2052:
```suggestion
				Item &candidate = player.SpdList[i];
```
clang format :D

AJenbo (approved):

Apocalypse612:
Very nice. I've been wanting this!

PR #4578, opened 2022-05-15
Add context to "gold" strings

### LINKED ISSUE #4009: Localization String overlap/omission: item prefix for Gold
The project only has localization string 417 for the word Gold, and it's used in many contexes.

It is also used for the prefix for "golden" which is a different string for greek and other languages with gender forms.
So, once a "ring of gold" dropped in our QA games this omission became apparent.

Please add prefix string "Gold" for items, (tagged as a prefix like others are in it's group)

ephphatha:
fixes #4009 

Went with the context strings mentioned in the issue as I wasn't sure it was worth setting context for all items and item prefixes. So this is only enough to split the two uses of gold plus keep the itemdat table in alignment.


AJenbo (changes requested):
This won't actually work afaik, it will present the strings for translators, but the actual translation part is unchanged. Also if you make the proper change with only some in the list being presented for contextual translation will result in the rest not being translated.

PR #4580, opened 2022-05-15
Floating numbers

qndel:
Side effect of this PR: Fixed using ingame colors in menu - that's why I've replaced 3 uses of Whitegold in menu with Orange16 - earlier Whitegold was actually orange in menu because of a different palette.

defilerfight
magmafight

image
image
image

Thx @StephenCWills and @obligaron  for help :)


Source/qol/floatingnumbers.cpp
```diff
@@ -0,0 +1,176 @@
...
+		Point endPos = pos + floatingNum.endOffset * mul;
+		if (endPos.x < 0 || endPos.x >= gnScreenWidth || endPos.y < 0 || endPos.y >= gnScreenHeight)
+			continue;
+		constexpr UiFlags allSizes = UiFlags::FontSize12 | UiFlags::FontSize24 | UiFlags::FontSize30 | UiFlags::FontSize42 | UiFlags::FontSize46;
+		UiFlags shadowStyle = (floatingNum.style & allSizes) | UiFlags::ColorBlack;
+		DrawString(out, floatingNum.text, endPos + Displacement { -1, -1 }, shadowStyle);
+		DrawString(out, floatingNum.text, endPos + Displacement { 1, 1 }, shadowStyle);
+		DrawString(out, floatingNum.text, endPos + Displacement { -1, 1 }, shadowStyle);
+		DrawString(out, floatingNum.text, endPos + Displacement { 1, -1 }, shadowStyle);
+		DrawString(out, floatingNum.text, endPos, floatingNum.style);
```
glebm, line 143: We use CEL for fonts now and CEL has a handy `CelBlitOutlineTo` method that is much faster than rendering 4 times. Let's add outline support to `DrawString` using `CelBlitOutlineTo`?

Source/qol/floatingnumbers.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+	value >>= 6;
+	if (*sgOptions.FloatingNumbers.bigThreshold != -1 && value >= *sgOptions.FloatingNumbers.bigThreshold)
+		return UiFlags::FontSize30;
+	if (*sgOptions.FloatingNumbers.mediumThreshold != -1 && value >= *sgOptions.FloatingNumbers.mediumThreshold)
+		return UiFlags::FontSize24;
+	return UiFlags::FontSize12;
+}
+
+UiFlags ShiftIndex(int index)
```
glebm, line 55: Perhaps `IndexToFlag`?

Source/qol/floatingnumbers.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+void AddFloatingNumber(bool isMyPlayer, Point pos, FloatingType type, int value, int index, UiFlags style)
+{
+	if (!*sgOptions.FloatingNumbers.enableFloatingNumbers)
+		return;
+	if (!*sgOptions.FloatingNumbers.floatingNumbersFromOthers && !isMyPlayer)
+		return;
+	double distanceX = gnScreenWidth * *sgOptions.FloatingNumbers.maxHorizontalDistance / 100.0;
+	double distanceY = gnScreenHeight * *sgOptions.FloatingNumbers.maxVerticalDistance / 100.0;
+	double angle = (500 + rand() % 500) * 0.0062831853;
```
glebm, line 97: What's 0.0062831853? (please add a comment)

glebm (changes requested):

glebm (dismissed):

qndel:
turns out the "optimization" commit broke things a bit - numbers moved slower if you moved towards them and faster if you moved away from them :D it looked bad lol - position needs to be checked every frame for accurate drawing = reverted "optimization"

StephenCWills (changes requested):

AJenbo:
why so many options :cry: I think it would be best if the colors where the same for all players so that screen shots will make sens :roll_eyes:

PR #4582, opened 2022-05-16
Enable gcov support in codecov uploader

ephphatha:
Doesn't appear to be active by default? Maybe this is why the reports weren't being uploaded


AJenbo:
Looks like you fixed it :+1:

PR #4587, opened 2022-05-17
Remove emoji from cmake output

AMDmi3:
This requires advanced unicode support and compatible fonts, so it's likely that more people see it like garbage than as a library icon
2022-05-17-172622_maim
2022-05-17-172839_maim
2022-05-17-172939_maim


glebm:
What distro doesn't come with emoji out of the box in 2022?

AMDmi3:
I'd guess that any of non-desktop-oriented ones. The screens are from FreeBSD.

AJenbo:
Isn't it possible to install a fallback unicode front on the system?

image

DakkJaniels:
Game should compile with https://github.com/AJenbo/devilutionX/tree/roguelike if desktop isn't available 😄

PR #4588, opened 2022-05-17
Fix UB in random generator

AMDmi3:
`abs(INT_MIN)` is [undefined behavior](https://en.cppreference.com/w/c/numeric/math/abs), so process this case without calling `abs()`

This problem manifestates itself on FreeBSD with clang 11.0 by breaking `QuestTest.SinglePlayerBadPools` and `RandomTest.ShiftModDistributionSignCarry`

### DIFF
--- a/Source/engine/random.cpp
+++ b/Source/engine/random.cpp
@@ -1,5 +1,7 @@
 #include "engine/random.hpp"
 
+#include <limits>
+
 #include "utils/stdcompat/abs.hpp"
 
 namespace devilution {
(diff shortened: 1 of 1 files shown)


Source/engine/random.cpp
```diff
@@ -29,7 +29,8 @@ uint32_t GetLCGEngineState()
 
 int32_t GetRndSeed()
 {
-	return abs(static_cast<int32_t>(sglGameSeed));
+	const int32_t seed = static_cast<int32_t>(sglGameSeed);
+	return seed == -2147483648 ? -2147483648 : abs(seed);
```
ephphatha, line 33:
```suggestion
	return seed == std::numeric_limits<int32_t>::min() ? std::numeric_limits<int32_t>::min() : abs(seed);
```

ephphatha (commented):
Good to see it fails on another system :D, makes the tests worthwhile. Beyond the nit about magic numbers it's good to be more explicit here.

AJenbo (approved):

PR #4590, opened 2022-05-18
Fix link with disabled sound

AMDmi3:
sgnMusicTrack is used from diablo.cpp regardless of NOSOUND value, so provide a constant dummy value

### DIFF
--- a/Source/sound.h
+++ b/Source/sound.h
@@ -51,7 +51,11 @@ struct TSnd {
 };
 
 extern bool gbSndInited;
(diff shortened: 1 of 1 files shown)


glebm:
I've fixed that a bit differently here https://github.com/diasurgical/devilutionX/pull/4544/files#diff-d04367a10138d45ec668a6610d36de57507add955933dc161d22589569e6b890R9, just waiting for that PR to get merged

PR #4591, opened 2022-05-18
Set system data paths on FreeBSD as well

AMDmi3:


AJenbo (dismissed):
Probably other BSDs should go here too

AMDmi3:
Added all BSDs I know.

Source/init.cpp
```diff
@@ -90,7 +90,7 @@ std::vector<std::string> GetMPQSearchPaths()
 	if (paths[0] == paths[1])
 		paths.pop_back();
 
-#if defined(__linux__) && !defined(__ANDROID__)
+#if (defined(__linux__) && !defined(__ANDROID__)) || defined(__FreeBSD__) || defined(__DragonFly__) || defined(__NetBSD__) || defined(__OpenBSD__)
```
ephphatha, line 93: Could this just use `__unix__ && !__ANDROID__`? Looks like all linux and most BSD systems define unix (macos doesn't but that's probably desirable anyway for this case?)

AMDmi3: Makes sense.

AJenbo: Seams good mac has its own scheme for this

dmitryluhtionov:
Сan this be added to the branch 1.4.1 ?

AJenbo:
> Сan this be added to the branch 1.4.1 ?

done

PR #4592, opened 2022-05-18
Removed mention of "Linux" because the game runs on BSD systems as well. Add Russian and Ukrainian tips

dmitryluhtionov:
Removed mention of "Linux" because the game runs on BSD systems as well
Add Russian and Ukrainian tips

PR #4593, opened 2022-05-18
Remove redundant "on Linux" from .desktop file comments

AMDmi3:
Not sure if it's the best solution. The thing is that I use these nix .desktop files for FreeBSD packaging where `on Linux` clause does not apply. I could copy .desktop files for e.g. Packaging/bsd, however IMO it would be better to avoid duplication, and just remove the clause, as even on Linux it's redundant.


qndel:
Duplicate of https://github.com/diasurgical/devilutionX/pull/4592

PR #4600, opened 2022-05-21
MonsterMHit and MonsterTrapHit split 1

k-bar:
This is part 1 of split of #4570
This is to lower complexity of change for review, especially of first original commit there.
In this part I chose only changes around the main function merge. Usually function extracts and convoluted code unwindling.

Please review this first.
Soon part 2 will be available to review - with main change.
Or alternatively, if this would get merge before - I will rebase original PR, to make left changes easier to see.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -251,26 +251,14 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 		monster._mFlags |= MFLAG_NOHEAL;
 
 	if (monster._mhitpoints >> 6 <= 0) {
-		if (monster._mmode == MonsterMode::Petrified) {
-			M_StartKill(m, pnum);
-			monster.Petrify();
-		} else {
-			M_StartKill(m, pnum);
-		}
+		M_StartKill(m, pnum);
+	} else if (resist) {
+		PlayEffect(monster, 1);
 	} else {
-		if (resist) {
-			PlayEffect(monster, 1);
-		} else if (monster._mmode == MonsterMode::Petrified) {
-			if (m > MAX_PLRS - 1)
-				M_StartHit(m, pnum, dam);
-			monster.Petrify();
-		} else {
-			if (MissilesData[t].mType == 0 && HasAnyOf(player._pIFlags, ItemSpecialEffect::Knockback)) {
-				M_GetKnockback(m);
-			}
-			if (m > MAX_PLRS - 1)
-				M_StartHit(m, pnum, dam);
-		}
+		if (monster._mmode != MonsterMode::Petrified && MissilesData[t].mType == 0 && HasAnyOf(player._pIFlags, ItemSpecialEffect::Knockback))
+			M_GetKnockback(m);
+		if (m > MAX_PLRS - 1)
+			M_StartHit(m, pnum, dam);
 	}
 
 	if (monster._msquelch == 0) {
@@ -959,23 +947,12 @@ bool MonsterTrapHit(int m, int mindam, int maxdam, int dist, missile_id t, bool 
 		monster._mhitpoints = 0;
 #endif
 	if (monster._mhitpoints >> 6 <= 0) {
-		if (monster._mmode == MonsterMode::Petrified) {
-			M_StartKill(m, -1);
-			monster.Petrify();
-		} else {
-			M_StartKill(m, -1);
-		}
+		M_StartKill(m, -1);
+	} else if (resist) {
+		PlayEffect(monster, 1);
 	} else {
-		if (resist) {
-			PlayEffect(monster, 1);
-		} else if (monster._mmode == MonsterMode::Petrified) {
-			if (m > MAX_PLRS - 1)
-				M_StartHit(m, -1, dam);
-			monster.Petrify();
-		} else {
-			if (m > MAX_PLRS - 1)
-				M_StartHit(m, -1, dam);
-		}
+		if (m > MAX_PLRS - 1)
+			M_StartHit(m, -1, dam);
 	}
 	return true;
 }
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -212,9 +212,8 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 	if (monster._mmode == MonsterMode::Petrified)
 		hit = 0;
 
-	bool ret = false;
-	if (CheckMonsterHit(monster, &ret))
-		return ret;
+	if (monster.TryLiftGargoyle())
+		return true;
 
 	if (hit >= hper) {
 #ifdef _DEBUG
@@ -938,10 +937,8 @@ bool MonsterTrapHit(int m, int mindam, int maxdam, int dist, missile_id t, bool 
 	int hit = GenerateRnd(100);
 	int hper = 90 - (BYTE)monster.mArmorClass - dist;
 	hper = clamp(hper, 5, 95);
-	bool ret;
-	if (CheckMonsterHit(monster, &ret)) {
-		return ret;
-	}
+	if (monster.TryLiftGargoyle())
+		return true;
 	if (hit >= hper && monster._mmode != MonsterMode::Petrified) {
 #ifdef _DEBUG
 		if (!DebugGodMode)
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -189,33 +189,8 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 {
 	auto &monster = Monsters[m];
 
-	bool resist = false;
-	if (monster.mtalkmsg != TEXT_NONE
-	    || monster._mhitpoints >> 6 <= 0
-	    || (t == MIS_HBOLT && monster.MType->mtype != MT_DIABLO && monster.MData->mMonstClass != MonsterClass::Undead)) {
+	if (!monster.IsPossibleToHit() || monster.IsImmune(t))
 		return false;
-	}
-	if (monster.MType->mtype == MT_ILLWEAV && monster._mgoal == MGOAL_RETREAT)
-		return false;
-	if (monster._mmode == MonsterMode::Charge)
-		return false;
-
-	uint8_t mor = monster.mMagicRes;
-	missile_resistance mir = MissilesData[t].mResist;
-
-	if (((mor & IMMUNE_MAGIC) != 0 && mir == MISR_MAGIC)
-	    || ((mor & IMMUNE_FIRE) != 0 && mir == MISR_FIRE)
-	    || ((mor & IMMUNE_LIGHTNING) != 0 && mir == MISR_LIGHTNING)
-	    || ((mor & IMMUNE_ACID) != 0 && mir == MISR_ACID))
-		return false;
-
-	if (((mor & RESIST_MAGIC) != 0 && mir == MISR_MAGIC)
-	    || ((mor & RESIST_FIRE) != 0 && mir == MISR_FIRE)
-	    || ((mor & RESIST_LIGHTNING) != 0 && mir == MISR_LIGHTNING))
-		resist = true;
-
-	if (gbIsHellfire && t == MIS_HBOLT && (monster.MType->mtype == MT_DIABLO || monster.MType->mtype == MT_BONEDEMN))
-		resist = true;
 
 	int hit = GenerateRnd(100);
 	int hper = 0;
@@ -257,14 +232,14 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 
 	const auto &player = Players[pnum];
 
-	if (MissilesData[t].mType == 0 && mir == MISR_NONE) {
+	if (MissilesData[t].mType == 0 && MissilesData[t].mResist == MISR_NONE) {
 		dam = player._pIBonusDamMod + dam * player._pIBonusDam / 100 + dam;
 		if (player._pClass == HeroClass::Rogue)
 			dam += player._pDamageMod;
 		else
 			dam += player._pDamageMod / 2;
 	}
-
+	bool resist = monster.IsResistant(t);
 	if (!shift)
 		dam <<= 6;
 	if (resist)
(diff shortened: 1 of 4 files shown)


Source/monster.cpp
```diff
@@ -5018,4 +4991,55 @@ bool Monster::IsWalking() const
...
+	    || _mmode == MonsterMode::Charge)
+		return false;
+	if ((MType->mtype >= MT_COUNSLR && MType->mtype <= MT_ADVOCATE) && _mgoal != MGOAL_NORMAL) {
+		GenerateRnd(100); // Vanilla Compatibility Note: Pushing RNG, as original function CheckMonsterHit was called after CTH roll
+		return false;
+	}
+	return true;
+}
+
+bool  Monster::TryLiftGargoyle()
```
AJenbo, line 5035:
```suggestion
bool Monster::TryLiftGargoyle()
```

Source/monster.cpp
```diff
@@ -1128,6 +1128,8 @@ void StartMonsterDeath(int i, int pnum, bool sendmsg)
 	M_FallenFear(monster.position.tile);
 	if ((monster.MType->mtype >= MT_NACID && monster.MType->mtype <= MT_XACID) || monster.MType->mtype == MT_SPIDLORD)
 		AddMissile(monster.position.tile, { 0, 0 }, Direction::South, MIS_ACIDPUD, TARGET_PLAYERS, i, monster._mint + 1, 0);
+	if (monster._mmode == MonsterMode::Petrified)
+		monster.Petrify();
```
AJenbo, line 1132: This won't work as line 1119 always sets the `_mmode = Death`, but instead you could make that line conditional on `_mmode != Petrified`

Source/player.cpp
```diff
@@ -995,22 +984,13 @@ bool PlrHitMonst(int pnum, int m, bool adjacentDamage = false)
...
-			M_StartKill(m, pnum);
-			monster.Petrify();
-		} else {
-			M_StartKill(m, pnum);
-		}
+		M_StartKill(m, pnum);
 	} else {
-		if (monster._mmode == MonsterMode::Petrified) {
-			M_StartHit(m, pnum, dam);
+		if (!(monster._mmode == MonsterMode::Petrified) && HasAnyOf(player._pIFlags, ItemSpecialEffect::Knockback))
```
AJenbo, line 989:
```suggestion
		if (monster._mmode != MonsterMode::Petrified && HasAnyOf(player._pIFlags, ItemSpecialEffect::Knockback))

```

Source/player.cpp
```diff
@@ -995,22 +984,13 @@ bool PlrHitMonst(int pnum, int m, bool adjacentDamage = false)
...
+			M_GetKnockback(m);
+		M_StartHit(m, pnum, dam);
+		if (monster._mmode == MonsterMode::Petrified)
 			monster.Petrify();
```
AJenbo, line 993: It appears that this was never needed as `M_StartHit()` doesn't appear to change `_mmode`.

Source/monster.cpp
```diff
@@ -5018,4 +4991,55 @@ bool Monster::IsWalking() const
...
+
+bool Monster::IsPossibleToHit() const
+{
+	if (_mhitpoints >> 6 <= 0
+	    || mtalkmsg != TEXT_NONE
+	    || (MType->mtype == MT_ILLWEAV && _mgoal == MGOAL_RETREAT)
+	    || _mmode == MonsterMode::Charge)
+		return false;
+	if ((MType->mtype >= MT_COUNSLR && MType->mtype <= MT_ADVOCATE) && _mgoal != MGOAL_NORMAL) {
+		GenerateRnd(100); // Vanilla Compatibility Note: Pushing RNG, as original function CheckMonsterHit was called after CTH roll
```
AJenbo, line 5029: No need to do so outside of DLRG or Item generation algorithms.

Source/missiles.cpp
```diff
@@ -1009,23 +950,14 @@ bool MonsterTrapHit(int m, int mindam, int maxdam, int dist, missile_id t, bool
...
 	} else {
-		if (resist) {
-			PlayEffect(monster, 1);
-		} else if (monster._mmode == MonsterMode::Petrified) {
-			if (m > MAX_PLRS - 1)
-				M_StartHit(m, -1, dam);
+		if (m > MAX_PLRS - 1)
+			M_StartHit(m, -1, dam);
+		if (monster._mmode == MonsterMode::Petrified)
 			monster.Petrify();
```
AJenbo, line 960: This again appears to have had no effect, and the proposal defiantly won't, so can just be removed.

Source/missiles.cpp
```diff
@@ -957,39 +922,15 @@ bool MonsterTrapHit(int m, int mindam, int maxdam, int dist, missile_id t, bool
 {
 	auto &monster = Monsters[m];
 
-	bool resist = false;
-	if (monster.mtalkmsg != TEXT_NONE) {
+	bool resist = monster.IsResistant(t);
```
AJenbo, line 925: Move to 944 since it's not used until then.

Source/player.cpp
k-bar: Actually `M_StartHit()` changes `_mode` to `MonsterMode::HitRecovery` when calling `StartMonsterGotHit`, but it isn't called when monster is alredy Petrified. So you are right :)

Source/missiles.cpp
k-bar: OK. I did't change this function much, because I change it a lot in main PR. In this split this line only hooks to function moved to monster.cpp.

Source/monster.cpp
k-bar: Very good find. Thank you!

Source/monster.cpp
k-bar: Yeah, I know. But I am afraid it can a bit involve those chance to hit dirrerent outcome chains when player is cornering advocate (or even lazarus) and he is constantly trying to escape the corner by fading in and out. 
And I am afraid changing this tiny thing would make a bit different experience when player hits or misses this fading advo.
Leaving it here just assure game works as it did in original. Even if it is not the most intuitive way.
(But we also have firebolt damage recalculated every tick so :man_shrugging: )

AJenbo: it wont make a perceptual difference, please remove it.

Source/monster.cpp
```diff
@@ -1294,28 +1297,23 @@ void MonsterAttackMonster(int i, int mid, int hper, int mind, int maxd)
...
+		if (monster.TryLiftGargoyle())
+			return;
+		if (hit < hper) {
 			int dam = (mind + GenerateRnd(maxd - mind + 1)) << 6;
 			monster._mhitpoints -= dam;
 			if (monster._mhitpoints >> 6 <= 0) {
-				if (monster._mmode == MonsterMode::Petrified) {
-					StartDeathFromMonster(i, mid);
+				StartDeathFromMonster(i, mid);
+				if (monster._mmode == MonsterMode::Petrified)
```
AJenbo, line 1311: This will always fail as `StartDeathFromMonster()` sets `_mmode = Death`. If you want to avoid the duplicate `StartDeathFromMonster()` line then use `bool wasPetrified = monster._mmode == MonsterMode::Petrified;` before, or change `StartDeathFromMonster()` to not change `_mmode` if it was `Petrified`

Source/monster.cpp
```diff
@@ -1294,28 +1297,23 @@ void MonsterAttackMonster(int i, int mid, int hper, int mind, int maxd)
...
 					monster.Petrify();
-				} else {
-					StartDeathFromMonster(i, mid);
-				}
 			} else {
-				if (monster._mmode == MonsterMode::Petrified) {
-					MonsterHitMonster(mid, i, dam);
+				MonsterHitMonster(mid, i, dam);
+				if (monster._mmode == MonsterMode::Petrified)
```
AJenbo, line 1315: `MonsterHitMonster()` won't change `_mmode`, no need for this call.

Source/monster.cpp
```diff
@@ -1116,7 +1116,8 @@ void StartMonsterDeath(int i, int pnum, bool sendmsg)
 
 	Direction md = pnum >= 0 ? GetMonsterDirection(monster) : monster._mdir;
 	NewMonsterAnim(monster, MonsterGraphic::Death, md, gGameLogicStep < GameLogicStep::ProcessMonsters ? AnimationDistributionFlags::ProcessAnimationPending : AnimationDistributionFlags::None);
```
AJenbo, line 1118: looks like `NewMonsterAnim` should also only be called if `_mmode != MonsterMode::Petrified`

Source/monster.cpp
AJenbo: You can remove this now that the mode isn't changed :)

Source/missiles.cpp
```diff
@@ -189,33 +189,9 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 {
 	auto &monster = Monsters[m];
 
-	bool resist = false;
-	if (monster.mtalkmsg != TEXT_NONE
-	    || monster._mhitpoints >> 6 <= 0
-	    || (t == MIS_HBOLT && monster.MType->mtype != MT_DIABLO && monster.MData->mMonstClass != MonsterClass::Undead)) {
+	bool resist = monster.IsResistant(t);
```
AJenbo, line 192: This should also be moved to the first use, which appear to be line 245

AJenbo (commented):
Nice work splitting this out. The chance size here is much easier to follow :+1: It's much clearer how much of a mess `CheckMonsterHit()` was :D

Source/monster.cpp
k-bar: You are probably right.
I didn't want to go that far into monster logic in this PR. 
I noticed similiar places where this logic was confoluted with else ifs and it was previosusly looking like that:
```
//simplified pseudocode

if (petrified)
    mosterDie();
    petrify();
else
    monsterDie();
    
```
this is equivalent to
```
monsterDie();
if petrified
   Petrify(); 
```
the same was for *pseudoMonsterHit* with additions to curly braces.
I will leave the rest of this logic to monster refactoring.
I only refactored obvious cases. When things go deep into animations - it is not yet analyzed well by me.
I will try to fix places where I messed up, and try do not touch it anymore in this PR.

AJenbo: Even better would be to change StartDeathFromMonster() to not call `NewMonsterAnim()` and set `_mmode = MonsterMode::Death` since then you do not need to call `monster.Petrify();` anywhere any longer.

Source/monster.cpp
k-bar: As above. but I am not sure what id does with `AnimInfo.IsPetrified` part of Petrify(); Don't want to go to deep to monster in this PR.

Source/monster.cpp
AJenbo:
```cpp
void monsterDie() {petrified = false; hp = 0;}
```

Thus the two examples are not the same
```cpp
if (petrified) { // This might be true
    monsterDie();
    petrify();
} else {
    monsterDie();
    
}
```

```cpp
monsterDie();
if (petrified) { // This will never be true
   Petrify(); 
}
```

Source/monster.cpp
AJenbo: I looked in to it as part of the reviews, if you just make the remaining suggested changes everything should be good and clean. No more calls to `monster.Petrify()` :)

Source/monster.cpp
k-bar: After I look again at this it loks like this:
Calls to `StartMonsterDeath`:
- `M_StartKill`  - had no petrify check, but there were such checks in function that called it, which I removed.
- `M_SyncStartKill` - had petrify check

seems StartMonsterDeath ALWAYS set `_mode` to `MonsterMode::Death` and calls `	NewMonsterAnim` which sets 
`AnimInfo.IsPetrified` to `true`.

That would mean - that those checks are always redundant. Therefore I remove this condition also for 
```
 if (monster._mmode != MonsterMode::Petrified)
		monster._mmode = MonsterMode::Death;
```
and leave just 
```
monster._mmode = MonsterMode::Death;
```
as it was before.
Is it good for you?

Source/monster.cpp
k-bar: I checked this and agree - I will just remove it it, as `StartDeathFromMonster` is always setting this, so that checks after calling it from `MonsterAttackMonster` are redundant too :| sorry for too long explanation.

Source/monster.cpp
k-bar: Yes, `_mmode` isn't changed. But here I am not sure if that Petrify() sets `AnimInfo.IsPetrified` isn't needed here just in case.
I would like to not to mess with it right now.

Source/missiles.cpp
k-bar: ok, btw, who shall resolve conversation, you or me?

Source/monster.cpp
AJenbo: Considering how close you are to having this all solved in a really nice way i think that would be a little sad :)

AJenbo: > That would mean - that those checks are always redundant. Therefore I remove this condition also for

They where never redundant before this PR as they happened before the call to `StartMonsterDeath`, only after you moved them to after the call did they become ineffective. Which is why you will need to make `NewMonsterAnim()` and `monster._mmode = MonsterMode::Death;` inside `StartMonsterDeath` conditional to not break stone curse.

Source/monster.cpp
```diff
@@ -5027,7 +5022,7 @@ bool Monster::IsPossibleToHit() const
 	    || _mmode == MonsterMode::Charge)
 		return false;
 	if ((MType->mtype >= MT_COUNSLR && MType->mtype <= MT_ADVOCATE) && _mgoal != MGOAL_NORMAL) {
-		GenerateRnd(100); // Vanilla Compatibility Note: Pushing RNG, as original function CheckMonsterHit was called after CTH roll
+		// Vanilla Compatibility Note: put GenerateRnd(100); here, as original function CheckMonsterHit was called after CTH roll
```
AJenbo, line 5025: We don't retain notes like this for the other changes that we have made, having one here could be misleading.

Source/monster.cpp
AJenbo: NewMonsterAnim is the one that changes IsPetrified, so since `MonsterHitMonster` never calls that for petrified monsters there is no need for it here. This is actually the one place where the check was always unnecessary :)

Source/missiles.cpp
AJenbo: Since I brought it up it would probably be me, unless you feel very sure that you have addressed it in a way that I would accept.

Source/monster.cpp
k-bar: Yes you are right, I will try to fix that :) sorry, I am getting experience right now

Source/monster.cpp
k-bar: Yes I was wrong, I fixed that to bool wasPetrified (commit soon) and I call Petrify at the end. I don't want to look at what else NewMonsterAnim() is doing. Seems many things, it is safer for me to keep it out of scope and not look at its internals much.
Is it ok for you?

Source/monster.cpp
k-bar: As previous conv above ^ Finally I was convinced to revemo it as you suggested. (commit soon)

k-bar: > NewMonsterAnim is the one that changes IsPetrified, so since `MonsterHitMonster` never calls that for petrified monsters there is no need for it here. This is actually the one place where the check was always unnecessary :)

Yes, I am probably too much phobic about it. I am afraid there may be some strange case when `_mode` is Petrified, but anim info` AnimInfo.IsPetrified` was not set yet to true.

AJenbo (dismissed):

AJenbo:
Inspired by your change I did some further clean up of MonsterDeath: https://github.com/diasurgical/devilutionX/pull/4601

PR #4601, opened 2022-05-21
Monster death

AJenbo:
- This fixes a bug where Spiderlords would not leave behind an acid puddle when killed by monsters.
- Cleans up redundant logic by moving it to a shared function.
- Don't take short cuts when calculating the proper heading for the death animation

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4999,13 +4999,11 @@ bool Monster::IsResistant(missile_id mName) const
 
 bool Monster::IsPossibleToHit() const
 {
-	if (_mhitpoints >> 6 <= 0
+	return !(_mhitpoints >> 6 <= 0
 	    || mtalkmsg != TEXT_NONE
-	    || MType->mtype == MT_ILLWEAV && _mgoal == MGOAL_RETREAT
+	    || (MType->mtype == MT_ILLWEAV && _mgoal == MGOAL_RETREAT)
 	    || _mmode == MonsterMode::Charge
-	    || MType->mtype >= MT_COUNSLR && MType->mtype <= MT_ADVOCATE && _mgoal != MGOAL_NORMAL)
-		return false;
-	return true;
+	    || (MType->mtype >= MT_COUNSLR && MType->mtype <= MT_ADVOCATE && _mgoal != MGOAL_NORMAL));
 }
 
 bool Monster::TryLiftGargoyle()
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1174,9 +1174,8 @@ void StartDeathFromMonster(int i, int mid)
 	dMonster[monster.position.tile.x][monster.position.tile.y] = mid + 1;
 	CheckQuestKill(monster, true);
 	M_FallenFear(monster.position.tile);
-	if (monster.MType->mtype >= MT_NACID && monster.MType->mtype <= MT_XACID)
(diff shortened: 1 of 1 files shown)


Source/monster.cpp
```diff
@@ -1095,92 +1095,72 @@ void MonsterHitMonster(int mid, int i, int dam)
...
 	assert(monster.MType != nullptr);
 
-	if (pnum >= 0)
-		monster.mWhoHit |= 1 << pnum;
-	if (pnum < MAX_PLRS && i >= MAX_PLRS) /// BUGFIX: i >= MAX_PLRS (fixed)
-		AddPlrMonstExper(monster.mLevel, monster.mExp, monster.mWhoHit);
+	if (pnum < MAX_PLRS) {
+		if (pnum >= 0)
+			monster.mWhoHit |= 1 << pnum;
+		if (pnum < MAX_PLRS && mid >= MAX_PLRS)
```
StephenCWills, line 1107: The outer condition makes this one unnecessary.
```suggestion
		if (mid >= MAX_PLRS)
```

Source/monster.cpp
```diff
@@ -1095,92 +1095,72 @@ void MonsterHitMonster(int mid, int i, int dam)
...
-	monster.position.future = monster.position.old;
-	M_ClearSquares(mid);
-	dMonster[monster.position.tile.x][monster.position.tile.y] = mid + 1;
-	CheckQuestKill(monster, true);
-	M_FallenFear(monster.position.tile);
-	if (monster.MType->mtype >= MT_NACID && monster.MType->mtype <= MT_XACID)
-		AddMissile(monster.position.tile, { 0, 0 }, Direction::South, MIS_ACIDPUD, TARGET_PLAYERS, mid, monster._mint + 1, 0);
-
+	Direction md = GetDirection(monster.position.tile, killer.position.tile);
+	MonsterDeath(mid, i, md, false);
```
StephenCWills, line 1161: Wasn't `sendmsg` true before?
```suggestion
	MonsterDeath(mid, i, md, true);
```

StephenCWills (changes requested):

Source/monster.cpp
AJenbo: My bad, I got to used to it being false from looking at all the Object code last week :sweat_smile:

StephenCWills (approved):

AJenbo:
Thanks :)

Source/monster.cpp
```diff
@@ -1095,92 +1095,72 @@ void MonsterHitMonster(int mid, int i, int dam)
...
 	assert(monster.MType != nullptr);
 
-	if (pnum >= 0)
-		monster.mWhoHit |= 1 << pnum;
-	if (pnum < MAX_PLRS && i >= MAX_PLRS) /// BUGFIX: i >= MAX_PLRS (fixed)
-		AddPlrMonstExper(monster.mLevel, monster.mExp, monster.mWhoHit);
+	if (pnum < MAX_PLRS) {
+		if (pnum >= 0)
+			monster.mWhoHit |= 1 << pnum;
+		if (mid >= MAX_PLRS)
```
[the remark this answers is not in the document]
AJenbo: On no, you missed it by 7 sec :(

I was thinking the same, but decided to do it as a separate clean up in case there where more of them in other places :)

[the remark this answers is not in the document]
obligaron: Yeah I also noticed I was a little too late. 😄

PR #4602, opened 2022-05-21
Apply type checking on monster type

AJenbo:
Follow up to @obligaron 's comments on #4601


Source/missiles.cpp
```diff
@@ -2101,19 +2101,17 @@ void InitMissileAnimationFromMonster(Missile &mis, Direction midir, const Monste
...
+	Monster &monster = Monsters[missile._misource];
 
 	MonsterGraphic graphic = MonsterGraphic::Special;
-	if (monster.MType->mtype < MT_HORNED || monster.MType->mtype > MT_OBLORD) {
-		if (monster.MType->mtype < MT_NSNAKE || monster.MType->mtype > MT_GSNAKE) {
-			graphic = MonsterGraphic::Walk;
-		} else {
-			graphic = MonsterGraphic::Attack;
-		}
+	if (IsAnyOf(monster.MType->mtype, MT_NSNAKE, MT_RSNAKE, MT_BSNAKE, MT_GSNAKE)) {
```
obligaron, line 2107:
```suggestion
	if (IsAnyOf(monster.MType->mtype, MT_NSNAKE, MT_RSNAKE, MT_BSNAKE, MT_GSNAKE)) {
		graphic = MonsterGraphic::Attack;
	} else if (IsNoneOf(monster.MType->mtype, MT_HORNED, MT_MUDRUN, MT_FROSTC, MT_OBLORD)) {
		graphic = MonsterGraphic::Walk;
```

The result is swapped or?
Only snakes use the attack animation. All other types use Walk.

We also could remove the negation:

```cpp
	MonsterGraphic graphic = MonsterGraphic::Walk;
	if (IsAnyOf(monster.MType->mtype, MT_HORNED, MT_MUDRUN, MT_FROSTC, MT_OBLORD)) {
		graphic = MonsterGraphic::Special;
	}
	else if (IsAnyOf(monster.MType->mtype, MT_NSNAKE, MT_RSNAKE, MT_BSNAKE, MT_GSNAKE)) {
		graphic = MonsterGraphic::Attack;
	}
```

obligaron (changes requested):
Found two copy&past errors.
Else it looks much nicer and allows searching for a monster type in the code. 👍

Source/missiles.cpp
AJenbo: Good catch, also much nicer :)

obligaron (approved):
lgtm 🙂

PR #4604, opened 2022-05-22
MonsterMHit and MonsterTrapHit Split 2

k-bar:
Please review this PR - as 2nd part of split of #4570.

This second part of MonsterMhit and MonsterTrapHit refactor I finally merged both function into TryHitMonster using template method pattern.
To do so I used function template to still achieve proper inheritance. 
Player specific code went to `PlayerMissile` class and Trap specific code went to `TrapMissile`.
Both classes are derived from abstract class `Collidable`, which I have plans for in the future :)

It is strongly recommended to view changes commit by commit, because there are many movings of code.

I am also looking forward for permission to merge [already implemented](https://github.com/k-bar/devilutionX/commit/cb92aec7ef0509f375514d063a0c0bf9246ed7d3) solution based on abstract class and pure virtual functions. If not now, then maybe in the future.

This PR should not change game mechanics.
Minor changes: 
- enabled damage calculations for Bone Spirit as viable trap missile.
- fixed Hellfire bug related to Ring of Fire trap vs monsters.
- firewall originating from Oily shrine calculates CTH the same as trap Ring of Fire.

<details>
  <summary>outdated</summary>
This PR will focus on 2 things

1st commit will extract Player and Trap specific code from MonsterMHit and MonsterTrapHit.

It won't be ready to merge yet!

After some review remarks I will commit 2nd commit in which I will merge those 2 almost identical right now functions.

Please note that number of parameters will drasticall change, and names will go shorter, as all function names starting with `PlayerMissileSomething` or `TrapMissileSomething` will be `PlayerMissile::Something` and `TrapMissile::Something`.
So you do not have to focus on that right now.
</details>

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -29,6 +29,12 @@ namespace devilution {
 std::list<Missile> Missiles;
 bool MissilePreFlag;
 
+// temporary declarations, for review purposes
+int PlayerMissileCalculateCTHAgainstMonster(int pnum, missile_id mName, int dist, Monster &monster);
+int PlayerMissileCalculateDamageAgainstMonster(int pnum, int minDam, int maxDam, missile_id mName, Monster &monster);
+void PlayerMissileHitMonster(int pnum, int dam, int m, missile_id mName);
+void MissileHitMonsterConsequences(int mid, int pnum, int dam, missile_id mName);
+
 namespace {
 
 int AddClassHealingBonus(int hp, HeroClass heroClass)
@@ -194,80 +200,26 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
 		return false;
 
 	int hit = GenerateRnd(100);
-	int hper = 0;
-	if (pnum != -1) {
-		const Player &player = Players[pnum];
-		if (MissilesData[t].mType == 0) {
-			hper = player.GetRangedPiercingToHit();
-			hper -= player.CalculateArmorPierce(monster.mArmorClass, false);
-			hper -= (dist * dist) / 2;
-		} else {
-			hper = player.GetMagicToHit() - (monster.mLevel * 2) - dist;
-		}
-	} else {
-		hper = GenerateRnd(75) - monster.mLevel * 2;
-	}
-
+	int hper = PlayerMissileCalculateCTHAgainstMonster(pnum, t, dist, monster);
 	hper = clamp(hper, 5, 95);
-
-	if (monster._mmode == MonsterMode::Petrified)
-		hit = 0;
 
 	if (monster.TryLiftGargoyle())
 		return true;
 
-	if (hit >= hper) {
+	if (hit >= hper && monster._mmode != MonsterMode::Petrified) {
 #ifdef _DEBUG
 		if (!DebugGodMode)
 #endif
 			return false;
 	}
 
-	int dam;
-	if (t == MIS_BONESPIRIT) {
-		dam = monster._mhitpoints / 3 >> 6;
-	} else {
-		dam = mindam + GenerateRnd(maxdam - mindam + 1);
-	}
-
-	const Player &player = Players[pnum];
-
-	if (MissilesData[t].mType == 0 && MissilesData[t].mResist == MISR_NONE) {
-		dam = player._pIBonusDamMod + dam * player._pIBonusDam / 100 + dam;
-		if (player._pClass == HeroClass::Rogue)
-			dam += player._pDamageMod;
-		else
-			dam += player._pDamageMod / 2;
-		if (monster.MData->mMonstClass == MonsterClass::Demon && HasAnyOf(player._pIFlags, ItemSpecialEffect::TripleDemonDamage))
-			dam *= 3;
-	}
-	bool resist = monster.IsResistant(t);
+	int dam = PlayerMissileCalculateDamageAgainstMonster(pnum, mindam, maxdam, t, monster);
 	if (!shift)
 		dam <<= 6;
-	if (resist)
+	if (monster.IsResistant(t))
 		dam >>= 2;
 
-	if (pnum == MyPlayerId)
-		monster._mhitpoints -= dam;
-
-	if ((gbIsHellfire && HasAnyOf(player._pIFlags, ItemSpecialEffect::NoHealOnMonsters)) || (!gbIsHellfire && HasAnyOf(player._pIFlags, ItemSpecialEffect::FireArrows)))
-		monster._mFlags |= MFLAG_NOHEAL;
-
-	if (monster._mhitpoints >> 6 <= 0) {
-		M_StartKill(m, pnum);
-	} else if (resist) {
-		PlayEffect(monster, 1);
-	} else {
-		if (monster._mmode != MonsterMode::Petrified && MissilesData[t].mType == 0 && HasAnyOf(player._pIFlags, ItemSpecialEffect::Knockback))
-			M_GetKnockback(m);
-		if (monster.MType->mtype != MT_GOLEM)
-			M_StartHit(m, pnum, dam);
-	}
-
-	if (monster._msquelch == 0) {
-		monster._msquelch = UINT8_MAX;
-		monster.position.last = player.position.tile;
-	}
+	PlayerMissileHitMonster(pnum, dam, m, t);
 
 	return true;
 }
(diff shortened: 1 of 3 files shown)


Source/missiles.cpp
```diff
@@ -29,6 +29,12 @@ namespace devilution {
 std::list<Missile> Missiles;
 bool MissilePreFlag;
 
+// temporary declarations, for review purposes
```
AJenbo, line 32: I don't get it?

Source/missiles.cpp
```diff
@@ -915,6 +876,100 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+		if (monster._mmode != MonsterMode::Petrified
+		    && MissilesData[mName].mType == 0
+		    && hasKnockback)
+			M_GetKnockback(mid);
+		if (mid > MAX_PLRS - 1)
+			M_StartHit(mid, pnum, dam);
+	}
+}
+
+int PlayerMissileCalculateCTHAgainstMonster(int pnum, missile_id mName, int dist, Monster &monster)
```
AJenbo, line 903: `Player` Considering that the function also accepts not being given a valid player id this function doesn't seam specific to players.

`AgainstMonster` is given by the last argument. This seams to be a good case for overloading where there could be a similar function that takes Player as the last argument.

Maybe a better name is `CalculateMissileChanceToHit()`, or even `Missile::CalculateChanceToHit()` so that it matches `Player::CalculateArmorPierce()` which it calls in turn.

Source/missiles.cpp
```diff
@@ -915,6 +876,100 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+		} else {
+			hper = player.GetMagicToHit() - (monster.mLevel * 2) - dist;
+		}
+	} else {
+		hper = GenerateRnd(75) - monster.mLevel * 2;
+	}
+	return hper;
+}
+
+int PlayerMissileCalculateDamageAgainstMonster(int pnum, int minDam, int maxDam, missile_id mName)
```
AJenbo, line 921: `int Missile(::)CalculateDamage(const Player &player, int minDam, int maxDam, missile_id mName)`

Source/missiles.cpp
```diff
@@ -915,6 +874,105 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+	return dam;
+}
+
+void PlayerMissileHitMonster(int pnum, int dam, int m, missile_id mName)
+{
+	auto &monster = Monsters[m];
+	if (pnum == MyPlayerId)
+		monster._mhitpoints -= dam;
+
+	const auto &player = Players[pnum];
```
AJenbo, line 943:
```suggestion
void PlayerMissileHitMonster(const Player &player, int dam, int m, missile_id mName)
{
	Monster &monster = Monsters[m];
	if (&player == MyPlayer)
		monster._mhitpoints -= dam;

```

Source/missiles.cpp
```diff
@@ -915,6 +874,105 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+		hper = GenerateRnd(75) - monster.mLevel * 2;
+	}
+	return hper;
+}
+
+int PlayerMissileCalculateDamageAgainstMonster(int pnum, int minDam, int maxDam, missile_id mName)
+{
+	int dam = minDam + GenerateRnd(maxDam - minDam + 1);
+
+	const auto &player = Players[pnum];
```
AJenbo, line 925: Take player by ref instead of index, also function name:
```suggestion
int Missile(::)CalculateDamage(const Player &player, int minDam, int maxDam, missile_id mName)
{
	int dam = minDam + GenerateRnd(maxDam - minDam + 1);
```

Source/missiles.cpp
```diff
@@ -915,6 +874,105 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+		if (monster._mmode != MonsterMode::Petrified
+		    && MissilesData[mName].mType == 0
+		    && hasKnockback)
+			M_GetKnockback(mid);
+		if (monster.MType->mtype != MT_GOLEM)
+			M_StartHit(mid, pnum, dam);
+	}
+}
+
+int PlayerMissileCalculateCTHAgainstMonster(int pnum, missile_id mName, int dist, Monster &monster)
```
AJenbo, line 903: `Player` Considering that the function also accepts not being given a valid player id this function doesn't seam specific to players.

`AgainstMonster` is given by the last argument. This seams to be a good case for overloading where there could be a similar function that takes Player as the last argument.

Maybe a better name is `CalculateMissileChanceToHit()`, or even `Missile::CalculateChanceToHit()` so that it matches `Player::CalculateArmorPierce()` which it calls in turn.

Source/missiles.cpp
```diff
@@ -915,6 +874,105 @@ Direction16 GetDirection16(Point p1, Point p2)
...
 }
 
+void MissileHitMonsterConsequences(int mid, int pnum, int dam, missile_id mName)
+{
+	auto &monster = Monsters[mid];
+	const auto &player = Players[pnum];
+
+	bool hasKnockback = false;
+	if (pnum >= 0) {
+		const auto &player = Players[pnum];
```
AJenbo, line 884: You already did this on line 880
```suggestion
```

Source/missiles.cpp
```diff
@@ -915,6 +874,105 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+			M_GetKnockback(mid);
+		if (monster.MType->mtype != MT_GOLEM)
+			M_StartHit(mid, pnum, dam);
+	}
+}
+
+int PlayerMissileCalculateCTHAgainstMonster(int pnum, missile_id mName, int dist, Monster &monster)
+{
+	int hper = 0;
+	if (pnum != -1) {
```
AJenbo, line 906: When you have a special case like this it's preferable to do an early return.
```cpp
if (pnum == -1) {
    return player.GetMagicToHit() - (monster.mLevel * 2) - dist;
}
```

But in this case I would actually suggest that you move the case out of the function so that the function can take `Player &player` as it's input rather then int `pnum`.

Source/missiles.cpp
```diff
@@ -915,6 +874,105 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+		const auto &player = Players[pnum];
+		if (MissilesData[mName].mType == 0) {
```
AJenbo, line 908: Not necessary for this PR, but adding it since it took me a bit of time to figure out. This check is to see if it's a physical arrow or a magic spell. It would be best replaced with something like:
```cpp
bool Missile::isArrow()
{
	IsAnyOf(_mitype, MIS_ARROW, MIS_FARROW, MIS_LARROW, MIS_SPECARROW);
}
```
(`MIS_KRULL` is technically also a physical arrow, but is unused and can be dropped)

With that `mType` could be doped from the `MissileData` struct as well.

Source/missiles.cpp
```diff
@@ -915,6 +874,105 @@ Direction16 GetDirection16(Point p1, Point p2)
...
+
+	if (monster._msquelch == 0) {
+		monster._msquelch = UINT8_MAX;
+		monster.position.last = player.position.tile;
+	}
+}
+
+int TrapMissileCalculateCTHAgainstMonster(int dist, Monster &monster)
+{
+	return 90 - (BYTE)monster.mArmorClass - dist;
```
AJenbo, line 957: The old WINAPI C-style cast can be removed as `mArmorClass` is defined as `uint8_t`.
```suggestion
	return 90 - monster.mArmorClass - dist;
```

Source/missiles.cpp
```diff
@@ -193,78 +199,31 @@ bool MonsterMHit(int pnum, int m, int mindam, int maxdam, int dist, missile_id t
...
 	if (monster.TryLiftGargoyle())
 		return true;
 
-	if (hit >= hper) {
+	if (hit >= hper && monster._mmode != MonsterMode::Petrified) {
 #ifdef _DEBUG
 		if (!DebugGodMode)
 #endif
 			return false;
 	}
```
AJenbo, line 213: This block could be better isolated:

```cpp
if (monster.TryLiftGargoyle())
	return true;

if (monster._mmode != MonsterMode::Petrified) {
	int hit = GenerateRnd(100);
	int hper = PlayerMissileCalculateCTHAgainstMonster(pnum, t, dist, monster);
	hper = clamp(hper, 5, 95);

	if (hit >= hper) {
	#ifdef _DEBUG
		if (!DebugGodMode)
	#endif	
			return false;
	}
}
```
(I'm aware that this moves when RNG is fired, but that isn't an issue here)

Maybe we could even extract this block at some point:
```cpp
bool DidMissilesHit(int pnum, missile_id t, dist, Monster &monster)
{
#ifdef _DEBUG
	if (DebugGodMode)
		return true;
#endif	

	if (monster._mmode == MonsterMode::Petrified)
		return true;

	int toHit = CalculateMissileToHit(pnum, t, dist, monster);

	return GenerateRnd(100) < clamp(toHit, 5, 95);
}

if (monster.TryLiftGargoyle())
	return true;

if (!DidMissilesHit(pnum, t, dist, monster))
	return false;
```

Source/missiles.cpp
k-bar: yes, thank you, the like 880 is the redundant one.

Source/missiles.cpp
```diff
@@ -923,37 +981,32 @@ bool MonsterTrapHit(int m, int mindam, int maxdam, int dist, missile_id t, bool
...
 
-	bool resist = monster.IsResistant(t);
-	int dam = mindam + GenerateRnd(maxdam - mindam + 1);
+	int dam;
+	if (t == MIS_BONESPIRIT)
+		dam = monster._mhitpoints / 3 >> 6;
+	else
+		dam = TrapMissileCalculateDamageAgainstMonster(mindam, maxdam);
```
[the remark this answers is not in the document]
k-bar: **Yes it is prep for later refactor which will be commited to this PR soon**, But I can totally move it to Player part of calculation damage only. Just would need to pass monster there as it totally ignore other thigs.

I was more thinking as first case of missile, which damage is calculated specially (like mods can have more of this). 
And if such missile has specially calculated damage then it won't call player or trap calculatedDamage.
So it is like settings place for such things - but also with one already implemented exception, as Bone Spirit is special already.

Source/missiles.cpp
k-bar: Functions parameters would be changed after merge of MonsterMHit and MonsterTrapHit together.
As for renaming part - let's wait after merge 2 functions together in following commit in this PR.
What brackets in (::) are for? I usually see this without brackets.

Source/missiles.cpp
k-bar: 1. MonsterMHit is called twice from CheckMissileCol
-  `_micaster == TARGET_MONSTERS` and `!missile.IsTrap()`
- `_micaster == TARGET_BOTH`

first - missile from Player only.
second - Walls, Runes, FireRing and Teleport ;)

I am not 100% sure, but 95% that this is Player's only, and even if there are rare cases - MonsterMHit acts like it would be called by Player Missile. (regardless if pnum = -1 exception. Used or not).

There is also MonsterTrapHit called, when mosnters missile hitting golem - and I will put it to TrapMissile class, because it is more about that it acts like a trap.

Let's get back to naming discussion after commit with 2 functions merge to 1 in this PR.

2. Yes I was thinking this may end like that finally. After `Plr2PlrMHit` and `PlayerMHit` refactor I plan, when this one would be succefully merged, those functions could be overloaded. But I will go with your suggestion now.

Source/missiles.cpp
k-bar: yes, I agree to early return, but I am not sure is it even used at all :joy: 

as for parameter list - it will be changed in 2 functions merge to 1 commit in this pr.

Source/missiles.cpp
AJenbo: (::) was to let you know of the two optional names :)

Source/missiles.cpp
AJenbo: You can freely change `MissileData` as it is static data that does not affect the save games.

Source/missiles.cpp
AJenbo: As a consequence of fixing bugs the exact sequence will change. This is also why two versions of the original are not able to join the same game. If someone wants things to be identical to 1.09 then they should use 1.0.x where no such changes had been made. The RNG can be manipulated by attacking the air, or by simply standing still in some cases, so the exact sequence isn't stable and shouldn't be though of as part of the game mechanics. For places where RNG needs to remain stable we write tests to asset this.

Source/missiles.cpp
k-bar: > You can freely change `MissileData` as it is static data that does not affect the save games.

right, I thought somehow about Missile struct :) but its not there.

Source/missiles.cpp
AJenbo: Latest release for 1.0.x is 1.0.3, which has backports from 1.1.0:
https://github.com/diasurgical/devilutionX/releases/tag/1.0.3
The v1.0 branch also has backports from 1.2.0, but no release was created as download numbers are low and the double release confuses some users:
https://github.com/diasurgical/devilutionX/tree/v1.0

Clean up and refactoring of v1.0 is not intended. As things stands more then anything it's missing someone willing to backport bits from 1.3.0 and 1.4.0. I think Blighthorn from Discord also had some requests, some of which might be acceptable. So feel free to work on that if you like.

k-bar:
MonsterMHit and MonsterTrapHit merged now.
Please suggest naming, placing in file and namespace and other remarks in review.
This version of solution is based on template function pattern using template method. But In my opinion it can be converted to use virtuals too. If not now - then later.
I also consider moving `MissileHitMonsterConsequences` into `Collidable` but then it won't be pure interface ;) but maybe it won't be - because of later `Plr2PlrMHit` and `PlayerMHit` similiar refactoring.
After resolve all concerns it should resolve #2350 issue imho.

k-bar:
For easier view of changes - look only at commit changes, because *Files changed* view mess this up.

k-bar:
For easier view of changes - look only at commit changes, because *Files changed* view mess this up.

k-bar:
if you wonder how it would look with virtuals instead function template:
https://github.com/k-bar/devilutionX/commit/cb92aec7ef0509f375514d063a0c0bf9246ed7d3
https://github.com/k-bar/devilutionX/commit/e075fc42f9f4ba024712f7ef1adbe27dd6dbcb8f

AJenbo:
not really, but i can tell you really like virtuals :)

k-bar:
> not really, but i can tell you really like virtuals :)

why not, go take a look it's prutty. 
But seriously I think it may be the way to go, if not now then later, when missiles go further refactoring.
Btw, some low-end devices might go test it now of there is any noticable overhead.

Source/missiles.cpp
k-bar: I did replace auto for Monster, the parameter part is outdated. But I will try to improve it even more.

AJenbo: I hope to continue the review sometime this weekend.

Source/missiles.cpp
```diff
@@ -861,7 +858,9 @@ void PlayerMissile::hitMonster(int mid, int dam) const
 
 int TrapMissile::calculateCTH(Monster &monster) const
 {
-	return 90 - monster.mArmorClass - colMissile->_midist;
+	return (colMissile->_mitype == MIS_FIREWALL
+	        ? GenerateRnd(75) - monster.mLevel * 2
```
k-bar, line 862: This CTH vs monster is now calculated this way also for Firewalls originating from Oily shrine.
I am not sure why even Hellfire devs wanted special CTH vs monster only for Fire Ring trap in first place... Probably it was late found for this missile being TARGET_BOTH, which in all other cases is used for missiles originating from player.
Let me know if we want this special CTH for it at all. I think it could be removed as well.

Source/missiles.cpp
k-bar: this is probably resolved now

Source/missiles.cpp
k-bar: this is probably resolved now

Source/missiles.cpp
k-bar: It turned out that the only exception was Ring of Fire as trap.
I moved this Hellfire exclusive logic out from `PlayerMissile::CalculateCTH` to `TrapMissile::CalculateCTH`.
Now all calls equivalents to MonsterMHit shall be treated as PlayerMissile.

Source/missiles.cpp
k-bar: this is probably resolved now

Source/missiles.cpp
k-bar: this is probably resolved now

Source/missiles.cpp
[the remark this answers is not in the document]
k-bar: this is probably resolved now

Source/missiles.cpp
k-bar: this is probably resolved now

Source/missiles.cpp
```diff
@@ -314,25 +314,25 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
 
 	bool isMonsterHit = false;
 	const int mid = dMonster[mx][my];
-	if (missile._micaster != TARGET_BOTH && !missile.IsTrap()) {
-		if (missile._micaster == TARGET_MONSTERS) {
-			if (mid != 0 && (mid > 0 || Monsters[abs(mid) - 1]._mmode == MonsterMode::Petrified)) {
+	if (mid != 0) {
```
k-bar, line 317: In addition to Fix Fire Ring trap bug, I have made this whole part a bit more readable.
Now, it is a bit clearer without bunch of elses.
It is also now ready to extract function (called when mid != 0).

Source/missiles.cpp
k-bar: > I hope to continue the review sometime this weekend.

 I guess I should ask which one is *this* weekend :cat:

AJenbo: Yeah that is totally fair. Things have been moving pretty fast lately, and there has been more going on in my life then usual so I have been unable to keep up with it all. I'm keeping an open tab for this so it's not forgotten. I also have a note to test your bug fix PRs so we can get them merged as well :)

PR #4607, opened 2022-05-27
Remove PANEL_ defines

obligaron:
## Content
- Introduce `GetUIRectangle` and use it everywhere where a fixed UI position or size is requested (for example main menu, settings menu, in-game chat log, in-game messages, ...). This is needed if main panel size is different then UI size.
- Removes (Main) PANEL_ defines and use `GetMainPanel()` instead.

## Notes
- This is a start to make replacing the main panel easier (for mods or perhaps later settings?).
- Perhaps in the future we can also make some parts of the UI more dynamic (for example a larger settings menu or chat log). 
- Much more refactoring is needed to achieve this.

### DIFF
--- a/Source/DiabloUI/credits.cpp
+++ b/Source/DiabloUI/credits.cpp
@@ -94,8 +94,9 @@ void CreditsRenderer::Render()
 	prev_offset_y_ = offsetY;
 
 	SDL_FillRect(DiabloUiSurface(), nullptr, 0x000000);
-	DrawArt({ PANEL_LEFT - 320, UI_OFFSET_Y }, &ArtBackgroundWidescreen);
-	DrawArt({ PANEL_LEFT, UI_OFFSET_Y }, &ArtBackground);
+	const Point uiPosition = GetUIRectangle().position;
+	DrawArt(uiPosition - Displacement { 320, 0 }, &ArtBackgroundWidescreen);
+	DrawArt(uiPosition, &ArtBackground);
 
 	const std::size_t linesBegin = std::max(offsetY / LINE_H, 0);
 	const std::size_t linesEnd = std::min(linesBegin + MAX_VISIBLE_LINES, linesToRender.size());
@@ -107,15 +108,15 @@ void CreditsRenderer::Render()
 	}
 
 	SDL_Rect viewport = VIEWPORT;
-	viewport.x += PANEL_LEFT;
-	viewport.y += UI_OFFSET_Y;
+	viewport.x += uiPosition.x;
+	viewport.y += uiPosition.y;
 	ScaleOutputRect(&viewport);
 	SDL_SetClipRect(DiabloUiSurface(), &viewport);
 
 	// We use unscaled coordinates for calculation throughout.
-	Sint16 destY = UI_OFFSET_Y + VIEWPORT.y - (offsetY - linesBegin * LINE_H);
+	Sint16 destY = uiPosition.y + VIEWPORT.y - (offsetY - linesBegin * LINE_H);
 	for (std::size_t i = linesBegin; i < linesEnd; ++i, destY += LINE_H) {
-		Sint16 destX = PANEL_LEFT + VIEWPORT.x + 31;
+		Sint16 destX = uiPosition.x + VIEWPORT.x + 31;
 
 		auto &lineContent = linesToRender[i];
 
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -436,6 +436,12 @@ void CalculatePanelAreas()
 		RightPanel.position.x = gnScreenWidth - RightPanel.size.width - LeftPanel.position.x;
 	}
 	RightPanel.position.y = LeftPanel.position.y;
+
+	gnViewportHeight = gnScreenHeight;
+	if (gnScreenWidth <= PANEL_WIDTH) {
+		// Part of the screen is fully obscured by the UI
+		gnViewportHeight -= PANEL_HEIGHT;
+	}
 }
 
 bool IsChatAvailable()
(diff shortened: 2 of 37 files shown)


Source/DiabloUI/credits.cpp
```diff
@@ -94,8 +94,9 @@ void CreditsRenderer::Render()
 	prev_offset_y_ = offsetY;
 
 	SDL_FillRect(DiabloUiSurface(), nullptr, 0x000000);
-	DrawArt({ PANEL_LEFT - 320, UI_OFFSET_Y }, &ArtBackgroundWidescreen);
-	DrawArt({ PANEL_LEFT, UI_OFFSET_Y }, &ArtBackground);
+	auto &uiPosition = GetUIRectangle().position;
```
AJenbo, line 97: It's small and shouldn't be changed.
```suggestion
	const Point uiPosition = GetUIRectangle().position;
```

Source/cursor.cpp
```diff
@@ -264,8 +264,9 @@ void CheckCursMove()
 			sx += GetScreenWidth() / 4;
 		}
 	}
-	if (sy > GetMainPanel().position.y - 1 && MousePosition.x >= GetMainPanel().position.x && MousePosition.x < GetMainPanel().position.x + PANEL_WIDTH && track_isscrolling()) {
-		sy = GetMainPanel().position.y - 1;
+	auto &mainPanel = GetMainPanel();
+	if (sy > mainPanel.position.y - 1 && MousePosition.x >= mainPanel.position.x && MousePosition.x < GetMainPanel().position.x + mainPanel.size.width && track_isscrolling()) {
```
AJenbo, line 268:
```suggestion
	if (mainPanel.Contains(MousePosition) && track_isscrolling()) {
```

Source/diablo.cpp
```diff
@@ -218,7 +218,7 @@ void LeftMouseCmd(bool bShift)
 {
 	bool bNear;
 
-	assert(MousePosition.y < GetMainPanel().position.y || MousePosition.x < GetMainPanel().position.x || MousePosition.x >= GetMainPanel().position.x + PANEL_WIDTH);
+	assert(MousePosition.y < GetMainPanel().position.y || MousePosition.x < GetMainPanel().position.x || MousePosition.x >= GetMainPanel().position.x + GetMainPanel().size.width);
```
AJenbo, line 221:
```suggestion
	assert(!GetMainPanel().Contains(MousePosition));
```

Source/control.cpp
```diff
@@ -1203,13 +1215,13 @@ void control_type_message()
 		return;
 
 	talkflag = true;
-	SDL_Rect rect = MakeSdlRect(PANEL_LEFT + 200, PANEL_Y + 22, 250, 39);
+	SDL_Rect rect = MakeSdlRect(GetMainPanel().position.x + 200, GetMainPanel().position.y + 22, 250, 39);
```
AJenbo, line 1218:
```suggestion
	SDL_Rect rect = MakeSdlRect(GetMainPanel().position + Displacement { 200, 22 }, 250, 39);
```

Source/utils/display.cpp
```diff
@@ -135,12 +135,7 @@ void AdjustToScreenGeometry(Size windowSize)
 {
 	gnScreenWidth = windowSize.width;
 	gnScreenHeight = windowSize.height;
-
-	gnViewportHeight = gnScreenHeight;
-	if (gnScreenWidth <= PANEL_WIDTH) {
-		// Part of the screen is fully obscured by the UI
-		gnViewportHeight -= PANEL_HEIGHT;
-	}
```
AJenbo, line 143: This bit of code was moved to `CalculatePanelAreas()` but you are calling `CalculateUIRectangle()` in it's replacment, is that intentional?

Source/interfac.cpp
```diff
@@ -188,6 +189,21 @@ void DrawCutscene()
...
+void CalculateUIRectangle()
+{
+	constexpr int UIWidth = 640;
+	constexpr int UIHeight = 480;
+	UIRectangle = {
+		{ (gnScreenWidth - UIWidth) / 2, (gnScreenHeight - UIHeight) / 2 },
+		{ UIWidth, UIHeight }
+	};
+}
+
```
AJenbo, line 206: Despite the friendly name of `interface.cpp` this file only deals with the load screens. This should probably live in `control.cpp` instead.

Source/DiabloUI/credits.cpp
obligaron: Changed it. Here and in other occurrences (point and rectangle and for mainpanel), 🙂

Source/cursor.cpp
obligaron: I modified so many lines that I didn't thought about this one. This make it much more simple. 🙂

Source/diablo.cpp
obligaron: Fixed 🙂

Source/utils/display.cpp
obligaron: I added it, cause if we want to support changing resolution while ingame it would make a difference.
Normally `InitControlPan` is called when entering a game, so this should make it always correct.
The tests still need to call it, cause the tests also don't call `AdjustToScreenGeometry` and before used the on-the-fly calculation.

Source/interfac.cpp
obligaron: Initially I did it here, cause `UI_OFFSET_Y` was also definied in `interfac.h` 😉 
Currently `GetUIRectangle` is used for menu UI and ingame UI.
That's why I wouldn't move it to `control.cpp` or `diablui.cpp`.

We could
- Move it to `ui_fwd.h`/`display.cpp` like `GetScreenWidth`
- Separate menu UI and ingame UI size (`GetIngameUIRectangle` and `GetMenuUIRectangle` ?)
- Leave it for a later refactoring
- Something else? 😛

AJenbo: Looks like I made the mistake back in https://github.com/diasurgical/devilutionX/commit/f450d6a125966ef12ecc0aa8bcb23627768a74ba

I guess move it to ui_fwd.h for now

obligaron: Done. 🙂

PR #4612, opened 2022-05-29
Fireball/Elemental: Make explosion calculation line of sight independent

### LINKED ISSUE #1679: Fireball and Elemental explosions do not hit if caster is not in line of sight
**To Reproduce**
1. Cast Fireball
2. Duck around a corner before impact

**Expected behavior**
The explosion should hit all orthogonally and diagonally adjacent spaces around the point of impact, as well as the point of impact, regardless of the caster's line of sight. Although, it shouldn't hit spaces through walls.

**Additional context**
This appears to be due to a rather naïve attempt to prevent explosions from damaging enemies through walls. See the calls to `CheckBlock()` in the processing logic for these missiles.

https://github.com/diasurgical/devilutionX/blob/efbe8a8338ab87258b406c813487463cc3b5ebd1/Source/missiles.cpp#L3659

obligaron:
Fixes #1679

Instead of using the current caster location, we can use the missiles start position to calculate if an explosion is blocked (for example by a wall).

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2950,28 +2950,28 @@ void MI_Fireball(Missile &missile)
 		}
 		MoveMissileAndCheckMissileCol(missile, minDam, maxDam, true, false);
 		if (missile._mirange == 0) {
-			Point m = missile.position.tile;
+			const Point missilePosition = missile.position.tile;
 			ChangeLight(missile._mlid, missile.position.tile, missile._miAnimFrame);
 
-			constexpr Displacement Pattern[] = { { 0, 0 }, { 0, 1 }, { 0, -1 }, { 1, 0 }, { 1, -1 }, { 1, 1 }, { -1, 0 }, { -1, 1 }, { -1, -1 } };
-			for (auto shift : Pattern) {
-				if (!CheckBlock(missile.position.start, m + shift))
-					CheckMissileCol(missile, minDam, maxDam, false, m + shift, true);
+			constexpr Displacement Offsets[] = { { 0, 0 }, { 0, 1 }, { 0, -1 }, { 1, 0 }, { 1, -1 }, { 1, 1 }, { -1, 0 }, { -1, 1 }, { -1, -1 } };
+			for (Displacement offset : Offsets) {
+				if (!CheckBlock(missile.position.start, missilePosition + offset))
+					CheckMissileCol(missile, minDam, maxDam, false, missilePosition + offset, true);
 			}
 
-			if (!TransList[dTransVal[m.x][m.y]]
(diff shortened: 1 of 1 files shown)


StephenCWills (approved):
Well done. You always make it look so simple. 😄 

The only real comment I have is about the weirdly inconsistent variable naming: `Pattern`/`m`/`shift` vs `Offsets`/`c`/`offset`. But it doesn't have much to do with the bugfix.

obligaron:
I can ensure working on it wasn't as easy as it may look. Especially the elemental explosion was ... interesting. 😜 

Yeah the naming is off, if you want I can add a commit to align them. 🙂 
Changes for this pr:
- always use `offset`/`Offsets`
- replace `m`/`c` with `missilePosition`
- replace `monster` with `nextMonster` (elemental only)

StephenCWills:
All I can say is that given how similar these two missiles are, at least in this respect, it was very strange seeing how different the code was between them. Reviewing the code for this PR felt a bit tricky as a result. I think those changes you describe are good regardless so I guess the question is whether we do them now or later as part of a larger cleanup. Since you did a small cleanup of `MI_Element` already that brought it a bit more in line with `MI_Fireball`, I thought you might be interested to go a bit further, so I say go for it. I'll be happy to review this again once it's in.

Source/missiles.cpp
```diff
@@ -2951,28 +2950,28 @@ void MI_Fireball(Missile &missile)
...
-			Point m = missile.position.tile;
+			const Point missilePosition = missile.position.tile;
 			ChangeLight(missile._mlid, missile.position.tile, missile._miAnimFrame);
 
-			constexpr Displacement Pattern[] = { { 0, 0 }, { 0, 1 }, { 0, -1 }, { 1, 0 }, { 1, -1 }, { 1, 1 }, { -1, 0 }, { -1, 1 }, { -1, -1 } };
-			for (auto shift : Pattern) {
-				if (!CheckBlock(p, m + shift))
-					CheckMissileCol(missile, minDam, maxDam, false, m + shift, true);
+			constexpr Displacement Offsets[] = { { 0, 0 }, { 0, 1 }, { 0, -1 }, { 1, 0 }, { 1, -1 }, { 1, 1 }, { -1, 0 }, { -1, 1 }, { -1, -1 } };
+			for (auto offset : Offsets) {
```
StephenCWills, line 2957:
```suggestion
			for (Displacement offset : Offsets) {
```

obligaron: Fixed 🙂

StephenCWills (approved):

AJenbo:
nice works guys :)

PR #4615, opened 2022-05-29
Set default sample rate to 22050 on Windows

AJenbo:


StephenCWills (approved):
I prefer it this way.

AJenbo:
I also appears to work best for the majority of users based on feedback. In the rare cases something unorthodox is needed it's better left to the user apparently.

qndel (approved):

PR #4616, opened 2022-05-29
Cleanup block_lvid checks

AJenbo:
Only consider the relevant flags for block_lvid, and use it as a flag instead of integer values.


StephenCWills (commented):
I guess bit 7 was just an unused bit they had reserved for `block_lvid`? Based on the logic before, if that bit was ever set, it would have disabled the other two flags. 😕

StephenCWills (approved):
I checked all the sol files and answered my own question. I guess they did set bit 7 for some tiles, but bits 5 and 6 were always unset in those cases anyway so this change should be equivalent.

AJenbo:
Interesting that it was set in some cases. I was not aware of that :) I wonder what they might have used it for in the past :thinking:

PR #4618, opened 2022-05-30
fix original Hellfire Berserk ranged bug

k-bar:
Seems like in original version of Hellfire there is a bug that prevents a missile originating from berserking monster hit other monster.
This is proposal to fix this.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -406,7 +406,7 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
 			Monster &attackingMonster = Monsters[missile._misource];
 			if ((attackingMonster._mFlags & MFLAG_TARGETS_MONSTER) != 0
 			    && mid > 0
-			    && (Monsters[mid - 1]._mFlags & MFLAG_GOLEM) != 0)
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
When solving another issue that could cause an OOB crash I ended up also solving this issue as part of of that.

Your solution appears fine, but I instead decided to solve it by checking if the monster being hit was the intended target of the firing monster, since it simplifies the condition.

So I'm going to both approve and close this PR which might seam a bit strange :)

I would appreciate if you could look at my solution here: https://github.com/diasurgical/devilutionX/pull/4800

k-bar:
> When solving another issue that could cause an OOB crash I ended up also solving this issue as part of of that.
> 
> Your solution appears fine, but I instead decided to solve it by checking if the monster being hit was the intended target of the firing monster, since it simplifies the condition.
> 
> So I'm going to both approve and close this PR which might seam a bit strange :)
> 
> I would appreciate if you could look at my solution here: #4800

Ye, I saw your solution, and I have some objection concernign intended target. I left a review comment at your PR.
I think my PR should be merged concerning this issue - since it solves it, without a problem.

k-bar:
Then you could rebase your PR on top of master containing my commit.

k-bar:
although it is my first review, so I did not know that I have to submit it. (it was pending) ;)

AJenbo:
> Then you could rebase your PR on top of master containing my commit.

I don't think there would be much point, it would be more work and it would completely remove all parts of this PR any way.

PR #4621, opened 2022-05-30
Simplify logic of UpdateMissilePos using helpers from 4620

ephphatha:
Quick proof of concept while I worked out what the fuck missiles were actually doing. Turns out it's simply a negated world transformation. There is a rounding error that I initially fixed but reintroduced because it was making it harder to determine if I was changing behaviour.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -181,7 +181,7 @@ void MoveMissilePos(Missile &missile)
 	auto target = missile.position.tile + moveDirection;
 	if (IsTileAvailable(Monsters[missile._misource], target)) {
 		missile.position.tile = target;
-		missile.position.offset += Displacement(moveDirection).WorldToScreen();
+		missile.position.offset += Displacement(moveDirection).worldToScreen();
 	}
 }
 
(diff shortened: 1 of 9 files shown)


Source/missiles.cpp
```diff
@@ -137,18 +137,15 @@ void PutMissile(Missile &missile)
...
-	int dy = 2 * my - mx;
-	int lx = dx / 8;
-	dx = dx / 64;
-	int ly = dy / 8;
-	dy = dy / 64;
-	missile.position.tile = missile.position.start + Displacement { dx, dy };
-	missile.position.offset.deltaX = mx + (dy * 32) - (dx * 32);
-	missile.position.offset.deltaY = my - (dx * 16) - (dy * 16);
-	ChangeLightOffset(missile._mlid, { lx - (dx * 8), ly - (dy * 8) });
+	Displacement pixelsTravelled = missile.position.traveled >> 16;
```
ephphatha, line 140: Doing the bitshift here causes the transformations to lose precision. It'd be more accurate to shift after calling `screenToMissile` etc, but that requires changing the calculations a bit for things like the light offsets and the subtile offset.

AJenbo (approved):

AJenbo:
Crazy stuff :D

PR #4623, opened 2022-05-30
Document/simplify the UpdateMissileVelocity logic

ephphatha:
The multiplication by `1 << 21` was unnecessary, with the way the division is performed it doesn't add precision. Missile velocities are 16 bit fixed values so split out the normalisation step in order to make it clearer what the multiplication and use of `v` ends up doing.


StephenCWills (approved):
❤️

PR #4625, opened 2022-05-30
Fix GuardianTryFireAt missing monster index check

obligaron:
Currently casting a guardian crashes the game, cause the fire check for the guardian access `Monsters` array with negative index.
A monster index check is missing.

Missed this when reviewing #4602 sorry. 😞
Thanks Constantine for reporting. 🙂


AJenbo:
Ok, i'll be sorry about making the mistake and then we can all be friends :)

PR #4634, opened 2022-05-31
Reintroduce c-str overload for SetIniValue()

StephenCWills:
Resolves an issue (introduced by #4544) where all keymapper settings, as well as a handful of other settings in various categories, were selecting the Boolean overload for `SetIniValue()` because they were not explicitly using the `string_view` type to represent section name, key name, and value.

I'm not sure if this is the best approach, but modifying all overloads to use `string_view` instead of `const char *` didn't seem to solve the problem so this is the best I could come up with. I guess C++ considers `bool` a better match for `const char *` than `string_view`, at least in MSVC.

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -197,14 +197,19 @@ void SetIniValue(const char *keyname, const char *valuename, float value)
 	GetIni().SetDoubleValue(keyname, valuename, value, nullptr, true);
 }
 
+void SetIniValue(const char *sectionName, const char *keyName, const char *value)
+{
+	IniChangedChecker changedChecker(sectionName, keyName);
+	auto &ini = GetIni();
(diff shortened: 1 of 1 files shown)


glebm:
> I think this might be a bit backwards. Perhaps the string_view overload should be calling the c-str one to avoid creating copies.

Strings returned by `string_view::c_str` are not guaranteed to be C-strings, i.e. they may not have a trailing `\0`.

StephenCWills:
> `char *` returned by `string_view::c_str` are not guaranteed to be C-strings, i.e. they may not have a trailing `\0`.

I didn't mean to suggest the `std::string` instances are not necessary, but rather that they will result in copying the C-strings from the other overload when it's clearly unnecessary. I'll push a change here in a second.

glebm (approved):

PR #4638, opened 2022-06-01
Select audio device in settings menu

StephenCWills:
Adds an option to the settings menu for selecting the audio device to be used by the game.

The device names are truncated to avoid character overflow with Windows' absurdly long device names, which you can see in the following screenshot.

image

The result looks more like this.

image
image

I was hoping to add an ellipsis to the end of truncated descriptions, but the `GetListDescription()` function of `OptionEntryBase` is const and returns `string_view` so there was no particularly sane way to return a newly allocated string with periods on the end of it. 

This replaces #917


AJenbo:
I'm not sure that the way Linux sets the names is much more sane, but they did fit :)

image

Source/options.cpp
```diff
@@ -812,6 +818,75 @@ void OptionEntryResampler::UpdateDependentOptions() const
...
+#if SDL_VERSION_ATLEAST(2, 0, 0)
+	return SDL_GetNumAudioDevices(false) + 1;
+#else
+	return 1;
+#endif
+}
+
+string_view OptionEntryAudioDevice::GetListDescription(size_t index) const
+{
+	constexpr size_t MaxChars = 32;
```
AJenbo, line 850: Considering that western symbols are not all the same width maybe it would be better to truncate based on width rather then symbol count?

StephenCWills: Well, here's how that looks with a max width of 500.

image
image

PR #4639, opened 2022-06-01
[debug] Export levels as dun maps

AJenbo:
A debug command to export the current level state to a dun map. This is useful both for mappers that want examples to work with (transparency is tricky to get right). And useful for generating fixtures for tests.

... should also become handy in debugging various map generating issues and creating fixes/tests for them.


Source/debug.cpp
```diff
@@ -294,6 +295,82 @@ std::string DebugCmdLoadMap(const string_view parameter)
...
+	std::ofstream dunFile;
+
+	std::string levelName = fmt::format("{}-{}.dun", currlevel, glSeedTbl[currlevel]);
+
+	dunFile.open(levelName, std::ios::out | std::ios::app | std::ios::binary);
+
+	char padding = 0;
+	uint16_t width = DMAXX;
+	uint16_t height = DMAXY;
+	dunFile.write((char *)&width, 2);
```
glebm, line 309: This is an unaligned read (UB) and not endian-correct.

Perhaps something like this:

```c++
void WriteLE16(std::fstream &out, uint16_t val)
{
	const uint16_t littleEndian = SDL_SwapLE16(val);
	char data[2];
	memcpy(data, &littleEndian, 2);
	out.write(data, 2);    
}
```

AJenbo: It wasn't all that important since it's only for debugging, but since you provided a function I'll add it when I get home 😊

AJenbo: There we go, much nicer :)

Source/debug.cpp
```diff
@@ -294,6 +295,84 @@ std::string DebugCmdLoadMap(const string_view parameter)
...
+			WriteLE16(dunFile, 0);
+		}
+	}
+
+	/** Monsters */
+	for (int y = 16; y < MAXDUNY - 16; y++) {
+		for (int x = 16; x < MAXDUNX - 16; x++) {
+			uint16_t monsterId = 0;
+			int mid = abs(dMonster[x][y]) - 1;
+			if (mid != -1) {
```
obligaron, line 336: This would write a moving monster twice and duplicates the monster in the dun file or? 
(one for mid positive and one for mid negative).
Do we want this?

Source/debug.cpp
```diff
@@ -294,6 +295,84 @@ std::string DebugCmdLoadMap(const string_view parameter)
...
+			WriteLE16(dunFile, monsterId);
+		}
+	}
+
+	/** Objects */
+	for (int y = 16; y < MAXDUNY - 16; y++) {
+		for (int x = 16; x < MAXDUNX - 16; x++) {
+			uint16_t objectId = 0;
+			int oid = abs(dObject[x][y]) - 1;
+			if (oid != -1) {
```
obligaron, line 353: Does this work with objects that occupy two tiles (for example Sarcophagus)?
Or get the object added twice?

obligaron (commented):
Some Questions regarding monsters/objects that occupy more then one tile, else looks good to me.

Source/debug.cpp
AJenbo: I did notice the issue for one of the Catacomb maps, but didn't sport the double Sarcophagus, but you are right about this as well :)

image

AJenbo: Fixed:
image

AJenbo:
Made an update to the mapping tools as well so that it can now load/export all the monsters which helped verify this PR as well:
https://github.com/diasurgical/modding-tools/commit/a073e54e00664ef0559bf693933f2caf03f5c0dd

obligaron (approved):

PR #4640, opened 2022-06-01
Test cathedral level generation

AJenbo:


test/drlg_l1_test.cpp
```diff
@@ -19,3 +26,132 @@ TEST(Drlg_l1, DRLG_Init_Globals)
...
+
+	for (int y = 0; y < DMAXY; y++) {
+		for (int x = 0; x < DMAXX; x++) {
+			auto tileId = static_cast<uint8_t>(SDL_SwapLE16(*tileLayer));
+			tileLayer++;
+			EXPECT_EQ(dungeon[x][y], tileId);
+		}
+	}
+
+	const uint16_t *transparantLayer = &dunData[2 + DMAXX * DMAXY * 13];
```
qndel, line 61:
```suggestion
	const uint16_t *transparentLayer = &dunData[2 + DMAXX * DMAXY * 13];
```

test/drlg_l1_test.cpp
```diff
@@ -19,3 +26,134 @@ TEST(Drlg_l1, DRLG_Init_Globals)
...
+		}
+	}
+
+	const uint16_t *transparentLayer = &dunData[2 + DMAXX * DMAXY * 13];
+
+	for (int y = 16; y < 16 + DMAXY * 2; y++) {
+		for (int x = 16; x < 16 + DMAXX * 2; x++) {
+			auto sectorId = static_cast<uint8_t>(SDL_SwapLE16(*transparentLayer));
+			transparentLayer++;
+			EXPECT_EQ(dTransVal[x][y], sectorId);
```
ephphatha, line 69: This is going to generate a lot of output if the level diverges, this should probably be ASSERT_EQ as well.
```suggestion
			ASSERT_EQ(dTransVal[x][y], sectorId) << "Room/region indexes don't match";
```

test/drlg_l1_test.cpp
```diff
@@ -19,3 +26,134 @@ TEST(Drlg_l1, DRLG_Init_Globals)
...
+{
+	TestCreateL5Dungeon(false, 1, 743271966, ENTRY_MAIN);
+	EXPECT_EQ(ViewPosition.x, 51);
+	EXPECT_EQ(ViewPosition.y, 82);
+	TestCreateL5Dungeon(false, 1, 743271966, ENTRY_PREV);
+	EXPECT_EQ(ViewPosition.x, 81);
+	EXPECT_EQ(ViewPosition.y, 47);
+}
+
+TEST(Drlg_l1, CreateL5Dungeon_diablo_2_1383137027)
```
ephphatha, line 84: Having a version of these quest level tests with the other state (diablo unavailable, hellfire available) would be great (in a later pr :D)

test/drlg_l1_test.cpp
```diff
@@ -19,3 +26,134 @@ TEST(Drlg_l1, DRLG_Init_Globals)
...
+	leveltype = DTYPE_CATHEDRAL;
+
+	CreateL5Dungeon(seed, entry);
+
+	std::string path = paths::BasePath();
+
+	paths::SetPrefPath(path);
+	std::string dunPath;
+	if (hellfire)
+		dunPath = fmt::format("../test/fixtures/hellfire/{}-{}.dun", level, seed);
```
ephphatha, line 46: Probably need to add the ability to choose files based on quest state but I wouldn't worry about that for now :D

test/drlg_l1_test.cpp
ephphatha: Also I don't think diverging here is necessarily a problem, as long as the rooms are the same shape they can be given different region IDs... but handling that requires a bit more care.

test/drlg_l1_test.cpp
```diff
@@ -19,3 +26,134 @@ TEST(Drlg_l1, DRLG_Init_Globals)
...
+	else
+		dunPath = fmt::format("../test/fixtures/diablo/{}-{}.dun", level, seed);
+	auto dunData = LoadFileInMem<uint16_t>(dunPath.c_str());
+	EXPECT_EQ(DMAXX, dunData[0]);
+	EXPECT_EQ(DMAXY, dunData[1]);
```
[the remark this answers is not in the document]
AJenbo: Since it's a dump of the generated level the size should match the levels that we generate. So it's sort of a sanity check in case we expand the size of maps or something like that. I guess ASSERT_EQ is a good choice for this so that we stop the test early if this changes.

```suggestion
	ASSERT_EQ(DMAXX, dunData[0]);
	ASSERT_EQ(DMAXY, dunData[1]);
```

test/drlg_l1_test.cpp
AJenbo: Yeah, was a bit random with what I exported since there was still a lot to figure out regarding what and how to test things. So yeah, very much meant to have several follow up PRs once we have the basics down :)

test/drlg_l1_test.cpp
```diff
@@ -19,3 +26,134 @@ TEST(Drlg_l1, DRLG_Init_Globals)
...
+
+	const uint16_t *tileLayer = &dunData[2];
+
+	for (int y = 0; y < DMAXY; y++) {
+		for (int x = 0; x < DMAXX; x++) {
+			auto tileId = static_cast<uint8_t>(SDL_SwapLE16(*tileLayer));
+			tileLayer++;
+			EXPECT_EQ(dungeon[x][y], tileId);
```
AJenbo, line 59:
```suggestion
			ASSERT_EQ(dungeon[x][y], tileId);
```

ephphatha (approved):

mewmew:
Really exciting to see this merged in DevilX!

Marcelo20XX:
Having been a long time follower of this awesome project, what are the new advantages of this new RNG system? I noticed on the current engine that it differs slightly from the Blizzard one, mainly the first level of the cathedral on normal difficulty were too large, I don't remember it was like that on the original. Then again I maybe remembering it wrong...

Thanks in advance for any answers...

qndel:
> Having been a long time follower of this awesome project, what are the new advantages of this new RNG system? I noticed on the current engine that the level generation differs slightly from the Blizzard one, mainly the cathedral levels on normal difficulty were too large, I don't remember it was like that on the original. Then again I maybe remembering it wrong...
> 
> Thanks in advance for any answers...

You remember wrong ... the whole point of this thing is to make sure we generate the same levels

Marcelo20XX:
Well, in that case is an excellent goal, what was wrong with the current engine? The level generation was slight or a lot off?

qndel:
For discussing stuff like this open a ... discussion or come to our discord, closed PRs are not really a place for this

PR #4646, opened 2022-06-01
Exctract duplicate code in ShowProgress

k-bar:
Extracted repeating duplicate code as steps.


AJenbo:
After some talking I think we have decided against this, it's not a good way to logically group things and adds lambdas which is icky :)

PR #4650, opened 2022-06-02
Use 0-based frame index for talk buttons

### LINKED ISSUE #4608: Crash when unmuting the two last buttons in the talk panel
**Important information**
3d2f783101580022d28217ca5a3220e372e834c4 and probably also 1.4.0

**Describe**
First muting and then unmuting the two last buttons in the chat panel will cause the game to crash because of rendering outside the buffer area.

```
    #0 0x55e4e55c1130 in RenderCelClipY<devilution::(anonymous namespace)::<lambda(uint8_t*, const uint8_t*, std::size_t)>, devilution::(anonymous namespace)::<lambda()> > Source/engine/render/cel_render.cpp:70
    #1 0x55e4e55c1130 in RenderCel<devilution::(anonymous namespace)::<lambda(uint8_t*, const uint8_t*, std::size_t)>, devilution::(anonymous namespace)::<lambda()> > Source/engine/render/cel_render.cpp:175
    #2 0x55e4e55c1130 in CelBlitSafeTo Source/engine/render/cel_render.cpp:553
    #3 0x55e4e55c47ff in devilution::CelDrawTo(devilution::Surface const&, devilution::Point, devilution::CelSprite, int) Source/engine/render/cel_render.cpp:602
    #4 0x55e4e4d62a0f in devilution::DrawTalkPan(devilution::Surface const&) Source/control.cpp:1129
    #5 0x55e4e54b067a in devilution::DrawAndBlit() Source/scrollrt.cpp:1729
    #6 0x55e4e4ddaa76 in RunGameLoop Source/diablo.cpp:782
```

StephenCWills:
Only the top button uses a different graphic because the demon statue's wing overlaps the button. So only the middle and bottom buttons would attempt to use the last frame of the CEL, which is the depressed mute button.

It's really hard to tell what's going on with the CEL because we draw over the buttons so they can be translated. It's easiest to see if you focus on the part of the button that overlaps with the wing.

This resolves #4608

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1139,12 +1139,12 @@ void DrawTalkPan(const Surface &out)
 		const Point talkPanPosition = mainPanelPosition + Displacement { 172, 84 + 18 * talkBtn };
 		if (WhisperList[i]) {
 			if (TalkButtonsDown[talkBtn]) {
-				int nCel = talkBtn != 0 ? 4 : 3;
+				int nCel = talkBtn != 0 ? 3 : 2;
 				CelDrawTo(out, talkPanPosition, *talkButtons, nCel);
 				DrawArt(out, talkPanPosition + Displacement { 4, -15 }, &TalkButton, 2);
 			}
 		} else {
-			int nCel = talkBtn != 0 ? 2 : 1;
+			int nCel = talkBtn != 0 ? 1 : 0;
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
I guess this is not the issue affecting 1.4.0 then :thinking: Or maybe myself600 was using a test build at the time.

PR #4652, opened 2022-06-02
Copy test fixtures to build dir (handle default VS2022 directory structure)

ephphatha:
Tests were failing because visual studio uses an intermediate build directory for each configuration, going up one level wasn't enough to reference the test source directory.

### DIFF
--- a/test/drlg_l2_test.cpp
+++ b/test/drlg_l2_test.cpp
@@ -21,10 +21,10 @@ void TestCreateL2Dungeon(int level, uint32_t seed, lvl_entry entry)
 	std::string path = paths::BasePath();
 
 	paths::SetPrefPath(path);
-	std::string dunPath = fmt::format("../test/fixtures/diablo/{}-{}.dun", level, seed);
+	std::string dunPath = fmt::format("test/fixtures/diablo/{}-{}.dun", level, seed);
 	auto dunData = LoadFileInMem<uint16_t>(dunPath.c_str());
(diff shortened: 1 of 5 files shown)


test/drlg_l2_test.cpp
```diff
@@ -50,41 +50,33 @@ void TestCreateL2Dungeon(int level, uint32_t seed, lvl_entry entry)
...
-	EXPECT_EQ(ViewPosition.y, 62);
+	EXPECT_EQ(ViewPosition, Point(26, 62));
 }
 
 TEST(Drlg_l1, CreateL2Dungeon_diablo_6_2034738122)
 {
 	TestCreateL2Dungeon(6, 2034738122, ENTRY_MAIN);
-	EXPECT_EQ(ViewPosition.x, 33);
-	EXPECT_EQ(ViewPosition.y, 26);
+	EXPECT_EQ(ViewPosition, Point(33, 26));
```
AJenbo, line 61: Thanks, I wasn't able to get this working when I first added the tests :)

ephphatha: I actually had to look up other tests to see how I'd been comparing points before, because I remember it working 😅 . `(Point {1 ,2})` prevents syntax errors from the macro expansion but it's ugly, so added the constructor to match Size/Displacement

AJenbo (approved):
Well now I feel embarased for asking 3 other people for suggestions on how to solve this issue and then you just do it :sweat_smile: 

I really love it when we get to improve on each others additions <3

PR #4653, opened 2022-06-03
Test basic Hell levels

AJenbo:
+1.19% coverage

image
image
image
image

level 16 is a bit more complicated to test so I will tackle that with the various quest levels.
(the 4th level here is level 15 with the gate open)


ephphatha (approved):
There's a fair bit of duplication after loading a .dun which would be good to clean up, but that can happen later :)

AJenbo:
My thoughts exactly, I was thinking I would do so once I work on quest maps as the will help to figure out how much is always reused

PR #4658, opened 2022-06-04
Fix crash when switching to English

### LINKED ISSUE #4461: Crash when loading translation
Only indication of language is the phone model: POCO X3

```
libc.so (memcmp+8)
libdevilutionx.so (offset 0x1000) (LanguagePluralTranslate(char const*, char const*, int))
libdevilutionx.so (offset 0x1000) (devilution::CheckInvHLight())
libdevilutionx.so (offset 0x1000) (devilution::CheckCursMove())
```

StephenCWills:
This resolves #4461

### DIFF
--- a/Source/utils/language.cpp
+++ b/Source/utils/language.cpp
@@ -335,7 +335,7 @@ void LanguageInitialize()
 			break;
 	}
 	if (rw == nullptr) {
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Oops :sweat_smile:

PR #4659, opened 2022-06-05
Check player hitpoints in InGameMenu()

### LINKED ISSUE #4427: [Virtual Gamepad] Flash missile interferes with controller logic
**Important information**
Reproduced in 1.4.0, but this should affect any version with touch controls

**Describe**
Pressing the primary action button while the local player's Flash missile is active hides the automap.
Pressing the cancel button while the local player's Flash missile is active brings up the game menu.

**To Reproduce**
1. Go to the dungeon
2. Open the automap
3. Cast the Flash spell
4. After the hit frame and before the spell has ended, press the primary action button
5. The automap should disappear

**Expected behavior**
Assuming no menus are open, these buttons should do nothing at all during the Flash spell.

**Video**
The following video demonstrates the issue twice, first around 20 seconds and then again around 30 seconds.

https://user-images.githubusercontent.com/9203145/163315747-0ce816e0-c399-4269-9968-36a3eef9cd29.mp4

**Additional context**
The primary action button generates a spacebar action when the player is in certain menus.

https://github.com/diasurgical/devilutionX/blob/a66ca446950bceb25ac31ee268dfc8e2ec9259ce/Source/controls/game_controls.cpp#L126-L138

It seems that the `inGameMenu` flag evaluates to `true` when the local player's Flash missile is active because the `InGameMenu()` function checks the `Player::_pInvincible` flag.

https://github.com/diasurgical/devilutionX/blob/a66ca446950bceb25ac31ee268dfc8e2ec9259ce/Source/controls/plrctrls.cpp#L54-L64

Reported on Discord by SinfulPain.

StephenCWills:
I analyzed all the locations in the codebase where `_pInvincible` is assigned, and it seems like my theory in https://github.com/diasurgical/devilutionX/issues/4427#issuecomment-1099145676 was probably correct. Most places that need to determine whether the player is dead check both this flag as well as the player's hitpoints so I did the same thing here.

This resolves #4427

---

Here are my notes for all the places where `_pInvincible` is assigned.

```
LoadPlayer()
    Remember what it was when game was saved

MI_Flash()
MI_Flash2()
    I-frames

StartNewLvl()
RestartTownLvl()
StartWarpLvl()
    MyPlayerId invincible when warping

InitPlayer()
    No longer invincible after warping

StartPlayerKill()
    Invincible when dying/dead

DoResurrect()
RestartTownLvl()
    No longer invincible when revived

GamemenuNewGame()
PrepDoEnding()
    All players invincible before leaving game
```

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -60,7 +60,7 @@ bool InGameMenu()
 	    || qtextflag
 	    || gmenu_is_active()
 	    || PauseMode == 2
-	    || MyPlayer->_pInvincible;
+	    || (MyPlayer->_pInvincible && MyPlayer->_pHitPoints == 0);
 }
 
 namespace {


Source/controls/plrctrls.cpp
```diff
@@ -60,7 +60,7 @@ bool InGameMenu()
 	    || qtextflag
 	    || gmenu_is_active()
 	    || PauseMode == 2
-	    || MyPlayer->_pInvincible;
+	    || (MyPlayer->_pInvincible && MyPlayer->_pHitPoints == 0);
```
k-bar, line 63: Should this be check for shifted hp?
or even <= 0 ?

StephenCWills: Of the places that check the `_pInvincible` flag, there is one location in monster.cpp that uses `player._pHitPoints >> 6 <= 0`. Nine other locations use `player._pHitPoints == 0`. Since it's simpler and doesn't appear to cause any problems, I went with this.

AJenbo (approved):

glebm (approved):

PR #4661, opened 2022-06-05
Refactor inventory/belt item presence and removal

### LINKED ISSUE #4583: Black mushroom quest wont complete if the potion is in your belt
**Important information**
Windows 10 / v1.4.0 Nightly build from 2 days ago / download

**Describe**
If you pick up the potion from Pepin after delivering the brain it can go into your belt. Then Adria won't take it from there. You have to put it into your inventory before ending the quest. Very small issue but maybe a simple fix

**To Reproduce**
1. Do the black mushroom quest
2. deliver the brain to Pepin
3. Pick up the potion with a free slot in your belt
4. go top Adria
5. try to complete the quest

**Expected behavior**
The potion should either not go to your belt, or complete the quest if it's in your belt

**Screenshots**

**Additional context**
It's a very small issue but maybe an easy fix

glebm:
Introduces new functions to check for presence of and remove player items.

These functions do not rely on or expose item indices.

They're implemented as free functions instead of Player methods due to complicated include dependencies between `player.h`, `inv.h`, and `inv_iterators.h`. We should probably look into cleaning this up at some point.

Also fixes #4583 -- the mushroom quest potion check

### DIFF
--- a/Source/inv.h
+++ b/Source/inv.h
@@ -8,6 +8,7 @@
 #include <cstdint>
 
 #include "engine/point.hpp"
+#include "inv_iterators.hpp"
 #include "items.h"
 #include "palette.h"
 #include "player.h"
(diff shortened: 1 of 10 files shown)


Source/inv.h
```diff
@@ -235,6 +235,153 @@ int CalculateGold(Player &player);
...
+{
+	return RemoveInventoryItemById(player, id) || RemoveBeltItemById(player, id);
+}
+
+/**
+ * @brief Removes the first inventory or belt scroll with the player's current spell.
+ *
+ * @return Whether a scroll was found and removed.
+ */
+inline bool removeCurrentSpellScroll(Player &player)
```
AJenbo, line 377:
```suggestion
inline bool RemoveCurrentSpellScroll(Player &player)
```

glebm: Done

AJenbo (approved):

PR #4662, opened 2022-06-05
Rename `inv_iterators` methods to `camelCase`

glebm:
Per clang-tidy config

Also fixes `RemoveBeltItemById` (follow-up to #4661)

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1948,13 +1948,10 @@ bool UseScroll(const spell_id spell)
 	if (pcurs != CURSOR_HAND)
 		return false;
 
-	Player &myPlayer = *MyPlayer;
-
 	if (leveltype == DTYPE_TOWN && !spelldata[spell].sTownSpell)
 		return false;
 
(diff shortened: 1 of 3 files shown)


AJenbo (approved):

Source/inv.cpp
```diff
@@ -1948,13 +1948,10 @@ bool UseScroll(const spell_id spell)
...
 		return false;
 
-	Player &myPlayer = *MyPlayer;
-
 	if (leveltype == DTYPE_TOWN && !spelldata[spell].sTownSpell)
 		return false;
 
-	const InventoryAndBeltPlayerItemsRange items { myPlayer };
-	return std::any_of(items.begin(), items.end(), [spell](const Item &item) {
+	return HasInventoryOrBeltItem(*MyPlayer, [spell](const Item &item) {
```
AJenbo, line 1954: I guess for simplicity we could have called it `HasItem`

glebm: There are also equipped items and stash so it may be better to be explicit

PR #4671, opened 2022-06-06
Fix issues with ears due to translation

### LINKED ISSUE #4641: Ears from players often get bugged
**Important information**
Win7 / DevX 1,4v. / Download

**Describe**
When I kill a player around 50% of the time the ear they drop has a different name, something like "i''or just some random letters and it can't be picked up. Sometimes the ear vanishes but if I kill that player again no ear will drop. I don't know what causes that except some desynch between the two players.

(Now so you don't rush a blame me for killing new people I'm not. I'm offering items in exchange for their ear)

**To Reproduce**
1. Kill a player
2. Watch for an ear with different nametag than the player
'3. See error

StephenCWills:
For ears, store the player's name in the `_iIName` field of the `Item` struct so that the name data can be reliably copied regardless of how translations modify the phrase "Ear of PlayerName".

This resolves #4641

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3222,6 +3222,7 @@ void RecreateEar(Item &item, uint16_t ic, int iseed, int id, int dur, int mdur, 
 	std::string itemName = fmt::format(_(/* TRANSLATORS: {:s} will be a Character Name */ "Ear of {:s}"), heroName);
 
 	CopyUtf8(item._iName, itemName, sizeof(item._iName));
+	CopyUtf8(item._iIName, heroName, sizeof(item._iIName));
 
 	item._iCurs = ((ivalue >> 6) & 3) + ICURS_EAR_SORCERER;
 	item._ivalue = ivalue & 0x3F;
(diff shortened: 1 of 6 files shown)


Source/items.cpp
```diff
@@ -3222,6 +3222,7 @@ void RecreateEar(Item &item, uint16_t ic, int iseed, int id, int dur, int mdur,
 	std::string itemName = fmt::format(_(/* TRANSLATORS: {:s} will be a Character Name */ "Ear of {:s}"), heroName);
 
 	CopyUtf8(item._iName, itemName, sizeof(item._iName));
+	CopyUtf8(item._iIName, heroName, sizeof(item._iIName));
```
AJenbo, line 3225: We should probably make LoadHeroItems() not load ears so make sure we don't get old ones where this fix has not been applied.

StephenCWills: Good thinking, I hadn't thought of that at all.

AJenbo (approved):

PR #4675, opened 2022-06-06
Test Hell quest sets using dummies

AJenbo:
This should cover all the quest maps. There are still some edge cases in the map generation that is not covered

image
image


glebm:
> make[2]: *** No rule to make target '../test/fixtures/hellfire/2-1180526547.dun', needed by 'test/fixtures/hellfire/2-1180526547.dun'.  Stop.

glebm (dismissed):

AJenbo:
Thanks, I totally missed that when looking at the log earlier :D

AJenbo:
> > make[2]: *** No rule to make target '../test/fixtures/hellfire/2-1180526547.dun', needed by 'test/fixtures/hellfire/2-1180526547.dun'.  Stop.

Actually that must be from an old build log, that test currently passes :/

AJenbo:
Alright figured it out, bad quest state again... but only triggered when run under `ctest`, not if run directly via `drlg_l4_test`. So I was unable to reproduce it for a while :/

PR #4677, opened 2022-06-06
replace shareware link with archive.org link

thepwrtank18:
This makes it so that modern browsers can download it

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ For a full list of changes see our [changelog](docs/CHANGELOG.md).
 
 # How to Install
 
-Note: You'll need access to the data from the original game. If you don't have an original CD then you can [buy Diablo from GoG.com](https://www.gog.com/game/diablo). Alternately you can use `spawn.mpq` from the [shareware](http://ftp.blizzard.com/pub/demos/diablosw.exe) version, in place of `DIABDAT.MPQ`, to play the shareware portion of the game.
+Note: You'll need access to the data from the original game. If you don't have an original CD then you can [buy Diablo from GoG.com](https://www.gog.com/game/diablo). Alternately you can use `spawn.mpq` from the [shareware](https://web.archive.org/web/20220303234203/http://ftp.blizzard.com/pub/demos/diablosw.exe) version, in place of `DIABDAT.MPQ`, to play the shareware portion of the game.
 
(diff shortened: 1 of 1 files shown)


StephenCWills:
What browsers are unable to use the current link? Is it an http vs https issue?

tsunamistate:
@StephenCWills Firefox, for example, removed FTP support in version 90 a year ago
Although downloading now it with version 101 works, after HTTP warning (I enabled strict HTTPS mode)

ChaosMarc:
But the current link is NOT ftp but http. The "ftp" in the url is just a subdomain. It should work with every browser there is (I've just tested it with a current Firefox and Chrome). Having a https link would be preferable and as Blizzard doesn't provice a valid certificate for their "ftp" download area using archive.org is a nice workaround in my opinion.

Trihedraf:
Maybe adding the archive.org link as a mirror/backup link would be a better idea so as to keep the official blizzard link.

README.md
```diff
@@ -25,7 +25,7 @@ For a full list of changes see our [changelog](docs/CHANGELOG.md).
 
 # How to Install
 
-Note: You'll need access to the data from the original game. If you don't have an original CD then you can [buy Diablo from GoG.com](https://www.gog.com/game/diablo). Alternately you can use `spawn.mpq` from the [shareware](http://ftp.blizzard.com/pub/demos/diablosw.exe) version, in place of `DIABDAT.MPQ`, to play the shareware portion of the game.
+Note: You'll need access to the data from the original game. If you don't have an original CD then you can [buy Diablo from GoG.com](https://www.gog.com/game/diablo). Alternately you can use `spawn.mpq` from the [shareware](https://web.archive.org/web/20220303234203/http://ftp.blizzard.com/pub/demos/diablosw.exe) version, in place of `DIABDAT.MPQ`, to play the shareware portion of the game.
```
AJenbo, line 28:
```suggestion
Note: You'll need access to the data from the original game. If you don't have an original CD then you can [buy Diablo from GoG.com](https://www.gog.com/game/diablo). Alternately you can use `spawn.mpq` from the [shareware](https://github.com/diasurgical/devilutionx-assets/releases/download/v2/spawn.mpq)[2](http://ftp.blizzard.com/pub/demos/diablosw.exe) version, in place of `DIABDAT.MPQ`, to play the shareware portion of the game.
```

PR #4678, opened 2022-06-06
Clean up door placment in Cathedral code

AJenbo:


Source/drlg_l1.cpp
```diff
@@ -502,6 +502,41 @@ BYTE CornerstoneRoomPattern[27] = {
  */
 BYTE L5ConvTbl[16] = { 22, 13, 1, 13, 2, 13, 13, 13, 4, 13, 1, 13, 2, 13, 16, 13 };
 
+enum CatedralTile : uint8_t {
```
ephphatha, line 505:
```suggestion
enum CathedralTile : uint8_t {
```

I think this could be  a scoped enum (`enum class`) without casting issues if `L5dflags` can use `CathedralTile` as the backing type? Haven't looked into how it's used to see how difficult that'll be.

Source/drlg_l1.cpp
```diff
@@ -502,6 +502,41 @@ BYTE CornerstoneRoomPattern[27] = {
...
+	VDoor          = 25,
+	HDoor          = 26,
+	HFenceVWall    = 27,
+	HDoorVDoor     = 28,
+	VDoorEnd       = 30,
+	HDoorEnd       = 31,
+	VFence         = 35,
+	HFence         = 36,
+	HWallVFence    = 37,
+	HRrchVDoor     = 40,
```
ephphatha, line 532:
```suggestion
	HArchVDoor     = 40,
```
?

Source/drlg_l1.cpp
```diff
@@ -517,58 +552,28 @@ void InitCryptPieces()
...
-				dungeon[x][y] = 25;
-			if (y != 1 && c == 7)
-				dungeon[x][y] = 31;
-			if (x != 1 && c == 6)
-				dungeon[x][y] = 30;
-		}
+	auto tileId = static_cast<CatedralTile>(dungeon[x][y]);
+	if (L5dflags[x][y] == (DLRG_VDOOR | DLRG_HDOOR)) {
+		if (tileId == CatedralTile::DWall)
+			dungeon[x][y] = CatedralTile::HDoorVDoor;
```
ephphatha, line 558: The only real difference I can see between this and the old logic is removing the checks for placing doors in the first row/column of the dungeon layer. Removing the checks seems fine since there's no valid way to have row 0 separated from row 2 that supports a door connection (that I can see anyway 😅 )  and there was no corresponding checks for the other near-edges, is this something that was ever triggered in dungeon generation?

AJenbo: Also it checks for row/column 1 when applying the `DLRG_VDOOR | DLRG_HDOOR` flag, so it would never happen on row/column 1 :)

AJenbo: After looking a bit closer I realize that none of the other cases can ever happen since it will always set the tile to h/vwall when adding the h/vdoor flag, and the reset of the code will avoid modifying it once the flag is set. So we can simply set the door tile directly when building the wall.

Source/debug.cpp
```diff
@@ -73,7 +73,7 @@ enum class DebugGridTextItem : uint16_t {
 	AutomapView,
 	dungeon,
 	pdungeon,
-	dflags,
+	Protected,
```
yuripourre, line 76: Very minor suggestion: Is protected the best word here? Maybe locked would sound better? Feel free to ignore this comment.

ephphatha (approved):
Happy with the wording for `Protected` given it matches the old symbol name, and the bugfix comment seems reasonable to document the flawed instruction.

Source/debug.cpp
ephphatha: I think there's gonna be a bit of work on the dungeon generation code now so plenty of chances to address this in the future :D

AJenbo: Yeah I was not in love with the name either, but couldn't come up with anything better. I sort of hope I can stick to some more user relevant changes, but going over this code is very interesting :)

Issue #4655, opened 2022-06-04
Having lightning and fire damage at the same time breaks melee


qndel:
**Important information**
1.4.0 / https://github.com/diasurgical/devilutionX/commit/0a71f1ffb39d3a3b8ae4e5c21f2fa96076b4daf6

**Describe**
Hitting a monster doesn't deal any fire/lightning damage but casts lightning north-west of player

**To Reproduce**
1. Find some item with lightning damage (not a weapon, for example sparking mail)
2. Find a melee weapon with fire damage
3. Hit a monster

**Expected behavior**
No idea, maybe apply both fire and lightning damage?

**Screenshots**
![thunder](https://user-images.githubusercontent.com/14297035/171993247-7b3388e5-351d-47d6-b70e-19b6f312e34d.gif)

**Additional context**

qndel:
![image](https://user-images.githubusercontent.com/14297035/172328490-f3634ec0-99df-437a-b374-a0b1e529ca4b.png)
Totally forgot about bard so this is even easier to break :<

StephenCWills:
I was able to trace the missile that crashes DevilutionX Diablo. It's spawned here, in `PlrHitMonst()`.

https://github.com/diasurgical/devilutionX/blob/b349e29058c7f6cd88639f267144f529cd5b370d/Source/player.cpp#L854-L857

PR #4681, opened 2022-06-08
Fix Special Arrows

DakkJaniels:
Add in SpecialArrrows as an ItemSpecialEffect for the Hellfire unique items.
Potential fix for #4655


qndel:
Pretty sure this won't work because you aren't using these in SaveItemData / LoadItemData and the effect will be gone when you do new game. At least that's what happened to me when I attempted to fix it.

PR #4683, opened 2022-06-08
Windows Tutorial Doc

ocampoke:
Added a Word Doc with step-by-step instructions with screenshots for installing DevilutionX in Windows.


DakkJaniels:
Probably should make this a pdf or write it in markdown so it can just be displayed in GitHub directly.

AJenbo:
Grate that you are willing to help with improving the documentation.

A Word document isn't really appropriate for this.
- It's not really possible to review a binary file like this
- It's harder to maintain
- Even correcting a single typo would grow the repo by 4MB
- Word documents are a known virus vector

This would also hold true for a PDF.

Maybe this would be best as either updates to the existing instructions, and as pages on the Wiki?

qndel:
I don't see a point when there's https://github.com/diasurgical/devilutionX/blob/master/docs/installing.md and installing on windows is literally copying diabdat.mpq into devilutionx folder or something like this.

PR #4684, opened 2022-06-09
Clean up PlaceMiniSet()

AJenbo:
This cleans up the 4 variants of `PlaceMiniSet()` to get them closer to each other in there implementation (that wasn't easy). It also turned out that there was quite a bit of dead code in them. While doing so I discovered a bug in Catacombs where a check for the quest pieces is done incorrectly in `Substitution()` leading to tile variations not being applied in a cross beam extending out from the quest area. Since it is at the very end of the DRLG and the substitutions are only graphical it was safe to correct the issue:

image

image

5 before:
image
5 after:
image

6 before
image
6 after
image

7 before
image
7 after
image


Source/drlg_l1.cpp
```diff
@@ -732,71 +732,67 @@ void ApplyShadowsPatterns()
...
+			sx++;
+			valid = false;
+		}
+		if (sy <= 12) {
+			sy++;
+			valid = false;
+		}
+		if (!valid) {
+			continue;
+		}
```
AJenbo, line 770: This code is just plain broken. It is causing candelabra and Poison Water entrance to never appear in the upper part of the map. It was probably not even intended to be activated since the input in drlg_l1 is always 0x0, but to deactivate it it should be -1x-1 like in the other levels. But we are pretty much stuck with it now.

image

AJenbo, line 770: This code is just plain broken. It is causing candelabra and Poison Water entrance to never appear in the upper part of the map. It was probably not even intended to be activated since the input in drlg_l1 is always 0x0, but to deactivate it it should be -1x-1 like in the other levels. But we are pretty much stuck with it now.

image

Source/drlg_l1.cpp
```diff
@@ -732,71 +732,68 @@ void ApplyShadowsPatterns()
...
-			for (int yy = 0; yy < sh && abort; yy++) {
-				for (int xx = 0; xx < sw && abort; xx++) {
-					if (miniset[ii] != 0 && dungeon[xx + sx][sy + yy] != miniset[ii])
-						abort = false;
-					if (Protected[xx + sx][sy + yy])
-						abort = false;
-					ii++;
-				}
-			}
+		// Limit the position of setpices for compatability with Diablo bug
```
qndel, line 759:
```suggestion
		// Limit the position of SetPieces for compatability with Diablo bug
```

Source/drlg_l1.cpp
```diff
@@ -830,45 +758,37 @@ void PlaceMiniSetRandom(const BYTE *miniset, int rndper)
...
 			}
 		}
 	}
 }
 
+void PlaceMiniSetRandom1x1(uint8_t search, uint8_t replace, int rndper)
+{
+	uint8_t miniSet[4] { 1, 1, 0, 0 };
+	miniSet[2] = search;
+	miniSet[3] = replace;
```
ephphatha, line 788:
```suggestion
	uint8_t miniSet[4] { 1, 1, search, replace };
```
should be supported? Is there a compiler (gcc 6?) that doesn't allow this?

Source/drlg_l1.cpp
```diff
@@ -732,71 +641,68 @@ void ApplyShadowsPatterns()
...
+			for (int xx = 0; xx < sw && success; xx++) {
+				if (miniset[ii] != 0 && dungeon[xx + sx][sy + yy] != miniset[ii])
+					success = false;
+				if (Protected[xx + sx][sy + yy])
+					success = false;
+				ii++;
 			}
 		}
+		if (success)
+			break;
```
ephphatha, line 695: `Miniset::matches({sx, sy})` would be usable here

Source/drlg_l1.cpp
```diff
@@ -732,71 +641,68 @@ void ApplyShadowsPatterns()
...
-				}
-				ii++;
+	for (int yy = 0; yy < sh; yy++) {
+		for (int xx = 0; xx < sw; xx++) {
+			if (miniset[ii] != 0) {
+				dungeon[xx + sx][sy + yy] = miniset[ii];
 			}
+			ii++;
 		}
 	}
```
ephphatha, line 707: `Miniset::replace({sx, sy})`

Source/drlg_l1.cpp
FitzRoyX: Wow, cool find. What happens to saves if you fix it, objects could spawn over the player and trap him? The floor torches are called "braziers" btw.

Source/drlg_l1.cpp
```diff
@@ -732,71 +641,68 @@ void ApplyShadowsPatterns()
...
-	}
-
-	for (int i = 0; i < numt; i++) {
-		sx = GenerateRnd(DMAXX - sw);
-		sy = GenerateRnd(DMAXY - sh);
-		bool abort = false;
-		int found = 0;
+	for (int bailcnt = 0;; bailcnt++) {
+		if (bailcnt > 4000)
+			return false;
```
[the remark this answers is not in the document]
ephphatha: See https://github.com/diasurgical/devilutionX/compare/PlaceMiniSet...ephphatha:PlaceMiniSet?expand=1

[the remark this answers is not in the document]
ephphatha: Reasonably confident the refactored loop is correct now, sorry if you were trying to review it while I was experimenting. Compared the output of https://github.com/diasurgical/devilutionX/compare/PlaceMiniSet...ephphatha:PlaceMiniSet-og-log?expand=1 and https://github.com/ephphatha/devilutionX/compare/PlaceMiniSet...ephphatha:PlaceMiniSet-log?expand=1 for the test cases I observed breaking to make sure they visit the same tiles in the same order.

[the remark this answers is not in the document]
AJenbo: > edit: turns out dungeon generation is even more broken than I thought joy . The increments on x mean that it actually has a min of 14 and also when y overflows it pushes the x bound in even more, skipping tiles in the first row in some circumstances.

Yeah I decided to just leave it for now and then write some more specific tests for figuring out exactly how to pre calculate the first valid position that matches this madness with out loops.

Source/drlg_l1.cpp
AJenbo: That was the planned next step :)

ephphatha (approved):
With the way it's currently written and the fun I've been having trying to refactor that loop exactly I can confidently say this behaves the same as the original, warts and all :D

AJenbo:
What fortunate timing, have a second part to this :D

Source/drlg_l1.cpp
AJenbo: You would get entirely different layouts so they could end up in a wall:
image

PR #4685, opened 2022-06-09
Load OFILE_CANDLE2 when the Poisoned Water Supply quest is available

StephenCWills:
Fixes an issue reported by Discord user ZeroContrast where the candles in front of Poisoned Water Supply had shown up as perpetually exploding barrels. Here is the savegame provided on Discord: [single_0.sv.zip](https://github.com/diasurgical/devilutionX/files/8873313/single_0.sv.zip). The following is a video they shared that showcases the perpetually exploding barrels.

https://user-images.githubusercontent.com/9203145/172928707-d9a8e327-5f8a-44ee-a61e-a0aa381cd369.mp4

FYI, if you load the savegame on master and go straight to dlvl 2 (`changelevel 2`), the game will attempt to render a null CEL buffer and the candles simply won't show up. If you go through dlvl 1 instead, you probably still will not see the exploding barrels, but the game may crash because it will load some garbage into memory and try to render that.

This will happen on any version of dlvl 2 that has Poisoned Water Supply and no shrines. Therefore, you can reproduce the issue easily enough on a new character by disabling randomized quests, using the following sequence of commands, and then teleporting through the wall to the northeast.

```
god
setspells 15
restart 2 1898487197
changelevel 2
```

The issue does not occur in 1.4.0. In case it matters, this was almost certainly introduced by 9872c7b.


AJenbo:
Ok, this is helpful. I had noticed it during testing of my object PR and have it as todo before setting it as ready for review 😅

PR #4687, opened 2022-06-10
Use templates instead of size member for Miniset

ephphatha:
More of a curiosity than anything, assuming the use of template functions is gonna destroy any possible size saving from only having exact size arrays. But maybe once the functions are unified (following #4684) this'll become worthwhile again?


AJenbo:
I also intend on moving some of the larger blocks out in to .dun files instead of it being baked in to static arrays.

vlolteanu:
This exact idea has crossed my mind. The array used to be 13x13 or something along those lines, and could cause needless cache misses when dealing with the smaller minisets.

However, now the array is 6x6, and probably about to shrink even further. There probably isn't much, if anything to be gained by templatizing it.

AJenbo:
From what I see 5x5 will be the needed size, unless we also make the search pattern in to .dun files. I don't know if that would be a good idear or not?

ephphatha:
Yeah, with the removal of the largest replacement patterns there's no real need for this anymore. There's potential for getting rid of the search pattern completely for things which only look for an area of floor of a given size (especially the ones that place a feature with a border) but that'll be for a separate PR :)

PR #4691, opened 2022-06-11
More DRLG cleanups :)

AJenbo:
@vlolteanu now is a good time to continue applying `Minisets` if you would like. If not then I might start doing so so that we can get this code fixed up :)

The fixture for level 10+anvil has been updated since I had to add a dummy map to use during tests:
Original:
image
Test variant:
image


Source/drlg_l3.cpp
```diff
@@ -2020,58 +1990,68 @@ void Fence()
...
 	}
 
-	setpc_x = sx;
-	setpc_y = sy;
-	setpc_w = sw;
-	setpc_h = sh;
+	// Hack to avoid rivers entering the island, reversed later
+	dungeon[setpc_x + 7][setpc_y + 5] = 2;
+	dungeon[setpc_x + 8][setpc_y + 5] = 2;
+	dungeon[setpc_x + 9][setpc_y + 5] = 2;
```
AJenbo, line 2055: This was applied directly in the static data previously

Source/drlg_l3.cpp
```diff
@@ -1433,74 +1377,113 @@ bool Spawn(int x, int y, int *totarea)
...
-					for (int xx = 0; xx < sw; xx++) {
-						if (miniset[kk] != 0) {
-							dungeon[xx + sx][yy + sy] = miniset[kk];
-						}
-						kk++;
-					}
 				}
 			}
+			if (!found)
+				continue;
```
ephphatha, line 1458: This can use miniset.matches?

Source/drlg_l3.cpp
```diff
@@ -1433,74 +1377,113 @@ bool Spawn(int x, int y, int *totarea)
...
+	    || ComparisonWithBoundsCheck(tile + Direction::SouthEast, tile + Direction::NorthWest)
+	    || ComparisonWithBoundsCheck(tile + Direction::SouthWest, tile + Direction::NorthWest)
+	    || ComparisonWithBoundsCheck(tile + Direction::NorthEast, tile + Direction::NorthWest)) {
+		return false;
+	}
+
+	return true;
+}
+
+bool PlaceMiniSetRandom(const Miniset &miniset, int rndper)
```
ephphatha, line 1437: A comment to describe the behaviour of this function would be good. Something like `/* Randomly places the given miniset throughout the dungeon wherever it would fit, returning true if at least one instance was placed */`?

Source/drlg_l3.cpp
```diff
@@ -1433,74 +1377,113 @@ bool Spawn(int x, int y, int *totarea)
...
-	int sw = miniset[0];
-	int sh = miniset[1];
+	int sw = miniset.size.width;
+	int sh = miniset.size.height;
+	int sx = GenerateRnd(DMAXX - sw) - 1;
+	int sy = GenerateRnd(DMAXY - sh);
+
+	for (int bailcnt = 0;; bailcnt++) {
+		if (bailcnt > 198)
+			return false;
```
ephphatha, line 1389: This is why cave level generation is so bad? It only checks ~6 rows to see if it can place the stairs before bailing 😭

vlolteanu:
This is just a drive-by comment; I'll take a closer look later.

 * I've already made a PR that applies Minisets to drlg_l3, but it's still a draft because I didn't get around to testing it. Feel free to copy code from there, if needed: https://github.com/diasurgical/devilutionX/pull/4471
 * In my PR there's no special case for 1x1 minisets.
 * I've noticed that you look at `search` and `replace`. I'm pretty certain you can use `Miniset::matches` in all instances, in drlg_l3

AJenbo:
@vlolteanu damn i missed that, would definatly have been helpful to just rebase yours for that part. But looks like there are still a few bits that could be used to improve this PR. I don't know if you have seen it, but we now have unit tests for asserting the level generation which makes refactoring all of this much easier.

The 1x1 function reads a lot cleaner once we apply an enum to the tile ids:

```cpp
PlaceRandom1x1(Tile::HWall, Tile::HWallCollapsed, 20);
```
But also it saved having to rewrite all the hellfire 1x1 sets :)

Source/drlg_l3.cpp
AJenbo: It at least explains why a large portion of seeds gets skipped, I wouldn't say it's necessarily bad, but it could cause the same levels to appear multiple times i suppose

Source/drlg_l1.cpp
```diff
@@ -679,34 +775,15 @@ bool PlaceMiniSet(const BYTE *miniset, bool setview)
...
-				if (miniset[ii] != 0 && dungeon[xx + sx][sy + yy] != miniset[ii])
-					success = false;
-				if (Protected[xx + sx][sy + yy])
-					success = false;
-				ii++;
-			}
-		}
-		if (success)
+		if (SetPiecesRoom.Contains({ sx, sy }))
+			continue;
```
ephphatha, line 779: I'd rather add braces here than remove them from the other instances 😂

ephphatha (approved):
Tests pass, ship it!

Source/drlg_l1.cpp
AJenbo: Normally I would agree, but for C++/DevilutionX I'm still wavering :)

glebm: Braces guard against indendation errors but in DevilutionX they can't happen thanks to clang-format, so it's a purely stylistic preference at this point (I prefer less visual noise over consistent braces)

PR #4695, opened 2022-06-12
Move Betrayer bugfix out of PlaceMiniSet

ephphatha:
So that the function is more similar to other dungeon generation versions, and so that it's not doing unnecessary work.

This only applies to singleplayer so I didn't add to the multiplayer dlvl15 test.


AJenbo (approved):
Thanks that's really helpful :)

PR #4696, opened 2022-06-12
Extract position logic from PlaceMiniSet

AJenbo:


Source/drlg_l1.cpp
```diff
@@ -741,18 +741,17 @@ void ApplyShadowsPatterns()
 	}
 }
 
-bool PlaceMiniSet(const Miniset &miniset, bool setview)
+bool PlaceMiniSet(const Miniset &miniset, Point *position = nullptr)
```
ephphatha, line 744: Returning `optional<Point>` was the approach I was considering for this earlier. You can use the return value as a bool with the existence check instead of using a parameter as a conditional out.

Source/drlg_l1.cpp
```diff
@@ -741,18 +741,17 @@ void ApplyShadowsPatterns()
...
 {
 	int sw = miniset.size.width;
 	int sh = miniset.size.height;
-	int sx = GenerateRnd(DMAXX - sw) - 1;
+	int sx = GenerateRnd(DMAXX - sw);
 	int sy = GenerateRnd(DMAXY - sh);
 
-	for (int bailcnt = 0;; bailcnt++) {
+	for (int bailcnt = 0;; bailcnt++, sx++) {
```
ephphatha, line 751: I was thinking this could operate over a range where the iterator operates something like:

```cpp
class DRLGIterator {
  Point pos;
  unsigned iterationCount;
  operator++() {
    iterationCount++;
    x++;
    if (x > MaxX) {
      x = 0;
      y++;
      if (y > MaxY) y = 0;
    }
  }
}

DRLGIterator begin() { return { GenerateRnd(MaxX), GenerateRnd(MaxY), 0 }; }

DRLGIterator end() { return { 0, 0, 4000}; }
```

Keeping the increment logic together makes it easier to refactor to that end.

Source/drlg_l1.cpp
```diff
@@ -2048,30 +2038,50 @@ void CryptFloor(int rndper)
...
 
 	// Place poison water entrance
 	if (Quests[Q_PWATER].IsAvailable()) {
-		if (!PlaceMiniSet(PWATERIN, entry == ENTRY_RTNLVL))
+		Point miniPosition;
+		if (!PlaceMiniSet(PWATERIN, &miniPosition)) {
 			success = false;
-		if (entry == ENTRY_RTNLVL)
-			ViewPosition += Displacement { 2, 3 };
+		} else {
```
ephphatha, line 2049: Returning optional would allow something like:

```cpp
optional<Point> position = PlaceMiniSet(PWATERIN);
if (position) {
    Point miniPosition = position.get();
```

Source/drlg_l1.cpp
```diff
@@ -2048,30 +2038,50 @@ void CryptFloor(int rndper)
...
+			int8_t t = TransVal;
+			TransVal = 0;
+			DRLG_MRectTrans(miniPosition.x, miniPosition.y + 2, miniPosition.x + 5, miniPosition.y + 4);
+			TransVal = t;
+			Displacement entranceOffset { 5, 7 };
+			Quests[Q_PWATER].position = miniPosition * 2 + entranceOffset + Displacement { 16, 16 };
```
ephphatha, line 2055: This comes up a lot, how about a helper in point.cpp :D

```cpp
Point Point::miniToWorld() {
    return {16 + 2 * x, 16 + 2 * y};
}
```

(been meaning to do a PR for this for a while)

AJenbo: Confusingly minisets are constructed out of mega tiles :)

Source/drlg_l1.cpp
AJenbo: It would be nice to get rid of the part that walks over the tiles, but I think I will leave such a change for a later PR, maybe someone else :)

I have unified the 4 versions of `PlaceMiniSet()` as part of this PR, which was the goal with this series, so it should at least be pretty easy to do later.

ephphatha (dismissed):

Source/drlg_l1.cpp
```diff
@@ -830,7 +772,7 @@ void PlaceMiniSetRandom(const Miniset &miniset, int rndper)
 		for (int sx = 0; sx < DMAXX - sw; sx++) {
 			if (!miniset.matches({ sx, sy }, false))
 				continue;
-			if (!CanReplaceTile(miniset.replace[0][0], { sx, sy }))
+			if (!CanReplaceTile(miniset.replace[0][0], { { sx, sy }, miniset.size }))
```
ephphatha, line 775: Feeding this function a rectangle doesn't seem appropriate, plus the PR had a nice self-contained purpose without that commit :P

AJenbo: Reverted.

AJenbo:
Your last comment sounded like a conditional approval, so merged with that change :)

PR #4699, opened 2022-06-13
Use common function for loading .dun tile data

AJenbo:


Source/monster.cpp
```diff
@@ -3899,7 +3886,7 @@ void SetMapMonsters(const uint16_t *dunData, Point startPosition)
 			auto monsterId = static_cast<uint8_t>(SDL_SwapLE16(monsterLayer[j * width + i]));
 			if (monsterId != 0) {
 				int mtype = AddMonsterType(MonstConvTbl[monsterId - 1], PLACE_SPECIAL);
-				PlaceMonster(ActiveMonsterCount++, mtype, i + startPosition.x + 16, j + startPosition.y + 16);
+				PlaceMonster(ActiveMonsterCount++, mtype, startPosition + Displacement { i + 16, j + 16 });
```
ephphatha, line 3889: Pretty sure you've made all callers of this function pass in world coordinates, the border offset has already been added
```suggestion
				PlaceMonster(ActiveMonsterCount++, mtype, startPosition + Displacement { i, j });
```

Source/objects.cpp
```diff
@@ -451,16 +451,14 @@ void AddCandles()
...
 	int cnt = 0;
-	int xp;
-	int yp;
+	Point position;
 	bool exit = false;
 	while (!exit) {
 		exit = true;
-		xp = GenerateRnd(80) + 16;
-		yp = GenerateRnd(80) + 16;
+		position = Point(GenerateRnd(80), GenerateRnd(80)) + Displacement { 16, 16 };
```
ephphatha, line 458: Pretty sure this is indeterminately sequenced due to calling the constructor function-style

>Every expression in a comma-separated list of expressions in a parenthesized initializer is evaluated as if for a function call (indeterminately-sequenced)

You can ensure the roll for x happens before the roll for y by using list initialisation:
>In [list-initialization](https://en.cppreference.com/w/cpp/language/list_initialization), every value computation and side effect of a given initializer clause is sequenced before every value computation and side effect associated with any initializer clause that follows it in the brace-enclosed comma-separated list of initalizers.
```suggestion
		position = Point { GenerateRnd(80), GenerateRnd(80) } + Displacement { 16, 16 };
```

Source/objects.cpp
```diff
@@ -451,16 +451,14 @@ void AddCandles()
...
 		for (int n = -2; n <= 2; n++) {
 			for (int m = -2; m <= 2; m++) {
-				if (!RndLocOk(xp + m, yp + n))
+				if (!RndLocOk(position.x + m, position.y + n))
 					exit = false;
 			}
 		}
```
ephphatha, line 464: Can use the point in rectangle range iterator for this :D
```cpp
for (Point nearPosition : PointInRectangleRange({ position, 2})) {
    if (!RndLocOk(nearPosition))
        exit = false;
}
```

Source/monster.cpp
AJenbo: Good catch

Source/objects.cpp
AJenbo: Going to do this as a follow up as it turns out to result in a lot of other changes that are better done as a separate cleanup.

Source/gendung.cpp
```diff
@@ -604,27 +607,25 @@ void DRLG_PlaceThemeRooms(int minSize, int maxSize, int floor, int freq, bool rn
...
-				themeLoc[themeCount].y = j + 1;
-				themeLoc[themeCount].width = themeW;
-				themeLoc[themeCount].height = themeH;
-				if (IsAnyOf(leveltype, DTYPE_CAVES, DTYPE_NEST))
-					DRLG_RectTrans(2 * i + 20, 2 * j + 20, 2 * (i + themeW) + 15, 2 * (j + themeH) + 15);
-				else
-					DRLG_MRectTrans(i + 1, j + 1, i + themeW, j + themeH);
+				themeLoc[themeCount].room = { { i + 1, j + 1 }, { themeW, themeH } };
+				if (IsAnyOf(leveltype, DTYPE_CAVES, DTYPE_NEST)) {
+					DRLG_RectTrans({ Point(i, j).megaToWorld() + Displacement { 4, 4 }, { themeW * 2 - 5, themeH * 2 - 5 } });
```
ephphatha, line 612: Fixing the loop bounds would let you make this symmetric
```suggestion
					DRLG_RectTrans({ Point(i, j).megaToWorld() + Displacement { 4, 4 }, { themeW * 2 - 4, themeH * 2 - 4 } });
```
However this is effectively doing the Mega -> World transformation manually, so you could also do:
```suggestion
					DRLG_MRectTrans(Point { i, j } + Displacement { 2, 2 }, { themeW - 2, themeH - 2 } });
```

Source/gendung.cpp
```diff
@@ -477,24 +474,26 @@ void DRLG_InitTrans()
...
-void DRLG_RectTrans(int x1, int y1, int x2, int y2)
+void DRLG_RectTrans(Rectangle area)
 {
-	for (int j = y1; j <= y2; j++) {
-		for (int i = x1; i <= x2; i++) {
+	Point position = area.position;
+	Size size = area.size;
+
+	for (int j = position.y; j <= position.y + size.height; j++) {
+		for (int i = position.x; i <= position.x + size.width; i++) {
```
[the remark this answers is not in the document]
AJenbo: Yeah there are a few more cases of this. It's something I would like to address in a later PR when in full call path for these regions have been cleaned up to use Point and Rectangle so that it is easier to see where things are being applied and have logic issues or off by one errors.

Source/gendung.cpp
AJenbo: Thanks that made this a lot cleaner :)

AJenbo: Wait that second one actually fails, since it's 1x1 shorter then the original :) I'll apply the first one and we can try to align things in a later PR :pray:

ephphatha: Yeah that still requires fixing the off-by-one in RectTrans (and removing the size - 1 in MRectRans) :D

AJenbo: I have a strong feeling that there are tons of bugs in how the sections are assigned. It's something we can look in to fixing but preferably where that is the main focus since it will require updating the figures and will need some additional testing since it can affect other parts of the game.

Source/gendung.cpp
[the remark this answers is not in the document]
ephphatha: Need to convert both extent and origin to world space. I ended up getting nerd-sniped and got this:
https://github.com/diasurgical/devilutionX/commit/125da655d587344c311e7997c9e4bb395dff8ea8

(no PR because despite declaring the constructor explicit it's still considered a candidate for an implicit rectangle declaration? previous commit shows the fun getting this to compile.)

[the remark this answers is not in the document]
AJenbo: Lol, that feeling when you just want something to be done :D

```cpp
void DRLG_MRectTrans(Point origin, Point extent)
{
	DRLG_MRectTrans({ origin, { extent.x - origin.x, extent.y - origin.y } });
}
```

Feel free to improve further then merged :joy: :white_flag:

Source/drlg_l3.cpp
```diff
@@ -1867,34 +1866,20 @@ bool PlaceAnvil()
...
-	const uint16_t *tileLayer = &dunData[2];
-	for (int yy = 0; yy < height; yy++) {
-		for (int xx = 0; xx < width; xx++) {
-			int tileId = tileLayer[xx + yy * width];
-			if (tileId != 0) {
-				dungeon[xx + sx][yy + sy] = tileId;
-			}
-		}
-	}
+	PlaceDunTiles(dunData.get(), { sx + 1, sy + 1 }, 7);
```
ephphatha, line 1877: Is this meant to be floorId = 0? The original loop didn't change dungeon if tileId was 0

AJenbo: (3 of them do require that the floor tile is specified so it's not possible to always leave it out either)

Source/quests.cpp
```diff
@@ -189,10 +162,7 @@ void DrawLTBanner(Point position)
 	int width = SDL_SwapLE16(dunData[0]);
 	int height = SDL_SwapLE16(dunData[1]);
 
-	setpc_x = position.x;
-	setpc_y = position.y;
-	setpc_w = width;
-	setpc_h = height;
+	SetPiece = { position, { SDL_SwapLE16(dunData[0]), SDL_SwapLE16(dunData[1]) } };
 
 	const uint16_t *tileLayer = &dunData[2];
```
ephphatha, line 167: I see this one is unique in that it doesn't protect the region, so later level generation steps can trample it.

ephphatha (approved):
I ended up pulling this locally and splitting the changes by function to make it easier to review (could not keep it straight with one commit changing so many loosely related things).

Works well enough and a nice cleanup (apart from what I already mentioned about the RectTrans functions :P)

Source/quests.cpp
AJenbo: This actually happens after the level has been fully generated. The position for it is decided during `FillChambers()` where `Banner1.dun` is placed (and then copied to `pdungeon[]`) before being overwritten with `Banner2.dun` at this stage, which is the one used for the initial level state before you complete the quest. Because it is always placed inside one of the 3 chambers it would already have been protected from the wall generators.

Functions that could potentially overwrite the `pdungeon[]` version:
`FixCornerTiles()`
`ApplyShadowsPatterns()`
`FillFloor()`

AJenbo:
Thanks for the extra effort in reviewing this. I'll try to take your suggestions more to hard and cut future change in to smaller change sets.

Source/quests.cpp
ephphatha: Oooooh, I see now. `pdungeon != dungeon` 😂 . No wonder that doesn't use `PlaceDunTiles`

AJenbo: yeah, this took me a while to figure out how i broke things. For some reason that one quest is loaded in reverse order of literally all other quests where `pdungeon` is handled in a general way.

PR #4700, opened 2022-06-13
Load quest level transparancy from files instead of static inline data 

AJenbo:


qndel:
wait so all this data was inside quest duns already but they hardcoded it anyway?!

AJenbo:
No, I created the two .dun that are included in the PR :)

But they absolutely could have added it to one of there existing ones. My guess is that they didn't have it implemented in there editor in the beginning. Then later when it was added they didn't see a point in revisiting the two affected levels.

glebm (approved):

PR #4711, opened 2022-06-15
Update RectTrans to operate on rectangle range like normal

ephphatha:
Using <= instead of < made this function behave differently to other loops over a Rectangle (it was correct when it operated on two points). Added a constructor for Rectangle taking two points to allow expressing the logic in both cases with natural behaviour each time.

This does mean there are now two constructors that could match `Rectangle { { int, int }, { int, int } }` so all uses of the initialiser list had to be updated to use an explicit Size constructor call or directly use an existing/new Size variable.


tsunamistate:
@ephphatha there were no updates on this since June, are you still interested in continuing this?

ephphatha:
Yeah, think I was waiting for other PRs to get merged before updating this one but I haven't had time to work on it since. Happy for you to have a go if you wanted to finish the cleanup

PR #4713, opened 2022-06-15
Render fonts as PCX (-420 KiB RAM)

glebm:
Reduces peak memory consuption in game by ~420 KiB.

This is because the PCX renderer applies the palette on the fly while rendering, meaning we do not need to duplicate the font for different palettes.

This approach is also a bit slower than precomposed palettes (still plenty fast).


AJenbo:
> This is because the PCX renderer applies the palette on the fly while rendering, meaning we do not need to duplicate the font for different palettes.

This is more inline with how the original worked and will better facilitate https://github.com/diasurgical/devilutionX/pull/4580

> This approach is also a bit slower than precomposed palettes (still plenty fast).

Good to hear was one of my worries when I originally implemented the current font rendering :+1:

PR #4714, opened 2022-06-15
Clean up how DRLG_L1 quest elements are placed

AJenbo:
I think I managed to keep the change set relatively clean this time :cold_sweat: 

This also saves 2K of ram during most of the level generation which I'm sure @glebm will appreciate :D

One possible down side here is that it will load the .dun file for each attempt at generating a valid level.

### DIFF
--- a/Source/drlg_l1.cpp
+++ b/Source/drlg_l1.cpp
@@ -804,11 +804,6 @@ void LoadQuestSetPieces()
 	}
 }
 
-void FreeQuestSetPieces()
-{
-	pSetPiece = nullptr;
-}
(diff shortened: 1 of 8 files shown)


glebm (dismissed):

glebm (dismissed):

AJenbo:
Got cold feet about loading the .dun files repeatedly for each attempt as it can be any where from 1 to hundreds of times.

Instead made all DRLGs load them in the same way and unload them as soon as a layout was successfully generated.

glebm (approved):

PR #4715, opened 2022-06-15
Define ApplyCryptShadowsPatterns

AJenbo:


glebm (approved):

glebm:
Oops, didn't realize this was on auto-merge

AJenbo:
Any reason not to merge it? If you where thinking that it should also be approved by the others that I requests reviews from then I'm actually just intending for any one to review it, not necessarily all :) Hope people are ok with the pings by the way, and feel free to not look at it if someone else already did :)

glebm:
OK! Yeah, that's indeed what I was thinking

PR #4716, opened 2022-06-16
Remove uses of `<locale>`

glebm:
1. The C++ `std::locale("")` constructor performs static initialization of all the facets on the first call.

    For example, this includes things like currency formatting.

    Avoid all of that by using the equivalent C locale call.

2. Replaces the deprecated `<codecvt>` machinery with a simple UTF-8 encoding function.

### DIFF
--- a/Source/platform/locale.cpp
+++ b/Source/platform/locale.cpp
@@ -18,7 +18,7 @@
 #elif defined(__APPLE__) and defined(USE_COREFOUNDATION)
 #include <CoreFoundation/CoreFoundation.h>
 #else
-#include <locale>
+#include <clocale>
 
 #include "utils/log.hpp"
 #endif
@@ -176,12 +176,22 @@ std::vector<std::string> GetLocales()
 
 	CFRelease(languages);
 #else
-	try {
-		std::string locale = std::locale("").name();
-		// strip off any encoding specifier, devX uses UTF8.
-		locales.emplace_back(locale.substr(0, locale.find('.')));
-	} catch (std::runtime_error &e) {
-		LogWarn("Locale detection failed: {}", e.what());
+	const auto from = [&locales](const char *localePtr) -> bool {
+		if (localePtr == nullptr)
+			return false;
+		string_view locale = localePtr;
+		const string_view lang = locale.substr(0, locale.find_first_of(".:,"));
+		if (lang.empty())
+			return false;
+		locales.emplace_back(std::string(lang));
+		return true;
+	};
+	if (!(from(std::getenv("LANGUAGE")) || from(std::getenv("LANG"))
+#ifdef LC_MESSAGES
+	        || from(setlocale(LC_MESSAGES, nullptr))
+#endif
+	        || from(setlocale(LC_CTYPE, nullptr)))) {
+		LogWarn("Locale detection failed");
 	}
 #endif
 	return locales;
--- a/test/utf8_test.cpp
+++ b/test/utf8_test.cpp
@@ -1,0 +1,38 @@
+
+#include <gtest/gtest.h>
+
+#include "utils/utf8.hpp"
+
+namespace devilution {
+namespace {
+
+TEST(AppendUtf8Test, OneByteCodePoint)
+{
+	std::string s = "x";
+	AppendUtf8(U'a', s);
+	EXPECT_EQ(s, "xa");
+}
+
+TEST(AppendUtf8Test, TwoByteCodePoint)
+{
+	std::string s = "x";
+	AppendUtf8(U'Á', s);
+	EXPECT_EQ(s, "xÁ");
+}
+
+TEST(AppendUtf8Test, ThreeByteCodePoint)
+{
+	std::string s;
+	AppendUtf8(U'€', s);
+	EXPECT_EQ(s, "€");
+}
+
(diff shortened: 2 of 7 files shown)


Source/platform/locale.cpp
```diff
@@ -176,12 +176,12 @@ std::vector<std::string> GetLocales()
...
 	CFRelease(languages);
 #else
-	try {
-		std::string locale = std::locale("").name();
-		// strip off any encoding specifier, devX uses UTF8.
-		locales.emplace_back(locale.substr(0, locale.find('.')));
-	} catch (std::runtime_error &e) {
-		LogWarn("Locale detection failed: {}", e.what());
+	const string_view locale = setlocale(0, nullptr);
+	const string_view lang = locale.substr(0, locale.find('.'));
```
ephphatha, line 180: This should probably just consider the messages (fallback to character type if not defined?) category instead of the whole locale. Guessing using LANGUAGE or LANG environment variables would be better here since locales are...almost useless from what I can see.
```suggestion
#ifdef LC_MESSAGES
	const string_view locale = setlocale(LC_MESSAGES, nullptr);
#else
	const string_view locale = setlocale(LC_CTYPE, nullptr);
#endif
	const string_view lang = locale.substr(0, locale.find('.'));
```
If a user has redefined part of their system locale `setlocale(0 /* LC_ALL*/, NULL)` gives a list of `key=value;key=value;...` pairs which will give a garbage first preference.

Having said that, don't bother changing it for this PR. I think it'd be better to rewrite this section to use `getenv(LANGUAGE) > getenv(LANG) > setlocal(LC_MESSAGES) > setlocale(LC_CTYPE)` as a preference order.

Source/platform/locale.cpp
```diff
@@ -176,12 +176,12 @@ std::vector<std::string> GetLocales()
...
+	if (!locale.empty()) {
+		locales.emplace_back(std::string(lang));
+	} else {
+		LogWarn("Locale detection failed: {}", locale);
```
ephphatha, line 184: This is logging a string we know is empty.
```suggestion
		LogWarn("Locale detection failed");
```

test/utf8_test.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "utils/utf8.hpp"
+
+namespace devilution {
+namespace {
+
+TEST(AppendUtf8Test, OneByteCodePoint)
+{
+	std::string s;
+	AppendUtf8(U'a', s);
+	EXPECT_EQ(s, "a");
```
ephphatha, line 13: Despite the function being called AppendUtf8 all current uses are effectively initialisation. If we decided to change the internals to assign instead of append this would be a significant behaviour difference (despite having no impact on the current code). I would expect at least one test to break.

```cpp
s = "test";
AppendUtf8(U'a', s);
EXPECT_EQ(s, "testa");
```

Source/platform/locale.cpp
glebm: Done

test/utf8_test.cpp
glebm: Done

Source/platform/locale.cpp
```diff
@@ -176,12 +176,19 @@ std::vector<std::string> GetLocales()
...
+			return false;
+		locales.emplace_back(std::string(lang));
+		return true;
+	};
+	if (!(from(std::getenv("LANGUAGE")) || from(std::getenv("LANG"))
+#ifdef LC_MESSAGES
+	        || from(setlocale(LC_MESSAGES, nullptr))
+#endif
+	        || from(setlocale(LC_CTYPE, nullptr)))) {
+		LogWarn("Locale detection failed");
```
ephphatha, line 191: `string_view{std::getenv("UNDEFINED_VAR")}` segfaults due to getenv returning a nullptr. Also we can handle languages a bit better given it's expected to be a list of locale codes with no character class suffix

```suggestion
	char* envVar = std::getenv("LANGUAGE");
	if (envVar != nullptr) {
		languages = envVar;
	} else {
		envVar = std::getenv("LANG");
		if (envVar != nullptr) {
			languages = envVar;
		} else {
#ifdef LC_MESSAGES
			languages = setlocale(LC_MESSAGES, nullptr);
#else
			languages = setlocale(LC_CTYPE, nullptr);
#endif
		}
		languages = languages.substr(0, languages.find_first_of("."));
	}

	while (!languages.empty()) {
		size_t separatorPos = languages.find_first_of(":");
		if (separatorPos != 0)
			locales.emplace_back(std::string(languages.substr(0, separatorPos)));

		if (separatorPos != languages.npos)
			languages.remove_prefix(separatorPos + 1);
		else
			break;
	}
```

ephphatha: With LANGUAGES undefined and LANG=en_AU.utf8 I get `en_AU`, if I define languages as "en_GB:zh_CN:ru" I get `en_GB, zh_CN, ru` (haven't tried to slot in LANG after languages, figure one or the other is good enough)

AJenbo:
Just need the new tests to parse :thinking:

Source/platform/locale.cpp
glebm: On my system I get `LANGUAGE=en_GB:en LANG=en_GB.UTF-8`

ephphatha (approved):

PR #4720, opened 2022-06-16
Reduce reliance on `bool done` for control flow (part 1 of x)

vlolteanu:


AJenbo:
This is going to cause some conflicts for sure :D

https://github.com/diasurgical/devilutionX/commit/1af53e57547a146e3876e18b8701223188cc135c

Source/inv.cpp
```diff
@@ -294,73 +294,67 @@ bool AutoEquip(int playerId, const Item &item, inv_body_loc bodyLocation, bool p
...
-	int j = cursorPosition.y;
-
 	Size itemSize = GetInventorySize(player.HoldItem);
 
-	if (!IsHardwareCursor()) {
-		// offset the cursor position to match the hot pixel we'd use for a hardware cursor
-		i += itemSize.width * INV_SLOT_HALF_SIZE_PX;
-		j += itemSize.height * INV_SLOT_HALF_SIZE_PX;
-	}
+	auto possibleSlot = [&]() -> std::optional<int> {
```
AJenbo, line 299: I'm neither a fan of arrow functions or optional, though they can have there place.

AJenbo (dismissed):
Much less aggressive then what I have been working on so a good base for making a follow up PR to :)

Source/inv.cpp
vlolteanu: We could get rid of both by introducing something like `SLOTXY_INVALID = 255`, but I'd rather not do that.

ephphatha: This could be a named function like `size_t FindSlotUnderCursor(Point hotPixelPosition)`. Return `SLOTXY_NOT_FOUND=-1` (or even `SLOTXY_END=NUM_XY_SLOTS`) and it matches the behaviour of other find algorithms like `string::find_first_of` or `std::find(it, end, criteria)`

ephphatha (approved):
some of the loops can be cleaned up further in future PRs, think this is ok as is.

Source/inv.cpp
vlolteanu: `NUM_XY_SLOTS` seems more sensible.

Done.

ephphatha (approved):
Still good😄, makes me wanna take a stab at InvRect after this gets merged 👀

Source/spells.cpp
```diff
@@ -67,42 +67,41 @@ void ClearReadiedSpell(Player &player)
 void PlacePlayer(int pnum)
 {
 	Player &player = Players[pnum];
-	Point newPosition = {};
 
-	if (player.plrlevel == currlevel) {
+	if (player.plrlevel != currlevel)
+		return;
+
+	Point newPosition = [&]() {
```
AJenbo, line 74: I wouldn't mind if this also became it's own function :)

Source/objects.cpp
```diff
@@ -927,35 +927,36 @@ void AddNakrulGate()
...
+		return;
+
+	AddObject(OBJ_STORYBOOK, *position);
+	AddObject(OBJ_STORYCANDLE, *position + Displacement { -2, 1 });
+	AddObject(OBJ_STORYCANDLE, *position + Displacement { -2, 0 });
+	AddObject(OBJ_STORYCANDLE, *position + Displacement { -1, -1 });
+	AddObject(OBJ_STORYCANDLE, *position + Displacement { 1, -1 });
+	AddObject(OBJ_STORYCANDLE, *position + Displacement { 2, 0 });
+	AddObject(OBJ_STORYCANDLE, *position + Displacement { 2, 1 });
 }
```
AJenbo, line 960: This should also be applied to `AddCryptStoryBook` (the first bit should probably be spun in to it's own function so it can be fully shared)

AJenbo:
Merged, with a couple of comments of desired followups :)

PR #4722, opened 2022-06-17
Use decimal separators for gold and XP

vlolteanu:
With the introduction of the stash, players can hoard huge amounts of gold. The total amount of gold can become illegible in the absence of decimal separators.
Screenshot_20220617_140658
Screenshot_20220617_142352

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -42,6 +42,7 @@
 #include "stores.h"
 #include "towners.h"
 #include "trigs.h"
+#include "utils/format_int.hpp"
 #include "utils/language.h"
 #include "utils/sdl_geometry.h"
 #include "utils/stdcompat/optional.hpp"
@@ -878,7 +879,7 @@ void DrawInfoBox(const Surface &out)
 	} else if (!myPlayer.HoldItem.isEmpty()) {
 		if (myPlayer.HoldItem._itype == ItemType::Gold) {
 			int nGold = myPlayer.HoldItem._ivalue;
-			InfoString = fmt::format(fmt::runtime(ngettext("{:d} gold piece", "{:d} gold pieces", nGold)), nGold);
+			InfoString = fmt::format(fmt::runtime(ngettext("{:s} gold piece", "{:s} gold pieces", nGold)), FormatInteger(nGold));
 		} else if (!myPlayer.CanUseItem(myPlayer.HoldItem)) {
 			InfoString = _("Requirements not met");
 		} else {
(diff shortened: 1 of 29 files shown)


AJenbo:
I like it, but i think locale is useless on most none Unix systems. Also @glebm removed it from usage recently so might be good to have some input from him as well.

glebm:
`std::locale("")` is the user's locale, it does not necessary match the game language, right?

AJenbo:
I think we have something for this already in the xp-bar code, maybe that can be reused for this.

glebm:
Yeah, `PrintWithSeparator` is good (can be improved further by splitting into groups of 4 instead of 3 in CJK)

vlolteanu:
XP would be another good candidate, but it doesn't fit in the box if you pass 1M.
Screenshot_20220617_212129

vlolteanu:
I've also left the gold withdrawal inputs untouched. It feels weird to have commas placed there for you and it can't go over 5K (inventory) / 200K (stash) / 400K (hellfire stash)

Translations/fr.po
```diff
@@ -10976,7 +10976,7 @@ msgid ""
 " \n"
 "Perhaps I can tell you more if we speak again. Good luck."
 msgstr ""
-"Dieu merci, vous êtes de retour !\n"
+"Dieu merci, vous êtes de retour !\n"
```
AJenbo, line 10979: Any idea what the change here is?

AJenbo (commented):
Thanks for also updating all the translation files

Translations/fr.po
vlolteanu: Hexdump says it used to be a non-breaking space, and it got replaced with a space. Come to think of it, should there be a space there at all?

vlolteanu:
@AJenbo No problem.

Translations/fr.po
AJenbo: I don't speak French, but maybe @Phalcon22 knows :)

vlolteanu:
I decreased the spacing for XP when over 1B, so that it fits.

Translations/fr.po
vlolteanu: Apparently it's correct. https://www.quora.com/Why-do-French-speakers-place-a-space-before-exclamation-points

Phalcon22Epita: I confirm it is correct.

Phalcon22: I confirm it is correct.

AJenbo: Ok, if you can revert that to be NBS then I think we can merge this :)

vlolteanu: Done.

Source/control.cpp
```diff
@@ -1057,10 +1058,10 @@ void DrawGoldSplit(const Surface &out, int amount)
 	const std::string description = fmt::format(
 	    fmt::runtime(ngettext(
 	        /* TRANSLATORS: {:d} is a number. Dialog is shown when splitting a stash of Gold.*/
```
ephphatha, line 1060:
```suggestion
	        /* TRANSLATORS: {:s} is a number with separators. Dialog is shown when splitting a stash of Gold.*/
```

FitzRoyX:
For some reason I don't like this. Maybe it's because the font is so small that commas and decimals look the same. Maybe it's the inconsistency of only applying it to big numbers, or the fact that the ui is space-starved and can't afford it. There's also creating the uncertainty of typing withdrawals with or without commas.

AJenbo (approved):

AJenbo:
If it helps you can pretend that it's also applied to all the small numbers since there is no difference with numbers that never go above 999

Issue #4723, opened 2022-06-17
Intermittent access violation when returning to SelHeroList from NameSelect


ephphatha:
**Important information**
Windows 11 / DevilutionX master / compiled from 31d43b3b84f8e9cf7c59fd05f2b4593ee0847319

**Describe**
Occasionally when returning to the Hero list (after trying to enter an invalid character name for a new multiplayer character) the UI graphics get corrupted and the game either crashes immediately or crashes on exit.

**To Reproduce**
This is intermittent, but I've been able to reproduce it about 30% of the time by:
1. Go to Multiplayer > loopback > New Hero
2. Type an invalid name (swear, use punctuation, or "admin" somewhere in the name)
3. Submit, you should see a screen saying the name is invalid with a single "OK" prompt and cursor.
4. Hit enter or select OK and if you're unlucky you'll see graphical corruption or a crash.

**Expected behavior**
To consistently go back to the name input so the player can try again.

**Screenshots**
![image](https://user-images.githubusercontent.com/357684/174308542-119fdbc3-295f-4e63-9852-4c769cd45cbd.png)

**Additional context**
Stacktrace on immediate crash points to DrawArt trying to blit to the UI (drawing the character image?):
```
 	SDL2d.dll!SDL_ListRemove(SDL_ListNode * * head, void * ent) Line 1071	C
 	SDL2d.dll!SDL_InvalidateMap(SDL_BlitMap * map) Line 1092	C
 	SDL2d.dll!SDL_UpperBlit_REAL(SDL_Surface * src, const SDL_Rect * srcrect, SDL_Surface * dst, SDL_Rect * dstrect) Line 727	C
 	SDL2d.dll!SDL_UpperBlit(SDL_Surface * a, const SDL_Rect * b, SDL_Surface * c, SDL_Rect * d) Line 497	C
>	devilutionx.exe!devilution::DrawArt(devilution::Point screenPosition, devilution::Art * art, int nFrame, unsigned short srcW, unsigned short srcH) Line 49	C++
 	devilutionx.exe!devilution::`anonymous namespace'::Render(const devilution::UiImage * uiImage) Line 824	C++
 	devilutionx.exe!devilution::`anonymous namespace'::RenderItem(devilution::UiItemBase * item) Line 944	C++
[...]

PR #4724, opened 2022-06-17
Use string_view for UiValidPlayerName

ephphatha:
Now that I've been able to confirm this isn't causing #4723 might as well get some eyes on it 😂 

The use of `std::string` for `tempName` isn't necessary, this could be reverted to use a char array and CopyUtf8 again with a bit of boilerplate to then wrap it in a string view. Didn't feel it was worth avoiding the heap allocation but I can change this if there's a strong preference to put it back on the stack.

Edit: Oh hey, thanks to GCC6 I have to wrap std::string in a string_view to be able to use `find(string_view)` anyway 🙃

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -621,19 +621,27 @@ void UiDestroy()
 	UnloadUiGFX();
 }
 
-bool UiValidPlayerName(const char *name)
+bool UiValidPlayerName(string_view name)
 {
-	if (strlen(name) == 0)
+	if (name.empty())
 		return false;
 
-	if (strpbrk(name, ",<>%&\\\"?*#/:") != nullptr || strpbrk(name, " ") != nullptr)
+	// Currently only allow saving PLR_NAME_LEN bytes as a player name, so if the name is too long we'd have to truncate it.
+	// That said the input buffer is only 16 bytes long...
+	if (name.size() > PLR_NAME_LEN)
 		return false;
 
-	for (BYTE *letter = (BYTE *)name; *letter != '\0'; letter++)
-		if (*letter < 0x20 || (*letter > 0x7E && *letter < 0xC0))
+	if (name.find_first_of(",<>%&\\\"?*#/: ") != name.npos)
+		return false;
+
+	constexpr auto isAsciiControlCharacter = [](char character) -> bool {
+		return (character >= 0x00 && character < 0x20) || character == 0x7F;
+	};
+	for (char letter : name)
+		if (isAsciiControlCharacter(letter) || !IsLeadUtf8CodeUnit(letter))
 			return false;
 
-	const char *const bannedNames[] = {
+	string_view bannedNames[] = {
 		"gvdl",
 		"dvou",
 		"tiju",
@@ -644,13 +652,13 @@ bool UiValidPlayerName(const char *name)
 		"benjo",
 	};
 
-	char tmpname[PLR_NAME_LEN];
-	CopyUtf8(tmpname, name, sizeof(tmpname));
-	for (size_t i = 0, n = strlen(tmpname); i < n; i++)
-		tmpname[i]++;
+	std::string buffer { name };
+	for (char &character : buffer)
+		character++;
 
-	for (const auto *bannedName : bannedNames) {
-		if (strstr(tmpname, bannedName) != nullptr)
+	string_view tempName { buffer };
+	for (string_view bannedName : bannedNames) {
+		if (tempName.find(bannedName) != tempName.npos)
 			return false;
 	}
 
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -634,12 +634,10 @@ bool UiValidPlayerName(string_view name)
 	if (name.find_first_of(",<>%&\\\"?*#/: ") != name.npos)
 		return false;
 
-	constexpr auto isAsciiControlCharacter = [](char character) -> bool {
-		return (character >= 0x00 && character < 0x20) || character == 0x7F;
-	};
-	for (char letter : name)
-		if (isAsciiControlCharacter(letter) || !IsLeadUtf8CodeUnit(letter))
-			return false;
+	// Only basic latin alphabet is supported for multiplayer characters to avoid rendering issues for players who do
+	// not have fonts.mpq installed
+	if (!std::all_of(name.begin(), name.end(), IsBasicLatin))
+		return false;
 
 	string_view bannedNames[] = {
 		"gvdl",
--- a/Source/utils/utf8.hpp
+++ b/Source/utils/utf8.hpp
@@ -32,11 +32,17 @@ inline char32_t ConsumeFirstUtf8CodePoint(string_view *input)
 /**
  * Returns true if this is a trailing byte in a UTF-8 code point encoding.
  *
- * A trailing byte is any byte that is not the heading byte.
+ * Trailing bytes all begin with 10 as the most significant bits, meaning they generally fall in the range 0x80 to
+ * 0xBF. Please note that certain 3 and 4 byte sequences use a narrower range for the second byte, this function is
+ * not intended to guarantee the character is valid within the sequence (or that the sequence is well-formed).
  */
 inline bool IsTrailUtf8CodeUnit(char x)
 {
-	return static_cast<signed char>(x) < -0x40;
+	// The following is equivalent to a bitmask test (x & 0xC0) == 0x80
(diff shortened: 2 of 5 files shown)


Source/DiabloUI/diabloui.cpp
```diff
@@ -621,19 +621,27 @@ void UiDestroy()
...
-	for (BYTE *letter = (BYTE *)name; *letter != '\0'; letter++)
-		if (*letter < 0x20 || (*letter > 0x7E && *letter < 0xC0))
+	if (name.find_first_of(",<>%&\\\"?*#/: ") != string_view::npos)
+		return false;
+
+	constexpr auto isAsciiControlCharacter = [](char character) -> bool {
+		return (character >= 0x00 && character < 0x20) || character == 0x7F;
+	};
+	for (auto letter = name.cbegin(); letter != name.cend(); letter++)
+		if (isAsciiControlCharacter(*letter) || !IsLeadUtf8CodeUnit(*letter))
```
ephphatha, line 641: This also seems busted. The old logic didn't allow multibyte characters in multiplayer hero names since it was excluding any string containing a trailing character? There's no point allowing multibyte leads in that case...

Source/utils/utf8.hpp
```diff
@@ -29,14 +29,29 @@ inline char32_t ConsumeFirstUtf8CodePoint(string_view *input)
...
  *
- * A trailing byte is any byte that is not the heading byte.
+ * Trailing bytes all begin with 10 as the most significant bits, meaning they generally fall in the range 0x80 to
+ * 0xBF. Please note that certain 3 and 4 byte sequences use a narrower range for the second byte, this function is
+ * not intended to guarantee the character is valid within the sequence (or that the sequence is well-formed).
  */
 inline bool IsTrailUtf8CodeUnit(char x)
 {
-	return static_cast<signed char>(x) < -0x40;
+	return (x & 0xC0) == 0x80;
```
glebm, line 54: The previous implementation did 1 comparison while this does an `&` and then a comparison. Any reason for this change?

https://godbolt.org/z/fTnGdW9jr

Source/DiabloUI/diabloui.cpp
```diff
@@ -621,19 +621,27 @@ void UiDestroy()
...
-	for (BYTE *letter = (BYTE *)name; *letter != '\0'; letter++)
-		if (*letter < 0x20 || (*letter > 0x7E && *letter < 0xC0))
+	if (name.find_first_of(",<>%&\\\"?*#/: ") != name.npos)
+		return false;
+
+	constexpr auto isAsciiControlCharacter = [](char character) -> bool {
+		return (character >= 0x00 && character < 0x20) || character == 0x7F;
+	};
+	for (char letter : name)
+		if (isAsciiControlCharacter(letter) || !IsLeadUtf8CodeUnit(letter))
```
glebm, line 641: Looks like this checks that *every* byte is a lead code unit, so would break support for non-ASCII names?

AJenbo: In multiplayer non-ASCII names where not permitted.

ephphatha: Yeah, the original code looks like it was only intended to support basic latin but it did this by looking for the presence of trailing code points specifically. Or it was assuming an ascii encoding and blocking half the extended range for some reason?

I didn't change it here as I didn't want to change behaviour but this could be simplified to a single check `isBasicLatin`.

Source/utils/utf8.hpp
ephphatha: This was mainly to express the criteria in a straightforward way, I was hoping that compilers would be able to optimise them to effectively the same instructions given the types involved :|

Oddly ARM as a target uses a different sequence for each version but ends up at the same instruction count. PowerPC uses fewer instructions for a bitmask test 😂 

I'll revert this with a comment explaining the reason

Source/DiabloUI/diabloui.cpp
ephphatha: Added a commit which changes this block of code to just check for basic latin characters. I think this conveys intent better given devilutionX uses utf8 strings.

Source/utils/utf8.hpp
ephphatha: So while playing around with this I realised this hits implementation defined behaviour on platforms where char is unsigned char by default due to unsigned to signed conversion not being strictly defined before C++20. I still reverted it but added a comment calling out the potential pitfall.

Or I think it can be avoided by casting x to `int` and comparing to a sign-extended int literal?

Source/utils/utf8.hpp
```diff
@@ -29,14 +29,44 @@ inline char32_t ConsumeFirstUtf8CodePoint(string_view *input)
...
+ * Well-formed UTF-8 sequences are described in table 3-7 of S3.9 of the Unicode Standard
+ * see: https://www.unicode.org/versions/Unicode14.0.0/ch03.pdf#G7404
+ *
+ * This is not an inverse of IsTrailUtf8CodeUnit. Byte values C0-C1, F5-FF are not valid anywhere in a UTF-8 sequence.
+ */
+constexpr bool IsLeadUtf8CodeUnit(char x)
+{
+	// single byte character || multibyte character leader
+	return (x >= '\x00' && x <= '\x7F') || (x >= '\xC2' && x <= '\xF4');
+}
```
ephphatha, line 44: Taking signed char explicitly here would also have triggered implementation defined behaviour. Thankfully clang and gcc both generate identical instructions for `signed char x; x >= 0 || ( x >= -62 && x <= -12);` and `char x; (x >= 0 && x <= 127) || (x >= -62 && x <= -12);`

MSVC generates different code but ends up with the same instruction count both ways, so figure it's close enough. Not gonna bother with performance testing this since it's not even used with the latest commit.

Source/utils/utf8.hpp
ephphatha: getting into the weeds a bit but something like the following could also be used to avoid implementation defined behaviour :D

```cpp
#if defined(__cplusplus) && (__cplusplus >= 202002L)
	// The following is equivalent to a bitmask test (x & 0xC0) == 0x80
	// On x86_64 architectures it ends up being one instruction shorter
	// This would invokes implementation defined behaviour on platforms where the underlying type of char is unsigned
	// char which don't support C++20. From C++20 onwards unsigned to signed conversion is well defined
	return static_cast<signed char>(x) < static_cast<signed char>('\xC0');
#else
	return (x & 0xC0) == 0x80;
#endif
```

AJenbo (approved):
font.mpq is only needed for CJK so we could check unicode code points instead.

Source/utils/utf8.hpp
glebm: I don't think we need to worry about it. The reason it became standard in C++20 is because almost everything (certainly everything DevilutionX can run on) is twos complement anyway: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0907r4.html#survey

glebm (approved):

glebm:
There is now a ton of warnings like this:

```
utils/utf8.hpp: In function ‘constexpr bool devilution::IsLeadUtf8CodeUnit(char)’:
utils/utf8.hpp:43:34: warning: comparison is always true due to limited range of data type [-Wtype-limits]
   43 |         return (x >= '\x00' && x <= '\x7F') || (x >= '\xC2' && x <= '\xF4');
      |                                ~~^~~~~~~~~
```

```
utils/utf8.hpp: In function ‘constexpr bool devilution::IsLeadUtf8CodeUnit(char)’:
utils/utf8.hpp:43:34: warning: comparison is always true due to limited range of data type [-Wtype-limits]
   43 |         return (x >= '\x00' && x <= '\x7F') || (x >= '\xC2' && x <= '\xF4');
      |                                ~~^~~~~~~~~
```

ephphatha:
```cpp
bool IsLeadUtf8CodeUnit(char x)
{
    if constexpr(std::is_signed_v<char>)
	    return (x >= '\x00') || (x >= '\xC2' && x <= '\xF4');
    else
	    return (x <= '\x7F') || (x >= '\xC2' && x <= '\xF4');
}
```
if constexpr isn't enough to avoid generating warnings on clang or gcc (even though it correctly compiles only the relevant code for both arm and x86_64). I'm assuming template specialisations/SFINAE would also leave the compiler parsing and warning about the unused test but I could not work out the syntax to enable a specialisation for a function that has an explicit signature. I also can't see any defines that can be used to guard the block.

I think at this point I just give up and cast to signed char 😂

glebm:
We could also just remove `IsLeadUtf8CodeUnit` because I think it's not actually used?

PR #4727, opened 2022-06-17
Operate directly on SOL data

AJenbo:
This also fixes some tiles along the walls in Hell incorrectly being marked as solid.

@glebm could you check memory usage in game before and after this PR, would be interesting to see if you prediction of ~40KiB holds true :)

Examples of tiles that are incorrectly marked as solid:
image

A good follow up to this would be to change dPiece to be zero-indexed so that the additional check and math isn't needed in the rest of the code. Doing so would require applying tile 218 as the default in town instead of 0.

### DIFF
--- a/Source/gendung.cpp
+++ b/Source/gendung.cpp
@@ -442,6 +442,7 @@ void LoadLevelSOLData()
 		break;
 	case DTYPE_HELL:
 		LoadFileInMem("Levels\\L4Data\\L4.SOL", SOLData);
+		SOLData[210] = TileProperties::None; // Tile is incorrectly marked as being solid
 		break;
 	case DTYPE_NEST:
 		LoadFileInMem("NLevels\\L6Data\\L6.SOL", SOLData);
--- a/Source/gendung.cpp
+++ b/Source/gendung.cpp
@@ -30,12 +30,7 @@ std::optional<OwnedCelSprite> pSpecialCels;
 std::unique_ptr<MegaTile[]> pMegaTiles;
 std::unique_ptr<uint16_t[]> pLevelPieces;
 std::unique_ptr<byte[]> pDungeonCels;
-std::array<uint8_t, MAXTILES + 1> block_lvid;
-std::array<bool, MAXTILES + 1> nBlockTable;
-std::array<bool, MAXTILES + 1> nSolidTable;
-std::array<bool, MAXTILES + 1> nTransTable;
-std::array<bool, MAXTILES + 1> nMissileTable;
-std::array<bool, MAXTILES + 1> nTrapTable;
+std::array<TileProperties, MAXTILES> SOLData;
 Point dminPosition;
 Point dmaxPosition;
 dungeon_type leveltype;
@@ -63,30 +58,6 @@ int themeCount;
 THEME_LOC themeLoc[MAXTHEMES];
 
 namespace {
-
-std::unique_ptr<uint8_t[]> LoadLevelSOLData(size_t &tileCount)
-{
-	switch (leveltype) {
-	case DTYPE_TOWN:
-		if (gbIsHellfire)
-			return LoadFileInMem<uint8_t>("NLevels\\TownData\\Town.SOL", &tileCount);
-		return LoadFileInMem<uint8_t>("Levels\\TownData\\Town.SOL", &tileCount);
-	case DTYPE_CATHEDRAL:
-		return LoadFileInMem<uint8_t>("Levels\\L1Data\\L1.SOL", &tileCount);
-	case DTYPE_CATACOMBS:
-		return LoadFileInMem<uint8_t>("Levels\\L2Data\\L2.SOL", &tileCount);
-	case DTYPE_CAVES:
-		return LoadFileInMem<uint8_t>("Levels\\L3Data\\L3.SOL", &tileCount);
-	case DTYPE_HELL:
-		return LoadFileInMem<uint8_t>("Levels\\L4Data\\L4.SOL", &tileCount);
-	case DTYPE_NEST:
-		return LoadFileInMem<uint8_t>("NLevels\\L6Data\\L6.SOL", &tileCount);
-	case DTYPE_CRYPT:
-		return LoadFileInMem<uint8_t>("NLevels\\L5Data\\L5.SOL", &tileCount);
-	default:
-		app_fatal("FillSolidBlockTbls");
-	}
-}
 
 bool WillThemeRoomFit(int floor, int x, int y, int minSize, int maxSize, int *width, int *height)
 {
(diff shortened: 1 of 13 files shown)


yuripourre:
@AJenbo (this is completely off-topic but here I go) based on @qndel's discovery about GOE (Good Or Evil) I am wondering, does SOL means Solid Or Liquid?

AJenbo:
Probably not since Liquid is not a thing in the game. I think it was simply short for Solid, considering that it is the 3 first types that where in it. Then later they probably added transparency and trap information since it was a convenient place to do so.

Source/gendung.cpp
```diff
@@ -30,12 +30,7 @@ std::optional<OwnedCelSprite> pSpecialCels;
 std::unique_ptr<MegaTile[]> pMegaTiles;
 std::unique_ptr<uint16_t[]> pLevelPieces;
 std::unique_ptr<byte[]> pDungeonCels;
-std::array<uint8_t, MAXTILES + 1> block_lvid;
-std::array<bool, MAXTILES + 1> nBlockTable;
-std::array<bool, MAXTILES + 1> nSolidTable;
-std::array<bool, MAXTILES + 1> nTransTable;
-std::array<bool, MAXTILES + 1> nMissileTable;
-std::array<bool, MAXTILES + 1> nTrapTable;
+std::unique_ptr<TileProperties[]> SOLData;
```
glebm, line 33: Is this moved to the heap in order to remove the `MAXTILES` limitation?

glebm:
The heap memory allocated by this function is 1.4 KiB according to heaptrack.

glebm (approved):

glebm:
Top 50 static data and function sizes with this PR:

```
nm --demangle --print-size --size-sort --reverse-sort build-reld-sdl1-nonet-nosound-gperf/devilutionx | head -50 | cut -d' ' -f2,4- | ruby -ne 'size, name = $_.split(" ", 2); puts "#{size.to_i(16).to_s.rjust(6)}\t#{name}"'
```

```
401408	devilution::dpiece_defs_map_2
151600	_ZN10devilution12_GLOBAL__N_18sgLevelsE.lto_priv.0
131072	devilution::paletteTransparencyLookupBlack16
107800	devilution::PopupData
 87200	devilution::Players
 65536	devilution::paletteTransparencyLookup
 56000	devilution::Monsters
 50176	devilution::dPiece
 44032	devilution::Items
 40000	devilution::(anonymous namespace)::sgLocals
 28800	devilution::(anonymous namespace)::path_nodes
 25088	devilution::dMonster
 20866	devilution::RenderTile(devilution::Surface const&, devilution::Point)
 16559	devilution::(anonymous namespace)::CreateLevel(devilution::lvl_entry) [clone .lto_priv.0]
 16512	devilution::storehold
 16384	devilution::(anonymous namespace)::lightblock
 16224	_ZN10devilution12_GLOBAL__N_15stextE.lto_priv.0
 15904	d_print_comp_inner
 15501	devilution::CelBlitOutlineTo(devilution::Surface const&, unsigned char, devilution::Point, devilution::CelSprite, int, bool)
 15359	_sub_I_65535_0.0
 14009	devilution::ParseCmd(int, devilution::TCmd const*)
 13824	devilution::LevelMonsterTypes
 13471	devilution::DrawAndBlit() [clone .part.0]
 13469	devilution::CheckTrigForce()
 13208	devilution::Objects
 12955	devilution::LoadGame(bool)
 12803	devilution::LoadGameLevel(bool, devilution::lvl_entry)
 12544	devilution::dTransVal
 12544	devilution::dPreLight
 12544	devilution::dSpecial
 12544	devilution::dPlayer
 12544	devilution::dObject
 12544	devilution::dCorpse
 12544	devilution::dLight
 12544	devilution::dFlags
 12544	devilution::dItem
 12544	devilution::(anonymous namespace)::dRendered
 12171	void devilution::(anonymous namespace)::RenderTileType<(devilution::(anonymous namespace)::TransparencyType)1, (devilution::(anonymous namespace)::LightType)1>(devilution::(anonymous namespace)::TileType, unsigned char*, int, unsigned char const*, unsigned int const*, unsigned char const*, devilution::(anonymous namespace)::Clip)
 12144	devilution::MonstersData
 11975	devilution::(anonymous namespace)::GenerateLevel(devilution::lvl_entry) [clone .lto_priv.2]
 11160	devilution::(anonymous namespace)::ProcessInput() [clone .lto_priv.0]
 11128	devilution::(anonymous namespace)::GameEventHandler(unsigned int, int, int) [clone .lto_priv.0]
 11009	devilution::Options::Options() [clone .constprop.0]
 10861	devilution::StartStore(devilution::talk_id)
 10534	devilution::game_loop(bool)
 10526	main
 10462	devilution::(anonymous namespace)::GenerateLevel(devilution::lvl_entry) [clone .lto_priv.1]
 10021	devilution::Cl2DrawOutline(devilution::Surface const&, unsigned char, int, int, devilution::CelSprite, int)
 10010	devilution::(anonymous namespace)::GenerateLevel(devilution::lvl_entry) [clone .lto_priv.0]
  9888	devilution::sgOptions
```

AJenbo:
oh i had remembered `bool` as being 32bit, not 8bit.

Source/gendung.cpp
AJenbo: My understanding of the various allocation types is still pretty weak. If i understand correctly fixed sized globals are pre allocated in a static block during boot up, but dynamically allocated memory would have to request general memory (heap?) from the OS at run time, making it the slowest.

I understand the stack as part of the statically allocated memory, reserved for variables used locally in functions. Meaning that the application can do it's own memory management with out context switching as long as it can fit within this size.

I actually thought loading the SOL data in a local variable would still result in a heap allocation, the only difference here being that it was de-allocated once the function returned. But you comment suggest to me that it would have been able to use the stack? If that is the case I guess there would be some logic to decide to do so only if it's small enough to fit in the stack?

I'm not sure what your question actually relates to, but I guess more of the lookup could have been per-calculated by the compiler if done against the static memory rather then the heap? If you think that would make a difference we can change it to be a fixed size, doing so would still end up saving us 10kb (down from 11kb if my math is correct).

AJenbo: I actually have no idea where the MAXTILES limit comes from, unless .cel has an inherent limit to how many frames it can contain I think this is just an arbitrary number and we are free to pick our own as long as it is large enough for the largest tileset (town at 1376).

glebm: `std::array<T, N>` is equivalent to a `T[N]`, so it's in the static memory.

> I actually thought loading the SOL data in a local variable would still result in a heap allocation

Previously the data was loaded into a static array, so there was no heap allocation going on there (other than the file I/O internals which may have some). There was no local variable there. If there were a local `std::array` and it was too large to fit on the stack, the program would simply crash (`std::array` is always stack-allocated).

If we don't plan to increase `MAXTILES`, it's better to change it from `std::array` to `std::unique_ptr`.

glebm:
Yep I tested in town

AJenbo:
I know @obligaron was looking at changing `sgLevels` in to using a map to avoid it eating up memory. Considering that it isn't even used in single player it can be pretty wistful. 

I think we could get rid of `dpiece_defs_map_2` if we constructed full tile strip on load. Alternatively I think there is no reason it is `[MAXDUNX][MAXDUNY] `rather then `[MAXTILES]`, what should reduce it to just 1/6 of the current size.

`Players` could be 1/4 in most cases if allocated it dynamically.

Source/gendung.cpp
glebm: Nevermind, it was indeed temporarily heap-allocated previously (I misread the code)

obligaron:
With #4738 `sgLevels` (Top 2) and `sgLocals` (Top 10) should be gone. 🙂

AJenbo:
I'll take a look at dpiece_defs_map_2 next 🙂

AJenbo:
P. S. thanks to @galaxyhaxz for pointing out that this was still an issue

PR #4728, opened 2022-06-17
Fix: Disable lightning in town

obligaron:
Fixes wrong lightning with new created town level:

grafik

Town doesn't support lightning.
Regressed with 6fa681e5675cdee7c8dbf34cc328d53c4a511cfb

Before `dLight` was initialized with 0 (in `CreateTown`):
`memset(dLight, 0, sizeof(dLight));`
In master `dLight` is initialized with 15 (in `InitGlobals`):
`memset(dLight, DisableLighting ? 0 : 15, sizeof(dLight));`

### DIFF
--- a/Source/gendung.cpp
+++ b/Source/gendung.cpp
@@ -382,7 +382,7 @@ void InitGlobals()
 	memset(dItem, 0, sizeof(dItem));
 	memset(dObject, 0, sizeof(dObject));
 	memset(dSpecial, 0, sizeof(dSpecial));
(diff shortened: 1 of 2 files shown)


AJenbo:
Thanks hadn't had time to look in to this yet, I originally assumed that DisableLighting was what controlled lighting in town :)

obligaron:
I also had the same impression. Initially I thought about setting `DisableLighting` to `true`. Only later I discovered that it was only used for debugging. 😁

PR #4730, opened 2022-06-17
Refactor MI_Apoca

vlolteanu:


Source/missiles.cpp
```diff
@@ -3604,12 +3604,8 @@ void MI_Infra(Missile &missile)
 
 void MI_Apoca(Missile &missile)
 {
-	int id = missile._misource;
-	bool exit = false;
-	int j;
-	int k;
-	for (j = missile.var2; j < missile.var3 && !exit; j++) {
-		for (k = missile.var4; k < missile.var5 && !exit; k++) {
```
obligaron, line 3612: Removing `&& !exist` has a side-effect.
Previous `j` or `k` was increment once more.
This means `var4` and `var6` don't change if the missile hits.
But this is needed to move the Apocalypse spell to the next target.
In this pr the monster will be hit multiple times.

Source/missiles.cpp
```diff
@@ -3619,20 +3615,16 @@ void MI_Apoca(Missile &missile)
...
 				continue;
+
+			int id = missile._misource;
 			AddMissile({ k, j }, { k, j }, Players[id]._pdir, MIS_BOOM, TARGET_MONSTERS, id, missile._midam, 0);
-			exit = true;
-		}
-		if (!exit) {
-			missile.var4 = missile.var6;
+			missile.var2 = j - 1;
+			missile.var4 = k;
```
obligaron, line 3622:
```suggestion
			missile.var2 = j;
			missile.var4 = k + 1;
```
With this changes you should have the old behavior (see other comment).

obligaron (changes requested):

Source/missiles.cpp
vlolteanu: Yikes! Thanks for catching this.

vlolteanu:
Squash please when merging.

obligaron (approved):

PR #4731, opened 2022-06-18
Fix PlaceUniqueMonsters off by one monstertype

obligaron:
Regressed with 4d6ff58c22b2195afc10e1072ec417b013241242
Didn't spot this one in #4729
Only results in a crash, if last monstertype on the level supports the uniques that is placed.

the `!done` condition was moved in the for-loop condition.
This is nicer but has a side-effect, cause the increment logic (`mt++`) of the for loop happens [before](https://en.cppreference.com/w/cpp/language/for) the for-condition is tested (`mt < LevelMonsterTypeCount && !done`).
That means's `mt` is increased one more time, even if `done` is true.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -543,11 +543,12 @@ void PlaceUniqueMonsters()
 		if (UniqueMonstersData[u].mlevel != currlevel)
 			continue;
 
-		bool done = false;
 		int mt;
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Thanks, it's unfortunate with all of these regressions :(

obligaron:
Happens. 😉

PR #4732, opened 2022-06-18
Apply MethodCase config to Rectangle

ephphatha:
Also included Circle and VirtualButton classes due to the shared use.

(Splitting a bunch of cleanups into dedicated PRs to make #4711 more focused)

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -70,7 +70,7 @@ Point FindSlotUnderCursor(Point cursorPosition)
 			InventorySlotSizeInPixels + 1
 		};
 
-		if (cell.Contains(cursorPosition)) {
+		if (cell.contains(cursorPosition)) {
 			return point;
 		}
 	}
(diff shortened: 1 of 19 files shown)


ephphatha:
No idea why the amiga build is failing. I'm pretty sure I didn't miss an instance while renaming these methods.

StephenCWills:
FYI, I reran it just to see if it was consistent, and it failed again. 😕

Source/qol/stash.cpp
```diff
@@ -400,7 +400,7 @@ uint16_t CheckStashHLight(Point mousePosition)
 			{ InventorySlotSizeInPixels.width + 1, InventorySlotSizeInPixels.height + 1 }
 		};
 
-		if (cell.Contains({ mousePosition })) {
+		if (cell.contains({ mousePosition })) {
```
ephphatha, line 403:
```suggestion
		if (cell.contains(mousePosition)) {
```

StephenCWills:
I ran some experiments in Docker, and it seems the amiga build fails because of `Center -> center`. If you don't rename that function, the build succeeds.

ephphatha:
That's annoying 😂 . Split out those 3 lines and I'll look at it when I've had some sleep. Or maybe it'll work in a separate build...

AJenbo:
Sorry for braking this, do you have time to rebase?

PR #4734, opened 2022-06-18
Add helpers for creating/using Rectangles in UI contexts

ephphatha:
Please note there are three behaviour differences included in this PR.

UiEdit text drawing regions now have their vertical dimension shrunk by 2 pixels. I don't think this matters in practice as when text overflows the clip region bounds it doesn't actually get clipped.

Similarly, spell book descriptions have a vertical dimension applied to their clip area now meaning they shouldn't overflow into the area for the next spell description. They still will because of clip regions not being true clip regions (unless the font size lines up reasonably well).

The last column of pixels of the last spell book tab buttons (for both Diablo and Hellfire) is now included in the hit area. The detection for buttons 3 and 4 in Diablo has been adjusted to match the alignment changes applied in DrawSpellBook. I'm not sure why there's a buffer page at the end of `SpellPages` since it shouldn't have been hit anyway but I've left removing that for a future PR.

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -802,18 +802,14 @@ namespace {
 
 void Render(const UiText *uiText)
 {
-	Rectangle rect { { uiText->m_rect.x, uiText->m_rect.y }, { uiText->m_rect.w, uiText->m_rect.h } };
-
 	const Surface &out = Surface(DiabloUiSurface());
-	DrawString(out, uiText->GetText(), rect, uiText->GetFlags() | UiFlags::FontSizeDialog);
+	DrawString(out, uiText->GetText(), MakeRectangle(uiText->m_rect), uiText->GetFlags() | UiFlags::FontSizeDialog);
 }
 
 void Render(const UiArtText *uiArtText)
 {
-	Rectangle rect { { uiArtText->m_rect.x, uiArtText->m_rect.y }, { uiArtText->m_rect.w, uiArtText->m_rect.h } };
-
 	const Surface &out = Surface(DiabloUiSurface());
-	DrawString(out, uiArtText->GetText(), rect, uiArtText->GetFlags(), uiArtText->GetSpacing(), uiArtText->GetLineHeight());
+	DrawString(out, uiArtText->GetText(), MakeRectangle(uiArtText->m_rect), uiArtText->GetFlags(), uiArtText->GetSpacing(), uiArtText->GetLineHeight());
 }
 
 void Render(const UiImage *uiImage)
@@ -865,10 +861,8 @@ void Render(const UiImageAnimatedPcx *uiImage)
 
 void Render(const UiArtTextButton *uiButton)
 {
-	Rectangle rect { { uiButton->m_rect.x, uiButton->m_rect.y }, { uiButton->m_rect.w, uiButton->m_rect.h } };
-
 	const Surface &out = Surface(DiabloUiSurface());
-	DrawString(out, uiButton->GetText(), rect, uiButton->GetFlags());
+	DrawString(out, uiButton->GetText(), MakeRectangle(uiButton->m_rect), uiButton->GetFlags());
 }
 
 void Render(const UiList *uiList)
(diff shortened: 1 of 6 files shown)


glebm (approved):

ephphatha:
I realised that the comment I added regarding hellfire/diablo spell book tab buttons was incorrect, I believe the description is now correct. Now the last column of pixels of tab 4 in Diablo is active and the boundary between 3 and 4 is visually aligned :D

glebm (dismissed):

Source/DiabloUI/diabloui.cpp
```diff
@@ -922,7 +916,8 @@ void Render(const UiEdit *uiEdit)
 {
 	DrawSelector(uiEdit->m_rect);
 
-	Rectangle rect { { uiEdit->m_rect.x + 43, uiEdit->m_rect.y + 1 }, { uiEdit->m_rect.w - 86, uiEdit->m_rect.h } };
+	// To simulate padding we inset the region used to draw text in an edit control
+	Rectangle rect = MakeRectangle(uiEdit->m_rect).inset({ 43, 1 });
```
obligaron, line 920: Note:
This changes the size height or?
Previous it wasn't reduced, now it is is reduced by 2.
I think this is a improvement, just wanted to mention it.

Source/panels/spell_book.cpp
```diff
@@ -184,16 +183,16 @@ void DrawSpellBook(const Surface &out)
...
+		yp += SpellBookDescription.height;
 	}
 }
 
 void CheckSBook()
 {
-	Rectangle iconArea = { GetPanelPosition(UiPanels::Spell, { 11, 18 }), { 48 - 11, 314 - 18 } };
-	Rectangle tabArea = { GetPanelPosition(UiPanels::Spell, { 7, 320 }), { 311 - 7, 349 - 320 } };
+	// Icons are drawn in a column near the left side of the panel and aligned with the spell book description entries
+	Rectangle iconArea = { GetPanelPosition(UiPanels::Spell, { 11, 18 }), Size { 37, SpellBookDescription.height * 7 } };
```
obligaron, line 193: I think this is intended (to make hit size of all buttons the same), but just to make sure:
The height is different between master and this pr
master: 314 - 18 = 296
pr: 43 x 7 = 301

obligaron (dismissed):
lgtm

Source/DiabloUI/diabloui.cpp
ephphatha: Yep, that's the first point I was calling out :D
> UiEdit text drawing regions now have their vertical dimension shrunk by 2 pixels. I don't think this matters in practice as when text overflows the clip region bounds it doesn't actually get clipped.

Source/panels/spell_book.cpp
ephphatha: Looking at this more I think the position of the hit area for spell book icons is also wrong, it should start at 11,12 instead of 11,18 to match how they're drawn. So the bottom edge was maybe correct previously but the top was off, in the current state of this PR the hit area has grown off the bottom when it should've grown from the top. I'll fix that now.

ephphatha: Ok, that turned out to be annoying but I'll leave it for the future.

image
This is the hit area for a spell icon button (in vanilla Diablo/Hellfire and devilutionX). The behaviour is (now) not changed by this PR, but the margin between one button and the next is included in the hit area for the topmost button. It should really be a set of 37x38 pixel hit regions.

Source/DiabloUI/diabloui.cpp
obligaron: Sorry my bad.

Source/panels/spell_book.cpp
obligaron: I also noticed that the selection behavior in master is not 100% pixel correct, but that's a really minor issue.
That's why I thought the extra 5 pixel don't make a big difference anyway.
Thanks for looking again. 🙂

obligaron (approved):

glebm (approved):

PR #4735, opened 2022-06-18
Optimize `FormatInteger`

glebm:


Source/utils/format_int.cpp
```diff
@@ -0,0 +1,39 @@
...
+std::string FormatInteger(int n)
+{
+	constexpr size_t GroupSize = 3;
+
+	char buf[40];
+	const char *begin = buf;
+	size_t len = fmt::format_to(buf, FMT_COMPILE("{}"), n) - begin;
+
+	std::string out;
+	out.reserve(len + (len - 1) / GroupSize + (n < 0 ? 1 : 0));
```
ephphatha, line 20: Should this account for translated contexts where the separator is a multibyte character (or string...)?

glebm: Oh right, will fix

glebm: Fixed along with other improvements

ephphatha (approved):

PR #4736, opened 2022-06-18
Fix build with NONET and no exceptions

glebm:
Fixes the build with `-fno-rtti` and `-fno-exceptions`.

Building with these flags reduces the binary size by more than 200 KiB.


AJenbo (approved):
Nice, been wanting this 😄

AJenbo:
I guess we can update the Amiga build flags now

PR #4737, opened 2022-06-18
Define sizes using Size type in control.cpp/stash.cpp

ephphatha:
Also extracted repeated redefinitions of the PointInRectangleRangeIterator to a common constexpr variable in stash.cpp. There are two uses in plrctrls.cpp that I didn't think were worth adding cruft to the stash.h header to support, maybe those functions can move into stash.cpp in the future?

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -591,7 +591,7 @@ void InitControlPan()
 	sbookflag = false;
 
 	InitSpellBook();
-	pQLogCel = LoadCel("Data\\Quest.CEL", SidePanelSize.width);
+	pQLogCel = LoadCel("Data\\Quest.CEL", static_cast<uint16_t>(SidePanelSize.width));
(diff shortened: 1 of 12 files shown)


glebm (dismissed):

Source/control.cpp
```diff
@@ -591,7 +591,7 @@ void InitControlPan()
 	sbookflag = false;
 
 	InitSpellBook();
-	pQLogCel = LoadCel("Data\\Quest.CEL", SPANEL_WIDTH);
+	pQLogCel = LoadCel("Data\\Quest.CEL", SidePanelSize.width);
```
ephphatha, line 594:
```suggestion
	pQLogCel = LoadCel("Data\\Quest.CEL", static_cast<uint16_t>(SidePanelSize.width));
```

ephphatha:
Sorry, noticed a call to LoadCel I missed a cast for.

glebm (approved):

PR #4739, opened 2022-06-18
Apply MethodCase config to Rectangle::center

ephphatha:
The bit that makes the amiga build break 🙃


StephenCWills:
Looks like `-O3` fails, but `-O2` succeeds. I tried using to see if I could pinpoint which optimization is causing problems, but I'm completely stumped. According to the output of `-Q --help=optimizers`, `-O3` is equivalent to `-O2 -fgcse-after-reload -finline-functions -fipa-cp-clone -fpredictive-commoning -fsplit-paths -ftree-loop-distribute-patterns -ftree-partial-pre -funswitch-loops -fvect-cost-model=dynamic`, but the error does not occur unless I use `-O3`. 🤷

PR #4740, opened 2022-06-18
Index Minis by tile instead of by coordinate

AJenbo:
So turned out that not only does reducing indexing MICROS by dPiece reduce the table size from 392KiB to 43KiB, it also means that the .min data is only needed at load time, saving an additional 9-44KiB (depending on the level). And as a bonus it remove the need to recompute the table as the level changes.

I also moved the call to `SetDungeonMicros()` before the call to `LoadLvlGFX()` to reduce peak memory usage.

DPieceMicros could be removed from runtime if SetDungeonMicros() instead computed a new CEL where the tiles are not split in to smaller chunks (see `DrawCell()`).

Doing so should also mean that we can drop much of the specialized render code and probably result in better CEL compression and render performance. But I feel this might be a bit more then I am able to do.

... also changed `dPiece` from int to `uint16_t` to gain 25KiB.

And unloaded `pMegaTiles` (3KiB) after loading town since it is not needed there. At run time it is only needed for dungeons that can change shape via `pdungeon`:
- SklKng2.DUN
- Bonecha1.DUN
- Vile2.DUN
- Level 4 + Banner
- Level 5 + Blood
- Level 6 + Bone
- Level 7 + Blind
- Level 16

I felt that unloading it for town made sens since town generally has a higher memory load then other levels. But it could also be unloaded for all Caves safely and from other levels by excluding the above list.

These levels are also the only ones where `pdungeon` (less then 1.6KiB) is needed, but this is currently static data.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1020,16 +1020,6 @@ void AddChest(int i, int t)
 void ObjSetMicro(Point position, int pn)
 {
 	dPiece[position.x][position.y] = pn;
-	pn--;
-
-	int blocks = leveltype != DTYPE_HELL ? 10 : 16;
-
-	uint16_t *piece = &pLevelPieces[blocks * pn];
-	MICROS &micros = dpiece_defs_map_2[position.x][position.y];
-
-	for (int i = 0; i < blocks; i++) {
-		micros.mt[i] = SDL_SwapLE16(piece[blocks - 2 + (i & 1) - (i & 0xE)]);
-	}
 }
 
 void InitializeL1Door(Object &door)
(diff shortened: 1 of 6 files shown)


glebm (approved):

glebm:
Nice!!

Source/objects.cpp
```diff
@@ -1020,16 +1020,6 @@ void AddChest(int i, int t)
...
 	dPiece[position.x][position.y] = pn;
-	pn--;
-
-	int blocks = leveltype != DTYPE_HELL ? 10 : 16;
-
-	uint16_t *piece = &pLevelPieces[blocks * pn];
-	MICROS &micros = dpiece_defs_map_2[position.x][position.y];
-
-	for (int i = 0; i < blocks; i++) {
-		micros.mt[i] = SDL_SwapLE16(piece[blocks - 2 + (i & 1) - (i & 0xE)]);
```
galaxyhaxz, line 1031: This is one of the nicest cleanups from removing/gutting the micro map. You wonder what was running through their minds when they wrote this code :D

PR #4741, opened 2022-06-19
Reduce reliance on `bool done` for control flow, part 2 of x

vlolteanu:
This also fixes a bug introduced in part 1, where `PlaceUniqueMonsters` uses a monster type that is off by 1.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -482,15 +482,18 @@ void PlaceUniqueMonst(int uniqindex, int miniontype, int bosspacksize)
 	PrepareUniqueMonst(monster, uniqindex, miniontype, bosspacksize, uniqueMonsterData);
 }
 
+int GetMonsterTypeIndex(_monster_id type)
+{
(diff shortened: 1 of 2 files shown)


Source/monster.cpp
```diff
@@ -543,12 +541,14 @@ void PlaceUniqueMonsters()
...
-				break;
-		}
-		if (mt >= LevelMonsterTypeCount)
+		int mt = [&]() {
+			for (int i = 0; i < LevelMonsterTypeCount; i++) {
+				if (LevelMonsterTypes[mt].mtype == UniqueMonstersData[u].mtype)
+					return i;
+			}
+			return LevelMonsterTypeCount;
+		}();
```
AJenbo, line 550: This looks identical to the block in `AddMonsterType`, maybe make them in to named functions and take type as an input. `int GetMonsterTypeIndex(uint8_t type)`

AJenbo (approved):
Nice work :)

PR #4744, opened 2022-06-19
Fix wide-screen background paths

glebm:
These files are lowercase when in a bundled filesystem.


AJenbo (approved):
Grate, I had been wondering why they didn't show up on android

PR #4745, opened 2022-06-19
Fix building unit tests with msvc (SOLData needs DVL_API_FOR_TEST)

obligaron:
Different compiler behavior 🎉


glebm (approved):

AJenbo:
Yeah this is a bit tricky to keep track of from the other side 😅

obligaron:
I know. This also happens for me with the old gcc. 😁

AJenbo:
At least you hit the build bot and not other people 😅

PR #4746, opened 2022-06-19
Move embedded dialog graphics into the MPQ

glebm:
If we can't find `devilutionx.mpq`, we can't render a dialog either, because the MPQ contains the fonts.


qndel:
any reason to use cryptic filenames that are hard to decipher? :D

glebm:
These previously-embedded graphics have similar versions in `diabdat.mpq` (with a few differences, e.g. diabdat.mpq button has 15 frames, and diabdat mpq error dialog has redundant transparent columns/rows on the right and bottom.

The names are the `diabdat.mpq` names with a `dvl_` prefix.

qndel:
imo it's a good moment to rename them to something that would indicate what they mean/do :D
dvl_lrpopup.pcx
I'm assuming lr means ... light red? if it's the error, maybe we can name it dvl_red_error_popup? Otherwise I read most of these names and have no clue what they do and have to check in the code every single time xD

glebm:
But if we rename them it'll be less obvious which `diabdat.mpq` assets they correspond to.

The other assets are also named similarly to the underlying `diabdat.mpq` assets.

Let's leave renaming for a separate PR.

AJenbo:
They are either from the shareware or where embeded into the executable.

AJenbo (approved):

qndel:
> They are either from the shareware or where embeded into the ui dll.

didn't recognize names so thought it was something we added :P

PR #4747, opened 2022-06-19
Fix PCX `LoadFrameOffsets`

glebm:


qndel (approved):

julealgon:
Can someone provide details about what this issue was and when it was introduced?

glebm:
It was introduced when we switched to rendering PCX fonts directly (instead of converting them to another format at load time).

The issue was that the padding pixel for fonts with odd widths was skipped not once per row but once per frame when generating the frame offsets table.

PR #4750, opened 2022-06-19
Player's Character Statistics

k-bar:
This PR introduces Player's Character Statistics to the game as a structure ready to expand.

Statistics are saved to savegame mpq as separate statistics file. Thus don't break compatibility with vanilla that will just ignore it.

**Ready to expand.** Currently only few statistisc are added. But the structure is ready to add as many as you like.
Just implement new ones in the game logic, and add in `statistics.cpp` in `LoadStatisticsFromMap()` and `SaveStatisticsToMap()` - by analogy to the currently added ones. Add to `Statistics` struct if needed.

**Forward and backward compatible.**
Thanks to saving as key->value map it is totally safe to load savegame originating from newer version of game that introcuded more statistics.
If savegame is loaded by older version of game that will not support particular statistics - not supported statistics will survive and just be saved without any changes. Game will modifiy only statistics known to it. The rest will stay untouched.

Currently implemented statistics:
- Player (character) Death Count
- Ingame time
- Monster Kill Count support in MP games.

Only Player Death Count is displayed right now (hover exp bar). For more statistics some statistics panel is needed as UI part - yet to be done. But nothing prevents you from adding statistics that will be gathered and displayed layer.
Monster Kill Count is displayed as always - but this time also saved per MP character as well.
(small modification by me here is count only kills done by My Player - as bug fix).

I experimented with few only, but as you can imagine sky is the limit. You can count and store everything.
some ideas:
How many times character asked Farnham for gossip, how many arrows she shooted. How many times he blocked. What was highest damage he dealt, or maybe received. What was his priciest buy. How many stairs he took, or town portals from other players. What was his farthest teleport distance. How many games he created or joined. How many messages he typed in a chat, or maybe used debug commands? How many times he said "#$%^&". What is his favourite spell or gear? Which shrines did he took how many times? Or hit by trap. How many items he too from stash, and how many placed in? How many small yellow potion he quaffed. How many times he killed or resurrected other players.
How many meters character walked, swim or jump :wink:
This also enables achievments (like kill 6666 goats).
Take note that it is possible to use it not only as counters. You can let's say store previous character names as coma separated list in one of map position (when change-name feature will be implemented). And many more.
At some point maybe save to txt could be added, to print and plant on a fridge. Or even display on some statistics website hosted on server connected to DevilutionX.

If you want more ideas - go take a look at statistics in World of Warcraft (it's free to play as trial with character level limitations). There are literally hundreds if not thousands.
It is even better as you can inspect other player statistics there too (as all is stored on server).

### DIFF
--- a/Source/qol/statistics.cpp
+++ b/Source/qol/statistics.cpp
@@ -2,50 +2,51 @@
 
 namespace devilution {
 
-Statistics myPlayerStatistics;
-std::map<std::string, std::string> statisticsFile;
+Statistics MyPlayerStatistics;
+std::map<std::string, std::string> StatisticsFile;
 
-void SaveMonsterKillCount();
 void LoadMonsterKillCount();
+void SaveMonsterKillCount();
 std::string GetMonsterName(uint16_t monsterID);
 void AddMonsterWeapon(std::string &monsterName, uint16_t monsterID);
 
 void InitializePlayerStatistics()
 {
-	myPlayerStatistics = {};
+	MyPlayerStatistics = {};
 	std::fill_n(MonsterKillCounts, MAXMONSTERS, 0);
+	StatisticsFile.clear();
 }
 
 void LoadStatisticsFromMap()
 {
-	if (statisticsFile.find("deathCount") != statisticsFile.end())
-		myPlayerStatistics.deathCount = std::stoul(statisticsFile.at("deathCount"));
+	if (StatisticsFile.find("deathCount") != StatisticsFile.end())
+		MyPlayerStatistics.deathCount = std::stoul(StatisticsFile.at("deathCount"));
 
-	if (statisticsFile.find("ingameTime") != statisticsFile.end())
-		myPlayerStatistics.ingameTime = std::stoul(statisticsFile.at("ingameTime"));
+	if (StatisticsFile.find("ingameTime") != StatisticsFile.end())
+		MyPlayerStatistics.ingameTime = std::stoul(StatisticsFile.at("ingameTime"));
 	LoadMonsterKillCount();
 }
 
 void SaveStatisticsToMap()
 {
-	statisticsFile["deathCount"] = std::to_string(myPlayerStatistics.deathCount);
-	statisticsFile["ingameTime"] = std::to_string(myPlayerStatistics.ingameTime);
+	StatisticsFile["deathCount"] = std::to_string(MyPlayerStatistics.deathCount);
+	StatisticsFile["ingameTime"] = std::to_string(MyPlayerStatistics.ingameTime);
 	SaveMonsterKillCount();
 }
 
 void CalculateInGameTime()
 {
 	uint32_t ticksNow = SDL_GetTicks();
-	myPlayerStatistics.ingameTime += (ticksNow - myPlayerStatistics.ticksSubstrahend);
-	myPlayerStatistics.ticksSubstrahend = ticksNow;
+	MyPlayerStatistics.ingameTime += (ticksNow - MyPlayerStatistics.ticksSubtrahend);
+	MyPlayerStatistics.ticksSubtrahend = ticksNow;
 }
 
 void LoadMonsterKillCount()
 {
 	for (uint16_t monsterID = 0; monsterID < _monster_id::NUM_MTYPES; monsterID++) {
 		std::string monsterName = GetMonsterName(monsterID);
-		if (statisticsFile.find("killed." + monsterName) != statisticsFile.end())
-			MonsterKillCounts[monsterID] = std::stoi(statisticsFile.at("killed." + monsterName));
+		if (StatisticsFile.find("killed." + monsterName) != StatisticsFile.end())
+			MonsterKillCounts[monsterID] = std::stoi(StatisticsFile.at("killed." + monsterName));
 	}
 }
 
@@ -53,14 +54,15 @@ void SaveMonsterKillCount()
 {
 	for (uint16_t monsterID = 0; monsterID < _monster_id::NUM_MTYPES; monsterID++) {
 		std::string monsterName = GetMonsterName(monsterID);
-		statisticsFile["killed." + monsterName] = std::to_string(MonsterKillCounts[monsterID]);
+		StatisticsFile["killed." + monsterName] = std::to_string(MonsterKillCounts[monsterID]);
 	}
 }
 
 std::string GetMonsterName(uint16_t monsterID)
 {
 	std::string monsterName = MonstersData[monsterID].mName;
-	monsterName.erase(std::remove_if(monsterName.begin(), monsterName.end(), [](unsigned char x) { return std::isspace(x); }), monsterName.end());
+	std::string remChars = "- ";
+	monsterName.erase(std::remove_if(monsterName.begin(), monsterName.end(), [remChars](const std::string::value_type &x) { return remChars.find(x) != std::string::npos; }), monsterName.end());
 	AddMonsterWeapon(monsterName, monsterID);
 	return monsterName;
 }
--- a/Source/qol/statistics.cpp
+++ b/Source/qol/statistics.cpp
@@ -10,38 +10,34 @@ void LoadMonsterKillCount();
 std::string GetMonsterName(uint16_t monsterID);
 void AddMonsterWeapon(std::string &monsterName, uint16_t monsterID);
 
-void InitializePlayerStatistics(Player &player)
+void InitializePlayerStatistics()
 {
 	myPlayerStatistics = {};
-	player.statistics = &myPlayerStatistics;
 	std::fill_n(MonsterKillCounts, MAXMONSTERS, 0);
 }
 
 void LoadStatisticsFromMap()
 {
-	Player &myPlayer = *MyPlayer;
 	if (statisticsFile.find("deathCount") != statisticsFile.end())
-		myPlayer.statistics->deathCount = std::stoul(statisticsFile.at("deathCount"));
+		myPlayerStatistics.deathCount = std::stoul(statisticsFile.at("deathCount"));
 
 	if (statisticsFile.find("ingameTime") != statisticsFile.end())
-		myPlayer.statistics->ingameTime = std::stoul(statisticsFile.at("ingameTime"));
+		myPlayerStatistics.ingameTime = std::stoul(statisticsFile.at("ingameTime"));
 	LoadMonsterKillCount();
 }
 
 void SaveStatisticsToMap()
 {
-	Player &myPlayer = *MyPlayer;
-	statisticsFile["deathCount"] = std::to_string(myPlayer.statistics->deathCount);
-	statisticsFile["ingameTime"] = std::to_string(myPlayer.statistics->ingameTime);
+	statisticsFile["deathCount"] = std::to_string(myPlayerStatistics.deathCount);
+	statisticsFile["ingameTime"] = std::to_string(myPlayerStatistics.ingameTime);
 	SaveMonsterKillCount();
 }
 
 void CalculateInGameTime()
 {
-	Statistics *stats = MyPlayer->statistics;
 	uint64_t ticksNow = SDL_GetTicks64();
-	stats->ingameTime += (ticksNow - stats->ticksSubstrahend);
-	stats->ticksSubstrahend = ticksNow;
+	myPlayerStatistics.ingameTime += (ticksNow - myPlayerStatistics.ticksSubstrahend);
+	myPlayerStatistics.ticksSubstrahend = ticksNow;
 }
 
 void LoadMonsterKillCount()
--- a/Source/qol/statistics.cpp
+++ b/Source/qol/statistics.cpp
@@ -5,10 +5,16 @@ namespace devilution {
 Statistics myPlayerStatistics;
 std::map<std::string, std::string> statisticsFile;
 
+void SaveMonsterKillCount();
+void LoadMonsterKillCount();
+std::string GetMonsterName(uint16_t monsterID);
+void AddMonsterWeapon(std::string &monsterName, uint16_t monsterID);
+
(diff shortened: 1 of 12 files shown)


julealgon:
Question: have you considered implementing these as custom metrics using OpenTelemetry?

I'm asking because that is something I wanted to do in my mod later on in C#.

k-bar:
> Question: have you considered implementing these as custom metrics using OpenTelemetry?

No. To be honest I wasn't familiar with it before.
I tried to implement small and simple yet functional and expandable solution to start gathering some measures now.
I considered only some composite structure and serialization before putting into savegame file. (Still achievable by parsing properly named metrics).
In future development of this area it may be good to use some standarized metrics like OpenTelemetry. So each metric would have label, description, unit, type (counter, measure, observer) and so on. And that way it would be better to show them to external world (like publish on web server maybe). Oh and logs!. Currently I preferred to keep it low and small save locally - as a start of new area. And working proof with hopes to merge it soon. (We do not even have any proper UI for it yet.)

Source/qol/xpbar.cpp
```diff
@@ -141,6 +142,7 @@ bool CheckXPBarInfo()
 	AddPanelString(fmt::format(fmt::runtime(_("Experience: {:s}")), FormatInteger(player._pExperience)));
 	AddPanelString(fmt::format(fmt::runtime(_("Next Level: {:s}")), FormatInteger(ExpLvlsTbl[charLevel])));
 	AddPanelString(fmt::format(fmt::runtime(_("{:s} to Level {:d}")), FormatInteger(ExpLvlsTbl[charLevel] - player._pExperience), charLevel + 1));
+	AddPanelString(fmt::format(fmt::runtime(_("Deaths: {:d}")), myPlayerStatistics.deathCount));
```
AJenbo, line 145: MP only?

Source/qol/statistics.cpp
```diff
@@ -0,0 +1,97 @@
...
+	statisticsFile["deathCount"] = std::to_string(myPlayerStatistics.deathCount);
+	statisticsFile["ingameTime"] = std::to_string(myPlayerStatistics.ingameTime);
+	SaveMonsterKillCount();
+}
+
+void CalculateInGameTime()
+{
+	uint32_t ticksNow = SDL_GetTicks();
+	myPlayerStatistics.ingameTime += (ticksNow - myPlayerStatistics.ticksSubstrahend);
+	myPlayerStatistics.ticksSubstrahend = ticksNow;
```
AJenbo, line 40: subtrahend

Also should we take pause in to consideration?

Source/qol/statistics.cpp
```diff
@@ -0,0 +1,97 @@
...
+		std::string monsterName = GetMonsterName(monsterID);
+		statisticsFile["killed." + monsterName] = std::to_string(MonsterKillCounts[monsterID]);
+	}
+}
+
+std::string GetMonsterName(uint16_t monsterID)
+{
+	std::string monsterName = MonstersData[monsterID].mName;
+	monsterName.erase(std::remove_if(monsterName.begin(), monsterName.end(), [](unsigned char x) { return std::isspace(x); }), monsterName.end());
+	AddMonsterWeapon(monsterName, monsterID);
```
AJenbo, line 64: This is not taking translations in to consideration, it's probably better have the full name in the switch rather then try to edit it, that way you can also return a string_view instead.

Source/qol/statistics.cpp
```diff
@@ -0,0 +1,97 @@
+#include "qol/statistics.h"
+
+namespace devilution {
+
+Statistics myPlayerStatistics;
+std::map<std::string, std::string> statisticsFile;
```
AJenbo, line 6: Globals starts with uppercase

Source/qol/statistics.cpp
AJenbo: Hm this might be for a text based file format, that feel a bit foreign compared to how all other data is saved. And also I'm not sure it's a good idea compared to using there id as the identifier.

k-bar: > This is not taking translations in to consideration, it's probably better have the full name in the switch rather then try to edit it, that way you can also return a string_view instead.

This should not take translations into considerations. I am building statistic row name here (key) for `statisticsFile` map, language independent. Its purpose isn't to display its name. But to be just english internal name for particular statistic row.
currently MonsterKillCounts is an array that stores killed monsters in order based on their ID.
I am taking this ID here, get the name from `MonstersData` correlating to it. And since no all names are unique I am making them unique.
1. First I trim the name from spaces.
2. AddMonsterWeapon function with switch is responsible only to make some non-unique names unique.

i.e. there are in `MonstersData` 2 positions with the same monster name:
- `MT_GGOATMC` `"Night Clan"` and 
- `MT_GGOATBW` `"Night Clan"`
I am making those statistics *key* (in map) name: 
- `killed.NightClan`
- `killed.NightClanArcher`

>Hm this might be for a text based file format, that feel a bit foreign compared to how all other data is saved. And also I'm not sure it's a good idea compared to using there id as the identifier.

yes `statisticsFile` is text base. It is <string, string> map.
keys in there looks like
`killed.NightClan`
`DeathCount`
etc.
and I later save it to file as `"killed.NightClan=666\n"` line - see `SaveStatistics` in loadsave.cpp
I am not using id as identifier at all.

Source/qol/statistics.cpp
k-bar: > subtrahend

yeah :)

>Also should we take pause in to consideration?

idk, should we?

Source/qol/xpbar.cpp
AJenbo: Yeah that was my point, don't show in SP as it doesn't make sens there.

Source/qol/statistics.cpp
AJenbo: Preferably not

Source/qol/statistics.cpp
AJenbo: Considering that we already have unique ids, and use them as a basis for the generated strings, this seams pointless/wistful.

Source/qol/statistics.cpp
k-bar: Ok, I will not take pause into consideration.

Source/qol/statistics.cpp
k-bar: > Considering that we already have unique ids, and use them as a basis for the generated strings, this seams pointless/wistful.

I wanted each key name to be meaningful. Also as follow-up to this PR I thought about adding counting unique monsters killed (like Lazarus) and handle them like that (`killed.BlackJade`). They are not counted separately right now by the game. They are not counted at all. And bosses have no unique identifier... They have only order in ` UniqueMonstersData`. So this would be more consistent to name all monster kills in the same way.

AJenbo: Unique also have an index :) I'm a bit unsure about using the display name to generate an index as it was not intended for it, and may be subject to change.

qndel: why?

DakkJaniels: because BlackJade is not necessarily a monster in a mod? Why not just use the index, so the key is not dependent on what the monster is (or was previously) called?

k-bar: It is not hardcoded. It takes name by index from monstersdat tables. Your mod will have different monster and unique monster names and statistics keyes saved in your mod will be named after the names of your monsters.
What's hardcoded here is only few exceptions in d1 that have several couples of monsters using the same name.

DakkJaniels: I think I misunderstood based on the conversation, since it does seem like it's going to pull the current monster name to generate the key, although the AddMonsterWeapon thing would be applicable to what I was saying - would just need to make sure to update the labels when modding I guess.

k-bar: Actually I think that all literals could be exported to translations. English too ;)
Or alternatively - fix typos only in english translations. But i guess there is no such place. Yet! :grin:

k-bar: Ok I have an idea.
How about change current indentifiers in enum `_monster_id` from like MT_GGOATBW` to a new ones `NightClanArcher` ?

AJenbo: Renaming `MT_GGOATBW` and other old names to nice descriptive names like `NightClanArcher` would be much appreciated :)

AJenbo: > yes, UniqueMonstersData is an array so yes there's index, but no unique id like normal monsters have and I could use.

Uh didn't even realize that :sweat:

StephenCWills: > > yes, UniqueMonstersData is an array so yes there's index, but no unique id like normal monsters have and I could use.
> 
> Uh didn't even realize that 😓

We can expand the `UMT_*` enum easily enough.

k-bar: > Renaming `MT_GGOATBW` and other old names to nice descriptive names like `NightClanArcher` would be much appreciated :)

I realized that this still doesn't solve the issue.
There is no clean way of converting enum to string without some convoluted tricks.
This is 2022 and reflection isn't in C++ yet. Its proposal is postponed again to C++23.
You can go here:
https://stackoverflow.com/questions/28828957/enum-to-string-in-modern-c11-c14-c17-and-future-c20
and choose which way would be the best fitting the project.
Amongst many solutions there are prominent:
- enum redeclaration by macro  (or x macro technique)
- using another lib like boost or magic enum or others
- enum generation from external file (even more subject to change)
- just retype all literals again in case switch or other way (which means double typing hardcoded strings).

At this point I would stay with current commited solution and wait for C++23 reflection. But pick one from the link if you really think it is better.

Source/qol/statistics.cpp
k-bar: just to be sure by not taking into consideration you mean:
ignore its existance and influence on game time. Or rather - not count game time when paused?

Source/qol/statistics.cpp
AJenbo: I was suggesting not using strings, since we don't do so for any other data type

StephenCWills: > * just retype all literals again in case switch or other way (which means double typing hardcoded strings).

We already use this strategy for other enums.

https://github.com/diasurgical/devilutionX/blob/39afeaf8b1a474cd662d5975994cb5cb939597f8/Source/dvlnet/packet.cpp#L54-L80

> At this point I would stay with current commited solution and wait for C++23 reflection. But pick one from the link if you really think it is better.

At the very least, we should avoid using `std::isspace()`. This quote is from @glebm on Discord in response to a bug we found recently from using `isblank()` in some parsing code.

> Don't use any of the `is` functions please. They're locale-dependent and basically super-broken because of that

StephenCWills: > I was suggesting not using strings, since we don't do so for any other data type

@k-bar We can define an enum that maps each statistic to an integer ID and use that as a key for the map. We can also set high bits to create categories for monsters and unique monsters.

What was the purpose for selecting `string` as the value type? Aren't basically all statistics numbers?

k-bar: > What was the purpose for selecting `string` as the value type? Aren't basically all statistics numbers?

Yes, most of the statistics that come to mind are numerical. And most of them will be counters. However, as I mentioned in the description of PR - I wanted the solution to be ready to write also non-numeric values. In addition to the example from the PR description, you can e.g. imagine some last string value (e.g. the name of the game, or the name of the last killed player, whatever).

The decision to use a map with any key in the form of a string was dictated by the need to have the save and read statistics disconnected from the game version.
This way, if a newer version of the game introduces a new statistic, the savegame from that version still works in the previous version, the statistics supported by the previous version are updated, and new, unknown ones are read but unused and not lost. They are stored in the map and saved without loss. The size of the map is also not predetermined in this way.
Such a design seemed to me the safest from data loss (overwriting, unsaved) and flexible in terms of what kind of statistics can be added in the future. It doesn't matter what kind. New keys will just be added to the map.
In case of resignation from the map solution string, string - first of all, the possibility of writing non-string statistics (for value) is given up. Secondly, the more flexible solution ensuring the compatibility of save files (key) is abandoned. (eg how many higher bits should be allocated to types of statistics, what if they run out etc. I assume that what you mentioned would also be a map only with a numeric key, how to save and parse it?). String keys seem like a much simpler solution.

k-bar: > We already use this strategy for other enums.

If this is correct approach for the project, let me know, and I will make proper readjustment.
I am not sure if @DakkJaniels would be happy :)

Source/qol/statistics.cpp
k-bar: I was wondering shouldn't I just access those 2 by pointer placed in `Player` struct...

Iniquitatis: @AJenbo Not quite on topic, but what led to this decision? In various C(++) projects, I hardly ever see globals starting with an uppercase letter, not to mention using the different case style from the rest of variables. Much more often they are just prefixed somehow (`g`, `g_`) and using the same case style as the members, locals, etc.

kphoenix137:
On behalf of the PVP community, we would like Players kills counter and Player deaths counter with Kill/Death Ratio

qndel:
> On behalf of the PVP community, we would like Players kills counter and Player deaths counter with Kill/Death Ratio. Perhaps disable incrementing the counter between players that share the same IP address? (To avoid people cheesing their stats with multiple instances of the game open). And it would be nice if the player kill counter would only increment if the target had player hostile set, and likewise, your player death counter would only increment if someone killed you while you had player hostile set.

that IP idea is so dumb, what will it achieve? dudes just farming KD with another person derp, don't bother

AJenbo:
Considering that we don't even know the IP well enough to display it on the host screen I think we can write off that idea anyway.

PR #4754, opened 2022-06-19
Make `Players` a vector

glebm:
Each Player struct is approximately 21 KiB, so this saves ~63 KiB in single-player.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -328,7 +328,7 @@ void ResetTalkMsg()
 
 	uint32_t pmask = 0;
 
-	for (int i = 0; i < MAX_PLRS; i++) {
+	for (size_t i = 0; i < Players.size(); i++) {
 		if (WhisperList[i])
 			pmask |= 1 << i;
 	}
@@ -1154,7 +1154,7 @@ void DrawTalkPan(const Surface &out)
 
 	x += 46;
 	int talkBtn = 0;
-	for (int i = 0; i < 4; i++) {
+	for (size_t i = 0; i < Players.size(); i++) {
 		Player &player = Players[i];
 		if (&player == MyPlayer)
 			continue;
@@ -1221,13 +1221,13 @@ void control_release_talk_btn()
 
 	int off = (MousePosition.y - (69 + mainPanelPosition.y)) / 18;
 
-	int p = 0;
-	for (; p < MAX_PLRS && off != -1; p++) {
-		if (p != MyPlayerId)
+	size_t playerId = 0;
+	for (; playerId < Players.size() && off != -1; ++playerId) {
+		if (playerId != MyPlayerId)
 			off--;
 	}
-	if (p <= MAX_PLRS)
-		WhisperList[p - 1] = !WhisperList[p - 1];
+	if (playerId > 0 && playerId <= Players.size())
+		WhisperList[playerId - 1] = !WhisperList[playerId - 1];
 }
 
 void control_type_message()
--- a/Source/cursor.cpp
+++ b/Source/cursor.cpp
@@ -423,52 +423,52 @@ void CheckCursMove()
 	if (leveltype != DTYPE_TOWN) {
 		if (pcurstemp != -1) {
 			if (!flipflag && mx + 2 < MAXDUNX && my + 1 < MAXDUNY && dMonster[mx + 2][my + 1] != 0 && IsTileLit({ mx + 2, my + 1 })) {
-				int mi = abs(dMonster[mx + 2][my + 1]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().selectionType & 4) != 0) {
+				const uint16_t monsterId = abs(dMonster[mx + 2][my + 1]) - 1;
+				if (monsterId == pcurstemp && Monsters[monsterId].hitPoints >> 6 > 0 && (Monsters[monsterId].data().selectionType & 4) != 0) {
 					cursPosition = Point { mx, my } + Displacement { 2, 1 };
-					pcursmonst = mi;
+					pcursmonst = monsterId;
 				}
 			}
 			if (flipflag && mx + 1 < MAXDUNX && my + 2 < MAXDUNY && dMonster[mx + 1][my + 2] != 0 && IsTileLit({ mx + 1, my + 2 })) {
-				int mi = abs(dMonster[mx + 1][my + 2]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().selectionType & 4) != 0) {
+				const uint16_t monsterId = abs(dMonster[mx + 1][my + 2]) - 1;
+				if (monsterId == pcurstemp && Monsters[monsterId].hitPoints >> 6 > 0 && (Monsters[monsterId].data().selectionType & 4) != 0) {
 					cursPosition = Point { mx, my } + Displacement { 1, 2 };
-					pcursmonst = mi;
+					pcursmonst = monsterId;
 				}
 			}
 			if (mx + 2 < MAXDUNX && my + 2 < MAXDUNY && dMonster[mx + 2][my + 2] != 0 && IsTileLit({ mx + 2, my + 2 })) {
-				int mi = abs(dMonster[mx + 2][my + 2]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().selectionType & 4) != 0) {
+				const uint16_t monsterId = abs(dMonster[mx + 2][my + 2]) - 1;
+				if (monsterId == pcurstemp && Monsters[monsterId].hitPoints >> 6 > 0 && (Monsters[monsterId].data().selectionType & 4) != 0) {
 					cursPosition = Point { mx, my } + Displacement { 2, 2 };
-					pcursmonst = mi;
+					pcursmonst = monsterId;
 				}
 			}
 			if (mx + 1 < MAXDUNX && !flipflag && dMonster[mx + 1][my] != 0 && IsTileLit({ mx + 1, my })) {
-				int mi = abs(dMonster[mx + 1][my]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().selectionType & 2) != 0) {
+				const uint16_t monsterId = abs(dMonster[mx + 1][my]) - 1;
+				if (monsterId == pcurstemp && Monsters[monsterId].hitPoints >> 6 > 0 && (Monsters[monsterId].data().selectionType & 2) != 0) {
 					cursPosition = Point { mx, my } + Displacement { 1, 0 };
-					pcursmonst = mi;
+					pcursmonst = monsterId;
 				}
 			}
 			if (my + 1 < MAXDUNY && flipflag && dMonster[mx][my + 1] != 0 && IsTileLit({ mx, my + 1 })) {
-				int mi = abs(dMonster[mx][my + 1]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().selectionType & 2) != 0) {
+				const uint16_t monsterId = abs(dMonster[mx][my + 1]) - 1;
+				if (monsterId == pcurstemp && Monsters[monsterId].hitPoints >> 6 > 0 && (Monsters[monsterId].data().selectionType & 2) != 0) {
 					cursPosition = Point { mx, my } + Displacement { 0, 1 };
-					pcursmonst = mi;
+					pcursmonst = monsterId;
 				}
 			}
 			if (dMonster[mx][my] != 0 && IsTileLit({ mx, my })) {
-				int mi = abs(dMonster[mx][my]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().selectionType & 1) != 0) {
+				const uint16_t monsterId = abs(dMonster[mx][my]) - 1;
+				if (monsterId == pcurstemp && Monsters[monsterId].hitPoints >> 6 > 0 && (Monsters[monsterId].data().selectionType & 1) != 0) {
 					cursPosition = { mx, my };
-					pcursmonst = mi;
+					pcursmonst = monsterId;
 				}
 			}
 			if (mx + 1 < MAXDUNX && my + 1 < MAXDUNY && dMonster[mx + 1][my + 1] != 0 && IsTileLit({ mx + 1, my + 1 })) {
-				int mi = abs(dMonster[mx + 1][my + 1]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().selectionType & 2) != 0) {
+				const uint16_t monsterId = abs(dMonster[mx + 1][my + 1]) - 1;
+				if (monsterId == pcurstemp && Monsters[monsterId].hitPoints >> 6 > 0 && (Monsters[monsterId].data().selectionType & 2) != 0) {
 					cursPosition = Point { mx, my } + Displacement { 1, 1 };
-					pcursmonst = mi;
+					pcursmonst = monsterId;
 				}
 			}
 			if (pcursmonst != -1 && (Monsters[pcursmonst].flags & MFLAG_HIDDEN) != 0) {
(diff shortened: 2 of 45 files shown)


Source/control.cpp
```diff
@@ -1132,7 +1132,7 @@ void DrawTalkPan(const Surface &out)
 
 	x += 46;
 	int talkBtn = 0;
-	for (int i = 0; i < 4; i++) {
+	for (size_t i = 0; i < Players.size(); i++) {
```
AJenbo, line 1135: This is fixed to the number of frames in TalkButton, this change will leave to a crash if player size is incrased past 4.
```suggestion
	for (size_t i = 0; i < 4; i++) {
```

test/path_test.cpp
```diff
@@ -110,7 +110,7 @@ void CheckPath(Point startPosition, Point destinationPosition, std::vector<int8_
 	// Die early if the wrong path length is returned as we don't want to read oob in expectedSteps
 	ASSERT_LE(pathLength, expectedSteps.size()) << "Path is longer than expected.";
 
-	for (auto i = 0; i < pathLength; i++) {
+	for (size_t i = 0; i < static_cast<size_t>(pathLength); i++) {
```
AJenbo, line 113: Why not int and avoid the case?

glebm: Done

Source/control.cpp
glebm: Done

Source/towners.cpp
```diff
@@ -808,7 +808,7 @@ bool IsTownerPresent(_talker_id npc)
 	case TOWN_COWFARM:
 		return gbIsHellfire && sgGameInitInfo.bCowQuest != 0;
 	case TOWN_GIRL:
-		return gbIsHellfire && sgGameInitInfo.bTheoQuest != 0 && Players->_pLvlVisited[17] && Quests[Q_GIRL]._qactive != QUEST_DONE;
+		return gbIsHellfire && sgGameInitInfo.bTheoQuest != 0 && MyPlayer->_pLvlVisited[17] && Quests[Q_GIRL]._qactive != QUEST_DONE;
```
AJenbo, line 811: wow nice find :D

Source/cursor.cpp
```diff
@@ -540,27 +540,27 @@ void CheckCursMove()
 	if (pcursmonst == -1) {
 		if (!flipflag && mx + 1 < MAXDUNX && dPlayer[mx + 1][my] != 0) {
 			int8_t bv = abs(dPlayer[mx + 1][my]) - 1;
```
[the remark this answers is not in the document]
glebm: It's `abs(...) - 1` so I'm guessing it could be negative?

NiteKat:
There are a large number of places the code is changed, so I'm having trouble telling, but does the Player vector resize in Multiplayer, or does it get set to 4 slots when in Multiplayer mode without regard for the number of active players?

I'm asking because there could be side effects on monster behavior that differ from today/vanilla if the vector is cleaned up based on number of active players, since monster targeting partially relies on player index (unless that has changed in DevX and I missed it).

glebm:
For multiplayer it always sets the size to 4 for now because I was concerned about the same thing (golem indices in particular).

AJenbo:
It would be nice if we can set it to 1 during loopback (it does know how many max players is for any connection type)

Source/cursor.cpp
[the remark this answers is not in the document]
glebm: Done

glebm:
> It would be nice if we can set it to 1 during loopback (it does know how many max players is for any connection type)

I think that may require further investigation -- I'll try it in a separate PR once this is merged

AJenbo:
Yeah, was just thinking out loud :)

Source/monster.cpp
```diff
@@ -3745,7 +3745,7 @@ void monster_some_crypt()
 void InitGolems()
 {
 	if (!setlevel) {
-		for (int i = 0; i < MAX_PLRS; i++)
+		for (size_t i = 0; i < Players.size(); i++)
```
obligaron, line 3748: Dis this break vanilla compatibility? 
With this change we only add one golem in single player.
That means the first normal monster could be at index 1.
But vanilla often checks for golems with ``if (monsterId < MAX_PLRS)``

AJenbo: Vanilla also broke it self since it would spawn 8 golems for set levels :sweat_smile:

obligaron: Thanks to qndel this is fixed. 👍 #2822
But this vanilla bug is different.
It simply creates more golems at their holding cell.
That creates garbage but don't harm or?

qndel: I'd keep the index at 4 like before, otherwise what happens when we load vanilla SP save in devx? golems 1-3 will be treated as monsters? What if you load devx save in vanilla? monsters 1-3 will be treatd as golems? :D

NiteKat: Wouldn't loading vanilla SP save in DevX with 1-3 as monsters instead of golems not matter, because 1-3 would never be spawned golems given there's no player ever in slots 1-3 in single player?

That said, loading a DevX save with monsters 1-3 existing could maybe introduce interesting behavior in vanilla. xD

AJenbo:
I'm going over the various parts of the code that leads to all the casting and cleaning it up so that it can be avoided. I'll rebase as the the cleanups are merged. First one his here: https://github.com/diasurgical/devilutionX/pull/4800

AJenbo:
Since the first one was merged so quickly I only had time to make a small follow up :) https://github.com/diasurgical/devilutionX/pull/4802

(I have a larger change set that cleans up about half of all the cases added in this PR)

AJenbo:
Here is another nice chunk of cleanups in player comparisons :)

https://github.com/diasurgical/devilutionX/pull/4803

AJenbo:
One more ready: https://github.com/diasurgical/devilutionX/pull/4805

Also I think I have everything cleaned up now and just need to split it in to individual PRs, so hopefully I can have it all ready some time tomorrow.

AJenbo:
Alright here is the batch: https://github.com/diasurgical/devilutionX/pull/4808 https://github.com/diasurgical/devilutionX/pull/4810 https://github.com/diasurgical/devilutionX/pull/4811

AJenbo:
After some more work this one now also helps out here https://github.com/diasurgical/devilutionX/pull/4813

Source/player.cpp
```diff
@@ -3285,8 +3276,8 @@ void StartWarpLvl(int pnum, int pidx)
 
 void ProcessPlayers()
 {
-	if ((DWORD)MyPlayerId >= MAX_PLRS) {
-		app_fatal("ProcessPlayers: illegal player %i", MyPlayerId);
+	if (MyPlayerId >= MAX_PLRS) {
```
AJenbo, line 3279: Why not compare it to `Players.size()`?

glebm: Yeah this should be `Players.size()`

Source/monster.cpp
glebm: Yeah let's revert this to `MAX_PLRS` just in case

Source/player.cpp
AJenbo: Ok looks like there is still quite a bit of that around. I'll try to update that later today.

Source/monster.cpp
glebm: This is now `MAX_PLAYERS` again

glebm:
Rebased

glebm:
Rebased again but let's merge #4976 first to avoid some mixed sign comparison warnings.

AJenbo:
When running the demo graphically things seams to workout correctly :thinking:

AJenbo:
```
/usr/include/c++/11/bits/stl_vector.h:1046:34: runtime error: reference binding to null pointer of type 'struct value_type'
/home/ajenbo/code/devilutionX/Source/pfile.cpp:342:31: runtime error: reference binding to null pointer of type 'struct Player'
/home/ajenbo/code/devilutionX/Source/pfile.cpp:344:15: runtime error: member call on null pointer of type 'struct Player'
/home/ajenbo/code/devilutionX/Source/player.h:217:10: runtime error: member access within null pointer of type 'struct Player'
AddressSanitizer:DEADLYSIGNAL
=================================================================
==256049==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f7d6e023ea2 bp 0x7ffed75b6ab0 sp 0x7ffed75b6a70 T0)
==256049==The signal is caused by a WRITE memory access.
==256049==Hint: address points to the zero page.
    #0 0x7f7d6e023ea2 in devilution::Player::operator=(devilution::Player&&) /home/ajenbo/code/devilutionX/Source/player.h:217
    #1 0x7f7d6e020c77 in devilution::pfile_ui_set_hero_infos(bool (*)(devilution::_uiheroinfo*)) /home/ajenbo/code/devilutionX/Source/pfile.cpp:344
    #2 0x560970af4da2 in RunTimedemo /home/ajenbo/code/devilutionX/test/timedemo_test.cpp:46
    #3 0x560970af5acb in Timedemo_WarriorLevel1to2_Test::TestBody() /home/ajenbo/code/devilutionX/test/timedemo_test.cpp:65
    #4 0x560970b2f81e in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (/home/ajenbo/code/devilutionX/build/timedemo_test+0x5b81e)
    #5 0x560970b235b5 in testing::Test::Run() (/home/ajenbo/code/devilutionX/build/timedemo_test+0x4f5b5)
    #6 0x560970b23734 in testing::TestInfo::Run() (/home/ajenbo/code/devilutionX/build/timedemo_test+0x4f734)
    #7 0x560970b23ce8 in testing::TestSuite::Run() (/home/ajenbo/code/devilutionX/build/timedemo_test+0x4fce8)
    #8 0x560970b243ee in testing::internal::UnitTestImpl::RunAllTests() (/home/ajenbo/code/devilutionX/build/timedemo_test+0x503ee)
    #9 0x560970b2fde6 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (/home/ajenbo/code/devilutionX/build/timedemo_test+0x5bde6)
    #10 0x560970b237fb in testing::UnitTest::Run() (/home/ajenbo/code/devilutionX/build/timedemo_test+0x4f7fb)
    #11 0x560970afe37e in RUN_ALL_TESTS() /usr/include/gtest/gtest.h:2490
    #12 0x560970afe151 in main /home/ajenbo/code/devilutionX/test/main.cpp:17
    #13 0x7f7d6ad57d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #14 0x7f7d6ad57e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #15 0x560970af4534 in _start (/home/ajenbo/code/devilutionX/build/timedemo_test+0x20534)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/ajenbo/code/devilutionX/Source/player.h:217 in devilution::Player::operator=(devilution::Player&&)
==256049==ABORTING
ajenbo@L4:~/code/devilutionX/build$ 

```

AJenbo:
It looks like things are actually running correctly but some state is being wiped before the test makes it's comparison.

If I disable headless mode in the test:
```cpp
	HeadlessMode = false;
	init_create_window();
```
I get the following in the console:
```
INFO: 4000 frames, 30.80 seconds: 129.9 fps
INFO: Timedemo: Same outcome as inital run. :)  <--- everything good in the appliction
test/timedemo_test.cpp:62: Failure <--- subsequent call to make the same comparision in the test fails
Expected equality of these values:
  pfile_compare_hero_demo(demoNumber)
    Which is: 4-byte object <02-00 00-00>
  HeroCompareResult::Same
    Which is: 4-byte object <01-00 00-00>
```

Source/multi.cpp
```diff
@@ -708,6 +712,8 @@ void NetClose()
 	SNetLeaveGame(3);
 	if (gbIsMultiplayer)
 		SDL_Delay(2000);
+	Players.clear();
```
AJenbo, line 715: This is the cause of the failing test. The player state is being wiped before the test makes the comparison between game state and reference save.

Source/multi.cpp
```diff
@@ -201,16 +201,16 @@ void MonsterSeeds()
 
 void HandleTurnUpperBit(int pnum)
 {
-	int i;
+	size_t i;
```
AJenbo, line 204: `i` should be `size_t`

Source/multi.cpp
```diff
@@ -530,10 +534,10 @@ void multi_send_msg_packet(uint32_t pmask, const byte *data, size_t size)
 
 void multi_msg_countdown()
 {
-	for (int i = 0; i < MAX_PLRS; i++) {
+	for (size_t i = 0; i < Players.size(); i++) {
 		if ((player_state[i] & PS_TURN_ARRIVED) != 0) {
 			if (gdwMsgLenTbl[i] == sizeof(int32_t))
-				ParseTurn(i, *(int32_t *)glpMsgTbl[i]);
+				ParseTurn(static_cast<int>(i), *(int32_t *)glpMsgTbl[i]);
```
AJenbo, line 540: `ParseTurn` should take `size_t`

Source/portal.cpp
```diff
@@ -152,7 +152,7 @@ void GetPortalLevel()
 		leveltype = Portals[portalindex].ltype;
 	}
 
-	if (portalindex == MyPlayerId) {
+	if (portalindex == static_cast<int>(MyPlayerId)) {
```
AJenbo, line 155: Make portalindex size_t

Source/player.cpp
```diff
@@ -2808,7 +2799,7 @@ void InitPlayer(Player &player, bool firstTime)
 	player._pNextExper = ExpLvlsTbl[player._pLevel];
 	player._pInvincible = false;
 
-	if (&player == &myPlayer) {
+	if (&player == MyPlayer) {
```
AJenbo, line 2802: ?

Source/monster.cpp
```diff
@@ -4140,7 +4140,7 @@ void GolumAi(int monsterId)
 
 void DeleteMonsterList()
 {
-	for (int i = 0; i < MAX_PLRS; i++) {
+	for (size_t i = 0; i < Players.size(); i++) {
```
AJenbo, line 4143: Should this be kept as MAX_PLRS since it works on Golems? Currently it should match but if we resize when a player drops it might not.
```suggestion
	for (size_t i = 0; i < MAX_PLRS; i++) {
```

Source/cursor.cpp
```diff
@@ -540,58 +540,58 @@ void CheckCursMove()
...
 			}
 		}
 		if (pcurs == CURSOR_RESURRECT) {
 			for (int xx = -1; xx < 2; xx++) {
 				for (int yy = -1; yy < 2; yy++) {
 					if (TileContainsDeadPlayer({ mx + xx, my + yy })) {
-						for (int i = 0; i < MAX_PLRS; i++) {
-							const Player &player = Players[i];
+						for (size_t playerId = 0; playerId < Players.size(); ++playerId) {
+							const Player &player = Players[playerId];
```
AJenbo, line 579: Could loop players instead of ids

Source/cursor.cpp
```diff
@@ -540,58 +540,58 @@ void CheckCursMove()
...
+			if (playerId != MyPlayerId) {
 				cursPosition = { mx, my };
-				pcursplr = bv;
+				pcursplr = static_cast<int8_t>(playerId);
 			}
 		}
 		if (TileContainsDeadPlayer({ mx, my })) {
-			for (int i = 0; i < MAX_PLRS; i++) {
-				const Player &player = Players[i];
+			for (size_t playerId = 0; playerId < Players.size(); ++playerId) {
```
AJenbo, line 566: Could loop players instead of ids

Source/multi.cpp
```diff
@@ -415,6 +416,8 @@ void UnregisterNetEventHandlers()
 
 bool InitSingle(GameData *gameData)
 {
+	Players.emplace_back();
```
AJenbo, line 419: Should this be Players.resize(1); to avoid issues with it no longer being cleared during demo playback?

glebm: Yeah I guess so

Source/sync.cpp
```diff
@@ -295,7 +295,7 @@ uint32_t OnSyncData(const TCmd *pCmd, int pnum)
 	if (gbBufferMsgs == 1) {
```
AJenbo, line 295: Change OnSyncData to take size_t

Source/spells.cpp
```diff
@@ -259,7 +259,7 @@ void CastSpell(int id, spell_id spl, int sx, int sy, int dx, int dy, int spllvl)
 
 void DoResurrect(int pnum, uint16_t rid)
```
AJenbo, line 260:
```suggestion
void DoResurrect(size_t pnum, uint16_t rid)
```

Source/multi.cpp
```diff
@@ -659,7 +664,7 @@ void multi_process_network_packets()
 
 void multi_send_zero_packet(int pnum, _cmd_id bCmd, const byte *data, size_t size)
 {
-	assert(pnum != MyPlayerId);
+	assert(pnum != static_cast<int>(MyPlayerId));
```
AJenbo, line 667: change this and DThreadPkt to size_t

Source/multi.cpp
```diff
@@ -598,7 +603,7 @@ void multi_process_network_packets()
 		ClearPlayerLeftState();
 		if (dwMsgSize < sizeof(TPktHdr))
 			continue;
-		if (dwID < 0 || dwID >= MAX_PLRS)
+		if (static_cast<size_t>(dwID) >= Players.size())
```
AJenbo, line 606: The underlying type being used in SNetReceiveMessage is uint8_t, so lets change dwID to that as well and set the default to max.

Source/multi.cpp
```diff
@@ -449,7 +454,7 @@ bool InitMulti(GameData *gameData)
 		gbSelectProvider = true;
 	}
 
-	if (playerId < 0 || playerId >= MAX_PLRS) {
+	if (static_cast<size_t>(playerId) >= Players.size()) {
```
AJenbo, line 457: underlying type is uint8_t

Source/engine/render/scrollrt.cpp
```diff
@@ -867,7 +867,7 @@ void DrawDungeon(const Surface &out, Point tilePosition, Point targetBufferPosit
 		DrawDeadPlayer(out, tilePosition, targetBufferPosition);
 	}
 	int8_t playerId = dPlayer[tilePosition.x][tilePosition.y];
-	if (playerId > 0 && playerId <= MAX_PLRS) {
+	if (static_cast<size_t>(playerId - 1) <= Players.size()) {
```
AJenbo, line 870: This will start to render the player at both the leaving and entering position. Only positive numbers should be accepted here.

glebm: This does accept only the positive numbers. static_cast<size_t>(-1) is a huge number.

However, the condition should be:

```diff
-	if (static_cast<size_t>(playerId - 1) <= Players.size()) {
+	if (static_cast<size_t>(playerId - 1) < Players.size()) {
```

Source/multi.cpp
AJenbo: On secound thought I'm not so sure about this one

Source/msg.cpp
```diff
@@ -2642,7 +2642,7 @@ void NetSendCmdGolem(uint8_t mx, uint8_t my, Direction dir, uint8_t menemy, int
 
 void NetSendCmdLoc(int playerId, bool bHiPri, _cmd_id bCmd, Point position)
```
AJenbo, line 2643: check if playerId makes more sens as a size_t

Source/msg.cpp
```diff
@@ -1461,15 +1461,15 @@ size_t OnKnockback(const TCmd *pCmd, int pnum)
...
 {
 	const auto &message = *reinterpret_cast<const TCmdParam1 *>(pCmd);
 
 	if (gbBufferMsgs == 1) {
 		SendPacket(pnum, &message, sizeof(message));
-	} else if (message.wParam1 < MAX_PLRS) {
+	} else if (message.wParam1 < Players.size()) {
 		DoResurrect(pnum, Players[message.wParam1]);
-		if (pnum == MyPlayerId)
+		if (static_cast<size_t>(pnum) == MyPlayerId)
```
AJenbo, line 1472: it's already size_t
```suggestion
		if (pnum == MyPlayerId)
```

AJenbo (approved):

PR #4755, opened 2022-06-19
Make path finding index-based

glebm:
Reduces the size of the `PathNodes` array from 28 KiB to just 8 KiB.

Also reduces the size of `pnode_tblptr` from `300 * sizeof(void *)` to `300 * 2` bytes.

### DIFF
--- a/Source/path.cpp
+++ b/Source/path.cpp
@@ -11,108 +11,137 @@
 #include "objects.h"
 
 namespace devilution {
-
 namespace {
 
+constexpr size_t MaxPathNodes = 300;
+
+struct PathNode {
+	static constexpr uint16_t InvalidIndex = std::numeric_limits<uint16_t>::max();
+	static constexpr size_t MaxChildren = 8;
+
+	int16_t x = 0;
+	int16_t y = 0;
+	uint16_t parentIndex = InvalidIndex;
+	uint16_t childIndices[MaxChildren] = { InvalidIndex, InvalidIndex, InvalidIndex, InvalidIndex, InvalidIndex, InvalidIndex, InvalidIndex, InvalidIndex };
+	uint16_t nextNodeIndex = InvalidIndex;
+	uint8_t f = 0;
+	uint8_t h = 0;
+	uint8_t g = 0;
+
+	[[nodiscard]] Point position() const
+	{
+		return Point { x, y };
+	}
+
+	void addChild(uint16_t childIndex)
+	{
+		size_t index = 0;
+		for (; index < MaxChildren; ++index) {
+			if (childIndices[index] == InvalidIndex)
+				break;
+		}
+		assert(index < MaxChildren);
+		childIndices[index] = childIndex;
+	}
+};
+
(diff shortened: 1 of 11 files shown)


Source/path.cpp
```diff
@@ -11,116 +11,143 @@
 #include "objects.h"
 
 namespace devilution {
-
 namespace {
 
+constexpr size_t MaxPathNodes = 300;
+
+struct PathNode {
+	static constexpr uint16_t InvalidIndex = std::numeric_limits<uint16_t>::max();
```
ephphatha, line 19: Is it worth using `uint8_t` as an index type here? values currently will always be in the range 0-24

```suggestion
	using IndexType = uint_fast8_t
	static constexpr IndexType InvalidIndex = std::numeric_limits<IndexType>::max()
```

Source/path.cpp
```diff
@@ -11,116 +11,143 @@
...
+
+	void addChild(uint16_t childIndex)
+	{
+		size_t index = 0;
+		for (; index < 8; ++index) {
+			if (childIndices[index] == InvalidIndex)
+				break;
+		}
+		childIndices[index] = childIndex;
+	}
```
ephphatha, line 43: While it's impossible to hit in the current code as tiles only have 8 neighbours (so this is called a max of 8 times for a given node) this tries to write past the end of the array if all child nodes are full. Would it be worth restructuring the function to set and early return in the loop? e.g.
```suggestion
	void addChild(uint16_t childIndex)
	{
		for (auto &index : childIndices) {
			if (index == InvalidIndex) {
				index = childIndex;
				return;
			}
		}
	}
```

Source/path.cpp
glebm: `MaxPathNodes` is 300, are you saying we can reduce `MaxPathNodes` to 24?

Source/path.cpp
glebm: We can detect the writes past-the-end with ASAN whereas that version would just do nothing silently. Should I add an assert?

Source/path.cpp
ephphatha: My mistake, I missed the define and use of MaxPathNodes😂

Source/path.cpp
ephphatha: an assert sounds good

glebm: Done

ephphatha (approved):

Source/platform/locale.cpp
```diff
@@ -11,6 +11,9 @@
 #elif defined(__3DS__)
 #include "platform/ctr/locale.hpp"
 #elif defined(_WIN32)
+// Suppress definitions of `min` and `max` macros by <windows.h>:
+#define NOMINMAX 1
```
AJenbo, line 15: This appears to already be there

glebm: Rebase accident, thanks @ephphatha for fixing!

PR #4756, opened 2022-06-19
scrollrt: Optimize `dRendered` size (-10 KiB)

glebm:


AJenbo:
Coudn't we use std::bitset for this?

glebm:
Yes, we probably should! And `std::bitset` is also a good idea for this.

AJenbo (changes requested):
Waiting for std:bitset :)

glebm:
Done! Migrated a few arrays to an `std::bitset`-based implementation.

AJenbo (approved):

qndel:
Just 1 thing - if we got `set`, it would be logical to have `get` instead of test :D

glebm:
The API mimics std::bitset

PR #4764, opened 2022-06-20
Fix finding SimpleIni

glebm:
The version in Ubuntu 22.04 comes without `ConvertUTF.{h,c}`.

We actually do not need `ConvertUTF.{h,c}` and starting from v4.19 it is possible to tell SimpleIni to not include it by setting `SI_NO_CONVERSION`.

Sets `SI_NO_CONVERSION` and raises the minimum required SimpleIni version to 4.19.

/cc @tsunamistate @YuviTz1


AJenbo (approved):

tsunamistate:
@glebm thank you for the attempt, but CMake **still** doesn't find vcpkg version on Windows, and uses FetchContent version instead
https://ci.appveyor.com/project/AJenbo/devilutionx/builds/43919689/job/efdbw0ya1e7ji375#L1103

glebm:
That could be because of the `find_library` call.

As I understand, simpleini is header-only so that call and the corresponding REQUIRED_VAR should be removed

glebm:
Oh, you're also only checking package config include dirs. Try also checking `SIMPLEINI_INCLUDE_DIRS`.

glebm:
I'll send a PR

PR #4767, opened 2022-06-21
Remove duplicated anvil size/area calc in PlaceAnvil

ephphatha:


AJenbo:
> maybe make a class like miniset to wrap DUN files?

That would be really useful since we are going to be using it a lot once we move all the minisets that only search for floor areas.

AJenbo (approved):

PR #4768, opened 2022-06-21
CMake: gc-sections + no-rtti/exceptions options

glebm:
Turns out gc-sections helps significantly even with LTO (-84 KiB for the MinSizeRel rg99 build).

Also adds options for disabling RTTI and exceptions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,9 @@ option(DEVILUTIONX_RESAMPLER_SDL "Build with SDL resampler" ON)
 
 option(DEVILUTIONX_PALETTE_TRANSPARENCY_BLACK_16_LUT "Whether to use a lookup table for transparency blending with black. This improves performance of blending transparent black overlays, such as quest dialog background, at the cost of 128 KiB of RAM." ON)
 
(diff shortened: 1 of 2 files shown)


AJenbo (approved):
What part of NONET is it that enables no-rtti?

glebm:
Actually I'm not sure that it does, although we do have a couple of `dynamic_cast` calls in tests.

I've also realized we already have the bit of code that disables RTTI and EXCEPTIONS for NONET, going to look a bit more into this

glebm:
Ah, that bit of code was in a branch I've never submitted as a PR :sweat_smile:

glebm:
Updated, PTAL

AJenbo:
Would dynamic linking aiso help with that?

glebm:
Possibly, though our own network code uses exceptions.

glebm:
Clarifying "possibly":

1. Only if asio doesn't use RTTI and exceptions in its headers.
2. Only on platforms that have dynamic linking at all (e.g. not psp if I remember right)

AJenbo (approved):

PR #4770, opened 2022-06-21
Use the same function when checking and actually dropping an item

### LINKED ISSUE #4763: Ctrl+Left Click Drop from Inventory when Standing on Item
**Important information**
Windows 10 / 1.4.0 / Download Release

**Describe**
When standing on an item and using ctrl+left click to drop items from inventory, there are three tiles the game cannot drop the items onto (will give audio feedback there's no where to place items if all other spots are taken and item will remain in inventory). The impacted tiles appear to be (-1, -1), (0, -1), and (-1, 0) relative to character position.

**To Reproduce**
1. Have an inventory with at least 7 items.
2. Drop one of the items such that it is on the same tile as your character.
3. Ctrl + Left Click on the other items in your inventory until the issue happens.

**Expected behavior**
If there is an open tile next to your character, you should be able to ctrl + left click drop items from inventory.

**Screenshots**
The described tiles in screenshot are empty, but cannot ctrl + left click drop items anymore.
image

ephphatha:
fixes #4763

Instead of using a function that only checks 6 or 7 tiles, extract the logic that checks all 9 tiles around a starting position which is used when dropping items and use it consistently.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1831,13 +1831,9 @@ bool TryDropItem()
 
 	Point position = myPlayer.position.future;
 	Direction direction = myPlayer._pdir;
-	if (!CanPut(position, direction)) {
-		direction = Opposite(direction);
-		// if we can't drop in front of the player, can we drop it behind?
-		if (!CanPut(position, direction)) {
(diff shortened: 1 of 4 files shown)


AJenbo (approved):
Seams like a good idea 😅

PR #4771, opened 2022-06-21
CMake: Clean up option definitions

glebm:
1. Make the vcpkg hack minimal by only defining the options it needs before `include(Platforms)`.

   Every dependent option needed by vcpkg has to be recalculated after `include(Platforms)`. This brings the count of such options down to 1 (`PACKET_ENCRYPTION`).

2. `include(Platforms)` before the rest of the options.

3. Group related options together into sections and add comments.

4. Fix `--gc-sections` for C files.

Note: `PIE` can now be `ON` even if `BUILD_TESTING` is `OFF`.

```
$ du -b build-rg99/devilutionx
1357720	build-rg99/devilutionx
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,9 @@ endif()
 if(POLICY CMP0111)
   cmake_policy(SET CMP0111 NEW)
 endif()
+if(POLICY CMP0126)
+  cmake_policy(SET CMP0126 NEW)
+endif()
 
 # Projects added via `add_subdirectory` or `FetchContent` may have a lower
 # `cmake_minimum_required` than we set here. Set policies that we require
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -203,11 +212,14 @@ endif()
 if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
   # For some reason, adding to CMAKE_CXX_FLAGS results in a slightly smaller
   # binary than using `add_compile/link_options`
-  set(_extra_flags "-ffunction-sections -fdata-sections -Wl,--gc-sections")
+  set(_extra_flags "-ffunction-sections -fdata-sections -Wl,--gc-sections -Wl,--as-needed")
```
AJenbo, line 215: Shouldn't this also work?
```suggestion
  set(_extra_flags "-ffunction-sections -fdata-sections -Wl,--gc-sections,--as-needed")
```

glebm: Probably, I'm not near a computer to try but feel free to make this change if it works! 👍

glebm: Done

AJenbo (approved):
Looks good, I originally had some concerns but after a proper reading I see that they where already handled :+1:

PR #4775, opened 2022-06-22
Enforce override virtual methods

kalebealvs:
I know this is working, but it kinda bugged me seeing all those virtual methods not being checked by the compiler. :smile: 
I don't believe it's a style thing since other newer code is not written like that, ie IsGameHost in the same file.


AJenbo (approved):
Nice, a lot of the network code could do with an overhaul :)

PR #4776, opened 2022-06-22
Move level generation to subfolder

AJenbo:


ephphatha (dismissed):
makes sense, once the include order is fixed so clang-format is happy it's good to go :D

PR #4779, opened 2022-06-24
AiRanged refactors

k-bar:
Refactor AiRanged changes.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -2070,36 +2070,27 @@ void AiAvoidance(int i, bool special)
 	monster.CheckStandAnimationIsLoaded(md);
 }
 
- missile_id GetMissileType(_mai_id ai)
+missile_id GetMissileType(_mai_id ai)
 {
 	switch (ai) {
 	case AI_GOATMC:
 		return MIS_ARROW;
-		break;
 	case AI_SUCC:
 		return MIS_FLARE;
-		break;
 	case AI_ACIDUNIQ:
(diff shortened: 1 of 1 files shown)


qndel:
if you don't want these commits to get squashed, how about fixing that typo `ger` -> `get`? in commit description? :D

Source/monster.cpp
```diff
@@ -2070,7 +2070,42 @@ void AiAvoidance(int i, bool special)
 	monster.CheckStandAnimationIsLoaded(md);
 }
 
-void AiRanged(int i, missile_id missileType, bool special)
+ missile_id GetMissileType(_mai_id ai)
+{
+	switch (ai) {
+	case AI_GOATMC:
+		return MIS_ARROW;
+		break;
```
qndel, line 2078: you don't need breaks after return

k-bar: isn't that good practice or something ? :D

glebm: Not after a return

DakkJaniels:
Are these kind of changes being playtested before merging? Looks like this might be causing Laz's helpers to fire arrows instead of bloodstars because their AI is AI_LAZHELP, not AI_SUCC

k-bar:
> Are these kind of changes being playtested before merging? Looks like this might be causing Laz's helpers to fire arrows instead of bloodstars because their AI is AI_LAZHELP, not AI_SUCC

yes, but unfortunately not through all kind of monsters.

PR #4780, opened 2022-06-24
AiRangedAvoidance rafactor

k-bar:
Refactor AiRangedAvoidance changes.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -2569,11 +2569,6 @@ void FallenAi(int i)
 		SkeletonAi(i);
 }
 
-void MagmaAi(int i)
-{
-	AiRangedAvoidance(i);
-}
-
 void LeoricAi(int i)
 {
 	assert(i >= 0 && i < MAXMONSTERS);
@@ -2815,11 +2810,6 @@ void SneakAi(int i)
 	}
 }
 
-void StormAi(int i)
-{
-	AiRangedAvoidance(i);
-}
-
 void GharbadAi(int i)
 {
 	assert(i >= 0 && i < MAXMONSTERS);
@@ -2865,11 +2855,6 @@ void GharbadAi(int i)
 		AiAvoidance(i, true);
 
 	monster.CheckStandAnimationIsLoaded(md);
-}
-
-void AcidAvoidanceAi(int i)
-{
-	AiRangedAvoidance(i);
 }
 
 void SnotSpilAi(int i)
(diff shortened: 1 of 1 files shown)


Source/monster.cpp
```diff
@@ -2156,8 +2165,11 @@ void AiRangedAvoidance(int i, missile_id missileType, bool checkdoors, int dam,
 	int mx = monster.position.tile.x - fx;
 	int my = monster.position.tile.y - fy;
 	Direction md = GetDirection(monster.position.tile, monster.position.last);
-	if (checkdoors && monster._msquelch < UINT8_MAX)
+	if ((monster._mAi == AI_MAGMA || monster._mAi == AI_STORM || monster._mAi == AI_BONEDEMON) && monster._msquelch < UINT8_MAX)
```
AJenbo, line 2168:
```suggestion
	if (IsAnyOf(monster._mAi, AI_MAGMA, AI_STORM, AI_BONEDEMON) && monster._msquelch < UINT8_MAX)

```

Source/monster.cpp
```diff
@@ -2156,8 +2165,11 @@ void AiRangedAvoidance(int i, missile_id missileType, bool checkdoors, int dam,
 	int mx = monster.position.tile.x - fx;
 	int my = monster.position.tile.y - fy;
 	Direction md = GetDirection(monster.position.tile, monster.position.last);
-	if (checkdoors && monster._msquelch < UINT8_MAX)
+	if ((monster._mAi == AI_MAGMA || monster._mAi == AI_STORM || monster._mAi == AI_BONEDEMON) && monster._msquelch < UINT8_MAX)
 		MonstCheckDoors(monster);
+	int lessmissiles = (monster._mAi == AI_ACID) ? 1 : 0;
+	int dam = (monster._mAi == AI_DIABLO) ? 40 : 4;
+	missile_id missileType = GetMissileType(monster._mAi);
```
AJenbo, line 2172: I think it would be better if the missile type was part of the monsterData, what do you think?

k-bar: I was thinking the same when I saw this:
https://github.com/diasurgical/devilutionX/blob/99181fd70981c81d879682a7b39616515731c9d0/Source/monster.cpp#L3703-L3744

Those are MissileSpriteData though, but quite correlated imho.
Could be part of another PR maybe :)

AJenbo (approved):

PR #4783, opened 2022-06-24
Fallen warcry dungeon inbound fix for vanilla bug

k-bar:
a bug found in Devilution
https://github.com/diasurgical/devilution/pull/2262

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -2537,7 +2537,8 @@ void FallenAi(int i)
 			for (int x = -rad; x <= rad; x++) {
 				int xpos = monster.position.tile.x + x;
 				int ypos = monster.position.tile.y + y;
-				if (InDungeonBounds({ x, y })) {
+				// BUGFIX: incorrect check of offset against limits of the dungeon (fixed)
+				if (InDungeonBounds({ xpos, ypos })) {
 					int m = dMonster[xpos][ypos];
 					if (m <= 0)
 						continue;


AJenbo (approved):
:+1:

Source/monster.cpp
```diff
@@ -2537,7 +2537,8 @@ void FallenAi(int i)
 			for (int x = -rad; x <= rad; x++) {
 				int xpos = monster.position.tile.x + x;
 				int ypos = monster.position.tile.y + y;
-				if (InDungeonBounds({ x, y })) {
+				// BUGFIX: incorrect check of offset against limits of the dungeon (fixed)
```
julealgon, line 2540: I understand this has already been merged, but I think it is a bad idea to add these kinds of comments to the source code. If we add such comments for every bug fix (and/or feature) that is made, the code will be littered with noise.

Having an issue that documents the bug and then link from the issue to the PR to the code would be vastly preferred.

AJenbo: Since the upstream comment hasn't been merged it's good to have this here so we know it's been addressed already. Other comments are generally being removed as the issues are fixed.

k-bar: @julealgon 
I was inspired by scavengerAi comments :)
@AJenbo 
Is it good idea to remove it in next Fallen_AI refactor iteration?

AJenbo: No, it should be keept untill i merge upstream.

PR #4778, opened 2022-06-23
Simplify loops in WillThemeRoomFit, rename to match use

ephphatha:
My brain is mush after looking at this code. It looks like it was intended to look for the largest/most balanced rectangular region within a given search area, but due to a weird way of picking when to use a particular bound it ended up getting smaller areas than strictly necessary if there were choices where both dimensions are over the minimum bound.

There's probably a way to make this even more straightforward but I have given up thought for today 😂 

The following are a few search spaces I dumped out so I could see what the old logic was doing and confirm this rewrite worked the same.
```
INFO: DebugCmd: changelevel 10 Result: Welcome to level 10.
DEBUG: +++++#####
DEBUG: +++++#####
DEBUG: +++++##-##
DEBUG: +++++--###
DEBUG: +++++--###
DEBUG: #-------##
DEBUG: ##------##
DEBUG: ##------##
DEBUG: ##-----###
DEBUG: -------###
DEBUG: found 5, 5
INFO: DebugCmd: changelevel 12 Result: Welcome to level 12.
DEBUG: +++++#####
DEBUG: +++++###-#
DEBUG: +++++----#
DEBUG: +++++---##
DEBUG: +++++--###
DEBUG: --###--###
DEBUG: -####----#
DEBUG: -#####--##
DEBUG: #-####--##
DEBUG: #####---##
DEBUG: found 5, 5
DEBUG: ++++++--##
DEBUG: ++++++-###
DEBUG: ++++++-###
DEBUG: ++++++-###
DEBUG: ++++++####
DEBUG: ++++++####
DEBUG: ----######
DEBUG: -#########
DEBUG: -#########
DEBUG: ##########
DEBUG: found 6, 6
The thread 0x66b8 has exited with code 0 (0x0).
INFO: DebugCmd: changelevel 7 Result: Welcome to level 7.
DEBUG: ++++++####
DEBUG: ++++++####
DEBUG: ++++++####
DEBUG: ++++++####
DEBUG: ++++++####
DEBUG: ++++++####
DEBUG: ++++++####
DEBUG: ++++++-###
DEBUG: ++++++-###
DEBUG: ++++++-###
DEBUG: found 6, 10
DEBUG: ++++++---# -- this should've been either a 6x9 or 9x6 area
DEBUG: ++++++---# -- since both width and height of both choices
DEBUG: ++++++---# -- were over the minSize value the original logic
DEBUG: ++++++---# -- ends up picking the intersection of both regions
DEBUG: ++++++---#
[...]

PR #4784, opened 2022-06-24
Use the largest available area when placing theme rooms

ephphatha:
Splitting this into a separate PR from #4778 as this does change the boundaries of theme rooms. It could potentially put someone in a wall if they load a save made from an old version (though they can walk off it), and could lead to bodies/objects/items/monsters floating in the air/inside walls.

For example:
cat_7
cat_7_fixed
cat_8
cat_8_fixed

But since it's one of the last phases of dungeon generation, it's relatively minor :D. Also got rid of the buggy bounds check since this changes behaviour anyway.

### DIFF
--- a/Source/engine/size.hpp
+++ b/Source/engine/size.hpp
@@ -91,6 +91,11 @@ struct Size {
 		return a;
 	}
 
+	constexpr int area() const
+	{
+		return width * height;
+	}
+
 #ifdef BUILD_TESTING
 	/**
 	 * @brief Format sizes nicely in test failure messages
--- a/Source/levels/gendung.cpp
+++ b/Source/levels/gendung.cpp
@@ -104,35 +104,45 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
 	int maxWidth = std::min(DMAXX - origin.x, maxSize);
 	int maxHeight = std::min(DMAXY - origin.y, maxSize);
 
-	// see if we can find a region at least as large as the minSize on each dimension
-	for (int yOffset = 0; yOffset < maxHeight; yOffset++) {
+	// Start out looking for the widest area at least as tall as minSize
+	for (int yOffset = 0; yOffset < minSize; yOffset++) {
 		for (int xOffset = 0; xOffset < maxWidth; xOffset++) {
-			// Start out looking for the widest area at least as tall as minSize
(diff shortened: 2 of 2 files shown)


AJenbo:
Yikes, I'm not to happy with the walls being moved around :/

julealgon:
It's not exactly clear to me why this change is needed: at first glance, it seems like an arbitrary (and ultimately subjective) behavior change. Can someone elaborate? Are we considering vanilla behavior "incorrect" for some reason?

ephphatha:
I'll preface this by saying that I'm not 100% sure what the true intent of the original code actually was, it was either attempting to find the largest total area or possibly the squarest region that avoids corners? It ends up with some odd behaviour with certain layouts.

For a caves level (minSize is 5) the vanilla logic chooses the following areas:
```
++++++++-#
++++++++--
++++++++--
++++++++--
++++++++--
++++++++-# Same width as one of the first 5 rows, so doesn't trigger the early break/return
--------## This row does trigger the early return, and because
--------## the width is not smaller than minSize also defines the boundary
-------###
-------###
```
```
+++++++++-
+++++++++-
+++++++++-
+++++++++-
+++++++++-
---------# Now it's the first corner encountered and the row is still not less than minSize, so this defines the area
--------## End result is a wider/shorter area despite the only difference being no wall tiles in the first 5 rows.
--------##
-------###
-------###
```
```
++++++++++
++++++++++
++++++++++
++++++++++
++++++++++
----####-# This time the row is less than minSize, so we use the full width instead?
--------##
--------##
-------###
-------###
```
```
+++++++###
+++++++---
+++++++---
+++++++---
+++++++---
+++++++--#
+++++++-##
+++++++-##
+++++++### Not detected as a corner since it's the same width as one of the first
+++++++### 5 rows, so the region continues to the full height?
```

Is it intentional that case 2 gives a smaller total area than case 3? Is it more appropriate to use an 8x6 area for case 1 instead of a 7x10 area like what ends up getting chosen for case 4?

I don't wanna call this a fix because it's not clear what the actual "correct" behaviour is for this code. It's one possible interpretation, another could be to pick out corners better. This PR is more of an experiment following a comment in discord about #4778, that PR keeps the original logic and hopefully just makes the code easier to understand.

ephphatha:
Gonna close this one since it does change the placement of wall tiles, keeping the original behaviour is simpler anyway

PR #4785, opened 2022-06-24
M_FallenFear refactor

k-bar:
- makes function check fear affected area tiles only instead of every monster in `ActiveMonsters`
- run distance based on Intelligence Factor instead of fallen type
 to makes things equal to vanilla - adjusted for Carver with spear, which intelligenceFactor is higher than usual monster type streak (0,1,2,3) and is 2 instead of expected 1
 
 as bonus - fixed fallen retreat direction bug introduced in https://github.com/diasurgical/devilutionX/pull/2354/commits/4c578c537843c7c9adb011941e0262b3d19ece56
now they will run properly away from killed monster

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4515,40 +4515,24 @@ void SyncMonsterAnim(Monster &monster)
 
 void M_FallenFear(Point position)
 {
-	for (int i = 0; i < ActiveMonsterCount; i++) {
-		auto &monster = Monsters[ActiveMonsters[i]];
+	const Rectangle fearArea = Rectangle { position, 4 };
+	for (const Point tile : PointsInRectangleRange { fearArea }) {
+		if (!InDungeonBounds(tile))
+			continue;
+		int m = dMonster[tile.x][tile.y];
+		if (m == 0)
+			continue;
+		Monster &monster = Monsters[m - 1];
+		if (monster._mAi != AI_FALLEN || monster._mhitpoints >> 6 <= 0)
+			continue;
 
-		if (monster._mAi != AI_FALLEN)
-			continue;
-		if (position.WalkingDistance(monster.position.tile) >= 5)
-			continue;
-		if (monster._mhitpoints >> 6 <= 0)
-			continue;
-
-		int rundist;
-		switch (monster.MType->mtype) {
-		case MT_RFALLSP:
-		case MT_RFALLSD:
-			rundist = 7;
-			break;
-		case MT_DFALLSP:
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Nice work :)

Source/monster.cpp
```diff
@@ -4515,40 +4516,24 @@ void SyncMonsterAnim(Monster &monster)
...
-		case MT_BFALLSD:
-			rundist = 2;
-			break;
-		default:
-			continue;
-		}
+		int intelligenceFactor = monster._mint;
+		if (monster.MType->mtype == MT_DFALLSP)
+			intelligenceFactor--;
+		int runDistance = std::max((7 - 2 * intelligenceFactor), 2);
```
AJenbo, line 4533: On closer inspection this might cause a change in behavior since Pukerat the Unclean, Bongo, Gutshank the Quick, Shadowcrow and there minions have altered intelligence compared to the base.

k-bar:
Oh gee, I really liked this :) 
Btw I guess this is making them harder a bit ;)
I will try to fry some solution for this.
Btw it is only Pukerat the Unclean and Bongo.
Since Gutshank the Quick has AI_BAT, and Shadowcrow AI_SNEAK. (And Bladeskin the Slasher is ok with 0-0 and Snotspil 3-3)

Source/monster.cpp
k-bar: Oh gee, I really liked this :)
Btw I guess this is making them harder a bit ;)
I will try to fry some solution for this.
Btw it is only Pukerat the Unclean and Bongo.
Since Gutshank the Quick has AI_BAT, and Shadowcrow AI_SNEAK. (And Bladeskin the Slasher is ok with 0-0 and Snotspil 3-3)

k-bar: Pukerat the Unclean easy - no mobs

k-bar: lol, got better idea
Fallen level 1
Carver level 3
Devil Kin level 5
Dark One level 7
:)
how nice is that

k-bar: unless it have to be otherwise :)
Fallen level 1 -> 7
Carver level 3 -> 5
Devil Kin level 5 -> 3
Dark One level 7 -> 2

k-bar: `max (8 - base_monster_level), 2)`

AJenbo: That might work

PR #4789, opened 2022-06-25
DebugCmdLoadQuestMap: Don't load entry quest lvl if the hero is already on this level

obligaron:
When accessing a quest map with debug commands, the level where the quest can be entered is loaded first (so ReturnPosition is set).
But this is not needed if we are already on the dungeon level.


AJenbo (approved):
I plan to fix how level entrances are calculated so that this wont be needed :)

obligaron:
That would be great. 👍

PR #4790, opened 2022-06-25
FallenFear retreat distance based on level

k-bar:
FallenFear retreat distance based on level
Previously (#4785) based on IntelligenceFactor would be off for Pukerat the Unclean and Bongo. (Thanks @AJenbo for noticing).
And this one is cleaner too.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4528,11 +4528,7 @@ void M_FallenFear(Point position)
 		if (monster._mAi != AI_FALLEN || monster._mhitpoints >> 6 <= 0)
 			continue;
 
-		int intelligenceFactor = monster._mint;
-		if (monster.MType->mtype == MT_DFALLSP)
-			intelligenceFactor--;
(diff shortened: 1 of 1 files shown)


Source/monster.cpp
```diff
@@ -4528,11 +4528,7 @@ void M_FallenFear(Point position)
 		if (monster._mAi != AI_FALLEN || monster._mhitpoints >> 6 <= 0)
 			continue;
 
-		int intelligenceFactor = monster._mint;
-		if (monster.MType->mtype == MT_DFALLSP)
-			intelligenceFactor--;
-		int runDistance = std::max((7 - 2 * intelligenceFactor), 2);
-
+		int runDistance = std::max((8 - monster.MData->mLevel), 2);
```
k-bar, line 4531: I am calling this explanatory variable. I prefer this above comments.
(since mgoalvar1 says nothing).

AJenbo: Yeah we are slowly trying to rename things to have sensible names.

AJenbo (approved):

PR #4799, opened 2022-06-25
Use binaryformat for demofile

obligaron:
This makes the demo recording smaller and git doesn't recognize it as a text file 😉  (see [comment](https://github.com/diasurgical/devilutionX/pull/2490#discussion_r895181404))

Note:
- The format will change (see #2733).
- That's why I didn't try to make it endian safe for now.
- Perhaps we could later use a abstraction to write binary that is used in demo recording and loadsave. But one step at a time.

### DIFF
--- a/Source/engine/demomode.cpp
+++ b/Source/engine/demomode.cpp
@@ -58,7 +58,7 @@ void PumpDemoMessage(DemoMsgType demoMsgType, uint32_t message, int32_t wParam, 
 }
 
 template <class T>
-T ReadFromStream(std::ifstream& stream)
+T ReadFromStream(std::ifstream &stream)
 {
 	T value;
 	stream.read(reinterpret_cast<char *>(&value), sizeof(value));
@@ -296,8 +296,8 @@ void NotifyGameLoopEnd()
 	}
 
 	if (IsRunning()) {
-		float secounds = (SDL_GetTicks() - StartTime) / 1000.0f;
-		SDL_Log("%d frames, %.2f seconds: %.1f fps", LogicTick, secounds, LogicTick / secounds);
+		float seconds = (SDL_GetTicks() - StartTime) / 1000.0f;
+		SDL_Log("%d frames, %.2f seconds: %.1f fps", LogicTick, seconds, LogicTick / seconds);
 		gbRunGameResult = false;
 		gbRunGame = false;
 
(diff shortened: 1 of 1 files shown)


Source/engine/demomode.cpp
```diff
@@ -290,7 +296,7 @@ void NotifyGameLoopEnd()
 	}
 
 	if (IsRunning()) {
-		float secounds = (SDL_GetTicks() - StartTime) / 1000.0;
+		float secounds = (SDL_GetTicks() - StartTime) / 1000.0f;
```
qndel, line 299:
```suggestion
		float seconds = (SDL_GetTicks() - StartTime) / 1000.0f;
```
pls fix while we are at it :D

AJenbo (approved):
Nice!

Source/engine/demomode.cpp
```diff
@@ -57,53 +57,50 @@ void PumpDemoMessage(DemoMsgType demoMsgType, uint32_t message, int32_t wParam,
 	Demo_Message_Queue.push_back(msg);
 }
 
+template <class T>
+T ReadFromStream(std::ifstream& stream)
```
AJenbo, line 61:
```suggestion
T ReadFromStream(std::ifstream &stream)
```

Source/engine/demomode.cpp
```diff
@@ -290,7 +296,7 @@ void NotifyGameLoopEnd()
 	}
 
 	if (IsRunning()) {
-		float secounds = (SDL_GetTicks() - StartTime) / 1000.0;
+		float secounds = (SDL_GetTicks() - StartTime) / 1000.0f;
 		SDL_Log("%d frames, %.2f seconds: %.1f fps", LogicTick, secounds, LogicTick / secounds);
```
AJenbo, line 300:
```suggestion
		float seconds = (SDL_GetTicks() - StartTime) / 1000.0f;
		SDL_Log("%d frames, %.2f seconds: %.1f fps", LogicTick, seconds, LogicTick / seconds);
```

AJenbo (approved):

Source/engine/demomode.cpp
```diff
@@ -57,53 +57,50 @@ void PumpDemoMessage(DemoMsgType demoMsgType, uint32_t message, int32_t wParam,
...
+{
+	T value;
+	stream.read(reinterpret_cast<char *>(&value), sizeof(value));
+	return value;
+}
+
+template <class T>
+void WriteToDemo(T value)
+{
+	DemoRecording.write(reinterpret_cast<char *>(&value), sizeof(value));
```
glebm, line 71: This is an unaligned write and endian-dependent
Please do writes and reads the same way we do them for other things, e.g saves

PR #4801, opened 2022-06-26
Remove spurious "I can't cast that here"

vlolteanu:


AJenbo:
Isn't this going to prevent users from casting identify/repair/recharge when pointing to the desired item?

vlolteanu:
You're right.

Fixed.

AJenbo:
Could you explain what condition this now changes :sweat_smile:

vlolteanu:
If you've readied some spell that needs targeting, such as Fireball, and you're in the dungeon, right-clicking on anything that is not the playing field (character sheet, inventory etc) won't do anything. However, if you're in town, the character says "I can't cast that here".

PR #4803, opened 2022-06-26
Clean up Operate Door

AJenbo:
The pnum is known at the same time as sendflag, so we can simply use that early on in stead of passing the check on to the end function. This gives a nice reduction in parameters and comparisons.


glebm (approved):
Nice simplification!

AJenbo:
Credit goes to @ephphatha who suggested it during the clean up i was working on :) https://github.com/diasurgical/devilutionX/pull/4571#discussion_r872918314

PR #4808, opened 2022-06-27
Pass players by reference instead of an index

AJenbo:
Mostly trivial swaps of `Players` index to references where https://github.com/diasurgical/devilutionX/pull/4754 makes changes.


ephphatha (approved):
Reminds me that I need to find time to finish cleaning up cleanupitems 😓

PR #4809, opened 2022-06-27
Clean up HandleTurnUpperBit

AJenbo:


Source/multi.cpp
```diff
@@ -197,17 +197,15 @@ void MonsterSeeds()
 
 void HandleTurnUpperBit(int pnum)
 {
-	int i;
-
-	for (i = 0; i < MAX_PLRS; i++) {
-		if ((player_state[i] & PS_CONNECTED) != 0 && i != pnum)
+	for (int i = 0; i < MAX_PLRS; i++) {
```
StephenCWills, line 200:
```suggestion
	if (pnum == MyPlayerId) {
		gbDeltaSender = MAX_PLRS;
	}
	for (int i = 0; i < MAX_PLRS; i++) {
```

StephenCWills (changes requested):
There's an edge case where `pnum == MyPlayerId` and no other player has the `PS_CONNECTED` flag set. In this case, it would set `gbDeltaSender = MAX_PLRS`, which may be important for...

https://github.com/diasurgical/devilutionX/blob/e47599c45f89cb257a9dc0d9b29e27f0c11d11b1/Source/msg.cpp#L292-L297

It may also be important so that it doesn't stay at its default value of uninitialized, or even the value it had in the previous game.

Source/multi.cpp
AJenbo: Thanks, I specifically requested you for this PR cause I thought you might know if this was relevant :)

PR #4810, opened 2022-06-27
Pass along bools instead of player index

AJenbo:


Source/objects.cpp
```diff
@@ -5168,7 +5168,7 @@ void SyncOpObject(int pnum, int cmd, int i)
 	case OBJ_LEVER:
 	case OBJ_L5LEVER:
 	case OBJ_SWITCHSKL:
-		OperateLever(pnum, i);
+		OperateLever(i, pnum == MyPlayerId);
```
obligaron, line 5171: nit: Could we introduce a löocal `sendmsg` variable (like in `OperateObject`)?
Then we would have only one comparison with `pnum` in `SyncOpObject` and I would expect we could need this more often in the future. 😉 
Note: Didn't use approve, else the pr gets automerged and I'm unsure if you want to change this or not. 🙂

Source/sync.cpp
```diff
@@ -312,7 +312,7 @@ uint32_t OnSyncData(const TCmd *pCmd, int pnum)
 				continue;
 
 			if (GetLevelForMultiplayer(*MyPlayer) == level) {
-				SyncMonster(pnum, monsterSyncs[i]);
+				SyncMonster(pnum > MyPlayerId, monsterSyncs[i]);
```
obligaron, line 315: Not directly related to this cleanup-pr:
What is the background of this comparison? I don't know any other part of the code that does this.

glebm (approved):

Source/sync.cpp
AJenbo: It's planned as one of the next things for one I resume the object cleanups :)

https://github.com/diasurgical/devilutionX/pull/4571#discussion_r872933663

Source/objects.cpp
AJenbo: I was thinking the number of comparisons here was exactly low enough that I didn't want to add it yet, but if I find one more later then I will add it :)

PR #4812, opened 2022-06-27
Random clean ups

AJenbo:


Source/msg.cpp
```diff
@@ -2087,7 +2087,7 @@ DWORD OnNakrul(const TCmd *pCmd)
 		}
 		IsUberRoomOpened = true;
 		Quests[Q_NAKRUL]._qactive = QUEST_DONE;
-		monster_some_crypt();
+		WeakenNaKrul();
```
yuripourre, line 2090:
```suggestion
		AwakenNaKrul();
```

Source/monster.h
```diff
@@ -276,7 +276,7 @@ void PrepareUniqueMonst(Monster &monster, int uniqindex, int miniontype, int bos
 void InitLevelMonsters();
 void GetLevelMTypes();
 void InitMonsterGFX(int monst);
-void monster_some_crypt();
+void WeakenNaKrul();
```
yuripourre, line 279:
```suggestion
void AwakenNaKrul();
```

yuripourre: I never played Hellfire but I think this suppose to enable NaKrul instead of reducing his power, right?

AJenbo: No. He is locked in a room. If you just open the door he will be a full strength but if you trigger spell books out side his room first he will be weakened making the fight much easier.

yuripourre (dismissed):

PR #4815, opened 2022-06-28
:recycle: Fix inconsistent monster id variables

canassa:
Hello!

This is my first PR here so I decided to start with something simple 😄 

The monster id naming convention was inconsistent with several names being used: i, mid, mon, mi, monsterId

This change normalises all names to monsterId


Source/monster.cpp
```diff
@@ -4829,13 +4829,13 @@ void TalktoMonster(Monster &monster)
 	}
 }
 
-void SpawnGolem(int i, Point position, Missile &missile)
+void SpawnGolem(int monsterId, Point position, Missile &missile)
```
AJenbo, line 4832: Should be maybe call this `id` since it's ambiguous if it's the player or monster id?

Source/monster.cpp
```diff
@@ -4305,13 +4305,13 @@ void ProcessMonsters()
 				raflag = MonsterFadeout(monster);
 				break;
 			case MonsterMode::RangedAttack:
-				raflag = MonaterRangedAttack(mi);
```
AJenbo, line 4308: Good catch

AJenbo (commented):
Nice work, maybe as a follow up you could start swapping `int monsterId` for `Monster &monster` where possible :)

Source/monster.cpp
```diff
@@ -1828,40 +1828,40 @@ void GroupUnity(Monster &monster)
...
 	return ok;
 }
 
-bool RandomWalk2(int i, Direction md)
+bool RandomWalk2(int monsterId, Direction md)
 {
 	Direction mdtemp = md;
-	bool ok = DirOK(i, md);    // Can we continue in the same direction
+	bool ok = DirOK(monsterId, md);    // Can we continue in the same direction
 	if (GenerateRnd(2) != 0) { // Randomly go left or right
```
AJenbo, line 1857:
```suggestion
	bool ok = DirOK(monsterId, md); // Can we continue in the same direction
	if (GenerateRnd(2) != 0) {      // Randomly go left or right
```

Source/monster.cpp
canassa: Fixed

Source/monster.cpp
canassa: Thanks!

canassa:
> 


> Nice work, maybe as a follow up you could start swapping int monsterId for Monster &monster where possible :)

@AJembo I can do that in a follow-up PR, but I am not sure how to proceed when the id is used for other things. Take `M_ClearSquares` for example:

```c
void M_ClearSquares(int monsterId)
{

	int mx = monster.position.old.x;
	int my = monster.position.old.y;
	int m1 = -(monsterId + 1);
	int m2 = monsterId + 1;

	for (int y = my - 1; y <= my + 1; y++) {
		for (int x = mx - 1; x <= mx + 1; x++) {
			if (InDungeonBounds({ x, y }) && (dMonster[x][y] == m1 || dMonster[x][y] == m2))
				dMonster[x][y] = 0;
		}
	}
}
```

How can I reference the `dMonster` without passing the id? Does the `&monster` contains a reference to its own id?

ephphatha:
> How can I reference the `dMonster` without passing the id? Does the `&monster` contains a reference to its own id?

I ended up refactoring this one as an example before realising you were already discussing it :D. That function can use pointer comparison since it's looking for tiles referencing the same monster instance. See #4816 .

Other functions are harder to refactor as they involve network messages, IDs are used to deal with desync issues, but ideally position would be the absolute identifier if we could guarantee monster positions stay in sync.

Source/monster.cpp
```diff
@@ -662,15 +662,15 @@ void LoadDiabMonsts()
 	}
 }
 
-void DeleteMonster(int i)
+void DeleteMonster(int monsterId)
```
ephphatha, line 665: This function doesn't take a monster ID, it's an index into the list of active monster IDs stored in ActiveMonsters. `i` is perfectly fine here, or `size_t activeIndex` if you want to be verbose.

canassa: Thanks! I went with `activeIndex` as I like verbose 😛

AJenbo (approved):

AJenbo:
The easiest is to just look for functions that do not use the int and solve them first :) The same change are needed for `Item`, `Player` and `Object` related function (I'm some what looking in to the player side of things atm).

PR #4817, opened 2022-06-28
Add getLeader helper for monsters in a pack

ephphatha:
Precursor to updating DirOk to take a reference

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -366,9 +366,7 @@ void PlaceGroup(int mtype, unsigned num, UniqueMonsterPack uniqueMonsterPack, Mo
 				minion.intelligence = leader->intelligence;
 
 				if (uniqueMonsterPack == UniqueMonsterPack::Leashed) {
-					minion.setLeader(*leader);
-					minion.leaderRelation = LeaderRelation::Leashed;
-					minion.ai = leader->ai;
+					minion.setLeader(leader);
 				}
 
 				if (minion.ai != AI_GARG) {
@@ -1646,11 +1644,25 @@ bool MonsterGotHit(Monster &monster)
 	return false;
 }
 
-void MonsterDeath(int monsterId)
+void ReleaseMinions(const Monster &leader)
 {
-	assert(static_cast<size_t>(monsterId) < MaxMonsters);
-	auto &monster = Monsters[monsterId];
+	for (int j = 0; j < ActiveMonsterCount; j++) {
+		auto &minion = Monsters[ActiveMonsters[j]];
+		if (minion.leaderRelation == LeaderRelation::Leashed && minion.getLeader() == &leader) {
+			minion.setLeader(nullptr);
+		}
+	}
+}
 
+void ShrinkLeaderPacksize(const Monster &monster)
+{
+	if (monster.leaderRelation == LeaderRelation::Leashed) {
+		monster.getLeader()->packSize--;
+	}
+}
+
+void MonsterDeath(Monster &monster)
+{
 	monster.var1++;
 	if (monster.type().type == MT_DIABLO) {
 		if (monster.position.tile.x < ViewPosition.x) {
@@ -1676,7 +1688,7 @@ void MonsterDeath(int monsterId)
 		dMonster[monster.position.tile.x][monster.position.tile.y] = 0;
 		monster.isInvalid = true;
 
-		M_UpdateLeader(monsterId);
+		M_UpdateRelations(monster);
 	}
 }
 
@@ -2344,8 +2356,7 @@ void ScavengerAi(int monsterId)
 		return;
 	if (monster.hitPoints < (monster.maxHitPoints / 2) && monster.goal != MonsterGoal::Healing) {
 		if (monster.leaderRelation != LeaderRelation::None) {
-			if (monster.leaderRelation == LeaderRelation::Leashed)
-				monster.getLeader()->packSize--;
+			ShrinkLeaderPacksize(monster);
 			monster.leaderRelation = LeaderRelation::None;
 		}
 		monster.goal = MonsterGoal::Healing;
@@ -3384,7 +3395,7 @@ bool UpdateModeStance(int monsterId)
 	case MonsterMode::HitRecovery:
 		return MonsterGotHit(monster);
 	case MonsterMode::Death:
-		MonsterDeath(monsterId);
+		MonsterDeath(monster);
 		return false;
 	case MonsterMode::SpecialMeleeAttack:
 		return MonsterSpecialAttack(monsterId);
@@ -3998,20 +4009,10 @@ void M_SyncStartKill(int monsterId, Point position, int pnum)
 	StartMonsterDeath(monster, pnum, false);
 }
 
-void M_UpdateLeader(int monsterId)
+void M_UpdateRelations(const Monster &monster)
 {
-	assert(static_cast<size_t>(monsterId) < MaxMonsters);
-	auto &monster = Monsters[monsterId];
-
-	for (size_t j = 0; j < ActiveMonsterCount; j++) {
-		auto &minion = Monsters[ActiveMonsters[j]];
-		if (minion.leaderRelation == LeaderRelation::Leashed && minion.getLeader() == &monster)
-			minion.leaderRelation = LeaderRelation::None;
-	}
-
-	if (monster.leaderRelation == LeaderRelation::Leashed) {
-		monster.getLeader()->packSize--;
-	}
+	ReleaseMinions(monster);
+	ShrinkLeaderPacksize(monster);
 }
 
 void DoEnding()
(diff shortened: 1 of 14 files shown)


k-bar:
I think this in `PlaceGroup` should be taken into consideration too:
https://github.com/diasurgical/devilutionX/blob/d1c489c09e47b9a0573739ef1aa940019855cb51/Source/monster.cpp#L367

k-bar:
Actually I think I was wrong. Because it is OK without any additional consideration ;)

k-bar:
on the other hand PlaceGroup is also called like that:
`PlaceGroup(mtype, na, UniqueMonsterPack::None, 0);`
with leaderID 0 (golem).

ephphatha:
`UniqueMonsterPack::None` means that `leaderId` and `leader` in `PlaceGroup()` are never used. Ideally the leader reference should not be made at all unless actually placing a group with a leader, but that's not the focus of this PR.

Source/qol/monhealthbar.cpp
```diff
@@ -135,7 +135,7 @@ void DrawMonsterHealthBar(const Surface &out)
 	DrawString(out, monster.mName, { position + Displacement { -1, 1 }, { width, height } }, style | UiFlags::ColorBlack);
 	if (monster._uniqtype != 0)
 		style |= UiFlags::ColorWhitegold;
-	else if (monster.leader != 0)
+	else if (monster.leader != Monster::NoLeader)
 		style |= UiFlags::ColorBlue;
 	else
 		style |= UiFlags::ColorWhite;
```
ephphatha, line 141: Previously monsters which were part of a group would keep the blue text on the health bar when their leader died. With the last commit they fall back to white. I didn't think this was worth trying to preserve, I'd rather not keep a reference to a dead monster around purely for an optional QoL feature.

AJenbo: They would still have increased stats, so would be best if there was still a way to indicate this.

ephphatha: Can add a flag to the Monster struct that gets set if they were spawned as part of a group, or maybe monhealthbar should be comparing stats to what is expected for that type of monster?

ephphatha: Was having a think about this and comparing stats *should* work for monhealthbar, but I need to read through the code to make sure every group does in fact have different int compared to when they're naturally spawned (there's a couple of unique bosses with 0 and normal spawns have 0, but they might not overlap). I'll revise this PR to leave the dangling leader reference with a comment about the use when I get time this evening.

Source/monster.cpp
```diff
@@ -479,6 +475,8 @@ void PlaceUniqueMonst(int uniqindex, int miniontype, int bosspacksize)
 		UberDiabloMonsterIndex = ActiveMonsterCount;
 	}
 	PlaceMonster(ActiveMonsterCount, uniqtype, position);
+	ActiveMonsterCount++;
```
AJenbo, line 478: Was this some sort of bug?

ephphatha: PrepareUniqueMonster was incrementing ActiveMonsterCount in response to the allocation in this function. This meant that the use in debug.cpp needed to compensate for it when spawning uniques as part of the debug command. PrepareUniqueMonster shouldn't modify ActiveMonsterCount at all so I moved the increment here.

Source/monster.cpp
```diff
@@ -4845,6 +4845,29 @@ void decode_enemy(Monster &monster, int enemyId)
...
+{
+	if (leader == nullptr) {
+		// really we should update this->leader to NoLeader to avoid leaving a dangling reference to a dead monster
+		// when passed nullptr, however Diablo 1 doesn't reuse monster instances. There's no chance this'll end up
+		// pointing to a live monster other than the original leader.
+		leaderRelation = LeaderRelation::None;
+		return;
+	}
+
+	this->leader = static_cast<uint8_t>(std::distance<const Monster *>(&Monsters[0], leader));
```
AJenbo, line 4866:
```suggestion
	this->leader = leader->getId();
```

Source/monster.cpp
```diff
@@ -4000,20 +4010,10 @@ void M_SyncStartKill(int monsterId, Point position, int pnum)
...
-		auto &minion = Monsters[ActiveMonsters[j]];
-		if (minion.leaderRelation == LeaderRelation::Leashed && minion.leader == monsterId)
-			minion.leaderRelation = LeaderRelation::None;
-	}
-
-	if (monster.leaderRelation == LeaderRelation::Leashed) {
-		Monsters[monster.leader].packSize--;
-	}
+	ReleaseMinions(monster);
+	ShrinkLeaderPacksize(monster);
```
AJenbo, line 4016: Looks like it only makes sens to run one or the other since a leader can't also be a minion.

```suggestion
	if (monster.leaderRelation == LeaderRelation::Leashed)
		Monsters[monster.leader].packSize--;
	else
		ReleaseMinions(monster);
```

AJenbo:
Alright finally had time to finish this review :)

Source/monster.cpp
```diff
@@ -4845,6 +4845,29 @@ void decode_enemy(Monster &monster, int enemyId)
...
+
+	return &Monsters[leader];
+}
+
+void Monster::setLeader(const Monster *leader)
+{
+	if (leader == nullptr) {
+		// really we should update this->leader to NoLeader to avoid leaving a dangling reference to a dead monster
+		// when passed nullptr, however Diablo 1 doesn't reuse monster instances. There's no chance this'll end up
+		// pointing to a live monster other than the original leader.
```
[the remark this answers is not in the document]
ephphatha: Thankfully those get reinitialised, the spawned monsters have no leader. I was thinking more along the lines of diablo 2 shaman which revive existing monsters.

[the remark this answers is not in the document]
ephphatha: ahh but those monsters could occupy the position of a group leader, which is the situation I was originally worried about. I'll update this comment.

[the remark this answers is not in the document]
AJenbo: Hellfire strikes back

Source/monster.cpp
ephphatha: While in practice there's no current situation where a unique monster has both a leader and minions there's no real reason why that couldn't happen. I've updated the PR to include a guard for the ReleaseMinions call as that's a relatively expensive function but I'd rather keep them as separate conditions.

ephphatha:
In order to properly detect if a monster is potentially the leader of other minions it required pulling in the logic that checks if a monster is unique and that the unique monster data says they spawn with leashed minions. Since that's intertwined with the previously unnammed enum holding unique monster types I made that a named and scoped enum, and updated all uses of it. This has blown out the line count significantly, going from ~50 to the current 386 changes -_-

Source/monster.h
```diff
@@ -48,20 +49,21 @@ enum monster_flag : uint16_t {
...
+	Lazarus,
+	RedVex,
+	BlackJade,
+	Lachdan,
+	WarlordOfBlood,
+	Butcher,
+	HorkDemon,
+	Defiler,
+	NaKrul,
+	None = static_cast<uint8_t>(-1),
```
AJenbo, line 66: can just change the type to `int8_t`, it's not like there are more then 128 uniques :)

Source/monster.cpp
```diff
@@ -4411,13 +4413,11 @@ void SyncMonsterAnim(Monster &monster)
...
-		monster.name = pgettext("monster", UniqueMonstersData[monster.uniqType - 1].mName).data();
-	else
-		monster.name = pgettext("monster", monster.data().mName).data();
-
-	if (monster.uniqType != 0)
+	if (monster.isUnique()) {
+		monster.name = pgettext("monster", UniqueMonstersData[static_cast<size_t>(monster.uniqueType)].mName).data();
 		InitTRNForUniqueMonster(monster);
+	} else
+		monster.name = pgettext("monster", monster.data().mName).data();
```
AJenbo, line 4420: Not a fan of mixed braces
```suggestion
	} else {
		monster.name = pgettext("monster", monster.data().mName).data();
	}
```

AJenbo:
> This has blown out the line count significantly, going from ~50 to the current 386 changes -_-

I noticed :D

Source/monster.cpp
AJenbo: The game distinctly has a rule that no unique can also be a minion
```cpp
	// No unique monster would be a minion of someone else!
	assert(!monster.isUnique());
```

Source/monster.cpp
```diff
@@ -625,24 +624,24 @@ void PlaceQuestMonsters()
 			UberDiabloMonsterIndex = -1;
 			size_t i1;
 			for (i1 = 0; i1 < LevelMonsterTypeCount; i1++) {
-				if (LevelMonsterTypes[i1].type == UniqueMonstersData[UMT_NAKRUL].mtype)
+				if (LevelMonsterTypes[i1].type == UniqueMonstersData[static_cast<size_t>(UniqueMonsterType::NaKrul)].mtype)
```
AJenbo, line 627: With this cast happening 17 times, maybe a helper is in order?

Source/monster.cpp
ephphatha: Only as a debug assert though? I was hoping to keep this separate as it doesn't save much to make them exclusive conditions. Fewer changes required for a mod which implements Red Vex and Black Jade as leashed minions of Lazarus as well.

Could still do an else on hasLeashedMinions, prefer not inlining the function entirely though

AJenbo (approved):

AJenbo:
`/root/repo/Source/monster.cpp:515:1: error: the value of 'devilution::UniqueMonstersData' is not usable in a constant expression`

Source/monster.cpp
AJenbo: nit

ephphatha:
> `/root/repo/Source/monster.cpp:515:1: error: the value of 'devilution::UniqueMonstersData' is not usable in a constant expression`

GCC 6.5 strikes again? Made those inline instead.

AJenbo:
I think also 3DS was mad about it and it uses GCC 12

PR #4818, opened 2022-06-29
Add helpers to anim info

k-bar:
- Added helpers to animationinfo (improves readability of it's usage in the code)
  added `triggerFrame`
- clean all clang-tidy remarks in animationinfo in the process (access/ case style / casts)

review remarks:
[animationinfo.h](https://github.com/diasurgical/devilutionX/blob/bf9c09cbd5198ee8d0cc44354d3bdb1c71a4bf12/Source/engine/animationinfo.h) <- here are vast majority of changes, but mainly access/sort/case style + added helpers
the rest of the code is derivates of these changes.

other change in the rest of the code is support for triggerFrame (load/save/monster/player).

### DIFF
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -12,92 +12,92 @@
 
 namespace devilution {
 
-int AnimationInfo::GetFrameToUseForRendering() const
+int AnimationInfo::getFrameToUseForRendering() const
 {
 	// Normal logic is used,
 	// - if no frame-skipping is required and so we have exactly one Animationframe per game tick
 	// or
 	// - if we load from a savegame where the new variables are not stored (we don't want to break savegame compatiblity because of smoother rendering of one animation)
-	if (RelevantFramesForDistributing <= 0)
-		return std::max(0, CurrentFrame);
+	if (relevantFramesForDistributing_ <= 0)
+		return std::max(0, currentFrame_);
 
-	if (CurrentFrame >= RelevantFramesForDistributing)
-		return CurrentFrame;
+	if (currentFrame_ >= relevantFramesForDistributing_)
+		return currentFrame_;
 
-	float ticksSinceSequenceStarted = TicksSinceSequenceStarted;
-	if (TicksSinceSequenceStarted < 0) {
+	float ticksSinceSequenceStarted = ticksSinceSequenceStarted_;
+	if (ticksSinceSequenceStarted_ < 0) {
 		ticksSinceSequenceStarted = 0.0F;
-		Log("GetFrameToUseForRendering: Invalid TicksSinceSequenceStarted {}", TicksSinceSequenceStarted);
+		Log("getFrameToUseForRendering: Invalid ticksSinceSequenceStarted_ {}", ticksSinceSequenceStarted_);
 	}
 
 	// we don't use the processed game ticks alone but also the fraction of the next game tick (if a rendering happens between game ticks). This helps to smooth the animations.
-	float totalTicksForCurrentAnimationSequence = GetProgressToNextGameTick() + ticksSinceSequenceStarted;
+	float totalTicksForCurrentAnimationSequence = getProgressToNextGameTick() + ticksSinceSequenceStarted;
 
-	int absoluteAnimationFrame = static_cast<int>(totalTicksForCurrentAnimationSequence * TickModifier);
-	if (SkippedFramesFromPreviousAnimation > 0) {
+	int absoluteAnimationFrame = static_cast<int>(totalTicksForCurrentAnimationSequence * tickModifier_);
+	if (skippedFramesFromPreviousAnimation_ > 0) {
 		// absoluteAnimationFrames contains also the Frames from the previous Animation, so if we want to get the current Frame we have to remove them
-		absoluteAnimationFrame -= SkippedFramesFromPreviousAnimation;
+		absoluteAnimationFrame -= skippedFramesFromPreviousAnimation_;
 		if (absoluteAnimationFrame < 0) {
 			// We still display the remains of the previous Animation
-			absoluteAnimationFrame = NumberOfFrames + absoluteAnimationFrame;
+			absoluteAnimationFrame = numberOfFrames_ + absoluteAnimationFrame;
 		}
-	} else if (absoluteAnimationFrame >= RelevantFramesForDistributing) {
+	} else if (absoluteAnimationFrame >= relevantFramesForDistributing_) {
 		// this can happen if we are at the last frame and the next game tick is due (gfProgressToNextGameTick >= 1.0f)
-		if (absoluteAnimationFrame >= (RelevantFramesForDistributing + 1)) {
+		if (absoluteAnimationFrame >= (relevantFramesForDistributing_ + 1)) {
 			// we should never have +2 frames even if next game tick is due
-			Log("GetFrameToUseForRendering: Calculated an invalid Animation Frame (Calculated {} MaxFrame {})", absoluteAnimationFrame, RelevantFramesForDistributing);
+			Log("getFrameToUseForRendering: Calculated an invalid Animation Frame (Calculated {} MaxFrame {})", absoluteAnimationFrame, relevantFramesForDistributing_);
 		}
-		return RelevantFramesForDistributing - 1;
+		return relevantFramesForDistributing_ - 1;
 	}
 	if (absoluteAnimationFrame < 0) {
-		Log("GetFrameToUseForRendering: Calculated an invalid Animation Frame (Calculated {})", absoluteAnimationFrame);
+		Log("getFrameToUseForRendering: Calculated an invalid Animation Frame (Calculated {})", absoluteAnimationFrame);
 		return 0;
 	}
 	return absoluteAnimationFrame;
 }
 
-float AnimationInfo::GetAnimationProgress() const
+float AnimationInfo::getAnimationProgress() const
 {
-	float ticksSinceSequenceStarted = TicksSinceSequenceStarted;
-	float tickModifier = TickModifier;
+	float ticksSinceSequenceStarted = ticksSinceSequenceStarted_;
+	float tickModifier = tickModifier_;
 
-	if (RelevantFramesForDistributing <= 0) {
-		// This logic is used if animation distribution is not active (see GetFrameToUseForRendering).
+	if (relevantFramesForDistributing_ <= 0) {
+		// This logic is used if animation distribution is not active (see getFrameToUseForRendering).
 		// In this case the variables calculated with animation distribution are not initialized and we have to calculate them on the fly with the given information.
-		ticksSinceSequenceStarted = static_cast<float>((CurrentFrame * TicksPerFrame) + TickCounterOfCurrentFrame);
-		tickModifier = 1.0F / static_cast<float>(TicksPerFrame);
+		ticksSinceSequenceStarted = static_cast<float>((currentFrame_ * ticksPerFrame_) + tickCounterOfCurrentFrame_);
+		tickModifier = 1.0F / static_cast<float>(ticksPerFrame_);
 	}
 
-	float totalTicksForCurrentAnimationSequence = GetProgressToNextGameTick() + ticksSinceSequenceStarted;
+	float totalTicksForCurrentAnimationSequence = getProgressToNextGameTick() + ticksSinceSequenceStarted;
 	float progressInAnimationFrames = totalTicksForCurrentAnimationSequence * tickModifier;
-	float animationFraction = progressInAnimationFrames / static_cast<float>(NumberOfFrames);
+	float animationFraction = progressInAnimationFrames / static_cast<float>(numberOfFrames_);
 	return animationFraction;
 }
 
-void AnimationInfo::SetNewAnimation(std::optional<CelSprite> celSprite, int numberOfFrames, int ticksPerFrame, AnimationDistributionFlags flags /*= AnimationDistributionFlags::None*/, int numSkippedFrames /*= 0*/, int distributeFramesBeforeFrame /*= 0*/, float previewShownGameTickFragments /*= 0.F*/)
+void AnimationInfo::setNewAnimation(std::optional<CelSprite> celSprite, int numberOfFrames, int ticksPerFrame, AnimationDistributionFlags flags /*= AnimationDistributionFlags::None*/, int numSkippedFrames /*= 0*/, int distributeFramesBeforeFrame /*= 0*/, float previewShownGameTickFragments /*= 0.F*/)
 {
-	if ((flags & AnimationDistributionFlags::RepeatedAction) == AnimationDistributionFlags::RepeatedAction && distributeFramesBeforeFrame != 0 && NumberOfFrames == numberOfFrames && CurrentFrame + 1 >= distributeFramesBeforeFrame && CurrentFrame != NumberOfFrames - 1) {
+	if ((flags & AnimationDistributionFlags::RepeatedAction) == AnimationDistributionFlags::RepeatedAction && distributeFramesBeforeFrame != 0 && numberOfFrames_ == numberOfFrames && currentFrame_ + 1 >= distributeFramesBeforeFrame && currentFrame_ != numberOfFrames_ - 1) {
 		// We showed the same Animation (for example a melee attack) before but truncated the Animation.
 		// So now we should add them back to the new Animation. This increases the speed of the current Animation but the game logic/ticks isn't affected.
-		SkippedFramesFromPreviousAnimation = NumberOfFrames - CurrentFrame - 1;
+		skippedFramesFromPreviousAnimation_ = numberOfFrames_ - currentFrame_ - 1;
 	} else {
-		SkippedFramesFromPreviousAnimation = 0;
+		skippedFramesFromPreviousAnimation_ = 0;
 	}
 
 	if (ticksPerFrame <= 0) {
-		Log("SetNewAnimation: Invalid ticksPerFrame {}", ticksPerFrame);
+		Log("setNewAnimation: Invalid ticksPerFrame {}", ticksPerFrame);
 		ticksPerFrame = 1;
 	}
 
-	this->celSprite = celSprite;
-	NumberOfFrames = numberOfFrames;
-	CurrentFrame = numSkippedFrames;
-	TickCounterOfCurrentFrame = 0;
-	TicksPerFrame = ticksPerFrame;
-	TicksSinceSequenceStarted = 0.F;
-	RelevantFramesForDistributing = 0;
-	TickModifier = 0.0F;
-	IsPetrified = false;
+	celSprite_ = celSprite;
+	numberOfFrames_ = numberOfFrames;
+	currentFrame_ = numSkippedFrames;
+	tickCounterOfCurrentFrame_ = 0;
+	ticksPerFrame_ = ticksPerFrame;
+	ticksSinceSequenceStarted_ = 0.F;
+	relevantFramesForDistributing_ = 0;
+	tickModifier_ = 0.0F;
+	isPetrified_ = false;
 
 	if (numSkippedFrames != 0 || flags != AnimationDistributionFlags::None) {
 		// Animation Frames that will be adjusted for the skipped Frames/game ticks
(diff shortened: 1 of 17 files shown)


ephphatha (dismissed):
Even though it's marked as a draft I think this makes sense, certainly makes it nicer to read

Source/monster.cpp
```diff
@@ -1221,8 +1221,8 @@ bool MonsterWalk(int monsterId, MonsterMode variant)
 	assert(monster.MType != nullptr);
 
 	// Check if we reached new tile
-	const bool isAnimationEnd = monster.AnimInfo.CurrentFrame == monster.AnimInfo.NumberOfFrames - 1;
-	if (isAnimationEnd) {
+	bool isWalkAnimationEnd = monster.AnimInfo.isLastFrame();
```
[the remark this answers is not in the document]
k-bar: ye, my mistake. I ommited this const without purpose. :)

Source/engine/animationinfo.cpp
```diff
@@ -98,6 +98,8 @@ void AnimationInfo::SetNewAnimation(std::optional<CelSprite> celSprite, int numb
 	RelevantFramesForDistributing = 0;
 	TickModifier = 0.0F;
 	IsPetrified = false;
+	triggerFrame = std::max(distributeFramesBeforeFrame - 1, 0);
```
k-bar, line 101: I am not happy about how distributeFramesBeforeFrame is named now with its new usage and meaning added.
Shall I rename it to something else?

k-bar: OK I screw this up. I need another argument pass here - because monsters actually do not load its anim the same like players... (considering trigger).

k-bar: Or I will just update trigger member value...

Source/engine/animationinfo.h
```diff
@@ -62,6 +62,18 @@ class AnimationInfo {
 	 */
 	bool IsPetrified;
 
+	uint8_t triggerFrame;
+
+	uint8_t lastFrame;
+
+	bool isLastFrame() const;
```
ephphatha, line 69: An example of what I meant in the other comment, you can declare and define functions in the header itself. Using constexpr is mainly a helper for optimisation passes (adds restrictions to the function such that it can't branch and a few other things):
```suggestion
	constexpr bool isLastFrame() const
	{
		return (CurrentFrame == lastFrame);
	}
```

but it's not particularly well supported in old GCC versions, so sometimes you have to drop that keyword and just use:
```suggestion
	bool isLastFrame() const
	{
		return (CurrentFrame == lastFrame);
	}
```

You might also see the keyword `inline` used in other headers, this is already implicitly applied to class member functions defined inside the class definition so doesn't need to be explicitly written here.

Source/engine/animationinfo.cpp
```diff
@@ -216,4 +218,24 @@ float AnimationInfo::GetProgressToNextGameTick() const
...
+}
+
+bool AnimationInfo::isTriggerFrame() const
+{
+	return (CurrentFrame == triggerFrame);
+}
+
+bool AnimationInfo::isAfterTriggerFrame() const
+{
+	return (CurrentFrame >= triggerFrame + 1);
```
ephphatha, line 238:
```suggestion
	return (CurrentFrame > triggerFrame);
```

Source/engine/animationinfo.h
k-bar: Thank you for short explanation. I will make them defined in header then :)

Source/engine/animationinfo.h
```diff
@@ -96,16 +129,15 @@ class AnimationInfo {
...
 	 * @brief Process the Animation for a game tick (for example advances the frame)
 	 * @param reverseAnimation Play the animation backwards (for example is used for "unseen" monster fading)
-	 * @param dontProgressAnimation Increase TickCounterOfCurrentFrame but don't change CurrentFrame
+	 * @param dontProgressAnimation Increase TickCounterOfCurrentFrame but don't change currentFrame
 	 */
 	void ProcessAnimation(bool reverseAnimation = false, bool dontProgressAnimation = false);
 
 private:
 	/**
-	 * @brief returns the progress as a fraction (0.0f to 1.0f) in time to the next game tick or 0.0f if the animation is frozen
```
k-bar, line 105: this is just moved below member variable

Source/loadsave.cpp
```diff
@@ -468,6 +468,11 @@ void LoadPlayer(LoadHelper &file, Player &player)
 	player._pBFrames = file.NextLE<int32_t>();
 	file.Skip(4); // skip _pBWidth
 
+	if (player._pmode == PM_SPELL)
+		player.AnimInfo.triggerFrame = player._pSFNum - 1;
+	if (player._pmode == PM_ATTACK || player._pmode == PM_RATTACK)
+		player.AnimInfo.triggerFrame = player._pAFNum - 1;
+
```
k-bar, line 475: I was wondering where to place it.
Alternative would be some init player or similiar.

Source/monster.cpp
```diff
@@ -371,7 +371,7 @@ void PlaceGroup(int mtype, int num, UniqueMonsterPack uniqueMonsterPack, int lea
 
 				if (minion._mAi != AI_GARG) {
 					minion.ChangeAnimationData(MonsterGraphic::Stand);
-					minion.AnimInfo.CurrentFrame = GenerateRnd(minion.AnimInfo.NumberOfFrames - 1);
+					minion.AnimInfo.setCurrentFrame(GenerateRnd(minion.AnimInfo.NumberOfFrames - 1));
```
k-bar, line 374: I am wondering if it's not a bug.
Shouldn't this be `GenerateRnd(minion.AnimInfo.NumberOfFrames)` ?

Source/monster.cpp
```diff
@@ -1429,22 +1433,22 @@ bool MonsterAttack(int monsterId)
...
 	assert(monster.MData != nullptr);
 
-	if (monster.AnimInfo.CurrentFrame == monster.MData->mAFNum - 1) {
+	if (monster.AnimInfo.isTriggerFrame()) {
 		MonsterAttackPlayer(monsterId, monster._menemy, monster.mHit, monster.mMinDamage, monster.mMaxDamage);
 		if (monster._mAi != AI_SNAKE)
 			PlayEffect(monster, 0);
 	}
-	if (IsAnyOf(monster.MType->mtype, MT_NMAGMA, MT_YMAGMA, MT_BMAGMA, MT_WMAGMA) && monster.AnimInfo.CurrentFrame == 8) {
+	if (IsAnyOf(monster.MType->mtype, MT_NMAGMA, MT_YMAGMA, MT_BMAGMA, MT_WMAGMA) && monster.AnimInfo.getCurrentFrame() == 8) {
```
k-bar, line 1441: hardcoded 2nd attack frame for melee magma (during the same swing)

Source/monster.cpp
```diff
@@ -1429,22 +1433,22 @@ bool MonsterAttack(int monsterId)
...
 		MonsterAttackPlayer(monsterId, monster._menemy, monster.mHit + 10, monster.mMinDamage - 2, monster.mMaxDamage - 2);
 		PlayEffect(monster, 0);
 	}
-	if (IsAnyOf(monster.MType->mtype, MT_STORM, MT_RSTORM, MT_STORML, MT_MAEL) && monster.AnimInfo.CurrentFrame == 12) {
+	if (IsAnyOf(monster.MType->mtype, MT_STORM, MT_RSTORM, MT_STORML, MT_MAEL) && monster.AnimInfo.getCurrentFrame() == 12) {
```
k-bar, line 1445: hardcoded 2nd attack frame for melee storms (during the same swing)

Source/player.cpp
```diff
@@ -1338,7 +1342,7 @@ bool DoSpell(int pnum)
 	}
 	Player &player = Players[pnum];
 
-	if (player.AnimInfo.CurrentFrame == player._pSFNum) {
+	if (player.AnimInfo.getCurrentFrame() == player.AnimInfo.triggerFrame + 1) {
```
k-bar, line 1345: spells are exception.
spell is casted actually 1 frame after triggerFrame.

Source/engine/animationinfo.cpp
```diff
@@ -172,11 +173,12 @@ void AnimationInfo::ChangeAnimationData(std::optional<CelSprite> celSprite, int
 	if (numberOfFrames != NumberOfFrames || ticksPerFrame != TicksPerFrame) {
 		// Ensure that the CurrentFrame is still valid and that we disable ADL cause the calculcated values (for example TickModifier) could be wrong
 		if (numberOfFrames >= 1)
-			CurrentFrame = clamp(CurrentFrame, 0, numberOfFrames - 1);
+			CurrentFrame = clamp(CurrentFrame, 0, (int)lastFrame);
```
ephphatha: Use `static_cast`, not c-style casts. Was this in response to a warning? unit8_t to int promotion is one of the guaranteed good conversions so shouldn't need a cast...

ephphatha:
Can you please rewrite the history to squash related commits and put the clang-tidy changes first? Having 14 commits with very similar names doing partial applications of a change makes it hard to review.

k-bar:
> Can you please rewrite the history to squash related commits and put the clang-tidy changes first? Having 14 commits with very similar names doing partial applications of a change makes it hard to review.

I think that since most changes are straigthforward like case style change, it is best to view as final work together. I could squash them all together. Reversing history would be meaning I had to write things I replaced by helpers in a way I did not write it here.

Source/engine/animationinfo.cpp
k-bar: Ye, I got rid finally of `lastFrame` at the end, if you look at `Changed Files` for review - which I recommend.

k-bar: It turned out, it would be used in really tiny number after I introduced better helpers.

k-bar:
Ye, I checked again.
It is best to inspect the header 1st. Like all those getters/setters/helpers. 
So if header is fine, the rest are derivatives of its changes.
Plus added handling for triggerFrame in the code.

k-bar:
Squashed and rebased on latest changed master.

Source/engine/animationinfo.h
```diff
@@ -37,41 +37,46 @@ enum AnimationDistributionFlags : uint8_t {
...
+	void setCelSprite(const std::optional<CelSprite> &celSprite) { celSprite_ = celSprite; }
+	void setCurrentFrame(const int frameNumber)                  { currentFrame_ = frameNumber; }
+	void setNumberOfFrames(const int numberOfFrames)             { numberOfFrames_ = numberOfFrames; }
+	void setTickCounterOfCurrentFrame(const int tickCounter)     { tickCounterOfCurrentFrame_ = tickCounter; }
+	void setTicksPerFrame(const int ticksPerFrame)               { ticksPerFrame_ = ticksPerFrame; } 
+	void setTriggerFrame(const int triggerFrame)                 { triggerFrame_ = triggerFrame; }
+
+	// helper setters
+	void emplaceCelSprite(const OwnedCelSprite &celSprite)  { celSprite_.emplace(celSprite); }
+	void setCurrentFrameAsLast()                            { currentFrame_ = numberOfFrames_ - 1; }
```
ephphatha, line 65: fairly awkward name for this function, something like `seekLastFrame`, `activateLastFrame`? a matching `seekFirstFrame`/`activateFirstFrame` function would be useful too.

Source/items.cpp
```diff
@@ -3410,18 +3410,18 @@ void ProcessItems()
...
-				item.AnimInfo.CurrentFrame = 0;
-			if (item._iSelFlag == 2 && item.AnimInfo.CurrentFrame == 20)
-				item.AnimInfo.CurrentFrame = 10;
+			if (item._iSelFlag == 1 && item.AnimInfo.getCurrentFrame() == 10)
+				item.AnimInfo.setCurrentFrame(0);
+			if (item._iSelFlag == 2 && item.AnimInfo.getCurrentFrame() == 20)
+				item.AnimInfo.setCurrentFrame(10);
 		} else {
-			if (item.AnimInfo.CurrentFrame == (item.AnimInfo.NumberOfFrames - 1) / 2)
+			if (item.AnimInfo.getCurrentFrame() == (item.AnimInfo.getNumberOfFrames() - 1) / 2)
```
ephphatha, line 3420: Might as well use isTriggerFrame here

Source/items.cpp
```diff
@@ -3410,18 +3410,18 @@ void ProcessItems()
...
+		item.AnimInfo.processAnimation();
 		if (item._iCurs == ICURS_MAGIC_ROCK) {
-			if (item._iSelFlag == 1 && item.AnimInfo.CurrentFrame == 10)
-				item.AnimInfo.CurrentFrame = 0;
-			if (item._iSelFlag == 2 && item.AnimInfo.CurrentFrame == 20)
-				item.AnimInfo.CurrentFrame = 10;
+			if (item._iSelFlag == 1 && item.AnimInfo.getCurrentFrame() == 10)
+				item.AnimInfo.setCurrentFrame(0);
+			if (item._iSelFlag == 2 && item.AnimInfo.getCurrentFrame() == 20)
+				item.AnimInfo.setCurrentFrame(10);
```
ephphatha, line 3418: I'm not sure this condition can ever be hit currently. Magic Rock has an animation length of 10. Starting with a currentFrame of 10 means it's immediately hitting the >= lastFrame condition in processAnimation and it gets reset to frame 0.

ephphatha (commented):
I haven't had time to go through it fully yet, so there's potentially more call sites which could be cleaned up or provide further context around the logic to put in helpers.

Source/engine/animationinfo.h
```diff
@@ -37,41 +37,46 @@ enum AnimationDistributionFlags : uint8_t {
...
+	[[nodiscard]] const std::optional<CelSprite>  &getCelSprite() const { return celSprite_; }
+	[[nodiscard]] int getCurrentFrame()                           const { return currentFrame_; }
+	[[nodiscard]] int getNumberOfFrames()                         const { return numberOfFrames_; }
+	[[nodiscard]] int getTickCounterOfCurrentFrame()              const { return tickCounterOfCurrentFrame_; }
+	[[nodiscard]] int getTicksPerFrame()                          const { return ticksPerFrame_; }
+	[[nodiscard]] int getTriggerFrame()                           const { return triggerFrame_; }
+
+	// helper getters
+	[[nodiscard]] bool isLastFrame()         const { return (currentFrame_ == numberOfFrames_ - 1); }
+	[[nodiscard]] bool isOver()              const { return (currentFrame_ >= numberOfFrames_ - 1); }
```
obligaron, line 51: I think both could be combined to one function.
Cause `currentFrame` should never be more then `numberOfFrames` else it's a bug.

Another note:
`isOver` is not 100% correct.
If you have a animation with 5 `NumberOfFrames` and each frame has a `TicksPerFrame` of 10.
Then if you have `currentFrame`=5 and `tickCounterOfCurrentFrame` = 0. The animation is only over after 9 more game ticks. 😉 
It's not a problem in the current codebase, cause on the locations where this check is performed all animations have `TicksPerFrame` = 1.
A solution would be to also check for `tickCounterOfCurrentFrame`, cause it shouldn't matter.
But better check twice, it was a little while away when I last checked all animations. 😁

Source/engine/animationinfo.h
```diff
@@ -37,41 +37,46 @@ enum AnimationDistributionFlags : uint8_t {
...
+	[[nodiscard]] bool isTriggerFrame()      const { return (currentFrame_ == triggerFrame_); }
+	[[nodiscard]] bool isAfterTriggerFrame() const { return (currentFrame_  > triggerFrame_); }
+
+	// generic setters
+	void setCelSprite(const std::optional<CelSprite> &celSprite) { celSprite_ = celSprite; }
+	void setCurrentFrame(const int frameNumber)                  { currentFrame_ = frameNumber; }
+	void setNumberOfFrames(const int numberOfFrames)             { numberOfFrames_ = numberOfFrames; }
+	void setTickCounterOfCurrentFrame(const int tickCounter)     { tickCounterOfCurrentFrame_ = tickCounter; }
+	void setTicksPerFrame(const int ticksPerFrame)               { ticksPerFrame_ = ticksPerFrame; } 
+	void setTriggerFrame(const int triggerFrame)                 { triggerFrame_ = triggerFrame; }
```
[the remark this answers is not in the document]
obligaron: This is a codestyle question, but still adding my own opinion:
I'm not in favor of adding simple getter/setter. They make the code harder to read and you have to write more code. (It would be something else if some external library would depend on our stuff.)

Small anecdote: When initially writing `AnimationInfo` I thought about making all members private and have only two methods that change/init a animation:
`SetNewAnimation` and `LoadAnimationFromSave`. But then I discovered more and more special cases. So I'm happy to see that this is picked up again.

Source/engine/animationinfo.h
```diff
@@ -83,45 +88,72 @@ class AnimationInfo {
...
+	 */
+	int ticksPerFrame_;
+	/**
+	 * @brief Number of game ticks after the current animation sequence started
+	 */
+	float ticksSinceSequenceStarted_;
+	/**
+	 * @brief Frame that triggers additional action
+	 */
+	int triggerFrame_;
```
obligaron, line 152: I think this is a good idea, but imho we shouldn't add this:
- It makes `AnimationInfo` larger (see #4849)
- It is only relevant for Players and irrelevant, perhaps confusing, for monsters and items (there I disocvered we never changed objects to `AnimationInfo` 😳)
- When changing the player mode, you also have to manually update the trigger frame. If they get out of sync => not good.

k-bar: It is relevant for players AND monsters
I made this update everytime just after set new animation - set in the same function that is called after player/monster mode is changed. So rather no decync should be here.
Go on - click `Files changed` and search for `trigger` to see where it applies.

I understand though, you are trying to minimalize AnimationInfo size.

I also do not like how magmas and storms 2nd trigger frame is actually hardcoded.
I think there could be some better solution to handle this. Perhaps not making this array of trigger frames only just for those 8 type of monsters (+ their bosses)? This 2nd trigger frame of them is not in MonstersData[] ...

Source/engine/animationinfo.h
k-bar: :shrug: what to do. Plenty of this condition in the code right now...
`currentFrame_ >= numberOfFrames_ - 1`
thus this helper. I think since there is distinction like that - I just maintained it. It doesn't change current behavior, nor make it better, than just add some readability.
I can imagine there are some cases where it exactly have to be equal. Changing this to equal or greater could maybe introduce some bugs, or some events occur more often, so I didn't want to touch the logic behind.

Source/engine/animationinfo.h
obligaron: Sorry my bad. I only did a fast review on the header and didn't notice that github didn't load the diff for `monster.cpp`.

Anyway the other problems stay the same.
Especially the last point concerns me (we have to ensure trigger frame and player/monster mode don't differ).
That's why we also needs the adjustments in `LoadPlayer` or `SyncMonsterAnim`.
Do we know in all places where we check the trigger frame in what player/monster mode we are in? Okay checked again `CanChangeAction` would really be simpler with a triggerFrame.

Another thing to think about is changing how the trigger frames are currently stored.
Before #4221 the check was
```cpp
if (player.AnimInfo.CurrentFrame >= player._pWFrames) {
```
now it's
```cpp
`if (player.AnimInfo.CurrentFrame >= player._pWFrames - 1) {`
```
Cause all frame indexes are zero-based. But we didn't updated the trigger frames. If we reduce all trigger frames by 1, we can have
```cpp
if (player.AnimInfo.CurrentFrame >= player._pWFrames) {
```
back. And I think that's much more readable. What do you think? 🙂 

Perhaps a `bool isAfterTriggerFrame(int8_t triggerFrame)` would help to clariffy it, even if we don't store it in `AnimationInfo`? 🤔 

There are many hardcoded places in diablo. If we could make them more generic/modding friendly that would be great. But I think that's often not a easy task.

Source/engine/animationinfo.h
obligaron: > I can imagine there are some cases where it exactly have to be equal.
I think we are pretty safe to change this. 🙂 
All occurrences that change `currentFrame` don't set it to something great then `numberOfFrames`.
But perhaps I miss something.

> Changing this to equal or greater could maybe introduce some bugs, or some events occur more often, so I didn't want to touch the logic behind.
I'm more concerned about `triggerFrame`, cause it needs to be updated every time we change the mode. 😉
But this is another discussions.

Source/engine/animationinfo.h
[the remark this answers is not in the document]
k-bar: > So I'm happy to see that this is picked up again.

This move to private was done only to conform clang-tidy rule, that was lifted after discussion on discord.
Since it's gone, I will revert it.

Source/engine/animationinfo.h
k-bar: `seekLastFrame` sounds awkward to me As it would be good for some function that returns it. I could agree to `activateLastFrame`... but maybe there will be even better propositions, like you know activate is like set something to active disactivate - non active state...? Actually `setCurrentFrameToLast` maybe? leaves not doubt for me. What are other options?

Source/items.cpp
k-bar: How exactly? it is kind of split number of frames in half. There is no setting TriggerFrame for items.

ephphatha: Monster attack animations have a trigger frame where damage is applied and the associated sound effects are played, same for player attack animations.

Item drop animations play a sound effect on this frame, seems like a trigger frame to me.

Source/engine/animationinfo.h
ephphatha: seek was based on the convention for file I/O methods, but there's probably something more suitable in the context of an animation sequence.

PR #4819, opened 2022-06-29
Take Player reference in FixPlrWalkTags

ephphatha:
Untested, not sure if there's still situations where player positions can get so out of sync in multiplayer games that a player disconnecting leaves an invalid index in `dPlayers`

### DIFF
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -237,7 +237,7 @@ void PlayerLeftMsg(int pnum, bool left)
 		return;
 	}
 
-	RemovePlrFromMap(pnum);
+	FixPlrWalkTags(player);
 	RemovePortalMissile(pnum);
 	DeactivatePortal(pnum);
 	delta_close_portal(pnum);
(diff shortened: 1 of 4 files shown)


ephphatha:
From #4874:
> Unless someone fixes `RemovePlrFromMap()` ...

Still haven't had time to playtest this, but it does completely remove `RemovePlrFromMap()` :D

Source/multi.cpp
```diff
@@ -237,7 +237,7 @@ void PlayerLeftMsg(int pnum, bool left)
 		return;
 	}
 
-	RemovePlrFromMap(pnum);
+	FixPlrWalkTags(player);
```
AJenbo, line 240: This does make the assumption that there was never a need to scan the full map for a bad player position. But maybe we should just track down such cases and fix them if they exists.

AJenbo:
I tried loading up a MP game and disconnect while walking. Then checked if i was able to target anything at the last know location or walk there. Everything worked as expected. Not sure if this is a good test or not, but couldn't think of anything better.

PR #4820, opened 2022-06-29
:recycle: Stores the monsterId inside the monster struct

canassa:
Hello,

I was trying to figure out how to remove the `monsterId` from the `StartWalk` functions parameter but the only solution I could come up with was storing the `monsterId` inside the `Monster` struct itself.

I am not sure if this is a good approach, but I tested it on both single-player and multiplayer and looks like nothing broke.

Please let me know what you think 😄

### DIFF
--- a/Source/monster.h
+++ b/Source/monster.h
@@ -168,7 +168,6 @@ struct CMonster {
 extern CMonster LevelMonsterTypes[MaxLvlMTypes];
 
 struct Monster { // note: missing field _mAFNum
-	int monsterId;
 	const char *mName;
 	std::unique_ptr<uint8_t[]> uniqueTRN;
 	AnimationInfo AnimInfo;
@@ -269,6 +268,13 @@ struct Monster { // note: missing field _mAFNum
 	{
 		return *type().data;
 	}
+
+	/**
+	 * @brief Returns the network identifier for this monster
+	 *
+	 * This is currently the index into the Monsters array, but may change in the future.
+	 */
+	[[nodiscard]] size_t getId() const;
 
 	/**
 	 * @brief Is the monster currently walking?
--- a/Source/monster.h
+++ b/Source/monster.h
@@ -168,6 +168,7 @@ struct CMonster {
 extern CMonster LevelMonsterTypes[MaxLvlMTypes];
 
 struct Monster { // note: missing field _mAFNum
+	int monsterId;
 	const char *mName;
 	std::unique_ptr<uint8_t[]> uniqueTRN;
 	AnimationInfo AnimInfo;
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -188,8 +188,9 @@ void InitMonsterTRN(CMonster &monst)
 	}
 }
 
-void InitMonster(Monster &monster, Direction rd, int mtype, Point position)
+void InitMonster(Monster &monster, Direction rd, int mtype, Point position, int monsterId)
 {
+	monster.monsterId = monsterId;
 	monster._mdir = rd;
 	monster.position.tile = position;
 	monster.position.future = position;
@@ -310,7 +311,7 @@ void PlaceMonster(int i, int mtype, Point position)
 	dMonster[position.x][position.y] = i + 1;
 
 	auto rd = static_cast<Direction>(GenerateRnd(8));
-	InitMonster(Monsters[i], rd, mtype, position);
+	InitMonster(Monsters[i], rd, mtype, position, i);
 }
 
 void PlaceGroup(int mtype, int num, UniqueMonsterPack uniqueMonsterPack, int leaderId)
@@ -808,14 +809,12 @@ void StartSpecialStand(Monster &monster, Direction md)
 	monster.position.old = monster.position.tile;
 }
 
-void StartWalk(int monsterId, int xvel, int yvel, int xadd, int yadd, Direction endDir)
+void StartWalk(Monster &monster, int xvel, int yvel, int xadd, int yadd, Direction endDir)
 {
-	auto &monster = Monsters[monsterId];
-
 	const auto fx = static_cast<WorldTileCoord>(xadd + monster.position.tile.x);
 	const auto fy = static_cast<WorldTileCoord>(yadd + monster.position.tile.y);
 
-	dMonster[fx][fy] = -(monsterId + 1);
+	dMonster[fx][fy] = -(monster.monsterId + 1);
 	monster._mmode = MonsterMode::MoveNorthwards;
 	monster.position.old = monster.position.tile;
 	monster.position.future = { fx, fy };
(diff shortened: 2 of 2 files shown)


k-bar:
Would be possible to use 
MonsterAtPosition
from here:  https://github.com/diasurgical/devilutionX/pull/4816
instead?

AJenbo:
Only for comparisons, but this is doing an assignment.

ephphatha:
I haven't had the time to go through this properly but from a quick skim I think this approach is fine.

The `Object::getId()` implementation is a dodgy hack that can be removed if we make all network messages and actions related to objects use position instead. That is made much easier because objects don't move.

The need to have a single entity occupy two tiles while moving makes monsters and players annoying without significant refactoring, this PR at least lets us restrict uses of the ID to where it is strictly required.

Source/monster.h
```diff
@@ -165,6 +165,7 @@ struct CMonster {
 };
 
 struct Monster { // note: missing field _mAFNum
+	int monsterId;
```
glebm, line 168: This makes the monster struct larger for no reason, I'd rather we don't do it like this.

AJenbo: @glebm would `std::distance` be an option instead?

canassa: I am not sure how `std::distance` works (not a C++ dev) but wouldn't it need to loop over `Monsters` until it finds the index? That would be much slower than just retrieving by index I think

glebm: `std::distance` is constant-time on random access iterators and pointers.

Still, if you need the ID, just pass the ID IMO.

ephphatha: `std::distance(first, last)` is "implementation defined" (magic :D) depending on what the types of `first` and `last` are. With `Monsters` being a simple c-style array and `this` being a pointer the implementation *should* be a subtraction of one pointer from another. If monsters ever changes to a `std::list` then we'd be forced to use bidirectional iterators and walk the chain (assuming you can even get an iterator to the current element without already walking the chain).

for example msvc uses the following definition:
```cpp
template <class _InIt>
_NODISCARD _CONSTEXPR17 _Iter_diff_t<_InIt> distance(_InIt _First, _InIt _Last) {
    if constexpr (_Is_random_iter_v<_InIt>) {
        return _Last - _First; // assume the iterator will do debug checking
    } else {
        _Adl_verify_range(_First, _Last);
        auto _UFirst             = _Get_unwrapped(_First);
        const auto _ULast        = _Get_unwrapped(_Last);
        _Iter_diff_t<_InIt> _Off = 0;
        for (; _UFirst != _ULast; ++_UFirst) {
            ++_Off;
        }

        return _Off;
    }
}
```
Pointers meet the criteria for random access iterators, so the simple case gets picked.

canassa: That's quite cool! I am learning a lot 😄

canassa: Let me know which solution you think it's best and I can change the MR.

vlolteanu: You can still have `monsterId` and not make the struct any larger. `uint16_t` is sufficient as its type, and you can place it where the compiler would add padding otherwise, e.g. before `pathcount`.

vlolteanu: Come to think about it, there are a few tens of monsters per level, so we are haggling over 100-200 bytes maybe. It seems a bit silly.

glebm: >  there are a few tens of monsters per level

The global `Monsters` array is currently at 56 KiB!

I'm of the "pass the ID if you want the ID" opinion not just for the size reasons but also because it's good to explicitly communicate that the function needs the ID specifically (everything else should pass by reference but that's an ongoing parallel refactoring).

AJenbo: `MAXMONSTERS` = 200. int = 4 bytes, so 800 bytes i guess.

AJenbo: > Very few functions actually need the ID.

The problem is that the few functions are at the end of a call chains leading to all of them having to pass the ID down the stack.

glebm: > The problem is that the few functions are at the end of a call chains leading to all of them having to pass the ID down the stack.

If there are a lot more functions than what's addressed in this PR then this is the implementation that I'd prefer:

```c++
size_t Monster::getId() const { return std::distance(&Monsters[0], this); }
```

Less stuff to keep track of.

vlolteanu: The extra memory usage was an afterthought. `struct Monster` still has padding where you can place an `uint16_t`, without enlarging it.

AJenbo: I think this is actually the case for the 60 renaming monster related functions. the of which 17 appears to use the ID.

vlolteanu: I'd place it outside of the class, i.e..:
```C++
size_t getMonsterId(Cmonster *monster) const { return std::distance(&Monsters[0], monster); }
```

canassa: The std::distance proposal seems a better solution than adding a new field to the struct. It’s less state to keep around and state adds complexity.

canassa: Bump.

I updated the PR. I removed the m`monsterId` from the struct and implemented the `GetId` method.

I play-tested the PR and it seems to be working

Source/monster.h
```diff
@@ -260,6 +260,13 @@ struct Monster { // note: missing field _mAFNum
 	 */
 	void Petrify();
 
+	/**
+	 * @brief Returns the network identifier for this monster
+	 *
+	 * This is currently the index into the Monsters array, but may change in the future.
+	 */
+	size_t GetId();
```
ephphatha, line 268: `readability-identifier-naming.MethodCase: camelBack` The existing methods use the wrong case style and haven't been fixed yet, new methods should use the new style. Also probably want to address modernize-use-nodiscard. Finally, this method doesn't change state so should be declared const.
```suggestion
	[[nodiscard]] size_t getId() const;
```

canassa: Hi, the compiler throws this error when I add `const`:

```
error: no matching function for call to 'distance'
        return std::distance(&Monsters[0], this);
```

It could be related to the iterator being mutable

ephphatha: `return std::distance(const_cast<const Monster *>(&Monsters[0]), this);`

ephphatha: https://en.cppreference.com/w/cpp/utility/as_const could also be useful if it's supported on GCC6.5? haven't tested that

AJenbo: From the changelog as_const appears to have been added in 7.0

glebm: `std::distance<const Monster *>(&Monsters[0], this);` should also work

AJenbo: yep, already doing it for players

canassa: PR updated

AJenbo (approved):

PR #4824, opened 2022-06-30
Non-int versions of `Point` and `Displacement`

glebm:
This will allow us to make some structs, such as `ActorPosition`, much smaller.

`sizeof(Monsters)`: 56K -> 40K

### DIFF
--- a/Source/engine/actor_position.hpp
+++ b/Source/engine/actor_position.hpp
@@ -1,25 +1,26 @@
 #pragma once
 
 #include "engine/point.hpp"
+#include "engine/world_tile.hpp"
 
 namespace devilution {
 
 struct ActorPosition {
-	Point tile;
+	WorldTilePosition tile;
 	/** Future tile position. Set at start of walking animation. */
-	Point future;
+	WorldTilePosition future;
 	/** Tile position of player. Set via network on player input. */
-	Point last;
+	WorldTilePosition last;
 	/** Most recent position in dPlayer. */
-	Point old;
-	/** Pixel offset from tile. */
-	Displacement offset;
-	/** Same as offset but contains the value in a higher range */
-	Displacement offset2;
-	/** Pixel velocity while walking. Indirectly applied to offset via _pvar6/7 */
-	Displacement velocity;
+	WorldTilePosition old;
 	/** Used for referring to position of player when finishing moving one tile (also used to define target coordinates for spells and ranged attacks) */
-	Point temp;
+	WorldTilePosition temp;
+	/** Pixel offset from tile. */
+	DisplacementOf<int8_t> offset;
+	/** Same as offset but contains the value in a higher range */
+	DisplacementOf<int16_t> offset2;
+	/** Pixel velocity while walking. Indirectly applied to offset via _pvar6/7 */
+	DisplacementOf<int16_t> velocity;
 };
 
 } // namespace devilution
--- a/Source/engine/displacement.hpp
+++ b/Source/engine/displacement.hpp
@@ -11,141 +11,102 @@
 
 namespace devilution {
 
-struct Displacement {
-	int deltaX;
-	int deltaY;
+template <typename DeltaT>
+struct DisplacementOf;
 
-	Displacement() = default;
+using Displacement = DisplacementOf<int>;
 
-	constexpr Displacement(int deltaX, int deltaY)
+template <typename DeltaT>
+struct DisplacementOf {
+	DeltaT deltaX;
+	DeltaT deltaY;
+
+	DisplacementOf() = default;
+
+	template <typename DisplacementDeltaT>
+	constexpr DisplacementOf(DisplacementOf<DisplacementDeltaT> other)
+	    : deltaX(other.deltaX)
+	    , deltaY(other.deltaY)
+	{
+	}
+
+	constexpr DisplacementOf(DeltaT deltaX, DeltaT deltaY)
 	    : deltaX(deltaX)
 	    , deltaY(deltaY)
 	{
 	}
 
-	explicit constexpr Displacement(int delta)
+	explicit constexpr DisplacementOf(DeltaT delta)
 	    : deltaX(delta)
 	    , deltaY(delta)
 	{
 	}
 
-	explicit constexpr Displacement(const Size &size)
+	explicit constexpr DisplacementOf(const Size &size)
 	    : deltaX(size.width)
 	    , deltaY(size.height)
 	{
 	}
 
-	explicit constexpr Displacement(Direction direction)
-	    : Displacement(fromDirection(direction))
+	explicit constexpr DisplacementOf(Direction direction)
+	    : DisplacementOf(fromDirection(direction))
 	{
 	}
 
-	constexpr bool operator==(const Displacement &other) const
+	template <typename DisplacementDeltaT>
+	constexpr bool operator==(const DisplacementOf<DisplacementDeltaT> &other) const
 	{
 		return deltaX == other.deltaX && deltaY == other.deltaY;
 	}
 
-	constexpr bool operator!=(const Displacement &other) const
+	template <typename DisplacementDeltaT>
+	constexpr bool operator!=(const DisplacementOf<DisplacementDeltaT> &other) const
 	{
 		return !(*this == other);
 	}
 
-	constexpr Displacement &operator+=(const Displacement &displacement)
+	template <typename DisplacementDeltaT = DeltaT>
+	constexpr DisplacementOf<DeltaT> &operator+=(DisplacementOf<DisplacementDeltaT> displacement)
 	{
 		deltaX += displacement.deltaX;
 		deltaY += displacement.deltaY;
 		return *this;
 	}
 
-	constexpr Displacement &operator-=(const Displacement &displacement)
+	template <typename DisplacementDeltaT = DeltaT>
+	constexpr DisplacementOf<DeltaT> &operator-=(DisplacementOf<DisplacementDeltaT> displacement)
 	{
 		deltaX -= displacement.deltaX;
 		deltaY -= displacement.deltaY;
 		return *this;
 	}
 
-	constexpr Displacement &operator*=(const int factor)
+	constexpr DisplacementOf<DeltaT> &operator*=(const int factor)
 	{
 		deltaX *= factor;
 		deltaY *= factor;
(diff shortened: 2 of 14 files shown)


Source/engine/actor_position.hpp
```diff
@@ -5,21 +5,21 @@
 namespace devilution {
 
 struct ActorPosition {
-	Point tile;
+	PointOf<uint8_t> tile;
```
ephphatha, line 8: `using WorldTile = PointOf<uint8_t>` would be useful

glebm: Done

Source/player.cpp
```diff
@@ -501,7 +502,7 @@ player_graphic GetPlayerGraphicForSpell(spell_id spellId)
 	}
 }
 
-void StartSpell(int pnum, Direction d, int cx, int cy)
```
AJenbo, line 504: This looks wrong. Should probably be:
```cpp
void StartSpell(int pnum, Direction d, WorldTileCoord position)
```

glebm: I'm not sure what you mean, I've simply changed `int` to `WorldTileCoord` to remove the need for `static_cast` inside the function

AJenbo: arh, i didn't notice that there where two types here.

Source/engine/displacement.hpp
```diff
@@ -11,143 +11,104 @@
 
 namespace devilution {
 
-struct Displacement {
-	int deltaX;
-	int deltaY;
+template <typename DeltaT>
+struct DisplacementOf;
```
isage, line 15: Maybe `DisplacementOff` would be better? off is more common shortening of offset, and DisplacementOf reads as 'displacement of something'

StephenCWills: I assume it's supposed to be `DisplacementOf<int>` = "displacement of int" in a similar fashion to how you might read `std::vector<int>` as "vector of int". The word "of" seems to be a bit redundant in this case, but that's probably because it would otherwise conflict with the alias established by `using Displacement = DisplacementOf<int>;`

glebm: Yep, that's exactly it

Source/engine/point.hpp
```diff
@@ -12,99 +13,81 @@
...
-	int y;
+template <typename CoordT>
+struct PointOf;
 
-	Point() = default;
+using Point = PointOf<int>;
 
-	constexpr Point(int x, int y)
+template <typename CoordT>
+struct PointOf {
```
vlolteanu, line 22: How about this?

```suggestion
template <typename CoordT = int>
struct Point {
```

Same goes for `DisplacementOf`.

glebm: For such types, `Point a;` is only legal in C++17. We're somewhere close to C++11 on the language support.

glebm:
Rebased and migrated CrawlTable

Source/player.cpp
```diff
@@ -417,7 +418,7 @@ void ChangeOffset(Player &player)
 		player.position.offset2 += player.position.velocity;
 	}
 
-	player.position.offset = { player.position.offset2.deltaX >> 8, player.position.offset2.deltaY >> 8 };
+	player.position.offset = DisplacementOf<int8_t> { static_cast<int8_t>(player.position.offset2.deltaX >> 8), static_cast<int8_t>(player.position.offset2.deltaY >> 8) };
```
AJenbo, line 421: Nit: This is becomming a bit hard to read... well it wasn't nice before either.
```suggestion
	player.position.offset = DisplacementOf<int8_t> {
		static_cast<int8_t>(player.position.offset2.deltaX >> 8),
		static_cast<int8_t>(player.position.offset2.deltaY >> 8)
	};
```

Source/msg.cpp
```diff
@@ -2405,12 +2405,13 @@ void DeltaLoadLevel()
...
 
 			M_ClearSquares(i);
-			int x = deltaLevel.monster[i]._mx;
-			int y = deltaLevel.monster[i]._my;
 			auto &monster = Monsters[i];
-			monster.position.tile = { x, y };
-			monster.position.old = { x, y };
-			monster.position.future = { x, y };
+			{
+				const WorldTilePosition position { deltaLevel.monster[i]._mx, deltaLevel.monster[i]._my };
```
AJenbo, line 2410: Could you change `DMonsterStr` to use `WorldTilePosition`, I don't think it's used a lot :)

Source/monster.cpp
```diff
@@ -844,22 +845,22 @@ void StartWalk2(int monsterId, int xvel, int yvel, int xoff, int yoff, int xadd,
...
 	auto &monster = Monsters[monsterId];
 
-	int fx = xadd + monster.position.tile.x;
-	int fy = yadd + monster.position.tile.y;
-	int x = mapx + monster.position.tile.x;
-	int y = mapy + monster.position.tile.y;
+	const auto fx = static_cast<WorldTileCoord>(xadd + monster.position.tile.x);
+	const auto fy = static_cast<WorldTileCoord>(yadd + monster.position.tile.y);
+	const auto x = static_cast<WorldTileCoord>(mapx + monster.position.tile.x);
+	const auto y = static_cast<WorldTileCoord>(mapy + monster.position.tile.y);
```
AJenbo, line 863: Would updating the attribute type not avoid the casting here?

Source/lighting.cpp
```diff
@@ -82,12 +82,11 @@ std::vector<Displacement> CrawlRow(int row)
...
-	std::vector<std::vector<Displacement>> ret;
-	for (int i = 0; i < 19; i++)
-		ret.push_back(CrawlRow(i));
-	return ret;
-}();
+const std::array<std::vector<DisplacementOf<int8_t>>, MaxCrawlRow + 1> CrawlTable {
+	std::vector<DisplacementOf<int8_t>> { { static_cast<int8_t>(0), static_cast<int8_t>(0) } },
+	CrawlRow(1), CrawlRow(2), CrawlRow(3), CrawlRow(4), CrawlRow(5), CrawlRow(6), CrawlRow(7), CrawlRow(8), CrawlRow(9),
+	CrawlRow(10), CrawlRow(11), CrawlRow(12), CrawlRow(13), CrawlRow(14), CrawlRow(15), CrawlRow(16), CrawlRow(17), CrawlRow(18)
+};
```
AJenbo, line 89: Are we porting to BIOSs now or why are we unrolling this loop?

glebm: Fixed

Source/engine/displacement.hpp
```diff
@@ -200,14 +161,14 @@ struct Displacement {
...
-	Displacement worldToNormalScreen() const
+	[[nodiscard]] Displacement worldToNormalScreen() const
 	{
 		// Most transformations between world and screen space take shortcuts when scaling to simplify the math. This
 		//  routine is typically used with missiles where we want a normal vector that can be multiplied with a target
 		//  velocity (given in pixels). We could normalize the vector first but then we'd need to scale it during
 		//  rotation from world to screen space. To save performing unnecessary divisions we rotate first without
 		//  correcting the scaling. This gives a vector in elevation projection aligned with screen space.
-		Displacement rotated { (deltaY - deltaX), -(deltaY + deltaX) };
+		Displacement rotated { static_cast<int>(deltaY) - static_cast<int>(deltaX), -(static_cast<int>(deltaY) + static_cast<int>(deltaX)) };
```
ephphatha, line 171: I think you can avoid the cast here by allowing this to be the current type, it's only the return value from `DisplacementOf<T>::normalized()` that needs an explicit type.
```suggestion
		DisplacementOf<DeltaT> rotated { deltaY - deltaX, -(deltaY + deltaX) };
```

though I guess it needs a static assert like other operations expecting DeltaT to be signed, I think having the assert is preferable personally but happy to go either way.

Source/monster.cpp
glebm: The xadd/mapx are `int` and can be negative, so it wouldn't by itself. Not sure if changing them to `int8_t` would but that's probably for another refactoring anyway.

Source/msg.cpp
glebm: Done

Source/player.cpp
glebm: Done

AJenbo: Thanks, now it fits on my 80 column monitor again :+1:

Source/engine/point.hpp
```diff
@@ -134,61 +118,107 @@ struct Point {
...
+		Displacement offset { abs(*this - other) };
 
 		return std::max<int>(offset.deltaX, offset.deltaY);
 	}
 
 	/**
 	 * @brief Converts a coordinate in megatiles to the northmost of the 4 corresponding world tiles
 	 */
-	constexpr Point megaToWorld() const
+	constexpr PointOf<CoordT> megaToWorld() const
```
ephphatha, line 149: One of the things I was hoping we could do in the future is make these functions part of the WorldTilePosition (and maybe a future MegaTilePosition?) specialisations instead of the base template. However not sure if that would be considered desirable given multi-use variables like quest positions storing both values at different stages of dungeon generation/quest progression.

Idle thought for the future anyway.

AJenbo (approved):

ephphatha (commented):
I can see the logic behind including the reordering of Monster member variables and adding explicit backing types to those enums, was a bit unexpected given the PR title but the commit message explains the reason at least :D

Consider this approval, giving a chance to address the nit before auto-merge kicks in

Source/engine/displacement.hpp
ephphatha: I was too slow 😂 . Followup maybe?

AJenbo:
Damn, didn't intend to suqash, this I was trying to hack the GitHub UI since there was still 6 PRs left before AppVeyor would come around to building this one :(

Source/engine/displacement.hpp
glebm: I've added the casts because of the compiler warnings but feel free to experiment and send a follow-up now that this is merged!

Source/engine/point.hpp
glebm: Sounds good, I was thinking the same thing, though rather than specializations a `struct WorldTilePosition : public Point<uint8_t>`

Issue #4669, opened 2022-06-06
Berserked ranged monster missile will always miss another not-berserked monster.


k-bar:
**Important information**
Original Hellfire bug

**Describe**
Seems like in original version of Hellfire there is a bug that prevents a missile originating from berserking monster hit other unberserked monster.

**To Reproduce**
1. Berserk a monster
2. Watch it missiles always miss another monster (unless it is also berserked).
![berserkRangedBad](https://user-images.githubusercontent.com/92504151/172091035-256fc891-7173-4ed0-8645-255654b84179.gif)

**Expected behavior**
Fixed in #4618:
![berserkRangedGood](https://user-images.githubusercontent.com/92504151/172091011-ee783579-5185-48fd-b769-dd079403ae45.gif)

PR #4826, opened 2022-06-30
Missile from berserked monster try to hit every monster

k-bar:
Thanks to this change, berserking ranged monster's missiles will try to hit all monsters they will collide with, not only 1 monster that currently is on target.

This is modification of fix for #4669 introduced in #4800

(First introduced as original fix #4618 for #4669)


AJenbo (approved):
Thanks :)

PR #4827, opened 2022-06-30
Monster global constants cleanup

aetn23:
Changed defines in files monster.cpp and monster.h to be constinit const int instead. Fixed some typos along the way. Changed passing the reference to the first element of an array to simply passing the array itself or its underlying data in the InitMonsterGFX.

Removing defines is good practice in modern C++. Other changes simplify code, or fix language errors in the comments.

### DIFF
--- a/Source/monster.h
+++ b/Source/monster.h
@@ -25,8 +25,8 @@ namespace devilution {
 
 struct Missile;
 
-#define MAXMONSTERS 200
-#define MAX_LVLMTYPES 24
+constinit const int MaxMonsters = 200;
+constinit const int maxLvlmtypes = 24;
 
 enum monster_flag : uint16_t {
 	// clang-format off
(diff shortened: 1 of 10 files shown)


Source/monster.h
```diff
@@ -25,8 +25,8 @@ namespace devilution {
 
 struct Missile;
 
-#define MAXMONSTERS 200
-#define MAX_LVLMTYPES 24
+constinit const int MaxMonsters = 200;
+constinit const int maxLvlmtypes = 24;
```
qndel, line 29: why lowercase

AJenbo:
`constinit` is C++20, we currently have a requirement slightly below C++17.

AJenbo (approved):
Nice :+1:

Source/monster.h
```diff
@@ -25,8 +25,8 @@ namespace devilution {
 
 struct Missile;
 
-#define MAXMONSTERS 200
-#define MAX_LVLMTYPES 24
+constexpr const int MaxMonsters = 200;
+constexpr const int MaxLvlMTypes = 24;
```
ephphatha, line 29: constexpr implies const, no need to repeat in these definitions.

AJenbo: @aetn23 would you be willing to do a follow up to correct this :)

aetn23: Sure

PR #4829, opened 2022-06-30
:bug: fix town boundaries

canassa:
Hello,

I noticed that the town boundaries were a bit weird:
Screenshot 2022-06-30 at 23 23 27

So I did a git bisect and I found that the issue was introduced at 4cc3a5264ce75511163081ee9372ca4f7fd1c1cc, this is how it was before:
Screenshot 2022-06-30 at 23 12 45

I am not sure if the change was intentional or not, but I looked into the tiles and I found that 426 is a black tile, it looks like the original:
Screenshot 2022-06-30 at 23 26 25

### DIFF
--- a/Source/levels/town.cpp
+++ b/Source/levels/town.cpp
@@ -187,10 +187,10 @@ void DrlgTPass3()
 {
 	for (int yy = 0; yy < MAXDUNY; yy += 2) {
 		for (int xx = 0; xx < MAXDUNX; xx += 2) {
-			dPiece[xx][yy] = 218;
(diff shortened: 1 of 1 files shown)


AJenbo:
As long as `426` is considered solid, else this will cause you to be able to walk through some houses and other obscured parts of town :)

You can check it ingame by typing `tiledata solid`

canassa:
This is the result of `tiledata solid`

Screenshot 2022-07-01 at 00 29 07

AJenbo:
Thanks, looks good :+1: 
3 means that it also blocks light.

PR #4831, opened 2022-06-30
Fix black background palette

glebm:
Diablo and Hellfire have different palettes in `ui_art\black.pcx`.

We now ship our own tiny `black.pcx` (as of https://github.com/diasurgical/devilutionX/pull/4746) that takes priority over the other ones.

The `black.pcx` that we ship has the Diablo palette. This resulted in an incorrect palette in the Settings screen, because Diablo and Hellfire logos use the upper part of the palette which are different.

Split our `black.pcx` file into `black_diablo.pcx` and `black_hellfire.pcx` and load the correct one depending on the mode.


Source/DiabloUI/diabloui.cpp
```diff
@@ -672,6 +672,11 @@ Sint16 GetCenterOffset(Sint16 w, Sint16 bw)
 	return (bw - w) / 2;
 }
 
+void UiLoadBlackBackground()
+{
+	LoadBackgroundArt(gbIsHellfire ? "ui_art\\black_hellfire.pcx" : "ui_art\\black_diablo.pcx");
```
AJenbo, line 677: Wouldn't this work just as well?
```suggestion
	LoadBackgroundArt(gbIsHellfire ? "ui_art\\black.pcx" : "ui_art\\black_diablo.pcx");
```
That way we don't need an embeded file for Hellfire.

glebm: Our embedded file is much smaller than what's shipped with Hellfire, at only 899 bytes vs 11.5 KiB, so that's a nice advantage of shipping our own as well

AJenbo: Speaking of background issues in the settings men :thinking: 

image

AJenbo: > Our embedded file is much smaller than what's shipped with Hellfire, at only 899 bytes vs 11.5 KiB, so that's a nice advantage of shipping our own as well

Is it worth it, since it will be unloaded before entering the game?

glebm: 1. Loading `ui_art/black.pcx` won't work well if `gbIsHellfire` is true and there is no `hellfire.mpq`.

2. The UI glitch always happens at 640x480 (because of how `UiClearScreen()` works together with our 1px black background).

I'll fix the glitch and clean up the black background stuff a bit more.

Memory thrashing before starting a game can still impact the game on super memory-constrained devices because of malloc fragmentation etc.

glebm: Done

Source/DiabloUI/diabloui.cpp
```diff
@@ -766,7 +779,7 @@ void DrawSelector(const SDL_Rect &rect)
 
 void UiClearScreen()
 {
-	if (gnScreenWidth > 640) // Background size
+	if (!ArtBackground || gnScreenWidth > PcxSpriteSheet { *ArtBackground }.width())
```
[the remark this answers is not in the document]
glebm: Done

AJenbo (approved):

PR #4833, opened 2022-07-01
Refactor CrawlTable

vlolteanu:
@glebm please switch `CrawlTable` to `const std::vector<std::vector<DisplacementOf<int8_t>>>` once https://github.com/diasurgical/devilutionX/pull/4824 gets merged.


ephphatha:
Have a look at [FindClosestValidPosition](https://github.com/diasurgical/devilutionX/blob/master/Source/engine/path.cpp#L436), that function aims to act the same as a loop over `CrawlTable`. I never got around to replacing every use of a loop over the lookup table but you can probably use it here.

vlolteanu:
@ephphatha I doubt the end result will be equivalent. CrawlTable tries to describe concentric rings, but does so imprecisely. I tried replacing it with functions describing the coordinates, but couldn't find the precise pattern. Some of the deviations seemed arbitrary.

ephphatha:
It's meant to be exactly equivalent. CrawlTable has a search space that looks like:

```
1

-2-
3-4
-1-

-526-
D9-AE
F---G
B7-8C
-314-
```

reorganising the table of offsets like the following makes the pattern clearer:
```
             {  0,  2 },              {  0,  -2 }, // cells vertically aligned with the start position, +0/-0 deduplicated
{  -1,  2 }, {  1,  2 }, { -1,  -2 }, {  1,  -2 }, // walking left/right
{  -1,  1 }, {  1,  1 }, { -1,  -1 }, {  1,  -1 }, // corner of the search space is inset one tile for visual effect
{  -2,  1 }, {  2,  1 }, { -2,  -1 }, {  2,  -1 }, // walking up/down
{  -2,  0 }, {  2,  0 },                           // cells horizontally aligned with the start position, +0/-0 deduplicated
```

Larger patterns have multiple "walking" rows, but the start/middle/end all look the same.

vlolteanu:
@ephphatha You're a life saver. *muah*

test/lighting_test.cpp
```diff
@@ -10,13 +10,12 @@ TEST(Lighting, CrawlTables)
 	bool added[40][40];
 	memset(added, 0, sizeof(added));
 
-	for (int j = 0; j < 19; j++) {
+	for (int j = 0; j < (int)CrawlTable.size(); j++) {
```
AJenbo, line 13:
```suggestion
	for (size_t j = 0; j < CrawlTable.size(); j++) {
```

AJenbo (approved):
Very nice :+1:

PR #4834, opened 2022-07-01
Refactor VisionCrawlTable & friends

vlolteanu:
EDIT:
There's some stuff to unpack for `DoVision`:

- It apparently never gets called for a position outside of the dungeon bounds.
- [x|y][1|2]adj act as displacements, but 1 and 2 are interchangeable, as they receive the same treatment
- If the displacement from the `VisionCrawlTable` has one coordinate equal to 0, all **adj variables are 0, and the last innermost `if` ends up looking at nCrawl[x|y]

### DIFF
--- a/Source/engine/displacement.hpp
+++ b/Source/engine/displacement.hpp
@@ -95,6 +95,14 @@ struct DisplacementOf {
 		return *this;
 	}
 
+	template <typename DeltaU>
+	constexpr DisplacementOf<DeltaT> &operator*=(const DisplacementOf<DeltaU> factor)
+	{
+		deltaX = static_cast<DeltaT>(deltaX * factor.deltaX);
+		deltaY = static_cast<DeltaT>(deltaY * factor.deltaY);
+		return *this;
+	}
+
 	constexpr DisplacementOf<DeltaT> &operator/=(const int factor)
 	{
 		deltaX /= factor;
@@ -279,6 +287,13 @@ constexpr DisplacementOf<DisplacementDeltaT> operator*(DisplacementOf<Displaceme
 
 template <typename DisplacementDeltaT>
 constexpr DisplacementOf<DisplacementDeltaT> operator*(DisplacementOf<DisplacementDeltaT> a, const float factor)
+{
+	a *= factor;
+	return a;
+}
+
+template <typename DisplacementDeltaT, typename DisplacementDeltaU>
+constexpr DisplacementOf<DisplacementDeltaT> operator*(DisplacementOf<DisplacementDeltaT> a, const DisplacementOf<DisplacementDeltaU> factor)
 {
 	a *= factor;
 	return a;
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -30,31 +30,31 @@ namespace {
  * X- Y-coordinate offsets of lighting visions.
  * The last entry-pair is only for alignment.
  */
-const uint8_t VisionCrawlTable[23][30] = {
+const DisplacementOf<int8_t> VisionCrawlTable[23][15] = {
 	// clang-format off
-	{ 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8, 0, 9, 0, 10,  0, 11,  0, 12,  0, 13,  0, 14,  0, 15,  0 },
-	{ 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8, 1, 9, 1, 10,  1, 11,  1, 12,  1, 13,  1, 14,  1, 15,  1 },
-	{ 1, 0, 2, 0, 3, 0, 4, 1, 5, 1, 6, 1, 7, 1, 8, 1, 9, 1, 10,  1, 11,  1, 12,  2, 13,  2, 14,  2, 15,  2 },
-	{ 1, 0, 2, 0, 3, 1, 4, 1, 5, 1, 6, 1, 7, 1, 8, 2, 9, 2, 10,  2, 11,  2, 12,  2, 13,  3, 14,  3, 15,  3 },
-	{ 1, 0, 2, 1, 3, 1, 4, 1, 5, 1, 6, 2, 7, 2, 8, 2, 9, 3, 10,  3, 11,  3, 12,  3, 13,  4, 14,  4,  0,  0 },
-	{ 1, 0, 2, 1, 3, 1, 4, 1, 5, 2, 6, 2, 7, 3, 8, 3, 9, 3, 10,  4, 11,  4, 12,  4, 13,  5, 14,  5,  0,  0 },
-	{ 1, 0, 2, 1, 3, 1, 4, 2, 5, 2, 6, 3, 7, 3, 8, 3, 9, 4, 10,  4, 11,  5, 12,  5, 13,  6, 14,  6,  0,  0 },
-	{ 1, 1, 2, 1, 3, 2, 4, 2, 5, 3, 6, 3, 7, 4, 8, 4, 9, 5, 10,  5, 11,  6, 12,  6, 13,  7,  0,  0,  0,  0 },
-	{ 1, 1, 2, 1, 3, 2, 4, 2, 5, 3, 6, 4, 7, 4, 8, 5, 9, 6, 10,  6, 11,  7, 12,  7, 12,  8, 13,  8,  0,  0 },
-	{ 1, 1, 2, 2, 3, 2, 4, 3, 5, 4, 6, 5, 7, 5, 8, 6, 9, 7, 10,  7, 10,  8, 11,  8, 12,  9,  0,  0,  0,  0 },
-	{ 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 5, 7, 6, 8, 7, 9, 8, 10,  9, 11,  9, 11, 10,  0,  0,  0,  0,  0,  0 },
-	{ 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 10, 11, 11,  0,  0,  0,  0,  0,  0,  0,  0 },
-	{ 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 5, 6, 6, 7, 7, 8, 8, 9,  9, 10,  9, 11, 10, 11,  0,  0,  0,  0,  0,  0 },
-	{ 1, 1, 2, 2, 2, 3, 3, 4, 4, 5, 5, 6, 5, 7, 6, 8, 7, 9,  7, 10,  8, 10,  8, 11,  9, 12,  0,  0,  0,  0 },
-	{ 1, 1, 1, 2, 2, 3, 2, 4, 3, 5, 4, 6, 4, 7, 5, 8, 6, 9,  6, 10,  7, 11,  7, 12,  8, 12,  8, 13,  0,  0 },
-	{ 1, 1, 1, 2, 2, 3, 2, 4, 3, 5, 3, 6, 4, 7, 4, 8, 5, 9,  5, 10,  6, 11,  6, 12,  7, 13,  0,  0,  0,  0 },
-	{ 0, 1, 1, 2, 1, 3, 2, 4, 2, 5, 3, 6, 3, 7, 3, 8, 4, 9,  4, 10,  5, 11,  5, 12,  6, 13,  6, 14,  0,  0 },
-	{ 0, 1, 1, 2, 1, 3, 1, 4, 2, 5, 2, 6, 3, 7, 3, 8, 3, 9,  4, 10,  4, 11,  4, 12,  5, 13,  5, 14,  0,  0 },
-	{ 0, 1, 1, 2, 1, 3, 1, 4, 1, 5, 2, 6, 2, 7, 2, 8, 3, 9,  3, 10,  3, 11,  3, 12,  4, 13,  4, 14,  0,  0 },
-	{ 0, 1, 0, 2, 1, 3, 1, 4, 1, 5, 1, 6, 1, 7, 2, 8, 2, 9,  2, 10,  2, 11,  2, 12,  3, 13,  3, 14,  3, 15 },
-	{ 0, 1, 0, 2, 0, 3, 1, 4, 1, 5, 1, 6, 1, 7, 1, 8, 1, 9,  1, 10,  1, 11,  2, 12,  2, 13,  2, 14,  2, 15 },
-	{ 0, 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 1, 8, 1, 9,  1, 10,  1, 11,  1, 12,  1, 13,  1, 14,  1, 15 },
-	{ 0, 1, 0, 2, 0, 3, 0, 4, 0, 5, 0, 6, 0, 7, 0, 8, 0, 9,  0, 10,  0, 11,  0, 12,  0, 13,  0, 14,  0, 15 },
+	{ { 1, 0 }, { 2, 0 }, { 3, 0 }, { 4, 0 }, { 5, 0 }, { 6, 0 }, { 7, 0 }, { 8, 0 }, { 9, 0 }, { 10,  0 }, { 11,  0 }, { 12,  0 }, { 13,  0 }, { 14,  0 }, { 15,  0 } },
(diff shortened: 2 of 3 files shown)


vlolteanu:
@ephphatha Don't tell me you've also figured out how `VisionCrawlTable` is generated :stuck_out_tongue_closed_eyes:.

Source/engine/displacement.hpp
```diff
@@ -116,6 +123,12 @@ struct Displacement {
 		return a;
 	}
 
+	constexpr friend Displacement operator*(Displacement a, const Displacement factor)
```
glebm, line 126: Hopefully https://github.com/diasurgical/devilutionX/pull/4824 will get merged first as this is going to cause conflicts there

ephphatha:
> @ephphatha Don't tell me you've also figured out how `VisionCrawlTable` is generated 😝.

Haven't looked at it before :P

That said, it looks like "rays" of 15 units in length, so pre-calculating the tiles touched accounting for diagonal movement. I seem to recall some discussion about this in a previous PR but I can't find it now.

Source/missiles.cpp
```diff
@@ -3074,13 +3073,12 @@ void MI_FireNova(Missile &missile)
...
-				AddMissile(src, src + offset, dir, MIS_FIRENOVA, en, id, dam, missile._mispllvl);
-			sx1 = k[6];
-			sy1 = k[7];
-		}
+		if (k[3] == s1)
+			continue;
+		Displacement offsets[] = { k[3], k[3].flipXY(), k[3].flipX(), k[3].flipY() };
+		for (Displacement offset : offsets)
+			AddMissile(src, src + offset, dir, MIS_FIRENOVA, en, id, dam, missile._mispllvl);
+		s1 = k[3];
```
ephphatha, line 3081: These *Nova loops are incredibly wasteful... they're going from `4, 0` to `4, 4` to `0, 4`. Can do that in 9 iterations instead of 23 and drop the `s1` temporary. But don't worry about it in this PR, that's for a future cleanup.

Source/lighting.cpp
```diff
@@ -283,90 +290,47 @@ void DoUnVision(Point position, int nRadius)
...
-					}
-					break;
-				case 2:
-					nCrawlX = position.x + VisionCrawlTable[j][k];
-					nCrawlY = position.y - VisionCrawlTable[j][k + 1];
-					if (VisionCrawlTable[j][k] > 0 && VisionCrawlTable[j][k + 1] > 0) {
-						x1adj = -1;
-						y2adj = 1;
-					}
+			int lineLen = radius - RadiusAdj[j];
```
ephphatha, line 310: You can get rid of RadiusAdj if you make VisionCrawlTable an array of vectors. The last entries of rays 8 and 14 (`13, 8` and `8, 13` respectively) don't get used, so can drop those and all the `0, 0` entries to match vanilla behaviour. The two innermost loops can then be range-based loops like
```cpp
for (auto ray : VisionCrawlTable)
    for (Displacement offset : ray)
```

Source/missiles.cpp
ephphatha: opened a pr for this

Source/lighting.cpp
```diff
@@ -93,31 +93,31 @@ const std::vector<std::vector<Displacement>> CrawlTable = [] {
  * X- Y-coordinate offsets of lighting visions.
  * The last entry-pair is only for alignment.
  */
-const uint8_t VisionCrawlTable[23][30] = {
+const Displacement VisionCrawlTable[23][15] = {
```
glebm, line 96: `DisplacementOf<int8_t>` once #4824 is merged

Source/lighting.cpp
ephphatha: Oh, looking at this again this is somewhat broken. the vision table is built using rays with magnitude 15, but when used with smaller radii the adjustment should be scaled to match. This becomes more noticeable when radius <= 8...

image
This table shows the magnitude of each vector between the origin and the tile rounded to the nearest integer. The cells with a red background have a magnitude > 15, if the intent was always to round down then they shouldn't be included. If the intent was to fudge the boundary for a more pleasing shape, 11,11 should probably be excluded.

For an example of the scaling impacting what area gets lit: using a radius of 4 ends up completely skipping the diagonal ray, though there's enough duplication in the other rows of the table that doesn't matter. There are still gaps because of the adjustments applied to other rows, the two highlighted cells are never checked when they should be included:
image

The "correct" adjustment would be something along the lines of
```cpp
size_t effectiveRadius = radius * ray.size() / 15; // using truncation towards zero
auto end = ray.begin() + effectiveRadius;
```

ephphatha: oooor, how about making the lookup table have 15 entries for each ray. I'll put a comment on the table with a suggestion of how that'd look

Source/lighting.cpp
```diff
@@ -93,31 +93,31 @@ const std::vector<std::vector<Displacement>> CrawlTable = [] {
...
+	{ { 0, 1 }, { 1, 2 }, { 1, 3 }, { 2, 4 }, { 2, 5 }, { 3, 6 }, { 3, 7 }, { 3, 8 }, { 4, 9 }, {  4, 10 }, {  5, 11 }, {  5, 12 }, {  6, 13 }, {  6, 14 }, {  0,  0 } },
+	{ { 0, 1 }, { 1, 2 }, { 1, 3 }, { 1, 4 }, { 2, 5 }, { 2, 6 }, { 3, 7 }, { 3, 8 }, { 3, 9 }, {  4, 10 }, {  4, 11 }, {  4, 12 }, {  5, 13 }, {  5, 14 }, {  0,  0 } },
+	{ { 0, 1 }, { 1, 2 }, { 1, 3 }, { 1, 4 }, { 1, 5 }, { 2, 6 }, { 2, 7 }, { 2, 8 }, { 3, 9 }, {  3, 10 }, {  3, 11 }, {  3, 12 }, {  4, 13 }, {  4, 14 }, {  0,  0 } },
+	{ { 0, 1 }, { 0, 2 }, { 1, 3 }, { 1, 4 }, { 1, 5 }, { 1, 6 }, { 1, 7 }, { 2, 8 }, { 2, 9 }, {  2, 10 }, {  2, 11 }, {  2, 12 }, {  3, 13 }, {  3, 14 }, {  3, 15 } },
+	{ { 0, 1 }, { 0, 2 }, { 0, 3 }, { 1, 4 }, { 1, 5 }, { 1, 6 }, { 1, 7 }, { 1, 8 }, { 1, 9 }, {  1, 10 }, {  1, 11 }, {  2, 12 }, {  2, 13 }, {  2, 14 }, {  2, 15 } },
+	{ { 0, 1 }, { 0, 2 }, { 0, 3 }, { 0, 4 }, { 0, 5 }, { 0, 6 }, { 0, 7 }, { 1, 8 }, { 1, 9 }, {  1, 10 }, {  1, 11 }, {  1, 12 }, {  1, 13 }, {  1, 14 }, {  1, 15 } },
+	{ { 0, 1 }, { 0, 2 }, { 0, 3 }, { 0, 4 }, { 0, 5 }, { 0, 6 }, { 0, 7 }, { 0, 8 }, { 0, 9 }, {  0, 10 }, {  0, 11 }, {  0, 12 }, {  0, 13 }, {  0, 14 }, {  0, 15 } },
```
ephphatha, line 120:
```suggestion
	{ { 1, 0 }, { 2, 0 }, { 3, 0 }, { 4, 0 }, { 5, 0 }, { 6, 0 }, { 7, 0 }, { 8, 0 }, { 9, 0 }, { 10,  0 }, { 11,  0 }, { 12,  0 }, { 13,  0 }, { 14,  0 }, { 15,  0 } },
	{ { 1, 0 }, { 2, 0 }, { 3, 0 }, { 4, 0 }, { 5, 0 }, { 6, 0 }, { 7, 0 }, { 8, 1 }, { 9, 1 }, { 10,  1 }, { 11,  1 }, { 12,  1 }, { 13,  1 }, { 14,  1 }, { 15,  1 } },
	{ { 1, 0 }, { 2, 0 }, { 3, 0 }, { 4, 1 }, { 5, 1 }, { 6, 1 }, { 7, 1 }, { 8, 1 }, { 9, 1 }, { 10,  1 }, { 11,  1 }, { 12,  2 }, { 13,  2 }, { 14,  2 }, { 15,  2 } },
	{ { 1, 0 }, { 2, 0 }, { 3, 1 }, { 4, 1 }, { 5, 1 }, { 6, 1 }, { 7, 1 }, { 8, 2 }, { 9, 2 }, { 10,  2 }, { 11,  2 }, { 12,  2 }, { 13,  3 }, { 14,  3 }, { 15,  3 } },
	{ { 1, 0 }, { 2, 1 }, { 3, 1 }, { 4, 1 }, { 5, 1 }, { 6, 2 }, { 7, 2 }, { 8, 2 }, { 9, 3 }, { 10,  3 }, { 11,  3 }, { 12,  3 }, { 13,  4 }, { 13,  4 }, { 14,  4 } },// {  0,  0 } },
	{ { 1, 0 }, { 2, 1 }, { 3, 1 }, { 4, 1 }, { 5, 2 }, { 6, 2 }, { 7, 3 }, { 8, 3 }, { 9, 3 }, { 10,  4 }, { 11,  4 }, { 12,  4 }, { 12,  4 }, { 13,  5 }, { 14,  5 } },// {  0,  0 } },
	{ { 1, 0 }, { 2, 1 }, { 3, 1 }, { 4, 2 }, { 5, 2 }, { 6, 3 }, { 7, 3 }, { 8, 3 }, { 9, 4 }, { 10,  4 }, { 10,  4 }, { 11,  5 }, { 12,  5 }, { 13,  6 }, { 14,  6 } },// {  0,  0 } },
	{ { 1, 1 }, { 2, 1 }, { 3, 2 }, { 4, 2 }, { 5, 3 }, { 6, 3 }, { 6, 3 }, { 7, 4 }, { 8, 4 }, {  9,  5 }, { 10,  5 }, { 11,  6 }, { 11,  6 }, { 12,  6 }, { 13,  7 } },// {  0,  0 }, {  0,  0 } },
	{ { 1, 1 }, { 2, 1 }, { 3, 2 }, { 4, 2 }, { 5, 3 }, { 5, 3 }, { 6, 4 }, { 7, 4 }, { 8, 5 }, {  9,  6 }, { 10,  6 }, { 10,  6 }, { 11,  7 }, { 12,  7 }, { 12,  8 } },// { 13,  8 }, {  0,  0 } },
	{ { 1, 1 }, { 2, 2 }, { 3, 2 }, { 3, 2 }, { 4, 3 }, { 5, 4 }, { 6, 5 }, { 7, 5 }, { 7, 5 }, {  8,  6 }, {  9,  7 }, { 10,  7 }, { 10,  8 }, { 11,  8 }, { 12,  9 } },// {  0,  0 }, {  0,  0 } },
	{ { 1, 1 }, { 2, 2 }, { 2, 2 }, { 3, 3 }, { 4, 4 }, { 4, 4 }, { 5, 5 }, { 6, 5 }, { 7, 6 }, {  8,  7 }, {  8,  7 }, {  9,  8 }, { 10,  9 }, { 11,  9 }, { 11, 10 } },// {  0,  0 }, {  0,  0 }, {  0,  0 } },
	{ { 1, 1 }, { 2, 2 }, { 2, 2 }, { 3, 3 }, { 4, 4 }, { 4, 4 }, { 5, 5 }, { 6, 6 }, { 7, 7 }, {  7,  7 }, {  8,  8 }, {  9,  9 }, {  9,  9 }, { 10, 10 }, { 11, 11 } },// {  0,  0 }, {  0,  0 }, {  0,  0 }, {  0,  0 } },
	{ { 1, 1 }, { 2, 2 }, { 2, 2 }, { 3, 3 }, { 4, 4 }, { 4, 4 }, { 5, 5 }, { 5, 6 }, { 6, 7 }, {  7,  8 }, {  7,  8 }, {  8,  9 }, {  9, 10 }, {  9, 11 }, { 10, 11 } },// {  0,  0 }, {  0,  0 }, {  0,  0 } },
	{ { 1, 1 }, { 2, 2 }, { 2, 3 }, { 3, 2 }, { 3, 4 }, { 4, 5 }, { 5, 6 }, { 5, 7 }, { 5, 7 }, {  6,  8 }, {  7,  9 }, {  7, 10 }, {  8, 10 }, {  8, 11 }, {  9, 12 } },// {  0,  0 }, {  0,  0 } },
	{ { 1, 1 }, { 1, 2 }, { 2, 3 }, { 2, 4 }, { 3, 5 }, { 3, 5 }, { 4, 6 }, { 4, 7 }, { 5, 8 }, {  6,  9 }, {  6, 10 }, {  6, 10 }, {  7, 11 }, {  7, 12 }, {  8, 12 } },// {  8, 13 }, {  0,  0 } },
	{ { 1, 1 }, { 1, 2 }, { 2, 3 }, { 2, 4 }, { 3, 5 }, { 3, 6 }, { 3, 6 }, { 4, 7 }, { 4, 8 }, {  5,  9 }, {  5, 10 }, {  6, 11 }, {  6, 11 }, {  6, 12 }, {  7, 13 } },// {  0,  0 }, {  0,  0 } },
	{ { 0, 1 }, { 1, 2 }, { 1, 3 }, { 2, 4 }, { 2, 5 }, { 3, 6 }, { 3, 7 }, { 3, 8 }, { 4, 9 }, {  4, 10 }, {  4, 10 }, {  5, 11 }, {  5, 12 }, {  6, 13 }, {  6, 14 } },// {  0,  0 } },
	{ { 0, 1 }, { 1, 2 }, { 1, 3 }, { 1, 4 }, { 2, 5 }, { 2, 6 }, { 3, 7 }, { 3, 8 }, { 3, 9 }, {  4, 10 }, {  4, 11 }, {  4, 12 }, {  4, 12 }, {  5, 13 }, {  5, 14 } },// {  0,  0 } },
	{ { 0, 1 }, { 1, 2 }, { 1, 3 }, { 1, 4 }, { 1, 5 }, { 2, 6 }, { 2, 7 }, { 2, 8 }, { 3, 9 }, {  3, 10 }, {  3, 11 }, {  3, 12 }, {  4, 13 }, {  4, 13 }, {  4, 14 } },// {  0,  0 } },
	{ { 0, 1 }, { 0, 2 }, { 1, 3 }, { 1, 4 }, { 1, 5 }, { 1, 6 }, { 1, 7 }, { 2, 8 }, { 2, 9 }, {  2, 10 }, {  2, 11 }, {  2, 12 }, {  3, 13 }, {  3, 14 }, {  3, 15 } },
	{ { 0, 1 }, { 0, 2 }, { 0, 3 }, { 1, 4 }, { 1, 5 }, { 1, 6 }, { 1, 7 }, { 1, 8 }, { 1, 9 }, {  1, 10 }, {  1, 11 }, {  2, 12 }, {  2, 13 }, {  2, 14 }, {  2, 15 } },
	{ { 0, 1 }, { 0, 2 }, { 0, 3 }, { 0, 4 }, { 0, 5 }, { 0, 6 }, { 0, 7 }, { 1, 8 }, { 1, 9 }, {  1, 10 }, {  1, 11 }, {  1, 12 }, {  1, 13 }, {  1, 14 }, {  1, 15 } },
	{ { 0, 1 }, { 0, 2 }, { 0, 3 }, { 0, 4 }, { 0, 5 }, { 0, 6 }, { 0, 7 }, { 0, 8 }, { 0, 9 }, {  0, 10 }, {  0, 11 }, {  0, 12 }, {  0, 13 }, {  0, 14 }, {  0, 15 } },
```

This inserts duplicate tiles along the original ray... casts? paths? with the aim of ensuring for each row/ray of VisionCrawlTable, that the displacement of `ray[n]` has magnitude approximately equal to `n`. Loops over a given radius can use
```cpp
for (const auto &ray : VisionCrawlTable)
    for (size_t i = 0; i < radius; i++)
        // do something with ray[i]
```
and don't need to use RadiusAdj.

Please note this doesn't exactly match existing behaviour for smaller radiuses, inserting duplicates at other positions could give better behaviour. This was just a quick and dirty approach.

vlolteanu:
Let's merge this as it is and figure out a cleaner solution later.

@ephphatha Why did you duplicate parts of the LightningCrawlTable in missiles.cpp?

AJenbo:
I have a lot more faith in merging this now that it's rebased on the timedemo test and passed :+1:

ephphatha:
> @ephphatha Why did you duplicate parts of the LightningCrawlTable in missiles.cpp?

For guardians it cuts out 90 duplicate checks (138 > 48 iterations) due to overlapping rays. Nova went from 23 to 9 iterations, however that code already had a check to skip duplicate tiles so it's not a major difference, that was mainly to make it clearer what it's trying to do.

AJenbo (approved):

PR #4835, opened 2022-07-01
DiabloUi: Render focus indicators as PCX

glebm:
They're slightly larger as PCX than CEL but are tiny overall.

This was the last use of `LoadPcxAssetAsCel`.

PCX->CEL conversion code is also removed in this PR.


AJenbo (approved):
As long as Render*Sprite(out, sprite, rect); exists this should all be compatible with future renderes :)

PR #4837, opened 2022-07-01
Replace most uses of `sprintf` with `fmt`

glebm:


AJenbo (approved):
Nice, and it reduces the site of monster data :)

PR #4839, opened 2022-07-01
Clarification of C++ standard usage

aetn23:
Add clarification in CMakelists.txt regarding C++ standard, as well as add new section in CONTRIBUTING.md.

This should help to clear the confusion about which standard is being used.

The reasons behind using C++17 while setting standard to C++20 are probably not described correctly - they have been taken from Discord discussion. As such, it would be great if somebody knowing the ins and outs of standard situation could clarify this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,6 +240,9 @@ if(GPERF)
   endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


qndel:
typo in commit name - should be standard instead of standared

AJenbo (approved):
LGTM :+1:

PR #4843, opened 2022-07-01
Avoid unnecessary iterations in missile routines

ephphatha:
Since I was looking at it anyway for #4834 :D

PR #4844, opened 2022-07-01
Pull distinct points from VisionCrawlTable used by guardians

ephphatha:
Opening this to highlight the odd targeting Guardians use more than anything. If we're happy to change behaviour this can be made way simpler by using `FindClosestValidPosition` instead of having guardians search for targets from far to near. Based off #4843 since it was the last remaining use of VisionCrawlTable outside of lighting.cpp

### DIFF
--- a/Source/lighting.h
+++ b/Source/lighting.h
@@ -131,6 +131,4 @@ auto Crawl(unsigned minRadius, unsigned maxRadius, F function) -> invoke_result_
 
 /* rdata */
 
-extern const uint8_t VisionCrawlTable[23][30];
-
 } // namespace devilution
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -3104,11 +3104,11 @@ void MI_NovaCommon(Missile &missile, missile_id projectileType)
 		en = TARGET_MONSTERS;
 	}
 
-	constexpr std::array<Displacement, 9> quarterRadius = { { { 4, 0 }, { 4, 1 }, { 4, 2 }, { 4, 3 }, { 4, 4 }, { 3, 4 }, { 2, 4 }, { 1, 4 }, { 0, 4 } } };
-	for (Displacement quarterOffset : quarterRadius) {
+	constexpr std::array<WorldTileDisplacement, 9> quarterRadius = { { { 4, 0 }, { 4, 1 }, { 4, 2 }, { 4, 3 }, { 4, 4 }, { 3, 4 }, { 2, 4 }, { 1, 4 }, { 0, 4 } } };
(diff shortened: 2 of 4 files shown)


Source/missiles.cpp
```diff
@@ -3294,26 +3294,38 @@ void MI_Guardian(Missile &missile)
...
-				found = GuardianTryFireAt(missile, { position.x + offset.deltaX, position.y + offset.deltaY })
-				    || GuardianTryFireAt(missile, { position.x - offset.deltaX, position.y - offset.deltaY })
-				    || GuardianTryFireAt(missile, { position.x + offset.deltaX, position.y - offset.deltaY })
-				    || GuardianTryFireAt(missile, { position.x - offset.deltaX, position.y + offset.deltaY });
-				if (!found) {
-					previous = offset;
-				}
+		// Guardians pick a target by working backwards along lines originally based on VisionCrawlTable.
+		// Because of their rather unique behaviour the points checked have been unrolled here
+		constexpr std::array<Displacement, 48> guardianArc {
```
glebm, line 3299: `DisplacementOf<int8_t>`?

ephphatha: I accidently changed the nova loops before realising what PR this comment was on -_-. Included both anyway

Source/lighting.h
```diff
@@ -131,6 +131,4 @@ auto Crawl(unsigned minRadius, unsigned maxRadius, F function) -> invoke_result_
 
 /* rdata */
 
-extern const uint8_t VisionCrawlTable[23][30];
```
AJenbo, line 134: VisionCrawlTable should also be wrapped by the anonymous name space :)

AJenbo:
Ok after see what pattern this uses I have nothing good to say about the existing way this is done...

Basically you have 4 fans rotating in a slightly erratic maner, with the fan blades being drawn from outside to in.

I have attached the primary quadrant for you viewing displeasure:
ezgif-3-e2683d7ece

ephphatha:
Yeah we had a bit of discussion about the use of VisionCrawlTable in https://github.com/diasurgical/devilutionX/pull/4834

When used in the context of determining sight/vision/light it was beneficial to check duplicate tiles as a nearby column could block 3 or 4 different rays quite easily, but for the way guardians use the lookup table the duplicates just wasted cycles.

AJenbo:
It also feals strange that the guardian prioritizes monsters in a NW/SE direction. I don't mind so much that it's fare to near, but that doesn't feel particularly meaning full either.

AJenbo:
> If we're happy to change behavior this can be made way simpler by using `FindClosestValidPosition`

I think that might actually be a good idea.

PR #4845, opened 2022-07-01
Migrate snprintf to fmt

glebm:
Follow-up to #4837

### DIFF
--- a/Source/appfat.cpp
+++ b/Source/appfat.cpp
@@ -25,20 +25,6 @@ bool Terminating = false;
 SDL_threadID CleanupThreadId;
 
 /**
- * @brief Displays an error message box based on the given format string and variable argument list.
- * @param pszFmt Error message format
- * @param va Additional parameters for message format
- */
-void MsgBox(const char *pszFmt, va_list va)
-{
-	char text[256];
-
-	vsnprintf(text, sizeof(text), pszFmt, va);
-
-	UiErrorOkDialog(_("Error"), text);
-}
-
-/**
  * @brief Cleans up after a fatal application error.
  */
 void FreeDlg()
@@ -66,25 +52,10 @@ void app_fatal(string_view str)
 	diablo_quit(1);
 }
 
-void app_fatal(const char *pszFmt, ...)
-{
-	va_list va;
-
-	va_start(va, pszFmt);
-	FreeDlg();
-
-	if (pszFmt != nullptr)
-		MsgBox(pszFmt, va);
-
-	va_end(va);
-
-	diablo_quit(1);
-}
-
 #ifdef _DEBUG
 void assert_fail(int nLineNo, const char *pszFile, const char *pszFail)
 {
-	app_fatal("assertion failed (%s:%i)\n%s", pszFile, nLineNo, pszFail);
+	app_fatal(fmt::format("assertion failed ({}:{})\n{}", pszFile, nLineNo, pszFail));
 }
 #endif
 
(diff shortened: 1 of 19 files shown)


Source/appfat.cpp
```diff
@@ -66,25 +52,10 @@ void app_fatal(string_view str)
 	diablo_quit(1);
 }
 
-void app_fatal(const char *pszFmt, ...)
-{
-	va_list va;
```
AJenbo, line 71: God riddens

AJenbo (dismissed):
Better merge this before we get conflicts :)


The following tests failed:
```
[ RUN      ] Appfat.ErrDlg
test/appfat_test.cpp:15: Failure
Death test: ErrDlg("Title", "Unknown error", "appfat.cpp", 7)
    Result: died but not with expected exit code:
            Terminated by signal 11 (core dumped)
```

```
[ RUN      ] Appfat.InsertCDDlg
test/appfat_test.cpp:20: Failure
Death test: InsertCDDlg("diabdat.mpq")
    Result: died but not with expected exit code:
            Terminated by signal 11 (core dumped)
```
 
```
[ RUN      ] Appfat.DirErrorDlg
test/appfat_test.cpp:25: Failure
Death test: DirErrorDlg("/")
    Result: died but not with expected exit code:
            Terminated by signal 11 (core dumped)
```

glebm:
Fixed

Source/engine/render/scrollrt.cpp
```diff
@@ -1333,7 +1336,7 @@ void DrawFPS(const Surface &out)
 		framerate = 1000 * frameend / frames;
 		frameend = 0;
 	}
-	snprintf(string, 12, "%i FPS", framerate);
+	*fmt::format_to_n(string, 12, FMT_COMPILE("{} FPS"), framerate).out = '\0';
```
ephphatha, line 1339: While it's unlikely anyone is ever going to render a billion frames a second, this could potentially write past the end of the array? How about removing the `string` temporary and inlining the `fmt::format` call or using std::string instead of char[]?

glebm: I'd like to minimize the per-frame allocations. Perhaps we can simply increase the buffer to the maximum number of digits.

Source/init.cpp
```diff
@@ -113,7 +113,7 @@ std::vector<std::string> GetMPQSearchPaths()
 		std::string message;
 		for (std::size_t i = 0; i < paths.size(); ++i) {
 			char prefix[32];
-			std::snprintf(prefix, sizeof(prefix), "\n%6u. '", static_cast<unsigned>(i + 1));
+			*fmt::format_to_n(prefix, sizeof(prefix) - 1, "\n{:6d}. '", static_cast<unsigned>(i + 1)).out = '\0';
 			message.append(prefix);
 			message.append(paths[i]);
 			message += '\'';
```
ephphatha, line 119: This should be able to be simplified to

```cpp
message.append(fmt::format("{:6d} '{}'", i+1, paths[i]));
```

Source/engine/render/scrollrt.cpp
glebm: Done

Source/init.cpp
glebm: That would allocate a string unnecessarily. We know the buffer is large enough, so replace `format_to_n` with `format_to`, making it easier to read.

ephphatha: That code is only ever run if verbose logging is enabled, and only 2 or 3 times when the user starts the game or once when they change the language in the settings menu to something other than english. I was hopeful it'd be worth the tradeoff to make it simpler to read and not call `string::append` so much.

glebm: Fair enough, I tend to write everything in the same efficient style so that I don't have to think too much about how often something is called.

If we had `absl::StrAppend`, this whole thing (and a number of other places) couldve been a single call, perhaps we could consider implementing something similar to it.

glebm: Done

ephphatha (approved):
Demo filename concatenation could possibly use inline calls to fmt::format instead of a buffer but that's pretty minor.

Also SetApplicationVersions() using two different ways of setting a c-string :P

PR #4846, opened 2022-07-01
Update Croatian translation

muzena:


AJenbo:

PR #4850, opened 2022-07-01
Correct members' names in monster.h

aetn23:
Correct names in AnimStruct and CMonster by fixing inconsistent use of uppercase. Remove m prefix from  members' names.

This is done to improve readability and bring consistency to naming scheme.


Source/dead.cpp
```diff
@@ -32,15 +32,15 @@ void InitCorpses()
...
+		if (mtypes[LevelMonsterTypes[i].type] != 0)
 			continue;
 
 		InitDeadAnimationFromMonster(Corpses[nd], LevelMonsterTypes[i]);
 		Corpses[nd].translationPaletteIndex = 0;
 		nd++;
 
-		LevelMonsterTypes[i].mdeadval = nd;
-		mtypes[LevelMonsterTypes[i].mtype] = nd;
+		LevelMonsterTypes[i].deadval = nd;
```
AJenbo, line 42: Could we find a better name then `deadvel`, maybe `corpseId`?

Source/monster.h
```diff
@@ -132,36 +132,37 @@ enum class LeaderRelation : uint8_t {
...
-	const AnimStruct &GetAnimData(MonsterGraphic graphic) const
+	const AnimStruct &getAnimData(MonsterGraphic graphic) const
 	{
-		return Anims[static_cast<int>(graphic)];
+		return anims[static_cast<int>(graphic)];
 	}
-	std::unique_ptr<TSnd> Snds[4][2];
-	int8_t mdeadval;
-	const MonsterData *MData;
+	std::unique_ptr<TSnd> snds[4][2];
```
AJenbo, line 163: `sounds` would be nicer and isn't hat long

Source/monster.h
```diff
@@ -25,8 +25,8 @@ namespace devilution {
 
 struct Missile;
 
-#define MAXMONSTERS 200
-#define MAX_LVLMTYPES 24
+constinit const int MaxMonsters = 200;
```
glebm: `constexpr`, because not all of our targets support `constinit`. Also, no need for `const`, it is implied by `constinit`/`constexpr`.

Source/dead.cpp
AJenbo: Well that was fast :eyes:

Source/monster.h
AJenbo: This isn't actually part of this PR and only there because @aetn23 merged rather then create a new branch/rebase after his previous PR got merged.

Source/monster.h
aetn23: Renamed.

Source/monster.h
aetn23: Yes, this was resolved a commit or two later and is already merged to master.

AJenbo:
Thanks, this is stuff I have been wanting to get done for ages but never had time to prioritize.

Going to squash merge since the history is containing both commits from your previous PR and the new ones. Would be good if you can either reset/rebase/re-branch when starting work on your next PR to make it easier to review.

aetn23:
@AJenbo  Sure, had no idea it's tougher to review when constantly making PRs from the same branch. Next monsters cleanups are coming in the near future.

glebm:
@aetn23 Please wait for #4824 to get merged before starting on renaming the `Monster` fields, to avoid having to deal with conflicts

PR #4852, opened 2022-07-02
Avoid casting unnecessarily in worldToNormalScreen

ephphatha:
This doesn't trigger warnings on msvc, not sure about other compilers. Also added asserts to all functions performing rotations to match Rotate()


AJenbo (approved):

AJenbo:
Didn't notice any GCC warnings in the CI log

PR #4853, opened 2022-07-02
Address int-arithmetic-overflow in loadsave.cpp

ephphatha:
clang-tidy config causes the linter to give a very helpful warning that `4*8` might overflow the range of `int` before being assigned to `size_t` 🙃 

we've got the template form of skip anyway so might as well use it.


AJenbo:
Nice cleanup :)

AJenbo (approved):

PR #4854, opened 2022-07-02
Only call fmt::format_to when the FPS value updates

ephphatha:
Don't even need to initialise the buffer since the last update time is set when DiabloInit is called, then as long as the loading screen takes more than a second the threshold gets hit to call format_to on the first frame of gameplay.

### DIFF
--- a/Source/engine/render/scrollrt.cpp
+++ b/Source/engine/render/scrollrt.cpp
@@ -1321,7 +1321,7 @@ void DrawView(const Surface &out, Point startPosition)
 void DrawFPS(const Surface &out)
 {
 	static int framesSinceLastUpdate = 0;
-	static fmt::memory_buffer buf {};
+	static string_view formatted {};
 
 	if (!frameflag || !gbActive) {
 		return;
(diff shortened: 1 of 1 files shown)


Source/engine/render/scrollrt.cpp
```diff
@@ -1322,23 +1320,24 @@ void DrawView(const Surface &out, Point startPosition)
  */
 void DrawFPS(const Surface &out)
 {
-	char buf[12];
+	static int framesSinceLastUpdate = 0;
+	static fmt::basic_memory_buffer<char, 12> buf {};
```
ephphatha, line 1324: I'm not sure if there's a way to specify a void allocator to prevent this from ever resizing? We can make this work with a static string_view and static char[] buffer instead of fmt::memory_buffer if that's desirable.

glebm: I think a static char buffer would be slightly simpler since it avoids all the capacity checks of `basic_memory_buffer`.

glebm (dismissed):
Thanks!

Source/engine/render/scrollrt.cpp
ephphatha: done :)

glebm (approved):

PR #4855, opened 2022-07-02
Turn CrawlTable into a single array

glebm:
Rather than an array of vectors, use a single `std::array` for the entire `CrawlTable`.

@vlolteanu

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -723,8 +723,8 @@ std::string DebugCmdSpawnUniqueMonster(const string_view parameter)
 
 	int spawnedMonster = 0;
 
-	for (const auto &table : CrawlTable) {
-		for (auto displacement : table) {
+	for (size_t row = 0; row < CrawlTableRows; ++row) {
+		for (DisplacementOf<int8_t> displacement : CrawlTableRow(row)) {
 			Point pos = myPlayer.position.tile + displacement;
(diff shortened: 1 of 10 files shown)


AJenbo:
C:\projects\devilutionx\Source\engine\displacement.hpp(8,10): fatal error C1083: Cannot open include file: 'fmt/core.h': No such file or directory [C:\projects\devilutionx\uwp-project\devilutionx.vcxproj]

vlolteanu:
@glebm I really don't see the point of this.

Maybe you're trying to get better cache locality, but CrawlTable is only seldomly looked at. Otherwise, this only complicates the code.

glebm:
It avoids heap allocations for something that we know the size of.

I find it to be cleaner than an array of vectors and prefer efficient code throughout.

There are no strong reasons for this otherwise -- better cache locality and less pointer indirection but this is used in a loop only in `MI_Chain`.

ephphatha:
CrawlTable is never used in a non-loop context. This basically takes it back to the original approach of a flat array chunked by lengths (stored in a separate array instead of embedded in the main list), but we really don't need the ability to arbitrarily index into the list (beyond specifying a min/max radius). I still think adapting the loop in FindClosestValidPosition to an iterator, or reworking the function to apply an action to `n` positions would be suitable, I don't think there's any need to actually store the table given the way it's used.

glebm:
Closing in favour of #4907

PR #4858, opened 2022-07-02
Add OptionalCelSprite: smaller than std::optional

glebm:
`CelSprite` data pointer can never be `nullptr`.

We implement a smaller `optional` for it by taking advantage of that.

Saves 8 bytes per optional `CelSprite`:

```c++
sizeof(OptionalCelSprite) == 16
sizeof(std::optional<OptionalCelSprite>) == 24
```

@ephphatha

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -85,7 +85,7 @@ Rectangle MainPanel;
 Rectangle LeftPanel;
 Rectangle RightPanel;
 std::optional<OwnedSurface> pBtmBuff;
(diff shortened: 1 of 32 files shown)


obligaron (approved):
lgtm, but can we remove some `#include "utils/stdcompat/optional.hpp"`?
For example `animationinfo.h` don't depend on `std::optional` anymore or?

glebm:
Done.

This optional is also cleared by assigning `std::nullopt`, so it's still needed in some places.

obligaron (approved):
lgtm 🙂

PR #4859, opened 2022-07-02
upload windows artifacts to second repo

qndel:
Will upload artifacts from every windows master build to https://github.com/artifacts-storage/devilutionx-artifacts
Should be pretty useful for bisecting in the future as builds take a pretty long time and after 1.4.0 a lot of them failed locally for me 
zip takes 9.45 mb, max github repo size is 10gb = artifacts from ~1k commits, let's worry about the limit when we get there :D

I've split exe and all other data - zip with exe takes 2.35mb, other data (DLLs, devilutionx.mpq, licenses) rarely changes so it shouldn't take much = we good for 4k commits? :D

### DIFF
--- a/.github/workflows/Windows_x64.yml
+++ b/.github/workflows/Windows_x64.yml
@@ -39,3 +39,35 @@ jobs:
       with:
         name: devilutionx_x64.zip
         path: build/devilutionx.zip
+        
+    - name: Move artifacts to new folder and split exe and other data into two folders
+      if: ${{ !env.ACT && github.ref == 'refs/heads/master' && github.event_name != 'pull_request' }}
+      working-directory: ${{github.workspace}}
+      shell: bash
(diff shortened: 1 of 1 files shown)


glebm:
Perhaps this should use `git lfs`?

glebm:
> If you use more than 1 GB of bandwidth per month without purchasing a data pack, Git LFS support is disabled on your account until the next month.

Maybe not

qndel:
Closingo while testing so I don't run build actions on devilutionx

qndel:
nvm forgot I shouldn't test on this branch anyway :D

AJenbo:
I was thinking it could be a separate shell script by probably not when it uses CI packages. Looks ok, is the key already there?

qndel:
> I was thinking it could be a separate shell script by probably not when it uses CI packages. Looks ok, is the key already there?

yup

PR #4861, opened 2022-07-02
Members' name changes and comments in Monster struct 

aetn23:
Change names of members in Monster: remove m or _m  prefix, correct camel case in most places. Members _mgoalvars as well as _mint were changed to more fitting names and provided with comments describing what they do. 

This is done to improve readability and provide more consistent naming scheme as well as document code better by giving more descriptive names and/or provide comments.


Source/dead.cpp
```diff
@@ -78,12 +78,12 @@ void SyncUniqDead()
...
 		auto &monster = Monsters[ActiveMonsters[i]];
-		if (monster._uniqtype == 0)
+		if (monster.uniqType == 0)
 			continue;
 		for (int dx = 0; dx < MAXDUNX; dx++) {
 			for (int dy = 0; dy < MAXDUNY; dy++) {
-				if ((dCorpse[dx][dy] & 0x1F) == monster._udeadval)
-					ChangeLightXY(monster.mlid, { dx, dy });
+				if ((dCorpse[dx][dy] & 0x1F) == monster.corpseId)
+					ChangeLightXY(monster.lid, { dx, dy });
```
AJenbo, line 86: `lightId`

aetn23: Fixed

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
-	monster_goal _mgoal;
-	uint8_t _pathcount;
+	uint8_t idx;
+	MonsterMode mode;
+	uint8_t pathCount;
 	/** Direction faced by monster (direction enum) */
-	Direction _mdir;
-	/** The current target of the monster. An index in to either the plr or monster array based on the _meflag value. */
-	uint8_t _menemy;
+	Direction dir;
```
AJenbo, line 209:
```suggestion
	Direction direction;
```

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
+	const MonsterData *data;
+	std::unique_ptr<uint8_t[]> uniqueMonsterTRN;
+	AnimationInfo animInfo;
+	/** Specifies current goal of the monster */
+	monster_goal goal;
+	/** Specifies monster's behaviour regarding moving and changing goals. */
+	int goalGeneral;
+	/** @brief Specifies turning direction for @p RoundWalk in most cases.
+	 *  Used in custom way by @p FallenAi, @p SnakeAi, @p M_FallenFear and @p FallenAi.
+	 */
```
AJenbo, line 180:
```suggestion
	/** 
	 * @brief Specifies turning direction for @p RoundWalk in most cases.
	 * Used in custom way by @p FallenAi, @p SnakeAi, @p M_FallenFear and @p FallenAi.
	 */

```

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
+	int goalTurning;
+	/** @brief Controls monster's behaviour regarding special actions.
+	 *  Used only by @p ScavengerAi and @p MegaAi.
+	 */
```
AJenbo, line 184:
```suggestion
	/**
	 * @brief Controls monster's behaviour regarding special actions.
	 * Used only by @p ScavengerAi and @p MegaAi.
	 */
```

Nice that you added a meaningful name and description to these.

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
+	_speech_id talkMsg;
 	ActorPosition position;
 
 	/** Usually corresponds to the enemy's future position */
 	WorldTilePosition enemyPosition;
-	uint8_t _mMTidx;
-	MonsterMode _mmode;
-	monster_goal _mgoal;
-	uint8_t _pathcount;
+	uint8_t idx;
```
AJenbo, line 205: `levelTypeId`, `typeId`, or maybe `levelType`

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
-	int8_t mWhoHit;
-	int8_t mLevel;
-	uint8_t mMinDamage;
-	uint8_t mMaxDamage;
-	uint8_t mMinDamage2;
-	uint8_t mMaxDamage2;
-	uint8_t mArmorClass;
+	_mai_id ai;
+	/** Specifies monster behaviour across various actions */
+	uint8_t behaviour;
```
ephphatha, line 220: Adding descriptions of how this is used like you've done for the goal vars would be great. This appears to be used as an "aggressiveness threshold", to determine how much damage the acid puddle does when left behind after a spitter attack, to determine how much health fallen gain when recovering, maybe more?

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
-	uint8_t _menemy;
+	Direction dir;
+	/** The current target of the monster. An index in to either the player or monster array based on the _meflag value. */
+	uint8_t enemy;
 
 	/**
 	 * @brief Contains information for current animation
 	 */
-	bool _mDelFlag;
+	bool delFlag;
```
AJenbo, line 216:
```suggestion
	bool delete;
```

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
-	int8_t mLevel;
-	uint8_t mMinDamage;
-	uint8_t mMaxDamage;
-	uint8_t mMinDamage2;
-	uint8_t mMaxDamage2;
-	uint8_t mArmorClass;
+	_mai_id ai;
+	/** Specifies monster behaviour across various actions */
+	uint8_t behaviour;
+	uint8_t squelch;
```
AJenbo, line 221: I'm not sure about this name, I always struggled to understand what it does, maybe simply `active` would be better?

Source/objects.cpp
```diff
@@ -3428,14 +3428,14 @@ void OperateBookCase(int pnum, int i, bool sendmsg)
...
-			zhar.mtalkmsg = TEXT_ZHAR2;
-			M_StartStand(zhar, zhar._mdir); // BUGFIX: first parameter in call to M_StartStand should be MAX_PLRS, not 0. (fixed)
-			zhar._mgoal = MGOAL_ATTACK2;
-			zhar._mmode = MonsterMode::Talk;
+		if (zhar.mode == MonsterMode::Stand // prevents playing the "angry" message for the second time if zhar got aggroed by losing vision and talking again
+		    && zhar.uniqType - 1 == UMT_ZHAR
+		    && zhar.squelch == UINT8_MAX
+		    && zhar.hitPoints > 0) {
+			zhar.talkMsg = TEXT_ZHAR2;
+			M_StartStand(zhar, zhar.dir); // BUGFIX: first parameter in call to M_StartStand should be MAX_PLRS, not 0. (fixed)
```
ephphatha, line 3436: Unrelated to this PR but this bugfix comment is completely irrelevant now 😂

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
+	uint8_t uniqType;
+	uint8_t uniqTrans;
```
AJenbo, line 223: Nit: Looks like this can be a bool

Source/monster.cpp
```diff
@@ -4915,47 +4915,47 @@ bool Monster::IsWalking() const
 	}
 }
 
-bool Monster::IsImmune(missile_id mName) const
+bool Monster::isImmune(missile_id mitype) const
```
AJenbo, line 4918:
```suggestion
bool Monster::isImmune(missile_id missileType) const
```

Source/monster.h
AJenbo: Preferably keep it as `inteligence`, behaviour is both very broad, not more explanatory and it changes what people that are already used to the code expects.

Source/missiles.cpp
```diff
@@ -2134,12 +2134,12 @@ void AddStone(Missile &missile, const AddMissileParameter &parameter)
 		return;
 	}
 
-	// Petrify the targeted monster
+	// petrify the targeted monster
```
AJenbo, line 2137:
```suggestion
	// Petrify the targeted monster
```

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
+	/** Specifies monster's behaviour regarding moving and changing goals. */
+	int goalGeneral;
+	/** @brief Specifies turning direction for @p RoundWalk in most cases.
+	 *  Used in custom way by @p FallenAi, @p SnakeAi, @p M_FallenFear and @p FallenAi.
+	 */
+	int goalTurning;
+	/** @brief Controls monster's behaviour regarding special actions.
+	 *  Used only by @p ScavengerAi and @p MegaAi.
+	 */
+	int goalSpecialAction;
```
AJenbo, line 185:
```suggestion
	int goalVar1;
	/**
	 * @brief Specifies turning direction for @p RoundWalk in most cases.
	 * Used in custom way by @p FallenAi, @p SnakeAi, @p M_FallenFear and @p FallenAi.
	 */
	int goalVar2;
	/**
	 * @brief Controls monster's behaviour regarding special actions. 
	 * Used only by @p ScavengerAi and @p MegaAi.
	 */
	int goalVar3;
```

This is also used for transferering an X coordinate in a X/Y pair with `goalTurning`, so best keep them as `goalVar1`,  `goalVar2`. `goalSpecialAction` is also not used enough for monsters special action that I don't think it warrants renaming when the others are still there.

Nice that you added a meaningful name and description to these :+1:

Source/monster.h
```diff
@@ -167,107 +166,115 @@ struct CMonster {
...
-	uint16_t mExp;
-	uint16_t mHit;
-	uint16_t mHit2;
-	uint16_t mMagicRes;
-	_speech_id mtalkmsg;
+	uint32_t aiSeed;
+	uint16_t exp;
+	uint16_t hit;
+	uint16_t hit2;
+	uint16_t magicRes;
```
AJenbo, line 199: I think it's ok to have the full word here :)
```suggestion
	uint16_t magicResistance;
```

AJenbo:
Alright I'm done reviewing now :)

Source/monster.h
qndel: @AJenbo delete is a reserved keyword ;)

Source/monster.h
qndel: iirc not really because of compatibility with vanilla - in vanilla it was used as an index into light table where monster TRNs were also stored, with this change I'd expect funny things to happen if you load a devilutionx save in vanilla ;D

Source/monster.h
qndel: what about `activityTicks` :D iirc it just stores how many game ticks the monster will remain active

Source/monster.h
aetn23: Yes, those members are problematic: they are used widely in monsters' ai function, sometimes to do seemingly unrelated things (`goalGeneral`  and `goalTurning` being used to calculate X and Y at one point - as you have said). 

I see the reasons to leave them as `goalVars` but those names are enigmatic, at least to me. Still, my solution in this commit reflects only a part of their usage - not ideal.

I checked the code and I think that X & Y calculation are done only by ScavengerAi - the rest seem to use `goalGeneral` as the name and description suggests. Still problematic to come up with an ideal name...

`goalSpecialAction` is used only by two Ais so maybe the name should reflect that?

aetn23:
Merged master into this branch.

FitzRoyX:
"hp" vs "maxhp" is a problem to me. Having the lack of a word serve as differentiator is a recipe for making mental mistakes. How about:

hitPoints -> remainingLife
maxHP -> maxLife

Source/monster.h
AJenbo: my comment makes no sens in context of this var, i wonder what i was originally looking at :thinking:

Source/monster.h
AJenbo: What do you think of `activeForTicks`? Sort of like timeToLive (TTL) which is a common it term. Or maybe `keepAlive`?

aetn23: Maybe `remainingTicks`? Or `ticksToLive`?

AJenbo: ticksToLive sounds fine to me

AJenbo:
hitPoints is well defined, it also implies a number, where life isn't countable. and yeah current/remaining is implied

qndel:
Personally I never liked the inconsistency between using `hitpoints` for current and `hp` for max, can't we use the same format? like hitPoints/maxHitPoints or currentHP/maxHP ? just to use hitpoints OR HP consistently :D

AJenbo:
I'm partial to `hitPoints`/`maxHitPoints`, but I haven't checked what is generally the case.

aetn23:
Renamed to `maxHitPoints`. The rest of the code seems to follow `hitPoints` but `maxHp` logic

FitzRoyX:
> hitPoints is well defined, it also implies a number, where life isn't countable. and yeah current/remaining is implied

And yet:

-currLife used in places as a variable name. Why bother with "curr" if it's not clearer in isolation?
-Life is used for the char sheet and potions
-Stolen life is called "life stealing" and not "hit point stealing"
-"Points" don't follow mana or durability, yet they are countable
-If "remaining" is implied, then we can drop "min" from "minDamage" and imply it from the lack of max?

Source/monster.h
FitzRoyX: That makes it sound like it's going to die at 0. I vote for activityTicks.

aetn23:
I second @AJenbo. `hitPoints` and `maxHitPoints` are clear enough. Also your example with with `minDamage` -> `damage` does not work since:
"The monster has 42 hit points" <-> "The monster has 42 hit points remaining" - that is true in most ordinary contexts, even if we forget about this particular case and game. As such, "remaining" can be skipped.
Meanwhile:
"The monster deals minimum 42 damage" and "The monster deals 42 damage" mean complete different things in all ordinary contexts. "minimum" cannot be skipped without changing the meaning.

Source/monster.h
aetn23: 'ticksToBeActive`? `activeForTicks` was mentioned previously and on second thought seems to be good one - coz it describes what is happening exactly and avoids the problem @FitzRoyX mentioned.

Source/monster.h
```diff
@@ -168,100 +167,115 @@ struct CMonster {
...
-	int _mgoalvar2;
-	int _mgoalvar3;
-	int _mVar1;
-	int _mVar2;
-	int _mVar3;
-	int _mmaxhp;
-	int _mhitpoints;
-	uint32_t _mFlags;
+	const char *name;
+	std::unique_ptr<uint8_t[]> uniqueMonsterTRN;
```
Iniquitatis, line 171: Why repeat `Monster` here? This member already belongs to the `Monster` struct.

Source/player.h
```diff
@@ -635,7 +635,7 @@ struct Player {
 	void RestorePartialLife();
 
 	/**
-	 * @brief Resets hp to maxHp
+	 * @brief Resets hp to maxHitPoints
```
AJenbo, line 638: I don't think you changed this?

Source/monster.h
```diff
@@ -168,100 +167,115 @@ struct CMonster {
...
-	uint8_t _uniqtrans;
-	int8_t _udeadval;
-	int8_t mWhoHit;
-	int8_t mLevel;
-	uint8_t mMinDamage;
-	uint8_t mMaxDamage;
-	uint8_t mMinDamage2;
-	uint8_t mMaxDamage2;
-	uint8_t mArmorClass;
+	bool invalidate;
```
AJenbo, line 217: I think this makes it more clear:
```suggestion
	bool isInvalid;
```

Source/monster.h
```diff
@@ -168,100 +167,115 @@ struct CMonster {
...
-	Direction _mdir;
-	/** The current target of the monster. An index in to either the plr or monster array based on the _meflag value. */
-	uint8_t _menemy;
+	Direction direction;
+	/** The current target of the monster. An index in to either the player or monster array based on the _meflag value. */
+	uint8_t enemy;
 
 	/**
 	 * @brief Contains information for current animation
 	 */
```
AJenbo, line 216: Nit: Not related to this PR, but this comment belong somewhere else.

Source/monster.h
StephenCWills: I'm not sure if the name `uniqueTRN` is enough to convey that this is the TRN for when the monster is a unique monster as opposed to the monster's very own TRN.

Source/player.h
aetn23: Oh, I must have gone a little too far with automatic renaming. Fixed

Source/monster.h
aetn23: Moved it to `animInfo`

Source/monster.h
aetn23: I agree with @StephenCWills  - this `monster` in `uniqueMonsterTRN` is attached to the specific monster being unique and not to every monster in general. Removing the monster would create more confusing name, implying the TRN is unique, not the monster.

Source/monster.h
AJenbo: Maybe revert these for now, then we can take them separately in a different PR as they are going to take some more debating before we will probably land on good meaningful names for them.

PR #4863, opened 2022-07-03
Take player by reference in UseMana

AJenbo:
@ephphatha already had this one ready when the other was merged, but decided to look for possible cases of missing `const` before sending it your way ;)

This leaves just 3 files with 47 lines to look through to se if we can reduce the diff for https://github.com/diasurgical/devilutionX/pull/4754


obligaron (approved):
lgtm

PR #4864, opened 2022-07-03
Change animation frames and tick per frame to int8_t

### LINKED ISSUE #4849: Make AnimationInfo smaller
`sizeof(AnimationInfo)` is currently 64 bytes, this is ~27% of the size of the `Monster` struct (after the optimizations in #4824).

@obligaron Can you please look into making it smaller?

One place to watch out for is this:

https://github.com/diasurgical/devilutionX/blob/ad7ca97ae1d6c0c23116f3960240e0501f9ce29a/Source/player.cpp#L1389

obligaron:
Fixes #4849

Notes:
- a signed type is need, cause
  - since #4221 we have a zero-based animation index. That means the start value of a animation is -1 (it will increase to 0 in the same game tick where it's starts).
  - in some cases the animation takes longer then we have animation frames (walking). So we could also end up with a negative index
- it's no breaking change to change the player death animation (first commit). the death animation is never stored in a save game, cause you can't save when you are dead and in multiplayer we don't save the world state.
- I changed all fields that hold a frame / tick per frame to `int8_t`, so that it's consistent over the codebase. This also reduces the size of `Player` even more.
- Reduces size of `AnimationInfo` by 16 bytes
- Reduces size of `Player` by additional 920 bytes
- Ensures that when loading data we don't overflow. It's only a precaution, cause frame / tick per frame should always fit in `int8_t`.

### DIFF
--- a/Source/engine/animationinfo.cpp
+++ b/Source/engine/animationinfo.cpp
@@ -12,14 +12,14 @@
 
 namespace devilution {
 
-int AnimationInfo::GetFrameToUseForRendering() const
+int8_t AnimationInfo::GetFrameToUseForRendering() const
 {
 	// Normal logic is used,
 	// - if no frame-skipping is required and so we have exactly one Animationframe per game tick
 	// or
 	// - if we load from a savegame where the new variables are not stored (we don't want to break savegame compatiblity because of smoother rendering of one animation)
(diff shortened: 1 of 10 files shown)


glebm (approved):
Thanks!

PR #4865, opened 2022-07-03
Monster: Remove MType/MData pointers

glebm:
These pointers are redundant, replaces them with methods.

`Monster` is down to 176 bytes, `Monsters` down to 33,600 bytes.


ephphatha (approved):
makes the `monster.type().type` uses stand out more, hopefully we can make that look nicer in the future 😂

PR #4874, opened 2022-07-03
Take player by reference in StartWalkStand and RemovePlrMissiles

AJenbo:
Unless someone fixes `RemovePlrFromMap()` then this may well be the last prep for https://github.com/diasurgical/devilutionX/pull/4754, once rebased I will change the remaining pnum from int to size_t to avoid the remaning static_casting.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -390,18 +390,13 @@ void ClearStateVariables(Player &player)
 	player.position.offset2 = { 0, 0 };
 }
 
-void StartWalkStand(int pnum)
+void StartWalkStand(Player &player)
 {
-	if ((DWORD)pnum >= MAX_PLRS) {
-		app_fatal(fmt::format("StartWalkStand: illegal player {}", pnum));
-	}
-	Player &player = Players[pnum];
-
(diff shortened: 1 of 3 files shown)


Source/player.cpp
```diff
@@ -3203,7 +3198,7 @@ void RemovePlrMissiles(int pnum)
 	}
 
 	for (auto &missile : Missiles) {
-		if (missile._mitype == MIS_STONE && missile._misource == pnum) {
+		if (missile._mitype == MIS_STONE && &Players[missile._misource] == &player) {
```
StephenCWills, line 3201: This made me a little nervous, but tracing the code for Rune of Stone suggests that `MIS_STONE` should indeed always have a player as its source. Note, I wasn't able to test in Hellfire due to an unrelated error.

StephenCWills (approved):

Source/player.cpp
AJenbo: Yeah, I'm hoping we can make a function on missiles that will return the player /monster or null

AJenbo:
Thanks for the review

PR #4877, opened 2022-07-03
Normalise the player and monster walking function naming

canassa:
Both player and monster walking functions are now named WalkNorthwards, WalkSouthwards and WalkSideways.

Also renames the PLT_MODE enum to follow the same pattern and removes unnecessary comments.


AJenbo (approved):
Especially nice with the enum

PR #4882, opened 2022-07-03
Remove monsterId from M_GetKnockback

canassa:
Also: 

- renamed `MonsterMHit` m parameter to `monsterId`
- Renamed `PlrHitMonst` m parameter to `monsterId`

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1434,8 +1434,7 @@ DWORD OnKnockback(const TCmd *pCmd, int pnum)
 	const auto &message = *reinterpret_cast<const TCmdParam1 *>(pCmd);
 
 	if (gbBufferMsgs != 1 && Players[pnum].isOnActiveLevel() && message.wParam1 < MaxMonsters) {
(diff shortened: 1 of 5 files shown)


Source/msg.cpp
```diff
@@ -1434,7 +1434,8 @@ DWORD OnKnockback(const TCmd *pCmd, int pnum)
 	const auto &message = *reinterpret_cast<const TCmdParam1 *>(pCmd);
 
 	if (gbBufferMsgs != 1 && Players[pnum].isOnActiveLevel() && message.wParam1 < MaxMonsters) {
-		M_GetKnockback(message.wParam1);
+		Monster &monster = Monsters[message.wParam1];
+		M_GetKnockback(monster);
```
AJenbo, line 1438: If it's not used for anything else might as well just inline it.
```suggestion
		M_GetKnockback(Monsters[message.wParam1]);
```

canassa: Done, thanks!

AJenbo (approved):

PR #4885, opened 2022-07-04
:recycle: Refactor NetSendCmdMonDmg to receive a Monster reference

canassa:


AJenbo (changes requested):
This operates purely on the id and should not be changed to depend on the object

Issue #4871, opened 2022-07-03
Unable to load saved game where animation counter has overflowed


ephphatha:
**Important information**
Win 11 / DevilutionX since https://github.com/diasurgical/devilutionX/commit/a6132a1aa81bf3ef36d05602465195b24556ce9c / compiled

**Describe**
Loading a save where the ticksPerAnimationFrame value have exceeded the limits of int8_t triggers an assert and crashes the game.

**To Reproduce**
1. Try to load the attached save (has a monster with 314 ticks on the current frame, > int8_t::max)
[single_1.zip](https://github.com/diasurgical/devilutionX/files/9034919/single_1.zip)

**Expected behavior**
The game loads

**Additional context**
fails at https://github.com/diasurgical/devilutionX/blob/17f8cdddc4d2ae9aa53b6c345be4203c39785c3f/Source/loadsave.cpp#L597

Perhaps this should truncate instead of crashing in this context?

aetn23:
I have read the code a little and I think that in this particular case removing the assert should work without breaking things. It will simply fallback to either min or max value available (because of clamp and arguments it gets in this function). In case of `TickCounterOfCurrentFrame` it will simply set a counter towards next frame of animation to max.

I wonder how it is even possible to have a save with monster with `TickCounterOfCurrentFrame = 314`? The type of this variable is int8_t, so it can't hold that big of a value. I think something can go south while saving, resulting in a monster having a variable with value that should be impossible. 

I tried reading saving/loading code and discovered that while saving the game `TickCounterOfCurrentFrame` is saved as as int32_t. I don't know if this is intentional or not (int8_t being saved as int32_t, then read as int8_t). Maybe this 314 is a result of some operations done while saving? 

Finally, I changed saving the `TickCounterOfCurrentFrame` to int8_t and after changing the loading similarly, I successfully managed to save and load, without any issues at first glance (of course first I have to do the saving, then save, then change loading, then load). This confuses me further regarding previous paragraph.

glebm:
The type used to be `int` a few commits ago, the save is probably from before it was changed to `int8_t`

AJenbo:
> I tried reading saving/loading code and discovered that while saving the game TickCounterOfCurrentFrame is saved as as int32_t. I don't know if this is intentional or not (int8_t being saved as int32_t, then read as int8_t). Maybe this 314 is a result of some operations done while saving?

[...]

PR #4890, opened 2022-07-04
Remove assert to check if value is in bounds in NextLENarrow.

aetn23:
Remove assert to check if value is in bounds in NextLENarrow. The behaviour now is to fallback to either minimum or maximum a type can hold (it was there before, as std::clamp is used. It was not possible for clamp to fallback to min/max, since the assert to prevent this was present).

This fixes issue  #4871. It potentially will fix other issues with older saves, in case where a variable ranged has been changed between versions - the save will load and fallback per std::clamp.

The fallback to min/max means that variables may get unexpected values assigned - always valid though. I didn't see a case where it could break things, but such possibility exists. Nonetheless, this change will allow older saves alike to the one in issue #4871 load - and then maybe break.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -148,7 +148,6 @@ public:
 		static_assert(std::numeric_limits<TSource>::min() < std::numeric_limits<TDesired>::min());
 		static_assert(std::numeric_limits<TSource>::max() > std::numeric_limits<TDesired>::max());
 		TSource value = SwapLE(Next<TSource>()) + modifier;
-		assert(value >= std::numeric_limits<TDesired>::min() && value <= std::numeric_limits<TDesired>::max());
 		return static_cast<TDesired>(clamp<TSource>(value, std::numeric_limits<TDesired>::min(), std::numeric_limits<TDesired>::max()));
 	}
 


obligaron (approved):
lgtm, didn't thought about gargoyle edge case (see #4893).

StephenCWills:
Maybe an assert is too strong, but it seems like a decent place for logging a warning.

obligaron:
How fast is logging?
The assert/warning should only happen for gargoyles (see #4893).
If we have 50 gargoyles on a level, how fast is logging 50x? 🤔 

If it's no problem I think this is a good idea. 🙂 

a little bit off topic:
Could we change [Golem-AI-Assert](https://github.com/diasurgical/devilutionX/blob/dab0517f21e1a996b5481bfb52ee65a90658d803/Source/monster.cpp#L4088) also to a log?
I often have to remove the assert when testing vanilla saves.

StephenCWills:
> How fast is logging? The assert/warning should only happen for gargoyles (see #4893). If we have 50 gargoyles on a level, how fast is logging 50x? 🤔
> 
> If it's no problem I think this is a good idea. 🙂

Depends on the platform. It could be very slow on Vita so maybe the log isn't even worth it.

I do think it's important to consider the gargoyle case here in terms of performance, because it does indeed seem reasonable to expect that a save could log 50x at load, and logging is enabled for release builds. However, this function is used for more than just `tickCounterOfCurrentFrame` and could conceivably be used more in the future, right?

> a little bit off topic: Could we change [Golem-AI-Assert](https://github.com/diasurgical/devilutionX/blob/dab0517f21e1a996b5481bfb52ee65a90658d803/Source/monster.cpp#L4088) also to a log? I often have to remove the assert when testing vanilla saves.

I've certainly never made an error that triggered that assert. Also, it seems like a very old assertion. Maybe it would be good to demote it to a log message.

AJenbo:
Yeah let's demote golem assert

aetn23:
What's the story behind GolumAi assert? It checks for monsterId < MAX_PLRS while other ais check for MaxMonsters. Why is that? What happens if monsterId >= MAX_PLRS?

Also this should be done in separate PR to avoid confusion, right?

glebm:
@aetn23 Each player can have at most 1 golem and they're stored at `Monsters[playerId]`.

StephenCWills:
> Also this should be done in separate PR to avoid confusion, right?

Yes, it was a little bit off-topic, but it was somewhat relevant to the discussion around this PR so we had it here. Sorry for any confusion.

PR #4892, opened 2022-07-04
artifacts as releases

qndel:
Since there doesn't seem to be any limit on releases, this should be the best way atm :D (Thx @Trihedraf for the idea)
https://github.com/artifacts-storage/devilutionx-artifacts
The question is if we want to also have linux artifacts (if yes, I'll push it into this PR so we have both windows and linux artifacts from the same moment)


Trihedraf (approved):

AJenbo (approved):
Afaik they factor in to the total repo limit

PR #4893, opened 2022-07-04
Avoid overflow of tickCounterOfCurrentFrame

obligaron:
When `MFLAG_ALLOW_SPECIAL` was set on a monster, the `tickCounterOfCurrentFrame` was increased but nothing more.

`MFLAG_ALLOW_SPECIAL` is set, when

- a the gargoyle sits on the ground
- or a mega (winged demon) sits on the ground and casts inferno

Both are specified to stop the animation from advancing.
The logic [exists in vanilla](https://github.com/diasurgical/devilution/blob/b86b8fd7c4966385e3a7dbcb29e6d7e05bdab95e/Source/monster.cpp#L5663).
This could result in overflowing the `tickCounterOfCurrentFrame` (for gargoyles, for mega the duration is too short).

With this pr `tickCounterOfCurrentFrame` isn't blindly increased anymore.
The logic is simplified by using the same approach as petrified.

Notes
- Backwards compatibility: A `tickCounterOfCurrentFrame` > `ticksPerFrame` is no problem. As long as it doesn't overflow in negative (the change in `loadsave.cpp` ensures that this doesn't happen even with vanilla saves).
- Without this pr a gargoyles `tickCounterOfCurrentFrame`, could overflow and let him awaken later (cause a overflown negative `tickCounterOfCurrentFrame` needs to be increased to a positive value again).
- In vanilla this could theoretical also happen (after ~29k hours in a level with a gargoyle) and then the gargoyle would need the same time to awaken again. 😜 
- #4890 is still needed, cause I didn't thought about this edge case when adding the assert.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1467,7 +1467,7 @@ bool MonsterRangedSpecialAttack(int monsterId)
 	assert(monsterId >= 0 && monsterId < MaxMonsters);
 	auto &monster = Monsters[monsterId];
 
-	if (monster.AnimInfo.currentFrame == monster.data().mAFNum2 - 1 && monster.AnimInfo.tickCounterOfCurrentFrame == 0 && monster._mVar2 == 0) {
+	if (monster.AnimInfo.currentFrame == monster.data().mAFNum2 - 1 && monster.AnimInfo.tickCounterOfCurrentFrame == 0 && (monster._mAi != AI_MEGA || monster._mVar2 == 0)) {
 		if (AddMissile(
 		        monster.position.tile,
 		        monster.enemyPosition,
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1467,7 +1467,7 @@ bool MonsterRangedSpecialAttack(int monsterId)
 	assert(monsterId >= 0 && monsterId < MaxMonsters);
 	auto &monster = Monsters[monsterId];
 
-	if (monster.AnimInfo.currentFrame == monster.data().mAFNum2 - 1 && monster.AnimInfo.tickCounterOfCurrentFrame == 0) {
+	if (monster.AnimInfo.currentFrame == monster.data().mAFNum2 - 1 && monster.AnimInfo.tickCounterOfCurrentFrame == 0 && monster._mVar2 == 0) {
 		if (AddMissile(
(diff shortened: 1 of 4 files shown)


Source/monster.cpp
```diff
@@ -1467,7 +1467,7 @@ bool MonsterRangedSpecialAttack(int monsterId)
 	assert(monsterId >= 0 && monsterId < MaxMonsters);
 	auto &monster = Monsters[monsterId];
 
-	if (monster.AnimInfo.currentFrame == monster.data().mAFNum2 - 1 && monster.AnimInfo.tickCounterOfCurrentFrame == 0) {
+	if (monster.AnimInfo.currentFrame == monster.data().mAFNum2 - 1 && monster.AnimInfo.tickCounterOfCurrentFrame == 0 && monster._mVar2 == 0) {
```
AJenbo, line 1470: What does var2 do here? Can we trust it wasn't set for some odd reason that we didn't think off?

obligaron: - For `_mVar2` Mega-AI it [counts the game ticks](https://github.com/diasurgical/devilutionX/blob/7612fb4bb5636f993afbad9e15fffa09e204ff89/Source/monster.cpp#L1482-L1488) that the mega sits when casting inferno.
- The added check ensures that Mega-AI casts only one inferno while sitting.
- `_mVar2` is initialized in [StartRangedSpecialAttack](https://github.com/diasurgical/devilutionX/blob/7612fb4bb5636f993afbad9e15fffa09e204ff89/Source/monster.cpp#L905). 
- `_mVar2` is also used for walking, delay and normal ranged attacks. But in these cases monsters have a different `_mmode`.

If wanted i can modify the check to `&& (monster._mAi != AI_MEGA || monster._mVar2 == 0)`.
This would make it even more safe, cause in case of Mega-AI  `_mVar2` must be correct, else the sitting would not work as expected.

AJenbo:
```suggestion
	if (monster.AnimInfo.currentFrame == monster.data().mAFNum2 - 1 && monster.AnimInfo.tickCounterOfCurrentFrame == 0 && (monster._mAi != AI_MEGA || monster._mVar2 == 0)) {
```

PR #4894, opened 2022-07-04
:recycle: Refactor AddMonster to return a monster pointer

canassa:
AddMonster now returns a Monster pointer or nullptr instead of a monster id. Similar to the existing AddMissile

Also refactored AddSkeleton, SpawnSkeleton and PreSpawnSkeleton to work with pointers instead of ids


Source/objects.cpp
```diff
@@ -1075,7 +1075,7 @@ void AddSarc(int i)
 	Objects[i]._oVar1 = GenerateRnd(10);
 	Objects[i]._oRndSeed = AdvanceRndSeed();
 	if (Objects[i]._oVar1 >= 8)
-		Objects[i]._oVar2 = PreSpawnSkeleton();
+		Objects[i]._oVar2 = PreSpawnSkeleton()->getId();
```
AJenbo, line 1078: previously if PreSpawnSkeleton() failed it would set ovar2 to -1, now it would result in UB, so we need to check if PreSpawnSkeleton() was successful.

Source/objects.cpp
```diff
@@ -2495,7 +2495,7 @@ void OperateSarc(int i, bool sendMsg, bool sendLootMsg)
 	if (Objects[i]._oVar1 <= 2)
 		CreateRndItem(Objects[i].position, false, sendLootMsg, false);
 	if (Objects[i]._oVar1 >= 8)
```
AJenbo, line 2497: best to also check that _oVar2 is in a valid range

Source/objects.cpp
```diff
@@ -1074,8 +1074,15 @@ void AddSarc(int i)
...
 	Objects[i]._oVar1 = GenerateRnd(10);
 	Objects[i]._oRndSeed = AdvanceRndSeed();
-	if (Objects[i]._oVar1 >= 8)
-		Objects[i]._oVar2 = PreSpawnSkeleton()->getId();
+	if (Objects[i]._oVar1 >= 8) {
+		Monster *monster = PreSpawnSkeleton();
+		if (monster != nullptr) {
+			Objects[i]._oVar2 =	monster->getId();
+		} else {
+			Objects[i]._oVar1 = 0;
```
ephphatha, line 1082: 0 is a valid monster ID (though it'll never be used in practice). `-1` like the function previously returned on failure is better.

AJenbo:
```suggestion
			Objects[i]._oVar1 = -1;
```

Source/objects.cpp
```diff
@@ -1127,8 +1134,14 @@ void AddBarrel(Object &barrel)
...
 	barrel._oVar3 = GenerateRnd(3);
 
-	if (barrel._oVar2 >= 8)
-		barrel._oVar4 = PreSpawnSkeleton()->getId();
+	if (barrel._oVar2 >= 8) {
+		Monster *skeleton = PreSpawnSkeleton();
+		if (skeleton != nullptr)
+			barrel._oVar4 = skeleton->getId();
+		else {
+			barrel._oVar2 = 0;
```
AJenbo, line 1142:
```suggestion
			barrel._oVar2 = -1;
```

Source/objects.cpp
```diff
@@ -1127,8 +1134,14 @@ void AddBarrel(Object &barrel)
 	barrel._oVar2 = barrel.isExplosive() ? 0 : GenerateRnd(10);
 	barrel._oVar3 = GenerateRnd(3);
 
-	if (barrel._oVar2 >= 8)
-		barrel._oVar4 = PreSpawnSkeleton();
+	if (barrel._oVar2 >= 8) {
+		Monster *skeleton = PreSpawnSkeleton();
+		if (skeleton != nullptr)
+			barrel._oVar4 = skeleton->getId();
+		else {
```
AJenbo, line 1141:
```suggestion
		if (skeleton != nullptr) {
			barrel._oVar4 = skeleton->getId();
		} else {
```

Source/objects.cpp
```diff
@@ -1074,8 +1074,15 @@ void AddSarc(int i)
...
-		Objects[i]._oVar2 = PreSpawnSkeleton();
+	if (Objects[i]._oVar1 >= 8) {
+		Monster *monster = PreSpawnSkeleton();
+		if (monster != nullptr) {
+			Objects[i]._oVar2 = monster->getId();
+		} else {
+			Objects[i]._oVar1 = 0;
+		}
+
+	}
```
AJenbo, line 1085:
```suggestion
		}
	}
```

Source/objects.cpp
```diff
@@ -1074,8 +1074,14 @@ void AddSarc(int i)
...
 	Objects[i]._oVar1 = GenerateRnd(10);
 	Objects[i]._oRndSeed = AdvanceRndSeed();
-	if (Objects[i]._oVar1 >= 8)
-		Objects[i]._oVar2 = PreSpawnSkeleton();
+	if (Objects[i]._oVar1 >= 8) {
+		Monster *monster = PreSpawnSkeleton();
+		if (monster != nullptr) {
+			Objects[i]._oVar2 = monster->getId();
+		} else {
+			Objects[i]._oVar1 = -1;
```
AJenbo, line 1082:
```suggestion
			Objects[i]._oVar2 = -1;
```

Source/objects.cpp
```diff
@@ -1127,8 +1133,14 @@ void AddBarrel(Object &barrel)
...
 	barrel._oVar3 = GenerateRnd(3);
 
-	if (barrel._oVar2 >= 8)
-		barrel._oVar4 = PreSpawnSkeleton();
+	if (barrel._oVar2 >= 8) {
+		Monster *skeleton = PreSpawnSkeleton();
+		if (skeleton != nullptr) {
+			barrel._oVar4 = skeleton->getId();
+		} else {
+			barrel._oVar2 = -1;
```
AJenbo, line 1141:
```suggestion
			barrel._oVar4 = -1;
```

Source/objects.cpp
AJenbo: Or was the intent here to set `_oVar1` to a number less then 8 to avoid the later loading, with out checking that `_oVar2` is out of range? If that is the case then it should indeed be set to 0, but a comment should be added to make this clear.

canassa: Yeah, that was the original intent. But I think that this approach is more clear, so let's go ahead with it 😄

canassa:
@AJenbo Let me comment here because GitHub is putting all my comments as "Pending" for some reason.

Yeah, that was the original intent was to set the variable that stores the result of `GenerateRnd(10)` to a number < 8 so that the code that spawns the skeleton not even runs. But now that I think about it, just storing -1 as the monsterId is more clear, even if it requires an extra check later

AJenbo:
> @AJenbo Let me comment here because GitHub is putting all my comments as "Pending" for some reason.

You have started a review by adding a comment as a review comment rather then commenting directly. You need to submit the review to send the pending comments before you can return to normal comments.

canassa:
I pushed the code that checks if the monsterId is -1 before spawning the skeleton.

I checked the `BreakBarrel` code again and messing around with the `_oVar2` wasn't a good idea. It probably changed the loot algorithm.

AJenbo:
Looks like it would have to be 2-7 to have no effect :)

PR #4895, opened 2022-07-05
Add StrCat and StrAppend

glebm:
Adds simple string / integer concatenation functions.

Many of the uses of `fmt::format` are simply concatenation of a few strings and integers.

`StrCat` is an easier-to-read alternative to such uses of `fmt`.

"Easier-to-read" is subjective. How do you feel about this?

@obligaron @ephphatha @AJenbo


Source/DiabloUI/selhero.cpp
```diff
@@ -78,12 +79,12 @@ void SelheroFree()
...
-	CopyUtf8(textStats[2], fmt::format("{}", selhero_heroInfo.magic), sizeof(textStats[2]));
-	CopyUtf8(textStats[3], fmt::format("{}", selhero_heroInfo.dexterity), sizeof(textStats[3]));
-	CopyUtf8(textStats[4], fmt::format("{}", selhero_heroInfo.vitality), sizeof(textStats[4]));
-	CopyUtf8(textStats[5], fmt::format("{}", selhero_heroInfo.saveNumber), sizeof(textStats[5]));
+	CopyUtf8(textStats[0], StrCat(selhero_heroInfo.level), sizeof(textStats[0]));
+	CopyUtf8(textStats[1], StrCat(selhero_heroInfo.strength), sizeof(textStats[1]));
+	CopyUtf8(textStats[2], StrCat(selhero_heroInfo.magic), sizeof(textStats[2]));
+	CopyUtf8(textStats[3], StrCat(selhero_heroInfo.dexterity), sizeof(textStats[3]));
+	CopyUtf8(textStats[4], StrCat(selhero_heroInfo.vitality), sizeof(textStats[4]));
+	CopyUtf8(textStats[5], StrCat(selhero_heroInfo.saveNumber), sizeof(textStats[5]));
```
ephphatha, line 87: I gather this was being used to force a copy, is it still necessary?

Source/utils/str_cat.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * @brief Appends the integer to the given string.
+ */
+void StrAppend(std::string &out, int value);
+
+/**
+ * @brief Appends the given string_view to the given buffer.
+ */
+inline char *BufAppend(char *out, string_view value)
```
ephphatha, line 25: This writes to the given pointer directly overwriting anything that was already there. The string version actually appends to the string, the BufAppend functions would be clearer with a different name imo.

```suggestion
 * @brief Copies the given string_view to the given buffer.
 *
 * Does not null terminate the resulting c-str, does not truncate if the string_view is larger than the buffer
 */
inline char *BufCopy(char *out, string_view value)
```

Source/utils/str_cat.hpp
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @brief Appends the given string_view to the given string.
+ */
+inline void StrAppend(std::string &out, string_view value)
+{
+	out.append(value.data(), value.size());
+}
+
+/**
+ * @brief Appends the given C string to the given buffer.
```
ephphatha, line 40:
```suggestion
 * @brief Appends the given C string to the given buffer.
 *
 * str must be a null-terminated C string, out will not be null terminated
```

Source/utils/str_cat.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include "utils/stdcompat/string_view.hpp"
+
+namespace devilution {
+
+/**
+ * @brief Appends the integer to the given buffer.
+ * @return char* end of the buffer
+ */
+char *BufAppend(char *out, int value);
```
ephphatha, line 15: Probably safest to make these take a buffer size argument as well. Would it impact performance much to make them write a null character to the end of the string even if most of the time the null gets overwritten by a chained call?

Source/DiabloUI/selhero.cpp
glebm: It's to convert an int to a string

Source/stores.cpp
```diff
@@ -1022,7 +1023,7 @@ void StoreConfirm(Item &item)
 		prompt = _("Are you sure you want to repair this item?");
 		break;
 	default:
-		app_fatal(fmt::format("Unknown store dialog {}", static_cast<int>(stextshold)));
+		app_fatal(StrCat("Unknown store dialog ", static_cast<int>(stextshold)));
```
ephphatha, line 1026: This sort of use I think is an improvement, a single static string followed by some runtime value

Source/pfile.cpp
```diff
@@ -316,19 +288,19 @@ void pfile_write_hero(bool writeGameData)
 
 void pfile_write_hero_demo(int demo)
 {
-	std::string savePath = GetSavePath(gSaveNumber, fmt::format("demo_{}_reference_", demo));
+	std::string savePath = GetSavePath(gSaveNumber, StrCat("demo_", demo, "_reference_"));
```
ephphatha, line 291: I'm not as fond of changes like this, but I think it's acceptable (especially if there's a comment outlining the benefits over using fmt::format).

Source/options.cpp
```diff
@@ -730,7 +731,7 @@ void OptionEntryResolution::CheckResolutionsAreInitialized() const
 	sizes.erase(std::unique(sizes.begin(), sizes.end()), sizes.end());
 
 	for (auto &size : sizes) {
-		resolutions.emplace_back(size, fmt::format("{}x{}", size.width, size.height));
+		resolutions.emplace_back(size, StrCat(size.width, "x", size.height));
```
ephphatha, line 734: Can we make this a string cast/conversion operator on size? Could probably get rid of the ostream function for testing if it's allowed as an implicit conversion... but not sure if that'd cause issues elsewhere in the code.

Source/DiabloUI/selhero.cpp
ephphatha: Ahh, at least the IDE should show the types involved 😂 . I was wondering if there could be a ToString function but I gather that's not gonna be an option without significant hassle around the varargs templates

Source/debug.cpp
```diff
@@ -97,16 +95,11 @@ void PrintDebugMonster(int m)
...
-	                monster.position.tile.y,
-	                monster._menemy,
-	                monster._mhitpoints,
-	                static_cast<int>(monster._mmode),
-	                monster._mVar1),
+	EventPlrMsg(StrCat(
+	                "Monster ", m, " = ", monster.mName,
+	                "\nX = ", monster.position.tile.x, ", Y = ", monster.position.tile.y,
+	                "\nEnemy = ", monster._menemy, ", HP = ", monster._mhitpoints,
+	                "\nMode = ", static_cast<int>(monster._mmode), ", Var1 = ", monster._mVar1),
```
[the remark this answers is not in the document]
glebm: One issue with the previous format was that it was difficult to check that the order of values is correct. Here the values are right next to the labels so it's immediately clear.

Source/options.cpp
glebm: It is not obvious that the string conversion would produce this format. I think it's better to keep it explicit here.

Source/utils/str_cat.hpp
glebm: Writing a '\0' would require larger buffers. Most of the time, we don't use null-terminated strings.

Source/utils/str_cat.hpp
glebm: Done

Source/utils/str_cat.hpp
glebm: We don't use this function for user input or anything else with unknown size but we can add a sized version if we need it at some point

ephphatha: Sure, makes sense. I think with the rename there's enough distinction between BufCopy and StrAppend that it shouldn't be surprising to programmers that BufCopy doesn't produce c-strings, could always expand the comments a bit more to call out the behaviour if necessary.

glebm:
This also makes the rg99 binary 7.3 KiB smaller (1369448 -> 1361928)

AJenbo (approved):
This does add more specialization to the code base, and for the most part there aren't any big wins for the readability. Pfile and CharPanel are the biggest winners here, pfile could probably have been cleaned up previously as well.

AJenbo:
I'm not against the monster path change either, sometimes it's nice to have an enforced structure as it prevents people from putting files in random places.

PR #4896, opened 2022-07-05
Remove #defines in player.h and add constexpr where applicable 

aetn23:
Remove #defines in player.h and add constexpr where applicable in player.h/cpp. 

This is done to bring the code closer to modern C++

### DIFF
--- a/Source/dvlnet/base_protocol.h
+++ b/Source/dvlnet/base_protocol.h
@@ -292,7 +292,7 @@ template <class P>
 void base_protocol<P>::recv_decrypted(packet &pkt, endpoint_t sender)
 {
 	if (pkt.Source() == PLR_BROADCAST && pkt.Destination() == PLR_MASTER && pkt.Type() == PT_INFO_REPLY) {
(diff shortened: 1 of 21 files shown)


Source/dvlnet/base_protocol.h
```diff
@@ -292,7 +292,7 @@ template <class P>
 void base_protocol<P>::recv_decrypted(packet &pkt, endpoint_t sender)
 {
 	if (pkt.Source() == PLR_BROADCAST && pkt.Destination() == PLR_MASTER && pkt.Type() == PT_INFO_REPLY) {
-		constexpr size_t sizePlayerName = (sizeof(char) * PLR_NAME_LEN);
+		constexpr size_t sizePlayerName = (sizeof(char) * PlayerNameLength);
```
ephphatha, line 295: You might as well replace `sizePlayerName` with `PlayerNameLength` here and in `recv_ingame`, `sizeof(char)` is defined as 1 and we're using UTF8 strings so the backing type is never going to change.

AJenbo (approved):

PR #4900, opened 2022-07-06
Rename Monster's goals

mikolaj-pirog:
Rename goals in Monster struct to more descriptive names. This is done to improve readability and make code less cryptic.

This PR is a result of discussion in  #4861

The names suggested are not the best ones: they do not cover every use case, namely `goalTurning` and `goalGeneral` is used in a custom way by some Ais, which is not reflected by their names. Because the goals are used in a various ways across monster code it is hard to come up with a great name. I think it would be best if we change the current names to something, since I believe that having descriptive name that covers most of the usage is better than having a general, but very cryptic name. 

The suggestions of other names and/or ideas are more than welcome.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -568,9 +568,9 @@ void LoadMonster(LoadHelper *file, Monster &monster)
 	monster.mode = static_cast<MonsterMode>(file->NextLE<int32_t>());
 	monster.goal = static_cast<monster_goal>(file->NextLE<uint8_t>());
 	file->Skip(3); // Alignment
-	monster.goalVar1 = file->NextLE<int32_t>();
-	monster.goalVar2 = file->NextLE<int32_t>();
-	monster.goalVar3 = file->NextLE<int32_t>();
+	monster.goalGeneral = file->NextLE<int32_t>();
+	monster.goalTurning = file->NextLE<int32_t>();
+	monster.goalSpecialAction = file->NextLE<int32_t>();
 	file->Skip(4); // Unused
 	monster.pathCount = file->NextLE<uint8_t>();
 	file->Skip(3); // Alignment
@@ -1310,9 +1310,9 @@ void SaveMonster(SaveHelper *file, Monster &monster)
 	file->WriteLE<int32_t>(static_cast<int>(monster.mode));
 	file->WriteLE<uint8_t>(monster.goal);
 	file->Skip(3); // Alignment
-	file->WriteLE<int32_t>(monster.goalVar1);
-	file->WriteLE<int32_t>(monster.goalVar2);
-	file->WriteLE<int32_t>(monster.goalVar3);
+	file->WriteLE<int32_t>(monster.goalGeneral);
+	file->WriteLE<int32_t>(monster.goalTurning);
+	file->WriteLE<int32_t>(monster.goalSpecialAction);
 	file->Skip(4); // Unused
 	file->WriteLE<uint8_t>(monster.pathCount);
 	file->Skip(3); // Alignment
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -217,9 +217,9 @@ void InitMonster(Monster &monster, Direction rd, int mtype, Point position)
 	monster.ai = monster.data().mAi;
 	monster.intelligence = monster.data().mInt;
 	monster.goal = MGOAL_NORMAL;
-	monster.goalVar1 = 0;
-	monster.goalVar2 = 0;
-	monster.goalVar3 = 0;
+	monster.goalGeneral = 0;
+	monster.goalTurning = 0;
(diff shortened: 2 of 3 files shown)


FitzRoyX:
goalVar3, at least for scavengers, appears to be "remainingTicksToEat" or something, right?

canassa:
I was thinking about how to improve that as well. In a green field project, we would probably design custom monsters as subclasses of Monster and add any extra field to the subclasses. 

Perhaps we could achieve something similar by creating a set of getter and setters around those 3 variables with the goal of splitting the game logic from the storage. e.g.:

```C++
void Monster::setLeoricMoveCounter(int counter) {
	goalVar1 = direction;
}

int Monster::getLeoricMoveCounter() {
	return goalVar1;
}

void Monster::setLeoricMoveDirection(int direction) {
	goalVar2 = direction;
}

int Monster::getLeoricMoveDirection() {
	return goalVar2;
}
```

Then this part of `LeoricAi()` could be rewritten from:

```C++
if (monster.goal != MGOAL_MOVE) {
	monster.goalVar1 = 0;
	monster.goalVar2 = GenerateRnd(2);
}
```

into

```c++
if (monster.goal != MGOAL_MOVE) {
	monster.setLeoricMoveCounter(0);
	monster.setLeoricMoveDirection(GenerateRnd(2));
}
```

AJenbo:
Maybe unions would be an option?

DakkJaniels:
I was thinking the same thing - but I wasn't sure if there would be any time when it would be set as one variable in the union and accessed as another... I was reading that some implementations of C++ may not like that? Although if they are all the same type size, maybe it doesn't matter?

aetn23:
What @canassa suggested would be problematic, since we would have to implement many many member functions in monster that would do similar things.

 The best solution would to be rewrite the whole monster to a nice OO, hierarchy, with abstract Monster class and particular monsters inheriting it - but that is not feasible. 

I think it would best to leave the logic of AIs to specific functions, since creating setters & getters for specific actions would result in a very big monster class. The logic itself is just too convoluted. 

I don't really see what others mean by using unions: split the `goalSpecialAction` to union of two vars? If that's it, it would introduce some ifs - but would make obvious what is being done in Ais.

aetn23:
I think I see what you mean. But wouldn't it mean we would have to implement a union of structs for every monster ai? And then every monster ai would just access its union, whenever it needs `goal...`? Or just a common denominator general struct and special structs within the union for monsters that do custom things? Introducing unions would also introduce manual management of them, and more complicated code. I think it's not worth it

StephenCWills:
> But wouldn't it mean we would have to implement a union of structs for every monster ai? And then every monster ai would just access its union, whenever it needs `goal...`?

Yes, it could very well mean that, assuming every AI uses these variables in a specific way that is distinct from how every other AI uses them.

> Or just a common denominator general struct and special structs within the union for monsters that do custom things?

Ideally, common usages could be grouped by a single struct, if such common usage can be clearly described. There's no sense duplicating the struct just to match the number of structs with the number of AI functions.

> Introducing unions would also introduce manual management of them, and more complicated code. I think it's not worth it

It's unclear to me what you mean by manual management of unions and how that is somehow a downside to this approach. It seemed like you were on board with the idea of an abstract `Monster` base class, which is very similar to this approach. Although, I also don't know why you said that it was not feasible.

The only thing that makes it more complicated is the fact that the same variable can have different names in different places in the code. It makes each location where it's used more understandable in isolation, but it's not immediately obvious that `monster.goalVars.leoric.moveCounter` refers to the same value as `monster.goalVars.common.goalVar1`. If you are, for instance, trying to figure out how Leoric's goal vars are captured in the save files, you can't just search for usages of `moveCounter` throughout the codebase.

Perhaps a simple compromise could be to use reference types to alias the variables in the functions themselves. However, this way you lose out on having descriptive variable names at the struct level so it's less clear what the variables are used for unless you explicitly search out their usages.

aetn23:
Thanks for clarification. By manual union management I meant that we would have to make sure we we reading and writing to active field each time we access the union. 

I described reworking monster structure as not feasible, since it would be a very big, paradigm shifting rework. If we want to do this, I am on board, since I believe that is the cleanest (and most radical) solution - but it would result in much clearer and manageable code.

I am now more convinced to the union approach - it would solve readability problem and is not as hard to do as OO rewrite.

heinermann:
The union approach is probably a simpler version of Starcraft's [here](https://github.com/bwapi/bwapi/blob/main/bwapi/BWAPI/Source/BW/CUnit.h#L152) (replica of official code with different naming). Unions are manually accessed based on unit type or class of unit types (in this case monster type or class of monsters). That code uses an anonymous union to bypass the extra reference. If you adopt the same mechanism then you could do i.e. `monster.leoric.moveCounter` but maybe you want to keep `goalVars`.

PR #4904, opened 2022-07-06
:memo: Fixes Doxygen warnings

canassa:
This fixes 200+ Doxygen warnings. There is only one left:

```
Generating docs for README.md:38: warning: unable to resolve reference to 'building-from-source' for \ref command
```

That's because Doxygen and GitHub are using different flavours of Markdown, so it can't be solved.

### DIFF
--- a/Source/engine.cpp
+++ b/Source/engine.cpp
@@ -163,23 +163,6 @@ void DrawHalfTransparentRectTo(const Surface &out, int sx, int sy, int width, in
 	DrawHalfTransparentBlendedRectTo(out, sx, sy, width, height);
 }
 
-/**
- * @brief Returns the direction a vector from p1(x1, y1) to p2(x2, y2) is pointing to.
- *
- *      W    SW     S
- *            ^
- *            |
- *     NW ----+---> SE
- *            |
- *            |
- *      N    NE     E
- *
- * @param x1 the x coordinate of p1
- * @param y1 the y coordinate of p1
- * @param x2 the x coordinate of p2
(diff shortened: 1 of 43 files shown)


Source/monster.cpp
```diff
@@ -774,6 +774,7 @@ void UpdateEnemy(Monster &monster)
 
 /**
  * @brief Make the AI wait a bit before thinking again
+ * @param monster The monster what will wait
```
AJenbo, line 777:
```suggestion
 * @param monster The monster that will wait
```

Source/engine.cpp
```diff
@@ -163,23 +163,6 @@ void DrawHalfTransparentRectTo(const Surface &out, int sx, int sy, int width, in
...
-/**
- * @brief Returns the direction a vector from p1(x1, y1) to p2(x2, y2) is pointing to.
- *
- *      W    SW     S
- *            ^
- *            |
- *     NW ----+---> SE
- *            |
- *            |
- *      N    NE     E
```
AJenbo, line 175: this should be moved to the header instead of being deleted

canassa: I removed it because the header already was documented:

```
/**
 * @brief Calculate the best fit direction between two points
 * @param start Tile coordinate
 * @param destination Tile coordinate
 * @return A value from the direction enum
 */
 ```

AJenbo: I guess that is actually a more accurate description, despite not having a nice graphics.

PR #4907, opened 2022-07-07
Eliminate CrawlTable

vlolteanu:
@ephphatha This PR should be right up your alley.

If merged, it makes https://github.com/diasurgical/devilutionX/pull/4855 obsolete.


Source/lighting.h
```diff
@@ -74,10 +74,113 @@ void ChangeVisionXY(int id, Point position);
...
+ *    |  314
+ *    | B7 8C
+ *    | F # G
+ *    | D9 AE
+ *    |  526
+ *    +-------> x
+ */
+
+template <typename F>
+auto Crawl(unsigned radius, F function) -> typename std::result_of<decltype(function)(Displacement)>::type
```
glebm, line 140: Consider implementing this as an iterator (e.g. input_iterator) rather than callback-style

Source/debug.cpp
```diff
@@ -723,28 +723,30 @@ std::string DebugCmdSpawnUniqueMonster(const string_view parameter)
...
+	auto ret = Crawl(0, MaxCrawlRadius, [&](auto displacement) -> std::optional<std::string> {
+		Point pos = myPlayer.position.tile + displacement;
+		if (dPlayer[pos.x][pos.y] != 0 || dMonster[pos.x][pos.y] != 0)
+			return {};
+		if (!IsTileWalkable(pos))
+			return {};
+
+		Monster *monster = AddMonster(pos, myPlayer._pdir, id, true);
+		if (monster == nullptr)
+			return fmt::format("I could only summon {} Monsters. The rest strike for shorter working hours.", spawnedMonster);
```
glebm, line 735: Conflict with #4895, let's see which gets merged first :sweat_smile:

glebm (approved):

Source/lighting.h
ephphatha, line 140:
```suggestion
auto Crawl(unsigned radius, F function) -> typename std::invoke_result_t<decltype(function), Displacement>
```

`std::result_of<F>` was deprecated in C++17 and removed in C++20, this is why the msvc builds are failing. Need to switch to `std::invoke_result_t<F, Args...>`. Also should `#include <type_traits>`. Beyond that I like the approach, decent enough given we can't use coroutine based ranges :(

ephphatha (approved):
Looks good :)

Source/lighting.h
```diff
@@ -74,10 +75,113 @@ void ChangeVisionXY(int id, Point position);
...
+ *    |  314
+ *    | B7 8C
+ *    | F # G
+ *    | D9 AE
+ *    |  526
+ *    +-------> x
+ */
+
+template <typename F>
+auto Crawl(unsigned radius, F function) -> invoke_result_t<decltype(function), Displacement>
```
AJenbo, line 141: Is there any reason this is unsigned if we are just going to cast it to `int` anyway?

ephphatha: it's cast to signed for use in a displacement that gets flipped and rotated, but the loops are expected to be always going from 0 (or a low positive value) to a larger value. Given the two contexts it's being used I think taking unsigned is appropriate here.

Source/lighting.h
```diff
@@ -74,10 +75,113 @@ void ChangeVisionXY(int id, Point position);
...
+ *    ^
+ *    |  314
+ *    | B7 8C
+ *    | F # G
+ *    | D9 AE
+ *    |  526
+ *    +-------> x
+ */
+
+template <typename F>
```
glebm, line 140: Perhaps we can avoid templates by making it accept a `FunctionRef<bool(Displacement)>` where bool indicates whether to abort or not. This will also result in less generated code. The return value stuff can be managed by the caller.

We'd need to add a `FunctionRef` class first (e.g. from https://github.com/TartanLlama/function_ref)

vlolteanu: We can, but this lends itself to a nastier API, i. e.:

glebm: We could have only a single templated function in the header, with the rest in the cpp file, e.g.:

```c++
void Crawl(unsigned minRadius, unsigned maxRadius, tl::function_ref<bool(Displacement)>);

template <typename T>
std::optional<T> Crawl(unsigned minRadius, unsigned maxRadius, tl::function_ref<std::optional<T>(Displacement)> fn) {
	std::optional<T> result
    Crawl(minRadius, maxRadius, [&result, &fn](Displacement displacement) {
      result = fn(displacement);
      return !result;
    });
  return result;
}

vlolteanu: I guess it's OK, but definitely not as part of this PR.

AJenbo (approved):

Source/lighting.h
glebm: I'll send a PR for this shortly

PR #4908, opened 2022-07-07
:sparkles: Introduces flipCoin()

canassa:
`GenerateRnd(2) != 0` and `GenerateRnd(2) == 0` were being used everwhere so I created a new function specifically for that.

The `flipCoin` name is debatable but I think that it shows a better intent than `randomBoolean`

### DIFF
--- a/Source/engine/random.cpp
+++ b/Source/engine/random.cpp
@@ -51,7 +51,7 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
-bool flipCoin()
+bool FlipCoin()
 {
 	return GenerateRnd(2) != 0;
 }
--- a/Source/engine/random.cpp
+++ b/Source/engine/random.cpp
@@ -51,4 +51,9 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
+bool flipCoin()
+{
+	return GenerateRnd(2) != 0;
+}
+
 } // namespace devilution
--- a/Source/levels/drlg_l1.cpp
+++ b/Source/levels/drlg_l1.cpp
@@ -954,10 +954,10 @@ void GenerateRoom(int x, int y, int w, int h, int dir)
 
 void FirstRoom()
 {
-	VerticalLayout = !flipCoin();
-	HasChamber1 = flipCoin();
-	HasChamber2 = flipCoin();
-	HasChamber3 = flipCoin();
+	VerticalLayout = !FlipCoin();
+	HasChamber1 = FlipCoin();
+	HasChamber2 = FlipCoin();
+	HasChamber3 = FlipCoin();
 
 	if (!HasChamber1 || !HasChamber3)
 		HasChamber2 = true;
@@ -1497,11 +1497,11 @@ Point SelectChamber()
 {
 	int chamber;
 	if (!HasChamber1)
-		chamber = flipCoin() ? 3 : 2;
+		chamber = FlipCoin() ? 3 : 2;
 	else if (!HasChamber2)
-		chamber = flipCoin() ? 1 : 3;
+		chamber = FlipCoin() ? 1 : 3;
 	else if (!HasChamber3)
-		chamber = flipCoin() ? 1 : 2;
+		chamber = FlipCoin() ? 1 : 2;
 	else
 		chamber = GenerateRnd(3) + 1;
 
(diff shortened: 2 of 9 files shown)


canassa:
Hmm, the tests broke because the existing code is using `GenerateRnd(2) != 0` and `GenerateRnd(2) == 0` interchangeably

Source/engine/random.cpp
```diff
@@ -51,4 +51,9 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
+bool flipCoin()
+{
+	return GenerateRnd(2) == 0;
```
AJenbo, line 56: There are 7 cases of this and 40 cases of the other variant, so change this to the common one and change the 7 other cases to use `!flipCoin()`

```suggestion
	return GenerateRnd(2) != 0;
```

AJenbo:
> I checked the tests and I think I could fix them, but let me know first if this change is welcome or not before I spend time on it. I am not sure if there are impacts on changing the level generation like that. With this change, a different dungeon will be generated for the same seed

The tests for levels and item generation is there to make sure that we stay compatible with the original game. It is critical that they do not change.

AJenbo:
> The `flipCoin` name is debatable but I think that it shows a better intent than `randomBoolean`

Fun fact I have Cuprolaminophobia :(

Source/engine/random.cpp
AJenbo: Alright just that one line fixed the 5 failing item generation tests :)

Source/objects.cpp
```diff
@@ -987,7 +987,7 @@ void DeleteObject(int oi, int i)
 
 void AddChest(int i, int t)
 {
-	if (GenerateRnd(2) == 0)
+	if (flipCoin())
```
AJenbo, line 990: I'm guessing this wasn't deliberate as it is the only one you didn't flip :)
```suggestion
	if (!flipCoin())
```

Source/engine/random.cpp
```diff
@@ -51,4 +51,9 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
+bool flipCoin()
```
AJenbo, line 54: I wasn't able to come up with a better name, but being a global the casing is wrong:
```suggestion
bool FlipCoin()
```

Source/levels/drlg_l1.cpp
```diff
@@ -954,10 +954,10 @@ void GenerateRoom(int x, int y, int w, int h, int dir)
...
 void FirstRoom()
 {
-	VerticalLayout = GenerateRnd(2) == 0;
-	HasChamber1 = (GenerateRnd(2) != 0);
-	HasChamber2 = (GenerateRnd(2) != 0);
-	HasChamber3 = (GenerateRnd(2) != 0);
+	VerticalLayout = !FlipCoin();
+	HasChamber1 = FlipCoin();
+	HasChamber2 = FlipCoin();
+	HasChamber3 = FlipCoin();
```
AJenbo, line 960: All though it seams strange that VerticalLayout uses an inverted check, I think you will agree that it's better then the original code :)

https://github.com/diasurgical/devilution/blob/master/Source/drlg_l1.cpp#L1486

Source/engine/random.cpp
```diff
@@ -51,4 +51,9 @@ int32_t GenerateRnd(int32_t v)
 	return AdvanceRndSeed() % v;
 }
 
+bool FlipCoin()
+{
+	return GenerateRnd(2) != 0;
+}
```
ephphatha, line 57: Structuring this check to be `== 0` is more natural, especially when dealing with frequencies other than 1 in 2. To fully address #2193 you could make this function take a frequency parameter with a default of 2. Then you can replace calls to GenerateRnd() like items.cpp:1084,1156,1157,1167,1224,1600, objects.cpp:504,538,542,546,550, etc.

Source/levels/drlg_l1.cpp
AJenbo: There we go: https://github.com/diasurgical/devilutionX/pull/4912

Source/engine/random.cpp
AJenbo: See https://github.com/diasurgical/devilutionX/pull/4908#discussion_r916160846 for why we flipped it. Not that we couldn't just add a lot of `!` instead.

ephphatha: Opened #4913 with the result of inverting the condition, can adjust most of the callers to suit :D

PR #4912, opened 2022-07-08
Clean up initial room generation

AJenbo:
`L4dungeon[80][80]`, `L5dungeon[80][80]`, and `dung[20][20]` have been combined in to `Bitset2d<40, 40>DungeonMask` as they all served to generate a boolean layout at the start of the level generation. `dung` was used to generate a quarter and then copy it to `L4dungeon`, instead this now happens directly in `DungeonMask` which sped things up a bit as well. The same was the case with `L5dungeon`, but here `dungeon` was used as the temp buffer. Previously this lead me to think that walls where being added at this stage, which was incorrect.

Both `L4dungeon` and `L5dungeon` where upscaled to 2x the dungeon and then down sampled with an offset to determiner what side of the wall the tile prepresents. But this would be done cleaner by simply probing the next tile when scanning.

Also reuse `DungeonMask` for DRLG3's tile lock map which seams to be used to similarly verify the size of the dungen.

In terms of data this should save 14600 bytes, but also generate less code.

### DIFF
--- a/Source/levels/drlg_l1.cpp
+++ b/Source/levels/drlg_l1.cpp
@@ -8,6 +8,7 @@
 #include "engine/load_file.hpp"
 #include "engine/point.hpp"
 #include "engine/random.hpp"
+#include "engine/rectangle.hpp"
 #include "levels/gendung.h"
 #include "player.h"
 #include "quests.h"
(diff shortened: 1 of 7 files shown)


glebm (approved):
Nice!

ephphatha (approved):
Now that I actually read this you already address what I was confused about with #4914 😂

AJenbo:
Thanks :)

> Now that I actually read this you already address what I was confused about with https://github.com/diasurgical/devilutionX/pull/4914 joy

Nice, that's a good win :)

PR #4913, opened 2022-07-08
Expand use of FlipCoin for unfair choices

ephphatha:
Opening as a draft till I finish going through the non-50/50 rolls.

### DIFF
--- a/Source/levels/themes.cpp
+++ b/Source/levels/themes.cpp
@@ -169,7 +169,7 @@ bool TFit_GoatShrine(int t)
 	return false;
 }
 
-bool CheckThemeObj3(Point origin, int8_t regionId, int frequency)
+bool CheckThemeObj3(Point origin, int8_t regionId, unsigned frequency)
 {
 	const PointsInRectangleRange searchArea { Rectangle { origin, 1 } };
 	return std::all_of(searchArea.cbegin(), searchArea.cend(), [regionId, frequency](Point testPosition) {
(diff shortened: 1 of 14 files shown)


Source/levels/themes.cpp
```diff
@@ -577,14 +574,26 @@ void Theme_MonstPit(int t)
 	PlaceThemeMonsts(t, monstrnd[leveltype - 1]);
 }
 
+namespace {
+void SpawnObjectOrSkeleton(unsigned frequency, _object_id objectType, Point tile)
```
AJenbo, line 578: Looks like `objectType` is alwasy `OBJ_BANNERL`

AJenbo:
Nice with all the improved variable names :)

Source/levels/themes.cpp
ephphatha: Shouldn't be, it's left right middle middle right left

AJenbo: Are you trying to use the Konami Code on me :eyes:

AJenbo (dismissed):

AJenbo:
There are still some cases like in GenerateRoom where this could be applied. I'll let you decide if you want to merge it as is or add more to it:
https://github.com/diasurgical/devilutionX/blob/master/Source/levels/drlg_l1.cpp#L898 (also in l4)

AJenbo:
I'm just going to merge this before we get any more conflicts :D You can always make a follow up for the remaining cases :)

glebm:
Getting this warning now:

```
Source/levels/themes.cpp:190:31: warning: comparison of integer expressions of different signedness: ‘const unsigned int’ and ‘int’ [-Wsign-compare]
  190 |                 if (frequency != -1 && FlipCoin(frequency)) {
      |                     ~~~~~~~~~~^~~~~
```

PR #4914, opened 2022-07-08
Attempt to document the behaviour of GenerateRoom through code

ephphatha:
Based on #4913 but split out because I wasn't sure how to express what it's trying to do clearly.


ephphatha:
Already addressed by #4912

PR #4915, opened 2022-07-08
Crawl: Move implementation from header to cpp

glebm:
As discussed in #4907


vlolteanu (approved):
Looks good to me.

Adding the extra dependency to avoid a few template instantiations is a bit overkill, though.

glebm:
`function_ref` is a pretty common class that we'll likely use elsewhere
it's also a standards proposal https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p0792r10.htm

PR #4919, opened 2022-07-08
Fix flash, town portal, and identify scroll descriptions

### LINKED ISSUE #4649: Flash, Town Portal and Identify scrolls have incorrect descriptions
**Important information**
Windows 10 / latest master

**Describe**
Flash scroll description says it needs a targeting click. It doesn't.
Town Portal description says it doesn't need a targeting click. It does.
Identify scroll descriptions says it doesn't need a targeting click. It does.

**Screenshots**
flash
identify
townportal

burningserenity:
Fixes #4649
Fix applies to existing items and new items.


AJenbo:
Won't this change how they are used?

burningserenity:
> Won't this change how they are used?

Not according to my test

AJenbo:
Ok I looked in to it and you are mostly right. There is one difference. When operating book stands it will only drop scrolls of type IMISC_SCROLL. This won't affect the two first scrolls since they are both IDROP_NEVER. But the consequence here is that you will no longer be able to find Identify or town portal but will start to find Flash instead. So we need to find a way to avoid this to apply this change.

burningserenity:
I should have read more closely :-D. I tried a different approach that checks for those spells specifically.

AJenbo:
Cold you please run clang-format to apply the correct code style, or do it manually :)

burningserenity:
> Cold you please run clang-format to apply the correct code style, or do it manually :)

Done

AJenbo (approved):
Despite the duplication here, I'm ok with the solution :)

PR #4925, opened 2022-07-09
cel_render: Remove unused `LineEndFn`

glebm:
This was used for stippled transparency which we no longer support.


AJenbo (approved):
:+1:

PR #4929, opened 2022-07-09
:recycle: Refactor M_WalkDir

canassa:
- :fire: Removed DumbWalk
- :truck: Renames M_WalkDir to Walk

This integrates the DirOK check into M_WalkDir itself, that makes the DumbWalk function useless and most places that were calling DirOK before M_WalkDir can now be simplified.

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1832,7 +1832,7 @@ bool RandomWalk(int monsterId, Direction md)
 		    || (md = Right(Right(mdtemp)), DirOK(monsterId, md));
 	}
 	if (ok)
-		M_WalkDir(monster, md);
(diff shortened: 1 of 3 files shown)


Source/monster.cpp
```diff
@@ -1832,7 +1832,7 @@ bool RandomWalk(int monsterId, Direction md)
 		    || (md = Right(Right(mdtemp)), DirOK(monsterId, md));
 	}
 	if (ok)
-		M_WalkDir(monster, md);
+		Walk(monster, md);
```
canassa, line 1835: RandomWalk and RandomWalk2 can be simplified. But I will do that on another PR

AJenbo (approved):

PR #4934, opened 2022-07-10
Fix FillChambers

AJenbo:
@glebm noticed that some Cathedral maps where a bit broken:

image

This PR fixes the issue (which was introduced by me in by not considering that single chamber levels where possible)

And also adds a test for this level seed to prevent regressions going forward.

image
(the floating fire is caused by the slightly different layout of the fixed level but the fire being objects in the save file)

Since I also ended up understanding the code a bit better I also ended up cleaning up `FillChambers()` to a point where I'm not satisfied with most of the code in DRLG_L1 :)

Note: I removed some nesting so about 50% of this PR is just indentation, probably ignore whitespace when viewing the diff.


Source/levels/drlg_l1.cpp
```diff
@@ -1003,44 +986,31 @@ void Substitution()
...
+		std::swap(hall1.x, hall1.y);
+		std::swap(hall2.x, hall2.y);
+	}
 
-		if (!HasChamber3)
-			GenerateChamber({ 14, 14 }, false, false, true, false);
-		else if (!HasChamber1)
-			GenerateChamber({ 14, 14 }, false, false, false, true);
-		else if (HasChamber1 && HasChamber2 && HasChamber3)
-			GenerateChamber({ 14, 14 }, false, false, true, true);
```
AJenbo, line 1015: The bug was here, i was always connecting chamber 2 to at least one other chamber. But instead of the multiple if's we can simply use the variable as the paramter which makes things a lot cleaner. At that point it also became feasible to merge vertical/horizontal layout and just pass on the layout direction instead of alternating between parameters 1&2 and 3&4 for the chambers.

Source/levels/drlg_l1.cpp
```diff
@@ -1003,44 +986,31 @@ void Substitution()
 
 void FillChambers()
 {
+	Point chamber1 { 14, 0 };
+	Point chamber3 { 14, 28 };
+	Point hall1 { 18, 12 };
+	Point hall2 { 18, 26 };
 	if (!VerticalLayout) {
```
ephphatha, line 993: Pretty minor but you can structure these checks as positive by making the horizontal layout the default.

```suggestion
	Point chamber1 { 0, 14 };
	Point chamber3 { 28, 14 };
	Point hall1 { 12, 18 };
	Point hall2 { 26, 18 };
	if (VerticalLayout) {
```

Same could be done in FirstRoom with a bit more care.

AJenbo: Sure why not

ephphatha (approved):

AJenbo:
Thanks :)

PR #4937, opened 2022-07-10
Use absolute coordinates for temp position when north-walking

StephenCWills:
This resolves an issue where position desync when walking north would cause the player's position to warp out of bounds, leading to disconnects in multiplayer games. This also fixes some errors in loadsave.cpp related to the type change from `Point` to `PointOf<WorldTileCoord>`.


AJenbo (approved):
Nice to not have the runtime be esoteric :)

AJenbo:
GitHub's buttons topped working again :face_with_spiral_eyes:

PR #4938, opened 2022-07-10
Take player by references

AJenbo:
With https://github.com/diasurgical/devilutionX/commit/1fb45a7f2548663294fe6e6996963033e6c70407 solved it unlocked a few more methods to take the player by reference rather then as an ID.

It looks like the rest are pretty much tied up in needing the player ID, we can solve that the same way we did for monsters, but I think this is a good place to cut this PR. I also changed comparisons from `playerId == MyPlayerId` when the function already had the player pointer since this will also simplify https://github.com/diasurgical/devilutionX/pull/4754

### DIFF
--- a/Source/cursor.cpp
+++ b/Source/cursor.cpp
@@ -540,14 +540,16 @@ void CheckCursMove()
 	if (pcursmonst == -1) {
 		if (!flipflag && mx + 1 < MAXDUNX && dPlayer[mx + 1][my] != 0) {
 			int8_t bv = abs(dPlayer[mx + 1][my]) - 1;
-			if (bv != MyPlayerId && Players[bv]._pHitPoints != 0) {
+			Player &player = Players[bv];
+			if (&player != MyPlayer && player._pHitPoints != 0) {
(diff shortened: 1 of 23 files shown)


Source/cursor.cpp
```diff
@@ -540,14 +540,16 @@ void CheckCursMove()
 	if (pcursmonst == -1) {
 		if (!flipflag && mx + 1 < MAXDUNX && dPlayer[mx + 1][my] != 0) {
 			int8_t bv = abs(dPlayer[mx + 1][my]) - 1;
-			if (bv != MyPlayerId && Players[bv]._pHitPoints != 0) {
+			Player &player = Players[bv];
```
glebm, line 543: This is going to cause conflicts because in #4754 I've renamed `bv` to `playerId`

glebm (approved):

Source/cursor.cpp
AJenbo: Yeah, it will be the second one here, defiantly the most painful conflicts, but ok handling it :)

PR #4941, opened 2022-07-10
Remove `MonsterData#has_trans`

glebm:
1. Removes the field because it is redundant, checks for `TransFile != nullptr` instead.
2. Differentiate between Skullwing (weaker) and Bone Demon (stronger) with a TRN

Skelt.trn (weak) | white.trn | no TRN (strong)
---|---|---
image | image | image


qndel (approved):

AJenbo:
I only learned recently that there is even a difference between the two monsters :)

PR #4944, opened 2022-07-10
Tests CMake: Use platform flags

glebm:


glebm:
Nevermind, this should not be necessary. libdevilutionx_so already has these definitions and options as PUBLIC

PR #4947, opened 2022-07-10
Replace AppVeyor with GitHub Actions

### LINKED ISSUE #4945: Use GitHub Actions instead of AppVeyor
Windows seems to be available on GitHub Actions now

https://github.blog/changelog/2021-11-16-github-actions-windows-server-2022-with-visual-studio-2022-is-now-generally-available-on-github-hosted-runners/

glebm:
Fixes #4945

1. We build in Debug mode (same as we did on AppVeyor).
2. We now also run the tests. We previously did not run any tests on Windows.
3. We cache the vcpkg output using [run-vcpkg](https://github.com/marketplace/actions/run-vcpkg)
4. We use [run-cmake](https://github.com/marketplace/actions/run-cmake) actions together with [CMake Presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).
   This has a very cool feature of reporting build warnings in the action summary and on the changed lines in the PR.
5. We use ccache to greatly speed up the build.

@AJenbo You'll need to disable AppVeyor integration in the repo settings / AppVeyor UI as well


StephenCWills:
CMakePresets.json may interfere with CMakeSettings.json in Visual Studio.
https://docs.microsoft.com/en-us/cpp/build/cmake-presets-vs?view=msvc-170

glebm:
Looks like Visual Studio ignores `CMakePresets.json` by default:

> `CMakePresets.json` integration isn't enabled by default in Visual Studio.

`CMakePresets.json` is the only way to use run-cmake

StephenCWills (approved):

AJenbo:
It still took a combined 25 minutes to build :D Still this comes with some nice QoL improvements and I think there is less of a waiting line, the two builds might even run in parallel? And hopefully the cache will kick in for subsequent builds. Nice work :+1:

AJenbo:
After a couple of builds it still takes 13min for MSVC and 11 for XBox One. Still with builds now running in parallel this is a huge improvement.

glebm:
Yeah I'll need to adjust the ccache setup

PR #4949, opened 2022-07-10
Fix monster goal variable references

StephenCWills:
Resolves an issue reported by FireIceTalon on Discord where Advocates are almost completely invulnerable to attacks. The error in `Monster::isPossibleToHit()` tipped me off to the possibility that there might have been more goal variable mistakes. I double-checked all the goal variable renames in #4861 and fixed quite a number of other issues as well.


AJenbo (approved):
Yikes, this looks like it was tedious to sort out.

PR #4952, opened 2022-07-10
Convert monster_goal in to a scoped enum

AJenbo:
As a follow up to https://github.com/diasurgical/devilutionX/pull/4949 to make sure that there aren't any cases where this is still broken and prevent it from happening in the future. Also it found a few cases in HorkAI where the enum hadn't been applied.


StephenCWills (approved):

AJenbo:
@aetn23 pinging you in case you might find this interesting :)

aetn23:
Nice! Much more readable. I think there are many enums like old `monster_goal` still in the code - could make for some nice PRs.

Issue #4957, opened 2022-07-11
Incorrect room size for Catacombs


AJenbo:
**Important information**
7bb401d9e8b0cca0f8ace40243fc9560eeb2282d changed how rooms in the catacombs are generated :sweat_smile: 

**To Reproduce**
Load [single_3.zip](https://github.com/diasurgical/devilutionX/files/9080298/single_3.zip)
See misplaced door tile:
![image](https://user-images.githubusercontent.com/204594/178172387-bb8f825a-1621-49ef-a4b0-1402d1e580c1.png)

**Additional context**
A test case for this issue https://github.com/diasurgical/devilutionX/pull/4958

PR #4958, opened 2022-07-11
Add test for broken Catacombs generatrion

AJenbo:
A failing test to go with https://github.com/diasurgical/devilutionX/issues/4957 to make it easier to work with :)


ephphatha (dismissed):
Helpful for this annoying corner case :D

PR #4959, opened 2022-07-11
Fix incorrect size determination in GetSizeForThemeRoom

### LINKED ISSUE #4957: Incorrect room size for Catacombs

ephphatha:
This was a corner case where a boundary was determined by the edge of the search region. Not 100% sure it's correct in all cases, aiming to add some tests to try cover every possible case.

Fixes #4957 and doesn't break existing tests however the cases in the last loop could be expressed better for sure.

### DIFF
--- a/Source/levels/gendung.cpp
+++ b/Source/levels/gendung.cpp
@@ -123,7 +123,7 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
 		}
 	}
 
-	// Worth out the tallest area we could potentially fill
+	// Work out the tallest area we could potentially fill
 	for (int xOffset = 0; xOffset < minSize; xOffset++) {
 		for (int yOffset = minSize; yOffset < maxHeight; yOffset++) {
 			if (dungeon[origin.x + xOffset][origin.y + yOffset] == floor)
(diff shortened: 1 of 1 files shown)


Source/levels/gendung.cpp
```diff
@@ -123,27 +123,26 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
 		}
 	}
 
-	// area is at least as high as necessary. If we wanted to find the largest rectangular area we should keep going
-	// and start checking for the largest total area we can find (could be the tallest region that maintains current
-	// width, or maybe a narrower but taller region has a larger area). Instead to match vanilla Diablo logic we
-	// trigger a break as soon as the width shrinks again.
-	for (int yOffset = minSize; yOffset < maxHeight; yOffset++) {
-		for (int xOffset = 0; xOffset < maxWidth; xOffset++) {
+	// Worth out the tallest area we could potentially fill
```
AJenbo, line 126: ?
```suggestion
	// Work out the tallest area we could potentially fill
```

Issue #4742, opened 2022-06-19
Vector-like container backed by array


vlolteanu:
The code base is full of stuff like:
```
Foo Bar[MAX_BAR]
int BarSize; 
```
Simply put, this is a poor man's container. It also makes it awkward to place any code with side-effects into `Foo`s constructor and/or destructor. (E.g. imagine calling `InitMonsterGFX` when constructing a `CMonster`.)

The obvious solution would be to use something similar to `boost::static_vector` (https://www.boost.org/doc/libs/1_57_0/doc/html/boost/container/static_vector.html).

In previous discussions, @AJenbo seemed turned off by the idea of pulling in boost/container as a dependency, but I've seen other implementations floating around the Internet. We could also roll our own.

ephphatha:
https://github.com/diasurgical/devilutionX/pull/2951 was the initial implementation I was working on. Still haven't got around to fleshing it out with a use case since there's plenty of cleanups to do for object/item usage before it can be dropped in.

vlolteanu:
I have a stupid implementation that piggybacks on `std::vector` to provide all the kosher C++ container stuff.

```
template <class T, size_t N>
class StaticAllocator
{
	char mem[sizeof(T) * N];

public:
	typedef T value_type;

	T *allocate(std::size_t n) {
		if (n > N)
			throw std::bad_array_new_length();
		return reinterpret_cast<T *>(mem);
	}

	void deallocate(T *, std::size_t) noexcept {
	}

	template <class U>
	struct rebind
	{
		using other = StaticAllocator<U, N>;
	};
};

template <class T, size_t N>
struct StaticVectorReserver
{
	StaticVectorReserver();
};

template <typename T, size_t N>
class StaticVector : public std::vector<T, StaticAllocator<T, N>>, StaticVectorReserver<T, N> {
	using std::vector<T, StaticAllocator<T, N>>::vector;
};

template<class T, size_t N>
StaticVectorReserver<T, N>::StaticVectorReserver()
{
	((StaticVector<T, N> *)this)->reserve(N);
}
```

glebm:
We already have https://github.com/diasurgical/devilutionX/blob/master/Source/utils/static_vector.hpp

vlolteanu:
> We already have https://github.com/diasurgical/devilutionX/blob/master/Source/utils/static_vector.hpp
> 
> It's a bit barebones but could be expanded

This reminds me of https://knowyourmeme.com/memes/we-have-food-at-home

[...]

PR #4964, opened 2022-07-11
Object oriented Monsters 

aetn23:
Begin changing monster hierarchy to be OO. Those are very much first steps - as such no ground breaking changes have been made. This is done to experiment with OO monster design. I believe fully OO monsters would be easier to maintain and to understand.

I make this PR as a draft this early, since I would like to know if this even makes any kind of sense - there is much work to be done and I wouldn't like to do it without being sure it is even wanted in the first place. The code proposed is very very WIP and as such it does not make a lot of sense.

Here is a quick list of changes in the code:
- Changed Monsters array to hold ptr to monster instead of monster
- Added default ctor to monster
- Added ctor to Monster that is a copy of InitMonster with initialization of fields that Init function did not touch
- Remove adding monster through InitMonsters, monsters are constructed via Monster ctor and added to Monsters array. Changed monster adding accordingly. As of now, when new monster is added, memory for it is allocated via new, and the memory for old monster in the Monsters array is deleted.

I am sure there are some bugs - my goal was to quickly hack some code to spark a discussion about OO monsters: mostly if the idea makes any sense.


glebm:
This kind of memory management (allocating each monster via new/delete) is a lot less efficient than the current approach. We have DevilutionX ported to some very constrained platforms that can barely manage 20 FPS as it is.

If we do decide to change how this is done, the implementation should still be efficient under the hood.
For previous discussion, see https://github.com/diasurgical/devilutionX/issues/4742

glebm:
An array of pointers means an extra indirection. If you're using placement new, you shouldn't need pointers.

The diff is also far too large to review because of all the `.` to `->` changes.

See https://github.com/diasurgical/devilutionX/blob/master/Source/utils/static_vector.hpp for an example of how to implement a direct array via `std::aligned_storage`.

At this point, it's hard to tell whether this particular approach is the one we want, so please don't go through too much effort on this.

aetn23:
The idea behind using array of pointers for `Monsters` is to later add each monster as a class inheriting base `Monster` and store them as a pointers to a base class

glebm:
Ah, inheritance, virtual functions, etc. Inheritance is usually not a good fit for complex software, especially games.

AJenbo:
maybe something can be done using composition?

aetn23:
> maybe something can be done using composition?

I thought about putting ai stuff into a separate struct. That would be mostly cosmetic, but would improve readability (it would be clear what monsters uses for ai). I also thought about maybe storing the ai function inside this struct, but it would increase the size of the struct. 

What other things regarding composition do you have on mind?

AJenbo:
Not much it was a loos thought :)

One ugly aspect is that in town towners are seen as monsters but must be looked up in a different table.

glebm:
> I thought about putting ai stuff into a separate struct. That would be mostly cosmetic, but would improve readability (it would be clear what monsters uses for ai). 

How will putting the AI stuff in a separate struct be more clear than having it in the same struct?

> I also thought about maybe storing the ai function inside this struct, but it would increase the size of the struct.

Let's not increase the struct size unnecessarily.

fwannmacher:
Polymorphism is not that good when thinking about performance. It is a good approach in theory but when we do have limited process power it will cause more issues than benefits.
Another thing is that some devices won't handle allocating memory outside the heap well. So using new will be very slow.

PR #4966, opened 2022-07-11
Clean up color cycling

AJenbo:


Source/engine/palette.cpp
```diff
@@ -360,44 +360,41 @@ void palette_update_caves()
...
 	}
+
+	CycleColorsReverse(16, 31);
+	palette_update(0, 31);
+	delayLava = !delayLava;
 }
 
+/**
+ * @brief Cycle the pond waves and bubles colors every 3rd frame
+ * Bubles have 8 colors and waves 7, so the full animation has 56th frames before it loops
```
AJenbo, line 383:
```suggestion
 * Bubles have 8 colors and waves 7, so the full animation has 56 frames before it loops
```

Source/engine/palette.cpp
```diff
@@ -360,44 +360,41 @@ void palette_update_caves()
...
 
+/**
+ * @brief Cycle the lava every other frame, and glow every frame
+ * Lava has 15 colors and the glow 16, so the full animation has 240 frames before it loops
+ */
 void palette_update_crypt()
 {
-	static int laveDelay = 0;
-	static int glowDelay = 0;
+	static bool delayLava = true;
```
AJenbo, line 369:
```suggestion
	static bool delayLava = false;
```

Source/engine/palette.cpp
```diff
@@ -360,44 +360,41 @@ void palette_update_caves()
...
-	} else {
-		laveDelay++;
-	}
-	if (glowDelay > 0) {
-		CycleColorsReverse(16, 31);
-		palette_update(0, 31);
-		glowDelay++;
-	} else {
-		glowDelay = 1;
+		delayLava = 0;
```
obligaron, line 373:
```suggestion
		delayLava = false;
```

Source/engine/palette.cpp
```diff
@@ -360,44 +360,41 @@ void palette_update_caves()
...
+	static bool delayLava = false;
 
-	if (laveDelay > 1) {
+	if (!delayLava) {
 		CycleColorsReverse(1, 15);
-		laveDelay = 0;
-	} else {
-		laveDelay++;
-	}
-	if (glowDelay > 0) {
```
obligaron, line 374: I think this doesn't matter but to be safe:
Removing this is not 100% the same.
In this pr glow will always cycle. In this master this is not always the case
- In the first run `glowDelay` is 0 and we don't cycle.
- `glowDelay` is never rest and that means after some time int overflows and we get back to a `glowDelay` of 0 that don't cycle.

obligaron (commented):
lgtm, only one nit 🙂

Source/engine/palette.cpp
AJenbo: Considering that the stutter only happens every 500 day of being in crypts with out restarting Diablo I consider that more of a bug then anything :)

PR #4971, opened 2022-07-11
:recycle: Use std::optional instead of nullptr on AddMonster

canassa:
I recently refactored AddMoster (#4894) so that it returns a pointer instead of a Monster id. AddMonster might fail to create a monster and returns a `nullptr` in that case.

The problem with that approach is that everyone who calls `AddMonster` should now be aware of null point references. Being a newbie in C++ myself, I am more used to type checkers that enforce the caller to check if the returned value is a Monster or not.  This code for example is 100% valid:

```c++
Monster *monster = AddMonster(pos, myPlayer._pdir, id, true);
// Forgets to check if monster is a nullprt
monster->getId();  // No compiler error or warning
```

 So I did a little bit of research and found `std::optional` which seems to solve the issue. It will be explicit to the caller that the return is optional and they will have to check it and unwrap it.
 
 Let me know what you think 😄


Source/monster.cpp
```diff
@@ -1719,7 +1720,7 @@ void MonsterPetrified(Monster &monster)
 	}
 }
 
-Monster *AddSkeleton(Point position, Direction dir, bool inMap)
+std::optional<Monster *> AddSkeleton(Point position, Direction dir, bool inMap)
```
AJenbo, line 1723: It looks like you are both returning an optional and using a pointer, instead you should return an optional reference.

AJenbo:
On a personal note I have a very hard time seeing how writing `.value().` instead of `->` prevents code errors.

Source/monster.cpp
canassa: I tried, but the compiler doesn't like it 😞 

https://stackoverflow.com/a/40383290/360829

Source/objects.cpp
```diff
@@ -1128,9 +1128,9 @@ void AddBarrel(Object &barrel)
...
 	if (barrel._oVar2 >= 8) {
-		Monster *skeleton = PreSpawnSkeleton();
-		if (skeleton != nullptr) {
-			barrel._oVar4 = skeleton->getId();
+		auto maybeSkeleton = PreSpawnSkeleton();
+		if (maybeSkeleton) {
+			barrel._oVar4 = maybeSkeleton.value()->getId();
 		} else {
 			barrel._oVar4 = -1;
 		}
```
aetn23, line 1136: What about `barrel._oVar4 = maybeSkeleton.value_or(-1);`?

glebm (changes requested):
You can't put a reference into an optional (unless you use `std::reference_wrapper`) and it's redundant anyway.
When it can be null, use a pointer, otherwise use a reference!

`std::optional` adds a bit of overhead in C++ as it carries around a boolean.

AJenbo:
Well if we are just returning a pointer in side an optional I don't think we have gained anything, that pointer might still be a null pointer.

@glebm do you claim to understand C++?

Source/objects.cpp
AJenbo: If I'm understanding you correctly here this would also require `SpawnSkeleton()` be changed to take an index intead of a pointer.

canassa:
> On a personal note I have a very hard time seeing how writing `.value().` instead of `->` prevents code errors.

Yeah, it's still not ideal. What I would like is that the compiler forced the caller to check if the value is valid or not, someone is still able to use .value() without checking if the value is there or not.

One benefit of it is that it works as live documentation, it says, at the type level, that the variable might contain not the desired value. `nullptr` is transparent.

But I am not sure how useful that is 😄

AJenbo:
> When a programmer sees that a function returns an optional they knows the function may contain nothing. So calling a `.value()` should indicate to the programmer that they should handle the case when optional does not hold a value.

When a programmer sees that a function returns a pointer they knows the function may return nothing. So calling `->` should indicate to the programmer that they should handle the case when the pointer is null :shrug:

aetn23:
> Well if we are just returning a pointer in side an optional I don't think we have gained anything, that pointer might still be a null pointer.

I don't see how this optional may hold null pointer, since the `AddMonster` returns optional with a value only if it managed to init a monster. 

Beside, with `value_or` (as in my comment somewhere up) we may end up simplyfing many `if ... else` blocks by simply doing `var = value.value_or(fallbackValue)`

aetn23:
> > When a programmer sees that a function returns an optional they knows the function may contain nothing. So calling a `.value()` should indicate to the programmer that they should handle the case when optional does not hold a value.
> 
> When a programmer sees that a function returns a pointer they knows the function may return nothing. So calling `->` should indicate to the programmer that they should handle the case when the pointer is `nullprt` shrug
> 
> > One benefit of it is that it works as live documentation, it says, at the type level, that the variable might contain not the desired value. `nullptr` is transparent.
> 
> I don't see how, `*` is just as clear an indication that you need to check for `nullprt` as `std::optional<>` is that you need to check the value.

Not necessarily, since some functions always return non null pointers

aetn23:
Yes, you are right. I forgot about that.

As such the only benefit would be `value_or` - but that can be done by a simple utility function, without `std::optional` overhead.

Still, maybe other functions that use place holder values (like -1) to indicate a failure could benefit from using `std::optional`?

canassa:
> @aetn23 Not sure if you saw my comment above. If you always return a non-null pointer, return a reference instead. Assume that a pointer can be null.

That's clear. If we always assume that pointers can be null then indeed this `std::optional` is not providing much value.

I still wonder if there is a way to detect potential nullptr problems around the code, which was my initial goal with the PR 🤔

PR #4973, opened 2022-07-11
Fix progress bar on already-visited levels

### LINKED ISSUE #4970: The progress bar doesn't complete when you load an already-visited level
**Important information**
Windows 10 / latest master / download 

**Describe**
The progress bar doesn't complete when you load an already-visited level. Seems like a fairly recent regression, don't recall it happening two weeks ago.

glebm:
Fixes #4970

### DIFF
--- a/Source/interfac.cpp
+++ b/Source/interfac.cpp
@@ -31,6 +31,8 @@ namespace devilution {
 
 namespace {
 
+constexpr uint32_t MaxProgress = 534;
+
 OptionalOwnedCelSprite sgpBackCel;
 
 bool IsProgress;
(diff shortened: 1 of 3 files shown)


Source/interfac.cpp
```diff
@@ -216,16 +218,23 @@ void interface_msg_pump()
...
 	DrawCutsceneForeground();
-	return sgdwProgress >= 534;
+}
+
+void CompleteProgress()
+{
+	if (!IsProgress)
+		return;
+	while (sgdwProgress < MaxProgress)
+		IncProgress();
```
AJenbo, line 237: Bummer I had previously managed to align the number of steps so the progressbar would be at the same stage at the same position for both code paths. I guess we got rid of some useless stuff since then :)

AJenbo (approved):

PR #4976, opened 2022-07-11
♻️ Make Monster counters `size_t`

glebm:


Source/sync.cpp
```diff
@@ -232,7 +232,7 @@ bool IsTSyncMonsterValidate(const TSyncMonster &monsterSync)
 {
 	const int monsterId = monsterSync._mndx;
 
-	if (monsterId < 0 || monsterId >= MaxMonsters)
+	if (static_cast<size_t>(monsterId) >= MaxMonsters)
```
AJenbo, line 235: `_mndx` is already unsinged (`uint8_t`), probably we should just change `monsterId` to `size_t`?

glebm: Done

Source/player.cpp
```diff
@@ -789,7 +789,7 @@ bool PlrHitMonst(int pnum, int monsterId, bool adjacentDamage = false)
 {
 	int hper = 0;
 
-	if (monsterId < 0 || monsterId >= MaxMonsters) {
+	if (static_cast<size_t>(monsterId) >= MaxMonsters) {
```
AJenbo, line 792: All calls to `PlrHitMonst()` verifies that `monsterId` is positive, so we can just make it size_t

Source/monster.cpp
```diff
@@ -348,7 +348,7 @@ void PlaceGroup(int mtype, int num, UniqueMonsterPack uniqueMonsterPack, int lea
 		int y1 = yp;
 
 		if (num + ActiveMonsterCount > totalmonsters) {
-			num = totalmonsters - ActiveMonsterCount;
+			num = static_cast<int>(totalmonsters - ActiveMonsterCount);
```
AJenbo, line 351: `num`, `totalmonsters`, and `placed` can all be unsigned types

AJenbo:
Looks reasonable :)

Source/monster.cpp
glebm: Done

Source/player.cpp
glebm: Done

AJenbo: Didn't expect that you would fully remove the check, but on further inspection it looks safe to do that as well :sweat_smile:

AJenbo (approved):

PR #4905, opened 2022-07-06
:recycle: Refactor RandomWalk to receive a Monster reference

canassa:
It's still using the monsterId internally since DirOK is waiting for #4817

PR #4978, opened 2022-07-12
:recycle: Refactor RandomWalk

canassa:
:sparkles: ~~Adds the Shuffle helper function~~
:sparkles: Adds WalkAny
:recycle: Changes RandomWalk to receive a monster reference (closes #4905)
:memo: Minor documentation fixes
:fire: Merges RandomWalk2 into RandomWalk

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1173,7 +1173,7 @@ void StartFadeout(Monster &monster, Direction md, bool backwards)
 /**
  * @brief Starts the monster healing procedure.
  *
- * The monster will be healed between 1.47% and 25% of its max HP. The healing amount is stored in _mVar1.
+ * The monster will be healed between 1.47% and 25% of its max HP. The healing amount is stored in var1.
  *
  * This is only used by Gargoyles.
  *
@@ -1560,7 +1560,6 @@ bool MonsterFadeout(Monster &monster)
  * This is triggered by StartHeal()
  *
  * @param monster The monster that will be healed.
- * @return
  */
 void MonsterHeal(Monster &monster)
 {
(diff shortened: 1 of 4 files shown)


Source/monster.cpp
```diff
@@ -1803,50 +1802,53 @@ void GroupUnity(Monster &monster)
...
+/**
+ * @brief Walks the monster in the direction md.
+ *
+ * Take the following monster for example:
+ *
+ * @code{.unparsed}
+ * ↖️ ⬆️ ↗️
+ * ⬅️ 😈 ➡️
+ * ↙️ ⬇️ ↘️
+ * @endcode
```
canassa, line 1814: GitHub is not rendering the emojis correctly for me, but it looks correct in both my editor (CLion) and in Doxygen:

image

canassa:
After submitting this PR I came up with this version that doesn't require the `Shuffle`:

```c++
bool RandomWalk2(Monster &monster, Direction md, bool narrow = false)
{
	// Tries to walk in the direction md.
	if (Walk(monster, md))
		return true;

	bool ok = false;
	// Tries the next adjacent directions.
	if (FlipCoin()) {
		ok = WalkAny(monster, { Left(md), Right(md) });
	} else {
		ok = WalkAny(monster, { Right(md), Left(md) });
	}

	// Tries the next adjacent directions.
	if (!ok && !narrow) {
		if (FlipCoin()) {
			ok = WalkAny(monster, { Left(Left(md)), Right(Right(md)) });
		} else {
			ok = WalkAny(monster, { Right(Right(md)), Left(Left(md)) });
		}
	}

	return false;
}
```

It's only slightly longer than the Shuffle version and it is still more readable than the original (I think). What do you guys think?

ephphatha:
Both uses of shuffle in this pr are on a list of two choices, it's not enough of an advantage to make up for the readability issues so I think your proposed version in the comment above is better.

canassa:
> Both uses of shuffle in this pr are on a list of two choices, it's not enough of an advantage to make up for the readability issues so I think your proposed version in the comment above is better.

I totally agree with you, unless the Shuffle could be used in other parts of the codebase I don't think this justifies adding it. I will update the MR with the second version

PR #4986, opened 2022-07-13
Remove monster name member from monster struct

aetn23:
Remove monster name member from monster struct. It is possible to fetch the name from `MonsterData` (or unique variant) global, using info that is already present in the struct.

This is done to save few bytes & improve readability (one less field to worry about)


Source/monster.h
```diff
@@ -280,6 +279,18 @@ struct Monster { // note: missing field _mAFNum
...
+	/**
+	 * @brief Returns monster's name
+	 * @return Monster's name
+	 */
+	const char *name() const
+	{
+		if (uniqType != 0)
+			return UniqueMonstersData[uniqType - 1].mName;
+
+		return data().mName;
```
AJenbo, line 291: you forgot the translation support:

```cpp
	if (monster.uniqType != 0)
		monster.name = pgettext("monster", UniqueMonstersData[monster.uniqType - 1].mName).data();
	else
		monster.name = pgettext("monster", monster.data().mName).data();
```

aetn23: Whoops, fixed.

Source/monster.h
```diff
@@ -285,6 +285,19 @@ struct Monster { // note: missing field _mAFNum
 		return *type().data;
 	}
 
+	/**
+	 * @brief Returns monster's name
+	 * Internally it returns a name stored in global array of monsters' data.
+	 * @return Monster's name
+	 */
+	const char *name() const
```
glebm, line 293: Return `string_view` here and remove `.data()` calls below and the explicit conversion to `string_view` elsewhere

Source/control.cpp
```diff
@@ -917,7 +917,7 @@ void DrawInfoBox(const Surface &out)
 			if (leveltype != DTYPE_TOWN) {
 				const auto &monster = Monsters[pcursmonst];
 				InfoColor = UiFlags::ColorWhite;
-				InfoString = string_view(monster.name);
+				InfoString = string_view(monster.name());
```
AJenbo, line 920:
```suggestion
				InfoString = monster.name();
```

glebm (approved):

glebm:
Yeah that's a pretty decent reduction. There can be more names displayed at a time though (item labels), so item names should probably be stored.

aetn23:
Also, much information in monster struct is basically a copy of info from `MonsterData` (members like `exp`, `level`) are written once, during `monster` initialization. Their values are taken directly from `MonsterData`. Sure, there are modified based on difficulty, but it happens once. What if we do the same to those fields, as this pr done to `name` - remove them, and fetch them from the array? The array would have to be initialized with proper values depending on game difficulty during initialization - but that would happen only once per run, during game loading. That move would remove many members from `monster`

AJenbo:
> But if we try to do the same for items, namely fetch the name from global array, don't we run into the problem with an ear item, that does not have predefined name, since it carries a player name?
> 
> Aside from that, why store it in the struct, if we could fetch it from global array?

Item names are generated, there can be 100s of 1000s of possible combination. So holding the up to 150 individual ones for the current level is probably more sensible, unless you would want to get creative but I don't think that is worth it.

AJenbo:
> The array would have to be initialized with proper values depending on game difficulty during initialization - but that would happen only once per run, during game loading. That move would remove many members from `monster`

Much of it could probably be moved to `LevelMonsterTypes`? The biggest issue is probably how to handle unique monsters.

aetn23:
I have counted how many members could be removed and the number is 5: `exp, toHitSpecial, level, minDmgSpecial, maxDmgSpecial`. It turns out that many of `monster` members are runtime depended: for example `PlaceGroup` changes some members, as well as spawning golems does. Also `WeakenNaKrul` changes some things. 

Let me know if removing those 5 members is worth the effort. 

Still, initializing `MonsterData` once, depending on difficulty, and then only copying the the fields while initializing monsters is better than what is done right now (do calculations for difficulty every time a monster is being initialized)

glebm:
Removing these 5 fields from `Monster` sounds good!

PR #4987, opened 2022-07-13
Use cursor position for position based network messages targeting objects

ephphatha:
pcursobj was used as a pointer so might as well make it an actual pointer. This also allows cleaning up some of the network messages to work off a position instead of ignoring the passed info and relying on index instead

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1215,29 +1215,18 @@ size_t OnTargetSpellTile(const TCmd *pCmd, Player &player)
 	return sizeof(message);
 }
 
-size_t OnObjectTileAction(const TCmd &cmd, Player &player, action_id action)
+size_t OnObjectTileAction(const TCmd &cmd, Player &player, action_id action, bool pathToObject = true)
 {
 	const auto &message = reinterpret_cast<const TCmdLoc &>(cmd);
 	const Point position { message.x, message.y };
 	const Object *object = ObjectAtPosition(position);
 
 	if (gbBufferMsgs != 1 && player.isOnActiveLevel() && object != nullptr) {
-		MakePlrPath(player, position, !object->_oSolidFlag && !object->_oDoorFlag);
+		if (pathToObject)
+			MakePlrPath(player, position, !object->_oSolidFlag && !object->_oDoorFlag);
 
 		player.destAction = action;
 		player.destParam1 = object->GetId();
-	}
-
-	return sizeof(message);
-}
-
-size_t OnOperateObjectTelekinesis(const TCmd *pCmd, Player &player)
-{
-	const auto &message = *reinterpret_cast<const TCmdParam1 *>(pCmd);
-
-	if (gbBufferMsgs != 1 && player.isOnActiveLevel() && message.wParam1 < MAXOBJECTS) {
-		player.destAction = ACTION_OPERATETK;
-		player.destParam1 = message.wParam1;
 	}
 
 	return sizeof(message);
@@ -3005,7 +2994,7 @@ size_t ParseCmd(int pnum, const TCmd *pCmd)
 	case CMD_DISARMXY:
 		return OnObjectTileAction(*pCmd, player, ACTION_DISARM);
 	case CMD_OPOBJT:
-		return OnOperateObjectTelekinesis(pCmd, player);
+		return OnObjectTileAction(*pCmd, player, ACTION_OPERATETK, false);
 	case CMD_ATTACKID:
 		return OnAttackMonster(pCmd, player);
 	case CMD_ATTACKPID:
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1215,29 +1215,17 @@ size_t OnTargetSpellTile(const TCmd *pCmd, Player &player)
 	return sizeof(message);
 }
 
-size_t OnOperateObjectTile(const TCmd *pCmd, Player &player)
+size_t OnObjectTileAction(const TCmd &cmd, Player &player, action_id action)
 {
-	const auto &message = *reinterpret_cast<const TCmdLocParam1 *>(pCmd);
+	const auto &message = reinterpret_cast<const TCmdLoc &>(cmd);
 	const Point position { message.x, message.y };
+	const Object *object = ObjectAtPosition(position);
 
-	if (gbBufferMsgs != 1 && player.isOnActiveLevel() && InDungeonBounds(position) && message.wParam1 < MAXOBJECTS) {
-		MakePlrPath(player, position, !Objects[message.wParam1]._oSolidFlag && !Objects[message.wParam1]._oDoorFlag);
-		player.destAction = ACTION_OPERATE;
-		player.destParam1 = message.wParam1;
-	}
+	if (gbBufferMsgs != 1 && player.isOnActiveLevel() && object != nullptr) {
+		MakePlrPath(player, position, !object->_oSolidFlag && !object->_oDoorFlag);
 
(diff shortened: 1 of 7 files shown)


Source/track.cpp
```diff
@@ -111,11 +109,8 @@ void RepeatMouseAction()
 			CheckPlrSpell(false);
 		break;
 	case MouseActionType::OperateObject:
-		if (pcursobj != -1) {
-			auto &object = Objects[pcursobj];
-			if (object.IsDoor())
-				break;
-			NetSendCmdLocParam1(true, CMD_OPOBJXY, object.position, pcursobj);
+		if (ObjectUnderCursor != nullptr && !ObjectUnderCursor->IsDoor()) {
+			NetSendCmdLocParam1(true, CMD_OPOBJXY, ObjectUnderCursor->position, ObjectUnderCursor->GetId());
```
AJenbo, line 113: Shouldn't this be changed to:
```suggestion
			NetSendCmdLoc(MyPlayerId, true, CMD_OPOBJXY, ObjectUnderCursor->position);
```

Source/msg.cpp
```diff
@@ -1218,44 +1218,34 @@ DWORD OnTargetSpellTile(const TCmd *pCmd, Player &player)
...
-	const auto &message = *reinterpret_cast<const TCmdLocParam1 *>(pCmd);
-	const Point position { message.x, message.y };
-
-	if (gbBufferMsgs != 1 && player.isOnActiveLevel() && InDungeonBounds(position) && message.wParam1 < MAXOBJECTS) {
-		MakePlrPath(player, position, !Objects[message.wParam1]._oSolidFlag && !Objects[message.wParam1]._oDoorFlag);
-		player.destAction = ACTION_DISARM;
-		player.destParam1 = message.wParam1;
-	}
+	return OnObjectTileAction(*reinterpret_cast<const TCmdLoc *>(pCmd), player, ACTION_OPERATE);
+}
```
AJenbo, line 1239: I think it would actually be simple to have `OnObjectTileAction` do the case and then manually inline the 3 calls (completely getting rid of the old functions).

Source/msg.cpp
```diff
@@ -1218,44 +1218,34 @@ DWORD OnTargetSpellTile(const TCmd *pCmd, Player &player)
...
 {
-	const auto &message = *reinterpret_cast<const TCmdLocParam1 *>(pCmd);
-	const Point position { message.x, message.y };
+	const Object *object = ObjectAtPosition({ message.x, message.y });
 
-	if (gbBufferMsgs != 1 && player.isOnActiveLevel() && InDungeonBounds(position) && message.wParam1 < MAXOBJECTS) {
-		MakePlrPath(player, position, !Objects[message.wParam1]._oSolidFlag && !Objects[message.wParam1]._oDoorFlag);
-		player.destAction = ACTION_OPERATE;
-		player.destParam1 = message.wParam1;
+	if (gbBufferMsgs != 1 && player.isOnActiveLevel() && object != nullptr) {
```
AJenbo, line 1225: nice, this is cleaner and provides stronger validation of the message sanity.

AJenbo:
looks like it's currently causing a gameplay change, let me know if you need help replaying the demo or narrow it down otherwise.

Source/msg.cpp
```diff
@@ -3013,11 +3001,9 @@ uint32_t ParseCmd(int pnum, const TCmd *pCmd)
 	case CMD_TSPELLXY:
 		return OnTargetSpellTile(pCmd, player);
 	case CMD_OPOBJXY:
-		return OnOperateObjectTile(pCmd, player);
 	case CMD_DISARMXY:
-		return OnDisarm(pCmd, player);
 	case CMD_OPOBJT:
-		return OnOperateObjectTelekinesis(pCmd, player);
+		return OnObjectTileAction(*reinterpret_cast<const TCmdLoc *>(pCmd), player);
```
AJenbo, line 3006: This might be the opposite of what I was thinking, so in case it was bad communication on my part here is a code sample instead :D

```cpp
	case CMD_OPOBJXY:
		return OnObjectTileAction(pCmd, player, ACTION_OPERATE);
	case CMD_DISARMXY:
		return OnObjectTileAction(pCmd, player, ACTION_DISARM);
	case CMD_OPOBJT:
		return OnObjectTileAction(pCmd, player, ACTION_OPERATETK, false);
```

ephphatha:
> looks like it's currently causing a gameplay change, let me know if you need help replaying the demo or narrow it down otherwise.

The demo desyncs about 2 minutes in while highlighting a monster, doesn't appear directly related to this change. Not sure why that's started failing :/

Source/msg.cpp
```diff
@@ -3013,11 +2988,11 @@ uint32_t ParseCmd(int pnum, const TCmd *pCmd)
...
 		return OnTargetSpellTile(pCmd, player);
 	case CMD_OPOBJXY:
-		return OnOperateObjectTile(pCmd, player);
+		return OnObjectTileAction(*reinterpret_cast<const TCmdLoc *>(pCmd), player, ACTION_OPERATE);
 	case CMD_DISARMXY:
-		return OnDisarm(pCmd, player);
+		return OnObjectTileAction(*reinterpret_cast<const TCmdLoc *>(pCmd), player, ACTION_DISARM);
 	case CMD_OPOBJT:
-		return OnOperateObjectTelekinesis(pCmd, player);
+		return OnObjectTileAction(*reinterpret_cast<const TCmdLoc *>(pCmd), player, ACTION_OPERATETK, false);
```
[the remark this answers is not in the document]
ephphatha: It was more to make it clear what the function actually operates on, but I can move it in.

AJenbo:
`NetSendCmdLoc` is throttled where I don't think that is the case for `NetSendCmdLocParam1`

Source/msg.cpp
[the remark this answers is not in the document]
AJenbo: I remember doing the same when we where working to get things bin exact :)

ephphatha:
Yeah, I pass in playerId as MyPlayerId to ensure the first case evaluates to the same result. LocParam1 always checks WasPlayerCmdAlreadyRequested, Loc only checks if playerId==MyPlayerId

Source/track.cpp
```diff
@@ -115,7 +115,8 @@ void RepeatMouseAction()
 			auto &object = Objects[pcursobj];
 			if (object.IsDoor())
 				break;
-			NetSendCmdLocParam1(true, CMD_OPOBJXY, object.position, pcursobj);
+			// This should probably be cursPosition so paths to large objects are consistent
+			NetSendCmdLoc(MyPlayerId, true, CMD_OPOBJXY, object.position);
```
ephphatha, line 119: Finally worked out what was breaking the timedemo test.

Large objects like sarcophagi can end up with different pathfinding depending on which tile the cursor is over when interacting with them. Using the object position isn't correct as this is always the southernmost tile, the cursor might be over another tile and the player might want to path to a different edge.

However repeated mouse actions break pathing and force navigating to the southernmost tile. This gets triggered twice in the demo, the pathfinding diverges if I try fix this.

AJenbo: Nice work! That can't have been easy to track down. I guess who ever fixes this have to also record a new demo.

AJenbo (approved):

Source/track.cpp
ephphatha: Which is why I'm trying to avoid fixing it 😂

AJenbo: At least you added a comment :)

PR #4989, opened 2022-07-13
CMake: Use the mold linker where available

glebm:
https://github.com/rui314/mold is quite stable these days and is insanely fast.

Let's use it by default if it's there.

Linking a debug `devilutionx` binary:

GNU ld: 1.82s
Mold: 0.03s

(mold 1.3.1 [installed from source](https://github.com/rui314/mold#compile-mold) on Ubuntu 22.04)


AJenbo (approved):
I heard gold was fast compared to ld :)

PR #4994, opened 2022-07-13
:recycle: Replace AddMissile _misource with pointers

canassa:
Removes the _misource from the Missile struct and replaces it with *sourcePlayer and *sourceMonster.

This should improve the code readability and can simplify a few parts.

I am opening this as a draft. While the code compiles, it's still crashing and I don't know how to handle load and save games.


Source/missiles.h
```diff
@@ -107,7 +107,8 @@ struct Missile {
 	bool _miPreFlag;
 	uint32_t _miUniqTrans;
 	int _mirange; // Time to live for the missile in game ticks, oncs 0 the missile will be marked for deletion via _miDelFlag
-	int _misource;
+	Monster *sourceMonster;
```
glebm, line 110: This makes the struct larger.

Rather than replacing the field with 2 pointers, you can add methods to the struct that return the pointers.

glebm (changes requested):

Source/missiles.h
canassa: Wouldn't that make the access to the pointers a bit slower?

AJenbo: It wouldn't be slower then the current code since it already had to do the lookup by index.

@glebm how about we do this as a union since the two pointers are mutually exclusive and will be the same size as the int was on 32bit systems?

glebm: We could but that index should probably just be 16-bit, not 32 (with 1 bit for type Monster vs Player). And you'd need a discriminator, so ~40 bits vs 16.

The access won't be slower (an addition of a known address + offset is trivial).

In the situations where making the struct smaller means one less cache line read, it will even make it *a lot* faster.

AJenbo: If you reuse the scheme from encode_enemy/decode_enemy you can fit it in a single `uint8_t`.

canassa: Do you guys mean something like this?

```c++
	[[nodiscard]] Monster* sourceMonster()
	{
		return _misource != -1 ? &Monsters[_misource] : nullptr;
	}

	[[nodiscard]] Player* sourcePlayer()
	{
		return _misource != -1 ? &Players[_misource] : nullptr;
	}
```

AJenbo: Something like
```cpp
uint8_t source;

Player* getPlayer() {
	if (source >= MAX_PLRS)
		return null;
	return Players[_misource];
}

Monster* getMonster() {
	if (source < MAX_PLRS || _misource == 255)
		return null;
	return Monsters[_misource - MAX_PLRS];
}

MissileSourceType getSourceType() {
	if (source == 255)
		return MissileSourceType::Trap;
	if (source < MAX_PLRS)
		return MissileSourceType::Player;
	return MissileSourceType::Monster;
}
```
(don't actually use a literal 255, use limits instead)

Fun fact, even a low end Epyc can fit the entire game in L3.

AJenbo: here is roughly what Glebm suggested if I'm not mistaking:
```cpp
uint16_t source;

MissileSourceType getSourceType() {
	if (source == 0xFFFF)
		return MissileSourceType::Trap;
	if ((source & (1 << 15)) != 0)
		return MissileSourceType::Player;
	return MissileSourceType::Monster;
}

Player* getPlayer() {
	if (getSourceType() != MissileSourceType::Player)
		return null;
	return Players[_misource];
}

Monster* getMonster() {
	if (getSourceType() != MissileSourceType::Monster)
		return null;
	return Monsters[_misource - MAX_PLRS];
}
```

canassa: I have two questions:

1- Are we assuming that the compiler will inline that code? otherwise, there will be added overhead of a function call, no?

2- Those `if` are all extra operations that needed to be performed every time getPlayer or getMonster is called, right? So if I understand correctly, we are trading CPU for memory?

AJenbo: For savegame:
int oldFormat = 0;
switch (missile.getSourceType()) {
	catch MissileSourceType::Player:
		oldFormat = missile.getPlayer().getId();
		break;
	catch MissileSourceType::Monster:
		oldFormat = missile.getMonster().getId();
		break;
}

glebm: 1. The compiler will inline this code. Define such function in the header to allow the compiler to inline them without LTO.
2. Not really. 
3.

canassa: Those are some really good ideas! thanks guys!

I really don't like humongous PRs and one really nice about adding the methods is that it allows for this change to be split into smaller chunks. I came up with this plan, let me know what you think:

### Step 1 : introduce `SourcePlayer()` and `SourceMonster()`
In Step 1 we will only add those methods to the Missile struct and nothing else. `_misource` will be kept as-is for now so the code will be backwards compatible

### Step 2: Migrate functions that are accessing `_misource` to use the new methods instead
This can be done in multiple PR's. The goal is to change the functions that are accessing `_misource` directly and replace them with calls to either `SourcePlayer()` or `SourceMonster()`

### Step 3: Replace the `id` parameter of `AddMissile()` with reference to either Player or Monster

### Step 4: Optimize
With every direct access to `_misource` removed, we can optimize how the Missile struct memory like you guys suggested by replacing the `int` with something that has a smaller footprint.

AJenbo: 1 will give you warnings that they are unused, so combine that with 2 and you have a good plan.

AJenbo: This should work for the current structure
```cpp
int _misource;

Player* getPlayer() {
	if (IsNoneOf(_micaster, TARGET_BOTH, TARGET_MONSTERS))
		return null;
	return Players[_misource];
}

Monster* getMonster() {
	if (_micaster != TARGET_PLAYERS)
		return null;
	return Monsters[_misource];
}

MissileSourceType getSourceType() {
	if (_misource == -1)
		return MissileSourceType::Trap;
	if (_micaster == TARGET_PLAYERS)
		return MissileSourceType::Monster;
	return MissileSourceType::Player;
}
```

Source/loadsave.cpp
```diff
@@ -727,7 +727,8 @@ void LoadMissile(LoadHelper *file)
 	missile._miPreFlag = file->NextBool32();
 	missile._miUniqTrans = file->NextLE<uint32_t>();
 	missile._mirange = file->NextLE<int32_t>();
-	missile._misource = file->NextLE<int32_t>();
+	// TODO: How to handle this?
+	//	missile._misource = file->NextLE<int32_t>();
```
AJenbo, line 731: Decode the old format, fetch the monster or player based on id, get a pointer to them and set that on the missile.

Source/loadsave.cpp
```diff
@@ -1447,7 +1448,8 @@ void SaveMissile(SaveHelper *file, const Missile &missile)
 	file->WriteLE<uint32_t>(missile._miPreFlag ? 1 : 0);
 	file->WriteLE<uint32_t>(missile._miUniqTrans);
 	file->WriteLE<int32_t>(missile._mirange);
-	file->WriteLE<int32_t>(missile._misource);
+	// TODO: How to handle this?
+	//	file->WriteLE<int32_t>(missile._misource);
```
AJenbo, line 1452: .getId() on the player/monster, or -1

AJenbo:
Closing as this is being done differently now.

PR #4995, opened 2022-07-14
Replace issue template with issue form

StephenCWills:
The hope is to reduce the frequency of bug reports that are missing crucial information.

Here's how the form looks.
https://github.com/StephenCWills/devilutionX/issues/new?assignees=&labels=&template=bug_report.yml

Here's an example issue.
https://github.com/StephenCWills/devilutionX/issues/1


AJenbo (approved):
The drop down options feel like they could potentially be a limit, but maybe I'm just being to cautious. Lets see how this works for starters :)

PR #4996, opened 2022-07-14
Restore legacy behaviour for GetSizeForThemeRoom

ephphatha:
Going back to the old and slow logic until I can sort out all the little quirks that end up with layout differences. The long running test case is to force multiple failures and see when things diverge/fall back in line. It's over the top but flags corner cases most runs.

### DIFF
--- a/Source/levels/gendung.cpp
+++ b/Source/levels/gendung.cpp
@@ -388,6 +388,13 @@ void InitGlobals()
 
 } // namespace
 
+#ifdef BUILD_TESTING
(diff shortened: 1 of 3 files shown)


qndel:
Is it a clean revert of a commit? If yes, pls mention which one, if not, ignore ;)

ephphatha:
Implemented in #5010

AJenbo:
well all wxcwpt thw one that takes multiple minutes to run ;)

PR #5000, opened 2022-07-14
Rename MonsterData members

aetn23:
Rename `MonsterData` members. Remove `m` prefix, make the names more clear and consistent with `Monster` members. 

This is done to improve readability & consistency.


Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
+	uint8_t armorClass;
+	MonsterClass monsterClass;
 	/** Using monster_resistance as bitflags */
-	uint8_t mMagicRes;
+	uint8_t magicResistance;
 	/** Using monster_resistance as bitflags */
-	uint8_t mMagicRes2;
-	int8_t mSelFlag; // TODO Create enum
+	uint8_t magicResistanceHellDiff;
+	int8_t sellFlag; // TODO Create enum
```
DakkJaniels, line 126: why sellFlag? I believe this has to do with the selection outline for the monster.

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
+	uint8_t maxDamage;
+	uint8_t hit2;
+	int8_t animFrameNum2;
+	uint8_t minDamage2;
+	uint8_t maxDamage2;
+	uint8_t armorClass;
+	MonsterClass monsterClass;
 	/** Using monster_resistance as bitflags */
-	uint8_t mMagicRes;
+	uint8_t magicResistance;
```
DakkJaniels, line 123: Probably remove "magic" from these, since this is for all resistances (fire and lightning too), not just magic.

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
+	uint16_t image;
+	bool hasSpecial;
+	bool sndSpecial;
+	int8_t frames[6];
+	int8_t rate[6];
+	int8_t minDunLvl;
+	int8_t maxDunLvl;
+	int8_t level;
+	uint16_t maxHitPointsMinimum;
+	uint16_t maxHitPointsMaximum;
```
DakkJaniels, line 107: is `max` really required at the start of these? since this is the basis for the creation of monsters, isn't it implied that this structure is not used to store current hitpoints?

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
-	uint8_t mMaxDamage2;
-	uint8_t mArmorClass;
-	MonsterClass mMonstClass;
+	uint16_t flags;
+	uint8_t intelligence;
+	uint8_t hit;
+	int8_t animFrameNum;
+	uint8_t minDamage;
+	uint8_t maxDamage;
+	uint8_t hit2;
```
DakkJaniels, line 116: Can these (L112 and L116) be toHit?

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
-	int8_t mAFNum;
-	uint8_t mMinDamage;
-	uint8_t mMaxDamage;
-	uint8_t mHit2;
-	int8_t mAFNum2;
-	uint8_t mMinDamage2;
-	uint8_t mMaxDamage2;
-	uint8_t mArmorClass;
-	MonsterClass mMonstClass;
+	uint16_t flags;
```
DakkJaniels, line 110: maybe call this one `abilityFlags`, or `abilities` and put in a comment that it's bitflags?

Source/monstdat.h
AJenbo: It defines what shape it will have in relation to the mouse. See https://github.com/diasurgical/devilutionX/blob/master/Source/cursor.cpp#L404

Source/monstdat.h
AJenbo: Yeah, that was my mistake last time, so if you can also change that for the monster struct while you are at it that would be good :)

Source/monstdat.h
AJenbo: > abilities and put in a comment that it's bitflags?

Sound nice :+1:

Source/monstdat.h
DakkJaniels: but why sell? shouldn't it be shape? or selectionShape?

AJenbo: Sorry I got confused because both you and @aetn23 have a similarly generic avatar on GitHub :D

Source/cursor.cpp
```diff
@@ -405,49 +405,49 @@ void CheckCursMove()
 		if (pcurstemp != -1) {
 			if (!flipflag && mx + 2 < MAXDUNX && my + 1 < MAXDUNY && dMonster[mx + 2][my + 1] != 0 && IsTileLit({ mx + 2, my + 1 })) {
 				int mi = abs(dMonster[mx + 2][my + 1]) - 1;
-				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().mSelFlag & 4) != 0) {
+				if (mi == pcurstemp && Monsters[mi].hitPoints >> 6 > 0 && (Monsters[mi].data().sellFlag & 4) != 0) {
```
qndel, line 408: if it's about selection, it should be sel, not sell

Source/monstdat.h
AJenbo: I agree with removing max.

We already settled on hitPoints in a previous cleanup. HP isn't good for searching either.

aetn23: I put `max` there to distinguish between field `maxHitPoints` in `Monster` and this one. Without `max` they would be similarly named, but they don't have a `=` correspondence. Those two vars in `MonsterData` are used for generating `maxHitPoints` for a given monster, so I tried to reflect that in those names. I think the name `hitPointsMaximum` could wrongly lead to an assumption that this field and the `maxHitPoints` field in monster are the same. 
 
 But maybe there are better names for those 2 altogether...

AJenbo: I think they are two separeret contexts

Source/monstdat.cpp
```diff
@@ -14,7 +14,7 @@ namespace devilution {
 
 namespace {
 
-// Returns an `mTreasure` value for the given item.
+// Returns an `treasure` value for the given item.
```
qndel, line 17:
```suggestion
// Returns a `treasure` value for the given item.
```

AJenbo:
Happy `#5000` :tada:

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
-	uint8_t mMaxDamage;
-	uint8_t mHit2;
-	int8_t mAFNum2;
-	uint8_t mMinDamage2;
-	uint8_t mMaxDamage2;
-	uint8_t mArmorClass;
-	MonsterClass mMonstClass;
+	uint16_t image;
+	bool hasSpecial;
+	bool sndSpecial;
```
AJenbo, line 100: hasSpecialSound

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
+	uint8_t armorClass;
+	MonsterClass monsterClass;
 	/** Using monster_resistance as bitflags */
-	uint8_t mMagicRes;
+	uint8_t resistance;
 	/** Using monster_resistance as bitflags */
-	uint8_t mMagicRes2;
-	int8_t mSelFlag; // TODO Create enum
+	uint8_t resistanceHellDiff;
+	int8_t selectionFlag; // TODO Create enum
```
AJenbo, line 126: Maybe `selectionType` since flag normally denotes bit fileds.

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
+	uint16_t abilityFlags;
+	uint8_t intelligence;
+	uint8_t toHit;
+	int8_t animFrameNum;
+	uint8_t minDamage;
+	uint8_t maxDamage;
+	uint8_t toHitSpecial;
+	int8_t animFrameNum2;
+	uint8_t minDamage2;
+	uint8_t maxDamage2;
```
AJenbo, line 119: `Special`?

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
+	bool sndSpecial;
+	int8_t frames[6];
+	int8_t rate[6];
+	int8_t minDunLvl;
+	int8_t maxDunLvl;
+	int8_t level;
+	uint16_t hitPointsMinimum;
+	uint16_t hitPointsMaximum;
+	_mai_id ai;
+	/** Using @p monster_flag as bitflags */
```
AJenbo, line 109: ?

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
 
 struct MonsterData {
-	const char *mName;
-	const char *GraphicType;
-	const char *sndfile;
-	const char *TransFile;
+	const char *name;
+	const char *graphicType;
+	const char *sndFile;
+	const char *transFile;
```
AJenbo, line 95: lets just call it `trnFile` since that is the file type, no need to have multiple acronyms for this.

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
 };
 
 struct MonsterData {
-	const char *mName;
-	const char *GraphicType;
-	const char *sndfile;
-	const char *TransFile;
+	const char *name;
+	const char *graphicType;
```
AJenbo, line 93:
```suggestion
	const char *assetsSuffix;
```
That way it will have less conflicts with https://github.com/diasurgical/devilutionX/pull/5001 :)

Source/monstdat.h
qndel: let's make it consistent with the `snd` below so either snd in both cases or sound

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
...
+	uint8_t maxDamage2;
+	uint8_t armorClass;
+	MonsterClass monsterClass;
 	/** Using monster_resistance as bitflags */
-	uint8_t mMagicRes;
+	uint8_t resistance;
 	/** Using monster_resistance as bitflags */
-	uint8_t mMagicRes2;
-	int8_t mSelFlag; // TODO Create enum
+	uint8_t resistanceHellDiff;
```
AJenbo, line 125: Nit: `resistanceHell` might also be enough, I don't particularly like "diff" as it's normally "difference" for me.

Source/monstdat.h
```diff
@@ -89,44 +89,44 @@ enum class MonsterAvailability : uint8_t {
 };
 
 struct MonsterData {
-	const char *mName;
-	const char *GraphicType;
-	const char *sndfile;
-	const char *TransFile;
+	const char *name;
+	const char *graphicType;
+	const char *sndFile;
```
AJenbo, line 94:
```suggestion
	const char *soundSuffix;
```

Source/monstdat.h
aetn23: Makes sense to me, it's shorter and clear enough

Source/monstdat.h
```diff
@@ -106,24 +106,27 @@ struct MonsterData {
 	uint16_t hitPointsMinimum;
 	uint16_t hitPointsMaximum;
 	_mai_id ai;
-	/** Using @p monster_flag as bitflags */
+	/**
+	 * @brief Denotes monster's abilities defined in @p monster_flag as bitflags
```
AJenbo, line 110: I don't understand `@p`, syntax error, brain out of context :shrug:

Source/monstdat.h
AJenbo: I don't understand `@p`, syntax error, brain out of context :shrug:

Source/monstdat.h
aetn23: This '@p` is to make `monster_flag` appear in typewriter font, since it is an enum

AJenbo: ok

PR #5001, opened 2022-07-15
Clean up LevelMonsterTypes related code

AJenbo:
Clean up `LevelMonsterTypes` related code.

@glebm there are a couple of call chains that I would have linked to change to take `CMonster` by reference rather than an index in to `LevelMonsterTypes`, but this isn't really possible now that Monster only has the index in to `LevelMonsterTypes`. How do you feel about if we change it to have a pointer to `CMonster` instead?


glebm:
> but this isn't really possible now that Monster only has the index in to LevelMonsterTypes

`monster.type()` returns a `const CMonster&`, isn't that what you need?

AJenbo:
No, i need to replace:

```cpp
void a(int i){ b(i); }
void b(int i){ c(i); }
void c(int i) {
    monster.monsterType = i;
}
```
with
```cpp
void a(const CMonster &i){ b(i); }
void b(const CMonster &i){ c(i); }
void c(const CMonster &i) {
    monster.type = &type;
}
```

Source/monstdat.cpp
```diff
@@ -56,7 +56,7 @@ const MonsterData MonstersData[] = {
...
+/* MT_INVILORD*/ { P_("monster", "Invisible Lord"),          "TSneak\\TSneak",   nullptr,          nullptr,             MonsterAvailability::Never,    128,    800, false,       false,       { 13, 13, 15, 11, 16,  0 }, { 2, 1, 1, 1, 1, 1 },        19,       20,     14,    278,    278, AI_SKELSD,                                     MFLAG_SEARCH | MFLAG_CAN_OPEN_DOOR,    3,   65,      8,         16,         30,     0,       0,           0,           0,          60, MonsterClass::Demon,    RESIST_MAGIC | RESIST_FIRE | RESIST_LIGHTNING | IMMUNE_NULL_40  , RESIST_MAGIC | RESIST_FIRE | RESIST_LIGHTNING | IMMUNE_NULL_40    ,  3,                   0,         2000 },
```
vlolteanu, line 59: I take it "TSneak\\Sneakl" does not exist in the MPQ, right?

Source/monster.cpp
```diff
@@ -192,7 +192,7 @@ void InitMonsterTRN(CMonster &monst)
 	}
 }
 
-void InitMonster(Monster &monster, Direction rd, int mtype, Point position)
+void InitMonster(Monster &monster, Direction rd, size_t mtype, Point position)
```
vlolteanu, line 195: How about replacing `size_t mtype` with `size_t typeIndex`? There's also `_monster_id type` sprinkled around, and the two are very different.

Source/monstdat.cpp
vlolteanu: Shouldn't we either keep stuff like this intact for potential modders, or remove it altogether? I don't see any reason why some fields should go, but others should not.

Source/monster.cpp
AJenbo: Tried cleaning up the various names a bit, let me know what you think :)

Source/monstdat.cpp
AJenbo: We should remove it. Modders can always fish out the data and partial routines and reimplement them.

The reasoning here is that I knew these to have no usage and no consequence for removing them, where I'm less sure about the rest. But having this removed should make the rest easier to reason about.

Source/monster.cpp
vlolteanu: LGTM.

AJenbo:
thanks

PR #5009, opened 2022-07-16
Remove strncmp

glebm:


Source/utils/language.cpp
```diff
@@ -241,36 +222,25 @@ void ParsePluralForms(const char *string)
...
-		}
-
-		val = StrTrimRight(val);
+	while (!metadata.empty() && ((delim = metadata.find(':')) != string_view::npos)) {
+		const string_view key = TrimLeft(string_view(metadata.data(), delim));
+		string_view val = TrimLeft(string_view(metadata.data() + delim + 1, metadata.size() - delim - 1));
 
-		if ((strcmp("Content-Type", key) == 0) && ((delim = strstr(val, "=")) != nullptr)) {
-			if (strcasecmp(delim + 1, "utf-8") != 0) {
-				Log("Translation is now UTF-8 encoded!");
```
AJenbo, line 265: What's the thinking for not warning about invalid translation files?

glebm: The previous version didn't warn either (the code was incorrect and never logged anything).
We control the build process for the translation files so they're always valid.

glebm: Shall I bring it back (and fix the message)?

AJenbo: It could potentially be useful for translators, maybe also reports of strange behavior although garbage text should probably tip us off. Many users also do not look in the log.
I did find a Russian forum where someone is editing the translation in order to improve it, but do not appear to be considering feed it back to the project (they may well be hex editing it).
Not sure any of this qualifies as a reason for keeping it. Especially not if it was previously broken. Besides we should not even be loading the translation if we are not able to use it.

So I'll just make a mental note that there is an option for validating data here. And see if i can get the translation validation to also check the encoding.

AJenbo (approved):

PR #5012, opened 2022-07-16
Xbox One: Package assets/ directory

glebm:


AJenbo (commented):
Best to also disable loading of devilutionx.mpq so we don't get issues with people upgrading or being extra judicial in adding files.

glebm:
Shall we simply make `assets/` take priority over the MPQs?

AJenbo:
Not sure, will have to think :)

glebm:
Turns out the appx file already has the assets!

PR #5013, opened 2022-07-16
Cleanup GetSizeForThemeRoom

AJenbo:
A conservative cleanup of GetSizeForThemeRoom. I just went about it robotically so I don't understand more then a couple of lines and not the full function so someone with a level head can probably make this even nicer :D

Big thanks to @ephphatha for documenting a couple of bugs in the original as well as writing a useful test for messing with the code :+1:

### DIFF
--- a/Source/levels/gendung.cpp
+++ b/Source/levels/gendung.cpp
@@ -102,58 +102,32 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
 		return {};
 	}
 
-	bool yFlag = true;
-	bool xFlag = true;
-	int xCount = 0;
-	int yCount = 0;
+	int maxWidth = std::min(maxSize, DMAXX - origin.x);
+	int maxHeight = std::min(maxSize, DMAXY - origin.y);
+
+	int yArray[20] = {};
+	for (int x = 0; x < maxWidth; x++) {
+		for (int y = origin.y; y < origin.y + maxHeight; y++) {
+			if (dungeon[origin.x + x][y] != floor) {
+				if (yArray[x] < minSize && x < minSize)
+					return {};
+				break;
+			}
+
+			yArray[x]++;
+		}
+	}
 
 	int xArray[20] = {};
-	int yArray[20] = {};
+	for (int y = 0; y < maxHeight; y++) {
+		for (int x = origin.x; x < origin.x + maxWidth; x++) {
+			if (dungeon[x][origin.y + y] != floor) {
+				if (xArray[y] < minSize && y < minSize)
+					return {};
+				break;
+			}
 
-	for (int ii = 0; ii < maxSize; ii++) {
-		if (xFlag && origin.y + ii < DMAXY) {
-			for (int xx = origin.x; xx < origin.x + maxSize && xx < DMAXX; xx++) {
-				if (dungeon[xx][origin.y + ii] != floor) {
-					if (xx >= /* origin.x + */ minSize) {
-						break;
-					}
-					xFlag = false;
-				} else {
-					xCount++;
-				}
-			}
-			if (xFlag) {
-				xArray[ii] = xCount;
-				xCount = 0;
-			}
-		}
-		if (yFlag && origin.x + ii < DMAXX) {
-			for (int yy = origin.y; yy < origin.y + maxSize && yy < DMAXY; yy++) {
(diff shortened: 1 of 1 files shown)


AJenbo:
Here is a version with out any optimizations which might make it easier to understand the "logic".

```cpp
	int maxWidth = std::min(maxSize, DMAXX - origin.x);
	int maxHeight = std::min(maxSize, DMAXY - origin.y);

	int yArray[20] = {};
	for (int x = 0; x < maxWidth; x++) {
		for (int y = 0; y < maxHeight; y++) {
			if (dungeon[origin.x + x][origin.y + y] != floor)
				break;
			yArray[x]++;
		}
	}

	int xArray[20] = {};
	for (int y = 0; y < maxHeight; y++) {
		for (int x = 0; x < maxWidth; x++) {
			if (dungeon[origin.x + x][origin.y + y] != floor)
				break;
			xArray[y]++;
		}
	}

	for (int ii = 0; ii < minSize; ii++) {
		if (yArray[ii] < minSize || xArray[ii] < minSize)
			return {};
	}
```

AJenbo:
Managed to figure out how to roll the whole thing in to a single main loope. This both clears out all the special cases and improves performance so quite happy with this solution.

Source/levels/gendung.cpp
```diff
@@ -95,83 +95,42 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
...
-		}
-	}
+	Size room { maxWidth, maxHeight };
 
-	for (int ii = 0; ii < minSize; ii++) {
-		if (xArray[ii] < minSize || yArray[ii] < minSize) {
-			return {};
+	for (int i = 0; i < maxSize; i++) {
+		int width = 0;
+		for (int x = 0; x < room.width && i < maxHeight; x++) {
```
glebm, line 109: `maxHeight` and `i` don't change within the loop so I think this will be clearer if you wrap the loop with `if (i < maxHeight)`

Source/levels/gendung.cpp
```diff
@@ -95,83 +95,42 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
...
-					yCount++;
-				}
-			}
-			if (yFlag) {
-				yArray[ii] = yCount;
-				yCount = 0;
-			}
-		}
-	}
+	int maxWidth = std::min(maxSize, DMAXX - origin.x);
```
glebm, line 102: `const`?

AJenbo:
Running the extensive test on the original takes 137 sec, with the change it's down to 117. So that should translated to at least a 29% performance increase (ignoring the overhead)

Source/levels/gendung.cpp
```diff
@@ -95,83 +95,46 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
...
+	const int maxWidth = std::min(maxSize, DMAXX - origin.x);
+	const int maxHeight = std::min(maxSize, DMAXY - origin.y);
+
+	Size room { maxWidth, maxHeight };
+
+	for (int i = 0; i < maxSize; i++) {
+		int width = 0;
+		if (i < maxHeight) {
+			for (int x = 0; x < room.width; x++) {
+				if (dungeon[origin.x + x][origin.y + i] != floor)
```
ephphatha, line 111: Arguable if it's any clearer but width and x are interchangable in this loop. Could just use width directly.
```suggestion
			for (; width < room.width;) {
				if (dungeon[origin.x + width][origin.y + i] != floor)
```

(same for the loop over y)

ephphatha (dismissed):
Annoyingly it's really the best way to keep the original behaviour. At least this version is faster than the original.

ephphatha (approved):

Source/levels/gendung.cpp
AJenbo: Good catch, that is actually a nice improvement :)

AJenbo:
> Annoyingly it's really the best way to keep the original behaviour. At least this version is faster than the original.

Aw I was hoping that you would like it :(
Do you prefer the multiple loop style from [earlier](https://github.com/diasurgical/devilutionX/pull/5013#issuecomment-1186134731)?

ephphatha:
Was hoping there was a way to express it while only checking a given x/y position once -_-. What you've got here is a vast improvement over the old logic, but the two inner loops are what I really wanted to try do away with.

PR #5015, opened 2022-07-16
Remove `MAX_PATH`

glebm:
Rather than using the Windows-like `MAX_PATH`, what we really want is 2 things:

1. Maximum path of a file in an MPQ. This is now `MaxMpqPathSize`.
2. Max size for the known monster paths. This is now 64 for sounds and TRNs, which is more than enough (picked arbitrarily).


AJenbo (approved):
WINAPI will not be missed. Thanks :)

PR #5018, opened 2022-07-16
Replace `DVL_VK` virtual key codes with `SDLK`

glebm:
Removes one of the 2 redundant translation layers between SDL and the game.

We do 2 things to be compatible with the demo files:

1. Keymapper still uses uppercase latin characters (SDLK_a is lowercase, so we map to uppercase there).
2. Message `wParam` is still constructed in a special way instead of simply being `SDL_GetModState()`.

Mouse buttons are mapped in the key mapper by setting the highest bit of the key to 1.

Reduces rg99 binary size by 4 KiB.


AJenbo (approved):
Nice work :+1:

PR #5020, opened 2022-07-16
Reduce double scans in GetSizeForThemeRoom

AJenbo:
Double scan goes from `n^2` to just `n`. Resulting in a further 13% performance boost. Hope that makes it feel a bit more ideal :)

The remaining over scan comes from both loops starting from the same diagonal coordinates.

Given the following map and a max of 10x10:
image
Black = floor
Red = wall

Here is the scan patterns from the original:
image
Blue = vertical
Orange = horizontal
Green = overlapping scans

The version in this PR instead has this pattern:
image

### DIFF
--- a/Source/levels/gendung.cpp
+++ b/Source/levels/gendung.cpp
@@ -351,6 +351,13 @@ void InitGlobals()
 
 } // namespace
 
+#ifdef BUILD_TESTING
+std::optional<Size> GetSizeForThemeRoom()
+{
+	return GetSizeForThemeRoom(0, { 0, 0 }, 5, 10);
+}
+#endif
+
 dungeon_type GetLevelType(int level)
 {
 	if (level == 0)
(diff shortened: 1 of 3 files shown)


Source/levels/gendung.cpp
```diff
@@ -115,7 +115,7 @@ std::optional<Size> GetSizeForThemeRoom(int floor, Point origin, int minSize, in
 			}
 		}
 
-		int height = 0;
+		int height = i < room.width ? i : 0;
 		if (i < maxWidth) {
```
AJenbo, line 119: The following can be done to avoid the remaining double scans but doing so will actually degrade performance (and readability).
```cpp
		if (i < maxWidth) {
			if (width > i)
				height++;
```

ephphatha (approved):
The case illustrated in your comment would be good in the tests, should only need the four points that the  search actually touches.

Issue #3474, opened 2021-11-11
Import character from single player/multiplayer


vlolteanu:
It would be nice to be able to import your characters from single player to multiplayer and vice-versa.

UI-wise, I would envision a new menu entry under "New hero" called "Import hero".

AJenbo:
What where you thinking should happen to the hero's equipment?

vlolteanu:
It should stay the same.

I take it the item seeds produce different items for single player and multiplayer?

AJenbo:
> I take it the item seeds produce different items for single player and multiplayer?

Yeah, so that's why I'm curious what your thoughts where regarding how to handle that :) Hellfire just simply stripped all items from the hero when converting from vanilla to Hellfire.

vlolteanu:
Unless there's an easy way of converting them (i.e. not brute-force), I see little point in this proposal.

AJenbo:
Nothing short of adding support for single player items in multiplayer (and vice versa), and saves that go threw this process will no longer be backwards compatible with vanilla.

julealgon:
> ...and saves that go threw this process will no longer be backwards compatible with vanilla.

What if DevilutionX introduced a third type of save, that is neither vanilla SP nor vanilla MP, but a new unified mode, where characters can be migrated _to_ but not _from_? Have you considered something like this?

vlolteanu:
> > ...and saves that go threw this process will no longer be backwards compatible with vanilla.
> 
> What if DevilutionX introduced a third type of save, that is neither vanilla SP nor vanilla MP, but a new unified mode, where characters can be migrated _to_ but not _from_? Have you considered something like this?

With https://github.com/diasurgical/devilutionX/pull/2930 on the horizon, this might be the prudent approach.

AJenbo:
We would need to rewrite how multiplayer works then. And it would make it harder to verify if some issue is vanilla or not

julealgon:
> And it would make it harder to verify if some issue is vanilla or not

Why is this aspect important? Shouldn't any issue be evaluated by its own merits and potentially fixed anyways?

AJenbo:
It affects how critical it is. Figure out what solution should be applied. How to document the change. And sometimes figure out if something is an issue at all or just miss remembering hot things worked originally.

PR #5022, opened 2022-07-16
Toward importing characters to/from multiplayer

vlolteanu:
This PR is an early proof-of-concept attempting to resurrect https://github.com/diasurgical/devilutionX/issues/3474. In a nutshell, I think importing characters to and from multiplayer would be awesome, and some of you probably do, as well.

The main problem that item seeds generate wildly different items between single player and multiplayer. All that would be needed to be able to use both kinds of items at the same time is an extra flag, indicating whether or not the item is "foreign".

As far as storing the foreign items in save files, we have three options:

- An extra flag in the `ItemPack` structure: There doesn't seem to be any room.
- Flags in the `PlayerPack` structure: There are 18B of reserved space, which are set to 0. We can use 10B (8b for equipped items, 8b for belt items, 40b for the regular inventory). The PR uses this approach,
- An extra file in the MPQ archive containing the packed foreign items: This would be, by far, the cleanest approach. Loading such a save in vanilla DIablo would simply strip all foreign items without a hitch. (The other two would cause foreign items to morph.)

This PR does not consider the cornerstone of the world, or items dropped on the ground in multiplayer.

The PR also features stub menu entries that show how I'd envision the UI going forward. The UI code needs some refactoring to make this work.

### DIFF
--- a/Source/DiabloUI/selhero.cpp
+++ b/Source/DiabloUI/selhero.cpp
@@ -167,6 +167,10 @@ void SelheroListSelect(int value)
 		return;
 	}
 
+	if (static_cast<std::size_t>(value) == selhero_SaveCount + 1) {
+		return;
+	}
+
 	if (selhero_heroInfo.hassaved) {
 		vecSelDlgItems.clear();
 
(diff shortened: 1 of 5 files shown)


vlolteanu:
UI-wise, another approach would be a an INI option that lists all characters in both modes. Joining a multiplayer game would erase one's progress in single player.

AJenbo:
One thing that seams problematic is that monsters and others stats are scaled differently the two game mode. So bringing heros between them seams like it could have some general issues.

vlolteanu:
Could you give me a few pointers as to what might turn out to be problematic? I'll look into it.

AJenbo:
different difficaulity, different xp, unique items that are not suposed to be in multiplaer

vlolteanu:
I see. It sounds like this is more of an issue of someone not exactly earning their character. I'll one-up you on this point: SP is a lot easier and you can do a lot more risky stuff because you can quicksave/load.

Other than that, are there any other technical impediments that you foresee?

PR #5034, opened 2022-07-17
:recycle: Remove _misource access from MI_Arrow

canassa:


AJenbo (approved):

AJenbo:
I would like to put in a request for InitMissiles, AddTown to be done next

canassa:
> I would like to put in a request for InitMissiles, AddTown to be done next

On it 😃

Source/missiles.cpp
```diff
@@ -2699,23 +2699,24 @@ void MI_Arrow(Missile &missile)
...
-			maxd = player._pIMaxDam;
-		} else {
-			auto &monster = Monsters[p];
-			mind = monster.minDamage;
-			maxd = monster.maxDamage;
-		}
-	} else {
+	switch (missile.sourceType()) {
+	case MissileSource::Player: {
+		const Player *player = missile.sourcePlayer();
```
glebm, line 2707: We already check the source type, and then `sourcePlayer/Monster` checks the type again!

Generally, for variant types, the method that returns a particular value is not checked, and the caller instead must check the type. Let's do the same?

AJenbo: I think this is mostly a temporary issue until everything has been moved over.

canassa: That means we can also change sourceMonster and sourcePlayer to return a reference instead of a pointer, right?

glebm: Yeah

PR #5036, opened 2022-07-17
:sparkles: Adds isSameSource Missile method

canassa:
This is a precursor to removing all _misource direct access from the AddTown function.

### DIFF
--- a/Source/missiles.h
+++ b/Source/missiles.h
@@ -153,6 +153,11 @@ struct Missile {
 		return &Monsters[_misource];
 	}
 
+	[[nodiscard]] bool isSameSource(Missile &missile)
+	{
+		return _misource == missile._misource;
+	}
+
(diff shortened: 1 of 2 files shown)


Source/missiles.h
```diff
@@ -153,6 +153,11 @@ struct Missile {
 		return &Monsters[_misource];
 	}
 
+	[[nodiscard]] bool isSameSource(Missile &missile)
+	{
+		return _misource == missile._misource;
```
AJenbo, line 158: out of context of the specific spell this would need to also check the type of caster or monsters and players could collide ... of cause golems do not cast spells or shoot bows.

canassa: I added a check for the same sourceType too, do you think that's enough?

AJenbo (approved):

PR #5042, opened 2022-07-18
Minimize passing of player indexes

AJenbo:
The last prep needed for #4754

Changed `NetSendCmdGItem()` to only take one player id. The first (master) is actually unused for this message, but the two values where also always asserted to be the MyPlayerId before calling.

CMD_PLROPOBJ has been changed to a 1 param message as player id is always part of the messages.


Source/msg.cpp
```diff
@@ -1465,7 +1465,7 @@ size_t OnResurrect(const TCmd *pCmd, int pnum)
 	if (gbBufferMsgs == 1) {
 		SendPacket(pnum, &message, sizeof(message));
 	} else if (message.wParam1 < MAX_PLRS) {
-		DoResurrect(pnum, message.wParam1);
```
glebm, line 1468: Perhaps this check for `< Players.size()`? Otherwise this is a potential OOB

glebm (approved):

PR #5048, opened 2022-07-18
:recycle: Remove _misource access from CheckMissileCol

canassa:


Source/missiles.cpp
```diff
@@ -412,17 +420,18 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
 	bool blocked = false;
 	const int8_t pid = dPlayer[mx][my];
 	if (pid > 0) {
+		Player &hitPlayer = Players[pid - 1];
```
AJenbo, line 423: Usually we call this `Player &target`

Source/missiles.cpp
```diff
@@ -412,17 +420,18 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
 	bool blocked = false;
 	const int8_t pid = dPlayer[mx][my];
 	if (pid > 0) {
+		Player &hitPlayer = Players[pid - 1];
 		if (missile._micaster != TARGET_BOTH && !missile.IsTrap()) {
-			if (missile._micaster == TARGET_MONSTERS) {
-				if ((pid - 1) != missile._misource)
-					isPlayerHit = Plr2PlrMHit(Players[missile._misource], pid - 1, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted, &blocked);
+			if (missileSource == MissileSource::Player) {
```
AJenbo, line 425: This looks like another switch-case for Player, Monster, Trap

Source/missiles.cpp
```diff
@@ -388,17 +388,25 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
...
+			break;
+		case MissileSource::Monster: {
+			Monster *sourceMonster = missile.sourceMonster();
+			if ((sourceMonster->flags & MFLAG_TARGETS_MONSTER) != 0
+			    && ((sourceMonster->flags & MFLAG_GOLEM) != 0 || (sourceMonster->flags & MFLAG_BERSERK) != 0)) {
+				isMonsterHit = MonsterTrapHit(hitMonster.getId(), minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
+			}
+		} break;
+		case MissileSource::Player:
+			isMonsterHit = MonsterMHit(missile._misource, hitMonster.getId(), minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
```
AJenbo, line 409:
```suggestion
		mid = abs(mid) - 1;
		switch (missileSource) {
		case MissileSource::Player:
			isMonsterHit = MonsterMHit(missile._misource, mid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
			break;
		case MissileSource::Monster:
			if ((missile.sourceMonster()->flags & MFLAG_TARGETS_MONSTER) == 0 || (Monsters[mid].flags & (MFLAG_GOLEM | MFLAG_BERSERK)) == 0) {
				break;
			// No Break
		case MissileSource::Trap:
			isMonsterHit = MonsterTrapHit(mid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
			break;
```

Source/missiles.cpp
canassa: I tried to convert it, but the first `else` happens when the _micaster is TARGET_BOTH or a trap

Source/missiles.cpp
canassa: Done

Source/missiles.cpp
AJenbo: Here is what I came up with:

```cpp
		pid -= 1;
		int earflag = 0;
		switch (missileSource) {
		case MissileSource::Player: {
			Player *player = missile.sourcePlayer();
			if (&Players[pid] != player) {
				isPlayerHit = Plr2PlrMHit(*player, pid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted, &blocked);
				break;
			}
			earflag = IsAnyOf(missile._miAnimType, MFILE_FIREWAL, MFILE_LGHNING) ? 1 : 0;
		} // No Break
		default: {
			Monster *monster = missileSource == MissileSource::Monster ? missile.sourceMonster() : nullptr;
			isPlayerHit = PlayerMHit(pid, monster, missile._midist, minDamage, maxDamage, missile._mitype, isDamageShifted, earflag, &blocked);
			break;
		}
		}
```

It's not much of a switch, but it is using it to do less comparisons and combine some calls.

Source/missiles.cpp
```diff
@@ -410,19 +416,24 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
...
+		case MissileSource::Player: {
+			Player *player = missile.sourcePlayer();
+			if (&Players[pid] != player) {
+				isPlayerHit = Plr2PlrMHit(*player, pid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted, &blocked);
+				break;
 			}
-		} else {
-			int earflag = (!missile.IsTrap() && (missile._miAnimType == MFILE_FIREWAL || missile._miAnimType == MFILE_LGHNING)) ? 1 : 0;
-			isPlayerHit = PlayerMHit(pid - 1, nullptr, missile._midist, minDamage, maxDamage, missile._mitype, isDamageShifted, earflag, &blocked);
+			earflag = IsAnyOf(missile._miAnimType, MFILE_FIREWAL, MFILE_LGHNING) ? 1 : 0;
```
AJenbo, line 430: There is a bug in the original Hellfire here. The play will not drop an ear if killed by a fire rune, but will for lightning runes.

Source/missiles.cpp
```diff
@@ -410,19 +416,24 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
...
-					isPlayerHit = Plr2PlrMHit(Players[missile._misource], pid - 1, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted, &blocked);
-			} else {
-				Monster &monster = Monsters[missile._misource];
-				isPlayerHit = PlayerMHit(pid - 1, &monster, missile._midist, minDamage, maxDamage, missile._mitype, isDamageShifted, 0, &blocked);
+		pid -= 1;
+		int earflag = 0;
+		switch (missileSource) {
+		case MissileSource::Player: {
+			Player *player = missile.sourcePlayer();
+			if (&Players[pid] != player && missile._micaster != TARGET_BOTH) {
```
AJenbo, line 426: I'm not sure exactly how to classify this, but it looks like firewalls damages players as if they where cast by traps even when cast by players because of the special case for `TARGET_BOTH`. This PR makes this behavior more clear. I'm not sure if it should be addressed or not.

canassa: Perhaps we should keep the behaviour as is for now and create a separate issue to handle this later?

Source/missiles.cpp
AJenbo: I don't know if you have seen the original 200 line monstrosity: https://github.com/diasurgical/devilution/blob/b86b8fd7c4966385e3a7dbcb29e6d7e05bdab95e/Source/missiles.cpp#L1131

Source/missiles.cpp
```diff
@@ -388,17 +388,23 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
...
-		        && (Monsters[missile._misource].flags & MFLAG_TARGETS_MONSTER) != 0
-		        && ((Monsters[mid].flags & MFLAG_GOLEM) != 0 || (Monsters[missile._misource].flags & MFLAG_BERSERK) != 0))) {
-			isMonsterHit = MonsterTrapHit(mid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
-		} else if (IsAnyOf(missile._micaster, TARGET_BOTH, TARGET_MONSTERS)) {
+		switch (missileSource) {
+		case MissileSource::Player:
 			isMonsterHit = MonsterMHit(missile._misource, mid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
+			break;
+		case MissileSource::Monster:
+			if ((missile.sourceMonster()->flags & MFLAG_TARGETS_MONSTER) == 0 || (Monsters[mid].flags & (MFLAG_GOLEM | MFLAG_BERSERK)) == 0)
```
ephphatha, line 402: previously this was checking if the attacking monster was beserked or the target monster is a golem, now it's checking if the target monster is beserked or a golem. Is that intentional?

canassa: Is this function unit testable?

AJenbo: It's not directly exposed, but you could create a test function that calls it. Similarly to how it was done for `GetSizeForThemeRoom` https://github.com/diasurgical/devilutionX/blob/master/Source/levels/gendung.cpp#L354

canassa: I mean regarding the global variables, this function depends on many of those. I checked the other tests for examples of how to test this but didn't find any.

AJenbo: Most can probably be ignored :)

Source/missiles.cpp
```diff
@@ -410,19 +416,24 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
...
+	int8_t pid = dPlayer[mx][my];
 	if (pid > 0) {
-		if (missile._micaster != TARGET_BOTH && !missile.IsTrap()) {
-			if (missile._micaster == TARGET_MONSTERS) {
-				if ((pid - 1) != missile._misource)
-					isPlayerHit = Plr2PlrMHit(Players[missile._misource], pid - 1, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted, &blocked);
-			} else {
-				Monster &monster = Monsters[missile._misource];
-				isPlayerHit = PlayerMHit(pid - 1, &monster, missile._midist, minDamage, maxDamage, missile._mitype, isDamageShifted, 0, &blocked);
+		pid -= 1;
```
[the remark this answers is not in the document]
canassa: Fixed

Source/missiles.cpp
ephphatha: I tried to express this one a bit clearer in https://github.com/diasurgical/devilutionX/pull/5107, as far as I could tell the "does the target have the golem flag" check was serving double duty for detecting if a monster is berserked or controlled by a player. Seems like the correct approach was to check if either monster has the berserk flag.

PR #5050, opened 2022-07-18
:recycle: Remove _misource access from SpawnLightning

canassa:


Source/missiles.cpp
```diff
@@ -641,8 +641,7 @@ void SpawnLightning(Missile &missile, int dam)
 	if (!TileHasAny(pn, TileProperties::BlockMissile)) {
 		if (position != Point { missile.var1, missile.var2 } && InDungeonBounds(position)) {
 			missile_id type = MIS_LIGHTNING;
-			if (!missile.IsTrap() && missile._micaster == TARGET_PLAYERS
-			    && IsAnyOf(Monsters[missile._misource].type().type, MT_STORM, MT_RSTORM, MT_STORML, MT_MAEL)) {
+			if (IsAnyOf(missile.sourceMonster()->type().type, MT_STORM, MT_RSTORM, MT_STORML, MT_MAEL)) {
```
AJenbo, line 644: You need to either check the sourceType or return value. Based on @glebm's suggestion with removing the internal checks from sourceMonster() i guess a check for sourceType would be best here.l

AJenbo (approved):

PR #5051, opened 2022-07-18
Fix `DVL_WM_RBUTTONUP`

glebm:
Missed one case in #5047


AJenbo (approved):
It's shows that the demo is using a Warrior :D

PR #5053, opened 2022-07-18
:recycle: Remove _misource access from AddReflection

canassa:


Source/missiles.cpp
```diff
@@ -1106,19 +1106,19 @@ void AddReflection(Missile &missile, const AddMissileParameter & /*parameter*/)
 {
 	missile._miDelFlag = true;
 
-	if (missile._misource < 0)
+	if (missile.sourceType() != MissileSource::Player)
 		return;
 
-	Player &player = Players[missile._misource];
+	Player *player = missile.sourcePlayer();
```
ephphatha, line 1112:
```suggestion
	Player &player = *missile.sourcePlayer();
```
You can dereference the pointer immediately since the type was already checked earlier.

AJenbo (approved):

canassa:
> Please also apply `Player &player = *missile.sourcePlayer();` to the other cases that where already merged.

I will convert sourcePlayer and sourceMonster to return a reference instead of a pointer as soon as #5048 is merged

PR #5054, opened 2022-07-18
:recycle: Changes sourceMonster and sourcePlayer to return references

canassa:


canassa:
@AJenbo It's probably better to give this one priority over the others, the others needs to be rebased once this one is merged

AJenbo (commented):
I'm not thrilled that this has the potential to now return a broken state and crash the game if used incorrectly. It feels like raging the guard off your chainsaw just to cut a little bit faster

canassa:
> I'm not thrilled that this has the potential to now return a broken state and crash the game if used incorrectly. It feels like raging the guard off your chainsaw just to cut a little bit faster

I was following the comments from here: https://github.com/diasurgical/devilutionX/pull/5034#discussion_r923019574

How shall I proceed with this one? Shall I close it?

AJenbo:
idk, it doesn't feel right to me, but I'm willing to be convinced other wise.

canassa:
> idk, it doesn't feel right to me, but I'm willing to be convinced other wise.

I don't have any strong opinions on this. May I suggest that we close this for now? This is currently blocking the rest of the _misource removal.

AJenbo:
Yeah that fine, it can always be reopened

glebm:
There should be an `assert` inside these methods.

> I'm not thrilled that this has the potential to now return a broken state and crash the game if used incorrectly

If this is used incorrectly all bets are off anyway.

The code that uses that has 2 patterns:

1. Just call `*missile.sourcePlayer/Monster`.
2. Check `sourceType` then call `sourcePlayer/Monster`.
3. Check if `missile.sourcePlayer/Monster()` is nullptr first.

Case 3 should become case 2.
In both of these cases, there will be a crash as it is if used incorrectly even with the pointer return type.
So these signatures and checks are redundant anyway.

PR #5059, opened 2022-07-18
Load all CEL as CL2

glebm:
Convert CEL files to CL2 at load time. CL2 format is more efficient and is about as fast to render.

CEL vs CL2 sizes, on dLvl 5: https://gist.github.com/glebm/9bbdd76962abcd4fd2405ecd3379af97

Memory:

* Peak memory (while loading): -300 KiB
* Memory in-game (dLvl5): -700 KiB
* RG99 binary size: -15 KiB (1333096 -> 1317192)

Performance on rg99:

* On average, -1 FPS in town.
* Same FPS in dungeon (20 FPS on dLvl 1).


AJenbo:
An average saving of 26% and at a reduced line count, that's pretty awesome. What was the reason not to do this for the dungeon tiles?

glebm:
Doing this for dungeon tiles may slow tile rendering and not add that much in terms of memory savings because there aren't that many potential fill runs in them from what I can see:

Also, since dungeon tiles use a [a different format](https://github.com/savagesteel/d1-file-formats/blob/master/PC-Mac/CEL.md#422-level-cel-files), additional rendering code may be needed. Not to mention the special masking.

Source/utils/cel_to_cl2.cpp
```diff
@@ -0,0 +1,208 @@
...
+	const uint32_t maybeNumFrames = LoadLE32(data);
+
+	std::vector<uint8_t> cl2Data;
+
+	// Most files become smaller with CL2. Allocate exactly enough bytes to avoid reallocation.
+	// The only file that becomes larger is Data\hf_logo3.CEL, by exactly 4445 bytes.
+	cl2Data.reserve(size + 4445);
+
+	// If it is a number of frames, then the last frame offset will be equal to the size of the file.
+	if (LoadLE32(&data[maybeNumFrames * 4 + 4]) != size) {
```
AJenbo, line 117: nice to have logic handle this rather then having it hardcoded for every sprite.

Source/utils/cel_to_cl2.cpp
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		WriteLE32(&cl2Data[cl2DataOffset + 4 * (1 + static_cast<size_t>(numFrames))], static_cast<uint32_t>(cl2Data.size() - cl2DataOffset));
+		data = srcEnd;
+	}
+
+	auto out = std::unique_ptr<byte[]>(new byte[cl2Data.size()]);
+	memcpy(&out[0], cl2Data.data(), cl2Data.size());
+#ifdef DEBUG_CEL_TO_CL2_SIZE
+	std::cout << "\t" << size << "\t" << cl2Data.size() << "\t" << std::setprecision(1) << std::fixed << (static_cast<int>(cl2Data.size()) - static_cast<int>(size)) / ((float)size) * 100 << "%" << std::endl;
```
AJenbo, line 203: A bit unusual considering we normaly use Log()?

glebm: It's easier to copy paste without the INFO prefix (also stdout so can simply redirect to file)

AJenbo (approved):

PR #5060, opened 2022-07-18
:recycle: Remove _misource access from AddBerserk

canassa:


Source/missiles.cpp
```diff
@@ -1124,7 +1124,7 @@ void AddBerserk(Missile &missile, const AddMissileParameter &parameter)
 {
 	missile._miDelFlag = true;
 
-	if (missile._misource < 0)
+	if (missile.sourceType() == MissileSource::Trap)
 		return;
```
AJenbo, line 1128: Just mentioning this as a side note. This is a bit odd as a little later we make the assumption that it was cast by a player but make no check for monsters. Probably a better check would be `!= Player`, though that leaves the spell broken for monster/trap and it should probably just be left as unable to handle that case.

AJenbo (approved):

PR #5064, opened 2022-07-19
Use object pointer for pcursobj

ephphatha:
This was mainly to remove uses of `Object::GetId` but with #4987 most uses are already gone. Maybe it's still worthwhile?

It's still used for two network messages and player.destParam1, need to refactor player actions and handle loading/saving before it can be completely removed.


AJenbo (approved):
Nice to get rid of all those random lookups

PR #5065, opened 2022-07-19
Dedupe operate object message handlers, use position for break object

ephphatha:
These messages are handled almost identically, the only difference is the SyncBreakObject special case function. Prep to switch object messages over to position based instead of ID based.

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -537,13 +537,16 @@ void DeltaLeaveSync(uint8_t bLevel)
 	LocalLevels.insert_or_assign(bLevel, AutomapView);
 }
 
-void DeltaSyncObject(int oi, _cmd_id bCmd, const Player &player)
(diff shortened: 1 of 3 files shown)


Source/msg.cpp
```diff
@@ -3038,15 +2994,12 @@ size_t ParseCmd(int pnum, const TCmd *pCmd)
 	case CMD_PLRDAMAGE:
 		return OnPlayerDamage(pCmd, player);
 	case CMD_OPENDOOR:
-		return OnOpenDoor(pCmd, pnum);
 	case CMD_CLOSEDOOR:
```
AJenbo, line 2997: I know there was some funny business going on with the message being sent multiple times. Hope you made extra sure that doors aren't flapping as a consequence :)

AJenbo (approved):
Nice. I knew there was a lot that could be simplified in relations to how doors work, but didn't even realize that also applied to the network side of things :)

PR #5067, opened 2022-07-19
Change exp member to uint32 to fix Diablo exp going overbounds

aetn23:
Changed exp member to uint32_t from uint16_t in monster data and monster. This is done to fix a bug where diablo exp goes overbounds on hell difficulty.

I make it as a draft, since it breaks previous saves and I don't know how to supply backwards compatibility. 

Also, I don't know if this change even makes sense, since as far as I know only diablo's exp behaves like that on hell, so maybe a simple fallback to max value of unit16_t is better?

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -638,7 +638,7 @@ void LoadMonster(LoadHelper *file, Monster &monster)
 	monster.whoHit = file->NextLE<int8_t>();
 	monster.level = file->NextLE<int8_t>();
 	file->Skip(1); // Alignment
-	monster.exp = file->NextLE<uint16_t>();
+	monster.exp = file->NextLE<uint32_t>();
 
 	if ((monster.flags & MFLAG_GOLEM) != 0) // Don't skip for golems
 		monster.toHit = file->NextLE<uint8_t>();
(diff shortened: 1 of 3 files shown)


Source/loadsave.cpp
```diff
@@ -638,7 +638,7 @@ void LoadMonster(LoadHelper *file, Monster &monster)
 	monster.whoHit = file->NextLE<int8_t>();
 	monster.level = file->NextLE<int8_t>();
 	file->Skip(1); // Alignment
-	monster.exp = file->NextLE<uint16_t>();
+	monster.exp = file->NextLE<uint32_t>();
```
ephphatha, line 641: you need to ensure that the same number of bytes are read/written to save files for compatibility with old versions and the base game. You could work around the issue by splitting the value into two locations or writing a truncated value here and a full value somewhere else.

e.g.
```cpp
// To allow for values which overflow the old 16bit space allocated to monster exp we store an extra 8 bits in a previously unused space in the file
monster.exp = file->NextLE<uint8_t>() << 16;
monster.exp |= file->NextLE<uint16_t>();
```

(and do the inverse when saving)

qndel: since diablo is the only monster that overflows it, I'd suggest just limiting his exp to max possible value, otherwise it's an overkill and who kills hell diablo for xp xD

DakkJaniels:
You can't cast to int32 for the exp calculation?

aetn23:
Makes sense to me. I will handle it in a separate PR.

PR #5068, opened 2022-07-19
Load PCX as CL2

glebm:
Similar to #5059, which converted CEL to CL2 at load time, we now do the same for PCX.

Some size stats: https://gist.github.com/glebm/067bf1ec73f9d14514932cfe237e4e8e

Notably, fonts RAM usage is decreased by ~35%.

The names of structs in `cel_sprite.hpp` have become a bit confusing with the addition of basic sheets but we can clean that up at a later point.


AJenbo:
> Notably, fonts RAM usage is decreased by ~35%.

That's awesome, I was just feeling sad today that they where different :D

Feel free to add support for loading from PNG as well :sweat_smile: 
(Touch UI should still remain 32bit SDL2 textures, so not to replace that)

glebm:
I don't think DevilutionX should convert our own images at runtime.

Instead, we should have tools (probably in a separate repo) to convert everything to CL2, and a tool to convert CL2 to PCX (for editing). Additionally that tools repo could have a tool to compress an existing diabdat.mpq (convert everything to CL2, optionally convert audio to MP3).

AJenbo:
> Additionally that tools repo could have a tool to compress an existing diabdat.mpq

As long as this is an optional step :)

Source/utils/pcx_to_cl2.hpp
```diff
@@ -0,0 +1,21 @@
...
+#include <SDL.h>
+
+#include "engine/cel_sprite.hpp"
+#include "utils/stdcompat/optional.hpp"
+
+namespace devilution {
+
+/** @brief Loads a PCX file as a CL2 sprite.
+ *
+ *
```
AJenbo, line 14:
```suggestion
 *
```

Source/utils/pcx_to_cl2.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+	// We process the PCX a whole frame at a time because the lines are reversed in CEL.
+	auto frameBuffer = std::unique_ptr<uint8_t[]>(new uint8_t[static_cast<size_t>(frameHeight) * width]);
+
+	const unsigned srcSkip = width % 2;
+	uint8_t *dataPtr = fileBuffer.get();
+	for (unsigned frame = 1; frame <= numFrames; ++frame) {
+		WriteLE32(&cl2Data[4 * static_cast<size_t>(frame)], static_cast<uint32_t>(cl2Data.size()));
+
+		// Frame header: 5 16-bit offsets to 32-pixel height blocks.
```
AJenbo, line 162: Regarding converting assets to Cl2 offline. If we on load decode the first block and divide by 32 then we would know the width? Would we need to also decode the full image to know the height?
I think pngs2cel might now encode this, but that can be ignored.

glebm: > If we on load decode the first block and divide by 32 then we would know the width? 

Yes but the width calculation doesn't work for CL2 files that are fewer than 32 lines tall.

> Would we need to also decode the full image to know the height?

Yes.

Let's keep the more general discussion outside of this PR (e.g. in the issue), unless you think it affects it somehow?

Source/engine/render/cl2_render.hpp
```diff
@@ -65,6 +73,14 @@ void Cl2DrawOutlineSkipColorZero(const Surface &out, uint8_t col, Point position
  */
 void Cl2DrawTRN(const Surface &out, Point position, CelSprite cel, int frame, uint8_t *trn);
 
+/**
+ * @brief Same as Cl2DrawTRN but position.y is the top of the sprite instead of the bottom.
+ */
+inline void RenderCl2SpriteWithTRN(const Surface &out, CelFrameWithHeight cel, Point position, uint8_t *trn)
```
AJenbo, line 79: Should we adapt the code to position them from the bottom? Or change rendering of other sprites?

glebm: This PR only switches from PCX to CL2.
The code that rendered PCX required height.

In the future, we can either:

1. Require that all sprites have height (that means hardcoding the height as well).
2. Migrate the few places that need height to position from the bottom. Note that, for fonts, we'll need to know the height for regardless but that's easy to hardcode as there is only a few font sizes.

Changing that and the decision on that is perhaps out of scope for this PR.

AJenbo: I REALLY do not like the idea of hard-coding the dimensions. If we go down this route we need to figure out a way to determiner the size instead.

AJenbo (approved):

PR #5069, opened 2022-07-19
:white_check_mark: Add test for RotateBlockedMissile

canassa:
Just a simple test for RotateBlockedMissile

### DIFF
--- a/test/missiles_test.cpp
+++ b/test/missiles_test.cpp
@@ -16,7 +16,7 @@ TEST(Missiles, RotateBlockedMissileArrow)
 	TestRotateBlockedMissile(missile);
 	EXPECT_EQ(missile._miAnimFrame, 16);
 	
-	SetRndSeed(/*find a seed which rotates the other direction*/);
+	SetRndSeed(3210);
 	TestRotateBlockedMissile(missile);
 	EXPECT_EQ(missile._miAnimFrame, 1);
 	
--- a/test/missiles_test.cpp
+++ b/test/missiles_test.cpp
@@ -7,12 +7,23 @@ using namespace devilution;
 
 TEST(Missiles, RotateBlockedMissileArrow)
 {
-	SetRndSeed(0);
 	Player &player = Players[0];
-	auto *missile = AddMissile({ 0, 0 }, { 0, 0 }, Direction::South, MIS_ARROW, TARGET_MONSTERS, player.getId(), 0, 0);
-	EXPECT_EQ(missile->_miAnimFrame, 1);
-	TestRotateBlockedMissile(*missile);
-	EXPECT_EQ(missile->_miAnimFrame, 16);
+	// missile can be a copy or a reference, there's no nullptr check and the functions that use it don't expect the instance to be part of a global structure so it doesn't really matter for this use.
+	Missile missile = *AddMissile({ 0, 0 }, { 0, 0 }, Direction::South, MIS_ARROW, TARGET_MONSTERS, player.getId(), 0, 0);
+	EXPECT_EQ(missile._miAnimFrame, 1);
+	
+	SetRndSeed(0);
+	TestRotateBlockedMissile(missile);
+	EXPECT_EQ(missile._miAnimFrame, 16);
+	
+	SetRndSeed(/*find a seed which rotates the other direction*/);
+	TestRotateBlockedMissile(missile);
+	EXPECT_EQ(missile._miAnimFrame, 1);
+	
+	missile._miAnimFrame = 5;
+	SetRndSeed(1234); // no idea what rotation this actually performs, might need to change the expectation to 4 instead of 6
+	TestRotateBlockedMissile(missile);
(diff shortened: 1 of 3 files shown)


test/missiles_test.cpp
```diff
@@ -4,6 +4,15 @@
 
 using namespace devilution;
 
+TEST(Missiles, RotateBlockedMissileArrow)
+{
+	Player &player = Players[0];
+	auto *missile = AddMissile({ 0, 0 }, { 0, 0 }, Direction::South, MIS_ARROW, TARGET_MONSTERS, player.getId(), 0, 0);
+	EXPECT_EQ(missile->_miAnimFrame, 1);
+	TestRotateBlockedMissile(*missile);
```
ephphatha, line 12: This relies on an RNG call, you'll need to seed the engine before calling TestRotateBlockedMissile to make it consistent. See drlg_l*_test, quests_test, or random_test for examples.

canassa: Thanks!

test/missiles_test.cpp
```diff
@@ -1,9 +1,20 @@
...
 
+TEST(Missiles, RotateBlockedMissileArrow)
+{
+	SetRndSeed(0);
+	Player &player = Players[0];
+	auto *missile = AddMissile({ 0, 0 }, { 0, 0 }, Direction::South, MIS_ARROW, TARGET_MONSTERS, player.getId(), 0, 0);
+	EXPECT_EQ(missile->_miAnimFrame, 1);
+	TestRotateBlockedMissile(*missile);
+	EXPECT_EQ(missile->_miAnimFrame, 16);
+}
```
ephphatha, line 16: The name of this test is fine in that it only tests the case where `animType == MFILE_ARROWS`, however expanding the tests to also capture overflow and a regular non-wrapping rotation would be good.

Also should try structure tests so the broad setup happens first, then test setup immediately before the relevant call.

```suggestion
TEST(Missiles, RotateBlockedMissileArrow)
{
	SetRndSeed(0);
	Player &player = Players[0];
	// missile can be a copy or a reference, there's no nullptr check and the functions that use it don't expect the instance to be part of a global structure so it doesn't really matter for this use.
	Missile missile = *AddMissile({ 0, 0 }, { 0, 0 }, Direction::South, MIS_ARROW, TARGET_MONSTERS, player.getId(), 0, 0);
	EXPECT_EQ(missile._miAnimFrame, 1);
	
	SetRndSeed(0);
	TestRotateBlockedMissile(missile);
	EXPECT_EQ(missile._miAnimFrame, 16);
	
	SetRndSeed(/*find a seed which rotates the other direction*/);
	TestRotateBlockedMissile(missile);
	EXPECT_EQ(missile._miAnimFrame, 1);
	
	missile._miAnimFrame = 5;
	SetRndSeed(1234); // no idea what rotation this actually performs, might need to change the expectation to 4 instead of 6
	TestRotateBlockedMissile(missile);
	EXPECT_EQ(missile._miAnimFrame, 6);
}
```

test/missiles_test.cpp
```diff
@@ -1,9 +1,31 @@
...
+	Player &player = Players[0];
+	// missile can be a copy or a reference, there's no nullptr check and the functions that use it don't expect the instance to be part of a global structure so it doesn't really matter for this use.
+	Missile missile = *AddMissile({ 0, 0 }, { 0, 0 }, Direction::South, MIS_ARROW, TARGET_MONSTERS, player.getId(), 0, 0);
+	EXPECT_EQ(missile._miAnimFrame, 1);
+	
+	SetRndSeed(0);
+	TestRotateBlockedMissile(missile);
+	EXPECT_EQ(missile._miAnimFrame, 16);
+	
+	SetRndSeed(/*find a seed which rotates the other direction*/);
```
AJenbo, line 19: Lets try just so we can start to get some ouput :D
```suggestion
	SetRndSeed(3210);
```

AJenbo:
@ephphatha did i just randomly guess a valid seed on first try?

test/missiles_test.cpp
```diff
@@ -1,9 +1,31 @@
...
+	SetRndSeed(0);
+	TestRotateBlockedMissile(missile);
+	EXPECT_EQ(missile._miAnimFrame, 16);
+	
+	SetRndSeed(3210);
+	TestRotateBlockedMissile(missile);
+	EXPECT_EQ(missile._miAnimFrame, 1);
+	
+	missile._miAnimFrame = 5;
+	SetRndSeed(1234); // no idea what rotation this actually performs, might need to change the expectation to 4 instead of 6
```
AJenbo, line 24:
```suggestion
	SetRndSeed(1234);
```

AJenbo:
Need to do a clang-format dance. Other then that I think this is ready.

AJenbo (approved):

AJenbo:
Nice, it's always good to have new tests :)

PR #5070, opened 2022-07-19
Exp overflow fix

aetn23:
Move monster exp calculation to `MonsterDeath`. This is done to prevent a bug where monster exp could overflow on hell if the monster is Diablo (per #5067). 

I know that right now `exp` member could be removed from monster, but I am working on a PR that will do that, among other things.

EDIT: removed `exp` field from `monster` since it simplifies things (no need to calculate the `exp` while loading)


Source/monster.cpp
```diff
@@ -3891,8 +3886,16 @@ void M_StartHit(Monster &monster, const Player &player, int dam)
 
 void MonsterDeath(Monster &monster, Direction md, bool sendmsg)
 {
+	int monsterExp = sgGameInitInfo.nDifficulty == DIFF_NIGHTMARE ? 2 * (monster.exp + 1000) : sgGameInitInfo.nDifficulty == DIFF_HELL ? 4 * (monster.exp + 1000)
+	                                                                                                                                   : monster.exp;
+	if (monster.isUnique()) {
+		monsterExp *= 2;
```
DakkJaniels, line 3892: Should this bug be fixed? Probably wasn't intended to run the difficulty adjustment again for unique monsters 🤷🏽‍♂️

qndel: + you should leave old values for vanilla compatibility or monsters will give 0 xp when you load devx save in vanilla xD

aetn23:
I know that, I will handle it in a separate PR that I am working on that will do that and remove some other monster's members

Source/monster.cpp
DakkJaniels: @qndel is this a response to my other comment? If so, that's why I was saying the old values (from the array) should be saved to the save file, but we don't need to reload them in DevX.

Source/monster.cpp
```diff
@@ -3884,18 +3883,23 @@ void M_StartHit(Monster &monster, const Player &player, int dam)
 	M_StartHit(monster, dam);
 }
 
-void MonsterDeath(Monster &monster, Direction md, bool sendmsg)
+int CalculateMonsterExp(uint16_t baseExp, bool isUnique)
 {
-	int monsterExp = sgGameInitInfo.nDifficulty == DIFF_NIGHTMARE ? 2 * (monster.exp + 1000) : sgGameInitInfo.nDifficulty == DIFF_HELL ? 4 * (monster.exp + 1000)
-	                                                                                                                                   : monster.exp;
-	if (monster.isUnique()) {
+	int monsterExp = sgGameInitInfo.nDifficulty == DIFF_NIGHTMARE ? 2 * (baseExp + 1000) : sgGameInitInfo.nDifficulty == DIFF_HELL ? 4 * (baseExp + 1000)
```
DakkJaniels, line 3888: I'm fine either way, but would switch/case be more readable here?

Source/monstdat.cpp
```diff
@@ -25,7 +25,7 @@ constexpr uint16_t Uniq(_unique_items item)
...
+// _monster_id   name,                                       assetsSuffix,        soundSuffix,          trnFile,           availability,                 width, image, hasSpecial, hasSpecialSound,   frames[6],                  rate[6],                  minDunLvl, maxDunLvl, level, hitPointsMinimum, hitPointsMaximum, ai,           abilityFlags,                          intelligence, toHit, animFrameNum, minDamage, maxDamage, toHitSpecial, animFrameNumSpecial, minDamageSpecial, maxDamageSpecial, armorClass, monsterClass, resistance,                                                resistanceHell,                                                  selectionType,     treasure,      expBase
```
DakkJaniels, line 28: shouldn't this stay as exp, since that's what it is called in MonsterData?

aetn23:
I removed the function to calculate exp, and simply calculate it in the monster `exp` member, since it is more readable.

aetn23:
If the general consensus is to fix the unique monster overflow bug, let me know and I will quickly fix it. If there is no consensus, maybe it's better to merge this to master, and start new PR with unique monster exp fix?

ephphatha:
> If the general consensus is to fix the unique monster overflow bug, let me know and I will quickly fix it. If there is no consensus, maybe it's better to merge this to master, and start new PR with unique monster exp fix?

If you want to do a quick interim fix capping the exp value when performing the calculation that can potentially overflow is going to be far easier to review and reason about, might as well do that in a separate PR and we can leave this one for a proper fix.

Source/monster.h
```diff
@@ -307,6 +307,29 @@ struct Monster { // note: missing field _mAFNum
 		return pgettext("monster", data().name);
 	}
 
+	int exp(_difficulty difficulty) const
+	{
+		int monsterExp = data().exp;
```
ephphatha, line 312: This should be unsigned, negative experience values aren't valid.

Source/loadsave.cpp
```diff
@@ -1388,7 +1389,7 @@ void SaveMonster(SaveHelper *file, Monster &monster)
 	file->WriteLE<int8_t>(monster.whoHit);
 	file->WriteLE<int8_t>(monster.level);
 	file->Skip(1); // Alignment
-	file->WriteLE<uint16_t>(monster.exp);
+	file->WriteLE<uint16_t>(monster.exp(sgGameInitInfo.nDifficulty));
```
ephphatha, line 1392: Should take this opportunity to fix the overflow when writing to the save file.

```suggestion
	file->WriteLE<uint16_t>(static_cast<uint16_t>(std::min<unsigned>(std::numeric_limits<uint16_t>::max(), monster.exp(sgGameInitInfo.nDifficulty))));
```

Source/loadsave.cpp
```diff
@@ -638,7 +639,7 @@ void LoadMonster(LoadHelper *file, Monster &monster)
 	monster.whoHit = file->NextLE<int8_t>();
 	monster.level = file->NextLE<int8_t>();
 	file->Skip(1); // Alignment
-	monster.exp = file->NextLE<uint16_t>();
+	file->Skip(2); // Skip exp - the monster used to have this field
```
ephphatha, line 642:
```suggestion
	file->Skip(2); // Skip exp - now calculated from monstdat when the monster dies
```

aetn23:
> If you want to do a quick interim fix capping the exp value when performing the calculation that can potentially overflow is going to be far easier to review and reason about, might as well do that in a separate PR and we can leave this one for a proper fix.

Okay, I think it's better to come up with a proper fix in this PR then. 

It seems like removing double unique monster calculation is a way to go, per reasons @DakkJaniels mentioned.

Source/monster.h
```diff
@@ -307,9 +307,9 @@ struct Monster { // note: missing field _mAFNum
 		return pgettext("monster", data().name);
 	}
 
-	int exp(_difficulty difficulty) const
+	uint32_t exp(_difficulty difficulty) const
 	{
-		int monsterExp = data().exp;
+		unit32_t monsterExp = data().exp;
```
ephphatha: `unsigned` is fine, this doesn't need to be a fixed width field. It'll eventually get added to `Player::experience` which is `uint32_t` but that's after a division and cap has been applied, and the fixed width type there is because it's a persistent data structure. For function return values it's not critical to specify the size.

aetn23:
I removed double unique monster exp calculation and commented exp function

Source/monster.cpp
```diff
@@ -3659,8 +3653,7 @@ void M_StartHit(Monster &monster, const Player &player, int dam)
 
 void MonsterDeath(Monster &monster, Direction md, bool sendmsg)
 {
-	if (!monster.isPlayerMinion())
```
qndel, line 3662: why did you remove this?

aetn23: It was removed by accident during merging, it is now reverted to how it should be

Source/monster.h
```diff
@@ -317,6 +317,29 @@ struct Monster { // note: missing field _mAFNum
 		return pgettext("monster", data().name);
 	}
 
+	/**
+	 * @brief Calculates monster's experience points.
+	 * Fetches base exp value from @p MonstersData array.
```
qndel, line 322:
```cpp
	 * Fetches base exp value from @p MonstersData array.
```
what's `@p` ?

aetn23: `@p` makes the word after it appear in typewriter font

qndel (approved):

PR #5072, opened 2022-07-19
:recycle: Remove monsterId from StartDeathFromMonster

canassa:


Source/monster.cpp
```diff
@@ -1067,24 +1067,19 @@ void MonsterHitMonster(Monster &monster, int i, int dam)
 	HitMonster(monster, dam);
 }
 
-void StartDeathFromMonster(int i, int mid)
+void StartDeathFromMonster(Monster &killerMonster, Monster &killedMonster)
```
ephphatha, line 1070: having variables names which differ by only a single letter is very confusing, `attacker` and `target` like other functions will be fine, we already know they're `Monster` instances due to the variable type.

Source/monster.cpp
```diff
@@ -1067,24 +1067,19 @@ void MonsterHitMonster(Monster &monster, int i, int dam)
...
-
-	delta_kill_monster(monster, monster.position.tile, *MyPlayer);
-	NetSendCmdLocParam1(false, CMD_MONSTDEATH, monster.position.tile, mid);
+	delta_kill_monster(killedMonster, killedMonster.position.tile, *MyPlayer);
+	NetSendCmdLocParam1(false, CMD_MONSTDEATH, killedMonster.position.tile, killedMonster.getId());
 
-	if (killer.type().type == MT_GOLEM)
-		monster.whoHit |= 1 << i;
+	if (killerMonster.type().type == MT_GOLEM)
+		killedMonster.whoHit |= 1 << killerMonster.getId();
```
ephphatha, line 1076: This is the owning player ID, while the monster network ID is equivalent in the current code it would be good to differentiate this from the concept of a monster network ID

Source/monster.cpp
canassa: Thanks, attacker and target are much better names indeed 😄

Source/monster.cpp
canassa: Shall I create a separate issue for this? I agree with the comment but I think that changing that is outside the scope of the PR as the getId() is already being used as the network id in order places now

ephphatha: This specific instance is different to other uses of Monster::getId. At least a comment calling out the difference would be good. It'll also need to be replicated in MonsterHitMonster as these two functions are the only instances where a player is given credit for a killing blow from their golem.

canassa: Understood. I will update the PR

canassa: I added getNetworkId, although I am not sure I understood 100% what you meant. Should calls like this all be updated to use `getNetworkId()`? 

```c++
NetSendCmdLocParam1(false, CMD_MONSTDEATH, target.position.tile, target.getId());
```

Source/monster.h
```diff
@@ -308,12 +308,19 @@ struct Monster { // note: missing field _mAFNum
...
 	 * This is currently the index into the Monsters array, but may change in the future.
 	 */
 	[[nodiscard]] size_t getId() const;
 
+	/**
+	 * @brief Returns the network identifier for this monster
+	 *
+	 * This is currently the index into the Monsters array, but may change in the future.
+	 */
+	[[nodiscard]] size_t getNetworkId() const;
```
ephphatha, line 322: Wrong way round. `Monster::getId()` *is* the network/lookup ID for a monster instance. The use in leashed monsters is consistent with the use in network actions so the more generic name is suitable for that function.

For the use in StartDeathFromMonster and MonsterHitMonster the important concepts are "is this monster controlled by a player" and "who is the controlling player". This function addresses the second requirement.
```suggestion
	/**
	 * @brief Returns the identifier for the player who owns/controls this monster
	 *
	 * In Diablo 1 players can only control a single monster. The first four slots of the Monsters array are reserved for
	 * player golems so we currently reuse the monster network id here, but provide a separate function to allow for
	 * possible future changes.
	 */
	[[nodiscard]] size_t getOwningPlayerId() const;
```

AJenbo: I don't think I really favor this since it would be easy to assume that the return value will be blank for monsters that aren't controlled.

ephphatha: Yeah, I was basing this implementation on the missile source by reference PR with the same assumption that every use would be guarded. It can return an optional instead but at that point this function really needs to be addressed in its own PR.

Source/monster.cpp
ephphatha: Lets just revert this back to `target.whoHit |= 1 << attacker.getId(); // really the id the player who controls this golem`

Can add helpers to make the use stand out later.

canassa: Thanks, I updated it

Source/monster.h
canassa: The change was reverted, I am resolving this one

ephphatha (approved):

Source/monster.cpp
AJenbo: the comments solves the misgivings i had :+1:

PR #5080, opened 2022-07-20
Tidy up ActivateSkeleton and take monster by reference

ephphatha:


Source/monster.cpp
```diff
@@ -4653,56 +4653,40 @@ bool IsGoat(_monster_id mt)
...
-			if (x2 == 3) {
-				x2 = 0;
-				y2++;
-				if (y2 == 3)
-					y2 = 0;
-			}
-		}
-	}
+	// this is used in the following loop to find the nth set bit.
+	int spawnChoice = GenerateRnd(15) % spawnOk.count();
```
AJenbo, line 4676: Maybe we could also use count to clean up https://github.com/diasurgical/devilutionX/blob/master/Source/levels/drlg_l1.cpp#L545 and it's siblings

AJenbo (approved):
nice, the old code was really ugly :)

PR #5081, opened 2022-07-20
Use object references in OperateLever/Chest

ephphatha:
OperateChest was an incidental inclusion... I can split that into a different PR if it helps

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -2334,32 +2334,32 @@ void OperateChamberOfBoneBook(Object &questBook)
 	InitQTextMsg(textdef);
 }
 
-void OperateChest(const Player &player, int i, bool sendmsg)
+void OperateChest(const Player &player, Object &chest, bool sendLootMsg)
 {
-	if (Objects[i]._oSelFlag == 0) {
+	if (chest._oSelFlag == 0) {
 		return;
 	}
 
-	PlaySfxLoc(IS_CHEST, Objects[i].position);
-	Objects[i]._oSelFlag = 0;
-	Objects[i]._oAnimFrame += 2;
-	SetRndSeed(Objects[i]._oRndSeed);
+	PlaySfxLoc(IS_CHEST, chest.position);
+	chest._oSelFlag = 0;
+	chest._oAnimFrame += 2;
+	SetRndSeed(chest._oRndSeed);
 	if (setlevel) {
-		for (int j = 0; j < Objects[i]._oVar1; j++) {
-			CreateRndItem(Objects[i].position, true, sendmsg, false);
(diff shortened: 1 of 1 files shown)


Source/quests.h
```diff
@@ -58,6 +58,13 @@ struct Quest {
 	uint8_t _qvar2;
 
 	bool IsAvailable();
+
+	void activateMultiStageQuest(uint8_t quest_stage = 1)
+	{
+		_qactive = QUEST_ACTIVE;
+		_qlog = true;
+		_qvar1 = quest_stage;
```
AJenbo, line 66: The paramter feels a bit odd, since it is only used by two quests. I think it would be clearer to simply update it immediately after calling this.

Source/quests.h
```diff
@@ -58,6 +58,13 @@ struct Quest {
 	uint8_t _qvar2;
 
 	bool IsAvailable();
+
+	void activateMultiStageQuest(uint8_t quest_stage = 1)
```
AJenbo, line 62: Are there single stage quests, do they need to be activated differently?

ephphatha: There seem to be a few different types of quests. Some like Q_ANVIL and Q_ROCK are made active independently of being marked visible in the log, Q_GRAVE uses qvar1 but appears to progress before the quest is made active in some cases. Q_GARBUD, Q_ZHAR, Q_VEIL (and possibly others) don't use qvar1. Q_SKELKING uses qvar1 but isn't marked visible in the log when it's made active.

Source/quests.h
ephphatha: were you thinking we make this function always set qvar1 to 1, then add the assignment to 2 immediately after for the two quests which use it? Or just make this function `activate() { _qactive = ACTIVE; _qlog=true; }` and set qvar1 explicitly for every use?

AJenbo: > were you thinking we make this function always set qvar1 to 1, then add the assignment to 2 immediately after for the two quests which use it?

Yeah that was my thinking. But I haven't had time to sit down and read the code so it might be miss guided if this isn't a common pattern

ephphatha: I'd actually done it that way originally then decided it looked nicer to have the default argument instead. That said it's an inline function anyway so compilers should be able to get rid of the first assignment to qvar1 pretty easily.

Source/quests.h
ephphatha: Sorry to break the conversation, I ended up splitting this change into #5085 so this one can be focused on the last couple of non-door OperateObject functions.

AJenbo (approved):

AJenbo:
Thanks for splitting I was just considering that as an option until I have better time to look at the quest part :)

PR #5083, opened 2022-07-22
Add Bitset2d::count() and use it when counting set Dungeonmask tiles

ephphatha:
from https://github.com/diasurgical/devilutionX/pull/5080#discussion_r926908430

### DIFF
--- a/Source/levels/drlg_l4.cpp
+++ b/Source/levels/drlg_l4.cpp
@@ -174,8 +174,8 @@ void InitDungeonFlags()
 
 void MapRoom(Rectangle room)
 {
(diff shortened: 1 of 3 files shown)


Source/levels/drlg_l4.cpp
```diff
@@ -843,7 +846,10 @@ void Substitution()
 	}
 }
 
-void UShape()
+/**
+ * @brief Sets up the inside borders of the first quadrant so there are valid paths after mirroring the layout
+ */
+void PrepareInnerBorders()
```
AJenbo, line 852: Good to know :)

PR #5084, opened 2022-07-22
use correct attribute for Bitset2d::count

ephphatha:
Should've read what I was typing...


glebm (approved):

AJenbo:
Oh that explains why we started getting warning for them 😅

PR #5086, opened 2022-07-22
Take references in operate door functions

ephphatha:
Bit of a monolith, it is possible to split this further but the use in `MonsterCheckDoors` ties them all together so that chunk gets changed in pretty much every commit. Next step is changing `CMD_OPENDOOR` and `CMD_CLOSEDOOR` to use position instead of ID in the network message, planning that for a followup so it's easier to test and be confident doors still work.


AJenbo (approved):
good size pr :+1:

PR #5098, opened 2022-07-23
:recycle: Remove _misource access from AddLightningWall

canassa:


Source/missiles.cpp
```diff
@@ -1375,12 +1375,18 @@ void AddLightningWall(Missile &missile, const AddMissileParameter &parameter)
...
-		missile.var1 = Players[missile._misource].position.tile.x;
-		missile.var2 = Players[missile._misource].position.tile.y;
+		break;
+	case MissileSource::Player: {
+		Player &player = *missile.sourcePlayer();
+		missile.var1 = player.position.tile.x;
+		missile.var2 = player.position.tile.y;
+	} break;
+	case MissileSource::Monster:
+		break;
```
canassa, line 1389: Perhaps it should raise an error instead?

ephphatha: exceptions are discouraged, but you could add a debug assert to document the expectation that this is never hit.

canassa: Something like this?

```c++
	case MissileSource::Monster:
		assert(false); // Monster source not supported
		break;
```

ephphatha: yeah that'd work, maybe make the comment "Monsters aren't expected to cast lightning wall"

canassa: Done. Thanks.

ephphatha (approved):

Source/missiles.cpp
```diff
@@ -1375,12 +1375,19 @@ void AddLightningWall(Missile &missile, const AddMissileParameter &parameter)
...
-		missile.var1 = Players[missile._misource].position.tile.x;
-		missile.var2 = Players[missile._misource].position.tile.y;
+		break;
+	case MissileSource::Player: {
+		Player &player = *missile.sourcePlayer();
+		missile.var1 = player.position.tile.x;
+		missile.var2 = player.position.tile.y;
+	} break;
+	case MissileSource::Monster:
+		assert(false); // Monsters aren't expected to cast lightning wall
```
AJenbo, line 1389: Good idea but better to make the assert `!= MissileSource::Monster` so that we would get a useful error, that would also remove the need for the comments.

canassa: Done

AJenbo (approved):

Source/missiles.cpp
```diff
@@ -1375,12 +1375,19 @@ void AddLightningWall(Missile &missile, const AddMissileParameter &parameter)
...
-		missile.var1 = Players[missile._misource].position.tile.x;
-		missile.var2 = Players[missile._misource].position.tile.y;
+		break;
+	case MissileSource::Player: {
+		Player &player = *missile.sourcePlayer();
+		missile.var1 = player.position.tile.x;
+		missile.var2 = player.position.tile.y;
+	} break;
+	case MissileSource::Monster:
+		assert(missile.sourceType() != MissileSource::Monster);
```
qndel, line 1389: I liked the one with false, if we got here, then we don't really have to check this :D

AJenbo: The problem is that you get no context in the error message:
image

If we hit this code we are already in a place we shouldn't be

canassa:
@AJenbo This merge was blocked due to the vita errors. I can force push a commit to rerun the pipeline but that will remove your approval. What do you prefer me to do?

PR #5100, opened 2022-07-23
Simplify MonsterOperateDoor

ephphatha:
No real need to loop over all level objects when we already check based on position, can just use ObjectAtPosition instead.

It could also use `considerLargeObjects = false` but doors are single tile objects anyway so I don't think it's necessary, makes the call simpler.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -4698,38 +4698,38 @@ void MonstCheckDoors(Monster &monster)
 	    || dObject[mx][my + 1] != 0
 	    || dObject[mx + 1][my + 1] != 0) {
 		for (int i = 0; i < ActiveObjectCount; i++) {
-			Object &object = Objects[ActiveObjects[i]];
-			if ((object._otype == OBJ_L1LDOOR || object._otype == OBJ_L1RDOOR) && object._oVar4 == 0) {
-				int dpx = abs(object.position.x - mx);
-				int dpy = abs(object.position.y - my);
-				if (dpx == 1 && dpy <= 1 && object._otype == OBJ_L1LDOOR)
-					OperateL1LDoor(object, true);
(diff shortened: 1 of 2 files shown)


Source/objects.cpp
```diff
@@ -4685,51 +4685,43 @@ void RedoPlayerVision()
...
+			continue;
+
+		Object &door = *object;
+		// Doors use _oVar4 to track open/closed state, non-zero values indicate an open door
+		// we don't use Object::isDoor since the later conditions will check the object is actually a door anyway.
+		if (door._oVar4 != 0)
+			continue;
+
+		// Strictly speaking these checks shouldn't be required, they're guarding against monsters
+		// standing in a wall opening an adjacent door. Shouldn't be possible?
```
AJenbo, line 4702: Hm I think you are right, even the doors in Valor are spaced one tile apart. Previously I thought it was to make sure that the monster was standing in front of the door, but that would have been `x == 0 && Y == 1`

Source/objects.cpp
```diff
@@ -4685,51 +4685,43 @@ void RedoPlayerVision()
...
+		if (IsNoneOf(dir, Direction::NorthEast, Direction::SouthWest)) {
+			if (door._otype == OBJ_L1LDOOR) {
+				OperateL1LDoor(door, true);
+			} else if (door._otype == OBJ_L2LDOOR) {
+				OperateL2LDoor(door, true);
+			} else if (door._otype == OBJ_L3RDOOR) {
+				// L3 doors are backwards, see OperateL3Door
```
AJenbo, line 4709: We should probably rename the enum an functions to avoid the confusion :)

AJenbo (approved):
This is one of the cases where the original turns out to be even more messy then you initially thought :)
Nice cleanup.

PR #5101, opened 2022-07-23
CLX: A new graphics format

### LINKED ISSUE #5176: Debug command r_drawvision is broken
### Operating System

Windows x64

### DevilutionX version

### Describe
image

### To Reproduce

.

### Expected Behavior

_No response_

### Additional context

_No response_

glebm:
The format is almost identical to [CL2](https://github.com/savagesteel/d1-file-formats/blob/master/PC-Mac/CL2.md), except it uses the frame header to store frame width and height instead of 5 32-line offsets.

CLX frame header (10 bytes, same as CL2):

Bytes |   Type   | Value
:-----:|:--------:|------------------------------------
 0..2  | uint16_t | offset to frame pixel data (always 10, same as CL2)
 2..4  | uint16_t | width
 4..6  | uint16_t | height
6..10 |    -          | unused (set to 0)

The CLX format is otherwise identical to CL2.

This means we always have access to frame dimensions, so we can use it as an on-disk format for our graphics as well.

We now also skip the rendering entirely if the sprite is out of bounds vertically (thanks to knowing the height).

The new `ClxSprite*` classes provide typed representations of various types of sprites:

1. `ClxSprite` -- a single sprite.
2. `ClxSpriteList` -- a list of sprites.
3. `ClxSpriteSheet` -- a sheet (list of lists) of sprites, e.g. one list per direction.

`Owned` and `Optional` versions of the above are available and used throughout as well.

Also fixes #5176


AJenbo (commented):
Do you think we can make a GIMP plugin for CLX?

glebm:
I've never implemented a GIMP plugin but it should be possible! One issue is that we'd have to load the palette separately because CLX doesn't store one.

Source/qol/stash.cpp
```diff
@@ -370,15 +370,14 @@ void DrawStash(const Surface &out)
...
 
 		const Point position = GetStashSlotCoord(item.position) + offset;
-		const CelSprite cel { GetInvItemSprite(frame) };
-		const int celFrame = GetInvItemFrame(frame);
+		const ClxSprite sprite = GetInvItemSprite(frame);
 
 		if (pcursstashitem == itemId) {
 			uint8_t color = GetOutlineColor(item, true);
-			Cl2DrawOutline(out, color, position, cel, celFrame);
+			ClxDrawOutline(out, color, position, sprite);
```
AJenbo, line 377: Glad we only have 1 function for drawing outlines at this point.

Source/engine/load_file.hpp
```diff
@@ -132,20 +132,21 @@ struct MultiFileLoader {
 		StaticVector<SFile, MaxFiles> files;
 		StaticVector<uint32_t, MaxFiles> sizes;
 		size_t totalSize = 0;
-		for (size_t i = 0; i < numFiles; ++i) {
+		for (size_t i = 0, j = 0; i < numFiles; ++i) {
```
AJenbo, line 135: I'm failing to see where j is being defined?

julealgon:
Aren't there more standard formats that the game could use instead of creating yet another custom format on top of an incredibly custom and unknown Diablo-specific format?

It just feels to me that the more custom any aspect of the codebase becomes, the harder it will be for people to interface with it moving forward.

AJenbo:
Think of it more like unifying how the 3 original graphics formats are handled internally. Most formats are either meant for on-disk storage or GPU rendering. Neither which fits the bill here. Unifying run time handling also open the door for loading standard formats from disk (see `SurfaceToCl2`).

Source/engine/load_file.hpp
glebm: Right here, you can define any number of variables in a for loop.

glebm:
Yeah I've investigated a ton of formats and Diablo's CL2 was the most suitable in the RAM usage vs performance tradeoff.
CLX is a slight modification to it. Most RAM is taken by CL2 and we can convert them to CLX on the fly without reallocation.

Source/utils/surface_to_cl2.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "engine/clx_sprite.hpp"
+#include "engine/surface.hpp"
+
+namespace devilution {
+
+OwnedClxSpriteListWithFrameHeight SurfaceToCl2(const Surface &surface, unsigned numFrames,
```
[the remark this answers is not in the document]
glebm: Oh this file shouldn't be here, I had it temporarily based on my surface-to-cel branch but it is unused

Source/engine/load_file.hpp
AJenbo: oh wait it's a `,` :D

AJenbo:
> We now also skip the rendering entirely if the sprite is out of bounds vertically (thanks to knowing the height).

nice!

Source/objects.h
```diff
@@ -20,44 +21,48 @@ namespace devilution {
...
+	int _oVar5 = 0;
+	uint32_t _oVar6 = 0;
 	/**
 	 * @brief ID of a quest message to play when this object is activated.
 	 *
 	 * Used by spell book objects which trigger quest progress for Halls of the Blind, Valor, or Warlord of Blood
 	 */
-	_speech_id bookMessage;
-	int _oVar8;
+	_speech_id bookMessage = TEXT_KING1; // TEXT_NONE default breaks timedemo.
```
AJenbo, line 64: Funny this exact same issues was there for the monsters and was one of the more challenging things to test for 1.4.0 :)

Source/objects.h
```diff
@@ -20,44 +21,48 @@ namespace devilution {
...
+	uint32_t _oAnimFlag = 0;
+	OptionalClxSpriteList _oAnimData;
+	int _oAnimDelay = 0;      // Tick length of each frame in the current animation
+	int _oAnimCnt = 0;        // Increases by one each game tick, counting how close we are to _pAnimDelay
+	uint32_t _oAnimLen = 0;   // Number of frames in current animation
+	uint32_t _oAnimFrame = 0; // Current frame of animation.
+
+	// TODO: Remove this field, it is unused and always equal to:
+	// (*_oAnimData)[0].width()
+	uint16_t _oAnimWidth = 0;
```
AJenbo, line 36: Did you intend to do this before marking the PR as ready, or is there a reason it should be done later?

Source/engine/render/clx_render.cpp
```diff
@@ -435,78 +397,83 @@ void RenderCl2OutlineClippedXY(const Surface &out, Point position, RenderSrcBack
...
-			auto v = static_cast<uint8_t>(*dst++);
-			nDataSize--;
-			assert(nDataSize >= 0);
-			if (!IsCl2Opaque(v))
-				continue;
-			if (IsCl2OpaqueFill(v)) {
-				nDataSize--;
-				assert(nDataSize >= 0);
-				*dst = static_cast<byte>(ttbl[static_cast<uint8_t>(*dst)]);
+	// A bit of a hack but this is the only place in the code where we need mutable sprites.
```
AJenbo, line 426: 🤔

Source/panels/spell_book.cpp
```diff
@@ -81,9 +81,9 @@ spell_type GetSBookTrans(spell_id ii, bool townok)
 
 void InitSpellBook()
 {
-	pSpellBkCel = LoadCelAsCl2("Data\\SpellBk.CEL", static_cast<uint16_t>(SidePanelSize.width));
-	pSBkBtnCel = LoadCelAsCl2("Data\\SpellBkB.CEL", gbIsHellfire ? 61 : 76);
-	pSBkIconCels = LoadCelAsCl2("Data\\SpellI2.CEL", 37);
+	pSpellBkCel = LoadCelAsClx("Data\\SpellBk.CEL", static_cast<uint16_t>(SidePanelSize.width));
+	pSBkBtnCel = LoadCelAsClx("Data\\SpellBkB.CEL", gbIsHellfire ? 61 : 76);
+	pSBkIconCels = LoadCelAsClx("Data\\SpellI2.CEL", 37);
```
AJenbo, line 86: Nit: At this point it would be implied that the graphics are being loaded as clx.

Source/objects.h
glebm: It should probably be done in a separate PR where we can also remove a few other fields.
Some extra work will be needed for `HeadlessMode` compatibility, which currently relies on this field as we don't actually load the sprites themselves.

Source/objects.h
glebm: Let's update the save game in a follow-up PR!

Source/panels/spell_book.cpp
glebm: Shall I remove all the `AsClx` suffixes throughout?

Source/engine/render/clx_render.cpp
glebm: I wasn't sure whether to introduce mutable sprites just for this one function.

Currently sprites store `const uint8_t *`, we could either make them mutable throughout (`uint8_t *`) or have separate mutable classes and conversions from mutable to const.

Other than this function, nothing else needs non-const `uint8_t *`.

Source/panels/spell_book.cpp
AJenbo: if you agree that it would be cleaner :)

glebm: Done!

Source/monster.cpp
```diff
@@ -156,19 +157,13 @@ size_t GetNumAnims(const MonsterData &monsterData)
...
-
 void InitMonsterTRN(CMonster &monst)
 {
 	char path[64];
-	std::array<uint8_t, 256> colorTranslations;
 	*BufCopy(path, "Monsters\\", monst.data->trnFile, ".TRN") = '\0';
-	LoadFileInMem(path, colorTranslations);
-
-	std::replace(colorTranslations.begin(), colorTranslations.end(), 255, 0);
+	uint8_t colorTranslations[256];
```
AJenbo, line 164: why are we using a c-style array?

glebm: No reason, now an `std::array` again

Source/utils/intrusive_optional.hpp
```diff
@@ -0,0 +1,81 @@
+#pragma once
+
+#include <type_traits>
+#include <utility>
+
+#include "appfat.h"
+#include "utils/stdcompat/optional.hpp"
+
+/// An optional that uses a field of the stored class and some value to store nullopt.
+#define DEFINE_INTRUSIVE_OPTIONAL_IMPL(OPTIONAL_CLASS, VALUE_CLASS, FIELD, NULL_VALUE, CONSTEXPR) \
```
AJenbo, line 10: if this thing exploads one day is it generic enough that it's trivial to replace with the std optional?

Source/engine/render/clx_render.cpp
AJenbo: Do we utilize this function during run time, or is it only for loading? Maybe we could embed it in the load routine before the sprite data is made `const`?

On a side note this very function is the main cause of crashes in 1.4.1 so fare so that makes a extra alarmed to have hacks in it.

Source/utils/intrusive_optional.hpp
glebm: @AJenbo Yes, the API is a subset of `std::optional`, it even uses `std::nullopt` assignment etc.

Source/engine/render/clx_render.cpp
glebm: > Maybe we could embed it in the load routine before the sprite data is made const?

The issue is that we do not have any types to represent non-const sprite data.
Anyway, this cast is completely safe, because the data did not start off as `const`.

> On a side note this very function is the main cause of crashes in 1.4.1

Interesting, we could replace this with runtime TRN application in a follow-up PR -- may cost some FPS but will have better RAM usage.

AJenbo (approved):

PR #5107, opened 2022-07-24
Add helper to check if a monster belongs to a player

ephphatha:
Spawned from the discussion in #5072, but relevant to the chat in discord around berserked monsters as well?

Please note I haven't captured every use of MFLAG_GOLEM and/or MT_GOLEM as some contexts I think still need to handle garbage data (like saves with golems in slots 4-19 from before #2822 was fixed).

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -640,7 +640,7 @@ void LoadMonster(LoadHelper *file, Monster &monster)
 	file->Skip(1); // Alignment
 	monster.exp = file->NextLE<uint16_t>();
 
-	if ((monster.flags & MFLAG_GOLEM) != 0) // Don't skip for golems
+	if (monster.isPlayerMinion()) // Don't skip for golems
 		monster.toHit = file->NextLE<uint8_t>();
 	else
 		file->Skip(1); // Skip hit as it's already initialized
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -393,9 +393,12 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
 	if (mid > 0 || (mid != 0 && Monsters[abs(mid) - 1].mode == MonsterMode::Petrified)) {
 		mid = abs(mid) - 1;
 		if (missile.IsTrap()
-		    || (missile._micaster == TARGET_PLAYERS
-		        && (Monsters[missile._misource].flags & MFLAG_TARGETS_MONSTER) != 0
-		        && ((Monsters[mid].flags & MFLAG_GOLEM) != 0 || (Monsters[missile._misource].flags & MFLAG_BERSERK) != 0))) {
+		    || (missile._micaster == TARGET_PLAYERS && (                                           // or was fired by a monster and
+		            Monsters[mid].isPlayerMinion() != Monsters[missile._misource].isPlayerMinion() //  the monsters are on opposing factions
+		            || (Monsters[missile._misource].flags & MFLAG_BERSERK) != 0                    //  or the attacker is berserked
+		            || (Monsters[mid].flags & MFLAG_BERSERK) != 0                                  //  or the target is berserked
+		            ))) {
+			// then the missile can potentially hit this target
 			isMonsterHit = MonsterTrapHit(mid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
 		} else if (IsAnyOf(missile._micaster, TARGET_BOTH, TARGET_MONSTERS)) {
 			isMonsterHit = MonsterMHit(missile._misource, mid, minDamage, maxDamage, missile._midist, missile._mitype, isDamageShifted);
@@ -595,7 +598,7 @@ bool GuardianTryFireAt(Missile &missile, Point target)
 	if (mid < 0)
 		return false;
 	const Monster &monster = Monsters[mid];
-	if (monster.type().type == MT_GOLEM)
+	if (monster.isPlayerMinion())
 		return false;
 	if (monster.hitPoints >> 6 <= 0)
 		return false;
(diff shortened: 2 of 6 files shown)


Source/loadsave.cpp
```diff
@@ -640,7 +640,7 @@ void LoadMonster(LoadHelper *file, Monster &monster)
 	file->Skip(1); // Alignment
 	monster.exp = file->NextLE<uint16_t>();
 
-	if ((monster.flags & MFLAG_GOLEM) != 0) // Don't skip for golems
+	if (monster.isPlayerOwned()) // Don't skip for golems
```
ephphatha, line 643: maybe this should be checking if the monster is berserked (or both?)

Source/missiles.cpp
```diff
@@ -251,7 +251,7 @@ bool MonsterMHit(int pnum, int monsterId, int mindam, int maxdam, int dist, miss
 	} else {
 		if (monster.mode != MonsterMode::Petrified && MissilesData[t].mType == 0 && HasAnyOf(player._pIFlags, ItemSpecialEffect::Knockback))
 			M_GetKnockback(monster);
-		if (monster.type().type != MT_GOLEM)
+		if (!monster.isPlayerOwned())
```
ephphatha, line 254: I'm assuming there are no non-player golems, might need to revert this one so they're still immune to knockback

Source/missiles.cpp
```diff
@@ -883,9 +886,8 @@ bool MonsterTrapHit(int monsterId, int mindam, int maxdam, int dist, missile_id
 		MonsterDeath(monster, monster.direction, true);
 	} else if (resist) {
 		PlayEffect(monster, 1);
-	} else {
-		if (monster.type().type != MT_GOLEM)
-			M_StartHit(monster, dam);
+	} else if (!monster.isPlayerOwned()) {
```
ephphatha, line 889: Same as the case for MonsterMHit, this allows non-player golems to be interrupted by trap hits.

Source/missiles.cpp
AJenbo: > I'm assuming there are no non-player golems

correct

Source/loadsave.cpp
AJenbo: Your change here seams correct. Baserked monsters should get there values from the monster table, but monsters spawned by the player is based on the values that where calculated then.

Source/missiles.cpp
```diff
@@ -393,9 +393,12 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
 	if (mid > 0 || (mid != 0 && Monsters[abs(mid) - 1].mode == MonsterMode::Petrified)) {
 		mid = abs(mid) - 1;
 		if (missile.IsTrap()
-		    || (missile._micaster == TARGET_PLAYERS
-		        && (Monsters[missile._misource].flags & MFLAG_TARGETS_MONSTER) != 0
-		        && ((Monsters[mid].flags & MFLAG_GOLEM) != 0 || (Monsters[missile._misource].flags & MFLAG_BERSERK) != 0))) {
+		    || (missile._micaster == TARGET_PLAYERS                                 // or was fired by a monster
+		        && (Monsters[missile._misource].flags & MFLAG_TARGETS_MONSTER) != 0 // which can target other monsters
```
AJenbo, line 397: the flag only indicates the type of the current target, not what the monster will possibly target
```suggestion
		        && (Monsters[missile._misource].flags & MFLAG_TARGETS_MONSTER) != 0 // which is target other monsters
```

Source/missiles.cpp
```diff
@@ -393,9 +393,12 @@ void CheckMissileCol(Missile &missile, int minDamage, int maxDamage, bool isDama
 	if (mid > 0 || (mid != 0 && Monsters[abs(mid) - 1].mode == MonsterMode::Petrified)) {
 		mid = abs(mid) - 1;
 		if (missile.IsTrap()
```
AJenbo, line 395: This might be better written as:
```cpp
if (missile.IsTrap()
    || (missile._micaster == TARGET_PLAYERS && ( // or was fired by a monster
        Monsters[mid].isPlayerOwned() != Monsters[missile._misource].isPlayerOwned() // between monster of different alignments
        || (Monsters[mid).flags & MFLAG_BERSERK) != 0 || (Monsters[missile._misource].flags & MFLAG_BERSERK) != 0)) { // or one of them is baserked
```

Source/missiles.cpp
AJenbo: This is because golems cannot be interupted, not because the player is controlling it.
```suggestion
	} else if (monster.type().type != MT_GOLEM) {
```

Source/monster.h
```diff
@@ -338,6 +338,12 @@ struct Monster { // note: missing field _mAFNum
 	bool isWalking() const;
 	bool isImmune(missile_id mitype) const;
 	bool isResistant(missile_id mitype) const;
+
+	/**
+	 * Is this a player's golem?
+	 */
+	[[nodiscard]] bool isPlayerOwned() const;
```
AJenbo, line 345: Somehow I don't much like the name. For one thing it makes me think that it might be specific to `MyPlayer`. Maybe `isPlayerMinion` or `IsPlayerControlled` would be better.

ephphatha: Haha, another case where I actually had it as `isPlayerControlled` then changed it because I was worried people might think it means a player is actually moving it around directly. `isPlayerMinion` sounds good.

AJenbo: Golems actually controlled by the player as they will walk in what ever cardinal direction the player is looking :)

ephphatha: I never realised that 🤯

AJenbo (approved):

PR #5110, opened 2022-07-24
Convert ZT gamename to lower case

DakkJaniels:


StephenCWills:
@glebm I'm reading that `std::tolower()` is locale-dependent. Is this another function we should be avoiding?

DakkJaniels:
>  I'm reading that std::tolower() is locale-dependent. Is this another function we should be avoiding?

I can just swap it to 
```cpp
selgame_Ip[i] = (selgame_Ip[i] >= 'A' && selgame_Ip[i] <= 'Z') ? selgame_Ip[i] + 'a' - 'A' : selgame_Ip[i];
```
if that's preferable

AJenbo:
This depend on us figuring out where the names are generated :)

StephenCWills:
https://github.com/diasurgical/devilutionX/blob/b21ebffaf2d20b40a25afe393a02713da0b81987/Source/dvlnet/protocol_zt.cpp#L315-L325

DakkJaniels:
was just going to post that lol

AJenbo:
Ok, well that confirms that they are all lowercase an limited to a-z :)

DakkJaniels:
could also swap to `std::tolower(selgame_Ip, std::locale("C"))` - I think that will force the conversion to "classic"

AJenbo (approved):

PR #5112, opened 2022-07-24
Improve demo test error message

glebm:
Example:

```
file "game" is different at byte 54251:
Expected: ... 00 00 00 00 00 00 00 00 30 00 00 00 50 00 00 00 ...
  Actual: ... FF FF FF FF 00 00 00 00 30 00 00 00 50 00 00 00 ...
```


AJenbo (approved):
I think @obligaron is also working on adding output that shows exactly what field was changed :)

PR #5114, opened 2022-07-24
♻️ Replaces direct access to _misource

canassa:
I grouped all the low-hanging fruits in a single PR.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -2372,7 +2372,7 @@ void AddNova(Missile &missile, const AddMissileParameter &parameter)
 
 void AddBlodboil(Missile &missile, const AddMissileParameter & /*parameter*/)
 {
-	Player &player = Players[missile._misource];
+	Player &player = *missile.sourcePlayer();
 
 	if (HasAnyOf(player._pSpellFlags, SpellFlag::RageActive | SpellFlag::RageCooldown) || player._pHitPoints <= player._pLevel << 6) {
 		missile._miDelFlag = true;
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1686,14 +1686,14 @@ void AddFirebolt(Missile &missile, const AddMissileParameter &parameter)
 		dst += parameter.midir;
 	}
 	int sp = 26;
-	if (missile._micaster == TARGET_MONSTERS) {
+	if (missile.sourceType() == MissileSource::Player) {
 		sp = 16;
 		if (!missile.IsTrap()) {
 			sp += std::min(missile._mispllvl * 2, 47);
 		}
 
 		if (parameter.pParent == nullptr || parameter.pParent->_mitype != MIS_GUARDIAN)
-			ConsumeSpell(Players[missile._misource], SPL_FIREBOLT);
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -1913,15 +1921,19 @@ void AddFlash(Missile &missile, const AddMissileParameter & /*parameter*/)
...
+		int dmg = missile.sourcePlayer()->_pLevel + 1;
+		dmg += GenerateRndSum(20, dmg);
+		missile._midam = ScaleSpellEffect(dmg, missile._mispllvl);
+		missile._midam += missile._midam / 2;
+	} break;
+	case MissileSource::Trap:
+		missile._midam = currlevel / 2;
+		break;
+	case MissileSource::Monster:
+		assert(missile.sourceType() == MissileSource::Monster);
```
ephphatha, line 1935: This wont fail as expected.
```suggestion
		assert(missile.sourceType() != MissileSource::Monster);
```

Source/missiles.cpp
```diff
@@ -1429,19 +1429,27 @@ void AddLightningArrow(Missile &missile, const AddMissileParameter &parameter)
...
-		missile.var1 = Players[missile._misource].position.tile.x;
-		missile.var2 = Players[missile._misource].position.tile.y;
+		break;
+	case MissileSource::Player: {
+		Player &player = *missile.sourcePlayer();
+		missile.var1 = player.position.tile.x;
+		missile.var2 = player.position.tile.y;
+	} break;
+	case MissileSource::Monster: {
+		assert(missile.sourceType() != MissileSource::Monster);
```
glebm, line 1443: This is a bit odd, perhaps just `app_fatal("Invalid missile source")`

Source/missiles.cpp
```diff
@@ -1209,7 +1209,7 @@ void AddJester(Missile &missile, const AddMissileParameter &parameter)
 		break;
 	case 6:
 		spell = MIS_TOWN;
-		UseMana(Players[missile._misource], SPL_TOWN);
+		UseMana(*missile.sourcePlayer(), SPL_TOWN);
```
glebm, line 1212: Still think `sourcePlayer/Monster` should return references and perform no checks (just an `assert`).
As I understand you plan to do this later, which will be another substantial diff? Why not do this now?

Source/missiles.cpp
canassa: There was already some discussion about that pattern on https://github.com/diasurgical/devilutionX/pull/5098#discussion_r928109064 https://github.com/diasurgical/devilutionX/pull/5098#discussion_r928305171

I followed the pattern from that one

Source/missiles.cpp
canassa: I opened a PR for that but I closed it since there wasn't a consensus with @AJenbo 

https://github.com/diasurgical/devilutionX/pull/5054

I am fine with both approaches. I have no strong opinion on this 🙂

Source/missiles.cpp
AJenbo: The app_fatal() is also fine and is how it's normally done, except it will change it to also crash in production.

PR #5115, opened 2022-07-25
Bump NDK version

AJenbo:
25 LTS gives SDL build errors (will be default in august).
24 gives a build error for one of our std polyfills.
23 works, but is not LTS.
21 LTS (current default) chokes on CMake 3.22 (which is now the version on Ubuntu, and also de latest in Android Studio).

Hopefully the SDL build errors for 25 will be solved in the not to distant future.

### DIFF
--- a/.github/workflows/Android.yml
+++ b/.github/workflows/Android.yml
@@ -26,11 +26,6 @@ jobs:
         distribution: 'adopt'
         cache: gradle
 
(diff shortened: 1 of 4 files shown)


tsunamistate:
Probably related to SDL build errors on NDK 25: https://github.com/libsdl-org/SDL/pull/5929

glebm (approved):

PR #5117, opened 2022-07-25
Move theme internals into anonymous namespace, misc cleanups

ephphatha:
Was looking at adding a helper for checking if a point matches a given regionId however I don't really want to introduce an unavoidable function call by using the same pattern as IsTileSolid and defining it in path.cpp. Probably move all that into a new header for dungeon tile checks?

This was the intent for CheckThemeObj5 but omitted for now:
```cpp
bool IsTileInRegion(Point position, int8_t regionId)
{
	return InDungeonBounds(position) && dTransVal[position.x][position.y] == regionId;
}
```

### DIFF
--- a/Source/levels/themes.cpp
+++ b/Source/levels/themes.cpp
@@ -36,22 +36,6 @@ bool bFountainFlag;
 
 /** Specifies the set of special theme IDs from which one will be selected at random. */
 theme_id ThemeGood[4] = { THEME_GOATSHRINE, THEME_SHRINE, THEME_SKELROOM, THEME_LIBRARY };
-/** Specifies a 5x5 area to fit theme objects. */
-int trm5x[] = {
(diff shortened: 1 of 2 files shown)


Source/levels/themes.cpp
```diff
@@ -97,51 +83,37 @@ bool TFit_Shrine(int i)
...
-				yp++;
-				if (yp == MAXDUNY) {
-					if (r == rs) {
-						return false;
-					}
-					yp = 0;
-				}
+bool TFit_Obj5(int t)
+{
+	AdvanceRndSeed();
```
ephphatha, line 104: Hmm, this seems to actually be discarding the first x number of potential spots, I'll need to restore that.

AJenbo (approved):
:+1:

PR #5118, opened 2022-07-25
:sparkles: Adds ignoreMovingMonsters to MonsterAtPosition

canassa:
I believe that most read access to the `dMonster` global could be replaced with calls to `MonsterAtPosition`, which should improve the code readability and reduce global variable usage.

But, some places are checking if the dMonster contains no monster at all (`== 0`) while other places are checking for moving monsters as well (`<= 0`) and the current version of `MonsterAtPosition` only supports `== 0`.

This PR adds a strict flag that allows the `MonsterAtPosition` caller to switch the function's behaviour.

~~Note: I don't think `strict` is a good name, if you guys like this proposal we should find a better name~~
The parameter was renamed to `ignoreMovingMonsters` thanks to the feedback

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4482,17 +4482,18 @@ void MissToMonst(Missile &missile, Point position)
 		return;
 	}
 
-	if (dMonster[oldPosition.x][oldPosition.y] <= 0)
+	Monster *target = FindMonsterAtPosition(oldPosition, true);
+
+	if (target == nullptr)
 		return;
 
-	Monster &target = *FindMonsterAtPosition(oldPosition);
-	MonsterAttackMonster(monster, target, 500, monster.minDamageSpecial, monster.maxDamageSpecial);
+	MonsterAttackMonster(monster, *target, 500, monster.minDamageSpecial, monster.maxDamageSpecial);
 
 	if (IsAnyOf(monster.type().type, MT_NSNAKE, MT_RSNAKE, MT_BSNAKE, MT_GSNAKE))
 		return;
 
 	Point newPosition = oldPosition + monster.direction;
-	if (IsTileAvailable(target, newPosition)) {
+	if (IsTileAvailable(*target, newPosition)) {
 		monsterId = dMonster[oldPosition.x][oldPosition.y];
 		dMonster[newPosition.x][newPosition.y] = monsterId;
 		dMonster[oldPosition.x][oldPosition.y] = 0;
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -3729,7 +3729,7 @@ void M_StartStand(Monster &monster, Direction md)
 void M_ClearSquares(const Monster &monster)
 {
 	for (Point searchTile : PointsInRectangleRange { Rectangle { monster.position.old, 1 } }) {
-		if (MonsterAtPosition(searchTile) == &monster)
+		if (FindMonsterAtPosition(searchTile) == &monster)
 			dMonster[searchTile.x][searchTile.y] = 0;
 	}
 }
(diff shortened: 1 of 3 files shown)


Source/monster.cpp
```diff
@@ -4503,20 +4502,20 @@ void MissToMonst(Missile &missile, Point position)
 	}
 }
 
-Monster *MonsterAtPosition(Point position)
+Monster *MonsterAtPosition(Point position, bool strict)
```
ephphatha, line 4505: `ignoreMovingMonsters` maybe? `-id` is used for the tile a monster is leaving (or moving into...) as well as charging monsters. Most of the time the caller still wants the tile the monster is moving into when they're walking normally though so there's still potential confusion there...

canassa: Isn't the -id also used for monsters that occupy multiple tiles?

qndel: > Isn't the -id also used for monsters that occupy multiple tiles?

are there any?

ephphatha: only cows in town, and town uses a special case for hit detection in cursor.cpp anyway which could be called out as an exception

glebm, line 4505: Rather than making it a flag, let's have 2 functions with different return types:

```c++
// Assumes `position` is in dungeon bounds and contains a monster.
Monster &MonsterAtPosition(Point);

// Returns the monster if the `position` is in dungeon bounds and contains a monster,
// `nullptr` otherwise.
Monster *MaybeMonsterAtPosition(Point);
```

Not sure about the names either but having 2 functions does seem sensible.

canassa: The "strict" version can also return a `nullptr`. The only different is that it won't check for walking monsters

canassa: Ah yes, cows 🐮. I remember seeing -id being used for that but I wasn't sure where 😄 

I agree that that's an exception so I will rename `strict` to `ignoreMovingMonsters`

canassa: Paramter renamed

ephphatha: With the way these functions are currently used I personally think that `ObjectAtPosition`, `PlayerAtPosition`, and `MonsterAtPosition` are more natural returning pointers. The position based lookups are similar to `map::find()` which tend to return `iterator::end` when the find fails, though the name does imply that it's equivalent to `map::at()`...

Putting the logic to filter moving players/monsters and large objects into `Is*AtPosition` (as an equivalent to `map::contains()` and making the at method just do a simple lookup could be worth it though.

canassa: I am a bit confused about what I should do here 😄 

This PR is not changing the return type. It's still returning pointers and not references. The only thing that the extra does is to indicate if `MonsterAtPosition` should ignore walking monsters or not.

I could also have it as a separate function instead of a flag, but then naming becomes a challenge. Perhaps `StandingMonsterAtPosition`?

ephphatha: Nah, keeping the one function for now is fine imo. `MonsterAtPosition` is barely used at the moment, need to expand the usage and work out if splitting it like `ObjectAtPosition`/`FindObjectAtPosition` makes sense.

If you want you could rename `MonsterAtPosition` to `FindMonsterAtPosition` (keep the param you've added, just rename all current uses). We can then add a version returning a reference if it's useful.

canassa: Alright! Function renamed

Source/monster.cpp
```diff
@@ -4486,7 +4485,7 @@ void MissToMonst(Missile &missile, Point position)
 	if (dMonster[oldPosition.x][oldPosition.y] <= 0)
 		return;
 
-	Monster &target = *MonsterAtPosition(oldPosition);
+	Monster &target = *FindMonsterAtPosition(oldPosition);
```
ephphatha, line 4488: Just noticed this can use `FindMonsterAtPosition(oldPosition, true)`

canassa: True, fixed on 039e7846cb2a5b6d5eb8318b2f035ce690967e56

ephphatha (approved):
Looks good to me, not gonna auto-merge to give glebm a chance to review though

glebm (approved):

PR #5120, opened 2022-07-25
fix jester's description

qndel:
Also for anyone wondering about average dmg:

0-99 = 0-99% dmg = average dmg = (0 + 99%) / 2 = 49.5% dmg
100 = 100% dmg = average dmg 100%
101-200 = 100% + (value - 100) * 5 dmg = average dmg = (105% + 600%) / 2 = 352.5%

final average:
100/201 * 49.5 +
1/201 * 100 +
100/201 * 352.5
= (100/201 * 49.5) + (1/201 * 100) + (100/201 * 352.5) = 200.497512438%


Source/items.cpp
```diff
@@ -3709,7 +3709,7 @@ bool DoOil(Player &player, int cii)
 	case IPL_PERIL:
 		return _("2x dmg to monst, 1x to you");
 	case IPL_JESTERS:
-		return std::string(_(/*xgettext:no-c-format*/ "Random 0 - 500% damage"));
+		return std::string(_(/*xgettext:no-c-format*/ "Random 0 - 600% damage"));
```
glebm, line 3712: Can you also update the translations?

julealgon: Is there any reason why the numbers are not parameterized in the localization? Having to update translations because of a numerical change like this seems sub-optimal.

qndel: iirc this was the only affix with a hardcoded number - we don't really expect it to change ;)

qndel: > Can you also update the translations?

done

glebm (approved):

Source/items.cpp
julealgon: > iirc this was the only affix with a hardcoded number - we don't really expect it to change ;)

Even if that's the case, it is still beneficial to parameterize it for translation purposes IMHO.

PR #5130, opened 2022-07-26
`Player#getGraphic`: Remove `MyPlayer` check

glebm:
We always have `_pSpell` when this function is called.


AJenbo (approved):

AJenbo:
if nothing else we should know once inventory is synced and double casting is fixed

PR #5131, opened 2022-07-26
Github Actions: Add Windows x64 Release Build

glebm:
Not sure if this works but needs to be merged first so that it can be triggered manually


AJenbo (commented):
I prefer the release builds to be with mingw rather then msvc since i understand it's dependencies better

Trihedraf:
Also MinGW is smaller and there are less dlls.

AJenbo:
> there are less dlls

Yeah that is my main reason. I strongly wanted to avoid "now install the latest .Net runtime framework update as of 2021 december including the SP4 pack" in the install instruktions :D

PR #5137, opened 2022-07-26
:recycle: Replace dMonster usage on DoAttack

canassa:


Source/player.cpp
```diff
@@ -1055,10 +1052,10 @@ bool DoAttack(Player &player)
 			}
 		}
 
-		if (dMonster[dx][dy] != 0) {
-			didhit = PlrHitMonst(player, abs(dMonster[dx][dy]) - 1);
-		} else if (dPlayer[dx][dy] != 0 && !player.friendlyMode) {
-			didhit = PlrHitPlr(player, Players[abs(dPlayer[dx][dy]) - 1]);
+		if (monster != nullptr) {
+			didhit = PlrHitMonst(player, *monster);
+		} else if (monster != nullptr && !player.friendlyMode) {
```
ephphatha, line 1057: monster  != nullptr is the wrong check here, with that change this case will never be true.
```suggestion
		} else if (IsPlayerAtPosition(position) && !player.friendlyMode) {
```

canassa: Thanks!

AJenbo: You will also have to implement `IsPlayerAtPosition`

canassa: I replaced it with `PlayerAtPosition`

Source/main.cpp
```diff
@@ -1,3 +1,6 @@
+#if (defined(_WIN64) || defined(_WIN32)) && defined(__GNUC__)
```
qndel, line 1: what's this

canassa: Rebase error with master, let me fix

Source/player.cpp
```diff
@@ -1055,10 +1052,10 @@ bool DoAttack(Player &player)
 			}
 		}
 
-		if (dMonster[dx][dy] != 0) {
-			didhit = PlrHitMonst(player, abs(dMonster[dx][dy]) - 1);
-		} else if (dPlayer[dx][dy] != 0 && !player.friendlyMode) {
-			didhit = PlrHitPlr(player, Players[abs(dPlayer[dx][dy]) - 1]);
+		if (monster != nullptr) {
+			didhit = PlrHitMonst(player, *monster);
+		} else if (PlayerAtPosition(position) != nullptr && !player.friendlyMode) {
```
ephphatha, line 1057: There's no nice way to structure this to avoid calling `PlayerAtPosition` twice(or doing it unnecessarily before the if statements :/). Would be worth reordering the conditions so the PVP flag is checked first but I'd do that in a separate PR and include things like the calls to `CanTalkToMonst` in the cleave section.

ephphatha (approved):

Source/player.cpp
canassa: There's the "If statement with initializer" from C++ 17 but I heard that that feature is not very popular 😁

ephphatha: yeah, I'd rate that a "not-nice" way of doing it :D

PR #5138, opened 2022-07-27
In-game menu: Fix slider click handling

glebm:
Previously, clicking on the slider close to the end or near the start did nothing -- you'd have to actually drag it to set the value to min or max.

Also cleans up the in-game menu code overall.

### DIFF
--- a/Source/gmenu.h
+++ b/Source/gmenu.h
@@ -18,6 +18,56 @@ struct TMenuItem {
 	uint32_t dwFlags;
 	const char *pszStr;
 	void (*fnMenu)(bool);
+
+	[[nodiscard]] bool enabled() const
+	{
+		return (dwFlags & GMENU_ENABLED) != 0;
+	}
+
+	[[nodiscard]] bool isSlider() const
+	{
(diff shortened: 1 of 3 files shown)


Source/gmenu.h
```diff
@@ -18,6 +18,56 @@ struct TMenuItem {
...
+		dwFlags &= 0xFFFFF000;
+		dwFlags |= step;
+	}
+
+	[[nodiscard]] uint16_t sliderSteps() const
+	{
+		return (dwFlags & 0xFFF000) >> 12;
+	}
+
+	void setSliderSteps(uint16_t steps)
```
ephphatha, line 48: `setSliderPosition` maybe? Seems very easy to confuse this with the other function which sets the step/increment value.

glebm: Not sure what the best name is but this sets the total number of steps (plus  one actually).

The component doesn't keep the difference between min and max, so for example for the `[25, 75]` range it'd be called with `50`.

The "step" granularity is always 1.

glebm:
Rebased

AJenbo (approved):

PR #4665, opened 2022-06-05
Clean up drlg tests

AJenbo:
In preparation for testing dungeons with quest elements

PR #5144, opened 2022-07-28
Don't recalculate leveltype (from currlevel) in SaveGameData

### LINKED ISSUE #4954: Crash when using town portal
**Important information**
Operating system (Debian Linux 5.18.0-2-amd64 x86_64) / DevilutionX (latest master)  / did you download a release or did you compile yourself? (compiled)

**Describe**
Started a game. Heavily shortened the first 4 levels and completely played through the fifth level.
Cleared also the sixth level and went to the bone chamber. Killed all monsters in the bone chamber.
Due to lack of space, left some items lying around and went to the town through a town portal.
Completed the necessary sales and attempted to return to the bone chamber through the portal.
A crash occurs when entering the town portal.
I tried to bisect the problem, but unfortunately the savegame must be defective because the crash occurs with every version of devilutionx. So I don't know how to help you except to provide the savegame.

[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/9079958/single_0.hsv.zip)

**To Reproduce**
1. Load attached game
2. Try to enter the town portal
3. See crash

**Expected behavior**
Never to crash

**Screenshots**

**Additional context**
I've noticed that a lot of cleanups have taken place in the monster and dungeon generation code.
I can therefore only assume that there may have been some new issues that I hope can be solved.

EDIT:
The game can be continued by returning to the bone chamber through the level entrances.

obligaron:
Fixes #4954

Setting `leveltype` from `GetLevelType(currlevel)` is wrong if we are on a quest map (`setlevel` is true).

Was [introduced](https://github.com/diasurgical/devilutionX/commit/54cd839cf6c7b4647f3b1906275f9dcae225b7a8#diff-eb2db6084d42c0224b56a122152915d07d944c27040e427d6193bc382ca754b3R2267) in #4665 /  54cd839cf6c7b4647f3b1906275f9dcae225b7a8 .

Many thanks to @Chance4us for finding the buggy commit and the nice reproduction videos. 🙂👍

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -2326,8 +2326,6 @@ void SaveGameData(MpqWriter &saveWriter)
 	file.WriteBE<uint32_t>(static_cast<uint32_t>(std::min(Missiles.size(), MaxMissilesForSaveGame)));
 	file.WriteBE<int32_t>(ActiveObjectCount);
 
-	leveltype = GetLevelType(currlevel);
-
 	for (uint8_t i = 0; i < giNumberOfLevels; i++) {
 		file.WriteBE<uint32_t>(glSeedTbl[i]);
 		file.WriteBE<int32_t>(getHellfireLevelType(GetLevelType(i)));


Chance4us:
The fix works for me.
Thanks a lot @obligaron :-)

Source/loadsave.cpp
```diff
@@ -2326,8 +2326,6 @@ void SaveGameData(MpqWriter &saveWriter)
 	file.WriteBE<uint32_t>(static_cast<uint32_t>(std::min(Missiles.size(), MaxMissilesForSaveGame)));
 	file.WriteBE<int32_t>(ActiveObjectCount);
 
-	leveltype = GetLevelType(currlevel);
```
AJenbo, line 2329: This was added for correcting crypt and hive being designated cathedral and caves in original hellfire save games. Do we not still need it for correcting this? Maybe we should just add `(!setLevel)` before it?

obligaron: When saving data we need to stay compatible with the old format.
This is done by calling `getHellfireLevelType` when writing to the save game.
The in-memory data is still in the new format (`DTYPE_CRYPT`&`DTYPE_NEST`).
And I thought saving shouldn't change the in-memory data or?

(When reading from the save game `(!setLevel)` is used).

AJenbo: Oh wait, I must have looked in the wrong spot, I thought this was in the LoadLevel function, not save data...

PR #5146, opened 2022-07-29
Clean up logic in enchanted shrine

ephphatha:
Ended up looking at this due to a discussion in discord. The way a spell is chosen to take a level (or two) from was wasteful and didn't really need to be kept as is since it happens after dungeon generation. Using the same approach as Stephen pioneered in #3314 we can just directly use the object seed to determine the effect.

While trying to confirm the loop range was correct I noticed a lot of magic numbers/hardcoded values that really related to `spell_id` enum members, updated those as I went along.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -2728,37 +2728,40 @@ void OperateShrineReligious(Player &player)
 	InitDiabloMsg(EMSG_SHRINE_RELIGIOUS);
 }
 
-void OperateShrineEnchanted(Player &player)
+void OperateShrineEnchanted(Player &player, uint32_t seed)
 {
 	if (&player != MyPlayer)
 		return;
 
-	int cnt = 0;
-	uint64_t spell = 1;
-	int maxSpells = gbIsHellfire ? MAX_SPELLS : 37;
-	uint64_t spells = player._pMemSpells;
-	for (int j = 0; j < maxSpells; j++) {
-		if ((spell & spells) != 0)
-			cnt++;
-		spell *= 2;
-	}
-	if (cnt > 1) {
-		spell = 1;
-		for (int j = SPL_FIREBOLT; j < maxSpells; j++) { // BUGFIX: < MAX_SPELLS, there is no spell with MAX_SPELLS index (fixed)
-			if ((player._pMemSpells & spell) != 0) {
-				if (player._pSplLvl[j] < MaxSpellLevel)
-					player._pSplLvl[j]++;
-			}
-			spell *= 2;
+	const size_t lastSpellIndex = static_cast<size_t>(gbIsHellfire ? spell_id::SPL_LAST : spell_id::SPL_LASTDIABLO);
+
+	std::bitset<MAX_SPELLS> knownSpells { player._pMemSpells & (~0ULL >> (64 - lastSpellIndex)) };
+	// If the player knows any spells
+	if (knownSpells.any()) {
+		// choose a spell at random to decrement (+1 as spells are 1-indexed in Player::_pSplLvl)
+		unsigned r = ShiftModDistribution(seed, knownSpells.count()) + 1;
+
+		// r currently represents the nth set bit, translate this to the actual bit position
+		for (size_t i = 0; i < r; i++) {
+			if (!knownSpells.test(i))
+				r++;
 		}
-		int r;
-		do {
-			r = GenerateRnd(maxSpells);
-		} while ((player._pMemSpells & GetSpellBitmask(r + 1)) == 0);
-		if (player._pSplLvl[r + 1] >= 2)
-			player._pSplLvl[r + 1] -= 2;
-		else
-			player._pSplLvl[r + 1] = 0;
+
+		for (size_t j = static_cast<size_t>(SPL_FIREBOLT); j <= lastSpellIndex; j++) {
+			if ((player._pMemSpells & GetSpellBitmask(static_cast<spell_id>(j))) != 0) {
+				if (j == r) {
+					// decrement the randomly chosen spell if it's above level 0
+					if (player._pSplLvl[r] >= MaxSpellLevel)
+						player._pSplLvl[r] -= 2; // by 2 levels if it was maxed
+					else if (player._pSplLvl[r] > 0)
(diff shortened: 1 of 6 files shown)


Source/objects.cpp
```diff
@@ -2733,32 +2733,30 @@ void OperateShrineEnchanted(Player &player)
...
+	const unsigned maxSpells = gbIsHellfire ? MAX_SPELLS : 37;
+
+	const uint64_t knownSpellMask = (0xFFFFFFFFULL >> (64 - maxSpells));
+	if ((player._pMemSpells & knownSpellMask) != 0) {
+		// If the player knows any spells
+		int r;
+		do {
+			// choose a spell at random to decrement
+			r = GenerateRnd(maxSpells) + 1;
+		} while ((player._pMemSpells & GetSpellBitmask(r)) == 0);
```
ephphatha, line 2745: Negative values returned by the RNG I think end up calling lshift with a negative index, resulting in a mask of 0x00000000, so the loop continues.

Source/objects.cpp
```diff
@@ -2733,32 +2733,30 @@ void OperateShrineEnchanted(Player &player)
...
+
 		for (int j = SPL_FIREBOLT; j < maxSpells; j++) { // BUGFIX: < MAX_SPELLS, there is no spell with MAX_SPELLS index (fixed)
```
ephphatha, line 2747: The cnt loop and random index can both choose a spell in index maxSpells, this loop could result in adding a level to all known spells and not decrementing a level from any in that case.

Source/objects.cpp
ephphatha: I'm not gonna bother fixing this properly in this PR, going to use the seed directly like in #3314 to pick the spell to decrement. Need to do a bit of cleanup around the way GetSpellBitmask and _pSplLvl is used to make it look nice though.

ephphatha: And I did it anyway -_- I actually specified the wrong mask for determining player spells so it was easier to fix it all together.

Source/objects.cpp
```diff
@@ -2728,37 +2728,40 @@ void OperateShrineReligious(Player &player)
...
+		unsigned r = ShiftModDistribution(seed, knownSpells.count()) + 1;
+
+		// r currently represents the nth set bit, translate this to the actual bit position
+		for (size_t i = 0; i < r; i++) {
+			if (!knownSpells.test(i))
+				r++;
+		}
+
+		for (size_t j = static_cast<size_t>(SPL_FIREBOLT); j <= lastSpellIndex; j++) {
+			if ((player._pMemSpells & GetSpellBitmask(static_cast<spell_id>(j))) != 0) {
```
StephenCWills, line 2751: This is pretty confusing. The first for loop isn't really getting the actual bit position, but rather the 1-based spell index. Also, is there some reason we shouldn't use the same approach in both for loops (`std::bitset` vs `GetSpellBitmask()`)?

ephphatha: we could make player._pMemSpells a bitset, it'd require a bit more refactoring to update the uses and save/load of that field as well though.

Regarding the 0-based/1-based confusion, player._pMemSpells is 0-based, but player._pSplLvl is 1-based as player._pSplLevel[0] corresponds to the class skill (and is always 0). It could be rewritten like the following to make it clearer:
```cpp
size_t r = ShiftModDistribution(knownSpells.count());

for (size_t i = 0; i < r; i++) {
    if (!knownSpells.test(i))
        r++;
}

r++; // translate from 0-based index as used in _pMemSpells to 1-based as used for _pSplLvl
```

Could also just use a single loop and use r as a decrementing counter of how many known spells to skip.

StephenCWills: I didn't necessarily mean to suggest we make `player._pMemSpells` a bitset, although maybe that would be an overall improvement. I'm really just talking about how the bitset is literally an equivalent representation of `player._pMemSpells` within the scope of this particular PR. Consider this implementation of the first loop.

```c++
for (size_t i = static_cast<size_t>(SPL_FIREBOLT); i <= r; i++) {
	if ((player._pMemSpells & GetSpellBitmask(static_cast<spell_id>(i))) == 0)
		r++;
}
```

One could similarly reimagine the second loop using the bitset instead of `GetSpellBitmask()`. IMO, either way you slice it, it doesn't make sense to use a bitset for the first loop and then `GetSpellBitmask()` for the second loop when they essentially do the same thing.

ephphatha: The bitset currently is pretty much only used to get the count of known spells for the currently active game mode. In most cases it'll be equivalent to `Player::_pMemSpells` however I wasn't sure whether loading a hellfire character into diablo masked off the hellfire specific spells on load (and didn't bother to check...). The bitmask is constructed by ANDing with a bitmask representing all valid spells for the given game mode as a safety measure. I should comment that better, but I'll have a go at refactoring `_pMemSpells` when I get time and revise this PR instead.

In the current state of this PR the first loop is primarily concerned with the bitset of known spells within the given game mode, which may not be the same as `Player::_pMemSpells`.

The second loop uses `spell_id` enum values as indexes into `Player::_pSplLvl` as the primary interaction. I kept it using `GetSpellBitmask()` to document the translation from `spell_id` as an index to the appropriate `Player::_pMemSpells` bit position.

StephenCWills: Upon reflection, I think the bitset itself is more of a minor problem compared to the sleight-of-hand pulled with the variable `r`. Ultimately, I just feel this code should just be written in a way that is more conceptually accurate.

* Your first comment claims we add 1 because the we need to use a 1-based index for `player._pSplLvl`. The very next comment immediately reinterprets it as the nth known spell, which I think is the more accurate interpretation based on how it's generated.
* `r` is then used as the upper bound for a for loop that uses a 0-based index for the bitset. This makes the least amount of sense because `i` doesn't really represent the ith known spell. This feels more like a micro-optimization made possible by the dual interpretation of the variable `r`.
* When the loop is complete, `r` somewhat magically points to the `spell_id` of the nth known spell. Conceptually, the loop is shifting `r` over by the number of unknown spells with a smaller index than the nth known spell, retroactively suggesting maybe the first comment was right all along.

PR #5147, opened 2022-07-30
HitMonster() cleanup

StephenCWills:
Cleanup in preparation for the fix described in #5109.

You can see from the commits that I tackled this piecemeal in the hopes that it would help me to avoid making mistakes. I'm leaving this unsquashed in case it helps with the review. Feel free to squash when merging.

### DIFF
--- a/Source/monster.h
+++ b/Source/monster.h
@@ -345,6 +345,7 @@ struct Monster { // note: missing field _mAFNum
 	[[nodiscard]] bool isPlayerMinion() const;
 
 	bool isPossibleToHit() const;
+	void tag(const Player &tagger);
 
 	[[nodiscard]] bool isUnique() const
 	{
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -887,6 +887,7 @@ bool MonsterTrapHit(int monsterId, int mindam, int maxdam, int dist, missile_id 
 	} else if (resist) {
 		PlayEffect(monster, 1);
 	} else if (monster.type().type != MT_GOLEM) {
+		PlayEffect(monster, 1);
 		M_StartHit(monster, dam);
 	}
 	return true;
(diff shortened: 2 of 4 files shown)


Source/monster.h
```diff
@@ -345,6 +345,7 @@ struct Monster { // note: missing field _mAFNum
 	[[nodiscard]] bool isPlayerMinion() const;
 
 	bool isPossibleToHit() const;
+	void Tag(const Player &tagger);
```
ephphatha, line 348: camelCase for class member functions

ephphatha:
Had a quick playtest and it seems fine, I'm getting a bit tripped up on why the hit sound call was pulled out of `M_StartHit(Monster&, int)` though. It seemed natural to have the sound effect play in the same function that determines if the monster should react/re-target in response to damage, but after the refactor only 2 of the 4 call sites actually play the sound before proceeding into the function. For the last 2 `ApplyMonsterDamage` means the sound gets called in monster.cpp:4447 immediately before a call to `M_StartHit(Monster&, int)`, and while slightly harder to trace the other use of `ApplyMonsterDamage` at monster.cpp:1196 results in the effect playing before a call to `M_StartHit` inside `MonsterHitMonster`.

You mentioned further cleanup is planned, are you looking to simplify the way sounds are triggered as well?

StephenCWills:
Hm, you might be right. My plan for this was to replace the two usages of `PlayEffect(monster, 1)` in missiles.cpp with `ApplyMonsterDamage()` to fix the bug with resists. I was also planning to use `ApplyMonsterDamage()` in other places like thorn and reflect. It makes sense to me that if a monster takes damage, the game should play the sound effect. However, that thought breaks down a bit due to network desync, in which case it might indeed be better to put the sound effect back into `M_StartHit()` and continue treating missiles.cpp as a special case.

ephphatha (approved):
Given you're planning on further cleanups having the sound play when damage is applied seems fine, lets `M_StartHit(Monster&, int)` be dedicated to follow on effects

PR #5148, opened 2022-07-30
:recycle: Pass objects by reference

canassa:
- Passing objects by reference
- Renamed a few functions

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1187,14 +1187,14 @@ void AddBloodFountain(Object &bloodFountain)
 	bloodFountain._oRndSeed = AdvanceRndSeed();
 }
 
-void AddPurifyingFountain(Object &purifyingFountain)
+void AddLargeFountain(Object &fountain)
 {
-	int ox = purifyingFountain.position.x;
-	int oy = purifyingFountain.position.y;
-	dObject[ox][oy - 1] = -(purifyingFountain.GetId() + 1);
-	dObject[ox - 1][oy] = -(purifyingFountain.GetId() + 1);
-	dObject[ox - 1][oy - 1] = -(purifyingFountain.GetId() + 1);
-	purifyingFountain._oRndSeed = AdvanceRndSeed();
+	int ox = fountain.position.x;
+	int oy = fountain.position.y;
+	dObject[ox][oy - 1] = -(fountain.GetId() + 1);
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -1067,33 +1067,32 @@ void InitializeMicroDoor(Object &door)
 	ObjSetMicro(door.position, pieceNumber);
 }
 
-void AddSarc(int i)
+void AddSarcophagus(Object &sarcophagus)
 {
-	dObject[Objects[i].position.x][Objects[i].position.y - 1] = -(i + 1);
```
ephphatha, line 1072: This is an initialisation function that marks the tiles occupied by a large object. Sarcophagi occupy two tiles, this needs to be kept in. This is breaking the demo test.

Source/objects.cpp
```diff
@@ -1164,140 +1163,140 @@ void AddShrine(int i)
...
 
-void AddCauldron(int i)
+void AddCauldron(Object &cauldron)
 {
-	Objects[i]._oRndSeed = AdvanceRndSeed();
+	cauldron._oRndSeed = AdvanceRndSeed();
 }
 
-void AddMurkyFountain(int i)
+void AddMurkyFountain(Object &murkyFountain)
```
ephphatha, line 1219: Have a look at #4571, a lot of these functions can be combined.

Source/objects.cpp
canassa: That was removed by mistake, let me put it back

canassa: Done

Source/objects.cpp
canassa: Done

Source/objects.cpp
```diff
@@ -1164,140 +1164,130 @@ void AddShrine(int i)
...
+void AddBookcase(Object &bookcase)
 {
-	Objects[i]._oRndSeed = AdvanceRndSeed();
-	Objects[i]._oPreFlag = true;
+	bookcase._oRndSeed = AdvanceRndSeed();
+	bookcase._oPreFlag = true;
 }
 
-void AddBookstand(int i)
+void AddBookstand(Object &bookstand)
```
ephphatha, line 1180: The other big cleanup that would be worth including from #4571 is combining all these functions which only set `Object::_oRndSeed` :D

glebm (approved):

PR #5150, opened 2022-07-30
Add detail log info to CompareSaves

obligaron:
With this change the save game comparison can display detailed info what fields changed (used for replay based unit tests).

Sample output:
> Diff found in game.player.velocityX count: 1
> Diff found in game.player.velocityY count: 1

Notes:
- I described the save game format in extern files (`txt` files under `memory_map`). This has some benefits
   - The txt files are only included when `BUILD_TESTING` is defined to reduce the amount of space they use.
   - Less code is needed.
   - Files could be used for some other use-cases (for example a save game browser/inspector).
   - This is currently only used in unit tests.
 - I'm not sure if I got every field correct. So there could be some errors in the data.
 - I tried to compare even files with different sizes (for example different number of monsters stored)
 - The diff output is aggregated to reduce output size (for example if 3 monsters all have a different velocity the output is `game.monsters.velocityX count: 3`)).
 - The format could be improved, but my motivation to work even more on this declined with the many fields that I needed to map. Perhaps this can be done in a future pr.

Format:
- R \<NumberOfBits> \<comment> (Read number of bytes)
- M \<Counter> \<NumberOfBits> \<comment> (Read multiple number of bytes (array etc.))
- LT \<NumberOfBits> \<comment> (Read leveltype, needed to check if this is a town level)
- LC/LC_LE \<NumberOfBits> \<comment> (Load counter (_LE = Little endian). This is used in M/C <Counter>)
- C \<Counter> \<memory_map> \<comment> (Calls in another memory_map file. For example game contains missiles).
- Suffix _HF (Only for Hellfire)
- Suffix _DA (Only for Diablo)
- Suffix _DL (Only for dungeon level)

I hope this helps when working with replay based unit tests. 🙂


AJenbo (commented):
Could we move this to just be part of the tests? Shipping this with release builds seam to not be that useful especially on devices where it's not possible to invoke commandline options

obligaron:
It's correct that this is only used for unit testing.
That's why the txt files aren't included in normal builds (cmake contains `if(BUILD_TESTING)` condition).
Do you think this is not enough or are you thinking about the `CreateDetailDiffs` function in `pfile.cpp`?

CMake/Assets.cmake
```diff
@@ -177,6 +177,24 @@ if(NOT USE_SDL1 AND NOT VITA)
     ui_art/menu.png)
 endif()
 
+if(BUILD_TESTING)
+  list(APPEND devilutionx_assets
```
glebm, line 181: It'd be better to place these in a separate directory, e.g. `test_assets`, so that they don't accidentally end up in a release.

They can then be opened using `SDL_RWFromFile` directory rather than going via `OpenAsset`.

glebm (changes requested):

CMake/Assets.cmake
glebm: Also, in the repo, they should probably live in `test/assets` or similar (instead of `Packaging/resources/assets`) and they should probably be set up in `test/CMakeLists.txt` (instead of `Assets.cmake`).

obligaron: I moved them to `test/fixtures` and `test/CMakeLists.txt`. 🙂

glebm (approved):
Thanks!

PR #5152, opened 2022-07-30
Inline functions which only set Object::_oRndSeed

ephphatha:


AJenbo (approved):
eventually i can close my PR instead of rebasing :D

PR #5155, opened 2022-07-31
Extract object/player reference clean-ups from #4571

ephphatha:
This leaves mainly the door logic clean-ups to be pulled from that PR.


AJenbo:
the door logic in #4571 is broken, i need to redo it from scratch :D

AJenbo (approved):

PR #5156, opened 2022-07-31
Clean up door sync

AJenbo:
Just the most basic change to make it simpler to understand the door logic later when refactoring it

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -3870,87 +3870,83 @@ void SyncPedestal(const Object &pedestal, Point origin, int width)
 	}
 }
 
-void SyncL1Doors(Object &door)
+void OpenDoor(Object &door)
 {
-	if (door._oVar4 == DOOR_CLOSED) {
-		door._oMissFlag = false;
-		return;
-	}
-
 	door._oMissFlag = true;
 	door._oSelFlag = 2;
 
-	bool isLeftDoor = door._otype == _object_id::OBJ_L1LDOOR; // otherwise the door is type OBJ_L1RDOOR
-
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -1426,7 +1434,7 @@ void ActivateTrapLine(int ttype, int tid)
 	for (int i = 0; i < ActiveObjectCount; i++) {
 		Object &trap = Objects[ActiveObjects[i]];
 		if (trap._otype == ttype && trap._oVar1 == tid) {
-			trap._oVar4 = 1;
+			trap._oVar4 = DOOR_OPEN;
```
ephphatha, line 1437: This isn't a door, this is setting a trap as being tripped/activated

Source/objects.cpp
```diff
@@ -1437,34 +1445,34 @@ void ActivateTrapLine(int ttype, int tid)
 void UpdateFlameTrap(Object &trap)
 {
 	if (trap._oVar2 != 0) {
-		if (trap._oVar4 != 0) {
+		if (trap._oVar4 != DOOR_CLOSED) {
```
ephphatha, line 1448: Same here, this function operates on a flamehole object, not a door.

AJenbo: Oops, lets try that again, but this time with a bit of clean up on top :D

ephphatha (approved):
Would prefer the door state enum to be a named and scoped enum but the usage makes that really messy 😂

AJenbo:
yeah, I'm not even sure it is currently being used in any place where we wouldn't need to do casting, at which point it becomes pretty pointless

PR #5157, opened 2022-07-31
ActorPosition: Remove offset, offset2 and velocity

obligaron:
This pr removes offset/offset2 and velocity from players and monsters (`ActorPosition`).

This reduces the size of `Monster` from 152 bytes to 144 and `Player` from 19776 to 19768. In total this is 1632 bytes.

offset was used for drawing walking players/monsters but with ADL we changed this to dynamic calculation (hight fps support).
Now offset is only used for lightning.

With this pr offset is also calculated dynamically for lightning and writing to the save game (save game compatibility with vanilla).
In the future the lightning part could also be obsolete when we have hardware lightning.

The changes in the demo recording is:

> file "game" has different content.
> Diff found in game.player.velocityX count: 1
> Diff found in game.player.velocityY count: 1

Powered by #5150 😉 

Commit `Add walking offset calculation to ActorPosition (and validate it)` contains my testing asserts that the dynamic calculation and the original logic is the same (checks was removed in last commit, cause the old logic was removed in it).

Note: The calculation is not 100% correct when town-jogging. That means when loading this in vanilla there could be a small visual glitch (one animation). But for devilutionX this is not relevant cause town doesn't use player lightning.


glebm:
How come test/fixtures/timedemo/WarriorLevel1to2/demo_0_reference_spawn_0.sv is 120 KiB larger?

glebm (approved):

obligaron:
> How come test/fixtures/timedemo/WarriorLevel1to2/demo_0_reference_spawn_0.sv is 120 KiB larger?

I'm not 100% sure, but I think it's cause save game was resized multiple times while testing.

What did I do?
I did run the test locally and copied my `demo_0_actual_spawn_0.sv` to `demo_0_reference_spawn_0.sv` after checking the diff with #5150.
Normally this shouldn't change the size much.
But the `demo_0_actual_spawn_0.sv` doesn't get reset when running the tests.
That means the last `demo_0_actual_spawn_0.sv` save game was reused and the save game could be resized (unused mpq blocks etc.).
That would explain the file size difference.
I redid my steps with a clean testing folder (without initial `demo_0_actual_spawn_0.sv`) and now the size is a little bit smaller.
Note: `CompareSaves` doesn't compare the compressed bytes but the uncompressed content. So it's understandable that there is no actual diff.

Anyway I think we should change the behavior for testing and creating the reference save game (not part of this pr).

AJenbo (approved):

Issue #5145, opened 2022-07-28
Input not working for split gold dialog


Chance4us:
It is no longer possible to drop a certain number of gold coins.
The enter key does not respond.
A detailed description will follow shortly.

Chance4us:
The issue appears since commit a4ac41cecef3df2d09db4d3958d5ff93bb920e48.

DakkJaniels:
what version are you running? It works fine for me in 1.4.

https://user-images.githubusercontent.com/6080734/182044540-72ba3821-5b01-4a12-95ed-3ceb6c5ef482.mp4

1

qndel:
@DakkJaniels master obviously since 1.4.1 doesn't contain https://github.com/diasurgical/devilutionX/commit/a4ac41cecef3df2d09db4d3958d5ff93bb920e48

DakkJaniels:
Is there an obvious way to tell that? That commit was 15 days ago and 1.4.1 was released 6 days ago. I know 1.4.1 doesn't contain everything, I'm just asking if there is an easy way to tell what commits were included and what were not.

StephenCWills:
If you click on the commit, GitHub will show which tags the commit applies to.

![image](https://user-images.githubusercontent.com/9203145/182045956-26663c46-3e46-40c3-b7d0-5b74cf288979.png)

AJenbo:
It's a bit less ovious just looking at this commit since it's not in any release yet:
![image](https://user-images.githubusercontent.com/204594/182046250-2ce575a1-3668-4d8e-985e-053a040c2157.png)

AJenbo:
@Chance4us https://github.com/Chance4us/devilutionX/blob/59482b5e9f9bb9c4d0b3a54eba5902f8df93998e/Source/diablo.cpp#L417

PR #5164, opened 2022-07-31
workaround drop gold input

Chance4us:
I would like to point out the issue so that it can be resolved quickly and professionally.

Workarounds issue #5145.

Creates a new issue:
When pressing the 0-9 keys except on the keypad, potions are consumed although you are in drop gold dialogue.


qndel:
pls don't open dummy pull requests like these, we are capable of finding relevant bits of code ourselves :) or you can post them in the issue

PR #5167, opened 2022-07-31
Merge more Doors logic

AJenbo:
Merge most of the high level door functions and split delta sync from the door functions. This revealed delta sync to be almost identical to `OpenDoor()`.

A bigger chunk of the door logic refactoring. Split in to a few logical commits. All in all this should still be mostly simple changes.

### Logic changes
`OpenDoor()` now sets
```cpp
door._oPreFlag = true;
door._oVar4 = DOOR_OPEN;
```

This was done during delta sync/direct operation so setting it on network sync should be correct.

-----

`CloseDoor()` now set:
```cpp
door._oMissFlag = false;
door._oSelFlag = 3;
```
This was done each frame by `UpdateDoor()` so L3 setting `true` and synced doors setting `2` was pretty pointless.

-----

`DeltaSyncOpObject()` is no longer called for `CMD_CLOSEDOOR`, doors are closed on spawn so this was probably just there to satisfy the call path being used previously.

P.s. this also completely removes the `deltaload` global

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1080,6 +1080,8 @@ void AddDoor(Object &door)
 	case OBJ_L3RDOOR:
 		ObjSetMicro(door.position, 533);
 		break;
+	default:
+		break;
 	}
 }
 
@@ -2048,6 +2050,8 @@ void OperateDoor(const Player &player, Object &door)
 		case OBJ_L5LDOOR:
 			OperateL5LDoor(door, true);
 			break;
+		default:
+			break;
 		}
 	}
 	if (dpx <= 1 && dpy == 1) {
@@ -2063,6 +2067,8 @@ void OperateDoor(const Player &player, Object &door)
 			break;
 		case OBJ_L5RDOOR:
 			OperateL5RDoor(door, true);
+			break;
+		default:
 			break;
 		}
 	}
@@ -2081,7 +2087,7 @@ bool AreAllLeversActivated(int leverId)
 	return true;
 }
 
-void DeltaOperateLever(Object &object)
+void UpdateLeverState(Object &object)
 {
 	if (object._oSelFlag == 0) {
 		return;
@@ -2110,7 +2116,7 @@ void OperateLever(Object &object, bool sendmsg)
 
 	PlaySfxLoc(IS_LEVER, object.position);
 
-	DeltaOperateLever(object);
+	UpdateLeverState(object);
 
 	if (currlevel == 24) {
 		PlaySfxLoc(IS_CROPEN, { UberRow, UberCol });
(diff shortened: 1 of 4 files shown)


Source/objects.cpp
```diff
@@ -2073,24 +2030,46 @@ void OperateL5LDoor(Object &door, bool sendflag)
...
+			OperateL3RDoor(door, true);
+			break;
+		case OBJ_L5LDOOR:
+			OperateL5LDoor(door, true);
+			break;
+		default:
+			break;
+		}
+	}
+	if (dpx <= 1 && dpy == 1 && door._otype == OBJ_L1RDOOR) {
```
ephphatha, line 2055:
```suggestion
	if (dpx <= 1 && dpy == 1) {
```

AJenbo: Must have pressed undo one to many times after trying to push teleflag in there :)

Source/objects.cpp
```diff
@@ -1404,23 +1395,33 @@ void ObjectStopAnim(Object &object)
...
 
+/**
+ * @brief Checks if an open door can be closed
+ *
+ * In order to be able to close a door the space where the closed door would be must be free of bodies, monsters, and items
+ *
+ * @param doorPosition Map tile where the door is in its closed position
+ * @return true if the door is free to be closed, false if anything is blocking it
+ */
+inline bool IsDoorClear(const Point &doorPosition)
```
ephphatha, line 1406: Where possible I've been trying to use constexpr over inline (doesn't make much difference, just acts as a stronger suggestion afaik), as long as the eventual `IsMonsterAtPosition` etc functions are also made constexpr this should be able to use that keyword.
```suggestion
constexpr bool IsDoorClear(const Point &doorPosition)
```

Source/objects.cpp
```diff
@@ -4891,29 +4801,24 @@ void OperateObject(Player &player, int i, bool teleFlag)
...
-		SyncOpL2Door(cmd, object);
-		break;
 	case OBJ_L3LDOOR:
 	case OBJ_L3RDOOR:
-		SyncOpL3Door(cmd, object);
-		break;
 	case OBJ_L5LDOOR:
 	case OBJ_L5RDOOR:
-		SyncOpL5Door(cmd, object);
+		object._oAnimFrame += 2;
```
ephphatha, line 4815: This is to capture a side-effect of `SyncOpL*Door` that's no longer triggered by the switch to `OpenDoor` right?

ephphatha (approved):
Looks much more readable, one question I wanted to double-check I understood the reason for since I couldn't see it called out in the commit messages.

Source/objects.cpp
AJenbo: `OpenDoor()` derives from a different call path then `SyncOpL*Door()`, but was the only meaningful difference. `_oAnimFrame` cannot be shared as it isn't idempotent an the other path already starts with the correct frame.

Source/objects.cpp
AJenbo: ill include it in the follow up as i have to touch this any way

AJenbo: turns out this doesn't build, so it's still `inline` in the other PR :)

PR #5169, opened 2022-08-01
Unify and clean up door logic

AJenbo:
Besides removing a lot of redundant code the now shared logic means that some original inconsistencies are now fixed:
- Doors could be drawn in the wrong order if opened by other players
- Crypt doors made the wrong sound when blocked

And one that we had introduced
- Crypt doors looked wrong after closing them

In total the logic has been reduced by over 500 lines.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1823,238 +1823,30 @@ void CloseDoor(Object &door)
 	SetDoorStateClosed(door);
 }
 
-void OperateL1RDoor(Object &door, bool sendflag)
+void OperateDoor(Object &door, bool sendflag)
 {
-	if (!IsDoorClear(door.position)) {
-		PlaySfxLoc(IS_DOORCLOS, door.position);
+	bool isCrypt = IsAnyOf(door._otype, OBJ_L5LDOOR, OBJ_L5RDOOR);
+
+	if (!IsDoorClear(door)) {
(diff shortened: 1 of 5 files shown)


Source/objects.cpp
```diff
@@ -4590,36 +4250,10 @@ void MonstCheckDoors(const Monster &monster)
...
-			} else if (door._otype == OBJ_L2RDOOR) {
-				OperateL2RDoor(door, true);
-			} else if (door._otype == OBJ_L3LDOOR) {
-				// L3 doors are backwards, see OperateDoor
-				OperateL3LDoor(door, true);
-			} else if (door._otype == OBJ_L5RDOOR) {
-				OperateL5RDoor(door, true);
-			}
-		}
+		OperateDoor(door, true);
```
ephphatha, line 4256: As an aside the only reason this loop operated in the specific order was to allow for branch prediction on the conditions that have now been removed. Could reorder that to go from `Direction::South` to `Direction::SouthEast` in a future cleanup.

ephphatha (approved):
Haven't playtested this but it looks decent from walking through the code. The bug in #5184 needs a fix implemented in msg.cpp:DeltaLoadLevel and objects.cpp:UpdateTrapState, which won't conflict with this PR.

AJenbo:
True, but I would still rather have this one merged first so I don't potentially have to debug in more places then necessary :)

PR #5174, opened 2022-08-01
Use position for object delta messages

### LINKED ISSUE #5172: Object state is not synced if you where not on the floor when the object was operated
Because object network messages are now validated based on object position `DeltaSyncObject()` will only be populated if you where present on a level with an object at the same location (most likely the same level)
https://github.com/diasurgical/devilutionX/blob/master/Source/msg.cpp#L1643

Broken in https://github.com/diasurgical/devilutionX/commit/de83d9440ef8b881945286a722cfa2d9fc947f2e#diff-3ca88f26b73a54abe9ce9eb981f4ab2ae1e454516047fa1854c37992d0afa7b8R1712

I'm not sure if it would be best to change deltasync to be x/y indexed or revert the change.

ephphatha:
In the worst case this ends up increasing the size of the DLevel object by 127*2 bytes (since we now need to track object position in the delta) + map overhead. Having said that deltas are only really needed for objects which actually allow interaction. Decorations no longer get a delta since they never change. Actual network traffic is compressed so should end up about the same in practice.

Static memory usage increases by 127*2+1 bytes unfortunately due to the use of `sgRecvBuf`. Hard to tell if we even need this chunk of memory to be on the stack given the use in `OnLevelData()`.

fixes #5172

### DIFF
--- a/Source/engine/world_tile.hpp
+++ b/Source/engine/world_tile.hpp
@@ -11,3 +11,18 @@ using WorldTileOffset = int8_t;
 using WorldTileDisplacement = DisplacementOf<WorldTileOffset>;
 
 } // namespace devilution
+
+namespace std {
+
+/**
+ * @brief Allows using WorldTilePosition as a map key for contexts where we want to lookup an entity by physical dungeon location
+ */
+template <>
+struct hash<devilution::WorldTilePosition> {
+	size_t operator()(const devilution::WorldTilePosition &position) const noexcept
+	{
+		return static_cast<size_t>(position.x) << 8 | position.y;
+	}
+};
+
+} // namespace std
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -72,7 +72,7 @@ struct DObjectStr {
 
 struct DLevel {
 	TCmdPItem item[MAXITEMS];
-	DObjectStr object[MAXOBJECTS];
+	std::unordered_map<WorldTilePosition, DObjectStr> object;
 	DMonsterStr monster[MaxMonsters];
 };
 
@@ -112,7 +112,10 @@ uint32_t sgdwRecvOffset;
 int sgnCurrMegaPlayer;
 std::unordered_map<uint8_t, DLevel> DeltaLevels;
 uint8_t sbLastCmd;
-byte sgRecvBuf[sizeof(DLevel) + 1];
+/**
+ * @brief buffer used to receive level deltas, size is the worst expected case assuming every object on a level was touched
+ */
+byte sgRecvBuf[1U + sizeof(DLevel::item) + sizeof(uint8_t) + (sizeof(WorldTilePosition) + sizeof(_cmd_id)) * MAXOBJECTS + sizeof(DLevel::monster)];
 _cmd_id sgbRecvCmd;
 std::unordered_map<uint8_t, LocalLevel> LocalLevels;
 DJunk sgJunk;
@@ -140,7 +143,8 @@ DLevel &GetDeltaLevel(uint8_t level)
 	auto emplaceRet = DeltaLevels.emplace(level, DLevel {});
 	assert(emplaceRet.second);
 	DLevel &deltaLevel = emplaceRet.first->second;
-	memset(&deltaLevel, 0xFF, sizeof(DLevel));
+	memset(&deltaLevel.item, 0xFF, sizeof(deltaLevel.item));
+	memset(&deltaLevel.monster, 0xFF, sizeof(deltaLevel.monster));
 	return deltaLevel;
 }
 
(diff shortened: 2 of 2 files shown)


Source/msg.cpp
```diff
@@ -2487,15 +2503,15 @@ void DeltaLoadLevel()
...
 	if (leveltype != DTYPE_TOWN) {
-		for (int i = 0; i < MAXOBJECTS; i++) {
-			switch (deltaLevel.object[i].bCmd) {
+		for (auto &objectDelta : deltaLevel.object) {
+			switch (objectDelta.second.bCmd) {
 			case CMD_OPENDOOR:
 			case CMD_OPERATEOBJ:
 			case CMD_PLROPOBJ:
-				DeltaSyncOpObject(Objects[i]);
+				DeltaSyncOpObject(ObjectAtPosition(objectDelta.first));
```
ephphatha, line 2511: This doesn't handle malicious messages that send an invalid position. Probably need to check an object actually exists and remove the delta if nothing is on the tile.

Source/engine/world_tile.hpp
```diff
@@ -11,3 +11,14 @@ using WorldTileOffset = int8_t;
...
 } // namespace devilution
+
+/**
+ * @brief Allows using WorldTilePosition as a map key for contexts where we want to lookup an entity by physical dungeon location
+ */
+template <>
+struct std::hash<devilution::WorldTilePosition> {
+	std::size_t operator()(const devilution::WorldTilePosition &position) const noexcept
+	{
+		return static_cast<size_t>(position.x) << 8 & position.y;
```
StephenCWills, line 22: Isn't this always zero?

StephenCWills:
```suggestion
		return static_cast<size_t>(position.x) << 8 | position.y;
```

Source/msg.cpp
```diff
@@ -140,6 +140,7 @@ DLevel &GetDeltaLevel(uint8_t level)
 	auto emplaceRet = DeltaLevels.emplace(level, DLevel {});
 	assert(emplaceRet.second);
 	DLevel &deltaLevel = emplaceRet.first->second;
+	// The following memset should not be necessary, DLevel was constructed three lines up, the default constructor should set appropriate values.
 	memset(&deltaLevel, 0xFF, sizeof(DLevel));
```
StephenCWills, line 144: The reason for the `memset()` is so that the logic can compare `TCmdPItem::bCmd` with `CMD_INVALID` to see if a specific delta has been initialized. 🤮 

https://github.com/diasurgical/devilutionX/blob/2c7e886c8503f7e0c04085d8ff0900c168095684/Source/msg.cpp#L556-L560

Source/engine/world_tile.hpp
AJenbo:
```suggestion
 		return static_cast<size_t>(position.x) << 8 | position.y; 
```

Source/msg.cpp
ephphatha: also uses monster.position.x being 255 as a test. Split the memsets for now.

Source/msg.cpp
```diff
@@ -2487,17 +2520,30 @@ void DeltaLoadLevel()
...
+				Object *object = FindObjectAtPosition(it->first);
+				if (object != nullptr) {
+					DeltaSyncBreakObj(*object);
+					it++;
+				} else {
+					it = deltaLevel.object.erase(it);
+				}
+			} break;
 			default:
+				it++;
```
ephphatha, line 2546:
```suggestion
				it = deltaLevel.object.erase(it); // discard invalid commands
```

Source/msg.cpp
```diff
@@ -2487,17 +2520,30 @@ void DeltaLoadLevel()
...
+				if (object != nullptr) {
+					DeltaSyncBreakObj(*object);
+					it++;
+				} else {
+					it = deltaLevel.object.erase(it);
+				}
+			} break;
 			default:
+				it = deltaLevel.object.erase(it); // discard invalid commands
 				break;
```
AJenbo, line 2547:
```cpp
			Object *object = FindObjectAtPosition(it->first);
			if (object == nullptr) {
				it = deltaLevel.object.erase(it);
				continue;
			}

			switch (it->second.bCmd) {
			case CMD_OPENDOOR:
			case CMD_OPERATEOBJ:
			case CMD_PLROPOBJ:
				DeltaSyncOpObject(*object);
				it++;
				break;
			case CMD_BREAKOBJ:
				DeltaSyncBreakObj(*object);
				it++;
				break;
			default:
				it = deltaLevel.object.erase(it); // discard invalid commands
				break;
			}
```

AJenbo (approved):

StephenCWills (approved):

glebm:
Getting a warning:

```
Source/msg.cpp:357:121: warning: operation on ‘src’ may be undefined [-Wsequence-point]
  357 |                 WorldTilePosition objectPosition { static_cast<WorldTileCoord>(*src++), static_cast<WorldTileCoord>(*src++) };
      |                                                                                                                      ~~~^~
```

ephphatha:
GCC warns on code that is valid in C++17 but was undefined before that standard, even if the language level is set to a modern version :/

Can use #5181 to avoid the warning

glebm:
This is actually useful for us because some platforms do not fully support C++17 (only Amiga by now actually)

PR #5177, opened 2022-08-02
Copy save file when creating reference or comparison save game

obligaron:
When writing save games for comparison (reference or actual) `pfile_write_hero` is called on a custom path.

This has two problems
1. `pfile_write_hero` writes doesn't write all data (only game state (including current level), hero and hotkeys etc.). But previous levels aren't written. That means the new created comparison save games don't contain all levels.
2. `pfile_write_hero` doesn't reset the save game, but only writes some data (see top 1). That means if somehow the save game existed before and did contain some other data (for example levels) they won't be erased.

This pr changes the behavior when creating comparison save games to copy the actual save game beforehand.
This way the save game includes all data (including levels) and no old data can be present in the save game.

Also fixes the issue described in this [comment](https://github.com/diasurgical/devilutionX/pull/5157#issuecomment-1200430673).

Diff
> file "perml00" is different size. Expected: 0 Actual: 25354
> file "perml01" is different size. Expected: 0 Actual: 160376

The levels are now correctly included. That's also why the new `demo_0_reference_spawn_0.sv` is bigger now.


AJenbo (dismissed):

AJenbo:
@obligaron could I get you to update the sv file since another PR updated it :/

glebm (approved):

obligaron:
No problem.
reference sv is updated 🙂

AJenbo (approved):

PR #5178, opened 2022-08-02
[android] Handle null pointers in Java

AJenbo:
Mostly minor issues. But the issue with accessing internal files have about 5 crashes related to it. Though I'm curious how you can end up with a device where the app doesn't have it's internal files folder... but hey we do not rely on it anyway.

### DIFF
--- a/android-project/app/src/main/java/org/diasurgical/devilutionx/DataActivity.java
+++ b/android-project/app/src/main/java/org/diasurgical/devilutionx/DataActivity.java
@@ -9,6 +9,7 @@ import android.content.IntentFilter;
 import android.database.Cursor;
 import android.net.Uri;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.text.method.LinkMovementMethod;
 import android.view.View;
 import android.widget.TextView;
@@ -162,7 +163,7 @@ public class DataActivity extends Activity {
 	 */
 	private class DownloadReceiver extends BroadcastReceiver {
 		@Override
-		public void onReceive(Context context, Intent intent) {
+		public void onReceive(@NonNull Context context, @NonNull Intent intent) {
 			long receivedID = intent.getLongExtra(DownloadManager.EXTRA_DOWNLOAD_ID, -1L);
 			DownloadManager mgr = (DownloadManager) context.getSystemService(Context.DOWNLOAD_SERVICE);
 
--- a/android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
+++ b/android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
@@ -5,8 +5,10 @@ import android.content.ClipData;
 import android.content.ContentResolver;
 import android.content.Intent;
 import android.net.Uri;
+import android.os.Build;
 import android.support.annotation.Nullable;
 import android.os.Bundle;
+import android.support.annotation.RequiresApi;
 import android.support.v4.provider.DocumentFile;
 import android.util.Log;
 
@@ -20,6 +22,7 @@ public class ImportActivity extends Activity {
 
 	private static final int IMPORT_REQUEST_CODE = 0xD1AB70;
 
+	@RequiresApi(api = Build.VERSION_CODES.KITKAT)
 	@Override
 	protected void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
@@ -60,7 +63,7 @@ public class ImportActivity extends Activity {
 		if (fileUri == null)
 			return;
 
-		DocumentFile file = DocumentFile.fromSingleUri(getApplicationContext(), fileUri);
+		DocumentFile file = Objects.requireNonNull(DocumentFile.fromSingleUri(getApplicationContext(), fileUri));
 		String fileName = file.getName();
 		String externalFilesDir = getExternalFilesDir(null).getAbsolutePath();
 		String externalFilesPath = externalFilesDir + "/" + fileName;
--- a/android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
+++ b/android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
@@ -90,7 +90,12 @@ public class ImportActivity extends Activity {
 					outputStream.close();
 			}
 		} catch (IOException exception) {
-			Log.e("importFile", Objects.requireNonNull(exception.getMessage()));
+			String message = exception.getMessage();
+			if (message == null) {
+				Log.e("importFile", "IOException", exception);
+			} else {
+				Log.e("importFile", message);
+			}
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
```diff
@@ -60,7 +63,7 @@ private void importFile(Uri fileUri) {
 		if (fileUri == null)
 			return;
 
-		DocumentFile file = DocumentFile.fromSingleUri(getApplicationContext(), fileUri);
+		DocumentFile file = Objects.requireNonNull(DocumentFile.fromSingleUri(getApplicationContext(), fileUri));
```
AJenbo, line 66: fromSingleUri only returns null if called on versions older then KitKat which we already guard against accessing this function, so simply throwing a null pointer exception at this point should never cause an issue.

tsunamistate:
@AJenbo is there a specific test case I should test for, or just do a smoke test? (Wow, that's a lot of "test")

AJenbo:
Strictly code review is just to see if you think the code looks sensible. But I won't stop you from also testing things ;) Especially as there can be a lot of variety in Android devices.

The changes here affects:
- Using the hold-app-icon and selecting import files to import a save game or mpq file
- Launching the game with save games from the legacy app on the device
- Launching the app with out any mpq files

If any of that seams easy to test go ahead, but don't do anything you think is a hazel.

tsunamistate:
Code review is NOT my thing, unfortunately. I am QA, not a programmer
Those cases are easy enough, I'll get back to you with the results soon

android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
```diff
@@ -20,6 +22,7 @@ public class ImportActivity extends Activity {
 
 	private static final int IMPORT_REQUEST_CODE = 0xD1AB70;
 
+	@RequiresApi(api = Build.VERSION_CODES.KITKAT)
```
yuripourre, line 25: I didn't look into the entire code but having this annotation on the onCreate code will prevent the app to run on devices prior to this. Maybe the part that handles the file could be extracted to a new method and this annotation moved there.

AJenbo: This is only accessible tought the shortcut list which is a feature that only exists in Android 25+, so preventing Android 18 from opening it isn't a problem.

android-project/app/build.gradle
```diff
@@ -65,4 +65,5 @@ android {
 dependencies {
 	implementation fileTree(include: ['*.jar'], dir: 'libs')
 	implementation 'com.android.support.constraint:constraint-layout:2.0.4'
+	implementation 'org.jetbrains:annotations:15.0'
```
ephphatha, line 68: Should this be a compileOnly dependency? Doesn't look like it's needed at runtime.

AJenbo: Don't know much about Java, it was added automatically by Android Studio (Jetbrain) when i added the annotation suggestion. If I had know that it was the automatic fixes that you guys thought where controversial I would have just left them out :joy:

ephphatha: 😂 surprised jetbrain doesn't use their own recommended config

https://github.com/JetBrains/java-annotations

```suggestion
	compileOnly 'org.jetbrains:annotations:23.0.0'
```

AJenbo: It did also initially add them in there long form and without importing the namespace, had to google a bit to make it compile :roll_eyes:

android-project/app/build.gradle
```diff
@@ -65,4 +65,5 @@ android {
 dependencies {
 	implementation fileTree(include: ['*.jar'], dir: 'libs')
 	implementation 'com.android.support.constraint:constraint-layout:2.0.4'
+	compileOnly 'org.jetbrains:annotations:23.0.0'
```
ephphatha, line 68: Looking at the imports in DataActivity/ImportActivity this package isn't being used, instead you're referencing android.support.annotation. The current version of that package is [androidx.annotation](https://mvnrepository.com/artifact/androidx.annotation/annotatio, not sure if it's a straightforward migration or something needs to be done to enable compiling for older targets?

AJenbo: Looks to work with out it, maybe I should just start ignoring there suggestions.

android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
```diff
@@ -87,7 +90,10 @@ private void importFile(Uri fileUri) {
 					outputStream.close();
 			}
 		} catch (IOException exception) {
-			Log.e("importFile", Objects.requireNonNull(exception.getMessage()));
+			String message = exception.getMessage();
+			if (message == null)
+				return;
+			Log.e("importFile", message);
```
yuripourre, line 96: When message is null, this call shouldn´t break but maybe would be more informative to print the entire exception so will output the stacktrace.

```suggestion
if (message == null) {
	Log.e("importFile", exception);
} else {
	Log.e("importFile", message);
}
```
Or simpler, we can ignore the message and print the exception only.
```java
	Log.e("importFile", exception);
```

android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
yuripourre: @AJenbo in that case, that's ok.

android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java
AJenbo: Thanks, this is my first time writing Java code :D

AJenbo: Apparently that isn't possible:

```
/home/runner/work/devilutionX/devilutionX/android-project/app/src/main/java/org/diasurgical/devilutionx/ImportActivity.java:97: error: incompatible types: IOException cannot be converted to String

	Log.e("importFile", exception);
> Task :app:compileDebugJavaWithJavac FAILED
```

ephphatha: try `log.e("importFile", "IOException", exception);`? https://developer.android.com/reference/android/util/Log#e(java.lang.String,%20java.lang.String,%20java.lang.Throwable)

yuripourre: @AJenbo sorry, @ephphatha is right, Log.e requires a third param to print an exception.
```java
if (message == null) {
	Log.e("importFile", "IOException", exception);
} else {
	Log.e("importFile", message);
}
```

android-project/app/src/main/java/org/diasurgical/devilutionx/DataActivity.java
```diff
@@ -9,6 +9,7 @@
 import android.database.Cursor;
 import android.net.Uri;
 import android.os.Bundle;
+import android.support.annotation.NonNull;
```
yuripourre, line 12: This import is better than the one suggested by IntelliJ, indeed.

yuripourre (approved):

PR #5180, opened 2022-08-02
Get rid of dthread

vlolteanu:
I'm 95% certain that dthread is useless. In vanilla, it used to add some delay in between some lower-priority messages on low baud connections. Now it's just a message queue, but all messages are sent asynchronously anyway. So I got rid of it.

A second opinion would be helpful.


Source/multi.cpp
```diff
@@ -333,14 +332,13 @@ void ProcessTmsgs()
 void SendPlayerInfo(int pnum, _cmd_id cmd)
 {
 	static_assert(alignof(PlayerPack) == 1, "Fix pkplr alignment");
```
AJenbo, line 334:
```suggestion
	static_assert(alignof(PlayerPack) == 1, "Fix pPack alignment");
```

Also is that still needed?

AJenbo:
Anyone up for helping with play testing this a bit :)

StephenCWills (dismissed):
Nice, I was also wondering why this thread exists. Good riddance.

Source/multi.cpp
vlolteanu: No. Fixed.

StephenCWills: Isn't it important for the `reinterpret_cast<byte *>(&pPack)`?

vlolteanu: It would only matter if `alignof(byte)` were larger than `alignof(PlayerPack)`.

StephenCWills (approved):

AJenbo (approved):

PR #5181, opened 2022-08-02
Prevent sequence-point warning in DeltaImportObjects when compiling with GCC

vlolteanu:
Doing `foo++` twice is a big no-no.

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -355,7 +355,8 @@ const byte *DeltaImportObjects(const byte *src, std::unordered_map<WorldTilePosi
 	dst.reserve(numDeltas);
 
 	for (unsigned i = 0; i < numDeltas; i++) {
-		WorldTilePosition objectPosition { static_cast<WorldTileCoord>(*src++), static_cast<WorldTileCoord>(*src++) };
+		WorldTilePosition objectPosition { static_cast<WorldTileCoord>(src[0]), static_cast<WorldTileCoord>(src[1]) };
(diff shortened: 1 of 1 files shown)


Source/msg.cpp
```diff
@@ -355,7 +355,8 @@ const byte *DeltaImportObjects(const byte *src, std::unordered_map<WorldTilePosi
 	dst.reserve(numDeltas);
 
 	for (unsigned i = 0; i < numDeltas; i++) {
-		WorldTilePosition objectPosition { static_cast<WorldTileCoord>(*src++), static_cast<WorldTileCoord>(*src++) };
+		WorldTilePosition objectPosition { static_cast<WorldTileCoord>(src[0]), static_cast<WorldTileCoord>(src[1]) };
+		src += 2;
```
ephphatha, line 359: This was/is using [braced list initialisation](https://en.cppreference.com/w/cpp/language/list_initialization).

> Every initializer clause is [sequenced before](https://en.cppreference.com/w/cpp/language/eval_order) any initializer clause that follows it in the braced-init-list.

This should guarantee that all side-effects of the first post-increment are evaluated (and stored) before the second post-increment is executed.

vlolteanu: I guess I learn something new every day. Cheers for that!

ephphatha:
Looks like we should implement this change to remove a defensive warning anyway :D

glebm (approved):

PR #5182, opened 2022-08-03
Split delta trap operation from realtime trap operation

StephenCWills:
Fixes an issue reported by user Malignus on Discord.

> When you go in a level where there are traps, if you go to town and take back the town portal, all the traps will re-activate themselves upon entry of TP.
> 
> So if you opened all chests and doors with traps fireball and arrows will be sent through the rooms.

The issue was introduced by #4805, where the condition that checked for `deltaload` was removed.


AJenbo (approved):
Thanks :)

PR #5183, opened 2022-08-03
Convert assets to CLX

glebm:
Converted using the following commands:

```
cd Packaging/resources/assets
pcx2clx --transparent-color 1 --num-sprites 256 fonts/*.pcx && rm fonts/*.pcx
pcx2clx --num-sprites 2 ui_art/dvl_but_sml.pcx && rm ui_art/dvl_but_sml.pcx
pcx2clx --transparent-color 1 data/hintbox.pcx data/hintboxbackground.pcx && rm data/hintbox.pcx data/hintboxbackground.pcx
pcx2clx --transparent-color 1 --num-sprites 6 data/hinticons.pcx && rm data/hinticons.pcx
pcx2clx --num-sprites 2 data/panel8buc.pcx data/dirtybuc.pcx data/dirtybucp.pcx && rm data/panel8buc.pcx data/dirtybuc.pcx data/dirtybucp.pcx
pcx2clx --transparent-color 1 data/healthbox.pcx && rm data/healthbox.pcx
pcx2clx --transparent-color 1 --num-sprites 6 data/resistance.pcx && rm data/resistance.pcx
pcx2clx --transparent-color 1 --num-sprites 5 data/monstertags.pcx && rm data/monstertags.pcx
pcx2clx --transparent-color 1 data/stash.pcx && rm data/stash.pcx
pcx2clx --transparent-color 1 --num-sprites 5 data/stashnavbtns.pcx && rm data/stashnavbtns.pcx
pcx2clx **/*.pcx && rm **/*.pcx
git checkout master -- ui_art/black_diablo.pcx ui_art/black_hellfire.pcx data/charbg.pcx data/talkbutton.pcx data/panel8bucp.pcx
rm ui_art/black_diablo.clx ui_art/black_hellfire.clx data/charbg.clx data/talkbutton.clx data/panel8bucp.clx
```

`pcx2clx` is from https://github.com/diasurgical/devilutionx-graphics-tools

`fonts.mpq` will need to be re-built (https://github.com/diasurgical/devilutionx-assets/pull/7).

Size of `Packaging/resources/assets/`: 6.1 MiB -> 4.7 MiB

PR #5188, opened 2022-08-04
More CLX rendering

glebm:
Fixes a couple of things missed in #5183 and converts pre-composed surfaces in `charpanel` and `mainpanel` to CLX.

Now the only remaining use of `Art` is the `ArtCursor` -- will be removed in a follow-up PR along with all the `Art` code.


Source/utils/surface_to_clx.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <iomanip>
+#include <iostream>
+#endif
+
+namespace devilution {
+
+OwnedClxSpriteList SurfaceToClx(const Surface &surface, unsigned numFrames,
+    std::optional<uint8_t> transparentColor)
+{
+	// CL2 header: frame count, frame offset for each frame, file size
```
AJenbo, line 19: Old comment

Source/panels/mainpanel.hpp
```diff
@@ -1,14 +1,12 @@
...
-extern Art TalkButton;
-
-void LoadMainPanel();
-void FreeMainPanel();
-
-} // namespace devilution
+#pragma once
+
+#include "engine/clx_sprite.hpp"
+namespace devilution {
```
AJenbo, line 4:
```suggestion
#include "engine/clx_sprite.hpp"

namespace devilution {
```

Source/panels/charpanel.cpp
```diff
@@ -269,42 +265,45 @@ void DrawStatButtons(const Surface &out)
...
-
-	const bool isSmallFontTall = IsSmallFontTall();
-	const int attributeHeadersY = isSmallFontTall ? 112 : 114;
-	for (unsigned i : AttributeHeaderEntryIndices) {
-		panelEntries[i].position.y = attributeHeadersY;
-	}
-	panelEntries[GoldHeaderEntryIndex].position.y = isSmallFontTall ? 105 : 106;
+	OptionalOwnedClxSpriteList background = LoadOptionalClx("data\\charbg.clx");
+	if (!background)
+		app_fatal(_("Please update devilutionx.mpq to the latest version"));
```
AJenbo, line 270: Is this specifically because `charbg.clx` was missed in the first round? I don't think we should really give error messages for in between version changes, we don't support development versions in general. So maybe check should be in the init where devilutionx.mpq is loaded, similar to how diabdat.mpq is tested for by probing title.pcx.

Source/panels/mainpanel.cpp
```diff
@@ -56,72 +57,85 @@ void RenderMainButton(int buttonId, string_view text, int frame)
...
-	RenderMainButton(2, _("map"), 1);
-	RenderMainButton(3, _("menu"), 0);
-	RenderMainButton(4, _("inv"), 1);
-	RenderMainButton(5, _("spells"), 0);
+	std::optional<OwnedSurface> out;
+	constexpr uint16_t NumButtonSprites = 6;
+	{
+		OptionalOwnedClxSpriteList background = LoadOptionalClx("data\\panel8bucp.clx");
+		if (!background)
+			app_fatal(_("Please update devilutionx.mpq to the latest version"));
```
AJenbo, line 72: Since this is a mandatory component the other check seams redundant?

Source/panels/charpanel.cpp
glebm: Done

Source/panels/mainpanel.cpp
glebm: Done

Issue #5187, opened 2022-08-03
lazarus lair teleports bug


bazilio-ua:
### Operating System

Mac

### DevilutionX version

Custom build (please specify commit ID)

### Describe

build from commit: 9a55b1334

### To Reproduce

In lazarus lair teleports into corners jail's and to lazarus room itself don't triggered
![Screenshot 2022-08-04 at 00 20 33](https://user-images.githubusercontent.com/2544903/182716090-6d4feb40-8868-4ddf-b9af-773678a864b2.png)
[single_0.sv.zip](https://github.com/diasurgical/devilutionX/files/9255170/single_0.sv.zip)

PR #5190, opened 2022-08-04
fix interacting with lazarus books while not on circles

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/5187
Basically reverts part of https://github.com/diasurgical/devilutionX/commit/56915974d60eded7d7199a915cb7ddb904ef842d

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1908,14 +1908,7 @@ void OperateBook(Player &player, Object &book)
 		return;
 	}
 
-	book._oSelFlag = 0;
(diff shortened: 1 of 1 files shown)


AJenbo:
Considering that this is 8 lines and the other one was 147 I don't think it is exactly reverting all of it :D

AJenbo (approved):

qndel:
> Considering that this is 8 lines and the other one was 147 I don't think it is exactly reverting all of it :D

image
;p

PR #5197, opened 2022-08-06
Ensure valid animation state when loading death animation in InitPlayer()

StephenCWills:
Fixes this error...
image

**To reproduce...**
1. On client 1, create a new warrior and start a TCP game
2. On client 2, create another new warrior and join the TCP game
3. On client 2, use the "New Game" option in the menu to leave the game
4. On client 2, immediately rejoin the game
5. You should see the error on client 1

---

The error was introduced with the recent CLX changes, but the conditions for the error existed long before that. This fix follows the pattern used in a handful of other places throughout the codebase that also load the `PM_DEATH` animation. For example...

https://github.com/diasurgical/devilutionX/blob/52780a37125452f9ac5cb5010af7071e0ecd10f1/Source/msg.cpp#L1836-L1838


obligaron (approved):
lgtm 🙂

PR #5199, opened 2022-08-06
CelToClx: Fix writing frame count with multiple groups

obligaron:
When converting cel to clx with a graphic that has multiple groups the number of frames (per group) is not correctly written.
This results in an [assert error](https://github.com/diasurgical/devilutionX/blob/master/Source/engine/clx_sprite.hpp#L130) when going to the cows in town.


AJenbo:
Also just noticed that this file still refer to it as CL2 and not CLX :)

PR #5200, opened 2022-08-06
Fix draw flag for MIS_HORKDMN (set to no drawing)

### LINKED ISSUE #5198: Hork Demon crash
The game crashes when the Hork Demon tries to generate hork spawns.
At least that was the impression.
More details will follow soon.

obligaron:
Fixes #5198 

The hork spawn missile has no animation associated (`MFILE_NONE`) but `MissilesData` says it should be rendered (`mDraw`).
This is incorrect (also in 1.4.1).

In 1.4.1 this doesn't result in a crash, cause of [checks in drawing](https://github.com/diasurgical/devilutionX/blob/1.4/Source/scrollrt.cpp#L312-L315). But we get a log warning:
> INFO: Draw Missile 2 type 95: NULL Cel Buffer

@glebm I'm unsure if we should change the behavior (in a separate pr) when drawing incorrect animations (change back to log instead of appfatal?).

Thanks @Chance4us for reporting. 🙂 

Side-Note:
When loading a original hellfire game and the missile is spawned this could still result in a crash, cause the incorrect `MissilesData` was used to initialize the `Missile::_miDrawFlag`.


glebm:
I think a crash is good, since we apparently missed this when we only logged it! Thanks for the fix!

AJenbo:
If anything I think an asset could be an option

PR #5204, opened 2022-08-06
Consistently send network messages when monsters are damaged

StephenCWills:
Replace instances of `monster.hitpoints -= damage` with `ApplyMonsterDamage(monster, damage)`.

This change ensures that network messages are sent to remote clients whenever the local client registers a source of damage on a monster, reducing cases of desync when a monster is hit by a trap, reflect, or a spell that the monster resists. It resolves the issue described in #5109.

Regarding the discussion in #5147, I chose to reassociate the sound effect with the hit recovery logic because of network desync. In most cases where damage is applied to monsters by a remote player, the damage is ignored but the hit recovery is not. This gives the game the appearance that the remote player's attacks are landing with the proper timing without amplifying damage by duplicating hit chance and damage rolls.

Note that some sources of damage are causing duplicate hit chance and damage rolls (traps, golems, reflect). This amplifies the deadliness of these effects, which was previously somewhat mitigated by not synchronizing these sources of damage.

Also, by moving the network messages from `M_StartHit()` to `ApplyMonsterDamage()`, godmode desyncs worse than it did before. I assume this is acceptable and so didn't bother complicating the logic just to better synchronize it.


AJenbo:
Before this pr I noticed that monsters where sometimes dying on the other end of the network before you had killed it locally. Do you know if that could be an issue with god mode? (I don't remember if it was on or not).

AJenbo:
Do you have any idea for how the potential double damage from traps can be resolved?

StephenCWills:
> Before this pr I noticed that monsters where sometimes dying on the other end of the network before you had killed it locally. Do you know if that could be an issue with god mode? (I don't remember if it was on or not).

I don't know for sure, but there is a possibility. Godmode doesn't discriminate against remote players the way that damage does. If client 1 hits a monster and client 2 has godmode on, the monster will die on client 2 regardless of what happens on client 1. In this situation, client 2 will fail to send the `CMD_MONSTDEATH` packet because `&player != MyPlayer`.

> Do you have any idea for how the potential double damage from traps can be resolved?

The simplest solution would be to select one client to handle all the damage packets for each trap. This could be done in a similar fashion to how one player is selected to send deltas to new players.

ephphatha:
> Do you have any idea for how the potential double damage from traps can be resolved?

The effects of traps are being resolved as a side effect each tick after running through the object processing loop, which means every client runs the logic locally and generates a damage event. I've been trying to move this to an explicit call when a trapped object is interacted with but it's made really annoying by two different types of traps applying to chests (and the multiple different objects which can potentially be trapped all having different update functions)... Eventually it should be that the player who interacted with the trapped object also is responsible for generating the damage event.

Issue #5205, opened 2022-08-06
"questmap" debug command is broken


qndel:
### Operating System

Windows x64

### DevilutionX version

Custom build (please specify commit ID)

### Describe

https://github.com/diasurgical/devilutionX/commit/e68f2b46da679f81082f5e89423ed3c066283e69

doesn't warp to quest levels - I suspect it's treating quest level as normal level - trying to warp to lazarus quest (ID 5) displayed catacombs loading screen while lazarus quest is in hell -

### To Reproduce

use "questmap" debug command

qndel:
Broken by https://github.com/diasurgical/devilutionX/commit/ba34bafb3fac32391c12aba5bf1bd8e5430ccde3

PR #5206, opened 2022-08-06
fix debug questmap command

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/5205
No idea why :D (Thanks @obligaron )


Source/debug.cpp
```diff
@@ -238,6 +238,7 @@ std::string DebugCmdLoadQuestMap(const string_view parameter)
 		if (!MyPlayer->isOnLevel(quest._qlevel)) {
 			StartNewLvl(*MyPlayer, (quest._qlevel != 21) ? interface_mode::WM_DIABNEXTLVL : interface_mode::WM_DIABTOWNWARP, quest._qlevel);
 			ProcessMessages();
+			ProcessMessages();
```
obligaron, line 241: I know it's only a debug command but perhaps we should add a explaining topic for this workaround?

```suggestion
			// Workaround for SDL_PollEvent:
			// StartNewLvl pushes a new event with SDL_PushEvent.
			// ProcessMessages calls SDL_PollEvent but SDL ignores the new pushed event.
			// Calling SDL_PollEvent again fixes this.
			ProcessMessages();
```

obligaron (approved):

Issue #5031, opened 2022-07-17
[UI/Enhancement] Display difficulty clear status for each character


adpa:
I recall the Battle.net lobby UI for Diablo displaying icons with dots for every difficulty that character has killed Diablo on.

To my knowledge, this was the only indication of a given character's difficulty clear status in game.

I think it would be rather nice if the same (or similar) icons could be added to the character select menu and/or the character status pane in game in order to be able to see difficulty clear status for single player and offline play.

AJenbo:
Single player does not automatically save so this information is only available for multiplayer heros

adpa:
Fair point on the single player limitation, but I still think the idea has merit for multiplayer characters. Given that DX doesn't support Battle.net (are the servers even up for the vanilla client at this point?), the difficulty clear status is kind of lost as it stands today.

I recall getting your dots as being sort of an aspirational thing for B.net players back in the day. It would be nice to recreate that feeling for DX.

AJenbo:
You mentioned specifically wanting this for offline play so didn't think you wanted it for multiplayer?

adpa:
Ah, sorry, I guess that's poor wording on my part. I intended for "offline" to refer to non-B.net multiplayer play. I mostly play my multiplayer characters solo via the direct connect mode, which doesn't require connecting to any external servers. Sorry for the confusion.

AJenbo:
Ok, In DevilutionX you have loopback for that, it doesn't even listen to a network interface. So yeah that is offline, but would exclude ZeroTier and possibly TCP depending on your definition.

adpa:
Fair point. It was not my intent to exclude the ZeroTier or TCP use cases from consideration.

PR #5210, opened 2022-08-07
show difficulty indicator on multiplayer characters

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/5031
image

```cpp
myPlayer.pDiabloKillLevel = std::max(myPlayer.pDiabloKillLevel, static_cast<uint8_t>(sgGameInitInfo.nDifficulty + 1));
```
this happens in PrepDoEnding  which means showing difficulty thingies works only in multiplayer - multi autosaves after outro so the value is preserved while in singleplayer you have no way to save after it got set.
That's why I've added 
```cpp
	if (!gbIsMultiplayer) {
		Player &myPlayer = *MyPlayer;
		myPlayer.pDiabloKillLevel = std::max(myPlayer.pDiabloKillLevel, static_cast<uint8_t>(sgGameInitInfo.nDifficulty + 1));
	}
```
to DiabloDeath to compensate for that - the variable will increase by 2 in singleplayer after outro, but since you have no way to save the second increase, it's all good. You can only save the increased variable if you save the game while diablo is dying. This will color a gem above your character's portrait.

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -44,6 +44,7 @@
 namespace devilution {
 
 OptionalOwnedClxSpriteList ArtLogo;
+std::array<OptionalOwnedClxSpriteList, 2> DifficultyIndicator;
 
 std::array<OptionalOwnedClxSpriteList, 3> ArtFocus;
 
@@ -562,6 +563,8 @@ void LoadUiGFX()
 	} else {
 		ArtLogo = LoadPcxSpriteList("ui_art\\smlogo.pcx", /*numFrames=*/15, /*transparentColor=*/250);
 	}
+	DifficultyIndicator[0] = LoadPcx("ui_art\\radio1.pcx");
+	DifficultyIndicator[1] = LoadPcx("ui_art\\radio3.pcx");
 	ArtFocus[FOCUS_SMALL] = LoadPcxSpriteList("ui_art\\focus16.pcx", /*numFrames=*/8, /*transparentColor=*/250);
 	ArtFocus[FOCUS_MED] = LoadPcxSpriteList("ui_art\\focus.pcx", /*numFrames=*/8, /*transparentColor=*/250);
 	ArtFocus[FOCUS_BIG] = LoadPcxSpriteList("ui_art\\focus42.pcx", /*numFrames=*/8, /*transparentColor=*/250);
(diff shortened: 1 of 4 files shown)


Source/DiabloUI/diabloui.cpp
```diff
@@ -563,6 +564,7 @@ void LoadUiGFX()
 	} else {
 		ArtLogo = LoadPcxSpriteList("ui_art\\smlogo.pcx", /*numFrames=*/15, /*transparentColor=*/250);
 	}
+	DifficultyIndicator = LoadPcxSpriteList("ui_art\\radio1.pcx", /*numFrames=*/1, /*transparentColor=*/250);
```
glebm, line 567: You can use `LoadPcx` when it's just a single frame

Source/DiabloUI/selhero.cpp
```diff
@@ -565,6 +565,13 @@ static void UiSelHeroDialog(
 		while (!selhero_endMenu && !selhero_navigateYesNo) {
 			UiClearScreen();
 			UiRenderItems(vecSelHeroDialog);
+			std::vector<std::unique_ptr<UiItemBase>> tmp;
+			const Point uiPosition = GetUIRectangle().position;
+			for (int i = 0; i < selhero_heroInfo.herorank; i++) {
+				SDL_Rect rect = { (Sint16)(uiPosition.x + 30) + i * 12, (Sint16)(uiPosition.y + 275), 12, 12 };
```
glebm, line 571: Use `MakeSdlRect` to avoid all the casts.

Source/DiabloUI/selhero.cpp
```diff
@@ -565,6 +565,13 @@ static void UiSelHeroDialog(
 		while (!selhero_endMenu && !selhero_navigateYesNo) {
 			UiClearScreen();
 			UiRenderItems(vecSelHeroDialog);
+			std::vector<std::unique_ptr<UiItemBase>> tmp;
+			const Point uiPosition = GetUIRectangle().position;
+			for (int i = 0; i < selhero_heroInfo.herorank; i++) {
+				SDL_Rect rect = { (Sint16)(uiPosition.x + 30) + i * 12, (Sint16)(uiPosition.y + 275), 12, 12 };
+				tmp.push_back(std::make_unique<UiImageAnimatedClx>(*DifficultyIndicator, rect, UiFlags::None));
```
glebm, line 572: That creates a vector on every frame. Let's just create / update it on hero change?

glebm (approved):

PR #5214, opened 2022-08-07
Move some monster initialization to initializing values in monsterData

aetn23:
Move some part of monster initialization from `InitMonster` to new functions in monstdata.cpp. This is done to open a door to removing some members from monster. 

Some information in monster is dependent solely on values form `monsterData` and difficulty. As such, it is unnecessary to calculate them every time a monster is created - it is sufficient to initialize `monsterData` every time a game is loaded (meaning rarely) and simply fetch the data from the array. This required some functions to change back and forth between difficulties.

Of course, some members of monster are now redundant, because they can simply be fetched form `monsterData` - I will address that in a separate PR since this is one is big already. Actually, the possibility of emoval of these values from monster is the reason this PR exists.

I wrote a test to make sure this new initialization produces the same results as the old one. It helped me catch few bugs already (including the one with exp overflow I addressed in other PR).

This PR has #5070 merged into it, since without this fix I would have to provide some ugly workaround.


glebm:
Windows build fails with:

```
  monster_init_test.cpp.obj : error LNK2001: unresolved external symbol "struct devilution::MonsterData * devilution::MonstersData" (?MonstersData@devilution@@3PAUMonsterData@1@A)
  monster_init_test.cpp.obj : error LNK2001: unresolved external symbol "struct devilution::CMonster * devilution::LevelMonsterTypes" (?LevelMonsterTypes@devilution@@3PAUCMonster@1@A)
    monster_init_test.cpp.obj : error LNK2001: unresolved external symbol "struct devilution::GameData devilution::sgGameInitInfo" (?sgGameInitInfo@devilution@@3UGameData@1@A)
```

All the globals used in tests must be marked with `extern DVL_API_FOR_TEST` for the MSVC linker to see them.

test/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ target_link_libraries(libdevilutionx_so PUBLIC libdevilutionx)
 target_include_directories(libdevilutionx_so INTERFACE "${DevilutionX_SOURCE_DIR}/Source")
 set_target_properties(libdevilutionx_so PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
 
-add_library(test_main STATIC main.cpp)
+add_library(test_main STATIC main.cpp monster_init_test.cpp)
```
AJenbo, line 9: This isn't normally where we add tests :thinking:

Source/monstdat.cpp
```diff
@@ -23,7 +24,7 @@ constexpr uint16_t Uniq(_unique_items item)
 } // namespace
 
 /** Contains the data related to each monster ID. */
-const MonsterData MonstersData[] = {
+MonsterData MonstersData[] = {
```
AJenbo, line 27: Manipulating the table during runtime doesn't sound like a good idea :thinking:

AJenbo: Also should use MonstersDataSize to specify the size if you are adding it so they don't go out of sync.

aetn23:
After looking at this with a fresh eye, and some valid points about not touching `MonstersData` during runtime, I think maybe it's better to drop this approach, and simply have a monster apply difficulty modifier on a fly (Basically doing what #5070 did for exp for other variables)? That would be less code, no global table would be modified, and it will easy to see what is happening just by looking at `Monster`. 

The only downside would be that a monster would perform some basic additions and ifs every time a game would need, say, its `armorClass`

What do you guys think?

glebm:
> The only downside would be that a monster would perform some basic additions and ifs every time a game would need, say, its `armorClass`

That seems reasonable, the time spent in such calculations is tiny.

aetn23:
I close this, as the approach mentioned earlier is a better one. Soon, I will add new PRsto address it.

PR #5217, opened 2022-08-08
Synchronize player inventories

StephenCWills:
Synchronizes changes to the player's inventory to remote clients. There are definitely some missing edge cases, but this can get us started on resolving many of the sources of desync discussed in #4074.

Here is a video demonstration of how the synchronization looks when testing in conjunction with https://github.com/StephenCWills/devilutionX/commit/5df65ff54f6ee6fef6d13cb23e01d9710f4a0060:

https://user-images.githubusercontent.com/9203145/183315099-ff6d09f9-d43a-4304-bc2a-72db5a11a4b3.mp4

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1767,6 +1767,34 @@ size_t OnDeleteInventoryItems(const TCmd *pCmd, int pnum)
 	return sizeof(message);
 }
 
+size_t OnChangeBeltItems(const TCmd *pCmd, int pnum)
+{
+	const auto &message = *reinterpret_cast<const TCmdChItem *>(pCmd);
+	Player &player = Players[pnum];
+
+	if (gbBufferMsgs == 1) {
+		SendPacket(pnum, &message, sizeof(message));
+	} else if (&player != MyPlayer && message.bLoc < MaxBeltItems && message.wIndx <= IDI_LAST) {
+		CheckBeltSwap(player, message.bLoc, message.wIndx, message.wCI, message.dwSeed, message.bId != 0, message.dwBuff);
+	}
+
+	return sizeof(message);
+}
+
+size_t OnDeleteBeltItems(const TCmd *pCmd, int pnum)
+{
+	const auto &message = *reinterpret_cast<const TCmdParam1 *>(pCmd);
+	Player &player = Players[pnum];
+
+	if (gbBufferMsgs == 1) {
+		SendPacket(pnum, &message, sizeof(message));
+	} else if (&player != MyPlayer && message.wParam1 < MaxBeltItems) {
+		player.RemoveSpdBarItem(message.wParam1);
+	}
+
+	return sizeof(message);
+}
+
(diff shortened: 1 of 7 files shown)


Source/msg.cpp
```diff
@@ -2893,6 +2949,47 @@ void NetSendCmdDelItem(bool bHiPri, uint8_t bLoc)
...
+	int8_t invListIndex = abs(MyPlayer->InvGrid[invGridIndex]) - 1;
+	Item &item = MyPlayer->InvList[invListIndex];
+
+	cmd.bCmd = CMD_CHANGEINVITEMS;
+	cmd.bLoc = invGridIndex;
+	cmd.wIndx = item.IDidx;
+	cmd.wCI = item._iCreateInfo;
+	cmd.dwSeed = item._iSeed;
+	cmd.bId = item._iIdentified ? 1 : 0;
+	cmd.dwBuff = item.dwBuff;
```
obligaron, line 2965: We have this logic now in three places.
Could we add a helper method `SetItemForTCmdChItem` or something like this?

Source/msg.cpp
```diff
@@ -2893,6 +2949,47 @@ void NetSendCmdDelItem(bool bHiPri, uint8_t bLoc)
 		NetSendLoPri(MyPlayerId, (byte *)&cmd, sizeof(cmd));
 }
 
+void NetSendCmdChInvItem(bool bHiPri, int invGridIndex)
+{
+	TCmdChItem cmd;
+
+	int8_t invListIndex = abs(MyPlayer->InvGrid[invGridIndex]) - 1;
+	Item &item = MyPlayer->InvList[invListIndex];
```
obligaron, line 2957:
```suggestion
	const Item &item = MyPlayer->InvList[invListIndex];
```

Source/msg.cpp
```diff
@@ -2893,6 +2949,47 @@ void NetSendCmdDelItem(bool bHiPri, uint8_t bLoc)
...
+		NetSendHiPri(MyPlayerId, (byte *)&cmd, sizeof(cmd));
+	else
+		NetSendLoPri(MyPlayerId, (byte *)&cmd, sizeof(cmd));
+}
+
+void NetSendCmdChBeltItem(bool bHiPri, int beltIndex)
+{
+	TCmdChItem cmd;
+
+	Item &item = MyPlayer->SpdList[beltIndex];
```
obligaron, line 2977:
```suggestion
	const Item &item = MyPlayer->SpdList[beltIndex];
```

obligaron (approved):
lgtm, only some nits 🙂 
Thanks for working on this ❤️

Source/msg.cpp
StephenCWills: I'm hesitant to do that just yet because I will probably need to replace two of these places with `TCmdPItem` instead.

obligaron: I'm fine with leaving them for now and refactor them later. 🙂

AJenbo (approved):

PR #5220, opened 2022-08-08
Migrate monstdat to a data file 

glebm:
We may want to migrate this to 1 file per monster but for now the format is as close to the hard-coded version as possible.

Sprites that are used by multiple monsters are only loaded from disk once.

With the CSV editor extension in VS Code:

image


glebm:
Rebased

glebm:
Something is wrong with the golem, the logged sizes are all over the place:

```
VERBOSE: Loaded monster graphics: golem\golem      909 KiB   x1
VERBOSE: Loaded monster graphics: golem\golem        0 KiB   x1
VERBOSE: Loaded monster graphics: golem\golem     1306 KiB   x1
```

AJenbo:
Yes!

Source/monstdat.cpp
```diff
@@ -335,6 +201,450 @@ const _monster_id MonstConvTbl[] = {
 	MT_LRDSAYTR,
 };
 
+// P_("monster", "A \"monster\"") => "A \"monster\""
+tl::expected<std::string, std::string> ParsePgettext(std::string_view value)
+{
+	constexpr std::string_view Prefix = "P_(\"";
+	constexpr std::string_view Suffix = "\")";
+	if (!value.starts_with(Prefix) || !value.ends_with(Suffix)) {
+		return tl::make_unexpected("Not a P_-string");
```
AJenbo, line 210: I think it would be better to simply return the string here, the monster name simply isn't translatable at that point. But that is also allowed now.

AJenbo:
@glebm I just have that one single comment, but also now that debug.cpp has been rewritten this is going to need a little rebase if you have the time :)

AJenbo:
Ok, it looks like you haven't made any adjustments for picking up the monster names when translation files are updated.
Do you have any thought about how that should be handled?
Maybe we could actually just crab the second column and do away with the handeling of `P_("monster", )`.

AJenbo:
> We may want to migrate this to 1 file per monster

I actually much prefer it as a table of monsters. I'm not sure how it would look if split up though, but it sounds like it would mostly just complicates editing them and be easy to make one of them not align with the others etc.

AJenbo:
I had a little chat with chatGPT and we came to the conclusion that we should make a python script that the developers run each time the tsv file is updated. This lets the translation process work with out changes and also addresses my only comment to your code and gives us a simpler data file.

I got excited and made an implementation for it, hope you like it :)

AJenbo (approved):

PR #5223, opened 2022-08-09
Inferno difficulty

kphoenix137:
4th Diablo Difficulty. Increases for monster stats are scaled to be a similar jump from Hell to Inferno as Nightmare to Hell.

### DIFF
--- a/Source/DiabloUI/selgame.cpp
+++ b/Source/DiabloUI/selgame.cpp
@@ -237,6 +237,9 @@ void selgame_GameSelection_Focus(int value)
 			case DIFF_HELL:
 				difficulty = _("Hell");
 				break;
+			case DIFF_INFERNO:
+				difficulty = _("Inferno");
+				break;
 			}
 			infoString.append(fmt::format(fmt::runtime(_(/* TRANSLATORS: {:s} means: Game Difficulty. */ "Difficulty: {:s}")), difficulty));
 			infoString += '\n';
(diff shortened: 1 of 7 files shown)


ChaosMarc:
How is this even remotely in the scope of DevilutionX? Imho this is clearly in the realm of mods

julealgon:
@ChaosMarc having a draft PR in place doesn't necessarily mean it will be merged.

glebm:
This PR is a useful evidence for consolidating all the difficulty adjustments in `Source/difficulty.{hpp,cpp}`, to make modding easier.

ChaosMarc:
ok, from a pure documentational point this PR it may be helpful but I still don't see the need for it. What will happen when all nessessary changes and reviews are done? will it just be closed and thats that? My understanding was that devilutionx's focus was purely on fixing bugs and providing a base for ports and mods but not providing blueprints for specific changes in mods. I absolutely don't want to be rude or something, but in my personal feeling there were some similar tickets in the past that got closed faster that you could look with a quick remark like "out of scope". Another point against this ticket is that it binds resources of the core team for reviews and such. even if the code never will be merged.

qndel:
not really, nobody reviews stuff in draft mode

ChaosMarc:
Ok never mind 👍

AJenbo:
> will it just be closed and thats that?

probably

> there were some similar tickets in the past that got closed faster

It depends on both what and who. We have let PRs stays for a while just as a way to highlight what turned in to someones custom version as well.

kphoenix137:
I opened this PR for 2 reasons:
In case the idea may eventually be accepted if it adheres well enough to the game without affecting game balance (perhaps it could be implemented as a "challenge mode" of sorts for endgame players, where we can mirror the exp gains and item/gold drops of hell difficulty, so it's of no actual benefit to the player, and only provides an extreme challenge for people bored with Hell difficulty)

And

Because submitting it as a PR was the quickest way I could get it to build on github for others to playtest and see what they thought of it.

PR #5228, opened 2022-08-10
CMake: An option to disable demomode support

glebm:
Useful for targets with limited RAM.
Reduces RG99 binary size by 8 KiB.


test/CMakeLists.txt
```diff
@@ -35,13 +35,16 @@ set(tests
...
   quests_test
   random_test
-  timedemo_test
   scrollrt_test
   stores_test
   utf8_test
   writehero_test
 )
 
+if(NOT DISABLE_DEMOMODE)
```
AJenbo, line 44: Could we make it incompatible with building tests instead?

obligaron (approved):
lgtm

One general question:
Should we change these kind of options from `DISABLE_XYZ` to something like `XYZ_ENABLED`. Positive conditions are often easier to read.

`#ifndef DISABLE_DEMOMODE`
=>
`#ifdef DEMOMODE_ENABLED`

test/CMakeLists.txt
glebm: Done, now setting `DISABLE_DEMOMODE` disables `BUILD_TESTING`.

glebm:
The way they're currently named is based on the default.
If a feature is enabled by default, then the option name is `DISABLE_<FEATURE>`.

We can change that if we think always using `ENABLE` is better but that should be in a separate PR.

AJenbo (approved):

PR #5229, opened 2022-08-10
Item effect cleanup

vlolteanu:


DakkJaniels:
I don't think effects that are implemented in the code should be eliminated, even if they aren't being used. Just making it more difficult for modders if they want to use them in the future, because then they would have reimplement all of that.

vlolteanu:
I don't feel strongly either way about keeping unused stuff for modders, but the precedent has been set (probably earlier than what I've linked here): https://github.com/diasurgical/devilutionX/pull/5001#discussion_r922273238

DakkJaniels:
Fair enough - I'm using `infravision` and `nohealmon` currently, so it's just a pain whenever I decide to update to the next release.

Source/itemdat.h
```diff
@@ -325,19 +325,19 @@ enum unique_base_item : int8_t {
 enum class ItemSpecialEffect {
 	// clang-format off
 	None                   = 0,
-	Infravision            = 1 << 0,
+	Unused0                = 1 << 0,
```
AJenbo, line 328: why not just remove them, they have values so won't change things.

Source/itemdat.h
```diff
@@ -507,14 +507,14 @@ enum item_effect_type : int8_t {
...
-	IPL_0x30,      // Unknown
-	IPL_0x31,      // Unknown
-	IPL_FIREBALL,  /* only used in hellfire */
-	IPL_0x33,      // Unknown
+	IPL_0x2F,     // unused
+	IPL_0x30,     // Unknown
+	IPL_0x31,     // Unknown
+	IPL_FIREBALL, /* only used in hellfire */
+	IPL_0x33,     // Unknown
 	IPL_ABSHALFTRAP,
```
AJenbo, line 515: Alternatively we could give a fixed number to the next item in the list and then remove the unused ones.

Source/itemdat.h
vlolteanu: There were already some `Unknown*` values that also weren't used, so I opted to be consistent. I can delete all of them.

AJenbo:
Also a surprising number of times removing some dead code removes a special case that makes it possible to clean up other parts of the code.

AJenbo (approved):

PR #5234, opened 2022-08-12
#3380 Use appropriate terms for mouse/gamepad/touch

burningserenity:
Show correct buttons for gamepads on item help; change the text for virtual gamepads ever so slightly; refactor logic with choosing how to display misc item text.

Currently, w/ a Dual Shock type controller, the words "triangle" and "square" are written out to represent the buttons, rather than using triangle and square characters to represent them. Using the actual characters causes the game to segfault, leaving this message:
```
ERROR: Missing font kerning: fonts\12-25.bin
ERROR: Missing file: fonts\12-25.pcx
ERROR: Error loading font: fonts\12-25.clx
[1]    16390 segmentation fault  ./build/devilutionx --verbose
```

Looking into this issue.

Issue #3380, opened 2021-11-03
Use appropriate terms for mouse/gamepad/touch


AJenbo:
Currently the help texts will always instruct users how to do things with mouse+keyboard, we should make the text dynamically fit the current input device.

`Right-click to read` -> `Press Y to read`

Additionally we should try and detect the controller layout to show △/Y/X for Sony/Microsoft/Nintendo layouts on gamepads.

I'm not sure how best to handle the texts for touch, maybe "Activate to read", or simply get rid of the info panel :eyes:

glebm:
The info box is done

There is no way to toggle the help screen from a gamepad, so I guess it's fine that it's not adapted?

daedrixxx2:
I could add that, in the given example above "press Y to read", the button doesn't currently change in the case the layout has been remapped (for instance I remapped Y to use magic, but the "read button" remained Y.

Also in the newer test builds, and in the incoming v1.5.0 I presume, the help screen can be mapped on a gamepad, so it makes sense that it (like any other on screen info like "press I for the Inventory") should also be adapted to whatever controller the game detects, right?

AJenbo:
it would be good to have a help screen for gamepad users, but we can take that as a separate task

PR #5235, opened 2022-08-12
 #3380 Use appropriate terms for mouse/gamepad/touch #5234 

burningserenity:
Adds support for controller buttons for gamepads. Shows XYBA for most, Square, Triangle, X, Circle for dual shock, and some text for virtual gamepads.

Currently dual shock buttons are represented with "Square", "Triangle" written out, as using symbols causes the game to segfault with this message:
```
ERROR: Missing font kerning: fonts\12-25.bin
ERROR: Missing file: fonts\12-25.pcx
ERROR: Error loading font: fonts\12-25.clx
[1]    16390 segmentation fault  ./build/devilutionx --verbose
```

Looking into this

### DIFF
--- a/Source/controls/devices/game_controller.cpp
+++ b/Source/controls/devices/game_controller.cpp
@@ -232,4 +232,34 @@ bool GameController::IsPressedOnAnyController(ControllerButton button, SDL_Joyst
 	return false;
 }
 
+GamepadLayout GameController::getLayout(const SDL_Event &event)
+{
+#if SDL_VERSION_ATLEAST(2, 0, 10)
+	const int index = event.cdevice.which;
+	const SDL_GameControllerType gamepadType = SDL_GameControllerTypeForIndex(index);
+	switch (gamepadType) {
+#if SDL_VERSION_ATLEAST(2, 0, 16)
+	case SDL_CONTROLLER_TYPE_GOOGLE_STADIA:
+	case SDL_CONTROLLER_TYPE_AMAZON_LUNA:;
+#endif
+	case SDL_CONTROLLER_TYPE_XBOXONE:
+	case SDL_CONTROLLER_TYPE_XBOX360:
+		return GamepadLayout::Xbox;
+#if SDL_VERSION_ATLEAST(2, 0, 14)
+	case SDL_CONTROLLER_TYPE_VIRTUAL:
+		return GamepadLayout::Virtual;
+	case SDL_CONTROLLER_TYPE_PS5:
+#endif
+	case SDL_CONTROLLER_TYPE_PS4:
+	case SDL_CONTROLLER_TYPE_PS3:
+		return GamepadLayout::Playstation;
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
+		return GamepadLayout::Nintendo;
+	case SDL_CONTROLLER_TYPE_UNKNOWN:
+		return GamepadLayout::Generic;
+#endif
+	}
+	return GamepadLayout::Generic;
+}
+
 } // namespace devilution
--- a/Source/controls/devices/game_controller.cpp
+++ b/Source/controls/devices/game_controller.cpp
@@ -239,11 +239,6 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 	const SDL_GameControllerType gamepadType = SDL_GameControllerTypeForIndex(index);
 	switch (gamepadType) {
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
-#if SDL_VERSION_ATLEAST(2, 0, 24)
-	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT:
-	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGHT:
-	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_PAIR:
-#endif
 		return GamepadLayout::Nintendo;
 	case SDL_CONTROLLER_TYPE_PS3:
 	case SDL_CONTROLLER_TYPE_PS4:
@@ -256,9 +251,6 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 #if SDL_VERSION_ATLEAST(2, 0, 16)
 	case SDL_CONTROLLER_TYPE_GOOGLE_STADIA:
 	case SDL_CONTROLLER_TYPE_AMAZON_LUNA:
-#endif
-#if SDL_VERSION_ATLEAST(2, 0, 24)
-	case SDL_CONTROLLER_TYPE_NVIDIA_SHIELD:
 #endif
 		return GamepadLayout::Xbox;
 #if SDL_VERSION_ATLEAST(2, 0, 14)
--- a/Source/controls/devices/game_controller.cpp
+++ b/Source/controls/devices/game_controller.cpp
@@ -239,7 +239,7 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 	const SDL_GameControllerType gamepadType = SDL_GameControllerTypeForIndex(index);
 	switch (gamepadType) {
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
-#if SDL_VERSION_ATLEAST(2, 0, 23)
+#if SDL_VERSION_ATLEAST(2, 0, 24)
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT:
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGHT:
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_PAIR:
@@ -257,7 +257,7 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 	case SDL_CONTROLLER_TYPE_GOOGLE_STADIA:
 	case SDL_CONTROLLER_TYPE_AMAZON_LUNA:
 #endif
-#if SDL_VERSION_ATLEAST(2, 0, 23)
+#if SDL_VERSION_ATLEAST(2, 0, 24)
 	case SDL_CONTROLLER_TYPE_NVIDIA_SHIELD:
 #endif
 		return GamepadLayout::Xbox;
--- a/Source/controls/devices/game_controller.cpp
+++ b/Source/controls/devices/game_controller.cpp
@@ -239,6 +239,7 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 	const SDL_GameControllerType gamepadType = SDL_GameControllerTypeForIndex(index);
 	switch (gamepadType) {
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
+#endif
 #if SDL_VERSION_ATLEAST(2, 0, 23)
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT:
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGH:
--- a/Source/controls/devices/game_controller.cpp
+++ b/Source/controls/devices/game_controller.cpp
@@ -238,23 +238,32 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 	const int index = event.cdevice.which;
 	const SDL_GameControllerType gamepadType = SDL_GameControllerTypeForIndex(index);
 	switch (gamepadType) {
-#if SDL_VERSION_ATLEAST(2, 0, 16)
-	case SDL_CONTROLLER_TYPE_GOOGLE_STADIA:
-	case SDL_CONTROLLER_TYPE_AMAZON_LUNA:;
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
+#if SDL_VERSION_ATLEAST(2, 0, 23)
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT:
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGH:
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_PAIR:
 #endif
+		return GamepadLayout::Nintendo;
+	case SDL_CONTROLLER_TYPE_PS3:
+	case SDL_CONTROLLER_TYPE_PS4:
+#if SDL_VERSION_ATLEAST(2, 0, 14)
+	case SDL_CONTROLLER_TYPE_PS5:
+#endif
+		return GamepadLayout::PlayStation;
 	case SDL_CONTROLLER_TYPE_XBOXONE:
 	case SDL_CONTROLLER_TYPE_XBOX360:
+#if SDL_VERSION_ATLEAST(2, 0, 16)
+	case SDL_CONTROLLER_TYPE_GOOGLE_STADIA:
+	case SDL_CONTROLLER_TYPE_AMAZON_LUNA:
+#endif
+#if SDL_VERSION_ATLEAST(2, 0, 23)
+	case SDL_CONTROLLER_TYPE_NVIDIA_SHIELD:
+#endif
 		return GamepadLayout::Xbox;
 #if SDL_VERSION_ATLEAST(2, 0, 14)
 	case SDL_CONTROLLER_TYPE_VIRTUAL:
-		return GamepadLayout::Virtual;
-	case SDL_CONTROLLER_TYPE_PS5:
 #endif
-	case SDL_CONTROLLER_TYPE_PS4:
-	case SDL_CONTROLLER_TYPE_PS3:
-		return GamepadLayout::Playstation;
-	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
-		return GamepadLayout::Nintendo;
 	case SDL_CONTROLLER_TYPE_UNKNOWN:
 		return GamepadLayout::Generic;
 	}
--- a/Source/controls/devices/game_controller.cpp
+++ b/Source/controls/devices/game_controller.cpp
@@ -257,8 +257,8 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 		return GamepadLayout::Nintendo;
 	case SDL_CONTROLLER_TYPE_UNKNOWN:
 		return GamepadLayout::Generic;
-#endif
 	}
+#endif
 	return GamepadLayout::Generic;
 }
 
--- a/Source/controls/devices/game_controller.cpp
+++ b/Source/controls/devices/game_controller.cpp
@@ -239,10 +239,9 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 	const SDL_GameControllerType gamepadType = SDL_GameControllerTypeForIndex(index);
 	switch (gamepadType) {
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
-#endif
 #if SDL_VERSION_ATLEAST(2, 0, 23)
 	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT:
-	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGH:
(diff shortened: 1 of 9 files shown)


AJenbo:
@burningserenity another option is to provide custom controller symbols in the private use range: U+E000..U+F8FF

This would allow you to render them in any given style unrelated to the generic Geometric Shapes. For example:
https://esquiredevmark.itch.io/generic-gamepad-font

AJenbo:
Also make sure to read: https://github.com/diasurgical/devilutionX/pull/5234#issuecomment-1212616596

AJenbo:
Here is a pixel art ui converted to Diablo's palette:
image
https://greatdocbrown.itch.io/gamepad-ui

AJenbo:
Let us know if you need help solving the build errors

burningserenity:
Okay, SDL doesn't have these variables in all the versions being built.The omission of better SDL1 handling is an obvious mistake, what's less obvious is how to best target SDL2. Is there a version of SDL2 that you reference for Devilutionx? Am I good if I stick with what ships with the latest version of Ubuntu?

AJenbo:
I'm pretty sure we support SDL2 all the way back to 2.0.8. We do implement some graceful fallback via `SDL_VERSION_ATLEAST` and `sdl2_to_1_2_backports.h`.

Source/controls/plrctrls.h
```diff
@@ -5,6 +5,7 @@
 #include <cstdint>
 
 #include <SDL.h>
+#include <SDL_gamecontroller.h>
```
AJenbo, line 8: SDL1 only has joystick support meaning which requires a bit more management of the individual elements that makes up the gamepad. All of that has been abstracted away in general though to if you implemented things in the right layer you shouldn't have issues with that.

```suggestion
#ifndef USE_SDL1
#include "controls/devices/game_controller.h"
#endif
```

AJenbo:
SDL2.0.12
SDL_GameControllerType

SDL2.0.14
SDL_CONTROLLER_TYPE_VIRTUAL
SDL_CONTROLLER_TYPE_PS5

SDL2.0.16
SDL_CONTROLLER_TYPE_AMAZON_LUNA
SDL_CONTROLLER_TYPE_GOOGLE_STADIA

SDL2.0.24 (future release)
SDL_CONTROLLER_TYPE_NVIDIA_SHIELD
SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT
SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGH
SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_PAIR

Source/items.cpp
```diff
@@ -1741,76 +1741,99 @@ void DrawUniqueInfoWindow(const Surface &out)
...
+	    || item._iSpell == SPL_TOWN
+	    || item._iSpell == SPL_IDENTIFY
+	    || TargetsMonster(item._iSpell);
+
+	if (ControlMode == ControlTypes::KeyboardAndMouse) {
+		printItemMiscKBM(item, isOil, isCastOnTarget);
+	} else if (ControlMode == ControlTypes::VirtualGamepad || GamepadType == SDL_CONTROLLER_TYPE_VIRTUAL) {
+		printItemMiscVirtualGamepad(item, isOil);
+	} else {
+		const bool usingDualShock = IsAnyOf(GamepadType, SDL_CONTROLLER_TYPE_PS3, SDL_CONTROLLER_TYPE_PS4, SDL_CONTROLLER_TYPE_PS5);
```
AJenbo, line 1835: Does a bool accurately represent the various types of layouts?

Is the brand name for a line of controllers introduced between 1997 and 2014 the best way to express a layout that was introduced by the original PlayStation in 1994 and still in use (and by other products)?

AJenbo:
I think the best solution here is to create a general function in `Source/controls/controller.cpp` which returns our own enum (create one) that just has the types that you have implemented support for. You can maybe base it off of `ToControllerButtonEvent()`. For KBCTRL and Joystick this should just return Generic as we do not have a way to detect them. Platforms that utilize SDL1 would then overwrite this at compile time (Nintendo 3DS).

For SDL older then 2.0.12, don't to any logic, simply return `generic`. And increment support from there using.
At a later point we could add additional logic for older SDL versions and more obscure controlers via `SDL_JoystickName()` and `SDL_GameControllerName()` if we feel it would be relevant.

Source/items.cpp
AJenbo: Did a bit of research, and there are 3 main types:

- Nintendo (SNES 1990)
- PlayStation (1994)
- Xbox (2001)

The Nintendo layout goes back to NES 1985, but since it only had 8 of the 12 buttons we need it's not really relevant. Also they experimented a lot with the controller layout between 1995 and 2006, but we do not really support any of these yet.

Here are the enums that the SDL API provides grouped in to the ones that they belong to.

```
Nintendo:
    SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT,
    SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGHT,
    SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_PAIR,
    SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO,

PlayStation:
    SDL_CONTROLLER_TYPE_PS3,
    SDL_CONTROLLER_TYPE_PS4,
    SDL_CONTROLLER_TYPE_PS5,

Xbox:
    SDL_CONTROLLER_TYPE_GOOGLE_STADIA,
    SDL_CONTROLLER_TYPE_AMAZON_LUNA,
    SDL_CONTROLLER_TYPE_NVIDIA_SHIELD,
    SDL_CONTROLLER_TYPE_XBOX360,
    SDL_CONTROLLER_TYPE_XBOXONE,

Generic:
    SDL_CONTROLLER_TYPE_UNKNOWN = 0,
    SDL_CONTROLLER_TYPE_VIRTUAL,
```

Source/controls/devices/game_controller.cpp
```diff
@@ -232,4 +232,34 @@ bool GameController::IsPressedOnAnyController(ControllerButton button, SDL_Joyst
...
+#endif
+	case SDL_CONTROLLER_TYPE_PS4:
+	case SDL_CONTROLLER_TYPE_PS3:
+		return GamepadLayout::Playstation;
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
+		return GamepadLayout::Nintendo;
+	case SDL_CONTROLLER_TYPE_UNKNOWN:
+		return GamepadLayout::Generic;
+#endif
+	}
```
AJenbo, line 261:
```suggestion
	}
#endif
```

Source/controls/devices/game_controller.h
```diff
@@ -8,6 +8,14 @@
 
 namespace devilution {
 
+enum GamepadLayout {
```
AJenbo, line 11: Lets start new enums as scoped so that there is strict checking on them :)
```suggestion
enum class GamepadLayout {
```

Source/controls/devices/game_controller.h
```diff
@@ -8,6 +8,14 @@
 
 namespace devilution {
 
+enum GamepadLayout {
+	Generic = 0,
+	Xbox,
+	Playstation,
+	Nintendo,
+	Virtual
```
AJenbo, line 16: Nit: Nether alphabetical, or historic order, feels like some sort of favoritism :D
```suggestion
	Generic = 0,
	Nintendo,
	Playstation,
	Xbox,
	Virtual
```

Source/controls/plrctrls.cpp
```diff
@@ -1495,32 +1546,51 @@ string_view ControlTypeToString(ControlTypes controlType)
...
+	case GamepadLayout::Generic:
+		return "Unknown";
+	case GamepadLayout::Nintendo:
+		return "Nintendo Switch Pro";
+	case GamepadLayout::Playstation:
+		return "Dual Shock";
+	case GamepadLayout::Virtual:
+		return "Virtual Gamepad";
+	case GamepadLayout::Xbox:
+		return "Xbox";
```
AJenbo, line 1561: Nit: Just for constancy
```suggestion
	case GamepadLayout::Xbox:
		return "Xbox";
	case GamepadLayout::Virtual:
		return "Virtual Gamepad";
```

Source/controls/plrctrls.cpp
```diff
@@ -11,10 +11,7 @@
 #include "automap.h"
 #include "control.h"
 #include "controls/controller_motion.h"
-#include "miniwin/misc_msg.h"
-#ifndef USE_SDL1
 #include "controls/devices/game_controller.h"
```
AJenbo, line 14: Removing this check is probably going to break SDL1
```suggestion
#ifndef USE_SDL1
#include "controls/devices/game_controller.h"
#endif
```

Source/controls/plrctrls.cpp
```diff
@@ -1409,37 +1408,75 @@ bool IsStickMovementSignificant()
...
-		return event.button.which == SDL_TOUCH_MOUSEID ? ControlTypes::VirtualGamepad : ControlTypes::KeyboardAndMouse;
-	if (event.type == SDL_MOUSEMOTION)
-		return event.motion.which == SDL_TOUCH_MOUSEID ? ControlTypes::VirtualGamepad : ControlTypes::KeyboardAndMouse;
-	if (event.type == SDL_MOUSEWHEEL)
-		return event.wheel.which == SDL_TOUCH_MOUSEID ? ControlTypes::VirtualGamepad : ControlTypes::KeyboardAndMouse;
-	if (IsAnyOf(event.type, SDL_FINGERDOWN, SDL_FINGERUP, SDL_FINGERMOTION))
-		return ControlTypes::VirtualGamepad;
+	GamepadLayout gamepad = GameController::getLayout(event);
+
+	if (gamepad > GamepadLayout::Generic) {
```
AJenbo, line 1423: As a human this doesn't make logical sens to me, what is better then a generic layout?

Maybe this what you intended to express:
```suggestion
	if (gamepad != GamepadLayout::Generic) {
```

Source/controls/plrctrls.cpp
```diff
@@ -1409,37 +1408,75 @@ bool IsStickMovementSignificant()
...
+		return {
+			.controlTypeCoarse = gamepad < GamepadLayout::Virtual ? ControlTypes::Gamepad : ControlTypes::VirtualGamepad,
```
AJenbo, line 1425: By the way, treating an enum as if it was an integer and assuming knolage of the underlying type will also fail when using scoped enums.
```suggestion
			.controlTypeCoarse = gamepad != GamepadLayout::Virtual ? ControlTypes::Gamepad : ControlTypes::VirtualGamepad,
```

If you mean to apply this to more then 1 option then use `IsAnyOf(gamepad, A, B, C)` or `IsNoneOf(gamepad, A, B, C)`

Source/controls/plrctrls.h
```diff
@@ -4,6 +4,7 @@
 #include <cstddef>
 #include <cstdint>
 
+#include "controls/devices/game_controller.h"
```
AJenbo, line 7: this is an SDL2 only feature, apply guard

Source/controls/devices/game_controller.h
AJenbo, line 11: This enum should live in the device agnostic layer, like `controls/game_controls.h`, as it is used in general and not specifically for SDL2 game controllers.

Source/controls/plrctrls.cpp
```diff
@@ -1409,37 +1408,75 @@ bool IsStickMovementSignificant()
 	    || rightStickX != 0 || rightStickY != 0;
 }
 
-ControlTypes GetInputTypeFromEvent(const SDL_Event &event)
+ControlTypeWithGamepad GetInputTypeFromEvent(const SDL_Event &event)
```
AJenbo, line 1411: Adding Gamepad layout to this function both feels a bit wrong since not every control type is a gamepad, but also it makes the function a lot more complicated (a lot of it being repetitive as well).

Maybe instead it should just be it own function that gets called separately.

Source/items.cpp
```diff
@@ -1741,76 +1741,99 @@ void DrawUniqueInfoWindow(const Surface &out)
...
 	}
-	if (IsAnyOf(item._iMiscId, IMISC_BOOK, IMISC_NOTE)) {
-		if (ControlMode == ControlTypes::KeyboardAndMouse) {
-			AddPanelString(_("Right-click to read"));
-		} else {
-			AddPanelString(_("Activate to read"));
-		}
+}
+
+void printItemMiscGamepad(const Item &item, const bool isOil, const bool isCastOnTarget, const bool usingDualShock)
```
AJenbo, line 1780: Left over param, and we don't `const` copied values.
```suggestion
void printItemMiscGamepad(const Item &item, bool isOil, bool isCastOnTarget)
```

Source/controls/devices/game_controller.h
burningserenity: Haha definitely not favoritism, or Nintendo would be first :-D

Source/controls/plrctrls.cpp
burningserenity: I thought I needed game_controller.h regardless of SDL version

Source/controls/plrctrls.h
burningserenity: Okay I get it. Ifndef.

Source/controls/plrctrls.cpp
```diff
@@ -1478,6 +1479,20 @@ void LogControlDeviceAndModeChange(ControlTypes newControlDevice, ControlTypes n
 	LogVerbose("Control: device {}, mode {}", DebugChange(ControlDevice, newControlDevice), DebugChange(ControlMode, newControlMode));
 }
 
+void LogGamepadChange(GamepadLayout newGamepad)
+{
+	if (SDL_LOG_PRIORITY_VERBOSE < SDL_LogGetPriority(SDL_LOG_CATEGORY_APPLICATION))
+		return;
+	if (GamepadType == newGamepad)
+		return;
```
AJenbo, line 1487: call already checks for this.

Source/controls/devices/game_controller.cpp
```diff
@@ -232,4 +232,34 @@ bool GameController::IsPressedOnAnyController(ControllerButton button, SDL_Joyst
 	return false;
 }
 
+GamepadLayout GameController::getLayout(const SDL_Event &event)
+{
+#if SDL_VERSION_ATLEAST(2, 0, 10)
```
AJenbo, line 237:
```suggestion
#if SDL_VERSION_ATLEAST(2, 0, 12)
```

Source/controls/plrctrls.h
AJenbo: Still missing

Source/items.cpp
```diff
@@ -1741,76 +1741,103 @@ void DrawUniqueInfoWindow(const Surface &out)
...
-		if (ControlMode == ControlTypes::KeyboardAndMouse) {
-			AddPanelString(_("Right-click to view"));
+		AddPanelString(_(y_icon + " to view"));
+	} else if (isOil) {
+		PrintItemOil(item._iMiscId);
+		if (!invflag) {
+			AddPanelString(_("Open inventory to use"));
 		} else {
-			AddPanelString(_("Activate to view"));
+			AddPanelString(_(y_icon + " to use"));
```
AJenbo, line 1804: `_()` is a translation function, you need to feed it string that can be picked up directly from the source or it will not match any string in the translation file. So it has to be either the ful finalized string, or a paramters inserted using `fmt::format()`.

Source/items.cpp
```diff
@@ -1741,76 +1741,103 @@ void DrawUniqueInfoWindow(const Surface &out)
...
-		} else {
-			AddPanelString(_("Activate to read"));
-		}
+}
+
+#ifndef USE_SDL1
+void printItemMiscGamepad(const Item &item, bool isOil, bool isCastOnTarget)
+{
+	std::string y_icon = "Y";
+	std::string x_icon = "X";
```
AJenbo, line 1784: Use generic terms like `Activate` and `Cast` for the fallback, X/Y can be meaningless or missleading.

```suggestion
	std::string activateButton = "Activate";
	std::string castButton = "Cast";
```

Source/items.cpp
```diff
@@ -1741,76 +1741,103 @@ void DrawUniqueInfoWindow(const Surface &out)
...
+	    || TargetsMonster(item._iSpell);
+
+	if (ControlMode == ControlTypes::KeyboardAndMouse) {
+		printItemMiscKBM(item, isOil, isCastOnTarget);
+	} else if (ControlMode == ControlTypes::VirtualGamepad) {
+		printItemMiscVirtualGamepad(item, isOil);
+#ifndef USE_SDL1
+	} else {
+		printItemMiscGamepad(item, isOil, isCastOnTarget);
+#endif
```
AJenbo, line 1840: This would have left SDL1 based gamepads with out any info

AJenbo: (fixed in my commit)

Source/controls/devices/game_controller.cpp
```diff
@@ -238,23 +238,32 @@ GamepadLayout GameController::getLayout(const SDL_Event &event)
 	const int index = event.cdevice.which;
 	const SDL_GameControllerType gamepadType = SDL_GameControllerTypeForIndex(index);
 	switch (gamepadType) {
-#if SDL_VERSION_ATLEAST(2, 0, 16)
-	case SDL_CONTROLLER_TYPE_GOOGLE_STADIA:
-	case SDL_CONTROLLER_TYPE_AMAZON_LUNA:;
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_PRO:
+#if SDL_VERSION_ATLEAST(2, 0, 23)
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_LEFT:
+	case SDL_CONTROLLER_TYPE_NINTENDO_SWITCH_JOYCON_RIGH:
```
burningserenity, line 244: Is it a good idea to support individual joycons? We lose two trigger buttons and a start/select button with just one joycon. Is this adequate for playing Diablo?

AJenbo: I'm not entirely sure how it works, my thinking was that it would still behave as the full gamepad. But maybe that is incorrect? We sort of would need someone to test this out. Feel free to ask people on Discord etc.

Source/utils/pcx.cpp
```diff
@@ -1,5 +1,6 @@
 #include "utils/pcx.hpp"
 
+#include <SDL_endian.h>
```
AJenbo, line 3: Looks like you cherry-picked, or rebased master on top of your branch, rather then the other way around?

Source/items.cpp
```diff
@@ -1741,76 +1741,98 @@ void DrawUniqueInfoWindow(const Surface &out)
...
+		if (!invflag) {
+			AddPanelString(_("Open inventory to use"));
 		} else {
-			AddPanelString(_("Activate to view"));
+			AddPanelString(_(fmt::format("{} to use", activateButton)));
 		}
+	} else if (isCastOnTarget) {
+		AddPanelString(_(fmt::format("Select from spell book, then {} to read", castButton)));
+	} else if (IsAnyOf(item._iMiscId, IMISC_BOOK, IMISC_NOTE, IMISC_SCROLL)) {
+		AddPanelString(_(fmt::format("{} to read", activateButton)));
```
AJenbo, line 1808: You translate the string litteral, not the resulting string. Else translators would have to predict the future when making translations :)

```suggestion
		AddPanelString(fmt::format(_("{} to read", activateButton)));

```

Source/items.cpp
```diff
@@ -1741,76 +1741,98 @@ void DrawUniqueInfoWindow(const Surface &out)
...
+	if (GamepadType == GamepadLayout::Xbox) {
+		activateButton = "Y";
+		castButton = "X";
+	} else if (GamepadType == GamepadLayout::PlayStation) {
+		activateButton = "Triangle";
+		castButton = "Square";
+	} else if (GamepadType == GamepadLayout::Nintendo) {
+		activateButton = "Y";
+		castButton = "X";
 	}
```
AJenbo, line 1794: If they are the same cases you can just do:
```suggestion
	if (IsAnyOf(GamepadType, GamepadLayout::Xbox, GamepadLayout::Nintendo)) {
		activateButton = "Y";
		castButton = "X";
	} else if (GamepadType == GamepadLayout::PlayStation) {
		activateButton = "Triangle";
		castButton = "Square";
	}
```

burningserenity: I was leaving this as a placeholder for the icons

Source/utils/pcx.cpp
burningserenity: I may have. I'll ping you on Discord.

AJenbo (approved):

PR #5239, opened 2022-08-13
Correct some header includes

AJenbo:


Source/utils/sdl2_to_1_2_backports.cpp
```diff
@@ -1,5 +1,6 @@
 #include "./sdl2_to_1_2_backports.h"
 
+#include <SDL_getenv.h>
```
AJenbo, line 3:
```suggestion
```

Source/multi.cpp
```diff
@@ -7,6 +7,7 @@
 #include <SDL.h>
 #include <config.h>
 
+#include <ctime>
```
AJenbo, line 10: This is causing builds to fail for PS2

glebm (approved):

AJenbo:
Thanks!

PR #5240, opened 2022-08-13
Update SDL2 to versoin 2.23.2

AJenbo:
Required for PS2 support, and probably Android NDK 25 (default version by November)


glebm (approved):
Do we need to update the Android Java code as well?

AJenbo:
Yes, probably a good idea

PR #5242, opened 2022-08-13
Use SDL for creating file during RW test

AJenbo:
It looks like this is mostly a left over from when file handeling was migrated to SDL

### DIFF
--- a/Source/restrict.cpp
+++ b/Source/restrict.cpp
@@ -8,17 +8,19 @@
 #include "utils/file_util.h"
 #include "utils/paths.h"
 
+#include <SDL.h>
+
 namespace devilution {
 
 void ReadOnlyTest()
 {
 	const std::string path = paths::PrefPath() + "Diablo1ReadOnlyTest.foo";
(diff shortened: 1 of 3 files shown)


Source/restrict.cpp
```diff
@@ -8,17 +8,19 @@
...
+
 namespace devilution {
 
 void ReadOnlyTest()
 {
 	const std::string path = paths::PrefPath() + "Diablo1ReadOnlyTest.foo";
-	FILE *f = FOpen(path.c_str(), "w");
-	if (f == nullptr) {
+	SDL_RWops *file = SDL_RWFromFile(path.c_str(), "w");
+	if (file == NULL) {
```
qndel, line 19: why not nullptr?

AJenbo: Cause we didn't do it in other places and I didn't feel like breaking the trend :shrug:

AJenbo: Looked at it a bit more and decided to change the other cases instead.

qndel (approved):

PR #4999, opened 2022-07-14
:recycle: Adds Missile::SourcePlayer and Missile::SourceMonster

canassa:
This is a follow up from the discussions on #4994

Adds the helper methods `SourcePlayer` and `SourceMonster` to the `Missile` struct.

The end goal is to replace every direct access to `_misource` with calls to `SourcePlayer` and/or `SourceMonster`

`AddFlash` already converted to use the new methods.

PR #5243, opened 2022-08-13
Fix mana usage from shrines

DakkJaniels:
Revert changing TARGET_PLAYERS to TARGET_MONSTERS under #4999 (switching it back to TARGET_PLAYERS under this PR). Changing it to TARGET_MONSTERS makes the spell use mana.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -2411,7 +2411,7 @@ void OperateShrineMagical(const Player &player)
 	    player.position.tile,
 	    player._pdir,
 	    MIS_MANASHIELD,
-	    TARGET_PLAYERS,
+	    TARGET_BOTH,
 	    player.getId(),
 	    0,
 	    2 * leveltype);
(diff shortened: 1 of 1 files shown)


DakkJaniels:
Switched to TARGET_BOTH so the spell still won't use mana, and SourceType() and sourcePlayer() will return the player (if that matters for these shrine spells)

StephenCWills:
I checked all the relevant missiles, and none of them depend on `_micaster` so this PR should be okay. However, I feel like it would be better if calls to `ConsumeSpell()` were removed from the missile functions and added to other more appropriate functions, such as `CastSpell()`. I didn't quite exhaustively search the codebase for all usages of `AddMissile()`, but I did find a couple functions that would likely need calls to `ConsumeSpell()`: `AddJester()` and `OperateBook()` (when learning Guardian in the Chamber of Bone).

StephenCWills:
I implemented my suggestion with #5418 which I believe makes this PR obsolete. Thanks for having a look at this.

Issue #5248, opened 2022-08-16
It's possible to make cursor items get overwriten by cursor changing spells


qndel:
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

Cursor items can disappear

### To Reproduce

1. Pick an item from the ground with inventory screen open
2. Right before picking the item cast a spell that changes cursor
3. If you timed it well, you will see the item get picked up then the spell change the cursor and destroy the item

### Expected Behavior

No idea, certainly not the item being eaten :D

### Additional context

Easiest to test with bard (identify skill)

PR #5249, opened 2022-08-16
fix cursor item being overwritten by cursor changing spells

qndel:
Fixes https://github.com/diasurgical/devilutionX/issues/5248


AJenbo (approved):
There are probably a few more things we could clean up to get the cursor to behave in a safe way, but this looks like a step in the right direction.

Issue #5250, opened 2022-08-16
Invisible Enemies' Death Animation is invisible


burningserenity:
### Operating System

Linux x64

### DevilutionX version

1.4.1 (latest release)

### Describe

When a monster that can turn invisible is killed while invisible, their death animation does not appear to play. This results in the sudden appearance of their corpse.

### To Reproduce

1. Find a monster which can turn invisible
2. Get it to turn invisible
3. Fire spells or arrows at it until it dies

### Expected Behavior

Death animation plays and is visible

### Additional context

https://user-images.githubusercontent.com/25830960/184985054-aea35195-ef72-4acd-938e-a38c957633b5.mp4

AJenbo:
I belive that this is how it was in vanilla?

FitzRoyX:
This could maybe be considered an oversight. It certainly makes more sense to me that it can only have the cloaking effect when it's alive.

burningserenity:
I can't speak to whether or not this behavior exists in vanilla, but I would think it would be hard to sustain something like active cloaking while dying. I guess we could argue about at what point in the dying process cloaking should deactivate, but to me, having it turn off only after the monster is finished dying looks silly.

PR #5253, opened 2022-08-18
break hidden invis on death

qndel:
Resolves https://github.com/diasurgical/devilutionX/issues/5250
Imo looks better than corpses just suddenly appearing after death animation has ended, duh.


AJenbo (commented):
Could we instead remove the flag on death?

qndel:
> Could we instead remove the flag on death?

I tried to find why they appear after animation ends at all but failed xD

AJenbo:
Because they are corpses and not monsters

AJenbo (approved):

PR #5254, opened 2022-08-20
Use Abseil

glebm:
Adds a dependency on https://github.com/abseil/abseil-cpp.

absl is portable, fast to compile, and battle-tested.

It can replace:
1. `string_view` -> `absl::string_view` (done in this PR)
2. `tl::function_ref` -> `absl::FunctionRef` (done in this PR)
3. `StrCat/StrAppend` -> `absl::StrCat/Append` (done in this PR)
4. `std::unordered_map/set` -> `absl::flat_hash_map/set` (faster and smaller, for a future PR)

It also provides `absl::Status/StatusOr` for exception-less error handling.


glebm:
Turns out that the incompatibility between `absl::string_view` and `fmt::string_view` on pre-C++17 compilers makes this somewhat infeasible. Can revisit when/if we switch to C++17.

PR #5256, opened 2022-08-20
Avoid DLevel copy constructor in GetDeltaLevel()

StephenCWills:
Fixes a crash due to data alignment on 3DS when attempting to create a Multiplayer game.

Some relevant notes from @glebm on Discord.

> Sounds like `DLevel` has a broken copy constructor
> may have something to do with `#pragma pack(push, 1)`
> older ARM CPUs crash on unaligned writes (and reads sometimes)


glebm (approved):

AJenbo:
Nice work!

PR #5257, opened 2022-08-20
Lowercase all file paths

glebm:
We want to be able to use unpacked MPQs on low-end platforms (PS2/rg99/etc).

This is tricky on case-sensitive filesystems. Avoids case issues by lowercasing all paths in the code (then we'll just need lowercased listfiles).


AJenbo (approved):
Opening .DUN was always such a struggle

PR #5259, opened 2022-08-21
[3DS] Bundle assets directly in romfs

StephenCWills:
Likely speeds up load times, measurably improves RAM usage, and cleans up the CMake logic quite a bit too. The only downside is larger binaries. They are about 4 MB larger without the MPQ compression.

image


AJenbo:
Forgot to check this before merging: https://github.com/diasurgical/devilutionX/pull/5260

PR #5260, opened 2022-08-21
Do not try to load devilutionx.mpq on 3DS

AJenbo:


StephenCWills (approved):
> In that case i think it would be best to drop the MPQ, it also avoids the upgrade issue for people that have an old one lying on the console.

Ahh, now I see what you meant.

PR #5265, opened 2022-08-21
Remove unnecessary code in InvDrawSlotBack

kphoenix137:
Removes code that does nothing.


kphoenix137:
PR No longer needed since #5266 was merged

PR #5266, opened 2022-08-21
[QOL] Colored slot backs

kphoenix137:
image

Changes the slot back colors of Magical and Unique quality items in Stash and Inventory to visually identify your items without needing to mouse-over.


kphoenix137:
Should I add a QOL toggle for this?

obligaron:
side-note:
#4536 also tries to use slot back colors to differ between own player and inspected player.
So we should think about for what we want to use the colors? 🤔

kphoenix137:
> side-note: #4536 also tries to use slot back colors to differ between own player and inspected player. So we should think about for what we want to use the colors? 🤔

Use same colors, but display other players on the left panel instead of right panel?

Chance4us:
> Should I add a QOL toggle for this?

There should be an option for the users who don't like it.
I don't think it's bad at first glance, even if it gives the game a slightly warmer atmosphere.
But there are players who like it classically sinister dark, and that's exactly why there should be an option.

julealgon:
> There should be an option for the users who don't like it.

There will always be people who don't like something. If we follow this logic, then every single little modification in the game would need to be behind a toggle.

I just don't think this is a good argument.

FitzRoyX:
The blue doesn't look too bad, but I can't bring myself to like the yellow. It's too light of a background or something, the item art blends with it too much and it ends up looking washed out.
Untitled-1dd

julealgon:
> but I can't bring myself to like the yellow. It's too light of a background or something, the item art blends with it too much and it ends up looking washed out.

I think it will vary wildly based on the item art itself: if it contains a lot of yellow/gold colors, then it will blend in more. This is to be expected really.

Items such as valor, undead crown, harlequin's, wizard's wrap, etc have a _ton_ of yellow/gold color in their art (some are mostly gold) and those will blend in the most.

I think we'd need a screenshot with more variability in items to judge this better, but even then, it will be somewhat subjective too.

I guess one option to mitigate this problem a bit, would be to, instead of having the whole tile(s) be of the item color, make it so that there is a gradient towards the item color that is only at the max value at the borders. Obviously, that would take a lot more work to implement though.

Alternatively, one could investigate adding a black border to the items and see if that works (kind of like we did for the automap to be able to see it over lava tiles).

FitzRoyX:
> Alternatively, one could investigate adding a black border to the items and see if that works (kind of like we did for the automap to be able to see it over lava tiles).

No, I think yellow is a lost cause. As you said, lots of silver and gold in items. The original red does its job well, which is to clearly show which slots are occupied without blending. Blue is a bit blendy with the dark gray of the slots, but is ok. Since many unique items have unique graphics, there's nothing an extra color does for them anyway.

You'll notice that in D2 they changed the orientation of large items so that the graphics themselves covered up the slots they occupied better. This could've been to make weapons look longer, but it also could've been to make blue less problematic.

kphoenix137:
> > Alternatively, one could investigate adding a black border to the items and see if that works (kind of like we did for the automap to be able to see it over lava tiles).
> 
> No, I think yellow is a lost cause. As you said, lots of silver and gold in items. The original red does its job well, which is to clearly show which slots are occupied without blending. Blue is a bit blendy with the dark gray of the slots, but is ok. Since many unique items have unique graphics, there's nothing an extra color does for them anyway.
> 
> You'll notice that in D2 they changed the orientation of large items so that the graphics themselves covered up the slots they occupied better. This could've been to make weapons look longer, but it also could've been to make blue less problematic.

Not as impactful for unique items, but more for normal items against magical items. During gameplay, you can easily identify then that you've accidentally picked up a normal item and can toss it out, without having to hover over every item in your inventory.

kphoenix137:
> The blue doesn't look too bad, but I can't bring myself to like the yellow. It's too light of a background or something, the item art blends with it too much and it ends up looking washed out. Untitled-1dd

Maybe I can explore shifting each color in the palette by 1 to make it darker, causing fewer blending issues

kphoenix137:
image

AJenbo:
> > The blue doesn't look too bad, but I can't bring myself to like the yellow. It's too light of a background or something, the item art blends with it too much and it ends up looking washed out. Untitled-1dd
> 
> Maybe I can explore shifting each color in the palette by 1 to make it darker, causing fewer blending issues

you make a good argument. I was previously a bit against it as it changes the mood a bit with a large portion of the screen having a different color, but I'm more for it now.

AJenbo (approved):

PR #5267, opened 2022-08-21
Make function to display multiple lines in panels; implement it as POC.

burningserenity:
Adds a function to display multiple lines of panel text, instead of having to call AddPanelString multiple tiimes.


Source/control.cpp
```diff
@@ -466,6 +466,40 @@ void AddPanelString(string_view str)
 		pnumlines++;
 }
 
+std::vector<string_view> splitStringViewByNewline(const string_view str, const char delim = '\n')
```
glebm, line 469: Let's not allocate a vector on every frame.
You can use this implementation instead: https://github.com/OpenDingux/gmenu2x/blob/master/src/split_by_char.h

burningserenity: I don't really understand what this is doing. It seems to me that I want to create a SplitByChar by constructor, get a SplitByCharIterator out of it by calling begin(), then I should be able to use the * operator with... nothing? That doesn't compile. It looks like I can move to the next string with ++, that makes sense, it's just actually pulling string_view objects out of this that doesn't.

AJenbo:
Since the text rendere can render multiple lines, maybe we can feed it just a single text, and just adjust the y offset of the first line and line height depending on the number of lines?

burningserenity:
> Since the text rendere can render multiple lines, maybe we can feed it just a single text, and just adjust the y offset of the first line and line height depending on the number of lines?

Is that under engine/, or somewhere else?

AJenbo:
https://github.com/diasurgical/devilutionX/blob/master/Source/engine/render/text_render.hpp#L187

Source/control.cpp
glebm:
```c++
for (string_view s : SplitByChar(str, '\n')) {
}
```

burningserenity: awesome

Source/utils/stdcompat/string_view_split.hpp
```diff
@@ -0,0 +1,92 @@
+#include "utils/stdcompat/string_view.hpp"
```
glebm, line 1: This is not a polyfill for a standard library functions, so it shouldn't be in stdcompat.
Please move to `util/string_split_by_char.hpp`

burningserenity:
> https://github.com/diasurgical/devilutionX/blob/master/Source/engine/render/text_render.hpp#L187

I played with breaking up text by newlines in the DrawString function, but the problem with this approach is that I need any panels drawn before and after a string with multiple lines to also have their y offset changed, but I don't have access to text from previous panels in DrawString. Am I missing something?

AJenbo:
Don't support a mix of multi line an single line with line breaks. Ideally we would only support a single entry with line breaks.

burningserenity:
So we would have a single entry comprised of all the text from various sources that would make up that info panel, feed them into DrawString, and break them all up in there?

AJenbo:
Might not even have to break them up, just count the line breaks, and adjust the line height and y offset before drawing the text.

burningserenity:
Okay so I have to figure out how to get all of the panel strings to be passed to DrawString. How does DrawString work in relation to AddPanelString?

AJenbo:
Well no, you just need get `PrintInfo()` to count the line breaks in `InfoString()`, instead of using `pnumlines`.

At that point (possibly as a follow up PR) multiple calls to `AddPanelString()` can be replaced by setting/concat `InfoString`.

```cpp
InfoString = string_view(item._iIName);
AddPanelString(_("Right-click to read, then"));
AddPanelString(_("left-click to target"));
```

```cpp
InfoString = StrCat(item._iIName, "\n", _("Right-click to read, then\nleft-click to target"));
```

burningserenity:
Okay that was a pretty simple change. The more complicated part is how to calculate the y offset. How does the y offset get calculated with AddPanelString?

AJenbo:
`LineStart[pnumlines]`, you already changed this meaning that it will be incorrect when using `AddPanelString()`:

https://github.com/diasurgical/devilutionX/pull/5267/files#diff-d59ae24c872d55375ef9c864346e2841b9075ceca2b2d24f95836c67a31669f8R270

burningserenity:
I see that scrolls with stat requirements display differently, due to the AddPanelString.

AJenbo:
A picture would be helpful

burningserenity:
Screenshot_2022-09-04_15:20:11
Screenshot_2022-09-04_15:20:22
Screenshot_2022-09-04_15:20:31
Screenshot_2022-09-04_15:20:37
Screenshot_2022-09-04_15:20:39


That's with the current value, `LineStart[infoStringLines + pnumlines`

burningserenity:
Think I finally grokked it:

Screenshot_2022-09-04_16:32:53
Screenshot_2022-09-04_16:32:55
Screenshot_2022-09-04_16:32:57
Screenshot_2022-09-04_16:32:58
Screenshot_2022-09-04_16:32:59
Screenshot_2022-09-04_16:33:05
Screenshot_2022-09-04_16:33:06

Source/items.cpp
```diff
@@ -1788,18 +1786,18 @@ void printItemMiscGamepad(const Item &item, bool isOil, bool isCastOnTarget)
 	}
 
 	if (item._iMiscId == IMISC_MAPOFDOOM) {
-		AddPanelString(fmt::format(fmt::runtime(_("{} to view")), activateButton));
+		InfoString = StrCat(item._iName, "\n", fmt::format(fmt::runtime(_("{} to view")), activateButton));
 	} else if (isOil) {
 		PrintItemOil(item._iMiscId);
 		if (!invflag) {
-			AddPanelString(_("Open inventory to use"));
+			InfoString = StrCat(item._iName, "\n", "Open inventory to use");
```
qndel, line 1793: u disabled translating that line

Source/engine/render/text_render.cpp
```diff
@@ -644,7 +644,7 @@ uint32_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
 
 	characterPosition.y += BaseLineOffset[size];
 
-	const int bytesDrawn = DoDrawString(out, text, rect, characterPosition, spacing, lineHeight, lineWidth, rightMargin, bottomMargin, flags, size, color);
+	int bytesDrawn = DoDrawString(out, text, rect, characterPosition, spacing, lineHeight, lineWidth, rightMargin, bottomMargin, flags, size, color);
```
AJenbo, line 647:
```suggestion
	const int bytesDrawn = DoDrawString(out, text, rect, characterPosition, spacing, lineHeight, lineWidth, rightMargin, bottomMargin, flags, size, color);
```

Source/control.cpp
```diff
@@ -469,6 +475,40 @@ void AddPanelString(string_view str)
...
+{
+	std::vector<string_view> panelLines = splitStringViewByNewline(str);
+	for (uint32_t i = 0; i < size(panelLines); i++) {
+		CopyUtf8(panelstr[pnumlines], panelLines[i], sizeof(*panelstr));
+
+		if (pnumlines < 4)
+			pnumlines++;
+	}
+}
+
```
AJenbo, line 511: Looks unused

AJenbo, line 511: Looks unused

Source/control.cpp
```diff
@@ -264,17 +265,22 @@ void PrintInfo(const Surface &out)
...
+	Rectangle line { GetMainPanel().position + Displacement { 177, LineStart[infoStringLines + pnumlines] }, { 288, 12 } };
 
 	if (!InfoString.empty()) {
-		DrawString(out, InfoString, line, InfoColor | UiFlags::AlignCenter | UiFlags::KerningFitSpacing, 2);
-		line.position.y += LineHeights[pnumlines];
+		for (string_view s : SplitByChar(InfoString, '\n')) {
+			DrawString(out, s, line, InfoColor | UiFlags::AlignCenter | UiFlags::KerningFitSpacing, 2);
+			line.position.y += LineHeights[pnumlines + infoStringLines];
+		}
```
AJenbo, line 276: 6th param of `DrawString()` is line height, is it not possible to use that instead of looping over `SplitByChar()`?

Source/control.cpp
```diff
@@ -264,17 +265,22 @@ void PrintInfo(const Surface &out)
...
 	}
 
 	for (int i = 0; i < pnumlines; i++) {
+
 		DrawString(out, panelstr[i], line, InfoColor | UiFlags::AlignCenter | UiFlags::KerningFitSpacing, 2);
-		line.position.y += LineHeights[pnumlines];
-	}
+		line.position.y += LineHeights[pnumlines + infoStringLines];
+	};
```
AJenbo, line 283:
```suggestion
	}
```

Source/control.cpp
burningserenity: Would I not still have to increment the y position for each line? How can I do that without looping?

burningserenity: I replaced the loop and its contents with what you suggested, but the result is that I only see the first line of InfoString rendered, and the line height is incorrect.
Screenshot_2022-09-07_18:46:12
Screenshot_2022-09-07_18:46:14
Screenshot_2022-09-07_18:46:16

burningserenity: Well the line height actually looks okay in 2/3

Source/control.cpp
```diff
@@ -264,16 +265,21 @@ void PrintInfo(const Surface &out)
...
-		DrawString(out, InfoString, line, InfoColor | UiFlags::AlignCenter | UiFlags::KerningFitSpacing, 2);
-		line.position.y += LineHeights[pnumlines];
+		for (string_view s : SplitByChar(InfoString, '\n')) {
+			DrawString(out, s, line, InfoColor | UiFlags::AlignCenter | UiFlags::KerningFitSpacing, 2);
+			line.position.y += LineHeights[pnumlines + infoStringLines];
+		}
 	}
 
 	for (int i = 0; i < pnumlines; i++) {
+
```
AJenbo, line 280:
```suggestion
```

Source/control.cpp
AJenbo: Ok, I'll try to have a look at it later. Other then that I think things looks good now.

burningserenity: I ran this through gdb and found that we're exiting early [here](https://github.com/diasurgical/devilutionX/blob/master/Source/engine/render/text_render.cpp#L386). The loop exits on the first newline.

Source/engine/render/text_render.cpp
```diff
@@ -631,8 +631,11 @@ uint32_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
 	else if (HasAnyOf(flags, UiFlags::AlignRight))
 		characterPosition.x += rect.size.width - lineWidth;
 
+	auto textLines = (int)std::count(text.begin(), text.end(), '\n');
+	int yMax = rect.position.y + rect.size.height * (textLines + 1);
```
burningserenity, line 635: Open to suggestions on this variable name

Source/control.cpp
burningserenity: Check out the changes :smiley:

Source/engine/render/text_render.cpp
```diff
@@ -631,8 +631,11 @@ uint32_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
 	else if (HasAnyOf(flags, UiFlags::AlignRight))
 		characterPosition.x += rect.size.width - lineWidth;
 
+	auto textLines = (int)std::count(text.begin(), text.end(), '\n');
+	int yMax = rect.position.y + rect.size.height * (textLines + 1);
+
 	int rightMargin = rect.position.x + rect.size.width;
-	const int bottomMargin = rect.size.height != 0 ? std::min(rect.position.y + rect.size.height, out.h()) : out.h();
+	const int bottomMargin = rect.size.height != 0 ? std::min(yMax, out.h()) : out.h();
```
StephenCWills, line 638: The rectangle defines the bounds for the string. It shouldn't depend on the number of newlines in the input text. Rather, it's up to the caller to determine what the appropriate bounds should be for the string that will be drawn.

I believe what you're going for here should be done by defining the full rectangle for the region inside the info panel and then using `UiFlags::VerticalCenter`.

Source/control.cpp
```diff
@@ -264,16 +265,19 @@ void PrintInfo(const Surface &out)
 	const int LineStart[] = { 70, 58, 52, 48, 46 };
 	const int LineHeights[] = { 30, 24, 18, 15, 12 };
 
-	Rectangle line { GetMainPanel().position + Displacement { 177, LineStart[pnumlines] }, { 288, 12 } };
+	auto infoStringLines = (int)std::count(InfoString.str().begin(), InfoString.str().end(), '\n');
+
+	Rectangle line { GetMainPanel().position + Displacement { 177, LineStart[infoStringLines + pnumlines] }, { 288, 12 } };
```
StephenCWills, line 270: I mentioned this in another PR, but the array access here is not bounds checked. `AddPanelString()` is applying validation to the input to ensure that `pnumlines` doesn't cause an overflow, and this change subverts it. To do this right, you'd probably need to combine all the strings using newlines and then truncate the result if it's more than 5 lines. Supporting both `InfoString` and `AddPanelString()` together definitely complicates this.

burningserenity: Took a stab at handling this

Source/engine/render/text_render.cpp
burningserenity: Worked better without the additional UI flag

Source/control.cpp
```diff
@@ -267,11 +267,33 @@ void PrintInfo(const Surface &out)
...
 
-	Rectangle line { GetMainPanel().position + Displacement { 177, LineStart[infoStringLines + pnumlines] }, { 288, 12 } };
+	int panelLines = pnumlines + infoStringLines;
+
+	std::string infoStringOut = std::string(InfoString.str());
+
+	if (panelLines >= 4) {
+		panelLines = 3;
+		pnumlines = panelLines - infoStringLines;
+		Log("PrintInfo unable to render everything - not enough lines");
```
glebm, line 277: This would be logged on every frame.

Source/control.cpp
```diff
@@ -267,11 +267,33 @@ void PrintInfo(const Surface &out)
 
 	auto infoStringLines = (int)std::count(InfoString.str().begin(), InfoString.str().end(), '\n');
 
-	Rectangle line { GetMainPanel().position + Displacement { 177, LineStart[infoStringLines + pnumlines] }, { 288, 12 } };
+	int panelLines = pnumlines + infoStringLines;
+
+	std::string infoStringOut = std::string(InfoString.str());
```
glebm, line 272: This is assigned here and then immediately reassigned on line 286

Source/control.cpp
```diff
@@ -267,11 +267,33 @@ void PrintInfo(const Surface &out)
...
+	}
+
+	Rectangle line { GetMainPanel().position + Displacement { 177, LineStart[panelLines] }, { 288, 12 } };
 
 	if (!InfoString.empty()) {
-		int lineHeight = LineHeights[pnumlines + infoStringLines];
-		DrawString(out, InfoString, line, InfoColor | UiFlags::AlignCenter | UiFlags::KerningFitSpacing, 2, lineHeight);
+		const int lineHeight = LineHeights[panelLines];
+		if (infoStringLines >= 4) {
+			infoStringLines = 3;
```
glebm: Rather than doing all these memory allocations on every frame, how about we only do this when `InfoString` is changed?

Source/control.cpp
```diff
@@ -267,11 +267,33 @@ void PrintInfo(const Surface &out)
...
+			infoStringOut = "";
+			int i = 0;
+			for (string_view s : SplitByChar(InfoString, '\n')) {
```
glebm, line 288: You can avoid having the `InfoStringOut` variable entirely, eliminating all these allocations on every frame, by simply doing this:


```c++
for (size_t end = 0; end < 3; ++end) {
  pos = InfoString.find('\n', end);
}
DrawString(out, InfoString.str().substr(0, pos), ...);
```

It'd be better to do this on assignment to InfoString rather than during rendering, so that the warning is only logged once (also more efficient).

glebm (changes requested):

Source/control.cpp
burningserenity: I took logging this only when the `InfoString` changes to mean overload the string_or_view = operator to validate the length of the string there, and print the info. I still get tons of logs about it though, it looks like it's still calling it on every frame.

burningserenity: I realize we probably don't want to change the = operator, since we probably need string_or_view for things besides the info panel :smile: 
What did you actually have in mind?

Source/utils/string_or_view.hpp
```diff
@@ -39,9 +40,23 @@ class StringOrView {
 
 	StringOrView &operator=(StringOrView &&other) noexcept
 	{
+		string_view::size_type end = 0;
```
glebm, line 43: `StringOrView` is a general class that is simply a union type between a `string` and a `string_view`. This behaviour is specific to `InfoString` and certainly doesn't belong here.

You can add a `SetInfoString(StringOrView&&)` (probably the easiest) function or add a separate class for `InfoString`.

Source/utils/string_or_view.hpp
```diff
@@ -39,9 +40,23 @@ class StringOrView {
...
+				break;
+		}
+		if (end < other.str().size()) {
+			LogWarn("PrintInfo unable to render everything - not enough lines");
+		}
+
 		if (owned_) {
 			if (other.owned_) {
-				str_ = std::move(other.str_);
+				str_ = std::string(other.str().substr(0, end));
```
glebm, line 59: This eliminates the benefit of the move constructor, please revert the changes in this file.

glebm (changes requested):

Source/control.cpp
glebm: Ah, just saw this, please see my new comment

StephenCWills (changes requested):

Source/utils/string_or_view.hpp
burningserenity: Do we need the view portion of StringOrView? I’m trying to make a new class that extends that one, but having some difficulty. Would it be okay if I made a new class based on std::string, or would that break something?

burningserenity: Never mind. It would make several things fail to compile.

StephenCWills:
@burningserenity FYI, you might want to have a look at #5367 and #5369. My reviews on this PR helped shape the work I did there so I hope you don't feel I'm undercutting your effort here. I also don't want you to be spinning your wheels on this for too long, and I'm hoping these two PRs will get us moving toward merging #5357.

burningserenity:
> @burningserenity FYI, you might want to have a look at #5367 and #5369. My reviews on this PR helped shape the work I did there so I hope you don't feel I'm undercutting your effort here. I also don't want you to be spinning your wheels on this for too long, and I'm hoping these two PRs will get us moving toward merging #5357.

Ah bummer. Oh well, I'm glad I could inspire you, at least. I'll close this and watch your two PRs for when to move on #5357 .

Issue #3831, opened 2021-12-22
Monster kill count reset to 0 when starting new game from 1.2.1 save


karaluh:
**Important information**
Win 10 / 1.3.0 / downloaded

**Describe**
After starting a new game in 1.3.0 from a 1.2.1 save monster kill count gets reset to 0.

**To Reproduce**
Use the attached save to start a new game.

**Additional context**
If you load the game first and then start a new one the counter works properly.

xGnoSiSx:
@AJenbo Just a reminder that this original issue, (that i did find, before opening mine) was/is for single player saves. Mine was exclusively for MP characters, and an enhancement over D1. Thanks!

AJenbo:
> @AJenbo Just a reminder that this original issue, (that i did find, before opening mine) was/is for single player saves. Mine was exclusively for MP characters, and an enhancement over D1. Thanks!

Both singleplayer and multiplayer is affected by this, in multiplayer you always start a new game.

PR #5269, opened 2022-08-22
Keep track of MonsterKillCounts across New Games and in MP

sixcy:
This patch preserves MonsterKillCounts values in MP and when using New
Game in SP.

It is implemented by adding a "monsterkillcounts" section in the save
archive.

This patch also changes the type of MonsterKillCounts elements to be
`int32_t` instead of `int`, for better compatibility across platforms,
as the `int` datatype might not always be 4 bytes.

Implements https://github.com/diasurgical/devilutionX/issues/3831

I tested it locally on dlvl1 in the following scenarios : 
- launch devilutionX, create a MP game, kill some monsters. Quit devilutionX. Relaunch it. Start a MP game: the kill counters were preserved.
- launch devilutionX, create a SP game, kill some monsters, Save Game, quit devilutionX, relaunch devilutionX, create New Game from SP menu. The kill counters seemed to be preserved.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -27,6 +27,7 @@
 #include "lighting.h"
 #include "menu.h"
 #include "missiles.h"
+#include "monster.h"
 #include "mpq/mpq_writer.hpp"
 #include "pfile.h"
 #include "qol/stash.h"
@@ -1971,6 +1972,13 @@ void LoadHeroItems(Player &player)
 	gbIsHellfireSaveGame = gbIsHellfire;
 }
 
+void LoadMonsterKillCounts()
+{
+	LoadHelper file(OpenSaveArchive(gSaveNumber), "monsterkillcounts");
+	for (int32_t &monsterKillCount : MonsterKillCounts)
+		monsterKillCount = file.NextBE<int32_t>();
+}
+
 constexpr uint8_t StashVersion = 0;
 
 void LoadStash()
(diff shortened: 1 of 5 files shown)


qndel:
I think that int -> `int32_t` change is an overkill, if a platform doesn't have a 4 byte int, I bet devilutionx wouldn't even work on it

sixcy:
> I think that int -> `int32_t` change is an overkill, if a platform doesn't have a 4 byte int, I bet devilutionx wouldn't even work on it
Perhaps - would there be any reason not to include that change?

qndel:
> > I think that int -> `int32_t` change is an overkill, if a platform doesn't have a 4 byte int, I bet devilutionx wouldn't even work on it
> 
> Perhaps - would there be any reason not to include that change?

Minimizing the diff? ;p

sixcy:
> > > I think that int -> `int32_t` change is an overkill, if a platform doesn't have a 4 byte int, I bet devilutionx wouldn't even work on it
> > 
> > 
> > Perhaps - would there be any reason not to include that change?
> 
> Minimizing the diff? ;p

I can remove these changes if it is required

sixcy:
The SP behavior is a bit odd, as I did not synchronize the kill count from SP and the one from the new entry.
I will work on that and close this PR for now

Source/loadsave.cpp
```diff
@@ -1971,6 +1972,13 @@ void LoadHeroItems(Player &player)
 	gbIsHellfireSaveGame = gbIsHellfire;
 }
 
+void LoadMonsterKillCounts()
+{
+	LoadHelper file(OpenSaveArchive(gSaveNumber), "monsterkillcounts");
```
qndel, line 1977: look how it's done in the function above, check if file is valid

PR #5270, opened 2022-08-22
Keep track of MonsterKillCounts across New Games and in MP

sixcy:
This patch preserves MonsterKillCounts values in MP and when using New
Game in SP.

It is implemented by adding a "monsterkillcounts" section in the save
archive.

This patch also changes the type of MonsterKillCounts elements to be
`int32_t` instead of `int`, for better compatibility across platforms,
as the `int` datatype might not always be 4 bytes.

Implements https://github.com/diasurgical/devilutionX/issues/3831

I tested it locally on dlvl1 in the following scenarios : 
- launch devilutionX, create a MP game, kill some monsters. Quit devilutionX. Relaunch it. Start a MP game: the kill counters were preserved.
- launch devilutionX, create a SP game, kill some monsters, Save Game, quit devilutionX, relaunch devilutionX, create New Game from SP menu. The kill counters seemed to be preserved.
- launch devilutionX, Save Game, kill some monsters, Load Game: the kill counters are appropriately reset to their previous value

Previous PR: https://github.com/diasurgical/devilutionX/pull/5269
Apologies for the duplicated PR, I wanted to reopen the 5269 but it is not possible to do after a force push. I am still discovering Github PR system (we use Phabricator at work). In the future i will either avoid closing the PR, or reopen before updating it.

### DIFF
--- a/Source/loadsave.cpp
+++ b/Source/loadsave.cpp
@@ -13,6 +13,7 @@
 #include <SDL.h>
 #include <fmt/compile.h>
 
+#include "DiabloUI/selgame.h"
 #include "automap.h"
 #include "codec.h"
 #include "control.h"
@@ -24,10 +25,13 @@
 #include "engine/random.hpp"
 #include "init.h"
 #include "inv.h"
+#include "levels/gendung.h"
 #include "lighting.h"
 #include "menu.h"
 #include "missiles.h"
+#include "monster.h"
 #include "mpq/mpq_writer.hpp"
+#include "multi.h"
 #include "pfile.h"
 #include "qol/stash.h"
 #include "stores.h"
@@ -1971,6 +1975,28 @@ void LoadHeroItems(Player &player)
 	gbIsHellfireSaveGame = gbIsHellfire;
 }
 
+constexpr const char *MonsterKillCountSection = "monsterKillCounts";
+constexpr uint32_t MonsterKillCountVersion = 0;
+
+static bool LoadMonsterKillCountsInto(int32_t *mkc, _difficulty difficulty)
+{
+	LoadHelper file(OpenSaveArchive(gSaveNumber), MonsterKillCountSection);
+	if (!file.IsValid() || file.NextBE<uint32_t>() > MonsterKillCountVersion)
+		return false;
+
+	file.Skip(difficulty * sizeof(MonsterKillCounts));
+	for (size_t m = 0; m < MaxMonsters; m++)
+		mkc[m] = file.NextBE<int32_t>();
+
+	return true;
+}
+
+void LoadMonsterKillCounts()
+{
+	if (!LoadMonsterKillCountsInto(MonsterKillCounts, nDifficulty))
+		return;
+}
+
 constexpr uint8_t StashVersion = 0;
 
 void LoadStash()
@@ -2095,7 +2121,7 @@ void LoadGame(bool firstflag)
 	ActiveMonsterCount = tmpNummonsters;
 	ActiveObjectCount = tmpNobjects;
 
-	for (int &monstkill : MonsterKillCounts)
+	for (int32_t &monstkill : MonsterKillCounts)
 		monstkill = file.NextBE<int32_t>();
 
 	if (leveltype != DTYPE_TOWN) {
@@ -2238,6 +2264,26 @@ void SaveHeroItems(MpqWriter &saveWriter, Player &player)
 		SaveItem(file, item);
 }
 
+void SaveMonsterKillCounts(MpqWriter &saveWriter)
+{
+	int32_t mkc[DIFF_LAST + 1][MaxMonsters] = { { 0 } };
+
+	for (int diff = DIFF_NORMAL; diff < DIFF_LAST; diff++) {
+		if (diff == nDifficulty) {
+			memcpy(mkc[diff], MonsterKillCounts, sizeof(MonsterKillCounts));
+			continue;
+		}
+		LoadMonsterKillCountsInto(mkc[diff], static_cast<_difficulty>(diff));
+	}
+
+	SaveHelper file(saveWriter, MonsterKillCountSection, sizeof(MonsterKillCountVersion) + (DIFF_LAST + 1) * sizeof(MonsterKillCounts));
+
+	file.WriteBE<uint32_t>(MonsterKillCountVersion);
+	for (int diff = DIFF_NORMAL; diff < DIFF_LAST; diff++)
+		for (size_t m = 0; m < MaxMonsters; m++)
+			file.WriteBE<int32_t>(mkc[diff][m]);
+}
+
 void SaveStash(MpqWriter &stashWriter)
 {
 	const char *filename;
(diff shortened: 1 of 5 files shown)


sixcy:
I thought there was some desynchronization of some kind but it works as intended: Load Game should not preserve the kill counters but instead reset them to where they were at the moment of saving. This behavior is preserved with this patch.

Note that the vanilla behavior of making SP New Game without saving (if you make a new character, kill some monsters, make New Game: the kill counters are preserved) is still there. Ideally in that scenario these kill counters should be reset to 0, but I do not think it matters much

sixcy:
Note that the vanilla behavior of making SP New Game without saving (if you make a new character, kill some monsters, make New Game: the kill counters are preserved) is still there. Ideally in that scenario these kill counters should be reset to 0, but I do not think it matters much.

As soon as the player saves once, this behavior or bug stops existing (making New Game without Saving resets the kill counter to where it was on the last save, instead of persisting)

kphoenix137:
Great change! Would love this

obligaron:
I think this is a good change. 👍
Especially when you look at the multiplayer workaround. 🥴 

Some questions related to the pr:
1. Should we have different kill-counter per difficulty?
2. Should the new file have a version number, if we want to change the format later? (for example see `additionalMissiles`)
3. Should the file be named camelCase `monsterKillCounter` (like `additionalMissiles`)?
4. Should we change the type from `int32_t` to `uint64_t`? There are some crazy people out there. 😜 
5. `MonsterKillCounts` array is defined with a size of 200/`MaxMonsters`. But there is a comment `BUGFIX: replace MonsterKillCounts[MaxMonsters] with MonsterKillCounts[NUM_MTYPES]`. Should the new format be more correct or be more flexible for modders? (for example it could be based on a `unordered_map` instead of a array)

sixcy:
Thanks for the feedback!

1) This is an interesting question. I have not thought about difficulties at all. It would be logical for the kill counts to be per difficulty - this also sounds doable programming wise. I can certainly look at it
2) Good idea, I will add a version number
3) I followed the style of "heroitems" (can be seen in loadsave.cpp/LoadHeroItems)
4) I do not think this is a good idea to change the bitwidth, because then the Save Game / Load Game format would also need to be changed, saves would become incompatible. However, the fear that someone would go over 2,147,483,647 is valid ; I think someone trying really hard might get that number eventually.  I will add a check to prevent overflow (the kill count would clamp to this maximum value).
5) Interesting take, but I would say it is beyond the scope of this PR

Note that changing `int32_t` to `uint64_t` could be done in a later patch if someone does reach the max value.

I have a side question: how do you handle GitHub PRs when the developer is planning changes? Do you close the PR and re-open it later? Or do you keep it open?

obligaron:
4, We could use `int32_t` when we write the old format for compatibility and only use `int64_t` for the new file.
We have done something similar in other cases.
5, If we need a compatibility layer for the old format anyway (see above), I think it's doable. But it was only a suggestion.

github question:
When the changes are done in the near future, I would keep the pr open.
When the changes are taking some time but will be done, I would convert the pr to draft to signal that no one should review it yet.
When no one plans to works on the pr and no new changes come in to address feedback, I would close the pr.

sixcy:
> We could use int32_t when we write the old format for compatibility and only use int64_t for the new file.
We have done something similar in other cases.

Do you mean to write/read int32_t or int64_t depending on some version number somewhere in the save file, or something else?

kphoenix137:
> I think this is a good change. 👍 Especially when you look at the multiplayer workaround. 🥴
> 
> Some questions related to the pr:
> 
>     1. Should we have different kill-counter per difficulty?
> 
>     2. Should the new file have a version number, if we want to change the format later? (for example see `additionalMissiles`)
> 
>     3. Should the file be named camelCase `monsterKillCounter` (like `additionalMissiles`)?
> 
>     4. Should we change the type from `int32_t` to `uint64_t`? There are some crazy people out there. 😜
> 
>     5. `MonsterKillCounts` array is defined with a size of 200/`MaxMonsters`. But there is a comment `BUGFIX: replace MonsterKillCounts[MaxMonsters] with MonsterKillCounts[NUM_MTYPES]`. Should the new format be more correct or be more flexible for modders? (for example it could be based on a `unordered_map` instead of a array)

There is no reason to use `uint64_t` for monster kill counts per monster type. It is physically impossible for a person to kill 4.28 billion of a single monster type. If you somehow killed 1000 Blood Knights per hour, it would take you 4,294,967 hours (490 years) of gameplay.

sixcy:
I am unsure about the physically impossible - killing 1000 BKs per hour sounds plausible to me (with fastest speed setting). Also, it's 2 billion not 4 because it's `int32_t`. But I agree this is far stretched, and in practice only someone running a bot of some kind would achieve that in the next 10 years.

Also, the save file won't broke if that ever happens. It will just go back to some garbage value (likely -2 billion).

kphoenix137:
> I am unsure about the physically impossible - killing 1000 BKs per hour sounds plausible to me (with fastest speed setting). Also, it's 2 billion not 4 because it's `int32_t`. But I agree this is far stretched, and in practice only someone running a bot of some kind would achieve that in the next 10 years.
> 
> Also, the save file won't broke if that ever happens. It will just go back to some garbage value (likely -2 billion).

We can adjust the numbers. Let's say we kill 20,000 Blood Knights per hour. That averages to killing 5.55 Blood Knights per second, for every second that elapses during 1 hour of gameplay. Pretty unrealistic since you spend time in town, and spend time doing other not-killing-Blood-Knight type things, but we'll give this the benefit of the doubt. with `int32_t`, you are able to kill up to 2,147,483,647 Blood Knights before this number overflows to a negative number. At killing 20,000 Blood Knights per hour, to hit the signed int32 limit, it will take you exactly 107,374.1824 hours of gameplay. At 24 hours in a day, that is 4,473.924266666667 days of gameplay, and at 365.25 days per year (factoring in leap years occuring once every 4 years getting an extra 1 day), that is 12.24893707506274 years of gameplay. If you play DevilutionX for 6 hours per day without missing a day every, it will take you 48.99574830025097 years to complete this task. So I guess physically impossible? No. Likely to happen? I wouldn't even bet a dollar on it.

obligaron:
> Do you mean to read int32_t when reading from an old savefile, but to write/read int64_t in the new section? This could be possible. I could also use this section for Load Game/Save Game instead of storing the kill counts in two separate places (in the new section and in the "load game" save file)

When running the game, in-memory there only exists the new logic (for example a `unordered_map` instead of a array).
When saving (`loadsave.cpp`) the new format is written and additionally the old format is also written for compatibility.
The same happens while loading. If the new format is present the old format isn't used and only the new format is used.
This way we can add new features and keep backwards compatibility. And the compatibility only happens when handling the save game reading/writing
We did something like this for missiles and hotkeys.

And for the difficulty support for this pr we also need this to some extend.
That's why I asked if we could think more freely about how we want to store the new data. 🙂 

btw: I have no problem with sticking with `int32_t`. I only wanted to ask to make sure, we thought about it. 😉

AJenbo:
A bot will have a limit of around 1500x game speed.

I don't think switching to uint32 would be a problem if this is critical to someone...

sixcy:
I updated the commit:
- the kill counts are now stored per difficulty. I tested that switching from one difficulty to the other does not reset the counter. It seems to work fine.
- changed the file name to camelCase: monsterKillCounts
- added a version number (uint32_t)
- rebased on top of master

These changes are ready to be reviewed

PR #5272, opened 2022-08-22
Only load Hellfire music when Hellfire is active

StephenCWills:
The game was attempting to load Nest and Crypt music in the menu when playing DevilutionX Diablo, causing silence in the menus because the assets failed to load.


AJenbo (approved):
I guess that's better then `#ifdef`

PR #5273, opened 2022-08-23
Support unpacked MPQs from devilutionx-mpq-tools

glebm:
TODO: Fix palette in menu by replace devilutionx's `black*.pcx` with `*.pal` files.


AJenbo:
`DEVILUTIONX_CEL_EXT` isn't a particularly nice solution, how about replacing `LoadCel()` with a function that simply takes the name, replaces the extension and passes it to `LoadClx()`?

glebm:
That would mean an extra allocation on every `Load*` call.
I suppose we could make a `LOAD_CEL` macro to avoid both the noise and the allocation. What do you think?

glebm:
Actually we can avoid the string allocations because we know that the paths are limited by mpq max path.

glebm:
> DEVILUTIONX_CEL_EXT is a bit noisy and not as readable, how about replacing LoadCel() with a function that simply takes the name, replaces the extension and passes it to LoadClx()?

Done, rather than replacing the extension we simply append, which also reduces the sizes of strings in the binary a bit.

AJenbo:
I like that solution a lot better.

glebm:
Rebased

AJenbo (approved):

PR #5274, opened 2022-08-23
Replace default palette PCX files with PAL files

glebm:
These PCX files were only used to load the default palette.
Replaces them with PAL files.


AJenbo:
Packaging/resources/assets/ui_art/hellfire.pal looks more like a pcx and not a pal at 899 bytes

glebm:
This seems to be a GitHub UI bug, the files are both at 768 bytes

PR #5276, opened 2022-08-23
Add _DEBUG definition to CMake's Debug release type

sixcy:
This patch defines _DEBUG when building with `DCMAKE_BUILD_TYPE=Debug`
to activate all the debug features for CMake users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,10 @@ if(MSVC AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT DISABLE_LTO)
   set(BUILD_TESTING OFF)
 endif()
 
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+  add_definitions(-D_DEBUG)
+endif()
+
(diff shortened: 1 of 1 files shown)


AJenbo:
What features have we been missing out on?

sixcy:
One example is using `+god` to activate godmode :
```
./devilutionX +god
```

This is only possible if you build with `_DEBUG` defined. Which does not seem to be the case right now when using CMake and Linux.

VS users (on Windows) already have `_DEBUG` defined (this is a feature from the VS compiler tools) so this only oncerns devilutionX developers/modders that are on Linux

AJenbo:
> This is only possible if you build with _DEBUG defined. Which does not seem to be the case right now when using CMake and Linux.

I use CMake on Linux, I don't have Windows or VS, god mode works fine for me.

sixcy:
I will close the PR until I figure out what is different between AJenbo, staphen and me

sixcy:
This PR can be permanently closed. This was a problem of CMake version. I had 3.16.3, it did not work. Upgraded to 3.22.1, now it works fine

PR #5279, opened 2022-08-24
Update Italian localization

M0Rf30:
As in title,
Regards


AJenbo (approved):
Thanks, I was wondering how much had changed since last :)

PR #5285, opened 2022-08-25
Fix macOS build with the SDL2 2.24.0

glebm:
SDL2 2.24.0 has a bug in its cmake for macOS: https://github.com/libsdl-org/SDL/issues/6119

This bug has already been fixed upstream. This is a workaround until the next version of SDL2 is released.


StephenCWills (approved):
Good timing, because I wanted to test the build in my fork.

PR #5305, opened 2022-08-29
3rdParty/SDL2: Disable SDL_TEST

glebm:
The fully-vendored source distribution removes all test files to reduce the size, which causes a CMake configuration error if `SDL_TEST` is on.


StephenCWills (approved):
Thanks, I wasn't sure why the tests were missing, but I was thinking of the exact same solution.

PR #5306, opened 2022-08-29
Add `XDG_DATA_DIRS` to MPQ search paths

### LINKED ISSUE #5304: Flatpak cannot find devilutionx.mpq because app does not search its preferred path
### Operating System

Linux x64

### DevilutionX version

1.4.1 (latest release)

### Describe

After talking to folks in Flatpak Matrix room, the reason why Flatpak cannot find devilutionx.mpq, even though it's installed into /app/share location, is because paths are hardcoded to /usr/share

https://github.com/diasurgical/devilutionX/blob/a3e94d6b558893afd6f6fcae4582a6168ab67e53/Source/init.cpp#L96-L97

From gasinvein in Flatpak Matrix room:

> The proper fix would be patching the app to search for the files in paths specified in XDG_DATA_DIRS. Or at least adding /app/... path to search list.

### To Reproduce

1. Install `flatpak` and `flatpak-builder` 
2. Install Freedesktop 21.08 runtime and SDK: `flatpak install flathub org.freedesktop.Platform//21.08 org.freedesktop.Sdk//21.08`
3. Copy https://github.com/diasurgical/devilutionX/blob/master/Packaging/nix/org.diasurgical.DevilutionX.yml to any empty writable folder and go to it in terminal
4. Build and install flatpak: `flatpak-builder --user --install build org.diasurgical.DevilutionX.yml --force-clean`
5. Launch flatpak from Start menu or run it from terminal: `flatpak run org.diasurgical.DevilutionX`

### Expected Behavior

Error dialogue is displayed to show data files are missing instead of game screen showing the error

### Additional context

_No response_

glebm:
Fixes #5304

### DIFF
--- a/Source/init.cpp
+++ b/Source/init.cpp
@@ -23,6 +23,7 @@
 #include "utils/language.h"
 #include "utils/log.hpp"
 #include "utils/paths.h"
+#include "utils/str_split.hpp"
 #include "utils/ui_fwd.h"
 #include "utils/utf8.hpp"
 
@@ -93,8 +94,21 @@ std::vector<std::string> GetMPQSearchPaths()
 		paths.pop_back();
 
 #if defined(__unix__) && !defined(__ANDROID__)
-	paths.emplace_back("/usr/share/diasurgical/devilutionx/");
-	paths.emplace_back("/usr/local/share/diasurgical/devilutionx/");
+	// `XDG_DATA_HOME` is usually the root path of `paths::PrefPath()`, so we only
+	// add `XDG_DATA_DIRS`.
+	const char *xdgDataDirs = std::getenv("XDG_DATA_DIRS");
+	if (xdgDataDirs != nullptr) {
+		for (const string_view path : SplitByChar(xdgDataDirs, ':')) {
+			std::string fullPath(path);
+			if (!path.empty() && path.back() != '/')
+				fullPath += '/';
+			fullPath.append("diasurgical/devilutionx/");
+			paths.push_back(std::move(fullPath));
+		}
+	} else {
+		paths.emplace_back("/usr/local/share/diasurgical/devilutionx/");
+		paths.emplace_back("/usr/share/diasurgical/devilutionx/");
+	}
 #elif defined(NXDK)
 	paths.emplace_back("D:\\");
 #elif (defined(_WIN64) || defined(_WIN32)) && !defined(__UWP__) && !defined(NXDK)
--- a/Source/utils/str_split.hpp
+++ b/Source/utils/str_split.hpp
@@ -1,0 +1,98 @@
+#pragma once
+
+#include <iterator>
+#include <type_traits>
+
+#include "utils/stdcompat/string_view.hpp"
+
+namespace devilution {
+class SplitByCharIterator {
+public:
+	using iterator_category = std::forward_iterator_tag;
+	using value_type = string_view;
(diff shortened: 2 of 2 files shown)


Source/init.cpp
```diff
@@ -93,8 +94,21 @@ std::vector<std::string> GetMPQSearchPaths()
...
+			std::string fullPath(path);
+			if (!path.empty() && path.back() != '/')
+				fullPath += '/';
+			fullPath.append("diasurgical/devilutionx/");
+			paths.push_back(std::move(fullPath));
+		}
+	} else {
+		paths.emplace_back("/usr/share/diasurgical/devilutionx/");
+		paths.emplace_back("/usr/local/share/diasurgical/devilutionx/");
+	}
```
gasinvein, line 111: According to the spec, the default value for `XDG_DATA_DIRS` is `/usr/local/share:/usr/share`, i.e. the same as these paths but in reverse order. So I guess you just assume `/usr/local/share:/usr/share` if `XDG_DATA_DIRS` is empty or unset.

glebm: Good point, I've fixed the order for the default to be the same as what's in `XDG_DATA_DIRS` by default.

StephenCWills:
Do we need to check `XDG_DATA_HOME` explicitly as well?

> https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
> `$XDG_DATA_DIRS` defines the preference-ordered set of base directories to search for data files in addition to the `$XDG_DATA_HOME` base directory. The directories in `$XDG_DATA_DIRS` should be seperated with a colon ':'.

glebm:
`SDL_GetPrefPath` defaults to `${XDG_DATA_HOME}/diasurgical/devilution` already, so no need.

tsunamistate:
@glebm could you please rebase this so that the branch includes https://github.com/diasurgical/devilutionX/pull/5305?

Source/utils/str_split.hpp
```diff
@@ -0,0 +1,98 @@
...
+		slice_ = slice_.substr(0, slice_.find(split_by_));
+		return *this;
+	}
+
+	SplitByCharIterator operator++(int) const
+	{
+		auto copy = *this;
+		++copy;
+		return copy;
+	}
```
StephenCWills, line 50: I'm not 100% sure, but this feels wrong.

```suggestion
	SplitByCharIterator operator++(int) const
	{
		auto copy = *this;
		++*this;
		return copy;
	}
```

glebm: Fixed!

tsunamistate:
@glebm it seems that #5305 didn't fix SDL2 in fully vendored tarballs.
Based tarball on https://github.com/diasurgical/devilutionX/commit/4285db299cb9c9a4d3aa5b5cdbbb71ed22858f05, but I still get:
```
CMake Error at dist/sdl2-src/CMakeLists.txt:3159
No SOURCES given to target: SDL2_test
```
on generation step. This blocks me from testing this :(

StephenCWills:
How bizarre. It seems like a waste to rerun the `linux_x86_64` job since it clearly passed.

glebm:
@tsunamistate, that's odd, can you try again now that it's been merged? Perhaps that commit is from before the rebase

tsunamistate:
@glebm Tried to base on master's 9feb3139e2af3bc7dfcaa922f9139ee89aeb708e which definitely includes #5305, but no dice, same error on generation step

PR #5307, opened 2022-08-29
Add killall, breakall, openall and clearall debug commands

sixcy:
In the current level:
- killall: kill all monsters
- openall: open all chests and sarcophagus
- breakall: break all barrels
- clearall: mix of the three above commands

This is an arranged cherry-pick from devilutionX-ironman 98bbbc337d9e83e6dedc358fe0832630b614d34d, where those debug commands have been useful to me

CC @obligaron who was interested in the PR

### DIFF
--- a/Source/debug.cpp
+++ b/Source/debug.cpp
@@ -4,6 +4,8 @@
  * Implementation of debug functions.
  */
 
+#include "missiles.h"
+#include "objects.h"
 #ifdef _DEBUG
 
 #include <fstream>
@@ -984,6 +986,67 @@ std::string DebugCmdChangeTRN(const string_view parameter)
 	return out;
 }
 
+std::string DebugCmdOpenAllChests(const string_view /*parameter*/)
+{
+	int countOpened = 0;
+	for (int i = 0; i < ActiveObjectCount; i++) {
+		Object &o = Objects[ActiveObjects[i]];
+		if ((o.IsChest() || o.isSarcophagus()) && !o.isOpen()) {
+			OperateObject(*MyPlayer, o);
+			countOpened++;
+		}
+	}
+	return StrCat(countOpened, " opened");
+}
+
+std::string DebugCmdBreakAllBarrels(const string_view /*parameter*/)
+{
+	int countBroken = 0;
+	for (int i = 0; i < ActiveObjectCount; i++) {
+		Object &o = Objects[ActiveObjects[i]];
+		if (o.IsBarrel() && !o.IsBroken()) {
+			BreakObject(*MyPlayer, o);
+			countBroken++;
+		}
+	}
+	return StrCat(countBroken, " broken");
+}
+
+std::string DebugCmdKillAllMonsters(const string_view parameter)
+{
+	int countKilled = 0;
+	for (size_t i = 0; i < ActiveMonsterCount; i++) {
+		int mi = ActiveMonsters[i];
+		auto &monster = Monsters[mi];
+		if ((monster.hitPoints >> 6) <= 0)
+			continue;
+		if (monster.position.tile == GolemHoldingCell)
+			continue;
+		M_StartKill(monster, *MyPlayer);
+		countKilled++;
+	}
+	return StrCat(countKilled, " killed");
+}
+
+std::string DebugCmdClearLevel(const string_view parameter)
+{
+	std::string subcmd { parameter };
+	std::transform(subcmd.begin(), subcmd.end(), subcmd.begin(), ::tolower);
+	bool doAll = subcmd == "all";
+	std::string msg {};
+	if (doAll || subcmd == "chests")
+		msg += DebugCmdOpenAllChests("");
+	if (doAll || subcmd == "barrels")
+		msg += (msg.empty() ? "" : ", ") + DebugCmdBreakAllBarrels("");
+	if (doAll || subcmd == "monsters")
+		msg += (msg.empty() ? "" : ", ") + DebugCmdKillAllMonsters("");
+
+	if (msg.empty())
+		return "Type not recognized.\nPlease write 'clear {type}'.\nPossible values for {type}: all, chests, barrels, monsters";
+
+	return msg;
+}
+
 std::vector<DebugCmdItem> DebugCmdList = {
 	{ "help", "Prints help overview or help for a specific command.", "({command})", &DebugCmdHelp },
 	{ "give gold", "Fills the inventory with gold.", "", &DebugCmdGiveGoldCheat },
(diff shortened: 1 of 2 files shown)


Source/debug.cpp
```diff
@@ -984,6 +985,55 @@ std::string DebugCmdChangeTRN(const string_view parameter)
...
+		int mi = ActiveMonsters[i];
+		auto &monster = Monsters[mi];
+		if ((monster.hitPoints >> 6) <= 0)
+			continue;
+		if (monster.position.tile == GolemHoldingCell)
+			continue;
+		M_StartKill(monster, *MyPlayer);
+		countKilled++;
+	}
+	return fmt::format("{:d} killed", countKilled);
```
glebm, line 1026: For simple cases of just concatenating some strings and numbers, consider using `StrCat`:

```c++
return StrCat(countKilled, " killed");
```

qndel:
> In the current level:
> 
> * killall: kill all monsters
> * openall: open all chests and sarcophagus
> * breakall: break all barrels
> * clearall: mix of the three above commands
> 
> This is an arranged cherry-pick from devilutionX-ironman [98bbbc3](https://github.com/diasurgical/devilutionX/commit/98bbbc337d9e83e6dedc358fe0832630b614d34d), where those debug commands have been useful to me
> 
> CC @obligaron who was interested in the PR

he was only interested in the "kill all monsters part", I can't imagine a situation where anyone would need a command to break barrels/open chests

Source/objects.h
```diff
@@ -191,6 +191,17 @@ struct Object {
...
 	}
 
+	[[nodiscard]] constexpr bool IsSarcophagus() const
+	{
+		return _otype == _object_id::OBJ_SARC;
+	}
+
+	[[nodiscard]] constexpr bool IsOpen() const
+	{
+		assert(IsChest() || IsSarcophagus());
```
qndel, line 201: ?

Source/debug.cpp
```diff
@@ -984,6 +985,55 @@ std::string DebugCmdChangeTRN(const string_view parameter)
...
+			countOpened++;
+		}
+	}
+	return fmt::format("{:d} opened", countOpened);
+}
+
+std::string DebugCmdBreakAllBarrels(const string_view /*parameter*/)
+{
+	int countBroken = 0;
+	for (auto &o : Objects) {
```
ephphatha, line 1004: Both of these loops should iterate over active objects, just in case this is called on a level which doesn't reinitialise the entire array.

Source/objects.h
sixcy: Please elaborate on your question whose meaning I cannot get

Source/debug.cpp
sixcy: Alright

Source/debug.cpp
sixcy: Will try

Source/objects.h
```diff
@@ -191,6 +191,17 @@ struct Object {
 		return IsAnyOf(_otype, _object_id::OBJ_CHEST1, _object_id::OBJ_CHEST2, _object_id::OBJ_CHEST3, _object_id::OBJ_TCHEST1, _object_id::OBJ_TCHEST2, _object_id::OBJ_TCHEST3);
 	}
 
+	[[nodiscard]] constexpr bool IsSarcophagus() const
```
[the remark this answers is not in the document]
sixcy: Will update

sixcy:
> he was only interested in the "kill all monsters part", I can't imagine a situation where anyone would need a command to break barrels/open chests

I can remove those if it is required

qndel:
I mean debug code is supposed to be useful and I really can't imagine a situation where I'd need to open chests/break barrels - it will be dead code that we have to maintain so I'm against including these. Unless someone else thinks they would be useful xD

StephenCWills:
> I mean debug code is supposed to be useful and I really can't imagine a situation where I'd need to open chests/break barrels - it will be dead code that we have to maintain so I'm against including these. Unless someone else thinks they would be useful xD

TBH, I had a need for it once when trying to verify my hypothesis for why Diablo RNG could potentially generate duplicate items in the dungeon. Rather than a debug command, though, I just hacked it into the code when I entered the dungeon level. This is a pretty odd use case so I'm not sure if it really justifies having a debug command.

StephenCWills:
@sixcy If no one is using them, then probably no one is going to notice if the commands break. By extension, no one will be maintaining them. I'd say the issue with including useless commands is more of a usability concern. Consider that the `help` command is already overrun with info, and you're adding more even commands for questionable value in return.

sixcy:
> @sixcy If no one is using them, then probably no one is going to notice if the commands break. By extension, no one will be maintaining them. I'd say the issue with including useless commands is more of a usability concern. Consider that the `help` command is already overrun with info, and you're adding even more commands for questionable value in return.

Thanks that is a good point

In that case I can bridge them into one unified command : `clear [monsters|barrels|chests|all]`.

StephenCWills:
> Would bridging them into one unified command solve the problem ? `clear [monsters|barrels|chests|all]`.

Seems okay to me.

sixcy:
Alright, I will do that then 👍

Source/objects.h
qndel: what's the point of this assert?

sixcy:
Sorry, I am still experimenting a bit with the review system. It's too bad that force pushing seems to discard previous reviews. I'm not sure what's the best way to handle updating a PR

I went with:
 - pushing the changes in a separate commit so you can notice what got changed
 - squashing the commits and force pushing to update the commit message and have it all in a neat commit

sixcy:
Except that I am just realizing that the second commit does not display. Oh well. Here are the list of changes from the initial code review:
    - `isSarcophagus()` in snakeCase
    - handle doors in `IsOpen()`. Fail with error if it's not a door, chest
      or sarcophagus
    - Write a doxygen description in IsOpen() to say the method is only for
      doors, sarcophagus and chests
    - Iterate over active objects instead of all objects
    - Use StrCat instead of fmt::format for simple cases
    - Agglomerate the new debug commands into a unified "clear" command to
      avoid adding too much of them

qndel:
squashing commits = merging them into 1 lol

AJenbo:
> It's too bad that force pushing seems to discard previous reviews

No cause if you completely change the code it needs to be reviewed again :D

Source/objects.h
```diff
@@ -191,6 +191,26 @@ struct Object {
...
+	}
+
+	/**
+	 * @brief Check if the object is open
+	 *
+	 * Only chests, sarcophagus and doors are handled.
+	 */
+	[[nodiscard]] constexpr bool isOpen() const
+	{
+		assert((IsChest() || isSarcophagus() || isDoor()) && "IsOpen can only be called on chests, sarcophagus and doors");
```
AJenbo, line 206:
```suggestion
		assert((IsChest() || isSarcophagus() || isDoor()) && "isOpen can only be called on chests, sarcophagus and doors");

```

sixcy:
> GitHub also shows you what was changed by the latest force push. Neat commits are preferred.
Alright, good to know

Source/objects.h
```diff
@@ -191,6 +191,27 @@ struct Object {
...
+	 * Only chests, sarcophagus and doors are handled.
+	 */
+	[[nodiscard]] constexpr bool isOpen() const
+	{
+		assert((IsChest() || isSarcophagus() || isDoor()) && "isOpen can only be called on chests, sarcophagus and doors");
+
+		if (IsChest() || isSarcophagus())
+			return _oSelFlag == 0U;
+		if (isDoor())
+			return _oSelFlag == 2U;
```
AJenbo, line 211:
```suggestion
			return door._oVar4 == DOOR_OPEN;
```

ephphatha: Blocked doors are open as well :D

AJenbo: Fixed

sixcy:
Yeah, I can include it in the PR

AJenbo:
Well you pretty much have to as it's currently causing all the tests to fail :)

obligaron:
Note: I'm still in favor to a `killall` or `clear monsters` command 😉 
This would helps testing the multiplayer quests.

PR #4011, opened 2022-01-29
Replaced magic numbers with Direction conversions in missiles.cpp

Sergentfox234:
It did not change the function of the code, just made it slightly faster and more easily readable.

PR #5311, opened 2022-08-29
Refactor missiles.cpp to use Direction enum instead of magic Point/Displacement constants.

zaid-ismail031:
This PR supersedes PR #4011 which was opened by @Sergentfox234.  This is a solution for issue #2260. @AJenbo suggested a number of changes to PR #4011 on January 29th, and since these changes have not been implemented yet, I opted to do it myself. 

Besides implementing the suggested changes, I have also made an additional change. I've added a direction called `NoDirection` to `enum class Direction`, and added this direction to the Displacement switch statement in `displacement.hpp` so that displacement `{ 0, 0 }` can be labelled in the code. To prevent warnings during compilation, I've added a case for `Direction::NoDirection` to switch statements in `monster.cpp` and `automap.cpp` (the case does nothing, just breaks).

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -468,6 +468,8 @@ void DrawAutomapPlr(const Surface &out, const Displacement &myPlayerOffset, int 
 		DrawHorizontalLine(out, point, AmLine(8) + 1, playerColor);
 		DrawMapLineSteepSE(out, point, AmLine(4), playerColor);
 	} break;
+	case Direction::NoDirection:
+	break;
 	}
 }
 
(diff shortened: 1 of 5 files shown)


AJenbo (approved):

AJenbo:
Please apply clang-format to missiles.cpp

AJenbo (approved):

zaid-ismail031:
@AJenbo Apologies for the formatting error in automap.cpp... did not notice it in the error logs first time round.

AJenbo:
Well it highlighted an issue with the github config which has not been fixed to prevent things being merged before all issues are fixed :P

PR #5312, opened 2022-08-30
CMake: Always strip the Release binary

glebm:


Trihedraf:
Will that include the MinSizeRel as well?

glebm:
I think it won't but the only platform where we use MinSizeRel (rg99) already does the stripping in the shell script. Not sure how to adjust the genex so that it applied to MinSizeRel as well.

AJenbo (approved):

PR #5316, opened 2022-08-30
Handle multiline strings in PrintInfo

sixcy:
Currently, using multiline strings in InfoString results in the text not
being centered, in addition to strings with 3 lines and more being just
cut even though there should be enough space in the box.

This patch counts the newlines in the InfoString to adjust accordingly
the text box: text with 2 or 3 lines can now be printed correctly.

Here are a few pictures in my mod that shows the difference:

2-line string, before:
twolines-before

2-line string, with this patch:
twolines-after

3-line string, before:
threelines-before

3-line string, after:
threelines-after

I believe this change could benefit modders by allowing them to print more text in the InfoString if they want to.
It could also benefit translations as well: some languages require more words to say the same thing. Perhaps some of the existing translations could be updated to benefit this new multiline string support.

Right now I multiply by 6 the number of newlines and that's the offset I apply to the text box. I pulled this magic number out of my hat, tried different numbers until it looked alright. I am open to have a more precise method if someone can think of one.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -256,6 +256,16 @@ void SetButtonStateDown(int btnId)
 	panbtndown = true;
 }
 
+int CountNewlines(std::string_view str)
+{
+	int count = 0;
+	for (char c : str) {
+		if (c == '\n')
+			count++;
+	}
+	return count;
+}
+
 void PrintInfo(const Surface &out)
 {
 	if (talkflag)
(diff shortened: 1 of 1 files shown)


AJenbo:
Could you also correct the line height so it matches what you would hey if only adding 2/3/4 lines?

sixcy:
> Could you also correct the line height so it matches what you would hey if only adding 2/3/4 lines?

You mean have it so that a string of two lines is displayed the same as two string of one line ?

I would have to take a look

Source/control.cpp
```diff
@@ -256,6 +256,16 @@ void SetButtonStateDown(int btnId)
 	panbtndown = true;
 }
 
+int CountNewlines(std::string_view str)
```
glebm, line 259: No need for this function, you can simply do:

```c++
std::count(str.begin(), str.end(), '\n')
```

sixcy: Thanks, I did not know about that function.

StephenCWills:
I feel like I'm missing something. Is there a reason you are assigning a multiline string to `InfoString` instead of calling `AddPanelString()` for each line you want to add?

sixcy:
> I feel like I'm missing something. Is there a reason you are assigning a multiline string to `InfoString` instead of calling `AddPanelString()` for each line you want to add?

I was not aware of the AddPanelString solution I will look into that.

AJenbo:
> I feel like I'm missing something. Is there a reason you are assigning a multiline string to `InfoString` instead of calling `AddPanelString()` for each line you want to add?

The idea here is to replace `AddPanelString()` with multi line string support to provide better texts for translators. I'm hoping we can drop `AddPanelString()` entirely.

StephenCWills:
With the merge of #5369, we now have support for multiline messages in `InfoString` so I think it's appropriate to close this one.

sixcy:
Thanks for working on this, cheers

PR #5317, opened 2022-08-30
[Documentation] Add instructions how to add "patch" to PATH for Windows

tsunamistate:
Since a430d99f2401b27416c85907a9ab3cc431833f3f `patch` utility is required to build, and it isn't in PATH when using Windows MSVC.
Add instructions on where to find it and how to add it to PATH


glebm:
Thanks, I guess we should use `find_program` in CMake and look there to make it easier for users

glebm:
https://github.com/diasurgical/devilutionX/pull/5320

AJenbo:
Fix by #5320 (please test)

PR #5320, opened 2022-08-31
CMake: Use FindPatch

glebm:
CMake's built-in FindPatch will look in the Git installation directory on Windows, eliminating the need for Windows users to modify PATH.

See https://github.com/Kitware/CMake/blob/master/Modules/FindPatch.cmake

### DIFF
--- a/3rdParty/discord/CMakeLists.txt
+++ b/3rdParty/discord/CMakeLists.txt
@@ -1,10 +1,12 @@
 include(functions/FetchContent_MakeAvailableExcludeFromAll)
 include(FetchContent)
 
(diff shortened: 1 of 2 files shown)


AJenbo (approved):

AJenbo:
Wonder what GitHub will do since I already applied the formatting fix :thinking:

AJenbo:
Does this replace #5317 ?

glebm:
Yeah, #5317 shouldn't be needed with this

tsunamistate:
Fix confirmed, CMake finds the `patch` utility without modifying PATH on Windows MSVC

PR #5321, opened 2022-08-31
Fix single room cathedral with quest piece

### LINKED ISSUE #5299: Broken level (Leorics tomb)
### Operating System

Linux x64

### DevilutionX version

1.4.1 (latest release)

### Describe

Started a new game on nightmare difficulty.
Played to level 3 and saw something like this.
leoric

I can't judge if it's a vanilla bug, for example a broken seed or something similar.
A vanilla bug would of course be better, otherwise there might be an undesirable issue in our beautiful level generation code.

### To Reproduce

Rare issue. Unclear whether reproducible.

### Expected Behavior

No broken level

### Additional context

Savegame:
[leoric.zip](https://github.com/diasurgical/devilutionX/files/9439350/leoric.zip)

EDIT: When walking around the tomb there is a "mirroring effect".

AJenbo:
Fixes #5299

SelectChamber was not handling the case where only chamber 2 is available.


AJenbo:
Not sure why tests are failing on CI, they work locally :/

test/drlg_l1_test.cpp
```diff
@@ -148,6 +148,20 @@ TEST(Drlg_l1, CreateL5Dungeon_hellfire_2_1180526547)
 	EXPECT_EQ(ViewPosition, Point(81, 45));
 }
 
+TEST(Drlg_l1, CreateL5Dungeon_hellfire_3_1369955278)
+{
+	LoadExpectedLevelData("hellfire/3-1369955278.dun");
+
```
glebm, line 154: Players.resize(1);
MyPlayer = &Players[0];

AJenbo: Arh implemented the test in a version before the breakage and this change and didn't think to look at the other test after rebasing :) Just got stomped when my local tests where passing...

Source/levels/drlg_l1.cpp
```diff
@@ -1258,14 +1258,14 @@ void PlaceMiniSetRandom(const Miniset &miniset, int rndper)
...
 		chamber = PickRandomlyAmong({ 2, 3 });
-	else if (!HasChamber2)
+	else if (HasChamber1 && !HasChamber2 && HasChamber3)
 		chamber = PickRandomlyAmong({ 3, 1 });
-	else if (!HasChamber3)
+	else if (HasChamber1 && HasChamber2 && !HasChamber3)
 		chamber = PickRandomlyAmong({ 2, 1 });
-	else
+	else if (HasChamber1 && HasChamber2 && HasChamber3)
 		chamber = GenerateRnd(3) + 1;
```
ephphatha, line 1269: How about reordering the tests so they're positive checks? That way we can explain the exceptional single-chamber case.

```cpp
int chamber;
if (HasChamber1 && HasChamber2 && HasChamber3) {
    chamber = GenerateRnd(3) + 1;
} else if (HasChamber1 && HasChamber2) {
    chamber = PickRandomlyAmong({ 2, 1 });
} else if (HasChamber1 && HasChamber3) {
    chamber = PickRandomlyAmong({ 3, 1 });
} else if (HasChamber2 && HasChamber3) {
    // Reversed list order compared to the previous 2 cases in order to match vanilla logic
    chamber = PickRandomlyAmong({ 2, 3 });
} else {
    // The dungeon generation logic ensures that chamber 2 is available if
    // either (or both of) 1 or 3 aren't, so if we ever end up with a single
    // chamber layout it's always chamber 2.
    chamber = 2;
}

File: Source/levels/drlg_l1.cpp [hunk 3]
```diff
@@ -1259,14 +1259,20 @@ void PlaceMiniSetRandom(const Miniset &miniset, int rndper)
 Point SelectChamber()
 {
 	int chamber;
-	if (!HasChamber1)
+	if (HasChamber1 && HasChamber2 && HasChamber3) {
+		chamber = PickRandomlyAmong({ 1, 2, 3 });
```
ephphatha, line 1263: This is not quite the same as `GenerateRnd(3) + 1`. It's possible for the former code to result in `-2 + 1`/`-1` which effectively acts the same as picking chamber 2. This code picks chamber 1 instead since PickRandomlyAmong corrects the original bias for the buggy case where GenerateRnd returns a negative value.

obligaron:
Note:
I also found a seed that generates different levels between master and the previous release (hellfire, dungeon level 2, seed 2380102989 with poisoned water quest).
With this pr this seed is fixed. 👍

Source/levels/drlg_l1.cpp
AJenbo:
```suggestion
		chamber = GenerateRnd(3) + 1;
```

Source/levels/drlg_l1.cpp
```diff
@@ -1259,14 +1259,20 @@ void PlaceMiniSetRandom(const Miniset &miniset, int rndper)
...
-		chamber = PickRandomlyAmong({ 3, 1 });
-	else if (!HasChamber3)
-		chamber = PickRandomlyAmong({ 2, 1 });
-	else
+	if (HasChamber1 && HasChamber2 && HasChamber3) {
 		chamber = GenerateRnd(3) + 1;
+	} else if (HasChamber1 && HasChamber2) {
+		chamber = PickRandomlyAmong({ 2, 1 }); // Reverse odrder to match vanilla
+	} else if (HasChamber1 && HasChamber3) {
+		chamber = PickRandomlyAmong({ 3, 1 }); // Reverse odrder to match vanilla
```
ephphatha, line 1267:
```suggestion
		chamber = PickRandomlyAmong({ 2, 1 }); // Reverse order to match vanilla
	} else if (HasChamber1 && HasChamber3) {
		chamber = PickRandomlyAmong({ 3, 1 }); // Reverse order to match vanilla
```

ephphatha (approved):
I think it's possible to prove whether or not the three chamber case can ever be hit in a way that the difference between `GenerateRnd(3) + 1` and `PickRandomlyAmong({1, 2, 3})` matters, but that's gonna take a bit to work out the exact set of RNG calls leading up to it :D

AJenbo:
I think this will do for now :)

PR #5325, opened 2022-09-04
Introduce AnimationInfo::isLastFrame and use it instead of _pXYZFrames

### LINKED ISSUE #5046: Equipment short cuts causes animation desync
### Operating System

Windows x64

### DevilutionX version

1.4.0 (latest release)

### Describe

When switching or changing equipment using ctrl+click or shift+click it can cause the players to appear to be walking in place on the other end.

See the following video around the 25min mark
https://www.twitch.tv/videos/1534790732

### To Reproduce

ctrl+click shield during multiplayer

### Expected Behavior

Player should continue to animate as normal.

### Additional context

_No response_

obligaron:
Fixes #5046

The cause of #5046 was that the `AnimationInfo` and player mode got out of sync.
This is mostly not problematic, but in some actions (for example `DoAttack`), we don't use `AnimationInfo::numberOfFrames` to check if an animation ends but instead the original value (for example `_pAFrames`). This can result in problems (see #5046).

With this pr a new helper (`AnimationInfo::isLastFrame`) is introduced to check for the last frame (inspired by #4818).
And this helper is used instead of `_pXYZFrames` (first commit; fixes the issue) and where we previous used `numberOfFrames` (second commit; makes the code more readable).

This pr makes the game more stable and is additional to #5324 (Both pr's fixes #5046 but #5324 is the cleaner way).

### DIFF
--- a/Source/engine/animationinfo.h
+++ b/Source/engine/animationinfo.h
@@ -66,6 +66,11 @@ public:
 		return (*sprites)[getFrameToUseForRendering()];
 	}
 
+	[[nodiscard]] bool isLastFrame() const
+	{
+		return currentFrame >= (numberOfFrames - 1);
+	}
+
 	/**
 	 * @brief Calculates the Frame to use for the Animation rendering
 	 * @return The Frame to use for rendering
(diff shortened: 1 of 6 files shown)


glebm (approved):
Nice!

PR #5326, opened 2022-09-04
Decode min-data on load

AJenbo:
I guess this is going to grow the memory usage a bit. But I think it improves readability, and it gets rid of a global.

Let me know if you have some suggestions.

Maybe not hardcode MICROS to 16, or DPieceMicros to MAXTILES?

### DIFF
--- a/Source/levels/gendung.h
+++ b/Source/levels/gendung.h
@@ -114,8 +114,79 @@ struct MegaTile {
 	uint16_t micro4;
 };
 
+/**
+ * Tile type.
+ *
+ * The tile type determines data encoding and the shape.
+ *
+ * Each tile type has its own encoding but they all encode data in the order
+ * of bottom-to-top (bottom row first).
+ */
+enum class TileType : uint8_t {
+	/**
+	 * 🮆 A 32x32 square. Stored as an array of pixels.
+	 */
+	Square,
+
+	/**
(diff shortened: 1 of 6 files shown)


Source/levels/gendung.h
```diff
@@ -114,8 +114,79 @@ struct MegaTile {
...
+	 * 🭞 Right-pointing 32x32 trapezoid: 32x16 rectangle and the 16x16 bottom part of `RightTriangle`.
+	 *
+	 * Begins with the triangle part, which uses the `RightTriangle` encoding,
+	 * and is followed by a flat array of pixels for the top rectangle part.
+	 */
+	RightTrapezoid,
+};
+
+struct MicroTile {
+	TileType type;
```
glebm, line 184: How about making this a `class` with a single `uint_16t` member and `type()` and `id()` methods?
It'll have all the readability benefits without memory usage increase.

glebm: I did this in https://github.com/diasurgical/devilutionX/commit/69c429bb683b6de54d8848066def7f83a97f54a1

glebm:
I guess we can close this now that we have the `LevelCelBlock` class that does the decoding on the fly?

AJenbo:
Ok 🙂

PR #5327, opened 2022-09-04
Synchronize ears and gold in player inventories

StephenCWills:
Continuation of #5217

* Introduce a struct to differentiate ears from other items in the network code
* Expand `TCmdChItem` to include the same level of item information as `TCmdGItem` and `TCmdPItem`
* Send network messages when adding gold to and removing gold from the player's inventory
* Fix an error that used the wrong inventory grid index when a remote player swaps items

While testing, I noticed that moving ears around in a player's inventory will cause remote clients to crash due to the inventory synchronization. This PR should fix that as well.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1848,25 +1848,48 @@ int SyncDropItem(Point position, int idx, uint16_t icreateinfo, int iseed, int i
 
 	dItem[position.x][position.y] = ii + 1;
 
-	if (idx == IDI_EAR) {
-		RecreateEar(item, icreateinfo, iseed, id, dur, mdur, ch, mch, ivalue, ibuff);
-	} else {
-		RecreateItem(*MyPlayer, item, idx, icreateinfo, iseed, ivalue, (ibuff & CF_HELLFIRE) != 0);
-		if (id != 0)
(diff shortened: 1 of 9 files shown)


Source/inv.cpp
```diff
@@ -1848,25 +1830,48 @@ int SyncDropItem(Point position, int idx, uint16_t icreateinfo, int iseed, int i
...
+	item.dwBuff = ibuff;
+	item.position = position;
+	RespawnItem(item, true);
+
+	if (currlevel == 21 && position == CornerStone.position) {
+		CornerStone.item = item;
+		InitQTextMsg(TEXT_CORNSTN);
+		Quests[Q_CORNSTN]._qlog = false;
+		Quests[Q_CORNSTN]._qactive = QUEST_DONE;
 	}
```
AJenbo, line 1855: Considering that the corner stone was never intended/tested for multiplayer and causes morphing of items transported between single player and multiplayer. I was thinking that we should simply disable it during multiplayer.

I guess this doesn't necessarily preclude still having the room and allowing for completing the quest. So maybe this comment is just irrelevant :)

AJenbo (approved):

PR #5337, opened 2022-09-10
Replace toHitSpecial member with a function & fix wrong arguments being passed

aetn23:
Replace toHitSpecial member with a function and fix wrong arguments being passed in `MonsterAttackEnemy`.
The replacement is done to reduce `Monster` in size and to make the code more readable, by having this value being calculated solely in one place. The latter is to fix a bug, where wrong values were being passed to a function.


Source/monster.h
```diff
@@ -31,6 +31,9 @@ struct Player;
 constexpr size_t MaxMonsters = 200;
 constexpr size_t MaxLvlMTypes = 24;
 
+constexpr int NightmareToHitBonus = 85;
+constexpr int HellToHitBonus = 120;
```
StephenCWills, line 35: These were previously defined in the anonymous namespace because they're not used outside of monster.cpp. It would make more sense to define `toHitSpecial()` in the cpp file rather than moving the variables to the header.

StephenCWills (changes requested):

Source/monster.h
aetn23: Yes, it makes more sense. Moved `toHitSpecial` implementations to the .cpp file along with globals

StephenCWills (dismissed):
Just noting that the to-hit bonus constants were moved below the AC bonus constants seemingly by accident. I don't think it's a big deal so I'm approving.

Source/monster.cpp
```diff
@@ -60,12 +60,12 @@ bool sgbSaveSoundOn;
...
 
-constexpr int NightmareToHitBonus = 85;
-constexpr int HellToHitBonus = 120;
-
 constexpr int NightmareAcBonus = 50;
 constexpr int HellAcBonus = 80;
 
+constexpr int NightmareToHitBonus = 85;
+constexpr int HellToHitBonus = 120;
+
```
AJenbo, line 68:
```suggestion
constexpr int NightmareToHitBonus = 85;
constexpr int HellToHitBonus = 120;

constexpr int NightmareAcBonus = 50;
constexpr int HellAcBonus = 80;

```

AJenbo (approved):

PR #5339, opened 2022-09-11
Implement town map

AJenbo:
It's not uncommon for new players to get a bit lost in town, to alleviate that a bit I have implemented a map for the town.

image
(shown here with all entrances open, only the church is shown at the start)

Since the town dun files are over sized and contains multipl hidden non nonsensical parts, a custom .dun needed to be created for the map portion. Since the town has over 255 tiles, i decided to instead base the automap on the crypt tile set to avoid having to expand the memory allocated by `uint8_t dungeon[DMAXX][DMAXY];` and `std::array<AutomapTile, 256> AutomapTypeTiles;
`, so instead it is actually based on the Cave tileset:

Behold Ünder Tristram:
image


Source/automap.cpp
```diff
@@ -504,23 +736,24 @@ void DrawAutomapText(const Surface &out)
...
+	std::string description;
+	switch (leveltype) {
+	case DTYPE_NEST:
+		description = fmt::format(fmt::runtime(_("Level: Nest {:d}")), currlevel - 16);
+		break;
+	case DTYPE_CRYPT:
+		description = fmt::format(fmt::runtime(_("Level: Crypt {:d}")), currlevel - 20);
+		break;
+	case DTYPE_TOWN:
+		description = std::string(_("Town"));
```
burningserenity, line 748: What if this was instead called "Tristram?" More specific. OTOH, it could be argued that all of these places under Tristram are also in Tristram.

glebm (approved):
Nice!

Source/automap.cpp
AJenbo: ill consider it for a follow up

Source/automap.cpp
```diff
@@ -340,6 +536,42 @@ void DrawAutomapTile(const Surface &out, Point center, Point map)
 	case AutomapTile::Types::Corner:
 	case AutomapTile::Types::None:
 		break;
+	case AutomapTile::Types::Bridg:
```
Iniquitatis, line 539: Bridge?

qndel: orly?

PR #5345, opened 2022-09-14
refactored GetItemSpace

kpyrkosz:
This patch refactors the function. The behavior is exactly the same, without the messy loops. As a bonus, one global is removed.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -523,7 +523,7 @@ bool GetItemSpace(Point position, int8_t inum)
 		}
 	}
 
-	int rs = GenerateRnd(15) + 1;
+	int rs = GenerateRnd(15);
 
 	if (candidate_count == 0)
 		return false;
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -147,8 +147,6 @@ Item curruitem;
 /** Holds item get records, tracking items being recently looted. This is in an effort to prevent items being picked up more than once. */
 ItemGetRecordStruct itemrecord[MAXITEMS];
 
-bool itemhold[3][3];
-
 /** Specifies the number of active item get records. */
 int gnNumGetRecords;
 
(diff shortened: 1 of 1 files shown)


AJenbo (dismissed):
Nice clean up 🙂

Source/items.cpp
```diff
@@ -516,50 +514,23 @@ void CalcSelfItems(Player &player)
...
-		yy++;
-		if (yy == 3)
-			yy = 0;
-	}
-
-	xx += position.x - 1;
-	yy += position.y - 1;
-	Items[inum].position = { xx, yy };
-	dItem[xx][yy] = inum + 1;
+	Point &destination_xy = xy_candidates[rs % candidate_count];
```
ephphatha, line 531: Removing the loop changes the way `rs` is used in a way that requires removing the +1 for correct results. At this point the expected range for rs is currently 1-15, since the range doesn't start at 0 the first potential position is considered less frequently than it was previously.

Source/items.cpp
```diff
@@ -516,50 +514,23 @@ void CalcSelfItems(Player &player)
...
-	for (int j = 0; j < 3; j++) {
-		for (int i = 0; i < 3; i++) { // NOLINT(modernize-loop-convert)
-			if (itemhold[i][j])
-				savail = true;
+			if (ItemSpaceOk({ i, j }))
+				xy_candidates[candidate_count++] = { i, j };
 		}
 	}
 
 	int rs = GenerateRnd(15) + 1;
```
ephphatha, line 526: It's possible for this to initialise rs to -7. The old code would set the item position to the top left space unconditionally, the new code tries to access out-of-bounds in xy_candidates on platforms with signed modulo arithmetic.

Source/items.cpp
ephphatha: This is also probably why the tests are failing, items are being dropped in the next candidate space in the search list than when the demo was recorded.

ephphatha (changes requested):

Source/items.cpp
```diff
@@ -516,50 +514,23 @@ void CalcSelfItems(Player &player)
 
 bool GetItemSpace(Point position, int8_t inum)
 {
-	int xx = 0;
-	int yy = 0;
+	Point xy_candidates[9];
+	int candidate_count = 0;
 	for (int j = position.y - 1; j <= position.y + 1; j++) {
-		xx = 0;
 		for (int i = position.x - 1; i <= position.x + 1; i++) {
```
ephphatha, line 520: While you're at it this can use the PointInRectangleRangeIterator :D

kpyrkosz:
The +1 was causing the issue. This version should return the same point as original version.

AJenbo:
The issue here is that it fails to connect two doors, walking in the opposite direction, hitting a wall and not turning around.

beginning 22x10, end 10x12, dir: 3
```
                                         
                                         
                                         
                                         
                                         
                                         
                                         
                                 C##E    
                         C###E   #..#    
             C#####E     #...#   #..#    
            ,D.....#     #...#   #..#    
             #.....#     #...#   B##A    
             #.....#     #...#           
             #.....#     #...#           
             B##D##A     #...#           
               ,,,       B###A           
               ,,,                       
               ,,,     C#######E         
               ,,,     #.......#         
               ,,,     #.......#         
               ,,,     #.......#         
       C###E   ,,,     B#######A         
       #...D   CDE                       
       #...#   #.#                       
       #...#   #.#                       
       B###A   #.#         C#######E     
               #.#         #.......#     
               BDA         #.......#     
                ,          #.......#     
               ,,,         #.......#     
                ,          #.......#     
            C###D####E     #.......#     
            #........#     #.......#     
      C###E #........#     #.......#     
      #...# #........#     B#######A     
      #...# #........#                   
      #...# #........#                   
      B###A #........#                   
            B########A                   
                                         
```

After 1600 iterations:
```
                                         
                                         
                                         
                                         
                                         
                                         
                                         
                                 C##E    
                         C###E   #..#    
             C#####E     #...#   #..#    
            ,D.....#     #...#   #..#    
             #.....#     #...#   B##A    
             #.....#     #...#           
             #.....#     #...#           
             B##D##A     #...#           
               ,,,       B###A           
               ,,,                       
               ,,,     C#######E         
               ,,,     #.......#         
               ,,,     #.......#         
               ,,,     #.......#         
       C###E   ,,,     B#######A         
       #...D,  CDE                       
       #...#,  #.#                       
       #...#,  #.#                       
       B###A,  #.#         C#######E     
            ,  #.#         #.......#     
            ,  BDA         #.......#     
            ,   ,          #.......#     
            ,  ,,,         #.......#     
            ,   ,          #.......#     
            C###D####E     #.......#     
            #........#     #.......#     
      C###E #........#     #.......#     
      #...# #........#     B#######A     
      #...# #........#                   
      #...# #........#                   
      B###A #........#                   
            B########A                   
                               
```

Source/items.cpp
AJenbo: Should we wrap it in `abs()`?

ephphatha: Calling std::max to clamp to 0 would be better, could even change the limit used by generaternd since it won't impact dungeon generation if someone is willing to re-record the demo

PR #5350, opened 2022-09-16
Spellbook text revision

kphoenix137:
Visual explanation:

image
image
image
image
image
image

Text explanation:
This changes the spellbook to be more uniform with the character panel. In the character panel, red is used to signify when stats are lower than base, and blue is used to signify when stats are higher than base. Also gold is used when a stat is maxed out. I used this method for the spellbook, as well and showing exactly how many spell levels are being gained or lost from items.

On top of that, I revised how spell information is shown for skill and staves. In vanilla, you were given no information other than the fact a skill is a skill, or how many charges a staff has remaining. This gives the player access to that missing information.

### DIFF
--- a/Source/panels/spell_book.cpp
+++ b/Source/panels/spell_book.cpp
@@ -132,19 +132,15 @@ void DrawSpellBook(const Surface &out)
 
 	int yp = 12;
 	const int textPaddingTop = 7;
-
 	for (int i = 1; i < 8; i++) {
-<<<<<<< HEAD
 		SpellID sn = SpellPages[sbooktab][i - 1];
-=======
-		spell_id sn = SpellPages[sbooktab][i - 1];
 
 >>>>>>> d87240e2d (Revise Spell Book Colors and Text)
 		if (IsValidSpell(sn) && (spl & GetSpellBitmask(sn)) != 0) {
 			SpellType st = GetSBookTrans(sn, true);
 			SetSpellTrans(st);
 			const Point spellCellPosition = GetPanelPosition(UiPanels::Spell, { 11, yp + SpellBookDescription.height });
-			DrawSpellCel(out, spellCellPosition, *pSBkIconCels, SpellITbl[sn]);
+			DrawSmallSpellIcon(out, spellCellPosition, sn);
 
 			bool isSkill = false;
 			bool isStaff = false;
--- a/Source/panels/spell_book.cpp
+++ b/Source/panels/spell_book.cpp
@@ -18,6 +18,10 @@
 #include "spelldat.h"
 #include "utils/language.h"
 #include "utils/stdcompat/optional.hpp"
+#include "utils/str_cat.hpp"
+#include <vector>
+
+#define SPLICONLAST (gbIsHellfire ? 51 : 42)
 
 namespace devilution {
 
@@ -45,7 +49,7 @@ void PrintSBookStr(const Surface &out, Point position, string_view text, UiFlags
 	    Rectangle(GetPanelPosition(UiPanels::Spell, position + Displacement { SPLICONLENGTH, 0 }),
 	        SpellBookDescription)
 	        .inset({ SpellBookDescriptionPaddingHorizontal, 0 }),
-	    UiFlags::ColorWhite | flags);
+	    flags);
 }
 
 SpellType GetSBookTrans(SpellID ii, bool townok)
@@ -109,6 +113,7 @@ void FreeSpellBook()
 void DrawSpellBook(const Surface &out)
 {
 	ClxDraw(out, GetPanelPosition(UiPanels::Spell, { 0, 351 }), (*pSpellBkCel)[0]);
+
 	if (gbIsHellfire && sbooktab < 5) {
 		ClxDraw(out, GetPanelPosition(UiPanels::Spell, { 61 * sbooktab + 7, 348 }), (*pSBkBtnCel)[sbooktab]);
 	} else {
@@ -119,6 +124,7 @@ void DrawSpellBook(const Surface &out)
 		}
 		ClxDraw(out, GetPanelPosition(UiPanels::Spell, { sx, 348 }), (*pSBkBtnCel)[sbooktab]);
 	}
+
 	Player &player = *MyPlayer;
 	uint64_t spl = player._pMemSpells | player._pISpells | player._pAblSpells;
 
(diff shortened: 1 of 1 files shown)


kphoenix137:
If this concept is acceptable for DevilutionX, should I remove the switch-case and consolidate all the logic and use if-else statements to decipher between skill/spell/staff? Any suggestions?

StephenCWills:
The concept seems pretty harmless to me. You don't necessarily have to remove the switch-case, but I would definitely recommend deduplicating much of that logic.

kphoenix137:
> The concept seems pretty harmless to me. You don't necessarily have to remove the switch-case, but I would definitely recommend deduplicating much of that logic.

I'll see what I can do

ChaosMarc:
What do you think of using gold as a color for skill levels above maximum? In your screenshots there is no way of knowing that the lvl18 skills are maxed out.

kphoenix137:
> What do you think of using gold as a color for skill levels above maximum? In your screenshots there is no way of knowing that the lvl18 skills are maxed out. 

With this logic we would want to display gold if the spell is maxed regardless if spell levels are being removed or added from items and wouldn't have a consistent usage of blue and red text. I can't select specific characters from a string to make a certain color so just making the first number gold and second number another color would be very difficult for spacing correctly. May be better to just make all the text for a maxed out spell gold and keep the blue and red for the colors when the spell level is being increased or decreased.

qndel:
why not keep the 18 gold and (+3) blue?

kphoenix137:
> why not keep the 18 gold and (+3) blue?

If you can help me figure out a way to code that and maintain correct text spacing then I'll do that

FitzRoyX:
This doesn't work because there is no "base" and "now" column like in the character panel. 18 (+3) can be read as either 18 or 21, and being all blue while something lower is gold is awkward.

ChaosMarc:
What about stopping at 15? So instead of 18 (+3) we print 15 (+3). 15 in gold and (+3) in blue

kphoenix137:
> This doesn't work because there is no "base" and "now" column like in the character panel. 18 (+3) can be read as either 18 or 21, and being all blue while something lower is gold is awkward.

I think it may work better to display entire spell text in gold if the spell is maxed, and just use blue or gold for the spell level number if it's adjusted without extra text showing the added levels

StephenCWills:
> This doesn't work because there is no "base" and "now" column like in the character panel. 18 (+3) can be read as either 18 or 21...

I think this is a non-issue. If you're confused, you can just dequip your gear. Once you see how it works, you won't be confused anymore.

kphoenix137:
I will work on a revision of what I have so far then post screenshots tonight for re-evaluation. I have some more ideas

kphoenix137:
image
Opinions?

qndel:
How about display the spell level without bonuses on the icon?

kphoenix137:
Sure

kphoenix137:
> 

image

kphoenix137:
Pushed changes

kphoenix137:
> How about display the spell level without bonuses on the icon? Otherwise you are just duplicating the level already displayed on the right side

Would it be more logical to display actual level over the icon and display base level with added level on the right instead?

kphoenix137:
image
Yeah I think this makes way more sense

kphoenix137:
image
Only maximums are shown as gold (20 on the left, 15 on the right)

yuripourre:
@kphoenix137 loved it!

kphoenix137:
Removed gold text. Icon numbers show up as white unless increased by items which will be shown as blue. Text on the right shows original spell level, along with the added levels next to it

ChaosMarc:
Why did you remove the gold text again? I think it's a useful information to display and really liked the implementation from your post before the last. Besides that the latest version Looks awsome

kphoenix137:
> Why did you remove the gold text again? I think it's a useful information to display and really liked the implementation from your post before the last. Besides that the latest version Looks awsome

I found it's too confusing to use since there are 2 actual different maximums (base level maximum and added spell maximum) and showing different numbers as gold can add a layer of extra confusion. I think the final iteration is the most straightforward and adds the much needed extra information for the player.

StephenCWills:
IMO, in the same way as the character screen doesn't care about the theoretical maximum with gear, the spellbook probably shouldn't either. There is one maximum, and that is slvl 15 without bonuses from gear. I personally think it would be better if we use gold text for it as that shouldn't be inconsistent with highlighting the base stat on the char panel when it reaches maximum.

kphoenix137:
> IMO, in the same way as the character screen doesn't care about the theoretical maximum with gear, the spellbook probably shouldn't either. There is one maximum, and that is slvl 15 without bonuses from gear. I personally think it would be better if we use gold text for it as that shouldn't be inconsistent with highlighting the base stat on the char panel when it reaches maximum.

image

kphoenix137:
New revision:
No more spell icon text. Level on right is base level (will show as gold for max spell level), and number in parentheses is current level (only displayed if spell level is adjusted by items).

image
image

kphoenix137:
Added damage reduction text for Mana Shield
image

kphoenix137:
Ready for review

Source/panels/spell_book.cpp
```diff
@@ -121,62 +122,94 @@ void DrawSpellBook(const Surface &out)
...
+			if (splLvlAdd > 0)
+				splLvlAddColor = UiFlags::ColorBlue;
+			else if (splLvlAdd < 0)
+				splLvlAddColor = UiFlags::ColorRed;
+
+			if (IsNoneOf(sn, SPL_REPAIR, SPL_RECHARGE, SPL_DISARM, SPL_IDENTIFY, SPL_BLODBOIL)) {
+				if (splLvlAdd != 0) {
+					PrintSBookStr(out, line0, fmt::format(fmt::runtime(pgettext(/* TRANSLATORS: UI constraints, keep short please.*/ "spellbook", "Level: {:d} ({:d})")), baseLvl, lvl), UiFlags::AlignRight | UiFlags::ColorWhite);
+					PrintSBookStr(out, line0, fmt::format(fmt::runtime(pgettext(/* TRANSLATORS: UI constraints, keep short please.*/ "spellbook", "{:d} ({:d})")), baseLvl, lvl), UiFlags::AlignRight | maxSplLvlColor);
+					PrintSBookStr(out, line0, fmt::format(fmt::runtime(pgettext(/* TRANSLATORS: UI constraints, keep short please.*/ "spellbook", "({:d})")), lvl), UiFlags::AlignRight | splLvlAddColor);
```
StephenCWills, line 176: Have you seen these functions?

https://github.com/diasurgical/devilutionX/blob/68d57bd2d742dd46799114dfb8a89a5888514837/Source/engine/render/text_render.hpp#L209-L229

kphoenix137: I haven't! Thank you. Will make another commit tomorrow night

ChaosMarc:
I have a question about the skill level of charges. Not quite sure about the actual mechanisms behind this but how can a have 6 charges for lvl0 mana shield? Do charges use your current skill level? What if you don't know the skill without the item? Do items that reduce all skill levels by x affect staffs?

kphoenix137:
> I have a question about the skill level of charges. Not quite sure about the actual mechanisms behind this but how can a have 6 charges for lvl0 mana shield? Do charges use your current skill level? What if you don't know the skill without the item? Do items that reduce all skill levels by x affect staffs? 

Staves use whatever spell level the player currently has. If the spell is not known, the staff casts at spell level 0.

kphoenix137:
> I have a question about the skill level of charges. Not quite sure about the actual mechanisms behind this but how can a have 6 charges for lvl0 mana shield? Do charges use your current skill level? What if you don't know the skill without the item? Do items that reduce all skill levels by x affect staffs? 

Also to get more in depth, the only reason level 0 spells don't work when cast as normal spells is because there is intentionally design to stop you from doing so. Spells still function at spell level 0, which we've been using for many years but most of us were under the impression staff spells without knowing the spell casts at level 1 which has been recently debunked.

ChaosMarc:
Oh wow. I didn't know that. Thanks for the explanation

Source/panels/spell_book.cpp
StephenCWills: Do you need some help with this?

kphoenix137:
not sure why tests are failing on amiaos

ikonomov:
Great idea, I think keeping the base value to white or gold and then blue or red for the bonus is the most logical way to do it.

kphoenix137:
I have a new approach that I think would be a hell of a lot better than how the spellbook currently works, which would make this PR obsolete as well as https://github.com/diasurgical/devilutionX/pull/5366

PR #5357, opened 2022-09-18
Gamepad icons

burningserenity:
Adds icons for gamepad buttons. Used icons from [greatdocbrown](https://greatdocbrown.itch.io/gamepad-ui) Includes icons for Xbox, Nintendo Switch, and PlayStation Dual Shock control schemes.
Constants are defined for each icon for easy reference in code.
Currently, icons are only displayed when showing info text for items.
Companion PR: [!8](https://github.com/diasurgical/devilutionx-assets/pull/8)

PlayStation:
Screenshot_2022-09-18_13:27:15
Screenshot_2022-09-18_13:53:32

Nintendo:
Screenshot_2022-09-18_13:28:25
Screenshot_2022-09-18_13:28:33

Xbox:
Screenshot_2022-09-18_13:29:32
Screenshot_2022-09-18_13:29:37


Source/items.cpp
```diff
@@ -1740,66 +1740,65 @@ void DrawUniqueInfoWindow(const Surface &out)
 void printItemMiscKBM(const Item &item, const bool isOil, const bool isCastOnTarget)
 {
 	if (item._iMiscId == IMISC_MAPOFDOOM) {
-		AddPanelString(_("Right-click to view"));
+		InfoString = StrCat(item._iName, "\n", _("Right-click to view"));
```
StephenCWills, line 1743: What is the purpose of replacing calls to `AddPanelString()` with a multiline `InfoString`? Seems like you're just inviting potential for OOB in `PrintInfo()` by bypassing the validation in `AddPanelString()`. Furthermore, you've changed a few strings that have likely already been translated into many languages so I'd like to understand the justification.

Source/controls/controller_buttons.h
```diff
@@ -36,4 +37,114 @@ inline bool IsDPadButton(ControllerButton button)
 	    || button == ControllerButton_BUTTON_DPAD_RIGHT;
 }
 
+namespace controllerButtonIcon {
+using namespace std;
```
StephenCWills, line 41: What is the purpose of `using namespace std;` here? Seems like you're referencing `std::string` explicitly anyway.

StephenCWills (changes requested):
Will you be submitting a PR for devilutionx-assets?
https://github.com/diasurgical/devilutionx-assets/tree/main/bundled-assets/fonts

Source/items.cpp
burningserenity: The multiline `InfoString` is to allow for multiline strings to be translated in one shot. Right now, for "Right-click to..., then" and "left-click to..." are sent to the translation function separately, which doesn't translate well. I could put the validations into PrintInfo, I could restrict this logic to just times when it's necessary to translate a multiline string, or both.

Source/controls/controller_buttons.h
burningserenity: There is no reason for it to be here.

burningserenity:
> Will you be submitting a PR for devilutionx-assets? https://github.com/diasurgical/devilutionx-assets/tree/main/bundled-assets/fonts

Yes I will. Does the pcx file go in both bundled-assets/fonts and assets/fonts?

StephenCWills:
No, just `bundled-assets/fonts`. That's for symbols that are included in DevilutionX releases (`devilutionx.mpq`).

The other folder is for `fonts.mpq` that provides extended symbols for CJK languages.

Source/items.cpp
StephenCWills: At a bare minimum, we should protect against the OOB. Honestly, though, it wasn't the only red flag here. The fact that you are introducing concatenation with the item name suggests to me that you are overwriting the text that was already assigned to `InfoString` by another part of the code.

I'd rather see a helper that handles multiline strings via `AddPanelString()`. You may also consider breaking this out into another PR in case it generates further discussion and blocks the merge.

StephenCWills: Actually, I just got a response from AJenbo on another PR so maybe my suggestion to use `AddPanelString()` is not the right approach. See https://github.com/diasurgical/devilutionX/pull/5316#issuecomment-1251655975

burningserenity: See [!5267](https://github.com/diasurgical/devilutionX/pull/5267)

burningserenity:
Awaiting merge of:
#5367 and #5369

AJenbo:
https://github.com/diasurgical/devilutionX/pull/5369 has now been merged :)

Source/items.cpp
StephenCWills: My last critique in regards to this is that maybe we should stick with `AddPanelString()` rather than replacing the entirety of `InfoString`. This will be especially important for oils, which have text inserted between the item name and the usage instruction.

image

burningserenity: My current code isn't working, so it needs to change in some way. What if I changed InfoString in PrintItemMiscOil as well, so that everything would still happen in the correct order? What do you think, @AJenbo ?

burningserenity: Yeah makes sense now that I can pass a multiline string to AddPanelString.

Source/control.cpp
```diff
@@ -47,6 +47,7 @@
 #include "utils/sdl_geometry.h"
 #include "utils/stdcompat/optional.hpp"
 #include "utils/str_cat.hpp"
+#include "utils/str_split.hpp"
```
StephenCWills, line 50: It looks like this include is no longer necessary.
```suggestion
```

Source/items.cpp
```diff
@@ -1764,27 +1763,27 @@ void printItemMiscVirtualGamepad(const Item &item, const bool isOil, bool isCast
...
 	} else if (GamepadType == GamepadLayout::PlayStation) {
-		activateButton = "Triangle";
-		castButton = "Square";
+		activateButton = Playstation_Triangle;
+		castButton = Playstation_Square;
 	} else if (GamepadType == GamepadLayout::Nintendo) {
-		activateButton = "Y";
-		castButton = "X";
+		activateButton = Nintendo_Y;
+		castButton = Nintendo_X;
```
StephenCWills, line 1786: Looking at this again, should the Nintendo hints be swapped?

StephenCWills (changes requested):
This is getting really close.

StephenCWills (approved):

Source/items.cpp
```diff
@@ -1764,27 +1763,27 @@ void printItemMiscVirtualGamepad(const Item &item, const bool isOil, bool isCast
...
+		AddPanelString(_("Select from spell book, then\ncast to read"));
+	} else {
 		AddPanelString(_("Activate to read"));
 	}
 }
 
 void printItemMiscGamepad(const Item &item, bool isOil, bool isCastOnTarget)
 {
+	using namespace controllerButtonIcon;
 	std::string activateButton = "Activate";
```
glebm, line 1775: These are not translatable in the current implementation

PR #5359, opened 2022-09-18
Upgrade libpng to 1.6.38

StephenCWills:
I'm pretty sure that https://github.com/diasurgical/libpng/commit/838a5dc474b896fec83f5a42508f739f23f66176 was originally necessary to get the Android build to work, but I didn't encounter any errors in my tests today so we're just going to try without it. 🤞

### DIFF
--- a/.github/workflows/Android.yml
+++ b/.github/workflows/Android.yml
@@ -30,8 +30,8 @@ jobs:
       uses: actions/cache@v2
       with:
         path: android-project/app/.cxx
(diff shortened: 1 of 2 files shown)


StephenCWills:
For a PR that changes just four lines of code, I made an awful lot of mistakes along the way. But it appears to be good now.

AJenbo:
Nice to be on upstream for this 🙂

PR #5361, opened 2022-09-19
Turn off modernize-use-nodiscard in .clang-tidy

aetn23:
Turn off `modernize-use-nodiscard` check in `.clang-tidy`. This check suggests to add `[[nodiscard]]` to some member functions, particularly to `const` members. `[[nodiscard]]` causes the compiler to issue a warning when the result is discarded. This of course is fine, but most of the times .clang-tidy suggests to apply this check to functions that even if ignored won't do any harm (for example `level` function of `Monster`), cluttering the code. I simply stick to the rational that `[[nodiscard]]` should be used only when discarding the result results in a disaster, like a memory leak, not when it simply does not make any sense, but is harmless (like calling `level` and discarding the result). Of course this is a matter of preference; simply put, I think that `[[nodiscard]]` unnecessarily clutters the code.

### DIFF
--- a/Source/.clang-tidy
+++ b/Source/.clang-tidy
@@ -34,6 +34,9 @@
 # -modernize-concat-nested-namespaces
 # -modernize-avoid-bind
 #   Compatibility with older compilers.
+# -modernize-use-nodiscard
+#   This checks suggests to add [[nodiscard]] to member functions,
+#   without a good reason to do so, cluttering the code.
 Checks: >
   -*,
   bugprone-*,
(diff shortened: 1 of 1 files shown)


StephenCWills:
> ... when it simply does not make any sense, but is harmless (like calling `level` and discarding the result).

I thought the whole point of the warning was that this type of error is not necessarily harmless. With or without a memory leak, the example you gave represents a potentially subtle but obvious error, and illustrates exactly the case in which you should want to have this warning.

To put it another way, if I ask myself when I would ever use a const function like `Monster::level()` and choose not to use the result, the answer is only if I made a mistake. If the usage is actually harmless, I may as well remove it. If it isn't harmless, I should be glad for the warning.

PR #5366, opened 2022-09-23
Add Spell Descriptions

kphoenix137:
to info panel while in spellbook

https://discord.com/channels/518540764754608128/518541192993046562/1035691020945461319

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -167,6 +167,61 @@ const char *const PanBtnStr[8] = {
 	"" // Player attack
 };
 
+const char *const SpellDescription[MAX_SPELLS] = {
+	N_("null"),
+	N_("Creates a magical flaming missile"),
+	N_("Instantly restores lost life"),
+	N_("Creates a powerful lightning bolt to lay waste to your enemies"),
+	N_("Releases a short-ranged but deadly burst of energy"),
+	N_("Identify a magic item's properties"),
+	N_("Creates a wall of flame that burns nearby enemies"),
+	N_("Creates a magic portal to town"),
+	N_("Petrifies an enemy, rendering it harmless and defenseless for a short time"),
+	N_("Vision pierces through walls"),
+	N_("A blind, unpredictable blink jump to a nearby location"),
+	N_("Creates a magical shield that consumes mana instead of health when you take damage"),
+	N_("Creates an explosive sphere of fiery death to engulf your enemies"),
+	N_("Summons a multi-headed beast of flame to reduce your enemies to ashes"),
+	N_("Unleashes multiple bolts of lightning into surrounding enemies"),
+	N_("Sends forth a moving wall of fire"),
+	N_("Unleashes twin serpents to hunt an enemy"),
+	N_("Restores spiritual energy at the cost of blood"),
+	N_("Creates an expanding ring of lightning to shock nearby enemies"),
+	N_("Renders the caster unseen, allowing to sneak past enemies"),
+	N_("Creates a continuous jet of flame to scorch your enemies"),
+	N_("Creates a golem from the earth to fight by your side"),
+	N_("Increases physical abilities for a short time"),
+	N_("Instantly moves to a destination within your line of sight"),
+	N_("Devastates the battlefield with unstoppable hellfire"),
+	N_("Fade into the Ether, becoming invulnerable to damage for a few moments"),
+	N_("Repairs an item at cost of reducing its maximum durability"),
+	N_("Replenishes charges in a spell staff at cost of decreasing maximum charge capacity"),
+	N_("Attempts to disarm a trap"),
+	N_("Summons a fiery apparition that charges an enemy and engulfs them in flames"),
+	N_("Creates multiple, randomly directed bolts of electrical energy"),
+	N_("A bolt of divine energy that damages undead"),
+	N_("Brings a deceased ally back to life"),
+	N_("Uses the power of your mind to pick up items, use objects, and knock back enemies"),
+	N_("Instantly restores lost life to an ally"),
+	N_("Creates a powerful bolt of magical energy at the cost of blood"),
+	N_("Summons a powerful spirit to seek the enemy at the cost of blood"),
+	N_("Refills one's spiritual force, restoring a portion of mana"),
+	N_("Completely refills one's spiritual force, restoring all mana"),
+	N_("Allows fate to choose the demise of enemies"),
+	N_("Creates a wall of electrical energy that shocks nearby enemies"),
+	N_("Releases balls of flame in all directions, engulfing enemies in fiery death"),
+	N_("Transports the caster to the nearest stairway, no matter explored or not"),
+	N_("Reflects part of damage taken, mirroring it on attackers"),
+	N_("Drives enemy amok, making it attack friend and foe alike"),
+	N_("Surrounds the caster with a circle of flame"),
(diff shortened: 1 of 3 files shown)


Source/panels/spell_book.cpp
```diff
@@ -130,6 +133,11 @@ void DrawSpellBook(const Surface &out)
 	const int textPaddingTop = 7;
 	for (int i = 1; i < 8; i++) {
 		spell_id sn = SpellPages[sbooktab][i - 1];
+		Rectangle iconArea = { GetPanelPosition(UiPanels::Spell, { 11, 18 }), Size { 37, SpellBookDescription.height * 7 - 5 } };
+		if (iconArea.contains(MousePosition)) {
+			InfoString = fmt::format(fmt::runtime(_("{:s}")), pgettext("spell", spelldescription[sn].sSpellDescription));
```
glebm, line 138: There is no need for fmt::format here:

```suggestion
			InfoString = pgettext("spell", spelldescription[sn].sSpellDescription);
```

FitzRoyX:
This is a mod imo

Source/control.cpp
```diff
@@ -176,6 +176,61 @@ const char *const PanBtnStr[8] = {
...
 
+const char *const SpellDescription[MAX_SPELLS] = {
+	N_("null"),
+	N_("Creates a magical flaming missile"),
+	N_("Instantly restores lost life"),
+	N_("Creates a powerful lightning bolt to lay waste to your enemies"),
+	N_("Releases a short-ranged but deadly burst of energy"),
+	N_("Identify a magic item's properties"),
+	N_("Creates a wall of flame that burns nearby enemies"),
+	N_("Create a magic portal to town"),
```
qndel, line 187:
```suggestion
	N_("Creates a magic portal to town"),
```
You went with "creates" everywhere else

glebm: Perhaps we should actually remove the `s` throughout? It's cleaner

kphoenix137: > Perhaps we should actually remove the `s` throughout? It's cleaner
It sounds more like an instruction than a description without the plural case

kphoenix137:
> This is a mod imo

what's the criterion?

FitzRoyX:
> > This is a mod imo
> 
> what's the criterion?

That's not up to me but I think the steps should be:

1. Does it resolve something that players have historically and frequently complained about?
2. Are there a narrow range of solutions or thousands of possibilities and highly personal artistic decisions to be made?
3. Can it be made optional?

So if I woke up one day and decided that Griswold doesn't have enough dialog, that wouldn't qualify. No one else has complained about it, and even if I wrote new text, that prose is going to be wildly different from everyone else's.

Source/control.cpp
```diff
@@ -976,6 +1031,8 @@ void FreeControlPan()
 	FreeModifierHints();
 }
 
+constexpr Size SpellBookDescription { 250, 43 };
```
glebm, line 1034: This can be moved inside the only function that uses it.

Source/control.cpp
```diff
@@ -1026,6 +1083,27 @@ void DrawInfoBox(const Surface &out)
...
+			const int spellSlots = 7;
+			spell_id snPos[spellSlots];
+			Rectangle iconAreas[spellSlots];
+
+			for (int i = 0; i < spellSlots; i++) {
+				spell_id sn = SpellPages[sbooktab][i];
+				snPos[i] = sn;
+				iconAreas[i] = { GetPanelPosition(UiPanels::Spell, { 11, 18 + (SpellBookDescription.height * i) }), Size { 37, SpellBookDescription.height - 5 } };
+			}
+			int panelWidth = 240;
```
glebm, line 1098: Doesn't look like this variable is ever modified? In this case, consider:

```suggestion
			constexpr int PanelWidth = 240;
```

Source/control.cpp
```diff
@@ -1026,6 +1083,27 @@ void DrawInfoBox(const Surface &out)
 			AddPanelString(fmt::format(fmt::runtime(_("{:s}, Level: {:d}")), _(ClassStrTbl[static_cast<std::size_t>(target._pClass)]), target._pLevel));
 			AddPanelString(fmt::format(fmt::runtime(_("Hit Points {:d} of {:d}")), target._pHitPoints >> 6, target._pMaxHP >> 6));
 		}
+		if (sbookflag) {
+			Player &player = *MyPlayer;
+			uint64_t spl = player._pMemSpells | player._pISpells | player._pAblSpells;
+			const int spellSlots = 7;
```
glebm, line 1089:
```suggestion
			constexpr size_t SpellSlots = 7;
```

Source/control.cpp
```diff
@@ -1026,6 +1083,27 @@ void DrawInfoBox(const Surface &out)
 			AddPanelString(fmt::format(fmt::runtime(_("{:s}, Level: {:d}")), _(ClassStrTbl[static_cast<std::size_t>(target._pClass)]), target._pLevel));
 			AddPanelString(fmt::format(fmt::runtime(_("Hit Points {:d} of {:d}")), target._pHitPoints >> 6, target._pMaxHP >> 6));
 		}
+		if (sbookflag) {
+			Player &player = *MyPlayer;
+			uint64_t spl = player._pMemSpells | player._pISpells | player._pAblSpells;
```
glebm, line 1088:
```suggestion
			const uint64_t spl = player._pMemSpells | player._pISpells | player._pAblSpells;
```

Source/control.cpp
```diff
@@ -1026,6 +1083,27 @@ void DrawInfoBox(const Surface &out)
...
 			AddPanelString(fmt::format(fmt::runtime(_("Hit Points {:d} of {:d}")), target._pHitPoints >> 6, target._pMaxHP >> 6));
 		}
+		if (sbookflag) {
+			Player &player = *MyPlayer;
+			uint64_t spl = player._pMemSpells | player._pISpells | player._pAblSpells;
+			const int spellSlots = 7;
+			spell_id snPos[spellSlots];
+			Rectangle iconAreas[spellSlots];
+
+			for (int i = 0; i < spellSlots; i++) {
```
glebm, line 1093:
```suggestion
			for (size_t i = 0; i < SpellSlots; i++) {
```

Source/control.cpp
```diff
@@ -1026,6 +1083,27 @@ void DrawInfoBox(const Surface &out)
...
+				spell_id sn = SpellPages[sbooktab][i];
+				snPos[i] = sn;
```
glebm, line 1095:
```suggestion
				snPos[i] = SpellPages[sbooktab][i];
```

kphoenix137: Thanks for the suggestions!

Source/control.cpp
kphoenix137: Unable to compile. Used uint8_t instead

ikonomov:
If those descriptions do not exist in the game I think they shouldn't be added.  If this PR is merged next thing we can decide is write out poems for how to use a shield.  It sets a precedent that shouldn't be allowed for this simple reason and also because it it's actually not needed.

kphoenix137:
Closed per request of @ikonomov

ikonomov:
I did not make such request.

kphoenix137:
> I did not make such request.

Sorry, will edit

ikonomov:
Thank you.

PR #5369, opened 2022-09-24
Refactor PrintInfo()

StephenCWills:
The goal of this PR is to clean up the way `PrintInfo()` handles multiline strings as a precursor to a larger cleanup. To do this, I've removed `panelstr`, `pnumlines`, and `ClearPanel()` as well as modified `AddPanelString()` to handle the concatenation of new lines onto `InfoString`. Calls to `AddPanelString()` need to be evaluated to determine whether it would be appropriate to combine them for the purposes of translation.

This is actually a single commit rebased on top of #5367, so keep that in mind when reviewing and/or merging.

Alternative to #5316
Alternative to #5267


StephenCWills:
One more thing I forgot to mention. This actually does away with the warning message in `AddPanelString()`. I'm not sure if it adds any value, considering you can just read the info panel to see if your string was cut off. However, it can easily be added back in `PrintInfo()`.

glebm (approved):

PR #5388, opened 2022-10-02
Restore item seed when converting potions to rejuvenation potions

ephphatha:
Items are initialised using either `GetItemAttrs` (which copies values from `AllItemAttrs` over whatever memory values were already present) or `InitializeItems` (which re-initialises the item struct explicitly then copies values from `AllItemAttrs` over the default/zero-init values).

We can be confident that potions have been spawned in a way that initialises their seed to some random value, so reusing the seed is not harmful.

Out of interest, the only items I saw which have a fixed seed are uniques spawned via `SpawnUnique` (seeds are either 0 from the earlier call to `AllocItem` or set to the `uid` inside `GetUniqueItem`) and the quest items Magic Rock/the combined note which both use a zero-initialised seed.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1279,11 +1279,12 @@ void AddStealPotions(Missile &missile, const AddMissileParameter & /*parameter*/
 
 		bool hasPlayedSFX = false;
 		for (int si = 0; si < MaxBeltItems; si++) {
-			int ii = -1;
-			if (player.SpdList[si]._itype == ItemType::Misc) {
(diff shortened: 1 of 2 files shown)


StephenCWills:
It looks to me like the potion-stealing trap may require the same fix. Can we get that added to this PR?

https://github.com/diasurgical/devilutionX/blob/c80fcf42c899171d2d27971baacf4a0010730810/Source/missiles.cpp#L1317-L1320

StephenCWills (approved):

PR #5394, opened 2022-10-02
TFit_Obj5: Fix theme room generation when only last skipped candidate was valid

obligaron:
Fixes #5360

The bug was introduced with #5117 / 08acd4aac05158e09655a54ee60fd5e15014993a

Previous `TFit_Obj5` only returned false when [no place for the (first candidate) theme room was found (`r == rs`)](https://github.com/diasurgical/devilutionX/commit/08acd4aac05158e09655a54ee60fd5e15014993a#diff-b22a42607df3745a75d7085a9f065fbd0be5b4522c3d24527480312e4e7a3bf5L128). In this case searching for a second candidate/position is not necessary, cause the entire dungeon can't fit any.
Else the function [returned true and used the coordinates for the last candidate](https://github.com/diasurgical/devilutionX/commit/08acd4aac05158e09655a54ee60fd5e15014993a#diff-b22a42607df3745a75d7085a9f065fbd0be5b4522c3d24527480312e4e7a3bf5L139-L142).

@ephphatha Could you take a look? That would be great. 🙂

Issue #5360, opened 2022-09-19
Crash when rendering invalid objects loaded from savegame


licaon-kter:
### Operating System

Linux x64

### DevilutionX version

Custom build (please specify commit ID)

### Describe

4cff046ba946859fca1d9e4cb07df950f9773928 on Debian Sid, nVidia closed driver

### To Reproduce

* Start game
* Load savegame

### Expected Behavior

Play game

### Additional context

I'm guessing it's related to https://github.com/diasurgical/devilutionX/pull/4931

There are 2 ways to trigger the bug:
* .INI has `Zoom=1` load the save game, then press Z (to zoom)
* .INI has `Zoom=0` load the save game, crashes right away

```
CRITICAL: Unable to find object_graphic_id 7 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 13 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 13 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 7 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 13 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 13 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 7 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 5 in list of objects to load, level generation error.
CRITICAL: Unable to find object_graphic_id 13 in list of objects to load, level generation error.
ERROR: Eroare
assertion failed (Source/engine/clx_sprite.hpp:130)
spriteIndex < numSprites()
```

AJenbo:
from the log it sounds more like the issue is an object not being loaded and failing to render, nothing actually related ti zoom, also its failing on a debug assert and not actually crashing

AJenbo:
what level is the save on, how was it created?

licaon-kter:
It's an old save that I use for random testing, will upload it asap.

licaon-kter:
Attached savegame:
[a.zip](https://github.com/diasurgical/devilutionX/files/9638985/a.zip)

StephenCWills:
I added a bit of logging and got these messages.

```
DEBUG: OBJ_SKFIRE (29,33)
CRITICAL: Unable to find object_graphic_id 7 in list of objects to load, level generation error.
DEBUG: OBJ_SKELBOOK (29,31)
[...]

PR #5395, opened 2022-10-02
Use a counter to track the number of candidates considered in TFit_Obj5

ephphatha:
Based on #5394, slight simplification of the logic involved.

I was going to add a test case but it seems that the theme array isn't initialised in headless mode? The seed in #5360 ends up with 5 themes on floor 1, THEME_SKELROOM, THEME_TREASURE, THEME_SKELROOM, THEME_MONSTPIT, THEME_SKELROOM. The bugged logic in #5117 fails to place all skeleton rooms which results in the skeleton tomes/banners not being loaded as the themes aren't marked as used. They exist in the save game objects list though which causes the crash when the game attempts to render them.

### DIFF
--- a/Source/levels/themes.cpp
+++ b/Source/levels/themes.cpp
@@ -101,28 +101,26 @@ bool CheckThemeObj5(Point origin, int8_t regionId)
 
 bool TFit_Obj5(int t)
 {
-	int skipCandidates = GenerateRnd(5);
-	if (skipCandidates < 0) {
+	const int targetCandidates = GenerateRnd(5);
+	if (targetCandidates < 0) {
 		// vanilla rng can return -3 for GenerateRnd(5), default behaviour is to set the output to 0,0 and return true in this case...
 		themex = 0;
 		themey = 0;
 		return true;
 	}
 
-	bool validFound = false;
-	for (int yp = 0; yp < MAXDUNY; yp++) {
-		for (int xp = 0; xp < MAXDUNX; xp++) {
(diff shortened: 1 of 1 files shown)


ephphatha:
Sorry, picked the wrong field when I was attempting to request a review -_-

Source/levels/themes.cpp
```diff
@@ -101,28 +101,26 @@ bool CheckThemeObj5(Point origin, int8_t regionId)
...
-			}
+	int candidatesFound = 0;
+	for (Point tile : PointsInRectangleRange { { { 0, 0 }, { MAXDUNX, MAXDUNY } } }) {
+		if (dTransVal[tile.x][tile.y] == themes[t].ttval && IsTileNotSolid(tile) && CheckThemeObj5(tile, themes[t].ttval)) {
+			// Use themex/y to keep track of the last candidate area found, in case we end up with fewer candidates than the target
+			themex = tile.x;
+			themey = tile.y;
+			candidatesFound++;
+			if (candidatesFound > targetCandidates)
+				break;
```
obligaron, line 120: nit style question: should we return here instead of break?
For me this would be a little bit more clear ("We found our candidate, get the hell out of here" 😓).
Note: you can discard the comment. It's only a style question not functional.

obligaron (approved):
lgtm, only a style question 🙂 
Tested with the reported save and another save with the same issue.

Source/levels/themes.cpp
ephphatha: Yeah, I was torn. I went with a break because the return value is a different condition, even though it's always true if this gets hit...

ephphatha: Needed to rebase anyway so went with the early return this time :D

AJenbo (approved):

PR #5396, opened 2022-10-03
monster.{h,cpp}: Minor refactor

glebm:
1. Extract some common code into functions.
2. Avoid some overly long lines.


AJenbo (approved):
nice catch, good cleanup :)

PR #5407, opened 2022-10-11
Reorder/optimize large structs

glebm:
[`pahole`](https://linux.die.net/man/1/pahole) showed a few simple reorderings to reduce struct sizes:

```
pahole --reorganize --show_reorg_steps --show_only_data_members -C AnimationInfo build/devilutionx
```

This PR also changes a couple of members to a smaller size.

This leads to significant reduction in the overall size of compound structs, e.g. `Player`: 18,304 bytes -> 16,848 bytes.
`Monster`: 112 -> 104


AJenbo (approved):
Nice :)

PR #5409, opened 2022-10-11
Xbox nxdk: Set gamepad type

glebm:
Set the gamepad type to Xbox for the original Xbox.


AJenbo (approved):
Forgot about this one when reviewing the previous PR :)

Issue #5025, opened 2022-07-17
Remove `DVL_WM_*`


glebm:
WIth the recent cleanups (#5018 and #5023), each SDL event now corresponds to at most 1 `DVL_WM_*` event.

We should remove the `misc_msg` code entirely and just use SDL events directly.

This will require changing the demo format to record e.g. the handled SDL events, which means re-recording or otherwise transforming the timedemo test file (perhaps similar to how it was done in #5047)

glebm:
`DVL_WM_*` stuff has been removed. SDL events handling should still be cleaned up a bit (happens in too many places) but that's a separate issue.

PR #5411, opened 2022-10-12
Add padmapper for gamepad customization

StephenCWills:
Adds a padmapper for customizing actions associated with gamepad buttons. The controls basically feel the same, as far as I've been able to tell. Hopefully the code is easier to follow now as well.

* Enables mappings for two-button combos
* Introduces gamepad actions for nearly all of the keyboard actions
* Introduces actions for "stand ground" and "toggle stand ground" to enable Shift+Click functionality for gamepad
* Implements the currently hardcoded control scheme as the padmapper defaults
* Swaps A/B and X/Y menu actions for non-Nintendo controllers based on the `GamepadLayout` (note, this may have a somewhat adverse effect on the Quest Log and Spellbook screens that should be addressed at some point)
* Removes settings for d-pad hotkeys and swap shoulder button mode as it should now be more or less possible to achieve the same result by remapping buttons

Some additional notes for reviewers...

* Options are currently hidden from the settings menu since the code to bind the gamepad buttons has not yet been written
* Most of the actions were straight-up copied from keymapper, and I haven't tested to make sure they work the same when mapped to the gamepad
* This PR sets the `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS` hint so that buttons are always positional regardless of the gamepad layout. This was done to ensure that the on-screen hints we display when holding Start or Select would be accurate to the position of the buttons on the gamepad. This is why it was necessary to add the `TranslateTo()` function to swap button codes around for Nintendo layout controllers. My hope for when the settings menu is implemented would be to add gamepad icons for the larger font so we can display the appropriate buttons based on the layout, but continue to save the positional layout to the INI.

### DIFF
--- a/Source/controls/game_controls.cpp
+++ b/Source/controls/game_controls.cpp
@@ -11,8 +11,10 @@
 #include "controls/plrctrls.h"
 #include "controls/touch/gamepad.h"
 #include "doom.h"
+#include "gamemenu.h"
 #include "gmenu.h"
 #include "options.h"
+#include "qol/stash.h"
 #include "stores.h"
 
 namespace devilution {
@@ -101,36 +103,6 @@ SDL_Keycode TranslateControllerButtonToSpellbookKey(ControllerButton controllerB
 	default:
 		return SDLK_UNKNOWN;
 	}
-}
-
-} // namespace
-
-ControllerButton TranslateTo(GamepadLayout layout, ControllerButton button)
-{
-	if (layout != GamepadLayout::Nintendo)
-		return button;
-
-	switch (button) {
-	case ControllerButton_BUTTON_A:
-		return ControllerButton_BUTTON_B;
-	case ControllerButton_BUTTON_B:
-		return ControllerButton_BUTTON_A;
-	case ControllerButton_BUTTON_X:
-		return ControllerButton_BUTTON_Y;
-	case ControllerButton_BUTTON_Y:
-		return ControllerButton_BUTTON_X;
-	default:
-		return button;
-	}
-}
-
-bool SkipsMovie(ControllerButtonEvent ctrlEvent)
-{
-	return IsAnyOf(ctrlEvent.button,
-	    ControllerButton_BUTTON_A,
-	    ControllerButton_BUTTON_B,
-	    ControllerButton_BUTTON_START,
-	    ControllerButton_BUTTON_BACK);
 }
 
 bool GetGameAction(const SDL_Event &event, ControllerButtonEvent ctrlEvent, GameAction *action)
@@ -241,6 +213,121 @@ bool GetGameAction(const SDL_Event &event, ControllerButtonEvent ctrlEvent, Game
 	return false;
 }
 
+void PressControllerButton(ControllerButton button)
+{
+	if (IsStashOpen) {
+		switch (button) {
+		case ControllerButton_BUTTON_BACK:
+			StartGoldWithdraw();
+			return;
+		case ControllerButton_BUTTON_LEFTSHOULDER:
+			Stash.PreviousPage();
+			return;
+		case ControllerButton_BUTTON_RIGHTSHOULDER:
+			Stash.NextPage();
+			return;
+		}
+	}
+
+	if (PadHotspellMenuActive) {
+		auto quickSpellAction = [](size_t slot) {
+			if (spselflag) {
+				SetSpeedSpell(slot);
+				return;
+			}
+			if (!*sgOptions.Gameplay.quickCast)
+				ToggleSpell(slot);
+			else
+				QuickCast(slot);
+		};
+		switch (button) {
+		case devilution::ControllerButton_BUTTON_A:
+			quickSpellAction(2);
+			return;
+		case devilution::ControllerButton_BUTTON_B:
+			quickSpellAction(3);
+			return;
+		case devilution::ControllerButton_BUTTON_X:
+			quickSpellAction(0);
+			return;
+		case devilution::ControllerButton_BUTTON_Y:
+			quickSpellAction(1);
+			return;
+		default:
+			break;
+		}
+	}
+
+	if (PadMenuNavigatorActive) {
+		switch (button) {
+		case devilution::ControllerButton_BUTTON_DPAD_UP:
+			PressEscKey();
+			LastMouseButtonAction = MouseActionType::None;
+			PadHotspellMenuActive = false;
+			PadMenuNavigatorActive = false;
+			gamemenu_on();
+			return;
+		case devilution::ControllerButton_BUTTON_DPAD_DOWN:
+			DoAutoMap();
+			return;
+		case devilution::ControllerButton_BUTTON_DPAD_LEFT:
+			ProcessGameAction(GameAction { GameActionType_TOGGLE_CHARACTER_INFO });
+			return;
+		case devilution::ControllerButton_BUTTON_DPAD_RIGHT:
+			ProcessGameAction(GameAction { GameActionType_TOGGLE_INVENTORY });
+			return;
+		case devilution::ControllerButton_BUTTON_A:
+			ProcessGameAction(GameAction { GameActionType_TOGGLE_SPELL_BOOK });
+			return;
+		case devilution::ControllerButton_BUTTON_B:
+			return;
+		case devilution::ControllerButton_BUTTON_X:
+			ProcessGameAction(GameAction { GameActionType_TOGGLE_QUEST_LOG });
+			return;
+		case devilution::ControllerButton_BUTTON_Y:
+#ifdef __3DS__
+			sgOptions.Graphics.zoom.SetValue(!*sgOptions.Graphics.zoom);
+			CalcViewportGeometry();
+#endif
+			return;
+		default:
+			break;
+		}
+	}
+
+	sgOptions.Padmapper.ButtonPressed(button);
+}
+
+} // namespace
+
+ControllerButton TranslateTo(GamepadLayout layout, ControllerButton button)
+{
+	if (layout != GamepadLayout::Nintendo)
+		return button;
+
+	switch (button) {
+	case ControllerButton_BUTTON_A:
+		return ControllerButton_BUTTON_B;
+	case ControllerButton_BUTTON_B:
+		return ControllerButton_BUTTON_A;
+	case ControllerButton_BUTTON_X:
+		return ControllerButton_BUTTON_Y;
+	case ControllerButton_BUTTON_Y:
+		return ControllerButton_BUTTON_X;
+	default:
+		return button;
+	}
+}
+
+bool SkipsMovie(ControllerButtonEvent ctrlEvent)
+{
+	return IsAnyOf(ctrlEvent.button,
+	    ControllerButton_BUTTON_A,
+	    ControllerButton_BUTTON_B,
+	    ControllerButton_BUTTON_START,
+	    ControllerButton_BUTTON_BACK);
+}
+
 bool IsSimulatedMouseClickBinding(ControllerButtonEvent ctrlEvent)
 {
 	ControllerButtonCombo leftMouseClickBinding1 = sgOptions.Padmapper.ButtonComboForAction("LeftMouseClick1");
@@ -256,6 +343,35 @@ bool IsSimulatedMouseClickBinding(ControllerButtonEvent ctrlEvent)
 AxisDirection GetMoveDirection()
 {
 	return GetLeftStickOrDpadDirection(true);
+}
+
+bool HandleControllerButtonEvent(const SDL_Event &event, GameAction &action)
+{
+	const ControllerButtonEvent ctrlEvent = ToControllerButtonEvent(event);
+	bool isGamepadMotion = ProcessControllerMotion(event, ctrlEvent);
+	DetectInputMethod(event, ctrlEvent);
+	if (isGamepadMotion) {
+		return true;
+	}
+
+	if (ctrlEvent.button != ControllerButton_NONE && ctrlEvent.button == SuppressedButton) {
+		if (!ctrlEvent.up)
+			return true;
+		SuppressedButton = ControllerButton_NONE;
+	}
+
+	if (GetGameAction(event, ctrlEvent, &action)) {
+		ProcessGameAction(action);
+		return true;
+	} else if (ctrlEvent.button != ControllerButton_NONE) {
+		if (!ctrlEvent.up)
+			PressControllerButton(ctrlEvent.button);
+		else
+			sgOptions.Padmapper.ButtonReleased(ctrlEvent.button);
+		return true;
+	}
+
+	return false;
 }
 
 } // namespace devilution
--- a/Source/controls/game_controls.cpp
+++ b/Source/controls/game_controls.cpp
@@ -2,47 +2,57 @@
 
 #include <cstdint>
 
-#include "controls/controller.h"
 #include "controls/controller_motion.h"
 #include "miniwin/misc_msg.h"
 #ifndef USE_SDL1
 #include "controls/devices/game_controller.h"
 #endif
 #include "controls/devices/joystick.h"
-#include "controls/menu_controls.h"
-#include "controls/modifier_hints.h"
 #include "controls/plrctrls.h"
 #include "controls/touch/gamepad.h"
 #include "doom.h"
 #include "gmenu.h"
 #include "options.h"
-#include "qol/stash.h"
 #include "stores.h"
 
 namespace devilution {
 
-bool start_modifier_active = false;
-bool select_modifier_active = false;
-const ControllerButton ControllerButtonPrimary = ControllerButton_BUTTON_B;
-const ControllerButton ControllerButtonSecondary = ControllerButton_BUTTON_Y;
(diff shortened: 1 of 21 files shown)


Source/controls/game_controls.cpp
```diff
@@ -56,50 +69,72 @@ SDL_Keycode TranslateControllerButtonToKey(ControllerButton controllerButton)
...
-	const bool startIsDown = IsControllerButtonPressed(ControllerButton_BUTTON_START);
-	const bool selectIsDown = IsControllerButtonPressed(ControllerButton_BUTTON_BACK);
-	start_modifier_active = !inGameMenu && startIsDown;
-	select_modifier_active = !inGameMenu && selectIsDown && !start_modifier_active;
-
-	// Tracks whether we've received both START and SELECT down events.
-	//
-	// Using `IsControllerButtonPressed()` for this would be incorrect.
-	// If both buttons are pressed simultaneously, SDL sends 2 events for which both buttons are in the pressed state.
-	// This allows us to avoid triggering START+SELECT action twice in this case.
```
glebm, line 72: Does the new code handle this case correctly?

Source/diablo.cpp
```diff
@@ -1690,6 +1781,463 @@ void InitKeymapActions()
...
+	    },
+	    [] {
+		    ControllerActionHeld = GameActionType_NONE;
+		    LastMouseButtonAction = MouseActionType::None;
+	    },
+	    CanPlayerTakeAction);
+	sgOptions.Padmapper.AddAction(
+	    "CancelAction",
+	    N_("Cancel action"),
+	    N_("Close menus, open speedbook."),
```
glebm, line 1870: Opening a speedbock should probably be on a separate action with the same mapping

Source/diablo.cpp
```diff
@@ -1690,6 +1781,463 @@ void InitKeymapActions()
...
+	    N_("Spellbook"),
+	    N_("Open Spellbook."),
+	    { ControllerButton_BUTTON_BACK, ControllerButton_AXIS_TRIGGERRIGHT },
+	    [] {
+		    ProcessGamepadEvents(GameAction { GameActionType_TOGGLE_SPELL_BOOK });
+	    },
+	    nullptr,
+	    CanPlayerTakeAction);
+	sgOptions.Padmapper.AddAction(
+	    "DisplaySpells",
```
glebm, line 1984: "OpenSpeedbook" and map to button A by default?

glebm:
> This PR sets the SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS hint so that buttons are always positional regardless of the gamepad layout.

As I understand, that would be `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS=0` but the PR currently sets it to 1, necessitating the `TranslateTo` function which we wouldn't need otherwise?

Source/options.cpp
```diff
@@ -1419,6 +1417,270 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
...
+{
+	std::array<char, 64> result;
+	if (!GetIniValue(category.data(), key.data(), result.data(), result.size())) {
+		SetValue(defaultInput);
+		return; // Use the default button combo if no mapping has been set.
+	}
+
+	std::string modName;
+	std::string buttonName;
+	for (string_view name : SplitByChar(result.data(), '+')) {
```
glebm, line 1487:
```suggestion
	for (string_view name : SplitByChar(result, '+')) {
```

Source/controls/game_controls.cpp
StephenCWills: I'm not sure this case matters much. Unless I misunderstood something here, it's trying to handle cases like when the player is holding `Select` and then accidentally presses `Start` and `A` at the exact same time. This code would swallow the `A` input and only trigger `Select+Start`. The padmapper would depend on what order SDL sends the events, but the case is highly unlikely to happen, and it shouldn't be at all surprising to the user if we handle the extra button.

Maybe I don't understand the use-case, though. I suppose there could be a device that sends spurious `A` events when `Start` and `Select` are pressed together or something..?

Source/diablo.cpp
StephenCWills: Hm, are you thinking that someone might want to map, say, the inventory to this button instead of the speedbook? I suppose we could make this could work if we tune the `enable` functions for these actions so that they are mutually exclusive. Otherwise, it will definitely result in some bad interactions between these two actions.

Source/diablo.cpp
StephenCWills: I used the name from the keymapper. Should we change it in both?

Source/diablo.cpp
glebm: The cancel action should always stop further handling of any other binding, not sure how easy it is to implement in the current system though

StephenCWills:
> As I understand, that would be `SDL_HINT_GAMECONTROLLER_USE_BUTTON_LABELS=0` but the PR currently sets it to 1, necessitating the `TranslateTo` function which we wouldn't need otherwise?

You're right that I should have set it to zero, but we do still need the `TranslateTo` function. Otherwise, approve and cancel menu actions would be backwards on Nintendo controllers.

Source/controls/game_controls.cpp
glebm: The case handled here is actually just pressing SELECT + START simultaneously. If you can, please give it try it with SDL2.
Pressing SELECT + START triggers 2 events: the first opened the menu and the second immediately closed it in the previous implementation before this workaround.

Source/diablo.cpp
glebm: Ah, probably not worth changing the name then as it'd break existing keybindings.

Source/options.cpp
```diff
@@ -1419,6 +1417,270 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
...
+		modName = buttonName;
+		buttonName = name;
```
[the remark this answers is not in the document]
StephenCWills: The intent is to consume strings of the form `Modifier+Button` where the modifier is optional. So it uses the part before the last occurrence of `+` as the modifier and the part after the last occurrence of `+` as the button. In theory, there should only ever be one occurrence of `+` so whatever we do in cases with multiple plus signs is effectively UB. If you prefer `find_last_of`, I can make it work with that.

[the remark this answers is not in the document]
glebm: Ah, I see, how about this then:

```c++
auto it = SplitByChar(result, '+');
modName = *it;
++it;
buttonName = *it;
```

Source/controls/game_controls.cpp
StephenCWills: Actually, I needed to reread the whole comment. If `Start+Select` are pressed simultaneously, it would trigger two events and the game menu would open and immediately close because the second event would be translated to the Escape key. I imagine this could happen more often on devices that aren't processing events at 150+ fps.

I was initially thinking this would be handled properly because the `Start` event would come in and trigger its own action, thereby becoming the modifier key for `Select`. However, when I eventually implemented `PadmapperOptions::ButtonPressed()`, I ended up using `IsControllerButtonPressed()` and falling into the same trap. I'll try to think of a way to iron this out.

Source/diablo.cpp
StephenCWills: I think I see what you mean. Otherwise, we'd have to make the speedbook action unconditionally operate only when there are no other menus open. In that case, my best idea would be to tune the `enable` function for the `CancelAction`, get rid of the `inputToAction` lookup, and have the padmapper scan the `actions` vector to select the first enabled action to execute.

Source/controls/game_controls.cpp
StephenCWills: This should be resolved. Refer to the logic related to the `SuppressedButton` variable.

Source/options.cpp
StephenCWills: This actually doesn't work so I left it as-is. `result` is an `std::array<char, 64>` so the call to `result.data()` is necessary to treat it as a C string.

Source/diablo.cpp
```diff
@@ -642,6 +727,12 @@ void GameEventHandler(const SDL_Event &event, uint16_t modState)
 			sgOptions.Keymapper.KeyReleased(event.button.button | KeymapperMouseButtonMask);
 		}
 		return;
+	case SDL_JOYBUTTONDOWN:
+		PressControllerButton(static_cast<ControllerButton>(event.jbutton.button));
```
glebm, line 731: 1. Why is this being handled here now?

2. This doesn't seem correct, `event.jbutton.button` and `ControllerButton` have different domains.
Isn't `event.jbutton.button` an arbitrary numeric code? We usually translate joystick events via `Joystick::ToControllerButton`.

Source/options.cpp
```diff
@@ -1419,6 +1418,249 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
...
+void PadmapperOptions::AddAction(string_view key, const char *name, const char *description, ControllerButtonCombo defaultInput, std::function<void()> actionPressed, std::function<void()> actionReleased, std::function<bool()> enable, unsigned index)
+{
+	actions.push_back(std::unique_ptr<Action>(new Action(key, name, description, defaultInput, std::move(actionPressed), std::move(actionReleased), std::move(enable), index)));
+}
+
+void PadmapperOptions::ButtonPressed(ControllerButton button)
+{
+	// To give preference to button combinations,
+	// first pass ignores mappings where no modifier is bound
+	for (auto it = actions.begin(); it != actions.end(); ++it) {
```
glebm, line 1593:
```suggestion
	for (Action *action_ptr : actions) {
	  Action &action = *action_ptr;
	}
```

Source/options.cpp
```diff
@@ -1419,6 +1418,249 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
...
+		if (action.enable && !action.enable())
+			continue;
+		if (action.actionPressed)
+			action.actionPressed();
+		SuppressedButton = combo.modifier;
+		buttonToReleaseAction.insert_or_assign(combo.button, action);
+		return;
+	}
+
+	for (auto it = actions.begin(); it != actions.end(); ++it) {
```
glebm, line 1611:
```suggestion
	for (Action *action_ptr : actions) {
	  Action &action = *action_ptr;
	}
```

Source/options.cpp
```diff
@@ -1419,6 +1418,249 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
...
+	if (!action.actionReleased || (action.enable && !action.enable()))
+		return;
+
+	action.actionReleased();
+	buttonToReleaseAction.erase(button);
+}
+
+string_view PadmapperOptions::InputNameForAction(string_view actionName) const
+{
+	for (const auto &action : actions) {
```
glebm, line 1646: A reference to a pointer seems like a typo

```suggestion
	for (const Action *action : actions) {
```

Source/options.cpp
```diff
@@ -1419,6 +1418,249 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
...
+		if (action->key == actionName && action->boundInput.button != ControllerButton_NONE) {
+			return action->GetValueDescription();
+		}
+	}
+	return "";
+}
+
+ControllerButtonCombo PadmapperOptions::ButtonComboForAction(string_view actionName) const
+{
+	for (const auto &action : actions) {
```
glebm, line 1656:
```suggestion
	for (const Action *action : actions) {
```

glebm (commented):
I haven't tested the PR yet but the code looks good overall, just left a few comments.

Source/diablo.cpp
StephenCWills: In a similar fashion to how we handle gamepad events as keystrokes, I modified `FetchMessage_Real()` to convert all gamepad events to `SDL_JOYBUTTON*` events specifically so they could be handled by the `GameEventHandler()`. In doing so, I capture the `ControllerButton` code as the `button` value for the converted event.

https://github.com/diasurgical/devilutionX/blob/beabeb0b8d43d94adaf53b890f931eca0c0c5145/Source/miniwin/misc_msg.cpp#L177-L179

As for why it was moved, it's because the padmapper actions should only run when `GameEventHandler` is registered. This is the same way that keymapper is used. I wouldn't say I'm particularly proud of the way this is handled, but the registered event handler can only handle SDL events so I figured this was a decent compromise.

glebm: If a real `SDL_JOYBUTTONDOWN` comes in (e.g. the extra buttons on the Switch controller), wouldn't this cast be invalid?

Perhaps we should move the gamepad handling out of `FetchMessage_Real` to the actual event handling code (such as `GameEventHandler` etc)?

Source/controls/modifier_hints.cpp
```diff
@@ -158,13 +158,10 @@ void DrawStartModifierMenu(const Surface &out)
 
 void DrawSelectModifierMenu(const Surface &out)
 {
-	if (!select_modifier_active || SimulatingMouseWithSelectAndDPad)
+	if (!PadHotspellMenuActive || SimulatingMouseWithPadmapper)
```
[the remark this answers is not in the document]
StephenCWills: This is the drawing function for the pad hostpell menu, so this check is probably what prevents it from being shown. It looks like the function needs to be renamed.

Source/options.h
```diff
@@ -682,6 +680,63 @@ struct KeymapperOptions : OptionCategoryBase {
...
+	    std::function<void()> actionReleased = nullptr,
+	    std::function<bool()> enable = nullptr,
+	    unsigned index = 0);
+	void ButtonPressed(ControllerButton button);
+	void ButtonReleased(ControllerButton button);
+	string_view InputNameForAction(string_view actionName) const;
+	ControllerButtonCombo ButtonComboForAction(string_view actionName) const;
+
+private:
+	std::vector<std::unique_ptr<Action>> actions;
```
glebm, line 734: Perhaps `std::forward_list<Action>`?

Source/diablo.cpp
StephenCWills: No, that shouldn't happen. The `FetchMessage_Real()` function takes the `event` pointer from the caller and then captures the result of `PollEvent()` to the `e` variable. Data has to be explicitly written to the `event` variable that was passed in for it to reach the registered event handler. This means it will only send an `SDL_JOYBUTTONDOWN` event to the registered event handler if the `ControllerButton` is recognized.

It might indeed make some sense to move the rest of the event handling code to the `GameEventHandler`, but I felt that `FetchMessage_Real()` appears to be specifically designed for filtering and translating so to me it seemed more appropriate to leave it there. This way, the keystroke that the gamepad event gets translated into would be the one handled by the `GameEventHandler`.

glebm: We ideally want to remove all the redundant translation layers and `misc_msg.cpp` in the future (I did some work towards that in https://github.com/diasurgical/devilutionX/issues/5025).

How about we make a separate function to do the gamepad-related stuff that is currently being done in `misc_msg.cpp` and call that from `GameEventHandler`?
Then we won't need the hacky `SDL_JOYBUTTON` at all!

fwannmacher:
I was working on this. So I guess I won't do it anymore xD. Nicely done :D

Source/options.cpp
```diff
@@ -1419,6 +1418,249 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
...
+	if (!GetIniValue(category.data(), key.data(), result.data(), result.size())) {
+		SetValue(defaultInput);
+		return; // Use the default button combo if no mapping has been set.
+	}
+
+	std::string modName;
+	std::string buttonName;
+	auto parts = SplitByChar(result.data(), '+');
+	auto it = parts.begin();
+	buttonName = std::string(*it);
```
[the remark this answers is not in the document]
StephenCWills: Hmm, I think you're right. I guess in the case of empty string, `SplitByCharIterator::begin() == SplitByCharIterator::end()`, but the `_slice` variable returned by `operator*()` would be `text.substr(text.size(), npos)`: an empty string. Thus it worked okay in my testing, but `*iterable.end()` is generally considered UB so we should check it.

Source/controls/modifier_hints.cpp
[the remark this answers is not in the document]
StephenCWills: I renamed both `DrawStartModifierMenu()` and `DrawSelectModifierMenu()` and added `SimulatingMouseWithPadmapper` to both, based on the assumption that this was added to keep those menus out of the way when trying to place the mouse.

Source/options.cpp
```diff
@@ -1419,6 +1418,255 @@ uint32_t KeymapperOptions::KeyForAction(string_view actionName) const
 	return SDLK_UNKNOWN;
 }
 
+PadmapperOptions::PadmapperOptions()
+    : OptionCategoryBase("Padmapping", N_("Padmapping"), N_("Padmapping Settings"))
+{
+	buttonToButtonName.emplace(ControllerButton_AXIS_TRIGGERLEFT, "LT");
```
glebm, line 1424: nit: Construct in one go, as in `: buttonToButtonName({{ControllerButton_AXIS_TRIGGERLEFT, "LT"}, ...)`

glebm (approved):

AJenbo:
Big improvement :+1:

Issue #1152, opened 2021-03-10
Unable to open chest in Ogden's Tavern Sign quest after talking to Snotspill


Tiddalick:
Running 64 bit build of devilutionX 3453, created automatically on circleci.

I am unable to open the Locked chest that holds Ogden's sign.
I have spoken to Snotspill who repeats his instruction to kill the 'Big Uglies'.
I already had the quest from Ogden originally.

I am unable to proceed with the quest and the game. I include a savegame at the location for testing.

I have enabled some of the additional ini settings, specifically the autogold pickup, monster health bars and the xp bar.

[single_0.zip](https://github.com/diasurgical/devilutionX/files/6116925/single_0.zip)

AJenbo:
intereatin thatsthe secoubd report about quests being stuck on 64bit windows. could you test with te 32bit build

AJenbo:
what does the 3453 number refer to

julealgon:
I think I heard in one of the streams that this happens in vanilla as well.

Question to OP: was this game started from scratch? Or did you play this character (and reach this quest) after playing with some other character/save in the same session, without quitting to desktop?

The Valor quest has a similar bug in vanilla where it becomes unoperable if you get there after having already played another character without quitting the game.

qndel:
I edited

AJenbo:
i know, but that doesn't answer the question :)

AJenbo:
@alanbirtles how was the faulty build compiled?

alanbirtles:
I'm pretty sure I was using a Circle Ci build from around the 3rd March (according to my download history)

AJenbo:
Ok, I think that isolates this to a Mingw64-64bit issue then.

alanbirtles:
Just tried a release visual studio build of the latest master with that save and it shows the same issue

qndel:
well, it  shows 
![image](https://user-images.githubusercontent.com/14297035/110688969-74606b00-81e2-11eb-9f08-21de2e9dc428.png) for me, the question is what caused that situation

AJenbo:
Yeah, the save is probably no good. Need to test with a fresh game.

AJenbo:
[...]

PR #5413, opened 2022-10-13
Reenable LTO for MinGW builds

StephenCWills:
I'm hoping these are no longer relevant.

LTO was disabled for these compilers because a reference to a C string pointer was being optimized to refer to a different memory location with the same content (maybe a bad implementation of `-fmerge-constants`??). https://github.com/diasurgical/devilutionX/issues/1152#issuecomment-798126255

Not only was that particular bug fixed in 8b8caf5, the compiler version was updated to 10.0.0 since then. Furthermore, the code has been updated to use `string_view` in a lot of places where string comparison is still required.


AJenbo:
I guess it's a good time to enable and test

PR #5416, opened 2022-10-16
trigs.cpp: Fix build error

glebm:


Source/levels/trigs.cpp
```diff
@@ -754,33 +754,41 @@ bool ForcePWaterTrig()
 
 bool ForceArenaTrig()
 {
-	int *checkList = nullptr;
+	const uint16_t *checkList = nullptr;
+	size_t len;
```
qndel, line 758:
```suggestion
	size_t len = 0;
```
just to give it a default value if we did it with checkList? :P

glebm: No need because we always assign it (or `return false`)

glebm: Removed the default value for `checkList` as well

ephphatha (approved):

Source/levels/trigs.cpp
qndel: that works too, just wanted consistency :p

PR #5418, opened 2022-10-17
Remove ConsumeSpell() from AddMissile() functions

### LINKED ISSUE #5237: PvP mod/master bug - Murkey Pools for the arena's cost the player 30 mana.
### Operating System

Windows x64

### DevilutionX version

### Describe

Here is a link to the build in question - https://github.com/diasurgical/devilutionX/suites/7789559185/artifacts/328190699

After the player casts telekineses to click on a Murky Pool shrine in either Arena 2 or 3, the pool removes 30 mana.

### To Reproduce

type /arena 2 or /arena 3 to proceed to the arenas, cast telekineses on any Murky Pool shrine.

### Expected Behavior

For normal mana cost (8) to be withdrawn from a normal telekinesis cast, but no mana withdrawn from actually touching the Murky Pool shrine

### Additional context

https://user-images.githubusercontent.com/98377923/184463020-b1085f09-1ff4-454a-9f5e-6ea9d84d708e.mov

StephenCWills:
Moves the invocation of `ConsumeSpell()` to the `CastSpell()` function. This ensures that the mana, scroll, or staff charge is only consumed when a player casts a spell, and avoids all the tricky logic in the `AddMissile()` functions to ensure that a player did indeed cast the spell (as opposed to a monster, trap, Guardian, etc). For missiles that would conditionally fail, such as Stone Curse when it can't find a target, I added the `Missile::_miSpellFizzle` variable to indicate that the spell should not consume mana. This works out well for `AddJester()` which previously relied on the call to `ConsumeSpell()` in the `AddMissile()` function for the random missile that it generated to consume the staff charge. Now it more appropriately invokes `ConsumeSpell(player, SPL_JESTER)` if the generated missile does not fizzle.

Alternative to #5243
This resolves #5237

### DIFF
--- a/Source/missiles.h
+++ b/Source/missiles.h
@@ -238,43 +238,44 @@ struct AddMissileParameter {
 	Point dst;
 	Direction midir;
 	Missile *pParent;
+	bool spellFizzled;
 };
 
-void AddHiveExplosion(Missile &missile, const AddMissileParameter &parameter);
-void AddFireRune(Missile &missile, const AddMissileParameter &parameter);
-void AddLightningRune(Missile &missile, const AddMissileParameter &parameter);
-void AddGreatLightningRune(Missile &missile, const AddMissileParameter &parameter);
-void AddImmolationRune(Missile &missile, const AddMissileParameter &parameter);
-void AddStoneRune(Missile &missile, const AddMissileParameter &parameter);
-void AddReflection(Missile &missile, const AddMissileParameter &parameter);
-void AddBerserk(Missile &missile, const AddMissileParameter &parameter);
+void AddHiveExplosion(Missile &missile, AddMissileParameter &parameter);
+void AddFireRune(Missile &missile, AddMissileParameter &parameter);
+void AddLightningRune(Missile &missile, AddMissileParameter &parameter);
+void AddGreatLightningRune(Missile &missile, AddMissileParameter &parameter);
+void AddImmolationRune(Missile &missile, AddMissileParameter &parameter);
+void AddStoneRune(Missile &missile, AddMissileParameter &parameter);
+void AddReflection(Missile &missile, AddMissileParameter &parameter);
+void AddBerserk(Missile &missile, AddMissileParameter &parameter);
 
 /**
  * var1: Direction to place the spawn
  */
-void AddHorkSpawn(Missile &missile, const AddMissileParameter &parameter);
-void AddJester(Missile &missile, const AddMissileParameter &parameter);
-void AddStealPotions(Missile &missile, const AddMissileParameter &parameter);
-void AddManaTrap(Missile &missile, const AddMissileParameter &parameter);
-void AddSpecArrow(Missile &missile, const AddMissileParameter &parameter);
-void AddWarp(Missile &missile, const AddMissileParameter &parameter);
-void AddLightningWall(Missile &missile, const AddMissileParameter &parameter);
-void AddRuneExplosion(Missile &missile, const AddMissileParameter &parameter);
-void AddFireNova(Missile &missile, const AddMissileParameter &parameter);
-void AddLightningArrow(Missile &missile, const AddMissileParameter &parameter);
-void AddMana(Missile &missile, const AddMissileParameter &parameter);
-void AddMagi(Missile &missile, const AddMissileParameter &parameter);
-void AddRing(Missile &missile, const AddMissileParameter &parameter);
-void AddSearch(Missile &missile, const AddMissileParameter &parameter);
-void AddCboltArrow(Missile &missile, const AddMissileParameter &parameter);
-void AddLArrow(Missile &missile, const AddMissileParameter &parameter);
-void AddArrow(Missile &missile, const AddMissileParameter &parameter);
-void AddRndTeleport(Missile &missile, const AddMissileParameter &parameter);
-void AddFirebolt(Missile &missile, const AddMissileParameter &parameter);
-void AddMagmaball(Missile &missile, const AddMissileParameter &parameter);
-void AddTeleport(Missile &missile, const AddMissileParameter &parameter);
-void AddLightball(Missile &missile, const AddMissileParameter &parameter);
-void AddFirewall(Missile &missile, const AddMissileParameter &parameter);
+void AddHorkSpawn(Missile &missile, AddMissileParameter &parameter);
+void AddJester(Missile &missile, AddMissileParameter &parameter);
+void AddStealPotions(Missile &missile, AddMissileParameter &parameter);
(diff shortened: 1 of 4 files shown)


Source/missiles.h
```diff
@@ -97,7 +97,8 @@ struct Missile {
 	MissilePosition position;
 	int _mimfnum; // The direction of the missile (direction enum)
 	int _mispllvl;
-	bool _miDelFlag; // Indicate whether the missile should be deleted
+	bool _miDelFlag;     // Indicate whether the missile should be deleted
+	bool _miSpellFizzle; // Indicate whether a spell-spawned missile should consume mana
```
obligaron, line 101: `_miSpellFizzle` is only used in `AddGuardian/AddBersersk/...` and `CastSpell` or?

`CastSpell` -> `AddMissile` -> `AddGuardian` and when `AddMissile` returns `CastSpell` checks `_miSpellFizzle`.

We could avoid this member in `Missile` and instead add it to `AddMissile` and `AddMissileParameter`.
`AddMissile` would need to return spell fizzle (for example a new function overload).
`AddMissile` would get spell fizzle from `AddMissileParameter` (there it is added as member).
`AddGuardian/AddBersersk/...` could set this new member in `AddMissileParameter`.

What do you think? 🙂

obligaron (commented):
I like this changes a lot. 👍🎉 
It makes the code much more readable and decouples missiles from player specific behavior (consume spell). 🙂
I have only some small open quests left (see comments).

Source/objects.cpp
```diff
@@ -1959,6 +1959,8 @@ void OperateBook(Player &player, Object &book)
 		    player.getId(),
 		    0,
 		    0);
+		player.executedSpell.spellType = RSPLTYPE_SPELL;
+		ConsumeSpell(player, SPL_GUARDIAN);
```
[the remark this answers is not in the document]
StephenCWills: Yeah, I was struggling with that question myself. There's a good chance this code simply called `UseMana()` as a side-effect of calling `AddMissile()`, and it's entirely possible that it could have consumed a staff charge or scroll as a result indicating that this was probably not intended behavior. Furthermore, the code always casts slvl 0 Guardian from the Bone Chamber book regardless of the player's spell level, which seems to suggest that it shouldn't really be the player's Guardian. I was kinda thinking the mana loss could still be justified as the cost of learning more about Guardian, but in hindsight that intent is obviously not reflected in the code and would be inconsistent with every other spell book in the game.

I suppose I would ultimately agree that the call to `ConsumeSpell()` should be removed here.

Source/missiles.h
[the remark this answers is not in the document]
StephenCWills: I have no objections, but if we end up moving `_miSpellFizzle` to `AddMissileParameter`, I'll probably just leave this struct alone.

Source/objects.cpp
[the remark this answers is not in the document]
obligaron: I agree. This makes more sense and we also got better code. 😉

Source/missiles.h
StephenCWills: It is true that the field is only used by `CastSpell()` to determine whether the spell cost should be applied. However, I would also argue that there should be other contexts in which it might be useful to know whether a spell fizzled, like maybe a fallback or retry with different constraints. It's possible that no such mechanics were developed for Diablo or Hellfire because Blizzard didn't develop a mechanism like this one for returning that information to the caller.

I wasn't sure about putting it in `AddMissileParameter` because that gets passed by const reference. In order to return the value of spell fizzle via `AddMissileParameter`, we'd have to pass by mutable reference, which isn't a big deal except maybe semantically. The const reference suggests that `AddMissileParameter` was intended for passing information into the `AddMissile()` function rather than for getting information out.

That said, I know that the const qualifier is sometimes applied simply because there is currently no use-case for mutability. Naturally, we can change the semantics if we need to, but since the `AddMissile()` function already returns a reference to the missile and `missile->_miSpellFizzle` makes some good sense semantically, this is ultimately what I settled on. If there's a compelling reason to change it then I could certainly be convinced, but a function overload with a different return value does sound considerably less intuitive to me.

obligaron: I added `AddMissileParameter` in #3957 and at that time I used `const` cause there was no use-case for mutability. 😉

I suggested the change, cause I thought about save game compatibility (the new flag is not part of the save game). Then I saw every time we set `_miSpellFizzle` we also set `_miDelFlag`. That means the missile don't life very long and I thought this new flag is more part of the missile creation then part of the missile behavior.
But yeah a new overload is also not very nice and intuitive.

Alternative idea:
Currently `AddMissile` only returns `nullptr` when we have more missiles then `size_t` allows (this is a theoretically limit. before that we would run out of memory).
Could we change `AddMissile` to return a `nullptr` if the missile creation fails / fizzle?
That way we don't need a new overload and `AddMIssile` would make more sense.
We could change the `AddGuardian/AddBersersk/...` to return a bool value (or alternative change `AddMissileParameter`).

🤔

StephenCWills: That alternative sounds pretty good to me. I guess I'd be more inclined to change `AddMissileParameter` mutability than return a bool since most `AddMissile()` functions don't have conditions in which they could fizzle.

obligaron: I had the same thought (that's why I stated the alternative).
But I was unsure what's more ideal.
Adding a return bool would be cleaner in some way "creation failed or not".
On the other hand using `AddMissileParameter` we would only need to change methods that are affected by it. And I'm unsure if a return bool is really more intuitive, cause we have some missiles that are directly deleted (for example golem or identify. In this case a return false would also make sense in a way (cause we don't need a `Missile` object). That means a fizzle member in `AddMissileParameter` would make it more cleaner. 🤔 
That's why I'm also in favor of using `AddMissileParameter` (but not a strong opinion).

StephenCWills: > And I'm unsure if a return bool is really more intuitive, cause we have some missiles that are directly deleted (for example golem or identify. In this case a return false would also make sense in a way (cause we don't need a `Missile` object).

Yeah, I 100% agree that the bool return value could be ambiguous. The reason I find the `nullptr` return value to be more acceptable is because the intent is clear from reading the code in the `AddMissile()` function. If we had every specialization return a bool instead, then the meaning of that bool would only be apparent when reading the code of certain functions and not others. This could perhaps be mitigated with comments, but I think the `AddMissileParameter` approach is more effective.

Anyway, I made the change so feel free to review and see how you like it.

obligaron: I looked at the code and I think it's good. 🙂 
Even if we want to change it later, it should be fairly easy.

obligaron (approved):
lgtm 🙂

PR #5423, opened 2022-10-21
Network drop Items: Always use passed coordinates for item position

### LINKED ISSUE #4124: Desync when killing a player with 18+ gold stacks

obligaron:
Fixes #4124

When syncing dropped items `SyncPutItem` is used on the remote client.
`SyncPutItem` used `FindAdjacentPositionForItem` and the player position to get the coordinates of the new item.
`FindAdjacentPositionForItem` use only the adjacent positions relative to the player.
If no space is free the item is not placed on the remote client.
This is okay for normal player drops, cause the same check is done in `LeftMouseDown` or `TryDropItem`.

But when items are dropped on death `FindAdjacentPositionForItem` is not used (see `DeadItem`).
That allows dropping on death to occupy more tiles then `FindAdjacentPositionForItem`.
But when syncing with other clients the remote client uses `SyncPutItem` and `SyncPutItem` only allows adjacent tiles.
That can result in desyncs.

The bugfix is that `SyncPutItem` don't use `FindAdjacentPositionForItem` and instead get the desired tile from the remote client.
To further ensure no item desyncs happens, we use the same logic when loading a level (a call to `GetItemPosition` that looks for the next free tile).

Last commit combines `SyncPutItem` and `SyncDropItem` and fixes the issue also for dropped ears.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -1981,14 +1981,13 @@ bool TryDropItem()
 		}
 	}
 
-	Point position = myPlayer.position.future;
-	Direction direction = myPlayer._pdir;
-	if (!FindAdjacentPositionForItem(position, direction)) {
+	std::optional<Point> itemTile = FindAdjacentPositionForItem(myPlayer.position.future, myPlayer._pdir);
+	if (!itemTile) {
 		myPlayer.Say(HeroSpeech::WhereWouldIPutThis);
 		return false;
 	}
 
-	NetSendCmdPItem(true, CMD_PUTITEM, position + direction, myPlayer.HoldItem);
+	NetSendCmdPItem(true, CMD_PUTITEM, *itemTile, myPlayer.HoldItem);
 	myPlayer.HoldItem.clear();
 	NewCursor(CURSOR_HAND);
 	return true;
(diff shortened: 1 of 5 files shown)


StephenCWills (approved):
This is my new favorite PR.

Issue #4491, opened 2022-04-19
Items deleted on pickup


owallace:
**Important information**
Win10, 1.4.0, downloaded myself.

**Describe**
Occasionally, items which are either on the ground or in my inventory are being deleted. It only happens after I click on the item a 2nd time, the first seems to have no effect. This has happened with items on the ground my friend had dropped or a book in my inventory I attempted to learn.

**To Reproduce**
Sadly I cannot consistently reproduce this issue.
1. Drop item on ground
2. Attempt to pick up (left-click)
3. Attempt to pick up a second time, item then disappears.

**Additional context**
Not sure if this is a core D1 issue. but it may be related to network problems.

owallace:
Any input on this by chance? At this point I'm just curious. 
Was playing a TCP game with 2 other friends, a similar situation occurred but this time the item was delete after attempting to place it in the same inventory slot.
My friend clicked on his head piece, cursor turned to an hour glass for a second, then when he tried to reequip it the item disappeared.  Would ZeroTier be less prone to this? Just trying to get some friends into the game.

StephenCWills:
What you're describing here is potentially multiple issues. The easiest way to reproduce the issue of items disappearing from the inventory is using the shops.

1. Select an item for purchase that, when bought, will be placed into your inventory (not on your belt)
2. Wait for the hourglass to appear, then disappear
3. Buy the item
4. Go into your inventory and click on the item

Here is a video where I forced a lag spike to reproduce the issue.

https://user-images.githubusercontent.com/9203145/168489685-67aa358c-9a28-4a72-b2d2-8a8bd8c4f8aa.mp4

This particular issue has already been fixed after the 1.4.0 release, but I haven't yet reproduced the issue of items disappearing from the ground. The closest I've gotten is using debug commands to spawn items, which causes the item to exist on one client and not the other. It's possible there may be edge cases involving deltas with players going back and forth between the dungeon that would cause this type of scenario, but I haven't been able to reproduce it by simply dropping an item and picking it back up.

AJenbo:
It looks to me like that situation might have already been possible in the original

StephenCWills:
I just noticed this is still open. I believe this has been fixed with #5423.

PR #5424, opened 2022-10-21
TimeoutCursor: Only restore previous cursor when cursor is not already changed

obligaron:
Contributes to #4491 (or fixes it? 🤔)

How to reproduce #4491 in master
- Try to pick up a item as a remote client (OnRequestItem).
- Add a sleep in OnRequestItem.
- Lag => `TimeoutCursor(true)` is called. It stores current hand cursor in `sgnTimeoutCurs` and sets hourglass cursor.
- Lag is gone => Messages gets handled => `OnGetItem` is called => `OnGetItem` sets item cursor => `TimeoutCursor(false)` is called and calls NewCursor with the stored cursor (`sgnTimeoutCurs` with the stored hand cursor) => `NewCursor` clears `HoldItem` => Item is gone.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1324,7 +1324,11 @@ void TimeoutCursor(bool bTimeout)
 		}
 		scrollrt_draw_game_screen();
 	} else if (sgnTimeoutCurs != CURSOR_NONE) {
-		NewCursor(sgnTimeoutCurs);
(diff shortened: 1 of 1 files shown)


StephenCWills (approved):
Really good catch. I had assumed that the hourglass couldn't be overwritten during a timeout because the game short-circuits the game logic, but the network packet handler could certainly do it.

PR #5426, opened 2022-10-22
Only color cycle animations if game logic is processed

### LINKED ISSUE #5282: Lava animation goes crazy during hourglass
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

seems like hell's lava color cycling animation gets broken during hourglass - it cycles every frame = looks super sped up and bad

### To Reproduce

get hourglass
watch lava in hell

### Expected Behavior

_No response_

### Additional context

https://clips.twitch.tv/SmoggyEntertainingDragonfruitChocolateRain-gxCq_qMb6-wTpRYh - 0:50

obligaron:
Fixes #5282

When a lag occurs the game tries to run `game_loop`. But fails to do so (cause of the lag).
Then the game is drawn and in the next main loop it tires to run `game_loop` again (and fails).

The problem is that at every try it executes `diablo_color_cyc_logic`. That's why the animations updates very fast/go crazy.

Solution: Only call `diablo_color_cyc_logic` when the game loop is successfully executed. If not don't update the animations (similar to game pause).

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -765,9 +765,9 @@ void RunGameLoop(interface_mode uMsg)
 			continue;
 		}
 
-		diablo_color_cyc_logic();
 		multi_process_network_packets();
-		game_loop(gbGameLoopStartup);
+		if (game_loop(gbGameLoopStartup))
+			diablo_color_cyc_logic();
 		gbGameLoopStartup = false;
 		if (drawGame)
 			DrawAndBlit();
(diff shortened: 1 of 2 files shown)


Trihedraf:
Is this the same thing that happens on mobile in multiplayer when you switch away from DevX and then come back?

StephenCWills:
If you're talking about the game running at super-speed in solo Multiplayer games after pausing the logic for some period of time, then it's not the same thing.

StephenCWills (approved):

AJenbo (approved):

PR #5429, opened 2022-10-22
Avoid dramatically speed up in solo multiplayer

obligaron:
Fixes speed up on android when doing solo multiplayer and switching apps (see [comment](https://github.com/diasurgical/devilutionX/pull/5426#issuecomment-1287829778)).

### DIFF
--- a/Source/nthread.cpp
+++ b/Source/nthread.cpp
@@ -216,9 +216,23 @@ bool nthread_has_500ms_passed()
 {
 	int currentTickCount = SDL_GetTicks();
 	int ticksElapsed = currentTickCount - last_tick;
-	if (!gbIsMultiplayer && ticksElapsed > gnTickDelay * 10) {
-		last_tick = currentTickCount;
-		ticksElapsed = 0;
+	// Check if we missed multiple game ticks (> 10)
+	if (ticksElapsed > gnTickDelay * 10) {
+		bool resetLastTick = true;
+		if (gbIsMultiplayer) {
+			for (size_t i = 0; i < Players.size(); i++) {
+				if ((player_state[i] & PS_CONNECTED) != 0 && i != MyPlayerId) {
+					// Reset last tick is not allowed when other players are connected, cause the elapsed time is needed to sync the game ticks between the clients
+					resetLastTick = false;
+					break;
+				}
+			}
+		}
+		if (resetLastTick) {
+			// Reset last tick to avoid caught up of all missed game ticks (game speed is dramatically increased for a short time)
+			last_tick = currentTickCount;
+			ticksElapsed = 0;
+		}
 	}
 	return ticksElapsed >= 0;
 }


Source/nthread.cpp
```diff
@@ -216,9 +216,23 @@ bool nthread_has_500ms_passed()
...
-	if (!gbIsMultiplayer && ticksElapsed > gnTickDelay * 10) {
-		last_tick = currentTickCount;
-		ticksElapsed = 0;
+	// Check if we missed multiple game ticks (> 10)
+	if (ticksElapsed > gnTickDelay * 10) {
+		bool resetLastTick = true;
+		if (gbIsMultiplayer) {
+			for (size_t i = 0; i < Players.size(); i++) {
+				if ((player_state[i] & PS_CONNECTED) != 0 && i != MyPlayerId) {
+					// Reset last tick is not allowed when other players are connected, cause the elapsed time is needed to sync the game ticks between the clients
```
obligaron, line 225: Note:
Locally I tested that the change also works with connected clients but I was unsure if my tests are really correct or if I missed something. That's why I only enabled it for solo multiplayer.
For the reported use-case (switching apps) fixing solo multiplayer should be good enough, cause in multiplayer with other clients there would be a timeout anyway.

Trihedraf:
Works on iOS 👍

Source/nthread.cpp
StephenCWills: Yeah, I wouldn't expect this to have any real effect on non-solo multiplayer. The `nthread_recv_turns()` function updates `last_tick` as soon as the hourglass resolves.

https://github.com/diasurgical/devilutionX/blob/d0c1aba463fef3444e0a0be09dffc17f47ad43cb/Source/nthread.cpp#L121-L132

I feel like we can get rid of `sgbTicsOutOfSync` by replacing that logic with what you have here. What do you think?

StephenCWills: Wait, actually, the purpose of this might be to force `nthread_has_500ms_passed()` to return true on every frame so that `game_loop()` can rapidly check if the missing turns have arrived. I guess it might be a bit more complicated to replace the `nthread_recv_turns()` logic.

obligaron: I also was unsure that's why I didn't touch that part.
For now I think this pr is good enough, cause we don't have extreme speed ups in lag situations.
My thought was to get single- and multiplayer to behave similar when playing solo multiplayer.

AJenbo (approved):

AJenbo:
nice with the comments, these functions are still a bit of voodoo

AJenbo:
nthread_has_500ms_passed also still has a name based on early guesswork and is missleading

AJenbo:
something that would also be worth looking in to is allowing the rendering to skip if we are behind on logic frames. This would prevent the game from looking odd when its behind, catch up faster, and work on lowend systems that cant sustain 20fps

obligaron:
> nice with the comments, these functions are still a bit of voodoo

For me the functions are also a little bit voodoo. 😉 So I hope all added comments are correct.

> nthread_has_500ms_passed also still has a name based on early guesswork and is missleading

I think there are more misleading variables and at least I added one 😟 (for example `last_tick` is more the next tick).

> something that would also be worth looking in to is allowing the rendering to skip if we are behind on logic frames. This would prevent the game from looking odd when its behind, catch up faster, and work on lowend systems that cant sustain 20fps

I think the game currently tries this.
In `game_loop` there is loop that processes 3 game ticks for every rendering (if it's behind and if it's multiplayer).
I think they used this logic so the game still updates (visual feedback) and responds to input (menu etc.).

PR #5443, opened 2022-10-28
Display correct Armor Class in Char Panel

kphoenix137:
In the hit calculation for a monster hitting a player, the player level is used, which equates to the player always having a 2 * clvl increase to what is displayed in the character panel. This adds it in.


AJenbo:
Amiga build is failing since this PR was based on an older version of the code

kphoenix137:
> Amiga build is failing since this PR was based on an older version of the code

Resolved

StephenCWills (approved):
In the PvE calculations, this is traditionally thought of as part of a computation where player level is pitted against monster level: `2 * (clvl - mlvl)`. However, while mlvl is variable based on the monster you're fighting, clvl is pretty static so it should be appropriate to factor this out and display it. Essentially, this gives you that sense that each clvl provides a +2 bonus against monster hit chance across the board.

glebm (approved):

AJenbo:
@StephenCWills thanks for the extra details :+1:

PR #5444, opened 2022-10-28
Minor cleanup for MonsterAttackPlayer()

kphoenix137:


Source/monster.cpp
```diff
@@ -1126,13 +1126,17 @@ void MonsterAttackPlayer(Monster &monster, Player &player, int hit, int minDam,
...
-	int minhit = 15;
-	if (currlevel == 14)
-		minhit = 20;
-	if (currlevel == 15)
-		minhit = 25;
-	if (currlevel == 16)
+	int minhit;
+	switch (currlevel) {
+	case 16:
 		minhit = 30;
```
glebm, line 1132: Each `case` branch needs a `break;`

kphoenix137: Oh my god lmao
I've been doing python for the past 2 months. Already getting rusty on c++

glebm: Alternativately, extract it into a function and replace assignment with `return`. Then there is no need for `break;`.

AJenbo: I wouldn't mind doing somethings in Rust :)

PR #5446, opened 2022-10-29
Fix compilation warnings

glebm:


AJenbo (approved):
Nice with the external lib warnings we might be able to fix all and treat them as errors

glebm:
There are still a ton of warnings from libzt that we can't silence because of their poorly written CMake file.
Also one warning from asio that we may be able to silence.

AJenbo:
I guess we could make a seperate CI build for that with TZ disabled

PR #5453, opened 2022-10-30
Overhaul backbuffer state handling

### LINKED ISSUE #5447: Rendering issue on rg350m
PXL_20221029_230854552

I'm guessing double-buffering-related

glebm:
When rendering directly to the output buffer, we need to maintain the state of what has been drawn and what needs redrawing per-buffer.

We previously tried to do it implicitly by checking `SDL_DOUBLEBUF` and other flags. The previous implementation was broken in several ways, resulting in rendering issues on devices that support 8-bit output directly.

Changes this mechanism to explicitly maintain buffer state per output buffer. The new mechanism doesn't require knowledge of the number of buffers, and thus also works correctly with triple-buffering.

Fixes #5447

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -96,6 +96,7 @@ set(libdevilutionx_SRCS
   engine/actor_position.cpp
   engine/animationinfo.cpp
   engine/assets.cpp
+  engine/backbuffer_state.cpp
   engine/direction.cpp
   engine/dx.cpp
   engine/load_cel.cpp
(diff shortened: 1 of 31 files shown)


AJenbo:
It's always a bit mind boggling when a time demo fails because of rendering changes :sweat_smile:

glebm:
Fixed! It was segfaulting due to `PalSurface` being `nullptr` in headless mode.

AJenbo (approved):

PR #5455, opened 2022-10-31
Fix mouse simulation

### LINKED ISSUE #5449: Select + D-Pad mouse simulation broken
1. Mouse moves in wrong directions and doesn't stop when d-pad button is released (while holding select).
2. The SELECT menu reappers when d-pad button is released (while holding select).

### LINKED ISSUE #5452: gamepad: Right-stick mouse cursor disappears on right stick click
Right stick-driven mouse cursor disappears on click with the right stick.
E.g. picking up a potion from the belt results in no mouse cursor until the next cursor movement.

StephenCWills:
The interactions here are so complex that it's difficult to summarize this. I hope, when paired with the commit messages, that the following bullet points should suffice.

* `PadmapperOptions::ActionNameTriggeredByButtonEvent()` is intended for use by mouse simulation functions because these run during input detection, before `PadmapperOptions::ButtonPressed()`
* `PadmapperOptions::IsActive()` is intended for basically anything else that runs after the `PadmapperOptions::ButtonPressed()`
* `SimulateRightStickWithDpad()` returned a value to indicate whether the pad button input represents a gamepad movement event in order to shortcut further processing. That return value was removed from `SimulateRightStickWithPadmapper()` because it now uses `PadmapperOptions::IsActive()` to determine whether other mouse movement actions were activated by previous events. In order for that to work, those button events need to make it through to the padmapper to activate those mouse movement actions.
* `GetLeftStickOrDpadDirection(false)` is intended for use with stores, in-game menus, and char/inv/etc panels. The `false` flag overrides the padmapper actions for moving the character so that the d-pad works as you might expect in these menus, but mouse simulation actually trumps these controls.
* `IsAutomapActive()` was removed because there is now a valid automap in town so it doesn't need to be excluded

This resolves #5449
This resolves #5452

### DIFF
--- a/Source/controls/controller_motion.cpp
+++ b/Source/controls/controller_motion.cpp
@@ -81,26 +81,32 @@ void SetSimulatingMouseWithPadmapper(bool value)
 // SELECT + D-Pad to simulate right stick movement.
 bool SimulateRightStickWithDpad(ControllerButtonEvent ctrlEvent)
 {
+	if (IsAnyOf(ctrlEvent.button, ControllerButton_NONE, ControllerButton_IGNORE))
+		return false;
+
 	ControllerButtonCombo upCombo = sgOptions.Padmapper.ButtonComboForAction("MouseUp");
 	ControllerButtonCombo downCombo = sgOptions.Padmapper.ButtonComboForAction("MouseDown");
 	ControllerButtonCombo leftCombo = sgOptions.Padmapper.ButtonComboForAction("MouseLeft");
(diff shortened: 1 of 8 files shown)


Trihedraf:
@StephenCWills This broke the touch buttons again.

StephenCWills:
Argh, it's probably this.

https://github.com/diasurgical/devilutionX/blob/680ab5ec40fcc2078ed6ad4a48bf081e68b1c9a1/Source/controls/game_controls.cpp#L366-L368

I forgot that `GetGameAction()` handles touch events too. Thanks for finding these so quickly each time.

PR #5465, opened 2022-11-04
Update japanese translations.

bubio:
- Corrected some translations
- Translated the added wording


AJenbo (approved):
Nice :)

AJenbo:

PR #5469, opened 2022-11-06
AnimationInfo: Use fixed point math for fractions

obligaron:
Addressed glebm's [comment](https://github.com/diasurgical/devilutionX/issues/4849#issuecomment-1173046136).

Uses fixed point math for fractions (game ticks, animations, ...).

The pros:
- reduces size of `AnimationInfo` in 32bit builds (from 20 to 16 bytes)
- better performance for devices that don't have hardware float support

The cons:
- a little bit harder to read
- a fraction can only contain 128 distinct values. That means if you have a device that can show more than 2560fps you could see a difference. 😜


glebm (approved):
Nice!

Issue #4188, opened 2022-03-20
Issues with Cow Quest in multiplayer


AJenbo:
Only the host is able to pick up the gray suit (and probably the brown one as well). Other players are able to see it, but the item will vanish if they attempt to pick it up.

Is probably caused by SpawnQuestItem() not syncing the item with the client, this also affects the Torn Notes and most single player quests. The solution should be similar to what was done in https://github.com/diasurgical/devilutionX/pull/4198

Issue #926, opened 2020-11-21
Enable all quests in Multiplayer, perhaps even Strange Map (Map of the Stars)


sedrosken:
I'm not sure quite how to tag these issues, or if I even can, but I believe this would be filed under "enhancement," perhaps even a long-term one.

In the original version of Diablo, due to time crunch, only four quests were enabled in Multiplayer mode: The Butcher, The Skeleton King, Archbishop Lazarus, and the final confrontation with Diablo himself. In the beta I believe it was said that all quests were enabled, but due to syncing issues (that I believe could have been solved had they had more time -- according to Diablo Evolution, the beta was mid-November of 1996, and the gold master disk was pressed on 27 December 1996) were subsequently disabled on release. I'm not sure why Blizzard themselves didn't patch this as time went on, maybe it was just the fact that it wasn't a bug but a choice made in the face of last-minute development issues, and back then, only bugs got patched. But I don't believe any but the absolute most hardcore of Diablo purists would mind these quests being enabled.

Another quest that I'd like to see revisited is the Strange Map quest from beta and earlier. That might be too big an ask, maybe MPQ modification would be needed there (which I understand is outside the scope of this particular project) as I bet a lot of related data was removed from the retail MPQ -- aside from the item itself, in a relatively functional state, that can be spawned in using cheats... For some background, Blizzard decided just before release that it caused a sense of panic and pacing issues as if you waited too long to face Diablo you'd be facing a stronger version of him. But I believe we're up to the challenge.

AJenbo:
Any quest that contains special maps do not work in multiplayer as the game is really confused bout where the player is when entering thease. It probably won't take to much work to fix and we would probably do it eventually.

sedrosken:
I hadn't thought of that. That's the main reason Leoric is encountered on dlvl 3 and Lazarus on dlvl 15 in multiplayer instead of in their own levels as in single player, right?

AJenbo:
correct

PR #5471, opened 2022-11-06
Start enabling singleplayer quests for multiplayer

obligaron:
Contributes to #926

With #4738 (and it's successors) quest/set-maps are multiplayer compatible.
This pr now starts with the ground work for syncing all quests and enabling two singleplayer quests for multiplayer (Lazarus and Poisoned Water Supply).

Notes:
- `UseMultiplayerQuests()` was introduced to switch between single- and multiplayer quests (In theory it could also be used to play the multiplayer quests in singleplayer 😉). Previous this was done with `gbIsMultiplayer` checks, but now that we start supporting singleplayer quests in multiplayer this is insufficient. Currently `UseMultiplayerQuests()` is identically to `gbIsMultiplayer`, but later it will be likely replaced by a setting that is synced between clients (like missile friendly fire). For my testing i replaced it with a hardcoded false. 😉 Side-note: This make understanding the code easier, cause previous many `gbIsMultiplayer` checks are quests specific and not really multiplayer specific.
- Some syncing of map changing is necessary that also didn't happen in singleplayer. For example: one player kills lazarus and another player stands in dungeon level 15 at the pentagram the pentagram must be updated on the fly (and not at level load time). Another example: the poisoned water fountain in town must be cleared if another player kills all monsters. Currently `ResyncQuests` is used for that.
- For other quests some general issues need to be fixed (for example spawning unique items from quests #4188)
- This pr is not 100% tested. There should be some edge-cases missing. But I hope this is a good start.
- After the pr is merged, it is hopefully possible to work independent and parallel on different quests. I killed Lazarus and the monsters in poisioned water supply a lot already 🥴, so any help is more then welcome. 🙂

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -2745,10 +2745,10 @@ void LoadGameLevel(bool firstflag, lvl_entry lvldir)
 				for (int y = 0; y < DMAXY; y++)
 					UpdateAutomapExplorer({ x, y }, MAP_EXP_SELF);
 		}
-		if (!gbIsMultiplayer)
-			ResyncQuests();
-		else
+		if (UseMultiplayerQuests())
 			ResyncMPQuests();
+		else
+			ResyncQuests();
 	} else {
 		LoadSetMap();
 		IncProgress();
@@ -2791,8 +2791,11 @@ void LoadGameLevel(bool firstflag, lvl_entry lvldir)
 		} else {
 			LoadLevel();
 		}
-		if (gbIsMultiplayer)
+		if (gbIsMultiplayer) {
 			DeltaLoadLevel();
+			if (!UseMultiplayerQuests())
+				ResyncQuests();
+		}
 
 		InitMissiles();
 		IncProgress();
--- a/Source/levels/drlg_l4.cpp
+++ b/Source/levels/drlg_l4.cpp
@@ -1112,10 +1112,10 @@ bool PlaceStairs(lvl_entry entry)
 		}
 	} else {
 		// Place hell gate
-		bool isGateOpen = UseMultiplayerQuests() || Quests[Q_DIABLO]._qactive == QUEST_ACTIVE;
-		position = PlaceMiniSet(isGateOpen ? L4PENTA2 : L4PENTA);
+		position = PlaceMiniSet(L4PENTA2);
 		if (!position)
 			return false;
+		Quests[Q_DIABLO].position = *position;
 		if (entry == ENTRY_PREV)
 			ViewPosition = position->megaToWorld() + Displacement { 6, 5 };
 	}
@@ -1180,6 +1180,10 @@ void GenerateLevel(lvl_entry entry)
 	DRLG_CheckQuests(SetPieceRoom.position);
 
 	if (currlevel == 15) {
+		bool isGateOpen = UseMultiplayerQuests() || Quests[Q_DIABLO]._qactive == QUEST_ACTIVE;
+		if (!isGateOpen)
+			L4PENTA.place(Quests[Q_DIABLO].position);
+
 		for (WorldTileCoord j = 1; j < DMAXY; j++) {
 			for (WorldTileCoord i = 1; i < DMAXX; i++) {
 				if (IsAnyOf(dungeon[i][j], 98, 107)) {
--- a/Source/levels/drlg_l4.cpp
+++ b/Source/levels/drlg_l4.cpp
@@ -1184,10 +1184,8 @@ void GenerateLevel(lvl_entry entry)
 			for (WorldTileCoord i = 1; i < DMAXX; i++) {
 				if (IsAnyOf(dungeon[i][j], 98, 107)) {
 					Make_SetPC({ WorldTilePosition(i - 1, j - 1), { 5, 5 } });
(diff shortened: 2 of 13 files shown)


AJenbo (approved):
A good start indeed :)

Caesar0007:
Please, can someone provide a compilled pre-relese 1.5 version for multiplayer quest testing. thanks.

StephenCWills:
We make compiled test builds available automatically. See: https://github.com/diasurgical/devilutionX#test-builds

obligaron:
Even in compiled test builds the feature is not active.

Reasons:
- Not all quests are implemented yet. But they are still active. Playing those broken quests can result in a broken game state.
- We need to discuss if we enable this features per default.
- We need to discuss how we handle quest items.
- We need to discuss how we handle unique drops.

The downside is that you can only test it when compiling a version for yourself. 😞

AJenbo:
Basically it's not ready for general testing yet :)

Caesar0007:
22 years ago I played Hellfire 1.01 via Rockwell 14.4 with my friend. We tried to dublicate items - for example - Rings +20 to all params. I they dropped to earth - game just kill one of them.
Also, I was able to import into Vein of Steel to my multiplayer hero via memory editing. It function without any problems.

Caesar0007:
1. We need to discuss if we enable this features per default => I played Hellfire a lot and always missed a single player quests in multiplayer mode. If this function will be turnable in game menu, that will be great.
2. We need to discuss how we handle quest items => is there any difference from single player? I checked many items from Single in Multiplayer - no problems was found, at least at Hellfire 1.01. Only one item can be dropped per game session - like a bomb for nest and used by one of the players.
3. We need to discuss how we handle unique drops => is there any conflict here? Each time Boss, that must drop uniq item - drop it. I think a possible solution is to check, If boss uniq item is present currently in players stash. If not = 100% drop.
p.s. this is just my personal opinion.

Caesar0007:
if possible - can someone provide latest test build with quest enabled in multiplayer. just for my personal testing. I will backup my save before use it. thanks.

obligaron:
# Enable singleplayer quests
I think this will get a setting to toggle it.
For me the only question is: should we enable it per default or not? 😉 

# Quest Items
You can carry them from one game to the next and have them multiple times in your inventory.
This can happen frequently in multiplayer.
Is this a good thing?
Should it stay this way or (for example) should remove them when joining a new game?

Another thing: If a player with a quest leaves the game the quest can't progress.
Should it stay this way or (for example) should we drop the item in town so other players can process the quest?

# Quest rewards (unique items)
Some items are (at the start9 powerful. Should we drop them?
Should we make replaying a quest more interesting? (For example drop the quest reward randomly only in x% of the drops?)

This is also my personal thoughts, that's why I'm also saying we need to discuss this further. 🙂

Caesar0007:
1. Enabled by default is the best option
2. if this item, that was inteded only for quest and can broke something => players must be checked and item must be removed before entering the session
3. I think, that they must be droped as was designed or at least you can do an option for this.

AJenbo:
> > We need to discuss how we handle unique drops
> is there any conflict here? Each time Boss, that must drop uniq item - drop it. I think a possible solution is to check, If boss uniq item is present currently in players stash. If not = 100% drop.

Why have it work differently then it does in single player?

> > if this item, that was inteded only for quest and can broke something
> players must be checked and item must be removed before entering the session

Again this breaks with how it's handled in Single Player. Lets avoid such arbitrary changes to the game.

>  For me the only question is: should we enable it per default or not? wink

Maybe I could start working on a better "create game" menu :)

> You can carry them from one game to the next and have them multiple times in your inventory.
This can happen frequently in multiplayer.

You can already do this in Single Player so lets just keep it like that.

> Another thing: If a player with a quest leaves the game the quest can't progress.

We can take a look at how D2 handels situations like this. But my feeling is that we should leave it like that.

StephenCWills:
> We can take a look at how D2 handels situations like this. But my feeling is that we should leave it like that.

If someone takes off with Ogden's Sign, it's not that big of a deal since you can just skip to dlvl 5 with the town entrance. If someone takes off with Staff of Lazarus, though, that could be annoying.

AJenbo:
Or drop by just to take the rune bomb before you are high enough level to access it your self...

Caesar0007:
Please! Someone, how have a knowledge to do this! Is it possible to compile and send to me latest beta with quest enabled in multiplayer?

StephenCWills:
It's absolutely inappropriate to beg for a build in a merged PR. We've already explained why the feature is not currently enabled. Please just be patient and wait for the feature to be completed.

obligaron:
> Maybe I could start working on a better "create game" menu :)

That would be great. 🙂👍 

> You can already do this in Single Player so lets just keep it like that.

I don't think this was a explicit decision from the orginal devs, cause they don't work correctly when reusing them in a new game.
You can't continue the quest where you left of.
You can get them multiple times.
For example: take ogden's sign from one game to the next. You can't return ogden's sign until you visited level 3. If you return it and visit later level 4, you see the path is open and the monsters attack you. But snotspill does nothing while his comrades get killed. 😉

> We can take a look at how D2 handels situations like this. But my feeling is that we should leave it like that.

D2 remembers the quest state between games.
We could go this route but this would be different from singleplayer and also much work, cause we need to handle (like D2) different player quest progress.

Issue #5474, opened 2022-11-07
Stash vector out of bounds from spurious loop?


StephenCWills:
### Discussed in https://github.com/diasurgical/devilutionX/discussions/5473

<div type='discussions-op-text'>

<sup>Originally posted by **hateborne** November  7, 2022</sup>
_(Hitting vector out of bounds issue with underflow itemId, which is what started all of this.)_

Why is the [InvDrawSlotBack](https://github.com/diasurgical/devilutionX/blob/680ab5ec40fcc2078ed6ad4a48bf081e68b1c9a1/Source/inv.cpp#L1020) call handled in a separate loop ([stash.cpp:355](https://github.com/diasurgical/devilutionX/blob/069ba7f1d4884a7abbe686db8827e1d97268ff1f/Source/qol/stash.cpp#L355)) instead of the loop right below it ([stash.cpp:363](https://github.com/diasurgical/devilutionX/blob/069ba7f1d4884a7abbe686db8827e1d97268ff1f/Source/qol/stash.cpp#L363))? The validation seems to be handled in the second loop and the first loop seems (from a relative outsider) problematic. 

Example of lines 355 - 381:
```cpp
	for (auto slot : StashGridRange) {
		StashStruct::StashCell itemId = Stash.GetItemIdAtPosition(slot);
		if (itemId == StashStruct::EmptyCell) {
			continue; // No item in the given slot
		}

		Item &item = Stash.stashList[itemId];

		if (Stash.IsItemAtPosition(slot)) {
			InvDrawSlotBack(out, GetStashSlotCoord(slot) + offset, InventorySlotSizeInPixels, item._iMagical);
		}

		if (item.position != slot) {
			continue; // Not the first slot of the item
		}

		int frame = item._iCurs + CURSOR_FIRSTITEM;

		const Point position = GetStashSlotCoord(item.position) + offset;
		const ClxSprite sprite = GetInvItemSprite(frame);

		if (pcursstashitem == itemId) {
			uint8_t color = GetOutlineColor(item, true);
			ClxDrawOutline(out, color, position, sprite);
		}

		DrawItem(item, out, position, sprite);
	}
```
</div>

StephenCWills:
The error is just an oversight introduced by #5266. The loops certainly could be combined. You do not need to check `Stash.IsItemAtPosition()` anymore since this logic is already checking `itemId == StashStruct::EmptyCell`.

PR #5475, opened 2022-11-07
Fix underflow issue on coloring in Stash

hateborne:
The addresses the issue in #5474.

> tl;dr - This moves **_entirely_** away from `StashGridRange`, prioritizing `StashList` as the sole source of authority.

# Stash specific
* All functions in header have doc blocks added, both to aid others in
  contributing and because I am a psychopath.
* Changed usage of itemId/iv to itemIndex where appropriate. As these
  were used interchangably, it took a while to untangle things.
  Anytime the literal index value is used, it should be `itemIndex`.
  In other places where the modified id is used, it should be `itemId`.
* Uses of `StashGridRange` were replaced with a much smaller loop over
  only the visible page's `StashGrid` or the `StashList`.
* `AddItemToStashGrid` was removed in favor of an action agnostic
  function named `UpdateStashGrid`. This can be used to set an itemId
  into the corresponding cells, or simply zero out the cells. The
  important note here is that the item position must be the bottom-left
  cell. The inconsistency of top-left vs bottom-left is what contributed
  to some of the issues with stashList getting salty. The
  `PointsInRectangle` usage always worked right (x) and down (y), which
  worked extremely well if the itemPosition was always passed as the
  bottom-left cell (which is converted within the function before
  changing things).
* `RemoveStashItem` now takes an `Item` pointer, the x/y `Point` of the
  item, and the item's index in the `StashList`. The reason for the
  passed position versus pulling it from the item is that the autoequip
  would change the items out before `RemoveStashItem` was called. This
  resulted in the itemPosition being incorrect and going out of bounds.
* Removed the multiple manual cursor slot detection loops in favor of
  the function that already exists to do just that.

# Misc
* Uniform invalid item id usage: This is an opinionated change to clarify the code at a glance. While it is functionally the same as `uint16_t(-1)`, it feels like a more obvious explanation of what is happening.
* Simple `==` and `!=` operators for the `Item` class.
* Removed the gold withdraw QOL change as it is out of scope for this.


Source/qol/stash.cpp
```diff
@@ -352,21 +352,19 @@ void DrawStash(const Surface &out)
...
 		if (itemId == StashStruct::EmptyCell) {
 			continue; // No item in the given slot
 		}
 
 		Item &item = Stash.stashList[itemId];
+
+		// provide item background coloring
+		if (Stash.IsItemAtPosition(slot)) {
+			InvDrawSlotBack(out, GetStashSlotCoord(slot) + offset, InventorySlotSizeInPixels, item._iMagical);
+		}
```
StephenCWills, line 366: https://github.com/diasurgical/devilutionX/blob/a2a8b1c706789a9f7ddc08c7e4941d69d43690aa/Source/qol/stash.h#L50-L53

This loop is already doing this same validation right at the start.

```suggestion
		InvDrawSlotBack(out, GetStashSlotCoord(slot) + offset, InventorySlotSizeInPixels, item._iMagical);
```

StephenCWills (changes requested):

hateborne:
(both images)

Good catch! It was tested and it looks ok. I have included an example image using the PR iteration, as well as one with the change you suggested. :)

PR version
pr-ver

With your suggested change
pr-with-chgs

StephenCWills:
> Good catch! It was tested and it looks ok. I have included an example image using the PR iteration, as well as one with the change you suggested. :)

Okay, so the reason it looks mostly correct is because the bottom-left corner of the item sprite is the slot at which the item is placed. It means the slot back is only being drawn over the bottom-right corner of the 2x2 and 2x3 items in your stash. You can only really see it with the studded leather in your screenshot, but it becomes pretty obvious if you highlight an item by hovering with the mouse.

image

DakkJaniels:
Shouldn't the two unrelated changes be in separate PR's? (@qndel surprised you didn't comment on that first)

Source/qol/stash.cpp
```diff
@@ -658,7 +654,11 @@ void GoldWithdrawNewText(string_view text)
 			int newGoldValue = WithdrawGoldValue * 10;
 			newGoldValue += digit;
 			if (newGoldValue <= InitialWithdrawGoldValue) {
+				// update to newly set value
 				WithdrawGoldValue = newGoldValue;
+			} else {
+				// lazily 'set & cap' to maximum gold value
+				WithdrawGoldValue = InitialWithdrawGoldValue;
 			}
```
ephphatha, line 662: Can simplify this using `min`

```suggestion
			// Mimic D2 stash behaviour, cap withdrawal amounts to maximum gold value
			WithdrawGoldValue = std::min(newGoldValue, InitialWithdrawGoldValue);
```

StephenCWills:
> Shouldn't the two unrelated changes be in separate PR's?

It would be nice if they were at least separated into different commits, but they're both sufficiently small changes that I don't think it's that big of a deal. Maybe someone else thinks differently, but I'm willing to let this slide.

Source/qol/stash.cpp
```diff
@@ -352,21 +352,17 @@ void DrawStash(const Surface &out)
 
 	constexpr Displacement offset { 0, INV_SLOT_SIZE_PX - 1 };
 
-	for (auto slot : StashGridRange) {
-		StashStruct::StashCell itemId = Stash.GetItemIdAtPosition(slot);
-		Item &item = Stash.stashList[itemId];
-		if (Stash.IsItemAtPosition(slot)) {
-			InvDrawSlotBack(out, GetStashSlotCoord(slot) + offset, InventorySlotSizeInPixels, item._iMagical);
-		}
-	}
```
ephphatha, line 361: Stephen pointed out the rendering issue on items two cells wide which was the reason for the backs being drawn first. A minimally invasive fix would be to move the code that performs the stashList lookup inside the if statement. It's more relevant for the scope that `item` is actually used as well.

If you did want to fix the rendering for inventory items so backs can be drawn in one pass that'd be awesome :D, seems like a fair bit of work required though.

hateborne:
> > Shouldn't the two unrelated changes be in separate PR's?
> 
> It would be nice if they were at least separated into different commits, but they're both sufficiently small changes that I don't think it's that big of a deal. Maybe someone else thinks differently, but I'm willing to let this slide.

I can absolutely do that and apologies for the breach of etiquette/procedure. :)

---

> Stephen pointed out the rendering issue on items two cells wide which was the reason for the backs being drawn first. A minimally invasive fix would be to move the code that performs the stashList lookup inside the if statement. It's more relevant for the scope that `item` is actually used as well.
> 
> If you did want to fix the rendering for inventory items so backs can be drawn in one pass that'd be awesome :D, seems like a fair bit of work required though.

I'll review this in the morning and be more diligent in testing before I pester again.

hateborne:
@StephenCWills  I really did not mean for this simple change to end up as a full blown crusade, but here we are. I have tested in every way that caused a crash previously and was unable to cause any issue (read: 0 crashes, 0 item corruption incidents). 

In addition to the tests below, the highlighting always worked and the background coloring always worked as expected.

```
=== Existing Character, Existing Stash ===
Manually placing medium+ size object in corners and cardinal direction 'borders' on blank page
Manually placing medium+ size object in corners and cardinal direction 'borders' on non-blank page
Manually placing medium+ size object over items in corners and cardinal direction 'borders'
CTRL clicking into blank page
CTRL clicking into non-blank page
CTRL clicking into full page
CTRL clicking out of non-blank page
CTRL clicking out of full page
SHIFT clicking out of non-blank page
SHIFT clicking out of full page
Pull gold out
Manually put gold in
CTRL clicking gold in
Saved game, then exited
Loaded game, confirmed changes from previous save & exit

## Deleted Character & Stash saves

=== New Character, New Stash ===
Manually placing medium+ size object in corners and cardinal direction 'borders' on blank page
Manually placing medium+ size object in corners and cardinal direction 'borders' on non-blank page
Manually placing medium+ size object over items in corners and cardinal direction 'borders'
CTRL clicking into blank page
CTRL clicking into non-blank page
CTRL clicking out of non-blank page
SHIFT clicking out of non-blank page
Pull gold out
Manually put gold in
CTRL clicking gold in
Saved game, then exited
Loaded game, confirmed changes from previous save & exit


=== New Character, Existing Stash ===
Confirmed stash exists in state from previous section
Manually placing medium+ size object in corners and cardinal direction 'borders' on blank page
Manually placing medium+ size object in corners and cardinal direction 'borders' on non-blank page
Manually placing medium+ size object over items in corners and cardinal direction 'borders'
CTRL clicking into blank page
CTRL clicking into non-blank page
CTRL clicking out of non-blank page
SHIFT clicking out of non-blank page
Pull gold out
Manually put gold in
CTRL clicking gold in
Saved game, then exited
Loaded game, confirmed changes from previous save & exit


## Restored original Stash save only
(Both new characters could access stash, save & exit, then load & confirm changes)
```

Source/items.h
```diff
@@ -449,6 +449,16 @@ struct Item {
...
 	void updateRequiredStatsCacheForPlayer(const Player &player);
+
+	bool operator==(const Item &other) const
+	{
+		return this->_iSeed == other._iSeed && this->_iCreateInfo == other._iCreateInfo;
+	}
+
+	bool operator!=(const Item &other) const
+	{
+		return this->_iSeed != other._iSeed && this->_iCreateInfo != other._iCreateInfo;
```
qndel, line 460: shouldn't it be || instead of &&?

Source/qol/stash.cpp
```diff
@@ -494,53 +449,64 @@ bool UseStashItem(uint16_t c)
...
 	}
 
-	// If the item at the end of stash array isn't the one we removed, we need to swap its position in the array with the removed item
+	// get item's size to figure slot changes
+	const Size itemSize = GetInventorySize(*item);
+
+	// reset claimed slot IDs to zero
+	UpdateStashGrid(Stash.GetPage(), itemPosition, itemSize);
+
+	// not the last item, swap with last item to keep IDs contiguous
```
ephphatha, line 476: this comment should move into the if statement

Source/qol/stash.cpp
```diff
@@ -494,53 +449,64 @@ bool UseStashItem(uint16_t c)
...
-						itemId = iv + 1;
-					}
-				}
-			}
-		}
+	if (lastItemIndex != itemIndex) {
+		Point swappedItemPosition = stashList[lastItemIndex].position;
+		const Size swappedItemSize = GetInventorySize(stashList[lastItemIndex]);
+		unsigned swappedItemPage = GetPageByItemId(lastItemIndex);
+		std::swap(stashList[itemIndex], stashList[lastItemIndex]);
```
ephphatha, line 482: std::swap is unnecessary here, `stashList[itemIndex]` can be treated as if it's a garbage value at this point.

```suggestion
		stashList[itemIndex] = std::move(stashList[lastItemIndex]);
```

Source/qol/stash.cpp
```diff
@@ -494,53 +449,64 @@ bool UseStashItem(uint16_t c)
...
 	Stash.dirty = true;
 }
 
+unsigned StashStruct::GetPageByItemId(StashCell itemId)
+{
+	if (itemId < stashList.size()) {
+		Point itemPosition = stashList[itemId].position;
+		unsigned currentPageIndex = 0;
+		while (currentPageIndex <= LastStashPage) {
+			if (stashGrids[currentPageIndex][itemPosition.x][itemPosition.y] == itemId + 1) {
```
ephphatha, line 500: This causes allocations if a player has skipped pages when navigating the stash. You can use `std::map::iterator` based loops instead.

```suggestion
		for (const auto &[currentPageIndex, stashGrid] : stashGrids) { // C++17, probably need to use the long form for GCC6 support
			if (stashGrid[itemPosition.x][itemPosition.y] == itemId + 1) {
```

Source/qol/stash.cpp
```diff
@@ -664,6 +630,20 @@ void GoldWithdrawNewText(string_view text)
 	}
 }
 
+void UpdateStashGrid(unsigned page, Point startPoint, Size itemSize, StashStruct::StashCell itemId /*= 0*/)
+{
+	if (itemId == InvalidItemId) {
```
ephphatha, line 635: This comparison will never be true in the current code. `itemId` is expected to be 1-indexed and defaults to 0, InvalidItemId is expected to be used in comparison with a 0-indexed value.

Source/qol/stash.h
```diff
@@ -68,30 +110,110 @@ constexpr Point InvalidStashPoint { -1, -1 };
...
 void GoldWithdrawNewText(string_view text);
 
+/**
+ * @brief Update the StashGrid on the provided page to account for item movement.
+ * @param page stashGrid page to act upon.
+ * @param startPoint x,y stashGrid position (from the bottom-left cell of the item!).
+ * @param itemSize x,y stashGrid size of the item (ex: 2x3).
+ * @param itemId item id exactly as it should be stored in the StashGrid.
+ */
+void UpdateStashGrid(unsigned page, Point startPoint, Size itemSize, StashStruct::StashCell itemId = 0);
```
ephphatha, line 215: This is an internal function that operates on a `StashStruct` instance, it should be a private member of that class.

Source/qol/stash.h
```diff
@@ -18,45 +18,87 @@ class StashStruct {
...
 	int gold;
 	bool dirty = false;
 
+	/**
+	 * @brief Remove the passed item from the Stash, handling grid cleanup and index updates.
+	 * @param item pointer to item being removed.
+	 * @param itemPosition position of the item as it may not currently be (ex: autoequipped).
+	 * @param itemIndex 0-based index of the item as it exists in the stashList.
+	 */
+	void RemoveStashItem(Item *item, Point itemPosition, StashCell itemIndex);
```
ephphatha, line 33: clang-tidy configuration is `methodCase=camelBase` for class methods, existing code still needs to be updated but new methods should follow the expected format.

Source/qol/stash.h
```diff
@@ -18,45 +18,87 @@ class StashStruct {
...
+
+	/**
+	 * @brief Get the current stash page.
+	 * @return page number as int.
```
ephphatha, line 37:
```suggestion
	 * @return page number as unsigned int.
```

Source/qol/stash.cpp
```diff
@@ -698,8 +678,9 @@ bool AutoPlaceItemInStash(Player &player, const Item &item, bool persistItem)
 			if (persistItem) {
 				Stash.stashList.push_back(item);
 				uint16_t stashIndex = static_cast<uint16_t>(Stash.stashList.size() - 1);
+				// move starting position to bottom-left cell
 				Stash.stashList[stashIndex].position = stashPosition + Displacement { 0, itemSize.height - 1 };
-				AddItemToStashGrid(pageIndex, stashPosition, stashIndex, itemSize);
+				UpdateStashGrid(pageIndex, Stash.stashList[stashIndex].position, itemSize, stashIndex + 1);
```
ephphatha, line 683: Ordinarily I'd be against requiring callers adjust the item index to match the specific implementation (i.e. doing the +1 here instead of in the method) however given `UpdateStashGrid` is effectively an internal function it seems fine as long as you mark it private.

Source/qol/stash.cpp
```diff
@@ -494,53 +449,64 @@ bool UseStashItem(uint16_t c)
...
-	}
-
-	if (stashList.empty()) {
+	if (item == nullptr) {
 		return;
 	}
 
-	// If the item at the end of stash array isn't the one we removed, we need to swap its position in the array with the removed item
+	// get item's size to figure slot changes
+	const Size itemSize = GetInventorySize(*item);
```
ephphatha, line 471: This is the only use of the `Item*` value. Should make this method take a `Rectangle` instead of position/size separately.

Source/qol/stash.h
ephphatha: Also should take `Rectangle` as a param instead of `Point, Size` separately.

Source/qol/stash.h
```diff
@@ -68,30 +110,110 @@ constexpr Point InvalidStashPoint { -1, -1 };
...
+
+/**
+ * @brief Rendering the gold withdraw popup and the current value.
+ * @param out SDL surface upon which to draw the output.
+ * @param amount entered withdraw amount to populate in the withdraw popup.
+ */
 void DrawGoldWithdraw(const Surface &out, int amount);
+
+/**
+ * @brief Stop the withdraw process, ending text input, without gold transfer.
```
ephphatha, line 198:
```suggestion
 * @brief Closes the gold withdraw popup and stop capturing text input.
 * Actually withdrawing gold is performed when the player confirms a value by pressing enter with the popup open, during WithdrawGoldKeyPress.
```

Source/qol/stash.h
```diff
@@ -68,30 +110,110 @@ constexpr Point InvalidStashPoint { -1, -1 };
...
 void CheckStashButtonRelease(Point mousePosition);
+
+/**
+ * @brief Store x,y screen position of mouse, setting StashButtonPressed for use in CheckStashButtonRelease.
+ * @param mousePosition x,y screen coordinates of mouse at press.
+ */
 void CheckStashButtonPress(Point mousePosition);
 
+/**
+ * @brief Being the gold withdraw process, initializing the withdraw dialog popup.
```
ephphatha, line 180:
```suggestion
 * @brief Begin the gold withdraw process, initializing the withdraw dialog popup.
```

Source/items.h
hateborne: Committed the following suggestion, marking this one as resolved :)

Source/qol/stash.cpp
hateborne: The intent was to create an informal barrier due to an accident that broke things. I accidentally allowed the 65535 value to be written to the grid, which caused all manner of weirdness. I added this to prevent that, admittedly unlikely, situation from occurring again.

That being said, if you feel strongly that is has no value, I am **_not_** arrogant enough to assume I know best and will remove it. :)

Source/qol/stash.h
hateborne: Moving it private is a fine suggestion. The reason for the separate point & size is the startPosition is move from the bottom-left cell to the top-left cell on [stash.cpp L640](https://github.com/hateborne/devilutionX/blob/nov2022_qol_tweak/Source/qol/stash.cpp#L640) (below). 
```cpp
void UpdateStashGrid(unsigned page, Point startPoint, Size itemSize, StashStruct::StashCell itemId /*= 0*/)
{
	if (itemId == InvalidItemId) {
		return;
	}

	// set start position to top-left corner as PointsInRectangle draws right (x) and down (y)
	Point adjustedStartPoint = startPoint - Displacement(0, itemSize.height - 1);

	for (Point point : PointsInRectangle(Rectangle { adjustedStartPoint, itemSize })) {
		Stash.stashGrids[page][point.x][point.y] = itemId;
	}
}
```

Source/qol/stash.cpp
hateborne: The intent here is to find an itemId that may not necessarily be on the current page. At the time, it seemed to be less complex to start from 0 than to start at the current page then wrap around.

Source/qol/stash.h
```diff
@@ -68,30 +110,111 @@ constexpr Point InvalidStashPoint { -1, -1 };
...
+
+/**
+ * @brief Rendering the gold withdraw popup and the current value.
+ * @param out SDL surface upon which to draw the output.
+ * @param amount entered withdraw amount to populate in the withdraw popup.
+ */
 void DrawGoldWithdraw(const Surface &out, int amount);
+
+/**
+ * @brief Closes the gold withdraw popup and stop capturing text input.
```
qndel, line 198:
```suggestion
 * @brief Close the gold withdraw popup and stop capturing text input.
```
Let's be consistent, closes/stops or close/stop

Source/qol/stash.cpp
hateborne: Just mentioned this above as well, but this is due to the start position being moved from the bottom-left cell to the top-left cell inside of `UpdateStashGrid`.

Source/qol/stash.cpp
hateborne: Done, thank you!

Source/qol/stash.cpp
ephphatha: That's fine, `stashGrid` is a `std::map` which orders keys by natural ordering. The range based loop will start from the first already allocated page and continue through all pages that already exist. Using `operator[]` to access a key that doesn't exist will end up inserting a blank page if someone skips from page 5 to 10.

As an example, lets say I order things in the stash so I have two pages of potions starting from page 1 (key 0), three pages of armour from page 10, and two pages of weapons from page 20. I then call this function with an index that doesn't exist (so the early return never happens). The range based loop will iterate over the pairs with keys [0, 1, 9, 10, 11, 19, 20] in that order (assuming the player hasn't flicked through the stash since last loading the game). The index based loop will iterate over [0..99], allocating 93 new empty grids.

Source/qol/stash.cpp
ephphatha: In that case the following would reflect your intent better:

```suggestion
	if (itemId > stashList.size()) {
```

Source/qol/stash.cpp
hateborne: (Apologies for the delay, work got messy)
I actually think the stash should be reworked a bit further as so many of the functions littered throughout the stash header/cpp files expected publicly accessible everything (which feels like it explains the struct vs class choice). I have no issue making these additional changes so it's a bit closer to a "proper" public/private usage model, but I wanted to make sure you/others agree that it is the right decision. 

A stopgap measure would be to make the function private, then add public "helper" functions that do the last second conversion (+1 or -1) before calling the private UpdateStashGrid function.

Source/qol/stash.cpp
hateborne: Excellent suggestion yet again! committing

Source/qol/stash.cpp
hateborne: I can take a look at batch rendering separately as that's likely something that may affect more than just the stash. :)

Source/qol/stash.h
hateborne: Making this happen in next push (just a few min)

Source/qol/stash.cpp
hateborne: Excellent explanation and I did change this just now to move away form range based iteration.

Source/inv.h
```diff
@@ -19,6 +19,7 @@ namespace devilution {
 #define INV_SLOT_HALF_SIZE_PX (INV_SLOT_SIZE_PX / 2)
 #define INV_ROW_SLOT_SIZE 10
 constexpr Size InventorySlotSizeInPixels { INV_SLOT_SIZE_PX };
+static constexpr uint16_t InvalidItemId = uint16_t(-1);
```
ephphatha, line 22: This constant is used in comparisons with an index into the global Stash's internal item list. Should probably be defined in stash.h instead of inv.h. It's used more like a nullptr value (or end iterator/`npos` style invalid position) so a name that doesn't reference "ID" would be nice.

Source/qol/stash.cpp
```diff
@@ -453,7 +407,8 @@ bool UseStashItem(uint16_t c)
 	if (stextflag != STORE_NONE)
 		return true;
 
-	Item *item = &Stash.stashList[c];
+	Item *item = &Stash.stashList[itemIndex];
+	Point itemPosition = item->position;
```
ephphatha, line 411: Nothing modifies item->position between this line and when the call to `Stash.RemoveStashItem()` occurs on line 459. This variable can be inlined or the declaration moved to before the first call that potentially modifies the position (if I've missed it).

Source/qol/stash.cpp
```diff
@@ -494,53 +449,62 @@ bool UseStashItem(uint16_t c)
 
 	UseItem(MyPlayerId, item->_iMiscId, item->_iSpell);
 
-	if (Stash.stashList[c]._iMiscId == IMISC_MAPOFDOOM)
+	if (Stash.stashList[itemIndex]._iMiscId == IMISC_MAPOFDOOM)
 		return true;
-	if (Stash.stashList[c]._iMiscId == IMISC_NOTE) {
+	if (Stash.stashList[itemIndex]._iMiscId == IMISC_NOTE) {
```
ephphatha, line 454: might as well use the `Item &item` alias here

Source/qol/stash.cpp
```diff
@@ -272,13 +241,14 @@ void InitStash()
...
+void TransferItemToInventory(Player &player, uint16_t itemIndex)
 {
-	if (itemId == uint16_t(-1)) {
+	if (itemIndex == InvalidItemId) {
 		return;
 	}
 
-	Item &item = Stash.stashList[itemId];
+	Item &item = Stash.stashList[itemIndex];
+	Point originalPosition = item.position;
```
ephphatha, line 251: Reading the position of an empty item isn't ideal. The declaration of `originalPosition` should move to immediately before the call to `AutoPlaceItemInInventory()`

hateborne: Suggestion applied in most recent push

Source/qol/stash.cpp
hateborne: Suggestion applied in most recent push

Source/qol/stash.cpp
hateborne: Suggestion applied in most recent push

Source/inv.h
hateborne: Suggestion applied in most recent push (removed all InvalidItemId references and reverted to `-1`)

ephphatha:
Did you have some more commits you intended to push? You've marked comments as resolved but there's no change to the source visible yet.

hateborne:
> Did you have some more commits you intended to push? You've marked comments as resolved but there's no change to the source visible yet.

Oh nuts, apologies! I forgot to push last two commits. Thank you for your continued patience in this

AJenbo:
it looks like something has changed game behavior which is why the test are currently failing for this PR.

Could you:
1. merge latest master and see if that helps
2. run the timedemo locally and see if you can see where things diverge (copy the .sv and .dmo from the test fixtures to your save game folder `devilutionx --demo 0 --spawn`)

Source/cursor.cpp
```diff
@@ -366,7 +366,7 @@ void CheckCursMove()
 	if ((sgbMouseDown != CLICK_NONE || ControllerActionHeld != GameActionType_NONE) && IsNoneOf(LastMouseButtonAction, MouseActionType::None, MouseActionType::Attack, MouseActionType::Spell)) {
 		InvalidateTargets();
 
-		if (pcursmonst == -1 && ObjectUnderCursor == nullptr && pcursitem == -1 && pcursinvitem == -1 && pcursstashitem == uint16_t(-1) && pcursplr == -1) {
+		if (pcursmonst == -1 && ObjectUnderCursor == nullptr && pcursitem == -1 && pcursinvitem == -1 && pcursstashitem == -1 && pcursplr == -1) {
```
obligaron, line 369: This line is likely failing the timedemo.

Assigning -1 to a uint16_t (unsigned int) is not the same as comparing a unsigned int 16 with a signed 32 bit int.

Run the following example:
```
uint16_t test = -1;
if (test == -1)
	Log("What not the same?!");
```
`What not the same?!` will be printed. 😉 
uin16_t will be changed/widened to `int32_t` for comparison.
But assigning will use `uint16_t` casted -1.

That's why
> Uniform invalid item id usage: This is an opinionated change to clarify the code at a glance. While it is functionally the same as uint16_t(-1), it feels like a more obvious explanation of what is happening.

is not correct.

But your statement that `uint16_t(-1)` is not obvious is still true.
My suggestion: instead of `uint16_t(-1)` we could use `StashStruct::EmptyCell`. This would make the code more readable and still works fine. 🙂

hateborne: Reverted to `-1` for now as I am not found of EmptyCell being used sporadically. I feel like a globally defined "invalid item id" should exist, but this PR has been going too long and needs to be completed without anymore scope creep. 

That aside, thank you for the excellent explanation!

PR #5479, opened 2022-11-11
Fix amiga build

arczi84:


Source/utils/paths.cpp
```diff
@@ -52,6 +52,9 @@ std::string FromSDL(char *s)
 		Log("{}", SDL_GetError());
 		SDL_ClearError();
 	}
+#ifdef defined(__MORPHOS__) || defined(__amigaos__)
+	AddTrailingSlash(result);
```
AJenbo, line 56: This should be moved to the Amiga implementation of SDL_GetBasePath and SDL_GetPrefPath

Source/utils/sdl2_to_1_2_backports.cpp
```diff
@@ -792,6 +795,8 @@ char *SDL_GetBasePath()
 		retval = readSymLink("/proc/curproc/exe");
 #elif defined(__QNXNTO__)
 		retval = SDL_LoadFile("/proc/self/exefile", NULL);
+#elif defined(__amigaos__)
+		//avoid compiling readSymLink
```
AJenbo, line 799: Since the previous id-def block returns for both 3DS and Amiga we should probably add an else to it that wraps the rest of the code instead of this.

Source/utils/paths.cpp
arczi84: I took this from Morphos, probably not needed as I've modified SDL_GetBasePath and SDL_GetPrefPath already.

Source/utils/paths.cpp
```diff
@@ -86,6 +89,8 @@ const std::string &PrefPath()
 		prefPath = FromSDL(IOSGetPrefPath());
 #elif defined(NXDK)
 		prefPath = NxdkGetPrefPath();
+#elif defined(__amigaos__)
+		prefPath = std::string("PROGDIR:");
```
AJenbo, line 93: Since you made SDL_GetPrefPath return this value is this overwrite really needed?

Source/utils/paths.cpp
```diff
@@ -103,6 +108,8 @@ const std::string &ConfigPath()
 		configPath = FromSDL(IOSGetPrefPath());
 #elif defined(NXDK)
 		configPath = NxdkGetPrefPath();
+#elif defined(__amigaos__)
+		prefPath = std::string("PROGDIR:");
```
AJenbo, line 112: Since you made SDL_GetPrefPath return this value is this overwrite really needed?

CMake/platforms/amiga.cmake
```diff
@@ -20,7 +20,7 @@ add_definitions(-Dfseeko=fseek)
 
 list(APPEND DEVILUTIONX_PLATFORM_LINK_LIBRARIES ZLIB::ZLIB)
 if(NOT WARPOS)
-  list(APPEND DEVILUTIONX_PLATFORM_LINK_LIBRARIES -ldebug)
+  list(APPEND DEVILUTIONX_PLATFORM_LINK_LIBRARIES -ldebug -fbbb=-)
```
AJenbo, line 23: Have you tested which optimization is breaking things?

https://eab.abime.net/showthread.php?t=85474&page=51

How does the issue materialize?

arczi84: Crashes system without it. I will report this to Amiga gcc maintainer. I did that many times before and he fixed it.

Source/utils/paths.cpp
arczi84: Right, haven't noticed SDL_GetPrefPath there.

Source/utils/paths.cpp
arczi84: Right, haven't noticed SDL_GetPrefPath there.

Source/utils/sdl2_to_1_2_backports.cpp
arczi84: True.

Source/utils/paths.cpp
arczi84: But because of `prefPath = std::string("." DIRECTORY_SEPARATOR_STR);`  it is still needed. 
"./" won't work on Amiga.

Source/utils/sdl2_to_1_2_backports.cpp
```diff
@@ -779,6 +779,9 @@ char *SDL_GetBasePath()
 #if defined(__3DS__)
 	retval = SDL_strdup("file:sdmc:/3ds/devilutionx/");
 	return retval;
+#elif defined(__amigaos__)
+	retval = SDL_strdup("PROGDIR:");
+	return retval;
```
AJenbo, line 784: Trailing space (this is what is causing clang-format to fail)
```suggestion
	return retval;
```

Source/utils/sdl2_to_1_2_backports.cpp
```diff
@@ -849,6 +854,9 @@ char *SDL_GetPrefPath(const char *org, const char *app)
 #if defined(__3DS__)
 	retval = SDL_strdup("sdmc:/3ds/devilutionx/");
 	return retval;
+#elif defined(__amigaos__)
+	retval = SDL_strdup("PROGDIR:");
+	return retval;
```
AJenbo, line 859:
```suggestion
	return retval;
```

Source/utils/sdl2_to_1_2_backports.cpp
AJenbo, line 799:
```suggestion
		//avoid compiling readSymLink
```

CMake/platforms/amiga.cmake
arczi84: Although this is wrong place for this switch.

Source/utils/paths.cpp
AJenbo: In that case preferably have the ifdef around the `if (FileExistsAndIsWriteable... }` block in that case.

arczi84: Done.

Source/utils/paths.cpp
```diff
@@ -88,9 +88,11 @@ const std::string &PrefPath()
 		prefPath = NxdkGetPrefPath();
 #else
 		prefPath = FromSDL(SDL_GetPrefPath("diasurgical", "devilution"));
+#if !defined(__amigaos__)
 		if (FileExistsAndIsWriteable("diablo.ini")) {
-			prefPath = std::string("." DIRECTORY_SEPARATOR_STR);
+			configPath = std::string("." DIRECTORY_SEPARATOR_STR);
```
glebm, line 93: Perhaps this should be `prefPath`?

Source/utils/paths.cpp
```diff
@@ -103,6 +105,8 @@ const std::string &ConfigPath()
 		configPath = FromSDL(IOSGetPrefPath());
 #elif defined(NXDK)
 		configPath = NxdkGetPrefPath();
+#elif defined(__amigaos__)
+		prefPath = std::string("PROGDIR:");
```
glebm, line 109: `configPath`?

Source/utils/paths.cpp
AJenbo:
```suggestion
			prefPath = std::string("." DIRECTORY_SEPARATOR_STR);
```

Source/utils/paths.cpp
AJenbo: This one should also be changed to wrap `FileExistsAndIsWriteable`

glebm: This is the `ConfigPath()` method, probably not?

AJenbo:
```cpp
#if !defined(__amigaos__)
		if (FileExistsAndIsWriteable("diablo.ini")) {
			prefPath = std::string("." DIRECTORY_SEPARATOR_STR);
		}
#endif
```

Source/utils/paths.cpp
arczi84: Yeah, copied wrong code.

MBeijer:
I'll investigate at what's causing the resulting binary from the CI process not to run properly when compiled through the docker image, as discussed with @arczi84 in Discord.

Issue #5192, opened 2022-08-05
Help window is missing its scrollbar


AJenbo:
When opening the in game help no scroll bar appears on the right side of the window.

Version 1.4.1, Windows

vlolteanu:
![image](https://user-images.githubusercontent.com/3142367/183944409-ea8de07d-8503-4d6d-a9ac-62a65b04b29a.png)

AJenbo:
![image](https://user-images.githubusercontent.com/204594/183955035-da2bad46-79b4-4d68-bf65-17aac85788e9.png)

Raphalex46:
A friend and I have started working on this as it seems like a good first issue for us (beginners for open source contributions and for devilutionX as well). Is that ok ?

AJenbo:
@Raphalex46 yes that is excellent, thank you for making a note of this

vlolteanu:
@Raphalex46 
![image](https://user-images.githubusercontent.com/3142367/200970427-5ed93cc9-31b2-443d-8a41-fcaa1d4bf2b9.png)

PR #5481, opened 2022-11-12
Add scrollbar to help window

Raphalex46:
This is my attempt at adding a scrollbar to the help window. It is heavily inspired by the code for the store sliders in `stores.cpp`.

I had to reduce the width of the text a bit to actually leave enough space for the scrollbar (also explicitly using this width value for the wrapping of the text).

Directly using the code from the stores UI code looks hard because it uses store-specific variables and the way the help window text and scroll works seems different as well.

This PR is for issue #5192

### DIFF
--- a/Source/help.cpp
+++ b/Source/help.cpp
@@ -7,6 +7,7 @@
 #include <vector>
 
 #include "DiabloUI/ui_flags.hpp"
+#include "engine/render/clx_render.hpp"
 #include "engine/render/text_render.hpp"
 #include "init.h"
 #include "minitext.h"
@@ -101,7 +102,7 @@ constexpr int PaddingTop = 32;
 constexpr int PaddingLeft = 32;
 
 constexpr int PanelHeight = 297;
-constexpr int ContentTextWidth = 577;
+constexpr int ContentTextWidth = 565;
 
 int LineHeight()
 {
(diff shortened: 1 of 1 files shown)


AJenbo:
The code in stores.cpp it a big mess begging to get rewritten... glad you didn't reuse it :)

Raphalex46:
Yeah, well what I did was kind of a mess as well then I think as I did it in the same style, using the information I had available about the help window.

However, the `help.cpp` code looks like a mess as well, it is already calling some functions from `stores.cpp`, so my changes are kind of building on the mess, which I don't like very much. Is the rewrite going to make it easier to re-use UI elements like this ?

Source/help.cpp
```diff
@@ -138,6 +139,29 @@ int NumVisibleLines()
...
+		ClxDraw(out, { sliderXPos, sliderCurrent }, (*pSTextSlidCels)[13]);
+	}
+	ClxDraw(out, { sliderXPos, sliderCurrent }, (*pSTextSlidCels)[10]);
+	// Subtract visible lines from the total number of lines to get the actual
+	// scroll range
+	const float scrollRange = HelpTextLines.size() - NumVisibleLines();
+	// Subtract the size of the arrow buttons to get the length of the interior
+	// part of the slider
+	const float sliderLength = sliderCurrent - 12 - sliderStart;
+	ClxDraw(out, { sliderXPos, sliderStart + (int)(((float)SkipLines * sliderLength) / scrollRange) }, (*pSTextSlidCels)[12]);
```
AJenbo, line 162: I'm not sure I see the point in having a float here, it's just subtracting and multiplying whole numberes. The cast back to int happens right after the division so the precision is discarded right away.

Raphalex46: You're 100% right, I think I wanted to do it some other way at first and thought for some reason we needed the cast, but we really don't.

AJenbo:
P.s. don't mind the failing ci/circleci: 3ds testflow, it's because a dependency ment 404 just a few minutes ago

Source/help.cpp
AJenbo: Ok, if you change that to integer math then I'm happy to merge this :)

Raphalex46: Should I add a commit to the PR ? Or is modifying the commit cleaner ? (I'm new to this as mentioned in the issue ahah)

Raphalex46:
Yeah I figured this small change couldn't break the 3DS build :)

Source/help.cpp
AJenbo: We prefer having things clean, but that also means that we will squash your PR in to one commit if it's messy so either is fine :)

AJenbo:
```suggestion
	const int sliderLength = sliderCurrent - 12 - sliderStart;
	ClxDraw(out, { sliderXPos, sliderStart + (SkipLines * sliderLength / scrollRange) }, (*pSTextSlidCels)[12]);
```

Raphalex46: Done :). Thank you for your quick answers !

AJenbo:
Oh, I might have been a bit to quick with merging, could you look at this error on some of the targets:

```

/home/runner/work/devilutionX/devilutionX/Source/help.cpp:162:29: error: non-constant-expression cannot be narrowed from type 'unsigned int' to 'int' in initializer list [-Wc++11-narrowing]
        ClxDraw(out, { sliderXPos, sliderStart + ((SkipLines * sliderLength) / scrollRange) }, (*pSTextSlidCels)[12]);
```

glebm:
The error was fixed in https://github.com/diasurgical/devilutionX/commit/4f21264a98ffb2fbd44b55ce887e0cca2234d274

PR #5485, opened 2022-11-12
Sync SpawnUnique (multiplayer unique quest rewards)

obligaron:
First commit refactors `SpawnItem` to make the behavior of `T_UNIQ` more clear.
Second commit allows to pick up unique quest reward items for all multiplayer clients (for example bovine plate).

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1309,9 +1309,6 @@ void GetItemBonus(const Player &player, Item &item, int minlvl, int maxlvl, bool
 
 _item_indexes RndUItem(Monster *monster)
 {
-	if (monster != nullptr && (monster->data().treasure & T_UNIQ) != 0 && !gbIsMultiplayer)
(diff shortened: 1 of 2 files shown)


Source/monster.cpp
```diff
@@ -1437,7 +1437,7 @@ void MonsterTalk(Monster &monster)
 			Quests[Q_VEIL]._qlog = true;
 		}
 		if (monster.talkMsg == TEXT_VEIL11 && (monster.flags & MFLAG_QUEST_COMPLETE) == 0) {
-			SpawnUnique(UITEM_STEELVEIL, monster.position.tile + Direction::South);
+			SpawnUnique(UITEM_STEELVEIL, monster.position.tile + Direction::South, true);
```
qndel, line 1440: sendmsg in SpawnUnique is true everywhere, maybe let's go with a default true value for it?

obligaron: Changed 🙂

AJenbo (approved):

PR #5487, opened 2022-11-13
Multiplayer: Spawn butcher's cleaver and undead crown

obligaron:
When singleplayer quests are activated in multiplayer butcher and skeleton kings drops their unique items (butcher's cleaver and undead crown).

- Only one item is dropped regardless how many player are present in the game.
- Currently the items drops every time you kill butcher/skeleton king (no limit per character).
- I checked that the items don't morph when loading the save game or dropping the item on the ground and pick them up with another character.

This change can be a bit controversial that's why it was separated from #5485

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3058,7 +3058,7 @@ void SpawnItem(Monster &monster, Point position, bool sendmsg)
 
 	bool dropsSpecialTreasure = (monster.data().treasure & T_UNIQ) != 0;
 
-	if (dropsSpecialTreasure && !gbIsMultiplayer) {
+	if (dropsSpecialTreasure && !UseMultiplayerQuests()) {
 		Item *uniqueItem = SpawnUnique(static_cast<_unique_items>(monster.data().treasure & T_MASK), position, false);
 		if (uniqueItem != nullptr && sendmsg)
 			NetSendCmdPItem(false, CMD_DROPITEM, uniqueItem->position, *uniqueItem);


AJenbo:
I don't think we should apply this change.

obligaron:
Why?
This pr would align to your requested requirement in this [comment](https://github.com/diasurgical/devilutionX/pull/5471#issuecomment-1328333963).
> Again this breaks with how it's handled in Single Player. Lets avoid such arbitrary changes to the game.
If we want to restore original singleplayer behavior we should take this pr. 😉 

I think we should handle all quest rewards (uniques & exlicier) the same. This included butcher's cleaver and undead crown.
If we don't want to take to have them in multiplayer that's also fine, but I think we should do this identical.

btw: In discord that was the idea to drop a magic item from the same typ (heave armor, helmet, ...) that normally the unique would be. Perhaps we should do this and let `SpawnUnique` handle this (that's what I experiment for #5554). With this pr this would also work for butcher and skeleton king. 🙂

AJenbo:
What I was arguing against in that comment was to enable unique items but try to add logic for multiple runs when none such logic exists in single player despite multiple runs being just as possible.

Undead crown and cleaver was explicity disabled for multiplayer. There appears to have been no other reason for this other then them not working well in a multiplayer settings. So I think there is a good precedence for this behavior.

>  If we don't want to take to have them in multiplayer that's also fine, but I think we should do this identical.

Yes we should not drop any of the unique singplayer items in multiplayer. Instead the quest reward should be a regular item like seen in the quests that where originally available in multiplayer.

> btw: In discord that was the idea to drop a magic item from the same typ (heave armor, helmet, ...) that normally the unique would be. Perhaps we should do this and let SpawnUnique handle this (that's what I experiment for https://github.com/diasurgical/devilutionX/pull/5554). With this pr this would also work for butcher and skeleton king. slightly_smiling_face

Yeah I like this idea

obligaron:
I think they were only disabled, cause they are buggy on vanilla multiplayer and no one had the time to fix them. 😉

For me the unique quest rewards should behave identically for all quests on singleplayer mode. That's why I would still prefer to merge this pr. 😁

AJenbo:
How where they buggy?

obligaron:
They are missing [NetSendCmdDItem](https://github.com/diasurgical/devilution/blob/master/Source/items.cpp#L3006) cause of the early [return](https://github.com/diasurgical/devilution/blob/master/Source/items.cpp#L2974).
I fixed this in #5485 that's why I know about it. 😜 
The multiplayer part was rushed and I think the current behavior was the easier/faster fix.
But this is only a assumption.

AJenbo:
Alright, I'll give it some more consideration :)

AJenbo (approved):

ikonomov:
Just reading the comments here makes it plain that the game is in safe and loving hands :)  I love everything you guys are doing.

PR #5489, opened 2022-11-13
Assets: Load files directly for UNPACKED_MPQS

glebm:
When using `UNPACKED_MPQS`, avoid all the SDL rwops machinery and use `<cstdio>` / `file_utils` instead.

Benefits:
1. Reduced indirection.
2. Can test for the file's existence and get the file size without opening it, which is much faster.

This also cleans up file handling a bit:
1. Eliminates the need for manual handle management.
3. Separates finding files from opening them.


Source/engine/assets.cpp
```diff
@@ -65,50 +67,140 @@ bool OpenMpqFile(const char *filename, MpqArchive **archive, uint32_t *fileNumbe
...
+#endif
+
+	return result;
+}
 #endif
 
-	return nullptr;
+AssetHandle OpenAsset(AssetRef &&ref, bool threadsafe)
+{
+#if UNPACKED_MPQS
```
AJenbo, line 162:
```suggestion
#ifdef UNPACKED_MPQS
```

Source/engine/assets.cpp
```diff
@@ -65,50 +67,140 @@ bool OpenMpqFile(const char *filename, MpqArchive **archive, uint32_t *fileNumbe
...
+}
+
+SDL_RWops *OpenAssetAsSdlRwOps(const char *filename, bool threadsafe)
+{
+#ifdef UNPACKED_MPQS
+	AssetRef ref = FindAsset(filename);
+	if (!ref.ok())
+		return nullptr;
+	return SDL_RWFromFile(ref.path, "rb");
+#else
```
AJenbo, line 201: Isn't the SDL path only for `#ifndef`?

AJenbo (changes requested):
A couple of the ifdefs look suspect

Source/engine/assets.cpp
glebm: We still use the SDL path for video rendering and music playback, which both accept only SDL_RWops

Source/engine/assets.cpp
glebm: Thanks, fixed

AJenbo (approved):

PR #5491, opened 2022-11-13
Multiplayer: Enable skeleton king's lair

obligaron:
Contributes to #926

When singleplayer quests are activated in multiplayer use singleplayer skeleton king's quest including his lair.

- DeltaLoadLevel: Item's check that they are not placed on a solid tiles. But objects can change the dungeon layout. That's why objects need to be synced/loaded before items else items could get placed on a wrong tile.
- Break crux wasn't synced in multiplayer (cause they were only used in skeleton king's lair).
- Ensures that entering skeleton king's lair plays skeleton kings speech in multiplayer. For this the already existing `PlayDungMsgs`/`sfxdelay` is used (normal `PlaySFX` don't work in the first 10 game ticks in multiplayer).
- For single- and multiplayer: Changed that skeleton kings speech is only played once when entering skeleton king's lair (not on every entrance).

SkeletongKingsLair_PR

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -2781,7 +2781,7 @@ void LoadGameLevel(bool firstflag, lvl_entry lvldir)
 			}
 		}
 		IncProgress();
-
+		PlayDungMsgs();
 		InitMultiView();
 		IncProgress();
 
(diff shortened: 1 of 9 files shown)


AJenbo:
Finally, over 2.5 decades and this is not just false avertissement :D
image

StephenCWills (approved):

PR #5493, opened 2022-11-15
Add SDL2_image-devel as Fedora dependency

yuripourre:
Otherwise the project cannot compile:
```
*** No rule to make target '/usr/lib64/libSDL2_image.so', needed by 'test/liblibdevilutionx_so.so'.  Stop.
```

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -36,7 +36,7 @@ sudo apt-get install smpq
 ### Installing dependencies on Fedora
 
 ```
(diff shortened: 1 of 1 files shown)


glebm (approved):
Hmm, it should be building SDL2_image from source when it's not available on the system. Did you do a clean build?

In any case, it is better to use a system dependency, approved.

AJenbo (approved):

yuripourre:
@glebm Good point, I was using a pretty old build in a fresh Fedora install. That might be the case, thanks.

PR #5497, opened 2022-11-17
Reset button mapping for MiyooMini to the current default layout

Trihedraf:
There is no way to confirm in the in-game menu with this being set. With this change both the start button and A button work for confirm. There is a dedicated Menu button on the MiyooMini making there be no reason the start button needs to be the menu button.


glebm:
> There is no way to confirm in the in-game menu with this being set.

Why doesn't A work when start is mapped to the menu?

glebm:
Ah, I see, start is actually `TAB`.
By the way, the menu can also be triggered with START + SELECT

glebm:
We've actually introduced remappable controls in 1.5.0, so it'd be better to change all the mappings to the actual buttons.

So e.g. instead of:

```diff
- set(KBCTRL_BUTTON_BACK SDLK_BACKSPACE)
+ set(KBCTRL_BUTTON_BACK SDLK_RCTRL)
```

This way they'll be configurable, for now in the ini, hopefully in the UI before 1.5.0.

Trihedraf:
Ok I've set all the buttons to the default layout, so it should all be remappable with the new system

glebm (approved):

PR #5501, opened 2022-11-18
amiga.cmake: Remove `find_package(ZLIB)`

glebm:
We're not using the system zlib on Amiga

### DIFF
--- a/3rdParty/zlib/CMake/FindZLIB.cmake
+++ b/3rdParty/zlib/CMake/FindZLIB.cmake
@@ -1,0 +1,1 @@
+# No-op, ZLIB::ZLIB is built from source.
--- a/3rdParty/zlib/CMakeLists.txt
+++ b/3rdParty/zlib/CMakeLists.txt
@@ -1,5 +1,7 @@
 include(functions/FetchContent_MakeAvailableExcludeFromAll)
 
+set(CMAKE_POLICY_DEFAULT_CMP0048 NEW)
+
 include(FetchContent)
 FetchContent_Declare(zlib
     URL https://www.zlib.net/zlib-1.2.13.tar.gz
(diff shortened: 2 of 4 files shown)


AJenbo:
Then why did the build fail :thinking:

StephenCWills:
libpng is looking for it. We may need to make use of the `PNG_BUILD_ZLIB` option.

https://github.com/glennrp/libpng/blob/0a158f3506502dfa23edfc42790dfaed82efba17/CMakeLists.txt#L45-L52

AJenbo:
The way things are currently PNG is only used for SDL2 targets so if it's not disabled there it probably should be.

glebm:
We should stub out the package finder when we build zlib from source.

glebm:
Filed #5504 for the separate issue of avoiding SDL_image and png when targeting SDL1.

arczi84 (dismissed):
Works fine.

glebm:
Thanks @StephenCWills! I won't have access to a PC until tomorrow night, please squash the commits when merging!

StephenCWills:
@glebm Should we add `find_package(zlib REQUIRED)` to `Dependencies.cmake`? All the other `DEVILUTIONX_SYSTEM_*` dependencies have it. Like, if you compare it to bzip in the following snippet.

https://github.com/diasurgical/devilutionX/blob/77ded34137f8937e0133572518c22f8e40459ad9/CMake/Dependencies.cmake#L6-L20

AJenbo (approved):

StephenCWills (approved):

glebm:
@StephenCWills Actually we should probably remove the `find_package(BZip2 REQUIRED)` call. It is not used by DevilutionX directly -- it's a dependency of libmpq, which also calls `find_package(BZip2 REQUIRED)`.

PR #5507, opened 2022-11-20
Fix Amiga build#2

arczi84:


Source/diablo.cpp
```diff
@@ -1027,6 +1027,9 @@ void SetApplicationVersions()
 
 void CheckArchivesUpToDate()
 {
+#if defined(__amigaos__)
```
glebm, line 1030: Why doesn't this work on Amiga?

glebm (changes requested):

StephenCWills:
Should this be closed on account of #5555 and #5556, or do we still need to update the dockerfile?

Source/diablo.cpp
arczi84: Actually it's works. I wasn't aware that assets v3 are needed.

PR #5508, opened 2022-11-20
Multiplayer: Enable Ogden's Sign Quest

obligaron:
Contributes to #926

- Sign chest was not synced
- Ogden now also talk about the quest when you only entered lvl 4 and skipped 3 (for example cause you just joined the game and other player made a portal for you)
- `TalktoMonster` now uses the player that talks to the monster to check behavior not the player that is the current target of the monster
- When a player talks to a monster this is currently played back for all players on the same level
- A player can still enter catacombs even if the quest is not completed

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -4473,9 +4473,8 @@ Monster *PreSpawnSkeleton()
 	return skeleton;
 }
 
-void TalktoMonster(Monster &monster)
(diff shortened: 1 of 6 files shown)


AJenbo (approved):
We should probably change things so that snortspil's dialog only plays for the player that triggered it.

PR #5510, opened 2022-11-20
CMake: Fixup BZip2 dependency

glebm:
1. Remove `find_package(BZip2 REQUIRED)` from Dependencies.cmake because it is not used by devilutionx directly (it's a libmpq dependency).
2. For emscripten, use the built-in version of bzip2.
3. Reverts 0b4705cf1c82db27f0da6a883b176b3ad813e112, which added a redundant `find_package(ZLIB REQUIRED)` to `Dependencies.cmake`.


StephenCWills (approved):
This makes more sense, I think.

PR #5512, opened 2022-11-20
CMake: Clean up dead code removal flags

glebm:
`-Wl,--gc-sections,--as-needed` only needs to be passed to the linker. Fixes an unused linker flags warning with clang.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -181,11 +181,14 @@ Make sure to install the workload `Desktop development with C++` and the individ
 
 Install [Git for Windows](https://gitforwindows.org/)
 
-Run the following commands in Command Prompt or Powershell  
-git clone https://github.com/microsoft/vcpkg  
-cd vcpkg  
-bootstrap-vcpkg.bat  
+Run the following commands in Command Prompt or Powershell
+
+~~~ bat
+git clone https://github.com/microsoft/vcpkg
+cd vcpkg
+bootstrap-vcpkg.bat
 vcpkg integrate install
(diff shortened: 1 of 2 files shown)


docs/building.md
```diff
@@ -181,10 +181,10 @@ Make sure to install the workload `Desktop development with C++` and the individ
...
 Install [Git for Windows](https://gitforwindows.org/)
 
-Run the following commands in Command Prompt or Powershell  
-git clone https://github.com/microsoft/vcpkg  
-cd vcpkg  
-bootstrap-vcpkg.bat  
+Run the following commands in Command Prompt or Powershell
+git clone https://github.com/microsoft/vcpkg
+cd vcpkg
+bootstrap-vcpkg.bat
```
StephenCWills, line 187: The spaces here are actually necessary to indicate that consecutive lines are not part of the same paragraph and should, in fact, be placed on separate lines. However, perhaps it would make more sense to put the commands in a code block.

StephenCWills (changes requested):

glebm:
I've fixed the PS4 linker flag passing in the SDK (<https://github.com/PacBrew/ps4-openorbis/pull/7>) but there is a crash in create-fself. Will remove gc-sections for PS4 for now as weren't using it previously anyway.

docs/building.md
glebm: Done

StephenCWills (approved):

PR #5514, opened 2022-11-20
PS4: Update SDK (1.0 -> 1.1)

glebm:
pacbrew-pacman is updated for Ubuntu 22.04 (the new version works on 22.04 but not on 20.04, the old version does not work on 20.04)

### DIFF
--- a/.github/workflows/PS4.yml
+++ b/.github/workflows/PS4.yml
@@ -15,10 +15,10 @@ concurrency:
 
 jobs:
   ps4:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


StephenCWills (approved):
It makes no sense to me why `ubuntu-latest` would be 20.04, but I guess that's how it is.

AJenbo:
GitHub is scared of breaking builds by upgrading it...

StephenCWills:
That's the point of having tags for specific versions. In my unprofessional opinion, they need to just rip off that band-aid.

AJenbo:
I much agree

PR #5516, opened 2022-11-20
Multiplayer: Enable Anvil of Fury Quest

obligaron:
Contributes to #926

- Quest is only activated when player was on level 9 or 10
- Removed checks for magic rock, cause in vanilla both quests (magic rock and anvil of fury) can't be present at the same time.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -901,6 +901,7 @@ void CheckQuestItem(Player &player, Item &questItem)
 	if (questItem.IDidx == IDI_ANVIL && Quests[Q_ANVIL]._qactive != QUEST_NOTAVAIL) {
 		if (Quests[Q_ANVIL]._qactive == QUEST_INIT) {
 			Quests[Q_ANVIL]._qactive = QUEST_ACTIVE;
+			NetSendCmdQuest(true, Quests[Q_ANVIL]);
 		}
 		if (Quests[Q_ANVIL]._qlog) {
 			myPlayer.Say(HeroSpeech::INeedToGetThisToGriswold, 10);
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -387,24 +387,24 @@ void TalkToBlackSmith(Player &player, Towner &blackSmith)
 			}
 		}
 	}
-	if (Quests[Q_ANVIL]._qactive != QUEST_NOTAVAIL) {
-		if (player._pLvlVisited[9] && Quests[Q_ANVIL]._qactive != QUEST_DONE) {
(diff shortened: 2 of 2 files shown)


AJenbo:
I'm pretty sure that anvil and magic rock are always available together?

Trihedraf:
Isn't it just blocking Griswold from trying to talk about both at the same time?

obligaron:
> I'm pretty sure that anvil and magic rock are always available together?

Magic Rock is in [Quest Group 2](https://github.com/diasurgical/devilutionX/blob/master/Source/quests.cpp#L107-L112) and Anvil of Fury is in [Quest Group 3](https://github.com/diasurgical/devilutionX/blob/master/Source/quests.cpp#L113-L118).
So both don't need to be present at the same time. But they can be together, so my sentence is incorrect. Sorry. 😞 

Anyway: Before it checked that the rock quest muss be active to start or complete anvil of fury. That means the player minimum have to be at level 4 and talk to griswold about it. But the player don't have to return the magic rock (`Quests[Q_ROCK]._qactive != QUEST_INIT`).
That didn't made much sense to me. If you need to magic rock to complete anvil of fury that would be fine, but only knowing about it is fine? 🤔 That's why I removed the check entirely.

> Isn't it just blocking Griswold from trying to talk about both at the same time?

This should be achieved by the `return` statements in the code. 🤔

StephenCWills (approved):
LGTM - my only comment is that maybe this is no longer necessary.

https://github.com/diasurgical/devilutionX/blob/20c27d92e4bc31d5abf5f221f2c1a1a9d4c8a592/Source/quests.cpp#L275-L276

Regarding Magic Rock, it seems that vanilla Diablo does require that you visit dlvl 4 and learn about Magic Rock before you can start Anvil of Fury. This does not seem to have anything to do with preventing Griswold from triggering both dialogs simultaneously as that seems to be what `towner[t]._tMsgSaid` is used for. I cannot begin to imagine what the intent was here, but it's worth noting that SP characters would have to pass through dlvl 4 to even arrive at dlvl 9 in any given game because the shortcuts are not active until you reach them for the first time starting from dlvl 1. This is obviously going to be problematic for MP, and the change obligaron is making here won't really change the behavior in SP.

It's also worth noting that it seems like vanilla Hellfire did it differently because they do activate the shortcuts at the start of a new game. Therefore, you can turn in the anvil whether you initiated the quest with Griswold or not. Currently, DevX does not attempt to match this logic and will therefore not allow you to turn in the anvil if Magic Rock is available unless you've visited dlvl 4 and initiated that quest first. I seem to recall some confusion on the Discord regarding that.

PR #5520, opened 2022-11-21
Fix gamepad icon size and alignment

StephenCWills:
* Update CLX based on changes submitted to devilutionx-assets.
* Update character widths to fix overlaps and spacing issues.

Companion PR: https://github.com/diasurgical/devilutionx-assets/pull/10


glebm (approved):

AJenbo:
@glebm should we look in to baking the frame size in to the CLX instead of having it loaded from the .bin file?

glebm:
Yeah that's something we could without changing the format, since CLX already supports frame sizes.
However, the converters do not yet support variable frame size.

AJenbo:
Yeah that is what I was thinking

PR #5524, opened 2022-11-22
Text rendering: Clip text to the given rect

glebm:
Previously, we only clipped to the given surface, without clipping to the given rect.

We do not actually need this for anything, I was just taken by surprise when this did not work as I'd expected.

### DIFF
--- a/Source/engine/render/text_render.cpp
+++ b/Source/engine/render/text_render.cpp
@@ -354,6 +354,16 @@ int GetLineHeight(string_view fmt, DrawStringFormatArg *args, std::size_t argsLe
 	return LineHeights[fontIndex];
 }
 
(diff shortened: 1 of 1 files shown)


StephenCWills:
Does this only clip the right and bottom edges of the rect?

If so, do we need to clip the left and top edges as well? What happens if text is centered horizontally or vertically?

glebm:
I assumed that text is only rendered left-to-right, top-to-bottom, so no need for left/top clipping.
If the text is centered, it is still centered within the same rect, so I think the clipping should stay the same?

StephenCWills:
Now that you mention it, I should have known that.. #5367 😅

StephenCWills (approved):

AJenbo:
Starting to feel like a real professionel here:

image

PR #5528, opened 2022-11-22
Show portal opening animation in town

obligaron:
When a player is in town and a new portal of another player opens the portal opening animation is now shown.
 
Example video:

https://user-images.githubusercontent.com/25415264/203418430-72bd198a-ba8c-4a27-aa3d-032b286a3271.mp4


AJenbo (approved):
nice with the video and comments in the code :)

PR #5529, opened 2022-11-22
rg99: Build with -O2 and tune buffer size

glebm:
`-O2` increases the binary size by 0.5 MiB which we can now afford. Increasing audio buffer size to `768` improves performance (seems to be the sweet spot).

15-20 FPS in dungeon on the max-mem OD fork.


AJenbo (approved):

AJenbo:
Not used to your PRs not being set to automerge :thinking:

PR #5530, opened 2022-11-23
Support demo playback in SDL1

glebm:
Useful for measuring performance in SDL1.
The demomode test now passes in SDL1 as well.


AJenbo (approved):

AJenbo:
> Useful for measuring performance in SDL1.

Very appealing

PR #5532, opened 2022-11-24
fixed some spanish translation inconsistencies

b3t0787:
I noticed some inconsistencies in the Spanish translation of _town_. In some text it was translated to _ciudad_ and in other text it was translated to _pueblo_. I am a native Spanish speaker and I think the most accurate translation is _pueblo_, so I removed all instances of _ciudad_ and changed them to _pueblo_.


AJenbo (approved):
Looks to fit from what I see in the directory 👍

PR #5538, opened 2022-11-27
Fix alignment of values and field names in monster data

StephenCWills:


StephenCWills:
I can't get GitHub to render the changes, so I'll just say that the changes are all whitespace except for monstdat.cpp line 339 where I had to fix some of the field names to match the `UniqueMonsterData` struct.

glebm (approved):

PR #5539, opened 2022-11-27
Debug: Add maxstats and minstats command

kphoenix137:
Maxes all player stats

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -2270,6 +2270,21 @@ size_t OnCheatExperience(const TCmd *pCmd, size_t pnum) // NOLINT(misc-unused-pa
 	return sizeof(*pCmd);
 }
 
+size_t OnCheatStats(const TCmd *pCmd, size_t pnum) // NOLINT(misc-unused-parameters)
+{
+#ifdef _DEBUG
+	if (gbBufferMsgs == 1)
+		SendPacket(pnum, pCmd, sizeof(*pCmd));
+	else {
+		ModifyPlrStr(Players[pnum], 255);
+		ModifyPlrMag(Players[pnum], 255);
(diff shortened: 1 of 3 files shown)


Source/msg.cpp
```diff
@@ -2270,6 +2270,21 @@ size_t OnCheatExperience(const TCmd *pCmd, size_t pnum) // NOLINT(misc-unused-pa
...
+size_t OnCheatStats(const TCmd *pCmd, size_t pnum) // NOLINT(misc-unused-parameters)
+{
+#ifdef _DEBUG
+	if (gbBufferMsgs == 1)
+		SendPacket(pnum, pCmd, sizeof(*pCmd));
+	else {
+		ModifyPlrStr(Players[pnum], 255);
+		ModifyPlrMag(Players[pnum], 255);
+		ModifyPlrDex(Players[pnum], 255);
+		ModifyPlrVit(Players[pnum], 255);
```
StephenCWills, line 2282: I don't think that `CMD_CHEAT_STATS` is necessary in this case, since these functions to modify player stats send their own network packets. You can just have the debug function call these directly.

Source/msg.h
```diff
@@ -198,6 +198,10 @@ enum _cmd_id : uint8_t {
 	// Cheat: increase active spell level of player.
 	//
 	// body (TCmd)
+	CMD_CHEAT_STATS,
+	// Cheat: max base stats of player.
+	//
+	// body (TCmd)
```
StephenCWills, line 204: If you remove the network message, you don't need to worry about this, but the comment should be above the corresponding enum value.

StephenCWills (changes requested):

StephenCWills (approved):

PR #5540, opened 2022-11-27
Multiplayer: Enable Arkaine's Valor Quest

obligaron:
Contributes to #926

- The pedestal of blood (stone) can be used multiple times (multiple stones get added). But in multiplayer only one operate object is stored/synced. That's why qvar2 is used to store the added stones (not in singleplayer).


Source/objects.cpp
```diff
@@ -4488,7 +4512,7 @@ void DeltaSyncOpObject(Object &object)
 	case OBJ_BLINDBOOK:
 	case OBJ_BLOODBOOK:
 	case OBJ_STEELTOME:
-		UpdateState(object, object._oVar6);
+		object._oAnimFrame = object._oVar6;
```
StephenCWills, line 4515: Nice catch.

StephenCWills (approved):

PR #1548, opened 2021-04-15
fancy stairs up/down

qndel:
image

all that's left is mapping all stairs up for catas/caves/hell/crypt/hive 🎉 🤮 
If anyone finds a way that doesn't require mapping all stairs manually, let me know :D

Issue #1362, opened 2021-04-04
Differentiate upstairs from downstairs on the automap


julealgon:
## Description

Stairs are currently represented using the same graphics on the automap, regardless if they go up, or down. This makes it harder to quickly tell whether a given stair on the map goes up or down without memorizing their positions a priori.

This is poor user experience and should be improved by indicating whether a stair goes up or down on the automap.

## Details
- Try to use raw graphics elements like the rest of the automap for consistency
- Discuss potential symbols to use to denote up or down

No feature flag should be needed for this as it is a simple QOL change that doesn't impact gameplay.

AJenbo:
I sent this a duplicate issue?

julealgon:
> I sent this a duplicate issue?

What?

AJenbo:
i think i need to drop grammarly...

anyway i guess we only talked about it on chat in the past

PR #5546, opened 2022-11-28
Automap: Differentiate stair types

glebm:
This is a re-implementation of @qndel's #1548 in the current codebase.

A current implementation should make it easier to implement whatever UI we decide on.

image

Refs #1362

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -29,6 +29,8 @@ enum MapColors : uint8_t {
 	MapColorsPlayer = (PAL8_ORANGE + 1),
 	/** color for bright map lines (doors, stairs etc.) */
 	MapColorsBright = PAL8_YELLOW,
+	/** color fo the stairs type indicator line */
+	MapColorsAccent = PAL8_RED,
 	/** color for dim map lines/dots */
 	MapColorsDim = (PAL16_YELLOW + 8),
 	/** color for items on automap */
@@ -65,7 +67,7 @@ struct AutomapTile {
 		RiverRightOut,
 	};
 
-	Types type;
+	Types type = Types::None;
 
 	/** Additional details about the given tile */
 	enum class Flags : uint8_t {
@@ -83,7 +85,7 @@ struct AutomapTile {
 		// clang-format on
 	};
 
-	Flags flags;
+	Flags flags = {};
 
 	constexpr bool HasFlag(Flags test) const
 	{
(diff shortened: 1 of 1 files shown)


Source/automap.cpp
```diff
@@ -29,6 +29,8 @@ enum MapColors : uint8_t {
 	MapColorsPlayer = (PAL8_ORANGE + 1),
 	/** color for bright map lines (doors, stairs etc.) */
 	MapColorsBright = PAL8_YELLOW,
+	/** color fo the stairs type indicator line */
```
qndel, line 32:
```suggestion
	/** color of the stairs type indicator line */
```

Source/automap.cpp
```diff
@@ -326,14 +328,95 @@ void DrawRiverForkOut(const Surface &out, Point center, uint8_t color)
...
+
+	// Hell: Parts of a staircase down
+	case 39: return StairsType::Down;
+	case 40: return StairsType::Part;
+	case 42: return StairsType::Down;
+	case 43:
+	case 44: return StairsType::Part;
+	case 46: return StairsType::Down;
+
+	// Parts of a staircase to town
```
qndel, line 390:
```suggestion
	// Hell: Parts of a staircase to town

```

julealgon:
> until someone send a PR with better graphics?

Is this even _possible_, though? Wouldn't it make sense to first add some sort of support for sprite-based map elements or something like that?

Right now these "graphics" are 100% raw line draw calls right? It's not an image that you can set or anything like that.

On another note, what about adding a legend to the map? Maybe a way to say "blue == Upstairs, red == Downstairs", have it on by default, and allow it to be toggled off for people who are already used to the conventions. Just a thought...

glebm:
> On another note, what about adding a legend to the map?

Yeah that's what I mean, let's punt any improvements to a future PR.

The idea is that having nothing at all is strictly worse than having some differentiation, even if confusing.

> Is this even possible, though? Wouldn't it make sense to first add some sort of support for sprite-based map elements or something like that?

It is entirely possible to draw a sprite there, simply calling `ClxDrawBlendedTRN` will render a sprite with 50% transparency (a version without `TRN` can be easily added). However, there is no support for zooming the sprite together with the map.

glebm:
Retargeted to `development` and enabled auto-merge

glebm:
Superseded by https://github.com/diasurgical/devilutionX/pull/6517

PR #5553, opened 2022-11-30
dun_render: Overhaul mask handling

glebm:
We notice that masks can be described by 2 parameters:

1. Whether they have 0 or 1 as their high bits.
2. Whether they shift to the left or to the right on the next line.

Describing masks this way allows us to lift them to template variables and simplify the code.
We also avoid handling the mask in the `RenderLine` loop entirely.

We also note that some masks are only valid for certain tile types:

The `Left` mask can only be applied to `LeftTrapezoid` and `TransparentSquare`.
The `Right` mask can only be applied to `RightTrapezoid` and `TransparentSquare`.
The `Left/RightFoliage` masks can only be applied to `TransparentSquare`.

We now only instantiate the render function templates for the valid tile type / mask combinations.

Also fixes a foliage rendering bug: Transparent foliage pixels were previously blended but they should have been simply skipped.

### DIFF
--- a/Source/engine/render/scrollrt.cpp
+++ b/Source/engine/render/scrollrt.cpp
@@ -54,6 +54,10 @@
 #include "debug.h"
 #endif
 
+#ifdef DUN_RENDER_STATS
+#include "utils/format_int.hpp"
+#endif
+
 namespace devilution {
 
 /**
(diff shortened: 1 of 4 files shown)


Source/engine/render/scrollrt.cpp
```diff
@@ -1171,11 +1171,21 @@ void DrawFPS(const Surface &out)
...
 		lastFpsUpdateInMs = runtimeInMs;
+// #define FRAC_FPS
+#ifdef FRAC_FPS
+		int framerate = 10000 * framesSinceLastUpdate / msSinceLastUpdate;
+#else
 		int framerate = 1000 * framesSinceLastUpdate / msSinceLastUpdate;
+#endif
 		framesSinceLastUpdate = 0;
 
+#ifdef FRAC_FPS
```
AJenbo, line 1182: just make it permanent instead

AJenbo (approved):

PR #5554, opened 2022-12-02
Simplify search for droppable item index

obligaron:
While experimenting some other changes I discovered that we can simplify the logging around finding droppable item indexes.

- Extract common logic to `GetItemIndexForDroppableItem`.
- Reduces memory usage, cause we only have one static/global item index list (`ril`) (previous we had 8 instances).
- Replicates vanilla bug for resurrect (see devilution [`RndItem`](https://github.com/diasurgical/devilution/blob/master/Source/items.cpp#L2667-L2673) and [`RndAllItems`](https://github.com/diasurgical/devilution/blob/master/Source/items.cpp#L2754-L2757). Note: devilutionX had the bug removed for `RndAllItems` with this pr it was reintroduced for same behavior as vanilla, but we can also decide to remove it entirely.


StephenCWills:
I'm not sure that the bug is significant enough to justify keeping or reintroducing it.

AJenbo:
It would actually be preferable to fix this bug. Since it happens at level generation it won't cause any BC issues.

obligaron:
The bug is no removed and the timedemo is updated.
I also added some improvements to the timedemo (in separated commits).

AJenbo:
Nice! I repacked the reference save to reduce it by 220kb. Here are the steps to do so if anyone else wants to do so in the future:

save.txt
```
additionalMissiles
game
hero
perml00
perml01
```

```bash
smpq -x -L save.txt demo_0_reference_spawn_0.sv
```
delete any files not in save.txt

```bash
smpq -c demo_0_reference_spawn_0.sv *
```

It reduces the file by leaving out unused data and avoiding bad compression, but it it also makes the save only compatible with devilutionx (Vanilla has hardcoded some parameters for the MPQ archive when reading save files).

AJenbo (approved):

glebm:
I'll rebase #5605 and #5604 once this is merged

AJenbo:
Was just about to mention it :)

obligaron:
Could we add the implement the logic for reducing save game size directly in devilutionX?

AJenbo:
Not sure, the MPQ writing logic is pretty primitive and it's only relevant for timedemo reference saves as the resulting save games is not backwards compatible.

If we do we should probably fix it so that it doesn't encrypt before compressing, that would make it even smaller.

PR #5555, opened 2022-12-02
Hopefully fix Amiga build.

arczi84:
Apparently docker image bug build bad executables is fixed now and this patch actually makes it faulty again.


AJenbo:
Have you tested that it works now?

arczi84:
It works locally.

AJenbo:
What if you download the one build by the CI :D

arczi84:
I did but still doesn't work. No idea..

PR #5557, opened 2022-12-03
Show FPS up to 1 decimal point

glebm:
If FPS < 100, print it with 1 decimal point.
Useful for measuring performance impact on low-end devices.


AJenbo:
if vsync is enabled this will be true for most. maybe set it to 50?

StephenCWills (approved):
Just tried it with V-sync, and it seems fine to me. Basing the level of precision on the number of digits that need to be displayed is a sensible choice.

PR #5563, opened 2022-12-04
dun_render: Avoid `Invalid` tile type

glebm:
Do not call RenderTile with invalid values to begin with.


AJenbo:
Nice, I was wondering if we should do this when I saw your last one :)

AJenbo (dismissed):

AJenbo (approved):

PR #5582, opened 2022-12-08
Optimize IsTileSafe

glebm:
Add dungeon flags for Fire/Lightning Wall and use them to check if the monster considers the tile safe.

### DIFF
--- a/Source/levels/gendung.h
+++ b/Source/levels/gendung.h
@@ -84,16 +84,17 @@ enum lvl_entry : uint8_t {
 
 enum class DungeonFlag : uint8_t {
 	// clang-format off
-	None        = 0, // Only used by lighting/automap
-	Missile     = 1 << 0,
-	Visible     = 1 << 1,
-	DeadPlayer  = 1 << 2,
-	Populated   = 1 << 3,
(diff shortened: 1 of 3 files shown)


StephenCWills (approved):
This looks good to me. Unless we had any alternative thoughts for how to use the remaining dungeon flags.

AJenbo:
On a side note this appears to remove the only use of `DungeonFlag::Missile` so i think we could remove that :)

StephenCWills:
It's used when placing runes, Town Portal, and Guardian as well.

AJenbo:
But is it ever read/tested? I only saw writes/set.

StephenCWills:
Each of these functions reads the flag to avoid placing them in the same tile as another missile.

`AddRune()`

https://github.com/diasurgical/devilutionX/blob/5fa554c2e638fc8631044af10bfd73ad0ab01233/Source/missiles.cpp#L586-L588

`AddTown()`

https://github.com/diasurgical/devilutionX/blob/5fa554c2e638fc8631044af10bfd73ad0ab01233/Source/missiles.cpp#L1917-L1919

`AddGuardian()`

https://github.com/diasurgical/devilutionX/blob/5fa554c2e638fc8631044af10bfd73ad0ab01233/Source/missiles.cpp#L2031-L2033

AJenbo:
Arh I missed that, thanks :)

Issue #5573, opened 2022-12-06
Test for packing and unpacking items is failing on BE systems


AJenbo:
### Operating System

Linux PowerPC

### DevilutionX version

https://github.com/diasurgical/devilutionX/commit/44c3bc403859e0d8b129fecd1a3c381a5e9a3f1f

### Describe

Most pack tests are filing on PowerPC

```
[ RUN      ] PackTest.UnPackItem_gold_small
test/pack_test.cpp:765: Failure
Expected equality of these values:
  id._ivalue
    Which is: 1000
  wvalue
    Which is: 59395
[  FAILED  ] PackTest.UnPackItem_gold_small (2 ms)
[ RUN      ] PackTest.UnPackItem_gold_medium
test/pack_test.cpp:765: Failure
Expected equality of these values:
  id._ivalue
    Which is: 1001
  wvalue
    Which is: 59651
[  FAILED  ] PackTest.UnPackItem_gold_medium (1 ms)
[ RUN      ] PackTest.UnPackItem_gold_large
test/pack_test.cpp:765: Failure
Expected equality of these values:
  id._ivalue
    Which is: 2500
  wvalue
    Which is: 50185
[  FAILED  ] PackTest.UnPackItem_gold_large (1 ms)
[ RUN      ] PackTest.UnPackItem_ear
test/pack_test.cpp:40: Failure
Expected equality of these values:
  item1_iSeed
    Which is: 1633955154
  item2_iSeed
    Which is: 1378706529
test/pack_test.cpp:45: Failure
Expected equality of these values:
  item1_iCreateInfo
    Which is: 17509
  item2_iCreateInfo
    Which is: 25924
[  FAILED  ] PackTest.UnPackItem_ear (2 ms)
```

### To Reproduce

./pack_test

glebm:
Packed items should use little-endian format even on big-endian systems.

The test fixtures are currently defined with native-endian format.
They instead should use little-endian format.

There may be other endian-related issues in item packing but step one is to fix the test fixtures so that we know what the other issue are

glebm:
The test failures above can also be seen in the CI run: https://github.com/diasurgical/devilutionX/actions/runs/3646395402/jobs/6157431868

PR #5583, opened 2022-12-08
More item packing fixes on big-endian

glebm:
Refs #5573


AJenbo (approved):
Nice that they where not all false positives :)

PR #5584, opened 2022-12-08
Pop test runner cache

AJenbo:


StephenCWills (approved):
I assume maybe we're still waiting on results from Codecov, but the change looks fine.

glebm:
Looks like the workaround is to set the token
Bug: https://github.com/codecov/codecov-action/issues/837
Example workaround https://github.com/sensuikan1973/pedax/pull/1033

AJenbo:
Pressed all the hidden "auth" buttons in GH and CC. The repo association looks to be fixed now :)

AJenbo:
Test coverage is now 29.10%, so the new demo increased it by about 2% it looks like.

PR #5586, opened 2022-12-08
Move item info from panel to Unique window

kphoenix137:
This PR repurposes the Unique Item Info Panel to display all item powers within the Unique window regardless of item quality. The control panel display continues to display base item details (item name, durability, item quality, armor, damage). I opted for this solution over transforming the control panel information into a floating box (a la Diablo 2) since such a method would render the obtuse black square obsolete, making it feel very confusing to players why it's there. This solution appears to be the least invasive way of displaying more item information without changing too much.

The purpose of this PR is to allow for more lines of text for item details. The hidden bonus for sword and maces VS animals and undead are additionally displayed in the window.

image
image
image
image
image

### DIFF
--- a/Source/cursor.cpp
+++ b/Source/cursor.cpp
@@ -387,7 +387,7 @@ void CheckCursMove()
 	pcursinvitem = -1;
 	pcursstashitem = uint16_t(-1);
 	pcursplr = -1;
-	ShowUniqueItemInfoBox = false;
+	ShowItemInfoBox = false;
 	panelflag = false;
 	trigflag = false;
 
--- a/Source/engine/render/scrollrt.cpp
+++ b/Source/engine/render/scrollrt.cpp
@@ -1116,8 +1116,8 @@ void DrawView(const Surface &out, Point startPosition)
 		DrawStash(out);
 	}
 	DrawLevelUpIcon(out);
-	if (ShowUniqueItemInfoBox) {
-		DrawUniqueInfo(out);
+	if (ShowItemInfoBox) {
+		DrawItemInfo(out);
 	}
 	if (qtextflag) {
 		DrawQText(out);
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1922,7 +1922,7 @@ int8_t CheckInvHLight()
 		InfoString = fmt::format(fmt::runtime(ngettext("{:s} gold piece", "{:s} gold pieces", nGold)), FormatInteger(nGold));
 	} else {
 		InfoColor = pi->getTextColor();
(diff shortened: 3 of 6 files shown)


glebm:
I'd down for this as a QoL addition. It also paves the way forward for a more compact infoboxless UI.
For that, the stat requirements should also go into the unique box.

AJenbo:
Agreed, I was secretly hoping KP would do something like this :)

obligaron:
Could me make this toggle able?
This changes the UI a lot (purist) but also give more info (great).
So I can't decide if I like this changes or not. 😝

kphoenix137:
> Could me make this toggle able?
> This changes the UI a lot (purist) but also give more info (great).
> So I can't decide if I like this changes or not. 😝 

Yes

kphoenix137:
I'm starting to feel like maybe it would be optimal for this PR to completely migrate all the item info into the top box and not use the control panel

AJenbo:
agree, toggle and move the info

kphoenix137:
> > such a method would render the obtuse black square obsolete, making it feel very confusing to players why it's there
> 
> Doesn't this box still show monster info? You definitely want a dynamically sized box that hovers over the inventory for less obstruction of the game world. Look at your cap screenshot.

I will create a new pr and experiment with the d2 style and we can then decide on one or the other, or neither

AJenbo:
> > such a method would render the obtuse black square obsolete, making it feel very confusing to players why it's there
> 
> Doesn't this box still show monster info? You definitely want a dynamically sized box that hovers over the inventory for less obstruction of the game world. Look at your cap screenshot.

One step at a time

obligaron:
Perhaps we should simply combine this with #5456? 🤔 

New setting "control panel" with currently two options:
- classic
- modern (or compact or something like this)

"Modern" would change the control panel and the info panel.

kphoenix137:
> Perhaps we should simply combine this with #5456? 🤔 
> 
> New setting "control panel" with currently two options:
> - classic
> - modern (or compact or something like this)
> 
> "Modern" would change the control panel and the info panel.

Good point. That pr is so outdated that I'm just going to create a new branch for this and close both of these PRs.
@AJenbo should I wait for Lua to create this PR, or do you think this is something you'd want prior to Lua?

AJenbo:
I thin this is fine without lua, but they should be kept as two PRs to make the introduction more gradual and reviews more manageable.

glebm:
This is starting to look good:

1. Please make this a setting!
2. Let's not use the infobox at all when the setting is on.

PR #5601, opened 2022-12-13
Update SDL2

glebm:
The new version fixes audio resampler quality issues, improves support for various gamepads, and fixes some threading issues on the 3DS.

I did not check for Android Java code updates -- some may be necessary

PR #5615, opened 2022-12-17
Update SDL2 and SDL2 Android code

glebm:
The previous SDL2 update in #5601 did not include the updated Android code.

This one does and should hopefully fix the Android startup issue pointed out by @Trihedraf in #5601.

With this new version, we get further improved gamepad support and audio device selection on Android.


AJenbo (approved):
Thanks, I noticed previously that SDL now checks for version matching between the Java and C components. So I expected some issues from only upgrading one, but didn't get around to fixing it yet :)

Issue #628, opened 2020-02-18
Unique Item Generation Bug (with possible fix)


cain05:
Hello!  The unique item generation bug in Diablo has always bothered me, so I made an attempt to fix it.  I thought I would share this information here in hope that it can aid others in possibly coming up with a more elegant solution.  For those not familiar with the bug, unique items with the same base type and ilvl, only one ever spawn in multiplayer games.  Even if you were to hack the item into your inventory, it would morph once you started a new game.

There's actually three parts to the bug, therefore three changes are required to get it working correctly.

Fix 1 - Fix UniqueItemList in itemdat.cpp so that there are no UIItemId and UIMinLvl conflicts.  This change is pretty straight forward.  Using UITYPE_LARGEAXE as an example, The Mangler and Sharp Beak both have a UIMinLvl of 2.  Because of the way unique items are selected (fix 2), The Mangler would never drop.  Simple fix was to give Sharp Beak a UIMinLvl of 3.

Fix 2 - Pick a random unique from the list of possible uniques.  This fix is also straight forward.  The way the game currently works is that if more than one unique of a given UIItemId is to be spawned, it always picks that last item in the order of indices in UniqueItemList.  This is why Sharp Beak would always spawn when the UIMinLvl was the same as The Manger.  I modified the selection code to pick an item at random when then generation routine was not set to "recreate".  Now that the The Mangler can be selected and spawned it would morph back in to Sharp Beak when a new game would be started.

Fix 3 - Stop items from morphing. This one took me a bit to figure out what was going on.  From what I can tell, the game only stores the UIItemId and the level of the monster that was killed which spawned the item.  So when a new game is started, unique items are regenerated.  So for example, if you kill a monster that drops UIMinLvl 10, when the items were re-created, it would go back to picking between The Mangler and Sharp Beak at random from fix 2.  What I ended up doing was storing the UIMinLvl of the item instead, and then used the old unique selection code (which always picks the last item in the list) when the "recreate" flag was set.  This code change I wasn't 100% sure it was the right change, but it seems to work.

Note that these changes are based on release 0.5.  When I merged in the 1.0 changes, I think something broke, or I screwed up the merge on my local repository.  I haven't had much time to look at it.

Relevant commits:
https://github.com/cain05/devilutionX/commit/b294140e70d348a02e09ed79e6291024dfca74f4
https://github.com/cain05/devilutionX/commit/79fb1f3024920936c48caca47433488bc1196062

[...]

PR #5619, opened 2022-12-17
In multiplayer randomly drop unique quest reward or magic item of same type

obligaron:
This pr changes how unique quest rewards (like Valor, Griswold's Edge or Bovine Plate) drop in multiplayer. Normal unique items that drop randomly from monsters are not affected.
The unique quest rewards only have a change of 25% to drop.
In the other 75% they drop a magic item of the same type (Sword, Axe, Helm, Light Armor, Medium Armor, ...).

Currently this is done for all unique quest rewards (including hellfire quest reward items like Bovine Plate).
The idea is to make the quests more interesting to play again in multiplayer and don't spam the unique quest reward items.

This pr could also be changed
- change drop change
- enable this only for the diablo unique quest rewards
- even change it to never drop unique quest reward uniques in multiplayer

This is up to a discussion.
I hope this pr helps to give us more options how to proceed with this topic (unique quest rewards). 🙂


qndel:
Cool idea!

AJenbo:
I like this, but I would do it 100% of the time in multiplayer :)

obligaron:
I would like something less then 100% 😉 
Cause I would like to see these items also in multiplayer. They add something to the game and make the quest more interesting.
How we want to define the change of dropping them is another topic. I would also be okay with 98% magic items and only 2% unique quest rewards. 😁

qndel:
> I would like something less then 100% 😉 Cause I would like to see these items also in multiplayer. They add something to the game and make the quest more interesting. How we want to define the change of dropping them is another topic. I would also be okay with 98% magic items and only 2% unique quest rewards. 😁

That'd be best imo :D

kphoenix137:
> I would like something less then 100% 😉 Cause I would like to see these items also in multiplayer. They add something to the game and make the quest more interesting. How we want to define the change of dropping them is another topic. I would also be okay with 98% magic items and only 2% unique quest rewards. 😁

I personally would like to see quest reward items dropped 100% on the first completion of the quest in multiplayer per character. Perhaps we could utilize some free space in the player struct with a bitmask? After the first quest completion, the game would generate a random magical/unique item. To avoid the issue of multiple players in a game and one quest reward dropping, a quest reward could be dropped for each player that hasn't completed the quest. If at least one player in the game has already completed the quest, a single magical/unique item would drop in addition to a quest item for each player that hasn't completed the quest.

For quests that have items already on the ground, maybe we can base it on the host's completion, and prevent the completion bit being set for non-host players when the quest is completed. (I think this only applies to Arkaine's Valor)

AJenbo:
I have been convinced that something like this is not bad, just have to think some more about how exactly it should work :)

kphoenix137:
> I would like something less then 100% 😉 Cause I would like to see these items also in multiplayer. They add something to the game and make the quest more interesting. How we want to define the change of dropping them is another topic. I would also be okay with 98% magic items and only 2% unique quest rewards. 😁

What about using unique monster item generation, (85% magical item, 15% unique item) with sp uniques having equal weight to all other uniques

fuckbeamdog:
I have a range of ideas really. One idea i had was to have some of the quests have the possibility of dropping "undropable uniques" as an alternative. Another thought was to have some of the higher tier quests drop "wirt only items". I also had the thought of their loot table only consisting of "logical" items - For example the butcher only dropping axes.

One worry i have is that some items will be more difficult to acquire after implementing this. The helm of spirits is most easily farmed on nm butcher in vanilla, although that obviously matters less if the undead crown is available.

My first thought was that they only drop the actual quest unique the first time you complete the quest with each character.

Some specific quest reward ideas:
**Black mushroom**
First time you get the Spectral elixir, second time you have a drop table ala:
- 10% chance of getting three random elixirs
- 25% chance of getting two random elixirs
- 65% chance of getting one random elixir

**Anvil of fury**
First time you get Griswold's edge. Second time it will give you a high tier weapon base item with high tier mods, possibly even wirt-only "Strange".

**Diablo**
This one is a bit strange, but i thought about a few high tier items just appearing in town next game after defeating him. It's a bit sad that there is currently no point in killing Diablo.

Just tossing ideas out there.

obligaron:
This pr only drops the same type as the original unique item (Sword, Axe, Helm, Light Armor, Medium Armor, ...).
So it's similar to what you described. 😉 

For the undroppable uniques we have a issue #628 (and also a pr open #2930)

For black mushroom we have a discussion in #5770 😉 

For everything "drop only once per play" It's not that easy 😢
Not every quest reward is triggered by a player, for example arkain's valor and halls of the blind (the items are on the ground when generating the level).
Another issue is that one player could "steal" the item from the other. But this could be more minor. And we need to store the state in a additional save game. But that should be doable.

For diablo:
This is already a issue in singleplayer and I think another topic. I checked if we have a issue for it, but couldn't find one. I only remember we discussed it somewhere (on github or discord) and most people thought it was to much of a mod change. But feel free to open a issue. 🙂

D1-Constantine:
> I would like something less then 100% wink Cause I would like to see these items also in multiplayer. They add something to the game and make the quest more interesting. How we want to define the change of dropping them is another topic. I would also be okay with 98% magic items and only 2% unique quest rewards. grin

Yeah I like that idea a lot. I think if Blizzard ever made quest for multiplayer they would have integrated them somehow. If you get Leorics Crown or Veil of Steel every time those quest become useless as you won't do them more than once. Repetition and farming is the name of the game in multiplayer. I think having less than 5% for the quest uniques rewards is more than generous as you'll probably do them thousands of times.

fuckbeamdog:
> For everything "drop only once per play" It's not that easy 😢 Not every quest reward is triggered by a player, for example arkain's valor and halls of the blind (the items are on the ground when generating the level). Another issue is that one player could "steal" the item from the other. But this could be more minor. And we need to store the state in a additional save game. But that should be doable.

I thought about checking the character and stash to see if they already have the quest reward in possession?

Source/items.cpp
```diff
@@ -3219,9 +3219,19 @@ Item *SpawnUnique(_unique_items uid, Point position, bool sendmsg /*= true*/)
...
+		GetUniqueItem(*MyPlayer, item, uid);
+		SetupItem(item);
+	} else {
+		if (level)
+			curlv = *level;
+		const ItemData &uniqueItemData = AllItemsList[idx];
+		_item_indexes idx = GetItemIndexForDroppableItem(false, [&uniqueItemData](const ItemData &item) {
+			return item.itype == uniqueItemData.itype;
+		});
+		SetupAllItems(*MyPlayer, item, idx, AdvanceRndSeed(), curlv * 2, 15, true, false, false);
```
AJenbo, line 3233: Would it work to set a 0% chance of unique so that the unique is the quest item?
```suggestion
		SetupAllItems(*MyPlayer, item, idx, AdvanceRndSeed(), curlv * 2, 0, true, false, false);
```

Source/items.cpp
```diff
@@ -3219,9 +3219,19 @@ Item *SpawnUnique(_unique_items uid, Point position, bool sendmsg /*= true*/)
 	while (AllItemsList[idx].iItemId != UniqueItems[uid].UIItemId)
 		idx++;
 
-	GetItemAttrs(item, static_cast<_item_indexes>(idx), curlv);
-	GetUniqueItem(*MyPlayer, item, uid);
-	SetupItem(item);
+	if (!gbIsMultiplayer || FlipCoin(15)) {
```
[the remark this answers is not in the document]
AJenbo:
```suggestion
	if (!gbIsMultiplayer || GenerateRnd(100) < 15) {
```

AJenbo:
With the two suggested changes and a rebase I'm all for merging this for 1.5.0 :)

AJenbo (approved):

Source/items.cpp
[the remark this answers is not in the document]
obligaron: Fixed 🙂

Source/items.cpp
obligaron: When changing `uper` to 0, we will also reduce the item base level `GetItemBLevel` and affect magical items.
Also the result could still be a unique, cause `GenerateRnd` could return 0 and the check in `CheckUnique` would still succeed (but quite unlikely).

Another idea could be to put `SetupAllItems` (inklusive `AdvanceRndSeed`) in a loop and run the loop until the item is not a unique.

Personally I wouldn't mind if another unique item is dropped/rewarded.

AJenbo: Ok, would 1 also have negative side effects on the generated item? I'm ok with it potentially being another item, but preferable a very low chance.

Source/items.cpp
```diff
@@ -3219,9 +3219,19 @@ Item *SpawnUnique(_unique_items uid, Point position, bool sendmsg /*= true*/)
...
+		GetUniqueItem(*MyPlayer, item, uid);
+		SetupItem(item);
+	} else {
+		if (level)
+			curlv = *level;
+		const ItemData &uniqueItemData = AllItemsList[idx];
+		_item_indexes idx = GetItemIndexForDroppableItem(false, [&uniqueItemData](const ItemData &item) {
+			return item.itype == uniqueItemData.itype;
+		});
+		SetupAllItems(*MyPlayer, item, idx, AdvanceRndSeed(), curlv * 2, 15, true, false, false);
```
AJenbo, line 3233:
```suggestion
	if (gbIsMultiplayer) {
		if (level)
			curlv = *level;
		const ItemData &uniqueItemData = AllItemsList[idx];
		_item_indexes idx = GetItemIndexForDroppableItem(false, [&uniqueItemData](const ItemData &item) {
			return item.itype == uniqueItemData.itype;
		});
		SetupAllItems(*MyPlayer, item, idx, AdvanceRndSeed(), curlv * 2, 15, true, false, false);
	}
	if (!gbIsMultiplayer || item._iMagical == ITEM_QUALITY_UNIQUE) {
		GetItemAttrs(item, static_cast<_item_indexes>(idx), curlv);
		GetUniqueItem(*MyPlayer, item, uid);
		SetupItem(item);
```

obligaron: I really like the idea, because we stick with the normal random loot logic instead of creating our own version of it.

I tested, if all uniques could be generated for the giving dungeon level where the quest is located.
But optic amulet don't drop in diablo, because there are no unique amulets in diablo that could be replaced. 😢

AJenbo: Should we just let anything drop of amulet ?

obligaron: You mean changing
```cpp
			return item.itype == uniqueItemData.itype;`
```
to
```cpp
			if (IsAnyOf(item.itype, ItemType::Gold, ItemType::Misc))
				return false;
			if (uniqueItemData.itype == ItemType::Amulet)
				return true;
			return item.itype == uniqueItemData.itype;
```
?
That would work (tested it with optic amulet in diablo).

We could also allow this for all uniques. But this would sometimes make little sense (Grisworlds Edge for example).

AJenbo (approved):

ikonomov:
Yes! :)

ikonomov:
I've been thinking about this for a while and came up with an idea that maybe worth considering as well for the condition of the quest item:

`int qitem = GenerateRnd(3);
if (sgGameInitInfo.nDifficulty == DIFF_NORMAL || (item._iMagical == ITEM_QUALITY_UNIQUE && qitem == 0))`

PR #5621, opened 2022-12-17
Fix OOB in ConsumeScroll()

AJenbo:


Source/inv.cpp
```diff
@@ -1942,13 +1942,13 @@ void ConsumeScroll(Player &player)
 	int itemSlot = player.executedSpell.spellFrom;
 	int itemIndex = 0;
 	Item *item;
```
obligaron, line 1944:
```suggestion
	Item *item = nullptr;
```

obligaron: Then we have to check `nullptr` later in the function.
Previous:
`if (!item->isEmpty() && isCurrentSpell(*item)) {`
Changed:
`if (item != nullptr && !item->isEmpty() && isCurrentSpell(*item)) {`

Source/inv.cpp
```diff
@@ -1942,13 +1942,13 @@ void ConsumeScroll(Player &player)
...
-		itemIndex = itemSlot - INVITEM_INV_FIRST;
-		item = &player.InvList[itemIndex];
-	} else {
-		itemIndex = itemSlot - INVITEM_BELT_FIRST;
-		item = &player.SpdList[itemIndex];
-	}
+        if (itemSlot >= INVITEM_INV_FIRST && itemSlot <= INVITEM_INV_LAST) {
+                itemIndex = itemSlot - INVITEM_INV_FIRST;
+                item = &player.InvList[itemIndex];
+        } else if (itemSlot >= INVITEM_BELT_FIRST) {
```
obligaron, line 1948: Should we also check for `INVITEM_BELT_LAST`? 🤔

Source/inv.cpp
```diff
@@ -1942,13 +1942,13 @@ void ConsumeScroll(Player &player)
...
+                itemIndex = itemSlot - INVITEM_BELT_FIRST;
+                item = &player.SpdList[itemIndex];
+        }
```
glebm, line 1951: else `app_fatal` because this should never happen

obligaron: Currently this can happen (0 is set) when using the spell list to cast a scroll.
Cause in this case no explicit scroll was selected/used.

glebm: Ah, makes sense

glebm:
Let's merge #5622, which adds 2 more tests and includes additional checks. It also calls `app_fatal` on an invalid item index.

PR #5622, opened 2022-12-18
  Fix ConsumeScroll

### LINKED ISSUE #5617: OOB access in ConsumeScroll()
### Operating System

Linux PowerPc

### DevilutionX version

https://github.com/diasurgical/devilutionX/commit/96c7b154be835bddcd80419163af2680a2944ec5

### Describe

```
[ RUN      ] InvTest.RemoveCurrentSpellScroll_inventory
/home/ajenbo/devilutionX/Source/inv.cpp:1947:35: runtime error: index -7 out of bounds for type 'Item [40]'
```

Source/inv.cpp:1947
```cpp
        int itemSlot = player.executedSpell.spellFrom;
        int itemIndex = 0;
        Item *item;
        if (itemSlot <= INVITEM_INV_LAST) {
                itemIndex = itemSlot - INVITEM_INV_FIRST;
                item = &player.InvList[itemIndex]; // <- 0 - 7
        } else {
                itemIndex = itemSlot - INVITEM_BELT_FIRST;
                item = &player.SpdList[itemIndex];
        }

```

### To Reproduce

run ./inv_test

glebm:
Fixes #5617

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1938,28 +1938,28 @@ void ConsumeScroll(Player &player)
 {
 	const spell_id spellId = player.executedSpell.spellId;
 
-	// Try to remove the scroll from selected inventory slot
-	int itemSlot = player.executedSpell.spellFrom;
-	int itemIndex = 0;
-	Item *item;
-	if (itemSlot <= INVITEM_INV_LAST) {
-		itemIndex = itemSlot - INVITEM_INV_FIRST;
-		item = &player.InvList[itemIndex];
-	} else {
-		itemIndex = itemSlot - INVITEM_BELT_FIRST;
-		item = &player.SpdList[itemIndex];
-	}
-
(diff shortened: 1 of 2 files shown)


AJenbo (approved):

obligaron:
I don't think this is the correct approach cause spellFrom can be 0 if you cast a scroll from spell list. 🤔 But I need to check again.

obligaron:
Check again: Using the spell list doesn't set `spellFrom`, so it can be 0.
That's why I would leave the unit tests as they are (or only change one of them) and would merge #5621.

glebm:
I see. We should test all 3 cases, and set it to 0 explicitly in the third case.

glebm:
I've added tests and incorporated the fixes from #5621

AJenbo (approved):
LGTM

PR #5629, opened 2022-12-19
Remove GNU style strip args

gancheff:
The --strip-all argument to the strip command breaks the build on BSD strip. Removing it altogether for now so it doesn't break non GNU builds. Possibly can be improved with selecting the proper argument depending on the implementation of strip.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,7 +307,7 @@ else()
         TARGET ${BIN_TARGET} POST_BUILD
         DEPENDS ${BIN_TARGET}
         COMMAND $<$<CONFIG:Release>:${CMAKE_STRIP}>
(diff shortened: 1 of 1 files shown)


glebm:
FreeBSD has it https://www.freebsd.org/cgi/man.cgi?query=strip&sektion=1&format=html
OpenBSD has it https://man.openbsd.org/strip.1
NetBSD has it https://man.netbsd.org/strip.1
DragonFly has it https://man.dragonflybsd.org/?command=strip&section=ANY
macOS does not have it

Since macOS is the only exception, let's just add a special case for it instead of removing it everywhere?

glebm:
Actually, IBM AIX strip does not have it either so it's fine to remove it by default
https://www.ibm.com/docs/en/aix/7.1?topic=s-strip-command

glebm (approved):

PR #5634, opened 2022-12-28
Multiplayer: Fix Poisoned Water Supply purify

obligaron:
Fixes that the water doesn't get purified for remote players.
The issue was that the kill count was checked on one client and the quest update message gets send and proceed on the remote client (and sets the new quest state) before the kill gets recognized on the remote client.
Thanks for reporting @kphoenix137 🙂


AJenbo (approved):
Nice, looks like things are coming together 😊

PR #5636, opened 2022-12-31
Clean up CheckReflect

AJenbo:


Source/monster.cpp
```diff
@@ -1114,7 +1114,6 @@ void CheckReflect(Monster &monster, Player &player, int &dam)
 	// reflects 20-30% damage
 	int mdam = dam * (GenerateRnd(10) + 20L) / 100;
 	ApplyMonsterDamage(monster, mdam);
-	dam = std::max(dam - mdam, 0);
```
ephphatha, line 1117: Doesn't this mean that players take full damage from an unblocked monster attack when reflect is up instead of the 70-80% they'd take previously? `dam` is taken by reference in this function.

AJenbo: Thanks i missed the by reference part, that's a bit tricky actually :)

PR #5637, opened 2022-12-31
Update CheckReflect to return applied damage

ephphatha:
Instead of modifying the damage value by reference

This does introduce a slight behaviour change, it was previously possible for reflect to only apply 12% of the incoming damage (`GenerateRnd(10)` returns `-8`, adding 20 gives 12). Now it'll always be between 20% and 29% inclusive.

### DIFF
--- a/Source/engine/random.hpp
+++ b/Source/engine/random.hpp
@@ -84,4 +84,28 @@ const T PickRandomlyAmong(const std::initializer_list<T> &values)
 	return *(values.begin() + index);
 }
 
+/**
+ * @brief Generates a random non-negative integer
+ *
+ * Effectively the same as GenerateRnd but will never return a negative value
+ * @param v upper limit for the return value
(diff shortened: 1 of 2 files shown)


AJenbo:
Thanks I was just thinking something like that would probably be better :)

I'm a bit fuzzy on the negative RNG issue, would `max(0)` not result in a bias towards `0` rather then `abs(-8)`?

ephphatha:
Treating the engine state as a signed value and taking the absolute value of it introduces a bias away from 0 (0 is only returned when the state/seed is 0, giving it a 1 in 2^32 chance of appearing while all other values have 2 in 2^32 chance of appearing). Clamping the negative value to zero corrects that bias.

AJenbo:
Thanks I had a feeling that was the case but the bug in RNG is still a bit of a black box to me :)

AJenbo (approved):

PR #5638, opened 2022-12-31
Remove No Monster Heal from Fire Arrows

kphoenix137:
This PR is meant to correct an alleged bug as a result of an alleged oversight in the development of Diablo.

The evidence:

To preface, the item power `ISPL_NOHEALMON` was originally used for the suffix "vileness" as shown here: https://gitlab.com/moralbacteria/diablo-prdemo-decomp/-/blob/main/src/itemdat.cpp on line 295. The intent was for this item power to set `MFLAG_NOHEAL`, preventing monsters from regenerating life.

This bug was corrected in Hellfire (https://github.com/diasurgical/devilution/blob/master/Source/missiles.cpp on line 753). 

Per Blighthorn and staphen, there is some evidence that this was an oversight. It appears that it's entirely possible that `plr[pnum]._pIFlags & MFLAG_NOHEAL` may have been used rather than `plr[pnum]._pIFlags & ISPL_NOHEALMON`. Since `_pIFlags` is not used for monster flags, this would have had the result of `ISPL_FIRE_ARROWS` being used rather than `ISPL_NOHEALMON`, since both `MFLAG_NOHEAL` and `ISPL_FIRE_ARROWS` resolve to `8`. (See https://github.com/diasurgical/devilution/blob/master/enums.h at line 3351 and line 1827).

There is a similar mistake found in https://github.com/diasurgical/devilution/blob/master/Source/player.cpp at line 3049, where the incorrect item power `ISPL_NOHEALPLR` is used rather than `ISPL_NOHEALMON`.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -247,9 +247,6 @@ bool MonsterMHit(int pnum, int monsterId, int mindam, int maxdam, int dist, miss
 	if (&player == MyPlayer)
 		ApplyMonsterDamage(monster, dam);
 
-	if (!gbIsHellfire && HasAnyOf(player._pIFlags, ItemSpecialEffect::FireArrows))
-		monster.flags |= MFLAG_NOHEAL;
-
 	if (monster.hitPoints >> 6 <= 0) {
 		M_StartKill(monster, player);
 	} else if (resist) {
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1381,12 +1381,6 @@ bool MonsterFadeout(Monster &monster)
  */
 void MonsterHeal(Monster &monster)
 {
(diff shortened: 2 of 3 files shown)


AJenbo (approved):

galaxyhaxz:
I did the opposite recently and restored the `vileness` suffix in my mod. I think the reason it was removed was because the flag to check it for both melee and missile weapons was wrong and it never worked, so they canned it before release since they didn't have time to fix/finish all the affixes. Evident by the fact `vileness` exists in Diablo II and does the same thing. The same applies to several other affixes like `spikes` that do more thorns damage, because they hardcoded thorns and didn't have any free item flag slots for both versions.

Of course, enabled/messing with any of the affixes will completely change all existing items.

PR #5639, opened 2023-01-01
Floating numbers

AJenbo:
This replaces #4580 as an initial implementation for floating numbers. The cleanups that have happened since it was originally created has helped simplify a few things. Additionally it replaces `enum missile_resistance` with `enum class DamageType` an breaks up the logic for calculating the worldtile to screen conversion which allows it to now have simpler and more accurate calculation for where to display the numbers. The code is still based on @qndel original work and retains him as the author.

Font colors are also reuses the existing system rather then implementing a new scheme. D1 Graphics Tool was helpful here for editing and previewing the TRN files.

This version does not have the many options and is instead simply on/off. I think it would be ok to introduce a few additional options like only showing number for high damage values. Maybe also toggles for healing and XP. But that can be done as a follow up once we have the basics merged. For that I would also prefer if we could spend some time breaking up the settings in to multiple pages and cleaning out some of other options.

One remaning issue is that the number will continue to move even when the game is paused since they are based on real time and not game time.


Trihedraf:
Just stating my preference, but the I don't like the numbers going out in different directions. The other PR with all the options there was the angle options which could be set to make the numbers go straight up which is much cleaner looking to me.

julealgon:
> Additionally it replaces `enum missile_resistance` with `enum class DamageType` an breaks up the logic for calculating the worldtile to screen conversion which allows it to now have simpler and more accurate calculation for where to display the numbers.

Would be nice if you could split these things into at least a separate commit.

AJenbo:
With the prep PRs now merged this is ready for review

Source/monster.cpp
```diff
@@ -3568,8 +3569,10 @@ void AddDoppelganger(Monster &monster)
 	}
 }
 
-void ApplyMonsterDamage(Monster &monster, int damage)
+void ApplyMonsterDamage(DamageType dmgType, Monster &monster, int damage)
```
StephenCWills, line 3572: Any reason we shouldn't spell all these out (`damageType`)?

StephenCWills (changes requested):

Source/qol/floatingnumbers.cpp
```diff
@@ -0,0 +1,198 @@
...
+		{ -100, -100 },
+	};
+
+	Displacement endOffset = goodAngles[rand() % 3];
+
+	if (damageToPlayer)
+		endOffset = -endOffset;
+
+	for (auto &num : FloatingQueue) {
+		if (num.reverseDirection == damageToPlayer && num.type == type && num.index == index && (SDL_GetTicks() - (int)num.lastMerge) <= 100) {
```
glebm, line 106: Use `static_cast`

Source/qol/floatingnumbers.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+		DrawString(out, floatingNum.text, Rectangle { screenPosition, { lineWidth, 0 } }, floatingNum.style);
+	}
+
+	ClearExpiredNumbers();
+}
+
+void ClearFloatingNumbers()
+{
+	srand(time(NULL));
```
glebm, line 193: Use `nullptr`

Source/engine/render/scrollrt.cpp
```diff
@@ -984,101 +985,94 @@ Displacement tileShift;
...
+	if (myPlayer.isWalking())
+		offset += GetOffsetForWalking(myPlayer.AnimInfo, myPlayer._pdir, true);
+
+	position += tileShift;
+
+	// Skip rendering parts covered by the panels
+	if (CanPanelsCoverView() && (IsLeftPanelOpen() || IsRightPanelOpen())) {
+		int multiplyer = (*sgOptions.Graphics.zoom) ? 1 : 2;
+		position += Displacement(Direction::East) * multiplyer;
+		offset.deltaX += -TILE_WIDTH * multiplyer / 2 / 2;
```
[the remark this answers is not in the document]
AJenbo: one for getting half the tile with, and one for half the resolution. I'm pretty sure the compiler will optimize it.

Source/qol/floatingnumbers.cpp
StephenCWills:
```suggestion
		if (num.reverseDirection == damageToPlayer && num.type == type && num.index == index && (SDL_GetTicks() - static_cast<int>(num.lastMerge)) <= 100) {
```

Source/qol/floatingnumbers.cpp
StephenCWills:
```suggestion
	srand(time(nullptr));
```

StephenCWills (dismissed):
Somehow, I broke all the GitHub Actions jobs with an error that makes no sense to me, but otherwise the PR looks good now. The only suggestion I have left is to maybe change `enableFloatingNumbers` to `floatingNumbersMode` with an enum for `None`, `Up`, and `Random`. That could satisfy @Trihedraf's request without introducing another separate setting.

AJenbo:
Probably because it spans the border between us upgrading the CI version

AJenbo:
> The only suggestion I have left is to maybe change `enableFloatingNumbers` to `floatingNumbersMode` with an enum for `None`, `Up`, and `Random`. That could satisfy @Trihedraf's request without introducing another separate setting.

I'm not dismissing this, but going to wait with it for a follow up

Source/monster.cpp
AJenbo: Fixed

Source/missiles.cpp
```diff
@@ -245,15 +246,15 @@ bool MonsterMHit(int pnum, int monsterId, int mindam, int maxdam, int dist, Miss
 		dam >>= 2;
 
 	if (&player == MyPlayer)
-		ApplyMonsterDamage(monster, dam);
+		ApplyMonsterDamage(MissilesData[static_cast<int8_t>(t)].damageType, monster, dam);
```
StephenCWills, line 249:
```suggestion
		ApplyMonsterDamage(missileData.damageType, monster, dam);
```

StephenCWills (changes requested):

StephenCWills (approved):

PR #5646, opened 2023-01-08
Reject games info of unexpected size

AJenbo:


Source/dvlnet/base_protocol.h
```diff
@@ -295,6 +295,8 @@ void base_protocol<P>::recv_decrypted(packet &pkt, endpoint_t sender)
 		size_t neededSize = sizeof(GameData) + (PlayerNameLength * MAX_PLRS);
 		if (pkt.Info().size() < neededSize)
 			return;
+		if (pkt.Info().data()[0] != sizeof(GameData))
+			return;
 		const GameData *gameData = (const GameData *)pkt.Info().data();
```
obligaron, line 300: `pkt.Info.data()` is defined as `unsigned char` but `GameData.size` is defined as `int32_t`.
That's why I think this check is not 100% correct. We only use the first byte of the `int32_t`.
For little-endian systems and the current size of `GameData` the difference should not matter but I think this would make it a little bit clearer:

```suggestion
		const GameData *gameData = (const GameData *)pkt.Info().data();
		if (gameData->size != sizeof(GameData))
			return;
```

Alternative we could change the check in line 296 from
`if (pkt.Info().size() < neededSize)`
to
`if (pkt.Info().size() == neededSize)`
cause currently the size is constant and always the same.

AJenbo: pkt.Info().size() == neededSize checks the actual package size, the package contains multiple structures so will be larger then game data. checking the size property lets us know that it is likely a version we understand and that we wont eat in to other structures

StephenCWills: So we should take obligaron's suggestion, right?

AJenbo: I think I got stuck on the text below the code sample, which wouldn't work (unless we also included names in the calculation for needed size)

AJenbo: The reason I didn't go with the suggestion originally was to not cast potential garbage to the structure, but we would be doing that anyway when assuming the first portion is somekind of number I guess.

obligaron (approved):

AJenbo:
Thanks :)

PR #5648, opened 2023-01-11
Removed ability to try and join incompatible ZeroTier games

DakkJaniels:
While the panel said the public game was incompatible, you could still try and join the game, which then caused the game you were trying to join to "hourglass" for the players in it. This only allows SNetJoinGame to be called for public games if host and client have compatible game versions.

Also fixed what may have been a bug with using *m_game_data (found by @StephenCWills), since it does not look like this was referring to the Gamelist gameData, but what is created locally.

### DIFF
--- a/Source/DiabloUI/selgame.cpp
+++ b/Source/DiabloUI/selgame.cpp
@@ -612,6 +612,9 @@ void selgame_Password_Select(int /*value*/)
 	SDL_ClearError();
 
 	if (selgame_selectedGame > 1) {
+		bool allowJoin = true;
+		if (selgame_selectedGame > 2)
+			allowJoin = IsGameCompatible(Gamelist[selgame_selectedGame - 3].gameData);
 		if (provider == SELCONN_ZT) {
 			for (unsigned int i = 0; i < (sizeof(selgame_Ip) / sizeof(selgame_Ip[0])); i++) {
 				selgame_Ip[i] = (selgame_Ip[i] >= 'A' && selgame_Ip[i] <= 'Z') ? selgame_Ip[i] + 'a' - 'A' : selgame_Ip[i];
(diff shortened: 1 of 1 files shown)


Source/DiabloUI/selgame.cpp
```diff
@@ -620,8 +623,8 @@ void selgame_Password_Select(int /*value*/)
 		} else {
 			strcpy(sgOptions.Network.szPreviousHost, selgame_Ip);
 		}
-		if (SNetJoinGame(selgame_Ip, gamePassword, gdwPlayerId)) {
-			if (!IsGameCompatibleWithErrorMessage(*m_game_data)) {
+		if (allowJoin && SNetJoinGame(selgame_Ip, gamePassword, gdwPlayerId)) {
+			if (!IsGameCompatibleWithErrorMessage(Gamelist[selgame_selectedGame - 3].gameData)) {
```
StephenCWills, line 627: This change only works for public games. We can use `m_game_data` here because it will be populated by `SNetJoinGame()` through a convoluted series of events starting with the `PT_JOIN_ACCEPT` handler and ending with the `EVENT_TYPE_PLAYER_CREATE_GAME` handler.

```suggestion
			if (!IsGameCompatibleWithErrorMessage(*m_game_data)) {
```

Source/DiabloUI/selgame.cpp
```diff
@@ -630,14 +633,22 @@ void selgame_Password_Select(int /*value*/)
 			UiInitList_clear();
 			selgame_endMenu = true;
 		} else {
+
```
StephenCWills, line 636:
```suggestion
```

StephenCWills (changes requested):

PR #5653, opened 2023-01-12
Use outline for spell icon texts

AJenbo:
Looks closer to the original and much simpler. Will be used in floating numbers as well


AJenbo:
image

glebm (approved):
Why are we skipping color index zero? The skipping is usually used for baked in shadows but fonts don't have any?

AJenbo:
I'm not 100% sure since it was @qndel who originally did this, but I guess it's because if it's black it doesn't look tood to draw black around it. But your probably right that there aren't any such cases (i didn't check)

PR #5657, opened 2023-01-12
Bump `cmake_minimum_required` to 3.15

glebm:
DevilutionX already does not build on 3.13 because of libzt.

libzt uses the `TARGET_OBJECTS` generator expression with non-OBJECT libraries. This is only supported in CMake 3.15+:
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/3178


AJenbo:
It's an optional dependency, but sure lets bump things :)

AJenbo (approved):

PR #5659, opened 2023-01-14
A tool to format a struct table

glebm:
Does not support subobjects, i.e. it treats them as a single value.

To format all the struct tables in the codebase, simply run the python file.

Numeric columns are right-aligned, everything else is left-aligned.

@kphoenix137


kphoenix137:
Add "Source/spelldat.cpp" and "Source/misdat.cpp"?

glebm:
Done

AJenbo (approved):
Nice work

PR #5660, opened 2023-01-15
Fix Friendly Fire name and description

kphoenix137:
Friendly Fire game option is backwards. When turning Friendly Fire game option to On, friendly fire in game is prevented, rather than allowed. The description is misleading as well, since it states that it allows friendly fire, even when friendly mode is enabled.

Note: The term "Friendly Fire" may be a bit misleading, which may have been the reason it got implemented this way in the first place, as one might assume it means "Firing becomes friendly, not hitting allies". The actual definition is "weapon fire coming from one's own side, especially fire that causes accidental injury or death to one's own forces."

As the default in Diablo is that friendly fire is allowed to happen, the game option should be "No Friendly Fire" or "Disable Friendly Fire", which can be changed to Yes to prevent friendly fire from happening. If the game options remains as "Friendly Fire", the default option would have to be Yes to match the default Diablo behavior. I believe changing it to No Friendly Fire would be the optimal option as No seems to be the default for most options to match vanilla behavior.

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -271,7 +271,7 @@ bool Plr2PlrMHit(const Player &player, int p, int mindam, int maxdam, int dist, 
 {
 	Player &target = Players[p];
 
-	if (sgGameInitInfo.bFriendlyFire == 0 && player.friendlyMode)
+	if (sgGameInitInfo.bNoFriendlyFire == 0 && player.friendlyMode)
 		return false;
 
 	*blocked = false;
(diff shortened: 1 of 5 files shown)


StephenCWills:
I tested this setting just now, and it does not seem to be backwards. It works the opposite of how you describe. Can you test again and confirm?

AJenbo:
The way it correctly works matches the original and the how friendly fire normally works in games (you take damage from stray projectiles fire by friends). The proposed change would have it switched off by default which i the opposite of how the original worked. Reducing the description also doesn't appear to help explain things.

PR #5666, opened 2023-01-15
Enable Xbox One builds

AJenbo:


fwannmacher:
Weird... Before pushing I have tested it locally and everything has worked.

I'll take a look

glebm:
This could be relevant, from the log:

```
'VsDevCmd.bat' is not recognized as an internal or external command,
operable program or batch file.
'makecert' is not recognized as an internal or external command,
operable program or batch file.
'pvk2pfx' is not recognized as an internal or external command,
operable program or batch file.
```

glebm:
The build was enabled in https://github.com/diasurgical/devilutionX/pull/5669

PR #5672, opened 2023-01-16
Update Discord SDK

glebm:


glebm:
@heinermann Any idea what's up with the build errors?

glebm:
I think we should simply disable DISCORD_INTEGRATION for mingw builds since it's not supported there anyway

heinermann:
What are the build errors?

AJenbo (approved):
Sounds like we should maybe consider not using MinGW for the next release :(

PR #5674, opened 2023-01-17
`enum spell_type` -> `enum class SpellType`

kphoenix137:
Changes enum spell_type to enum class SpellType

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1451,7 +1451,7 @@ size_t OnSpellWall(const TCmd *pCmd, Player &player)
 	if (wParam1 > SPL_LAST)
 		return sizeof(message);
 	const uint16_t wParam2 = SDL_SwapLE16(message.wParam2);
-	if (wParam2 > static_cast<uint8_t>(SpellType::Invalid))
+	if (wParam2 > static_cast<uint8_t>(SpellType::Last))
 		return sizeof(message);
 
 	auto spell = static_cast<spell_id>(wParam1);
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1492,7 +1492,7 @@ size_t OnSpellTile(const TCmd *pCmd, Player &player)
 	if (wParam1 > SPL_LAST)
 		return sizeof(message);
 	const uint16_t wParam2 = SDL_SwapLE16(message.wParam2);
-	if (wParam2 > static_cast<uint8_t>(SpellType::Invalid))
+	if (wParam2 > static_cast<uint8_t>(SpellType::Last))
 		return sizeof(message);
 
 	auto spell = static_cast<spell_id>(wParam1);
@@ -1642,7 +1642,7 @@ size_t OnSpellMonster(const TCmd *pCmd, Player &player)
 	if (wParam2 > SPL_LAST)
 		return sizeof(message);
 	const uint16_t wParam3 = SDL_SwapLE16(message.wParam3);
-	if (wParam3 > static_cast<uint8_t>(SpellType::Invalid))
+	if (wParam3 > static_cast<uint8_t>(SpellType::Last))
 		return sizeof(message);
 
 	auto spell = static_cast<spell_id>(wParam2);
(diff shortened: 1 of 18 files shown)


Source/panels/spell_list.cpp
```diff
@@ -200,22 +200,22 @@ std::vector<SpellListItem> GetSpellListItems()
 	int x = mainPanelPosition.x + 12 + SPLICONLENGTH * SPLROWICONLS;
 	int y = mainPanelPosition.y - 17;
 
-	for (int i = RSPLTYPE_SKILL; i < RSPLTYPE_INVALID; i++) {
+	for (int i = static_cast<uint8_t>(SpellType::Skill); i < static_cast<uint8_t>(SpellType::Invalid); i++) {
```
AJenbo, line 203: we should maybe add a first/last alias

Source/panels/spell_list.cpp
```diff
@@ -324,26 +324,26 @@ void DoSpeedBook()
...
-			case RSPLTYPE_CHARGES:
+			case SpellType::Charges:
 				spells = myPlayer._pISpells;
 				break;
 			}
 			uint64_t spell = 1;
 			for (int j = 1; j < MAX_SPELLS; j++) {
 				if ((spell & spells) != 0) {
-					if (j == myPlayer._pRSpell && i == myPlayer._pRSplType) {
+					if (j == myPlayer._pRSpell && i == static_cast<uint8_t>(myPlayer._pRSplType)) {
```
AJenbo, line 346: nit: i wold cast i to the enum to have a more consistant and strict operation

Source/msg.cpp
```diff
@@ -1451,7 +1451,7 @@ size_t OnSpellWall(const TCmd *pCmd, Player &player)
 	if (wParam1 > SPL_LAST)
 		return sizeof(message);
 	const uint16_t wParam2 = SDL_SwapLE16(message.wParam2);
-	if (wParam2 > RSPLTYPE_INVALID)
+	if (wParam2 > static_cast<uint8_t>(SpellType::Invalid))
```
AJenbo, line 1454: if you add last this wold be a good place to use it also

Source/panels/spell_list.cpp
ephphatha: > we should maybe add a first/last alias

That'd be good, enum_traits.h defines a couple of template functions that could be used to iterate over enum values if the aliases use `FIRST`/`LAST` as the identifiers.

Source/msg.cpp
```diff
@@ -1642,7 +1642,7 @@ size_t OnSpellMonster(const TCmd *pCmd, Player &player)
 	if (wParam2 > SPL_LAST)
 		return sizeof(message);
 	const uint16_t wParam3 = SDL_SwapLE16(message.wParam3);
-	if (wParam3 > RSPLTYPE_INVALID)
+	if (wParam3 > static_cast<uint8_t>(SpellType::Invalid))
```
AJenbo, line 1645:
```suggestion
	if (wParam3 > static_cast<uint8_t>(SpellType::Last))
```

Source/msg.cpp
```diff
@@ -1681,7 +1681,7 @@ size_t OnSpellPlayer(const TCmd *pCmd, Player &player)
...
```
AJenbo, line 1684:
```suggestion
	if (wParam3 > static_cast<uint8_t>(SpellType::Last))
```

AJenbo:
Also check if enum_traits can help with iterating over the enum

kphoenix137:
> Also check if enum_traits can help with iterating over the enum

I need help with this

ephphatha:
You need to declared the identifiers as `FIRST`/`LAST` (all caps) as that is what enum_traits expects. You can then use code like `for (auto i : enum_values<SpellType>()) { /* do something with i, which is a SpellType value */ }` to iterate over the values in that range.

The enum values iterator returns all the values from FIRST to LAST inclusive, so you need to define `LAST = Charges` instead of `Invalid` to keep the same behaviour.

kphoenix137:
> for (auto i : enum_values<SpellType>())

Are you sure `LAST = Charges` instead of `LAST = Invalid`? Invalid is checked in the switch case inside the for loop

ephphatha:
Not at my computer to check the code but I saw 4 loops over spell type values, all of which were either `<= Charges` or `< Invalid`. For example the `case SpellType::Invalid` section in `GetSpeedListItem` is unreachable. Another example is the loop in `DoSpeedBook`, it operates over the same range and doesn't have a case for `SpellType::Invalid`.

AJenbo (approved):

DakkJaniels:
Game is crashing when you try and open speedbook as a result of this PR.

PR #5682, opened 2023-01-18
`UNPACKED_SAVES`: Saves without MPQs

glebm:
Reduces rg99 binary size by ~70 KiB.

The unpacked saves are ~30% larger than the MPQ saves (though does not it matter for minimizing binary footprint):

```bash
$ du -sh single_0_hsv single_0.hsv
3.7M	single_0_hsv
2.8M	single_0.hsv
```

Existing saves can be unpacked using https://github.com/diasurgical/devilutionx-mpq-tools/


Source/pfile.h
```diff
@@ -8,12 +8,76 @@
...
+	{
+		::devilution::RenameFile((dir_ + from).c_str(), (dir_ + to).c_str());
+	}
+
+	void RemoveHashEntry(const char *path)
+	{
+		RemoveFile((dir_ + path).c_str());
+	}
+
+	void RemoveHashEntries(bool (*fnGetName)(uint8_t, char *));
```
AJenbo, line 70: Ideally the method would be called RemoveFile(s) and for the MpqWriter it does it by removing hash entires.

AJenbo (approved):

PR #5696, opened 2023-01-21
Add item graphics to store menus

### LINKED ISSUE #1359: Draw item graphics on vendor screens
## Description

Standard vendor pages are very barebones. We can improve on them without changing their logic too dramatically by rendering the item icons alongside the item descriptions.

## Details
- Draw the icons to the left of the item description
- Investigate potential impact on item template height

Behavior should be standard as there is zero gameplay impact. No need for feature flags.

StephenCWills:
Adds item graphics to the store menu, based on mockups posted in #1359.

image

Graphics are downscaled by half to fit in the available space to the left of the item description. The `BilinearDownscaleByHalf8()` function was provided courtesy of @glebm. I had to modify it a bit to handle transparency.

This resolves #1359.

### DIFF
--- a/Source/utils/sdl_bilinear_scale.cpp
+++ b/Source/utils/sdl_bilinear_scale.cpp
@@ -148,4 +148,37 @@ void BilinearScale32(SDL_Surface *src, SDL_Surface *dst)
 	}
 }
 
+void BilinearDownscaleByHalf8(SDL_Surface *src, const uint8_t (*paletteBlendingTable)[256], SDL_Surface *dst, uint8_t transparentIndex)
+{
+	for (unsigned y = 0; y < static_cast<unsigned>(dst->h); ++y) {
+		auto *srcPixels = static_cast<uint8_t *>(src->pixels) + 2 * y * src->pitch;
+		auto *dstPixels = static_cast<uint8_t *>(dst->pixels) + y * dst->pitch;
+		for (unsigned x = 0; x < static_cast<unsigned>(dst->w); ++x) {
+			uint8_t quad[] = {
+				srcPixels[0],
+				srcPixels[1],
+				srcPixels[src->pitch],
+				srcPixels[src->pitch + 1]
+			};
+			// Attempt to avoid blending with transparent pixels
+			if (quad[0] == transparentIndex)
+				quad[0] = quad[1];
+			if (quad[1] == transparentIndex)
+				quad[1] = quad[0];
+			if (quad[2] == transparentIndex)
+				quad[2] = quad[3];
+			if (quad[3] == transparentIndex)
+				quad[3] = quad[2];
+			uint8_t top = paletteBlendingTable[quad[0]][quad[1]];
+			uint8_t bottom = paletteBlendingTable[quad[2]][quad[3]];
+			if (top == transparentIndex)
+				top = bottom;
+			if (bottom == transparentIndex)
+				bottom = top;
+			*dstPixels++ = paletteBlendingTable[top][bottom];
+			srcPixels += 2;
+		}
+	}
+}
+
 } // namespace devilution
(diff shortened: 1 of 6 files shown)


ikonomov:
If possible might be a good idea to have the red overlay for the graphics of the items the character can't use for consistency.  I'm absolutely overjoyed that you guys were able to make this work.  Looks fantastic!

Source/utils/sdl_bilinear_scale.cpp
```diff
@@ -148,4 +148,37 @@ void BilinearScale32(SDL_Surface *src, SDL_Surface *dst)
...
+				quad[2] = quad[3];
+			if (quad[3] == transparentIndex)
+				quad[3] = quad[2];
+			uint8_t top = paletteBlendingTable[quad[0]][quad[1]];
+			uint8_t bottom = paletteBlendingTable[quad[2]][quad[3]];
+			if (top == transparentIndex)
+				top = bottom;
+			if (bottom == transparentIndex)
+				bottom = top;
+			*dstPixels++ = paletteBlendingTable[top][bottom];
```
AJenbo, line 178: I think it might be better to test for 2 or more transparent pixels and in that case mark it as fully transparent. This is how we handle transparency in pngs2cel and d1gt.

This also lets us make the assumption in the second part that at most 1px will be transparent. My suggestion here might be a bit over optimized and could be done shorter, but I couldn't decide of which of 3 implementations to got with so went with what I think is the fastest one.

```suggestion
			if ((quad[0] == transparentIndex && quad[1] == transparentIndex)
				|| (quad[0] == transparentIndex && quad[2] == transparentIndex)
				|| (quad[0] == transparentIndex && quad[3] == transparentIndex)
				|| (quad[1] == transparentIndex && quad[2] == transparentIndex)
			) {
				*dstPixels++ = transparentIndex;
				continue;
			}

			// Avoid blending with transparent pixels
			uint8_t top;
			uint8_t bottom;
			if (quad[0] == transparentIndex)
				top = quad[1];
				bottom = paletteBlendingTable[quad[2]][quad[3]];
			else if (quad[1] == transparentIndex)
				top = quad[0];
				bottom = paletteBlendingTable[quad[2]][quad[3]];
			else if (quad[2] == transparentIndex)
				top = paletteBlendingTable[quad[0]][quad[1]];
				bottom = quad[3];
			else if (quad[3] == transparentIndex)
				top = paletteBlendingTable[quad[0]][quad[1]];
				bottom = quad[2];
			else
				top = paletteBlendingTable[quad[0]][quad[1]];
				bottom = paletteBlendingTable[quad[2]][quad[3]];

			*dstPixels++ = paletteBlendingTable[top][bottom];
```

AJenbo: i guess yours have the benifit of keeping thin details like bow strings

ikonomov: 213886403-5b4300f6-d1c0-405c-b104-7d8971716204
A sample of how it might look with red overlay for items that are not usable by the character.

AJenbo (approved):

Source/utils/sdl_bilinear_scale.cpp
AJenbo: @ikonomov They start your own thread, this one is about how to sample the transparent pixels :D

StephenCWills: > i guess yours have the benifit of keeping thin details like bow strings, but it appears to be biased against horizontal transparancy

I think there's no horizontal bias. If there are any non-transparent pixels, my implementation would blend those pixels and ignores the transparent ones. If anything, it has a bias against transparency altogether. Your idea of applying transparency if two or more source pixels are transparent would be a good approach to correcting that bias.

TBH, I settled on this implementation after looking at the bow strings.

AJenbo: The way I read your it will ignore any single transparent pixel in the top or bottom, but if top or bottom is fully transparent the result will also be.

But if there is 1 transparent and 1 solid in the top, and 1 transparent and 1 solid in the bottom, both gets ignored despite there overall being 50% transparency.

So transparency of 50% horizontally will result in transparent pixels, but diagonal or vertical transparency of 50% will result in a colored pixel.

image

AJenbo: No this is just me misreading line 174->178. You actually having everything turn in to a solid pixel if they are 75% transparent.

ikonomov:
Just attaching the image from the discussion of the alignment change.
up

PR #5699, opened 2023-01-22
Recalculate cursor position after level change

### LINKED ISSUE #5571: Staircase Teleport Bug
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

On any character, if you go down the stairs to the next level and rapidly use teleport, it will behave largely like phasing instead and catapult the character to a random location on the current level. This is an old trick that was used primarily on Normal/Hell by Sorcerers to stumble upon the next level ultra fast.

### To Reproduce

1- Go to Hell with any character
2- Loading screen appears
3-Spam or Use Teleport, see how far you lend from initial stairs

### Expected Behavior

The bug should just be fixed. This was clearly unintended and teleport should behave normally upon entering a new level.

### Additional context

_No response_

obligaron:
Fixes #5571

How does the bug happen:
1. Player initializes level transition
2. Player clicks right mouse button
3. New level is loaded. First game loop is processed.
4. `HandleMessage` is called. It sees the right mouse click sdl event and adds `CMD_SPELLXY` Message. Here are the old coordinates used (`cursPosition`). The coordinates get updated in `ProcessInput` that is called after `HandleMessage`.

With this pr `CheckCursMove` is called after every level change/load and before the normal game loop runs.
This ensures that that all cursor related information are initialized correctly.
`ResetItemlabelHighlighted` and `sgbMouseDown` must be called so that `CheckCursMove` always calculates a new value (even if previous a item was selected or the player locked on a target).

Alternative fixes:
- Discard all messages after level change (`HandleMessage` don't add new commands).
- Call `ProcessInput`/`CheckCursMove` before `HandleMessage`.

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -747,8 +747,6 @@ void GameEventHandler(const SDL_Event &event, uint16_t modState)
 			nthread_ignore_mutex(true);
 			PaletteFadeOut(8);
 			sound_stop();
-			LastMouseButtonAction = MouseActionType::None;
-			sgbMouseDown = CLICK_NONE;
 			ShowProgress(GetCustomEvent(event.type));
 
 			RedrawEverything();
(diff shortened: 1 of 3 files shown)


AJenbo:
Nice work, you really know your way around the engine at this point :D

AJenbo (approved):

PR #5700, opened 2023-01-22
Add option for Floating Numbers to go vertical only

Trihedraf:


AJenbo:
I think I would like to gather some feedback to see if this should just be the way it works rather then having an option. Still keep this open in case that is what we end up with.

kphoenix137:
I think both are good. Would be nice to have both options.

PR #5701, opened 2023-01-22
Store graphics: Scale sprites only once

glebm:
Prepare downscaled sprites once instead of doing it on every frame.

Note that we do this lazily in `StartStore` rather than `SetupTownStores` because we need the palette blending table to be available when downscaling.


glebm:
Actually let's move this to cursor.cpp

AJenbo (dismissed):

glebm:
Moved

AJenbo (approved):

PR #5703, opened 2023-01-22
Store items: Draw unusable items in red

glebm:
image

We generate separate downscaled sprites for the red versions because simply applying TRN after downscaling leads to some non-red pixels (because of blending during downscaling).


ikonomov:
I think the red overlay helps to make the addition of the thumbnails to the store lists better integrated with the rest of the game.

AJenbo (approved):

PR #5704, opened 2023-01-22
Support mouse back button in stores/dialogues

glebm:


obligaron (approved):

AJenbo:
Nice hopefully this will also help on Android... or at least we should do the same in some sens there.

PR #5705, opened 2023-01-22
Do not use PKWare with no MPQs + NONET

glebm:
Reduces RG99 binary size by 12 KiB


AJenbo (approved):
Wish we could avoid it in general :)

PR #5706, opened 2023-01-22
`CreateAulibResampler`: Check the real sample rate

glebm:
The requested sample rate is merely preferred and the actual rate may be different.


AJenbo (approved):

AJenbo:
Arh that makes good sens

PR #5707, opened 2023-01-22
`StaticVector`: Replace `std::aligned_storage_t`

glebm:
`std::aligned_*` types are deprecated in C++23 per https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1413r3.pdf


AJenbo (approved):
even looks cleaner

Issue #5709, opened 2023-01-23
Severe: Game crash


kphoenix137:
### Operating System

Windows x64

### DevilutionX version

Other (please specify version number)

### Describe

Commit 1401e130

Game crashes with assertion failure.

### To Reproduce

1. Toggle on "Show Graphics in Stores"
2. Create multiplayer game
3. Talk to Griswold

Trihedraf:
Also happening for me. Debug minGW x86.

AJenbo:
What is the assertion?

AJenbo:
```

ERROR: Error
assertion failed (C:\Users\Eric\Documents\GitHub\devilutionX\Source\engine/clx_sprite.hpp:616)
value_.data_ != nullptr
The thread 0x3758 has exited with code 0 (0x0).
avcore\audiocore\client\audioclient\audioclientcore.cpp(1907)\AUDIOSES.DLL!00007FF9D77B3EF5: (caller: 00007FF993683F12) ReturnHr(1) tid(47c8) 80070057 The parameter is incorrect.
```

PR #5711, opened 2023-01-23
Fix latest crash with Shop Item Graphics

kphoenix137:
Fixes: #5709


glebm:
Let's merge `#5712` instead because it avoids going through the branch in `GetInvItemSprite`.

PR #5719, opened 2023-01-23
Filter bad words from chat Toggle

kphoenix137:
Implements a toggle in the Settings menu to filter bad words displayed in chat. Currently only contains English words.

Toggle is off by default.


glebm:
To avoid triggering adult content filters elsewhere, it is usually common practice to apply rot13 or another obfuscation method to the swear words in source code

glebm:
For example, using something like https://github.com/adamyaxley/Obfuscate

glebm:
Nah that library is not good, it only obfuscates strings in the binary but not in the source code. Just rot13 it I guess.

AJenbo:
What I did for the bot is punt the list to a file that the user is responsible for populating:
https://github.com/diasurgical/devilutionx-gamelist/blob/main/discord_bot.py#L180

That way to enable it they have fille it with what they consider unacceptable rather then us being helt to some standard.

PR #5722, opened 2023-01-25
Reduce `ObjectData` size: 24 bytes -> 11 bytes

glebm:
1. Reduces sizes of integral fields.
2. Merges boolean flags into a single field.


AJenbo (approved):

AJenbo:
i kinda like it better this way as well

glebm:
Yeah, I'll do the same for other tables after the enum renaming PRs are merged.

PR #5728, opened 2023-01-28
Fix Guardian damage

ikonomov:
The intended damage for the Guardian spell in the AddGuardian function is never used because it uses the function MI_Firebolt to create the missiles which have their own damage formula.


AJenbo:
@ikonomov do you consider this issue solved with https://github.com/diasurgical/devilutionX/pull/5731 merged?

ikonomov:
Yes, thank you! :)

PR #5730, opened 2023-01-29
fix speedbook crash

DakkJaniels:
Fixes speed book crash introduced under #5674

### DIFF
--- a/Source/panels/spell_list.cpp
+++ b/Source/panels/spell_list.cpp
@@ -216,7 +216,7 @@ std::vector<SpellListItem> GetSpellListItems()
 			mask = myPlayer._pISpells;
 			break;
 		default:
-			break;
+			continue;
 		}
 		int8_t j = SPL_FIREBOLT;
 		for (uint64_t spl = 1; j < MAX_SPELLS; spl <<= 1, j++) {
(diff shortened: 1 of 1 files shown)


Source/panels/spell_list.cpp
```diff
@@ -339,6 +339,8 @@ void DoSpeedBook()
 			case SpellType::Charges:
 				spells = myPlayer._pISpells;
 				break;
+			default:
+				spells = 0;
```
StephenCWills, line 343:
```suggestion
			default:
				spells = 0;
				break;
```

StephenCWills:
This does fix the crash, but the speedbook does still have a problem.

image

StephenCWills (dismissed):

StephenCWills (dismissed):

AJenbo:
With https://github.com/diasurgical/devilutionX/pull/5732 merged should we close this?

StephenCWills:
With that one merged, this one is unnecessary. However, it's not a bad thing to have valid default cases for these switch statements.

StephenCWills:
I screwed up a merge and ended up squashing to fix it.

StephenCWills (approved):

PR #5731, opened 2023-01-29
Rename MI_Firebolt to MI_Projectile and calculate damage when missile added

DakkJaniels:
As a result of this change, guardian damage now uses the damage calculated by AddGuardian. 

I also included using the guardian spell level for the firebolt damage, rather than the caster's firebolt level, but this can be switched back if preferred.


ikonomov:
I know this is a fantastic improvement to the logic of the code, I wish it somehow magically makes missiles.cpp easier for me to understand, but it seems magics only work inside the game and not here on GitHub. :D  Dakk, fantastic work!  Thank you so much for making this change.

As discussed, once (and if) it gets merged, we should probably consider moving the random damage calculation for each missile from Guardian.

AJenbo (approved):

PR #5733, opened 2023-01-29
Refactor Missile Function Names

kphoenix137:


AJenbo:
i was ok with the Add* names, and leaving them means we will have to do less translation when looking at old versions etc. do you think the name are better this way?

Other then that i like the changes to the MI_* names. Could you rebase?

AJenbo (approved):

PR #5736, opened 2023-01-30
Derive viewport geometry from player position

### LINKED ISSUE #1498: Garbage graphics at screen top when zoomed at 1280x960
**Describe**
Broken graphics near top border

**To Reproduce**
Throw a lot of gold or other stuff in town, walk up/down, watch top border

paddinggobrrr

StephenCWills:
Basically, issue #1498 occurs because the logic in `CalcViewportGeometry()` selects the wrong tile from which to start rendering. Given that `CalcTileOffset()` was effectively producing the wrong offset, and that the logic for this function was already attempting to make a few adjustments to fix it, I decided to take a more straightforward approach and just compute all the values from scratch. I'm hoping this change should make the code in this function easier to understand and less error prone.

This resolves #1498

### DIFF
--- a/Source/engine/render/scrollrt.cpp
+++ b/Source/engine/render/scrollrt.cpp
@@ -1467,46 +1467,49 @@ void TilesInView(int *rcolumns, int *rrows)
 
 void CalcViewportGeometry()
 {
+	const int zoomFactor = *sgOptions.Graphics.zoom ? 2 : 1;
+	const int screenWidth = GetScreenWidth() / zoomFactor;
+	const int screenHeight = GetScreenHeight() / zoomFactor;
+	const int panelHeight = GetMainPanel().size.height / zoomFactor;
+	const int pixelsToPanel = screenHeight - panelHeight;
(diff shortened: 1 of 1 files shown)


Source/engine/render/scrollrt.cpp
```diff
@@ -1467,46 +1467,49 @@ void TilesInView(int *rcolumns, int *rrows)
 
 void CalcViewportGeometry()
 {
-	tileShift = { 0, 0 };
+	int zoomFactor = *sgOptions.Graphics.zoom ? 2 : 1;
+	int screenWidth = GetScreenWidth() / zoomFactor;
+	int screenHeight = GetScreenHeight() / zoomFactor;
+	int panelHeight = GetMainPanel().size.height / zoomFactor;
+	int pixelsToPanel = screenHeight - panelHeight;
+	Point playerPosition { screenWidth / 2, pixelsToPanel / 2 };
```
glebm, line 1475: Please mark constant variables as `const`

PR #5737, opened 2023-01-30
Optimize AddPlrExperience()

kphoenix137:
`AddPlrExperience()` is called whenever the player performs an action that would give them more experience points. This is unncessary when the player is at the hardcoded maximum already. `return` added near beginning of function, as well as some magic numbers replaced.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2716,7 +2716,7 @@ void AddPlrExperience(Player &player, int lvl, int exp)
 		RedrawEverything();
 	}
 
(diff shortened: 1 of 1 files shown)


DakkJaniels:
As part of this, should we just remove the extra line from the experience table? It has no effect on gameplay, and makes it a bit confusing if you are modifying the experience table as part of a mod. Why not just have the number of entries equal to the max character level?

ikonomov:
I really hope as part of this update we can include the PR here https://github.com/diasurgical/devilutionX/pull/3928

AJenbo:
Would be a lot easier to do if @kphoenix137 can reopen it

PR #5744, opened 2023-02-03
Change default monster missile damage to zero

StephenCWills:
Recently saw a recording on Discord where a Barbarian with 50% resist in Hell/Hell was consistently receiving exactly 2 damage from every Blood Star missile. Testing reveals this was a bug, introduced by #5731. That PR interprets non-zero damage as a custom amount provided to the missile, like in the case of the Guardian spell. This PR uses zero to indicate that there is no custom damage amount for missiles produced by monsters, apart from Diablo's Apocalypse attack.

I checked all the missile types returned by `GetMissileType(MonsterAIID)`, and I wasn't able to determine why 4 was being used as the default missile damage for monsters. Maybe the original developers used it for testing at some point during development. 🤷

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1905,9 +1905,9 @@ void AiRanged(Monster &monster)
 			if (LineClearMissile(monster.position.tile, monster.enemyPosition)) {
 				MissileID missileType = GetMissileType(monster.ai);
 				if (monster.ai == MonsterAIID::AcidUnique)
-					StartRangedSpecialAttack(monster, missileType, 4);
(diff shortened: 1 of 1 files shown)


galaxyhaxz:
What a coincidence! I too was looking at this exact thing last night. The damage values here are not used, Flash for mages also uses 4. It also uses the melee damage values for some types like Inferno, instead of having a custom amount.

DakkJaniels:
Whoops, I had checked a few and thought they were listed as 0, I guess I missed some. Good catch!

AJenbo (approved):

PR #5746, opened 2023-02-03
Clean up missile data access

glebm:
Replace enum static casts with `GetMissile(Sprite)Data`.


Source/missiles.cpp
```diff
@@ -3440,7 +3440,7 @@ void ProcessGuardian(Missile &missile)
 	if (missile.var2 > 0) {
 		missile.var2--;
 	}
-	if (missile._mirange == missile.var1 || (missile._mimfnum == static_cast<uint8_t>(MissileGraphicID::Guardian) && missile.var2 == 0)) {
+	if (missile._mirange == missile.var1 || (missile._mimfnum == 2 && missile.var2 == 0)) {
```
AJenbo, line 3443: I would see it as a benefit that we don't just have a magic number?

glebm: `_mifnum` is not a `MissileGraphicID`. I'm not sure whether it's a `Direction` either but it's certainly not a graphic ID.
Bette a magic number than the wrong constant that just happens to have the same value.

AJenbo (approved):

Source/missiles.cpp
kphoenix137: > `_mifnum` is not a `MissileGraphicID`. I'm not sure whether it's a `Direction` either but it's certainly not a graphic ID. Better a magic number than the wrong constant that just happens to have the same value.

I noticed this as well while doing the MissileGraphicID PR, didn't seem like the right thing was being used, although it didn't seem appropriate for the scope of the PR to attempt to fix it.

PR #5747, opened 2023-02-03
Optimize `MissileFileData` size: 80 bytes -> 32 

glebm:


Source/misdat.cpp
```diff
@@ -128,113 +128,152 @@ MissileData MissilesData[] = {
...
+	Repeat(14),
+	Repeat(15),
+	Repeat(16),
+	Repeat(17),
+	Repeat(19),
+	Repeat(20),
+	{ 9, 4 },
+	{ 15, 4, 3 },
+	{ 13, 11 },
+	{ 16, 16, 16, 16, 16, 16, 16, 16, 8 }
```
AJenbo, line 167: it should also be possible to split the sprites for these missiles to avoid the special cases

PR #5748, opened 2023-02-03
Reenable tests on MSVC build

StephenCWills:
It seems the tests aren't running in the MSVC CI jobs since we made the switch to `RelWithDebInfo`.

https://github.com/diasurgical/devilutionX/blob/25e9b7320d7a265dfec1cc9de81d55bb3dd25716/CMakeLists.txt#L161-L166

### DIFF
--- a/Packaging/windows/CMakePresets.json
+++ b/Packaging/windows/CMakePresets.json
@@ -21,7 +21,7 @@
 		},
 		{
 			"name": "ninja-vcpkg-debug",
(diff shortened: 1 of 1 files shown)


glebm (approved):

glebm:
To make it work with LTO we'd need to add `DVL_API_FOR_TEST` to all symbols (functions) that are used in tests, and then remove this line:
https://github.com/diasurgical/devilutionX/blob/7e11d63004a10b899506b9ac0663208a504465e1/test/CMakeLists.txt#L7

PR #5749, opened 2023-02-04
Provide default selections for resolution on systems with only one display mode

### LINKED ISSUE #5723: Provide sensible resolution when upscaling is enabled
For Android and iOS only the max resolution is suggested isn't the OS doesn't provide a method for changing the screen resolution. This means that users only have the option between 640x480 and 4k on some devices, not very useful.

So in case the OS only reports a single resolution it would be nice if we can check if up-scaling is enabled and in that case generate some sensible alternatives below that size.

This can also help to mitigate that the UI will sometimes overlap on devices that aren't 21:9

StephenCWills:
This PR provides the following options for my 2280x1080 Pixel 4 device.

Screenshot_20230203-221557

This resolves #5723

### DIFF
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -293,7 +293,7 @@ void OptionEnemyHealthBarChanged()
 }
 
 #if !defined(USE_SDL1) || defined(__3DS__)
-void OptionFitToScreenChanged()
+void ResizeWindowAndUpdateResolutionOptions()
 {
 	ResizeWindow();
 #ifndef __3DS__
@@ -705,10 +705,13 @@ void OptionEntryResolution::CheckResolutionsAreInitialized() const
 	float scaleFactor = GetDpiScalingFactor();
 
 	// Add resolutions
+	bool supportsAnyResolution = false;
 #ifdef USE_SDL1
 	auto *modes = SDL_ListModes(nullptr, SDL_FULLSCREEN | SDL_HWPALETTE);
 	// SDL_ListModes returns -1 if any resolution is allowed (for example returned on 3DS)
-	if (modes != nullptr && modes != (SDL_Rect **)-1) {
+	if (modes == (SDL_Rect **)-1) {
+		supportsAnyResolution = true;
+	} else if (modes != nullptr) {
 		for (size_t i = 0; modes[i] != nullptr; i++) {
 			if (modes[i]->w < modes[i]->h) {
 				std::swap(modes[i]->w, modes[i]->h);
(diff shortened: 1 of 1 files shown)


StephenCWills:
@glebm Do you know of any SDL1 devices where providing additional options like this might cause a problem?

AJenbo:
Yeah SDL1 doesn't have upscaling support, there is no good reason to provide it with options that have to be scaled.

Source/options.cpp
```diff
@@ -719,9 +719,25 @@ void OptionEntryResolution::CheckResolutionsAreInitialized() const
...
+		// Attempt to provide sensible options for 4:3 and the native aspect ratio
+		const int width = sizes[0].width;
+		const int height = sizes[0].height;
+		const int commonHeights[] = { 480, 540, 720, 960, 1080, 1440, 2160 };
+		for (int commonHeight : commonHeights) {
+			if (commonHeight > height)
+				break;
+			sizes.emplace_back(Size { commonHeight * 4 / 3, commonHeight });
+			if (commonHeight * width % height == 0)
+				sizes.emplace_back(Size { commonHeight * width / height, commonHeight });
```
AJenbo, line 732: Should we maybe just provide them as 4/3 and let fit-to-screen deal with this case instead of having them in there explicitly? It will appear as if there is no difference in picking half of them to some users.

StephenCWills: I had considered that. I thought it might feel a bit odd to have only 4:3 resolutions to choose from on a widescreen device. Some resolutions won't divide evenly at the native resolution so listing the ones that do could potentially be helpful. Having them listed should give the user more freedom when experimenting with the `Fit to Screen` option, hopefully helping them discover how it works.

On the flip side, I also found it strange to see 1440x1080 in the list. Maybe it makes some sense to list it given that the game was originally 4:3, but players will almost certainly be more familiar with the 16:9 resolution at that height. It might be better to limit 4:3 resolutions up to 960p and then add some 16:9 variants starting from 720p. If we do that, we could maybe do without the native resolutions.

Ultimately I decided that it wouldn't hurt to include the native resolutions so I added them to this PR, but we can certainly take them out if you prefer. Thoughts?

AJenbo: Maybe as a separate adjustment we could have it only show the unique heights with a p added when fit-to-screen is selected. The underlysning would then be the one with the most narrow width since that then hows an immediate difference when disabling fit-to-screen

StephenCWills:
I didn't change much. Mainly, I added a `supportsAnyResolution` flag that is true on SDL2 systems when upscale is true, and true on SDL1 systems if `SDL_ListModes()` returns -1. It still attempts to add resolutions at the native aspect ratio, which I figure is now okay since the options will look different when Fit to Screen is on.

AJenbo (approved):

PR #5752, opened 2023-02-05
Fix wrong flip sound: Tavern Sign

kphoenix137:
Tavern Sign was set to use the inventory placement sound, rather than the flip sound.

What it sounds like with the fix:

https://user-images.githubusercontent.com/68359262/216803557-320a096b-4676-4c1a-b700-1a5addcbd4ad.mp4


AJenbo (approved):

ikonomov:
[flipsign.zip](https://github.com/diasurgical/devilutionX/files/10700865/flipsign.zip)

PR #5753, opened 2023-02-05
add town automap toggle

ikonomov:
A toggle intended to hide the map in town when automap is enabled.

### DIFF
--- a/Source/automap.cpp
+++ b/Source/automap.cpp
@@ -13,6 +13,7 @@
 #include "engine/render/automap_render.hpp"
 #include "levels/gendung.h"
 #include "levels/setmaps.h"
+#include "options.h"
 #include "player.h"
 #include "utils/language.h"
 #include "utils/stdcompat/algorithm.hpp"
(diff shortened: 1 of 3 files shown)


ikonomov:
I've left the original "No automap available in town" message and a check to disable it in SP as it seems unnecessary and possibly even preferable to have it the way it is in MP.

Source/automap.cpp
```diff
@@ -775,7 +776,10 @@ std::unique_ptr<AutomapTile[]> LoadAutomapData(size_t &tileCount)
 {
 	switch (leveltype) {
 	case DTYPE_TOWN:
-		return LoadFileInMem<AutomapTile>("levels\\towndata\\automap.amp", &tileCount);
+		if (*sgOptions.Graphics.showTownMap)
+			return LoadFileInMem<AutomapTile>("levels\\towndata\\automap.amp", &tileCount);
+		else
+			return nullptr;
```
AJenbo, line 782: I don't think there is a need for this as it won't have an effect when the map isn't being rendered any way.
```suggestion
		return LoadFileInMem<AutomapTile>("levels\\towndata\\automap.amp", &tileCount);
```

Source/automap.cpp
```diff
@@ -869,6 +873,10 @@ void AutomapZoomOut()
 
 void DrawAutomap(const Surface &out)
 {
+	if (!*sgOptions.Graphics.showTownMap && leveltype == DTYPE_TOWN) {
+		DrawAutomapText(out);
+		return;
+	}
```
AJenbo, line 879: Maybe we could instead split this function up in one that draws the text and one that draws the map and then have the one that draws the map do an early return if disabled. That way we don't duplicate code and make it harder to follow.

PR #5754, opened 2023-02-05
Add `GetSpellData(SpellID)`

glebm:


Source/spelldat.h
```diff
@@ -223,6 +223,11 @@ struct SpellData {
 	uint16_t sStaffCost;
 };
 
-extern const SpellData spelldata[];
+extern const SpellData SpellsData[];
+
+inline const SpellData &GetSpellData(SpellID spellId)
+{
+	return SpellsData[static_cast<std::underlying_type<SpellID>::type>(spellId)];
```
ephphatha, line 230: `static_cast<size_t>` seems more appropriate given the use, also do you think there's a need to `assert(IsValidSpell(spellId))` here?

glebm: Perhaps surprisingly, the binary comes out smaller if we always cast to the underlying type rather than `size_t`.
We can add a debug assert here and in other similar functions (`GetMissileData`, `GetMissileSpriteData`) in a follow up PR (there was none previously).

ephphatha: Really surprised it makes a difference to the binary size :D

ephphatha (approved):

PR #5755, opened 2023-02-05
Optimize `SpellData` size: 40 bytes -> 24

glebm:


Source/spelldat.h
```diff
@@ -203,24 +205,56 @@ enum class MissileID : int8_t {
...
+	}
+
+	[[nodiscard]] uint32_t bookCost() const
+	{
+		return bookCost100 * 100;
+	}
+
+	[[nodiscard]] uint16_t staffCost() const
+	{
+		return staffCost10 * 10;
```
AJenbo, line 246: this feels very agressive

DakkJaniels: I think it makes sense. Why use a 32-bit number so you can store a bunch of zeroes? As long as it's not changing the cost of things from what they were originally, I don't see the issue.

AJenbo: imo it's harder to read and modefy

glebm: True but the struct is exactly 24 bytes. Adding even a single byte to this struct will increase its size to 32 bytes (on 64-bit systems) due to alignment requirements for the `sNameText` pointer.

glebm: We could change `bookCost100` to `bookCost10` for free though.

Source/items.cpp
```diff
@@ -638,15 +638,21 @@ void GetBookSpell(Item &item, int lvl)
...
-	else if (GetSpellData(bs).sType == MagicType::Lightning)
+		break;
+	case MagicType::Lightning:
 		item._iCurs = ICURS_BOOK_BLUE;
-	else if (GetSpellData(bs).sType == MagicType::Magic)
+		break;
+	case MagicType::Magic:
 		item._iCurs = ICURS_BOOK_GREY;
+		break;
+	}
```
ephphatha, line 655: what about something like
```cpp
enum class SpellDataFlags : uint8_t {
	Fire = 1U << 0,
	Lightning = 1U << 1,
	Targeted = 1U << 2,
	AllowedInTown = 1U << 3,
};
use_enum_as_flags(SpellDataFlags);

struct SpellData {
  [[nodiscard]] bool hasFireAspect() const
  {
    return HasAnyOf(flags, SpellDataFlags::Fire);
  }
...
};

if (spellData.hasFireAspect()) {
  item._iCurs = ICURS_BOOK_RED;
} else if (spellData.hasLightningAspect()) {
  item._iCurs = ICURS_BOOK_BLUE;
} else {
  // non-elemental spell
  item._iCurs = ICURS_BOOK_GREY;
}
```

same can be applied to the `GetPlayerGraphic` function and then the `MagicType` enum can be dropped, leaving two flags in `SpellDataType` for fire/lightning elemental effects.

`MagicType::Magic` is the most common default case, I don't think it'll cause any problem having all spells without an elemental type can fall back to that. The value isn't persisted anywhere that I can see so changing it should be fine.

Source/spelldat.h
AJenbo: Are we memory strapped on 64bit systems?

Source/items.cpp
glebm: I prefer the enum over a series of if statements because `switch` exhaustiveness is checked by the compiler

Source/spelldat.h
glebm: Alignment requirements are still there on 32-bit systems.
On a 32-bit system, the struct is currently exactly 20 bytes; adding even a single a byte will increase the size to 24 bytes.

Source/spells.cpp
```diff
@@ -245,8 +245,9 @@ void CastSpell(int id, SpellID spl, int sx, int sy, int dx, int dy, int spllvl)
 	}
 
 	bool fizzled = false;
-	for (int i = 0; i < 3 && GetSpellData(spl).sMissiles[i] != MissileID::Null; i++) {
-		Missile *missile = AddMissile({ sx, sy }, { dx, dy }, dir, GetSpellData(spl).sMissiles[i], TARGET_MONSTERS, id, 0, spllvl);
+	const SpellData &spellData = GetSpellData(spl);
+	for (size_t i = 0; i < 2 && spellData.sMissiles[i] != MissileID::Null; i++) {
```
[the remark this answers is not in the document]
glebm: Done

AJenbo (approved):

PR #5761, opened 2023-02-05
Debug: Add `takexp` command

kphoenix137:
Based on #5759 

Adds a debug command that reduces player level and adjusts life/mana accordingly. Does not refund stat points.

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -2440,6 +2440,22 @@ size_t OnSyncQuest(const TCmd *pCmd, size_t pnum)
 	return sizeof(message);
 }
 
+size_t OnCheatRemoveExperience(const TCmd *pCmd, size_t pnum) // NOLINT(misc-unused-parameters)
+{
+#ifdef _DEBUG
+	if (gbBufferMsgs == 1)
+		SendPacket(pnum, pCmd, sizeof(*pCmd));
+	else if (Players[pnum]._pLevel > 1) {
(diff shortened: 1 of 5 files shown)


AJenbo:
rebase

Source/msg.h
```diff
@@ -190,6 +190,10 @@ enum _cmd_id : uint8_t {
 	// body (TCmdParam1):
 	//    int16_t portal_num
 	CMD_WARP,
+	// Cheat: give player level down.
```
AJenbo, line 193: give down?

Source/player.cpp
```diff
@@ -2629,6 +2629,52 @@ int CalcStatDiff(Player &player)
 	return diff;
 }
 
+void PreviousPlrLevel(Player &player)
```
AJenbo, line 2632: can this be moved to debug.cpp?

Source/msg.cpp
```diff
@@ -3439,6 +3455,8 @@ size_t ParseCmd(size_t pnum, const TCmd *pCmd)
 		return OnFriendlyMode(pCmd, player);
 	case CMD_SYNCQUEST:
 		return OnSyncQuest(pCmd, pnum);
+	case CMD_CHEAT_REMOVE_EXPERIENCE:
+		return OnCheatRemoveExperience(pCmd, pnum);
```
AJenbo, line 3459: Could we do this without sending a network message? Even move CMD_CHEAT_EXPERIENCE ot of here as well?

PR #5763, opened 2023-02-06
Add `playerdat`

kphoenix137:
Moves a bunch of code out of `player.cpp` into `playerdat.cpp`, and centralizes player related table data.


Source/playerdat.h
```diff
@@ -0,0 +1,64 @@
+/**
+ * @file playerdat.h
```
glebm, line 2: We use `.hpp` for new headers.

Source/playerdat.cpp
```diff
@@ -0,0 +1,109 @@
...
+	// clang-format on
+};
+
+/** Contains the data related to each monster ID. */
```
glebm, line 82:
```suggestion
/** Contains the data related to each player class. */
```

Source/player.cpp
```diff
@@ -1890,18 +1759,19 @@ int Player::GetCurrentAttributeValue(CharacterAttribute attribute) const
...
+	switch (attribute) {
+	case CharacterAttribute::Strength:
+		return plrData.maxStr;
+	case CharacterAttribute::Magic:
+		return plrData.maxMag;
+	case CharacterAttribute::Dexterity:
+		return plrData.maxDex;
+	case CharacterAttribute::Vitality:
+		return plrData.maxVit;
+	default:
```
glebm, line 1772: Best not have a `default` branch in enum switches.
Without the default branch, if a new value is added to the enum and is not handled, the compiler will warn you.
Can put `app_fatal` after the `switch`

Source/playerdat.cpp
kphoenix137: 😅

kphoenix137:
Since `DFrames` and `TNFrames` have the same value regardless of character class, should these variables be removed from PlayersData[] and just set each to `20` in `SetPlrAnims()`?

Source/player.cpp
```diff
@@ -2362,112 +2232,68 @@ void NewPlrAnim(Player &player, player_graphic graphic, Direction dir, Animation
...
+	PlayerData plrData = PlayersData[static_cast<uint8_t>(pc)];
+	PlayerAttackAnimData plrAtkAnimData = PlayersAttackAnimData[static_cast<uint8_t>(pc)];
+	auto gn = static_cast<PlayerWeaponGraphic>(player._pgfxnum & 0xFU);
 
 	if (leveltype == DTYPE_TOWN) {
-		player._pNFrames = PlrGFXAnimLens[static_cast<std::size_t>(pc)][7];
-		player._pWFrames = PlrGFXAnimLens[static_cast<std::size_t>(pc)][8];
-		player._pDFrames = PlrGFXAnimLens[static_cast<std::size_t>(pc)][4];
-		player._pSFrames = PlrGFXAnimLens[static_cast<std::size_t>(pc)][5];
+		player._pNFrames = plrData.TNFrames;
```
kphoenix137, line 2240: @AJenbo I meant doing something like
`player._pNFrames = 20;`

AJenbo: ik

kphoenix137: Oh okay! I'll leave them as ia

Source/player.h
```diff
@@ -54,6 +54,7 @@ enum {
 	// clang-format on
 };
 
+// todo: move to playerdat.hpp?
```
AJenbo, line 57: Did you decide on this?

Source/player.cpp
```diff
@@ -3600,12 +3388,8 @@ void ModifyPlrMag(Player &player, int l)
...
 	player._pBaseMag += l;
 
-	int ms = l << 6;
-	if (player._pClass == HeroClass::Sorcerer) {
-		ms *= 2;
-	} else if (player._pClass == HeroClass::Bard) {
-		ms += ms / 2;
-	}
+	int ms = l;
+	ms *= PlayersData[static_cast<size_t>(player._pClass)].chrMana;
```
AJenbo, line 3392: I don't think there is a reason to multiply?
```suggestion
	int ms = PlayersData[static_cast<size_t>(player._pClass)].chrMana;
```

Source/player.cpp
```diff
@@ -3641,10 +3425,8 @@ void ModifyPlrVit(Player &player, int l)
 	player._pVitality += l;
 	player._pBaseVit += l;
 
-	int ms = l << 6;
-	if (player._pClass == HeroClass::Warrior || player._pClass == HeroClass::Barbarian) {
-		ms *= 2;
-	}
+	int ms = l;
+	ms *= PlayersData[static_cast<size_t>(player._pClass)].chrLife;
```
AJenbo, line 3429:
```suggestion
	int ms = PlayersData[static_cast<size_t>(player._pClass)].chrLife;
```

Source/playerdat.hpp
```diff
@@ -0,0 +1,136 @@
...
+	uint8_t swFire;
+	/* Sprite width: Cast Magic Spell */
+	uint8_t swMagic;
+	/* Sprite width: Death */
+	uint8_t swDeath;
+};
+
+struct PlayerAttackAnimData {
+	/* Unarmed frame count */
+	int8_t AFramesUnarmed;
```
glebm, line 94: The fields should be named in `camelCase`, for example `framesUnarmed`.
If you use VS Code, it should be complaining about this btw.

AJenbo: Done

Source/player.h
AJenbo: Lets not move it, to messy

AJenbo:
@glebm looks like this will allow us to shrink the player struct by about 9bytes by just referering to PlayerAnimData instead of copying it to the player object?

AJenbo (approved):

PR #5764, opened 2023-02-06
added instructions for Xbox Retail Mode

fwannmacher:
Added instructions for Retail Mode as it is quite different than Dev Mode


AJenbo:
Is it still currently possible to install it in retail mode?

The doe the installation instructions here still apply?
https://github.com/diasurgical/devilutionX/blob/master/docs/installing.md

AJenbo (approved):

PR #5765, opened 2023-02-06
Multiplayer: Enable Black Mushroom Quest

obligaron:
Contributes to #926

This one was annoying. 😄

Notes:
- Ensures spectral elixir is only usable when quest is completed (old comment: `/// BUGFIX: This will cause the elixir to be usable in the next game`)
- qmsg (displayed quest message) is now synced, cause this is the first quest that frequently changes it
- Spawn Fungal Tome during level generation to avoid desyncs
- Fixes a bug where quest items can't get picked up after dropping (`CF_PREGEN` was set to late)
- Brain now spawns differently on multiplayer to avoid desyncs (monster can now drop a normal item and the brain)
- Sync quest state and make mushroom patch state
- Spectral elixir still give the full bonus. In a follow-up pr we could change it to some random effect (similar to #5619). For now I would leave it to avoid the discussions in this pr. 😉


Source/itemdat.cpp
```diff
@@ -70,7 +70,7 @@ ItemData AllItemsList[] = {
...
 /*IDI_FUNGALTM   */ { IDROP_NEVER,   ICLASS_QUEST,  ILOC_UNEQUIPABLE, ICURS_FUNGAL_TOME,                 ItemType::Misc,        UITYPE_NONE,        N_("Fungal Tome"),                 nullptr,              0,            0,        0,        0,       0,       0,        0,        0,        0, ItemSpecialEffect::None,            IMISC_NONE,      SpellID::Null,           false,          0 },
-/*IDI_SPECELIX   */ { IDROP_NEVER,   ICLASS_MISC,   ILOC_UNEQUIPABLE, ICURS_SPECTRAL_ELIXIR,             ItemType::Misc,        UITYPE_ELIXIR,      N_("Spectral Elixir"),             nullptr,             15,            0,        0,        0,       0,       0,        0,        0,        0, ItemSpecialEffect::None,            IMISC_SPECELIX,  SpellID::Null,           false,          0 },
+/*IDI_SPECELIX   */ { IDROP_NEVER,   ICLASS_MISC,   ILOC_UNEQUIPABLE, ICURS_SPECTRAL_ELIXIR,             ItemType::Misc,        UITYPE_ELIXIR,      N_("Spectral Elixir"),             nullptr,             15,            0,        0,        0,       0,       0,        0,        0,        0, ItemSpecialEffect::None,            IMISC_SPECELIX,  SpellID::Null,           true,           0 },
```
AJenbo, line 73: can we make this table const now?

Source/player.h
```diff
@@ -372,6 +372,8 @@ struct Player {
 
 	bool CanUseItem(const Item &item) const
 	{
+		if (item.IDidx == IDI_SPECELIX && Quests[Q_MUSHROOM]._qactive != QUEST_DONE)
```
AJenbo, line 375: i guess the down side is that if you bring the elexier with you in to a new game it can no longer be used. Still that is better then the current situation.

Source/items.cpp
```diff
@@ -2999,11 +3003,21 @@ void SpawnItem(Monster &monster, Point position, bool sendmsg)
...
 		// Unqiue monster is killed => use better item base (for example no gold)
 		idx = RndUItem(&monster);
-	} else if (Quests[Q_MUSHROOM]._qactive == QUEST_ACTIVE && Quests[Q_MUSHROOM]._qvar1 == QS_MUSHGIVEN) {
+	} else if (dropBrain && !gbIsMultiplayer) {
 		// Normal monster is killed => need to drop brain to progress the quest
-		idx = IDI_BRAIN;
 		Quests[Q_MUSHROOM]._qvar1 = QS_BRAINSPAWNED;
+		NetSendCmdQuest(true, Quests[Q_MUSHROOM]);
+		// brain replaces normal drop
+		idx = IDI_BRAIN;
```
AJenbo, line 3011: since this is only for single player do we need to call netsend?

Source/itemdat.cpp
obligaron: Good point. Fixed 🙂

Source/player.h
obligaron: The worst part of the old behavior has many disadvantages even in singleplayer when you finished the quest:
- when you didn't use the elixir, saved and restarted the application you couldn't use it
- when you loaded a old save state where you had the elixir but didn't finished the quest you could use it and got a broken quest

Source/items.cpp
obligaron: Strictly speaking no we don't need it.
I would still keep it. It doesn't harm and when you search if you send all quest updates over the wire you don't have to think about this spot.
We also have another one that is not strictly necessary, cause it's only used in singleplayer (when spawning the Fungal Tome).

AJenbo (approved):

PR #5766, opened 2023-02-06
Damage vs. demons description string

ikonomov:
Pulling out the number out the description string.  This makes it easier for modding, resolving issues with the translation.  The same can be done for a few other affixes, if we decide this is something that is desirable.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3547,7 +3547,7 @@ bool DoOil(Player &player, int cii)
 	case IPL_KNOCKBACK:
 		return _("knocks target back");
 	case IPL_3XDAMVDEM:
-		return _(/*xgettext:no-c-format*/ "+200% damage vs. demons");
+		return fmt::format(fmt::runtime(_(/*xgettext:no-c-format*/ "+{:d}% damage vs. demons")), 200);
 	case IPL_ALLRESZERO:
 		return _("All Resistance equals 0");
 	case IPL_STEALMANA:
--- a/Translations/bg.po
+++ b/Translations/bg.po
@@ -4273,8 +4273,8 @@ msgstr "избутва целта назад"
 
 #: Source/items.cpp:3666
 #, no-c-format
-msgid "+200% damage vs. demons"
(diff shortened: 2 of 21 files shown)


ikonomov:
The other affixes' descriptions that I think we can edit the same way are:
IPL_THORNS
IPL_STEALMANA
IPL_STEALLIFE
IPL_PERIL
IPL_JESTERS
IPL_MANATOLIFE
IPL_LIFETOMANA

Source/items.cpp
```diff
@@ -3547,7 +3547,7 @@ bool DoOil(Player &player, int cii)
 	case IPL_KNOCKBACK:
 		return _("knocks target back");
 	case IPL_3XDAMVDEM:
-		return _(/*xgettext:no-c-format*/ "+200% damage vs. demons");
+		return fmt::format(fmt::runtime(_(/*xgettext:no-c-format*/ "+{:d}% damage vs. demons")), 200);
```
AJenbo, line 3550: Please derive this from it's source instead of hardcoding it a secound time

ikonomov: ok, great idea!

ikonomov: Correction to my above comment.  A great idea for another PR.
AJenbo, my PR was meant to change not the hardcoding of the affix itself, but its description only as it relates to the translation files.  Maybe it would be nice to have the values pulled from itemdat.cpp, but I currently do not know how to do this.  I'm trying to follow how it's done for the other affixes, for example item._iPLToHit and I see entries of this in many places including loadsave.cpp.  Unless I'm missing something simple that can allow me to pass those values, it seems that this is not going to be as simple as what I have done.  Does changing the hardcoding for the translations only absolutely require that we remove the hardcoding for the values themselves?  If so I guess what I have done here is not good enough and is destined to wait because I'm simply not able to do what you are asking and highly doubt anybody else will have any interest to do this.

ikonomov: Also I don't know why the checks are failing, I was able to build this in VS.

PR #5768, opened 2023-02-07
Description string for fixed value affixes

ikonomov:
Removes the hardcoding for translations of descriptions to:
`IPL_3XDAMVDEM`
`IPL_STEALMANA`
`IPL_STEALLIFE`
`IPL_MANATOLIFE` fixed with added *no-c-format*
`IPL_LIFETOMANA` fixed with added *no-c-format*

Also removed added *no-c-format* from `IPL_JESTERS` since the % number was not used in the translation files.


ikonomov:
Please remove the Draft mark, whenever somebody is able.  Thank you!

PR #5769, opened 2023-02-07
Refactor PrintItemPower()

kphoenix137:
Removes hardcoding for IPL_THORNS, IPL_STEALMANA, IPL_STEALLIFE, IPL_MANATOLIFE, and IPL_LIFETOMANA.

Jester's and Peril don't use their parameters and it seems like it would be a complicated manner to make them, so I left them as their hardcoded values.


ikonomov:
You might want to add `/*xgettext:no-c-format*/ ` to IPL_MANATOLIFE and IPL_LIFETOMANA as well as `#, no-c-format` to their translation lines.  I'm changing them now in my PR, https://github.com/diasurgical/devilutionX/pull/5768/, I am happy with either solution.  Thank you KP for doing this.

ikonomov:
Also /*xgettext:no-c-format*/ is not needed for Jester's.

Source/items.cpp
```diff
@@ -3441,6 +3443,45 @@ bool DoOil(Player &player, int cii)
 
 [[nodiscard]] StringOrView PrintItemPower(char plidx, const Item &item)
 {
+	int param1, param2 = 0;
```
qndel, line 3446: why only assign 0 to param2? param1 is still uninitialized

kphoenix137: Mistake

PR #5770, opened 2023-02-07
Multiplayer: Spectral Elixir only randomly gives full bonus

obligaron:
Follow-up of #5765 and similar to #5619

Notes
- Spectral Elixir only has a 10% to get the full bonus (+3 to all attributes)
- In 90% of all uses it adds +3 to a random attribute
- When selecting the random attributes max stats are ignored
- Change doesn't affect singleplayer

All numbers above aren't fix. They can be discussed and be changed.
I hope with this pr can give us more options how to balance Spectral Elixir in multiplayer. 🙂

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -3887,10 +3887,38 @@ void UseItem(size_t pnum, item_misc_id mid, SpellID spl)
 		NewCursor(CURSOR_OIL);
 		break;
 	case IMISC_SPECELIX:
-		ModifyPlrStr(player, 3);
-		ModifyPlrMag(player, 3);
-		ModifyPlrDex(player, 3);
-		ModifyPlrVit(player, 3);
+		if (!gbIsMultiplayer || FlipCoin(10)) { // Don't give the full bonus every time in multiplayer
+			ModifyPlrStr(player, 3);
+			ModifyPlrMag(player, 3);
+			ModifyPlrDex(player, 3);
+			ModifyPlrVit(player, 3);
+		} else {
+			// Find attributes that can increased
+			std::vector<CharacterAttribute> increasableAttributes;
+			for (auto attribute : enum_values<CharacterAttribute>()) {
+				if (player.GetBaseAttributeValue(attribute) >= player.GetMaximumAttributeValue(attribute))
+					continue; // attribute can't be increased
+				increasableAttributes.push_back(attribute);
+			}
+			if (increasableAttributes.empty())
+				break; // no attributes to increase
+			size_t index = static_cast<size_t>(RandomIntLessThan(increasableAttributes.size()));
+			CharacterAttribute attributeToIncrease = increasableAttributes[index];
+			switch (attributeToIncrease) {
+			case CharacterAttribute::Strength:
+				ModifyPlrStr(player, 3);
+				break;
+			case CharacterAttribute::Magic:
+				ModifyPlrMag(player, 3);
+				break;
+			case CharacterAttribute::Dexterity:
+				ModifyPlrDex(player, 3);
+				break;
+			case CharacterAttribute::Vitality:
+				ModifyPlrVit(player, 3);
+				break;
+			}
+		}
 		break;
 	case IMISC_RUNEF:
 		player._pTSpell = SpellID::RuneOfFire;


ikonomov:
I see this as necessary.  I think we can also consider doing a random generation on each stat bonus.  Maybe the randomization can give bigger chances to get no bonus, and +3 would be the hardest to get.  Say Random 0-9, 4 out of 10 give no bouns, 3 give +1, 2 gives +2 and 1 gives +3.

kphoenix137:
I'm not entirely convinced on changing the mechanics arbitrarily. I think the multiplayer behavior should match the single player behavior. In Single Player, you'd also be able to farm this elixir for the result you're trying to prevent in multiplayer. I can see that this is indeed a balance change to multiplayer (as are all single player exclusive items being introduced into MP), although I don't think the functionality of these items should be modified in any capacity. Either remove them entirely, reduce the drop/reward rate which you get them at, or full send.

yuripourre:
@kphoenix137 I like your approach. My two cents: maybe multiplayer could give you a standard elixir (str,dex,mag,vit) and sometimes a spectral.

kphoenix137:
> @kphoenix137 I like your approach. My two cents: maybe multiplayer could give you a standard elixir (str,dex,mag,vit) and sometimes a spectral.

I'd prefer this, and it's in line with the other quest rewards he has set up, where the sp item only drops a small percentage of the time, while a random item drops most of the time.

obligaron:
Initially I also thought about dropping a normal elixir. But it's not that easy.
- Pepin spawns the spectral elixir
- You need to pick it up and bring it to Adria
- Adria checks that you have the spectral elixir in your inventory and unlocks it (now it's usable and the quest is finished)

To proceed the quest the spectral elixir is needed (Adria checks the inventory). To make the quest work normally:

- We could change the elixir when Adria unlocks it (replace spectral elixir with a normal one)
- We could let pepin drop a normal elixir that is not entirely normal cause you need it to flag it somehow to proceed the quest
- We could let the quest finish after Pepin spawns the elixir
- Something else 😜

All of the above options don't match well with the dialogs that the npcs have or are confusing (normal elixir that is not a normal elixir).
That's why I used the approach to randomize the effect of the give spectral elixir, cause I thought it's a easy and the less confusing approach. But I don't have a strong opinion on this.

Another note: If we want to make the effect more clear, we could show a message in multiplayer what the attributes the spectral elixir changed (like shrine messages).

ikonomov:
I agree with obligaron, even if all issues are resolved the dialogs talking about a special elixir that is just a regular elixir is something that should be avoided.

FitzRoyX:
I think most people will be happy to just have working quests. Why not worry about the balance and flavor stuff later?

obligaron:
> I think most people will be happy to just have working quests. Why not worry about the balance and flavor stuff later?

I made separated pr's for the enabling and the balancing, so we can discuss this separately without blocking any progress in making the quests working. 😉

Personally I think there is a difference between singelplayer and multiplayer when playing the quests

- Multiplayer don't save the world state (including quest state).
- Singleplayer often is only played once but multiplayer gets replayed more. Imho that means in sinlgeplayer you should get all content (quest rewards) in the first run and multiplayer should be more interesting to replay.
- Diablo singleplayer doesn't allow you to use town shortcuts to enter specify dungeons after starting a new game, but multiplayer does.
- For multiplayer the quests are new content that no vanilla dev ever balanced.

On the other hand this balances are only arbitrary changes, but enabling the quest is the same. 😉

If we conclude that we don't want to change the behavior of the quests I'm also fine to close the pr's.
But I think it's worth a discussion. 🙂

kphoenix137:
> > I think most people will be happy to just have working quests. Why not worry about the balance and flavor stuff later?
> 
> I made separated pr's for the enabling and the balancing, so we can discuss this separately without blocking any progress in making the quests working. 😉
> 
> Personally I think there is a difference between singelplayer and multiplayer when playing the quests
> 
> - Multiplayer don't save the world state (including quest state).
> - Singleplayer often is only played once but multiplayer gets replayed more. Imho that means in sinlgeplayer you should get all content (quest rewards) in the first run and multiplayer should be more interesting to replay.
> - Diablo singleplayer doesn't allow you to use town shortcuts to enter specify dungeons after starting a new game, but multiplayer does.
> - For multiplayer the quests are new content that no vanilla dev ever balanced.
> 
> On the other hand this balances are only arbitrary changes, but enabling the quest is the same. 😉
> 
> If we conclude that we don't want to change the behavior of the quests I'm also fine to close the pr's.
> But I think it's worth a discussion. 🙂

Spectral Elixir as is would certainly pose a balance problem in MP considering the town warps and how easy it is to complete this quest. I also don't think the elixir itself should be changed. Honestly i think this quest is the most difficult one in question to balance correctly. I'd almost suggest disallowing the quest from MP or not having the option to have all quests in every multiplayer game (and subsequently reducing the rate at which this quest appears) in order to keep the quest fully intact but prevent consist farming of the quest. I can see the downsides to that as well.

Trihedraf:
I think we should just change it when Adria checks it. And I think it should be 90% a regular elixir and 10% a spectral elixir. Maybe is there a way that we could add line to the regular elixir description that says like (spectral elixir) or something.

kphoenix137:
> I think we should just change it when Adria checks it. And I think it should be 90% a regular elixir and 10% a spectral elixir. Maybe is there a way that we could add line to the regular elixir description that says like (spectral elixir) or something.

Same result and fewer steps to change MP Spectral Elixir to pick one random stat and raise it by 1

StephenCWills:
Personally, I think Spectral Elixir is at most just as unbalanced as, say, shopping MBoS, KSoH, books, or regular elixirs. Stats are still capped for each class regardless of how you choose to farm. I don't really see how it clearly poses a balance problem.

kphoenix137:
> Personally, I think Spectral Elixir is at most just as unbalanced as, say, shopping MBoS, KSoH, books, or regular elixirs. Stats are still capped for each class regardless of how you choose to farm. I don't really see how it clearly poses a balance problem.

Its the ease of access, not the item itself. Its a free 12 elixirs packed into one that you could in theory farm one of every 3-5 minutes

Trihedraf:
Well except for vitality you can just Adria farm probably faster than doing the whole quest. So maybe it's not really that big of a deal if it always just drops the standard spectral elixir.

StephenCWills:
Currently, to get 12 elixirs, you only have to farm 60k gold and Adria. The main difference is that you have access to this quest at a much lower clvl. But twinking was already a thing, and clvl is also important to become OP so.. what is the actual balance implication here?

kphoenix137:
> Currently, to get 12 elixirs, you only have to farm 60k gold and Adria. The main difference is that you have access to this quest at a much lower clvl. But twinking was already a thing, and clvl is also important to become OP so.. what is the actual balance implication here?

I changed my mind. I agree it should be added as is

obligaron:
If we think are rebalance for multiplayer is not needed, I'm fine with closing this pr.
If no one objects in the next week, I will close it.

obligaron:
No objections => closed 😉

qndel:
:(

obligaron:
@qndel you could have sad something in this week 😉

btw: We don't forbid a discussion. If a majority wants a balance change I'm fine with reopening this pr.

PR #5771, opened 2023-02-07
Bugfix: Hellfire "Life to mana" and "Mana to life" 

### LINKED ISSUE #6875: [Issue Report]: Overflow with Acolyte Amulet?
### Operating System

Windows x64

### DevilutionX version

1.5.1

### Describe

I play as a Sorcerer and get to high level, in Nightmare I find the Acolyte's Amulet, which is a Hellfire item.
I think "Wow, this is perhaps the least useful drop yet!" and put it in my stash.

Fast forward some days, and I make a Warrior, he's level 1 and I go look in my stash for fun stuff I can start him out with. Seeing the Acolyte's Amulet, which transfers 50% of mana to health, I try it out for the fun of it. Expecting the meager mana to add some extra life, instead I get THIS result.

Screenshot from 2023-12-21 12-50-09
Now, this doesn't look right at all. I don't know if this is some sort of overflow or if the amulet decided to remember my Sorcerer's massive mana pool for some reason and apply that on my Warrior. After I took a screenshot and went to write this, the game also ended up freezing/crashing, so I assume that it created some situation the .exe did not like.

Loading the game up again, there he is, but this time the health and HP values are normal, so it seems self-correcting, in a way.

### To Reproduce

.

### Expected Behavior

_No response_

### Additional context

_No response_

kphoenix137:
Fixes: https://github.com/diasurgical/devilutionX/issues/6875

Hellfire writes the Life and Mana directly to the item. This is a problem since it does not dynamically change the values, but keeps the values from when the item is originally generated. This becomes less of a problem the more frequently the item with one of these powers is regenerated, however it is a terrible way of doing it.

This PR rather sets the Life and Mana values dynamically so they stay accurate to the item description.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -2648,6 +2648,7 @@ void CalcPlrItemVals(Player &player, bool loadgfx)
 					int portion = ((player._pMaxHPBase >> 6) * 40 / 100) << 6;
 					ihp -= portion;
 					imana += portion;
+				}
 				if (HasNoneOf(item._iDamAcFlags, ItemSpecialEffectHf::ManaToLife, ItemSpecialEffectHf::LifeToMana)) {
 					ihp += item._iPLHP;
 					imana += item._iPLMana;
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1035,11 +1035,15 @@ int SaveItemPower(const Player &player, Item &item, ItemPower &power)
 		item._iDamAcFlags |= ItemSpecialEffectHf::ACAgainstUndead;
 		break;
 	case IPL_MANATOLIFE: {
+		item._iDamAcFlags |= ItemSpecialEffectHf::ManaToLife;
+		// Calculate for reverse compatibility
 		int portion = ((player._pMaxManaBase >> 6) * 50 / 100) << 6;
 		item._iPLMana -= portion;
 		item._iPLHP += portion;
 	} break;
 	case IPL_LIFETOMANA: {
+		item._iDamAcFlags |= ItemSpecialEffectHf::LifeToMana;
+		// Calculate for reverse compatibility
 		int portion = ((player._pMaxHPBase >> 6) * 40 / 100) << 6;
 		item._iPLHP -= portion;
 		item._iPLMana += portion;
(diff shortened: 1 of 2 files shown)


Source/items.cpp
```diff
@@ -2375,8 +2371,21 @@ void CalcPlrItemVals(Player &player, bool loadgfx)
...
+				// Check for Acolyte's Amulet and Gladiator Ring to apply bonuses as life and mana changes, rather than getting static bonuses from the item data
+				if (IsAnyOf(item._iDamAcFlags, ItemSpecialEffectHf::ManaToLife)) {
+					int portion = ((player._pMaxManaBase >> 6) * 50 / 100) << 6;
+					imana -= portion;
+					ihp += portion;
+				} else if (IsAnyOf(item._iDamAcFlags, ItemSpecialEffectHf::LifeToMana)) {
+					int portion = ((player._pMaxHPBase >> 6) * 40 / 100) << 6;
+					ihp -= portion;
+					imana += portion;
+				} else {
```
DakkJaniels, line 2384: Should this not be included in the if/else? wouldn't you want to apply item bonuses after doing the ManaToLife or LifeToMana? Also, what if you are wearing both?

DakkJaniels: Nevermind my bad, I was thinking this was the player loop, this is looping through items, so it shouldn't be an issue.

Source/items.cpp
```diff
@@ -1010,16 +1010,12 @@ int SaveItemPower(const Player &player, Item &item, ItemPower &power)
...
-		int portion = ((player._pMaxManaBase >> 6) * 50 / 100) << 6;
-		item._iPLMana -= portion;
-		item._iPLHP += portion;
-	} break;
-	case IPL_LIFETOMANA: {
-		int portion = ((player._pMaxHPBase >> 6) * 40 / 100) << 6;
-		item._iPLHP -= portion;
-		item._iPLMana += portion;
-	} break;
+	case IPL_MANATOLIFE:
```
DakkJaniels, line 1013: Since we skip the hardcoded values when calculating the items for  DevX, why not just hardcode in the values to maintain reverse compatibility?

Source/items.cpp
kphoenix137: This wouldn't be a problem under normal circumstances, but for the sake of modding, if someone wanted to put this effect on an item as well as another effect that modifies hp and/or life, this would be a problem.

PR #5772, opened 2023-02-08
Add ItemSpecialEffect::Indestructible

kphoenix137:
Adds a new flag for Indestructible, so that 255 durability doesn't need to be used as indestructible. Indestructible items are given both the new flag, and also have their durability set to 255 for reverse compatibility. Also added code in `CalcPlrItemVals()` to force set the flag on any existing item if it has `IPL_INDESTRUCTIBLE` but doesn't already have the flag (for reverse compatibility).


Source/objects.cpp
```diff
@@ -2349,24 +2349,25 @@ void OperateShrineHidden(Player &player)
...
 			player.InvBody[r]._iMaxDur -= 20;
-			if (player.InvBody[r]._iDurability <= 0)
-				player.InvBody[r]._iDurability = 1;
-			if (player.InvBody[r]._iMaxDur <= 0)
-				player.InvBody[r]._iMaxDur = 1;
+
 			break;
 		}
+		for (auto &item : player.InvBody) {
+			item._iDurability = clamp(item._iDurability, 1, 255);
```
yuripourre, line 2368:
```suggestion
			item._iDurability = clamp(item._iDurability, 1, 255);
                        item._iDurability = clamp(item._iDurability, 1, DUR_MAX);
```

Source/items.cpp
```diff
@@ -2656,6 +2674,8 @@ void CalcPlrItemVals(Player &player, bool loadgfx)
 		}
 	}
 
+
```
yuripourre, line 2677: Are those blank lines intended?

Source/objects.cpp
kphoenix137: Could have sworn I committed this exact thing before getting off for the night lol. Guess I forgot to push

Source/items.cpp
```diff
@@ -2385,6 +2386,23 @@ void CalcPlrItemVals(Player &player, bool loadgfx)
 				lmax += item._iLMaxDam;
 			}
 		}
+
+		// Creates reverse compatibility, since items with IPL_INDESTRUCTIBLE from older versions/vanilla won't have ItemSpecialEffect::Indestructible
```
StephenCWills, line 2390: I feel like this should be done in `UnPackItem()` and `LoadItemData()`.

StephenCWills (changes requested):

Source/pack.cpp
```diff
@@ -315,6 +315,23 @@ void UnPackItem(const ItemPack &packedItem, const Player &player, Item &item, bo
...
+				}
+			}
+			break;
+		case ITEM_QUALITY_MAGIC:
+			if (item._iSufPower == IPL_INDESTRUCTIBLE) {
+				item._iFlags |= ItemSpecialEffect::Indestructible;
+			}
+			break;
+		}
+
```
StephenCWills, line 334: Now that I'm looking at this, it occurs to me that we shouldn't need it because `RecreateItem()` will set the flag when calling `SaveItemPower()` during regeneration. We only the one in `LoadItemData()` to fix SP save data and `heroitems` data.
```suggestion
```

StephenCWills (dismissed):

AJenbo:
Won't this break Hellfire items that have been oiled to be indestructable?

StephenCWills:
> Won't this break Hellfire items that have been oiled to be indestructable?

The only thing I can think to do about that is to add a flag to `icreateinfo_flag2` so we can differentiate between items before and after this PR. Otherwise, we can't tell the difference between oiled vanilla SP Hellfire items and DevX SP Hellfire items with 255 durability.

Well, either that, or we can simply ignore the problem. Continue to allow SP Hellfire players to abuse Hidden shrines since they have access to Oil of Permanence anyway.

PR #5773, opened 2023-02-08
Bugfix: Items over 255 durability

kphoenix137:
Some items with the `of Structure` power can exceed 255 durability, causing the item to drop to a low durability when regenerated. This clamps the bonus given from the affix so that it doesn't exceed 255.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -855,7 +855,7 @@ int SaveItemPower(const Player &player, Item &item, ItemPower &power)
 		RedrawComponent(PanelDrawComponent::Mana);
 		break;
 	case IPL_DUR: {
-		int bonus = std::min(r * item._iMaxDur / 100, DUR_INDESTRUCTIBLE - item._iMaxDur);
+		int bonus = std::min(r * item._iMaxDur / 100, DUR_INDESTRUCTIBLE - item._iMaxDur - 1);
 		item._iMaxDur += bonus;
 		item._iDurability += bonus;
 	} break;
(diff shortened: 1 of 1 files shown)


StephenCWills (dismissed):

AJenbo:
Should it not be clamped to 254?

StephenCWills:
Hm, maybe yes, if you assume #5772 won't be merged any time soon. I was kinda hoping it was close to ready, though I can't say I know what we should do about Oil of Permanence.

kphoenix137:
approved, when merge?

Source/items.cpp
```diff
@@ -855,7 +855,7 @@ int SaveItemPower(const Player &player, Item &item, ItemPower &power)
 		RedrawComponent(PanelDrawComponent::Mana);
 		break;
 	case IPL_DUR: {
-		int bonus = r * item._iMaxDur / 100;
+		int bonus = std::min(r * item._iMaxDur / 100, DUR_INDESTRUCTIBLE - item._iMaxDur);
```
AJenbo, line 858: I don't think we want to make them indestructible, so this PR either depends on https://github.com/diasurgical/devilutionX/pull/5772 or the following change should be made:

```suggestion
		int bonus = std::min(r * item._iMaxDur / 100, DUR_INDESTRUCTIBLE - item._iMaxDur - 1);
```

AJenbo:
> approved, when merge?

See previous comments, or the one I just made.

PR #5775, opened 2023-02-08
Fix item levels for set/quest-maps

### LINKED ISSUE #5757: Bad rewards from the chests in the quest "Leoric's Tomb" and "Chamber of bone"
I've noticed for a while that there are very low quality items in the chests from the above quests.
The value of the items is often only slightly more than the value of 100 gold coins.
I'm relatively sure that the value of the items should be much higher.
I would ask all contributors to at least confirm the issue, so we can narrow it down soon.

obligaron:
Fixes #5757

After #5196 `currlevel` is always identically with `setlvlnum` for set/quest-maps.
But `currlevel` is used in [`ItemsGetCurrlevel`](https://github.com/diasurgical/devilutionX/blob/98294e0ad57865331202b916e2f71cd7fa6c82ae/Source/items.cpp#L388-L397) to determinate the value of the item.
With this pr we always use the entry level of the set/quest-map for determinating the value of a item.

Note:
This can also happen in vanilla if you open a town portal in the set/quest-map, go to town, go back and open a chest/kill a monster.
See #5196 for further analysis. :wink:

Thanks @Chance4us for reporting the issue. 🙂

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -387,6 +387,21 @@ bool IsSuffixValidForItemType(int i, AffixItemType flgs)
 
 int ItemsGetCurrlevel()
 {
+	if (setlevel) {
+		switch (setlvlnum) {
+		case SL_SKELKING:
+			return Quests[Q_SKELKING]._qlevel;
+		case SL_BONECHAMB:
+			return Quests[Q_SCHAMB]._qlevel;
+		case SL_POISONWATER:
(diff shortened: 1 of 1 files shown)


Chance4us:
Thanks for fixing @obligaron :smiley:

ikonomov:
Fantastic work guys, catching and fixing this easy to miss this but very important issue!

AJenbo (approved):

PR #5777, opened 2023-02-08
Fix Life/Mana bugs

kphoenix137:
Fixes: https://github.com/diasurgical/devilutionX/issues/6520

This PR fixes problems related to Life and Mana regarding negative numbers.

- Max Life will never go under 1 (Fixes town death bug)
- Max Mana will never go under 0 (I would have outright removed ALL negative numbers for life and mana, however it's absolutely essential to keep a relative relationship between Current Mana and Max Mana to avoid having Current Mana changed when equipping and unequipping items)
- Current Mana visually appears to never go under 0 to the player (Not able to actually set Current Mana to 0, since this would allow the player to exploit this mechanic and fully restore their Mana.)
- Current Life and Current Mana can no longer exceed Max Life and Max Mana

The fixes to Max Life and Max Mana will not cause an exploit where the user can gain Current Life or Current Mana by equipping cursed items and unequipping them (excluding gaining Current Life while in town, an already exiting exploit, however not really an issue because the Healer can be used to gain Life for free).

Players can still die in the dungeon by equipping cursed items/unequipping +life items, however there isn't a good way around this, since min capping Current Life at 1 when equipping/unequipping items would allow the player to fully restore their Life.

The town death bug occurs when a player dies with a negative Max Life. If the player still has negative Max Life while dead, respawning in Town will cause you to appear dead at the start location, and never be able to come back to life unless someone casts Resurrect on you. A similar bug appears while in the dungeon. If you are alive with negative Max Life, casting Healing or having someone else cast Heal Other on you will kill you. Haven't tested, but it's likely that Potion of Healing and Potion of Full Healing would also kill you.

### DIFF
--- a/Source/engine/render/scrollrt.cpp
+++ b/Source/engine/render/scrollrt.cpp
@@ -1687,10 +1687,10 @@ void DrawAndBlit()
 	DrawXPBar(out);
 	if (*sgOptions.Gameplay.showHealthValues)
 		DrawFlaskValues(out, { mainPanel.position.x + 134, mainPanel.position.y + 28 }, MyPlayer->_pHitPoints >> 6, MyPlayer->_pMaxHP >> 6);
-	if (*sgOptions.Graphics.showManaValues)
-		DrawFlaskValues(out, { mainPanel.position.x + mainPanel.size.width - 138, mainPanel.position.y + 28 }, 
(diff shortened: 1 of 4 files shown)


ikonomov:
A very important fix for potentially a game breaking bug!

AJenbo:
Can you rebase it and get it back in working shape :)

AJenbo (approved):

PR #5778, opened 2023-02-09
Fix mana and health gain at character level 50

ikonomov:
https://github.com/diasurgical/devilutionX/pull/3928 was opened about a year ago, not merged because it was supposed be added as part of https://github.com/diasurgical/devilutionX/pull/1626 that was opened two years ago.  It looks like people have different ideas about the colors of the character's panel and what numbers should be shown there, but in the mean time, if possible let's consider fixing the health and mana gain at character level 50 so that people can get the benefit of this gameplay fix.


ikonomov:
fixed in https://github.com/diasurgical/devilutionX/pull/5782

PR #5779, opened 2023-02-09
Debug: Add command to give experience points

kphoenix137:
Gives experience points. Ignores power level cap.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2679,7 +2679,7 @@ void NextPlrLevel(Player &player)
 	CalcPlrInv(player, true);
 }
 
-void AddPlrExperience(Player &player, int lvl, int exp)
+void AddPlrExperience(Player &player, int lvl, int exp, bool debug /*= false*/)
 {
 	if (&player != MyPlayer) {
 		return;
@@ -2699,7 +2699,7 @@ void AddPlrExperience(Player &player, int lvl, int exp)
 	uint32_t clampedExp = std::max(static_cast<int>(exp * (1 + (lvl - player._pLevel) / 10.0)), 0);
 
 	// Prevent power leveling
-	if (gbIsMultiplayer) {
+	if (gbIsMultiplayer && !debug) {
 		const uint32_t clampedPlayerLevel = clamp(static_cast<int>(player._pLevel), 1, MaxCharacterLevel);
 
 		// for low level characters experience gain is capped to 1/20 of current levels xp
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2679,7 +2679,7 @@ void NextPlrLevel(Player &player)
 	CalcPlrInv(player, true);
 }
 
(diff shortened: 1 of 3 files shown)


Source/player.cpp
```diff
@@ -2679,7 +2679,7 @@ void NextPlrLevel(Player &player)
 	CalcPlrInv(player, true);
 }
 
-void AddPlrExperience(Player &player, int lvl, int exp)
+void AddPlrExperience(Player &player, int lvl, int exp, bool debug /*= false*/)
```
AJenbo, line 2682: Not happy about modifying release code with debug code

Source/player.cpp
```diff
@@ -2699,7 +2699,7 @@ void AddPlrExperience(Player &player, int lvl, int exp)
 	uint32_t clampedExp = std::max(static_cast<int>(exp * (1 + (lvl - player._pLevel) / 10.0)), 0);
 
 	// Prevent power leveling
-	if (gbIsMultiplayer) {
+	if (gbIsMultiplayer && !debug) {
 		const uint32_t clampedPlayerLevel = clamp(static_cast<int>(player._pLevel), 1, MaxCharacterLevel);
```
AJenbo, line 2703:
```suggestion
#ifndef debug
	if (gbIsMultiplayer) {
		const uint32_t clampedPlayerLevel = clamp(static_cast<int>(player._pLevel), 1, MaxCharacterLevel);
#endif
```

Source/player.h
```diff
@@ -791,7 +791,7 @@ int CalcStatDiff(Player &player);
 #ifdef _DEBUG
 void NextPlrLevel(Player &player);
 #endif
-void AddPlrExperience(Player &player, int lvl, int exp);
+void AddPlrExperience(Player &player, int lvl, int exp, bool debug = false);
```
AJenbo, line 794:
```suggestion
void AddPlrExperience(Player &player, int lvl, int exp);
```

Source/player.cpp
qndel: It would

qndel:
Why not just increase player's xp directly? It's a debug command, it doesn't have to be super clean ;) Even if it doesn't trigger a level up, there's a separate command for that. Or expand it to fully emulate regular xp gain (specify "monster level" and xp it'd give etc.)

kphoenix137:
> Why not just increase player's xp directly? It's a debug command, it doesn't have to be super clean ;) Even if it doesn't trigger a level up, there's a separate command for that. Or expand it to fully emulate regular xp gain (specify "monster level" and xp it'd give etc.)

I originally set it up that way but changed it because I needed something to accurately help me test changing things in AddPlrExperience and NextPlrLevel

Source/player.cpp
kphoenix137: > It would
> 
> 

Maybe we can revisit this command if someone else has a need for it specifically that they can't use the command thats already there for

PR #5784, opened 2023-02-10
Black Mushroom Quest: Fix missing quest dialog

### LINKED ISSUE #5781: Black Mushroom quest behavior
The default behavior is as follows.
https://diablo-archive.fandom.com/wiki/Adria_(Diablo_I)#Black_Mushroom

Citation Quest Successful:
    "What? Now you bring me that elixir from the healer? I was able to finish my brew without it. Why don't you just keep it..."

In master the above sentence is missing, as soon you finish the quest.
What is also new is that the elixir turns red if it cannot/may not be used, but that is not a problem.

Spectral elixir red:
spectral_elixir

Missing speech/sentence:

https://user-images.githubusercontent.com/45902449/217844617-4c0afe1b-4106-4ce7-af87-680ff4a4ae76.mp4

Savegame:
[spectral_elixir.zip](https://github.com/diasurgical/devilutionX/files/10698326/spectral_elixir.zip)

obligaron:
Fixes #5781 

- Adds missing quest dialog back
- Doesn't show spectral elixir in red when quest is not finished

Thanks @Chance4us for reporting 🙂

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1058,7 +1058,7 @@ bool CanBePlacedOnBelt(const Item &item)
 	return FitsInBeltSlot(item)
 	    && item._itype != ItemType::Gold
 	    && MyPlayer->CanUseItem(item)
-	    && AllItemsList[item.IDidx].iUsable;
+	    && item.isUsable();
 }
 
 void FreeInvGFX()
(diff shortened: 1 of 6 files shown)


AJenbo (approved):
thanks :) looks nicer

PR #5787, opened 2023-02-13
Add toggles for disabling "Sturdy" and "Burgundy"

kphoenix137:
Adds toggles to Gameplay options that will visually appear to change "Sturdy" and "Burgundy" back to "Fine" and "Crimson", respectively. This does not write to item data, so it can be toggled on and off without permanently changing the names on items.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -1013,10 +1013,13 @@ void DrawInfoBox(const Surface &out)
 		} else if (!myPlayer.CanUseItem(myPlayer.HoldItem)) {
 			InfoString = _("Requirements not met");
 		} else {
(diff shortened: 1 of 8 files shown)


ikonomov:
If possible, could we just consider reverting back Sturdy into Fine to avoid items named Sturdy of Sturdiness.  I think nothing else is needed.

kphoenix137:
> If possible, could we just consider reverting back Sturdy into Fine to avoid items named Sturdy of Sturdiness. I think nothing else is needed.

How about toggles default to true, and people can enable the changed names if they please

Trihedraf:
I don't think there should be a toggle for this. It either should be fixed/changed or not.

Issue #5791, opened 2023-02-13
ConsumeScroll: Invalid item index 2


ChaosMarc:
### Operating System

Windows x64

### DevilutionX version

Custom build (please specify commit ID)

### Describe

When using the town portal scroll from the right click menu the game crashes with the following error:

![image](https://user-images.githubusercontent.com/1343751/218476913-f04a6b46-0d90-4255-a52f-33c142b30d77.png)

When right clicking the scroll in the inventory everything is fine.

I'm using the latest Windows MinGW x64 master build (b46482f41481826e562477619820c25c2a0bd244) from here: https://github.com/diasurgical/devilutionX/actions/runs/4155463518

Savegame: [single_3.zip](https://github.com/diasurgical/devilutionX/files/10722838/single_3.zip)

### To Reproduce

1. load the attached savegame
2. right click somewhere to open a town portal
3. crash

qndel:
4f9e9e04b1086141cda4361413a0c20dd03b72c7 is the first bad commit
commit 4f9e9e04b1086141cda4361413a0c20dd03b72c7
Author: Gleb Mazovetskiy <glex.spb@gmail.com>
Date:   Sun Dec 18 15:01:02 2022 +0000

    Fix ConsumeScroll

galaxyhaxz:
> Date: Sun Dec 18 15:01:02 2022 +0000
Fix ConsumeScroll

It's hilarious how a commit with that name caused this bug :P

qndel:
Using a scroll by selecting it in spells and then right clicking doesn't set `spellFrom`, in this case 2 came from loading the save.

qndel:
Casting nova crashes everyone else with index 3

glebm:
item slot 2 is `INVITEM_RING_RIGHT` and 3 is `INVITEM_AMULET` :sweat_smile:

obligaron:
This was fixed with #5988 and #5799

PR #5799, opened 2023-02-14
msg.cpp: Always set `spellFrom`

glebm:
Note that in a few places it is always being set to 0.

This is probably wrong and should instead be passed in the network message as a parameter.

For now, this at least fixes the crash observed in #5791.

/cc @obligaron

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -1525,6 +1525,7 @@ size_t OnSpellTile(const TCmd *pCmd, Player &player)
 	player.destParam3 = SDL_SwapLE16(message.wParam3);
 	player.queuedSpell.spellId = spell;
 	player.queuedSpell.spellType = static_cast<SpellType>(wParam2);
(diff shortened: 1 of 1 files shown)


obligaron (approved):
Personally I would still remove the `app_fatal` in `ConsumeScroll`.
`spellFrom` is read from the save.
I'm not sure what vanilla writes to `spellFrom`.
If we load a vanilla save at a "bad" point (casting is queued but not executed yet) we could still get the crash.

I still approve, cause for devilutionX this should fix the problem. 🙂

AJenbo:
We had similar crashes in a release from loading past saves with bad values for the current selected spell so probably would be good to sanity check that somehow.

glebm:
@obligaron Perhaps we should fix up the value when loading the save file.

For network stuff, we should make sure the value is actually passed in the message, otherwise the players' inventories can get out of sync.

Issue #5796, opened 2023-02-13
iOS not generating documents folder


616dustin:
### Operating System

iOS

### DevilutionX version

1.4.1 (latest release)

### Describe

Sideloaded via Altstore, launched DevilutionX, received a "Data File Error" warning. "Unable to open main data archive (diabdat.mpq or spawn.mpq

Make sure it is in the game folder"

i click OK and close the app then open files.app and go to "On My iPhone" and there is no DevilutionX folder. 

I tried creating the folder and adding the mpq's but i receive the same error.

### To Reproduce

1. Download devilutionx-ios.ipa
2. Upload devilutionx-ios.ipa to icloud folder.
3. In AltStore add devilutionx-ios.ipa for sideloading
4. launch devilutionx
5. receive missing mpq error
6. close devilutionx
7. open files.app
8.  Under Locations choose "On My iPhone"
9. no folder found for devilutionx
10. also tried searching in files.app for devilutionx and no results returned.

### Expected Behavior

there should be a devilutionx folder to upload the mpq's to.

### Additional context

iPhone 14 Pro Max
iOS Version: 16.0.3

AJenbo:
@bubio do you have an idea what could be the issue here?

bubio:
@616dustin Try to make the file name diabdat.mpq or spawn.mpq all uppercase! (DIABDAT.MPQ)

DakkJaniels:
> @616dustin Try to make the file name diabdat.mpq or spawn.mpq all uppercase! (DIABDAT.MPQ)

How would that help if there is no Devilution folder?

bubio:
@616dustin

I tried with the master repository, but there is no folder.
Try Method 2 or Method 3 in the following document.

https://github.com/diasurgical/devilutionX/blob/master/docs/installing.md

bubio:
@AJenbo 

If the Document folder in the app bundle is empty, files.app doesn't seem to see the folder.
At present, Method 1 cannot be used.
It may be good if the generation timing of diablo.ini can be advanced.

616dustin:
@bubio Method 2 worked for me. Oddly when i first launched it i received a warning that not all hellfire mpq's were found, even though i have the 4 hellfire mpqs. upon restarting the app everything seems to work great, thank you!

bubio:
@AJenbo 

If the Document folder is empty, I guess one solution would be to create an empty hidden file.


```objc

char *IOSGetPrefPath()
{
[...]

PR #5801, opened 2023-02-14
Strikeout iOS installation method 1

DakkJaniels:
Based on #5796 method 1 may not work.

### DIFF
--- a/docs/installing.md
+++ b/docs/installing.md
@@ -77,17 +77,17 @@ Once the App is installed, launch it once. It will say that it cannot find the d
 
 There are 3 way to get install the MPQs. Make sure to have installed the app and have launched it once before attempting to copy over the MPQ files.
 
-Method 1: Using the files.app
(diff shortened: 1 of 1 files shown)


DakkJaniels:
It looks like maybe file.app doesn't see the folder if it is empty. I only crossed it out for now so people know that this isn't necessarily going to work, but it can be restored easily enough if something changes in the future. I don't know if/when this will be fixed, just thought it would be a good idea to update the install instructions so people are aware of something that doesn't currently work. If the fix is imminent, then this is not required.

AJenbo:
I was also hesitant, but it will probably be long enough, and will need a release, and we can just revert this at that point.

jorycz:
iOS installation method 1 Is working. Just tried and I’m playing already. iPadOS 17.6.1.
I renamed file to diabdat.mpq (lowecase first d) and then kill app and open it again.

PR #5802, opened 2023-02-14
Stash Gold Limit

ikonomov:
Prevents overflow of the stash gold that happens around 2B gold.  The solution was provided during a discussion on discord.  Currently this change causes the game to crash when accessing the stash in a debug build as outlined here https://github.com/diasurgical/devilutionX/issues/5474.  I was able to test building a release version as well as implementing https://github.com/diasurgical/devilutionX/pull/5475 in debug (and setting a lower limit for the test).  Both placing gold above the limit as well as Ctrl + clicking gold to move it caused no issues and the game behaved as expected, limiting placing gold above the limit.

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -96,6 +96,8 @@ void CheckStashPaste(Point cursorPosition)
 	}
 
 	if (player.HoldItem._itype == ItemType::Gold) {
+		if (Stash.gold > std::numeric_limits<int>::max() - player.HoldItem._ivalue)
+			return;
 		Stash.gold += player.HoldItem._ivalue;
 		player.HoldItem.clear();
 		PlaySFX(IS_GOLD);
@@ -668,6 +670,8 @@ void GoldWithdrawNewText(string_view text)
 bool AutoPlaceItemInStash(Player &player, const Item &item, bool persistItem)
 {
 	if (item._itype == ItemType::Gold) {
+		if (Stash.gold > std::numeric_limits<int>::max() - item._ivalue)
+			return false;
 		if (persistItem) {
 			Stash.gold += item._ivalue;
 			Stash.dirty = true;


DakkJaniels:
Shouldn't you put stash gold to max before returning?

StephenCWills:
> Shouldn't you put stash gold to max before returning? Or allow up to the limit and keep the rest of the gold in hand?

I told him it wasn't necessary, because no one is going to hit this limit anyway. If you want, you can do that and test it and submit the change yourself.

ikonomov:
> Shouldn't you put stash gold to max before returning? Or allow up to the limit and keep the rest of the gold in hand?

Sure, except I didn't know how to code this complication.  With the current 2B limit and only 5000 gold pile size I thought the point of this PR is mostly theoretical.

Seems like Stephen has beat me to it.  Thank you.

AJenbo:
So this PR fixes an edge case that will never happen, but introduces another edge case that will never happen, but triggers an edge case that often happens?

ikonomov:
> So this PR fixes an edge case that will never happen, but introduces another edge case that will never happen, but triggers an edge case that often happens?

If possible, let's keep it open until somebody can fix the introductions.

ikonomov:
> So this PR fixes an edge case that will never happen, but introduces another edge case that will never happen, but triggers an edge case that often happens?

And no, the edge case never happening is an incorrect assessment.  Dan is already at above 100M since the stash was introduced.  It will happen for some people, the only question is not if, but when.

StephenCWills:
> And no, the edge case never happening is an incorrect assessment. Dan is already at above 100M since the stash was introduced. It will happen for some people, the only question is not if, but when.

Stash was introduced March last year, so you'll have to give him another 20 years at the current pace.

DakkJaniels:
> I told him it wasn't necessary, because no one is going to hit this limit anyway. If you want, you can do that and test it and submit the change yourself.

I was going to, but I can't access my stash to test 🤣

ikonomov:
When we discussed this edit in discord and I tested it and discovered the trigger I think nobody had the impression that this should be merged.  In fact I said right then that I should wait with this PR until the trigger problem is resolved.  A few days passed and after another discussion I had with Stephen where he said it's better to submit PRs and then just allow the reviews to get to them at their own pace I decided it's better to just submit it now so it doesn't get lost and maybe somebody can look at this in the future (or distant future) whenever time permits.  I'm sorry, but I do not see the point of the bad criticism or the laughing alongside it.

DakkJaniels:
Not sure what your comment is about, I was laughing because I was just mentioning in Discord that I'm getting crashes when I try and access stash. I would look into doing what I suggested, but I can't actually test anything because I can't access my stash (like I said in the comment).

ikonomov:
> Not sure what your comment is about, I was laughing because I was just mentioning in Discord that I'm getting crashes when I try and access stash. I would look into doing what I suggested, but I can't actually test anything because I can't access my stash (like I said in the comment).

Ah, sorry then about my laughing comment. :(  I really thought you were trying to sing along to AJenbo's tune.  You can test your edit in a release build, or you could implement https://github.com/diasurgical/devilutionX/pull/5475 like I did, and test with that.  I thought about doing what you suggested, but there is quite a bit of complexity (that's beyond my abilities) when holding the gold so I gave up on the idea.

AJenbo:
> Not sure what your comment is about, I was laughing because I was just mentioning in Discord that I'm getting crashes when I try and access stash. I would look into doing what I suggested, but I can't actually test anything because I can't access my stash (like I said in the comment).

try creating a new one with 1.4.1? ... or maybe just look at the bug that is causing you to not being able to access it now.

AJenbo (approved):

PR #5803, opened 2023-02-14
revert Fine to Sturdy change

ikonomov:
Reverts back the change from https://github.com/diasurgical/devilutionX/commit/99a37a41130f09b95259133179a680f1dd84db9e in order to avoid duplicate affix names on the same item (Sturdy *item* of Sturdiness).


AJenbo:
it should be done with an actual revert

ikonomov:
> it should be done with an actual revert

I can't do "an actual revert".  I did what I can.

AJenbo (approved):

galaxyhaxz:
I agree actually with this. Though instead I kept sturdy and changed it to "rigidness". They kept Sturdy in Diablo 2 but got rid of the durability suffixes and used Craftsmanship for damage, instead.

AJenbo:
Fit's nicely since D1 players are very "rigidness"

ikonomov:
> 
I renamed Sturdy in my mod to Stout since I saw that they've used it in D2 for defense.  I'm happy we have an agreement.  I think it's ok having two prefixes with the same name, since they don't appear on the same items.  Lightning is another example, used for a prefix and suffix.

galaxyhaxz:
> Lightning is another example, used for a prefix and suffix.

Yep, I agree. Diablo 2 had a few quirks where you get something like "Berserker" as a prefix, then you get charges of the *skill* Berserk as a suffix, resulting in "Berserker's Item of Berserk". Cases like these should be reduced/eliminated as much as possible.

galaxyhaxz:
> Lightning is another example, used for a prefix and suffix.

Also I overlooked this completely. IMO, this is bad as a prefix name. You don't say "Lightning Sword of Shock" instead "Shocking Sword of Lightning" sounds better. In general the grammar choices need to be updated, and some words have to be changed, inevitably.

PR #5811, opened 2023-02-16
[Android] Select the external files directory judiciously

StephenCWills:
On Android, this enables the game to use secondary external storage if the user moves all their data/config/save files to it.

### DIFF
--- a/android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
+++ b/android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
@@ -1,13 +1,9 @@
 package org.diasurgical.devilutionx;
 
-import android.Manifest;
-import android.annotation.SuppressLint;
 import android.content.Intent;
-import android.content.pm.PackageManager;
 import android.graphics.Rect;
 import android.os.Build;
 import android.os.Bundle;
-import android.os.Environment;
 import android.util.Log;
 import android.view.SurfaceHolder;
 import android.view.SurfaceView;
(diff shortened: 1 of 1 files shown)


AJenbo (commented):
One question though. What happens if someone tried the SD first, then later got it working with the internal one. Will the game discover that original file and prefer the SD now? Wouldn't that result in them not having asses to there existing saves?

Maybe a bit of an edge case, but there is about 20k "testers" :cold_sweat:

AJenbo (approved):

StephenCWills:
Uhh, yeah. If there are files in both locations, it will prefer the last one, which is probably going to be the SD card. We can (and probably should) break out of the loop early so that it prefers the first one it finds, which is actually what I meant to do given how the commit was worded. 😓

AJenbo:
Maybe we should mirror the logic from the C++ code where we prefer which ever folder has a diablo.ini?

StephenCWills:
> Maybe we should mirror the logic from the C++ code where we prefer which ever folder has a diablo.ini?

I guess we could do that, but how are Android users going to create that INI file? There's no native support for creating text files in the file manager. I figured this approach would be a bit better because the existence of any files in these folders represents a direct effort by the user to put one there. Plus, you only have to put the MPQ in your preferred storage location and you're off to the races.

However, I suppose we might have plans to produce the INI file sooner for iOS users which could throw a wrench in my assumptions. In that case, I'm not sure what would be the best approach.

AJenbo:
I'm thinking we could prioritize folders this way:
- Has a diablo.ini
- Has any file
- What we think is the SD card

StephenCWills:
Okay, I'm pretty sure this does what you suggested. Crossing my fingers that the lambda compiles. It did on my system.

AJenbo (approved):
Nice work :)

PR #5812, opened 2023-02-17
[Android] Use correct external files dir in all activities

StephenCWills:
It seems that the various activities weren't sharing their `externalDir` state. This extracts the logic for managing files in external storage to the `ExternalFilesManager` class so that all activities can share the same logic.

### DIFF
--- a/android-project/app/src/main/java/org/diasurgical/devilutionx/ExternalFilesManager.java
+++ b/android-project/app/src/main/java/org/diasurgical/devilutionx/ExternalFilesManager.java
@@ -1,0 +1,103 @@
+package org.diasurgical.devilutionx;
+
+import android.content.Context;
+import android.os.Build;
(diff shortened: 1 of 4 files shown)


android-project/app/src/main/java/org/diasurgical/devilutionx/ExternalFilesManager.java
```diff
@@ -0,0 +1,103 @@
...
+		} catch (IOException exception) {
+			Log.e("copyFile", Objects.requireNonNull(exception.getMessage()));
+			if (dst.exists()) {
+				//noinspection ResultOfMethodCallIgnored
+				dst.delete();
+			}
+			return false;
+		}
+
+		return  true;
```
AJenbo, line 101:
```suggestion
		return true;
```

PR #5813, opened 2023-02-17
Fix OOB access when accessing stash

### LINKED ISSUE #5474: Stash vector out of bounds from spurious loop?

StephenCWills:
This resolves #5474

### DIFF
--- a/Source/qol/stash.cpp
+++ b/Source/qol/stash.cpp
@@ -355,10 +355,11 @@ void DrawStash(const Surface &out)
 
 	for (auto slot : StashGridRange) {
 		StashStruct::StashCell itemId = Stash.GetItemIdAtPosition(slot);
-		Item &item = Stash.stashList[itemId];
-		if (Stash.IsItemAtPosition(slot)) {
-			InvDrawSlotBack(out, GetStashSlotCoord(slot) + offset, InventorySlotSizeInPixels, item._iMagical);
+		if (itemId == StashStruct::EmptyCell) {
+			continue; // No item in the given slot
 		}
+		Item &item = Stash.stashList[itemId];
+		InvDrawSlotBack(out, GetStashSlotCoord(slot) + offset, InventorySlotSizeInPixels, item._iMagical);
 	}
 
 	for (auto slot : StashGridRange) {


AJenbo (approved):
Thanks you <3

ikonomov:
Thank you! :)

PR #5238, opened 2022-08-13
Build for PS2

AJenbo:
```bash
docker run -it -w /project -v $(pwd):/project ps2dev/ps2dev sh

apk add git make cmake gmp mpc1 mpfr

git config --global --add safe.directory /project

cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=/usr/local/ps2dev/ps2sdk/ps2dev.cmake

cmake --build build -j $(nproc)
```

if you need to build the SDK (don't forget to clone and build ps2_drivers) you will also need this:
```bash
apk add musl-dev gcc bash
```

image

### todo
- [x] `ReadOnlyTest()` is creating a folder instead of a file, and fails to delete it, failing next lauch
- [x] Resolution is 640x448 ~~instead of 640x480~~ which is the max PS2 can do 
- https://github.com/ps2dev/ps2sdk/pull/322
- https://github.com/ps2dev/ps2sdk/pull/397
- https://github.com/ps2dev/ps2sdk/pull/398
- https://github.com/ps2dev/ps2sdk/pull/404
- https://github.com/ps2dev/ps2sdk/pull/405
- https://github.com/ps2dev/ps2sdk/pull/406
- https://github.com/ps2dev/ps2sdk/pull/407
- https://github.com/ps2dev/ps2sdk/pull/408
- https://github.com/ps2dev/ps2sdk/pull/410
- [ ] Remove unused player sounds to get below the 2MB audio limit
  - [ ] https://github.com/diasurgical/devilutionX/pull/5820
- [ ] No audio on videos (works via audiolib, maybe just enable that again)
- [ ] Writing saves fails on hardware
- [ ] Wrong sounds are being played in various cases on hardware and there are clicks and pops

### Side note
Initially we tried using SDL1.2, but the version provided by ps2sdk is 1.2.8 and we expect 1.2.10. There is also an issue with there port not correctly implementing Uint64. Solving that wasn't to big of an issue, but the game gets stuck during SDL_JoystickInit, so instead we for now have opted for SDL2 which went much smoother.

Issue #3532, opened 2021-11-17
line played in English with Polish sound pack on


twojstaryzdomu:
**Important information**
Android 11 / 1.3.0 / github release

**Describe**
The line is played in English with the inventory full ('I have no room') instead of a Polish equivalent. No idea whether a translation is present in the Polish sound pack for that line though. So far this has been the only line played in English with the Polish sound pack on.

**To Reproduce**
1. Run Hellfire with the Polish sound pack, pick single user, play Warrior. 
2. Keep adding items to the inventory until full.
3. The line played in English ('I have no room') indicates the problem.

**Expected behavior**
Expected to hear a Polish equivalent from the sound pack.

qndel:
That's because they didn't translate all sounds so whenever we tried to add some flavor with one of the unused ones, you'll get results like these.

twojstaryzdomu:
There are other voice lines that still sound out in Polish for the said full inventory.
Can you disable that one line so that it adds no foreign flavour to an otherwise flawless sound pack?

qndel:
> There are other voice lines that still sound out in Polish for the said full inventory. Can you disable that one line so that it adds no foreign flavour to an otherwise flawless sound pack? Or at least add an option to blanket deny foreign sounds consistently.

Not really, that's why I recommended them to translate all sounds 🤷

twojstaryzdomu:
> > There are other voice lines that still sound out in Polish for the said full inventory. Can you disable that one line so that it adds no foreign flavour to an otherwise flawless sound pack? Or at least add an option to blanket deny foreign sounds consistently.
> 
> Not really, that's why I recommended them to translate all sounds 🤷

That would be ideal but until that happens the game will feature those occassional annoyances.

qndel:
"until that happens" - it won't happen, they aren't working on the dubbing anymore and consider it finished

twojstaryzdomu:
That's a shame, that makes it one more argument to allow excluding the few English lines that didn't make it via an ini param.

qndel:
> That's a shame, that makes it one more argument to allow excluding the few English lines that didn't make it via an ini param.

As I've said, it's not something we will be dealing with from our side, you can disable the sound from playing by replacing it with an empty one. 😛

StephenCWills:
FYI, here's a silent audio file. You should be able to use this to override the untranslated voice lines.
https://github.com/diasurgical/devilutionX/blob/master/Packaging/ctr/audio_silent.wav

[...]

PR #5820, opened 2023-02-19
Remove unused Warrior sounds

glebm:
List taken from the PS2 PR (https://github.com/diasurgical/devilutionX/pull/5238) and these 2 files:

1. sfx/warrior/warior96.wav
2. sfx/warrior/warior97.wav

PR for devilutionx-mpq-tools: https://github.com/diasurgical/devilutionx-mpq-tools/pull/9


AJenbo:
The same sounds can be removed from all heroes, I didn't get around to making a PR for it since I didn't check to make sure that they where all fully unused. But I guess since you removed them that takes care of validating that they are unused :)

glebm:
> But I guess since you removed them that takes care of validating that they are unused

I didn't actually check anything, I simply copied the list from the PR!

FitzRoyX:
I feel like this is one of many reasons why portathon to outdated hardware is problematic. Any modder that wants to use unused assets now has to laboriously add them back in, all because the PS2 has some ridiculously low ram limit? How many people have a ps2 but not a pc?

DakkJaniels:
> Any modder that wants to use unused assets now has to laboriously add them back in, all because the PS2 has some ridiculously low ram limit?

I kind of agree - can't we use a preprocessor directives for a low ram build, or just comment out the lines so they can be added back in easily enough?

glebm:
We already removed a ton of unused stuff from the codebase, ifdefs here alone would not be enough.
Eventually we'd probably want to move this from the code into configuration files anyway.

AJenbo:
yeah having this cleaned out would make it easier to solve issues like https://github.com/diasurgical/devilutionX/issues/3532 and making things data driven which should make modding a lot easier all in all.

DakkJaniels:
Fair enough - I guess if everything is being pulled at some point and we've already pulled a lot of this stuff, no sense in keeping this around now.

FitzRoyX:
> yeah having this cleaned out would make it easier to solve issues like #3532 and making things data driven which should make modding a lot easier all in all.

So why did we add dialog for the unused quest wavs then such as this https://github.com/diasurgical/devilutionX/pull/1092 if we're now removing the wavs themselves?

#3532 doesn't even look like a real issue, it just looks like a mod issue. They could've easily done all the wavs.

AJenbo:
Ok , feel free to record the missing lines

StephenCWills:
> So why did we add dialog for the unused quest wavs then such as this #1092 if we're now removing the wavs themselves?

Perhaps if the lines being removed here were providing any real value at all, you might have a point. It's all just `PS_WARR#` enum values and the corresponding array entries that reference the `warior#.wav` files. We could comment them like the speech entries in the PR you referenced, but I fail to see how that would be useful in this case.

At best, you can claim that this PR would be eliminating the mapping that enables you to determine which wav file each `HeroSpeech` enum value can be found in. However, if you look a bit closer, you'll notice that `herosounds` just lays out the sound files in ascending order by number so it's still not hard to reconstruct it.

It's a bit annoying for modders that have to add these back in, but it's not a huge amount of work. As has been already mentioned, we have future plans for making things easier on modders anyway.

glebm:
Retargeted to development

PR #5823, opened 2023-02-19
Multiplayer: Enable Zhar the Mad Quest

obligaron:
Contributes to #926

Monster `talkmsg` and `flags` is not synced in multiplayer.
In multiplayer `_qvar1` is used to sync Zhar's state.

Remaining quests:
- Warlord of Blood
- Lachdanan

### DIFF
--- a/Source/quests.cpp
+++ b/Source/quests.cpp
@@ -430,6 +430,7 @@ void CheckQuestKill(const Monster &monster, bool sendmsg)
 		myPlayer.Say(HeroSpeech::ImNotImpressed, 30);
 	} else if (monster.uniqueType == UniqueMonsterType::Zhar) { //"Zhar the Mad"
 		Quests[Q_ZHAR]._qactive = QUEST_DONE;
+		NetSendCmdQuest(true, Quests[Q_ZHAR]);
 		myPlayer.Say(HeroSpeech::ImSorryDidIBreakYourConcentration, 30);
 	} else if (monster.uniqueType == UniqueMonsterType::Lazarus) { //"Arch-Bishop Lazarus"
 		auto &betrayerQuest = Quests[Q_BETRAYER];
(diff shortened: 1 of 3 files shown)


Source/quests.h
```diff
@@ -43,6 +43,14 @@ enum {
 	QS_GHARBAD_ATTACKING,
 };
 
+/** @brief States of the zhar the ma the week quest for multiplayer sync */
```
qndel, line 46:
```suggestion
/** @brief States of Zhar the Mad quest for multiplayer sync */
```

obligaron: Thanks 🙂

Source/quests.cpp
```diff
@@ -752,6 +753,27 @@ void ResyncQuests()
...
+				zhar->goal = MonsterGoal::Talking;
+				break;
+			case QS_ZHAR_ANGRY:
+				zhar->talkMsg = TEXT_ZHAR2;
+				zhar->goal = MonsterGoal::Inquiring;
+				break;
+			case QS_ZHAR_ATTACKING:
+				zhar->talkMsg = TEXT_NONE;
+				zhar->goal = MonsterGoal::Normal;
+				zhar->activeForTicks = UINT8_MAX;
```
glebm, line 772: Use std::numeric_limits?

obligaron: I simply copied it from the [ZharAI](https://github.com/diasurgical/devilutionX/blob/31b0f2147a3939feefa8eec0f97e47e1c3ca6081/Source/monster.cpp#L2717), cause it does reproduce (sync) the same state.
For consistency I wouldn't change it now and we should do a cleanup-pr that replaces all occurrences of `UINT8_MAX` with `std::numeric_limits`. What do you think? 🙂

AJenbo (approved):
your flying through these :)

PR #5828, opened 2023-02-22
updated gameshell buildscripts

danie1kr:
recent changes made it necessary to add new packages to be installed and build options to be adapted


AJenbo (approved):
Thanks :)

PR #5830, opened 2023-02-24
Remove references to Bard-specific speech IDs

### LINKED ISSUE #5829: Valor quest log plays King Leoric script
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

In single player mode, I received the Valor quest in the Catacombs. Clicking the Book of Blood played the quest activation dialog as expected, "...And so, locked beyond the Gateway of Blood and past the Hall of Fire, Valor awaits for the Hero of Light to awaken..."

A few minutes later, I clicked on Valor in my quest log, and rather than replaying the Valor quest dialog, it instead played the following, "The dead who walk among the living follow the cursed King. He holds the power to raise yet more warriors for an ever-growing army of the Undead. If you do not stop his reign, he will surely march across this land and slay all who still live here." The Fandom wiki tells me that's Adria's comment on The Curse of King Leoric.

### To Reproduce

1. Activate the Valor quest (single player Catacombs level 5)
2. Open the quest log (Q)
3. Click on Valor within the quest to replay the quest dialog

### Expected Behavior

I expected to see and hear the quest dialog for the Valor quest

### Additional context

_No response_

StephenCWills:
From what I can tell, the only changes the Hellfire devs made to these speeches was to double the spaces between sentences in the text. As far as backwards compatibility is concerned, it just means the vanilla Hellfire client might play the Rogue's speech ID instead of the Bard's.

The only reason I can think of to prefer another solution would be if anyone considers it important for us to attempt to correct speech IDs from saves made by prior DevX versions. I don't consider that particularly important so I prefer this solution.

This resolves #5829


obligaron:
> From what I can tell, the only changes the Hellfire devs made to these speeches was to double the spaces between sentences in the text.

Should we remove this extra space?
This would make the work for translators more easy, cause the text would be the same and they only need to translate one of them and not two. 🤔

StephenCWills:
I have no objections.

AJenbo (commented):
Wont this cause the Polish dub to now use the Rouge speech for the Bard?

StephenCWills:
> Wont this cause the Polish dub to now use the Rouge speech for the Bard?

Does the `PS_ROGUE1` identifier somehow resolve to something different when applied to the Bard's speech IDs?

StephenCWills:
I just checked the Polish dub, and these texts are recited in English. They didn't provide translations for `rogue01.wav`, `rogue10.wav`, `rogue11.wav`, or `rogue12.wav`.

Also, all the Bard' voice lines play the `rogue##.wav` sfx. As far as I can tell, there's no logic to play the Bard's voice lines even though they were added to `pl.mpq`. Am I doing something wrong?

AJenbo:
> I just checked the Polish dub, and these texts are recited in English. They didn't provide translations for rogue01.wav, rogue10.wav, rogue11.wav, or rogue12.wav.

Kinda shocked that they aren't there. @qndel do you know if I fucked up and left them out or something?

> Does the PS_ROGUE1 identifier somehow resolve to something different when applied to the Bard's speech IDs?

Are forgot there was an extra step to how this works. Honestly a much better way to handle the speeches would for them to be generic for the hero and have InitQTextMsg use `herosounds[]` if it's a line spoken by the hero. Of cause that is fare beyond the scope of this PR/issue.

AJenbo (approved):

qndel:
@AJenbo are we talking specifically about rogue1/10/11/12.wav? what texts are these? Iirc they didn't translate unused stuff

AJenbo:
@qndel yes, Valor quest for one

StephenCWills:
> are we talking specifically about rogue1/10/11/12.wav? what texts are these? Iirc they didn't translate unused stuff

These are for "book levers". Chamber of Bone, Arkaine's Valor, Halls of the Blind, and Warlord of Blood in that order.

AJenbo:
Maybe they only did the multiplayer portion of the game :thinking:

StephenCWills:
Hm, seems like they did some Single Player content. They translated `bmaid04.wav`, which is for Poisoned Water Supply.

PR #5831, opened 2023-02-24
Multiplayer: Enable Lachdanan Quest

obligaron:
Contributes to #926

Monster `talkmsg` and `goal` is not synced in multiplayer.
In multiplayer `_qvar2` is used to sync Lachdanan's state.
I changed where the unique is spawned (from `MonsterTalk` to `TalktoMonster`), so I can check that only one player (that talks to lachdanan) spawns the Veil of Steel.

Remaining quest:
- None 🚀🎉

... but many bugs. 😁 

Next step after this pr is adding a toggle for switching between classic multiplayer quests and singleplayer quests in multiplayer.

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -2309,8 +2309,12 @@ void InitItems()
 		if (sgGameInitInfo.bCowQuest != 0 && currlevel == 19)
 			SpawnQuestItem(IDI_GREYSUIT, { 25, 25 }, 3, 1, false);
 		// In multiplayer items spawn during level generation to avoid desyncs
-		if (gbIsMultiplayer && Quests[Q_MUSHROOM].IsAvailable())
-			SpawnQuestItem(IDI_FUNGALTM, { 0, 0 }, 5, 1, false);
+		if (gbIsMultiplayer) {
+			if (Quests[Q_MUSHROOM].IsAvailable())
+				SpawnQuestItem(IDI_FUNGALTM, { 0, 0 }, 5, 1, false);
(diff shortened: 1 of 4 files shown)


AJenbo (approved):
I like that you added enums for the quest, we should probably clean up the other ones in the same way as well.

Chance4us:
Good day dear friends, before too much work goes into it upfront, I'd like to throw in my guess that it probably makes sense to provide users with an enable option for each individual quest. :question:

Chance4us:
Well, it seems I was terribly mistaken.
Nevertheless, I would like to thank @obligaron for his great work.

AJenbo:
I get the appeal but i would like to keep it closer to the spirit of the original where random quest was part of the game design, and also this would keep things a bit simpler in the ui and code :)

AJenbo:
yes, they where actually fixed in 1.4.1 i think, since Hellfire never had a multiplayer mode they never differentiated the two

Chance4us:
@AJenbo Thanks and I have one more question related to multiplayer games.
What do you think of the idea of implementing the functionality to be able to save a multiplayer game,
which could be continued later by the creator of it?

I would find it interesting to be able to continue a partially completed multiplayer game in such a way
that other players could join or not have to continue the game if they wished.

Chance4us:
And a second idea that I don't want to withhold and that is the possibility of an AI based multiplayer co-player.
An AI based co-player would make testing of the multiplayer mode much easier.
At least a rudimentary AI would be useful insofar as he can at least walk alongside.

AJenbo:
yes, both are in the future plans.

You can checkout https://www.twitch.tv/nitekat for what is sorta possible in terms of AI

PR #5838, opened 2023-02-26
Multiplayer: Prevent trapped doors to fire twice

### LINKED ISSUE #5184: Trapped doors reset if you close the door then re-enter the level in multiplayer
### Operating System

Windows x64

### DevilutionX version

Other (please specify version number)

### Describe

`UpdateTrapState` returns early if it uses a door as a trigger and the door is closed. This means that in a multiplayer game a player can set off a trap twice by leaving the level and replaying the delta

### To Reproduce

Start or join a multiplayer game
find a level with a trapped door
open the door, close it, and observe the door is "Closed door" and not "Trapped door" (if playing as a rogue)
Leave the level then re-enter
Observe the door label is "Trapped door" (if playing as a rogue)
Open the door, note the trap fires again

### Expected Behavior

Traps on doors should only trigger once

### Additional context

I suspect this can happen in vanilla but haven't found a trapped door yet to be able to test.

obligaron:
Fixes #5184

Now `CMD_DOORCLOSE` is stored in delta messages.
When we found this message we set `_oTrapFlag`.
And when `_oTrapFlag` is not set the trap will not activate.

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -700,10 +700,7 @@ void DeltaSyncObject(WorldTilePosition position, _cmd_id bCmd, const Player &pla
 
 	sgbDeltaChanged = true;
 	auto &objectDeltas = GetDeltaLevel(player).object;
-	if (bCmd == _cmd_id::CMD_CLOSEDOOR)
-		objectDeltas.erase(position);
-	else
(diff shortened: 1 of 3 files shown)


AJenbo (approved):
nice fix

PR #5849, opened 2023-03-01
Disable corner stone of the world in multiplayer

### LINKED ISSUE #4193: Disable corner stone of the world in multiplayer
I don't think there is any way to make this mechanic work in multiplayer.

- What happens if both players have an item on the corner stone?
- Should players be able to take others players items?
- The data is shared between multiplayer and singleplayer which will cause the item to morph.

obligaron:
Fixes #4193

First commit adds helper function for all places that check for the corner stone.
Second commit removes it for multiplayer.


AJenbo (approved):
considering all the work for adding quests its funny to see you remove one :)

PR #5853, opened 2023-03-03
Ignore build workflows if only change is markdown

DakkJaniels:
No need to rebuild everything if the only change is to documentation.


AJenbo (approved):
sweet!

PR #5860, opened 2023-03-04
MP: Send delta message for lachdanans death

### LINKED ISSUE #5858: MP: Lachdanan reviving after coming back from town
### Operating System

Windows x64

### DevilutionX version

### Describe

After completing the quest tied to him, TPing out of the level and coming back, Lachdanan will be displayed as if you had never completed the quest and will give you the very same item he was supposed to drop (not the veil of steel, the other one), this can be reproduced indefinitely.
For reference, I have randomized quests turned off and of course quests in multiplayer turned on.

### To Reproduce

1. Complete Lachdanan's quest in MP
2. TP out to town
3. Go back
4. Witness Lachdanan being there again
5. Talk to him, he'll give you the dialogue right before he dies
6. It drops the very same magical item he gave you before

### Expected Behavior

Lachdanan should have stayed dead after completing the quest

### Additional context

All encountered on commit [a629aa9](https://github.com/diasurgical/devilutionX/commit/a629aa91535483f12280f599b8134229d53a6336)

obligaron:
Fixes #5858

Let Lachdanan rest in peace🙏

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -2885,8 +2885,10 @@ void LachdananAi(Monster &monster)
 			if (!effect_is_playing(USFX_LACH3) && monster.goal == MonsterGoal::Talking) {
 				monster.talkMsg = TEXT_NONE;
 				Quests[Q_VEIL]._qactive = QUEST_DONE;
-				MonsterDeath(monster, monster.direction, true);
 				NetSendCmdQuest(true, Quests[Q_VEIL]);
+				MonsterDeath(monster, monster.direction, true);
+				delta_kill_monster(monster, monster.position.tile, *MyPlayer);
+				NetSendCmdLocParam1(false, CMD_MONSTDEATH, monster.position.tile, monster.getId());
 			}
 		}
 	}


AJenbo (approved):
Makes sens, kinda interesting though :)

obligaron:
Initially I thought `MonsterDeath` would do the job... but I was wrong. 😛

glebm:
Perhaps it should? It's kinda odd that it sends messages for loot spawning etc but not for the death itself

obligaron:
A refactoring would be good. 👍 
But it's a rabit hole. You start by one function and come to the next.
So I think this is outside of this issue. Perhaps we can should open a issue for it?

glebm:
Yeah, it's out of scope for the fix, a separate issue would make sense

AJenbo:
Yeah it's a bit of an issue since it's also what is called when receiving `CMD_MONSTDEATH` so that has to be taken in to account to avoid an infinit loop :)

It might actually make sens to find a general way to solve this for all network vs local events.

glebm:
> so that has to be taken in to account to avoid an infinit loop :)

That's why it already has a `sendmsg` argument from what I can tell.

PR #5861, opened 2023-03-05
Arkaine's Valor: Don't check player inventory when processing CMD_OPERATEOBJ

### LINKED ISSUE #5855: MP: Valor, only the player opening the chambers will see them open

When placing the blood stones only the player who activates it will see the chambers opening, other players will just see the blood stone dropping near where the chamber should be and will see the player an monsters walking through walls.

StephenCWills:
The player who operates the pedestal will invoke `OperatePedestal()` from `OperateObject()`. This will remove the Blood Stone from their inventory and send `CMD_OPERATEOBJ`, `CMD_SYNCQUEST`, and `CMD_SPAWNITEM` commands to remote clients.

Remote players who receive the `CMD_OPERATEOBJ` command for the pedestal will invoke `OperatePedestal()` from `SyncOpObject()` and blindly accept that the quest state has advanced. This will only play the sound effect and update the dungeon layout without sending any additional commands to remote clients.

The `CMD_OPERATEOBJ` command was moved up in `OperateObject()` so that it arrives to remote clients before `CMD_SPAWNITEM`. Otherwise, the Blood Stones and Arkaine's Valor will spawn before the dungeon layout changes and they'll end up dropping outside the rooms they were meant to drop in.

This resolves #5855

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -2266,8 +2266,17 @@ void OperatePedestal(Player &player, Object &pedestal, bool sendmsg)
 		return;
 	}
 
-	if (pedestal._oVar6 == 3 || !RemoveInventoryItemById(player, IDI_BLDSTONE)) {
+	if (pedestal._oVar6 == 3 || (sendmsg && !RemoveInventoryItemById(player, IDI_BLDSTONE))) {
 		return;
+	}
+
+	if (sendmsg) {
+		NetSendCmdLoc(MyPlayerId, false, CMD_OPERATEOBJ, pedestal.position);
+		if (gbIsMultiplayer) {
+			// Store added stones to pedestal in qvar2, cause we get only one CMD_OPERATEOBJ from DeltaLoadLevel even if we add multiple stones
+			Quests[Q_BLOOD]._qvar2++;
+			NetSendCmdQuest(true, Quests[Q_BLOOD]);
+		}
 	}
 
 	pedestal._oAnimFrame++;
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -2266,10 +2266,19 @@ void OperatePedestal(Player &player, Object &pedestal, bool sendmsg)
 		return;
 	}
 
-	if (pedestal._oVar6 == 3 || !RemoveInventoryItemById(player, IDI_BLDSTONE)) {
+	if (pedestal._oVar6 == 3 || (sendmsg && !RemoveInventoryItemById(player, IDI_BLDSTONE))) {
```
obligaron, line 2269: `OperatePedestal` is called from two call-sites (`OperateObject` and `SyncOpObject`).
`OperateObject` is also called for remote clients in `CheckNewPath`.
In this case `sendmsg` is false but we still need to check if the item is in the inventory, cause it's not validated (or alternative do nothing in `OperateObject` for remote players).
So we must distinguish if `OperatePedestal` got called from `SyncOpObject` or `OperateObject`. That's why other functions got two parameters (`sendmsg` and `sendLootMsg`).
I think we need to do the same here.

This is another issue where multiplayer logic is complex. 😕 
In the long run we should avoid running `OperateObject` for remote players (through `CheckNewPath`) and should always rely on `SyncOpObject`. But this is far outside of the scope of this pr.

StephenCWills: This change prevents `OperateObject()` from calling `OperatePedestal()` when `sendmsg = false`, and also prevents `SyncOpObject()` from calling `OperatePedestal()` if `sendmsg = true`. Only the local player will invoke it from `CheckNewPath() > OperateObject() > OperatePedestal()` and only remote players will invoke it via `OnOperateObject() > SyncOpObject() > OperatePedestal()`. If you try to handle it in `OperatePedestal()` with two boolean values, you'll end up with something that doesn't make much sense logically, like...

```c++
if (sendmsg != sendLootMsg) {
    return;
}
```

obligaron: Sorry I missed the `if (sendmsg)` check in `OperateObject`. 😢 
This goes in the way that `OperateObject` is not used for remote players that's a nice direction imho.

StephenCWills: > (or alternative do nothing in `OperateObject` for remote players)

To be clear, if the local player processes their own `CMD_OPERATEOBJ` commands, the whole quest would be finished immediately because he wouldn't be checking for Blood Stone. If remote players process the local player's `CheckNewPath()`, then the game could end up advancing the quest twice depending on the race condition I mentioned in the issue. IMO, your proposed alternative is the only reasonable approach here, and that's what this PR does.

obligaron (approved):

AJenbo (approved):

PR #5862, opened 2023-03-05
UK TL fixes

tsunamistate:
* TL new settings
* Fix mistakes I found during gameplay

Also, for some reason there are two entries for Warlord of Blood book narration, weird


AJenbo:
> Also, for some reason there are two entries for Warlord of Blood book narration, weird

We are actually working on cleaning that up atm :) It happens because some of the heros had a different number of spaces for the same texts.

PR #5863, opened 2023-03-05
Use herosounds for class-specific quest speech IDs

### LINKED ISSUE #5856: MP: Quest dialogs initiated from the quest menu will always read in the rogue's voice

During multiplayer when opening the quest panel and selecting a quest entry it will play the dialog in the voice of the rogue even when playing as a different hero.

The game was started by a barbarian

StephenCWills:
Use Warrior sfx IDs when defining `Speech::sfxnr` for class-specific sound effects, then use the appropriate sound effect from `herosounds` in `InitQTextMsg()`.

This resolves #5856


obligaron (approved):
lgtm 🙂
I think this should be compatible with vanilla.

While checking if all texts are identically for all classes, I found one small difference.
The text for `WarlordOfBloodLore` for warrior/rogue/sorcerer and monk/bard differ:
> Angels and **men** alike
 vs
> Angels and **man** alike
I think we should fix this (not necessary in this pr).

PR #5864, opened 2023-03-05
Sync staff charges after recharging

### LINKED ISSUE #5857: MP: Sorc's spells are often not seen by other players

Often when playing with a sorcerer other players won't see them casting magic but will just see monsters dropping dead from the effect. This is likely caused by the player not being synced to where casting the spell would be possible for them on the other end. 

This most often seamed to happen at the start of a level, but that may just have been user bias.

StephenCWills:
This resolves #5857


AJenbo (approved):
Nice work :)

PR #5865, opened 2023-03-06
Update Italian localization

M0Rf30:
As in title,
Regards


AJenbo (approved):
Thanks :)

P.s. around April there are probably going to be a few new lines in preperation for the next release.

PR #5866, opened 2023-03-08
Add screen reader support

AJenbo:
This PR add screen reader integration for Linux and Windows.

This is the first step to making the game accessible for blind gamers.

The reader is currently being fed just the main menu and in-game info panel. This enough that a blind person can load a save game and fight a few monsters in the dungeon.

Shortcomings (will be addressed in later PRs):
- Creating a new hero is hard for them since it does not announce when it expects the name input.
- Navigating town is hard since there are basically no hints on how to make your way from the starting position to the church.
- Settings- and in-game menus are not read.
- None of the side panels are read.
- It only reads when the highlighted text changes, not when context does, so multiple objects with the same name (barrels) can be to identify

For Linux `libspeechd` is used, this will read the text regardless if the user has enabled the screen reader. Ther for integration is not being biuld by default.

Recording of the current implementation (Linux using Orca): https://www.youtube.com/watch?v=SdSjSZzhNBQ

On Windows it uses https://github.com/sig-a11y/tolk

Things I have not done (and don't think I will):
- Setup the option for MSVC since I can't easily verify that it works proper
- Added any of the liberies to the vendored src script.
- Looked in to if this would work on Xbox One/Series
- Looked in to if this will work on macOS (libspeechd can be build for it afaik, but it's unclear if it works)


Source/diablo.cpp
```diff
@@ -136,6 +139,18 @@ extern void plrctrls_after_check_curs_move();
 extern void plrctrls_every_frame();
 extern void plrctrls_after_game_logic();
 
+SPDConnection *Speechd;
+std::string SpokenText;
+void SpeakText(const char* text)
+{
+	if (!SpokenText.empty() && SpokenText.compare(text) == 0)
```
glebm, line 146: `SpokenText == text`

Source/diablo.cpp
```diff
@@ -6,6 +6,9 @@
 #include <array>
 
 #include <fmt/format.h>
+#ifndef WIN32
```
glebm, line 9: We probably want something like `#ifdef DEVILUTIONX_USE_LIBSPEECHD`

glebm (approved):

PR #5873, opened 2023-03-12
Replace ReturnLevel/LevelType/LvlPosition

obligaron:
The pr removes the need of `ReturnLevel`, `ReturnLevelLevelType` and `ReturnLevelLvlPosition`.
They are now only used for backwards compatibility with vanilla.

The pr fixes three issues:
- Wrong return level and position in multiplayer, when
   - entering a portal to a quest/set map from another player
   - didn't enter the quest/set map before
   - try to return the original level (for example level 3 when in skeleton king's lair)
   - now you will return in town at position 0/0 or whatever was last in `ReturnLevel` and friends
- Wrong return level and position in single- and multiplayer, when
   - going to a quest map (for example skeleton king's lair)
   - making a portal in quest map
   - going to another quest map (for example chamber of bone)
   - going to town and entering portal
   - try to return the original level (for example level 3 when in skeleton king's lair)
   - now you will return to the previous quest map return level and not the current (for example level 6)
   - this should also happen in vanilla
- Crash when using portal to arenas from another player when you never entered the arena


Source/quests.cpp
```diff
@@ -337,8 +337,7 @@ void CheckQuests()
 	    && quest._qvar1 >= 2
 	    && (quest._qactive == QUEST_ACTIVE || quest._qactive == QUEST_DONE)
 	    && (quest._qvar2 == 0 || quest._qvar2 == 2)) {
-		// Move the quest trigger into world space, then spawn a portal at the same location
-		quest.position = quest.position.megaToWorld();
```
AJenbo, line 341: will this still work if the game was started on vanilla?

AJenbo:
nice, this is a cleanup i was thinking should be done even before knowing for sure that it was causing issues

Source/quests.cpp
obligaron: `quest.position` is initialized in `GenerateLevel` (in `drlg_l4.cpp`).
This logic runs every time a level is changed or loaded.
So in vanilla and in devilutionX this variable should be initialized every time.
Another evidence is that this transformation is only possible once but when returning from lazarus' lair it's executed every time.
That's why I came to the conclusion that it's safe to move `megaToWorld` to `GenerateLevel`. 🙂 
But I didn't test it vanilla.

obligaron:
> nice, this is a cleanup i was thinking should be done even before knowing for sure that it was causing issues

I felt the same. 🙂 
With quest map multiplayer support it couldn't hide it's strange behavior anymore 😁

PR #5874, opened 2023-03-12
Fix Flash hits north east direction

### LINKED ISSUE #5872: Flash not being able to damage players in the NE direction
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

In the picture bellow the 1'st Warrior(using the chain-mail) will not be able to hit the 2'nd Warrior(using the plate armor) while using Flash. We tested it and it seems Flash will damage your opponent in any direction except this one.

### To Reproduce

Position two characters like the picture
Cast Flash
The character position in the North-East direction won't get damage

### Expected Behavior

Flash should always deal damage to all direction

### Additional context

_No response_

obligaron:
Fixes #5872

Flash did not hit north east direction but instead hit north direction twice. 😉 

Was introduced with #5311

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -3385,7 +3385,7 @@ void ProcessFlashTop(Missile &missile)
 
 	constexpr Direction Offsets[] = {
 		Direction::North,
-		Direction::North,
+		Direction::NorthEast,
 		Direction::East
 	};
 	for (Direction offset : Offsets)


glebm:
`Direction::NorthWest,` is also in `ProcessFlashTop` 🤔

obligaron:
`Direction::NorthWest` is only used in `ProcessFlashBottom` or?
This pr adds `Direction::NorthEast` (not `Direction::NorthWest`) to `ProcessFlashTop`.

glebm:
Makes sense, thanks

AJenbo (approved):

PR #5879, opened 2023-03-12
Avoid dependency on fmt/chrono

glebm:
We only used it in one place and it doesn't work on Amiga GCC-6 with the latest libfmt master (https://github.com/fmtlib/fmt/issues/3344)

Also handles the case where `localtime` returns `nullptr`.

As a bonus, this reduces RG-99 binary size by 32 KiB (chrono formatting is a bit heavy it seems).

### DIFF
--- a/Source/capture.cpp
+++ b/Source/capture.cpp
@@ -5,8 +5,9 @@
  */
 #include <cstdint>
 #include <cstdio>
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Lgtm

PR #5883, opened 2023-03-13
GitHub CI: Remove git SHA from cache key

glebm:
We're hitting the GitHub cache limit because we create a new cache for every commit.

Reuse the same cache instead.

image

### DIFF
--- a/.github/workflows/Android.yml
+++ b/.github/workflows/Android.yml
@@ -35,11 +35,10 @@ jobs:
         cache: gradle
 
     - name: Cache CMake build folder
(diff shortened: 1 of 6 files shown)


AJenbo (commented):
any way we could limit it so they are only created by master?

glebm:
Nor sure but it should be OK to have it for each branch. The branches will load the master cache if it's missing. Doing it for every commit was a bit much though.

There is a way to force delete merged branch caches by setting up a workflow to do that, but hopefully won't be necessary.

PR #5884, opened 2023-03-14
installing.md update - 3DS install note 

DakkJaniels:
add note to 3DS install instructions regarding lowercase letters.


AJenbo (approved):
Also make a note to saying to read the notes

PR #5890, opened 2023-03-17
items.cpp: fix inconsistent localization for miscellanous items

lkk7:
* For some reason, in `items.cpp`, `"Select from spell book, then\ncast spell to read"` was at some point changed to `"Select from spell book, then\ncast to read"`, breaking most of the translations.
* Update translations to be a single string when expected


AJenbo:
looks like the issue was introduced here https://github.com/diasurgical/devilutionX/commit/cf33230293004be048916574b94bdf3b8afa467c

AJenbo (approved):
thanks

PR #5894, opened 2023-03-18
Demomode: Store/override gameplay options

glebm:
The advantage of having a specific list rather than storing the whole `diablo.ini` file is that non-gameplay options, such as floating numbers, can still be changed in the config and visually affect the demo.

### DIFF
--- a/Source/engine/demomode.cpp
+++ b/Source/engine/demomode.cpp
@@ -75,6 +75,34 @@ bool Timedemo = false;
 int RecordNumber = -1;
 bool CreateDemoReference = false;
 
+// These options affect gameplay and are stored in the demo file.
+struct {
+	uint8_t tickRate = 20;
+	bool runInTown = false;
+	bool theoQuest = false;
+	bool cowQuest = false;
+	bool autoGoldPickup = false;
+	bool autoElixirPickup = false;
+	bool autoOilPickup = false;
+	bool autoPickupInTown = false;
+	bool adriaRefillsMana = false;
+	bool autoEquipWeapons = false;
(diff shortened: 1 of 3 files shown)


glebm:
@AJenbo Let's merge this?

glebm:
Rebased

Source/engine/demomode.cpp
```diff
@@ -75,6 +75,36 @@ bool Timedemo = false;
 int RecordNumber = -1;
 bool CreateDemoReference = false;
 
+// These options affect gameplay and are stored in the demo file.
+struct {
+	uint8_t tickRate = 20;
```
AJenbo, line 80: isn't this more of a playback speed, we do allow changing it during playback

Source/engine/demomode.cpp
```diff
@@ -75,6 +75,36 @@ bool Timedemo = false;
...
 bool CreateDemoReference = false;
 
+// These options affect gameplay and are stored in the demo file.
+struct {
+	uint8_t tickRate = 20;
+	bool runInTown = false;
+	bool theoQuest = false;
+	bool cowQuest = false;
+	bool testBard = false;
+	bool testBarbarian = false;
```
AJenbo, line 85: these shoudn't affect it

Source/engine/demomode.cpp
glebm: Done

Source/engine/demomode.cpp
glebm: Done

AJenbo (approved):

PR #5896, opened 2023-03-18
GitHub CI: Fix cache strategy

glebm:
GitHub caches are immutable, so #5883 was wrong (caches were never updated).

Goes back to the previous caching strategy but with a cleanup workflow to delete caches for merged/closed PRs.

The cleanup code comes from the example in GitHub docs: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries


AJenbo (approved):
I had a feeling that was the case :) Thanks for looking in to this.

PR #5897, opened 2023-03-18
use `Point`s instead of `x, y`s in GetRndObjLoc (objects.cpp) and portal-related functions (portal.cpp)

lkk7:


Source/objects.cpp
```diff
@@ -1491,37 +1491,37 @@ void AddTorturedBody(Object &torturedBody)
...
 
 void AddMushPatch()
 {
-	int y;
-	int x;
-
 	if (ActiveObjectCount < MAXOBJECTS) {
 		int i = AvailableObjects[0];
-		GetRndObjLoc(5, &x, &y);
+		auto [x, y] = GetRndObjLoc(5);
```
AJenbo, line 1524: We usually keep the use of auto to a minimum (lines where the type is already declared, such as casts).

```suggestion
		Point [x, y] = GetRndObjLoc(5);
```

glebm: This requires C++17 and won't work on some of the compilers that we still support, such as GCC 6.

lkk7: Sorry, it escaped my mind that we support older compilers.

https://github.com/diasurgical/devilutionX/blob/master/docs/CONTRIBUTING.md might be a bit misleading then, saying that the oldest compiler used (GCC 6.5) supports most of C++17. Looks like most of the features are supported since GCC 7.

glebm: Yeah that should be adjusted. It supports all of C++11 and some C++17 libraries via `std::experimental` (`string_view`, `optional`, etc).
C++17 language features, such as destructuring assignments, fold expressions, lambda auto arguments etc, are not supported by GCC 6.5.

AJenbo: > Sorry, it escaped my mind that we support older compilers.

If you can allow CircleCI to run on your PR then it will check with the oldest relevant compiler so you don't have to think about it.

Source/objects.cpp
```diff
@@ -1491,41 +1491,41 @@ void AddTorturedBody(Object &torturedBody)
...
+		y = GenerateRnd(MAXDUNY);
 		bool failed = false;
 		for (int i = 0; i < randarea && !failed; i++) {
 			for (int j = 0; j < randarea && !failed; j++) {
-				failed = !RndLocOk(i + *xx, j + *yy);
+				failed = !RndLocOk(i + x, j + y);
 			}
 		}
 		if (!failed)
 			break;
```
AJenbo, line 1515: This should now be possible
```cpp
		for (int i = 0; i < randarea; i++) {
			for (int j = 0; j < randarea; j++) {
				if (!RndLocOk(i + x, j + y))
					return { x, y };
			}
		}
```

lkk7: Wouldn't that instantly return from the function on first failure?

AJenbo: Oh I made a mistake, I thought it was breaking out of all 3 loops when failed turns to true, but it's only tow of them the third is inverted. So yeah this probably isn't possible.

AJenbo (approved):

AJenbo:
Nice cleanup, thanks.

PR #5900, opened 2023-03-19
Create a new build script for RG350 devices.

TmDobbin:
The Issue:
The Packaging/OpenDingux/build.sh script for whatever reason simply does not work at this time. There's a few different errors it throws, including it trying to create a /ccache directory and failing as it doesn't have the required permissions. I've tried this script on my main machine (Arch Linux) and two separate Ubuntu installs, and it was unsuccessful every time.

My proposed solution:
A new build script! It's based on the workflow file for the git actions OpenDingux build. Currently I've only made it work with the RG350, and it's admittedly very simple. If the script looks good, I'd be happy to write one for the other targets. I think that my script could likely be merged with the existing build.sh script, as it seems to be just the downloading that the current script fails on.

I've tested this script on the 3 machines mentioned above and it went successfully every time, producing a working .opk file. Please let me know if there's anything further I can do!

(Edited a mistake about the functionality of the current script)


AJenbo:
@glebm would you be willing to take a look at this PR?

TmDobbin:
Just to be clear, the download *does* succeed, but it's the steps after it fails on (i.e. the file is downloaded, but the toolchain seems to fail to build.)
I could probably have a look and fix the current script, if you'd like.

TmDobbin:
Just for a bit more clarification on the actual error - it seems the file is downloaded to the home directory, then it attempts to download ccache and places it in root, for some reason. It doesn't have permissions to place it in root, so it fails. I can't for the life of me find where the code is to save the file to a directory, hence my adaptation of the workflow file here on github. I'll have a look later today at adapting my new script to work for the rest of the targets.

glebm:
> I'll have a look later today at adapting my new script to work for the rest of the targets.

Your script is missing a number of options vs the current one and also downloads a prebuilt toolchain rather than building it from scratch, which won't work on ARM etc. It is better to fix the existing script.

The script itself doesn't do anything with ccache -- buildroot does though. The fix may involve setting some environment variables if that's what causes it to create `/ccache`. (ref: https://github.com/OpenDingux/buildroot/blob/035527ac16b66dd3980627f94a64a78dc3fd0e1b/configs/od_rs90_defconfig#L12)

glebm:
I've sent #5902 with a fix. The issue was that the `shared-dl` directory was missing.

PR #5901, opened 2023-03-19
Multiplayer: Fix desync when teleporting

obligaron:
in master a desync can happen when using a spell that changes your position (teleport, phasing or warp).
When the the desyncs happens the remote client starts/sees a walking one tile to the old position (before teleport).
After that the remote client sees the player walk back to the correct position.

Example Video:

https://user-images.githubusercontent.com/25415264/226195322-13e4dae6-bca8-499b-806a-8a9cedfc5ba8.mp4

How does this happen?
- Background: In multiplayer the "player data" is frequently send to the remote client.
  - These "player data" include a "target"-position (current position or position where the player is moving to) along with hp and other stuff.
  - This data is used for a syncing/position logic: the remote client tries to walk the player to this position (`MakePlrPath`). This ensures that the player always stands on the correct tile. Normally the "target"-position is in sync with the calculated position on the remote client.
- When starting a teleport a `CMD_SPELLXY` is send with the desired coordinates. But the "target"-position isn't set to something new (the player didn't changed position yet and is not moving).
- When the teleport is finished on the client, A new "player data" with a new "target"-position is send to the remote client.
- But when this package is too slow, the remote client uses the old "target"-position for it's syncing/position logic and starts walking towards the old position (before the teleport).
- After this it's receives a new "player data" info and walks back to the correct/new tile.

Fix:
Don't send desired "target"-position when the player will change his position soon.

### DIFF
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -129,7 +129,11 @@ byte *ReceivePacket(TBuffer *pBuf, byte *body, size_t *size)
 void NetReceivePlayerData(TPkt *pkt)
 {
 	const Player &myPlayer = *MyPlayer;
-	const Point target = myPlayer.GetTargetPosition();
+	Point target = myPlayer.GetTargetPosition();
+	// Don't send desired target position when we will change our position soon.
+	// This prevents a desync where the remote client starts a walking to the old target position when the teleport is finished but the the new position isn't received yet.
+	if (myPlayer._pmode == PM_SPELL && IsAnyOf(myPlayer.executedSpell.spellId, SpellID::Teleport, SpellID::Phasing, SpellID::Warp))
+		target = {};
 
 	pkt->hdr.wCheck = HeaderCheckVal;
 	pkt->hdr.px = myPlayer.position.tile.x;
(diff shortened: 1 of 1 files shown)


StephenCWills (approved):

Source/multi.cpp
```diff
@@ -129,7 +129,11 @@ byte *ReceivePacket(TBuffer *pBuf, byte *body, size_t *size)
 void NetReceivePlayerData(TPkt *pkt)
 {
 	const Player &myPlayer = *MyPlayer;
-	const Point target = myPlayer.GetTargetPosition();
+	Point target = myPlayer.GetTargetPosition();
+	// Don't send desired target position when we will change our position soon.
+	// This prevents a desync where the remote client starts a walking to the old target position when the teleport is finished but the the new position isn't received yet.
+	if (myPlayer._pmode == PM_SPELL && IsAnyOf(myPlayer.executedSpell.spellId, SpellID::Teleport, SpellID::Phasing, SpellID::Warp))
+		target = {};
```
glebm, line 136: Using `(0, 0)` as the "null" value seems a bit dangerous, consider using a bool or `optional`.

StephenCWills: We're writing `target.x` and `target.y` to the packet header so we'd either need a sentinel value like this one or add an additional byte to the packet headers. I believe the assumption is that `(0, 0)` is always at the edge of the dungeon and therefore unreachable by the player. But maybe `Point { 0xFF, 0xFF }` would be better, since that's basically guaranteed to be out of bounds of the dungeon.

AJenbo: maybe that should be a general change then, there are lots of other cases where 0,0 is used as the null position.

AJenbo: maybe that should be a general change then, there are lots of other cases where 0,0 is used as the null position.

PR #5903, opened 2023-03-20
Multiplayer: Add arena potions

obligaron:
This pr add's arena potions.

Notes:
- they can be generated by `/arenapot {number}` command
- they only work in arenas
- they are not persisted in save games. this ensures backwards compatibility.
- they can't be moved to the stash

Thanks @StephenCWills for helping with the save game logic 🙂 

Example video:

https://user-images.githubusercontent.com/25415264/226454115-1637ee31-6b2c-49cb-87b7-54a1d7c5789d.mp4


StephenCWills (changes requested):
I think we should add `IMISC_ARENAPOT` here so that the gamepad hotkeys will work with them. Not that I think anyone will necessarily be playing PvP on gamepad, but those hotkeys were also added to keymapper.

https://github.com/diasurgical/devilutionX/blob/4826e5819e9ae3dc8d6a4cf61af5df7945a30250/Source/controls/plrctrls.cpp#L1821

I suppose this would potentially cause problems with those hotkeys if you carry arena potions around in your belt outside of the arenas. The logic would select the arena potion to be consumed, but then the potion wouldn't be usable and nothing would happen. So maybe it actually needs a special case.

Just an additional note: touch controls will need to be adjusted to display the new graphic. Since touch controls are well outside the current use case for PvP, I think we can merge this first, and I can have a look at them later.

StephenCWills (approved):

StephenCWills:
There was a flurry of activity a few weeks back while they were planning their tournament, but it must have died down pretty quick because it was canceled due to a lack of interest. 🤷

PR #5913, opened 2023-03-21
dun_renderer: Avoid calling 0-width `Blit*`

glebm:
A slight optimization to avoid calling blit functions entirely when the width is zero.


AJenbo (approved):
Looks complicated, good thing it's inside the blit render code :)

PR #5917, opened 2023-03-23
Clean up FindSlotUnderCursor

AJenbo:
@ephphatha am I missing something here :)

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -302,37 +302,23 @@ bool AutoEquip(Player &player, const Item &item, inv_body_loc bodyLocation, bool
 
 int FindSlotUnderCursor(Point cursorPosition, Size itemSize)
 {
-	int i = cursorPosition.x;
-	int j = cursorPosition.y;
-
+	Point hotSpot = cursorPosition;
 	if (!IsHardwareCursor()) {
 		// offset the cursor position to match the hot pixel we'd use for a hardware cursor
-		i += itemSize.width * INV_SLOT_HALF_SIZE_PX;
-		j += itemSize.height * INV_SLOT_HALF_SIZE_PX;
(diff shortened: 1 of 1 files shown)


AJenbo:
Oh that makes sens, thanks I'll correct that and update the PR.

Source/inv.cpp
```diff
@@ -302,37 +302,23 @@ bool AutoEquip(Player &player, const Item &item, inv_body_loc bodyLocation, bool
...
-			if (itemSize.width % 2 == 0)
-				i -= INV_SLOT_HALF_SIZE_PX;
-			if (itemSize.height % 2 == 0)
-				j -= INV_SLOT_HALF_SIZE_PX;
-		}
-		if (r == SLOTXY_INV_LAST && itemSize.height % 2 == 0)
-			j += INV_SLOT_HALF_SIZE_PX;
+		Rectangle slot {
+			panel + Displacement { InvRect[r].x, InvRect[r].y - InventorySlotSizeInPixels.height },
+			InventorySlotSizeInPixels
```
ephphatha, line 316: This also introduces an off-by one. The hit areas used in the old code are inclusive on both ends for the width. This is actually important because the definition of `InvRect` is pretty garbage, there's a line of pixels between the left and right columns of the weapon slots. With the old cold the hit area is enlarged by a single pixel closing this gap, this PR makes it possible to fail to equip a weapon by clicking on that line. The gutter between cells in the inventory and belt also becomes inactive unless you use `InventorySlotSizeInPixels + 1`.

The hit areas for body slots in general are a mess -_-, I ended up pulling out the proper regions and making #5925 for it, but this is what is currently covered by InvRect in master:
2023-03-25T154617_devilutionx_markup

Apart from the gap in the middle of the weapon slots the ring areas are offset slightly left of where they should be, and head/body/right arm graphics are larger than the actual hit area.

AJenbo: i did notice some of these issues yestaday when play testing some other tthings. in general there are still a lot that can be cleaned up about the inventory code :)

PR #5920, opened 2023-03-24
Blitter: Minor optimization

glebm:
`src < end - 3` results in one more instruction than `src + 3 < end`.

Godbolt: https://godbolt.org/z/PPfhYYWG8

RG-99 binary becomes 224 bytes smaller.


AJenbo (approved):
Wild 😄

PR #5923, opened 2023-03-24
Add cutttw widescreen load screen

AJenbo:


AJenbo:
> It opens in d1g so it should be okay.

All the broken ones were made using d1g :D

StephenCWills:
> All the broken ones were made using d1g :D

Yeah, but they also wouldn't load in d1g, so I still trust it for this.

AJenbo:
Odd I seam to remember they did load in the version i was using

StephenCWills:
I'm using a version from December.

PR #5925, opened 2023-03-25
Use rectangles for inventory/belt hit areas

ephphatha:
Controller movement should be working as normal now but I still haven't tested in depth, this was a quick proof of concept for a comment on #5917 

`inv_xy_slot` and `inv_item` can be deduplicated, haven't done that yet either.

Pretty much all the rectangles were defined as the coordinate of the bottom left corner, not the top left as would be normal for the way the UI coordinates work. This seems to have been to avoid adding an offset when rendering, but there was a -1 y offset anyway... changing that to the top left corner makes the inventory code way simpler.

### DIFF
--- a/Source/controls/plrctrls.cpp
+++ b/Source/controls/plrctrls.cpp
@@ -584,32 +584,32 @@ Point InvGetEquipSlotCoord(const inv_body_loc invSlot)
 	Point result = GetPanelPosition(UiPanels::Inventory);
 	switch (invSlot) {
 	case INVLOC_HEAD:
-		result.x += ((InvRect[SLOTXY_HEAD_FIRST].x + InvRect[SLOTXY_HEAD_LAST].x) / 2);
-		result.y += ((InvRect[SLOTXY_HEAD_FIRST].y + InvRect[SLOTXY_HEAD_LAST].y) / 2);
+		result.x += InvRect[SLOTXY_HEAD].Center().x;
+		result.y += InvRect[SLOTXY_HEAD].Center().y;
 		break;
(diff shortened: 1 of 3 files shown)


AJenbo:
i think this is the right thing to do, it should also simplify some of the controller code

ephphatha:
I can't run the demo locally but pretty sure that test was failing because I dropped the case for pasting an item in the right hand slot, good thing it was simulating mouse input in that case :D

Source/controls/plrctrls.cpp
```diff
@@ -584,32 +584,32 @@ Point InvGetEquipSlotCoord(const inv_body_loc invSlot)
 	Point result = GetPanelPosition(UiPanels::Inventory);
 	switch (invSlot) {
 	case INVLOC_HEAD:
-		result.x += ((InvRect[SLOTXY_HEAD_FIRST].x + InvRect[SLOTXY_HEAD_LAST].x) / 2);
-		result.y += ((InvRect[SLOTXY_HEAD_FIRST].y + InvRect[SLOTXY_HEAD_LAST].y) / 2);
+		result.x += InvRect[SLOTXY_HEAD].Center().x;
+		result.y += InvRect[SLOTXY_HEAD].Center().y;
```
AJenbo, line 588: We can do this now right?
```suggestion
		result += InvRect[SLOTXY_HEAD].Center();
```

ephphatha:
Think it needs an explicit cast to displacement, but otherwise yeah that
should work.

AJenbo (approved):

PR #5940, opened 2023-04-01
aarch64-linux-gnu cross-compilation

glebm:
Allows one to easily cross-compile for aarch64 on a Debian/Ubuntu host.
Also adds a GitHub Actions build with automated upload on release.


Trihedraf:
Looks way better than mine 👍 🤦
I don't know if the appimage works for arm, it gives me a `cannot execute binary file: Exec format error`.
And the regular executable is missing libfmt8 as only libfmt7 is in default raspbian repos. I don't know if that's changeable when building or just needs to be upgraded on the pi manually.

glebm:
I've updated it to not use the system libfmt, so it should work now.
As for AppImage, turns out it produces an x86_64 binary so I've disabled it for now.

AJenbo:
Not finding libfmt was a common complaint from Pi users.

AJenbo:
nice that you also fixed the appimage script

Trihedraf (approved):

AJenbo (approved):

PR #5941, opened 2023-04-02
Translate/Generate item names on the fly

### LINKED ISSUE #4382: Ground item names not regenerated on language change
This is a somewhat tricky and minor issue

If an item on the ground was generated with a different language, its name is not updated until it is picked up and then dropped.

E.g.:

1. Set language to Polish
2. Kill a monster, leave the item on the ground, save game
3. Set language to English, load game
4. Item on the ground uses the Polish name
5. Pick up the item -- still Polish name
6. Drop the item -- name is now English (as it should be)

obligaron:
Fixes #4382

With this PR all item names are
- translated on the fly
- always stored/saved in english

Notes:
- `Item::getName` handles the translation and is used on all UI related places.
- The translation logic tries to interpret the info stored in the `Item` class where possible (for normal and unique items this is always good enough).
- For magical items the logic of `RecreateItem` is replicated in a minimal way.
- Common logic between name and item generation is extracted (for example `GenerateStaffNameMagical`).
- Could rename hacked items (this items would also morph when recreating) 😉 
- The translated string could be stored in `Item` and not generated on every call if desired. But this would need a new member in `Item` and probably the performance impact isn't that great.
- Item name generation runs also for english, cause we can have existing items with stored translated names. Also this ensures that everyone sees the same result, if a bug in the generation logic exists.
- Generation is tested with all packed tests (`pack_test.cpp` / `TestItemNameGeneration`).
- I tried to tested with this to diablo converted vanilla hellfire save games. But couldn't reproduce the issues described in [AJenbos comment](https://github.com/diasurgical/devilutionX/issues/4382#issuecomment-1094291169). Probably I miss something and I would be grateful for tips to reproduce this problem. 🙂


Source/items.cpp
```diff
@@ -4531,6 +4733,20 @@ void Item::updateRequiredStatsCacheForPlayer(const Player &player)
 	_iStatFlag = player.CanUseItem(*this);
 }
 
+std::string Item::getName() const
+{
+	if (isEmpty()) {
```
glebm, line 4738: Let's generate the name once and store it on the item, returning a `string_view` here to avoid doing this on every frame?

AJenbo:
Hellfire items do not receive the is_hellfire flag when convert from single player afaik which is why they would morph on regen. If they actually do get it then I'm wrong but also surprised.

Source/items.cpp
obligaron: This has two down-sides
- Make `Item` larger, cause of a new std::string member
- We would need to ensure to invalidate/regenerate the name when the `Item` object changes

To reduce the size impact, we could only store the currently shown name (respect identified flag).
To ensure that the name is correctly invalidated we could introduce a new `Item::initialize` and `Item::identify` method.
The name would only be recreated on-demand (first call to `Item::getName`).

Any thoughts?

AJenbo: We could probably also drop the internal/English name and just generate that when saving to save some memory

obligaron: For the a little while I would keep `_iName` and `_iiName`, cause I think a little bit testing of the changes would be good.
Not that there is a bug and the correct names are gone when we are dropping and regenerating the names. 😅 

Also the generation the logic should fast for normal and unidentified items and that's are most of the items on the ground.

obligaron:
> Hellfire items do not receive the is_hellfire flag when convert from single player afaik which is why they would morph on regen. If they actually do get it then I'm wrong but also surprised.

Yeah you are right. 👍 
I tired to add a logic that tries to set the correct flag for these items.
This should also help a little to avoid morphing the items when regenerating.

Thanks for the help on discord. 🙂

Source/items.cpp
```diff
@@ -1459,6 +1503,28 @@ void ItemRndDur(Item &item)
...
+	if (GenerateRnd(100) <= 10 || GenerateRnd(100) <= lvl) {
+		iblvl = lvl;
+	}
+	if (iblvl == -1 && miscId == IMISC_STAFF) {
+		iblvl = lvl;
+	}
+	if (iblvl == -1 && miscId == IMISC_RING) {
+		iblvl = lvl;
+	}
+	if (iblvl == -1 && miscId == IMISC_AMULET) {
```
AJenbo, line 1518: is there a need for checking iblvl here? it should just set the same vale, later conditions do no appear to check either.

Source/items.cpp
```diff
@@ -2217,6 +2267,166 @@ _item_indexes RndItemForMonsterLevel(int8_t monsterLevel)
 	});
 }
 
+std::string GetTranslatedItemName(const Item &item)
```
AJenbo, line 2270: Since some names like oils and non magical items are taken directly from the translation table maybe we could return `StringOrView` to avoid string copying in a decent amount of case?

Source/items.cpp
obligaron: The code was only moved to a function and not modified. 😉 
I don't think this is necessary, as it always set iblvl to lvl.
Anyway I simplified the logic. Please take a look. 🙂

AJenbo (approved):

Source/items.cpp
```diff
@@ -4531,6 +4731,20 @@ void Item::updateRequiredStatsCacheForPlayer(const Player &player)
...
 
+StringOrView Item::getName() const
+{
+	if (isEmpty()) {
+		return StringOrView(string_view(""));
+	} else if (!_iIdentified || _iCreateInfo == 0 || _iMagical == ITEM_QUALITY_NORMAL) {
+		return GetTranslatedItemName(*this);
+	} else if (_iMagical == ITEM_QUALITY_UNIQUE) {
+		const auto &baseItemData = AllItemsList[static_cast<size_t>(IDidx)];
+		return std::string(_(UniqueItems[_iUid].UIName));
```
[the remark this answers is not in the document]
obligaron: Nice catch. Forgot this while changing to `StringOrView`.

PR #5944, opened 2023-04-03
Add Wii and Wii U builds

AJenbo:
Just enough to get them building, I don't have any way to test it atm


heinermann:
And here I am trying to get rid of my Wii U.

PR #5945, opened 2023-04-03
Remove devilutionx.mpq from Linux release archive

glebm:
`devilutionx.mpq` is already present in the deb/rpm.


glebm:
Good point!

PR #5957, opened 2023-04-08
Multiplayer: Enable Slain Hero

### LINKED ISSUE #5955: MP: Enable Slain Hero
```cpp
if (currlevel == 9 && !gbIsMultiplayer)
			AddSlainHero();
```

obligaron:
Fixes #5955

Note:
To avoid desync for the spawned loot `_oRndSeed` is used (similar to chests).
Due to the additional `AdvanceRndSeed` in `AddObject` this PR can result in a different object layout even in singleplayer.
As far as I know we have multiple reasons why monster and object layout can differ from vanilla, so it shouldn't be a problem.
For existing singleplayer saves the layout is stored in the save.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -2188,28 +2188,30 @@ void OperateInnSignChest(const Player &player, Object &questContainer, bool send
 	}
 }
 
-void OperateSlainHero(const Player &player, Object &corpse)
+void OperateSlainHero(const Player &player, Object &corpse, bool sendmsg)
 {
 	if (corpse._oSelFlag == 0) {
 		return;
 	}
 	corpse._oSelFlag = 0;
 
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -3821,7 +3823,7 @@ void InitObjects()
 	} else {
 		ApplyObjectLighting = true;
 		AdvanceRndSeed();
-		if (currlevel == 9 && !gbIsMultiplayer)
+		if (currlevel == 9 && !UseMultiplayerQuests())
 			AddSlainHero();
```
ephphatha, line 3827: can move the `AdvanceRndSeed()` call here to an else statement to end up with the same number of rng calls

AJenbo (approved):
Nice, that was quick :)

PR #5962, opened 2023-04-09
Remove optimization that skips quest sync at game start

StephenCWills:
This resolves an issue where clients with different `Randomize Quests` settings do not agree on which quests are active in the same game.

The idea here is that the `randomizeQuests` setting does not need to be synced in game init info because `DeltaSyncJunk()` is always called shortly after `InitQuests()` to sync quest state between clients. However, `sgbDeltaChanged` was being used to skip `CMD_DLEVEL_JUNK` if dlvl or quest state hadn't changed since the game was created. The optimization is only relevant if a player joins a game that was newly created, which must have seemed sensible in 1996 to speed up load time and reduce bandwidth for cases where players coordinate their play session in advance or use a game session solely to trade items. However, it has negligible impact on performance these days and skips quest synchronization so I've removed it.

### DIFF
--- a/Source/msg.cpp
+++ b/Source/msg.cpp
@@ -227,7 +227,6 @@ byte sgRecvBuf[1U + sizeof(DLevel::item) + sizeof(uint8_t) + (sizeof(WorldTilePo
 _cmd_id sgbRecvCmd;
 std::unordered_map<uint8_t, LocalLevel> LocalLevels;
 DJunk sgJunk;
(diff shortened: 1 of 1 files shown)


obligaron (approved):
lgtm
thanks for the fix and it's also more readable 🙂

AJenbo:
Nice, I doubt it was ever much of a difference to have this off. And it's at a time where you probably expect a delay anyway.

PR #5965, opened 2023-04-10
Multiplayer: Sync circles in Lazarus' Lair

### LINKED ISSUE #5963: MP: Lazarous chamber doesn't sync
When one player (player 2) steps on the teleport pads and uses the books this doesn't open the two small rooms for the other player (player 1) in the quest level.
Both players where present on the level at the same time.

I think this was also the case for the main room but could be mistaken.

obligaron:
Fixes #5963

Notes:
- Check circle state independent of `MyPlayer` (also makes the circle red for remote players)
- In `OperateBook` only check if the player stands on the circle if it's the local client, cause on remote clients this can desync
- Call `OperateBook` also in `SyncOpObject` without the position check

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1563,9 +1563,9 @@ void UpdateObjectLight(Object &light, int lightRadius)
 
 void UpdateCircle(Object &circle)
 {
-	Player &myPlayer = *MyPlayer;
+	Player *playerOnCircle = PlayerAtPosition(circle.position);
 
-	if (myPlayer.position.tile != circle.position) {
+	if (!playerOnCircle) {
 		if (circle._otype == OBJ_MCIRCLE1)
 			circle._oAnimFrame = 1;
 		if (circle._otype == OBJ_MCIRCLE2)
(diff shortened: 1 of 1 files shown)


Source/objects.cpp
```diff
@@ -1913,32 +1915,26 @@ void OperateBook(Player &player, Object &book, bool sendmsg)
...
 
-			if (doAddMissile) {
-				questObject._oVar6 = 4;
-				ObjectAtPosition({ 35, 36 })._oVar5++;
-				AddMissile(player.position.tile, target, Direction::South, MissileID::Phasing, TARGET_BOTH, player.getId(), 0, 0);
-				missileAdded = true;
-			}
-		}
-		if (!missileAdded) {
+		// Only verfiy that the player stands on the circle when it's the local player (sendmsg), cause for remote players the position could be desynced
```
qndel, line 1930:
```suggestion
		// Only verfiy that the player stands on the circle when it's the local player (sendmsg), because for remote players the position could be desynced
```

Source/objects.cpp
```diff
@@ -1913,32 +1915,26 @@ void OperateBook(Player &player, Object &book, bool sendmsg)
...
 
-			if (questObject._otype == OBJ_MCIRCLE2 && questObject._oVar6 == 1) {
-				target = { 27, 29 };
-				doAddMissile = true;
-			}
-			if (questObject._otype == OBJ_MCIRCLE2 && questObject._oVar6 == 2) {
-				target = { 43, 29 };
-				doAddMissile = true;
-			}
+		Object &circle = ObjectAtPosition(book.position + Direction::SouthWest);
```
AJenbo, line 1927: Nice

AJenbo:
Good use of comments

PR #5968, opened 2023-04-10
 Re-encode all the bundled CLX assets 

glebm:
Following the CLX encoding fix in https://github.com/diasurgical/devilutionx-graphics-tools/pull/3

Total size, bytes: 5,154,051 -> 5,063,902 (-1.7%)

```
find Packaging/resources/assets -type f -name '*.clx' -exec du -cb {} + | tail -1
```


AJenbo:
Thats about the same total size saving as 46-00.clx, not bad :)

PR #5975, opened 2023-04-14
Refactor earflag

obligaron:
Replaces `earflag` with a new enum `DeathReason` and refactor `StartPlayerKill`.
Also added some comments to make the behavior hopefully clearer.


AJenbo (approved):
Nice work, at this point I think ears is kinda understandable. Thanks everyone :)

PR #5977, opened 2023-04-15
Use padmapper bindings on spell icons

StephenCWills:
Had to limit button names to two characters so they'd fit in the spell icons. Here's a couple screenshots.

Generic gamepad:
image

PlayStation:
image


AJenbo:
nice :eyes: !

AJenbo (approved):

PR #5984, opened 2023-04-16
Update gamepad motion state regardless of the active event handler

### LINKED ISSUE #5946: [Controller] Players keeps walking the same direction after loading screen
### Operating System

Android

### DevilutionX version

### Describe

 Version   #7602

With a controller only, if the player changes level, he will automatically walks toward the same direction as before the loading screen even if the stick is no longer pressed. This only occurs when using the joystick, using the dpad works perfectly.

The issue is not present with touch controls which also makes use of a joystick to move around. It could be interesting to see how the issue is avoided there.

### To Reproduce

1. Enter the dungeon by pressing the joystick (not Dpad)
2. Release the joystick during the loading screen
3. Witness how the character keeps moving the same direction

### Expected Behavior

_No response_

### Additional context

_No response_

StephenCWills:
This PR splits `ProcessControllerMotion()` up so that gamepad thumbstick state can be updated consistently in `PollEvent()` regardless of the event handler currently in use by the game.

* Reduce `ProcessControllerMotion()` to only updating thumbstick state based on an SDL event
* Remove unimplemented function `ProcessKbCtrlAxisMotion()`
* Introduce `IsControllerMotion()` to indicate whether an SDL event represents controller motion, taking the place of the bool previously returned by `ProcessControllerMotion()`
* Call `SimulateRightStickWithPadmapper()` directly in the game event handler

This resolves #5946

### DIFF
--- a/Source/DiabloUI/settingsmenu.cpp
+++ b/Source/DiabloUI/settingsmenu.cpp
@@ -479,7 +479,7 @@ void UiSettingsMenu()
 
 				StaticVector<ControllerButtonEvent, 4> ctrlEvents = ToControllerButtonEvents(event);
 				for (ControllerButtonEvent ctrlEvent : ctrlEvents) {
-					bool isGamepadMotion = ProcessControllerMotion(event, ctrlEvent);
+					bool isGamepadMotion = IsControllerMotion(event);
 					DetectInputMethod(event, ctrlEvent);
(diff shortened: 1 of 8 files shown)


AJenbo:
seems a better way to handle things as well

AJenbo (approved):

Issue #5970, opened 2023-04-11
Monster health on Hell difficulty too low


Miz4r:
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

I started Hell difficulty on single player for Diablo 1 on my lvl 28 Warrior and I was surprised to oneshot everything except bosses doing only 50-60 character screen dmg. Monster health seems to be lower than should be for this difficulty. I checked the official guides for Diablo 1 and most monsters I am oneshotting should have at least 100 hp. See attached screenshot below for an example.

![DevilutionXhell](https://user-images.githubusercontent.com/6607310/231214584-ba648f58-4ebe-48b9-ab06-ede936a8a11b.jpg)

### To Reproduce

1. Start a single player game for Diablo 1 (not Hellfire)
2. New game on Hell difficulty with high level character
3. Kill normal monsters on dlvls 1-4 and observe how they die too easily
4. Check health numbers once you killed enough of a certain type and see how they don't represent their official values for Hell difficulty

### Expected Behavior

I expected monsters on Hell difficulty to have more health

StephenCWills:
I'm not sure what official guide you're referring to, but Single Player vanilla Diablo does not officially have HP values for Hell difficulty because it was only possible to play in Hell difficulty due to a bug. What you've shown in your screenshot is approximately half of the HP value listed for Scavengers in Jarulf's Guide, and that seems about right to me.

![image](https://user-images.githubusercontent.com/9203145/231230765-6697a55c-a1f0-4a32-a6de-62bae077f771.png)

galaxyhaxz:
_**This is actually a bug**_: in patches 1.00-1.04 monsters were supposed to get +100 life on nightmare and +200 for hell (shown when hovering over monster). However, when adding those values to their life they forgot to shift it to leave room for the fractional bits. This means they would get 100/64 (**1**) and 200/64 (**3**) life instead.

Whoever at Blizzard made patch 1.07 likely misunderstood the issue, and changed the life additions to actually be +1 and +3 accordingly and reflect that in the hover over. Since the life scaling formula just simply multiplies their base life then adds this, monsters which had a low base life will not scale properly to the difficulty and be far too easy.

Hellfire fixed it... though they cut the bonus in half for single player.

julealgon:
Would you advocate for attempting to do the original intended change on this one @galaxyhaxz ?

galaxyhaxz:
[...]

PR #5987, opened 2023-04-16
Fix monster health to +100 for NM and +200 for Hell instead of +1 and +3

Trihedraf:
As discussed in #5970


AJenbo (approved):
What a mess of a history for this, but sems to checkout :+1:

PR #5989, opened 2023-04-16
When player position desyncs ensure position.old is also updated

### LINKED ISSUE #5981: Crash on player changing level because of left over in dPlayer[][]
It looks like the player entry in dPlayer[][] is not being properly cleared when changing levels.
This can lead to crashes in the render where it tries to render a player who is nolonger on the current level.
In 1.4.1 the whole level was cleared, but in master it is now only clearing in an area around where the player was last seen.

obligaron:
Could fix #5981

`FixPlrWalkTags` uses `player.position.old` but in some cases the old position is not updated correctly:
- when joining a game
- when desyncs happen and the remote player position is instant updated/"teleported"
- when player is not on the same level (this is relevant when another player joins the level)

This PR resolves the desyncs from above and add's a debug routine to help diagnostic this kind of problems earlier.

Steps to reproduce #5981 in master:
- In StartSpell change NewPlrAnim to: `NewPlrAnim(player, GetPlayerGraphicForSpell(player.queuedSpell.spellId), d, animationFlags, -100, player._pSFNum);` note: this change is only needed to have enough time for the cast between the two clients 😉
- Player 1 goes to dungeon
- Player 1 walks 5 tiles away from upstairs entry
- Player 1 cast teleports to stairs
- Player 2 goes down to dungeon
- Player 2 sees Player 1 standing (note: cast is not synced)
- Player 1 finishes teleport cast
- Player 1 starts town warp
- Player 2 still sees Player 1 in dungeon
- ==> crash

Note:
Perhaps there are other reasons why `player.position.old` is out of sync.
So it is still possible #5981 is not fixed with this PR.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2588,6 +2588,7 @@ void InitPlayer(Player &player, bool firstTime)
 		}
 
 		player.position.future = player.position.tile;
+		SetPlayerOld(player);
 		player.walkpath[0] = WALK_NONE;
 		player.destAction = ACTION_NONE;
 
@@ -2701,6 +2702,16 @@ void FixPlrWalkTags(const Player &player)
 			dPlayer[searchTile.x][searchTile.y] = 0;
 		}
 	}
+
+#ifdef _DEBUG
+	// Checks that FixPlrWalkTags really removes player from all dPlayer tiles
+	// FixPlrWalkTags could fail if player.position.old is not updated correctly
+	for (int y = 0; y < MAXDUNY; y++) {
+		for (int x = 0; x < MAXDUNX; x++) {
+			assert(PlayerAtPosition({ x, y }) != &player);
+		}
+	}
+#endif
 }
 
 void StartPlrHit(Player &player, int dam, bool forcehit)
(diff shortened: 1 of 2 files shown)


AJenbo:
> Perhaps there are other reasons why player.position.old is out of sync.

Considering that I was playing a Warrior with almost no ability to cast magic this is likely true :)

Source/player.cpp
```diff
@@ -2701,6 +2702,16 @@ void FixPlrWalkTags(const Player &player)
...
 	}
+
+#ifdef _DEBUG
+	// Checks that FixPlrWalkTags really removes player from all dPlayer tiles
+	// FixPlrWalkTags could fail if player.position.old is not updated correctly
+	for (int y = 0; y < MAXDUNY; y++) {
+		for (int x = 0; x < MAXDUNX; x++) {
+			assert(PlayerAtPosition({ x, y }) != &player);
+		}
+	}
```
AJenbo, line 2713: This might indeed be very helpful for catching any further cases. Good thinking :+1:

AJenbo (approved):

AJenbo:
Been testing this for a few rounds. It used to happen multiple times during a session, but since this was merged I haven't see the issue once. Nicely done :+1:

PR #5996, opened 2023-04-18
Fix light brightness pop on load

AJenbo:
Light radios for 3 objects did not match between `addObject` and `updateObject`. This caused some lights (especially in the cathedral) to fade in with the lower brightness on level loading, then pop to full brightness once the transition is over.

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -256,10 +256,11 @@ void DoLighting(Point position, uint8_t radius, int Lnum)
 		maxY = MAXDUNY - position.y;
 	}
 
-	if (IsNoneOf(leveltype, DTYPE_NEST, DTYPE_CRYPT)) {
-		SetLight(position, 0);
-	} else if (GetLight(position) > LightFalloffs[radius][0]) {
+	// Allow for dim lights in crypt and nest
+	if (IsAnyOf(leveltype, DTYPE_NEST, DTYPE_CRYPT)) {
 		SetLight(position, LightFalloffs[radius][0]);
+	} else if (GetLight(position) > LightFalloffs[radius][0]) {
+		SetLight(position, 0);
 	}
 
 	for (int i = 0; i < 4; i++) {
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1532,7 +1532,7 @@ void AddMushPatch()
 bool IsLightVisible(Object &light, int lightRadius)
 {
 #ifdef _DEBUG
-	if (!DisableLighting)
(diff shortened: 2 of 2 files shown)


obligaron:
This changes behavior in non set-maps.
Because in normal levels most lights are static and `UpdateObjectLight` does nothing.
That means it still uses the old light radius of 5.
But in set-maps it uses the light radius of 10, cause `UpdateObjectLight` is executed.

I think it is desirable to make the light radius always the same, but I would take the lower light radius as master and change set-maps light radius.

Master
grafik

PR
grafik

AJenbo:
Odd I was testing on dlvl 1 and getting the inital image during transition but then it would jump to the second one after the transition. So I was not seeing a change at run time, just a fix to the transition.

I'll test things some more.

Source/objdat.cpp
```diff
@@ -275,7 +275,7 @@ const ObjectData AllObjects[109] = {
 /*OBJ_URN*/          { OFILE_URN,           21,      24, DTYPE_NONE,      THEME_NONE,              Q_INVALID,  Solid | MissilesPassThrough | Light | Breakable,          1,        9,         96,        3 },
 /*OBJ_URNEX*/        { OFILE_URNEX,         21,      24, DTYPE_NONE,      THEME_NONE,              Q_INVALID,  Solid | MissilesPassThrough | Light | Breakable,          1,       10,         96,        3 },
 /*OBJ_L5BOOKS*/      { OFILE_L5BOOKS,       21,      24, DTYPE_NONE,      THEME_NONE,              Q_INVALID,  Solid | MissilesPassThrough | Light,                      1,        0,         96,        3 },
-/*OBJ_L5CANDLE*/     { OFILE_L5CANDLE,      21,      23, DTYPE_NONE,      THEME_NONE,              Q_INVALID,  Animated | Solid | MissilesPassThrough | Light,           2,        4,         96,        0 },
+/*OBJ_L5CANDLE*/     { OFILE_L5CANDLE,      21,      23, DTYPE_NONE,      THEME_NONE,              Q_INVALID,  Animated | Solid | MissilesPassThrough,                   2,        4,         96,        0 },
```
AJenbo, line 278: These emit a very faint light and should not be fully lit.
```suggestion
/*OBJ_L5CANDLE*/     { OFILE_L5CANDLE,      21,      23, DTYPE_NONE,      THEME_NONE,              Q_INVALID,  Animated | Solid | MissilesPassThrough | Light,                   2,        4,         96,        0 },

```

StephenCWills (approved):

PR #5997, opened 2023-04-18
Only reset _oVar1 for dynamic lights

obligaron:
With #3292 dynamic lights _oVar1 are reset, so that they can be recreated after changing the level.
But it also changed _oVar1 for static lights (they have _oVar1 = -1).
Because of this after going back to a already visited level (_oVar1 was saved with 0) a additional light source was added.
This also manifests in the timedemo test shown difference.


AJenbo (approved):
Was wondering about this as well, I'm currently working on some other corrections for this code :)

PR #5998, opened 2023-04-18
Consistently use uint8_t for lighting values

AJenbo:
This gets rid of a few more cases of `char` which may or may not be signed depending on the platform.


glebm (approved):
Nice!

PR #6001, opened 2023-04-18
Debug: Improve rendering of light gird

AJenbo:
LightsMax is the darkest value, and the most common so lets not have that rendered all over the screen.


qndel:
Looks like this would cause 0's to not get displayed anywhere. How about return false directly in the switch if value is LightsMax?

PR #6005, opened 2023-04-18
Unify vision and lights

AJenbo:
Vision will never grow to more then 4, since there is only 4 players... so why have a list of 32 of them.

Since vision needs to stay in sync with your light, might as well drop all the duplicate code. The only special case is remote players where the lighting should not be rendered.

Also optimize the Light struct.

All the changes to data structures should save us about 3KB ram, but there should also be some savings in terms of runtime processing and less code generation.

Use `Rectangle` logic in `DoUnLight()` and `DoUnVision()`, and take a `Point` as an input.

Have `ProcessLightList()` free deleted lights before applying lighting to simplify the process. And use early return to avoid indentation of the function body (so probably view with "ignore white-space").

Remove `DisableLighting` from release builds.

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -16,8 +16,6 @@
 
 namespace devilution {
 
-std::array<bool, MAXVISION> VisionActive;
-Light VisionList[MAXVISION];
 Light Lights[MAXLIGHTS];
 std::array<uint8_t, MAXLIGHTS> ActiveLights;
 int ActiveLightCount;
@@ -29,6 +27,7 @@ std::array<uint8_t, 256> PauseTable;
 bool DisableLighting;
 #endif
 bool UpdateLighting;
+bool UpdateVision;
 
 namespace {
 
@@ -68,7 +67,6 @@ const DisplacementOf<int8_t> VisionCrawlTable[23][15] = {
 constexpr size_t NumLightRadiuses = 16;
 /** Falloff tables for the light cone */
 uint8_t LightFalloffs[NumLightRadiuses][128];
-bool UpdateVision;
 /** interpolations of a 32x32 (16x16 mirrored) light circle moving between tiles in steps of 1/8 of a tile */
 uint8_t LightConeInterpolations[8][8][16][16];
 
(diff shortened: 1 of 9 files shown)


Source/lighting.cpp
```diff
@@ -114,26 +114,16 @@ uint8_t GetLight(Point position)
 	return dLight[position.x][position.y];
 }
 
-void DoUnLight(int nXPos, int nYPos, int nRadius)
+void DoUnLight(Point position, uint8_t radius)
```
glebm, line 117: The argument named "radius" is not actually a radius.
Consider changing the arguments to a single `WordTileRectangle area`

Source/lighting.cpp
```diff
@@ -290,23 +315,20 @@ void DoLighting(Point position, int nRadius, int lnum)
 	}
 }
 
-void DoUnVision(Point position, int nRadius)
+void DoUnVision(Point position, uint8_t radius)
```
glebm, line 318: Same, accept a `WordTileRectangle area`?

Source/lighting.cpp
AJenbo: The way I read it it is a radius, but for simplicity the implementation updates the tiles in a bounding square rather then a circle. Since it's the implementation that makes use of a Rectangle I would find it a little odd to make the function take that as the input, especially since we do not have a Rectangle at the call site and would need to construct one any way then.

AJenbo:
Please note that a lot of this has been punted of in to https://github.com/diasurgical/devilutionX/pull/6014 while I work on resolving the failing test.

Source/lighting.cpp
AJenbo: ? are we reading the same code.

The original code has this:
```cpp
	int minX = nXPos - nRadius;
	int maxX = nXPos + nRadius;
	int minY = nYPos - nRadius;
	int maxY = nYPos + nRadius;
```
That very much to me looks like it's using the point as the center. Which also aligns with how DoLight() applies full brightness to the given point which visually appears in the center of the light.

We also call it a radius in the RectableOf constructor:
https://github.com/diasurgical/devilutionX/blob/3304e8a5429e5158b26c2c492e0dc080796df06e/Source/engine/rectangle.hpp#L29

> And the radius of the circle inscribed into that rectangle is actually radius/2, which makes that argument more like a diameter than a radius.

Are you confusing radius and diameter?
https://en.wikipedia.org/wiki/Diameter
https://en.wikipedia.org/wiki/Radius

StephenCWills: There is this other constructor, which takes the top-left corner position and size.

https://github.com/diasurgical/devilutionX/blob/3304e8a5429e5158b26c2c492e0dc080796df06e/Source/engine/rectangle.hpp#L15

However, I agree with AJenbo that this code appears to be calling the constructor that takes center position and radius. We should be avoiding ambiguity since the following constructor for `SizeOf` is marked `explicit`, right?

https://github.com/diasurgical/devilutionX/blob/3304e8a5429e5158b26c2c492e0dc080796df06e/Source/engine/size.hpp#L22

glebm: Ah, right, I confused it with the other constructor, sorry!

AJenbo:
Turns out that Vision is consistent in lacking behind the player in all walking directions. But for light this is only true when walking in northern direction. But for southern it moves instantly like the player position.
image

For left and right the light behaves a bit odd in that it is offset by a row as well. For this the player position instead matches that of the vision.
image

My conclusion here is that none of these are really ideal. But vision is at least consistent. So lets close this until the day that we have fixed things to where all 3 position systems instead matches the visual representation instead (moving to the next tile once the animation is 50% done).

AJenbo:
Reimplemented the loop optimization in https://github.com/diasurgical/devilutionX/pull/6017

PR #6010, opened 2023-04-20
Update timedemo test to include improved lightning

obligaron:
Update timedemo to include improved lightning from #6002


AJenbo (approved):
<3

PR #6018, opened 2023-04-22
Apply u8path() to paths passed into std::filesystem API calls

StephenCWills:
Resolves an issue where multi-byte code points in file paths were being misinterpreted on Windows in functions that made use of the `std::filesystem` API. I went ahead and applied this to all paths passed into `std::filesystem` API calls regardless of whether they are used on Windows since it seems like the correct way to use the API.


AJenbo (approved):

glebm:
Sadly `u8path` is deprecated in C++20 but its replacement, `char8_t`, is not available everywhere, so this does indeed seem to be the best solution. Thanks! 👍

PR #6025, opened 2023-04-23
Fix lighting on quest maps

AJenbo:
This fixes lights being visibly off at a distance and turning on when getting closer.
(If you resolution is high enough to see past the normal render distance)
It also provides a performance boost since these lights no longer have to be updated each frame.
This removes the need for `ApplyObjectLighting` as lighting should always be applied to objects.

Also applies a fix for Lazarus and his minions glowing inside the wall:
image

fixes https://github.com/diasurgical/devilutionX/issues/6028

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -241,8 +241,9 @@ void DoLighting(Point position, uint8_t radius, DisplacementOf<int8_t> offset)
 
 	// Allow for dim lights in crypt and nest
 	if (IsAnyOf(leveltype, DTYPE_NEST, DTYPE_CRYPT)) {
-		SetLight(position, LightFalloffs[radius][0]);
-	} else if (GetLight(position) > LightFalloffs[radius][0]) {
+		if (GetLight(position) > LightFalloffs[radius][0])
+			SetLight(position, LightFalloffs[radius][0]);
+	} else {
 		SetLight(position, 0);
 	}
 
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -578,6 +578,8 @@ void ProcessLightList()
 			i--;
 			continue;
 		}
+		if (TileHasAny(dPiece[light.position.tile.x][light.position.tile.y], TileProperties::Solid))
+			continue; // Monster hidden in a wall, don't spoil the surprise
 		DoLighting(light.position.tile, light.radius, light.position.offset);
 	}
 
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -109,19 +109,6 @@ uint8_t GetLight(Point position)
 	return dLight[position.x][position.y];
 }
 
-void DoUnLight(Point position, uint8_t radius)
-{
-	radius++;
-	radius++; // If lights moved at a diagonal it can result in some extra tiles being lit
-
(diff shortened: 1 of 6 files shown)


Source/objects.cpp
```diff
@@ -1345,14 +1342,35 @@ void AddTrap(Object &trap)
...
+	case OBJ_TORCHR:
+	case OBJ_TORCHL2:
+	case OBJ_TORCHR2:
+		radius = 8;
+		break;
+	default:
+		return;
 	}
+
+	DoLighting(object.position, radius, {});
```
obligaron, line 1372: Question:
`DoLighting` expects that `LoadingMapObjects` is true (else the light would be removed when updating these tiles).
But `LoadingMapObjects` is not always true when calling `AddObjectLight`.
Examples: `InitObjects()` and `CreateThemeRooms()` when loading a game.
Is it relevant?

obligaron, line 1372: When porting to lazarus (books are activated) `LoadingMapObjects` is true.
That means `DoLighting` update `dPreLight`.
But `dLight` is not yet updated. The new light source is visible only after another lightning update happens for the relevant tiles (for example hero or monster moving).

grafik

Source/lighting.cpp
```diff
@@ -578,6 +579,8 @@ void ProcessLightList()
 			i--;
 			continue;
 		}
+		if (TileHasAny(dPiece[light.position.tile.x][light.position.tile.y], TileProperties::Solid))
+			continue; // Monster hidden in a wall, don't spoil the surprise
```
obligaron, line 583: Do we need to update `dLight` when the tile gets visible again?
(For example setting `Light::hasChanged` and `UpdateLighting`)

obligaron (commented):
I like this changes.
They make object lightning consistent and improve performance.
But some edge-cases need to be addressed.

Source/objects.cpp
AJenbo: >  DoLighting expects that LoadingMapObjects is true (else the light would be removed when updating these tiles).

This is only true during gameplay, while initially generating the level lighting is applied to `dLIght`, when the level is fully generated `dLight` will be copied to `dPreLight`. So the global should be off during level generation, but on during gameplay.

Source/lighting.cpp
AJenbo: All lighting should be invalidated when the map changes, which should take care of this.

Source/objects.cpp
AJenbo: fixed :)

obligaron: Forgot about this one.

Source/lighting.cpp
obligaron: Out of curiosity I tested it (because I couldn't find the corresponding sources).
I disabled `AddObjectLight` (removed `DoUnLight`) and lights from players.
This is the result after teleporting to Lazarus:
grafik
The lighting is not updated.
When `AddObjectLight` works normally, this is irrelevant, because the other light sources ensure lighting is recalculated for Lazarus.
So feel free to ignore this comment, if you want. It should be only relevant for modders.

AJenbo: I don't think we can protect people from deleting important parts of the code :)

PR #6026, opened 2023-04-23
Use symlinks to work around lack of Unicode support for paths in ZeroTier

StephenCWills:
In cases where players on Windows OS have chosen a username with non-ASCII characters, attempts to create or join games on ZeroTier fail with the error message `Unknown network error`. This PR resolves that error by working around limitations in the ZeroTier library for Windows.

On Windows systems, the default config path resides in `C:\Users\[username]\AppData\Roaming\diasurgical\devilution` which is a path that could have non-ASCII characters based on the player's username. DevilutionX encodes the path as UTF-8 and then passes it to ZeroTier which proceeds to interpret it as an ANSI-encoded path. For paths with non-ASCII characters, this resolves to an entirely different directory under `C:\Users` which does not exist and must therefore be created in order for ZeroTier to store its configuration there. Since users don't typically have rights to create folders under `C:\Users` without elevated privileges, the game simply fails to initialize ZeroTier and players receive a generic error message when attempting to create or join games.

The workaround creates an alias for the config path under the "Common Application Data" directory which is typically located in `C:\ProgramData`. The original config path is hashed and then converted to hex to generate the name of the symbolic link that targets that config path. Symlinks are written to `C:\ProgramData\diasurgical\devilution` and then the path to the symlink is passed into ZeroTier. This ensures that the path passed to ZeroTier has no non-ASCII characters and will therefore be the same whether encoded using UTF-8 or ANSI.

For the vast majority of cases in which paths already do not have non-ASCII characters, no symlinks will be created.

### DIFF
--- a/Source/dvlnet/zerotier_native.cpp
+++ b/Source/dvlnet/zerotier_native.cpp
@@ -9,6 +9,21 @@
 #include "utils/sdl2_backports.h"
 #endif
 
+#if (defined(_WIN64) || defined(_WIN32)) && !defined(NXDK)
+#include "utils/stdcompat/filesystem.hpp"
+#ifdef DVL_HAS_FILESYSTEM
+#define DVL_ZT_SYMLINK
+#endif
+#endif
+
+#ifdef DVL_ZT_SYMLINK
+#include <shlobj.h>
+#include <sodium.h>
+
+#include "utils/str_cat.hpp"
+#include "utils/utf8.hpp"
+#endif
+
 #include <ZeroTierSockets.h>
 #include <cstdlib>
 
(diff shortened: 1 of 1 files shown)


Source/dvlnet/zerotier_native.cpp
```diff
@@ -20,6 +35,88 @@
...
+namespace {
+
+#ifdef DVL_ZT_SYMLINK
+bool HasMultiBytesChars(string_view path)
+{
+	for (char c : path) {
+		if (IsTrailUtf8CodeUnit(c))
+			return true;
+	}
+	return false;
```
glebm, line 47:
```suggestion
	return std::any_of(path.begin(), path.end(), IsTrailUtf8CodeUnit);
```

Source/dvlnet/zerotier_native.cpp
```diff
@@ -20,6 +35,88 @@
...
+	    reinterpret_cast<const unsigned char *>(path.data()), path.size(),
+	    nullptr, 0);
+
+	if (status != 0)
+		return "";
+
+	char alternateFolderName[hashSize * 2 + 1];
+
+	for (size_t i = 0; i < hashSize; i++)
+		snprintf(&alternateFolderName[i * 2], 3, "%02x", hash[i]);
```
glebm, line 65: Use `fmt::format_to`?

Source/dvlnet/zerotier_native.cpp
```diff
@@ -20,6 +35,87 @@
...
+	if (status != 0)
+		return "";
+
+	std::string alternateFolderName(hashSize * 2, '\0');
+	auto iter = alternateFolderName.begin();
+	const auto end = alternateFolderName.end();
+	for (size_t i = 0; i < hashSize; i++) {
+		const auto fmtResult = fmt::format_to_n(iter, end - iter, "{:02x}", hash[i]);
+		iter = fmtResult.out;
+	}
```
glebm, line 64: I think we may even be able to do something like:

```c++
const std::string alternateFolderName = fmt::format("{:02x}", fmt::join(hash));
```

glebm: Yep https://godbolt.org/z/fh79ohxvb

Source/dvlnet/zerotier_native.cpp
```diff
@@ -20,6 +35,87 @@
...
+
+	return alternateFolderName;
```
glebm, line 66:
```suggestion
	return fmt::format("{:02x}", fmt::join(hash, ""));
```

StephenCWills: Nice find! I like this much better.

Source/dvlnet/zerotier_native.cpp
```diff
@@ -9,6 +9,21 @@
 #include "utils/sdl2_backports.h"
 #endif
 
+#if defined(_WIN64) || defined(_WIN32)
```
glebm, line 12: Xbox nxdk currently does not have `std::filesystem` but let's guard against it being added in the future:

```suggestion
#if (defined(_WIN64) || defined(_WIN32)) && !defined(NXDK)
```

glebm (approved):

AJenbo:
Nice work, there are probably a good number of people that were stuck on this one.

PR #6027, opened 2023-04-23
Sync learned/changed spells

### LINKED ISSUE #5980: New learned spells aren't synced
If a hero learns a new spell during a session other players will not see them cast it since they only know of the spells the player new when joining the session, and the spell cast command will there for fail on there end.

obligaron:
Fixes #5980

When changing a spell level a new command `CMD_CHANGE_SPELL_LEVEL` is used.
This is a similar approach to adding stat points (`CMD_ADDSTR`).

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -1979,10 +1979,11 @@ void OperateBook(Player &player, Object &book, bool sendmsg)
 	}
 
 	if (setlvlnum == SL_BONECHAMB) {
-		player._pMemSpells |= GetSpellBitmask(SpellID::Guardian);
-		if (player._pSplLvl[static_cast<int8_t>(SpellID::Guardian)] < MaxSpellLevel)
-			player._pSplLvl[static_cast<int8_t>(SpellID::Guardian)]++;
 		if (sendmsg) {
+			uint8_t newSpellLevel = player._pSplLvl[static_cast<int8_t>(SpellID::Guardian)] + 1;
+			if (newSpellLevel <= MaxSpellLevel)
+				NetSendCmdParam2(true, CMD_CHANGE_SPELL_LEVEL, static_cast<uint16_t>(SpellID::Guardian), newSpellLevel);
+
 			Quests[Q_SCHAMB]._qactive = QUEST_DONE;
 			NetSendCmdQuest(true, Quests[Q_SCHAMB]);
 		}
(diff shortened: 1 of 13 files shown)


test/player_test.h
```diff
@@ -30,6 +30,16 @@ static int Count8(int8_t *ints, int n)
...
 }
 
+static int CountU8(uint8_t *ints, int n)
+{
+	int count = n;
+	for (int i = 0; i < n; i++)
+		if (ints[i] == 0)
+			count--;
+
+	return count;
```
glebm, line 40:
```suggestion
	return std::count_if(ints, ints + n, [](uint8_t x) { return x != 0; });
```

obligaron: I also updated all other Count*-function to `std::count_if`

glebm (approved):

Source/objects.cpp
```diff
@@ -2512,20 +2512,23 @@ void OperateShrineEnchanted(Player &player)
 	if (cnt > 1) {
 		spell = 1;
 		for (int j = static_cast<int8_t>(SpellID::Firebolt); j < maxSpells; j++) { // BUGFIX: < MAX_SPELLS, there is no spell with MAX_SPELLS index (fixed)
-			if ((player._pMemSpells & spell) != 0) {
-				if (player._pSplLvl[j] < MaxSpellLevel)
-					player._pSplLvl[j]++;
+			if ((player._pMemSpells & spell) != 0 && player._pSplLvl[j] < MaxSpellLevel) {
+				NetSendCmdParam2(true, CMD_CHANGE_SPELL_LEVEL, static_cast<uint16_t>(j), static_cast<uint8_t>(player._pSplLvl[j] + 1));
```
AJenbo, line 2516: Should we maybe set this one to low pri, so that it is not sent before all of them are changed, and won't risk a race condition where the one that lowers the level arrives before the one that increases it?

```suggestion
				NetSendCmdParam2(false, CMD_CHANGE_SPELL_LEVEL, static_cast<uint16_t>(j), static_cast<uint8_t>(player._pSplLvl[j] + 1));
```

obligaron: I updated the code to make it more straight forward and only update every spell only once.
Also it fixes the bug that a maxed spell is reduced by 2 instead of 1.

StephenCWills: If `gbShouldValidatePackage` is false when sending a high-priority packet, the game creates a new packet to be sent out to all other players. It copies as much data as it can from the high-priority buffer, then the low-priority buffer, and finally fills the rest with monster sync data. Anything from the high-priority or low-priority buffers that couldn't fit into the packet will stay behind to be sent with the next packet. This means that high-priority packets could be sent before low-priority packets regardless of the order in which `NetSendCmd*` functions were called, but all packets with the same priority as each other should be sent in the same order as the function calls.

StephenCWills: It occurs to me after explaining this that it wouldn't just create a race condition. The packet sent on line 2531 would _always_ be sent before the ones sent on this line because these would only be written to the low-priority buffer. At the earliest, they would be sent to remote clients after 2531 writes its packet to the high-priority buffer.

AJenbo (approved):

Source/objects.cpp
AJenbo: Thanks for explaining that.

Issue #3601, opened 2021-11-24
Tiles disappearing while on screen


NikoVP:
**Important information**
Windows 10 / DevilutionX 273b9e7 / Github actions build

While walking in the Crypt in lower part of the screen you can clearly notice floor tiles disappearing.

It appears to be part of 1.3.0 or later
Use save state.

[single_1.zip](https://github.com/diasurgical/devilutionX/files/7596682/single_1.zip)

Walk around the floor. I also checked release 1.2.0 and this wasn't occurring there

Tiles shouldn't disappear while on screen.

![Animation2](https://user-images.githubusercontent.com/60973634/143266424-1a0fe858-451d-458c-b1cc-0e6b51b71af4.gif)

Chance4us:
@NikoVP I didn't really understand which tiles, but possibly discovered something similar.
Could you check if you have been experiencing the problem since commit 92db72719d9a8efcc24292f96be9866f2e0c0c13?

NikoVP:
What I was referring to was the ground floor tiles. On the animation you can see some of the flame tiles on the crypt floor dissappearing and reappearing. I wasn't able to record a good video, had some trouble even doing that one. Still, you can see how the exir reappears suddenly (after having dissappeared while on screen) instead of just coming into view from the edge of the screen.

AJenbo:
Regarding the issue not appearing in 1.2.0, it's true that things did not pop out of view as much at that point, but there where still many visual artifacts and things did pop into view. These issues have now been resolved and will be part of 1.5.0

PR #6029, opened 2023-04-23
Hellfire: Fix light fall off not within its radius

AJenbo:
Helfire makes various modifications to the lighting system. One of them is to calculate a new falloff table for lights.
Unfortunately this table isn't adhering to the radius bound which results in several visual artifacts (some can be seen here: https://github.com/diasurgical/devilutionX/issues/3601) and light smearing.
On clear destination form the normal lighting is that it allows for dim lights which is used for the lava candles.

It's possible that the overshot lighting was somewhat intentional in order to apply a glow to the floor, but since it will vary in intensity based on the players light radius this isn't very consistent. A much better way to handle it is to apply illumination to the lave tiles in the same way as for caves (but with a lower intensity). I intent to do so in a follow up to this PR.

Hellfire:
Screenshot from 2023-04-23 11-45-52
(a lot of things outside visual range are being lit, this will cause issues as these areas are not consistently updated)

Diablo:
Screenshot from 2023-04-23 12-08-40
(notice the bright candles) 

PR:
Screenshot from 2023-04-23 11-52-13

The X axis is distance from light, with left being at the light source. Y is the brightness value, with top being complete darkness at the bottom being fully lit.

I have also plotted the 3 decay models.
The X axis is distance from light, with left being at the light source. Y is the brightness value, with top being complete darkness at the bottom being fully lit.
Notice how Hellfires never reach fully dark (15 on the Y axis) for most light radii, only 0 and 1 tile lights strength do so, the rest will cause lighting glitches outside of there set range.

Diablo:
image
Hellfire:
image
PR:
image

### DIFF
--- a/Source/lighting.cpp
+++ b/Source/lighting.cpp
@@ -369,29 +369,26 @@ void MakeLightTable()
 	LoadFileInMem("gendata\\pause.trn", PauseTable);
 
 	// Generate light falloffs ranges
-	if (IsAnyOf(leveltype, DTYPE_NEST, DTYPE_CRYPT)) {
-		for (size_t j = 0; j < NumLightRadiuses; j++) {
-			double fa = (sqrt((double)(16 - j))) / 128;
-			fa *= fa;
-			for (int i = 0; i < 128; i++) {
-				uint8_t color = 15 - static_cast<uint8_t>(fa * (double)((128 - i) * (128 - i)));
-				if (color > 15)
-					color = 0;
-				color -= static_cast<uint8_t>((NumLightRadiuses - j - 1) / 2);
-				if (color > 15)
-					color = 0;
-				LightFalloffs[NumLightRadiuses - j - 1][i] = color;
-			}
-		}
-	} else {
-		for (size_t j = 0; j < NumLightRadiuses; j++) {
-			for (size_t i = 0; i < 128; i++) {
-				if (i > (j + 1) * 8) {
-					LightFalloffs[j][i] = 15;
+	const float maxDarkness = 15;
+	const float maxBrightness = 0;
+	for (size_t radius = 0; radius < NumLightRadiuses; radius++) {
+		size_t maxDistance = (radius + 1) * 8;
+		for (size_t distance = 0; distance < 128; distance++) {
+			if (distance > maxDistance) {
+				LightFalloffs[radius][distance] = 15;
+			} else {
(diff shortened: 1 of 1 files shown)


Source/lighting.cpp
```diff
@@ -369,29 +369,19 @@ void MakeLightTable()
...
-				} else {
-					double fs = (double)15 * i / ((double)8 * (j + 1));
-					LightFalloffs[j][i] = static_cast<uint8_t>(fs + 0.5);
+	for (size_t radius = 0; radius < NumLightRadiuses; radius++) {
+		size_t maxDistance = (radius + 1) * 8;
+		for (size_t distance = 0; distance < 128; distance++) {
+			if (distance > maxDistance) {
+				LightFalloffs[radius][distance] = 15;
+			} else {
+				double factor = (double)distance / (double)maxDistance;
```
glebm, line 378: `double` seems like overkill, `float` hopefully results in the same values?

```suggestion
				const float factor = static_cast<float>(distance) / maxDistance;
```

Source/lighting.cpp
```diff
@@ -369,29 +369,19 @@ void MakeLightTable()
...
+				double scaled = factor * 15.0F;
```
glebm, line 379:
```suggestion
				float scaled = factor * 15.0F;
```

Source/lighting.cpp
```diff
@@ -369,29 +369,19 @@ void MakeLightTable()
...
+				if (IsAnyOf(leveltype, DTYPE_NEST, DTYPE_CRYPT)) {
+					int maxIntensity = radius * 1.45;
```
glebm, line 381:
```suggestion
					const int maxIntensity = radius * 1.45;
```

Source/lighting.cpp
```diff
@@ -369,29 +369,19 @@ void MakeLightTable()
...
+					scaled = std::max<double>(factor * maxIntensity + (15.0F - (double)maxIntensity), 0.0F);
```
glebm, line 382:
```suggestion
					scaled = std::max(factor * maxIntensity + (15.0F - maxIntensity), 0.0F);
```

StephenCWills (approved):
Good analysis. It makes way more sense to me in this form.

AJenbo:
Thanks for everyone who helped evaluate this.

PR #6044, opened 2023-04-25
Change Na-Krul's animation rates from 0 to 1

### LINKED ISSUE #6042: Crash when reentering Crypt level 4
Game crashes when reentering crypt level 4.
To reproduce load attached game and go through the town portal.
I can only speculate that the savegame is broken for some reason.

[CRASH.zip](https://github.com/diasurgical/devilutionX/files/11315217/CRASH.zip)

StephenCWills:
TBH, I'm just blindly making this change based on the work done in #2244, and I can only hope Na-Krul wasn't skipped for a reason.

This resolves #6042


obligaron (approved):
lgtm 🙂 
There is no reason why he should have a rate of 0.
I think I simply overlooked it, cause of the `Monster Block end` comment.

Issue #6041, opened 2023-04-24
Crash on player leaving and rejoining


AJenbo:
3 players
Player 1, 2, 3 all on the same level
player 3 leaves the game
player 3 is now standing frozen on screen (getting hit animation)
player 3 rejoins
player 1 and 2 gets an assert error as player 3's graphics is unloaded as he enters town

AJenbo:
Managed to capture a stack trace from another one.

2 players in game
both on lvl 13
enter staris for 14 at the same time
game dies after fading in to level 14

![image](https://user-images.githubusercontent.com/204594/234156720-28a437f5-0fa3-4129-85e1-1bcbbb2a9899.png)

AJenbo:
Player 2 was on lvl 16, plr 1 warped there with TP

![image](https://user-images.githubusercontent.com/204594/234162778-5204580f-2d89-46a3-a405-d67235592876.png)
_pLvlChanging = false
plrlevel = 0

PR #6045, opened 2023-04-25
Move back to an exhaustive clearing of players from the current level

AJenbo:
Related to #6041

To not delay 1.5.0 any further lets move back to fully remove the player from the current level. We can revert this an try to solve the underlying issue after the release.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -2678,23 +2678,17 @@ void StartPlrBlock(Player &player, Direction dir)
 	SetPlayerOld(player);
 }
 
+/**
+ * @todo Figure out why clearing player.position.old sometimes fails
+ */
 void FixPlrWalkTags(const Player &player)
 {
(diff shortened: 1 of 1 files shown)


Source/player.cpp
```diff
@@ -2678,23 +2678,17 @@ void StartPlrBlock(Player &player, Direction dir)
...
-		}
-	}
-
-#ifdef _DEBUG
-	// Checks that FixPlrWalkTags really removes player from all dPlayer tiles
-	// FixPlrWalkTags could fail if player.position.old is not updated correctly
 	for (int y = 0; y < MAXDUNY; y++) {
 		for (int x = 0; x < MAXDUNX; x++) {
-			assert(PlayerAtPosition({ x, y }) != &player);
+			if (PlayerAtPosition({ x, y }) != &player)
```
StephenCWills, line 2688:
```suggestion
			if (PlayerAtPosition({ x, y }) == &player)
```

StephenCWills (approved):

PR #6047, opened 2023-04-25
GitHub actions: Automate i386-linux-gnu release

glebm:
Also updates aarch64 to be basically the same as the i386 one.


AJenbo:
I guess that makes sens as a fare well to 18.04 :)

AJenbo (approved):

PR #6048, opened 2023-04-26
Update to gradle 8.0.0

AJenbo:
Follow up to #6046

### DIFF
--- a/.github/workflows/Android.yml
+++ b/.github/workflows/Android.yml
@@ -27,10 +27,10 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: set up JDK 11
+    - name: set up JDK 17
       uses: actions/setup-java@v3
       with:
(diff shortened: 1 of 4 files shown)


glebm (dismissed):

glebm:
Looks like we need to bump the Java version here to at least 17:
https://github.com/diasurgical/devilutionX/blob/77dda702cf26346e3e27ceae7a19845266f0bfbf/.github/workflows/Android.yml#L30-L35

Might as well go for the latest Java for the CI.

AJenbo:
Doesn't look like they support versions later then 17:
https://github.com/actions/setup-java#supported-version-syntax

I'll try 20 and see if it works any way :P

AJenbo:
20 didn't work, moving to 17

glebm (approved):

PR #6052, opened 2023-04-26
Update Italian localization

M0Rf30:
As in the title,
Regards

### DIFF
--- a/Translations/it.po
+++ b/Translations/it.po
@@ -13,8 +13,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: DevilutionX\n"
-"POT-Creation-Date: 2023-03-06 13:30+0100\n"
-"PO-Revision-Date: 2023-03-06 13:45+0100\n"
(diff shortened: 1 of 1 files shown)


AJenbo:
Good timing

AJenbo:
There's going to be and update to DevilutionX soon so now would be a good time to update the translation if you have time.

M0Rf30:
A guy from devilutionX italian discord contacted me to know if it's possible to localize only the speeches, without involving itemDescriptions or itemNames or monsterNames, only dialogues between characters and things strictly related to the narration.

do you think it is feasible?

Why: Motivation is closely related to deep habits around the names of entities, reported here and there on wikis and communities.

TLDR
filter out entities localizations and only localize narration dialogues

AJenbo:
Hm yeah I think that might be possible. I think it should be a general option not distinctly related to any given language.

Could you create an issue for it?
We should also have an option for not using the voice over translations in the same menu.

PR #6053, opened 2023-04-27
Update Ukrainian TL

tsunamistate:
This PR translates new strings for player inspection and arena pots


qndel:
Finally someone making a commit for updating and a commit for the actual translation ❤️

AJenbo:

Issue #6057, opened 2023-04-28
Arbitrary screenshots like in vanilla


myself600:
### Operating System

Other (please specify)

### DevilutionX version

1.4.1 (latest release)

### Describe

I can't make screenshots using the built-in feature from

- menu
- chat (Enter)
- when removing gold pieces in inventory (right-click)
- loading screen
- cutscenes? (not sure how this was handled in vanilla, as the PrtScr key stops the video)

Reasoning:

This is needed to grab the screen before it is upscaled to 1920x1080

### To Reproduce

In the description

### Additional context

**Operating system**: Manjaro ARM 23.02

### Examples

![Screenshot from 2023-04-28 18-17-17](https://user-images.githubusercontent.com/3542623/235201551-57344fe7-846f-417e-a7fd-b78a059aca61.gif)
Original PCX file - 374,2 KiB (169,3 KiB after recompression with WebP lossless)

![Screenshot_20230428_181757](https://user-images.githubusercontent.com/3542623/235201648-30e96b0d-acc8-442f-a948-6d66d8c9fb06.gif)
Original PNG screenshot captured by the OS - 2,5 MiB (1,9 MiB after recompression with WebP lossless)

AJenbo:
The original did not support taking screen shots during load screens or cut scenes.

AJenbo:
Current implementation does not take keymapper in to account and has some issue with capturing from the menu on some systems.

DakkJaniels:
I fixed the keymapper issue at the very least. Capturing from menu, in terms of having accurate/complete screenshots, sounds like a different issue all together.

AJenbo:
@DakkJaniels can we disable menu screen shots for now?

DakkJaniels:
I think the problem that I haven't worked out yet was the SDL1 part, not the menu screenshots

StephenCWills:
There are still three problems, right?

* `KeymapperOptions::IsTextEntryKey()` is not consistent with the logic in `KeymapperOptions::KeyReleased()`.
* SDL1 doesn't support the same keycodes.
* The menu screenshots are missing some or all of the on-screen visual elements on some platforms.

DakkJaniels:
I'll try and wrap it up at least the first two items this week. The third I'm not really sure about.

AJenbo:
if you can just disable screen shots in the menu that would be good enough :)

StephenCWills:
[...]

PR #6058, opened 2023-04-28
Add print screen to event handler for UI

DakkJaniels:
#6057 

I think this should also let you take a screenshot while in the drop gold menu, but I didn't test that specifically.

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -3,7 +3,6 @@
 #include <algorithm>
 #include <string>
 
-#include "diablo.h"
 #include "DiabloUI/button.h"
 #include "DiabloUI/dialogs.h"
 #include "DiabloUI/scrollbar.h"
@@ -11,6 +10,7 @@
 #include "controls/input.h"
 #include "controls/menu_controls.h"
 #include "controls/plrctrls.h"
+#include "diablo.h"
 #include "discord/discord.h"
 #include "engine/assets.hpp"
 #include "engine/clx_sprite.hpp"
@@ -462,7 +462,7 @@ void UiHandleEvents(SDL_Event *event)
 		return;
 	}
 
-		if (event->type == SDL_KEYUP && event->key.keysym.sym == SDLK_PRINTSCREEN) {
+	if (event->type == SDL_KEYUP && event->key.keysym.sym == SDLK_PRINTSCREEN) {
 		PrintScreen(SDLK_PRINTSCREEN);
 		return;
 	}
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <string>
 
+#include "diablo.h"
 #include "DiabloUI/button.h"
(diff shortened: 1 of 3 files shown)


DakkJaniels:
Now it works while in drop gold menu

myself600:
I still can't make screenshots as described in #6057 with the latest build.

DakkJaniels:
> I still can't make screenshots as described in #6057 with the latest build.

In every single instance you described, or only specific instances? It works for me in menus and the gold drop menu. It was not working in chat.

myself600:
> > I still can't make screenshots as described in #6057 with the latest build.
> 
> In every single instance you described, or only specific instances? It works for me in menus and the gold drop menu. It was not working in chat. Also doesn't work for loading screens and cutscenes, but that is vanilla behavior.

Every single mentioned.

DakkJaniels:
I can't imagine how this is the case if you are on the latest build. I just downloaded the latest build from github and it worked for me.

AJenbo:
> Every single mentioned.

Did you update your build ...
Gold drop:
image

I will complain about one thing thoug. Menu:
image

DakkJaniels:
> I will complain about one thing thoug. Menu:
Is that the resulting screenshot? This was mine with the github build (pcx opened in GIMP).
image

Source/DiabloUI/diabloui.cpp
```diff
@@ -461,6 +462,11 @@ void UiHandleEvents(SDL_Event *event)
 		return;
 	}
 
+	if (event->type == SDL_KEYUP && event->key.keysym.sym == SDLK_PRINTSCREEN) {
```
AJenbo, line 465: Instead of hard coding against SDLK_PRINTSCREEN we should look at what has been bound to "PRINT" in the keymapper.

myself600: I have it bound to F9 as Fn+Print Screen is not registered by DevilutionX as PRINT key

DakkJaniels: > Instead of hard coding against SDLK_PRINTSCREEN we should look at what has been bound to "Screenshot" in the keymapper.

My bad for some reason I thought it adjusted what SDLK_PRINTSCREEN is bound to when it initializes the keys.

myself600:
> Did you update your build ...

I've redownloaded the build from Github Actions (instead of nightly.link) and tried Hellfire also. No, it doesn't work for me.

AJenbo:
I can partially capture some menus. Probably this is related how the driver handles buffering and how rendering is done in the menu. You can see here that I can actually partially capture the Hero selection screens:

image

Source/DiabloUI/diabloui.cpp
AJenbo: Yeah that explains why it's not working for you

DakkJaniels:
> There's no flickering red screen or .pcx files in `~/.local/share/diasurgical/devilution/` when pressing F9 (key bound to screenshots) in menu/chat/gold removal screens

You don't need to keep respond over and over that it doesn't work. It is understood that you rebound the print screen key and the code needs to be adjusted to account for that.

PR #6060, opened 2023-04-28
Fix print screen for different print screen key bindings

DakkJaniels:
Fixes issue with #6058 
Also allows for print screen while in chat.

### DIFF
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -11,6 +11,7 @@
 #include "controls/input.h"
 #include "controls/menu_controls.h"
 #include "controls/plrctrls.h"
+#include "controls/remap_keyboard.h"
 #include "diablo.h"
 #include "discord/discord.h"
 #include "engine/assets.hpp"
@@ -463,8 +464,14 @@ void UiHandleEvents(SDL_Event *event)
 		return;
 	}
 
-	if (event->type == SDL_KEYUP && sgOptions.Keymapper.KeyForAction("Screenshot") == event->key.keysym.sym) {
-		PrintScreen(event->key.keysym.sym);
+	SDL_Keycode vkey = event->key.keysym.sym;
+	remap_keyboard_key(&vkey);
+	bool textKey = sgOptions.Keymapper.IsTextEntryKey(vkey);
+	if (vkey >= SDLK_a && vkey <= SDLK_z)
+		vkey -= 'a' - 'A';
+
+	if (event->type == SDL_KEYUP && sgOptions.Keymapper.KeyForAction("Screenshot") == vkey && (!textInputActive || !textKey)) {
+		PrintScreen(vkey);
 		return;
 	}
 
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -11,7 +11,6 @@
 #include "controls/input.h"
 #include "controls/menu_controls.h"
 #include "controls/plrctrls.h"
-#include "controls/remap_keyboard.h"
 #include "diablo.h"
 #include "discord/discord.h"
 #include "engine/assets.hpp"
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -464,17 +464,6 @@ void UiHandleEvents(SDL_Event *event)
 		return;
 	}
 
-	SDL_Keycode vkey = event->key.keysym.sym;
-	remap_keyboard_key(&vkey);
-	bool textKey = sgOptions.Keymapper.IsTextEntryKey(vkey);
-	if (vkey >= SDLK_a && vkey <= SDLK_z)
-		vkey -= 'a' - 'A';
-
-	if (event->type == SDL_KEYUP && sgOptions.Keymapper.KeyForAction("Screenshot") == vkey && (!textInputActive || !textKey)) {
-		PrintScreen(vkey);
-		return;
-	}
-
 	if (event->type == SDL_KEYDOWN && event->key.keysym.sym == SDLK_RETURN) {
 		const Uint8 *state = SDLC_GetKeyState();
 		if (state[SDLC_KEYSTATE_LALT] != 0 || state[SDLC_KEYSTATE_RALT] != 0) {
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -463,8 +463,8 @@ void UiHandleEvents(SDL_Event *event)
 		return;
 	}
 
-	if (event->type == SDL_KEYUP && event->key.keysym.sym == SDLK_PRINTSCREEN) {
-		PrintScreen(SDLK_PRINTSCREEN);
+	if (event->type == SDL_KEYUP && sgOptions.Keymapper.KeyForAction("Screenshot") == event->key.keysym.sym) {
+		PrintScreen(event->key.keysym.sym);
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


Source/options.cpp
```diff
@@ -1473,7 +1473,7 @@ void KeymapperOptions::KeyReleased(uint32_t key) const
 
 	// Check that the action can be triggered and that the chat textbox is not
 	// open.
-	if (!action.actionReleased || (action.enable && !action.enable()) || talkflag)
+	if (!action.actionReleased || (action.enable && !action.enable()) || ((talkflag || dropGoldFlag) && action.key != "Screenshot"))
```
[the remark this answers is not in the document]
DakkJaniels: is p hardcoded to something somewhere?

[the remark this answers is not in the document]
AJenbo: No but it could cause some issues when trying to write a chat message "ick u the otions before using the ortal"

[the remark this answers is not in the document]
DakkJaniels: 😂 good point. I guess it can be checked against characters that can be typed first to decided whether or not print screen will work. 

But for what it's worth, I did check `p` and it didn't seem to work in main menu while F9 was working. I can look into this more tomorrow.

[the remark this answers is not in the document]
AJenbo: thanks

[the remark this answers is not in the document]
DakkJaniels: I think this is resolved now, feel free to take a look when you get a chance.

Source/options.cpp
```diff
@@ -1465,20 +1465,33 @@ void KeymapperOptions::KeyPressed(uint32_t key) const
...
+	// open. If either of those textboxes are open, only return if the key can be used for entry into the box
+	if (!action.actionReleased || (action.enable && !action.enable()) || ((talkflag && IsTextEntryKey(key)) || (dropGoldFlag && IsNumberEntryKey(key))))
 		return;
 
 	action.actionReleased();
 }
 
+bool KeymapperOptions::IsTextEntryKey(SDL_Keycode vkey) const
+{
+	return ((vkey >= 0 && vkey < 128) || vkey == SDLK_LSHIFT || vkey == SDLK_RSHIFT || vkey == SDLK_KP_BACKSPACE);
```
AJenbo, line 1487: I'm a big ignorant to how this work so just asking. Does this take international symbols in to account or only American keyboards with a-Z?

DakkJaniels: I'm not sure either. Could do the extended ASCII up to 255?

StephenCWills: Basically, it's all of these.
https://github.com/libsdl-org/SDL/blob/ac13ca9ab691e13e8eebe9684740ddcb0d716203/include/SDL_keycode.h#L52-L127

And also the delete key.
https://github.com/libsdl-org/SDL/blob/ac13ca9ab691e13e8eebe9684740ddcb0d716203/include/SDL_keycode.h#L150

Looking at the table, we might want to include caps lock, a bunch more of the `SDLK_KP_*` keys to support numeric keypads, and maybe also the `www` key (just in case?).

As for international symbols, most should be entered using the same keys located in the ASCII range so I think the 0-127 range is probably fine. We need to add `SDLK_RALT` since it seems that "Alt Gr" is used as a modifier to enter those symbols.

I'm not too sure about the more complex input systems, like those used for Japanese symbols. I do know that there are various keys for international keyboards to help enter those types of characters. SDL provides scancodes for these, but not keycodes. So we would likely receive those inputs as `SDLK_UNKNOWN`.

StephenCWills: Oh, and we should probably also include `SDLK_UP` and `SDLK_DOWN` as those can be used to cycle through message history.

AJenbo:
Looks like you have angered the SDL1 builds :(

StephenCWills:
> Looks like you have angered the SDL1 builds :(

@DakkJaniels You can refer to this for the keycodes supported in SDL1.
https://github.com/libsdl-org/SDL-1.2/blob/release-1.2.15/include/SDL_keysym.h

Source/options.cpp
StephenCWills: Looking at this more closely, it looks like you're only applying this function to `KeymapperOptions::KeyReleased()` because the print screen action only acts on key release and not key press. However, there is a similar function that filters (and acts on) key pressed events which should probably be kept consistent.

https://github.com/diasurgical/devilutionX/blob/5a12f7b0d543f7aed528c2d78b36e098f409d3de/Source/control.cpp#L1483-L1510

If we make any changes to better support international symbols or numeric keypads, we should apply them in both places.

StephenCWills:
> I agree that it should be unified, but I think that means moving it to the keymapper, which maybe can be handled separately?

I don't know if it makes sense to unify since they do different things. I just think that if a key is considered a text entry key in one function, the other one should agree.

Source/options.cpp
```diff
@@ -1463,22 +1464,35 @@ void KeymapperOptions::KeyPressed(uint32_t key) const
...
+	// open. If either of those textboxes are open, only return if the key can be used for entry into the box
+	if (!action.actionReleased || (action.enable && !action.enable()) || ((talkflag && IsTextEntryKey(key)) || (dropGoldFlag && IsNumberEntryKey(key))))
 		return;
 
 	action.actionReleased();
 }
 
+bool KeymapperOptions::IsTextEntryKey(SDL_Keycode vkey) const
+{
+	return ((static_cast<Sint32>(vkey) >= 0 && static_cast<Sint32>(vkey) < 128) || vkey == SDLC_KEYSTATE_LEFTSHIFT || vkey == SDLC_KEYSTATE_RIGHTSHIFT || vkey == SDLK_BACKSPACE);
```
StephenCWills, line 1488: `SDLK_BACKSPACE` is 8, so it doesn't need a separate condition. For SDL2, `SDLC_KEYSTATE_LEFTSHIFT` is actually a reference to the scancode, not the keycode. What you want is `SDLK_LSHIFT` which is available in both SDL2 and SDL1.

But you can probably ignore all that because, to remain consistent with `control_presskeys()`, I recommend the following.

```suggestion
	return IsAnyOf(vkey, SDLK_ESCAPE, SDLK_RETURN, SDLK_KP_ENTER, SDLK_BACKSPACE, SDLK_DOWN, SDLK_UP) || (vkey >= SDLK_SPACE && vkey <= SDLK_z);
```

If you want to include left shift and right shift, you should modify `control_presskeys()` as well.

StephenCWills (changes requested):

DakkJaniels:
I don't see why amiga is failing here

Source/DiabloUI/diabloui.cpp
```diff
@@ -11,6 +11,7 @@
 #include "controls/input.h"
 #include "controls/menu_controls.h"
 #include "controls/plrctrls.h"
+#include "controls/remap_keyboard.h"
```
StephenCWills, line 14: This should no longer be necessary without the call to `remap_keyboard_key()`.
```suggestion
```

Source/options.cpp
```diff
@@ -29,6 +29,7 @@
 #include "utils/language.h"
 #include "utils/log.hpp"
 #include "utils/paths.h"
+#include "utils/sdl_compat.h"
```
StephenCWills, line 32: This should no longer be necessary without the references to `SDLC_KEYSTATE_LEFTSHIFT` and `SDLC_KEYSTATE_RIGHTSHIFT`.
```suggestion
```

StephenCWills (changes requested):

StephenCWills (approved):

StephenCWills:
> I don't see why amiga is failing here Looks like it's a different issue, I see #6142 also failed to build on amiga, looks like similar errors.

Surely it's unrelated to this PR. Looks like the Docker image was updated about 25 hours ago.

AJenbo:
tanks

PR #6063, opened 2023-04-29
Correct OOB check in PrepareInnerBorders()

### LINKED ISSUE #6061: Stuck in PrepareInnerBorders() when entering level 14
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

Game crashes when entering level 14

here is the save file just before entering
[save.zip](https://github.com/diasurgical/devilutionX/files/11358811/save.zip)

### To Reproduce

Click on the Entrance of the level 14

### Expected Behavior

_No response_

### Additional context

The issue also occurs in the latest master version

AJenbo:
Fixes #6061

The old check was wrong because the second check is inverted.

Issue introduced in a1b7b8522a528cc5fd309b4295bbb3aa255bea30

Thanks to @obligaron for helping with debugging this

This caused the following level mask to send the next part to become an infinit loop
```
00000000000000000000
00000000000000000000
00000000111111000000
00011111111111111111
00011111111111111111
00000110111111001100
00000111111110001100
00000111111110001100
00000110000000001100
00011111111110001100
00011111111110001100
00000000011110001100
00111100011110001100
00111100011110001100
00111100011110001100
00111111111111001100
00001111111111001100
00001111111111001100
00000011111111000000
00000000000000000000
```


AJenbo:
Taking the thumbs up and test to be the same as an approval in this case

AJenbo:
Here's the mas overlayed on the quardrant:

image

The se area will be reworked somewhat for the final version of the level so i had to manually revert that part.

Here is the final level layout:

image

PR #6070, opened 2023-05-02
Use SDL_SetWindowDisplayMode() to resize fullscreen windows

### LINKED ISSUE #6059: Switching between upscaling and not upscaling requires restart to take effect
If you go to the menu and switch the state of upscaling the game will no longer change the resolution and instead end up rendering at the wrong resolution.
For example if you run with upscaling enabled on a 1920x1080 monitor with a 480p game resolution the, the game will render at 1080p after switching upscaling off.
After restarting the game it will correctly switch the screen resolution.

Replicate:
1: go to settings -> graphics
2: toggle upscaling
3: notice how the game is now running a mix between the old and descried state

Extra info:
As an alternative to restarting the app you can also switch between full screen and windowed

StephenCWills:
If you use `Alt+Enter` to toggle fullscreen while browsing any settings menu type besides `ShownMenuType::Settings`, the `FullscreenChanged()` handler will attempt to access an item in the empty `vecOptions` vector. The first commit fixes that.

As for the rest, it seems that `SDL_SetWindowDisplayMode()` is the appropriate function to call when resizing a window with the `SDL_WINDOW_FULLSCREEN` flag set. This required some additional logic in both `ResizeWindow()` and `SetFullscreenMode()`. Hopefully the comments help to explain a lot of the nuance better than I can here.

Fair warning: I only tested this on Windows.

This resolves #6059

### DIFF
--- a/Source/DiabloUI/settingsmenu.cpp
+++ b/Source/DiabloUI/settingsmenu.cpp
@@ -318,7 +318,7 @@ void FullscreenChanged()
 
 	for (auto &vecItem : vecDialogItems) {
 		int vecItemValue = vecItem->m_value;
-		if (vecItemValue < 0)
+		if (vecItemValue < 0 || vecItemValue >= vecOptions.size())
 			continue;
 
 		auto *pOption = vecOptions[vecItemValue];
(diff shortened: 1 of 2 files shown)


glebm (approved):

StephenCWills:
I don't have that problem on Windows.

AJenbo:
It only happens when up-scaling is disabled, ie. "real" fullscreen is in use.

PR #6071, opened 2023-05-02
Play Pepin's quest intro dialog if the player skips the intro

ephphatha:
If the player discovers and enters the poison water set level before talking to Pepin the poison water quest does not appear in the quest log and his first set of dialog doesn't play. Handle this case specifically to ensure the player gets the message and can see progress.

### DIFF
--- a/Source/quests.cpp
+++ b/Source/quests.cpp
@@ -355,13 +355,15 @@ void CheckQuests()
 	}
 
 	if (setlevel) {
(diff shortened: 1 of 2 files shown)


glebm (approved):

ephphatha:
Sorry, just realised there was still a way to skip having the log entry appear.

glebm (approved):

PR #6081, opened 2023-05-06
Remove C++11 compatibility code

glebm:
With this, we will require C++17 at the minimum (except for std::filesystem which is still optional)
Let's see if anything other than Amiga breaks


glebm:
Looks like it's just the Amiga!

PR #6083, opened 2023-05-06
Overhaul version handling

glebm:
Stop using git for the version tag.
The new process is as follows:

1. The `VERSION` file contains the version. For the upcoming release it should be set to "1.5.0". Immediately after the release, it should be changed to "1.6.0-dev". The post release 1.5 branch should use "1.5.1-dev".
2. The build type is included in the version string unless it is "Release". The commit hash is now included into the version string whenever the `VERSION` file has a suffix (i.e. 1.5.0-dev rather than just 1.5.0). The full version string for development builds now looks like this: `1.5.0-dev-Debug-b102dccf8`.

This means we longer need git or `-DVERSION_NUM` to get a sensible version number.

The in-development versions are now always clearly indicated with `-dev` no matter how DevilutionX was built.


AJenbo:
A bit less automatically then I originally hoped when we where started using git for versions. But a lot less error prone 👍

AJenbo (dismissed):

AJenbo (approved):

PR #6090, opened 2023-05-07
Cleanup random.cpp and add function for cases where random values are discarded

ephphatha:
Extracted from #6079 since this is a no impact change that hopefully makes the code easier to reason about. The helper to discard items gets rid of about half the uses of `AdvanceRndSeed()` as well.

### DIFF
--- a/Source/engine/random.hpp
+++ b/Source/engine/random.hpp
@@ -30,6 +30,11 @@ void SetRndSeed(uint32_t seed);
 uint32_t GetLCGEngineState();
 
 /**
+ * @brief Advances the global RandomNumberEngine state and returns the new value
+ */
+uint32_t GenerateSeed();
+
+/**
  * @brief Generates a random non-negative integer (most of the time) using the vanilla RNG
  *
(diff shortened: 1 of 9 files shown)


Source/engine/random.hpp
```diff
@@ -29,6 +29,19 @@ void SetRndSeed(uint32_t seed);
  */
 uint32_t GetLCGEngineState();
 
+/**
+ * @brief Advance the global RandomNumberEngine state by the specified number of rounds
+ *
+ * Only used to maintain vanilla compatibility until logic requiring reproducable random number generation is isolated.
+ * @param count How many values to discard
+ */
+void DiscardRandomValues(unsigned count);
```
AJenbo, line 38: i guess we should markAdvanceRndSeed as nodiscard at this point

glebm (approved):

Source/engine/random.hpp
ephphatha: Want me to do that in this PR? Easy enough to rewrite the second commit :D

AJenbo: if you have time :)

PR #6092, opened 2023-05-07
Automate macOS x86_64 release

glebm:
Also cleans up CPack files in other workflows after building (to avoid saving them to cache)


.github/workflows/macOS_x86_64.yml
```diff
@@ -0,0 +1,67 @@
...
+      - '*.md'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: macos-latest
```
AJenbo, line 25: I'm wondering if we should use older builds for compat on release

AJenbo (approved):

.github/workflows/macOS_x86_64.yml
glebm: That might be a good idea

PR #6095, opened 2023-05-10
Add missing <cstdint> includes

glebm:
Done with the following script:

```ruby
Dir["Source/**/*.{h,c,cc,cpp,hpp}"].each do |path|
  v = File.read(path)
  next if !v.include?("uint32_t") || v.include?("cstdint")

  lines = v.lines
  line_num = if lines[2].start_with?(" *")
    lines.index { |l| l.start_with?(" */") } + 3
  else
    3
  end

  lines.insert(line_num, "#include <cstdint>\n")

  File.write(path, lines.join(""))
end
```

then fixed-up manually

Refs https://github.com/bebbo/gcc/issues/201#issuecomment-1541884125

### DIFF
--- a/Source/levels/drlg_l3.cpp
+++ b/Source/levels/drlg_l3.cpp
@@ -1,4 +1,7 @@
+#include "levels/drlg_l3.h"
+
 #include <algorithm>
+#include <cstdint>
 
 #include "engine/load_file.hpp"
 #include "engine/points_in_rectangle_range.hpp"
--- a/Source/DiabloUI/diabloui.cpp
+++ b/Source/DiabloUI/diabloui.cpp
@@ -1,6 +1,7 @@
 #include "DiabloUI/diabloui.h"
 
 #include <algorithm>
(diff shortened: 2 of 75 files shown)


Source/levels/drlg_l3.cpp
```diff
@@ -1,4 +1,7 @@
+#include "levels/drlg_l3.h"
+
 #include <algorithm>
+#include <cstdint>
```
AJenbo, line 4: Since it is included in the drlg_l3.h should it still be included here?

AJenbo (approved):

Source/levels/drlg_l3.cpp
glebm: Yes, the header includes may change in the future and we don't want the to break the cc code

AJenbo: Ok, was just thinking that it would be a way to have them change in the same way instead of having to change it back and forth individually.

PR #6100, opened 2023-05-11
Correct use of local instead of global lightTableIndex

AJenbo:
It looks like this was mistakenly using LightTableIndex instead of lightTableIndex. Found by @qndel


AJenbo:
Qndel privatly approved

PR #6101, opened 2023-05-11
debug path command

qndel:
image

### DIFF
--- a/Source/engine/render/dun_render.hpp
+++ b/Source/engine/render/dun_render.hpp
@@ -262,11 +262,10 @@ string_view MaskTypeToString(MaskType maskType);
  * @param position Target buffer coordinates
  * @param levelCelBlock The MIN block of the level CEL file.
  * @param maskType The mask to use,
- * @param lightTableIndex The light level to use for rendering (index into LightTables).
- * @param tbl Optional TRN for a tile.
+ * @param tbl LightTable or TRN for a tile.
  */
 void RenderTile(const Surface &out, Point position,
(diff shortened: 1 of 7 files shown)


ephphatha (approved):
really useful command, thanks :D

Source/engine/render/scrollrt.cpp
```diff
@@ -505,6 +505,12 @@ void DrawCell(const Surface &out, Point tilePosition, Point targetBufferPosition
 	const uint16_t levelPieceId = dPiece[tilePosition.x][tilePosition.y];
 	const MICROS *pMap = &DPieceMicros[levelPieceId];
 
+	uint8_t *tbl = nullptr;
```
glebm, line 508:
```suggestion
	const uint8_t *tbl = nullptr;
```

Source/engine/render/dun_render.hpp
```diff
@@ -263,9 +263,10 @@ string_view MaskTypeToString(MaskType maskType);
  * @param levelCelBlock The MIN block of the level CEL file.
  * @param maskType The mask to use,
  * @param lightTableIndex The light level to use for rendering (index into LightTables).
+ * @param tbl Optional TRN for a tile.
```
glebm, line 266: Perhaps we should remove the `lightTableIndex` and simply always pass a non-null `tbl`?

AJenbo:
could you block it out with ifdef so it only affects debug builds

glebm (approved):

PR #6104, opened 2023-05-13
Fix monster light offset after walk animation ends

### LINKED ISSUE #6102: Monster light source lags behind 1 tile when walking southern directions
### Operating System

Windows x64

### DevilutionX version

Other (please specify version number)

### Describe

Started happening after the light code cleanup I believe. Load this save, do nothing, and see issue.
[single_3.zip](https://github.com/diasurgical/devilutionX/files/11456876/single_3.zip)

### To Reproduce

1. Load savefile

### Expected Behavior

_No response_

### Additional context

_No response_

StephenCWills:
Looks like, before 2493f06, `monster.position.offset` was used to determine the offset for the light. A few functions such as `M_StartStand()` and `MonsterDeath()` would set `monster.position.offset = { 0, 0 }` to prevent stale values from offsetting the position of the monster's light. After that commit, we just always calculate the walking offset based on the current animation info, regardless of the type of animation is being used. This means that calling `M_StartStand()` at the end of the walking animation followed by `SyncLightOffset()` would result in the light being placed as if the start of the stand animation was actually the start of the walking animation, causing the light to be placed at the wrong offset.

This resolves #6102

### DIFF
--- a/Source/monster.cpp
+++ b/Source/monster.cpp
@@ -1016,7 +1016,7 @@ void SyncLightPosition(Monster &monster)
 	if (monster.lightId == NO_LIGHT)
 		return;
 
-	const WorldTileDisplacement offset = monster.position.CalculateWalkingOffset(monster.direction, monster.animInfo);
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Nice work figuring this out

PR #6112, opened 2023-05-17
GitHub Actions: Automate Android release

glebm:


AJenbo:
I don't think there is a point to doing this. I need to build the aab package and sign it with my publisher key in any case, and then Google will generate a signed apk that we can download and upload to the release section for the few people that prefer manually downloading an apk.

glebm:
Ah OK makes sense

Issue #4668, opened 2022-06-05
Walking animation stuck


qndel:
It's possible to get character stuck in walking animation
![walkinplacedevx](https://user-images.githubusercontent.com/14297035/172058890-400d04e7-0ad2-431f-99d8-8e83ac5e6d74.gif)
![walkinplacedevx](https://user-images.githubusercontent.com/14297035/172059037-c6465ec8-1de5-4c10-8f5a-6aa86c0015cf.gif)

I tried to reproduce it with town NPCs but failed so probably can only be done in dungeon.

To reproduce:
1. Find a quest monster you can talk to (I tested on Gharbad, no idea if it works on other monsters)
2. Move next to it and queue next movement and talk to monster
3. Your walking animation should get stuck

it might take a few tries to pull it off :D

PR #6115, opened 2023-05-19
fix walking animation getting stuck

qndel:
Fixes https://github.com/diasurgical/devilutionX/issues/4668
Walking seems to work fine, can't get it stuck anymore, would appreciate if someone else looked at it too
Looks like StartStand does **everything** that StartWalkStand did and more.

### DIFF
--- a/Source/player.cpp
+++ b/Source/player.cpp
@@ -191,16 +191,6 @@ void ClearStateVariables(Player &player)
 	player.queuedSpell.spellLevel = 0;
 }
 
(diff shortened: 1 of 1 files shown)


obligaron (approved):
lgtm
I tested that it doesn't mess with animations.
Also I like the change, because I dislike when shown animation and player mode diverge.

Another interesting question is: how can the game calculated a walk path that is invalid? This could be another possible fix for the issue.

But this PR is still valid and fine. 🙂

Issue #5618, opened 2022-12-17
Big orange cursor


qndel:
### Operating System

Windows x64

### DevilutionX version

Custom build (please specify commit ID)

### Describe

Commit https://github.com/diasurgical/devilutionX/commit/96c7b154be835bddcd80419163af2680a2944ec5
https://cdn.discordapp.com/attachments/702267057772822590/1053712799882281091/VID_20221217_173751.mp4

### To Reproduce

Go to settings/support screen
Go back
Cursor is bigger and orange for like 1 frame

### Additional context

Default settings

[Graphics]
Width=640
Height=480
Fullscreen=0
Fit to Screen=1
Upscale=1
Scaling Quality=2
Integer Scaling=0
Vertical Sync=1
Gamma Correction=100
Zoom=0
FPS Limiter=1
Show FPS=0
Show health values=0
Show mana values=0
Color Cycling=1
Alternate nest art=0
Hardware Cursor=1
Hardware Cursor For Items=0
Hardware Cursor Maximum Size=128

kphoenix137:
Does this bug still occur?

qndel:
> Does this bug still occur?

I think I was the only one that had it happen and it seems it no longer happens

PR #6121, opened 2023-05-20
fix big orange hw cursor

qndel:
Fixes https://github.com/diasurgical/devilutionX/issues/5618

I have no clue what's going on here 🤷 but everything works for me with this change, both in menu and in game


qndel:
Works fine in UI after this change but breaks after entering the game

glebm:
This actually just makes it use software cursor in the menu, so converting it back to draft for now

AJenbo:
Looks like this is better solved elsewhere

PR #6126, opened 2023-05-21
Delay reinitializing hwcursor until fade 1

glebm:
With fade 0 the cursor is never visible because everything is black.

While this is a very minor optimization, it works around
one of the instances of a bug in certain drivers as seen in https://github.com/diasurgical/devilutionX/issues/5618.


qndel:
Fixes https://github.com/diasurgical/devilutionX/issues/5618
Great job!

qndel:
with this PR it's like this:
fullscreen = broken
windowed small window = broken
windowed maximized = works

without PR it's broken in all cases so an improvement I guess 😅

AJenbo (approved):

PR #6135, opened 2023-05-22
Use _talker_id when initialising towners to match use when gossiping.

### LINKED ISSUE #6132: some gossips got mixed up
If you click on the npcs and want to gossip, the conversations of other npcs are often played back. For example. Ogden speaks as Cain, Cain speaks as Farnham. Not all npcs and no causing commit tested.

ephphatha:
fixes #6132

### DIFF
--- a/Source/towners.cpp
+++ b/Source/towners.cpp
@@ -39,15 +39,20 @@ void NewTownerAnim(Towner &towner, ClxSpriteList sprites, uint8_t numFrames, int
 	towner._tAnimDelay = delay;
 }
 
-void InitTownerInfo(int i, const TownerData &townerData)
+void InitTownerInfo(const TownerData &townerData)
 {
-	auto &towner = Towners[i];
+	// _talker_id is used to select the active towner when selecting store dialog options, if the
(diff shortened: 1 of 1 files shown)


ephphatha:
This does introduce potential for broken behaviour. If someone saves a game while they're in the process of walking towards a town NPC to talk to them in a previous version then loads the game in this version `player.destParam1` will be off-by-one and they'll get the menu/dialog for a different NPC. Quitting the menu and talking to the NPC again will show the expected dialog options.

The previous broken behaviour was seeding the global RNG with unintialised memory.

AJenbo:
By this version are you refering to the curent development (mater) builds? If so they are not supported as stable releases.

ephphatha:
any build after this is merged. Pepin and Griswold are always assigned ID 1 and 2 in `dMonster`, in 1.09 and every devX release all other towners get shifted back one if deadguy isn't available (Cain can be ID 4 or 5 for example). Opened #6136 as a backwards compatible approach, I think there was some code which used to do that approach but doesn't appear it was ever used for gossip?

AJenbo:
yeah there is something like that in other places, but i forgot for what :)

PR #6136, opened 2023-05-22
Map from _talker_id to towner pointer when using gossip option

ephphatha:
Backwards compatible fix for #6132

### DIFF
--- a/Source/stores.cpp
+++ b/Source/stores.cpp
@@ -2026,7 +2026,9 @@ void TalkEnter()
 	}
 
 	if (stextsel == sn - 2) {
-		InitQTextMsg(Towners[talker].gossip);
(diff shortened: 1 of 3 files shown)


Source/stores.cpp
```diff
@@ -2026,7 +2026,9 @@ void TalkEnter()
 	}
 
 	if (stextsel == sn - 2) {
-		InitQTextMsg(Towners[talker].gossip);
+		// Assumed we'll always have a valid towner here.
+		Towner *target = GetTowner(talker);
```
AJenbo, line 2030:
```suggestion
		Towner *target = GetTowner(talker);
		assert(target != null_ptr);
```

AJenbo:
Thanks :)

PR #6139, opened 2023-05-22
Fixes spawned items from Zhar and Gharbard

### LINKED ISSUE #5979: Zhar's first book does not sync when dropped
When player 1 talks to Zhar the dropped book cannot be picked up by player 2

obligaron:
Fixes #5979

Notes

- In master spawning logic for both happens in `MonsterTalk` and this runs for both clients
- But the spawning logic uses `CMD_DROPITEM` that should only be send by one player (the "source" of the action)
- Also the used seed for the spawned items can differ between clients (and therefore also the item)
- This PR uses only `CMD_SPAWNITEM` that doesn't depend on the same logic (and same seed) is executed on both clients
- `CMD_SPAWNITEM` is only send by the player that talks to the quest monster
- The item spawning logic should be refactored at some point (we currently have `sendmsg`, `delta` and now `spawn`)

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -1555,7 +1555,7 @@ void SetupAllItems(const Player &player, Item &item, _item_indexes idx, uint32_t
 	SetupItem(item);
 }
 
-void SetupBaseItem(Point position, _item_indexes idx, bool onlygood, bool sendmsg, bool delta)
+void SetupBaseItem(Point position, _item_indexes idx, bool onlygood, bool sendmsg, bool delta, bool spawn = false)
 {
 	if (ActiveItemCount >= MAXITEMS)
 		return;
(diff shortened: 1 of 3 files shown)


Source/items.cpp
```diff
@@ -3226,7 +3228,7 @@ Item *SpawnUnique(_unique_items uid, Point position, bool sendmsg /*= true*/)
 	return &item;
 }
 
-void SpawnItem(Monster &monster, Point position, bool sendmsg)
+void SpawnItem(Monster &monster, Point position, bool sendmsg, bool spawn /*= false*/)
```
AJenbo, line 3231: SpawnItem() having an option not to spawn confused me, but it's only related to spawning on other clients now that i read it while awake.

AJenbo (approved):

Source/items.cpp
obligaron: It's still confusing that's why I mentioned a possible refactoring in the notes 🙂

PR #6140, opened 2023-05-22
Sync break crux correctly

### LINKED ISSUE #6022: MP: Skeleton cross hit by monsters are not synced
When players interact with the 3 skeleton crosses in the tomb it is correctly synced, but if a monster triggers one (stray arrow) only one client will notice the change leading to the object and level (once all 3 have been triggered) being out of sync.

obligaron:
Fixes #6022

Notes:
- SyncBreakObj was missing handling for crux
- Simplified logic for breaking crux. Now all missiles will sync breaking cruxes, but multiple events (from different clients) will be ignored
- Would be great if someone else could test it, because it's hard to get a small desync alone 😉

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -3497,8 +3497,11 @@ bool AreAllCruxesOfTypeBroken(int cruxType)
 	return true;
 }
 
-void BreakCrux(const Player *player, Object &crux)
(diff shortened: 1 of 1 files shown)


AJenbo (approved):
Will keep and eye on it during the next play test :)

PR #6148, opened 2023-05-25
Gnat Sting Fix

wkdgmr:
This commit fixes an issue with Gnat Sting that causes the bow to only fire single shots if the player attacks too quickly.  There is nothing in the description of the bow that indicates that this was intended.  It seems like this was coded in a way to properly play sound effects for both the single shot and side shots, but had the unintended consequence of causing Gnat Sting to override itself if fired too quickly, causing the bow to only fire single shots instead of multiple shots as it is intended to do.


StephenCWills:
Rejecting based on established precedent: #5598 and #5595.

PR #6151, opened 2023-05-26
Translation spanish - Improvement and simplification of town strings

thiagol-fenix:
- Corrections in some text strings.
- Simplification of pnj interactions to approach diablo 2 style.


AJenbo (approved):
Please also translate the few new strings

PR #6167, opened 2023-06-02
Spawn Optic Amulet and Arkaine's Valor always at the same tile

### LINKED ISSUE #6165: mp: can't compleat quest that requires picking up unique items
### Operating System

Windows x64

### DevilutionX version

1.4.1 (latest release)

### Describe

Since the unique item drop randomly it not possible to complete quest that require picking up the unique item.
- Valor
- Halls of the blind

### To Reproduce

start mp game, play the two quests, check quest log

### Expected Behavior

_No response_

### Additional context

Suggest triggering quest completion in MP when spawning the two items

obligaron:
Fixes #6165

## Notes
- In master the position of Optic Amulet and Arkaine's Valor is slightly randomized (+/- 1 Tile). This PR changes that so both items gets always spawned at the same (relative) location.
- In Multiplayer to complete Halls of Blind and Arkaine's Valor Quests also checks these exact drop positions.
- The old check (for quest item type) could be removed and always the new logic can be used (check position). But I tried to minimize the changes (and risks) so close to release of 1.5.0 version.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -867,7 +867,9 @@ void CheckQuestItem(Player &player, Item &questItem)
 {
 	Player &myPlayer = *MyPlayer;
 
-	if (questItem.IDidx == IDI_OPTAMULET && Quests[Q_BLIND]._qactive == QUEST_ACTIVE) {
+	if (Quests[Q_BLIND]._qactive == QUEST_ACTIVE
+	    && (questItem.IDidx == IDI_OPTAMULET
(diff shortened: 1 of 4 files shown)


AJenbo:
Good solution :+1:

obligaron:
I didn't thought about that edge-case.
But it's not that likely. The time window is short where this would be affected.
On the other hand the added code isn't that complex, so better be safe then sorry.

PR #6169, opened 2023-06-02
Drop quest item based on difficulty

AJenbo:


obligaron (approved):
I think this is a good solution. 🙂 
For new players they get the same experience as in vanilla (nostalgic) and it's compatible with what all websites state about the quests.
For people that grind they get more interesting quests to replay.

Thanks @ikonomov and @AJenbo

ikonomov:
Do you think it might be a good idea to have some chance to get the quest rewards in Nightmare and Hell as well?  I was thinking if somebody was to do a playthrough (without having all quests enabled setting) and decide to go into Nightmare after killing Diablo on Normal they will not be able to find all the quest items.  What about giving those quest items some chance to be able to drop even on NM and Hell, say 10% or something similar?
`if (sgGameInitInfo.nDifficulty == DIFF_NORMAL || FlipCoin(10))`

AJenbo:
no

ikonomov:
Yeah, I think this is probably the right call.  I wanted to have both vanilla's mechanic as well as have the random idea by obligaron, but I think it makes more sense to have it one way or the other.

AJenbo:
well we do, it's on NM and HL, mine was replacing unique for quest

PR #6175, opened 2023-06-03
Use `std::from_chars` instead of `std::stoi`

glebm:
Adds a `ParseInt` function which uses `std::from_chars`.

From `std::from_chars` documentation:
> Unlike other parsing functions in C++ and C libraries,
> std::from_chars is locale-independent, non-allocating, and non-throwing.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -48,6 +48,7 @@
 #include "utils/format_int.hpp"
 #include "utils/language.h"
 #include "utils/log.hpp"
+#include "utils/parse_int.hpp"
 #include "utils/sdl_geometry.h"
 #include "utils/stdcompat/optional.hpp"
 #include "utils/str_case.hpp"
@@ -388,9 +389,9 @@ std::string TextCmdArena(const string_view parameter)
 		return ret;
 	}
 
-	int arenaNumber = atoi(parameter.data());
-	_setlevels arenaLevel = static_cast<_setlevels>(arenaNumber - 1 + SL_FIRST_ARENA);
-	if (arenaNumber < 0 || !IsArenaLevel(arenaLevel)) {
+	const ParseIntResult<int> parsedParam = ParseInt<int>(parameter, /*min=*/0);
+	const _setlevels arenaLevel = parsedParam.ok() ? static_cast<_setlevels>(parsedParam.value - 1 + SL_FIRST_ARENA) : _setlevels::SL_NONE;
+	if (!IsArenaLevel(arenaLevel)) {
 		StrAppend(ret, _("Invalid arena-number. Valid numbers are:"));
 		AppendArenaOverview(ret);
 		return ret;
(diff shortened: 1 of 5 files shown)


Source/debug.cpp
```diff
@@ -505,7 +532,8 @@ std::string DebugCmdQuest(const string_view parameter)
 
 std::string DebugCmdLevelUp(const string_view parameter)
 {
-	int levels = std::max(1, atoi(parameter.data()));
+	const ParseIntResult<int> parsedArg = ParseInt<int>(parameter, /*min=*/1);
+	const int levels = parsedArg.ok() ? parsedArg.value : 1;
```
ephphatha, line 536: `ParsedIntResult<T>::value_or(T value_if_error)` would be a great helper for cases like this, matches https://en.cppreference.com/w/cpp/utility/expected/value_or if c++23 support becomes a possibility :D

glebm: Done! We might as well add `tl::expected` once we've switched to C++17

Source/control.cpp
```diff
@@ -411,10 +412,15 @@ std::string TextCmdArenaPot(const string_view parameter)
 		StrAppend(ret, _("Arenas are only supported in multiplayer."));
 		return ret;
 	}
+	int numPots = 1;
+	const ParseIntResult<int> parsedParam = ParseInt<int>(parameter, /*min=*/1);
+	if (parsedParam.ok()) {
+		numPots = parsedParam.value;
+	}
```
ephphatha, line 419:
```suggestion
	int numPots = ParseInt<int>(parameter, /*min=*/1).value_or(1);
```

Source/debug.cpp
```diff
@@ -191,8 +192,11 @@ std::string DebugCmdTakeGoldCheat(const string_view parameter)
...
 {
 	Player &myPlayer = *MyPlayer;
-	auto level = atoi(parameter.data());
-	if (level < 0 || level > (gbIsHellfire ? 24 : 16))
+	const ParseIntResult<int> parsedParam = ParseInt<int>(parameter, /*min=*/0);
+	if (!parsedParam.ok())
+		return "Specify the level number";
+	const int level = parsedParam.value;
+	if (level > (gbIsHellfire ? 24 : 16))
 		return StrCat("Level ", level, " is not known. Do you want to write a mod?");
```
ephphatha, line 200: This one is getting into the weeds a bit but given status allows discriminating between failure cases you can keep the old behaviour of returning one message for out-of-range values while also putting a distinct message for missing/not a number values.
```suggestion
	const ParseIntResult<int> parsedParam = ParseInt<int>(parameter, /*min=*/0, /*max=*/(gbIsHellfire ? 24 : 16));
	if (!parsedParam.ok()) {
		switch (parsedParam.status) {
		case ParseIntStatus::ParseError:
			return "Specify the level number";
		case ParseIntStatus::OutOfRange:
			return StrCat("Level ", level, " is not known. Do you want to write a mod?");
		}
	}
	const int level = parsedParam.value;
```

There's quite a few places where the value is later checked if it's in bounds which could use a similar approach. I'd leave this for a later refactor (I'm not really happy about using a switch for this...), it'd be a nice activity if `ParseIntResult` is made more like a `std::expected` specialisation.

ephphatha (commented):
One other potential use of `value_or()`, the second comment is just a thought about error handling that doesn't warrant changing now. Consider this approval with a chance to accept the first suggestion :D

Source/debug.cpp
glebm: This is a non-exhaustive switch, so it will result a warning.
Let's keep further improvements for a separate PR!

glebm:
Rebased

glebm:
@AJenbo Rebased on top of development and removed pre-C++17 fallback

PR #6176, opened 2023-06-03
vcpkg.json: Add builtin-baseline

glebm:
This field is now required, otherwise we get this error:

> error: this vcpkg instance requires a manifest with a specified baseline in order to interact with ports.
> Please add 'builtin-baseline' to the manifest or add a 'vcpkg-configuration.json' that redefines the default registry.

The hash here is a commit hash from https://github.com/microsoft/vcpkg/commits/master

https://learn.microsoft.com/en-us/vcpkg/users/examples/versioning.getting-started#builtin-baseline

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -6,6 +6,7 @@
         "bzip2",
         "simpleini"
     ],
+    "builtin-baseline": "662dbb50e63af15baa2909b7eac5b1b87e86a0aa",
     "features": {
         "sdl1": {
             "description": "Use SDL1.2 instead of SDL2",


AJenbo (approved):

StephenCWills:
FYI, this requires that the baseline commit be fetched in order to build the project. Anyone receiving the following error needs to run `git fetch` in the directory where they installed vcpkg.

```
1> [CMake] -- Running vcpkg install
1> [CMake] error: while checking out baseline from commit '662dbb50e63af15baa2909b7eac5b1b87e86a0aa', failed to `git show` versions/baseline.json. This may be fixed by fetching commits with `git fetch`.
1> [CMake] error: git failed with exit code: (128).
1> [CMake] fatal: path 'versions/baseline.json' exists on disk, but not in '662dbb50e63af15baa2909b7eac5b1b87e86a0aa'
1> [CMake] 
1> [CMake] You can use the current commit as a baseline, which is:
1> [CMake] 	"builtin-baseline": "582a7ef7a0dc5cdda5c015e98f57c97218e18c32"
1> [CMake] -- Running vcpkg install - failed
```

glebm:
Was this not the right fix to the CI build failure?
Should we do something else instead?

StephenCWills:
We could have used an older commit, but there's no telling whether anyone's been keeping their vcpkg up to date. AFAIK, this is about as good as anything we could have done I think.

PR #6177, opened 2023-06-03
Remove uses of `std::tolower`

glebm:
`std::tolower` relies on the C locale, which we try to avoid.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -161,6 +161,7 @@ set(libdevilutionx_SRCS
   utils/sdl_bilinear_scale.cpp
   utils/sdl_thread.cpp
   utils/str_cat.cpp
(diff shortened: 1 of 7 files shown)


StephenCWills (approved):

StephenCWills:
I guess the only thing that might be cause for concern would be diacritics in player names. However, we've already started discussing some of the issues with using player names in the `/inspect` command, and expanding the scope of `AsciiStrToLower()` to include non-ASCII characters wouldn't fix it anyway.

AJenbo:
isn't MP names limited to ASCII?

StephenCWills:
Looks like you are correct. I didn't realize that.

AJenbo:
it's one of the reasons i wasn't so bothered by /inspect taking a name :)

PR #6178, opened 2023-06-03
Empty cursor when closing inventory

### LINKED ISSUE #4205: closing stash while holding an item

AJenbo:
Fixes #4205

An alternative to #4226

When closing the stash while an item is held we now first try to drop it, place it in the belt, or inventory.
If that fails we assume that we can stick it in the stash since it's huge.

The item put sound is played to let the player know what happened.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -2081,8 +2081,35 @@ bool UseInvItem(int cii)
 void CloseInventory()
 {
 	CloseGoldWithdraw();
-	IsStashOpen = false;
+	CloseStash();
 	invflag = false;
+}
+
+void CloseStash()
+{
+	if (!IsStashOpen)
+		return;
+
+	Player &myPlayer = *MyPlayer;
+	if (!myPlayer.HoldItem.isEmpty()) {
+		std::optional<Point> itemTile = FindAdjacentPositionForItem(myPlayer.position.future, myPlayer._pdir);
+		if (itemTile) {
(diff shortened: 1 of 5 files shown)


Source/inv.cpp
```diff
@@ -2081,10 +2081,37 @@ bool UseInvItem(int cii)
...
+	if (!myPlayer.HoldItem.isEmpty()) {
+		std::optional<Point> itemTile = FindAdjacentPositionForItem(myPlayer.position.future, myPlayer._pdir);
+		if (itemTile) {
+			NetSendCmdPItem(true, CMD_PUTITEM, *itemTile, myPlayer.HoldItem);
+		} else {
+			if (!AutoPlaceItemInBelt(myPlayer, myPlayer.HoldItem, true)
+			    && !AutoPlaceItemInInventory(myPlayer, myPlayer.HoldItem, true)
+			    && !AutoPlaceItemInStash(myPlayer, myPlayer.HoldItem, true)) {
+				// This can fail for max gold, arena potions and a stash that has been arranged to not have room for the item
+				// all 3 cases are exteamly unlikely
```
StephenCWills, line 2103:
```suggestion
				// all 3 cases are extremely unlikely
```

StephenCWills (approved):

StephenCWills (approved):

PR #6181, opened 2023-06-03
Move seedinfo to text commands

StephenCWills:
Moves the `seedinfo` debug command to text commands so it can be used in release builds. The extra information about mid/end seeds is still only included in debug builds. Here's how it looks in a release build.

image

Thought I might slip this in as an option for 1.5.0 in case we want to be able to ask players of release builds to share seed info, but it does add some new strings for translation so may be better to defer. 🤷 

I considered attempting to translate some of the output strings, but devs are ultimately the ones who need to be able to read it so probably not a great idea.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -462,11 +462,49 @@ std::string TextCmdInspect(const string_view parameter)
 	return ret;
 }
 
+std::string TextCmdLevelSeed(const string_view parameter)
+{
+	string_view levelType = setlevel ? "set level" : "dungeon level";
+
+	char gameId[] = {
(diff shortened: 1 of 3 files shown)


StephenCWills:
I guess one downside to this is it no longer gets logged to stderr. I usually take screenshots after using the command, but I'm not sure if anyone prefers stderr for copy/paste.

AJenbo:
I haven't been good about notifying translators about the upcoming release so translations are a bit spotty in terms of new strings since 1.4.1

AJenbo (approved):

PR #6186, opened 2023-06-07
Fix Item::getName() for Doppelganger's prefix

### LINKED ISSUE #6185: Suffix "of the moon" does not give points to all attributes
Not sure if the item is valid.
mace

Savegame:
[single_0.hsv.zip](https://github.com/diasurgical/devilutionX/files/11676456/single_0.hsv.zip)

The width of the text is not good either.

StephenCWills:
Because of the fallthrough...

https://github.com/diasurgical/devilutionX/blob/555dac5d22bf4cd0e1d672c526f94b3e0db62c77/Source/items.cpp#L735-L742

Doppelganger's should discard the same amount of random numbers as `IPL_TOHIT_DAMP` when computing the translated name of the item.

This resolves #6185

image


Chance4us:
Many thanks for fixing.

PR #6187, opened 2023-06-07
Add --ini-file cmdline option

Helldiver-RX:
Makes it possible to change the name of the ini configuration file. The default name is still "diablo.ini".


StephenCWills:
Is there an advantage to using this over `--config-dir`?

Source/diablo.cpp
```diff
@@ -995,6 +996,12 @@ void DiabloParseFlags(int argc, char **argv)
 				diablo_quit(64);
 			}
 			paths::SetConfigPath(argv[++i]);
+    } else if (arg == "--ini-file") {
+			if (i + 1 == argc) {
+				PrintFlagsRequiresArgument("--ini-file");
+				diablo_quit(64);
+			}
+			paths::SetIniFile(argv[++i]);
```
qndel, line 1004: garbage formatting

StephenCWills:
> It is just an extra cmdline option and does something different: --config-dir sets the directory where "diablo.ini" is and --ini-file sets an optional new name for the file.

I'm not so much asking about what it does as opposed to what it's used for. In other words, I get that it changes the name of the file, but why do you want to change the name of the file?

DakkJaniels:
I forget, does `--config-dir` use the save games in that directory? If so, I guess that could be one possible use case (use different configuration, but still use same save games). Not sure how often that's really needed though.

Helldiver-RX:
> > It is just an extra cmdline option and does something different: --config-dir sets the directory where "diablo.ini" is and --ini-file sets an optional new name for the file.
> 
> I'm not so much asking about what it does as opposed to what it's used for. In other words, I get that it changes the name of the file, but why do you want to change the name of the file?

One possible use would be to keep several ini files in the same folder, each one with a different set of configuration values for the program.

Helldiver-RX:
> I forget, does `--config-dir` use the save games in that directory? If so, I guess that could be one possible use case (use different configuration, but still use same save games). Not sure how often that's really needed though.

Yes, that is what i had in mind with this.

StephenCWills:
> One possible use would be to keep several ini files in the same folder, each one with a different set of configuration values for the program.

Is this how you intend to use it? And is there an advantage to this over having multiple config directories in the same folder?

FitzRoyX:
> One possible use would be to keep several ini files in the same folder, each one with a different set of configuration values for the program.

You can already manually swap files out. Why would you need to flip between configs quickly like that? If this was added, you'd be the only person using it and mainline features should be something that at least a third of users will use.

Issue #4436, opened 2022-04-14
Unusable stairs


vlolteanu:
**Important information**
Linux / 1.4.0 / official build

**Describe**
The stairs up to level 7 are unusable in the attached save. The stairs on the other levels seem to be working just fine.

**Additional context**
Saved game here: 
[single_2.zip](https://github.com/diasurgical/devilutionX/files/8492071/single_2.zip)

PR #6191, opened 2023-06-09
When checking chamber of bone entry also check the current dungeon level

### LINKED ISSUE #110: Locked upstairs entry
For some reason i got locked the upstairs entry "up to level 6".
It could be happened when reading the "book of the blind".
A picture and the save game are attached.
entry
[single_4.sv.zip](https://github.com/diasurgical/devilutionX/files/3173865/single_4.sv.zip)

obligaron:
Fixes #110 and #4436

This is a vanilla issue.
The entry to chamber of bone reuses the dungeon pieces for going upwards.
That's why `InitL2Triggers` checks that this is not the entry to chamber of bone before placing a upwards-trigger on it.
And this check is faulty, because it's missing a dungeon level check.
The saves in #110 and #4436 were unlucky to have the same position for entry to chamber and the upwards pieces in dungeon level 7.

This PR adds the missing dungeon level check.

### DIFF
--- a/Source/levels/trigs.cpp
+++ b/Source/levels/trigs.cpp
@@ -170,7 +170,7 @@ void InitL2Triggers()
 	numtrigs = 0;
 	for (WorldTileCoord j = 0; j < MAXDUNY; j++) {
 		for (WorldTileCoord i = 0; i < MAXDUNX; i++) {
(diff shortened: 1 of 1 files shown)


Chance4us:
That's great, I never thought that someone would find the cause.
Thanks a lot for fixing.

AJenbo:
Oh no, now i have to start all over with writing the change log :D

PR #6200, opened 2023-06-13
CMakeLists.txt: Adjust version string

glebm:
For versions without a suffix, e.g. "1.5.0" rather than "1.5.0-dev", only include the build type if it is a debug build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,7 @@ if(NOT VERSION_SUFFIX)
   else()
     # For versions without a suffix, e.g. "1.5.0" rather than "1.5.0-dev",
     # only include the build type if it is a debug build.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -65,13 +67,17 @@ if(NOT VERSION_SUFFIX)
     if(GIT_COMMIT_HASH)
       set(VERSION_SUFFIX "${VERSION_SUFFIX}-${GIT_COMMIT_HASH}")
     endif()
+  else()
+    # For versions without a suffix, e.g. "1.5.0" rather than "1.5.0-dev",
+    # only include the build type if it is a debug build.
+    set(VERSION_SUFFIX "$<IF:$<CONFIG:Debug>,$<CONFIG>,>")
```
glebm, line 73:
```suggestion
    set(VERSION_SUFFIX "$<$<CONFIG:Debug>:$<CONFIG>>")
```

PR #6207, opened 2023-06-14
Update README.md

broocard:
The README is updated to clarify up-front that if the game is installed from GoG.com, a multi-step process may be needed to utilize the game with the port. This is useful for novice gamers (or programmers) to put them on notice that going the GoG.com route could lead them down a rabbit hole they do not want to go. Additional clarification regarding downloading the latest release was also included, simply to make the instructions more user friendly. Here is a copy of what the README would look like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ For a full list of changes see our [changelog](docs/CHANGELOG.md).
 
 # How to Install
 
-Note: You'll need access to the data from the original game. If you don't have an original CD then you can [buy Diablo from GoG.com](https://www.gog.com/game/diablo). Alternately you can use `spawn.mpq` from the [shareware](https://github.com/diasurgical/devilutionx-assets/releases/download/v2/spawn.mpq) [[2]](http://ftp.blizzard.com/pub/demos/diablosw.exe) version, in place of `DIABDAT.MPQ`, to play the shareware portion of the game.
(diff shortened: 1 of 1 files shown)


Trihedraf:
You can also just install the game and take the MPQs from the install instead of extracting them.

StephenCWills:
You should just be able to install the game and then DevX will automatically detect the location of the GOG install to get the MPQs.

AJenbo:
I guess @broocard was not installing on Windows, although I would be surprised if someone looking for a port downloaded one not with the targeted system in it's name.

One problem with adding a lot of notices is that the more text there is the less likely people are to read any of it.
Which is why we have moved such details in to the [Installation Instructions](https://github.com/diasurgical/devilutionX/blob/master/docs/installing.md) which are linked below the offending text.

The majority of people will have to do nothing (Windows with an existing GOG installation), but if you intend to install it on an iPad and you only have a Mac then things can get a lot more complicated of cause. Things being very different depending on your use case is also another good reason to keep the generic instructions to just the common parts and leave the specifics for the "more detailed instructions" section for those that prefer being informed before diving in to something as to not confuse the people that the notes are not relevant for.

DakkJaniels:
It might be worthwhile to say something about not needing to move MPQs if you have the gog installation, but yeah, I agree that having less text, at least up front, is better than more.

AJenbo:
Thanks for the suggestion but I think we will close this PR

PR #6209, opened 2023-06-14
Bump VERSION to 1.6.0-dev

glebm:


AJenbo:
This should probably be more like a revert of https://github.com/diasurgical/devilutionX/commit/0438be6cf3c92706d8c94a9a012b69febcf1b5e3

glebm:
Ah yeah the libzt address should change. I'll let you do that.

PR #6232, opened 2023-06-17
Ensure library theme bookcases are solid to match normal bookshelves

### LINKED ISSUE #6691: [Issue Report]: Can't cast some spells when standing at a bookshelf
### Operating System

Other (please specify)

### DevilutionX version

1.5.1

### Describe

In title

**Operating System**: Manjaro ARM (aarch64)

### To Reproduce

1. Find a bookshelf (church/catacombs)
2. Step on its tile
3. Cast spells like Charged Bolt, Elemental, Nova, Bone Spirit and Blood Star
4. The spell doesn't affect the surroundings of the tile

### Expected Behavior

Spells shouldn't hit an invisible wall

### Additional context

**Screenshot**

Screenshot from 2023-10-07 18-28-05

ephphatha:
Fixes #6691

I'm not sure if there's ever a situation where a library theme room is unnavigable if the bookcase is solid, seems more likely that this was an oversight and both bookcases should have been solid to prevent issues like missiles colliding on spawn.


ephphatha:
Actually turns out the library versions were always non-solid, I was just looking at the wrong column in the table.

PR #6241, opened 2023-06-18
Fix net validation of durability for items without durability

StephenCWills:
Potions, scrolls, books, etc do not have durability and would therefore cause an error because the max value is smaller than the min value in the call to `clamp()`. I introduced a new function, following the pattern for validation of CTH and max damage.


AJenbo (approved):
Nice :+1:

PR #6255, opened 2023-06-22
Do not crash on missing fonts

### LINKED ISSUE #6247: Regression: Unknown symbols will cause a crash
### Operating System

Other (please specify)

### DevilutionX version

1.5.0 (latest release)

### Describe

Entering a symbol that the font file is not available for will crash the game

### To Reproduce

1:  Copy a chinese symbol
2: start a single player hero
3: when entering the name paste the symbol

### Expected Behavior

_No response_

### Additional context

_No response_

AJenbo:
Fixes #6247

Fall back to the original PCX fonts was broken as it looked for `file.pcx.pcx` instead of `file.pcx`. This may have resulted in a crash instead of showing the error message about upgrading if the user also had translations enabled.

If a symbol is not available it will now instead fall back to `?`. Only in case that is also unavailable do we terminate with a `Missing fonts` (missing devilutionx.mpq or application assets).

`LoadFont()` now returns a reference instead of a pointer so that we can be sure it's not null once we pass that point and can safely dereference it, as well as use `LoadFont()` direction.

### DIFF
--- a/Source/engine/render/text_render.cpp
+++ b/Source/engine/render/text_render.cpp
@@ -37,7 +37,6 @@ namespace {
 
 constexpr char32_t ZWSP = U'\u200B'; // Zero-width space
 
-using Font = const OwnedClxSpriteList;
 std::unordered_map<uint32_t, OptionalOwnedClxSpriteList> Fonts;
 
 std::unordered_map<uint32_t, std::array<uint8_t, 256>> FontKerns;
@@ -144,7 +143,7 @@ bool IsSmallFontTallRow(uint16_t row)
 
 void GetFontPath(GameFontTables size, uint16_t row, string_view ext, char *out)
 {
-	*fmt::format_to(out, R"(fonts\{}-{:02x}.{})", FontSizes[size], row, ext) = '\0';
+	*fmt::format_to(out, R"(fonts\{}-{:02x}{})", FontSizes[size], row, ext) = '\0';
 }
 
 uint32_t GetFontId(GameFontTables size, uint16_t row)
@@ -162,7 +161,7 @@ std::array<uint8_t, 256> *LoadFontKerning(GameFontTables size, uint16_t row)
 	}
 
 	char path[32];
-	GetFontPath(size, row, "bin", &path[0]);
+	GetFontPath(size, row, ".bin", &path[0]);
 
 	auto *kerning = &FontKerns[fontId];
 
@@ -181,7 +180,7 @@ std::array<uint8_t, 256> *LoadFontKerning(GameFontTables size, uint16_t row)
 	return kerning;
 }
 
-const OwnedClxSpriteList *LoadFont(GameFontTables size, text_color color, uint16_t row)
+OptionalClxSpriteList LoadFont(GameFontTables size, text_color color, uint16_t row)
 {
 	if (ColorTranslations[color] != nullptr && !ColorTranslationsData[color]) {
 		ColorTranslationsData[color].emplace();
@@ -191,11 +190,11 @@ const OwnedClxSpriteList *LoadFont(GameFontTables size, text_color color, uint16
 	const uint32_t fontId = GetFontId(size, row);
 	auto hotFont = Fonts.find(fontId);
 	if (hotFont != Fonts.end()) {
-		return &*hotFont->second;
+		return OptionalClxSpriteList(*hotFont->second);
 	}
 
 	char path[32];
-	GetFontPath(size, row, "clx", &path[0]);
+	GetFontPath(size, row, ".clx", &path[0]);
 
 	OptionalOwnedClxSpriteList &font = Fonts[fontId];
 	font = LoadOptionalClx(path);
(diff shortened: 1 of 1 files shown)


Source/engine/render/text_render.cpp
```diff
@@ -147,6 +147,11 @@ void GetFontPath(GameFontTables size, uint16_t row, string_view ext, char *out)
 	*fmt::format_to(out, R"(fonts\{}-{:02x}.{})", FontSizes[size], row, ext) = '\0';
 }
 
+void GetFontBaseNamePath(GameFontTables size, uint16_t row, char *out)
+{
+	*fmt::format_to(out, R"(fonts\{}-{:02x})", FontSizes[size], row) = '\0';
```
glebm, line 152: I'm guessing `GetFontPath` is no longer used with this? Let's delete it?

Source/engine/render/text_render.cpp
```diff
@@ -411,15 +418,21 @@ uint32_t DoDrawString(const Surface &out, string_view text, Rectangle rect, Poin
 		if (next == ZWSP)
 			continue;
 
-		const uint32_t unicodeRow = GetUnicodeRow(next);
+		uint32_t unicodeRow = GetUnicodeRow(next);
 		if (unicodeRow != currentUnicodeRow || font == nullptr) {
 			kerning = LoadFontKerning(size, unicodeRow);
-			font = LoadFont(size, color, unicodeRow);
+			font = &LoadFont(size, color, unicodeRow);
+			if (!*font) {
```
glebm, line 425: This doesn't quite make sense. `*font` is a `const OwnedClxSpriteList&`. Applying an implicit boolean conversion to it is not meaningful.

Source/engine/render/text_render.cpp
```diff
@@ -754,10 +767,16 @@ void DrawStringWithColors(const Surface &out, string_view fmt, DrawStringFormatA
 			continue;
 		}
 
-		const uint32_t unicodeRow = GetUnicodeRow(next);
+		uint32_t unicodeRow = GetUnicodeRow(next);
 		if (unicodeRow != currentUnicodeRow || font == nullptr) {
 			kerning = LoadFontKerning(size, unicodeRow);
-			font = LoadFont(size, color, unicodeRow);
+			font = &LoadFont(size, color, unicodeRow);
+			if (!*font) {
```
glebm, line 774: Likewise, `!*font` doesn't make sense.

Source/engine/render/text_render.cpp
AJenbo: `GetFontBaseNamePath()` is used for PCX and `GetFontPath()` is used for clx because the  `loaderPcx()` appends it's own extension, unlike `loadpcx()`. This is a biproduct of the unpacked mpq support.

Source/engine/render/text_render.cpp
glebm: Oh nvm `font` is a pointer here.

AJenbo: It's been changed to `OptionalOwnedClxSpriteList` since the fail case was returning a `OwnedClxSpriteList` with null data and not a `nullptr`.

Source/engine/render/text_render.cpp
AJenbo: https://github.com/diasurgical/devilutionX/pull/6255#discussion_r1240499596

Source/engine/render/text_render.cpp
glebm: nit: Let's use `OptionalClxSpriteList` instead of a pointer?

AJenbo: Wasn't sure how to do so with out causing a copy. Or do you think doing so would be ok?

Source/engine/render/text_render.cpp
glebm: I see. Let's remove the `.` from the format string in `GetFontPath` and pass the extension with the dot instead (e.g. ".bin").
Then we won't need `GetFontBaseNamePath`, as we can just pass `/*ext=*/""`.

Source/engine/render/text_render.cpp
glebm: `OptionalClxSpriteList` is free to copy, it's represented as a single pointer.
`OptionalClxSpriteList` ~ `OptionalOwnedClxSpriteList` relationship is similar to `std::optional<std::string_view>` ~ `std::optional<std::string>`.

Source/engine/render/text_render.cpp
```diff
@@ -686,7 +712,7 @@ uint32_t DrawString(const Surface &out, string_view text, const Rectangle &rect,
 		ClxDraw(clippedOut, characterPosition + Displacement { 0, lineHeight - BaseLineOffset[size] }, sprite);
 	} else if (HasAnyOf(flags, UiFlags::TextCursor) && GetAnimationFrame(2, 500) != 0) {
 		MaybeWrap(characterPosition, 2, rightMargin, initialX, lineHeight);
-		DrawFont(clippedOut, characterPosition, LoadFont(size, color, 0), color, '|', outlined);
+		DrawFont(clippedOut, characterPosition, *LoadFont(size, color, 0), color, '|', outlined);
```
glebm, line 715: Does this call to `LoadFont` always return a non-nullopt value?

Source/engine/render/text_render.cpp
```diff
@@ -750,26 +774,26 @@ void DrawStringWithColors(const Surface &out, string_view fmt, DrawStringFormatA
...
 		}
 
-		const uint32_t unicodeRow = GetUnicodeRow(next);
-		if (unicodeRow != currentUnicodeRow || font == nullptr) {
-			kerning = LoadFontKerning(size, unicodeRow);
-			font = LoadFont(size, color, unicodeRow);
-			currentUnicodeRow = unicodeRow;
+		if (!font.load(size, color, next)) {
+			next = U'?';
+			if (font.load(size, color, next)) {
```
glebm, line 783: `!font.load`?

Source/engine/render/text_render.cpp
```diff
@@ -205,21 +204,50 @@ const OwnedClxSpriteList *LoadFont(GameFontTables size, text_color color, uint16
...
 		LogError("Error loading font: {}", path);
-		return nullptr;
 	}
 
-	return &(*font);
+	return OptionalClxSpriteList(*font);
 }
 
-void DrawFont(const Surface &out, Point position, const OwnedClxSpriteList *font, text_color color, int frame, bool outline)
+class Font {
```
glebm, line 218: Perhaps `CurrentFont` to make it more obvious that it maintains some state?

Source/engine/render/text_render.cpp
AJenbo: lol, good catch.

Source/engine/render/text_render.cpp
AJenbo: It can technically fail if the base font is missing from the assets (devilutionx.mpq). It's an edge case we have had for a while as we did not check inside `DrawFont()` either. But I could wrap this in a check to avoid the fail case.

AJenbo: Done

glebm (approved):

PR #6256, opened 2023-06-23
Fix "Keep Strike Team as Name" in de.po

WebSnke:


ChaosMarc:
Why is it that the translation should be kept English? In German there is the the exakt translation "Einsatzkommando" for strike team (as an improvement to "Einsatzteam")

WebSnke:
> Why is it that the translation should be kept English? In German there is the the exakt translation "Einsatzkommando" for strike team (as an improvement to "Einsatzteam")

Because it specifically says "Keep Strike Team as Name".

ChaosMarc:
> Because it specifically says "Keep Strike Team as Name".

Yes. Sorry I should have specified that I'm not questioning your PR (which does what the instruction says) but the instruction itself to not translate it.

qndel:
Not providing a justification for such a thing is the best way to get your PR closed

ChaosMarc:
@qndel could you please take another look. This PR does exactly what the instruction to translators in the po file says: Not to translate "Diablo Strike Team". Either someone ignored/didn't see this instruction in the past or it was added later on. Nevertheless it has been translated and this PR reverts this fact to fulfill the requirement of not translating it. I was just asking if someone of the dev team knows WHY it should not be translated.

qndel:
Okay, wasn't aware of it, was a bit hard to get it from this conversation ;)

AJenbo:
> @qndel could you please take another look. This PR does exactly what the instruction to translators in the po file says: Not to translate "Diablo Strike Team". Either someone ignored/didn't see this instruction in the past or it was added later on. Nevertheless it has been translated and this PR reverts this fact to fulfill the requirement of not translating it. I was just asking if someone of the dev team knows WHY it should not be translated.

I think it was added as it's a general translation rule for the languages that they spoke. If this does not apply for German you should overrule it.

ChaosMarc:
I think this is very subjective. I play the Diablo in English like every other game, so I don't have a strong opinion about this. If all other languages don't translate it and if there are no plans to change this I wouldn't translate it to German neither. If we were to remove the instruction to not translate it for all languages and update them respectively, everything should stay the same for the current German translation as my gut feeling that "team" should be changed to "kommando" was wrong as this would translate to "task force". Military slang is obviously not my strongest area of expertise 🙈

AJenbo (approved):

PR #6262, opened 2023-06-24
Player Death Messages and Player Hostile/Friendly alert

kphoenix137:
Adds chat messages for when a player dies, showing the thing that killed the player. Many messages are randomized for fun and variety. Also shows a chat message when a player toggles Friendly or Hostile. Will also show Friendly or Hostile for players joining into a game for each player already in the game. Also added player class to the message that shows player names and level when joining games.

Downside: Needs translation for every language.

### DIFF
--- a/Source/msg.h
+++ b/Source/msg.h
@@ -203,9 +203,10 @@ enum _cmd_id : uint8_t {
 	CMD_MONSTDAMAGE,
 	// Player death.
 	//
-	// body (TCmdParam2):
+	// body (TCmdParam3):
 	//    int16_t ear_flag
-	//    int16_t death_source_index
+	//    int16_t death_source
+	//    int16_t is_monster_unique
 	CMD_PLRDEAD,
 	// Lift item to hand request.
 	//
@@ -505,6 +506,20 @@ struct TCmdParam5 {
 	uint16_t wParam5;
 };
 
+struct TCmdPDeath {
+	_cmd_id bCmd;
+	uint16_t wDeathReason;
+	uint16_t wDeathSourceIndex;
+	uint16_t wMonsterUid;
+	bool bWasKilledByUnique;
+};
+
+struct TCmdMDeath {
+	_cmd_id bCmd;
+	uint16_t wMonsterUid;
+	bool bIsMonsterUnique;
+};
+
 struct TCmdGolem {
 	_cmd_id bCmd;
 	uint8_t _mx;
@@ -752,6 +767,8 @@ void NetSendCmdLocParam5(bool bHiPri, _cmd_id bCmd, Point position, uint16_t wPa
 void NetSendCmdParam1(bool bHiPri, _cmd_id bCmd, uint16_t wParam1);
 void NetSendCmdParam2(bool bHiPri, _cmd_id bCmd, uint16_t wParam1, uint16_t wParam2);
 void NetSendCmdParam5(bool bHiPri, _cmd_id bCmd, uint16_t wParam1, uint16_t wParam2, uint16_t wParam3, uint16_t wParam4, uint16_t wParam5);
+void NetSendCmdPDeath(bool bHiPri, _cmd_id bCmd, uint16_t wDeathReason, uint16_t wDeathSourceIndex, uint16_t wMonsterUid, bool bWasKilledByUnique = false);
+void NetSendCmdMDeath(bool bHiPri, _cmd_id bCmd, uint16_t wMonsterUid, bool bIsMonsterUnique = false);
 void NetSendCmdQuest(bool bHiPri, const Quest &quest);
 void NetSendCmdGItem(bool bHiPri, _cmd_id bCmd, uint8_t pnum, uint8_t ii);
 void NetSendCmdPItem(bool bHiPri, _cmd_id bCmd, Point position, const Item &item);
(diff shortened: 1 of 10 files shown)


AJenbo:
I don't get the need for the unique monster

kphoenix137:
> I don't get the need for the unique monster 

For funsies

sskras:
Unique monsters are like a milestones. If you kill one, it's as like you complete some chapter. I find this to be very informative to have.

FitzRoyX:
Aren't you supposed to just see some of these things happen by fighting with the people you're playing with?

kphoenix137:
> Unique monsters are like a milestones. If you kill one, it's like you have completed some chapter. I find logging this event to be very informative.

I agree. Also I like that it will announce when boss enemies die, lets everyone else in the game see what's going on in case they are in another location. I would have preferred to show exactly which monster or player killed someone, however that data isn't available in the functions I wrote these in. It might be possible to pass that information but I'm not sure, since it seems like damage packets and death packets are mutually exclusive.

kphoenix137:
Big changes. 

* Revises DeathSource enum class to differentiate Monster and Trap. 
* Sends an extra variable over the network with the player death cmd containing the index of the player or monster that caused the death. The index variable is passed down starting at ApplyPlrDamage()
* Changed some damage sources from MonsterOrTrap to Unknown, such as for Rage skill, Blood Star/Bone Spirit, Life Drain so we get accurate chat messages and avoid a bug where it attempts to write a null name to chat.

kphoenix137:
* Removed unique monster death message
* Fixed bug with wrong monster name showing if players are in different levels
* Death messages are now red

kphoenix137:
@AJenbo ready for review

kphoenix137:
* Added random death messages

ikonomov:
I think parts of this PR are easier to evaluate than others and it might be better to separate them.  I see a benefit of having 1a and 1b and potential for annoyance after using 2a and 2b for a long period of time.

1. a) Chat messages for Friendly/Hostile toggle.
    b) More detailed chat messages when joining that includes Friendly/Hostile and also class and level.
2. a) Chat messages for being killed in PvP.
    b) Chat messages for being killed by a monster.

ikonomov:
Maybe instead of messages we can get an indication of dead players on the map by changing the arrow into an X.

kphoenix137:
> Maybe instead of messages we can get an indication of dead players on the map by changing the arrow into an X.

Thank you for taking the time to review my pull request and sharing your feedback. I appreciate your perspective and the points you've raised. I would like to address each of your suggestions individually:

1a) Chat messages for Friendly/Hostile toggle: I agree that adding chat messages to indicate when a player toggles between Friendly and Hostile can be a useful addition. It provides clarity for other players and enhances the overall communication within the game.

1b) More detailed chat messages when joining: I understand your concerns about potential annoyance after prolonged use. I believe that including the player's class and level in the chat messages when joining a game can provide valuable information and enhance the player's experience. However, I'm open to discussing alternative ways to present this information that might mitigate any potential annoyance.

2a) Chat messages for being killed in PvP: I believe including chat messages to inform players when they are killed in player-versus-player (PvP) encounters can be beneficial for maintaining engagement and providing feedback.

2b) Chat messages for being killed by a monster: Similarly, notifying players through chat messages when they are killed by a monster can enhance their understanding of the game dynamics and add an element of immersion.

Regarding your suggestion about indicating dead players on the minimap by changing the arrow into an X, I appreciate your alternative approach. However, I believe that chat messages serve as a more direct and immediate means of communication. The proposed X on the minimap might not convey the same level of information and could be less noticeable to players.

In conclusion, I believe that chat messages offer an effective and informative way to communicate important events within the game. However, I'm open to refining the implementation to address any potential concerns and ensure a positive user experience.

Once again, I appreciate your input and look forward to further discussing and refining these features for the benefit of the DevilutionX community.

Source/msg.h
```diff
@@ -751,6 +762,8 @@ void NetSendCmdLocParam5(bool bHiPri, _cmd_id bCmd, Point position, uint16_t wPa
 void NetSendCmdParam1(bool bHiPri, _cmd_id bCmd, uint16_t wParam1);
 void NetSendCmdParam2(bool bHiPri, _cmd_id bCmd, uint16_t wParam1, uint16_t wParam2);
 void NetSendCmdParam5(bool bHiPri, _cmd_id bCmd, uint16_t wParam1, uint16_t wParam2, uint16_t wParam3, uint16_t wParam4, uint16_t wParam5);
+void NetSendCmdPDeath(bool bHiPri, _cmd_id bCmd, uint16_t wDeathReason, uint16_t wDeathSourceIndex, uint16_t wMonsterUid, bool bWasKilledByUnique = false);
+void NetSendCmdMDeath(bool bHiPri, _cmd_id bCmd, uint16_t wMonsterUid, bool bIsMonsterUnique = false);
```
kphoenix137, line 766: pending deletion

ikonomov: https://tenor.com/view/cat-typing-typing-on-computer-computer-work-laptop-gif-21481919 eventually something will get through...

ikonomov:
https://media.tenor.com/bCfpwMjfAi0AAAAC/cat-typing.gif eventually something will get through...

PR #6265, opened 2023-06-25
MSVC: Create config for building release with debug info in visual studio

obligaron:
I used `RelWithDebInfo` when doing some performance profiling and this ensures the pdb's are generated.

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -76,6 +76,24 @@
       ]
     },
     {
+      "name": "x64-RelWithDebInfo",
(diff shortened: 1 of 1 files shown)


CMakeSettings.json
```diff
@@ -75,6 +75,24 @@
         }
       ]
     },
+    {
+      "name": "x64-Release-WithDebInfo",
```
glebm, line 79:
```suggestion
      "name": "x64-RelWithDebInfo",
```

glebm (approved):

PR #6266, opened 2023-06-25
Fix PoolFix() to always replace dirt ceiling tiles if they are adjacent to a ground lava tile

### LINKED ISSUE #3130: [vanilla] Defective lava with mirroring effect
The issue looks similar to #477 but is not the same.
The savegame also looks the same in vanilla. 
[mirroring.zip](https://github.com/diasurgical/devilutionX/files/7346719/mirroring.zip)
lava

Or could there be a regression? Somehow it looks exactly the same.

obligaron:
Fixes #3130

In masster `PoolFix()` checks if a ceiling/dirt tile is surrounded by lava ground tiles (id 25 to 41).
This doesn't work when two ceiling/dirt tiles are needed to be filled with lava.

This PR simplify the algorithm on the following basis:
A ground lava tile can never be directly connected to a ceiling tile.
There must always be a kind of transition tile between (from ground to ceiling).

That means a ceiling/dirt tile is part of a lava pool (and was missed to fill in `River()`), when it's adjacent to a ground lava tile.
In this case we change the tile to a ground lava tile.

### DIFF
--- a/Source/levels/drlg_l3.cpp
+++ b/Source/levels/drlg_l3.cpp
@@ -1501,23 +1501,25 @@ bool PlacePool()
 	return PlaceLavaPool();
 }
 
+/**
+ * @brief Fill lava pools correctly, cause River() only generates the edges.
+ */
 void PoolFix()
 {
-	for (int duny = 1; duny < DMAXY - 1; duny++) {     // BUGFIX: Change '0' to '1' and 'DMAXY' to 'DMAXY - 1' (fixed)
-		for (int dunx = 1; dunx < DMAXX - 1; dunx++) { // BUGFIX: Change '0' to '1' and 'DMAXX' to 'DMAXX - 1' (fixed)
-			if (dungeon[dunx][duny] == 8) {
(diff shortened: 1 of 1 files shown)


Source/levels/drlg_l3.cpp
```diff
@@ -1501,23 +1501,25 @@ bool PlacePool()
...
-				}
+	for (Point tile : PointsInRectangle(Rectangle { { 1, 1 }, { DMAXX - 2, DMAXY - 2 } })) {
+		// Check if the tile is a the default dirt ceiling tile
+		if (dungeon[tile.x][tile.y] != 8)
+			continue;
+
+		for (Point adjacentTiles : PointsInRectangle(Rectangle { tile - Displacement(1, 1), { 3, 3 } })) {
+			int tileId = dungeon[adjacentTiles.x][adjacentTiles.y];
+			// Check if the adjacent tile is a ground lava tile
+			if (25 <= tileId && tileId <= 41) {
```
AJenbo, line 1517: a bit confusing
```suggestion
			if (tileId >= 25 && tileId <= 41) {
```

AJenbo (approved):

PR #6267, opened 2023-06-25
Remove Trap-Flag from OBJ_SIGNCHEST

### LINKED ISSUE #3810: Chest [ogden's tavern sign] can look like it is equipped with a trap
The chest with the ogden's tavern sign can look like it is equipped with a trap.
The chest also had a red outline as long as it was not open.
When you open the chest there is no trap.
It is very likely a vanilla bug.

In the wall you can see an opening for a possible projectile.
trap

Savegame:
[tavern_sign_chest.zip](https://github.com/diasurgical/devilutionX/files/7747983/tavern_sign_chest.zip)

obligaron:
Fixes #3810

Without the Trap-Flag the game doesn't try to place a trap on the chest.
Alternative fix would be handling the placed trap correctly in `OperateInnSignChest`.


AJenbo (approved):

AJenbo:
i like this fix

PR #6276, opened 2023-06-28
Floating Info Box

kphoenix137:
https://github.com/diasurgical/devilutionX/assets/68359262/6cbf710b-653a-4131-9abc-a95d7ee2de6e

Adds toggle option for a floating info box (only for inventory currently). Features both moving the control panel item info to a floating box near the item, as well as expanded information since it's no longer limited to the space confinements of the control panel.

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1940,10 +1940,15 @@ int8_t CheckInvHLight()
 	} else {
 		InfoColor = pi->getTextColor();
 		InfoString = pi->getName();
+
 		if (pi->_iIdentified) {
-			PrintItemDetails(*pi);
+			if (!*sgOptions.Gameplay.enableFloatingInfoBox) {
+				PrintItemDetails(*pi);
+			}
 		} else {
-			PrintItemDur(*pi);
+			if (!*sgOptions.Gameplay.enableFloatingInfoBox) {
+				PrintItemDur(*pi);
+			}
 		}
 	}
 
--- a/Source/options.cpp
+++ b/Source/options.cpp
@@ -1083,6 +1083,7 @@ GameplayOptions::GameplayOptions()
               { FloatingNumbers::Random, N_("Random Angles") },
               { FloatingNumbers::Vertical, N_("Vertical Only") },
           })
+    , enableFloatingInfoBox("Enable Floating Info Box", OptionEntryFlags::None, N_("Enable Floating Info Box"), N_("Moves information from the control panel to a floating box."), false)
 {
 	grabInput.SetValueChangedCallback(OptionGrabInputChanged);
 	experienceBar.SetValueChangedCallback(OptionExperienceBarChanged);
@@ -1128,6 +1129,7 @@ std::vector<OptionEntryBase *> GameplayOptions::GetEntries()
 		&disableCripplingShrines,
 		&adriaRefillsMana,
 		&grabInput,
+		&enableFloatingInfoBox,
 	};
 }
 
--- a/Source/qol/floatinginfobox.cpp
+++ b/Source/qol/floatinginfobox.cpp
@@ -1,0 +1,731 @@
+/**
+ * @file floatinginfobox.cpp
+ *
+ * Adds floating info box QoL feature
+ */
+
+#include "qol/floatinginfobox.hpp"
+
+#include "itemlabels.h"
+
+#include <algorithm>
+#include <limits>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include <fmt/format.h>
+
+#include "control.h"
+#include "controls/plrctrls.h"
+#include "cursor.h"
+#include "engine/point.hpp"
+#include "engine/render/clx_render.hpp"
+#include "gmenu.h"
+#include "inv.h"
+#include "options.h"
+#include "qol/stash.h"
+#include "utils/format_int.hpp"
+#include "utils/language.h"
+#include "utils/stdcompat/string_view.hpp"
+
+namespace devilution {
+
+namespace {
+
+bool extraInfoKeyPressed = false;
+
+} // namespace
+
+void ExtraInfoKeyPressed(bool pressed)
+{
+	extraInfoKeyPressed = pressed;
+}
+
+enum class ItemStatType {
+	Damage,
+	Armor,
+	Durability
+};
+
+UiFlags GetItemBonusColors(const ItemStatType stat, const Item &item)
+{
+	ItemData iData = AllItemsList[item.IDidx];
+	UiFlags color = UiFlags::ColorWhite;
+
+	if (item._iMagical == ITEM_QUALITY_UNIQUE) {
+		const UniqueItem &uitem = UniqueItems[item._iUid];
+		assert(uitem.UINumPL <= sizeof(uitem.powers) / sizeof(*uitem.powers));
+		for (const auto &power : uitem.powers) {
+			switch (power.type) {
+			case IPL_ACP:
+				if (stat == ItemStatType::Armor)
+					color = UiFlags::ColorBlue;
+				break;
+			case IPL_ACP_CURSE:
+				if (stat == ItemStatType::Armor)
+					color = UiFlags::ColorRed;
+				break;
+			case IPL_SETAC:
+			case IPL_AC_CURSE:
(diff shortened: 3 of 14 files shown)


kphoenix137:
Then the question becomes what do we do with that disgusting black void at the bottom of the screen?

Trihedraf:
Get that new ui on there and get rid of the wasted space

kphoenix137:
> Get that new ui on there and get rid of the wasted space

that's a whole nother ball game and that will DEFINITELY some some massive pushback

FitzRoyX:
Nice, been waiting a long time for this. I don't think it should be an option, but whatever. Make town map the option, that thing is getting annoying since I know where everything is.

Source/qol/floatinginfobox.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+} // namespace
+
+void PrintFloatingItemDetails(const Item &item)
+{
+	if (HeadlessMode)
+		return;
+
+	//Player &player = *MyPlayer;
+
```
AJenbo, line 332:
```suggestion
```

Source/qol/floatinginfobox.cpp
```diff
@@ -0,0 +1,480 @@
...
+				}
+				if (i != lines.size() - 1) {
+					linesBase.append("\n");
+				}
+			}
+
+			previousLineHadColon = currentLineHasColon;
+		}
+
+		std::string_view linesBaseView(linesBase.data(), linesBase.length());
```
AJenbo, line 287: Use the polyfill until 1.6.0-dev
```suggestion
		string_view linesBaseView(linesBase.data(), linesBase.length());
```

Chance4us:
> Then the question becomes what do we do with that disgusting black void at the bottom of the screen?

I thought that this black area is also needed for other descriptions, for example for the Monter type or when there are items on the ground. But if this is no longer the case, then this area could be made transparent. Or we could use AI to paint over the surface with a suitable motif. But I would rather vote that we would create a larger battlefield.

AJenbo:
The monster HP bar can show most of the other relevant informations. In combination they could provide for a more minimal main panel:

image

Chance4us:
Nice! It's really very compact.

kphoenix137:
> The monster HP bar can show most of the other relevant informations. In combination they could provide for a more minimal main panel:
> 
> image
> 
> 

Needs spell select

kphoenix137:
Bug: New player default gear doesn't show durability

kphoenix137:
Bug: Bottom half inventory item boxes display an inventory slot height too high

FitzRoyX:
First post indicates it is working for stash items, but I can't get it to. Known?

kphoenix137:
> First post indicates it is working for stash items, but I can't get it to. Known?

Not implemented yet

FishGaming:
This looks really nice!

kphoenix137:
Currently broken. Can't figure out how to format these strings correctly (for damage range and durability range), and I can't figure out how to grab the item power string from PrintItemPower(). It bugs out hard when trying to display the strings.

Source/engine/render/text_render.cpp
```diff
@@ -574,6 +574,25 @@ int GetLineHeight(string_view text, GameFontTables fontIndex)
 	return LineHeights[fontIndex];
 }
 
+int GetLongestLineWidth(string_view text, GameFontTables size, int spacing)
```
kphoenix137, line 577: Don't think this is used anymore. Will likely delete

Source/qol/stash.cpp
```diff
@@ -446,11 +447,17 @@ uint16_t CheckStashHLight(Point mousePosition)
 	}
 
 	InfoColor = item.getTextColor();
-	InfoString = item.getName();
```
kphoenix137, line 449: Need to check why this line was removed. Possibly bad rebase

Source/options.cpp
```diff
@@ -1128,6 +1129,8 @@ std::vector<OptionEntryBase *> GameplayOptions::GetEntries()
 		&disableCripplingShrines,
 		&adriaRefillsMana,
 		&grabInput,
+		&enableFloatingNumbers,
```
kphoenix137, line 1132: WTF? Why is this here

Source/inv.cpp
```diff
@@ -1188,6 +1230,15 @@ void DrawInvBelt(const Surface &out)
 
 		DrawItem(myPlayer.SpdList[i], out, position, sprite);
 
+		if (*sgOptions.Gameplay.enableFloatingInfoBox) {
+			// Draw info box
+			if (pcursinvitem == i + INVITEM_BELT_FIRST) {
+				if (ControlMode == ControlTypes::KeyboardAndMouse || invflag) {
```
kphoenix137, line 1236: I think this needs to be fixed or else non K&M users won't be able to have floating info box?

Source/inv.cpp
```diff
@@ -1155,6 +1156,47 @@ void DrawInv(const Surface &out)
 			DrawItem(myPlayer.InvList[ii], out, position, sprite);
 		}
 	}
+
+	if (*sgOptions.Gameplay.enableFloatingInfoBox) {
```
kphoenix137, line 1160: Add stash functionality here

Source/qol/floatinginfobox.cpp
```diff
@@ -0,0 +1,485 @@
...
+	// Add the lines vector contents into the linesWithColor vector, which contains both the strings and the colors that they will be
+	std::vector<DrawStringFormatArg> linesWithColor;
+
+	int16_t modifiedVals[3] = {
+		item._iMinDam,
+		item._iMaxDam,
+		item._iAC,
+	};
+
+	if (item._iMagical == ITEM_QUALITY_UNIQUE) {
```
kphoenix137, line 190: Should extract to own function

Source/qol/floatinginfobox.cpp
```diff
@@ -0,0 +1,485 @@
...
+			break;
+		default:
+			break;
+		}
+	}
+
+	return itemBonusColors;
+}
+
+std::map<ItemRequirementType, UiFlags> GetItemRequirementColors(const Item &item)
```
kphoenix137, line 145: Completely stupid way to handle this. Should just refactor this function so I can call it and specify which exact thing I want item requirement color for (STR, DEX, or MAG) and have it return what I need instead of having it return all 3 and having to write them to a new map variable.

Source/qol/floatinginfobox.cpp
```diff
@@ -0,0 +1,485 @@
...
+	DUR
+};
+
+enum ItemRequirementType {
+	STR,
+	DEX,
+	MAG
+};
+
+std::map<ItemStatType, UiFlags> GetItemBonusColors(const Item &item)
```
kphoenix137, line 46: Need a refactor too (return the exact stat bonus type I want a color for)

Source/qol/floatinginfobox.cpp
```diff
@@ -0,0 +1,485 @@
...
+#include "inv.h"
+#include "options.h"
+#include "qol/stash.h"
+#include "utils/format_int.hpp"
+#include "utils/language.h"
+#include "utils/stdcompat/string_view.hpp"
+
+namespace devilution {
+
+enum ItemStatType {
```
kphoenix137, line 34: Can probably do something cleaner than this...

kphoenix137:
Okay so at the moment, everything is functional and floating info boxes appear for both stash and inventory items (belt too). Next step would probably be to add floating boxes for anything else UI related. Anything not related to the UI should probably expand on the code used to display ground items.

kphoenix137:
Ah, we have a bug with displaying correct magical weapon damage

julealgon:
@kphoenix137 while you are actively working on the change (finding bugs, sending commits, etc), please consider always marking the PRs as "draft". Then, once you are good to go, remove it.

Otherwise each new comment here triggers a notification for people listening on issues.

kphoenix137:
Altered the base AC of quest items that have unique data that manually sets the durability. Therefore there shouldn't be any compatibility issues, and the end-user won't notice. This change was made to improve the colored numbers displayed for these items for AC.

kphoenix137:
https://github.com/diasurgical/DevilutionX/pull/8000

PR #6280, opened 2023-06-28
Add fields to PlayerNetPack for validation

StephenCWills:
FYI, because `TItem` can't handle ears, they get filtered out in `PackNetItem()` and `UnPackNetItem()`.


AJenbo (approved):
Nice work I only have two comments but I think they are more in terms of what should be done in a follow up:
- We should probably add support for ears
- Many of the added fields do not need to be 32bit and could be lowered to 16 or 8, but this should also be done on the player struct at the same time to have things be clean.

PR #6288, opened 2023-06-29
Refactor CreatePlayer()

kphoenix137:
Revising this PR explanation:
* Adds new member variables to the `Player` struct to calculate base max life and base max mana
* Cleans up the code in `CreatePlayer()`

Future uses:
* Validating life and mana from network players

Modding uses:
* Calculating how much life has been lost as a result of Black Death zombies by getting the difference between the return value of the member function and player max base life
* Calculating how much mana has been lost as a result of Shrines by getting the difference between the return value of the member function and player max base mana


AJenbo (dismissed):

FitzRoyX:
>For example, a modder decides Warrior gets 3 life points per base vitality, and then uses shrines to lose all 25 base vitality down to 0. The Warrior would lose 75 base life, but only has 70

Can you help me understand how this is true? Jalruf seems to indicate that starting life is based on a formula and not some preset number. So if you changed the amount of life per vit, the Warrior would start with 95 life, not 70.
(vit25x3) + (clvl1x2) + 18 = 95

kphoenix137:
> >For example, a modder decides Warrior gets 3 life points per base vitality, and then uses shrines to lose all 25 base vitality down to 0. The Warrior would lose 75 base life, but only has 70
> 
> Can you help me understand how this is true? Jalruf seems to indicate that starting life is based on a formula and not some preset number. So if you changed the amount of life per vit, the Warrior would start with 95 life, not 70.
> (vit25x3) + (clvl1x2) + 18 = 95

```        player._pHitPoints = (player._pVitality + 10) << 6; 
         if (player._pClass == HeroClass::Warrior || player._pClass == HeroClass::Barbarian) { 
                 player._pHitPoints *= 2; 
         } else if (player._pClass == HeroClass::Rogue || player._pClass == HeroClass::Monk || player._pClass == HeroClass::Bard) { 
                 player._pHitPoints += player._pHitPoints / 2; 
         }```
Vanilla: 25 + 10 + ((25 +10) * 2) = 70
If you change Warrior from 2 points gained per vitality to 3, it's still the same result as 70. Then if you shrine your vit down to 0, you lose 25*3 life

**Comment by kphoenix137** [comment]
> >For example, a modder decides Warrior gets 3 life points per base vitality, and then uses shrines to lose all 25 base vitality down to 0. The Warrior would lose 75 base life, but only has 70
> 
> Can you help me understand how this is true? Jalruf seems to indicate that starting life is based on a formula and not some preset number. So if you changed the amount of life per vit, the Warrior would start with 95 life, not 70.
> (vit25x3) + (clvl1x2) + 18 = 95

The formula in Jarulfs guide technically isn't wrong but that isn't how base life and mana get set for creating new characters

**Comment by FitzRoyX** [comment]
I feel like changing to a switch statement using jalruf's formula would be preferred by modders. Vit being disconnected from the starting life value is too confusing. And this prevents having to break stuff out into new if statement the moment you change something. 

```
switch(player._pClass)
   case HeroClass::Warrior:
      player._pHitPoints = (player._pVitality * 2) + 20;
   case HeroClass::Rogue:
      player._pHitPoints = player._pVitality + 25;
   case HeroClass::Sorcerer:
      player._pHitPoints = player._pVitality + 10;
   case HeroClass::Monk:
      player._pHitPoints = player._pVitality + 25;
   case HeroClass::Bard:
      player._pHitPoints = player._pVitality + 25;
   case HeroClass::Barbarian:
      player._pHitPoints = (player._pVitality * 2) + 20;
```

**Comment by AJenbo** [comment]
We want to move to a data driven model, not alternative ways of hardcoding values

**Comment by kphoenix137** [comment]
> 

I don't understand how this would prevent the problem I outlined that this PR solves.

**Comment by kphoenix137** [comment]
> player._pHitPoints = (player._pVitality * 2) + 20;

Why player vitality * 2? Because Warrior gets 2 points of life per base vitality? By hardcoding the 2, you force the modder to have to change the 2 in that code block when changing the 2 located in playerdat.cpp

**Comment by FitzRoyX** [comment]
> > player._pHitPoints = (player._pVitality * 2) + 20;
> 
> Why player vitality * 2? Because Warrior gets 2 points of life per base vitality? By hardcoding the 2, you force the modder to have to change the 2 in that code block when changing the 2 located in playerdat.cpp
> 
> Also, just noticed that what you did is the same as what I did except I didn't hardcode as much

You're right. I think your PR looks good, I should've looked at it harder. Although I would rename "StartLife" to "LifeBonus" and put it at the end of the expression. Every time I hear "StartLife" I think of the cumulative number you start with.

**Comment by kphoenix137** [comment]
> > > player._pHitPoints = (player._pVitality * 2) + 20;
> > 
> > Why player vitality * 2? Because Warrior gets 2 points of life per base vitality? By hardcoding the 2, you force the modder to have to change the 2 in that code block when changing the 2 located in playerdat.cpp
> > 
> > Also, just noticed that what you did is the same as what I did except I didn't hardcode as much
> 
> You're right. I think your PR looks good, I should've looked at it harder. Although I would rename "startLife" to "lifeBonus" and put it at the end of the expression. Every time I hear "StartLife" I think of the cumulative number you start with.
> 
> I also think chrLife is a bad name. I'd rename that to "lifePerVit" or "vitMultiplier"

There are 2 vitality multipliers. Base vit and item vit. Also that would be for a different PR. This PR only adds startLife

**Review by yuripourre — APPROVED** [review]

**Comment by AJenbo** [comment]
@yuripourre thanks

File: Source/player.cpp [hunk 1]
```diff
@@ -2055,6 +2055,18 @@ void Player::UpdatePreviewCelSprite(_cmd_id cmdId, Point point, uint16_t wParam1
 	}
 }
 
+int32_t Player::calculateBaseLife() const
+{
+	const PlayerData playerData = PlayersData[static_cast<size_t>(_pClass)];
```
glebm, line 2060: This copies `playerData`, which is expensive.

Here and below:

```diff
- const PlayerData playerData = PlayersData[static_cast<size_t>(_pClass)];
+ const PlayerData &playerData = PlayersData[static_cast<size_t>(_pClass)];
```

PR #6291, opened 2023-06-30
InfoString Refactor

kphoenix137:
This PR is intended on changing how InfoString is written to for future usage with Floating Info Boxes. This adds functions for writing to InfoString so we no longer write directly to the variable, as well as a function to clear InfoString. NewPanelString() and AddPanelString() will utilize a vector that will keep track of the purpose of each line contained inside InfoString, so that lines and text within lines can be colored based on their purpose.

### DIFF
--- a/Source/control.cpp
+++ b/Source/control.cpp
@@ -755,6 +755,21 @@ void AddPanelString(std::string &&str)
 		InfoString = StrCat(InfoString, "\n", str);
 }
 
+void NewPanelString(string_view str)
+{
+	InfoString = str;
+}
+
+void NewPanelString(std::string &&str)
+{
+	InfoString = std::move(str);
+}
+
+void ClearPanelString()
+{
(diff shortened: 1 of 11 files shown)


kphoenix137:
Currently it contains a pretty bad bug with the control panel for displaying magical items for some reason. can't figure it out yet

AJenbo:
Why not add them with the right color to start with?

kphoenix137:
> Why not add them with the right color to start with?

There is a lot of code in the floating info box PR to decide what the right color is

kphoenix137:
I'm not sure if this PR is enough for what I need as is, since each line in InfoString for items contains specific information that all gets colored separately based on conditions. I think I was overthinking it completely. Maybe it's possible for me to just construct the contents of the floating item box in place directly reading item data instead of trying to deconstruct the InfoString that has item data written to it.

PR #6293, opened 2023-06-30
D2 Style Experience Sharing

kphoenix137:
Adds shared experience using the formula from Diablo 2. Each player in the dlvl receives a portion of the experience gain from the monster equal to clvl divided by the total clvl of each player in the dlvl. This has been a recurring requested feature from players, so I thought I'd make it. Haven't bothered with the demo stuff and likely won't unless there's some sort of indication this feature will be approved (Although I doubt it'll break the demo since experience gains are only adjusted in multiple person games)

### DIFF
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1815,9 +1815,10 @@ void InitKeymapActions()
 	    'V',
 	    [] {
 		    EventPlrMsg(fmt::format(
-		                    fmt::runtime(_(/* TRANSLATORS: {:s} means: Character Name, Game Version, Game Difficulty. */ "{:s} {:s}")),
+		                    fmt::runtime(_(/* TRANSLATORS: {:s} means: Character Name, Game Version, Game Difficulty. */ "{:s} {:s} {:d}")),
 		                    PROJECT_NAME,
-		                    PROJECT_VERSION),
+		                    PROJECT_VERSION,
+							sgGameInitInfo.bSharedExperience),
 		        UiFlags::ColorWhite);
 	    },
 	    nullptr,
--- a/Source/diablo.cpp
+++ b/Source/diablo.cpp
@@ -1815,10 +1815,9 @@ void InitKeymapActions()
 	    'V',
 	    [] {
 		    EventPlrMsg(fmt::format(
-		                    fmt::runtime(_(/* TRANSLATORS: {:s} means: Character Name, Game Version, Game Difficulty. */ "{:s} {:s} {:d}")),
+		                    fmt::runtime(_(/* TRANSLATORS: {:s} means: Character Name, Game Version, Game Difficulty. */ "{:s} {:s}")),
 		                    PROJECT_NAME,
-		                    PROJECT_VERSION,
-							sgGameInitInfo.bSharedExperience),
+		                    PROJECT_VERSION),
 		        UiFlags::ColorWhite);
 	    },
 	    nullptr,
--- a/Source/engine/demomode.cpp
+++ b/Source/engine/demomode.cpp
@@ -95,6 +95,7 @@ struct {
 	bool showItemLabels = false;
 	bool autoRefillBelt = false;
 	bool disableCripplingShrines = false;
+	bool enableSharedExperience = false;
 	uint8_t numHealPotionPickup = 0;
 	uint8_t numFullHealPotionPickup = 0;
 	uint8_t numManaPotionPickup = 0;
@@ -135,6 +136,7 @@ void ReadSettings(FILE *in, uint8_t version)
 		DemoSettings.showItemLabels = ReadByte(in) != 0;
 		DemoSettings.autoRefillBelt = ReadByte(in) != 0;
 		DemoSettings.disableCripplingShrines = ReadByte(in) != 0;
+		DemoSettings.enableSharedExperience = ReadByte(in) != 0;
 		DemoSettings.numHealPotionPickup = ReadByte(in);
 		DemoSettings.numFullHealPotionPickup = ReadByte(in);
 		DemoSettings.numManaPotionPickup = ReadByte(in);
@@ -167,6 +169,7 @@ void WriteSettings(FILE *out)
 	WriteByte(out, *sgOptions.Gameplay.showItemLabels);
 	WriteByte(out, *sgOptions.Gameplay.autoRefillBelt);
 	WriteByte(out, *sgOptions.Gameplay.disableCripplingShrines);
+	WriteByte(out, *sgOptions.Gameplay.enableSharedExperience);
 	WriteByte(out, *sgOptions.Gameplay.numHealPotionPickup);
 	WriteByte(out, *sgOptions.Gameplay.numFullHealPotionPickup);
 	WriteByte(out, *sgOptions.Gameplay.numManaPotionPickup);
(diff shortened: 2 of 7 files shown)


FishGaming:
I REALLY hope this does get approved into the main game!

DakkJaniels:
Doing this the "right" way, it should probably be a game option when making the game or something. If someone has this on, and another person doesn't, could that cause the monster to give out more experience than it normally would? Person who doesn't have it on get's full experience, then the person who has it on gets whatever proportion they would get?

kphoenix137:
> Doing this the "right" way, it should probably be a game option when making the game or something. If someone has this on, and another person doesn't, could that cause the monster to give out more experience than it normally would? Person who doesn't have it on get's full experience, then the person who has it on gets whatever proportion they would get?

It is

FishGaming:
Well I tested this and it works! Thing is the experience is dramatically lower. The main game gives someone 20-30 xp for each kill in the first dungeon level while with only one other person in the group (total of 2) we are only getting 5-10 xp per kill. That would mean that for each person that is added it would get to the point where we would only be getting a tenth each of xp resulting in much more of a grind in the game and changing the gameplay experience a considerable amount. In your lines of changes, is there a way to change it so that each person would have the original amount of experience per kill? If so I would like to attempt to compile the mod that way.

kphoenix137:
I'm going to revise it for accuracy. Also diablo has code that reduces exp gain based on the difference between character level and monster level which is untouched.

DakkJaniels:
> It is

Ah ok, my bad.

kphoenix137:
Should I take this a step further to make it copy the d2 exp sharing functionality people keep requesting and make experience sharing only work within a certain distance? The person who killed the monster would get a share of experience always, and it would check for any players within a set distance of the monster. If any players exist in range, then their levels are factored into the experience share and they get a cut.

Trihedraf:
The difference is that d2 is five big worlds, whereas each d1 level isn’t really that big.

kphoenix137:
> The difference is that d2 is five big worlds, whereas each d1 level isn’t really that big.

But diablo 1 goes 16 stories deep in the ground where most of d2 doesn't go nearly that deep

yuripourre:
@kphoenix137 my 2 cents. In don't think people would mind sharing if other players are in the same level.

kphoenix137:
> @kphoenix137 my 2 cents. In don't think people would mind sharing if other players are in the same level.

I think it may be more of a matter of the idea of a party that's working together are all gaining experience from being in the same combat situation. In my opinion, receiving exp from monsters killed by others on the complete other side of the level doesn't make sense in practical terms.

AJenbo:
that would also encourage people to not split up

kphoenix137:
I'd like to share some info as well that was discussed in discord. 

So there are 4 main functionalities involved with this:
* player has to hit monster
Vs
* player has to be in level

* player can be anywhere in level
Vs
* player needs to be within a distance of the monster

* player friendly/Hostile status doesn't matter
* player Hostile players cannot be involved with experience sharing

* experience is split evenly between players involved
Vs
* experience is weighted based on player level and the total level of all players involved

The first option for all 4 of these are derived from diablo 1, and second option for all 4 of these things are directly derived from diablo 2. The default toggle of course would be all the diablo 1 options, and we could additionally have a mix of d1 and d2 options, or maybe even multiple choices to toggle for handling experience share that mix and match d1 and d2 functionalities, although I think it would be far less complicated for players to settle on just the default and one alternative option. Personally, I think the best way to go is all diablo 2 functionality choices for all 4 functionalities, and I'll explain why in detail.

1. Players should receive experience if they are in the level instead of having to hit the monster.

This one is the obvious decision as it is the entire reason for implementing this PR.

2. Players should be within a certain distance of the monster to receive a share of experience.

Sharing experience is a facet of co-operative gameplay, and allowing players to receive experience from monsters killed by others somewhere far away doesn't promote cooperation and teamwork. Logically, it doesn't make sense either. If you are not actively involved in a combat situation with a group, how would your player become more skilled and powerful from someone else's far away combat? Allowing players to gain experience while far away also promotes a meta where low level players may afk in a safe area and allow a teammate to kill everything so they can passively soak up experience. The idea here is not to restructure the game meta, but rather to allow players to play how they currently do, but benefit more by not having to intentionally go out of their way to allow teammates to get a single hit in so they don't miss out on experience.

3. Players who are Hostile should not share experience nor have experience shared with them.

This functionality isn't as impactful and important as the others, however it makes sense and allows a way to prevent experience sharing without having to exit the game and toggle the vanilla setting.

4.

ikonomov:
From a gamer's perspective I see an issues with both approaches.
Vanilla:  With no sharing there's the issue that Sorcerer is able to gain much more experience than other classes.  The real problem is the damage output, however, not really the experience being gained.  If the damage was low for area of effect spells then I can see why this would be an issue, but currently Sorcerer can do a massive amount of damage AND do it to multiple enemies at the same time.  The experience sharing alone I don't think can really improve how bad a Sorcerer is when it comes to coop gameplay quality.
Sharing: Unless somebody is actually attacking there's no real reason for them to get experience.  About the only other thing they can be doing is using Healing Other spell, and that alone doesn't really make a compelling case for it.

All things considered, I don't really think there's a big advantage to be gained here.  I've played plenty of coop games from level 1 to level 50 with all three classes playing together with players using all three classes.  Up to around level 30 this is a non-issue.  Only after about level 30 spells for Sorcerer and to a lesser extend Rogue allows the scales to tip where this becomes somewhat of a problem.  However, it's not really that big of a problem because there's plenty of monster combination in Hell/Hell that allow the Warrior to shine and somewhat balance things.  Maybe the way D2 has done it with sharing is subjectively better, but in my opinion it does change the tone of cooperative gameplay, something that might be worth preserving considering the limited gains to be had for the types of characters and coop gameplay that is available in the game.

FitzRoyX:
>The experience sharing alone I don't think can really improve how bad a Sorcerer is when it comes to coop gameplay quality.

Bingo. What's fun about Diablo is killing stuff, not watching your hero level. Lategame rogue/sorc will kill almost everything before a warrior/barb can slowly walk up to it. D2 heroes have more aoe and move way faster relative to missiles. Major (mod) changes would have to be made to make D1 mp fun for mixed class groups. Passive xp gain for melee classes is spitting in the ocean.

AJenbo:
How does that make this a bad change. It would have to start somewhere. If you don't like the balance of the game I would suggest you make a mod since that isn't the goal of DevilutionX it's off topic.

ikonomov:
If you see this change as a start of something I'd be really curious to learn where you think it is going.  My problem is that I see it as a fairly significant change that can offer very little in terms of improvement to the gameplay.  9 points for change, 1 point for improvement.  There are an infinite number of changes that can be done to improve the game, the question is what we want that game at the end of those changes.

ikonomov:
And nobody is talking about liking or not liking the balance of the game.  I simply compared the classes in terms of their damage potential to point out that if that's one of the justifications for this change, it is simply not a very good one.

ikonomov:
I just re-read the first post and the follow-ups.  I think I wrongly jumped to the conclusion that this was going to be implemented without a toggle.  If so I have to correct myself, since I think this would be a great addition to the game assuming it is implemented with a toggle.

PR #6300, opened 2023-07-01
Rename: Loopback -> Offline

kphoenix137:
Change Loopback to Offline (visually for end-user) and revise Automap text

Credit goes to @ikonomov for this suggestion

### DIFF
--- a/Source/multi.cpp
+++ b/Source/multi.cpp
@@ -62,7 +62,7 @@ uint8_t gbDeltaSender;
 bool sgbNetInited;
 uint32_t player_state[MAX_PLRS];
 Uint32 playerInfoTimers[MAX_PLRS];
-bool isLoopback;
+bool IsLoopback;
 
 /**
  * Contains the set of supported event types supported by the multiplayer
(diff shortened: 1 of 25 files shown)


Translations/bg.po
```diff
@@ -367,8 +367,8 @@ msgstr "Отказ"
 msgid "Client-Server (TCP)"
 msgstr "Клиент-сървър (TCP)"
 
-#: Source/DiabloUI/selconn.cpp:15
-msgid "Loopback"
+#: Source/DiabloUI/selconn.cpp:15 # Needs updated translation
```
AJenbo, line 370: Translations should be marked as fussy, this looks like a manually added comment?

kphoenix137: Not sure what you mean. Yeah I added those comments manually

AJenbo: https://til.hashrocket.com/posts/wcmc04nort-fuzzy-translation-key-merging-with-gettext

Source/multi.cpp
```diff
@@ -62,6 +62,7 @@ uint8_t gbDeltaSender;
 bool sgbNetInited;
 uint32_t player_state[MAX_PLRS];
 Uint32 playerInfoTimers[MAX_PLRS];
+bool isLoopback;
```
AJenbo, line 65: Globals starts with capitals:
```suggestion
bool IsLoopback;
```

Preferably there would be a way to check the network type, but I'm not sure if that is possible and probably a bit hairy to get working. So if you correct this bit I think it's good enough to merge.

kphoenix137: Got it!

PR #6301, opened 2023-07-01
Rename Loopback to Offline and hide in-game info

ikonomov:
Rename Loopback to Offline to help make the menu option a little more direct and clear to the end user. Also the in-game info overlay for Loopback doesn't need a Password info and the name of the game.

LoopbackGame still needs a definition


ikonomov:
https://github.com/diasurgical/devilutionX/pull/6300

PR #6310, opened 2023-07-05
Consume Resource After Successful Spell Usage

kphoenix137:
When using any spell/scroll/charge that has a cursor, resources do not get consumed until the spell successfully is used.

Currently has a problem that allows you to still cast a cursor mana-consuming spell if your mana drops below the mana cost for the spell and you already have the spell active.

### DIFF
--- a/Source/spells.cpp
+++ b/Source/spells.cpp
@@ -212,24 +212,29 @@ void CastSpell(int id, SpellID spl, int sx, int sy, int dx, int dy, int spllvl)
 {
 	Player &player = Players[id];
 	Direction dir = player._pdir;
+
 	if (IsWallSpell(spl)) {
 		dir = player.tempDirection;
 	}
 
 	bool fizzled = false;
 	const SpellData &spellData = GetSpellData(spl);
+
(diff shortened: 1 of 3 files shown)


kphoenix137:
I need help with the time demo

StephenCWills:
It's broken because your PR is broken. Go cast a Heal scroll and then try to do something after.

kphoenix137:
Any brave souls willing to playtest this? Also I cannot figure out how to correctly handle oils. Any help would be appreciated. The problematic line has `// FIXME` by it

Source/spells.cpp
```diff
@@ -212,25 +212,28 @@ void CastSpell(int id, SpellID spl, int sx, int sy, int dx, int dy, int spllvl)
...
 		for (int i = (spllvl / 2) + 3; i > 0; i--) {
 			Missile *missile = AddMissile({ sx, sy }, { dx, dy }, dir, MissileID::ChargedBolt, TARGET_MONSTERS, id, 0, spllvl);
 			fizzled |= (missile == nullptr);
 		}
 	}
-	if (!fizzled) {
+
+	if (!fizzled)
 		ConsumeSpell(player, spl);
-	}
```
qndel, line 233: imo avoid adding empty diff like in this file - changes not really relevant to the PR

Issue #6271, opened 2023-06-26
Books wont update magic requirement after read 1


AdriaTheWitch1:
### Operating System

Windows x64

### DevilutionX version

1.5.0 (latest release)

### Describe

if u have 2x same bookspell in inventory. if u read 1 the other wont update the magic requirement, u have to update interface for change.
IE; i had 2healing book with spell lvl0(req 20magic to read) . i readed 1 but still the other book wasonly req 20magic until i update interface with in/out book from inv

### To Reproduce

savefile here [multi_0.sv.zip](https://github.com/diasurgical/devilutionX/files/11873829/multi_0.sv.zip)
if u open the save, u will have 2x holy bolt book that require 20magic to read
if u read one , the other book magic requirement wont change n will stay at 20magic till u update change/move something in your inventory

### Expected Behavior

if i read i book, i expect the next book samespell to immediatly update magic requirement

AdriaTheWitch1:
just notice another bug with book refreshing magic requirement, its about same ,
 if i have books in inventory and i take enchanted shrine, books dont refresh requirement in inventory till inv refresh. 
 havent tested for ornate shrine, but i suppose its do same for holy bolt book.

PR #6312, opened 2023-07-05
Fix Book requirements not updating

kphoenix137:
When you read a Book, it attempts to update the Spell Level by sending a network command, and then attempts to recalculate the Magic requirement. The network command doesn't resolve the Spell Level before the function calls `updateRequiredStatsCacheForPlayer()`, resulting in the incorrect requirement being displayed until this function gets called again at a later time, observed by the call to `CheckInvCut()` in the attached issue report. To solve this, I added a line to locally change the Spell Level to `newSpellLevel`, which will be overwritten anyway when the network command resolves.

Fixes: https://github.com/diasurgical/devilutionX/issues/6271

### DIFF
--- a/Source/items.cpp
+++ b/Source/items.cpp
@@ -4100,8 +4100,10 @@ void UseItem(size_t pnum, item_misc_id mid, SpellID spellID, int spellFrom)
 		break;
 	case IMISC_BOOK: {
 		uint8_t newSpellLevel = player._pSplLvl[static_cast<int8_t>(spellID)] + 1;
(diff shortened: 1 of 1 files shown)


AJenbo (approved):

AdriaTheWitch1:
just notice another bug with book refreshing magic requirement, its about same ,
 if i have books in inventory and i take enchanted shrine, books dont refresh requirement in inventory till inv refresh. 
 havent tested for ornate shrine, but i suppose its do same for holy bolt book.

AJenbo:
K plz create a new issue

PR #6314, opened 2023-07-06
Fix inventory alignment

kphoenix137:
* Fixed item alignment in bottom inventory (items and back slot colors were shifted 1 px down)
* Refactored `DrawInv()` to remove hardcoded tables and add `Point` where needed
* Improved the back slot color coverage in equipment slots
* Added a function to adjust the sprites of graphics that have poor horizontal alignment

### DIFF
--- a/Source/inv.cpp
+++ b/Source/inv.cpp
@@ -1066,11 +1066,7 @@ void InitInv()
 		break;
 	}
 }
-/**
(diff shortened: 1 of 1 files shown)


kphoenix137:
Will need to call the sprite adjustment function for stash as well

PR #6322, opened 2023-07-07
Fix Book requirements not updating from Objects (Part 2)

kphoenix137:
Builds upon https://github.com/diasurgical/devilutionX/pull/6316 where I mistakenly forgot to refresh item requirements. The other PR updates the player's Spell Level immediately which is necessary for updating magic requirements, but then we need to iterate all inventory and stash items to update requirements.

### DIFF
--- a/Source/objects.cpp
+++ b/Source/objects.cpp
@@ -35,6 +35,7 @@
 #include "missiles.h"
 #include "monster.h"
 #include "options.h"
+#include "qol/stash.h"
 #include "stores.h"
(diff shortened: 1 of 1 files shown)


qndel:
I think it could be worth it to move it to a separate function

kphoenix137:
> I think it could be worth it to move it to a separate function

I'll do that

AJenbo (approved):
Please make a follow up to make a helper to avoid the duplicate code

PR #6323, opened 2023-07-08
Only warp to Lazarus' chamber when red portal is active

StephenCWills:


AJenbo (approved):
Crazy this thing was so broken :D

Issue #4656, opened 2022-06-04
Sparking mail breaks thunderclap's effect


qndel:
**Important information**
1.4.0 / https://github.com/diasurgical/devilutionX/commit/0a71f1ffb39d3a3b8ae4e5c21f2fa96076b4daf6

**Describe**
If you are wearing sparking mail, thunderclap won't fire a charged bolt in the direction of enemy that you attacked but will fire a holy-bolt looking missile north-west instead.

**To Reproduce**
1. Equip sparking mail
2. Equip thunderclap
3. Hit an enemy

**Expected behavior**
Keep the original effect and maybe add sparking mail's damage to charged bolts?

**Screenshots**
![thunderclap](https://user-images.githubusercontent.com/14297035/171994208-3eef45c5-5983-4aca-8311-a3d6e365a69d.gif)

**Additional context**
Missile looks like holy bolt but hits everything so it doesn't seem like it's actually holy bolt

wkdgmr:
Ya know... what is actually interesting is HolyBoltBow is explicitly coded to be picked up on switch 3 in the case/switch in ProcessSpectralArrow().  And there is no other way to access the HolyBoltBow logic than having Thunderclap and Sparking Mail.  I think this was an intended easter egg by the Hellfire devs.

Damn... very cool if that was the case.   I just realized this now I thought it was unfinished and not implemented.

PR #6327, opened 2023-07-09
Bugfix: Missiles calculating damage in Process functions instead of Add functions

kphoenix137:
Moves damage calculations from the Process functions into the Add functions, so that damage is added to the missile struct rather than calculated every time the Process functions are called, which can also lead to undefined behavior if the source of the missile dies or leaves the game.

Also fixes: https://github.com/diasurgical/devilutionX/issues/4656

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1953,6 +1953,20 @@ void AddLightningControl(Missile &missile, AddMissileParameter &parameter)
 	UpdateMissileVelocity(missile, parameter.dst, 32);
 	missile._miAnimFrame = GenerateRnd(8) + 1;
 	missile._mirange = 256;
+
+	switch (missile.sourceType()) {
+	case MissileSource::Player: {
+		const Player &player = *missile.sourcePlayer();
+		missile._midam = (GenerateRnd(2) + GenerateRnd(player._pLevel) + 2) << 6;
+	} break;
+	case MissileSource::Monster: {
+		const Monster &monster = *missile.sourceMonster();
+		missile._midam = 2 * (monster.minDamage + GenerateRnd(monster.maxDamage - monster.minDamage + 1));
+	} break;
+	case MissileSource::Trap:
+		missile._midam = (2 * currlevel) + GenerateRnd(currlevel);
+		break;
+	}
 }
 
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -1393,11 +1393,10 @@ void AddStealMana(Missile &missile, AddMissileParameter & /*parameter*/)
 
 void AddSpectralArrow(Missile &missile, AddMissileParameter &parameter)
 {
+	Player &player = *missile.sourcePlayer();
```
glebm, line 1396: This dereferences a null pointer if `missile.sourceType() == MissileSource::Player`.
This line should be moved back to where it was before.

Source/missiles.cpp
```diff
@@ -1396,7 +1396,7 @@ void AddSpectralArrow(Missile &missile, AddMissileParameter &parameter)
 	int av = 0;
 
 	if (missile.sourceType() == MissileSource::Player) {
-		const Player &player = *missile.sourcePlayer();
+		Player &player = *missile.sourcePlayer();
```
glebm, line 1399: Why remove the `const`? I don't think `player` is modified her?

kphoenix137: Can't iterate through equipment if it's const

kphoenix137: Would it be better to use a traditional for loop to retain player as const?

glebm: Ah, right, this will be easy to fix once we have C++17 but for now non-const is fine.

kphoenix137: I'll add a comment by it so we know to readd const at a later time

Source/missiles.cpp
```diff
@@ -1956,10 +2033,19 @@ void AddMissileExplosion(Missile &missile, AddMissileParameter &parameter)
 void AddWeaponExplosion(Missile &missile, AddMissileParameter &parameter)
 {
 	missile.var2 = parameter.dst.x;
-	if (parameter.dst.x == 1)
+
+	const Player &player = *missile.sourcePlayer();
```
kphoenix137, line 2037: Need to verify that missile source is player first before defining player

PR #6336, opened 2023-07-10
Calculate Missile damage from Items

kphoenix137:
Prevents read/writing to the lightning and fire damage member variables in `Player` struct, and calculates lightning and fire damage by iterating the items and directly writing to the `Missile` struct. Based on https://github.com/diasurgical/devilutionX/pull/6327

### DIFF
--- a/Source/missiles.cpp
+++ b/Source/missiles.cpp
@@ -1953,6 +1953,20 @@ void AddLightningControl(Missile &missile, AddMissileParameter &parameter)
 	UpdateMissileVelocity(missile, parameter.dst, 32);
 	missile._miAnimFrame = GenerateRnd(8) + 1;
 	missile._mirange = 256;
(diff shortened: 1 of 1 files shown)


Source/missiles.cpp
```diff
@@ -1956,10 +2054,36 @@ void AddMissileExplosion(Missile &missile, AddMissileParameter &parameter)
 void AddWeaponExplosion(Missile &missile, AddMissileParameter &parameter)
 {
 	missile.var2 = parameter.dst.x;
-	if (parameter.dst.x == 1)
+
+	Player &player = *missile.sourcePlayer();
```
kphoenix137, line 2058: Move this into next if block scope

kphoenix137: Sorry

PR #6341, opened 2023-07-10
deduplicate NPC sell/buy code

qndel:


qndel:
This changes the code a bit:
adds
```cpp
} else {
			l -= 4;
		}
```
to places that didn't have it before + different value depending on if the item is identified or not, but I don't expect it to cause any problems - stores only sell identified items after all

Source/stores.cpp
```diff
@@ -348,6 +348,32 @@ bool StoreAutoPlace(Item &item, bool persistItem)
 	return AutoPlaceItemInInventory(player, item, persistItem);
 }
 
+void ScrollVendorStore(Item *itemData, int storeLimit, int idx, int selling = true)
```
Iniquitatis, line 351: Um, `int selling`? Why not `bool`?

qndel: > Um, `int selling`? Why not `bool`?

I think I was messing with other parameters and this must've been a leftover, nice catch ^^

PR #6345, opened 2023-07-11
Fixes testing characters from Hellfire showing up in Diablo

viniciusferrao:
This commit restricts the use of Test Barbarian and Test Barb to the Hellfire expansion only.


qndel:
No need to disable what we've enabled on purpose...

viniciusferrao:
Hi @qndel I didn't realise that was on purpose. It seemed strange to me that I can use test characters from Hellfire in the non-expansion Diablo game.

If that's on purpose may I ask if the Monk can be enabled also? I know that are missing assets, but we can just use mage assets instead? If that's the spirit of using this characters on the game without expansion.

I'm not questioning, just curious about the reasoning.

DakkJaniels:
Just using the Sorcerer's assets isn't that simple, but feel free to implement it and submit a PR 🙂

AJenbo:
Just fyi it's not for sure that we would merge a PR for something like that. It would have to be very minimalistisk at the least.

PR #6349, opened 2023-07-12
Re-encode CL2 on load

glebm:
Original Blizzard encoder is slightly less optimal than our encoder.

While size in RAM in less of a concern for the non-`UNPACKED_MPQS` build,
it is still a benefit because smaller files are faster to render.

Savings for unpacked and minified MPQs:
* diabdat.mpq: 918,311 bytes.
* hellfire.mpq: 313,882 bytes.

Example player graphics (note that only a few are loaded at any given time for single player):
* diabdat/plrgfx/warrior/: 366,564 bytes.

Example monster graphics savings:

* diabdat/monsters/skelbow: 5,391 bytes.

Based on the implementation from https://github.com/diasurgical/devilutionx-graphics-tools/pull/6

### DIFF
--- a/Source/engine/load_cl2.hpp
+++ b/Source/engine/load_cl2.hpp
@@ -61,13 +61,13 @@ OwnedClxSpriteSheet LoadMultipleCl2Sheet(tl::function_ref<const char *(size_t)> 
 			FailedToOpenFileError(paths[i].data(), handle.error());
 		}
 		WriteLE32(&data[i * 4], accumulatedSize);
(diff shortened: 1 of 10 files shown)


AJenbo (approved):

glebm:
Oh, didn't see the 1.6.0 milestone, sorry about the merge. We can revert this if needed.

AJenbo:
i prefer not to revert, but let's hope it doesn't expload.

glebm:
Should be fine hopefully

PR #6351, opened 2023-07-12
Move hp/mana display and item graphics to gameplay options

DakkJaniels:
Experience bar, enemy health bar, floating numbers, etc. are all under gameplay, these options should be as well.


AJenbo (approved):
Wanted to do this last minute for 1.5.0 but gave up on it since it's all a bit complicated. Thanks.

PR #6354, opened 2023-07-13
Fix iCreateInfo comment

kphoenix137:


DakkJaniels:
But numbers were correct before, but referenced from 0 I think, weren't they? First part should have said bits 0 through 5 instead of first 6 bits, to be a little less confusing.

AJenbo (approved):

PR #6363, opened 2023-07-14
Log validation errors in UnPackNetPlayer()

StephenCWills:
Implementation for the `ValidateField()` and `ValidateFields()` macros was heavily inspired by the implementation for the `assert()` macro.

**Example log message:**
```
DEBUG: Remote player validation failed: ValidateFields(player._pStrength: 15, SDL_SwapLE32(packed.pStrength): 250, player._pStrength == SDL_SwapLE32(packed.pStrength))
```

**Example player notification:**
image

One additional thought I had just before submitting this PR was to move `CopyUtf8(player._pName, packed.pName, sizeof(player._pName))` to the top of the `UnPackNetPlayer()` function so it can be used in `EventFailedJoinAttempt()` in an attempt to show the name of the player that transmitted invalid data. Does anyone else think that might be a good idea?

### DIFF
--- a/Source/pack.cpp
+++ b/Source/pack.cpp
@@ -21,7 +21,7 @@
 	do {                                                           \
 		if (!(condition)) {                                        \
 			LogFailedJoinAttempt(#condition, #logValue, logValue); \
-			EventFailedJoinAttempt();                              \
+			EventFailedJoinAttempt(player._pName);                 \
 			return false;                                          \
 		}                                                          \
 	} while (0)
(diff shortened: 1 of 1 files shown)


Source/pack.cpp
```diff
@@ -11,10 +11,20 @@
 #include "init.h"
 #include "loadsave.h"
 #include "playerdat.hpp"
+#include "plrmsg.h"
 #include "stores.h"
 #include "utils/endian.hpp"
+#include "utils/log.hpp"
 #include "utils/utf8.hpp"
 
+#define ValidateField(logValue, condition)                                          \
```
glebm, line 20: Consider using the `do while(0)` trick rather than the comma operator:

```c++
#define ValidateField(logValue, condition) \
  do { \
    if (!condition) { \
      LogFailedJoinAttempt(#condition, #logValue, logValue); \
      return false; \
    } \
  } while (0)
```

It's a bit more verbose but perhaps easier to read.

AJenbo:
Yeah the change you made previously makes it safe to have the name be copied at any point, so makes sens that we copy it first.

StephenCWills:
**New event message:**
image

AJenbo (approved):

PR #6364, opened 2023-07-14
Fonts: Use pre-cropped CLX sprites

### LINKED ISSUE #6361: Replace font `.bin` files with CLX frame widths
CLX natively supports frame widths so we can get rid of the `.bin` files if we encode the widths into the CLX files themselves.
This will make font files smaller and will improve rendering performance.

glebm:
No more "kerning" files and the built-in font size is decreased by 121,587 bytes (2,287,332 -> 2,165,745).

[Screencast from 2023-07-14 14-20-58.webm](https://github.com/diasurgical/devilutionX/assets/216339/b71a9362-d70b-430f-8363-089a6b5a04c5)

We will need to release a new `fonts.mpq` and `devilutionx.mpq` once this is merged.

Related PRs:

* [x] https://github.com/diasurgical/devilutionx-graphics-tools/pull/7
* [ ] https://github.com/diasurgical/devilutionx-assets/pull/14

Fixes #6361


AJenbo:
All looks good, but since we had a few issues with fonts in 1.5.0, which I think are all addressed in master I would like to push this change to 1.6.0 to avoid adding more noise to the signal when we release 1.5.1.

glebm:
I think I'll use the extra time to add a VERSION file for fonts.mpq, because the way we do out-of-date detection currently is far too complicated.

AJenbo:
If you can spare a thought for how that will work with Android that would be much appreciated :sweat_smile:

glebm:
Updated the Android code. It is still file size based, but someone more familiar with Android could implement a bit of code that asks C++ from Java whether the fonts are up to date.

android-project/app/src/main/java/org/diasurgical/devilutionx/DevilutionXSDLActivity.java
```diff
@@ -79,11 +79,12 @@ private boolean isMissingGameData() {
 		if (lang.startsWith("ru") && !fileManager.hasFile("ru.mpq"))
 			return true;
 		if (lang.startsWith("ko") || lang.startsWith("zh") || lang.startsWith("ja")) {
-			if (!fileManager.hasFile("fonts.mpq"))
+			if (!fileManager.hasFile("fonts.mpq") ||
+					fileManager.fileSize("fonts.mpq") == 70471463 /* v1 */ ||
+					fileManager.fileSize("fonts.mpq") == 53991069 /* v2 */ ||
+					fileManager.fileSize("fonts.mpq") == 58488019 /* v3 */) {
```
StephenCWills, line 85: `fonts.mpq` version was deliberately checked regardless of language settings because an out-of-date file needs to be updated regardless.

glebm:
@StephenCWills added JNI-based version checking in https://github.com/glebm/devilutionX/pull/8 and I've deduplicated the checking code in a subsequent commit.

Source/platform/android/android.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include <jni.h>
+
+namespace devilution {
+namespace {
+
+bool AreExtraFontsOutOfDateForMpqPath(const char *mpqPath)
+{
+	int32_t error = 0;
+	std::optional<MpqArchive> archive = MpqArchive::Open(mpqPath, error);
+	return error == 0 && archive && !AreExtraFontsOutOfDate(*archive);
```
StephenCWills, line 13:
```suggestion
	return error == 0 && archive && AreExtraFontsOutOfDate(*archive);
```

Source/init.cpp
```diff
@@ -169,8 +171,48 @@ std::vector<std::string> GetMPQSearchPaths()
...
+bool CheckExtraFontsVersion(AssetRef &&ref)
+{
+	const size_t size = ref.size();
+	AssetHandle handle = OpenAsset(std::move(ref), false);
+	if (!handle.ok())
+		return true;
+
+	std::unique_ptr<char[]> version_contents { new char[size] };
+	if (!handle.read(version_contents.get(), size))
+		return false;
```
StephenCWills, line 183: Is it better to assume the fonts MPQ is out of date if we're unable to read from the VERSION file? We do the same for if the asset handle is not okay.
```suggestion
	std::unique_ptr<char[]> version_contents { new char[size] };
	if (!handle.read(version_contents.get(), size))
		return true;
```

AJenbo:
Well now this is starting to look more appealing then what we have...

AJenbo (dismissed):

Source/platform/android/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+include(functions/devilutionx_library)
+add_devilutionx_object_library(libdevilutionx_android android.cpp)
+
+target_link_libraries(libdevilutionx_android PUBLIC
+  DevilutionX::SDL
+  SDL_audiolib::SDL_audiolib
+)
```
StephenCWills, line 7: Thanks to the changes you made to remove duplicate code, `android.cpp` no longer includes `engine/assets.hpp` which was the reason why I had to link `SDL_audiolib`.
```suggestion
target_link_libraries(libdevilutionx_android PUBLIC DevilutionX::SDL)
```

glebm:
@AJenbo Do you mean to get it in for 1.5.1?

AJenbo:
Not initially but the better version checks for Android changed my min, since the size checking is one of the things that kept me up at night.

AJenbo:
Damit, I keep forgetting to edit the changelog within the PR before merging...

glebm:
I've created a pre-release for devilutionx-assets: v4 https://github.com/diasurgical/devilutionx-assets/releases/tag/v4
We can just mark it as the latest release once 1.5.1 is out

